PR #3301, opened 2018-03-15
style(meta) use 'or' instead of ternary operator

kidd:
### Summary

Style fix . change a ternary operation by a simple "or" keeping the same behavior

### Full changelog

meta.lua - change a ternary operation by an "or"


thibaultcha:
@kidd Thanks!

PR #3309, opened 2018-03-16
chore(admin) remove unneeded code from endpoints.generate_endpoints

bungle:
### Summary

Just removes a few lines of code that were not used for anything. Something that was left there before that part of code was refactored (it used to be recursive).


thibaultcha:
Nice!

PR #3310, opened 2018-03-17
feat: check database version compatibility

thibaultcha:
Add a check to the DAO's `init()` method to ensure the compatibility of the underlying database (PostgreSQL or Cassandra). This check is ran:
1. In the `kong start` command
2. In `init_by_lua`

We support "minimum recommended version", and an _optional_ "oldest deprecated version", which as of today and 0.12.3 are:

PostgreSQL
- current minimum recommended: 9.5
- deprecated: 9.4

Cassandra
- current minimum recommended: 2.2
- deprecated: 2.1

---

When using a deprecated datastore, the CLI shows:
```
$ bin/kong start                 
2018/03/16 17:38:18 [warn] Currently using PostgreSQL 9.4 which is considered deprecated, please use 9.5 or greater
Kong started
```

It is also shown in the nginx logs when starting Kong from `nginx` (e.g. Docker images).

---

When using a version that is below the minimum recommended, and below the oldest deprecated one, we error out from the CLI (and from `init_by_lua`):
```
$ bin/kong start
Error: Kong requires PostgreSQL 9.5 or greater (currently using 9.4)

```

### DIFF
--- a/kong/dao/factory.lua
+++ b/kong/dao/factory.lua
@@ -1,6 +1,11 @@
 local DAO = require "kong.dao.dao"
+local log = require "kong.cmd.utils.log"
 local utils = require "kong.tools.utils"
+local version = require "version"
+local constants = require "kong.constants"
 local ModelFactory = require "kong.dao.model_factory"
+
+local fmt = string.format
 
 local CORE_MODELS = {
   "apis",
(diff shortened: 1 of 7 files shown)


bungle:
Should there be some way to start Kong in case the version is below the oldest deprecated one. I mean I would not want to end up with situation with many old Windows software installers that just refuse to install because of a faulty check (even if the software was just fine with newer version of Windows). Also there might be PostgreSQL / Cassandra compatible client interface in some other databases that just report something different as a version.

thibaultcha:
> Should there be some way to start Kong in case the version is below the oldest deprecated one.

I would be wary of potentially corrupting the database, but well, if the option is obvious enough that it implies danger, why not, we could.

> Also there might be PostgreSQL / Cassandra compatible client interface in some other databases that just report something different as a version

Good point, especially with implementations like ScyllaDB.

Maybe a `--ignore-db-compat` flag? Or `no-db-check`?

kong/dao/factory.lua
```diff
@@ -128,8 +133,50 @@ function _M.new(kong_config)
...
 
+function _M:check_version_compat(min, deprecated)
+  local db_infos = self:infos()
+  if db_infos.version == "unknown" then
+    return nil, "could not check database compatibility: version " ..
+                "is unknown (did you call ':init'?)"
+  end
+
+  local db_v = version.version(db_infos.version)
+  local min_v = version.version(min)
```
Tieske: you can update the 'version.lua' library to 1.0 and change those into `version(min)` without the double 'version' name

thibaultcha: Yes, I saw; but as usual when possible, we'll make that a separate contribution

PR #3311, opened 2018-03-17
fix(*) use dedicated shm for rate-limiting plugins

### LINKED ISSUE #3124: Rate Limiting plugin memory leak issue
### Summary

After set config.second property in ratelimit plugin. The second cache (e.g. dict = kong_cache, key = ratelimit:561f23d4-23d2-4283-9f4c-7c077b5c77d3:118.116.111.5:1514260849000:second, value = 1) will take all cahce memory and never expired. When new API add in or be called. No memory for new API and plugins. Then no memory error show up.

### Steps To Reproduce

1. vim kong.conf
mem_cache_size = 100k
2. kong start --vv
3. create 6 APIs
4. Add ratelimit plugin set second 
$ curl -X POST http://kong:8001/apis/{api}/plugins \
    --data "name=rate-limiting" \
    --data "config.second=100" \
    --data "config.hour=10000"
5. keep calling the first 5 APIs.
6. After few minutes with no error.
7. call 6th API. you will see the error in error.log 
[error] 19343#0: *2427 [lua] responses.lua:107: send_HTTP_INTERNAL_SERVER_ERROR(): failed to get from node cache: could not write to lua_shared_dict: no memory, client: 118.116.111.5, server: kong, request: "GET /V1/ben_test_1221?auth_key=097C993DF185099B2B7633F4C1778023&t=1514259888796 HTTP/1.1", host: "ec2-13-56-255-95.us-west-1.compute.amazonaws.com:8000"

### Additional Details & Logs

- Kong version (0.11.0)
- Kong debug-level startup logs
[startup.log](https://github.com/Kong/kong/files/1586503/startup.log)
[error.log](https://github.com/Kong/kong/files/1586509/error.log)
- Kong configuration (registered APIs/Plugins & configuration file)
- Operating System
ubuntu 16.04 LTS

### LINKED ISSUE #3241: Rate Limiting plugin memory consumption in "local" strategy
https://github.com/Kong/kong/issues/3124 was closed.
I'd like to submit a test which demonstrates the memory leak.
thibaultcha is this okay?

Best,
Dave

thibaultcha:
This is part of a series of fixes:
- https://github.com/thibaultcha/lua-resty-mlcache/pull/41
- https://github.com/thibaultcha/lua-resty-mlcache/pull/42
- https://github.com/Kong/kong/pull/3311
- https://github.com/Kong/kong/pull/3341

Context
-------
In the `local` mode of the rate-limiting plugins, storing the
rate-limiting counters in the same shm used by Kong's database cache is
too invasive for the underlying shm, especially when the rate-limiting
plugins are used with a `seconds` precision.

On top of exhausting the database cache slots, this approach also
generates some form of fragmentation in the shm. This is due to the
side-by-side storage of values with sizes of different orders of
magnitude (JSON strings vs. an incremented double) and the LRU eviction
mechanism. When the shm is full and LRU kicks-in, it is highly probable
that several rate-limiting counters will be evicted (due to their
proliferation), thus not freeing enough space to store the retrieved
data, causing a `no memory` error to be reported by the shm.

Solution
--------
Declaring shms that are only used by some plugins is not very elegant.
Now, all users (even those not using rate-limiting plugins) have to pay
a memory cost (although small).
Unfortunately, and in the absence of a more dynamic solution to shm
configuration such as a more dynamic templating engine, or a
`configure_by_lua` phase, this is the safest solution.

Size rationale
--------------
Running a script generating similar keys and storing similar values
(double) indicates that an shm with 12Mb should be able to store about
~48,000 of those values at once. It is important to remind ourselves
that one Consumer/IP address might use more than one key (in fact, one
per period configured on the plugin), and both the rate-limiting and
response-ratelimiting plugins at once, and they use the same shms.

Even considering the above statements, ~48,000 keys per node seems
somewhat reasonable, considering keys of `second` precision will most
likely fill up the shm and be candidates for LRU eviction. Our concern
lies instead around long-lived limits (and thus, keys) set by the user.

Additionally, a future improvement upon this will be the setting of the
`init_ttl` argument for the rate-limiting keys, which will help **quite
considerably** in reducing the footprint of the plugins on the shm. As
of this day, this feature has been contributed to ngx_lua but not
released yet:

openresty/lua-nginx-module#1226

See also
--------
Another piece of the fixes for the `no memory` errors resides in the
behavior of the database caching module upon a full shm. See:

thibaultcha/lua-resty-mlcache#41

This patch reduces the likeliness of a full shm (by a lot!), but does
not remove it. The above patch ensures a somewhat still sane behavior
would the shm happen to be full again.

Fix #3124
Fix #3241

### DIFF
--- a/kong/constants.lua
+++ b/kong/constants.lua
@@ -82,5 +82,6 @@ return {
     "kong_process_events",
     "kong_cluster_events",
     "kong_healthchecks",
+    "kong_rate_limiting_counters",
   },
 }
--- a/kong/plugins/rate-limiting/policies/init.lua
+++ b/kong/plugins/rate-limiting/policies/init.lua
@@ -6,7 +6,7 @@ local reports = require "kong.core.reports"
 
 
 local ngx_log = ngx.log
-local shm = ngx.shared.kong_cache
+local shm = ngx.shared.kong_rate_limiting_counters
 local pairs = pairs
 local fmt = string.format
 
--- a/kong/plugins/response-ratelimiting/policies/init.lua
+++ b/kong/plugins/response-ratelimiting/policies/init.lua
@@ -6,7 +6,7 @@ local reports = require "kong.core.reports"
 
 
 local ngx_log = ngx.log
-local shm = ngx.shared.kong_cache
+local shm = ngx.shared.kong_rate_limiting_counters
 local pairs = pairs
 local fmt = string.format
 
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -33,6 +33,7 @@ lua_shared_dict kong_cache          ${{MEM_CACHE_SIZE}};
 lua_shared_dict kong_process_events 5m;
 lua_shared_dict kong_cluster_events 5m;
 lua_shared_dict kong_healthchecks   5m;
+lua_shared_dict kong_rate_limiting_counters 12m;
 > if database == "cassandra" then
 lua_shared_dict kong_cassandra      5m;
 > end
--- a/spec/fixtures/custom_nginx.template
+++ b/spec/fixtures/custom_nginx.template
@@ -46,6 +46,7 @@ http {
     lua_shared_dict kong_process_events 5m;
     lua_shared_dict kong_cluster_events 5m;
     lua_shared_dict kong_healthchecks   5m;
+    lua_shared_dict kong_rate_limiting_counters 12m;
 > if database == "cassandra" then
     lua_shared_dict kong_cassandra      5m;
 > end


kong/constants.lua
```diff
@@ -70,6 +70,7 @@ return {
   DICTS = {
     "kong",
     "kong_cache",
+    "kong_rl_counters",
```
bungle: I am sure you thought about it a lot, but this feature seems to make rate-limiting plugin a core feature, while it is still a plugin. It feels a bit bad that plugins inject stuff in core. About naming, I'd like `kong_rate_limiting` as `dict` name better as it pollutes core a bit less.

kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -4,7 +4,7 @@ local redis = require "resty.redis"
 local policy_cluster = require "kong.plugins.rate-limiting.policies.cluster"
 local reports = require "kong.core.reports"
 local ngx_log = ngx.log
-local shm = ngx.shared.kong_cache
+local shm = ngx.shared.kong_rl_counters
```
bungle: Other option could be adding a new `config` parameter to rate-limiting plugin where you can configure the `shm` per plugin configuration. Not sure though, should we still supply a default.

thibaultcha: With `configure_by_lua`, yep! Otherwise, it becomes harder to maintain a custom nginx configuration in sync with the plugins configured in the database...

kong/constants.lua
thibaultcha: > I am sure you thought about it a lot, but this feature seems to make rate-limiting plugin a core feature, while it is still a plugin. It feels a bit bad that plugins inject stuff in core.

Yep. This is also why the fix took so long to appear... We have little choice for now though.

 > I'd like kong_rate_limiting as dict name better as it pollutes core a bit less.

We can do that. Probably `kong_rate_limiting_counter` for completesness's sake then.

kong/plugins/rate-limiting/policies/init.lua
thibaultcha: (That or a more powerful templating system that we do not have and tbh, aren't sure of ever wanting at all, as it would make custom nginx templates harder and harder to maintain...)

Issue #2858, opened 2017-08-31
Allow JWT header `kid` to identify the credentials in use


Floby:
### Summary

Currently, the JWT plugin looks at a configured claim (`iss` by default). I would like to have the key being used to sign the token to be specified as the [`kid`](https://tools.ietf.org/html/rfc7516#section-4.1.6) header. This seems like a incomplete implementation since this header is specifically designed for this use-case. If this is a feature that has a go, I'm happy to write a PR.

### Steps To Reproduce

1. Create an API
2. Enable the JWT plugin
3. Make a request to this API with a signed JWT with a kid header.

### Additional Details & Logs

- Kong version (`0.11`)
- Kong configuration (registered APIs/Plugins & configuration file)

```yaml
host: 'localhost:8001'
headers: []
apis:
  - name: Admin
    plugins:
      - name: jwt
        attributes:
          enabled: true
          config:
            claims_to_verify: {}
            key_claim_name: kid
            anonymous: ''
            secret_is_base64: false
            uri_param_names: {}
    attributes:
      uris:
        - /admin
      strip_uri: true
      preserve_host: false
      upstream_url: 'http://nginx-lb:8001'
      retries: 5
      upstream_connect_timeout: 60000
      upstream_read_timeout: 60000
      upstream_send_timeout: 60000
      https_only: false
      http_if_terminated: false
consumers:
  - username: Floby's app
    custom_id: floby
    acls: []
    credentials:
      - name: jwt
        attributes:
          algorithm: HS256
          key: floby-sym
          secret: salut
plugins: []
```

thibaultcha:
> If this is a feature that has a go, I'm happy to write a PR.

Happy to receive one!

jfalkbjer:
Any feedback about this feature?
Right now I send the kid both in the JWT header and body. It would be great to be able to remove it from the body.

Floby:
Sorry, I don't have time anymore to look into this :/ but it didn't seem very complicated

alexandermontgomery:
Hello!
I am looking to use Kong's JWT with [Firebase](https://firebase.google.com/docs/auth/admin/verify-id-tokens). Am I right in assuming that Kongs JWT can support multiple public keys, each with its own `config.key_claim_name`? If yes, then it seems like moving the kid to the header would fix this and we can work on PR, but wanted to confirm this is actually the case.

thibaultcha:
This should have been addressed by #3313 - thank you for the suggestion!

[...]

PR #3313, opened 2018-03-19
feat(jwt) support for config.key_claim_name in header

brycehemme:
Addresses the issue discussed in #2858. 

## Context

The JWT plug-in previously only supported setting the config.key_claim_name parameter to a claim in the body and not the header. Many identity providers who use JWKS (such as Amazon Cognito) support signing their tokens with more than one public key. This results in the need to support the "kid" (key ID) claim which identifies the public key required to validate the signature of the token. Due to the JWKS defining multiple public keys, the standard "iss" (issuer) claim is not unique. The "kid" claim is typically set in the header of the token and therefore requires also checking the token header for claims. The change that was made is fully backwards compatible due to keeping the functionality that checks the token body claims first and if the config.key_claim_name is not found there it then checks the header. An additional test case was also written to confirm that setting config.key_claim_name to a claim that is in the header succeeds.

## Solution

The current release of the JWT plugin [only checks the token body (claims)](https://github.com/Kong/kong/blob/master/kong/plugins/jwt/handler.lua#L120-L125) for the config.key_claim_name parameter. While this works for most scenarios, in the aforementioned scenario of the "kid" claim in the header, this does not suffice. Adding an [additional check against the header of the token](https://github.com/brycehemme/kong/blob/master/kong/plugins/jwt/handler.lua#L121-L123) if the config.key_claim_name parameter isn't found in the body resolves this issue.

After making this change, an IdP that uses more than one public key can be configured for a single consumer as follows - 

```bash
# Configure the plugin and ensure config.key_claim_name is set to the appropriate header claim -
curl -X POST http://localhost:8001/apis/myapi/plugins --data "name=jwt" --data "config.claims_to_verify=exp" --data "config.key_claim_name=kid"

# Configure the FIRST public key for the consumer. Each of the parameters in curly braces should be replaced to fit your scenario -
curl -i -X POST http://localhost:8001/consumers/{MY_CONSUMER}/jwt -F "algorithm=RS256" -F "rsa_public_key=@{PATH_TO_MY_FIRST_PUBLIC_KEY.pem}" -F "key={MY_FIRST_PUBLIC_KEY_KID}"

# Configure the SECOND public key for the consumer. Each of the parameters in curly braces should be replaced to fit your scenario -
curl -i -X POST http://localhost:8001/consumers/{MY_CONSUMER}/jwt -F "algorithm=RS256" -F "rsa_public_key=@{PATH_TO_MY_SECOND_PUBLIC_KEY.pem}" -F "key={MY_SECOND_PUBLIC_KEY_KID}"
```

## See also

JWKS standard: https://tools.ietf.org/html/rfc7517#section-5
Related issue: https://github.com/Kong/kong/issues/2858

### DIFF
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -118,8 +118,9 @@ local function do_authentication(conf)
   end
 
   local claims = jwt.claims
-
-  local jwt_secret_key = claims[conf.key_claim_name]
+  local header = jwt.header
+  
+  local jwt_secret_key = claims[conf.key_claim_name] or header[conf.key_claim_name]
   if not jwt_secret_key then
     return false, {status = 401, message = "No mandatory '" .. conf.key_claim_name .. "' in claims"}
   end
--- a/spec/03-plugins/17-jwt/03-access_spec.lua
+++ b/spec/03-plugins/17-jwt/03-access_spec.lua
@@ -12,7 +12,7 @@ local PAYLOAD = {
 }
 
 describe("Plugin: jwt (access)", function()
-  local jwt_secret, base64_jwt_secret, rsa_jwt_secret_1, rsa_jwt_secret_2, rsa_jwt_secret_3
+  local jwt_secret, jwt_secret_2, base64_jwt_secret, rsa_jwt_secret_1, rsa_jwt_secret_2, rsa_jwt_secret_3
   local proxy_client, admin_client
 
   setup(function()
@@ -20,7 +20,7 @@ describe("Plugin: jwt (access)", function()
 
     local apis = {}
 
-    for i = 1, 9 do
+    for i = 1, 10 do
       apis[i] = assert(helpers.dao.apis:insert({
         name         = "tests-jwt" .. i,
         hosts        = { "jwt" .. i .. ".com" },
@@ -34,6 +34,7 @@ describe("Plugin: jwt (access)", function()
     local consumer3 = assert(cdao:insert({ username = "jwt_tests_rsa_consumer_1" }))
     local consumer4 = assert(cdao:insert({ username = "jwt_tests_rsa_consumer_2" }))
     local consumer5 = assert(cdao:insert({ username = "jwt_tests_rsa_consumer_5" }))
+    local consumer6 = assert(cdao:insert({ username = "jwt_tests_consumer_2" }))
     local anonymous_user = assert(cdao:insert({ username = "no-body" }))
 
     local pdao = helpers.dao.plugins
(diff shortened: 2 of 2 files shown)


kong/plugins/jwt/handler.lua
```diff
@@ -118,8 +118,9 @@ local function do_authentication(conf)
   end
 
   local claims = jwt.claims
-
-  local jwt_secret_key = claims[conf.key_claim_name]
+  local header = jwt.header
+  
+  local jwt_secret_key = claims[conf.key_claim_name] or header[conf.key_claim_name]
```
bungle: If the claim is found in both `payload` and `header`, the `header` one is never used. And I think you did this for better backward compatibility. I was just thinking, could we try both in that scenario and if one works (to signature verification) then continue. For that you may want to extract signature verification (and cache/db key loading) to its own function(s). And then just try both. If one works, then it is ok. @brycehemme Any thoughts on this (yes, it makes it a bit more complex, but also a bit more robust. Of course in most cases the `payload` or the `header` one is `nil` so you will still in most cases try signature verification just once?

brycehemme: That makes sense although it does seem to add unnecessary complexity, in my opinion. It seems like a major edge case where the same claim would exist in the body and the header, each having a different value, and one of them being intended to be used for looking up the proper key for checking the signature. Regardless, if you think this is necessary I can certainly make the change.

Additionally, I had considered instead handling the naming of the value for the "config.key_claim_name" as explicitly referring to the header. So, something like this -

```
config.key_claim_name=header.kid
```
So, if the value is prefixed with "header." then it will explicitly look at the header claims and if it isn't prefixed with "header." then it will use the standard claims. 

Thoughts on this approach vs. what you suggested vs. the way I've already implemented it?

Also, thanks for kicking the build!

brycehemme: @bungle - Can I get your thoughts on the above comments? I'm happy to address however you feel is appropriate. Thanks!

bungle: I think we have a few options:

1. leave it as is
2. use prefix `header.` or something to refer claims in JWT headers
3. try both if first one didn't work
4. add new config value, e.g. `config.key_claim_in=payload|header` (default being `payload`)
5. similar to 4. but ordered array (of these two possible values) `config.key_claim_in=header,payload`
6. A boolean value like `config.key_claim_in_header=true|false (`false` being default)

Maybe there are other options as well. @brycehemme what do you think? I don't have a favourite.

brycehemme:
I like option 2. It should be fully backwards compatible assuming people don’t have existing claims using that exact format. I’ll get this changed up today.

kong/plugins/jwt/handler.lua
brycehemme: I would opt for leaving it as is. I implemented #2 above but after I completed it I didn't really like having the logic in the code. Although it would work just fine, it just makes things more confusing than necessary, in my opinion. As you can tell, I prefer reducing complexity unless it is absolutely necessary. Any issues with accepting this PR in this case?

spec/03-plugins/17-jwt/03-access_spec.lua
```diff
@@ -150,6 +156,20 @@ describe("Plugin: jwt (access)", function()
...
+      local res = assert(proxy_client:send {
+        method  = "GET",
+        path    = "/request",
+        headers = {
+          ["Authorization"] = authorization,
+          ["Host"]          = "jwt10.com",
+        }
+      })
+      assert.res_status(200, res)
+    end)
```
thibaultcha: I don't think this test should be in the "refusals" section. The below test for "returns 200 on OPTIONS" either, but that is another issue.

thibaultcha:
Rebased and merged with the necessary modifications, thank you for the patch!

PR #3010, opened 2017-11-03
Add config option for ssl_protocols

francois-maillard:
That's useful for those poor people with client that still
support TLSv1 only.

### Summary

Add config option in kong.conf for ssl_protocols

### Full changelog

* Nginx's ssl_protocols option can now be configured from kong.conf
* Add related tests

Issue #2355, opened 2017-04-06
[community feedback requested] A new way to install and configure Kong


thibaultcha:
### Summary

1. Problem
2. Proposed new approach
  a. Installing
  b. Configuring
3. Pros and Cons
4. Community Feedback

### 1. Problem

Today's Kong can be quite un-intuitive and painful to both install and configure:

- For a source install, it requires installing separately several third-party software components (Serf/OpenResty/LuaRocks).
- Said source installs need a few flags during the Nginx compilation that Kong requires and that users need to add manually.
- Customizing the Nginx configuration requires creating a Penlight template with its own syntax and quirks, and users must update their template manually between Kong versions if new directives are added.
- When using a custom Nginx configuration, users have to maintain 2 configuration files: `kong.conf`, and `nginx.conf`.
- The Kong CLI is a process sitting between the user and the Nginx master process, and makes things like `stdout` logging or non-daemon mode harder for containers setups.
- Several Kong instances cannot run on the same machine at the same time.
- Installing the Kong Lua sources and dependencies requires an Internet access to luarocks.org and hosting code services like github.com.

Source installs are quite common because they allow building Nginx with custom flags (and custom Nginx C modules), and custom Nginx configurations are even more common, because `kong.conf` does not (and should not because it is not "scalable") support each and every Nginx directive.
 
#### a. Installing

Release Kong as an archive like `kong-0.12.0.tar.gz` containing:
- The OpenResty sources (which themselves contain the Nginx sources)
- LuaRocks
- The Kong core Lua sources
- The Kong community-edition, OSS plugins (currently, they live in the same repository but not for long)

The installation process for Kong would consist of:
```
$ wget http://getkong.org/download/kong-0.12.0.tar.gz
$ tar -xvf kong-0.12.0.tar.gz
$ cd kong-0.12.0
$ ./configure --prefix=/usr/local/kong
$ make
$ [sudo] make install
```

And that's it. Let's walk through the steps and what they achieved here:

```
$ wget http://getkong.org/download/kong-0.12.0.tar.gz
$ tar -xvf kong-0.12.0.tar.gz
$ cd kong-0.12.0
```

Here, we just downloaded the latest release archive and extracted it. Such an archive should be PGP signed by the core Kong contributors for each release, ensuring users get what they want.

```
$ ./configure --prefix=/usr/local/kong
```

[...]

PR #3323, opened 2018-03-20
Fix nginx ssl params

peakyblinder:

### Summary

Parameterize nginx SSL params, so the user can override the values as required

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX


thibaultcha:
Hi,

Thank you for the suggestion. However, I do not believe we can accept it. See [this comment](https://github.com/Kong/kong/pull/3010#issuecomment-342708588) for context (it is about a similar patch), and the [subsequent one](https://github.com/Kong/kong/pull/3010#issuecomment-342880170) as well.

thibaultcha:
Long story short, we try to shy away from making every nginx directive configurable because it is unmaintainable and unreasonable. We will try to come up with a different solution. Our ideal one is presented in #2355 if you are curious, but the end result might be a mix of `kong.conf` and this issue.

peakyblinder:
yes, that makes sense!

PR #3324, opened 2018-03-20
tests(spec-old-api) backport upstreams upper bound test updates

hishamhm:
Backport of a16537c into `spec-old-api` for `next`.


hishamhm:
I added the test update to `next` as you asked, but the actual commit that changes the upper bound is on `master` :laughing: 

I cherry-picked it here to make the test pass.

thibaultcha:
Ok, then I think it might be better to merge master in next now

thibaultcha:
Ok, I just merged master into next, and cherry-picked your spec-old-api commit there. Thanks!

PR #3328, opened 2018-03-21
feat(db) handle unpartitioned tables in Cassandra

kikito:


thibaultcha:
@kikito It seems like we removed one-too-many of those `ALLOW FILTERING` statements? See the CI results. I'm curious to know which one btw

kikito:
@thibaultcha fixed (there was also an issue with a bad Cassandra 2.x colum name)

kong/db/strategies/cassandra/init.lua
```diff
@@ -60,6 +60,128 @@ local function extract_pk_values(schema, entity)
...
+    cql = fmt([[
+      SELECT * FROM system.schema_columns
+      WHERE keyspace_name = '%s'
+      AND columnfamily_name = '%s'
+      AND column_name = 'partition'
+      ALLOW FILTERING;
+    ]], self.connector.keyspace, self.schema.name)
+  end
+
+  local rows = assert(self.connector:query(cql, {}, nil, "read"))
```
thibaultcha: our constructors always follow the `return nil, err` pattern, and never throw errors (except programmer errors). This should be returned all the way up to https://github.com/Kong/kong/blob/master/kong/db/strategies/init.lua#L40-L43

kong/db/strategies/cassandra/init.lua
```diff
@@ -60,6 +60,128 @@ local function extract_pk_values(schema, entity)
...
+  local cql
+
+  if self.connector.major_version == 3 then
+    cql = fmt([[
+      SELECT * FROM system_schema.columns
+      WHERE keyspace_name = '%s'
+      AND table_name = '%s'
+      AND column_name = 'partition'
+      AND kind = 'partition_key'
+      ALLOW FILTERING;
```
thibaultcha: Instead, it might be better to use:
```
SELECT * FROM system_schema.columns
  WHERE keyspace_name = 'kong_tests'
  AND table_name = 'routes'
  AND column_name = 'partition'
  AND kind = 'partition_key';
```

And when we retrieve the row:
```lua
return column_row.kind == "partition_key"
```

And adapt the C* 2.x table the same way (it can probably avoid this `ALLOW FILTERING` clause as well).

thibaultcha: @kikito ping. Thoughts on this?

kikito: Hi, the code you wrote provokes an `ALLOW FILTERING` error in cqlsh:

```
SELECT * FROM system_schema.columns
   ...   WHERE keyspace_name = 'kong_tests'
   ...   AND table_name = 'routes'
   ...   AND column_name = 'partition'
   ...   AND kind = 'partition_key';
InvalidRequest: Error from server: code=2200 [Invalid query] message="Cannot execute this query as it might involve data filtering and thus may have unpredictable performance. If you want to execute this query despite the performance unpredictability, use ALLOW FILTERING"
```

I have worked under the assumption that what you meant was this instead:

```
SELECT * FROM system_schema.columns
  WHERE keyspace_name = 'kong_tests'
  AND table_name = 'routes'
  AND column_name = 'partition';
```

That one works without ALLOW FILTERING.

> And when we retrieve the row:
> ```
> return column_row.kind == "partition_key"
> ```

That works in Cassandra 3 but not in 2, so I made different exit conditions for this as well.

kong/db/strategies/cassandra/init.lua
kikito: Understood. Removed the `assert` and added a bunch of `if err then return nil, err end`

kikito:
@thibaultcha made the changes you requested in Cassandra, apologies for the delay (I didn't notice the mentions amongst all the github noise).

kong/db/strategies/cassandra/init.lua
thibaultcha: Ah yes, that is what I meant (and check the `.kind` column afterwards).

thibaultcha:
It looks like the tests are broken now?

thibaultcha:
@kikito Ping

kikito:
@thibaultcha apologies, I did not notice your ping. It was a stupid typo, which I have fixed.

PR #3330, opened 2018-03-22
feat(router) expose router as a singleton

kidd:
Exposing the router as a singleton allows for more introspection of the defined api routes programmatically

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -170,6 +170,7 @@ function Kong.init()
   singletons.loaded_plugins = assert(load_plugins(config, dao))
   singletons.dao = dao
   singletons.configuration = config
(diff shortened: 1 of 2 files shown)


kong/init.lua
```diff
@@ -170,6 +170,7 @@ function Kong.init()
   singletons.loaded_plugins = assert(load_plugins(config, dao))
   singletons.dao = dao
   singletons.configuration = config
+  singletons.router = nil
```
thibaultcha: This should go in `singletons.lua` instead, where this table is (pre-)allocated for self-documentation purposes

thibaultcha:
Manually merged with minor edits!

PR #3331, opened 2018-03-22
added support for maximum expiration of a JWT token

mvanholsteijn:
### Summary

Support for limiting the expiration period on JWT tokens.

In the JWT plugin you can set  the property 'maximum_expiration'  to a positive integer, indicating the maximum number of seconds the 'exp' claim in the token may be ahead in the future.

The purpose of this feature is to prevent the use of tokens with extreme long lifetimes.

### Full changelog
- Added the 'maximum_expiration' to the jwt plugin schema
- Added a check in the handler
- Added unit tests for handler and schema.

### DIFF
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -165,6 +165,15 @@ local function do_authentication(conf)
     return false, {status = 401, message = errors}
   end
 
+  -- Verify the JWT registered claims
+  if conf.maximum_expiration ~= nil and conf.maximum_expiration > 0 then
+    local ok, errors = jwt:check_maximum_expiration(conf.maximum_expiration)
+    if not ok then
+      return false, {status = 403, message = errors}
+    end
+  end
+
+
   -- Retrieve the consumer
   local consumer_cache_key = singletons.db.consumers:cache_key(jwt_secret.consumer_id)
   local consumer, err      = singletons.cache:get(consumer_cache_key, nil,
--- a/kong/plugins/jwt/jwt_parser.lua
+++ b/kong/plugins/jwt/jwt_parser.lua
@@ -282,6 +282,23 @@ function _M:verify_registered_claims(claims_to_verify)
   return errors == nil, errors
 end
 
+--- check the maximum allowed expiration 
+-- @param maximum_expiration of the claim
+-- @return A Boolean indicating true, if the claim the maximum allowed expiration
+-- @return error if any
+function _M:check_maximum_expiration(maximum_expiration)
+  if maximum_expiration <= 0 then
+     return true
+  end
+
+  local exp = self.claims["exp"]
+  if exp == nil or (exp - ngx_time()) > maximum_expiration then
+    return false, {exp = "exceeds maximum allowed expiration"}
+  end
+
+  return true
+end
+
 _M.encode = encode_token
 
 return _M
--- a/kong/plugins/jwt/schema.lua
+++ b/kong/plugins/jwt/schema.lua
@@ -1,4 +1,5 @@
 local utils = require "kong.tools.utils"
+local Errors = require "kong.dao.errors"
 
 local function check_user(anonymous)
   if anonymous == "" or utils.is_valid_uuid(anonymous) then
@@ -6,6 +7,13 @@ local function check_user(anonymous)
   end
 
   return false, "the anonymous user must be empty or a valid uuid"
+end
+
+local function check_positive(v)
+  if v < 0 then
+    return false, "should be 0 or greater"
+  end
+  return true
 end
 
 return {
@@ -18,5 +26,23 @@ return {
     claims_to_verify = {type = "array", enum = {"exp", "nbf"}},
     anonymous = {type = "string", default = "", func = check_user},
     run_on_preflight = {type = "boolean", default = true},
+    maximum_expiration = {type = "number", default = 0, func = check_positive},
   },
+  self_check = function(schema, plugin_t, dao, is_update)
+    if plugin_t.maximum_expiration ~= nil and plugin_t.maximum_expiration > 0 then
+      local has_exp = false
+      if plugin_t.claims_to_verify then
+        for index, value in ipairs(plugin_t.claims_to_verify) do
+          if value == "exp" then
+            has_exp = true
+            break
+          end
+        end
+      end
+      if not has_exp then
+        return false, Errors.schema "to set maximum_expiration, you need to add 'exp' in claims_to_verify"
+      end
+    end
+    return true
+  end
 }
--- a/spec/03-plugins/17-jwt/03-access_spec.lua
+++ b/spec/03-plugins/17-jwt/03-access_spec.lua
@@ -28,7 +28,7 @@ for _, strategy in helpers.each_strategy() do
 
       local routes = {}
 
-      for i = 1, 10 do
+      for i = 1, 11 do
         routes[i] = bp.routes:insert {
           hosts = { "jwt" .. i .. ".com" },
         }
@@ -103,6 +103,12 @@ for _, strategy in helpers.each_strategy() do
         name     = "jwt",
         route_id = routes[10].id,
         config   = { key_claim_name = "kid" },
+      })
+
+      plugins:insert({
+        name     = "jwt",
+        route_id = routes[11].id,
+        config   = { claims_to_verify = {"nbf", "exp"}, maximum_expiration = 300 },
       })
 
       plugins:insert({
@@ -253,6 +259,39 @@ for _, strategy in helpers.each_strategy() do
         })
         local body = assert.res_status(401, res)
         assert.equal([[{"message":"Unauthorized"}]], body)
+      end)
+      it("returns 403 if the token exceeds the maximum allowed expiration limit", function()
+        local payload = {
+          iss = jwt_secret.key,
+          exp = os.time() + 3600,
+          nbf = os.time() - 30
+        }
+        local jwt = jwt_encoder.encode(payload, jwt_secret.secret)
+        local res = assert(proxy_client:send {
+          method = "GET",
+          path = "/request/?jwt=" .. jwt,
+          headers = {
+            ["Host"] = "jwt11.com"
+          }
+        })
+        local body = assert.res_status(403, res)
+        assert.equal('{"exp":"exceeds maximum allowed expiration"}', body)
+      end)
+      it("accepts a JWT token within the maximum allowed expiration limit", function()
+        local payload = {
+          iss = jwt_secret.key,
+          exp = os.time() + 270,
+          nbf = os.time() - 30
+        }
+        local jwt = jwt_encoder.encode(payload, jwt_secret.secret)
+        local res = assert(proxy_client:send {
+          method = "GET",
+          path = "/request/?jwt=" .. jwt,
+          headers = {
+            ["Host"] = "jwt11.com"
+          }
+        })
+        assert.res_status(200, res)
       end)
     end)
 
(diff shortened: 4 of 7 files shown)


jeremyjpj0916:
I like the idea here, but not all JWT's will include the exp and nbf, why not also make an if validating on if exp >= nginx.time + conf.valid_time or w/e you have if you want to keep token expiry an hour or less? It's something we implemented too to prevent clients from being irresponsible with their tokens.

mvanholsteijn:
I thought it would be cheaper to do the (exp -nbf) calculation. In addition, with (exp-nbf) I did not have to deal with clock skew.

I will change it, to use exp only. leaving the configuration of maximum_expiration to deal with possible clock skew..

Cheers,

Mark

mvanholsteijn:
@jeremyjpj0916 can you please review my pull request? Once it is integrated, I can stop merging the changes from master into this one 😃

jeremyjpj0916:
@mvanholsteijn I don't have write access, I chimed in only because I was implementing a similar change personally and thought to bring up the scenario where just exp is exposed and to compare against ngx.time() , which is looks like you added that so overall I think this is a nice feature! Hopefully kong team agrees and will consider implementing it as it makes sense from a security perspective.

mvanholsteijn:
@jeremyjpj0916 I did not know you did not have write access :-)

We implemented this feature after an external security audit was performed on our API. They recommended to limit the lifetime of the jwt token.

mvanholsteijn:
@thibaultcha  Do you have any idea on when this PR might be reviewed?

spec/03-plugins/17-jwt/03-access_spec.lua
```diff
@@ -254,6 +260,39 @@ for _, strategy in helpers.each_strategy() do
         local body = assert.res_status(401, res)
         assert.equal([[{"message":"Unauthorized"}]], body)
       end)
+      it("returns 403 if the token exceeds the maximum_expiration limit", function()
+        local payload = {
+	  iss = jwt_secret.key,
+	  exp = os.time() + 3600,
+	  nbf = os.time() - 30
```
thibaultcha: style: indentation is off here and in several other places in this file

kong/plugins/jwt/schema.lua
```diff
@@ -18,5 +26,20 @@ return {
...
+  self_check = function(schema, plugin_t, dao, is_update)
+    if plugin_t.maximum_expiration ~= nil and plugin_t.maximum_expiration > 0 then
+      local has_exp = false
+      if plugin_t.claims_to_verify then
+	for index, value in ipairs(plugin_t.claims_to_verify) do
+	  has_exp = has_exp or value == "exp"
+	end
+      end
+      if not has_exp then
+	 return false, Errors.schema "when you specify maximum_expiration, 'exp' must be in claims_to_verify"
```
thibaultcha: style: indentation is off here and above in the same function

spec/03-plugins/17-jwt/03-access_spec.lua
```diff
@@ -254,6 +260,39 @@ for _, strategy in helpers.each_strategy() do
...
+          exp = os.time() + 270,
+          nbf = os.time() - 30
+        }
+        local jwt = jwt_encoder.encode(payload, jwt_secret.secret)
+        local res = assert(proxy_client:send {
+        method = "GET",
+        path = "/request/?jwt=" .. jwt,
+        headers = {
+          ["Host"] = "jwt11.com"
+         }
```
thibaultcha: ditto: indentation

kong/plugins/jwt/schema.lua
```diff
@@ -18,5 +26,20 @@ return {
     claims_to_verify = {type = "array", enum = {"exp", "nbf"}},
     anonymous = {type = "string", default = "", func = check_user},
     run_on_preflight = {type = "boolean", default = true},
+    maximum_expiration = {type = "number", default = -1, func = check_maximum_expiration_positive},
```
thibaultcha: I think a more common pattern is to treat `0` as the "indefinite expiration", and any value below zero as invalid (as if it were an unsigned integer). The semantics are somewhat more obvious (`-1` has a strong "error-related" connotation and also feels strange in a time-related value).

kong/plugins/jwt/schema.lua
thibaultcha: Trying to come up with a more standard/consistent error message:
```
"claims_to_verify must contain 'exp' when specifying maximum_expiration"
```

kong/plugins/jwt/handler.lua
```diff
@@ -165,6 +165,15 @@ local function do_authentication(conf)
     return false, {status = 401, message = errors}
   end
 
+  -- Verify the JWT registered claims
+  if conf.maximum_expiration ~= nil and conf.maximum_expiration > 0 then
```
thibaultcha: The second check of this condition seems unnecessary, since you included a safeguard to `check_maximum_expiration()` already (`if max_exp <= 0`)

kong/plugins/jwt/jwt_parser.lua
```diff
@@ -282,6 +282,25 @@ function _M:verify_registered_claims(claims_to_verify)
...
+     return true, nil
+  end
+  local exp = self.claims["exp"]
+  if exp then
+    if (exp - ngx_time()) > maximum_expiration then
+      return false, {exp = "exceeds maximum expiration"}
+    end
+  else
+      return false, {exp = "is missing"}
+  end
```
thibaultcha: What do you think about: "exceeds maximum _allowed_ expiration"?

thibaultcha: Also, as per our contributing guide, we recommend hoisting short branches with early exits for clarity, and a few other minor nitpicks:
```lua
local exp = self.claims.exp
if not exp then
  return false, { exp = "missing" }
end

if exp - ngx_time() > maximum_expiration then
  return false, { exp = "exceeds maximum allowed expiration" }
end

return true
```

kong/plugins/jwt/schema.lua
```diff
@@ -8,6 +9,13 @@ local function check_user(anonymous)
   return false, "the anonymous user must be empty or a valid uuid"
 end
 
+local function check_maximum_expiration_positive(v)
+  if v < -1 or v == 0 then
+    return false, "maximum_expiration should be -1 or greater than 0"
```
thibaultcha: No need to repeat the name of the property in this error message, it will already be contained in the key for this error. We are also lacking a test for this validation.

kong/plugins/jwt/schema.lua
```diff
@@ -18,5 +26,20 @@ return {
...
     anonymous = {type = "string", default = "", func = check_user},
     run_on_preflight = {type = "boolean", default = true},
+    maximum_expiration = {type = "number", default = -1, func = check_maximum_expiration_positive},
   },
+  self_check = function(schema, plugin_t, dao, is_update)
+    if plugin_t.maximum_expiration ~= nil and plugin_t.maximum_expiration > 0 then
+      local has_exp = false
+      if plugin_t.claims_to_verify then
+	for index, value in ipairs(plugin_t.claims_to_verify) do
+	  has_exp = has_exp or value == "exp"
```
thibaultcha: A more convenient pattern is to break out of the loop:
```lua
if value == "exp" then
  has_exp = true
  break
end
```

spec/03-plugins/17-jwt/03-access_spec.lua
mvanholsteijn: fixed

kong/plugins/jwt/schema.lua
mvanholsteijn: changed to "To set maximum_expiration, you need to add 'exp' in claims_to_verify"

spec/03-plugins/17-jwt/03-access_spec.lua
mvanholsteijn: fixed.

kong/plugins/jwt/schema.lua
mvanholsteijn: you are right. fixed.

kong/plugins/jwt/handler.lua
mvanholsteijn: I wanted to avoid calling the function as  I know it is irrelevant.

kong/plugins/jwt/jwt_parser.lua
mvanholsteijn: I changed it to the message "exceeds maximum allowed expiration"  if the 'exp' is missing.

mvanholsteijn:
@thibaultcha I processed all (but 1) remarks.

kong/plugins/jwt/schema.lua
```diff
@@ -18,5 +26,23 @@ return {
...
+      if plugin_t.claims_to_verify then
+       for index, value in ipairs(plugin_t.claims_to_verify) do
+          if value == "exp" then
+            has_exp = true
+            break
+          end
+       end
+      end
+      if not has_exp then
+        return false, Errors.schema "To set maximum_expiration, you need to add 'exp' in claims_to_verify"
```
thibaultcha: A convention followed by absolutely all error messages in Kong is to not capitalize the sentences. Better not deviate from it (hence my previous suggestion).

kong/plugins/jwt/schema.lua
```diff
@@ -8,6 +9,13 @@ local function check_user(anonymous)
   return false, "the anonymous user must be empty or a valid uuid"
 end
 
+local function check_positive(v)
+  if v < 0 then
+    return false, "should be >= 0"
```
thibaultcha: Another convention our error messages follow is to use words instead of signs. In this case:
```
should be 0 or greater
or:
cannot be negative
```

kong/plugins/jwt/jwt_parser.lua
```diff
@@ -282,6 +282,23 @@ function _M:verify_registered_claims(claims_to_verify)
...
+-- @return A Boolean indicating true, if the claim the maximum allowed expiration
+-- @return error if any
+function _M:check_maximum_expiration(maximum_expiration)
+  if maximum_expiration <= 0 then
+     return true, nil
+  end
+
+  local exp = self.claims["exp"]
+  if exp == nil or (exp - ngx_time()) > maximum_expiration then
+      return false, {exp = "exceeds maximum allowed expiration"}
```
thibaultcha: style: the indentation in this codebase is 2 spaces (this one uses 4 and the above branch uses 3)

kong/plugins/jwt/jwt_parser.lua
```diff
@@ -282,6 +282,23 @@ function _M:verify_registered_claims(claims_to_verify)
   return errors == nil, errors
 end
 
+--- check the maximum allowed expiration 
+-- @param maximum_expiration of the claim
+-- @return A Boolean indicating true, if the claim the maximum allowed expiration
+-- @return error if any
+function _M:check_maximum_expiration(maximum_expiration)
+  if maximum_expiration <= 0 then
+     return true, nil
```
thibaultcha: `return true` would be enough here as per our return conventions.

kong/plugins/jwt/schema.lua
```diff
@@ -18,5 +26,23 @@ return {
     claims_to_verify = {type = "array", enum = {"exp", "nbf"}},
     anonymous = {type = "string", default = "", func = check_user},
     run_on_preflight = {type = "boolean", default = true},
+    maximum_expiration = {type = "number", default = 0, func = check_positive},
   },
+  self_check = function(schema, plugin_t, dao, is_update)
+    if plugin_t.maximum_expiration ~= nil and plugin_t.maximum_expiration > 0 then
+      local has_exp = false
+      if plugin_t.claims_to_verify then
+       for index, value in ipairs(plugin_t.claims_to_verify) do
```
thibaultcha: style: indentation is weird here as well (1 space?)

thibaultcha:
Merged with some modifications (testing, formatting, errors), thanks again for your contribution

mvanholsteijn:
@thibaultcha thank you for your support!

PR #3333, opened 2018-03-23
fix(plugin) http-log use lua-resty-http ensure http protocal completed

### LINKED ISSUE #3332: bug http-log losing log
### Summary

http-log losing log

Plugin does not fully implement http protocol resulting in partial log loss.

### Reason

If Kong http plugin has multiple concurrent log requests, some logs will be lost. Because http log not ensure http roundtrip completed before keepalive. So socket reuse a http session not completed the log will fail.

https://github.com/Kong/kong/blob/master/kong/plugins/http-log/handler.lua

mengskysama:
### Summary

https://github.com/Kong/kong/issues/3332

### Full changelog

* use lua-resty-http

### Issues resolved

Fix #3332

### DIFF
--- a/kong/plugins/http-log/handler.lua
+++ b/kong/plugins/http-log/handler.lua
@@ -86,7 +86,10 @@ local function log(premature, conf, body, name)
     ngx.log(ngx.ERR, name .. "request failed: " .. req_url .. "status: " .. tostring(res.status), err)
   end
 
-  httpc:set_keepalive(conf.timeout)
+  local ok, err = httpc:set_keepalive(conf.keepalive)
+  if not ok then
+    ngx.log(ngx.ERR, name .. "failed to set_keepalive", err)
+  end
 end
 
 -- Only provide `name` when deriving from this class. Not when initializing an instance.
--- a/kong/plugins/http-log/handler.lua
+++ b/kong/plugins/http-log/handler.lua
@@ -1,6 +1,7 @@
 local basic_serializer = require "kong.plugins.log-serializers.basic"
 local BasePlugin = require "kong.plugins.base_plugin"
 local cjson = require "cjson"
+local http = require "resty.http"
 local url = require "socket.url"
 
 local string_format = string.format
@@ -13,34 +14,6 @@ HttpLogHandler.VERSION = "0.1.0"
 
 local HTTP = "http"
 local HTTPS = "https"
-
--- Generates the raw http message.
--- @param `method` http method to be used to send data
--- @param `content_type` the type to set in the header
--- @param `parsed_url` contains the host details
--- @param `body`  Body of the message as a string (must be encoded according to the `content_type` parameter)
--- @return raw http message
-local function generate_post_payload(method, content_type, parsed_url, body)
-  local url
-  if parsed_url.query then
-    url = parsed_url.path .. "?" .. parsed_url.query
-  else
-    url = parsed_url.path
-  end
-  local headers = string_format(
-    "%s %s HTTP/1.1\r\nHost: %s\r\nConnection: Keep-Alive\r\nContent-Type: %s\r\nContent-Length: %s\r\n",
-    method:upper(), url, parsed_url.host, content_type, #body)
-
-  if parsed_url.userinfo then
-    local auth_header = string_format(
-      "Authorization: Basic %s\r\n",
-      ngx.encode_base64(parsed_url.userinfo)
-    )
-    headers = headers .. auth_header
-  end
-
-  return string_format("%s\r\n%s", headers, body)
-end
 
 -- Parse host url.
 -- @param `url` host url
@@ -57,6 +30,9 @@ local function parse_url(host_url)
   if not parsed_url.path then
     parsed_url.path = "/"
   end
+  if parsed_url.params then
+    parsed_url.params = "?" .. parsed_url.params
+  end
   return parsed_url
 end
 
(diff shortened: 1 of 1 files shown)


kong/plugins/http-log/handler.lua
```diff
@@ -72,37 +48,45 @@ local function log(premature, conf, body, name)
...
+  if res.status >= 400 then
+    ngx.log(ngx.ERR, name .. "request failed: " .. req_url .. "status: " .. tostring(res.status), err)
   end
 
-  ok, err = sock:setkeepalive(conf.keepalive)
-  if not ok then
-    ngx.log(ngx.ERR, name .. "failed to keepalive to " .. host .. ":" .. tostring(port) .. ": ", err)
-    return
-  end
+  httpc:set_keepalive(conf.timeout)
```
hishamhm: I noticed this replaced the use of `conf.keepalive` with `conf.timeout` -- was this intentional? Also, error checking here was removed. `lua-resty-http` will return non-error both when the connection is kept alive, or when it is properly closed.

mengskysama: It's my mistake to change `conf.keepalive` to `conf.timeout`. 

The test case of this plugin was disabled (It looks like the mockbin has an exception) so I not sure this change can pass the test. But verified in our production environment.

mengskysama:
@hishamhm Thanks for correcting, I have push a new commit.

kong/plugins/http-log/handler.lua
```diff
@@ -72,36 +48,47 @@ local function log(premature, conf, body, name)
...
+  if parsed_url.userinfo then
+    -- for lua-resty-http client Authorization
+    headers = {
+      ["Content-Type"] = conf.content_type,
+      ["Authorization"] = string_format(
+        "Basic %s", ngx.encode_base64(parsed_url.userinfo))
+    }
+    req_url = parsed_url.scheme .. "://" .. parsed_url.host .. ":" .. parsed_url.port
+      .. parsed_url.path .. (parsed_url.params or "")
+  else
```
thibaultcha: style: please jump a line above this `else` token, as per our updated style guide. Also, the indentation for the above concatenation operation could be improved and made more readable ideally.

kong/plugins/http-log/handler.lua
```diff
@@ -72,36 +48,47 @@ local function log(premature, conf, body, name)
...
+  httpc:set_timeout(conf.timeout)
+  local res, err = httpc:request_uri(req_url, {
+      ssl_verify = false,
+      method = conf.method:upper(),
+      body = body,
+      headers = headers
+  })
+
+  if not res then
+    ngx.log(ngx.ERR, name .. "request failed: " .. req_url, err)
```
thibaultcha: Need better formatting. This will not insert a space between `req_url` and `err`. We should also stop using this pattern of Lua-land string concatenations, let's update it while we're at it.
```lua
ngx.log(ngx.ERR, name, "request failed (", req_url, "): ", err)
```

kong/plugins/http-log/handler.lua
```diff
@@ -72,36 +48,47 @@ local function log(premature, conf, body, name)
...
 
-  ok, err = sock:send(generate_post_payload(conf.method, conf.content_type, parsed_url, body))
-  if not ok then
-    ngx.log(ngx.ERR, name .. "failed to send data to " .. host .. ":" .. tostring(port) .. ": ", err)
+  if res.status >= 400 then
+    ngx.log(ngx.ERR, name .. "request failed: " .. req_url .. "status: " .. tostring(res.status), err)
   end
 
-  ok, err = sock:setkeepalive(conf.keepalive)
+  local ok, err = httpc:set_keepalive(conf.keepalive)
```
thibaultcha: The [request_uri()](https://github.com/pintsized/lua-resty-http/blob/master/lib/resty/http.lua#L792) interface does the `set_keepalive()` automatically for us, so technically this would not be necessary.

However, using the `request_uri()` interface also renders the `conf.keepalive` parameter useless. We should probably use the lower-level `request()` interface instead. This also mean we will need to handle the SSL handshake ourselves, something that the `request_uri()` interface was doing for us.

kong/plugins/http-log/handler.lua
```diff
@@ -72,36 +48,47 @@ local function log(premature, conf, body, name)
...
-    if err then
-      ngx.log(ngx.ERR, name .. "failed to do SSL handshake with " .. host .. ":" .. tostring(port) .. ": ", err)
-    end
+  local httpc = http.new()
+  httpc:set_timeout(conf.timeout)
+  local res, err = httpc:request_uri(req_url, {
+      ssl_verify = false,
+      method = conf.method:upper(),
+      body = body,
+      headers = headers
```
thibaultcha: style: indentation is 2 spaces, not 4. We also encourage the use of a trailing comma in table declarations.

kong/plugins/http-log/handler.lua
mengskysama: I PR for http-lua-resty add an option for keepalive, I think that way simple and grace.

thibaultcha: That works as well, unfortunately this PR will have to stay stale until/if it is merged upstream and Kong updates its dependency.

Tieske: for reference, this is the PR: https://github.com/pintsized/lua-resty-http/pull/145

hishamhm: would it be too ugly to peek inside `httpc` and do `httpc.sock:set_keepalive` in the mean time?

thibaultcha: Yeah, let's avoid going down that road... If this has been undetected for so long it can stay like this a little bit longer and avoid us setting dangerous precedents.

mengskysama:
`lua-resty-http` owner didn't accept the PR until now.  So I make a new commit  for fix httpc:set_keepalive.

PR #3338, opened 2018-03-26
tests(balancer) increase speed and reliability of balancer tests

hishamhm:
* Removes timed dependencies based on `ngx.sleep`: using the new `/health` endpoint, we can poll-wait on health changes and write tests that are timing-independent
* Avoids restarting Kong on every test: instead, this generates unique ids so that each test works on different entities.

Local results show a speed up from ~1min26s down to ~11s. :tada: 

The wrapper `describe` block is left unintented in the `spec-old-api` version to reduce `git diff` and to make the actual changes clearer (otherwise the whole file would be marked as changed). The one in `spec` is the one we will want to keep long-term and that one has been indented properly.

(Eventually we should remove the "postgres, cassandra" iteration from inside the tests anyway and just run one of each database in separate Travis jobs.)

### DIFF
--- a/spec-old-api/02-integration/05-proxy/09-balancer_spec.lua
+++ b/spec-old-api/02-integration/05-proxy/09-balancer_spec.lua
@@ -2,9 +2,12 @@
 -- for dns-record balancing see the `dns_spec` files
 
 local helpers = require "spec.helpers"
-local PORT = 21000
 local utils = require "kong.tools.utils"
 local cjson = require "cjson"
+
+local FIRST_PORT = 20000
+local SLOTS = 10
+local HEALTHCHECK_INTERVAL = 0.01
 
 local healthchecks_defaults = {
   active = {
@@ -52,7 +55,7 @@ local TEST_LOG = false    -- extra verbose logging of test server
 local function direct_request(host, port, path)
   local pok, client = pcall(helpers.http_client, host, port)
   if not pok then
-    return nil, "pcall"
+    return nil, "pcall: " .. client .. " : " .. host ..":"..port
   end
   if not client then
     return nil, "client"
@@ -91,17 +94,22 @@ end
 
 -- Modified http-server. Accepts (sequentially) a number of incoming
 -- connections and then rejects a given number of connections.
--- @param timeout Server timeout.
 -- @param host Host name to use (IPv4 or IPv6 localhost).
 -- @param port Port number to use.
 -- @param counts Array of response counts to give,
 -- odd entries are 200s, event entries are 500s
 -- @param test_log (optional, default fals) Produce detailed logs
 -- @return Returns the number of succesful and failure responses.
-local function http_server(timeout, host, port, counts, test_log)
+local function http_server(host, port, counts, test_log)
+
+  -- This is a "hard limit" for the execution of tests that launch
+  -- the custom http_server
+  local hard_timeout = 300
+  local expire = ngx.now() + hard_timeout
+
   local threads = require "llthreads2.ex"
   local thread = threads.new({
-    function(timeout, host, port, counts, TEST_LOG)
+    function(expire, host, port, counts, TEST_LOG)
       local function test_log(...)
         if not TEST_LOG then
           return
@@ -127,7 +135,6 @@ local function http_server(timeout, host, port, counts, test_log)
 
       local handshake_done = false
 
-      local expire = socket.gettime() + timeout
       assert(server:settimeout(0.5))
       test_log("test http server on port ", port, " started")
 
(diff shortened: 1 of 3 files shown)


spec-old-api/02-integration/05-proxy/09-balancer_spec.lua
```diff
@@ -235,234 +237,396 @@ local function http_server(timeout, host, port, counts, test_log)
...
 local localhosts = {
   ipv4 = "127.0.0.1",
-  ipv6 = "0000:0000:0000:0000:0000:0000:0000:0001",
+  ipv6 = "[0000:0000:0000:0000:0000:0000:0000:0001]",
+  hostname = "localhost",
 }
 
 
-for ipv, localhost in pairs(localhosts) do
+dao_helpers.for_each_dao(function(kong_config)
```
thibaultcha: minor but just as a reminder: while we still iterate over the DBs in the tests, and because this helper comes from `spec` (and not `old-spec`), then we should use the new iterator provided by `spec/helpers.lua` and meant to replace this callback approach. The iterator is easier to use and also more flexible (it allows one to easily run only the `postgres` or `cassandra` version of the tests by specifying an argument.

Since this is the old-spec suite that is bound to disappear, it may be ok to use it there for now...

spec/02-integration/05-proxy/09-balancer_spec.lua
```diff
@@ -371,194 +454,24 @@ local localhosts = {
 }
 
 
-for mode, localhost in pairs(localhosts) do
+dao_helpers.for_each_dao(function(kong_config)
```
thibaultcha: Here however, we really should rely on the proper utility. Combined with `get_db_utils()`, we get:
```lua
for _, strategy in helpers.each_strategy(strategy?) do
describe("<component> [#" .. strategy .. "]", function()

  setup(function()
    local bp, db, dao = helpers.get_db_utils(strategy) -- also runs migrations for us

    assert(helpers.start_kong({
      database   = strategy,
      nginx_conf = "spec/fixtures/custom_nginx.template",
    }))
  end)
end)
end
```

spec-old-api/02-integration/05-proxy/09-balancer_spec.lua
```diff
@@ -235,234 +237,396 @@ local function http_server(timeout, host, port, counts, test_log)
...
+  delete_plugin = function(api_name, plugin_id)
+    local path = "/apis/" .. api_name .. "/plugins/" .. plugin_id
+    assert.same(204, api_send("DELETE", path, {}))
   end
-  api_client:close()
-  return res.status
+end
+
+
+local function truncate_relevant_tables()
```
thibaultcha: this could then receive an argument, `dao` and run the truncations from there instead of the very legacy `helpers.dao` instance

spec/helpers.lua
```diff
@@ -1163,8 +1173,13 @@ return {
...
   stop_kong = function(prefix, preserve_prefix, preserve_tables)
     prefix = prefix or conf.prefix
+
+    local running_conf = get_running_conf(prefix)
+    if not running_conf then return end
+
     local ok, err = kong_exec("stop --prefix " .. prefix)
-    wait_pid(conf.nginx_pid, nil)
+
+    wait_pid(running_conf.nginx_pid)
```
thibaultcha: This is a great fix. Maybe this warrants `--wait` flag in the CLI to allow for other users to not have to rely on custom logic when scripting `kong stop`? Maybe this should even be the default behaviour of `kong quit` (since it is the "graceful" stop).

hishamhm:
@thibaultcha Thank you for the review!

* reworded the helpers commit
* I added an optional `no_truncate` argument to `get_db_utils` because that truncates every table by default (speed difference was noticeable locally)
* changed to use `each_strategy` as you suggested, in both `spec` and `spec-old-api`
* got rid of `helpers.dao` in `truncate_relevant_tables`
* now that the code went through the initial round of reviews, I corrected the indentation of the `spec-old-api` version so that both files are in sync (now they are identical from `each_strategy` onwards)

> This is a great fix. Maybe this warrants --wait flag in the CLI to allow for other users to not have to rely on custom logic when scripting kong stop? Maybe this should even be the  default behaviour of kong quit (since it is the "graceful" stop).

To double-check if I understood this: you mean eventually moving `wait_pid` out of `helpers` (into `kong.tools.utils` or something) for use by `kong stop --wait` and `kong quit`?

thibaultcha:
Yay!

thibaultcha:
> To double-check if I understood this: you mean eventually moving wait_pid out of helpers (into kong.tools.utils or something) for use by kong stop --wait and kong quit?

Yes, this is what I meant (as a very low priority work item :sweat_smile:)

PR #3339, opened 2018-03-26
fix(hmac-auth) use ngx var request_uri instead uri

shashiranjan84:
### Summary

`ngx.var.uri` is normalized and so `ngx.var.request_uri` should be
used to get request origional uri with arguments

### Full changelog

* signature generator logic now use ngx.var.request_uri to
  get original request uri

### DIFF
--- a/spec/03-plugins/20-hmac-auth/03-access_spec.lua
+++ b/spec/03-plugins/20-hmac-auth/03-access_spec.lua
@@ -4,6 +4,8 @@ local helpers = require "spec.helpers"
 local utils = require "kong.tools.utils"
 local resty_sha256 = require "resty.sha256"
 
+local fmt = string.format
+
 
 local hmac_sha1_binary = function(secret, data)
   return openssl_hmac.new(secret, "sha1"):final(data)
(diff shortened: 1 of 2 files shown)


kidd:
lgtm

spec/03-plugins/20-hmac-auth/03-access_spec.lua
```diff
@@ -1078,6 +1078,188 @@ for _, strategy in helpers.each_strategy() do
...
+            ["HOST"]                = "hmacauth5.com",
+            date                    = date,
+            ["proxy-authorization"] = hmacAuth,
+            ["content-md5"]         = "md5",
+          },
+        })
+        assert.res_status(200, res)
+      end)
+
+      it("should pass with GET with request-line having encoded query param", function()
```
thibaultcha: what do we mean here by "encoded"? The below tests don't URL-encode the querystring which has an (invalid) space character.

If you wish to test a URL-encoded value, you have to encode it yourself (`:send()` doesn't do it for you)

spec/03-plugins/20-hmac-auth/03-access_spec.lua
```diff
@@ -1078,6 +1078,188 @@ for _, strategy in helpers.each_strategy() do
...
+            ["HOST"]                = "hmacauth5.com",
+            date                    = date,
+            ["proxy-authorization"] = hmacAuth,
+            ["content-md5"]         = "md5",
+          },
+        })
+        assert.res_status(200, res)
+      end)
+
+      it("should pass with GET with request-line having encode path param", function()
```
thibaultcha: ditto

spec/03-plugins/20-hmac-auth/03-access_spec.lua
shashiranjan84: Thanks @thibaultcha, I made bad assumption. I have encoded the parameters.

spec/03-plugins/20-hmac-auth/03-access_spec.lua
```diff
@@ -1078,6 +1080,195 @@ for _, strategy in helpers.each_strategy() do
...
+            ["content-md5"]         = "md5",
+          },
+        })
+        assert.res_status(200, res)
+      end)
+
+      it("should pass with GET with request-line having encoded query param", function()
+        local date = os.date("!%a, %d %b %Y %H:%M:%S GMT")
+        local escaped_uri = fmt("/request?name=foo bar",
+                                ngx.escape_uri("foo bar"))
```
thibaultcha: That won't work, the template string should contain `%s` in lieu of the hard-coded argument

shashiranjan84: Indeed, I am blind, fixed now.

spec/03-plugins/20-hmac-auth/03-access_spec.lua
```diff
@@ -1078,6 +1080,198 @@ for _, strategy in helpers.each_strategy() do
...
+            ["content-md5"]         = "md5",
+          },
+        })
+        assert.res_status(200, res)
+      end)
+
+      it("should pass with GET with request-line having encoded query param", function()
+        local date = os.date("!%a, %d %b %Y %H:%M:%S GMT")
+        local escaped_uri = fmt("/request?name=%s",
+                                ngx.escape_uri("foo bar"))
```
thibaultcha: btw, we could use `ngx.encode_args()` here, but that's fine

PR #3348, opened 2018-03-27
fix(admin) do not corrupt the plugin table when reading the schema

### LINKED ISSUE #3222: Invalid property values in response on POST request using a nested parameter
### Summary

Making a `POST` request to create a `key-auth` plugin using _nested params_, creates a response where the configuration property `key_names` value becomes `[ 'function' ]`

### Steps To Reproduce

**Note** `HTTPie 0.9.8` will properly send the request required to replicate as it will send `null` values.

Make a POST request using JSON with a nested parameter:

```bash
http --debug :8001/plugins name=key-auth config.run_on_preflight=true config.key_names:=null
```

Get back a response which contains the invalid output:

```json
{
    "config": {
        "anonymous": "",
        "hide_credentials": false,
        "key_in_body": false,
        "key_names": [
            "function"
        ],
        "run_on_preflight": true
    },
    "created_at": 1518551355000,
    "enabled": true,
    "id": "6e07f3fe-5502-43e8-8e26-c1d9e3981855",
    "name": "key-auth"
}
```

Where:

```
        "key_names": [
            "function"
        ],
```

**Bonus Points**: Caching issue

Delete the Plugin you just created and run the following request:

```bash
http --debug :8001/plugins name=key-auth config.run_on_preflight=true
```

You will get the same result.

Delete the plugin again, and re-run the request above. You will get the proper result. Seems like there are retained values in memory.

### Additional Details & Logs

- Kong version (`0.11+`)
- Ubuntu

hishamhm:
Make `remove_functions` produce a copy instead of modifying the plugin
schema table in-place.

FIxes #3222.

### DIFF
--- a/kong/api/routes/plugins.lua
+++ b/kong/api/routes/plugins.lua
@@ -7,14 +7,13 @@ local singletons = require "kong.singletons"
 -- Remove functions from a schema definition so that
 -- cjson can encode the schema.
 local function remove_functions(schema)
+  local copy = {}
   for k, v in pairs(schema) do
-    if type(v) == "function" then
-      schema[k] = "function"
-    end
-    if type(v) == "table" then
-      remove_functions(schema[k])
-    end
+    copy[k] =  (type(v) == "function" and "function")
+            or (type(v) == "table"    and remove_functions(schema[k]))
+            or v
   end
+  return copy
 end
 
 return {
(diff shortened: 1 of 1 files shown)


thibaultcha:
Thanks!

PR #3351, opened 2018-03-28
fix(responses) to not set content type header for 204 (no content)

### LINKED ISSUE #3342: Kong adds `Content-Type: text/plain; charset=UTF-8` header to 204 No Content replies of the upstream
### Summary

Upstream server replies *204 No Content* like this:
```
HTTP/1.1 204 No Content
Date: Tue, 27 Mar 2018 09:29:00 GMT
Server: xxx
```

And after response leaves Kong, it looks like this:
```
HTTP/1.1 204 No Content
Content-Type: text/plain; charset=UTF-8
Date: Tue, 27 Mar 2018 09:29:00 GMT
X-Kong-Upstream-Latency: 8
X-Kong-Proxy-Latency: 1
Via: kong/0.11.2
X-App-Server: kong/10.20.6.145:18000
Cache-Control: must-revalidate
Cache-Control: no-cache
Expires: 0
Pragma: no-cache
```

Where `Content-Type: text/plain; charset=UTF-8` is added without any reason.

### Steps To Reproduce

1. Get a http server who replies 204 without content-type
2. Request it through Kong

### Additional Details & Logs

- Kong version 0.11.2
- Kong configuration: registered APIs -- only custom made; plugins: none
- Operating System: Ubuntu 16.04

registered APIs: cmp-consents/consents
plugins: none
OS: Ubuntu 16.04

bungle:
### Summary

When there is no content (204 status code), there is no need to set Content-Type header. It may even be misleading.

### Issues resolved

Fix #3342


kong/tools/responses.lua
```diff
@@ -123,7 +123,11 @@ local function send_response(status_code)
     end
 
     ngx.status = status_code
-    ngx.header["Content-Type"] = "application/json; charset=utf-8"
+
+    if status_code ~= 204 then
```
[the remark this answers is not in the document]
bungle: @aleksandr-vin thanks!

kong/tools/responses.lua
```diff
@@ -123,7 +123,11 @@ local function send_response(status_code)
     end
 
     ngx.status = status_code
-    ngx.header["Content-Type"] = "application/json; charset=utf-8"
```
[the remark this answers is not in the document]
bungle: @aleksandr-vin good point, does this come out of Nginx custom error pages, then?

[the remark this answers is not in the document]
bungle: @aleksandr-vin I made a new push, this one solves `text/plain` case as well.

kong/templates/nginx_kong.lua
```diff
@@ -1,5 +1,6 @@
 return [[
 charset UTF-8;
+default_type '';
```
thibaultcha: unfortunately any configuration template change means it is breaking (for custom nginx configs) and should be opened against `next` instead...

kong/tools/responses.lua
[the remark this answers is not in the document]
thibaultcha: I don't think this should be removed? This is Kong-produced responses, not upstream responses.

kong/tools/responses.lua
```diff
@@ -144,6 +143,7 @@ local function send_response(status_code)
         ngx.log(ngx.ERR, "[admin] could not encode value: ", err)
       end
 
+      ngx.header["Content-Type"] = "application/json; charset=utf-8"
```
thibaultcha: Oh, you moved it here... Better not moving it down here unnecessarily? This instruction is currently voluntarily placed before the response body is produced for cognitive reasons.

kong/templates/nginx_kong.lua
thibaultcha: Also, all changed applied to `nginx_kong.lua` must be applied to the test template as well. And also, it would be better to nest this setting under the proxy `server {}` block.

kong/tools/responses.lua
bungle: Do you feel it needs to be added even when:
https://github.com/Kong/kong/blob/f65b9fa5cf463c9f8287fad3b923499d9f69da61/kong/tools/responses.lua#L139

(there is no content)?

kong/templates/nginx_kong.lua
bungle: Thanks for reminding me about the test ones.

bungle: Test template is updated. This is now rebased on `next`

kong/tools/responses.lua
thibaultcha: Yeah, maybe better to keep it in this branch then (but if we hit an error with the JSON-encoding, we might not want to set this header either).

kong/templates/nginx_kong.lua
thibaultcha: I think we should still have it moved to the proxy `server {}` block?

bungle: I moved it to proxy `location /` block. I think it is fine like that.

kong/tools/responses.lua
bungle: Yes, I made it only happen when we are actually outputting someting.

kong/tools/responses.lua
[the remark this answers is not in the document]
bungle: ^- that was discussed below.

kong/templates/nginx_kong.lua
thibaultcha: Yep, there is good!

kong/tools/responses.lua
```diff
@@ -142,9 +141,11 @@ local function send_response(status_code)
                                   {message = content})
       if not encoded then
         ngx.log(ngx.ERR, "[admin] could not encode value: ", err)
-      end
 
-      ngx.say(encoded)
+      else
+        ngx.header["Content-Type"] = "application/json; charset=utf-8"
+        ngx.say(encoded)
```
thibaultcha: :+1: 

We may need some unit tests for this (they exist for this module already)

thibaultcha: Ah, my bad, I see you already added it :)

thibaultcha:
Let's wait on CI and merge if it passes, thanks!

PR #3354, opened 2018-03-28
fix(ldap) cache key to work when plugin is configured globally

### LINKED ISSUE #3335: Cannot use ldap-auth as a global plugin
### Summary

In 0.13 the ldap_auth plugin will throw an error when configured as a global plugin

### Steps To Reproduce

1. Configure ldap plugin as global
2. any simple authentication attempt fails

### Additional Details & Logs

This is because the authenticate function attempts to use the route_id or api_id as part of the
cache key and there isn't one when the plugin is global:

local function authenticate(conf, given_credentials)
  local given_username, given_password = retrieve_credentials(given_credentials, conf)
  if given_username == nil then
    return false
  end

  local route_id = conf.route_id or conf.api_id <----- this is nil

  local cache_key = "ldap_auth_cache:" .. route_id .. ":" .. given_username
  local credential, err = singletons.cache:get(cache_key, {

sh-4.2# kong version
0.13.0

2018/03/24 17:27:44 [error] 31#0: *27888 [lua] responses.lua:121: access(): /usr/local/share/lua/5.1/kong/plugins/ldap-auth/access.lua:94: attempt to concatenate local 'route_id' (a nil value), client: 10.36.0.0, server: kong, request: ....

bungle:
### Summary

Changes the way how LDAP cache key in constructed. Previously it didn't share cache with similarly configured LDAP plugins. And more importantly, it was broken when configured as a global plugin (that was reported with #3335)

### Full changelog

* Implement cache key based on plugin configuration

### Issues resolved

Fix #3335

### DIFF
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -7,10 +7,12 @@ local match = string.match
 local lower = string.lower
 local find = string.find
 local sub = string.sub
+local fmt = string.format
 local ngx_log = ngx.log
 local request = ngx.req
 local ngx_error = ngx.ERR
 local ngx_debug = ngx.DEBUG
+local md5 = ngx.md5
 local decode_base64 = ngx.decode_base64
 local ngx_socket_tcp = ngx.socket.tcp
 local ngx_set_header = ngx.req.set_header
@@ -18,6 +20,10 @@ local tostring =  tostring
 
 local AUTHORIZATION = "authorization"
 local PROXY_AUTHORIZATION = "proxy-authorization"
+
+
+local ldap_config_cache = setmetatable({}, { __mode = "k" })
+
 
 local _M = {}
 
@@ -37,7 +43,6 @@ end
 local function ldap_authenticate(given_username, given_password, conf)
   local is_authenticated
   local err, suppressed_err, ok
-  local who = conf.attribute .. "=" .. given_username .. "," .. conf.base_dn
 
   local sock = ngx_socket_tcp()
   sock:settimeout(conf.timeout)
@@ -58,6 +63,7 @@ local function ldap_authenticate(given_username, given_password, conf)
     end
   end
 
+  local who = conf.attribute .. "=" .. given_username .. "," .. conf.base_dn
   is_authenticated, err = ldap.bind_request(sock, who, given_password)
 
   ok, suppressed_err = sock:setkeepalive(conf.keepalive)
@@ -84,15 +90,29 @@ local function load_credential(given_username, given_password, conf)
   return {username = given_username, password = given_password}
 end
 
+
+local function cache_key(conf, username)
+  if not ldap_config_cache[conf] then
+    ldap_config_cache[conf] = md5(fmt("%s:%u:%s:%s:%u",
+      lower(conf.ldap_host),
+      conf.ldap_port,
+      conf.base_dn,
+      conf.attribute,
+      conf.cache_ttl
+    ))
+  end
+
+  return fmt("ldap_auth_cache:%s:%s", ldap_config_cache[conf], username)
+end
+
+
 local function authenticate(conf, given_credentials)
   local given_username, given_password = retrieve_credentials(given_credentials, conf)
   if given_username == nil then
     return false
   end
 
-  local route_id = conf.route_id or conf.api_id
-  local cache_key = "ldap_auth_cache:" .. route_id .. ":" .. given_username
-  local credential, err = singletons.cache:get(cache_key, {
+  local credential, err = singletons.cache:get(cache_key(conf, given_username), {
     ttl = conf.cache_ttl,
     neg_ttl = conf.cache_ttl,
   }, load_credential, given_username, given_password, conf)
--- a/spec/03-plugins/21-ldap-auth/01-access_spec.lua
+++ b/spec/03-plugins/21-ldap-auth/01-access_spec.lua
@@ -2,8 +2,21 @@ local helpers = require "spec.helpers"
 local utils   = require "kong.tools.utils"
 
 
-local function acl_cache_key(route_id, username)
-  return "ldap_auth_cache:" .. route_id .. ":" .. username
+local lower   = string.lower
+local fmt     = string.format
+local md5     = ngx.md5
+
+
+local function cache_key(conf, username)
+  local prefix = md5(fmt("%s:%u:%s:%s:%u",
+    lower(conf.ldap_host),
+    conf.ldap_port,
+    conf.base_dn,
+    conf.attribute,
+    conf.cache_ttl
+  ))
+
+  return fmt("ldap_auth_cache:%s:%s", prefix, username)
 end
 
 
@@ -38,6 +51,10 @@ for _, strategy in helpers.each_strategy() do
 
       local route5 = bp.routes:insert {
         hosts = { "ldap5.com" },
+      }
+
+      bp.routes:insert {
+        hosts = { "ldap6.com" },
       }
 
       local anonymous_user = bp.consumers:insert {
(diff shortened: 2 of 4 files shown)


StephenLord:
I will give it a go, may take a day or so to get there, my kong deployments are currently done by pulling a docker image, so I need to work out how to drop this into my setup.

kong/plugins/ldap-auth/access.lua
```diff
@@ -84,14 +86,31 @@ local function load_credential(given_username, given_password, conf)
...
+
 local function authenticate(conf, given_credentials)
   local given_username, given_password = retrieve_credentials(given_credentials, conf)
   if given_username == nil then
     return false
   end
 
-  local route_id = conf.route_id or conf.api_id
-  local cache_key = "ldap_auth_cache:" .. route_id .. ":" .. given_username
+  local cache_key = acl_cache_key(conf, given_username)
```
Tieske: this is calculated on every request. We better cache the calculated key.

spec/03-plugins/21-ldap-auth/01-access_spec.lua
```diff
@@ -2,8 +2,24 @@ local helpers = require "spec.helpers"
...
+      conf.ldap_port,
+      conf.base_dn,
+      conf.attribute,
+      username,
+      conf.cache_ttl,
+    }, ":"))
+  }
+
+  return concat(cache_key, ":")
 end
```
Tieske: maybe also add a test that reproduces the original problem?

Tieske:
pinging @shashiranjan84 since he did the original work on this.

@bungle maybe add a small comment in the `schema.lua` file that if any properties are added/removed/updated, that the hash creation might need to be updated.

kong/plugins/ldap-auth/access.lua
bungle: Not sure, the calculation is not very intensive. Maybe caching is even slower than calculating MD5 out of small string? Not sure if it is worth considering added complexity?

kong/plugins/ldap-auth/access.lua
```diff
@@ -84,14 +86,31 @@ local function load_credential(given_username, given_password, conf)
   return {username = given_username, password = given_password}
 end
 
+
+local function acl_cache_key(conf, username)
```
shashiranjan84: Is this name intentional `acl_cache_key`, not a big deal though.

kong/plugins/ldap-auth/access.lua
Tieske: if you leave the username out of the hash, you can create the entire key and store it in a weak table. That would be a single table lookup, followed by the username concatenation.

kong/plugins/ldap-auth/access.lua
Tieske: oops... how did I miss that... good catch

kong/plugins/ldap-auth/access.lua
```diff
@@ -84,14 +86,31 @@ local function load_credential(given_username, given_password, conf)
...
+      conf.ldap_host,
+      conf.ldap_port,
+      conf.base_dn,
+      conf.attribute,
+      username,
+      conf.cache_ttl,
+    }, ":"))
+  }
+
+  return concat(cache_key, ":")
```
thibaultcha: It is a lot more efficient to follow the approach taken by `dao:cache_key()` rather than paying the cost of 2 shirt-lived tables, the double concatenation operation, and the `md5` hash. Any reason for now using a simple `string.format` instead? Combined with @Tieske's comment about caching the cache key prefix (that of the plugin configuration), and simply use the `username` as the dynamic component?

kong/plugins/ldap-auth/access.lua
bungle: I copied that name from unit tests. I can rename both.

kong/plugins/ldap-auth/access.lua
bungle: Sure, I will change that. A lot more in this case might be a few microseconds, but of course better to do it like that. Not sure though, should we use something else than `MD5`. Maybe requiring `lua-resty-string` and using `SHA256` instead.

thibaultcha: Well, as always, it isn't about making that particular piece of the code run a few microseconds faster, but rather prevent the spreading of anti-patterns and ensure that _overall_, the proxy still performs relatively well in hot code paths. To avoid accumulating performance penalties when they can be avoided. It is a continuous effort.
In this case if we cache the key and only rotate the `username` part, we can get away with the current approach, sure.

I'm not sure why the hashing is required though?

Tieske: I think the resulting string is a lot smaller?

thibaultcha: That's the only benefit I am seeing as well (in which case md5 should be just fine?)

bungle: A key without hashing:
```
ldap_auth_cache:ec2-54-172-82-117.compute-1.amazonaws.com:389:ou=scientists,dc=ldap,dc=mashape,dc=com:uid:2:einstein
```

A key with hashing:
```
ldap_auth_cache:aec022a3d1d6590c4432b4bc65433e2e:einstein
```

bungle:
@Tieske, @thibaultcha, @shashiranjan84, 

Ok, I tried to fix all your concerns. Let me know what you think about it now.

thibaultcha:
The CI build seems to consistently error on the following step:
```
spec-old-api/03-plugins/21-ldap-auth/01-access_spec.lua @ 326: Plugin: ldap-auth (access) caches LDAP Auth Credential
```

I restarted both builds to make sure that this wasn't a flaky setup.

kong/plugins/ldap-auth/access.lua
```diff
@@ -84,15 +90,29 @@ local function load_credential(given_username, given_password, conf)
...
+
+local function cache_key(conf, username)
+  if not ldap_config_cache[conf] then
+    ldap_config_cache[conf] = md5(fmt("%s:%u:%s:%s:%u",
+      conf.ldap_host,
+      conf.ldap_port,
+      conf.base_dn,
+      conf.attribute,
+      conf.cache_ttl
+    ))
```
Tieske: we should remove case sensitivity from the cache-key for those elements that are not case sensitive (eg. hostname, haven't looked into the ldap specifc ones).

Tieske:
@thibaultcha somewhat related to this, in the light of plugin-api spec, we could consider dropping the weak table pattern and use the config table instead.

- currently the config table is considered read-only, hence we use the table as a key in a weak table to cache intermediary results.

we could define it as:

- the plugin is the owner of the config table. Allowing the plugin to update the table.
- Kongs responsibility will only be to fetch the plugin configuration and take care of invalidation and caching
- Any changes made to the table will never be persisted, and are local to a worker process
- The lifetime of the table may differ from the worker process based on lru or similar cache evictions

we might even consider to add a `initialize_config()` method to the plugins to create the intermediary updates right after fetching it, and forego on any runtime checks.

None of this for this PR though.

bungle:
@tieske, yes I was looking forward for that, but couldn't find a place currently to do that. Your other comments are good (case insensitivity), also I need to check that erroring test.

kong/plugins/ldap-auth/access.lua
bungle: LDAP specific ones are **usually** `case-insensitive`, but you may possibly configure them to be `case-sensitive`.

Tieske: then the safe way would be to keep them case-sensitive, so that only leaves the hostname to update.

bungle: http://ldapwiki.com/wiki/Distinguished%20Name%20Case%20Sensitivity

(I don't have a courage to make other than `host` case-insensitive).

Tieske: aren't the host and port properties only used to connect to the ldap server?

PR #3355, opened 2018-03-28
fix(dao) cassandra to handle ngx.null with check_*_constraints

### LINKED ISSUE #3308: Cannot clear consumer configuration from plugins when using Cassandra
### Summary

Kong [allows PATCHes with parameters set to `null` to clear configuration](https://github.com/Kong/kong/commit/fc82dd73918bed332f46a66cdd8204d24e6e3fcb) where appropriate. Clearing consumers from a plugin in this fashion works when the database backend is Postgres, but not Cassandra.

### Steps To Reproduce

1. On a Cassandra-backed Kong, create a plugin associated with a consumer, e.g. `http -v post http://localhost:8001/plugins/ name="request-size-limiting" config.allowed_payload_size=128 consumer_id=<ID> api_id=<ID>`
2. Attempt to clear the consumer configuration, e.g. `http -v patch http://localhost:8001/plugins/<ID> consumer_id:=null`
3. Kong will respond a 500 and `"message": "An unexpected error occurred"`
4. Kong will log `2018/03/16 18:54:17 [error] 45#0: *2749 [lua] responses.lua:121: handler(): [Invalid] Invalid null value in condition for column id, client: 10.142.0.3, server: kong_admin, request: "PATCH /plugins/<ID> HTTP/1.1", host: "127.0.0.1:8001"`

### Additional Details & Logs

- Kong version: 0.12.3
- Kong debug logs: https://gist.github.com/rainest/5a40605678cb4f79429d7d19a956fa5a
- Operating System: standard Kong Docker image on GCE

Initial comment from @thibaultcha: Yep, there should be a check here to use the null uuid instead of NULL since in the C* table, `consumer_id` is part of the partition key iirc

bungle:
### Summary

Fix some issues with C* when doing verifications on constraints while the value was `ngx.null`.

### Full changelog

* Implement check for `ngx.null` in `check_unique_constraints`
* Implement check for `ngx.null` in `check_foreign_constaints`

### How to reproduce the problem

```bash
KONG_DATABASE=cassandra bin/kong start
http :8001/apis name=test upstream_url=http://www.google.com/ uris=/
http :8001/consumers username=test
http :8001/plugins name=request-transformer api_id=<API_ID> consumer_id=<CONSUMER_ID>
http patch :8001/plugins/<PLUGIN_ID> consumer_id:=null
```

### Issues resolved

Fix #3308

### DIFF
--- a/kong/dao/db/cassandra.lua
+++ b/kong/dao/db/cassandra.lua
@@ -357,7 +357,7 @@ local function check_unique_constraints(self, table_name, constraints, values, p
 
   for col, constraint in pairs(constraints.unique) do
     -- Only check constraints if value is non-null
-    if values[col] ~= nil then
+    if values[col] ~= nil and values[col] ~= ngx.null then
       local where, args = get_where(constraint.schema, {[col] = values[col]})
       local query = select_query(table_name, where)
       local rows, err = self:query(query, args, nil, constraint.schema)
@@ -393,7 +393,7 @@ local function check_foreign_constaints(self, values, constraints)
   for col, constraint in pairs(constraints.foreign) do
     -- Only check foreign keys if value is non-null,
     -- if must not be null, field should be required
-    if values[col] ~= nil then
+    if values[col] ~= nil and values[col] ~= ngx.null then
       local res, err = self:find(constraint.table, constraint.schema, {
         [constraint.col] = values[col]
       })
--- a/kong/dao/schemas_validation.lua
+++ b/kong/dao/schemas_validation.lua
@@ -277,9 +277,19 @@ function _M.validate_entity(tbl, schema, options)
       end
 
       if errors == nil and type(schema.self_check) == "function" then
+        local nil_c = {}
+        for column in pairs(schema.fields) do
+          if t[column] == ngx.null then
+            t[column] = nil
+            table.insert(nil_c, column)
+          end
+        end
         local ok, err = schema.self_check(schema, t, options.dao, options.update)
         if ok == false then
           return false, nil, err
+        end
+        for _, column in ipairs(nil_c) do
+          t[column] = ngx.null
         end
       end
     end
--- a/spec/02-integration/04-admin_api/04-plugins_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/04-plugins_routes_spec.lua
@@ -1,187 +1,230 @@
 local helpers = require "spec.helpers"
-local cjson = require "cjson"
+local cjson   = require "cjson"
 
-describe("Admin API", function()
-  local client
-  setup(function()
-    assert(helpers.dao:run_migrations())
-    assert(helpers.db:truncate())
+for _, strategy in helpers.each_strategy() do
 
-    assert(helpers.start_kong())
-    client = helpers.admin_client()
-  end)
-  teardown(function()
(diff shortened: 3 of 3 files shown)


thibaultcha:
No regression test case? :/

kong/dao/db/cassandra.lua
```diff
@@ -357,7 +357,7 @@ local function check_unique_constraints(self, table_name, constraints, values, p
 
   for col, constraint in pairs(constraints.unique) do
     -- Only check constraints if value is non-null
-    if values[col] ~= nil then
+    if values[col] ~= nil and values[col] ~= ngx.null then
```
thibaultcha: I thought it was the `schema_validator`'s responsibility to nullify the column value if `ngx.null` (JSON NULL) is given? As per the existing implementation [here](https://github.com/Kong/kong/blob/master/kong/dao/schemas_validation.lua#L102-L104) and [here](https://github.com/Kong/kong/blob/master/kong/dao/schemas_validation.lua#L120-L122)?

bungle: Well, here we need to pass something to DAO. If we `nil` then that field will not be updated (or `nullified`). If we set it to `cassandra.null` then the generic implementations needs to know about the database. `ngx.null` is common way to tell the strategies to do something. Serialize function already handles this: https://github.com/Kong/kong/blob/12244ec4e9bb3587a9e34608ff83c9deaf4a8ec3/kong/dao/db/cassandra.lua#L314

thibaultcha: Ah true. Alright, then I think all we need would be a regression test for this?

bungle: That's right. I will be adding it.

bungle: @thibaultcha the test was added.

kong/dao/schemas_validation.lua
```diff
@@ -277,10 +277,20 @@ function _M.validate_entity(tbl, schema, options)
...
+            table.insert(nil_c, column)
+          end
+        end
         local ok, err = schema.self_check(schema, t, options.dao, options.update)
         if ok == false then
           return false, nil, err
         end
+        for _, column in ipairs(nil_c) do
+          t[column] = ngx.null
+        end
```
bungle: I don't like this, but there is no other way around it. Only way is to update all the `self_check`s to work with `ngx.null`s?

thibaultcha: Agreed, this change makes sense for backwards-compatibility purposes.

spec/02-integration/04-admin_api/04-plugins_routes_spec.lua
```diff
@@ -1,187 +1,230 @@
...
-      local res = assert(client:send {
-        method = "GET",
-        path = "/plugins/enabled",
-      })
-      local body = assert.res_status(200, res)
-      local json = cjson.decode(body)
-      assert.is_table(json.enabled_plugins)
-      assert.True(#json.enabled_plugins > 0)
+      dao:truncate_tables()
+      db:truncate()
```
thibaultcha: These operations are now handled by `get_db_utils()`. Will update locally but good for the next time :)

thibaultcha:
Merged!

PR #3358, opened 2018-03-29
fix(api) parsing service url default to 443 for https

### LINKED ISSUE #3357: service 'url' creation defaults to port 80 for https
example:
```shell
$ http -f post :8001/services name=test15 path[]=/test15 url=https://test.host.com/test-path
HTTP/1.1 201 Created
Connection: keep-alive
Date: Thu, 29 Mar 2018 11:51:06 GMT
Server: kong/0.13.0
Transfer-Encoding: chunked

{
    "connect_timeout": 60000,
    "created_at": 1522324266,
    "host": "test.host.com",
    "id": "4daed54b-136b-436c-b738-9a90b40dd2b1",
    "name": "test15",
    "path": "/test-path",
    "port": 80,
    "protocol": "https",
    "read_timeout": 60000,
    "retries": 5,
    "updated_at": 1522324266,
    "write_timeout": 60000
}
```

Since the `url` parameter provided is using the `https` scheme, it is expected that the `port` property ends up being `443` by default.

If specified separately then I get that the default for `port` is 80. But when parsed from `url` the default should be based on the scheme.

Tieske:
The `url` convenience property was parsed, but defaulted
to port 80, even for 'https' schemes. Updated to 443.

fixes #3357

### DIFF
--- a/spec/02-integration/04-admin_api/21-services_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/21-services_routes_spec.lua
@@ -97,6 +97,31 @@ for _, strategy in helpers.each_strategy() do
             assert.equals(60000, json.read_timeout)
           end
         end)
+
+        it_content_types("'port' defaults to 443 when 'url' scheme is https", function(content_type)
+          return function()
+            local res = client:post("/services", {
+              body = {
+                url = "https://service.com/",
+              },
+              headers = { ["Content-Type"] = content_type },
+            })
+            local body = assert.res_status(201, res)
+            local json = cjson.decode(body)
+
+            assert.is_string(json.id)
+            assert.is_number(json.created_at)
+            assert.is_number(json.updated_at)
+            assert.equals(cjson.null, json.name)
+            assert.equals("https", json.protocol)
+            assert.equals("service.com", json.host)
+            assert.equals("/", json.path)
+            assert.equals(443, json.port)
+            assert.equals(60000, json.connect_timeout)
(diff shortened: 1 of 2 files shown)


hishamhm:
@Tieske Travis failures are related: `url.parse` may not return a `scheme` in invalid URLs. 

Also, no need to do `:lower()` and in fact I'd prefer if it didn't:  the schema validator won't allow uppercase schemes anyway (the lenience there makes the code misleading, as it makes it seem that uppercase inputs _are_ accepted).

bungle:
@hishamhm I think at some point I asked that should schemas have more intelligent defaults? Eg. instead of fixing this here, should schema do that?

bungle:
I also agree that `lower` can be removed. And we need to add test or too., as usual

Tieske:
@bungle @hishamhm comments addressed

spec/02-integration/04-admin_api/21-services_routes_spec.lua
```diff
@@ -97,6 +97,31 @@ for _, strategy in helpers.each_strategy() do
             assert.equals(60000, json.read_timeout)
           end
         end)
+
+        it_content_types("creates a service with url (https based)", function(content_type)
```
thibaultcha: Isn't the proper name for this test: `'port' defaults to 443 when 'url' scheme is https`?

Tieske:
description updated, test fixed, and 'tonumber' left alone since it is required.

PR #3362, opened 2018-03-31
fix(dao) cassandra boolean serialization

salazar:
### Summary

Correctly serialize boolean values in Cassandra.

### Issues resolved

Queries filtering on boolean values would result in an error. At the admin API level, this can be reproduced by issuing a request that filters on a boolean field - e.g., `GET /apis?preserve_host=false`, which will result in a 500 response.


thibaultcha:
I think an integration test in the DAO or Admin API suite would be appropriate?

thibaultcha:
Manually merged, thank you!

PR #3364, opened 2018-04-02
feat(schema) friendlier behavior for at_least_one_of

### LINKED ISSUE #3361: Updating a route requires path, host, method to be set
### Summary

According to the documentation, it stats that path, host, method are conditionally required as such that one of the 3 must be set. However all 3 are required when doing an update.

### Steps To Reproduce

1. Create a Service
> request
```
curl -X POST \
  http://172.30.0.102:8001/services \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/x-www-form-urlencoded' \
  -H 'postman-token: b56e1fd6-c4c9-1e3f-50e4-533e3249e982' \
  -d 'name=bug-report&protocol=https&host=127.0.0.1&path=%2F&port=443'
```
> response
```
{
    "host": "127.0.0.1",
    "created_at": 1522350788,
    "connect_timeout": 60000,
    "id": "be8493de-27f4-49c1-9606-cc3e24e8a7e4",
    "protocol": "https",
    "name": "bug-report",
    "read_timeout": 60000,
    "port": 443,
    "path": "/",
    "updated_at": 1522350788,
    "retries": 5,
    "write_timeout": 60000
}
```
2. Create a Route
> request
```
curl -X POST \
  http://172.30.0.102:8001/routes \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/x-www-form-urlencoded' \
  -H 'postman-token: ffbdedb3-b935-7774-938c-70b8fa634239' \
  -d 'protocols%5B%5D=http&protocols%5B%5D=https&paths%5B%5D=%2Ftest&service.id=be8493de-27f4-49c1-9606-cc3e24e8a7e4'
```
> response
```
{
    "created_at": 1522350912,
    "strip_path": true,
    "hosts": null,
    "preserve_host": false,
    "regex_priority": 0,
    "updated_at": 1522350912,
    "paths": [
        "/test"
    ],
    "service": {
        "id": "be8493de-27f4-49c1-9606-cc3e24e8a7e4"
    },
    "methods": null,
    "protocols": [
        "http",
        "https"
    ],
    "id": "1e61ab66-8ce7-4036-85a7-cdd0089ea3c7"
}
```
3. Update route by supplying only a path, host or method, example shown with method but all 3 produce similar results.
> request
```
curl -X PATCH \
  http://172.30.0.102:8001/routes/1e61ab66-8ce7-4036-85a7-cdd0089ea3c7 \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/x-www-form-urlencoded' \
  -H 'postman-token: 36e12a0e-6dfe-e9d0-4dd5-db74d72f2b64' \
  -d methods%5B%5D=POST
```
> response
```
{
    "fields": {
        "hosts": "field required for entity check",
        "paths": "field required for entity check"
    },
    "name": "schema violation",
    "code": 2,
    "message": "2 schema violations (hosts: field required for entity check; paths: field required for entity check)"
}
```

### Additional Details & Logs

- Kong version 0.13.0

hishamhm:
Make the behavior of the validation of routes friendlier:

Without this patch, if a user performs a PATCH request without giving explicit values to all three of `methods`, `hosts` or `paths`, we reject the request because we need to cross-check all three fields to ensure that at least one of those is not-null. If only one of those
is to be set, the user needs to explicitly set the other two to `null` in the request.

With this patch, if a user performs a PATCH request giving one non-empty value for at least one of `methods`, `hosts` or `paths`, we now accept it even if the other fields are missing, because this request cannot put these fields in an invalid state.

This was implementing in a very simple way, by adding a new internal property to `entity_checkers`, called `run_with_missing_fields`. Currently, the only entity checker that can run with missing fields is `at_least_one_of`, which is used by the Route entity.

This PR includes a "progression" test ( :) ). Note that this causes a couple of error messages to change, but I don't this should be considered a serious breaking change (especially since users are way less likely to bump into these errors with this patch).

Closes #3361.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -64,6 +64,9 @@ local validation_errors = {
   ENTITY_CHECK              = "failed entity check: %s(%s)",
   ENTITY_CHECK_N_FIELDS     = "entity check requires %d fields",
   CHECK                     = "entity check failed",
+  CONDITIONAL               = "failed conditional validation",
+  AT_LEAST_ONE_OF           = "at least one of these fields must be non-empty: %s",
+  ONLY_ONE_OF               = "only one of these fields must be non-empty: %s",
   -- schema error
   SCHEMA_NO_DEFINITION      = "expected a definition table",
   SCHEMA_NO_FIELDS          = "error in schema definition: no 'fields' table",
@@ -296,24 +299,15 @@ end
 
 --- Produce a nicely quoted list:
 -- Given `{"foo", "bar", "baz"}` and `"or"`, produces
--- `"'foo', 'bar' or 'baz'"`.
+-- `"'foo', 'bar', 'baz'"`.
 -- @param words an array of strings.
--- @param conjunction The conjunction to use before the last item.
 -- @return The string of quoted words.
-local function quoted_list(words, conjunction)
+local function quoted_list(words)
   local msg = {}
-  local next_to_last = #words - 1
-  for i, word in ipairs(words) do
-    insert(msg, "'")
-    insert(msg, word)
-    insert(msg, "'")
-    if i == next_to_last then
-      insert(msg, " " .. conjunction .. " ")
-    elseif i < next_to_last then
-      insert(msg, ", ")
-    end
+  for _, word in ipairs(words) do
+    insert(msg, ("'%s'"):format(word))
   end
-  return concat(msg)
+  return concat(msg, ", ")
 end
 
 
@@ -374,8 +368,7 @@ Schema.entity_checkers = {
         end
       end
 
-      return nil, "at least one of " .. quoted_list(field_names, "or")
-                  .. " must be non-empty"
+      return nil, quoted_list(field_names)
     end,
   },
 
@@ -397,8 +390,7 @@ Schema.entity_checkers = {
       if ok then
         return true
       end
-      return nil, "only one of " .. quoted_list(field_names, "or")
-                  .. " must be non-empty"
+      return nil, quoted_list(field_names)
     end,
   },
 
@@ -426,7 +418,7 @@ Schema.entity_checkers = {
       -- Handle `required`
       if arg.then_match.required == true and then_value == null then
         local field_errors = { [arg.then_field] = validation_errors.REQUIRED }
-        return nil, validation_errors.IF_THEN, field_errors
+        return nil, validation_errors.CONDITIONAL, field_errors
       end
 
       local then_merged = merge_field(schema.fields[arg.then_field], arg.then_match)
@@ -434,7 +426,7 @@ Schema.entity_checkers = {
       ok, err = Schema.validate_field(schema, then_merged, then_value)
       if not ok then
         local field_errors = { [arg.then_field] = err }
-        return nil, validation_errors.IF_THEN, field_errors
+        return nil, validation_errors.CONDITIONAL, field_errors
       end
 
       return true
@@ -738,7 +730,7 @@ local function run_entity_check(self, name, input, arg)
   for _, fname in ipairs(fields_to_check) do
     if input[fname] == nil then
       if not checker.run_with_missing_fields then
-        local err = validation_errors.REQUIRED_FOR_ENTITY_CHECK:format(fname)
+        local err = validation_errors.REQUIRED_FOR_ENTITY_CHECK
         field_errors[fname] = err
         ok = false
       end
@@ -762,7 +754,7 @@ local function run_entity_check(self, name, input, arg)
     return true
   end
 
-  err = err or validation_errors[name:upper()]
+  err = validation_errors[name:upper()]:format(err)
   if not err then
     local data = pretty.write({ name = arg }):gsub("%s+", " ")
     err = validation_errors.ENTITY_CHECK:format(name, data)
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1067,7 +1067,23 @@ end
 -- indexed numerically for general errors, and by field name for field errors.
 -- In all cases, the input table is untouched.
 function Schema:validate_update(input)
-  return self:validate(input, false)
+
+  -- Monkey-patch some error messages to make it clearer why they
+  -- apply during an update. This avoids propagating update-awareness
+  -- all the way down to the entity checkers (which would otherwise
+  -- defeat the whole purpose of the mechanism).
+  local rfec = validation_errors.REQUIRED_FOR_ENTITY_CHECK
+  local aloo = validation_errors.AT_LEAST_ONE_OF
+  validation_errors.REQUIRED_FOR_ENTITY_CHECK = rfec .. " when updating"
+  validation_errors.AT_LEAST_ONE_OF = "when updating, " .. aloo
+
+  local ok, err, err_t = self:validate(input, false)
+
+  -- Restore the original error messages
+  validation_errors.REQUIRED_FOR_ENTITY_CHECK = rfec
+  validation_errors.AT_LEAST_ONE_OF = aloo
+
+  return ok, err, err_t
 end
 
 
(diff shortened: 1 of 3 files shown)


spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
```diff
@@ -395,6 +395,26 @@ for _, strategy in helpers.each_strategy() do
           assert.same(route, new_route)
         end)
 
+        it("accepts a partial update to routing criteria when not null", function()
```
thibaultcha: we can make this test case name more explicit for our future selves I think: "when at least one of the required fields it not null"

spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
```diff
@@ -410,10 +430,15 @@ for _, strategy in helpers.each_strategy() do
...
-            }
+            message  = unindent([[
+              schema violation
+              (at least one of 'methods', 'hosts' or 'paths' must be non-empty)
+            ]], true, true),
+            fields   = {
+              ["@entity"] = {
+                "at least one of 'methods', 'hosts' or 'paths' must be non-empty",
+              }
+            },
```
thibaultcha: It seems like this test becomes the same one as the above "fails trying to unset an interdependent field by itself":

1. we create a valid entity
2. we send an update with all included "semi-required" fields set to null

It seems like it becomes "fails trying to unset many interdependent field by themselves", which feels redundant with the previous test? It only differs in what the original entity has prior to the update, which should not change the result of the test because it focuses on the update operation.

The above test also deserves a comment that I will add here for brevity: it creates an entity with both `hosts` and `methods`, and sends an update nullifying `methods`. But the received error would be hard to swallow as a user: `at least one of 'methods', 'hosts' or 'paths' must be non-empty`. Yes, that is still the case. But the user does not know we do not do a read-before-write when updating this entity. It might be worth returning:

```lua
{
  ["@entity"] = "at least one of '', '', '' must be non-empty when updating",
  methods = "field required for entity check when updating",
  hosts = "field required for entity check when updating",
  paths = "field required for entity check when updating",
}
```

In other words, I think the behavior would be:
- In insertion
  - always require at least one of the three
  - error out with the "at least one of..."
- In update
  - if one of the required values is `nil` or `null`, require a non-null value
  - if all values are `nil` or `null`, error out with the more descriptive "field required for entity check when updating" for the user to be less confused
- Remove the existing "field required for entity check" error seems it seems it comes unused? Or update it to the above, more descriptive, update-aware message

This ended up being a bit longer than I thought, but I hope it makes sense.

hishamhm: I definitely appreciate the thought given to the matter!

I modified the behavior to match what you described (removing the existing "field required for entity check" error in those cases, since the overall entity message now does a better job at describing what's going on). I also renamed the test, your description is better.

I got a bit confused about the redundancy of the tests -- I added it because the existing ones didn't cover the behavior I wanted to change. Perhaps it now covers both the behavior I wanted to test and that of another test? In other words, which particular test are you asking to be removed?

kong/db/schema/init.lua
```diff
@@ -64,6 +64,9 @@ local validation_errors = {
   ENTITY_CHECK              = "failed entity check: %s(%s)",
   ENTITY_CHECK_N_FIELDS     = "entity check requires %d fields",
   CHECK                     = "entity check failed",
+  CONDITIONAL               = "failed conditional validation",
+  AT_LEAST_ONE_OF           = "at least one of these fields must be non-empty: %s",
+  ONLY_ONE_OF               = "only one of these fields must be non-empty: %s",
```
thibaultcha: I am slightly confused because I am not seeing `ONLY_ONE_OF` being used anywhere, and `CONDITIONAL` being tested nowhere I think? :fearful: Was `IF_THEN` missing then?
I am afraid I am either missing something, or that this was your intent? I don't know. Could you help clarifying this for me? I am not necessarily seeing this as a blocker, but rather confused and looking for what your intent here is with those fields, and what that means with regards to the maintainability of the schema validation. Or if this is future errors, or anything else really.

spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
```diff
@@ -526,7 +551,7 @@ for _, strategy in helpers.each_strategy() do
           setup(function()
             assert(db:truncate())
 
-            for i = 1, 2000 do
+            for i = 1, 1002 do
```
thibaultcha: :+1: a good move

spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
```diff
@@ -410,10 +430,15 @@ for _, strategy in helpers.each_strategy() do
...
-            }
+            message  = unindent([[
+              schema violation
+              (when updating, at least one of these fields must be non-empty: 'methods', 'hosts', 'paths')
+            ]], true, true),
+            fields   = {
+              ["@entity"] = {
+                "when updating, at least one of these fields must be non-empty: 'methods', 'hosts', 'paths'",
+              }
+            },
```
thibaultcha: Moving the previous conversation here, but not considering it a blocker either (bigger fish to fry):

What I was saying is that it seems that both this test and the one at L.350 look alike:

- They both create a Service with `n` amount of semi-required fields
- They both issue an update which sets those `n` semi-required fields to `null`

This _seems_ to be a case of "1 feature breaks and now N tests are failing" because they were testing the same thing.

Maybe you want to test both one and many semi-optional fields being unset, in which case maybe those names can help:
- `fails trying to unset an interdependent field by itself`
- `fails trying to unset several interdependent fields at once`

Or I am missing something else that differentiates those tests.

Again, only clarifying my previous point, I think the main concern I had with this PR personally was the lack of "when updating ..." context in the error message, which is now addressed.

kong/db/schema/init.lua
hishamhm: they are used here: https://github.com/Kong/kong/pull/3364/files#diff-7135c27fc8fde710af0a85f4dc622d13R757

`IF_THEN` was indeed missing, but it mostly harmless because there's a fallback `"failed entity check: %s(%s)"` message (which adds further data). The idea is that entity checkers (added by us or in the future by others) will always produce messages, whether they return a custom message, a default message, or just a boolean.

spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
hishamhm: Oh, I see.

These two tests as different because they document different aspects of system behavior:

* the test at 350 starts with `hosts[]=example.com methods[]=GET` and tries to update with `methods=null`. A user could reasonably expect this to work, but it doesn't because we don't do read-before-write. This test documents this.

* the test at 418 starts with `hosts[]=example.com methods[]=GET` and tries to update with `hosts=null methods=null`. A user could **not** reasonably expect this to work because it would break the constraint. This test verifies that indeed it does not work.

So, the second test fails for the same reason of the first test, but in the future the system behavior could be changed so that the first test becomes o a success-test but the second will still need to be a fail-test.

kong/db/schema/init.lua
hishamhm: (and just to clarify, the `conditional` entity checker is tested)

spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
thibaultcha: Ok, that is what I thought (that they currently test the same thing), lgtm if intentional then. I think I will just move the tests closer to each other and name them similarly then.

kong/db/schema/init.lua
thibaultcha: Oh I see, it is tested but does not have an assertion for its error message content. Alright, good enough it seems unrelated to the scope of this PR anyway.

thibaultcha:
Alright, merged!

PR #3373, opened 2018-04-03
feat(dao) add 2 cassandra migration helpers

kikito:
Migration helpers for Cassandra extracted from the SSL Certificates-to-new-DAO branch so they can be examined independently.

### DIFF
--- a/kong/dao/migrations/helpers.lua
+++ b/kong/dao/migrations/helpers.lua
@@ -1,8 +1,12 @@
 local json_decode = require("cjson.safe").decode
+local cassandra = require("cassandra")
+local utils = require "kong.tools.utils"
 
 
 local _M = {}
 
+local fmt = string.format
+local table_concat = table.concat
 
 -- Iterator to update plugin configurations.
 -- It works indepedent of the underlying datastore.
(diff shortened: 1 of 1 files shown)


kong/dao/migrations/helpers.lua
```diff
@@ -95,5 +99,205 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+        end
+      end
+
+      local question_marks = string.sub(string.rep("?, ", len), 1, -3)
+
+      local insert_cql = fmt("INSERT INTO %s (%s) VALUES (%s)",
+                             destination_table_def.name,
+                             table_concat(column_names, ", "),
+                             question_marks)
+      local _, err = coordinator:execute(insert_cql, values, nil, "write")
```
thibaultcha: `execute()` from lua-cassandra only takes `(query, args, options)` as arguments, need to remove the two trailing ones, or eventually pass `{ consistency = cql.consistencies.all }` (as do other migrations).

kong/dao/migrations/helpers.lua
```diff
@@ -95,5 +99,205 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+
+The function takes the "source row" as parameter, so it could be used to do things
+like merging two fields together into one, or putting a string in uppercase.
+--]]
+function _M.cassandra.copy_records(dao,
+                                   source_table_def,
+                                   destination_table_def,
+                                   columns_to_copy)
+
+  local coordinator = dao.db:get_coordinator()
```
thibaultcha: I think we need to catch the error potentially returned by `get_coordinator()` here (which is susceptible to a bug, as noticed by @salazar, but this is out of scope for now)

kong/dao/migrations/helpers.lua
```diff
@@ -95,5 +99,205 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+  local cql = fmt("SELECT * FROM %s ALLOW FILTERING", source_table_def.name)
```
thibaultcha: We can get rid of the `ALLOW FILTERING` clause if there is no `WHERE` clause

kong/dao/migrations/helpers.lua
```diff
@@ -95,5 +99,205 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+        key        = "text",
+        created_at = "timestamp",
+      },
+      partition_keys = { "id" },
+    }
+  --]]
+
+  function _M.cassandra.add_partition(dao, table_def)
+
+    local coordinator = dao.db:get_coordinator()
```
thibaultcha: ditto: catching the error

kong/dao/migrations/helpers.lua
```diff
@@ -95,5 +99,205 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+    if err then
+      return nil, err
+    end
+
+    local _, err = _M.copy_records(dao, table_def, aux_table_def, columns_to_copy)
+    if err then
+      return nil, err
+    end
+
+    local _, err = drop_table(coordinator, table_def.name)
```
thibaultcha: should we remove both tables at the last minute, to ensure none of them are lost in case one of the other operation fails?

kong/dao/migrations/helpers.lua
```diff
@@ -95,5 +99,205 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+    if err then
+      return nil, err
+    end
+    error("stop")
```
thibaultcha: seems like a leftover that we'd better remove

kong/dao/migrations/helpers.lua
thibaultcha: Also, we should probably make this `drop_table_if_exists`, to make this migration safe to re-run in case it previously failed halfway through. If it failed after a delete table but before finishing, it seems like it would never be able to fully complete (since all subsequent runs of the migration hit a roadblock trying to delete a non-existing table).

kong/dao/migrations/helpers.lua
thibaultcha: Ok, #3374 will now help us here :)

kikito:
@thibaultcha thanks for the review, all changes have been addressed!

kong/dao/migrations/helpers.lua
```diff
@@ -95,5 +99,213 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+      return nil, err
+    end
+
+    table_def = utils.deep_copy(table_def)
+
+    local aux_table_def = create_aux_table_def(table_def)
+    local columns_to_copy = get_columns_to_copy(table_def)
+    columns_to_copy.partition = function() return cassandra.text(table_def.name) end
+
+    local _, err = create_table(coordinator, aux_table_def)
```
thibaultcha: This helper needs to be able to re-run if it failed halfway through in a previous attempt, just in case. I think this operation should be `create_table_if_exists`, otherwise it would never succeed a re-run, right?

We should fix all other similar places which prevent this migration from being re-run imho.

kikito: Thanks! I've done the change (`create_table_if_not_exists` instead of `create_table_if_exists`).

> This helper needs to be able to re-run

This made me think. If the helper collapses in the middle of the record copying, it will be difficult to recuperate it (the records would have to be removed from the destination database, otherwise they would be duplicated). I see two ways of approaching this:

* Deleting all records from the destination table before copying the records.
* Not do the initial delete, and instead each time a record is copied from source to destination, the record is then removed from source.

The first option seems more conservative because it delays erasing stuff. The second option seems faster.

But there is a catch: if the first option fails precisely because there is "too much data" (i.e. the memory on the coordinator is spent, or too many operations provoke a timeout) then running the migration again won't help, because every time it has to copy all the records. If the second option fails for the same reason, when it is run again the records will already be partially copied, so it is more likely to succeed in case of "too much data" errors.

For this reason I am partial to the second option. But maybe I am overthinking this. Please let me know your thoughts.

kikito:
@thibaultcha implemented your changes, but this:

> This helper needs to be able to re-run

This made me think.

If the migration fails while the record copying is taking place, it will be difficult to recuperate it (the records would have to be removed from the destination database, otherwise they would be duplicated). I see two ways of approaching this:

* Deleting all records from the destination table before copying the records.
* Not do the initial delete, and instead each time a record is copied from source to destination, the record is then removed from source.

The first option seems more conservative because it delays erasing stuff. The second option seems faster.

But there is a catch: if the first option fails precisely because there is "too much data" (i.e. the memory on the coordinator is spent, or too many operations provoke a timeout) then running the migration again won't help, because every time it has to copy all the records. If the second option fails for the same reason, when it is run again the records will already be partially copied, so it is more likely to succeed in case of "too much data" errors.

For this reason I am partial to the second option. But maybe I am overthinking this. Please let me know your thoughts.

thibaultcha:
Mhhh I don't think we should delete data as it is being inserted into the new table. We should allow the migration to re-run as much as we can, but also prevent loss of data in case something goes horribly wrong. We just need to ensure this migration will be both idempotent *and* somewhat re-entrant upon errors. The moving of the data to a new temporary table is already somewhat preventing us from ensuring this, but we can still ensure that the 1st and 2nd halves of the migration can be re-entrant.

If we are already in the second half of the migration (data has been migrated to the temporary table) and we fail there, it seems like the subsequent run will error in the 1st half? The first `create_table_if_not_exists()` will succeed (and not re-create the temporary table), but the subsequent `copy_records()` will fail since the original table does not exist anymore (it was deleted in the previous run when it completed its first half)?

kikito:
> but the subsequent copy_records() will fail since the original table does not exist anymore (it was deleted in the previous run when it completed its first half)?

The migration deletes the original table and then immediately (literally the next step) is re-creating it (with a new `partition` field).

The scenario you are suggesting can happen but only if the failure happens exactly between the destruction of the table and its re-creation on the second half. In that case the first `copy_records` will error.

If the error happens between the table re-creation and the first record is copied "back", then we are re-entrant: the first `copy_records` will find an empty table and not attempt to copy anything. Then it will delete the table and re-create it again.

If the error happens while the second `copy_records` takes place, we also will have a problem: the first `copy_record` will see the records in that table and attempt to "re-copy" them on the "aux" table. This will either raise an error (for example on duplicate unique values) or make copies of registers.

Maybe the right thing to do is doing each `copy_records` on a separate migration. That way once one copy succeeds, we "never have to deal with it again". This would mean leaving `copy_records` on the helper and removing everything else.

thibaultcha:
> the first copy_record will see the records in that table and attempt to "re-copy" them on the "aux" table.

I had the same realization, but was then thinking that if the PK is the same, then it should simply UPSERT. Still not desirable though...

> Maybe the right thing to do is doing each copy_records on a separate migration.

Yes, either this, either making the aux table "special" (via a comment on the table for example), so that the first half only runs if the source table is the original one, and not the aux one.

kikito:
> either making the aux table "special" (via a comment on the table for example),

I tried doing this, but getting the comment back (from an actual cql command, not from the cqlsh console) was difficult and changed between Cassandra versions.

I have done something different: I count the number of rows on the "destination" table before doing copy_rows. If it's greater than 0, I skip `copy_rows` entirely.

This avoids the problems of "copying the same stuff more than once". The two outsanding issues remain:

* If the migration fails between the first `delete_table` and the subsequent `create_table`, the migration won't be re-entrant (there will be an error "table not found").
* If the migration fails while `copy_tables` is being executed, the migration will continue (possibly even succeed) but there will be data losses unless someone goes in and finishes moving the data before running the migration.

I don't know if we can do something regarding those two.

thibaultcha:
It seems like linting is failing now?

kikito:
@thibaultcha linting problem fixed, apologies.

thibaultcha:
tbh, I am very worried about the described approach to decide if a migration is re-entering or not... `copy_records()` is the place where most of this migration will be spending its time at, and is therefore the mist highly likely place where it can fail. And if it does (during the first or second half or the migration), we can lost enormous amounts of data because the next run of the migration considers that step done... Are you not concerned by this?

kikito:
> Are you not concerned by this?

I am, but honestly I don't know what else we can do. Besides going back to the "export to CSV, add column to CSV, import CSV" approach.

thibaultcha:
Alright, merged! :tada:

PR #3374, opened 2018-04-04
fix(dao) get_coordinator return value

salazar:
### Summary

`get_coordinator` would return a second return value (error message) even if `coordinator` was non-nil.


thibaultcha:
Merged, thanks!

PR #3380, opened 2018-04-06
feat(datadog) include tags on request metric

### LINKED ISSUE #2054: Pass consumer_id as a tag in Datadog plugin
### Summary

A follow up to https://github.com/Mashape/kong/issues/998. Inside the Datadog plugin, would it be possible to add an option to place the consumer_id inside a dogstatsd tag rather than the metric name? The advantage is that this makes it easier in Datadog to do things like plot the top 10 API users or worst performing APIs.

So an option meaning `kong.<api_name>.request.count` would end up having a dogstatsd tag with the consumer_id. This could easily be extended to all the other metrics, for example adding this tag to `kong.<api_name>.<http_status_code>.count` would mean it could also be broken down by response code.

mathematician:
### Summary

- Every request_count metric will include a consumer_id tag if available

### Full changelog

- Update datadog request_count schema to include consumer_identifier
- Update datadog request_count migrations to include consumer_identifier
- Update tests to include tag if consumer is used

### Issues resolved

Fix #2054


mathematician:
Is there any way I can get a review on this? I'm not sure who maintains plugins like this. @thibaultcha 

I'm fine with needing to provide updates for this functionality as well, just haven't heard back from anyone.

kikito:
Closing this in favor of #5154 , which implements similar changes.

PR #2675, opened 2017-07-06
add extend points in nginx/kong conf templates

passos:
### Summary

Add three extendsion points in nginx configuration template so that user can put their extend configuration in their own directory instead of update the template file in system library.

### Full changelog

* kong/templates/nginx.lua
* kong/templates/nginx_kong.lua

PR #3382, opened 2018-04-06
feat (config) allow the user the ability to change

maguec:
### Summary

large_client_header_buffers nginx configuration option in kong.conf

Nginx returns "400 Request Header Or Cookie Too Large" when cookies are
too large, so the large_client_header_buffers needs to be configured to
allow large cookie request through Kong.

### Full changelog

* set nginx default in kong_defaults.lua
* allow overrides in the templating
* tests to ensure that changes git picked up

### DIFF
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -67,6 +67,8 @@ local CONF_INFERENCES = {
   real_ip_header = {typ = "string"},
   real_ip_recursive = {typ = "ngx_boolean"},
   client_max_body_size = {typ = "string"},
+  large_client_header_buffer_count = {typ = "number"},
+  large_client_header_buffer_size = {typ = "string"},
   client_body_buffer_size = {typ = "string"},
   error_default_type = {enum = {"application/json", "application/xml",
                                 "text/html", "text/plain"}},
--- a/kong/templates/kong_defaults.lua
+++ b/kong/templates/kong_defaults.lua
@@ -31,6 +31,8 @@ trusted_ips = NONE
 real_ip_header = X-Real-IP
 real_ip_recursive = off
 client_max_body_size = 0
+large_client_header_buffer_count = 4
+large_client_header_buffer_size = 8k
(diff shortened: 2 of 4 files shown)


thibaultcha:
Thanks for the suggestion,

Unfortunately, I am afraid this falls under the same category as #3010, #3323, and a few other ones. See [this comment/thread](https://github.com/Kong/kong/pull/3010#issuecomment-342708588) for context on our current stance with regards to customizing the nginx directives from the Kong template.

We'd rather introduce a nicer way to solve this issue altogether (otherwise, the list of configuration variables will grow out of control). Our options are:

1. #2355 (the cleanest imho)
2. Hooks into the nginx configuration, such as proposed by #2675
3. Another generic way of overriding _any_ nginx directive from `kong.conf`. I have given this some thought as a temporary solution, or compromise until/if we ever get to 1.

thibaultcha:
Another approach we have been discussing with @maguec that does not involve managing different configuration files (as per suggestion 2.), and still allows for granular nginx directive override (with both `kong.conf` and environment variables), is the declaration of "flexible" configuration values in `kong.conf` (therefore picked up by env variables as well):

There would exist at least 3 prefixes that I can think of for now, in order to inject those directives in various levels of the nginx configuration:
```
# injected into the http {} block
nginx_directive_http_<directive_name> = <directive value>

# injected into the proxy server {} block
nginx_directive_proxy_<directive_name> = <directive value>

# injected into the admin server {} block
nginx_directive_admin_<directive_name> = <directive value>
```

Translated to environment variables, those would give:
```
KONG_NGINX_DIRECTIVE_HTTP<DIRECTIVE_NAME>=<directive value>
```

This addition would require nginx template changes (as would the proposal 2. above), and some changes in the configuration loader (`conf_loader.lua`).

thibaultcha:
Closing this since some internal decisions and progress has been made on the above proposal. Thanks for the suggestion @maguec

PR #3383, opened 2018-04-07
Started at enclosed in ""

harryparmar:
Sending logs to elastic search through HTTP-LOG plugin, don't work with Kibana because timestamp is not able to work with default started at variable in pipeline that was setup in Elastic Search. Doing debug in elastic search we can see problem it is expecting a string. I have been manually adding this code since version 8, it will be nice if Kong can add this to official distribution.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX

### DIFF
--- a/kong/plugins/log-serializers/basic.lua
+++ b/kong/plugins/log-serializers/basic.lua
@@ -43,7 +43,7 @@ function _M.serialize(ngx)
     api = ngx.ctx.api,
     consumer = ngx.ctx.authenticated_consumer,
     client_ip = ngx.var.remote_addr,
-    started_at = ngx.req.start_time() * 1000
+    started_at = ngx.req.start_time() * 1000,
     started_at_unix_ms = "" .. ngx.req.start_time() * 1000 .. ""
   }
 end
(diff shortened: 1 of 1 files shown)


hbagdi:
@harryparmar Thanks for opening this PR but did you try to convert the field to string using a ruby filter in logstash?

harryparmar:
@hbagdi dont use logstash I have pipeline on ES itself. we have apis that do more than million calls per hour would prefer if variable was available from source in right format rather than doing these conversions and thus stressing indexing? If this cant be done not big deal i can add the variable on our local instance as currently i have been doing

thibaultcha:
Thanks for the patch.

This is similar to previous proposals we saw, and decided to reject under the rationale that we do not want to duplicate information contained in the log object for the purpose of the presentation layer. It is your backend's job (or log collector) to transform the data sent by Kong in whatever format/type makes sense for you.

harryparmar:
@thibaultcha @hbagdi  that makes sense, realized after making PR that from your (Kong) point of view this is something that should be handled locally.

thibaultcha:
Alright, closing this then :)

PR #3386, opened 2018-04-09
feat(dao) move ssl certificates/snis to new DAO

kikito:
This PR continues #3315:
### Migrations
* `ssl_certificates` table renamed to `certificates`
* `ssl_servers_names` table renamed to `snis`. Added a new field (`id`, UUID) and made that the PK. But also added indexes for the `name` field.
* In postgresql these changes involved altering the tables. In Cassandra the changes involved copying all the records from the original tables to new tables (this was done with a custom migration helper which has its own PR).

### Admin API
* `/snis` is automatically generated by the new DAO admin api generator-from-schema.
* `/certificates` is also generated, but it has some overrides, to deal with three "special" things: 
  * the `server_names` accepts a comma-separated string or an array of strings. It
    allows "setting the server names in one go" when creating (POST) or updating 
    (PATCH) a certificate.
  * the `server_names` attribute is automatically added to the certs when getting
    each one individually or when getting a list of them.
  * a certificate can be referenced by its id or also by the `name` attribute of
    any of its associated `server_name`s.
* `PUT` is not allowed any more.
* Deactivates some `/snis` endpoints that are not needed any more:
  * `GET /snis` is not needed any more now that `GET /certificates` also returns the list of server names associated to every cert
  * `POST` /snis` is not needed since `POST /certificates/id/server_names` is easier to use and does the same
  * `GET /snis/[name]/certificate was not needed since you can do `GET /certificates/[name]` to get the same info
  * The only active server names endpoints are `DELETE /snis/x` (to remove a server name quickly), `PATCH /snis/x` (to move server names quickly between certificates), `GET /snis/x` (for symmetry with DELETE and PATCH).

### DAO
* `ssl_certificates` and ssl_servers_names' schemas translated to new DAO's schemas (renamed to `certificates` and `snis`).
* I had to include a read-before-delete in all the generated `dao:delete*` methods. This read was needed in order to propagate the structure through the worker events so it could be deleted properly (worker events can use any attribute
from the entitity, for example for the server names you need their `name` attribute. So if you delete a server name using its `id` you still need to read it in order to propagate an event with the `name` field).
* `certificates` has a custom DAO to deal with the `server_names` pseudo-attribute and for "looking for a certificate using one of its server names' `name`"
* `server_names` also has a custom DAO for "using the server names as if they were lists of strings", which is easier to do than "handling them as instances of the `server_names` schema".

### Tests
* the `DAO:delete*` methods now returns the entity which was erased instead of `true`, so some tests had to be changed from `is_true` to `is_truthy`.
* the tests for `PUT` where just removed
* I also removed the admin api tests from spec-old-api because they were a carbon copy of the new ones (there are no APIs, Services or Routes involved in SSL). I double-checked that we don't do anything "extra" on it before removing.
* Added tests for the new admin API endpoints.

### Related work
* The migrations are tested in https://github.com/Kong/kong-upgrade-tests/pull/5
* Since we're doing the read-before-delete on the DAO we probably don't need the `delete_by_field` methods on the strategies (we can use the PK, we have it because of the "read"). We could remove those methods.


kong/api/routes/certificates.lua
```diff
@@ -1,317 +1,80 @@
-local crud = require "kong.api.crud_helpers"
-local utils = require "kong.tools.utils"
-local cjson = require "cjson"
+local endpoints   = require "kong.api.endpoints"
+local utils       = require "kong.tools.utils"
```
bungle: One more new line here.

kong/api/routes/certificates.lua
```diff
@@ -1,317 +1,80 @@
...
 
+local function get_cert_by_server_name_or_id(self, db, helpers)
+  local id = self.params.certificates
```
bungle: New line maybe not needed, similar to pattern:
```
local ok, err = func()
if not ok then
end
```

E.g. where test follows immediately the declaration.

kong/api/routes/server_names.lua
```diff
@@ -0,0 +1,16 @@
+local function method_not_allowed(self, db, helpers)
+  return helpers.responses.send_HTTP_METHOD_NOT_ALLOWED()
+end
+
+return {
+  -- GET / PATCH / DELETE /server_names/server_name are the only methods allowed
+
+  ["/server_names"] = {
+    before = method_not_allowed,
```
bungle: If they are not allowed, is `404` a better one? E.g. is there any endpoint `/server_names` that you can call with any method? If not with any, then I would go with `404`. We could also add some support for autogenerator to remove endpoints, but not in this PR.

kong/api/routes/server_names.lua
```diff
@@ -0,0 +1,16 @@
...
+  },
+
+  ["/server_names/:server_names/certificate"] = {
+    before = method_not_allowed,
```
bungle: Same here, if this is the only:
/server_names/server_name

Endpoint, maybe we don't want to give reply that says that it is not allowed, users might try then other methods, and these seems no method allowed here. So I think it would be better to just have `404`.

kong/dao/migrations/postgres.lua
```diff
@@ -692,4 +694,76 @@ return {
...
+        return err
+      end
+
+      local fmt = string.format
+
+      for _, row in ipairs(rows) do
+        local sql = fmt("UPDATE server_names SET id = '%s' WHERE name = '%s';",
+                        utils.uuid(),
+                        row.name)
+        local _, err = dao.db:query(sql)
```
bungle: Instead of executing these one-by-one, could up make up an transaction and do them in one `query` that is wrapped in transaction?

kong/db/dao/certificates.lua
```diff
@@ -0,0 +1,209 @@
...
+local utils      = require "kong.tools.utils"
+
+-- Get an array of server names from either a string (split+sort),
+-- an array(sort) or ngx.null(return {})
+-- Returns an error if the list has duplicates
+-- Returns nil if input is falsy.
+local function parse_name_list(input, errors)
+  local name_list
+  if type(input) == "string" then
+    name_list = utils.split(input, ",")
```
bungle: new line after this

kong/db/dao/certificates.lua
```diff
@@ -0,0 +1,209 @@
...
+  elseif type(input) == "table" then
+    name_list = utils.shallow_copy(input)
```
bungle: new line after this

kong/db/dao/certificates.lua
```diff
@@ -0,0 +1,209 @@
...
+  cert, err, err_t = assert(self:insert(cert))
+  if not cert then
+    return nil, err, err_t
+  end
+  cert.server_names = name_list or cjson.empty_array
+
+  if name_list then
+    local ok, err, err_t = db.server_names:insert_list({id = cert.id}, name_list)
+    if not ok then
+      return nil, err, err_t
```
bungle: This leads to interesting situations, like certificate added, but no server_names added. Do we have any concerns of that?

kong/db/dao/init.lua
```diff
@@ -158,6 +166,7 @@ function _M.new(schema, strategy, errors)
     schema   = schema,
     strategy = strategy,
     errors   = errors,
+    super    = DAO, -- allows custom daos to do self.super.delete(self, ...)
```
bungle: I see, but this feels like OO terminology. Maybe just call this `dao`.

kong/db/dao/server_names.lua
```diff
@@ -0,0 +1,159 @@
...
+-- Truthy if all the names on the list don't exist on the db or exist but are
+-- associated to the given certificate
+function _ServerNames:check_list_is_new_or_in_cert(cert_pk, name_list)
+  for i=1, #name_list do
+    local row, err, err_t = self:select_by_name(name_list[i])
+    if err then
+      return nil, err, err_t
+    end
+    if row and row.certificate.id ~= cert_pk.id then
+      local msg = "Server Name '" .. row.name ..
```
bungle: On line 47 it says `"Server name already exists` and here `Server Name `. I think in general these messages should be all lower case (because they are injected usually in different places), comments on that?

kong/db/init.lua
```diff
@@ -49,10 +52,10 @@ function DB.new(kong_config, strategy)
       local entity_schema = require("kong.db.schema.entities." .. entity_name)
 
       -- validate core entities schema via metaschema
-      local ok, err = MetaSchema:validate(entity_schema)
+      local ok, err_t = MetaSchema:validate(entity_schema)
       if not ok then
         return nil, fmt("schema of entity '%s' is invalid: %s", entity_name,
-                        err)
+                        pl_pretty.write(err_t))
```
bungle: Do we want this, or was this for debugging?

kong/api/routes/certificates.lua
kikito: This is done

kong/db/dao/init.lua
kikito: I see what you say but I think calling it `dao` would also be confusing as well ("I'm using the DAO while building my DAO").

May I suggest `base` instead?

kong/db/dao/certificates.lua
kikito: I would rather not have that situation, but I honestly didn't know what else to do (we don't have transactions at this level, because Cassandra).

This code reflects what the old DAO code did - it makes as much "checks" as possible before writing anything to the db.

kikito:
Thanks for the review @bungle I have added changes for most things. Answering to your general comments: My initial PR was much smaller - it moved only `certificates` to the new DAO, leaving `ssl_server_names` in the old DAO and not touching the admin API. But I was told to include all the other changes as well.

Both you and @hishamhm have mentioned that you prefer other name instead of `server_names`. I can rename `server_names` to `snis` everywhere (table name, DAO, admin api). Would you consider this better?

hishamhm:
> Both you and @hishamhm have mentioned that you prefer other name instead of server_names. I can rename server_names to snis everywhere (table name, DAO, admin api). Would you consider this better?

This would make the endpoints more compatible with the current ones, right? If so, I think it would be a better idea. We can think of better names when/if we merge the entities in the future.

kong/db/dao/init.lua
kikito: (we spoke offline and we decided to leave it as `super`)

kikito:
Renamed `server_names` to `snis` pretty much everywhere, tests passing.

kong/db/dao/certificates.lua
```diff
@@ -0,0 +1,211 @@
...
+  for _, name in ipairs(name_list) do
+    if found[name] then
+      local msg   = "duplicate sni in request: " .. name
+      local err_t = errors:invalid_input(msg)
+      return nil, tostring(err_t), err_t
+    end
+    found[name] = true
+  end
+
+  table.sort(name_list)
```
thibaultcha: So we wish to sort the snis alphabetically?

kong/db/dao/certificates.lua
```diff
@@ -0,0 +1,211 @@
...
+  elseif type(input) == "table" then
+    name_list = utils.shallow_copy(input)
+
+  elseif input == ngx.null then
+    name_list = {}
+  end
+
+  if not name_list then
+    return nil
+  end
```
thibaultcha: The code would be simplified if we avoided returning `nil` from this function, and instead always return a table (empty if no names):

```lua
if type(input) == "table" then
  name_list = utils.shallow_copy(input)

elseif input == ngx.null then
  name_list = {}

else
  -- maybe throw an error here, looks like a programming error
end

-- ...

return setmetatable(name_list, cjson.empty_array_mt)
```

This way, the below `insert_with_name_list()` method could do: 
```lua
if #name_list > 0 then

end

-- ...

cert.snis = name_list -- no need for cjson.empty_array fallback
```

kong/db/dao/snis.lua
```diff
@@ -0,0 +1,158 @@
...
+  local rows, err, err_t = self:for_certificate(cert_pk)
+  if err then
+    return nil, err, err_t
+  end
+  for i = 1, #rows do
+    name_list[i] = rows[i].name
+  end
+
+  table.sort(name_list)
+  return name_list
```
thibaultcha: We should set the `cjson.empty_array_mt` metatable here, otherwise the Admin API will JSON-encode this as `{}` instead of `[]`.

kong/db/dao/certificates.lua
```diff
@@ -0,0 +1,211 @@
...
+
+
+local _Certificates = {}
+
+-- Creates a certificate
+-- If the provided cert has a field called "snis" it will be used to generate server
+-- names associated to the cert, after being parsed by parse_name_list.
+-- Returns a certificate with the snis sorted alphabetically.
+function _Certificates:insert_with_name_list(cert)
+  local db = singletons.db
```
thibaultcha: This is one of my biggest concerns with this PR. We need to find a way to avoid this at **all costs**.

kong/db/dao/init.lua
```diff
@@ -352,14 +361,22 @@ function DAO:delete(primary_key)
     return nil, tostring(err_t), err_t
   end
 
+  local entity, err, err_t = self:select(primary_key)
+  if err then
+    return nil, err, err_t
+  end
+  if not entity then
+    return 0
```
thibaultcha: Our rationale was to not return the number of affected rows from `delete()` (because of the lack of support from Cassandra for this). Is the rationale behind this return value to indicate that no row was deleted?

kong/db/errors.lua
```diff
@@ -24,14 +24,16 @@ end
...
-  DATABASE_ERROR        = 8, -- connection refused or DB error (HTTP 500)
+  INVALID_PRIMARY_KEY   =  1,
+  SCHEMA_VIOLATION      =  2,
+  PRIMARY_KEY_VIOLATION =  3, -- primary key already exists (HTTP 400)
+  FOREIGN_KEY_VIOLATION =  4, -- foreign entity does not exist (HTTP 400)
+  UNIQUE_VIOLATION      =  5, -- unique key already exists (HTTP 409)
+  NOT_FOUND             =  6, -- WHERE clause leads nowhere (HTTP 404)
+  INVALID_OFFSET        =  7, -- page(size, offset) is invalid
+  DATABASE_ERROR        =  8, -- connection refused or DB error (HTTP 500)
+  CONFLICTING_INPUT     =  9, -- user-provided data generated a conflict (HTTP 409)
```
thibaultcha: Let's maybe make the description more precise: "a conflict _with already inserted DB values_"?

I recall discussing this with @bungle, and since this will result from `UNIQUE` constraints set in PostgreSQL, we wanted to name this value `UNIQUE_VIOLATION` to stay consistent with other `_VIOLATION` error types we already defined.

Since the certs/sni entities are already abnormal in the sense that they validate their own values and produce their own errors, I would rather avoid spreading that abnormality to the errors as well. So in other words, the conflict errors produced by certs/sni validations are not because of DB/schema UNIQUE constraints, but I don't think that matters and I think they can throw `UNIQUE_VIOLATION` which has the same semantics.

kong/api/routes/certificates.lua
```diff
@@ -1,317 +1,80 @@
...
-        return -- avoid tail call
-      end
-
-      -- id present in body
-      update_certificate(self, dao_factory, helpers)
+      return helpers.responses.send_HTTP_CREATED(data)
     end,
   },
 
+  ["/certificates/:certificates"] = {
```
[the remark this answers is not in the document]
bungle: I think you need to use what @kikito used to override autogenerated.

kong/db/errors.lua
```diff
@@ -24,14 +24,16 @@ end
...
+  INVALID_INPUT         = 10, -- user-provided data is not valid (HTTP 400)
```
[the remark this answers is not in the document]
hishamhm: > This looks conflicting with the above `SCHEMA_VIOLATION` error

I have mixed feelings about reusing the error messages for errors detected via different sources.

On the one hand, for the user perspective I agree that the way bad input parameters were caught shouldn't affect the error message, it's an implementation detail. On the other hand, debugging gets harder when we get reports of `SCHEMA_VIOLATION` from the field and these don't come from the schema library.

I don't think it's the case for certificates, but if something like this starts happening in other parts of the code it could easily trip up plugin authors ("but my schema looks okay!"). But as you said, certs/snis are "abnormal" in that sense, so it's probably fine to make an exception for it in the name of better user-visible consistency.

In that case, the `SCHEMA_VIOLATION` should be formatted in a way consistent with those produced by the schema library (e.g. listing the offending fields, etc).

> I also dislike that this error producer accepts any error message

Definitely agree with this. It shouldn't just forward the full error message, it should construct a message like the other functions in this module.

kong/api/routes/certificates.lua
```diff
@@ -1,317 +1,80 @@
...
-
-    DELETE = function(self, dao_factory, helpers)
-      return crud.delete({
-        id = self.ssl_certificate_id
-      }, dao_factory.ssl_certificates)
-    end,
-  }
+  ["/certificates/:certificates/snis"] = {
+    before = get_cert_by_sni_or_id,
+  },
```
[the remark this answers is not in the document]
kikito: You can do `GET` and `POST` on it.
```
GET /certificates/xxx/snis
POST /certificates/xxx/snis { name: "potato.com" }
```

`GET` lists the "real sni entities" associated with the certificate, not "magical strings". It's the only place where you can get all the ids of the snis associated to a cert via the admin api.

`POST` allows adding a single sni to the existing list of snis (so if a cert has 100 snis, you can use this instead of doing a PATCH on the cert and passing 101 snis as an array).

kong/api/routes/certificates.lua
```diff
@@ -1,317 +1,80 @@
...
-  -- update certificate if necessary
-  if self.params.key or self.params.cert then
-    self.params.created_at = ssl_cert.created_at
-
-    ssl_cert, err = dao_factory.ssl_certificates:update(self.params, {
-      id = self.params.id,
-    }, { full = true })
-    if err then
-      return helpers.yield_error(err)
+local function get_cert_by_sni_or_id(self, db, helpers)
```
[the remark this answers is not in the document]
kikito: What this function does is changing the params called `certificates` (as in `/certificates/:certificates`). If the param looks like a uuid it does nothing. Otherwise, it is assumed that the param is an sni name. If the sni for that name is found, the `certificates` param takes the value of the sni's `certificate_id`.

I have renamed the function to `get_cert_id_from_sni`, probably this will make it a bit more clear.

kong/api/routes/certificates.lua
```diff
@@ -1,317 +1,80 @@
...
-          return helpers.responses.send_HTTP_NOT_FOUND()
-        end
+    -- override to include the snis list when getting an individual certificate
+    GET = function(self, db, helpers)
+      local pk = { id = self.params.certificates }
 
-        -- cache certificate row id
-
-        self.ssl_certificate_id = row.ssl_certificate_id
+      local cert, _, err_t = db.certificates:select_with_name_list(pk)
```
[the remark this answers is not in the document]
kikito: You are right, this was doing an extra select that was not needed. This has been fixed, thanks.

kong/db/dao/certificates.lua
kikito: The reason I sort them out is that when I added ids to Cassandra the SNIs started being returned in random order (because the random uuid generation). This made writing the tests very painful, and I thought it would confuse users as well.

kong/api/routes/snis.lua
```diff
@@ -1,53 +1,16 @@
...
-    POST = function(self, dao_factory)
-      crud.post(self.params, dao_factory.ssl_servers_names)
-    end,
+  ["/snis"] = {
+    before = not_found,
   },
 
+  ["/snis/:snis/certificate"] = {
+    before = not_found,
+  },
```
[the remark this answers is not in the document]
bungle: Yes, currently the custom endpoints supports:
1. overriding (and optionally calling original, send as a parameter to customized)
2. adding new

But it has no mechanism to remove routes. On option could be that custom endpoint returns:
```lua
return {
  -- overrides
  ["/services"] = {
    POST = function(self, _, _, parent)
      api_helpers.resolve_url_params(self)
      return parent()
    end,
  },
  remove = {
    "/services/:services",
  }
}
```

Other option is to have some hints to endpoint auto generator, possibly in schemas to not generate something in a first place. E.g. here schema could have:

```lua
fields = {
  { id           = typedefs.uuid, },
  { name         = { type = "string", required = true, unique = true }, },
  { created_at   = { type = "integer", timestamp = true, auto = true }, },
  { certificate  = { type = "foreign", reference = "certificates", required = true, admin_api = false }, },
},
```

We can think about this a bit. I think originally people were not that convinced about the current way to override, but at that point we decided to left it on table.

kong/db/dao/certificates.lua
kikito: The idea I was going for was this:

* If you don't set the `snis` variable at all, it remains as it was (no programming error, maybe you where just missing a char on the key and are updating just that).
* If you set the `snis` variable, it will create/delete whatever snis is necessary. This includes passing an empty table (all snis will be removed)
* If you pass a json null, it is as if you passed an empty table.

If I do the change you propose, there are two drawbacks:
* Either we consider that an empty table means "no changes" (which I think is confusing).
* Or we require the user to pass the `snis` param every time they want to make a change on a cert, even for changing the cert or key (otherwise we set of an error).

If you think this is worth it, I will implement it, but I wanted to warn you about the costs first.

kong/db/dao/certificates.lua
kikito: I think I can change `kong.db.dao.init` so that `db` is an attribute of `self` here, so we can do `self.db.snis:whatever` instead of `singletons.db.snis:whatever`. Would that be ok?

kong/db/dao/init.lua
kikito: Yes. With the read-before-delete you know which rows will (should) be affected by a delete. The `0` means "I could not find any rows". I can replace it with a `true` if you prefer.

kong/api/routes/certificates.lua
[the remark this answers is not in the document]
kikito: I confirm that is the case. The auto-generated route is `"certificates/certificates"` so I must use the same name in order to override stuff.

kong/db/dao/snis.lua
kikito: Done! (at the top of the function)

kong/db/errors.lua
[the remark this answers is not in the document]
kikito: I have removed both my custom errors and used schema violations everywhere. All the errors happen on the context of the `snis` magical pseudo-attribute, so it does not look bad (all errors appear "scoped" by `snis`).

kong/api/routes/certificates.lua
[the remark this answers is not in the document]
kikito: That is exactly the reason. The auto-generated route is `/certificates/:certificates`.

We would have had to add a param called "singular_name" or something similar to the schema in order to be able to do fancy things like sometimes using singulars instead of plurals everywhere.

kong/db/errors.lua
kikito: I have chosen to use schema violations instead. The errors always happen in the context of the `snis` magical meta-param, so it "fits" to just use schema violations for those (so they are scoped inside `snis`).

kikito:
I made a change that makes `kong/admin/routes/certificates.lua` smaller: instead of "adding new methods" to the Certificates DAO (like `Certificates:page_with_name_list` or `Certificates:insert_with_name_list`) I just overrode the default (`Certificates:page` and `Certificates:insert`). These are the methods the generated admin api uses, so I didn't have to override those any more.

I did not dare changing `Certificates:select` because it is used in other places besides the Admin API, generating more queries for getting all the SNIS every time a cert was selected. So I still had to override one of the generated methods.

kong/dao/migrations/cassandra.lua
```diff
@@ -628,5 +631,116 @@ return {
...
+        name text,
+        certificate_id uuid,
+        created_at timestamp,
+        PRIMARY KEY (partition, id)
+      );
+
+      CREATE INDEX IF NOT EXISTS snis_name_idx ON snis(name);
+      CREATE INDEX IF NOT EXISTS snis_certificate_id_idx ON snis(certificate_id);
+    ]],
+    down = nil
```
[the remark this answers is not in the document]
bungle: @salazar I made other changes to this except this. I see it is always "defined" in other places, sometimes with empty function and sometimes with `nil`. Go figure. Didn't feel too bad for leaving them, :-)

kong/dao/migrations/helpers.lua
```diff
@@ -100,6 +100,219 @@ function _M.plugin_config_iterator(dao, plugin_name)
...
+    end
+
+    local _, err = drop_table_if_exists(coordinator, aux_table_def.name)
+    if err then
+      return nil, err
+    end
+  end
+
+end
+
```
thibaultcha: This entire file has duplicate contents (open it locally on this branch) this absolutely needs to be fixed prior to the merge.

bungle:
The `PUT` was moved to later work, and it is not blocking this merge.

Ehekatl:
```POST /snisis not needed sincePOST /certificates/id/server_names` is easier to use and does the same
```

where I can find the related doc?

PR #3388, opened 2018-04-10
fix(dao) properly serialize string booleans

nijikokun:
### Summary

Correctly serialize string based boolean values for Cassandra.

### Issues resolved

Filtering on boolean values through API queries would always result in a
list of truthy based results. This is because the value passed to
serialize_arg is of type string. This can be reproduced by issuing a
request against entities with true fields with a query filter of false.

### DIFF
--- a/kong/dao/db/cassandra.lua
+++ b/kong/dao/db/cassandra.lua
@@ -324,9 +324,9 @@ local function serialize_arg(field, value)
   elseif field.type == "boolean" then
     if type(value) == "boolean" then
       return cassandra.boolean(value)
-    else
-      return cassandra.boolean(value == 'true')
     end
+
+    return cassandra.boolean(value == "true")
   elseif field.type == "table" or field.type == "array" then
     return cjson.encode(value)
   else
--- a/kong/dao/db/cassandra.lua
+++ b/kong/dao/db/cassandra.lua
@@ -322,7 +322,11 @@ local function serialize_arg(field, value)
   elseif field.type == "timestamp" then
     return cassandra.timestamp(value)
   elseif field.type == "boolean" then
-    return cassandra.boolean(value)
+    if type(value) == "boolean" then
+      return cassandra.boolean(value)
+    else
+      return cassandra.boolean(value == 'true')
+    end
   elseif field.type == "table" or field.type == "array" then
     return cjson.encode(value)
   else
--- a/spec/02-integration/03-dao/03-crud_spec.lua
+++ b/spec/02-integration/03-dao/03-crud_spec.lua
@@ -407,11 +407,14 @@ helpers.for_each_dao(function(kong_config)
 
       local run_boolean_checks = true
 
+      -- Filtering on non-primary key columns only supported in C* 3.5+
+      -- Since we do not maintain an index on https_only we can
+      -- only run the boolean checks against Cassandra 3.5+
       if kong_config.database == "cassandra" then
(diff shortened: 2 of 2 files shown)


kong/dao/db/cassandra.lua
```diff
@@ -322,7 +322,11 @@ local function serialize_arg(field, value)
   elseif field.type == "timestamp" then
     return cassandra.timestamp(value)
   elseif field.type == "boolean" then
-    return cassandra.boolean(value)
+    if type(value) == "boolean" then
+      return cassandra.boolean(value)
+    else
+      return cassandra.boolean(value == 'true')
```
thibaultcha: style: we should stick to double quotes and follow the early-exit paradigm we try to enforce for short if/else branches. 
```lua
if type(value) == "boolean" then
  return cassandra.boolean(value)
end

return cassandra.boolean(value == "true")
```

It is fine to have this fix here considering this is legacy code, but part of the reason why this doesn't really belong here (cc @salazar) is because when given a value like `bool_attr="hello"`, we will return rows with `bool_attr=false`, instead of returning HTTP 400, which can be surprising.

spec/02-integration/03-dao/03-crud_spec.lua
```diff
@@ -326,10 +326,15 @@ helpers.for_each_dao(function(kong_config)
...
+          local https_only = true
+          if i == 5 then
+            https_only = false
+          end
           local api, err = apis:insert {
             name = "fixture_" .. i,
             hosts = { "fixture" .. i .. ".com" },
-            upstream_url = "http://fixture.org"
+            upstream_url = "http://fixture.org",
+            https_only = https_only
```
thibaultcha: can just be `https_only = i == 5` but that's fine

spec/02-integration/03-dao/03-crud_spec.lua
```diff
@@ -402,7 +407,15 @@ helpers.for_each_dao(function(kong_config)
...
         assert.same(first_api, rows[1])
       end)
-      it("filter supports a boolean value", function()
+      it("filter supports a truthy boolean value", function()
+        local rows, err, _ = apis:find_page {
+          name = "fixture_5",
+          https_only = "true"
+        }
+        assert.falsy(err)
+        assert.is_table(rows)
```
thibaultcha: Now that we test both values in boolean attributes, and to avoid the below test to produce a false positive (returning values with `https_only=true`, we really should update both of these to ensure that all fetched rows indeed have `https_only` set to `true` or `false`, accordingly.

kong/dao/db/cassandra.lua
nijikokun: This is a fair assessment, I wasn't sure where we wanted to stand with strings that aren't equal to their counterpart so took the safest approach, I was also unsure about whether we wanted to support another paradigm you see in boolean based query strings values such as using `1` and `0`.

spec/02-integration/03-dao/03-crud_spec.lua
nijikokun: I support this idea, that was the reason for ensuring only one row has a true value. Ensuring the values would be a great step further.

spec/02-integration/03-dao/03-crud_spec.lua
nijikokun: Correct, we can use `i ~= 5` here.

PR #3408, opened 2018-04-23
Feat/consumers 2 db

kikito:
Moves consumers to the new DAO. Notes:

* Includes @bungle's changes to `endpoints.lua` that allow param parsing using schemas (https://github.com/Kong/kong/compare/feat/admin-schema-aware-parser), so that one should probably be merged before this PR.
* The consumers admin api has been replaced by auto-generated endpoints, except for plugin-related endpoints:
  * /consumers/:username_or_id/plugins/
  * /consumers/:username_or_id/plugins/:id
* `PUT` is gone (for now, @bungle is working on adding it back)
* "filtering via fields" (`/consumers/?created_at=xxx`) is no longer possible


kong/api/routes/consumers.lua
```diff
@@ -1,38 +1,6 @@
...
-    PATCH = function(self, dao_factory)
-      crud.patch(self.params, dao_factory.consumers, self.consumer)
-    end,
-
-    DELETE = function(self, dao_factory)
-      crud.delete(self.consumer, dao_factory.consumers)
-    end
-  },
 
   ["/consumers/:username_or_id/plugins/"] = {
```
bungle: Can this be changed to `/consumers/:consumers/plugins` for future (not also that no `/` at the end)?

kong/db/schema/entities/consumers.lua
```diff
@@ -0,0 +1,18 @@
+local typedefs = require "kong.db.schema.typedefs"
+
+return {
+  name        = "consumers",
+  primary_key = { "id" },
```
bungle: here we need to add ` endpoint_key = "username",` (see feat/admin-schema-aware-parser)

kong/api/routes/consumers.lua
```diff
@@ -1,40 +1,8 @@
...
-      crud.patch(self.params, dao_factory.consumers, self.consumer)
-    end,
-
-    DELETE = function(self, dao_factory)
-      crud.delete(self.consumer, dao_factory.consumers)
-    end
-  },
 
-  ["/consumers/:username_or_id/plugins/"] = {
+  ["/consumers/:username_or_id/plugins"] = {
```
bungle: This should also be `/consumers/:consumers/plugins` for future changes, when plugins are transferred (probably at that point this needs to be removed).

kikito: I have removed the extra "/". Thanks!

kikito: I also renamed `username_or_id` to `consumers` on the route.

PR #3414, opened 2018-04-24
feat(tools) add `verbatim` parameter to `responses.send`

salazar:
### Summary
`send_response` unconditionally encodes responses as JSON. This PR adds support for verbatim responses, extending the `send` function with a 4th argument `verbatim`, which, if true, results in the `body` argument being interpreted as a string and returned verbatim in the response body.

### DIFF
--- a/kong/tools/responses.lua
+++ b/kong/tools/responses.lua
@@ -97,7 +97,7 @@ local response_default_content = {
 -- Return a closure which will be usable to respond with a certain status code.
 -- @local
 -- @param[type=number] status_code The status for which to define a function
-local function send_response(status_code)
+local function send_response(status_code, verbatim)
   -- Send a JSON response for the closure's status code with the given content.
   -- If the content happens to be an error (500), it will be logged by ngx.log as an ERR.
   -- @see https://github.com/openresty/lua-nginx-module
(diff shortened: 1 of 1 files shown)


p0pr0ck5:
Just wondering, might it allow for more flexibility if the response generator could accept many content types, similar to `kong.core.error_handlers`?

salazar:
@p0pr0ck5 the closure returned by `send_response` takes in a headers table, over which we can pass the content type header.

p0pr0ck5:
Yes but the body formatting is handled differently in each case, no (as evidenced by this PR)?

salazar:
@p0pr0ck5 what this PR seeks is allowing the caller to return a certain response as is - without it being modified in any way (like core error handlers do, by applying a certain template depending on the response type); so it's that exact body formatting that it avoids

p0pr0ck5:
Yep, I understand that :)

My question was, for consistency sake (as well as for future flexibility), should Kong-synthesized responses support a variety of content types? For example, I have a custom plugin that writes out data in XML- it would be great if I could use this module to do that, rather than writing my own content writer and duplicating a bunch of code. Again, just thinking out loud here.

salazar:
@p0pr0ck5 that would be a nice contribution - providing Kong-synthesized, templated, responses _and_ a verbatim mode - e.g., to be used in plugins that want absolute control over the response ! :)

thibaultcha:
We are working on the plugins SDK these days which will have response-producing capabilities. Fir this particular PR, better to use `ctx.delayed_response_callback` as it was implemented for this precise use-case, and this PR is a re-implementation of it (but without delayed responses which is a step backwards).

Issue #87, opened 2015-03-21
[keyauth] Add auto expiration to keys


subnetmarco:
For example, to auto-expire in `2d` or `1y` since the creation date.

sinzone:
+1

kstigen:
+1

tunglam14:
+1

zihaozhu93:
+1

hizorro88:
+1

subutux:
Hi,
As a company looking into api gateways, this would much improve our use case for kong.

Is there any progress on this?

cesar-blog:
+1
Is there any progress on this?

azisaka:
+1

SuperMasterBlasterLaser:
+1

I know that on OAuth plugin there is an expiration, but it could be very usefull on other Authentications

wpjunior:
+1

thibaultcha:
An implementation of this has been proposed at #4239. It leverages the native TTL support of our new DAO implementation, which means that TTLs are specified in seconds with this first pass.

larryhu:
+1

bungle:
I am glad to close this PR now on behalf of @eskerda who fixed it by submitting a PR #4984. It took some four and half years to get this in, but we finally did, ;-). Perhaps it was worth to wait, like all the good stuff. It feels better if it just doesn't come given. Haha... well thanks again @eskerda! Closing this now.

maraswrona:
@bungle is this feature documented somewhere? I can see the `ttl` field present in Kong API response but don't know how to set it.

PR #3421, opened 2018-04-27
feat(key-auth) add expires_in to key-auth

zhouzhuojie:
### Summary
This pr adds the `expires_in` field to key-auth plugin. And its default value is `0`.
This pr also requires a migration, i.e. adding the new column `expires_in`. Let me know if it's ok to base on master or next branch.

### Full changelog
* Implement expires_in for key-auth plugin

### Issues resolved
Fix https://github.com/Kong/kong/issues/87

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -1,5 +1,6 @@
 local responses = require "kong.tools.responses"
 local constants = require "kong.constants"
+local timestamp = require "kong.tools.timestamp"
 local singletons = require "kong.singletons"
 local public_tools = require "kong.tools.public"
 local BasePlugin = require "kong.plugins.base_plugin"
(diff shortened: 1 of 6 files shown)


kong/plugins/key-auth/handler.lua
```diff
@@ -187,6 +188,15 @@ local function do_authentication(conf)
     return false, {status = 403, message = "Invalid authentication credentials"}
   end
 
+  -- check expires_in
```
zhouzhuojie: roughly copied from oauth2's plugin

zhouzhuojie:
@thibaultcha @thefosk Can I get some help on the reviews? thanks

kong/plugins/key-auth/migrations/postgres.lua
```diff
@@ -44,4 +44,13 @@ return {
     end,
     down = function(_, _, dao) end  -- not implemented
   },
+  {
+    name = "2018-04-26_key-auth_expires_in",
+    up = [[
+      ALTER TABLE keyauth_credentials ADD COLUMN expires_in int;
```
klamontagne: Should the default be set in the schema? (`SET DEFAULT 0`)
Should negative values be accepted? (`CHECK (expires_in > 0)`)

nijikokun: Setting a default value would cause a deviation in consistency between Postgres and Cassandra since Cassandra does not support default values on columns.

hishamhm:
We're going to be adding auto-expiration with #4984 so this one can probably be closed once that one is merged.

zhouzhuojie:
Close it with #4984

PR #3427, opened 2018-05-01
enabling lambda functions to return custom status, body, and headers

aloisbarreras:

### Summary

Currently, the AWS Lambda plugin will only return a 200, 202, or 204 HTTP status response code on successful requests. This adds the ability for Lambda functions to define custom response status codes by returning a specific object. For example returning the following from a Lambda function will set custom values on the Kong response.

```
{
    statusCode = "201",
    body = {
      key1 = "some_value_post1",
      key2 = "some_value_post2",
      key3 = "some_value_post3"
    },
    headers = {
      ["X-Custom-Header"] = "Hello world!"
    }
}
```

Returning this from your Lambda function would set that status to 201, the response body to `body` and add the custom headers in `headers`.

This same functionality is [already present with API Gateway](https://stackoverflow.com/a/39936210/5419691), and it would be great to replicate it here. It would make switching from API Gateway to Kong seamless :).

### Full changelog

* Update plugin to look for a statusCode, body, and headers in the Lambda response and apply them to the Kong response.
* Added test.

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -12,6 +12,7 @@ local public_utils = require "kong.tools.public"
 
 
 local tostring             = tostring
+local tonumber             = tonumber
 local pairs                = pairs
 local type                 = type
 local fmt                  = string.format
@@ -37,6 +38,53 @@ local server_header = meta._SERVER_TOKENS
 
 local AWS_PORT = 443
 
+
+--[[
+  Response format should be
+  {
+      "statusCode": httpStatusCode,
+      "headers": { "headerName": "headerValue", ... },
+      "body": "..."
+  }
+--]]
+local function validate_custom_response(response)
+  if type(response.statusCode) ~= "number" then
+    return nil, "statusCode must be a number"
+  end
+
+  if response.headers ~= nil and type(response.headers) ~= "table" then
+    return nil, "headers must be a table"
+  end
+
+  if response.body ~= nil and type(response.body) ~= "string" then
+    return nil, "body must be a string"
+  end
+
+  return true
+end
+
+
+local function extract_proxy_response(content)
+  local serialized_content, err = cjson.decode(content)
+  if not serialized_content then
+    return nil, err
+  end
+
+  local ok, err = validate_custom_response(serialized_content)
+  if not ok then
+    return nil, err
+  end
+
+  local headers = serialized_content.headers or {}
+  local body = serialized_content.body or ""
+  headers["Content-Length"] = #body
+
+  return {
+    status_code = tonumber(serialized_content.statusCode),
+    body = body,
+    headers = headers,
+  }
+end
 
 local function send(status, content, headers)
   ngx.status = status
@@ -187,13 +235,27 @@ function AWSLambdaHandler:access(conf)
   end
 
   local status
-  if conf.unhandled_status
-     and headers["X-Amz-Function-Error"] == "Unhandled"
-  then
-    status = conf.unhandled_status
 
-  else
-    status = res.status
+  if conf.is_proxy_integration then
+    local proxy_response, err = extract_proxy_response(content)
+    if not proxy_response then
+      return responses.send_HTTP_BAD_GATEWAY(err)
+    end
+
+    headers = utils.table_merge(headers, proxy_response.headers)
+    content = proxy_response.body
+    status = proxy_response.status_code
+  end
+
+  if not status then
+    if conf.unhandled_status
+      and headers["X-Amz-Function-Error"] == "Unhandled"
+    then
+      status = conf.unhandled_status
+
+    else
+      status = res.status
+    end
   end
 
   local ctx = ngx.ctx
--- a/kong/tools/responses.lua
+++ b/kong/tools/responses.lua
@@ -59,6 +59,7 @@ local _M = {
     HTTP_CONFLICT = 409,
     HTTP_UNSUPPORTED_MEDIA_TYPE = 415,
     HTTP_INTERNAL_SERVER_ERROR = 500,
+    HTTP_BAD_GATEWAY = 502,
     HTTP_SERVICE_UNAVAILABLE = 503,
   }
 }
@@ -73,6 +74,7 @@ local _M = {
 -- @field status_codes.HTTP_INTERNAL_SERVER_ERROR Always "Internal Server Error"
 -- @field status_codes.HTTP_METHOD_NOT_ALLOWED Always "Method not allowed"
 -- @field status_codes.HTTP_SERVICE_UNAVAILABLE Default: "Service unavailable"
+-- @field status_codes.HTTP_BAD_GATEWAY Always: "Bad Gateway"
 local response_default_content = {
   [_M.status_codes.HTTP_UNAUTHORIZED] = function(content)
     return content or "Unauthorized"
@@ -91,6 +93,9 @@ local response_default_content = {
   end,
   [_M.status_codes.HTTP_SERVICE_UNAVAILABLE] = function(content)
     return content or "Service unavailable"
+  end,
+  [_M.status_codes.HTTP_BAD_GATEWAY] = function(content)
+    return "Bad Gateway"
   end,
 }
 
@@ -116,7 +121,7 @@ local function send_response(status_code)
       coroutine.yield()
     end
 
-    if status_code == _M.status_codes.HTTP_INTERNAL_SERVER_ERROR then
+    if status_code == _M.status_codes.HTTP_INTERNAL_SERVER_ERROR or status_code == _M.status_codes.HTTP_BAD_GATEWAY then
       if content then
         ngx.log(ngx.ERR, tostring(content))
       end
--- a/spec/03-plugins/23-aws-lambda/01-access_spec.lua
+++ b/spec/03-plugins/23-aws-lambda/01-access_spec.lua
@@ -1,3 +1,4 @@
+local cjson = require "cjson"
 local helpers = require "spec.helpers"
 local meta    = require "kong.meta"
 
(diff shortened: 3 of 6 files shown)


aloisbarreras:
Hey guys! Any chance on this making the next release?

gaffney:
Thanks for the sweet contribution! Works for me on my now super-patched Kong Lambda plugin.

Python syntax of the above:
```python
return {
        'statusCode': 201,
        'body': {
          'key1': 'some_value_post1',
          'key2': 'some_value_post2',
          'key3': 'some_value_post3'
        },
        'headers': {
            'X-Custom-Header': 'Hello world!'
        }
    }
```

Curistofa:
AWS plugin throws an error if the lambda doesn't return a response body 
`2018/07/09 17:02:27 [error] 6295#0: *20927 [lua] public.lua:64: could not read request body (ngx.req.get_body_data() returned nil), client:, server: kong, request: "GET /test-api/testcustomers/v1/customers HTTP/1.1", host: ""
2018/07/09 17:02:27 [error] 6295#0: *20927 lua coroutine: runtime error: .../local/share/lua/5.1/kong/plugins/aws-lambda/handler.lua:155: attempt to index local 'serialized_body' (a userdata value)
stack traceback:
coroutine 0:
.../local/share/lua/5.1/kong/plugins/aws-lambda/handler.lua: in function <.../local/share/lua/5.1/kong/plugins/aws-lambda/handler.lua:34>
coroutine 1:
[C]: in function 'resume'
coroutine.wrap:21: in function <coroutine.wrap:21>`

aloisbarreras:
hey @thibaultcha this one missed the last release. I've rebased it, so I'd love some feedback and to get this merged in!

@Curistofa it should no longer throw any errors if Lambda doesn't return a body.

aloisbarreras:
FYI: I'm designing this to be in line with the AWS API Gateway functionality. For a custom response to work with API Gateway, the requirements are that the function must return a JSON object that looks like:

- statusCode (required)
- The 'body' property  must be a JSON string (optional)
- The 'headers' property is for custom response headers in addition to standard ones (optional)

aloisbarreras:
Not sure why the tests are failing. All the tests for the lambda plugin pass, but it looks like a test with the udp plugin failed as well as some cassandra ring health checks. Do those sometimes spontaneously fail @thibaultcha @bungle ?

thibaultcha:
Some tests eventually fail yes, that happens. Btw, I am away for a couple of weeks. But @bungle can help you.

aloisbarreras:
@bungle any feedback on this?

aloisbarreras:
any thoughts?

robjr:
Nice job! That's exactly what I was looking for.

thibaultcha:
@aloisbarreras We would also ask of you that you provide documentation for this change, in the https://github.com/Kong/docs.konghq.com repository - thank you!

aloisbarreras:
Okay @thibaultcha  I've addressed your feedback, and I think everything looks good.

_However_, I was just reading more about how AWS API Gateway works with Lambda and I didn't realize that you have to pick `Lambda Proxy Integration` when setting up the proxy route. It does not automatically look for this custom response by default.

That begs the question, should I add an option to the plugin configuration that enables this? I'd love some feedback from you and other community members on what they would like to see.

If we do go with that option, then the required format of the Lambda response is described [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format).

There's an `isBase64Encoded` encoded option I'll add as well once we decide if we want to add an option to the plugin config.

thibaultcha:
@aloisbarreras Thanks for verifying. I do feel like this should be an option considering the opinionated behaviour of the plugin on such responses, but I never used it myself.

Yes, we should gather feedback from the rest of the community here. Considering the attention this PR has received so far, it shouldn't be too hard to gather a few opinions I hope :) Please chime in if you are waiting for this PR!

kong/plugins/aws-lambda/handler.lua
```diff
@@ -9,6 +9,8 @@ local meta = require "kong.meta"
 local http = require "resty.http"
 local cjson = require "cjson.safe"
 local public_utils = require "kong.tools.public"
+local aws_lambda_proxy_schema = require "kong.plugins.aws-lambda.proxy_response_schema"
+local schemas = require "kong.dao.schemas_validation"
```
thibaultcha: Please do not use this module, as it is deprecated in favour of the new schema module written for the new DAO.

Overall, I question the dependency on either schema modules to perform such validations in this plugin (or any other plugin). The schemas (and their validation functions) are not meant for public consumption since there is no guarantee that their API or behaviour won't break in the future. They are meant to declaratively specify DAO entities only, and I fear that using them in plugins breaks the single responsibility principle.

aloisbarreras: Okay I'll change that up. It just seemed like the easiest way to do some validation.

Have any recommendations for how to validate the response format? I can roll my own function with some type() checks and if statements if that works.

kong/plugins/aws-lambda/handler.lua
```diff
@@ -37,6 +37,49 @@ local server_header = meta._SERVER_TOKENS
...
+
+  local headers = serialized_content.headers or {}
+  local body = serialized_content.body or ""
+  headers["Content-Length"] = #body
+
+  return {
+    status_code = tonumber(serialized_content.statusCode),
+    body = body,
+    headers = headers,
+  }, nil
```
thibaultcha: ditto, no need for the second `nil` return value here

spec/03-plugins/23-aws-lambda/01-access_spec.lua
```diff
@@ -1,3 +1,4 @@
+local cjson = require "cjson.safe"
```
thibaultcha: Better not using the safe module here, and instead:
```lua
require "cjson"
```

Because using the safe module requires to do handle errors manually, which isn't done in the tests this is used in.

RomainDeSaJardim:
According to [Lambda proxy integration](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format), the body is not specified to be optionnal. However headers is clearly specified to be optionnal.
_In the output, headers can be unspecified if no extra response headers are to be returned._

I'm working on the documentation for this PR

thibaultcha:
@aloisbarreras Hey there, just a nudge: will you be able to address the review comments?

aloisbarreras:
hey @thibaultcha . Yes! Sorry work has been crazy. I will finish this by EOW.

kong/plugins/aws-lambda/handler.lua
```diff
@@ -38,6 +38,51 @@ local server_header = meta._SERVER_TOKENS
...
+  if not ok then
+    return nil, err
+  end
+
+  local headers = serialized_content.headers or {}
+  local body = serialized_content.body or ""
+  headers["Content-Length"] = #body
+
+  return {
+    status_code = tonumber(serialized_content.statusCode),
```
thibaultcha: Please cache the `tonumber` global lookup at the beginning of this file, just like other globals that are accessed :)

spec/03-plugins/23-aws-lambda/01-access_spec.lua
```diff
@@ -527,5 +553,93 @@ for _, strategy in helpers.each_strategy() do
...
+        method  = "POST",
+        path    = "/post",
+        headers = {
+          ["Host"]         = "lambda11.com",
+          ["Content-Type"] = "application/json"
+        },
+        body = {
+          statusCode = 201,
+          body = 1234,
+        }
```
thibaultcha: We might be missing a test case when the actual response from the Lambda is _not_ a JSON payload, as pointed by my earlier comment). Here, only the value of `body` is not JSON.

kong/plugins/aws-lambda/handler.lua
```diff
@@ -38,6 +38,51 @@ local server_header = meta._SERVER_TOKENS
...
+      "body": "..."
+  }
+--]]
+local function validate_custom_response(response)
+  if type(response.statusCode) ~= "number" then
+    return nil, "statusCode must be a number"
+  end
+
+  local headers = response.headers
+  local body = response.body
```
thibaultcha: There is no point in de-referencing those values it seems. The below code even indexes `response.body` still. It seems concise enough.

kong/tools/responses.lua
```diff
@@ -59,6 +59,7 @@ local _M = {
     HTTP_CONFLICT = 409,
     HTTP_UNSUPPORTED_MEDIA_TYPE = 415,
     HTTP_INTERNAL_SERVER_ERROR = 500,
+    HTTP_BAD_GATEWAY = 502,
```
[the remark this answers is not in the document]
aloisbarreras: Yeah that makes sense. I updated the logic and test, let me know if you have any other feedback.

thibaultcha:
Merged to master! Thank you for your hard work on this @aloisbarreras! :tada: :+1: 

I made a few tweaks locally, added a few test cases, extended the commit message, and did some cleanup/squashing. Kong 1.0 rc2 will ship with this new Lambda Proxy Integration (between now and the next couple of weeks). Thanks to everybody who helped and to those who expressed their interest in this as well (by emojis or even in person).

The docs PR opened by @romdsj can be seen [here](https://github.com/Kong/docs.konghq.com/pull/836), and will be published when Kong 1.0 is released. We will be refining it, and until then, the PR can be used as documentation for this yet to be released feature :)

On the "NYI" (not yet implemented) list for this feature, I noted the lack of support for `isBase64Encoded` in the commit message. We would welcome a contribution for this feature as well!

PR #3442, opened 2018-05-08
feat(aws-lambda) add forward_request_uri_path_params config option

aloisbarreras:
### Summary

When defining a Kong Route with a regex with capturing groups such as:

```/version/(?<version>\d+)/users/(?<user>\S+)```

The matched params will be available inside the plugin on ```local router_matches = ngx.ctx.router_matches```.

This adds the config option to forward those matched path params to the Lambda event body under ```request_uri_path_params```.

Therefore, the above route regex will produce a Lambda event body of:

```
{
  request_uri_path_params = {
    user = "batman",
    ["1"] = "12",
    ["2"] = "batman",
    version = "12",
    ["0"] = "/versions/12/users/batman"
  },
}
```

### Full changelog

* Added forward_request_uri_path_params config option and passing params to upstream body
* Add related tests.

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -38,6 +38,7 @@ function AWSLambdaHandler:access(conf)
 
   if conf.forward_request_body or conf.forward_request_headers
     or conf.forward_request_method or conf.forward_request_uri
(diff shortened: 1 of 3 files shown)


hbagdi:

PR #3450, opened 2018-05-10
fix(api) correctly handle sni in PUT /certificates/foo.com

kikito:
Previously this method did create an extra certificate, but not the
accompanying SNI included on the URL. See the spec for details.


thibaultcha:
Merged with `hotfix(admin)` as per commit message conventions!

PR #2735, opened 2017-07-26
feat(rate-limit) Add the `X-RateLimit-Reset` header

grahamar:
NOTE: Please read the CONTRIBUTING.md guidelines before submitting your patch,
and ensure you followed them all:

https://github.com/Mashape/kong/blob/master/CONTRIBUTING.md#contributing

### Summary

While there's no standards on rate limiting headers, that I can find, a convention does exist
across a number of popular APIs (e.g. Github, Twitter, Vimeo, Reddit) using variations of the following.

- X-RateLimit-Limit – The limit that you cannot surpass in a given amount of time
- X-RateLimit-Remaining – The number of calls you have available until a given reset time stamp, or calculated given some sort of sliding time window.
- X-RateLimit-Reset – The timestamp in UTC formatted to HTTP spec per RFC 1123 for when the limits will be reset.

The `Retry-After` header is mentioned in the IETF RFCs (e.g. [RFC-7231](https://tools.ietf.org/html/rfc7231#section-7.1.3) and serves
a very similar purpose to the `X-RateLimit-Reset` header, the only functionality that's missing from the current set of headers.

### Full changelog

* Add `X-RateLimit-Reset` header
* Add test

### Issues resolved

Fix #233

PR #3451, opened 2018-05-10
Fixes #233: Add generic X-RateLimit-* and X-RateLimit-Reset headers too.

### LINKED ISSUE #233: [ratelimiting] X-RateLimit-Reset

| Header Name | Description |
| --- | --- |

ioggstream:
### Summary

In case of 429, returns `Retry-After` as per https://tools.ietf.org/html/rfc6585#section-4

Return the lowest limit header as X-RateLimit-Remaining and
the associate X-RateLimit-Reset.

### Full changelog

* Add `Retry-After` when clients are throttled out
* Add X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset using the lower values between the configured ones

### Issues resolved

Fix #233

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -15,6 +15,9 @@ local timer_at = ngx.timer.at
 local EMPTY = {}
 local RATELIMIT_LIMIT = "X-RateLimit-Limit"
 local RATELIMIT_REMAINING = "X-RateLimit-Remaining"
+local RATELIMIT_RESET = "X-RateLimit-Reset"
+local RETRY_AFTER = "Retry-After"
+local EXPIRATIONS = policies["EXPIRATIONS"]
 
 local RateLimitingHandler = BasePlugin:extend()
 
@@ -115,6 +118,8 @@ function RateLimitingHandler:access(conf)
     -- Adding headers
     if not conf.hide_client_headers then
       local headers = {}
+      -- Find the closer limit.
+      local lowest = nil
       for k, v in pairs(usage) do
         if stop == nil or stop == k then
           v.remaining = v.remaining - 1
@@ -122,13 +127,28 @@ function RateLimitingHandler:access(conf)
 
         headers[RATELIMIT_LIMIT .. "-" .. k] = v.limit
         headers[RATELIMIT_REMAINING .. "-" .. k] = max(0, v.remaining)
+
+        if (lowest == nil
+            or usage[lowest].remaining > v.remaining
+            or (usage[lowest].remaining == v.remaining and EXPIRATIONS[k] > EXPIRATIONS[lowest])
+           ) then
+          lowest = k
+        end
+
       end
+
+      -- Set the normalized ratelimit values.
+      headers[RATELIMIT_LIMIT] = headers[RATELIMIT_LIMIT .. "-" .. lowest]
+      headers[RATELIMIT_REMAINING] = headers[RATELIMIT_REMAINING .. "-" .. lowest]
+      -- Set the expected reset time.
+      headers[RATELIMIT_RESET] = EXPIRATIONS[lowest] - (current_timestamp/1000 % EXPIRATIONS[lowest])
 
       kong.ctx.plugin.headers = headers
     end
 
     -- If limit is exceeded, terminate the request
     if stop then
+      kong.ctx.plugin.headers[RETRY_AFTER] = kong.ctx.plugin.headers[RATELIMIT_RESET]
       return kong.response.exit(429, { message = "API rate limit exceeded" })
     end
   end
--- a/kong/plugins/rate-limiting/policies/init.lua
+++ b/kong/plugins/rate-limiting/policies/init.lua
@@ -59,6 +59,7 @@ local EXPIRATIONS = {
 
 
 return {
+  ["EXPIRATIONS"] = EXPIRATIONS,
   ["local"] = {
     increment = function(conf, limits, identifier, current_timestamp, value)
       local periods = timestamp.get_timestamps(current_timestamp)
--- a/spec/03-plugins/23-rate-limiting/04-access_spec.lua
+++ b/spec/03-plugins/23-rate-limiting/04-access_spec.lua
@@ -257,6 +257,9 @@ for _, strategy in helpers.each_strategy() do
         wait(3)
       end)
 
+      -- print("limit", res.headers["x-ratelimit-limit"])
+      -- print("remaining", res.headers["x-ratelimit-remaining"])
+      -- print("reset", res.headers["x-ratelimit-reset"])
       describe("Without authentication (IP address)", function()
         it("blocks if exceeding limit", function()
           for i = 1, 6 do
@@ -266,6 +269,10 @@ for _, strategy in helpers.each_strategy() do
 
             assert.are.same(6, tonumber(res.headers["x-ratelimit-limit-minute"]))
             assert.are.same(6 - i, tonumber(res.headers["x-ratelimit-remaining-minute"]))
+            -- Common headers
+            assert.are.same(6, tonumber(res.headers["x-ratelimit-limit"]))
+            assert.are.same(6 - i, tonumber(res.headers["x-ratelimit-remaining"]))
+            -- assert.is_true(tonumber(res.headers["x-ratelimit-reset"]) <= 60, tonumber(res.headers["x-ratelimit-reset"])) -- X 
           end
 
           -- Additonal request, while limit is 6/minute
@@ -276,6 +283,9 @@ for _, strategy in helpers.each_strategy() do
           local body = assert.res_status(429, res)
           local json = cjson.decode(body)
           assert.same({ message = "API rate limit exceeded" }, json)
+          -- retry after 60 - 6 seconds 
+          print("reset", res.headers["retry-after"])
+          -- assert.is_true(tonumber(res.headers["retry-after"]) <= 60)
         end)
 
         it("counts against the same service register from different routes", function()
@@ -286,6 +296,10 @@ for _, strategy in helpers.each_strategy() do
 
             assert.are.same(6, tonumber(res.headers["x-ratelimit-limit-minute"]))
             assert.are.same(6 - i, tonumber(res.headers["x-ratelimit-remaining-minute"]))
+            -- Common headers
+            assert.are.same(6, tonumber(res.headers["x-ratelimit-limit"]))
+            assert.are.same(6 - i, tonumber(res.headers["x-ratelimit-remaining"]))
+            -- assert.is_true(tonumber(res.headers["x-ratelimit-reset"]) < 60)
           end
 
           for i = 4, 6 do
@@ -295,6 +309,10 @@ for _, strategy in helpers.each_strategy() do
 
             assert.are.same(6, tonumber(res.headers["x-ratelimit-limit-minute"]))
             assert.are.same(6 - i, tonumber(res.headers["x-ratelimit-remaining-minute"]))
+            -- Common headers
+            assert.are.same(6, tonumber(res.headers["x-ratelimit-limit"]))
+            assert.are.same(6 - i, tonumber(res.headers["x-ratelimit-remaining"]))
+            -- assert.is_true(tonumber(res.headers["x-ratelimit-reset"]) < 60)
           end
 
           -- Additonal request, while limit is 6/minute
@@ -304,6 +322,7 @@ for _, strategy in helpers.each_strategy() do
 
           local json = cjson.decode(body)
           assert.same({ message = "API rate limit exceeded" }, json)
+          -- assert.is_true(tonumber(res.headers["retry-after"]) < 60)
         end)
 
         it("handles multiple limits", function()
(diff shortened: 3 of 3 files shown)


kong/plugins/rate-limiting/handler.lua
```diff
@@ -104,6 +111,18 @@ function RateLimitingHandler:access(conf)
...
         ngx.header[RATELIMIT_REMAINING .. "-" .. k] = math.max(0, (stop == nil or stop == k) and v.remaining - 1 or v.remaining) -- -increment_value for this current request
       end
+
+      -- Find the closer limit.
+      local closer_limit = 0
+      for k, v in pairs(usage) do
+        if (closer_limit == 0 or closer_limit > v.remaining) then
+          closer_limit = v.remaining
+          ngx.header[RATELIMIT_LIMIT] = v.remaining
+          ngx.header[RATELIMIT_REMAINING] = v.limit
```
thibaultcha: Those headers duplicate what is already available (and more detailed) to the client. If the client wishes to compute what is the closest limit, they can do so themselves. Those new headers are unnecessary here.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -104,6 +111,18 @@ function RateLimitingHandler:access(conf)
...
+          -- Add a little jitter to the expected reset time.
+          ngx.header[RATELIMIT_RESET] = math.random(0,10) + SECONDS_IN_PERIOD[k] - (current_timestamp/1000 % SECONDS_IN_PERIOD[k])
```
thibaultcha: A few notes here:
1. What is the reasoning behind the "jitter" inserted here? It seems unnecessary. The fixed window approach followed by this plugin should already guarantee that there is a given second at which the counters will be reset.
2. It is inefficient to call `ngx.header` within a loop. Instead, better to assign this value to a local, and assign `ngx.header` to this local _after_ the loop.
3. `SECONDS_IN_PERIODS` is missing `day`, `month`, and `year` values. This code will try to do arithmetic on a `nil` value and throw interpreter errors.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -104,6 +111,18 @@ function RateLimitingHandler:access(conf)
         ngx.header[RATELIMIT_LIMIT .. "-" .. k] = v.limit
         ngx.header[RATELIMIT_REMAINING .. "-" .. k] = math.max(0, (stop == nil or stop == k) and v.remaining - 1 or v.remaining) -- -increment_value for this current request
       end
+
+      -- Find the closer limit.
+      local closer_limit = 0
+      for k, v in pairs(usage) do
```
thibaultcha: This loops is unnecessary, we could simply add its body to the above one.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -12,6 +12,13 @@ local ngx_timer_at = ngx.timer.at
 
 local RATELIMIT_LIMIT = "X-RateLimit-Limit"
 local RATELIMIT_REMAINING = "X-RateLimit-Remaining"
+local RATELIMIT_RESET = "X-RateLimit-Reset"
+
+local SECONDS_IN_PERIOD = {
+  second = 1,
+  minute = 60,
+  hour = 3600
+}
```
thibaultcha: Maybe better to expose the [EXPIRATIONS constant](https://github.com/ioggstream/kong/blob/89afad90c6bb1d9ed2de666ceadeb438a543cadb/kong/plugins/rate-limiting/policies/init.lua#L54-L61) instead of duplicating it (it is also incomplete, see below comment).

kong/plugins/rate-limiting/handler.lua
```diff
@@ -104,6 +111,18 @@ function RateLimitingHandler:access(conf)
         ngx.header[RATELIMIT_LIMIT .. "-" .. k] = v.limit
         ngx.header[RATELIMIT_REMAINING .. "-" .. k] = math.max(0, (stop == nil or stop == k) and v.remaining - 1 or v.remaining) -- -increment_value for this current request
       end
+
+      -- Find the closer limit.
+      local closer_limit = 0
+      for k, v in pairs(usage) do
+        if (closer_limit == 0 or closer_limit > v.remaining) then
```
thibaultcha: This condition seems flawed to me.

1. It will find the "smallest limit", and not the "closest to quota" one.
2. If `v.remaining` is `0` in the first iteration (and any subsequent ones), then we enter the branch when in fact, the value is the same. While 1. is a bigger concern, this points just highlights inefficiency in the implementation of 1.

It is my belief that your intent is to show this header only for the closest limit anyway, so the condition to retrieve this limit needs to be reworked.

kong/plugins/rate-limiting/handler.lua
thibaultcha: My biggest concern with this PR relates to [a comment](https://github.com/Kong/kong/pull/2735/files#r134082010) I made last year in a similar PR (see the second part of the comment).

I believe we should show two headers to clients:
1. A `X-Rate-Limit-Retry`, which should only appear in `429` responses, and tells the client when will they be able to make a request again. This header should be inserted after having found the shortest limit that triggered a rejection (e.g. `second` over `minute`).
2. A `X-Rate-Limit-Reset`, which could be present in all responses sent to the client, and indicate when will _all_ of their counters be reset. In this case, this header would need to find the _largest_ limit configured (e.g. `minute` over `second`).

This way, a client has a way of knowing:
1. When is the soonest they can retry a request once they have been blocked
2. When can they expect their quota to be fully reset (i.e. when does the current time window ends?)

Here, is seems like you are trying to implement 1., but the chosen name and behavior contradicts those goals. Thoughts?

kong/plugins/rate-limiting/handler.lua
ioggstream: I created a different loop just to isolate better the patch. It's fine to integrate that in the loop ;)

kong/plugins/rate-limiting/handler.lua
ioggstream: Sure! Will do!

kong/plugins/rate-limiting/handler.lua
ioggstream: Ok, I just didn't want to modify the actual code. It's fine to modify the loop.

kong/plugins/rate-limiting/handler.lua
ioggstream: - agree on 2. initializing  `closer_limit = -1` is better
- I want to just publish the lowest limit.

kong/plugins/rate-limiting/handler.lua
ioggstream: While "labeled" ratelimits are very useful, most of the clients will just honor `X-RateLimit-Limit|Remaining` with no suffixes. 

My goal is to provide the "vanilla" headers.

Delegating

ioggstream:
@thibaultcha Hi! And thank you very much for your time your comments.

My goal is:

  - expose just 3 headers with the first limit that will fire on every call
     `X-RateLimit-Limit|Remaining|Reset`
      - the reset should be expressed in seconds
      - my clients are free to  enjoy "labeled" headers (eg. X-RateLimit-Limit-minute, X-RateLimit-Limit-hour, ..)
      - my clients must not be forced to iterate between different "labeled" headers
  - expose one more `Retry-After` + 429

kong/plugins/rate-limiting/handler.lua
thibaultcha: * Yes to `Retry-After`, let's use this name if it is part of the RFC. I believe, as pointed out in my initial comment, that it should contain the value indicating to the client when, at the soonest, will they be able to make another request.

And as per the [RFC](https://tools.ietf.org/html/rfc6585#section-4), this seems check out:

>  The response representations SHOULD include details explaining the
   condition, and MAY include a Retry-After header indicating how long
   to wait before making a new request.

* `X-RateLimit-Reset`: imho, this should indicate to the client when _all_ of their counters will be reset (and thus provides the information `Retry-After` does not already).

* `X-RateLimit-Limit`: I still do not see the point of this header, and find it confusing because it fluctuates between time periods (minutes, seconds, etc...) depending on which one is the closest to `0`. I have trouble understanding how a client could make efficient use of such a header, since it cannot reasonably come up with an execution plan for future requests, because the time window of the counter is unspecified. Would you mind clarifying?
And again, my point about duplication of information still stands (this value is already present in the headers and can be deducted by clients).

thibaultcha: Ah, I think it just struck me that you wish `X-RateLimit-Limit` to be interpreted _alongside_ the `X-RateLimit-Reset` header, thus indicating to the client which time window is currently the most restrictive to them, and when such a time window will be reset... I am still not sure if this is free of confusion for instances of this plugin configured with different time windows... And my point about this information already being available to the clients still stands...

ioggstream: @thibaultcha The rationale of adding the couple (Limit, Reset) to the "labeled" headers is here:

  - https://github.com/Kong/kong/pull/3451#issuecomment-388553141

Thx++, R

kong/plugins/rate-limiting/handler.lua
ioggstream: done in ff678caf5

kong/plugins/rate-limiting/handler.lua
ioggstream: done in ff678caf5

kong/plugins/rate-limiting/handler.lua
ioggstream: done in ff678caf5. Feedback welcome ;)

kong/plugins/rate-limiting/handler.lua
ioggstream: The logic is the same, but honoring the largest period ff678caf5

HTH. Thanks again for all your time, R:

kong/plugins/rate-limiting/handler.lua
```diff
@@ -12,6 +12,17 @@ local ngx_timer_at = ngx.timer.at
...
+local RETRY_AFTER = "Retry-After"
+
+local EXPIRATIONS = {
+  second = 1,
+  minute = 60,
+  hour = 3600,
+  day = 86400,
+  month = 2592000,
+  year = 31536000,
+}
```
thibaultcha: Why not exposing the table from the `policies` module as previously discussed?

kong/plugins/rate-limiting/handler.lua
```diff
@@ -14,15 +14,7 @@ local RATELIMIT_LIMIT = "X-RateLimit-Limit"
...
-
-local EXPIRATIONS = {
-  second = 1,
-  minute = 60,
-  hour = 3600,
-  day = 86400,
-  month = 2592000,
-  year = 31536000,
-}
+local EXPIRATIONS = policies["EXPIRATIONS"]
```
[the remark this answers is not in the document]
thibaultcha: @ioggstream Yes, that is fine, thanks.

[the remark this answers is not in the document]
ioggstream: @thibaultcha thanks for the review. Let me know if I can help the merge in some way!

ioggstream:
Hi @thibaultcha, any news?

thibaultcha:
@ioggstream Hi,

Sorry, nothing for now. We are busy preparing the 0.14 release. Unfortunately this PR is still lacking tests, for starters (as previously mentioned).

ioggstream:
@thibaultcha Good work for the 0.14!

Should tests be implemented in `04-access_spec` or in a new file?

Do you think those tests will suffice? 

  - https://github.com/Kong/kong/blob/master/spec/03-plugins/24-rate-limiting/04-access_spec.lua#L262
    expect: X-RateLimit-Limit,Remaining equals to the previous headers
    expect: X-RateLimit-Reset in seconds
  - https://github.com/Kong/kong/blob/master/spec/03-plugins/24-rate-limiting/04-access_spec.lua#L271 
    expect: Retry-After when 429
  - https://github.com/Kong/kong/blob/master/spec/03-plugins/24-rate-limiting/04-access_spec.lua#L323
    expect: X-RateLimit-* referencing the closer limit
    expect: X-RateLimit-Reset = min(x-ratelimit-reset-*)
  - https://github.com/Kong/kong/blob/master/spec/03-plugins/24-rate-limiting/04-access_spec.lua#L323
    expect: Retry-After when 429 with min(x-ratelimit-reset-*)

ioggstream:
@thibaultcha Hi! Can you please tell me if the tests suggested in https://github.com/Kong/kong/issues/233#issuecomment-397551093  are fine?

ioggstream:
Hi @thibaultcha I added some tests but:

  -  all `04-access_spec`  are marked as `#flaky` and are not executed by the CI

Is that the expected behavior?

ioggstream:
Hi @thibaultcha. I'm working on this I-D on rate-limiting. Let me know if this is of some interest for you.

https://ioggstream.github.io/draft-polli-ratelimit-headers/draft-polli-ratelimit-headers.html

bungle:
the #5335 is now merged, thus closing this.

PR #3457, opened 2018-05-13
ScyllaDB compatibility: explicitly drop index before dropping table

PAStheLoD:
### Summary

ScyllaDB compatibility patch

### Full changelog

Added an explicit `DROP INDEX`, because otherwise Scylla refuses to drop `nodes` with the following error:

```
migrating core for keyspace kong
Error: [cassandra error] Error during migration 2017-05-19-173100_remove_nodes_table: [Invalid] Cannot drop table when materialized views still depend on it (kong.{nodes_cluster_listening_address_idx_index})
```

### DIFF
--- a/kong/dao/migrations/cassandra.lua
+++ b/kong/dao/migrations/cassandra.lua
@@ -487,7 +487,7 @@ return {
   {
     name = "2017-05-19-173100_remove_nodes_table",
     up = [[
-      DROP INDEX IF EXISTS nodes_cluster_listening_address_idx_index;
+      DROP INDEX IF EXISTS nodes_cluster_listening_address_idx;
       DROP TABLE nodes;
     ]],
   },
--- a/kong/dao/migrations/cassandra.lua
+++ b/kong/dao/migrations/cassandra.lua
@@ -487,7 +487,7 @@ return {
   {
     name = "2017-05-19-173100_remove_nodes_table",
     up = [[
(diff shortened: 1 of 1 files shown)


thibaultcha:
SGTM :)

@kikito In the recent Cassandra migration helpers for new DAO entities (https://github.com/Kong/kong/blob/next/kong/dao/migrations/helpers.lua#L104), this might be something we'd want to add, otherwise dropped tables with indexes would cause the same error for ScyllaDB (which isn't an officially supported database, granted, but still cheap to fix)

kong/dao/migrations/cassandra.lua
```diff
@@ -487,6 +487,7 @@ return {
   {
     name = "2017-05-19-173100_remove_nodes_table",
     up = [[
+      DROP INDEX IF EXISTS nodes_cluster_listening_address_idx_index;
```
thibaultcha: In my local C* 3.9 cluster, this index is named  `nodes_cluster_listening_address_idx`. Oversight?

PAStheLoD: Thanks for catching this! Yes, I mixed up the name for the materialized view and for the actual index.

thibaultcha:
@PAStheLoD Thank you! If you wish to, you can redeem a [Kong Contributor T-shirt](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt) :)

PR #3461, opened 2018-05-16
Match route 'host' inclusive of port

james-callahan:
## Summary

Currently a route is matched based on the hostname. This PR changes it to match on a host+port pair, as sent in the `Host` header.

I think the current situation came about due to the confusing identifier of `host`: in some places it is used to mean "host name", while in others it is used to mean "form of a Host header". As ports are optional in host headers, it is easy to conflate the two.

## Open Questions

1.  When a HTTP request is sent to the nominal port (80 for http; 443 for https), the host header does not include the port.

      - <s>What route should be used if a client sends `Host: myhost.com:80`?</s>
          - <s>What about if it was a https request?</s>
      - <s>Does nginx strip or modify things in any way?</s>
      - If an admin adds a route with host `myhost.com:80`, when should it match?

2. Should X-Forwarded-Port be removed?

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -15,13 +15,9 @@ local type = type
 
 
 local function validate_host(host)
-  local res, err_or_port = utils.normalize_ip(host)
-  if type(err_or_port) == "string" and err_or_port ~= "invalid port number" then
+  local _, err_or_port = utils.normalize_ip(host)
+  if type(err_or_port) == "string" then
     return nil, "invalid value: " .. host
-  end
-
-  if err_or_port == "invalid port number" or type(res.port) == "number" then
-    return nil, "must not have a port"
   end
 
   return true
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -907,7 +907,7 @@ function _M.new(routes)
     end
   end
 
-  local function find_route(req_method, req_uri, req_host, ngx,
+  local function find_route(req_method, req_uri, req_host, req_scheme, ngx,
                             src_ip, src_port, dst_ip, dst_port, sni)
     if req_method and type(req_method) ~= "string" then
       error("method must be a string", 2)
@@ -917,6 +917,9 @@ function _M.new(routes)
     end
     if req_host and type(req_host) ~= "string" then
       error("host must be a string", 2)
+    end
+    if req_scheme and type(req_scheme) ~= "string" then
+      error("scheme must be a string", 2)
     end
     if src_ip and type(src_ip) ~= "string" then
       error("src_ip must be a string", 2)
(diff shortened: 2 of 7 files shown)


thibaultcha:
+1 on this change from my side.

james-callahan:
Rebased and fixed performance concern.

james-callahan:
Answering two of the open questions:

>   - What route should be used if a client sends Host: myhost.com:80?
>      -  What about if it was a https request?

Answer: trim off 80 or 443 based on req_scheme.

>   - Does nginx strip or modify things in any way?

Yes: trailing whitespace is stripped.
This is safe behaviour for us.

james-callahan:
Needs tests.

francois-maillard:
> Should X-Forwarded-Port be removed?

If not removed, it should be possible to overwrite it since in some cases setting the X-Forward-Port to the current listening port is wrong and leads to routing issues. For instance:
- when using an [OSI layer 4 load-balancer](https://discuss.konghq.com/t/x-forwarded-port-and-load-balancing/1588)
- when using of the [docker port mapping feature](https://docs.docker.com/compose/compose-file/#ports)

thibaultcha:
The new support for wildcards in hosts? Wildcard hosts have been supported for a very long time, can you elaborate?

james-callahan:
> The new support for wildcards in hosts? Wildcard hosts have been supported for a very long time, can you elaborate?

Ah my mistake, they aren't *that* new.
The problem is that the `validate_host_with_wildcards` type needs to be expanded for port support. Possibly with wildcards for port allowed too.

p0pr0ck5:
Is this still relevant?

thibaultcha:
Yes.

laurentverbruggen:
Would really like to see this implemented!

When deploying Kong as K8S ingress controller there is port mapping involved, resulting in an **incorrect X-Forwarded-Port**. This causes problems for certain applications that rely on that header to calculate correct redirect URLs. The location header contains the correct host but not the correct port (8000 instead of 80). 

Removing the header completely would solve my problem but I don't know if others might need the header. Overriding would also solve the problem and including it in the host header too. So whatever works for you guys, as long as something changes :)

subnetmarco:
@james-callahan @thibaultcha what is the status of this PR?

miles-:
@james-callahan @thibaultcha Any updates on this PR? I have the some use-case as @laurentverbruggen.

rveede:
Same here :)

hishamhm:
Given this PR is not currently in a mergeable state, and it looks like it would take a sort of rewrite to get it in sync with current `next`, we're closing it for now. @james-callahan if you feel like getting this up to speed, feel free to reopen it.

davinwang:
Same here

trollr:
Any update here? Need it cuz fusionauth does csrf protection and Im not able to set the correct x-forwarded-port header.

hishamhm:
@trollr Happy to report that this functionality is supported since Kong 1.5 and 2.0. It was implemented in PR #5102.

trollr:
Most of it. The X-forwarded-port is still there. That's the reason for my issues :(

PR #3037, opened 2017-11-16
Allow the api name to appear in tags instead of the metric name

kbacha:
### Summary

Data dog prefers to use tags to query and divide statistics up by.
By putting the api name in the key, we cannot create dashboards that
are keyed off the api name. This is accomplished via a new plugin
setting called `tag_api_name`. It is a boolean field that when set
to true will remove the api_name from the metric key and add it as
a tag. It will append this tag to whatever tags are defined by the
metric configuration.

### Full changelog

* Implemented schema changes
* Implemented test for api name inclusion in key set
* Implemented appended api name to key set
* Refactored plugin to help readability.

PR #3465, opened 2018-05-17
Alter datadog plugin to tag metric data per datadog best-practices

AndrewFarley:
### Summary

Currently all metrics for datadog are pushed to differently named metrics (eg: kong.api_name.latency, kong.api2_name.latency).  This is bad practice on Datadog and it makes it impossible to make a single dashboard in Datadog with variable options (eg: choosing which API) to change which API you'd like to analyze.  With the current implementation you have to manually create a dashboard for every API you ever make.  Datadog best-practices dictate that you use a static metric name and use tags to differentiate your data within' that metric.

### Full changelog

* Made all datadog metric names static
* Added proper tagging to datadog metrics instead of variable metric names
* Removed a now-obsolete metric (totals) since you can glean this from a single metric without specifying an app_name tag.

### Additional notes

* Because the way this plugin works is fundamentally changing (improving) any/all blogs and articles that point to the old way this works needs to be updated.  These are more external, but the one I used was: https://www.datadoghq.com/blog/monitor-kong-datadog/ .  I've submitted a handful of improvement requests to their articles lately... I'm sure another one or two wouldn't hurt.
* Admittedly, I'm not a Lua programmer.  Maintainers, please edit/change/modify as necessary.  I tried to fit within' the contribution and code style guidelines, but I could be way off in some ways.

### Relevant Links

[Datadog - Tagging](https://docs.datadoghq.com/getting_started/tagging/)
[Datadog - Using Tags](https://docs.datadoghq.com/getting_started/tagging/using_tags/)
[Datadog - Why to use tags](https://www.datadoghq.com/blog/the-power-of-tagged-metrics/)
[Datadog - Dashboard Template Variables](https://docs.datadoghq.com/graphing/faq/correlate-metrics-and-events-using-dashboard-template-variables/)

### DIFF
--- a/kong/plugins/datadog/handler.lua
+++ b/kong/plugins/datadog/handler.lua
@@ -31,26 +31,22 @@ local get_consumer_id = {
 
 local metrics = {
   status_count = function (api_name, message, metric_config, logger)
-    local fmt = string_format("%s.request.status", api_name,
-                       message.response.status)
-
-    logger:send_statsd(string_format("%s.%s", fmt, message.response.status),
+    logger:send_statsd("api.request.status",
                        1, logger.stat_types.counter,
-                       metric_config.sample_rate, metric_config.tags)
-
-    logger:send_statsd(string_format("%s.%s", fmt, "total"), 1,
-                       logger.stat_types.counter,
-                       metric_config.sample_rate, metric_config.tags)
+                       metric_config.sample_rate, 
+                       {
+                          "app:kong",
+                          "api_name:" .. api_name,
+                          "response_code:" .. message.response.status
+                       })
   end,
   unique_users = function (api_name, message, metric_config, logger)
     local get_consumer_id = get_consumer_id[metric_config.consumer_identifier]
     local consumer_id     = get_consumer_id(message.consumer)
 
     if consumer_id then
-      local stat = string_format("%s.user.uniques", api_name)
-
-      logger:send_statsd(stat, consumer_id, logger.stat_types.set,
-                         nil, metric_config.tags)
+      logger:send_statsd("api.user.uniques", consumer_id, logger.stat_types.set,
+                         nil, {"app:kong", "api_name:" .. api_name} )
     end
   end,
   request_per_user = function (api_name, message, metric_config, logger)
@@ -58,10 +54,13 @@ local metrics = {
     local consumer_id     = get_consumer_id(message.consumer)
 
     if consumer_id then
-      local stat = string_format("%s.user.%s.request.count", api_name, consumer_id)
-
-      logger:send_statsd(stat, 1, logger.stat_types.counter,
-                         metric_config.sample_rate, metric_config.tags)
+      logger:send_statsd("api.user.request.count", 1, logger.stat_types.counter,
+                         metric_config.sample_rate, 
+                         {
+                           "app:kong",
+                           "api_name:" .. api_name,
+                           "consumer_id:" .. consumer_id
+                         })
     end
   end,
   status_count_per_user = function (api_name, message, metric_config, logger)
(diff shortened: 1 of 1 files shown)


AndrewFarley:
I guess I (or someone) needs to go fix the tests to conform to the new standards.  I don't really have the cycles right now... so, maybe will get around to it later... hopefully someone else can chip in to help clean this up.

AndrewFarley:
Other quick last minute comments from utilizing this patch already in the wild... there is one more element I (or someone) can remove when they have time... the "count" metric is also redundant now, since similar to "total" you can simply SUM() the metric on datadog.  :P

kong/plugins/datadog/handler.lua
```diff
@@ -112,12 +109,12 @@ local function log(premature, conf, message)
...
-    latency          = name .. ".latency",
-    upstream_latency = name .. ".upstream_latency",
-    kong_latency     = name .. ".kong_latency",
-    request_count    = name .. ".request.count",
+    request_size     = "api.request.size",
+    response_size    = "api.response.size",
+    latency          = "api.latency",
+    upstream_latency = "api.upstream_latency",
+    kong_latency     = "api.kong_latency",
+    request_count    = "api.request.count",
```
thibaultcha: This 0.13.0, APIs are deprecated in favor of Services and Routes. Why not using a more generic name?

kong/plugins/datadog/handler.lua
```diff
@@ -146,7 +143,8 @@ local function log(premature, conf, message)
 
       logger:send_statsd(stat_name, stat_value,
                          logger.stat_types[metric_config.stat_type],
-                         metric_config.sample_rate, metric_config.tags)
+                         metric_config.sample_rate, 
+                         {"app:kong", "api_name:" .. name})
```
thibaultcha: ditto, `api_name` is incorrect in the context of Services/Routes

kong/plugins/datadog/handler.lua
```diff
@@ -31,53 +30,51 @@ local get_consumer_id = {
...
-                         metric_config.sample_rate, metric_config.tags)
-
-      logger:send_statsd(string_format("%s.%s", fmt,  "total"),
+      logger:send_statsd("api.user.request.status",
                          1, logger.stat_types.counter,
-                         metric_config.sample_rate, metric_config.tags)
+                         metric_config.sample_rate,
+                         {
+                           "app:kong",
+                           "api_name:" .. api_name,
```
thibaultcha: `api_name` is incorrect in the context of Services/Routes

kong/plugins/datadog/handler.lua
AndrewFarley: Some re-alignment is necessary before merging this, thanks for the feedback.  At this point 'api' is just a namespace difference to denote the difference between data gathered from the stats endpoint by Datadog's agent (that don't have this extra 'api' namespace, and data that is pushed via statsd with this plugin.  This could easily be "statsd" to show the user that this is data pushed from statsd, or this could be "service" or "route" if that makes it make more sense.  Thoughts?

kong/plugins/datadog/handler.lua
AndrewFarley: Recommendations...?  "route"?  "service"?  What's more appropriate?

shashiranjan84:
@AndrewFarley PR https://github.com/Kong/kong/pull/3037 already provide the option of making metric name static.

AndrewFarley:
@shashiranjan84 Ahh, great, our commit/changelog is disturbingly similar also.  :)  Great minds think alike... thanks.  Go ahead and close this then, look forward to that merge.  And please, can we get that merged already?  It's already been >6 months on that PR.

PR #3467, opened 2018-05-17
perf(core) parallelize plugin iterator

kikito:


thibaultcha:
@kikito We should use the `perf` prefix for this commit message here, as usual.

kong/runloop/plugins_iterator.lua
```diff
@@ -85,129 +85,139 @@ local function load_plugin_configuration(route_id,
...
+  local    route_id = route    and    route.id or nil
+  local  service_id = service  and  service.id or nil
+  local consumer_id = consumer and consumer.id or nil
 
-    local api          = self.api
-    local route        = self.route
-    local service      = self.service
-    local consumer     = ctx.authenticated_consumer
+  local plugin_name = plugin.name
+  local threads = {}
```
thibaultcha: Let's preallocate this table with 9 array slots

kong/runloop/plugins_iterator.lua
```diff
@@ -85,129 +85,139 @@ local function load_plugin_configuration(route_id,
...
 
-      if consumer_id then
-        plugin_configuration = load_plugin_configuration(nil, nil, consumer_id, plugin_name, nil)
-        if plugin_configuration then
-          break
-        end
-      end
+    if cfg then
+      return cfg
+    end
```
thibaultcha: Seems like this could just be:
```
return cfg
```

(without the surrounding branch)?

kikito: No. `cfg` can be `nil` without having an error (if config found for the given parameters on the thread) and we want to keep iterating through the loop in that case.

kikito:
I renamed the commit to `perf` as well

kong/runloop/plugins_iterator.lua
```diff
@@ -85,129 +93,139 @@ local function load_plugin_configuration(route_id,
...
-      if route_id and service_id then
-        plugin_configuration = load_plugin_configuration(route_id, service_id, nil, plugin_name, nil)
-        if plugin_configuration then
-          break
-        end
-      end
+  for i = 1, len do
+    local _, cfg, err = assert(ngx_thread_wait(threads[i]))
+    if err then
+      ngx.ctx.delay_response = false
```
bungle: can this just be `ctx.delay_response = false`

kong/runloop/plugins_iterator.lua
```diff
@@ -85,129 +93,139 @@ local function load_plugin_configuration(route_id,
...
-        end
-      end
+  -- perform the most expensive query last, only if needed
+  -- TODO this needs real perf testing to see when it helps.
+  -- Right now it penalizes every single request with global plugins enabled,
+  -- but also uses less CPU. Since we cache negative results inside
+  -- load_plugin_configuration, that could negate the gains
+  local cfg, err = load_plugin_configuration(plugin_name, nil, nil, nil, nil)
+  if err then
+    ngx.ctx.delay_response = false
```
bungle: can this just be ctx.delay_response = false

thibaultcha:
Internal testing showed that this was not improving cache churning scenarios after all, and saw a considerable increase of CPU usage. Another attempt without using the `ngx.thread` API on cache hits preserved the performance and resource consumption offered by the current runloop, but made performance under cache churning scenarios worse than that of the current runloop (increased CPU usage and reduced req/s). Closing this as we will explore a few more ideas, and eventually revisit the plugins runloop anyway in the future.

PR #3470, opened 2018-05-17
docs(config) add lua_package_path default

darrenjennings:
### Summary

lua_package_path has a default in kong_defaults.lua but not reflective in kong.conf.default

### Full changelog
* docs(config) add lua_package_path default

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -485,10 +485,10 @@
                                  # This includes the certificates configured
                                  # for Kong's database connections.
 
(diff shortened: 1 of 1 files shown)


thibaultcha:
That was actually done on purpose for various historical reasons, since this value is primarily to make Kong developers' life easier. But I guess it's fine to add it for now, at least it will be informative. Eventually this default value should disappear and developers should maintain their own environments, we'll revisit then.

thibaultcha:
Manually merged with minor edits

PR #3472, opened 2018-05-17
Add hash on cookie

PepperTeasdale:
### Summary

Add ability to hash on cookie. Per a [discussion on the feature suggestion board](https://discuss.konghq.com/t/support-sticky-sessions-for-websockets/937/9) with @Tieske , this adds a `cookie` option for the `hash_on` feature, which works very similarly to the `header option. The use case for this is to allow load balancing with cookie injection, creating a sticky session for API consumers who all behind the same NAT gateway. Specifically this is useful for websockets.

### Full changelog

* Implement cookie source for hashing algorithm
  * Upstream config can use `cookie` for `hash_on` and `hash_fallback` value
  * Upstream config uses `hash_on_cookie` or `hash_fallback_cookie` for the name of the cookie to use
  * Upstream config uses `hash_on_cookie_path` or `hash_fallback_cookie_path` to add the path to the `Set-Cookie` header
* Add related unit and integration tests

### Notes

I'm new to Lua (and, hence, this project), so feel encouraged to tell me everything that's wrong with my code ;) I'm not sure if there's a place for me to add documentation.
I will clean up git history after making any requested changes/ getting final approval.

### DIFF
--- a/kong/dao/migrations/cassandra.lua
+++ b/kong/dao/migrations/cassandra.lua
@@ -748,5 +748,16 @@ return {
       CREATE INDEX IF NOT EXISTS ON consumers(custom_id);
       CREATE INDEX IF NOT EXISTS ON consumers(username);
     ]]
+  },
+  {
+    name = "2018-05-17-173100_hash_on_cookie",
+    up = [[
+      ALTER TABLE upstreams ADD hash_on_cookie text;
+      ALTER TABLE upstreams ADD hash_on_cookie_path text;
+    ]],
+    down = [[
+      ALTER TABLE upstreams DROP hash_on_cookie;
+      ALTER TABLE upstreams DROP hash_on_cookie_path;
+    ]]
   }
 }
--- a/kong/dao/schemas/upstreams.lua
+++ b/kong/dao/schemas/upstreams.lua
@@ -168,6 +168,7 @@ return {
         "consumer",
         "ip",
         "header",
+        "cookie",
       },
     },
     hash_fallback = {
@@ -179,6 +180,7 @@ return {
         "consumer",
         "ip",
         "header",
+        "cookie",
       },
     },
     hash_on_header = {
@@ -188,6 +190,15 @@ return {
     hash_fallback_header = {
       -- header name, if `hash_fallback == "header"`
       type = "string",
+    },
+    hash_on_cookie = {
+      -- cookie name, if `hash_on` or `hash_fallback` == `"cookie"`
+      type = "string",
+    },
+    hash_on_cookie_path = {
+      -- cookie path, if `hash_on` or `hash_fallback` == `"cookie"`
+      type = "string",
+      default = "/",
     },
     slots = {
       -- the number of slots in the loadbalancer algorithm
@@ -230,6 +241,20 @@ return {
       end
     end
 
+    if config.hash_on_cookie then
+      local ok, err = utils.validate_cookie_name(config.hash_on_cookie)
+      if not ok then
+        return false, Errors.schema("Cookie name: " .. err)
+      end
+    end
+
+    if config.hash_on_cookie_path then
+      local ok, err = check_http_path(config.hash_on_cookie_path)
+      if not ok then
+        return false, Errors.schema("Cookie path: " .. err)
+      end
+    end
+
     if (config.hash_on == "header"
         and not config.hash_on_header) or
        (config.hash_fallback == "header"
@@ -238,11 +263,25 @@ return {
                                   "but no header name provided")
     end
 
+    if (config.hash_on == "cookie" or config.hash_fallback == "cookie")
+        and not config.hash_on_cookie then
+      return false, Errors.schema("Hashing on 'cookie', " ..
+                                  "but no cookie name provided")
+    end
+
     if config.hash_on and config.hash_fallback then
       if config.hash_on == "none" then
         if config.hash_fallback ~= "none" then
           return false, Errors.schema("Cannot set fallback if primary " ..
                                       "'hash_on' is not set")
+        end
+
+      elseif config.hash_on == "cookie" then
+        if config.hash_fallback ~= "none" then
+          -- Cookie value will be set with the server response if not present,
+          -- so the fallback value would never be evaluated.
+          return false, Errors.schema("Cannot set `hash_fallback` if primary " ..
+                                      "`hash_on` is set to cookie")
         end
 
       else
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -1,5 +1,6 @@
 local pl_tablex = require "pl.tablex"
 local singletons = require "kong.singletons"
+local utils = require "kong.tools.utils"
 
 -- due to startup/require order, cannot use the ones from 'singletons' here
 local dns_client = require "resty.dns.client"
@@ -656,6 +657,25 @@ local create_hash = function(upstream)
       if type(identifier) == "table" then
         identifier = table_concat(identifier)
       end
+
+    elseif hash_on == "cookie" then
+      identifier = ngx.var["cookie_" .. upstream.hash_on_cookie]
+
+      -- If the cookie doesn't exist, create one and store in `ctx`
+      -- to be added to the "Set-Cookie" header in the response
+      if not identifier then
+        identifier = utils.uuid()
+
+        -- TODO: This should be added the `ngx.ctx.balancer_address`
+        -- structure, where other balancer-related values are stored,
+        -- when that is renamed/ refactored.
+        ctx.balancer_hash_cookie = {
+          key = upstream.hash_on_cookie,
+          value = identifier,
+          path = upstream.hash_on_cookie_path
+        }
+      end
+
     end
 
     if identifier then
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -7,6 +7,7 @@
 --
 -- In the `access_by_lua` phase, it is responsible for retrieving the route being proxied by
 -- a consumer. Then it is responsible for loading the plugins to execute on this request.
+local ck          = require "resty.cookie"
 local utils       = require "kong.tools.utils"
 local Router      = require "kong.router"
 local ApiRouter   = require "kong.api_router"
@@ -705,6 +706,18 @@ return {
 
           elseif matches then
             header[upstream_status_header] = matches[0]
+          end
+        end
+
+        local cookie_hash_data = ctx.balancer_hash_cookie
+        if cookie_hash_data then
+          local cookie = ck:new()
+          local ok, err = cookie:set(cookie_hash_data)
+
+          if not ok then
+            log(ngx.WARN, "failed to set the cookie for hash-based load balancing: ", err,
+                          " (key=", cookie_hash_data.key,
+                          ", path=", cookie_hash_data.path, ")")
           end
         end
       end
--- a/kong/tools/utils.lua
+++ b/kong/tools/utils.lua
@@ -810,4 +810,22 @@ _M.validate_header_name = function(name)
               "', allowed characters are A-Z, a-z, 0-9, '_', and '-'"
 end
 
+--- Validates a cookie name.
+-- Checks characters used in a cookie name to be valid
+-- a-z, A-Z, 0-9, '_' and '-' are allowed.
+-- @param name (string) the cookie name to verify
+-- @return the valid cookie name, or `nil+error`
+_M.validate_cookie_name = function(name)
+  if name == nil or name == "" then
+    return nil, "no cookie name provided"
+  end
+
+  if re_match(name, "^[a-zA-Z0-9-_]+$", "jo") then
+    return name
+  end
+
+  return nil, "bad cookie name '" .. name ..
+              "', allowed characters are A-Z, a-z, 0-9, '_', and '-'"
+end
+
 return _M
--- a/spec/01-unit/011-balancer_spec.lua
+++ b/spec/01-unit/011-balancer_spec.lua
@@ -506,6 +506,27 @@ describe("Balancer", function()
       })
       assert.are.same(crc32(value), hash)
     end)
+    describe("cookie", function()
+      it("uses the cookie when present in the request", function()
+        local value = "some cookie value"
+        ngx.var.cookie_Foo = value
+        local hash = balancer._create_hash({
+          hash_on = "cookie",
+          hash_on_cookie = "Foo",
+        })
+        assert.are.same(crc32(value), hash)
+        assert.is_nil(ngx.ctx.balancer_hash_cookie)
+      end)
+      it("creates the cookie when not present in the request", function()
+        balancer._create_hash({
+          hash_on = "cookie",
+          hash_on_cookie = "Foo",
+          hash_on_cookie_path = "/",
+        })
+        assert.are.same(ngx.ctx.balancer_hash_cookie.key, "Foo")
+        assert.are.same(ngx.ctx.balancer_hash_cookie.path, "/")
+      end)
+    end)
     it("multi-header", function()
       local value = { "some header value", "another value" }
       headers.HeaderName = value
@@ -561,6 +582,29 @@ describe("Balancer", function()
             hash_fallback_header = "HeaderName",
         })
         assert.are.same(crc32(table.concat(value)), hash)
+      end)
+      describe("cookie", function()
+        it("uses the cookie when present in the request", function()
+          local value = "some cookie value"
+          ngx.var.cookie_Foo = value
+          local hash = balancer._create_hash({
+            hash_on = "consumer",
+            hash_fallback = "cookie",
+            hash_on_cookie = "Foo",
+          })
+          assert.are.same(crc32(value), hash)
+          assert.is_nil(ngx.ctx.balancer_hash_cookie)
+        end)
+        it("creates the cookie when not present in the request", function()
+          balancer._create_hash({
+            hash_on = "consumer",
+            hash_fallback = "cookie",
+            hash_on_cookie = "Foo",
+            hash_on_cookie_path = "/",
+          })
+          assert.are.same(ngx.ctx.balancer_hash_cookie.key, "Foo")
+          assert.are.same(ngx.ctx.balancer_hash_cookie.path, "/")
+        end)
       end)
     end)
   end)
--- a/spec/02-integration/04-admin_api/07-upstreams_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/07-upstreams_routes_spec.lua
@@ -87,6 +87,29 @@ describe("Admin API: #" .. kong_config.database, function()
           assert.are.equal("HeaderFallback", json.hash_fallback_header)
         end
       end)
+      it_content_types("creates an upstream with hash_on cookie parameters", function(content_type)
+        return function()
+          local res = assert(client:send {
+            method = "POST",
+            path = "/upstreams",
+            body = {
+              name = "my.upstream",
+              hash_on = "cookie",
+              hash_on_cookie = "CookieName1",
+              hash_on_cookie_path = "/foo",
+            },
+            headers = {["Content-Type"] = content_type}
+          })
+          assert.response(res).has.status(201)
+          local json = assert.response(res).has.jsonbody()
+          assert.equal("my.upstream", json.name)
+          assert.is_number(json.created_at)
+          assert.is_string(json.id)
+          assert.are.equal("cookie", json.hash_on)
+          assert.are.equal("CookieName1", json.hash_on_cookie)
+          assert.are.equal("/foo", json.hash_on_cookie_path)
+        end
+      end)
       it_content_types("creates an upstream with 2 header hashes", function(content_type)
         return function()
           local res = assert(client:send {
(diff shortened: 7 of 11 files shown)


kong/dao/migrations/cassandra.lua
```diff
@@ -748,5 +748,16 @@ return {
       CREATE INDEX IF NOT EXISTS ON consumers(custom_id);
       CREATE INDEX IF NOT EXISTS ON consumers(username);
     ]]
+  },
+  {
+    name = "2018-05-17-173100_hash_on_cookie",
```
PepperTeasdale: Is there some automated way contributors are generating these migrations? I just manually created the timestamp/ name and it seemed to work.

kong/runloop/balancer.lua
```diff
@@ -656,6 +657,9 @@ local create_hash = function(upstream)
       if type(identifier) == "table" then
         identifier = table_concat(identifier)
       end
+
+    elseif hash_on == "cookie" then
+      identifier = ngx.var["cookie_" .. upstream[cookie_field_name]]
```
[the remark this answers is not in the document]
PepperTeasdale: Oh great, I misunderstood and thought that a plugin would manage the generation or the cookie, but I think this is a lot better. Thanks for the quick feedback!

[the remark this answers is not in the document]
PepperTeasdale: @Tieske I'm having a little trouble finding out how to add request cookies when proxying. All of the documentation I'm finding is for the latter part, adding cookies to the response in the `header-filter` phase.
Also, do you have any opinions about adding the [resty-cookie library](https://github.com/cloudflare/lua-resty-cookie) as a dependency? I'm noticing it's quite popular for this kind of use-case, but I don't like adding dependencies if not necessary (and it might not be).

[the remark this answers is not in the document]
PepperTeasdale: I _think_ I have something that will work, but would like some early feedback on it:
```lua
local get_request_cookies = function()
  local cookies = ngx.header["Cookie"] or {}

  if type(cookies) == "string" then
    cookies = {cookies}
  end

  return cookies

end

local create_hash = function(upstream)
  -- ... existing logic
  elseif hash_on == "cookie" then
  -- If the cookie doesn't exist, create it
  identifier = ngx.var["cookie_" .. upstream[cookie_field_name]]

  if not identifier then
    identifier = utils.uuid()
    ngx.ctx.HASH_ON_COOKIE = identifier

    local cookies = get_request_cookies()
    table.insert(cookies, cookie_field_name .. "=" .. identifier)
    ngx.header["Cookie"] = table_concat(cookies, "; ")
  end
end
```

PepperTeasdale:
@Tieske I _think_ this handles what you were asking for, please take a look. I will add tests of course, but I want to make sure this is on the right path before I go further.

kong/runloop/handler.lua
```diff
@@ -707,6 +708,17 @@ return {
             header[upstream_status_header] = matches[0]
           end
         end
+
+        if ctx.SET_HASH_ON_COOKIE then
```
PepperTeasdale: I put this in the `before` hook, presumably this will give plugins the ability to access/ modify the cookie value if there is some use case for it.

kong/runloop/handler.lua
```diff
@@ -707,6 +708,17 @@ return {
...
+          local cookies = header["Set-Cookie"] or {}
+
+          if type(cookies) == "string" then
+            cookies = {cookies}
+          end
+
+          insert(cookies, ctx.SET_HASH_ON_COOKIE)
+          header['Set-Cookie'] = cookies
+        end
```
Tieske: in the header:
```lua
local ck = require "resty.cookie"
```

and here we do something like:
```lua
local cookie_hash_data = ctx.SET_HASH_ON_COOKIE
if cookie_hash_data then
  local cookie, ok, err
  cookie, err = ck:new()
  if cookie then
    ok, err = cookie:set(cookie_hash_data)
  end
  if not ok then
    log(ngx.WARN, "failed to set hashing cookie, key=`", cookie_hash_data.key, 
                             "`, value=`", cookie_hash_data.value, 
                             "`, path=`", cookie_hash_data.path,
                             "` : ", err)
  end
end
```

kong/dao/schemas/upstreams.lua
```diff
@@ -189,6 +191,14 @@ return {
...
       type = "string",
     },
+    hash_on_cookie = {
+      -- cookie name, if `hash_on == "cookie"`
+      type = "string",
+    },
+    hash_fallback_cookie = {
+      -- cookie name, if `hash_fallback == "cookie"`
+      type = "string",
+    },
```
Tieske: some validation on the contents of the cookie name and path are needed here imo

kong/dao/schemas/upstreams.lua
```diff
@@ -238,6 +248,14 @@ return {
...
     end
 
+    if (config.hash_on == "cookie"
+        and not config.hash_on_cookie) or
+       (config.hash_fallback == "cookie"
+        and not config.hash_fallback_cookie) then
+      return false, Errors.schema("Hashing on 'cookie', " ..
+                                  "but no cookie name provided")
+    end
+
```
Tieske: also relevant here: if the primary is set to "cookie" then the fallback must be "none". Because if the cookie isn't set, it will be created, hence the fallback will never be invoked in that case.

kong/runloop/balancer.lua
```diff
@@ -656,6 +669,23 @@ local create_hash = function(upstream)
...
+        ngx.ctx.SET_HASH_ON_COOKIE = cookie_field_name ..
+                                     "=" ..
+                                     identifier ..
+                                     "; Path=" ..
+                                     upstream[cookie_path_field_name]
+
+        local cookies = get_request_cookies()
+        table.insert(cookies, cookie_field_name .. "=" .. identifier)
+        ngx.header["Cookie"] = table_concat(cookies, "; ")
+      end
```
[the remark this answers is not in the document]
PepperTeasdale: Right, that makes sense. Thanks! Although, I think we don't want to set the cookie again if it already exists, correct?

kong/tools/utils.lua
```diff
@@ -797,16 +797,16 @@ end
 -- a-z, A-Z, 0-9 and '-' are allowed.
 -- @param name (string) the header name to verify
 -- @return the valid header name, or `nil+error`
-_M.validate_header_name = function(name)
+_M.validate_http_token = function(name, token_type)
```
PepperTeasdale: [Cookie name spec](https://tools.ietf.org/html/rfc2616#section-2.2) technically allows a handful of other characters, but the same is true of headers. Seems fair to generalize these rules for both. Unless there are any objections?

spec/02-integration/05-proxy/09-balancer_spec.lua
```diff
@@ -1331,37 +1331,110 @@ for _, strategy in helpers.each_strategy() do
...
+                  local _, count1 = server1:done()
+                  local _, count2 = server2:done()
+
+                  -- verify
+                  assert(count1 == 0 or count1 == requests, "counts should either get 0 or ALL hits")
+                  assert(count2 == 0 or count2 == requests, "counts should either get 0 or ALL hits")
+                  assert(count1 + count2 == requests)
+                end)
+
+                it("when the cookie is not set", function()
```
PepperTeasdale: There's a lot of redundancy between these two new examples. Not sure we really need to test when the cookie is already set, if we're testing when it's not set. Thoughts?

kong/runloop/balancer.lua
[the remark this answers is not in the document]
Tieske: yes, you're right! good catch

kong/tools/utils.lua
Tieske: for the purpose it serves, I don't see any issues with that

kong/plugins/key-auth/schema.lua
```diff
@@ -12,7 +12,7 @@ end
 
 local function check_keys(keys)
   for _, key in ipairs(keys) do
-    local res, err = utils.validate_header_name(key, false)
+    local res, err = utils.validate_http_token(key, "header")
```
Tieske: where did this come from?

kong/tools/utils.lua
PepperTeasdale: Understood, and I was on the fence about this, too. Do you prefer me to just duplicate the logic for the cookie name validation? The only difference would be the error message.

kong/dao/schemas/upstreams.lua
```diff
@@ -189,6 +191,20 @@ return {
...
+    hash_on_cookie = {
+      -- cookie name, if `hash_on == "cookie"`
+      type = "string",
+    },
+    hash_on_cookie_path = {
+      -- cookie name, if `hash_on == "cookie"`
+      type = "string",
+      default = "/",
+    },
+    hash_fallback_cookie_path = {
```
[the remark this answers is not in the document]
PepperTeasdale: Should this not support falling back to a cookie value if the primary `hash_on` is `header` or `ip`? I can't think of any use cases for it, but currently `header` falling back on `ip` and vice-versa are supported.

kong/tools/utils.lua
Tieske: yes, let's do that for now. Let's get it compete functionality/test wise first and then look at some final things like style and stuff like this.

kong/dao/schemas/upstreams.lua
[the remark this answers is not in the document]
Tieske: yes it should. But with headers we can have 2 different headers, one for primary, one for fallback. With cookies, there is never more than one since the cookie is created when absent.

Hence having `hash_on_cookie` and `hash_on_cookie_path` is enough. We do not need `hash_fallback_cookie_path` (actually in the current state, `hash_fallback_cookie` is missing).

kong/dao/schemas/upstreams.lua
```diff
@@ -189,6 +191,20 @@ return {
       -- header name, if `hash_fallback == "header"`
       type = "string",
     },
+    hash_on_cookie = {
+      -- cookie name, if `hash_on == "cookie"`
```
Tieske: should be

```lua
-- cookie name, if `hash_on` or `hash_fallback` == `"cookie"`
```

kong/dao/schemas/upstreams.lua
```diff
@@ -189,6 +191,20 @@ return {
...
+      type = "string",
+    },
+    hash_on_cookie_path = {
+      -- cookie name, if `hash_on == "cookie"`
```
Tieske: should be

```lua
-- cookie path, if `hash_on` or `hash_fallback` == `"cookie"`
```

PepperTeasdale:
@Tieske Thanks for that excellent feedback! I think the latest changes address your requests, as well as add some more test coverage. PTAL

kong/tools/utils.lua
```diff
@@ -810,4 +810,22 @@ _M.validate_header_name = function(name)
               "', allowed characters are A-Z, a-z, 0-9, '_', and '-'"
 end
 
+--- Validates a cookie name.
+-- Checks characters used in a header name to be valid
+-- a-z, A-Z, 0-9 and '-' are allowed.
+-- @param name (string) the header name to verify
+-- @return the valid cookie name, or `nil+error`
+_M.validate_cookie_name = function(name)
```
PepperTeasdale: I've considered other possible ways to DRY this up, but I think they would mostly hurt the readability of the validation. It could be generalized to not refer to "cookie" or "header" at all in the validation message, and give the responsibility of the schema that is applying this validation to a field.

spec/02-integration/05-proxy/09-balancer_spec.lua
```diff
@@ -1331,37 +1331,126 @@ for _, strategy in helpers.each_strategy() do
...
+                local _, count2 = server2:done()
+
+                -- verify
+                assert(count1 == 0 or count1 == requests, "counts should either get 0 or ALL hits")
+                assert(count2 == 0 or count2 == requests, "counts should either get 0 or ALL hits")
+                assert(count1 + count2 == requests)
+              end)
+
+              describe("hashing on cookie", function()
+                it("when the cookie is set", function()
```
PepperTeasdale: I can't think of any edge cases that testing when a cookie _is_ set catch that are not caught in the next example (where the cookie is _not_ set). I would favor removing this test as redundant, but defer to your judgement.

spec/01-unit/011-balancer_spec.lua
```diff
@@ -506,6 +506,15 @@ describe("Balancer", function()
...
       assert.are.same(crc32(value), hash)
     end)
+    it("cookie", function()
+      local value = "some cookie value"
+      ngx.var.cookie_Foo = value
+      local hash = balancer._create_hash({
+          hash_on = "cookie",
+          hash_on_cookie = "Foo",
+      })
+      assert.are.same(crc32(value), hash)
```
Tieske: can we also add validation of the `ctx` variables?, so in this case that `ngx.ctx.SET_HASH_ON_COOKIE` is NOT created.

And add another test without a cookie set, that shows that `ngx.ctx.SET_HASH_ON_COOKIE` is properly created?

spec/02-integration/04-admin_api/07-upstreams_routes_spec.lua
```diff
@@ -246,7 +269,7 @@ describe("Admin API: #" .. kong_config.database, function()
             local json = cjson.decode(body)
             assert.same({ message = "Header: bad header name 'not a <> valid <> header name', allowed characters are A-Z, a-z, 0-9, '_', and '-'" }, json)
 
-            -- Invalid header
+            -- Invalid fallback header
```
Tieske: 👍

spec/02-integration/04-admin_api/07-upstreams_routes_spec.lua
```diff
@@ -311,6 +351,68 @@ describe("Admin API: #" .. kong_config.database, function()
...
+                hash_on = "cookie",
+                hash_on_cookie = "not a <> valid <> cookie name",
+              },
+              headers = {["Content-Type"] = content_type}
+            })
+            body = assert.res_status(400, res)
+            local json = cjson.decode(body)
+            assert.same({ message = "Cookie name: bad cookie name 'not a <> valid <> cookie name', allowed characters are A-Z, a-z, 0-9, '_', and '-'" }, json)
+
+            -- Invalid cookie
```
Tieske: duplicate of the one above? maybe this is supposed to be the fallback?

spec/02-integration/05-proxy/09-balancer_spec.lua
Tieske: agreed. But there is no test that validates that no cookie will be set if it was already provided, so maybe update this test to do that?

spec/02-integration/04-admin_api/07-upstreams_routes_spec.lua
PepperTeasdale: Yeah, copy/ paste error. I meant to check validating the cookie path on hash fallback.

PepperTeasdale:
@Tieske just addressed the last round of feedback, PTAL :)

Tieske:
@PepperTeasdale looks good to me now. Since I already did a lot of reviews on this, I requested @hishamhm for a final review.

Final thing to do would be to update the docs at https://github.com/Kong/getkong.org

spec/01-unit/011-balancer_spec.lua
```diff
@@ -506,6 +506,27 @@ describe("Balancer", function()
...
+      it("creates the cookie when not present in the request", function()
+        balancer._create_hash({
+          hash_on = "cookie",
+          hash_on_cookie = "Foo",
+          hash_on_cookie_path = "/",
+        })
+        assert.are.same(ngx.ctx.balancer_hash_cookie.key, "Foo")
+        assert.are.same(ngx.ctx.balancer_hash_cookie.path, "/")
+      end)
+    end)
```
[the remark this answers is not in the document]
PepperTeasdale: Good catch! Didn't see that :)

PepperTeasdale:
OK @hishamhm thank you for that feedback! This ought to be good now, and I've cleaned up the git history. My only question pertains to updating the docs. I don't see a `next` branch in the [website repo](https://github.com/Kong/getkong.org/branches). I did see the `release/ce-0.14.0` branch that has a no-merge PR in the project. Should I be branching off that or what? Thanks again!

spec/01-unit/011-balancer_spec.lua
```diff
@@ -562,6 +583,27 @@ describe("Balancer", function()
         })
         assert.are.same(crc32(table.concat(value)), hash)
       end)
+      describe("cookie", function()
```
hishamhm: It looks like this block is the same as the above, it should exercise `"cookie"` as the `hash_fallback` entry.

PepperTeasdale: Woops! Fixed

kong/runloop/handler.lua
```diff
@@ -707,6 +709,20 @@ return {
             header[upstream_status_header] = matches[0]
           end
         end
+
+        if cookie_hash_data then
+          local cookie, ok, err
```
thibaultcha: There is no benefits here in separating the declaration and initialization of those variables. Let's keep things more concise and declare/initialize them on the same line (by simply moving the `local` keyword to the below lines).

kong/runloop/handler.lua
```diff
@@ -707,6 +709,20 @@ return {
...
+
+          cookie = ck:new()
+          ok, err = cookie:set(cookie_hash_data)
+
+          if not ok then
+            log(ngx.WARN, "failed to set hashing cookie, key=`", cookie_hash_data.key,
```
thibaultcha: None of the logs produced by Kong use or should use backticks, instead, they use single quotes `'`. Let's update this to keep it consistent.

kong/runloop/handler.lua
```diff
@@ -690,6 +691,7 @@ return {
     before = function(ctx)
       local var = ngx.var
       local header = ngx.header
+      local cookie_hash_data = ctx.balancer_hash_cookie
```
thibaultcha: It seems like there is no point in retrieving this value outside of the `ctx.KONG_PROXIED` branch. It would also be more readable if we retrieved it just above the `if cookie_hash_data` branch created below.

kong/runloop/balancer.lua
```diff
@@ -656,6 +657,20 @@ local create_hash = function(upstream)
...
         identifier = table_concat(identifier)
       end
+
+    elseif hash_on == "cookie" then
+      identifier = ngx.var["cookie_" .. upstream.hash_on_cookie]
+
+      -- If the cookie doesn't exist, create it
+      if not identifier then
+        identifier = utils.uuid()
+        ngx.ctx.balancer_hash_cookie = {
```
thibaultcha: `ctx` is already locally cached above in this function, so better use it instead of invoking `ngx.ctx` again here.

kong/runloop/handler.lua
thibaultcha: Additionally, this gives the impression that it belongs to the `ngx.ctx.balancer_address` structure, in which we already store all balancer-related values for decision making and introspection purposes (its name should be revisited already). Doing so also allows us to reduce usage of the top-level `ctx` table (which is sensitive to naming conflicts), and self-documentation by adding this field to the ones documented with its [initialization code](https://github.com/Kong/kong/blob/next/kong/runloop/handler.lua#L525-L534). cc @Tieske @hishamhm

kong/runloop/handler.lua
thibaultcha: Additionally, I am not sure logging the value (just a UUID) is very insightful here. It mostly feels like noise.

A more concise error message which could be more readable and clearer from a user's POV would also mention the failure is related to "load balancing". Something similar to:
```
log(ngx.WARN, "failed to set the cookie for hash-based load balancing: ",
              err, " (key=", key, ", path=", path, ")")
```

Not the format and conciseness allows us to drop quotes altogether and still remain redable imho.

spec/02-integration/05-proxy/09-balancer_spec.lua
thibaultcha: note: it is good practice to write the intended result of a test in its description, rather than relying on the reader to infer the proper test results by reading the test case... In this case, a name such as `does not reply with Set-Cookie if cookie is already set` or something.

PepperTeasdale:
@thibaultcha Thanks for the good CR. I'll await your discussion on storing balancer cookie data the balancer structure. What would be a more descriptive name than `balancer_address` if so? `balancer_(config|data)` perhaps?

hishamhm:
I'd just go with `balancer_data`

thibaultcha:
@PepperTeasdale Choosing the proper name, if we were to rename this structure, would belong to another discussion, and should not be included in the scope of this PR. We'll take care of it later :)

Tieske:
> any thoughts about including the cookie parameters as values in the balancer structure instead?

Either way is fine. That structure needs to be renamed (and maybe torn apart?), so imo that is part of/follow up of the SDK. Until that refactor, in which we can take this one along, it doesn't really matter whether it lives inside a badly named structure, or lives outside with a proper name.

kong/runloop/balancer.lua
```diff
@@ -656,6 +657,25 @@ local create_hash = function(upstream)
...
+      identifier = ngx.var["cookie_" .. upstream.hash_on_cookie]
+
+      -- If the cookie doesn't exist, create one and store in `ctx`
+      -- to be added to the "Set-Cookie" header in the response
+      if not identifier then
+        identifier = utils.uuid()
+
+        -- TODO: This should be added the `ngx.ctx.balancer_address`
+        -- structure, where other balancer-related values are stored,
+        -- when that is renamed/ refactored.
```
PepperTeasdale: Are you ok with this comment to address the need to refactor the balancer structure? cc: @thibaultcha @Tieske

thibaultcha:
@PepperTeasdale Thanks for the patch. By the way, you are now eligible to receive your very own [Contributor T-shirt](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt) (details in the link). Enjoy! :)

PepperTeasdale:
Thank you! A+ contributor experience. I look forward contributing more to this and other kong-related projects :)

PR #3491, opened 2018-05-28
feat(JWT Plugin) Inject claims to request header for upstream servers

maudrid:
### Summary
JWT Plugin - Inject claims from token into request header.

The user can now configure any of the claims contained in the verified token to inject into the request header "x-{claim_name}" before it is forwarded to the API.

This allows the API to know details of the requester (eg. user_name) without having to decode the token.

This is meant for services that are secure behind the reverse proxy.

### Full changelog

* Added a new field (the claims that should be injected) to the jwt plugin configuration (schema) and added migration code to set defaults for postgres and casandra.

* Added the code that injects the configured claims into the header in the handler.

### DIFF
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -189,6 +189,12 @@ local function do_authentication(conf)
 
   set_consumer(consumer, jwt_secret, token)
 
+  for _, claim_name in ipairs(conf.inject_claims) do
+    if claims[claim_name] then
+      ngx_set_header("x-" .. claim_name, claims[claim_name])
+    end
+  end
+  
   return true
 end
 
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -189,9 +189,11 @@ local function do_authentication(conf)
 
   set_consumer(consumer, jwt_secret, token)
 
-  for _, claim_name in ipairs(conf.inject_claims) do
-    if claims[claim_name] then
-      ngx_set_header("x-" .. claim_name, claims[claim_name])
+  if conf.inject_claims then
+    for _, claim_name in ipairs(conf.inject_claims) do
+      if claims[claim_name] then
+        ngx_set_header("x-" .. claim_name, claims[claim_name])
+      end
     end
   end
   
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -95,6 +95,23 @@ local function set_consumer(consumer, jwt_secret, token)
 
 end
 
+local function iter(config_array)
+  return function(config_array, i, previous_name, previous_value)
+    i = i + 1
+    local current_pair = config_array[i]
+    if current_pair == nil then -- n + 1
+      return nil
+    end
+
+    local current_name, current_value = current_pair:match("^([^:]+):*(.-)$")
+    if current_value == "" then
+      current_value = nil
+    end
+
+    return i, current_name, current_value
+  end, config_array, 0
+end
+
 local function do_authentication(conf)
   local token, err = retrieve_token(ngx.req, conf)
   if err then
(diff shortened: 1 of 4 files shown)


maudrid:
Question, I have made the same code change for the master branch.
Should I do a pull request for that?

maudrid:
I can't tell if the CI failure is because of my code.
Can someone please help to see if I need to change anything?

kong/plugins/jwt/schema.lua
```diff
@@ -28,6 +28,7 @@ return {
     anonymous = {type = "string", default = "", func = check_user},
     run_on_preflight = {type = "boolean", default = true},
     maximum_expiration = {type = "number", default = 0, func = check_positive},
+	inject_claims = {type = "array", default = {}},
```
thibaultcha: style: please watch the indentation here :)

kong/plugins/jwt/migrations/postgres.lua
```diff
@@ -104,4 +104,22 @@ return {
...
   },
+  {
+    name = "2018-05-28-151700_jwt_inject_claims_default",
+    up = function(_, _, dao)
+      for ok, config, update in plugin_config_iterator(dao, "jwt") do
+        if not ok then
+          return config
+        end
+        if config.inject_claims == nil then
+          config.inject_claims = {}
```
thibaultcha: It would result in faster and safer migrations if this was included in the above migration instead. We should also rename it to something like `jwt_max_exp_and_inject_claims`.

kong/plugins/jwt/handler.lua
```diff
@@ -189,6 +189,12 @@ local function do_authentication(conf)
 
   set_consumer(consumer, jwt_secret, token)
 
+  for _, claim_name in ipairs(conf.inject_claims) do
```
thibaultcha: Better wrap this loop within a branch checking for the existence of `conf.inject_claims`, we've had related issues in the past:
```lua
if conf.inject_claims then

end
```

kong/plugins/jwt/handler.lua
```diff
@@ -189,6 +189,12 @@ local function do_authentication(conf)
...
+    if claims[claim_name] then
+      ngx_set_header("x-" .. claim_name, claims[claim_name])
```
thibaultcha: We should maybe prefix this header with something like `x-jwt` or `x-jwt-claim`. What do you think? Have you seen this done anywhere else, is it part of any standard you are aware of?

thibaultcha:
> I have made the same code change for the master branch.
> Should I do a pull request for that?

At this point in the release cycle, we would be more comfortable releasing this as part of 0.14, which means that targeting the next branch is appropriate. Note also that because of the migration, it is required to target the next branch anyway.

thibaultcha:
Oops, mis-clicked. At least this should answer your second question:

> I can't tell if the CI failure is because of my code.

Likely not, since you did not add tests. But the CI will re-run now, and ideally get rid of any spurious failures.

kong/plugins/jwt/schema.lua
maudrid: Will fix in the next commit

kong/plugins/jwt/handler.lua
maudrid: I had a look into this now.
RFC 6648 of 2012 states that it discourages custom headers to be prefixed with x-.
Though, the set_consumer function in the same plugin is adding the consumer id, consumer custom id and consumer user name using the same x- prefix.
So if we go with the way that headers are currently added by the same plugin, then just x- should be used and not have the 'jwt' added. 

I'm happy to do it anyway you choose.

maudrid:
I took @bungle 's advice and made it possible to pick the name of the header.
It works in the same way as the request/response transformer plugins, by using the format claimName:headerName.

The tests are still outstanding.
This is my first time using Lua and contributing to this project, so it may take some time to do the tests.
All other raised issues should be resolved now.

tklein:
Hi,
I'd like to know how we could solve the outstanding issue to get this merged.

We'd love to have that feature also...

TIA

maudrid:
> Hi,
> I'd like to know how we could solve the outstanding issue to get this merged.
> 
> We'd love to have that feature also...
> 
> TIA

Honestly I don't have the time or even know where to begin to please the merge gods.
This was done on such an old version of Kong.
At my company we just have our own fork with this one change in it.

If I had to do this again, I would just make this a separate extension that can be build without having to merge it into the main line.
The changes are quite trivial to do, so maybe have a look at what was done and decide how you want to continue.

tklein:
@maudrid : Thanks for the quick reply. 

Offerering praise and sacrifices to the gods of merge doesn't them to help though...

I'll check out how to proceed and start digging into the changes first.

hbagdi:
@maudrid Would it be possible to rebase this once more? Apologies for the delay on our side.

PR #3492, opened 2018-05-28
chore(config) set default db_cache_ttl to 0

kikito:


Tieske:
also need an update to https://github.com/Kong/kong/blob/master/kong.conf.default#L411

PR #3496, opened 2018-05-30
chore(rockspec) bump lua-resty-healthcheck to 0.4.2

hishamhm:


thibaultcha:
@hishamhm Could we send this to master maybe?

hishamhm:
But we don't bump dependencies on point releases, right?

hishamhm:
Or is it fine to bump in a point-release, given that it is itself a point-release? (That would make sense.)

thibaultcha:
We do, but it depends on the content of the dependency bump. In this case, we have a fix and a refactor, so it is a bit hard to send it to master indeed...

thibaultcha:
Pushed to master!

PR #3501, opened 2018-05-31
Chore/update openssl

kikito:


kong/templates/nginx_kong.lua
```diff
@@ -79,7 +79,7 @@ server {
 > if proxy_ssl_enabled then
     ssl_certificate ${{SSL_CERT}};
     ssl_certificate_key ${{SSL_CERT_KEY}};
-    ssl_protocols TLSv1.1 TLSv1.2;
+    ssl_protocols TLSv1.1 TLSv1.2 TLSv1.3;
```
thibaultcha: TLS 1.3 is only available with OpenSSL 1.1.1, which is still in beta, so this actually doesn't belong here (yet).

PR #3505, opened 2018-06-01
tests(helpers) try harder to extract the OpenResty version

hishamhm:
This doesn't look like it should have been necessary, but at least one strange CI failure could have been avoided by this: https://travis-ci.com/Kong/kong-private/jobs/126566137


Tieske:
may I suggest we comment out your solution, and add extra debugging info to be displayed, such that we can fix the root-cause when it reoccurs?

PR #3507, opened 2018-06-04
feat/core add options to responses.send functions

albertored:
In some situations one has an already encoded body and wants to send it
as a response. Actually the only way of sending it with the `responses`
module is to decode the body to a table and pass it to the `send`
function that encodes it back to a json string.

Setting `encoded_content = true` in the new `options` parameter
now takes the content without modifying it, no more extra
decoding/encoding. This table can also be used in future if we need
more configuration options.

If you think this can be useful I will also add a couple of tests.

### DIFF
--- a/kong/tools/responses.lua
+++ b/kong/tools/responses.lua
@@ -103,7 +103,7 @@ local function send_response(status_code)
   -- @see https://github.com/openresty/lua-nginx-module
   -- @param content (Optional) The content to send as a response.
   -- @return ngx.exit (Exit current context)
-  return function(content, headers)
+  return function(content, headers, options)
     local ctx = ngx.ctx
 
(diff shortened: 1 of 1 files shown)


thibaultcha:
Hi,

Thanks for the suggestion. We are about to roll out the first SDK to develop Kong plugins, which will include high-level functions to manipulate the request/response, and produce responses from plugins. The `kong.responses` module was always a temporary solution (and originally not intended for usage by external plugins).

As such, we would rather avoid introducing new features to a legacy API. The new SDK is scheduled to be released with 0.14, in just a few weeks. And it will offer the behaviour you implemented here (custom encoding).

Thanks!

albertored:
ok, thanks for the info!

Is there any article/thread/discussion with the feature this new SDK will include? I'm curious now :-)

PR #3511, opened 2018-06-05
refactor(meta) add meta._SERVER_TOKENS

bungle:
### Summary

Add `kong.meta._SERVER_TOKENS` and change codebase to use that everywhere.

### DIFF
--- a/kong/tools/responses.lua
+++ b/kong/tools/responses.lua
@@ -17,7 +17,8 @@
 --
 --    -- Raw send() helper:
 --    return responses.send(418, "This is a teapot")
-
+local singletons = require "kong.singletons"
+local constants = require "kong.constants"
 local cjson = require "cjson.safe"
 local meta = require "kong.meta"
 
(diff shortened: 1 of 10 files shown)


kong/plugins/aws-lambda/handler.lua
```diff
@@ -150,6 +153,10 @@ function AWSLambdaHandler:access(conf)
     ngx.status = res.status
   end
 
+  if singletons.configuration.enabled_headers[constants.HEADERS.VIA] then
+    ngx.header["Via"] = meta._SERVER_TOKENS
```
thibaultcha: We should also test this (or make it part of a separate commit)

kong/tools/responses.lua
```diff
@@ -123,7 +123,18 @@ local function send_response(status_code)
...
+      if singletons.configuration.enabled_headers[constants.HEADERS.SERVER] then
+        ngx.header["Server"] = server_header
+
+      else
+        ngx.header["Server"] = nil
+      end
+
+    else
+      ngx.header["Server"] = server_header
+    end
```
thibaultcha: Ditto: probably belong to a separate commit (i.e. in the PR already opened for delayed response)

bungle: I think you wanted to comment about request-termination plugin changes here? I will make separate PR for the two plugins.

spec/01-unit/009-responses_spec.lua
```diff
@@ -192,4 +192,33 @@ describe("Response helpers", function()
       assert.spy(s).was.called_with(ngx.ctx)
     end)
   end)
+
+  describe("server tokens", function()
+    it("are send by default", function()
```
hishamhm: typo: `are sent by default` (likewise below)

hishamhm:
I guess the first commit should be `refactor` instead of `feat` since this is an internal code reorganization without visible functionality change

PR #3513, opened 2018-06-05
fix(request-termination) delayed response

bungle:
### Summary

Fix `request-termination` delayed response

### DIFF
--- a/kong/plugins/request-termination/handler.lua
+++ b/kong/plugins/request-termination/handler.lua
@@ -1,40 +1,80 @@
 local BasePlugin = require "kong.plugins.base_plugin"
+local singletons = require "kong.singletons"
 local responses = require "kong.tools.responses"
+local constants = require "kong.constants"
 local meta = require "kong.meta"
 
-local server_header = meta._NAME .. "/" .. meta._VERSION
+
+local ngx = ngx
+
+
+local server_header = meta._SERVER_TOKENS
+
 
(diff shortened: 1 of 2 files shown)


kong/plugins/request-termination/handler.lua
```diff
@@ -1,40 +1,82 @@
...
+        content_type = conf.content_type,
+      }
 
-    ngx.say(body)
+      ctx.delayed_response_callback = flush
 
-    return ngx.exit(status_code)
-   else
-    return responses.send(status_code, message)
+      coroutine.yield()
```
thibaultcha: Same comment as the azure-functions plugin

kong/plugins/request-termination/handler.lua
```diff
@@ -1,40 +1,82 @@
...
+  end
+
+  ngx.status = status
+
+  if singletons.configuration.enabled_headers[constants.HEADERS.SERVER] then
+    ngx.header[constants.HEADERS.SERVER] = server_header
+
+  else
+    ngx.header[constants.HEADERS.SERVER] = nil
+  end
```
thibaultcha: This plugin should already have tests ensuring that this header is set/unset when terminating a request, and I am not seeing them. Perhaps this underlying implementation change warrants adding those tests.

bungle: Added

PR #3515, opened 2018-06-05
fix(plugins) raise ip-restriction priority to 3000

nijikokun:
### Summary

Raises the priority of the `ip-restriction` plugin to `3000` so that it runs before other plugins, like `bot-detection` to prevent code execution that is not required. 

See: https://getkong.org/docs/0.13.x/plugin-development/custom-logic/#plugins-execution-order

### DIFF
--- a/kong/plugins/ip-restriction/handler.lua
+++ b/kong/plugins/ip-restriction/handler.lua
@@ -18,7 +18,7 @@ local cache = {}
 
 local IpRestrictionHandler = BasePlugin:extend()
 
-IpRestrictionHandler.PRIORITY = 990
+IpRestrictionHandler.PRIORITY = 3000
 IpRestrictionHandler.VERSION = "0.1.0"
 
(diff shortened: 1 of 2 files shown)


thibaultcha:
@nijikokun This needs to go against `next`. Let's not close and reopen the PR, simply changing its base branch (after rebase) will do.

Also, we will need you to update the 0.14 version of [this document](https://github.com/Kong/getkong.org/blob/master/app/docs/0.13.x/plugin-development/custom-logic.md) to reflect this change. To be opened against the 0.14 branch of getkong.org. You should just copy the `plugin-development/custom-logic.md` into the `0.14.x` tree of this branch.

Thanks!

nijikokun:
Sounds good to me :+1:

thibaultcha:
> Let's *not* close and reopen the PR

If so then please delete the unmerged branch then :)

nijikokun:
I can't believe I misread that. Apologies!

PR #3516, opened 2018-06-05
fix(plugins) raise ip-restriction priority to 3000

nijikokun:
### Summary

Raises the priority of the `ip-restriction` plugin to `3000` so that it runs before other plugins, after Zipkin, just as `bot-detection` does to prevent code execution that is not required. 

See: https://getkong.org/docs/0.13.x/plugin-development/custom-logic/#plugins-execution-order

### DIFF
--- a/kong/plugins/ip-restriction/handler.lua
+++ b/kong/plugins/ip-restriction/handler.lua
@@ -18,7 +18,7 @@ local cache = {}
 
 local IpRestrictionHandler = BasePlugin:extend()
 
-IpRestrictionHandler.PRIORITY = 990
+IpRestrictionHandler.PRIORITY = 3000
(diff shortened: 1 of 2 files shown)


Tieske:
Since this plugin can be attached to a consumer, it MUST run after the auth plugions.

thibaultcha:
What would be the use case for specifying this plugin to a Consumer?

Tieske:
I don't know really, but it will break expectations if it doesn't do that. So the alternative is to add the `no_consumer` flag to the IP-restriction plugin.

Either way something needs to be updated.

thibaultcha:
Update for visibility purposes: this was reverted in https://github.com/Kong/kong/commit/380535a6dd220f985d53b3075bb0009938c554ec, before the 0.14.0 release.

Issue #3521, opened 2018-06-06
Response transformer causes runtime error


michael-min-h:
### Steps To Reproduce

1. enable plugin basic auth
2. enable plugin response transformer and set config.remove.json=ip
3. call the api without auth information
4. the below error can be found in error.log
[error] 107#0: *2494522 failed to run body_filter_by_lua*: ...re/lua/5.1/kong/plugins/response-transformer/handler.lua:37: bad argument #1 to 'table_concat' (table expected, got nil)
stack traceback:
	[C]: in function 'table_concat'
	...re/lua/5.1/kong/plugins/response-transformer/handler.lua:37: in function 'body_filter'
	/usr/local/share/lua/5.1/kong/init.lua:410: in function 'body_filter'

### Additional Details & Logs
- Kong version (0.12.3)
- Use kong docker
- I found the function ResponseTransformerHandler:access will not be invoked if authentication failed, so ctx.rt_body_chunks is not initialized. However the function ResponseTransformerHandler:body_filter is still invoked, ctx.rt_body_chunks is used in below code:

    if eof then
      local body = body_filter.transform_json_body(conf, table_concat(ctx.rt_body_chunks))
      ngx.arg[1] = body
    else
      ctx.rt_body_chunks[ctx.rt_body_chunk_number] = chunk
      ctx.rt_body_chunk_number = ctx.rt_body_chunk_number + 1
      ngx.arg[1] = nil
    end

thibaultcha:
Hi,

Thank you for the report! I will label this as a bug and we will get to it as soon as we can manage. We are currently quite busy with the upcoming 0.14 release.

We would also very much welcome a contributed patch to fix this. Kong contributors get credited in our changelog, receive their very own [Contributor T-shirt](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt), and of course, enjoy the satisfaction given by contributing back to an open source community :)

PR #3524, opened 2018-06-08
fix(response-transformer) initialize `rt_body_chunks` and `rt_body_ch…

nateslo:
### Summary

 When `ResponseTransformerHandler:access` is not executed, `ResponseTransformerHandler:body_filter` attempts to access nil `ctx.rt_body_chunks` and nil `ctx.rt_body_chunk_number`. This PR adds checks for both and initializes them to defaults if nil.  

A test case has been added to verify there is no error if `ResponseTransformerHandler:access` is not executed.

### Full changelog

* Add defaults for `ctx.rt_body_chunks`  and `ctx.rt_body_chunk_number` in `ResponseTransformerHandler:body_filter`
* Add test case (unauthorized request, skips `ResponseTransformerHandler:access`)
* Remove `access` handler - ctx variables now initialized in `body_filter`

### Issues resolved

Fix [#3521](https://github.com/Kong/kong/issues/3521)

### DIFF
--- a/kong/plugins/response-transformer/handler.lua
+++ b/kong/plugins/response-transformer/handler.lua
@@ -13,15 +13,6 @@ function ResponseTransformerHandler:new()
   ResponseTransformerHandler.super.new(self, "response-transformer")
 end
 
-function ResponseTransformerHandler:access(conf)
-  ResponseTransformerHandler.super.access(self)
-
-  local ctx = ngx.ctx
-
-  ctx.rt_body_chunks = {}
-  ctx.rt_body_chunk_number = 1
-end
-
 function ResponseTransformerHandler:header_filter(conf)
   ResponseTransformerHandler.super.header_filter(self)
   header_filter.transform_headers(conf, ngx.header)
--- a/kong/plugins/response-transformer/handler.lua
+++ b/kong/plugins/response-transformer/handler.lua
@@ -40,6 +40,6 @@ function ResponseTransformerHandler:body_filter(conf)
 end
 
 ResponseTransformerHandler.PRIORITY = 800
-ResponseTransformerHandler.VERSION = "0.1.0"
+ResponseTransformerHandler.VERSION = "0.1.1"
 
 return ResponseTransformerHandler
--- a/kong/plugins/response-transformer/handler.lua
+++ b/kong/plugins/response-transformer/handler.lua
@@ -33,6 +33,13 @@ function ResponseTransformerHandler:body_filter(conf)
   if is_body_transform_set(conf) and is_json_body(ngx.header["content-type"]) then
     local ctx = ngx.ctx
     local chunk, eof = ngx.arg[1], ngx.arg[2]
+
+    -- Initializes context here in case this plugin's access phase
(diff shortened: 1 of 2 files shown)


kong/plugins/response-transformer/handler.lua
```diff
@@ -33,6 +33,13 @@ function ResponseTransformerHandler:body_filter(conf)
   if is_body_transform_set(conf) and is_json_body(ngx.header["content-type"]) then
     local ctx = ngx.ctx
     local chunk, eof = ngx.arg[1], ngx.arg[2]
+
+    -- Initializes context here in case this plugin's access phase
+    -- did not run - and hence `rt_body_chunks` and `rt_body_chunk_number`
+    -- were not initialized
+    ctx.rt_body_chunks = ctx.rt_body_chunks or {}
+    ctx.rt_body_chunk_number = ctx.rt_body_chunk_number or 1
```
Tieske: The access handler above does nothing but setting the defaults, so that entire handler can be removed now.

nateslo: @Tieske `access` handler removed

kong/plugins/response-transformer/handler.lua
```diff
@@ -33,6 +24,13 @@ function ResponseTransformerHandler:body_filter(conf)
   if is_body_transform_set(conf) and is_json_body(ngx.header["content-type"]) then
     local ctx = ngx.ctx
     local chunk, eof = ngx.arg[1], ngx.arg[2]
+
+    -- Initializes context here in case this plugin's access phase
+    -- did not run - and hence `rt_body_chunks` and `rt_body_chunk_number`
+    -- were not initialized
```
Tieske: these comments are now outdated

thibaultcha:
@nateslo Could you bump the version to `0.1.1`? Thanks!

spec/03-plugins/16-response-transformer/04-filter_spec.lua
```diff
@@ -97,5 +116,19 @@ for _, strategy in helpers.each_strategy() do
         assert.equals([[this is a / test]], json.uri_args)
       end)
     end)
+
+    describe("unauthorized", function()
```
thibaultcha: Could you rename this block to "regressions"? Thanks

spec/03-plugins/16-response-transformer/04-filter_spec.lua
```diff
@@ -97,5 +116,19 @@ for _, strategy in helpers.each_strategy() do
...
+      it("doesn't err when unauthorized", function()
```
thibaultcha: Within this block, it'd be nice to include a link to the issue this regression test fixes just like https://github.com/Kong/kong/blob/master/spec-old-api/01-unit/010-router_spec.lua#L927

thibaultcha: We should also be more explicit in the name of this test:
```
it("does not throw an error when request was short-circuited in access phase")
```

thibaultcha:
Manually merged to master, thank you!

PR #3526, opened 2018-06-10
fix(ip-restriction) disallow plugin on Consumers

thibaultcha:
A followup commit to dcb253643db67d775d58ec09ec7c97cef32f2961, which
made the ip-restriction plugin run before authentication ones. Thus, it
does not make sense to allow this plugin to be added to a Consumer
anymore (since they will always be identified after this plugin runs).

It is arguable whether there is any value at all in allowing this plugin
to run on specific Consumers, since L7 authentication and IP restriction
reside on different levels of the stacks altogether. Most use-cases
would require this plugin to run first, as per the aforementioned
commit.

If this were to be reverted, then so should
dcb253643db67d775d58ec09ec7c97cef32f2961.

### DIFF
--- a/kong/plugins/ip-restriction/schema.lua
+++ b/kong/plugins/ip-restriction/schema.lua
@@ -14,6 +14,7 @@ local function validate_ips(v, t, column)
 end
 
 return {
+  no_consumer = true,
   fields = {
     whitelist = {type = "array", func = validate_ips},
     blacklist = {type = "array", func = validate_ips}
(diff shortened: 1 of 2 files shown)


Tieske:
I have no preference for allowing or disallowing it on consumers. But how to deal with existing plugins configured on consumers?

thibaultcha:
Not much to do, except documentation and _eventually_ a migration to detect those, although the timeline is somewhat short for this. Happy to merge one if you have time to contribute one :+1:

thibaultcha:
Reverting dcb2536 and not going through with this change by fear of breaking existing workflows. Thanks @Tieske for the feedback on the initial PR!

PR #3530, opened 2018-06-11
feat(conf) add support to configure nginx directives via kong.conf

hbagdi:
Problem
-------
Kong ships with an NGINX template which is rendered when Kong starts.
There exists no mechanisms to add/update any NGINX directive to the
`nginx.conf` used to run Kong. To change or add any directive, user has
to use a custom NGINX template which has to be synced with Kong for a
release which introduces changes to Kong's template.
Including options in `kong.conf` to configure NGINX directives is not a
good solution since the list will be endless.
This problem can be seen in #3010, #3323 and #3382.

Solution
--------
There needs to be a flexible  way to specify any NGINX directive
via Kong's config file without Kong needing to maintain a list of all
NGINX directives.
While a clean and ideal solution would be #2355, this commit
adopts a simpler as discussed like the one proposed in #2675.

NGINX directives can be specified using config variables with prefixes,
which help determine the block in which to place a directive.
eg:

`nginx_proxy_add_header=Header-Name header-value` will add a `add_header
Header-Name header-value;` directive in the proxy server block of Kong.

`nginx_http_lua_shared_dict=custom_cache 2k` will add a a
`lua_shared_dict custom_cache 2k;` directive to HTTP block of Kong.


kong/conf_loader.lua
```diff
@@ -448,6 +455,24 @@ local function parse_listeners(values)
...
+local function parse_nginx_directives(prefix, conf)
+  local directives = {}
+
+  if type(prefix) ~= "string" or type(conf) ~= "table" then
+    return directives
+  end
+
+  for k, v in pairs(conf) do
+    if type(k) == "string" then
+      local _ , _ , directive= string.find(k, prefix .. "(.+)")
```
hishamhm: minor: spacing (`local _, _, directive = string.find`...)

hishamhm:
A wild idea that came to mind for supporting duplicated nginx directives: how about taking the key and runnig `k:gsub("__", " ")` at the appropriate place, so one can add multiple entries of the same nginx key using syntax like this: `KONG_NGINX_HTTP_LUA_SHARED_DICT__CUSTOM_CACHE=200k`, producing `lua_shared_dict custom_cache 200k;` in the nginx config file.

(The downside is that the arguments would always become lowercase when sent via environment variables (which may or may not be okay for headers), but for kong.conf that wouldn't be a problem.)

kong/cmd/utils/env.lua
```diff
@@ -0,0 +1,35 @@
+local pl_utils = require "pl.utils"
+local log = require "kong.cmd.utils.log"
+local fmt = string.format
+
+
+local cmd = [[ printenv ]]
```
thibaultcha: Maybe there is no need for this extra variable anymore, it is used only once.

kong/cmd/utils/env.lua
```diff
@@ -0,0 +1,35 @@
...
+
+
+-- returns table and err
+local function read_all()
+  log.debug("reading environment variables: %s", cmd)
+  local vars = {}
+  local success, ret_code, stdout, stderr = pl_utils.executeex(cmd)
+  if not success or ret_code ~= 0 then
+    return nil, fmt("could not read environment variables (exit code: %d): %s",
+           ret_code, stderr)
```
thibaultcha: style: alignment

kong/cmd/utils/env.lua
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  for line in stdout:gmatch("[^\r\n]+") do
+    local i = string.find(line, "=") -- match first =
+    if i then
+      local k = string.sub(line,1, i-1)
```
thibaultcha: style: `sub(line, 1, i - 1)`

kong/cmd/utils/nginx_signals.lua
```diff
@@ -107,6 +107,25 @@ function _M.start(kong_conf)
...
+  end
+
+  local cmd = fmt("KONG_NGINX_CONF_CHECK=true %s -t -p %s -c %s",
+                  nginx_bin, kong_conf.prefix, "nginx.conf")
+  log.debug("checking nginx conf: %s", cmd)
+
+  local ok, retcode, _, stderr = pl_utils.executeex(cmd)
+  if not ok then
+    return false, ("failed to validate nginx configuration " ..
+           "(exit code %d):\n%s"):format(retcode ,stderr)
```
thibaultcha: style: alignment and space _after_ the comma

kong/cmd/utils/nginx_signals.lua
```diff
@@ -107,6 +107,25 @@ function _M.start(kong_conf)
...
+  end
+
```
thibaultcha: Let's also add a debug log when the validation succeededs, otherwise the user just sees "checking nginx conf" and does not know what happened after that.

kong/conf_loader.lua
```diff
@@ -514,6 +539,48 @@ local function load(path, custom_conf)
...
+      end
+    end
+
+    for _, prefix in pairs(CONF_KEY_PREFIXES) do
+      find_dynamic_keys(custom_conf, prefix)
+      find_dynamic_keys(kong_env_vars, prefix)
+      find_dynamic_keys(from_file_conf, prefix)
+    end
+
+    defaults = tablex.merge(defaults, dynamic_keys, true)
```
thibaultcha: Can you add a comment that this is building the union? This way the intent is clearer (as discussed) to readers who see `tablex.merge()` usage several times in this module. See below:
```lua
conf = tablex.merge(conf, defaults) -- intersection (remove extraneous properties)
```

kong/init.lua
```diff
@@ -29,6 +29,13 @@ require "resty.core"
 local constants = require "kong.constants"
 
 do
+  -- if we're running nginx -t then don't initialize
+  if os.getenv("KONG_NGINX_CONF_CHECK") then
+    return {
+      init = function()
+      end,
+    }
+  end
```
thibaultcha: This should go below the check for required shms, otherwise configs lacking the shm would error during the actual `start`, and yet configuration validation would report a success...

kong/cmd/utils/nginx_signals.lua
hbagdi: UNIX philosophy of no output is good?

kong/cmd/utils/env.lua
hbagdi: It is used for the debug log as well.

spec/02-integration/05-proxy/15-custom_nginx_directive_spec.lua
```diff
@@ -0,0 +1,58 @@
...
+
+
+describe("Custom NGINX directives", function()
+  local proxy_client
+  local bp
+
+  local function start(config)
+    return function()
+      bp.routes:insert {
+        hosts = { "headers-inspect.com" },
```
hishamhm: Ah, a super minor thing but which I think we should start paying attention in our tests (perhaps worth an addition to the style guide even?) — I suggest we stop using "valid-looking" URLs in tests, even if we're not hitting them (we've had in the past tests that used an innocent-looking URL which ended up pointing to a porn site!) — we can use instead `example.com` or any URL using the reserved `.test` TLD (which then allows for "descriptive" test hostnames such as `headers-inspect.test`)

hbagdi: Thanks for sharing pointing this out!
I like the idea of using the `.test` TLD. I've updated the code to use `headers-inspect.test`.

p0pr0ck5: +1

> we've had in the past tests that used an innocent-looking URL which ended up pointing to a porn site

I've also seen cases of test/example code in public documentation using "domain.com", which at one point hosted malware; customers who blindly copy-pasted were in for a nasty surprise :p 

IANA has reserved `example.com` for cases like this, FWIW. https://www.iana.org/domains/reserved

Just my $0.02.

thibaultcha:
I like @hishamhm's idea of specifying multiple directives as such. Another approach we discussed with @hbagdi was to use semicolons to separate multiple values for a directive (`ngx_http_lua_shared_dict=cats 12k; dogs24k`). Yesterday we decided to postpone support for multiple instances of the same directive for now, in order to comfortably ship this in 0.14 and re-assessing later (also based on potential feature requests for this). So, delaying this for now unless someone else sees it as a blocker (or anything else), I am polishing this branch locally and am planning on pushing to next very soon.

thibaultcha:
Manually merged, thanks!

TijoVarghese:
I see that if we are trying to add multiple values for a directive, the later overrides the former value.
For eg: 
`nginx_http_log_format = a 'aa'
  nginx_http_log_format = b 'bb'
`
b will gets overrided for log_format

tzssangglass:
> 我看到，如果我们尝试为一个指令添加多个值，则后者会覆盖前一个值。
> 例如：
> `nginx_http_log_format = a 'aa' nginx_http_log_format = b 'bb'`
> b将被log_format覆盖。是否有任何解决方案？我需要同时定义log_format a和b


Have you found a solution to this problem？

PR #3079, opened 2017-12-08
feat(core) execute subsequent phases upon short-circuited requests

thibaultcha:
Background context
------------------

Prior to this change, if a plugin short-circuits a request (e.g. auth
plugin with HTTP 401, or rate limiting with HTTP 429) then plugins
running in the NGX_HTTP_LOG_PHASE would not be executed.

Arguably, this is an issue if one wishes to track such requests that are
short-circuited by Kong, i.e. via one of our available log plugins.

The purpose of this patch is to allow other plugins in general to run
after a plugin short-circuited a request in the access phase.

Underlying issue
----------------

Our current plugins' run-loop is implemented in such a way that it both
constructs the list of plugins to execute, and yields to its caller (for
loop) at the same time. Once yielded, a plugin is instantly executed by
the caller (for loop). If the plugin uses the `ngx.exit()` API, then the
execution of the access phase is interrupted, and our run-loop never has
the chance to add logging plugins to the list of plugins to execute for
the current request (that is because logging plugins have a lower
priority than our authentication plugins, which must run first).

Possible solutions
------------------

One could think of several solutions to this issue:

1. Increase the priority of the logging plugins, so that they run
   earlier than auth plugin, and will be added to the list of plugins to
   execute for this request before the access phase is short-circuited.
2. Re-implement our run-loop (`plugins_iterator.lua` and
   `kong/init.lua`) so that it somehow builds the list of plugins to
   execute for a request first, then execute said plugin _after_.
3. Force the run-loop to rebuild the entire list of plugins inside of
   the logging phase.

However, none of these solutions comes without a trade-off.

1. By updating the priority order of each plugin, we run the risk of
   unnecessarily breaking logic depending on the current order of
   execution. We also risk not fixing this issue for custom plugins
   without those plugins also bumping their priority order, which
   could cause cascading issues if other plugins depend on those
   plugins being executed at later phases.
2. While this is definitely a long-term goal, the breaking change
   nature of this solution should tell us that we would rather
   post-pone it until a better case study is made against it.
   Re-implementing our run-loop will benefit Kong in many ways
   (easier to write plugins, more fine-grained hooks, more
   sandboxing...), but doing so now would be disruptive for current
   plugins. One of this reasons behind this is that such a new run-loop
   should probably not follow the same paradigm of building itself and
   yielding at the same time. Instead, we should think of a run-loop
   executing global plugins first, then authentication plugins, then
   API/Consumer-specific plugin. Such an approach as of today would be
[...]

Issue #892, opened 2016-01-19
*-log - 4xx response from Kong are not logged


ahmadnassri:
when Kong responds in a `4xx` _(e.g. `401`, `403`)_ the logging plugins (all `*-log` and Galileo) are not logging those requests.

this might affect runscope as well.

thefosk:
I think this happens when a plugin terminates the request, but not when the usptream service returns a `4**` response.

thibaultCha:
Yes, logging plugins have always been intended to log *successful* requests.

See #490 and https://groups.google.com/forum/#!topic/konglayer/Dx45HAHdf8k for why.

nijikokun:
Perhaps it should be configurable based on the client's needs. Such as per *method*, *status code*, *path*, and *headers*.

Not sure whether we want to get into filtering out certain strings and data that is being passed but I know for a fact that certain logging utilities (raygun) already provide such a service in their agents and libraries.

ahmadnassri:
> See #490 and https://groups.google.com/forum/#!topic/konglayer/Dx45HAHdf8k for why.

we'll have to come up with a possible solution for logging errors in this scenario, a production system would be lacking if those types of errors are not captured in logs.

ryanwilsonperkin:
Hey all, I'd love to weigh in on this since it's a feature we'd be interested in taking advantage of. We're hoping to use the rate-limiting plugin and use Galileo for visualizing of how many requests are being rate limited, and especially being able to break those down by header or path. We'd also love to visualize this for other plugins, such as how many requests are being rejected for authentication reasons.

If there's any other information we can provide, feel free to ask. @ahmadnassri thanks for filing this issue.

nicklanng:
Hi, I've just been working on a PoC of Kong and run into this very issue - has there been any more work done in this area? Thanks.

BrianHutchison:
+1 for letting us control how Kong logs in potential DDOS scenarios.

We would like to enable alerting in our production environment if we are denying requests (past some threshold) based on reaching a rate limit.

@ahmadnassri Have you come up with a work-around?

Tieske:
Code wise it happens when a plugin (or any other piece of Kong) exits prematurely like @thefosk mentioned. To catch those we could try and wrap those ngx functions/methods and catch the errors in between.

[...]

Issue #490, opened 2015-08-19
Kong error response are not getting logged


tamizhgeek:
When there is a error raised by other plugins, for example say ratelimiting, those are not logged in any of the log plugins. The stop_phases check fails and none of them get executed. I think we need to remove this check at least for the 'function _M.exec_plugins_log()'. Or add the log plugins to a special list, which will execute no matter what. Let me know your thoughts. Will be happy to contribute a patch.

thibaultCha:
The logging plugins are for logs of the requests (successfully proxied). Check the Kong error file for kong logging.

thefosk:
@tamizhgeek The Kong error file is the nginx error file, and if you are using Kong in the default working directory is should be located at `/usr/local/kong/logs/error.log`.

tamizhgeek:
But, see the logging plugins are mostly used as a input for the analytics
systems. In analytics we would definitely want to know how much of requests
got ratelimited/quota exceeded. Won't it be good have them logged in the
plugins also?

tamizhgeek:
@thefosk @thibaultCha any more comments on this?

thefosk:
@tamizhgeek the initial reason why the rate-limiting plugins have logging disabled was because effectively if there is a DDOS attack and logging is enabled, that will DDOS also the logging output.

Maybe we could add a boolean flag `enable_log` when adding a rate-limiting plugin?

jeevanrd:
@thefosk Any progress made on this?

sinzone:
@tamizhgeek if all you need is "how much requests, etc" then you should use mashape analytics plugin for that.

thefosk:
@sinzone, @tamizhgeek was asking for a different feature.

ahmadnassri:
this is a valid issue and should not have been closed, now being tracked in https://github.com/Mashape/kong/issues/892

amankumarjain:
How to log request payload and response body using any logging plugin or using some other way, is it allowed to do it?

Tieske:
@amankumarjain please do not reply on closed issues, just accidentally caught this one. Please check the [gitter community channel](https://gitter.im/Mashape/kong) for questions like this.

PR #3533, opened 2018-06-12
feat(runloop) execute plugins on Nginx-produced errors

### LINKED ISSUE #3193: Kong plugins cannot log read and connection timeouts
### Summary

Has been discussed in the forum but since its a bug figured it would be good to make a git issue since that is easier to track long term:
https://discuss.konghq.com/t/kong-http-log-plugin-errors/445/10 

Basically  when it comes to read and connection timeouts Kong can't invoke any of the logging plugins on transactions. 

### Steps To Reproduce

1. Enable any of the logging plugins globally/on an api/on a consumer

2. Create an API, set the upstream read timeout property to 1ms or something crazy to force the timeout.

3. Watch as you will see in proxy error logs the read timeout with nginx, but no plugins at the Kong layer can trap this problem currently, thibaultcha informed us: 

"
kong.log() call will not execute plugins indeed, for another reason: NGINX’s error_page does an internal redirect, which causes ngx.ctx (the value storing the API and plugins to execute for this request) to be reset. Thus, no plugins will be run (Kong even lost the information relating to which API was matched in that particular request at this point).

There seems to be no easy fix for this, we’ll have to add this to our backlog and think of a workaround.
"

### Additional Details & Logs
- Kong version - All versions are impacted, we specifically are running Kong 0.12.1
- Operating System -Alpine 3.6(Kong's docker image). 

Thanks,
-Jeremy

thibaultcha:
Context
-------

Before this patch, several scenarios would make Kong/Nginx produce
errors that did not execute plugins:

1. Nginx-produced client errors (HTTP 4xx) on invalid requests per the
   Nginx settings (e.g. headers too large, URI too large, etc...)
2. Proxy TCP/HTTP errors on the last balancer try (e.g. connection
   refused, connection timeout, read/write timeout, etc...)
3. Any other Nginx-produced server errors (HTTP 5xx), if any

Because plugins are not executed in such cases, logging plugins in
particular do not get a chance to report such errors, leaving Kong
administrator blind when relying on bundled or homegrown logging
plugins. Additionally, other plugins, e.g. transformation, do not get a
chance to run either and could lead to scenarios where those errors
produce headers or response bodies that are undesired by the
administrator.

This issue is similar to that fixed by #3079, which handled the
following case:

4. Requests short-circuited by Kong itself (vs Nginx in 1, 2, and 3.)

See also #490 & #892 for issues related to 4. This patches addresses 1,
2, and 3.

Problem
-------

Nginx-produced errors (4xx and 5xx) are redirected to the
`/kong_error_handler` location by way of the `error_page` directive,
but the runloop is not given a chance to run, since the error handler
is only instructed to override the Nginx-defined HTML responses, and
produces Kong-friendly ones.

In short, a patch solving this issue must enable the runloop within the
error handler.

Proposed solution
-----------------

Several (many) possibilities were evaluated and explored to fix this.
The proposed patch is the result of many attempts, observations, and
adjustments to initial proposals. In a gist, it proposes:

* Executing the response part of the runloop in the error_page location.
* Ensure that if the request part of the runloop did not have a chance
  to run, we do build the list of plugins to execute (thus, global plugins
  only since the request was not processed).
* Ensure that if the request part of the runloop did run, we preserve
  its `ngx.ctx` values.

A few catches:

* Currently, `ngx.req.get_method()` in the `error_page` location block
  will return `GET` even if the request method was different (only
  `HEAD` is preserved). A follow up patch for Nginx is being prepared.
  As such, several tests are currently marked as "Pending".
* When the `rewrite` phase did not get a chance to run, some parts of
  the request may not have been parsed by Nginx (e.g. on HTTP 414), thus
  some variables, such as `$request_uri` or request headers may not have
  been parsed. Code should be written defensively against such
  possibilities, and logging plugins may report incomplete request data
  (but will still report the produced error).
* The `uninitialized_variable_warn` directive must be disabled in the
  `error_page` location block for cases when the `rewrite` phase did not
  get a chance to run (Nginx-produced 4xx clients errors).
* The `ngx.ctx` variables must be copied by reference for plugins to be
  able to access it in the response phases when the request was
  short-circuited by Nginx (e.g. HTTP 502/504 from the upstream module).
* Using a named location for the `error_page` handler is not possible,
  since in some cases the URI might not be parsed by Nginx, and thus Nginx
  refuses to call named locations in such cases (e.g. on HTTP 414).

Resulting behavior
------------------

The `03-plugins_triggering_spec.lua` test suite is now a good reference
for the behavior observed upon Nginx-produced errors. This behavior can
be classified in 2 categories:

1. When the request is not parsed by Nginx (HTTP 4xx)
    a. The `error_page` directive is called (`rewrite` and `access` are
       skipped)
    b. The runloop retrieves the list of **global** plugins to run (it
       cannot retrieve more specific plugins since the request could not be
       parsed)
    c. The error content handler overrides the Nginx body with a Kong-friendly
       one (unchanged from previous behavior)
    d. It executes the `header_filter` phase of each global plugin.
    e. It executes the `body_filter` phase of each global plugin.
    f. It executes the `log` phase of each global plugin.

2. When the error was produced during upstream module execution (HTTP 5xx)
    a. The `rewrite` and `access` phases ran normally (we stashed the
       `ngx.ctx` reference)
    b. The `error_page` directive is called
    c. The error content handler applies the original `ngx.ctx` by
       reference
    d. It does **not** run the plugins loop, since `ngx.ctx` contains
       the list of plugins to run (global **and** specific plugins in
       this case)
    e. The error content handler overrides the Nginx body with a Kong-friendly
       one (unchanged from previous behavior)
    f. It executes the `header_filter` phase of each global plugin.
    g. It executes the `body_filter` phase of each global plugin.
    h. It executes the `log` phase of each global plugin.

Fix #3193

### DIFF
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -112,6 +112,7 @@ server {
 > end
 
     location / {
+        set $ctx_ref                     '';
         set $upstream_host               '';
         set $upstream_upgrade            '';
         set $upstream_connection         '';
@@ -159,8 +160,22 @@ server {
 
     location = /kong_error_handler {
         internal;
+        uninitialized_variable_warn off;
+
         content_by_lua_block {
             kong.handle_error()
+        }
+
+        header_filter_by_lua_block {
+            kong.header_filter()
+        }
+
+        body_filter_by_lua_block {
+            kong.body_filter()
+        }
+
+        log_by_lua_block {
+            kong.log()
         }
     }
 }
--- a/kong-0.13.1-0.rockspec
+++ b/kong-0.13.1-0.rockspec
@@ -60,6 +60,7 @@ build = {
     ["kong.templates.nginx_kong"] = "kong/templates/nginx_kong.lua",
     ["kong.templates.kong_defaults"] = "kong/templates/kong_defaults.lua",
 
+    ["kong.resty.ctx"] = "kong/resty/ctx.lua",
     ["kong.vendor.classic"] = "kong/vendor/classic.lua",
 
     ["kong.cmd"] = "kong/cmd/init.lua",
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -61,6 +61,7 @@ local singletons = require "kong.singletons"
 local DAOFactory = require "kong.dao.factory"
 local kong_cache = require "kong.cache"
 local ngx_balancer = require "ngx.balancer"
+local kong_resty_ctx = require "kong.resty.ctx"
 local plugins_iterator = require "kong.runloop.plugins_iterator"
 local balancer_execute = require("kong.runloop.balancer").execute
 local kong_cluster_events = require "kong.cluster_events"
@@ -383,6 +384,8 @@ function Kong.balancer()
 end
 
 function Kong.rewrite()
+  kong_resty_ctx.stash_ref()
+
   local ctx = ngx.ctx
   runloop.rewrite.before(ctx)
 
@@ -452,6 +455,14 @@ function Kong.log()
 end
 
 function Kong.handle_error()
+  kong_resty_ctx.apply_ref()
+
+  if not ngx.ctx.plugins_for_request then
+    for plugin, plugin_conf in plugins_iterator(singletons.loaded_plugins, true) do
+      -- just build list of plugins
+    end
+  end
+
   return kong_error_handlers(ngx)
 end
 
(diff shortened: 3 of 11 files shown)


kong/templates/nginx_kong.lua
```diff
@@ -147,11 +148,25 @@ server {
         }
     }
 
-    location = /kong_error_handler {
+    location /kong_error_handler {
```
thibaultcha: Need to keep the `=`

PR #3538, opened 2018-06-14
hotfix(cli) read environment from **environ instead of printenv

thibaultcha:
Follow up fix of ac0f3f6fe3048210a68f93ee00d88f10eb77274c

### DIFF
--- a/kong/cmd/utils/env.lua
+++ b/kong/cmd/utils/env.lua
@@ -1,35 +1,46 @@
-local pl_utils = require "pl.utils"
+-- Parts of this file are adapted from the ljsyscall project
+-- The ljsyscall project is licensed under the MIT License,
+-- and copyrighted as:
+--   Copyright (C) 2011-2016 Justin Cormack. All rights reserved.
+
+local ffi = require "ffi"
 local log = require "kong.cmd.utils.log"
-local fmt = string.format
 
 
-local cmd = [[ printenv ]]
+ffi.cdef [[
+  extern char **environ;
+]]
+
+
+local environ = ffi.C.environ
 
 
 local function read_all()
-  log.debug("reading environment variables: %s", cmd)
+  log.debug("reading environment variables")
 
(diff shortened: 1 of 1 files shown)


kong/cmd/utils/env.lua
```diff
@@ -1,35 +1,44 @@
...
 
-  local vars = {}
-  local success, ret_code, stdout, stderr = pl_utils.executeex(cmd)
-  if not success or ret_code ~= 0 then
-    return nil, fmt("could not read environment variables (exit code: %d): %s",
-                    ret_code, stderr)
+  local env = {}
+
+  if not environ then
+    return env
```
thibaultcha: We should add a debug log in this branch

kong/cmd/utils/env.lua
```diff
@@ -1,35 +1,45 @@
...
 
-  local vars = {}
-  local success, ret_code, stdout, stderr = pl_utils.executeex(cmd)
-  if not success or ret_code ~= 0 then
-    return nil, fmt("could not read environment variables (exit code: %d): %s",
-                    ret_code, stderr)
+  local env = {}
+
+  if not environ then
+    log.warn("could not access **environ")
```
hbagdi: This is user visible, the error message should be more comprehensible for a user of any background.

But this can also be a little confusing since only NGINX directive variables (for which key is not known) are read from this code path and other environment variables are looked up from a different code path (`os.getnev`). But if this code path is relied upon in future for all settings then this should be on `error` log level.

thibaultcha: It should be unlikely to fail though, hence why the "low-level" message. Unless you have a possibility in mind?

hbagdi: If you can't read this then there are bigger problems to worry about and so this should be okay.
I can't see why this would fail unless the underlying OS is having trouble with memory management.

PR #3547, opened 2018-06-15
Feat/prometheus

hbagdi:


kong/conf_loader.lua
```diff
@@ -657,6 +657,23 @@ local function load(path, custom_conf)
...
   end
 
+  -- inject an shm for prometheus plugin if needed
+  -- temporary solution until an API to configure shms for plugins is developed
+  if conf.loaded_plugins["prometheus"] then
+    local shm_value = conf["nginx_http_directives"]["lua_shared_dict"]
+
+    if shm_value then
+      if not string.match(shm_value, "prometheus_metrics") then
+        shm_value = shm_value .. "; lua_shared_dict prometheus_metrics 5m"
```
thibaultcha: This is making me regret that the loop injected in the template is not iterating over an array, which would be future proof, would not make this hack necessary, and would yield predictable results if we ever do or allow some sorting of the directives...

thibaultcha:
Merged! Thank you

PR #3548, opened 2018-06-15
fix(schema) catch properly when foreign keys are null

hishamhm:
Thanks to @fffonion for reporting it and to @bungle for the fix!

Includes both a general regression test for the schema library and a regression test specific for the Routes entity.


kong/db/schema/init.lua
```diff
@@ -858,7 +859,7 @@ function Schema:validate_primary_key(pk, ignore_others)
     pk_set[k] = true
     local field = self.fields[k]
     local v = pk[k]
-    if not v then
+    if v == nil or v == null then
```
bungle: The only concern I have with this, is that if we have say `plugins` in new model that can have `null` in `consumer_id`, `route_id`, `service_id`, `api_id`, does this prevent that? Just double checking.

thibaultcha: Likely, but I think we should address the concern when the issue arises, and refactor/add new attributes only then.

PR #3551, opened 2018-06-15
feat(serverless) bundle serverless function plugins

nijikokun:
Bundles serverless function plugins into Kong CE.

**Caveats**

- `pre-function` runs before `zipkin`, it is advisable to use `post-function` when using `zipkin` for accurate metric reporting.
  - If we want to change this behavior, I would advise that we raise the priority on the `zipkin` plugin to `math.huge`, and set the `pre-function` value close nearby to stay consistent with documentation.


nijikokun:
@thibaultcha that is correct, I have made a note of this in the PR description.

thibaultcha:
@nijikokun My commented was made on the commit and pre-dated the PR - hid it as "obsolete".

coopr:
Is there any uncertainty as to "if we want to change this"? I can't guess why we would not want to change this - it seems necessary that Zipkin run before `pre-function` for the reason stated.

nijikokun:
@coopr that's what the purpose of that text is for, to decide such. I'm not sure whether it is necessary. There are two plugins (pre / post), one can run after Zipkin (post), one (pre) explicitly states it runs before all others.

Therefore, if you want to use a function and zipkin, you can use `post`.

coopr:
Ah, thanks for the clarification.

In my opinion, if you are running Zipkin (or, really, anything else that monitors Kong's execution of Kong's code), you'd want it to start running as early as possible. 

Running Zipkin before `pre-function` can't, as far as I understand, affect the execution of the pre-function in any way.

PR #3555, opened 2018-06-17
fix(conf) support adding same directives multiple times

hbagdi:


kong/conf_loader.lua
```diff
@@ -660,19 +661,14 @@ local function load(path, custom_conf)
...
   -- TODO: allow plugins to declare shm dependencies that are automatically
   -- injected
-  if conf.loaded_plugins["prometheus"] then local shm_value =
-    conf["nginx_http_directives"]["lua_shared_dict"]
-
-    if shm_value then
-      if not string.match(shm_value, "prometheus_metrics") then
-        shm_value = shm_value .. "; lua_shared_dict prometheus_metrics 5m"
-      end
+  if conf.loaded_plugins["prometheus"] then local shm_values =
```
hishamhm: style: `then local` in the same line (I realize it's in the existing code, sorry for missing it the last time around)

kong/templates/nginx_kong.lua
```diff
@@ -45,8 +45,10 @@ lua_ssl_verify_depth ${{LUA_SSL_VERIFY_DEPTH}};
 > end
 
 # injected nginx_http_* directives
-> for k, v in pairs(nginx_http_directives)  do
-$(k) $(v);
+> for key, values in pairs(nginx_http_directives)  do
+>   for _, value in pairs(values) do
```
thibaultcha: Better use an array structure (of which we can control its order) and use a simpler loop in the template:
```
> for _, el in ipairs(nginx_http_directives) do
$(el.name) $(el.value);
> end
```

hbagdi: To make sure I understand you correctly, you are asking to change the current layout from:
```lua
local nginx_http_directives = {
  lua_shared_dict = { "custom_cache1 1m", "custom_cache2 1m" }
}
```

to

```lua
local nginx_http_directives = { 
  {
    name = "lua_shared_dict",
    value = "custom_cache1 1m"
  }, 
  {
    name = "lua_shared_dict",
    value = "custom_cache2 1m"
  } 
}
```

thibaultcha: Yes, or any other layout you may think of that results in a simpler change in the template and allows ordering.

hbagdi: IMHO the current implementation makes it easy to sort at two levels (first name, second value) easily but results in a slightly more complicated template.
The proposed approach makes reading all values of the same directive a little more complicated as well.

Willing to make the changes but would like to know why nested loops in the template are being considered complicated?

thibaultcha: > Willing to make the changes but would like to know why nested loops in the template are being considered complicated?

The more Lua mixed in the template, the more complicated to maintain for users? Can we implement the proposed solution? Thanks.

thibaultcha: The table could contain both an array and a hash part, I don't think the goals of simplifying the template, allowing the sort of directives, and the retrieval of values for recurrences of a directive get in the way of each other. Plus the latter can be done in the configuration table, and the former (array of directives) can be built especially for the purpose of compiling the nginx template.

hbagdi: Yes, they are not going to get in each other's way. To keep things small, I'll update the PR to switch the data layout from hash to array of directives.

hbagdi: @thibaultcha I've updated the code with the recent push.

kong/templates/nginx_kong.lua
```diff
@@ -209,8 +209,8 @@ server {
 > end
 
     # injected nginx_admin_* directives
-> for k, v in pairs(nginx_admin_directives)  do
-    $(k) $(v);
+> for _, el in pairs(nginx_admin_directives)  do
```
thibaultcha: We should also switch the iterator to `ipairs` - again, the benefit here is that we can guarantee the order of insertion of each directive across architectures (or even sort it in the future, or guarantee the same order as that of insertion in the configuration file, etc...).

kong/conf_loader.lua
```diff
@@ -665,19 +668,23 @@ local function load(path, custom_conf)
...
-      if not string.match(shm_value, "prometheus_metrics") then
-        shm_value = shm_value .. "; lua_shared_dict prometheus_metrics 5m"
+  if conf.loaded_plugins["prometheus"] then
+    local http_directives = conf["nginx_http_directives"]
+    local found = false
+
+    for _, directive in pairs(http_directives) do
+      if directive.name == "lua_shared_dict"
+         and string.match(directive.value, "prometheus_metrics") then
+         found = true
```
thibaultcha: We can `break` below this line.

kong/conf_loader.lua
```diff
@@ -665,19 +668,23 @@ local function load(path, custom_conf)
...
-    if shm_value then
-      if not string.match(shm_value, "prometheus_metrics") then
-        shm_value = shm_value .. "; lua_shared_dict prometheus_metrics 5m"
+  if conf.loaded_plugins["prometheus"] then
+    local http_directives = conf["nginx_http_directives"]
+    local found = false
+
+    for _, directive in pairs(http_directives) do
+      if directive.name == "lua_shared_dict"
+         and string.match(directive.value, "prometheus_metrics") then
```
thibaultcha: Usually, we prefer `string.find(str, pat, nil, true)` for such use cases.

kong/conf_loader.lua
```diff
@@ -473,7 +473,7 @@ local function parse_nginx_directives(dyn_key_prefix, conf)
           v = string.format("%q", v)
         end
 
-        directives[directive] = v
+        table.insert(directives, { name = directive, value = v })
```
thibaultcha: We could have preserved this behavior to keep the configuration output of the `/` Admin API endpoint in a key/value fashion, and make the array ordering an implementation detail for now, but well.

thibaultcha: Let's ignore this for now.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -93,6 +93,7 @@ local function compile_conf(kong_config, conf_template)
   local compile_env = {
     _escape = ">",
     pairs = pairs,
```
hbagdi: Haven't replaced `pairs` since it doesn't hurt to keep it here.

thibaultcha:
Merged, thanks! Used "refactor" to signify the intended change before the release of this feature.

PR #3558, opened 2018-06-19
fix(db) entity checks on put and cassandra upsert checks

bungle:
### Summary

It fixes this call (maybe something else too):
```
http put :8001/consumers/test
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - [Scheduled](#scheduled)
 - [Released](#released)
+    - [0.14.0rc1](#0140rc1---yyyymmdd)
     - [0.13.1](#0131---20180423)
     - [0.13.0](#0130---20180322)
     - [0.12.3](#0123---20180312)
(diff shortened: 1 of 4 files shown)


hishamhm:
@bungle what is expected to happen with `http put :8001/consumers/foo username=bar` (if `foo` exists and if `foo` doesn't exist)?

bungle:
You cannot change url with `PUT`.

bungle:
You have to use `http put :8001/consumers/<uuid> username=bar`.

hishamhm:
@bungle do we have test cases to verify these two scenarios fail (with consistent error messages in both strategies)? if not, it would be a good idea to add to this PR.

bungle:
If we want to have this included in release then we are running out of time to add anything.

hishamhm:
@bungle I ran the queries locally (with your two PRs applied), and `http put :8001/consumers/foo username=bar` gives a 200 (upserts and ignores `username=bar`) in both Postgres and Cassandra.

Is that the correct behavior or should it instead give an error?

bungle:
@hishamhm yes, put ignores stuff in params so that it stays idempotent.

bungle:
For example:
`PUT /consumers/bob username=jack` <- the username cannot be changed because that would change the the URI as well.

`PATCH /consumers/bob username=jack` <-- should work

`PUT` is always a full replace, `PATCH` is partial, though you can update all parameters.

PR #3564, opened 2018-06-20
Feat/rewrite key auth plugin using kong's PDK

kikito:


kong/plugins/key-auth/handler.lua
```diff
@@ -50,68 +39,50 @@ local function load_consumer(consumer_id, anonymous)
...
+  kong.service.request.set_headers(new_headers)
 end
 
 
 local function do_authentication(conf)
+  local log_err = kong.log.err
   if type(conf.key_names) ~= "table" then
-    ngx.log(ngx.ERR, "[key-auth] no conf.key_names set, aborting plugin execution")
-    return false, {status = 500, message= "Invalid plugin configuration"}
+    log_err("no conf.key_names set, aborting plugin execution")
```
thibaultcha: Why not `kong.log.err` here?

kong/plugins/key-auth/handler.lua
```diff
@@ -193,11 +149,12 @@ local function do_authentication(conf)
...
 
-  local consumer_cache_key = singletons.db.consumers:cache_key(credential.consumer_id)
-  local consumer, err = cache:get(consumer_cache_key, nil, load_consumer,
-                                  credential.consumer_id)
+  local consumer_cache_key = kong.db.consumers:cache_key(credential.consumer_id)
+  local consumer, err      = cache:get(consumer_cache_key, nil, load_consumer,
+                                       credential.consumer_id)
   if err then
-    return responses.send_HTTP_INTERNAL_SERVER_ERROR(err)
+    log_err(err)
```
thibaultcha: It seems like this pattern is repeated in multiple places? Any particular reason why?

kikito:
@thibaultcha thanks for the review. When I did the "un-localization" I missed those spots with `kong.log`. I have unlocalized them too.

To answer to your question: 

> It seems like this pattern is repeated in multiple places? Any particular reason why?

The old `responses.lua` does different things depending on the debug level used. On errors, it logs the errors in the logfile and then returns the 500 to the client. Our methods are less magical: kong.exit "only" exits, and so on. So in order to replicate the old functionality as close as possible, the old calls to `responses.lua` sometimes result in two calls to our pdk (kong.log + kong.exit). That's why the pattern appears several times in the file, I think.

PR #3567, opened 2018-06-21
docs(pdk) replace ldoc tag with custom tag

kikito:


config.ld
```diff
@@ -1,15 +0,0 @@
...
-  './kong/dao/',
-  './kong/tools/',
-  './spec/',
-}
-dir='doc'
---readme='readme.md'
-sort=true
-sort_modules=true
---not_luadoc=true
-all=false
```
thibaultcha: :+1:

PR #3570, opened 2018-06-22
docs(legal) create copyright file

coopr:
This file is a concatenation of all the licenses and copyright notices from all libraries that are used in Kong CE and EE. As such, some notices in this file may not apply to this Kong CE repo, as they cover elements present only in Kong EE.


thibaultcha:
Thanks @coopr, very much appreciated!!

> As such, some notices in this file may not apply to this Kong CE repo, as they cover elements present only in Kong EE.

That is unfortunate, because I do not think those should be included in this copyright file.

coopr:
Roger that @thibaultcha - I'm preparing a CE-only COPYRIGHT file, and will revise this PR soon...

rainest:
Replaced COPYRIGHT with version that only includes CE libraries.

COPYRIGHT
```diff
@@ -0,0 +1,1865 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+%%%%%%%%%
+
+kong-lapis
+https://github.com/leafo/lapis
```
thibaultcha: It seems like we are lacking the license for Lapis (should be MIT)

COPYRIGHT
```diff
@@ -0,0 +1,1865 @@
...
+          pertain to any part of the Derivative Works, in at least one
+          of the following places: within a NOTICE text file distributed
+          as part of the Derivative Works; within the Source form or
+          documentation, if provided along with the Derivative Works; or,
+          within a display generated by the Derivative Works, if and
+          wherever such third-party notices normally appear. The contents
+          of the NOTICE file are for informational purposes only and
+          do not modify the License. You may add Your own attribution
+          notices within Derivative Works that You distribute, alongside
+          or as an ad
```
thibaultcha: We do not ship CE with libnettle

thibaultcha: We do not ship CE with lua-resty-nettle

thibaultcha: We do not ship CE with lua-resty-session afaik?

thibaultcha: I believe we only ship libgmp in EE, along with nettle?

thibaultcha: This should probably be removed

thibaultcha: None of these have their copyright and licence notices :/

hishamhm:
If we're adding runtime dependencies like Nginx/OpenResty/Lua to the list, then it needs to include LuaRocks as well (it's not only a build dependency, as the Kong code does `require("luarocks.loader")`)

thibaultcha:
@hishamhm Good point - LuaRocks was indeed in the list I handed off to @coopr a few days ago; forgot about it here.

COPYRIGHT
coopr: They weren't found with a quick search. Now must occur a slower search...

coopr:
I'm not sure how LuaRocks got dropped from the list - I've added it back, and added https://github.com/luarocks/luarocks/issues/829 to make LuaRocks' licensing more "normal"

COPYRIGHT
coopr: I've removed libraries that are from https://github.com/Kong/

COPYRIGHT
coopr: The lapis repo https://github.com/leafo/lapis does not have a LICENSE file or other statement of license that allows Github's API to return the contents of lapis' license. This COPYRIGHT file is generated programmatically, and such automation depends on each library's license content being retrievable via a URL (eg. pgmoon's license can is available at https://raw.githubusercontent.com/leafo/pgmoon/master/LICENSE)

COPYRIGHT
coopr: will be fixed in the next version of COPYRIGHT

coopr: will be fixed in the next version of COPYRIGHT

coopr: will be fixed in the next version of COPYRIGHT

coopr: will be fixed in the next version of COPYRIGHT

coopr: I'm not sure which lines you are referring to - I think Github's web UI is showing me your comment in the wrong place

COPYRIGHT
```diff
@@ -0,0 +1,1090 @@
...
+
+In addition, the MIME types contained in the Software were
+originally obtained from the Python 2.7.1 ``mimetypes.py`` module,
+though they have been considerably modified and augmented.
+Said file was made available under the Python Software Foundation
+license (http://python.org/psf/license/).
+
+%%%%%%%%%
+
+Open-Resty
```
thibaultcha: Could we spell OpenResty correctly? :/

COPYRIGHT
```diff
@@ -0,0 +1,1090 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
thibaultcha: Our lua-resty-mediador library is a fork from https://github.com/simbiose/mediador of which we are missing the license (MIT)

lpack (on top of which our lua_pack library is implemented) is in the public domain, so nothing to do here I believe.

COPYRIGHT
coopr: That is how the library identifies itself, it seems.

COPYRIGHT
coopr: Sorry, I'm unclear if you are requesting a change or not @thibaultcha thanks.

COPYRIGHT
hishamhm: "OpenResty®", which is how they call it in every single use of the name in their front page :laughing:

thibaultcha: It is registered since last year or so, after all :)

thibaultcha: Where? I could not find any occurence of OpenResty spelled "Open-Resty". This is certainly not how it is generally spelled (in fact, I have never seen it spelled that way, ever). The copyright is in fact attributed to "Yichun "agentzh" Zhang, OpenResty Inc." (no dashes).

COPYRIGHT
thibaultcha: imho, we should have the following copyright for lua-resty-mediador:
```
Copyright (c)
    2015 simbiose
    2017-2018 Kong, Inc.
```

Or in the current form given by the [LICENSE file](https://github.com/Kong/lua-resty-mediador/blob/master/LICENSE):
```
Copyright (c) 2015 simbiose, 2017 Mashape, Inc.
```

(But slightly outdated)

lua_pack seems to be :ok_hand:

thibaultcha: And I believe that would be the last update before this be merged! We can still revisit later if we forgot something.

thibaultcha:
Did not have much time today because of rc2, but will take a final look at this tomorrow!

coopr:
Confirmed, this is in an acceptable state for merging.

PR #3576, opened 2018-06-27
fix(dao) fix error message double wrapping

shashiranjan84:
Lower level strategy module already wraps error message as table and
we should not wrap it again.


thibaultcha:
Manually merged, thanks!

PR #3583, opened 2018-07-02
doc(pdk) add ldoc for pdk ip module

kikito:


kong/pdk/ip.lua
```diff
@@ -1,5 +1,25 @@
...
+-- Both ipv4 and ipv6 are supported.
+--
+-- @function kong.ip.is_trusted
+-- @phases
+-- @tparam string address A string representing an IP address
+-- @treturn boolean `true` if the IP is trusted, `false` otherwise
+-- @usage
+-- if kong.ip.is_trusted("1.1.1.1") then
+--   kong.log("The IP is trusted")
+-- end
```
thibaultcha: I think we also need to edit `kong.pdk.init` to add this module to the list of singletons?

kong/pdk/ip.lua
```diff
@@ -1,5 +1,25 @@
+---
+-- Trusted IPs module
+--
+-- This module allows querying whether an IP is trusted or not for X-Forwarded-* headers.
+-- It depends on the `trusted_ips` option in the Kong config file
```
thibaultcha: "the `trusted_ips` configuration property" is how we generally refer to... configuration properties (at least on the website). Since they do not necessarily come from the config file (env variables and CLI arguments too). Ditto in the below paragraph.

kong/pdk/ip.lua
```diff
@@ -1,5 +1,25 @@
+---
+-- Trusted IPs module
+--
+-- This module allows querying whether an IP is trusted or not for X-Forwarded-* headers.
```
thibaultcha: A more accurate description of this module's duty could be:
```
This module can be used to determine whether or not a given IP address is
in the range of trusted IP addresses defined by the `trusted_ips` configuration
property.
Trusted IP addresses are those that are known to send correct replacement
addresses for clients (as per the chosen header field, e.g. X-Forwarded-*).

See https://docs.konghq.com/latest/configuration/#trusted_ips 
```

kikito:
@thibaultcha I did all the changes you requested, thanks!

PR #3588, opened 2018-07-04
fix(schema) fix validation of nested records in metaschema

hishamhm:
Includes a regression test in the metaschema testsuite, and a testcase in the
schema testsuite (which demonstrates that the nested records issue was a
metaschema-only problem).

Includes some minor corrections to the metaschema specification, detected
using the metaschema to attempt to validate itself... unfortunately the schema
validator doesn't implement a type system strong enough to fully validate the
metaschema without falling into an infinite loop -- in fact, the schema library isn't strong enough to validate schemas without having the metaschema
"cheat" and perform some of its validation of schemas using the
`check` escape-hatch function to recurse into records, as we see in this
commit. This is by design, because the schema library is meant to be only as complex as necessary for handling Kong entities; the additional complexity is tucked away in the metaschema, which remains a non-essential nice-to-have. Still, the metaschema is useful enough to catch syntax errors in schema definitions, so it's good to keep it around.

As a bonus, there's an extra commit that simplifies some code of the schema module.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1194,32 +1194,19 @@ end
 
 --- Cycle-aware table copy.
 -- To be replaced by tablex.deepcopy() when it supports cycles.
-local function copy(t, c, cache)
+local function copy(t, cache)
   if type(t) ~= "table" then
     return t
   end
   cache = cache or {}
-  c = c or {}
+  if cache[t] then
+    return cache[t]
+  end
+  local c = {}
+  cache[t] = c
   for k, v in pairs(t) do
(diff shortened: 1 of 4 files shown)


thibaultcha:
Maybe `refactor` instead of `optimize`?

hishamhm:
@thibaultcha I went with `optimize` cause I'm pretty sure it moves less bits around :)

thibaultcha:
Then `perf` is our scope of choice :) (Actually that's my bad, I meant to write `perf` but my comment mentions `refactor` for some reason).

PR #3594, opened 2018-07-06
chore(dev) auto-convert Windows line-endings

Tieske:
Git clients on Windows tend to convert line endings to Windows
format, such that Windows editors can be used (is reverted on commit).
This doesn't work if mounting those files from a Unix machine for
development for example (Kong-Vagrant). Lua is agnostic to the format,
but the Unix shell is not. Hence only the scripts in `./kong/bin` are
affected.

This change will have the makefile automatically convert those scripts
back to Unix format. This allows Windows users to work on the Kong code
and custom plugins without other configuration.

### Issues resolved

Fix https://github.com/Kong/kong-vagrant/issues/84

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,11 +65,10 @@ pdk-phase-checks:
 
 fix-windows:
 	@for script in $(WIN_SCRIPTS) ; do \
-	  if [ $$(grep -obUaPc '\015' $$script) -gt 0 ] ; then \
-	    echo Converting Windows file $$script ; \
-	    mv $$script $$script.win ; \
-	    tr -d '\015' <$$script.win >$$script ; \
-	    rm $$script.win ; \
-	  fi \
+	  echo Converting Windows file $$script ; \
+	  mv $$script $$script.win ; \
+	  tr -d '\015' <$$script.win >$$script ; \
+	  rm $$script.win ; \
+	  chmod 0755 $$script ; \
 	done;
 
--- a/Makefile
+++ b/Makefile
@@ -11,12 +11,12 @@ else
 OPENSSL_DIR ?= /usr
 endif
 
-.PHONY: install dev lint test test-integration test-plugins test-all win_scripts
+.PHONY: install dev lint test test-integration test-plugins test-all fix-windows
 
-install: win_scripts
+install:
 	@luarocks make OPENSSL_DIR=$(OPENSSL_DIR) CRYPTO_DIR=$(OPENSSL_DIR)
 
-dev: win_scripts
+dev:
 	-@luarocks remove kong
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -61,3 +62,14 @@ pdk-phase-checks:
 	luacov -c t/phase_checks.luacov
 	grep "ngx\\." t/phase_checks.report
 	grep "check_" t/phase_checks.report
+
+win_scripts:
+	@for script in $(WIN_SCRIPTS) ; do \
+	  if [ $$(grep -obUaPc '\015' $$script) -gt 0 ] ; then \
```
hishamhm: grep is notoriously inconsistent across platforms; the "alphabet soup" of those flags is dangerous (especially if is a dependency of all build targets above).

In particular, I just tested it locally and `-P` doesn't work on my machine. Other distros may have other surprises.

How about making this `win_scripts` unconditional (ie, removing the `if [ $$(grep...` line) and not running it as a dependency of every target (but keeping it in the makefile so it's available there for a quick fixup if needed by Windows users)?

hutchic:
I don't love this change ... majority of Windows development tools are configurable

Makefile
Tieske: That's actually a good idea, I'll update it

Makefile
```diff
@@ -61,3 +62,14 @@ pdk-phase-checks:
 	luacov -c t/phase_checks.luacov
 	grep "ngx\\." t/phase_checks.report
 	grep "check_" t/phase_checks.report
+
+fix-windows:
+	@for script in $(WIN_SCRIPTS) ; do \
+	  if [ $$(grep -obUaPc '\015' $$script) -gt 0 ] ; then \
```
hishamhm: The `if grep` is still here, this can go away

hishamhm:
Pushed the necessary tweaks myself, squashed and merged. Thanks!

Tieske:
@hishamhm Ahead of me... I just did some minor updates on the github ui, hadn't had a chance to test it. I should have labelled it as WIP, sorry about that.

Makefile
```diff
@@ -61,3 +62,13 @@ pdk-phase-checks:
...
 	grep "ngx\\." t/phase_checks.report
 	grep "check_" t/phase_checks.report
+
+fix-windows:
+	@for script in $(WIN_SCRIPTS) ; do \
+	  echo Converting Windows file $$script ; \
+	  mv $$script $$script.win ; \
+	  tr -d '\015' <$$script.win >$$script ; \
+	  rm $$script.win ; \
+	  chmod 0755 $$script ; \
```
Tieske: Was this needed btw? I never had an issue without it

hishamhm: Yes. `>` produces a new file without the executable bit set.

PR #3601, opened 2018-07-09
misc(travis) use travis-ci to build / release kong

hutchic:
### Summary

Uses a private github repository that builds / releases kong artifacts via travis-ci

### Testing

- I made / pushed a tag
- The travis deploy stage here ( https://travis-ci.org/Kong/kong/jobs/401935053 ) successfully
- Triggered the kong-distributions build ( https://travis-ci.com/Kong/kong-distributions/builds/78491101 )
- Which I killed because I didn't want it to actually build / publish the artifacts
- Deleted the stray tags

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,14 @@ before_install:
 install:
   - make dev
 
+jobs:
+  include:
+    - stage: deploy
+      if: tag =~ 0.
+      script: .ci/trigger-travis.sh $TRAVIS_ACCESS_TOKEN
+      env:
+        - TEST_SUITE=none
+
 script:
   - .ci/run_tests.sh
 
--- a/.ci/trigger-travis.sh
+++ b/.ci/trigger-travis.sh
@@ -1,0 +1,133 @@
+#!/bin/sh -f
+
+# Trigger a new Travis-CI job.
+# Ordinarily, a new Travis job is triggered when a commit is pushed to a
+# GitHub repository.  The trigger-travis.sh script provides a programmatic
+# way to trigger a new Travis job.
+
+# Usage:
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -50,6 +53,14 @@ install:
 script:
   - .ci/run_tests.sh
 
+jobs:
+  include:
+    - stage: deploy
+      if: tag =~ 0.
```
hishamhm: @hutchic does this mean "only tags starting with `0.`"? If so, that should be good enough for our pre-1.0 tag use, as long as we're aware that those tags have this side-effect :+1:

.travis.yml
```diff
@@ -10,13 +10,16 @@ notifications:
   email: false
 
 addons:
-  postgresql: "9.4"
+  postgresql: "9.5"
```
hishamhm: This is unrelated to the scope of this PR (the `next` branch already sets it to 9.5)

.travis.yml
hutchic: @hishamhm yes this will match `0.*`. We could make this more explicit so we don't accidentally build a release for accidentally pushed tags ie we could do `if: tag =~ ^release/0.` which would only match tags that are `release/0.*` ?

hishamhm: Currently, the branch names follow the `release/` pattern but the release tags themselves are plain version numbers. I don't know if we can check both tag and branch, but if possible I guess that would be ideal.

hutchic: hmm maybe not then :/

I've seen issues where tag name == a branch name

hishamhm:
@hutchic the PR currently contains only the `.ci/trigger-travis.sh` file, looks like the `.travis.yml` changes are gone, please check.

PR #3603, opened 2018-07-10
feat(schema) add support for ttl

bungle:
### Summary

Adds support for entity schemas for `ttl = true` and `ttl = false` which enables or disables `ttl` support for the entitity.

This is first PR in series of PRs that implement the `ttl` support in new model.

### DIFF
--- a/kong/db/schema/metaschema.lua
+++ b/kong/db/schema/metaschema.lua
@@ -132,11 +132,14 @@ local meta_errors = {
   ATTRIBUTE = "field of type '%s' cannot have attribute '%s'",
   REQUIRED = "field of type '%s' must declare '%s'",
   TABLE = "'%s' must be a table",
+  BOOLEAN = "'%s' must be a boolean",
   TYPE = "missing type declaration",
   FIELDS_ARRAY = "each entry in fields must be a sub-table",
   FIELDS_KEY = "each key in fields must be a string",
   ENDPOINT_KEY = "value must be a field name",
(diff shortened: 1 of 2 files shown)


hishamhm:
The PR looks good! A question about the reserved field, since it affects the strategy implementation: would it make sense to make `expires_at` reserved too in case of `ttl = true`, to allow for both kinds of implementations?

hishamhm:
Followup: @bungle explained that the `ttl` name can be used for both relative-time and absolute-time implementations (and it will remain hidden as a strategy-level implementation detail), so no need to reserve another name.

Given that, the PR LGTM!

kong/db/schema/metaschema.lua
```diff
@@ -132,11 +132,14 @@ local meta_errors = {
   ATTRIBUTE = "field of type '%s' cannot have attribute '%s'",
   REQUIRED = "field of type '%s' must declare '%s'",
   TABLE = "'%s' must be a table",
+  BOOLEAN = "'%s' must be a boolean",
   TYPE = "missing type declaration",
   FIELDS_ARRAY = "each entry in fields must be a sub-table",
   FIELDS_KEY = "each key in fields must be a string",
   ENDPOINT_KEY = "value must be a field name",
   ENDPOINT_KEY_UNIQUE = "endpoint key must be a unique field",
+  TTL_RESERVED = "ttl is a reserved field name when ttl is enabled",
```
Tieske: this message seems a bit vague to me?

hishamhm: In context it makes sense. :)

PR #3604, opened 2018-07-10
feat(http-log) add buffering support using a generic facility

hishamhm:
This PR introduces a generic logging buffer module, refactored out of the Galileo-specific logging buffer and uses it in the http-log plugin.

This adds three new optional properties to the configuration of http-log:

* `retry_count` (default value: 10)
* `queue_size` (default value: 1)
* `flush_timeout` (default value: 2)

Their behaviors are the same as the identically-named options in the `galileo` plugin.

Unlike in `galileo`, the default of `queue_size` is set to 1 so that the default behavior of `http-log` remains backwards-compatible.

The output format of `http-log` is dependent on the `queue_size` parameter:

* When `queue_size` is 1, it makes one HTTP request per logged entry, containing a single JSON object in each. This is compatible with the previous releases of the plugin.
* When `queue_size` is greater than 1, the HTTP connection will send a JSON array containing one or more JSON objects in it, up to the configured queue size.

The new fields are not required, and when absent their defaults mean that they behave the same as the http-log plugin always did, meaning this requires no migrations.

This PR also fixes the http-log test suite, which was entirely disabled due to reliance on an unavailable feature of mockbin.com. It adds a mock log server to our custom Nginx configuration of the test suite.

Includes a test case for http-log buffering using the mock log server.


kong-0.14.0-0.rockspec
```diff
@@ -195,6 +195,10 @@ build = {
     ["kong.plugins.oauth2.daos"] = "kong/plugins/oauth2/daos.lua",
     ["kong.plugins.oauth2.api"] = "kong/plugins/oauth2/api.lua",
 
+    ["kong.plugins.log-buffering.json_serializer"] = "kong/plugins/log-buffering/json_serializer.lua",
```
kikito: The naming is at odds with other previous pieces, and is somewhat confusing.

Since the `json_serializer` and `lua_serializer` both are called "serializer", that makes me think that they have something to do with `kong.plugins.log-serializers.basic`. (But then, why the differences in location and names)

Upon looking at their source code, I see that they are completely different. I would recommend finding a different word than "serializer" for them (maybe "producers"?)

kong/plugins/log-buffering/json_serializer.lua
```diff
@@ -0,0 +1,66 @@
...
+-- that already represent JSON-encoded data.
+-- If `encoded` is `false`, entries are assumed to be Lua objects
+-- that need to be encoded during serialization.
+function json_serializer.new(encoded)
+  if encoded ~= nil and type(encoded) ~= "boolean" then
+    error("arg 2 (encoded) must be boolean")
+  end
+
+  local self = {
+    buffer = {},
```
kikito: I find troublesome the fact that the serializers have an attribute called `buffer` but at the same time they are required by a Buffer. I think the Buffer could be renamed to Queue to make it a bit less confusing.

kong-0.14.0-0.rockspec
hishamhm: I reused the naming of the Galileo log buffering module, which used `serializers`. I agree it is confusing with `log-serializers`. Thanks for the suggestion, will switch to "producers".

kong/plugins/log-buffering/json_serializer.lua
hishamhm: the Buffer objects also have a sending queue, so that's not much better. :) Renaming those to `self.output` :)

kong/plugins/http-log/sender.lua
```diff
@@ -0,0 +1,114 @@
...
+local ngx_encode_base64 = ngx.encode_base64
+local ERR = ngx.ERR
+
+
+-- Parse host url.
+-- @param `url` host url
+-- @return `parsed_url` a table with host details like domain name, port, path etc
+local function parse_url(host_url)
+  local parsed_url = url.parse(host_url)
+  if not parsed_url.port then
```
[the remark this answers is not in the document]
hishamhm: Most of this is code pulled straight from the Galileo plugin buffer. The `git blame` of the file was unfortunately lost when the file move and the code changes I made were squashed in a single commit (Git assumes the file was "rewritten").

I will fixup the commit with some cleanups.

hishamhm:
@kikito addressed all concerns, please re-review

PR #3607, opened 2018-07-11
feat(conf) make nginx ssl/tls protocols configurable

romainDavaze:
### Summary

When using https protocol to proxy requests, Kong should provide the
ability to configure which version of ssl/tls are enabled.

This introduces a new variable named SSL_PROTOCOLS to Kong's nginx
template file.
The default supported protocols are TLS1.1 and TLS1.2.


bungle:
@romainDavaze mind you add tests too? Also while you are on it, could you also add this: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_protocols

Then I think we need another PR here:
https://github.com/Kong/docs.konghq.com

I know, it is a lot of effort ;-). Please let us know if you need furher advice.

bungle:
BTW. at some point we wish to make more of the configs dynamic (e.g. that `proxy_ssl_protocols` based on `upstream`), but until then I think we should have these configurable on config level.

thibaultcha:
Thanks @romainDavaze for the contribution. Unfortunately we have other plans for these configuration parameters that translate into Nginx directives. Since 0.14 and the new [injected Nginx directives](https://docs.konghq.com/0.14.x/configuration/#injecting-nginx-directives), this directive you modified should become the appropriate Nginx directives with the appropriate defaults. That is our plan regarding those so we would avoid having to break in the future and be more aligned with our vision regarding Nginx directives.
Maybe @bungle can help you going down that path?

romainDavaze:
Thanks for the quick response @thibaultcha . I guess that's what happens when you don't read the changelog in full details ... My bad !

I just tried the newly added dynamic nginx directives.
In my case, when i set the `KONG_NGINX_PROXY_SSL_PROTOCOLS` variable, it gets well added in the Kong nginx configuration file.
The problem is that the initial directive, that is already there by default, is not replaced. So the configuration file has duplicated entries and it seems like Nginx does not like it when it tries to start, which seems normal (it sometimes crashes when starting for some variables but does not for others).

Maybe I'm doing it wrong or maybe this is something that should be looked at. Anyways, I guess that is not the topic of this PR anymore and since it now seems to be pointeless , maybe i should close it.

thibaultcha:
Yes, as of today this is expected. What I was trying to explain is that the directive that is currently hard-coded into the Nginx configuration file (as well as many other ones) should be removed and replaced by a default provided by Kong, and overridable withba custom kong.conf or other environment variable, as usual with all other configuration values.

thibaultcha:
Closing this as there seems to be no activity. We will backlog this on our side, and I hope the above thread is insightful enough for anyone who may want to implement this feature :)

Thank you

PR #3611, opened 2018-07-11
misc(travis) only run travis on master and next branches

hutchic:
to reduce the travis queue lets only run travis on master and next branches and all PR merge results

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,11 @@ notifications:
 services:
   - redis-server
 
+branches:
+  only:
+  - master
+  - next
+
 addons:
   postgresql: "9.5"
   apt:


bungle:
@hutchic did I understand correctly that if I want to merge branch X to either master or next the tests will still run before merging? If I have PR and want to merge in any other branch than master or next, it won't run tests? Are we on the same page?

hutchic:
@bungle based on the travis docs I believe that to be accurate.

For sure:
the only branches that get tested are next / master
pr's to next / master get tested

Somewhat uncertain:
do pr's to branches not next / master get tested.

bungle:
Yes, and this PR I think is a good case, We only have one:
> continuous-integration/travis-ci/pr — The Travis CI build passed

Instead of normal two.

hishamhm:
Merged with the scope changed to `chore(travis)`. Thanks!

PR #3614, opened 2018-07-11
feat(db) add ttl support to postgres strategy

bungle:
### Summary

Implements Postgres strategy support for TTL (schema feature `ttl = true`).

See also: #3603, #3620

### DIFF
--- a/kong/db/strategies/postgres/init.lua
+++ b/kong/db/strategies/postgres/init.lua
@@ -14,9 +14,11 @@ local ipairs        = ipairs
 local pairs         = pairs
 local error         = error
 local upper         = string.upper
+local type          = type
 local null          = ngx.null
 local load          = load
 local find          = string.find
+local time          = ngx.time
 local rep           = string.rep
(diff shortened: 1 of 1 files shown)


kong/db/strategies/postgres/init.lua
```diff
@@ -27,6 +28,8 @@ local log           = ngx.log
 local NOTICE        = ngx.NOTICE
 local LIMIT         = {}
 local UNIQUE        = {}
+local CREATED_AT    = "created_at"
+local TTL           = "ttl"
```
hishamhm: minor: I believe we're moving away from the pattern of making variables for string constants (because if the constant is changed the variable name would have to be changed as well for consistency, so these don't gain us much).

hishamhm:
@bungle Sorry I couldn't grasp from the code at first, so I thought I would just ask: is the `ttl` field being returned by the `select*` functions or is it being kept fully internal?

bungle:
@hishamhm it is kept internal.

PR #3618, opened 2018-07-13
fix(ldap-auth) remove bin.mod and negative cache

shashiranjan84:
This PP fixes two issues with plugin:
- One where request added negative cache for incorrect password.
- Second when request results in 500 when very long password being used. Now `%` used instead of `bin.mod` as `bin.mod` does not exist


hishamhm:
@shashiranjan84 changes and regression tests look good! CI failure was unrelated, I restarted it. Will merge when we get the green light from Travis!

PR #3622, opened 2018-07-16
Feat/targets to new dao

kikito:
The new admin API is fairly similar to the previously existing one, but there are some *breaking changes*:

* Lists (`GET /upstreams` or `GET /upstreams/{upstream}/targets`) don't accept query-attributes for filtering any more
* Pagination: the `total` attribute is gone
* Format of error messages has changed to new DAO errors.
* `PUT` actions now are applied to entities instead of to collections. (`PUT /upstreams/foo` instead of `PUT /upstreams`
* Some error codes have changed:
  * Attempting to use an invalid name name (such an IP address) when updating an upstream has become a `500` error instead of a `400`.
  * `PUT` actions return a `200` on creation instead of `201`.
* timestamps (like `created_at`) are floating point numbers now (in all entities - not just upstreams and services).

*New features*:

* There is a attribute called `next` on paginated admin API results.
* When listing the targets for an upstream with `db.targets:for_upstream(u)`, the list only includes active targets, and is reverse-sorted by `created_at+id`. This is the method used in `/upstreams/{upstream}/targets`
* There is also a `db.targets:for_upstreams_raw(u)` which returns all targets, including those with `weight=0`, associated with the upstream. That's what used in `/upstreams/{upstream}/targets/all`
* `db.targets:for_upstream_with_health(u)` is the same list as `db.targets:for_upstream(u)` but also includes health information (so it is slower to call). It is used in `/upstreams/{upstream}/health`.
* In addition to those there's `db.targets:for_upstream_sorted(u)` which returns the targets sorted by `created_at+id`. It is used internally in a couple places for things like cleaning up and building the cache.
* The `db.targets:delete` method creates a new target with `weight = 0`. The only way to really delete a target from the database is by inserting targets (this will eventually trigger a `clean_history` call).
* The the `db.targets:post_health` is the one-place for changing health status of a target. It deals with balancer changes and sending events to the cluster. It is used in `POST /upstreams/{upstream}/targets/{target}/healthy|unhealthy`.

*Prerequisites*:

* Cassandra migrations where not re-entrant (they can not "catch exceptions and continue" like postgres can). In the past the tests where run only in postgresql. I had to add a way to "ignore errors" in order to allow re-entrant migrations in Cassandra.
* There was a lurking bug related with partial updates of record-type fields.
* Needed a new kind of validator to allow validating the `hash_on_header` is not equal to `hash_fallback_header`
* The `target.target` attribute is an endpoint key but it is not unique. Had to remove that restriction from the system.
* Timestamps should have millisecond-precision, but until we updated the targets to the new DAO we didn't realize that we where storing them with second-precision in the database.


hishamhm:
@kikito please remove from this PR the commits that were already merged via other PRs (I'm not doing it myself to avoid getting your local branch out of sync)

kikito:
@hishamhm done - I have left the timestamps-as-float commit because it is needed for the tests to pass.

PR #3623, opened 2018-07-16
feat(db) call post_process for delete too

nitzanharel:
Having the ability to make a callback on delete of an entity is
important for things like invalidating applicative caches.

### DIFF
--- a/kong/api/crud_helpers.lua
+++ b/kong/api/crud_helpers.lua
@@ -265,7 +265,7 @@ end
 
 --- Delete an entity.
 -- The DAO requires to be given a table containing the full primary key of the entity
-function _M.delete(primary_keys, dao_collection)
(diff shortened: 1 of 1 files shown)


thibaultcha:
@nitzanharel Is your entity implemented with the DAO layer provided by Kong? Because if so, a callback is already available and used for cache invalidation purposes. It is much more reliable and powerful than having it at the admin API level. The post processing of rows is mostly used for transformations when retrieving values from the database as far as I remember, or some occasional anonymous events reporting; it isn’t quite the same use case.

nitzanharel:
@thibaultcha - thanks for the feedback. What is this callback you are referring to and how/when/where is it called?

thibaultcha:
Closing this then, thank you!

PR #3628, opened 2018-07-17
fix(tests) fix flaky tests

kidd:
This PR cleans 2 flaky tests.

* join an opened thread to clean test environment
* Allow 1 extra ms more for a test to pass (probably an split ms
  issue)
* More verbose logging in case of failure starting Kong

### DIFF
--- a/spec-old-api/03-plugins/08-datadog/01-log_spec.lua
+++ b/spec-old-api/03-plugins/08-datadog/01-log_spec.lua
@@ -223,7 +223,7 @@ describe("Plugin: datadog (log)", function()
   end)
 
   it("should not return a runtime error (regression)", function()
-    helpers.udp_server(9999, 1, 1)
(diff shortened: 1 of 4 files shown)


hishamhm:
Squashed and merged adjusting the commit scope/message. Thank you, @kidd!!

spec-old-api/03-plugins/08-datadog/01-log_spec.lua
```diff
@@ -237,5 +237,6 @@ describe("Plugin: datadog (log)", function()
 
     local err_log = pl_file.read(helpers.test_conf.nginx_err_logs)
     assert.not_matches("attempt to index field 'api' (a nil value)", err_log, nil, true)
+    thread:join()
```
Tieske: this should be wrapped in a `finally` call, since currently it won't get executed if the assertion fails.

PR #3630, opened 2018-07-18
feat(schema) add support for subschemas

hishamhm:
This resolves three difficulties that would appear when porting the Plugins entity over to the new DAO: loading the plugin configuration schema dynamically, the fact that this schema depends on a value of the entity row itself, and the fact that the specific plugin type can impose further constraints on values.

The old DAO implemented this by allowing records to have a `schema` field that would be either a table or a function evaluated at validation time, and using an ad-hoc validation that checked for a `no_consumer` field to restrict the value of `consumer_id` to `null`.

The new DAO implements this by allowing an entity (such as Plugins) to declare that it is extensible via **subschemas**:

* a schema may declare that it allows subschemas, tagged by one specific string key of the schema, with the top-level `subschema_key` property.
  * essentially, the set of all plugins taken as one works like a tagged union.
  * for Plugins, `subschema_key` will be `"name"`
* a schema may declare where it gets subschemas from, with the `subschema_module`.
  * for Plugins, `subschema_module` will be `"kong.plugins.?.schema"`
* the parent schema may declare fields as "abstract", which are meant to be specialized by the subschema. Abstract records don't need their `fields` property to be set, and they must be given in the specialized ones.
  * for Plugins, `config` will be `abstract`
* the set of fields considered when validating an entity is that of the parent schema
  * subschemas cannot introduce new fields, those need to be declared in the parent schema ("abstract" if needed)
  * the data model of strategies can be inferred from parent schema alone
* subschemas may override both abstract and non-abstract fields.
  * they cannot change the type of a field
  * a field declared in the subschema overrides the parent schema's field completely (constraints are not merged)
  * for Plugins, plugins will override `config`, and for the `no_consumer` feature, they can override `consumer` using the `eq = null` validator

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -71,12 +71,17 @@ local validation_errors = {
   -- schema error
   SCHEMA_NO_DEFINITION      = "expected a definition table",
   SCHEMA_NO_FIELDS          = "error in schema definition: no 'fields' table",
-  SCHEMA_MISSING_ATTRIBUTE  = "error in schema definition: missing attribute",
+  SCHEMA_MISSING_ATTRIBUTE  = "error in schema definition: missing attribute %s",
   SCHEMA_BAD_REFERENCE      = "schema refers to an invalid foreign entity: %s",
   SCHEMA_TYPE               = "invalid type: %s",
   -- primary key errors
   NOT_PK                    = "not a primary key",
   MISSING_PK                = "missing primary key",
+  -- subschemas
+  SUBSCHEMA_UNKNOWN         = "unknown type: %s",
+  SUBSCHEMA_BAD_PARENT      = "entities of type '%s' cannot have subschemas",
+  SUBSCHEMA_UNDEFINED_FIELD = "error in schema definition: abstract field was not specialized",
+  SUBSCHEMA_BAD_TYPE        = "error in schema definition: cannot change type in a specialized field",
 }
 
 
@@ -468,6 +473,10 @@ function Schema:validate_field(field, value)
     end
   end
 
+  if field.abstract == true then
+    return nil, validation_errors.SUBSCHEMA_UNDEFINED_FIELD
+  end
+
   if field.type == "array" then
     if not is_sequence(value) then
       return nil, validation_errors.ARRAY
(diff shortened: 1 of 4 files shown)


kikito:
I am not sure the parent needs to know where his subschemas are: the  `"kong.plugins.?.schema"` part is not needed as far as I know - as long as we "keep track of the list of subschemas" when a new subschema is found.

Similarly, I don't think the `abstract` field abstraction is needed. It seems to me that the only reason to have them is to raise an error if someone wants to instantiate a Plugin without the correct subschema. But in that case we can still raise an error by detecting that the `subschema_key` doesn't correspond with any of the loaded subschemas.

hishamhm:
> I am not sure the parent needs to know where his subschemas are: the "kong.plugins.?.schema" part is not needed as far as I know - as long as we "keep track of the list of subschemas" when a new subschema is found.

We need that otherwise the knowledge of this path ends up hardcoded somewhere. This is already a problem with core entities because foreign relationships assume that other entities are core entities: https://github.com/Kong/kong/blob/master/kong/db/schema/init.lua#L1182-L1183 (@james-callahan has had problems with that line already). I don't want to repeat this. For foreign-key relationships we could have made it such that the `reference` field needs a full module path, but for subschemas we don't have this luxury because the "name" field of plugins needs to be backwards-compatible.

Unless you are proposing that we load every plugin beforehand, including their config schemas and eventual custom DAOs, and store them all in the same flat/single namespace. I think keeping these namespaces separate is less prone to conflict.

For example, one possible (far off) future application of subschemas would be in a redesign of the Upstream entity into a more general Balancer entity where instead of a bunch of fields for hashing method, where we need to add fields each time we add a new method (e.g. `hash_on_cookie`), we could have a `Balancer.config` field and implement `kong.runloop.balancer.methods.?.schema` without worrying if there's another plugin called `cookie` out there when you create a `cookie` hashing method. (Of course one could fall into the age-old "but then let's create a convention that every X is prefixed by Y" which is always true for single-namespace propositions, but namespaces exist for a reason. :) )

> Similarly, I don't think the abstract field abstraction is needed. It seems to me that the only reason to have them is to raise an error if someone wants to instantiate a Plugin without the correct subschema. But in that case we can still raise an error by detecting that the subschema_key doesn't correspond with any of the loaded subschemas.

The reason it is there is so that the parent schema alone can used to produce the DDL stuff for creating the database table in the strategy, without knowledge of the concrete ones, and for forcing certain fields to be specialized (`config`, in Plugins case).

hishamhm:
@kikito I pushed a fixup commit (to be squashed when merged), based on the feedback you gave me today. It removes the `subschema_module` property, making the (pre-)loading of subschemas a separate concern from the schema library itself.

I think it reaches a nice middle-ground between the various approaches we suggested. Let me know what you think!

PR #3638, opened 2018-07-20
feat(db) implement ttl support for kong.db and new admin api

bungle:
### Summary

* Implements TTL support in Cassandra strategy
* Implements TTL support in Postgres strategy
* Implements TTL support in Generic DAO
* Implements TTL support in Admin API Endpoints
* Implements TTL cleanup timer for Postgres strategy
* Renames `kong.db.init_connector` to `kong.db.init`

### DIFF
--- a/kong/db/errors.lua
+++ b/kong/db/errors.lua
@@ -3,10 +3,12 @@ local pl_keys = require("pl.tablex").keys
 
 
 local type         = type
+local null         = ngx.null
 local error        = error
 local upper        = string.upper
 local fmt          = string.format
 local pairs        = pairs
+local ipairs       = ipairs
 local tostring     = tostring
 local setmetatable = setmetatable
 local concat       = table.concat
@@ -26,12 +28,15 @@ end
 local ERRORS            = {
   INVALID_PRIMARY_KEY   =  1,
   SCHEMA_VIOLATION      =  2,
-  PRIMARY_KEY_VIOLATION =  3, -- primary key already exists (HTTP 400)
-  FOREIGN_KEY_VIOLATION =  4, -- foreign entity does not exist (HTTP 400)
-  UNIQUE_VIOLATION      =  5, -- unique key already exists (HTTP 409)
-  NOT_FOUND             =  6, -- WHERE clause leads nowhere (HTTP 404)
-  INVALID_OFFSET        =  7, -- page(size, offset) is invalid
-  DATABASE_ERROR        =  8, -- connection refused or DB error (HTTP 500)
+  PRIMARY_KEY_VIOLATION =  3,  -- primary key already exists (HTTP 400)
+  FOREIGN_KEY_VIOLATION =  4,  -- foreign entity does not exist (HTTP 400)
+  UNIQUE_VIOLATION      =  5,  -- unique key already exists (HTTP 409)
+  NOT_FOUND             =  6,  -- WHERE clause leads nowhere (HTTP 404)
+  INVALID_OFFSET        =  7,  -- page(size, offset) is invalid
+  DATABASE_ERROR        =  8,  -- connection refused or DB error (HTTP 500)
+  INVALID_SIZE          =  9,  -- page(size, offset) is invalid
+  INVALID_UNIQUE        =  10, -- unique field value is invalid
+  INVALID_OPTIONS       =  11, -- invalid options given
 }
 
 
@@ -47,6 +52,9 @@ local ERRORS_NAMES               = {
   [ERRORS.NOT_FOUND]             = "not found",
   [ERRORS.INVALID_OFFSET]        = "invalid offset",
   [ERRORS.DATABASE_ERROR]        = "database error",
+  [ERRORS.INVALID_SIZE]          = "invalid size",
+  [ERRORS.INVALID_UNIQUE]        = "invalid unique %s",
+  [ERRORS.INVALID_OPTIONS]       = "invalid options",
 }
 
 
@@ -56,8 +64,8 @@ local ERRORS_NAMES               = {
 local _err_mt = {
   __tostring = function(err_t)
     local message = err_t.message
-    if message == nil or message == ngx.null then
-      message = err_t.name
+    if message == nil or message == null then
+       message = err_t.name
     end
 
     return fmt("[%s] %s", err_t.strategy, message)
@@ -78,7 +86,7 @@ local _M = {
 }
 
 
-local function new_err_t(self, code, message, errors)
+local function new_err_t(self, code, message, errors, name)
   if not code then
     error("missing code")
   end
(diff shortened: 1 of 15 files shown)


kong/db/strategies/postgres/init.lua
```diff
@@ -483,22 +479,60 @@ local function execute(strategy, statement_name, attributes, is_update)
...
+             fields.created_at.auto then
+        ttl_value = escape_literal(connector, ttl_value + attributes.created_at, fields.ttl)
 
-    if i == argc and is_update and attributes[UNIQUE] then
-      value = attributes[UNIQUE]
+      elseif update and attributes.updated_at and
+             fields.updated_at                and
+             fields.updated_at.timestamp      and
+             fields.updated_at.auto then
+        ttl_value = escape_literal(connector, ttl_value + attributes.updated_at, fields.ttl)
```
thibaultcha: Why is the TTL value derived from `updated_at`? Shouldn't a TTL always be derived form the creation date?

kong/db/strategies/postgres/init.lua
```diff
@@ -483,22 +479,60 @@ local function execute(strategy, statement_name, attributes, is_update)
...
+
+      else
```
thibaultcha: Should this be `elseif not update then`? Not sure, but we'd rather avoid touching the TTL if we are updating a row without editing its `created_at` field, right? Unless I missed something that already prevents that, or that makes this behaviour desirable.

kong/db/strategies/postgres/init.lua
```diff
@@ -483,22 +479,60 @@ local function execute(strategy, statement_name, attributes, is_update)
 
   clear_tab(argv)
 
-  for i = 1, argc do
-    local name  = argn[i]
-    local value
+  local update = options and options.update
```
thibaultcha: my 2cts: `is_update` (which you previously came up with) was more readable :sweat_smile:

kong/db/strategies/postgres/init.lua
bungle: If you want to patch entity a new ttl we use updated_at If present. You can also remove ttl (ttl=0) with patch.

kong/db/strategies/postgres/init.lua
bungle: We won't touch ttl If not requested on updates.

kong/db/schema/metaschema.lua
```diff
@@ -141,7 +141,6 @@ local meta_errors = {
   FIELDS_KEY = "each key in fields must be a string",
   ENDPOINT_KEY = "value must be a field name",
   TTL_RESERVED = "ttl is a reserved field name when ttl is enabled",
-  TTL_CREATED_AT = "ttl can only be enabled on entities that have a 'created_at' timestamp field",
```
thibaultcha: Why are we removing this constraint?

thibaultcha:
Had some more concerns but pretty minor at this point, we can practically merge this as-is imho. Left a few notes for you though @bungle

kong/db/schema/metaschema.lua
bungle: At first I thought I need that when I created PR for `ttl=true` for the schema. But when I implemented Postgres strategy, I kinda saw it as optional. PG strategy will still count TTL if there is `created_at` and on updates `updated_at`, but there is also a fallback for entities that have `ttl=true`, but no `created_at`, in which case PG ttl is counted like `ngx.time() + ttl`. Cassandra never really needed or used `created_at` for `ttl` as it just accepts the number as is. With PG there is no native support as of now, so we decided to store `expires_at` (absolute time) to `ttl` field, as it makes it easier to cleanup, and exclude expired rows.

kong/db/init.lua
```diff
@@ -111,7 +111,7 @@ local function prefix_err(self, err)
 end
 
 
-function DB:init_connector()
+function DB:init()
```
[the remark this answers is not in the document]
bungle: Ok, I will remove:
https://github.com/Kong/kong/pull/3638/commits/83a556ee27f000b10500107644b1b5a80497703f

The original intent was that we have these:
1. `kong.db.connector:init()`
2. `kong.db.init()` (it calls also the connector:init() — meant to be used in init phase)
3. `kong.db.init_worker()` — (it calls also the connector:init_worker() — meant to be used in init_worker phase)

Just to make kong.db follow similar patterns as we use elsewhere (and in old DAO). And I though it was nicer to think that `kong.db.init` was not called than `kong.db.init_connector` was not called. And as we already had direct way to init connector via `kong.db.connector:init()` (same number of chars as with `init_connector`), I though well lets add init phase handler to db too. You never know if we only need to init connector or do other stuff as well, and if yes, later on we can find ourselves doing this:

```lua
local db = kong.db.new()
db:init_connector()
db:init()
```

vs. just:

```lua
local db = kong.db.new()
db:init()
```

Though even that has been argued that should the `.new` do it all, or perhaps have different constructors or parameters to `new` that also does initialization.

So the `init_worker` was missing, but we had:
1. `kong.db.connector:init()`
2. `kong.db.init_connector()`

Those are pretty much equal. There is only a bit of error handling in `init_connector`.

Then I needed init_worker in connector. Should I add:
1. `kong.db.init_worker_connector` OR
2. `kong.db.init_connector_worker` OR
3. `kong.db.init_worker` 

But it looks like what you sugges is to just have:

`kong.db.init_connector`
`kong.db.init_worker`

So I will be making those changes.

kong/db/errors.lua
```diff
@@ -317,4 +328,60 @@ function _M:database_error(err)
...
+  local buf = {}
+  local len = 0
+
+  for _, option_name in ipairs(sorted_keys(errors)) do
+    local option_errors = errors[option_name]
+    if type(option_errors) == "table" then
+      for _, sub_option in ipairs(sorted_keys(option_errors)) do
+        len = len + 1
+        buf[len] = fmt("%s.%s: %s", option_name, sub_option,
+          option_errors[sub_option])
```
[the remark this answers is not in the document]
bungle: Yes, I'll try to correct that and somehow setup the IDE to do that for me.

kong/db/strategies/postgres/init.lua
```diff
@@ -483,22 +479,62 @@ local function execute(strategy, statement_name, attributes, is_update)
...
+             attributes.updated_at and
+             fields.updated_at and
+             fields.updated_at.timestamp and
+             fields.updated_at.auto then
+        ttl_value = escape_literal(connector, ttl_value + attributes.updated_at, fields.ttl)
 
-    if i == argc and is_update and attributes[UNIQUE] then
-      value = attributes[UNIQUE]
+      else
+        ttl_value = escape_literal(connector, ttl_value + time(), fields.ttl)
```
thibaultcha: It seems to me like it would be fine to just have this branch? The only purpose the above branches seem to serve is to not repeat the `ngx.time()` call, and use `created_at`, or `updated_at` (which were generated by `ngx.time()` since they have the `auto` attribute). That would simplify the implementation quite a bit, unless those branches serve another purpose that I missed, of course.

PR #2777, opened 2017-08-04
feat(aws-lambda) Lambda authentication via EC2 IAM roles

Mgutjahr:
### Summary
Adds support for querying temporary credentials from EC2 metadata service instead of fetching long-lived key-pairs from the database.

The credentials are fetched from the EC2 metadata service and are cached for 60 seconds. I assume that many users are running kong on EC2 instances. For these users this is the recommended way to authenticate requests for AWS services as key-rotation and credential provisioning are managed by AWS.

For further details refer to the [AWS documentation about EC2 roles](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html)

### Full changelog

* Added `kong.plugins.aws-lambda.iam-role-credentials` for  fetching and caching temporary credentials from the metadata service (default 169.254.169.254 port 80) 
* Added plugins config flag `use_ec2_iam_role` to `kong.plugins.aws-lambda.schema` and declared `aws_key` and `aws_secret` as optional
* Added `self_check` function to check whether either `use_ec2_iam_role` is set  to true or both aws_key and aws_secret as configured
* Added if branch in `kong.plugins.aws-lambda.handler` to either fetch credentials from the metadata service or use credentials stored in database

PR #3639, opened 2018-07-23
Iam roles with proxy

rkbasa:

### Summary

This PR is an extension to the PR #2777 raised by @Mgutjahr. Extended the plugin to enabled proxy option for enterprises where requests are routed via enterprise proxy.

### Full changelog

Added plugins config flag proxy_url and proxy_scheme to kong.plugins.aws-lambda.schema
Added self_check function to check whether proxy_url is set, if yes, check whether proxy_scheme is set.
Updated kong.plugins.v4.lua to use openssl library instead of crpto
Updated kong.plugins.iam-role-credentials.lua to make 169.254.169.254 to go via proxy only. 
Added check for proxy url. if set, http connect uses connect_proxy else uses normal connect. Also changed the upstream_body attributes to match/align that of aws lambda implementations i.e upstream_body.httpMethod, upstream_body.headers, upstream_body.path, upstream_body.queryStringParameters and upstream_body.body

### DIFF
--- a/kong-0.11.0-0.rockspec
+++ b/kong-0.11.0-0.rockspec
@@ -275,6 +275,7 @@ build = {
     ["kong.plugins.aws-lambda.handler"] = "kong/plugins/aws-lambda/handler.lua",
     ["kong.plugins.aws-lambda.schema"] = "kong/plugins/aws-lambda/schema.lua",
     ["kong.plugins.aws-lambda.v4"] = "kong/plugins/aws-lambda/v4.lua",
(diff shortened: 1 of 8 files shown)


thibaultcha:
Hi @rkbasa,

Thank you for your contribution! We will need some time to review it as we are sorting out a few other priorities right now, but I just wanted to drop you a note that it isn't being ignored.

Cheers

Tieske:
closing this, has been added to the aws-lambda plugin after it was spun out into its own git repo.

PR #3646, opened 2018-07-26
Allow plugins to opt out of per-api/route/service

james-callahan:
### Summary

Allow plugins to opt out of getting enabled per-api/route/service
This allows a plugin author to e.g. force that their plugin only runs globally

### Full changelog

* Add  support for `.no_api`, `.no_route`, and `.no_service`


kong/runloop/plugins_iterator.lua
```diff
@@ -97,17 +97,24 @@ local function get_next(self)
...
+      api = nil
+    end
+    if route and schema.no_route then
+      route = nil
+    end
+    if service and schema.no_service then
+      service = nil
+    end
+    if consumer and schema.no_consumer then
+      consumer = nil
```
Tieske: I see that the original code had the same checks and balances, but isn't this a migration problem? Those cases should not exist, so if they do, something else failed.

If the code stays, it should generate warnings maybe?

@thibaultcha what do you think?

thibaultcha: > Those cases should not exist

Which cases are you referring to?

Tieske: My thought was: if a plugin schema has `no_route == true` then we should not need to set `route = nil` since it already should be `nil`.

rereading the code it seems that `route` holds the current request-route, not the configured one for the plugin.

Sorry for the noise.

coopr:
Does this have a matching docs PR?

thibaultcha:
@coopr For now, this is needed to unblock @james-callahan's work. If anybody has the time to come up with the docs PR, that'd be great! If not, we'll come to it when the release changelog will be drafed and docs updated scheduled as part of the release checklist.

PR #3649, opened 2018-07-26
feat: configurable phase handlers

p0pr0ck5:
This commit provides a Kong config option to enable/disable
various phase handlers. This behavior can result in substantial
performance gains when use to ignore certain phase handlers in which
no substantial work is done by Kong, but various core behavior such
as the plugins iterator result in CPU cycles spent on a nop behavior.

* Implement a loop generator to write configured phase handlers into
  the generated Nginx config
* Test relevant conf_loader behaviors

### DIFF
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -126,14 +126,12 @@ server {
         set $upstream_x_forwarded_host   '';
         set $upstream_x_forwarded_port   '';
 
-        rewrite_by_lua_block {
-            Kong.rewrite()
+> for _, phase in ipairs(phase_handlers) do
+        $(phase)_by_lua_block {
+            Kong.$(phase)()
         }
 
-        access_by_lua_block {
-            Kong.access()
-        }
-
+> end
         proxy_http_version 1.1;
         proxy_set_header   Host              $upstream_host;
         proxy_set_header   Upgrade           $upstream_upgrade;
@@ -147,18 +145,6 @@ server {
         proxy_pass_header  Date;
         proxy_ssl_name     $upstream_host;
         proxy_pass         $upstream_scheme://kong_upstream$upstream_uri;
-
-        header_filter_by_lua_block {
-            Kong.header_filter()
-        }
-
-        body_filter_by_lua_block {
-            Kong.body_filter()
-        }
-
-        log_by_lua_block {
-            Kong.log()
-        }
     }
 
     location = /kong_error_handler {
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -35,6 +35,14 @@ local DYNAMIC_KEY_PREFIXES = {
   ["nginx_admin_directives"] = "nginx_admin_",
 }
 
+local KONG_PHASE_HANDLERS = {
+  "rewrite",
+  "access",
+  "header_filter",
+  "body_filter",
+  "log",
+}
+
 local PREFIX_PATHS = {
   nginx_pid = {"pids", "nginx.pid"},
   nginx_err_logs = {"logs", "error.log"},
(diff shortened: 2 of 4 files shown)


thibaultcha:
This sounds like giving a (rather large) gun to users for them to shoot themselves with. There are no safeguards,  Also, while the benefits are clear, this optimization seems to be so low-level from a user's POV that it doesn't seem like it should be up to them to tweak it, but rather up to the runloop (and the runloop is scheduled for a refactor, this is not news). Finally, I would expect a large number of setups to barely be able to tweak this, considering a "regular" setup with a few plugins is likely going to use all phases from access to log...

I am having a hard time justifying this change.

kong/templates/nginx_kong.lua
```diff
@@ -147,18 +145,6 @@ server {
...
-            Kong.header_filter()
-        }
-
-        body_filter_by_lua_block {
-            Kong.body_filter()
-        }
-
-        log_by_lua_block {
-            Kong.log()
-        }
```
thibaultcha: fwiw, the same phases are also defined below under the `/kong_error_handler` location.

p0pr0ck5:
@thibaultcha the biggest push here is to be able to cleanly remove the `rewrite` phase handler (and its associated call to the plugins iterator); manually removing this phase handler in the Nginx config shows a ~15% increase in throughput. I assumed that a flag to simply remove this handler wasn't a good idea, so this was a first crack to start a conversation. It's an admittedly blunt hammer. :) Let me know if you want to see data/flamegraphs on the tests we've run. Apart from this, what recommendations do you have regarding this, apart from running with a custom Nginx config?

coopr:
@james-callahan I'm curious if you foresee this offering notable performance benefits in typical Kong Service Mesh deployments?

p0pr0ck5:
Sorry, one other piece of context here for everyone that I forgot to mention (cc @coopr and @james-callahan in addition to @thibaultcha)- _none_ of the plugins that ship with Kong CE uses this handler, and the before/after meta handlers are also essential nops (timing stats unused elsewhere). This call burns a nontrivial amount of CPU time for 0 gain.

thibaultcha:
@p0pr0ck5 My recommendation - and strong opinion - would be to avoid introducing such low level settings users will become entitled to, and wait the proper refactor of the runloop.

thibaultcha:
As an alternative, we could take advantage of the fact that the list of enabled plugins is static (plugins that will potentially be configured, i.e. the `plugins` configuration property) and will only change when Kong restart. From there, we could loop through the ones currently configured and determine if any implements a `rewrite` handler in their interface. If not, a flag can prevent the rewrite phase from running the plugins runloop altogether.

james-callahan:
> As an alternative, we could take advantage of the fact that the list of enabled plugins is static (plugins that will potentially be configured, i.e. the plugins configuration property) and will only change when Kong restart. From there, we could loop through the ones currently configured and determine if any implements a rewrite handler in their interface. If not, a flag can prevent the rewrite phase from running the plugins runloop altogether.

Why not do it each time a plugin is enabled/disabled?

Tieske:
Imo there are 2 ways to go about this:

- do a proper research spike in how we can optimize the runloop and refactor (this is in the backlog)
- take a stab at optimizing some stuff ahead of the previous point if it can't wait

If taking the 2nd option, then @thibaultcha s suggestion makes sense. It shouldn't be hard to do since `rewrite` handlers are only executed on global plugins anyway.

thibaultcha:
> Why not do it each time a plugin is enabled/disabled?

The static solution is likely easier to implement and covers most deployments already. In practice, if at least one plugin implements the `rewrite` phase, it is likely not shipped with the default CE bundle. Yes it covers the last 5% of use-cases, but at an increased complexity in terms of implementation. A static solution would already help most use-cases and thus less susceptible to bugs as well (bugs in the runloop can be quite harmful).

If @p0pr0ck5 or anyone else is willing to implement either of the static or dynamic "skip rewrite plugins" patches, they'd be welcome!

thibaultcha:
Closing this for now given that the proposed suggestion is not to be scheduled anytime soon by core maintainers (and would warrant another PR anyway).

Contributions welcome!

PR #3652, opened 2018-07-27
feat(db) add operation to truncate single table in new DAO

hishamhm:
Introduces the ability to truncate a single table in the new DAO, using either
`entities:truncate()` or `db.truncate("entities")`.

Adjusts all tests to avoid truncating all tables at once deliberately. This is
done to speed up the test suite.

Note that the tests still truncate the entire database once at the beginning
of each file when they use `helpers.get_db_utils` in the top-level `setup`
directive.

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -222,6 +222,11 @@ function _M.new(db, schema, strategy, errors)
 end
 
 
+function DAO:truncate()
+  return self.strategy:truncate()
+end
+
+
 function DAO:select(primary_key)
   if type(primary_key) ~= "table" then
     error("primary_key must be a table", 2)
--- a/kong/db/init.lua
+++ b/kong/db/init.lua
@@ -158,8 +158,18 @@ function DB:reset()
 end
 
 
-function DB:truncate()
-  local ok, err = self.connector:truncate()
+function DB:truncate(table_name)
+  if table_name ~= nil and type(table_name) ~= "string" then
+    error("table_name must be a string", 2)
+  end
+  local ok, err
+
+  if table_name then
+    ok, err = self.connector:truncate_table(table_name)
+  else
+    ok, err = self.connector:truncate()
+  end
+
   if not ok then
     return nil, prefix_err(self, err)
   end
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -301,4 +301,28 @@ function CassandraConnector:truncate()
 end
 
 
+function CassandraConnector:truncate_table(table_name)
+  local ok, err = self:connect()
+  if not ok then
+    return nil, err
+  end
+
+  local cql = string.format("TRUNCATE TABLE %s.%s",
+                            self.keyspace, table_name)
+
+  local ok, err = self:query(cql)
+  if not ok then
+    self:setkeepalive()
+    return nil, err
+  end
+
+  ok, err = self:setkeepalive()
+  if not ok then
+    return nil, err
+  end
+
+  return true
+end
+
+
 return CassandraConnector
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -284,7 +284,7 @@ function CassandraConnector:truncate()
       local cql = string.format("TRUNCATE TABLE %s.%s",
                                 self.keyspace, table_name)
 
-      local ok, err = self:query(cql)
+      local ok, err = self:query(cql, nil, nil, "write")
       if not ok then
         self:setkeepalive()
         return nil, err
@@ -302,26 +302,10 @@ end
 
 
 function CassandraConnector:truncate_table(table_name)
-  local ok, err = self:connect()
-  if not ok then
-    return nil, err
-  end
-
   local cql = string.format("TRUNCATE TABLE %s.%s",
                             self.keyspace, table_name)
 
-  local ok, err = self:query(cql)
-  if not ok then
-    self:setkeepalive()
-    return nil, err
-  end
-
-  ok, err = self:setkeepalive()
-  if not ok then
-    return nil, err
-  end
-
-  return true
+  return self:query(cql, nil, nil, "write")
 end
 
 
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -313,6 +313,20 @@ function _mt:truncate()
 end
 
 
+function _mt:truncate_table(table_name)
+  local truncate_statement = {
+    "TRUNCATE TABLE ", table_name, " RESTART IDENTITY CASCADE;"
+  }
+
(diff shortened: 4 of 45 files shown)


kong/db/init.lua
```diff
@@ -130,8 +130,16 @@ function DB:reset()
...
+function DB:truncate(table_name)
+  if table_name ~= nil and type(table_name) ~= "string" then
+    error("table_name must be a string", 2)
+  end
+
+  if table_name then
+    return self.connector:truncate_table(table_name)
+  else
+    return self.connector:truncate()
+  end
```
thibaultcha: Could we avoid nesting the branch if our previous branch already returns, as usual?

kong/db/strategies/postgres/connector.lua
```diff
@@ -313,6 +313,21 @@ function _mt:truncate()
 end
 
 
+function _mt:truncate_table(table_name)
+
```
thibaultcha: why the blank line? :thinking:

spec/helpers.lua
```diff
@@ -122,8 +122,25 @@ do
...
+    else
+      db:truncate(t)
+    end
+  end
+end
+
+local function get_db_utils(strategy, tables)
   strategy = strategy or conf.database
+  if tables ~= nil and type(tables) ~= "table" then
+    error("arg #2 must be a list of tables to truncate")
```
thibaultcha: very minor because a spec utils, but this error call and the next one could use a second argument for the stack level - but ok to ignore

spec/02-integration/04-admin_api/02-apis_routes_spec.lua
```diff
@@ -20,7 +20,7 @@ pending("Admin API #" .. kong_config.database, function()
 
   setup(function()
     local _
-    _, db, dao = helpers.get_db_utils(kong_config.database)
+    _, db, dao = helpers.get_db_utils(kong_config.database, {})
```
thibaultcha: It seems like the second argument is optional, yet was added, empty, to many calls of `get_db_utils()` - for explicitness?

hishamhm:
Commenting down here, as the links above don't work since the PR was re-pushed:

The second argument for `get_db_utils is the list of tables to truncate: when not given, it truncates them all. The files with `{}` are the ones for which I reviewed their truncations and deemed safe to not truncate any tables upon startup (so yeah `{}` is different than none).

kong/db/dao/init.lua
```diff
@@ -221,6 +221,11 @@ function _M.new(db, schema, strategy, errors)
 end
 
 
+function DAO:truncate()
```
kikito: I think this function is never used. I am not sure it belongs in the DAO level, either - maybe it belongs on the "instantiated strategy" (combination of schema + strategy) level.

kong/db/strategies/cassandra/connector.lua
```diff
@@ -301,4 +301,28 @@ function CassandraConnector:truncate()
...
+function CassandraConnector:truncate_table(table_name)
+  local ok, err = self:connect()
+  if not ok then
+    return nil, err
+  end
+
+  local cql = string.format("TRUNCATE TABLE %s.%s",
+                            self.keyspace, table_name)
+
+  local ok, err = self:query(cql)
```
thibaultcha: I believe we should specify this query as a write operation, as per the 4th argument of this function?

kong/db/strategies/cassandra/connector.lua
```diff
@@ -301,4 +301,28 @@ function CassandraConnector:truncate()
 end
 
 
+function CassandraConnector:truncate_table(table_name)
+  local ok, err = self:connect()
```
thibaultcha: I believe the `:connect()` and `:setkeepalive()` calls aren't necessary, since `:query()` will do these for you? If there are no loops anyway, I think we should stay away from this pattern and let `:query()` do the work.

hishamhm:
@thibaultcha comments addressed, and rebased into `master`

kong/db/strategies/cassandra/connector.lua
```diff
@@ -284,44 +279,22 @@ function CassandraConnector:truncate()
...
-        self:setkeepalive()
         return nil, err
       end
     end
   end
 
-  ok, err = self:setkeepalive()
-  if not ok then
-    return nil, err
-  end
```
thibaultcha: In `:truncate()`, this is done on purpose (sorry if my previous comment wasn't clear enough about that). Because we loop over a significant number of rows, we avoid multiple, unnecessary, connect/keepalive operations. Especially useful in contexts when cosockets are unavailable, and `:setkeepalive()` is closing the underlying LuaSocket connection.

hishamhm: but it calls `self:query` in the loop, which does `setkeepalive` that as well, no?

thibaultcha: Not when the connection is opened prior to calling `:query()`:

https://github.com/Kong/kong/blob/master/kong/db/strategies/cassandra/connector.lua#L196

kong/db/init.lua
```diff
@@ -158,7 +158,15 @@ function DB:reset()
...
 
 
-function DB:truncate()
+function DB:truncate(table_name)
+  if table_name ~= nil and type(table_name) ~= "string" then
+    error("table_name must be a string", 2)
+  end
+
+  if table_name then
+    return self.connector:truncate_table(table_name)
```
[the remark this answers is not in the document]
hishamhm: fixed

spec/03-plugins/26-oauth2/04-invalidations_spec.lua
```diff
@@ -12,15 +12,16 @@ for _, strategy in helpers.each_strategy() do
...
     before_each(function()
-      assert(db:truncate())
-      dao:truncate_tables()
+      assert(db:truncate("routes"))
+      assert(db:truncate("services"))
+      assert(db:truncate("consumers"))
+      dao:truncate_table("plugins")
+      dao:truncate_table("oauth2_tokens")
+      dao:truncate_table("oauth2_credentials")
+      dao:truncate_table("oauth2_authorization_codes")
```
thibaultcha: Did this commit use to have all of the tests changes, or were they supposed to be part of the subsequent commit (reduce unnecessary truncations)? Either if fine if it was the goal, just making sure, because I seem to recall those spec changes were not in this commit

hishamhm: The spec changes related to the `truncate` calls were in this commit, and the spec changes related to the other modifications I made to the helpers ware in the other commit. But I ended up applying more and more truncation cleanups so they might have mixed up a bit, so I just moved all spec changes to the `tests(*)` commit now.

thibaultcha: Alright!

thibaultcha:
Manually merged (minor commit message edits)! Merging into next now

PR #3655, opened 2018-07-30
Fix wrong type exception in key-auth plugin

gjedeer:
add tostring() around consumer.username

### Summary

kong-community-edition-0.14.0-1.noarch on CentOS 7 fails with:

```
2018/07/30 11:22:21 [error] 17610#0: *1895 lua coroutine: runtime error: /usr/local/share/lua/5.1/kong/plugins/key-auth/handler.lua:65: invalid header value for "X-Consumer-Username": got userdata, expected string, number, boolean or array of strings
stack traceback:
coroutine 0:
        [C]: in function 'error'
        /usr/local/share/lua/5.1/kong/pdk/private/checks.lua:92: in function 'validate_headers'
        /usr/local/share/lua/5.1/kong/pdk/service/request.lua:370: in function 'set_headers'
        /usr/local/share/lua/5.1/kong/plugins/key-auth/handler.lua:65: in function 'set_consumer'
        /usr/local/share/lua/5.1/kong/plugins/key-auth/handler.lua:165: in function 'do_authentication'
        /usr/local/share/lua/5.1/kong/plugins/key-auth/handler.lua:188: in function </usr/local/share/lua/5.1/kong/plugins/key-auth/handler.lua:171>
coroutine 1:
        [C]: in function 'resume'
        coroutine.wrap:21: in function <coroutine.wrap:21>
        /usr/local/share/lua/5.1/kong/init.lua:468: in function 'access'
        access_by_lua(nginx-kong.conf:86):2: in function <access_by_lua(nginx-kong.conf:86):1>,
```

I've converted the userdata object to string to prevent the crash from happening.

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -46,7 +46,7 @@ local function set_consumer(consumer, credential)
   local new_headers = {
     [const.CONSUMER_ID]        = consumer.id,
     [const.CONSUMER_CUSTOM_ID] = tostring(consumer.custom_id),
-    [const.CONSUMER_USERNAME]  = consumer.username,
+    [const.CONSUMER_USERNAME]  = tostring(consumer.username),
   }
 
   kong.ctx.shared.authenticated_consumer = consumer -- forward compatibility


Tieske:
I don't think this is the right approach, simply because the `username` ought to be `nil` or a string. So question is why it has a userdata as value.

The only thing that comes to mind is that it might be a `null` value, because the `consumers` were moved over to the new DB instances. Maybe @hisham knows?

hishamhm:
@Tieske yes, `consumer.username ~= ngx.null and consumer.username` should be the correct fix (also to the `custom_id` line above). @kikito, does that sound right?

gjedeer:
I have no idea how that happened, after migration to 0.14 my Kong instance stopped working and started giving this error.

Tieske:
@hishamhm if that is the case (as I suspected), then this problem will most likely be in many places in the code base

thibaultcha:
This is indeed not a valid approach for this issue. Nonetheless, thanks @gjedeer for the patch and also raising this to our attention.

Another stab at this has been given by @bungle in #3700, but we might even make a more radical change to avoid having this issue spread throughout the codebase, and requiring both `~= nil` and `~= ngx.null` checks like so.

Closing this for now; thanks!

PR #3656, opened 2018-07-30
LDAP Auth: Fix WWW-Authenticate header with custom cred type

francois-maillard:

### Summary

When using custom credential type for LDAP authentication (the `config.header_type` setting),
the WWW-Authenticate header that is sent back when auth failed should be accordingly set.

Using `Basic` as the header type should allow the use of any web browser as a client. When there's no (or an invalid) authentication data in the query, Kong replies with the `WWW-Authenticate` header. Currently, this header doesn't use  `config.header_type` so it sets `WWW-Authenticate: LDAP`, which is not understood by the browser, hence **the login popup won't show**. 

### Full changelog

* Fix in `kong/plugins/ldap-auth/access.lua`
* Add related tests in `spec/03-plugins/21-ldap-auth/01-access_spec.lua`

### Issues resolved

I don't think a proper issue has been open, but it has been reported [on discuss.konghq.com](https://discuss.konghq.com/t/how-to-enable-ldap-auth-browser-popup/1579)

### DIFF
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -5,6 +5,7 @@ local ldap = require "kong.plugins.ldap-auth.ldap"
 
 local match = string.match
 local lower = string.lower
+local upper = string.upper
 local find = string.find
 local sub = string.sub
 local fmt = string.format
@@ -165,7 +166,14 @@ local function do_authentication(conf)
 
   -- If both headers are missing, return 401
   if not (authorization_value or proxy_authorization_value) then
-    ngx.header["WWW-Authenticate"] = 'LDAP realm="kong"'
+    local scheme = conf.header_type
+    if scheme == "ldap" then
+      -- ensure backwards compatibility (see GH PR #3656)
+      -- TODO: provide migration to capitalize older configurations
+      scheme = upper(scheme)
+    end
+
+    ngx.header["WWW-Authenticate"] = scheme .. ' realm="kong"'
     return false, {status = 401}
   end
 
--- a/spec/03-plugins/21-ldap-auth/01-access_spec.lua
+++ b/spec/03-plugins/21-ldap-auth/01-access_spec.lua
@@ -123,7 +123,7 @@ for _, strategy in helpers.each_strategy() do
           start_tls = false,
           base_dn   = "ou=scientists,dc=ldap,dc=mashape,dc=com",
           attribute = "uid",
-          header_type = "basic",
+          header_type = "Basic",
         }
       }
 
@@ -359,6 +359,20 @@ for _, strategy in helpers.each_strategy() do
         }
       })
       assert.response(r).has.status(200)
+    end)
+    it("returns 'invalid credentials' and www-authenticate header when the credential is missing and custom credential type is configured", function()
+      local res = assert(proxy_client:send {
+        method  = "GET",
+        path    = "/get",
+        headers = {
+          host  = "ldap5.com"
+        }
+      })
+      assert.response(res).has.status(401)
+      local value = assert.response(res).has.header("www-authenticate")
+      assert.are.equal('Basic realm="kong"', value)
+      local json = assert.response(res).has.jsonbody()
+      assert.equal("Unauthorized", json.message)
     end)
     it("fails if custom credential type is invalid in post request", function()
       local r = assert(proxy_client:send {


kong/plugins/ldap-auth/access.lua
```diff
@@ -165,7 +165,7 @@ local function do_authentication(conf)
 
   -- If both headers are missing, return 401
   if not (authorization_value or proxy_authorization_value) then
-    ngx.header["WWW-Authenticate"] = 'LDAP realm="kong"'
+    ngx.header["WWW-Authenticate"] = conf.header_type .. ' realm="kong"'
```
Tieske: existing configurations will mostly have the old default set `ldap` (lowercase) so wouldn't this break an existing configuration after an upgrade?

Adding a migration that converts `ldap` to `LDAP` should fix the defaults. Not sure though about folks not using the default. But probably since no one complained about this issue before, it's safe to just convert `ldap` to `LDAP`?

What do you think @bungle ?

francois-maillard: I must say I had overlooked this issue. Yet, I'm not sure any client actually does anything with the LDAP type in `WWW-Authenticate`, since it's not listed in the [list maintained by IANA](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml)

francois-maillard: I capitalized the default value so as to avoid a breaking change from `WWW-Authenticate: LDAP` to `WWW-Authenticate: ldap` in the response (It doesn't impact the `Authentication: ldap xxxxxxx` header in a valid request since this check is case insensitive). That was not a good option since I overlooked the fact that the default value is stored in the DB and so will only be right in a fresh install.

Here are the options I see:
1. Keep the original default value `ldap`, in which case the default `WWW-Authenticate` header will change to lowercase
2. Change the default value to `LDAP`, and, as @Tieske proposed, add a migration step
3. Change my code to `ngx.header["WWW-Authenticate"] = upper(conf.header_type) .. ' realm="kong"'` but in the case of basic auth, Kong would answer `WWW-Authenticate: BASIC` while the official answer should be `WWW-Authenticate: Basic`

2 seems like the best option... If you agree with that, I can write the migration.

Regarding @Tieske remark:
> Not sure though about folks not using the default.

If someone does not use the default, then the `WWW-Authenticate: LDAP` is not a valid response, so we'll fix a bug for them (actually, that's exactly the issue I have with basic auth).

thibaultcha: 2. is more work but sounds the sanest indeed. Although a migration means that this _must_ target next (which it does) and will only make it into the next major release (far from now).

A possible solution could be to keep the updated default to LDAP:
```lua
local scheme = conf.header_type
if scheme == "ldap" then
  -- ensure backwards compatibility (see GH PR #3656)
  -- TODO: provide migration to capitalize older configurations
  scheme = upper(scheme)
end

ngx.header["WWW-Authenticate"] = scheme .. ' realm="kong"'
```

This way, this PR can go to master. A subsequent PR (that you can contribute later at your own pace) would introduce the migration (there are other similar migrations to copy-paste and tweak, it should be fairly simple), remove this piece code, and target the next branch.

francois-maillard:
hmm... @thibaultcha , I followed your last proposal, and reworked my commits and targeted the master branch. It looks like it broke our discussion stream though: our messages don't appear anymore in the discussion. Sorry about that :(

kong/plugins/ldap-auth/access.lua
```diff
@@ -165,7 +165,14 @@ local function do_authentication(conf)
 
   -- If both headers are missing, return 401
   if not (authorization_value or proxy_authorization_value) then
-    ngx.header["WWW-Authenticate"] = 'LDAP realm="kong"'
+    local scheme = conf.header_type
+    if scheme == "ldap" then
+      -- ensure backwards compatibility (see GH PR #3656)
+      -- TODO: provide migration to capitalize older configurations
+      scheme = upper(scheme)
```
thibaultcha: `upper` isn't a valid Lua function, `string.upper` is. I was suggesting caching it at the top of this module, like `string.lower`

francois-maillard: yep... sorry, did it in a rush before leaving the office.
The plugin unit tests did pass, though the [cassandra integration job](https://travis-ci.org/Kong/kong/jobs/411478395) failed with an error that does not seem related to by patch:
```
[  FAILED  ] spec/02-integration/02-cmd/02-start_stop_spec.lua @ 114: kong start/stop /etc/hosts resolving in CLI resolves #cassandra hostname
```

thibaultcha: @francois-maillard Spurious failures can happen in the CI, in particular on PR builds. It is something we are working towards fixing, no worries for now, I'll restart the jobs if those failures are unrelated.

francois-maillard: All tests are green!

thibaultcha:
Manually merged, thank you!

As discussed, will you now be able to provide a subsequent patch for the migration towards the capitalized default value @francois-maillard? That'd be greatly appreciated, thanks!

thibaultcha:
@francois-maillard PS: cool picture :)

francois-maillard:
Leaving in the south of france can be tough ;)
I'll try do so next week, if I can't find the time, it might then take longer as I'll be off for two weeks.

PR #3658, opened 2018-07-31
feat(nightly) make rhel:6/7 and amazonlinux nightly as well

hutchic:
add rhel:6, rhel:7 and amazonlinux to the trigger-travis job that builds kong


thibaultcha:
Thanks!

PR #3659, opened 2018-07-31
Feat/groundwork for targets to new dao

kikito:
These are a few uncontroversial changes that are needed in order to merge feat/targets-to-new-dao

### DIFF
--- a/kong/dao/factory.lua
+++ b/kong/dao/factory.lua
@@ -417,6 +417,14 @@ local function migrate(self, identifier, migrations_modules, cur_migrations, on_
       err = migration.up(self.db, self.kong_config, self)
     end
 
+    if err and migration.ignore_error then
+      if type(migration.ignore_error) == "boolean" or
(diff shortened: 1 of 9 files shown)


kong/dao/factory.lua
```diff
@@ -417,6 +417,14 @@ local function migrate(self, identifier, migrations_modules, cur_migrations, on_
       err = migration.up(self.db, self.kong_config, self)
     end
 
+    if err and migration.ignore_error then
+      if type(migration.ignore_error) == "boolean" or
+         (type(migration.ignore_error) == "string" and
+          string.find(tostring(err), migration.ignore_error)) then
```
hishamhm: I suggest chaning this to `string.find(tostring(err), migration.ignore_error, 1, true)` so that it does a plain-text match, just so this doesn't fail mysteriously someday in the future when we have a message that happens to have a hyphen.

PR #3660, opened 2018-07-31
fix(schema) declare timestamps as number to keep millisecond precision

kikito:
Currently the `created_at` the field in the new DAO uses second-level numbers instead of millisecond-level numbers (since we use `ngx.time` instead of `ngx.now`)

This changes that and makes all timestamps be stored with millisecond precision (no changes to the database are required, just that we replaced `ngx.time` by `ngx.now`).

However this is a breaking change: until now the `created_at` field looked like an integer, and now it will start looking like a float. 

This change is required by #3622

### DIFF
--- a/kong/db/schema/entities/certificates.lua
+++ b/kong/db/schema/entities/certificates.lua
@@ -7,7 +7,7 @@ return {
 
   fields = {
     { id = typedefs.uuid, },
-    { created_at     = { type = "integer", timestamp = true, auto = true }, },
+    { created_at     = typedefs.auto_timestamp },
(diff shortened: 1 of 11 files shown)


Tieske:
My guess would be that this is because of the time order required by the targets?

An alternative would be to replace the UUID by a [ULID](https://github.com/ulid/spec). In a ULID, part of the randomness is sacrificed and replaced with a time stamp. The way this is done ensures that you can lexically sort based on the ULID, and get historical order.

A ULID is compatible (as a format) with UUID

But I don't know what the impact of that would be regarding breaking changes.

PR #3661, opened 2018-07-31
Feat/init connector

kikito:
Calls `init_connector` on `DB.new`.

Needed for #3622


kikito:
I ended up changing #3622 so this is no longer needed. Closing down this PR since it is no longer needed.

PR #3663, opened 2018-08-01
fix(travis) adjust travis syntax so it runs deploy jobs

hutchic:


thibaultcha:
@hutchic Could you describe why the change is needed and what it does? It’d make merging easier and faster and allows everybody to maybe learn something.

hutchic:
@thibaultcha I should have made this in kong-private (or tagged it WIP) sorry. Basically travis didn't run the deploy job to build the nightly releases like it was supposed to so I'm trying to tweak the `travis.yml` so it matches exactly what's expected.

Exactly what's expected is a little mysterious though ... my hypothesis is travis doesn't like mixing build stages with non-build stages

PR #3671, opened 2018-08-03
fix(api) no longer hide the strategy in returned errors

hishamhm:
**Note:** `next` is currently broken because tests expect `strategy` to appear but the code doesn't produce it; this is one possible solution: changing the tests. The other one is in an alternative PR, #3670 — one should be merged, the other dropped. :)

About this PR:

Avoid hiding the strategy name in errors returned by the Admin API.


thibaultcha:
Closing this since #3670 was merged. Thanks!

PR #3675, opened 2018-08-06
Bump lua-resty-healthcheck version

gaetanfl:
### Summary

This PR bumps lua-resty-healthcheck version to 0.5.0. The 0.5.0 version includes https active checks.

### Full changelog

* Bump lua-resty-healthcheck version

### DIFF
--- a/kong-0.14.0-0.rockspec
+++ b/kong-0.14.0-0.rockspec
@@ -31,7 +31,7 @@ dependencies = {
   "lua-resty-dns-client == 2.1.0",
   "lua-resty-worker-events == 0.3.3",
   "lua-resty-mediador == 0.1.2",
-  "lua-resty-healthcheck == 0.4.2",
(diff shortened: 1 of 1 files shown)


hishamhm:
Thank you for the PR! We'll need a bit more in the Kong side to fully support https healthchecks, since the schema validation for Upstream entities needs to be updated accordingly. We are working on converting the Upstream entities to the new schema format and DAO in https://github.com/Kong/kong/pull/3622, so it will probably make sense to add this once that is merged.

hishamhm:
We could merge this version bump as-is, though (with the usual adjustments to the commit message), since the library update itself is backwards compatible.

gaetanfl:
I wasn't aware of that, thanks for the explanation It can wait indeed.

thibaultcha:
@gaetanfl Thanks! This is good to merge early indeed, especially for our CI to run with 0.5.0 from now on.

PR #3677, opened 2018-08-06
fix(ldap-auth) use username, password for cache key

shashiranjan84:
### Summary

This change adds password as part of cache key, doing so
we can cache invalid credential without blocking valid credential

### Full changelog

*  Update cache key generation logic to use password
*  Update invalidation spec to test changes
*  Minor log refactor

### DIFF
--- a/spec/03-plugins/21-ldap-auth/02-invalidations_spec.lua
+++ b/spec/03-plugins/21-ldap-auth/02-invalidations_spec.lua
@@ -1,5 +1,4 @@
 local helpers = require "spec.helpers"
-local cjson = require "cjson"
 local fmt = string.format
 local lower = string.lower
 local md5 = ngx.md5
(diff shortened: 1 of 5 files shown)


spec/03-plugins/21-ldap-auth/02-invalidations_spec.lua
```diff
@@ -28,29 +27,36 @@ for _, strategy in helpers.each_strategy() do
           ldap_port = "389",
           start_tls = false,
           base_dn   = "ou=scientists,dc=ldap,dc=mashape,dc=com",
-          attribute = "uid"
+          attribute = "uid",
+          cache_ttl = 5,
```
thibaultcha: Could we use a slightly shorter TTL, to avoid slowing down our tests? 5s isn't much, but when every single test relying on TTL behaviour in the suite waits for 5s between each test, we end up with a problem... Maybe 1? 2?

shashiranjan84: sure, will update it.

shashiranjan84: I set it to one, please review

PR #3679, opened 2018-08-07
feat(origins) Add ability to override hosts from kong configuration on a per-kong-node basis

james-callahan:
### Summary

This allows (via configuration, to facilitate it on a per-node basis instead of cluster-wide) overriding the scheme, host and port before resolution occurs in the balancer phase.

### Full changelog

  - feat(utils) normalize ipv6 address when formatting
  - feat(core) add 'origins' configuration property


james-callahan:
Needs tests.

kong/runloop/balancer.lua
```diff
@@ -752,6 +752,14 @@ end
 -- which do not support sending it)
 -- @return true on success, nil+error message+status code otherwise
 local function execute(target, ctx)
+  -- Check for KONG_ORIGINS override
+  local origin = singletons.origins[target.scheme .. "://" .. utils.format_host(target)]
```
thibaultcha:
```lua
local origin = singletons.origins[target.scheme .. "://" .. 
                                  utils.format_host(target)]
```

_Maybe_
```lua
local origin = target.scheme .. "://" .. utils.format_host(target)
origin = singletons.origins[origin]
```

kong/runloop/balancer.lua
```diff
@@ -752,6 +752,16 @@ end
 -- which do not support sending it)
 -- @return true on success, nil+error message+status code otherwise
 local function execute(target, ctx)
+  do -- Check for KONG_ORIGINS override
+    local origin_key = target.scheme .. "://" .. utils.format_host(target)
```
Tieske: Since this comparison might be expensive maybe cache the results?

Does this cater for matching different formats? casing in names, shortening in ipv6? could use some accompanying tests

spec/02-integration/05-proxy/16-origins_spec.lua
```diff
@@ -0,0 +1,121 @@
+local helpers = require "spec.helpers"
```
Tieske: this file could use some more vertical white space

kong/runloop/balancer.lua
james-callahan: What do you mean "comparison" here? What would you cache?

Tieske: sorry bad wording. Creating the key and then looking it up (=comparison) in the `origins` table.

so if `origins` has

```lua
{
  ["http://[::1]:80"] = {
    scheme = "http",
    host = "somewhere.else",
    port = 80,
  }
}
```

and incoming is `http://[0000::1]:80` then it doesn't match, so you need to normalize. That is expensive, and that is something you could cache.

kong/runloop/handler.lua
```diff
@@ -646,14 +647,17 @@ return {
       end
 
       do
+        local balancer_data = ctx.balancer_data
+
+        -- set the upstream scheme as balancer selected
+        local upstream_scheme = balancer_data.scheme
+        var.upstream_scheme = upstream_scheme
```
thibaultcha: note 1: this is a breaking change for any plugin out there that relies on `ngx.var.upstream_scheme = "..."`, they now need to do `ngx.ctx.balancer_data.scheme = "..."`.

note 2: a good way to avoid such breaking changes would be to provide a PDK API for this operation (none currently exist in the `kong.service` module); this is outside the scope of this PR of course.

note 3: the `var.upstream_scheme = match_t.upstream_scheme` assignment in the `access.before` handler could arguably be removed, since plugins making use of it are already broken.

It is hard to know if such a change would have any impact - I myself am not sure if we offer any plugins making use of this variable (forward-proxy seems to hard-code `http://`, luckily). I do have a vague memory of it being mentioned from community users though.

kong/runloop/balancer.lua
```diff
@@ -752,6 +752,17 @@ end
 -- which do not support sending it)
 -- @return true on success, nil+error message+status code otherwise
 local function execute(target, ctx)
+  do -- Check for KONG_ORIGINS override
+    local origin_key =
+      target.scheme .. "://" .. utils.format_host(utils.normalize_ip(target))
```
thibaultcha: Something caught my eye, not sure what I may have missed: `utils.normalize_ip()` expects a string, but `target` is a table here, correct?

spec/02-integration/05-proxy/16-origins_spec.lua
```diff
@@ -0,0 +1,121 @@
...
+      bp, db, dao = helpers.get_db_utils()
+    end)
+
+    after_each(function()
+      if proxy_client then
+        proxy_client:close()
+      end
+
+      dao:truncate_tables()
+      assert(db:truncate())
```
thibaultcha: Better to let the `get_db_utils()` utility do this type of work (above line as well) for speed and debug reasons.

spec/02-integration/05-proxy/16-origins_spec.lua
```diff
@@ -0,0 +1,121 @@
...
+        nginx_conf = "spec/fixtures/custom_nginx.template",
+      }))
+
+      proxy_client = helpers.proxy_client()
+      local res = proxy_client:get("/request", {
+        headers = { Host = "mock_upstream" }
+      })
+      assert.res_status(502, res)
+      proxy_client:close()
+      helpers.stop_kong(nil, nil, true)
```
thibaultcha: `stop_kong()` only accept 2 arguments nowadays, this can be written as `helpers.stop_kong()` now

kong/conf_loader.lua
```diff
@@ -733,6 +734,33 @@ local function load(path, custom_conf)
...
+      -- Validate 'from'
+      local from_authority, err =
+        utils.format_host(utils.normalize_ip(from_host_port))
+      if not from_authority then
+        return nil, err
+      end
+      -- Check for duplicates
+      local from_origin = from_scheme .. "://" .. from_authority
+      if seen_origins[from_origin] then
+        return nil, "duplicate origin"
```
thibaultcha: This message is a bit unfriendly, isn't it?

1. It does not specify which origin is duplicated
2. If more than one origin is duplicated, the user will only find out after several trial and errors. Something that the `conf_loader` tries hard to avoid. Instead, it always tries to be a good citizen and report all errors at once, to avoid back and forth from the user.

kong/runloop/handler.lua
james-callahan: After some research, the PDK *does* have a function to modify the scheme. `request.set_scheme`.

thibaultcha: Indeed; after our irl this does mean that the patch in its current form breaks the PDK. Options were discussed, for now that is a blocker for this PR.

kong/tools/utils.lua
```diff
@@ -786,7 +786,7 @@ _M.format_host = function(p1, p2)
     return nil, "cannot format type '" .. t .. "'"
   end
   if typ == "ipv6" and not find(host, "[", nil, true) then
-    return "[" .. host .. "]" .. (port and ":" .. port or "")
+    return "[" .. _M.normalize_ipv6(host) .. "]" .. (port and ":" .. port or "")
```
Tieske: This is not ok.

There are 3 functions for normalizing/validating, and there is 1 for formatting (this one). With this change the formatter:

- normalizes/validates IPv6
- does not normalize/validate IPv4 nor names
- suddenly becomes a lot more expensive, due to the 1st point above

spec/01-unit/004-utils_spec.lua
```diff
@@ -479,8 +479,8 @@ describe("Utils", function()
       it("correctly formats addresses", function()
         assert.are.equal("1.2.3.4", utils.format_host("1.2.3.4"))
         assert.are.equal("1.2.3.4:80", utils.format_host("1.2.3.4", 80))
-        assert.are.equal("[::1]", utils.format_host("::1"))
-        assert.are.equal("[::1]:80", utils.format_host("::1", 80))
+        assert.are.equal("[0000:0000:0000:0000:0000:0000:0000:0001]", utils.format_host("::1"))
+        assert.are.equal("[0000:0000:0000:0000:0000:0000:0000:0001]:80", utils.format_host("::1", 80))
```
Tieske: this should be formatting only, not normalizing

spec/02-integration/05-proxy/16-origins_spec.lua
```diff
@@ -0,0 +1,61 @@
+local helpers = require "spec.helpers"
+
+describe("origins config option", function()
```
Tieske: Please add tests for the proper matching of incoming (non-normalized) addresses to the normalized ones in `origins`.

Tieske: @James Callahan with my current understanding of the code it will simply fail if you configure a service with an upstream url `http://192.168.001.001/somewhere` due to a lack of normalizing the ipv4.
Also be aware that plugins might change the upstream properties, so the normalization must be done in each request. Just updating the service entity is not enough.

Tieske: fyi have a look at this example code: https://github.com/Tieske/ncache.lua/blob/master/examples/ipcache.lua#L241-L258

kong/tools/utils.lua
james-callahan: All of our other "normalization" functions don't actually do any normalization. They just do validation.

spec/01-unit/004-utils_spec.lua
james-callahan: Thibault and I discussed yesterday and decided that formatting is allowed to include normalisation.

spec/02-integration/05-proxy/16-origins_spec.lua
james-callahan: IPv4 addresses are not allowed to contain leading zeros: that should be a syntax error according to RFC 3986. If we don't check that on input then that is where the fix needs to be.

kong/tools/utils.lua
Tieske: If we change behaviour, it should be consistent. Currently it moved the other direction

spec/02-integration/05-proxy/16-origins_spec.lua
Tieske: That does not matter. Since the utility functions are now inconsistent. And those functions serve a bigger purpose than just this change.

And even then the test cases should be added to show it is validated in the right places.

And finally the function is expensive and should not be on a hot code path imo.

kong/tools/utils.lua
Tieske: Also the documentation is no longer correct add the behaviour of the function changed.

kong/conf_loader.lua
```diff
@@ -425,6 +426,46 @@ local function check_and_infer(conf)
...
+    if not from_scheme then
+      errors[#errors + 1] = "an origin must be of the form " ..
+                            "'from_scheme://from_host:from_port=" ..
+                            "to_scheme://to_host:to_port', got '" ..
+                            v .. "'"
+
+    else
+      -- Validate 'from'
+      local from_authority, err =
+        utils.format_host(utils.normalize_ip(from_host_port))
```
Tieske: This code now normalized an ipv6 address twice. Once before calling format and once inside the format call.

PR #3680, opened 2018-08-07
Bump luasec to 0.7

prettykingking:
### Summary

The dependency luasec 0.6 is not compatible with OpenSSL 1.1.0, thus could not compile.

See the issue
[https://github.com/brunoos/luasec/issues/101](url)

### Issues resolved

- Bump dependency luasec to 0.7


thibaultcha:
Merged in master - thank you!

PR #3681, opened 2018-08-07
 feat(core) add kong.resty.getssl module 

james-callahan:
### Summary

This allows fetching the `SSL*` object for the current client as a luaossl object

### DIFF
--- a/kong/resty/getssl.lua
+++ b/kong/resty/getssl.lua
@@ -1,0 +1,32 @@
+local ffi = require "ffi"
+local pushssl = require "openssl.ssl".pushffi -- will define SSL* in ffi
+local get_ssl_pointer = require "ngx.ssl".get_ssl_pointer
+
+
+local getssl
+if get_ssl_pointer == nil then
+  local err_msg = "OpenResty patch missing. See https://github.com/Kong/openresty-patches"
+
+  ngx.log(ngx.WARN, err_msg)
+
+  function getssl()
+    return nil, err_msg
+  end
+else
+  local cast = ffi.cast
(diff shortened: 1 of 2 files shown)


kong/resty/getssl.lua
```diff
@@ -0,0 +1,23 @@
...
+local function getssl()
+  local ptr, err = get_ssl_pointer()
+  if not ptr then
+    return nil, err
+  end
+  ptr = cast(SSLp, ptr)
+  return pushssl(ptr)
+end
+
+return {
```
thibaultcha: style: please respect the 2 lines jump between logical blocks in this codebase - thanks!

kong/resty/getssl.lua
```diff
@@ -0,0 +1,23 @@
+local ffi = require "ffi"
+local pushssl = require "openssl.ssl".pushffi -- will define SSL* in ffi
+local get_ssl_pointer = require "ngx.ssl".get_ssl_pointer
+
+if get_ssl_pointer == nil then
+  error("Openresty patch missing. See https://github.com/Kong/openresty-patches")
```
thibaultcha: typo: OpenResty

More importantly: we should _gracefully_ fail if the patch is not available, but not prevent the module from loading altogether, which would certainly prevent Kong from even starting; while in fact, this feature might not even be used.

james-callahan: I'm not sure how this *should* fail gracefully: if the patch is not applied and you're using this module, then you've likely got a security issue.

We *could* create a 'dummy' function that always returns `nil, "OpenResty patch missing"`. But would it get checked correctly? I guess that will do for now....

PR #3683, opened 2018-08-07
chore(travis) separate the cassandra / postgres tests

hutchic:
### Summary

Separates the cassandra and postgres tests in test code as well as when ran via travis-ci


.travis.yml
```diff
@@ -29,59 +30,62 @@ env:
...
+    - stage: lint and unit
+      script:
+      - luacheck -q .
+      - bin/busted -v -o gtest spec/01-unit
+      - bin/busted -v -o gtest spec-old-api/01-unit
       env:
-        - TEST_SUITE=none
-    - stage: nightly deploy
-      if: TRAVIS_EVENT_TYPE == cron
+        - KONG_TEST_PG_DATABASE=foobar
```
thibaultcha: Why do we need this? Any particular unit test are relying on this not being set maybe? If so, could it be that this belongs to one of the setup scripts which could call `unset KONG_TEST_PG_DATABASE`? If not, then at the very least we could probably find a more elegant name for this value.

hutchic: It's the unit tests that are expecting `foobar`. We can move this to the `run_tests.sh` if we want I'm not sure of the benefits I always prefer things as ENV personally

thibaultcha: @hutchic Which test specifically? I don't think we ever needed to set this variable for the unit tests to succeed? :thinking:

hutchic: https://travis-ci.org/Kong/kong/jobs/413320585#L1794

```
[ RUN      ] spec/01-unit/003-prefix_handler_spec.lua @ 418: NGINX conf compiler prepare_prefix() dump Kong conf (custom conf)
spec/01-unit/003-prefix_handler_spec.lua:426: Expected objects to be the same.
```

thibaultcha: Something else must be broken with it, because this test sets its own value via the configuration loader arguments: https://github.com/Kong/kong/blob/master/spec/01-unit/003-prefix_handler_spec.lua#L420 It certainly shouldn't need a custom environment variable to be able to pass (that'd be a pretty terrible unit test...)

.travis.yml
```diff
@@ -36,14 +36,10 @@ env:
...
-    - KONG_TEST_PG_DATABASE=travis
-    - KONG_TEST_CASSANDRA_KEYSPACE=kong_tests
-    - KONG_TEST_DB_UPDATE_PROPAGATION=1
   matrix:
-    - KONG_TEST_DATABASE=postgres
-    - KONG_TEST_DATABASE=cassandra CASSANDRA=2.2
-    - KONG_TEST_DATABASE=cassandra CASSANDRA=3.9
+    - KONG_TEST_DATABASE=postgres KONG_TEST_PG_DATABASE=travis KONG_TEST_PG_USER=postgres KONG_TEST_PG_DATABASE=travis
+    - KONG_TEST_DATABASE=cassandra CASSANDRA=2.2 KONG_TEST_CASSANDRA_KEYSPACE=kong_tests KONG_TEST_DB_UPDATE_PROPAGATION=1
+    - KONG_TEST_DATABASE=cassandra CASSANDRA=3.9 KONG_TEST_CASSANDRA_KEYSPACE=kong_tests KONG_TEST_DB_UPDATE_PROPAGATION=1
```
thibaultcha: I think I have suggested it before, but those variables should be added to `.ci/setup_env.sh` _or_ `.ci/run_tests.sh`. Since they are very specific to the matrix itself, and duplicated all over the place if we keep them here (especially when we start adding new rows to this build matrix).

PR #3685, opened 2018-08-08
feat(db) implement the db:cluster_mutex() utility

thibaultcha:
This function can be called to set a cluster-wide lock (via the
underlying DB, currently C* only) to execute some code in a
"cluster-wide mutex".

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -309,4 +309,87 @@ function CassandraConnector:truncate_table(table_name)
 end
 
 
+function CassandraConnector:setup_locks(default_ttl)
+  local ok, err = self:connect()
+  if not ok then
+    return nil, err
+  end
+
+  local cql = string.format([[
+    CREATE TABLE IF NOT EXISTS locks(
+      key text PRIMARY KEY,
+      owner text
+    ) WITH default_time_to_live = %d
+  ]], default_ttl)
+
+  local ok, err = self:query(cql)
+  if not ok then
+    self:setkeepalive()
+    return nil, err
+  end
+
+  ok, err = self.cluster:wait_schema_consensus(self.connection)
+  if not ok then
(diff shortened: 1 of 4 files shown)


kong/db/strategies/cassandra/connector.lua
```diff
@@ -309,4 +309,87 @@ function CassandraConnector:truncate_table(table_name)
...
+  if not ok then
+    return nil, err
+  end
+
+  local cql = string.format([[
+    CREATE TABLE IF NOT EXISTS locks(
+      key text PRIMARY KEY,
+      owner text
+    ) WITH default_time_to_live = %d
+  ]], default_ttl)
```
Tieske: Shouldn't the above be in a migration? Not as much for this instance as well as creating a precedent.

Unless the intent is to use this lock, during migrations?

thibaultcha: > Unless the intent is to use this lock, during migrations?

Yes indeed

spec/02-integration/000-new-dao/03-db_cluster_mutex_spec.lua
```diff
@@ -0,0 +1,112 @@
...
+        local t2 = ngx.thread.spawn(function()
+          local _, err = db:cluster_mutex("my_key_2", { owner = "2" }, cb2)
+          assert.is_nil(err)
+        end)
+
+        ngx.thread.wait(t1)
+        ngx.thread.wait(t2)
+
+        assert.spy(cb2).was_called()
+        assert.spy(cb1).was_not_called()
```
hishamhm: since cb2 takes 0.2 to run and t1 waits 0.2 until it acquires, isn't there a chance we get unlucky in this test and t2 releases the lock before t1 tries to acquire? (in other words, shouldn't the sleep in cb2 be a little longer than t1 to be on the safe side, or am I misunderstanding the test?)

PR #3686, opened 2018-08-08
Provide with migrations following #3656

francois-maillard:
1. Add postgres/cassandra migrations
2. Change default config.header_type value from ldap to LDAP
3. Remove temporary code in ldap-auth/access.lua

### Summary

Provide with migrations following #3656

### Full changelog

* Add postgres/cassandra migrations to update config.header_type value from ldap to LDAP
* Change default config.header_type value from ldap to LDAP
* Remove temporary code added in #3656 in ldap-auth/access.lua

### Issues resolved

Cleanup following #3656

### DIFF
--- a/kong/plugins/ldap-auth/migrations/cassandra.lua
+++ b/kong/plugins/ldap-auth/migrations/cassandra.lua
@@ -20,4 +20,35 @@ return {
     end,
     down = function(_, _, dao) end  -- not implemented
   },
+  {
+    name = "2018-08-08-100000_header_type_default_uppercase",
+    up = function(_, _, dao)
+      for ok, config, update in plugin_config_iterator(dao, "ldap-auth") do
+        if not ok then
+          return config
+        end
+        if config.header_type == 'ldap' then
+          config.header_type = 'LDAP'
+          local _, err = update(config)
+          if err then
+            return err
+          end
+        end
+      end
+    end,
+    down = function(_, _, dao)
+      for ok, config, update in plugin_config_iterator(dao, "ldap-auth") do
+        if not ok then
+          return config
+        end
+        if config.header_type == 'LDAP' then
+          config.header_type = 'ldap'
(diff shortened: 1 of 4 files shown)


kong/plugins/ldap-auth/migrations/cassandra.lua
```diff
@@ -20,4 +20,35 @@ return {
...
     down = function(_, _, dao) end  -- not implemented
   },
+  {
+    name = "2018-08-08-100000_header_type_default_uppercase",
+    up = function(_, _, dao)
+      for ok, config, update in plugin_config_iterator(dao, "ldap-auth") do
+        if not ok then
+          return config
+        end
+        if config.header_type == 'ldap' then
```
thibaultcha: style: will you please use double quotes instead of singles? Thanks!

francois-maillard:
It seems like #3731 removes the need for this PR. If the `ldap` string is hardcoded, then it is logical that the `if scheme == "ldap"` test stays hardcoded as well.

There's an option though: we could just remove the capitalization check and keep:
```
  if not (authorization_value or proxy_authorization_value) then
    local scheme = conf.header_type
    ngx.header["WWW-Authenticate"] = scheme .. ' realm="kong"'
    return false, {status = 401}
  end
```
That would be a breaking change (the default `WWW-Authenticate` header would change from `LDAP kong` to `ldap kong`, but I don't think anyone would notice since, from my understanding, this code is only triggered when no authentication header is passed, which would only be the case in a human interactive call... and since neither `LDAP` nor `ldap` are official types, no browser/client should be using it.

What do you think... breaking change that should not be seen, or 3 "dirty" lines of code? Or maybe another option?

bungle:
I am closing this as it feels like #3731 gave alternative way to fix it. If there is still work to be done here, could you please open a new one and rebase on latest code.

Issue #3634, opened 2018-07-19
Spike in postgres connections when routes are added or updated


ganeshs:
We are observing a spike in database connections whenever we add a route to a service or update an existing route. We are running kong version 0.14.0 on 8 (4 core ec2) machines with default configuration. 

On regular periods, the number of connections to database is 4 per node and we see 32 connections to the db overall. But during this issue, the connection goes beyond 2000 in the db. It lasts for less than a minute. During this period, the "active connections" spikes in kong and we see slowness in the calling systems. Below is our configuration, I have attached the kong connection metrics and rds connection metrics as well

<img width="908" alt="screen shot 2018-07-19 at 1 43 26 pm" src="https://user-images.githubusercontent.com/1185338/42930218-0b8456d6-8b5a-11e8-98bd-3e041476b205.png">
<img width="837" alt="screen shot 2018-07-19 at 1 44 23 pm" src="https://user-images.githubusercontent.com/1185338/42930219-0bd38f80-8b5a-11e8-9888-ebef56be9bd9.png">

[...]

PR #3696, opened 2018-08-13
Rebuild the router through events rather than at the access phase 

ganeshs:

### Summary
Made a fix to not rebuild the router during the access phase. The approach I have tried is to trigger a worker_events.post to rebuild on all workers in the local node when there's a change in route. And also subscribe to cluster_events invalidations on 'router:version' to rebuild the router

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix https://github.com/Kong/kong/issues/3634

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -264,6 +264,23 @@ return {
         end
       end, "rebuild", "routes")
 
+      local ok, err = cluster_events:subscribe("invalidations", function(key)
+        log(ngx.DEBUG, "received invalidate event from cluster for key: '", key, "'")
+        if key == "router:version" then
+          log(ngx.DEBUG, "[events] Router updated, rebuilding router in this worker")
+          local version, err = cache:get(key, CACHE_ROUTER_OPTS, utils.uuid)
+          local ok, err = build_router(singletons.db, version)
+          if not ok then
+             router_err = err
+             log(ngx.CRIT, "could not rebuild router: ", err)
+          end
+        end
+      end)
+      if not ok then
+        return nil, "failed to subscribe to invalidations cluster events " ..
+                    "channel: " .. err
+      end
+
       worker_events.register(function(data)
         if data.operation ~= "create" and
            data.operation ~= "delete"
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -239,8 +239,30 @@ return {
       worker_events.register(function()
         log(DEBUG, "[events] Route updated, invalidating router")
         cache:invalidate("router:version")
+        local version, err = cache:get("router:version", CACHE_ROUTER_OPTS, utils.uuid)
+        local ok, err = build_router(singletons.db, version)
+        if not ok then
+           router_err = err
+           log(ngx.CRIT, "could not rebuild router: ", err)
+        end
+        local ok, err = worker_events.post("rebuild", "routes", {
+           version = version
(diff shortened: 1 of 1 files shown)


ganeshs:
@thibaultcha Thanks for your response. Our throughput is pretty high ~50000 rps going through 10+ kong nodes. Addition/Removal of routes happen once or twice a day and the current implementation blocks the incoming requests as the db connections go for a toss. This results in resp times increasing for the callers upto a minute in the worst case. 

 I didn't clearly understand the concerns related to eventual consistency. Are you referring to the event getting lost in some of the workers? I can see this happen when there's an error in processing the callback. I agree this need to be handled. 

> Overall, the proper "fix" here would probably be to allow for the router to receive atomic updates, e.g. being more granular when adding Routes/Services, and not require a complete rebuild for every CRUD operation exercised on these entities. That is an optimization on our TODO list, but we have yet to take care of it yet.
We are kind of blocked on this change. Have applied this patch to our production servers and its been running smooth for the past 2 weeks without any issues. While I understand the patch is far from a proper fix, do you have any suggestions how we can live with this till we have proper fix. Thanks

thibaultcha:
> I didn't clearly understand the concerns related to eventual consistency. Are you referring to the event getting lost in some of the workers?

I am referring to the time it takes for the even to propagate amongst the workers.

> We are kind of blocked on this change. Have applied this patch to our production servers and its been running smooth for the past 2 weeks without any issues.

While this patch may fix your use-case (which is great! yay open source!), it isn't something we can consider on our side given the concerns cited above. There are other ways to reduce the impact made to the DB when applying Services/Routes changes but they are all non-trivial to develop and test.

ganeshs:
> I am referring to the time it takes for the even to propagate amongst the workers.

Sorry my understanding of kong internals is limited. But can't this be controlled by setting polling frequency to 1s? Or are we talking about workers not being consistent during that 1s interval?

While we can live with the patch for interim, is there any timeline that I can wait for an official fix to be released. I would be glad to help out with the fix if we know of an approach.

thibaultcha:
> Or are we talking about workers not being consistent during that 1s interval?

Indeed. Which, as I've said, has been an issue in the past for a number of confused users and caused us a recurring flow of support requests and GitHub issues.

> While we can live with the patch for interim, is there any timeline that I can wait for an official fix to be released.

We're not focusing our efforts on the router at this point, but a few possibilities to consider:

- Use lua-resty-worker-events `poll()` function as previously suggested
- Cache Routes/Services into a shared memory zone (lua_shared_dict/kong.cache) to limit the number of database queries on a rebuild
- Improve the router to accept granular updates of Routes/Services

All of these solutions come with their own pros and cons, and you are welcome to try any of these, but the results would ideally require careful testing and benchmarking (i.e. does the cost of calling `poll()` - which is _usually_ cheap - on the proxy code path is actually worth it, will this even make a difference on a busy node, that is, in 1s, there is little chance that no workers would not process any requests and thus trigger a rebuild from the access phase anyway, how would a shared memory zone cache implementation behave when all of the Services/Routes do not fit in it, etc...).

I will leave this open for discussion for some more time if anyone else wants to jump in, but labelling it as such!

thibaultcha:
Closing this, as I feel like the discussion has reached a stall point, and there has been no development in the last 2 weeks. Hopefully this PR stays visible, along with this thread!

Issue #3609, opened 2018-07-11
Kong Oauth2 backend headers NULL and discussion point.


jeremyjpj0916:
### Summary

When backend receives valid Oauth2 Client, headers that are null should not even be set and passed.

X-Consumer-ID: 8aa31296-a4ec-4cb4-bfc7-64a5428b3c65
**X-Consumer-Custom-ID: userdata: NULL**
X-Consumer-Username: admin
X-Consumer-Groups: c3af0477-4747-4700-b9bc-520d158fdc15, 614c8995-a153-408b-b67d-06f0fc524898

### Steps To Reproduce

1. Create service + route + acl + oauth2 on a proxy and use client credentials flow
2. Inspect the headers backend received on the proxy
3. Notice X-Consumer-Custom-ID (which is not set in Kong) getting weirdly null populated on back-end, I think this to be sort of buggy/bad practice.

One other thing I wanted to bring up, X-Consumer-Groups is not a great header, would kong accept a PR to Oauth2 that disables this header? Think about it. A consumer at scale with an ACL to each proxy(if you do it like that) ends up being 100-200+ uuids thrown in this header, not super performant. Would Kong accept a PR to enable/disable that in the Oauth2 schema?

### Additional Details & Logs
- Kong version 0.14.0

bungle:
`X-Authenticated-Groups`(?), do you mean `X-Authenticated-Scope`?

bungle:
@jeremyjpj0916 I would give a `+` for such PR on `ACL` plugin. I think we also need `ngx.authenticated_groups` that any plugin can set (such as OpenID Connect or LDAP plugins), so that they can also be used together with `ACL` plugin.

jeremyjpj0916:
@bungle yeah I was looking at the OAuth2 code the other day and realized X-Consumer-Groups header had to deal with ACL plguin code and not OAuth2 in my suggestion. So glad to hear you think that would be an appropriate change as each API Provider does not need to know all the groups(which really represent either access to one or many proxies) a member has access to(must be a specific use case yall were thinking about when it was designed). And as for X-Consumer-Custom-ID: userdata: NULL being populated by the OAuth2 plugin you would consider that to be a bug right and a PR to the OAuth2 plugin checking for nil before assignment is appropriate?

bungle:
[...]

PR #3697, opened 2018-08-13
Optional X-Consumer-Groups ACL Header Improvement

jeremyjpj0916:
### Summary

As discussed here : https://github.com/Kong/kong/issues/3609

One of my biggest concerns with Kong's ACL functionality as it stands today is including the arbitrarily growing in size list of a consumers entire ACL groups as a header. In our specific implementation of Kong every route gets is own ACL group as the routes existing uuid, so we can then give consumers access to specific routes by adding that uuid to their ACL, I suspect this will be a popular pattern going forward for many users who implement Kong. 

I propose a boolean that will act as a switch and enable/disable sending the consumer groups as a header, as I suspect for plenty of use cases that the arbitrary group info in the header is not necessary and as it grows in size could one day break an otherwise working proxy by increasing the header size past what a API Provider will accept (since many API providers have a 8kb header size limit by default like Apache Tomcat). I would like a predictable header size going to the API provider that is static when it comes to size generally, and by enabling this functionality on the ACL plugin we can achieve that with Kong.

### Full changelog

1. Modify schema.lua to include new boolean property 'hide_groups_header' defaulted to false as to not break existing functionality

2. Modify handler.lua to account for 'hide_groups_header' boolean setting.

### Issues resolved

Fixes concern for growing Kong header size as ACL groups list grow. Also improves bytes across the wire to exclude potentially extraneous info of a consumers entire ACL group list.

---

Let me know if I am missing anything, not super familiar with the PR process in general but would like to see this functionality in one of Kong's upcoming releases as I have a few Kong consumer accounts at 50+ ACL groups and that extra header is already irking me a bit 😄 .

### DIFF
--- a/kong/plugins/acl/schema.lua
+++ b/kong/plugins/acl/schema.lua
@@ -4,7 +4,8 @@ return {
   no_consumer = true,
   fields = {
     whitelist = { type = "array" },
-    blacklist = { type = "array" }
+    blacklist = { type = "array" },
+    set_groups_header = { type = "boolean", default = true }
   },
   self_check = function(schema, plugin_t, dao, is_update)
     if next(plugin_t.whitelist or {}) and next(plugin_t.blacklist or {}) then
--- a/kong/plugins/acl/schema.lua
+++ b/kong/plugins/acl/schema.lua
@@ -5,7 +5,7 @@ return {
   fields = {
     whitelist = { type = "array" },
     blacklist = { type = "array" },
-    set_groups_header = { type = "boolean", default = true }
+    hide_groups_header = { type = "boolean", default = false }
   },
   self_check = function(schema, plugin_t, dao, is_update)
     if next(plugin_t.whitelist or {}) and next(plugin_t.blacklist or {}) then
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -81,7 +81,9 @@ function ACLHandler:access(conf)
     return responses.send_HTTP_FORBIDDEN("You cannot consume this service")
   end
 
-  set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
+  if conf.set_groups_header then
+    set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
+  end
 end
 
 return ACLHandler
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -81,7 +81,7 @@ function ACLHandler:access(conf)
     return responses.send_HTTP_FORBIDDEN("You cannot consume this service")
   end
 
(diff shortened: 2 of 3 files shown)


kong/plugins/acl/schema.lua
```diff
@@ -4,7 +4,8 @@ return {
   no_consumer = true,
   fields = {
     whitelist = { type = "array" },
-    blacklist = { type = "array" }
+    blacklist = { type = "array" },
+    set_groups_header = { type = "boolean", default = true }
```
thibaultcha: The default for this value should probably respect the already existing behaviour, that is to _include_ the groups header. Thus, `default = false`. Now, to semantically align with the authentication plugins, which provide a `hide_credentials` option (whose default is `false`), this option should probably be updated to `hide_groups_header = { type = "boolean", default = false }`. Thoughts?

jeremyjpj0916: And then just do this right?
```
  if not conf.hide_groups_header then
    set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
  end
```
This works fine for me 👍 , whatever Kong prefers honestly. Glad to see yall are cool with adding this functionality.

thibaultcha: @jeremyjpj0916 Yes, exactly.

jeremyjpj0916:
Added related documentation PR here: https://github.com/Kong/docs.konghq.com/pull/821

jeremyjpj0916:
@thibaultcha We attempted to rebase to master... but doing it without a new PR from the UI. This is not my strong suit, I am not a smart man 😆 ...

jeremyjpj0916:
I think I did a proper testcase here: https://github.com/Kong/kong/pull/3697/commits/e004a02149b3215ee1f8f1548bfc580ab88d878d , I include basic test for if true then make sure that header value equals nil, I also include a if false make sure the header is still accounted for.

jeremyjpj0916:
Based on my reading here it seems its true you cannot change the source branch of a PR: https://stackoverflow.com/questions/42381557/changing-source-branch-for-pr-on-github , why don't I just close this one unimplemented and make a newer one since that seems easiest?

thibaultcha:
@jeremyjpj0916 You can change the base branch for your PR if you click on the "edit" button on top of it. You can then chose "master", and rebase/cherry-pick/etc... your commit(s) on top of the master branch, and force push the branch. Or reopen a PR is fine too if that's too much trouble.

jeremyjpj0916:
Sorry to disappoint man 😅 , but I do understand the latter half method, sorry for the extra overhead closing and opening a new one adds but its easiest on me at least.

image

All I see to do so in UI is to point to different target. Lets give this one more go on a diff PR. I assure it will be perfecto!

thibaultcha:
Yeah that is the right button

PR #3698, opened 2018-08-13
Get_last_failure does not catch errors such as POST, PATCH, which can…

kingljl:
… affect the retry mechanism because the default value of proxy_next_upstream, error timeout, does not contain non_idempotent

The following is the explanation of proxy_next_upstream.
proxy_next_upstream
Specifies in which cases a request should be passed to the next server:
Default:
proxy_next_upstream error timeout;

non_idempotent
normally, requests with a non-idempotent method (POST, LOCK, PATCH) are not passed to the next server if a request has been sent to an upstream server (1.9.13); enabling this option explicitly allows retrying such requests;

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX

### DIFF
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -59,6 +59,7 @@ init_worker_by_lua_block {
     Kong.init_worker()
 }
 
+proxy_next_upstream timeout error  non_idempotent;
 
 > if #proxy_listeners > 0 then
 upstream kong_upstream {
@@ -81,7 +82,7 @@ server {
     error_log ${{PROXY_ERROR_LOG}} ${{LOG_LEVEL}};
 
     client_body_buffer_size ${{CLIENT_BODY_BUFFER_SIZE}};
-
+    
 > if proxy_ssl_enabled then
     ssl_certificate ${{SSL_CERT}};
     ssl_certificate_key ${{SSL_CERT_KEY}};


thibaultcha:
@kingljl Thank you for the patch! Catching failures on unsafe HTTP methods requests should definitely happen. However, I don't think we wish to enforce the best practice provided by Nginx (rightfully so) which is to not retry such requests... Overall, this PR would need some regression tests proving this behaviour.

Ping @hishamhm @Tieske - this feels like an important discussion to have.

Tieske:
iirc correctly the failures are still caught in the logging phase.

This change is dangerous though, since we do not want non-idempotent requests to be retried by default. When the retries were implemented it was a deliberate choice to stick to the existing nginx directive (from a MVP feature perspective).

The proper way to do this is to recreate the nginx retry logic in Kong. Copy the configuration option into the `service` entity, such that they are configurable on a per service basis (effectively have nginx retry _every_ request, and have Kong decide when to stop).

thibaultcha:
@hishamhm I was suggesting a test that asserts the upstream server (Target )is marked as unhealthy when a non-idempotent request fails (i.e. not any logging plugin, but specifically the healthcheck core component). If anything mostly to ensure we do not experience any regression in this scenario and demonstrate to users such as @kingljl that such failures are properly caught by the balancer/healthchecks.

It seems like this is already the case anyway (I originally misinterpreted the patch as implementing the behavior described in my above paragraph). And retrying non-idempotent requests is a no-no at this time (unanimously). Let's close this then!

@kingljl Thank you for the suggestion; I believe the thread to be insightful enough as to why we unfortunately cannot incorporate it.

PR #3699, opened 2018-08-13
fix(hmac-auth) add deprecation transition for pr #3339

### LINKED ISSUE #3672: PR #3339 in Kong 0.13.1 breaks existing hmac auth clients
PR #3339 breaks existing hmac auth clients

Before Kong 0.13.1, Kong calculated signatures using `request-line` that did not include the query string. This Pull Request released in 0.13.1 significantly changes the signature calculation to include the query string which breaks existing clients. What is the recommended approach to upgrade to Kong 0.13.1 without breaking existing clients? There doesn’t appear to be a way to configure Kong allow both signing algorithms for an appropriate transition period.

I’m in the process of upgrading our clusters to 0.13.1 but this is a major breaking change which breaks all existing clients so we’re basically stuck on Kong 0.13.0 or below because the older Kong versions don’t support the new algorithm and Kong 0.13.1+ doesn’t support the old algorithm. “Cutting-over" the new signing algorithm is basically impossible to coordinate with all existing clients simultaneously without incurring significant amounts of downtime.

Please help

mlehner616:
### Summary
Add deprecation transition for PR #3339 by adding support for older signatures generated from `request-line` without the query_string. This is intended to be a patch on 0.13.1 but may apply to 0.14+ although I haven't tested this. 

### Full changelog
* add deprecation hmac-auth transition for pr #3339 

### Issues resolved

Fix #3672

### DIFF
--- a/kong/plugins/hmac-auth/access.lua
+++ b/kong/plugins/hmac-auth/access.lua
@@ -190,12 +190,13 @@ end
 
 local function validate_signature(request, hmac_params, headers)
   local signature_1            = create_hash(request, hmac_params, headers)
-  local signature_1_deprecated = create_hash_deprecated(request, hmac_params, headers)
   local signature_2            = ngx_decode_base64(hmac_params.signature)
 
   if signature_1 == signature_2 then
     return true
   else
+    local signature_1_deprecated = create_hash_deprecated(request, hmac_params, headers)
+
     return signature_1_deprecated == signature_2
   end
 end
--- a/kong/plugins/hmac-auth/access.lua
+++ b/kong/plugins/hmac-auth/access.lua
@@ -157,11 +157,47 @@ local function create_hash(request, hmac_params, headers)
   return hmac[hmac_params.algorithm](hmac_params.secret, signing_string)
 end
 
-local function validate_signature(request, hmac_params, headers)
-  local signature_1 = create_hash(request, hmac_params, headers)
-  local signature_2 = ngx_decode_base64(hmac_params.signature)
+-- plugin assumes the request parameters being used for creating
+-- signature by client are not changed by core or any other plugin
+-- DEPRECATED BY: https://github.com/Kong/kong/pull/3339
+local function create_hash_deprecated(request, hmac_params, headers)
+  local signing_string = ""
+  local hmac_headers = hmac_params.hmac_headers
+  local count = #hmac_headers
 
-  return signature_1 == signature_2
+  for i = 1, count do
+    local header = hmac_headers[i]
+    local header_value = headers[header]
+
+    if not header_value then
+      if header == "request-line" then
+        -- request-line in hmac headers list
+        local request_line = fmt("%s %s HTTP/%s", ngx.req.get_method(), ngx.var.uri, ngx.req.http_version())
+
+        signing_string = signing_string .. request_line
+      else
+        signing_string = signing_string .. header .. ":"
+      end
+    else
+      signing_string = signing_string .. header .. ":" .. " " .. header_value
+    end
+    if i < count then
+      signing_string = signing_string .. "\n"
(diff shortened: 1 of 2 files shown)


thibaultcha:
@mlehner616 Thanks for taking the initiative here!

@shashiranjan84 Will you help review the proposed patch for this behaviour? Thanks!

kong/plugins/hmac-auth/access.lua
```diff
@@ -157,11 +157,48 @@ local function create_hash(request, hmac_params, headers)
...
+      signing_string = signing_string .. "\n"
+    end
+  end
+  return hmac[hmac_params.algorithm](hmac_params.secret, signing_string)
+end
+
 local function validate_signature(request, hmac_params, headers)
-  local signature_1 = create_hash(request, hmac_params, headers)
-  local signature_2 = ngx_decode_base64(hmac_params.signature)
+  local signature_1            = create_hash(request, hmac_params, headers)
```
shashiranjan84: we don't need extra space before assignment operator.

kong/plugins/hmac-auth/access.lua
```diff
@@ -157,11 +157,48 @@ local function create_hash(request, hmac_params, headers)
...
+  local signature_2            = ngx_decode_base64(hmac_params.signature)
```
shashiranjan84: ditto

shashiranjan84: Also we don't need to create both the signature at once, one should be fall back and only be created if needed.

mlehner616: @shashiranjan84 Whitespace removed, My commit from yesterday should have included the deprecated signature creation into the conditional. This looks like it would do what you're describing.

mlehner616:
@shashiranjan84 I made the changes you requested, I haven't had a chance to write tests for this yet.

kong/plugins/hmac-auth/access.lua
shashiranjan84: You are right. Thanks again for your effort, we just need add test, which would be very similar to https://github.com/Kong/kong/blob/master/spec/03-plugins/20-hmac-auth/03-access_spec.lua#L1150, you would just need to create request signature without query-param

mlehner616: Test Added

shashiranjan84:
@mlehner616 thanks, we will review it.

spec/03-plugins/20-hmac-auth/03-access_spec.lua
```diff
@@ -1080,6 +1080,47 @@ for _, strategy in helpers.each_strategy() do
         assert.res_status(200, res)
       end)
 
+      it("should pass with GET with request-line having query param but signed without query param", function()
```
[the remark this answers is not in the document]
mlehner616: done

mlehner616:
@thibaultcha @shashiranjan84  Thanks for the feedback. I've committed changes that address your latest suggestions. Let me know if there are any other tweaks needed. Is there any chance this PR will get backported to a 0.13.2+ release? We're actually blocked on the 0.14 versions right now since we're not ready to drop API support.

thibaultcha:
@mlehner616 We do not have plans for a 0.13.2 version so far.

> not ready to drop API support.

Not sure what you mean by that. In 0.14, the API entity is still supported (`/apis` on the Admin API). or are you talking about other specific changes in the Admin API itself?

mlehner616:
Sorry, I thought I remember seeing 0.14 dropped API entities so it looks like I misread that section. Thanks for the clarification, I'll wait for the release to get cut and give it a shot.

thibaultcha:
@mlehner616 Great, thank you!

PR #3703, opened 2018-08-14
Optional X-Consumer-Groups ACL Header Improvement

jeremyjpj0916:
### Summary

As discussed here : https://github.com/Kong/kong/issues/3609

One of my biggest concerns with Kong's ACL functionality as it stands today is including the arbitrarily growing in size list of a consumers entire ACL groups as a header. In our specific implementation of Kong every route gets is own ACL group as the routes existing uuid, so we can then give consumers access to specific routes by adding that uuid to their ACL, I suspect this will be a popular pattern going forward for many users who implement Kong. 

I propose a boolean that will act as a switch and enable/disable sending the consumer groups as a header, as I suspect for plenty of use cases that the arbitrary group info in the header is not necessary and as it grows in size could one day break an otherwise working proxy by increasing the header size past what a API Provider will accept (since many API providers have a 8kb header size limit by default like Apache Tomcat). I would like a predictable header size going to the API provider that is static when it comes to size generally, and by enabling this functionality on the ACL plugin we can achieve that with Kong.

### Full changelog

1. Modify schema.lua to include new boolean property 'hide_groups_header' defaulted to false as to not break existing functionality
2. Modify handler.lua to account for 'hide_groups_header' boolean setting.
3. Modify /spec/03-plugins/19-acl/02-access_spec.lua for false and true testcase. 

PR to Documentation: https://github.com/Kong/docs.konghq.com/pull/822

### Issues resolved

Fixes concern for growing Kong header size as ACL groups list grow. Also improves bytes across the wire to exclude potentially extraneous info of a consumers entire ACL group list.

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -81,7 +81,9 @@ function ACLHandler:access(conf)
     return responses.send_HTTP_FORBIDDEN("You cannot consume this service")
   end
 
-  set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
+  if not conf.hide_groups_header then
+    set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
+  end
 end
 
 return ACLHandler
--- a/kong/plugins/acl/schema.lua
+++ b/kong/plugins/acl/schema.lua
@@ -4,7 +4,8 @@ return {
   no_consumer = true,
   fields = {
     whitelist = { type = "array" },
-    blacklist = { type = "array" }
+    blacklist = { type = "array" },
+    hide_groups_header = { type = "boolean", default = false }
   },
   self_check = function(schema, plugin_t, dao, is_update)
     if next(plugin_t.whitelist or {}) and next(plugin_t.blacklist or {}) then
--- a/kong/plugins/acl/schema.lua
+++ b/kong/plugins/acl/schema.lua
@@ -5,7 +5,7 @@ return {
   fields = {
     whitelist = { type = "array" },
     blacklist = { type = "array" },
-    hide_groups_header = { type = "boolean", default = false }
+    hide_groups_header = { type = "boolean", default = false },
   },
   self_check = function(schema, plugin_t, dao, is_update)
     if next(plugin_t.whitelist or {}) and next(plugin_t.blacklist or {}) then
--- a/spec/03-plugins/19-acl/02-access_spec.lua
+++ b/spec/03-plugins/19-acl/02-access_spec.lua
@@ -221,6 +221,44 @@ for _, strategy in helpers.each_strategy() do
           anonymous = anonymous.id,
         }
       }
+          
+      local route9 = bp.routes:insert {
+        hosts = { "acl9.com" },
+      }
+
+      bp.plugins:insert {
+        name     = "acl",
+        route_id = route9.id,
+        config   = {
+          whitelist = "admin",
+          hide_groups_header = true    
+        }
+      }
+
+      bp.plugins:insert {
+        name     = "key-auth",
+        route_id = route9.id,
+        config   = {}
+      }
+          
+      local route10 = bp.routes:insert {
+        hosts = { "acl10.com" },
+      }
+
+      bp.plugins:insert {
+        name     = "acl",
+        route_id = route10.id,
+        config   = {
+          whitelist = "admin",
+          hide_groups_header = false    
+        }
+      }
+
+      bp.plugins:insert {
+        name     = "key-auth",
+        route_id = route10.id,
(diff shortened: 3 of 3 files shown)


kong/plugins/acl/schema.lua
```diff
@@ -4,7 +4,8 @@ return {
   no_consumer = true,
   fields = {
     whitelist = { type = "array" },
-    blacklist = { type = "array" }
+    blacklist = { type = "array" },
+    hide_groups_header = { type = "boolean", default = false }
```
thibaultcha: style: will you please add a trailing comma at the end of this line? This is a part of the codebase that isn't yet compliant with our formalized codestyle, but it allows to reduce the size of the diff when adding new fields in the future. Thanks!

jeremyjpj0916: Done! Thx.

thibaultcha:
@jeremyjpj0916 Awesome! Thank you!

jeremyjpj0916:
Glad you took the initiative to squeeze this in so soon! Thanks a ton.

kong/plugins/acl/handler.lua
```diff
@@ -81,7 +81,9 @@ function ACLHandler:access(conf)
     return responses.send_HTTP_FORBIDDEN("You cannot consume this service")
   end
 
-  set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
+  if not conf.hide_groups_header then
+    set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
+  end
```
Tieske: In the case when we're not setting the header, we should set a dummy value in the cache. We're trying to save bandwidth in the request, but we should also reduce Kong memory pressure.

so above do:
```lua
    if to_be_blocked == false then
       -- we're allowed, so go and convert 'false' to the header value, if we need it.
       -- if we don't need it, set a dummy value to save memory for potentially very long strings
       to_be_blocked = conf.hide_groups_header and "" or table_concat(consumer_groups, ", ")
     end
```

kong/plugins/acl/schema.lua
```diff
@@ -4,7 +4,8 @@ return {
   no_consumer = true,
   fields = {
     whitelist = { type = "array" },
-    blacklist = { type = "array" }
+    blacklist = { type = "array" },
+    hide_groups_header = { type = "boolean", default = false },
```
Tieske: this needs another PR against `next`, to convert existing records from `nil` to `false`. Since the schema says the default is `false`, it is expected to always be a boolean value, and not an absent one. To keep the data in line with the schema.

kong/plugins/acl/handler.lua
jeremyjpj0916: This seems nice as a further improvement. Since this has already been merged to master branch would the proper approach be to submit a new PR with this additional enhancement? Didn't know you could do  *true* and "" or *table of consumers_groups* would cause to_be_blocked  == "".

kong/plugins/acl/schema.lua
jeremyjpj0916: Is it problematic for this to go into the 0.14.1 minor version as is? I think in my testing of custom plugins the addition of new schema fields are treated as nil which will at least produce the desired results until converting nil to false right? For some reason initially I was thinking Kong core recognized those defaults and would take action on populating them by default in all plugins it sees, maybe a nice future Kong enhancement to automate that process.

image

thibaultcha:
@Tieske All in for the flag to prevent the header from being constructed. 

Regarding the migration, there has been some internal discussions in the last week. We might reconsider our policy to provide migrations for schema changes. In any case, I did not want to bother Jeremy with contributing a migration just yet anyway, especially if there might not be the need for it. We should have a chat.

thibaultcha:
@jeremyjpj0916 Yes, this is indeed scheduled to be part of 0.14.1, no problem. 

If you are able to provide the subsequent improvement suggested by @Tieske by then (before next week), that’d be great, but not necessary.

jeremyjpj0916:
@thibaultcha I'm wfh and have a few mins to tackle this before next meeting. Question is would "" be better for the setting or nil on mem 😆? 

```
to_be_blocked = conf.hide_groups_header and "" or table_concat(consumer_groups, ", ")
```

Or 

```
to_be_blocked = conf.hide_groups_header and nil or table_concat(consumer_groups, ", ")
```
If we want to get picky with it hah.

thibaultcha:
Strings in Lua are interned, so `””` is probably fine. But `nil` might be more appropriate to carry the intended meaning here; I’d probably go for `nil`.

jeremyjpj0916:
👍 roger. Thx! PR for that incoming to master 2 mins.

PR #3706, opened 2018-08-15
fix(ci) adjust deploy condition per travis docs

hutchic:
the travis docs are a little ambiguous :(

travis uses travis-conditional for `if` logic that's applies to builds, stages and jobs but the deploy job `on` conditional has it's own way of working.

Since we're using `if` per the `travis-conditional` library documentation

```
The following known attributes are available:

type (the current event type, known event types are: push, pull_request, api, cron)
```

**References**
https://github.com/travis-ci/travis-conditions
https://docs.travis-ci.com/user/deployment#conditional-releases-with-on

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ stages:
   - pdk tests
   - test
   - name: deploy
-    if: tag =~ 0. || TRAVIS_EVENT_TYPE == cron
+    if: tag =~ 0. || type == cron
 
 jobs:
   include:


thibaultcha:
@hutchic Please use the `chore` commit message for changes to the CI/build mechanism/etc... _Even_ for fixes. Same for _docs_ and typo fixes. The idea here is to be able to browse the git history easily and be assured that feat/fix/hotfix/perf commits are related to the code itself, and ignore all other scopes. This helps a lot for git archeology, changelog, etc... Thanks!

PR #3707, opened 2018-08-15
Optional X-Consumer-Groups ACL Header Improvement, secondary memory footprint enhancement

jeremyjpj0916:
### Summary

Update to reduce Kong memory pressure when ACL X-Consumer-Groups header has been hidden as suggested by @Tieske . Related to Original PR https://github.com/Kong/kong/pull/3703

### Full changelog

* Modify kong/plugins/acl/handler.lua to_be_blocked = nil if conf.hide_groups_header enabled.

### Issues resolved
Improve memory footprint with ACL plugin.

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -71,7 +71,7 @@ function ACLHandler:access(conf)
     if to_be_blocked == false then
       -- we're allowed, so go and convert 'false' to the header value, if we need it.
       -- if we don't need it, set a dummy value to save memory for potentially very long strings
-      to_be_blocked = conf.hide_groups_header and nil or table_concat(consumer_groups, ", ")
+      to_be_blocked = conf.hide_groups_header and "" or table_concat(consumer_groups, ", ")
     end
 
     -- update cache
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -69,8 +69,9 @@ function ACLHandler:access(conf)
     end
 
     if to_be_blocked == false then
-      -- we're allowed, so go and convert 'false' to the header value
-      to_be_blocked = table_concat(consumer_groups, ", ")
+      -- we're allowed, so go and convert 'false' to the header value, if we need it.
+      -- if we don't need it, set a dummy value to save memory for potentially very long strings
+      to_be_blocked = conf.hide_groups_header and nil or table_concat(consumer_groups, ", ")
     end
 
     -- update cache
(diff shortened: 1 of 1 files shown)


kong/plugins/acl/handler.lua
```diff
@@ -69,8 +69,9 @@ function ACLHandler:access(conf)
     end
 
     if to_be_blocked == false then
-      -- we're allowed, so go and convert 'false' to the header value
-      to_be_blocked = table_concat(consumer_groups, ", ")
+      -- we're allowed, so go and convert 'false' to the header value, if we need it.
+      -- if we don't need it, set a dummy value to save memory for potentially very long strings
+      to_be_blocked = conf.hide_groups_header and nil or table_concat(consumer_groups, ", ")
```
Tieske: no, we cannot set it to `nil`, because the resulting value is cached, and in the cache `nil` means "nothing is cached" (see line 62). The tests all pass, because it will just regenerate every time, without ever caching the results.

You can also leave it at the `false` value it already has. In the example I used an empty string because it gets interned, and because it matches the comments on line 59-60. If you don't use an empty string, then please also update the comments at 59-60.

thibaultcha: Nice catch. Hadn't looked at this yet, but indeed, this cannot be set to `nil`. Using `false` or `""` is a better fit in this context!

jeremyjpj0916: Fixed, thanks @Tieske for the further details, I should have guessed you considered nil and knew the implications of it! Back to "" 👍  .

kong/plugins/acl/handler.lua
```diff
@@ -69,8 +69,9 @@ function ACLHandler:access(conf)
     end
 
     if to_be_blocked == false then
-      -- we're allowed, so go and convert 'false' to the header value
-      to_be_blocked = table_concat(consumer_groups, ", ")
+      -- we're allowed, so go and convert 'false' to the header value, if we need it.
+      -- if we don't need it, set a dummy value to save memory for potentially very long strings
+      to_be_blocked = conf.hide_groups_header and "" or table_concat(consumer_groups, ", ")
```
thibaultcha: @jeremyjpj0916 Will you bring these lines back to being under 80 chars long? Thanks!

jeremyjpj0916:
```
      -- we're allowed, convert 'false' to the header value, if needed.
      -- if not needed, set dummy value to save mem for potential long strings.
      to_be_blocked = conf.hide_groups_header and "" 
                      or table_concat(consumer_groups, ", ")
```
Tried to shorten wording and newline the long line of code properly, this work or any further change desired?

thibaultcha:
@jeremyjpj0916 Thanks!

PR #3708, opened 2018-08-15
docs(readme) DRY License

coopr:
DRY License by linking to it, rather than duplicating part of it.


coopr:
Nevermind, closing, going to do a different adjustment in a different PR...

Issue #3287, opened 2018-03-12
BUG: Client(Chrome)->Load Balancer(F5)->Kubernetes 1.5-Kong Gateway(Upstream Targets)->Biz(HttpREST) ???


taotao2014:
BUG: Client->Load Balancer(F5)->Kubernetes 1.5-Kong Gateway(Upstream Targets)->Biz(HttpREST) ???

Client error, request sometimes ok, sometimes bad
1.ERR_CONNECTION_RESET
2.ERR_CONNECTION_ABORTED
3.HTTP 504
4.HTTP 404

 Clinet->Kong Gateway(Upstream Targets)->Biz(HttpREST),Do not pass F5 without any errors.

WHY??? help....

-Kubernetes: 1.5
- Kong version :0.12.1
- OS: Linux version 3.10.0-693.11.1.el7.x86_64 (builder@kbuilder.dev.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-16) (GCC) ) #1 SMP Mon Dec 4 23:52:40 UTC 2017
- Kong error logs :

error list:
2018/03/12 19:26:38 [error] 29#0: *14715969 recv() failed (104: Connection reset by peer) while reading response header from upstream, client: 10.129.5.51, server: kong, request: "GET /TenantBaseDimissionMRest/100000/User HTTP/1.1", upstream: "http://10.129.8.15:6046/TenantBaseDimissionMRest/100000/User", host: "cloud.abc.link", referrer: "http://cloud.abc.link/TenantBaseDimissionMRest/api-docs/index.html"
2018/03/12 19:26:38 [error] 29#0: *14715969 recv() failed (104: Connection reset by peer) while reading response header from upstream, client: 10.129.5.51, server: kong, request: "GET /TenantBaseDimissionMRest/100000/User HTTP/1.1", upstream: "http://10.129.8.15:6046/TenantBaseDimissionMRest/100000/User", host: "cloud.abc.link", referrer: "http://cloud.abc.link/TenantBaseDimissionMRest/api-docs/index.html"
[...]

PR #3716, opened 2018-08-18
Feature: Allow keepalive to be disabled when set to 0

pryorda:
### Summary

Allow Engineers to disable keeaplive by setting upstream_keepalive to 0. In our infrastructure we did not want to have nginx workers keeping the backend connections alive that were to externally lb services. This fixed various connection resets for us. 

### Full changelog

* Implemented ability to disable upstream_keepalive

### Issues resolved

Possibly: #3287

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -213,6 +213,9 @@
                                  # preserved in the cache of each worker
                                  # process. When this number is exceeded, the
                                  # least recently used connections are closed.
+                                 # A value of `0` will disable this behavior 
+                                 # altogether, forcing each upstream request 
+                                 # to open a new connection
 
 #headers = server_tokens, latency_tokens
                                  # Specify Kong-specific headers that should
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -66,7 +66,9 @@ upstream kong_upstream {
     balancer_by_lua_block {
         Kong.balancer()
     }
+> if upstream_keepalive > 0 then 
     keepalive ${{UPSTREAM_KEEPALIVE}};
+> end
 }
 
 server {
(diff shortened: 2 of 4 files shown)


pryorda:
Let me know if I missed something. This is my first lua pull.

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+# Would be awesome to see this use docker-compose and build a fully
+# functional test environment
+
+COPY . /build
+
+WORKDIR /build
+RUN apk add --no-cache openssl openssl-dev git
+RUN make dev && make test
+
```
thibaultcha: Please remove unrelated changes from this PR, such as this Dockerfile and the Makefile rule added below, along with the README change. Thank you!

kong/templates/nginx_kong.lua
```diff
@@ -66,7 +66,9 @@ upstream kong_upstream {
     balancer_by_lua_block {
         Kong.balancer()
     }
+> if upstream_keepalive > 0 then 
     keepalive ${{UPSTREAM_KEEPALIVE}};
+> end
```
thibaultcha: We would need you to provide a test for this behaviour. Nginx config template test can be found here: https://github.com/Kong/kong/blob/master/spec/01-unit/003-prefix_handler_spec.lua#L55

kong.conf.default
```diff
@@ -213,6 +213,8 @@
                                  # preserved in the cache of each worker
                                  # process. When this number is exceeded, the
                                  # least recently used connections are closed.
+                                 # 0 will remove the entry conf nginx-kong.conf
+                                 # when using the default template
```
thibaultcha: Will you please update this description to something like:
```
A value of `0` will disable this behavior altogether,
forcing each upstream request to open a new connection.
```

`nginx-kong.conf` and "the default template" are implementation details that should not be mentioned here. Additionally, the proposed text is more explicit as to what the behavior is when the given value is `0`.

pryorda:
@thibaultcha Should be good to go now. Let me know if I missed anything or did it wrong.

thibaultcha:
@pryorda Manually merged to next with a few edits, thanks!

Will you be able to provide documentation for this change as well? Thank you!

pryorda:
@thibaultcha Created here: https://github.com/Kong/docs.konghq.com/pull/827

Thanks

PR #3717, opened 2018-08-20
feat(file-log) add log_level to file-log plugin

af-inet:
Implements a log_level concept in the file-log plugin, similar to syslog or loggly.

## Summary

Use Case: if you rely on the file-log plugin, and you'd like to log only errors from Kong (such is the case at my company).

## Question / Discussion

Maybe this should be a separate plugin, to preserve compatibility ?

### DIFF
--- a/kong/plugins/file-log/handler.lua
+++ b/kong/plugins/file-log/handler.lua
@@ -24,6 +24,17 @@ int write(int fd, const void * ptr, int numbytes);
 -- fd tracking utility functions
 local file_descriptors = {}
 
+local LOG_LEVELS = {
+  debug = 7,
+  info = 6,
+  notice = 5,
+  warning = 4,
+  err = 3,
+  crit = 2,
+  alert = 1,
+  emerg = 0
+}
+
 -- Log to a file. Function used as callback from an nginx timer.
 -- @param `premature` see OpenResty `ngx.timer.at()`
 -- @param `conf`     Configuration table, holds http endpoint details
@@ -55,7 +66,18 @@ local function log(premature, conf, message)
     end
   end
 
-  ffi.C.write(fd, msg, #msg)
+  local severity
+  if message.response.status >= 500 then
+    severity = conf.server_errors_severity
+  elseif message.response.status >= 400 then
+    severity = conf.client_errors_severity
+  else
+    severity = conf.successful_severity
+  end
+
+  if LOG_LEVELS[severity] <= LOG_LEVELS[conf.log_level] then
+    ffi.C.write(fd, msg, #msg)
+  end
 end
 
 local FileLogHandler = BasePlugin:extend()
--- a/kong/plugins/file-log/schema.lua
+++ b/kong/plugins/file-log/schema.lua
@@ -1,5 +1,6 @@
 local pl_file = require "pl.file"
 local pl_path = require "pl.path"
+local ALLOWED_LEVELS = { "debug", "info", "notice", "warning", "err", "crit", "alert", "emerg" }
 
 local function validate_file(value)
   -- create file in case it doesn't exist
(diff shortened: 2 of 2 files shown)


kikito:
Hi,

Thanks a lot for taking the time to send this PR.

I think it does something simple with more ceremony than it needs to. If I'm understanding the code correctly, the end result of this plugin can only be 4 options:

* Server errors are logged if `conf.level` >= `conf.server_errors_severity`
* Client errors are logged if `conf.level` >= `client_errors_severity`
* The remaining responses are logged if `conf.log_level` >= `conf.successful_severity`

Introducing the concept of "log levels" in order to do this filter doesn't seem to be necessary to me. I would rather code the filter directly instead.

I think if you are logging client errors, you probably also want to log server errors. And if you are logging successes you probably also want to log errors. So I would change this to a single variable:

``` lua
local ALLOWED_LEVELS = { "everything", "errors", "server_errors" }

fields = {
   allowed = { type = "string", enum = ALLOWED_LEVELS, default = "everything" },
```

This would allow to log all requests (`everything`), client and server errors (`errors`) and server errors only (`server_errors`). We could add a couple more possible values there if more granularity is needed, but I think this would be it.

thibaultcha:
Thanks @kikito for the provided suggestions here. On top of that, I would also stress the importance of distinguishing downstream/Kong/upstream errors. Is `server_errors` going to log upstream 5xx errors, or Kong-generated 5xx errors as well? What about `client_errors` (more explicit than `errors`), should they include client requests 4xx, or Kong to upstream 4xx errors as well (e.g. a Kong transformation resulting in an invalid upstream request)? What if I wish to only log upstream and downstream errors, and ignore Kong-generated ones?

In my opinion, introducing the behaviour seeked by this patch raises the above questions in users' mind, and we probably should think about them ahead of time.

Thoughts?

af-inet:
@kikito @thibaultcha thanks for your feedback.

> I think it does something simple with more ceremony than it needs to. If I'm understanding the code correctly, the end result of this plugin can only be 4 options:

I was originally aiming for consistency with the other plugins that implement similar functionality, but if that's not a priority than I 100% agree - the interface would be a lot cleaner with something like you've described.

To combine this idea with @thibaultcha suggestion (regarding upstream/downstream/kong errors), imagine a schema that looks more like this

```
fields = {
  log_client_errors = { type = "boolean", default = true },
  log_server_errors = { type = "boolean", default = true },
  log_upstream_errors = { type = "boolean", default = true },
  log_downstream_errors = { type = "boolean", default = true },
  log_kong_errors = { type = "boolean", default = true },
  ...
}
```

I think this gives the user a clear & flexible interface.

I'm happy to refactor this PR if ya'll agree, thanks!

af-inet:
I would also note that with my above suggestion...

you can imagine more granular filters which would be impossible to implement

   for example: "only log kong server errors & upstream client errors"

...but I think this is an inherent tradeoff between convenience vs flexibility, and this limited set of options is a good balance.

thibaultcha:
> I was originally aiming for consistency with the other plugins that implement similar functionality

@af-inet Which plugins and what functionalities of these plugins are you referring to exactly?

> imagine a schema that looks more like this

In the schema you proposed, I am having a hard time distinguishing the following:

- client_errors vs. downstream_errors
- server_errors vs. kong_errors or upstream_errors

Would you mind clarifying?

af-inet:
@thibaultcha 

> @af-inet Which plugins and what functionalities of these plugins are you referring to exactly?

source: https://docs.konghq.com/plugins/syslog/

source: https://docs.konghq.com/plugins/loggly/

Notice both plugins implement the following config values:

```
config.successful_severity
config.client_errors_severity
config.server_errors_severity
config.log_level
```

> In the schema you proposed, I am having a hard time distinguishing the following (...)

client_errors = `message.response.status >= 400 && message.response.status < 500`

server_errors = `message.response.status >= 500`

I think I've misunderstood your original question regarding kong_errors vs. upstream_errors vs. downstream_errors, so I'm not 100% sure how to address this - perhaps this snippet from the file-log documentation sheds some light?

```
This logging plugin will only log HTTP request and response data. If you are looking for the Kong process error file (which is the nginx error file), then you can find it at the following path: {prefix}/logs/error.log
```

source: https://docs.konghq.com/plugins/file-log

> Is server_errors going to log upstream 5xx errors, or Kong-generated 5xx errors as well?

I believe both.

> What about client_errors (more explicit than errors), should they include client requests 4xx, or Kong to upstream 4xx errors as well (e.g. a Kong transformation resulting in an invalid upstream request)?

I haven't tested this, but I'm guessing that in both cases `message.response.status >= 400 && message.response.status < 500` will be `true`.

> What if I wish to only log upstream and downstream errors, and ignore Kong-generated ones?

I don't think there's a distinction made in the file-log plugin, the only distinction made is what status code we're sending in the response.

PR #3723, opened 2018-08-23
refactor(db) define default/max page sizes as constants

thibaultcha:


kong/db/strategies/postgres/init.lua
```diff
@@ -509,8 +509,6 @@ end
 
 
 local function page(self, size, token, foreign_key, foreign_entity_name)
-  size = min(size or 100, 1000)
```
thibaultcha: Arguments already went through sanitization in the DAO layer before hitting the strategy (deliberately, as per our our design pattern).

PR #3729, opened 2018-08-28
feat(aws-lambda) add static response template to optionnaly match

### LINKED ISSUE #3208: AWS Lambda Handled Exceptions
### Summary

Currently Kong AWS Lambda plugin supports only Unhanded exceptions, that is errors caught by Lambda engine,  mapped via `config.unhandled_status`, which is sufficient for this case. 

At the same time, there is Handled exception, that is error returned by lambda code itself and something that might return custom error in text or JSON body. This is handled as well via `X-Amz-Function-Error` header. 

I propose we introduce additional config option, `config.handled_status_regexp`, if configured, that would try to match body, end extract status code. I think regexp is better as it will support different content types other then just JSON.

I'm trying to test if you going to accept such a PR and/or start discussion.

I'll do all the work.

### Steps To Reproduce

Return Handled Error from lambda, it will go with default flow.

RomainDeSaJardim:
### Summary

Allow lambda's response to set statusCode and add headers to Kong's output response.

If Content-Type is "application/json" and response from AWS Lambda match the following JSON :
{
  status_code: <my_status_code>,
  body: <my_json_encoded_body>,
  headers: <my_headers>
}
Then the response from Kong will set the statusCode value to <my_status_code> and the body to <my_json_encoded_body> and add <my_headers> to response

### Full changelog

* `kong/plugins/aws-lambda/handler.lua`

### Issues resolved

Fix #3208

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -21,7 +21,6 @@ local ngx_req_get_uri_args = ngx.req.get_uri_args
 local ngx_req_get_headers  = ngx.req.get_headers
 local ngx_encode_base64    = ngx.encode_base64
 
-
 local new_tab
 do
   local ok
@@ -179,6 +178,7 @@ function AWSLambdaHandler:access(conf)
   end
 
   local content = res:read_body()
+
   local headers = res.headers
 
   local ok, err = client:set_keepalive(conf.keepalive)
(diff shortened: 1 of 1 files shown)


thibaultcha:
Hi,

Thank you for the patch. However, it seems like it is a duplicate of #3427?

RomainDeSaJardim:
Oh ... Sorry I missed that. I close this PR

RomainDeSaJardim:
Oh ... Sorry I missed that. I close this PR

PR #3730, opened 2018-08-28
chore(deps) bump DNS lib to 2.2.0, introducing ttl-override

Tieske:
This bump fixes a number of minor issues (see
https://github.com/Kong/lua-resty-dns-client#history), and
introduces a new option to override the DNS record ttl.

### DIFF
--- a/kong-0.14.1-0.rockspec
+++ b/kong-0.14.1-0.rockspec
@@ -28,7 +28,7 @@ dependencies = {
   "luaossl == 20171028",
   "luasyslog == 1.0.0",
   "lua_pack == 1.0.5",
-  "lua-resty-dns-client == 2.1.0",
+  "lua-resty-dns-client == 2.2.0",
   "lua-resty-worker-events == 0.3.3",
   "lua-resty-mediador == 0.1.2",
   "lua-resty-healthcheck == 0.5.0",
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -486,6 +486,11 @@
                                  # specified name). The format is a (case
                                  # insensitive) comma separated list.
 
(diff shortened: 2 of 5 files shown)


kong.conf.default
```diff
@@ -486,6 +486,12 @@
                                  # specified name). The format is a (case
                                  # insensitive) comma separated list.
 
+#dns_valid_ttl =                 # TTL in seconds for any valid DNS record.
+                                 # If not specified the DNS resolver will
+                                 # honour the TTL specified in the DNS record.
+                                 # If given, it will override the TTL for any
+                                 # valid record received.
```
thibaultcha: This description is somewhat confusing to me. How about a more concise approach, less repetitive:
```
By default, DNS records are cached using the TTL value
of a response. If this property receives a value (in seconds),
it will override the TTL for all records.
```

kong-0.14.1-0.rockspec
```diff
@@ -28,7 +28,7 @@ dependencies = {
   "luaossl == 20171028",
   "luasyslog == 1.0.0",
   "lua_pack == 1.0.5",
-  "lua-resty-dns-client == 2.1.0",
+  "lua-resty-dns-client == 2.2.0",
```
thibaultcha: It would have been nice to split the bump and the feature into 2 separate commits :)

On top of the commit atomicity benefits, we would likely avoid forgetting to update the documentation at release time (if you do not happen to do it by then), since a `feat(conf)` commit is quite visible when drafting the changelog.

Tieske: yes, good point. Adjusted it.

thibaultcha:
@Tieske Thanks, merged!

PR #2963, opened 2017-10-16
Ability to use "Authorization: WhaTeVer base64(username:password)"

francois-maillard:
NOTE: Please read the CONTRIBUTING.md guidelines before submitting your patch,
and ensure you followed them all:

https://github.com/Mashape/kong/blob/master/CONTRIBUTING.md#contributing

### Summary

Ability to use `Authorization: WhaTeVer base64(username:password)` has a authorization header for the ldap-auth plugin.

In particular, when not using the basic-auth plugin, using `Authorization: Basic base64(username:password)` instead of "Authorization: LDAP base64(username:password)` will allow you to use any client (web browser, curl) without having to fool around the headers manually.

### Full changelog

* The ldap-auth plugin can now use any `Authorization type` you want, such as `Authorization: LDAP base64(username:password)`, `Authorization: Basic base64(username:password)`, `Authorization: WhaTeVer base64(username:password)`. This can be set through the new `config.header_type` configuration option for the ldap-auth plugin. Its default value is `ldap`

PR #3731, opened 2018-08-28
fix(ldap-auth) hardcode ldap header so schema can change

darrenjennings:
### Summary

Hard coding header default so that schema can change in later versions of Kong without affecting behavior of migration.


thibaultcha:
Related to #2963 (introducing the `header_type` option), 58942b59a1816036066b329b32cb459f9d6773ea (creating the migration), #3656 (a fix requiring the option to be left lower-cased by default but upper-cased in WWW-Authenticate), and #3686 (the migration capitalizing the default for `header_type`).

PR #3733, opened 2018-08-29
conf_loader: minor refactor + update style

thibaultcha:


thibaultcha:
cc @nateslo - see the first commit which avoids allocating to many identical metatables :)

PR #3739, opened 2018-09-03
Groundwork for porting the Plugins core entity to kong.db

hishamhm:
This is the first half of the work porting the Plugins core entity to kong.db — Plugins is the last pending core entity using the old DAO; after this, the only core entity using the old DAO will be the deprecated Apis entity, as well as the non-core entities from Plugins, which are on the way to being ported as well.

This PR contains a number of features and fixes needed to port Plugins to `kong.db`, the new DAO infrastructure. This contains all the work that can be merged prior to "flicking the switch" and deleting the entity from `kong.dao` and adding the entity in `kong.db`.

This contains the minimal changes to the test suite to keep tests passing. Tests exercising the new behaviors added will appear in the second PR, in which the Plugins entity is ported over.

### DIFF
--- a/kong/db/init.lua
+++ b/kong/db/init.lua
@@ -25,6 +25,7 @@ local CORE_ENTITIES = {
   "snis",
   "upstreams",
   "targets",
+  "apis",
 }
 
 
@@ -47,7 +48,7 @@ function DB.new(kong_config, strategy)
 
   -- load errors
 
-  local errors = Errors.new(strategy or kong_config.database)
+  local errors = Errors.new(strategy)
 
   local schemas = {}
 
@@ -65,8 +66,12 @@ function DB.new(kong_config, strategy)
         return nil, fmt("schema of entity '%s' is invalid: %s", entity_name,
                         tostring(errors:schema_violation(err_t)))
       end
-
-      schemas[entity_name] = Entity.new(entity_schema)
+      local entity, err = Entity.new(entity_schema)
+      if not entity then
+        return nil, fmt("schema of entity '%s' is invalid: %s", entity_name,
+                        err)
+      end
+      schemas[entity_name] = entity
     end
   end
 
@@ -79,7 +84,6 @@ function DB.new(kong_config, strategy)
   end
 
   local daos = {}
-
 
   local self   = {
     daos       = daos,       -- each of those has the connector singleton
@@ -96,7 +100,6 @@ function DB.new(kong_config, strategy)
       if not strategy then
         return nil, fmt("no strategy found for schema '%s'", schema.name)
       end
-
       daos[schema.name] = DAO.new(self, schema, strategy, errors)
     end
   end
--- a/kong/db/strategies/cassandra/init.lua
+++ b/kong/db/strategies/cassandra/init.lua
@@ -411,10 +411,14 @@ function _M.new(connector, schema, errors)
 
       -- store in constraints to subsequently retrieve the inverse relation
 
-      _constraints[field.reference] = {
+      if not _constraints[field.reference] then
+        _constraints[field.reference] = {}
+      end
+      insert(_constraints[field.reference], {
         schema     = schema,
         field_name = field_name,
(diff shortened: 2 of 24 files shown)


kong/db/strategies/cassandra/init.lua
```diff
@@ -519,44 +506,43 @@ function _mt:insert(entity, options)
...
-      return nil, err
-    end
+  local partitioned, err = is_partitioned(self)
+  if err then
+    return nil, err
+  end
 
-    cql = fmt(cql, ttl)
+  local cols = {}
+  local binds = {}
```
thibaultcha: Please initialize these variables below the `args` local declaration? And we should use the `new_tab(fields_len)` allocation (caching the `#schema.fields` value), even if we may end up using less slots that allocated.

kong/db/strategies/cassandra/init.lua
```diff
@@ -575,9 +561,19 @@ function _mt:insert(entity, options)
       end
 
       insert(args, serialize_arg(field, entity[field_name]))
+      insert(cols, field_name)
+      insert(binds, "?")
```
thibaultcha: Respecting the `insert(cols); insert(binds); insert(args)` order (as per the `is_partitioned` branch) might be helpful for the reader.

kong/db/strategies/cassandra/init.lua
```diff
@@ -575,9 +561,19 @@ function _mt:insert(entity, options)
...
     end
+
+    ::continue::
   end
 
+  local cql = fmt("INSERT INTO %s(%s) VALUES(%s)%s",
+                  schema.name,
+                  concat(cols, ", "),
+                  concat(binds, ", "),
+                  ttl and fmt(" USING TTL %d", ttl) or "")
```
thibaultcha: This function suddenly became more costly for all Cassandra versions. Was there no way to bind [cassandra.unset](https://thibaultcha.github.io/lua-cassandra/modules/cassandra.html#type_serializers) instead?

kong/db/init.lua
```diff
@@ -65,8 +66,12 @@ function DB.new(kong_config, strategy)
         return nil, fmt("schema of entity '%s' is invalid: %s", entity_name,
                         tostring(errors:schema_violation(err_t)))
       end
-
-      schemas[entity_name] = Entity.new(entity_schema)
+      local entity, err = Entity.new(entity_schema)
+      if not entity then
+        return nil, fmt("schema of entity '%s' is invalid: %s", entity_name,
```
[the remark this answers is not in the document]
hishamhm: > This would mean that there are errors that the MetaSchema won't catch, but Entity.new will?

Yes. The MetaSchema checks that it's a valid Schema, not that it's a valid Entity. This was already the case before this PR, we just didn't catch those errors. (I hit this during development when I wrote a bad entity).

> If so, maybe the MetaSchema should be responsible for validating such subclasses itself? This way we don't duplicate error handling here, which seems like a design flaw as it stands.

My initial thought about this is that a MetaSchema-based validation of Entity structures would be doable with a "MetaEntity" subschema for the MetaSchema, in which the additional constraints are specified (which would be conceptually a pretty cool solution). That would require a little structural tweaking, so I would prefer to implement this improvement on a further PR though.

kong/db/strategies/cassandra/init.lua
hishamhm: I haven't tried that approach, I just went with copying the old DAO approach which worked on C* 2.2 — giving cassandra.unset a spin now

thibaultcha: It seems like this moved to query-build-time anyway

thibaultcha:
Builds succeeded even though the status is not updated in the PR - probably due to the current Travis issues.

thibaultcha:
BTW, it is nice to use a merge commit for such long-lived branches, as it makes the history somewhat more readable in those cases, and reverts get easier as well!

PR #3745, opened 2018-09-05
fix(dao) make drop_schema aware of new-dao tables

kidd:
consumers, certificates, and snis are dropped when drop_schema is
called (for example, when `kong migrations reset` is called from cli)


kong/dao/db/cassandra.lua
```diff
@@ -30,7 +30,14 @@ _M.dao_insert_values = {
...
 }
 
-_M.additional_tables = { "cluster_events", "routes", "services" }
+_M.additional_tables = {
+  "cluster_events",
+  "routes",
+  "services",
+  "consumers",
+  "certificates",
+  "snis"
```
thibaultcha: Mind adding the trailing comma on these tables? Thanks!

PR #3749, opened 2018-09-06
refactor(test) adds testcases for (trailing) slash handling

Tieske:
Added cases where the incoming uri is longer than the matched uri.

The tests match current behaviour. But please note case 37, it injects an extra slash. This seems counter intuitive to me, I would have expected `/fee/bors` as the final uri.


Tieske:
Updated after discussion with @thibaultcha 

Will work on a follow-up PR to fix the extra slash behaviour, currently marked as "TODO".

Issue #2674, opened 2017-07-06
Kong doesn't remove old Cassandra nodes


braedon:
### Summary
Kong keeps trying to send requests to old Cassandra peers that have left the cluster, resulting in slow responses as it waits for the request to timeout. If the Cassandra nodes don't have static IPs (or the cluster is simply shrunk) Kong instances will accumulate these old peers, increasing the percentage of requests that have to be retried.

After a quick look into lua-cassandra, I think the issue is that Kong never calls `refresh()` on the cluster after initial initialisation.

### Steps To Reproduce

1. Remove/replace a Cassandra node from the cluster, such that an IP that previously had a node no longer does,
2. Make requests that require talking to Cassandra (e.g. `GET /apis/`) until one takes ~10s,
3. Check logs for messages like `[lua] cluster.lua:146: set_peer_down(): [lua-cassandra] setting host at {old_node_ip} DOWN`.

### Additional Details & Logs

- Kong version: 0.10.3

thibaultcha:
Yes indeed, that is a known issue that has surfaced from time to time but stays in our backlog. It affects both scaling up and down of the Cassandra cluster.

Eventually, such a refresh operation should probably be triggered manually from the Admin API as it can be quite expensive (there is no way as of today to notify Kong from Cassandra that the cluster has been scaled up or down).

braedon:
For a bit of context, we're running Kong in a Kubernetes cluster, currently backed by Postgres (also in Kubernetes). We were hoping to switch to Cassandra to make it HA - we use Kong for auth so can't tolerate DB downtime.

Cassandra is also deployed in Kubernetes, which means that the nodes get a new IP each time they restart (updates, failures, node maintenance, etc.), even though the hostnames remain the same.

For the purposes of HA, manually triggering refresh doesn't seem like it would be feasible. It could be managed when doing planned upgrades/maintenance, but requiring a manual step to recover after a failure leaves too much opportunity for subsequent failures to ruin your day before the manual step can be performed.

Plus, the current reconnection policy results in dead nodes being used for subsequent real requests after a fixed interval, despite the node never passing a health check in the meantime. While these requests are retried, the extra 10s delay can cause the downstream request to time out (or simply be unacceptably slow).

If it's too expensive to trigger a refresh whenever a node times out, is it feasible to do an explicit health check before marking them as up, rather than effectively using real requests as health checks? That would at least minimise the requests hit with the delay of trying a down node until refresh can be called.

[...]

PR #3752, opened 2018-09-08
Add manual datastore refresh endpoint

rainest:
### Summary

Add manual datastore refresh functionality.

### Full changelog

* Adds refresh functions for Cassandra and Postgres. The latter is a dummy function for compatibility.
* Adds Admin API route to call datastore refresh functions.

### Background

Related to issue #2674. Although this is not a complete fix (ideally refreshes should not require manual intervention) it's a better workaround than completely restarting a node.

For Cassandra, this adds a simple means of calling the existing Cluster:refresh() function provided by the lua-cassandra driver. Postgres does not have similar needs, so its function is a dummy that the Admin API route can always call successfully.

No tests added as lua-cassandra should handle everything important and test it independently. I performed basic testing by starting Kong, adding a node to its Cassandra cluster, calling refresh, and confirming that the new node's IP was selected by the DB load balancing policy. I then removed the node from the cluster, observed that `set_peer_down()` was called on the node IP, refreshed again, and confirmed that calls to `set_peer_down()` had stopped.

### DIFF
--- a/kong/dao/db/cassandra.lua
+++ b/kong/dao/db/cassandra.lua
@@ -5,6 +5,7 @@ local Errors = require "kong.dao.errors"
 local db_errors = require "kong.db.errors"
 local utils = require "kong.tools.utils"
 local cjson = require "cjson"
+local semaphore = require "ngx.semaphore"
 
 local tonumber = tonumber
 local concat = table.concat
@@ -110,6 +111,9 @@ function _M.new(kong_config)
     return nil, err
   end
 
+  local refresh_sema = semaphore.new(1)
+
+  self.refresh_sema = refresh_sema
   self.cluster = cluster
   self.query_options = query_opts
   self.cluster_options = cluster_options
@@ -302,6 +306,15 @@ function _M:query(query, args, options, schema, no_keyspace)
   local coordinator_opts = {}
   if no_keyspace then
     coordinator_opts.no_keyspace = true
+  end
+
+  if self.refresh_sema:count() < 1 then
+    local ready, err = self.refresh_sema:wait(10)
+
+    if not ready then
+      return nil, 'query timed out while waiting on refresh: ' .. err
+    end
+    self.refresh_sema:post(1)
   end
 
   if coordinator then
@@ -835,7 +848,12 @@ function _M:reachable()
 end
 
 function _M:refresh()
+  local ready, err = self.refresh_sema:wait(10)
+  if not ready then
+    return nil, 'failed to obtain refresh semaphore: ' .. err
+  end
   local ok, err = self.cluster:refresh()
+  self.refresh_sema:post(1)
   if not ok then
     return nil, Errors.db(err)
   end
--- a/kong/dao/db/cassandra.lua
+++ b/kong/dao/db/cassandra.lua
@@ -834,4 +834,13 @@ function _M:reachable()
   return true
 end
 
+function _M:refresh()
+  local ok, err = self.cluster:refresh()
+  if not ok then
+    return nil, Errors.db(err)
+  end
+
+  return true
+end
+
 return _M
--- a/kong/dao/db/postgres.lua
+++ b/kong/dao/db/postgres.lua
@@ -631,4 +631,15 @@ function _M:reachable()
   return true
 end
 
+function _M:refresh()
+  -- This is a dummy function to provide compatibility.
+  -- Refresh only applies to Cassandra.
+  local ok, err = true, true
+  if not ok then
+    return nil, Errors.db(err)
+  end
+
+  return true
+end
+
 return _M
--- a/kong/dao/db/postgres.lua
+++ b/kong/dao/db/postgres.lua
@@ -634,11 +634,6 @@ end
 function _M:refresh()
   -- This is a dummy function to provide compatibility.
   -- Refresh only applies to Cassandra.
-  local ok, err = true, true
-  if not ok then
-    return nil, Errors.db(err)
-  end
-
   return true
 end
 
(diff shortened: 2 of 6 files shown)


kong/dao/db/postgres.lua
```diff
@@ -631,4 +631,15 @@ function _M:reachable()
...
 
+function _M:refresh()
+  -- This is a dummy function to provide compatibility.
+  -- Refresh only applies to Cassandra.
+  local ok, err = true, true
+  if not ok then
+    return nil, Errors.db(err)
+  end
+
+  return true
```
hishamhm: minor suggestion: no need to keep the "error handling template" in this Postgres implementation. In similar situations we just keep it simple, with the body keeping only `return true` and the comment.

rainest:
There seems to be an underlying issue in lua-cassandra that prevents this from working as expected. After refreshing, Kong's load balancing policy selects the correct nodes, but other actions will still attempt to use old addresses and fail. Reviewing how refresh modifies the driver's data structures to see why this occurs.

rainest:
Now improved with lua-resty-worker-events to run refresh on all workers. Previously, only one worker would run refresh, and other workers would continue to use outdated load balancer tables unsuccessfully.

While testing, I used [a modified version](https://gist.github.com/rainest/0a59608cf6c86d84b113c26b465b5419) of the lua-cassanda cluster object to litter debug logging statements everywhere and confirm that refresh was run on multiple workers. There are no functional changes there, however, so I have not made a separate PR against lua-cassandra.

kong/runloop/handler.lua
```diff
@@ -165,9 +165,16 @@ return {
       local worker_events  = singletons.worker_events
       local cluster_events = singletons.cluster_events
 
+      -- database topology refresher
 
-      -- events dispatcher
+      worker_events.register(function()
+        local ok, err = dao.db:refresh()
+        if not ok then
+          ngx.log(ERR, "[events] failed to schedule DB refresh: ", err)
```
thibaultcha: If the cluster topology fails to refresh, it warrants a higher logging level. That is because more than one transaction/request might be affected by this error. In fact, this error may trigger cascading failures in the long run. A logging level set to `ngx.CRIT` seems more appropriate here.

Also, the error message seems wrong: it isn't the scheduling that failed in this case, but the refresh itself. And the other way around: the message currently showed in the Admin API failure branch seems to be wrong; in this case, it is indeed the _scheduling_ that failed, and not the refresh itself.

thibaultcha (changes requested):
I have concerns about the use of `refresh()` at runtime, as detailed in the related comment. This method was not implemented as a complete solution back then, and was never completed to support true runtime topology refresh... Let me know your thoughts.

kong/dao/db/cassandra.lua
```diff
@@ -834,4 +834,13 @@ function _M:reachable()
...
 end
 
+function _M:refresh()
+  local ok, err = self.cluster:refresh()
+  if not ok then
+    return nil, Errors.db(err)
+  end
+
+  return true
+end
```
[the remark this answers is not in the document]
rainest: Updated to include equivalent functions in the new DAO and call both until the old DAO is removed.

angrosz:
When are you going to have the final solution of this problem? I have the same problem as described in #2674.

angrosz:
This solution is a manual action. Is it possible to implement an automatic cassandra cluster refresh?

rainest:
@thibaultcha added [a locking mechanism to block queries](https://github.com/Kong/kong/pull/3752/commits/3f23b2540bcf869a2d47e28643b46bb931e207d5) at the Kong level. I initially attempted a driver-level solution, but wasn't sure if there was a way to implement it at the [host query level](https://github.com/thibaultcha/lua-cassandra/blob/664422d2e58fbd7e7cc30e2f6a576fd518c171ce/lib/cassandra/init.lua#L485). Kong sends queries directly to hosts in some cases (see [this example](https://github.com/Kong/kong/blob/master/kong/db/strategies/cassandra/connector.lua#L193) in the new DAO versus a [query sent to the cluster](https://github.com/Kong/kong/blob/master/kong/dao/db/cassandra.lua#L316) in the old DAO) and I didn't see an existing link back to the cluster object (where refresh grabs its semaphore). I still have several open questions/issues:
* Should this use the existing `cassandra_timeout` or its own (currently hardcoded to 10s arbitrarily)?
* Should queries always attempt to grab the semaphore? I added a check to `count()` first to ensure they only did so if someone invokes `refresh()`, as otherwise queries could block each other, albeit only briefly (they only grab the resource for a preclearance check/do not hold onto it until the query completes). Checking via `count()` before running `wait()` means that there isn't a true guarantee of exclusivity, though it's a stronger safeguard than not checking anything.
* Is there a practical means of adding an automated test for this? I don't know of a means to guarantee a test query will attempt to run while `refresh()` holds the semaphore. I do see expected behavior with a [modified version](https://gist.github.com/rainest/777673a5edbfada6ff2c2795ed5a6a1b#file-connector-lua-L337-L338) that intentionally delays refresh completion enough that I can manually issue queries then.

@angrosz this should work as-is if you want to make a custom build that includes it, though it may cause some queries to fail during a brief period. The commit mentioned above should prevent those failures.

A fully-automated solution is possible but would require new functionality in the driver. We do not yet have support for handling [CQL EVENT messages](https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L729-L739). Alternately, adding a PostStart hook that calls the Kong refresh endpoint to Cassandra pods would likely handle automating this for reschedule or respawn events, though manual downscaling would still require a manual refresh.

thibaultcha:
Closing in favor of https://github.com/Kong/kong/pull/5071

PR #3754, opened 2018-09-08
Fix/Retries times setting's idempotent method checking

jacobslei:
### Summary

Optimized balancer's set_more_tries's conditions, added  idempotent method condition

### Full changelog

* Added idempotent  checking for Calling set_more_tries .

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -401,7 +401,7 @@ function Kong.balancer()
     local retries = balancer_data.retries
     -- non-idempotent method should not be retried
     local method = ngx.req.get_method()
-    if retries > 0 and method ~= "POST" and method ~= "PATCH" and method ~= "LOCK"then
+    if retries > 0 and method ~= "POST" and method ~= "PATCH" and method ~= "LOCK" then
       set_more_tries(retries)
     end
   end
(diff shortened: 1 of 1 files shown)


thibaultcha:
@jacobslei Do you see non-idempotent methods currently being retried, or is this a suggestion for a performance improvement? As of today, non-idempotent methods should already not be retried, as per the default values of the [proxy_next_upstream](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream) directive. I suspect the latter, but your PR's title mentions that it is a "fix".

jacobslei:
@thibaultcha  our cases: Under the situation that w had set proxy_next_upstream directive for upstream 500 error and set retries times more then 0 for balancer, when we sent a post method  request and the upstream server return 500,  the next time's retry, then the OpenResty server returned    internal  error page.

jacobslei:
"As of today, non-idempotent methods should already not be retried", it's right.  
I meant, if you use ngx.balancer.set_more_retries  with input parameter greater than 0 for non-idempotent request , when the upstream server return error, the OpenResty's internal error will occur.

thibaultcha:
@jacobslei Oh, interesting! That would be a nice catch. Would you mind contributing a test for this behavior as well then? It will serve both as a regression test and as a way to demonstrate the issue :) And we require it in order to be able to merge your PR, accessorily... Such a test should probably live in https://github.com/Kong/kong/blob/master/spec/02-integration/05-proxy/09-balancer_spec.lua

PR #3760, opened 2018-09-11
feat(db) arguments parser to decode and infer json arguments

bungle:
### Summary

Our arguments parser on `admin api` does decode and infer `application/x-www-form-urlencoded` but it expects `application/json` to be real and rightly typed JSON structure. This makes it cumbersome to use tools like `HTTPie` which by default send `application/json`. So I added decoding and inferring of JSON arguments to this library too in sense of `being liberal in what you accept`. Not sure though what will others agree with this patch, but as it was so easy to do, I did it, and we can have conversation here about whether or not to merge it.


hishamhm:
I may be just not caffeinated enough yet, but I didn't get it — could you give examples of inputs that are not accepted without this patch but are accepted with it?

bungle:
Some examples:

```
$ http :8001/services url=http://example.org/
$ http :8001/services url=http://example.org/ -f
```

Both work, which is fine.

```
$ http :8001/routes paths=/ service.id=<uuid>
$ http :8001/routes paths=/ service.id=<uuid> -f
```
The first one does not work. The second one fails as it sends this:

```json
{
    "paths": "/",
    "service.id": "<uuid>"
}
```

And the error is:
```json
{
    "code": 2,
    "fields": {
        "paths": "expected an array",
        "service": "required field missing",
        "service.id": "unknown field"
    },
    "message": "3 schema violations (paths: expected an array; service: required field missing; service.id: unknown field)",
    "name": "schema violation"
}
```

So the first `HTTPie` call need to be changed to:

```
$ http :8001/routes paths:='["/"]' service:='{ "id":"<uuid>" }' -v
```

So that it sends:

```json
{
    "paths": [
        "/"
    ],
    "service": {
        "id": "72e30a1e-2476-4bc0-a1a8-9ceee7b4cd44"
    }
}
```

Which is really hard to do without making any typos. Also you cannot send anything other than strings with `url=http://example.org/` as the types are not inferred, e.g. `retries=5` need to be written as `restries:=5`.

This patch kinda makes `application/json` calls decoded and inferred similarly to `application/x-www-form-urlencoded`. It still allows you to send right structures too. It is just more liberal.

hishamhm:
I admit my comment above did remind me of this meme, sorry about that :)

Code reviews meme

bungle:
I do see your point. The old Admin API did this. This change is limited to admin api only. But sure, maybe it creates more problems than it solves. It is just freaking annoying to do many things without `-f` and then to keep your sanity, add it to some other calls. Yes, you could also create alias to always add `-f` option or perhaps use https://httpie.org/doc#config-file-location.

I guess no-one dares to defend this pr after reading your wall of text, :-). But let's see...

bungle:
And it seems like Travis does not agree with this either? That is fixable, but if the concept is not wanted, then it doesn't make sense to spend time to get details just right.

Tieske:
I like the flexibility, but had the exact same itch about unknown edgecases.

Would it be possible to provide a better error message in those cases? (see [this issue](https://discuss.konghq.com/t/array-of-strings-in-plugin-configuration-order-not-respected/1466/6?u=tieske) for a live case)

bungle:
We could also find a sweet spot. E.g. infer types, but not do any this decoding stuff?

PR #3762, opened 2018-09-11
chore(deps) bump luaossl to 20180708

thibaultcha:
* adds `ssl.pushffi()`

Previous bump attempted by e59c031c3a0be6827d52c756aedb04614973c04d, but
reverted due to an issue with our Alpine image and musl libc, which is
fixed in this new version.


thibaultcha:
Already in next

PR #3765, opened 2018-09-12
fix(cors) set missing vary=origin

marckhouzam:
When Access-Control-Allow-Origin is not *, the cors plugins normally
sets Vary=Origin.  In the case where the plugin sets
Access-Control-Allow-Credentials the Vary=Origin is missing.

When that happens, the browser is told to uses its cache even when the
origin is different and then CORS fails by rejecting the non-matching
origin.

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -79,6 +79,7 @@ local function configure_credentials(ngx, conf)
     if req_origin then
       ngx.header["Access-Control-Allow-Origin"]      = req_origin
       ngx.header["Access-Control-Allow-Credentials"] = "true"
+      ngx.header["Vary"] = "Origin"
     end
   end
 end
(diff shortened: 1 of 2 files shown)


thibaultcha:
@marckhouzam Thank you for the patch! Would you mind adding a test for this fix as well? We would need it in order for us to merge your PR. Thank you :)

marckhouzam:
I'll give it a try.

nijikokun:
We should review this further before allowing it to go through as there are implications when it isn't the origin. This blog post describes it:

https://www.fastly.com/blog/caching-cors

marckhouzam:
@nijikokun I read through the documentations you posted and it does look like the cache handling through the use of Vary could be improved.  However, I am thinking that currently the CORS plugin has a bug in its handling of the cache when credentials=true and this PR has value in fixing that. What do you think?

With that in mind, I modified the CORS tests in a second commit.  After looking at the existing tests for the cors plugin, I thought that adding a check in the different relevant tests for header Vary=Origin would cover the most ground.  Since there was already a test where credentials=true, then checking for Vary=Origin in there shows that the test fails with the current master branch while it passes once my PR is applied.

I could add an is_nil() check for header Vary in all the other CORS tests if you would like.

Thanks

marckhouzam:
Any comments on this updated PR?
Thanks!

thibaultcha:
Thanks @marckhouzam, we'll have a look at it soon!

thibaultcha:
@marckhouzam Thank you! 

Btw, feel free to claim your [contributor T-shirt](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt) now that your first PR has been merged :)

marckhouzam:
Thanks for the quick turnaround!
It makes it a great experience to contribute to your project.

PR #3767, opened 2018-09-13
Minor core refactors

james-callahan:


spec/01-unit/002-conf_loader_spec.lua
```diff
@@ -766,16 +766,6 @@ describe("Configuration loader", function()
...
-    it("rejects an unknown scheme", function()
-      local conf, err = conf_loader(nil, {
-        origins = "http://foo:42=ftp://example.com",
-      })
-      assert.is_nil(conf)
-      assert.equal("an origin must be of the form " ..
-                   "'from_scheme://from_host:from_port=" ..
-                   "to_scheme://to_host:to_port', got " ..
-                   "'http://foo:42=ftp://example.com'", err)
-    end)
```
thibaultcha: We should probably keep this test, and make it fail differently now. Ideally in a few different ways, even.

james-callahan: Fixed.

thibaultcha:
@james-callahan The unit test seems to be failing

PR #3769, opened 2018-09-14
refactor(tests) introduce extra setting for PDK testing

Tieske:
The pdk tests use local files for sockets and those cannot be
located on a mounted/shared filesystem. Hence an extra environment
variable for setting the location. This will allow to run the PDK
tests from Vagrant and Docker for example.

Variable is named TEST_NGINX_NXSOCK and should point to a
directory.

### DIFF
--- a/t/01-pdk/04-request/00-phase_checks.t
+++ b/t/01-pdk/04-request/00-phase_checks.t
@@ -3,8 +3,7 @@ use warnings FATAL => 'all';
 use Test::Nginx::Socket::Lua;
 use t::Util;
 
-$ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
-$ENV{TEST_NGINX_NXSOCK}   ||= html_dir();
+$ENV{TEST_NGINX_NXSOCK} ||= html_dir();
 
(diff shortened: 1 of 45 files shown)


Tieske:
currently the default value is the old value (http-web-dir). But I'm wondering whether or not we should update that to $TMPDIR for example, that would make it work out of the box, without even specifying the new variable on Vagrant and Docker

t/01-pdk/04-request/00-phase_checks.t
```diff
@@ -4,6 +4,7 @@ use Test::Nginx::Socket::Lua;
 use t::Util;
 
 $ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
```
thibaultcha: This variable becomes unused; this patch should be getting rid of it.

Tieske: Good catch, I didn't check on those. Will update.

hishamhm:
Was about to approve and squash this and wondered, should we use `tests(pdk)` in squashed commit msg?

thibaultcha:
@hishamhm Yes

Tieske:
👍

PR #3772, opened 2018-09-15
refactor(db) a more consistent options handling

bungle:
### Summary

Refactored `kong.db.dao` to be more consistent with the options table.

This was originally proposed in
#3710 

And this is one of the series of PRs that were previously implemented in that branch.

### DIFF
--- a/kong/db/dao/targets.lua
+++ b/kong/db/dao/targets.lua
@@ -1,9 +1,16 @@
-local cjson = require "cjson"
 local balancer = require "kong.runloop.balancer"
 local utils = require "kong.tools.utils"
+local cjson = require "cjson"
+
+
+local setmetatable = setmetatable
+local tostring = tostring
+local ipairs = ipairs
+local assert = assert
+local table = table
+
 
 local _TARGETS = {}
-
 local DEFAULT_PORT = 8000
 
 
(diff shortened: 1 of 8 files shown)


kong/db/dao/targets.lua
```diff
@@ -138,11 +145,11 @@ end
 -- including entries that have been since overriden, and those
 -- with weight=0 (i.e. the "raw" representation of targets in
 -- the database)
-function _TARGETS:select_by_upstream_raw(upstream_pk)
+function _TARGETS:select_by_upstream_raw(upstream_pk, ...)
```
kikito: I think the `...` can be replaced with the param `options` here.

kikito:
@bungle I don't know how or why, but this seems to be triggering a "memory corruption" problem on nginx on travis.

Two possible things that could fix this (I am not sure):
* maybe replace the varargs (`...`) with `options` as I mentioned before
* try not localizing all the functions? Or localizing them in the scope they are used, instead of at the top of the file (shooting blind here)

bungle:
@kikito memory corruption problem is unrelated to this PR. It is a problem in the Travis OpenSSL builds triggered by OpenSSL / luaossl bump.

PR #3774, opened 2018-09-15
Feat/oauth2 daos.lua & api.lua to new db

kikito:
This change has some previous groundwork, then the commit itself which deals with oauth2, and then some cleanup. Each part has its own separate commit, with explanations. I recommend reviewing the PR commit-by-commit.


spec/01-unit/000-new-dao/01-schema/01-schema_spec.lua
```diff
@@ -1276,10 +1277,10 @@ describe("schema", function()
       }))
       Test.primary_key = { "f" }
       local ok, errs = Test:validate_primary_key({
-        f = { id = ngx.null },
+        f = { },
```
hishamhm: More of a question than a comment: I failed to understand how the changes in this commit affected the behavior of this test requiring this change. Could you elaborate a bit on this here, just so we're on the same page?

kong/plugins/oauth2/access.lua
```diff
@@ -136,7 +135,9 @@ local function retrieve_scopes(parameters, conf)
     return false, {[ERROR] = "invalid_scope", error_description = "You must specify a " .. SCOPE}
```
hishamhm: if the above `false` returns were changed to `nil`, perhaps this one should as well?

kong/plugins/oauth2/access.lua
```diff
@@ -163,13 +164,13 @@ local function authorize(conf)
       end
 
       -- Check scopes
-      local ok, scopes = retrieve_scopes(parameters, conf)
-      if not ok then
-        response_params = scopes -- If it's not ok, then this is the error message
+      local scopes, err = retrieve_scope(parameters, conf)
+      if err then
+        response_params = err -- If it's not ok, then this is the error message
```
hishamhm: now that you changed the return type to `scope | nil, err` this comment (and the similar one below) aren't necessary :)

kong/plugins/oauth2/migrations/cassandra.lua
```diff
@@ -74,37 +75,41 @@ return {
   },
   {
     name = "2016-04-14-283949_serialize_redirect_uri",
-    up = function(_, _, factory)
-      local json = require "cjson"
-      local apps, err = factory.oauth2_credentials.db:find_all('oauth2_credentials', nil, nil);
+    up = function(_, _, dao)
+      local rows, err = dao.db:query([[
```
hishamhm: This SELECT construct will not return all elements. For row-by-row updates in Cassandra migrations, you need to use the `iterate` construct as shown here:
https://github.com/Kong/kong/blob/next/kong/dao/migrations/cassandra.lua#L586-L603

Likewise in the migrations below.

kong/db/dao/plugins.lua
```diff
@@ -268,16 +268,16 @@ function Plugins:load_plugin_schemas(plugin_set)
         local Strategy = require(string.format("kong.db.strategies.%s", db.strategy))
         for name, schema_def in pairs(daos_schemas) do
           if name ~= "tables" and schema_def.name then
-            ngx_log(ngx_DEBUG, string.format("Loading custom plugin entity: '%s.%s'", plugin, name))
+            ngx_log(ngx_DEBUG, string.format("Loading custom plugin entity: '%s.%s'", plugin, schema_def.name))
```
hishamhm: Took me a while to see how this added support for arrays (namely, that then the `name` variable becomes the numeric index returned in arbitrary order, and that it goes unused in this loop except for comparing against `"tables"`). I just wonder if keeping this around for backwards compatibility sake is worth the "sneaky" code — plus, this code assumes that pairs always traverses the numeric keys of a table in order, right? I believe this is true in practice in LuaJIT for tables specified in literals as is usually the case in daos.lua, but IIRC it is not a formal guarantee.

And AFAICT the backwards-compat will be partial anyway, since plugins with dependencies among custom daos like oauth2 does will require using ordered arrays.

BTW, thanks for the explanation in the commit message, otherwise I would have totally missed the intent!

hishamhm:
Reviewed commit-by-commit. I think the overall port of Oauth2 looks good! I think the fix for the Cassandra migrations is the only blocker.

kong/plugins/oauth2/migrations/cassandra.lua
kikito: OMG I always forget that. Thanks.

kong/db/dao/plugins.lua
kikito: Removing the backward-compatibility and using `ipairs` instead of `pairs` would involve changing all the plugins' `daos.lua` files to return an array instead of a hash. That is the main reason for keeping the "sneaky" code here.

If you don't think that's a good enough reason I will change all the `daos.lua` to arrays and use `ipairs` here.

spec/01-unit/000-new-dao/01-schema/01-schema_spec.lua
kikito: You were right to raise questions about this!

The test were failing so I changed them to something "equivalent" without thinking.

It turns out there was actually a problem: the `_cache` variable I created was a bit "overzealous" in the way it "protected the cache": if you created a schema named `A` and later a second schema named the same way, *the cache would not replace the old schema with the new one*. Since the tests's reuse the same name again and again, the foreign keys were pointing to schemas from previous tests.

I've changed the cache so that it is less "protectionist" and now the tests work as before. Thanks!

kong/plugins/oauth2/access.lua
kikito: Yes, changed all the `return false` to `return nil`

hishamhm:
@kikito Regarding the "sneaky" loop, my suggestion then is to add a comment above the pairs() loop with something like
```
-- If the daos list is declared as an array, we assume below that pairs()
-- will traverse them in order and initialize the schemas
-- such that foreign field relationships are resolved correctly
```

...or use this order-safe iterator instead? :) (couldn't resist! (untested of course) )

```lua
local function each_dao(daos)
  local phase = 1
  local ak = 1
  local hk = next(daos)
  return function()
    if phase == 1 then
      local v = daos[ak]
      ak = ak + 1
      if v == nil then
        phase = 2
      else
        return v
      end
    end
    --- phase == 2:
    while type(hk) == "number" then
      hk = next(daos, hk)
    end
    if hk == "tables" then
      hk = next(daos, hk)
    end
    v = daos[hk]
    hk = next(daos, hk)
    return v
  end
end
```

Issue #3738, opened 2018-09-01
upload certificates fail


lesliechung88:
kong response:

"POST /certificates/ HTTP/1.1" 400 37 "-" "Go-http-client/1.1"

docker-compose.yaml:

 kong:
    image: "kong:0.13.0"
    container_name: kong
    hostname: kong
    networks:
        - edgex-network
    ports:
        - "8000:8000"
        - "8001:8001"
        - "8443:8443"
        - "8444:8444"
    environment:
        - 'KONG_DATABASE=postgres'
        - 'KONG_PG_HOST=kong-db'
        - 'KONG_PROXY_ACCESS_LOG=/dev/stdout'
        - 'KONG_ADMIN_ACCESS_LOG=/dev/stdout'
        - 'KONG_PROXY_ERROR_LOG=/dev/stderr'
        - 'KONG_ADMIN_ERROR_LOG=/dev/stderr'
        - 'KONG_ADMIN_LISTEN=0.0.0.0:8001, 0.0.0.0:8444 ssl'

golang code:

	cert, key, err := getCertKeyPair(config, secretBaseURL, c)
	if err != nil {
		return err
	}
	body := &CertInfo{
		Cert: cert,
		Key:  key,
		Snis: []string{config.SecretService.SNIS},
	}
	lc.Info(config.SecretService.SNIS)
	lc.Info("Trying to upload cert to proxy server.")
	req, err := sling.New().Base(url).Post(CertificatesPath).BodyJSON(body).Request()
	resp, err := c.Do(req)
	if err != nil {
		lc.Error("Failed to upload cert to proxy server with error %s", err.Error())
		return err
	}

--------------------------------------------------------------
NOTE: GitHub issues are reserved for **bug reports only**. For anything else,
please join the conversation in Kong Nation https://discuss.konghq.com/.

Please read the CONTRIBUTING.md guidelines to learn on which channels you can
seek for help and ask general questions:

https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#where-to-seek-for-help

### Summary

SUMMARY_GOES_HERE

### Steps To Reproduce

1.
2.
3.
4.

### Additional Details & Logs

- Kong version (`$ kong version`)
- Kong debug-level startup logs (`$ kong start --vv`)
- Kong error logs (`<KONG_PREFIX>/logs/error.log`)
- Kong configuration (the output of a GET request to Kong's Admin port - see
  https://getkong.org/docs/latest/admin-api/#endpoint)
- Operating system

[...]

Issue #3610, opened 2018-07-11
Error uploading certificate on 0.14.0 : schema violation, required field missing


pixeye33:
### Summary

Can't upload a certificate on a brand new install of kong (docker alpine)

### Steps To Reproduce

1. try to add a certificate on a brand new install of kong (docker alpine)

here are the files i'm trying to upload
fullchain.cer (DER BASE64 formated) :
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----

private.key (DER BASE64 formated) :
-----BEGIN RSA PRIVATE KEY-----
-----END RSA PRIVATE KEY-----

both files are obtained via acme.sh and letsencrypt.

### Additional Details & Logs

- Kong version : 0.14.0
- Kong configuration (pretty much default out of the box) : 
[...]

PR #3776, opened 2018-09-15
feat(admin) add multipart/form-data support to admin api (80% solution)

bungle:
### Summary

This adds `multipart/form-data` support to Admin API. But this is not a 100% solution as there turns our to be edge cases because of Lapis Framework that are not very easy to solve without rather complex changes.

How Lapis Reads the `multipart/form-data`:

1. On OpenResty Lapis uses `lua-resty-upload` to stream read `multipart/form-data` upload.
2. `lua-resty-upload` consumes the request body without writing it back, which means that you cannot read request body in our `arguments` parser.
3. That means that only place we can access `multipart/form-data` params in admin api is through Lapis `self.req.params_post`.
4. Lapis in turn does parsing of `multipart/form-data` to a key/value table. And because of that you cannot send same name in `multipart/form-data` twice, e.g. `paths=/first` and `paths=/second` OR `paths[]=/first` and `paths[]=/second`. But you can work-around that on client side by sending: `paths[1]=/first` and `paths[2]=/second`. That's why I call this 80% solution.

Should we wait for 100% solution? In my opinion not. This 80% solution is already more than enough to solve issues such as:
#3738
#3610

We can re-visit this later with 100% solution.

### DIFF
--- a/kong/api/arguments.lua
+++ b/kong/api/arguments.lua
@@ -653,17 +653,22 @@ local function load(opts)
     end
 
   elseif options.multipart and find(content_type_lower, "multipart/form-data", 1, true) == 1 then
-    local pargs, err = parse_multipart(options, content_type)
-    if pargs then
-      if options.decode then
-        args.post = decode(pargs, options.schema)
+    if options.request and options.request.params_post then
+      args.post = decode(options.request.params_post, options.schema)
 
-      else
-        args.post = pargs
+    else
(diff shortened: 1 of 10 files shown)


bungle:
@kikito I added `multipart/form-data` to many `it_content_types` tests, but because issues in client, and because of this 80% solution, I needed to skip it in some tests. But in general this adds a lot of tests there.

PR #3779, opened 2018-09-17
FEATURE - Rate limit by key

xiejiangzhi:
### Summary

Rate limit by a custom key. Migrate a app to Kong is easier, we can use rate limit with app authorization

### Full changelog

* Add `key` to limit_by and add `key_name`, `key_in_body` to rate limit. Allow limit by this key, we can find the key from header, query and body


.gitignore
```diff
@@ -3,6 +3,7 @@
 .buildpath
 .project
 .idea
+*.swp
```
thibaultcha: Please remove such unrelated changes from your commit - thank you!

kong/plugins/rate-limiting/handler.lua
```diff
@@ -29,6 +29,18 @@ local function get_identifier(conf)
     end
   elseif conf.limit_by == "credential" then
     identifier = ngx.ctx.authenticated_credential and ngx.ctx.authenticated_credential.id
+  elseif conf.limit_by == "key" then
+    identifier = kong.request.get_headers()[conf.key_name] or kong.request.get_query()[conf.key_name]
```
thibaultcha: [kong.request.get_header()](https://docs.konghq.com/latest/pdk/kong.request/#kong_request_get_header) and [kong.request.get_query_arg()](https://docs.konghq.com/latest/pdk/kong.request/#kong-request-get_query_arg) would be more appropriate here.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -29,6 +29,18 @@ local function get_identifier(conf)
...
+
+    if not identifier and conf.key_in_body then
+      local body, err = kong.request.get_body()
+
+      if err then
+        ngx_log(ngx.ERR, "Cannot process request body: ", tostring(err))
```
thibaultcha: Please write the above as:
```lua
local body, err = kong.request.get_body()
if err then
  ngx_log(...)
  return
end

identifier = ...
```
 
As per our style guide and rules around early-exit branches. Also, error logs should _not_ be capitalized. Thanks!

kong/plugins/rate-limiting/schema.lua
```diff
@@ -10,9 +10,11 @@ return {
     day = { type = "number" },
     month = { type = "number" },
     year = { type = "number" },
-    limit_by = { type = "string", enum = {"consumer", "credential", "ip"}, default = "consumer" },
+    limit_by = { type = "string", enum = {"consumer", "credential", "key", "ip"}, default = "consumer" },
     policy = { type = "string", enum = {"local", "cluster", REDIS}, default = "cluster" },
     fault_tolerant = { type = "boolean", default = true },
+    key_name = { type = "string" },
+    key_in_body = { type = "boolean", default = false },
```
thibaultcha: This schema should also ensure that `key_in_body` cannot be set if `key` is not the current policy?

kong/plugins/rate-limiting/schema.lua
```diff
@@ -59,6 +61,10 @@ return {
       end
     end
 
+    if plugin_t.limit_by == "key" and not plugin_t.key_name then
+      return false, Errors.schema "You need to specify a key name"
+    end
```
thibaultcha: These schema updates will need to be ported to the new format _after_ https://github.com/Kong/kong/pull/3778 is merged. It converts our plugins schemas to the new schema library we have developed for Services and Routes, and is becoming the new schema library for Kong 1.0.

thibaultcha (changes requested):
@xiejiangzhi Thank you for the patch! Left a few comments. Also, we will have to wait for https://github.com/Kong/kong/pull/3778 to be merged first, as described in my comments :)

kong/plugins/rate-limiting/handler.lua
thibaultcha: Right. I should have expanded the diff! Nonetheless, we should avoid the empty line between the call and the error handling :)

kong/plugins/rate-limiting/schema.lua
xiejiangzhi: The `key_in_body` is invalid if `key` is not the current policy.  It will be used when policy is `key` only. 
Do you have any advice?

thibaultcha: Yes, I am advising that if `key_in_body` is configured while `limit_by` is not `key`, then `self_check` returns an error; because as you said, `key_in_body` is invalid if `key` is not the current limiter.

In the new schema library, this would translate with what we call "entity checkers", that can ensure that a field if set if and only if _another_ field is set a particular way. See an example [here](https://github.com/Kong/kong/blob/next/kong/plugins/rate-limiting/schema.lua#L26-L27). Of course, this PR still has to remain in a pending state until the new schemas are merged into master, at which point, the above behavior can be implemented in this PR.

thibaultcha:
Updated this PR with the "WIP" label to reflect the need to wait for #3778 to be merged before moving forward here. Thank you for your understanding!

kong/plugins/rate-limiting/schema.lua
xiejiangzhi: Thanks, I'll update it after the new schema is merged.

james-callahan:
FYI #3778 has now been merged; are you able to rebase+update this PR?

xiejiangzhi:
Hi @james-callahan #3778 just merged into `next` branch, so, should I rebase based on `next` and change this pr merge into `next`?

xiejiangzhi:
Done

kong/plugins/rate-limiting/schema.lua
```diff
@@ -3,7 +3,6 @@ local typedefs = require "kong.db.schema.typedefs"
 
 local ORDERED_PERIODS = { "second", "minute", "hour", "day", "month", "year"}
 
-
```
james-callahan: There's a couple of misc whitespace changes in this PR.

spec/03-plugins/24-rate-limiting/01-schema_spec.lua
```diff
@@ -15,6 +15,12 @@ describe("Plugin: rate-limiting (schema)", function()
     assert.truthy(ok)
     assert.is_nil(err)
   end)
+  it("proper config validates (limit by key)", function()
+    local config = {second = 10, limit_by = "key", key_name = "token"}
+    local ok, _, err = validate_entity(config, plugin_schema)
```
james-callahan: This function doesn't exist (see failing travis run):
```
Checking spec/03-plugins/24-rate-limiting/01-schema_spec.lua 4 warnings
    spec/03-plugins/24-rate-limiting/01-schema_spec.lua:20:24: accessing undefined variable validate_entity
    spec/03-plugins/24-rate-limiting/01-schema_spec.lua:20:48: accessing undefined variable plugin_schema
    spec/03-plugins/24-rate-limiting/01-schema_spec.lua:47:26: accessing undefined variable validate_entity
    spec/03-plugins/24-rate-limiting/01-schema_spec.lua:47:50: accessing undefined variable plugin_schema
```

kong/plugins/rate-limiting/schema.lua
```diff
@@ -38,14 +37,17 @@ return {
...
-              one_of = { "consumer", "credential", "ip" },
+              one_of = { "consumer", "credential", "key", "ip" },
           }, },
           { policy = {
               type = "string",
               default = "cluster",
               len_min = 0,
               one_of = { "local", "cluster", "redis" },
           }, },
+
```
james-callahan: nit: misc whitespace

spec/03-plugins/24-rate-limiting/01-schema_spec.lua
```diff
@@ -29,13 +35,18 @@ describe("Plugin: rate-limiting (schema)", function()
       assert.falsy(ok)
       assert.equal("The limit for year(50.0) cannot be lower than the limit for month(60.0)", err.config)
     end)
-
```
james-callahan: nit: misc whitespace

james-callahan (changes requested):
Just had a quick look over and commented on a couple of things that jumped out at me. I haven't thought about this PR in a wider scope yet.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -29,6 +29,17 @@ local function get_identifier(conf)
     end
   elseif conf.limit_by == "credential" then
     identifier = ngx.ctx.authenticated_credential and ngx.ctx.authenticated_credential.id
+  elseif conf.limit_by == "key" then
+    identifier = kong.request.get_header(conf.key_name) or kong.request.get_query_arg(conf.key_name)
+
+    if not identifier and conf.key_in_body then
+      local body, err = kong.request.get_body()
+      if err then
+        ngx_log(ngx.ERR, "cannot process request body: ", tostring(err))
```
james-callahan: Could you use `kong.log.err` here?
https://docs.konghq.com/0.14.x/pdk/kong.log/#kong-log-level

xiejiangzhi:
Also has some problem in the schema, I think it isn't my side.
the access testing is passed.

```
 if_field = "limit_by", if_match = { eq = "key" },
 then_field = "key_name", then_match = { required = true },
```
and 

```
 if_field = "config.limit_by", if_match = { eq = "key" },
 then_field = "config.key_name", then_match = { required = true },
```

has a different result

I need some help.

xiejiangzhi:
It seems that the testing has some errors. I saw `local ok, _, err = v(config, schema_def)` and `local ok, err = v(config, schema_def)` in your code. according to that helper code, it should be 2 arguments, right?

codifierr:
Can this be used for rate-limiting oauth2 refresh token api?

PR #3780, opened 2018-09-17
fix(router) do not force an extra segment in the path

Tieske:
### Issues resolved

Fixes the issue/todo identified in #3749

### Read carefully

The way this is handled is by changing the way the router constructs the outgoing path. The path is constructed from these elements in order:

1. the upstream prefix (the `path` property in the `service` entity)
2. the path match (based on the `paths` property in the `route` entity)
    - if matching on path
    - if not stripping the path
3. the request postfix (the part after the matched path, or if not matched by path, the full request path)

In all cases, the initial `/` of the incoming requested path will be ignored/removed. This allows the user to control the presence of the `/` in the outgoing path.

The reason for this change is because in the existing code, different inputs lead to the same output, always injecting the extra `/`. See [these lines for example](https://github.com/Kong/kong/blob/master/spec/01-unit/010-router_spec.lua#L1689-L1698). The first set of 5 tests, yield the exact same upstream path as the second set of 5 lines, despite the `upstream_path` being different. The [updated tests (now a block of 12) start here](https://github.com/Kong/kong/pull/3780/files#diff-1770b4b763fc513f92257840b039392eR1691).

This PR also adds quite a number of test cases that were not previously covered. 

The bad news is that this is a breaking change, since it changes the way the upstream path is constructed. The good news is that in the entire Kong test suite there was [only 1 occurrence of a test failing](https://github.com/Kong/kong/pull/3780/files#diff-e1afd87867200fd0cc7977cf95e039a6R131) because of this.


kong/router.lua
```diff
@@ -388,15 +388,7 @@ do
           end
 
           -- plain or prefix match from the index
-          if route_t.strip_uri then
```
thibaultcha: This branch was an optimization around the two `sub` and the concatenation operations that it enclosed. Depending on the answer to the comment right above this one (about `ctx.matches.stripped_uri`), we may or may not need bring it back.

kong/router.lua
```diff
@@ -768,49 +750,35 @@ function _M.new(routes)
...
-               and matches.stripped_uri
-            then
-              upstream_uri = matches.stripped_uri
-            end
+            -- Path construction
 
-            -- uri trailing slash logic
+            -- if we do not have a path-match, then the postfix is simply the
+            -- incoming path, without the initial slash
+            local request_postfix = matches.stripped_uri or sub(req_uri, 2, -1)
```
thibaultcha: This variable should only be needed within the below `if matched_route.strip_uri` branch. Let's move it within it then. Doing so would also avoid executing `sub(req_uri, 2, -1)` twice.

kong/router.lua
```diff
@@ -768,49 +750,35 @@ function _M.new(routes)
...
 
-            local upstream_url_path = upstream_url_t.path
-            local upstream_url_file = upstream_url_t.file
-
-            if upstream_url_path and upstream_url_path ~= "/" then
-              if upstream_uri ~= "/" then
-                upstream_uri = upstream_url_file .. upstream_uri
+            if matched_route.strip_uri then
+              -- we drop the matched part, replacing it with the upstream path
+              if sub(upstream_base, -1, -1) == "/" and sub(request_postfix, 1, 1) == "/" then
```
thibaultcha: The first statement of this condition (`if sub(upstream_base, -1, -1) == "/"`) could be avoided if we relied on `upstream_url_t.file`, which we sanitize only once, [here](https://github.com/Kong/kong/blob/fix/path-handling/kong/router.lua#L265-L266), to ensure that it does not contain a trailing slash.

And when `upstream_base == "/"` (initialized to `file` instead of `path`), we should then do: `upstream_uri = request_postfix`. That would be cheaper. 

Besides, it seems like the third argument in the first `sub()` call is unnecessary, but that's irrelevant if superseded by the above.

spec/01-unit/010-router_spec.lua
```diff
@@ -1681,64 +1681,89 @@ describe("Router", function()
 
     describe("slash handling", function()
       local checks = {
-        -- upstream url    paths            request path    expected path           strip uri
+        -- upstream url    paths           request path    expected path           strip uri
```
thibaultcha: Unrelated to this PR, but I truly wish the order of those were: request path, route path, upstream url, expected url... Even despite the fact that the upstream URL is built from upstream+path+request. If anybody with some time on their hands is reading this...

kong/router.lua
thibaultcha: style: mind keeping the line length below 80 characters please?

spec/02-integration/05-proxy/01-router_spec.lua
```diff
@@ -1035,10 +1087,12 @@ for _, strategy in helpers.each_strategy() do
           config = "(strip_path = off)"
         end
 
-        it(config .. " is not appended to upstream url " .. args[1] ..
-                     " (with uri "                       .. args[2] .. ")" ..
-                     " when requesting "                 .. args[3],
-          check(i, args[3], args[4]))
+        it("(" .. i .. ") " .. config ..
+           " is not appended to upstream url " .. args[1] ..
+           " (with " .. (args[2] and ("uri " .. args[2]) or ("host test" .. i .. ".domain.org")) .. ")" ..
```
thibaultcha: style: ditto (line length)

thibaultcha:
@Tieske Why does this PR have a "do not merge" label? This label signifies that a PR is _rejected_. If you are worried that this PR gets merged without proper review/testing, that should not happen as per in the first place as per our process.

kong/router.lua
Tieske: Considering this code will be moved out of this pre-access code, into post-access, I'd rather not as it seems a premature optimization doing that now. Once a follow up PR puts this in the proper place, we can optimize it there imo.

That said, afaik, the `file` thing is only used here in the router, should it be removed? Once the path construction moves to after-access, we cannot rely on it any more.

spec/01-unit/010-router_spec.lua
Tieske: yes, I had that itch as well 😄

kong/router.lua
Tieske: see comment below, fix it in after-access

kong/router.lua
```diff
@@ -371,7 +371,7 @@ do
 
             if m then
               if m.stripped_uri then
-                ctx.matches.stripped_uri = "/" .. m.stripped_uri
+                ctx.matches.stripped_uri = m.stripped_uri
```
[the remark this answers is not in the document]
thibaultcha: > The variable should probably be renamed

The name is fine as is, it is still the stripped part of the URL. I was mostly worried about the breaking change; but it might be a necessary evil here.

kong/router.lua
thibaultcha: It wasn't clear that there would be a subsequent PR. In this case, I'd rather see the final result of these changes at once, in the same PR (separate commits is fine). I'd argue that aiming towards commit atomicity does mean that each commit should be written optimally: say we end up reverting commit number 2, now commit number 1 does not include the optimizations that were only added in 2.

> That said, afaik, the file thing is only used here in the router, should it be removed? 

When a variable becomes stale (and only then), please, by all means, remove it!

kong/router.lua
```diff
@@ -260,19 +260,12 @@ local function marshall_route(r)
...
   -- TODO: service.path is not supported in new model
   local path = service.path or null
-  local file = path
-  if path ~= null then
-    if sub(path, -1) == "/" and path ~= "/" then
-      file = sub(path,  1, -2)
-    end
 
-  else
+  if path == null then
```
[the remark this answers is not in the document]
Tieske: my assumption here was that the TODO comment above indicated that in the future this value comes in as `null` and not `nil`, hence I retained the existing logic with the extra check. And it seemed a bad combination to also fix that here?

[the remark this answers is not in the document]
bungle: There was once a time when new dao returned `null` instead of `nil`, but that was changed to configurable. Today it returns `nil` by default. Only the admin API uses `null`.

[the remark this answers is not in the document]
Tieske: Fixed after consulting @bungle, see https://github.com/Kong/kong/pull/3780/commits/d1cad14432c14801d23280ea3f6bace5f9186ea3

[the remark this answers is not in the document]
thibaultcha: Yes, just like @bungle said. Thank you!

spec/02-integration/05-proxy/01-router_spec.lua
```diff
@@ -939,107 +939,228 @@ for _, strategy in helpers.each_strategy() do
...
+        {  "/fee/bor/",    nil,            "/foo/bar/",    "/fee/bor/foo/bar/",    false     },
+        {  "/",            nil,            "/",            "/",                    true      }, -- 70
+        {  "/",            nil,            "/foo/bar",     "/foo/bar",             true      },
+        {  "/",            nil,            "/foo/bar/",    "/foo/bar/",            true      },
+        {  "/fee/bor",     nil,            "/",            "/fee/bor",             true      },
+        {  "/fee/bor",     nil,            "/foo/bar",     "/fee/borfoo/bar",      true      },
+        {  "/fee/bor",     nil,            "/foo/bar/",    "/fee/borfoo/bar/",     true      },
+        {  "/fee/bor/",    nil,            "/",            "/fee/bor/",            true      },
+        {  "/fee/bor/",    nil,            "/foo/bar",     "/fee/bor/foo/bar",     true      },
+        {  "/fee/bor/",    nil,            "/foo/bar/",    "/fee/bor/foo/bar/",    true      },
```
thibaultcha: Are we not testing with regex paths as we discussed? By simply copy-pasting a prefix match block, and making "dumb" regexes?

thibaultcha: Great then

thibaultcha:
If good to merge on your end too, let's proceed **with a merge commit**

Tieske:
yay 🎉   @thibaultcha please go ahead and merge as you see fit

PR #3782, opened 2018-09-17
perf(router) wrap router rebuilds in a mutex

### LINKED ISSUE #3634: Spike in postgres connections when routes are added or updated

p0pr0ck5:
Guard calls to build_router in a worker-level mutex (via
ngx.semaphore) in order to prevent database dogpiling during
rebuild events in high-concurrency traffic patterns. Requests
on a worker process rebuilding the router will be queued and resumed
once the rebuild is complete.

### Full Changelog

* wrap build_router calls in a worker-level mutex

### Issues resolved

Fix #3634

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -15,6 +15,7 @@ local ApiRouter   = require "kong.api_router"
 local reports     = require "kong.reports"
 local balancer    = require "kong.runloop.balancer"
 local constants   = require "kong.constants"
+local semaphore   = require "ngx.semaphore"
 local responses   = require "kong.tools.responses"
 local singletons  = require "kong.singletons"
 local certificate = require "kong.runloop.certificate"
@@ -45,6 +46,10 @@ local EMPTY_T = {}
 local router, router_version, router_err
 local api_router, api_router_version, api_router_err
 local server_header = meta._SERVER_TOKENS
+
+
+local build_router_semaphore
+local build_api_router_semaphore
 
 
 local function get_now()
@@ -408,6 +413,23 @@ return {
         balancer.init()
       end)
 
+      do
+        local err
+
+        -- semaphore (mutex) for build_api_router
+        build_api_router_semaphore, err = semaphore.new()
+        if err then
+          log(ngx.CRIT, "failed to create build_api_router_semaphore: ", err)
+        end
+        build_api_router_semaphore:post(1)
+
+        -- semaphore (mutex) for build_router
+        build_router_semaphore, err = semaphore.new()
+        if err then
+          log(ngx.CRIT, "failed to create build_router_semaphore: ", err)
+        end
+        build_router_semaphore:post(1)
+      end
     end
   },
   certificate = {
@@ -435,13 +457,55 @@ return {
         log(ngx.CRIT, "could not ensure API router is up to date: ", err)
 
       elseif api_router_version ~= version then
-        log(DEBUG, "rebuilding API router")
 
-        local ok, err = build_api_router(singletons.dao, version)
-        if not ok then
-          api_router_err = err
-          log(ngx.CRIT, "could not rebuild API router: ", err)
+        -- wrap api_router rebuilds in a per-worker mutex (via ngx.semaphore)
+        -- this prevents dogpiling the database during rebuilds in
+        -- high-concurrency traffic patterns
+        -- requests that arrive on this process during a api_router rebuild will be
+        -- queued. once the semaphore resource is acquired we re-check the
+        -- api_router version again to prevent unnecessary subsequent rebuilds
+        --
+        -- note the ten-second wait() timeout. we should be able to safely assume
+        -- that a api_router rebuild event should never take this long
+
+        -- for postgres-backed installations, wait up to 60 seconds (the default
+        -- socket timeout) to acquire the mutex when attempting to build the
+        -- router. cassandra-backed installations rely on the 'cassandra_timeout'
+        -- configuration to better align with the configured timeout behavior 
+        -- that other cassandra queries use
+        local timeout = 60
+        if singletons.configuration.database == "cassandra" then
+          -- cassandra_timeout is defined in ms
+          timeout = singletons.configuration.cassandra_timeout / 1000
         end
+
+        local ok, err = build_api_router_semaphore:wait(timeout)
+        if not ok then
+          return responses.send_HTTP_INTERNAL_SERVER_ERROR(
+            "error attempting to acquire build_api_router lock: " .. err
+          )
+        end
+
+        -- lock acquired but we might not need to rebuild the api_router (if we were
+        -- not the first request in this process to enter this code path)
+        -- check again and rebuild if necessary
+
+        version, err = cache:get("api_router:version", CACHE_ROUTER_OPTS, utils.uuid)
+        if err then
+          log(ngx.CRIT, "could not ensure api_router is up to date: ", err)
+
+        elseif api_router_version ~= version then
+          -- api_router needs to be rebuilt in this worker
+          log(DEBUG, "rebuilding api_router")
+
+          local ok, err = build_api_router(singletons.db, version)
+          if not ok then
+            api_router_err = err
+            log(ngx.CRIT, "could not rebuild api_router: ", err)
+          end
+        end
+
+        build_api_router_semaphore:post(1)
       end
 
       if not api_router then
@@ -456,14 +520,55 @@ return {
         log(ngx.CRIT, "could not ensure router is up to date: ", err)
 
       elseif router_version ~= version then
-        -- router needs to be rebuilt in this worker
-        log(DEBUG, "rebuilding router")
 
-        local ok, err = build_router(singletons.db, version)
-        if not ok then
-          router_err = err
-          log(ngx.CRIT, "could not rebuild router: ", err)
+        -- wrap router rebuilds in a per-worker mutex (via ngx.semaphore)
+        -- this prevents dogpiling the database during rebuilds in
+        -- high-concurrency traffic patterns
+        -- requests that arrive on this process during a router rebuild will be
+        -- queued. once the semaphore resource is acquired we re-check the
+        -- router version again to prevent unnecessary subsequent rebuilds
+        --
+        -- note the ten-second wait() timeout. we should be able to safely assume
+        -- that a router rebuild event should never take this long
+
+        -- for postgres-backed installations, wait up to 60 seconds (the default
+        -- socket timeout) to acquire the mutex when attempting to build the
+        -- router. cassandra-backed installations rely on the 'cassandra_timeout'
+        -- configuration to better align with the configured timeout behavior 
+        -- that other cassandra queries use
+        local timeout = 60
+        if singletons.configuration.database == "cassandra" then
+          -- cassandra_timeout is defined in ms
+          timeout = singletons.configuration.cassandra_timeout / 1000
         end
+
+        local ok, err = build_router_semaphore:wait(timeout)
+        if not ok then
+          return responses.send_HTTP_INTERNAL_SERVER_ERROR(
+            "error attempting to acquire build_router lock: " .. err
+          )
+        end
+
+        -- lock acquired but we might not need to rebuild the router (if we were
+        -- not the first request in this process to enter this code path)
+        -- check again and rebuild if necessary
+
+        version, err = cache:get("router:version", CACHE_ROUTER_OPTS, utils.uuid)
+        if err then
+          log(ngx.CRIT, "could not ensure router is up to date: ", err)
+
+        elseif router_version ~= version then
+          -- router needs to be rebuilt in this worker
+          log(DEBUG, "rebuilding router")
+
+          local ok, err = build_router(singletons.db, version)
+          if not ok then
+            router_err = err
+            log(ngx.CRIT, "could not rebuild router: ", err)
+          end
+        end
+
+        build_router_semaphore:post(1)
       end
 
       if not router then


p0pr0ck5:
I only wrapped this around `build_router` as `build_api_router` is due to be nixed I assume?

thibaultcha:
@p0pr0ck5 `build_api_router` is still around, and will be for yet another release; we'd need the semaphone around it as well

p0pr0ck5:
@thibaultcha ah gotcha thanks; wasnt 100% on the timeline. re-up'd with wrappers around both

kong/runloop/handler.lua
```diff
@@ -82,7 +87,6 @@ local function build_api_router(dao, version)
   return true
 end
 
-
```
thibaultcha: Mind re-adding this line back?

kong/runloop/handler.lua
```diff
@@ -408,6 +412,23 @@ return {
...
+        build_api_router_semaphore, err = semaphore.new()
+        if err then
+          log(ERR, "failed to create build_api_router_semaphore: ", err)
+        end
+        build_api_router_semaphore:post(1)
+
+        -- semaphore (mutex) for build_router
+        build_router_semaphore, err = semaphore.new()
+        if err then
+          log(ERR< "failed to create build_router_semaphore: ", err)
```
thibaultcha: A typo here

thibaultcha: I am wondering if those logs should have a higher level than `error`, and I believe so. Probably `crit`, as they would make the system operate in what we would consider as critical conditions for its current lifecycle.

kong/runloop/handler.lua
```diff
@@ -435,13 +456,45 @@ return {
...
+        -- lock acquired but we might not need to rebuild the api_router (if we were
+        -- not the first request in this process to enter this code path)
+        -- check again and rebuild if necessary
+
+        version, err = cache:get("api_router:version", CACHE_ROUTER_OPTS, utils.uuid)
+        if err then
+          log(ngx.CRIT, "could not ensure api_router is up to date: ", err)
+
+        elseif api_router_version ~= version then
+
```
thibaultcha: Mind getting rid of this blank line (in both Routes and APIs code paths)?

kong/runloop/handler.lua
p0pr0ck5: yeah, CRIT definitely makes sense. Done :+1:

kong/runloop/handler.lua
p0pr0ck5: Whups, unintentional. Fixed

kong/runloop/handler.lua
p0pr0ck5: done!

kong/runloop/handler.lua
```diff
@@ -408,6 +413,23 @@ return {
         balancer.init()
       end)
 
+      do
+        local err
+
+        -- semaphore (mutex) for build_api_router
+        build_api_router_semaphore, err = semaphore.new()
+        if err then
+          log(ERR, "failed to create build_api_router_semaphore: ", err)
```
thibaultcha: This log is still using the `ngx.ERR` level instead of `CRIT` - mind updating it?

p0pr0ck5: Shoot, my bad!

kong/runloop/handler.lua
```diff
@@ -435,13 +457,44 @@ return {
...
+        -- api_router version again to prevent unnecessary subsequent rebuilds
+        --
+        -- note the ten-second wait() timeout. we should be able to safely assume
+        -- that a api_router rebuild event should never take this long
+
+        local ok, err = build_api_router_semaphore:wait(10)
         if not ok then
-          api_router_err = err
-          log(ngx.CRIT, "could not rebuild API router: ", err)
+          return responses.send_HTTP_INTERNAL_SERVER_ERROR(
```
thibaultcha: As I was going through one last go over this, I started wondering if the best course of action in this scenario would be to log a warning/error, and fallthrough to the cache checking/building. We would still prevent stampede effects to the database in almost all cases, but if a situation ever occurs where a particular rebuild gets slowed down, we don't prevent another request to rebuild the router on its own - on Cassandra, a subsequent router rebuild would hit another peer, for instance. And a client with a large enough timeout would avoid having to deal with a 500. At the very least, I think we should not send this error message to the client (as done by `send_HTTP_INTERNAL_SERVER_ERROR(err)`), but just logging it for the administrator to see.

thibaultcha: > This is already the case; send_HTTP_INTERNAL_SERVER_ERROR does not write the err message to the client:

Oh yes, that's great then! I should know, I wrote this module... -_-

> I'm not sure I see the value in retrying the router build following a mutex acquisition failure

The point that I am trying to convey here is that we should follow a "best effort" practice with regards to the proxy's resiliency - at all times. I feel like hard-coding the semaphore timeout at 10 seconds is very arbitrary, and combined with the hard failure approach taken in case of a timeout, becomes a recipe that is bound to at some point provoke 500 errors for clients, when we could maybe avoid it.
When confronted to this sort of problem (need for a retry, which could cause a cascading failure), I usually think of implementing the proper retry policy. Something much more precautionary with regards to retrying a failed operation (with the proper retry delays, ideally increasing exponentially). In this particular case (and Saturday morning), I failed to elaborate, and wanted to spark a discussion about the topic in general (making the best effort).

In this particular case, I think that granting one or a few more resources to the semaphore (instead of suddenly unlocking every waiting request) would allow for another request to attempt rebuilding the router, and maybe successfully so, thus preventing all waiting requests from resulting in a 500 just because the first try failed (or is lagging behind).
Granted, 10 seconds is a large timeout in the first place, but:

1. most clients would be willing to wait up to 30s, which would allow for the "retry" to work
2. the retry approach would allow us to reduce the initial timeout value, and trigger the "retry mode" faster

Again, a retry mode can be useful because it would connect to a different Cassandra peer than the first one (due to the load balancing policy), and it certainly wouldn't hurt to retry a query on postgres either, especially if some middleware is involved between Kong and the instance, as it can be the case. The "best-practise" approach (that we need to follow in the proxy, but even more so in the core, and even more so in the router component), combined with the above arguments, kinds of add up and made me raise this point.

p0pr0ck5: Talked IRL yesterday; my concern here is that introducing complex retry logic wrapped around sempahore resources introduces more possibility for buggy code- if we make a mistake and lose track of something, we could end up with very odd behavior (or lose the ability for a worker to acquire a resource at all). The solution we've potentially settled on is keeping the timeout + mutex logic, and using the Kong configuration's cassandra_timeout when Kong is talking to Cassandra (for Postgres, we can use a default value of 60s since pgmoon doesn't support setting socket timeouts).

Gonna sit on this today and think it through, then will update the PR accordingly.

thibaultcha: @bungle Oh nice! And... I just realized _I_ contributed this feature to pgmoon a couple years ago, and I totallty forgot about it and did not document it nor implemented the option in Kong :man_facepalming: :man_facepalming: :man_facepalming: Damn...

thibaultcha: Docs PR for `pgmoon:settimeout()`: https://github.com/leafo/pgmoon/pull/75

In the meantime, Kong does not have a `pg_timeout` config option, so we can still go ahead with the plan described by @p0pr0ck5 until we add support for it.

Thanks @bungle!

p0pr0ck5: @thibaultcha both `:wait` calls have been updated accordingly

bungle:
Do we also want this to be rebased on `next`?

thibaultcha:
This is consistently failing CI when rebased on top of next (https://travis-ci.org/Kong/kong/builds/434843989), so I reverted the merge attempt, and pushed it into a `next-merge-mutex` for now, until I or someone else can investigate.

thibaultcha:
That was it:
```diff
diff --git a/kong/runloop/handler.lua b/kong/runloop/handler.lua
index 188310d5..a8e9e102 100644
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -496,7 +496,7 @@ return {
           -- api_router needs to be rebuilt in this worker
           log(DEBUG, "rebuilding api_router")
 
-          local ok, err = build_api_router(singletons.db, version)
+          local ok, err = build_api_router(singletons.dao, version)
           if not ok then
             api_router_err = err
             log(ngx.CRIT, "could not rebuild api_router: ", err)
```

thibaultcha:
Merged!

PR #3784, opened 2018-09-17
fix(request-termination) allow null value for body or message

shashiranjan84:
### Summary
Allow user to unset either message or body.

### Full changelog
- update schema self_check to support null value for `message` or `body`
- add relevant spec

### DIFF
--- a/kong/plugins/request-termination/schema.lua
+++ b/kong/plugins/request-termination/schema.lua
@@ -14,14 +14,18 @@ return {
       end
     end
 
-    if plugin_t.message then
-      if plugin_t.content_type or plugin_t.body then
+    if plugin_t.message and plugin_t.message ~= ngx.null then
+      if plugin_t.content_type or (plugin_t.body and plugin_t.body ~= ngx.null) then
         return false, Errors.schema("message cannot be used with content_type or body")
       end
     else
       if plugin_t.content_type and not plugin_t.body then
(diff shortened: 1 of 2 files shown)


thibaultcha:
In the `next` branch, we already ported plugins to the new DAO, and we currently are in the process of convert plugins schemas to the new schema library (https://github.com/Kong/kong/pull/3778). The new schema library/Admin API endpoints should support JSON `null` un-setting out of the box. Still, it is nice to have tests for this.

1. We should probably wait until #3778 is done, before reworking this PR to reflect the desired changes, and introduce those new tests
2. There should be a test that makes two subsequent requests, the first one un-setting `message`, and the second one un-setting `body`). That should not be possible (as per the "either one of message or body should be set" error).

thibaultcha:
"WIP" label applied to reflect the need for #3778 to be merged first, before porting those changes to the new schema. Keeping the PR open in the meantime!

shashiranjan84:
> There should be a test that makes two subsequent requests, the first one un-setting message, and the second one un-setting body). That should not be possible (as per the "either one of message or body should be set" error).

good catch, wondering if it possible to figure out what plugin state saved in DB from self check?

shashiranjan84:
Also we needed this fix  for Kong-ee 0.34, which would be based of Kong 1.0. But for now I will create a separate PR for it.

PR #3787, opened 2018-09-18
fix(aws-plugin) handle empty json arrays

kidd:
Empty tables are decoded as empty json arrays instead of emtpy
objects.

### DIFF
--- a/kong/tools/public.lua
+++ b/kong/tools/public.lua
@@ -8,7 +8,9 @@ local _M = {}
 
 do
   local multipart = require "multipart"
-  local cjson     = require "cjson.safe"
+  local cjson     = (require "cjson.safe").new()
+  cjson.decode_array_with_array_mt(true)
+
   local utils     = require "kong.tools.utils"
 
 
--- a/spec/03-plugins/23-aws-lambda/01-access_spec.lua
+++ b/spec/03-plugins/23-aws-lambda/01-access_spec.lua
@@ -273,6 +273,19 @@ for _, strategy in helpers.each_strategy() do
       assert.is_string(res.headers["x-amzn-RequestId"])
       assert.equal("some_value_json1", body.key1)
     end)
+    it("passes empty json arrays unmodified", function()
+      local res = assert(proxy_client:send {
+        method  = "POST",
+        path    = "/post",
+        headers = {
+          ["Host"]         = "lambda.com",
+          ["Content-Type"] = "application/json"
+        },
+        body = '[{}, []]'
+      })
+      assert.res_status(200, res)
+      assert.equal('[{},[]]', string.gsub(res:read_body(), "\n",""))
+    end)
     it("invokes a Lambda function with POST and both querystring and body params", function()
       local res = assert(proxy_client:send {
         method  = "POST",


thibaultcha:
@kidd Thanks for the PR. I believe we would need tests as well? This is a tricky fix and a test would certainly prevent a future regression, especially when we remove `kong.tools.public` and have to switch to the PDK (which does not implement this feature yet and we must not forget so).

aloisbarreras:
My knowledge of concurrency in lua and how nginx handles worker threads is still new, so this may be an unfounded concern, but could modifying the global cjson object cause unintended side effects? For example, setting `cjson.decode_array_with_array_mt(true)` and then another thread uses the cjson object before it's set back to false.

Should we create a local instance of cjson here with `cjson.new`? We could modify the `public.get_body_args` function to accept an optional cjson instance parameter and use that, otherwise defaulting to the original behavior of using main one.

thibaultcha:
Doing some digging, it turns out that `ngx.req.read_body()` can actually yield. In which case, modifying the global cjson module isn't recommended indeed. That said, to be perfectly honest, `decode_array_with_array_mt(true)` is actually something that we'd want at all times. It _shouldn't_ be breaking anything if we were to turn it on by default, but it also isn't something that really is in the scope of this PR, so if we were to find any issues by enabling it globally, we would put the burden of fixing it upon this PR, which we normally try to avoid.

> Should we create a local instance of cjson here with `cjson.new`?

That is a great suggestion, and probably how we should fix this moving forward. However, `public.get_body_args()` is deprecated and is scheduled for removal for the final 1.0 release, so it isn't recommended to make modifications to it at this point in time. Instead, plugins should rely on [kong.request.get_body()](https://docs.konghq.com/latest/pdk/kong.request/#kong_request_get_body). 

There isn't a way to specify a custom cjson instance for the PDK to use. We do have plans for a JSON library to be included in the PDK, and I can see us developing an interface making per-plugin instances and customization (e.g. `kong.json.new(); kong.json.decode_array_with_array_mt(true)`), but that does not exist yet... @kidd @aloisbarreras Is that something one of you would be willing to explore maybe?

kidd:
To keep the scope of this PR narrow, and keep the changes localized to this particular use case, what do you think about "manually" tagging the body_args in the aws-lambda plugin itself?

```lua
   function tag_table(t)
     for i,v in pairs(t) do
       if type(v) == "table" then
         t[i] = tag_table(v)
         if not next(v) then
           setmetatable(t[i], cjson.empty_array_mt)
         end
       end
     end
     return t
   end
```

This passes the tests and doesn't rush into defining interfaces for the json pdk. At some point we will use the de-facto solution provided by the pdk, but this PR should not require to change when the pdk json supports the array_mt.

thibaultcha:
@kidd The issue with this approach is that it would unconditionally set the `empty_array_mt` metatable on _all_ table elements, but that may not be what we want.

For example, take this script:
```lua
local cjson = require "cjson"

local function tag_table(t)
    for i,v in pairs(t) do
        if type(v) == "table" then
            t[i] = tag_table(v)
            if not next(v) then
                setmetatable(t[i], cjson.empty_array_mt)
            end
        end
    end
    return t
end

local json = [[
{
    "tags": [],
    "meta": {}
}
]]

-- bad
local t = cjson.decode(json)
tag_table(t)
print(cjson.encode(t))

-- good
t = cjson.decode(json)
print(cjson.encode(t))
```

Here is the result:
```
$ resty test.lua
{"tags":[],"meta":[]}
{"tags":[],"meta":{}}
```

So this change could be breaking... This is why `decode_array_with_array_mt()` was a crucial addition to lua-cjson.

thibaultcha:
Just a note: the `kong.tools.public` utility is also used in the oauth2 plugin - although this should be ok.

thibaultcha:
This should also become two commits now:
```
feat(utils) set 'decode_array_with_array_mt' on get_body_*()
tests(aws-lambda) ensure JSON arrays are re-encoded as such
```

kidd:
reworked the commits as asked. Hope it's ok.

PR #3792, opened 2018-09-19
hotfix(db) postgres ttl cleanup timer

bungle:
### Summary

Fixes unfortunate bug in Postgres `TTL` cleanup  timer where it tried to delete rows from tables that didn't have `ttl=true`.

### DIFF
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -198,7 +198,7 @@ function _mt:init_worker(strategies)
         local name = schema.name
         graph:add(name)
         for _, field in schema:each_field() do
-          if field.type == "foreign" then
+          if field.type == "foreign" and field.schema.ttl then
             graph:add(name, field.schema.name)
           end
         end


thibaultcha:
@bungle I'd be useful to see the error/stacktrace that this sort of fixes help prevent! This way this gives better insight into the fix to the reviewer/reader.

bungle:
@thibaultcha here you go:

```
2018/09/20 10:39:02 [warn] 78048#0: *296 [lua] connector.lua:235: unable to clean expired rows from postgres database (ERROR: column "ttl" does not exist (174)), context: ngx.timer
```

bungle:
I think this only showed up after we moved plugins to new dao.

bungle:
Also this:

```
2018/09/20 10:42:02 [warn] 78048#0: *1173 [lua]
  connector.lua:235: unable to clean expired rows from postgres database
  (ERROR: current transaction is aborted, commands ignored until end of transaction block),
  context: ngx.timer

2018/09/20 10:42:02 [error] 78048#0: *1177 [lua]
  postgres.lua:230: [postgres] could not cleanup TTLs: ERROR: current transaction is aborted, 
  commands ignored until end of transaction block, context: ngx.timer
```

Issue #3768, opened 2018-09-14
Kong excessively ERR logs on OAuth2.0 protected non json/form url encoded HTTP requests


jeremyjpj0916:
### Summary

As noted here:
https://discuss.konghq.com/t/seeing-this-error-flood-our-kong-logs-lately/1864 

### Steps To Reproduce

1. Make proxy with some of the plugins(I don't think its because of plugins this error occurs so you may be able to leave them off.)

"
Plugins existing directly on this route/service:
acl
oauth2

Global plugins:
Variant of the HTTP Logger
StatsD
Rate Limiting set to local 500 tps
Kong upstream JWT: https://github.com/Optum/kong-upstream-jwt 
correlation id
"
2. Send traffic to said proxy with Request Header and some dummy xml http body:
Content-Type 'text/xml; charset=utf-8' or Content-Type 'application/xml'

3. Watch the Kong error logs start flooding stdout per tx about how it does not know how to parse the request body.

```
2018/09/14 01:24:20 [error] 36#0: *14978818 [lua] public.lua:139: get_body_info(): don't know how to parse request body (can't decode Content-Type 'application/xml'), client: 10.xxx.x.x, server: kong, request: "POST /api/test/service/v1 HTTP/1.1", host: "gateway.company.com"
```

Here is what I propose for changing public.lua (my first take at a quick glance public.lua's main point really is informational logging about the request):

Replace
https://github.com/Kong/kong/blob/master/kong/tools/public.lua#L3

With local DEBUG = ngx.DEBUG and then all instances of ERR with DEBUG so this only shows up in logger when running Kong in debug.

That way if someone runs Kong in debug they can see they have tx's flowing through them that Kong can't necessarily do HTTP Body transformations on and other things of that nature. But the benefit here is now Kong is not just logging to be logging it when running in notice mode(which is what I run as for a  normal mode). We have lots of JSON / XML / Multi part file based API's and currently the few XML services we have are really running up our stdout logs with unhelpful logs classified as Errors (As the tcp level nginx errors thrown are what I really like to see in that stdout). 

It may be even better to wrap those log lines with an if check of sorts so the log line does not even get executed unless in debug mode. 

And I think it makes more sense to log this as a DEBUG because its not an error per say with the proxy and nothing broke with kong or nginx, its debuggable info for users who wanna figure out what some service transacting against their gateway can't support having a request transformer plugin which expects json to manipulate xml.

Thoughts on the approach? I could PR the change if it seems agreeable.

[...]

PR #3793, opened 2018-09-20
perf(oauth2) do not read request body if unnecessary

jeremyjpj0916:
Improve OAuth2.0 by checking headers first and not reading every request body into memory by default. 

Also is a nice fix for https://github.com/Kong/kong/issues/3768 since everyone within our company uses token in the header(Which IMO for the record is the right way 😄).

### Summary

OAuth2.0 Plugin currently reads the request body to parse for an access_token prematurely, this PR optimizes performance in the event users are not sticking their access_token in the HTTP Body but rather as a Header.

### Full changelog
* modify access.lua  parse_access_token() function for improvements. 

### Issues resolved
Fixes https://github.com/Kong/kong/issues/3768

Existing tests with the plugin will validate out this logic since nothing changed, otherwise if for some reason newer tests must be made let me know.

### DIFF
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -450,19 +450,21 @@ end
 
 local function parse_access_token(conf)
   local found_in = {}
-  local result = retrieve_parameters()["access_token"]
-  if not result then
-    local authorization = ngx.req.get_headers()[conf.auth_header_name]
-    if authorization then
-      local parts = {}
-      for v in authorization:gmatch("%S+") do -- Split by space
-        table.insert(parts, v)
-      end
-      if #parts == 2 and (parts[1]:lower() == "token" or parts[1]:lower() == "bearer") then
-        result = parts[2]
-        found_in.authorization_header = true
-      end
+  local access_token = ngx.req.get_headers()[conf.auth_header_name]
+  if access_token then
+    if type(access_token) == "table" then --Take the first found
+      access_token = access_token[1]
     end
+    local parts = {}
+    for v in access_token:gmatch("%S+") do -- Split by space
+      table.insert(parts, v)
+    end
+    if #parts == 2 and (parts[1]:lower() == "token" or parts[1]:lower() == "bearer") then
+      access_token = parts[2]
+      found_in.authorization_header = true
+    end
+  else
+    access_token = retrieve_parameters()[ACCESS_TOKEN]  
   end
 
   if conf.hide_credentials then
@@ -489,7 +491,7 @@ local function parse_access_token(conf)
     end
   end
 
-  return result
+  return access_token
 end
 
 local function load_oauth2_credential_into_memory(credential_id)
--- a/spec/03-plugins/26-oauth2/03-access_spec.lua
+++ b/spec/03-plugins/26-oauth2/03-access_spec.lua
@@ -1718,6 +1718,19 @@ for _, strategy in helpers.each_strategy() do
         })
         assert.res_status(200, res)
       end)
+      it("works when a correct access_token is being sent in duplicate custom headers", function()
+        local token = provision_token("oauth2_11.com",nil,"clientid1011","secret1011")
+
+        local res = assert(proxy_ssl_client:send {
+          method = "GET",
+          path = "/request",
+          headers = {
+            ["Host"] = "oauth2_11.com",
+            ["custom_header_name"] = { "bearer " .. token.access_token, "bearer " .. token.access_token },
+          }
+        })
+        assert.res_status(200, res)
+      end)
       it("fails when a correct access_token is being sent in the wrong header", function()
         local token = provision_token("oauth2_11.com",nil,"clientid1011","secret1011")
 


kong/plugins/oauth2/access.lua
```diff
@@ -450,19 +450,18 @@ end
...
-    if authorization then
-      local parts = {}
-      for v in authorization:gmatch("%S+") do -- Split by space
-        table.insert(parts, v)
-      end
-      if #parts == 2 and (parts[1]:lower() == "token" or parts[1]:lower() == "bearer") then
-        result = parts[2]
-        found_in.authorization_header = true
-      end
+  local access_token = ngx.req.get_headers()[conf.auth_header_name]
```
bungle: While you are at it, could you also fix issue with this code where it actually crashes runtime, if you happen to send the header more than once. Making `local access_token` be `table` instead of string. What to do in such case, well just pick the first for example `access_token[1]`.

jeremyjpj0916: Crashed when you tested? I am not able to reproduce such a failure, do I have to run in debug mode to see it?

bungle: I think on line 456:
```lua
 for v in access_token:gmatch("%S+") do
```

If the access token is actually a table it will crash, as `table`s in Lua don't have `gmatch`. It is something that is only in `string`.

As you see here:
https://github.com/openresty/lua-nginx-module#ngxreqget_headers

The example where it sends headers:
```http
 Foo: foo
 Foo: bar
 Foo: baz
```

The `ngx.req.get_headers()["Foo"] ` will return a table instead of string containing one of those values:
```lua
 {"foo", "bar", "baz"}
```

There is other way, which is `ngx.var.http_Foo` that always returns a single value, but I think using `ngx.req.get_headers()` there is fine.

bungle: What I mean is to add code like this:
```
if type(access_token) == "table" then
  access_token = access_token[1]
end
```

To make code more bullet-proof.

jeremyjpj0916: You prefer the above extra check rather than forcing it to 1(requires no extra conditional/assignment?:
access_token = ngx.req.get_headers(1)["Foo"] ? 

I am going to retest again too in debug because I never saw that failure print out to stdout with the lua code, I would have expected to see something of a failure on the gmatch like you mentioned.

bungle: `max_headers` parameter is not about how many `Foo` headers you get. It is just how many headers you will read in total.

jeremyjpj0916: Oh, right duh that makes more sense. I just wanted to get fancy with it :P , testing again now just so I can produce the failure too then will go with the table check suggestion.

jeremyjpj0916: Yeah tested my code again with debug enabled too, I see this in logs when I run through it:

```
2018/09/20 19:58:22 [info] 33#0: *1692 client sent duplicate header line: "Authorization: Bearer SHj7zxYf7mhcyKAb5uLG1A1FtzbtG93Y", previous value: "Authorization: Bearer SHj7zxYf7mhcyKAb5uLG1A1FtzbtG93Y" while reading client request headers, client: 10.xxx.xx.x, server: kong, request: "POST /api/fasthttp/v1.0 HTTP/1.1"
```

But that was it for logs so must be a silent failure to not make it to docker std out, regardless I am fine with just taking the first and continuing the call. Thx.

jeremyjpj0916: @bungle Mr. Bungle sir I believe I have somehow managed to fix the PR with the changes and maintain one commit with a fixed commit message as per Contribution guidelinse as requested by @Tieske earlier using a combination of git commit --amend's and git push -f . Let me know if anything else is off. And I think you are spot on with that evaluation, I was leaving it to the default and nginx must be handling that gracefully before lua land was seeing it.

Thanks!

thibaultcha: @jeremyjpj0916 Appreciate the effort :) The proper prefix however is `perf(oauth2)` (as per the [contributing docs](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format): `<type>(<scope>)`. We also enforce a non-capitalization of the git message headers :)

Ideally, I believe that a clearer message for the intent behind this commit would be something similar to:
```
perf(oauth2) do not read request body if unnecessary

`access_token` retrieval used to unconditionally read the request body.
We now evaluate if the token is contained in the request headers before
attempting to read the body.

Fix ...
```

This way, the commit message is roughly organized as:
```
<type>(<scope>) <intent>

<before patch>
<new behavior>

<links>
```

jeremyjpj0916: I think I modified everything to your request 👍  . It may not be today, or tomorrow, or within the month. But one day I hope I can aspire to get it right well enough that my time spent editing the commit messages and PR takes less time that the code itself lol.

thibaultcha: @jeremyjpj0916 Thank you! I do spend time editing the commit messages/style issues myself when I find some in PRs, but:

1. I am a fervent believer that contributors should learn how to "fit" into a codebase/process and distillate the knowledge and practices themselves to future newcomers :)
2. If you are to contribute more and more frequently, "teach a man how to fish..." yada yada, right? :)

kong/plugins/oauth2/access.lua
```diff
@@ -450,19 +450,21 @@ end
...
+  if access_token then
+    if type(access_token) == "table" then --Take the first found
+      access_token = access_token[1]
```
thibaultcha: This sort of change (which is a great addition!) would require a test to protect ourselves from future regressions.

thibaultcha:
@jeremyjpj0916 I agree with @bungle that this PR is ready in practice, but would also ask of you that you provide a test to ensure we don't regress on the fix that you implemented as well. _Ideally_, two commits should be made here: `fix` (with the regression test), and the `perf` commit. Thank you!

kong/plugins/oauth2/access.lua
jeremyjpj0916: Thought this might pop up as I was thinking about that latest enhancement. Would a test showing how a request containing 2 of the same correct Bearer token headers yielding success be enough? Something similar to the lines of https://github.com/Kong/kong/blob/master/spec/03-plugins/26-oauth2/03-access_spec.lua#L1334 
```
        it("returns success with multiple custom duplicate authorization headers", function()
          local res = assert(proxy_ssl_client:send {
            method  = "POST",
            path    = "/oauth2/token",
            body    = {
              provision_key        = "provision123",
              authenticated_userid = "id123",
              scope                = "email",
              grant_type           = "password"
            },
            headers = {
              ["Host"]             = "oauth2_5.com",
              ["Content-Type"]     = "application/json",
              CustomAuthHeader = "Basic Y2xpZW50aWQxMjM6c2VjcmV0MTIz",
              CustomAuthHeader = "Basic Y2xpZW50aWQxMjM6c2VjcmV0MTIz"
            }
          })
          assert.is_table(ngx.re.match(body, [[^\{"refresh_token":"[\w]{32,32}","token_type":"bearer","access_token":"[\w]{32,32}","expires_in":5\}$]]))
        end)
```

Since we know nginx handles dup Authorization headers itself only time the code prior would have failed would be on a declared custom Authorization header in the oauth2 plugin config.

thibaultcha: @jeremyjpj0916 Yes, this is the test I had in mind. However, in order to specify the same header multiple times, you would have to specify it as such:
```lua
headers = {
  ["Header-Name"] = { "val1", "val2" },
}
```

jeremyjpj0916: Feeling this snippit:
```
      it("works when a correct access_token is being sent in duplicate custom headers", function()
        local token = provision_token("oauth2_11.com",nil,"clientid1011","secret1011")

        local res = assert(proxy_ssl_client:send {
          method = "GET",
          path = "/request",
          headers = {
            ["Host"] = "oauth2_11.com",
            ["custom_header_name"] = { "bearer " .. token.access_token, "bearer " .. token.access_token },
          }
        })
        assert.res_status(200, res)
      end)
```

jeremyjpj0916:
Made another commit with the additional test(hopefully it passes 😄). Hope everyone enjoys their weekend!

PR #3794, opened 2018-09-20
perf(plugins-iterator) execute iterator only on configured plugins

p0pr0ck5:
### Summary

Reduce the number of invocations of the plugins iterator by attempting
to load configurations only for plugins we know are configured on the
cluster. A map of distinct names of configs plugins is maintained by
iterating over each row in the plugins table. Plugins CRUD events are
handled and delete a node cache entry representing the current version
of the plugins map; in the request path this value is used to determine
if a rebuild of the map is necessary. This logic mimics that of the
router rebuild mechanism. Additionally, plugin maps rebuilds are wrapped
in a worker-level mutex (implemented via ngx.semaphore); this prevents
stampeding database connections during rebuild events in high-concurrency
situations.

### Full Changelog

* Execute iterator only on configured plugins
* Add integration test to validate plugin map cache key invalidation

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -65,6 +65,7 @@ local utils = require "kong.tools.utils"
 local lapis = require "lapis"
 local runloop = require "kong.runloop.handler"
 local responses = require "kong.tools.responses"
+local semaphore = require "ngx.semaphore"
 local singletons = require "kong.singletons"
 local DAOFactory = require "kong.dao.factory"
 local kong_cache = require "kong.cache"
@@ -98,8 +99,94 @@ local ffi = require "ffi"
 local cast = ffi.cast
 local voidpp = ffi.typeof("void**")
 
+local plugins_map_version
+local plugins_map_semaphore
+
+local PLUGINS_MAP_CACHE_OPTS = { ttl = 0 }
+
+local configured_plugins
 local loaded_plugins
 local schema_state
+
+local function build_plugins_map(db, version)
+  local map = {}
+
+  for plugin, err in db.plugins:each() do
+    if err then
+      return nil, tostring(err)
+    end
+
+    map[plugin.name] = true
+  end
+
+  if version then
+    plugins_map_version = version
+  end
+
+  configured_plugins = map
+
+  return true
+end
+
+local function plugins_map_wrapper()
+  local version, err = kong.cache:get("plugins_map:version",
+                                      PLUGINS_MAP_CACHE_OPTS,
+                                      utils.uuid)
+  if err then
+    ngx_log(ngx_CRIT, "could not ensure plugins map is up to date: ", err)
+
+    return false
+
+  elseif plugins_map_version ~= version then
+    -- try to acquire the mutex (semaphore)
+
+    local timeout = 60
+    if singletons.configuration.database == "cassandra" then
+      -- cassandra_timeout is defined in ms
+      timeout = singletons.configuration.cassandra_timeout / 1000
+
+    elseif singletons.configuration.database == "postgres" then
+      -- pg_timeout is defined in ms
+      timeout = singletons.configuration.pg_timeout / 1000
+    end
+
+    local ok, err = plugins_map_semaphore:wait(timeout)
+
+    if ok then
+      -- we have the lock but we might not have needed it. check the
+      -- version again and rebuild if necessary
+
+      version, err = kong.cache:get("plugins_map:version",
+                                    PLUGINS_MAP_CACHE_OPTS,
+                                    utils.uuid)
+      if err then
+        ngx_log(ngx_CRIT, "could not ensure plugins map is up to date: ", err)
+
+        plugins_map_semaphore:post(1)
+
+        return false
+
+      elseif plugins_map_version ~= version then
+        -- we have the lock and we need to rebuild the map. go go gadget!
+
+        ngx_log(ngx_DEBUG, "rebuilding plugins map")
+
+        local ok, err = build_plugins_map(kong.db, version)
+        if not ok then
+          ngx_log(ngx_CRIT, "could not rebuild plugins map: ", err)
+        end
+      end
+
+      plugins_map_semaphore:post(1)
+    else
+      ngx_log(ngx_CRIT, "could not acquire plugins map update mutex: ", err)
+
+      return false
+    end
+  end
+
+  return true
+end
 
 -- Kong public context handlers.
 -- @section kong_handlers
@@ -232,6 +319,19 @@ function Kong.init()
   loaded_plugins = assert(db.plugins:load_plugin_schemas(config.loaded_plugins))
   sort_plugins_for_execution(config, db, loaded_plugins)
 
+  local err
+  plugins_map_semaphore, err = semaphore.new()
+  if not plugins_map_semaphore then
+    error("failed to create plugins map semaphore: " .. err)
+  end
+
+  plugins_map_semaphore:post(1) -- one resource, treat this as a mutex
+
+  local _, err = build_plugins_map(db, "init")
+  if err then
+    error("error building initial plugins map: ", err)
+  end
+
   assert(runloop.build_router(db, "init"))
   assert(runloop.build_api_router(dao, "init"))
 end
@@ -360,6 +460,14 @@ function Kong.init_worker()
     return
   end
 
+  local ok, err = cache:get("plugins_map:version", { ttl = 0 }, function()
+    return "init"
+  end)
+  if not ok then
+    ngx_log(ngx_CRIT, "could not set plugins map version in cache: ", err)
+    return
+  end
+
 
   -- LEGACY
   singletons.cache          = cache
@@ -396,7 +504,8 @@ function Kong.ssl_certificate()
 
   runloop.certificate.before(ctx)
 
-  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins, true) do
+  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins,
+                                              configured_plugins, true) do
     kong_global.set_namespaced_log(kong, plugin.name)
     plugin.handler:certificate(plugin_conf)
     kong_global.reset_log(kong)
@@ -508,10 +617,16 @@ function Kong.rewrite()
 
   runloop.rewrite.before(ctx)
 
+  local ok = plugins_map_wrapper()
+  if not ok then
+    return responses.send_HTTP_INTERNAL_SERVER_ERROR()
+  end
+
   -- we're just using the iterator, as in this rewrite phase no consumer nor
   -- api will have been identified, hence we'll just be executing the global
   -- plugins
-  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins, true) do
+  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins,
+                                              configured_plugins, true) do
     kong_global.set_named_ctx(kong, "plugin", plugin_conf)
     kong_global.set_namespaced_log(kong, plugin.name)
 
@@ -532,7 +647,8 @@ function Kong.access()
 
   ctx.delay_response = true
 
-  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins, true) do
+  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins,
+                                              configured_plugins, true) do
     if not ctx.delayed_response then
       kong_global.set_named_ctx(kong, "plugin", plugin_conf)
       kong_global.set_namespaced_log(kong, plugin.name)
@@ -564,7 +680,8 @@ function Kong.header_filter()
 
   runloop.header_filter.before(ctx)
 
-  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins) do
+  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins,
+                                              configured_plugins) do
     kong_global.set_named_ctx(kong, "plugin", plugin_conf)
     kong_global.set_namespaced_log(kong, plugin.name)
 
@@ -581,7 +698,8 @@ function Kong.body_filter()
 
   local ctx = ngx.ctx
 
-  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins) do
+  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins,
+                                              configured_plugins) do
     kong_global.set_named_ctx(kong, "plugin", plugin_conf)
     kong_global.set_namespaced_log(kong, plugin.name)
 
@@ -598,7 +716,8 @@ function Kong.log()
 
   local ctx = ngx.ctx
 
-  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins) do
+  for plugin, plugin_conf in plugins_iterator(ctx, loaded_plugins,
+                                              configured_plugins) do
     kong_global.set_named_ctx(kong, "plugin", plugin_conf)
     kong_global.set_namespaced_log(kong, plugin.name)
 
@@ -615,7 +734,7 @@ function Kong.handle_error()
 
   local ctx = ngx.ctx
   if not ctx.plugins_for_request then
-    for _ in plugins_iterator(ctx, loaded_plugins, true) do
+    for _ in plugins_iterator(ctx, loaded_plugins, configured_plugins, true) do
       -- just build list of plugins
     end
   end
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -113,7 +113,7 @@ local function build_plugins_map(db, version)
 
   for plugin, err in db.plugins:each() do
     if err then
-      return nil, tostring(err)
+      return nil, err
     end
 
     map[plugin.name] = true
@@ -133,9 +133,7 @@ local function plugins_map_wrapper()
                                       PLUGINS_MAP_CACHE_OPTS,
                                       utils.uuid)
   if err then
-    ngx_log(ngx_CRIT, "could not ensure plugins map is up to date: ", err)
-
-    return false
+    return nil, "failed to retrieve plugins map version: " .. err
 
   elseif plugins_map_version ~= version then
     -- try to acquire the mutex (semaphore)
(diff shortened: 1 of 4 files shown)


bungle:
@p0pr0ck5 It seems this is going to land on `0.15.0`, so can you please rebase on `next`? @thibaultcha, am I correct?

kong/init.lua
```diff
@@ -428,6 +496,56 @@ function Kong.balancer()
...
+                                            utils.uuid)
+  if err then
+    ngx_log(ngx.CRIT, "could not ensure plugins_map is up to date: ", err)
+
+    return false
+
+  elseif plugins_map_version ~= version then
+    -- try to acquire the mutex (semaphore)
+
+    local ok, err = plugins_map_semaphore:wait(10)
```
bungle: Should the timeout be configurable?

kong/init.lua
```diff
@@ -90,8 +91,56 @@ local set_current_peer = ngx_balancer.set_current_peer
...
+
+    if not rows then
+      return nil, tostring(err)
+    end
+
+    for _, row in ipairs(rows) do
+      map[row.name] = true
+    end
+
+    if offset == nil then
```
bungle: Can this ever be `false`?

kong/runloop/plugins_iterator.lua
```diff
@@ -92,6 +92,10 @@ local function get_next(self)
 
   self.i = i
 
+  if not self.configured_plugins[plugin.name] then
```
bungle: Wondering, can be somehow join `loaded_plugins` and `configured_plugins` so that we don't needlessly loop them?

kong/init.lua
```diff
@@ -90,8 +91,56 @@ local set_current_peer = ngx_balancer.set_current_peer
 local set_timeouts     = ngx_balancer.set_timeouts
 local set_more_tries   = ngx_balancer.set_more_tries
 
+local plugins_map_version
+local plugins_map_semaphore
+
+local PLUGINS_MAP_CACHE_OPTS = { ttl = 0 }
+local PLUGINS_MAP_PAGE_SIZE = 1000
```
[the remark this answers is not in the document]
p0pr0ck5: trying to avoid magic numbers :) a meaningful description is useful IMO?

kong/init.lua
```diff
@@ -323,6 +382,14 @@ function Kong.init_worker()
   kong.db:set_events_handler(worker_events)
   kong.dao:set_events_handler(worker_events)
 
+  plugins_map_version, err = cache:get("plugins_map:version",
+                                       PLUGINS_MAP_CACHE_OPTS,
+                                       function() return "init" end)
```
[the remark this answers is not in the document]
p0pr0ck5: same thing, that means declaring and sharing a function to return a string 'init', which doesnt seem useful

kong/init.lua
p0pr0ck5: yes

kong/runloop/plugins_iterator.lua
p0pr0ck5: i had initially considered that, but doing so makes the code very difficult to re-use. re-executing the loop here during initialization is worth it IMO to keep things sane

kong/init.lua
p0pr0ck5: Indeed there's some relevant discussion about the same kind of behavior at https://github.com/Kong/kong/pull/3782#discussion_r219679253; will update this PR to match that behavior once that's settled and approved.

kong/init.lua
bungle: `offset` can be `false`, `nil` or something else. ok.

kong/runloop/plugins_iterator.lua
bungle: Fair.

kong/init.lua
```diff
@@ -90,8 +91,56 @@ local set_current_peer = ngx_balancer.set_current_peer
 local set_timeouts     = ngx_balancer.set_timeouts
 local set_more_tries   = ngx_balancer.set_more_tries
 
+local plugins_map_version
+local plugins_map_semaphore
+
+local PLUGINS_MAP_CACHE_OPTS = { ttl = 0 }
```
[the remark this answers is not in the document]
bungle: What do you mean by different module?

https://github.com/Kong/kong/blob/perf/plugins-iterator/kong/init.lua#L354-L387

E.g.:
```lua
local ok, err = cache:get("router:version", { ttl = 0 }, function()
    return "init"
  end
```

Could be written, as this repeates in 3 places (ok, api_router is going away) now:
```lua
local ok, err = cache:get("router:version", CACHE_OPTS, CACHE_INIT)
```

And that:
```lua
  plugins_map_version, err = cache:get("plugins_map:version",
                                       PLUGINS_MAP_CACHE_OPTS,
                                       function() return "init" end)
```

changed to
```lua
  plugins_map_version, err = cache:get("plugins_map:version", CACHE_OPTS, CACHE_INIT)
```

But not a big deal. Just repeating here so that we were in a same page.

p0pr0ck5:
@bungle just FYI this is rebased off `next`. Seeing how CI plays with it and then will adjust the remaining hard-coded semaphore timeout in a bit.

kong/init.lua
```diff
@@ -98,9 +99,85 @@ local ffi = require "ffi"
...
+                                      utils.uuid)
+  if err then
+    ngx_log(ngx_CRIT, "could not ensure plugins map is up to date: ", err)
+
+    return false
+
+  elseif plugins_map_version ~= version then
+    -- try to acquire the mutex (semaphore)
+
+    local ok, err = plugins_map_semaphore:wait(10)
```
p0pr0ck5: I need to update this to use the same logic presented in #3820

p0pr0ck5:
@bungle this is ready for another round of review following rebasing to `next` :)

kong/init.lua
```diff
@@ -98,9 +99,95 @@ local ffi = require "ffi"
...
+local PLUGINS_MAP_CACHE_OPTS = { ttl = 0 }
+
+local configured_plugins
 local loaded_plugins
 local schema_state
 
+local function build_plugins_map(db, version)
+  local map = {}
+
+  for plugin, err in db.plugins:each() do
```
thibaultcha: I am concerned by this, in the sense that it can significantly increase the startup time for Kong. Routes and Services in Cassandra are partitioned together (i.e. they reside on the same node), but Plugins are not partitioned. This means that for every 100 plugins, we will make subsequent queries to the cluster to retrieve all rows from all partitions. On top of being potentially slow (we know of some users having in the (tens of) thousands of plugins), this also increases the chances of a failure to start due to a query timeout or such. We are most likely not in a position to experience any pain from this change, but I am fairly confident that some users will.

The impact of the query at runtime is less concerning I believe, as it will rarely be needed to rebuild the plugins map in actual production usage (and other operations esp. in the load balancer are making use of a complete iterator like this one as well).

I would suggest making a custom query for this operation (the new DAO is quite extensible and supports such custom operation per entity), and that would be helpful for PostgreSQL as well, **however**, I am really unsure there actually _is_ any improved query we could make for Cassandra... I am referring to something like:

```
SELECT DISTINCT name from plugins;
```

Unfortunately, our partitioning of the plugins column family isn't allowing us to do so in Cassandra... Ugh. I haven't found a workaround quite yet, or thought of an alternative approach.

What concerns do you have about the concern raised here?

kong/init.lua
```diff
@@ -98,9 +99,95 @@ local ffi = require "ffi"
...
+    if err then
+      return nil, tostring(err)
```
thibaultcha: The new DAO guarantees that this value will be a string (no trick, ever, promise!). It is one of its design goals to get rid of such patterns like `tostring(err)` on the consumer-side (from the caller's POV). Can we remove the `tostring` wrap?

kong/init.lua
```diff
@@ -98,9 +99,95 @@ local ffi = require "ffi"
...
+          ngx_log(ngx_CRIT, "could not rebuild plugins map: ", err)
+        end
+      end
+
+      plugins_map_semaphore:post(1)
+    else
+      ngx_log(ngx_CRIT, "could not acquire plugins map update mutex: ", err)
+
+      return false
+    end
```
thibaultcha: Would you mind hoisting this branch up below the `:wait()` call? We always try to write failure branches first (which are much shorter), as per our style guide, and the rest of this codebase. Also, this avoids the need for two branches to be written, since instead, we can do:
```lua
local ok, err = ...
if not ok then
  -- ...
  return
end

-- rest goes here, unindented
```

kong/init.lua
```diff
@@ -508,10 +617,16 @@ function Kong.rewrite()
 
   runloop.rewrite.before(ctx)
 
+  local ok = plugins_map_wrapper()
```
thibaultcha: Any reason why we are not invoking this from `ssl_certificate` as well? My understanding is that `rewrite` catches non-SSL connections (that's great), but when a new connection is opened just after a new plugin was configured, then are we failing to execute it during the SSL handshake?

kong/init.lua
```diff
@@ -98,9 +99,95 @@ local ffi = require "ffi"
...
+  if version then
+    plugins_map_version = version
+  end
+
+  configured_plugins = map
+
+  return true
+end
+
+local function plugins_map_wrapper()
```
thibaultcha: The design of this function is slightly unaligned with our current best practices. 

Instead of logging and returning `false`, it'd be preferable if it were returning `nil, err`.
This way, the caller decides what to do of the error message, and in this particular case, at this particular time, we want to log it with a `CRIT` level, indeed.

So in the current form of this patch, the `rewrite` phase call would be written as:
```lua
local ok, err = plugins_map_wrapper()
if not ok then
  ngx.log(ngx.CRIT, "could not ensure plugins map is up to date: ", err)
  return HTTP_500
end
```

(This came from a weird feeling after reading the `local ok =  build_map(); if not ok then` branch of the `rewrite` phase, and I think this concern explains why this original feeling).

And this function returns one of those:

* `return nil, err`
* `"return nil, failed to acquire plugins map mutex"`

spec/02-integration/06-invalidations/03-plugins_map_invalidation_spec.lua
```diff
@@ -0,0 +1,262 @@
...
+      -- insert single fixture Service
+      service_fixture = bp.services:insert()
+
+      local db_update_propagation = strategy == "cassandra" and 0.1 or 0
+
+      assert(helpers.start_kong {
+        log_level             = "debug",
+        prefix                = "servroot1",
+        database              = strategy,
+        proxy_listen          = "0.0.0.0:8000, 0.0.0.0:8443 ssl",
```
thibaultcha: side note: ouch, it is a realization to me that other existing test suites already make use of the default ports used by Kong, which breaks the test suite when a development instance is running... No action required here, just a note...

spec/02-integration/06-invalidations/03-plugins_map_invalidation_spec.lua
```diff
@@ -0,0 +1,262 @@
...
+      proxy_client_1 = helpers.http_client("127.0.0.1", 8000)
+      proxy_client_2 = helpers.http_client("127.0.0.1", 9000)
+
+      wait_for_propagation = function()
+        ngx.sleep(POLL_INTERVAL * 2 + db_update_propagation * 2)
+      end
+    end)
+
+    teardown(function()
+      helpers.stop_kong("servroot1", true)
```
thibaultcha: The `true` arguments here are a debug-only variable, the current policy in tests is to cleanup servroot prefixes after the tests run. Mind removing them?

kong/init.lua
```diff
@@ -98,9 +99,95 @@ local ffi = require "ffi"
...
+        return false
+
+      elseif plugins_map_version ~= version then
+        -- we have the lock and we need to rebuild the map. go go gadget!
+
+        ngx_log(ngx_DEBUG, "rebuilding plugins map")
+
+        local ok, err = build_plugins_map(kong.db, version)
+        if not ok then
+          ngx_log(ngx_CRIT, "could not rebuild plugins map: ", err)
```
thibaultcha: Is this branch lacking a `return` statement? It seems like it will fallthriough to the `return true` at the bottom of this function, and not thus not trigger then unhappy code path of the caller?

spec/02-integration/06-invalidations/03-plugins_map_invalidation_spec.lua
bungle: Yes, I have hit this many times :-(.

kong/init.lua
p0pr0ck5: This was a copy paste leftover from old code. Will fix.

kong/init.lua
p0pr0ck5: Gooood catch. I didn’t think about this, largely because we have no plugins that uses ssl phase. 

Tangentially- to my understanding the ssl phase historically was provided for plugins so we could use dynamic ssl before ssl_by_lua was a Thing. Given that we have certs and snis as first class objects, should this phase handler for plugins go away?

spec/02-integration/06-invalidations/03-plugins_map_invalidation_spec.lua
p0pr0ck5: This test was a copypasta of the boilerplate code in https://github.com/Kong/kong/blob/next/spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua.

spec/02-integration/06-invalidations/03-plugins_map_invalidation_spec.lua
p0pr0ck5: Happy to. Again as a note, this test copies the boilerplate from https://github.com/Kong/kong/blob/next/spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua

spec/02-integration/06-invalidations/03-plugins_map_invalidation_spec.lua
thibaultcha: Yep, that's how I realized we were already using those ports for existing test suites.

kong/init.lua
bungle: Yes, what you suggested for C* is basically what you usually do with proper NoSQL design, e.g. you have to start denormalize when you need to have read performance. But atm I do not see us doing that a lot as it introduces insert/update complexities.

About custom strategies, here is example:
https://github.com/Kong/kong/blob/master/kong/db/strategies/cassandra/routes.lua

And here is example of doing generic dao customization:
https://github.com/Kong/kong/blob/master/kong/db/dao/consumers.lua

hbagdi: Just leaving a thought and not a review.

The concerns raised here are valid but if I'm not wrong the pattern already exists in the code.

While going through this thread, I recalled that when Kong starts up, we check if all plugins configured in DB are enabled or not, to guard against:
- A custom plugin which was enabled before but is not anymore
- With `plugins` property in Kong 0.14.0, someone could disable a bundled plugin after configuring it in Kong.

In both cases, Kong will fail to start up.

The code I'm referring to in master branch is [here](https://github.com/Kong/kong/blob/master/kong/init.lua#L110) and the next branch is [here](https://github.com/Kong/kong/blob/next/kong/db/dao/plugins.lua#L43). I see that it has undergone change with the plugins being moved to the new DAO.

Since both, this PR and the check plugins on startup logic need a set of plugin names that are configured in the DB, we could optimize this and load the plugins names from DB once (rather than in both places as happening currently), cache them and then use them to configure this PR and also verify if Kong should exit.

This is just a thought on how we can optimize the code and not an ask of any sorts.

Thoughts?

thibaultcha: Ugh, yes indeed, that is correct! I am not sure if it is a good thing or a bad thing, tbh, but at the very least we could take advantage of it, indeed!

p0pr0ck5: The iterator invoked in the `next` branch for the plugins load uses the same iterator in question here. A large part of the design of this PR was to use existing facilities in preparation for a stable release- if these facilities are not sufficient, we can address that. 

I would be much more concerned about optimizing the runtime query times as opposed to startup- the data access model in Kong already means that slow start concerns are the responsibility of infrastructure, and not software, architects (and I would agree). That given, the complexity and confusion to optimize the load-time query pattern is not worth it.

Here’s my semi final thought in this: if we are deeply concerned about performance in this context, we should build strategy-specific solutions, both of which had been alluded to here. Otherwise, we should acknowledge the utility and value of Kong’s underlying data access patterns, and leave it at that(and any potential optimization’s to that central code can come outside the scope of this pr).

kong/init.lua
p0pr0ck5: adjusted

p0pr0ck5:
@thibaultcha ive pushed up a separate commit addressing your comments for easy of review.

kong/init.lua
```diff
@@ -504,6 +500,12 @@ function Kong.ssl_certificate()
 
   runloop.certificate.before(ctx)
 
+  local ok, err = plugins_map_wrapper()
+  if not ok then
+    ngx_log(ngx_CRIT, "could not ensure plugins map is up to date: ", err)
+    return responses.send_HTTP_INTERNAL_SERVER_ERROR()
```
thibaultcha: This will be trouble in this context (`certificate`). Mind following the same pattern as we do for this phase? See: https://github.com/Kong/kong/blob/0.14.1/kong/runloop/certificate.lua#L106-L107

kong/init.lua
```diff
@@ -152,37 +150,35 @@ local function plugins_map_wrapper()
 
     local ok, err = plugins_map_semaphore:wait(timeout)
 
-    if ok then
-      -- we have the lock but we might not have needed it. check the
-      -- version again and rebuild if necessary
+    if not ok then
+      return nil, err
```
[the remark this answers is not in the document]
p0pr0ck5: Ah gotcha. I wasn't sure about the double colon design; felt odd to me but i see the understanding (now i know why it lives in other places in the codebase). And i missed the initial part of that comment, my bad.

Fixed!

kong/init.lua
```diff
@@ -152,37 +150,35 @@ local function plugins_map_wrapper()
...
 
-      plugins_map_semaphore:post(1)
-    else
-      ngx_log(ngx_CRIT, "could not acquire plugins map update mutex: ", err)
+      local ok, err = build_plugins_map(kong.db, version)
+      if not ok then
+        plugins_map_semaphore:post(1)
 
-      return false
+        return false, err
```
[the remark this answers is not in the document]
p0pr0ck5: missed this one, thanks :)

kong/init.lua
p0pr0ck5: whups! thanks!

thibaultcha:
Merged! :tada:

thibaultcha:
The resulting commit on `next` is 9ea6d0b89cc14900af9973269c49edb4ca209bab

Issue #91, opened 2015-03-25
Refactor Applications


thefosk:
Applications should support the following features:

* Whitelisted APIs
* Blacklisted APIs
* #87 

And should be more intuitive to use. @montanaflynn brought up the following issue: it is possible to create applications even if no authentication is being enabled, and the application properties depend on the authentication type that has been configured. 

For the last point maybe Application's should be authentication-related by supporting a list of `authentication_types`, as discussed at #18. More or less `authentication_types` can be added to an Application depending what authentication schemes the API has installed through the relative Plugins.

thibaultCha:
Since yesterday we decided to ditch out the concept of applications in favour of `consumers` + still have multiple keys per authentication type, this issue kind of becomes obsolete. Update with yesterday's discussion in #18. #87 is still relevant.

PR #3803, opened 2018-09-24
Override config settings from kubernetes mounted file secrets

jdevalk2:
### Summary

This is related to bug #91 in the docker_kong repo.

https://github.com/Kong/docker-kong/issues/91

Though we had a need to read the secrets from Kubernetes mounted secret files, I have not tested with barebones Docker images but it should be very similar.

In our Kubernetes setup we did not want to pass the database password and other sensitive details through environment variables. Limitations of environment variables in general are that they can be  dumped in a log in case of an error and this would cause the database password to be visible.

This current version is pretty bare bones as it does what we need it to. The secrets are mounted under /etc/kong/secrets if they exist and conf loader will look for them there.

### Full changelog

I have added a couple of unit tests and some methods to inject the secrets path into the config loader.

### Issues resolved

Allows the reading / override of config values from kubernetes file secrets.

You would use it something like this in K8S yaml:
This loads postgres database PG_USER, PG_HOST, PG_PASSWORD etc.

```
   spec:
      volumes:
      - name: secrets-volume
        secret:
          secretName: kongapigateway-db-secret    
          items:
          - key: dbUser
            path: pg_user  
          - key: dbPassword
            path: pg_password
          - key: dbHost
            path: pg_host
          - key: dbPort
            path: pg_port
          - key: dbDatabase
            path: pg_database          
      containers:
      - name: kong-proxy
        volumeMounts:
        - name: secrets-volume
          mountPath: /etc/kong/secrets
          readOnly: true
        image: "kong:0.14.1-centos"
        imagePullPolicy: Always
```


kong/conf_loader.lua
```diff
@@ -23,6 +23,7 @@ local DEFAULT_PATHS = {
   "/etc/kong.conf",
 }
 
+local secret_path = "/etc/kong/secrets"
```
p0pr0ck5: would it make sense to make this value configurable?

jdevalk2: It would, though looking at that I found it resulted in a "chicken and egg" problem. 

Since the override with file secrets is done inside the confloader object, a configurable file secret path would itself be part of the configuration file.  So while loading the configuration the configuration is referenced.  

I couldn't find any instances of the configuration referencing itself yet, so I figured a constant file path would be a good first step.

Any suggestions on this would be welcome.

pamiel: Hi @jdevalk2 

Sorry to comment this Pull Request as a pure potential user of this feature, as I’m not an official reviewer!

I was about to comment that having a hard coded path for secret files is not that a good idea as security constraints on deployment, or rules for hardening Docker images (in case Kong is deployed on Docker) might mandate such configuration information to be located in specific directories other than `/etc/kong/secrets/`.

However, digging into your code, I have a second comment (first one remains applicable :P ): your proposed implementation looks to me not homogeneous with the configuration management already in place in Kong. While configuration is currently based on a file containing `tag = value` pairs, your proposal is to get sensitive data in a different manner: value of _each_ configuration parameter would be stored in an _individual_ file named `/etc/kong/secrets/<parameter value>`. 10 sensitive parameters => 10 files (as far as I understood; sorry if my understanding is incorrect).

The feature of providing (sensitive) configuration data as Kubernetes secrets definitely makes sense, but why not addressing it simply by the capability to get several configuration files as input to Kong (a kind of `-c <file1> -c <file2>`)? That would be a deployment decision to put some configuration data in clear text in `<file1>` (and in directory A), and the sensitive configuration data in a "protected" file `<file2>`, made available for example as a K8S secret file mounted as a volume (in directory B).
Priority would be linked to the precedence of the file within the command line (`<file1>` is in first position in the command line, so with higher priority in case data are part of both files… or the reverse ;) ). Usage of envvars would anyway be with the highest priority (a difference with your code).

This looks more in line with current configuration mechanism, and less linked to a pure Kubernetes use case, by the way.

jdevalk2: I think one distinction to make in the discussion is between Configuration and Secrets.
Secrets are a type of Configuration. But a different class of information.

Secrets like username / password have a different lifecycle than normal configuration.
For example I can put my username + password on rotation every week, or I can revoke a password.

Further as stated I am a personal believer that secrets should not be passed into environment variables.
It will not "break the bank" but I am a believer of Defense in Depth strategies and for purpose of environment dumps and that its so easy for them to end up in log files, they should not be in there.


To best of my knowledge (as I am not a Kubernetes guru)
In scope of Kubernetes, configmaps are used for configuration and non-secure data.
Kubernetes secrets are used for secure bits of data. Secrets are stored as a different class of data in Kubernetes (and also Docker)
And third party products exist to manage secrets and plug into kubernetes like Hashicorp Vault.


If we provide a configfile to Kong on the command line
Kong.exe -c baseconf.conf -c secretconf.conf


The control of which config overrides another would be part of the command line. 
The secretconf.conf would have to come from somewhere. 
Two ways could be it would have to be mounted as a file by a configmap, or it would have to be created by an init container.

As the configmap is not really meant for secret data, I feel the better would be init container.

In which case one could mount the secrets in the init container and have it produce that file.

I think I just argued against my own pull request!
Anyway from a docker / kubernetes perspective it would be nice to have an alternate other than having secrets passed into ENV variables.   I will explore the init container route a bit further as well.

pamiel:
Hi @jdevalk2 ,

I think we are not so far !
I agree with you on the following points:
- Life cycle of sensitive information is not (may not be) the same as other configuration data
- No sensitive data should be put in envars
- In the context of Kubernetes, ConfigMap is to be used for plain text configuration, and Secret is suitable for sensitive configuration data (we may discuss on how “secure” Kubernetes Secrets are… but that’s another story!)
- There are other technical solutions than K8S Secrets to handle sensitive data storage (and Vault is quite an interesting one).

However there is one point I’m not in line with you: you said `Two ways could be [secrets] would have to be mounted as a file by a configmap, or it would have to be created by an init container.` Indeed no: Secrets could be mounted… as a Secret file/volume! => neither a plain ConfigMap nor with the absolute need of an Init Container !
Let me explain:

**A Secret is... a Secret, not a ConfigMap**
Maybe you’ve been confused by the example provided by Kubernetes.io documentation https://kubernetes.io/docs/concepts/configuration/secret/ where they put one sensitive data per file; but you can also have a single file with a set of `<tag> = <value>` lines, and create a Secret (`kubectl create secret generic secretConfig.conf --from-file=./<the single file>`) containing your bunch of sensitive data, as a single file.
And then:
-	Making the K8S Secret be available to a Container is very similar to making a ConfigMap be available to this Container… but this is a Secret, not a ConfigMap!
-	Application within the Container (Kong in our use case) will read the file as it was a normal file, and look for `<tag> = <value>` content in it

That’s why I think it is easy to keep the same pattern of configuration file (with a set of `<tag> = <value>` lines) for plain configuration data and for sensitive data (my initial comment was mainly addressing this point: in your PR, the sensitive data are presented in a different way (1 parameter = 1 file, parameter name being the file name) while it is possible to be in a similar format than the `konf.conf` file).

**Usage of an Init Container**
Usage of an Init Container might be useful if the data are not stored directly in a Kubernetes Secret, but for example in a Vault as you mentioned. However, Kong would most likely not want to create an explicit dependency on Vault by using the Vault API; so patters well documented on Internet present a solution where an Init Container would first get the data from Vault (so only this Init Container would have the dependency on the Vault API), and then create a file containing the sensitive data, so that the business Container could finally read them.
Why not… but in any case, Kong would have the need to read several configuration files (the `kong.conf`, plus the one created by the Init Container), and I think it would be preferable that these 2 files have the same format.

**A last word about envars?**
I was talking about envars taking the precedence to files just because this is the way it is done today when Kong reads its configuration: we might not touch this behavior at all and just recommend to not put password or any other sensitive data as envars (or let people that don’t care do that !)

jdevalk2:
Hi @pamiel,

I concur that Kubernetes secrets can also be mounted as a file with multiple key value pairs, rather than just 1 secret per file.

For the purpose of getting the secrets config inside Kong, at this moment I now prefer init container.
This is based on the work I have done over the last couple of days and diving further into how Kong loads config.  
  
  
One secret per file or many per file

In the end there must be a piece of logic that maps 1, 2, to N secrets into a Kong config file.

So whatever Kubernetes approach is chosen, 1 per file, or multiple imported as a file, the piece of logic that creates that file (merges) with secrets must live somewhere.

Then once that file is created, it has to be passed onto Kong.


I like the idea of being able to give Kong multiple config files, however outside the scope of this particular use case I do not know of other use cases where this might be used.

Looking at Kong configuration code, it already merges a base configuration with the provided configuration.
So the merging of configurations is something it does already.

However if this is refactored to take multiple configurations I have the feeling it complicates that config loading code further and it might result in some undesired side effects if config files are merged on top of eachother multiple times.
At least it would result in a considerable expansion of the unit tests to test all permutations.

If we have the piece of logic that creates the secret config file, it's not too far to also create a single entire Kong config file.
In our case we now have a Kong template and we merge the secrets through shell script.

Our shell script then produces 1 kong config file which we then inject by writing it to the default config path /etc/kong where it gets picked up by kong on startup.

So the need to give Kong multiple config files to merge is not there anymore.


For our use case init container makes a lot of sense now.
In the init container the Kubernetes secrets are mounted and then injected into a kong configuration template.


A further benefit of this approach came when we had a use for a Kong custom plugin, where the init container can also be used to setup the plugin in the kong container.

At this point I am thinking about closing this pull request as there is a viable workaround I found. 
If the multiple configuration file loading option is explored then probably this should be it's own pull request, and not branch of this one?

jdevalk2:
PS: One thing I thought I'd point out

What the init container does not address is when a database password is changed, that it would then pick up the credential change and reconfigure Kong automatically with "Kong restart". It would have to be a side car container then in Kubernetes that does this.  

Or maybe this should be part of Kong Ingress Controller project. 

That would be a manual restart with the init container approach.

PR #3806, opened 2018-09-26
test(tcp-log-plugin) fix flaky test

kidd:
In some cases, there's 1 split ms that makes the test fail.
Allowing for 1 extra ms.

### DIFF
--- a/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
+++ b/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
@@ -100,7 +100,15 @@ for _, strategy in helpers.each_strategy() do
       local log_message = cjson.decode(res)
 
       assert.True(log_message.latencies.proxy < 3000)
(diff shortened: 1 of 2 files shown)


spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
```diff
@@ -100,7 +100,12 @@ for _, strategy in helpers.each_strategy() do
       local log_message = cjson.decode(res)
 
       assert.True(log_message.latencies.proxy < 3000)
-      assert.True(log_message.latencies.request >= log_message.latencies.kong + log_message.latencies.proxy)
+
+      local is_latencies_sum_adding_up =
+        1+log_message.latencies.request >= log_message.latencies.kong +
```
thibaultcha: @kidd Mind adding a comment that explains why the `+ 1` is needed here? Otherwise this looks quite obscure to the reader. Thanks!

kidd: sure, I thought the git commit msg was enough but I guess the clearer the better.

PR #3808, opened 2018-09-27
 feat(db) add pg_timeout configuration parameter to postgres strategy

bungle:
### Summary

Adds support for `pg_timeout` configuration parameter. Also fixes `major_minor_version` of Postgres according to Postgres versioning policy: https://www.postgresql.org/support/versioning/.

### DIFF
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -138,6 +138,10 @@ local function connect(config)
   connection.convert_null = true
   connection.NULL         = null
 
+  if config.timeout then
+    connection:settimeout(config.timeout)
+  end
+
   local ok, err = connection:connect()
   if not ok then
     return nil, err
@@ -230,6 +234,7 @@ function _mt:init()
   if major < 10 then
     self.major_version       = tonumber(fmt("%u.%u", major, floor(ver / 100 % 100)))
     self.major_minor_version = fmt("%u.%u.%u", major, floor(ver / 100 % 100), ver % 100)
+
   else
     self.major_version       = major
     self.major_minor_version = fmt("%u.%u", major, ver % 100)
@@ -761,6 +766,7 @@ function _M.new(kong_config)
   local config = {
     host       = kong_config.pg_host,
     port       = kong_config.pg_port,
+    timeout    = kong_config.pg_timeout,
     user       = kong_config.pg_user,
     password   = kong_config.pg_password,
     database   = kong_config.pg_database,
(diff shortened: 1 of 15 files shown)


kong/db/strategies/postgres/connector.lua
```diff
@@ -215,25 +222,18 @@ local _mt = {}
...
-    return nil, "failed to retrieve server_version"
-  end
-
-  self.major_minor_version = extract_major_minor(res[1].server_version)
-  if not self.major_minor_version then
-    return nil, "failed to extract major.minor version from '" ..
-                res[1].server_version .. "'"
+  local major = floor(ver / 10000)
+  if major < 10 then
+    self.major_minor_version = fmt("%u.%u.%u", major, floor(ver / 100 % 100), ver % 100)
```
thibaultcha: `major_minor` is supposed to only contain `X.Y` from the `X.Y.Z` version string, on purpose. It looks like this would break this expectation? The old DAO has tests for this for both C* and Postgres (that were not yet ported to the new DAO): https://github.com/Kong/kong/blob/next/spec/01-unit/01-db/03-postgres_spec.lua

bungle: So question is, should we follow meaning of Major/Minor of PG, or not, and follow semver style, even when it doesn't match the real world.

bungle: And I can port the tests to new db too.

bungle: I have pushed changes to this branch.

kong/db/strategies/postgres/connector.lua
```diff
@@ -215,25 +223,20 @@ local _mt = {}
...
-    return nil, "failed to retrieve server_version"
-  end
-
-  self.major_minor_version = extract_major_minor(res[1].server_version)
-  if not self.major_minor_version then
-    return nil, "failed to extract major.minor version from '" ..
-                res[1].server_version .. "'"
+  local major = floor(ver / 10000)
+  if major < 10 then
+    self.major_version       = tonumber(fmt("%u.%u", major, floor(ver / 100 % 100)))
```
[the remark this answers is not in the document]
bungle: I followed Cassandra strategy here, where `self.major_version` is `number` and `self.major_minor_version` is `string`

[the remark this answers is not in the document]
thibaultcha: I recalled wrong then :) Yes, the C* connector needs to know if the major version is 2 or 3 in order to make some queries slightly differently... Well, let's say both variables are due for a rename then!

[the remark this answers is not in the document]
thibaultcha: (Outside of the scope of this PR, of course)

thibaultcha:
Merged!

PR #3815, opened 2018-09-29
feat(healthchecks) add support for HTTPS in active health checks

hishamhm:
* Introduces three new fields to `active` healthchecks configuration:
  * `type` - "tcp", "http" or "https", mirrorring lua-resty-healthcheck 0.6.0, bumped dependency
  * `https_verify_certificate` - boolean, mirrorring lua-resty-healthcheck 0.6.0
  * `https_sni` - explicitly give an SNI, overriding the hostname configured in the Target
* Adds tests using HTTPS, based on lua-http (new dependency)
  * lua-http is used for HTTPS tests, the mock server based on LuaSocket remains used for HTTP tests for performance reasons (faster startup on separate threads which are repeatedly spawned for various short-lived tests)
  * This PR also updates the certs in spec/fixtures, which were expired


kong/db/schema/entities/upstreams.lua
```diff
@@ -55,11 +55,30 @@ local header_name = Schema.define {
...
+local NO_DEFAULT = {}
+
+
+local healthchecks_config = {
   active = {
+    type = "http",
     timeout = 1,
     concurrency = 10,
     http_path = "/",
+    https_sni = NO_DEFAULT,
```
thibaultcha: `https_sni` expects a `host` typedef (which in turn, if of type `string`), but the default is a table? Doesn't this feel strange?

thibaultcha: It seems like the `host` typedef allows for IP addresses (https://github.com/Kong/kong/blob/master/kong/db/schema/typedefs.lua#L10-L19), while as per [RFC 6066](https://tools.ietf.org/html/rfc6066#section-3), an SNI can only contain DNS hostnames:

>  Currently, the only server names supported are DNS hostnames; [...] Literal IPv4 and IPv6 addresses are not permitted in "HostName".

This probably warrants the creation of a new `sni` typedef.

kong/db/schema/entities/upstreams.lua
```diff
@@ -100,26 +121,37 @@ local types = {
...
+
 local function gen_fields(tbl)
   local fields = {}
   local count = 0
   for name, default in pairs(tbl) do
     local typ = types[name]
-    local def
+    local def, required
+    if default == NO_DEFAULT then
+      default = nil
```
thibaultcha: This answers the question in my first comment; ignore it :)

kong/db/schema/typedefs.lua
```diff
@@ -57,6 +57,24 @@ local function validate_name(name)
...
+  if err_or_port == "invalid port number" or type(res.port) == "number" then
+    return nil, "must not have a port"
+  end
+
+  if res.type ~= "name" then
+    return nil, "must not be an IP"
+  end
+
+  return true
+end
```
thibaultcha: If I were to specify `ip:port`, I would first receive an error stating that I should not specify a port, and after retrying, I would get an error that I should also not specify an IP. I think the `"most not be an IP address"` error should be thrown first.

Also, are we lacking tests for this function?

thibaultcha:
@hishamhm Ping on the failing CI suite - it looks related.

thibaultcha:
@hishamhm Ping on the failing CI suite - it looks related.

kong/db/schema/typedefs.lua
hishamhm: > Also, are we lacking tests for this function?

It was being tested indirectly via the upstreams tests (as is the case for some other typedefs).

hishamhm:
Re-pushed. Apparently `snis` needs to accept non-hostname `name` entries otherwise Kong's own certs are not stored.

PR #3819, opened 2018-10-01
fix(migrations) postgres migrations tuning and json->jsonb, timestamp->timestamptz

bungle:
### Summary

- ~This PR changes base migration of `plugins` table `config` column type from `JSON` to `JSONB` (it was already converted on 0.14.0 to 0.15.0 migration)~
- ~It also changes `upstreams` table `healtchecks` column type from `JSON` to `JSONB` and adds `0.14.0` to `0.15.0` migration for the column type.~
- This PR changes `JSON` types to `JSONB` types on `0.14.0` to `0.15.0` migrations
- This PR changes `TIMESTAMP WITHOUT TIMEZONE` types to `TIMESTAMP WITH TIME ZONE` types on `0.14.0` to `0.15.0` migrations
- This PR cleans cruft left behind from **pre** `0.15.0` migrations to match the new database created with `kong migrations bootstrap && kong migrations up && kong migrations finish` on `0.15.0`.


kong/db/migrations/core/000_base.lua
```diff
@@ -152,7 +152,7 @@ return {
         "hash_on_cookie"        TEXT,
         "hash_on_cookie_path"   TEXT,
         "slots"                 INTEGER                      NOT NULL,
-        "healthchecks"          JSON
+        "healthchecks"          JSONB
```
thibaultcha: This migration should truly reflect the state of the database as of 0.14, and we should not diverge away from it. The `14_to_15` migration will still run for freshly bootstrapped databases as well, so there is no need for this change anyway?

kong/db/migrations/core/001_14_to_15.lua
```diff
@@ -33,6 +33,10 @@ return {
       $$;
 
       ALTER TABLE IF EXISTS ONLY "plugins" ALTER "config" TYPE JSONB USING "config"::JSONB;
+
+
+
+      ALTER TABLE IF EXISTS ONLY "upstreams" ALTER "healthchecks" TYPE JSONB USING "healthchecks"::JSONB;
```
thibaultcha: Out of curiosity: do you know what happens _if_ (big if) the type of this column is already `JSONB`?

bungle: I am not sure what happens internally, but I guess pretty much nothing happens. I tried it and, well nothing happened.

kong/db/migrations/core/000_base.lua
bungle: Thanks for for this. I fully 100% agree.

kong/plugins/rate-limiting/migrations/000_base_rate_limiting.lua
```diff
@@ -108,7 +108,7 @@ return {
         period      text,
         identifier  text,
         value       counter,
-        PRIMARY KEY ((route_id, service_id, api_id, identifier, period_date, period))
```
thibaultcha: Why removing those? They serve the purpose of specifying a compound partition key, and are essential. I recall similar attempt to remove them already in the past - unfortunately we cannot add comments in this migration...

kong/plugins/rate-limiting/migrations/001_14_to_15.lua
```diff
@@ -0,0 +1,16 @@
...
+        ALTER "period_date" TYPE TIMESTAMP WITH TIME ZONE USING "period_date" AT TIME ZONE 'UTC';
+    ]],
+  },
+
+  cassandra = {
+    up = [[
+      BEGIN BATCH
+        -- nothing to migrate
+      APPLY BATCH;
+    ]],
```
thibaultcha: It should be possible (and preferable) to specify an empty string for these NOP migrations; have you tried?

kong/plugins/rate-limiting/migrations/000_base_rate_limiting.lua
bungle: Oh, it really looked like a tupo as it wasn't found anywhere else. I even tried to find this on C* docs, but couldn't. I should have tried harder. Thanks for educating me on this. I will fix it. Siirry.

kong/plugins/rate-limiting/migrations/001_14_to_15.lua
bungle: I tried, but it gave an error. We can fix why it errored, or keep this.

thibaultcha: Could we fix the error then? Specifying the migration is required for the sake of explicitness as per [the migrations schema](https://github.com/Kong/kong/blob/next/kong/db/schema/others/migrations.lua#L11-L12), but the connectors should just NOP on empty migrations

bungle: @thibaultcha I added `len_min = 0` to migrations schema.

PR #3820, opened 2018-10-01
feat(core) use 'pg_timeout' as router rebuild semaphore timeout

thibaultcha:
Complementary commit for:

* 11d69be feat(conf) add 'pg_timeout' property
* 85946ba perf(router) wrap router rebuilds in a worker mutex

We now use the configured timeout for PostgreSQL as the timeout to our
router rebuild semaphore. Note that the default value of `60` is
preserved to be friendly to any fork out there that may support other
configuration backends.

See #3794
See #3808

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -466,15 +466,14 @@ return {
         -- semaphore resource is acquired we re-check the api_router version
         -- again to prevent unnecessary subsequent rebuilds
 
-        -- for postgres-backed installations, wait up to 60 seconds (the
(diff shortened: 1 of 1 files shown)


p0pr0ck5:
Does keeping a default assignment of `60` make sense, given that we automatically assign a 5s timeout for psql: https://github.com/Kong/kong/commit/11d69be02a996b66ccd79bd0c2969dbba2b1adc6#diff-82c78a715b27a7bade741c59e02437a9R42 ?

p0pr0ck5:
> may support other configuration backends.

My bad, misread this :) nvm

PR #3824, opened 2018-10-03
feat(migrations) 0.14.0 to 0.15.0 migrations with added checks

bungle:
### Summary

**Original description:**

When running 'kong migrations up' from a 0.15 install (this ignores plugins):

- If we detect an existing Kong install (0.14 only, detected via previously ran migrations), then, no need for 'bootstrap step'. In an 'if' statement somewhere, follow a special migration path as such:
- Invoke DAO's 'bootstrap' step. Now, the schema_meta and locks tables are available. We assume that the rest of the DB is of the same state as if we ran the `init` migration of each subsystem.
- In a cluster_mutex (so that multiple nodes can be running this 0.14 to 0.15 migration), run the 014_to_015 migration of each subsystem (but still record the init migration as having ran). Now, both the init and the 014_to_015 migrations are recorded as pending/executed in `schema_meta` (but only the latter really ran).

At this point, we have:
- The `schema_meta` table with `init` and `014_to_015` migrations pending/executed (allowing the user to start 0.15 nodes)
- The `schema_migrations` table with all 0.14 migrations (allowing the user to start 0.14 nodes while still migrating)
- A DB schema similar to 0.14, but with some 0.15 improvements (from the `014_to_015` migrations)

Eventually, the user runs `kong migrations teardown` (just like the new framework). The same detection is done again on the `schema_migrations` table, and we detect that we are in a 0.14 to 0.15 migration path. Instead of running the above steps again, this time, we *drop* the `schema_migrations` table, and then run all `teardown` phases of the `14_to_15` migrations (if any). Once this operation is completed, the schema is only valid for 0.15 and above.

This implements the new 2-steps migrations framework (up + teardown), but for a 0.14 to 0.15 migration.

TODO: there needs to be special handling for custom plugins with migrations.

**About the implementation:**

While it doesn't strictly work exactly as described above, e.g. it runs `base` migration always as it is harmless, but that's the way we can get exactly the schema as what we get from fresh bootstrap (as the `base` migration adds some `indexes` and updates some triggers and functions on Postgres, not worth to move them I think), and this together with #3819 will make schemas the same (except column order in table might be different, but it doesn't matter, and there is hardly anything we can do about it).

### DIFF
--- a/kong/db/init.lua
+++ b/kong/db/init.lua
@@ -739,20 +739,60 @@ do
                               self.strategy, mig.name)
         end
 
+        if t.subsystem == "core" and mig.name == "000_base" then
+          local res, err = self.connector:is_014()
+          if err then
+            return nil, fmt_err(self, "unable to detect database version (%s)",
+                                err)
+          end
+
+
+
+          if not res.is_eq_014 and not res.is_gt_014 then
+            if res.missing_migration then
+              return nil, fmt_err(self,
+                                  "Migration to 1.0 can only be performed "  ..
+                                  "from a 0.14.x %s %s, but the current "    ..
+                                  "one seems to be older (missing "          ..
+                                  "migration '%s' for '%s'). Migrate to "    ..
+                                  "0.14.x first, or install 1.0 on "         ..
+                                  "a fresh %s.",
+                                  self.strategy, self.infos.db_desc,
+                                  res.missing_migration, res.missing_component,
+                                  self.infos.db_desc)
+
+            elseif res.missing_component then
+              return nil, fmt_err(self,
+                                  "Migration to 1.0 can only be performed "  ..
+                                  "from a 0.14.x %s %s, but the current "    ..
+                                  "one seems to be older (missing "          ..
+                                  "migrations for '%s'). Migrate to 0.14.x " ..
+                                  "first, or install 1.0 on a fresh %s.",
+                                  self.strategy, self.infos.db_desc,
+                                  res.missing_component, self.infos.db_desc)
+            end
+
+            return nil, fmt_err(self,
+                                "Migration to 1.0 can only be performed "    ..
+                                "from a 0.14.x %s %s, but the current one "  ..
+                                "seems to be older (missing migrations)."    ..
+                                "Migrate to 0.14.x first, or install 1.0 "   ..
+                                "on a fresh %s.", self.strategy,
+                                self.infos.db_desc, self.infos.db_desc)
+          end
+        end
+
         log.debug("running migration: %s", mig.name)
 
         if run_up then
           -- kong migrations bootstrap
           -- kong migrations up
-
-          local ok, err = self.connector:run_up_migration(mig.name,
-                                                          strategy_migration.up)
+          ok, err = self.connector:run_up_migration(mig.name, strategy_migration.up)
           if not ok then
             self.connector:close()
             return nil, fmt_err(self, "failed to run migration '%s' up: %s",
                                 mig.name, err)
           end
-
 
           local state = "executed"
           if strategy_migration.teardown then
@@ -761,8 +801,7 @@ do
             n_pending = n_pending + 1
           end
 
-          local ok, err = self.connector:record_migration(t.subsystem,
-                                                          mig.name, state)
+          ok, err = self.connector:record_migration(t.subsystem, mig.name, state)
           if not ok then
             self.connector:close()
             return nil, fmt_err(self, "failed to record migration '%s': %s",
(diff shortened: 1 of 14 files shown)


bungle:
How to test:

### make fresh install

1. `git checkout checkout feat/migration_14_to_15`
2. `./bin/kong/migrations reset`
3. `./bin/kong/migrations bootstrap`
4. `./bin/kong/migrations up` (not needed as bootstrap already does this for fresh install)
5. You can no try to start Kong (actually you can try that in any step, to see if it works or not, and why)
6. `./bin/kong/migrations finish` (not needed as bootstrap already does this for fresh install)

### upgrade from 0.14.1

1. `git checkout checkout 0.14.1`
2. `./bin/kong/migrations reset`
3. `./bin/kong/migrations up`
4. `git checkout checkout feat/migration_14_to_15`
5. `./bin/kong/migrations bootstrap`
6. `./bin/kong/migrations up`
7. At this point `schema_migrations`table is still around, and `0.14.0` nodes should still work.
8. Start kong
6. `./bin/kong/migrations finish` (now the `schema_migrations` table is gone)


### upgrade from 0.14.1

1. `git checkout checkout 0.13.1`
2. `./bin/kong/migrations reset`
3. `./bin/kong/migrations up`
4. `git checkout checkout feat/migration_14_to_15`
5. `./bin/kong/migrations bootstrap`
6. `./bin/kong/migrations up` (doesn't work)
7. `git checkout checkout 0.14.1`
8. `./bin/kong/migrations up`
9. `git checkout checkout feat/migration_14_to_15`
10. `./bin/kong/migrations up` (does work)
11. At this point `schema_migrations`table is still around, and `0.14.0` nodes should still work.
12. Start kong
13. `./bin/kong/migrations finish` (now the `schema_migrations` table is gone)

kong/db/init.lua
```diff
@@ -739,30 +740,50 @@ do
...
+            return nil, fmt_err(self, "unable to detect database version (%s) ", err_0_14)
+          end
+
+          if not is_0_14 and not is_gt_0_14 then
+            if missing_0_14_migration then
+              return nil, fmt_err(self, "migration '%s' can only be executed on " ..
+                                  "a %s database that is migrated to version 0.14.x, " ..
+                                  "but database is missing '%s' migration for '%s'",
+                                  mig.name, self.strategy, missing_0_14_migration,
+                                  missing_0_14_id)
```
thibaultcha: It seems like in this code path, `missing_0_14_id` may sometimes be `nil`. See https://github.com/Kong/kong/pull/3824/files#diff-aa25782656ddfb572924702f3d4e838cR990, in which we complain that an entire subsystem's migrations have not been executed (in prior framework). Note that we return 4 arguments, not 5. I believe using a `res` table to hold that information will prevent the confusing API with the multiple return values.

I am not sure if the error message is properly written either, as it seems to be expecting a migration's name, but in fact, _all_ migrations of this subsystem are missing, afaict. Thus, the proper error message seems to be:
```
... but [database] is missing migrations for '%s' component.
```

kong/db/init.lua
```diff
@@ -739,30 +740,50 @@ do
...
+            else
+              return nil, fmt_err(self, "migration '%s' can only be executed on " ..
+                                  "a %s database that is migrated to version 0.14.x, " ..
+                                  "but database is missing '%s'", mig.name, self.strategy,
+                                  missing_0_14_id)
```
thibaultcha: I feel like both of these error messages can be improved. They should start by explaining what they are trying to do (migrate to 1.0), and should have some sort of closure, to explain to the user which conclusion our heuristics took:
```
Migration to 1.0 can only be executed from a fresh or 0.14.x %s [database/keyspace], but the %s [database/keyspace] is missing some migrations (%s), indicating an older version. Migrate to 0.14.x first.
```

Where the first `%s` is "Postgres/Cassandra", the second one is the name of the database/keyspace itself, and the third one is the name of the migration detected as missing.

The message may be refined, of course. Ultimately, I don't even think we need to separate both possible error cases, which would result in a single, identical error message for both "missing component" and "missing migration(s) in component". Unless we do for another reason>

kong/db/init.lua
```diff
@@ -739,30 +739,67 @@ do
...
+            if res.missing_migration then
+              return nil, fmt_err(self,
+                                  "Migration to 1.0 can only be executed " ..
+                                  "from a fresh or 0.14.x %s %s, but the " ..
+                                  "%s %s is missing some migrations (%s) " ..
+                                  "for %s component, indicating an older " ..
+                                  "version. Migrate to 0.14.x first.",
+                                  self.strategy, self.infos.db_desc,
+                                  self.strategy, self.infos.db_desc,
+                                  res.missing_migration, res.missing_component)
```
thibaultcha: The messages are still somewhat verbose. Verbose error messages tend to be ignored by users... May I suggest avoiding some repetition (in all 3 of those messages):
```
Migration to 1.0 can only be performed from a 0.14.x postgres database, but
the current one seems to be older (missing migration '%s' for '%s'). Migrate to
0.14.x first, or install 1.0 on a fresh database.
```

thibaultcha:
Manually merged to next!

PR #3825, opened 2018-10-03
Merge branch 'master' into next

hishamhm:
Not sure if this will merge the way we want, or if it will make a double `merge` commit

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,26 +110,11 @@ patch](#submitting-a-patch) section.
 
 The documentation hosted at https://docs.konghq.com is open source and built
 with [Jekyll](https://jekyllrb.com/). You are welcome to propose changes to it
-(correct typos, add examples or clarifications...)!
+(correct typos, add examples or clarifications...) and contribute to the
+[Kong Hub](https://docs.konghq.com/hub/)!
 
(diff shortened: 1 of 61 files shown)


thibaultcha:
It seems like the CI is failing due to some missing updates to the AWS Lambda tests about the new migrations in next.

hishamhm:
@thibaultcha it's green now. Do we attempt merging this?

thibaultcha:
@hishamhm Absolutely! I believe that a "rebase and merge" would result in the equivalent of a direct push from a local merge (with only the "Merge branch 'master' into next" commit showing up in the next branch).

hishamhm:
What, _now_ it started complaining due to conflicts :confounded:

hishamhm:
and _without_ a list of conflicting files!

hishamhm:
@thibaultcha oh, "Rebase and merge" is blocked, but the "Create a merge commit" option is available. Do we shrug and go with that?

hishamhm:
If Github's command line instructions are to be believed as the equivalent of what the green button would do, the "Merge pull request" option will indeed create a double merge commit entry (because of its `--no-ff` option).

hishamhm:
Merging manually without `--no-ff` worked!

PR #3826, opened 2018-10-03
feat(pdk) add kong.node module with '.get_id()' util

thibaultcha:
This new module provides node-level utilities, such as
`kong.node.get_id()`, which is a replacement to the old
`kong.tools.public` module's `get_node_id()` utility.

Reminder: the `kong.tools.public` module is to be stripped away for the
1.0 stable release.

This new implementation **throws** errors instead of returning them, for
two reasons:

1. Simplicity of usage: now that this function is exposed for public
consumption (and already used by plugins such as Zipkin), avoiding for
the need to check for errors simplifies its usage.
2. If such an error occurs, it will be during the initialization of the
master process, and will then properly prevent Kong from starting. At
runtime, no such error should occur since the id is cached, but if they
do, they should be very obvious and provide their stacktraces.

The `kong.node` module could also contain the following utilities in the
future:

* get_hostname()
* get_system_infos()
* ...

### DIFF
--- a/kong/pdk/node.lua
+++ b/kong/pdk/node.lua
@@ -1,0 +1,56 @@
+--- Node-level utilities
+--
+-- @module kong.node
+
+local utils = require "kong.tools.utils"
+
+
+local NODE_ID_KEY = "kong:node_id"
+
+
+local node_id
+
+
+local function new(self)
+  local _NODE = {}
+
+
+  ---
+  -- Returns the id used by this node to describe itself.
+  --
+  -- @function kong.node.get_id
+  -- @treturn string The v4 UUID used by this node as its id
+  -- @usage
+  -- local id, err = kong.node.get_id()
+  function _NODE.get_id()
+    if node_id then
+      return node_id
+    end
+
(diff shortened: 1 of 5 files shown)


kong/pdk/node.lua
```diff
@@ -0,0 +1,56 @@
...
+  local _NODE = {}
+
+
+  ---
+  -- Returns the id used by this node to describe itself.
+  --
+  -- @function kong.node.get_id
+  -- @treturn string The v4 UUID used by this node as its id
+  -- @usage
+  -- local id, err = kong.node.get_id()
```
hishamhm: The example says `get_id` may return a second `err` argument, but that's not the case. I suggest changing to `local node_id = assert(kong.node.get_id())` (no hard feelings either way about the assert, just an idea to convey "should always work")

thibaultcha: Ugh, a leftover from prior to this change! Thanks!

thibaultcha:
> oh I was writing the review when this was merged

Oops, my bad!

PR #3827, opened 2018-10-04
feat(schema) add `certificate` and `key` typedefs

james-callahan:
Adds `certificate` and `key` typedefs

Used by the forthcoming service mesh CA entity.
However also useful for validating e.g. certificate entity.

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -1,6 +1,9 @@
 --- A library of ready-to-use type synonyms to use in schema definitions.
 -- @module kong.db.schema.typedefs
 local utils = require "kong.tools.utils"
+local openssl_pkey = require "openssl.pkey"
+local openssl_x509 = require "openssl.x509"
+
 local Schema = require("kong.db.schema")
 local socket_url = require("socket.url")
 
@@ -89,6 +92,28 @@ local function validate_url(url)
 
   if not parsed_url.scheme then
     return nil, "missing scheme in url"
+  end
+
+  return true
+end
+
+
+local function validate_certificate(cert)
+  local ok
+  ok, cert = pcall(openssl_x509.new, cert)
+  if not ok then
+    return nil, "invalid certificate: " .. cert
+  end
+
+  return true
+end
+
+
+local function validate_key(key)
+  local ok
+  ok, key = pcall(openssl_pkey.new, key)
+  if not ok then
+    return nil, "invalid key: " .. key
   end
 
   return true
@@ -212,6 +237,18 @@ typedefs.name = Schema.define {
 typedefs.sni = Schema.define {
   type = "string",
   custom_validator = validate_sni,
+}
+
+
+typedefs.certificate = Schema.define {
+  type = "string",
+  custom_validator = validate_certificate,
+}
+
+
+typedefs.key = Schema.define {
+  type = "string",
+  custom_validator = validate_key,
 }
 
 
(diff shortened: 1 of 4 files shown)


kong/db/schema/typedefs.lua
```diff
@@ -75,6 +78,28 @@ local function validate_sni(host)
 end
 
 
+local function validate_certificate(cert)
+  local ok
+  ok, cert = pcall(openssl_x509.new, cert)
+  if not ok then
+    return nil, "invalid cert: " .. cert
```
bungle: should we use `invalid certificate` instead of the `invalid cert` here?

spec/01-unit/000-new-dao/01-schema/03-typedefs_spec.lua
```diff
@@ -19,6 +21,50 @@ describe("typedefs", function()
...
+    do
+      local key = openssl_pkey.new { bits = 2048 }
+      local crt = openssl_x509.new()
+      crt:setPublicKey(key)
+      crt:sign(key)
+      assert.truthy(Test:validate({ f = crt:toPEM() }))
+    end
+    assert.falsy(Test:validate({ f = "not a certificate" }))
+    assert.falsy(Test:validate({ f = 42 }))
+    assert.falsy(Test:validate({ f = [[
```
thibaultcha: It's unfortunate that we aren't testing the actual error messages in all of these failures. We really should, to ensure that the test is failing (or rather, succeeding) where it should be.

james-callahan: I was following the form of the other tests in this file. Is there a particular pattern you'd like to see here?
(and in-fact I was essentially copying the general structure of commit 3176e2f1292f2943b1396fd17d51a882efdd6019)

thibaultcha: Hmm.. If you ask me, this file is guilty of the same incompleteness... Mind updating this patch and help in preventing the pattern from spreading around too much? We always explicitly test error messages throughout our test suite, to ensure we avoid regressions not only in their triggering, but also in their contents...

hishamhm:
@james-callahan LGTM, apart from the Travis failure (just looks like some test is using an invalid cert that needs to be fixed with the proper fixture).

Side comment — from the current Travis failure:

```
...integration/04-admin_api/06-certificates_routes_spec.lua:499: Invalid response status code.
Status expected:
(number) 200
Status received:
(number) 400
Body:
(string) '{"message":"2 schema violations (cert: invalid certificate: x509.cert.new: asn1_lib.c:101:error:0D07207B:asn1 encoding routines:ASN1_get_object:header too long; key: invalid key: pkey.new: tasn_dec.c:1130:error:0D0680A8:asn1 encoding routines:asn1_check_tlen:wrong tag)","name":"schema violation","fields":{"cert":"invalid certificate: x509.cert.new: asn1_lib.c:101:error:0D07207B:asn1 encoding routines:ASN1_get_object:header too long","key":"invalid key: pkey.new: tasn_dec.c:1130:error:0D0680A8:asn1 encoding routines:asn1_check_tlen:wrong tag"},"code":2}'
200
stack traceback:
	...integration/04-admin_api/06-certificates_routes_spec.lua:499: in function <...integration/04-admin_api/06-certificates_routes_spec.lua:490>
```

This is exposing a lot of internal implementation data in the error message. Is this information useful? (honest question, just want to know if these messages make sense — anyway, not a blocker)

james-callahan:
> This is exposing a lot of internal implementation data in the error message. Is this information useful? (honest question, just want to know if these messages make sense — anyway, not a blocker)

Some of it. The `x509.cert.new` prefix? No. The rest: mostly. The error numbers and locations in there are very googlable to track down the problem with the certificate.

james-callahan:
Fixed the misc failing test: was a raw-url-encoding one I missed.
Should be good to merge now? I've rebased on top of current next.

PR #3839, opened 2018-10-09
Feat/plugins code to new db

kikito:
Replaces usage of old-style DAO with new-style db in all plugins:

* Noticed a bug in the ACL plugin cache invalidation code. Fixed here.
* The main purpose of this task was replacing `singletons.dao` with calls to `kong.db`. <del>I also replaced `singletons.cache` with `kong.cache` and `singletons.configuration` with `kong.configuration`</del>.

### DIFF
--- a/spec-old-api/03-plugins/25-response-rate-limiting/02-policies_spec.lua
+++ b/spec-old-api/03-plugins/25-response-rate-limiting/02-policies_spec.lua
@@ -13,10 +13,10 @@ describe("Plugin: response-ratelimiting (policies)", function()
     local dao
 
     setup(function()
-      dao = select(3, helpers.get_db_utils())
+      local _, db
+      _, db, dao = helpers.get_db_utils()
 
-      local singletons = require "kong.singletons"
-      singletons.dao = dao
+      _G.kong = _G.kong or { db = db }
 
       dao:truncate_tables()
     end)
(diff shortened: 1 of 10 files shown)


kong/plugins/aws-lambda/handler.lua
```diff
@@ -100,7 +99,7 @@ local function send(status, content, headers)
     ngx.header["Content-Length"] = #content
   end
 
-  if singletons.configuration.enabled_headers[constants.HEADERS.VIA] then
+  if kong.configuration.enabled_headers[constants.HEADERS.VIA] then
```
hishamhm: this is going to have some minor conflicts when #3844 is merged...

spec-old-api/03-plugins/25-response-rate-limiting/02-policies_spec.lua
```diff
@@ -13,10 +13,10 @@ describe("Plugin: response-ratelimiting (policies)", function()
     local dao
 
     setup(function()
-      dao = select(3, helpers.get_db_utils())
+      local _, db
+      _, db, dao = helpers.get_db_utils("postgres", {})
```
hishamhm: is this test meant to be postgres-only?

kikito: It was implicitly pg-only (when `get_db_utils` gets no parameters it uses the one that's on the test config file, which is pg). I made it explicit while trying to debug why the tests were failing on my machine. I have undone this change (notice this is on `spec-old-api`, so no big deal).

PR #3841, opened 2018-10-10
 refactor(balancer) factor out pre-access and post-access balancer setup 

james-callahan:
Preparatory refactor for stream support.

This factors out the bits required to setup Kong's `balancer_by_lua` phase; leaving the HTTP-only bits behind.


thibaultcha:
Merged!

PR #3842, opened 2018-10-10
Feat/groundwork for zipkin

kikito:
This PR adds a couple features to the PDK:
* `kong.request.get_start_time()`, which wraps `ngx.req.start_time()`
* `kong.request.get_raw_path_and_query()`

It also fix a (possible?) problem when ngx.request_uri is nil.

It is needed for the zipkin update: https://github.com/Kong/kong-plugin-zipkin/pull/24

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -279,7 +279,7 @@ local function new(self)
   function _REQUEST.get_path()
     check_phase(PHASES.request)
 
-    local uri = ngx.var.request_uri
+    local uri = ngx.var.request_uri or ""
     local s = find(uri, "?", 2, true)
     return s and sub(uri, 1, s - 1) or uri
   end
@@ -298,7 +298,7 @@ local function new(self)
   -- kong.request.get_raw_path_and_query() -- "/v1/movies?movie=foo"
   function _REQUEST.get_raw_path_and_query()
     check_phase(PHASES.request)
-    return ngx.var.request_uri
+    return ngx.var.request_uri or ""
   end
 
   ---
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -641,6 +641,18 @@ local function new(self)
     end
   end
 
+  ---
+  -- Returns the time at which the request was created
+  --
+  -- @function kong.request.get_start_time
+  -- @phases rewrite, access, header_filter, body_filter, log
+  -- @treturn number the distance in seconds from epoch, with milliseconds as the decimal part.
+  -- @usage
+  -- kong.request.get_start_time() -- 290132101.123
+  function _REQUEST.get_start_time()
+    check_phase(PHASES.request)
+    return ngx.req.start_time()
+  end
 
   return _REQUEST
 end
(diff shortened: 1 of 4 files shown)


kong/pdk/request.lua
```diff
@@ -640,6 +640,19 @@ local function new(self)
...
   end
 
+  ---
+  -- Returns the time at which the request was created
+  --
+  -- @function kong.request.get_start_time
+  -- @phases rewrite, access, header_filter, body_filter, log
+  -- @treturn number the distance in seconds from epoch, with milliseconds as the decimal part.
+  -- @usage
+  -- kong.request.get_start_time() -- 290132101.123
```
hishamhm: wow, that's a very old request :)

kong/runloop/handler.lua
```diff
@@ -210,6 +210,11 @@ local function balancer_setup_stage1(ctx, scheme, host_type, host, port,
   ctx.route            = route
   ctx.balancer_data    = balancer_data
   ctx.balancer_address = balancer_data -- for plugin backward compatibility
+
+  -- plugin forwards-compatible way
+  kong.ctx.shared.service       = service
+  kong.ctx.shared.balancer_data = balancer_data
```
hishamhm: I've seen a WIP commit floating around that changed the name (and possibly content?) of this `balancer_data` structure. Not sure what will happen of that structure, but just a heads up. Also, I don't know what the plugin is doing with it, but we did add functions in `kong/pdk/service.lua` to avoid having plugins work on `balancer_data` directly. (I just noticed we only added setters and not getters — what does the plugin need?)

kikito: The trick is getting this one merged first then xD

> what does the plugin need?

It seems it gets statistics about each try, the total count, and the ip involved: https://github.com/Kong/kong-plugin-zipkin/blob/master/kong/plugins/zipkin/opentracing.lua#L185-L201

kikito: I have rolled back the usage of kong.ctx by zipkin - it was a bit premature. So I will also roll back this particular change too.

kong/pdk/request.lua
```diff
@@ -279,12 +279,28 @@ local function new(self)
...
+  -- Otherwise it returns just the path. No transformations/normalizations are done.
+  --
+  -- @function kong.request.get_path_and_querystring()
+  -- @phases rewrite, access, header_filter, body_filter, log
+  -- @treturn string the path and querystring
+  -- @usage
+  -- -- Given a request to https://example.com:1234/v1/movies?movie=foo
+  --
+  -- kong.request.get_path_and_querystring() -- "/v1/movies?movie=foo"
+  function _REQUEST.get_path_and_querystring()
```
bungle: I left comment about the name here:
https://github.com/Kong/kong/pull/3859#pullrequestreview-165268435

bungle:
I am not sure if `ngx.var.request_uri` can ever be `nil`, at least with `http` module. How could it be?

thibaultcha:
Merged to next with adjustments; except for the timing API commit

PR #3845, opened 2018-10-10
(refactor) plugins to remove dependencies to kong.tools and use more PDK

bungle:
### Summary

This contains changes to following plugins and their test suite? This is rebased on @hishamhm's https://github.com/Kong/kong/tree/feat/plugins-to-new-dao-api branch which should be merged before this.

- acl
- basic-auth
- correlation-id
- hmac-auth
- jwt
- ldap-auth
- rate-limiting (still has some flaky tests)
- request-termination

Also contains a small fix to a typo on key-auth plugin.

### DIFF
--- a/kong/plugins/acl/api.lua
+++ b/kong/plugins/acl/api.lua
@@ -1,8 +1,10 @@
 local endpoints = require "kong.api.endpoints"
-local responses = require "kong.tools.responses"
 
+
+local kong             = kong
 local acls_schema      = kong.db.acls.schema
 local consumers_schema = kong.db.consumers.schema
+
 
 return {
   ["/consumers/:consumers/acls/"] = {
@@ -24,7 +26,7 @@ return {
           return endpoints.handle_error(err_t)
         end
         if not consumer then
-          return responses.send_HTTP_NOT_FOUND()
+          return kong.response.exit(404, { message = "Not found" })
         end
 
         self.consumer = consumer
@@ -36,7 +38,7 @@ return {
 
         if self.req.cmd_mth ~= "PUT" then
           if not acl or acl.consumer.id ~= consumer.id then
-            return responses.send_HTTP_NOT_FOUND()
+            return kong.response.exit(404, { message = "Not found" })
           end
           self.acl = acl
           self.params.acls = acl.id
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -1,27 +1,28 @@
 local BasePlugin = require "kong.plugins.base_plugin"
-local responses = require "kong.tools.responses"
 local constants = require "kong.constants"
-local pl_tablex = require "pl.tablex"
+local tablex = require "pl.tablex"
 local groups = require "kong.plugins.acl.groups"
 
-local table_concat = table.concat
-local set_header = ngx.req.set_header
-local ngx_error = ngx.ERR
-local ngx_log = ngx.log
-local EMPTY = pl_tablex.readonly {}
+
+local setmetatable = setmetatable
+local concat = table.concat
+local kong = kong
+
+
+local EMPTY = tablex.readonly {}
 local BLACK = "BLACK"
 local WHITE = "WHITE"
 
+
 local mt_cache = { __mode = "k" }
 local config_cache = setmetatable({}, mt_cache)
-
 
 
 local ACLHandler = BasePlugin:extend()
 
 
 ACLHandler.PRIORITY = 950
-ACLHandler.VERSION = "0.1.1"
+ACLHandler.VERSION = "0.1.2"
 
 
 function ACLHandler:new()
@@ -44,16 +45,17 @@ function ACLHandler:access(conf)
   -- get the consumer/credentials
   local consumer_id = groups.get_current_consumer_id()
   if not consumer_id then
-    ngx_log(ngx_error, "[acl plugin] Cannot identify the consumer, add an ",
-                       "authentication plugin to use the ACL plugin")
-    return responses.send_HTTP_FORBIDDEN("You cannot consume this service")
+    kong.log.err("Cannot identify the consumer, add an authentication " ..
+                 "plugin to use the ACL plugin")
+    return kong.response.exit(403, { message = "You cannot consume this service" })
   end
 
   -- get the consumer groups, since we need those as cache-keys to make sure
   -- we invalidate properly if they change
   local consumer_groups, err = groups.get_consumer_groups(consumer_id)
   if not consumer_groups then
-    return responses.send_HTTP_INTERNAL_SERVER_ERROR(err)
+    kong.log.err(err)
+    return kong.response.exit(500, { message = "An unexpected error occurred" })
   end
 
   -- 'to_be_blocked' is either 'true' if it's to be blocked, or the header
@@ -71,8 +73,8 @@ function ACLHandler:access(conf)
     if to_be_blocked == false then
       -- we're allowed, convert 'false' to the header value, if needed
       -- if not needed, set dummy value to save mem for potential long strings
-      to_be_blocked = conf.hide_groups_header and "" 
-                      or table_concat(consumer_groups, ", ")
+      to_be_blocked = conf.hide_groups_header and ""
+                      or concat(consumer_groups, ", ")
     end
 
     -- update cache
@@ -80,12 +82,14 @@ function ACLHandler:access(conf)
   end
 
   if to_be_blocked == true then -- NOTE: we only catch the boolean here!
-    return responses.send_HTTP_FORBIDDEN("You cannot consume this service")
+    return kong.response.exit(403, { message = "You cannot consume this service" })
   end
 
   if not conf.hide_groups_header then
-    set_header(constants.HEADERS.CONSUMER_GROUPS, to_be_blocked)
+    kong.service.request.set_header(constants.HEADERS.CONSUMER_GROUPS,
+                                    to_be_blocked)
   end
 end
+
 
 return ACLHandler
--- a/kong/plugins/basic-auth/basicauth_credentials.lua
+++ b/kong/plugins/basic-auth/basicauth_credentials.lua
@@ -1,5 +1,20 @@
 local crypto = require "kong.plugins.basic-auth.crypto"
-local utils  = require "kong.tools.utils"
+
+
+local type = type
+local re_find = ngx.re.find
+
+
+local uuid_regex = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+
+
+local function is_valid_uuid(str)
+  if type(str) ~= 'string' or #str ~= 36 then
+    return false
+  end
+
+  return re_find(str, uuid_regex, 'ioj') ~= nil
+end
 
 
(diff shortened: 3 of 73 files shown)


kong/plugins/acl/api.lua
```diff
@@ -1,9 +1,11 @@
 local endpoints = require "kong.api.endpoints"
-local responses = require "kong.tools.responses"
 
+
+local kong             = kong
```
hishamhm: Is it okay to `local`-ize the `kong` global? I am not sure if it is, because in the future different versions of the PDK may be active, and the appropriate PDK version might be determined only after the module is `require`d... or are we avoiding this issue via metatable indirection? do we have a policy on that? pinging @thibaultcha

kong/plugins/key-auth/handler.lua
```diff
@@ -137,7 +137,7 @@ local function do_authentication(conf)
                                     key)
   if err then
     kong.log.err(err)
-    return kong.response.exit(500, "An unexpected error ocurred")
+    return kong.response.exit(500, "An unexpected error occurred")
```
hishamhm: :100:

kong/plugins/basic-auth/handler.lua
```diff
@@ -1,20 +1,24 @@
...
+
 function BasicAuthHandler:access(conf)
   BasicAuthHandler.super.access(self)
   access.execute(conf)
 end
 
+
 BasicAuthHandler.PRIORITY = 1001
-BasicAuthHandler.VERSION = "0.1.0"
+BasicAuthHandler.VERSION = "0.1.1"
```
hishamhm: oh, my PR did not upgrade the VERSION identifier of any plugins. Should we do that?

kong/plugins/correlation-id/handler.lua
```diff
@@ -1,78 +1,91 @@
...
+        var.server_addr,
+        var.server_port,
+        worker_pid,
+        var.connection, -- connection serial number
+        var.connection_requests, -- current number of requests made through a connection
+        now() -- the current time stamp from the nginx cached time.
+      )
+    end,
+  }, { __index = function(_, generator)
+      kong.log.err("Invalid generator: " .. generator)
```
hishamhm: This extra metatable that's added for error-checking only is effectively dead code, because the schema guarantees that `conf.generator` is always valid. I think we could remove this and make `generators` a simple table.

hishamhm: (I realize this was in the original code already, just noticed that and thought we could clean it up)

kong/plugins/hmac-auth/access.lua
```diff
@@ -28,6 +24,7 @@ local DIGEST = "digest"
 local SIGNATURE_NOT_VALID = "HMAC signature cannot be verified"
 local SIGNATURE_NOT_SAME = "HMAC signature does not match"
 
+
 local new_tab
```
hishamhm: new_tab can be removed and replaced by `kong.table.new`

kong/plugins/hmac-auth/access.lua
```diff
@@ -95,82 +94,87 @@ local function validate_params(params, conf)
...
+
+local function retrieve_hmac_fields(authorization_header)
   local hmac_params = {}
-  local authorization_header = headers[header_name]
+
   -- parse the header to retrieve hamc parameters
   if authorization_header then
-    local iterator, iter_err = ngx_gmatch(authorization_header, "\\s*[Hh]mac\\s*username=\"(.+)\",\\s*algorithm=\"(.+)\",\\s*headers=\"(.+)\",\\s*signature=\"(.+)\"")
+    local iterator, iter_err = re_gmatch(authorization_header,
+                                          "\\s*[Hh]mac\\s*username=\"(.+)\",\\s*algorithm=\"(.+)\",\\s*headers=\"(.+)\",\\s*signature=\"(.+)\"")
```
hishamhm: you can reduce the line by breaking this long string into shorter ones connected by `..` (don't worry, the compiler's constant-subexpression-elimination will make it into a single string)

kong/plugins/hmac-auth/access.lua
```diff
@@ -95,82 +94,87 @@ local function validate_params(params, conf)
...
+
   return hmac[hmac_params.algorithm](hmac_params.secret, signing_string)
 end
 
-local function validate_signature(request, hmac_params, headers)
-  local signature_1 = create_hash(request, ngx.var.request_uri, hmac_params, headers)
-  local signature_2 = ngx_decode_base64(hmac_params.signature)
+
+local function validate_signature(hmac_params)
+  local signature_1 = create_hash(ngx.var.request_uri, hmac_params)
```
hishamhm: It's unfortunate we still need to rely on `ngx.var.request_uri` here, but I guess it's because we don't have an exact equivalent in the PDK, right? Would this be our best shot sticking to the current PDK API?

```lua
local function get_path_and_query()
  local p = ngx.request.get_path()
  local q = ngx.request.get_raw_query()
  return (q ~= "") and (p .. "?" .. q) or p
end
```
(which is inefficient because `get_path` strips the query from `ngx.var.request_uri` and then this adds it back, I know.)

Is this pattern common enough to warrant a `kong.request.get_path_and_query` function? (The aws-lambda plugin does the same as the above, but that's mostly for backwards compat with the "request_uri" info being forwarded)

kong/plugins/request-termination/handler.lua
```diff
@@ -59,21 +67,29 @@ function RequestTerminationHandler:access(conf)
   local content = conf.body
 
   if content then
-    local ctx = ngx.ctx
-    if ctx.delay_response and not ctx.delayed_response then
-      ctx.delayed_response = {
+    local shared_ctx = kong.ctx.shared
```
hishamhm: The whole delayed_response logic (and flush function) can be replaced by `kong.response.exit`, which takes care of it. It is possible to just do

```lua
if content then
  return kong.response.exit(status, content, {
    ["Content-Type"] = conf.content_type
  })
end
```

instead of this whole block, also removing the `flush` function above.

kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -1,21 +1,65 @@
...
+local timetable = luatz.timetable
 local pairs = pairs
+local floor = math.floor
+local null = ngx.null
+local shm = ngx.shared.kong_rate_limiting_counters
 local fmt = string.format
 
 
-local NULL_UUID = "00000000-0000-0000-0000-000000000000"
+local log_err = kong and kong.log.err or function(...)
```
hishamhm: Is this here because of tests where `kong` is not available? If that's the case, I'd rather just populate `kong.log.err` in the test case before testing this module instead of leaking `ngx.log` here.

kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -1,21 +1,65 @@
...
+local function get_timetable(now)
+  if now > ms_check then
+    return timetable.new_from_timestamp(now / 1000)
+  end
+
+  return timetable.new_from_timestamp(now)
+end
+
+
+local function get_timestamps(now)
```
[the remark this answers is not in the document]
bungle: I can get back to just using `kong.tools.timestamp` module.

kong/plugins/acl/api.lua
thibaultcha: It should be fine yes :+1: The reference to the `kong` global should always be the same, and underlying metatables should be responsible for redirecting callers to the appropriate PDK version (or cached modules that are rotated, like the current logging facilities implementation). The "don't" would be to cache individual fields of the `kong` global (e.g. `kong.pdk_version`, or `kong.request.get_path`).

kong/plugins/acl/handler.lua
```diff
@@ -44,16 +45,17 @@ function ACLHandler:access(conf)
   -- get the consumer/credentials
   local consumer_id = groups.get_current_consumer_id()
   if not consumer_id then
-    ngx_log(ngx_error, "[acl plugin] Cannot identify the consumer, add an ",
-                       "authentication plugin to use the ACL plugin")
-    return responses.send_HTTP_FORBIDDEN("You cannot consume this service")
+    kong.log.err("Cannot identify the consumer, add an authentication " ..
```
thibaultcha: It is still preferred to use the variadic form of this function, (`,` instead of `..`).

kong/plugins/basic-auth/basicauth_credentials.lua
```diff
@@ -1,5 +1,20 @@
...
+local uuid_regex = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+
+
+local function is_valid_uuid(str)
+  if type(str) ~= 'string' or #str ~= 36 then
+    return false
+  end
+
+  return re_find(str, uuid_regex, 'ioj') ~= nil
+end
```
thibaultcha: Why are we duplicating `is_valid_uuid`? This is a blocker. The scope of this work is not to get rid of `kong.tools.utils`.

kong/plugins/hmac-auth/access.lua
```diff
@@ -1,24 +1,20 @@
...
-local ngx_decode_base64 = ngx.decode_base64
-local ngx_encode_base64 = ngx.encode_base64
-local ngx_parse_time = ngx.parse_http_time
-local ngx_set_header = ngx.req.set_header
-local ngx_get_headers = ngx.req.get_headers
-local ngx_log = ngx.log
-local req_read_body = ngx.req.read_body
-local req_get_body_data = ngx.req.get_body_data
-local ngx_hmac_sha1 = ngx.hmac_sha1
-local split = utils.split
```
thibaultcha: ditto - using `utils.split` is more consistent, because changing the underlying implementation of `split` can be done in a single place. Again, the goal isn't to get rid of `kong.tools.utils` here.

kong/plugins/jwt/asn_sequence.lua
```diff
@@ -1,128 +1,165 @@
...
+      input = sub(input, 2, -1)
     else
       break
     end
   end
+
   return input
 end
 
+
```
thibaultcha: Such vendor files ideally should be left untouched and are not subject to our coding style, as pulling fixes from upstream becomes harder.

kong/plugins/ldap-auth/handler.lua
```diff
@@ -1,18 +1,23 @@
...
+
 function LdapAuthHandler:access(conf)
   LdapAuthHandler.super.access(self)
   access.execute(conf)
 end
 
+
 LdapAuthHandler.PRIORITY = 1002
-LdapAuthHandler.VERSION = "0.1.0"
+LdapAuthHandler.VERSION = "0.1.1"
```
thibaultcha: All of these massive updates warrant increasing the second digit (e.g. `0.2.0`) in all of these plugins, instead of the patch digit.

kong/plugins/basic-auth/handler.lua
bungle: Probably doesn't matter really. I updated the version as it is different than the previous. Only the minor though.

kong/plugins/hmac-auth/access.lua
bungle: Fixed already.

PR #3848, opened 2018-10-11
feat(schema) do not run custom_entity_check if fields are invalid

hishamhm:
This is inspired by #3846.

For the generic checkers we had so far, running them on invalid fields and
reporting everything at once was okay, and avoided the "get one error, fix
thing, get another error" experience. But for a custom checker, it's safer to
only run it on valid fields (especially thinking of making things less
error-prone for third-party plugin authors).

The commit is lengthy because the `errors` table is now passed along
and is used to check whether there are recorded errors by the time
entity checkers run; a side effect of this is that the code is now
simpler, and produces fewer intermediate tables.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -460,6 +460,7 @@ Schema.entity_checkers = {
 
   at_least_one_of = {
     run_with_missing_fields = true,
+    run_with_invalid_fields = true,
     fn = function(entity, field_names)
       for _, name in ipairs(field_names) do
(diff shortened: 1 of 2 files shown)


james-callahan:
Is there a way that a custom validator could specify `run_with_invalid_fields` itself?

hishamhm:
@james-callahan not at this point, I didn't want to expose this unless it becomes actually needed, to avoid feature bloat/documentation overhead

PR #3860, opened 2018-10-16
refactor(db) replace db.old_dao:infos() with db.infos

bungle:
### Summary

Very tiny patch to replace `db.old_dao:infos` with `db.infos`.


thibaultcha:
Waiting for the tests to pass before merging.

PR #3861, opened 2018-10-17
feat(db) add generate_admin_api schema option

james-callahan:
Still needs tests.

### DIFF
--- a/kong/api/init.lua
+++ b/kong/api/init.lua
@@ -240,7 +240,7 @@ do
 
   -- Auto Generated Routes
   for _, dao in pairs(singletons.db.daos) do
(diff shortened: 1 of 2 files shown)


bungle:
Can you elaborate about the context? It seems like a decent patch otherwise.

james-callahan:
> Can you elaborate about the context?

Sorry for not explaining; this patch came out of a phone call with @hishamhm.

We don't want to expose the new CA entity (see https://github.com/Kong/kong/compare/next...james-callahan:feat/mesh-ca-bootstrap) via the admin api. This PR introduces a flag to opt-out.

PR #3862, opened 2018-10-17
fix(db) handle nils in copy_cassandra_records migration helper

hishamhm:
Fixes `copy_cassandra_records` to handle `nil` values and use `cassandra.null` instead of calling the cassandra type constructor (`cassandra.uuid(x)`, etc.)

(Does not include a regression test in this PR as this is in the process of being tested in https://github.com/kong/kong-upgrade-tests )

### DIFF
--- a/kong/db/migrations/helpers.lua
+++ b/kong/db/migrations/helpers.lua
@@ -186,7 +186,11 @@ function _M:copy_cassandra_records(source_table_def,
                             dest_column_name, source_table_def[dest_column_name])
           end
 
-          source_value = type_converter(source_value)
+          if source_value == nil then
+            source_value = cassandra.unset
+          else
+            source_value = type_converter(source_value)
+          end
 
         elseif type(source_value) == "function" then
           source_value = source_value(source_row)


kong/db/migrations/helpers.lua
```diff
@@ -186,7 +186,11 @@ function _M:copy_cassandra_records(source_table_def,
                             dest_column_name, source_table_def[dest_column_name])
           end
 
-          source_value = type_converter(source_value)
+          if source_value == nil then
+            source_value = cassandra.null
```
thibaultcha: An old problem that keeps resurfacing that I recalled: we really should use `cassandra.unset` instead, which helps with newer Cassandra versions:

http://thelastpickle.com/blog/2016/09/15/Null-bindings-on-prepared-statements-and-undesired-tombstone-creation.html

https://thibaultcha.github.io/lua-cassandra/modules/cassandra.html#type_serializers

thibaultcha:
```suggestion
            source_value = cassandra.unset
```

PR #3865, opened 2018-10-18
fix(db) Fix incorrect check for primary key violations under cassandra

james-callahan:
The old code would never throw a primary key violation.


bungle:
Great catch! Is there any possibility to make a test for it?

james-callahan:
> Is there any possibility to make a test for it?

There is a test that relies on this behaviour in my ca-bootstrap branch.
I guess we could pick an existing dao item and write a test for it if you'd like?

bungle:
@james-callahan sounds like a plan! Yes, lets have official test for it so that we don't hit this again without noticing.

PR #3872, opened 2018-10-18
fix(cors) properly validate flat strings

### LINKED ISSUE #3832: Cors plugin doesn't validate "flat strings" appropriately
### Summary

The Cors plugin states you can provide "flat strings or PCRE regexes" for the `config.origins` value.  However, flat string origins are not validated safely.

From what I can tell flat strings are evaluated with `re_find()` (see [here](https://github.com/Kong/kong/blob/acdd89d1ebb2d804913bd204ba6efc4fda351bbd/kong/plugins/cors/handler.lua#L54)). Since there are no line start/end (`^`, `$`) symbols surrounding the flat string, as long as an origin starts with the same characters it will be validated.

### Steps To Reproduce

1. Setup an clean Kong environment. I'm using Docker with `kong:0.13.0-alpine` and `postgres:10.1-alpine`, but I believe this bug exists in all versions of the plugin (see `re_find()` link above).
2. Setup the Cors plugin like this:
```
curl -X POST \
  http://localhost:8001/plugins \
  -H 'Content-Type: application/json' \
  -d '{
    "name": "cors",
    "config": {
      "origins": [
        "http://my-site.com",
        "http://my-other-site.com"
      ]
    }
  }'
```
3. Ping the Proxy with **your legitimate** origin and confirm **you do** get the `Access-Control-Allow-Origin` response header ✅ 
```
curl -X OPTIONS \
  http://localhost:8000 \
  -H 'Origin: http://my-site.com' 
```
4. Ping the Proxy with a **clearly illegitimate** origin and confirm **you do not** get the `Access-Control-Allow-Origin` response header ✅ 
```
curl -X OPTIONS \
  http://localhost:8000 \
  -H 'Origin: http://bad-guys.com' 
```
5. Ping the Proxy with a **tricky illegitimate** origin and confirm **you do not** get the `Access-Control-Allow-Origin` response header 😞 
```
curl -X OPTIONS \
  http://localhost:8000 \
  -H 'Origin: http://my-site.com.bad-guys.com' 
```

subnetmarco:
Closes #3832.

This pull request changes the behavior of the CORS plugin as follows:

* Forces a full string validation to avoid the issue in #3832 
* When a both valid and matched domain is processed, returns an `Access-Control-Allow-Origin` header value that equals the string sent by the client in the original `Origin` value
* Validates that the scheme, host and port also match the `Origin` value as per [RFC6454 Section 5](https://tools.ietf.org/html/rfc6454#section-5)

These changes seem reasonable to me, but happy to hear feedback.

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -1,7 +1,8 @@
 local BasePlugin = require "kong.plugins.base_plugin"
 local responses  = require "kong.tools.responses"
+local lrucache   = require "resty.lrucache"
 
-
+local url             = require "socket.url"
 local req_get_method  = ngx.req.get_method
 local re_find         = ngx.re.find
 local concat          = table.concat
@@ -14,6 +15,31 @@ local CorsHandler = BasePlugin:extend()
 
 CorsHandler.PRIORITY = 2000
 CorsHandler.VERSION = "0.1.0"
+
+
+-- per-worker cache of parsed origins
+local CACHE_SIZE = 10 ^ 4
+local parsed_domains
+
+
+local function parse_origin_domain(domain)
+  local parsed_obj = url.parse(domain)
+  if parsed_obj and parsed_obj.host then
+    local port = parsed_obj.port
+    if not port and parsed_obj.scheme then
+      if parsed_obj.scheme == "http" then
+        port = 80
+      elseif parsed_obj.scheme == "https" then
+        port = 443
+      end
+    end
+    return (parsed_obj.scheme and parsed_obj.scheme .. "://" or "") .. 
+            parsed_obj.host .. 
+            (port and ":" .. port or "")
+  else
+    return domain
+  end
+end
 
 
 local function configure_origin(ngx, conf)
@@ -50,8 +76,21 @@ local function configure_origin(ngx, conf)
 
   local req_origin = ngx.var.http_origin
   if req_origin then
+
+    local parsed_req_origin = parse_origin_domain(req_origin)
+
     for _, domain in ipairs(conf.origins) do
-      local from, _, err = re_find(req_origin, domain, "jo")
+      if not parsed_domains then
+        parsed_domains = lrucache.new(CACHE_SIZE)
+      end
+
+      local parsed_domain = parsed_domains:get(domain)
+      if not parsed_domain then
+        parsed_domain = parse_origin_domain(domain)
+        parsed_domains:set(domain, parsed_domain)
+      end
+
+      local from, _, err = re_find(parsed_req_origin, "^" .. parsed_domain .. "$", "jo")
       if err then
         ngx.log(ngx.ERR, "[cors] could not search for domain: ", err)
       end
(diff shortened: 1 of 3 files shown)


kong/plugins/cors/handler.lua
```diff
@@ -50,8 +50,11 @@ local function configure_origin(ngx, conf)
 
   local req_origin = ngx.var.http_origin
   if req_origin then
+
+    local parsed_req_origin = url.parse(req_origin).host or req_origin
+
     for _, domain in ipairs(conf.origins) do
-      local from, _, err = re_find(req_origin, domain, "jo")
+      local from, _, err = re_find(parsed_req_origin, "^"..(url.parse(domain).host or domain).."$", "jo")
```
p0pr0ck5: Can we cache the value of url.parse here?

kong/plugins/cors/handler.lua
```diff
@@ -16,6 +16,9 @@ CorsHandler.PRIORITY = 2000
 CorsHandler.VERSION = "0.1.0"
 
 
+local parsed_domains = { }
```
subnetmarco: Not sure if putting this variable here is the intended style.

kong/plugins/cors/handler.lua
```diff
@@ -16,6 +17,11 @@ CorsHandler.PRIORITY = 2000
 CorsHandler.VERSION = "0.1.0"
 
 
+-- per-worker cache of parsed origins
+local CACHE_SIZE = 10 ^ 4
```
subnetmarco: Will store at `10000` origins values at a time, seems reasonable.

kong/plugins/cors/handler.lua
```diff
@@ -50,8 +56,25 @@ local function configure_origin(ngx, conf)
 
   local req_origin = ngx.var.http_origin
   if req_origin then
+
+    local parsed_req_origin_obj = url.parse(req_origin)
+    local parsed_req_origin = parsed_req_origin_obj and parsed_req_origin_obj.host
```
thibaultcha: [RFC 6454](https://tools.ietf.org/html/rfc6454#page-11) says on "Comparing Origins" that:

>  Two origins are "the same" if, and only if, they are identical.  In
   particular: 
> If the two origins are scheme/host/port triples, the two origins
      are the same if, and only if, they have identical schemes, hosts,
      and ports.

Also see:

>  Q: Why not just use the host?
>   A: Including the scheme in the origin tuple is essential for
   security.  If user agents did not include the scheme, there would be
   no isolation between http://example.com and https://example.com
   because the two have the same host.  However, without this isolation,
   an active network attacker could corrupt content retrieved from
   http://example.com and have that content instruct the user agent to
   compromise the confidentiality and integrity of content retrieved
   from https://example.com, bypassing the protections afforded by TLS
   [RFC5246].

The comparisons we are doing now do not take scheme and port into consideration (nor do we have tests asserting them). See section 3.2.1 of RFC 6454 for examples of resources with similar and different origins, we could use this list to write our tests.

thibaultcha:
I have added a commit to this PR. Should be good to go now, but waiting for any additional review just in case. Will be merged for 1.0rc4.

PR #3876, opened 2018-10-19
fix(migrations) allow `kong migrations up` to go between 0.14 & 0.15

kikito:


hishamhm:
Oh, just saw this here. I don't think this _should_ version check at all (see my PR #3877), because there is no reason why running `kong migrations up` and `kong migrations finish` in a green-field Kong installation shouldn't work just as well as `kong migrations bootstrap`. Or am I missing something?

kikito:
Ok, closing this one down, let's continue on #3877

PR #3878, opened 2018-10-19
Fix/0.14.x plugins iterator

kikito:


hishamhm:
This will not work for 1.0 because kong.dao will be gone.

kikito:
You mean it will be `nil`, right? I'm counting on that.

kikito:
Oh you mean `kong.dao` *itself* will be nil so `kong.dao.plugins` will throw an error. I added an extra check for that. Thanks!

kikito:
Closing this down in favor of #3898

PR #3883, opened 2018-10-22
fix(db) type 'any' shouldn't cause metaschema validation errors

james-callahan:
Trying the proposal at https://github.com/Kong/kong/commit/3365902caa785fef3fcdbf723f2547342e336971#r30944834 I ran into an issue where `eq` was not accepted by the metaschema validator (though it seems to be accepted by the normal schema validator?)

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -736,8 +736,8 @@ function Schema:validate_field(field, value)
       return nil, validation_errors[field.type:upper()]
     end
 
-  -- if type is "skip" (an internal marker), run validators only
-  elseif field.type ~= "skip" then
+  -- if type is "any" or "skip" (an internal marker), run validators only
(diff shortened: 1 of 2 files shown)


hishamhm:
On second thought, I think this fix actually works because the types of fields are constrained by `Schema.valid_types`. Could you add the following test in `spec/01-unit/000-new-dao/01-schema/02-metaschema_spec.lua`?
 
```lua
  it("a schema cannot have a field of type 'any'", function()
    local s = {
      name = "hello",
      primary_key = { "foo" },
      fields = {
        { foo = { type = "any" } } } }
    local ok, err = MetaSchema:validate(s)
    assert.falsy(ok)
    assert.match("expected one of", err.fields.type)
  end)
```

hishamhm:
Rebasing it into current `next` should make tests pass.

PR #3884, opened 2018-10-22
feat(conf) add support for 'transparent' in listeners

james-callahan:
Requires https://github.com/Kong/openresty-patches/pull/10
Related to https://trac.nginx.org/nginx/ticket/287

### DIFF
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -545,7 +545,7 @@ end
 -- `port` (number), `ssl` (bool), `http2` (bool), `listener` (string, full listener)
 local function parse_listeners(values)
   local list = {}
-  local flags = { "ssl", "http2", "proxy_protocol" }
+  local flags = { "ssl", "http2", "proxy_protocol", "transparent" }
   local usage = "must be of form: [off] | <ip>:<port> [" ..
                 concat(flags, "] [") .. "], [... next entry ...]"
 
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -110,8 +110,11 @@
                          #   enabled.
                          # - `http2` will allow for clients to open HTTP/2
                          #   connections to Kong's proxy server.
-                         # - Finally, `proxy_protocol` will enable usage of the
+                         # - `proxy_protocol` will enable usage of the
                          #   PROXY protocol for a given address/port.
+                         # - `transparent` will cause kong to listen to, and
+                         #   respond from, any and all IP addresses and ports
+                         #   you configure in iptables.
                          #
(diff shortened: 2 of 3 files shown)


Tieske:
Shouldn't this include the output in the Nginx template (and related tests)?

james-callahan:
> Shouldn't this include the output in the Nginx template (and related tests)?

Tests would fail if the patch is not applied; @thibaultcha mentioned he didn't want them as a requirement if possible.

Tieske:
hmmm, they would fail if you'd try to use the listeners with that option? but just implementing it in the temnplates and testing whether the template gets the proper flags passed shouldn't be a problem right?

kong/conf_loader.lua
```diff
@@ -545,7 +545,7 @@ end
 -- `port` (number), `ssl` (bool), `http2` (bool), `listener` (string, full listener)
 local function parse_listeners(values)
   local list = {}
-  local flags = { "ssl", "http2", "proxy_protocol" }
+  local flags = { "ssl", "http2", "proxy_protocol", "transparent" }
```
Tieske: This would also add the "transparent" option to the http listeners no? shouldn't they be limited to the TCP listeners?

james-callahan: We want it available on both kinds. Infact stream support isn't in master/next right now, so this PR is *only* adding it for HTTP.

bungle:
One thing about this patch is that it needs usually `root` privileges to start a Kong. And that is a pretty huge change from the previous. There is also an option to make it work even without `root` when this is applied:
```
setcap cap_net_admin=eip /usr/local/openresty/nginx/sbin/nginx
```

This works at least on Linux and we may need to consider adding it to our distributions when this `transparent` feature gets applied.

james-callahan:
> When the patch is not applied to nginx, this option should be disabled, and an error printed to the user that a patch is missing.

How can we feature-detect this missing patch?

thibaultcha:
I was envisioning the patch introducing a new dummy function on the C side, and a protected call to load it via the FFI in our Lua. If the loading fails, we now the patch wasn't applied. Since our CLI runs with the `resty` interpreter, this would work for our configuration loading phase.
It is a bit hacky, but a price I'd be willing to pay to support vanilla OpenResty with good error messages for the neophytes, in case the patch is not applied (instead of the cryptic nginx configuration parsing error message saying `transparent is an unrecognized option`.

Anyway, due to our timing constraints, we can merge this as-is and consider the parsing error good enough, so I will be dismissing my review.

PR #3887, opened 2018-10-22
feat(migrations) add migration translation for plugins:select_by_cache_key

hishamhm:
:warning: **Note**: This is ready for review, but it still contains one `[wip]` commit forcing `db:schema_bootstrap()` in the `kong migrations up` command. Once `kong migrations bootstrap` is fixed to support migrations from 0.14, this can be removed and re-tested with kong-upgrade-tests.

-----

This implements a translation function for the migration of the cache_key value in the Plugins, so that the plugin runloop can work while the migration is not complete.

It attempts to use the proper `kong.db` mechanisms first in case the data is already available in the new format, and in case it isn't it, tries a lower-level approach to read the data in the old format (and in Cassandra's case, in the temporary table as well).

This PR also includes a couple of related bugfixes in additional commits.


kong/init.lua
```diff
@@ -329,6 +329,20 @@ function Kong.init()
...
+  local list = {}
+  for _, t in ipairs(migtable) do
+    local mignames = {}
+    for _, mig in ipairs(t.migrations) do
+      table.insert(mignames, mig.name)
+    end
+    table.insert(list, string.format("%s (%s)", t.subsystem,
+                       table.concat(mignames, ", ")))
+  end
+  return table.concat(list, ", ")
```
thibaultcha: A blank separator (`" "`) was the correct separator here (since `,` is used to separate migrations within the parenthesis). Only one of the two previous logging logics got it right :sweat_smile:

kong/init.lua
```diff
@@ -329,6 +329,20 @@ function Kong.init()
   assert(runloop.build_api_router(dao, "init"))
 end
 
+local function list_migrations(migtable)
```
thibaultcha: style: missing blank line above this function

kong/db/dao/plugins.lua
```diff
@@ -309,4 +309,21 @@ function Plugins:load_plugin_schemas(plugin_set)
...
+    return entity
+  end
+
+  -- try old format
+  local row, old_err = self.strategy:select_by_cache_key_migrating(key)
+  if row then
+    return self:row_to_entity(row)
+  end
+
+  return nil, (new_err and old_err) and new_err or old_err
```
thibaultcha: If we are trying the old format, isn't is most likely because `new_err` is set (e.g. to "invalid query: column 'cache_key' does not exist"). As such, wouldn't this condition supersede `old_err` (which may be a legitimate error from the legacy call) with this error from the new format call?

In short, shouldn't this be:
```lua
return nil, (new_err and old_err) and old_err or new_err
```

I am really unsure, maybe I missed another case this order covers.

kong/db/strategies/postgres/plugins.lua
```diff
@@ -0,0 +1,57 @@
...
+    local id = parts[i + 2]
+    if id ~= "" then
+      insert(qbuild, field .. " = '" .. id .. "'")
+    else
+      insert(qbuild, field .. " IS NULL")
+    end
+  end
+  local query = table.concat(qbuild, " AND ")
+
+  -- perform query, trying both temp and old table
```
thibaultcha: This comment is incorrect for the PostgreSQL strategy

kong/db/strategies/postgres/plugins.lua
```diff
@@ -0,0 +1,57 @@
...
+  local errs = 0
+  local last_err
+  for row, err in self.connector:iterate(query) do
```
thibaultcha: Do we need an iterator at all in this strategy? Since this is using SQL's `AND` condition instead of CQL's, are we ever going to get more than one row as a result to this query? The below condition (`if row then`) seems to indicate that it isn't the case.

kong/db/strategies/postgres/plugins.lua
```diff
@@ -0,0 +1,57 @@
...
+    if row then
+      row.cache_key = nil
+      return row
+    end
+
+    if err then
+      -- some errors here may happen depending of migration stage
+      errs = errs + 1
+      last_err = err
+    end
```
thibaultcha: Do we need this logic at all in this strategy?

kong/db/strategies/cassandra/plugins.lua
```diff
@@ -0,0 +1,74 @@
...
+    else
+      parts[i + 2] = nil
+    end
+  end
+  local query = table.concat(qbuild, " AND ") .. " ALLOW FILTERING"
+
+  -- perform query, trying both temp and old table
+  local errs = 0
+  local last_err
+  for _, tbl in ipairs({"plugins_temp", "plugins"}) do
```
thibaultcha: style: missing spaces inside of the table declaration

kong/db/dao/plugins.lua
hishamhm: I scratched my head for a while on this too, and I don't think there's a perfect answer. Considering the cases:

1. row is found
   a. migrated: new format returns row
   b. unmigrated: old format returns row
2. there is no row
  a. migrated: new_err and old_err are nil
  b. unmigrated: new_err is "cache_key doesn't exist", old_err is nil
3. new format produces real error
  a. migrated: most likely old will error too? so both new_err and old_err will be set
  b. unmigrated: new_err is "cache_key doesn't exist", old_err is something (case you brought up)
4. old format produces real error
  a. migrated: new_err is nil, old_err has error
  b. unmigrated: new_err is "cache_key doesn't exist", old_err has the error

The original order gives a bad message on case 3b indeed. But flipping it as you suggest would fix 3b but produce a bad message on case 3a, _I think_. Since "migrated" is the common case, I went with that one.

When we manage to disable this translator after the migration is complete, then it would essentially run on the unmigrated case, so flipping the condition as you suggested would be ideal.

kong/init.lua
hishamhm: But doesn't `foo (hello, world), bar (baz, bla), ble (bli, blo, blu)` actually look nicer? :)

hishamhm: à chacun son goût! ;) all right a space it is

PR #3893, opened 2018-10-23
hotfix(db) fix Plugins cache_key translation function for C* 2.x

hishamhm:
Complex queries (even with `ALLOW FILTERING`) can cause a Java exception in Cassandra 2. This PR simplifies the query performed by the migration translation function for `plugins:select_by_cache_key` when using Cassandra 2, to only filter by a single non-null value -- either one of the ids set, or the plugin name as a fallback -- and then further filter the results in Lua.

### DIFF
--- a/kong/db/strategies/cassandra/plugins.lua
+++ b/kong/db/strategies/cassandra/plugins.lua
@@ -19,9 +19,11 @@ function Plugins:select_by_cache_key_migrating(key)
   -- unpack cache_key
   local parts = split(key, ":")
 
+  local c3 = self.connector.major_version >= 3
+
   -- build query and args
-  local qbuild = { "SELECT * FROM %s WHERE name = ?" }
-  local args = { cassandra.text(parts[2]) }
+  local qbuild = {}
+  local args = {}
   for i, field in ipairs({
     "route_id",
     "service_id",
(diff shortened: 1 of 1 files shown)


thibaultcha:
I think a simpler patch could result in:
```lua
  -- build query and args
  local qbuild = { "SELECT * FROM %s WHERE name = ?" }
  local args = { parts[2] }

  if self.connector.major_version >= 3 then
    for i, field in ipairs({
      "route_id",
      "service_id",
       "consumer_id",
       "api_id",
    }) do
      local id = parts[i + 2]

      if id ~= "" then
        insert(qbuild, field .. " = ?")
        insert(args, cassandra.uuid(id))

      else
        parts[i + 2] = nil
      end
    end
  end

  local query = table.concat(qbuild, " AND ") .. " ALLOW FILTERING"
```

With a single branch for the C* >= 3 case

hishamhm:
> Edited to be optimized with at least one additional argument on top of name for C* 2.x as this patch already attempts to do

Not on top of, but instead of. The problem here is when we do `SELECT * BY plugins WHERE foo = ? AND bar = ? ALLOW FILTERING` (and that doesn't match an index), C* 2 doesn't like it. What I'm doing is, if we have a non-null id, use that only (we have indexes for each individual id, and for name) and it only scans by name if no id is given (e.g. when scanning for global plugins).

PR #3898, opened 2018-10-24
feat/self-deactivating translation functions

kikito:


kikito:
@thibaultcha I needed a place to put a function to "check if an individual migration is finished". I tried it adding it to db first:

```
db:is_schema_migration_executed(schema_state, subsystem, migration_name)
```

But then I saw that we already had this other method on db:

```
db:is_schema_up_to_date(schema_state)`
```

This, along with the fact that the `db.init` file was somewhat large, made me think that the schema_state was now "important enough" to be promoted to its own class. Extracting it from db was also quite clean: the local functions / requires moved almost verbatim, and I had to change `is_schema_up_to_date` in only 1 place.

Also, I thought was going to need to add caching on top of the `load` function at first. @hishamhm made me realize that for our current use that wasn't needed, since the plugin iterator *already* wraps the db access with a cache layer anyway. But in future migrations we'll probably need to cache (& invalidate) the `load` method. I did not want to add cache to `db.init`.

kong/db/init.lua
```diff
@@ -882,5 +614,9 @@ do
   --]]
 end
 
+<<<<<<< Updated upstream
```
hishamhm: There are remnants of a bad merge here

hishamhm:
I think it's a very clean refactor, and will help understandability of the codebase a lot (especially given the very similar names of the unrelated concepts of "kong.db schemas" and "database schemas" and the fact that the implementation of the latter actually uses the former!). I am very much in favor of splitting it into a separate `kong.db.migrations.state` module.

kong/db/dao/plugins.lua
```diff
@@ -310,19 +310,35 @@ end
...
+
+    -- if migration is complete, disable this translator function and return
+    if schema_state:is_migration_executed("core", "001_14_to_15") then
+      Plugins.select_by_cache_key = self.super.select_by_cache_key
+      return entity
+    end
   end
 
-  -- try old format
+  -- otherwise, we're either not started migrating, or we're migrating but
```
hishamhm: grammar: `we're either not started migrating, or we're migrating but` -> `we either have not started migrating, or we're migrating but`

PR #3900, opened 2018-10-24
chore(travis) compile with stream_ssl_preread_module

bungle:
### Summary

Compiles OpenResty with added `--with-stream_ssl_preread_module`.

This is needed for Service Mesh to be fully functional.


thibaultcha:
The proper scope for this should be `chore(ci)`, instead of `travis`. Was hoping to fix that at merge time.

PR #3909, opened 2018-10-25
feat(pdk) add consumer, service, route handling

Tieske:
Adds getters/setters to the PDK for the `consumer` (+`credential`), `service`, and `route` entities.


Tieske:
superseded by #3916

PR #3916, opened 2018-10-26
feat(pdk) add consumer, service, route handling

Tieske:
- Adds getters/setters to the PDK for the `consumer` (+`credential`), `service`, and `route` entities.
- refactor plugins to use the new PDK functions

Replaces #3909


kong/pdk/client.lua
```diff
@@ -121,6 +127,71 @@ local function new(self)
...
+  ---
+  -- Sets the authenticated consumer (and credential) for the current request.
+  -- @function kong.client.set_consumer
+  -- @phases access
+  -- @tparam table consumer The consumer to set, this can be `nil`.
+  -- @tparam table credential The credential to set, this can be `nil`.
+  -- @usage
+  -- -- assuming `credential` and `consumer` have been set by some authentication code
+  -- kong.client.set_consumer(consumer, credentials)
+  function _CLIENT.set_consumer(consumer, credential)
```
bungle: Where is `set_credential`? E.g. OpenID Connect Plugin can set `credential` without `consumer` for some reasons:
1. rate-limiting can then rate-limit by that
2. the `if conf.anonymous and ngx.ctx.authenticated_credential then ...` works with 3rd party authenticating plugins too.

Also I am not totally sure should we have `credential` here as a second argument, if we add `set_credential`.

Also I don't think these Kong concepts belong to `client`? Maybe `kong.consumer.set` and `kong.consumer.get` and `kong.credential.set` and `kong.credential.get`.

Tieske: I was wondering on `set_credential` as well. But updating the plugins (2nd commit) showed that this actually is ok. Because you should always set them both, if only a credential, then set the consumer to `nil`. Only makes the code more explicit.

wrt to `client` or `consumer` namespace, I think `client` is fine, auth is a downstream property imo.

bungle: Yes, fair enough. As we approach the higher level functionality in PDK it is impossible to be very unopionionated. Still thinking where should we place these.`kong.auth` is another option. Should these functions also take care of setting headers?

bungle: If consumer is sometimes optional should we then just have set_credential that takes consumer as second arg. Or do you see use case for consumer without credential? Maybe setting of anonymous consumer? But is it responsibility of these functions to set the upstream headers? Or should we patch core to do that?

Tieske: Setting the headers is a complex matter, because it involves removing, adding, and replacing headers. And the sets of headers differ per auth method.
That said, it is out of scope for this PR, I don't even think we can/should solve that now, since we'd first need to update the plugin-runloop to have something like an "auth" phase (basically moving the and/or logic for multiple-auth out of the individual plugins into core)

having independent set_consumer and set_credential is not a good idea imo.

bungle: Yes, that's why I was suggesting having just `set_credential`, e.g. possibly `kong.auth.set_credential(credential or , consumer or nil)` Or perhaps some table to set this:*
```lua
{
  consumer = ...,
  credential = ...,
}
```

Because of all the comments here (e.g. regarding to runloop), it would probably make sense to stop and think about this for a moment. As with PDK we want proper backward and forward compatibility.

kong/pdk/client.lua
```diff
@@ -121,6 +127,73 @@ local function new(self)
...
+  -- kong.client.authenticate(consumer, credentials)
+  function _CLIENT.authenticate(consumer, credential)
+    check_phase(PHASES.access)
+
+    if not TABLE_OR_NIL[type(consumer)] then
+      error("consumer must be a table or nil", 2)
+    elseif not TABLE_OR_NIL[type(credential)] then
+      error("credential must be a table or nil", 2)
+    end
+
```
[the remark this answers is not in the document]
Tieske: I removed the option, since it isn't used yet. Easy to relax later or add a `clear_auth` method or something similar.

kong/pdk/client.lua
```diff
@@ -121,6 +127,75 @@ local function new(self)
...
+  -- end
+  function _CLIENT.get_consumer()
+    check_phase(AUTH_AND_LATER)
+
+    return ngx.ctx.authenticated_consumer
+  end
+
+
+  ---
+  -- Sets the authenticated consumer (and credential) for the current request.
```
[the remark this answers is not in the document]
kikito: I think the description also needs to have a note "While both `consumer` and `credential` can be nil, it is required that at least one if them exists. Otherwise this function will throw an error".

PR #3930, opened 2018-10-29
Feat/plugins mesh field 1/2

kikito:
This PR includes:
* A fix on the schema, found by chance while studying how to best implement the new field
* A new field called `mesh_mode` to tag plugins with `inbound`, `outbound` or `two-way`. Note that `two-way` plugin instances can be individually tagged as `inbound` or `outbound` on creation, but plugins tagged as `inbound` can only have `inbound` instances, and the same goes for `outbound` pluggins.

I have tagged 15 plugins as `two-way` on this PR. There are multiple commits with explanations for my reasons.

### DIFF
--- a/kong/db/migrations/core/001_14_to_15.lua
+++ b/kong/db/migrations/core/001_14_to_15.lua
@@ -123,6 +123,15 @@ return {
       END;
       $$;
 
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "plugins" ADD "run_on" TEXT;
+      EXCEPTION WHEN DUPLICATE_COLUMN THEN
+        -- Do nothing, accept existing state
+      END;
+      $$;
+
+      CREATE INDEX IF NOT EXISTS "plugins_run_on_idx" ON "plugins" ("run_on");
 
 
       ALTER TABLE IF EXISTS ONLY "apis"
@@ -190,7 +199,6 @@ return {
       CREATE INDEX IF NOT EXISTS routes_name_idx ON routes(name);
 
 
-
       CREATE TABLE IF NOT EXISTS plugins_temp(
         id uuid,
         created_at timestamp,
@@ -198,6 +206,7 @@ return {
         route_id uuid,
         service_id uuid,
         consumer_id uuid,
+        run_on text,
         name text,
         config text, -- serialized plugin configuration
         enabled boolean,
@@ -241,6 +250,7 @@ return {
           route_id = "uuid",
           service_id = "uuid",
           consumer_id = "uuid",
+          run_on = "text",
           created_at = "timestamp",
           enabled = "boolean",
           cache_key = "text",
@@ -285,6 +295,7 @@ return {
           route_id uuid,
           service_id uuid,
           consumer_id uuid,
+          run_on text,
           name text,
           config text, -- serialized plugin configuration
           enabled boolean,
@@ -298,6 +309,7 @@ return {
         CREATE INDEX IF NOT EXISTS ON plugins(service_id);
         CREATE INDEX IF NOT EXISTS ON plugins(consumer_id);
         CREATE INDEX IF NOT EXISTS ON plugins(cache_key);
+        CREATE INDEX IF NOT EXISTS ON plugins(run_on);
       ]]))
 
       plugins_def = {
@@ -310,6 +322,7 @@ return {
           route_id = "uuid",
           service_id = "uuid",
           consumer_id = "uuid",
+          run_on = "text",
           created_at = "timestamp",
           enabled = "boolean",
           cache_key = "text",
@@ -325,6 +338,7 @@ return {
         route_id = "route_id",
         service_id = "service_id",
         consumer_id = "consumer_id",
+        run_on = "run_on",
         created_at = "created_at",
         enabled = "enabled",
         cache_key = "cache_key",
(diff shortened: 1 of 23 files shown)


hishamhm:
@kikito You have added a new top-level field to the Plugins entity in the db, but in the schemas you added it as a subfield of the `config` entry. If `mesh_mode` is a top-level Plugins field, it should be alongside `config` in the schemas, and not inside it.

kong/db/schema/typedefs.lua
```diff
@@ -252,4 +252,31 @@ typedefs.key = Schema.define {
...
+  default = "outbound",
+  eq = "outbound",
+  nullable = false,
+}
+
+
+typedefs.mesh_two_way = Schema.define {
+  type = "string",
+  default = "two-way",
+  one_of = { "inbound", "outbound", "two-way" },
```
Tieske: **\<opinion\>** use `twice` or `both` instead of `two-way` **\</opinion\>**

Tieske:
The auth plugins are all `two-way`, but I think their default value should be outbound. The functional effect will be the same. You want it to run on `outbound` to stop 'unwanted' traffic asap, and if it passes on one end, it will also pass on the other end (inbound), so there is no use in authenticating again.

Conceptually: currently the available values are tied to the default value. And I think they should be separate.

coopr:
> The auth plugins are all `two-way`, but I think their default value should be outbound. The functional effect will be the same. You want it to run on `outbound` to stop 'unwanted' traffic asap, and if it passes on one end, it will also pass on the other end (inbound), so there is no use in authenticating again.

I was imagining all the auth plugins would be `inbound` - that is conceptually most similar to how Kong API Gateway works, and I think it is easier to reason about (eg. "OK, I'm configuring an auth plugin that'll decide who gets past the proxy and gets to talk to my application..." - that makes it all about inbound traffic to the local proxy+application, and makes no assumptions about the source of the traffic)

bungle:
In service mesh it is different. The forward proxy already knows that authentication is going to fail, so it stops the traffic on a sidecar (Kong proxy) before even entering the network. Eg. traffic is stopped on forward proxy's incoming traffic and not on the callee service reverse proxy incoming traffic. I'd like to call these forward and reverse proxies, e.g. plugin is executed on forward proxy or reverse proxy or both. Most of the auth plugins will be executed on service mesh only on forward proxy e.g. The traffic that is trying to enter the mesh. After The traffic has entered the mesh, it is trusted traffic by mutual tls. All The plugins still have same incoming and outgoing nginx phases. It is just a matter that will the plugin run on one proxy (and which) or both. I would use `forward proxy`, `reverse proxy` and `both|always|...`, e.g. `run-on="forward-proxy"`. You could call forward and reverse proxies as `exit` and `enter` If you think about network structure. Traffic exits the service a and enters service b. But forward-proxy and reverse-proxy are My favourites.

bungle:
After talking with James, I got new info, so new proposal for naming:
 
`run-on="gateway|mesh|both"`

Because we don't know before actual connection is made whether or not we are in service mesh or not we call everything by default a gateway. All forward proxies in service mesh are gateways (gw to the mesh), all reverse proxies in service mesh are mesh nodes. So only the incoming can be a mesh node in service mesh. If there is no service mesh, then all Kong nodes are gateways, and usually they are reverse proxies (e.g. ingress). The role cannot change between different Nginx phases (by design).

So if you think about this, then all our current plugins should be marked run-on="gateway". But we may adjust some (the zipkin tracing to be "both"). We do not have mesh only plugins, e.g. run only on incoming (reverse proxy) mesh.

bungle:
^ the above is the default hard coded plugin level settings. The next question obviously is that do we allow change these defaults on plugin instance level, e.g. plugin.config.run_on=mesh overrides the hardcoded plugin.run_on=gateway.

bungle:
Okay, another take on this:

`run_on="untrusted|trusted|both"`
(pick a better name for `run_on`)

The `untrusted` is the **default**. E.g. all the traffic that enters Kong will be untrusted. And all the plugins are by default configured `untrusted`. It makes sense to run e.g. auth plugins on untrusted connections, e.g. service 1 connecting its own side car proxy (that is untrusted). The traffic can only become trusted when the connection is made in service mesh and authenticated with mutual tls.

The `trusted` can thus only be the receiving end of the mesh in this simple exercise of `SERVICE-A -> KONG-FORWARD-PROXY (a sidecar for SERVICE-A) -> KONG-REVERSE-PROXY (a sidecar for SERVICE-B) -> SERVICE-B`. Yes the `KONG-REVERSE-PROXY` is the `trusted` and `KONG-FORWARD-PROXY` is the `untrusted`.

And then we have `both` (possibly for `zipkin` only for now) which just means that plugin runloop will execute regardless if the source of the request (it doesn't matter if the connection was trusted or untrusted).

And again we can have `hard-coded` default on `plugin` `schema`:
`plugin.run_on="untrusted"` (probably all except zipkin)

and then a runtime override on plugin configuration (if the user for some weird reason wants to run it differently than the sane default):
`plugin.config.run_on="both"`

kong/db/migrations/core/001_14_to_15.lua
```diff
@@ -123,6 +123,15 @@ return {
       END;
       $$;
 
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "plugins" ADD "run_on" TEXT UNIQUE;
```
bungle: I don't think this should be `UNIQUE`.

kikito: fixed, thanks!

PR #3947, opened 2018-11-03
fix(http-log) remove res nil pointer possibility

jeremyjpj0916:
### Summary
res can be nil during bad tx, wrap such in if/else, added additional retry logic for failed requests.

Side note, not sure if Kong wants it, there may be value in logging when a non 200 response comes back too, my snippet is here. Can probably do without the tostring() for perf if you want to bring it into the PR:
```
    if not res then
      log(ERR, "failed request to ", host, ":", tostring(port), ": ", err)
      return false
    else
      res:read_body()
      if res.status ~= 200 then
         log(ERR, "Http log host returned status code: ",  tostring(res.status))
      end      
    end
```

Another consideration may be not doing res:read_body() in the code at all as it serves no purpose at the moment that I can see.

### Full changelog
* Modify /kong/plugins/http-log/sender.lua


subnetmarco:
_Note: We should check the other logging plugins to find out if they should also be updated._

kong/plugins/http-log/sender.lua
```diff
@@ -75,12 +75,13 @@ local function send(self, bodies)
     })
     if not res then
       log(ERR, "failed request to ", host, ":", tostring(port), ": ", err)
+      return false
+    else
```
Tieske: since the previous block already returns, the else branch is unnecessary

thibaultcha:
> Another consideration may be not doing res:read_body() in the code at all as it serves no purpose at the moment that I can see.

We need it since we keep the connection alive and we need to consume the data from the buffer before it becomes reused by another request.

kong/plugins/http-log/sender.lua
jeremyjpj0916: Cleaned up!

kong/plugins/http-log/sender.lua
```diff
@@ -75,6 +75,7 @@ local function send(self, bodies)
     })
     if not res then
       log(ERR, "failed request to ", host, ":", tostring(port), ": ", err)
+      return false
```
thibaultcha: Before exiting from this function, we still need the call to `set_keepalive()`.

Something like this may work:
```lua
local res, err = httpc:request(...)
if not res then
  log(ERR, ...)

else
  res:read_body()
end

ok, err = httpc:set_keepalive(...)
if not ok then
  log(ERR, ...)
end

return res ~= nil
```

thibaultcha:
>  So here: https://github.com/Kong/kong/blob/033eb94043ca5e743abab92b0becf19f41bbf13c/kong/plugins/http-log/sender.lua#L45 does not count for a new instantiation of the http client every time?

On the Lua-land, it is a new object, but since previous requests put the socket into the connection pool via `set_keepalive()`, we cannot assume that the underlying socket is new, it may come from the connection pool. See https://github.com/openresty/lua-nginx-module#tcpsocksetkeepalive and note:

> When the system receive buffer for the current connection has unread data, then this method will return the "connection in dubious state" error message (as the second return value) because the previous session has unread data left behind for the next session and the connection is not safe to be reused.

kong/plugins/http-log/sender.lua
jeremyjpj0916: I dig this, will commit this version of it.

jeremyjpj0916:
Committed my latest version. This moves the keepalive setting above the request, which feels fine to me to get "settings" out of the way prior to request. Lemme know if any issues seen.

kong/plugins/http-log/sender.lua
thibaultcha: I really should have expanded the diff here :joy:

kong/plugins/http-log/sender.lua
```diff
@@ -58,7 +58,12 @@ local function send(self, bodies)
         return false
       end
     end
-
+    
+    ok, err = httpc:set_keepalive(self.keepalive)
```
thibaultcha: `set_keepalive` must be called _after_ `request`, or else the connection will be considered closed from the Lua-land socket object by the time the request is made. See the documentation:

> This method also makes the current cosocket object enter the "closed" state, so there is no need to manually call the close method on it afterwards.

jeremyjpj0916: I just keep taking L's lol. But at least I did recognize the nil pointer 😸 . Did another commit, should take care of everything now. If yall have a cleaner preference let me know.

kong/plugins/http-log/sender.lua
```diff
@@ -73,18 +73,19 @@ local function send(self, bodies)
       },
       body = body,
     })
+    ok, err2 = httpc:set_keepalive(self.keepalive)
```
thibaultcha: So, this is the same issue: if you call `set_keepalive` before calling `receive` (what `read_body` does under the hood), you'll get an error because the Lua-land socket object will be flipped to the "closed" state.

I have updated my previous suggestion to reflect how we usually handle this scenario (keepalive on a failed request), plus updated the return statement which should only be called if `res == nil`.

It seems like a test is warranted here, since the CI failed to catch any issue with the last few proposals around this patch, which is worrying if we want to prevent future regressions...

kong/plugins/http-log/sender.lua
```diff
@@ -58,7 +58,7 @@ local function send(self, bodies)
         return false
       end
     end
-
+
```
thibaultcha: This patch introduced some undesired whitespaces here.

jeremyjpj0916:
I think the code is in a good state right now. On tests not sure how to best capture these, any tests already out there around this new http-log plugin and what you are thinking would be good to see?

kong/plugins/http-log/sender.lua
jeremyjpj0916: Fixed.

thibaultcha:
The best way would probably be to use a faulty HTTP server that would accept a connection, but then produce an invalid HTTP response or cause a timeout. Maybe our mock HTTP server can do the job here (see [this endpoint](https://github.com/Kong/kong/blob/master/spec/fixtures/custom_nginx.template#L311)). See how we make use of it in some integration tests: https://github.com/Kong/kong/blob/master/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua#L87 and how you can retrieve its URL from the `spec.helpers` module: https://github.com/Kong/kong/blob/master/spec/helpers.lua#L1151-L1152

In this case, we would configure an http-log plugin to point to this faulty HTTP endpoint, and configure a very low timeout, so that the test runs in a timely fashion (slow tests asserting 10s timeout are in effect are no fun...).

jeremyjpj0916:
Also followup here after re-reading some of the thread got me thinking.

What was suggested
```
local res, err = httpc:request(...)
if not res then
  log(ERR, ...)
  httpc:set_keepalive(...) -- ignore errors
  return false
end

res:read_body()

local ok, err = httpc:set_keepalive(...)
if not ok then
  log(ERR, ...)
end
```
You mentioned:
if you call set_keepalive before calling receive (what read_body does under the hood), you'll get an error because the Lua-land socket object will be flipped to the "closed" state.
I have updated my previous suggestion to reflect how we usually handle this scenario (keepalive on a failed request)

Isn't calling the set_keepalive going to fail then in the not res block always because we are doing so before a proper read_body could be executed? Would it not just be dead code and returning false before even attempting setting the keep-alive be more appropriate?

thibaultcha:
No, the connection is still opened but the socket received no data yet since the request was not sent.

jeremyjpj0916:
@thibaultcha I tried adding a new test to capture these sorts of things... It feels incomplete but I am not sure what do to here:

I added a new route and a http-log config that would be faulty(calling the delay 1 second mock, while setting a timeout on the http-log plugin itself to 1 millisecond):
```
      local route6 = bp.routes:insert {
        hosts   = { "https_logging_faulty.test" },
        service = service2
      }

      bp.plugins:insert {
        route = { id = route6.id },
        name     = "http-log",
        config   = {
          http_endpoint = "https://" .. helpers.mock_upstream_ssl_host
                                     .. ":"
                                     .. helpers.mock_upstream_ssl_port
                                     .. "/delay/1",
          timeout = 1
        }
      }
```

Then well for the test, I call the route(which under the hood would invoke the faulty http-log plugin config with the delay endpoint which should run into the timeout. Which old code would throw a nil pointer exception trying to read res:body(), BUT does a unit test like this see a nil pointer exception and fail the test? I imagine it has to be achieved with asserts but I don't know what could go in this test.

```
    it("gracefully handles layer 4 failures", function()
      local res = assert(proxy_client:send({
        method  = "GET",
        path    = "/status/200",
        headers = {
          ["Host"] = "https_logging_faulty.test"
        }
      }))
      assert.res_status(200, res)
    end)
```

thibaultcha:
I’m on a flight now, so this will be a short reply, sorry. 

Parsing the log files and making sure that no error were written to it is an option, we do this in a few test cases (unfortunately we do not have a help function for that, but there is previous set to draw inspiration from).

thibaultcha:
I’m on a flight now, so this will be a short reply, sorry. 

Parsing the log files and making sure that no error were written to it is an option, we do this in a few test cases (unfortunately we do not have a help function for that, but there is previous set to draw inspiration from).

jeremyjpj0916:
@thibaultcha short reply was all that was needed, found the code snippet you were describing in a test case and added it to mine.

jeremyjpj0916:
Also adding one more thing you will want from another issue: #3970, I fixed it in my splunk code but didn't think to add it here. More the merrier to knock out getting yalls codebase ship shape here. 

Was going to do something like this as a test case for it too, but is it possible to declare a global version of the plugin in this unit test that would not impact the other unit tests, is there a different unit test that gets used 

```
    it("successfully executes globally when routes don't exist", function()
       --This assumes http-log is already enabled on a global level for the kong node!

    	-- setup: cleanup logs
      local test_error_log_path = helpers.test_conf.nginx_err_logs
      os.execute(":> " .. test_error_log_path)

      local res = assert(proxy_client:send({
        method  = "GET",
        path    = "/nonexistant/proxy/path",
        headers = {
          ["Host"] = "https_no_exist.test"
        }
      }))
      assert.res_status(404, res)

      -- Assertion: there should be no [error], including no error
      -- resulting from attempting to reference the id on
      -- a route or api when no such value exists

      local pl_file = require "pl.file"
      local logs = pl_file.read(test_error_log_path)

      for line in logs:gmatch("[^\r\n]+") do
        assert.not_match("[error]", line, nil, true)
      end
    end)
```

Its true you can eyeball the change and automatically know why that extra ``` or "somevalue" ``` is necessary if Kong is invoked on a route that does not exist and the plugin is global.

kong/plugins/http-log/handler.lua
```diff
@@ -47,7 +47,7 @@ end
 function HttpLogHandler:log(conf)
   HttpLogHandler.super.log(self)
 
-  local route_id = conf.route_id or conf.api_id
+  local route_id = conf.route_id or conf.api_id or "global"
```
thibaultcha: Mind making this part of _another_ commit (could be the same PR, could be a different one), with its own test case (most importantly)? Thank you.

jeremyjpj0916: That code change requires a global test case to produce a situation that could cause regression. Can I just make a new test case lua file where the plugin is defined as global? My most recent PR comment was around that problem.

thibaultcha: No need, you can write this test within a `describe` block that has a `setup` phase clearing up the database.

jeremyjpj0916: @thibaultcha how we lookin now? I think I figured out how to make a new describes block and I believe to enable a plugin globally using this bp.plugins:insert{} you just drop setting the route id's, I struggled to find globally enabled plugin examples though during a 1 minute sweep searching the word global. Also broken out into 2 diff commits.

spec/03-plugins/03-http-log/01-log_spec.lua
```diff
@@ -269,4 +312,67 @@ for _, strategy in helpers.each_strategy() do
       end, 10)
     end)
   end)
-end
+
+  describe("Plugin: http-log (log) [#" .. strategy .. "]", function()
+    local proxy_client
+
+    setup(function()
```
thibaultcha: Note: we now use `lazy_setup` (see the recent commits on master)

spec/03-plugins/03-http-log/01-log_spec.lua
```diff
@@ -269,4 +312,67 @@ for _, strategy in helpers.each_strategy() do
       end, 10)
     end)
   end)
-end
+
+  describe("Plugin: http-log (log) [#" .. strategy .. "]", function()
```
thibaultcha: This describe block needs to play nicely with its parent block. As of now, this test’s name would be:

```
Plugin: http-log (log) [#postgres] Plugin: http-log (log) [#postgres] successfully executes gobally [...]
```

Something like this would be preferable:
```
Plugin: http-log (log) [#postgres] enables globallly executes successfully
```

With:
```
describe("enabled globally")
it("executes successfully")
```

This way, the test are properly scopes within a describe block about usage of this plugin when enabled globally, and the name is readable, plus the door is open for more `it()` test cases (e.g. `it("sets the name to 'global'")`)

spec/03-plugins/03-http-log/01-log_spec.lua
```diff
@@ -269,4 +312,67 @@ for _, strategy in helpers.each_strategy() do
...
+
+    after_each(function()
+      if proxy_client then
+        proxy_client:close()
+      end
+    end)
+
+    it("successfully executes globally when route does not exist", function()
+    	-- setup: cleanup logs
+      local test_error_log_path = helpers.test_conf.nginx_err_logs
```
thibaultcha: We should not do this ourselves, a previous call to `stop_kong` (from the sibling describe block for example) should take care of it for us.

spec/03-plugins/03-http-log/01-log_spec.lua
```diff
@@ -269,4 +312,67 @@ for _, strategy in helpers.each_strategy() do
...
+      -- Assertion: there should be no [error], including no error
+      -- resulting from attempting to reference the id on
+      -- a route or api when no such value exists
+
+      local pl_file = require "pl.file"
+      local logs = pl_file.read(test_error_log_path)
+
+      for line in logs:gmatch("[^\r\n]+") do
+        assert.not_match("[error]", line, nil, true)
+      end
```
thibaultcha: It seems like this test is not asserting that the plugin actually ran? Only that no error was thrown

kong/plugins/http-log/handler.lua
thibaultcha: It’s looking good @jeremyjpj0916! Just a few tweaks to adjust here and there :) I am confident this will make it in time for 1.0 :) Thank you!

spec/03-plugins/03-http-log/01-log_spec.lua
jeremyjpj0916: Agreed, whats the best way to validate a global plugin executed? I just sort of assumed it ran 😆, but certainly makes sense to validate that is the case.

thibaultcha: By copying any of the existing sanity tests in this suite (like “logs to http”) :)

spec/03-plugins/03-http-log/01-log_spec.lua
jeremyjpj0916: I think my last commit resolved this.

spec/03-plugins/03-http-log/01-log_spec.lua
jeremyjpj0916: Added lazy_setup/lazy_tear_down, I am all for lazy :D .

spec/03-plugins/03-http-log/01-log_spec.lua
jeremyjpj0916: So i just committed what I think you wanted, basically don't do the cleanup part at the top anymore but just reference the helpers nginx_err_logs down below. That's what you wanted right?

spec/03-plugins/03-http-log/01-log_spec.lua
jeremyjpj0916: Attempt made if I am reading the code right. Thanks for the patience! Just a newb wanting to help make Kong great 👍 . Don't work too hard this Thanksgiving yo 😃 .

jeremyjpj0916:
Good stuff. Thanks to all involved in progressing this one 👍 .

PR #3951, opened 2018-11-05
refactor(oauth2) drop support for api entity

bungle:
### Summary

This PR removes `API` entity support from the `OAuth 2.0 Plugin` and removes the associated columns and indexes from a database.


kong/plugins/oauth2/migrations/001_14_to_15.lua
```diff
@@ -100,10 +100,11 @@ return {
     ]],
 
     teardown = function(connector)
+      assert(connector:connect_migrations())
```
thibaultcha: It seems this patch belongs to the `next` branch? Will you backport it there?

thibaultcha:
Merged!

PR #3956, opened 2018-11-07
refactor(ratelimiting-plugins) drop support for api entity

bungle:
### Summary

This PR removes API entity support from the `rate-limiging` and `response-ratelimiting` plugins and removes the associated columns and indexes from a database.

It also drops the old migrations for these plugins.

One **important note**:

For Postgres it migrates the table's primary key without dropping the data, but for Cassandra it re-creates the table and drops the data. We might re-consider this after `RC3`.


thibaultcha:
Merged

Issue #3924, opened 2018-10-29
Kong 1.0rc2 migrations from 0.14.1 to 1.0rc2 issues


jeremyjpj0916:
Figured Kong may want some external feedback on what migrations looks like from an outside user. So here is my initial thought based on what I see: 

### Summary
Generally I expect conclusive results at the end of a migrations informing me everything completed successfully or not. My first run of the migrations from a Cassandra database running the 0.14.1 Kong using Cassandra 3.11.x seems to have some mixed bag of results. Log detailed below

```
2018/10/29 04:22:23 [verbose] Kong: 1.0.0rc2
2018/10/29 04:22:23 [debug] ngx_lua: 10013
2018/10/29 04:22:23 [debug] nginx: 1013006
2018/10/29 04:22:23 [debug] Lua: LuaJIT 2.1.0-beta3
2018/10/29 04:22:23 [verbose] no config file found at /etc/kong/kong.conf
2018/10/29 04:22:23 [verbose] no config file found at /etc/kong.conf
2018/10/29 04:22:23 [verbose] no config file, skip loading
2018/10/29 04:22:23 [debug] reading environment variables
2018/10/29 04:22:23 [debug] KONG_CASSANDRA_USERNAME ENV found with "dbUsername"
2018/10/29 04:22:23 [debug] KONG_PG_USER ENV found with ""
2018/10/29 04:22:23 [debug] KONG_CASSANDRA_DATA_CENTERS ENV found with "DC1:3,DC2:3"
2018/10/29 04:22:23 [debug] KONG_CASSANDRA_PASSWORD ENV found with "******"
2018/10/29 04:22:23 [debug] KONG_PG_HOST ENV found with ""
2018/10/29 04:22:23 [debug] KONG_CASSANDRA_SSL ENV found with "on"
2018/10/29 04:22:23 [debug] KONG_CASSANDRA_PORT ENV found with "9042"
2018/10/29 04:22:23 [debug] KONG_DATABASE ENV found with "cassandra"
2018/10/29 04:22:23 [debug] KONG_PG_DATABASE ENV found with ""
2018/10/29 04:22:23 [debug] KONG_CASSANDRA_LOCAL_DATACENTER ENV found with "DC1"
2018/10/29 04:22:23 [debug] KONG_CASSANDRA_CONSISTENCY ENV found with "LOCAL_QUORUM"
2018/10/29 04:22:23 [debug] KONG_PG_SSL_VERIFY ENV found with ""
[...]

PR #3957, opened 2018-11-07
feat(migrations) implement post up and teardown functionality in connectors

bungle:
### Summary

This implements a way to run a strategy specific code after `up` and `teardown` migration phases. For Cassandra it also implements `a wait for consensus` on those calls.

It is made according to this comment on issue #3924:
https://github.com/Kong/kong/issues/3924#issuecomment-436343538

### DIFF
--- a/kong/db/init.lua
+++ b/kong/db/init.lua
@@ -570,6 +570,14 @@ do
       end
     end
 
+    if run_teardown then
+      ok, err = self.connector:post_run_teardown_migrations()
+      if not ok then
+        self.connector:close()
+        return nil, prefix_err(self, err)
+      end
+    end
+
     self.connector:close()
 
     return true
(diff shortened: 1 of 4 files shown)


kong/db/init.lua
```diff
@@ -570,6 +570,14 @@ do
       end
     end
 
+    if run_teardown then
+      ok, err = self.connector:post_run_teardown_migrations()
+      if not ok then
+        self.connector:close()
+        return nil, prefix_err(self, err)
+      end
+    end
```
thibaultcha: When both `run_up` and `run_teardown` are given, this will wait for schema consensus twice (and print the verbose logs twice as well). Will update locally.

kong/db/strategies/postgres/connector.lua
```diff
@@ -770,7 +770,12 @@ function _mt:record_migration(subsystem, name, state)
 end
 
 
-function _mt:post_up_migrations()
+function _mt:post_run_up_migrations()
+  return true
+end
+
+
+function _mt:post_run_teardown_migrations()
```
thibaultcha: We should probably not bother with those since the parent class defines these... Will update as well.

thibaultcha:
Merged!

PR #3965, opened 2018-11-09
chore(deps) update dns lib to 3.0.2

Tieske:
Bumps the DNS client and loadbalancer lib to 3.0.2

- Fix (DNS lib + Kong): a potential issue with the healthchecks not being able to set the status of a peer. This happens when there are multiple levels of dns, eg: SRV -> A -> ip-address.
- Fix (DNS lib): the (long time) outstanding fix for the error `attempt to index local 'address' (a nil value)`

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -195,7 +195,7 @@ end
 
 local create_balancer
 do
-  local ring_balancer = require "resty.dns.balancer"
+  local ring_balancer = require "resty.dns.balancer.ring"
 
   local create_healthchecker
   do
@@ -206,10 +206,11 @@ do
     -- or removed to a balancer.
     -- @param balancer the ring balancer object that triggers this callback.
     -- @param action "added" or "removed"
+    -- @param address balancer address object
     -- @param ip string
     -- @param port number
     -- @param hostname string
-    local function ring_balancer_callback(balancer, action, ip, port, hostname)
+    local function ring_balancer_callback(balancer, action, address, ip, port, hostname)
       local healthchecker = healthcheckers[balancer]
       if action == "added" then
         local ok, err = healthchecker:add_target(ip, port, hostname)
@@ -264,21 +265,24 @@ do
       -- The lifetime of the healthchecker is based on that of the balancer.
       healthcheckers[balancer] = hc
 
-      balancer.report_http_status = function(ip, port, status)
+      balancer.report_http_status = function(handle, status)
+        local ip, port = handle.address.ip, handle.address.port
         local _, err = hc:report_http_status(ip, port, status, "passive")
         if err then
           log(ERR, "[healthchecks] failed reporting status: ", err)
         end
       end
 
-      balancer.report_tcp_failure = function(ip, port)
+      balancer.report_tcp_failure = function(handle)
+        local ip, port = handle.address.ip, handle.address.port
         local _, err = hc:report_tcp_failure(ip, port, nil, "passive")
         if err then
           log(ERR, "[healthchecks] failed reporting status: ", err)
         end
       end
 
-      balancer.report_timeout = function(ip, port)
+      balancer.report_timeout = function(handle)
+        local ip, port = handle.address.ip, handle.address.port
         local _, err = hc:report_timeout(ip, port, "passive")
         if err then
           log(ERR, "[healthchecks] failed reporting status: ", err)
@@ -797,16 +801,17 @@ local function execute(target, ctx)
     end
   end
 
-  local ip, port, hostname
+  local ip, port, hostname, handle
   if balancer then
     -- have to invoke the ring-balancer
-    ip, port, hostname = balancer:getPeer(hash_value,
-                                          target.try_count,
-                                          dns_cache_only)
+    ip, port, hostname, handle = balancer:getPeer(dns_cache_only,
+                                          target.balancer_handle,
+                                          hash_value)
     if not ip and port == "No peers are available" then
       return nil, "failure to get a peer from the ring-balancer", 503
     end
     target.hash_value = hash_value
+    target.balancer_handle = handle
 
   else
     -- have to do a regular DNS lookup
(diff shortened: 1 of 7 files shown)


kong/runloop/handler.lua
```diff
@@ -773,13 +774,14 @@ return {
       if ctx.KONG_PROXIED == true then
         -- Report HTTP status for health checks
         if balancer_data and balancer_data.balancer and balancer_data.ip then
```
hishamhm: shouldn't this test now test for `balancer_handle` instead of `ip`?

hishamhm:
A few comments:
* IIRC, the code changes shouldn't be in the `chore` commit, but in a separate one, but I see how this would cause an intermediate broken state — maybe it's a good case for keeping separate commits, but merging them with an explicit merge-commit instead of rebasing (@thibaultcha could confirm)
* is it possible to add a regression test that exercises the scenario that this bump fixes? (I understand it could be difficult due to requiring a specific DNS configuration -- but I guess it could it be mocked?)
* Is the `lua-resty-dns-client` API final? I think I mentioned this before, but it is inconsistent that the descriptor-type handle sometimes is the first argument and sometimes not (in general, descriptor-type arguments are always the first argument)

kong/runloop/handler.lua
Tieske: Either one is ok

thibaultcha:
@hishamhm Either is fine imho when updating the usage of a module to fit the new API. However, if this fixes a bug, a regression test is also very much welcome since it would avoid the bug from slipping away form the changelog, which makes a "multi-commit" PR a better candidate.

Tieske:
@hishamhm  @thibaultcha 

> I understand it could be difficult due to requiring a specific DNS configuration -- but I guess it could it be mocked?

exactly that is the reason. We can't mock a dns record in an integration test. The one thing we could do is set up some DNS test records somewhere and use those in a test. But I don't think we should add that type of dependency (outside the codebase) to the testsuite. Let me know if you want it anyway or if you see other possibilities.

> Is the lua-resty-dns-client API final? I think I mentioned this before, but it is inconsistent that the descriptor-type handle sometimes is the first argument and sometimes not (in general, descriptor-type arguments are always the first argument)

Since we're only using it, we can update and re-tag the lib. The choices I made were because I didn't want to break the API more than necessary. The references are;

- `setPeerStatus(available, handle/address/ip, port)`: since the parameter is combined, it doesn't make sense to change the order
- `handle:release(ignore)`: for completeness, since your comment doesn't apply to this reference
- `ip, port, host, handle = getPeer(cacheOnly, handle, ....)`: here they are kept as they were, and (imho) ordered on importance, which makes sense to me.

hishamhm:
quick update on this: just had an out-of-band discussion with @Tieske on how to mock the DNS resolver library to write a regression test for this. He'll add the test and then we'll proceed to merge this PR.

kong/runloop/balancer.lua
```diff
@@ -264,21 +265,24 @@ do
       -- The lifetime of the healthchecker is based on that of the balancer.
       healthcheckers[balancer] = hc
 
-      balancer.report_http_status = function(ip, port, status)
+      balancer.report_http_status = function(handle, status)
+        local ip, port = handle.address.ip, handle.address.port
```
dliberman: This line is returning an error on every request when I test this change combined with [fix for issue 49](https://github.com/Kong/lua-resty-dns-client/issues/49) over Kong CE 1.0.3.

```
2019/03/08 13:59:40 [error] 5245#0: *166 failed to run log_by_lua*: /usr/local/share/lua/5.1/kong/runloop/balancer.lua:269: attempt to index field 'address' (a nil value)
stack traceback:
        /usr/local/share/lua/5.1/kong/runloop/balancer.lua:269: in function 'report_http_status'
        /usr/local/share/lua/5.1/kong/runloop/handler.lua:952: in function 'after'
        /usr/local/share/lua/5.1/kong/init.lua:772: in function 'log'
        log_by_lua(nginx-kong.conf:125):2: in function <log_by_lua(nginx-kong.conf:125):1> while logging request, client: 192.168.121.1, server: kong, request: "GET /v1/f174909b-40f1-4d71-ae5f-4fa86702b4ab.json HTTP/1.1", upstream: "http://52.95.164.33:80/v1/f174909b-40f1-4d71-ae5f-4fa86702b4ab.json", host: "192.168.121.46:8000"
```

Complete log available here: https://drive.google.com/file/d/1248kdT5u6oK0N5_bxUH1ZaPxu-kmyAC3/view?usp=sharing

For this test I just set up the server as described on my [comment for issue 49](https://github.com/Kong/lua-resty-dns-client/issues/49#issuecomment-463680659), and sent 10 requests (with a 1sec sleep between then), nothing in parallel. There was no error 500, but this also seems a bit weird in the logs.

Tieske: thx for testing, but you we're too early. I rebased this PR, but didn't get to fix the test failures yet. Sorry for non mentioning that.

dliberman: Sure, I understand. Just trying to help, since this fix is a big deal for us here.

Tieske:
PR complete now. A helper for mocking DNS records has been added, which allows to test the 2-level dns healthcheck failure.

Found another minor bug in the dns lib, hence that was bumped to 3.0.2

@hishamhm please review, and would be nice if this makes into the next RC

shenruisi:
I meet this problem at the Kong 1.0.0 CE version. After running a period time. The Kong Gateway throws 
```
stack traceback:
coroutine 0:
        /usr/local/share/lua/5.1/resty/dns/balancer.lua: in function 'getPeer'
        ./kong/runloop/balancer.lua:798: in function 'execute'
        ./kong/runloop/handler.lua:306: in function 'balancer_setup_stage2'
        ./kong/runloop/handler.lua:826: in function 'after'
        ./kong/init.lua:718: in function 'access'
```

And then I tried to update Kong to the latest version. I checked that seems Kong (1.0.3) just depend on lua-restry-dns-client(2.0.0). How can I just update the lua-restry-dns-client to 3.0.2 with Kong (1.0.3)

shenruisi:
Seem the fix dns lib to 3.0.2 missed on the master now?  I check the kong-1.0.3-0.rockspec still 2.2.0 on master.

thibaultcha:
@shenruisi This fix shouldn't have been targeting the master branch, as it includes a major bump in the DNS dependency, which we generally try to avoid. It has been pushed to [next](https://github.com/Kong/kong/blob/next/kong-1.1.0rc2-0.rockspec#L33) instead, and will be included in Kong 1.1.0 in the coming week.

PR #3293, opened 2018-03-13
fix(plugin) reuse redis conn for different db & auth optimize

mengskysama:
### Summary

If user configure response-ratelimiting、response-ratelimiting use same redis instance with `db0` and `other` will get unexpected results.

### Full changelog

* red:select each time.
* red:auth only once.

### Issues resolved

Fix #3292

PR #3973, opened 2018-11-12
fix(plugins) don't call redis:select when not necessary

fffonion:
### Summary

In plugin rate-limiting and response-ratelimiting, a redis:select is
called when connection is reused, even though the current configured
redis database is 0, to clear the previously selected database. This
will be problem for some use case like twemproxy where `select` is
not supported, and sending `select` call will cause the connection to
be closed.

The patch used a host+port+database based redis connection pool, so that
we always know the current connection is using correct database. The 
behaviour of redis:auth only once is not changed.

### Full changelog

* Use database-based redis connection pool to reduce unnecessary `select` call

### Issues resolved

Related to #3293

### DIFF
--- a/kong/plugins/rate-limiting/policies/init.lua
+++ b/kong/plugins/rate-limiting/policies/init.lua
@@ -51,13 +51,7 @@ local get_local_key = function(conf, identifier, period_date, name)
 end
 
 
-local function get_redis_pool_name(conf)
-  -- return a special pool name only if redis_database is set to non-zero
-  -- otherwise use the default pool name host:port
-  if conf.redis_database ~= 0 then 
-    local key = fmt("%s:%d:%d", conf.redis_port, conf.redis_port, conf.redis_database)
-  end
-end
+local sock_opts = {}
 
 
 local EXPIRATIONS = {
@@ -145,8 +139,13 @@ return {
     increment = function(conf, limits, identifier, current_timestamp, value)
       local red = redis:new()
       red:set_timeout(conf.redis_timeout)
+      -- use a special pool name only if redis_database is set to non-zero
+      -- otherwise use the default pool name host:port
+      sock_opts.pool = conf.redis_database and
+                       conf.redis_host .. ":" .. conf.redis_port .. 
+                       ":" .. conf.redis_database
       local ok, err = red:connect(conf.redis_host, conf.redis_port,
-                                  { pool = get_redis_pool_name(conf) })
+                                  sock_opts)
       if not ok then
         ngx_log(ngx.ERR, "failed to connect to Redis: ", err)
         return nil, err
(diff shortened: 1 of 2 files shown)


kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -51,6 +51,15 @@ local get_local_key = function(conf, identifier, period_date, name)
 end
 
 
+local function get_redis_pool_name(conf)
+  -- return a special pool name only if redis_database is set to non-zero
+  -- otherwise use the default pool name host:port
+  if conf.redis_database ~= 0 then 
+    local key = fmt("%s:%d:%d", conf.redis_port, conf.redis_port, conf.redis_database)
```
thibaultcha: The second argument here should be `conf.redis_host`

kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -136,7 +145,8 @@ return {
     increment = function(conf, limits, identifier, current_timestamp, value)
       local red = redis:new()
       red:set_timeout(conf.redis_timeout)
-      local ok, err = red:connect(conf.redis_host, conf.redis_port)
+      local ok, err = red:connect(conf.redis_host, conf.redis_port,
+                                  { pool = get_redis_pool_name(conf) })
```
thibaultcha: We could reuse this table and just override its `pool` property, to save some GC overhead, and some function call overhead as well:
```lua
sock_opts.pool = get_redis_pool_name(conf) -- could even be inlined
```

kong/plugins/rate-limiting/policies/init.lua
thibaultcha: Additionally, prior art has proven that a simple concat operation could perform as well or better than `string.format`, we should maybe follow this new rule of thumb.

thibaultcha:
Nice catch :) Meant to request some changes, but mis-clicked. Additionally, note that the linter is failing in CI.

fffonion:
@thibaultcha Thank you for reviewing this, I've addressed the comments and CI is green now. Could you take another look when you got chance?

kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -148,27 +157,24 @@ return {
...
-        -- can skip this roundtrip.
+        if conf.redis_database ~= 0 then
+          -- Only call select first time, since we know the connection is shared
+          -- between instances that use the same redis database
 
-        local ok, err = red:select(conf.redis_database or 0)
-        if not ok then
-          ngx_log(ngx.ERR, "failed to change Redis database: ", err)
-          return nil, err
+          local ok, err = red:select(conf.redis_database or 0)
```
p0pr0ck5: probably dont need the `or 0` here?

thibaultcha:
LGTM, however we should wait until our Travis CI issues are resolved before merging this.

thibaultcha:
@fffonion It should be fine if you rebase this branch on the latest master.

hishamhm:
Running green on `master`!

PR #3976, opened 2018-11-13
refactor(plugins) make plugins defensive against possible errors between different phases

bungle:
### Summary

This PR consists of several fixes and commits:

**The plugin changes**

1. chore(rate-limiting) make plugin defensive against possible errors and nil header values
2. chore(response-transformer) make plugin defensive against possible errors and nil header values
3. chore(response-ratelimiting) make plugin defensive against possible errors and nil header values
4. chore(request-transformer) make plugin defensive against possible errors and nil header values

Relates to: #3924


hishamhm:
@bungle changes LGTM! how complicated would it be to cause those error situations in a test so that we could we could regression-test this to ensure the plugins will remain defensive in the future?

hishamhm:
Closing this one as the fixes for the plugin issues this fixes were split into #4006 (in rc3), #4038 and #4039

PR #3984, opened 2018-11-15
feat(schema) remove default port from services scheme

james-callahan:
The default of 80 was unsuitable for HTTPS and even more unsuitable for
stream support.

This cause minimal breakage as a default is injected via
`api_helpers.resolve_url_params` when using the /services url argument.

Note: the router also has a default for the port if the protocol is absent.

### DIFF
--- a/kong/db/schema/entities/services.lua
+++ b/kong/db/schema/entities/services.lua
@@ -15,7 +15,7 @@ return {
     -- { tags          = { type = "array", array = { type = "string" } }, },
     { protocol        = typedefs.protocol { required = true, default = "http" } },
     { host            = typedefs.host { required = true } },
-    { port            = typedefs.port { required = true, default = 80 }, },
+    { port            = typedefs.port { required = true } },
     { path            = typedefs.path },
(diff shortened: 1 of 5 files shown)


hishamhm:
> The default of 80 was unsuitable for HTTPS and even more unsuitable for
stream support.

Agreed. Still, the default value for `protocol` is still `"http"`. I understand that these defaults are unsuitable for HTTPS and stream, and that things will continue working if users are using the `url` shortcut. However, if users are setting just `host` and expecting HTTP proxying to work, then we're breaking people's working setups for little gain (stream users will still have to set `port` with this default set or not, and they will still get a broken setup if they don't).

I'd like to hear more opinions, but this doesn't sound like a must-have change.

james-callahan:
> I'd like to hear more opinions, but this doesn't sound like a must-have change.

The problem is that when you post `url` we default to `80` for http or `443` for `https`. With tcp and tls we have no sensible default to use in the url parser: so instead it defaults to the `80` from the default in the schema.

PR #3986, opened 2018-11-15
feat(schema) add new conditional_at_least_one_of entity checker

james-callahan:
This code was originally written by @thibaultcha, after some testing with @bungle I found that updating (e.g. via `PATCH`) a record with this entity checker results in unexpected behaviour.
I split it out of its original PR so we can review separately and solve the issue.
The final commit of this branch currently contains a failing test, please let me know if the issue is with the test or the entity checker.


spec/01-unit/000-new-dao/01-schema/01-schema_spec.lua
```diff
@@ -1272,6 +1272,19 @@ describe("schema", function()
...
+
+          -- Can update to whole valid record
+          assert.truthy(Test:validate_update({ a = 123, b = "foo" }))
+
+          -- Empty update works
+          assert.truthy(Test:validate_update({ }))
+
+          -- Can update if_field without respecifying other fields
+          assert.truthy(Test:validate_update({ a = 123 }))
+        end)
```
thibaultcha: 1. No `a` defined, so the entity checker does not run: valid
2. `a` defined and `>= 0`, entity checker runs and succeeds: valid
3. No `a` defined, entity checker does not run: valid
4. `a` defined and `>= 0`, entity checker runs and fails correctly (missing required fields if `a` is specified): invalid

From the schema library's point of view, there is nothing wrong with the above.

Performing validations on updates (and implementing a read-before-write + merged values) is the responsibility of the DAO, and we might want to add some entity checkers as conditions for a read-before-write trigger, this way we avoid the scenario described in the comment for 4. (cc @hishamhm): https://github.com/Kong/kong/blob/next/kong/db/schema/init.lua#L1346-L1357

hishamhm: > we might want to add some entity checkers as conditions for a read-before-write trigger

Yes, this would be a welcome addition in the future, but we've been living with this kind of over-restriction in entity checkers for a while now. I say we merge `conditional_at_least_one_of` as is (inverting the assert in L1285, and adding the "when updating " caveat to the error string in `validate_update" as we do for a couple other errors), as it contains the necessary features for the incoming PRs, and we improve the read-before-write detection later.

bungle: The thing is, with `POST` and `PUT` we get full entity. With `PATCH` we get partial thing. We can kinda expect that the database row is `OK` when `PATCHING`. Of course it could lead to thing that you make two `PATCH` request and set fields to `NULL` or something, and ending up with `NON-VALID` entity. That is active work, and you are really shooting to foot there (or maybe you do that just to temporary disable something). But making `PATCH` to require you to send old values just to please validator is in my opinion worse.

bungle: read-before-write has the possibility to race condition if we care about that, though it is quite rare.

hishamhm: > I think we should pay the read-before-write cost to avoid an invalid database state, anything else will result in tricky support issues!

If I'm following the discussion correctly, failing that fourth `validate_update` case will cause benign PATCHes (that wouldn't put the DB in an invalid state) to be rejected, but will not allow malign PATCHes to pass.

As far as I can tell, the current behavior is over-restrictive, not over-lenient. If it's over-restrictive, we can fix it later.

hishamhm: Yeah, I think it's good to go, I made that rebased branch with the test added with tgr one last assert reverted and the minor "when updating" tweak to the error message. Will merge it tomorrow, thanks!

hishamhm:
Closed in favor of #3994.

PR #3991, opened 2018-11-16
tests(*) lazy-setup and teardown

Tieske:
The default `setup()` and `teardown()` always run, even if there are no tests to be executed within the describe block. The lazy-ones (`lazy_setup()` and `lazy_teardown()`) only execute when there is actually a test to run.

Effectively this means that tests that run migrations and are executed with the `#postgres` tag, would still also run the migrations on C*.

This should improve the test-suite execution time.

### DIFF
--- a/spec-old-api/01-unit/010-router_spec.lua
+++ b/spec-old-api/01-unit/010-router_spec.lua
@@ -617,14 +617,14 @@ describe("Router", function()
       end)
 
       describe("root / [uri]", function()
-        setup(function()
+        lazy_setup(function()
           table.insert(use_case, 1, {
             name = "api-root-uri",
             uris = {"/"},
           })
         end)
 
-        teardown(function()
+        lazy_teardown(function()
           table.remove(use_case, 1)
         end)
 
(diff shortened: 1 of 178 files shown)


thibaultcha:
One thing: while `spec-old` is still in this repository, we should also maintain it, and avoid having it divert from `spec`. Would you update the initializers in `spec-old` as well? Thanks.

kidd:
busted has the flag `--lazy` that overwrites setup and teardown to be lazy. 

If the plan is to use it everywhere it might make sense to use it.

heads up: making all setups lazy might tests when run individually (that rely on previous setup phases of other 'contexts').

Tieske:
imo the default should be lazy, but I also like the explicitness.

Tieske:
@thibaultcha updated to include the old tests

hishamhm:
> heads up: making all setups lazy might tests when run individually (that rely on previous setup phases of other 'contexts').

I thought of that, but I think if it happens, then this exposes lack of atomicity/incorrect dependencies in our tests (they should be fixed if it's the case — either by making them properly atomic or, at the very least, making the specific ones non-lazy, the explicit option gives us this granularity). But I hope in most cases the dependency relations are properly captured by the nesting of `describe` blocks.

So, LGTM!

PR #3994, opened 2018-11-19
feat: conditional_at_least_one_of

hishamhm:
This is a variant of #3986 rebased on top of current `next`,  with the last assert for the error condition inverted, and adding the "when updating " caveat to the error string in `validate_update` as we do for a couple other errors.


james-callahan:
Merged in #4009

PR #3996, opened 2018-11-19
perf(key-auth) unroll header sets to improve JITability

p0pr0ck5:
### Summary

The PDK's 'set_headers' function relies on NYI LuaJIT operations that result in the JIT trace being aborted, forcing OpenResty to fall back to interpreted code. This commit unrolls the header
definitions into individual PDK 'set_header' calls, improving JITability of this chunk.

In lab tests we observed that Kong's maximum throughput (in a CPU-bound workload) increased 10% with this change. Backtrace flamegraphs of non-JITed captures show that this change results in non-JITed key-auth handler code reduced from 32% of traces to 1% of traces:

rc2-nojit
key-auth-jit-nojit

### Full changelog

* Replace PDK `set_headers` call in `key-auth` with appropriate `set_header`.

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -41,23 +41,22 @@ end
 local function set_consumer(consumer, credential)
   local const = constants.HEADERS
 
-  local new_headers = {
-    [const.CONSUMER_ID]        = consumer.id,
-    [const.CONSUMER_CUSTOM_ID] = tostring(consumer.custom_id),
-    [const.CONSUMER_USERNAME]  = consumer.username,
-  }
-
   kong.client.authenticate(consumer, credential)
 
+  kong.service.request.set_header(const.CONSUMER_ID, consumer.id)
+  kong.service.request.set_header(const.CONSUMER_CUSTOM_ID, tostring(consumer.custom_id))
+  kong.service.request.set_header(const.CONSUMER_USERNAME, consumer.username)
+
   if credential then
-    new_headers[const.CREDENTIAL_USERNAME] = credential.username
+    if credential.username then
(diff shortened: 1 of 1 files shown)


kong/plugins/key-auth/handler.lua
```diff
@@ -41,23 +41,22 @@ end
...
-    [const.CONSUMER_ID]        = consumer.id,
-    [const.CONSUMER_CUSTOM_ID] = tostring(consumer.custom_id),
-    [const.CONSUMER_USERNAME]  = consumer.username,
-  }
-
   kong.client.authenticate(consumer, credential)
 
+  kong.service.request.set_header(const.CONSUMER_ID, consumer.id)
+  kong.service.request.set_header(const.CONSUMER_CUSTOM_ID, tostring(consumer.custom_id))
+  kong.service.request.set_header(const.CONSUMER_USERNAME, consumer.username)
```
james-callahan: `username is optional`, you will need a guard here.

bungle: There is no test with `consumer` with username, and the `tostring` there is I think wong, it should be something like:

```lua
if consumer.custom_id then
  kong.service.request.set_header(const.CONSUMER_CUSTOM_ID, consumer.custom_id)
else
  kong.service.request.clear_header(const.CONSUMER_CUSTOM_ID, consumer.custom_id)
end
```

The issue is that `set_header` in `pdk` does not allow `nil`, as did the `OpenResty`'s one. I am also making changes here:
https://github.com/Kong/kong/pull/3976

I can take your changes and port them there too, is that ok?

p0pr0ck5: @bungle yeah that’s be great, thanks. It looks like your work is already covering this anyhow. Thanks!

PR #3999, opened 2018-11-21
fix delete target can not offline old target issue

zjkyz8:
Currently, if we delete a target from upstream by Admin API, the old target is still there, and new target with same upstream id and target and 0 weight will be created, this will cause the target still server the request. So this PR use update to set weight to 0, make the target server offline, and keeping history there.

### DIFF
--- a/kong/api/routes/upstreams.lua
+++ b/kong/api/routes/upstreams.lua
@@ -263,11 +263,11 @@ return {
       clean_history(self.upstream.id, dao_factory)
 
       -- this is just a wrapper around POSTing a new target with weight=0
-      local _, err = dao_factory.targets:insert({
+      local _, err = dao_factory.targets:update({
         target      = self.target.target,
         upstream_id = self.upstream.id,
         weight      = 0,
-      })
+      }, { id = self.target.id})
       if err then
         return app_helpers.yield_error(err)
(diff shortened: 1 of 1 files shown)


thibaultcha:
> Currently, if we delete a target from upstream by Admin API, the old target is still there, and new target with same upstream id and target and 0 weight will be created

That is the intended behavior. As reflected by the failing test in the CI build (`Admin API /upstreams/{upstream}/targets/{target} DELETE acts as a sugar method to POST a target with 0 weight (by target)`).

> this will cause the target still server the request

Can you provide a minimal reproducible test case for this behavior you are describing? Thanks.

zjkyz8:
@thibaultcha, thank you very much to reply, below is steps:
image
At the beginning, I have one target which belows to upstream 493ff8e1-26f7-4e68-a729-c89bb6bc20b5.

Then call DELETE API by some REST client, I use Chrome extension REST client, like below:
image

Current behavior like below, two targets appear.
image

Excepted behavior, only one target with 0 weight should be there.

thibaultcha:
The behavior you are seeing is the expected one. See https://docs.konghq.com/0.14.x/admin-api/#delete-target

I will thus be closing this now.

PR #4012, opened 2018-11-22
feat(core) add simulate_access_after function

kidd:
Plugins that generate content need to set latency headers
methods. This function centralizes these cleanups so that plugins can
call it.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -53,6 +53,22 @@ local function get_now()
 end
 
 
+-- Plugins that generate content need to set latency headers. This
+-- function centralizes these cleanups so that plugins can call it.
+local function simulate_access_after(ctx)
+  local now = get_now()
+
+  ctx.KONG_ACCESS_TIME = now - ctx.KONG_ACCESS_START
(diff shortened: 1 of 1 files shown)


thibaultcha:
That doesn’t seem right... can you elaborate on the use-case here? An example usage maybe?

thibaultcha:
Ok, I supposed plugins that gracefully terminate the request from signing Kong itself? E.g. authentication rejected, rate-limiting exceeded, request-termination, etc...

Unfortunately plugins should not be responsible for calling this, which is prompt to plugins authors errors and just very unfriendly. Additionally, proxy latency should not be set (since not proxying happened), and the ctx variable will confuse the kong.response.get_source function, which is a no-no. And lastly the code duplication makes things worse (we would at the very least need some tests). 

Instead, those metrics should be exposed via a module (ideally a PDK module), which is on the PDK roadmap. Such a module would return sane defaults and be the source of truth for both the core and plugins. Introducing this work around sets us back and increases our tech debt, while prolongating it because of the 1.0 milestone which would force us to support this function in the long run.

kidd:
Ok, I understand.

Closing this one.

PR #4019, opened 2018-11-26
tests(*) speed up test suite, reduce flakiness

hishamhm:
I spent some time catching some low-hanging fruit to speed up our test suite and reduce some of the flakiness.

It seems to reduce "total time" from around 45 to 37 min, which translate in clock time to tests going down from 13-14 (best case, no flaky restarts) to something like 11-12 mins. After this is merged and the `spec-old-api` tests are finally removed in 1.0, I hope we can get to a sub-10 minute test suite (at least before we add more tests!).

### DIFF
--- a/spec/02-integration/02-cmd/02-start_stop_spec.lua
+++ b/spec/02-integration/02-cmd/02-start_stop_spec.lua
@@ -2,7 +2,7 @@ local helpers = require "spec.helpers"
 
 describe("kong start/stop", function()
   lazy_setup(function()
-    helpers.get_db_utils() -- runs migrations
+    helpers.get_db_utils(nil, {}) -- runs migrations
     helpers.prepare_prefix()
   end)
   after_each(function()
--- a/spec/03-plugins/24-rate-limiting/02-policies_spec.lua
+++ b/spec/03-plugins/24-rate-limiting/02-policies_spec.lua
@@ -10,12 +10,11 @@ for _, strategy in helpers.each_strategy() do
       local conf       = { route = { id = uuid() }, service = { id = uuid() } }
 
       local db
-      local dao
       local policies
 
       lazy_setup(function()
         local _
-        _, db, dao = helpers.get_db_utils(strategy)
+        _, db = helpers.get_db_utils(strategy, {})
 
         if _G.kong then
           _G.kong.db = db
@@ -27,9 +26,8 @@ for _, strategy in helpers.each_strategy() do
         policies = require "kong.plugins.rate-limiting.policies"
       end)
 
-      after_each(function()
-        assert(db:truncate())
-        dao:truncate_tables()
+      before_each(function()
+        assert(db:truncate("ratelimiting_metrics"))
       end)
 
       it("returns 0 when rate-limiting metrics don't exist yet", function()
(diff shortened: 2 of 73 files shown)


spec/fixtures/admin_api.lua
```diff
@@ -0,0 +1,73 @@
...
+  local res, err = api_client:send({
+    method = method,
+    path = path,
+    headers = {
+      ["Content-Type"] = "application/json"
+    },
+    body = body,
+  })
+  if not res then
+    return nil, err
```
thibaultcha: Ideally we'd close the client before returning here.

spec/fixtures/admin_api.lua
```diff
@@ -0,0 +1,73 @@
...
+  end
+  local resbody = res.status ~= 204 and res:read_body()
+  api_client:close()
+  if res.status == 204 then
+    return nil
+  elseif res.status < 300 then
+    return cjson.decode(resbody)
+  else
+    return nil, "Error " .. tostring(res.status) .. ": " .. resbody
+  end
```
thibaultcha: Suggestion to simplify the branching and reduce the `204` comparisons:
```lua
if res.status == 204 then
  return nil
end

local resbody = res:read_body()
if res.status < 300 then
  return cjson.decode(resbody)
end

return nil, "Error " .. tostring(res.status) .. ": " .. resbody
```

spec/fixtures/admin_api.lua
```diff
@@ -0,0 +1,73 @@
...
+  "plugins",
+  "routes",
+  "services",
+  "jwt_secrets",
+  "oauth2_credentials",
+  "oauth2_tokens",
+  "oauth2_authorization_codes",
+  "keyauth_credentials",
+  "hmacauth_credentials",
+}
```
thibaultcha: This duplication could become annoying in the long run... It seems like a DAO `pairs()` could produce the same `admin_api_as_db` without duplicating the entities slugs?

thibaultcha:
Some commits mention plugins tests, but other plugins speed ups are in their own commits (e.g. `tests(oauth2)`). Some rebasing is needed maybe? E.g. spec helpers, integrations, plugins (more or les)?

spec/03-plugins/24-rate-limiting/02-policies_spec.lua
```diff
@@ -10,12 +10,11 @@ for _, strategy in helpers.each_strategy() do
       local conf       = { route = { id = uuid() }, service = { id = uuid() } }
 
       local db
-      local dao
       local policies
 
       lazy_setup(function()
         local _
-        _, db, dao = helpers.get_db_utils(strategy)
+        _, db = helpers.get_db_utils(strategy, {})
```
kikito: What is the reasoning behind this change? It looks superfluous.

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -2,7 +2,7 @@ local helpers = require "spec.helpers"
 
 describe("kong start/stop", function()
   lazy_setup(function()
-    helpers.get_db_utils() -- runs migrations
+    helpers.get_db_utils(nil, {}) -- runs migrations
```
kikito: same question as before - is this change needed?

spec/03-plugins/24-rate-limiting/02-policies_spec.lua
hishamhm: this means "empty list of truncations" to run, as opposed to the default, which is truncate all tables.

PR #4023, opened 2018-11-27
feat(rate-limiting) global rate-limiting to protect back-end

julienlau:
This minor feature enable to perform global rate-limiting over a given endpoint aggregating count globally on the whole kong instance in case you want to protect the back-end to exceed its limits.

### Summary

Current rate-limiting limit_by modes are: consumer, credential, ip.
Default is ip.
A new mode, is implemented and named "global".
It was tested and validated only using policy=local, but should also work on cluster or REDIS as soon as nginx server ip differs between knog/nginx instances.

### Full changelog

I did not add new tests. Existing luacheck tests are OK.

### Issues resolved

It is linked to the following forum thread: 
https://discuss.konghq.com/t/common-rate-limiting-for-all-consumers/2218/5
https://discuss.konghq.com/t/rate-limiting-behaviour/2081/2

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -29,10 +29,12 @@ local function get_identifier(conf)
     end
   elseif conf.limit_by == "credential" then
     identifier = ngx.ctx.authenticated_credential and ngx.ctx.authenticated_credential.id
+  elseif conf.limit_by == "ip" then
+    identifier = ngx.var.remote_addr
   end
 
   if not identifier then
-    identifier = ngx.var.remote_addr
+    identifier = ngx.var.server_addr .. "-" .. ngx.var.server_name .. ngx.var.pid .. "-" .."-global"
   end
 
   return identifier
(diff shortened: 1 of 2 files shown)


kong/plugins/rate-limiting/handler.lua
```diff
@@ -29,10 +29,12 @@ local function get_identifier(conf)
     end
   elseif conf.limit_by == "credential" then
     identifier = ngx.ctx.authenticated_credential and ngx.ctx.authenticated_credential.id
+  elseif conf.limit_by == "ip" then
+    identifier = ngx.var.remote_addr
   end
 
   if not identifier then
-    identifier = ngx.var.remote_addr
+    identifier = ngx.var.server_addr .. "-" .. ngx.var.server_name .. ngx.var.pid .. "-" .."-global"
```
bungle: Is this really a per worker ratelimit as there is `ngx.var.pid` (provided that no worker in cluster ever uses a same process id, which you have no gurantee)? Or do I read this incorrectly?

julienlau: Hi,
Thank you for the correction, my will is to have a limit per "kong" instance. It makes sense to simplify the identifier like that as long as this value is never empty or valuated at 127.0.0.1 ?
I am not familiar with nginx / kong internals.

hbagdi:

PR #4029, opened 2018-11-28
fix(cors) follow mozilla guidelines for preflight request

### LINKED ISSUE #4008: CORS Preflight OPTIONS request should return 200 instead of 204
### Summary

CORS Preflight OPTIONS request should return 200 instead of 204
Actual code here
https://github.com/Kong/kong/blob/master/kong/plugins/cors/handler.lua#L143

### Steps To Reproduce
It seems it will only block the GET request. If you set your own header in a GET request, chrome will send a preflight OPTIONS first and get 204 response.
The browser will not continue to send the actual GET request since it's NO_CONTENT.

### Possible fix
instead of returning 204, just return 200 with Content-Length header set to 0.
Instructed by this guideline. 
https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

aslafy-z:
### Summary

Follow the mozilla guidelines for preflight request (https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) for the cors plugin implementation.

### Full changelog

* On OPTIONS requests
	* Send 200 instead of 204
	* Set `Content-Length` to `0`

### Issues resolved

Fixes #4008

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -8,7 +8,7 @@ local tostring = tostring
 local ipairs   = ipairs
 
 
-local NO_CONTENT = 204
+local HTTP_OK = 200
 
 
 local CorsHandler = BasePlugin:extend()
@@ -137,7 +137,7 @@ function CorsHandler:access(conf)
     set_header("Access-Control-Max-Age", tostring(conf.max_age))
   end
 
-  return kong.response.exit(NO_CONTENT)
+  return kong.response.exit(HTTP_OK)
 end
 
 
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -137,6 +137,8 @@ function CorsHandler:access(conf)
     set_header("Access-Control-Max-Age", tostring(conf.max_age))
   end
 
+  set_header("Content-Length", "0")
+
   return kong.response.exit(HTTP_OK)
 end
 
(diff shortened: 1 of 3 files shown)


kong/plugins/cors/handler.lua
```diff
@@ -137,7 +137,9 @@ function CorsHandler:access(conf)
     set_header("Access-Control-Max-Age", tostring(conf.max_age))
   end
 
-  return kong.response.exit(NO_CONTENT)
+  set_header("Content-Length", "0")
```
thibaultcha: `kong.response.exit` without a body will already set this header appropriately, we can remove this.

aslafy-z: applied with https://github.com/Kong/kong/pull/4029/commits/58949d2d24c38ea79a588a43327102d3a2ed8624

thibaultcha:
Thank you! Don't forget to claim your [Contributor T-shirt](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt) :)

dzmitry-lahoda:
I do not get it. Link in description says 204 is valid response

image

but merged code uses 200

i do not get it ...

thibaultcha:
The page was updated since then. See its contents on Sept 30th, 2018: https://web.archive.org/web/20180930031917/https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

PR #4030, opened 2018-11-28
Template bugfix + healthcheck test for stream router

hishamhm:
A basic test for healthchecks using the stream router, plus a fix to a bug that was uncovered when writing it.

### DIFF
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -893,6 +893,15 @@ local function load(path, custom_conf)
   end
 
   do
+    -- is ssl_preread compiled in OpenResty?
+    conf.ssl_preread_enabled = false
+    local nginx_configuration = ngx.config.nginx_configure()
+    if nginx_configuration:find("with-stream_ssl_preread_module", 1, true) then
+      conf.ssl_preread_enabled = true
+    end
+  end
+
+  do
     -- load headers configuration
     local enabled_headers = {}
 
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -291,11 +291,26 @@ do
       if not healthcheck then
         healthcheck = require("resty.healthcheck") -- delayed initialization
       end
+
+      -- Do not run active healthchecks in `stream` module
+      local checks = upstream.healthchecks
+      if ngx.config.subsystem == "stream"
+         and (checks.active.healthy.interval ~= 0
+              or checks.active.unhealthy.interval ~= 0)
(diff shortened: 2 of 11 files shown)


kong/conf_loader.lua
```diff
@@ -892,6 +892,15 @@ local function load(path, custom_conf)
...
   end
 
+  do
+    -- is ssl_preread compiled in OpenResty?
+    conf.ssl_preread_enabled = false
+    local nginx_configuration = ngx.config.nginx_configure()
+    if nginx_configuration:find("with-stream_ssl_preread_module", 1, true) then
+      conf.ssl_preread_enabled = true
+    end
+  end
```
thibaultcha: :clap:

kong/runloop/balancer.lua
```diff
@@ -291,11 +291,21 @@ do
       if not healthcheck then
         healthcheck = require("resty.healthcheck") -- delayed initialization
       end
+
+      -- Do not run active healthchecks in `stream` module
```
thibaultcha: Would you mind elaborating this comment for future selves?

kong/runloop/balancer.lua
```diff
@@ -291,11 +291,21 @@ do
...
+      local checks = upstream.healthchecks
+      if ngx.config.subsystem == "stream" then
+        checks = pl_tablex.deepcopy(checks)
+        checks.active.healthy.interval = 0
+        checks.active.unhealthy.interval = 0
+      end
```
thibaultcha: Note: If we ever allow the stream module to be enabled without enabling the HTTP module (we currently always enable the HTTP module), then we would need to allow the inverse of this logic.

spec/02-integration/05-proxy/00-proxy_spec.lua
```diff
@@ -74,3 +75,43 @@ describe("Proxy interface listeners", function()
...
+      stream_listen = "off",
+    }))
+    assert.equals(1, count_server_blocks(helpers.test_conf.nginx_kong_stream_conf))
+    assert.is_nil(get_listeners(helpers.test_conf.nginx_kong_stream_conf).kong)
+  end)
+
+  it("multiple", function()
+    assert(helpers.start_kong({
+      stream_listen = "127.0.0.1:9011, 127.0.0.1:9012",
+      plugins = "bundled",
```
thibaultcha: Maybe this option can be removed (not needed for the test, and is also the default).

PR #3481, opened 2018-05-23
Fix/lambda http2

rainest:
### Summary

Modifies the AWS Lambda plugin to fix issues with HTTP/2 clients

### Full changelog

* Replace `ngx.say` with `ngx.print` for body output, as the trailing newline added by `say` causes a mismatch between the `Content-Length` reported by Lambda and the actual output sent by Kong.
* Strip `Connection` header for HTTP/2 clients per RFC 7540 8.1.2.2

WIP/review requested as I'm unsure how to test this. I did not see any current HTTP/2 client tests, and the resty-http library used for most client integration tests doesn't support HTTP/2 yet (there is an [alternate Lua HTTP client library](https://github.com/daurnimator/lua-http/) that does).

PR #4032, opened 2018-11-28
fix(aws-lambda) strip Connection header if client is using HTTP/2

### LINKED ISSUE #2991: Issue with invoking Lambda using plugin.
Hello 
I have a kong setup in aws with ELB. (https). 
I tried to use the lambda plugin.
when I change the loadbalancer to http i do not see any issue. But with https I get 
"curl: (92) HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)"
Can I get some help or pointers on how to fix this. i tried using the http port from elb to kong instance as well. It does not work
How can i have https enabled 

My Setup
3 kong ec2 boxes 0.11.0
One Application loadbalancer with https listner sending traffic to port 8443 in each box.
I also tried sending traffic to 8000(http). 
But i get the same error with curl. i tried to call from postman and request just timesout.
with firefox or chrome, i see the response but the request does not get disconnected.
if i add --http1.1 option to curl it works.

Thanks

rainest:
### Summary
Strip Connection headers from AWS Lambda responses if client is using HTTP per [RFC 7540](https://tools.ietf.org/html/rfc7540#section-8.1.2.2)

### Issues resolved

Fix #2991

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -236,9 +236,7 @@ function AWSLambdaHandler:access(conf)
     headers["Keep-Alive"] = ""
     headers["Proxy-Connection"] = ""
     headers["Upgrade"] = ""
-    if headers["Transfer-Encoding"] ~= "trailers" then
-      headers["Transfer-Encoding"] = ""
-    end
+    headers["Transfer-Encoding"] = ""
   end
 
   local ok, err = client:set_keepalive(conf.keepalive)
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -233,11 +233,11 @@ function AWSLambdaHandler:access(conf)
 
   if var.http2 then
     headers["Connection"] = ""
-    response_headers["Keep-Alive"] = ""
-    response_headers["Proxy-Connection"] = ""
-    response_headers["Upgrade"] = ""
-    if response_headers["Transfer-Encoding"] ~= "trailers" then
-      response_headers["Transfer-Encoding"] = ""
+    headers["Keep-Alive"] = ""
+    headers["Proxy-Connection"] = ""
+    headers["Upgrade"] = ""
+    if headers["Transfer-Encoding"] ~= "trailers" then
+      headers["Transfer-Encoding"] = ""
     end
   end
 
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -129,6 +129,7 @@ function AWSLambdaHandler:access(conf)
   AWSLambdaHandler.super.access(self)
 
   local upstream_body = new_tab(0, 6)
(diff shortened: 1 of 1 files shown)


rainest:
Previously https://github.com/Kong/kong/pull/3481, which did not like my attempt to rebase. The current version of the plugin no longer uses `ngx.say`, so only the change to strip Connection remains relevant. 

At present we lack an HTTP/2 client library in the test suite, so this does not include new tests.

thibaultcha:
Thanks @rainest. Before I check myself, have you researched (i.e. `grep` or such) the codebase for potential other instances of this issue?

thibaultcha:
@rainest Thanks for tracking them down. It indeed looks like Azure would be guilty of this as well (https://github.com/Kong/kong-plugin-azure-functions/blob/master/kong/plugins/azure-functions/handler.lua#L118). Will you be able to submit a fix for it as well?

wbcustc:
May worth directly create the PR to [next](https://github.com/Kong/kong/blob/next/kong/plugins/aws-lambda/handler.lua) branch. The plugin structure is largely refactored in the [next](https://github.com/Kong/kong/blob/next/kong/plugins/aws-lambda/handler.lua) branch.

thibaultcha:
No, bug fixes are to be submitted against the `master` branch. The fix is identical in the next branch version of the plugin anyways.

kong/plugins/aws-lambda/handler.lua
```diff
@@ -230,6 +231,16 @@ function AWSLambdaHandler:access(conf)
   local content = res:read_body()
   local headers = res.headers
 
+  if var.http2 then
+    headers["Connection"] = ""
+    response_headers["Keep-Alive"] = ""
```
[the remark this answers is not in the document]
rainest: Fixed. Apologies, bad copy-paste from the Azure handler, which uses a different name.

kong/plugins/aws-lambda/handler.lua
```diff
@@ -129,6 +129,7 @@ function AWSLambdaHandler:access(conf)
   AWSLambdaHandler.super.access(self)
 
   local upstream_body = new_tab(0, 6)
+  local var = ngx.var
```
thibaultcha: Mind removing the below `local var = ngx.var` line that now shadows this definition?

kong/plugins/aws-lambda/handler.lua
```diff
@@ -230,6 +231,14 @@ function AWSLambdaHandler:access(conf)
   local content = res:read_body()
   local headers = res.headers
 
+  if var.http2 then
+    headers["Connection"] = nil
+    headers["Keep-Alive"] = nil
+    headers["Proxy-Connection"] = nil
+    headers["Upgrade"] = nil
+    headers["Transfer-Encoding"] = nil
+  end
```
thibaultcha: Note for future reference: https://tools.ietf.org/html/rfc7540#section-8.1.2.2

PR #4038, opened 2018-11-30
fix(plugins) defend transformer plugins against possible errors and nil values

hishamhm:
Fixes by @bungle for request-transformer and response-transformer, defending them against possible errors and nil values. Cherry-picked from #3976.


kikito:
I approved this in order to accelerate merging, but have one question: the `spec-old-api` specs have been removed. Does that belong here? I thought API-related code would be removed on its own PR.

PR #4039, opened 2018-11-30
fix(pdk) allow kong.response.exit on header_filter with no body

kikito:
This fix on kong.response.exit fixes unintended failures on the response-ratelimiting plugin which were introduced when it was translated to the PDK.

I temporarily activated the response-ratelimiting specs on travis and they worked (although I had to to re-launch them many times, they are still quite flaky).

This PR also includes some doc comments on the pdk_phase_checking function, please verify that those comments are correct.

### DIFF
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -239,12 +239,12 @@ local function new(self, major_version)
       return "error"
     end
 
-    if ctx.KONG_PROXIED then
-      return "service"
-    end
-
     if ctx.KONG_EXITED then
       return "exit"
+    end
+
+    if ctx.KONG_PROXIED then
+      return "service"
     end
 
     return "error"


t/Util.pm
```diff
@@ -34,6 +34,33 @@ our $HttpConfig = <<_EOC_;
...
+        -- checking deactivated, and the second time with phase checking activated.
+        -- Params:
+        -- * phase: the phase we want to test, i.e. "access"
+        -- * skip_fnlist controls a check: by default, this method
+        --         will check that the provided list of methods is "complete" - that
+        --         all the methods inside the `mod` (see below) are covered. Setting
+        --         `skip_fnlist` to `true` will skip that test (so the `mod` can have
+        --         methods that go untested)
+        --
+        -- This method also reads from 2 globals, which could be params instead:
```
hishamhm: These are not parameters because they are defined in the test's `init` block and read by other blocks, and those only share globals. Since communication is happening via globals anyway, passing the parameter there would be just boilerplate added to every test. No point in adding a parameter if its value is always the same global.

t/Util.pm
```diff
@@ -34,6 +34,33 @@ our $HttpConfig = <<_EOC_;
...
+        -- * phase_check_module is just a string used to determine the "module"
+        --   For exmaple, if `phase_check_module` is kbecomesong.response, then `mod` is response
```
hishamhm: two typos in this sentence (also, `response` should be in quotes)

t/Util.pm
```diff
@@ -34,6 +34,33 @@ our $HttpConfig = <<_EOC_;
         local private_phases = require("kong.pdk.private.phases")
         local phases = private_phases.phases
 
+        -- This method executes 1 or more pdk methods twice: the first time with phase
```
hishamhm: s/method/function/ :)

t/Util.pm
```diff
@@ -34,6 +34,33 @@ our $HttpConfig = <<_EOC_;
...
+        --      init_worker   = false,     -- expected to always throw an error on init_worker phase
+        --      certificate   = "pending", -- ignored phase
+        --      rewrite       = true,      -- expected to work with and without the phase checker
+        --      access        = true,
+        --      header_filter = "forced false", -- exit will only error with the phase_checks active
+        --      body_filter   = false,
+        --      log           = false,
+        --      admin_api     = true,
+        --    }
+        --
```
hishamhm: Overall description of the function is very good! Thanks for documenting :)

spec/03-plugins/25-response-rate-limiting/04-access_spec.lua
```diff
@@ -92,7 +92,7 @@ end
 for _, strategy in helpers.each_strategy() do
 for _, policy in ipairs({"local", "cluster", "redis"}) do
 
-describe(fmt("#flaky Plugin: response-ratelimiting (access) with policy: #%s [#%s]", policy, strategy), function()
+describe(fmt("flaky Plugin: response-ratelimiting (access) with policy: #%s [#%s]", policy, strategy), function()
```
hishamhm: Since you verified that tests pass, these can be reverted so that it's `#flaky` again

t/01-pdk/08-response/00-phase_checks.t
```diff
@@ -262,7 +263,76 @@ GET /t
...
+
+    #ssl_certificate_by_lua_block {
+    #    phase_check_functions(phases.certificate)
+    #}
+}
+--- config
+    location /t {
+        proxy_pass http://unix:$TEST_NGINX_NXSOCK/nginx.sock;
+        set $upstream_uri '/t';
+        set $upstream_scheme 'http';
```
thibaultcha: Are those two variables used in this test case?

bungle:
What does `ngx.ctx.KONG_EXITED` mean after this change? It was about to signal early exit. E.g. a plugin that short-circuited proxying (usually in access phase and I think not in response phases, e.g. after proxying).

thibaultcha:
> What does ngx.ctx.KONG_EXITED mean after this change?

`KONG_EXITED` is `true`, and so is `KONG_PROXIED`. That is Kong indeed proxied the request (no short-circuit, but Kong produced the response itself (i.e. the response seen by the client is not that which was received from the upstream). That is where `kong.response.get_source()` loses some of its precisions, but I don't believe the control variables to be incorrectly set.

PR #4040, opened 2018-12-01
perf(acl) properly cache intermediary config

thibaultcha:
Prior to this patch, the `if not config` branch was hit during every
request, effectively nullifying the plugin's config cache optimization.


thibaultcha:
I ran a quick check against bundled plugins and no other seem to have the same defect. Please have a look and confirm @Tieske, thanks.

thibaultcha:
By reusing the pattern for the commit I added to https://github.com/Kong/kong/pull/3872 - noticed we never saved the table. Thanks for taking a look.

PR #4044, opened 2018-12-04
hotfix(admin_api) ensure offset is a string for targets list endpoint

### LINKED ISSUE #4043: Kong 1.0.0rc3 target list endpoint returns offset as a number
### Summary

The request `GET /upstreams/foo/targets?size=1` returns `offset` as an integer and not string, as is the case with other endpoints.

### Steps To Reproduce

1. Start Kong 1.0.0rc3, I'm using the docker image
2. `http :8001/upstreams name=foo`
3. `http :8001/upstreams/foo/targets target=10.1.1.42`
4. `http :8001/upstreams/foo/targets target=10.1.1.43`
5. `http :8001/upstreams/foo/targets?size=1`
```
http ":8001/upstreams/foo/targets?size=1"
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 261
Date: Mon, 03 Dec 2018 00:24:01 GMT
Server: kong/1.0.0rc3

{
    "data": [
        {
            "created_at": 1543796635.211,
            "id": "55f94788-930f-45f6-8f43-da6eb033f479",
            "target": "10.1.1.42:8000",
            "upstream": {
                "id": "bab45234-1421-4684-acf2-90000dd3e982"
            },
            "weight": 100
        }
    ],
    "next": "/upstreams/bab45234-1421-4684-acf2-90000dd3e982/targets?offset=1",
    "offset": 1
}
```
`offset` should be a string and not an integer.

Another oddity with the endpoint is that the response contains `data` as `{} ` and not `[]` when there are no targets.
Steps to reproduce:
1. `http :8001/upstreams name=bar`
2. `http :8001/upstreams/bar/targets`
```
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 24
Date: Mon, 03 Dec 2018 00:31:00 GMT
Server: kong/1.0.0rc3

{
    "data": {},
    "next": null
}
```

hbagdi:
`offset` is a string on the Admin API for all list endpoints.
This change ensures that `/upstreams/:upstream/targets` endpoint
conforms to this expectation.

Fix #4043

### DIFF
--- a/kong/db/dao/targets.lua
+++ b/kong/db/dao/targets.lua
@@ -214,7 +214,7 @@ function _TARGETS:page_for_upstream(upstream_pk, size, offset, options)
 
   local next_offset
   if all_active_targets[size + offset + 1] then
-    next_offset = size + offset
+    next_offset = tostring(size + offset)
   end
 
   return page, nil, nil, next_offset
--- a/spec/02-integration/04-admin_api/08-targets_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/08-targets_routes_spec.lua
@@ -461,6 +461,16 @@ describe("Admin API #" .. strategy, function()
         local json = assert.response(res).has.jsonbody()
         assert.equal(10, #json.data)
       end)
+      it("offset is a string", function()
+        local res = assert(client:send {
+          method = "GET",
+          path = "/upstreams/" .. upstream.name .. "/targets",
+          query = {size = 3},
+        })
+        assert.response(res).has.status(200)
+        local json = assert.response(res).has.jsonbody()
+        assert.is_string(json.offset)
+      end)
       it("paginates a set", function()
         local pages = {}
         local offset
--- a/spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
+++ b/spec/02-integration/000-new-dao/02-db_core_entities_spec.lua
@@ -19,6 +19,8 @@ for _, strategy in helpers.each_strategy() do
         "routes",
         "services",
         "basicauth_credentials",
+        "upstreams",
+        "targets",
       })
     end)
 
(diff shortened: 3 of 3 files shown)


spec/02-integration/04-admin_api/08-targets_routes_spec.lua
```diff
@@ -461,6 +461,17 @@ describe("Admin API #" .. strategy, function()
...
+      it("offset is a string", function()
+        local res = assert(client:send {
+          method = "GET",
+          path = "/upstreams/" .. upstream.name .. "/targets",
+          query = {size = 3},
+        })
+        assert.response(res).has.status(200)
+        local json = assert.response(res).has.jsonbody()
+        assert.not_nil(json.offset)
+        assert.equal("string", type(json.offset))
```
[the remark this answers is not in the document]
hbagdi: 👍 updated.

kong/db/dao/targets.lua
```diff
@@ -214,7 +214,7 @@ function _TARGETS:page_for_upstream(upstream_pk, size, offset, options)
 
   local next_offset
   if all_active_targets[size + offset + 1] then
-    next_offset = size + offset
+    next_offset = tostring(size + offset)
```
thibaultcha: In Cassandra, `offset` is a base64-encoded string. Has this change been tested with the C* strategy?

hbagdi: Good point. I've tested this only with Postgres manually and had wrongly assumed that the test ran both Postgres and C*. Let me test with C*.

hbagdi: Actually I take that back. The tests run using both the strategies and pass.

thibaultcha: Oh, this endpoint defines its own logic to generate the offset, I see... Well, nowadays in the new DAO, PostgreSQL should have base64-encoded offsets as well, so this endpoint is somewhat of an oddity now...

thibaultcha: Let's change this to a `string`, indeed. But for one additional reason, which is that all `offset` return values from the DAO are string, and are tested as such (see tests in https://github.com/Kong/kong/blob/next/spec/02-integration/000-new-dao/02-db_core_entities_spec.lua).

We should add a DAO-level test asserting that the return `offset` value is a string (on top of the Admin API one that we can keep as a regression test).

thibaultcha:
Manually merged with local edits. Thanks.

PR #4046, opened 2018-12-04
fix(templates) turn on TLSv1.3

james-callahan:
This acts as a workaround for https://github.com/openssl/openssl/issues/7660


kong/templates/nginx_kong.lua
```diff
@@ -87,7 +87,7 @@ server {
 > if proxy_ssl_enabled then
     ssl_certificate ${{SSL_CERT}};
     ssl_certificate_key ${{SSL_CERT_KEY}};
-    ssl_protocols TLSv1.1 TLSv1.2;
+    ssl_protocols TLSv1.1 TLSv1.2 TLSv1.3;
```
thibaultcha: Please make the same change in the tests template as well.

james-callahan: Just spec? or spec-old-api as well?

thibaultcha: Both is fine.

bungle: In some another PR, I wish we could make it configurable.

thibaultcha: @bungle This is directive is precisely one of the reasons why we introduced dynamically injected nginx directives. Adding each and every nginx directive is not a scalable solution. Let's not forget that. https://github.com/Kong/kong/pull/3607

bungle: OK, maybe then making this injection friendly?

bungle: What I mean that injection is great for adding, but not so great for changing existing.

thibaultcha: The goal detailed above is to add a default injection from `kong.conf` that one can edit, or override via the equivalent environment variable.

PR #4054, opened 2018-12-05
fix(core) fix getssl to not cache library functions

bungle:
### Summary

This should fix the problem with crashing.


kong/resty/getssl.lua
```diff
@@ -1,6 +1,6 @@
 local ffi = require "ffi"
-local pushssl = require "openssl.ssl".pushffi -- will define SSL* in ffi
-local get_ssl_pointer = require "ngx.ssl".get_ssl_pointer
+local openssl = require "openssl.ssl"
```
james-callahan: This should be named `openssl_ssl` or similar.

PR #4063, opened 2018-12-07
fix(pdk) make kong.request.get_body() to return nil, err on simple json values

### LINKED ISSUE #4007: OAuth2 Protected Proxy yeilds HTTP 500 error
### Summary

We have a proxy protected with ACL and OAUTH2.0, recently a consumer came to us with a specific proxy path on their wildcarded proxy returning HTTP 500's. Our logs indicate it stopped @ the gateway and never proxied. Checking Kong logs I see this error.

```
2018/11/21 13:54:58 [error] 84#0: *10614388 lua coroutine: runtime error: /usr/local/share/lua/5.1/kong/tools/utils.lua:430: bad argument #1 to 'pairs' (table expected, got number)
stack traceback:
coroutine 0:
               [C]: in function 'pairs'
               /usr/local/share/lua/5.1/kong/tools/utils.lua:430: in function 'retrieve_parameters'
               /usr/local/share/lua/5.1/kong/plugins/oauth2/access.lua:453: in function 'parse_access_token'
               /usr/local/share/lua/5.1/kong/plugins/oauth2/access.lua:531: in function 'do_authentication'
               /usr/local/share/lua/5.1/kong/plugins/oauth2/access.lua:603: in function 'execute'
               /usr/local/share/lua/5.1/kong/plugins/oauth2/handler.lua:12: in function </usr/local/share/lua/5.1/kong/plugins/oauth2/handler.lua:10>
coroutine 1:
               [C]: in function 'resume'
               coroutine.wrap:21: in function <coroutine.wrap:21>
               /usr/local/share/lua/5.1/kong/init.lua:468: in function 'access'
access_by_lua(nginx.conf:119):2: in function <access_by_lua(nginx.conf:119):1>, client: 10.***.*.***, server: kong, request: "POST /api/test2/provider/v1/bill/Service HTTP/1.1", host: "gateway.company.com"
```

OAuth2.0 Plugin on route config(pretty standard):
```
{
  "created_at": 1537827592725,
  "config": {
    "refresh_token_ttl": 1209600,
    "token_expiration": 7200,
    "mandatory_scope": false,
    "accept_http_if_already_terminated": false,
    "hide_credentials": false,
    "enable_implicit_grant": false,
    "global_credentials": true,
    "enable_client_credentials": true,
    "enable_password_grant": false,
    "anonymous": "",
    "enable_authorization_code": false,
    "provision_key": "function",
    "auth_header_name": "authorization"
  },
  "id": "91b369b6-fb78-4562-9138-a20fc0814b5a",
  "name": "oauth2",
  "route_id": "cc488204-fcf6-4826-b30e-3bbe8bdaee3f",
  "enabled": true
}
```

My initial analysis reveals this:

1. They could not be passing the token as Authorization: Bearer <token> header otherwise the first catch in the parse_access_token would be hit and handled properly. So it goes to retrieve_parameters call which checks the http body and the uri for query parameters.

My guess is that at this spot: https://github.com/Kong/kong/blob/master/kong/plugins/oauth2/access.lua#L113 the https://github.com/Kong/kong/blob/master/kong/tools/utils.lua#L420 , table merge call fails to handle certain edge cases based on the uri_args that were collected 🔥  .

### Steps To Reproduce

1. One way to reproduce is in the HTTP Body just post a raw numeric string with nothing else. Leave off the Authorization: Bearer {token} if you are running latest Kong master branch with my PR that looks for header first before body. If you run the official 14.1 Kong release it looks for body before caring about token in header so error will occur even if you send proper token in header.

### Additional Details & Logs
- Kong version 14.1

bungle:
### Summary

Changes `kong.request.get_body()` to return `nil, err` on `application/json` simple types as a root (and only) content:
- strings
- numbers
- booleans
- nulls

So you can after this is merged only use this to read following JSON content:
- arrays
- objects

as a root content.

Fix #4007

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -638,7 +638,7 @@ local function new(self)
       cjson.decode_array_with_array_mt(true)
       local json = cjson.decode(body)
       cjson.decode_array_with_array_mt(false)
-      if not json then
+      if type(json) ~= "table" then
         return nil, "invalid json body", CONTENT_TYPE_JSON
       end
 
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -500,9 +500,6 @@ local function parse_access_token(conf)
   local found_in = {}
   local access_token = kong.request.get_header(conf.auth_header_name)
   if access_token then
-    if type(access_token) == "table" then --Take the first found
-      access_token = access_token[1]
-    end
     local parts = {}
     for v in access_token:gmatch("%S+") do -- Split by space
       table.insert(parts, v)
@@ -513,6 +510,9 @@ local function parse_access_token(conf)
     end
   else
     access_token = retrieve_parameters()[ACCESS_TOKEN]
+    if type(access_token) ~= "string" then
+      return
+    end
   end
 
   if conf.hide_credentials then
--- a/t/01-pdk/04-request/16-get_body.t
+++ b/t/01-pdk/04-request/16-get_body.t
@@ -174,7 +174,227 @@ error: invalid json body
 
 
 
-=== TEST 7: request.get_body() content type value is case-insensitive
+=== TEST 7: request.get_body() returns error with null json body
+--- http_config eval: $t::Util::HttpConfig
+--- config
+    location = /t {
+        access_by_lua_block {
+            local PDK = require "kong.pdk"
+            local pdk = PDK.new()
+
+            local json, err = pdk.request.get_body()
+
+            ngx.say("type: ", type(json))
+            ngx.say("error: ", err)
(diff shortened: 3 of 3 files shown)


kong/pdk/request.lua
```diff
@@ -638,7 +638,7 @@ local function new(self)
       cjson.decode_array_with_array_mt(true)
       local json = cjson.decode(body)
       cjson.decode_array_with_array_mt(false)
-      if not json then
+      if type(json) ~= "table" then
```
Tieske: lgtm, but this is quite opinionated, so maybe @thibaultcha has some thoughts on this as well?

PR #4065, opened 2018-12-07
feat(core) add ctx.is_service_mesh_request context value with tls mesh

bungle:
### Summary

This PR makes it possible for code to check if the Kong node is running as a `service-mesh` node or not with `mutual tls` connection between two Kong nodes in a same cluster using stream module.


bungle:
@james-callahan I think it is only set to `true` with `http` too. This is internal, so lets start with this. I can make a card to rename it, maybe we do that before 1.0. But for rc4 this is ok. Thanks for suggesting that.

PR #4066, opened 2018-12-07
feat(core) enable plugin.run_on support on plugins iterator

bungle:
### Summary

This will add support for `run_on` attribute on Plugins to make them run on:
1. `all` Kong nodes regardless
2. `first` Kong node in `gateway` or `service-mesh` modes only
3. `second` Kong node in `service-mesh` mode only

This is manually tested to be working with latest (not yet released) `luaossl` and our (released) `openresty-patches`, but this does not include `busted` tests yet. If that is a blocker for `release` of a release candidate, then I will update this PR on monday.

### DIFF
--- a/kong/runloop/plugins_iterator.lua
+++ b/kong/runloop/plugins_iterator.lua
@@ -45,16 +45,31 @@ local function load_plugin_configuration(ctx,
     return ngx.exit(ngx.ERROR)
   end
 
-  if plugin ~= nil and plugin.enabled then
-    local cfg = plugin.config or {}
-
-    cfg.api_id      = plugin.api and plugin.api.id
-    cfg.route_id    = plugin.route and plugin.route.id
-    cfg.service_id  = plugin.service and plugin.service.id
-    cfg.consumer_id = plugin.consumer and plugin.consumer.id
-
-    return cfg
+  if not plugin or not plugin.enabled then
+    return
   end
+
+  if plugin.run_on ~= "all" then
+    if ctx.is_service_mesh_request then
+      if plugin.run_on == "first" then
+        return
+      end
+
+    else
+      if plugin.run_on == "second" then
+        return
+      end
+    end
+  end
+
+  local cfg = plugin.config or {}
+
+  cfg.api_id      = plugin.api and plugin.api.id
+  cfg.route_id    = plugin.route and plugin.route.id
+  cfg.service_id  = plugin.service and plugin.service.id
+  cfg.consumer_id = plugin.consumer and plugin.consumer.id
+
+  return cfg
 end
 
 
--- a/spec/02-integration/05-proxy/03-plugins_triggering_spec.lua
+++ b/spec/02-integration/05-proxy/03-plugins_triggering_spec.lua
@@ -27,6 +27,7 @@ for _, strategy in helpers.each_strategy() do
         "keyauth_credentials",
       }, {
         "error-handler-log",
+        "error-generator-pre",
       })
 
       local consumer1 = bp.consumers:insert {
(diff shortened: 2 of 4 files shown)


kong/runloop/plugins_iterator.lua
```diff
@@ -45,7 +45,11 @@ local function load_plugin_configuration(ctx,
     return ngx.exit(ngx.ERROR)
   end
 
-  if plugin ~= nil and plugin.enabled then
+  if plugin ~= nil and plugin.enabled and
+    (plugin.run_on == "all" or
+    (ctx.is_service_mesh_request ~= true and plugin.run_on == "first") or
```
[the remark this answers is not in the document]
bungle: Is it now ok?

kong/runloop/plugins_iterator.lua
```diff
@@ -45,7 +45,11 @@ local function load_plugin_configuration(ctx,
     return ngx.exit(ngx.ERROR)
   end
 
-  if plugin ~= nil and plugin.enabled then
+  if plugin ~= nil and plugin.enabled and
+    (plugin.run_on == "all" or
```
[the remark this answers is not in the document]
bungle: I tried to move but it seems like Github didn't make this outdated.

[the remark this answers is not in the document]
bungle: oh it actually did.

thibaultcha:
The `chore` commit can use the more descriptive scope `perf`.

thibaultcha:
Ok, no big deal! Let’s keep it as-is then.

kong/runloop/plugins_iterator.lua
```diff
@@ -45,7 +45,10 @@ local function load_plugin_configuration(ctx,
     return ngx.exit(ngx.ERROR)
   end
 
-  if plugin ~= nil and plugin.enabled then
+  if plugin ~= nil and plugin.enabled and (plugin.run_on == "all"    or
+       (not ctx.is_service_mesh_request and plugin.run_on == "first") or
+           (ctx.is_service_mesh_request and plugin.run_on == "second")) then
```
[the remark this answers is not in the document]
bungle: I didn't like it either, I just send a new one, which I think is better.

kong/runloop/plugins_iterator.lua
```diff
@@ -45,16 +45,26 @@ local function load_plugin_configuration(ctx,
...
+  if not plugin or not plugin.enabled then
+    return
+  end
 
-    return cfg
+  if plugin.run_on ~= "all" then
+    if ctx.is_service_mesh_request and plugin.run_on == "first" then
+      return
+    elseif plugin.run_on == "second" then
+      return
```
thibaultcha: Seems like plugins with `run_on == "second"` would never run, since they'd enter this branch every time.

spec/02-integration/05-proxy/03-plugins_triggering_spec.lua
```diff
@@ -1000,5 +1001,111 @@ for _, strategy in helpers.each_strategy() do
...
+          assert(bp.plugins:insert {
+            name = "error-generator-pre",
+            route = { id = route2.id },
+            config = {
+              access = true,
+            },
+            run_on = "first",
+          })
+
+          local route3 = assert(db.routes:insert {
```
kikito: It seems this route is not used on any tests.

kong/runloop/plugins_iterator.lua
```diff
@@ -45,16 +45,31 @@ local function load_plugin_configuration(ctx,
...
+    if ctx.is_service_mesh_request then
+      if plugin.run_on == "first" then
+        return
+      end
 
-    return cfg
+    else
+      if plugin.run_on == "second" then
+        return
+      end
```
kikito: I think this nested if could be simplified to:

```
if (plugin.run_on == "first" and ctx.is_service_mesh_request)
or (plugin.run_on == "second" and not ctx.is_service_mesh_request)
then
  return
end
```

bungle: It originally was like that. Then a lot of comments about spacing, then I came with current, which is easiest to understand, imo.

spec/02-integration/05-proxy/03-plugins_triggering_spec.lua
bungle: @kikito, yes, I forgot to push latest, it should be now there.

spec/02-integration/05-proxy/03-plugins_triggering_spec.lua
```diff
@@ -1000,5 +1001,123 @@ for _, strategy in helpers.each_strategy() do
...
+
+        assert.res_status(500, res)
+      end)
+
+      it("= second does not get executed when running on first", function()
+        local res = assert(proxy_client:send {
+          method = "GET",
+          path = "/status/200",
+          headers = {
+            ["Host"] = "run-on-second.org",
```
james-callahan: Avoid potentially real domain names in tests? Can we prefer subdomains of `example.com`/org/net?

p0pr0ck5: +1 for `example.com` please

bungle: It already went in., sorry :-(.

PR #4068, opened 2018-12-07
fix(mesh) set up SSL_CTX in the init phase

james-callahan:
With OpenSSL 1.1.1 setting up the `SSL_CTX*` in the certificate
phase no longer works.

This commit introduces a new kong.resty.config module which
uses the patches at https://github.com/Kong/openresty-patches/pull/31 and https://github.com/Kong/openresty-patches/pull/32

### DIFF
--- a/kong/resty/config.lua
+++ b/kong/resty/config.lua
@@ -1,0 +1,64 @@
+local ffi = require "ffi"
+local openssl_ssl_context = require "openssl.ssl.context"
+
+local has_patch = pcall(ffi.cdef, [[
+  void* ngx_http_lua_get_server_block(uintptr_t);
+  ngx_str_t* ngx_http_lua_server_block_server_name(void*);
+  SSL_CTX* ngx_http_lua_ssl_get_SSL_CTX(void*);
+]])
+
+
+local ngx_lua_get_server_block
+local ngx_lua_server_block_server_name
+local ngx_lua_get_SSL_CTX
+if has_patch and ngx.config.subsystem == "http" then
+  ngx_lua_get_server_block = ffi.C.ngx_http_lua_get_server_block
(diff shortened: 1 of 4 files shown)


kong/resty/config.lua
```diff
@@ -0,0 +1,57 @@
...
+    if srv == nil then
+      return nil
+    end
+    i = i + 1
+    return srv
+  end
+end
+
+return {
+  get_server_block = get_server_block;
```
bungle: shoudn't we use `,` here instead of `;`

kong/resty/config.lua
```diff
@@ -0,0 +1,57 @@
...
+  each_server_block = each_server_block;
```
bungle: oudn't we use `,` here instead of `;`

kong/runloop/mesh.lua
```diff
@@ -155,7 +152,16 @@ local function init()
   mesh_server_ssl_ctx:setCertificate(node_cert)
   mesh_server_ssl_ctx:setStore(ca_store)
   mesh_server_ssl_ctx:setVerify(openssl_ssl.VERIFY_PEER + openssl_ssl.VERIFY_FAIL_IF_NO_PEER_CERT)
-  mesh_server_ssl_ctx:setAlpnSelect(simple_mesh_alpn_select, mesh_alpn)
+  if ngx.config.subsystem == "http" then
+    for server_block in resty_config.each_server_block() do
+      if server_block:get_server_name() == "kong" then -- Assume server_name for proxy is "kong"
+        local ssl_ctx = server_block:get_ssl_ctx()
+        ssl_ctx:setAlpnSelect(nginx_mesh_alpn_select, mesh_server_ssl_ctx, mesh_alpn)
```
bungle: I guess `mesh_server_ssl_ctx` needs to be `ssl_ctx`?

Issue #4049, opened 2018-12-05
1.0.0rc2 update to 1.0.0.rc3


ZTzha:
### Summary

step 1:
use old database kong
kong can't start

Error: 
./kong/cmd/start.lua:63: ./kong/cmd/start.lua:52: nginx: [warn] [lua] getssl.lua:10: OpenResty patch missing. See https://github.com/Kong/openresty-patches
nginx: [warn] [lua] init.lua:260: init(): missing "ngx.balancer".set_ssl_ctx API. Dynamic client SSL_CTX* will be unavailable
nginx: [error] init_by_lua error: ./kong/init.lua:292: [postgres] 错误: 字段 "run_on" 不存在 (154)
stack traceback:
        [C]: in function 'assert'
        ./kong/init.lua:292: in function 'init'
        init_by_lua:3: in main chunk

stack traceback:
        [C]: in function 'error'
        ./kong/cmd/start.lua:63: in function 'cmd_exec'
        ./kong/cmd/init.lua:87: in function <./kong/cmd/init.lua:87>
        [C]: in function 'xpcall'
        ./kong/cmd/init.lua:87: in function <./kong/cmd/init.lua:44>
        /usr/local/bin/kong:7: in function 'file_gen'
        init_worker_by_lua:47: in function <init_worker_by_lua:45>
        [C]: in function 'xpcall'
        init_worker_by_lua:54: in function <init_worker_by_lua:52>

step 2:
drop database kong; 
then
CREATE DATABASE kong OWNER kong;
kong can't start

### Steps To Reproduce

1.kong migrations bootstrap
2.kong migrations up
3.kong start --vv
4.

### Additional Details & Logs

- Kong version 1.0.0rc3
- Kong debug-level startup logs 
Error: 
/usr/local/share/lua/5.1/kong/cmd/start.lua:63: /usr/local/share/lua/5.1/kong/cmd/start.lua:52: nginx: [warn] [lua] getssl.lua:10: OpenResty patch missing. See https://github.com/Kong/openresty-patches
nginx: [warn] [lua] init.lua:260: init(): missing "ngx.balancer".set_ssl_ctx API. Dynamic client SSL_CTX* will be unavailable
nginx: [error] init_by_lua error: /usr/local/share/lua/5.1/kong/runloop/mesh.lua:158: attempt to call method 'setAlpnSelect' (a nil value)
stack traceback:
        /usr/local/share/lua/5.1/kong/runloop/mesh.lua:158: in function 'init'
[...]

Issue #4042, opened 2018-12-02
Can't update kong from 1.0.0rc2 to 1.0.0rc3


pocketdigi:
### Summary
Can't update kong from 1.0.0rc2 to 1.0.0rc3
Current version:1.0.0rc2
Database type:PostgreSQL
OS: CentOS 7

### Steps To Reproduce
After I install 1.0.0rc3, and execute
```
kong migrations up
```
it just show `database is already up-to-date`,but nothing happened,then I start Kong
```
kong start
```
show an error:

```
Error: /usr/local/share/lua/5.1/kong/cmd/start.lua:52: nginx: [warn] [lua] getssl.lua:10: OpenResty patch missing. See https://github.com/Kong/openresty-patches
nginx: [warn] [lua] init.lua:260: init(): missing "ngx.balancer".set_ssl_ctx API. Dynamic client SSL_CTX* will be unavailable
nginx: [error] init_by_lua error: /usr/local/share/lua/5.1/kong/init.lua:292: [postgres] ERROR: column "run_on" does not exist (154)
stack traceback:
	[C]: in function 'assert'
	/usr/local/share/lua/5.1/kong/init.lua:292: in function 'init'
	init_by_lua:3: in main chunk


  Run with --v (verbose) or --vv (debug) for more details
```

If I use a new database, Kong can start normally, so I think 1.0.0rc3 is install successfully,but the migration script may have some problems.

ZTzha:
i have the same problem

asomir:
i have the same problem

asomir:
A hope, it help for you
[root@kong-01 ~]# kong migrations reset -c /etc/kong/kong.conf
> Are you sure? This operation is irreversible. [Y/n] y
resetting database...
database successfully reset
[root@kong-01 ~]# systemctl start kong
Job for kong.service failed because the control process exited with error code. See "systemctl status kong.service" and "journalctl -xe" for details.
[root@kong-01 ~]# less /usr/local/kong/logs/error.log
[root@kong-01 ~]# kong migrations bootstrap -c /etc/kong/kong.conf
bootstrapping database...
migrating core on database 'kong'...
core migrated up to: 000_base (executed)
core migrated up to: 001_14_to_15 (executed)
migrating oauth2 on database 'kong'...
oauth2 migrated up to: 000_base_oauth2 (executed)
oauth2 migrated up to: 001_14_to_15 (executed)
migrating acl on database 'kong'...
[...]

Issue #4033, opened 2018-11-29
Kong 1.0 rc3 Postgres migrations failing in custom tests


mars:
*This issue is similar in context to #4031, but a different issue.*

### Summary

When running custom tests with a Postgres backend in [this Kong app](https://github.com/heroku/heroku-kong/tree/kong-1.0/spec) (uses the `spec/helpers.lua` & a `bin/busted` runner modeled after the one in Kong master), the tests error.

Locally these tests work with a local Postgres database (solved in #4031), but fail in this different way when run by Heroku CI tests with hosted, cloud Postgres:

```
$ bin/test -o gtest -v
[==========] Running tests from scanned files.
[----------] Global test environment setup.
[----------] Running tests from spec/integration/kong/plugins/hello-world-header_spec.lua
2018/11/28 23:39:25 [warn] 2876#0: *2 [lua] connector.lua:196: setkeepalive(): unable to close postgres connection, context: ngx.timer
2018/11/28 23:39:25 [warn] 2876#0: *2 [lua] connector.lua:196: setkeepalive(): unable to close postgres connection, context: ngx.timer
./spec/helpers.lua:156: [postgres error] Error during migration 2016-05-04-160000_resp_trans_schema_changes: [postgres] ERROR: column "run_on" does not exist (154)
stack traceback:
	./spec/helpers.lua:156: in function 'get_db_utils'
	spec/integration/kong/plugins/hello-world-header_spec.lua:11: in function <spec/integration/kong/plugins/hello-world-header_spec.lua:10>
[----------] 0 tests from spec/integration/kong/plugins/hello-world-header_spec.lua (3966.10 ms total)
```

I've confirmed that the database is empty before the tests run.

Why would these test migrations fail like this?

### Steps To Reproduce

1. Deploy to Heroku a Kong app from [this commit](https://github.com/heroku/heroku-kong/tree/74c1f7fbd6d09c4959a08ed7c0872048d828d459) (in the kong-1.0 branch)
2. Add the app to a new Heroku Pipeline
3. Connect the Pipeline with your own GitHub fork of [heroku-kong](https://github.com/heroku/heroku-kong)
3. Enable Tests in the Pipeline
4. Start a new test on the `kong-1.0` branch
5. Watch the test fail

### Additional Details & Logs

- Kong version **1.0.0rc3**
- Operating system **Heroku-18 (Ubuntu 18.04)**

PR #4071, opened 2018-12-10
feat(migrations) add --force flag

hishamhm:
Adds a `--force` flag to `kong migrations up` to allow for migrating from one release candidate to the next (since the name of the migration file is unchanged). 

A more robust solution should be in place for future rc cycles, but this might be sufficient for now. Tested locally by:

1. checking out 1.0.0rc2
2. resetting/migrating the database
3. checking out `next`
4. running `kong migrations up --force`
5. running the test suite

Verified to fail with the errors seen by users when skipping step 4 and to work with Postgres and Cassandra when including it.

See #4033, #4042, #4049.

### DIFF
--- a/kong/cmd/utils/migrations.lua
+++ b/kong/cmd/utils/migrations.lua
@@ -116,7 +116,7 @@ local function up(schema_state, db, opts)
 
     else
       -- fresh install: must bootstrap (which will run migrations up)
-      error("can't run migrations: database needs bootstrapping; " ..
+      error("cannot run migrations: database needs bootstrapping; " ..
             "run 'kong migrations bootstrap'")
     end
   end
@@ -128,8 +128,19 @@ local function up(schema_state, db, opts)
       error("database has pending migrations; run 'kong migrations finish'")
     end
 
+    if opts.force then
+      log.debug("forcing re-execution of these migrations:\n%s",
+                schema_state.executed_migrations)
+
+      assert(db:run_migrations(schema_state.executed_migrations, {
+        run_up = true,
+      }))
+    end
+
     if not schema_state.new_migrations then
-      log("database is already up-to-date")
+      if not opts.force then
+        log("database is already up-to-date")
+      end
       return
     end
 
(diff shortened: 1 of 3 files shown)


kong/cmd/utils/migrations.lua
```diff
@@ -151,16 +160,18 @@ local function up(schema_state, db, opts)
...
     log("can't run migrations: database not bootstrapped")
     return
   end
 
-  local opts = {
-    ttl = ttl,
-    no_wait = true, -- exit the mutex if another node acquired it
-  }
+  if opts.force then
+    log("cannot use --force with 'finish'")
```
thibaultcha: let's keep the language consistent: cannot vs can't (above). Which should pick one and stick with it (any of the two is fine imho).

kong/cmd/utils/migrations.lua
```diff
@@ -128,6 +128,15 @@ local function up(schema_state, db, opts)
       error("database has pending migrations; run 'kong migrations finish'")
     end
 
+    if opts.force then
+      log.debug("forcing re-execution of these migrations:\n%s",
+                schema_state.executed_migrations)
+
+      assert(db:run_migrations(schema_state.executed_migrations, {
+        run_up = true,
+      }))
```
thibaultcha: The error messages following this branch would look slightly weird imho. E.g.:
```
forcing re-execution of these migrations: [...]
...
...
database is already up-to-date
```

And:
```
forcing re-execution of these migrations: [...]
...
...
migrations to run: [...]
```

It would be enough to slightly update these logs so that their text makes sense in both contexts (forced and not forced migrations). Or to wrap them in appropriate branches determining if the `--force` flag is in use.

hishamhm: That module already had 2 `can't`s and 2 `cannot`s in other messages :) Will standardize them on `cannot` for :robot: computer voice :robot:  reasons, then

hishamhm: regarding the messages, I think just removing the `database is already up-to-date` message in the forced case would be enough.

kong/cmd/utils/migrations.lua
hishamhm: That module already had 2 `can't`s and 2 `cannot`s in other messages :) Will standardize them on `cannot` for :robot: computer voice :robot:  reasons, then

thibaultcha: Awesome!

kong/cmd/utils/migrations.lua
thibaultcha: Yep, sounds good to me!

hishamhm: updated

PR #4075, opened 2018-12-12
test(rate-limiting) fix rate-limiting access tests and remove old dao

bungle:
### Summary

Simple fix that makes `rate-limiting` tests work rather un-flaky (also removes the need for old dao too).


hishamhm:
Please add a separate commit to this PR that removes the `#flaky` tag so that we can experiment running this on Travis! (this commit will be temporary so we can easily drop it if we decide to merge but keep the tag.) Doing this will let us see both the improvements in flakiness and also how slow this test suite currently is.

If we observe that it improves flakiness but is still super slow, I think we should merge this but not remove the tag yet, before we make the running time reasonable (I think some individual tests in there can wait up to a full minute).

PR #4077, opened 2018-12-12
fix(admin) avoid Lapis handler when handler doesn't exit explicitly

hishamhm:
If an Admin API handler did not exit explicitly using `kong.response.exit` (or
`ngx.exit`), it would end up concatenating HTML produced by Lapis (for an
example of this behavior, remove `kong.response.exit(200)` from the `collect`
function in kong-prometheus-plugin 0.3.2).

This ensures that only the handler function executes without falling into
Lapis.

Includes a plugin fixture and a regression test.

### DIFF
--- a/kong/api/endpoints.lua
+++ b/kong/api/endpoints.lua
@@ -240,7 +240,7 @@ local function post_collection_endpoint(schema, foreign_schema, foreign_field_na
     if post_process then
       entity, _, err_t = post_process(entity)
       if err_t then
-        handle_error(err_t)
+        return handle_error(err_t)
       end
     end
 
--- a/kong/api/init.lua
+++ b/kong/api/init.lua
@@ -44,7 +44,12 @@ local function parse_params(fn)
 
     self.params = api_helpers.normalize_nested_params(self.params)
 
-    return fn(self, ...)
+    local res = fn(self, ...)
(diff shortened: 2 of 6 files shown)


hishamhm:
This was prompted by @hbagdi's issue with the prometheus plugin here: https://github.com/Kong/kong-plugin-prometheus/issues/32

Once he removed `kong.response.exit` he started getting HTML produced by Lapis. I have an alternate branch that makes `kong.response.exit` work like `ngx.exit` in that case (which would fix his problems) but I don't think this is the right way to go: sending `ngx.exit(200)` there "works", but the argument is silently ignored because the headers and body have already been sent. Our behavior of failing `kong.response.exit(200)` is more correct in that regard.

However, I don't know what the handler method there in `kong.api.init` is supposed to be returning anyway (if anything). I'd like to check with someone more knowledgeable about the flow here if my solution there is correct.

hishamhm:
@bungle The patch as-is is breaking a lot of tests, I think it is breaking the delayed-response mechanism. But the general idea is that, if the plugin handler in `api.lua` doesn't return explicitly (with `kong.responses.exit` or `ngx.exit`) the code is currently falling through to Lapis and producing HTML. We want it to not do it.

PR #4081, opened 2018-12-13
fix(db) store connection in request-aware context

thibaultcha:
Calling `db:connect()` used to store the connection in as an instance
attribute in `self`, therefore in a worker singleton (`kong.db`), which
would produce `bad request` errors when used at runtime (i.e.
non-migration contexts).

This error presents itself when within the `plugins:select_by_cache_key`
special DAO method, in which the schema_state module performs a
`connect()` and subsequent operations to determine the database schema
state.

Storing the connection object in the request aware context (`ngx.ctx`)
prevents this issue. We still store the connection as an instance
attribute in non-request contexts.

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -9,14 +9,15 @@ CassandraConnector.__index = CassandraConnector
 
 
 local function wait_for_schema_consensus(self)
-  if not self.connection then
+  local conn = self:get_stored_connection()
+  if not conn then
     error("no connection")
   end
 
   log.verbose("waiting for Cassandra schema consensus (%dms timeout)...",
               self.cluster.max_schema_consensus_wait)
 
-  local ok, err = self.cluster:wait_schema_consensus(self.connection)
+  local ok, err = self.cluster:wait_schema_consensus(conn)
 
   log.verbose("Cassandra schema consensus: %s",
               ok and "reached" or "not reached")
@@ -172,8 +173,9 @@ end
 
 
 function CassandraConnector:connect()
-  if self.connection then
-    return true
+  local conn = self:get_stored_connection()
+  if conn then
+    return conn
   end
 
   local peer, err = self.cluster:next_coordinator()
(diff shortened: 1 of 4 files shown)


hishamhm:
@thibaultcha Thanks for the write-up! For later cross-referencing, where did this come about? (Github issue, Kong Nation link, did you bump into it yourself?)

kong/db/strategies/postgres/connector.lua
```diff
@@ -690,7 +702,9 @@ function _mt:run_up_migration(name, up_sql)
     error("up_sql must be a string", 2)
   end
 
-  if not self.connection or not self.connection.sock then
+  local conn = self:get_stored_connection()
+  if not conn then
+    error("no connection")
```
hishamhm: this line ended up duplicated

kong/db/strategies/cassandra/connector.lua
```diff
@@ -9,14 +9,15 @@ CassandraConnector.__index = CassandraConnector
 
 
 local function wait_for_schema_consensus(self)
-  if not self.connection then
+  local conn = self.super.get_stored_connection(self)
```
hishamhm: the Cassandra strategy does `self.super.get_stored_connection`; the Postgres strategy simply does `self:get_stored_connection`. AFAICT they seem functionally equivalent. Is the explicit `super` (and additional metatable logic) necessary?

hishamhm: (No strong feeling on this — I'm more asking whether there was a technical reason or if it was a stylistic choice, though consistency between strategies would be nice if the difference is unnecessary)

thibaultcha:
The explicit `super` is intended, here is a leftover as I was working through the fix.

kong/db/strategies/postgres/connector.lua
thibaultcha: Oops

thibaultcha:
@hishamhm No specific link to reference, or else I would have included it of course :)

kong/db/strategies/cassandra/connector.lua
hishamhm: Got it, I see it now

PR #4087, opened 2018-12-13
feat(core) router rebuild timer (instead of rebuilding on request)

bungle:
### Summary

This is a PoC of semaphore-less router rebuilding that happens on `ngx.timer` instead of rebuilding a new router during the request and queuing the requests. 

Recently we merged #4084, but I never have liked that we do semaphores during requests, for me it is an anti-pattern, but of course it cannot be always avoided. Here is a PoC for discussion that avoids the request based semaphore for router rebuilding. Edit: not totally as I added a backup there, and our test suite in its current form cannot pass as it does not wait for routers to rebuild, so I added 1 sec wait in request phase, still a lot less than 5 secs (though it was configurable), and if 1 sec (maybe make it configurable) goes it will use whatever router there is.

Things to know about this:
1. it always ensures that router represents a stable database state (e.g. it fixes the possible `each` issue too)
2. it does not check router version on request phase, but if you need that it is really easy to add, so this implementation is eventually consistent. During the rebuilding the old stable router is used.

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -359,7 +359,7 @@ function Kong.init()
     error("error building initial plugins map: ", err)
   end
 
-  assert(runloop.build_router(db, "init"))
+  assert(runloop.build_router("init"))
   assert(runloop.build_api_router(dao, "init"))
 end
 
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -23,6 +23,7 @@ local certificate = require "kong.runloop.certificate"
 
 
 local kong        = kong
+local pcall       = pcall
 local tostring    = tostring
 local tonumber    = tonumber
 local sub         = string.sub
@@ -38,6 +39,7 @@ local unpack      = unpack
 
 
 local ERR         = ngx.ERR
+local CRIT        = ngx.CRIT
 local DEBUG       = ngx.DEBUG
 
 
@@ -45,7 +47,7 @@ local CACHE_ROUTER_OPTS = { ttl = 0 }
 local EMPTY_T = {}
 
 
-local get_router, build_router
+local get_router, build_router, rebuild_router, router_version
 local api_router, api_router_version, api_router_err
 local server_header = meta._SERVER_TOKENS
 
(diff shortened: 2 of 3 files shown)


p0pr0ck5:
I will note (only note, not driving any discussion) that this approach was previously proposed and rejected in https://github.com/Kong/kong/pull/3696.

bungle:
@p0pr0ck5 that's why with this approach you can add `semaphore:wait` to `get_router` to wait `0-n` seconds for `timer` to finish.

bungle:
The point is that we should not build router during a request because that will mean it is always build on request, so there is always a requests that it affects. With timer there isn't. But we can still make requests to queue if we want. The key is that rebuilding happens whether or not there is requests.

thibaultcha:
I haven't looked at this PR yet, but as long as we guarantee node consistency (that is, that all workers share the same router starting from a given request), then we are fine. If not, then that is an absolute blocker for this to move forward.

p0pr0ck5:
@thibaultcha I will note very briefly that that's not the behavior we see now, because router rebuild events are triggered based on a one-second poll of the events shm. See https://github.com/Kong/kong/issues/4055#issuecomment-445313185 (and if I'm missing something, apologies, but that behavior has been my observation during stress testing).

thibaultcha:
@p0pr0ck5 Correct. The delay is acceptable (in years that it has been in place, no new issues about workers inconsistency has been reported ever again) and will be present for as long as we do not follow the `ipc:update()` path, which the intent was to start using eventually.

thibaultcha:
We must also start being very careful with such timers on a critical code path. We've started receiving reports of some nodes hitting the maximum (currently hard-coded) number of timers in some cases. Ideally such a critical path as the router (vs, stay, logging plugins creating timers) should be resilient to timer scheduling failures. Ideally.

Also, not to be discouraging (all improvements in this area is very much welcome, as were the plugins map and router semaphore!), but such a change probably shouldn't ship in 1.0 - just so we all are on the same page here :)

bungle:
this pr creates only a single timer per worker, if one is running, other one will not start.

thibaultcha:
Sure, that is great. What if the timer cannot start?

bungle:
@thibaultcha good point, I changed it to create then a timer on request phase too. Still it will not make request phase to queue or wait. Yes this is for discussion and experimenting, not for 1.0.0.

bungle:
I will close this. It was after several iterations a failed approach.

Conclusion:
- it is bad that we rely router rebuilding on request and queue requests
- it is bad that we don't give any router to request if we cannot get the most recent
- our paging may introduce a flakiness
- it is hard to change this without making workers less consistent (e.g. using stale routers)
- if we change this to be eventually consistent, a lot of our test suite breaks
- making this more granular is not easy

PR #4092, opened 2018-12-14
docs(upgrade) add the 1.0.0 upgrade path

hishamhm:


hbagdi:
Can we clarify what happens to entities that are related to APIs in Kong?
For example, as part of migration to Kong 1.0, do we delete plugins that are enabled for a specific API? What happens to plugins that are enabled for a specific API and a specific consumer?

hishamhm:
@hbagdi added a paragraph about migrations on the presence of apis in the "Preliminary checks" section. Please take a look!

hbagdi:
@hishamhm Thanks for adding that in! Very helpful!

PR #4098, opened 2018-12-15
feat: router mutex fallthrough

thibaultcha:
Upon semaphore acquisition timeout (due to another coroutine being slow to release it while performing its own router rebuild), we now fallthrough to rebuild the router in order to ensure a best effort strategy.

I considered adding a second level semaphore (e.g. releasing a few more resources for some other coroutines instead of _all_ waiting coroutines), but this behavior is close to 0.14 all the while preserving the new router mutex in most cases (as long as they don't timeout). This alleviates some of the issues observed in #4055.

Also see discussion [here](https://github.com/Kong/kong/pull/4084#discussion_r241475200) (and other references there) for context around this behavior.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -38,6 +38,7 @@ local unpack      = unpack
 
 
 local ERR         = ngx.ERR
+local WARN        = ngx.WARN
 local DEBUG       = ngx.DEBUG
 
 
@@ -48,7 +49,7 @@ local EMPTY_T = {}
 local get_router, build_router
 local api_router, api_router_version, api_router_err
 local server_header = meta._SERVER_TOKENS
-
+local _set_check_router_rebuild
 
 local build_router_semaphore
 local build_api_router_semaphore
@@ -194,6 +195,12 @@ do
   end
 
 
+  -- for unit-testing purposes only
+  _set_check_router_rebuild = function(f)
+    check_router_rebuild = f
+  end
+
+
   get_router = function()
     local version, err = singletons.cache:get("router:version",
                                               CACHE_ROUTER_OPTS,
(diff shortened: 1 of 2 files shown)


kong/runloop/handler.lua
```diff
@@ -342,6 +348,9 @@ end
 return {
   build_router     = build_router,
   build_api_router = build_api_router,
+  set_check_router_rebuild = function(f)
```
hishamhm: Tiny suggestion: in the `balancer` module, we prefixed with a `_` the functions that were exported for unit-testing purposes only (along with a comment); this helps people reading the module to understand that this is not really a "configurable" bit of the module. I can make the change when merging.

hishamhm: > the general pattern in Lua probably predates Kong itself

Yeah, I believe it comes from the Python convention of treating `_`-prefixed methods in a class as private-by-convention since they have OOP but no visibility rules

hishamhm: > writing this module in the _M form would avoid the need for such a function

but then `check_router_rebuild` would be public. I was sad to even see it escape from the `do` block :) The router logic is delicate, anything to help the reader understand the dependency structure helps!

thibaultcha: Yes, but testability trumps readability imho (the router rebuild is slightly complicated to the neophyte, but isn’t _too_ cluttered, it is still fairly ok to grasp). And adding tests-only public functions makes readability worse than private functions being public (since the former makes the latter possible anyways, plus adds more code).

hishamhm: Sure, don't get me wrong, I don't like test-only functions either. On the readability aspect, I think it's just that whenever I see a public function it adds this extra mental cost of going through "is this used somewhere else? does someone else depends on this?", so I guess I was thinking more about "understandability" than readability.

PR #4101, opened 2018-12-16
perf(runloop) prefetch services when rebuilding router

p0pr0ck5:
### Summary

Trade space for time to improve performance when rebuilding the router by loading the service associated with a given route from a local table, rather than requiring a separate query for each route in the cluster. This reduces the number of queries needed to rebuild the router with N routes from N to roughly N / 1000 * 2.

This is a proof-of-concept commit that needs further discussion (for example, should we cache services in memory over a longer lifetime, rather than using a function-local table for potentially several thousand objects).

### Full changelog

* prefetch services when rebuilding router

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -103,8 +103,35 @@ do
   local router_version
 
 
+  local function gather_services(db)
+    local services_t = {}
+
+    for service, err in db.services:each(1000) do
+      if err then
+        return nil, "could not load services: " .. err
+      end
+
+      if not services_t[service.id] then
+        services_t[service.id] = service
+      end
+    end
+
+    return services_t
+  end
+
+
   build_router = function(db, version)
     local routes, i = {}, 0
+
+    -- trade space for time by pre-fetching services in an efficient
+    -- manner (paging through them, using the same page size as
+    -- as implemented by the Route pager below). this lets us build
+    -- the Route <> Service relationship needed by the router with
+    -- minimal database traffic
+    local services, err = gather_services(db)
+    if err then
+      return nil, err
+    end
 
     for route, err in db.routes:each(1000) do
       if err then
@@ -117,7 +144,8 @@ do
         return nil, "route (" .. route.id .. ") is not associated with service"
       end
 
-      local service, err = db.services:select(service_pk)
+      -- service_pk is a table of { id = <uuid> }
+      local service = services[service_pk.id]
       if not service then
         return nil, "could not find service for route (" .. route.id .. "): " ..
                     err


p0pr0ck5:
Given the comments in https://github.com/Kong/kong/issues/4055#issuecomment-447695888, I'm changing the label to this to 'please review'. I think this is a valid change that should be pushed as part of 1.0 as it involves no additional necessary configuration and no breaking client change.

bungle:
TBH, I think deciding on these these will be `post-1.0.0`. The issue of `500/401` is already solved.

p0pr0ck5:
The goal here was to implement an improvement with as little behavioral change as possible, to be tested in accordance with the traffic patterns noted in #4055. Caching Services potentially imbues non-trivial memory pressure on a given node, depending on the size of the cluster config; I was hoping to have this change tested with as little divergence from current/RC4 code as possible. But definitely, caching object has value if it's not too big a burden to bear in the shm. I've no strong feelings about caching Service objects in-memory vs fetching them, but tbvh I don't have the bandwidth the implement and test a larger changeset at this point. Hopefully someone can review this and help drive it forward into 1.0. I'll be happy to help with a pair of hands where and when i can :)

p0pr0ck5:
>The issue of 500/401 is already solved.

I'm not sure that it is, per #4055. The latest test results from @jeremyjpj0916 for code that is released (available as of #4084; see https://github.com/Kong/kong/issues/4055#issuecomment-447205972) indicated that the constant-500 errors were solved (due to the incorrect mutex handling), but potentially-sporadic 500 responses are still very possible (see https://github.com/Kong/kong/issues/4055#issuecomment-447209585). This PR attempts to solve that problem and that problem alone - future optimizations (timer-based rebuild, configurable consistency, etc) are out of scope here.

bungle:
We already presented a maybe theoretical problem with paging on `routes`. Your PR introduces another (possibly theoretical) issue: think about new route and being added while building a router using paging. You get new `route` in results, but that is not found in your gathered services, so your code will return an error.

p0pr0ck5:
Iterative results. This approach is a pragmatic- send less expensive network traffic where we don't need. In odd cases where _many_ Routes are added near-simultaneously, the router likely breaks as-is. This PR doesn't try to change this; it just makes talking to the DB (and in particular, strongly-consistent C* installations, the behavior of which is easily forgotten during development) less expensive. Nothing more, nothing less.

bungle:
This PR changes that as it makes hitting here more possible:
https://github.com/Kong/kong/pull/4101/files#diff-1f0f137350308286de76ec59740c8edaR150

At least you could add a fallback to normal select in that case.

bungle:
As I am obviously biased as I would merge #4102 (or take caching strategy out of it and merge just that) instead of this, so I ask others to do review and possible merge (I do not have anything against it, though I do not see why it cannot wait).

p0pr0ck5:
This will get handled elsewhere

PR #4117, opened 2018-12-18
fix(plugins) call redis:select only necessary in usage() as well

fffonion:
### Summary

In https://github.com/Kong/kong/pull/3973 the logic of selective call redis:select is implemented for `increment` but not `usage`. This patch applies the same logic to `usage` as well.

### Full changelog

* Use database-based redis connection pool to reduce unnecessary select call in `usage()` call.

### DIFF
--- a/kong/plugins/response-ratelimiting/policies/init.lua
+++ b/kong/plugins/response-ratelimiting/policies/init.lua
@@ -208,8 +208,13 @@ return {
     usage = function(conf, identifier, name, period, current_timestamp)
       local red = redis:new()
       red:set_timeout(conf.redis_timeout)
-
-      local ok, err = red:connect(conf.redis_host, conf.redis_port)
+      -- use a special pool name only if redis_database is set to non-zero
+      -- otherwise use the default pool name host:port
+      sock_opts.pool = conf.redis_database and
+                       conf.redis_host .. ":" .. conf.redis_port ..
+                       ":" .. conf.redis_database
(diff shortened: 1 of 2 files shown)


kong/plugins/response-ratelimiting/policies/init.lua
```diff
@@ -208,8 +208,13 @@ return {
     usage = function(conf, identifier, name, period, current_timestamp)
       local red = redis:new()
       red:set_timeout(conf.redis_timeout)
-
-      local ok, err = red:connect(conf.redis_host, conf.redis_port)
+      -- use a special pool name only if redis_database is set to non-zero
```
thibaultcha: The below condition isn't "non-zero", but "non-nil". Either one is probably fine, there is no harm in always specifying a pool name.

kong/plugins/response-ratelimiting/policies/init.lua
```diff
@@ -208,8 +208,13 @@ return {
...
+      -- otherwise use the default pool name host:port
+      sock_opts.pool = conf.redis_database and
```
thibaultcha: Related to my above comment (when mentioning "always specifying a pool name): if `redis_database` is `nil`, this call to `connect` may use a previously-set `sock_opts.pool` value (since `sock_opts` is an upvalue and we don't clear the field when `conf.redis_database == nil`). That's a bug.

thibaultcha: Oh, my bad, it should be fine actually.

PR #4148, opened 2019-01-03
 feat(stream) nginx directive injections for stream config

bungle:
### Summary

Makes it possible to inject Nginx directives also to the stream configurations with configuration or environment variables:

`KONG_NGINX_STREAM_*` (see `KONG_NGINX_HTTP`)
`KONG_NGINX_SPROXY_*` (see `KONG_NGINX_PROXY`)

E.g.

```
KONG_NGINX_SPROXY_PROXY_BIND = 127.0.0.1 transparent
```

**Some Notes:**

TLDR; using `SPROXY` prefix comes from the limitations of original design introduced prior stream routing.

I didn't enjoy using `SPROXY` but I needed some unique prefix that did not conflict with already used prefixes. I wanted to use prefix such as `STREAM_PROXY` or `STREAM_SERVER` or `PROXY_STREAM` but those would conflict with already taken `STREAM_`  and `PROXY_`  prefixes. It would make code a bit more complicated if we would like to have longest prefix match first semantics.

If I was to rename these prefixes I would probably name them according to Nginx blocks, e.g.:
* `KONG_NGINX_HTTP`
* `KONG_NGINX_HTTP_SERVER` (or perhaps `KONG_NGINX_HTTP_PROXY`)
* `KONG_NGINX_HTTP_ADMIN`
* `KONG_NGINX_STREAM`
* `KONG_NGINX_STREAM_SERVER`  (or perhaps `KONG_NGINX_STREAM_PROXY`)

But that would be a breaking change and also require longest prefix match first semantics.

### DIFF
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -39,6 +39,8 @@ local HEADER_KEY_TO_NAME = {
 local DYNAMIC_KEY_PREFIXES = {
   ["nginx_http_directives"] = "nginx_http_",
   ["nginx_proxy_directives"] = "nginx_proxy_",
+  ["nginx_stream_directives"] = "nginx_stream_",
+  ["nginx_sproxy_directives"] = "nginx_sproxy_",
   ["nginx_admin_directives"] = "nginx_admin_",
 }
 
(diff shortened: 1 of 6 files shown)


hishamhm:
@bungle Agreed with the sentiments. `SPROXY` is kind of ugly but no better non-breaking option comes to mind right now...

kikito:
Tentatively approved, let's see if we can get Travis to pass...

PR #4149, opened 2019-01-03
feat(stream) add support for stream real ip configuration

bungle:
### Summary

A stream module supports real ip when using `proxy_protocol` flag on stream `listen`. This enables the `trusted ips` to work with stream module.

### DIFF
--- a/kong/templates/nginx.lua
+++ b/kong/templates/nginx.lua
@@ -20,7 +20,9 @@ events {
 }
 
 http {
+> if #proxy_listeners > 0 or #admin_listeners > 0 then
     include 'nginx-kong.conf';
+> end
 }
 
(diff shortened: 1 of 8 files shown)


hishamhm:
@bungle We are not building nginx/OpenResty with `--with-stream_realip_module` — I think this would be necessary for this feature to work, right?

hutchic:
if we didn't configure nginx/OpenResty correctly why did this PR pass?

bungle:
@hutchic I think it is because `stream` block is disabled by default.

bungle:
> @bungle We are not building nginx/OpenResty with `--with-stream_realip_module` — I think this would be necessary for this feature to work, right?

We added support just a week ago everywhere.

Issue #3094, opened 2017-12-14
Incorrect routing priority with wildcard host


dsem:
### Summary

I believe the incorrect routing priority applied when there are two APIs configured, one with the `hosts` defined with a preceding wildcard, e.g., `*.domain.com`, and another with a matching host _and_ a `uris` defined, e.g., `hosts: subdomain.domain.com` `uris: /api`.

### Steps To Reproduce

1. Create an API with a wildcard host rule:
    ```
    POST /apis
    {
        "name": "api-1",
        "hosts": "*.domain.com",
        "upstream_url": "upstream_one.com"
    }
    ```
2. Create an API with a host rule that uses a subdomain with the same domain and a uris rule:
    ```
    POST /apis
    {
       "name": "api-2",
        "hosts": "subdomain.domain.com",
        "uris": "/api",
        "upsteam_url": "upstream_two.com"
    }
    ```
3. Make a request to `subdomain.domain.com/api`. This request matches both APIs. Since api-2 defines two rules and api-1 only defines one, the request should be routed to api-2. However, in my tests, it's getting routed to api-1 at `upstream_one.com`

### Additional Details & Logs

I observe this on a new orchestration of Kong following the guide in the kong-dist-kubernetes repo. This is running kong 0.11.2 backed by cassandra.

crejb:
I've encountered a problem which appears to be similar. The scenario is the same except that both APIs have uris configured. The API with the subdomain specified in the host works, but the API with the wildcard specified does not work.

**Steps To Reproduce**

1. Create an API with a wildcard host rule:
```
POST /apis
{
    "name": "api-1",
    "hosts": ["*.domain.com"],
    "uris": ["/api/api1"],
    "upstream_url": "http://upstream_one.com/"
}
```

2. Create an API with a specific subdomain in the host rule:
```
POST /apis
{
    "name": "api-2",
    "hosts": ["subdomain.domain.com"],
    "uris": ["/api/api2"],
    "upstream_url": "http://upstream_two.com/"
}
```

3. Request to `subdomain.domain.com/api/api1` returns `404 no API found with those values`
4. Request to `subdomain.domain.com/api/api2` works

bungle:
I tried to reproduce  what @dsem reported with `1.0.0` without success. So I guess it might have been fixed already.

The issue with @crejb is what I can reproduce, so I mark this as a bug.

thibaultcha:
[...]

PR #4152, opened 2019-01-03
fix(router) wildcard host matching not considered with plain host match

bungle:
### Summary

When two routes are specified with wildcard host and plain host, the wildcard host may be skipped in some cases. See the test suite change that hilights the problem.

### Issues resolved

Fixes issue reported here:
https://github.com/Kong/kong/issues/3094#issuecomment-363659917

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -512,7 +512,8 @@ end
 do
   local matchers = {
     [MATCH_RULES.HOST] = function(route_t, ctx)
-      local host = route_t.hosts[ctx.hits.host or ctx.req_host]
+      local host = route_t.hosts[ctx.req_host] or
+                   route_t.hosts[ctx.hits.host]
       if host then
         ctx.matches.host = host
 
@@ -705,7 +706,8 @@ end
 do
   local reducers = {
     [MATCH_RULES.HOST] = function(category, ctx)
-      return category.routes_by_hosts[ctx.hits.host]
+      return category.routes_by_hosts[ctx.req_host] or
+             category.routes_by_hosts[ctx.hits.host]
     end,
 
     [MATCH_RULES.URI] = function(category, ctx)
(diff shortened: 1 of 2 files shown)


thibaultcha:
It seems like a couple of wildcard routing unit tests are currently failing in the current CI run?

bungle:
@thibaultcha thanks, it should be fixed now.

kong/router.lua
```diff
@@ -986,23 +988,24 @@ function _M.new(routes)
 
     -- host match
 
-    if plain_indexes.hosts[ctx.req_host] then
-      req_category = bor(req_category, MATCH_RULES.HOST)
```
thibaultcha: This change can be highly detrimental in many scenarios.

kong/router.lua
```diff
@@ -512,7 +512,8 @@ end
 do
   local matchers = {
     [MATCH_RULES.HOST] = function(route_t, ctx)
-      local host = route_t.hosts[ctx.hits.host or ctx.req_host]
+      local host = route_t.hosts[ctx.req_host] or
+                   route_t.hosts[ctx.hits.host]
```
thibaultcha: Here, we are ignoring the matched hint, which is thus rendered useless. In which case we could simply remove it. This change can be detrimental in some scenarios.

spec/02-integration/05-proxy/01-router_spec.lua
```diff
@@ -882,6 +882,63 @@ for _, strategy in helpers.each_strategy() do
...
+            ["kong-debug"] = 1,
+          }
+        })
+
+        assert.res_status(200, res)
+
+        assert.equal(routes[2].id,           res.headers["kong-route-id"])
+        assert.equal(routes[2].service.id,   res.headers["kong-service-id"])
+        assert.equal(routes[2].service.name, res.headers["kong-service-name"])
+      end)
```
thibaultcha: This test is a duplicate of [this unit test](https://github.com/Kong/kong/blob/master/spec/01-unit/08-router_spec.lua#L519).

kong/router.lua
bungle: What do you mean? If you remove `route_t.hosts[ctx.hits.host]` the tests will not pass, and if you remove `route_t.hosts[ctx.req_host]` the tests won't pass either. So it is not ignored.

thibaultcha:
Replaced by https://github.com/Kong/kong/pull/4775

Issue #4121, opened 2018-12-21
kong proxy match order


zhangxing406:
NOTE: GitHub issues are reserved for **bug reports only**. For anything else,
please join the conversation in Kong Nation https://discuss.konghq.com/.

Please read the CONTRIBUTING.md guidelines to learn on which channels you can
seek for help and ask general questions:

https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#where-to-seek-for-help

### Summary

SUMMARY_GOES_HERE

i configured some routes,but the true matching rule is different from the docs tells. i think docs says the matching  rule is:
1) full uri match
2) prefix match
3) regex match
for example, two rules:
a: {path:'/api/account'}
b: {path:'/api/.*'}
when requesting '/api/account/getAll',it matched b.

that is to say  the truth is:
1) full uri match
2) regex match
3) prefix match 

i use the default kong config and db type is postgres. hope for your answer.

### Additional Details & Logs

- Kong version (`$ kong 0.14.0`)
- Kong debug-level startup logs (`$ kong start --vv`)
- Kong error logs (`<KONG_PREFIX>/logs/error.log`)
- Kong configuration (the output of a GET request to Kong's Admin port - see
  https://docs.konghq.com/latest/admin-api/#retrieve-node-information)
- Operating system

RobSerafini:
The matching can definitely be complex and counter intuitive and will depend on the exact examples you are using. Please review the following discussions and documentation and reply with the exact routes and requests you are using.

https://discuss.konghq.com/t/a-few-words-on-regex-uris-usage-and-common-pitfalls/506

and 

https://docs.konghq.com/1.0.x/proxy/#matching-priorities

Regards,
Rob

bungle:
I think it is a bug. I will send a PR soon.

thibaultcha:
This is not a bug, but a known missing feature. The feature would be the equivalent of the nginx `location` directive's `^~` modifier, which means "don't evaluate this regex if a previous prefix was already matched. See [ngx_http_core_module's location directive docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#location).

The order of evaluation _is_:
1. full paths
2. prefix paths
3. regex paths

But in the case of multiple matches, the _priority_ order is:
1. full paths
2. regex paths
3. prefix paths

[...]

Issue #3111, opened 2017-12-19
Kong 0.11.1 evaluation order of proxy is not clear between Prefix URI and Regex URI


monday0rsunday:
At this time, Kong 0.11.1 [say that](https://getkong.org/docs/0.11.x/proxy/#evaluation-order): `URI prefixes are always evaluated first `, but it is not clear.

In https://github.com/Kong/kong/issues/2899, @checky has raised that in issue 1

If I understand correctly `URI prefixes are always evaluated first `, following expected result should be correct:

### Setting

- /some/prefix--> API 1
- /some/prefix/{regex1}/some/postfix --> API2

### Expected result

- /some/prefix--> API 1
- /some/prefix/1 --> API1
- /some/prefix/1/some/postfix --> API1
- /some/prefix/1/some/postfix/1 --> API1

But actual result is

### Actual result

- /some/prefix--> API 1
- /some/prefix/1 --> API1
- /some/prefix/1/some/postfix --> API2
- /some/prefix/1/some/postfix/1 --> API2

So I don't know whether it is a bug, or I mis-understand something?

monday0rsunday:
I've found that there is a test case which describe my example

https://github.com/Kong/kong/blob/master/spec/01-unit/010-router_spec.lua#L311

So maybe it is not a bug, but the documentation is not clear, and not completed.

For Prefix URI, there're two possible cases:
- match exactly: /some/prefix
- match prefix: /some/prefix/some/postfix

For Regex URI, there're also two possible cases:
- match exactly: /some/prefix/1/some/postfix
- match prefix: /some/prefix/1/some/postfix/some/postfix_too

As far as I know, we have 3 rules:
1) Prefix URI evaluation is ordered by length
2) Regex URI evaluation is ordered by API creation time
3) ** Prefix URI are always evaluate first**

But from the test case, rule 3 is not enough

thibaultcha:
Hi,

There might indeed be some lack of documentation in this case. The behavior we follow is identical to that of NGINX's `location` block:

1. match plain prefixes by descending order of length
2. match regexes (NGINX does so in the order they are defined in the file, we do so in creation order in the database)
3. if a regex match was found in step 2, override the match from step 1

There is one crucial piece to your precise use-case that we have not yet implemented, which is the `^~` prefix of NGINX's location blocks:

> If the longest matching prefix location has the “^~” modifier then regular expressions are not checked. 

See: https://nginx.org/en/docs/http/ngx_http_core_module.html#location

[...]

PR #4153, opened 2019-01-04
fix(router) regex path matching not considered with exact path match

bungle:
### Summary

This is a sister-PR for #4152, and makes similar adjustments to `path` matching.

The added test case shows the problem.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -522,7 +522,8 @@ do
 
     [MATCH_RULES.URI] = function(route_t, ctx)
       do
-        local uri_t = route_t.uris[ctx.hits.uri or ctx.req_uri]
+        local uri_t = route_t.uris[ctx.req_uri] or
+                      route_t.uris[ctx.hits.uri]
 
         if uri_t then
           if uri_t.is_regex then
@@ -709,7 +710,8 @@ do
     end,
 
     [MATCH_RULES.URI] = function(category, ctx)
-      return category.routes_by_uris[ctx.hits.uri or ctx.req_uri]
+      return category.routes_by_uris[ctx.req_uri] or
+             category.routes_by_uris[ctx.hits.uri]
     end,
 
     [MATCH_RULES.METHOD] = function(category, ctx)
@@ -1007,30 +1009,32 @@ function _M.new(routes)
 
     -- uri match
 
-    if plain_indexes.uris[req_uri] then
-      req_category = bor(req_category, MATCH_RULES.URI)
-
-    else
+    do
+      local prefix_match
       for i = 1, #prefix_uris do
-        if find(req_uri, prefix_uris[i].value, nil, true) == 1 then
-          hits.uri     = prefix_uris[i].value
-          req_category = bor(req_category, MATCH_RULES.URI)
+        prefix_match = find(req_uri, prefix_uris[i].value, nil, true) == 1
+        if prefix_match then
+          hits.uri = prefix_uris[i].value
           break
         end
       end
 
+      local regex_match, _, err
       for i = 1, #regex_uris do
-        local from, _, err = re_find(req_uri, regex_uris[i].regex, "ajo")
+        regex_match, _, err = re_find(req_uri, regex_uris[i].regex, "ajo")
         if err then
           log(ERR, "could not evaluate URI regex: ", err)
           return
         end
 
-        if from then
-          hits.uri     = regex_uris[i].value
-          req_category = bor(req_category, MATCH_RULES.URI)
+        if regex_match then
+          hits.uri = regex_uris[i].value
           break
         end
+      end
+
+      if prefix_match or regex_match or plain_indexes.uris[req_uri] then
+        req_category = bor(req_category, MATCH_RULES.URI)
       end
     end
 
--- a/spec/02-integration/05-proxy/02-router_spec.lua
+++ b/spec/02-integration/05-proxy/02-router_spec.lua
@@ -882,6 +882,119 @@ for _, strategy in helpers.each_strategy() do
       end)
     end)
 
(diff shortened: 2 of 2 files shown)


spec/02-integration/05-proxy/01-router_spec.lua
```diff
@@ -882,6 +882,119 @@ for _, strategy in helpers.each_strategy() do
...
+          }
+
+        }
+      end)
+
+      lazy_teardown(function()
+        remove_routes(routes)
+      end)
+
+      it("prioritizes regexes over prefixes", function()
```
hishamhm: The documented Kong semantics is that it proritizes prefixes over regexes. See https://docs.konghq.com/1.0.x/proxy/#evaluation-order where it says:

> Prefix paths are always evaluated before regex paths.

bungle: @hishamhm 

Please read:
https://github.com/Kong/kong/issues/4121
and
https://github.com/Kong/kong/issues/3111

The docs are a bit misleading. The docs describe the evaluation process, not the matching order. Matching order is for now: exact, regex, prefix. Until we add support for `^~`.

spec/02-integration/05-proxy/01-router_spec.lua
```diff
@@ -882,6 +882,119 @@ for _, strategy in helpers.each_strategy() do
...
+            ["kong-debug"] = 1,
+          }
+        })
+
+        assert.res_status(200, res)
+
+        assert.equal(routes[2].id,           res.headers["kong-route-id"])
+        assert.equal(routes[2].service.id,   res.headers["kong-service-id"])
+        assert.equal(routes[2].service.name, res.headers["kong-service-name"])
+      end)
```
thibaultcha: This test is a duplicate of [this unit test](https://github.com/Kong/kong/blob/master/spec/01-unit/08-router_spec.lua#L452). Btw, it is preferred that router tests be added to the unit test suite pointed above, instead here. This test suite, while it may have _some_ overlap with the above unit suite, relates more to integration of the router within Kong at runtime.

kong/router.lua
```diff
@@ -1007,31 +1009,33 @@ function _M.new(routes)
 
     -- uri match
 
-    if plain_indexes.uris[req_uri] then
-      req_category = bor(req_category, MATCH_RULES.URI)
```
thibaultcha: This change could be highly detrimental in many cases.

kong/router.lua
```diff
@@ -709,7 +710,8 @@ do
     end,
 
     [MATCH_RULES.URI] = function(category, ctx)
-      return category.routes_by_uris[ctx.hits.uri or ctx.req_uri]
+      return category.routes_by_uris[ctx.req_uri] or
+             category.routes_by_uris[ctx.hits.uri]
```
thibaultcha: This change can be detrimental in many cases (that is, instead of taking the hint into consideration, and prioritizing the evaluation of Routes with the hinted path, we now overlook the hint and perform the lookup based on the request path. The hint serves no purpose anymore.

kong/router.lua
```diff
@@ -522,7 +522,8 @@ do
 
     [MATCH_RULES.URI] = function(route_t, ctx)
       do
-        local uri_t = route_t.uris[ctx.hits.uri or ctx.req_uri]
+        local uri_t = route_t.uris[ctx.req_uri] or
+                      route_t.uris[ctx.hits.uri]
```
thibaultcha: Ditto: while the impact is lesser, we still disregard the hint (which serves no purpose anymore) in favor of the request path.

spec/02-integration/05-proxy/01-router_spec.lua
```diff
@@ -882,6 +882,119 @@ for _, strategy in helpers.each_strategy() do
...
+            ["kong-debug"] = 1,
+          }
+        })
+
+        assert.res_status(200, res)
+
+        assert.equal(routes[3].id,           res.headers["kong-route-id"])
+        assert.equal(routes[3].service.id,   res.headers["kong-service-id"])
+        assert.equal(routes[3].service.name, res.headers["kong-service-name"])
+      end)
```
thibaultcha: Note for reference: after testing this on the master branch and seeing it pass, it has become evident to me that this test is to assert we preserve the behavior broken by the removal at line 1011 of the router. Maybe this test could be added to the unit test suite pointed above, regardless of the future of this PR?

spec/02-integration/05-proxy/01-router_spec.lua
```diff
@@ -882,6 +882,119 @@ for _, strategy in helpers.each_strategy() do
...
+            paths      = { "/(path)" },
+          },
+        }
+      end)
+
+      lazy_teardown(function()
+        remove_routes(routes)
+      end)
+
+      it("matches regex path even when there is exact and prefix match", function()
```
thibaultcha: We can see that the second Route in this test's fixture data set defines a host, so I don't think this is a valid test. In fact, the previous test case already establishes that an _exact_ match is favored vs a regex match. This test should probably be about prefix match vs regex matches.

thibaultcha:
It seems to me like we are better off renaming the Route's `priority` attribute and let the user decide for themselves where regex routes have a priority. Not only is it more intuitive, but also avoids the performance degradations noted here, it has also made (almost) consensus in #4121.

bungle:
What I mean that the current router matches route with rule x, then it thinks this is a route to only after further checks finds out that not it doesn't match with rule y of the same route, and then it is 404. Even when there is another route that mathes the request.

thibaultcha:
Replaced by https://github.com/Kong/kong/pull/4775

PR #4162, opened 2019-01-07
fix(api) fix output of records in /plugins/schema

### LINKED ISSUE #4137: Plugin schema inconsistentcy for request-transformer
### Summary

Plugin schema for request transformer contains fields in a nested object as opposed to an array.

### Steps To Reproduce
```
http :8001/plugins/schema/request-transformer
-- stripped out  
      {
            "remove": {
                "fields": {
                    "1": {
                        "body": {
                            "default": {},
                            "elements": {
                                "type": "string"
                            },
                            "type": "array"
                        }
                    },
                    "2": {
                        "headers": {
                            "default": {},
                            "elements": {
                                "type": "string"
                            },
                            "type": "array"
                        }
                    },
                    "3": {
                        "querystring": {
                            "default": {},
                            "elements": {
                                "type": "string"
                            },
                            "type": "array"
                        }
                    },
                    "body": {
                        "default": {},
                        "elements": {
                            "type": "string"
                        },
                        "type": "array"
                    },
                    "headers": {
                        "default": {},
                        "elements": {
                            "type": "string"
                        },
                        "type": "array"
                    },
                    "querystring": {
                        "default": {},
                        "elements": {
                            "type": "string"
                        },
                        "type": "array"
                    }
                },
                "required": true,
                "type": "record"
            }
        },
-- stripped out
```

The `fields` property of schema of an object should be an array and not an object.
Also, as can be seen, the fields are again nested into `1`, `2`, `3`, and are also present in the fields object itself, which seems ambiguous.
 
### Additional Details & Logs

- Kong version (`$ kong version`) 1.0.0
- Kong error logs (`<KONG_PREFIX>/logs/error.log`) none
- Operating system Ubuntu

hishamhm:
* Ensure that plugin schemas return sub-fields of nested records in the same
format as the toplevel `fields` entry (i.e., as an array of single-fields
objects, so that the field order can be retrieved).
* Ensure `fields` in `/plugins/schema` always returns a JSON array,
even when it is empty (as in the prometheus plugin).

This produces an incompatible output from that of 1.0.0, but I'm opening this against `master` because the one in 1.0.0 is arguably broken, so this produces the intended result.

Fixes #4137.

### DIFF
--- a/kong/api/routes/plugins.lua
+++ b/kong/api/routes/plugins.lua
@@ -70,14 +70,52 @@ end
 -- cjson can encode the schema.
 local schema_to_jsonable
 do
-  local function fdata_to_jsonable(fdata)
+  local insert = table.insert
+  local ipairs = ipairs
+  local next = next
+
+  local fdata_to_jsonable
+
+
+  local function fields_to_jsonable(fields)
     local out = {}
-    for k, v in pairs(fdata) do
+    for _, field in ipairs(fields) do
+      local fname = next(field)
+      local fdata = field[fname]
+      insert(out, { [fname] = fdata_to_jsonable(fdata, "no") })
+    end
+    setmetatable(out, cjson.array_mt)
+    return out
+  end
+
+
+  -- Convert field data from schemas into something that can be
+  -- passed to a JSON encoder.
+  -- @tparam table fdata A Lua table with field data
+  -- @tparam string is_array A three-state enum: "yes", "no" or "maybe"
+  -- @treturn table A JSON-convertible Lua table
+  fdata_to_jsonable = function(fdata, is_array)
+    local out = {}
+    local iter = is_array == "yes" and ipairs or pairs
+
+    for k, v in iter(fdata) do
+      if is_array == "maybe" and type(k) ~= "number" then
+        is_array = "no"
+      end
+
       if k == "schema" then
         out[k] = schema_to_jsonable(v)
 
       elseif type(v) == "table" then
-        out[k] = fdata_to_jsonable(v)
+        if k == "fields" and fdata.type == "record" then
+          out[k] = fields_to_jsonable(v)
+
+        elseif k == "default" and fdata.type == "array" then
+          out[k] = fdata_to_jsonable(v, "yes")
+
+        else
+          out[k] = fdata_to_jsonable(v, "maybe")
+        end
 
       elseif type(v) == "number" then
         if v ~= v then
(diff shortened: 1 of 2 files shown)


kong/api/routes/plugins.lua
```diff
@@ -70,14 +70,50 @@ end
...
+  local fdata_to_jsonable
+
+  local function fields_to_jsonable(fields)
+    local out = {}
+    for _, field in ipairs(fields) do
+      local fname = next(field)
+      local fdata = field[fname]
+      insert(out, { [fname] = fdata_to_jsonable(fdata, "no") })
+    end
+    setmetatable(out, cjson.empty_array_mt)
```
thibaultcha: Here, we now that `fields` is always an Array. We should use [cjson.array_mt](https://github.com/openresty/lua-cjson#array_mt) instead, which is is preferred over `empty_array_mt`. The latter is less efficient and, while not documented, is scheduled for deprecation some time in the future since `array_mt` is a better pattern.

kong/api/routes/plugins.lua
```diff
@@ -70,14 +70,50 @@ end
...
+    return out
+  end
+
```
thibaultcha: style: missing line break in between functions declarations

kong/api/routes/plugins.lua
```diff
@@ -94,20 +130,14 @@ do
         out[k] = v
       end
     end
+    if is_array == "yes" or is_array == "maybe" then
+      setmetatable(out, cjson.empty_array_mt)
```
thibaultcha: Consider using `array_mt` if possible. The issues with `empty_array_mt` were partially described in my other comment, but overall, it makes for unpredictable results (or hard to predict) since this will leave tables to be encoded as an Object or an Array solely depending on whether they are empty or not... It could be that the `maybe` branch isn't sure how to encode the table, but we better be certain if possible. The same way, the table could deserve to be encoded as an empty Object in some cases, while we are making it an empty Array with this metatable. And vice-versa if the table _has_ fields.

hishamhm:
@thibaultcha Review addressed! Thanks for the tip about `array_mt`!

spec/02-integration/04-admin_api/04-plugins_routes_spec.lua
```diff
@@ -241,6 +241,19 @@ for _, strategy in helpers.each_strategy() do
             assert.is_table(json.fields)
           end
         end)
+        it("returns nested records as arrays", function()
+          for plugin, _ in pairs(helpers.test_conf.loaded_plugins) do
+            local res = assert(client:send {
+              method = "GET",
+              path = "/plugins/schema/request-transformer",
```
thibaultcha: The plugin name is hard-coded here. The surrounding loop is useless (and the below assertions about `default:[]` need to be conditionally triggered).

hishamhm: welp, I thought I copied only the inside of the block of the previous test, not the for loop

PR #4165, opened 2019-01-08
Always set SSL verify depth

rainest:
### Summary

The existing template will only set `lua_ssl_verify_depth` if `lua_ssl_trusted_certificate` is set also. It should be possible to set verification depth regardless of whether custom CAs are in use. Note that this change will always include this directive, even if using the default (1). This should be the case, as Kong's default depth may diverge from OpenResty's in the future.

### Full changelog

* Modify nginx_kong.lua template to always include `lua_ssl_verify_depth` directive.
* Modify tests to check `lua_ssl_verify_depth` and `lua_ssl_trusted_certificates` settings independently.

### Background

https://github.com/Kong/kong/commit/0b180bfba59e8ac170d92b1f3ec254887869c8de originally added this. It looks like it was originally only intended for use with Cassandra when Cassandra's certs are signed by several intermediates before a custom CA, but the [setting in question applies to any OpenResty TLS socket](https://github.com/openresty/lua-nginx-module#tcpsocksslhandshake). It's necessary to apply this, for example, if the identity provider used with OIDC uses a chain longer than the default, as Azure AD does.


kong/templates/nginx_kong.lua
```diff
@@ -42,8 +42,8 @@ lua_shared_dict kong_cassandra      5m;
 lua_socket_log_errors off;
 > if lua_ssl_trusted_certificate then
 lua_ssl_trusted_certificate '${{LUA_SSL_TRUSTED_CERTIFICATE}}';
-lua_ssl_verify_depth ${{LUA_SSL_VERIFY_DEPTH}};
 > end
+lua_ssl_verify_depth ${{LUA_SSL_VERIFY_DEPTH}};
```
thibaultcha: Will you also apply the same change to the tests configuration template under `spec/`? Thanks!

spec/01-unit/003-prefix_handler_spec.lua
```diff
@@ -181,22 +181,26 @@ describe("NGINX conf compiler", function()
...
       end)
     end)
-    it("does not include lua_ssl_trusted_certificate/lua_ssl_verify_depth by default", function()
+    it("sets lua_ssl_verify_depth", function()
+      local conf = assert(conf_loader(helpers.test_conf_path, {
+        lua_ssl_verify_depth = "2"
+      }))
+      local kong_nginx_conf = prefix_handler.compile_kong_conf(conf)
+      assert.matches("lua_ssl_verify_depth 2;", kong_nginx_conf, nil, true)
+    end)
```
thibaultcha: Will you also add a test that asserts that `lua_ssl_verify_deps is set by default`? Thanks.

PR #4166, opened 2019-01-08
chore(tests) tests renamed to have no sequence gaps

bungle:
### Summary

Renames test suite files and folders to have a sequence without gaps, also renames `new-dao` to just `db`.


bungle:
Question, If we want to merge this, should we merge it to `next` or `master` (currently based on `master`, but I can `rebase` it if needed)?

kikito:
@bungle I think we should do this in both simultaneously. Or first in master and then rebase master into next.

thibaultcha:
I just merged master into next to synchronize those changes.

bungle:
I guess many of the current PRs need to be rebased, but well, that was expected.

PR #4167, opened 2019-01-08
feat(db) add support for postgres schemas

### LINKED ISSUE #4145: 1.0 Kong migration doesn't allow non-public Postgres schemas
### Summary

When trying to start the 1.0 version of Kong on a database using a non-public Postgres schema, it always errors out with: `database needs bootstrap; run 'kong migrations bootstrap'`.

### Steps To Reproduce

1. Start fresh Postgres instance with non-public schema
2. Configure Kong 1.0 instance to point to above database
3. Run `kong migrations bootstrap` (runs successfully)
4. Run `kong start`
5. Receive error claiming you need to bootstrap (step 3)

### Additional Details & Logs

I got so frazzled that I went and tracked the problem down to this line, which says 'public' directly: https://github.com/Kong/kong/blob/master/kong/db/strategies/postgres/connector.lua#L33

I hope I provided enough information on this issue. Thank you for any help.

bungle:
### Summary

Adds support for Postgres schemas. Makes it possible to use other schemas, and not only the `public` schema in Postgres with an explicit configuration option.

Introduces a new configuration parameter:
```
pg_schema
````

The `unspecified` is the default value of the parameter.

### Issues resolved

Fix #4145

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -372,6 +372,10 @@
 #pg_password =                   # The password to authenticate if required.
 #pg_database = kong              # The database name to connect to.
 
+#pg_schema =                     # The database schema to use. If unspecified,
+                                 # Kong will respect the `search_path` value of
+                                 # your PostgreSQL instance.
+
 #pg_ssl = off                    # Toggles client-server TLS connections
                                  # between Kong and PostgreSQL.
 
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -30,7 +30,7 @@ local WARN                          = ngx.WARN
 local SQL_INFORMATION_SCHEMA_TABLES = [[
 SELECT table_name
   FROM information_schema.tables
- WHERE table_schema = 'public';
+ WHERE table_schema = CURRENT_SCHEMA;
 ]]
 local PROTECTED_TABLES = {
   schema_migrations = true,
@@ -148,7 +148,19 @@ local function connect(config)
   end
 
   if connection.sock:getreusedtimes() == 0 then
-    ok, err = connection:query("SET TIME ZONE 'UTC';");
+    if config.schema == "" then
+      local res = connection:query("SELECT CURRENT_SCHEMA AS schema")
+      if res and res[1] and res[1].schema and res[1].schema ~= null then
+        config.schema = res[1].schema
+      else
+        config.schema = "public"
+      end
+    end
+
+    ok, err = connection:query(concat {
+      "SET SCHEMA ",    connection:escape_literal(config.schema), ";\n",
+      "SET TIME ZONE ", connection:escape_literal("UTC"), ";",
+    })
     if not ok then
       setkeepalive(connection)
       return nil, err
@@ -325,10 +337,11 @@ function _mt:infos()
   end
 
   return {
-    strategy = "PostgreSQL",
-    db_name  = self.config.database,
-    db_desc  = "database",
-    db_ver   = db_ver or "unknown",
+    strategy  = "PostgreSQL",
+    db_name   = self.config.database,
+    db_schema = self.config.schema,
+    db_desc   = "database",
+    db_ver    = db_ver or "unknown",
   }
 end
 
@@ -441,12 +454,15 @@ end
 
 
 function _mt:reset()
+  local schema = self:escape_identifier(self.config.schema)
   local user = self:escape_identifier(self.config.user)
+
   local ok, err = self:query(concat {
     "BEGIN;\n",
-    "  DROP SCHEMA IF EXISTS public CASCADE;\n",
-    "  CREATE SCHEMA IF NOT EXISTS public AUTHORIZATION ", user, ";\n",
-    "  GRANT ALL ON SCHEMA public TO ", user, ";\n",
+    "  DROP SCHEMA IF EXISTS ", schema ," CASCADE;\n",
+    "  CREATE SCHEMA IF NOT EXISTS ", schema, " AUTHORIZATION ", user, ";\n",
+    "  GRANT ALL ON SCHEMA ", schema ," TO ", user, ";\n",
+    "  SET SCHEMA ",  self:escape_literal(self.config.schema), ";\n",
     "COMMIT;",
   })
 
@@ -501,7 +517,7 @@ end
 
 
 function _mt:setup_locks(_, _)
-  logger.verbose("creating 'locks' table if not existing...")
+  logger.debug("creating 'locks' table if not existing...")
 
   local ok, err = self:query([[
 BEGIN;
@@ -517,7 +533,7 @@ COMMIT;]])
     return nil, err
   end
 
-  logger.verbose("successfully created 'locks' table")
+  logger.debug("successfully created 'locks' table")
 
   return true
 end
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: kong/conf_loader.lua [hunk 1]
```diff
@@ -106,13 +106,16 @@ local CONF_INFERENCES = {
...
+  pg_database = { typ = "string" },
+  pg_schema = { typ = "string" },
   pg_timeout = { typ = "number" },
   pg_password = { typ = "string" },
   pg_ssl = { typ = "boolean" },
   pg_ssl_verify = { typ = "boolean" },
 
   cassandra_contact_points = { typ = "array" },
   cassandra_port = { typ = "number" },
+  cassandra_keyspace = { typ = "string" },
```
thibaultcha: `string` configuration values were actually left out intentionally. We only specify `string` for some configuration properties when we want to avoid the configuration parser from inferring the value as a number. For example `nginx_worker_processes`. It's cannot hurt to be more specific, but it increases this commit's diff unecessarily.

kong/db/strategies/postgres/connector.lua
```diff
@@ -461,7 +463,13 @@ end
 function _mt:truncate()
   local i, table_names = 0, {}
 
-  for row in self:iterate(SQL_INFORMATION_SCHEMA_TABLES) do
+  local sql = concat {
+    "SELECT table_name\n",
+    "  FROM information_schema.tables\n",
+    " WHERE table_schema = ", self:escape_literal(self.config.schema), ";"
+  }
```
thibaultcha: Since we are reusing this statement below, it'd be both less repetitive and more efficient to do:
```lua
-- top-level upvalue
local SQL_INFORMATION_SCHEMA_TABLES = [[
  SELECT table_name FROM information_schema.tables
    WHERE table_schema = %s;
]]

function _mt:truncate()
  local sql = string.format(SQL_INFORMATION_SCHEMA_TABLES,
                            self:escape_literal(self.config.schema))
end 
```

kong/db/strategies/postgres/connector.lua
```diff
@@ -634,15 +648,37 @@ function _mt:schema_migrations()
 end
 
 
-function _mt:schema_bootstrap(kong_config, default_locks_ttl)
+function _mt:schema_bootstrap(_, default_locks_ttl)
```
thibaultcha: There is no reason for this change since our linter ignores unused arguments. However, this changes makes it harder for future developers (and selves) to remember what this argument is. It is less explicit and thus less self-documented.

spec/02-integration/000-new-dao/01-db_spec.lua
```diff
@@ -50,10 +51,33 @@ for _, strategy in helpers.each_strategy() do
           error("unknown database")
         end
       end)
+
+      if strategy == "postgres" then
+        it("initializes infos with custom schema", function()
+          local conf = utils.table_merge(helpers.test_conf)
```
thibaultcha: Did you mean to use `deep_copy()` maybe? The intent would be clearer (ditto in similar places in this PR).

thibaultcha:
Agreed. Especially since this is how we have been supporting (or rather, our relationship to) Postgres schemas so far. 

The issue with the configuration though is that there is no easy way to unset a value... Unless we introduce some value with a special meaning.

kong/conf_loader.lua
hishamhm: Oh — are unknown entries (i.e. typos) ignored then? I never noticed it!

kong/db/strategies/postgres/connector.lua
bungle: It now uses `CURRENT_SCHEMA` in that query, so it is changed to module level static string.

kong/db/strategies/postgres/connector.lua
bungle: Reverted that change.

hishamhm:
> The issue with the configuration though is that there is no easy way to unset a value... Unless we introduce some value with a special meaning.

The obvious choice, I think, would be to have empty string to mean "don't set the schema at all".

Also, making the schema optional would allow us to backport this to 1.0.x (without exposing the pg_schema conf key) and make Kong 1.0.1 work with non-public schemas.

spec/02-integration/000-new-dao/01-db_spec.lua
bungle: Changed to `deep_copy`.

bungle:
I changed it to support empty `""` or ` nil` values for `pg_schema`. And in that case it tries to lookup one from the connection using `SELECT CURRENT_SCHEMA`.

bungle:
Also changed the default value to be `unspecified`.

kong/db/strategies/postgres/connector.lua
```diff
@@ -148,7 +148,24 @@ local function connect(config)
   end
 
   if connection.sock:getreusedtimes() == 0 then
-    ok, err = connection:query("SET TIME ZONE 'UTC';");
+    if not config.schema or config.schema == "" then
+      local res = connection:query("SELECT CURRENT_SCHEMA AS schema")
+      if res and res[1] and res[1].schema and res.schema ~= null then
```
thibaultcha: Did you mean `res[1].schema ~= null`?

kong/db/strategies/postgres/connector.lua
```diff
@@ -148,7 +148,24 @@ local function connect(config)
   end
 
   if connection.sock:getreusedtimes() == 0 then
-    ok, err = connection:query("SET TIME ZONE 'UTC';");
+    if not config.schema or config.schema == "" then
```
thibaultcha: We assume that `config.schema` is not `nil` everywhere else in this patch. Plus given it is retrieved from the configuration object, it should not in the first place. In the worst case, keeping the `schema = kong_config.pg_schema or ""` in `new()` would be a good idea (e.g. when instantiating by hand in tests or such) and guarantee that we can get rid of this part of the condition.

kong.conf.default
```diff
@@ -371,6 +371,7 @@
 #pg_user = kong                  # The username to authenticate if required.
 #pg_password =                   # The password to authenticate if required.
 #pg_database = kong              # The database name to connect to.
+#pg_schema =                     # The database schema to be used.
```
thibaultcha: Let's clarify what happens when this value is empty (as with the default) in the comment. Something like:

> If unspecified, Kong will respect the `search_path` value of your PostgreSQL instance.

Or something simple like this.

kong/db/strategies/postgres/connector.lua
bungle: Thanks. Fixed.

bungle:
I made a small change, and now it behaves like this:
1. use schema if specified (do not check existence - it might be migration that is creating it)
2. if not specified try to find `CURRENT_SCHEMA` (an existing schema in `search_path` and use it, by default it is `public`)
3. If there is no existing schema, then force it to be `public`.

kong/db/strategies/postgres/connector.lua
bungle: Changed that.

kong.conf.default
bungle: Added.

PR #4169, opened 2019-01-08
perf(pdk) track the size of the Lua table via local Lua variable inst…

moonming:
…ead of table.insert.

### Summary

The `table.insert` is an O(n) operation due to the  lj_tab_len call inside LuaJIT. 
We can avoid it in hot loops for better performance.

### Full changelog

* track the size of the Lua table via local Lua variable instead of table.insert.


thibaultcha:
Hi!

This is really not a hot code path. We only invoke this function once in the entire lifecycle of Kong (in `init_by_lua`) [here](https://github.com/Kong/kong/blob/master/kong/init.lua#L59).

PR #4172, opened 2019-01-09
fix(pdk) prevent memory leaks in 'kong.service.response' & 'kong.ctx'

thibaultcha:
#### Leak 1

Introduced in eb20498, this index
metamethod attachment for response headers tables maintains a references
to itself. This was verified by increasing upvalues, functions, and
table object counts increasing when using this API over time.

The newer implementation is also more efficient and avoids creating
closures at runtime.

#### Leak 2

See `kong.global` module's `set_named_ctx` API. The third argument
(`key`) will be a plugin's configuration table. It is important not to
hold references to those configuration tables, as already observed
across many parts of the codebase.

### DIFF
--- a/kong/pdk/service/response.lua
+++ b/kong/pdk/service/response.lua
@@ -25,10 +25,14 @@ local header_body_log = phase_checker.new(PHASES.header_filter,
                                           PHASES.log)
 
 
-local function headers(response_headers, err)
-  local mt = getmetatable(response_headers)
-  local index = mt.__index
-  mt.__index = function(_, name)
+local attach_mt_index
+
+
+do
+  local resp_headers_orig_mt_index
+
+
+  local function resp_headers_mt_index(t, name)
     if type(name) == "string" then
       local var = fmt("upstream_http_%s", gsub(lower(name), "-", "_"))
       if not ngx.var[var] then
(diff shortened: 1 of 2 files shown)


kong/pdk/service/response.lua
```diff
@@ -25,21 +25,36 @@ local header_body_log = phase_checker.new(PHASES.header_filter,
...
 
-local function headers(response_headers, err)
-  local mt = getmetatable(response_headers)
-  local index = mt.__index
-  mt.__index = function(_, name)
+local attach_mt_index
+
+
+do
+  local resp_headers_orig_mt_index
```
hishamhm: Ahh, so all returns of ngx.resp.get_headers in a Lua state share the same metatable then? Good catch! That explains the leak, as the chain of functions is formed. I mistakenly assumed both the returned table and metatable would be owned by the caller (then the original code wouldn't leak).

Since the metatable is unique, I guess it would be a better idea to produce a new table here, rather than patch the existing one, otherwise we're changing the semantics of `ngx.resp.get_headers`.

thibaultcha: I have considered that, but find this pre-existing approach more efficient.

hishamhm: We can create our own metatable and reuse it, essentially keeping the efficiency. This will prevent surprises in scenarios mixing PDK and raw-ngx plugins.

thibaultcha: Which example do you have in mind that is currently broken between the two?

hishamhm: If some code in a plugin is using raw ngx.resp.get_headers and does not expect the additional logic that we add via the metatable to be there.

thibaultcha: Oh, gotcha. Yeah of course, that makes sense. Somehow yesterday night I overlooked that this overrides the ngx_lua metatable...

PR #4176, opened 2019-01-09
feat(aws-lambda) add AWS China region to plugin schema

wubins:
Add   "cn-north-1" and  "cn-northwest-1" to the list of allowed regions to permit AWS Lambda
functions to be invoked from the AWS China region.

@thibaultcha Thanks.

### DIFF
--- a/kong/plugins/http-log/sender.lua
+++ b/kong/plugins/http-log/sender.lua
@@ -7,7 +7,6 @@ local sender = {}
 
 local ngx_encode_base64 = ngx.encode_base64
 local ERR = ngx.ERR
-print('1')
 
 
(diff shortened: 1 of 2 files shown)


kong/plugins/http-log/sender.lua
```diff
@@ -7,6 +7,7 @@ local sender = {}
 
 local ngx_encode_base64 = ngx.encode_base64
 local ERR = ngx.ERR
+print('1')
```
thibaultcha: Will you please remove this undesired change? Thanks!

wubins: OK

PR #4180, opened 2019-01-09
A fix and a feature for kong.log.inspect

hishamhm:
Two things that came up as I was using `kong.log.inspect` to debug another issue. (I can put it in separate PRs if the feature must go into `next`, but it shouldn't affect any documented behavior — and `kong.log.inspect` is not meant for production — so I think it could safely go into `master` and spare the merge conflict in test cases.)

**Fix:**  In`kong.log.inspect`, we use the `inspect` function as a `tostring` replacement. The `tostring` function ignores additional arguments, but `inspect` doesn't (it reads them as a table of options). This avoid crashes when inspecting a table which happens to have keys understood by `inspect` with incompatible values.

**Feature:** It's not every day I get to post a PR with a picture, so let me indulge :)

window

The `kong.log.inspect` function is designed as a development tool for inspecting Lua objects, many of which are complex tables. It leverages @kikito's `inspect` library, which pretty-prints tables in a nice multi-line indented presentation.

This commits aims to preserve that presentation, by splitting multiline logs done with `kong.log.inspect` into multiple nginx log invocations and pretty-printing them with spaces and a border. This results in much easier to read tables, and allows you to read the entire contents, since the logs have a short limit per log call. (Often, I had to resort to writing a loop and print each key/value pair at once to see everything, defeating the purpose of `kong.log.inspect`.)

The `kong.log.inspect` function is designed as a development-time tool, so there are no production-time concerns about performance or printing over-long messages.


t/01-pdk/02-log/04-inspect.t
```diff
@@ -3,7 +3,7 @@ use warnings FATAL => 'all';
 use Test::Nginx::Socket::Lua;
 use t::Util;
 
-plan tests => repeat_each() * (blocks() * 4);
+plan tests => repeat_each() * (blocks() * 4 + 6);
```
thibaultcha: Here, I moved the `+ 6` inside of the `blocks()` multiplication parenthesis. That is because this plan would become incorrect if we were to update `repeat_each()`.

Issue #4164, opened 2019-01-07
Kong 1.0 first time invoking any unique route path slow loading reported in access_phase


jeremyjpj0916:
### Summary

Have to run in 5 min, so posting this quick with more followup later tonight. 

Basically my last vetting of 1.0 has one last outstanding issue. On 1.0 against my existing C* database nodes I have used with 0.14.1 without major issue it seems Kong on any given unique exposed route that has not been called yet takes a really long time to respond. Opentracing reveals it to be in the access_phase, StatsD shows that it takes that long, and even running a curl client locally from the kong node itself reveals it taking that long. After the first call everything is super fast as expected out of this awesome application.

### Steps To Reproduce

1. Create a few routes and services on Kong 1.0, I have like 2 or so consumers with creds oauth2/jwt too.
2. Try calling a given route for first time(note this proxy has no custom plugins or anything on it, just a bare route and service! I even tried disabling the global zipkin/statsd plugins too as a test but that didn't help either), should see a major delay the first call and every other call fast. I see it taking anywhere around 11-14 seconds on the first call. After we go to 48 milliseconds or less end to end.

Picture for some ammo/proof:
![kongfirstclientrequesttoagivenroute](https://user-images.githubusercontent.com/31913027/50799012-503b3380-12d2-11e9-91bd-b575216fbbc9.png)

Edit - And if it helps, and I call a route protected by oauth/jwt the slowdown occurs too before I get back an "unauthorized". Makes me still think something about the service/router building/local caching??? Note I also never made any change to the config after kong started up, no edits on the fly or anything.

### Additional Details & Logs

- Kong version 1.0
- Kong debug-level Startup/debug logs reveal nothing
- Kong error logs Nothing in the error logs
- Kong configuration Will post tonight, its same as every other github issue I post though
- Operating system: Alpine 3.6

p0pr0ck5:
This is pretty well-known and expected behavior. When kong is first started, most of the db entities (consumers, plugin configs) are not cached in memory, and so require database access. With 1.0 C* read consistency set to a weaker level can reduce database access times, but unless data is ore fetched from the database on node initialization, the behavior noted above is fundamental to Kong’s architecture.

jeremyjpj0916:
[...]

PR #4182, opened 2019-01-10
perf(db) reuse first call of schema_state in translation function

hishamhm:
The translation function for plugins:select_by_cache_key needs to poll
schema_state to decide when to stop using the "migrating" version of the
operation.

Kong checks the schema state once on startup. This commit adds a new
operation, `DB:last_schema_state` to retrieve the last value obtained
(`DB:schema_state` remains the same, re-querying the live database). The
`plugins:select_by_cache_key` translation function first checks
`last_schema_state` before performing the more expensive operation using
`schema_state`.

In the usual scenario where Kong is properly migrated before startup, the use
of `last_schema_state` will be sufficient for disabling the translation
function.

See #4164.

### DIFF
--- a/kong/db/init.lua
+++ b/kong/db/init.lua
@@ -411,8 +411,19 @@ do
   local MigrationHelpers = require "kong.db.migrations.helpers"
   local MigrationsState = require "kong.db.migrations.state"
 
+
+  local last_schema_state
+
+
   function DB:schema_state()
-    return MigrationsState.load(self)
+    local err
+    last_schema_state, err = MigrationsState.load(self)
+    return last_schema_state, err
+  end
+
+
+  function DB:last_schema_state()
+    return last_schema_state or self:schema_state()
   end
 
 
--- a/spec/01-unit/01-db/07-db.lua
+++ b/spec/01-unit/01-db/07-db.lua
@@ -1,0 +1,99 @@
+local mocker = require("spec.fixtures.mocker")
+
+
+local function setup_it_block()
+  mocker.setup(finally, {
+    modules = {
+      {"kong.db.strategies", {
+        new = function()
+          local connector = {
+            infos = function()
+              return {}
+            end,
+            connect_migrations = function()
+              return true
+            end,
+            schema_migrations = function()
+              return {}
+            end,
+            is_014 = function()
+              return { is_014 = false }
+            end,
+            close = function()
+            end,
+          }
+          local strategies = mocker.table_where_every_key_returns({})
+          return connector, strategies
+        end,
+      }},
+      {"kong.db", {}},
+    }
+  })
+end
+
+
+describe("DB", function()
+
+  describe("schema_state", function()
+
+    it("returns the state of migrations", function()
+      setup_it_block()
+
+      local DB = require("kong.db")
+
+      local kong_config = {
+        loaded_plugins = {},
+      }
+      local db, err = DB.new(kong_config, "mock")
+      assert.is_nil(err)
+      assert.is_table(db)
+
+      local state = db:schema_state()
+      assert.is_table(state)
+    end)
+
+  end)
+
+  describe("last_schema_state", function()
+
+    it("returns the last fetched state of migrations", function()
+      setup_it_block()
+
+      local DB = require("kong.db")
+
+      local kong_config = {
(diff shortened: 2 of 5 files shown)


p0pr0ck5:
Just a quick note, I can confirm that this does indeed reduce the latency envelope as [expected](https://github.com/Kong/kong/issues/4164#issuecomment-453225899). Nice :eyes:!

thibaultcha:
Let's add some tests for this new method as well.

hishamhm:
> Let's add some tests for this new method as well.

Sure! Apparently the old method (or the DB module at large) didn't have any explicit tests either, so I added the beginnings of a unit test suite for `kong/db/init.lua` with a test for `last_schema_state`.

I also split the commit into `feat()` and `perf()` sections. It is still a fit for `master` in spite of `feat()` because the feature is internal.

spec/01-unit/01-db/07-db.lua
```diff
@@ -0,0 +1,99 @@
...
+      local state_2 = db:schema_state()
+      assert.is_table(state_2)
+
+      -- schema_state always returns a new object
+      assert(state ~= state_2)
+
+      local last_state_3 = db:last_schema_state()
+
+      -- the latest object created by schema_state is the one cached
+      assert(state_2 == last_state_3)
```
thibaultcha: `assert.equal` is preferred since it produces a graceful busted failure, while this assertion produces an error. Failed and error-ed tests are counted differently in busted.

hishamhm: No, this produces as graceful failure as well (I just checked by changing to `~=`. I wanted to be explicit I'm checking for pointer equality.

Tieske: `equal` does pointer equality, the `same` assertion only compares contents.

In this case a failed assertion will just give an 'assertion failed' error, where the `equal` assertion would provide feedback of the actual contents of the variables.

hishamhm: > feedback of the actual contents of the variables.

And that feedback is misleading, because this test is not about their contents. In the case of two tables that are different objects but structurally equivalent (which is the case I'm trying to catch), it results in a message like

```
spec/01-unit/01-db/07-db.lua:93: Expected objects to be equal.
Passed in:
(table) {
  [legacy_is_014] = false
  [needs_bootstrap] = false
  [new_migrations] = {
    [1] = {
      [migrations] = { ... more }
      [namespace] = 'kong.db.migrations.core'
      [subsystem] = 'core' } } }
Expected:
(table) {
  [legacy_is_014] = false
  [needs_bootstrap] = false
  [new_migrations] = {
    [1] = {
      [migrations] = { ... more }
      [namespace] = 'kong.db.migrations.core'
      [subsystem] = 'core' } } }

stack traceback:
	spec/01-unit/01-db/07-db.lua:93: in function <spec/01-unit/01-db/07-db.lua:59>
```

If one reads the message above, what will they assume? First, they will assume the some deep field that's hidden under `... more` is different. If the tables _are_ structurally different, then one might assume that making them strucuturally identical would fix it, because it's the structures being shown and not the raw pointers, which is not the case.

Compare that to getting an `assertion failed`, going straight to the offending line in the test file and seeing that tables are being compared with `==` (plus, these assertions have comments).

I knew about `assert.equal` and `assert.same`, my choice for `==` was deliberate.

But sure, I can change the comments into assertion messages. :)

Tieske: All opinions here. Best solution would be to change the assertion message and including the table-id in the `equal` assertion in Luassert.

kong/db/init.lua
```diff
@@ -411,8 +411,18 @@ do
   local MigrationHelpers = require "kong.db.migrations.helpers"
   local MigrationsState = require "kong.db.migrations.state"
 
+
+  local last_schema_state
+
+
   function DB:schema_state()
-    return MigrationsState.load(self)
+    last_schema_state = MigrationsState.load(self)
```
bungle: The `MigrationsState.load` may return more than one return value. Do we care about them? Perhaps better error reporting / logging?

hishamhm: Excellent catch! Fixed!

PR #4187, opened 2019-01-11
chore(*) replace internal 'unindent' helper with pl utility

Tieske:
ran into this code while updating something else. There is an external function in Penlight available, so no need to carry our own here.


Tieske:
nvm sorry for the noise, the existing one is smarter, so it's not a replacement.

PR #4192, opened 2019-01-13
Update README.md

Yash-Garg:


thibaultcha:
Hi,

Thank you for the patch. The current wording reads just fine, however.

PR #4195, opened 2019-01-14
fix(conf) kong crashes on start with only stream listener

bungle:
### Summary

To reproduce the issue, start Kong with:
```
KONG_PROXY_LISTEN=off \
KONG_ADMIN_LISTEN=off \
KONG_STREAM_LISTEN=0.0.0.0:9000 kong start --vv
```

This PR fixes the above to work.

### DIFF
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -247,8 +247,8 @@ local function prepare_prefix(kong_config, nginx_custom_template_path)
   end
 
   -- generate default SSL certs if needed
-  if kong_config.proxy_ssl_enabled and not kong_config.ssl_cert and
-     not kong_config.ssl_cert_key then
+  if not kong_config.ssl_cert and not kong_config.ssl_cert_key and
+    (kong_config.proxy_ssl_enabled or kong_config.stream_listeners[1] ~= nil) then
     log.verbose("SSL enabled, no custom certificate set: using default certificate")
     local ok, err = gen_default_ssl_cert(kong_config)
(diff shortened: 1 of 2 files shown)


kong/cmd/utils/prefix_handler.lua
```diff
@@ -247,8 +247,8 @@ local function prepare_prefix(kong_config, nginx_custom_template_path)
   end
 
   -- generate default SSL certs if needed
-  if kong_config.proxy_ssl_enabled and not kong_config.ssl_cert and
-     not kong_config.ssl_cert_key then
+  if (#kong_config.stream_listeners > 0 or kong_config.proxy_ssl_enabled) and
```
james-callahan: `kong_config.stream_listeners[1] ~= nil` is probably a better check.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -247,8 +247,8 @@ local function prepare_prefix(kong_config, nginx_custom_template_path)
...
+    not kong_config.ssl_cert and not kong_config.ssl_cert_key then
```
james-callahan: Maybe put these two checks first? They seem like the more important condition.

kong/cmd/utils/prefix_handler.lua
bungle: Changed.

kong/cmd/utils/prefix_handler.lua
bungle: Changed.

PR #4196, opened 2019-01-14
feat(db) add migration for 1.1.0 & truncate created_at for some entities

### LINKED ISSUE #4183: Kong 1.0 migration bug with timestamp of Upstream entity
### Summary

The `created_at` timestamp is expected to be an int as per the API docs.
The field is an int for upstreams created in Kong 0.14.1 but is returned as a float in Kong 1.0.0.

Please note that the upstreams created in 1.0.0 return the field as int as well, only the upstreams migrated from 0.14.1 to 1.0.0 return the created_at fields as float.
This is likely due to change in precision from millisecond to microsecond.

### Steps To Reproduce

I've detailed the steps to reproduce the issue here:
https://gist.github.com/hbagdi/258e2fdcf878cacfb73565cee37a8d56

kikito:
Fixes #4183


hishamhm:
I believe the plan is for patch releases to not have migrations; however, this apparently fixes a mistake in the 1.0.0 migration. @thibaultcha, opinion on the matter?

bungle:
Schema could also enforce such rule(?)

hishamhm:
@bungle @hbagdi sounds like the best approach indeed. @kikito, could you rebase this to `next` and open a new one with the via-code fix for `master`?

thibaultcha:
> I believe the plan is for patch releases to not have migrations; however, this apparently fixes a mistake in the 1.0.0 migration. @thibaultcha, opinion on the matter?

In the very early days, the "no migrations on patch releases" policy originated from the flakiness of our previous migrations framework. When this decision was made, we established that if the migrations were less disruptive (no downtime) and more robust, we could ship patch releases including migrations.

If we feel like the new migrations framework fulfills those requirements, then I see no reason not to start shipping them in patch releases. Two concerns are still on my mind though:

1. The migrations are not automatic; one still needs to run `kong migrations up`, and we have never trained our user-base to doing so for minor releases. This migration could come up as a surprise and maybe restrict the adoption of 1.0.1.
2. I can see a distinction between "large" migrations such as the plugins one from 0.14 to 0.15 (requiring B/G deployments, cluster mutex lock, proxy path backwards-compatibility work on our side, and careful scripting or execution from our users), and small, idempotent migrations such as this one, which could be executed automatically, by any number of nodes, with little harm. In a world where we decide to ship migrations in patch releases, those better be as easy and automated as possible (on top of a robust framework which avoids migration errors and DB corruption at all costs), and my thinking so far would be to introduce such "automated" (idempotent, no-teardown) migrations (they can be labelled as such in their schema definition).

kong/db/migrations/core/003_1.0.0_to_1.0.1.lua
```diff
@@ -0,0 +1,23 @@
...
+
+      UPDATE consumers SET created_at = DATE_TRUNC('seconds', created_at);
+      UPDATE plugins   SET created_at = DATE_TRUNC('seconds', created_at);
+      UPDATE upstreams SET created_at = DATE_TRUNC('seconds', created_at);
+      UPDATE targets   SET created_at = DATE_TRUNC('milliseconds', created_at);
+
+    ]],
+
+    teardown = function(connector, helpers)
+    end,
```
thibaultcha: Just as a general guideline for the future: there would be no need to specify the `teardown` functions if those are empty. It's an optional attribute of the migrations schema.

hishamhm:
@thibaultcha Makes sense!

> This migration could come up as a surprise and maybe restrict the adoption of 1.0.1.

I share the feeling. I think we should avoid migrations for 1.0.1 if we can. In this particular case, it should be easy enough to `math.floor` all values of type `"integer"` in the schema library and avoid the migration.

kong/db/migrations/core/init.lua
```diff
@@ -2,4 +2,5 @@ return {
   "000_base",
   "001_14_to_15",
   "002_15_to_1",
+  "003_1.0.0_to_1.1.0",
```
bungle: What convention should we follow?
1. `003_1.0.0_to_1.1.0`
2. `003_1_to_1_1`
3. `003_1_to_1.1`
4. `003_1_to_11`

I think when I did a PR that added first `002_15_to_1`, I named it as `002_15_to_10`, but someone renamed it to `002_15_to_1`. My reasoning was to always use two digits, and never semver beyond 9 (and just leave `001` and `002` as exception to this rule). Did someone have an idea to this?

thibaultcha: 5. `003_100_to_110` is the most precise, concise, and future-proof notation (e.g. introducing migrations in future patch releases as elaborated in my previous comment).

> and never semver beyond 9

No big deal to do so. Besides, if mixing up digits is a concern, we should consider using an integral value in these file names (like [kong.version_num](https://github.com/Kong/kong/blob/master/kong/global.lua#L12-L15)).

I used `1` and not `10` in the 1.0 release so as to avoid confusion with `0.10`, but if we use `110`, and if we agree to this convention, this is not really a concern anymore.

hishamhm: Let's do `110` then. This is just for human visual inspection. Nobody's going to get `4312` confused between `4.31.2` or `4.3.12` given that the files are ordered in sequence with prefixes like `004_` anyway.

kikito: Ok, doing `110`.

When we get to Kong v10 migrations will probably have changed again anyway :)

PR #4197, opened 2019-01-14
feat(oauth2) authenticate consumer on token generate

jeremyjpj0916:
### Summary

As from here: https://discuss.konghq.com/t/how-to-set-kong-authenticated-consumer-context-during-oauth2-bearer-token-generation/2514/2

Basically this will enable logging plugins to know who is generating tokens when they call the "/oauth2/token" path, also provides early mover benefit of doing some caching at the time of token generation as opposed to only when the token gets passed to proxy as a Bearer token header for actual service invocation.

See log example below and how Consumer is now populated(which before we had no knowledge of who it was that just generated an oauth2 token in logs):
image

### Full changelog

* New function(load_credential_and_consumer_into_memory_from_credential) to prevent code duplication(in token evaluate logic vs token generate logic) https://github.com/jeremyjpj0916/kong/blob/447af55761fb9342ad9e27b529f614409d80ed13/kong/plugins/oauth2/access.lua#L75

### DIFF
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -50,6 +50,49 @@ local function internal_server_error(err)
   return kong.response.exit(500, { message = "An unexpected error occurred" })
 end
 
+local function load_oauth2_credential_into_memory(credential_id)
+  local result, err = kong.db.oauth2_credentials:select { id = credential_id }
+  if err then
+    return nil, err
+  end
+
+  return result
+end
+
+local function load_consumer_into_memory(consumer_id, anonymous)
+  local result, err = kong.db.consumers:select { id = consumer_id }
+  if not result then
+    if anonymous and not err then
+      err = 'anonymous consumer "' .. consumer_id .. '" not found'
+    end
+
+    return nil, err
+  end
+
+  return result
+end
+
+local function load_credential_and_consumer_into_memory_from_credential(credential)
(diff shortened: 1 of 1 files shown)


thibaultcha:
@jeremyjpj0916 The error is visible here: https://travis-ci.org/Kong/kong/jobs/479492527#L2438.

jeremyjpj0916:
@thibaultcha Oh interesting, I scrolled to the bottom and saw the error but did not notice that detail.
```
spec/03-plugins/25-oauth2/03-access_spec.lua:1023: attempt to index field 'authenticated_consumer' (a nil value)
```
So does this point to something underlying wrong with the test suite specifically then? Because I am testing locally now just overwriting that oauth2 access.lua against 1.0 and its doing what it should (calling kong.client.authenticate(consumer, cred) under the hood and setting it during token generation calls so my http log plugin has that info available to it).

jeremyjpj0916:
Hmm even after I pushed dropping my unit test that caused a nil looks like the ci/cd unit tests failed due to nothing directly with my PR. Flakiness in the test suite maybe? I removed my unit test addition because it seems unsupported to do what is needed for checking ngx.ctx to ensure the authorized consumer is set in the context of the token generation event. Checked other spec/03-plugins for examples and none leverage pdk.client.* or ngx.ctx for checking the context of what is set, only req/resp/header evaluations generally or db/cache related checks.

PR #4199, opened 2019-01-14
feat(schema) verify the validity of plugin schemas

hishamhm:
### feat(schema) verify the validity of plugins schema.lua files

Introduces a "MetaSubSchema" for validating subschemas.

For Kong 1.0.x, this will only raise a warning log and not prevent plugins accepted by Kong 1.0.0 from loading.

To add the warning log, I also needed this:

### refactor(*) ensure kong global is always available

The `kong` global, which offers the PDK APIs such as `kong.log`, was available from some parts of the Kong core and not from others. Because of this, we would often end up recoursing to `ngx` APIs instead of the improved `kong` ones.

This refactor loads the `kong` global for two environments where it wasn't being loaded, namely the test files and the `bin/kong` CLI.

Because the Kong core initializes a PDK instance and that loaded `resty.core` (which patches `ngx` fields) and `kong.globalpatches` also patches `ngx` fields for the CLI and tests, it was necessary to move the loading of `resty.core` earlier than the PDK initialization. (An `assert` was kept to verify that the PDK always runs with `resty.core` enabled.)

### DIFF
--- a/spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
+++ b/spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
@@ -1,4 +1,5 @@
 local Schema = require "kong.db.schema"
+local helpers = require "spec.helpers"
 local MetaSchema = require "kong.db.schema.metaschema"
 
 
@@ -145,6 +146,7 @@ describe("metaschema", function()
         { a = { type = "number" } },
         { b = { type = "number" } },
       },
+      primary_key = { "a" },
       entity_checks = {
         { custom_entity_check = {
             field_sources = { "a" },
@@ -155,9 +157,8 @@ describe("metaschema", function()
         },
       }
     }
-    local ok, errs = MetaSchema:validate(s)
-    assert.falsy(ok)
-    assert.truthy(errs)
+    local ok = MetaSchema:validate(s)
+    assert.truthy(ok)
   end)
 
   it("demands a primary key", function()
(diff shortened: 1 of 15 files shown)


hishamhm:
Adding this to `next` with an actual error raised is the end-goal. However, I think adding this to `master` (with a warning only, to ensure no behavioral change) could be useful to aid plugin writers who are porting plugins. Ping @thibaultcha

spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
```diff
@@ -501,3 +502,219 @@ describe("metaschema", function()
...
+describe("metasubschema", function()
+  it("rejects a bad schema", function()
+    local s = {
+      name = "bad",
+      fields = {
+        { foo = "bar", },
+      },
+      primary_key = { "foo" },
+    }
+    assert.falsy(MetaSchema.MetaSubSchema:validate(s))
```
kikito: I think we should test the error messages as well (not the whole returned table, maybe just the relevant entry - `assert.equals("some error", err.fields.foo)`

spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
```diff
@@ -501,3 +502,219 @@ describe("metaschema", function()
...
+      name = "bad",
+      fields = {
+        { foo = { type = "number" }, },
+      },
+      primary_key = { "foo" },
+      entity_checks = {
+        foo = { "bar" },
+      }
+    }
+    assert.falsy(MetaSchema.MetaSubSchema:validate(s))
```
kikito: Same: I think the error message should be tested too.

spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
```diff
@@ -501,3 +502,219 @@ describe("metaschema", function()
...
+      primary_key = { "foo" },
+      entity_checks = {
+        { only_one_of = { "a", "b" },
+          at_least_one_of = { "c", "d" },
+        },
+      }
+    }
+    local ok, errs = MetaSchema.MetaSubSchema:validate(s)
+    assert.falsy(ok)
+    assert.truthy(errs)
```
kikito: Let's test the `errs` a bit more. Again, not necessarily the whole returned table.

spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
```diff
@@ -501,3 +502,219 @@ describe("metaschema", function()
...
+            type = "array",
+            elements = {
+              { foo = "bar", },
+            }
+          }
+        }
+      },
+      primary_key = { "foo" },
+    }
+    assert.falsy(MetaSchema.MetaSubSchema:validate(s))
```
kikito: Errors testing required.

spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
```diff
@@ -501,3 +502,219 @@ describe("metaschema", function()
...
+  end)
+
+  it("rejects a bad schema matching validators and types", function()
+    local s = {
+      name = "bad",
+      fields = {
+        {
+          foo = {
+            type = "array",
+            -- will cause error because `uuid` must be used with `strings`
```
kikito: Transform this comment into an assert over `errs` at the end.

hishamhm: Not needed given the updated error check.

PR #4203, opened 2019-01-15
fix(core) http2 support

bungle:
### Summary

The `http2` did not work with Kong `1.0.0`, and it was also reported here:
https://discuss.konghq.com/t/kong-http2-support/2600/3

### DIFF
--- a/kong/runloop/mesh.lua
+++ b/kong/runloop/mesh.lua
@@ -11,6 +11,7 @@ local cluster_ca_tools = require "kong.tools.cluster_ca"
 
 
 local encode_base64 = ngx.encode_base64
+local http2_enabled
 
 
 local function simple_mesh_alpn_select(ssl, protos, mesh_alpn)
@@ -33,8 +34,9 @@ local function nginx_mesh_alpn_select(ssl, protos, mesh_server_ssl_ctx, mesh_alp
       -- https://github.com/openssl/openssl/issues/1652#issuecomment-384660673
       ssl:setVerify(mesh_server_ssl_ctx:getVerify()) -- to set e.g. VERIFY_FAIL_IF_NO_PEER_CERT
       return v
-    -- elseif v == "h2" -- TODO: figure out if current proxy listener directive has http2 allowed
-    elseif v == "http/1.1" then
+    elseif v == "http/1.1" or (v == "h2" and http2_enabled) then
+      -- TODO: Figure out if current proxy listener directive has http2 allowed.
+      -- For now this is better than not having http2 support at all.
       return v
     end
   end
@@ -60,6 +62,13 @@ local function init()
   -- This should run in init phase (in master, not worker)
 
   ngx.log(ngx.INFO, "initialising cluster ca...")
+
+  for _, listener in ipairs(singletons.configuration.proxy_listeners) do
+    if listener.http2 then
+      http2_enabled = true
+      break
+    end
+  end
 
   local ca_cert
   local node_cert


bungle:
tests are missing still.

thibaultcha:
We have a need for an http2 client in our test suite (AWS lambda with http2 fix from a few months ago, this fix, etc...). I told @travenichka to assume that curl is installed during the tests a few months back, but no tests were added in the end. 
Today, there exists an http2 lua client recently published on the OpenResty mailing list. We could install it as a testing dependency?

bungle:
Also the comment says `figure out if current proxy listener directive has http2 allowed` this PR does not do that yet as what it does is that it checks if any proxy listened has http2 allowed.

hutchic:
manually testing this looks good
```
docker run -t --network=host --rm badouralix/curl-http2 -I https://localhost:8443 --insecure
HTTP/2 404
date: Tue, 15 Jan 2019 21:06:49 GMT
content-type: application/json; charset=utf-8
content-length: 48
server: kong/1.0.0
```

kong/runloop/mesh.lua
```diff
@@ -139,7 +147,11 @@ local function init()
     -- We (ab)use the OpenSSL hostname callback for doing this
     kong.default_client_ssl_ctx:setHostNameCallback(client_hostname_callback, mesh_client_ssl_ctx, mesh_alpn)
     if ngx.config.subsystem == "http" then
-      kong.default_client_ssl_ctx:setAlpnProtos { mesh_alpn, "http/1.1" }
+      if http2_enabled then
+        kong.default_client_ssl_ctx:setAlpnProtos { mesh_alpn, "h2", "http/1.1" }
```
bungle: @james-callahan should this change be removed?

kong/runloop/mesh.lua
```diff
@@ -33,8 +34,8 @@ local function nginx_mesh_alpn_select(ssl, protos, mesh_server_ssl_ctx, mesh_alp
       -- https://github.com/openssl/openssl/issues/1652#issuecomment-384660673
       ssl:setVerify(mesh_server_ssl_ctx:getVerify()) -- to set e.g. VERIFY_FAIL_IF_NO_PEER_CERT
       return v
-    -- elseif v == "h2" -- TODO: figure out if current proxy listener directive has http2 allowed
-    elseif v == "http/1.1" then
+    elseif v == "http/1.1" or (v == "h2" and http2_enabled) then
+      -- TODO: figure out if current proxy listener directive has http2 allowed
```
bungle: @james-callahan as this runs on `init` there is no way to check `current`, or can you elaborate what you mean about `current`?

james-callahan: This doesn't run on init, it is called for each incoming connection (this is a callback set to be run for each TLS negotiation)
The question is how do we map from e.g. an `ssl` object to the related proxy listener.

kong/runloop/mesh.lua
james-callahan: no? I think it looks okay.

Need to test with multiple proxys (one with http2, one without) going to different upstreams (one supporting http2, one not)

bungle: proxy module does not support http2, afaik.

kong/runloop/mesh.lua
bungle: oh...  but I tried it with http 1.1 and http 2 and it seems to work still.

kong/runloop/mesh.lua
james-callahan: ah yes indeed. as nginx doesn't support http2 to the upstream then we shouldn't advertise support here.

bungle: Removed that.

bungle:
@hishamhm / @kikito If we want this for `1.0.1` (which I think we should) the tests need to be added later when we have a working `http2` client available. I already added task for that.

kong/runloop/mesh.lua
hishamhm: what happens if we just do `v == "http/1.1" or v == "h2"` here without the `http2_enabled` check?

hishamhm: (asking as I would assume that if `http2` is not enabled then it would probably fail somewhere else and letting both pass here would be harmless... I guess?)

hishamhm: I was explained out-of-band that this would result in degraded behavior

usernkey:
This same or similar issue might have reappeared. https://discuss.konghq.com/t/kong-ingress-issue-with-grpc-service/6448/3

PR #4204, opened 2019-01-15
fix(db) floor values for integer fields on the schema

kikito:
Related with #4183


bungle:
Interesting though that it failed on tests.

PR #4222, opened 2019-01-19
feat(basic-auth) haveibeenpwned integration

p0pr0ck5:
### Summary

This commit integrates the PwnedPasswords API from the haveibeenpwned.com service, with the basic-auth plugin. Passwords sent to the Admin API as part of create/update/upsert operations on a given credential are validated for presence within the haveibeenpwned data set. If a password hash is found to be present, the API request is rejected.

For more information on haveibeenpwned see

  https://haveibeenpwned.com/Passwords

and

  https://haveibeenpwned.com/API/v2#PwnedPasswords

### Full changelog

* integrate haveibeenpwned API logic with basic-auth credentials Admin API endpoints
* add mock haveibeenpwned passwords range endpoint
* override basic-auth meta in `spec/03-plugins/10-basic-auth/04-invalidations_spec.lua` to avoid calling the production haveibeenpwned service in unrelated tests


p0pr0ck5:
Turning this on unconditionally seems like a bit of a big hammer; I'm not sure of the best approach here (generate a warning when an insure password is used? toggle this integration via admin api/config file? simply don't use this?). I'll admit this is largely a toy commit (good way to kill an hour ;) ), but I do think it has utility.

PR #4230, opened 2019-01-23
break the plugins iterator as soon as possible

ms2008:
If the delayed_response has triggered, we should  break the plugins iterator as soon as possible.


bungle:
We need to run through plugins iterator to get list of plugins for the request. The line there:
```
if not ctx.delayed_response then
```

means that we will not run the plugins (if one there was one short circuiting), we just iterate. So I will close this. Please reopen if I got something wrong about it. Yes, I agree it is quite non-obvious what the code does here and why. Thank you for proposal though.

ms2008:
Indeed, nvm sorry for the noise.

PR #4231, opened 2019-01-23
fix(core) pass the context table to mesh.rewrite

bungle:
### Summary

Originally done here:
https://github.com/Kong/kong/commit/40a073ea57155c5e99e717724dae86dbc0f02508#diff-1f0f137350308286de76ec59740c8eda

Bug introduced here (the next commit after ^):
https://github.com/Kong/kong/commit/ba3f4e46d122f1e134d6bff4024e9a9af8f1baeb#diff-1f0f137350308286de76ec59740c8eda

Yes, this is quite severe bug.


thibaultcha:
Good catch.

PR #4243, opened 2019-01-25
tests(*) add tests for run-on support on mesh and non-mesh (tls/plain)

bungle:
### Summary

Adds more tests for plugins triggering on supported subsystems.

The Travis tests will fail on this PR with Postgres as it highlights a bug with Nginx `co-sockets` on `preread` phase.


spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
```diff
@@ -16,7 +16,7 @@ for _, strategy in helpers.each_strategy() do
     local db
     local bp
 
-    lazy_setup(function()
+    setup(function()
```
hishamhm: is this change necessary for proper operation or was it temporary debugging?

spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
```diff
@@ -1002,120 +1006,1372 @@ for _, strategy in helpers.each_strategy() do
...
+
+            --
+
+            local first_on_first = assert(db.routes:insert {
+              hosts     = { "first-on-first.org" },
+              protocols = { "http" },
+              service   = mock_service,
+            })
+
+            assert(bp.plugins:insert {
```
kikito: The `assert` around `bp.*:insert` are not necessary. The blueprints throw an error if an `insert` fails.

spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
bungle: oh, nice catch.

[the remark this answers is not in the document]
bungle: Yes, probably just some testing leftover

PR #4248, opened 2019-01-28
Feat/protocols

kikito:
This PR adds a protocols field to the Plugin entity, and sets its default for all plugins. In addition to this:

* It filters out plugins on the plugins iterator according to their `protocols` field
* It throws an error when trying to associate a Plugin to a Route which is not compatible, `protocols`-wise, or to a Service with no compatible routes.

### DIFF
--- a/kong/db/dao/plugins.lua
+++ b/kong/db/dao/plugins.lua
@@ -22,6 +22,96 @@ local ngx_WARN = ngx.WARN
 local ngx_DEBUG = ngx.DEBUG
 
 
+local function has_a_common_protocol_with_route(plugin, route)
+  local plugin_prot = plugin.protocols
+  local route_prot = route.protocols
+  -- plugin.protocols and route.protocols are both sets provided by the schema
+  -- this means that they can be iterated as over an array, and queried as a hash
+  for i = 1, #plugin_prot do
+    if route_prot[plugin_prot[i]] then
+      return true
+    end
+  end
+end
+
+
+local function has_a_compatible_route_in_service(self, plugin, service_pk)
+  for route, err, err_t in self.db.routes:each_for_service(service_pk) do
+    if not route then
+      return nil, err, err_t
+    end
+
+    if has_a_common_protocol_with_route(plugin, route) then
+      return true
+    end
+  end
+end
+
+
+local function check_protocols_match(self, plugin)
+  if type(plugin.protocols) ~= "table" then
+    return true
+  end
+
+  if type(plugin.route) == "table" then
+    local route = self.db.routes:select(plugin.route) -- ignore error
+    if route and not has_a_common_protocol_with_route(plugin, route) then
+      local err_t = self.errors:schema_violation({
+        protocols = "must match the associated route's protocols",
+      })
+      return nil, tostring(err_t), err_t
+    end
+  end
+
+  if type(plugin.service) == "table" then
+    if not has_a_compatible_route_in_service(self, plugin, plugin.service) then
+      local err_t = self.errors:schema_violation({
+        protocols = "must match the protocols of at least one route " ..
+                    "pointing to this Plugin's service",
+      })
+      return nil, tostring(err_t), err_t
+    end
+  end
+
+  return true
+end
+
+
+function Plugins:insert(entity, options)
+  local ok, err, err_t = check_protocols_match(self, entity)
+  if not ok then
+    return nil, err, err_t
+  end
+  return self.super.insert(self, entity, options)
+end
+
+
+function Plugins:update(primary_key, entity, options)
+  local rbw_entity = self.strategy:select(primary_key, options) -- ignore errors
+  if rbw_entity then
+    entity = self.schema:merge_values(entity, rbw_entity)
+  end
+  local ok, err, err_t = check_protocols_match(self, entity)
+  if not ok then
+    return nil, err, err_t
+  end
+
+  return self.super.update(self, primary_key, entity, options)
+end
+
+
+function Plugins:upsert(primary_key, entity, options)
+  local rbw_entity = self.strategy:select(primary_key, options) -- ignore errors
+  if rbw_entity then
+    entity = self.schema:merge_values(entity, rbw_entity)
+  end
+  local ok, err, err_t = check_protocols_match(self, entity)
+  if not ok then
+    return nil, err, err_t
+  end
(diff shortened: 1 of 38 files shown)


kong/db/dao/plugins.lua
```diff
@@ -4,7 +4,7 @@ local utils = require "kong.tools.utils"
 local Entity = require "kong.db.schema.entity"
 local DAO = require "kong.db.dao"
 local MetaSchema = require "kong.db.schema.metaschema"
-
+local Set = require "pl.Set"
```
Tieske: I would not be using the Penlight set object if I could prevent it. Since here you don't use the complex methods the object has, I'd simply write my own code here

kikito: That was actually my first implementation. Then I thought "man, someone is going to look at my `create_set` function and ask why I haven't used one of the `utils` or penlight functions instead of reinventing the wheel" xD

hishamhm: I'll go one step further: objects of type `"set"` in our schemas already ship with a metatable for set-style access. :)

However, I just double checked and their implementation is O(n). So I pushed a branch to fix this. You may want to cherry-pick this commit into your branch:

https://github.com/Kong/kong/commit/5e94c0ff0391c841855f7f1341775d9aaa991bd1

spec/02-integration/03-db/03-plugins_spec.lua
```diff
@@ -60,6 +72,23 @@ for _, strategy in helpers.each_strategy() do
...
                       [["},consumer=null}']], err_t.message)
         end)
+
+        it("returns an error when inserting mismatched plugins", function()
+          local plugin, _, err_t = db.plugins:insert({ name = "key-auth",
+                                                       protocols = { "http" },
+                                                       route = { id = route.id },
+                                                     })
+          assert.is_nil(plugin)
+          assert.equals(err_t.fields.protocols, "Plugin protocols must match the associated route's protocols")
```
hishamhm: suggestions on the error message:
* `Plugin protocols` is redundant given its context

spec/02-integration/03-db/03-plugins_spec.lua
```diff
@@ -60,6 +72,23 @@ for _, strategy in helpers.each_strategy() do
...
+
+          local plugin, _, err_t = db.plugins:insert({ name = "key-auth",
+                                                       protocols = { "tcp" },
+                                                       service = { id = service.id },
+                                                     })
+          assert.is_nil(plugin)
+          assert.equals(err_t.fields.protocols,
+                        "Plugin protocols must match at least one of the service's route's protocols")
```
hishamhm: "service's route's protocols" is a mouthful, I don't think it's usual form to double "'s" like this to "nest" ownership. I suggest `"must match the protocols of at least one route pointing to this Plugin's service"`.

To confirm, the intent of this verification is for user-friendliness only and not actual _validation_, right? Because if I follow correctly, I can add `key-auth` to route `R` with { protocols = {"http"} }` and then PATCH `R` to `{ protocols = {"tcp"} }` and the plugin will remain plugged to it, breaking the invariant that plugins are only added to routes of the correct protocol.

I mean, there is no actual _harm_ if that happens right? (Other than having a plugin attached to route `R` that will never be triggered) — I'm not advocating to add checks in the Routes and Services DAOs to verify that the constraint is never broken, just trying to understand whether we really care if the invariant is held or not.

kong/db/dao/plugins.lua
kikito: Ok, since the schema sets already provide the functionality, I disposed of penlight and used them directly. I also cherry-picked the perf change.

spec/02-integration/03-db/03-plugins_spec.lua
kikito: I updated the message as requested.

> To confirm, the intent of this verification is for user-friendliness only and not actual validation, right?

Yes, the compatibility errors are 100% UX.

> I mean, there is no actual harm if that happens right?

Nothing bad should happen.

* First of all, the plugin should not run on the same Lua space where the route is matched, given the first `perf` commit on this PR, which filters them out.
* Secondly, even if they somehow managed to run together, stream plugins have different phases than http plugins.

kong/db/dao/plugins.lua
```diff
@@ -22,6 +21,99 @@ local ngx_WARN = ngx.WARN
...
+      return true
+    end
+  end
+end
+
+
+local function has_at_least_one_compatible_route_in_service(self, plugin, service_pk)
+  local routes, err_t, offset
+  repeat
+    routes, err_t, offset = self.db.routes:page_for_service(service_pk, nil, offset)
```
hishamhm: any particular reason for not using `each_for_service` with a `for` loop instead? I saw the repeat/until and had to do a double-take

kikito: Because I'm a dork, that's why :P . Rewrote using `each_for_service`

kong/db/migrations/core/003_100_to_110.lua
```diff
@@ -7,14 +7,26 @@ return {
...
 
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "plugins" ADD "protocols" TEXT[];
+      EXCEPTION WHEN DUPLICATE_COLUMN THEN
+        -- Do nothing, accept existing state
+      END;
+      $$;
+
+      CREATE INDEX IF NOT EXISTS "plugins_protocols_idx" ON "plugins" ("protocols");
```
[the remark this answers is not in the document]
kikito: Agreed. Removed those indexes.

spec/02-integration/03-db/03-plugins_spec.lua
```diff
@@ -6,14 +6,25 @@ local UUID_PATTERN = "%x%x%x%x%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%x
...
         "services",
         "plugins",
       })
+
+    end)
+
+    before_each(function()
+      assert(db:truncate("routes"))
+      assert(db:truncate("services"))
+      assert(db:truncate("plugins"))
```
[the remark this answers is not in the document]
hishamhm: I tested this locally and force-pushed the change. Test times (pg + C*) went from 5s to 1.5s :+1:

PR #3821, opened 2018-10-02
renaming internal `balancer_data` structure

Tieske:
The initial attempt to refactor the request construction (see #3744) is now phased into multiple PR's

First one being the refactor of the request path building (#3780 ), this is the second one where the internal structures are renamed and expanded. The routing properties will now be set AFTER the access phase.

PR #4253, opened 2019-01-28
fix(request-transformer) handle Host header correctly

### LINKED ISSUE #4252: Kong 1.0.2 Request Transformer forces "preserve_host" on Host header
As reported by @jeremyjpj0916 :

> Host: header does not get set correctly when sending to upstream servers with request-transformation plugin enabled. The host header is being set as the gateways hostname itself and NOT the upstream URLs host.
> 
> Bandaid we implemented on 1.0.2 was a hardcode technique, replacing host header with the upstreams host leveraging the request transformer plugin.

## Steps to reproduce

> check host header sent from Kong when proxying to upstream, host header is not being set properly.

(this issue was forked from #4244)

hishamhm:
Source of the problem: the plugin was getting the full set of headers from `kong.request.get_headers`, which includes the request Host value, and was setting it (after transformations) to `kong.service.request.get_headers`. Effectively, this performed the equivalent of the Route attribute's `preserve_host = true` behavior all the time.

This commit restores the input value of the `Host` header from the service request to the table prior to the transformations. It also removes special-case handling of the `Host` header that was present in the plugin, which is now handled by the PDK itself.

A change in this behavior as opposed to 0.14 is that now the
`request-transformer` plugin _is_ able to modify the value of `Host`:
* if `preserve_host` is false, you can modify it using `add`;
* if `preserve_host` is true, you can modify it using `replace`.

Fixes #4252.

### DIFF
--- a/spec/03-plugins/14-request-transformer/02-access_spec.lua
+++ b/spec/03-plugins/14-request-transformer/02-access_spec.lua
@@ -5,6 +5,7 @@ local cjson   = require "cjson"
 for _, strategy in helpers.each_strategy() do
   describe("Plugin: request-transformer (access) [#" .. strategy .. "]", function()
     local proxy_client
+    local upstream_host = helpers.mock_upstream_host .. ':' .. helpers.mock_upstream_port
 
     lazy_setup(function()
       local bp = helpers.get_db_utils(strategy, {
@@ -17,8 +18,44 @@ for _, strategy in helpers.each_strategy() do
         hosts = { "test1.com" },
       })
 
-      local route2 = bp.routes:insert({
-        hosts = { "test2.com" },
+      local route_nph = bp.routes:insert({
+        hosts = { "no-preserve-host.test" },
+        preserve_host = false,
+      })
+
+      local route_ph = bp.routes:insert({
+        hosts = { "preserve-host.test" },
+        preserve_host = true,
+      })
+
+      local route_ph_ah = bp.routes:insert({
+        hosts = { "preserve-host-add-host.test" },
+        preserve_host = true,
+      })
+
+      local route_nph_ah = bp.routes:insert({
+        hosts = { "no-preserve-host-add-host.test" },
+        preserve_host = false,
+      })
+
+      local route_ph_rh = bp.routes:insert({
+        hosts = { "preserve-host-replace-host.test" },
+        preserve_host = true,
+      })
+
+      local route_nph_rh = bp.routes:insert({
+        hosts = { "no-preserve-host-replace-host.test" },
+        preserve_host = false,
+      })
+
+      local route_rename_host = bp.routes:insert({
+        hosts = { "rename-host.test" },
+        preserve_host = false,
+      })
+
+      local route_append_host = bp.routes:insert({
+        hosts = { "append-host.test" },
+        preserve_host = false,
       })
 
       local route3 = bp.routes:insert({
@@ -62,11 +99,81 @@ for _, strategy in helpers.each_strategy() do
       }
 
       bp.plugins:insert {
-        route = { id = route2.id },
+        route = { id = route_nph.id },
         name     = "request-transformer",
         config   = {
-          add    = {
-            headers = {"host:mark"}
+          add = {
+            headers = {"x-foo:bar"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_ph.id },
+        name     = "request-transformer",
+        config   = {
+          add = {
+            headers = {"x-foo:bar"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_nph_ah.id },
+        name     = "request-transformer",
+        config   = {
+          add = {
+            headers = {"host:added"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_ph_ah.id },
+        name     = "request-transformer",
+        config   = {
+          add = {
+            headers = {"host:added"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_nph_rh.id },
+        name     = "request-transformer",
+        config   = {
+          replace = {
+            headers = {"host:replaced"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_ph_rh.id },
+        name     = "request-transformer",
+        config   = {
+          replace = {
+            headers = {"host:replaced"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_rename_host.id },
+        name     = "request-transformer",
+        config   = {
+          rename = {
+            headers = {"host:foo"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_append_host.id },
+        name     = "request-transformer",
+        config   = {
+          append = {
+            headers = {"host:appended"}
           }
         }
       }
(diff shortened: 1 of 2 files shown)


hishamhm:
@thibaultcha About this last bit:

> A change in this behavior as opposed to 0.14 (demonstrated by the added test cases) is that now the request-transformer plugin is able to modify the value of Host, if preserve_host is false. If preserve_host is set, it still takes precedence, and preserves the host from the request.

Do you know of any gotchas that would make this behavior unwanted?

spec/03-plugins/14-request-transformer/02-access_spec.lua
```diff
@@ -166,7 +197,7 @@ for _, strategy in helpers.each_strategy() do
     end)
 
     lazy_teardown(function()
-      helpers.stop_kong()
+      helpers.stop_kong(nil, true, true)
```
thibaultcha: This helper only accepts 2 arguments as of the current master branch. Also, better not to modify the default behavior for debug purposes in the PR itself

spec/03-plugins/14-request-transformer/02-access_spec.lua
```diff
@@ -189,6 +220,9 @@ for _, strategy in helpers.each_strategy() do
           }
         })
         assert.response(res).has.status(200)
+        local host = assert.request(res).has.header("host")
+        assert.equals(upstream_host, host)
```
thibaultcha: Adding this condition to _all_ subsequent (unrelated) tests really does not respect tests atomicity... If this behavior was to be broken, we'd end up with dozens of broken tests instead of one (or a few)...

hishamhm: I did this because at first I tried to write an isolated regression test and I didn't hit the issue. So when I added to other tests, some presented the issue and others didn't (and it wasn't clear to me at first what the criteria was). So I ended up spreading it as an overall sanity check, a la verifying that the status remains 200.

No strong feelings about it, I'll pick one of the failing ones and make an isolated test case for it.

spec/03-plugins/14-request-transformer/02-access_spec.lua
```diff
@@ -71,6 +82,26 @@ for _, strategy in helpers.each_strategy() do
...
+            headers = {"x-added-foo:bar"}
+          }
+        }
+      }
+
+      bp.plugins:insert {
+        route = { id = route_phch.id },
+        name     = "request-transformer",
+        config   = {
+          add    = {
```
thibaultcha: If this was `replace` instead of `add`, would the plugin still not be able to override the Host header when `preserve_host = true`? It seems to me like it would.

hishamhm: haha, I had just added these test cases as I was reorganizing the patch! Check out the latest version :)

spec/03-plugins/14-request-transformer/02-access_spec.lua
```diff
@@ -653,6 +730,39 @@ for _, strategy in helpers.each_strategy() do
...
+              host             = "preserve-host-replace-host.test"
+            }
+          })
+          assert.response(res).has.status(200)
+          local json = assert.response(res).has.jsonbody()
+          local value = assert.has.header("host", json)
+          assert.equals("replaced", value)
+        end)
+
+        it("preserve_host = false, host change in plugin: upstream host", function()
```
thibaultcha: So, now to answer this question:

> Do you know of any gotchas that would make this behavior unwanted?

Which I was about to do in the previous patch, but then noted that it was because the tests were using `add` instead of `replace`.

To answer the question, I believe that it isn't an unreasonable assumption for a user to expect a plugin to override a Route's settings. As of this patch, we only allow replacing the Host header if `preserve_host = false` (which, arguable is even more confusing that the previous assumed behavior - which was incorrectly described, but that is of no importance).

Additionally, there is a deeper issue at hand here I think, which is that the reason why this test case passes (and why the plugin cannot replace a Host header when `preserve_host = false`, is because we rely on `ngx.var.upstream_host`, which itself relies on the router's behavior. So the reason why this does not work (or why this test case succeeds) is because we directly rely on router behavior [see upstream_host is nil](https://github.com/Kong/kong/blob/master/kong/router.lua#L1159) which causes [ngx.var.upstream_host to be nil](https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L751), which causes this plugin to not act on a `replace = { "host:foo" }` configuration value, because we guard the replace branch with `if headers[name] ~= nil then`.

In my opinion, this plugin could safely set the upstream header regardless of whether `preserve_host` is `true` or `false`. In doing so, we would avoid relying on core behavior as we currently do, which isn't regression-safe at all.
We should find another way than the `headers.host = ngx.var.upstream_host`. In my opinion, we are lacking a thin layer of abstraction between the core and this plugin (we may use existing PDK functions that act as this layer, or we may come up with a new PDK abstraction). This is one of these problems that #3821 is trying to solve, iirc.

hishamhm: Yeah, I had a bad feeling about this particular case you highlighted as well. Since the original issue was about corrupting `Host` when transformations were not being applied on it and the semantics of the behavior of transforming `Host` were unclear anyway, I pushed on with that original approach.

> In my opinion, this plugin could safely set the upstream header regardless of whether preserve_host is true or false.

I like these simpler semantics. 

> In doing so, we would avoid relying on core behavior as we currently do, which isn't regression-safe at all.

In the latest iteration, I had to special-case the `Host` header a bit more in the plugin again, but it does not rely on core behavior (only on the assumptions that a `Host` header must always exist [to forbid `add` and `rename`] and cannot be set multiple times [to forbid `append`] ). 

> In my opinion, we are lacking a thin layer of abstraction between the core and this plugin (we may use existing PDK functions that act as this layer, or we may come up with a new PDK abstraction). This is one of these problems that #3821 is trying to solve, iirc.

Agreed, because since `kong.request` represents the incoming request, its value of `get_header("host")` is correct. Perhaps what we really want is `kong.service.request.get_headers`, which would represent the state of headers as they are being prepared to be sent to the outgoing request (but of course there are difficulties in crafting such an API and keeping `ngx` and the PDK interoperable).

I think for now the approach in this PR is sufficient to fix the related issue.

thibaultcha: > In the latest iteration, I had to special-case the Host header a bit more in the plugin again, but it does not rely on core behavior (only on the assumptions that a Host header must always exist [to forbid add and rename] and cannot be set multiple times [to forbid append] ).

This iteration looks better in its behavior. But it's still somewhat hand-crafted based on deep Core behavior, and a thin abstraction would be nicer. Nonetheless, the patch looks good for now :+1:

> Perhaps what we really want is kong.service.request.get_headers, which would represent the state of headers as they are being prepared to be sent to the outgoing request

This has been proposed in #3840 already - there has been a few asks around it and it seems like a valuable addition to the PDK.

thibaultcha: Actually, I take that back; I think the current level of decoupling between the core and this plugin is fine as per this iteration :+1:

spec/03-plugins/14-request-transformer/02-access_spec.lua
```diff
@@ -858,6 +1016,69 @@ for _, strategy in helpers.each_strategy() do
...
+              host             = "preserve-host-add-host.test"
+            }
+          })
+          assert.response(res).has.status(200)
+          local json = assert.response(res).has.jsonbody()
+          local value = assert.has.header("host", json)
+          assert.equals("preserve-host-add-host.test", value)
+        end)
+
+        it("preserve_host = false, add fails for Host: upstream host", function()
```
thibaultcha: Aren't those tests just reflecting that `add` for Host is simply a no-op as well? `add fails for Host` makes me think it would produce an error of some sort, but the code just skips it, like `remove` for example.

PR #4255, opened 2019-01-29
fix(router) ensure rebuilds when Routes have 'regex_priority = nil'

### LINKED ISSUE #4254: "no router to route request" (regex_priority is nil)
For Kong version 1.0.0-1.0.2 configured with a Route with null/nil `regex_priority`, Kong throws a 500 error when receiving any request, logging the error (formatted for readability):

```
2019/01/28 22:47:06 [error] 54#0: *33046 [kong] handler.lua:687
  no router to route request (reason:
    /app/kong-runtime/share/lua/5.1/kong/runloop/handler.lua:116:
      attempt to compare nil with number), 
  client: x.x.x.x, 
  server: kong, 
  request: "GET /favicon.ico HTTP/1.1", 
  host: "xxxxx.example.com", 
  referrer: "https://xxxxx.example.com/"
```

Looking at [Kong source](https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L116) location referenced in that error:

```lua
    sort(routes, function(r1, r2)
      r1, r2 = r1.route, r2.route
      if r1.regex_priority == r2.regex_priority then
        return r1.created_at < r2.created_at
      end
      return r1.regex_priority > r2.regex_priority  -- ← THIS IS LINE #116
    end)
```

So it seems to be routing code comparing route priorities.

I check the Postgres database to see what the route contains for that field:

```
-[ RECORD 2 ]--+-------------------------------------
id             | b4e8ae92-8e4c-40a4-87d9-98af04969da0
created_at     | 2019-01-28 23:07:05+00
updated_at     | 2019-01-28 23:07:05+00
service_id     | 36a43882-84ba-4fc8-aa2a-94569c4bf2b7
protocols      | {https}
methods        | 
hosts          | {mars-micro-wasabi.happ.host}
paths          | 
regex_priority |                                      ← THIS FIELD IS NULL
strip_path     | t
preserve_host  | f
name           | 
snis           | 
sources        | 
destinations   | 
```

**The field is null.**

So, it seems like at least one of these solutions would be good for Kong:

* that handler/routing code should gracefully accept `nil` (treat like `0`)
* the data access layer should fill in a default of `0`
* the API should reject Route configuration requests that contain `"regex_priority": null`

thibaultcha:
If Routes have `regex_priority` properties mixed between numbers and nil
(e.g. one Route was created with a default `regex_priority`, the other
was created with a purposefully NULL `regex_priority`, the comparison in
the router rebuild sorting of Routes would error out, because we are
comparing one Route's `nil` value against another Route's `number`.

Fix #4254

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -110,10 +110,15 @@ do
 
     sort(routes, function(r1, r2)
       r1, r2 = r1.route, r2.route
-      if r1.regex_priority == r2.regex_priority then
+
+      local rp1 = r1.regex_priority or 0
+      local rp2 = r2.regex_priority or 0
+
+      if rp1 == rp2 then
         return r1.created_at < r2.created_at
       end
-      return r1.regex_priority > r2.regex_priority
+
+      return rp1 > rp2
     end)
 
     local err
--- a/spec/02-integration/05-proxy/02-router_spec.lua
+++ b/spec/02-integration/05-proxy/02-router_spec.lua
@@ -49,6 +49,10 @@ local function insert_routes(routes)
 end
 
 local function remove_routes(routes)
+  if not routes then
+    return
+  end
+
   local services = {}
 
   for _, route in ipairs(routes) do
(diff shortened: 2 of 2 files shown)


kong/runloop/handler.lua
```diff
@@ -110,10 +110,15 @@ do
 
     sort(routes, function(r1, r2)
       r1, r2 = r1.route, r2.route
-      if r1.regex_priority == r2.regex_priority then
+
+      local reg_prio1 = r1.regex_priority or 0
```
james-callahan: Can `regex_priority` ever be `ngx.null` rather than `nil`?

thibaultcha: Not in the proxy, where we purposefully don't use the `opts.nulls` option.

PR #4257, opened 2019-01-29
fix(schema) fix errors when encoding default empty arrays

kikito:
Previously, a schema such as this:
```
{ type = "array",
  default = {},
  elements = { type = "string" },
}
```
Would be encoded in json as a `{}` instead of as a `[]` (the Lua table given for the default value would be copied, and its metatable would remain)

A similar thing would happen with sets.

This change sets the metatable to `cjson.array_mt` and also adds tests for other possible values given to the default (i.e. `cjson.empty_array`).

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -855,7 +855,14 @@ end
 -- @param field The field definition table.
 local function handle_missing_field(field, value)
   if field.default ~= nil then
-    return tablex.deepcopy(field.default)
+    local copy = tablex.deepcopy(field.default)
+    if (field.type == "array" or field.type == "set")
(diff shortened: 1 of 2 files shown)


kong/db/schema/init.lua
```diff
@@ -855,7 +855,11 @@ end
 -- @param field The field definition table.
 local function handle_missing_field(field, value)
   if field.default ~= nil then
-    return tablex.deepcopy(field.default)
+    local copy = tablex.deepcopy(field.default)
+    if field.type == "array" and type(copy) == "table" and not getmetatable(copy) then
```
hishamhm: I think we need this for field.type == "set" as well

PR #4259, opened 2019-01-29
fix(db) workaround starttls bug by reverting pg to luasockets on preread

bungle:
### Summary

Makes `preread` phase work on tests shown in #4243. But this is a work-around, as it forces `LuaSockets` on `proxy`. On the other hand, this is the current, and wrong, behavior of `lua-cassandra` as well.

This was made for `1.0.3` as it is better to have working implementation (even with `luasockets`) than a broken implementation. If `starttls` can be corrected so that tests run correctly with tests added in #4243, this can be closed.

### Issues resolved

Workaround: #4243

### DIFF
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -123,7 +123,10 @@ local setkeepalive
 
 local function connect(config)
   local phase  = get_phase()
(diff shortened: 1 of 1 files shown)


kikito:
merged since it doesn't seem we'll have a starttls fix for 1.0.3

PR #4261, opened 2019-01-29
fix(cors) improve behavior of regex matching

hishamhm:
This reverts a regression introduced in #3872, in which regexes were being
forcibly anchored and matched against normalized domains, leading to a
breaking change with regard to the 0.x behavior.

In 0.x, regexes such as `(.*[.])?foo\.test` would accept subdomain entries
(but were subject to bug #3832 as it would also accept `foo.test.evil.test`);
in 1.0.2, the latter is not accepted, but the regular uses failed as well,
because regexes were translated to `^(.*[.])?foo\.test$` but were then matched
against the normalized domains in a way that always included the port.

With this commit, both configured origins and the input header value are
normalized so that default ports are not an issue for both regex and non-regex
cases. We verify if each configured origin is a regex or not, then:

* for non-regex entries, we do plain equality matching against the normalized
  domain
* For regex entries,
  * if the regex contains `:`, we do an anchored match against the normalized
    domain
  * otherwise, we do an anchored match against the host component only
    (to account for the 0.x behavior where ports were not considered)

Matching domains with regexes must be done with care (for starters,
note that dots must be escaped), so we recommend using plain-text full-domain
matching whenever possible.

This change in behavior is arguably a breaking change, but it is a compromise
between a workable behavior and backwards compatibility. Good domain-matching
regexes such as `(.*[.])?foo\.test` will remain matching against the host
component as intended (without being subject to #3832). Naive regexes such as
`.foo.test` will stop "working", but these were vulnerable to #3832 anyway. In
particular, thorough regexes such as
`^https?://(.*[.])?foo\\.test(:(80|90))?$` that performed their own anchoring
remain working as well.

### DIFF
--- a/spec/03-plugins/13-cors/01-access_spec.lua
+++ b/spec/03-plugins/13-cors/01-access_spec.lua
@@ -1,9 +1,233 @@
 local helpers = require "spec.helpers"
 local cjson = require "cjson"
+local inspect = require "inspect"
+local tablex = require "pl.tablex"
+
+
+local function sortedpairs(t)
+  local ks = tablex.keys(t)
+  table.sort(ks)
+  local i = 0
+  return function()
+    i = i + 1
+    return ks[i], t[ks[i]]
+  end
+end
+
 
 for _, strategy in helpers.each_strategy() do
   describe("Plugin: cors (access) [#" .. strategy .. "]", function()
     local proxy_client
+
+    local regex_testcases = {
+      {
+        -- single entry, host only: ignore value, always return configured data
+        origins = { "foo.test" },
+        tests = {
+          ["http://evil.test"]          = "foo.test",
+          ["http://foo.test"]           = "foo.test",
+          ["http://foo.test.evil.test"] = "foo.test",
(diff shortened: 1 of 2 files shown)


spec/03-plugins/13-cors/01-access_spec.lua
```diff
@@ -204,20 +428,73 @@ for _, strategy in helpers.each_strategy() do
...
+                ["Host"] = host,
+                ["Origin"] = origin,
+              }
+            })
+
+            assert.res_status(200, res)
+
+            if accept then
+              assert.equal("GET,HEAD,PUT,PATCH,POST,DELETE", res.headers["Access-Control-Allow-Methods"])
+              assert.equal(accept == true and origin or accept, res.headers["Access-Control-Allow-Origin"])
```
kikito: `accept` can be `true`, `false` or a domain. That was ... surprising. I initially thought this tertiary operator was unnecessary (given the `if accept` clause two lines above).

thibaultcha:
Manually merged, thanks!

PR #4274, opened 2019-02-01
misc(ci) bump kong-build-tools

hutchic:
Relevant diff of kong-build-tools is [here](https://github.com/Kong/kong-build-tools/compare/c2692df..cc761c3)

which includes the following PR's:
- [fix(tests) move test setup out of travis.yml into Makefile](https://github.com/Kong/kong-build-tools/pull/19)
- [feat(tests) use taverntesting for functional testing](https://github.com/Kong/kong-build-tools/pull/19)

The test setup is a change we require for functional testing to work correctly and is necessary for the package step of Kong to work. The taverntesting is a simple copy/paste of the existing bash tests into python/tavern


thibaultcha:
Could we eventually switch this to a branch such as master, so that we avoid having to bump this dependency here for every additional work being done in the kong-build-tools repository? That seems burdensome.

hutchic:
👍 I think that's a good idea while kong-build-tools is going through a lot of churn. In the future I like the idea of pinning the kong-build-tools version as it will give us a reliable way to build past versions of Kong / hotfix patches.

hutchic:
moving this change to be a part of https://github.com/Kong/kong/pull/4279

PR #4275, opened 2019-02-01
feat(core) implement entity tagging

fffonion:
This PR adds support for list entities by combination of tags, and listing all entity IDs with a specific tag.

### Summary

Handling of tags per entity is done in different way depending on the strategy. For either strategy, a `tags` table is created to store tags for entities, each entity table also have `tags` column.

- With Postgres, the `tags` column of an entity table is indexed with GIN index type:
  - When selecting entities by tags, the strategy queries the `tags` column with Postgres array operator, which will hit index.
  - When doing insert/update/upsert/delete, the strategy set the `tags` column of the table and update the columns in `tags` table as well. It uses trigger to ensure consistency.
- With cassandra:
  - When selecting entities by tags, the strategy queries the `tags` table with corresponding `entity_id` and `entity_name`, then select from the entity table with the returned entity_ids.
  - When doing insert/update/upsert/delete, the strategy set the `tags` column of the table and update the columns in `tags` table as well. It uses the `cassandra:batch` function to ensure atomicity and consistency.

For a custom entity that needs tags functionality, developer will need to set `tags` field to be true in the schema, add a `tags` field to fields defination, and add migrations as [kong/db/migrations/core/003_100_to_110.lua](https://github.com/Kong/kong/blob/120ff930eaa35d37f501a349777234971881362a/kong/db/migrations/core/003_100_to_110.lua).

### Full changelog
- Implement storing tags for entities and listing entities by combination of tags.
- Implement listing all tags or specific tag with related entity UUIDs.
- Implement the `batch` function in the Cassandra connector.


kong/db/dao/init.lua
```diff
@@ -145,6 +146,24 @@ local function validate_options_value(options, schema, context)
     errors.ttl = fmt("cannot be used with '%s'", schema.name)
   end
 
+  if schema.tags == true and options.tags ~= nil then
+    if context ~= "select" then
```
hishamhm: does the `options.tags` filter also work in generated methods for relationships, such as `db.targets:each_for_upstream({ id = U.id }, nil, { tags = T })`? (i.e. "give me all targets of upstream X that have tag T") — there needs to be a test in the suite answering this question :)

spec/02-integration/04-admin_api/14-tags_spec.lua
```diff
@@ -0,0 +1,91 @@
+local helpers = require "spec.helpers"
+local cjson = require "cjson"
+
+-- We already test the functionality of page() when filtering by tag in
```
hishamhm: there are no tests for `/tags`?

* what happens with `/tags` with no arguments
* what happens with `/tags/valid_tag`
* what happens with `/tags/invalid@tag`
* what happens when mixing `/tags` and `?tags=`
  * what happens with `/tags/valid_tag?tags=another_tag`
  * what happens with `/tags/valid_tag?tags=another_tag,yet_another`
  * what happens with `/tags/valid_tag?tags=another_tag/yet_another`
  * what happens with `/tags/valid_tag?tags=invalid@tag`
* anything else?

kong/db/dao/init.lua
```diff
@@ -145,6 +146,24 @@ local function validate_options_value(options, schema, context)
...
+      errors.tags = fmt("option can only be used with selects and pages, not with '%ss'",
+                       tostring(context))
```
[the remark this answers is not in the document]
bungle: > 2\. (I hope we don't need to support `/` in tag names in the future!)

I think the trick is to:
1. split non url-decoded
2. then url-decode each tag

which means you can have `/` (in a future) if you url encode them.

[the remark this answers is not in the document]
bungle: > magic `tags = true` attribute

I think it is that we wanted that to be `option`, but it is now an implicit field (or column). Also the original PG one didn't have it as a field. But well, now both C* and PG it is a field, so perhaps that means that it should be field in schemas too. Though a lot of automation will be harder to do. As we cannot check if entity schema is taggable. We can check if it has `tags` field, but we cannot be sure if it was meant for tagging or just a field. Now it is quite easy to check `if schema.tags then -- create triggers, add tags field etc. `.

One option is to have both, then we don't need that implicit field.

kong/db/strategies/postgres/init.lua
```diff
@@ -1409,22 +1428,22 @@ function _M.new(connector, schema, errors)
...
-      "  RETURNING ",  select_expressions, ";",
+      "  RETURNING ", select_expressions, ";",
     }
 
     update_statement = concat {
       "   UPDATE ",  table_name_escaped, "\n",
       "      SET ",  update_expressions, "\n",
       "    WHERE (", pk_escaped, ") = (", update_placeholders, ")\n",
-      "RETURNING ",  select_expressions , ";"
+      "RETURNING ", select_expressions, ";",
```
james-callahan: I don't think we want these stylistic changes?

kong/db/strategies/postgres/tags.lua
```diff
@@ -0,0 +1,140 @@
...
+  return page(self, size, token, options, tag)
+end
+
+-- Overwrite the page function for /tags
+function Tags:page(size, token, options)
+  return page(self, size, token, options)
+end
+
+
+return Tags
```
james-callahan: needs newline at EOF

spec/02-integration/03-db/02-db_core_entities_spec.lua
```diff
@@ -487,6 +487,7 @@ for _, strategy in helpers.each_strategy() do
             regex_priority  = 0,
             preserve_host   = false,
             strip_path      = true,
+            tags            = ngx.null,
```
james-callahan: Should we return an empty set rather than `null`?

kong/db/strategies/postgres/tags.lua
bungle: this is a quick way to configure editor:
https://github.com/Kong/kong/blob/master/.editorconfig

kong/db/dao/init.lua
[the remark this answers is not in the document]
hishamhm: > But well, now both C* and PG it is a field, so perhaps that means that it should be field in schemas too. 

Yes, that was my reasoning.

> We can check if it has tags field, but we cannot be sure if it was meant for tagging or just a field.

We could add a small validation for that in `kong.db.schema.entity` to make sure that any `tags` field has the correct "type signature".

kong/db/strategies/postgres/init.lua
kikito: We have no strict rule for that. On the one hand improving the code as you modify it is considered a good practice. On the other hand, ideally style changes should be on their own commit (with a `style` namespace). On the third hand, these changes were so small that adding an extra commit for them feels overkill.

kong/api/endpoints.lua
```diff
@@ -249,14 +280,24 @@ end
 -- /services
 local function get_collection_endpoint(schema, foreign_schema, foreign_field_name, method)
   return not foreign_schema and function(self, db, helpers)
+    local next_page_tags = ""
+
+    -- FIXME: saner api to avoid getting args here
+    local args = self.args.uri
+    if args.tags then
+      next_page_tags = "&tags=" .. concat(args.tags, args.tags.cond == "and" and "," or "/")
```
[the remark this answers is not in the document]
fffonion: my brain is fried when writing this, it should now be just

`next_page_tags = "&tags=" .. args.tags`

spec/02-integration/03-db/07-tags_spec.lua
```diff
@@ -0,0 +1,356 @@
...
+
+    end)
+
+
+    local func = pending
+    if strategy == "postgres" then
+      func = describe
+    end
+    func("trigger defined for all core entities", function()
+      -- TODO: test basic insert on other entities on tags table
```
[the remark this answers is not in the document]
fffonion: This test is intended to ensure that we write all `CREATE TRIGGER` for core entities correctly. Since the tests above only run for `services` and for postgres, the behaviour may change if we don't write correct trigger for other entities. It will not be necessary in Cassandra since the it is transparent to the strategy on what the entity is when updating tags table.

The correctness of "updating a tags field will also update the tags table" is tested above.

spec/02-integration/03-db/02-db_core_entities_spec.lua
fffonion: This will follow the same pattern as for example `paths` for `routes`, which if empty will be set to `ngx.null`

kong/db/dao/init.lua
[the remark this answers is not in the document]
fffonion: If I read the conversation here correctly, is the path forward with the `tags=true` attribute  having both the attribute and the field definition in schema?

spec/02-integration/03-db/07-tags_spec.lua
```diff
@@ -0,0 +1,356 @@
...
+          tags = { "team_a", "level_"..fmod(i, 5), "service"..i }
+        }
+        local row, err, err_t = bp.services:insert(service)
+        assert.is_nil(err)
+        assert.is_nil(err_t)
+        assert.same(service.tags, row.tags)
+      end
+    end)
+
+    it("list all entities attached with tag", function()
```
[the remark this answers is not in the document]
fffonion: yes, `have tags` will describe it clearer

kong/db/dao/init.lua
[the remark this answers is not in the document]
hishamhm: Just a regular field definition in the schema, no attribute. Then instead of checking for tags with `if schema.tags`, do `if schema.fields.tags`, should work just as well.

If we were concerned that the `tags` fields in entities is indeed of the correct type, we could add a check for that in `Entity.new` in `kong/db/schema/entity.lua`. But I don't really think that's necessary. Just using `{ tags = typedefs.tags }` in schema definitions should be enough, like we do for `id`, `created_at`, etc.

kong/db/dao/init.lua
fffonion: each_for_* will not support tags because it will not parse "?tags=" from Admin API, and uses a different pager `page_for_*` other than the tag-enabled function `page` used by normal pagination. I'll add a unit test on each_for_upstream to ensure the functionality doesn't exist.

fffonion:
Since the changes that address the above comments are small and sparse, i'll make them a single commit. Once this PR looks good I'll squash them into previous commit for better commit history.

spec/02-integration/04-admin_api/14-tags_spec.lua
fffonion: Added tests for /tags and /tags/:tags in spec/02-integration/04-admin_api/14-tags_spec.lua 👍

spec/02-integration/03-db/02-db_core_entities_spec.lua
james-callahan: Those are different though: adding a path constricts the behaviour: it says "only consider this route when one of these array elements match" -> an empty path array would be effectively turning a route off.
Tags are different: an empty list of tags is the same as no tags, and so IMO should *always* be an array.
In practice this lets us remove lots of type checks, both inside of Kong, and in users of Kong.

hishamhm: @james-callahan while I agree in principle that we should have been using empty sets rather than nulls everywhere (and I even disagree about your sub-point that paths are rightfully null, one could easily pick a semantics where empty set means no constraint added to paths), but this battle was [fought in 2017 and lost](https://github.com/Kong/kong/commit/cd748a3c9c14865c48de88e8b99f139fc6002419) (Ctrl+F "billion"), and now null is the default in all aggregate empty-accepting fields.

So, in short, @fffonion's design is fine, as it is consistent with the behavior of other attributes.

fffonion:
Hi, I have squashed commits. The tests are passing expect one https://travis-ci.org/Kong/kong/jobs/492317694, but I'm not sure how to understand this error.

hishamhm:
@fffonion Merged! \o/

PR #4279, opened 2019-02-04
misc(ci) adjust how defaults are set

hutchic:
build requirements were getting split between `git://kong/.requirements` and `git://kong-build-tools/Makefile` instead have one sole source of truth that is .requirements of the relevant source

proof of concept that this is sensible:

.requirements
```
hello=world
```

Makefile
```
hello ?= `grep hello .requirements | awk -F"=" '{print $$2}'`

test:
        @echo $(hello)
```

```
make test
world
export hello=foo
make test
foo
```


Makefile
```diff
@@ -11,34 +11,18 @@ else
 OPENSSL_DIR ?= /usr
 endif
 
-include .requirements
 .PHONY: install dev lint test test-integration test-plugins test-all fix-windows
 
+KONG_NETTLE_VERSION ?= `grep KONG_NETTLE_VERSION .requirements | awk -F"=" '{print $$2}'`
```
thibaultcha: Since we removed nettle from `.requirements` (because unused in this repository), can we also remove mentions of it from the Makefile as well?

jeveleth:
Aside from the commit type, this seems good to me.

thibaultcha:
This could go to master as well?

PR #4284, opened 2019-02-06
Feat/declarative config db import

kikito:


kong-1.0.3-0.rockspec
```diff
@@ -76,11 +77,13 @@ build = {
     ["kong.cmd.quit"] = "kong/cmd/quit.lua",
     ["kong.cmd.start"] = "kong/cmd/start.lua",
     ["kong.cmd.check"] = "kong/cmd/check.lua",
+    ["kong.cmd.config"] = "kong/cmd/config.lua",
     ["kong.cmd.reload"] = "kong/cmd/reload.lua",
     ["kong.cmd.restart"] = "kong/cmd/restart.lua",
     ["kong.cmd.prepare"] = "kong/cmd/prepare.lua",
     ["kong.cmd.migrations"] = "kong/cmd/migrations.lua",
     ["kong.cmd.health"] = "kong/cmd/health.lua",
+    ["kong.cmd.config"] = "kong/cmd/config.lua",
```
Tieske: just passing by.... this line is in here twice, see 80 above

kong/cmd/config.lua
```diff
@@ -0,0 +1,108 @@
...
+
+local lapp = [[
+Usage: kong config COMMAND [OPTIONS]
+
+Use declarative configuration files with Kong.
+
+The available commands are:
+  import <file>                 Import a declarative config file into Kong.
+
+  parse <file>                  Parse a declarative config file (check
```
Tieske: parse = lint? or validate?

bungle: test, validate, verify, check... parsing and linting are more like programmer terms there.

kong/db/schema/topological_sort.lua
```diff
@@ -0,0 +1,106 @@
...
+
+
+-- aux function for topological_sort
+local function visit(current, neighbors_map, visited, marked, sorted)
+  if visited[current] then
+    return true
+  end
+
+  if marked[current] then
+    return nil, "Cycle detected, can not sort topographically"
```
bungle: `can not` -> `cannot`

kong/cmd/config.lua
Tieske: we have a cli command `kong check` so maybe reuse that terminology? eg. `kong config check <file>`

hishamhm:
Please do not merge this PR. This is based on the ongoing declarative config work which is still in flux.

kong/db/schema/topological_sort.lua
hishamhm: also, `topographically` -> `topologically`

kong-1.0.3-0.rockspec
```diff
@@ -128,6 +129,7 @@ build = {
     ["kong.db.dao.snis"] = "kong/db/dao/snis.lua",
     ["kong.db.dao.targets"] = "kong/db/dao/targets.lua",
     ["kong.db.dao.plugins"] = "kong/db/dao/plugins.lua",
+    ["kong.db.declarative"] = "kong/db/declarative/init.lua",
```
hishamhm: this also should not be in this commit, but rather in the commit where the module is introduced

spec/02-integration/03-db/07-declarative_spec.lua
```diff
@@ -0,0 +1,175 @@
...
+
+    before_each(function()
+      db.acls:truncate()
+      db.plugins:truncate()
+      db.routes:truncate()
+      db.services:truncate()
+      db.snis:truncate()
+      db.certificates:truncate()
+      db.consumers:truncate()
+    end)
```
hishamhm: We should avoid this pattern of truncating on `before_each`. It's better to write tests that don't require truncating the database (i.e. make different tests make unrelated uuids/entities). If you _need_ to truncate a database (i.e. to verify a table remains empty) do so only for the needed tables in the specific tests.

With this `before_each` block it will incur in a significant time penalty for the tests: each time a new test is added to this file, even if it has nothing to do with some of these tables. And currently, with only one test, it is truncating everything twice, since `start_kong` is already doing it in `lazy_setup`.

(Truncating the necessary tables only at the beginning of a test file is fine to ensure a clean-slate, as `start_kong` above does already).

spec/fixtures/declarative_config.yaml
```diff
@@ -0,0 +1,25 @@
+_format_version: "1.1"
```
hishamhm: I expect many different declarative config tests to exist in the test suite (I know I'm adding a bunch in my branch!), so a single yaml file called "the" `spec/fixtures/declarative_config.yml` will be confusing (or will grow huge as people reusing adding more stuff to test). 

Since this is not a shared file anyway, could you embed this file in the test where it is used (e.g. as a long-string that's dumped to a temp file and then use it instead)? It would be easier to read the test as well, cross-checking the config against the result. Something like
```lua
local function make_yaml_file(content)
  local filename = os.tmpname()
  os.rename(filename, filename .. ".yml")
  filename = filename .. ".yml"
  local fd = assert(io.open(yaml_file, "w"))
  assert(fd:write(content))
  assert(fd:close())
  return filename
end

it("my test", function()
  local yaml_file = make_yaml_file [[
    _format_version: "1.1"
    ...
  ]]
  finally(function()
    os.remove(yaml_file)
  end)
```

kong/cmd/config.lua
hishamhm: @bungle @Tieske all valid concerns; I'm going to mark "resolve" in this thread because this comment pertains to the base `feat/declarative-config` branch that the db-import work is based on. We'll revisit this question in another PR.

kong/cmd/config.lua
```diff
@@ -19,6 +21,8 @@ local function execute(args)
   }))
   log.enable()
 
+  log("potato")
```
hishamhm: :laughing:

PR #4285, opened 2019-02-06
misc(ci) remove distributions we no longer support because they've be…

hutchic:
Debian 7 and Ubuntu 12.04 are both EOL so we no longer build / support these distributions


thibaultcha:
@hutchic Oh, just noticing, but I this can go to master can't it?

PR #4286, opened 2019-02-06
feat(core) make service on routes optional (transparent proxying)

bungle:
### Summary

This is first implementation of so called `allow_by_default` routing. E.g. routes that act as a transparent routes. Which means you don't need to create service entity, but instead you can use HTTP proxying or `iptables` transparent proxying rules.

### DIFF
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -104,7 +104,7 @@ return {
                        },
                      }, },
     { tags             = typedefs.tags },
-    { service = { type = "foreign", reference = "services", required = true }, },
+    { service = { type = "foreign", reference = "services" }, },
   },
 
   entity_checks = {
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -13,6 +13,7 @@ local upper         = string.upper
 local lower         = string.lower
 local find          = string.find
 local sub           = string.sub
+local tonumber      = tonumber
 local ipairs        = ipairs
 local pairs         = pairs
 local error         = error
@@ -65,6 +66,9 @@ local MATCH_RULES = {
   SRC             = 0x00000002,
   DST             = 0x00000001,
 }
+
+
+local EMPTY_T = {}
 
 
 local match_route
@@ -333,7 +337,9 @@ local function marshall_route(r)
     route_t.upstream_url_t.scheme = protocol
   end
 
-  local host = service.host
+  local s = service or EMPTY_T
+
+  local host = s.host
   if host then
     route_t.upstream_url_t.host = host
     route_t.upstream_url_t.type = hostname_type(host)
@@ -342,7 +348,7 @@ local function marshall_route(r)
     route_t.upstream_url_t.type = hostname_type("")
   end
 
-  local port = service.port
+  local port = s.port
   if port then
     route_t.upstream_url_t.port = port
 
@@ -356,7 +362,7 @@ local function marshall_route(r)
   end
 
   if route_t.type == "http" then
-    route_t.upstream_url_t.path = service.path or "/"
+    route_t.upstream_url_t.path = s.path or "/"
   end
 
   return route_t
@@ -1129,14 +1135,13 @@ function _M.new(routes)
             local upstream_url_t = matched_route.upstream_url_t
             local matches        = ctx.matches
 
-
             -- Path construction
 
             if matched_route.type == "http" then
               -- if we do not have a path-match, then the postfix is simply the
               -- incoming path, without the initial slash
               local request_postfix = matches.uri_postfix or sub(req_uri, 2, -1)
-              local upstream_base = upstream_url_t.path
+              local upstream_base = upstream_url_t.path or "/"
 
               if matched_route.strip_uri then
                 -- we drop the matched part, replacing it with the upstream path
@@ -1229,11 +1234,24 @@ function _M.new(routes)
       -- debug HTTP request header logic
 
       if ngx.var.http_kong_debug then
-        ngx.header["Kong-Route-Id"]   = match_t.route.id
-        ngx.header["Kong-Service-Id"] = match_t.service.id
+        if match_t.route then
+          if match_t.route.id then
+            ngx.header["Kong-Route-Id"] = match_t.route.id
+          end
 
-        if match_t.service.name then
-          ngx.header["Kong-Service-Name"] = match_t.service.name
+          if match_t.route.name then
+            ngx.header["Kong-Route-Name"] = match_t.route.name
+          end
+        end
+
+        if match_t.service then
+          if match_t.service.id then
+            ngx.header["Kong-Service-Id"] = match_t.service.id
+          end
+
+          if match_t.service.name then
+            ngx.header["Kong-Service-Name"] = match_t.service.name
+          end
         end
       end
 
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -7,7 +7,6 @@ local bit           = require "bit"
 local hostname_type = utils.hostname_type
 local re_match      = ngx.re.match
 local re_find       = ngx.re.find
-local null          = ngx.null
 local insert        = table.insert
 local sort          = table.sort
 local upper         = string.upper
@@ -89,18 +88,22 @@ local protocol_subsystem = {
 }
 
 local function marshall_route(r)
-  local route    = r.route          or null
-  local service  = r.service        or null
-  local headers  = r.headers        or null
-  local paths    = route.paths      or null
-  local methods  = route.methods    or null
-  local protocol = service.protocol or null
-  local sources  = route.sources  or null
-  local destinations = route.destinations or null
-  local snis     = route.snis or null
+  local route        = r.route
+  local service      = r.service
+  local headers      = r.headers
+  local paths        = route.paths
+  local methods      = route.methods
+  local snis         = route.snis
+  local sources      = route.sources
+  local destinations = route.destinations
 
-  if not (headers ~= null or methods ~= null or paths ~= null or
-          sources ~= null or destinations ~= null or snis ~= null) then
+
+  local protocol
+  if service then
+    protocol = service.protocol
+  end
+
+  if not (headers or methods or paths or snis or sources or destinations) then
     return nil, "could not categorize route"
   end
 
@@ -125,7 +128,7 @@ local function marshall_route(r)
   -- headers
 
 
-  if headers ~= null then
+  if headers then
     if type(headers) ~= "table" then
       return nil, "headers field must be a table"
     end
@@ -174,7 +177,7 @@ local function marshall_route(r)
   -- paths
 
 
-  if paths ~= null then
+  if paths then
     if type(paths) ~= "table" then
       return nil, "paths field must be a table"
     end
@@ -219,7 +222,7 @@ local function marshall_route(r)
   -- methods
 
 
-  if methods ~= null then
+  if methods then
     if type(methods) ~= "table" then
       return nil, "methods field must be a table"
     end
@@ -238,7 +241,7 @@ local function marshall_route(r)
   -- sources
 
 
-  if sources ~= null then
+  if sources then
     if type(sources) ~= "table" then
       return nil, "sources field must be a table"
     end
@@ -271,7 +274,7 @@ local function marshall_route(r)
   -- destinations
 
 
-  if destinations ~= null then
+  if destinations then
     if type(destinations) ~= "table" then
       return nil, "destinations field must be a table"
     end
@@ -304,7 +307,7 @@ local function marshall_route(r)
   -- snis
 
 
-  if snis ~= null then
+  if snis then
     if type(snis) ~= "table" then
       return nil, "snis field must be a table"
     end
@@ -323,16 +326,15 @@ local function marshall_route(r)
   end
 
 
-  if protocol ~= null then
-    route_t.upstream_url_t.scheme = protocol
-  end
-
-
   -- upstream_url parsing
 
 
-  local host = service.host or null
-  if host ~= null then
+  if protocol then
+    route_t.upstream_url_t.scheme = protocol
+  end
+
+  local host = service.host
+  if host then
     route_t.upstream_url_t.host = host
     route_t.upstream_url_t.type = hostname_type(host)
 
@@ -340,8 +342,8 @@ local function marshall_route(r)
     route_t.upstream_url_t.type = hostname_type("")
   end
 
-  local port = service.port or null
-  if port ~= null then
+  local port = service.port
+  if port then
     route_t.upstream_url_t.port = port
 
   else
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -21,6 +21,7 @@ local concurrency = require "kong.concurrency"
 
 
 local kong        = kong
+local ipairs      = ipairs
 local tostring    = tostring
 local tonumber    = tonumber
 local sub         = string.sub
@@ -40,7 +41,9 @@ local DEBUG       = ngx.DEBUG
 
 
 local CACHE_ROUTER_OPTS = { ttl = 0 }
+local SUBSYSTEMS = constants.PROTOCOLS_WITH_SUBSYSTEM
 local EMPTY_T = {}
+
 
 
 local get_router, build_router
@@ -58,15 +61,41 @@ end
 
 do
   -- Given a protocol, return the subsystem that handles it
-  local protocol_subsystem = {
-    http = "http",
-    https = "http",
-    tcp = "stream",
-    tls = "stream",
-  }
-
   local router
   local router_version
+
+  local function should_process_route(route)
+    for _, protocol in ipairs(route.protocols) do
+      if SUBSYSTEMS[protocol] == subsystem then
+        return true
+      end
+    end
+
+    return false
+  end
+
+  local function get_service_for_route(db, route)
+    local service_pk = route.service
+    if not service_pk then
+      return nil
+    end
+
+    local service, err = db.services:select(service_pk)
+    if not service then
+      return nil, "could not find service for route (" .. route.id .. "): " ..
+                  err
+    end
+
+    -- TODO: this should not be needed as the schema should check it already
+    if SUBSYSTEMS[service.protocol] ~= subsystem then
+      log(WARN, "service with protocol '", service.protocol,
+                "' cannot be used with '", subsystem, "' subsystem")
+
+      return nil
+    end
+
+    return service
+  end
 
   build_router = function(db, version)
     local routes, i = {}, 0
@@ -76,26 +105,25 @@ do
         return nil, "could not load routes: " .. err
       end
 
-      local service_pk = route.service
+      if should_process_route(route) then
+        local service, err = get_service_for_route(db, route)
+        if err then
+          return nil, err
+        end
 
-      if not service_pk then
-        return nil, "route (" .. route.id .. ") is not associated with service"
-      end
-
-      local service, err = db.services:select(service_pk)
-      if not service then
-        return nil, "could not find service for route (" .. route.id .. "): " ..
-                    err
-      end
-
-      local stype = protocol_subsystem[service.protocol]
-      if subsystem == stype then
         local r = {
           route   = route,
           service = service,
         }
 
-        if stype == "http" and route.hosts then
+        local service_subsystem
+        if service then
+          service_subsystem = SUBSYSTEMS[service.protocol]
+        else
+          service_subsystem = subsystem
+        end
+
+        if service_subsystem == "http" and route.hosts then
           -- TODO: headers should probably be moved to route
           r.headers = {
             host = route.hosts,
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -25,18 +25,22 @@ local ipairs      = ipairs
 local tostring    = tostring
 local tonumber    = tonumber
 local sub         = string.sub
+local find        = string.find
 local lower       = string.lower
 local fmt         = string.format
 local sort        = table.sort
 local ngx         = ngx
 local log         = ngx.log
 local ngx_now     = ngx.now
+local re_match    = ngx.re.match
+local re_find     = ngx.re.find
 local update_time = ngx.update_time
 local subsystem   = ngx.config.subsystem
 local unpack      = unpack
 
 
 local ERR         = ngx.ERR
+local WARN        = ngx.WARN
 local DEBUG       = ngx.DEBUG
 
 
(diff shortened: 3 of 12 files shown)


bungle:
@james-callahan can you do a small feasability check on this? Am I on a right page, or do you see some problems? If everything is fine, I think I can move to adding tests.

kong/runloop/handler.lua
```diff
@@ -77,34 +78,49 @@ do
         return nil, "could not load routes: " .. err
       end
 
-      local service_pk = route.service
-
-      if not service_pk then
-        return nil, "route (" .. route.id .. ") is not associated with service"
+      local process_route
+      for _, protocol in ipairs(route.protocols) do
+        if protocol_subsystem[protocol] == subsystem then
```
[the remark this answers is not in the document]
bungle: Replaced.

kong/router.lua
```diff
@@ -1229,11 +1236,14 @@ function _M.new(routes)
...
-        ngx.header["Kong-Route-Id"]   = match_t.route.id
-        ngx.header["Kong-Service-Id"] = match_t.service.id
+        ngx.header["Kong-Route-Id"] = match_t.route.id
+
+        if match_t.service then
+          ngx.header["Kong-Service-Id"] = match_t.service.id
 
-        if match_t.service.name then
-          ngx.header["Kong-Service-Name"] = match_t.service.name
+          if match_t.service.name then
```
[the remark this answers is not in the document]
bungle: Protected that and the route as well.

kong/runloop/handler.lua
```diff
@@ -737,25 +725,82 @@ return {
...
+      if not service then
+        local service_scheme
+        local service_host
+        local service_port
+
+        local http_host = var.http_host
+        if http_host then
+          http_host = lower(http_host)
+
+          local s = find(http_host, ":", 2, true)
```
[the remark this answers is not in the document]
bungle: Added comment above.

kong/router.lua
```diff
@@ -323,38 +330,41 @@ local function marshall_route(r)
...
-  end
+  -- upstream_url parsing
 
 
-  -- upstream_url parsing
+  if protocol then
+    route_t.upstream_url_t.scheme = protocol
+  end
 
+  do
```
james-callahan: Why the `do/end` block here?

kong/runloop/handler.lua
```diff
@@ -721,25 +718,86 @@ return {
 
       local trusted_ip = kong.ip.is_trusted(realip_remote_addr)
       if trusted_ip then
-        forwarded_proto = var.http_x_forwarded_proto or var.scheme
-        forwarded_host  = var.http_x_forwarded_host  or var.host
-        forwarded_port  = var.http_x_forwarded_port  or var.server_port
+        forwarded_proto = var.http_x_forwarded_proto          or scheme
+        forwarded_host  = var.http_x_forwarded_host           or host
+        forwarded_port  = tonumber(var.http_x_forwarded_port) or port
```
james-callahan: Pass a base of `10` to `tonumber`?

kong/runloop/handler.lua
```diff
@@ -721,25 +718,86 @@ return {
...
+        if http_host then
+          http_host = lower(http_host)
+
+          -- Micro-optimization to make sure we use JITable functions in this
+          -- potentially hot path, e.g. this simple code doesn't warrant us
+          -- using something like socket.url.parse:
+          -- https://github.com/diegonehab/luasocket/blob/043e99771352aff47680b99f09b66a32f0cc3ef5/src/url.lua#L143-L192
+          local s = find(http_host, ":", 2, true)
+          if s then
+            local p = tonumber(sub(http_host, s + 1))
```
james-callahan: Pass base to `tonumber`?

kong/router.lua
bungle: Removed.

kong/runloop/handler.lua
bungle: Passing it now.

kong/runloop/handler.lua
bungle: Passing it now. Thank you @james-callahan for the review!

kong/router.lua
```diff
@@ -1150,7 +1157,7 @@ function _M.new(routes)
               else
                 -- we retain the incoming path, just prefix it with the upstream
                 -- path, but skip the initial slash
-                upstream_uri = upstream_base .. sub(req_uri, 2, -1)
+                upstream_uri = upstream_base .. sub(req_uri, 2)
```
james-callahan: Superficial change intended?

bungle: Reverted that.

kong/runloop/handler.lua
```diff
@@ -721,25 +718,86 @@ return {
...
+        if http_host then
+          http_host = lower(http_host)
+
+          -- Micro-optimization to make sure we use JITable functions in this
+          -- potentially hot path, e.g. this simple code doesn't warrant us
+          -- using something like socket.url.parse:
+          -- https://github.com/diegonehab/luasocket/blob/043e99771352aff47680b99f09b66a32f0cc3ef5/src/url.lua#L143-L192
+          local s = find(http_host, ":", 2, true)
+          if s then
+            local p = tonumber(sub(http_host, s + 1), 10)
```
james-callahan: Make sure you check for a non-whole number e.g. user provides `1000.5`

kong/runloop/handler.lua
```diff
@@ -721,25 +718,86 @@ return {
...
+            service_port = var.server_port
+          end
+        end
+
+        if service_port == 443 then
+          service_scheme = "https"
+        elseif service_port == 80 then
+          service_scheme = "http"
+        else
+          service_scheme = scheme
```
james-callahan: Shouldn't we always set `service_scheme` to `scheme`?

kong/runloop/handler.lua
bungle: Added check (one line below this)

bungle: Now there is a problem with `10.` too :-)

bungle: And also scientific notation like `12e3`.

kong/runloop/handler.lua
bungle: I am not sure, let's say you connect to Kong using `https` and have a `host` header of `Host: www.google.com:80`, Should we still connect to `www.google.com` on port `80` with `http` or `https`?

james-callahan: I don't think we should add special cases here: simpler to just have a rule "scheme in is scheme out". 

Note that in future we plan on having a route property like "terminate_tls" (recall https://github.com/Kong/kong/blob/aa034cda1be49a05751cb2695b1b56bbbcfafd1e/kong/runloop/handler.lua#L627), using that would probably be the right place to let a user configure http vs https.

kong/runloop/handler.lua
```diff
@@ -721,25 +724,114 @@ return {
...
+        local service_host
+        local service_port
+
+        -- 1. try to find information from a request-line
+        if var.http2 == "" then
+          local request_line = var.request
+          if request_line then
+            local matches = ngx.re.match(request_line, [[\w+ (\S+)]], "ajos")
+            if matches and matches[1] then
+              local uri = url.parse(matches[1])
```
james-callahan: Could this be done with a regex that extracts matches?

I think this could be a hot path. But as it's new functionality we could just fix it later.
maybe add a `-- TODO: optimize`?

kong/runloop/handler.lua
```diff
@@ -721,25 +724,114 @@ return {
...
+              local uri_scheme = uri and uri.scheme
+              if uri_scheme then
+                uri_scheme = lower(uri_scheme)
```
james-callahan: Interesting: I this should be done by luasocket; I think the lack of it should be considered a luasocket bug.

kong/runloop/handler.lua
```diff
@@ -721,25 +724,114 @@ return {
...
+                if uri_scheme ~= "http" or uri_scheme ~= "https" then
```
james-callahan: For a websocket connection, do we need to check for `"ws"` and `"wss"`?

kong/runloop/handler.lua
```diff
@@ -21,15 +22,18 @@ local certificate = require "kong.runloop.certificate"
...
 local tonumber    = tonumber
 local sub         = string.sub
+local find        = string.find
 local lower       = string.lower
 local fmt         = string.format
 local sort        = table.sort
 local ngx         = ngx
 local log         = ngx.log
 local ngx_now     = ngx.now
+local rematch     = ngx.re.match
```
james-callahan: Is this a conventional alias for this function? I would have expected `re_match`. (I got confused when reading further down)

kong/runloop/handler.lua
bungle:
```
> = require "socket.url".parse("HttP://www.google.com").scheme
HttP
```

james-callahan: Yes. I would consider that a luasocket bug.

Fine to work around it as you have.

kong/runloop/handler.lua
```diff
@@ -77,34 +74,49 @@ do
...
+        local service_subsystem
+
+        if service_pk then
+          service, err = db.services:select(service_pk)
+          if not service then
+            return nil, "could not find service for route (" .. route.id .. "): " ..
+                        err
+          end
+
+          service_subsystem = constants.PROTOCOLS_WITH_SUBSYSTEM[service.protocol]
```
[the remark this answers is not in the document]
bungle: This is possible ATM:
```
 http PUT :8001/services/test url=http://www.google.com/
http POST :8001/services/test/routes destinations[1].ip=0.0.0.0/0 protocols=tcp -f 
```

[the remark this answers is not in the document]
james-callahan: Ah. It's not meant to be. We ran out of time to add the restriction for this sort of thing to the metaschema. @hishamhm probably recalls the conversation?

So I guess leave this branch in here for now, but perhaps leave a TODO saying "should be enforced by schema"?

kong/db/schema/entities/routes.lua
```diff
@@ -104,7 +104,7 @@ return {
                        },
                      }, },
     { tags             = typedefs.tags },
-    { service = { type = "foreign", reference = "services", required = true }, },
+    { service = { type = "foreign", reference = "services", required = false }, },
```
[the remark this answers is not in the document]
hishamhm: I'd prefer leaving it out, but no big issue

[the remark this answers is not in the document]
hishamhm: (maybe it could be confusing for newcomers though)

kong/runloop/handler.lua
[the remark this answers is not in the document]
hishamhm: You mean doing something similar to what @kikito did in 608872dbfbcc1dbaec6e680f7dac52160ff0de73 for checking that protocols in plugins and routes match?

Do we need something similar for services and routes?

kong/db/schema/entities/routes.lua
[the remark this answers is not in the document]
bungle: removed.

kong/runloop/handler.lua
bungle: I rewrote that without `socket.url`.

kong/runloop/handler.lua
bungle: Worked this to remove dependency on `socket.url`.

kong/runloop/handler.lua
bungle: It now checks for `ws` and `wss` too.

kong/runloop/handler.lua
bungle: changed to `re_match`.

kong/runloop/handler.lua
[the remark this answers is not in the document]
bungle: @hishamhm so yes, I think so. I am not sure can we downgrade `http/https` to `tcp/tls`, and I am not sure if we can upgrade `tcp/tls` to `http/https`, at least we don't route such (I didn't invent `service.protocol` check here in router rebuild, I just added route `subsystem` check, which is quite obvious thing to do as there we don't have to deal with such question anyway.

kong/runloop/handler.lua
```diff
@@ -77,34 +74,49 @@ do
         return nil, "could not load routes: " .. err
       end
 
-      local service_pk = route.service
-
-      if not service_pk then
-        return nil, "route (" .. route.id .. ") is not associated with service"
+      local process_route
```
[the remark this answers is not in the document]
bungle: The helper was added.

kong/runloop/handler.lua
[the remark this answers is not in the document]
james-callahan: @hishamhm yes.

kong/runloop/handler.lua
```diff
@@ -59,16 +62,37 @@ end
...
-
   local router
   local router_version
 
+  local function should_process_route(route)
+    for _, protocol in ipairs(route.protocols) do
+      if constants.PROTOCOLS_WITH_SUBSYSTEM[protocol] == subsystem then
+        return true
+      end
+    end
```
james-callahan: Add `return false` rather than falling off?

kong/runloop/handler.lua
```diff
@@ -59,16 +62,37 @@ end
...
+
+    local service, err = db.services:select(service_pk)
+    if not service then
+      return nil, "could not find service for route (" .. route.id .. "): " ..
+                  err
+    end
+
+    -- TODO: this should not be needed as the schema should check it already
+    local service_subsystem = constants.PROTOCOLS_WITH_SUBSYSTEM[service.protocol]
+    if service_subsystem == subsystem then
```
james-callahan: I'd prefer to invert this check and return a good error message.
That way we `return` for the happy case

kong/runloop/handler.lua
```diff
@@ -721,25 +728,117 @@ return {
...
+        if var.http2 == "" then
+          local request_line = var.request
+          if request_line then
+            local matches = re_match(request_line, [[\w+ (https?|wss?)://([^/?#\s]+)]], "ajos")
+            if matches then
+              local uri_scheme = lower(matches[1])
+              if uri_scheme == "https" or uri_scheme == "http" then
+                service_scheme = uri_scheme
+                service_host   = lower(matches[2])
+              elseif uri_scheme == "wss" then
```
james-callahan: I wasn't 100% sure these paths were required, it was just an idea: did you verify what occurs if a websocket client uses a http proxy?

bungle: @james-callahan, it needs testing. The devil is in details in this PR, so thank you for all the feedback so far. We need to get this right. Currently adding more tests to it.

kong/runloop/handler.lua
bungle: @james-callahan 

Do you consider logging it is enough? I mean that if we forcibly `return` the router rebuilding will never work. I know that the db should not be in that state ever, but still, I think it is better to just log it and move on.

kong/runloop/handler.lua
bungle: Added that.

kong/runloop/handler.lua
bungle: Added logging.

kong/runloop/handler.lua
bungle: There is now more tests added. Lets figure out the websocket thing later as I think it needs in most cases `Connect`, and that is what we (or Nginx) don't support currently.

kong/runloop/handler.lua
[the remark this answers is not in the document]
bungle: I mark this as resolved here as it is not part of this PR.

kong/runloop/handler.lua
james-callahan: logging is good. I'd still like to inverse this `if` statement.

kong/runloop/handler.lua
```diff
@@ -698,25 +710,118 @@ return {
...
+        -- Serviceless HTTP / HTTPS / HTTP2 route
+        -----------------------------------------------------------------------
+        local service_scheme
+        local service_host
+        local service_port
+
+        -- 1. try to find information from a request-line
+        local request_line = var.request
+        if request_line then
+          local matches = re_match(request_line, [[\w+ (https?)://([^/?#\s]+)]], "ajos")
```
james-callahan: does this pattern need anchoring?

spec/02-integration/05-proxy/18-serviceless_proxying_spec.lua
```diff
@@ -0,0 +1,468 @@
...
+
+          assert(helpers.start_kong({
+            database      = strategy,
+            nginx_conf    = "spec/fixtures/custom_nginx.template",
+            stream_listen = "off",
+            admin_listen  = "off",
+          }))
+
+        else
+          yaml_file = helpers.make_yaml_file [[
```
james-callahan: use `unindent` helper?

spec/02-integration/05-proxy/18-serviceless_proxying_spec.lua
```diff
@@ -0,0 +1,468 @@
...
+  - paths:
+    - /
+]]
+
+          assert(helpers.start_kong({
+            database           = strategy,
+            nginx_conf         = "spec/fixtures/custom_nginx.template",
+            declarative_config = yaml_file,
+            stream_listen      = "off",
+            admin_liten        = "off",
```
james-callahan: typo: liten vs listen

spec/02-integration/05-proxy/18-serviceless_proxying_spec.lua
```diff
@@ -0,0 +1,468 @@
...
+      lazy_teardown(function()
+        if yaml_file then
+          os.remove(yaml_file)
+        end
+
+        helpers.stop_kong(helpers.test_conf.prefix, true)
+      end)
+
+      it("proxies http to http (request-line)", function()
+        local request = assert(http_request.new_from_uri(HTTP_UPSTREAM_URI))
```
james-callahan: this doesn't need an assert

spec/02-integration/05-proxy/18-serviceless_proxying_spec.lua
```diff
@@ -0,0 +1,468 @@
...
+        request.proxy = HTTP_PROXY_URI
+        request.version = 1.1
+        request.tls = false
+        local headers, stream = request:go()
+
+        assert.equal(200, tonumber(headers:get(":status")))
+        assert.equal(meta._SERVER_TOKENS, headers:get("via"))
```
james-callahan: I think you need to wrap the `:get` in parenthesis.

spec/02-integration/05-proxy/18-serviceless_proxying_spec.lua
```diff
@@ -0,0 +1,468 @@
...
+        ctx:setVerify(openssl_ssl.VERIFY_NONE)
+
+        local request = assert(http_request.new_from_uri(HTTPS_UPSTREAM_URI))
+        request.proxy = HTTP_PROXY_URI
+        request.version = 1.1
+        request.tls = true
+        request.ctx = ctx
+        local headers, stream = request:go()
+
+        print(stream)
```
james-callahan: leftover print?

kong/runloop/handler.lua
bungle: `a` means it is anchored.

kong/runloop/handler.lua
```diff
@@ -58,16 +62,41 @@ end
...
+    if not service then
+      return nil, "could not find service for route (" .. route.id .. "): " ..
+                  err
+    end
+
+    -- TODO: this should not be needed as the schema should check it already
+    local service_subsystem = constants.PROTOCOLS_WITH_SUBSYSTEM[service.protocol]
+    if service_subsystem ~= subsystem then
+      log(WARN, "service with protocol '", service.protocol,
+                "' cannot be used with '", subsystem, "' subsystem")
```
thibaultcha: Should we bail out here? `return` with no error, simply log the warning and ignore the Route. Otherwise we may end up in situations where a particular subsystem's Route needs to Route into another subsystem (http to stream or vice-versa), which we are not ready to handle yet.

Or what's the rationale for keeping the Service attached to the Route, and do we have any tests covering this edge case (I assume not, since the schema should not even allow this, in which case, we are introducing untested/undefined behavior in our core logic).

Additionally, this way, we don't have to return 3 values from this function anymore, and callers of this function may use `subsystem` (the top-chunk upvalue) instead of `service_subsystem`.

kong/runloop/handler.lua
```diff
@@ -698,25 +710,118 @@ return {
...
+          if pos then
+            service_host = sub(service_host, pos + 1)
+          end
+
+          pos = find(service_host, ":", 2, true)
+          if pos then
+            service_port = sub(service_host, pos + 1)
+            service_host = sub(service_host, 1, pos - 1)
+
+            if re_match(service_port, [[[1-9]{1}\d{0,4}$]], "adjo") then
```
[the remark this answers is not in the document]
thibaultcha: perf: using `ngx.re.find` should be cheaper and achieve the same result in this context.

kong/runloop/handler.lua
bungle: Yes, it was like that, but @james-callahan wanted it to be inverted and it introduced this bug. Damnit. I'll fix it.

kong/runloop/handler.lua
```diff
@@ -698,25 +710,118 @@ return {
...
-        forwarded_port  = var.server_port
+        forwarded_proto = scheme
+        forwarded_host  = host
+        forwarded_port  = port
       end
 
-      local protocols = route.protocols
-      if (protocols and
-          protocols.https and not protocols.http and forwarded_proto ~= "https")
+      local protocols = route and route.protocols
```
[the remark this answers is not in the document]
bungle: removed the `route and `.

kong/runloop/handler.lua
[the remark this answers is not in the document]
bungle: Added error check that logs with warning.

kong/runloop/handler.lua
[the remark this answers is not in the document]
bungle: Added error check that logs with warning, and replaced match with find.

kong/runloop/handler.lua
bungle: Made it return service OR nil, err OR nil (serviceless route).

kong/runloop/handler.lua
```diff
@@ -698,25 +713,125 @@ return {
...
+          if err then
+            log(WARN, "pcre runtime error when matching a request-line: ", err)
+
+          elseif matches then
+            local uri_scheme = lower(matches[1])
+            if uri_scheme == "https" or uri_scheme == "http" then
+              service_scheme = uri_scheme
+              service_host   = lower(matches[2])
+            end
+            --[[ TODO: check if these make sense here?
```
hutchic: left over comment?

bungle: I wanted to leave them here, maybe we return back to that later. I believe we need `Connect` for that to work, but I am not 100% sure.

PR #4288, opened 2019-02-06
fix(admin) proper support for `PUT /{entities}/{entity}/plugins/{plugin}`

thibaultcha:
Prior to this patch, `PUT` requests to plugins under their associated
resources (e.g. `/services/{service}/plugins/{plugin}`) would
systematically return `404`, instead of creating the plugin if it did
not exist. This was because the `before` Lapis hook attempts to retrieve
the plugin in the database before modifying it. Those endpoints are
preserved for backwards-compatibility purposes (being too nested to be
auto-generated), so that workaround is fine.

This patch ensures that we skip this logic with `PUT` requests, plus
makes sure that we set the `self.args.post` arguments according to the
`self.params` arguments (path segments with foreign UUIDs).

Thanks @gszr for the report.

### DIFF
--- a/kong/api/routes/plugins.lua
+++ b/kong/api/routes/plugins.lua
@@ -13,12 +13,15 @@ local setmetatable = setmetatable
 
 
 local get_plugin = endpoints.get_entity_endpoint(kong.db.plugins.schema)
-local put_plugin = endpoints.put_entity_endpoint(kong.db.plugins.schema)
 local delete_plugin = endpoints.delete_entity_endpoint(kong.db.plugins.schema)
 
 
 local function before_plugin_for_entity(entity_name, plugin_field)
   return function(self, db, helpers)
+    if kong.request.get_method() == "PUT" then
+      return
+    end
+
     local entity, _, err_t = endpoints.select_entity(self, db, kong.db[entity_name].schema)
     if err_t then
       return endpoints.handle_error(err_t)
@@ -44,8 +47,10 @@ local function before_plugin_for_entity(entity_name, plugin_field)
 end
 
 
-local function fill_plugin_data(args, plugin)
-  local post = args.post
+local function fill_plugin_data(self, plugin)
+  plugin = plugin or {}
+
+  local post = self.args.post
 
   post.name = post.name or plugin.name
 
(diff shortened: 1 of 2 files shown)


thibaultcha:
@gszr Will you make sure this works on your side? Thanks!

kong/api/routes/plugins.lua
```diff
@@ -13,12 +13,15 @@ local setmetatable = setmetatable
 
 
 local get_plugin = endpoints.get_entity_endpoint(kong.db.plugins.schema)
-local put_plugin = endpoints.put_entity_endpoint(kong.db.plugins.schema)
 local delete_plugin = endpoints.delete_entity_endpoint(kong.db.plugins.schema)
 
 
 local function before_plugin_for_entity(entity_name, plugin_field)
   return function(self, db, helpers)
+    if ngx.req.get_method() == "PUT" then
```
hishamhm: `kong.request.get_method`? :)

bungle:
> . Those endpoints are preserved for backwards-compatibility purposes (being too nested to be auto-generated), so that workaround is fine

I think we might want to make this also autogenerated for all the entities. Let me explain:
- /services/{service}/plugins/

This is the level where autogenerator stops. Now this:
- /services/{service}/plugins/{plugin}

Is not autogenerated but extended. I think we could perhaps open that url just for `PUT`. My reason is that:

- /services/{service}/plugins/
- /services/{service}/plugins/{plugin}

can be considered collection endpoints when doing `PUT`. This makes it possible to do:
`PUT /services/demo/routes/demo-route`

Which is nice. I wouldn't (though I have no strong feeling about it) support entity endpoints on that path anymore (e.g. PATCH, GET), just PUT (but that is debatable). PUT would just complete the POST that we already support.

kong/api/routes/plugins.lua
bungle: `self.req.method`?

thibaultcha: The Admin API code doesn't use `kong.request.get_method()` anywhere yet, and uses more `ngx.req.get_method()` than it uses `self.req.method` (which is actually not documented anywhere I could find in Lapis, hence my aversion against using it).

PR #4290, opened 2019-02-07
fix(ci) build stage order

hutchic:
Need to explicitey define the stage order for travis-ci

Before:

capture

After:

capture2


hutchic:
PR'd to the wrong branch ignore #sorry

PR #4296, opened 2019-02-12
 fix(db) resolve Cassandra contact points when provided as hostnames 

thibaultcha:
Resolve contact points before instantiating cluster.  The driver does
not support hostnames in the contact points list, but we accidentally
support this feature by relying on our globalpatches.lua override of
`tcpsock:connect()`, in which we perform DNS resolution. This could lead
to mismatches between the driver policies and the peers info stored in
the shm.

This scenario is described in details here:

thibaultcha/lua-cassandra@8b9fcd9

We do not rely on `kong.dns` since the `kong` global is not always fully
instantiated in all of our components yet (notably the tests and the
CLI).

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -9,9 +9,36 @@ CassandraConnector.__index = CassandraConnector
 
 
 function CassandraConnector.new(kong_config)
+  local resolved_contact_points = {}
+
+  do
+    -- Resolve contact points before instantiating cluster, since the
+    -- driver does not support hostnames in the contact points list.
+    local dns_tools = require "kong.tools.dns"
+    local dns = dns_tools(kong_config)
+
+    for i, cp in ipairs(kong_config.cassandra_contact_points) do
+      local ip, err = dns.toip(cp)
+      if not ip then
+        log.error("could not resolve Cassandra contact point '%s': %s",
+                  cp, err)
+
+      else
+        log.debug("resolved Cassandra contact point '%s' to: %s", cp, ip)
+        resolved_contact_points[i] = ip
+      end
+    end
+  end
+
+  if #resolved_contact_points == 0 then
+    return nil, "could not resolve any of the provided Cassandra " ..
+                "contact points (cassandra_contact_points = '" ..
+                table.concat(kong_config.cassandra_contact_points, ", ") .. "')"
+  end
+
   local cluster_options       = {
     shm                       = "kong_cassandra",
-    contact_points            = kong_config.cassandra_contact_points,
+    contact_points            = resolved_contact_points,
     default_port              = kong_config.cassandra_port,
     keyspace                  = kong_config.cassandra_keyspace,
     timeout_connect           = kong_config.cassandra_timeout,
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -433,6 +433,14 @@
 
 #cassandra_contact_points = 127.0.0.1  # A comma-separated list of contact
                                        # points to your cluster.
+                                       # You may specify IP addresses or
+                                       # hostnames. Note that the port
+                                       # component of SRV records will be
+                                       # ignored in favor of `cassandra_port`.
+                                       # When connecting to a multi-DC cluster,
+                                       # ensure that contact points from the
+                                       # local datacenter are specified first
+                                       # in this list.
 
 #cassandra_port = 9042           # The port on which your nodes are listening
                                  # on. All your nodes and contact points must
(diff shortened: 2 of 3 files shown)


Tieske:
this resolves each contactpoint to a single ip address. Will this meet expectations? eg. wouldn't someone use a single name resolving to an SRV that carries all the ip's of all the cassandra nodes?

Or will the driver contact the first node only and then get all cassandra peers from that first node?

thibaultcha:
> Or will the driver contact the first node only and then get all cassandra peers from that first node?

This one :) And we only care that the contact points aren't given to the driver as hostnames, there isn't any concern about hosts having multiple records. If the contact point used to retrieve the peers was reachable (it was, since peers are retrieved), the driver will store that IP address, and move on :)

kong/db/strategies/cassandra/connector.lua
```diff
@@ -9,9 +9,36 @@ CassandraConnector.__index = CassandraConnector
...
+  local resolved_contact_points = {}
+
+  do
+    -- Resolve contact points before instantiating cluster, since the
+    -- driver does not support hostnames in the contact points list.
+    local dns_tools = require "kong.tools.dns"
+    local dns = dns_tools(kong_config)
+
+    for i, cp in ipairs(kong_config.cassandra_contact_points) do
+      local ip, err = dns.toip(cp)
```
Tieske: `toip` returns also a port number (in case of SRV records) that is not handled here.

Signature is:
```lua
local ip, port, try_list = toip(name, port, cache_only, try_list)
-- nil, err, try_list will be returned on an error
```

So you probably also want to log the `try_list` in case of errors.

thibaultcha: > `toip` returns also a port number (in case of SRV records) that is not handled here.

We should rely on the `cassandra_port` property for that, it should be the same for all peers in a cluster and properly configured.

Errors are already logged individually by the CLI (see the `log.error` calls), so logging them once again would be redundant, hence we remind the user of which contact points they configured, but try to not be too noisy with a very long error message that will be ignored.

Tieske: > We should rely on the `cassandra_port` property for that, it should be the same for all peers in a cluster and properly configured.

In other places (see the [global patches](https://github.com/Kong/kong/blob/master/kong/globalpatches.lua#L281-L283) for example), the specifically configured port is the default port for A and AAAA records, that port gets overridden by the `toip` provided port in case of an SRV, eg:

```lua
local ip, port, try_list = toip(name, DEFAULT_PORT)
```

If we skip that here, we should note probably that the names in the Cassandra contact points will not be supporting SRV records.

wrt the errors, the calls below will just log "3 - name error", which has limited debugging use. My 2cts (from lessons learnt): append the `tostring(try_list)` to the message (or maybe add another log line, on DEBUG level only)

thibaultcha: > In other places (see the [global patches](https://github.com/Kong/kong/blob/master/kong/globalpatches.lua#L281-L283) for example), the specifically configured port is the default port for A and AAAA records, that port gets overridden by the `toip` provided port in case of an SRV, eg:

Yeah, I am aware. In the case of Cassandra, it is by accident (and it is harmful) that we support the port component of SRV records, as since we've encountered related issues in the past, the decision was made that every peer in a cluster should be listening on the same port, as intended by Cassandra and by other Datastax drivers: https://github.com/thibaultcha/lua-cassandra/commit/1f6668567e55bbb5489d21dd3fd6c10e1105bfc7.

> If we skip that here, we should note probably that the names in the Cassandra contact points will not be supporting SRV records.

Yes, we can also make it explicitly supported (and documented in `kong.conf`) that the contact points list may actually contain hostnames. We can also document there how the order in which the contact points is specified matters, as described in recent issues.

> wrt the errors, the calls below will just log "3 - name error", which has limited debugging use. My 2cts (from lessons learnt): append the `tostring(try_list)` to the message (or maybe add another log line, on DEBUG level only)

This:
```
2019/02/12 10:40:58 [error] could not resolve Cassandra contact point 'unknown': dns server error: 3 name error
```

is much more readable and likely to be read/understood by a user than this:
```
2019/02/12 10:40:47 [error] could not resolve Cassandra contact point 'unknown': ["(short)unknown:(na) - cache-miss","loc:33 - cache-miss/scheduled/querying/dns server error: 3 name error","unknown:1 - cache-miss/scheduled/querying/dns server error: 3 name error","unknown:5 - cache-miss/scheduled/querying/dns server error: 3 name error"]
```

Tieske: > is much more readable and likely to be read/understood by a user than this:

more likely to be read, yes. To be understood no, because there is little to understand. And usually that ends up with some sort of support request, and for us there is no info to help the user out. Hence my suggestion to add an additional line with `DEBUG` level, so we can at least ask the user to retry with that level of logging and get the details.

I leave at your discretion to dismiss the review and merge.

thibaultcha: Added a debug level log after the error one which show the `try_list` :+1:

PR #4297, opened 2019-02-12
chore(ci) fix nightly builds

hutchic:
kong-build-tools includes a functional test for the Kong version that's in the header `Via: "kong/x.y.z`. https://github.com/Kong/kong-build-tools/blob/master/test/test_smoke.tavern.yaml#L82

To make this test pass when building a nightly build we `sed` the `meta.lua` file so the Kong version matches. This also has the benefit that nightly builds will return the date/time stamp when one runs `Kong version` instead of the last released kong version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,19 +73,19 @@ jobs:
       script: .ci/run_tests.sh
     - stage: deploy daily build
       install: skip
-      script: make release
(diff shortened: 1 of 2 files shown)


thibaultcha:
Thank you!

PR #4300, opened 2019-02-12
refactor(db) de-hardcode custom strategies list

hishamhm:
The implementation of a core entity consists of up to three parts:

1. the schema of the entity (`kong.db.schema.entities.my_entity`)
2. a DB-agnostic DAO module implementing custom methods for that entity or overriding standard methods in a DB-agnostic manner (`kong.db.dao.my_entity`)
3. one or more DB-specific modules implementing the database backend
   for custom methods, or overriding standard methods in a DB-specific manner (`kong.db.strategies.postgres.my_entity` and    `kong.db.strategies.cassandra.my_entity`)

Instead of using a hardcoded list of customized DB-specific override modules (item 3 above) with this commit we simply attempt to load the module for every entity, and use it if it is available.

This is the backwards-compatible solution to be merged to `master`. Once this is in, a subsequent PR will be made to `next` that adds the ability of loading DB-specific strategies for custom entities in plugins (hi @p0pr0ck5!) in about 3 lines :)

### DIFF
--- a/kong/db/strategies/init.lua
+++ b/kong/db/strategies/init.lua
@@ -1,3 +1,6 @@
+local utils = require("kong.tools.utils")
+
+
 local fmt = string.format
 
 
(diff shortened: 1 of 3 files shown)


kong/db/strategies/init.lua
```diff
@@ -51,31 +51,29 @@ function _M.new(kong_config, strategy, schemas, errors)
...
-            -- override
-            local f = custom_strategy[k]
-            if f then
-              return f
-            end
-
-            -- parent fallback
-            return parent_mt[k]
+    local custom_strat = fmt("kong.db.strategies.%s.%s", strategy, schema.name)
+    local pok, mod = pcall(require, custom_strat)
```
thibaultcha: We have a [load_module_if_exists](https://github.com/Kong/kong/blob/master/kong/tools/utils.lua#L576) utility for this which is more helpful, since it will throw an error when the module fails to load for other reasons than not existing (e.g. syntax errors).

hishamhm: good call!

kong/runloop/balancer.lua
```diff
@@ -473,8 +473,13 @@ do
   -- and upstream entity tables as values, or nil+error
   local function load_upstreams_dict_into_memory()
     local upstreams_dict = {}
-    for up in singletons.db.upstreams:each(1000) do
     -- build a dictionary, indexed by the upstream name
+    for up, err in singletons.db.upstreams:each(1000) do
+      if err then
+        log(ngx.CRIT, "could not ensure obtain list of upstreams: ", err)
```
thibaultcha: This doesn't read very nicely; was a word omitted?

hishamhm: oops, it wasn't meant to be in this PR

thibaultcha: Even better :+1:

PR #4301, opened 2019-02-12
fix(balancer) avoid crash when failing to obtain list of upstreams

hishamhm:
Avoid a stacktrace when the database fails to obtain the list of upstreams (as
seen in an error log provided by @hutchic): instead report a CRIT-level
message and fallback to an empty list.

```
2019/02/12 17:38:18 [crit] 36#0: *7 [lua] balancer.lua:719: init(): failed loading initial list of upstreams: failed to get from node cache: callback threw an error: /usr/local/share/lua/5.1/kong/runloop/balancer.lua:476: attempt to index local 'up' (a
boolean value)
stack traceback:
       /usr/local/share/lua/5.1/kong/runloop/balancer.lua:476: in function </usr/local/share/lua/5.1/kong/runloop/balancer.lua:472>
       [C]: in function 'xpcall'
       /usr/local/share/lua/5.1/resty/mlcache.lua:766: in function 'get'
       /usr/local/share/lua/5.1/kong/cache.lua:126: in function 'get'
       /usr/local/share/lua/5.1/kong/runloop/balancer.lua:489: in function 'get_all_upstreams'
       /usr/local/share/lua/5.1/kong/runloop/balancer.lua:717: in function 'init'
       /usr/local/share/lua/5.1/kong/runloop/handler.lua:565: in function </usr/local/share/lua/5.1/kong/runloop/handler.lua:564>, context: ngx.timer
```

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -466,15 +466,15 @@ end
 
 local get_all_upstreams
 do
-  ------------------------------------------------------------------------------
-  -- Implements a simple dictionary with all upstream-ids indexed
-  -- by their name.
-  -- @return The upstreams dictionary, a map with upstream names as string keys
-  -- and upstream entity tables as values, or nil+error
   local function load_upstreams_dict_into_memory()
     local upstreams_dict = {}
-    for up in singletons.db.upstreams:each(1000) do
     -- build a dictionary, indexed by the upstream name
+    for up, err in singletons.db.upstreams:each(1000) do
+      if err then
+        log(CRIT, "could not obtain list of upstreams: ", err)
+        return nil
+      end
+
       upstreams_dict[up.name] = up.id
     end
     return upstreams_dict
@@ -482,19 +482,22 @@ do
   _load_upstreams_dict_into_memory = load_upstreams_dict_into_memory
 
 
+  local opts = { neg_ttl = 10 }
+
+
   ------------------------------------------------------------------------------
-  -- Finds and returns an upstream entity. This function covers
-  -- caching, invalidation, db access, et al.
-  -- @param upstream_name string.
-  -- @return upstream table, or `false` if not found, or nil+error
+  -- Implements a simple dictionary with all upstream-ids indexed
+  -- by their name.
+  -- @return The upstreams dictionary (a map with upstream names as string keys
+  -- and upstream entity tables as values), or nil+error
   get_all_upstreams = function()
-    local upstreams_dict, err = singletons.cache:get("balancer:upstreams", nil,
+    local upstreams_dict, err = singletons.cache:get("balancer:upstreams", opts,
                                                 load_upstreams_dict_into_memory)
     if err then
       return nil, err
     end
 
-    return upstreams_dict
+    return upstreams_dict or {}
   end
 end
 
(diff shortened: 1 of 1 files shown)


kong/runloop/balancer.lua
```diff
@@ -473,8 +473,13 @@ do
   -- and upstream entity tables as values, or nil+error
   local function load_upstreams_dict_into_memory()
     local upstreams_dict = {}
-    for up in singletons.db.upstreams:each(1000) do
     -- build a dictionary, indexed by the upstream name
+    for up, err in singletons.db.upstreams:each(1000) do
+      if err then
+        log(CRIT, "could not obtain list of upstreams: ", err)
+        return upstreams_dict
```
thibaultcha: It may be more semantically correct to return `nil, err` here, since this is an mlcache callback. This way, we will avoid caching the result, and properly bubble up the error to `get_balancer()`, which will return `nil, err` as well, instead of `false`. Or was the return value of `get_balancer()` intentional? Currently, if other mlcache callback fail, they will make `get_balancer()` return `nil` and thus produce a 500 error.

Another approach would be to return `nil` here (no `err`), and thus respect a very short `neg_ttl` value (which is presently equal to `ttl` but we could override it for this call). Outside of the callback (maybe in `get_upstream_by_name()`), we can handle the `nil` return value appropriately.

thibaultcha: The idea that I get here is that we are attempting to fallback into a "no balancer" mode (as if no upstream was configured for this Service). Which is great, but since it is sub-optimal behaviour (for resiliency purposes), I feel like the second proposed approach is favourable.

The first approach only makes sense when we know there is an upstream, but we were just unable to fetch it (that said, we could be resilient and fallback to "no balancer" mode there as well, after all) :+1:

hishamhm: You just read my mind: I was just going to ask whether it would be possible to not cache this to allow for retries or to set a ttl so that we could retry in case of an intermittent failure without hammering the database.

I pushed an updated version. What's a reasonable "very short `neg_ttl`"?

thibaultcha: > What's a reasonable "very short `neg_ttl`"?

Not a question that can be answered simply... tldr; 10s (the default you chose) seems reasonable. Otherwise, retry strategies play an important part in building resilient software (the C* driver has had constant and exponential retry strategies for a long time, but there are other Kong components, or first class citizens, that need it as well). Until then...

kong/runloop/balancer.lua
```diff
@@ -483,18 +483,19 @@ do
...
-  -- caching, invalidation, db access, et al.
-  -- @param upstream_name string.
-  -- @return upstream table, or `false` if not found, or nil+error
+  -- Implements a simple dictionary with all upstream-ids indexed
+  -- by their name.
+  -- @return The upstreams dictionary (a map with upstream names as string keys
+  -- and upstream entity tables as values), or nil+error
   get_all_upstreams = function()
-    local upstreams_dict, err = singletons.cache:get("balancer:upstreams", nil,
+    local opts = { neg_ttl = 10 }
```
thibaultcha: We could avoid extra GC cycles if we didn't allocate this table inside this relatively hot function.

hishamhm: ah yes, for a second I misread where this function was being called from and thought it wasn't worth it — you're right, fixed

PR #4304, opened 2019-02-13
Add kubernetes sidecar injector plugin

james-callahan:
This PR adds the kubernetes sidecar injector plugin

Released here: https://github.com/Kong/kubernetes-sidecar-injector/releases/tag/v0.1.0

PR #4305, opened 2019-02-13
feat(plugins) add kubernetes-sidecar-injector plugin

bungle:
### Summary

Adds `kong-plugin-kubernetes-sidecar-injector` to dependencies.


james-callahan:
What was wrong with #4304?

PR #4307, opened 2019-02-13
feat(pdk) add get_subsystem and get_protocol functions

kikito:
Adds `kong.request.get_subsystem` and `kong.request.get_protocol`

### DIFF
--- a/kong/pdk/client.lua
+++ b/kong/pdk/client.lua
@@ -199,6 +199,22 @@ local function new(self)
   end
 
 
+  ---
+  -- Returns the Nginx subsystem used by the client. It can be
+  -- "http" or "stream"
+  --
+  -- @function kong.client.get_subsystem
+  -- @phases access, header_filter, body_filter, log
+  -- @treturn string a string with either `"http"` or `"stream"`
+  -- @usage
+  -- kong.client.get_subsystem() -- "http"
+  function _CLIENT.get_subsystem()
+    check_phase(AUTH_AND_LATER)
+
+    return ngx.config.subsystem
+  end
+
+
   return _CLIENT
 end
 
--- a/kong/pdk/client.lua
+++ b/kong/pdk/client.lua
@@ -7,6 +7,7 @@
 -- @module kong.client
 
 
+local utils = require "kong.tools.utils"
 local phase_checker = require "kong.pdk.private.phases"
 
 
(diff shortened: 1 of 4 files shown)


bungle:
The only concern with me is that `request` module is very `http`, even to the point that I would say, use it only with `http`. `Streams` don't have requests or responses. That's why I was thinking maybe we need to make another one, like `kong.connection.get_subsystem()` / `kong.connection.get_protocol()`.

I already mentioned this to @kikito, so I think this is more a question to other members or community?

kikito:
An alternative to @bungle's suggestion would be to put these in the `kong.client` module, that might also work.

hishamhm:
+1 to `kong.client`, that works IMO

kikito:
Moved functions to `kong.client`

hishamhm:
@kikito there's a PDK phase check error in the testsuite (which doesn't look like a flaky test)

kong/pdk/client.lua
```diff
@@ -199,6 +200,63 @@ local function new(self)
...
+  -- @function kong.client.get_subsystem
+  -- @phases access, header_filter, body_filter, log
+  -- @treturn string a string with either `"http"` or `"stream"`
+  -- @usage
+  -- kong.client.get_subsystem() -- "http"
+  function _CLIENT.get_subsystem()
+    check_phase(AUTH_AND_LATER)
+
+    return ngx.config.subsystem
+  end
```
thibaultcha: Which subsystem is in use is an implementation detail of Nginx, not a client's choice (unlike choosing a protocol). It makes sense that `get_protocol()` be in this module, but `get_subsystem()` should be a top-level API: `kong.get_subsystem()` answers the question "which Nginx subsystem is my code currently executing within?"

t/01-pdk/05-client/09-get_protocol.t
```diff
@@ -0,0 +1,314 @@
...
+                        ngx.var.server_addr    .. " " ..
+                        ngx.var.remote_port    .. " " ..
+                        ngx.var.server_port    .. "\r\n"
+        sock:send(request)
+        ngx.print(sock:receive())
+    }
+--- stream_response chop
+protocol=tls
+--- no_error_log
+[error]
```
thibaultcha: We are missing test cases ensuring that these APIs (to be called in very hot code paths) actually JIT compile.

kong/pdk/client.lua
```diff
@@ -199,6 +200,63 @@ local function new(self)
...
+  -- Returns the Nginx subsystem used by the client. It can be
+  -- "http" or "stream"
+  --
+  -- @function kong.client.get_subsystem
+  -- @phases access, header_filter, body_filter, log
+  -- @treturn string a string with either `"http"` or `"stream"`
+  -- @usage
+  -- kong.client.get_subsystem() -- "http"
+  function _CLIENT.get_subsystem()
+    check_phase(AUTH_AND_LATER)
```
thibaultcha: We should be able to call this API from `init_by_lua`, `init_worker_by_lua` as well, which further highlights the need for this API to be a root API.

kong/pdk/client.lua
james-callahan: I thought some of the PDK was meant to be decoupling from nginx?
It feels like an implementation detail that http and stream workers aren't shared.

thibaultcha: That is precisely my point.

james-callahan: I don't follow your argument: shouldn't code be asking "what is the relevant request type for the current client?" rather than "what subsystem is this code running in?"
The former is always a sensible question to ask; the latter only makes sense as long as there are different lua VMs for http vs stream.

thibaultcha: > what is the relevant request type for the current client?

get_protocol() should answer this question.

However, the subsystem isn’t tied to a client, but is an nginx/OpenResty implementation detail. It is necessary to know the current subsystem in contexts where’s no clients are connected. See my other comment about this.

This necessity for 2 APIs (protocol _and_ subsystem) was discussed during a Core Team spoke a couple of months ago. Ultimately subsystem should be less and less relevant. It is 11PM on a Sunday, so forgive the brevity here.

PR #4309, opened 2019-02-14
Bump lua-http

james-callahan:
lua-http 0.3 was released yesterday, this PR bumps it in our rockspec

lua-http maintains lists of mozilla recommended cipher suites, e.g. https://daurnimator.github.io/lua-http/0.3/#http.tls.old_cipher_list
I've replaced `kong.tools.ciphers` with use of their lists.


Tieske:
Is the contents of the ciphers list the same? if not identical, what could be the negative side effects of replacing this? could it be breaking for some users?

james-callahan:
> Is the contents of the ciphers list the same?

yes.

james-callahan:
> if not identical, what could be the negative side effects of replacing this? could it be breaking for some users?

Some extra context: the "modern/intermediate/old" cipher suites terminology refer to those in this mozilla maintained page: https://wiki.mozilla.org/Security/Server_Side_TLS
Any user specifying "modern" should expect a recent revision of whatever is on that mozilla page and whatever breakage that entails.

PR #4315, opened 2019-02-14
Declarative config for DB-import and DB-less Kong

### LINKED ISSUE #4264: [community feedback requested] Kong native Declarative Config format
Hi everyone,

We are in the process of designing a native declarative configuration format to be used in Kong (a long time community ask: #528, #1465, ...). We just posted a first draft in Kong Nation:

https://discuss.konghq.com/t/rfc-kong-native-declarative-config-format/2719

Any and all feedback is welcome!

hishamhm:
Opening this branch for review.
This PR contains 3 main parts:

1. **Declarative config format** ([subset of changes](https://github.com/kong/kong/compare/next...667f777)) - implementation of the declarative config format presented in #4264. This exercised the schema library intensely, so a number of fixes to it emerged from this work. Some notable changes:
    * it no longer stops at the first failed element of an array, so this caused some changes in the error messages
    * the `url` pseudo-attribute is no longer handled in the Admin API code, but in the schema library code as a new feature, "shorthands", so that its logic can be reused in the Admin API and declarative config
    * introduces a new dependency, lyaml, which uses the C library libyaml.
2. **Database import** ([subset of changes](https://github.com/kong/kong/compare/667f777...cce7064)) - This PR includes the contents of #4284, which was already reviewed and included in this branch.
3. **Database-less mode** ([subset of changes](https://github.com/kong/kong/compare/cce7064...feat/dbless-cache)) - A new mode, `database=off`, which allows to run Kong off a declarative config file only, without a backing database, using the in-memory cache as the only object store.
    * introduces a new `kong.concurrency` module, which removes duplication of delicate concurrency code that was reused in multiple places in the codebase, making it easier to get concurrency right and avoid bugs
    * introduces the `dc_blueprints` helper for tests, which builds up a declarative config allowing for tests to be adapted for DB-less mode with minimal changes — the abstraction introduced by the Blueprints library pays for itself once again!
   * it restricts the Admin API of the DB-less Kong node into a "read-only mode": this was implemented in #4308, and was already reviewed and merged into this branch.

User-visible changes:

* CLI:
  * new `kong config` command for declarative config and DB-import
    * `kong config parse` for validating config files
    * `kong config import` for DB-import (#4284)
* Configuration:
  * `database=off` for selecting DB-less mode
  * `declarative_config=<filename>` in `kong.conf` for selecting a declarative config file in DB-less mode
* Admin API:
  * `/configure` endpoint for loading or reloading the declarative configuration in DB-less mode
  * the Admin API is read-only when using `database=off` (#4308)

Closes #4264.


.travis.yml
```diff
@@ -37,6 +37,7 @@ env:
     - KONG_TEST_DATABASE=postgres TEST_SUITE=integration
     - KONG_TEST_DATABASE=cassandra CASSANDRA=2.2 TEST_SUITE=integration
     - KONG_TEST_DATABASE=cassandra CASSANDRA=3.9 TEST_SUITE=integration
+    - KONG_TEST_DATABASE=off TEST_SUITE=dbless
```
kikito: Presumably db-less is faster than pg or C*? Maybe we could make Travis-ci to run db-less first, and if it passes, then run the other dbs, in a similar way to what it does with unit tests.

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,108 @@
...
+
+local off = {}
+
+
+local _mt = {}
+_mt.__index = _mt
+
+
+local function page_for_key(self, key, size, offset)
+  offset = offset and tonumber(offset) or 1
```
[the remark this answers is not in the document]
bungle: my commit already changes it.

kong/db/schema/init.lua
```diff
@@ -1359,59 +1384,82 @@ function Schema:process_auto_fields(input, context, nulls)
   end
   --]]
 
-  for key, field in self:each_field(input) do
+  if self.shorthands then
+    for _, shorthand in ipairs(self.shorthands) do
+      local sname = next(shorthand)
+      local sfunc = shorthand[sname]
```
james-callahan: This should just be `local sname, sfunc = next(shorthand)`

kong/db/schema/init.lua
```diff
@@ -1823,14 +1871,37 @@ function Schema.new_subschema(self, key, definition)
...
 
   local subschema, err = Schema.new(definition, true)
   if not subschema then
     return nil, err
   end
 
+  local parent_by_name = {}
+  for _, f in ipairs(self.fields) do
+    local fname = next(f)
+    local fdata = f[fname]
```
james-callahan: Should be `local fname, fdata = next(f)`

kong/db/schema/others/declarative_config.lua
```diff
@@ -0,0 +1,438 @@
...
+    local out = {}
+    for _, k in ipairs(schema.primary_key) do
+      table.insert(out, tostring(object[k]))
+    end
+    return table.concat(out, ":")
+  end
+end
+
+
+--[[
```
james-callahan: Do we want to leave this in?

kong/db/schema/plugin_loader.lua
```diff
@@ -0,0 +1,238 @@
...
+local insert = table.insert
+local ipairs = ipairs
+
+
+--- Check if a string is a parseable URL.
+-- @param v input string string
+-- @return boolean indicating whether string is an URL.
+local function validate_url(v)
+  if v and type(v) == "string" then
+    local url = require("socket.url").parse(v)
```
james-callahan: Move this `require` to top of file?

kong/db/schema/topological_sort.lua
```diff
@@ -0,0 +1,106 @@
...
+-- schema B has a foreign key to A, then B appears after A
+-- The function returns an error if cycles are found in the schemas
+-- (i.e. A has a foreign key to B and B to A)
+-- @tparam array schemas an array with zero or more schemas
+-- @treturn array|nil an array of schemas sorted topologically, or nil if cycle was found
+-- @treturn nil|string nil if the schemas were sorted, or a message if a cycle was found
+-- @usage
+-- local res = topological_sort({ services, routes, plugins, consumers })
+-- assert.same({ consumers, services, routes, plugins }, res)
+local topological_sort = function(schemas)
```
james-callahan: Why define like this instead of the normal `local function topological_sort(schemas)`

kong/db/strategies/off/connector.lua
```diff
@@ -0,0 +1,82 @@
...
+
+
+function OffConnector:query()
+  return nil, "cannot perform queries without a database"
+end
+
+
+do
+  -- migrations
+
```
james-callahan: What is the purpose of this `do` block?
Also, the whitespace here seems weird to me, though it might be to style guide.

kong/db/strategies/off/init.lua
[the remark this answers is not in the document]
bungle: @hishamhm / @kikito but where it is now?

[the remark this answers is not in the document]
bungle: See: https://github.com/Kong/kong/pull/4308/files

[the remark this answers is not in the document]
hishamhm: @bungle sorry, messed up a force-push in this branch on Friday. All should be back to normal.

kong/runloop/mesh.lua
```diff
@@ -82,6 +82,12 @@ local function init()
     if not ca_row and err == nil then
       -- No CA cert in DB, we are the first cluster member to start
 
+      if kong.db.strategy == "off" then
+        ngx.log(ngx.WARN, "no cluster_ca in declarative configuration: " ..
+                          "cannot use node in mesh mode")
+        return
```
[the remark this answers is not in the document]
hishamhm: No, because otherwise you can't use DB-less in non-mesh mode without a huge `cluster_ca` entity in the YAML file.

kong/api/routes/configure.lua
```diff
@@ -0,0 +1,28 @@
...
+        kong.response.exit(400, {
+          message = "this endpoint is only available when Kong is " ..
+                    "configured to not use a database"
+        })
+      end
+
+      local dc = declarative.init(kong.configuration)
+
+      local yaml = self.params.config
+      local entities, err = dc:parse_string(yaml, "config.yml", { yaml = true })
```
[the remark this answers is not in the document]
hishamhm: YAML is a superset of JSON, so we actually accept both here. Lua is not accepted here because of security reasons (loading a Lua file can hang the process; in the CLI this is not a problem).

I'd still like to get the actual filename from the request though to trigger proper JSON errors. I'll mark this a TODO.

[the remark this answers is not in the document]
Tieske: Iirc Penlight comes with a sandbox, can't we use that and have parity between cli and rest interfaces?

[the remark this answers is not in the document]
hishamhm: I can take a look at it later, but any sandbox for safely loading Lua depends on the `debug` library for hooks, which is notoriously buggy in LuaJIT. For now I'll keep Lua for the CLI only.

kong/init.lua
```diff
@@ -337,7 +377,9 @@ function Kong.init()
     certificate.init()
   end
 
-  mesh.init()
+  if kong.configuration.database ~= "off" then
```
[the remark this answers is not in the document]
hishamhm: this is not disallowing service mesh. For DB-less, `mesh.init()` is just called from a different place because of the different order things need to happen.

kong/db/declarative/init.lua
```diff
@@ -136,4 +175,97 @@ function declarative.load_into_db(dc_table)
 end
 
 
+local function remove_nulls(tbl)
```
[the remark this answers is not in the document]
hishamhm: I considered it, but:

* it is used only once in Kong proper (the other use is in the testsuite)
* I don't want people to think that this is the normal way of getting rid of `ngx.null` in responses (it is `opts.nulls`). The reason why we do this here is because this is a special situation (I didn't want to mess with the invariant in the schema library and DAO that `"insert"` mode always produces nulls).
* I'm trying not to fill `kong.tools.utils` with more things since it is already the bag of random stuff :)

kong/db/schema/others/declarative_config.lua
hishamhm: it was pretty useful for debugging and the code in this module is tricky. I think the next person who needs to make changes to this module would appreciate having that around. In very particular cases we have kept commented-out code around, I thought this was one of them. But I can remove it if truly necessary.

kong/db/strategies/off/connector.lua
hishamhm: it was just to mirror the structure of `kong.db.strategies.cassandra.connector`.

> Also, the whitespace here seems weird to me, though it might be to style guide.

Agreed and yes.

kong/db/strategies/off/connector.lua
```diff
@@ -0,0 +1,82 @@
...
+    strategy = "off",
+    db_name = "kong_off",
+    db_desc = "cache",
+    db_ver = meta._VERSION,
+  }
+end
+
+
+function OffConnector:connect_migrations(opts)
+  -- FIXME what is this used for, again?
```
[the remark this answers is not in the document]
hishamhm: Done, thanks!

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,108 @@
...
+    connector = connector, -- instance of kong.db.strategies.off.connector
+    schema = schema,
+    errors = errors,
+    page = page,
+    select = select,
+    select_by_field = select_by_field,
+    truncate = function() return true end,
+  }
+
+  -- TODO implement page_for_* methods,
```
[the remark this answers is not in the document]
hishamhm: Yep! Removed, thanks!

kong/init.lua
```diff
@@ -232,6 +233,45 @@ local function flush_delayed_response(ctx)
...
+  return concurrency.with_worker_mutex(opts, function()
+    local _, _, value = kong.cache:probe("declarative_config:loaded")
+    if value then
+      return true
+    end
+
+    local dc = declarative.init(config)
+    local entities, err = dc:parse_file(config.declarative_config)
+    if not entities then
+      return nil, config.declarative_config .. ": " .. require"inspect"(err)
```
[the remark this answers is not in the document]
hishamhm: Fixed to the best extent I could! For the ideal result, we'd need something like https://github.com/gvvaughan/lyaml/issues/26

.travis.yml
hishamhm: The dbless suite is very incomplete as opposed to the others right now so it wouldn't really gain us a lot of insight. Also, splitting phases has an overhead in which it delays the startup of the next jobs, which is considerable. So mayyybe something to consider once the dbless suite is more complete, but I wouldn't do it right now.

kong/db/schema/init.lua
hishamhm: Fixed! I often forget about `next`'s second return!

kong/db/schema/init.lua
hishamhm: ditto

kong/db/schema/plugin_loader.lua
hishamhm: done

kong/db/schema/others/declarative_config.lua
```diff
@@ -0,0 +1,438 @@
...
+
+  schema.known_entities = known_entities
+  schema.flatten = flatten
+
+  return schema, nil, def
+end
+
+
+return DeclarativeConfig
+
```
james-callahan: extra new line at end of file

kong/db/schema/topological_sort.lua
```diff
@@ -0,0 +1,106 @@
+local table_insert = table.insert
+
+-- Given an array of schemas, build a table were the keys are schemas and the values are
```
james-callahan: This seems like quite a general topological sort module. Perhaps move to utils? (or even it's own luarock at some point?)

kong/init.lua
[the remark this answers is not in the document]
james-callahan: What is the different order required? It was placed here because some plugins needed mesh initialised before they could load.

kong/db/schema/topological_sort.lua
bungle: We have another here:
https://github.com/Kong/kong/blob/master/kong/db/strategies/postgres/connector.lua#L65-L109

Which is about the same as:
https://github.com/bungle/lua-resty-tsort

There is also:
https://github.com/starius/toposort/

kong/init.lua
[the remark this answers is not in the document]
hishamhm: We can only have entities in DB-less after we have a cache, and we only have the cache in `init_worker`. When mesh is used, the `cluster_ca` object needs to be fed in the `config.yml` file (as the nodes don't talk to each other, there's no DB or cluster_events).

hishamhm:
All comments so far were addressed. 

Also, renamed `kong config import` to `kong config db_import` and the `/configure` endpoint to `/config`, as per conversations with @thibaultcha and @guanlan.

kong/init.lua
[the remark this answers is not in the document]
james-callahan: The "cluster_ca" entity is singular and global, does that limitation still apply?

kong/db/schema/topological_sort.lua
hishamhm: good to know! in the interest of expediency, let's consider refactoring this later

kong/db/strategies/off/connector.lua
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  return setmetatable(self, OffConnector)
+end
+
+
+function OffConnector:infos()
+  return {
+    strategy = "off",
+    db_name = "kong_off",
```
thibaultcha: `"in memory"`?

kong/db/strategies/off/connector.lua
```diff
@@ -0,0 +1,80 @@
...
+
+
+function OffConnector:is_014()
+  return {
+    is_014 = false,
+  }
+end
+
+
+function OffConnector:get_timeout()
```
thibaultcha: This seems unused (I was about to point out that it need to be added to the `kong/db/connector/init` super class otherwise)

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,150 @@
...
+  else
+    offset = 1
+  end
+
+  if not kong.cache then
+    return {}
+  end
+
+  local list = kong.cache:get(key, nil, function()
+    return {}
```
thibaultcha: Should we return `nil, err` here? So that we do not cache the return value, and we actually enter the below `if not list` branch?

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,150 @@
...
+    end
+
+    offset = token
+
+  else
+    offset = 1
+  end
+
+  if not kong.cache then
+    return {}
```
thibaultcha: Why not returning an error in those cases?

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,150 @@
...
+  end
+
+  local list = kong.cache:get(key, nil, function()
```
thibaultcha: No error handling here?

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,150 @@
...
+    return {}
+  end)
+  if not list then
+    return nil
```
thibaultcha: Are this branch and this return value correct? Are we missing the second return value from `:get()` being returned from here as well? Should we return `{}`? Maybe it'd help if you could clarify your intentions here (so far I am confused by this function's implementation, it seems).

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,150 @@
...
+  local name = self.schema.name
+  for i = 0, size - 1 do
+    local id = list[i + offset]
+    if not id then
+      offset = nil
+      break
+    end
+
+    local ck = name .. ":" .. id .. "::::"
+    local entry = kong.cache:get(ck, nil, function()
```
thibaultcha: I strongly suggest making those callbacks local functions of their own, instead of declaring anonymous closures in such relatively hot code paths.

kong/db/declarative/init.lua
```diff
@@ -0,0 +1,295 @@
...
+      end
+    end
+
+    local ids = {}
+    for id, item in pairs(items) do
+      table.insert(ids, id)
+
+      local cache_key = dao:cache_key(id)
+      item = remove_nulls(item)
+      kong.cache:get(cache_key, nil, function()
```
thibaultcha: No error handling here? Also, we should populate from the mlcache instance directly, which seems more correct (and will also avoid error messages such as `"failed to get from node cache: ..."` in cases of failures during db-less writes.

kong/db/schema/init.lua
```diff
@@ -1288,39 +1295,56 @@ local function make_set(set)
 end
 
 
+local function should_recurse_record(context, value, field)
+  if context == "update" then
+    return value ~= null and value ~= nil
+  else
```
thibaultcha: Unnecessary branch here

kong/db/strategies/off/init.lua
```diff
@@ -0,0 +1,150 @@
...
+function off.new(connector, schema, errors)
+  local unsupported = function(operation)
+    local err = fmt("cannot %s '%s' entities when not using a database", operation, schema.name)
+    return function()
+      return nil, errors:operation_unsupported(err)
+    end
+  end
+
+  local unsupported_by = function(operation)
+    local err = fmt("cannot %s '%s' entities by '%s' when not using a database", operation, schema.name, '%s')
```
thibaultcha: style: respect the 80 columns limit?

kong/init.lua
```diff
@@ -247,6 +234,57 @@ local function flush_delayed_response(ctx)
...
+
+  if not config.declarative_config then
+    -- no configuration yet, just build empty plugins map
+    build_plugins_map(kong.db, utils.uuid())
+    return true
+  end
+
+  local opts = {
+    name = "declarative_config",
+    timeout = 3600,
```
thibaultcha: We need to specify `exptime` here as well, or else this timeout setting will not take effect. Also, a 1h timeout? Really? I am very surprised.

kong/init.lua
```diff
@@ -352,21 +390,42 @@ function Kong.init()
...
+      timeout = timeout,
+    }
   end
 
-  local _, err = build_plugins_map(db, "init")
-  if err then
-    error("error building initial plugins map: ", err)
+  if kong.configuration.database == "off" then
+    local err
+    declarative_entities, err = parse_declarative_config(config)
```
thibaultcha: Nitpick that will help future selves: please us `kong.configuration` here. Typically we use `kong_config` when we refer to the `kong.conf` parsed table, for clarity. For the sake of consistency (and in the same spirit of explicitness), we should keep avoiding using `config` nowdays.

kong/init.lua
```diff
@@ -247,6 +234,57 @@ local function flush_delayed_response(ctx)
 end
 
 
+local function parse_declarative_config(config)
```
thibaultcha: As per my comment below, please rename this argument to `kong_config`

kong/runloop/mesh.lua
```diff
@@ -82,6 +82,12 @@ local function init()
     if not ca_row and err == nil then
       -- No CA cert in DB, we are the first cluster member to start
 
+      if kong.db.strategy == "off" then
+        ngx.log(ngx.WARN, "no cluster_ca in declarative configuration: " ..
```
thibaultcha: Better use the variadic form of this function for the sake of writing reference code

kong/runloop/mesh.lua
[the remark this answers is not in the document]
thibaultcha: Besides, throwing an error in `init_worker` (when `mesh.init()` is invoked when using DB-less) will not stop execution, the worker will log it and move on.

kong/init.lua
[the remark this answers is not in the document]
hishamhm: yes, because the cache is the backing store for db-less. Which plugin you have in mind that won't work with the current implementation?

kong/cmd/config.lua
```diff
@@ -0,0 +1,108 @@
...
+  local dc, err = declarative.init(conf)
+  if not dc then
+    log(err)
+    os.exit(1)
+  end
+
+  if args.command == "import" or args.command == "parse" then
+    if not args.file then
+      log("expected a declarative declarative configuration file; see `kong config --help`")
+      os.exit(1)
```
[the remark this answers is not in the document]
hishamhm: got it. this was a slip, in other places I was using `log.error` and `os.exit(1)`. Will switch.

kong/db/dao/plugins.lua
```diff
@@ -143,156 +137,96 @@ function Plugins:check_db_against_config(plugin_set)
...
-    insert(new_schema.fields, { consumer = typedefs.no_consumer })
+
+  ngx_log(ngx_DEBUG, "Loading plugin: " .. plugin)
+
+  plugin_list[#plugin_list+1] = {
+    name = plugin,
+    handler = handler(),
+  }
+
+  if db.strategy then -- skip during tests
```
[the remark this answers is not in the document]
hishamhm: I suppose it's for unit tests. This code was already here, I did not change this line.

kong/init.lua
hishamhm: > We need to specify exptime here as well, or else this timeout setting will not take effect. 

The library sets `exptime`.

> Also, a 1h timeout? Really? I am very surprised.

This is initialization code and there's no recovery from a timeout in this situation, so the timeout just needs to be high enough to cover every possible valid scenario, so this is effectively meant "no timeout". The arbitrary default used elsewhere in the code is 1 minute. I can just use that, so at least the user gets an error message sooner if they have the patience to wait while the system fails to start.

thibaultcha: > The library sets `exptime`.

Oh, your abstraction in `concurrency.lua`, I see :+1:

kong/db/init.lua
```diff
@@ -318,90 +306,74 @@ do
       ttl = DEFAULT_LOCKS_TTL
     end
 
-    local rlock, err = resty_lock:new("kong_locks", {
-      exptime = ttl,
+    local mutex_opts = {
+      name = key,
```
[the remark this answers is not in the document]
hishamhm: it is set by the library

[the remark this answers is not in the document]
thibaultcha: By your abstraction of it* :)

kong/db/schema/plugin_loader.lua
```diff
@@ -0,0 +1,239 @@
...
+  end
+
+  return schema
+end
+
+
+function plugin_loader.load_entity_schema(plugin, schema_def)
+  local _, err_t = MetaSchema:validate(schema_def)
+  if err_t then
+    return nil, fmt("schema of custom plugin entity '%s.%s' is invalid: %s",
```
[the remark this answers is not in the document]
hishamhm: fixed

kong/db/init.lua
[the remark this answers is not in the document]
hishamhm: the kong.concurrency library

kong/db/strategies/off/init.lua
hishamhm: no, this is meant to return `{}`. If you are trying to paginate on an entity that has no elements (e.g. you have no consumers and you hit `/services/mysrv/routes` and `mysrv` has no routes), the cache will be empty with no `list` entry for it, and the correct answer is an empty list.

kong/db/schema/others/declarative_config.lua
```diff
@@ -0,0 +1,437 @@
...
+  "services",
+  "routes",
+  "consumers",
+  "plugins",
+  "upstreams",
+  "targets",
+  "certificates",
+  "snis",
+  "cluster_ca",
+}
```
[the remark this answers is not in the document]
hishamhm: Implemented the suggested change, thank you!

tokingHong:
> * loading

How to use /configure to load config file?What should I put in http body?

PR #4316, opened 2019-02-14
feat(proxy) clear hop-by-hop headers

bungle:
### Summary

Removes the following request and response headers:

- `Proxy` (see: https://httpoxy.org/)
- `Proxy-Connection`
- `Proxy-Authenticate`
- `Proxy-Authorization`
- `TE`
- `Transfer-Encoding`
- `Trailer`
- `Upgrade`
    
Headers specified in the value of the Connection header are cleared as well.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -7,43 +7,54 @@
 --
 -- In the `access_by_lua` phase, it is responsible for retrieving the route being proxied by
 -- a consumer. Then it is responsible for loading the plugins to execute on this request.
-local ck          = require "resty.cookie"
-local meta        = require "kong.meta"
-local utils       = require "kong.tools.utils"
-local Router      = require "kong.router"
-local reports     = require "kong.reports"
-local balancer    = require "kong.runloop.balancer"
-local mesh        = require "kong.runloop.mesh"
-local constants   = require "kong.constants"
-local singletons  = require "kong.singletons"
-local certificate = require "kong.runloop.certificate"
-local concurrency = require "kong.concurrency"
-local ngx_re      = require "ngx.re"
+local ck           = require "resty.cookie"
+local meta         = require "kong.meta"
+local utils        = require "kong.tools.utils"
+local Router       = require "kong.router"
+local reports      = require "kong.reports"
+local balancer     = require "kong.runloop.balancer"
+local mesh         = require "kong.runloop.mesh"
+local constants    = require "kong.constants"
+local singletons   = require "kong.singletons"
+local certificate  = require "kong.runloop.certificate"
+local concurrency  = require "kong.concurrency"
+local ngx_re       = require "ngx.re"
 
 
-local kong        = kong
-local ipairs      = ipairs
-local tostring    = tostring
-local tonumber    = tonumber
-local sub         = string.sub
-local find        = string.find
-local lower       = string.lower
-local fmt         = string.format
-local sort        = table.sort
-local ngx         = ngx
-local log         = ngx.log
-local ngx_now     = ngx.now
-local re_match    = ngx.re.match
-local re_find     = ngx.re.find
-local re_split    = ngx_re.split
-local update_time = ngx.update_time
-local subsystem   = ngx.config.subsystem
-local unpack      = unpack
+local kong         = kong
+local ipairs       = ipairs
+local tostring     = tostring
+local tonumber     = tonumber
+local sub          = string.sub
+local find         = string.find
+local lower        = string.lower
+local fmt          = string.format
+local sort         = table.sort
+local ngx          = ngx
+local arg          = ngx.arg
+local var          = ngx.var
+local log          = ngx.log
+local exit         = ngx.exit
+local header       = ngx.header
+local ngx_now      = ngx.now
+local timer_at     = ngx.timer.at
+local re_match     = ngx.re.match
+local re_find      = ngx.re.find
+local re_split     = ngx_re.split
+local update_time  = ngx.update_time
+local subsystem    = ngx.config.subsystem
+local start_time   = ngx.req.start_time
+local clear_header = ngx.req.clear_header
+local starttls     = ngx.req.starttls
+local unpack       = unpack
 
 
-local ERR         = ngx.ERR
-local WARN        = ngx.WARN
-local DEBUG       = ngx.DEBUG
+local ERR          = ngx.ERR
+local INFO         = ngx.INFO
+local WARN         = ngx.WARN
+local CRIT         = ngx.CRIT
+local DEBUG        = ngx.DEBUG
+local ERROR        = ngx.ERROR
 
 
 local CACHE_ROUTER_OPTS = { ttl = 0 }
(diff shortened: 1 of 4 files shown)


kong/runloop/handler.lua
```diff
@@ -777,6 +778,29 @@ return {
...
+        if header_names then
+          for i=1, #header_names do
+            if header_names[i] ~= "" and lower(header_names[i]) ~= "close" then
+              ngx.req.clear_header(header_names[i])
+            end
+          end
+        end
+      end
+
+      ngx.req.clear_header("Keep-Alive")
```
p0pr0ck5: localize `ngx.req.clear_header`?

bungle: @p0pr0ck5, the `ngx` is localized, so I just skipped localizing the rest. I can change if that is recommended. I am not sure about it, do you know if there are any benefits of doing so (when the global access is already avoided).

p0pr0ck5: Ah heck, i didn't realize the `ngx` is localized. Ignore this :upside_down_face:

james-callahan:
Does nginx not do this for us? What hop by hop headers does nginx manage automatically?

bungle:
@james-callahan it doesn't seem to do it, at least for all the heades. I guess `TE` and `Transfer-Encoding` it might do, but others, I think not. Check this branch and disable modifications in `handler.lua` and see it failing on header tests.

bungle:
There is this question:
https://github.com/Kong/kong/pull/4316/files#diff-1f0f137350308286de76ec59740c8edaR854

That I don't really know how to answer. Kinda like chicken-egg situation.

bungle:
When OpenResty is updated to `Nginx 1.3.10` or above we can add proper `trailers` support I think.

bungle:
Some pointers to this work:

- https://golang.org/src/net/http/httputil/reverseproxy.go
- https://www.mnot.net/blog/2011/07/11/what_proxies_must_do

kong/runloop/handler.lua
```diff
@@ -18,6 +18,7 @@ local constants   = require "kong.constants"
 local singletons  = require "kong.singletons"
 local certificate = require "kong.runloop.certificate"
 local concurrency = require "kong.concurrency"
+local regex       = require "ngx.re"
```
hishamhm: sorry about a "style" request, but could you name this `ngx_re` (or alternative localize the function as `re_split = ngx.re.split`) to match how ngx.re is localized in other parts of the code? I was reading the other part of the patch and for a moment I thought `regex` was something else.

kong/runloop/handler.lua
```diff
@@ -934,6 +941,58 @@ return {
       -- ngx.req.start_time() is kept in seconds with millisecond resolution.
       ctx.KONG_PROXY_LATENCY   = now - ngx.req.start_time() * 1000
       ctx.KONG_PROXIED         = true
+
+      -- clear hop-by-hop request headers:
```
hishamhm: I think these additions should go above the line with `now = get_now()` (i.e., time doing this should be included in the KONG_ACCESS_TIME calculation)

hishamhm:
> There is this question:
> https://github.com/Kong/kong/pull/4316/files#diff-1f0f137350308286de76ec59740c8edaR854
> 
> That I don't really know how to answer. Kinda like chicken-egg situation.

I got confused following the link (not sure if the link above is pointing to the right place or if I just didn't understand) — could you please elaborate on what the open question is?

bungle:
> > There is this question:
> > https://github.com/Kong/kong/pull/4316/files#diff-1f0f137350308286de76ec59740c8edaR854
> > That I don't really know how to answer. Kinda like chicken-egg situation.
> 
> I got confused following the link (not sure if the link above is pointing to the right place or if I just didn't understand) — could you please elaborate on what the open question is?

Yes, I think that was later resolved.

kong/runloop/handler.lua
bungle: Changed that.

kong/runloop/handler.lua
bungle: Changed the order.

kong/runloop/handler.lua
```diff
@@ -574,6 +576,12 @@ return {
   rewrite = {
     before = function(ctx)
       ctx.KONG_REWRITE_START = get_now()
+
+      -- special handling for proxy-authorization and te headers in case
+      -- the plugin(s) want to specify them (store the original)
+      ctx.http_proxy_authorization = ngx.var.http_proxy_authorization
```
dndx: We should be able to cache `ngx.var` at module level variable since they are frequently used. They work through the magic of metamethods and can be safely cached.

kong/runloop/handler.lua
```diff
@@ -925,6 +933,58 @@ return {
...
+            if header_names[i] ~= "" then
+              local header_name = lower(header_names[i])
+              -- some of these are already handled by the proxy module,
+              -- proxy-authorization being an exception that is handled
+              -- below with special semantics.
+              if header_name ~= "close" and
+                 header_name ~= "upgrade" and
+                 header_name ~= "keep-alive" and
+                 header_name ~= "proxy-authorization" then
+                ngx.req.clear_header(header_names[i])
```
dndx: Maybe we should cache the `clear_header` method as well since it is on a potentially hot code path.

p0pr0ck5: Just to note, this was mentioned here as well: https://github.com/Kong/kong/pull/4316#discussion_r257044383

kong/runloop/handler.lua
bungle: Yes, `ngx.var` is not a problem, but `ngx.var.value` is. With `ngx.ctx` it is the otherway around. So I think caching that at the module level doesn't gain much as `ngx` is already cached (like @p0pr0ck5 mentions below).

dndx: Not sure what you mean exactly, but the `ngx.var` table itself does not contain the actual value of the variables, they are retrieved dynamically using the `__index` metamethod. So the table is always empty and caching it will make all access to it having one less indirections, and should be faster than just caching `ngx` alone.

bungle: Local caching of globals is what is usually recommended, but caching local table (in this case localized ngx) functions is what I don't think adds anything (it could add tinyt bits, but very very tiny that goes to error margin of measuremebt). This is explained here: http://luajit.org/ext_ffi_tutorial.html#cache

bungle: @dndx / @p0pr0ck5,

I made another commit to localize all the things in `runloop/handler.lua`:
https://github.com/Kong/kong/pull/4316/commits/4502523ce96b4881f4025f83b06abb14472545bc

Please let me know if we want or not want this. Also @thibaultcha may have something to say about it.

bungle: Crap, our unit test suite is not happy with localizing funcs:
https://travis-ci.org/Kong/kong/builds/521608375

bungle: Oh, actually a typo at localizing one function. Fixed now. What do you think? With or without?

PR #4319, opened 2019-02-15
fix(db) fix target cleanup and cascade-delete for Targets

### LINKED ISSUE #4317: Cannot delete upstreams on Kong 1.0.2
### Summary

Cannot delete upstreams on Kong 1.0.2

### Steps To Reproduce

1. Create one upstream
2. Create some targets belongs to the  upstream
3. Delete all the targets
4. Delete the upstream will raise error "an existing 'targets' entity references this 'upstreams' entity"

### Additional Details & Logs

- Kong version 1.0.2
- Operating system Debian 4.9.130-2 (2018-10-27) x86_64
- Database cassandra

hishamhm:
This fixes two problems:

1) Target cleanup was happening in the wrong order. This was not a problem for
the target being inserted (because cleanup happened before insertion), but it
would cause the other targets of the list to be corrupted in case they had a
target history themselves.

   Since cleanup happened before insertion (meaning before a weight=0 entry was
inserted upon deletion), this meant that deleting targets never truly emptied
the table -- this meant that Cassandra could not delete upstreams, because the
cascade-delete annotation was missing from the entity schema. Merely adding
`on_delete = "cascade"` to `targets.upstream` would fix the problem in #4317
illustrated by the regression test added to
spec/02-integration/04-admin_api/07-upstreams_routes_spec.lua but would not
fix the other problem, illustrated by the test added to
spec/02-integration/04-admin_api/08-targets_routes_spec.lua.

2) We also add `on_cascade = "delete"` to the schema, which fixes the inconsistency in behavior between Postgres and Cassandra, as the former included cascade-delete in its migration already.

Fixes #4317.

### DIFF
--- a/kong/db/dao/targets.lua
+++ b/kong/db/dao/targets.lua
@@ -27,8 +27,8 @@ end
 
 
 local function clean_history(self, upstream_pk)
-  -- when to cleanup: invalid-entries > (valid-ones * cleanup_factor)
-  local cleanup_factor = 10
+  -- when to cleanup: when less than 10 percent of the entries are valid
+  local cleanup_factor = 0.1
 
   --cleaning up history, check if it's necessary...
   local targets, err, err_t = self:select_by_upstream_raw(upstream_pk, 1000)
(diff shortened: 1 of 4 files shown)


kong/db/dao/targets.lua
```diff
@@ -111,7 +115,7 @@ function _TARGETS:delete(pk)
     return nil, err, err_t
   end
 
-  return self:insert({
+  self:insert({
```
bungle: Should we handle possible errors in this case?

hishamhm:
totally! that was an editing mistake. good catch, thanks!

kong/db/dao/targets.lua
hishamhm: fixed!

PR #4333, opened 2019-02-20
feat(core) add support for stream ipv6 routes

bungle:
### Summary

The validator used ipv4 only `resty.iputils` which caused errors in validating ipv4 addresses in stream routes with destinations and sources. The router itself used `resty.mediador.proxy` which does
have ipv6 support.

The `ip-restriction` plugin is still left to use `resty.iputils` partly because it is very efficient implementation (faster than mediador). We may need in future look for nice optimized way to verify ips and cidr blocks, and match cidr blocks. We also have some tools in `kong.tools.utils` which are a bit restrictive in their implementation.

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -7,8 +7,11 @@ local iputils = require "resty.iputils"
 local Schema = require("kong.db.schema")
 local socket_url = require("socket.url")
 local constants = require "kong.constants"
+local px = require "resty.mediador.proxy"
 
 
+local pairs = pairs
+local pcall = pcall
 local match = string.match
 local gsub = string.gsub
 local null = ngx.null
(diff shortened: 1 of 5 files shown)


kong/db/schema/typedefs.lua
```diff
@@ -174,11 +188,15 @@ typedefs.ip = Schema.define {
...
 
 
-typedefs.cidr = Schema.define {
+typedefs.ip_or_cidr = Schema.define {
   type = "string",
-  custom_validator = validate_cidr,
+  custom_validator = validate_ip_or_cidr,
 }
 
+typedefs.cidr_v4 = Schema.define {
```
hishamhm: could you add 
```
-- deprecated alias
typedefs.cidr = typedefs.cidr_v4
```
to support it as a deprecated alias? just in case a third-party plugin uses the typedef with the old name.

bungle: Added.

PR #4339, opened 2019-02-22
feat(db) implement foreign entity unique / endpoint key field resolver

bungle:
### Summary

This work was originally done by @hbagdi (thank you):
https://github.com/hbagdi/kong/commit/d627649ee3aad55d7a0c94a09e627f356ea3994e

This commit makes it possible to use names (`unique fields` or `endpoint key`) in `insert`/`update`/`upsert`/`update_by_*`/`upsert_by_*`
kong db calls and admin api calls, e.g.:

```
http :8001/routes paths=/ service.name=test -f
```

(previously you needed to use `service.id=<uuid>`)

### DIFF
--- a/kong/api/endpoints.lua
+++ b/kong/api/endpoints.lua
@@ -20,18 +20,19 @@ local split        = utils.split
 
 -- error codes http status codes
 local ERRORS_HTTP_CODES = {
-  [Errors.codes.INVALID_PRIMARY_KEY]   = 400,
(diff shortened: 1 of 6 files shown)


bungle:
Btw. it is really easy to add foreign entity creation too, but that is part of another PR. We may implement creation only with some entities, e.g. by creating route, we can create service, but maybe not allow creating routes or consumers when creating a plugin instance.

PR #4342, opened 2019-02-22
hotfix(dbless) fix initial load of declarative config

hishamhm:
This fixes the initial loading of the declarative config file based on the `declarative_config` configuration variable.

Includes a regression test (triggering the issue running that test on `next` may be flaky (though it did trigger consistently On My Machine™️, but running the test on the fixed version should be stable — best I could get for an intermittent test).

### DIFF
--- a/spec/02-integration/02-cmd/02-start_stop_spec.lua
+++ b/spec/02-integration/02-cmd/02-start_stop_spec.lua
@@ -220,6 +220,56 @@ describe("kong start/stop #" .. strategy, function()
     end)
   end)
 
+  if strategy == "off" then
+    describe("declarative config start", function()
+      it("starts with a valid declarative config file", function()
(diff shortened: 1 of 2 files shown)


spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -220,6 +220,56 @@ describe("kong start/stop #" .. strategy, function()
     end)
   end)
 
+  if strategy == "off" then
+    describe("declarative config start", function()
+      it("#only starts with a valid declarative config file", function()
```
thibaultcha: We should remove the `#only` tag

hishamhm: fixed!

PR #4357, opened 2019-03-04
chore(functional tests) run kong-build-tools functional tests during Kong CI

hutchic:
Runs the python/tavern functional tests against Kong packaged from the current source code and running on k8s via helm


hutchic:
Capture

This can be sped up but until it is perhaps we want to restrict when travis-ci runs the tests?

PR #4358, opened 2019-03-05
feat(pdk) move 'get_subsystem()' to a new 'kong.nginx' module 

thibaultcha:


kong/pdk/init.lua
```diff
@@ -211,7 +211,7 @@ assert(package.loaded["resty.core"])
 
 local MAJOR_VERSIONS = {
   [1] = {
-    version = "1.0.0",
+    version = "1.1.0",
```
hishamhm: A comment: I initially thought we would add a `"1.0.0"` entry as well to MAJOR_VERSIONS with the same list of modules minus `"nginx"`, plus a version check in `kong.client.get_protocol` so that it is only available in `kong.pdk.client` when `version >= "1.1.0"`, but we currently don't offer a way for plugins to specify the desired PDK version anyway, so this would be of little use at this point.

PR #4102, opened 2018-12-16
feat(core) router rebuild timer (revisited)

bungle:
### Summary

This PR refactors Kong's router rebuilding. Previously our router rebuilding happened synchronously only on request time which is good in sense that it maintains rather good consistency between workers. The problem is that it makes workers to wait for a new router (and queue requests), and it really shows up in P99 etc (and I don't think there is a way to make rebuilding really fast on request time whatever we do — except maybe granular updates). There is also a smaller issue when having large numbers of routes that db state changes when router building is paging through the routes and meanwhile the changes happen to routes or services.

This PR tries to fix issues, and instead of trying to be one size fits all solution (like the previous implementation) there is now a configuration value:

```
router_timeout = <milliseconds>
```

The default value for this parameter is `5000`. `0` means that whenever request acquires a router rebuild mutex, it starts a rebuild in a background timer (and serves the request with the old one). If timer creation fails, it runs rebuilding synchronously (before routing request). Usually this should not happen though as the rebuilding now also rebuilds when it gets worker events for `router:version` changes. It also now broadcasts the rebuild across all the workers and the cluster (pre-warming the new router).

If you set the value to negative e.g. `-1` the router rebuilding fully relies on background worker events to rebuild a router (there is no request time penalty or mutexes).

The previous implementation is pretty much the same as setting the value to (this is also a default value so that expectations won't change on this — it is also noted elsewhere that we should not use pg or cassandra timeout here for this as we currently do):
```
router_timeout = 5000
```

So there you go:
1. most request performance, set this to `-1` (if timer creation fails, new router will not be build, but you can trigger a new try with modifying routes or services)
2. good balance, set this to `0` (we also try to rebuild with a timer on request, and fallback to synchronous building)
3. bit more consistenty set this to `1` (1 ms wait for mutex before doing synchronous rebuilding on request)
4. more queuing and less db hammering, set this to `1000` or perhaps `5000` (the old and new default)

This PR also implements caching of the `service` entities, needed for rebuilding the router. 

This is continuation of the previous failed attempt: #4087 to fix #4055.

There are some other work recently merged or discussed:
#4101 (does prefetching instead of caching as this pr)
#4098 (mutex fallthrough that also this pr does when config value is > 0)
#4084 (mutex is always released)

### Issues resolved

Fix #4055

PR #4364, opened 2019-03-05
fix(router) ignore failures on unexisting services when building router

ealogar:
### Summary

The router is being built querying all routes and iterating over each of them by searching the related service.
This is a potentially inconsistency as a service may be deleted during the process, the router won't be built and a 500 error will be returned.
Instead this error is ignored and the router is being built with the rest of routes. After all if a route is not found it will fail with a 404 (that it normal) or the request will be progressed as the service deletion was different.
The request won't be lost in most of the cases.

Moreover: when this process happend (without this fix) the  dao returned nil in err field and this was causing the log to fail:
```
2019/03/05 03:46:38 [error] 35#0: *40775 [kong] handler.lua:689 no router to route request (reason: /usr/local/share/lua/5.1/kong/runloop/handler.lua:89: attempt to concatenate local 'err' (a nil value)), client: 100.98.128.0, server: kong, request: "GET /my-route HTTP/1.1", host: "example.com"
```
What it's also a minor bug that is solved as well.

### Full changelog

* ignore failures when building routes and a service is deleted

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -69,8 +69,29 @@ do
   local router
   local router_version
 
+  local function get_route_for_service(service, route)
+    local r
+    local stype = protocol_subsystem[service.protocol]
+    if subsystem == stype then
+      r = {
+        route = route,
+        service = service,
+      }
+
+      if stype == "http" and route.hosts then
+        -- TODO: headers should probably be moved to route
+        r.headers = {
+          host = route.hosts,
+        }
+      end
+    end
+    return r
+  end
+
+
   build_router = function(db, version)
     local routes, i = {}, 0
+    local r
 
     for route, err in db.routes:each(1000) do
       if err then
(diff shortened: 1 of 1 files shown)


bungle:
This is a good reminder that we need to move these forward:
https://github.com/Kong/kong/pull/4202
https://github.com/Kong/kong/pull/4102

They only fix similar issues and then some. Our current target for "merging" them will be `1.2.0`. Let's keep this PR open as well as it is related, and perhaps we can include some of work done here to that upcoming work as well.

bungle:
I marked it `for discussion` as it closely relates to other work already done in this area.

thibaultcha:
We could pull this in 1.1.0 as a bugfix for the time being, taking a look at it.

bungle:
Also, this touches a little bit the same place as well:
https://github.com/Kong/kong/pull/4286

jeremyjpj0916:
Looks like the problem I have seen often has reared its ugly head again 😄 . Luckily @bungle has a great solution for it imo already tested and vetted around router/plugin map rebuild running in prod by us(me/Optum).

bungle:
I think we may get some of the work to `1.1`. Things like:
1. use same db connection when rebuilding router
2. either cache services or preload them (caching feels more correct as the services may change between requests)
3. add a timer for rebuild but still leave the synchronous rebuilding on proxy time for worker consistency (make it configurable on `1.2` so that you don't affect proxy performance of p99)
4. do the atomic rebuilds of router and plugins map (if things change during build, e.g. because of paging and separate service selects, stop, and start again, as the paging cannot be trusted anymore).

thibaultcha:
@bungle Will you propose a simplified version of your patches as well so we can compare approaches?

ealogar:
I include a commit as the code was failing when a service is not found, I can reproduce locally deleting services. A Lua concat error was raised instead of building the router with the valid routes.
I will rebase it if you finally approve this.

ealogar:
@thibaultcha @bungle I see that this problem may appear in release 1.1.0 released today, are you planning in adding this or a different approach to solve it ?

PR #4369, opened 2019-03-07
feat(proxy) support wildcard certificates

ontherunvaro:
### Summary

Only if there isn't a 1-to-1 match between a requested SNI and the
available certificates, substitute the first member of the SNI for a
wilcard char and try the search again.

### Full changelog

* Add wildcard SNI support to runloop/certificate.lua

### DIFF
--- a/kong/runloop/certificate.lua
+++ b/kong/runloop/certificate.lua
@@ -26,9 +26,18 @@ local function fetch_certificate(sni_name)
   end
 
   if not row then
-    log(DEBUG, "no SNI registered for client-provided name: '",
+    -- no direct match:  find if there's a wildcard available
+    local _, index = string.find(sni_name, '%.')
+    local wildcard_domain = '*' .. string.sub(sni_name, index);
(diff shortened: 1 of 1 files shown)


thibaultcha:
Hi,

Thank you for proposing the patch. Unfortunately, as I am working on support for wildcard SNI myself on my spare time, I can tell that this patch won't be a complete solution to the problem, sorry. From this patch as it is, we are missing:

- wildcard suffix
- cache invalidation
- tests

If you are willing to wait a few more days, I have a branch in my local development environment that has all of the above. At this point, I am working on a lua-resty-mlcache feature that will make the wildcard SNI fetching much more efficient that what I currently have. I expect to push my branch in the next week or couple of weeks at most.

thibaultcha:
In the meantime, we can continue a discussion here, so I won't be closing this right away, but only labelling it as a no-go for the time being.

ontherunvaro:
Alright, I only pushed this because it seems like a common use case that wasn't being met, and I saw no indication that it was planned. Thank you for your work!

PR #4378, opened 2019-03-09
fix(*) cassandra contact points initialization

bungle:
### Summary

C* doesn't initialize contact points correctly when the contact points need to be resolved with DNS. DNS on the other hand uses Nginx cosockets that cannot be used with init phase. Nginx cosockets are not fully compatible with lua sockets. This PR is a quick fix for the issue.

This is based on `next` to prevent some merge conflicts. I can rebase on master if that is needed.

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -14,7 +14,77 @@ function CassandraConnector.new(kong_config)
   do
     -- Resolve contact points before instantiating cluster, since the
     -- driver does not support hostnames in the contact points list.
+
+    -- TODO: this is an ugly hack to force lua sockets on a third party library
+
+    local tcp_old = ngx.socket.tcp
+    local udp_old = ngx.socket.udp
+
+    local dns_no_sync_old = kong_config.dns_no_sync
+
+    package.loaded["socket"] = nil
+    package.loaded["kong.tools.dns"] = nil
+    package.loaded["resty.dns.client"] = nil
+    package.loaded["resty.dns.resolver"] = nil
+
+    ngx.socket.tcp = function(...)
+      local tcp = require("socket").tcp(...)
+      return setmetatable({}, {
+        __newindex = function(_, k, v)
+          tcp[k] = v
+        end,
+        __index = function(_, k)
+          if type(tcp[k]) == "function" then
+            return function(_, ...)
+              if k == "send" then
+                local value = select(1, ...)
+                if type(value) == "table" then
+                  return tcp.send(tcp, table.concat(value))
+                end
+
+                return tcp.send(tcp, ...)
+              end
+
+              return tcp[k](tcp, ...)
+            end
+          end
+
+          return tcp[k]
+        end
+      })
+    end
+
+    ngx.socket.udp = function(...)
+      local udp = require("socket").udp(...)
+      return setmetatable({}, {
+        __newindex = function(_, k, v)
+          udp[k] = v
+        end,
+        __index = function(_, k)
+          if type(udp[k]) == "function" then
+            return function(_, ...)
+              if k == "send" then
+                local value = select(1, ...)
+                if type(value) == "table" then
+                  return udp.send(udp, table.concat(value))
+                end
+
+                return udp.send(udp, ...)
+              end
+
+              return udp[k](udp, ...)
+            end
+          end
+
+          return udp[k]
+        end
+      })
+    end
+
     local dns_tools = require "kong.tools.dns"
+
+    kong_config.dns_no_sync = true
+
     local dns = dns_tools(kong_config)
 
     for i, cp in ipairs(kong_config.cassandra_contact_points) do
@@ -28,6 +98,16 @@ function CassandraConnector.new(kong_config)
         resolved_contact_points[i] = ip
       end
     end
+
+    kong_config.dns_no_sync = dns_no_sync_old
+
+    package.loaded["resty.dns.resolver"] = nil
+    package.loaded["resty.dns.client"] = nil
+    package.loaded["kong.tools.dns"] = nil
+    package.loaded["socket"] = nil
+
+    ngx.socket.udp = udp_old
+    ngx.socket.tcp = tcp_old
   end
 
   if #resolved_contact_points == 0 then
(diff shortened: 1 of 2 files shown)


Tieske:
Wondering whether the cassandra code is the right place to fix this. Wouldn't the dns client be the right place to fix it?

Also wondering, doesn't this same issue exist for Postgres? Why do we connect to Cassandra in this phase, but not to Postgres?

hutchic:
@Tieske Based on my tests of rc1 postgres does not suffer the same issue

bungle:
@tieske the question is that do we want to have this hack available anywhere or just here. I guess you could perhaps inject such hack at some point to your dns client if we need it work anywhere. Postgres does not run anything on init, not sure if it should test connection there(?).

Because we are about to release the rc2 soon, I wanted to isolate this hack so that it has a smaller surface base.

thibaultcha:
> Also wondering, doesn't this same issue exist for Postgres? Why do we connect to Cassandra in this phase, but not to Postgres?

Both Postgres and Cassandra strategies will connect to their respective instances during `init_by_lua`. The former to retrieve the Postgres version, the latter to refresh the cluster topology and retrieve the peers' versions as well. However, both have mechanisms to fallback to LuaSocket for their queries in this phase. Both Postgres instances and Cassandra peers could be specified as hostname thanks to our `globalpatches.lua` or the LuaSocket fallback in the case of `init_by_lua` (this is, as far as I remember, not tested and involuntarily supported). However, since we updated the C* strategy to use our DNS client in `init_by_lua`, but the tests rely on `/etc/hosts` values (which do not yield), this slipped away.

thibaultcha:
(And the tests run in `ngx.timer` anyway because of rbusted (resty-cli) which supports yielding. Hence why the integration tests starting Kong itself in this PR).

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -57,6 +57,20 @@ describe("kong start/stop #" .. strategy, function()
...
+
+  if strategy == "cassandra" then
+    it("start resolves cassandra contact points", function()
+      assert(helpers.kong_exec("start --prefix " .. helpers.test_conf.prefix, {
+        database = strategy,
+        cassandra_contact_points = "localhost",
+        cassandra_keyspace = helpers.test_conf.cassandra_keyspace,
+      }))
+      assert(helpers.kong_exec("stop", {
+        prefix = helpers.test_conf.prefix
```
thibaultcha: It shouldn't be necessary to specify the prefix here (`helpers.test_conf.prefix` already is the default for tests), unless we specify it in both `start` and `stop` (so as to reduce confusion, for explicitness' sake).

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -342,6 +356,20 @@ describe("kong start/stop #" .. strategy, function()
...
+      it("errors when cassandra contact points cannot be resolved", function()
+        local ok, stderr = helpers.kong_exec("start --prefix " .. helpers.test_conf.prefix, {
+          database = strategy,
+          cassandra_contact_points = "invalid.inexistent.host",
+          cassandra_keyspace = helpers.test_conf.cassandra_keyspace,
+        })
+
+        assert.False(ok)
+        assert.matches("could not resolve any of the provided Cassandra contact points " ..
+                       "(cassandra_contact_points = 'invalid.inexistent.host')", stderr, nil, true)
```
thibaultcha: A `finally` hook with: 

```lua
finally(function()
  helpers.kill_all()
  pcall(helpers.dir.rmtree)
end)
```

would be welcome and make this test more robust (what if the underlying behavior disappears, the test fails, and now we leave this test instance running? It could impact subsequent test cases).

kong/db/strategies/cassandra/connector.lua
```diff
@@ -14,19 +14,105 @@ function CassandraConnector.new(kong_config)
...
+      package.loaded["resty.dns.client"] = nil
+      package.loaded["kong.tools.dns"] = nil
+      package.loaded["socket"] = nil
+
+      ngx.socket.udp = udp_old
+      ngx.socket.tcp = tcp_old
+    end)
+
+    if not pok then
+      return nil, perr
```
thibaultcha: We should return a more descriptive error here, maybe something like:

```lua
return nil, "error while patching resty.dns.resolver and resolving contact points: " .. perr
```

I actually even wonder what the rationale is behind protecting this block? It hides the stacktraces and a returned error produces the same result as a runtime one: Kong cannot start because it has no DAO. What is the rationale/how does it make the booting sequence more resilient?

spec/02-integration/02-cmd/02-start_stop_spec.lua
bungle: removed.

spec/02-integration/02-cmd/02-start_stop_spec.lua
bungle: added finally.

kong/db/strategies/cassandra/connector.lua
bungle: removed pcall.

kong/db/strategies/cassandra/connector.lua
```diff
@@ -14,7 +14,77 @@ function CassandraConnector.new(kong_config)
   do
     -- Resolve contact points before instantiating cluster, since the
     -- driver does not support hostnames in the contact points list.
+
+    -- TODO: this is an ugly hack to force lua sockets on a third party library
```
thibaultcha: May be worth being more explicit here:

- we force sync option for resty.dns.client to not spawn a timer
- we force LuaSocket for resty.dns.resolver to not spawn a cosocket in init_by_lua

Can add at merge time, in fact let me take care of it, no worries :+1:

kong/db/strategies/cassandra/connector.lua
```diff
@@ -14,7 +14,77 @@ function CassandraConnector.new(kong_config)
...
+              return tcp[k](tcp, ...)
+            end
+          end
+
+          return tcp[k]
+        end
+      })
+    end
+
+    ngx.socket.udp = function(...)
```
thibaultcha: Note for posterity:

ideally we should add support for `ngx.socket.udp` to lua-resty-socket and avoid duplicating the fallback logic there and here (lua-resty-socket has its own test suite and benefits other projects).

bungle: Oh yes, looks like it already has all thing covered. Lets add dependency to it with 1.1+.

PR #4380, opened 2019-03-11
hotfix(dbless) make tags compatible with database=off

kikito:


kong/db/declarative/init.lua
```diff
@@ -304,13 +322,27 @@ function declarative.load_into_cache(entities)
         end
       end
 
-      for fname, ref in pairs(foreign_fields) do
-        if item[fname] then
-          local fschema = kong.db[ref].schema
+      if id then
```
hishamhm: `if id` is not necessary, every iteration in this loop will have a truthy `id` value

PR #4381, opened 2019-03-11
feat(dao) re-introduce immutable field validator to DAO

Jlawlzz:
### Summary
This PR re-introduces the `immutable` field validator to the DAO, allowing Kong to prevent certain fields from being updated. 

### Full changelog
* **schema/init.lua** Introduced `validate_immutable_fields` schema method
* **schema/init.lua** Expanded checks in `process_auto_fields` to include searching each field for `immutable` attribute
* **dao/init.lua** Integrated new functionality into `check_update`
* **spec/01-unit/01-db/04-dao.lua** Added Unit Tests

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -83,6 +83,7 @@ local validation_errors = {
   REQUIRED                  = "required field missing",
   NO_FOREIGN_DEFAULT        = "will not generate a default value for a foreign field",
   UNKNOWN                   = "unknown field",
+  IMMUTABLE                 = "immutable field cannot be updated",
   -- entity checks
   REQUIRED_FOR_ENTITY_CHECK = "field required for entity check",
   ENTITY_CHECK              = "failed entity check: %s(%s)",
@@ -1432,6 +1433,7 @@ function Schema:process_auto_fields(data, context, nulls)
   local now_ms = ngx_now()
   local read_before_write = false
   local cache_key_modified = false
+  local check_immutable_fields = false
 
   data = tablex.deepcopy(data)
 
@@ -1520,6 +1522,15 @@ function Schema:process_auto_fields(data, context, nulls)
        and data[key] ~= nil and data[key] ~= null then
       read_before_write = true
     end
+
+    if context == 'update' and field.immutable then
+      -- if entity schema contains immutable fields,
+      -- we need to preform a read-before-write and
+      -- check the existing record to insure update
+      -- is valid.
+      read_before_write = true
+      check_immutable_fields = true
+    end
   end
 
   --[[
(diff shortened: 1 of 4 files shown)


kong/db/schema/init.lua
```diff
@@ -1599,6 +1610,33 @@ function Schema:validate(input, full_check)
...
+-- against entity values to detirmine whether input is valid.
+-- @param input The input table.
+-- @param entity The entity update will be performed on.
+-- @return True on success.
+-- On failure, it returns nil and a table containing all errors by field name.
+-- In all cases, the input table is untouched.
+function Schema:validate_immutable_fields(input, entity)
+  local errors = {}
+
+  input = tablex.deepcopy(input)
```
thibaultcha: Any reason for deep-copying this table (a fairly expensive operation) since we are not modifying its values?

kong/db/schema/init.lua
```diff
@@ -1599,6 +1610,33 @@ function Schema:validate(input, full_check)
...
+
+  for key, field in pairs(input) do
```
thibaultcha: Have you considered using `each_field()` to iterate over this input table, as the more native way to iterate over fields of a schema-based entity?

```lua
for name, field in self:each_field() do
  local value = input[name]
  if field.immutable and ... then

  end
end
```

kong/db/schema/init.lua
```diff
@@ -1599,6 +1610,33 @@ function Schema:validate(input, full_check)
...
+    if self.fields[key].immutable and entity[key] ~= nil and input[key] ~= entity[key] then
```
thibaultcha: Understand this PR is a draft implementation, but we'll need to assess how aggregate types integrate with `immutable` (not at all/they do but then we need more logic before asserting a table is identical to one already in DB, as comparing memory address is not enough in this context).

kong/db/schema/init.lua
Jlawlzz: 👍 good call

kong/db/schema/init.lua
Jlawlzz: The concept of this type of comparison was not established in the previous version of Kong but definitely seems necessary/needed in this re-implementation.  I see no reason why immutability should not apply to aggregate types as well.  Will look into a helper function which will assess likeness.

kong/db/schema/init.lua
```diff
@@ -81,6 +81,7 @@ local validation_errors = {
   REQUIRED                  = "required field missing",
   NO_FOREIGN_DEFAULT        = "will not generate a default value for a foreign field",
   UNKNOWN                   = "unknown field",
+  IMMUTABLE                 = "field cannot be updated",
```
nijikokun: We should state that immutability is the reason why the field cannot be updated to be consistent with other error messages self identifying.

Jlawlzz: 👍

Jlawlzz:
@james-callahan: The purpose of immutability in this case is to preserve important relationships that could otherwise be tampered with, not prevent a database item from having a particular field (so deleting and replacing a record isn't problematic or what this is trying to solve).

PR #4383, opened 2019-03-12
fix(db) properly serialize maps in C* strategy

thibaultcha:
A map does not have `elements`, but `keys` and `values`.

This unfortunately does not come with a test, as no core entity has a
map field as of today.


thibaultcha:
And thanks @gszr for reporting this one :)

PR #4387, opened 2019-03-13
fix(db) is part of primary key was incorrectly checked on foreign fields

bungle:
### Summary

A fix to Postgres strategy that corrects the check on foreign fields whether or not the foreign field is part of the primary key as well. This is not needed for core entities (so perhaps not that urgent), but custom entities may need it.


kikito:
Thanks for separating the style changes from the functionality changes. It made the review much easier for me :)

PR #4390, opened 2019-03-13
Release: 1.1.0rc2

hishamhm:
# What's new since Kong 1.1.0rc1:

### Core
* **Transparent proxying** - the `service` attribute on Routes is now optional; a Route without an assigned Service will proxy transparently (#4286)

### Core
* Resolve hostnames in Cassandra contact points (#4378)
* Fix health checks for Targets that need two-level DNS resolution (e.g. SRV → A → IP) (#4386)
* Fix serialization of map types in the Cassandra backend (#4368)

### Admin API
* Fix Admin API inferencing of map types using form-encoded (#4368)

### PDK
* PDK version identifier was bumped to 1.1.0
* `kong.client.get_subsystem` was moved to `kong.nginx.get_subsystem` (#4358)

### DB-less / Declarative config
* fix the behavior of the `/tags` endpoint and `?tags=` query string from the Admin API when using `database=off` (#4380)
* fix initial load of the declarative config file (#4342)
* reloading the declarative config via `/config` refreshes the Upstream balancers (#4372)

## Dependencies

* Bump OpenSSL to 1.1.0b
* Bump lua-resty-dns-client to 3.0.2

### DIFF
--- a/kong-1.1.0rc2-0.rockspec
+++ b/kong-1.1.0rc2-0.rockspec
@@ -1,9 +1,9 @@
 package = "kong"
-version = "1.1.0rc1-0"
+version = "1.1.0rc2-0"
 supported_platforms = {"linux", "macosx"}
 source = {
   url = "git://github.com/Kong/kong",
-  tag = "1.1.0rc1"
+  tag = "1.1.0rc2"
 }
 description = {
   summary = "Kong is a scalable and customizable API Management Layer built on top of Nginx.",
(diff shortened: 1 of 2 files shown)


Tieske:
I think it's worth mentioning the fix for the ["address is nil"](https://github.com/Kong/lua-resty-dns-client/issues/49) error in the change log as well. We have had many reports over time for this bug

PR #4393, opened 2019-03-15
Enable least-connections loadbalancer

Tieske:
Enables the least-connections algorithm in the loadbalancer


kong/db/schema/entities/upstreams.lua
```diff
@@ -212,6 +213,12 @@ local r =  {
 
     -- different headers
     { distinct = { "hash_on_header", "hash_fallback_header" }, },
+
+    -- if algorithm is "least" then hash_on must be "none"
+    { conditional = {
+      if_field = "algorithm", if_match = { match = "^least$" },
+      then_field = "hash_on", then_match = { one_of = { "none" }, },
```
kikito: I suggest using `eq = "none"` instead of `one_of = { "none" }` here

Tieske: done

kong/db/schema/entities/upstreams.lua
```diff
@@ -154,6 +154,7 @@ local r =  {
     { id = typedefs.uuid, },
     { created_at = typedefs.auto_timestamp_s },
     { name = { type = "string", required = true, unique = true, custom_validator = validate_name }, },
+    { algorithm = { type = "string", default = "consistent", one_of = { "consistent", "least" }, }, },
```
bungle: The docs [1] say:

> By default a ring-balancer will use a weighted-round-robin scheme. The alternative would be to use the hash-based algorithm. The input for the hash can be either none, consumer, ip, header, or cookie. When set to none the weighted-round-robin scheme will be used, and hashing will be disabled.

Should we consider naming this parameter to `lb_ algorithm` or `balancing_algorithm` or `lb_strategy` or something like that, with values such as: `weighted-round-robin`, `least-connections` (no hard feelings if s suffix is removed), `consistent-hashing`?

[1] https://docs.konghq.com/1.0.x/loadbalancing/#balancing-algorithms

kong/db/schema/entities/upstreams.lua
```diff
@@ -212,6 +213,12 @@ local r =  {
 
     -- different headers
     { distinct = { "hash_on_header", "hash_fallback_header" }, },
+
+    -- if algorithm is "least" then hash_on must be "none"
```
bungle: Why it must be none? Why cannot `least` just ignore it?

hishamhm: I understand @Tieske's motivation for adding this constraint, but I think @bungle has a point there because there are already other fields that only make sense for certain configurations (e.g. `hash_on_cookie` only makes sense to be set if `hash_on == "cookie"`; we check that `hash_on_cookie` is required when hash mode is "cookie", but we don't check that it is _absent_ when hash mode is _not_ "cookie"). 

In the end, what @Tieske is really trying to represent with all those conditionals is a [tagged union](https://en.wikipedia.org/wiki/Tagged_union#1970s_&_1980s), but our schema library has not reached the state-of-the-art of the 1970s yet. ;)

kong/db/schema/entities/upstreams.lua
hishamhm: > weighted-round-robin (if that is still supported),

Currently, weighted-round-robin is implicitly set by using `hash_on` == `"none"` (and is the default). Now that we'll have three algorithms, it would be tempting to remove `"none"` from `hash_on` and add `weighted-round-robin` as the default option — but that would be a breaking change, so a no-no for 1.x. (Definitely something to consider if we redesign the Upstream entity in 2.x — perhaps even renaming it to LoadBalancer as we once discussed?)

As for my 2 cents on the new names, I like @bungle's suggestion of more descriptive values, and would prefer a name using "algorithm" in it rather than "strategy".

bungle: This can be made bC:

1. check algorithm
2. check this `none` value, and deprecate `none` option

hishamhm: It's tricky. The current default load balancing algorithm is indeed weighted-round-robin (due to `hash_on` being `none` by default).

If we add `algorithm` with a `weighted-round-robin` option as a default, then to set consistent-hashing one would need to set two attributes (`algorithm` and `hash_on`) instead of one, and that would break backwards compatibility.

To maintain the current behavior (i.e. so that `http :8001/upstreams name=foo hash_on=header hash_on_header=my-header` is sufficient to set up a consistent-hashing Upstream), setting `hash_on` to any value other than `none` would have to magically force `algorithm` to `consistent-hashing`. It would be doable using the new "shorthands" feature...

```lua
  shorthands = {
    { hash_on = function(value)
        return { algorithm = value == "none"
                             and "weighted-round-robin"
                             or "consistent-hashing" }
      end
    },
    -- optional:
    { algorithm = function(value)
        if value == "weighted-round-robin" then
          return { hash_on = "none" }
        end
      end
    },
  }
```

(Shorthand transformations run before the usual schema validation (and only when the value is actually given by the user), so all the rest usual schema validation rules still apply — we need to fix shorthands to _not_ run when context is `"select"`, though)

I'm not a super fan of the extra magic, but if we can pull this off, then having the three options in the algorithm attribute could make it easier to understand and document (and do away with the sort of confusion that just arised about weighted-round-robin).

bungle: I see, I do not feel too strongly about it. But having value of `consistent` in algorithm, but actual algorithm being `weighted-round-robin` feels some kind of strange magic too. What about making bc-incompatible change and writing a migration?

Tieske: problem is we have 2 balancer-types; ring-balancer, and least-connections. The latter one implements (only) the least-connections algorithm. Whereas the former one implements both weighted-round-robin, as well as consistent-hashing algorithms.

This cannot be cleanly separated, as the weighted-round-robin is the fallback for consistent-hashing (either when no hash is provided, as well as when an upstream is marked unhealthy).

kong/db/schema/entities/upstreams.lua
Tieske: I don't mind leaving it out. Only from a usability prespective, when a user selects "least" and provides a "hash_on" value, he might expect a consistent-hashing approach to be used. An error would make it explicit that it isn't supported.

bungle: but on the otherhand makes switching algorithms back and forth cumbersome (if using other than none).

bungle: but on the otherhand makes switching algorithms back and forth cumbersome (if using other than none).

kong/db/schema/entities/upstreams.lua
bungle: ok, so weighted round robin is actually a mode for consistent. thank you for explaining.

Tieske:
closing this in favor of #4528 

git history locally and github got out of sync, hence I created a new branch

PR #4394, opened 2019-03-15
docs(upgrade) add clarifications on Blue/Green

hishamhm:
As discussed with @UkiahSmith based on customer feedback, we thought some further clarifications on the migrations process from 0.14 to 1.0 would be helpful.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -256,19 +256,32 @@ See the section above on Plugins for information on plugin compatibility.
 
 Kong 1.0 introduces a new, improved migrations framework.
 It supports a no-downtime, Blue/Green migration model for upgrading
-from 0.14.x. The full migration is now split into two steps,
-which are performed via commands `kong migrations up` and
-`kong migrations finish`.
+from 0.14.x. This means that while the migration is ongoing, you will
+have two Kong clusters running, sharing the same database. The "Blue" cluster
+is your existing cluster running 0.14.x, the "Green" cluster is the new one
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -256,19 +256,32 @@ See the section above on Plugins for information on plugin compatibility.
...
 It supports a no-downtime, Blue/Green migration model for upgrading
-from 0.14.x. The full migration is now split into two steps,
-which are performed via commands `kong migrations up` and
-`kong migrations finish`.
+from 0.14.x. This means that while the migration is ongoing, you will
+have two Kong clusters running, sharing the same database. The "Blue"
+cluster is running 0.14.x, the "Green" cluster is running Kong 1.0.
+
+The migrations are designed so that there is no need to fully copy
+the data, but this also means that it designed in such a way so that
```
kikito: Faulty English on "it designed". 

Since the paragraph starts with "The migrations are designed " I think the "it" could be replaced by a "they're".

UPGRADE.md
```diff
@@ -256,19 +256,32 @@ See the section above on Plugins for information on plugin compatibility.
...
+the new version of Kong is able to use the data as it is migrated,
+and to do it in a way so that the old Kong cluster keeps working until
```
kikito: Called "Blue" before, but now called "old" here.

hishamhm: I addressed the concern by making a minor change to the first paragraph.

PR #4396, opened 2019-03-15
fix(schema) adjustments for legacy schema auto-converter

### LINKED ISSUE #4392: Inconsistency in plugin schema handling
### Summary

When plugin schema has a field with type=table the plugin cannot be added to a route. Once the plugin is added to the route without that field the schema can be updated to contain the field with type=table, and the plugin's configuration contains the default values.
Details below

### Steps To Reproduce

1. Create a plugin schema with a field that has type=table
2. Try to add the plugin to a route -> fails with 500
3. Remove the table field from the schema
4. Add the plugin to the route -> 201
5. Add the table field back to the schema
6. Restart Kong
7. The table field is in the config of the plugin when queried with curl

### Additional Details & Logs

- Kong version 1.0.3 (kong alpine 1.0.3 container image)
- Operating system: Linux kongdev 4.15.0-1040-azure #44-Ubuntu SMP Thu Feb 21 14:24:01 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

###Details 

I have the following schema definition for my plugin:

return {
  no_consumer = true, 
  fields = {
    max_msg_size = {type = "number", required = false, default=4194304},
    capability_upstream = {type = "table", required = false, default={["string1:substring1"] = "127.0.0.1:8082", ["string2:substring2"] = "127.0.0.1:8083"}}
  },
  self_check = function(schema, plugin_t, dao, is_updating)
    -- perform any custom verification
    return true
  end
}

I would like to add the plugin to a route:
curl -i -X POST --url http://x.x.x.x:8001/routes/d3b9fb5e-8d00-493f-97f3-4217a56980a1/plugins/ --data 'name=plugin_name'

I get 500 as response, and the following traceback:
2019/03/15 12:22:29 [error] 45#0: *1157 lua coroutine: runtime error: /usr/local/share/lua/5.1/kong/db/schema/init.lua:1304: attempt to index local 'field_schema' (a nil value)
stack traceback:
coroutine 0:
        /usr/local/share/lua/5.1/kong/db/schema/init.lua: in function 'adjust_field_for_context'
        /usr/local/share/lua/5.1/kong/db/schema/init.lua:1389: in function 'adjust_field_for_context'
        /usr/local/share/lua/5.1/kong/db/schema/init.lua:1389: in function 'process_auto_fields'
        /usr/local/share/lua/5.1/kong/db/dao/init.lua:685: in function 'insert_entity'
        /usr/local/share/lua/5.1/kong/api/endpoints.lua:324: in function 'fn'
        /usr/local/share/lua/5.1/kong/api/init.lua:45: in function </usr/local/share/lua/5.1/kong/api/init.lua:28>
coroutine 1:
        [C]: in function 'resume'
        /usr/local/share/lua/5.1/lapis/application.lua:393: in function 'handler'
        /usr/local/share/lua/5.1/lapis/application.lua:130: in function 'resolve'
        /usr/local/share/lua/5.1/kong/api/init.lua:130: in function 'handler'
        /usr/local/share/lua/5.1/lapis/application.lua:130: in function 'handler'
        /usr/local/share/lua/5.1/lapis/application.lua:163: in function </usr/local/share/lua/5.1/lapis/application.lua:159>
        [C]: in function 'xpcall'
        /usr/local/share/lua/5.1/lapis/application.lua:159: in function 'dispatch'
        /usr/local/share/lua/5.1/lapis/nginx.lua:215: in function 'serve_admin_api'
        content_by_lua(nginx-kong.conf:177):2: in function <content_by_lua(nginx-kong.conf:177):1>, client: 10.244.0.1, server: kong_admin, request: "POST /routes/125b5f06-249e-4a52-812c-0a8157803679/plugins/ HTTP/1.1", host: "10.98.19.169:8001"
2019/03/15 12:22:29 [error] 45#0: *1157 [lua] init.lua:156: handle_error(): /usr/local/share/lua/5.1/lapis/application.lua:397: /usr/local/share/lua/5.1/kong/db/schema/init.lua:1304: attempt to index local 'field_schema' (a nil value)
stack traceback:
        /usr/local/share/lua/5.1/kong/db/schema/init.lua: in function 'adjust_field_for_context'
        /usr/local/share/lua/5.1/kong/db/schema/init.lua:1389: in function 'adjust_field_for_context'
        /usr/local/share/lua/5.1/kong/db/schema/init.lua:1389: in function 'process_auto_fields'
        /usr/local/share/lua/5.1/kong/db/dao/init.lua:685: in function 'insert_entity'
        /usr/local/share/lua/5.1/kong/api/endpoints.lua:324: in function 'fn'
        /usr/local/share/lua/5.1/kong/api/init.lua:45: in function </usr/local/share/lua/5.1/kong/api/init.lua:28>

stack traceback:
        [C]: in function 'error'
        /usr/local/share/lua/5.1/lapis/application.lua:397: in function 'handler'
        /usr/local/share/lua/5.1/lapis/application.lua:130: in function 'resolve'
        /usr/local/share/lua/5.1/kong/api/init.lua:130: in function 'handler'
        /usr/local/share/lua/5.1/lapis/application.lua:130: in function 'handler'
        /usr/local/share/lua/5.1/lapis/application.lua:163: in function </usr/local/share/lua/5.1/lapis/application.lua:159>
        [C]: in function 'xpcall'
        /usr/local/share/lua/5.1/lapis/application.lua:159: in function 'dispatch'
        /usr/local/share/lua/5.1/lapis/nginx.lua:215: in function 'serve_admin_api'
        content_by_lua(nginx-kong.conf:177):2: in function <content_by_lua(nginx-kong.conf:177):1>, client: y.y.y.y, server: kong_admin, request: "POST /routes/125b5f06-249e-4a52-812c-0a8157803679/plugins/ HTTP/1.1", host: "x.x.x.x:8001"
10.244.0.1 - - [15/Mar/2019:12:22:29 +0000] "POST /routes/125b5f06-249e-4a52-812c-0a8157803679/plugins/ HTTP/1.1" 500 42 "-" "curl/7.47.0"

If I delete the capability_upstream field from the schema the creation of the plugin is successful.

And here the twist comes: after enabling the plugin successfully on the route I add the "capability_upstream" field back to the schema definition and I re-start Kong. And the schema is accepted, and the default values are filled in the plugin data:

$ curl x.x.x.x:8001/plugins
{"next":null,"data":[{"created_at":1552652729,"consumer":null,"id":"ad34b0bd-1cf8-43fd-a656-fb65aff05f12","service":null,"enabled":true,"run_on":"first","name":"plugin_name","route":{"id":"125b5f06-249e-4a52-812c-0a8157803679"},"config":{"max_msg_size":4194304,"capability_upstream":{"string1:substring1":"127.0.0.1:8082","string2:substring2":"127.0.0.1:8083"}}}]}

I think it is strange. Is it the way it should work? Thank you!

hishamhm:
When auto-converting a legacy schema from a plugin, we need to determine if something of type "table" is an array, map, set or record. When we failed to pin this down, we were producing an incomplete schema, which led to runtime crashes.

if we have a table type that wasn't further specified in any way, we now make it a map from strings to strings, which is the most generic type (apart from abusing "any", since we don't want dynamically-typed plugin schemas).

To prevent similar problems from happening again, this PR also adds a defensive measure: it runs the MetaSchema validator on the schemas generated by the auto-converter as well. This validation should never fail, because the best-effort legacy schema converter should either produce a valid schema or report failure in advance. However, in case it misbehaves, the MetaSchema validator should be able to catch that, so that an invalid schema does not slip through resulting in crashes further down in the schema library.

:warning: **Note**: This PR targets `master`, but I also have a separate [working branch](https://travis-ci.org/Kong/kong/builds/506828213) with the same fixes [targeting `next`](https://github.com/Kong/kong/tree/fix/table-heuristic), because the plugin loader has been refactored for Kong 1.1. If this is merged into `master`, the best approach is to reuse the code from that other branch for the fixups when merging `master` into `next`, instead of trying to resolve the conflicts manually.

### Issues resolved

Fixes #4392.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -94,8 +94,8 @@ local validation_errors = {
   -- schema error
   SCHEMA_NO_DEFINITION      = "expected a definition table",
   SCHEMA_NO_FIELDS          = "error in schema definition: no 'fields' table",
-  SCHEMA_MISSING_ATTRIBUTE  = "error in schema definition: missing attribute %s",
   SCHEMA_BAD_REFERENCE      = "schema refers to an invalid foreign entity: %s",
+  SCHEMA_CANNOT_VALIDATE    = "error in schema prevents from validating this field",
   SCHEMA_TYPE               = "invalid type: %s",
   -- primary key errors
   NOT_PK                    = "not a primary key",
@@ -712,9 +712,6 @@ function Schema:validate_field(field, value)
     if not is_sequence(value) then
       return nil, validation_errors.ARRAY
     end
-    if not field.elements then
-      return nil, validation_errors.SCHEMA_MISSING_ATTRIBUTE:format("elements")
-    end
 
     field.elements.required = true
     for _, v in ipairs(value) do
@@ -734,9 +731,6 @@ function Schema:validate_field(field, value)
     if not is_sequence(value) then
       return nil, validation_errors.SET
     end
-    if not field.elements then
-      return nil, validation_errors.SCHEMA_MISSING_ATTRIBUTE:format("elements")
-    end
 
     field.elements.required = true
     local set = {}
@@ -753,12 +747,6 @@ function Schema:validate_field(field, value)
   elseif field.type == "map" then
     if type(value) ~= "table" then
       return nil, validation_errors.MAP
-    end
-    if not field.keys then
-      return nil, validation_errors.SCHEMA_MISSING_ATTRIBUTE:format("keys")
-    end
-    if not field.values then
-      return nil, validation_errors.SCHEMA_MISSING_ATTRIBUTE:format("values")
     end
 
     field.keys.required = true
@@ -778,9 +766,6 @@ function Schema:validate_field(field, value)
   elseif field.type == "record" then
     if type(value) ~= "table" then
       return nil, validation_errors.RECORD
-    end
-    if not field.fields then
-      return nil, validation_errors.SCHEMA_MISSING_ATTRIBUTE:format("fields")
     end
 
     local field_schema = get_field_schema(field)
@@ -958,14 +943,19 @@ validate_fields = function(self, input)
     local err
     local field = self.fields[tostring(k)]
     if field and field.type == "self" then
-      field = input
+      local pok
+      pok, err, errors[k] = pcall(self.validate_field, self, input, v)
+      if not pok then
+        errors[k] = validation_errors.SCHEMA_CANNOT_VALIDATE
+        kong.log.debug(errors[k], ": ", err)
+      end
     else
       field, err = resolve_field(self, k, field, subschema)
-    end
-    if field then
-      _, errors[k] = self:validate_field(field, v)
-    else
-      errors[k] = err
+      if field then
+        _, errors[k] = self:validate_field(field, v)
+      else
+        errors[k] = err
+      end
     end
   end
 
@@ -1697,9 +1687,6 @@ end
 -- @return A schema object, or nil and an error message.
 local function get_foreign_schema_for_field(field)
   local ref = field.reference
-  if not ref then
-    return nil, validation_errors.SCHEMA_MISSING_ATTRIBUTE:format("reference")
-  end
 
   local foreign_schema = _cache[ref] and _cache[ref].schema
   if not foreign_schema then
--- a/kong/db/dao/plugins.lua
+++ b/kong/db/dao/plugins.lua
@@ -181,6 +181,14 @@ local function convert_legacy_schema(name, old_schema)
     if new_fdata.type == "array" then
       new_fdata.elements = elements
     end
+
+    if (new_fdata.type == "map" and new_fdata.keys == nil)
+       or (new_fdata.type == "record" and new_fdata.fields == nil) then
+      new_fdata.type = "map"
+      new_fdata.keys = { type = "string" }
+      new_fdata.values = { type = "string" }
+    end
+
     if new_fdata.type == nil then
       new_fdata.type = "string"
     end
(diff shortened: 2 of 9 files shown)


kong/db/schema/init.lua
```diff
@@ -94,8 +94,8 @@ local validation_errors = {
   -- schema error
   SCHEMA_NO_DEFINITION      = "expected a definition table",
   SCHEMA_NO_FIELDS          = "error in schema definition: no 'fields' table",
-  SCHEMA_MISSING_ATTRIBUTE  = "error in schema definition: missing attribute %s",
   SCHEMA_BAD_REFERENCE      = "schema refers to an invalid foreign entity: %s",
+  SCHEMA_CANNOT_VALIDATE    = "error in schema prevents from validating this field",
```
thibaultcha: Due diligence here: will said error be visible to the user somewhere?

hishamhm: The error in case is a crash in the schema library caught by `pcall` (a840a61 has the full story). I chose not to expose the pcall error as that won't really tell the user what the problem is (and could make the user think there was a bug in Kong instead), but at least this SCHEMA_CANNOT_VALIDATE message will be matched to the specific key that caused it, so at least that's _some_ indication on where their problem is.

(The pcall error could help _us_ to explain the user what their problem is, though...)

thibaultcha: Maybe worth exposing it in a debug log?

hishamhm: ohh, good idea

hishamhm: Done!

PR #4400, opened 2019-03-18
feat(migrations) auto-detect migrations files

Tieske:
Instead of creating a file and updating the list of files, now just
adding the file will suffice. The file will be auto-detected and
added to the list of migrations.

Filename needs to have an initial number component indicating the
order (numerically sorted, so `001 == 1 == 01`).

### DIFF
--- a/kong/db/migrations/core/init.lua
+++ b/kong/db/migrations/core/init.lua
@@ -1,5 +1,54 @@
-return {
-  "000_base",
-  "001_14_to_15",
-  "002_15_to_1",
-}
+local pl_path = require "pl.path"
+local pl_dir = require "pl.dir"
+
+local function get_files()
+  -- find script path remember to strip off the starting @
+  -- should be like: 'kong/db/migrations/core/init.lua' but
+  -- detecting it dynamically for robustness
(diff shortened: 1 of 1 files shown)


hishamhm:
This solution is interesting, but it was hardcoded to the `core` subsystem of migrations — the migrations system supports multiple subsystems (including those in plugins). It feels like such a change would be best implemented in the migrations library itself, otherwise we end up with inconsistent behavior (`core` magically doesn't need explicit entries, but other migrations need it).

kikito:
I have an additional concern wrt this one: one must remember to include the migration file in the rockspec file. If we add this "automation" I am concerned that people will add the file to the folder but not to the rockspec.

Tieske:
> I have an additional concern wrt this one: one must remember to include the migration file in the rockspec file. If we add this "automation" I am concerned that people will add the file to the folder but not to the rockspec.

there is a test for that, it would fail.

hishamhm:
I still think that having it in one file and not the others would be confusing. Better to have this as a feature of the migrations system or not at all.

kikito:
Hi Thijs, thanks for this, we decided the advantages provided by the PR don't out-weight the extra complexity involved so we decided to close this one down.

PR #4403, opened 2019-03-18
chore(ci) kong next daily build naming fix

hutchic:
the inclusion of `.requirements` doesn't let us specify package names when building releases. Example of this is https://bintray.com/beta/#/kong/kong-nightly/next?tab=overview where the package is always being named for what's in the `.requirements` file and not using the ENV override

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ else
 OPENSSL_DIR ?= /usr
 endif
 
-include .requirements
 .PHONY: install dev lint test test-integration test-plugins test-all fix-windows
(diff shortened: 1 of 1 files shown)


hishamhm:
those values are all already being grepped by the `?=` calls there so the `include` was never really necessary even on `master`, right?

hutchic:
@hishamhm correct 👍

hishamhm:
@hutchic thanks, just making sure I was understanding what I was reviewing :)

PR #4404, opened 2019-03-18
docs(cmd) add help option to cli output

darrenjennings:
### Summary

It is not immediately obvious from the kong cli to see how you might see info about a specific command. However, you can pass `-h` and it will not give warning about command not being known:

```bash
$ kong check -h
```

Output (already a feature, this PR just adds the doc)
```
Usage: kong check <conf>

Check the validity of a given Kong configuration file.

<conf> (default /etc/kong/kong.conf) configuration file

Options:
 -h, --help       help
 --v              verbose
 --vv             debug
```

### DIFF
--- a/kong/cmd/init.lua
+++ b/kong/cmd/init.lua
@@ -6,7 +6,7 @@ local pl_app = require "pl.lapp"
 local log = require "kong.cmd.utils.log"
 
 local options = [[
- -h               help
+ -h, --help       help
  --v              verbose
  --vv             debug
 ]]
--- a/kong/cmd/init.lua
+++ b/kong/cmd/init.lua
@@ -6,6 +6,7 @@ local pl_app = require "pl.lapp"
 local log = require "kong.cmd.utils.log"
 
 local options = [[
+ -h               help
  --v              verbose
  --vv             debug
 ]]


thibaultcha:
This would be a nice to have!

I don't think the implementation is quite there yet though. There are a couple of tweaks that we should make to the CLI to properly support the semantics of a global/per command `-h` flag.

i.e.

```
$ kong -h
No such command: -h

Usage: kong COMMAND [OPTIONS]

The available commands are:
 check
 config
 health
 migrations
 prepare
 quit
 reload
 restart
 roar
 start
 stop
 version

Options:
 -h, --help       help
 --v              verbose
 --vv             debug
$ echo $?
1
```

Notice how the flag was actually not recognized by Kong's CLI, and produces a `No such command: -h` output, plus exited with status code `1`.

With subcommands, the output is slightly different:
```
$ kong check -h
Usage: kong check <conf>

Check the validity of a given Kong configuration file.

<conf> (default /etc/kong/kong.conf) configuration file

Options:
 -h, --help       help
 --v              verbose
 --vv             debug
$ echo $?          
1
```

The exit status code is the same, but we don't see a "No such command" output (that's expected, as per the current implementation of our CLI), and instead, we see that `pl.lapp` was unable to make sense of a `-h` option, and thus printed the usage of the command.

Also, note that this output is sent to `stderr` instead of `stdout`, as is typical of `--help` flags:

```
$ kong check -h 2>/dev/null
$ 
```

Note how the global flags (`-v` and `--vv`) are currently implemented (we sort of override `pl.lapp` behavior):

https://github.com/Kong/kong/blob/master/kong/cmd/init.lua#L75

thibaultcha:
Oh, and of course, such behavior would require tests in this suite: https://github.com/Kong/kong/tree/master/spec/02-integration/02-cmd - or else we cannot assume the behavior of the `-h` flag is stable. The need for tests is further underlined due to our heavy customization of the `pl.lapp` module (sub-commands, global flags, stdout/stderr and exit status code expectations...).

darrenjennings:
Yea it's not global for sure. It seemed to be ok to show `-h, --help` in options for normal `kong` output, since it the `[OPTIONS]` is specified only when COMMAND is also present. For example. `kong --v` also produces `No such command: --v`.

You're right that the exit code is `1`, which might be purposeful since penlight is "bailing" before any other args are parsed and it does have special `-h` and `--help` handling:
https://github.com/stevedonovan/Penlight/blob/b9cc7822c11288704b8757c2d18244eccd3393d4/lua/pl/lapp.lua#L212

From https://github.com/stevedonovan/Penlight/commit/c222e422274cf152cf8202ee9e3861fa6e9cce6d
```
Also handles -h and --help early, before parsing command line arguments
 Otherwise it would be impossible to show the help, if conversion of the
 default value of some argument fails.
```

Agree also on the tests, and will maybe try and take another stab at supporting it "globally".

darrenjennings:
Closing PR as it cannot suffice until help functionality is better.

thibaultcha:
@darrenjennings Do you have plans to follow up with a revised version of this PR? I'd be happy to have a chat with you if you have questions or want to debate some implementation-specific details!

PR #4407, opened 2019-03-19
perf(router) cache service loading in router

kikito:
The current implementation of the router loads everything from the
database, all the time.

Since it iterates over routes, and then it fetches the relevant
service for each route, this means that given a single service with
1000 routes, it will make 1000 requests for the database, returning
the same service.

Some notes regarding "Services changing while the router is being
rebuilt":

* This PR sacrifices some "correctness" for speed: when a service
  is changed, it takes time for that change to be propagated to other
  workers and kong nodes. If a service is changed it will take some
  seconds before that change is made effective in the router.

* Consider that even on the current case (were no caching is used),
  if the Service changes while the router iterates over route number
  501, the first 500 routes will be "wrong" anyway, until the router
  gets refreshed - so at the end we don't loose that much with this
  change.

* This PR assumes that kong.cache is/can be prewarmed by an external
  prewarming service. Prewarming the cache is considered out of the 
  scope of this PR.

Extracted from #4488 (with improvements)

Co-authored-by: Aapo Talvensaari <aapo.talvensaari@gmail.com>

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -91,17 +91,66 @@ do
     return false
   end
 
-  local function get_service_for_route(db, route)
+
+  local function load_service_from_db(service_pk)
+    local service, err = kong.db.services:select(service_pk)
+    -- kong.db.services:select returns a third value, which we discard here:
+    return service, err
+  end
+
+
+  local function build_services_init_cache(db)
+    local services_init_cache = {}
+
+    for service, err in db.services:each(1000) do
+      if err then
+        return nil, err
+      end
+
+      services_init_cache[service.id] = service
+    end
+
+    return services_init_cache
+  end
+
+
+  local function get_service_for_route(db, route, services_init_cache)
     local service_pk = route.service
     if not service_pk then
       return nil
     end
 
-    local service, err = db.services:select(service_pk)
-    if not service then
-      return nil, "could not find service for route (" .. route.id .. "): " ..
-                  err
+    local id = service_pk.id
+    local service = services_init_cache[id]
+    if service then
+      return service
     end
+
+    local err
+
+    -- kong.cache is not available on init phase
+    if kong.cache then
+      local cache_key = db.services:cache_key(service_pk.id)
+      service, err = kong.cache:get(cache_key, CACHE_ROUTER_OPTS,
+                                    load_service_from_db, service_pk)
+
+    else -- init phase, not present on init cache
+
+      -- A new service/route has been inserted while the initial route
+      -- was being created, on init (perhaps by a different Kong node).
+      -- Load the service individually and update services_init_cache with it
+      service, err = load_service_from_db(service_pk)
+      services_init_cache[id] = service
+    end
+
+    if err then
+      return nil, "error raised while finding service for route (" .. route.id .. "): " ..
+                  err
+
+    elseif not service then
+      return nil, "could not find service for route (" .. route.id .. ")"
+    end
+
 
     -- TODO: this should not be needed as the schema should check it already
     if SUBSYSTEMS[service.protocol] ~= subsystem then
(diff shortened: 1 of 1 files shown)


kong/runloop/handler.lua
```diff
@@ -78,13 +78,26 @@ do
     return false
   end
 
+
+  local function load_service_from_db(service_pk)
+    return kong.db.services:select(service_pk)
```
thibaultcha: Better discard the third return value here, which is interpreted as an overridden TTL by mlcache. mlcache plays safe and will discard it if not a number, but let's be explicit about this and not rely on such low-level behavior.

kong/runloop/handler.lua
```diff
@@ -78,13 +78,26 @@ do
...
   local function get_service_for_route(db, route)
     local service_pk = route.service
     if not service_pk then
       return nil
     end
 
-    local service, err = db.services:select(service_pk)
+    local service, err
+    if kong.cache then -- the kong.cache module is sometimes missing during tests
+      local cache_key = kong.db.services:cache_key(service_pk.id)
```
thibaultcha: `db` is already given as an argument to this function, better skip the `kong.` lookup cost

kong/runloop/handler.lua
```diff
@@ -78,13 +78,26 @@ do
...
+      service, err = kong.cache:get(cache_key, CACHE_ROUTER_OPTS,
+                                    load_service_from_db, service_pk)
+    else
+      service, err = load_service_from_db(service_pk)
+    end
+
     if not service then
       return nil, "could not find service for route (" .. route.id .. "): " ..
```
thibaultcha: Sometimes as we've seen, `service` will be `nil` not because of an error, but because it is not found (eventual consistency between the Routes/Services relationship). We should probably distinguish `if err then ... elseif not service then ... end` and produce a different log in each branch (avoiding the concatenation of a `nil` value which supersede the underlying error. Something to keep in mind as we work on other router-related PRs.

kong/runloop/handler.lua
```diff
@@ -78,13 +78,26 @@ do
     return false
   end
 
+
+  local function load_service_from_db(service_pk)
+    return kong.db.services:select(service_pk)
+  end
+
```
thibaultcha: style: missing line jump

kong/runloop/handler.lua
```diff
@@ -78,13 +78,26 @@ do
...
       return nil
     end
 
-    local service, err = db.services:select(service_pk)
+    local service, err
+    if kong.cache then -- the kong.cache module is sometimes missing during tests
+      local cache_key = kong.db.services:cache_key(service_pk.id)
+      service, err = kong.cache:get(cache_key, CACHE_ROUTER_OPTS,
+                                    load_service_from_db, service_pk)
+    else
```
thibaultcha: style: missing line jump above

kong/runloop/handler.lua
```diff
@@ -78,13 +78,26 @@ do
...
+
   local function get_service_for_route(db, route)
     local service_pk = route.service
     if not service_pk then
       return nil
     end
 
-    local service, err = db.services:select(service_pk)
+    local service, err
+    if kong.cache then -- the kong.cache module is sometimes missing during tests
```
thibaultcha: imho, tests should be explicit and require/mock a cache instance. This is a dangerous precedent being made here, in which production (proxy path) code is being tested differently during tests than it is at runtime.

kikito: You are absolutely right about this one. I went back and removed the if, in order to do just that...

... and then I discovered that we create the initial router on `init` phase, but the cache is only available after `init_worker` phase. So on the first iteration this was still doing the 1000 db queries to the database.

I have changed the code so that it falls back to a regular lua table on init phase, and also left comments in the code explaining that the `if kong.cache ...` happens because of this, not because of tests, as I first thought.

Thanks for your feedback!

thibaultcha: Awesome! :100:

PR #4408, opened 2019-03-19
POC - [do not merge] - feat(runloop) cache prewarm for entities (configurable)

kikito:
This preloads all plugins & services from the db into the cache during
the init_worker phase. It should accelerate the first run of the
plugins_iterator without affecting any existing functionality or tests.

It should also improve the performance of the router, provided that
the services are loaded from the cache instead of directly from the
database - See #4407

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -75,6 +75,7 @@ local declarative = require "kong.db.declarative"
 local ngx_balancer = require "ngx.balancer"
 local kong_resty_ctx = require "kong.resty.ctx"
 local certificate = require "kong.runloop.certificate"
+local cache_prewarm = require "kong.runloop.cache_prewarm"
 local concurrency = require "kong.concurrency"
 local plugins_iterator = require "kong.runloop.plugins_iterator"
 local balancer_execute = require("kong.runloop.balancer").execute
@@ -284,6 +285,21 @@ local function load_declarative_config(kong_config, entities)
 
     mesh.init()
 
+    return true
+  end)
+end
+
+
+local function execute_cache_prewarm(kong_config)
+  if kong_config.database == "off" then
+    return true
+  end
+
+  return concurrency.with_worker_mutex({ name = "cache_prewarm" }, function()
+    local ok, err = cache_prewarm.execute()
+    if not ok then
+      return nil, err
+    end
     return true
   end)
 end
@@ -534,6 +550,11 @@ function Kong.init_worker()
     return
   end
 
+  ok, err = execute_cache_prewarm(kong.configuration)
+  if not ok then
+    ngx_log(ngx_CRIT, "error prewarming cache: ", err)
+    return
+  end
 
   runloop.init_worker.before()
 
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -290,13 +290,13 @@ local function load_declarative_config(kong_config, entities)
 end
 
 
-local function execute_cache_prewarm(kong_config)
+local function execute_cache_prewarm(kong_config, configured_plugins)
   if kong_config.database == "off" then
     return true
   end
 
   return concurrency.with_worker_mutex({ name = "cache_prewarm" }, function()
-    local ok, err = cache_prewarm.execute()
+    local ok, err = cache_prewarm.execute(configured_plugins)
     if not ok then
       return nil, err
     end
@@ -550,7 +550,7 @@ function Kong.init_worker()
     return
   end
 
-  ok, err = execute_cache_prewarm(kong.configuration)
+  ok, err = execute_cache_prewarm(kong.configuration, configured_plugins)
   if not ok then
     ngx_log(ngx_CRIT, "error prewarming cache: ", err)
     return
--- a/kong/runloop/cache_prewarm.lua
+++ b/kong/runloop/cache_prewarm.lua
@@ -1,5 +1,15 @@
+local constants = require "kong.constants"
+
+
+local CORE_ENTITIES = {}
+do
+  for _, entity_name in ipairs(constants.CORE_ENTITIES) do
+    CORE_ENTITIES[entity_name] = true
+  end
+end
+
+
 local tostring = tostring
-local ipairs = ipairs
 local math = math
 local ngx = ngx
 
@@ -7,18 +17,7 @@ local ngx = ngx
 local cache_prewarm = {}
 
 
-local ENTITIES_TO_PREWARM = {
-  "services",
-  "plugins",
-}
-
-
-local function cache_prewarm_single_entity(entity_name)
-  local dao = kong.db[entity_name]
-  if not dao then
-    return nil, "Invalid entity name found when prewarming the cache: " .. tostring(entity_name)
-  end
-
+local function cache_prewarm_single_entity(entity_name, dao)
   ngx.log(ngx.NOTICE, "Preloading '" .. entity_name .. "' into the cache ...")
 
   local start = ngx.now()
@@ -55,25 +54,35 @@ end
 -- access. This function is intented to be used during worker initialization
 -- The list of entities to be loaded is defined by the ENTITIES_TO_PREWARM
 -- variable.
-function cache_prewarm.execute()
+function cache_prewarm.execute(configured_plugins)
+
   -- kong.db and kong.cache might not be active while running tests
   if not kong.db or not kong.cache then
     return true
   end
 
-  for _, entity_name in ipairs(ENTITIES_TO_PREWARM) do
-    local ok, err = cache_prewarm_single_entity(entity_name)
-    if not ok then
-      if err == "no memory" then
-        kong.log.warn("cache prewarming has been stopped because cache ",
-                      "memory is exhausted, please consider increasing ",
-                      "the value of 'mem_cache_size' (currently at ",
-                      kong.configuration.mem_cache_size, ")")
-
-        return true
+  for entity_name, dao in pairs(kong.db.daos) do
+    if dao.schema.prewarm then
+      local prewarm = CORE_ENTITIES[entity_name]
+      if not prewarm and dao.plugin_name then
+        prewarm = configured_plugins[dao.plugin_name]
       end
 
-      return nil, err
+      if prewarm then
+        local ok, err = cache_prewarm_single_entity(entity_name, dao)
+        if not ok then
+          if err == "no memory" then
+            kong.log.warn("cache prewarming has been stopped because cache ",
+                          "memory is exhausted, please consider increasing ",
+                          "the value of 'mem_cache_size' (currently at ",
+                          kong.configuration.mem_cache_size, ")")
+
+            return true
+          end
+
+          return nil, err
+        end
+      end
     end
   end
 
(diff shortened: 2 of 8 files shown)


kong/init.lua
```diff
@@ -289,6 +290,21 @@ local function load_declarative_config(kong_config, entities)
 end
 
 
+local function execute_cache_prewarm(kong_config)
+  if kong_config.database == "off" then
+    return true
+  end
+
+  return concurrency.with_worker_mutex({ name = "cache_prewarm" }, function()
```
bungle: If we do `with_worker_mutex` does it mean that cache for `L3` is only warmed on single worker?

p0pr0ck5:
Just curious, how well does this handle cases where there are more routes/services than can fit in `mem_cache_size`?

p0pr0ck5:
Also, this likely has very little impact on the plugins iterator performance, as the iterator has cache keys related to each relevant plugins (e.g., `route` + `service` + `consumer` + `plugin`) - prewarming single entities like services and plugins won't have an impact here sadly.

bungle:
@p0pr0ck5,

that's good point that cache keys on plugins iterator point to combinations, so you need to cache plugins for every combination. @thibaultcha, it would perhaps be a good idea to make support for `mlcache` to support cache key aliases?

kikito:
Marking this PR as "do not merge" because for now it's a POC. There are other considerations, including a partial reimplementation of the router, to take into account before we merge this. 

Will leave the PR open for a while, in order to have a place for discussion.

kikito:
@p0pr0ck5:

> Just curious, how well does this handle cases where there are more routes/services than can fit in mem_cache_size?

This is a quick POC so I have not run extensive analysis on this (feel free if you can spare the cycles, and please share :)!).

My gut feeling (which is just that, a feeling) is that using some form of caching (vs querying the db once per route) will have more impact than anything we do on this PR. In other words, #4407, or your own proposal from several days ago, should improve the router one order of magnitude more than anything we do here.

Now, to answer your question directly: yes, if the cache is exhausted, be it because there are too many services or because other parts of Kong fill it up, Kong's overall performance might go down. Not only in the router - everywhere where cache is used.

> Also, this likely has very little impact on the plugins iterator performance, as the iterator has cache keys related to each relevant plugins (e.g., route + service + consumer + plugin) - prewarming single entities like services and plugins won't have an impact here sadly.

I'm not sure I follow. The call this PR is trying to accelerate is this one:

https://github.com/Kong/kong/blob/next/kong/runloop/plugins_iterator.lua#L30

Say, for example, that a plugin is configured on a service (there is something in the `plugin.service_id` column on the db). The prewarm will store it in the cache under an entry which will look something like `plugins:my_plugin:<service_id>::`.

When a request arrives to the plugin iterator, it will go through [this loop](https://github.com/Kong/kong/blob/next/kong/runloop/plugins_iterator.lua#L113). For simplicity, let's assume no auth, so we can ignore everything `consumer`-related. Then the iterator will try fetching the plugin from the following keys, on this order:
* The `if route_id then ...` conditional will look for `plugins:my_plugin:<route_id>`. It will be a cache miss, prompting a db query, returning an empty guard object.
* Then the `if service_id then ...` conditional will look for `plugins:my_plugin:<service_id>`. It will be a cache hit, and *will not* query the database.

So the presented cache prewarming might have *some* impact on the plugins iterator after all. How big of an impact will depend on the ratio of plugins vs routes, services and consumers. For a someone with lots of plugins but few of the others it could make a difference. The worst case scenario would be someone with a single plugin and a lot of services, routes and consumers - in that case the advantages brought by prewarming would be lost in a sea of cache misses.

We could improve our odds by caching more combinations, as @bungle says. I think that might be overstepping outside of the boundaries of a 2-points POC. Will discuss during planning today.

p0pr0ck5:
> The prewarm will store it in the cache under an entry which will look something like plugins:my_plugin:<service_id>::.

Are we sure that's the case? From what I can tell each entity is loaded individually, but I don't see how compound cache keys are getting generated? Apologies if I'm missing something :)

thibaultcha:
> Just curious, how well does this handle cases where there are more routes/services than can fit in mem_cache_size?

We mentioned several times before that under this scenarios, Kong should notify the user. The first thing we thought about was, of course, printing a warning log to the user. Other suggestions welcome (e.g. Admin API `/` endpoint notice, or even, in an ideal world, keeping track of our ratio of evictions and periodically remind the user that the cache is churning, which would be useful not only for pre-warm, but also runtime).

Pre-warming is only a piece of the puzzle. On top of the router rebuild PRs from @bungle, we have additional work to do in the plugins iterator (e.g. skip NOP function calls), separating proxy and Admin API C* consistencies, and separating the core cache (and cache miss) from the client-facing cache are all equally important pieces of work that need to happen.

kikito:
> Are we sure that's the case? From what I can tell each entity is loaded individually, but I don't see how compound cache keys are getting generated? Apologies if I'm missing something :)

Ah, yes, I see the problem. The [`DAO:cache_key` method](https://github.com/Kong/kong/blob/master/kong/db/dao/init.lua#L942) is a bit more complex than it used to be: if you pass several strings to it, it will still just concatenate them with `:` and return the result.

But it has new powers now. When you pass a *table* on the first parameter, `cache_key` will assume it is dealing with an entity, and extract the "cache key strings" from it. It will do this by using the schema's `cache_key` property, which is `{ "name", "route", "service", "consumer" }` [on plugins](https://github.com/Kong/kong/blob/master/kong/db/schema/entities/plugins.lua#L8)

Hence:
```
db.plugins:cache_key(plugin_with_route_id) => `plugins:my_plugin:<route_id>::`
db.plugins:cache_key(plugin_with_service_id) => `plugins:my_plugin::<service_id>:`
db.plugins:cache_key(plugin_with_consumer_id) => `plugins:my_plugin:::<consumer_id>`
```
And a plugin with both a service and a consumer will have both on its cache_key.

On this PR I am using this second version of the `cache_key` method exclusively when preloading the entities.

kong/init.lua
kikito: I answered this concern on private but forgot to reflect the answer here:

The reasoning behind this PR is simple: minimizing db access as much as possible, replacing it with either cache or raw Lua memory when possible.

Both the router and the plugins iterator are obvious candidates for such optimizations because they retrieve plugins and services "piecemeal" - with one database request (or more) per service/plugin.

In the case of services, for the scenario of a single Service with 1000 routes, the current implementation will do 1000 database request to retrieve such service - once per route. With the proposed POC, only 1 request is made, and then cached (assuming #4407).

In a different scenario, when there are 1000 services, each with 1 route, there will be 1000 database requests to retrieve those 1000 services. Using the cache (again, in consonance with #4407) minimizes the amount of repeated piecemeal requests to the minimal possible, but they would still happen more often than needed when the cache isn't hot. Prewarming would potentially do 1 or 2 db accesses instead of several thousands.

A similar thing happens with the plugins iterator. The main difference here is that the way multiple keys in the cache can retrieve a single plugin. The provided POC only prewarms *some* of the keys that might be used.

kong/runloop/cache_prewarm.lua
```diff
@@ -0,0 +1,83 @@
...
+  -- kong.db and kong.cache might not be active while running tests
+  if not kong.db or not kong.cache then
+    return true
+  end
+
+  for _, entity_name in ipairs(ENTITIES_TO_PREWARM) do
+    local ok, err = cache_prewarm_single_entity(entity_name)
+    if not ok then
+      if err == "no memory" then
+        kong.log.warn("cache memory is exhausted, please consider increasing ",
```
dndx: Maybe we can make this less alarming by rephrasing it as:

`"cache prewarming has been stopped because cache memory is exhausted, please consider increasing..."`

To emphasize that this does not affect Kong's ability to start and serve requests (versus loading declarative configs where same condition is a fatal error).

hbagdi:
> We mentioned several times before that under this scenarios Kong should notify the user. The first thing we thought about was, of course, printing a warning log to the user. Other suggestions welcome (e.g. Admin API / endpoint notice, or even, in an ideal world, keeping track of our ratio of evictions and periodically remind the user that the cache is churning, which would be useful not only for pre-warm, but also runtime).

This is where it becomes very important that we have a range of health in Kong.
Health of Kong can never be binary. Making sure the process is running is not enough, depending on the configuration Kong will operate better/worse given a set of configuration values.

We should give users metrics to decide this for themselves, which will be leveraged by pro and serious users. In some  cases, users will be using Kong but won't have in-depth knowledge of how it works and in that case we need a coarse metric which sums up how Kong is functioning.

To start the discussion, here is what I propose (we can move this to a Github Issue or Kong Nation to avoid hijacking this PR):

- We should have a health status in Kong as part of public API in Lua land and in Admin API, which has three states: `green`, `yellow`, `red`. The states represent how is Kong functioning. If there are a lot of cache misses on router rebuilds or other critical components, we should be `red` state. If we have DB reachability issues, we need to slowly transition from `green` to `yellow` to `red`, based on the rate and severity of errors that occur due to DB reachability.
- Please note that this is separate from a `200 OK` response on the Admin API endpoint `/status` which is merely a check for orchestration systems to detect if Kong process is running or not.


Summing up, error logs in this PR are obviously important to communicate _what_ is wrong but we need to add in feature to notify users when something _is_ wrong.

bungle:
I think everyone agrees with @hbagdi, and this probably points out a problem of where thingdös are discussed. Yes the context of discussion is close to this PR, but on the other hand, the scope of discussion is not something that fits into this PR. I can take a note on our backlog so that the comments Harry made won't be lost and forgotten when this is merged.

kong/runloop/cache_prewarm.lua
bungle: Changed that.

kong/init.lua
hishamhm: Getting back to @bungle's original question:

> If we do with_worker_mutex does it mean that cache for L1 is only warmed on a single worker?

No, `with_worker_mutex` here means that it will still run for all workers, but only one at a time. It also means that the one doing the DB access to perform the prewarming will end up holding back the initialization of all workers. (Which would well be a nice side-effect, so that workers only become responsive once prewarmed.)

However, it also means all workers are doing the `dao:each()` in `cache_prewarm_single entity` (sequentially among workers even!), which is probably not what we want. Given this, doing it only for a single worker using `if ngx.worker.id() == 0` would be best (yes, this would mean that L1 is only warm in a single worker, but what we care here is DB access).

p0pr0ck5:
One thing I think would be very helpful (well, I have some use cases already) would be exposition of the status of percentage or number of entities cached during initialization. That is, Kong exposes a value in the Admin API that indicates the progress of prewarming work. That would allow a smart downstream balancer to balance inbound traffic to Kong nodes based on a percentage of "readiness" to accept queries that will not result in anomalous latency (this of course is a pretty dumb strategy without some complex traffic analysis, but it's a valid use case still :) ). I don't think that change is in scope here, but if folks here think it's a worthwhile addition I'd be contribute a patch :)

kikito:
closing, done differently in other requests

PR #4410, opened 2019-03-19
feat(reports) report anonymous usage data of dbless feature

dndx:
### Summary
This PR reports certain usage of the new DB-less or declarative config feature.

* Add tests for hourly "ping" report when `database = off`
* Generate `dbless-reconfigure` event when `/config` API is used to reconfigure with tests
* Generate `config-db-import ` event when `kong config db_import` cmd is used to import config from `.yml` into database with tests
* Fix small styling issue in `/config` API endpoint

### DIFF
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -14,7 +14,7 @@ return {
   ["/config"] = {
     POST = function(self, db)
       if kong.db.strategy ~= "off" then
-        kong.response.exit(400, {
+        return kong.response.exit(400, {
           message = "this endpoint is only available when Kong is " ..
                     "configured to not use a database"
         })
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -1,4 +1,5 @@
 local declarative = require("kong.db.declarative")
+local reports = require("kong.reports")
 local kong = kong
 
 
@@ -7,6 +8,10 @@ local kong = kong
 local accept = {
   yaml = true,
   json = true,
+}
+
+local _reports = {
+  decl_fmt_version = false,
 }
 
 
@@ -24,9 +29,9 @@ return {
 
       local config = self.params.config
       -- TODO extract proper filename from the input
-      local entities, err = dc:parse_string(config, "config.yml", accept)
-      if err then
-        return kong.response.exit(400, { error = err })
+      local entities, err_or_ver = dc:parse_string(config, "config.yml", accept)
+      if not entities then
+        return kong.response.exit(400, { error = err_or_ver })
       end
 
       local ok, err = declarative.load_into_cache(entities)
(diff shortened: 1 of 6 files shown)


kong/api/routes/config.lua
```diff
@@ -9,12 +11,16 @@ local accept = {
   json = true,
 }
 
+local _reports = {
+    _format_version = false,
```
thibaultcha: style: indentation should be 2 spaces

kong/api/routes/config.lua
```diff
@@ -1,5 +1,7 @@
 local declarative = require("kong.db.declarative")
+local reports = require("kong.reports")
 local kong = kong
+local reports_send = reports.send
```
thibaultcha: There should not be any need for localizing this function in such a cold code path? Less upvalues is better in those scenarios

kong/cmd/config.lua
```diff
@@ -97,6 +97,16 @@ local function execute(args)
...
       log("import successful")
 
+      -- send anonymous report if reporting is not disabled
+      if conf.anonymous_reports then
+        local kong_reports = require "kong.reports"
+        kong_reports.configure_ping(conf)
+        kong_reports.toggle(true)
+
+        local report = { _format_version = err_or_ver }
+        kong_reports.send("dbless-import", report)
```
thibaultcha: This mode is not db-less, since it imports the declarative config file in the DB itself; this signal should be something like `db-import`

kong/api/routes/config.lua
```diff
@@ -35,6 +41,9 @@ return {
         return kong.response.exit(500, { message = "An unexpected error occurred" })
       end
 
+      _reports._format_version = err_or_ver
```
thibaultcha: I suggest a more descriptive name for this attribute, since the receiving end may be confused (it somewhat lacks of context). Maybe something like `decl_fmt_version`? Also, `_` prefixed values were for booleans (yes/no enabled listeners etc). For this attribute, better removing it.

spec/02-integration/02-cmd/11-config_spec.lua
```diff
@@ -62,16 +68,25 @@ describe("kong config", function()
...
+    }))
+
+    local thread = helpers.udp_server(constants.REPORTS.STATS_PORT)
 
     assert(helpers.kong_exec("config db_import " .. filename, {
       prefix = helpers.test_conf.prefix,
     }))
 
+    local _, res = assert(thread:join())
+    assert.matches("signal=dbless%-import", res)
```
thibaultcha: FYI: here, you could use a more readable expected string like so:
```lua
assert.matches("signal=dbless-import", res, nil, true)
```

It respects the `string.match` arguments.

spec/02-integration/04-admin_api/11-reports_spec.lua
```diff
@@ -299,5 +314,28 @@ for _, strategy in helpers.each_strategy() do
       assert.match("name=tcp%-log", reports_data[1])
     end)
 
+    if strategy == "off" then
+      it("#only reports declarative reconfigure via /config", function()
```
thibaultcha: Better remove this `#only` tag before merging this PR

PR #4411, opened 2019-03-19
tests(*) plugin schema helper

gszr:
### Summary

- Override `Entity.new_subschema` in schema instance returned by `Entity.new`
- Use `Entity.new` to instantiate plugins schemas in spec helpers

These changes allow the helper `validate_plugin_config_schema` to be consistent with how plugins schemas are loaded in the plugins DAO. In particular, this fixes an issue where default values for nested records are not populated in the entity validated and returned by `validate_plugin_config_schema`.

### DIFF
--- a/kong/db/schema/entity.lua
+++ b/kong/db/schema/entity.lua
@@ -23,6 +23,14 @@ local function make_records_required(field)
   end
 end
 
+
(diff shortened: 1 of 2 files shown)


hishamhm:
I discussed these changes with @gszr — patch looks good, waiting for CI.

PR #4413, opened 2019-03-20
feat(api) expose schemas of db entities on the Admin API

hbagdi:
Kong currently exposes schemas of plugin configurations at
`/plugins/schema/:plugin_name`.

This commit exposes schema of all core and custom entities in Kong at
`/schemas/:name`.

### DIFF
--- a/kong/api/routes/kong.lua
+++ b/kong/api/routes/kong.lua
@@ -2,6 +2,7 @@ local utils = require "kong.tools.utils"
 local singletons = require "kong.singletons"
 local conf_loader = require "kong.conf_loader"
 local cjson = require "cjson"
+local api_helpers = require "kong.api.api_helpers"
 
 local sub = string.sub
 local find = string.find
@@ -15,6 +16,18 @@ local knode  = (kong and kong.node) and kong.node or
 local tagline = "Welcome to " .. _KONG._NAME
 local version = _KONG._VERSION
 local lua_version = jit and jit.version or _VERSION
+
+
+local strip_foreign_schemas = function(fields)
+  for _, field in ipairs(fields) do
+    local fname = next(field)
+    local fdata = field[fname]
+    if fdata["type"] == "foreign" then
+      fdata.schema = nil
+    end
+  end
+end
+
 
 return {
   ["/"] = {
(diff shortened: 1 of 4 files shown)


hbagdi:
This is currently targeting `next` but should target `master` once Kong 1.1 is released and `next` is merged into `master`.

hbagdi:
> Piggy backing on our conversation from last week, this endpoint could be later extended to support the POST /schemas/:name/ endpoint (or POST /schemas/:name/validate), which would validate a given payload against a given endpoint.

@thibaultcha I've got a local working commit for `POST /schemas/:name/validate`, but I need to write tests for it. I'll PR that work once this gets merged in.

thibaultcha:
@hbagdi Great to hear!

hishamhm:
> This is currently targeting next but should target master once Kong 1.1 is released and next is merged into master.

@hbagdi it should continue to target `next`, as this will be included in Kong 1.2 :)

hbagdi:
With this endpoint, along with `/plugins/schema/:name` endpoint, all the schema information from Kong will be exposed to end user except [root level fields inside plugins](https://github.com/Kong/kong/blob/master/kong/plugins/key-auth/schema.lua#L8-L9), like `run_on` and `protocols`, which are fairly new.

Since, we can't break `/plugin/schema/:name`, an alternate solution here could be to expose the entire schema(not just config) for each plugin at `/schemas/plugins/:plugin_name`.

Thoughts?

thibaultcha:
@hbagdi Great idea! We will need to come up with a deprecation plan for the old endpoint, but it will likely stay around for the remaining lifecycle of the 1.0 family, no rush here.

kong/api/routes/kong.lua
```diff
@@ -118,5 +119,30 @@ return {
...
+  },
+  ["/schemas/:name"] = {
+    GET = function(self, db, helpers)
+      local entity = kong.db[self.params.name]
+      local schema = entity and entity.schema or nil
+      if not schema then
+        return kong.response.exit(404, { message = "No entity named '"
+                                      .. self.params.name .. "'" })
+      end
+      --TODO: should foreign entity schema be included or stripped?
```
hbagdi: TODO: Discuss this point.

hbagdi:
@thibaultcha I've implemented the `/schemas/plugins/:name` endpoint, and added tests for it.

kong/api/routes/kong.lua
```diff
@@ -16,6 +17,18 @@ local tagline = "Welcome to " .. _KONG._NAME
 local version = _KONG._VERSION
 local lua_version = jit and jit.version or _VERSION
 
+
+local strip_foreign_schemas= function(fields)
```
hishamhm: typo, missing space `= `

kong/api/routes/plugins.lua
```diff
@@ -240,12 +161,15 @@ return {
 
   ["/plugins/schema/:name"] = {
     GET = function(self, db, helpers)
+      kong.log.warn("DEPRECATED: /plugins/schema/:name endpoint " ..
+                    "is deprected, please use /schemas/plugins/:name " ..
```
hishamhm: typo: deprecated

PR #4414, opened 2019-03-20
feat(db) add mutually_required and mutually_exclusive_sets

gszr:
### Summary

Add new schema validators:
- `mutually_required`: all declared fields must be set if any of them is
set
- `mutually_exclusive_sets`: if either of the fields declared is set, none of
the remaining can be set

### TODO

- ~~Naming: I'm still not 100% happy with the names; in particular, I think
argument names in the `mutually_exclusive` validator could be better.
Suggestions very welcome~~

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -9,6 +9,7 @@ local re_match     = ngx.re.match
 local re_find      = ngx.re.find
 local concat       = table.concat
 local insert       = table.insert
+local format       = string.format
 local assert       = assert
 local ipairs       = ipairs
 local pairs        = pairs
@@ -91,6 +92,8 @@ local validation_errors = {
   CONDITIONAL_AT_LEAST_ONE_OF = "at least one of these fields must be non-empty: %s",
   ONLY_ONE_OF               = "only one of these fields must be non-empty: %s",
   DISTINCT                  = "values of these fields must be distinct: %s",
+  MUTUALLY_REQUIRED         = "all or none of these fields must be set: %s",
+  MUTUALLY_EXCLUSIVE_SETS   = "these sets are mutually exclusive: %s",
   -- schema error
   SCHEMA_NO_DEFINITION      = "expected a definition table",
   SCHEMA_NO_FIELDS          = "error in schema definition: no 'fields' table",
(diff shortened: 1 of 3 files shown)


james-callahan:
Re: `mutually_required`, we currently have the pattern of 
```
    { conditional = {
      if_field = "foo", if_match = { eq = "blah" },
      then_field = "bar", then_match = { required = true },
    } },
```
https://github.com/Kong/kong/blob/d1bfdf96962959acc5e9183ad1f5a5cc05c5059b/kong/plugins/response-ratelimiting/schema.lua#L78-L81

hishamhm:
> Re: mutually_required, we currently have the pattern of

@james-callahan yes, this is slightly different because it is based on presence of the fields, and not on their values, but it would also be doable with `conditional`, indeed.

I had the idea for these validators in this PR during a call with @gszr and suggested he implement them. We were going through his use-case, where he had a bunch of interdependencies. It _would_ be doable with a lot of `conditional` validators, but it be super cumbersome (effectively the cartesian product of all possibilities).

In fact, this is again an instance of really wanting an algebraic data type (effectively a tagged union) as was @Tieske's use-case the other day. As is the bunch of `conditional`s currently in the Routes entity.

The practical design problem I'm running into when coming up with a proper solution for this is that each existing entity uses a different ad hoc discriminant for the cases (Plugin uses a field (`name`), Routes uses values in a set (`protocols`), @gszr's use-case uses the mere presence of some of the fields, @Tieske's use-case for Upstream uses a mix of both, but the `algorithm` entry makes it more similar to the Plugin case). I have some thoughts on how to make subschema selection smart enough to handle all these cases, but we might to wait until Kong 2.0 to make it happen cleanly.

In the meantime, I think these validators will do the trick. 

I suggest `mutually_exclusive_sets` because otherwise it sounds like it will do the same as `only_one_of`. As for the field names, `set1` and `set2` are ugly but descriptive. It irked me a bit too when I first looked at them, but I can't think of anything really better. (And given the technical constraints in the entity checker implementation that I'm well aware of, I know that the other obvious option of turning it into an array of arrays would not be easy — and IMO, less readable in fact).

kong/db/schema/init.lua
```diff
@@ -655,8 +655,57 @@ Schema.entity_checkers = {
...
+          insert(nonempty, name)
+        end
+      end
+
+      if #nonempty == 0 or #nonempty == #field_names then
+        return true
+      end
 
+      return nil, "All or none of " .. quoted_list(field_names) .. " must be set. "
+                  .. "Only " .. quoted_list(nonempty) .. " found"
```
hishamhm: Two very small things here:
* the error string needs to start with lowercase (`all or`...)
* please add the error string to `validation_errors` use `string.format` on that (it keeps the schema error messages consistent, and who knows, someday we might even internationalize them!)

kong/db/schema/init.lua
```diff
@@ -655,8 +655,57 @@ Schema.entity_checkers = {
...
+
+      for _, name in ipairs(args.set2) do
+        if is_nonempty(get_field(entity, name)) then
+          insert(nonempty2, name)
+        end
+      end
+
+      if #nonempty1 > 0 and #nonempty2 > 0 then
+        return nil, quoted_list(nonempty1) .. " must not be set with " ..
+                    quoted_list(nonempty2)
```
hishamhm: same here

kong/db/schema/init.lua
gszr: Done. I slightly changed the errors messages to make them consistent with other entity checkers.

kong/db/schema/init.lua
gszr: Ditto; this one in particular required me to use a string.format call to make it more readable/descriptive, which deviates from how it's done in other entity checkers. Let me know what you think.

PR #4417, opened 2019-03-21
feat(api) add entity schema validation endpoints

hbagdi:
A new endpoint is introduced on the Admin API:
`POST /schemas/:entity_name/validate`
which can be used to validate an instance of any entity type in Kong
without creating the entity.

Please note that all the schema validations are run but database
constraints like foreign key associations are not validated.

### DIFF
--- a/kong/api/routes/kong.lua
+++ b/kong/api/routes/kong.lua
@@ -3,6 +3,8 @@ local singletons = require "kong.singletons"
 local conf_loader = require "kong.conf_loader"
 local cjson = require "cjson"
 local api_helpers = require "kong.api.api_helpers"
+local Schema = require "kong.db.schema"
+local Errors = require "kong.db.errors"
 
 local sub = string.sub
 local find = string.find
@@ -11,6 +13,7 @@ local tonumber = tonumber
 local kong = kong
 local knode  = (kong and kong.node) and kong.node or
                require "kong.pdk.node".new()
+local errors = Errors.new()
 
 
 local tagline = "Welcome to " .. _KONG._NAME
@@ -143,6 +146,26 @@ return {
       local copy = api_helpers.schema_to_jsonable(schema)
       strip_foreign_schemas(copy.fields)
       return kong.response.exit(200, copy)
+    end
+  },
+  ["/schemas/:db_entity_name/validate"] = {
+    POST = function(self, db, helpers)
+      local db_entity_name = self.params.db_entity_name
+      -- What happens when db_entity_name is a field name in the schema?
+      self.params.db_entity_name = nil
+      local entity = kong.db[db_entity_name]
+      local schema = entity and entity.schema or nil
+      if not schema then
+        return kong.response.exit(404, { message = "No entity named '"
+                                  .. db_entity_name .. "'" })
+      end
+      local schema = assert(Schema.new(schema))
+      local _, err_t = schema:validate(schema:process_auto_fields(
+                                        self.params, "insert"))
+      if err_t then
+        return kong.response.exit(400, errors:schema_violation(err_t))
+      end
+      return kong.response.exit(200, { message = "entity is valid" })
     end
   },
   ["/schemas/plugins/:name"] = {
--- a/spec/02-integration/04-admin_api/02-kong_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/02-kong_routes_spec.lua
@@ -199,7 +199,7 @@ describe("Admin API - Kong routes", function()
         local json = cjson.decode(body)
         assert.same({ message = "No entity named 'not-present'" }, json)
       end)
-      it("does not return #only schema of a foreign key", function()
+      it("does not return schema of a foreign key", function()
         local res = assert(client:send {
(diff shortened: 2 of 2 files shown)


hbagdi:
The following is valid, meaning entities which must require foreign relations (like consumer.id is required here), will be valid even if one doesn't supply that relation.

I'm not sure how to handle cases like this.
My thinking is:
- if an entity is valid, it means that it satisfies the schema of an entity but it doesn't mean that it's a valid entity that is ready to be inserted into the database.
- Database constraints (like enforcing uniqueness, foreign key validation) shouldn't be checked at this endpoint.
- In the following case however, `consumer.id` is missing, which should be reported as an error.

```
http :8001/schemas/acls/validate group=sdf
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 29
Date: Thu, 21 Mar 2019 17:33:33 GMT
Server: kong/1.1.0rc2

{
    "message": "entity is valid"
}
```

hbagdi:
Something to discuss:
Should this endpoint respond back with the entity filled with default?

hbagdi:
@Kong/team-core Thoughts?

hishamhm:
> In the following case however, consumer.id is missing, which should be reported as an error.

In the schema for `consumer` in `acls` it says `default = ngx.null`, and it does not say `required = true`. Missing required fields, foreign or not, should be a regular schema error.

> Should this endpoint respond back with the entity filled with default?

I think this would be informative. But it also stresses the point that schema validations must not have side-effects. (Thinking out loud: we are thinking of doing things like allowing fields to be hashed on input, symmetric encryption, and the like. In that case we would want to make sure the input is successfully encoded and decoded before being returned — all the while causing no side-effects.)

hbagdi:
> I think this would be informative. But it also stresses the point that schema validations must not have side-effects. (Thinking out loud: we are thinking of doing things like allowing fields to be hashed on input, symmetric encryption, and the like. In that case we would want to make sure the input is successfully encoded and decoded before being returned — all the while causing no side-effects.)

Should we refrain from adding the entity in the response (and keep this PR as is) and only respond back with errors or success message for now? We can increase the API surface in future if there is an ask in future for it.

hishamhm:
> Should we refrain from adding the entity in the response (and keep this PR as is) and only respond back with errors or success message for now?

I'm all for incremental improvement. Let's just keep in mind that if we add the extra functionality in Kong 1.x we'll have to do it in a non-breaking manner (e.g. via another endpoint or by adding a query arg to the endpoint to specify that we want the entity returned — the latter sounds like a sensible approach for a future PR, IMO).

spec/02-integration/04-admin_api/02-kong_routes_spec.lua
```diff
@@ -163,4 +163,38 @@ describe("Admin API - Kong routes", function()
       end)
     end)
   end
+
+  describe("#only /schemas/:db_entity_name/validate", function()
```
hishamhm: please remove `#only`

hishamhm:
@hbagdi pending conflicts against current `next` in this branch

hbagdi:
@hishamhm Rebased and resolved the conflicts.

I added tests to check if schema checks run for plugins as well properly (which were the initial motivation for this feature) and caught a bug. `/schemas/:db_entity_name/validate` had to be moved before `/schemas/plugins/:name`, else Lapis would match the `/schemas/plugins/validate` endpoint at `/schemas/plugins/:name` and it would result it a 405.

kong/api/routes/kong.lua
```diff
@@ -11,6 +13,7 @@ local tonumber = tonumber
 local kong = kong
 local knode  = (kong and kong.node) and kong.node or
                require "kong.pdk.node".new()
+local errors = Errors.new()
```
hutchic: `Errors / errors` concerns me but I don't know what our linting rules / guidelines are. I'm guessing it's not possible to `require "kong.db.errors".new()`

hishamhm:
@dndx user-friendliness, I guess? I get the "silence means success" unixism, but maybe not everyone does. (I'd feel more reassured myself with seeing the confirmation message.)

dndx:
@hishamhm Reason I brought this up is because it appears to me that returning message to indicate only the success of operation is not a common pattern among Kong APIs, but I may be wrong.

dndx:
I guess another reason for the argument is that this additional field makes success scenario ambitious. User could just check the status code but now they will likely also check content of the `message` field. Potentially makes the logic more complicated. That is, unless we document that `message = "entity is valid"` always follows a `200` response, but that kinda defeats the purpose of this message.

kong/api/routes/kong.lua
bungle: In most of the code I think we refer to it as `err_t`

hishamhm:
> we document that message = "entity is valid" always follows a 200 response, but that kinda defeats the purpose of this message.

@dndx We can document that the status code is what matters and that message is purely informative.

hishamhm:
Merged manually, thanks!

PR #4418, opened 2019-03-21
fix(key-auth) return unauthorized on empty header

### LINKED ISSUE #4405: key-auth plugin returns error when key is set with empty value
### Summary
The `key-auth` plugin returns an error (shown below) when the `apikey` header is set with an empty value. While the fix could be to ensure the client does not pass an empty `apikey`, I feel the expected result should be either:

1. Kong resolves and continues with the `anonymous` consumer (if defined); or
2. Throw a 40X as if the `apikey` was never set

### Steps To Reproduce

1. Configure the `key-auth` plugin globally, or on a specific service/route
2. Configure a consumer & key as per plugin instructions (https://docs.konghq.com/hub/kong-inc/key-auth/)
3. Send a request to the route's URL with a present (but empty) `apikey` header

### Additional Details & Logs
Kong version: 1.0.3
Error thrown in the logs: `[kong] handler.lua:159 [key-auth] failed to get from node cache: [postgres] length must be at least 1`

hbagdi:
Fix #4405

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -143,7 +143,7 @@ local function do_authentication(conf)
   end
 
   -- this request is missing an API key, HTTP 401
-  if not key then
+  if not key or key == "" then
     kong.response.set_header("WWW-Authenticate", _realm)
     return nil, { status = 401, message = "No API key found in request" }
   end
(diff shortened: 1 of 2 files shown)


thibaultcha:
Merged with minor tweaks and a test case for empty key as querystring.

PR #4422, opened 2019-03-22
feat(rate-limiting) set TTL when incrementing counters

cb372:
### Summary

The "local" rate-limiting policy stores its counters in a shared-memory dictionary (shdict).

This PR modifies that behaviour so the counters now expire after one period length (e.g. if the counter is for requests in the current minute, the TTL is set to one minute).

There's a little more detail in my [forum post](https://discuss.konghq.com/t/what-happens-when-the-rate-limiting-counters-dict-fills-up/3052).

### Full changelog

* update code in the local policy's `increment` method to pass the `init_ttl` parameter to `shm:incr` 
* add a plugin test for this behaviour
* update the dummy shm implementation in `globalpatches` to support the `ttl` method


kong/globalpatches.lua
```diff
@@ -157,6 +166,24 @@ return function(options)
...
+        else
+          local expire_at = item.info.expire_at
+          if expire_at == nil then
+            return 0
+          else
+            local remaining = expire_at - ngx.now()
+            if remaining < 0 then
+              return nil, "not found"
+            else
+              return remaining
```
javierguerragiraldez: what will happen when `remaining == 0` ?  would it expire at the next cycle, or maybe "TTL == 0  ⇒ never expire" ?

please ensure the case is covered by a test

PR #4424, opened 2019-03-22
feat: configurable https redirects

hbagdi:
TODO:

- Break up commits to respect commit atomicity

### DIFF
--- a/kong/db/migrations/core/004_110_to_120.lua
+++ b/kong/db/migrations/core/004_110_to_120.lua
@@ -7,10 +7,20 @@ return {
       EXCEPTION WHEN UNDEFINED_COLUMN THEN
         -- Do nothing, accept existing state
       END$$;
+
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "routes" ADD "https_redirect_status_code" INTEGER;
+      EXCEPTION WHEN DUPLICATE_COLUMN THEN
+        -- Do nothing, accept existing state
+      END;
+      $$;
     ]],
   },
 
   cassandra = {
-    up = [[]],
+    up = [[
+      ALTER TABLE routes ADD https_redirect_status_code int;
+    ]],
   },
 }
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -760,9 +760,21 @@ return {
       if (protocols and protocols.https and not protocols.http and
           forwarded_proto ~= "https")
       then
-        header["connection"] = "Upgrade"
-        header["upgrade"]    = "TLS/1.2, HTTP/1.1"
-        return kong.response.exit(426, { message = "Please use HTTPS protocol" })
+        local redirect_status_code = route.https_redirect_status_code or 426
+
+        if redirect_status_code == 426 then
+          ngx.header["connection"] = "Upgrade"
+          ngx.header["upgrade"]    = "TLS/1.2, HTTP/1.1"
+          return kong.response.exit(426, { message = "Please use HTTPS protocol" })
+        end
+
+        if redirect_status_code == 301 or
+          redirect_status_code == 302 or
+          redirect_status_code == 307 or
+          redirect_status_code == 308 then
+          ngx.header["Location"] = "https://" .. forwarded_host .. var.request_uri
+          return kong.response.exit(redirect_status_code)
+        end
       end
 
       if not service then
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -74,6 +74,10 @@ return {
                            },
                          }
                        }, },
+    { https_redirect_status_code = { type = "integer",
+                                     one_of = { 426, 301, 302, 307, 308 },
+                                     default = 426, required = true,
+                                   }, },
     { regex_priority = { type = "integer", default = 0 }, },
     { strip_path     = { type = "boolean", default = true }, },
     { preserve_host  = { type = "boolean", default = false }, },
--- a/spec/01-unit/01-db/01-schema/06-routes_spec.lua
+++ b/spec/01-unit/01-db/01-schema/06-routes_spec.lua
@@ -114,6 +114,17 @@ describe("routes schema", function()
     assert.truthy(errs["protocols"])
   end)
 
+  it("invalid https_redirect_status_code produces error", function()
+
+    local route = Routes:process_auto_fields({ protocols = { "http" },
+                                               https_redirect_status_code = 404,
+                                             }, "insert")
+    local ok, errs = Routes:validate(route)
+    assert.falsy(ok)
+    assert.truthy(errs["https_redirect_status_code"])
+  end)
+
+
   it("produces defaults", function()
     local route = {
       protocols = { "http" },
@@ -133,6 +144,7 @@ describe("routes schema", function()
     assert.same(0,             route.regex_priority)
     assert.same(true,          route.strip_path)
     assert.same(false,         route.preserve_host)
+    assert.same(426,           route.https_redirect_status_code)
   end)
 
   it("validates the foreign key in entities", function()
@@ -250,13 +262,13 @@ describe("routes schema", function()
 
     -- acceptance
     it("accepts an apex '/'", function()
-      local route = {
+      local route = Routes:process_auto_fields({
         protocols = { "http" },
         service = { id = a_valid_uuid },
         methods = {},
         hosts = {},
         paths = { "/" },
-      }
+      }, "insert")
 
       local ok, err = Routes:validate(route)
       assert.is_nil(err)
@@ -264,13 +276,13 @@ describe("routes schema", function()
     end)
 
     it("accepts unreserved characters from RFC 3986", function()
-      local route = {
+      local route = Routes:process_auto_fields({
         protocols = { "http" },
         service = { id = a_valid_uuid },
         methods = {},
         hosts = {},
         paths = { "/abcd~user~2" },
-      }
+      }, "insert")
 
       local ok, err = Routes:validate(route)
       assert.is_nil(err)
@@ -281,13 +293,13 @@ describe("routes schema", function()
       local valid_paths = { "/abcd%aa%10%ff%AA%FF" }
 
       for i = 1, #valid_paths do
-        local route = {
+        local route = Routes:process_auto_fields({
           protocols = { "http" },
           service = { id = a_valid_uuid },
           methods = {},
           hosts = {},
           paths = { valid_paths[i] },
-        }
+        }, "insert")
 
         local ok, err = Routes:validate(route)
         assert.is_nil(err)
@@ -296,13 +308,13 @@ describe("routes schema", function()
     end)
 
     it("accepts trailing slash", function()
-      local route = {
+      local route = Routes:process_auto_fields({
         protocols = { "http" },
         service = { id = a_valid_uuid },
         methods = {},
         hosts = {},
         paths = { "/ovo/" },
-      }
+      }, "insert")
 
       local ok, err = Routes:validate(route)
       assert.is_nil(err)
@@ -445,13 +457,13 @@ describe("routes schema", function()
       }
 
       for i = 1, #valid_hosts do
-        local route = {
+        local route = Routes:process_auto_fields({
           protocols = { "http" },
           service = { id = a_valid_uuid },
           methods = {},
           paths = {},
           hosts = { valid_hosts[i] },
-        }
+        }, "insert")
 
         local ok, err = Routes:validate(route)
         assert.is_nil(err)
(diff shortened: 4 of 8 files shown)


p0pr0ck5:
Can we support 307 and 308 as well?

kong/db/migrations/core/004_110_to_120.lua
```diff
@@ -0,0 +1,39 @@
+return {
+  postgres = {
+    up = [[
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "routes" ADD "https_redirect_status_code" INTEGER DEFAULT 426;
```
thibaultcha: It is known that altering a table by adding a new column with a default value will lock the table (if it is large enough, maybe for a significant amount of time). There are many resources describing the issue online, you probably have heard of it yourself before. http://www.joshuakehn.com/2017/9/9/postgresql-alter-table-and-long-transactions.html

Considering the potential size of the routes table (up to a few thousand rows), combined with our B/G migrations path, it seems we should give some extra attention to a possible scenario, in which new green nodes are starting up and attempting to query the routes table (to build their router) at the same time the physical rewrite is still taking place in Postgres.
Another argument can be made in favour of writing referential code in our core migrations, since we have learned a long time ago that core patterns tend to spread in tangential code (e.g. plugins). It is not unrealistic to expect future similar migrations (adding new columns) to other tables, potentially much larger ones (e.g. oauth2 access tokens). Writing best-practices migrations for core could help preventing issue down the road.

hishamhm: @hbagdi If you set declare a field in the schema with a `default` value AND `required = true` (to forbid `null` as a valid value), the schema library will fill in the default for you with no need to set it in the migration.

hbagdi: > If you set declare a field in the schema with a default value AND required = true (to forbid null as a valid value), the schema library will fill in the default for you with no need to set it in the migration.

Is this something we have done in the past and is it safe to do so?
Imagine we take the approach, now, the value will not be set in the database until there is an update operation for routes. And if the default value changes (unlikely in this case, but for sake of broader discussion), then the value of field will change across upgrades, assuming the user has not updated the route between migrations, right?

hbagdi: > Considering the potential size of the routes table (up to a few thousand rows), combined with our B/G migrations path, it seems we should give some extra attention to a possible scenario, in which new green nodes are starting up and attempting to query the routes table (to build their router) at the same time the physical rewrite is still taking place in Postgres.

I did consider this problem but considering there would be a few thousand routes or tens of thousand entries, this operation would be fairly quick on any modern hardware. I do understand the concern around how this pattern can (and will) spread and we should do something about it.

@thibaultcha, what do you propose in that case? I'm thinking of writing a migration similar to the Cassandra migration where we update the rows individually.

thibaultcha: @hbagdi My thinking was to set the default on Lua-land (e.g. in the proxy code itself `ngx.status = route.https_redirect_status_code or 426`. I _seem_ to recall that we followed the schema approach suggested by @hishamhm before, but I can't remember where. It's a good approach too imho, in this case I don't see the default value changing anytime soon, and we can worry about the default value changing if the issue ever arises. Most likely though, the "old default" value will still be accepted (backwards-compatibility), and if the "new default" changed, it's probably fine as per the new code. Now up to us to instruct users on how to best handle blue nodes, which we'll assess when and if the time comes.

hbagdi: > My thinking was to set the default on Lua-land (e.g. in the proxy code itself ngx.status = route.https_redirect_status_code or 426

Right, so I already added this piece to handle the C* case because v1.2 nodes would come up and start before we finish writing the default value in the column and the new column would be nil on v1.2 nodes. This was before @hishamhm's comment on using `default` along with `required`.

Here is what we could do:
1. We add the column in both Postgres (no default value) and Cassandra during migration.
2. We do not write the default value to the DB during the `teardown` process (I'm not sure if I'm implying this correctly from the above comment or not, so please let me know if incorrect).
3. We add `required = true` in the schema so that the it is always filled in with the default value.

hishamhm: @hbagdi :+1: yes, that's the approach we've been taking in Core with this kind of migration (I also don't remember where exactly we did this, it was most likely in some plugin — but it was definitely something that was part of a conversation on B/G-friendly migration patterns). It avoids having to hardcode the default in proxy code and keeps the schema as source of truth. (Plus, note that letting the schema fill the default is no better or worse than doing it in Lua-land as far as "what if the defaults change" is concerned.)

hbagdi:
@hishamhm @thibaultcha I've updated the PR with the migration as we discussed.

thibaultcha:
@hbagdi Hi! Could you rebase your branch on top of the latest upstream's `next`? It currently has a few conflicts due to recently merged PRs. Thanks!

hbagdi:
@thibaultcha Done.

kong/runloop/handler.lua
```diff
@@ -760,9 +760,20 @@ return {
...
-        header["upgrade"]    = "TLS/1.2, HTTP/1.1"
-        return kong.response.exit(426, { message = "Please use HTTPS protocol" })
+        local redirect_status_code = route.https_redirect_status_code or 426
+
+        if redirect_status_code == 426 then
+          ngx.header["connection"] = "Upgrade"
+          ngx.header["upgrade"]    = "TLS/1.2, HTTP/1.1"
+          return kong.response.exit(426, { message = "Please use HTTPS protocol" })
+
+        elseif redirect_status_code == 301 or
```
thibaultcha: style: avoid `elseif` if the previous branch returns

kong/runloop/handler.lua
```diff
@@ -760,9 +760,20 @@ return {
...
+          redirect_status_code == 302 or
+          redirect_status_code == 307 or
+          redirect_status_code == 308 then
+          ngx.header["Location"] = "https://" .. forwarded_host .. var.request_uri
```
thibaultcha: Do we not handle non-standard ports? I.e. != 80 for http and != 443 for HTTPs.

hbagdi: I'm not sure what you mean here.

Are you saying that the configuration should be able to redirect to HTTPS on a specific port rather than the default 443?

thibaultcha: > Are you saying that the configuration should be able to redirect to HTTPS on a specific port rather than the default 443?

Yes, if Kong's ssl proxy directive is listening on non-443 (or a previous proxy, as indicated by the forwarded port header). No?

hbagdi: > Yes, if Kong's ssl proxy directive is listening on non-443 (or a previous proxy, as indicated by the forwarded port header). No?

It is possible that the proxy in-front of Kong is not the only proxy between the real client and Kong, in which case this might break, depending on if the proxy in front of Kong respects the forwarded port or not.  

Most common use case here is automated browser and clients which follow automated redirects. I have not seen any implementations which support or any issues pointing to the need for it yet (and if we have those, we can add it right now). 

We can add it if there comes an actual need for this in future.

thibaultcha: > depending on if the proxy in front of Kong respects the forwarded port or not.

Well, we know that is should respect it, so let's make the implementation safe for proxies that respect the specs, at least. If they don't, then it is not Kong's issue, but those proxies.

Currently this is broken not only for those proxies that respect the specifications, but also if Kong is listening on non-443 ports for HTTPS connections. It's a small change to make (along with an integration test) and I don't see a reason for not adding it _before_ shipping the feature, and having to come back to this further down the road.

hbagdi: Okay, I went to see if this could be implemented and then I recalled having another config property (`https_redirect_port`) for this.

What happens is following: Client talks to LB or Kong via HTTP and then based on configuration, Kong decides to redirect.

Now, at this point, Kong is aware of the forwarded HTTP port that the LB is using or Kong is using but not about the SSL port.
To overcome this problem, in the beginning, I had introduced another property in the Route entity, which dictated the port to use for redirection.
I dropped that field as it seemed exposing too much configuration to the end user.

> Currently this is broken not only for those proxies that respect the specifications, but also if Kong is listening on non-443 ports for HTTPS connections.

We can use Kong's SSL port listener but then, we run into other problems:
- how do we handle the case when Kong doesn't have an SSL port and LB is doing TLS termination
- how do we handle the case when Kong has an ssl port of 8443 and the LB has 443 or vice-versa.

thibaultcha: > What happens is following: Client talks to LB or Kong via HTTP and then based on configuration, Kong decides to redirect.

That is the tricky case. In this case, I think it would be acceptable to assume that the LB should be sending the redirect on its own, before the request gets proxied through Kong in the first place. If we don't assume that then the property you suggested is relevant.

>     * how do we handle the case when Kong doesn't have an SSL port and LB is doing TLS termination

If Kong does not accept HTTPS connections but a Route is configured to be `https` only, that is certainly a larger issue that warrants warning logs and/or additional steps during initialization/router rebuilding. That said, it feels outside the immediate scope of this PR. Secondly, if the LB is doing TLS termination, then the requests to Kong will always be HTTP by definition, so this branch should not even be entered in the first place.

>     * how do we handle the case when Kong has an ssl port of 8443 and the LB has 443 or vice-versa.

In such cases, the terminating LB needs to set the forwarded port header, and Kong will consider it in priority when building the Location URL. In the "vice-versa" case, then Kong is the terminating LB, and any port accepting HTTPS should be fine (at least in Gateway mode), as long as there exists one.

hbagdi: > If Kong does not accept HTTPS connections but a Route is configured to be https only, that is certainly a larger issue that warrants warning logs and/or additional steps during initialization/router rebuilding.

I might be wrong here but is it not the case that Kong will respect the forwarded proto header which will be `HTTP` or `HTTPS` and Kong will respect that while matching the route? So, it can happen that Kong routes traffic based on HTTP or HTTPS even if it itself is not configured to listen on a SSL port.


> In such cases, the terminating LB needs to set the forwarded port header, and Kong will consider it in priority when building the Location URL.

That is not possible because we want to use a different port in the redirect, right?
Let's say we have the following configuration:

LB (port 80 HTTP) -> Kong (port 8000 HTTP)
LB (port 443 HTTP) -> Kong (port 8443 HTTPS)

In such a case, when an HTTP request comes to LB and then to Kong, the forwarded port will be set to `80`. In case Kong decides to do a HTTPS redirect, it will want to use the SSL port of the LB, and not the HTTP port of the LB but Kong has no knowledge of that port while processing the request.

thibaultcha: > I might be wrong here but is it not the case that Kong will respect the forwarded proto header which will be HTTP or HTTPS and Kong will respect that while matching the route? So, it can happen that Kong routes traffic based on HTTP or HTTPS even if it itself is not configured to listen on a SSL port.

Correct! For trusted IPs at least... If none is configured (which is the case by default), then Kong will never be able to proxy those HTTPS-only Routes.

> In such a case, when an HTTP request comes to LB and then to Kong, the forwarded port will be set to 80. In case Kong decides to do a HTTPS redirect, it will want to use the SSL port of the LB, and not the HTTP port of the LB but Kong has no knowledge of that port while processing the request.

That is the case that warrants either assuming that the LB will perform the redirect on its own, or makes the case for your new `https_redirect_port` attribute (I wanted to highlight this in my previous comment but it got removed during editing I think).

We still don't cover the following case:

- `Client -> Kong (8000 HTTP + 8443 HTTPS)` need to redirect to `8443`.

Other cases I could think of:

- `Client -> LB (443) -> Kong (8000 HTTP, trusted)` LB forwarded proto, so the request will be considered "HTTPS".
- `Client -> LB (80) -> Kong (8000 HTTP, trusted)` LB should redirect, or Kong should have prior knowledge of the LB HTTPS port.
- `Client -> LB (443) -> Kong (8000 HTTP, not trusted)` LB forwarded proto, but isn't trusted. Misconfiguration.
- `Client -> LB (80) -> Kong (8000 HTTP, not trusted)` LB should redirect, or Kong should have prior knowledge of the LB HTTPS port.

thibaultcha: As mentioned during our brief f2f chat:

- You made the point that if a TCP LB is sitting in front of Kong, it becomes hard to determine whether Kong is the first hop in the server chain or not, and thus to smartly advertise a port to the client.
- On the `https_redirect_port` property: unless asked for, we feel it is outside of the scope of this PR and whether it will become needed at all or not. 

So for now, let's make the assumption that in order for HTTPS redirect to be effective, users must configure Kong to listen for HTTPS traffic on the standard 443 port. Once other minor concerns have been addressed here, let's merge this :+1:

thibaultcha:
Thanks, big deal patch!

PR #4429, opened 2019-03-25
fix(dao) correctly manage primary keys of type "foreign" 

kidd:
This PR allows primary keys to be foreign keys (meaning there's a composite pk)

It assumes some conventions (maybe too many of them) as the field names and suffixes added for foreign keys. I'm happy to improve it if needed but don't want to step on other's toes.

### DIFF
--- a/kong/db/strategies/cassandra/init.lua
+++ b/kong/db/strategies/cassandra/init.lua
@@ -1041,7 +1041,7 @@ do
           if err then
             return nil, self.errors:database_error("could not gather " ..
                                                    "associated entities " ..
-                                                   "for delete cascade: ", err)
+                                                   "for delete cascade: " .. err)
           end
 
           local row_pk = constraint.schema:extract_pk_values(row)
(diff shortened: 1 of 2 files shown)


kong/db/strategies/cassandra/init.lua
```diff
@@ -280,6 +286,9 @@ local function serialize_arg(field, arg)
   elseif field.type == "record" then
     serialized_arg = cassandra.text(cjson.encode(arg))
 
+  elseif field.type == "foreign" then
+    serialized_arg = cassandra.uuid(arg[field.schema.primary_key[1]])
```
hishamhm: This is still assuming that the key will always be of type `uuid`. I suggest resolving the field and recursing this function instead (untested)

```lua
local fk_pk = field.schema.primary_key[1]
local fk_field = field.schema.fields[pk]
serialized_arg = serialize_arg(fk_field, arg[fk_pk])
```

kong/db/strategies/cassandra/init.lua
```diff
@@ -226,6 +230,8 @@ local function get_query(self, query_name)
 end
 
 
+local serialize_foreign_pk
```
hishamhm: This forward declaration looks unnecessary (it would be useful if we were supporting composite FKs in PKs, but as it stands it looks like a leftover).

kong/db/strategies/cassandra/init.lua
kidd: changed accordingly

PR #4430, opened 2019-03-25
fix(schema) change timeout to be at least 1ms

hbagdi:
Timeouts in Service objects can be currently set to 0, but the proxy
calls then return 500s as OpenResty API doesn't allow for zero or
negative timeouts:
https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/balancer.md#set_timeouts

`typedefs.timeout` is not currently used anywhere except for defining
timeouts in the `Service` entity.

A `0` value for a timeout is a valid, and for such a use-case, a user
can define their own type value using the Schema library.

### DIFF
--- a/kong/db/schema/entities/services.lua
+++ b/kong/db/schema/entities/services.lua
@@ -1,4 +1,11 @@
 local typedefs = require "kong.db.schema.typedefs"
+local Schema   = require("kong.db.schema")
+
+
+local nonzero_timeout = Schema.define {
+  type = "integer",
+  between = { 1, math.pow(2, 31) - 2 },
+}
 
 
 return {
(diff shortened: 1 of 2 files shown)


hbagdi:
I thought about this needing a migration for cases where this is already set to 0 but skipped it since Kong won't be working for those services in the first place so it's unlikely there are such configurations out there.

kong/db/schema/typedefs.lua
```diff
@@ -214,7 +214,7 @@ typedefs.header_name = Schema.define {
 
 typedefs.timeout = Schema.define {
   type = "integer",
-  between = { 0, math.pow(2, 31) - 2 },
+  between = { 1, math.pow(2, 31) - 2 },
```
thibaultcha: This is considered part of the public API of 1.0 (even though unfortunately not documented yet). We may not know, but third-party plugins could be relying on this behaviour. Some of our external plugins (bundled at packaging time) may also make use of this typedef (worth checking...).

Most likely, as you were saying, a value of `0` for a timeout is acceptable. As such, why not making that the general use-case (and keep `0` here, not only for backwards-compatibility, but also commodity), and update the timeout fields of the Service entity to _not_ accept `0`?

hishamhm: > Some of our external plugins (bundled at packaging time) may also make use of this typedef (worth checking...).

I checked, none are.

> Most likely, as you were saying, a value of 0 for a timeout is acceptable. As such, why not making that the general use-case (and keep 0 here, not only for backwards-compatibility, but also commodity), and update the timeout fields of the Service entity to not accept 0?

I would accept the PR as is, but I'm fine with this option too.

thibaultcha: I don't see why we wouldn't follow the above approach. Timeouts in general can have values of `0`, it just so happens that the timeout attributes of this particular Service entity cannot. Plus as I said, we preserve backwards-compatibility for any use-case out there we may not be aware of.

hbagdi: I've updated the PR to avoid the breaking change. Please re-review, thanks!

PR #4436, opened 2019-03-26
chore(make) adds a new Make target to install dev tools only

Tieske:
The current 'make dev' reinstalls Kong and the dev dependencies.
The new 'make dev-only' only installs dev dependencies without
touching the Kong installation itself.
This enables plugin testing without Kong source code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,8 @@ else
 OPENSSL_DIR ?= /usr
 endif
 
-.PHONY: install dev lint test test-integration test-plugins test-all fix-windows
+.PHONY: install remove dependencies dev \
+	lint test test-integration test-plugins test-all fix-windows
 
 KONG_GMP_VERSION ?= `grep KONG_GMP_VERSION .requirements | awk -F"=" '{print $$2}'`
 RESTY_VERSION ?= `grep RESTY_VERSION .requirements | awk -F"=" '{print $$2}'`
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,9 +53,12 @@ release: setup-release
 install:
 	@luarocks make OPENSSL_DIR=$(OPENSSL_DIR) CRYPTO_DIR=$(OPENSSL_DIR)
 
-dev:
+remove:
 	-@luarocks remove kong
-	@luarocks make OPENSSL_DIR=$(OPENSSL_DIR) CRYPTO_DIR=$(OPENSSL_DIR)
+
+reinstall: remove install
```
kikito: The PR and commit messages seem to indicate that they add a single Make target, but it seems they more than one: `make dev-only`, `make remove` & `make reinstall`.

Makefile
```diff
@@ -52,9 +53,12 @@ release: setup-release
...
+
+dev-only:
```
kikito: I think having `make dev` and `make dev-only` would be very confusing. I suggest renaming `dev-only` to `dependencies`.

Makefile
```diff
@@ -64,6 +68,8 @@ dev:
 	  fi \
 	done;
 
+dev: reinstall dev-only
```
kikito: It seems `reinstall` is only used by this `dev` target. I suggest removing `reinstall` completely and just using the two targets manually. This coupled with the rename of `dev-only` would give:
```
dev: remove install dependencies
```

PR #4439, opened 2019-03-26
[wip] feat(conf) add prometheus_listen

hbagdi:
Prometheus plugin exposes metrics related to Kong at an HTTP endpoint.
If Kong's Admin API is disabled, metrics cannot be exposed unless a
custom nginx conf is used.

An alternative solution exists where users can include an NGINX server
block using Kong's directive injection mechanism but that has proved to
be not user-friendly.

This commit adds support to setup a Prometheus scrape endpoint in Kong
natively.


hbagdi:
I've discussed a more generic solution that solves this problem not just for Prometheus but also for other plugins that might leverage a separate endpoint.
I will work on the feature when I can find time and open a PR.
Closing this one.

PR #4441, opened 2019-03-26
fix(error-handler) removing the ternary operator used by message as redundant

ms2008:
### Summary

This ternary operator is unnecessary and can be removed.


thibaultcha:
@ms2008 Thank you!

PR #4450, opened 2019-03-28
fix(migration) fix an issue in core migration that resulted in bad sc…

dndx:
…hema after migrating C*

backed Kong cluster from 0.14 to 1.1.0.

In `kong/db/migrations/core/003_100_to_110.lua`, we added new fields `tags` and `protocols`
to the `plugins` table in the `up` phase, however, the `plugins` table
was recreated in `kong/db/migrations/core/001_14_to_15.lua`'s `teardown`
phase without those two fields. This results green clusters to throw HTTP
500 error for all proxy requests after `kong migrations finish` is
executed.

Thanks @thibaultcha for helping on determining the fix.

### DIFF
--- a/kong/db/migrations/core/001_14_to_15.lua
+++ b/kong/db/migrations/core/001_14_to_15.lua
@@ -338,6 +338,8 @@ return {
           config text, -- serialized plugin configuration
           enabled boolean,
           cache_key text,
+          protocols set<text>, -- added in 1.1.0
+          tags set<text>, -- added in 1.1.0
           PRIMARY KEY (id)
         );
 


thibaultcha:
Thanks for the patch!

A few notes for the reader and future selves:
- We typically consider editing existing migrations bad practises, but not in this case, since we are in reality working around a limitation of the `copy_cassandra_records()` helper, and fixing an issue with the migration's code itself (and not the result of its DDL statements).
- We could have also decided to not support the 0.14 -> 1.1 upgrade path, but we decided to keep supporting it considering the fix is minimal.
- We must keep in mind that Admin API writes to green nodes are still not supported (as documented in our Upgrade Path). As pointed out by @dndx, such writes in between the two steps migrations will produce HTTP 500 errors. We should ideally implement the "migration state mode" preventing such writes in a more user-friendly fashion.
- The same issue may reappear in the future. To overcome it, we must make the `copy_cassandra_records()` helper smarter and not rely on a hard-coded schema in the migration itself, but able to interpret the schema as it lives in the keyspace, at time of migration, and recreate it in its temporary, and new tables.

:clap:

thibaultcha:
Note: waiting on a new run of the migrations tests from @dndx (and CI for this PR) before merging. This will be part of a 1.1.1 hotfix release targeted for 0.14 users with a Cassandra backend migrating directly to 1.1.

dndx:
@thibaultcha The upgrade tests are successful:

| Migration path        | Postgres | C* |
|-----------------------|:--------:|:--:|
| 1.0.0 -> this branch  |     ✅    |  ✅ |
| 0.14.1 -> this branch |     ✅    |  ✅ |

thibaultcha:
Thank you!

PR #4451, opened 2019-03-29
kong config init should work without a prefix set up

james-callahan:
`kong config init` should work without running `kong prepare` or `kong start && kong stop` first.

I added a 2nd commit here to allow specifying a filename, as it kept causing me issues when running kong from a read-only directory. In future I think we should add some convenience to sending it to stdout (makes it much easier to use with docker). Feel free to cherry-pick just the hotfix if you want.

### DIFF
--- a/kong/cmd/config.lua
+++ b/kong/cmd/config.lua
@@ -29,6 +29,11 @@ end
 
 
 local function execute(args)
+  if args.command == "init" then
+    generate_init()
+    os.exit(0)
+  end
+
   log.disable()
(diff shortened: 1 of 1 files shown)


thibaultcha:
Fix cherry picked to master for 1.1.1, thank you! I'm leaving this open as you'll probably want to rebase the feature commit on top of master.

hishamhm:
@james-callahan please rebase the `fix` commit alone on top of `next`

hishamhm:
Superseded by #5104

PR #4454, opened 2019-03-29
chore(tests) enable plugin specific nginx server/stream mocks

Tieske:
The mocks are injected into the nginx conf and rendered with the
Kong configuration. The Lambda plugin has been converted to this
format.

Please check commit scope/type, as I wasn't sure

### DIFF
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -349,6 +349,7 @@ end
 
 return {
   prepare_prefix = prepare_prefix,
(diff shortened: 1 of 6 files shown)


hishamhm:
I like this change! `custom_nginx.template` keeps growing, it's getting unwieldy.

> Please check commit scope/type, as I wasn't sure

I would name them `tests(dns)` and `tests(*)`

Also, if you rebase the PR on the latest `master` those PDK flaky tests should go away.

Tieske:
@hishamhm all updated!

PR #4455, opened 2019-03-29
docs(conf) document database=off and declarative_config

hishamhm:
Add missing documentation in `kong.conf.default` for features introduced in Kong 1.1.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -400,19 +400,30 @@
 # DATASTORE
 #------------------------------------------------------------------------------
 
-# Kong will store all of its data (such as Routes, Services, Consumers, and Plugins) in
-# either Cassandra or PostgreSQL, and all Kong nodes belonging to the same cluster
-# must connect themselves to the same database.
+# Kong can run with a database to store coordinated data between Kong nodes in
+# a cluster, or without a database, where each node stores its information
+# independently in memory.
+#
+# When using a database, Kong will store data for all its entities (such as
+# Routes, Services, Consumers, and Plugins) in either Cassandra or PostgreSQL,
+# and all Kong nodes belonging to the same cluster must connect themselves
+# to the same database.
 #
 # Kong supports the following database versions:
 # - **PostgreSQL**: 9.5 and above.
 # - **Cassandra**: 2.2 and above.
 #
+# When not using a database, Kong is said to be in "DB-less mode": it will keep
+# its entities in memory, and each node needs to have this data entered via a
+# declarative configuration file, which can be specified through the
+# `declarative_config` property, or via the Admin API using the `/config`
+# endpoint.
+#
 
 #database = postgres             # Determines which of PostgreSQL or Cassandra
                                  # this node will use as its datastore.
-                                 # Accepted values are `postgres` and
-                                 # `cassandra`.
+                                 # Accepted values are `postgres`,
+                                 # `cassandra`, and `off`.
 
 #pg_host = 127.0.0.1             # Host of the Postgres server.
 #pg_port = 5432                  # Port of the Postgres server.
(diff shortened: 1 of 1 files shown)


kong.conf.default
```diff
@@ -400,19 +400,30 @@
...
+# `declarative_config` directive, or via the Admin API using the `/config`
+# endpoint.
+#
 
 #database = postgres             # Determines which of PostgreSQL or Cassandra
                                  # this node will use as its datastore.
-                                 # Accepted values are `postgres` and
-                                 # `cassandra`.
+                                 # Accepted values are `postgres`,
+                                 # `cassandra` or `off`.
```
thibaultcha: a comma before `or` would read better

kong.conf.default
```diff
@@ -521,6 +532,17 @@
...
+#declarative_config =           # The path to the declarative configuration
+                                # file which holds the specification of all
+                                # entities (Routes, Services, Consumers, etc.)
+                                # to be used when Kong runs without a database.
+                                #
+                                # Entities are stored in the cache, so it
+                                # must have enough space to store all entities
+                                # (see the `mem_cache_size` directive),
+                                # and should be configured to never expire
+                                # (i.e. `db_cache_ttl` should be kept at 0).
```
thibaultcha: "it" sort of lacks context here. Maybe a better way to phrase this:
```
Entities are stored in Kong's in-memory cache, so you must
ensure that enough memory is allocated to it via the
`mem_cache_size` property. You must also ensure that items
in the cache never expire, which means that `db_cache_ttl`
should preserve its default value of 0.
```

Note that throughout our entire documentation, we refer to Kong's config properties as "properties", while using "directives" for Nginx directives only.

kong.conf.default
```diff
@@ -521,6 +532,17 @@
                                              # This value is only used during
                                              # migrations.
 
+#declarative_config =           # The path to the declarative configuration
+                                # file which holds the specification of all
+                                # entities (Routes, Services, Consumers, etc.)
+                                # to be used when Kong runs without a database.
```
thibaultcha: Should we specify that `database=off` should be in effect for this property to have a meaningful impact?

kong.conf.default
hishamhm: That works too!

> Note that throughout our entire documentation, we refer to Kong's config properties as "properties", while using "directives" for Nginx directives only.

Funny, I _did_ look it up to see how other attributes were being called, and my first hit was seeing `log_level` (which is a `kong.conf` property) being called a "directive" so I went with that.

thibaultcha: Ugh... We should change it then!

kong.conf.default
hishamhm: ah yes, I forgot that the doc style guide now enforces Oxford comma. Fixing it!

(As to whether it reads better... let's just skip that debate :P — as they say, it's tabs vs spaces for English :) )

kong.conf.default
hishamhm: I thought about it, but we don't do it for the Postgres or Cassandra options (though those are more obviously prefixed, granted). It does spell out "when Kong runs without a database", though, so my take was that it would feel a bit redundant.

kong.conf.default
hishamhm: Can do. Should all the "See <nginx URL> for a description of this directive" be changed as well? The "this" there seems to refer to the property (pedantically speaking the most technically correct would be "See <nginx URL> for a description of the Nginx directive exposed by this property")

kong.conf.default
thibaultcha: The feature is relatively new, and users may not even be aware of how Kong needs to be configured to run without a database yet. Or rather, any new user even in the future may not be immediately aware that Kong can run without a database, and how. I wouldn't call it redundant if we reminded users how to do so here, especially since it is a requirement for this property to take effect. In other properties (including the other two edited in this patch) we aren't shying away from mentioning sibling properties (e.g. "ensure the in-memory cache is large enough via the `mem_cache_size` property).

kong.conf.default
hishamhm: Scratch that, these entries mention "directive of the same name" above — my bad.

kong.conf.default
hishamhm: :+1:

PR #4457, opened 2019-03-29
feat(ssl) support for wildcard SNI matching for dynamic certificates

thibaultcha:
Our `ssl_certificate_by_lua` phase (and stream preread) is now able to
match an SNI against any registered wildcard SNI. This is helpful for
users who maintain many subdomains and currently have to create an SNI
entity for each of them (sometimes several dozens all serving the same
certificate).

Wildcard SNIs can have up to one leftmost or rightmost wildcard. They
can be created like so (via Admin API):

    curl -XPOST http://localhost:8001/snis \
        -d name='*.mydomain.com' \
        -d certificate.id='<cert_id>'

The SNI will be matched (and its associated certificate, served) when
the client provides it. E.g.:

    curl -k -v \
        --resolve www.mydomain.com:8443:127.0.0.1 \
        https://www.mydomain.com:8443

In the above request, the SNI will be `www.mydomain.com`.

The following wildcard SNIs are considered valid and supported:

- mydomain.*
- *.mydomain.com
- *.www.mydomain.com


kong/runloop/certificate.lua
```diff
@@ -4,91 +4,139 @@ local http_tls = require "http.tls"
...
+    -- *.example.com lookup
+    local wild_sni, n, err = re_sub(sni, [[([^.]+)(\.[^.]+\.\S+)]], "*$2",
+                                    "ajo")
+    if err then
+      log(ERR, "could not create SNI wildcard for SNI lookup: ", err)
+
+    elseif n > 0 then
+      wild_prefix_sni = wild_sni
+    end
   end
```
locao: I believe the remaining code in this function could be moved to inside this block, except the `return` statement.
```suggestion

    -- example.* lookup
    wild_sni, n, err = re_sub(sni, [[([^.]+\.)([^.]+)$]], "$1*", "jo")
    if err then
      log(ERR, "could not create SNI wildcard for SNI lookup: ", err)

    elseif n > 0 then
      wild_suffix_sni = wild_sni
    end

  elseif wild_idx == #sni then
    wild_suffix_sni = sni
  end
```

spec/01-unit/01-db/01-schema/11-snis_spec.lua
```diff
@@ -0,0 +1,90 @@
...
+
+      for _, name in ipairs(names) do
+        local ok, err = validate({ name = name, certificate = certificate })
+        assert.is_nil(err)
+        assert.is_true(ok)
+      end
+    end)
+
+    it("rejects wrong wildcard placements", function()
+      local names = { "foo.*.com", "foo.*.wildcard.com" }
```
locao: Just a few more cases:
```suggestion
      local names = { "foo.*.com", "foo.*.wildcard.com", "*.example.*",
                      "*.foo.*.wildcard.com" }
```

spec/01-unit/01-db/01-schema/11-snis_spec.lua
```diff
@@ -0,0 +1,90 @@
...
+
+      for _, name in ipairs(names) do
+        local ok, err = validate({ name = name, certificate = certificate })
+        assert.is_nil(ok)
+        assert.same({ name = "wildcard must be leftmost or rightmost character" }, err)
```
locao: Changes needed for the added cases:
```suggestion
        assert.is_true(err.name == "wildcard must be leftmost or rightmost character"
                       or err.name == "only one wildcard must be specified")
```

kong/runloop/certificate.lua
thibaultcha: Handling those cases separately draws a nicer distinction between leftmost SNI and rightmost SNI. Additionally, we explicitly avoid to nest our code in order to preserve readability (we favor early-exit patterns) and 80 chars limit on column length (as per our contribution and style guidelines). The code was written with both intentents mind.

spec/01-unit/01-db/01-schema/11-snis_spec.lua
thibaultcha: Nice catch. Although this case is tested in the Admin API integration suite, it is nice to have a unit test for the model as well.

guanlan:
@locao Does this ready for merge?
cc @hishamhm

locao:
Yes, it is.

thibaultcha:
I'd rather split the added test scenarios in their own case, please hold. I will go ahead and merge this then.

thibaultcha:
I also found another sanitization case that is not properly covered.

premaind:
Hi @thibaultcha,
Please let us know whether this feature is available in Enterprise Version of Kong ? If so, can you help with the version number of Kong where SNI wildcards are supported.

PR #4459, opened 2019-04-01
feat(Kongsumers)

StewartW:

### Summary

Changing from consumer to **kong**sumer

### Full changelog

* Changed all references of consumer to kongsumer

### Issues resolved

Fixes a missed opportunity for a great pun.

Happy April Fools <3 Skyscanner


Tieske:
Thank you for your contributions. Unfortunately they are missing migrations.

Since this is a hot debate, I think we'd like to see two branches for further discussion; one for migrations in the no-deal Brexit and one for the softer alternative. Please mind that in both cases there are edge-cases since Kongsumers from Mexico are not allowed to migrate to the US. Likewise any North-African Kongsumers should at a minimum be equipped with life-vests for the whole migration.

PR #4462, opened 2019-04-02
POC - [do not merge] - feat(cache) prewarming cache exhausted check plus cache monitoring

bungle:
### Summary

This PR implements three things:

1. cache prewarming exhausted check
   - writes log entry and skips rest of the warming

2. declarative config cache exhausted check
   - returns early in such case with an error

3. cache monitoring
   - runs every minute and writes to error log information
     about cache usage when it has grows over 60% percent
     (with different log levels: debug, info, notice, warn)

### DIFF
--- a/kong/cache.lua
+++ b/kong/cache.lua
@@ -1,16 +1,31 @@
-local resty_mlcache = require "resty.mlcache"
+local resty_mlcache    = require "resty.mlcache"
 
 
-local type    = type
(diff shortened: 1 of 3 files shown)


bungle:
I'll close this as `monitor`ing was a bit controversial, and #4468 already does pre-warm cache filled check more reliably, and the declarative check will follow.

PR #4468, opened 2019-04-03
feat(cache) prewarming cache exhausted check

bungle:
### Summary

This PR implements cache prewarming exhausted check:

- writes log entry and skips rest of the pre-warming in case the cache
  memory is exhausted

### DIFF
--- a/kong/cache.lua
+++ b/kong/cache.lua
@@ -11,6 +11,8 @@ local DEBUG   = ngx.DEBUG
 
 
 local SHM_CACHE = "kong_db_cache"
+
+
 --[[
 Hypothesis
 ----------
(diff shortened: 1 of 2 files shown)


kong/runloop/cache_prewarm.lua
```diff
@@ -52,6 +64,13 @@ function cache_prewarm.execute()
   for _, entity_name in ipairs(ENTITIES_TO_PREWARM) do
     local ok, err = cache_prewarm_single_entity(entity_name)
     if not ok then
+      if err == "no memory" then
+        kong.log.warn("cache memory is exhausted, please consider raising the ",
+                      "'mem_cache_size=", kong.configuration.mem_cache_size, "'")
```
hishamhm: The message reads a bit odd. I suggest:

```lua
  kong.log.warn("cache memory is exhausted, please consider increasing ",
                "the value of 'mem_cache_size' (currently at ",
                kong.configuration.mem_cache_size, ")")

Comment (bungle) [reply kong/runloop/cache_prewarm.lua]: changed.

File: kong/cache.lua [hunk 2]
```diff
@@ -132,6 +209,14 @@ function _M:get(key, opts, cb, ...)
 end
 
 
+function _M:safe_set(key, value)
+  local str_marshalled = marshall_for_shm(value, self.mlcache.ttl,
```
thibaultcha: This is missing error checks

kong/runloop/cache_prewarm.lua
```diff
@@ -23,6 +29,12 @@ local function cache_prewarm_single_entity(entity_name)
     end
 
     local cache_key = dao:cache_key(entity)
+
+    local ok, err = kong.cache:safe_set(cache_key, entity)
+    if not ok then
+      return nil, err
+    end
+
```
thibaultcha: Now that we are directly setting those values in L2, there is no point in calling `:get()` like we do below anymore. Unless I missed something?

kong/runloop/cache_prewarm.lua
```diff
@@ -52,6 +64,14 @@ function cache_prewarm.execute()
   for _, entity_name in ipairs(ENTITIES_TO_PREWARM) do
     local ok, err = cache_prewarm_single_entity(entity_name)
     if not ok then
+      if err == "no memory" then
+        kong.log.warn("cache memory is exhausted, please consider increasing ",
```
thibaultcha: It would have been helpful to explain that this happened during cache warming, and maybe explain what the implications of this are. It's important users don't get mislead into believing that a full cache could mean that Kong would fail to serve requests or any such thing...

kong/runloop/cache_prewarm.lua
bungle: It will warm L1, but perhaps there is `lock` involved? Is there? That may cause problems in `init_worker`?

thibaultcha: I don't see any benefits in warming up L1 here. There would COW benefits be if this was being executed in `init`, but that isn't the case as-is. This will likely slow down startup due to the worker mutexes.

bungle: @thibaultcha is there a mutex when reading value from L2? I am pretty sure there will not. The callback is never called. I changed return entity to error('should never be called').

bungle: here: https://github.com/Kong/kong/pull/4408/commits/6085ac2065eb7817c4c5b8847d168cc105a3d236#diff-f88344d1a956d42966b81639f8302e26R39

thibaultcha: Ok, that's fine. Nonetheless, warming up L1 has close to zero benefits here, unless L1 serialization would be doing I/O or something, which it isn't.

dndx: Agreed. L1 cache is really for reducing GC pressure from having to recreate objects every time L2 hits. Which in this scenario, provides close to no benefits from a performance perspective.

bungle: It will do it in first need, which could be argued with same sentence as well, e.g. zero to no benefit? Why not do it if it is cheap as hell to do it before a first request comes in? I mean what is the downside, it should add almost nothing to warmup time? I surely can comment out those 3 lines if that is a real issue.

PR #4470, opened 2019-04-03
fix(basic-auth) ignore password if nil on basic auth credential patch

nateslo:
### Summary

This PR fixes an issue when patching basic auth credentials: `PATCH consumers/:consumers/basic-auth/:basicauth_credentials`. 

Currently, if `password` is not sent in the PATCH request, nil is hashed and saved as the password. The new behavior is to ignore nil password, preserving the current password if one is not sent. 

### Full changelog

* `_BasicauthCredentials:update` only calls `hash_password` if password is not nil
* `_BasicauthCredentials:update_by_username` only calls `hash_password` if password is not nil
* Adds tests

### DIFF
--- a/kong/plugins/basic-auth/basicauth_credentials.lua
+++ b/kong/plugins/basic-auth/basicauth_credentials.lua
@@ -46,9 +46,11 @@ end
 
 
 function _BasicauthCredentials:update(cred_pk, cred, options)
-  local ok, err, err_t = hash_password(self, cred_pk.id, cred)
-  if not ok then
-    return nil, err, err_t
+  if cred.password ~= nil then
+    local ok, err, err_t = hash_password(self, cred_pk.id, cred)
+    if not ok then
+      return nil, err, err_t
+    end
   end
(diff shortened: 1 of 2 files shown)


hishamhm:
@nateslo I have a hunch that the same problem exists in `update_by_username`, the function defined after `update` (I don't think any code currently uses that function, but it does take a potentially partial `cred`, and it should be there for completeness of the DAO overrides). The upserts should be fine because they only take complete entities.

hishamhm:
(Side note to @Kong/team-core : The bug fixed by this PR is another example of why we need a better high-level interface for general transformations in the DAO, so that users don't have to implement all of `insert`, `update`, `upsert` and all `update_by_*`, `upsert_by_*` when they want to hash a password or do some other encoding on input or output — and also a data point to consider when deciding to provide transformations on a per entity or per field basis)

nateslo:
@hishamhm I didn't see `update_by_username` used anywhere but I think it makes sense for the behavior to be the same. I will update shortly.

PR #4477, opened 2019-04-04
feat(cache) prewarm some negative cache entries for plugins

kikito:
This PR prewarms *some* cache entries for plugins, in addition to the default one (which was provided by `kong.db.plugins:cache_key(plugin)`).

This code is repetitive on purpose (it could be made terser with tables and loops) to make it fast - we want prewarm to be as quick as possible.

### DIFF
--- a/kong/runloop/cache_prewarm.lua
+++ b/kong/runloop/cache_prewarm.lua
@@ -8,7 +8,6 @@ do
   end
 end
 
-
 local tostring = tostring
 local math = math
 local ngx = ngx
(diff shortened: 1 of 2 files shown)


kong/runloop/cache_prewarm.lua
```diff
@@ -17,27 +16,267 @@ local ngx = ngx
...
+
+  local start = ngx.now()
+
+  local ok
+  for entity, err in dao:each(1000) do
+    if err then
+      return nil, err
+    end
+
+    ok, err = write_entity_in_cache(entity_name, dao, entity)
```
hishamhm: _It took me so long to write this comment that by the time I hit "Send" the patch was outdated and the first problem I pointed out had already been fixed by switching to use `safe_add`. :)_

_But still, I think the general point I will make below stands, and the patch does not work as we intend:_

~~I am afraid the behavior is not correct.~~

~~For example, assume you have two `key-auth` plugins, one set to `service`, and another set to both `service` and `consumer`. The second one will overwrite the cache for the first one with `nil` (line 165). But even if you fix this to avoid overwriting existing entries with `nil`,~~ this is only storing nils for entries that are _below_ the plugin's spot in the plugin iterator loop: in the happy path, those additional entries are never tested; we care about the ones _above_ the entry that hits.

If we consider the unhappy cases as well (when the plugin iterator runs through every combination and finds nothing), we would need a complete negative cache:

* for every possible combination of route+service+consumer:
  * if does _not_ appear in some plugin instance:
    * store a `nil` in cache.

That of course explodes in size real fast. But it's the only way to give the Strong Guarantee "every request to plugins_iterator responds without a DB query".

Perhaps a lesser but still useful guarantee would be the Weaker Guarantee "every _positive response_ to plugins_iterator is done without a DB query". That means a potentially smaller negative cache:

* for every combination defined in a plugin
  * for all combinations that may appear _before it_ in the plugins_iterator loop AND also do not appear in some other plugin instance
     * store a `nil` in cache

There are 8 combinations(*) to consider:

* given a R+S+C plugin, no negative cache is needed
* Given R+C, only `nil` for R+S+C is needed (_if_ a valid R+S+C plugin doesn't also exist)
* Given S+C, we need `nil` for R+S+C and R+C (for every possible route that points to that service)
* and so on...

The "for every possible route" there starts to give an idea that this is not that much better. And indeed: given a global plugin, this method ends up requiring the same number of cache entries as the Strong Guarantee. :(

Given all this, I'm starting to doubt the viability of prewarming a negative cache.

One thing we _could_ do is, if we are _certain_ that all plugins were cached, just don't bother hitting the DB at all for un-cached entries and just return `nil` (almost like DB-less works). This would entail:

* prewarm all plugins without negative caching
  * but store plugins and later their negative cache in two separate caches, so nothing else can evict plugins after prewarming
  * if successful (no evictions happened), set `all_plugins_cached = true` in the shm
* if we get an invalidation for a plugin, set `all_plugins_cached = false`, the above assumption no longer holds
* in plugins_iterator, check `all_plugins_cached`:
   * if true, make `load_plugin_from_memory` return `nil`
   * if false, make `load_plugin_from_memory` hit the DB as usual

This would give us the Strong Guarantee, at least as long as the plugin configurations don't change. (an alternative to keep this working regardless of changes would be to handle CRUD events and adjust the cache accordingly; we do something like that for keeping the in-memory representation of Upstreams already.) But it still would take separate positive and negative caches to prevent eviction, which I think we don't currently support. So this proposal is more like an idea for the future.

(* @thibaultcha, a thought that crossed my mind more than once: why do we have the "route + service" combination? given that a route always point to a single service (or nil), this is functionally equivalent to "route" only, right? We could get rid of two combinations in the plugins iterator (R+S and R+S+C) — or is there a scenario I'm not seeing?)

bungle: @hishamhm I think you have a good last point, but how to get rid of it? There is no migration that we can do to get rid of that. It is still usable, e.g. if you want to switch route service back and fort with patch, e.g. service v1 and service v2..

bungle: ^ that said, consumers should not even be a core entity in my opinion :-).

thibaultcha: @hishamhm We can probably get rid of it.

bungle: @hishamhm, and yes I have been saying it from the start that we should have all the plugins in L2/L1 all the time. That will make negative caching go away as it is just about as quickly to look if certain plugin is among the cached ones or not.

kikito:
> It took me so long to write this comment that by the time I hit "Send" the patch was outdated and the first problem I pointed out had already been fixed by switching to use safe_add. :)

Yes, I was subconsciously going over it and at midnight I realized that there was a problem so I needed to add the `safe_add`.

> That of course explodes in size real fast.

Not only in size - also in time. Just *iterating* over all those combinations, even if we don't spend memory on them, may be too much.

> The "for every possible route" there starts to give an idea that this is not that much better.

Yes, exactly. Again, combinatory explosion. That's why I settled down with "filling the cache as much as I could without doing any extra db requests" on this PR.

> We can probably get rid of it (the route + service combination).

I can give that a try. It would reduce the amount of negative caching used.

hishamhm:
Closing this as we're exploring avenues on perf other than negative pre-caching

PR #4481, opened 2019-04-05
perf(plugins) remove route service from plugins iterator

kikito:
Possible plugins_iterator optimization.

As a result of a talk in the comments of #4477.


kikito:
@bungle pointed out a tricky scenario to me: While rare, it is possible that a user has already a setup with two plugins:

* One plugin for just the route
* Another instance of the same plugin plugin for both the route and its service

I don't think we can introduce this change without forcing a backwards-incompatible version number.

PR #4488, opened 2019-04-08
perf(runloop) asynchronous router and plugins rebuilds

### LINKED ISSUE #4194: Kong 1.0 "5K resource challenge/Plugin+Router map build"  findings/errors(Pending Bungle PR fixes this)
### Summary

Made it my weekend mission to vet Kong 1.0 as hard as I could in our real sandbox environment. Many awesome PRs have brought great improvements based on earlier reports. I decided to crank it up to a final test, I build out my Sandbox Kong to:

5k service/route pairs. Each with oauth2 + acl with a whitelist entry (so 2 plugins per route).
5k consumers each with an acl whitelist group added and a pair of oauth2/jwt creds

Additional PRs I included in my Kong 1.0 patched version:
4102
4179
4178
4177
4182

I did so with a little script in a loop calling the admin API from a remote client. Of the resources being built out I would say maybe 5 times total a few 500 level error responses occurred directly from the Admin API calls or from my lb health check pings during the execution of the admin api calls, figured I would list some of those here if they bring any insight or room for improvements:

First error came from my standard 5 second health-check ping calls against the gateway while under resource creation load:
```
2019/01/13 21:07:05 [crit] 35#0: *1415187 [lua] init.lua:639: rewrite(): could not ensure plugins map is up to date: failed to rebuild plugins map: [cassandra] could not execute page query: [Server error] java.lang.NullPointerException, client: 10.xxx.x.x, server: kong, request: "GET /F5/status HTTP/1.1", host: "gateway-dev-core-dc2.company.com"
```
Any way to add retry/hardening logic around this, strange seeing a null pointer exception being thrown(from C* itself considering is a logged java error?). Easy to say well C* caused it not Kong's problem, but is it possible Kong could safely and effectively ascend to attempt to remediate low level db query failures between client->server?

Another one looked like this:
```
2019/01/13 22:15:06 [error] 38#0: *1518092 [kong] init.lua:86 could not execute page query: [Server error] java.lang.NullPointerException, client: 10.xxx.xx.x, server: kong_admin, request: "GET /acls?offset=EKEtFbKFb0ThvMhuC02YnisA8H%2F%2F%2Bb8A HTTP/1.1", host: "kong:8001"
```

Worker-event failed to post due to no memory and another timed out? Is the first an exhaustion of the Lua VM memory? My node as a whole is only using 1.2GB of 5GB mem so seems a bit odd to me, any potential for internal remediation around failed worker-events or ones that may have timed out, I don't see any kong configurations available to the conf/env variables that lets us configure the timeout on worker event data(or what is the seemingly baked in wait for worker event data)?:
```
2019/01/13 22:50:10 [error] 37#0: *1569611 [lua] events.lua:254: post(): worker-events: failed posting event "healthy" by "lua-resty-healthcheck [test_upstream]"; no memory, context: ngx.timer
2019/01/13 22:50:10 [error] 34#0: *1569615 [lua] events.lua:345: post(): worker-events: dropping event; waiting for event data timed out, id: 34317, client: 127.0.0.1, server: kong_admin, request: "PUT /routes/76bb1716-14d0-46fc-a4d3-9b6441e4fa6f HTTP/1.1", host: "localhost:8001"
2019/01/13 22:50:10 [error] 38#0: *1569619 [lua] events.lua:345: poll(): worker-events: dropping event; waiting for event data timed out, id: 34317, context: ngx.timer
2019/01/13 22:50:10 [error] 36#0: *1569621 [lua] events.lua:345: poll(): worker-events: dropping event; waiting for event data timed out, id: 34317, context: ngx.timer
2019/01/13 22:50:10 [error] 35#0: *1569623 [lua] events.lua:345: poll(): worker-events: dropping event; waiting for event data timed out, id: 34317, context: ngx.timer
2019/01/13 22:50:11 [error] 33#0: *1569632 [lua] events.lua:345: poll(): worker-events: dropping event; waiting for event data timed out, id: 34317, context: ngx.timer
2019/01/13 22:50:11 [error] 37#0: *1569634 [lua] events.lua:345: poll(): worker-events: dropping event; waiting for event data timed out, id: 34317, context: ngx.timer
```

Another interesting error, seems almost like my C* node failed the ssl handshake one time?:
```
2019/01/14 00:16:07 [crit] 36#0: *1704447 [lua] init.lua:520: ssl_certificate(): could not ensure plugins map is up to date: failed to rebuild plugins map: [cassandra] could not execute page query: [Server error] java.lang.NullPointerException, context: ssl_certificate_by_lua*, client: 10.xxx.x.x, server: 0.0.0.0:8443
2019/01/14 00:16:07 [crit] 36#0: *1704446 SSL_do_handshake() failed (SSL: error:1417A179:SSL routines:tls_post_process_client_hello:cert cb error) while SSL handshaking, client: 10.xxx.x.x, server: 0.0.0.0:8443
```

Overall I have been impressed with Kong under this 5k stress challenge. I slightly assumed all sort of timers and rebuilding logic would fail due to resource growth but I am not seeing that behavior with the PR's applied and also leveraging -1 on the router perf pr(so I suppose I am partly leveraging async router building). Proxying still happens quickly and my chaos test that kept adding and deleting a service/route pair while the node was under load worked fine.

Feel free to close this whenever. I just thought to relay what behavior I saw. Not too shabby by any means! I think core Kong is getting really close to hitting all my check boxes in recent weeks 👍 . 5k resources with good perf against a real db cluster w network latency should meet 99% of the worlds needs around serving as a gateway imo. Maybe one day in a year or two I will give the 10k challenge a spin 😆.

### Additional Details & Logs
- Kong version 1.0 with pending/merged perf PRs

bungle:
### Summary

Latest performance 🚀 patches that we need to figure out how to get into `1.2.0`.

This PR replaces previous #4202 and #4102, and it changes the implementation a little bit.

Now you cannot configure the `rebuild_timeout` anymore as a number but only `on/off`. There is now only one configuration parameter that is used for both router and plugins rebuilds. The plugins map is turned into plugins map and plugins cache. The plugins iterator doesn't anymore rely on Kong entity level cache. It also adds DNS cache prewarming for `services`. In my simple test it removed 49ms from initial request of 50ms Kong latency. It also runs plugins code only when plugin has a custom function for a context/phase (e.g. `init_worker`). It also includes @hishamhm's #4494.

To enable `asynchronous` rebuilds on `router` and `plugins` you have to turn it on:
```
KONG_ASYNC_REBUILDS=true
```

Even without that the rebuilds can happen asynchronously, but that setting ensures that rebuilds will never happen synchronously during request.

I think the biggest issue here is that this contains just one big commit. But on the other hand it is not that big: 1290 additions and 814 deletions. But sure, separating this to multiple is perhaps required (?). Hard to make all those separate commits atomic. Another is that this touches the very core of Kong's critical `hot path` code.

### Issues resolved

Fixes #4194


hishamhm:
Whoa, big changes here! Reusing the data loaded by the plugin map as a cache sounds like a win for most scenarios — but it makes one big assumption that we were not making before: that all plugin instances can fit in memory at one time. (Currently, we work under that assumption for Upstreams, but not for Plugins — given use-cases for per-consumer Plugins, I am not sure this is an assumption we are able to make.)

But before I try to understand the whole PR myself, could you explain a bit why the initalization of declarative config was moved out of `kong.init`? At first glance, this change doesn't seem related to anything from the PR description above.

bungle:
declarative was moved because it is almost same stuff as router building and plugins map. easier to follow the code and keeps kong.init clean.

bungle:
If you can affort millions of plugins, i think you can affort enough ram was the reasoning.

bungle:
If the `plugins won't fit into memory` is a real a concern (I agree that it changes the current behavior a bit), then I suggest that we add an option, e.g. `KONG_PLUGINS_IN_MEMORY=true` (`false` would mean current behavior, and the plugins iterator should use code similar to current). Or perhaps some intelligent auto-adjusting.

I understand that memory requirement is `number_of_workers` times `configured plugins`. We need to measure how many plugins configurations can you fit into say 128 MB. In most cases single plugin takes a few kBs (at most), let's say it eats 5 kB (scaling that number up), and you have 8 core machine. Then it is 40 kB per configured plugin on all 8 workers combined. So in 1 MB you can have 25. Kong default mem cache size is 128 MB, which is not used for plugins anymore, but what current behavior uses (that is shared between workers, but workers also have their own L1 caches in current).  So you can fit some 25000 (need to check for real numbers) plugins in Kong cache currently until LRU kicks in.

RAM is a new SSD, and SSD is a new HDD, which is the old Tape, with things like Intel Optane coming, :-). Well sure at the same time there is IoT, but I don't think IoT is going to have many plugins, and I think with IoT you actually want to fit everything in memory (as little as needed). Kong's use case is more on server side though.

kong/runloop/handler.lua
```diff
@@ -152,95 +767,233 @@ do
...
+    if wait and wait > 0 then
+      rebuild_sync(callback, current_version, semaphore)
+      release_semaphore(semaphore)
+
+    else
+      ok = rebuild_async(callback, current_version, semaphore)
+      if not ok and wait == 0 then
+        rebuild_sync(callback, current_version, semaphore)
+        release_semaphore(semaphore)
+      end
```
p0pr0ck5: are we missing a `release_semaphore` in this branch?

bungle: @p0pr0ck5 I think we are. Good catch.

bungle: Fixed now.

bungle:
Now that we have added the plugin combos:
1. https://github.com/Kong/kong/pull/4488/commits/384f9eaac9a4727054836223c24256c273551df8
2. https://github.com/Kong/kong/pull/4488/commits/d8ec15b5ba0fad99f8a1a708790bd9e35aa29248

I don't consider it fully neccessary to cache all the plugins in a memory. These two combo patches greatly limit the plugins iterator to only check the necessary plugin bindings (service, route, consumer, global), and in most cases we don't endup touching the database even if we want to revert back to previous state where plugins could be evicted from the cache. Of course caching plugins fully (like done here) makes it more predictable (e.g. gurantees that on hot proxy path we do not go to database at any point to fetch plugins, like we don't do with routes). It also makes it pretty safe to turn off database when you have these patches applied, e.g. this would not happen: https://github.com/Kong/kong/issues/4520

ophusky:
I have tested this branch, and it works  fine, It solved my problem([#4520](https://github.com/Kong/kong/issues/4520)).

Under this code branch，I have tested by folling steps:
1.  Default config and all normal run.
2. stop postgresql.
3. Do some request, and all successed.

Does this branch code  mean that all data in database(route，service，consumer, plugin...)  is cached at startup?

bungle:
> Does this branch code mean that all data in database(route，service，consumer, plugin...)
> is cached at startup?

Routes / plugins are loaded at startup on all workers, and rebuilds usually happen on background timer. but by default may also happen on request time. It is also a bit more resilient to errors and contains some additional code to reduce possible db lookups and patches to make rebuilds faster.

The main point is to do asynchronous rebuilds so that they don't affect proxying latency. It is also pro-active, e.g. if you change the route, all the nodes and all the workers on node will start rebuilding those core structures, and not wait for request to come in to trigger a rebuild. It means that even without `KONG_ASYNC_REBUILDS=true` the workers may have build those structures already when the request comes in (that setting just means: NEVER do rebuilds while processing client request / connections). In worst case it takes about 1 second for workers on same node to start rebuilding, and 6 seconds to workers on other nodes to start rebuilds as well. Usually less (it is all about timing and cluster notification is configurable: with `KONG_DB_UPDATE_FREQUENCY`).

kong/runloop/handler.lua
```diff
@@ -152,95 +779,274 @@ do
...
+    local current_version = get_version(name)
+    if current_version == version then
+      return
+    end
 
+    local ok = acquire_semaphore(semaphore, wait)
     if not ok then
-      return nil, err
+      if wait and wait > 0 then
+        rebuild_sync(callback, current_version, semaphore)
```
p0pr0ck5: there are multiple calls to `rebuild_sync` that define a third param `semaphore`, but it does not seem used?

jeremyjpj0916:
I would say a fair assumption on how to run Kong in a simple manner and probably most common would be this from Gateway view(I know all the rave seems to be mesh, but I think Gateways are still the most prominent implementation out in the wild right now):

1x route to 1x service, (where its safe to assume between the two of them 2-5 plugins will be enabled, 1 of two generally being ACL, 2 of 2 being Auth type, leaving 3 others for potential oddity plugins the proxy requires). We like to put our plugins strictly at the route layer and leave service clean of plugins.

The ACL plugin likely runs a white list with a value unique to that proxy that Consumers add to their acl list for access.

1-10 global plugins depending on features people like to execute on all proxy paths
For each Route/Service pair anywhere from 1-20 consumers tied to it authorized to call the proxy.

Consumers with a few different types of creds(maybe oauth2/jwt) + their acl groups.

Taking that as a base case you get something like this for bloat(worst case):
1000 "Proxies"(service/route pairs)
5010 plugins (5000 for proxies 10 for global)
20000 Consumers each with their creds and lists of ACL access list

5000 "Proxies"(service/route pairs)
25010 plugins (5000 for proxies 10 for global)
100000 Consumers each with their creds and lists of ACL access list

Then lets think kinda max case for a general shared gateway(10,000 proxies I think is accurate for 99.9% of any use case):
10000 "Proxies"(service/route pairs)
50010 plugins (50000 for proxies 10 for global)
200000 Consumers each with their creds and lists of ACL access lists

I don't ever expect to cross the 5k boundary personally, but if Kong reaches a design where my max case above generally does not cause problems during resource change or latency increases beyond say 10ms from the baseline(1k proxies) I think it would be huge achievement of Kong with regards to performance of the product.

I personally have not gotten a chance to test all this but do hope to do so when I get a breath of fresh air!

Even with assurances the timer model is likely going to meet the need I would probably still run:
KONG_ASYNC_REBUILDS=true because 100% consistency across all my Kong nodes in an exact given second is never going to be a need for my use case. 

What changes the most generally are new consumers, new/modified routes/services and then new/modified plugins on top of that. And when we stand up new services we never have a need for it to be available in 1-5 seconds(which I understand is like worst case anyways?) realistically. Takes minutes before a consumer or that proxy gets on boarded into process and use generally. I think the most critical piece is just ensuring change/chaos does not impact existing resources called frequently in their hot paths. And I think its getting solved in v1.2 with these awesome changes if they can get pushed through 🦍 .

kong/runloop/handler.lua
bungle: @p0pr0ck5 as always, thank you for this! I'll remove that.

bungle: Fixed now.

hishamhm:
Continues in #4639! :rocket:

PR #4489, opened 2019-04-08
fix(schema) address issue where field type "record" nested values reset on update

Jlawlzz:
This PR enables nested `record` field types to merge as expected on `PATCH`.  Previously any unset values within the incoming record object would reset previously set values to either a default value or null. This change allows the previously set values to be respected during a partial update by conditionally setting `read_before_write` to `true`.


kikito:
Approved, but I think this should be done against `master`, not `next`.

PR #4494, opened 2019-04-09
perf(runloop) only check plugin combinations in use

hishamhm:
Our plugin iterator checks for up to 8 possible combinations of foreign keys when looking for plugins to trigger ("route + service + consumer" (RSC), "route + service" (RS), RC, SC, R, S, C, and global plugin). This means that on an unprimed cache, it takes up to 8 database accesses.

It is unlikely that a Kong installation will use all 8 plugin configuration styles, meaning that some of the combinations above will never produce a valid result.

The plugin iterator uses a "plugins map" we create at start-up (and recreate when Kong is reconfigured), which scans all plugins in the database to determine which plugins are in use.

While we are at it, we can check which combinations are not ever used by any plugin, and we can disable those in the plugin iterator. This is what this commit does, picking a low-hanging fruit from the current plugin iterator design without changing any of its underlying assumptions. (More radical changes could have been made to the plugin iterator exploiting this knowledge, such as generating the iterator function using metaprogramming to avoid a few in-memory comparisons, but this version gives the DB-access improvement for practical scenarios with a low code-change footprint).

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -115,6 +115,7 @@ local PLUGINS_MAP_MUTEX_OPTS
 local declarative_entities
 local plugins_map_version
 local configured_plugins
+local configured_plugin_combos
 local loaded_plugins
 local schema_state
 
@@ -132,6 +133,7 @@ end
 
 local function build_plugins_map(db, version)
   local map = {}
+  local combos = {}
 
   for plugin, err in db.plugins:each(1000) do
     if err then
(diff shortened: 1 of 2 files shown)


bungle:
Decent stuff, I will include this to #4488 with slight modifications to fit the other changes made there.

bungle:
@hishamhm,

I think we can do this also by `plugin.name`, see:
https://github.com/Kong/kong/pull/4488/commits/c5a40cfdb40418bfeb8ffa549dacb3f1ff73832a

hishamhm:
Reworked by @kikito and @bungle in https://github.com/Kong/kong/commit/549247f663e6676f50ece0ca9c63d5389f48ad8a

PR #4503, opened 2019-04-12
DELETE /cache/:key should invalidate cache in all nodes

Maples7:

### Summary

DELETE /cache/:key should invalidate cache in all nodes

### Full changelog

* DELETE /cache/:key uses invalidate rather than invalidate_local

### DIFF
--- a/kong/api/routes/cache.lua
+++ b/kong/api/routes/cache.lua
@@ -20,7 +20,7 @@ return {
     end,
 
     DELETE = function(self)
-      kong.cache:invalidate_local(self.params.key)
+      kong.cache:invalidate(self.params.key)
 
       return kong.response.exit(204) -- no content
     end,


thibaultcha:
What is motivating this change? The invalidation is local to a node by design, and should (for the use-cases solved by this endpoint) stay like so.

Maples7:
We have developed our own custom plugins which uses `kong.cache` widely as a way to improve performance and we deploy it with multiple nodes as a Kong cluster in a Kubernetes environment. From my perspective, should not `DELETE /cache/:key` invalidate cache in all nodes to keep data consistency in all Kong nodes? For instance, once we change some data in database, we are expected to invalidate the cache in all nodes by default to make it always consistent for every requests from users' clients. 

Even though that is not the default behavior, I think we should provide an Admin API to achieve it. Of course, we can also extend the Admin API to make it happen.

thibaultcha:
Cluster-wide cache invalidation is already automatically handled at the DAO layer, there is no need to interact with the Admin API for that. Many plugins rely on this mechanism (see the `cache_key` value in their schema) entirely transparently. It is quite nice. 

The cache admin api endpoint is a _debug_ endpoint for individual nodes in case something went wrong and a cache invalidation event did not properly propagate in a production scenario, for example. I never heard of a situation that even required using it, as the cache invalidation mechanism is quite robust (so long as the database is healthy and nodes are correctly connected to it, as per the clustering documentation).

https://docs.konghq.com/1.1.x/clustering/#interacting-with-the-cache-via-the-admin-api

Maples7:
I see... but the bad news is we do not use Kong’s database to store our own data so we also do not make use of the DAO layer in our custom plugins.

Actually, we retrieve user data from a remote Redis and cache them in L1 memory with the help of kong.cache. So we need to manually invalidate the cache in all nodes once it has been changed in Redis. Maybe we misused Kong...

Maples7:
How about supporting this feature to make custom plugin developing more friendly? Or can you give us some other advices about how to adjust our usage to meet the standard of Kong?

Many thanks to you! 🙏

thibaultcha:
Alas, such an endpoint makes little sense for Kong's current architecture and plugins ecosystem: as I said, it is intended for debug purposes on a single node level. Adding a cluster-wide invalidation endpoint would only further increase misusage of the cache invalidation mechanism, and serve little purpose in the first place given how Kong's database caching works.

Caching data from an external data source sounds like an interesting use-case, but should not require an endpoint in the core Admin API, or at least not without additional abstraction. Why not calling `kong.cache:invalidate(key)` directly from your plugin's Lua code for now? It is available from both `handler.lua` and `api.lua` code. If the cache invalidation is triggered by a source external to Kong, then you can implement your own endpoint in your plugin's `api.lua` file and call the invalidation mechanism there.
A slightly cleaner abstraction for plugins who wish to rely on the caching mechanism for external data sources would be to instantiate their own `kong.cache` (or `mlcache`) instance, and hook it to their own `kong.cluster_events` broadcasting channel for invalidations. But the [PDK](https://docs.konghq.com/1.1.x/plugin-development/) does not yet allow for this level of customization.

Maples7:
I see, we can make use of api.lua to extend Admin API to achieve that outer service is able to invalidate data. Thanks!

Issue #4502, opened 2019-04-12
Kong 1.1 cannot bootstrap or reset schema on Azure Postgres


rainest:
### Summary

Kong 1.1 fails to bootstrap using [Azure PostgreSQL](https://docs.microsoft.com/en-us/azure/postgresql/). Attempting to bootstrap will fail with an error like

```
Error: [PostgreSQL error] failed to bootstrap database: ERROR: role "kong@kong" does not exist
  Run with --v (verbose) or --vv (debug) for more details
```
Kong 1.0 is able to bootstrap without issues, and bootstrapping on 1.0 and then upgrading/running `kong migrations up` works/creates a viable 1.1 environment. This appears to be the best workaround at present due to the nature of the issue.

### Steps To Reproduce

1. Create an Azure Postgres instance and a user for Kong.
2. Attempt to run `kong migrations bootstrap` from a 1.1 instance configured to use the Postgres instance created in (1). Bootstrapping will fail with a `role "kong@kong" does not exist` error.

### Additional Details & Logs

This issue was introduced by #4167, specifically via the query added at https://github.com/Kong/kong/blob/1.1.1/kong/db/strategies/postgres/connector.lua#L633-L639

Azure Postgres uses a non-standard login system for Postgres, and requires that clients authenticate using a `user@dbname` username. This appears to be handled by some additional layer in front of Postgres proper, as the same `user@dbname` format is not used for Postgres internals:

```
root@ubuntu-bionic:/home/vagrant# psql -Ukong@kong -h "example.postgres.database.azure.com" kong
...
kong=> select rolname from pg_roles where rolname like '%kong%';
 rolname 
---------
 kong
(1 row)

kong=> select usename from pg_user where usename like '%kong%';
 usename 
---------
 kong
(1 row)
```
Although Kong is able to successfully connect to the Postgres instance, queries that attempt to reference its user will fail because the user does not actually exist, at least not under the name Kong will use. Kong will attempt to run `CREATE SCHEMA IF NOT EXISTS public AUTHORIZATION kong@kong; GRANT ALL ON SCHEMA public TO kong@kong;` whereas Postgres expects `CREATE SCHEMA IF NOT EXISTS public AUTHORIZATION kong; GRANT ALL ON SCHEMA public TO kong;`.

rainest:
[...]

PR #4506, opened 2019-04-12
feat(db) pg username indirection (e.g. azure) and limited permissions support

### LINKED ISSUE #4483: Issue running Migration Bootstrap with limited database permissions
### Summary 
When running the Kong Migration with Postgres schema level permissions, the migration fails. We ran in to this trying to run Kong with least privilege within the Postgres database. The issue is that Kong can not migrate without database level permissions. We are making the user and the Postgres schema in steps that happen before running the Kong migration.

### Steps To Reproduce
1. Create Kong App User in PostgresDB.
2. Create schema in Postgres database that is named the same as the Kong app user and grant authorization to that schema for the Kong app user.
3. Grant the needed permissions for the Kong user with the correct permissions *within the schema*. in our case we granted `ALL` to our Kong user in our schema.
4. Run the `kong migration bootstrap` on the database to build out the database that Kong will use.
5. The migration should fail because the Kong app user should not have permission to preform any DB level actions.

### Additional Details & Logs

- Kong version: `1.1.1`
- Failure on the `kong migration bootstrap --vv`: 
```
2019/04/05 18:31:15 [info] bootstrapping database...
Error: 
/usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:55: [PostgreSQL error] failed to bootstrap database: ERROR: permission denied for database kong
stack traceback:
	[C]: in function 'assert'
	/usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:55: in function 'bootstrap'
	/usr/local/share/lua/5.1/kong/cmd/migrations.lua:118: in function 'cmd_exec'
	/usr/local/share/lua/5.1/kong/cmd/init.lua:88: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:88>
	[C]: in function 'xpcall'
	/usr/local/share/lua/5.1/kong/cmd/init.lua:88: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:45>
	/usr/local/bin/kong:7: in function 'file_gen'
	init_worker_by_lua:48: in function <init_worker_by_lua:46>
	[C]: in function 'xpcall'
	init_worker_by_lua:55: in function <init_worker_by_lua:53>
2019/04/05 18:31:15 [debug] creating 'kongsomerandomsha' schema if not existing...
```
- Error from Postgres:
```
2019-04-05 15:57:14.092 UTC [40] ERROR:  permission denied for database kong
2019-04-05 15:57:14.092 UTC [40] STATEMENT:  BEGIN;
	  CREATE SCHEMA IF NOT EXISTS "kongsomerandomsha" AUTHORIZATION "kongsomerandomsha";
	  GRANT ALL ON SCHEMA "kongsomerandomsha" TO "kongsomerandomsha";
	  SET SCHEMA 'kongsomerandomsha';
	COMMIT;
2019-04-05 15:57:
```
- OS: alpine
- Postgres: 10.4

## Closing thoughts
This seems to be caused by the Kong app needing permissions outside the scope of its schema to set up the database. Normally if you are going one Kong instance per One database this could be fine to have the kong app with CreateDB, in our case we would like to be able to double dip on the PostgresDB to allow for warming up new versions of kong or being able to keep older versions running.

A possible fix fo this would be to check for the schema without using the `CREATE` command. If we were to check for the schema with a `SELECT` command, we would not need to grant `CREATEDB` to the kong app user with something like this: 

```
IF NOT EXISTS (
    SELECT *
    FROM information_schema.schemata
    where schema_name = 'kong_schema_name') THEN
    CREATE SCHEMA kong_user_name AUTHORIZATION kong_user_name;
    GRANT ALL ON SCHEMA kong_user_name TO kong_user_name;
    SET SCHEMA 'kong_schema_name';
END IF;
```

This should allow for checking if the schema exists with out needing the database level permissions to create a schema. If the schema doesn't exist, and Kong tries to create the schema, it would be up to the user to provide high enough permission to allow Kong to create the schema or create their schema themselves.

bungle:
### Summary

This PR tries to make Postgres schema bootstrapping and resetting more robust in scenarios where the username used to connect Postgres database has limited permissions (e.g. unable to drop or create schemes), see #4483.

It also fixes issue with (Azure Postgres) username indirection, see: #4502.

### Issues resolved

Fix #4483, #4502

### DIFF
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -118,6 +118,61 @@ local function iterator(rows)
 end
 
 
+local function get_table_names(self, excluded)
+  local i = 0
+  local table_names = {}
+  for row, err in self:iterate(SQL_INFORMATION_SCHEMA_TABLES) do
+    if err then
+      return nil, err
+    end
+
+    if not excluded or not excluded[row.table_name] then
+      i = i + 1
+      table_names[i] = self:escape_identifier(row.table_name)
+    end
+  end
+
+  return table_names
+end
+
+
+local function reset_schema(self)
+  local table_names, err = get_table_names(self)
+  if not table_names then
+    return nil, err
+  end
+
+  local drop_tables
+  if #table_names == 0 then
+    drop_tables = ""
+  else
+    drop_tables = concat {
+      "    DROP TABLE IF EXISTS ", concat(table_names, ", "), " CASCADE;\n"
+    }
+  end
+
+  local schema = self:escape_identifier(self.config.schema)
+  local ok, err = self:query(concat {
+    "BEGIN;\n",
+    "  DO $$\n",
+    "  BEGIN\n",
+    "    DROP SCHEMA IF EXISTS ", schema, " CASCADE;\n",
+    "    CREATE SCHEMA IF NOT EXISTS ", schema, " AUTHORIZATION CURRENT_USER;\n",
+    "    GRANT ALL ON SCHEMA ", schema ," TO CURRENT_USER;\n",
+    "  EXCEPTION WHEN insufficient_privilege THEN\n", drop_tables,
+    "  END;\n",
+    "  $$;\n",
+    "    SET SCHEMA ",  self:escape_literal(self.config.schema), ";\n",
+    "COMMIT;",  })
+
+  if not ok then
+    return nil, err
+  end
+
+  return true
+end
+
+
 local setkeepalive
 
 
@@ -196,7 +251,9 @@ setkeepalive = function(connection)
 end
 
 
-local _mt = {}
+local _mt = {
+  reset = reset_schema
+}
 
 
 _mt.__index = _mt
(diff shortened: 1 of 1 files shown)


kong/plugins/oauth2/migrations/001_14_to_15.lua
```diff
@@ -54,12 +54,24 @@ return {
...
       CREATE INDEX IF NOT EXISTS "oauth2_tokens_credential_id_idx"        ON "oauth2_tokens" ("credential_id");
       CREATE INDEX IF NOT EXISTS "oauth2_tokens_service_id_idx"           ON "oauth2_tokens" ("service_id");
-      CREATE INDEX IF NOT EXISTS "oauth2_tokens_api_id_idx"               ON "oauth2_tokens" ("api_id");
 
+      DO $$
+      BEGIN
+        CREATE INDEX IF NOT EXISTS "oauth2_authorization_api_id_idx"
+                                ON "oauth2_authorization_codes" ("api_id");
+      EXCEPTION WHEN UNDEFINED_COLUMN THEN
+        -- Do nothing, accept existing state
```
hishamhm: Adding this won't hurt, but I'm confused: this is `14_to_15`, which assumes a 0.14 state (where `api_id` exists) migrating into 0.15 (where `api_id` _still_ exists). How is it possible that this runs and `api_id` does not exist? Ahhh, we run all `core` migrations (bringing the core entities to 1.x) before running any `plugin` migrations, right? That explains it.

Err, that's a bit problematic. This means that every time we delete a column we have to go back to old migrations and fix them this way. One possible answer would be that core and migrations need to run migrations in lockstep going version-by-version, but if we consider a third party plugin that is installed _after_ Kong 1.x is installed, and has migrations to run, they do not have this luxury. _Every_ migration touching foreign fields would have to be written assuming that they might not be there in the future...

Ok, got it. (Thanks for reading my stream-of-consciousness rambling as I evolve my understanding of it. :) ) 

So let's do this — but something similar to this probably needs to be in every other plugin that uses `api_id` in migrations, right? A quick `grep` in `kong/plugins` shows lots of `api_id` in migrations. Perhaps that should be in a separate PR will all of them?

PR #4510, opened 2019-04-13
fix(http-log) simplify queueing mechanism

kikito:
This PR replaces the plugins/log-buffering folder with a single kong/tools/batch_queue.lua file, fixing one bug in http-log along the way.

Fixed bug: the old implemenation of buffers used a module-level global table called retry_delays to store delays per-buffer, using the buffer ids (a string) as key on that table. The http-log module created many buffers (one per route) bug gave all the same name (`http-log`). As a result, the `retry_delays` value was overwritten by many buffers. This could lead to loss of data.

Highlights of current refactor:

* Renamed "LogBuffer" to "BatchQueue", and "send" to "process".
* The module-level `retry_delays` variable was replaced by an instance-level
  variable: self.retry_delays
* As a result, queues don't need to have names any more.
* BatchQueue has only 1 mandatory parameter, called `process`. This replaces
  the old "sender" abstraction.
* The optional "opts" parameter can be used to provide configuration. Two
  options have been renamed (they remain unchanged in the http-log
  config for backwards compatibility):
  * `queue_size` is now `max_batch_size`
  * `send_delay` is now `process_delay`
* The "producers" abstraction has been replaced by `self.current_batch`
* Queues do not concern themselves with byte sizes any more, they only deal
  with "entities", in the abstract. The 2MB size limit has been removed.
* Compacting several entities into a single body/payload" (i.e. for sending)
  is now a responsibility of the `process` function. Previously this was
  done in the "producers". Notice how the `process` function is created
  in the http-log handler.
* `add_entry` was renamed to `add`
* `self.timer_flush_pending` was renamed to `self.flush_scheduled` and
  `self.timer_send_pending` was renamed to `self.process_scheduled`, to
  better reflect how they function
* `delayed_flush` was renamed to `flush` and `run_sender` was renamed
  to `process`.
* The `adjust_retries` private function was merged into the `send`
  private function.
* The public method `Buffer:send` was merged into the `Buffer:flush` method.
* Logging is now done via `kong.log` in the plugin and via `ngx.log` in
  the queue. This is no longer a config option


kong/tools/batch_queue.lua
```diff
@@ -0,0 +1,317 @@
...
+
+  else
+    ngx_log(ERR, "failed to process entries: ", tostring(err))
+
+    batch.retries = batch.retries + 1
+    if batch.retries < self.retry_count then
+      -- queue our data for processing again, at the end of the queue
+      self.batch_queue[#self.batch_queue + 1] = batch
+    else
+      ngx_log(WARN, fmt("entry batch was already tried %d times, dropping it",
```
Tieske: this should be an ERR probably

kong/tools/batch_queue.lua
```diff
@@ -0,0 +1,317 @@
...
+
+  local next_retry_delay
+
+  local ok, err = self.process(batch.entries)
+  if ok then -- success, set retry_delays to 1
+    self.retry_delay = 1
+    next_retry_delay = 1
+
+  else
+    ngx_log(ERR, "failed to process entries: ", tostring(err))
```
Tieske: this should go in the `if` below and be a WARN?

kong/tools/batch_queue.lua
```diff
@@ -0,0 +1,317 @@
...
+
+---------
+-- Queue
+---------
+
+
+-------------------------------------------------------------------------------
+-- Initialize a batch queue with background retryable processing
+-- @param process function, invoked to process every payload generated
+-- @param opts table, optinally including
```
Tieske: optinally

kong/tools/batch_queue.lua
```diff
@@ -0,0 +1,317 @@
...
+    end
+
+    self.retry_delay = self.retry_delay + 1
+    next_retry_delay = min(RETRY_MAX_DELAY, self.retry_delay * self.retry_delay)
+  end
+
+  if #self.batch_queue > 0 then -- more to process?
+    ngx_log(DEBUG, fmt("processing oldest data, %d still queued",
+                   #self.batch_queue - 1))
+    local oldest_batch = remove(self.batch_queue, 1)
```
Tieske: remove is not `nil` safe, so when adding entries you should validate `entry ~= nil`

kong/tools/batch_queue.lua
```diff
@@ -0,0 +1,317 @@
...
+
+  return setmetatable(self, Queue_mt)
+end
+
+
+-------------------------------------------------------------------------------
+-- Add data to the queue
+-- @param payload included in the queue
+-- @return true, or nil and an error message.
+function Queue:add(entry)
```
Tieske: assert here that `entry ~= nil` since table manipulation is not nil-safe

kong/plugins/http-log/handler.lua
```diff
@@ -18,68 +17,146 @@ HttpLogHandler.PRIORITY = 12
...
-  local buf = buffers[route_id]
-  if not buf then
-
-    if conf.queue_size == nil then
-      conf.queue_size = 1
-    end
+  local entry = cjson_encode(basic_serializer.serialize(ngx))
 
+  local route_id = conf.route_id or "global"
+  local q = queues[route_id]
```
Tieske: caching it by route is weird? shouldn't it be cached by plugin `conf` table?

Tieske: Also `conf.route_id` (slightly above) is always `nil`. `route_id` lives on the top level. (it is passed on the management api as `route_id` and not as `config.route_id`)

kong/plugins/http-log/handler.lua
```diff
@@ -18,68 +17,146 @@ HttpLogHandler.PRIORITY = 12
...
 
 
--- serializes context data into an html message body.
--- @param `ngx` The context table for the request being logged
--- @param `conf` plugin configuration table, holds http endpoint details
--- @return html body as string
-function HttpLogHandler:serialize(ngx, conf)
-  return cjson_encode(basic_serializer.serialize(ngx))
+local function json_array_concat(entries)
+  return "[" .. table.concat(entries, ",") .. "]"
```
Tieske: maybe cache local the `table.concat`?

kong/plugins/http-log/handler.lua
```diff
@@ -18,68 +17,146 @@ HttpLogHandler.PRIORITY = 12
...
+      ["Content-Length"] = #payload,
+      ["Authorization"] = parsed_url.userinfo and (
+        "Basic " .. ngx_encode_base64(parsed_url.userinfo)
+      ),
+    },
+    body = payload,
+  })
+  if not res then
+    kong.log.err("failed request to ", host, ":", tostring(port), ": ", err)
+    httpc:set_keepalive(keepalive)
```
Tieske: should we set keepalive in case of an error? Not entirely sure about keepalive behaviour...

kong/plugins/http-log/handler.lua
```diff
@@ -18,68 +17,146 @@ HttpLogHandler.PRIORITY = 12
...
+    return false
+  end
+
+  -- read and discard response body
+  -- TODO should we fail if response status was >= 500 ?
```
Tieske: at least log an error in case >= 500 ?

kong/plugins/http-log/handler.lua
```diff
@@ -18,68 +17,146 @@ HttpLogHandler.PRIORITY = 12
...
--- Only provide `name` when deriving from this class. Not when initializing an instance.
-function HttpLogHandler:new(name)
-  name = name or "http-log"
-  HttpLogHandler.super.new(self, name)
+-- Parse host url.
+-- @param `url` host url
+-- @return `parsed_url` a table with host details:
+-- scheme, host, port, path, query, userinfo
+local function parse_url(host_url)
+  local parsed_url = url.parse(host_url)
```
Tieske: maybe add a cache table for the parsed urls here? since its config data, they are static

kong/plugins/http-log/handler.lua
p0pr0ck5: If this `res` failed, that's likely an indication there was a transport layer issue. I don't think we should try to keep the connection alive here as the socket may be a dead/dubious state.

Also, I thought we should be cautious about always setting the connection to keepalive. In cases where the remote server doesn't respond with a `Connection: keep-alive` header, calling `httpc:set_keepalive` will fail. Though it doesn't have any logical impact, it generates a lot of unnecessary log noise. Ref:

https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L189-L197
https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L690

kong/plugins/http-log/handler.lua
kikito: I have reworked this ( >= 400 instead of 500). 

The plugin will now consider it a "failure". An error will be logged and this function will return `false`, so the batch will be retried.

@hishamhm you might consider this one of those "slightly breaking changes" we have recently talked about (before, a 500 status would not trigger a retry).

kong/plugins/http-log/handler.lua
kikito: I understand why you might think that. You got trumped by the magic of the plugins iterator. `conf.route_id` it is populated by it here: https://github.com/Kong/kong/blob/master/kong/runloop/plugins_iterator.lua#L64

I also think it is a bit weird to do this by route instead of by other criteria, but that's how it was done before this refactor so I decided to leave that fight for another day.

> caching it by route is weird? shouldn't it be cached by plugin conf table?

I would cache it by plugin.id if I had it. I'm hesitant to use the conf table since I am not 100% sure those tables are never duplicated.

kikito:
@Tieske thanks for your review. I have implemented most of the changes you recommended. I marked the trivial ones as resolved and have left comments on the ones that I think need some explanation.

kong/plugins/http-log/handler.lua
Tieske: since this uses the mlcache to store it I guess its shared with other workers and serialized?, hence using the `conf` table itself as a key doesn't work I guess.

I think what really makes sense here is to create 1 queue per log server-config. So maybe construct the cache key from that info?

hishamhm: > I think what really makes sense here is to create 1 queue per log server-config. 

This makes sense, but there's a mismatch in that `queue_size` is defined per plugin. Agree with @kikito that `plugin_id` would be ideal if available (given current public interface for the plugin), but yeah, this should suffice for now.

kong/plugins/http-log/handler.lua
hishamhm: > before, a 500 status would not trigger a retry

I believe this behavior was never really documented, and it's not the kind of thing that users would "rely on", so I think it's fine to change.

kong/plugins/http-log/handler.lua
kikito: After talking privately with @Tieske and @hishamhm about this we decided to go with an alternative approach which uses "a serialized form of the plugin config" as a queue key instead of the `route_id` which was used previously.

kong/plugins/http-log/handler.lua
```diff
@@ -18,68 +19,172 @@ HttpLogHandler.PRIORITY = 12
...
+  return fmt("%s:%s:%s:%s:%s:%s",
+             conf.http_endpoint,
+             conf.method,
+             conf.content_type,
+             conf.timeout,
+             conf.keepalive,
+             conf.retry_count,
+             conf.queue_size,
+             conf.flush_timeout)
+end
```
Tieske: Not sure it is worth it, but maybe cache the created key?

```lua
local function create_queue_id(conf)
  return fmt("%s:%s:%s:%s:%s:%s",
             conf.http_endpoint,
             conf.method,
             conf.content_type,
             conf.timeout,
             conf.keepalive,
             conf.retry_count,
             conf.queue_size,
             conf.flush_timeout)
end

local queue_id_cache = setmetatable({},{
  __mode = "k",
  __index = function(self, conf)
    self[conf] = create_queue_id(conf)
    return self[conf]
  end,
}
```

kong/plugins/http-log/handler.lua
```diff
@@ -18,68 +19,172 @@ HttpLogHandler.PRIORITY = 12
...
 
-  local route_id = conf.route_id or "global"
-  local buf = buffers[route_id]
-  if not buf then
-
-    if conf.queue_size == nil then
-      conf.queue_size = 1
+  local entry = cjson_encode(basic_serializer.serialize(ngx))
+
+  local queue_id = get_queue_id(conf)
```
Tieske: and change this to:

```lua
  local queue_id = queue_id_cache[conf]
```

kikito: I'm going to opt out to this one for the same reason I mentioned before: I'm not 100% positive the tables are not duplicated.

PR #4512, opened 2019-04-14
feat(request-transformer) validate header values in plugin configuration

rune-chan:
### Summary

Header-related configurations of request transformer are not validated. Invalid header value configurations would be accepted but the invalid headers are not properly transformed previously. This commit adds validations for header fields.

### Full changelog
* Add request-trasnformer plugin validation
* Add related tests


hishamhm:
@rune-chan merged (changing the scope from `feat` to `fix` during the squash-merge), thank you!

PR #4518, opened 2019-04-16
feat(dao) support loading custom DAO strategies for plugins

p0pr0ck5:
### Summary

This commit introduces new functionality into the plugin_loader module to support attempting to load a strategy for custom plugin DAO implementations. This would allow plugins to define a strategy at a path like

`kong/plugins/<plugin>/db/strategies/<strategy>/<entity>.lua`

This is essentially a copypasta of the custom strategy loader found in kong.db.strategies.

### Full changelog

* support loading custom DAO strategies for plugins

### DIFF
--- a/kong/db/dao/plugins.lua
+++ b/kong/db/dao/plugins.lua
@@ -150,12 +150,39 @@ local function load_plugin_handler(plugin)
 end
 
 
-local function load_plugin_entity_strategy(schema, db)
+local function load_plugin_entity_strategy(schema, db, plugin)
   local Strategy = require(fmt("kong.db.strategies.%s", db.strategy))
   local strategy, err = Strategy.new(db.connector, schema, db.errors)
   if not strategy then
     return nil, err
   end
+
+  local custom_strat = fmt("kong.plugins.%s.strategies.%s.%s",
+                           plugin, db.strategy, schema.name)
+  local exists, mod = utils.load_module_if_exists(custom_strat)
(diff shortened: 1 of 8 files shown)


kong/db/dao/plugins.lua
```diff
@@ -156,6 +156,32 @@ local function load_plugin_entity_strategy(schema, db)
   if not strategy then
     return nil, err
   end
+
+  local custom_strat = fmt("kong.db.strategies.%s.%s", db.strategy, schema.name)
```
p0pr0ck5: actually im wondering if this should something like

```
local custom_strat = fmt("kong.plugins.%s.db.strategies.%s.%s",
plugin, db.strategy, schema.name)
```
?

hishamhm: I'd go without the `.db.`, since we don't use it for `migrations`:

```
local custom_strat = fmt("kong.plugins.%s.strategies.%s.%s",
plugin, db.strategy, schema.name)
```

(I think this was the path originally suggested by @thibaultcha too)

p0pr0ck5: Updated, thanks!

hishamhm:
I was going to suggest how to add a test for this but as I started to write the comment I realized it would be quicker to just write the test, so I took the liberty and pushed a commit to this PR :) (thanks @bungle for the comment that inspired it ;) )

PR #4519, opened 2019-04-16
tests(env) use new kong-ngx-build script to set up base components fo…

dndx:
…r testing

This PR uses the reusable https://github.com/Kong/openresty-build-tools tool for bootstrapping base environments. Alongside a few improvements to how caching is done in Travis.

The `setup_env.sh` file contains logic we copy over several other repos. This makes code reuse difficult. Also the `openresty-build-tools` tool correctly sets the concurrency level when running `make` and should make the compilation process slightly faster. It also supports producing debug builds which may be useful in the future.

Note that we got rid of the cache for `$DOWNLOAD_CACHE` in `.travis.yml` file. Caching the intermediate objects and source code provides little value for speeding up the compilation process, but significantly increases the cache size. According to Travis [doc](https://docs.travis-ci.com/user/caching/):

> Note that this makes our cache not network-local, it is still bound to network bandwidth and DNS resolutions. That impacts what you can and should store in the cache. If you store archives larger than a few hundred megabytes in the cache, it is unlikely that you’ll see a big speed improvement.

This change reduces the cache size from more than 3GB in `master` (and whooping 8GB in `next`) right now to around 300MB, which should make downloading, updating and unarchiving the cache much faster.

The updated `setup_env.sh` takes 300ms to complete when all dependencies has been cached already.

### DIFF
--- a/.ci/setup_env.sh
+++ b/.ci/setup_env.sh
@@ -1,112 +1,34 @@
 #!/usr/bin/env bash
-set -e
+# set -e
 
 #---------
 # Download
 #---------
 
 DEPS_HASH=$(cat .ci/setup_env.sh .travis.yml | md5sum | awk '{ print $1 }')
+BUILD_TOOLS_DOWNLOAD=$DOWNLOAD_ROOT/openresty-build-tools
 
-OPENSSL_DOWNLOAD=$DOWNLOAD_CACHE/$DEPS_HASH/openssl-$OPENSSL
-OPENRESTY_DOWNLOAD=$DOWNLOAD_CACHE/$DEPS_HASH/openresty-$OPENRESTY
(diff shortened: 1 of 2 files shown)


.ci/setup_env.sh
```diff
@@ -1,112 +1,34 @@
 #!/usr/bin/env bash
-set -e
+# set -e
```
hutchic: did you mean to nuke that?

dndx: Yes. This conflicts with the Travis exit handler and causes builds to fail without any error message shown. Which was a big pain while I was troubleshooting.

dndx:
Build succeeded: https://travis-ci.org/Kong/kong/builds/521931774?utm_source=github_status&utm_medium=notification, merged to `master`.

PR #4524, opened 2019-04-18
style(router) remove passing of ngx and localize functions and magic tables

bungle:
### Summary

Just a small style change on `router` code to get rid of `mocking` related parameters on public interfaces. I am fully ok, if there are objections at merging this.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -5,8 +5,13 @@ local bit           = require "bit"
 
 
 local hostname_type = utils.hostname_type
+local subsystem     = ngx.config.subsystem
+local get_method    = ngx.req.get_method
 local re_match      = ngx.re.match
 local re_find       = ngx.re.find
+local header        = ngx.header
+local var           = ngx.var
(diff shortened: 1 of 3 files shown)


kong/router.lua
```diff
@@ -5,8 +5,13 @@ local bit           = require "bit"
 
 
 local hostname_type = utils.hostname_type
+local subsystem     = ngx.config.subsystem
+local get_method    = ngx.req.get_method
```
hishamhm: One concern I have about this is that it is easy to add another localized `ngx.` function here and forget to add the equivalent code in `_set_ngx` below (and end up with a "partially mocked" `ngx`). If we are going the path of `_set_ngx`, I would rather declare these variables with no value, and then call `_set_ngx(ngx)` at the top-level of the module, to make it more evident that the initialization of those variables is special.

bungle: partially mocking well may be the intent of the user?

Issue #4504, opened 2019-04-12
Request Transformer remove header is case-sensitive


A-Juma:
### Summary

The request transformer plugin implementation has changed, in that it is now case sensitive when trying to remove headers

Example:

If you wanted to remove the Accept header from requests then the below configuration for request transformer would remove the Accept header (note it is configured in lowercase).
`
{
  "name": "request-transformer",
  "config": {
    "remove": {
      "headers": ["accept"]
    }
}
`

However if you configure it as **Accept** (with an uppercase character) it will not remove the header. 
`
{
  "name": "request-transformer",
  "config": {
    "remove": {
      "headers": ["Accept"]
    }
}
`

### Steps To Reproduce

1. Configure a route
2. Configure the request-transformer plugin to remove a header (Configure the header with the first character in uppercase (e.g. 'Accept))
3. Make a request with an 'Accept' header
4. The header will not be removed.

From an initial look, it appears the issue lies with the use of kong.request.get_headers() in the request-transformer, in that it returns all headers in lowercase. However the config for the plugin will have stored the header names in whatever case was used when applying the configuration.

### Additional Details & Logs

- Kong version (`1.0.3`)

RobSerafini:
I can reproduce this.

`http POST :8001/routes/anything/plugins name=request-transformer config.remove.headers\[\]=accept -f`

Header is removed

http POST :8001/routes/anything/plugins name=request-transformer config.remove.headers\[\]=Accept -f

Header will not be removed.

guanlan:
We are working on the fix.

locao:
Closing as it has been fixed here: e1d510987d0e5d8c71ad17ab149865de08ba1c17

PR #4526, opened 2019-04-18
fix(request-transformer) plugin is case insensitive for header names now

locao:
ngx_http_lua_module treats all header names as lower case strings, so
when changing something from request they must always be compared this
way.

Issue reported in https://github.com/Kong/kong/issues/4504

### DIFF
--- a/kong/plugins/request-transformer/access.lua
+++ b/kong/plugins/request-transformer/access.lua
@@ -113,6 +113,7 @@ local function transform_headers(conf)
 
   if remove then
     for _, name, _ in iter(conf.remove.headers) do
+      name = lower(name)
       if headers[name] ~= nil then
         headers[name] = nil
         clear_header(name)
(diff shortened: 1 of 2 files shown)


hishamhm:
@locao Please add a regression test in `spec/03-plugins/15-response-transformer/01-header_transformer_spec.lua`

kong/plugins/request-transformer/access.lua
```diff
@@ -142,7 +144,7 @@ local function transform_headers(conf)
 
   if replace then
     for _, name, value in iter(conf.replace.headers) do
-      if headers[name] ~= nil or lower(name) == "host" then
+      if headers[name] ~= nil or name == "host" then
```
hishamhm: Just noticed this: I think this `lower(name)` needs to stay, right? (it is taking the input from the user's conf, so that may include uppercase). The headers coming from ngx are the ones that are always normalized to lowercase.

locao: You're right, thanks for pointing that out.

PR #4528, opened 2019-04-19
feat(balancer) implement least-connections balancer

Tieske:
replaces #4393 

Since this relies on the new mock helpers, the PR is against `master`, so should NOT BE MERGED. Once the mock helpers get merged into `next` I'll rebase this PR against `next`.

But please do review (some relevant comments are left behind in #4393), please let me know how you want them fixed.

Additionally the name of the new test file is not ok, how to proceed with that? make it 11 and renumber all existing ones, or move it to the end? please advise.


Tieske:
@hishamhm @bungle continuing the [discussion here](https://github.com/Kong/kong/pull/4393#discussion_r266414103)

We could rename the entries for the `algorithm` field to: `rr-consistent`, `connections`. With potential future additions like `latency`. (using `rr` here to keep is short)

Also: input is needed on the name of the new test file...

spec/02-integration/05-proxy/10-balancer/02-least-connections_spec.lua
```diff
@@ -0,0 +1,147 @@
+local helpers = require "spec.helpers"
+
+-- create two servers, one double the delay of the other
+local fixtures = {
+  http_mock = {
+    lambda_plugin = [[
```
Tieske: rename: copy-paste thing

spec/02-integration/05-proxy/10-balancer/02-least-connections_spec.lua
```diff
@@ -0,0 +1,147 @@
...
+        end
+      end
+
+      -- start the threads
+      for i = 1, thread_max do
+        threads[#threads+1] = ngx.thread.spawn(handler)
+      end
+
+      -- wait while we're executing
+      local finish_at = ngx.now() + 5
```
Tieske: why not sleep 5?

Tieske: 1.5 should be save

Tieske:
naming `least-connections` and `consistent`

Tieske:
@hishamhm after discussion I tried to find slightly more intelligent back-off mechanisms. Problem is that most of those require application knowledge/client side behaviour.

To improve the following could be implemented;

1. apply a global retry-limit on a balancer (max 10 retries per second, or 200 per minute, etc). This would always be in effect.
2. Drop relative traffic (so 20% unhealthy means randomly dropping 20% of traffic) and stop handling retries all together. Applicable from a first threshold, eg. health drops below 70%.
3. Drop all traffic to stop the cascade. Applicable from a second threshold, eg. health drops below 50%.

This PR implements health metrics, and option 3 above. The other ones could be added later, since I consider the third one the most important (and most protective).

Tieske:
@hishamhm updated as discussed

hishamhm:
@Tieske there's a conflict now in the test file, could you rebase? thanks!

PR #4535, opened 2019-04-23
tests(log_spec) fix flapping test caused by timeout error log produce…

dndx:
…d by http-log not being considered in the test

Ran the Cassandra tests multiple times afterwards, no failures so far.

We should exclude the "timeout" error in the assertion. As `http-log`'s timeout was set to 1ms for this test and it could definitely happen.

### DIFF
--- a/spec/03-plugins/03-http-log/01-log_spec.lua
+++ b/spec/03-plugins/03-http-log/01-log_spec.lua
@@ -132,7 +132,7 @@ for _, strategy in helpers.each_strategy() do
           http_endpoint = "https://" .. helpers.mock_upstream_ssl_host
                                      .. ":"
                                      .. helpers.mock_upstream_ssl_port
-                                     .. "/delay/1",
+                                     .. "/delay/5",
           timeout = 1
         }
       }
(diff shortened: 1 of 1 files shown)


spec/03-plugins/03-http-log/01-log_spec.lua
```diff
@@ -281,7 +281,13 @@ for _, strategy in helpers.each_strategy() do
...
 
       for line in logs:gmatch("[^\r\n]+") do
-        assert.not_match("[error]", line, nil, true)
+        if not line:find("send(): failed request to " ..
+                         helpers.mock_upstream_ssl_host .. ":" ..
+                         helpers.mock_upstream_ssl_port .. ": timeout",
+                         0, true)
+        then
+          assert.not_match("[error]", line, nil, true)
+        end
```
thibaultcha: Based on the test's setup (timeout of 1 second, and posting the log to `/delay/1`, it actually sounds like we should assert that the timeout error _was_ actually hit and logged here. How about increasing the chances of hitting the timeout, and asserting the error is logged to the error log by way of `helpers.wait_until()` or even `kong stop` (to ensure logs are written to disk before reading them)?

dndx: Yes, @hishamhm suggested this during the standup as well. I will push a commit shortly.

PR #4540, opened 2019-04-25
feat(dao) introduce the no_broadcast_crud_event option

p0pr0ck5:
### Summary

This commit introduces the 'quiet' flag in the options table passed to DAO CRUD methods. When set, the quiet flag skips broadcasting a local 'dao:crud' worker event, which in turn generates a cluster event
regarding the entity, as well as a local cache invalidation of the entity.

This mimics the 'quiet' functionality present in the legacy DAO layer.

### Full changelog

* reintroduce the quiet option

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -407,9 +407,9 @@ local function find_cascade_delete_entities(self, entity)
 end
 
 
-local function propagate_cascade_delete_events(entries)
+local function propagate_cascade_delete_events(entries, options)
   for _, entry in ipairs(entries) do
-    entry.dao:post_crud_event("delete", entry.entity)
+    entry.dao:post_crud_event("delete", entry.entity, nil, options)
   end
 end
 
@@ -596,7 +596,7 @@ local function generate_foreign_key_methods(schema)
           return nil, err, err_t
         end
 
-        self:post_crud_event("update", row, rbw_entity)
+        self:post_crud_event("update", row, rbw_entity, options)
 
         return row
       end
@@ -632,7 +632,7 @@ local function generate_foreign_key_methods(schema)
           return nil, err, err_t
         end
 
-        self:post_crud_event("update", row)
+        self:post_crud_event("update", row, nil, options)
 
         return row
       end
@@ -676,8 +676,8 @@ local function generate_foreign_key_methods(schema)
           return nil, tostring(err_t), err_t
         end
 
-        self:post_crud_event("delete", entity)
-        propagate_cascade_delete_events(cascade_entries)
+        self:post_crud_event("delete", entity, nil, options)
+        propagate_cascade_delete_events(cascade_entries, options)
 
         return true
       end
@@ -855,7 +855,7 @@ function DAO:insert(entity, options)
     return nil, err, err_t
   end
 
-  self:post_crud_event("create", row)
+  self:post_crud_event("create", row, nil, options)
 
   return row
 end
@@ -893,7 +893,7 @@ function DAO:update(primary_key, entity, options)
     return nil, err, err_t
   end
 
-  self:post_crud_event("update", row, rbw_entity)
+  self:post_crud_event("update", row, rbw_entity, options)
 
   return row
 end
@@ -928,7 +928,7 @@ function DAO:upsert(primary_key, entity, options)
     return nil, err, err_t
   end
 
-  self:post_crud_event("update", row)
+  self:post_crud_event("update", row, nil, options)
 
   return row
 end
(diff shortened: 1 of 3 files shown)


thibaultcha:
Can we use a more explicit name for this option than the one we used last time? Like `no_crud_broadcast` or similar.

bungle:
> Like no_crud_broadcast or similar.

Perhaps `propagate = false`.

p0pr0ck5:
Pushed up another commit; the option is `broadcast_crud_event` and the check looks for an explicit boolean `false` (likely the value would be `nil` if `options` is present).

thibaultcha:
That sounds like a worst API design than “`falsy` as default” to me, but why not.

p0pr0ck5:
Inverted the logic ^

kikito:
Approved. Please cleaup the commit history and rename field in the PR title / message before merging.

p0pr0ck5:
@kikito I don't have the permissions to merge this :D

p0pr0ck5:
Thanks @thibaultcha!

Regarding the ACL DAO, I suspect we should update this DAO override? I don't see any other overrides of `post_crud_event`, so would it be safe to update this function signature for third-parties, given that it's not a breaking change?

thibaultcha:
@p0pr0ck5 Yes, updating the function signature sounds ok to me. It's not ideal but I _highly_ doubt any other DAO out there overrides `post_crud_event` at this point.

hishamhm:
Since this is a new DAO feature, this PR should be rebased to target the `next` branch.

kong/db/dao/init.lua
```diff
@@ -754,7 +754,7 @@ function DAO:insert(entity, options)
     return nil, err, err_t
   end
 
-  self:post_crud_event("create", row)
+  self:post_crud_event("create", row, options)
```
bungle:
```suggestion
  self:post_crud_event("create", row, nil, options)
```

kong/db/dao/init.lua
```diff
@@ -846,7 +846,7 @@ function DAO:upsert(primary_key, entity, options)
     return nil, err, err_t
   end
 
-  self:post_crud_event("update", row)
+  self:post_crud_event("update", row, options)
```
bungle:
```suggestion
  self:post_crud_event("update", row, nil, options)
```

kong/db/dao/init.lua
```diff
@@ -890,8 +890,8 @@ function DAO:delete(primary_key, options)
     return nil, tostring(err_t), err_t
   end
 
-  self:post_crud_event("delete", entity)
-  propagate_cascade_delete_events(cascade_entries)
+  self:post_crud_event("delete", entity, options)
```
bungle:
```suggestion
  self:post_crud_event("delete", entity, nil, options)
```

kong/db/dao/init.lua
```diff
@@ -556,8 +556,8 @@ local function generate_foreign_key_methods(schema)
           return nil, tostring(err_t), err_t
         end
 
-        self:post_crud_event("delete", entity)
-        propagate_cascade_delete_events(cascade_entries)
+        self:post_crud_event("delete", entity, options)
```
bungle:
```suggestion
        self:post_crud_event("delete", entity, nil, options)
```

kong/db/dao/init.lua
```diff
@@ -512,7 +512,7 @@ local function generate_foreign_key_methods(schema)
           return nil, err, err_t
         end
 
-        self:post_crud_event("update", row)
+        self:post_crud_event("update", row, options)
```
bungle:
```suggestion
        self:post_crud_event("update", row, nil, options)
```

kong/db/dao/init.lua
```diff
@@ -267,9 +267,9 @@ local function find_cascade_delete_entities(self, entity)
 end
 
 
-local function propagate_cascade_delete_events(entries)
+local function propagate_cascade_delete_events(entries, options)
   for _, entry in ipairs(entries) do
-    entry.dao:post_crud_event("delete", entry.entity)
+    entry.dao:post_crud_event("delete", entry.entity, options)
```
bungle:
```suggestion
    entry.dao:post_crud_event("delete", entry.entity, nil, options)
```

thibaultcha:
@p0pr0ck5 Will you have some time to address @bungle's review and rebase this against `next` in the next couple of days? We have plans for 1.2.0rc1 to ship on Monday. Let us know if not, thanks!

p0pr0ck5:
@thibaultcha sorry for the delay on this. I've been stretched thin and likely won't have bandwidth to complete this before next Monday (and I am out next week as well). I also wonder if there's more work that needs to be done here, particularly for `snis` objects which also leverage `self:post_crud_event`. Haven't had time to look closely at it unfortunately.

p0pr0ck5:
I will try to make these changes tomorrow on my flight home, or next week after the holiday!

hishamhm:
Please hold merging this PR for a bit, I will push tests to it first.

hishamhm:
Merged manually into `release/1.2.0` for inclusion in 1.2.0rc2!

PR #4541, opened 2019-04-25
feat(db): allow swagger path parameters

ealogar:
### Summary

Allow '$', '{' and '}' as allowed characters in upstream paths as they are used in swagger
to represent path variables.
This is not agains RFC 3986 and allow the use of swagger syntax for defining routers

### Full changelog

* [Feat] - Allow swagger path variables when defining upstreams in routes

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -56,7 +56,8 @@ end
 
 
 local function validate_path(path)
-  if not match(path, "^/[%w%.%-%_~%/%%]*$") then
+  -- Accept '$''{''}' and all characters defined in RFC 3986
+  if not match(path, "^/[%w%.%-%_~%/%%${}]*$") then
     return nil,
            "invalid path: '" .. path ..
            "' (characters outside of the reserved list of RFC 3986 found)",
@@ -304,7 +305,7 @@ typedefs.run_on_first = Schema.define {
 
 typedefs.tag = Schema.define {
   type = "string",
-  required = true, 
+  required = true,
   match = "^[%w%.%-%_~]+$",
 }
 


thibaultcha:
I do not understand the purpose of this change. I do understand Swagger syntax and path variables, but defining a Route with a path such as `/foo/{var}`, means that you would have to send a request with the following URL to match it: `/foo/%7Bvar%7D` (at least, that is how most clients would encode it), and I doubt that this is the intent you had in mind. Were you attempting to perform path parameter extraction maybe?

As for the specifics of the validation rules regarding this property: it currently only accepts "unreserved" characters from RFC 3986:
```
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
```

`{` and `}` are considered "unwise" characters as per [RFC 2996](https://www.ietf.org/rfc/rfc2396.txt), and as such, must be escaped. As for `$`, it is considered a reserved character (sub-delim) by [RFC 3986 2.2](https://tools.ietf.org/html/rfc3986#section-2.2), and may be found in application specific use-cases, although I've never encountered one.

That said, we could make the validation rules more lenient, and support matching on a wider character-set (regardless of their percent-encoded representation in the HTTP request). But that is another topic, as it appears this PR's intent is something else, which is still obscure to me.

A good way to show the intent of your PR (and a requirement to see it merged) is to write test cases for it :)

ealogar:
@thibaultcha I will try writing some test to ilustrate the use of this.
E.G: you may add a route with this path /user/${user_id}/data and some plugin or even the upstream may understand user_id as a path variable that should be replaced with appropiate data (e.g, a header value or the claim of a JWT)
Give me tiem to write tests

kikito:
Hi, I am closing this issue because of lack of activity. If you think it is still unresolved, please reopen with the requested information. Thanks!

Tieske:
@ealogar if you can give this another spin, I'd be interested in this as well

PR #4543, opened 2019-04-25
docs(contributing) add code style guidelines on quotes usage

thibaultcha:


CONTRIBUTING.md
```diff
@@ -646,6 +646,27 @@ local t = { foo = "hello", bar = "world" }
...
+```lua
+-- bad
+local str = 'hello'
+
+-- good
+local str = "hello"
+```
+
+If a string contains double quotes, **do** favor long bracket strings:
+
```
Tieske: Nitpick; can we make it a bit more generic?

**do** prefer alternative quoting to over escape sequences:

```lua
-- bad
local str = "message: \"hello\""

-- good
local str = 'message: "hello"'
local str = [[message: "hello"]]
```

If preferred, I'll create a PR for this

thibaultcha: The above guideline states that we should keep away from single quotes. Hence, better not giving an example with single quotes considered “good” just below, that is the whole point why long bracket strings are preferred (in the codebase so far but also in those guidelines): as little code as possible uses single quotes, thus reducing confusion around single vs double quotes and when to use one vs the other. 
The cases for single quotes (to avoid some of the semantics of long bracket strings) are very rare.

Tieske: Should have been more clear: the text I proposed was the second option, so first one stays: double quotes. But then add the exception that other quoting is preferred over escape sequences.

The reason to go for single ones is that they have the same escaping rules as double ones. The long bracket strings differ in their escaping, and hence take more cognitive energy. But that is rather opinionated.

Feel free to leave it at this, since it was a nitpick to start with 😄 .

PR #4547, opened 2019-04-26
feat(*) better consistency of DNS resolution error messages

thibaultcha:
Additions:

- Log the `try_list` JSON string upon DNS resolution failures in the
  Cassandra connector.

Changes for consistency:

- All DNS resolution failures contain the string "DNS resolution
  failed" which is something end-users can understand better than
  implementation details (e.g. "`toip()` called failed"). It can also be
  searched for much more easily.
- The Cassandra connector log is prefixed with the strategy name,
  `[cassandra]`, as all other CLI error messages already are.
- Cosockets DNS resolution failures are prefixed with `[cosocket]`,
  which is much easier to search for, and more obvious for seasoned users
  or maintainers. Since the "DNS resolution failed" string is also in
  the message, we consider this message insightful enough for end-users.
- The balancer's logs don't use a prefix (alas), so no prefix was added
  to the balancer's DNS resolution failure log.

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -97,10 +97,10 @@ function CassandraConnector.new(kong_config)
     local dns = dns_tools(kong_config)
 
     for i, cp in ipairs(kong_config.cassandra_contact_points) do
-      local ip, err = dns.toip(cp)
+      local ip, err, try_list = dns.toip(cp)
       if not ip then
-        log.error("could not resolve Cassandra contact point '%s': %s",
-                  cp, err)
+        log.error("[cassandra] DNS resolution failed for contact " ..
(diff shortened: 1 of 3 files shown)


thibaultcha:
@Tieske Postgres will resolve once we entered the driver already (via our cosocket's globalpatches), so alas, there is no simple way of catching the error and prefixing it, unless we perform DNS resolution ahead of time (outside the scope of this PR, since it already is the current behavior). It will currently show as a `[cosocket]` error.

thibaultcha:
@p0pr0ck5 It's possible, indeed. Not sure if reasonable though, so I dismissed that thought. We have changed error logs formats or messages countless times already. Our dependencies and libraries could even change their error messages at anytime as well. The `feat` commit scope ensures this change will appear in the Changelog, at the very least.

PR #4550, opened 2019-04-27
feat(db) add index on psql cluster_events expire_at

p0pr0ck5:
### Summary

The cleanup mechanism for cluster events on Postgres uses an on-insert trigger. In busy cases, this cleanup function can cause an unnecessary strain on the database.

### Full changelog

* add index on psql cluster_events expire_at


p0pr0ck5:
cc @RobSerafini

kong/db/migrations/core/004_110_to_120.lua
```diff
@@ -0,0 +1,16 @@
...
+      BEGIN
+        CREATE INDEX IF NOT EXISTS cluster_events_expire_at_idx ON cluster_events(expire_at);
+      EXCEPTION WHEN UNDEFINED_COLUMN THEN
+        -- Do nothing, accept existing state
+      END$$;
+    ]],
+  },
+
+  cassandra = {
+    up = [[]],
```
Tieske: is this empty entry necessary? can't it be left out all together?

p0pr0ck5: Migrations have their own schema, which requires this field: https://github.com/Kong/kong/blob/master/kong/db/schema/others/migrations.lua

PR #4551, opened 2019-04-28
feat(db) query semaphore for Postgres connector

p0pr0ck5:
### Summary

This commit introduces an optional semaphore locking mechanism within the Postgres connector :query() method. This provides a solution to the longstanding problem of OpenResty not providing a ceiling on the number of connections that can be established to a remote TCP host. As Postgres is sensitive about the number of connections opened to a given server, this provides a workaround solution until a native connection cap implementation is made available for OpenResty.

This commit introduce two new config options, 'pg_semaphore_max' and 'pg_semaphore_timeout'. The former defines the number of resources allocated to a `ngx.semaphore` instance that is bound to a Postgres connector object. This effectively defines the maximum number of concurrent queries to a Postgres server that can be issued from a given worker process (the more natural number of maximum number of queries issued by a given Kong node can then be extrapolated by multiplying this value by the number of Kong worker processes - using the native ngx semaphore mechanism seems the more appropriate solution vs. creating a node-wide semaphore (via some messy shared dict approach) that provided a slightly more human-natural accounting).

The default 'pg_semaphore_max' value of 0 results in no semaphore object created in the connector, and no lock acquisition performed before executing a query. This leaves the existing behavior in place.

### Full changelog

* add semaphore object to postgres connector
* post/wait on semaphore resources within `:query()`
* add unit and integration tests for semaphore resource acquisition timeout behavior (returning `nil, err`)


bungle:
Looks great! Thanks @p0pr0ck5!

kong.conf.default
```diff
@@ -447,6 +447,16 @@
                                  # See the `lua_ssl_trusted_certificate`
                                  # setting to specify a certificate authority.
 
+#pg_semaphore_max = 0            # Sets the number of semaphore resources made
+                                 # available to the semaphore used to limit
+                                 # concurrent connections to the Postgres
+                                 # database. The default value of 0 removes the
+                                 # semaphore concurrency limiter.
```
thibaultcha: Should we rename this property to something more explicit for end-users? E.g. `pg_max_concurrent_queries` or so. Ultimately, "semaphore" is an implementation detail, and "max semaphore resources" does not provide a hint on what said resources are used for.

Also, we should note in the comment that this property will limit concurrent queries/connections on a per-worker basis only.

kong.conf.default
```diff
@@ -447,6 +447,16 @@
...
+
+#pg_semaphore_timeout = 60       # Defines the timeout, in seconds, after which
+                                 # Postgres query semaphore resource acquisition
+                                 # attempts will fail.
```
thibaultcha: We should describe what happens when resource acquisition fails. Limiting the number of connections to Postgres isn't a hard problem because of implementation (e.g. semaphore which was suggested a while ago already, or native connection queuing in OpenResty 1.15.8.1), but because of the behavior to follow in unhappy cases. The problem isn't solved by connection queuing, but only "moved". Two options:

1. Failing hard, and rejecting the PostgreSQL query, which could result in an HTTP 503 error from the Gateway. Imho, most users would not be ok with such as behavior. An easily overwhelmed API Gateway seems to be antithetical with the sheer concept of an API Gateway.
2. Gracefully fail, and proceed with the query nonetheless to ensure resiliency. Which completely defeats the purpose for such a limit in the first place.

It is my understanding that users who wish for the second behavior can achieve it with `pg_semaphore_max = 0` (i.e. not providing a limit in the first place). Nonetheless, when a limit is provided, better be upfront about what the potential downsides of it are.

kong/db/strategies/postgres/connector.lua
```diff
@@ -371,9 +372,46 @@ function _mt:setkeepalive()
 end
 
 
+function _mt:acquire_lock()
+  if get_phase() == "init" or get_phase() == "init_worker" then
```
thibaultcha: [wait()](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/semaphore.md#wait) can only be called from contexts that can yield, so I think more restrictions should be applied here. Also, better cache the result of `get_phase()`, especially if we add more checks.

kong/db/strategies/postgres/connector.lua
```diff
@@ -371,9 +372,46 @@ function _mt:setkeepalive()
 end
 
 
+function _mt:acquire_lock()
```
thibaultcha: I suggest that we rename this function to something else, such as `acquire_query_resource` or something similar. Otherwise, using `lock` makes it sound like we are employing a some sort of mutex to execute queries.

kong/db/strategies/postgres/connector.lua
```diff
@@ -371,9 +372,46 @@ function _mt:setkeepalive()
...
+  if get_phase() == "init" or get_phase() == "init_worker" then
+    return true
+  end
+
+  if not self.sem then
```
thibaultcha: We should perform this check before the `get_phase()` checks, to avoid unnecessary CPU cycles when `pg_semaphore_max = 0`.

kong/db/strategies/postgres/connector.lua
```diff
@@ -996,24 +1037,36 @@ local _M = {}
...
   }
 
   local db = pgmoon.new(config)
 
+  local sem
+  if config.sem_max > 0 then
+    local err
+    sem, err = semaphore.new(config.sem_max)
+    if not sem then
+      error("Failure creating Postgres connector semaphore: " .. err)
```
thibaultcha: Note: all error messages in Kong are supposed to be lower-case. I also suggest that we do not prevent initialization in case the semaphore allocation fails, but that we instead issue a `CRIT` error message. The reason being that in case of a change in `ngx.semaphore` semantics in the future, or other unexpected failure scenarios, we avoid a situation in which many users are unable to even _start_ their Kong instances. Since a lack of semaphore will not prevent Kong from running, but will prevent it from running in _optimal_, _predictable_ conditions, a `CRIT` log is appropriate.

Note: other instantiation failures such as `kong.dao` or `kong.cache` should ideally make Kong's initialization fail, but currently do not, alas, due to the worker-events library only being instantiated in `init_worker`. Something that has been communicated about and is awaiting for a fix to initialize worker-events in `init`. Currently, such failures to initialize singletons are logged a `CRIT`, but really should be logged with `EMERG`.

kong/conf_loader.lua
```diff
@@ -119,6 +119,8 @@ local CONF_INFERENCES = {
   pg_password = { typ = "string" },
   pg_ssl = { typ = "boolean" },
   pg_ssl_verify = { typ = "boolean" },
+  pg_semaphore_max = { typ = "number" },
+  pg_semaphore_timeout = { typ = "number" },
```
thibaultcha: I suggest ensuring that none of these values can be negative, which would result in failures to initialize the semaphore. Along with updates to the `conf_loader` unit tests suite to ensure that is the case.

p0pr0ck5: Ah yeah, I missed the `conf_loader` changes. Fixed up!

kong.conf.default
p0pr0ck5: I had conflicting thoughts, and the details are spelled out in the initial commit message. The config here is indeed part of an implementation detail, but that's a detail that users need to be aware of. `pg_max_concurrent_queries` as a config option, to me, would read as the number of concurrent queries that a Kong node could issue, which is not accurate. Something more accurate like `pg_max_worker_concurrent_queries` seems like a mouthful, and is a bit clunky- it almost reads like it's talking about a pg worker, which isn't really a thing.

So, I'd vote we keep the names relevant to the implementation detail to avoid confusion, and spell out well (with examples, diagrams, etc) the impacts of various configs in the documentation.

kong.conf.default
p0pr0ck5: I definitely do think we should detail this out very well. I'm not sure how far into detail to go here. The online docs seems like a better place to go into depth about various impacts and expected behaviors (as well as the current default ones). To me trying to explain it all in the reference config file would be too wordy and make this file unwieldy. That said this could go into a bit more detail about what would happen, although that occurs on a case by case basis, so the explanation here would be a bit vague.

To be super clear, when I talk about spelling things out in docs, I mean in a separate guide, not just the config file reference page at https://docs.konghq.com/1.1.x/configuration/

p0pr0ck5:
> OpenResty 1.15.8.1 ships with connection queuing for cosockets. Have you given any thoughts as to what those configuration properties will become if/once we switch our underlying implementation?

I had thought about just going with this approach for this patch series. That would require updating pgmoon and lua-cassandra to leverage that new API. I think ultimately it's a good long-term goal, but I'm not sure what the timeline would be to add that. Once upon a time we had a pgmoon fork IIRC- not sure we want to go down that route again?

> Should we issue a warning when pg_semaphore_max > lua_socket_pool_size? There is no value afaict in configuring Kong as such, but the performance penalty will be visible.

There is a [Kong-configurable value](https://github.com/Kong/kong/blob/master/kong.conf.default#L709) for this. However, I'm not sure issuing a warning is warranted or correct. I'm not sure what the performance penalty would be, apart from needing to open up new TCP connections. While there is overhead there, I don't know that I would call it a penalty, especially in the context of the queries that Kong runs (we have seen `cluster_events` inserts take several hundred milliseconds to execute on a loaded server - this is a pathological case and was solved in part by #4550, but the few-milliseconds overhead of opening a new TCP connection is not of the same magnitude). These are advanced options for operators that need to tune Kong, and issuing a warning might just cause more confusion? The way to silence that warning would be to adjust `lua_socket_pool_size`, which would affect all socket pools by default, not just the postgres driver pool.

kong/db/strategies/postgres/connector.lua
p0pr0ck5: In generally, the caller (`:query`) can also only be called in functions that yield, or init/init_worker. Given that we don't make restrictions in which phases `:query` can be called, is a separate check needed here as well? And, should that check live here, or in `:query`? I'd be happy to make a separate patchset to do some more phase sanity checking in `:query`, that sounds like a good defensive step.

kong.conf.default
thibaultcha: My thinking here was to use `pg_max_concurrent_queries` and specify in the comment for this property that this is a per-worker value. There is precedent here already with the `lua_socket_pool_size` property (which is, alas, not specified as being "per-worker" which I believe is an oversight).

kong.conf.default
thibaultcha: Indeed. My comment here was only to add a brief note: "Failure to acquire a resource can result in a failure to proxy a request. Find out more in the xyz guide at ...". It is really important we at least hint at the downsides of this model in this file.

thibaultcha:
> That would require updating pgmoon and lua-cassandra to leverage that new API.

I see no reason why either of those projects would reject a patch for connection queuing support in OpenResty >= 1.15.8.1! Our pgmoon fork has always been a temporary one, for at times it has been necessary to ship patches faster than they were merged upstream. We can go down the same road again if needed.

> There is a [Kong-configurable value](https://github.com/Kong/kong/blob/master/kong.conf.default#L709) for this.

I know, hence my comment. There is more overhead than a few ms when you take, e.g. low bandwidth networks or TLS-enabled database connections. A warning could also be a `notice` log. But I'm just as fine with not issuing any kind of log, especially when considering that the property affects all socket pools.

kong.conf.default
p0pr0ck5: I've updated the variable name per this suggestion. To me it makes the following `pg_semaphore_timeout` name a bit incongruous as it also doesn't hide an implementation detail? But maybe it's worth the detail here? :man_shrugging:

kong.conf.default
p0pr0ck5: I've updated the variable name per this suggestion. To me it makes the following `pg_semaphore_timeout` name a bit incongruous as it also doesn't hide an implementation detail? But maybe it's worth the detail here? :man_shrugging:

kong.conf.default
p0pr0ck5: Added in a bit more detail, with a hanging reference to the docs (once the docs PR is merged I will add a link here)

p0pr0ck5:
> I see no reason why either of those projects would reject a patch for connection queuing support in OpenResty >= 1.15.8.1!

Yes, I don't think that these projects shouldn't be patched to use this, only that the approach here is a first step. We can push this out and then keep the same `pg_max_concurrent_queries` config value, just with a different implementation, if/when pgmoon uses native queueing support.

kong.conf.default
```diff
@@ -447,6 +447,23 @@
                                  # See the `lua_ssl_trusted_certificate`
                                  # setting to specify a certificate authority.
 
+#pg_max_concurrent_queries = 0   # Sets the maximum number of concurrent queries
+                                 # that can be executing at any given time. This
+                                 # limit is enforced per worker process; the
+                                 # total number of queries concurrenct queries
+                                 # that can execute is equal to this value,
```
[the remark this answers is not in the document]
p0pr0ck5: Yes I completely butchered the writing here :upside_down_face: fixed!

kong.conf.default
```diff
@@ -447,6 +447,23 @@
...
+                                 # multiplied by the number of worker processes.
+                                 # The default value of 0 removes this
+                                 # concurrency limitation.
+
+#pg_semaphore_timeout = 60       # Defines the timeout, in seconds, after which
+                                 # Postgres query semaphore resource acquisition
+                                 # attempts will fail. Such failure will
```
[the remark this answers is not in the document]
p0pr0ck5: Yep, typo. thanks!

kong/db/strategies/postgres/connector.lua
```diff
@@ -996,24 +1043,36 @@ local _M = {}
...
   }
 
   local db = pgmoon.new(config)
 
+  local sem
+  if config.sem_max > 0 then
+    local err
+    sem, err = semaphore.new(config.sem_max)
+    if not sem then
+      ngx.log(ngx.CRIT, "failure creating Postgres connector semaphore: " .. err)
```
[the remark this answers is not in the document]
p0pr0ck5: Fixed!

kong/db/strategies/postgres/connector.lua
```diff
@@ -391,9 +436,11 @@ function _mt:query(sql)
   end
 
   if res then
+    self:release_query_semaphore_resource()
```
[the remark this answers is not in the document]
p0pr0ck5: Ah yes, my bad :)

kong/db/strategies/postgres/connector.lua
thibaultcha: The paradox here comes from the fact that `:query` is compatible in non-yieldable contexts (header_filter, body_filter, log...) with the magical fallback to LuaSocket. In those cases, the call to this semaphore will throw a Lua error, which is what I was concerned about and was asking for a way to cover those cases, for the programmer's sake (it is not a mistake to call this semaphore in those phases if `:query` is supported in those phases).

However, `:query` should _not_ be supported in those phases. There is a pending work item which is to *unload* LuaSocket once initialization has completed, in order to ensure that we can never block the NGINX runloop unknowingly (better fail hard and early than letting such a fallback ship to users and destroy performance under unknown cases. E.g. in a plugin with a particular configuration or other such twisted cases that make it hard to catch in pre-release benchmarking and smoke testing).

kong.conf.default
```diff
@@ -447,6 +447,22 @@
...
+                                 # total number of concurrent queries for this node
+                                 # will be will be `pg_max_concurrent_queries * nginx_worker_processes`.
+                                 # The default value of 0 removes this
+                                 # concurrency limitation.
+
+#pg_semaphore_timeout = 60       # Defines the timeout, in seconds, after which
+                                 # Postgres query semaphore resource acquisition
+                                 # attempts will fail. Such failures will
+                                 # generally result in the associated proxy or
+                                 # Admin API request failing with an HTTP 500.
```
thibaultcha: This makes me think that a desired user behavior _could be_ for Kong to catch such acquisition failure errors and exit with HTTP 503. Speculating here but at least such a behavior would be predictable for administrators and more explicit for clients than HTTP 500 (and more correct).

thibaultcha: (Yes, I did mean for the PostgreSQL connector to catch those failures and bluntly do something like: `if in_proxy then return ngx.exit(503)`)

p0pr0ck5: This feels like a reeeeally big hammer.

Do we even have such a `in_proxy` context/function? Also, why wouldn't we want to apply the same logic to failed acquisitions on the Admin API?

thibaultcha: Yes to the Admin API. No to the CLI or in-tests queries. I wonder how such "expected unexpected behavior" would be perceived by users, and if this would get any adoption at all in its current state.

p0pr0ck5: > if this would get any adoption at all in its current state.

Is the implication here that no one would use this feature at all unless semaphore acquisition failures returned a 503 instead of a 500? That seems like a really big stretch (apologies if I'm misreading).

I can make the change if it's a hard blocker for you. It seems like a big hammer and it's not the route I would take, but I'll defer to you here. Let me know.

thibaultcha: No, I don't have string feelings about it. Also got carried away a bit: even in its current state, we consistently return an error and the HTTP 500 will be predictable (arguably not an ideal status code but again, no strong feelings here), so I got unnecessarily worried.

thibaultcha:
@p0pr0ck5 Oh, something else: I think this should go to `next`. Mind rebasing your branch?

p0pr0ck5:
@thibaultcha done, thanks!

kong/db/strategies/postgres/connector.lua
hishamhm: @thibaultcha do you still want this restriction to be tightened prior to merging? Otherwise I think this is good to go.

thibaultcha: I'd say it is fine for now, at least a minor concern in comparison to the HTTP 500 error we return to clients (vs. a more appropriate 503 status code imho). What do you think?

kong.conf.default
```diff
@@ -447,6 +447,22 @@
                                  # See the `lua_ssl_trusted_certificate`
                                  # setting to specify a certificate authority.
 
+#pg_max_concurrent_queries = 0   # Sets the maximum number of concurrent queries
+                                 # that can be executing at any given time. This
+                                 # limit is enforced per worker process; the
+                                 # total number of concurrent queries for this node
+                                 # will be will be `pg_max_concurrent_queries * nginx_worker_processes`.
```
thibaultcha: It'd be nice to stick to the 80 chars limit in this file

p0pr0ck5: I want to say I found a few existing exceptions and based my thoughts on this. Let me 3x check and adjust if I'm off base

thibaultcha: The only exception we let slip through has been for links (or _should_ still be, if other oversights slipped through...)

p0pr0ck5: Yep. In this case moving the string "\`pg_max_concurrent_queries * nginx_worker_processes\`." still results in exceeding the limit based on how far we're indented. I (or the merger) can split this into multiple lines, but that feels a little awkward.

kong/db/strategies/postgres/connector.lua
hishamhm: > What do you think?

I agree with the current state of the PR in which it does not add the check in this function, as it is only called by `:query` and the real issue is where can `:query` be called. If I were to suggest anything, it would be to make the acquire and release functions `local function`s to the module passing an explicit `self` (instead of having them `_mt:` functions) to make their private status more evident. But this is not a blocker.

thibaultcha:
Leaving a comment here to express my thoughts on the discussion regarding error handling in case of semaphore resource acquisition failure.

### 500 vs 503

The current proxy implementation combined with this patch will make Kong return HTTP 500 errors when a semaphore resource cannot be acquired. The semantics of a 500 error (read: "an _unexpected_ error") don't really apply to this particular failure:

1. Said failure is (at least theoretically) predictable: there exist a predetermined number of clients allows to connect concurrently to PostgreSQL. Trying to open more connections will produce a very much expected error.
2. Additionally, we should convey to the Kong clients that the request failed at this given point in time, but that said request may succeed if retried, hence why "HTTP 503".

At this point, I think that this point wasn't a disagreement topic in our 1-1 discussions @p0pr0ck5 and @hishamhm.

However, there are valid implementation concerns:

- A "catch-all" `if no_semaphore_resource then ngx.exit(503)` in the DAO is bad separation of concerns.
- Updating all proxy-path DAO error handling code to return 503 on this exception isn't really an option either.

### DAO abstraction layer for the proxy path

This conversation has strengthened my opinion that there should exist an abstraction layer to use the DAO on the proxy path (or other parts of the codebase). There exist at least 3 behavorial differences when using the DAO in the proxy path:

1. Proxy, Admin API, and migrations/CLI have different consistency requirements for C* read queries (availability vs consistency tradeoffs)
2. Proxy and Admin API have different "global" options, such as `options.nulls`
3. Proxy/Admin API and other users of the DAO have different error handling policies (e.g. proxy handles "semaphore resource acquisition" by responding with HTTP 503)

### Path forward for this PR

Let's merge this patch as-is and consider the above concerns outside of the scope of this particular PR. Those concerns are a pressing topic though, particularly because of the point stressed out in this comment (on returning 500 not being an ideal situation).

thibaultcha:
With the above said, I will pull this and merge it now!

thibaultcha:
Merged!

PR #4555, opened 2019-04-29
fix(migrations) fix migrations related error messages; migrations up -f fix; migrations finish -f addition

bungle:
### Summary

This PR is meant to fix issues reported here:
https://discuss.konghq.com/t/automating-database-migration-for-a-cd-pipeline/2920

Schema state checking is improved on:
- `kong migrations list`
- `kong start`
- and kong's init

This PR now also includes two extra commits:
1. `fix(migrations) migrations up with -f will also run the executed teardowns`
2. `feat(migrations) migrations finish with -f will also run the executed migrations`

### DIFF
--- a/kong/cmd/migrations.lua
+++ b/kong/cmd/migrations.lua
@@ -1,6 +1,7 @@
 local DB = require "kong.db"
 local log = require "kong.cmd.utils.log"
 local tty = require "kong.cmd.utils.tty"
+local meta = require "kong.meta"
 local conf_loader = require "kong.conf_loader"
 local kong_global = require "kong.global"
 local migrations_utils = require "kong.cmd.utils.migrations"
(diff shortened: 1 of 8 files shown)


kong/cmd/start.lua
```diff
@@ -30,20 +34,44 @@ local function execute(args)
...
+      if schema_state.missing_migrations then
+        log.warn("database is missing some migrations:\n%s",
+                 schema_state.missing_migrations)
+      end
 
-      else
-        migrations_utils.print_state(schema_state)
+      if schema_state.pending_migrations then
+        log.info("database has pending migrations:\n%s",
+                 schema_state.pending_migrations)
```
thibaultcha: I am not a fan of duplicating this logic in both the CLI _and_ in the `init_by_lua` phase. Could we keep it exclusively in `init_by_lua` instead? Also, the `init_by_lua` logic as per this patch isn't as precise when listing migrations state as this command is.

kong/cmd/start.lua
```diff
@@ -30,20 +34,44 @@ local function execute(args)
...
 
   local schema_state = assert(db:schema_state())
-
   local err
 
   xpcall(function()
     assert(prefix_handler.prepare_prefix(conf, args.nginx_conf))
 
+    if not schema_state:is_up_to_date() and args.run_migrations then
+      migrations_utils.up(schema_state, db, {
```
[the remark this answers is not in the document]
bungle: I suggest keeping it where it is, as the `up` that it is running can possibly change the schema state, e.g. being up to date.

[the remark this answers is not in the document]
thibaultcha: It looks like the below branch I was referring to was removed in a subsequent revision of the patch, marking this as resolved then

kong/db/init.lua
```diff
@@ -497,10 +498,12 @@ do
           end
 
           local state = "executed"
-          if strategy_migration.teardown then
-            -- this migration has a teardown step for later
-            state = "pending"
-            n_pending = n_pending + 1
+          if not suppress_teardown then
+            if strategy_migration.teardown then
```
thibaultcha: Could we make this a single branch for brevity's sake?
```lua
if not suppress_teardown and strategy_migration.teardown then
  -- ...
end
```

kong/cmd/migrations.lua
```diff
@@ -94,21 +95,78 @@ local function execute(args)
...
+    end
+
+    if schema_state.new_migrations then
+      log("%snew migrations available:\n%s", r, schema_state.new_migrations)
+      r = "\n"
     end
 
-    migrations_utils.print_state(schema_state)
+    if schema_state.pending_migrations and schema_state.new_migrations then
+      log("%s[warn] database has pending migrations from a previous upgrade, " ..
```
thibaultcha: There exists `log.warn` to set the proper logging level. In verbose mode, this would show up as `[notice] [warn]` (and logging level would not be respected for this warning).

bungle: I know, but it puts timestamp (I can revert to that though).

kong/cmd/utils/migrations.lua
```diff
@@ -145,6 +134,7 @@ local function up(schema_state, db, opts)
 
       assert(db:run_migrations(schema_state.executed_migrations, {
         run_up = true,
+        suppress_teardown = true,
```
thibaultcha: Why is `force` only executing the `up` state of migrations? Since executed migrations have previously ran both steps (`up` and `teardown`), saying that `force` "forces re-execution of these migrations" is incorrect. It will only force re-execution of the `up` step of these migrations, but the user may need to re-execute the `teardown` phase instead? I really do not like this `--force` option.

kong/cmd/utils/migrations.lua
```diff
@@ -175,7 +165,19 @@ end
...
+      return
+    end
+
+    if schema_state.legacy_is_014 then
+      log("cannot run migrations on a legacy %s; run 'kong migrations up' " ..
+          "to proceed", db.infos.db_desc)
+      return
+    end
+
+    log("cannot run migrations; run 'kong migrations bootstrap' instead")
```
thibaultcha: It is worth noting that those should be error logs (which will be written to stderr instead of stdout). This is why we call `error()` (instead of `log.err` + `return`). I see the below `opts.force` branch is already guilty of this

kong/cmd/utils/migrations.lua
```diff
@@ -87,38 +75,39 @@ local function up(schema_state, db, opts)
...
                 "on a fresh %s.", db.strategy, db.infos.db_desc,
-                db.infos.db_desc))
+                db.infos.db_desc, db.infos.db_desc), 2)
     end
 
     if schema_state.legacy_is_014 then
       local present, err = db:are_014_apis_present()
       if err then
-        error(err)
+        error(err, 2)
```
thibaultcha: Why are we increasing the stack level in those errors? In our CLI code, we never do so, as it allows us to accurately pinpoint where they were thrown. The trace doesn't appear for users, only when `--vv` is specified (and we are the ones usually having a look and we want to know where it was thrown from, precisely. Especially in such cases where the error isn't prefixed with any sort of static strings.

kong/cmd/migrations.lua
thibaultcha: Yes, better use `log.warn` here (the newline can be put after the last line of the previous call to `log()`). Another option is to subsequently patch the logging utility to output a non-verbose log on `warn` and above levels (it currently is the desired behavior).

kong/cmd/utils/migrations.lua
bungle: Do you mean `suppress_teardown`, Or do you mean that it only passes `run_up=true`. Do you mean that I should change what it originally does?

bungle: Yes, I can hear that it is confusing. I guess it could be because the `teardowns` are not as re-entrant as `ups`.

bungle: Though on the other hand, it can be simply wrong to not execute `teardown` too.

bungle: It was implemented here: https://github.com/Kong/kong/pull/4071

bungle: I have now changed `migrations up -f` and also added `migrations finish -f`

thibaultcha:
Manually merged!

PR #4538, opened 2019-04-24
feat(router) add path_separated property to service entity, fix #4469

bungle:
### Summary

Adds `path_separated` property to `service` entity to allow path handling.

**Description:**

This was reported here:
https://github.com/Kong/kong/issues/4469

In short the breaking change was introduced between 0.14 and 0.15:
https://github.com/Kong/kong/commit/5be033566c6fc35a7db8a64f61c4cc71bd01eba2

The problem is that current behavior depends on client and all the cases cannot be enforced on server.

e.g. take url:
1. http://httpbin.org/anything and
2. http://httpbin.org/anything/something and
3. http://httpbin.org/anything/

the first two work, but the last one doesn’t work.

right now we can create service with path
1. `/anything` or
2. `/anything/`.

The problem with first is that if you request `/something` you will endup at:
`/anythingsomething` while you wanted to end up at `/anything/something`

The problem with second is that if you request with `/` you will endup at:
`/anything/` while you wanted to end up at `/anything`

By setting the `service.path_separated=true` it will work as expected.

Reasoning, while trying to perform client intent as closely as possible, the behavior should also be enforceable on proxy 

### Issues resolved

Fix #4469

PR #4564, opened 2019-05-02
tests(router) fix slash handling integration tests

bungle:
### Summary

Fixes router slash handling integration tests spotted while doing the "waiting to be resolved": #4538.

### DIFF
--- a/spec/01-unit/08-router_spec.lua
+++ b/spec/01-unit/08-router_spec.lua
@@ -1859,15 +1859,23 @@ describe("Router", function()
 
       for i, args in ipairs(checks) do
 
-        local config = args[5] == true and "(strip = on, plain)" or "(strip = off, plain)"
+        local config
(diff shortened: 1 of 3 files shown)


thibaultcha:
What does this fix, exactly?

bungle:
> What does this fix, exactly?

See:
https://github.com/Kong/kong/pull/4564/files#diff-0c4a62cf167ef5217b1a0131a916c1adL1100

And:
https://github.com/Kong/kong/pull/4564/files#diff-0c4a62cf167ef5217b1a0131a916c1adL1075

The current test suite is only testing whether the code can create a function, it never executes the function.

thibaultcha:
Got it. It'd be great to make this more explicit in the future to avoid unnecessary back and forth. Descriptive commit messages are very important for reviews but also for future selves.

PR #4565, opened 2019-05-02
feat(cache) introduce cache warmup for entities and DNS

hishamhm:
Introduces a new configuration property, `db_cache_warmup_entities`, which allows the user to specify which entities should be warmed up at initialization. 

From the documentation at `kong.conf.default`:

```
#db_cache_warmup_entities = services, plugins
                                 # Entities to be pre-loaded from the datastore
                                 # into the in-memory cache at Kong start-up.
                                 # This speeds up the first access of endpoints
                                 # that use the given entities.
                                 #
                                 # When the `services` entity is configured
                                 # for warmup, the DNS entries for values in
                                 # its `host` attribute are pre-resolved
                                 # asynchronously as well.
                                 #
                                 # Cache size set in `mem_cache_size` should
                                 # be set to a value large enough to hold all
                                 # instances of the specified entities.
                                 # If the size is insufficient, Kong will log
                                 # a warning.
```

The default value is `services, plugins`: the `routes` entity does not need to be warmed up because it is cached via the in-memory router object; specifying this in this configuration property is a no-op.

Includes unit tests for the `kong.cache_warmup` module.

This is a port of work produced by @bungle and kikito in #4408, #4462, #4468, #4488. Thanks @thibaultcha for the preliminary feedback and suggestions on the property name and specifics of the behavior.

(Note that, at @thibaultcha's suggestion, this PR replaces the "cache prewarm" terminology for the more usual "cache warmup" in all commits.)

Closes #4408.

### DIFF
--- a/kong/cache_warmup.lua
+++ b/kong/cache_warmup.lua
@@ -2,23 +2,25 @@ local cache_warmup = {}
 
 
 local tostring = tostring
-local ipairs = ipairs
+local pairs = pairs
 local math = math
 local kong = kong
 local ngx = ngx
 
 
 local ENTITIES_TO_WARMUP = {
-  "services",
-  "plugins",
+  ["services"] = true,
+  ["plugins"] = true,
 }
 
 
-local function cache_warmup_single_entity(entity_name)
-  local dao = kong.db[entity_name]
-  if not dao then
-    return nil, "Invalid entity name found when warming up the cache: " .. tostring(entity_name)
-  end
+local function fail_cb()
+  error("this should never be called as L2 should already be warmed")
+end
+
+
+local function cache_warmup_single_entity(dao)
+  local entity_name = dao.schema.name
 
   ngx.log(ngx.NOTICE, "Preloading '", entity_name, "' into the cache ...")
 
@@ -36,9 +38,8 @@ local function cache_warmup_single_entity(entity_name)
       return nil, err
     end
 
-    local ok, err = kong.cache:get(cache_key, nil, function()
-      return entity
-    end)
+    -- NOTE: this is just for warming up L1
+    ok, err = kong.cache:get(cache_key, nil, fail_cb)
     if not ok then
       return nil, err
     end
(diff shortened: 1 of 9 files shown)


kong/cache_warmup.lua
```diff
@@ -0,0 +1,140 @@
...
+  local entity_name = dao.schema.name
+
+  ngx.log(ngx.NOTICE, "Preloading '", entity_name, "' into the cache ...")
+
+  local start = ngx.now()
+
+  local hosts, names, host_count
+  if entity_name == "services" then
+    hosts = {}
+    names = {}
```
kikito: `names` is a very bad name for this variable. I suggest replacing it with `visited_hosts`.

kong/cache_warmup.lua
```diff
@@ -0,0 +1,140 @@
...
+-- Loads entities from the database into the cache, for rapid subsequent
+-- access. This function is intented to be used during worker initialization.
+function cache_warmup.execute(entities)
+  -- kong.db and kong.cache might not be active while running tests
+  if not kong.db or not kong.cache then
+    return true
+  end
+
+  for _, entity_name in ipairs(entities) do
+    if entity_name == "routes" then
```
kikito: I'd rather not skip Routes silently. Either of these options would would do it for me:
* We allow the user to fill the cache with routes (maybe they do something with routes in a custom plugin?)
* We forbid them explicitly (by making them an invalid option in `conf.db_cache_warmup_entities`)

If none of these options is possible, we should at least put a `WARN` in the logs indicating that routes are never prewarmed.

kong/cache_warmup.lua
```diff
@@ -0,0 +1,140 @@
...
+      -- do not spend shm memory by caching individual Routes entries
+      -- because the routes are kept in-memory by building the router object
+      goto continue
+    end
+
+    local dao = kong.db[entity_name]
+    if not (type(dao) == "table" and dao.schema) then
```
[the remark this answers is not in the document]
hishamhm: It can't because the plugins (and therefore custom entities) are not loaded by the time `conf_loader` runs.

kong/cache_warmup.lua
bungle: Because it is an explicit setting, and not enabled by default, then I would go with:
```
We allow the user to fill the cache with routes (maybe they do something with routes in a custom plugin?)
```

bungle:
This is what we need to debate:
> #db_cache_warmup_entities = services, plugins

The `async` patches in its current form:
1. load services on init, and do cache active services on route rebuilds, also does DNS warmup when getting things from db, but also when getting events.
2. load all the plugins in memory

This kinda makes:
```
#db_cache_warmup_entities =
```

by default, otherwise we will loop through entities twice on start.

kong/cache_warmup.lua
```diff
@@ -0,0 +1,140 @@
...
+  ngx.log(ngx.NOTICE, "finished preloading '", entity_name,
+                      "' into the cache (in ", tostring(elapsed), "ms)")
+  return true
+end
+
+
+-- Loads entities from the database into the cache, for rapid subsequent
+-- access. This function is intented to be used during worker initialization.
+function cache_warmup.execute(entities)
+  -- kong.db and kong.cache might not be active while running tests
```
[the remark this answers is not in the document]
hishamhm: You tell me: this line was taken from a commit you wrote yourself. :) 
https://github.com/Kong/kong/pull/4408/commits/41cd363c670fac439817423fc1249884dff78ff5

kong/cache_warmup.lua
hishamhm: The current behavior was added at @thibaultcha's request. @thibaultcha, what do you say?

> If none of these options is possible, we should at least put a WARN in the logs indicating that routes are never prewarmed.

Adding `routes` to the list in the current behavior is harmless, so `warn` is IMO overreacting, but I wouldn't mind adding a `notice` level log.

hishamhm:
> Though, not sure where to discuss it, if not here.

@bungle I would say "not here", because these are changes that are not being made by this PR. We should focus on the code that's under review here: scope creep in the discussion is the path of PR Review Hell. Let's keep both the PRs and the discussion in manageable-sized pieces. (Any changes to this motivated by future PRs should be made in those future PRs — and in the particular cases you brought up, we are talking about a one-line change to the defaults, so it's absolutely not a problem to merge the defaults as-is here (as they make sense for this code) and change them later if it proves necessary.)

I'm not saying those aren't relevant topics, but we'll cross those bridges when we get there.

kong/cache_warmup.lua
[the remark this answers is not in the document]
kikito: I replaced it with `if not kong.cache then` later (maybe on a different branch) on and changed the comment to "during the init phase". That's why I know it's not needed. Please change that.

kong/cache_warmup.lua
hishamhm: In case we go down that route (no pun intended!), I already added the `kong.log.notice` log.

kikito:
Approved the changes as the only open one is the routes one and there's a log now.

PR #4571, opened 2019-05-06
Key authentication with signature validation

svenwal:
### Summary

This pull requests adds the optional function to verify a signature when using key-authentication.

When enabled each request needs an API key as well as a signature based on the api key, a key secret and the current timestamp. This provides an extra layer of security if an attacker sniffs the API key on the line.

The signature is created by generating a SHA256 hash of a concatenated string of api_key + key_secret + current_UTC_timestamp. Timestamp may be up to 5 minutes off

### Full changelog

* Added the "secret" property to the consumer/key_auth DAO and added migrations
* Added following parameters to key_auth plugin:
  * signature_names: similar to key_names an array of possible header/body_field names to search for the signature
  * signature_in_body: similar to key_in_body if the signature shall be searched in body
  * validate_signature: boolean value if signature check is required (false by default which results in same behaviour as previous plugin version)
* Checks on signature (validate_signature is set to true)
  * reject if empty
  * reject if signature does not meet the sha256 of api_key + key_secret + current_UTC_timestamp (timestamp +/- 5 minutes to current time)


Tieske:
Is this based on some standard?

p0pr0ck5:
Would the HMAC plugin not already satisfy what this PR tries to accomplish?

svenwal:
HMAC is close but different. It's about securing a auth-header and the integrity of the content where this patch is about making sure the apikey not being compromised / stolen. I have seen such checks at various gateway users in the past and know some gateways to it in this way.

kong/plugins/key-auth/handler.lua
```diff
@@ -148,6 +158,54 @@ local function do_authentication(conf)
     return nil, { status = 401, message = "No API key found in request" }
   end
 
+   -- search for signatire in headers & querystring
+   if conf.validate_signature then
+    for i = 1, #conf.signature_names do
```
Tieske: indentation is of here, only 1 space

kong/plugins/key-auth/handler.lua
```diff
@@ -148,6 +158,54 @@ local function do_authentication(conf)
...
+        -- duplicate API key
+        return nil, { status = 401, message = "Duplicate signature found" }
+      end
+    end
+
+    -- this request is missing a signature, HTTP 401
+    if not signature or signature == "" then
+      kong.response.set_header("WWW-Authenticate", _realm)
+    return nil, { status = 401, message = "No signature found in request" }
+  end
```
Tieske: bad indentation

kong/plugins/key-auth/handler.lua
```diff
@@ -90,13 +92,21 @@ local function do_authentication(conf)
     return nil, { status = 500, message = "Invalid plugin configuration" }
   end
 
+  if conf.validate_signature then
+    if type(conf.signature_names) ~= "table" then
+      kong.log.err("no conf.signature_names set, aborting plugin execution")
+      return nil, { status = 500, message = "Invalid plugin configuration" }
+    end
+  end
```
Tieske: I see you copied the one from above, but this should be part of the schema validation imo.

kong/plugins/key-auth/daos.lua
```diff
@@ -11,6 +11,7 @@ return {
       { created_at = typedefs.auto_timestamp_s },
       { consumer = { type = "foreign", reference = "consumers", default = ngx.null, on_delete = "cascade", }, },
       { key = { type = "string", required = false, unique = true, auto = true }, },
+      { secret = { type = "string", required = false, unique = true, auto = true }, },
```
Tieske: the `key` needs to be unique to identify a consumer, but the `secret` has no such constraint?

kong/plugins/key-auth/handler.lua
```diff
@@ -179,9 +237,43 @@ local function do_authentication(conf)
     return nil, { status = 500, message = "An unexpected error occurred" }
   end
 
-  set_consumer(consumer, credential)
+  if conf.validate_signature then
+    -- 4. Verify sha256
+    local secret = credential.secret
+    local now = math.floor(ngx.time())
```
Tieske: `ngx.time` already returns an integer value (as aopposed to `ngx.now`) so no need for the `math.floor` here.

kong/plugins/key-auth/handler.lua
```diff
@@ -179,9 +237,43 @@ local function do_authentication(conf)
     return nil, { status = 500, message = "An unexpected error occurred" }
   end
 
-  set_consumer(consumer, credential)
+  if conf.validate_signature then
+    -- 4. Verify sha256
```
Tieske: this comment seems a copy-paste left over.

Also this entire check should be moved up, before the `Success, this request is authenticated` comment.

kong/plugins/key-auth/handler.lua
```diff
@@ -179,9 +237,43 @@ local function do_authentication(conf)
...
+    end
 
-  return true
+    if verify_sig(now) then
+      -- authenticated by the current second
+      set_consumer(consumer, credential)
+      return true
+    end
+
+    for distance = 1, 300 do
```
Tieske: shouldn't this value of 300 be configurable as a "clock_skew" parameter?

kong/plugins/key-auth/handler.lua
```diff
@@ -179,9 +237,43 @@ local function do_authentication(conf)
...
+    local ks = key .. secret
+
+
+    local verify_sig = function(time)
+      sha:reset()
+      sha:update(ks)
+      sha:update(tostring(time))
+      local final = to_hex(sha:final())
+      --print("  verifying - sha(" .. ks .. " .. " .. tostring(time) .. ") = " .. final)
+      -- print(signature .. " == " .. final)
```
Tieske: cleanup

kong/plugins/key-auth/handler.lua
```diff
@@ -179,9 +237,43 @@ local function do_authentication(conf)
...
-  return true
+    if verify_sig(now) then
+      -- authenticated by the current second
+      set_consumer(consumer, credential)
+      return true
+    end
+
+    for distance = 1, 300 do
+      if verify_sig(now + distance) or verify_sig(now - distance) then
+        print(" ****  Verified")
```
Tieske: left-over debug statement

kong/plugins/key-auth/schema.lua
```diff
@@ -16,9 +16,17 @@ return {
               elements = typedefs.header_name,
               default = { "apikey" },
           }, },
+          { signature_names = {
+            type = "array",
+            required = true,
+            elements = typedefs.header_name,
+            default = { "signature" },
+        }, },
```
Tieske: indentation

thibaultcha:
@svenwal Such new features are to be opened against the `next` branch. Will you please rebase your PR on top of `next` and change the target branch? Many thanks.

kong/plugins/key-auth/handler.lua
```diff
@@ -90,13 +92,21 @@ local function do_authentication(conf)
     return nil, { status = 500, message = "Invalid plugin configuration" }
   end
 
+  if conf.validate_signature then
+    if type(conf.signature_names) ~= "table" then
+      kong.log.err("no conf.signature_names set, aborting plugin execution")
+      return nil, { status = 500, message = "Invalid plugin configuration" }
```
thibaultcha: It should be avoided to return HTTP 500 from a plugin if possible. Here, our posture is to heavily rely on schema validation to ensure that invalid plugin configurations are impossible at runtime.

If schema validation of the plugin's configuration cannot enforce such guarantees, a hard-coded default behavior in the plugin should be resilient (i.e. ignoring the feature altogether as if it was never configured, possibly with a warning log or other way to notify the operators).

In this case, let's try to rely more heavily on schema validation to ensure this type of mis-configuration is never encountered at runtime.

kong/plugins/key-auth/handler.lua
```diff
@@ -148,6 +158,54 @@ local function do_authentication(conf)
...
+        return nil, { status = 401, message = "Duplicate signature found" }
+      end
+    end
+
+    -- this request is missing a signature, HTTP 401
+    if not signature or signature == "" then
+      kong.response.set_header("WWW-Authenticate", _realm)
+    return nil, { status = 401, message = "No signature found in request" }
+  end
+  end
```
thibaultcha: This duplicates almost all of the above code to find the API key. Could you please try to factorise it in its own utility function instead, so we avoid duplicating such a large amount of (tricky) code? Thanks.

kong/plugins/key-auth/handler.lua
```diff
@@ -148,6 +158,54 @@ local function do_authentication(conf)
...
+
+
+
+
+
```
thibaultcha: style: please stay consistent in line jumps. Our code style has [specific guidelines](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#modules) about those and we strictly enforce them.

kong/plugins/key-auth/handler.lua
```diff
@@ -179,9 +237,44 @@ local function do_authentication(conf)
...
+    local verify_sig = function(time)
+      sha:reset()
+      sha:update(ks)
+      sha:update(tostring(time))
+      local final = to_hex(sha:final())
+      --print("  verifying - sha(" .. ks .. " .. " .. tostring(time) .. ") = " .. final)
+      -- print(signature .. " == " .. final)
+      return signature == final
+    end
```
thibaultcha: Closures don't play well with the JIT and this may negatively (and _significantly_) impact the performance of this plugin. Please find a way to write it in a more JIT-friendly way (avoiding closures, that is).

kong/plugins/key-auth/handler.lua
```diff
@@ -179,9 +237,44 @@ local function do_authentication(conf)
...
+      if verify_sig(now + distance) or verify_sig(now - distance) then
+        print(" ****  Verified")
+        -- authenticated with `distance` seconds
+        set_consumer(consumer, credential)
+        return true
+      end
+    end
+
+    return nil, { status = 500, message = "Invalid signature" }
+  else
```
thibaultcha: style: please avoid `else` branches if the previous branch already returns, as enforced in [conditional expressions](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#conditional-expressions) throughout the codebase.

kong/plugins/key-auth/migrations/init.lua
```diff
@@ -1,4 +1,5 @@
 return {
   "000_base_key_auth",
   "001_14_to_15",
+  "002_add_signature_validation"
```
thibaultcha: Please respect the naming scheme for new migrations: version to version, and not "feature-centric". All migrations required between two versions are now part of the same migration.

kong/plugins/key-auth/migrations/002_add_signature_validation.lua
```diff
@@ -0,0 +1,15 @@
...
+    ]],
+  },
+
+  cassandra = {
+    up = [[
+      ALTER TABLE IF EXISTS ONLY "keyauth_credentials"
+        ADD COLUMN "secret" text;
+    ]],
+  },
+}
```
thibaultcha: Please add new files to the rockspec, or installation will not be possible.

kong/plugins/key-auth/schema.lua
```diff
@@ -16,9 +16,17 @@ return {
...
           { hide_credentials = { type = "boolean", default = false }, },
           { anonymous = { type = "string", uuid = true, legacy = true }, },
           { key_in_body = { type = "boolean", default = false }, },
+          { signature_in_body = { type = "boolean", default = false }, },
+          { validate_signature = { type = "boolean", default = false }, },
```
thibaultcha: I think `verify` is a more common and accurate name than `validate` (e.g. "verify" an SSL certificate).

kong/plugins/key-auth/migrations/002_add_signature_validation.lua
```diff
@@ -0,0 +1,15 @@
...
+    up = [[
+      ALTER TABLE IF EXISTS ONLY "keyauth_credentials"
+        ADD COLUMN "secret" TEXT;
+    ]],
+  },
+
+  cassandra = {
+    up = [[
+      ALTER TABLE IF EXISTS ONLY "keyauth_credentials"
+        ADD COLUMN "secret" text;
```
thibaultcha: This is not a valid CQL statement, this migration will not run in Cassandra (hence the importance of integration tests in PRs!). `ALTER TABLE IF EXISTS` is not supported.

hishamhm:
@svenwal since this is a new feature, please `git rebase` this to `next` in your local branch, push it to the PR with `git push --force` and edit the PR settings so that the base branch is `next`. The PR currently has a conflict, some more might pop up when rebasing. Once the rebase and conflicts are done we will proceed with the review! Thanks!

svenwal:
Thank you - I hope to get some spare time before the end of the week. Already fixed some of the remarks on my system (like Cassandra migration) but other work is distracting me :)

hishamhm:
@svenwal ping :) any updates on this one?

svenwal:
Expect some updated soon - just got way to much work the last weeks but from Wednesday on it should get better again :)

kong/plugins/key-auth/daos.lua
svenwal: Yes, that's on purpose. A username (secret) needs to be unique, a password/secret does not need to be unique, right? I could even think of some kind of attack where someone create his own consumer (let's say in the developer portal) and tries all secrets and every time he gets a violation he already knows someone else has used this password / secret

kong/plugins/key-auth/handler.lua
svenwal: It's now signature_distance_seconds

kong/plugins/key-auth/daos.lua
Tieske: yes, but the secret currently has `unique = true`, so that should go?

svenwal: You are right - it's false now (in the next commit)

kong/plugins/key-auth/handler.lua
svenwal: Will be in next commit

kong/plugins/key-auth/handler.lua
svenwal: Fixed in next commit

kong/plugins/key-auth/migrations/002_130_to_140.lua
```diff
@@ -27,11 +27,22 @@ return {
...
+      END$$;
     ]],
   },
   cassandra = {
     up = [[
       ALTER TABLE keyauth_credentials ADD tags set<text>;
+
+
+      ALTER TABLE "keyauth_credentials" ADD "secret" text;
+      CREATE INDEX IF NOT EXISTS ON keyauth_credentials(secret);
```
p0pr0ck5: Why is there an index added here?

kong/plugins/key-auth/handler.lua
```diff
@@ -168,15 +129,45 @@ local function do_authentication(conf)
...
+    local secret = credential.secret
+    local now = ngx.time()
+    local sha = sha256:new()
+    local ks = key .. secret
+
+
+    local verify_sig = function(time)
+      sha:reset()
+      sha:update(ks)
+      sha:update(tostring(time))
```
p0pr0ck5: This construction is vulnerable to a length extension attack (https://en.m.wikipedia.org/wiki/Length_extension_attack). This should instead use HMAC, likely in the form of `hmac(secret, key .. time)`

hishamhm:
@svenwal There's a key-auth test still failing in the suite, can you check?

svenwal:
Yes, I have spent close to a full day to find and adopt the many already existing tests but did not find this one until now (it comes from the previous wish to not replicate code in the plugin so I created a function). Will check later

hishamhm:
@svenwal Ping? :)

hishamhm:
@svenwal This PR has stayed open with no activity for a couple of months by now — we often see it in our regular review of the PR queue, so we think we're going to close it if there's no further activity happening here, but feel free to reopen it if you'd like to resume working on this. How does that sound?

PR #4574, opened 2019-05-06
hotfix(core) address misc code review comments

hishamhm:
Addresses several code review comments left after some recent PRs were merged. Some important catches there, thanks @thibaultcha for the :eyes:!

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -45,7 +45,7 @@ local update_time  = ngx.update_time
 local subsystem    = ngx.config.subsystem
 local start_time   = ngx.req.start_time
 local clear_header = ngx.req.clear_header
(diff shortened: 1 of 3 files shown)


kong/runloop/handler.lua
```diff
@@ -173,7 +176,8 @@ do
     if not kong.cache then
       services_init_cache, err = build_services_init_cache(db)
       if err then
-        return nil, "could not build services init cache: " .. err
+        services_init_cache = {}
+        log(WARN, "could not build services init cache: " .. err)
```
thibaultcha: Better use the variadic form here, always (for the sake of referential code as well, or else such pattern rapidly spread)

thibaultcha:
@hishamhm Thanks for addressing those!

PR #4577, opened 2019-05-07
refactor(runloop) async groundwork

kikito:
This PR contains reordenings and refactors in preparation for the async reloading work.

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -115,11 +115,11 @@ local declarative_entities
 local schema_state
 
 
-local function execute_plugins(ctx, phase)
-  local plugins = runloop.get_plugins()
-  local phase_plugins = plugins.phases[phase]
+local function execute_plugins_plan(ctx, phase)
+  local plugins_plan = runloop.get_plugins_plan()
+  local phase_plugins = plugins_plan.phases[phase]
 
-  for plugin, configuration in plugins_iterator(ctx, phase, plugins) do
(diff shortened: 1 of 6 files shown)


kong/init.lua
```diff
@@ -199,16 +199,16 @@ end
 
 
 local function update_plugins()
-  local version, err = kong.cache:get("plugins_map:version", TTL_ZERO, utils.uuid)
+  local version, err = kong.cache:get("plugins:version", TTL_ZERO, utils.uuid)
```
hishamhm: Was there a practical necessity for this? I ask because cache keys for entities are often built as `<entity>:<data>`, so this ends up looking like a regular entity key (storing a `plugins` entity value), rather than a special key. I'd rather keep the `plugins:` namespace reserved for the `plugins` DAO.

bungle: plugins contains:
1. plugins.map
2. plugins.combos
3. plugins.loaded
4. plugins.phases
(and originally plugins.configured, which we dropped)

So plugins_map doesn't seem like a descriptive or best name for it anymore. Perhaps just call it plugins.version or plugins/version or plugins#version. But then it differs from router:version.

kikito:
I have made several change after a talk with @hisham. Highlights:

* Split a Plugin DAO custom function into two (explained reasoning in commit) and used it to load/reload on init and also to get the list of plugins (handlers) separatedly.
* Removed the init.after function from init
* Move DC and warmup back to init
* Rename `plugins` to `plugins_plan` everywhere (both code and cache key)

PR #4579, opened 2019-05-08
feat(dbless) enable atomic update with eviction control

hishamhm:
This PR allows performing DB-less updates atomically and with cache eviction control.

This is done by adding support for "double-buffering" in `kong.cache`: the cache object is able to maintain two lower-level mlcache instances internally, one being the "active page" and another being the "shadow page", writing to the shadow page, and flip between them in one go.

* we enable two cache pages when using DB-less mode
* we load the new configuration into the shadow page, and only if the entire declarative config is loaded successfully the cache pages are flipped.
* when the cache is not large enough to hold a declarative config input, the `/config` endpoint fails with HTTP 413 Request Too Large, and the previous DB-less configuration is retained
* the PR includes tests that demonstrate the HTTP 413 scenario, by using a small cache

Note that this enables two additional shms in the configuration template for the double-buffering (shadow hits and misses), but only when DB-less is used.

### DIFF
--- a/kong/cache.lua
+++ b/kong/cache.lua
@@ -10,7 +10,6 @@ local NOTICE  = ngx.NOTICE
 local DEBUG   = ngx.DEBUG
 
 
-local SHM_CACHE = "kong_db_cache"
 --[[
 Hypothesis
 ----------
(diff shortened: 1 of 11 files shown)


kong/cache.lua
```diff
@@ -226,14 +245,21 @@ function _M:get_bulk(bulk, opts)
 end
 
 
-function _M:safe_set(key, value)
+function _M:safe_set(key, value, shadow_page)
```
kikito: I'm not a super fan of boolean parameters, and I would have added two new functions (`shadow_safe_set(key, value)` and `shadow_purge()`). But this is not critical.

hishamhm: I know you're not :) I have considered this before, but then I'd have to either duplicate code or make a private function that the two public functions dispatch to. Both seemed overkill, given the limited use. I did use a `SHADOW = true` constant in the calling code to make use of the shadow-page more obvious.

Issue #4379, opened 2019-03-11
rate-limiting plugin works randomly (in redis/cluster mode) 


wbarantt:
### Summary
The rate-limiting plugin fails to throttle when configured in redis/cluster mode.
It seems that under a load the kong node does not decrease the limit and/or does not respond with the X-RateLimit-* headers.

### Steps To Reproduce
1. Create a test service/route pair with a rate-limiting plugin running in redis or cluster mode
2. Perform a N synchronous GET requests on that endpoint
3. Check if all the responses contain X-RateLimit-* headers
4. Check if the remaining limit has been decreased by N

### Additional Details & Logs
- Kong version 1.0.2

It seems that the code defined outside of RateLimitingHandler:access is sometimes just not run.
It affects both the RateLimitingHandler:header_filter (which results in no rate-limiting headers being sent in the response) and the RateLimitingHandler:log (which results in no limit being consumed by the request).

h3xar0n:
Hi @wbarantt , could you please share the following:
* the environment in which you're running Kong
* `kong.conf` values
* Kong error logs

wbarantt:
It's just stock a stock kong with no custom configuration.
I have configured a per consumer rate-limitimg on my test route and started the GET traffic.

No errors in the logs.
If I break the redis host/port configuration then it will fail with 500 or return 200 (depending on the fault tolerant setting) so it does not seem to be a problem with kong<>redis communication.
It works fine if there's no other traffic (going through another route) but once the number of requests increases the plugin execution seems to become disordered.

p0pr0ck5:
@wbarantt we'd still like to know a bit more about your Kong configuration :) Are you deploying a cluster of Kong nodes behind an LB? Can you please share the exact plugin config you've enabled? And how many distinct consumers are sending requests to Kong when you observe this behavior?

TBH this sounds a bit like https://github.com/Kong/kong/issues/3329, but we'd love to learn more about your use case to get a better understanding of what's going on. Thanks!

wbarantt:
This is my whole test config:
[kong_config.txt](https://github.com/Kong/kong/files/2961217/kong_config.txt)

This is how I test it:
for i in `seq -w 1 20`;do echo -e "RUN:\t=====\t#${i}\t====="; curl -sD - -o /dev/null -H "Host:httpbin.org" http://localhost:8000/httpbin_01/headers;done

This is what I get with no other traffic coming to this kong node:
RUN:    =====   #11     =====
[...]

PR #4581, opened 2019-05-08
feat(plugins) log request TLS version

subnetmarco:
The following PR logs the TLS version used by the request into the `basic` serializer for every logging plugin.

### DIFF
--- a/kong/plugins/log-serializers/basic.lua
+++ b/kong/plugins/log-serializers/basic.lua
@@ -1,4 +1,5 @@
 local tablex = require "pl.tablex"
+local ngx_ssl = require "ngx.ssl"
 
 local _M = {}
 
@@ -17,6 +18,12 @@ function _M.serialize(ngx)
     }
   end
 
+  local request_tls
+  local request_tls_ver = ngx_ssl.get_tls1_version_str()
+  if request_tls_ver then 
+    request_tls = { version = request_tls_ver } 
+  end
+  
   local request_uri = var.request_uri or ""
 
   return {
@@ -26,7 +33,8 @@ function _M.serialize(ngx)
       querystring = req.get_uri_args(), -- parameters, as a table
       method = req.get_method(), -- http method
       headers = req.get_headers(),
-      size = var.request_length
+      size = var.request_length,
+      tls = request_tls
     },
     upstream_uri = var.upstream_uri,
     response = {
--- a/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
+++ b/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
@@ -7,7 +7,7 @@ local TCP_PORT = 35001
 
 for _, strategy in helpers.each_strategy() do
   describe("Plugin: tcp-log (log) [#" .. strategy .. "]", function()
-    local proxy_client
+    local proxy_client, proxy_ssl_client
 
     lazy_setup(function()
       local bp = helpers.get_db_utils(strategy, {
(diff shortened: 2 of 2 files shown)


kong/plugins/log-serializers/basic.lua
```diff
@@ -26,7 +26,10 @@ function _M.serialize(ngx)
       querystring = req.get_uri_args(), -- parameters, as a table
       method = req.get_method(), -- http method
       headers = req.get_headers(),
-      size = var.request_length
+      size = var.request_length,
+      tls = {
+        version = ctx.KONG_TLS_VERSION
```
subnetmarco: For some reason this value is always `nil`, even when properly being set in `certificate.lua` with:

```lua
ngx.ctx.KONG_TLS_VERSION = ver
```

bungle: It reminds me of this:
https://github.com/Kong/kong/issues/4379

bungle: Perhaps contact `upstream` and ask if `ngx.ctx` should work with `ssl_certicate`:
https://github.com/openresty/lua-nginx-module#ngxctx

On my testing `rewrite` resets `ngx.ctx`.

subnetmarco:
This is ready to merge (after review), but we can go back at it and use the original implementation (with context setting in `ssl_certificate_by_lua*`) once https://github.com/openresty/lua-nginx-module/pull/1514 has been merged.

p0pr0ck5:
Should we have a test that defines the behavior for plaintext requests? So we don't break the log format in the future.

subnetmarco:
@p0pr0ck5 done

kong/plugins/log-serializers/basic.lua
```diff
@@ -26,7 +28,10 @@ function _M.serialize(ngx)
       querystring = req.get_uri_args(), -- parameters, as a table
       method = req.get_method(), -- http method
       headers = req.get_headers(),
-      size = var.request_length
+      size = var.request_length,
+      tls = {
```
hishamhm: this will always send an empty `tls` array even in non-tls requests. If this is okay, we leave it as is, otherwise we change it to `tls = request_tls_ver and {` (to only add the sub-table when sending TLS info

spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
```diff
@@ -80,6 +81,9 @@ for _, strategy in helpers.each_strategy() do
       -- Making sure it's alright
       local log_message = cjson.decode(res)
       assert.equal("127.0.0.1", log_message.client_ip)
+
+      -- Since it's over HTTP, let's make sure there are no TLS information
+      assert.is_nil(log_message.request.tls.version)
```
hishamhm: If the change described above is applied, we change this to `assert.is_nil(log_message.request.tls)`

kong/plugins/log-serializers/basic.lua
subnetmarco: I don't know, what do you think? Should we keep it empty or not?

subnetmarco:
@hishamhm I have addressed your review.

spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
subnetmarco: Done

hishamhm:
@subnetmarco Thanks! Merged!

PR #4585, opened 2019-05-08
tests(admin) run tests for routes in dbless mode

hbagdi:
- Simplify the test file to setup Kong only once for each strategy to
  speed up tests
- Add `off` strategy for this file. Kong Ingress Controller will start
  using these endpoints once 1.2 in db-less mode. The tests will ensure
  that we don't run regress in future.

### DIFF
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -17,7 +17,9 @@ if [ "$TEST_SUITE" == "integration" ]; then
     eval "$TEST_CMD" spec/02-integration/
 fi
 if [ "$TEST_SUITE" == "dbless" ]; then
-    eval "$TEST_CMD" spec/02-integration/02-cmd spec/02-integration/05-proxy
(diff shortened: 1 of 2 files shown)


hishamhm:
@hbagdi to enable a testsuite file for DB-less on Travis, edit `.ci/run_tests.sh` and add it to the list in the DB-less `eval` entry: https://github.com/Kong/kong/blob/5c9b8bc3e020912edc09855e349281086ece6935/.ci/run_tests.sh#L20 — We're running only a subset of the full testsuite with DB-less on Travis, that is the line that decides what to run.

hishamhm:
@hbagdi while you're there, `spec/02-integration/04-admin_api/15-off_spec.lua` would be safe to add too :sweat_smile:

hbagdi:
@hishamhm Done!

TIL we do that in our CI.

hishamhm:
@hbagdi Thanks!

PR #4588, opened 2019-05-08
fix(migrations) oauth2 tokens ttl was incorrectly migrated, fix #4572

### LINKED ISSUE #4572: migration from 0.14 oauth2 token lost
i successfully run migration from 0.14 to latest kong 1.1.2 all seems fine when i look in db oauth2 tokens are intact, but after few minutes 1 or 2 it deletes all the tokens which has expiration: 0 and only leaves tokens which have expiration more then 0.

i am running in docker, only 2 containers are running kong and kong-db no requests are made to kong after migration i just wait and it deletes tokens

bungle:
### Summary

Fixes an issue reported at #4572, where migrations could set a 0 ttl for `oauth2_tokens` that have never expiring `0` tokens, leading to deletion of such tokens.

### Issues resolved

Fix #4572

### DIFF
--- a/kong/plugins/oauth2/migrations/001_14_to_15.lua
+++ b/kong/plugins/oauth2/migrations/001_14_to_15.lua
@@ -36,7 +36,8 @@ return {
       END$$;
 
       UPDATE "oauth2_tokens"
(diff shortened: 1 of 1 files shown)


hishamhm:
@bungle Approved — thinking out loud: is this something we could easily test in kong-upgrade-tests?

PR #4590, opened 2019-05-09
perf(*) plugins handlers do not have to inherit from BasePlugin anymore

thibaultcha:
Plugins handlers have always been inheriting from the `BasePlugin`
class, which is itself inheriting from the `Object` class provided by
`kong.vendor.classic`.

This design predates even the very first public release of Kong in 2015,
and was influenced by the OOP background of its original authors.

Today, the case can be made against this design:

- At best, it makes instances with `debug` log levels extremely noisy.
- At worst, it has a non-negligible performance impact due to the
  constant metatable lookups of a plugin's handlers methods and
  because of the non-JITable `ngx.log` calls sitting on hot code paths.
- All in all, its only purpose was to provide a `_name` attribute which
  was only used by the `BasePlugin` parent class and redundant with the
  `handler.name` property injected by the plugins loader.

This commit removes the need for plugins handlers to inherit from the
`BasePlugin` class, but still ensures backwards-compatibility for
plugins doing so.
Fixture plugins in our test suites were not updated and still inherit
from the `BasePlugin` class as a means to maintain regression tests for
this backwards-compatible behavior.

The `BasePlugin` is *not* deprecated by this commit.

In some initial benchmarks (sample size 3), this change increased
req/s by up to 6% when using a single plugin (key-auth). The more
plugins are configured for a request, the higher the gains.

### DIFF
--- a/kong/db/dao/plugins.lua
+++ b/kong/db/dao/plugins.lua
@@ -2,6 +2,7 @@ local constants = require "kong.constants"
 local utils = require "kong.tools.utils"
 local DAO = require "kong.db.dao"
 local plugin_loader = require "kong.db.schema.plugin_loader"
+local BasePlugin = require "kong.plugins.base_plugin"
 
 
 local Plugins = {}
@@ -269,7 +270,14 @@ function Plugins:load_plugin_schemas(plugin_set)
     local handler, err = load_plugin(self, plugin)
 
     if handler then
-      handlers[plugin] = handler()
+      if getmetatable(handler) == BasePlugin then
+        -- Backwards-compatibility for 0.x and 1.x plugins inheriting from the
+        -- BasePlugin class.
+        -- TODO: deprecate & remove
+        handler = handler()
+      end
+
+      handlers[plugin] = handler
 
     else
       errs = errs or {}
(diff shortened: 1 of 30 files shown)


kong/init.lua
```diff
@@ -182,15 +182,15 @@ local function build_new_plugins(version)
   for _, plugin in ipairs(loaded_plugins) do
     if new_plugins.combos[plugin.name] then
       for phase_name, phase in pairs(new_plugins.phases) do
-        if plugin.handler[phase_name] ~= BasePlugin[phase_name] then
+        if type(plugin.handler[phase_name]) == "function"
```
hishamhm: just to make sure I follow: why was this change necessary, does any plugin declare a non-function field with a name clashing with a phase handler?

thibaultcha: The check could have been `if plugin.handler[phase_name] ~= nil`, but this is safer (most likely, plugins handlers should be subject to schema validation, just like migrations are, for instance).

If a handler is not implemented, the `plugin.handler[phase_name] ~= BasePlugin[phase_name]` condition would have been valid, and the phase incorrectly enabled by our runloop.

hishamhm: ahhh got it

PR #4592, opened 2019-05-10
feat(admin) add memory statistics to '/status' + fix(core) reset 'kong' shm on reload

thibaultcha:
## fix: reset `kong` shm on reload

This is a fix that is long overdue, and would cause the results of the `/` Admin API endpoints to show PRNG seeds of previous NGINX workers (previously terminated).

## feat(admin) add memory statistics to '/status'

The response of the `/status` endpoint now contains a `memory` property
which contains:

* `lua_shared_dicts` (hash map)
    * `capacity`: the total capacity of the shared memory zone
    * `allocated_slabs`: the currently allocated memory slabs (suffixed
      "slabs" to insist on the NGINX slab allocator particularity, which
      should be further detailed in the documentation for this
      endpoint)
* `workers_lua_vms` (array)
    * `pid`: PID of the worker process
    * `allocated_gc`: Lua VM allocated GC memory as per the return value of
    `collectgarbage("count")` (sampled every 10s as per 6de1aa0)
  
Additionally, the endpoint supports two optional querystring arguments:

* `unit`: can be one of `b/B`, `k/K`, `m/M`, `g/G`, which will return
   results in bytes, kibibytes, mebibytes, or gibibytes,
   respectively. Defaults to `k`.
* `scale`: the number of digits to the right of the decimal points when
  in the new human-readable memory strings. Defaults to `2`.

Example response for the `/status` endpoint:
    
    {
        "database": {
            "reachable": true
        },
            "memory": {
                "lua_shared_dicts": {
                    "kong": {
                        "allocated_slabs": "0.04 MiB",
                        "capacity": "5.00 MiB"
                    },
                    "kong_cluster_events": {
                        "allocated_slabs": "0.04 MiB",
                        "capacity": "5.00 MiB"
                    },
                    "kong_db_cache": {
                        "allocated_slabs": "0.77 MiB",
                        "capacity": "128.00 MiB"
                    },
                    "kong_db_cache_miss": {
                        "allocated_slabs": "0.08 MiB",
                        "capacity": "12.00 MiB"
                    },
                    "kong_healthchecks": {
                        "allocated_slabs": "0.04 MiB",
                        "capacity": "5.00 MiB"
                    },
                    "kong_locks": {
                        "allocated_slabs": "0.06 MiB",
                        "capacity": "8.00 MiB"
                    },
                    "kong_process_events": {
                        "allocated_slabs": "0.04 MiB",
                        "capacity": "5.00 MiB"
                    },
                    "kong_rate_limiting_counters": {
                        "allocated_slabs": "0.08 MiB",
                        "capacity": "12.00 MiB"
                    },
                    "prometheus_metrics": {
                        "allocated_slabs": "0.04 MiB",
                        "capacity": "5.00 MiB"
                    }
                },
                "workers_lua_vms": [
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16069"
                },
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16070"
                },
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16068"
                },
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16073"
                },
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16071"
                },
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16072"
                },
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16074"
                },
                {
                    "allocated_gc": "0.02 MiB",
                    "pid": "16067"
                }
                ]
            },
            "server": {
                "connections_accepted": 1,
                "connections_active": 1,
                "connections_handled": 1,
                "connections_reading": 0,
                "connections_waiting": 0,
                "connections_writing": 1,
                "total_requests": 1
            }
    }


kong/api/routes/kong.lua
```diff
@@ -129,8 +168,51 @@ return {
...
+        local count, err = ngx.shared.kong:get("kong:mem:" .. pid)
+        if err then
+          ngx.log(ngx.ERR, "could not get Lua VM allocated memory (pid: ",
+                           pid, "): ", err)
+        end
+
+        if count then
+          table.insert(status_response.memory.workers_lua_vms, {
+            pid = pid,
+            allocated = utils.bytes_to_str(count, unit, scale)
```
thibaultcha: Might be more descriptive if renamed to `gc_allocated`.

kong/init.lua
```diff
@@ -115,6 +115,31 @@ local declarative_entities
 local schema_state
 
 
+local reset_kong_shm
+do
+  local preserve_keys = {
+    "events:requests",
+    "kong:node_id",
```
thibaultcha: Need to discuss how to handle declarative config here.

spec/02-integration/02-cmd/03-reload_spec.lua
```diff
@@ -81,9 +82,78 @@ describe("kong reload", function()
...
+      local res = assert(client:get("/"))
+      local body = assert.res_status(200, res)
+      local json = cjson.decode(body)
+      local prng_seeds_2 = json.prng_seeds
+      client:close()
+
+      for k in pairs(prng_seeds_1) do
+        assert.is_nil(prng_seeds_2[k])
+      end
+  end)
```
thibaultcha: Need to handle `database = off` in this test case (https://travis-ci.org/Kong/kong/jobs/530544583).

kong/init.lua
hishamhm: Nothing needs to be done. :tada: If we make it remove the `declarative_config:loaded` key, then `Kong.init()` will naturally reload the declarative config YAML file from `declarative_config` in `kong.conf`, just as we wanted.

hishamhm: FWIW, I just tested this locally in `database=off` with and without a `declarative_config` entry, by adding `ngx.shared.kong:delete("declarative_config:loaded")` at the beginning of `Kong.init()`.

With a `declarative_config` entry:

1. `KONG_DATABASE=off KONG_DECLARATIVE_CONFIG=kong.yml KONG_PREFIX=foo kong start`
2. `sed -i 's/my-service/mon-service/g' kong.yml`
3. `kong reload -p foo`
4. `http :8001/services | grep mon-service`

Without a `declarative_config` entry, it still preserves whatever is in memory, which is nice:

1. `KONG_DATABASE=off KONG_PREFIX=foo kong start`
2. `http :8001/config config=@kong.yml`
3. `kong reload -p foo`
4. `http :8001/services | grep mon-service`

(Now that I think of it, these would be nice test cases to have)

PR #4594, opened 2019-05-10
refactor(runloop) plugins_iterator refactors

kikito:
This PR contains 3 refactors:

* Merges plugins_plan and plugins_iterator into a single entity
* Replaces the `repeat ... until` fake loop inside plugins_iterator by `goto`s
* Removes unnecessary metatables in plugins iterator

These refactors will be used as a base for async work

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -77,7 +77,6 @@ local kong_resty_ctx = require "kong.resty.ctx"
 local certificate = require "kong.runloop.certificate"
 local concurrency = require "kong.concurrency"
 local cache_warmup = require "kong.cache_warmup"
-local plugins_iterator = require "kong.runloop.plugins_iterator"
 local balancer_execute = require("kong.runloop.balancer").execute
 local kong_error_handlers = require "kong.error_handlers"
 local migrations_utils = require "kong.cmd.utils.migrations"
@@ -142,11 +141,11 @@ do
 end
 
 
-local function execute_plugins_plan(ctx, phase)
-  local plugins_plan = runloop.get_plugins_plan()
-  local phase_plugins = plugins_plan.phases[phase]
+local function execute_plugins_iterator(ctx, phase)
+  local plugins_iterator = runloop.get_plugins_iterator()
+  local phase_plugins = plugins_iterator.phases[phase]
 
-  for plugin, configuration in plugins_iterator(ctx, phase, plugins_plan) do
+  for plugin, configuration in plugins_iterator:iterate(ctx, phase) do
     if phase_plugins[plugin.name] then
       kong_global.set_named_ctx(kong, "plugin", configuration)
       kong_global.set_namespaced_log(kong, plugin.name)
@@ -221,10 +220,10 @@ local function load_declarative_config(kong_config, entities)
   end
 
   if not kong_config.declarative_config then
-    -- no configuration yet, just build empty plugins plan
-    local ok, err = runloop.build_plugins_plan(utils.uuid())
+    -- no configuration yet, just build empty plugins iterator
+    local ok, err = runloop.build_plugins_iterator(utils.uuid())
     if not ok then
-      error("error building initial plugins plan: " .. err)
+      error("error building initial plugins iterator: " .. err)
     end
     return true
   end
@@ -246,9 +245,9 @@ local function load_declarative_config(kong_config, entities)
     kong.log.notice("declarative config loaded from ",
                     kong_config.declarative_config)
 
-    ok, err = runloop.build_plugins_plan(utils.uuid())
+    ok, err = runloop.build_plugins_iterator(utils.uuid())
     if not ok then
-      error("error building initial plugins plan: " .. err)
+      error("error building initial plugins iterator: " .. err)
     end
 
     assert(runloop.build_router(kong.db, "init"))
@@ -384,9 +383,9 @@ function Kong.init()
     end
 
   else
-    local ok, err = runloop.build_plugins_plan("init")
+    local ok, err = runloop.build_plugins_iterator("init")
     if not ok then
-      error("error building initial plugins: " .. err)
+      error("error building initial plugins: " .. tostring(err))
     end
 
     assert(runloop.build_router(db, "init"))
@@ -492,10 +491,10 @@ function Kong.init_worker()
 
 
   -- run plugins init_worker context
-  runloop.update_plugins_plan()
-  local plugins_plan = runloop.get_plugins_plan()
-  local phase_plugins = plugins_plan.phases.init_worker
-  for _, plugin in ipairs(plugins_plan.loaded) do
+  runloop.update_plugins_iterator()
+  local plugins_iterator = runloop.get_plugins_iterator()
+  local phase_plugins = plugins_iterator.phases.init_worker
+  for _, plugin in ipairs(plugins_iterator.loaded) do
     if phase_plugins[plugin.name] then
       kong_global.set_namespaced_log(kong, plugin.name)
       plugin.handler:init_worker()
@@ -511,13 +510,13 @@ function Kong.ssl_certificate()
 
   runloop.certificate.before(ctx)
 
-  local ok, err = runloop.update_plugins_plan()
+  local ok, err = runloop.update_plugins_iterator()
   if not ok then
-    ngx_log(ngx_CRIT, "could not ensure plugins plan is up to date: ", err)
+    ngx_log(ngx_CRIT, "could not ensure plugins iterator is up to date: ", err)
     return ngx.exit(ngx.ERROR)
   end
 
-  execute_plugins_plan(ctx, "certificate")
+  execute_plugins_iterator(ctx, "certificate")
 end
 
 function Kong.balancer()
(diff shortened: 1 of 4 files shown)


kong/init.lua
```diff
@@ -219,10 +214,10 @@ local function load_declarative_config(kong_config, entities)
   end
 
   if not kong_config.declarative_config then
-    -- no configuration yet, just build empty plugins plan
-    local ok, err = runloop.build_plugins_plan(utils.uuid())
+    -- no configuration yet, just build empty plugins iterator
+    local ok, err = runloop.build_plugins_iterator(utils.uuid())
```
bungle:
```suggestion
    local ok, err = runloop.build_plugins_iterator("init")
```

kong/init.lua
```diff
@@ -244,9 +239,9 @@ local function load_declarative_config(kong_config, entities)
     kong.log.notice("declarative config loaded from ",
                     kong_config.declarative_config)
 
-    ok, err = runloop.build_plugins_plan(utils.uuid())
+    ok, err = runloop.build_plugins_iterator(utils.uuid())
```
bungle:
```suggestion
    ok, err = runloop.build_plugins_iterator("init")
```

kong/runloop/plugins_iterator.lua
```diff
@@ -20,6 +23,42 @@ local MUST_LOAD_CONFIGURATION_IN_PHASES = {
   content     = true,
 }
 
+
+local subsystem    = ngx.config.subsystem
```
bungle:
```suggestion
local subsystem = ngx.config.subsystem
```

kong/runloop/plugins_iterator.lua
```diff
@@ -89,14 +128,15 @@ end
 local function get_next(self)
   local i = self.i + 1
 
-  local plugin = self.plugins_plan.loaded[i]
+  loaded_plugins = loaded_plugins or get_loaded_plugins()
```
bungle:
```suggestion
  if not loaded_plugins then
    loaded_plugins = get_loaded_plugins()
  end
```

Though, should it be in `self.iterator.loaded_plugins` to not check it on each iteration? Or why is this needef at all?

kong/runloop/plugins_iterator.lua
```diff
@@ -123,110 +163,176 @@ local function get_next(self)
...
 
     local name   = plugin.name
-    local combos = self.plugins_plan.combos[name]
+    local combos = self.iterator.combos[name]
 
     local plugin_configuration
 
-    repeat
+    if not combos then
+      goto after_combos
```
bungle: Not sure, but is it possible to do this with function like `self:find_configuration()`, with early returns instead of `gotos`?

kong/runloop/plugins_iterator.lua
```diff
@@ -123,110 +163,176 @@ local function get_next(self)
...
 -- Iterate over the plugin loaded for a request, stored in
 --`ngx.ctx.plugins`.
 --
--- @param[type=table] ctx Nginx context table
 -- @param[type=string] phase Plugins iterator execution phase
--- @param[type=table] plugins Plugins table
 -- @treturn function iterator
-local function plugins_iterator(ctx, phase, plugins_plan)
+local function iterate(self, phase)
+  local ctx = ngx.ctx
```
bungle: I think the `ctx` should be passed here as optimization to avoid `ngx.ctx`.

kong/init.lua
kikito: I tried that change in isolation here, but some declarative tests failed: https://travis-ci.org/Kong/kong/builds/530721676 (the branch is erased now, but it only contained that change - "init" instead of `utils.uuid()`)

kong/init.lua
kikito: Same as in the other case.

kong/runloop/plugins_iterator.lua
kikito: Yes, I did this several times and forgot to undo this, thanks

kong/runloop/plugins_iterator.lua
kikito: I considered that, but I figured this would be faster than that (one less function call). Given that this is hot code I didn't want to make it slower.

kong/runloop/plugins_iterator.lua
kikito: done

kong/init.lua
bungle: yes, those issues where all fixed the on async branch

bungle: if i remember correctly the declarative uses the uuid to determine later on that it has changed and will start to build again. I can check this on monday.

kong/runloop/plugins_iterator.lua
bungle: I am pretty sure this extra func call will be meaningless. And if performance is a thing, the repeat might have been even faster. How did you figure this would be faster in any meaningfull way?

bungle: thought above comment, I don't feel strongly about it. It was just something in lines of this pr to make it more structured and easier to understand.

kong/init.lua
bungle: Ok, you refer to this (so let it be like it is then):
https://github.com/Kong/kong/pull/4595

kong/runloop/plugins_iterator.lua
```diff
@@ -211,22 +253,85 @@ local plugins_iterator_mt = { __call = get_next }
...
+      combos[plugin.name][combo_key] = true
+    end
+  end
+
+  local phase_handler
+  for _, plugin in ipairs(loaded_plugins) do
+    if combos[plugin.name] then
+      for phase_name, phase in pairs(phases) do
+        phase_handler = plugin.handler[phase_name]
+        if phase_handler and phase_handler ~= BasePlugin[phase_name] then
```
bungle: Should this check for `"function"`, like here:
https://github.com/Kong/kong/pull/4590/files#diff-1f0f137350308286de76ec59740c8edaR412

kong/runloop/plugins_iterator.lua
```diff
@@ -123,86 +162,89 @@ local function get_next(self)
...
 
-      if combos[COMBO_GLOBAL] then
-        plugin_configuration = load_configuration(self, name, nil, nil, nil)
-      end
+    if combos[COMBO_GLOBAL] then
+      plugin_configuration = load_configuration(ctx, name, nil, nil, nil)
+    end
 
-    until true
+    ::after_combos::
```
bungle: Can this be moved inside that next `if` clause? Perhaps also rename it to `::configuration_found::`.

kong/runloop/plugins_iterator.lua
```diff
@@ -123,86 +162,89 @@ local function get_next(self)
...
 
-  -- return the plugin configuration
-  if ctx.plugins[plugin.name] then
+  -- Return the plugin configuration if:
+  -- * The plugin must have a customized phase handler function (different from Base Plugin)
+  -- * The current phase is init_worker, (plugin name at least. config might be nil because no route_id/service_id etc in init_worker)
+  -- * A combo has been found and a conf has been loaded
+  local phase = self.iterator.phases[self.phase]
+  if phase and phase[plugin.name]
+  and (self.phase == "init_worker" or ctx.plugins[plugin.name]) then
```
bungle: `init_worker` is such a special phase, only executed once on start on each worker, so I was thinking, could we remove that after the `init_worker` is done?

kong/runloop/plugins_iterator.lua
kikito: It can also mean `nothing found` - if `combos` is nil, you go here too. If we name it something like `found` it would be misleading.

kong/runloop/plugins_iterator.lua
kikito: You are right. I will switch the order in the statement.

kong/runloop/plugins_iterator.lua
kikito: Sure, I will make the change

kong/runloop/plugins_iterator.lua
bungle: in not combos, then we should perhaps jump over the whole block?

bungle: e.g. replace this `do` with:
https://github.com/Kong/kong/blob/10d2e9c70e1f48565e520175be24c5a634143ff1/kong/runloop/plugins_iterator.lua#L152

```
if not combos then
```

hishamhm: > Then all the gotos will only jump when config was found.

Not really, you can go through every `if` block there and still not have a plugin configuration found at the point of the `found` label (that's why there's an `if` right after that).

hishamhm: As @bungle [originally suggested](https://github.com/Kong/kong/pull/4594#discussion_r283080554), I think changing the whole `goto` dance with 

```
local plugin_configuration
if combos then
  plugin_configuration = try_combos(ctx, combos, name, route_id, service_id, consumer_id)
end
```

should simplify things a lot. @kikito, I understand your intuition about the cost of a function call, but LuaJIT is a trace compiler and the one thing trace compilers are really good at is getting rid of function calls. And given @lukego's recent comments about the weirdness of `goto` behavior, I really don't think we should assume that making a plain call there would be necessarily slower.

bungle: @hishamhm it is changed now according to my comments, and `if` can't jump inside `if`. so that check is still there, it was never meant to be removed, see my original comment, but you cannot but label inside if.

hishamhm: @bungle oh, what I meant was just that now, as of 6c17b13, it is possible to reach label `found` with the config not really being found at that point, so the label is misleading.

kikito: Ok, changing to function calls

PR #4595, opened 2019-05-10
perf(declarative) avoid invalidating the cache on init_worker

kikito:
If the declarative config is being loaded during the init_worker
phase the cache must already be empty and there should be no router
to invalidate.

Extracted from #4488


hishamhm:
When refreshing the declarative config with `kong reload`, init_worker is called on a non-empty cache, so this change is not correct. In any case, this logic will be superseded by the atomic update PR, which modifies the purging logic.

PR #4599, opened 2019-05-12
feat(schema) make `key` field of `certificate` entity to be optional

dndx:
So that we may use it to store certificates that does not have private key available (for example, CA certs)

Note: PR based on `1.0.2` tag. Will provide a version for `master` as well.

### DIFF
--- a/kong/db/dao/certificates.lua
+++ b/kong/db/dao/certificates.lua
@@ -56,7 +56,7 @@ function _Certificates:insert(cert, options)
 
   if (not cert.key or cert.key == null) and name_list and #name_list ~= 0 then
     local err_t = self.errors:schema_violation({
-      snis = "private key is not present, \"snis\" field should be absent",
+      snis = "cannot be specified because the 'key' attribute is not set",
     })
 
     return nil, tostring(err_t), err_t
@@ -118,10 +118,19 @@ function _Certificates:update(cert_pk, cert, options)
   if cert.key or cert.cert then
     if cert.key == null then
       -- user tries to remove private key from cert
-      if (name_list and #name_list ~= 0) or #snis ~= 0 then
+      if #snis > 0 and not name_list then
         local err_t = self.errors:schema_violation({
-          key = "one or more SNI names are still associated with this" ..
-                 " certificate, private key cannot be removed",
+          key = "cannot be removed, since one or more SNIs are still" ..
+                " associated to this Certificate",
+        })
+
+        return nil, tostring(err_t), err_t
+      end
+
+      if name_list and #name_list ~= 0 then
+        local err_t = self.errors:schema_violation({
+          snis = "cannot associate SNIs to this Certificate" ..
+                 " because the 'key' attribute is not set",
         })
 
         return nil, tostring(err_t), err_t
@@ -166,7 +175,7 @@ function _Certificates:upsert(cert_pk, cert, options)
 
   if (not cert.key or cert.key == null) and name_list and #name_list ~= 0 then
     local err_t = self.errors:schema_violation({
-      snis = "private key is not present, \"snis\" field should be absent",
+      snis = "cannot be specified because the 'key' attribute is not set",
     })
 
     return nil, tostring(err_t), err_t
--- a/kong/db/dao/certificates.lua
+++ b/kong/db/dao/certificates.lua
@@ -7,6 +7,7 @@ local tostring = tostring
 local ipairs = ipairs
 local table = table
 local type = type
+local null = ngx.null
 
 
 -- Get an array of SNI names from either
@@ -51,6 +52,14 @@ function _Certificates:insert(cert, options)
   local name_list, err, err_t = parse_name_list(cert.snis, self.errors)
   if err then
     return nil, err, err_t
+  end
+
+  if (not cert.key or cert.key == null) and name_list and #name_list ~= 0 then
+    local err_t = self.errors:schema_violation({
+      snis = "private key is not present, \"snis\" field should be absent",
+    })
+
+    return nil, tostring(err_t), err_t
   end
 
   if name_list then
@@ -99,8 +108,26 @@ function _Certificates:update(cert_pk, cert, options)
     end
   end
 
+  local snis
+  snis, err, err_t = self.db.snis:list_for_certificate(cert_pk)
+  if not snis then
+    return nil, err, err_t
+  end
+
   -- update certificate if necessary
   if cert.key or cert.cert then
+    if cert.key == null then
+      -- user tries to remove private key from cert
+      if (name_list and #name_list ~= 0) or #snis ~= 0 then
+        local err_t = self.errors:schema_violation({
+          key = "one or more SNI names are still associated with this" ..
+                 " certificate, private key cannot be removed",
+        })
+
+        return nil, tostring(err_t), err_t
+      end
+    end
+
     cert.snis = nil
     cert, err, err_t = self.super.update(self, cert_pk, cert, options)
     if err then
@@ -117,10 +144,7 @@ function _Certificates:update(cert_pk, cert, options)
     end
 
   else
-    cert.snis, err, err_t = self.db.snis:list_for_certificate(cert_pk)
-    if not cert.snis then
-      return nil, err, err_t
-    end
+    cert.snis = snis
   end
 
   return cert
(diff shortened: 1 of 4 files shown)


kong/db/dao/certificates.lua
```diff
@@ -99,8 +108,26 @@ function _Certificates:update(cert_pk, cert, options)
...
+  end
+
   -- update certificate if necessary
   if cert.key or cert.cert then
+    if cert.key == null then
+      -- user tries to remove private key from cert
+      if (name_list and #name_list ~= 0) or #snis ~= 0 then
+        local err_t = self.errors:schema_violation({
+          key = "one or more SNI names are still associated with this" ..
+                 " certificate, private key can not be removed",
```
hishamhm: typo: `can not` -> `cannot`

kong/db/dao/snis.lua
```diff
@@ -18,6 +18,31 @@ local function invalidate_cache(self, old_entity, err, err_t)
 end
 
 
+local function check_private_key_exists(self, sni)
+  if not sni.certificate or not sni.certificate.id then
+    -- just let validator do it's job
```
hishamhm: typo: `it's` -> `its`

kong/db/dao/snis.lua
```diff
@@ -129,7 +154,12 @@ end
 
 -- invalidates the *old* name when updating it to a new name
 function _SNIs:update(pk, entity, options)
-  local _, err, err_t = invalidate_cache(self, self:select(pk))
+  local ok, err, err_t = check_private_key_exists(self, entity)
+  if not ok then
+    return nil, err, err_t
+  end
+
+  _, err, err_t = invalidate_cache(self, self:select(pk))
```
hishamhm: `local _` missing, breaks lint

kong/db/schema/entities/certificates.lua
```diff
@@ -11,13 +12,18 @@ return {
...
     { cert           = typedefs.certificate { required = true }, },
-    { key            = typedefs.key         { required = true }, },
+    { key            = typedefs.key, },
   },
 
   entity_checks = {
     { custom_entity_check = {
       field_sources = { "cert", "key" },
       fn = function(entity)
+        if not entity.key or entity.key == null then
```
hishamhm: `if entity.key == null` should be sufficient; this function is not called on a partial update if `cert` and `key` are not given.

hishamhm:
> Will provide a version for `master` as well.

This extends the DAO API, so it should go to `next`.

kong/db/dao/snis.lua
dndx: I always thought `_` had special meaning on its own, until today... Thanks for pointing it out!

kong/db/dao/snis.lua
dndx: I always thought _ had special meaning on its own, until today... Thanks for pointing it out!

kong/db/dao/certificates.lua
```diff
@@ -53,6 +54,14 @@ function _Certificates:insert(cert, options)
     return nil, err, err_t
   end
 
+  if (not cert.key or cert.key == null) and name_list and #name_list ~= 0 then
+    local err_t = self.errors:schema_violation({
+      snis = "private key is not present, \"snis\" field should be absent",
```
thibaultcha: One recommendation I'd like to make is to reformulate those errors to be a bit more explicit (instead of "present", "absent", and a couple of semantic/quoting issues). Maybe something like:
```
snis = "cannot be specified since Certificate has no 'key' attribute"
```

(Since `snis` already is the field name in the error object, this is closer to our other schema errors for individual fields. We also use single quotes to refer to other attributes in the error itself, and capitalize entities' names, as per our writing guidelines)

kong/db/dao/certificates.lua
```diff
@@ -99,8 +108,26 @@ function _Certificates:update(cert_pk, cert, options)
...
+  end
+
   -- update certificate if necessary
   if cert.key or cert.cert then
+    if cert.key == null then
+      -- user tries to remove private key from cert
+      if (name_list and #name_list ~= 0) or #snis ~= 0 then
+        local err_t = self.errors:schema_violation({
+          key = "one or more SNI names are still associated with this" ..
+                 " certificate, private key cannot be removed",
```
thibaultcha: In the same vein:
```
key = "cannot be removed, since one or more SNIs are still associated to this Certificate"
```

(Since `key` already is the field name in the error object, this is closer to our other schema errors for individual fields)

kong/db/dao/certificates.lua
```diff
@@ -99,8 +108,26 @@ function _Certificates:update(cert_pk, cert, options)
...
+  snis, err, err_t = self.db.snis:list_for_certificate(cert_pk)
+  if not snis then
+    return nil, err, err_t
+  end
+
   -- update certificate if necessary
   if cert.key or cert.cert then
+    if cert.key == null then
+      -- user tries to remove private key from cert
+      if (name_list and #name_list ~= 0) or #snis ~= 0 then
```
thibaultcha: Should we distinguish those two cases?

```lua
if #name_list > 0 then
if #snis > 0 then
```

since the below error message specifies "still associated with this Certificate", but the first case means that we _want_ to associate an SNI to this Certificate, which is an important nuance for a reader of this error. 

Since `check_list_is_new()` will return `true` if all given SNIs are already associated with this Certificate, we can update the logic like such to still ensure that we differentiate both cases:
```lua
if #snis > 0 then
  -- "SNIs are still associated to this Certificate"
end

if #name_list > 0 then
  -- "cannot associate SNIs to this Certificate"
end
```

kong/db/dao/certificates.lua
hishamhm: not to bikeshed, but `has no 'key' attribute` is also slightly confusing, as the entity itself _has_ the attribute, the instance doesn't (and people sometimes confuse those things). `cannot be specified because the 'key' attribute is not set`, perhaps?

kong/db/dao/certificates.lua
dndx: I looked at this again, it seems that this should really be:

`if (name_list and #name_list ~= 0) and #snis ~= 0 then` instead of `or`.

It is possible for the cert to have both SNI and `key` associated with it, then user tries to update and remove both `key` and `sni` at the same time, this current logic will prevent it from happening.

Here is a test case I just added to show this, it currently fails:

```lua
it("#only removing private key and SNIs at the same time from certificate should succeed", function()
        local res = client:patch("/certificates/" .. cert_foo.id, {
          body    = {
            key = cjson.null,
            snis = cjson.null,
          },
          headers = { ["Content-Type"] = "application/json" },
        })
```

How about:

```lua
if #snis > 0 and not name_list then
  -- "SNIs are still associated to this Certificate"
end

if #name_list > 0 then
  -- "cannot associate SNIs to this Certificate"
end
```

kong/db/dao/certificates.lua
dndx: Changed to `snis = "cannot be specified because the 'key' attribute is not set"`, I think this sounds the best and aligns with both @thibaultcha and @hishamhm 's suggestions :)

kong/db/dao/certificates.lua
dndx: This commit contains the change and new tests to cover the case:

https://github.com/Kong/kong/pull/4599/commits/ec97799a4b8411331a2f0411654d249cb3d24af4

@hishamhm @thibaultcha mind taking another look? Thanks!

kong/db/dao/snis.lua
```diff
@@ -18,6 +18,31 @@ local function invalidate_cache(self, old_entity, err, err_t)
...
+  local cert, err, err_t = self.db.certificates:select({
+    id = sni.certificate.id,
+  })
+  if not cert then
+    return nil, err, err_t
+  end
+
+  if not cert.key then
+    local err_t = self.errors:schema_violation({
+      certificate = "must have the \"key\" field (private key) set to" ..
```
thibaultcha: We should also fix this message (i.e. quotes and wording)

dndx: Changed to `"certificate: cannot be used by SNI because specified Certificate does not have the 'key' attribute set"`.

hishamhm:
@dndx Only thing missing is rebasing/retargeting the PR to `next`, otherwise looks good!

thibaultcha:
Closing since a new PR has been opened to target `next` (#4608). In the future @dndx, you can change the base branch of a PR from the GitHub UI (maybe you already knew this).

dndx:
@thibaultcha It was necessary to open a new PR because this one has conflicts with the `next` branch, but yeah it's okay to close it :)

thibaultcha:
@dndx You can force push a rebased branch and then edit the base branch in the GitHub UI. But that's fine :)

PR #4609, opened 2019-05-14
hotfix(declarative) fix cache flip + feat(declarative) /cache?check_hash=1

hishamhm:
Instead of opening 2 PRs (one of which would be a single hotfix commit) which would require a rebase and conflict-fix once merged, I'm pushing these combined for review. If desired, I can open `hotfix/cache-flip` as a separate PR. Each commit has its own description and can be reviewed separately (they do not undo work from previous commits).

## hotfix(declarative) propagate cache flip to workers

https://github.com/Kong/kong/commit/2767d9bae41dc2c6451fd16a5e3ef72348c603b8

Thanks @bungle for the report!

This hotfix fixes two problems at once, both related to incorrect behavior flipping cache pages:

* we now propagate worker events so that all workers flip their cache pages and point to the currently-filled shm
* we avoid flipping pages at startup, to prevent dependencies with worker-event propagation order at initialization

Two regression tests are included, one for each scenario. The worker propagation test is probabilistic, but given the configuration of the test, it fails consistently on my machine
without the patch, and passes consistently with it. The startup test is deterministic.

## feat(declarative) hash checking of declarative configuration

The remaining of the the PR is a new feature, motivated by the investigation of the behavior of declarative config with regard to `kong reload`. In #4592, the reset of the `kong ` shm made so that a `kong reload` correctly rescans the declarative file configured in `declarative_config`. This PR adds tests in aa8d5fa that verify that this behavior is correct. While working on the behavior of `kong reload`, I added hash-checking support to the declarative config loader. It made sense to add this functionality to the `/config` endpoint. 

This introduces a new query argument so that the hash of a given declarative configuration is checked against the hash of the current configuration before loading it. For backward compatibility, this was added as a new query argument:

```
$ http :8001/config?check_hash=1 config=@config.yml   # load a config
201 Created
{
   ...
}

$ http :8001/config?check_hash=1 config=@config.yml   # reload the same config
304 Not Modified
```

Hash checking only happens when the value of `check_hash` is `1`. (This design was chosen to allow for future extensibility, if we add new hashing behaviors, those can be set to different numbers.)

### DIFF
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -1,6 +1,7 @@
 local declarative = require("kong.db.declarative")
 local concurrency = require("kong.concurrency")
 local reports = require("kong.reports")
+local errors = require("kong.db.errors")
 local kong = kong
 local dc = declarative.new_config(kong.configuration)
 
@@ -27,17 +28,16 @@ return {
         })
       end
 
-      local entities, err_or_ver
+      local entities, err_or_ver, err_t
       if self.params._format_version then
-        entities, err_or_ver = dc:parse_table(self.params)
+        entities, err_or_ver, err_t = dc:parse_table(self.params)
       else
         local config = self.params.config
-        -- TODO extract proper filename from the input
-        entities, err_or_ver = dc:parse_string(config, "config.yml", accept)
+        entities, err_or_ver, err_t = dc:parse_string(config, nil, accept)
       end
 
       if not entities then
-        return kong.response.exit(400, { error = err_or_ver })
+        return kong.response.exit(400, errors:declarative_config(err_t))
       end
 
       local ok, err = concurrency.with_worker_mutex({ name = "dbless-worker" }, function()
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -28,12 +28,12 @@ return {
         })
       end
 
-      local entities, err_or_ver, err_t
+      local entities, _, err_t, vers
       if self.params._format_version then
-        entities, err_or_ver, err_t = dc:parse_table(self.params)
+        entities, _, err_t, vers = dc:parse_table(self.params)
       else
         local config = self.params.config
-        entities, err_or_ver, err_t = dc:parse_string(config, nil, accept)
+        entities, _, err_t, vers = dc:parse_string(config, nil, accept)
       end
 
       if not entities then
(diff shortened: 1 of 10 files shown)


kong/db/declarative/init.lua
```diff
@@ -395,34 +395,32 @@ local function cache_entities(entities)
...
+
+  post_upstream_crud_delete_events()
 
-  local ok, err = cache_entities(entities)
+  local ok
+  ok, err = declarative.load_into_cache(entities, SHADOW)
   if ok then
-    kong.cache:flip()
+    kong.worker_events.post("declarative", "flip_config", true)
+    kong.worker_events.poll()
```
kikito: Shouldn't `kong.worker_events.poll()` be checked for errors like it's done before? If not, I suggest adding a comment explaining that errors are suppressed on purpose.

hishamhm: @Tieske is it guaranteed as per the contract of the worker-events API that the worker who is sending the event will execute the event callback _immediately_ (i.e., before `post` returns)? That way I would not have to `poll` here (the remainder of this function requires the operation triggered by this post to have been executed).

Tieske: doing a 'post' implicitly does a 'poll' (it returns the results from that `poll` call). This can return any of the following values:

- `true`: the event has been posted and handlers executed (queue is empty)
- `false`: the event was queued, but recursively from an event-handler, so hasn't been executed. Will be done after the current handler returns.
- `nil+error`: something went wrong

So the `poll` call here is useless, since the previous call to `post` already called `poll`.

The caveat is that the events will only be executed if the current code does NOT run itself within an event-handler.

Tieske: (the contract here is that every worker executes the events in the exact same order, guaranteed)

hishamhm: ok, "post implies poll" is the bit I needed — and I just realized it is in the docs, d'oh. Sorry about the noise! @kikito, I will remove this line and add error checking in `:post`. Thanks for the catch!

kong/api/routes/config.lua
```diff
@@ -27,21 +28,32 @@ return {
         })
       end
 
-      local entities, err_or_ver
+      local check_hash, old_hash
+      if tostring(self.params.check_hash) == "1" then
```
p0pr0ck5: just curious, why are we checking explicitly for `1`? any reason not to check for `true` (meaning no param value needs to be given, just the presence of the param key), or for any non-nil value?

hishamhm: From the PR description text: "This design was chosen to allow for future extensibility, if we add new hashing behaviors, those can be set to different numbers." :)

p0pr0ck5: Apologies for not opening my eyes :) thanks!

PR #4616, opened 2019-05-15
fix(healthchecks) fix active healthchecks in stream module

hishamhm:
Instead of disabling active healthchecks in the stream module altogether (which was done to avoid duplicate active healthchecks), what we need to do is to run the "tcp" type healthchecks in the stream module, and the other ones ("http", "https") in the http module.

For active healthchecks on stream to work, we need to bump lua-resty-healthcheck. 

This PR also adds a test for active healthchecks using stream, which also serves like a regression test for Kong/lua-resty-healthcheck#19

### DIFF
--- a/spec/02-integration/05-proxy/10-balancer_spec.lua
+++ b/spec/02-integration/05-proxy/10-balancer_spec.lua
@@ -524,6 +524,25 @@ local function wait_for_router_update(bp, old_rv, localhost, proxy_port, admin_p
 end
 
 
+local function tcp_client_requests(nreqs, host, port)
+  local fails, ok1, ok2 = 0, 0, 0
+  for _ = 1, nreqs do
+    local sock = ngx.socket.tcp()
+    assert(sock:connect(host, port))
+    assert(sock:send("hello\n"))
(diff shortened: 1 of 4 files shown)


spec/02-integration/05-proxy/10-balancer_spec.lua
```diff
@@ -524,6 +524,26 @@ local function wait_for_router_update(bp, old_rv, localhost, proxy_port, admin_p
...
+local function tcp_client_requests(nreqs, host, port)
+  local fails, ok1, ok2 = 0, 0, 0
+  for _ = 1, nreqs do
+    local sock = ngx.socket.tcp()
+    assert(sock:connect(host, port))
+    assert(sock:send("hello\n"))
+    local response, err = sock:receive()
+    if err then
+      fails = fails + 1
+      print(err)
```
bungle: Is this leftover?

hishamhm: oops, fixed!

Tieske:
nice follow up! didn't know this was disabled. 👍

PR #4618, opened 2019-05-15
chore(plugins) bump version fields

gszr:
### Summary

Bump plugins versions to account for recent internal changes their
handlers (a29c355). Said changes removed plugins handlers dependency
on the base plugin class; there were changes in the plugins DAO
(`load_plugin_schemas`) to accommodate for the base class removal, making
it non-backwards compatible, warranting a major version bump in plugins
versions.

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -39,7 +39,7 @@ local ACLHandler = {}
 
 
 ACLHandler.PRIORITY = 950
-ACLHandler.VERSION = "1.0.0"
+ACLHandler.VERSION = "2.0.0"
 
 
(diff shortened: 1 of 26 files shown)


bungle:
@gszr I dismissed my approval because of a question: Should we make version to `1.2.0` instead as the `1.1.0` is already released an `next` points to upcoming `1.2.0`?

gszr:
@bungle Good point. Will hold to hear more thoughts on this, but IMO it makes sense to bump to 1.2.0

gszr:
Updated the PR; the base class removal is not backwards-compatible (there were changes also in the plugins DAO), so a major bump is warranted. Please re-review @bungle. Thanks!

PR #4619, opened 2019-05-15
feat: move configured plugin map into singletons so that other code c…

dndx:
…an also

access it


thibaultcha:
This part of the runloop has changed a lot in `next`. I am not sure if we should make such a change in master (given the next release will be based on `next` anyway, and this PR will create merge conflicts).

PR #4620, opened 2019-05-16
fix(ldap-auth) reuse connection if starttls has been already started

artomal:
### Summary

This patch fixes STARTTLS functionality which hasn't been working because plugin starts TLS connection even for existing connections from the pool and then fails.

This patch separates connection pool for TLS connections because STARTTLS extention works on the same port as non-secure connections and two plugins configured with different settings can end up using secure connection or worse non-secure.

This patch adds additional check if TLS connection is being reused or not, that helps to prevent getting error messages from LDAP when plugins tries to STARTTLS connection even if it is already started.

This patch supplements tests with additional configuration for multiple LDAP connection strategies (non-secure, starttls).

### Full changelog

* Add separate connection pool for secure connections
* Add additional check for TLS connections to know if connection is reused
* Add configuration for tests to run on multiple LDAP connection strategies

### Issues resolved

Fix STARTTLS connection initialization and connections re-usage from connection pool.

### DIFF
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -50,7 +50,15 @@ local function ldap_authenticate(given_username, given_password, conf)
 
   sock:settimeout(conf.timeout)
 
-  ok, err = sock:connect(conf.ldap_host, conf.ldap_port)
+  local opts = {}
+
+  -- keep TLS connections in a separate pool to avoid reusing non-secure
+  -- connections and vica versa, because StartTLS use the same port 
+  if conf.start_tls then
+    opts.pool = conf.ldap_host .. ":" .. conf.ldap_port .. ":starttls"
+  end
+
+  ok, err = sock:connect(conf.ldap_host, conf.ldap_port, opts)
   if not ok then
     kong.log.err("failed to connect to ", conf.ldap_host, ":",
                    tostring(conf.ldap_port), ": ", err)
@@ -58,9 +66,12 @@ local function ldap_authenticate(given_username, given_password, conf)
   end
 
   if conf.start_tls then
-    local success, err = ldap.start_tls(sock)
-    if not success then
-      return false, err
+    -- convert connection to a StarTLS connection only if it is a new connection
+    if conf.start_tls and sock:getreusedtimes() == 0 then
+      local success, err = ldap.start_tls(sock)
+      if not success then
+        return false, err
+      end
     end
 
     _, err = sock:sslhandshake(true, conf.ldap_host, conf.verify_ldap_host)
--- a/spec/03-plugins/20-ldap-auth/01-access_spec.lua
+++ b/spec/03-plugins/20-ldap-auth/01-access_spec.lua
@@ -22,687 +22,695 @@ end
 
 local ldap_host_aws = "ec2-54-172-82-117.compute-1.amazonaws.com"
 
+local ldap_strategies = {
+  non_secure = { name = "non-secure", start_tls = false },
+  start_tls = { name = "starttls", start_tls = true }
+}
 
-for _, strategy in helpers.each_strategy() do
-  describe("Plugin: ldap-auth (access) [#" .. strategy .. "]", function()
-    local proxy_client
-    local admin_client
-    local route2
-    local plugin2
-
-    lazy_setup(function()
-      local bp = helpers.get_db_utils(strategy, {
-        "routes",
-        "services",
-        "plugins",
-        "consumers",
-      })
-
-      local route1 = bp.routes:insert {
-        hosts = { "ldap.com" },
-      }
-
-      route2 = bp.routes:insert {
-        hosts = { "ldap2.com" },
-      }
-
-      local route3 = bp.routes:insert {
-        hosts = { "ldap3.com" },
-      }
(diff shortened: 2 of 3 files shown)


kong/plugins/ldap-auth/access.lua
```diff
@@ -50,17 +50,28 @@ local function ldap_authenticate(given_username, given_password, conf)
 
   sock:settimeout(conf.timeout)
 
-  ok, err = sock:connect(conf.ldap_host, conf.ldap_port)
+  local opts = {}
+
+  -- keep TLS connections in a separate pool to avoid reusing non-secure
+  -- connections and vica versa, because StartTLS use the same port
```
hishamhm: typo: vice-versa

artomal: @hishamhm good eye, thanks for catching that!

kong/plugins/ldap-auth/access.lua
```diff
@@ -50,17 +50,28 @@ local function ldap_authenticate(given_username, given_password, conf)
...
                    tostring(conf.ldap_port), ": ", err)
     return nil, err
   end
 
   if conf.start_tls then
-    local success, err = ldap.start_tls(sock)
-    if not success then
-      return false, err
+    -- convert connection to a StarTLS connection only if it is a new connection
+    if conf.start_tls and sock:getreusedtimes() == 0 then
```
thibaultcha: Can we avoid repeating this condition? `conf.start_tls` is already satisfied because we are in this branch in the first place (L.68).

kong/plugins/ldap-auth/access.lua
```diff
@@ -50,17 +50,28 @@ local function ldap_authenticate(given_username, given_password, conf)
...
+      local success, err = ldap.start_tls(sock)
```
thibaultcha: Please use the standard `local ok, err` return value, thanks!

kong/plugins/ldap-auth/access.lua
```diff
@@ -50,17 +50,28 @@ local function ldap_authenticate(given_username, given_password, conf)
...
+      if not success then
+        return false, err
```
thibaultcha: On failure, and by convention, our Lua functions should return `nil, err`, and not `false`.

kong/plugins/ldap-auth/access.lua
```diff
@@ -50,17 +50,28 @@ local function ldap_authenticate(given_username, given_password, conf)
 
   sock:settimeout(conf.timeout)
 
-  ok, err = sock:connect(conf.ldap_host, conf.ldap_port)
+  local opts = {}
```
thibaultcha: We could save this allocation and alleviate GC pressure if `conf.start_tls` isn't truthy.

kong/plugins/ldap-auth/access.lua
```diff
@@ -50,17 +50,28 @@ local function ldap_authenticate(given_username, given_password, conf)
...
+
+  -- keep TLS connections in a separate pool to avoid reusing non-secure
+  -- connections and vica versa, because StartTLS use the same port 
+  if conf.start_tls then
+    opts.pool = conf.ldap_host .. ":" .. conf.ldap_port .. ":starttls"
```
thibaultcha: Is there a reason why we would only want to set this pool option for TLS connections only? Is there a reason why we cannot use the default `pool` key generated by ngx_lua?

thibaultcha:
Alas, this was merged...

kong/plugins/ldap-auth/access.lua
dndx: The reason is that the default `pool` key only has IP and ports and since `STARTTLS` uses the same IP and port with cleartext connections, mixing them will create issues in reused connections (cleartext connection returned for `STARTTLS` request and already `STARTTLS`'d connections returned for cleartext connections).

artomal: The reason for pool separation is that STARTTLS use non-secure port to initialize connections and connection pool name is generated by default in [HOST:PORT] pattern so it might be the case where non-secure and TLS connections be mixed up in the same pool, so thought it is a good thing to separate this particular pool with [HOST:PORT:starttls] name pattern.

kong/plugins/ldap-auth/access.lua
artomal: Didn't want to change what was working before and do refactoring, but will keep in mind, thanks!

hishamhm:
@thibaultcha Oops, race condition! The UI showed me @dndx's review but not yours.

kong/plugins/ldap-auth/access.lua
artomal: Same here, didn't want to change existing code which was working but thanks for the notice!

dndx:
@hishamhm I think it was submitted right after your merge..

thibaultcha:
Indeed, I pressed the submit button and less than a second later this showed up as merged :(

kong/plugins/ldap-auth/access.lua
thibaultcha: > same IP and port with cleartext connections

Ugh... Ok, perfect :+1:

PR #4625, opened 2019-05-16
tests(tcp) reduce flakiness using a handshake in tcp_server

hishamhm:
Add the same kind of handshake in `helpers.tcp_server` that we already do in `helpers.udp_server`. Hopefully, this will reduce flakiness in tests that use the `tcp_server` such as the frequent failures in the plugin tests for `tcp-log`.

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -532,7 +532,7 @@ end
 -- @param `port`    The port where the server will be listening to
 -- @param `opts     A table of options defining the server's behavior
 -- @return `thread` A thread object
-local function tcp_server(port, opts, ...)
+local function tcp_server(port, opts)
   local threads = require "llthreads2.ex"
   opts = opts or {}
   local thread = threads.new({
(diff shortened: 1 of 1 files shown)


spec/helpers.lua
```diff
@@ -561,16 +563,44 @@ local function tcp_server(port, opts, ...)
...
+  -- not necessary for correctness because we do the handshake,
+  -- but avoids harmless "connection error" messages in the wait loop
+  -- in case the client is ready before the server below.
+  ngx.sleep(0.001)
+
+  local sock = ngx.socket.tcp()
+  sock:settimeout(0.01)
+  while true do
+    local c, er = sock:connect("localhost", port)
+    print(c, er)
```
thibaultcha: Is this `print()` intentional?

hishamhm: Dammit :laughing: Second time this week! That's it, I'm adding a pre-commit hook!

hishamhm:
```bash
#!/usr/bin/env bash

files=($(git status --porcelain | grep "^[AM]" | cut -b4-))

lua_files=($(echo "${files[@]}" | tr ' ' '\n' | grep "lua$"))
if [ "${#lua_files[@]}" -gt 0 ]
then
   luacheck -q "${lua_files[@]}"
fi

spec_files=($(echo "${files[@]}" | tr ' ' '\n' | grep "spec/"))

fail=0
function complain() {
   echo -e "\033[1;31m *** $@\033[0m"
   echo
   fail=1
}


if ! [ "$WIP" ]
then
   for file in "${spec_files[@]}"
   do
      grep -w 'print' $file && complain "$file has a print statement!"
      grep -w '#only' $file && complain "$file has a #only comment!"
   done
fi

if [ "$fail" = 1 ]
then
   echo "use 'WIP=1 git commit' to override"
   exit 1
fi

exit 0
```

thibaultcha: I'd also recommend running `grep -R -w -i -- '--- ONLY' t/`

PR #4626, opened 2019-05-16
feat(log) adding more TLS information in basic serializer

subnetmarco:
Adding the following entries in the `basic` log serializer:

* `request.tls.cipher`
* `request.tls.client_verify`

### DIFF
--- a/kong/plugins/log-serializers/basic.lua
+++ b/kong/plugins/log-serializers/basic.lua
@@ -20,10 +20,14 @@ function _M.serialize(ngx)
 
   local request_tls
   local request_tls_ver = ngx_ssl.get_tls1_version_str()
-  if request_tls_ver then 
-    request_tls = { version = request_tls_ver } 
+  if request_tls_ver then
+    request_tls = {
+      version = request_tls_ver,
+      cipher = var.ssl_cipher,
+      client_verify = var.ssl_client_verify,
+    }
   end
-  
+
   local request_uri = var.request_uri or ""
 
   return {
--- a/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
+++ b/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
@@ -155,7 +155,7 @@ for _, strategy in helpers.each_strategy() do
       })
 
       assert.response(r).has.status(200)
-      
+
       -- Getting back the TCP server input
       local ok, res = thread:join()
       assert.True(ok)
@@ -164,6 +164,8 @@ for _, strategy in helpers.each_strategy() do
       -- Making sure it's alright
       local log_message = cjson.decode(res)
       assert.equal("TLSv1.2", log_message.request.tls.version)
+      assert.is_string(log_message.request.tls.cipher)
+      assert.equal("NONE", log_message.request.tls.client_verify)
     end)
 
   end)


kong/plugins/log-serializers/basic.lua
```diff
@@ -21,7 +21,12 @@ function _M.serialize(ngx)
   local request_tls
   local request_tls_ver = ngx_ssl.get_tls1_version_str()
   if request_tls_ver then 
-    request_tls = { version = request_tls_ver } 
+    request_tls = { 
+      version = request_tls_ver, 
+      cipher = var.ssl_cipher,
+      supported_client_ciphers = var.ssl_ciphers,
```
thibaultcha: Is it truly necessary to log client supported ciphers like so? What is the use-case? This will significantly increase the payload size for those logs. Example running the test locally with OpenSSL 1.1.0j and TLS 1.2:
```
supported_client_ciphers = "ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:
DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:
DHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:
DHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:
DHE-RSA-AES256-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:
DHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:
DHE-RSA-AES256-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES128-SHA:DHE-RSA-AES128-SHA:
AES256-GCM-SHA384:AES128-GCM-SHA256:AES256-SHA256:AES128-SHA256:AES256-SHA:
AES128-SHA:0x00ff"
```

TLS 1.3 has a much more restricted list of supported ciphers which would make for more reasonably size payloads, but then one can ask what value logging them would bring, if we can reasonably expect what they are going to be?

subnetmarco: Understanding in large deployments what’s preventing the adoption of more modern ciphers (old clients). In general Kong should be logging everything and also - when the log size is too long - have a way to dynamically exclude certain fields (but this is out of the scope of this PR).

james-callahan: Note that a client can send *hundreds* of supported ciphers: this could quickly overwhelm most logging formats.

subnetmarco: Okay I have removed `supported_client_ciphers` in my latest commit to speed up merging this PR.

kong/plugins/log-serializers/basic.lua
```diff
@@ -21,7 +21,11 @@ function _M.serialize(ngx)
   local request_tls
   local request_tls_ver = ngx_ssl.get_tls1_version_str()
   if request_tls_ver then 
-    request_tls = { version = request_tls_ver } 
+    request_tls = { 
+      version = request_tls_ver, 
+      cipher = var.ssl_cipher,
+      client_verify = var.ssl_client_verify
```
dndx: Do we really need the `$ssl_client_verify` variable here? Even for mTLS use case this variable is not used as we do validation dynamically using `luaossl` with client supplied certificate. Not sure if `$ssl_client_verify` will contain any useful information in this case.

thibaultcha: @dndx It seems like `$ssl_client_verify` is [changeable](http://lxr.nginx.org/source/src/http/modules/ngx_http_ssl_module.c#0343). I think it would make sense to update it in our own mTLS logic to reflect client cert verification that occurred in our own logic. This way, we could merge this PR as-is and expect this variable to work today for native NGINX mTLS support, and forward-compatibility later with Kong's mTLS. Any thoughts on that?

thibaultcha: We will go ahead with the current implementation of this PR, and @dndx will make sure that our mTLS support considers `$ssl_client_verify`. This is good to go now.

thibaultcha:
Pushed some manual changes improving tests resiliency and a couple of style fixes.

PR #4627, opened 2019-05-16
perf(runloop) asynchronous rebuilds of router and plugin iterator

kikito:
This is an adaptation of the main commit in #4488 .

It contains:
* Some small refactors (only the ones needed for this PR to work)
* A medium-sized commit introduces the first part of asynchronous loading:
  * timer-based refreshes of the router and plugins iterator
  * replaces the `concurrency` abstraction with semaphores
  * it introduces the concept of *rebuilding*, different from just *updating* in
    that the rebuild can be synchronous or async depending on the parameters.
* Another change, this one dealing with how errors are handled. This is again to provide a bit more stability when dealing with the scenarios mentioned by @jeremyjpj0916 in #4194

This PR is expanded with an extra commit in #4639 , where the rebuilds are (can) completely uncoupled from the request processing via a config option.

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -246,7 +246,7 @@ local function load_declarative_config(kong_config, entities)
       error("error building initial plugins iterator: " .. err)
     end
 
-    assert(runloop.build_router(kong.db, "init"))
+    assert(runloop.build_router("init"))
 
     mesh.init()
 
@@ -389,7 +389,7 @@ function Kong.init()
       error("error building initial plugins: " .. tostring(err))
     end
 
-    assert(runloop.build_router(db, "init"))
+    assert(runloop.build_router("init"))
   end
 
   db:close()
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -141,8 +141,7 @@ do
 end
 
 
-local function execute_plugins_iterator(ctx, phase)
-  local plugins_iterator = runloop.get_plugins_iterator()
+local function execute_plugins_iterator(plugins_iterator, ctx, phase)
   for plugin, configuration in plugins_iterator:iterate(ctx, phase) do
     kong_global.set_named_ctx(kong, "plugin", configuration)
     kong_global.set_namespaced_log(kong, plugin.name)
@@ -514,13 +513,8 @@ function Kong.ssl_certificate()
 
   runloop.certificate.before(mock_ctx)
 
-  local ok, err = runloop.update_plugins_iterator()
-  if not ok then
-    ngx_log(ngx_CRIT, "could not ensure plugins iterator is up to date: ", err)
-    return ngx.exit(ngx.ERROR)
-  end
-
-  execute_plugins_iterator(mock_ctx, "certificate")
+  local plugins_iterator = runloop.get_updated_plugins_iterator()
+  execute_plugins_iterator(plugins_iterator, mock_ctx, "certificate")
 end
 
 function Kong.balancer()
@@ -628,15 +622,14 @@ function Kong.rewrite()
   runloop.rewrite.before(ctx)
 
   -- On HTTPS requests, the plugins iterator is already updated in the ssl_certificate phase
-  if ngx.var.https ~= "on" then
-    local ok, err = runloop.update_plugins_iterator()
-    if not ok then
-      ngx_log(ngx_CRIT, "could not ensure plugins iterator is up to date: ", err)
-      return kong.response.exit(500, { message  = "An unexpected error occurred" })
-    end
+  local plugins_iterator
+  if ngx.var.https == "on" then
+    plugins_iterator = runloop.get_plugins_iterator()
+  else
+    plugins_iterator = runloop.get_updated_plugins_iterator()
   end
 
-  execute_plugins_iterator(ctx, "rewrite")
+  execute_plugins_iterator(plugins_iterator, ctx, "rewrite")
 
   runloop.rewrite.after(ctx)
 end
@@ -648,13 +641,8 @@ function Kong.preread()
 
   runloop.preread.before(ctx)
 
-  local ok, err = runloop.update_plugins_iterator()
-  if not ok then
-    ngx_log(ngx_CRIT, "could not ensure plugins iterator is up to date: ", err)
-    return kong.response.exit(500, { message  = "An unexpected error occurred" })
-  end
-
-  execute_plugins_iterator(ctx, "preread")
+  local plugins_iterator = runloop.get_updated_plugins_iterator()
+  execute_plugins_iterator(plugins_iterator, ctx, "preread")
 
   runloop.preread.after(ctx)
 end
@@ -702,7 +690,8 @@ function Kong.header_filter()
 
   runloop.header_filter.before(ctx)
 
-  execute_plugins_iterator(ctx, "header_filter")
+  local plugins_iterator = runloop.get_plugins_iterator()
+  execute_plugins_iterator(plugins_iterator, ctx, "header_filter")
 
   runloop.header_filter.after(ctx)
 end
@@ -712,7 +701,8 @@ function Kong.body_filter()
 
   local ctx = ngx.ctx
 
-  execute_plugins_iterator(ctx, "body_filter")
+  local plugins_iterator = runloop.get_plugins_iterator()
+  execute_plugins_iterator(plugins_iterator, ctx, "body_filter")
 
   runloop.body_filter.after(ctx)
 end
@@ -722,7 +712,8 @@ function Kong.log()
 
   local ctx = ngx.ctx
 
-  execute_plugins_iterator(ctx, "log")
+  local plugins_iterator = runloop.get_plugins_iterator()
+  execute_plugins_iterator(plugins_iterator, ctx, "log")
 
   runloop.log.after(ctx)
 end
@@ -735,8 +726,8 @@ function Kong.handle_error()
   ctx.KONG_UNEXPECTED = true
 
   if not ctx.plugins then
-    runloop.update_plugins_iterator()
-    local plugins_iterator = runloop.get_plugins_iterator()
+
+    local plugins_iterator = runloop.get_updated_plugins_iterator()
     for _ in plugins_iterator:iterate(ctx, "content") do
       -- just build list of plugins
     end
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -66,9 +66,8 @@ local TTL_ZERO = { ttl = 0 }
 
 local get_plugins_iterator, build_plugins_iterator, update_plugins_iterator
 
-local get_router, build_router
+local get_router, build_router, update_router
 local server_header = meta._SERVER_TOKENS
-local _set_check_router_rebuild
 
 local build_router_timeout
 
@@ -522,7 +521,8 @@ do
     return service
   end
 
-  build_router = function(db, version)
+  build_router = function(version)
+    local db = kong.db
     local routes, i = {}, 0
 
     local err
@@ -585,11 +585,12 @@ do
       return rp1 > rp2
     end)
 
-    local err
-    router, err = Router.new(routes)
-    if not router then
+    local new_router, err = Router.new(routes)
+    if not new_router then
       return nil, "could not create router: " .. err
     end
+
+    router = new_router
 
     if version then
       router_version = version
@@ -601,7 +602,7 @@ do
   end
 
 
-  local function check_router_rebuild()
+  update_router = function()
     -- we might not need to rebuild the router (if we were not
     -- the first request in this process to enter this code path)
     -- check again and rebuild only if necessary
@@ -617,38 +618,6 @@ do
       return true
     end
 
-    -- router needs to be rebuilt in this worker
-    log(DEBUG, "rebuilding router")
-
-    local ok, err = build_router(singletons.db, version)
-    if not ok then
-      log(CRIT, "could not rebuild router: ", err)
-      return nil, err
-    end
-
-    return true
-  end
-
-
-  -- for unit-testing purposes only
-  _set_check_router_rebuild = function(f)
-    check_router_rebuild = f
-  end
-
-
-  get_router = function()
-    local version, err = singletons.cache:get("router:version",
-                                              CACHE_ROUTER_OPTS,
-                                              utils.uuid)
-    if err then
-      log(CRIT, "could not ensure router is up to date: ", err)
-      return nil, err
-    end
-
-    if version == router_version then
-      return router
-    end
-
     -- wrap router rebuilds in a per-worker mutex:
     -- this prevents dogpiling the database during rebuilds in
     -- high-concurrency traffic patterns;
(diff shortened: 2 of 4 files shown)


kong/init.lua
```diff
@@ -629,7 +630,9 @@ function Kong.rewrite()
 
   -- On HTTPS requests, the plugins iterator is already updated in the ssl_certificate phase
   if ngx.var.https ~= "on" then
-    local ok, err = runloop.update_plugins_iterator()
+
+    local timeout = runloop.get_rebuild_timeout()
+    local ok, err = runloop.rebuild_plugins_iterator(timeout)
```
bungle: I see that you call `runloop.get_rebuild_timeout()` and immediately after `runloop.rebuild_plugins_iterator(timeout)`.  Doesn't `runloop` already know the `timeout`? I am fine with this argument, but when passed as `nil` could `runloop` then just get the timeout on it's own?

So could this be changed to just:
```
local ok, err = runloop.rebuild_plugins_iterator()
```

kong/init.lua
```diff
@@ -735,7 +739,9 @@ function Kong.handle_error()
   ctx.KONG_UNEXPECTED = true
 
   if not ctx.plugins then
-    runloop.update_plugins_iterator()
+
+    local timeout = runloop.get_rebuild_timeout()
+    runloop.rebuild_plugins_iterator(timeout) -- ignore error
```
bungle: Same here.

kong/init.lua
```diff
@@ -514,7 +514,8 @@ function Kong.ssl_certificate()
 
   runloop.certificate.before(mock_ctx)
 
-  local ok, err = runloop.update_plugins_iterator()
+  local timeout = runloop.get_rebuild_timeout()
+  local ok, err = runloop.rebuild_plugins_iterator(timeout)
```
bungle: Same here.

kong/runloop/handler.lua
```diff
@@ -365,6 +372,190 @@ local function register_events()
...
+
+
+local function get_version(name)
+  if not kong.cache then
+    return "init"
+  end
+
+  local version, err = kong.cache:get(name .. ":version", TTL_ZERO, utils.uuid)
+  if err then
+    return nil, "could not ensure " .. name .. " is up to date: " .. err
```
bungle: I see you changed this (internal) function to `return nil, err` which leads the calling code to handle the error (which is going to repeat same code all over?)

kong/runloop/handler.lua
```diff
@@ -365,6 +372,190 @@ local function register_events()
...
+  -- nil otherwise (safe_rebuild was neither called successfully nor enqueued).
+  -- Notice that this function might return true *even when it fails to acquire
+  -- the semaphore or invoke the async callback* - it will try to rebuild
+  -- synchronously when a wait is specified and the semaphore isn't acquired
+  -- or the timer-based call fails.
+  -- @returns error message as a second return value in case of failure
+  rebuild = function(name, callback, version, semaphore, wait)
+    local current_version, err = get_version(name)
+    if not current_version then
+      return nil, err
```
bungle: You made the ´get_version` as a fatal error. I am not sure if that is a right call. I would in that case try to build a new router as it is done below.

kong/runloop/handler.lua
```diff
@@ -365,6 +372,190 @@ local function register_events()
...
+      return true
+    end
+
+    local ok = acquire_semaphore(semaphore, wait)
+    if not ok then
+      if wait and wait > 0 then
+        log(NOTICE, "could not acquire semaphore, attempting sync rebuild")
+        local ok, err = safe_rebuild(name, callback, current_version)
+        if not ok then
+          return nil, err
```
bungle: Again here it is marked as fatal for the request. I question that a bit as I think it is better to serve request with `stale` router/plugins than refuse to serve request at all. This affects usually only one request. If Kong never is going to make a new rebuild even on other requests, you just got a lot of lines `CRIT` lines in your log, but it will still allow you to proxy requests with old router, until someone acts upon it. With making these fatal and rebuild never succeeding, you practically make all request stop working. Think about this happening middle of a night. That was the original reason that I made these to log with CRIT/ERR.

Of course this is debatable. Yours is more correct in a way that Kong immediately refuses to serve requests with stale router/plugins. But I would say that mine is more resilient on error and allows at least `99.99%` of the traffic still be processed 100% correct.

Though I am not sure if you handled ^ elsewhere, making this comment obsolete, but you got the idea?

kong/runloop/handler.lua
```diff
@@ -365,6 +372,190 @@ local function register_events()
...
+        return true
+      end
+
+      return
+    end
+
+    current_version, err = get_version(name)
+    if not current_version then
+      release_semaphore(semaphore)
+      return nil, err
```
bungle: Again, not sure if this error case should be handled as a fatal. I know it makes the interface a bit strange. My original `rebuild` didn't return anything as it handled the errors accordingly, e.g. it logged them and then just let the other code continue (it formed a border where rebuilding related errors were handled and not pushing those issues to calling code). Again, somewhat debatable, but just explaining the reasons in my original implementation, because that might not be obvious. What I mean here is that I try to follow bit different error handling than in normal code, because this is a hot path, and some errors should not in my opinion halt processing the requests. What I mean here is that we try hard (even on each following request to rebuild everything), but that is our best effort. It could be that there is a bug or maybe temporary memory issue etc., but because we cannot handle all the error cases we just try our best. Failing early just makes everything halt. I do consider there kinda of errors very severe, but not that severe, that it warrants to halt processing.

kong/runloop/handler.lua
```diff
@@ -365,6 +372,190 @@ local function register_events()
...
+    end
+    if current_version == version then
+      release_semaphore(semaphore)
+      return true
+    end
+
+    if wait and wait > 0 then
+      local ok, err = safe_rebuild(name, callback, current_version)
+      release_semaphore(semaphore)
+      return ok, err
```
bungle: See above.

kong/runloop/handler.lua
```diff
@@ -365,6 +372,190 @@ local function register_events()
...
+    else
+      ok, err = rebuild_async(name, callback, current_version, semaphore)
+      if not ok then
+        if wait and wait == 0 then
+          log(NOTICE, "could not rebuild asynchronously:",
+                       err,
+                       ", attempting synchronous rebuild")
+          local ok, err = safe_rebuild(name, callback, current_version)
+          release_semaphore(semaphore)
+          return ok, err
```
bungle: As here too.

kong/runloop/handler.lua
```diff
@@ -405,44 +422,59 @@ do
...
 
-    local version, err = kong.cache:get(name .. ":version", TTL_ZERO, utils.uuid)
-    if err then
-      return nil, "could not ensure " .. name .. " is up to date: " .. err
-    end
 
-    return version
-  end
+  local function safe_rebuild(name, callback, version)
+    local max_attempts = 6
```
bungle: I think in original code I tried 5 times, but this is just a number thrown out of hat.

kong/runloop/handler.lua
```diff
@@ -364,6 +369,161 @@ local function register_events()
 end
 
 
+local function get_rebuild_timeout()
```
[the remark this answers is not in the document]
kikito: Yes, agreed. I have replaced this by a single REBUILD_TIMEOUT constant in runloop/handler.lua, initialized on init_worker

kong/init.lua
kikito: I have replaced this with `runloop.get_updated_plugins_iterator()`, which will:
* get the updated plugins iterator using the proper timeout (now `REBUILD_TIMEOUT`)
* If updating fails, it logs the failure with `CRIT` and just returns the non-updated plugins iterator.
* So this function never returns `nil, err`.

The other function, `runloop.get_plugins_iterator`, still exists and is still being used when we just want to get the current iterator without (maybe) updating it first.

kong/init.lua
kikito: Answered in other thread, closing this one to keep the conversation focused.

kong/init.lua
kikito: Answered in other thread, closing this one to keep the conversation focused.

kong/runloop/handler.lua
kikito: I have changed how this works now (by using the `get_updated_plugins_iterator` field, which logs and serves the old version if needed).

kong/runloop/handler.lua
kikito: Using this to handle the "returning errors instead of logging inside the function" here.

As I see it, the purpose of this function is a single one: to trigger an update (which can be async or not). The rest (version checking, semaphore acquiring) are preparatory steps, needed to accomplish that. If one of them fails, then the function should make that manifest by erroring.

If `rebuild` itself starts deciding what to do with errors, we get into this situation in which the caller handles some of the errors, and `rebuild` handles others. I'd rather have one or the other do it, but not both.

For now I have decided to let the caller handle errors. It can just log the error and continue (like we do in `get_updated_plugins_iterator` now). Or it can decide to ignore one particular kind of error, like we do in timer-based rebuilds, where we ignore `"timeout"` errors which occur when attempting to get the lock.

kong/runloop/handler.lua
kikito: I see only one scenario where getting the version can fail: there is a problem with the cache.

The `:version` cache entry will rarely get evicted, and even if it does, a new `uuid` will be generated. So if you get a `nil` there must be a problem in the cache system (corrupted memory, that kind of thing).

If that's the case, then I think we should terminate stuff quickly. If there's a problem in the cache and there are a lot of requests arriving, we risk rebuilding the plugins iterator and the router once per request.

kong/runloop/handler.lua
kikito: I actually agree with part of what you are saying. This function failing should not mean always returning a bad request.

I just think that the decision of halting or not halting the request should not be done somewhere else. `rebuild` should just concern itself about rebuilding.

kong/runloop/handler.lua
kikito: I replaced this with a parameter. The parameter is set to 5 on timer-based rebuilds and to 1 on request-based rebuild. Request-based rebuilds use stale result if a semaphore fails, or a version fails.

kong/runloop/handler.lua
```diff
@@ -379,56 +532,50 @@ do
   end
 
 
-  update_plugins_iterator = function()
+  update_plugins_iterator = function(check_for_versions)
     local version, err = kong.cache:get("plugins_iterator:version", TTL_ZERO, utils.uuid)
```
bungle: It seems that versions are checked still here even when adding `check_for_versions`. Should we skip checking versions here too if requested?

kong/runloop/handler.lua
```diff
@@ -379,56 +532,50 @@ do
...
     if err then
       return nil, "failed to retrieve plugins iterator version: " .. err
     end
 
-    if not plugins_iterator or plugins_iterator.version ~= version then
+    if plugins_iterator and plugins_iterator.version == version then
```
bungle: same here?

kong/runloop/handler.lua
```diff
@@ -365,12 +375,155 @@ local function register_events()
...
+    if not current_version then
+      return nil, err
+    end
+    if current_version == version then
+      return true
+    end
+
+    local ok, err = acquire_semaphore(semaphore, timeout)
+    if not ok then
+      return nil, err
```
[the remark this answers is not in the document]
kikito: Done, now it NOTICEs on timeouts and CRITs everywhere else.

kong/runloop/handler.lua
kikito: Perhaps `check_for_versions` is not a good name.

What it means is: *while actually doing the build* (while making the `db.plugins:each(1000)`, or with routes), check from time to time if the version has changed.

Usually you want to do it but there are cases where you don't want to - when initializing stuff in declarative for example.

kong/runloop/handler.lua
```diff
@@ -672,8 +672,8 @@ do
 end
 
 
-local function balancer_setup_stage1(ctx, scheme, host_type, host, port,
-                                     service, route)
+local function balancer_prepare(ctx, scheme, host_type, host, port,
```
bungle: The `balancer_setup_stage2` is still there, perhaps that needs to be renamed to `balancer_execute` now too?

kong/runloop/handler.lua
```diff
@@ -705,7 +828,7 @@ local function balancer_setup_stage1(ctx, scheme, host_type, host, port,
 end
 
 
-local function balancer_setup_stage2(ctx)
+local function balancer_execute(ctx)
```
hishamhm: Regardless of whether this is a good change or not, why are these balancer functions being renamed (with no functionality change) in this PR? This seems totally unrelated.

bungle: @kikito, feel free to extract that to other pr.

kikito: Ok, extracted that commit as well as the localization to #4636

kikito:
Added some more changes after feedback:
* Removed the "rebuild" functionality completely - it overcomplicated things a lot and the benefit was not clear.
* Merged down the "auxiliary functions" used by `reload`
* Made REBUILD_TIMEOUT be built with direct instructions instead of by a function (the function might have confused other people thinking that they should use it)
* Replaced increase from 0.3 to 0.5 of `POLL_INTERVAL` in spec/02-integration/06-invalidations/03-plugins_iterator_invalidation_spec.lua by a more explicit `TIMER_REBUILDS` variable used when waiting for propagation.

kikito:
I had a talk with @bungle and he convinced me to port a small part of the changes we removed when removing the "retry" functionality: we don't retry any more, but we *do* wrap rebuilds with kong.db:connect() and kong.db:setkeepalive(). This allegedly helps with issues under high load, detailed in #4194

kikito:
After further discussion, we have decided to remove the commit which called `db:connect / db:keepalive` while doing rebuilds.

While we understand that this helped in some cases with Cassandra problems under load, we think that the issue should be solved by making changes to the `lua-cassandra` library, which has built-in "retry policies". Furthermore, there have been new releases of that library since the tests in #4194 were done. //cc @jeremyjpj0916 @bungle 

In addition to that, a bug was fixed which made unnecessary rebuilds of the plugins iterator (once per request).

jeremyjpj0916:
Specifically I think the issue was presented once paging has to occur so when you have 1000's of resources. Where the problem presented its self when Kong didn't sticky persists to a given C* node on those lookups for me. We noticed nil pointers in the C* logs from it, the prior issue tracks all of what we found though and I will be happy to retest in a RC what the behavior of 1.2 without the db keepalive (and well if problems still persist I will probably just drop in the db:connect/ db:keepalive manually into my 1.2 deploy when I am upgraded to it). This would be a non-issue to postgres kong instances as well so I could see the hesitation in doing so at a global anyways but it might be needed for C* at the moment. Ofc there are benefits to keepalive like less tls handshaking between kong on all these longer running queries when dealing with 1000's of resources as well. 

Looking forward to testing it all!
Cheers,
Jeremy

bungle:
@jeremyjpj0916,

The `NPE` problem should be handled without need to call `connect`, in most scenarios by using request aware lb (`cassandra_lb_policy`): https://github.com/Kong/kong/blob/master/kong.conf.default#L490-L503

jeremyjpj0916:
@bungle Just as an FYI the policy we have been running even when getting the errors was in fact "RequestDCAwareRoundRobin" as it seemed ever since it came out to be the most performant option based on description.

bungle:
@jeremyjpj0916,

There is still a slight chance to run to `NPE` especially when performance testing. That practically happens when all Cassandra connections on a node are used, and you need to grab a fresh new connection, which may end up in different node. Though we don't know if that is a real-world problem. Perhaps, and we can always add the `:connect` if needed. It has some tradeoffs.

hishamhm:
Continued in #4639! We're almost there!

hishamhm:
(Closing the PRs but leaving the branches around for reference now, will cleanup the branches once the router async project is fully merged)

PR #4632, opened 2019-05-18
feat(pdk) implemented 'kong.node.get_memory_stats()'

thibaultcha:
- Move memory reporting into a standardized PDK function
- Some updates to the memory reporting result including better semantics for clients (Lua numbers instead of strings)
- Update Admin API `/status` endpoint to use aforementioned PDK function


kikito:
Rebased this branch over `next`, fixing a conflict.

PR #4633, opened 2019-05-18
feat(*) HTTPS Routes can now be matched by SNI

thibaultcha:


spec/01-unit/01-db/01-schema/06-routes_spec.lua
```diff
@@ -924,7 +953,7 @@ describe("routes schema", function()
         assert.falsy(ok)
         assert.same({
           ["@entity"] = {
-            "must set one of 'methods', 'hosts', 'paths' when 'protocols' is 'http' or 'https'"
+            "must set one of 'methods', 'hosts', 'paths', 'snis' when 'protocols' is 'http' or 'https'"
```
hishamhm: this message is not exactly true (or, at the very least, can lead to a wrong interpretation: when protocols is `http` then `snis` is not an option). I think a different check order in the schema could give us better error messages, I'll try my hand at it.

thibaultcha: There could also be a way to rephrase this to carry the nuance without needing much fancier validation (`snis` is already validated as being `https`/`tls` only).

PR #4639, opened 2019-05-21
perf(runloop) pure async rebuilds with config option

kikito:
This is a continuation of #4627, but has been kept separate in order to make merging them separately easier.

This PR introduces the `router_consistency` config option which, if set to `eventual` completely decouples the building of the router and plugins iterator from the treatment of each request - Kong instead relies on the fact that rebuilds are triggered every second via a timer. The default remains `strict`, which matches the same level of router consistency of as that of past Kong releases.

This was extracted from #4488

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -66,9 +66,9 @@ local TTL_ZERO = { ttl = 0 }
 
 local get_plugins_iterator, build_plugins_iterator, update_plugins_iterator
 
-local get_router, build_router
+local get_router, build_router, update_router
 local server_header = meta._SERVER_TOKENS
-local _set_check_router_rebuild
+local _set_build_router
 
 local build_router_timeout
 
@@ -522,7 +522,8 @@ do
     return service
   end
 
-  build_router = function(db, version)
+  build_router = function(version)
+    local db = kong.db
     local routes, i = {}, 0
 
     local err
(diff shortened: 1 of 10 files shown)


kong/runloop/handler.lua
```diff
@@ -708,13 +708,18 @@ do
...
+    return rebuild("router", update_router, router_version, router_semaphore, timeout)
   end
 
 
-  rebuild_router = function(timeout)
-    return rebuild("router", update_router, router_version, router_semaphore, timeout)
+  get_updated_router = function()
+    local ok, err = rebuild_router(REBUILD_TIMEOUT)
+    if not ok then
+      log(ERR, "no router to route request (reason: ", err, ")")
```
kikito: There's a difference here between `get_updated_plugins_iterator` and this new `get_updated_router` and : the former logs the error *and returns the old plugins iterator*, while this one cancels the request and returns a 500 error right away. I am not sure which one is the best approach: erroring is more "correct" while returning the old one is "more resilient".

bungle: I would expect consistency on both. I have no hard opinion on this, but would lean toward resiliency (the new build is tried again).

kikito: After reading #4194 and discussing this privately with you, I decided to make the change so that both behave in the same way (logging and returning old instead of failing the request)

I have also decided to move that change to #4627 instead of keeping it here, because it looks non-controversial enough to me.

hishamhm:
Here are some preliminary performance testing results, made using [miniperf](http://github.com/kong/miniperf): 

When running load testing with `wrk` and a background task constantly creating routes and plugins on the background, Kong 1.1.2 (as a baseline) had a ~40% reduction in overall RPS compared to running without the "background noise"; with this branch, with `strict` consistency set had a ~25% slowdown (benefiting from the background async rebuilds even with strict consistency), with `eventual` consistency it had a ~15% slowdown. Most importantly, max latency went down by an order of magnitude.

PR #4649, opened 2019-05-24
feat(*) support custom logging serializers

p0pr0ck5:
### Summary

This commit introduces a new DAO entity, 'log_serializers', to support logging plugins writing output in customizable, arbitrary format.

Previously, bundled logging plugins made use of a serializer class to generate a JSON-encodable log message, to be shipped out the respective interface (TCP/UDP socket, file, etc). This provided a simple format to develop against, but limited flexibility in usage due to the inability to customize this format as defined as Kong source code.

The log_serializers objects are comprised of identifiers (ID, name) and a chunk field that shall contain a base64-encoded Lua string. Chunks are expected to expose a function 'serialize' that returns data to be written by the logging plugins. Plugins themselves are given a new config field 'serializer', denoting the
name of the serializer that should be used on the plugin. The legacy serializer module at  `kong.plugins.log-serializers.basic` is kept in the project for backwards compatibility, and plugins fall
back to this serializer when no serializer is defined on the plugin (the default behavior, and the behavior inherited during upgrade).

Note that given the imminent release of 1.2, I imagine this would be targeted for 1.3, hence creating a new migrations file.

### Full changelog

* Add `log_serializers` DAO
* Update `datadog`, `file-log`, `http-log`, `statsd`, `syslog`, `tcp-log`, and `udp-log` plugins to accept a config field `serializer`, which references the above object
* Write tests for `log_serializers` custom validation and cached `loadstring` results invalidations.


p0pr0ck5:
with this design, fetching the serializer from the db during the access phase is a bit anachronistic and introduces more latency (when used) to the proxy handler. I feel this is acceptable for a few reasons:

- We attempt to be as correct as possible based on the users explicitly desired configuration
- The serializer could be prefetched, either by the core utility that does cache fetches, or by the plugins via an init_by_worker handler, so the access phase calls in each plugin would essentially be nops
- Doing the fetch in the background thread would require we pass (copy) all of `ngx` To the timer scope, which I don’t even is possible

hishamhm:
@p0pr0ck5 before diving into any more specific code review, I'm going to leave some general architectural comments here. @thibaultcha and I discussed this PR and agreed that this feature is very welcome, and that the whole of functionality that this brings (extensibility through custom serializers + loading extensions via the DB) would be better split in two parts.

So the suggestion is to just do the first part here (extensibility through custom serializers), and just load the code via a plain `require`, and drop the new DAO entity. Instead, the `serializer` entry in configuration for plugins would be a simple string with the serializer name, and then plugins would issue a `require` based on that name (not 100% sure where exactly to store it — but if `basic` is any indication, I'd assume `"kong.plugins.log-serializers." .. name`). Logging serializers could then be installed via LuaRocks just like plugins (essentially they would be "plugins for (logging) plugins").

Once we have separately-loadable serializers like this, I think pretty soon people will want their custom serializers to emit data in different formats (XML, etc.) — as we talked about before, right now the `basic_serializer.serialize` is a misnomer, because it doesn't really serialize, it returns a Lua table with the info to be said, and then various log plugins are doing `cjson.encode` on that, locking them to JSON. It would make sense to have the JSON-encoding to be part of the serializer's concern. For backwards compatibility, we can't just change `basic_serializer.serialize` (plus, some plugins like statsd actually need the data in table form), so a suggestion is to make serializers expose two functions, `my_serializer.to_table` (that does what `basic_serializer.serialize` does, so `serialize` would be a backwards-compat alias to it), and `my_serializer.to_string`, which would do the proper encoding (names are open to bikeshedding, it just felt weird to write `serialize_to_table` ;) ).

The second thing that this PR does (loading Lua extensions via the DB) is something that could be handled in a more general manner (possibly extending to plugins in general even), outside of the scope of this PR.

hishamhm:
@p0pr0ck5 Ping :) Are you planning to move this forward? Or should we close it in the meantime and then revisit it some other time?

p0pr0ck5:
@hishamhm thanks for the ping. I won't have time to focus on this in the near future. I will close it and we can revisit later on. Thanks.

PR #4653, opened 2019-05-27
refactor(router) move path construction to after_access

Tieske:
The router was responsible for path construction but this would
happen in the before-access phase. Allowing path manipulation to
only be done on the full path, not on the individual components.

This commit changes it by creating an intermediate table with the
path components (for plugins to modify) and does the final
construction only in the after-access phase

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -1185,35 +1185,15 @@ function _M.new(routes)
           end
 
           if matched_route then
+            local request_postfix
             local upstream_host
-            local upstream_uri
             local upstream_url_t = matched_route.upstream_url_t
             local matches        = ctx.matches
-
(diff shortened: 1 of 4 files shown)


hishamhm:
@Tieske could you take a look at the failing tests before we review?

hishamhm:
@Tieske looks like the tests are still failing

hishamhm:
@Tieske ping?

Tieske:
leaving the branch around, just-in-case

windmgc:
@Tieske Just out of curiosity, why this PR didn't get merged?

Tieske:
because it never passed and got outdated (again), and hasn't gotten priority yet. It is still on the radar however

PR #4657, opened 2019-05-27
fix(declarative) fix db_import upsert via deterministic v5 UUIDs

hishamhm:
Whenever an entity was inserted via declarative config without an explicit `id` given in the YAML file, the loader auto-generated a random UUID. When using `kong config db_import`, this meant that each time one loaded a file, it would generate new entries, even though the implementation of `db_import` used `upsert`.

With this commit, when an `id` is not given but the "endpoint-key" unique identifier is given to an entity (e.g. `name` for `service`, `username` for `consumer`, etc.), this endpoint-key is used to generate a deterministic UUID v5.

This allows upserts to work as expected when using `kong config db_import`.

As discussed with @kikito, we cannot easily use `upsert_by_<unique>` here because the id's are auto-generated at a different step, so it is not easy to tell apart the id's given by the user from the auto-generated ones (the code doing so would be much more convoluted).

The first two commits contain the fixes and the tests and make it easier to review the code changes separately. The other three are code cleanups with no behavior changes (these can be dropped or merged later if deemed unnecessary or risky for 1.2.0).

### DIFF
--- a/kong/db/schema/others/declarative_config.lua
+++ b/kong/db/schema/others/declarative_config.lua
@@ -1,3 +1,4 @@
+local uuid = require("resty.jit-uuid")
 local utils = require("kong.tools.utils")
 local Errors = require("kong.db.errors")
 local Entity = require("kong.db.schema.entity")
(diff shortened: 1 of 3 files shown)


kong/db/schema/others/declarative_config.lua
```diff
@@ -339,6 +340,110 @@ local function validate_references(self, input)
...
+      parent_fk = parent_schema:extract_pk_values(input)
+      child_key = foreign_children[parent_entity][entity]
+    end
+
+    local entity_schema = all_schemas[entity]
+    for _, item in ipairs(input[entity]) do
+
+      local key
+      -- if entity schema has a non-composite PK
+      if #entity_schema.primary_key == 1 then
```
kikito: Not a deal breaker, but I think this block of nested ifs are complex enough to be their own function - one which returns the `key` (either from the endpoint_key or from `build_cache_key`).

```
local pk_name = entity_schema.primary_key[1]

local key = extract_key_from_entity(entity, item, entity_schema, parent_fk, child_key)
if key then
  item[pk_name] = generate_uuid(entity_schema.name, key)
end
```

kong/db/schema/others/declarative_config.lua
```diff
@@ -392,55 +404,37 @@ local function generate_uuid(namespace, name)
...
-    local child_key
-    if parent_entity then
-      local parent_schema = all_schemas[parent_entity]
-      if parent_schema.fields[entity] then
-        goto continue
-      end
-      parent_fk = parent_schema:extract_pk_values(input)
-      child_key = foreign_children[parent_entity][entity]
-    end
+    local schema = all_schemas[entity]
```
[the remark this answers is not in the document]
hishamhm: I chose not to do this because this would make the new parameter dependent on the first (given an `entity` string, there would only ever be one correct value for `schema`, so this is not really a "parameter" as in "something you can tweak"). The fact that both uses of `traverse_nested` use that same value is an implementation coincidence of the functions, not a property of the `traverse_nested` contract.

kong/db/schema/others/declarative_config.lua
hishamhm: Changed the code, looks much nicer!

hishamhm:
I rebased this to `release/1.2.0` for rc2 and dropped the refactor commits. I will bring them back in a separate PR!

PR #4660, opened 2019-05-28
chore(ci) bump LuaRocks to 3.1.2

thibaultcha:


hutchic:
at some point hopefully these values only exist in one place but right now I think `.requirements` as well and they seem to have diverged

thibaultcha:
@hutchic Good point, thanks!

PR #4674, opened 2019-05-31
perf(runloop) warmup dns on router rebuilds

bungle:
### Summary

Just an alternative approach to #4656.

- Warms up all the workers DNS on startup (not just worker 0)
- Warms up all services DNS on router rebuild

Not sure if it is worth it. DNS will get warmed on use after all. But if someone wants to benchmark it here is a PR that you can try. You should perhaps benchmark the initial latency of a first request and latency of request after adding a service with a new host.

But... I don't know if this is really wanted, and if not I propose even removing DNS warming on worker 0 during `service` entities warming.

### DIFF
--- a/kong/cache_warmup.lua
+++ b/kong/cache_warmup.lua
@@ -1,19 +1,12 @@
 local utils = require "kong.tools.utils"
 
 
-local cache_warmup = {}
-
-
 local tostring = tostring
 local ipairs = ipairs
 local math = math
+local type = type
 local kong = kong
 local ngx = ngx
-
-
-function cache_warmup._mock_kong(mock_kong)
-  kong = mock_kong
-end
 
 
 local function warmup_dns(premature, hosts, count)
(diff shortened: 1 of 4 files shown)


kong/cache_warmup.lua
```diff
@@ -128,4 +113,36 @@ function cache_warmup.execute(entities)
...
+function cache_warmup.warm_services_dns()
+  if not kong.cache then
+    return true
+  end
+
+  local host_count = 0
+  local hosts_array = {}
+  local hosts_set = {}
+
+  for entity, err in kong.db.services:each(1000) do
```
p0pr0ck5: I think we’ve already paginated through all relevant services as part of the router rebuild, it would be less pressure on the database to have this function receive a set of hostnames, rather than generate it again without context.

bungle: @p0pr0ck5 yes, there are 3 places:
1. router initial build (which happens with db on init, so cannot be used)
2. services warmup (the warmup in general is limited to worker=0, so could return the data, but makes interface a bit ugly).

Robert, if you have any time, I guess best way to bring this forward is to have some numbers. The numbers I have is that it makes any new request basically latency 0, also on router rebuilds. But does it really matter... perhaps not. Certainly may look better in some tests, but gains in a realworld are not that huge I guess, or are they?

PR #4681, opened 2019-06-04
feat(api) create service when creating a route

bungle:
### Summary

Creates service when creating an route, e.g. this works:
```
http :8001/routes paths=/ service.url=http://www.example.org/ -f
```

This has shortcomings, e.g. this:
```
http :8001/routes paths=/ service.name=existing-service service.url=http://www.example.org/ -f
```

May resolve by name to existing-service (which may have different url than `http://www.example.org/`, and the `url` will not be updated, should it)?

Also
```
http :8001/routes paths=/ service.id=<inexisting-uuid> service.url=http://www.example.org/ -f
```

Gives error about id not found (this can be fixed with read-before write, or perhaps by doing upsert).

So this is more for discussion. It is usable as is, but may cause make API bit strange on some edges. Perhaps we can work it out here.

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -198,15 +198,16 @@ local function resolve_foreign(self, entity)
       local value = entity[field_name]
       if value and value ~= null then
         if not schema:validate_primary_key(value, true) then
+          local resolved
           local resolve_errors = {}
           local has_resolve_errors
+          local dao = self.db[schema.name]
           for unique_field_name, unique_field in schema:each_field() do
             if unique_field.unique or unique_field.endpoint_key then
               local unique_value = value[unique_field_name]
               if unique_value and unique_value ~= null and
                  schema:validate_field(unique_field, unique_value) then
 
-                local dao = self.db[schema.name]
                 local select = dao["select_by_" .. unique_field_name]
                 local foreign_entity, err, err_t = select(dao, unique_value)
                 if err_t then
@@ -214,6 +215,7 @@ local function resolve_foreign(self, entity)
                 end
 
                 if foreign_entity then
+                  resolved = true
                   entity[field_name] = schema:extract_pk_values(foreign_entity)
                   break
                 end
@@ -226,6 +228,14 @@ local function resolve_foreign(self, entity)
 
                 has_resolve_errors = true
               end
+            end
+          end
+
+          if not resolved and field.creatable then
+            local foreign_entity = dao:insert(value)
+            if foreign_entity then
+              entity[field_name] = schema:extract_pk_values(foreign_entity)
+              has_resolve_errors = false
             end
           end
 
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -108,7 +108,7 @@ return {
                        },
                      }, },
     { tags             = typedefs.tags },
-    { service = { type = "foreign", reference = "services" }, },
+    { service = { type = "foreign", reference = "services", creatable = true }, },
   },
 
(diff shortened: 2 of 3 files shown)


subnetmarco:
@bungle is this limited to `services` and `routes` or can the dot notation be used for any other child entity?

bungle:
@subnetmarco the implementation is generic, but you need to enable it, you need to add that `creatable` to the foreign field in schema, like:
```
{ service = { type = "foreign", reference = "services", creatable = true } },
```

It is implemented in `DAO` level so you can run code like this:
```
kong.db.routes:insert{
  paths = { "/" },
  service = {
    url = "http://www.example.org/"
  }
}
```

But it will only work with foreign-fields that are marked as `creatable`.

bungle:
Dot notation can be used with things like `record` field, like plugin configuration. The thing with `application/x-www-form-urlencoded` and to some extend `multipart/form-data` (the current implementation for `multipart/form-data` does not support multiple fields of same name, because of Lapis).

With those mime-types we also do type inferring.
https://github.com/Kong/kong/blob/next/kong/api/arguments.lua#L173-L232

E.g. sending:
```
http :8001/plugins name=imaginary-plugin config.mark[1].relation[1].condition[1].mark_key=3
```
Is equal to (because of inferring):
```
http :8001/plugins name=imaginary-plugin config.mark.relation.condition.mark_key=3
```
And can also be specified as:
```
http :8001/plugins name=imaginary-plugin config.mark[].relation[].condition[].mark_key=3
```

bungle:
For the reviewers:

Let's discuss about it. Is the direction right, and how can we make it even better. Don't try to verify this as it is going right to `next`. There are tests to be added, but first some discussion about it. Ideas what we should do for short-comings, etc.

kikito:
My first comment that the path is `/routes`, but I think the parent should go first. Several reasons:

* It would also be more aligned with declarative config, where we do parent-first.
* By doing child-first, we are forcing it to create only 1 parent with 1 children. Most relationships in Kong allow for multiple children.

Nothing prevents us from accepting a `routes` pseudo-attribute which takes an array:

```
http POST :8001/services url=http://example.net routes:='[{"paths": ["/"]}]'
```

This can be easily extended to `service.plugins` as well. (Or `upstream.targets` and `certificate.snis`. And it allows more than 1 children.

---

Taking this a bit further in the declarative config direction, we could have some sort of endpoint which just accepts a kind-of-declarative-config file:

```
http POST :8001/bulk '{ "services": [ { "url": "http://example.net, routes: [{ "paths": ["/"]}] } ]'
```

Or even:

```
echo << EOF
services:
  url: http://www.example.net
  routes:
    paths: [ "/" ]
EOF | http POST :8001/bulk Content-Type:application/x-yaml
```

or

```
http POST :8001/bulk Content-Type:application/x-yaml @myfile.yaml
```

bungle:
@kikito, I think we can do both. The current proposal follows the usual pattern as we have right now:
```
http post :8001/services/example/routes paths=/ -f
```
or 
```
http post :8001/routes paths=/ service.name=example -f
```

Where the referenced entity is created first.

kong/db/dao/init.lua
```diff
@@ -229,6 +231,14 @@ local function resolve_foreign(self, entity)
             end
           end
 
+          if not resolved and field.creatable then
+            local foreign_entity = dao:insert(value)
```
hishamhm: what if the creating the route fails?

hishamhm:
This was valuable discussion which we'll keep in the archives, but the actual implementation for this is most likely going to be different, so we'll close this one for now.

PR #4683, opened 2019-06-04
Feat/change rate lim opts for dbless

locao:
### Summary

rate-limiting plugin now accepts only "local" or "redis" as policy options when kong is running in dbless mode.

### Full changelog

* rate-limiting plugin schema now checks the database strategy to define acceptable policies.
* Added test case to validate policies usage in different database strategies.

### DIFF
--- a/kong/plugins/rate-limiting/schema.lua
+++ b/kong/plugins/rate-limiting/schema.lua
@@ -1,3 +1,4 @@
+local kong = kong
 local typedefs = require "kong.db.schema.typedefs"
 
 
@@ -22,6 +23,10 @@ local function validate_periods_order(config)
   return true
 end
 
+local conf_db = kong and kong.configuration and kong.configuration.database
+local default_policy = conf_db == "off" and "local" or "cluster"
+local valid_policies = conf_db == "off" and { "local", "redis" } or
+                      { "local", "cluster", "redis" }
 
 return {
   name = "rate-limiting",
@@ -44,9 +49,9 @@ return {
           }, },
           { policy = {
               type = "string",
-              default = "cluster",
+              default = default_policy,
               len_min = 0,
-              one_of = { "local", "cluster", "redis" },
+              one_of = valid_policies,
           }, },
           { fault_tolerant = { type = "boolean", default = true }, },
           { redis_host = typedefs.host },
--- a/spec/03-plugins/23-rate-limiting/01-schema_spec.lua
+++ b/spec/03-plugins/23-rate-limiting/01-schema_spec.lua
@@ -1,3 +1,4 @@
+local helpers = require "spec.helpers"
 local schema_def = require "kong.plugins.rate-limiting.schema"
 local v = require("spec.helpers").validate_plugin_config_schema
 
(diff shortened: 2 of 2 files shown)


kong/plugins/rate-limiting/schema.lua
```diff
@@ -44,9 +48,9 @@ return {
           }, },
           { policy = {
               type = "string",
-              default = "cluster",
+              default = "local",
```
hbagdi: Thinking out loud: can this be classified as a breaking change?

To be clear, I'm not advocating against it but would like to clarify the compat promise that Kong 1.0 brings with it.

hishamhm: Agreed, the default should change only on DB-less.

locao: Thanks for pointing that, @hbagdi. The default policy is `"local"` only for DB-less now.

kong/plugins/rate-limiting/schema.lua
```diff
@@ -22,6 +23,10 @@ local function validate_periods_order(config)
   return true
 end
 
+local conf_db = kong and kong.configuration and kong.configuration.database
+local default_policy = conf_db == "off" and "local" or "cluster"
+local valid_policies = conf_db == "off" and { "local", "redis" } or
+                      { "local", "cluster", "redis" }
```
thibaultcha: Many other plugins will require a similar update. An approach I have suggested before that would make updating those plugins easier and more consistent between each other would be to implement the proper typedefs for DB strategies (as a single typedef or several ones, depending on what's best).

Such a typedef would also be implemented with Kong schema idioms providing appropriate error messages in case of mis-configuration (e.g. "cannot configure policy 'cluster' when 'database = off'").

Where this gets much better, is that this typedef would be able to implement precedence between plugin config specified values (e.g. `config.policy.postgres.host = ...`) and `kong.conf` values (e.g. a fallback to `pg_host = ...`). This is what would enable "complete" declarative configuration support for DB-less Kong nodes, in which rate-limiting and/or authentication plugins would still be able to connect to a database if needed to. This is, naturally, and as we expected, a highly-requested feature for DB-less nodes (I am talking about the ability to configure a DB-less Kong node and have plugins still connect to a database).

In short, such a typedef could, in order of implementation/iterations:

1. Be implemented as `typedefs.db_policy`, and conditionally change its `default` and accepted values based on `kong.conf`.
2. Be updated to support `typedefs.db_policy(...?)` where if a given plugin does not support a given policy, the list of arguments would restrict the list of supported policies for this field.
3. Be able to receive the connection properties of a given DB instance, and fallback to `kong.conf` specified values if desired. (Plugins would need to be able to instantiate DAOs or lightweight DB accessors)

hishamhm: This plan is very interesting, but also has a much larger scope. In the spirit of incremental improvement, let's go on with this PR as is, and eventually plan for this larger scheme for DB-aware schemas.

thibaultcha: I’d argue that in the spirit of incremental improvement, the above plan was split in steps, of which the first would be a relatively fast implementation, and would allow to update multiple plugins at once before the next release without duplication and setting up the proper foundations for a larger vision. 

But up to you guys.

thibaultcha: Adding to the above that proper error messages when trying to configure unsupported storage strategies in DB/DB-less modes should be a must-have imho, instead of changing the hard-coded schema, which is providing an incoherent behavior between each mode to users. Was the intent to ship this in 1.2.0?

PR #4685, opened 2019-06-04
feat(plugin) add session plugin

darrenjennings:
Adds [session](https://github.com/Kong/kong-plugin-session) plugin to bundled plugins:
Relevant Session Plugin PR: https://github.com/Kong/kong-plugin-session/pull/9

### DIFF
--- a/kong-1.1.2-0.rockspec
+++ b/kong-1.1.2-0.rockspec
@@ -43,7 +43,7 @@ dependencies = {
   "kong-plugin-serverless-functions ~> 0.2",
   "kong-prometheus-plugin ~> 0.3",
   "kong-proxy-cache-plugin ~> 1.2",
-  "kong-plugin-session == 1.0.2-3",
+  "kong-plugin-session ~> 2.0",
 }
 build = {
   type = "builtin",
--- a/kong-1.1.2-0.rockspec
+++ b/kong-1.1.2-0.rockspec
@@ -43,6 +43,7 @@ dependencies = {
   "kong-plugin-serverless-functions ~> 0.2",
   "kong-prometheus-plugin ~> 0.3",
   "kong-proxy-cache-plugin ~> 1.2",
+  "kong-plugin-session == 1.0.2-3",
 }
 build = {
   type = "builtin",
--- a/kong/constants.lua
+++ b/kong/constants.lua
@@ -33,6 +33,7 @@ local plugins = {
   "post-function",
   "prometheus",
   "proxy-cache",
+  "session",
 }
 
 local plugin_map = {}
--- a/spec/01-unit/12-plugins_order_spec.lua
+++ b/spec/01-unit/12-plugins_order_spec.lua
@@ -51,6 +51,7 @@ describe("Plugins", function()
       "zipkin",
       "bot-detection",
       "cors",
+      "session",
       "kubernetes-sidecar-injector",
       "jwt",
       "oauth2",


kong-1.1.2-0.rockspec
```diff
@@ -43,6 +43,7 @@ dependencies = {
   "kong-plugin-serverless-functions ~> 0.2",
   "kong-prometheus-plugin ~> 0.3",
   "kong-proxy-cache-plugin ~> 1.2",
+  "kong-plugin-session == 1.0.2-3",
```
thibaultcha: We don't pin these plugins versions since we control their release cycle, and leaving the patch version unspecified allows us to release bugfixes to those installation methods that rely on `luarocks install` without having to cut a whole new Kong release (we also never include the revision number since it should be left to the maintainer's discretion to update if necessary). Could we specify `~> 1.0` here instead?

thibaultcha:
The plugin seems to be tested against Kong's master branch (https://github.com/Kong/kong-plugin-session/blob/master/.travis.yml#L51) - could we run a sanity job against Kong's next branch to make sure it won't be broken in 1.2.0?

thibaultcha:
Given that the plugin's version was bumped to 1.0, and assuming that other environments depending on this plugin have pinned it to a 0.x version, we should consider removing the `BasePlugin` inheritance (and `:new()` implementation along with `.super` parent calls) in this plugin's 1.0 version (see https://github.com/Kong/kong-plugin-session/blob/master/kong/plugins/session/handler.lua#L9-L16). The pattern is not required anymore in Kong 1.2 (next branch) and plays a role in the 1.2 performance improvements. It is a very minor change to the plugin, see https://github.com/Kong/kong/commit/a29c3553f43dd53cfc78d618e653e3e7ae484dbb for an example on this repository's plugins.

We should also consider updating the plugin to use the PDK, which is a relatively easy task and brings a number of improvements to the plugin (e.g. abstracted req/resp getters/setters, prefixed logs or namespaced `ctx` tables, etc...), but that wouldn't be a blocker on my end.

darrenjennings:
@thibaultcha thanks for the great feedback. I have a working PR with pdk updates and removing the `BasePlugin` inheritance: https://github.com/Kong/kong-plugin-session/pull/9

WRT the version, will removing the base plugin constitute a minor bump of the plugin version to `1.1.0-1`? As I understand, the plugin inheritance is a breaking change so the new kong-plugin-session@1.1 will not work with Kong < 1.2?

thibaultcha:
> WRT the version, will removing the base plugin constitute a minor bump of the plugin version to `1.1.0-1`? As I understand, the plugin inheritance is a breaking change so the new kong-plugin-session@1.1 will not work with Kong < 1.2?

I noticed yesterday that the plugin was recently bumped to 1.x. My suggestion would be:

- If no other component requires version 1.x of this plugin, consider the `BasePlugin` class removal part of the 1.x release, and re-tag the release
- If another component already bumped its dependency on this plugin to 1.x, then release 2.x (since we follow semver in our bundled plugins' versioning, we should bump the major version number).

What do you think?

bungle:
@darrenjennings Minor bump is alright, if it doesn't get into `EE` or some older version that requires `BasePlugin`.

darrenjennings:
It is currently tagged in EE with explicit `"kong-plugin-session == 1.0.2-3"` so not a big deal for us. The breaking change is due to the Kong 1.2 peer dependency, so would be best if we could specify peer dependency version in luarocks, but we cannot. In npm you can specify peer dependency and it will error during install. So I am OK with minor bump as well per @bungle's suggestion, and updating the comment https://github.com/Kong/kong-plugin-session/blob/1bb43edc52357903e54e939655d20cd0af3afb16/kong-plugin-session-1.0.2-3.rockspec#L21

thibaultcha:
@darrenjennings I'd strongly advise against shipping breaking changes in a patch/minor release.

thibaultcha:
@darrenjennings Along with this change, we also need a docs update to [this page](https://github.com/Kong/docs.konghq.com/blob/release/1.2.x/app/1.2.x/plugin-development/custom-logic.md#plugins-execution-order) (note the document linked to points to the 1.2.x version of this page, which is where the change should be pushed to).

darrenjennings:
thanks for the link. I've bumped the session plugin to 2.x per your advice and latest changes for session PR is back up here: https://github.com/Kong/kong-plugin-session/pull/9

Also added a PR for the session docs:
https://github.com/Kong/docs.konghq.com/pull/1318

darrenjennings:
Also, the tests are passing on the next branch:
https://github.com/Kong/kong-plugin-session/runs/142688711

thibaultcha:
Great! Since the 2.0 version of the plugin is not published yet, it seems like CI is failing, otherwise this looks good!

hishamhm:
If we're adding this in 1.2.0, let's hold from merging it here and I'll cherry-pick it into `release/1.2.0`

hishamhm:
> The breaking change is due to the Kong 1.2 peer dependency, so would be best if we could specify peer dependency version in luarocks, but we cannot. In npm you can specify peer dependency and it will error during install.

Fun fact: LuaRocks does in fact have "peer dependencies" since 2008(!). The syntax is prepending the operator with `@` (as in `kong @> 1.2`) and this was added back in the day for Sputnik plugins: http://lists.luaforge.net/pipermail/luarocks-developers/2008-July/000759.html

It seems we just never documented the feature!? I can't find it in the LuaRocks wiki, at least. We do not have dependency alternatives, though (i.e. `or` operator in dependencies), so adding the above might not be advisable for plugins if we end up having different version constraints for Kong and Kong Enterprise.

PR #4687, opened 2019-06-05
feat(conf) allows to specify `ssl_protocols` in configuration

bungle:
### Summary

You can specify configuration as:
`KONG_SSL_PROTOCOLS=` -> we will not output `ssl_protocols` in config (uses Nginx defaults)
`KONG_SSL_PROTOCOLS="TLSv1, SSLv3"`
`KONG_SSL_PROTOCOLS="tlsv1   sslv3"`

Doesn't affect the stream module, as many of the SSL related options
doesn't currently.

Uses same protocols suite on both admin and proxy, didn't feel like
we need separate for admin api.


p0pr0ck5:
I imagine we’d want separate protocol definitions for admin and proxy for two reasons:

1. It keeps with the existing pattern we have
2. There’s a valid use case for separate definitions, eg supporting legacy proxy clients on TLS v1 but requiring stronger encryption for clients managing the admin api

spec/01-unit/03-conf_loader_spec.lua
```diff
@@ -670,6 +670,44 @@ describe("Configuration loader", function()
           assert.is_nil(err)
           assert.same(http_tls.modern_cipher_list, conf.ssl_ciphers)
         end)
+        it("defines ssl_protocols by default", function()
+          local conf, err = conf_loader(nil, {})
+          assert.is_nil(err)
+          -- looks kinda like a cipher suite
```
p0pr0ck5: Pedantic note that this comment doesn’t make sense/is incorrect here

spec/01-unit/03-conf_loader_spec.lua
```diff
@@ -670,6 +670,44 @@ describe("Configuration loader", function()
...
+        end)
+        it("explicitly defines multiple ssl_protocols (space separated)", function()
+          local conf, err = conf_loader(nil, {
+            ssl_protocols = "SSLV3  tlsv1"
+          })
+          assert.is_nil(err)
+          -- looks kinda like a cipher suite
+          assert.equal("SSLv3 TLSv1", conf.ssl_protocols, nil, true)
+        end)
+        it("doesn't define ssl_protocols when no protocols specified (uses nginx defaults)", function()
```
p0pr0ck5: In this test case it seems we fall back to the default value of http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_protocols, which defines TLS v1 (unlike our current posture where we set a minimum protocol version of 1.1). Do we want to stick with the existing default, rather than the nginx defaults (such a change would probably catch folks deploying 1.2 off guard if they regularly test their handshake capabilities)

bungle:
> I imagine we’d want separate protocol definitions for admin and proxy for two reasons:
> 
> 1. It keeps with the existing pattern we have
> 2. There’s a valid use case for separate definitions, eg supporting legacy proxy clients on TLS v1 but requiring stronger encryption for clients managing the admin api

There is no existing pattern. Both are seen in configs.

spec/01-unit/03-conf_loader_spec.lua
bungle: Existing default is still used as default. To fallback to Nginx default you need to actively ask for it.

bungle:
@p0pr0ck5, I refer to this:
https://github.com/Kong/kong/blob/master/kong/templates/nginx_kong.lua#L98
https://github.com/Kong/kong/blob/master/kong/templates/nginx_kong.lua#L210

Same thing is used on both Admin and Proxy, and I felt like this would probably go in same. If we get people asking about different protocols enabled on admin and proxy, then we can add admin later on? Or is there huge push for it now? It is minimal change.

p0pr0ck5:
Ah I was thinking of certificates- different default certs on the admin and proxy side. I guess for ciphers and protocol it probably makes sense to be consistent 👍

kong/templates/nginx_kong.lua
```diff
@@ -93,7 +93,9 @@ server {
 > if proxy_ssl_enabled then
     ssl_certificate ${{SSL_CERT}};
     ssl_certificate_key ${{SSL_CERT_KEY}};
-    ssl_protocols TLSv1.1 TLSv1.2 TLSv1.3;
+> if ssl_protocols then
+    ssl_protocols ${{SSL_PROTOCOLS}};
```
thibaultcha: In several PRs already, we have discussed the topic of adding new Kong configuration properties mapping to Nginx directives. It has been established each time that this model does not work well for us, nor for our users.

If I recall correctly, we had agreed at least twice already that:

1. If new Nginx directives were to be added to our template, we should leverage *dynamically injected nginx directives*: add a new "Nginx directives" section to the `kong.conf.default` template, and add a new default property of, e.g. `nginx_http_large_client_header_buffers = 8 24k`. This not only allows us to specify our own defaults for Nginx directives, but it also advertises dynamically injected Nginx directives, *and* allows us to move away from the current paradigm of adding a configuration property for each directive, which is painful for several reasons.
2. That we should deprecate our existing 1-1 Kong configuration properties in favor of new default `nginx_*` properties resulting in underlying dynamic injections out of the box.

Can we invest some time in the above plan instead of digging ourselves a slightly deeper hole? :sweat_smile:

PR #4688, opened 2019-06-05
feat(conf) allows to specify `ssl_dhparam` from pre-defined groups

bungle:
### Summary

Allows to specify `ssl_dhparam` from pre-defined groups or from absolute path.

The autogenerated and predefined groups are:
- `ffdhe2048`
- `ffdhe3072`
- `ffdhe4096`

See: https://wiki.mozilla.org/Security/Server_Side_TLS#Pre-defined_DHE_groups

By default it is set to `ffdhe2048`, but users can remove the parameter
from being injected to Kong configuration by specifying:
`KONG_SSL_DHPARAM=""`


p0pr0ck5:
My only note would be that the default behavior of Kong is to use the `modern` predefined TLS cipher suite, which does not specify any ciphers that perform a DH handshake, so defining these params are unused in the default behavior. But having them in isn't problematic, and I doubt it's worth the hassle to _not_ do this when `modern` cipher suite is enabled.

kong/templates/nginx_kong.lua
```diff
@@ -102,6 +102,9 @@ server {
     ssl_session_timeout 10m;
     ssl_prefer_server_ciphers on;
     ssl_ciphers ${{SSL_CIPHERS}};
+> if ssl_dhparam then
+    ssl_dhparam ${{SSL_DHPARAM}};
```
thibaultcha: Here, I simply want to link to the comment I wrote at https://github.com/Kong/kong/pull/4687#discussion_r290872846 which applies just as much to this change.

PR #4691, opened 2019-06-05
chore(deps) bump several plugins on removal of `BasePlugin` dependency

bungle:
### Summary

These were bumped:

- `kong-plugin-azure-functions` to `~> 0.4`
- `kong-plugin-kubernetes-sidecar-injector` to `~> 0.2`
- `kong-plugin-serverless-functions` to `~> 0.3`
- `kong-prometheus-plugin` to `~> 0.4`,

The `kong-plugin-zipkin` is still using `BasePlugin` and was not
bumped here as it uses multiple levels of inheritance.

There was no need to bump these plugins:

- `kong-proxy-cache-plugin`
- `kong-plugin-request-transformer`

Because they were moved to Kong as of `1.2.0`.

### DIFF
--- a/kong-1.2.0rc2-0.rockspec
+++ b/kong-1.2.0rc2-0.rockspec
@@ -37,11 +37,11 @@ dependencies = {
   "lua-resty-cookie == 0.1.0",
   "lua-resty-mlcache == 2.4.0",
   -- external Kong plugins
-  "kong-plugin-azure-functions ~> 0.3",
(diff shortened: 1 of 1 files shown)


bungle:
There was also another PR released to be merged to `next` by @hbagdi. If you want to merge that separately, then it can be removed from here. Also this PR cannot be merged before respective rocks are released. The prometheus is tagged and released already, but others are waiting for review. Let me also know if you want this to be rebased to `next` instead, @hishamhm?

PR #4692, opened 2019-06-05
feat(conf) add `deferred`, `bind` and `reuseport` listen flags

### LINKED ISSUE #4452: kong proxy listen cannot use reuseport?
### Summary

Kong start with error when I config proxy_listen with reuseport like this: 
```
KONG_PROXY_LISTEN: "0.0.0.0:8000 reuseport"
```
the error is:
```
Error: proxy_listen must be of form: [off] | <ip>:<port> [ssl] [http2] [proxy_protocol] [transparent], [... next entry ...]
```

### Steps To Reproduce

1. Configurate `kong.conf` with `proxy_listen: 0.0.0.0:8000 reuseport`

Thanks

bungle:
### Summary

On `http` you can now give new listen flags:
- `deferred`
- `bind`
- `reuseport`

On `stream` you can now give new listen flags:
- `bind`
- `reuseport`

Deferred is not supported on `stream` that's why it is not there.

### Issues resolved

Fix #4452

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -130,6 +130,14 @@
                          # - `transparent` will cause kong to listen to, and
                          #   respond from, any and all IP addresses and ports
                          #   you configure in iptables.
+                         # - `deferred` instructs to use a deferred accept on
+                         #   Linux (the TCP_DEFER_ACCEPT socket option).
+                         # - `bind` instructs to make a separate bind() call
+                         #   for a given address:port pair.
+                         # - `reuseport` instructs to create an individual
+                         #   listening socket for each worker process
+                         #   allowing a kernel to distribute incoming
+                         #   connections between worker processes
                          #
                          # This value can be set to `off`, thus disabling
                          # the HTTP/HTTPS proxy port for this node.
@@ -165,6 +173,12 @@
                          # - `transparent` will cause kong to listen to, and
                          #   respond from, any and all IP addresses and ports
                          #   you configure in iptables.
+                         # - `bind` instructs to make a separate bind() call
+                         #   for a given address:port pair.
+                         # - `reuseport` instructs to create an individual
+                         #   listening socket for each worker process
+                         #   allowing a kernel to distribute incoming
+                         #   connections between worker processes
                          #
                          # **Note:** The `ssl` suffix is not supported,
                          # and each address/port will accept TCP with or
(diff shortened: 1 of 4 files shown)


kong.conf.default
```diff
@@ -130,6 +130,14 @@
...
                          #   respond from, any and all IP addresses and ports
                          #   you configure in iptables.
+                         # - `deferred` instructs to use a deferred accept on
+                         #   Linux (the TCP_DEFER_ACCEPT socket option).
+                         # - `bind` instructs to make a separate bind() call
+                         #   for a given address:port pair.
+                         # - `reuseport` instructs to create an individual
+                         #   listening socket for each worker process
+                         #   allowing a kernel to distribute incoming
+                         #   connections between worker processes
```
dndx: We should mention "This should results in better load distributions between Kong workers and is highly recommended."

kong.conf.default
```diff
@@ -165,6 +173,12 @@
                          # - `transparent` will cause kong to listen to, and
                          #   respond from, any and all IP addresses and ports
                          #   you configure in iptables.
+                         # - `bind` instructs to make a separate bind() call
+                         #   for a given address:port pair.
+                         # - `reuseport` instructs to create an individual
+                         #   listening socket for each worker process
+                         #   allowing a kernel to distribute incoming
+                         #   connections between worker processes
```
dndx: Same here.

kong.conf.default
bungle: Yes, all opinions, should we enable by default?

kong.conf.default
bungle: Yes, again, should it be enabled by default? Does this concern us: http://man7.org/linux/man-pages/man7/socket.7.html

And yes, previously we did not even give option for these, so in some ways enabling them may lead to unwanted results. Perhaps in most cases to wanted results, though. I leave decision to others.

bungle: I tested the `reuseport` with our `miniperf` and here are results on my 2010 Mac Pro:
```
without reuseport: 25000 req/s
with reuseport: 8500 req/s
```

So it seems like on this machine the reuseport actually kills the performance. Other setting didn't make measurable difference.

dndx: @bungle It should be fine to enable it by default. The security implication refers to cases where people runs Nginx using incorrect user permission setup, which in that case is an even bigger security risk thank `SO_REUSEPORT` is.

The way `SO_REUSEPORT` is guarded is that the Kernel requires subsequent socket who tries to listen on the same port to belong to the same user as the first process who opened it. This should be enough to prevent any port hijacking from happening.

Other softwares, such as HAProxy, already enables it by default: https://www.haproxy.org/download/1.8/doc/configuration.txt (search `noreuseport`)

bungle: @dndx did you see above how it makee the performance suck? I need to retest on Linux.

kikito:
Approving this as it is: making the feature available but not enabled by default is a safe bet. The discussion about whether it should be enabled by default, or the performance tests, can be done in a different PR.

PR #4693, opened 2019-06-05
Various tweaks for 1.2.0

thibaultcha:


kong/concurrency.lua
```diff
@@ -83,8 +83,7 @@ function concurrency.with_coroutine_mutex(opts, fn)
     local err
     semaphore, err = ngx_semaphore.new()
     if err then
-      kong.log.crit("failed to create ", opts.name, " lock: ", err)
-      return nil, "critical error"
+      return nil, "failed to create " .. opts.name .. " lock: " .. err
```
hishamhm: Just to clarify, the original rationale on this one was that failure to create a semaphore is a pretty low-level failure situation, so the `"critical error"` result was meant as something that client code could check for (akin to `"timeout"` string results on other Lua APIs). As you said, current client code does not use that currently so it's useless. I'm thinking now that this library needs pcall-style returns that distinguish between errors produced by it (timeout, no semaphore) and errors produced by the callback. Something to fix in the future — in the meantime, this change is good.

thibaultcha: I agree, and had the same understanding (`"critical error"` being similar to a `"timeout"`). Yes, am improved API would be better in the longer term was my thinking as well.

hishamhm:
All of these are low-risk changes, I'm comfortable with merging them for 1.2.0 final.

thibaultcha:
Higher risk changes will be part of another PR :)

PR #4698, opened 2019-06-07
feat(db) support unique constraint on foreign fields

bungle:
### Summary

Adds support for unique foreign key fields in `schema` like:
```lua
{ service = { type = "foreign", reference = "services", unique = true, }, },
```

Also generates all the normal `unique` `DAO` functions for `unique` foreign keys:
- `select_by_[service]`
- `update_by_[service]`
- `upsert_by_[service]`
- `delete_by_[service]` 

Where `[service]` is the foreign field with `unique=true`.

This PR also adds new error `invalid foreign key` and changes the code to use it where it should.

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -439,7 +439,8 @@ local function generate_foreign_key_methods(schema)
   local methods = {}
 
   for name, field in schema:each_field() do
-    if field.type == "foreign" then
+    local field_is_foreign = field.type == "foreign"
+    if field_is_foreign then
       validate_foreign_key_is_single_primary_key(field)
 
       local page_method_name = "page_for_" .. name
@@ -527,8 +528,7 @@ local function generate_foreign_key_methods(schema)
           validate_options_type(options)
         end
 
-        local ok, errors = field.schema:validate_primary_key(foreign_key)
-
+        local ok, errors = schema:validate_field(field, foreign_key)
         if not ok then
           local err_t = self.errors:invalid_primary_key(errors)
           return iteration.failed(tostring(err_t), err_t)
@@ -550,8 +550,9 @@ local function generate_foreign_key_methods(schema)
 
         return iteration.by_row(self, pager, size, options)
       end
+    end
 
-    elseif field.unique or schema.endpoint_key == name then
+    if field.unique or schema.endpoint_key == name then
       methods["select_by_" .. name] = function(self, unique_value, options)
         validate_unique_type(unique_value, name, field)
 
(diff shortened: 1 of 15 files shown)


bungle:
Changed it to `wip` as I found out that `*_by_*` doesn't support `table` values such as `foreign keys` currently. The problem actually exists in `strategies`.

bungle:
I removed `wip` label as:
https://github.com/Kong/kong/blob/master/kong/db/dao/init.lua#L431

That line has been there for long, so we don't actually support any foreign fields with composite key.

hishamhm:
Let's double-check that `select_by_[service]` and friends work for DB-less

bungle:
@hishamhm dbless unique foreign now implemented, and tests added.

kong/db/dao/init.lua
```diff
@@ -458,14 +459,13 @@ local function generate_foreign_key_methods(schema)
           validate_options_type(options)
         end
 
-        local ok, errors = self.schema:validate_field(field, foreign_key)
+        local ok, err = self.schema:validate_field(field, foreign_key)
```
hishamhm: Looks like you changed a lot of `errors` variables to `err` in this module: these were named `errors` on purpose, to match their naming in the `kong.db.schema.init` module, where `errors` is used to refer for the errors table with the same type produced by that module, and differentiate them from `err`, which is used for a single error (of type string), and from `err_t` which is the error table (of a different type) produced by the `kong.db.errors` module.

There are three error things of different types going on in `kong.db.dao.init` (two of which are tables with different structures). The names were kept different to make it easier to keep track which is which. (Yeah, it sucks we don't have types to do this work properly, so we try to fake it with naming conventions ;) )

I personally would prefer these naming changes reverted, but I don't consider this a blocker for merging if you prefer to keep this change (the above was more of an explanation as to why the names were like this in the first place).

kong/db/declarative/init.lua
```diff
@@ -356,7 +363,13 @@ function declarative.load_into_cache(entities, hash, shadow_page)
 
       for _, unique in ipairs(uniques) do
         if item[unique] then
-          local cache_key = entity_name .. "|" .. unique .. ":" .. item[unique]
+          local unique_key = item[unique]
+          if type(unique_key) == "table" then
+            local _
+            _, unique_key = next(unique_key)
```
hishamhm: Wondering if we should add a comment here like
```
-- this assumes that foreign keys are not composite
```
but I guess various other places in the DAO make that assumption currently already.

spec/02-integration/04-admin_api/15-off_spec.lua
```diff
@@ -699,3 +699,135 @@ describe("Admin API #off", function()
...
+      }
+    })
+
+    local _, db = helpers.get_db_utils("off", {}, {
+      "unique-foreign"
+    })
+
+    local i = 1
+    while true do
+      local n, v = debug.getupvalue(db.unique_references.strategy.select_by_field, i)
```
hishamhm: Wow, you went deep to check this behavior! :grinning:

kong/db/dao/init.lua
bungle: Ok, made those changes, also tried to find other places to make it consistent.

kong/db/declarative/init.lua
bungle: Added that comment.

PR #4713, opened 2019-06-11
feat(api) add support for one more level of nesting to entity endpoints

### LINKED ISSUE #4329: cant not query endpoint for /oauth2
kong-ee can using query param like `/oauth2?client_id={client_id}`, but in kong-1.0.3, it can't effect.

image

bungle:
### Summary

1. Add support for one more level of nesting to entity endpoints

Auto-generates uris like (where you can send normal `GET`, `PUT`,
`PATCH`, and `DELETE`):

```
/services/:services/routes/:routes
```

E.g.
```
/services/example/routes/root
```

2. Add `admin_api_name` and `admin_api_nested_name` properties to `metaschema`

Enables admin api auto-generator to generate nicer urls with different
path components that what the `dao.name` is.

Also makes it possible for many authentication plugins to get rid of
their custom `api.lua`s.

3. Add admin api support for `admin_api_name` and `admin_api_nested_name`

Makes admin api auto-generator to leverage the newly added schema
properties `admin_api_name` and `admin_api_nested_name`.

4. Refactor plugins to use `admin_api_name` and `admin_api_nested_name`

Allows us to get rid of a bunch of custom `api.lua`s as well.

### Fixes

#4329


kong/db/schema/metaschema.lua
```diff
@@ -393,6 +393,18 @@ local MetaSchema = Schema.new({
         default = true,
       },
     },
+    {
+      path = {
```
hishamhm: I suggest renaming `path` and `nested_path`, because it's not clear those are related to Admin API endpoints, plus these are not paths, but path segments. My suggestion is names using `endpoint_*`, to align them with the existing `endpoint_key`.

* `path` → `endpoint_collection_name`
* `nested_path` →`endpoint_entity_name`

It's verbose, I know, but we or the users won't writing those things all the time (and the alternative of using `admin_api_` instead of `endpoint_` was even longer :) ).

kong/api/endpoints.lua
```diff
@@ -321,8 +323,14 @@ local function get_collection_endpoint(schema, foreign_schema, foreign_field_nam
...
 
     local foreign_key = self.params[foreign_schema.name]
-    local next_page = offset and fmt("/%s/%s/%s?offset=%s", foreign_schema.name,
-                                     foreign_key, schema.name, escape_uri(offset)) or null
+    local next_page = offset and fmt("/%s/%s/%s?offset=%s",
+                                     foreign_schema.path or
+                                     foreign_schema.name,
+                                     foreign_key,
+                                     schema.nested_path or
+                                     schema.path or
```
hishamhm: I like that `name` is used as a fallback for the `foreign_schema` entry :+1: , but here I think `path` should not be used as a fallback to the `nested_path` entry (the reason becomes clearer when we consider the names I suggested below — a collection name should not be used as a fallback for an entity name).

kong/db/schema/metaschema.lua
hishamhm: A follow-up to this based on our out-of-band conversation: looks like the `nested_path` is currently used for both entities (e.g. `GET /consumers/:consumers/key-auth/:keyauth_credentials`) and collections (e.g.  GET `/consumers/:consumers/key-auth`). My other two concerns remain (it's not clear those are related to Admin API endpoints, plus these are not paths, but path segments).

In these case, I sugest these uncool but technically correct names:

* `path` → `admin_api_name` (as in "the name this entity appears as in the Admin API") 
* `nested_path` → `admin_api_nested_name` (as in this entity appears as when nested in the Admin API")

(I avoided `endpoint_name` to prevent confusion with `endpoint_key` — but note that all `*_key` schema attributes refer to field names).

With this suggested rename, I think this PR can be merged.

bungle: They are not `path segments`path segment is `/<path_segment>/<second-path-segment>`, `path` and `nested_path` can be `something/:lapis_param/anything`.

Nested path is not used for both entities and collections (it always fits to position of `collection`:
```
/consumers/:consumers/key-auth/:keyauth_credentials
```
that can be translated to:
```
/collection/:entity/colection/:entity
```
that can be translated to:
```
/<path_or_name>/:entity/<nested_path_or_path_or_name>/:entity
```

Sure. after collection there can be `entity` endpoint (and this forth level adds one more entity level).

`nested` can also be seen `referenced`, e.g. `consumers` are referenced from `keyauth_credentials`.

I will do the rename, as the suggested names are ok with me.

bungle: The branch is now rebased (on latest `next`), and `path` is renamed to `admin_api_name` and `nested_path` is renamed to `admin_api_nested_name`.

Issue #4601, opened 2019-05-12
Kong freezes with database transaction error


ox160d05d:
### Summary

Kong periodically freezes (about once a week) without any visible reason. HTTP requests to kong gateway returns "503 Temporary unavailable". I see the following message in logs:
`
...
kong-gateway-develop_kong.1.yw8dfsgbvtz3@preprod-3    | 2019/04/30 04:24:16 [warn] 38#0: *854908 [lua] connector.lua:288: unable to clean expired rows from postgres database (ERROR: current transaction is aborted, commands ignored until end of transaction block), context: ngx.timer`

### Steps To Reproduce
Not clear, the issue appears periodically. 

### Additional Details & Logs

- Kong version (`1.1.2`)
- Kong error logs (empty)
- Kong configuration:

[...]

PR #4716, opened 2019-06-11
perf(db) delete expired rows without transaction with postgres

bungle:
### Summary

The reason is that transaction may acquire unneccessary locks, and is
also non-resilient to errors, e.g. if one table is missing `ttl` column
no cleanups will be done to other tables either.

This changes the code so that cleanups are ran separately on each table.

It might have something to do with #4601.

### DIFF
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -318,9 +318,11 @@ function _mt:init_worker(strategies)
 
     local sorted_strategies = graph:sort()
     local ttl_escaped = self:escape_identifier("ttl")
-    local cleanup_statement = {}
-    for i, table_name in ipairs(sorted_strategies) do
-      cleanup_statement[i] = concat {
+    local cleanup_statements = {}
+    local cleanup_statements_count = #sorted_strategies
+    for i = 1, cleanup_statements_count do
+      local table_name = sorted_strategies[i]
+      cleanup_statements[i] = concat {
         "  DELETE FROM ",
         self:escape_identifier(table_name),
         " WHERE ",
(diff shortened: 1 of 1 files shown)


kong/db/strategies/postgres/connector.lua
```diff
@@ -329,23 +329,22 @@ function _mt:init_worker(strategies)
...
-      if not ok then
-        if err then
-          log(WARN, "unable to clean expired rows from postgres database (", err, ")")
-        else
-          log(WARN, "unable to clean expired rows from postgres database")
+      for i, statement in ipairs(cleanup_statements) do
+        local ok, err = self:query(statement)
+        if not ok then
+          if err then
+            log(WARN, 'unable to clean expired rows from table "',
```
thibaultcha: style: can we stick to double quotes? We use single quotes in the text of each log line we write.

kong/db/strategies/postgres/connector.lua
```diff
@@ -329,23 +329,22 @@ function _mt:init_worker(strategies)
...
         return
       end
 
-      local ok, err = self:query(cleanup_statement)
-      if not ok then
-        if err then
-          log(WARN, "unable to clean expired rows from postgres database (", err, ")")
-        else
-          log(WARN, "unable to clean expired rows from postgres database")
+      for i, statement in ipairs(cleanup_statements) do
```
thibaultcha: I definitely agree with getting rid of the transaction for this operation, but I think that a unique query (or at least, a single round trip from Kong to PostgreSQL) would be better than sending many smaller queries, especially considering the lack of control over reused connections and additional network roundtrip this approach would cause regardless. Any reason for the individual queries?

bungle: I thought about it, but reporting the table name was easier with multiple. I can look other ways.

bungle: I tried different approaches, unfortunately, e.g. this:
```psql
DELETE FROM "oauth2_authorization_codes" WHERE "ttl" < CURRENT_TIMESTAMP AT TIME ZONE 'UTC';
DELETE FROM "oauth2_tokens" WHERE "ttl" < CURRENT_TIMESTAMP AT TIME ZONE 'UTC';
DELETE FROM "sessions" WHERE "ttl" < CURRENT_TIMESTAMP AT TIME ZONE 'UTC';
```

Would require `BEGIN EXCEPTION` handling on each statement which can be more expensive than it is beneficial. Also makes it hard to know and report what query errored. If for example the second will error, the third one will not be executed without proper exception handling.

So I think the current n-roundtrips is the best solution for this, unless someone has something better to offer?

thibaultcha: I wonder if there is anything to be done at the driver level for such use-cases (e.g. see https://github.com/leafo/pgmoon#result-err-partial-num_queries--postgresqueryquery_string and similar, although they don't do quite we want here as far as I can tell).

Nonetheless, we can keep the current n-roundtrips implementation which is already an improvement :+1:

thibaultcha: We could also run this query only when the schema has `ttl = true`, since we already assume that the TTL postgres column be present in the table when that is the case. This way we significantly reduce our margin for error and the logs noisiness. Actually, if we don't already do this, can we update this patch to do it?

bungle: I tried that `partial` pgmoon thing, but problem is that it stops on first error. Then the later statements don't get executed. We could have some logic to try `single roundtrip` if not working (obviously should always work), revert to `n-queries`. The code already checks the `ttl=true` on schemas.

thibaultcha: Alright, the current approach sounds good enough then given the constraints of other approaches :+1:

bungle:
@thibaultcha I made a slight change. Now it tries to execute everything in single roundtrip, if that fails it continues from the failed statement (to get error properly logged) and then continues on those that were after the errored one.

PR #4717, opened 2019-06-11
fix(dao) convert null/nil to empty string when generating cache_key from table

nateslo:
### Summary

This PR corrects an issue with cache_key generation when a schema defines a composite cache_key with optional fields that are not type `foreign`. The issues occurs when calling `DAO:cache_key` with `key` parameter of type `table` containing properties with value `nil` or `null`:

Example Schema
```
local schema = {
  name = "Foo",
  primary_key = { "a" },
  cache_key = { "b", "u" },
  fields = {
    { a = { type = "number" }, },
    { b = { type = "string" }, }, -- optional cache_key component
    { u = { type = "string" }, }, -- optional cache_key component
  },
}
```

nil
```
dao:cache_key({a = 1, b = "foo", u = ngx.null})
 -- 'Foo:foo:userdata: NULL:::'
```

null
```
dao:cache_key({a = 1, b = "foo", u = nil})
 -- 'Foo:foo:nil:::'
```

This PR converts nil/null cache_key components to an empty string, resulting in a cache_key output: 

nil/null
```
'Foo:foo::::'
```

### Full changelog

* convert null/nil to empty string when generating cache_key from table
* add regression test for nil/null in composite cache_key

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -1103,13 +1103,11 @@ function DAO:cache_key(key, arg2, arg3, arg4, arg5)
   for _, name in ipairs(source) do
     local field = self.schema.fields[name]
     local value = key[name]
-    if field.type == "foreign" then
+    if value == null or value == nil then
+      value = ""
+    elseif field.type == "foreign" then
       -- FIXME extract foreign key, do not assume `id`
-      if value == null or value == nil then
-        value = ""
-      else
-        value = value.id
-      end
+      value = value.id
     end
     values[i] = tostring(value)
     i = i + 1
--- a/spec/01-unit/01-db/04-dao_spec.lua
+++ b/spec/01-unit/01-db/04-dao_spec.lua
@@ -37,6 +37,17 @@ local non_nullable_schema_definition = {
   }
 }
 
+local optional_cache_key_fields_schema = {
+  name = "Foo",
+  primary_key = { "a" },
+  cache_key = { "b", "u" },
+  fields = {
+    { a = { type = "number" }, },
+    { b = { type = "string" }, },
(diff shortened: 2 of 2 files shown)


thibaultcha:
@nateslo Nice one! One question: considering we currently (since 1.0) insert plugins with a `cache_key` column that we generate based on this method, have you conducted tests or any sort of analysis determining if there may exist, today, rows with such invalid cache keys in the wild? If so, we need to write a migration for such rows, or else they will not be able to be selected by the DAO anymore, given this PR changes the generated cache keys (for the better of course, we definitely identified this as a bug).

nateslo:
> @nateslo Nice one! One question: considering we currently (since 1.0) insert plugins with a `cache_key` column that we generate based on this method, have you conducted tests or any sort of analysis determining if there may exist, today, rows with such invalid cache keys in the wild? If so, we need to write a migration for such rows, or else they will not be able to be selected by the DAO anymore, given this PR changes the generated cache keys (for the better of course, we definitely identified this as a bug).

Yes! Investigating a migration to correct existing rows now.

nateslo:
After speaking with @thibaultcha, it seems that a migration is not needed. This PR should be good to go as is.

hishamhm:
@nateslo since this is a bugfix (and the fix does not bring a significant compatibility concern as the chances of people depending on this behavior are slim), I think we could rebase this to `master` so we could ship it in Kong 1.2.1. Apart from that, it's a +1 from me!

thibaultcha:
Agreed with @hishamhm, let's rebase this on master!

nateslo:
@hishamhm sounds good, will rebase on master shortly.

PR #4719, opened 2019-06-12
fix(proxy) do not clear upgrade header with websockets, fix #4718

### LINKED ISSUE #4718: Websocket Upgrade header missing after upgrade to 1.2.0
### Summary

When publishing a an application that uses web sockets (via deepstream.io 3.2.4) after upgrading kong 1.1.2 to 1.20 the Upgrade response header is not passed to the client meaning a websocket connection fails to be established.

### Steps To Reproduce

1. Deploy websocket application (that uses websocket upgrade protocol)  through kong 1.1.2
2. Test that it works and if necessary capture http request logs using browser developer tools
3. Follow the standard kong upgrade process to 1.2.0 with identical service and route configuration
4. Test the same after the upgrade and note websockets will not be connected.

### Additional Details & Logs

- Kong version 1.2.0

Network logs

bungle:
### Summary

Ensures that `Upgrade` header is not cleared with `websockets`.

### Issues resolved

Fix #4718

### DIFF
--- a/spec/02-integration/05-proxy/09-websockets_spec.lua
+++ b/spec/02-integration/05-proxy/09-websockets_spec.lua
@@ -38,6 +38,59 @@ for _, strategy in helpers.each_strategy() do
       return wc
     end
 
+    local function check_headers(host, port)
+      local is_kong = (host == helpers.get_proxy_ip(false) and
+                       port == helpers.get_proxy_port(false)) or
+                      (host == helpers.get_proxy_ip(true) and
+                       port == helpers.get_proxy_port(true))
+
+      local encode_base64 = ngx.encode_base64
+      local rand = math.random
+      local char = string.char
+      local sock = assert(ngx.socket.tcp())
+      local path = is_kong and "/up-ws" or "/ws"
+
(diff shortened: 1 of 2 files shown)


spec/02-integration/05-proxy/09-websockets_spec.lua
```diff
@@ -38,6 +38,57 @@ for _, strategy in helpers.each_strategy() do
...
+      local key = encode_base64(bytes)
+      local req = "GET /up-ws HTTP/1.1\r\nUpgrade: websocket\r\nHost: "
+        .. host .. ":" .. port
+        .. "\r\nSec-WebSocket-Key: " .. key
+        .. "\r\nSec-WebSocket-Version: 13"
+        .. "\r\nConnection: Upgrade\r\n\r\n"
+
+      assert(sock:send(req))
+
+      local header_reader = sock:receiveuntil("\r\n\r\n")
```
thibaultcha: Can we close the socket once we are done here?

bungle: Sure. Thanks.

PR #4720, opened 2019-06-12
fix(db) snis dao to set correct array metatable on non-empty arrays

bungle:
### Summary

Fixes `snis` dao to set correct metatable on non-empty arrays:
`cjson.empty_array_mt` to `cjson.array_mt`

### DIFF
--- a/kong/db/dao/snis.lua
+++ b/kong/db/dao/snis.lua
@@ -85,7 +85,7 @@ end
 
 -- Returns the name list for a given certificate
 function _SNIs:list_for_certificate(cert_pk, options)
-  local name_list = setmetatable({}, cjson.empty_array_mt)
+  local name_list = setmetatable({}, cjson.array_mt)
 
   for sni, err, err_t in self:each_for_certificate(cert_pk, 1000, options) do
     if err then
--- a/spec/02-integration/03-db/02-db_core_entities_spec.lua
+++ b/spec/02-integration/03-db/02-db_core_entities_spec.lua
@@ -3,6 +3,7 @@ local utils   = require "kong.tools.utils"
 local helpers = require "spec.helpers"
 local cjson   = require "cjson"
 
+
 local fmt      = string.format
 local unindent = helpers.unindent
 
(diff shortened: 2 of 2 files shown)


kong/db/dao/snis.lua
```diff
@@ -85,18 +85,26 @@ end
...
     table.insert(name_list, sni.name)
+    name_count = name_count + 1
   end
 
   table.sort(name_list)
 
+  if name_count == 0 then
+    setmetatable(name_list, cjson.empty_array_mt)
+  else
+    setmetatable(name_list, cjson.array_mt)
```
thibaultcha: You can forget about `cjson.empty_array_mt` and simply stick to setting `array_mt` unconditionally (this could just be a one line change).

bungle:
> Do we have a test case for this fix? It's an important one to have for these tricky edge-cases that have bit us repeatedly.

Test-case added.

spec/02-integration/03-db/02-db_core_entities_spec.lua
```diff
@@ -1951,5 +1954,27 @@ for _, strategy in helpers.each_strategy() do
...
+          local json = cjson.encode(snis)
+          assert.equal("[]", json)
+          assert.equal(cjson.array_mt, getmetatable(snis))
+
+          db.snis:update_list(certificate, {"example.org"})
+          snis = assert(db.snis:list_for_certificate(certificate))
+          json = cjson.encode(snis)
+          assert.equal('["example.org"]', json)
+          assert.equal(cjson.array_mt, getmetatable(snis))
+        end)
```
thibaultcha: This test passes with the old `cjson.empty_array_mt` metatable, so long as we remove the `assert.equal(cjson.array_mt, ...)` assertions. I wonder if this is a fix for a known issue (in which we don't JSON-encode the result properly) with an incomplete regression test, or a refactor to use `array_mt` (in which case, it should be bundled with #4724)?

bungle: @thibaultcha, the other functions reset the metatable. I will bundle it to #4724, and rebase that to master. I will close this.

PR #4724, opened 2019-06-13
refactor(db) replace deprecated cjson.empty_array_mt with cjson.array_mt

bungle:
### Summary

1. `dao:rows_to_entities` will now properly set `array_mt` on returned entities
2. schema's `make_array` will now set `array_mt` even on non-empty arrays
3. schema's `make_set` will now set `array_mt`, it still sets `Set_mt` for non-empty arrays which the `cjson.encode` does not support, but  it works as long as the value is not turned to e.g. hash afterwards.

This is continuation of #4720.

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -1018,7 +1018,7 @@ end
 function DAO:rows_to_entities(rows, options)
   local count = #rows
   if count == 0 then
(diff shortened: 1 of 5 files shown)


thibaultcha:
@bungle Could we rebase this on master?

PR #4730, opened 2019-06-18
refactor(pdk) use cjson.new and decode_array_with_array_mt(true)

bungle:
### Summary

Instead of switching back and forth with global `cjson`, we instantiate
a new module level `cjson` and set `decode_array_with_array_mt` to true.

Thanks @kidd and @Tieske for the pointers.

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -5,7 +5,7 @@
 -- @module kong.request
 
 
-local cjson = require "cjson.safe"
+local cjson = require "cjson.safe".new()
 local multipart = require "multipart"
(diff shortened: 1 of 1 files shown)


Tieske:
I have no idea how "expensive" it is to instantiate new cjson instances, but shouldn't the instance actually live in `kong.pdk.json` and that be reused from other places?

bungle:
@Tieske, later, if we decide so. The interface of `kong.json` is probably a bit different than `cjson`, so for now this is just a quick patch to remove that back-and-forth. It is still global, but scoped for module.

PR #4742, opened 2019-06-20
feat(scripts) introduce semi-automated merge-master-into-next script

hishamhm:
This is a development tool for performing merges from the `master` branch into `next` correctly. Up to this day, these merges had to be done manually and that process was error-prone (multiple people got it wrong in different occasions, myself included).

A key problem is that the Github green button offers three options but none of them does the right thing we need (you end up either with an incorrect rebase, or with a double merge commit), therefore the branch needs to be approved in the web UI and merged from the terminal.

This script makes this process more consistent. It requires `hub` (a CLI interface to the Github API).

As a demonstration of this script, the latest merge from `master` into `next` was done entirely using this script:

* I ran `scripts/merge-master-into-next create` and it created the branch locally
* Then `scripts/merge-master-into-next submit` and it automatically opened PR #4741 (including the "do not merge" label, etc.)
* @gszr approved it in the web UI
* Then I ran `scripts/merge-master-into-next merge`, the script merged the branch, closed the PR and deleted the repo in Github.

### DIFF
--- a/scripts/merge-master-into-next
+++ b/scripts/merge-master-into-next
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+set -e
+
+if [ "$1" = "-h" ] || [ "$1" = "--help" ] || ! [ "$1" ]
+then
+   echo "Merge Kong's 'master' branch into the 'next' branch using this script:"
+   echo ""
+   echo "Usage:"
+   echo " Step 1) create the branch"
+   echo "    $0 create"
+   echo ""
+   echo " Step 2) submit the PR"
+   echo "    $0 submit"
+   echo ""
+   echo " Step 3) get approval for the PR -- do NOT merge it on Github!"
+   echo "    $0 approve"
+   echo ""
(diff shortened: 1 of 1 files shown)


scripts/merge-master-into-next
```diff
@@ -0,0 +1,120 @@
...
+         fi
+      done
+      browser https://github.com/Kong/kong/pull/$prnum
+      ;;
+   merge)
+      git checkout "$branch"
+      git pull
+      git checkout next
+      git reset --hard "$branch"
+      git push
```
gszr: Sanity check: this `git push` operation will be rejected in case new commits have been merged to `next` after `$branch` was branched off of it, correct?

hishamhm: Yes, because that would require a force-push. This `git push` only works from the CLI because that SHA is approved via the web UI.

hishamhm:
> CONTRIBUTING

If we are to document this, I'd rather have a separate doc for maintainers. This info is useful only to the limited number of people with commit access.

PR #4749, opened 2019-06-25
rate-limiting plugin support at service level

wuguangkuo:
rate-limiting plugin support at service level

The original code can't support rate limiting at service level, it's not satisfied the plugin usage.

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -25,7 +25,9 @@ RateLimitingHandler.VERSION = "2.0.0"
 local function get_identifier(conf)
   local identifier
 
-  if conf.limit_by == "consumer" then
+  if conf.limit_by == "service" then
+    identifier = ""
+  elseif conf.limit_by == "consumer" then
     identifier = (kong.client.get_consumer() or
                   kong.client.get_credential() or
                   EMPTY).id
@@ -33,9 +35,11 @@ local function get_identifier(conf)
   elseif conf.limit_by == "credential" then
     identifier = (kong.client.get_credential() or
                   EMPTY).id
+  elseif conf.limit_by == "ip" then
+    identifier = kong.client.get_forwarded_ip()
   end
 
-  return identifier or kong.client.get_forwarded_ip()
+  return identifier
 end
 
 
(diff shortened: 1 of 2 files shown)


wuguangkuo:
@gszr

kong/plugins/rate-limiting/handler.lua
```diff
@@ -25,17 +25,21 @@ RateLimitingHandler.VERSION = "2.0.0"
...
 
   elseif conf.limit_by == "credential" then
     identifier = (kong.client.get_credential() or
                   EMPTY).id
+  elseif conf.limit_by == "ip" then
+    identifier = kong.client.get_forwarded_ip()
   end
 
-  return identifier or kong.client.get_forwarded_ip()
+  return identifier
```
Tieske: this removes the automatic fallback on IP, that's not ok, please revert this.

kong/plugins/rate-limiting/schema.lua
```diff
@@ -39,8 +39,8 @@ return {
           { year = { type = "number", gt = 0 }, },
           { limit_by = {
               type = "string",
-              default = "consumer",
-              one_of = { "consumer", "credential", "ip" },
+              default = "service",
```
Tieske: please do not change the default, that is a major breaking change in behaviour

kong/plugins/rate-limiting/handler.lua
```diff
@@ -25,17 +25,21 @@ RateLimitingHandler.VERSION = "2.0.0"
 local function get_identifier(conf)
   local identifier
 
-  if conf.limit_by == "consumer" then
+  if conf.limit_by == "service" then
+    identifier = ""
```
Tieske: shouldn't this be the service id? if using a constant then there will be 1 global rate limit no?

wuguangkuo: the get_local_key method in init.lua had concat service id and route id(if existed)

`local get_local_key = function(conf, identifier, period, period_date)
  local service_id, route_id = get_service_and_route_ids(conf)

  return fmt("ratelimit:%s:%s:%s:%s:%s", route_id, service_id, identifier,
             period_date, period)
end`

Tieske: yes, that is correct. But please add a test that validates the behaviour anyway.

PR #4752, opened 2019-06-25
fix(routes/upstreams) healthcheck support by IP:port

locao:
- Added routes to set healthy/unhealthy status for each record's IP entry.
- Report health status for each record's IP entry individually.

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -183,7 +183,7 @@ local function populate_healthchecker(hc, balancer)
         -- Get existing health status which may have been initialized
         -- with data from another worker, and apply to the new balancer.
         local tgt_status = hc:get_target_status(ipaddr, port)
-        balancer:setPeerStatus(tgt_status, ipaddr, port, hostname)
+        balancer:setAddressStatus(tgt_status, ipaddr, port, hostname)
 
       else
         log(ERR, "[healthchecks] failed adding target: ", err)
@@ -206,9 +206,9 @@ do
 
     ------------------------------------------------------------------------------
     -- Callback function that informs the healthchecker when targets are added
-    -- or removed to a balancer.
+    -- or removed to a balancer and when targets health status change.
     -- @param balancer the ring balancer object that triggers this callback.
-    -- @param action "added" or "removed"
+    -- @param action "added", "removed", or "health"
     -- @param address balancer address object
     -- @param ip string
     -- @param port number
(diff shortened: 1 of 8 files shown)


spec/01-unit/09-balancer_spec.lua
```diff
@@ -451,17 +451,18 @@ describe("Balancer", function()
...
+                    health_info[t.host .. ":" .. t.port].addresses[1].health)
       end
     end)
 
     it("requires hostname if that was used in the Target", function()
-      local ok, err = balancer.post_health(upstream_ph, "127.0.0.1", 1111, true)
+      local ok, err = balancer.post_health(upstream_ph, "127.0.0.1", nil, 1111, true)
       assert.falsy(ok)
-      assert.match(err, "target not found for 127.0.0.1:1111")
+      assert.match(err, "No host found by: '127.0.0.1:1111'")
```
Tieske: it should probably remain "target" as that is the Kong vocabulary.

Tieske:
The version bump of the library, and taking care of the breaking changes should probably be a separate commit. Breaking changes are here: https://github.com/Kong/lua-resty-dns-client#400-26-jun-2019

The extra callback (health status) should probably do nothing but log a warn level message that the balancer changed state. The method name change can probably be a simple search and replace.

And, it would be nice to also expose the new `healthThreshold` property in the `upstream` object (since this PR is already targetted at `next`)

spec/01-unit/09-balancer_spec.lua
locao: This message comes straight from the lua-resty-dns-client library, it seems better to leave as is.

kong/runloop/balancer.lua
```diff
@@ -224,9 +224,13 @@ do
       elseif action == "removed" then
         local ok, err = healthchecker:remove_target(ip, port)
         if not ok then
-          log(ERR, "[healthchecks] failed adding a target: ", err)
+          log(ERR, "[healthchecks] failed removing a target: ", err)
         end
 
+      elseif action == "health" then
+        log(WARN, "[healthchecks] healthy status changed for target: ",
+                  hostname, " - ", ip, ":", port)
```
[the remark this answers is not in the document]
locao: Fixed that, thanks!

kong/runloop/balancer.lua
```diff
@@ -224,8 +224,18 @@ do
       elseif action == "removed" then
         local ok, err = healthchecker:remove_target(ip, port)
         if not ok then
-          log(ERR, "[healthchecks] failed adding a target: ", err)
+          log(ERR, "[healthchecks] failed removing a target: ", err)
+        end
+
+      elseif action == "health" then
+        local health_status
```
Tieske: can we change the variable name to `balancer_status` that's more appropriate imo

kong/runloop/balancer.lua
```diff
@@ -224,8 +224,18 @@ do
...
+        if address then
+          health_status = "HEALTHY"
+        else
+          health_status = "UNHEALTHY"
         end
+        log(WARN, "[healthchecks] healthchecker for ", healthchecker.name,
+            " reported health status changed to ", health_status)
```
Tieske: is `healthchecker.name` the same as the balancer name? I'd make the message more expclicit that it's the balancer that switches health state:

```lua
log(WARN, "[healthchecks] balancer ", healthchecker.name,
            " changed health status to ", health_status)
```

locao:
The content of this PR is part of #4792.

PR #4759, opened 2019-06-26
feat(reports) add prd field to detect product type

hbagdi:
## Summary

Adds  a `prd=kong` key-value to the reports to detect the type of product.


kong/reports.lua
```diff
@@ -41,6 +41,8 @@ do
     _buffer[#_buffer + 1] = k .. "=" .. v
   end
 
+  _buffer[#_buffer + 1] = "prd=kong"
```
gszr: If we intend to customize the value in other places (e.g., Enterprise), could we define the value in `kong.constants` (or obtain from `kong.meta`)?

PR #4761, opened 2019-06-28
add feature to jwt plugin:allow set headers from claim

comdeng:
### Summary

Add feature to `jwt` plugin: allow set header from claims.

### Full changelog

* [Implement] jwt plugin: allow set header from claims
   if you configure `headers_to_set`(set) , headers with prefix "X-Jwt-Claim-" will be set to the response. For example, if 'sub' and 'app' are configured, and both of them are found in jwt's claims, two headers will be set to the response: `X-Jwt-Claim-Sub` and `X-Jwt-Claim-App`.

### DIFF
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -9,12 +9,16 @@ local ipairs = ipairs
 local tostring = tostring
 local re_gmatch = ngx.re.gmatch
 
+local consts = {
+  JWT_CLAIM_HEADER_PREFIX = "X-Jwt-Claim"
+}
+
 
 local JwtHandler = {}
 
 
 JwtHandler.PRIORITY = 1005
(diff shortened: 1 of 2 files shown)


hishamhm:
@comdeng we're wondering whether it would make sense to allow the user to set the header name instead of using a hardcoded prefix. Making the configuration like this would work:

```
"headers_to_set": [
  { "header": "x-jwt-claim-foo", "claim": "foo" }
]
```

How does that sound?

amihalj:
what is the state of this pull request? will this be available any time soon? This seems like a good feature since we can extract more info from token (Eg if we use cognito token) than standard functionality offers.

thanks

hbagdi:
How does this relate to https://github.com/Kong/kong/pull/3491?
@comdeng are you still interested in pursuing this feature?

PR #4768, opened 2019-07-02
fix(utils) recursively url encode args

nateslo:
# Summary

This PR modifies `utils.encode_args` to recursively encode nested args. Previously, only a single level of nesting was supported. 

### Full changelog

* Adds `recursive_encode_args` function, called from `utils.encode_args` when it encounters a table value
* Adds tests

### DIFF
--- a/kong/api/api_helpers.lua
+++ b/kong/api/api_helpers.lua
@@ -1,10 +1,8 @@
-local pl_string = require "pl.stringx"
 local utils = require "kong.tools.utils"
 local cjson = require "cjson"
 
 local type = type
 local pairs = pairs
-local remove = table.remove
 
 local _M = {}
 
@@ -15,6 +13,45 @@ local function parse_value(v)
   return type(v) == "table" and v.content or v -- Handle multipart
 end
 
+local NO_ARRAY_INDEX_MARK = {}
+
+-- given a string like "x[1].y", return an array of indices like {"x", 1, "y"}
+-- the path parameter is an output-only param. the keys are added to it in order
+local function key_to_path(key, path)
+  -- try to match an array access like x[1].
+  -- the left side of the [] is mandatory
+  -- the array index can be omitted (the key will look like x[]).
+  -- if that's the case we mark the path entry with a special key
+  local left, array_index = key:match("^(.+)%[(%d*)]$")
+  if left then
+    key_to_path(left, path)
+    path[#path + 1] = tonumber(array_index) or NO_ARRAY_INDEX_MARK
+    return path
+  end
+
+  -- if no match, try a hash access like x.y (both x and y are mandatory)
+  -- the left side of the dot is called left and the other side is right
+  local left, right = key:match("^(.+)%.(.+)$")
+  if left then
+    key_to_path(left, path)
+    key_to_path(right, path)
+    return path
+  end
+
+  -- if no match found, append the whole key to the path as a single string
+  path[#path + 1] = key
+  return path
+end
+
+-- when NO_ARRAY_INDEX is encountered, replace it with the length of the node being parsed
+local function transform_no_array_index_mark(path_entry, node)
+  if path_entry == NO_ARRAY_INDEX_MARK then
+    return #node + 1
+  end
+  return path_entry
+end
+
+
 -- Put nested keys in objects:
 -- Normalize dotted keys in objects.
 -- Example: {["key.value.sub"]=1234} becomes {key = {value = {sub=1234}}
@@ -22,25 +59,14 @@ end
 -- @return `normalized_object`
 function _M.normalize_nested_params(obj)
   local new_obj = {}
-
-  local function attach_dotted_key(keys, attach_to, value)
-    local current_key = keys[1]
-
-    if #keys > 1 then
-      if not attach_to[current_key] then
-        attach_to[current_key] = {}
-      end
-      remove(keys, 1)
-      attach_dotted_key(keys, attach_to[current_key], value)
-    else
-      attach_to[current_key] = value
-    end
-  end
+  local is_array
 
   for k, v in pairs(obj) do
+    is_array = false
     if type(v) == "table" then
       -- normalize arrays since Lapis parses ?key[1]=foo as {["1"]="foo"} instead of {"foo"}
       if utils.is_array(v) then
+        is_array = true
         local arr = {}
         for _, arr_v in pairs(v) do arr[#arr+1] = arr_v end
         v = arr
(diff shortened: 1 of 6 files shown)


kong/tools/utils.lua
```diff
@@ -270,6 +270,28 @@ do
     end
   end
 
+
+  -- Recursively URL escape and format key and value
+  -- Handles nested arrays and tables
+  local function recursive_encode_args(parent_key, value, raw, no_array_indexes, query)
+    for sub_key, sub_value in pairs(value) do
+      if no_array_indexes then
+        sub_key = parent_key
```
thibaultcha: Better not to override the loop-provided variable within the loop's body; using a local is fine and preferred.

spec/02-integration/01-helpers/01-helpers_spec.lua
```diff
@@ -61,6 +61,23 @@ for _, strategy in helpers.each_strategy() do
         assert.equals("bob",    params["names[2]"])
         assert.equals("casius", params["names[3]"])
       end)
+
+      it("encodes nested tables and arrays in a way Lapis-compatible way when using form-urlencoded content-type", function()
```
thibaultcha: Note that we decode our Admin API request bodies [ourselves](https://github.com/Kong/kong/blob/master/kong/api/init.lua#L38) (and https://github.com/Kong/kong/blob/master/kong/api/api_helpers.lua#L23), and do not rely on Lapis for it. We should write unit tests asserting that the those decoders can indeed parse content encoded with nested tables. 

Also, note the typo (repeated "way") in this test's title.

spec/02-integration/01-helpers/01-helpers_spec.lua
```diff
@@ -61,6 +61,23 @@ for _, strategy in helpers.each_strategy() do
...
+        local r = proxy_client:get("/", {
+          headers = {
+            ["Content-type"] = "application/x-www-form-urlencoded",
+            host             = "mock_upstream",
+          },
+          body    = {
```
thibaultcha: Let's align those properly, `body = {` is fine

spec/01-unit/05-utils_spec.lua
```diff
@@ -229,11 +229,12 @@ describe("Utils", function()
         local str = utils.encode_args {
           array = {"hello, world"},
           hash = { answer = 42 },
+          nested = { array = { "one", "two" } },
           hello = "world",
           falsy = false,
           ["multiple values"] = true
         }
-        assert.equal("array%5b1%5d=hello%2c%20world&falsy=false&hash%2eanswer=42&hello=world&multiple%20values=true", str)
+        assert.equal("array%5b1%5d=hello%2c%20world&falsy=false&hash%2eanswer=42&hello=world&multiple%20values=true&nested%2earray%5b1%5d=one&nested%2earray%5b2%5d=two", str)
```
thibaultcha: We should also add tests asserting that the `raw` argument is still respected for nested table values.

kong/tools/utils.lua
```diff
@@ -270,6 +270,28 @@ do
     end
   end
 
+
+  -- Recursively URL escape and format key and value
+  -- Handles nested arrays and tables
+  local function recursive_encode_args(parent_key, value, raw, no_array_indexes, query)
+    for sub_key, sub_value in pairs(value) do
```
thibaultcha: I think that we should duplicate the keys sorting logic at the beginning of `encode_args()` in this function as well. We want the result of the encoding to be identical in all platforms, which means that we cannot rely on `pairs`.

spec/01-unit/05-utils_spec.lua
thibaultcha: Another _unit_ test case that seems to be missing is nested hash table cases (non-integer-indexed tables) and even maybe mixed tables.

thibaultcha: Other missing test case: nested tables with `no_array_indexes`

kong/tools/utils.lua
```diff
@@ -297,24 +319,7 @@ do
...
-
-            else
-              query[#query+1] = encode_args_value(("%s.%s")
-                                  :format(key, tostring(sub_key)), sub_value,
-                                          raw)
-
-            end
-          end
-        end
+        recursive_encode_args(key, value, raw, no_array_indexes, query)
```
thibaultcha: At the top of `encode_args()` is a comment documenting its behavior and differences with `ngx.encode_args()`. We should add a mention for such nested array parameters now.

thibaultcha:
Also, the scope of this commit/PR should definitely be `feat` and not `fix`

spec/02-integration/01-helpers/01-helpers_spec.lua
kikito: It turns out our api_helper was not able to handle "full" nested structures, only a subset. I have added tests for this and the changes needed in a separate commit. I also removed the "way way" typo.

spec/01-unit/05-utils_spec.lua
kikito: I have added a bunch of tests - `array`, `hash`, `array_array`, `hash_hash`, `array_hash`, `hash_array` and `hybrid`.

kong/api/api_helpers.lua
```diff
@@ -15,6 +13,57 @@ local function parse_value(v)
   return type(v) == "table" and v.content or v -- Handle multipart
 end
 
+
+-- given a string like "x[1].y", return an array of indices like {"x", 1, "y"}
+local function key_to_path(key)
```
hishamhm: I really liked the inductive approach implementing this!

kong/tools/utils.lua
```diff
@@ -270,13 +270,56 @@ do
...
+      sub_keys[#sub_keys + 1] = sk
+    end
+    sort(sub_keys, compare_keys)
+
+    local sub_value, next_sub_key
+    for _, sub_key in ipairs(sub_keys) do
+      sub_value = value[sub_key]
+
+      if no_array_indexes then
+        next_sub_key = parent_key
```
thibaultcha: Are we missing tests for `no_array_indexes` or did I miss them?

kikito: You were completely right on this one, thanks.

There were no tests for this and it turns out that the implementation we had did not do it correctly (it should return `array[]` instead of `array[1]`, but it was returning `array` instead).

I added tests for `no_array_indexes`. I also made the admin api params parser able to cope with parameters with missing array indexes too, with tests.

thibaultcha: @kikito Great! It seems like a key-auth integration test is failing; likely due to a recent change though (it fails with form-urlencoded payloads)

kikito: Yep, noticed yesterday but could not fix because I was busy preparing presentations. I have fixed key-auth on a separate commit. It is green now.

kong/plugins/key-auth/handler.lua
```diff
@@ -112,6 +112,10 @@ local function do_authentication(conf)
     -- search the body, if we asked to
     if not v and conf.key_in_body then
       v = body[name]
+      -- duplicated keys in form-urlencoded requests
+      if not v and (body[name .. "[]"] or body[name .. "[1]"]) then
+        return nil, { status = 401, message = "Duplicate API key found" }
+      end
```
thibaultcha: I don't think that we should have such a branch in this plugin. We parse our `form-urlencoded` payloads with [ngx.req.get_post_args()](https://github.com/Kong/kong/blob/master/kong/pdk/request.lua#L625), and it doesn't support the notation we use to encode arrays; rightfully-so since it isn't standard anyway, and only used as a practical mean to mimic `application/json` payloads in the Admin API.

In my opinion, we should either:

1. Not add `[]` or `[1]` on leaf array nodes, which would preserve the previous encoding behavior and cause `ngx.req.get_post_args()` to return `{ "value1", "value2" }`, thus avoiding any changes in this plugin or in our test suite.
2. Disable the "duplicated API keys" test case for `form-urlencoded` payloads. Only run the test with `application/json` (it already is disabled for `multipart/form-data` content types.

kikito: I have done something a bit different: I have left `key-auth` code code untouched, and instead added more tests which exactly document how things work now, without doing changes to key-auth:

* In json, if they key is duplicated, the error is "Duplicated key"
* In form-url-encoded:
  * `apikey=foo&apikey=foo` throws "Duplicated key"
  * `apikey[1]=foo&apikey[2]=foo` throws "Key not found"
  * `apikey[]=foo&apikey[]=foo` throws "Key not found"

I have not used automatic encoding and instead used direct strings for all the `form-url-encoded` tests. I think that makes things easier to follow.

hishamhm: @kikito Got it. I reviewed the new `tests()` commit and it LGTM. I think this PR is good to go.

thibaultcha: @kikito Works too. Albeit our test client’s form encoder isn’t able to encode arrays that ngx_lua can decode anymore, which may cause us issues down the road (lack of test coverage when maintainers skip those tests for form-encoded payloads or even future changes to plugins similar to the one you originally made due to a lack of understanding around this new limitation).

Note that my above comment also implied that no changes were to be made to the plugin itself.

Issue #4769, opened 2019-07-03
Kong and Cassandra, observed behaviors/paging issues


jeremyjpj0916:
### Summary

Continuation cleaned up of https://github.com/Kong/kong/issues/4194 , it was getting way too long and had old discussions on older Kong version vs modern issues trying to triage.

With keepalive problem we had had pretty much figured out now(https://github.com/Kong/openresty-patches/pull/40), this is the only hurdle we see left with running Kong, so lets crack the last mystery so I can crown this Gorilla 🐒.

Summary of issues seen:

1. Once route/service/plugins maps get big enough(couple thousand records each), the paging through C* can produce NPEs(https://github.com/Kong/kong/issues/4194#issuecomment-497572751). Seemingly occurs when Kong pages across multiple Cassandra nodes in the cluster for us.

Oddly something that seems to fix it is to sticky persist to a C* node during the rebuild. I accomplished this by modifying the ```concurrency.lua``` file as described here:
https://github.com/Kong/kong/issues/4194#issuecomment-498001854

This is a bandaid approach though and I don't wanna run this if I can avoid it, its right and proper for Kong to be able to jump around when doing all this paging querying and jumping node to node should not be problem. 

I think we can agree C* throwing a NPE is worthy of a JIRA task to them, I have an ASF Slack convo I had with some of them: 
https://the-asf.slack.com/archives/CJZLTM05A/p1559321422028200
And bits of the convo I pulled and posted here too: https://github.com/Kong/kong/issues/4194#issuecomment-497858824

I will try to raise that, but I don't know if I can give a good repro for them so they might be flying blind. But also per discussion could be a Kong DAO or Driver issue as well to nail down.

2. Even with the above in place, there are portions I can't do sticky persist logic in design, think like the admin API specifically. So I observe when standing up service + route + plugins on rout, then tearing down route(which also deletes the plugins automagically). Then the call to delete the service will fail as described here:
https://github.com/Kong/kong/issues/4194#issuecomment-498002915
https://github.com/Kong/kong/issues/4194#issuecomment-498117978
https://github.com/Kong/kong/issues/4194#issuecomment-498533977   

Seems like specifically the cascading hot path on the service delete when its trying page_for_service(which does not seem to get invoked often in the path if I wait a second to delete the service after deleting a given route).

Diving deeper down the rabbit hole at this stage I have taken @thibaultcha shout on where in the C* driver I can log the queries a bit and given this snippit a go in the ```init.lua```:

```
[...]

PR #4770, opened 2019-07-03
perf(db) cassandra cascade delete to use 1000 as page size

bungle:
### Summary

Changes Cassandra to use `1000` as a page size for cascading deletes instead of the default `100`. This change may have some impact on #4769.

### DIFF
--- a/kong/db/strategies/cassandra/init.lua
+++ b/kong/db/strategies/cassandra/init.lua
@@ -1461,7 +1461,7 @@ do
         local pager = function(size, offset)
           return strategy[method](strategy, primary_key, size, offset)
         end
-        for row, err in iteration.by_row(self, pager) do
+        for row, err in iteration.by_row(self, pager, 1000) do
           if err then
             return nil, self.errors:database_error("could not gather " ..
                                                    "associated entities " ..
@@ -1522,7 +1522,7 @@ function _mt:delete_by_field(field_name, field_value, options)
 
   local pk = self.schema:extract_pk_values(row)
 
-  return self:delete(pk)
+  return self:delete(pk, options)
 end
 
 
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -532,11 +532,11 @@ local function generate_foreign_key_methods(schema)
 
         local strategy = self.strategy
 
-        local pager = function(size, offset)
+        local pager = function(size, offset, options)
           return strategy[page_method_name](strategy, foreign_key, size, offset, options)
         end
 
-        return iteration.by_row(self, pager, size)
+        return iteration.by_row(self, pager, size, options)
       end
 
     elseif field.unique or schema.endpoint_key == name then


kong/db/strategies/cassandra/init.lua
```diff
@@ -1461,7 +1461,7 @@ do
         local pager = function(size, offset)
           return strategy[method](strategy, primary_key, size, offset)
         end
-        for row, err in iteration.by_row(self, pager) do
+        for row, err in iteration.by_row(self, pager, 1000) do
```
thibaultcha: Why not `options.page_size`? And having it being specified/defaulted by the DAO (not the strategy). This will allow to eventually make the page_size configurable (humanly or via heuristics) on a use-case basis.

thibaultcha: (Such an option would of course be a NOP for relevant PostgreSQL queries - non-SELECT DML queries, that is)

bungle: > There seems to be a lot of places in which we hard-coded the default/max page size to 1000

That is true. The default was picked up for admin api, and then after that a lot of places we have changed it for performance reasons. Originally it should have defaulted to `1000` and `admin api` should have changed it to `100`.

> Why not options.page_size?

The problem is that sometimes user calls, e.g.:
`dao:page` and that results additional queries. It is a bit problematic to pass that `options` table to those sub-queries too. Here we just optimized the cascade deletes. I think it is fine hard-coded.

bungle: What I also mean that It is also a bit unintuitive to pass `options.page_size` to `delete`. The cascade deletion is a side effect of `delete`, and we just want to make it faster by iterating a bigger result set at once.

thibaultcha: > The default was picked up for admin api, and then after that a lot of places we have changed it for performance reasons.

I am very much aware, having initially chosen those values :) I was only mentioning it to say that now that the pattern has spread in a lot of hard-coded magic numbers, we should correct that and introduce a variable. Maybe a constant at first, and something we can configure later on, as stated in my above comment. 

It is better hard-coded as a constant, maybe an attribute of the connector (possibly set in the connector's parent class) or DAO. E.g. `options.page_size_subqueries`.

> I think it is fine hard-coded.

No; besides the obvious gains from removing magic numbers in favor of a named constant, it also doesn't close the door to subsequent performance improvements, if the day comes that we want to make it configurable.

thibaultcha: Merging as-is for now, we can skip the above improvements for a separate PR with its own scope.

thibaultcha:
@bungle I think this can go in master (thus, I will be manually merging it)

kong/db/strategies/cassandra/init.lua
bungle: @thibaultcha yes, i have commits almost ready for that. I will push tomorrow a new pr.

PR #4772, opened 2019-07-03
feat(conf) cassandra_consistency to cassandra_consistency_proxy|admin

bungle:
#### cassandra_consistency to cassandra_consistency_proxy|admin 

Spits `cassandra_consistency` configuration option to:
1. `cassandra_consistency_proxy`
2. `cassandra_consistency_admin`

and deprecates the `cassandra_consistency` configuration option.

The configuration options can take 1 or 2 values, for example:

```
cassandra_consistency_proxy = read:ONE, write:QUORUM
cassandra_consistency_proxy = QUORUM
```

#### allow cassandra connector queries to override consistency

A small commit to fix Cassandra connector to allow cassandra `connector:query` to override consistency, which seemed like this was already used in some places:

- `CassandraConnector:insert_lock`
- `CassandraConnector:read_lock`
- `CassandraConnector:remove_lock`

But the options were not respected by the `CassandraConnector:query` or `CassandraConnector:batch` (they were actually overridden).

#### enable logging on restart even when stopping errors

Previously if you ran `kong restart` and the Kong was not running, the logging would be turned off. This commit enables the logging even on such case.

#### log config warnings only on startup (and restart)

The new deprecated config warnings were a bit too verbose, and also a wrong (e.g. it warned users on `kong stop` to stop using deprecated config values even when user didn't specify one). This commit fixrs that.


thibaultcha:
I recall we discussed several times in the past that we should split such consistency settings between Proxy and Admin API. We can then offer granularity between reads and writes if we judge it necessary. E.g.:

1. `cassandra_consistency_proxy = ONE` and `cassandra_consistency_admin= QUORUM`.
2. Use the conf_loader deprecation helper: `cassandra_consistency -> cassandra_consistency_proxy + cassandra_consistency_admin`.
3. Implement some warning logs in the conf_loader when we detect misconfigurations e.g. using `QUORUM` instead of `LOCAL_QUORUM` if the LB policy is datacenter-aware.
4. Eventually, also offer R/W granularity (really not sure about this one for now, we need to dig into specific use-cases): `cassandra_consistency_admin = read:ONE,write:QUORUM` (or other similar notation).

We also need to keep in mind that the consistency settings need to be overridable anytime, e.g., migrations need to be run with consistency over availability. This should already be the case but we should keep it in mind.

The scope of this PR should be 1. and 2. (3. being a nice to have, but not a blocker).

More thoughts on the topic of Cassandra configuration properties but not enough time to write them all down at the moment, sorry!

bungle:
The PR implements 1. to 4. from @thibaultcha's list above.

kong/conf_loader.lua
```diff
@@ -123,12 +143,14 @@ local CONF_INFERENCES = {
     deprecated = {
       replacement = "nginx_http_upstream_keepalive",
       alias = function(conf)
-        -- called before check_and_infer(), must parse ourselves
-        if tonumber(conf.upstream_keepalive) == 0 then
-          conf.nginx_http_upstream_keepalive = "NONE"
+        if conf.upstream_keepalive then
```
kikito: It seems that this fix to `conf.upstream_keepalive` and `conf.nginx_http_upstream_keepalive` should be on a separate commit, if I'm reading this correctly it has nothing to do with `cassandra_consistency_*`

spec/02-integration/03-db/01-db_spec.lua
```diff
@@ -662,6 +662,57 @@ for _, strategy in helpers.each_strategy() do
...
+      local three = require "cassandra".consistencies.three
+
+      local conf = utils.deep_copy(helpers.test_conf)
+
+      conf.cassandra_contact_points = { "localhost" }
+      conf.cassandra_consistency_proxy = {
+        "read:QUORUM",
+        "write:THREE",
+        read  = "quorum",
+        write = "three",
```
kikito: I understand why `read = "quorum", write = "three"` are there - that's how they look after the config has been parsed. Why are `"read:QUORUM"` and `"write:THREE"` here too?

kong.conf.default
```diff
@@ -579,9 +579,24 @@
...
 #cassandra_keyspace = kong       # The keyspace to use in your cluster.
 
-#cassandra_consistency = ONE     # Consistency setting to use when reading/
-                                 # writing to the Cassandra cluster.
-                                 #
+#cassandra_consistency_proxy = read:ONE, write:ONE  # Consistency setting to
+                                                    # use when reading/writing
+                                                    # to the Cassandra cluster
+                                                    # on proxy.
+                                                    #
```
thibaultcha: extra line

kong.conf.default
```diff
@@ -579,9 +579,24 @@
...
+#cassandra_consistency_admin = read:ONE, write:ONE  # Consistency setting to
+                                                    # use when reading/writing
+                                                    # to the Cassandra cluster
+                                                    # on admin API.
+                                                    #
+                                                    #
```
thibaultcha: extra line as well.

kong/conf_loader.lua
thibaultcha: This check is unnecessary, since we have `if conf[property_name] ~= nil` at https://github.com/Kong/kong/blob/next/kong/conf_loader.lua#L822. Besides, it may even harmful since exposing an anti-pattern that could be copy-pasted for those boolean configuration properties.

kong/conf_loader.lua
```diff
@@ -162,17 +184,25 @@ local CONF_INFERENCES = {
...
-                              "THREE",
-                              "LOCAL_ONE",
-                            }
-                          },
+  cassandra_consistency = {
+    enum = CASSANDRA_CONSISTENCY,
+    deprecated = {
+      replacement = "cassandra_consistency_proxy and cassandra_consistency_admin",
+      alias = function(conf)
+        if conf.cassandra_consistency then
```
thibaultcha: Ditto, should not be necessary since we have the proper safeguards in the aliasing caller code.

kong/conf_loader.lua
```diff
@@ -357,6 +389,99 @@ local function check_and_infer(conf, opts)
...
+                                  "write:%s)", name, concat(option, ", "),
+                                  concat(CASSANDRA_CONSISTENCY, ", read:"),
+                                  concat(CASSANDRA_CONSISTENCY, ", write:"))
+      end
+
+    elseif number_of_values == 1 then
+      local found
+      for _, name in ipairs(CASSANDRA_CONSISTENCY) do
+        if name == option[1] then
+          found = true
```
thibaultcha: Missing a `break` statement after this assignment

kong/conf_loader.lua
```diff
@@ -357,6 +389,99 @@ local function check_and_infer(conf, opts)
...
+        end
+
+      else
+        read_consistency  = option[1]
+        write_consistency = option[1]
+      end
+    end
+
+    if not read_consistency then
+      read_consistency  = conf.cassandra_consistency
```
thibaultcha: style: extra space before `=`

kong/conf_loader.lua
```diff
@@ -357,6 +389,99 @@ local function check_and_infer(conf, opts)
...
+    end
+
+    if not write_consistency then
+      write_consistency = conf.cassandra_consistency
+    end
```
thibaultcha: Why do we have those two fallbacks? 

- If the user specified `cassandra_consistency` and none of those above two properties, the deprecated alias should assign `_admin` and `_proxy` (and log the warning).
- If the user specified did not specify `cassandra_consistency`, then he or she removed it from a config file or env variable, and we should fully rely on the new `_admin` and `_proxy` settings (since they have defaults and we removed `cassandra_consistency` from the defaults list as well).

What I am trying to say is that if specified, the _deprecated_ property should **still have precedence** over the new one(s), and the warning logs should be printed. As soon as the user removes the deprecated property, then the new properties reign supreme, and their validation rules apply, with no fallback whatsoever to deprecated properties.

This feels wrong here. `_proxy` and `_admin` should not be allowed to be empty in this code path without `errors` being populated. This fallback should only be in the `alias` function, and invoked _before_ `check_and_infer()`.

kong/conf_loader.lua
```diff
@@ -260,7 +260,7 @@ local _nop_tostring_mt = {
 
 -- Validate properties (type/enum/custom) and infer their type.
 -- @param[type=table] conf The configuration table to treat.
-local function check_and_infer(conf)
+local function check_and_infer(conf, opts)
```
thibaultcha: This would have been more correct:

```lua
if not opts.starting then
  log.disable()
end

-- ...

-- end of the function
```lua
if not opts.starting then
  log.enable() -- re-enable
end
```
```

File: kong/cmd/restart.lua [hunk 10]
```diff
@@ -20,7 +20,10 @@ local function execute(args)
 
   log.enable()
 
-  pcall(stop.execute, args, { quiet = true })
+  local ok = pcall(stop.execute, args, { quiet = true })
+  if not ok then
+    log.enable()
+  end
```
thibaultcha: `log.enable()` is idempotent, so we could call it unconditionally, we don't have to check for `pcall()`'s return value. In fact, given `quiet = true` will call `log.disable` in `kong stop`'s `execute()` method, we should move the above `log.enable()` on L.21 down here instead, since it is a no-op where it is above right now.

kong/conf_loader.lua
```diff
@@ -357,6 +389,99 @@ local function check_and_infer(conf, opts)
...
+      write_consistency = conf.cassandra_consistency
+    end
+
+    if read_consistency and write_consistency then
+      if opts.starting and is_cassandra and is_dc_aware then
+        local suggestion
+        if read_consistency == "QUORUM" or
+           read_consistency  == "ONE" then
+          if write_consistency == "QUORUM" or
+             write_consistency == "ONE" then
```
thibaultcha: This logic should be "if not contains('LOCAL')". Currently this skips other consistency settings such as `TWO`, `THREE`, or even `ALL`.

PR #4775, opened 2019-07-03
fix(router) several fixes related to sub-matching priorities

### LINKED ISSUE #3094: Incorrect routing priority with wildcard host

thibaultcha:
This patch intents to replace the following PRs:

- https://github.com/Kong/kong/pull/4152
- https://github.com/Kong/kong/pull/4153

### Summary

It provides more appropriate fixes to each of the above issues. There may be more tests to write, which is something I will give more thoughts to when I have some more time on my hands. There are also several things we could do with the router to increase tests coverage and maintainability, but I am running short of time to write down ideas at the moment.

### Full changelog

- fix: edge-case resulting in failure to match regex paths
- fix: edge-case resulting in failure to match wildcard hosts
- perf: refine hosts reducer by accounting for the request's Host header
- tests: fix many potential false-positive and confused usage of luassert assertions

### Notes

1. The nature of these changes introduce breaking changes in the router, hence the PR is based on `next`. Nonetheless, those changes do fix edge-cases causing the router to not behave as expected/documented ("expected" meaning here that the behavior broke once the scenario was slightly more complexed than already tested; nonetheless basic matching rules were not being respected).
2. As of today (July 3rd, 2019), our [proxy guide documentation](https://docs.konghq.com/1.2.x/proxy/#evaluation-order) states that plain paths are evaluated before regex paths, and implies that the former have precedence over the latter, which should not be the case, as already asserted by existing test cases. Note as well the invalid link resulting in plain markdown inlined in the rendered document (cc @Kong/team-docs).

### Notes to reviewers 

- Reviewing commits individually and in chronological order will help (some changes will be invisible in the final diff, but should be looked at).
- We need to think about more test cases for sub-matching priorities edge-cases and catch potentially other broken use-cases fixed by this patch; and add regression tests for those. More can be done for testing and maintainability, but that is outside the scope of this PR.
- Each commit has some background and a detailed description of the issue.

### Issues resolved

Fix #3094 
Replaces #4152 
Replaces #4153


thibaultcha:
cc @Tieske @RobSerafini - FYI

kong/router.lua
```diff
@@ -168,6 +173,8 @@ local function marshall_route(r)
     preserve_host  = route.preserve_host == true,
     match_rules    = 0x00,
     match_weight   = 0,
+    match_weight2  = 0,
```
thibaultcha: Could be renamed to `submatch_weight` - open to suggestions.

PR #4780, opened 2019-07-05
fix(proxy) do not clear upgrade header (case-insensitive), fix #4779

### LINKED ISSUE #4779: kong 1.2.1 websocket fail
image
image

bungle:
### Summary

Ensures that `Upgrade` header is not cleared on `websocket`. This is the second patch after the initial patch PR #4719. It seems like some servers send back `Upgrade: WebSocket` instead of `Upgrade: websocket` which caused the proxy to clear the `Upgrade` header. This ensures that the value is compared case-insensitive.

The test is missing as it is a bit tricky to implement that as our mock websocket server always sends the `Upgrade` header back as lower case `websocket` which does not cause this issue. But the fix in itself is rather simple.

### Issues resolved

Fix #4779

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1349,7 +1349,7 @@ return {
       end
 
       if var.upstream_http_upgrade and
-         var.upstream_http_upgrade ~= var.upstream_upgrade then
+         lower(var.upstream_http_upgrade) ~= lower(var.upstream_upgrade) then
         header["Upgrade"] = nil
       end
 


kong/runloop/handler.lua
```diff
@@ -1349,7 +1349,7 @@ return {
       end
 
       if var.upstream_http_upgrade and
-         var.upstream_http_upgrade ~= var.upstream_upgrade then
+         lower(var.upstream_http_upgrade) ~= var.upstream_upgrade then
```
Tieske: @bungle is `var.upstream_upgrade` guaranteed to be lowercase? since we're only one-sided lowering...

bungle: @Tieske that's right.

bungle: Though, as it is publicly accessible, perhaps it makes sense to do it on both sides.

bungle: @Tieske changed to `lower` on both sides.

jeremyjpj0916:
@bungle @thibaultcha @gszr Was the intent to release 1.2.2? Out websocket services are broken in Stage on 1.2.1 and we need this adjustment I reckon. I checked luarocks, docker-kong repo and bintray and 1.2.2 was never released.

PR #4788, opened 2019-07-09
fix(conf) cassandra consistency changes that were reported after #4772 was merged

bungle:
### Summary

Fixes cassandra consistency changes that were reported after #4772 was merged.

### DIFF
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -143,14 +143,12 @@ local CONF_INFERENCES = {
     deprecated = {
       replacement = "nginx_http_upstream_keepalive",
       alias = function(conf)
-        if conf.upstream_keepalive then
-          -- called before check_and_infer(), must parse ourselves
-          if tonumber(conf.upstream_keepalive) == 0 then
-            conf.nginx_http_upstream_keepalive = "NONE"
+        -- called before check_and_infer(), must parse ourselves
+        if tonumber(conf.upstream_keepalive) == 0 then
+          conf.nginx_http_upstream_keepalive = "NONE"
 
-          elseif conf.nginx_http_upstream_keepalive == nil then
-            conf.nginx_http_upstream_keepalive = tostring(conf.upstream_keepalive)
-          end
+        elseif conf.nginx_http_upstream_keepalive == nil then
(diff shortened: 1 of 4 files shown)


kong/conf_loader.lua
```diff
@@ -425,60 +421,35 @@ local function check_and_infer(conf, opts)
...
+    if is_cassandra and is_dc_aware and ((
+      read_consistency ~= "LOCAL_QUORUM"  and read_consistency ~= "LOCAL_ONE"
+    ) or (
+      write_consistency ~= "LOCAL_QUORUM" and write_consistency ~= "LOCAL_ONE"
+    )) then
+      log.warn("Cassandra is configured to use a data center aware load "    ..
+               "balancing strategy '%s' while the consistency of '%s' is "   ..
+               "set to '%s'. It is recommended to use 'LOCAL_QUORUM' or "    ..
+               "'LOCAL_ONE' when specifying read/write consistency in such " ..
+               "scenario.", conf.cassandra_lb_policy, name,
```
thibaultcha: Should be `scenarios` (or "scenarii" if we wanted to be pedantic about it, but let's avoid it) .

Or maybe we could just simplify this warning:

```
A datacenter-aware Cassandra load balancing policy is in use ('cassandra_lb_policy=%s')
but some consistency levels are not datacenter-aware (e.g. LOCAL_QUORUM, LOCAL_ONE, or EACH_QUORUM).
```

And we only print the warning once, even if both `cassandra_consistency_admin` and `cassandra_consistency_proxy` are mis-configured.

hishamhm: (Off-topic, but since I was linguistics-nerd-sniped into it: [scenarii](https://english.stackexchange.com/questions/11775/what-is-the-plural-of-scenario) is a case of [hypercorrection](https://en.wikipedia.org/wiki/Hypercorrection) — "scenarios" is in fact the formally correct form.)

kikito:
Moving this out of the 1.3.0 milestone, replacing by #4811 instead.

hishamhm:
These changes should go in when the original PR this builds on gets reworked.

PR #4790, opened 2019-07-10
refactor(db) use 1000 as default page size for iteration

bungle:
### Summary

It was noted at #4770 that we should get rid of repeating pattern of adding `1000` as a magic value whenever we want to make iteration more performant over database rows with `each[_*]`.

This PR changes so that:
`each` and `each_for_*` will always use `1000` as a default page size for each iteration.

The `100` is still used as a default for `page` and `page_for_*`.

During changing that, I also refactored postgres cluster events to use similar iteration strategy as Cassandra (which was marked as TODO).


hishamhm:
That's a nice cleanup! (And I just had a branch that was using `each()` and will benefit immediately from this)

thibaultcha:
Taking a look (I think there was some confusion as to what the point of my comment was, but will look in more details in a bit)

thibaultcha:
This should go in master, since #4770 was merged there.

kong/db/iteration.lua
```diff
@@ -15,7 +15,7 @@ end
 
 local function page_iterator(pager, size, options)
   local page = 1
-  size = size or 100
+  size = size or 1000
```
thibaultcha: It is a bit unfortunate that the C* strategy uses this low-level module for cascade deletes, because it prevents us from being able to specify the default fallback (`1000`) in the DAO module only, making it authoritative, and enabling a few other nice-to-haves.

Besides, the approach of having the constants module loaded all of our DAO/strategies/cluster events modules isn't ideal either, as it reduces encapsulation of all of these modules and prevents us from **configuring** this property (at _runtime_), which is still the ultimate goal we should be looking at here.

---

The structure I had in mind when looking at #4770 was to have the `kong.db` module be the authoritative source for default values such as `default_page_size` and `max_page_size`. E.g. a table stored as an attribute: `kong.db.defaults`, and given when instantiating it:

```lua
-- kong/init.lua
local db = DB.new(config, nil, {
  page_size = constants.default_page_size,
  max_page_size = constants.max_page_size,
})
```

or:

```lua
-- kong/db/init.lua
function DB.new(kong_config, strategy)
  -- ...

  local self = {
    -- ...
    defaults = {
      page_size = constants.default_page_size,
      max_page_size = constants.max_page_size,
    },
  }

  -- ...
end
```

Why?

- It allows us to give a reference to this table to the DAOs (`kong.db.dao`) and the strategies or low-level helpers (`kong.db.iteration`) as needed.
- In the future, the values in this table can be updated at runtime as necessary based on the size of the database tables.
- In `options` table of all of our DAO functions, we accept a new `options.defaults` table. Then the DAO's fallback code in `each()` and `each_for_*()` becomes: `page_size = options.defaults.page_size or self.defaults.page_size`. 
- The [Admin API DAO options](https://github.com/Kong/kong/blob/master/kong/api/endpoints.lua#L127-L129) could would contain a new `defaults = { page_size = 100, max_page_size = 1000 }` attribute, given to all of the `each()` calls in the Admin API.
- In this low-level `kong.db.iteration` module, we `assert(page_size ~= nil)`, and the Cassandra strategy's cascade delete code grabs the `self.defaults.page_size` value, which can never be `nil`.

kong/cluster_events/strategies/cassandra.lua
```diff
@@ -26,7 +27,7 @@ local mt = { __index = _M }
 function _M.new(db, page_size, event_ttl)
   local self  = {
     cluster   = db.connector.cluster,
-    page_size = page_size or 100,
+    page_size = page_size or constants.DEFAULT_ITERATION_SIZE,
```
thibaultcha: The `kong.cluster_events` module should be authoritative with regards to this value. We should probably guarantee that the cluster_events strategies always receive a `page_size` argument.

kong/constants.lua
```diff
@@ -133,4 +133,6 @@ return {
   },
   PROTOCOLS = protocols,
   PROTOCOLS_WITH_SUBSYSTEM = protocols_with_subsystem,
+  DEFAULT_ITERATION_SIZE = 1000,
+  DEFAULT_PAGE_SIZE = 100,
```
thibaultcha: I wouldn't be against `MAX_PAGE_SIZE` and `CLUSTER_EVENTS_PAGE_SIZE` as separate constants.

kong/cluster_events/strategies/cassandra.lua
bungle: Was there by the way a reason for this being `100` instead of `1000`. Should this PR still use `100` instead?

kong/db/iteration.lua
bungle: @thibaultcha I tried to do that, but there were some dependencies, that would take time to figure out just right, and I could not finish on that. In this current shape I think we at least get a bit closer to the goal.

kong/constants.lua
bungle: these were added.

kikito:
I have removed this PR from the 1.3.0 milestone.

bungle:
Changed this to be based on `next` as it was taken away from `1.3.0`.

PR #4792, opened 2019-07-10
feat(balancer) healthcheck uses ip+port+hostname combo

locao:
- Changes needed for lua-resty-healthcheck 1.0.0 which requires that hostnames
are used combined with IP and port to know exactly which host is being checked.
- lua-resty-healthcheck version bumped in rockspeck.

### DIFF
--- a/kong/db/dao/targets.lua
+++ b/kong/db/dao/targets.lua
@@ -250,9 +250,22 @@ function _TARGETS:page_for_upstream_with_health(upstream_pk, ...)
     -- Note that lua-resty-dns-client does retry by itself,
     -- meaning that if DNS is down and it eventually resumes working, the
     -- library will issue the callback and the target will change state.
-    target.health = health_info
-                   and (health_info[target.target] or "DNS_ERROR")
-                   or  "HEALTHCHECKS_OFF"
+    if health_info[target.target] ~= nil and
+      #health_info[target.target].addresses > 0 then
+      target.health = "UNHEALTHY"
+      -- If any of the target addresses are healthy, then the target is
+      -- considered healthy.
+      for _, address in ipairs(health_info[target.target].addresses) do
+        if address.health == "HEALTHY" then
+          target.health = "HEALTHY"
+          break
+        end
+
+      end
+    else
+      target.health = "DNS_ERROR"
+    end
+    target.data = health_info[target.target]
   end
 
   return targets, nil, nil, next_offset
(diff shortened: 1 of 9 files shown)


spec/02-integration/04-admin_api/08-targets_routes_spec.lua
```diff
@@ -377,7 +377,14 @@ describe("Admin API #" .. strategy, function()
           assert.equal(targets[3].target, res.data[2].target)
           assert.equal(targets[2].target, res.data[3].target)
           for i = 1, n do
-            assert.equal(health, res.data[i].health)
+            if res.data[i].data ~= nil and res.data[i].data.addresses ~= nil then
+              for j = 1, #res.data[i].data.addresses do
+                assert.equal(health, res.data[i].data.addresses[j].health)
+              end
```
hishamhm: I think we should _still_ check `assert.equal(health, res.data[i].health)` when we enter this branch and check for `res.data[i].data.addresses`. This ensures backwards compatibility.

kong/db/dao/targets.lua
```diff
@@ -250,9 +250,22 @@ function _TARGETS:page_for_upstream_with_health(upstream_pk, ...)
     -- Note that lua-resty-dns-client does retry by itself,
     -- meaning that if DNS is down and it eventually resumes working, the
     -- library will issue the callback and the target will change state.
-    target.health = health_info
-                   and (health_info[target.target] or "DNS_ERROR")
-                   or  "HEALTHCHECKS_OFF"
```
hishamhm: What happened to the HEALTHCHECKS_OFF case in this test? Does `get_upstream_health` always return a `health_info` in line 238 above now? From my reading below it seems that it would end up defaulting to `UNHEALTHY`...?

I'm not saying this is broken because we do have a test in the suite that verifies that HEALTHCHECKS_OFF is still returned and that is passing, but I'm confused and I'd like to understand the code changes a bit better.

kong/runloop/balancer.lua
```diff
@@ -217,9 +217,9 @@ do
 
     ------------------------------------------------------------------------------
     -- Callback function that informs the healthchecker when targets are added
-    -- or removed to a balancer.
+    -- or removed to a balancer and when targets health status change.
     -- @param balancer the ring balancer object that triggers this callback.
-    -- @param action "added" or "removed"
+    -- @param action "added", "removed", or "health"
```
hishamhm: Is the "health" option here being handled just for logging purposes? I believe lua-resty-healthcheck already logs health changes, so these would produce duplicated log lines. If this was done just to avoid the warning in line 252, I'd consume the `"health"` actions and do nothing with it, here, or lower them to DEBUG in case we're concerned that the lua-resty-healthcheck logs and balancer responses don't match.

hishamhm:
You can squash 6b9d8ec to the commit that is fixes, to avoid adding broken states to the git history (makes it easier to use `git bisect` later).

hishamhm:
minor: in the commit message for 10a0e60: `fix(balancer.lua)` -> `fix(balancer)`

kong/db/dao/targets.lua
locao: Thanks, it was really broken. I added a test for that and fixed the test.

kong/runloop/balancer.lua
locao: Log message lowered to DEBUG.

spec/02-integration/04-admin_api/08-targets_routes_spec.lua
locao: Fixed!

hishamhm:
note to the team: let's hold merge until the merge-commit above is removed

PR #4798, opened 2019-07-16
Revert "Certificate" DAO optional key change and add new core entity "CA Certificates"

dndx:
Background
----------
There has been cases were storing certificates without respective
private key may be needed. For example, when validating upstream
certificate.

One may thinks that we can allow the existing "Certificate"
entity to accept a `key = null`, but that will be problematic
because we will have to write large amount of custom DAO
methods (and respective tests) which will be hard to maintain as the
usage grows.

Solution
--------
We add a new entity type "CA Certificates", which is principally the same as the
existing "Certificate" entity but without the `key` field. The new
entity is stored in a new table called `"ca_certificates"`.


kong/db/schema/entities/cas.lua
```diff
@@ -0,0 +1,41 @@
...
+
+  entity_checks = {
+    { custom_entity_check = {
+      field_sources = { "cert", },
+      fn = function(entity)
+        local cert = openssl_x509.new(entity.cert)
+        local not_before, not_after = cert:getLifetime()
+        local now = ngx_time()
+
+        if not_before > now then
```
hishamhm: is this a desired constraint? isn't it a case that someone may be preparing a certificate so it kicks in at some point in the future? (I assume that's the use of the `Not Before` attribute). In that case I don't see a problem in storing it in the database.

kong/db/schema/entities/cas.lua
```diff
@@ -0,0 +1,41 @@
...
+          return nil, "certificate unusable, \"Not Before\" time is " ..
+                      "in the future"
+        end
+
+        if not_after < now then
```
hishamhm: for this one I can see the obvious application, though I always get a "what about clock skew" kneejerk reaction when I see things like this. :) But I can't think up of a problem scenario right now (even considering short-lived certificates), and it adds user-friendliness. :+1:

hishamhm:
Not a blocker if everyone else is fine with it, but I'm not a super fan of the name `cas` (I'm sure people bumping into it will ask what that is). I looked around and people do use the name "CA certificates" for those things (after all the CA is the real-world entity (Verisign, etc.) who _issues_ the CA certificate). How about `ca_certificates`?

kong/db/schema/entities/cas.lua
dndx: Yeah. Regardless the reason a CA that has expired is no use for Kong, so this check definitely helps with avoiding some of the surprises.

PR #4800, opened 2019-07-16
Upstream mutual TLS handshake support

dndx:
This is achieved by specifying the Certificate to be used during TLS
handshake via the newly added `client_certificate` property on the Service
core entity.

Note: Based on https://github.com/Kong/kong/pull/4797, review that PR first.


kong/pdk/service/request.lua
```diff
@@ -417,6 +419,49 @@ local function new(self)
   end
 
 
+  if ngx.config.subsystem == "http" then
```
james-callahan: Why not support this in stream mode?
Should we provide a stub/error function in stream mode until if we are delaying the work?

kong/runloop/certificate.lua
```diff
@@ -119,15 +119,23 @@ local function fetch_sni(sni, i)
 end
 
 
-local function fetch_certificate(sni)
-  local certificate, err = singletons.db.certificates:select(sni.certificate)
+local function fetch_certificate(pkey, sni_name)
```
james-callahan: I was confused by this variable name: `pkey` is frequently used to mean "private key" in certificate handling code.

kong/db/schema/entities/cas.lua
```diff
@@ -0,0 +1,41 @@
+local typedefs = require "kong.db.schema.typedefs"
+local openssl_x509 = require "openssl.x509"
+local ngx_time = ngx.time
```
thibaultcha: Note that we don't have to cache those global lookups in the Admin API server, it being a very cold path compared to our proxy server.

kong/db/schema/entities/cas.lua
```diff
@@ -0,0 +1,41 @@
...
+
+return {
+  name        = "cas",
+  primary_key = { "id" },
+
+  fields = {
+    { id             = typedefs.uuid, },
+    { created_at     = typedefs.auto_timestamp_s },
+    { cert           = typedefs.certificate { required = true }, },
+    { tags           = typedefs.tags },
```
thibaultcha: style: we find it to be easier not to align `=` signs anymore, if you care to revert the change... Not a blocker for me.

kong/db/schema/entities/cas.lua
```diff
@@ -0,0 +1,41 @@
...
+          return nil, "certificate unusable, \"Not Before\" time is " ..
+                      "in the future"
+        end
+
+        if not_after < now then
+          return nil, "certificate expired, \"Not After\" time is in the past"
+        end
+
+        if not cert:getBasicConstraints("CA") then
+          return nil, "certificate does not have \"CA\" basic constrain set"
```
thibaultcha: Should we produce a more "non SSL expert" error message? Something like `"certificate does not appear to be a CA (missing "CA" basic constraint)"`.

Regardless, we should fix the typo too ("constraint").

kong/constants.lua
```diff
@@ -87,9 +87,9 @@ return {
   -- schemas of dependencies need to be loaded first.
   CORE_ENTITIES = {
     "consumers",
+    "certificates",
```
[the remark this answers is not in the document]
thibaultcha: This list is not supposed to be alphabetically sorted (not should it give the impression that it is) since its order is semantically meaningful?

kong/runloop/certificate.lua
thibaultcha: Agreed, we use `pk` or something similar in our DAO code when we mean "primary key". We should not use `pkey` here (I too got confused)

kong/runloop/certificate.lua
```diff
@@ -119,15 +119,23 @@ local function fetch_sni(sni, i)
...
+  local certificate, err = singletons.db.certificates:select(pkey)
   if err then
-    return nil, "failed to fetch certificate for '" .. sni.name .. "' SNI: " ..
-                err
+    if sni_name then
+      return nil, "failed to fetch certificate for '" .. sni_name .. "' SNI: " ..
+                  err
+    end
+
+    return "failed to fetch certificate " .. pkey.id
```
thibaultcha: A big mistake here I believe: missing `return nil` (first return value)? This would make the above string the valid, cached return value of `fetch_certificate`

kong/db/schema/entities/cas.lua
dndx: This has been reverted.

kong/db/schema/entities/cas.lua
dndx: Changed to `certificate does not appear to belong to a CA (missing the "CA" basic constraint)`.

kong/constants.lua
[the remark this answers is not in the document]
dndx: Because `service` has FK reference to `certificates`, it has to load before `service` entity schema is loaded.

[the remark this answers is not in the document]
thibaultcha: Yep, that was clear. Was just not agreeing with @james-callahan on this one that it had to be alphabetically sorted.

kong/runloop/certificate.lua
dndx: Fixed. Thanks for pointing out!

kong/runloop/certificate.lua
dndx: Good catch @thibaultcha, that was fixed!

kong/pdk/service/request.lua
dndx: Talked with @thibaultcha, it should be alright to skip this for now until someone complaints (and it is not a silent failure)...

dndx:
All comments addressed. Also pushed a new commit https://github.com/Kong/kong/pull/4800/commits/d1c8121be0404d8da69c010aaa369a217bd40042 that adds additional schema constraints for the new field. Please take a peek @thibaultcha @hishamhm :).

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -55,9 +68,10 @@ return {
         PRIMARY KEY (partition, id)
       );
 
-
-
       ALTER TABLE routes ADD headers map<text,frozen<set<text>>>;
+      ALTER TABLE services ADD client_certificate_id uuid;
+
+      CREATE INDEX IF NOT EXISTS services_client_certificate_id_idx ON services(client_certificate_id);
```
thibaultcha: style: in terms of migrations styling, we follow a rule to have 3 line breaks between statements related to a given feature (hence why the line breaks you removed here. We apply this for both Postgres and Cassandra migrations.

spec/02-integration/05-proxy/18-client_certificate_spec.lua
```diff
@@ -0,0 +1,135 @@
+local helpers = require "spec.helpers"
+local ssl_fixtures = require "spec.fixtures.ssl"
+
+
+local fixtures = {
+  http_mock = {
+    upstream_mtls = [[
+      server {
+          server_name example.com;
+          listen 16798 ssl;
```
thibaultcha: If possible, it'd be better to use a unix socket here, but not a blocker.

thibaultcha:
Agreed with @kikito on the git history cleanup regarding fix commits as well

spec/02-integration/05-proxy/18-client_certificate_spec.lua
dndx: I don't think we create Services with unix domain sockets as upstream in the test suite so most likely it is not supported.

PR #4809, opened 2019-07-18
feat(cmd) implement `kong config db_export` command

kikito:

### DIFF
--- a/kong/cmd/config.lua
+++ b/kong/cmd/config.lua
@@ -18,6 +18,37 @@ local accepted_formats = {
 }
 
 
+local function db_export(filename, conf)
+  if pl_file.access_time(filename) then
+    error(filename .. " already exists. Will not overwrite it.")
+  end
+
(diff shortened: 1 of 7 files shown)


hbagdi:
`basic-auth` credentials are hashed in the database. Should we skip exporting them?

kikito:
@hbagdi Unsure about this. If I'm understanding the [crypto algorithm](https://github.com/Kong/kong/blob/master/kong/plugins/basic-auth/crypto.lua) correctly, it uses the consumer id as salt, and such ids are also exported. So exported and reimported credentials should work, if they accompany their Consumers.

If your concern is not whether we *could* but whether we *should* ... well that's where I don't know.

I guess we could add a filter option to the export command line (something like `kong config db_export except=basic_auth_credentials,snis`). But for 1.3.0 we're definitively going with the "getting the basic functionality out and running". It is expected that we'll iterate over this a bit in future versions, or at least that's my understanding.

PR #4810, opened 2019-07-18
chore(deps) bump worker-events to 1.0.0

Tieske:
This version has the `shm_retries` option (default 5) to fix
potential 'no memory' errors, as reported by @hbagdi .

Additionally the return values of `post`, `post_local` and `poll`
behave more Lua-ish. So we can now check on `ok` instead of on `err`

### DIFF
--- a/kong/cache.lua
+++ b/kong/cache.lua
@@ -175,7 +175,11 @@ function _M.new(opts)
             end
           end,
           broadcast = function(channel, data)
-            opts.worker_events.post(channel_name, channel, data)
+            local ok, err = opts.worker_events.post(channel_name, channel, data)
+            if not ok then
+              log(ERR, "failed to post event '", channel_name, "', '",
+                       channel, "': ", err)
+            end
           end
         }
       })
--- a/kong-1.2.1-0.rockspec
+++ b/kong-1.2.1-0.rockspec
@@ -31,7 +31,7 @@ dependencies = {
   "luasyslog == 1.0.0",
   "lua_pack == 1.0.5",
   "lua-resty-dns-client == 3.0.2",
-  "lua-resty-worker-events == 0.3.3",
+  "lua-resty-worker-events == 1.0.0",
   "lua-resty-mediador == 0.1.2",
   "lua-resty-healthcheck == 0.6.2",
   "lua-resty-cookie == 0.1.0",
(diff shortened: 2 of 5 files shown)


kong/cache.lua
```diff
@@ -176,6 +176,7 @@ function _M.new(opts)
           end,
           broadcast = function(channel, data)
             opts.worker_events.post(channel_name, channel, data)
+            -- TODO: do we need to log a failed post?
```
hishamhm: I think so — this is something that shouldn't ever happen in normal operation, right?

Tieske: worker events will also log the fact that it couldn't post, but not where the data originated from. So it still adds value I think

hishamhm:
When merging, let's squash this into a single `chore(deps)` commit adjusting the API. The `fix(declarative)` commit is misleading for anyone in the future reading Git history because it would make it seem like it was fixing a bug in a past Kong release (when in fact it's fixing up the `chore(deps)` change right above — so we can just squash it in the same commit).

kong/cache.lua
Tieske: added a log line

hishamhm:
Squashed and rebased this on `next`, waiting for CI.

kong/cache.lua
```diff
@@ -175,7 +175,11 @@ function _M.new(opts)
             end
           end,
           broadcast = function(channel, data)
-            opts.worker_events.post(channel_name, channel, data)
+            local ok, err = opts.worker_events.post(channel_name, channel, data)
+            if not ok then
+              log(ERR, "failed to post event '", channel_name, "', '",
+                  channel, "': ", err)
```
thibaultcha: A minor styling nitpick: note that our preferred style with this particular function call is to align new-lines with the _second_ argument:
```lua
log(ERR, "failed to post event '", channel_name, "', '",
         channel, "': ", err)
```

Not a blocker ofc.

Tieske:
why `next` ?

hishamhm:
because it's a major-version dependency bump? Anyway, the next planned release is 1.3.0, so it won't make a difference.

hishamhm:
But I see what you mean, in the sense that this shouldn't have user-visible impact other than solving the shm problem. This could have arguably gone on `master`, I stand corrected.

kong/cache.lua
Tieske: done

PR #4817, opened 2019-07-22
fix(declarative) compound key for endpoints that can be identical

### LINKED ISSUE #4778: Getting "failure to get a peer from the ring-balancer" error after upgrading to 1.2.x
### Summary

I'm getting "failure to get a peer from the ring-balancer" error when calling the API's of my service after upgrading Kong to 1.2.1 (also tried 1.2.0). The exact same config works well in 1.1.2.
After some troubleshooting, I realized the problem has something to do with the following in my Kong yaml:
- Route1 points to Service1 with Upstream1 and PortX
- Route2 points to Service2 with Upstream2 and PortX (i.e. same as Route1)
- Upstream1 and Upstream2 have the same Target config (i.e. same IP and same Port)

I have this setup because I have my own plugin with different settings for Route1 and Route2 respectively; underlying they should both go to the same microservice and the same port. And I'm using a library to auto-generate the complex Kong yaml. 

Apparently the cause of the error is having multiple Upstreams sharing the same Target (same IP and same Port). If I remove one of the Routes, or simply point Route2 to Service1 and get rid of the thus excess Upstream2 etc., the error will go away.

### Steps To Reproduce

1. See Summary. Define two Routes pointing to different Services with different Upstreams but having the same Target setting.

### Additional Details & Logs

- Kong 1.2.1. No issue with the same Kong yaml in 1.1.2. 
- I am using Kong with declarative config (i.e. DB-less mode) and I am deploying Kong on K8S, just in case this matters.

locao:
# Summary

Endpoints that could be identical under different parents, e.g. identical targets under different upstreams, were being overwritten when using a declarative config. This PR makes auto-generated primary keys to be compound by parent's foreign keys, avoiding that issue.

### Issues resolved

Fix #4778

### DIFF
--- a/kong/db/schema/others/declarative_config.lua
+++ b/kong/db/schema/others/declarative_config.lua
@@ -409,10 +409,24 @@ local function get_key_for_uuid_gen(entity, item, schema, parent_fk, child_key)
     return
   end
 
-  if schema.endpoint_key and
-     item[schema.endpoint_key] ~= nil then
+  if schema.endpoint_key and item[schema.endpoint_key] ~= nil then
+    local key = item[schema.endpoint_key]
+
+    -- If this item has foreign keys with on_delete "cascade", it is inferred
+    -- that its endpoint is not necessarily unique, so its key must be composed
+    -- by parent's key, avoiding that it is overwritten by identical endpoints
+    -- under other parents.
+    for _, field in schema:each_field(item) do
+      if field.type == "foreign" and field.on_delete == "cascade" then
+        local foreign_key_keys = all_schemas[field.reference].primary_key
+        for _, fk_pk in ipairs(foreign_key_keys) do
+          key = key .. ":" .. parent_fk[fk_pk]
+        end
+
+      end
+    end
     -- generate a PK based on the endpoint_key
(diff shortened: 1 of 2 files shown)


kong/db/schema/others/declarative_config.lua
```diff
@@ -409,10 +409,24 @@ local function get_key_for_uuid_gen(entity, item, schema, parent_fk, child_key)
...
 
-  if schema.endpoint_key and
-     item[schema.endpoint_key] ~= nil then
+  if schema.endpoint_key and item[schema.endpoint_key] ~= nil then
+    local key = item[schema.endpoint_key]
+
+    -- If this item has foreign keys with on_delete "cascade", it is infered
+    -- that its endpoint is not necesserally unique, so its key must be composed
+    -- by parent's key, avoiding that it is overwritten by identical endpoints
+    -- under other parents.
```
kikito: Typos: "infered" (inferred) and "necesserally" (necessarily)

locao: Fixed that, thanks!

PR #4822, opened 2019-07-23
chore(makefile) pin the kong-build-tools version

hutchic:
pin the kong-build-tools version so we have more success making reproducible builds / the build scripts are easily tracked.

update the setup task


Makefile
```diff
@@ -23,13 +23,11 @@ KONG_BUILD_TOOLS ?= `grep KONG_BUILD_TOOLS .requirements | awk -F"=" '{print $$2
 KONG_VERSION ?= `cat kong-*.rockspec | grep tag | awk '{print $$3}' | sed 's/"//g'`
 
 setup-release:
-	if cd kong-build-tools; \
-	then git pull; \
-	else git clone https://github.com/Kong/kong-build-tools.git; fi
+	-rm -rf kong-build-tools; \
+	git clone https://github.com/Kong/kong-build-tools.git; fi
```
dndx: We can do `git clone -b $(KONG_BUILD_TOOLS) https://github.com/Kong/kong-build-tools.git; fi` to save the `git reset --hard` later.

hutchic: that won't work if we set `KONG_BUILD_TOOLS` to a sha

Issue #4542, opened 2019-04-25
Basic-Auth does not work with Kong 1.1.2 DB-Less


carnei-ro:
Basic-Auth plugin / consumers does not work similar, comparing DB-Less and Kong with Postgres

**kong.yaml** For databaseless
```
---
_format_version: '1.1'
services:
- name: httpbin
  url: http://httpbin.org/anything

routes:
- name: httpbin
  service: httpbin
  paths:
  - "/httpbin"
  plugins:
  - name: basic-auth

consumers:
- username: Leandro
  basicauth_credentials:
  - username: carneiro
    password: abc123
```

**Commands for kong with database:**
```
http POST ${KONG}:8001/consumers username=Leandro

http POST ${KONG}:8001/consumers/Leandro/basic-auth username=carneiro password=abc123

http POST http://${KONG}:8001/services/ \
 name="httpbin" \
 url="http://httpbin.org/anything"
 
http POST http://${KONG}:8001/services/httpbin/routes \
 paths:='["/httpbin"]' \
 name=httpbin

http POST http://${KONG}:8001/routes/httpbin/plugins \
 name=basic-auth
```

**curl -iu carneiro:abc123 ${KONG}:8000/httpbin**
Kong DBLess output:
```
HTTP/1.1 401 Unauthorized
Date: Thu, 25 Apr 2019 17:53:41 GMT
Content-Type: application/json; charset=utf-8
Connection: keep-alive
Content-Length: 48

{"message":"Invalid authentication credentials"}
```

Kong with database can authenticate.

Comparing the consumer endpoint (**http GET localhost:8001/consumers/Leandro/basic-auth**):
password in Databaseless have not been hashed

hishamhm:
Hi @carnei-ro, I have been able to reproduce, and the problem is indeed as you describe. The `basicauth_credentials` DAO is performing the hashing in database mode to keep the password hashed at rest, but the declarative config parser is not doing that. It's something we need to fix on our end, thank you for reporting.

In the mean time, what does work is keeping the password hashed in the YAML file. This script produces a hash:

**hash_basicauth.lua**
```lua
#!/usr/bin/env resty

if not arg[1] then
  print("usage:   hash_basicauth.lua <password> [<uuid>]")
  print("example: hash_basicauth.lua abc123")
  os.exit(1)
end

local crypto = require "kong.plugins.basic-auth.crypto"
local utils = require "kong.tools.utils"

local id = arg[2] or utils.uuid()
[...]

PR #4830, opened 2019-07-24
fix(db_export) skip basicauth_credentials on db_export

kikito:
This change makes basicauth_credentials non-exportable by db_export.

Explanation: basicauth_credentials have a built-in encryption system - the passwords are encrypted in the database.

If these are exported, they cannot be imported back to the database (they would be re-encrypted, rendering them useless).

The recommended way to import credentials for now is using plain, unencrypted passwords (which can not be obtained from the database).

### DIFF
--- a/kong/plugins/basic-auth/daos.lua
+++ b/kong/plugins/basic-auth/daos.lua
@@ -8,6 +8,9 @@ return {
     primary_key = { "id" },
     cache_key = { "username" },
     endpoint_key = "username",
+    -- Passwords are hashed on insertion, so the exported passwords would be encrypted.
+    -- Importing them back would require "plain" unencrypted passwords instead
+    db_export = false,
     admin_api_name = "basic-auths",
     admin_api_nested_name = "basic-auth",
     fields = {
--- a/spec/02-integration/03-db/08-declarative_spec.lua
+++ b/spec/02-integration/03-db/08-declarative_spec.lua
@@ -161,6 +161,13 @@ for _, strategy in helpers.each_strategy() do
     describe("export_from_db", function()
       it("exports base and custom entities with associations", function()
 
+        -- Add a basicauth_credential to make sure it is not exported
+        db.basicauth_credentials:insert({
+          username = "some_username",
+          password = "secret",
+          consumer = { id = consumer_def.id },
+        })
+
         local fake_file = {
           buffer = {},
           write = function(self, str)
@@ -173,7 +180,7 @@ for _, strategy in helpers.each_strategy() do
         local exported_str = table.concat(fake_file.buffer)
         local yaml = lyaml.load(exported_str)
 
-        -- ensure tags and cluster_ca are not being exported
+        -- ensure tags, cluster_ca & basicauth_credentials are not being exported
         local toplevel_keys = {}
         for k in pairs(yaml) do
           toplevel_keys[#toplevel_keys + 1] = k


hishamhm:
Since #4542 is not fixed yet, the credentials are _not_ re-encrypted on `db_import`. So I suspect that the current behavior of `next` for `db_export` (exporting them encrypted) is the one actually compatible with the current behavior of `db_import`. In other words, I think the roundtrip `db_export -> db_import` without this PR would actually work — did you give it a try?

kikito:
@hishamhm My tests where manual, but I have retried twice, to make sure. Here's the latest attempt:

Creation of the credential (notice how the initial password, `secret`, gets transformed into `3ebbff...`)

```
❯❯❯ http -b :8001/consumers username=peter
{
    "created_at": 1563985712,
    "custom_id": null,
    "id": "2b384366-a753-4c18-9950-7d12170c5a41",
    "tags": null,
    "username": "peter"
}
❯❯❯ http -b :8001/consumers/peter/basic-auth username=peter password=secret
{
    "consumer": {
        "id": "2b384366-a753-4c18-9950-7d12170c5a41"
    },
    "created_at": 1563985916,
    "id": "5d29a8b1-a091-465f-b69f-f0f891bd5089",
    "password": "3ebbff69c69f647dadf5e68ff5711df66015922f",
    "username": "peter"
}
```

Exporting:
```
❯❯❯ bin/kong config db_export kong.yml && cat kong.yml
_format_version: '1.1'
consumers:
- created_at: 1563985712
  id: 2b384366-a753-4c18-9950-7d12170c5a41
  username: peter
basicauth_credentials:
- created_at: 1563985916
  consumer: 2b384366-a753-4c18-9950-7d12170c5a41
  id: 5d29a8b1-a091-465f-b69f-f0f891bd5089
  password: 3ebbff69c69f647dadf5e68ff5711df66015922f
  username: peter
```

Emptying the database:
```
❯❯❯ bin/kong migrations reset -y
Resetting database...
Database successfully reset
❯❯❯ bin/kong migrations bootstrap
...
❯❯❯ http -b :8001/consumers
{
    "data": [],
    "next": null
}
```

Importing:
```
❯❯❯ bin/kong config db_import kong.yml
2019/07/24 18:34:11 [warn] the 'upstream_keepalive' configuration property is deprecated, use 'nginx_http_upstream_keepalive' instead
parse successful, beginning import
import successful
❯❯❯ http -b :8001/consumers/peter/basic-auth
{
    "data": [
        {
            "consumer": {
                "id": "2b384366-a753-4c18-9950-7d12170c5a41"
            },
            "created_at": 1563985916,
            "id": "5d29a8b1-a091-465f-b69f-f0f891bd5089",
            "password": "fbe6df64797e4074c3215470b684eb5e604bc221",
            "username": "peter"
        }
    ],
    "next": null
}
```

Notice how the password has changed from `3ebbff...` to `fbe6df...`. As far as I can tell the other fields have remained the same.

kikito:
(Obviously in order for this to work I had to deactivate the `db_export = false,` line from `kong/plugins/basic-auth/daos.lua`)

hishamhm:
@kikito Oh, you're right! `db_import` is encrypting the passwords... it is DB-less import that isn't. It still leaves us in a strange asymmetric state of affairs. I don't think disabling `db_export` for `basicauth_credentials` is the right thing to do, but it may be a stopgap measure to avoid confusion until #4542 is fixed.

kikito:
Yes, a more flexible solution would involve a way to import *both* encrypted and unencrypted passwords. Lots of options: we could use an environment variable ( `KONG_ENCRYPT_PASSWORDS=true kong config db_import ...`).

My personal preference would be that we never encrypt the `password` field on `db_import`, and instead we add a new "pseudo-field" called `unencrypted_password`, used only for imports and such:

```
basicauth_credentials:
- created_at: 1563985916
  consumer: 2b384366-a753-4c18-9950-7d12170c5a41
  id: 5d29a8b1-a091-465f-b69f-f0f891bd5089
  unencrypted_password: secret
  username: peter
```

PR #4831, opened 2019-07-24
tests(*) address grpc review

gszr:
### Summary

Minor changes pending from #4801.


spec/02-integration/04-admin_api/09-routes_routes_spec.lua
```diff
@@ -282,11 +282,6 @@ for _, strategy in helpers.each_strategy() do
               headers = { ["Content-Type"] = content_type }
             })
 
-            -- TODO: For some reason the body which arrives to the server is
```
gszr: @kikito This comment doesn't apply here (we are not passing the `methods` field), but I'm happy to look into this in other tests where the comment applies.

PR #4834, opened 2019-07-25
fix(balancer) added round-robin to algorithms enum

locao:
### Summary

- `"round-robin"` is now an available balancing algorithm, with the same behavior that was found in `algorithm="consistent"` and `hash_on="none"` combination before.
- Added migrations to set new `algorithm` field values, according to existent `hash_on` values.
- Added `shorthands` to the schema to keep backward compatibility in admin API and declarative config, avoiding inconsistent settings.

### DIFF
--- a/kong/db/migrations/core/005_120_to_130.lua
+++ b/kong/db/migrations/core/005_120_to_130.lua
@@ -55,6 +55,38 @@ return {
         -- Do nothing, accept existing state
       END$$;
     ]],
+    teardown = function(connector)
+      assert(connector:connect_migrations())
+      for upstream, err in connector:iterate([[SELECT * FROM "upstreams";]]) do
+        if err then
+          return nil, err
+        end
+
+        if type(upstream.algorithm) == "string" and #upstream.algorithm > 0 then
+          goto continue
+        end
+
+        local algorithm
+        if upstream.hash_on == "none" then
+          algorithm = "round-robin"
+        else
+          algorithm = "consistent-hashing"
+        end
+        local update_query = string.format([[
+          UPDATE "upstreams"
+          SET "algorithm" = '%s'
+          WHERE "id" = '%s';
+        ]], algorithm, upstream.id)
+
+        local _, err = connector:query(update_query)
+        if err then
+          return nil, err
+        end
+
+        ::continue::
+      end
+    end,
+
   },
 
   cassandra = {
@@ -81,5 +113,45 @@ return {
       ALTER TABLE services ADD client_certificate_id uuid;
       CREATE INDEX IF NOT EXISTS services_client_certificate_id_idx ON services(client_certificate_id);
     ]],
+    teardown = function(connector)
+      local cassandra = require "cassandra"
+      local coordinator = assert(connector:connect_migrations())
+
+      for rows, err in coordinator:iterate([[SELECT * FROM upstreams;]]) do
+        if err then
+          return nil, err
+        end
+
+        for i = 1, #rows do
+          local upstream = rows[i]
+          if type(upstream.algorithm) == "string" and #upstream.algorithm > 0 then
+            goto continue
+          end
+
+          local algorithm
+          if upstream.hash_on == "none" then
+            algorithm = "round-robin"
+          else
+            algorithm = "consistent-hashing"
+          end
+
+          local _, err = connector:query([[
+            UPDATE upstreams
+            SET algorithm = ?
+            WHERE id = ?
+          ]], {
+            algorithm,
+            cassandra.uuid(upstream.id),
+          })
+          if err then
+            return nil, err
+          end
+
+          ::continue::
+        end
+
+      end
+    end,
+
   },
 }
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -207,8 +207,9 @@ end
 local create_balancer
 do
   local balancer_types = {
-    ["consistent"] = require("resty.dns.balancer.ring"),
+    ["consistent-hashing"] = require("resty.dns.balancer.ring"),
     ["least-connections"] = require("resty.dns.balancer.least_connections"),
+    ["round-robin"] = require("resty.dns.balancer.ring"),
   }
 
   local create_healthchecker
@@ -677,7 +678,7 @@ end
 -- @return integer value or nil if there is no hash to calculate
 local create_hash = function(upstream, ctx)
   local hash_on = upstream.hash_on
-  if hash_on == "none" then
+  if hash_on == "none" or hash_on == nil or hash_on == ngx.null then
     return -- not hashing, exit fast
   end
 
--- a/kong/db/schema/entities/upstreams.lua
+++ b/kong/db/schema/entities/upstreams.lua
@@ -1,6 +1,7 @@
 local Schema = require "kong.db.schema"
 local typedefs = require "kong.db.schema.typedefs"
 local utils = require "kong.tools.utils"
+local null = ngx.null
 
 
 local validate_name = function(name)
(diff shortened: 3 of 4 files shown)


kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,139 @@
...
+    INSERT INTO upstreams(id, created_at, name, hash_on, hash_fallback,
+      hash_on_header, hash_fallback_header, hash_on_cookie,
+      hash_on_cookie_path, slots, healthchecks, tags, algorithm)
+    VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+  ]]
+  local delete_query = [[
+      DELETE FROM upstreams WHERE id = ?
+  ]]
+
+  local rows, err = connector:query(select_query)
```
hishamhm: even if you do `SELECT *`, this does not guarantee that all rows are returned in Cassandra. You have to use `iterate` (which returns pages) and then iterate the returned `rows` in an inner loop.

hishamhm: (see `plugins/acl/migrations/001_14_to_15.lua` for an example)

kong/runloop/balancer.lua
```diff
@@ -209,6 +209,7 @@ do
   local balancer_types = {
     ["consistent"] = require("resty.dns.balancer.ring"),
```
hishamhm: As discussed in https://github.com/Kong/kong/commit/3860151f2a2d9ef020d4c20496206aa5f922be9c#r33908240, let's also rename `consistent` to `consistent-hashing`

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,139 @@
...
+    end
+
+    local insert_query = string.format([[
+      INSERT INTO upstreams(id, created_at, name, hash_on, hash_fallback,
+        hash_on_header, hash_fallback_header, hash_on_cookie,
+        hash_on_cookie_path, slots, healthchecks, tags, algorithm)
+      VALUES(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
+    ]], id, created_at, name, hash_on, hash_fallback, hash_on_header,
+      hash_fallback_header, hash_on_cookie, hash_on_cookie_path, slots,
+      healthchecks, tags, algorithm)
```
hishamhm: Would it be possible to do the Postgres migration using UPDATE, to avoid the window between DELETE and INSERT?

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,139 @@
...
+    SELECT * FROM upstreams;
+  ]]
+
+  for row, err in connector:iterate(select_query) do
+    local algorithm
+    if row.hash_on == "none" then
+      algorithm = escape_literal("round-robin")
+    else
+      algorithm = escape_literal("consistent")
+    end
```
hishamhm: Migrations should be idempotent: if an `algorithm` field already has a value, it should not be overwritten (in that case the iteration can just `goto continue`).

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,139 @@
...
+
+
+local function add_upstream_algorithm_to_postgress(connector)
+  assert(connector:connect_migrations())
+
+  local select_query = [[
+    SELECT * FROM upstreams;
+  ]]
+
+  for row, err in connector:iterate(select_query) do
```
hishamhm: should test for `err` at the toop of the for-loop inner block

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,139 @@
...
+    return err
+  end
+
+  for _, upstream in ipairs(rows) do
+    local algorithm
+    if upstream.hash_on == "none" then
+      algorithm = "round-robin"
+    else
+      algorithm = "consistent"
+    end
```
hishamhm: same idempotency comment as above here

kong/db/migrations/core/005_120_to_130.lua
hishamhm: (I think it would also make the function a lot shorter, since you won't have to escape and store all fields)

hishamhm:
Also, FYI `spec/02-integration/05-proxy/10-balancer/02-least-connections_spec.lua` failed on Travis (see https://api.travis-ci.org/v3/job/563698245/log.txt )

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,139 @@
...
+    local slots = upstream.slots or cassandra.unset
+    local healthchecks = upstream.healthchecks or cassandra.unset
+    local tags = upstream.tags or cassandra.unset
+
+    local _, err = connector:query(delete_query, {id})
+    if err then
+      return err
+    end
+
+    local _, err = connector:query(insert_query, {id, created_at, name, hash_on,
```
hishamhm: same in Cassandra, we can UPDATE here.

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,88 @@
...
+    SELECT * FROM "upstreams";
+  ]]
+
+  for upstream, err in connector:iterate(select_query) do
+    if err then
+      return nil, err
+    end
+
+    if upstream.algorithm ~= nil and upstream.algorithm ~= ngx.null
+      and type(upstream.algorithm) == "string" and upstream.algorithm:len() > 0 then
```
thibaultcha: This expression can use much simpler conditions:
```lua
if type(upstream.algorithm) == "string" and #upstream.algorithm > 0 then

end
```

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,88 @@
+
+local function add_upstream_algorithm_to_postgress(connector)
```
thibaultcha: I don't see a reason for these to be local functions. It implies that the only teardown step for 1.2 to 1.3 migration is the addition of the `algorithm` field, and it could lead future migrations copy/pasted from this one to propagate this pattern. Better follow the pattern established in previous core migrations, and not name specific teardown steps in local functions.

Also, noting the (minor) typo: `_postgress`

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,88 @@
...
+
+  for rows, err in coordinator:iterate(select_query) do
+    if err then
+      return nil, err
+    end
+
+    for i = 1, #rows do
+      local upstream = rows[i]
+      if upstream.algorithm ~= nil and upstream.algorithm ~= cassandra.unset
+        and type(upstream.algorithm) == "string" and upstream.algorithm:len() > 0 then
```
thibaultcha: ditto: simpler conditions would produce the same behavior

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -1,3 +1,88 @@
...
+  local coordinator = assert(connector:connect_migrations())
+
+  local select_query = [[
+    SELECT * FROM upstreams;
+  ]]
+  local update_query = [[
+    UPDATE upstreams
+    SET algorithm = ?
+    WHERE id = ?
+  ]]
```
thibaultcha: A minor complaint, but throughout this file, we write the SQL/CQL queries in local variables, but I think it makes the code less readable, especially when the query is defined many lines above its use (e.g. here with `update_query`). We could just inline those queries in the `iterate()` and `query()` calls arguments.

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -55,6 +55,38 @@ return {
...
       END$$;
     ]],
+    teardown = function(connector)
+      assert(connector:connect_migrations())
+      for upstream, err in connector:iterate([[SELECT * FROM "upstreams";]]) do
+        if err then
+          return nil, err
+        end
+
+        if type(upstream.algorithm) == "string" and upstream.algorithm:len() > 0 then
```
thibaultcha: Let's avoid `:len()`, please (as highlighted in my previous review)

Tieske:
Nice!

PR #4839, opened 2019-07-26
fix(hmac-auth) correctly report http 1.0

### LINKED ISSUE #4833: (hmac-auth) http_version in request-line is wrong
when i use http1.0 to send request to kong, the request-line string before hmac is 
`GET / HTTP/1` instead of `GET / HTTP/1.0`, this bug make my client always been rejected

the problem is this line, `kong.request.get_http_version()` return `ngx.req.http_version()` 
https://github.com/Kong/kong/blob/88a530c68e09e0bde74a1904ee908054718f6195/kong/plugins/hmac-auth/access.lua#L133-L137

this is my test with `resty`
```
# /usr/local/openresty/bin/resty -e 'print(string.format("HTTP/%s", 1.0))'
HTTP/1
# /usr/local/openresty/bin/resty -e 'print(string.format("HTTP/%.01f", 1.0))'
HTTP/1.0
```

kikito:
Fixes #4833

### DIFF
--- a/kong/plugins/hmac-auth/access.lua
+++ b/kong/plugins/hmac-auth/access.lua
@@ -132,10 +132,11 @@ local function create_hash(request_uri, hmac_params)
     if not header_value then
       if header == "request-line" then
         -- request-line in hmac headers list
-        local request_line = fmt("%s %s HTTP/%s", kong.request.get_method(),
-                                 request_uri, kong.request.get_http_version())
+        local request_line = fmt("%s %s HTTP/%.01f",
+                                 kong.request.get_method(),
+                                 request_uri,
+                                 assert(kong.request.get_http_version()))
         signing_string = signing_string .. request_line
-
       else
         signing_string = signing_string .. header .. ":"
       end
--- a/spec/03-plugins/19-hmac-auth/03-access_spec.lua
+++ b/spec/03-plugins/19-hmac-auth/03-access_spec.lua
@@ -449,6 +449,31 @@ for _, strategy in helpers.each_strategy() do
         assert.res_status(200, res)
       end)
 
+      it("should encode http-1 requests as http/1.0", function()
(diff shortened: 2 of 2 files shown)


kong/plugins/hmac-auth/access.lua
```diff
@@ -132,8 +132,12 @@ local function create_hash(request_uri, hmac_params)
     if not header_value then
       if header == "request-line" then
         -- request-line in hmac headers list
+        local http_version = kong.request.get_http_version()
+        if http_version then
```
hishamhm: Can `kong.request.get_http_version` _not_ return an HTTP version? 

* If yes, then line 140 below would end up writing `HTTP/nil`
* If no, then you should be able to just change the format string to below to `"%s %s HTTP/%.01f"` and leave the PDK call inside the `fmt`

kikito: Well, we return nil ["for unrecognized values"](https://github.com/openresty/lua-nginx-module#ngxreqhttp_version) ... when I read that imagined someone sending `GET /foo HTTP/potato` instead of `GET /foo HTTP/1.0` over a telnet, but now that you mention it, I am not so sure such a request would reach this point. Not sure what "unrecognized values" mean then.

I've digged a bit more, and in the C source code there seems to be an allowance for it returning `nil`. I could not find a test or an explanation about when can it actually be nil:

https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_headers.c#L65

Leaving that aside, there's also the `http/nil`. Should we perhaps just error if the http version is not recognized?

kikito: After consulting @dndx, it seems that these "unrecognized values" are not for regular http access, but for things like timers. We can be pretty sure it will not raise a `nil`, and if it does, an `assert` would suffice.

PR #4855, opened 2019-07-30
feat(core) stash init_worker error and print them during request processing

thibaultcha:
Errors occurring during the `init_worker` phase can be critical to Kong's
proper runtime functioning. If any such error is encountered, it is
currently logged as a `CRIT` level warning, indicating an unhealthy
node. Unfortunately, those startup logs are often ignored by users,
causing corrupted worker processes to be left untouched, which can be
catastrophic in many cases.

This change ensures that all `init_worker` errors susceptible to produce
a corrupted worker process will be logged every time such a worker were
to accept a connection/request, with an `ALERT` level warning.

Note that a better initialization design would be to update the
lua-resty-worker-events library so that it can be initialized in the
`init` phase (and its polling later started in `init_worker`).
Doing so would allow us to initialize many of these singleton modules in
the `init` phase, and entirely prevent Kong from starting if any such
error is encountered, as is done today when checking for migrations or
initial database connectivity.

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -91,9 +91,10 @@ local ngx              = ngx
 local var              = ngx.var
 local header           = ngx.header
 local ngx_log          = ngx.log
+local ngx_ALERT        = ngx.ALERT
+local ngx_CRIT         = ngx.CRIT
 local ngx_ERR          = ngx.ERR
 local ngx_WARN         = ngx.WARN
-local ngx_CRIT         = ngx.CRIT
 local ngx_DEBUG        = ngx.DEBUG
 local ipairs           = ipairs
 local assert           = assert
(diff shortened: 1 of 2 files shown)


kong/init.lua
```diff
@@ -123,6 +124,47 @@ local declarative_entities
...
+    end
+
+    err = tostring(err)
+
+    if not init_worker_errors then
+      init_worker_errors = {}
+    end
+
+    table.insert(init_worker_errors, err)
+    init_worker_errors_str = table.concat(init_worker_errors, ", ")
```
dndx: Really minor suggestion: we can save this into `log_init_worker_errors` like:

```lua
    if not init_worker_errors or ngx.ctx[ctx_k] then
      return
    end

    if not init_worker_errors_str then
      init_worker_errors_str = table.concat(init_worker_errors, ", ")
    end
```

to avoid some unnecessary string creations. But a really minor point and certainly not a blocker.

thibaultcha: It's a tradeoff; the current approach saves the hot code path from executing `table.concat` entirely. Since `stash_init_worker_error` is only called during `init_worker`, we expect very few strings to be created at all, and they would be GC'ed. While the approach you suggested lazily calls `table.concat` on the proxy's hot code path, which is fine as well.

I'll leave the code as-is given both approaches are quite similar given the tradeoffs.

PR #4885, opened 2019-08-09
hotfix(schema) allow 'host' sub-string in headers

hbagdi:
### Summary

An exact match of `host` header should not be allowed in the headers
property of routes but a header with `host` as a sub-string in the key should be
allowed.

### Full changelog

* hotfix(schema) allow 'host' substrings in headers

### DIFF
--- a/spec/01-unit/01-db/01-schema/03-typedefs_spec.lua
+++ b/spec/01-unit/01-db/01-schema/03-typedefs_spec.lua
@@ -175,5 +175,15 @@ describe("typedefs", function()
     assert.falsy(Test:validate({ f = 123 }))
   end)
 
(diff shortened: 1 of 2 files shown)


spec/01-unit/01-db/01-schema/03-typedefs_spec.lua
```diff
@@ -175,5 +175,15 @@ describe("typedefs", function()
     assert.falsy(Test:validate({ f = 123 }))
   end)
 
+  it("headers", function()
```
thibaultcha: Mind making this test's name a bit more explicit? E.g. `"rejects 'host' but accepts 'host' substring"`?

PR #4887, opened 2019-08-09
 fix(pdk) response.exit to error when using table body with content-type

bungle:
### Summary

This contains more refinements to PDK's `kong.response.exit` for better gRPC support and compatibility with a current code and gRPC (a bit like best effort). Also explicitly errors on cases where explicitly outputting grpc content that we cannot encode.

* fix(pdk) remove the grpc_mode check, checking content types is enough. Also fixes warning that is logged when using this on admin api.
* fix(pdk) do not replace headers when specified with response.exit
* fix(pdk) response.exit to error when using table body with content-type.  Also logs warning when using table body that is not the common:
```
{
  message = "<string>"
}
```
and then basically ignores that body for better compatibility withcurrent Kong core and plugins that are not necessarily build for grpc in mind.

### DIFF
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -14,7 +14,6 @@
 
 local cjson = require "cjson.safe"
 local meta = require "kong.meta"
-local constants = require "kong.constants"
 local checks = require "kong.pdk.private.checks"
 local phase_checker = require "kong.pdk.private.phases"
 
@@ -35,7 +34,6 @@ local check_phase = phase_checker.check
 
 
 local PHASES = phase_checker.phases
-local GRPC_PROXY_MODES = constants.GRPC_PROXY_MODES
 
 
 local header_body_log = phase_checker.new(PHASES.header_filter,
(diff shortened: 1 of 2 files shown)


kong/pdk/response.lua
```diff
@@ -536,10 +532,16 @@ local function new(self, major_version)
     local json
     if type(body) == "table" then
       if is_grpc then
-        if type(body.message) == "string" then
+        if is_grpc_output then
```
gszr: Won't this catch part of the cases where `body.message` should be returned in the header - i.e., should we move the following `elseif` up?

bungle: @gszr we decided with @hishamhm that if there is response content type given for grpc output (either in exit headers orngx.header), we don't support this magic table. magic table is used only for this compatibility thing, e.g. currently there ia no such plugin that asks grpc response type but there are pkugins and core code tgat sends { message = <string> }.

bungle: is_grpc_output = send body in body

PR #4889, opened 2019-08-11
 fix(response-transformer) allow specify type for config value

### LINKED ISSUE #4880: Response-Transformer plugin problem: {“code”: -1} needed instead of {“code”, “-1”}
So I have been trying this plugin by reading this post: https://docs.konghq.com/hub/kong-inc/response-transformer/#examples
The problem I encountered is that I want to transform the response to something like {“code”: -1} and I tried “config.add.json”:[“code:-1”], but the response becomes {“code”, “-1”}, -1 turning out to be a string.
Does anyone have any idea on how to fix this?

tkbky:
### Summary

Response transformer is turning every value into string, which is not something expected. This change updates response transformer to convert values into its expected type accordingly.

### Issues resolved

Fix #4880

### DIFF
--- a/kong/plugins/response-transformer/body_transformer.lua
+++ b/kong/plugins/response-transformer/body_transformer.lua
@@ -19,12 +19,23 @@ local noop = function() end
 local _M = {}
 
 
-local function cast_value(value)
-  local num_v = tonumber(value)
-  if num_v == nil then
+local function toboolean(value)
+  if value == "true" then
+    return true
+  else
+    return false
+  end
+end
+
+
+local function cast_value(value, value_type)
+  if value_type == "number" then
+    return tonumber(value)
+  elseif value_type == "boolean" then
+    return toboolean(value)
+  else
     return value
   end
-  return num_v
 end
 
 
@@ -39,17 +50,16 @@ local function append_value(current_value, value)
   local current_value_type = type(current_value)
 
   if current_value_type  == "string" then
-    return {current_value, cast_value(value) }
+    return {current_value, value }
   end
 
   if current_value_type  == "table" then
-    insert(current_value, cast_value(value))
+    insert(current_value, value)
     return current_value
   end
 
-  return { cast_value(value) }
+  return { value }
 end
-
 
 local function iter(config_array)
   if type(config_array) ~= "table" then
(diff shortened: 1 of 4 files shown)


bungle:
@tkbky, perhaps we need to do this other way, as the same issue is with booleans. I don't like that we turn numbers and booleans to strings in a first place.

tkbky:
@bungle 
It seems like we're accepting those [configured values as string](https://github.com/Kong/kong/blob/master/kong/plugins/response-transformer/schema.lua). Unless you're saying the [original values from the response](https://github.com/Kong/kong/blob/master/kong/plugins/response-transformer/body_transformer.lua#L71), I don't think they are turned into string by the transformer.

Still feel that we need to convert the configured values to its appropriate type somehow, and I found out it could be done in the [iterator](https://github.com/Kong/kong/blob/master/kong/plugins/response-transformer/body_transformer.lua#L60).

I'm still very new to the code base, so I could have missed few things that might lead to neater solution.

bungle:
@tkbky I think the bug stars from here:
https://github.com/Kong/kong/blob/next/kong/plugins/response-transformer/body_transformer.lua#L84

I was thinking about changing this:

```lua
  for _, name, value in iter(conf.replace.json) do
    local v = cjson.encode(value)
    if v and sub(v, 1, 1) == [["]] and sub(v, -1, -1) == [["]] then
      v = gsub(sub(v, 2, -2), [[\"]], [["]]) -- To prevent having double encoded quotes
    end

    v = v and gsub(v, [[\/]], [[/]]) -- To prevent having double encoded slashes
    if json_body[name] and v then
      json_body[name] = v
    end
  end
  -- and similar changes to other cases.
  ...
```

to:

```lua
  for _, name, value in iter(conf.replace.json) do
    local v = cjson.encode(value)
    if vt == "string" then
      local vt = type(v)
      if v and sub(v, 1, 1) == [["]] and sub(v, -1, -1) == [["]] then
        v = gsub(sub(v, 2, -2), [[\"]], [["]]) -- To prevent having double encoded quotes
      end

      v = v and gsub(v, [[\/]], [[/]]) -- To prevent having double encoded slashes
     end
   
     if json_body[name] and v then
        json_body[name] = v
      end
  end
  -- and similar changes to other cases.
  ...
```

tkbky:
@bungle 
Hmm, but isn't `value` always be `string`?
e.g. the config pair is in this format: "name:value", and it's [matched here](https://github.com/Kong/kong/blob/next/kong/plugins/response-transformer/body_transformer.lua#L55)

bungle:
@tkbky, you are right, my bad, I held a wrong assumption.

hishamhm:
@tkbky Thanks for the PR! We have a couple of additional concerns with this PR, however. One is backward compatibility, since for users that have working setups where the data comes in and out as strings, they would have an unexpected change in behavior. Another one is whether all number-like strings should really be changed into numbers. Number conversion can be tricky, and things that don't look like numbers at first (like hexadecimal floats) that would be converted. If something like this was added, it would be safer to have a way to explicitly state the desired type.

tkbky:
@hishamhm 
yes, this change isn't backward compatible, and those are fair concerns. I think this might turned into a feature rather than just a simple fix, e.g. change of schema.

Let me know how would you like to proceed.

hishamhm:
> yes, this change isn't backward compatible, and those are fair concerns. I think this might turned into a feature rather than just a simple fix, e.g. change of schema.

@tkbky  _Adding_ to the schema allows it to be backwards compatible. While not ideal, I think a backwards compatible way of specifying this would be something like

```
{
   "json": ["foo:xxx", "bar:-1"],
   "json_types": ["string", "number"],
}
```

It's a bit ugly, but it's backwards compatible and solves the issue. How does that sound to you?

tkbky:
@hishamhm Yes, that sounds good to me.

tkbky:
@hishamhm I've made some changes but the build failed, can't seems to figure out why.

kong/plugins/response-transformer/body_transformer.lua
```diff
@@ -82,39 +101,58 @@ function _M.transform_json_body(conf, buffered_data)
...
     end
 
     v = v and gsub(v, [[\/]], [[/]]) -- To prevent having double encoded slashes
+
+    if conf.replace.json_types then
+      local v_type = conf.replace.json_types[i]
+      v = cast_value(v, v_type)
+    end
+
     if json_body[name] and v then
```
hishamhm: When `cast_value` converts `v` to boolean, this will be `false` and will not enter this branch. The fix is the following:

```suggestion
    if json_body[name] and v ~= nil then
```

(Note that this implicitly also means that when a conversion fails, the value stays unchanged (and sent as a string)).

Perhaps we need to add test cases for `true` and `false` values as well? It would have caught this.

hishamhm:
@tkbky The test suite failure seems unrelated (more likely a transient problem in the CI).

Thanks for the changes to the PR, they look good! Left a small comment there about a case to be handled.

kong/plugins/response-transformer/body_transformer.lua
tkbky: Thanks for the catch, I have made some changes as suggested and added some tests [in this commit](https://github.com/Kong/kong/pull/4889/commits/cad6b59c3824645c3d899f97ff5e145f6e9f15e8) to cover that.

tkbky:
@hishamhm Take a look?

hishamhm:
@tkbky Looks great! It's now merged, thank you!

(BTW, this got us thinking that it would be cool to have it in request-transformer as well... :) )

tkbky:
@hishamhm I'm happy to make a PR for that.

hishamhm:
@tkbky Great, thank you!!

PR #4892, opened 2019-08-12
feat/config-cache function

Tieske:
### Summary

adds a function to the PDK to generate a configuration cache table

### Full changelog

* new PDK function `kong.table.new_cache`
* added tests for PDK
* implemented the function on existing plugins

### DIFF
--- a/kong/pdk/table.lua
+++ b/kong/pdk/table.lua
@@ -2,6 +2,9 @@
 --
 -- @module kong.table
 
+local rawset = rawset
+local setmetatable = setmetatable
+local assert = assert
 
 local new_tab
 local clear_tab
@@ -55,6 +58,7 @@ end
 --- Merges the contents of two tables together, producing a new one.
 -- The entries of both tables are copied non-recursively to the new one.
 -- If both tables have the same key, the second one takes precedence.
+-- @function kong.table.merge
 -- @tparam table t1 The first table
 -- @tparam table t2 The second table
 -- @treturn table The (new) merged table
@@ -78,11 +82,65 @@ local function merge_tab(t1, t2)
 end
 
 
+--- Creates a new cache table (or memoize).
+-- A Cache table is typically used to convert configuration values once at
+-- load-time. To avoid expensive operations at run-time. The cache will
+-- have weak-keys, so the lifetime of the values is bound to the lifetime
+-- of the keys.
+--
+-- Whenever a key is not found on lookup, the `convert_func` will be invoked to
+-- generate the value.
+--
+-- @function kong.table.merge
+-- @tparam function convert_func the conversion function
+-- @treturn table The cache table
+-- @usage
+-- local config_cache = kong.table.new_cache(function(conf)
+--   -- assume we have a config table that has an array with strings
+--   -- but at runtime we'd like a reverse lookup table instead of iterating
+--   -- over the table on every request
+--   local hash = {}
+--   for i, v in ipairs(conf.array) do
+--     hash[value] = i
+--   end
+--   return {
+--     hash = hash,
+--     array = conf.array,
+--   }
+-- end)
+--
+-- function my_plugin.access(conf)
+--   conf = config_cache[conf]  -- generates it if not found
+--
+--   local value = ngx.req.get_headers()["My-Header"]
+--   if conf.hash[value] then  -- reverse-lookup instead of array traversal
+--     ngx.log(ngx.WARN, "the configured value was actually found")
+--   end
+-- end
+local function new_cache(convert_func)
+  return setmetatable({}, {
+    __mode = "k",
+    __index = function(self, key)
+      if type(key) ~= "table" then
+        error(("the key must be a table, got '%s'"):format(type(key)), 2)
+      end
+
+      local value = assert(convert_func(key))
+      assert(value ~= key, "the value cannot be the same table") -- memory leak
+
+      rawset(self, key, value)
+      return value
+    end,
+  })
+end
+
+
 local function new(self)
   return {
     new = new_tab,
     clear = clear_tab,
     merge = merge_tab,
+    new_cache = new_cache,
   }
 end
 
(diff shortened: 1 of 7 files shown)


hishamhm:
@Tieske given LuaJIT's issues with weak tables (the lack of ephemeron table support added in Lua 5.2) which has caused us some serious bugs more than once, I am _very_ wary about hiding weak-tables into a utility function like that. Plugin authors using that function indiscriminately would bump into the same issues. I think people need to be aware of the muddy waters they're stepping into when they deal with setmetatable applying weak tables.

Tieske:
@hishamhm I see the issue, yet we have this type of code in many places. We're not taking the concern away by not putting it in a PDK function.

That said, there is an alternative solution; getting away from configuration table being owned by the cache, and making them owned by the plugin itself.

This is just semantic, but changes the responsibilities. Currently the stance is that the `conf` table a plugin gets is owned by the cache and hence is (by convention, not enforced) read-only. So we create a weak table with the `conf` as key and the alternate config table as value.
If the stance would change that the plugin owns the `conf` table, then we could update the table in-place, and forego on the extra caching table.

hishamhm:
> We're not taking the concern away by not putting it in a PDK function.

Where `setmetatable` is used, we're making the concern explicit.

> That said, there is an alternative solution; getting away from configuration table being owned by the cache, and making them owned by the plugin itself.

What you really want is a (caching) storage space that is private to your plugin and has a lifetime that spans more than one request. It sounds like a missing piece of the PDK, since what we currently have is:

| Scope \ Lifetime | single request | multiple requests (caching) |
|--------------------------|-------------------------|-------------------------|
| single plugin | kong.ctx.plugin | ?
| all plugins | kong.ctx.shared | kong.cache

The fact that this storage space is keyed by `conf` is an implementation detail, that IMO should be taken care of by the PDK. I personally think this would be a welcome addition, but the API for that needs to be carefully designed and I think it shouldn't use the approach of this PR.

Tieske:
regarding the "footgun"; the specific implementation is using single references. I presume the issues you're mentioning are the ones with the balancer/healthchecks we had? That was a spaghetti of references. Totally unlike this usecase.

> What you really want is a (caching) storage space that is private to your plugin and has a lifetime that spans more than one request.

No. Definitions are important here, caching that is private to the plugin is already available as module level upvalues.

So the missing piece is a 'storage space' private to the '_**plugin instance**_' (which effectively is a 1-on-1 with the plugin `conf` table)

The two options I listed in this PR are the ones I know of;

1. weak-table cache based on `conf`
2. changing ownership definition of `conf`

This does not cater for all cases, because sometimes the caches are nested (see the ACL plugin for example).

I'm wondering whether extra PDK magic as you're proposing can beat the simple nr. 2 solution.

There is one more solution, which is adding a Lua-land callback to "deserialize" the plugin conf table when it is moved into the Lua-land cache. iirc @thibaultcha implemented the callback in `mlcache` but we're not using it here. (note that this solves the common usecase, but not nested caching like ACL uses)

thibaultcha:
The matrix for cache lifetimes we had in mind when developing the PDK was slightly larger:

| Lifetime                 | single request      | multiple requests  |
|--------------------------|-------------------------|--------------------------|
| single plugin, single config         | kong.ctx.plugin     | ? (this PR)
| single plugin, all configs             | N/A                       | ? (module upvalue)
| all plugins (all configs) | kong.ctx.shared | kong.cache

Our internal PDK TODO list mentions the need for a caching layer spanning multiple requests for a given plugin (what this PR proposes to solve). I recall discussions on the topic already and some ideas around naming, but cannot find any written trace at the moment.

To move this forward, I would suggest:

1. Making it always available to the user as a "magic table" similar to `kong.ctx`.
2. Finding an appropriate name.

A few suggestions:

- `kong.plugin` (mirroring `kong.ctx.plugin`)
- `kong.plugin_cache`
- Renaming `kong.ctx` to `kong.req_ctx` and introduce `kong.plugin_ctx` (this would have been nice to have, but it is too late now imho)

hishamhm:
> Totally unlike this usecase.

I agree. But when you give a super-general construct such as `kong.table.new_cache` to the user, you are giving the user the building block to construct spaghettis like those. That's what I want to avoid. Simplifying plugin development and helping to prevent footguns is the reason the PDK exists, otherwise we would have kept to the more powerful but lower-level OpenResty APIs.

> 1. Making it always available to the user as a "magic table" similar to kong.ctx.

+1 to that.

> 2. Finding an appropriate name.

From the suggestions, I like the sound of `kong.plugin_cache`. It's easy to explain as a generic key-value cache that a plugin instance has at its disposal.

The underlying implementation could be similar to this PR (a strong table stored in a a weak-key table keyed by the plugin `conf` — the main advantage over this PR is that the weak table itself is inaccessible to the plugin author, hence no chance of backreferences and unexpected weakref issues).

Tieske:
Can we do it in a way that we recombine the plugin code and the conf table to make a proper object? because the distinction we now have (that the `conf` table is actually sort-of the "instance" of a plugin) is a weird one.

If we take `conf` and add the plugin table as the metatable index. Such that `conf:access()` would invoke the `access` handler. Then we would make it more transparent imo.

Not sure it can be done backward compatible, but for 2.0 it would be nice.

Another approach could be to add a handler to the plugins, eg. `unpack_config(conf)`, and if that method is provided, it will be invoked once, weak-cached, and then the returned `conf` table will be passed to the other handlers. (similar to this PR, but all behind the scenes)

hishamhm:
@Tieske there's definitely room for further design discussion on how to approach this, and yes, we can consider even more alternative approaches for Kong 2.0 (keeping in mind the possible disruption to our plugin ecosystem and weighing the trade-offs) — but it's definitely the time to be discussing this.

In any case, I don't think we will be following the approach of this particular PR. Should we close this one?

Tieske:
Yep, seems right for now

Tieske:
> we can consider even more alternative approaches for Kong 2.0

@hishamhm seems we missed the mark. Is it time to revive this discussion, if so how/where?

PR #4894, opened 2019-08-12
feat(reports) include traffic entry types

kikito:
Previously the anonymous reports ping only included a single counter for the requests the server did:

```
requests=100
```

With this PR, it also differentiates between types:
```
requests=100;http_reqs=10;https_reqs=0;grpc_reqs=20;grpcs_reqs=0;ws_reqs=70;wss_reqs=0
```

### DIFF
--- a/kong/reports.lua
+++ b/kong/reports.lua
@@ -17,8 +17,7 @@ local sub = string.sub
 
 local PING_INTERVAL = 3600
 local PING_KEY = "events:reports"
-local BUFFERED_REQUESTS_COUNT_KEYS = "events:requests"
-
+local REQUEST_COUNT_KEY       = "events:requests"
 
 local _buffer = {}
 local _ping_infos = {}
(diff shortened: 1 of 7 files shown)


kong/reports.lua
```diff
@@ -150,6 +156,83 @@ local function create_timer(...)
 end
 
 
+local function get_counter(key)
+  local count, err = kong_dict:get(key)
+  if err then
+    log(WARN, "could not get " .. key .. " from 'kong' shm: ", err)
```
thibaultcha: Please use the variadic arguments form instead of Lua-land concatenations

kong/reports.lua
```diff
@@ -150,6 +156,83 @@ local function create_timer(...)
...
+  end
+  return count or 0
+end
+
+
+local function reset_counter(key, amount)
+  local ok, err = kong_dict:incr(key, -amount, amount)
```
thibaultcha: Better use `:safe_set(key, 0)` to avoid race-conditions in between the previous `get()` and this `incr()` step.

kikito: It puzzled me a bit too, but I think we *want* to preserve these "race conditions".

This is my understanding of what's happening now:

```
                Flow of Time
                    ||| 
                    VVV 

      Worker A       |     Worker B
                     |
                     |
get_counter -> 100   |
                     |  <log phase> incr_counter(1) -> 101
                     |
reset_counter(-100)  |
============================================
                     
Final counter value: 1 (correct, the incr_counter from Worker B was preserved)
```

Now with what you propose:
```
                Flow of Time
                    ||| 
                    VVV 

      Worker A       |     Worker B
                     |
                     |
get_counter -> 100   |
                     |  <log phase> incr_counter(1) -> 101
                     |
safe_set(0)          |
============================================
                     
Final counter value: 0 (incorrect, the incr_counter from Worker B was lost)
```

kikito:
@james-callahan I believe that's not possible currently: you would have to be able to increase a shared dict in stream mode and then read it in http mode.

kong/reports.lua
thibaultcha: @kikito You are correct! My bad. I should know, this was my implementation -_- https://github.com/Kong/kong/commit/2ace69179fba2b3c193bd3cfb73f4102514ab306#diff-10418ffd13cdf2bc868b10cce41a8781R169

kikito: I have added the explanation as a comment in the code. Hopefully it will help whomever visits this code in another 3 years :)

PR #4895, opened 2019-08-12
hotfix(conf) allow injecting the `ssl_protocols` directive

thibaultcha:
This directive is provided as an example for dynamically-injected NGINX
directives in the `kong.conf.default` file. To avoid potential users
following the example and configuring such a directive from getting
confused, we now port this directive to a default injected NGINX
directive.
Customizing the TLS protocols being a common ask, this is likely
avoiding us many future support tickets.

This directive is now always included even when no listener is
configured with the `ssl` option.


hishamhm:
@thibaultcha does this mean we need to include a section on modifying custom templates in UPGRADE.md for 1.3.0?

thibaultcha:
@hishamhm There was already the need for such a section I believe (given changes to support the gRPC routing, but also to the `upstream_keepalive` directive. This PR means that a new entry needs to be added to mention the ability to configure such supported TLS protocols (similar to existing 1.3 Changelog entries on upstream keepalive configurations).

PR #4897, opened 2019-08-13
feat(conf) add router_update_frequency configuration option

### LINKED ISSUE #4872: router_consistency job period
router_consistency=eventual    : the router will be rebuilt asynchronously via a recurring background job running every second inside of each worker.

hello,  i need  every 10 second to run this job,  please add the job period config to kong.conf

- Kong version (1.2.0)

bungle:
### Summary

Prior this change the router and plugins where checked for changes every second (starting two timers per second on each worker). The issue #4872 reported some issues that frequent updates may cause on proxying performance and suggested we allow configuring the value. This PR adds `router_update_frequency` configuration option, which still default to that same once per second. This is mostly usable with `router_consistency=eventual`. But it will affect `strict` too (but raising this value with `strict` will make it more likely that the routers are rebuild on request time (making a timers bit useless, though it may be ok with slow traffic scenarios).

### Issues resolved

Fix #4872

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -842,6 +842,15 @@
                                  # for a short period of time after Routes and
                                  # Services updates.
 
+#router_update_frequency = 1     # Defines how often the router changes are
+                                 # checked with a background job. When a change
+                                 # is detected, a new router will be built. By
+                                 # default we check for changes every second.
+                                 # Raising this value will decrease the load on
+                                 # database servers and result in less jitter
+                                 # in proxy latency, with downside of longer
+                                 # converge time for router updates.
+
 #------------------------------------------------------------------------------
 # DEVELOPMENT & MISCELLANEOUS
 #------------------------------------------------------------------------------
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -877,7 +877,9 @@ return {
         balancer.init()
       end)
 
-      timer_every(1, function(premature)
+      local router_update_frequency = kong.configuration.router_update_frequency or 1
+
+      timer_every(router_update_frequency, function(premature)
         if premature then
           return
         end
@@ -892,7 +894,7 @@ return {
         end
       end)
 
-      timer_every(1, function(premature)
+      timer_every(router_update_frequency, function(premature)
         if premature then
           return
         end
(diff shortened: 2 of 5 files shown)


kong/runloop/handler.lua
```diff
@@ -813,11 +813,15 @@ return {
...
 
-      timer_every(1, function(premature)
+      local router_update_frequency = kong.configuration.router_update_frequency or 1
+
+      timer_every(router_update_frequency, function(premature)
         if premature then
           return
         end
 
+        kong.log.err("updating router #", ngx.worker.id())
```
dndx: This info does not seems to warrant an `error`.

kong.conf.default
```diff
@@ -829,6 +829,15 @@
                                  # for a short period of time after Routes and
                                  # Services updates.
 
+#router_update_frequency = 1     # Defines how often the router changes are
+                                 # checked with a background job. When a change
+                                 # is detected, a new router will be built. By
+                                 # default we check for changes every second.
+                                 # Raising this value may allow more requests
```
dndx: Instead of "allow more requests to flow", how about "Raising this value will decrease the load on database servers and results in less jitter in proxy latency, with downside of longer waits for router update to converge.".

kong/runloop/handler.lua
bungle: oh, a leftover.

bungle:
> It's fine to add a new configuration property for this use-case imho. However we should also ensure that this value always be `>= 0`.

I am sure you meant `> 0`.

bungle:
@thibaultcha,

I added a check and some tests for it.

kong.conf.default
```diff
@@ -842,6 +842,15 @@
                                  # for a short period of time after Routes and
                                  # Services updates.
 
+#router_update_frequency = 1     # Defines how often the router changes are
+                                 # checked with a background job. When a change
+                                 # is detected, a new router will be built. By
+                                 # default we check for changes every second.
+                                 # Raising this value will decrease the load on
+                                 # database servers and results in less jitter
```
thibaultcha: typo: result

thibaultcha:
Commit scope should be `feat(conf)` btw

PR #4903, opened 2019-08-13
perf(proxy) cache plugins route and service ids

### LINKED ISSUE #4807: Kong 1.2.1 returns 502 when Postgres is unreachable
### Summary

I'm running chaos test on Kong V1.2.1 with a Postgres server.
When I make the Postgres server unreachable by modifying iptables, some request return 502(`$status` in `access.log` is `000`)
Here is a suspicious error log:
```
2019/07/17 07:07:10 [error] 21424#0: *12422286 [lua] plugins_iterator.lua:97: load_configuration(): failed to get from node cache: [postgres] timeout, client: 222.72.156.190, server: kong, request: "GET / HTTP/1.1", ...."
```

I looked into the code base, and found that this error happens when the related plugin cache is not ready. Maybe it's introduced by this [commit](https://github.com/Kong/kong/commit/e1ad79dbe8ceef3da91783296ccb3f8d451632c2). Is that okay to skip plugin loading when db is unreachable?

### Steps To Reproduce

1. Prepare a Kong server with 3 routes `route-a`, `route-b`, `route-c`.  Enable a plugin of `route-a`
```
curl -i -X POST \
  --url http://localhost:8001/services/ \
  --data 'name=example-service' \
  --data 'url=http://mockbin.org'

curl -i -X POST \
  --url http://localhost:8001/services/example-service/routes \
  --data 'hosts[]=routea.com' \
  --data 'name=routea'

curl -i -X POST \
  --url http://localhost:8001/routes/routea/plugins \
  --data 'name=prometheus'

curl -i -X POST \
  --url http://localhost:8001/services/example-service/routes \
  --data 'hosts[]=routeb.com'

curl -i -X POST \
  --url http://localhost:8001/services/example-service/routes \
  --data 'hosts[]=routec.com'
```
2. Restart Kong maybe.
3. Access `route-a` to warm up plugin cache
```
curl localhost:8000 -H 'Host: routea.com' -I
curl localhost:8000 -H 'Host: routeb.com' -I
```
4. Make the Postgres unreachable by `sudo iptables -A OUTPUT -j DROP -d $pg_ip -p tcp --dport 5432`
5. Request `route-a`, `route-b` and `route-c`. The response of `route-c` should have error. And similar error log can be found.
```
curl localhost:8000 -H 'Host: routec.com' -I
curl: (52) Empty reply from server
```
### Additional Details & Logs

- Kong version 1.2.1
- Operating system
```Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.1 LTS
Release:	16.04
Codename:	xenial
```
- Kong configuration
```
db_update_frequency = 60
dns_stale_ttl = 300
-- other configurations are just default values
```

bungle:
### Summary

At the moment plugins make Kong not to work very well when database connection is not available. Even when we do pre-warming, because we don't do negative cache pre-warming. That in general results a DB call.

Previous attempt to fix this was to load all the plugins in memory, but that raised concerns of consumers and plugins applied to them. This PR changes plugin iterator to cache route and service ids. Sure there can be a lot of routes and services too, but those are the entities that we already cache quite extensively.

It allows me to:

1. start kong
2. stop postgres
3. proxy all those requests that were mentioned in #4807

### Issues resolved

Fix #4807

### DIFF
--- a/kong/runloop/plugins_iterator.lua
+++ b/kong/runloop/plugins_iterator.lua
@@ -143,29 +143,40 @@ local function load_configuration_through_combos(ctx, combos, plugin)
   local  service_id = service  and  service.id or nil
   local consumer_id = consumer and consumer.id or nil
 
-  if route_id and service_id and consumer_id and combos[COMBO_RSC] then
-    plugin_configuration = load_configuration(ctx, name, route_id, service_id, consumer_id)
+  if route_id and service_id and consumer_id and combos[COMBO_RSC]
+    and combos.both[route_id] == service_id
+  then
+    plugin_configuration = load_configuration(ctx, name, route_id, service_id,
+                                              consumer_id)
     if plugin_configuration then
       return plugin_configuration
     end
   end
 
-  if route_id and consumer_id and combos[COMBO_RC] then
-    plugin_configuration = load_configuration(ctx, name, route_id, nil, consumer_id)
+  if route_id and consumer_id and combos[COMBO_RC]
+    and combos.routes[route_id]
+  then
+    plugin_configuration = load_configuration(ctx, name, route_id, nil,
+                                              consumer_id)
     if plugin_configuration then
       return plugin_configuration
     end
   end
 
-  if service_id and consumer_id and combos[COMBO_SC] then
-    plugin_configuration = load_configuration(ctx, name, nil, service_id, consumer_id)
+  if service_id and consumer_id and combos[COMBO_SC]
+    and combos.services[service_id]
+  then
+    plugin_configuration = load_configuration(ctx, name, nil, service_id,
+                                              consumer_id)
     if plugin_configuration then
       return plugin_configuration
     end
   end
 
-  if route_id and service_id and combos[COMBO_RS] then
-    plugin_configuration = load_configuration(ctx, name, route_id, service_id, nil)
+  if route_id and service_id and combos[COMBO_RS]
+    and combos.both[route_id] == service_id
+  then
+    plugin_configuration = load_configuration(ctx, name, route_id, service_id)
     if plugin_configuration then
       return plugin_configuration
     end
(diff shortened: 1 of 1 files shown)


bungle:
We can refine the code, e.g. do these id hash-table lookups after combos lookups for example.

bungle:
> We can refine the code, e.g. do these id hash-table lookups after combos lookups for example.

I rebased the code, and made it more readable.

kong/runloop/plugins_iterator.lua
```diff
@@ -157,28 +157,28 @@ local function load_configuration_through_combos(ctx, combos, plugin)
   local  service_id = service  and  service.id or nil
   local consumer_id = consumer and consumer.id or nil
 
-  if route_id and service_id and consumer_id and combos[COMBO_RSC] then
+  if route_id and service_id and consumer_id and combos[COMBO_RSC] and combos.both[route_id] == service_id then
```
thibaultcha: Could we limit the line length appropriately in those conditions now, please?

kong/runloop/plugins_iterator.lua
```diff
@@ -321,14 +321,28 @@ function PluginsIterator.new(version)
...
+      combos[name].routes   = combos[name].routes   or {}
+      combos[name].services = combos[name].services or {}
+
+      combos[name][combo_key] = true
+
+      if plugin.route and plugin.service then
+        combos[name].both[plugin.route.id] = plugin.service.id
+      elseif plugin.route then
+        combos[name].routes[plugin.route.id] = true
+      elseif plugin.service then
```
thibaultcha: style: ideally we should be adding line breaks before the `elseif` branches, as usual.

kong/runloop/plugins_iterator.lua
```diff
@@ -321,14 +321,28 @@ function PluginsIterator.new(version)
...
                       + (plugin.service  and 2 or 0)
                       + (plugin.consumer and 4 or 0)
 
-      combos[plugin.name] = combos[plugin.name] or {}
-      combos[plugin.name][combo_key] = true
+      combos[name]          = combos[name]          or {}
+      combos[name].both     = combos[name].both     or {}
+      combos[name].routes   = combos[name].routes   or {}
+      combos[name].services = combos[name].services or {}
+
```
thibaultcha: Is the reason behind not including `consumers` in this new heuristic because of the high-cardinality of the Consumer ids attribute? Because as the PR stands, the same issue with DB connectivity would occur when the request is authenticated, but the plugin currently being evaluated by the runloop doesn't apply to the authenticated Consumer.

bungle: @thibaultcha, yes, the high cardinality was the reason. In one proposal at one point I proposed that we just load all the plugin configurations in memory, but that was not accepted at that time because of the high-cardinality. Of course we were then speaking about caching the whole plugin configurations, and not just ids, as what this PR does. So with caching just ids the memory pressure is not that big, but well, high-cardinality of consumers may still be seen as a problem. Perhaps we can leave that decision later, and first release without, and rethink about caching the plugin consumer.ids later (just thinking that perhaps moving one step at time is safer, and we may get feedback on it). I can create a backlog item on that. One obviously good thing about caching the ids is that it makes negative caching not needed, so if we add consumer ids to this cache too, we will not need negative caching with plugins at all.  If you feel strongly about it, let me know, the caching of consumer ids is rather easy to add.

kong/runloop/plugins_iterator.lua
bungle: line breaks added.

kong/runloop/plugins_iterator.lua
bungle: line length limited to max 80 chars.

bungle:
> not only for performance reasons, but resiliency concerns most of all

Yes, I agree. On internal discussion I have referred it for resiliency improvements. I don't think this adds any significant perf improvements.

PR #4907, opened 2019-08-13
fix(balancer) target entities default to port 80

locao:
When target entities don't have a port `create_balancer` was not being able to parse them and was trying to use empty values for hostname and port to add a new host to dns client.

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -134,7 +134,12 @@ do
       -- split `target` field into `name` and `port`
       local port
       target.name, port = string.match(target.target, "^(.-):(%d+)$")
-      target.port = tonumber(port)
+      if target.name and port then
+        target.port = tonumber(port)
+      else
+        target.name = target.target
+        target.port = 80
+      end
     end
 
     return target_history
--- a/spec/02-integration/04-admin_api/15-off_spec.lua
+++ b/spec/02-integration/04-admin_api/15-off_spec.lua
@@ -440,6 +440,49 @@ describe("Admin API #off", function()
       assert.response(res).has.status(201)
     end)
 
+    it("can reload upstreams (regression test)", function()
+      local config = [[
+        _format_version: "1.1"
+        services:
+        - host: foo
+          routes:
+          - paths:
+            - "/"
+        upstreams:
+        - name: "foo"
+          targets:
+          - target: 10.20.30.40
+      ]]
+      local res = assert(client:send {
+        method = "POST",
+        path = "/config",
+        body = {
+          config = config,
+        },
+        headers = {
+          ["Content-Type"] = "application/json"
+        }
+      })
+
+      assert.response(res).has.status(201)
+
+      client:close()
+      client = helpers.admin_client()
+
+      local res = assert(client:send {
+        method = "POST",
+        path = "/config",
+        body = {
+          config = config,
+        },
+        headers = {
+          ["Content-Type"] = "application/json"
+        }
+      })
+
+      assert.response(res).has.status(201)
+    end)
+
     it("returns 304 if checking hash and configuration is identical", function()
       local res = assert(client:send {
         method = "POST",


hbagdi:
Thought/question: should this not default to the port in the corresponding service definition instead?

choaryzhang:
hi,I am an user of kong ,from China~

Tieske:
> Thought/question: should this not default to the port in the corresponding service definition instead?

@hbagdi no, this is by design. Kong trusts the DNS server, so in cases of a balancer or SRV record, the port in the `service` will be overridden by the data from the DNS

Tieske:
@locao I think we also need a more specific test. The regression test added showed the behaviour but only expeosed it because it accidentally had no port. The error can probably be triggered in a DB scenario as well. Correct? or should this test just be renamed/relocated?

kong/runloop/balancer.lua
```diff
@@ -134,7 +134,12 @@ do
       -- split `target` field into `name` and `port`
       local port
       target.name, port = string.match(target.target, "^(.-):(%d+)$")
-      target.port = tonumber(port)
+      if target.name and port then
+        target.port = tonumber(port)
```
bungle: what if `port` is greater than `65535`?

bungle:
> @hbagdi no, this is by design. Kong trusts the DNS server, so in cases of a balancer or SRV record, the port in the `service` will be overridden by the data from the DNS

Sure, but if it is not in DNS, should it fallback to service?

kong/runloop/balancer.lua
Tieske: isn't that caught by schema validations?

locao: Yes, schema uses `kong.tools.utils.normalize_ip` to validate IP and port.

locao:
> The error can probably be triggered in a DB scenario as well. Correct? or should this test just be renamed/relocated?

You are right, I'll write a test for that. Even if the error wasn't triggered, the tests should cover this case.

hishamhm:
> Thought/question: should this not default to the port in the corresponding service definition instead?

Upstreams and Targets are independent of Service. You can have multiple Services with different `port` values that have `host` values that point to the same Upstream.

Sure, those could reuse the Service's `port` when proxying... but what about active healthchecks? That is an operation on Targets that does not involve Services at all.

hishamhm:
https://github.com/Kong/kong/blob/1.2.1/kong/db/dao/targets.lua#L15

Ah! I _knew_ we had code somewhere which picked an arbitrary default port for targets somewhere.

So that answers my question: if a raw IP without a port is given for a Target, the default is 8000. (And generally people shouldn't be using raw IPs without ports in Targets anyway.)

hishamhm:
Ok, so now things start to make sense.

There shouldn't be targets without ports in the database _at all_.

The reason why `kong.runloop.balancer` is seeing targets without ports in DB-less mode is because the default port is being set at `Targets:insert`, in the DAO for targets.

But DB-less `load_into_cache` does not use the DAO; it just schema-validates the input and inserts it straight into the cache.

This is the same kind of problem we observed with DB-less and `basicauth_credentials` and the full proper solution is to have a transformations API added to the DAO (it's already in the plans, we should be getting that in 1.4.0).

Another thing I observe, though is that `DEFAULT_PORT` is only used by `format_target`, which is only used by `Targets:insert`... meaning that `update` and `upsert` are probably going to be able to set a `target` value without a port (i.e. making the problem reproducible in DB mode by using PATCH or POST). Again, the solution is to have the default port transformatino done in a future `on_input` transformation method that is used automatically by `insert`/`update`/`upsert`/DB-less.

For now, I think a better stopgap solution is to modify `kong.db.dao.targets` to make it call `format_target` in more places (such as `select_by_upstream_raw`, and I guess `post_health` as well?), then `kong.runloop.balancer` won't need to be modified at all.

hbagdi:
> So that answers my question: if a raw IP without a port is given for a Target, the default is 8000. (And generally people shouldn't be using raw IPs without ports in Targets anyway.)

I had vaguely recalled this behavior before my comment but I thought my memory was faulty.
@hishamhm, thank you for connecting all the dots here!

locao:
Made the changes in targets DAO (#4920), as it should be, so I'm closing this PR.

PR #4908, opened 2019-08-13
tests(proxy) fix flappy tests caused by instability in DNS resolving

dndx:
This should eliminate the `Status expected: (number) 504 Status received: (number) 503 Body: (string) '{"message":"name resolution failed"}` failure we have been seeing recently.

Note on the selection of IP address. It is just one of the RFC 1918 address that is likely not being used. This may or may not work depends on the network environment but as far as Travis, Docker, VMWare or VirtualBox are concerned, it works well (and much better than the domain we had in there).

### DIFF
--- a/spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
+++ b/spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
@@ -588,21 +588,21 @@ for _, strategy in helpers.each_strategy() do
 
         do
           -- service to mock HTTP 504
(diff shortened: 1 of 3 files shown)


dndx:
Ran the integration tests 6+ more times, no more flakiness observed 🚀

kikito:
Force-pushed a small change: instead of using `10.255.255.255`, I put it on helpers so everyone can use `helpers.blackhole_host` instead.

PR #4911, opened 2019-08-14
tests(proxy) mark integration test as #flaky

kikito:


bungle:
Seems like travis did not pass more reliably :-(.

kikito:
I'm going to close this down. I didn't notice that Datong's fix for this was done against `master`, not against `next`.

PR #4921, opened 2019-08-15
feat(jwt) add none algorithm

Ray-Eldath:
### Summary

Add none algorithm described in RFC 7515 to the JWT plugin.

see: 
 - **https://discuss.konghq.com/t/none-algorithm-for-jwt-plugin/4212**
 - https://tools.ietf.org/html/rfc7515#appendix-A.5

### Full changelog

* Add none algorithm to the JWT plugin.
* Add a related test.

### Issues resolved
 
N/A


spec/03-plugins/16-jwt/01-jwt_parser_spec.lua
```diff
@@ -31,9 +31,9 @@ describe("Plugin: jwt (parser)", function()
     end)
     it("should properly encode using HS384", function()
       local token = jwt_parser.encode({
+        sub   = "1234567890",
         name  = "John Doe",
-        admin = true,
-        sub   = "1234567890"
+        admin = true
```
Tieske: these changes are unrelated, please revert. Also the ones below at line 48

Ray-Eldath:
Command addressed, please take a look.

kong/plugins/jwt/handler.lua
```diff
@@ -175,8 +175,9 @@ local function do_authentication(conf)
     return false, {status = 401, message = "Invalid algorithm"}
   end
 
-  local jwt_secret_value = algorithm ~= nil and algorithm:sub(1, 2) == "HS" and
-                           jwt_secret.secret or jwt_secret.rsa_public_key
+  local jwt_secret_value = algorithm ~= nil and algorithm == "none" or
+                           algorithm ~= nil and algorithm:sub(1, 2) == "HS" and jwt_secret.secret or
+                           jwt_secret.rsa_public_key
```
Tieske: this is becoming very hard to read. Maybe we can change this into some if-then statements to make it easier to read/maintain

p0pr0ck5:
The `none` algorithm has some pretty substantial security implications; this is a well-known aspect of the JWT standard. https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/ is a pretty good write on this; are we sure we want to support this functionality?

Tieske:
@p0pr0ck5 see the kongnation post in the op. I think it makes sense, but warrants some clear warnings in the docs.

Ray-Eldath:
Edited. Valid now?
Tests all pasted on my fork, in here it's pending :-)

Ray-Eldath:
is this valid now?

Tieske:
@bungle mind having a look and merge if ok?

p0pr0ck5:
I'm still not sure this makes sense. From what I can gather (and apologies in advance if I'm missing something), this potentially allows a malicious actor to pass a JWT token to Kong that doesn't contain a digital signature, and the user can arbitrarily choose to have not Kong validate the token based on the contents of the request. That seems like a fairly major design flaw, given that operators of a Kong cluster would have no way to not allow unsigned JWT tokens to be sent.

Regardless of the transport mechanism, the `none` mechanism doesn't really seem to have any benefit. The first full page of results for a Google search on "JWT none signature" involves security issues with the `none` signature. @Ray-Eldath is there a specific use case you're hoping to solve with this?

At the very least, this plugin should contain a config option to allow `none` JWTs to be processed (and that option should be `false` by default).

Ray-Eldath:
My use case is that we have enforced HTTPS for all of our services, so the header that JWT using is cryptographically protected by it, and therefore no need for another signature, as described in IETF's document.

By the way, the IETF's document also required that `none` shouldn't be the default value. It may, in many cases, unsafe, but I think we must provide an option for some cases described.

>That said, if a JWT is cryptographically protected by a transport layer, such as TLS using cryptographically current algorithms, there may be no need to apply another layer of cryptographic protections to the JWT. In such cases, the use of the “none” algorithm can be perfectly acceptable. The “none” algorithm should only be used when the JWT is cryptographically protected by other means. JWTs using “none” are often used in application contexts in which the content is optionally signed; then the URL-safe claims representation and processing can be the same in both the signed and unsigned cases. JWT libraries SHOULD NOT generate JWTs using “none” unless explicitly requested to do by the caller.

p0pr0ck5:
Yes TLS prevents the contents of the JWT from tampering in flight, but the transport mechanism doesn't prevent a malicious user from _sending_ an unsigned JWT to Kong (nor does this PR enforce that TLS was used as the transport when processing `none` requests).

Again, at the very least this PR should provide an opt-in schema config that allows processing `none` tokens. Kong should not unconditionally process JWTs with the `none` signature without the operations explicit permission.

Tieske:
for reference: https://insomniasec.com/blog/auth0-jwt-validation-bypass

especially: _‘Alg “none” strikes again’_

PR #4924, opened 2019-08-16
feat(*) do not expose openresty version

### LINKED ISSUE #3908: Kong exposes openresty version
### Steps To Reproduce
```
docker run -d --name kong-database \
    -p 5432:5432 \
    -e "POSTGRES_USER=kong" \
    -e "POSTGRES_DB=kong" \
    postgres:9

docker run --rm --link kong-database:kong-database \
    -e "KONG_DATABASE=postgres" \
    -e "KONG_PG_HOST=kong-database" \
    kong:0.14.1-alpine kong migrations up

docker run -d \
    --link kong-database:kong-database \                                                                              
    -e "KONG_DATABASE=postgres" \
    -e "KONG_PG_HOST=kong-database" \
    -e "KONG_PROXY_ACCESS_LOG=/dev/stdout" \
    -e "KONG_ADMIN_ACCESS_LOG=/dev/stdout" \
    -e "KONG_PROXY_ERROR_LOG=/dev/stderr" \
    -e "KONG_ADMIN_ERROR_LOG=/dev/stderr" \
    -e "KONG_HEADERS=off" \
    -p 8000:8000 \
    kong:0.14.1-alpine

curl -X TRACE localhost:8000
```

### Additional Details & Logs
Output of curl:
```
* Rebuilt URL to: localhost:8000/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8000 (#0)
> TRACE / HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 405 Not Allowed
< Server: openresty/1.13.6.2
< Date: Thu, 25 Oct 2018 13:39:30 GMT
< Content-Type: text/html; charset=UTF-8
< Content-Length: 179
< Connection: close
< 
<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>openresty/1.13.6.2</center>
</body>
</html>
* Closing connection 0
```

bungle:
### Summary

Previously, e.g. when doing `TRACE` request:
```
http TRACE :8000
```

Nginx terminated the request with the following headers and body:

```html
HTTP/1.1 405 Not Allowed
Connection: close
Content-Length: 163
Content-Type: text/html; charset=UTF-8
Date: Fri, 16 Aug 2019 09:19:10 GMT
Server: kong/1.2.1

<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>openresty/1.15.8.1</center>
</body>
</html>
```

With this change, the output doesn't expose OpenResty version anymore, although it is to be noted, that it still exposes `openresty`, which can be considered a bug in Nginx/OpenResty default response body templates as they don't seem to respect `server_tokens=off` that this commit adds
to our server templates.

This also affects other default response bodies, such as those sent
with:

```lua
ngx.redirect("<url>");
```

Further development on this, we should perhaps add a white-labeling patch to OpenResty patches to patch the default templates to get rid of OpenResty labeling (which of course does not mean we are not proud of the OpenResty, but just to be more consistent with our product).

### Issues resolved

#3908

### DIFF
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -1,5 +1,6 @@
 return [[
 charset UTF-8;
+server_tokens off;
 
 > if anonymous_reports then
 ${{SYSLOG_REPORTS}}
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -151,7 +151,9 @@ local function do_authentication(conf)
                                     key)
   if err then
     kong.log.err(err)
-    return kong.response.exit(500, "An unexpected error occurred")
+    return kong.response.exit(500, {
+      message = "An unexpected error occurred"
+    })
   end
 
   -- no credential in DB, for this key, it is invalid, HTTP 401
--- a/spec/02-integration/05-proxy/13-error_handlers_spec.lua
+++ b/spec/02-integration/05-proxy/13-error_handlers_spec.lua
@@ -37,4 +37,16 @@ describe("Proxy error handlers", function()
     assert.matches("kong/", res.headers.server, nil, true)
     assert.equal("Request Header Or Cookie Too Large\n", body)
   end)
+
+  it("does not expose OpenResty version", function()
+    local res = assert(proxy_client:send {
+      method = "TRACE",
+      path = "/",
+    })
+
+    assert.res_status(405, res)
+    local body = res:read_body()
+    assert.matches("kong/", res.headers.server, nil, true)
+    assert.not_matches("openresty/", body, nil, true)
+  end)
 end)
--- a/spec/fixtures/custom_nginx.template
+++ b/spec/fixtures/custom_nginx.template
@@ -14,6 +14,7 @@ events {}
 http {
 > if #proxy_listeners > 0 or #admin_listeners > 0 then
     charset UTF-8;
+    server_tokens off;
 
     error_log logs/error.log ${{LOG_LEVEL}};
 


hutchic:
clarifying: this only removes it from the response body in the special cases but will still be in the via header correct?

bungle:
@hutchic,

Server token is Nginx/OpenResty concept. Kong usually adds `Server` header or `Via` header. I can hardly see any use for `Nginx` or `OpenResty` popping up in `bodies` or `headers` (and mind you, their versions).

In someways I would say this is a bit of a white-labeling work. I went ahead and added PR for a patch too: https://github.com/Kong/openresty-patches/pull/51

This and the patch are rather safe to have (I considering patching `nginx.h` too, but that probably needs a bit more thought, including `ps` listings to list `kong` instead of `nginx`, and I didn't want to get rid of Nginx/OpenResty versions totally as they are also usable to us). But this PR and the patches PR will make our responses a bit more consistent.

kong/templates/nginx_kong.lua
```diff
@@ -1,5 +1,6 @@
 return [[
 charset UTF-8;
+#server_tokens off;
```
hishamhm: this is commented out here in the real template, but it is not commented in the `spec/fixtures/custom_nginx.template` below. Is this intentional?

thibaultcha:
Sounds good to me. My only question is the same as @hishamhm's.

kong/templates/nginx_kong.lua
bungle: Oh, good catch.

PR #4925, opened 2019-08-16
fix(Makefile) this make target no longer exists

hutchic:


Makefile
```diff
@@ -46,7 +46,7 @@ nightly-release: setup-kong-build-tools
 	$(MAKE) build-kong && \
 	$(MAKE) release-kong
 
-release: setup-release
+release:
```
hishamhm: shouldn't this instead be: 

```suggestion
release: setup-kong-build-tools
```

PR #4926, opened 2019-08-16
chore(deps) bump lyaml from 6.2.3 to 6.2.4

bungle:
### Summary

Changelog:

- luke works with upgraded bootstrap luarocks version of require

(whatever that means)

### DIFF
--- a/kong-1.3.0-0.rockspec
+++ b/kong-1.3.0-0.rockspec
@@ -25,7 +25,7 @@ dependencies = {
   "luatz == 0.4",
   "http == 0.3",
   "lua_system_constants == 0.1.3",
-  "lyaml == 6.2.3",
+  "lyaml == 6.2.4",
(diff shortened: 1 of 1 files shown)


hishamhm:
It's just a build process fix on their end, there aren't even code changes in that tag. I don't think we're having any build problems with lyaml, so I guess the upgrade is not needed? ("if it ain't broke...")

bungle:
@hishamhm, just keeping things up-to-date, and so that someone else will not propose this same later.

thibaultcha:
Agreed with @bungle. However, let's merge this _after_ the 1.3.0 formal release, because @hishamhm also has a point :)

hutchic:
This might require bumping it here https://github.com/Kong/kong-build-tools/blob/master/Makefile#L47 as well

thibaultcha:
Manually merged to master.

PR #4927, opened 2019-08-16
chore(deps) bump luaossl from 20190612 to 20190731

bungle:
### Summary

Changes:

- introduce workarounds for LuaJIT 47bit userdata


hishamhm:
Since it seems this release contains only the 47bit changes, this is effectively a no-op update for x86_64.

PR #4930, opened 2019-08-16
chore(proxy) only log init worker errors once with tls request

bungle:
### Summary

Previously on TLS/SSL requests we logged init worker errors twice, this changes it so that we will only log it once.

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -665,7 +665,10 @@ function Kong.balancer()
 end
 
 function Kong.rewrite()
-  log_init_worker_errors()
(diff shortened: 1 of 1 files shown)


kong/init.lua
```diff
@@ -665,7 +665,10 @@ function Kong.balancer()
 end
 
 function Kong.rewrite()
-  log_init_worker_errors()
+  local is_https =  ngx.var.https == "on"
```
thibaultcha: style: there's an extra space after `=`

PR #4932, opened 2019-08-16
feat(pdk) add get_authenticated_groups

darrenjennings:
### Summary
Add authenticated groups to pdk kong.client.

Pulling get_authenticated_groups functionality out of acl plugin to be
used in other plugins that need access to groups (e.g. canary, session)

### Issues resolved

Fixes INTF-1954

### DIFF
--- a/kong/pdk/client.lua
+++ b/kong/pdk/client.lua
@@ -244,12 +244,12 @@ local function new(self)
 
   ---
   -- Returns a table of `authenticated_groups` of the currently authenticated
-  -- consumer.
+  -- consumer, or nil if groups is not set.
   -- @function kong.client.get_authenticated_groups
   -- @phases access, header_filter, body_filter, log
-  -- @treturn table,nil|err
-  -- The table will have an array part to iterate over, and a hash part where each
-  -- group name is indexed by itself. Eg.
+  -- @treturn table|nil
+  -- The table will have an array part to iterate over, and a hash part where
+  -- each group name is indexed by itself. Eg.
   -- {
   --   [1] = "users",
   --   [2] = "admins",
@@ -257,8 +257,8 @@ local function new(self)
   --   admins = "admins",
   -- }
   -- @usage
-  -- local groups, err = kong.client.get_authenticated_groups()
-  -- if not err then
+  -- local groups = kong.client.get_authenticated_groups()
+  -- if not groups then
   --   for i = 1, #groups_to_check do
   --     if groups[i] == 'group1' then
   --       return true
@@ -270,7 +270,7 @@ local function new(self)
 
     local authenticated_groups = ngx.ctx.authenticated_groups
     if authenticated_groups == nil then
-      return {}
+      return nil
     end
 
     assert(type(authenticated_groups) == "table",
--- a/kong/pdk/client.lua
+++ b/kong/pdk/client.lua
@@ -12,6 +12,7 @@ local phase_checker = require "kong.pdk.private.phases"
 
 
 local ngx = ngx
+local type = type
 local tonumber = tonumber
 local check_phase = phase_checker.check
 local check_not_phase = phase_checker.check_not
@@ -239,6 +240,50 @@ local function new(self)
     local is_tls = balancer_data and balancer_data.scheme == "tls" and balancer_data.ssl_ctx
 
     return is_tls and "tls" or "tcp"
+  end
+
+  ---
+  -- Returns a table of `authenticated_groups` of the currently authenticated
+  -- consumer.
+  -- @function kong.client.get_authenticated_groups
+  -- @phases access, header_filter, body_filter, log
+  -- @treturn table,nil|err
+  -- The table will have an array part to iterate over, and a hash part where each
+  -- group name is indexed by itself. Eg.
+  -- {
+  --   [1] = "users",
+  --   [2] = "admins",
+  --   users = "users",
+  --   admins = "admins",
+  -- }
+  -- @usage
+  -- local groups, err = kong.client.get_authenticated_groups()
+  -- if not err then
+  --   for i = 1, #groups_to_check do
+  --     if groups[i] == 'group1' then
+  --       return true
+  --     end
+  --   end
+  -- end
+  function _CLIENT.get_authenticated_groups()
+    check_phase(AUTH_AND_LATER)
+
+    local authenticated_groups = ngx.ctx.authenticated_groups
+    if authenticated_groups == nil then
+      return nil
+    end
+
+    if type(authenticated_groups) ~= "table" then
+      return nil, "invalid authenticated_groups, a table was expected"
+    end
+
+    local groups = {}
+    for i = 1, #authenticated_groups do
+      groups[i] = authenticated_groups[i]
+      groups[authenticated_groups[i]] = authenticated_groups[i]
+    end
+
+    return groups
   end
 
 
--- a/kong/pdk/client.lua
+++ b/kong/pdk/client.lua
@@ -270,12 +270,11 @@ local function new(self)
 
     local authenticated_groups = ngx.ctx.authenticated_groups
     if authenticated_groups == nil then
-      return nil
+      return {}
     end
 
-    if type(authenticated_groups) ~= "table" then
-      return nil, "invalid authenticated_groups, a table was expected"
-    end
+    assert(type(authenticated_groups) == "table",
+           "invalid authenticated_groups, a table was expected")
 
     local groups = {}
     for i = 1, #authenticated_groups do
--- a/kong/plugins/acl/groups.lua
+++ b/kong/plugins/acl/groups.lua
@@ -5,7 +5,6 @@ local EMPTY = tablex.readonly {}
 
 
 local kong = kong
-local type = type
 local mt_cache = { __mode = "k" }
 local setmetatable = setmetatable
 local consumer_groups_cache = setmetatable({}, mt_cache)
--- a/kong/plugins/acl/groups.lua
+++ b/kong/plugins/acl/groups.lua
@@ -124,45 +124,10 @@ local function get_current_consumer_id()
 end
 
 
---- Returns a table with all group names.
--- The table will have an array part to iterate over, and a hash part
--- where each group name is indexed by itself. Eg.
--- {
---   [1] = "users",
---   [2] = "admins",
---   users = "users",
---   admins = "admins",
--- }
--- If there are no authenticated_groups defined, it will return nil
--- @return table with groups or nil
-local function get_authenticated_groups()
-  local authenticated_groups = kong.ctx.shared.authenticated_groups
-  if type(authenticated_groups) ~= "table" then
-    authenticated_groups = ngx.ctx.authenticated_groups
-    if authenticated_groups == nil then
-      return nil
-    end
-
-    if type(authenticated_groups) ~= "table" then
-      kong.log.warn("invalid authenticated_groups, a table was expected")
-      return nil
-    end
-  end
-
-  local groups = {}
-  for i = 1, #authenticated_groups do
-    groups[i] = authenticated_groups[i]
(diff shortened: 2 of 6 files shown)


kong/pdk/client.lua
```diff
@@ -241,6 +242,50 @@ local function new(self)
...
+
+    if type(authenticated_groups) ~= "table" then
+      return nil, "invalid authenticated_groups, a table was expected"
+    end
+
+    local groups = {}
+    for i = 1, #authenticated_groups do
+      groups[i] = authenticated_groups[i]
+      groups[authenticated_groups[i]] = authenticated_groups[i]
+    end
```
Tieske: this only copies the array part, not the hash table part. So it doesn't match the doc comments above.

More over by returning a copy of the table we will not be. able to cache anything by the table. So any code using this will have to iterate the table on every request, without cache option.

kong/pdk/client.lua
```diff
@@ -241,6 +242,50 @@ local function new(self)
...
+    check_phase(AUTH_AND_LATER)
+
+    local authenticated_groups = ngx.ctx.authenticated_groups
+    if authenticated_groups == nil then
+      return nil
+    end
+
+    if type(authenticated_groups) ~= "table" then
+      return nil, "invalid authenticated_groups, a table was expected"
+    end
```
Tieske: do we need to test for this? if it's not a table a hard error will occur, which is ok, since this is a issue that should surface with a stacktrace probably.

kong/pdk/client.lua
darrenjennings: > this only copies the array part, not the hash table part. So it doesn't match the doc comments above.

This part was taken from the acl plugin, which creates both the array part, and the hash part. Test output shows this but maybe I miss your point:
https://github.com/Kong/kong/pull/4932/files#diff-ad1fe51fe0f72502d5325b508826376dR35

>More over by returning a copy of the table we will not be. able to cache anything by the table. So any code using this will have to iterate the table on every request, without cache option.

Yes this is true, and for the acl plugin nothing changes (i.e. cache is used by acl in other functions, should it also be used here?). I see canary plugin uses more of the acl plugin functionality than I am needing for https://github.com/Kong/kong-plugin-session/pull/13 so wondering if we expand the features to support canary, if we would use `client` or a new pdk module `groups`. The scope of this PR was to simply get the groups from the context. If we have path for where groups can live, I can use this PR to transplant the acl groups module.

Tieske: > This part was taken from the acl plugin, which creates both the array part, and the hash part. Test output shows this but maybe I miss your point:

Beats me?? you're doing a numerical loop in the code, so that does not copy the hash entries....

Tieske: duh, sorry for the noise, you create a second entry by name.... i was just looking for `pairs`. 

my bad !  😞 🔫

Tieske: > Yes this is true, and for the acl plugin nothing changes (i.e. cache is used by acl in other functions, should it also be used here?). I see canary plugin uses more of the acl plugin functionality than I am needing for Kong/kong-plugin-session#13 so wondering if we expand the features to support canary, if we would use client or a new pdk module groups.

This touches on my general comment; the whole `groups.lua` from ACL and Canary, should be moved into the PDK. That was the reason at the time to refactor the code out of ACL into a separate file that could be reused by Canary. (final step of removing the copy and have canary use the file included with ACL never was completed)

So imo all of that functionality should go into the PDK. Maybe an argument can be added to instruct to return the original table? eg.

```lua
function _CLIENT.get_authenticated_groups(no_copy)
```

Such that it still can be used for caching purposes if needed. Maybe @thibaultcha or @hishamhm have some ideas regarding the design of the API?

Ideally also the DAO entities would be moved into core imo.

kong/pdk/client.lua
```diff
@@ -241,6 +242,50 @@ local function new(self)
...
+  --       return true
+  --     end
+  --   end
+  -- end
+  function _CLIENT.get_authenticated_groups()
+    check_phase(AUTH_AND_LATER)
+
+    local authenticated_groups = ngx.ctx.authenticated_groups
+    if authenticated_groups == nil then
+      return nil
```
Tieske: consideration: you could return an empty table in this case, since this is not necessarily an error condition. The upside would be that error handling becomes easier since you can check the first result, and do not have to check the second one.

Tieske: currently it also does not comply with the doc comments, returning an empty table would.

kong/pdk/client.lua
```diff
@@ -241,6 +242,50 @@ local function new(self)
...
+    end
+
+    if type(authenticated_groups) ~= "table" then
```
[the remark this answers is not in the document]
darrenjennings: @bungle originally had warning during errors  https://github.com/Kong/kong/pull/4013/files#diff-7af3620da927a2715e38b2dc23f8d737R147 so I chose to do as few edits as possible. Any issue with `assert`ing @bungle?

kong/pdk/client.lua
```diff
@@ -241,6 +242,50 @@ local function new(self)
...
+      return nil, "invalid authenticated_groups, a table was expected"
+    end
+
+    local groups = {}
+    for i = 1, #authenticated_groups do
```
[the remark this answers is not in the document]
hishamhm: The PDK generally chooses safe over fast (its target audience is not the typical OpenResty programmer, but someone potentially new to both Kong and Lua). It's not a hard rule, but here I don't know if we want to allow the user to modify the set of authenticated groups by fiddling with the table returned by the getter.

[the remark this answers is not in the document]
Tieske: so our own plugins should not use the PDK then?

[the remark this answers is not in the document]
hishamhm: @Tieske I'm afraid I don't follow. The implementation of `get_authenticated_groups` that is currently in the released version of `kong.plugins.acl.groups` already returns a copy.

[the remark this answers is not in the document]
Tieske: that comment wasn't based on the code, but on your comment:

> The PDK generally chooses safe over fast (its target audience is not the typical OpenResty programmer, but someone potentially new to both Kong and Lua)

If that's what it is, then we shouldn't be using it, right?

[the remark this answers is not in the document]
Tieske: > The implementation of get_authenticated_groups that is currently in the released version of kong.plugins.acl.groups already returns a copy.

The original didn't: https://github.com/Kong/kong-plugin-enterprise-canary/blob/master/kong/plugins/canary/groups.lua#L65

As commented [here](https://github.com/Kong/kong/pull/4013/files#r247037159) and [here](https://github.com/Kong/kong/pull/4932#discussion_r316240061) the whole idea was to have a single set of methods in the PDK. Initial steps were taken, but then reverted by changing the implementation in ACL (diverted from Canary again), instead of moving it into the PDK.

Now we're moving yet another implementation into the PDK???

Can we please do it right this time, and consolidate the code and move it into the PDK. Or if PDK is only for end users, move the base-code into Kong and add safe-wrapper of those methods in the PDK.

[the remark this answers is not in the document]
hishamhm: > If that's what it is, then we shouldn't be using it, right?

No, we _should_ be using it, because reading the code from our plugins is how users learn how to write their own. Part of the work on introducing the PDK was porting over our plugins that used raw OpenResty calls to the PDK.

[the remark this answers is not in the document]
dndx: Even if we decide to return a copy, we should use the [`table.clone`](https://github.com/openresty/luajit2#tableclone) method from OpenResty LuaJIT to make it faster and shorter than doing a manual copy.

In my view if we want more of our code to use the PDK functions we should make it at least possible to not incur additional overheads for all the callers. I'm okay keeping it "safe" by default, but having an optional argument like `no_copy` would make it much more suitable for core code's consumption (especially those that has already been accessing the group object directly and doesn't need a copy).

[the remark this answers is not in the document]
hishamhm: I may be wrong, but I believe the explicit loop there was for adding both array and hash-style access to the resulting table (not sure if this is kept in the original, otherwise, sure, table.clone is a good choice — we should probably get rid of `kong.tools.utils.shallow_copy` as well).

I'm also okay with providing non-default fast-paths with reduced safety in the PDK if those prove necessary, but I'm just not sure if it's worth it here, because as far as I can see core code doesn't actually use authenticated groups, it's something in the realm of plugins only.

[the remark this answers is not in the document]
Tieske: I had a quick look around and couldn't find PDK functions that returns copies. Even getting credentials and consumers returns the real thing. 
Disclaimer: I didn't check everything.

If we return copies, we should also implement that in other places to be consistent across the board. If some return copies and others not, that's even more cumbersome.

kong/pdk/client.lua
hishamhm: Is there a scenario where `ngx.ctx.authenticated_groups` may be an empty table today? (As in, "authentication was checked and there are no groups" or something?) I'm unfamiliar with the details.

I ask this because looking at the code that uses this API in `kong/plugins/acl/handler.lua`, `nil` (not error) and `{}` would fall into two different paths today.

If the API is made to return `{}` when it currently returns `nil`, will client code need to be adjusted accordingly by testing for `if #authenticated_groups > 0` instead of `if authenticated_groups`?

Another way of putting this is: is there a semantic difference today between `{}` and `nil` when returned by `kong.plugins.acl.groups.get_authenticated_groups`? (As in, for example, "groups were checked and none match" vs. "groups were not checked yet") If so, does this distinction matter?

darrenjennings: @hishamhm you bring up a good point. The tests do seem to assume that `nil` is semantically different. For instance, currently one test is failing that is checking `nil`ness of groups here:
https://github.com/kong/kong/blob/304c3dfd947011cead2a47a6f6769afb9e98e325/kong/plugins/acl/handler.lua#L72 but merely substituting length of groups everywhere we do authenticated_groups check to check length of array breaks other tests so changing it back unless @bungle has some background he would like to provide.

darrenjennings: It's also worth noting that now we have `assert`, we no longer need to return err, so `table|nil` for return value is much simpler than `table|nil,err|nil`.

kong/pdk/client.lua
```diff
@@ -241,6 +242,49 @@ local function new(self)
...
+  function _CLIENT.get_authenticated_groups()
+    check_phase(AUTH_AND_LATER)
+
+    local authenticated_groups = ngx.ctx.authenticated_groups
+    if authenticated_groups == nil then
+      return nil
+    end
+
+    assert(type(authenticated_groups) == "table",
+           "invalid authenticated_groups, a table was expected")
```
thibaultcha: Please don't use `assert`, but use `error(..., 2)` (with the proper stacktrace level) instead. Assert will not report the stacktrace from the caller's stack level.

kong/pdk/client.lua
thibaultcha: > Ideally also the DAO entities would be moved into core imo.

Yes. In fact, there is on-going discussions around the "Consumer Groups" entity (there has been many mentions of it in the past already, even as early as 2017 when designing the Services/Routes model). I would suggest we hold off this PR until the entity has been implemented in Core (which may influence how the ACL plugin behaves...).

kong/pdk/client.lua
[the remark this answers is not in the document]
dndx: Agreed. @thibaultcha what's your view on this?

kong/pdk/client.lua
Tieske: I think assert is fine here. The error condition is not the users fault, it's a bug elsewhere, and in that case you'll want the full trace.

If a bad parameter is passed in to the function, then doing an error at level 2 would make sense

hishamhm:
@darrenjennings As discussed out-of-band, we're going to take a different approach on this, so let's close this one.

PR #4940, opened 2019-08-20
perf(proxy) remove regex from hop-by-hop removal on a hot path

bungle:
### Summary

Recently we found out that regex used with hop-by-hop headers can be a 5 to 10% of our hot path code. This PR removes the regex usage on those places.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -18,15 +18,17 @@ local constants    = require "kong.constants"
 local singletons   = require "kong.singletons"
 local certificate  = require "kong.runloop.certificate"
 local concurrency  = require "kong.concurrency"
-local ngx_re       = require "ngx.re"
 local PluginsIterator = require "kong.runloop.plugins_iterator"
 
 
 local kong         = kong
+local type         = type
 local ipairs       = ipairs
 local tostring     = tostring
 local tonumber     = tonumber
 local sub          = string.sub
+local byte         = string.byte
+local gsub         = string.gsub
 local find         = string.find
 local lower        = string.lower
 local fmt          = string.format
@@ -41,7 +43,6 @@ local timer_at     = ngx.timer.at
 local timer_every  = ngx.timer.every
 local re_match     = ngx.re.match
 local re_find      = ngx.re.find
-local re_split     = ngx_re.split
 local update_time  = ngx.update_time
 local subsystem    = ngx.config.subsystem
 local start_time   = ngx.req.start_time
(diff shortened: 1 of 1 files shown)


javierguerragiraldez:
the code is nice, but there are a few intermediate strings created (because of the space trimming).  I've found code like this is usually compiled very efficiently (UNTESTED!)
```
local csv_token do
	local COMMA, SPACE = (", "):byte(1, -1)
	
	function csv_token(s, b)
		b = b or 1
		local first, last
		for i = b, #s do
			local c = s:byte(i)
			if c == COMMA then
				if first and last then
					return s:sub(first, last), i+1
				else
					return '', i+1
				end
			end
			if c ~= SPACE then
				if not first then first = i end
				last = i
			end
		end
		if first then
			return s:sub(first), #s+1
		end
	end
end
```

dndx:
@javierguerragiraldez nice suggestion! I would say for our proxy path more than 90% of the headers we see will be `Connection: keep-alive`, so I did some benchmarks for this particular scenario as well. It is ran on my machine using latest OpenResty and 50M iterations:

Original `re.split` with DFA mode: 7.758s
`re.split` with normal mode: 6.293s
`csv_iterator.lua`: 0.035s
`csv_token.lua`: 1.965s
`csv_iterator.lua` with JIT=off: 7.554s
`csv_token.lua` with JIT=off: 24.741s

I have attached the file I used in https://gist.github.com/dndx/623a29aea034a90b005d6c153155e661. They can be run using the `resty` utility directly.

Not that this method is very scientific but it seems that `csv_iterator.lua` is more JIT friendly. @javierguerragiraldez please correct me if I'm wrong on this but it seems that `csv_iterator.lua` is more suitable for typical proxy path.

kong/runloop/handler.lua
```diff
@@ -1284,38 +1340,23 @@ return {
...
-            end
-          end
+      for _, header_name in csv(var.http_connection) do
+        -- some of these are already handled by the proxy module,
+        -- proxy-authorization being an exception that is handled
+        -- below with special semantics.
+        if header_name ~= "close" and
+           header_name ~= "upgrade" and
+           header_name ~= "keep-alive" and
+           header_name ~= "proxy-authorization" then
```
dndx: Style: we usually write long `if` conditions like this:

```
if header_name ~= "close" and
   header_name ~= "upgrade" and
   header_name ~= "keep-alive" and
   header_name ~= "proxy-authorization"
then
  -- blah
end
```

to make it more readable.

kong/runloop/handler.lua
```diff
@@ -1376,20 +1417,11 @@ return {
...
-              local header_name = lower(header_names[i])
-              if header_name ~= "close" and
-                 header_name ~= "upgrade" and
-                 header_name ~= "keep-alive" then
-                header[header_names[i]] = nil
-              end
-            end
-          end
+      for _, header_name in csv(var.upstream_http_connection) do
+        if header_name ~= "close" and
```
dndx: Same here.

bungle:
Ok, I will do the change from `string.sub` to `string.byte`. I am also thinking about making `keep-alive` check before iterating.

javierguerragiraldez:
> it seems that csv_iterator.lua is more suitable for typical proxy path.

heh, that's the value of benchmarking with representative data!  I agree that the absolute best would be to hardcode the most common cases, and just split if it's not one of them.

bungle:
@dndx / @javierguerragiraldez,

I have made the changes you suggested. Thanks for the feedback.

javierguerragiraldez:
the test failure seems spurious to me ("address already in use").  if so, I'd say this is ready for merging

PR #4943, opened 2019-08-20
chore(deps) dependency pinning

hutchic:
- pin all dependencies for CI testing and releasing in one spot
- speed up releases by skipping the install phase which is not needed


hutchic:
going to approach this differently

PR #4946, opened 2019-08-20
feat(schema) make `ca_certificates.cert` field unique

dndx:
Thanks @hbagdi for pointing it out!

This is a very small change and ideally goes out in 1.3.0 to avoid any future migration pains.

### DIFF
--- a/kong/db/migrations/core/005_120_to_130.lua
+++ b/kong/db/migrations/core/005_120_to_130.lua
@@ -14,7 +14,7 @@ return {
       CREATE TABLE IF NOT EXISTS "ca_certificates" (
         "id"          UUID                       PRIMARY KEY,
         "created_at"  TIMESTAMP WITH TIME ZONE   DEFAULT (CURRENT_TIMESTAMP(0) AT TIME ZONE 'UTC'),
-        "cert"        TEXT NOT NULL,
+        "cert"        TEXT NOT NULL              UNIQUE,
         "tags"        TEXT[]
       );
 
@@ -103,6 +103,8 @@ return {
         tags set<text>,
         PRIMARY KEY (partition, id)
       );
+
+      CREATE INDEX IF NOT EXISTS ca_certificates_cert_idx ON ca_certificates(cert);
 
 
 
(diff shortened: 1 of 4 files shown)


hishamhm:
@dndx CI failures seem related. Should this be labeled as `fix()` (or `hotfix()` even), since I believe was the originally intended design?

I'm ok with pushing this to 1.3.0 final once tests pass.

dndx:
@hishamhm Looking into the error (strange it passed on my machine locally).

I was tempted to not call it a `fix`, because the `Certificate` object has the same problem today but obviously 1.3.0 is not a good time for addressing that yet. Since we claimed that the `CA Certificate` behavior is exactly the same as `Certificate` without a `key` field, you see the problem here :)

hbagdi:
Looks good.
@dndx Thank you for the quick turn-around here!

kong/db/migrations/core/005_120_to_130.lua
```diff
@@ -104,6 +104,8 @@ return {
         PRIMARY KEY (partition, id)
       );
 
+      CREATE INDEX IF NOT EXISTS ca_certificates_cert_idx ON ca_certificates(cert);
```
bungle: Please remove this:

The `UNIQUE` constraint is implemented in `Postgres` as unique index.

Tieske: curious @bungle : so if you add a UNIQUE constraint, then Postgres automatically adds an index already on that field?

bungle: @Tieske that's right. But in this case the index was for Cassandra.

bungle: @Tieske see also: https://www.postgresql.org/docs/12/indexes-unique.html
> There's no need to manually create indexes on unique columns; doing so would just duplicate the automatically-created index.

Or https://www.postgresql.org/docs/9.4/indexes-unique.html:
> Note: The preferred way to add a unique constraint to a table is ALTER TABLE ... ADD CONSTRAINT. The use of indexes to enforce unique constraints could be considered an implementation detail that should not be accessed directly. One should, however, be aware that there's no need to manually create indexes on unique columns; doing so would just duplicate the automatically-created index.

hishamhm:
@dndx since 1.3.0 is just around the corner, could you also submit a PR for kong-upgrade-tests exercising this? Thank you!

PR #4958, opened 2019-08-23
docs(admin-api) document the fourth level of auto-generated admin apis

bungle:
### Summary

This PR adds the following:

1. a generic description about auto-generated endpoint
2. support for `admin_api_name` and `admin_api_nested_name` schema parameters
3. support for the fourth level of entity endpoint

### DIFF
--- a/autodoc/data/admin-api.lua
+++ b/autodoc/data/admin-api.lua
@@ -1527,11 +1527,11 @@ return {
     endpoint_w_ek = [[
       ##### ${Active_verb} ${Entity}
 
-      <div class="endpoint ${method}">/${entities_url}/{${endpoint_key} or id}</div>
+      <div class="endpoint ${method}">/${entities_url}/{${entity} ${endpoint_key} or id}</div>
 
       Attributes | Description
       ---:| ---
-      `${endpoint_key} or id`<br>**required** | The unique identifier **or** the ${endpoint_key} of the ${Entity} to ${active_verb}.
+      `${entity} ${endpoint_key} or id`<br>**required** | The unique identifier **or** the ${endpoint_key} of the ${Entity} to ${active_verb}.
     ]],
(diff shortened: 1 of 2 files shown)


autodoc/data/admin-api.lua
```diff
@@ -60,6 +60,28 @@ return {
         care should be taken when setting up Kong environments to avoid undue public
         exposure of this API. See [this document][secure-admin-api] for a discussion
         of methods to secure the Admin API.
+
+        The most of the Kong admin API is auto-generated based on entity schemas,
```
hishamhm: This description is great, but I think it should be in the plugin developer guide, not the Admin API reference (for a Kong end-user who never customizes Kong, this is an internal implementation detail; for a Kong plugin author who creates entities, this is super relevant — this also makes me think that we should at one point auto-generate Admin API references for the custom plugin entities in the documentation pages for plugins (heads up @Kong/team-docs!)). I suggest we open a PR in docs.konghq.com with this section added to the Plugin development guide.

hishamhm:
Script changes look good! I also generated the docs locally and the changes look correct. (CI failures are unrelated.)

My only suggestion above is to move the explanation of how the Admin API is auto-constructed to the Plugin development guide rather than the Admin API reference. (I think reference docs generally are more about the "list of what is available" than "explain how it is built", @Kong/team-docs may want to clarify.)

autodoc/data/admin-api.lua
bungle: The idea was that when user sees this at the beginning, he can in most cases just skip the rest of the admin api docs.

bungle: @hishamhm, do you think we should still remove it? It is one commit to be removed, so not a big deal.

bungle: I have a backup of that. And will send a PR later to docs.

PR #4959, opened 2019-08-23
feat(upstream) add host_header to upstream entities

locao:
Add support to hostname attribute in upstream entities.

This attribute is used as the `Host` header when proxying requests through Kong. It takes precedence over the `Host` that clients send to Kong.


hbagdi:
Question: Will active HTTP health-check set the host header to this new property (if present) as well?

locao:
> Will active HTTP health-check set the host header to this new property (if present) as well?

No, it won't. I will change this behavior, thanks for pointing that, @hbagdi!

Tieske:
This change doesn't make sense to me.

afaik we have 2 possible values for the upstream host-header; the incoming host-header, or the hostname of the upstream server.

This subject is complicated because we do not cleanly separate teh construction of the upstream request in our flow. As mentioned elsewhere; we need to collect info in `pre-access` phase and only build teh upstream properties in `post-access`.

Adding this extra property only makes thing more complicated imo.

In pseudo code:
```lua
-- pre-access
host_header_incoming = request.headers["Host"]
host_header_upstream = nil
host_header_preserve = route.preserve_hostname

-- pdk
function set_upstream_host_header(name)
  host_header_upstream = name
end
function set_host_preserve(bool)
  host_header_preserve = bool
end

--post-access
ip, port, resolved_name = dns.resolve()  -- includes balancers, returns hostname resolved, or target name if balancer

host_header_upstream = host_header_upstream or
                       (host_header_preserve and host_header_incoming) or
                       resolved_name
```

Does this make sense?

kong/runloop/balancer.lua
```diff
@@ -164,15 +164,22 @@ local function apply_history(rb, history, start)
...
+    if history[i].upstream then
+      local upstream = get_upstream_by_id(history[i].upstream.id)
+      if upstream and upstream.hostname then
+        hostname = upstream.hostname
+      end
+    end
 
     if target.weight > 0 then
-      assert(rb:addHost(target.name, target.port, target.weight))
+      assert(rb:addHost(hostname, target.port, target.weight))
```
hishamhm: One thing is confusing me a lot here. It was my understanding that the `upstream.hostname` attribute would only change the `Host` header sent with proxied requests, but not change the balancer logic otherwise. With this change, it seems that when `upstream.hostname` is set, this is no longer adding targets with their different IPs to the ring-balancer object (but instead is adding hostname-based entries to the object which are then subjected to DNS resolution).

I thought the idea was that, if I create a setup like this:

```
upstreams:
- name: myupstreaam
  hostname: example.com
  targets:
  - target: 10.0.0.1:80
  - target: 10.0.0.2:80
```

I want requests to be proxied either `10.0.0.1:80` or `10.0.0.2:80` with a header saying `Host: example.com` but not proxied to `example.com:80`.

Am I missing something?

spec/02-integration/05-proxy/10-balancer/01-ring-balancer_spec.lua
```diff
@@ -1310,6 +1421,77 @@ for _, strategy in helpers.each_strategy() do
...
+              local server2_oks = math.floor(requests / 4)
+              local server1 = http_server("localhost", port1,
+                { requests - server2_oks }, "false", "http", "true")
+              local server2 = http_server("localhost", port2, { server2_oks },
+                "false", "http", "true")
+
+              -- configure healthchecks
+              begin_testcase_setup(strategy, bp)
+              local upstream_name, upstream_id = add_upstream(bp, {
+                hostname = "localhost",
```
hishamhm: Can you make this a value other than `localhost`, some invalid name like `upstream-hostname.test` to make sure this test does not use the underlying system's name resolution in any way?

kong/runloop/balancer.lua
```diff
@@ -271,12 +278,30 @@ do
...
+        local hostname = tgt.hostname
         local ok, err
-        ok, err = balancer:setAddressStatus(status, tgt.ip, tgt.port, tgt.hostname)
+        ok, err = balancer:setAddressStatus(status, tgt.ip, tgt.port, hostname)
+
+        if not ok then
+          -- Fallback: target was not found in balancer with hostname reported
+          -- by healtchecker, will look for it using the name it was added
+          local upstream = get_upstream_by_id(upstream_id)
+          local history = fetch_target_history(upstream)
```
hishamhm: Instead of scanning the target history on every healthchecker event, you can pass `upstream.hostname` as an argument to `attach_healthchecker_to_balancer` and use that inside the `hc_callback`.

PR #4960, opened 2019-08-25
fix(cors) handle empty list of allowed methods

marckhouzam:
### Summary

An empty list of 'headers' or of 'exposed headers' is treated like a
null list by the CORS plugin.  The same should be done for the list
of 'methods'.

This is valuable because some frontends for Kong automatically create
an empty list for methods instead of a null list.  I saw this with
pantsel/Konga version 0.14.1, which is the latest version at this time.
Although this is a bug of the frontend, the CORS plugin should treat
the 'methods' list just like for 'headers' and 'exposed-headers'.

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -214,7 +214,7 @@ function CorsHandler:access(conf)
     end
   end
 
-  local methods = conf.methods and concat(conf.methods, ",")
+  local methods = conf.methods and #conf.methods > 0 and concat(conf.methods, ",")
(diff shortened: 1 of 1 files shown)


marckhouzam:
Darn, the last time I contributed, there was no CLA.
Since I submitted this patch on behalf of the City of Montreal, I have to get them to sign the CLA.  This will take months due to bureaucracy.

If you agree that this PR is correct, I suggest someone else code it again (it's one obvious line) and commit it to avoid months of delay.

BTW, the contributing documentation does not mention the CLA.

marckhouzam:
What I'm going to do instead is provide the contribution personally.

PR #4961, opened 2019-08-25
fix(cors) handle empty list of allowed methods

marckhouzam:
### Summary

An empty list of 'headers' or of 'exposed headers' is treated like a
null list by the CORS plugin.  The same should be done for the list
of 'methods'.

This is valuable because some frontends for Kong automatically create
an empty list for methods instead of a null list.  I saw this with
pantsel/Konga version 0.14.1, which is the latest version at this time.
Although this could be seen as a bug of the frontend, the CORS plugin 
should treat the 'methods' list just like for 'headers' and 'exposed-headers'.

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -214,7 +214,7 @@ function CorsHandler:access(conf)
     end
   end
 
-  local methods = conf.methods and concat(conf.methods, ",")
+  local methods = conf.methods and #conf.methods > 0 and concat(conf.methods, ",")
                   or "GET,HEAD,PUT,PATCH,POST,DELETE,OPTIONS,TRACE,CONNECT"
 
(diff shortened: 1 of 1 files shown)


hishamhm:
> Unless someone configures a global CORS plugin and then wants to block all CORS for one particular API? In that case, I guess they could achieve this by adding the plugin to that API (service or route) and specify an empty list ([]) of methods to block all CORS for that particular API? (This could also be achieved by using a bogus origin list, I believe.)

The trouble with backwards compatibility is that, given that the scenario you described is _possible_, we have no way of knowing if some user out there is already using it like this, and if changing this would break things for them (especially concerning that it would break things by making disallowed things becoming allowed)...

marckhouzam:
Ok, so due to backwards-compatibility issues, I'll just close this PR.

Thanks for taking the time to discuss it.

PR #4963, opened 2019-08-25
feat(*) http2 by default in ssl listeners

gszr:
### Summary

Enabling HTTP/2 by default in SSL listeners is a backwards-compatible
change, since ALPN allows clients that don't support HTTP/2 to negotiate
HTTP/1.1. Non-SSL listeners are left with HTTP/1.1.


bungle:
+1 for this to `1.4`. We skipped it for `1.3` as we were already on rc-phase. @thibaultcha, is it ok to get this to `next`?

hishamhm:
After conversation with @thibaultcha we are left with a few questions. Are all clients guaranteed to implement ALPN? This needs more investigation on compatibility. (Our own testsuite doesn't seem to be exercising this.)

dndx:
@hishamhm @gszr It works the other way around, only clients who implement ALPN **and** supports HTTP/2 will be handled as HTTP/2 traffic. For cases you mentioned HTTP 1 would be used.

This is by design otherwise ALPN itself would be a breaking change.

PR #4968, opened 2019-08-26
fix(dao) remove options.ttl and options.tags context checks

eskerda:
# Summary

This PR removes `options.ttl` and `options.tags` context validity checks (ie: failing a select with ttl on the options). These checks will fail, for instance, during an update of an entity, when options are being proxied from the update to a select during a `read_before_write`. 

Since there's already magic in place on the dao for checking these options depending on the context, these checks are an extra safeguard. What this proposes is to silently ignore options that do not make sense on a given context.

### Full changelog
* Remove `options.ttl` context checks on options validation from dao
* Remove `options.tags` context checks on options validation from dao

### Issues resolved
Makes it possible to include `ttl` on compound update queries

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -138,13 +138,7 @@ local function validate_options_value(options, schema, context)
   local errors = {}
 
   if schema.ttl == true and options.ttl ~= nil then
-    if context ~= "insert" and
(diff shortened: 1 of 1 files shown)


eskerda:
@hishamhm done!

PR #4969, opened 2019-08-26
fix(admin) return empty JSON array when consumer is not found

hbagdi:


spec/02-integration/04-admin_api/03-consumers_routes_spec.lua
```diff
@@ -269,6 +269,11 @@ describe("Admin API (#" .. strategy .. "): ", function()
         assert.equal(1, #json.data)
         assert.same(c, json.data[1])
       end)
+      it("returns emtpy json array when consumer does not exist #harry", function()
```
Tieske: Left over: `#harry`

PR #4971, opened 2019-08-27
feat(admin) implement GET /config for db-less

hbagdi:
The endpoint exports the currently running configuration for DB-less
nodes.
The configuration is returned as YAML string encoded inside a JSON
object with a `config` key.
This is chose for two reasons:
- The output of this endpoint can then be directly fed in as well,
  which helps in debugging, the primary use-case for this endpoint:
  ```
  http :8001/config -o response.json    # export
  cat response.json | http :8001/config # re-import on different node
  ```
- It would be idea to send back client a JSON configuration in the body
  but that's not possible yet as the export method only supports YAML
  and adding that is outside the scope of this work.
  Once JSON is supported as an output format, it can be added to this
  endpoint in a non-breaking change fashion by looking up the
  content-type in the `accept` header sent in the request.

### DIFF
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -20,6 +20,29 @@ local _reports = {
 
 return {
   ["/config"] = {
+    GET = function(self, db)
+      if kong.db.strategy ~= "off" then
+        return kong.response.exit(400, {
+          message = "this endpoint is only available when Kong is " ..
+                    "configured to not use a database"
+        })
+      end
+
+      local file = {
+        buffer = {},
+        write = function(self, str)
+          self.buffer[#self.buffer + 1] = str
+        end,
+      }
+
+      local ok, err = declarative.export_from_db(file)
+      if not ok then
+        kong.log.err("failed exporting config from cache: ", err)
+        return kong.response.exit(500, { message = "An unexpected error occurred" })
+      end
+
+      return kong.response.exit(200, { config = table.concat(file.buffer) })
+    end,
     POST = function(self, db)
       if kong.db.strategy ~= "off" then
         return kong.response.exit(400, {
--- a/spec/02-integration/04-admin_api/15-off_spec.lua
+++ b/spec/02-integration/04-admin_api/15-off_spec.lua
@@ -186,148 +186,51 @@ describe("Admin API #off", function()
   end)
 
   describe("/config", function()
-    it("accepts configuration as JSON body", function()
-      local res = assert(client:send {
-        method = "POST",
-        path = "/config",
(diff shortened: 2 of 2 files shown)


kong/api/routes/config.lua
```diff
@@ -20,6 +20,29 @@ local _reports = {
...
+        buffer = {},
+        write = function(self, str)
+          self.buffer[#self.buffer + 1] = str
+        end,
+      }
+
+      local ok, err = declarative.export_from_db(file)
+      if not ok then
+        kong.log.err("failed exporting config from cache: ", err)
+        return kong.response.exit(500, { message = "An unexpected error occurred" })
```
Tieske: A while ago we discussed not using this, but instead throwing an error. Or did I misunderstand that?

bungle: afaik throwing error will call kong_error_handlers which is very different than response.exit. that should be fixed first I think.

bungle: Also with throwing errors, you need to be very careful where you do so. E.g. if you do it on `header_filter` it will just cause Nginx to close the connection immediately, and that is not what we want. We should revisit this as a whole at some point.

Tieske: Afaik throwing an error results in a 500, and always a proper table error as above (we've seen occurrences where the manual errors were just strings, so we'd prevent that), and also a stack trace in the logs (so also no need for the extra manual log line just above).

@bungle why wouldn't you want to close the connection immediately in the header_filter phase? If an error occurs, then the request is in an undetermined state, so imo nothing else you can do, but close the connection.

bungle: @Tieske currently we catch errors in plugins on some phases and I think with admin api, but nit all phases, like header_filter. I think it would be better to give a proper 500 error on header_filter too rather than just closing connection and responding with nothing.

Tieske: Yes, a proper 500 is a requirement.

That said; why the inconsistency in error handling?

bungle: @Tieske originally we didn't handle errors in any phase, but when we added delayed response, we wrote it with `coroutine:wrap` that is a `pcall` with steroids. I am not sure if we want to be 100% consistent. E.g. handling error in body_filter is tricky, we cannot change headers anymore, and perhaps some content was already sent. And in log phase, we should perhaps continue with next plugin even if some other plugin fails. But... well this is getting off-topic, sorry for hijacking.

Tieske: this is tracked elsewhere, so resolving this.

PR #4977, opened 2019-08-28
feat: implemented a new status interface

hbagdi:
Problem
-------

Kong's Admin API exposes `/status` endpoint, which is consumed by other
services in the infrastructure to monitor Kong's health.

While this has worked for Kong for a long time,
we run into following issues with the current design:
- Users want to health-check Kong and also protect Kong's Admin API at
  the same time. If Kong's Admin API is protected using an auth plugin,
  health-checking becomes difficult. The reason being not all
  orchestration/platform tools performing health-checks support adding
  arbitrary headers to health-check requests.
- Users want to expose Kong's health outside a private network but keep
  the Admin API locked down.
- It is hard to perform HTTP based health-checks for a Kong node running
  only the proxy interface i.e. the Admin interface is disabled.
- A plugin could also potentially need to expose an endpoint which should not
  protected by any authentication.
  Prometheus plugin wants to expose `/metrics`, which should be not
  protected by any auth.
- Provision Kong Enterprise with RBAC enabled (health check won't pass):
  With Kong Enterprise, users want to enable RBAC from the get go,
  meaning when Kong is deployed, RBAC should be enabled. When this is
  done, health-checks for Kong using the current `/status` would fail
  (as we don't have any RBAC user while provisioning). This creates a
  chicken and egg where Kong needs to be up and running but that won't
  happen if status check is failing.

To work around this, users have resorted to injecting a custom server block
inside Kong using the NGINX directive injection feature, which adds
complexity and overhead in setting up Kong in a production environment.
Kong's Ingress Controller creates an additional file, mounts that as a volume
in the Docker container and then enables health-checking.

Health-checks and observability are common enough
(and simple for most deployments) that they should supported
out of the box and not require customizations.

Solution
--------

We introduce a `health` listener in Kong, which exposes insensitive health,
metrics and error read-only information from Kong.
Out of the box, only `/status` endpoint from Admin interface is available,
but plugins can add endpoints to this interface, as they can do for the Admin
interface.
This endpoint is strictly meant for machine-to-machine communication and
should not require any authentication/authorization.

The health server is disabled by default for
backwards-compatibility and not surprise existing users.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -59,6 +59,19 @@
                                           # is adjusted by the `log_level`
                                           # property.
 
+#status_access_log = off                  # Path for Status API request access
+                                          # logs. The default value of `off`
+                                          # implies that loggin for this API
+                                          # is disabled by default.
+                                          # If this value is a relative path,
+                                          # it will be placed under the
+                                          # `prefix` location.
+
+#status_error_log = logs/status_error.log # Path for Status API request error
+                                          # logs. The granularity of these logs
+                                          # is adjusted by the `log_level`
+                                          # property.
+
 #plugins = bundled               # Comma-separated list of plugins this node
                                  # should load. By default, only plugins
                                  # bundled in official distributions are
@@ -234,7 +247,20 @@
                          # capabilities) pulling its configuration changes
                          # from the database.
                          #
-                         # Example: `stream_listen = 127.0.0.1:8444 http2 ssl`
+                         # Example: `admin_listen = 127.0.0.1:8444 http2 ssl`
+
+ #status_listen = off    # Comma-separated list of addresses and ports on
+                         # which the Status API should listen.
+                         # The Status API is a read-only endpoint
+                         # allowing monitoring tools to retrieve metrics,
+                         # healthiness, and other non-sensitive information
+                         # of the current Kong node.
+                         #
+                         # This value can be set to `off`, disabling
+                         # the Status API for this node.
+                         #
+                         # Example: `status_listen = 0.0.0.0:8100`
+
 
 #nginx_user = nobody nobody      # Defines user and group credentials used by
                                  # worker processes. If group is omitted, a
--- a/kong/api/api_helpers.lua
+++ b/kong/api/api_helpers.lua
@@ -1,10 +1,21 @@
 local utils = require "kong.tools.utils"
 local cjson = require "cjson"
+local pl_pretty = require "pl.pretty"
+local tablex = require "pl.tablex"
+local app_helpers = require "lapis.application"
+local arguments = require "kong.api.arguments"
+local Errors = require "kong.db.errors"
+local singletons = require "kong.singletons"
 
-local type = type
-local pairs = pairs
+local ngx      = ngx
+local sub      = string.sub
+local find     = string.find
+local type     = type
+local pairs    = pairs
+local ipairs   = ipairs
 
 local _M = {}
+local NO_ARRAY_INDEX_MARK = {}
 
 -- Parses a form value, handling multipart/data values
 -- @param `v` The value object
@@ -13,7 +24,6 @@ local function parse_value(v)
   return type(v) == "table" and v.content or v -- Handle multipart
 end
 
-local NO_ARRAY_INDEX_MARK = {}
 
 -- given a string like "x[1].y", return an array of indices like {"x", 1, "y"}
 -- the path parameter is an output-only param. the keys are added to it in order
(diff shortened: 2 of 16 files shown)


hishamhm:
> (b) I don't know if Lua module hierarchy is part of 1.0 compat promise or not.

Sort of. At the very least, `kong.api.endpoints` is a _de facto_ API used by plugins.

kong/api/api_helpers.lua
```diff
@@ -194,4 +204,227 @@ do
...
+    if not content_length then
+      local _, err = ngx.req.socket()
+      if err == "no body" then
+        return
+      end
+    end
+
+  elseif sub(content_type, 1, 16) == "application/json"                  or
+         sub(content_type, 1, 19) == "multipart/form-data"               or
+         sub(content_type, 1, 33) == "application/x-www-form-urlencoded" then
```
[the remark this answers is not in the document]
hbagdi: This is part of a refactor commit and is not new code that is being added in this PR.
Can we address this in a later commit/PR if needed?

kong/api/api_helpers.lua
```diff
@@ -194,4 +204,227 @@ do
...
+  if err.strategy then
+    err.strategy = nil
+  end
+
+  if err.code == Errors.codes.SCHEMA_VIOLATION
+  or err.code == Errors.codes.INVALID_PRIMARY_KEY
+  or err.code == Errors.codes.FOREIGN_KEY_VIOLATION
+  or err.code == Errors.codes.INVALID_OFFSET
+  or err.code == Errors.codes.FOREIGN_KEYS_UNRESOLVED
+  then
```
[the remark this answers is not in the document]
hbagdi: Same. Part of a refactor commit.

kong/api/routes/health.lua
```diff
@@ -0,0 +1,88 @@
...
+        if not pok then
+          return kong.response.exit(400, { message = perr })
+        end
+      end
+
+      -- nginx stats
+
+      local r = ngx.location.capture "/nginx_status"
+      if r.status ~= 200 then
+        return kong.response.exit(500, { message = "An unexpected error happened" })
```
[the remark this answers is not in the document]
hbagdi: Not sure. This is part of the refactor commit as well.

kong/api/api_helpers.lua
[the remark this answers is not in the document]
javierguerragiraldez: :+1:

hishamhm:
@hbagdi one small but important "blocker" for this PR is to decide on a name. We all seem to agree that this is going to be eventually useful for more than node health. @thibaultcha suggested `cluster_listen`, which is better but _maybe_ too generic? I've considered `internal_listen`, but @kidd mentioned that this might confuse with the notion of `internal` in NGINX. `cluster_internal_listen` maybe? (or too clunky?) Any ideas?

hbagdi:
As discussed internally, we are going with the name `status_listen` for this feature.
I've updated the PR to reflect the new name.

javierguerragiraldez:
the test failure "Upstream foo has no balancer" seems legit.  maybe it's missing some config?

kong/api/routes/health.lua
```diff
@@ -0,0 +1,88 @@
+local utils = require "kong.tools.utils"
+
+local find = string.find
+local select = select
+local tonumber = tonumber
+local kong = kong
+local knode  = (kong and kong.node) and kong.node or
+               require "kong.pdk.node".new()
```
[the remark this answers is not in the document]
javierguerragiraldez: left for "some day.."  fixups

kong/api/routes/health.lua
```diff
@@ -0,0 +1,88 @@
...
+
+      -- nginx stats
+
+      local r = ngx.location.capture "/nginx_status"
+      if r.status ~= 200 then
+        return kong.response.exit(500, { message = "An unexpected error happened" })
+      end
+
+      local var = ngx.var
+      local accepted, handled, total = select(3, find(r.body, "accepts handled requests\n (%d*) (%d*) (%d*)"))
```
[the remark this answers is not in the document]
javierguerragiraldez: left for a rainy day

kong.conf.default
```diff
@@ -234,7 +234,19 @@
                          # capabilities) pulling its configuration changes
                          # from the database.
                          #
-                         # Example: `stream_listen = 127.0.0.1:8444 http2 ssl`
+                         # Example: `admin_listen = 127.0.0.1:8444 http2 ssl`
+
+ #status_listen = off    # Comma-separated list of addresses and ports on
+                         # which the Status interface should listen.
+                         # Status interface allows you to
```
thibaultcha: typo: The* Status interface. Also not a fan of addressing directly to the reader ("you"), but the pattern is already established in a few other configuration properties...

kong.conf.default
```diff
@@ -234,7 +234,19 @@
...
+                         # read health, metrics and other non-sensitive
```
thibaultcha: s/read/retrieve?

Suggestion:

> The status interface is a read-only API allowing monitoring tools to retrieve metrics, healthiness statuses, and other non-sensitive information from the current Kong node.

kong/api/routes/health.lua
```diff
@@ -0,0 +1,88 @@
...
+local kong = kong
+local shms = {}
+
+for shm_name, shm in pairs(ngx.shared) do
+  table.insert(shms, {
+    zone = shm,
+    name = shm_name,
+    capacity = shm:capacity(),
+  })
+end
```
thibaultcha: The above loop and `shms` variable are unused and seem to be copied from the PDK's `node.lua` module? That doesn't seem to be part of the refactor commit (or rather, it shouldn't), given I don't currently see it in today's Admin API code.

kong/conf_loader.lua
```diff
@@ -70,6 +70,10 @@ local DYNAMIC_KEY_NAMESPACES = {
     injected_conf_name = "nginx_admin_directives",
     prefix = "nginx_admin_", -- TODO: nginx_http_admin (optional)
   },
+  {
+    injected_conf_name = "nginx_status_directives",
+    prefix = "nginx_status_",
```
thibaultcha: Note that the "correct" prefix for this new category of injected directives should be `nginx_http_status_` (note the TODO comments in the previous prefixes). It was unfortunate that we weren't more specific in the naming when we first introduced them, but we haven't updated their name yet for backwards-compatibility reasons.

We should either directly introduce this new prefix with the chosen future naming convention, or also include a TODO comment for a future refactor of the directives injection prefixes. The latter might be preferable for now. This isn't a blocker, but it's important to keep it in mind.

kong/init.lua
```diff
@@ -1152,4 +1152,22 @@ function Kong.admin_header_filter()
 end
 
 
+function Kong.serve_status_api(options)
+  kong_global.set_phase(kong, PHASES.admin_api)
+
+  local options = options or {}
```
thibaultcha: This argument is never given to `serve_status_api`, let's not introduce it only to remove it later on. A subsequent commit can remove the one from `admin_content` later on.

kong/init.lua
```diff
@@ -1152,4 +1152,22 @@ function Kong.admin_header_filter()
...
+
+  if ngx.req.get_method() == "OPTIONS" then
+    header["Access-Control-Allow-Methods"] = "GET, HEAD, PUT, PATCH, POST, DELETE"
+    header["Access-Control-Allow-Headers"] = "Content-Type"
+
+    return ngx.exit(204)
+  end
+
+  return lapis.serve("kong.status", options)
+end
```
thibaultcha: Now that the Admin API always returns latency headers, so should this new status API. Imho this would be the largest blocker for this PR today.

kong/templates/nginx_kong.lua
```diff
@@ -252,4 +252,37 @@ server {
...
 > end
+
+> if #status_listeners > 0 then
+server {
+    server_name kong_status;
+> for i = 1, #status_listeners do
+    listen $(status_listeners[i].listener);
+> end
+
+    access_log off;
```
thibaultcha: Wouldn't it be preferable to write access logs to the same location as the Admin API access logs? In any case, disabling them unconditionally seems a bit extreme and users may feel blind spotted.

hbagdi: Given that this is a health-check and metrics endpoint in the current form, in most deployments, it is likely users will not want access logs for requests to this endpoint to be logged.
We can move this as an injected directive inside `kong.conf`.
It seems like a sane and an important default.
Thoughts?

kong/conf_loader.lua
hbagdi: Thank you for catching this!
Changed to `nginx_http_status_`

hbagdi:
@thibaultcha 
Updated with the following changes:
- includes latency tokens now
- injected directives are now prefixed with `nginx_http_status_` and added a test
- ensured parity in the refactor commit for `/status` endpoint
- updated the kong.conf description.

kong/templates/nginx_kong.lua
javierguerragiraldez: this seems to be the only missing point.  is this a blocker to merge?

kong.conf.default
```diff
@@ -234,7 +234,20 @@
                          # capabilities) pulling its configuration changes
                          # from the database.
                          #
-                         # Example: `stream_listen = 127.0.0.1:8444 http2 ssl`
+                         # Example: `admin_listen = 127.0.0.1:8444 http2 ssl`
+
+ #status_listen = off    # Comma-separated list of addresses and ports on
+                         # which the Status interface should listen.
+                         # The Status interface is a read-only endpoint
```
thibaultcha: I can't help but notice that the capitalized "Status" reads somewhat strangely. Beside that, I also can't help but notice that the terminology we use to refer to each listener interface is getting out of hand. I know this topic was debated internally, and I agree with "status" as a name. What I have an issue with is its strange capitalization and the qualifier ("interface").

As of this PR, we'd have:

- "proxy server"
- "Admin API"
- "stream mode" (also referred to as "stream proxy")
- "Status interface"

We should eventually address the naming inconsistencies among existing listeners, but we can make sure that this new listener uses a name we are comfortable with today, so that we avoid changing it later on.

I suggest "Status API" since it exposes a RESTful API just like the Admin API, while the proxy and stream listeners are reverse-proxy servers. If we were to stick with "interface", I'd suggest not capitalizing "status".

kong/templates/nginx_kong.lua
thibaultcha: I'm not a fan of adding yet another `*_access_log` config property, especially given that it can easily be added with an NGINX dynamic directive (`nginx_http_status_access_log = logs/status_access.log`). However, I'm also not totally satisfied with having to let users figure this out by themselves.

Over the last few years, I have often seen users being particularly demanding about keeping access logs enabled for all exposed listeners for observability's sake.

**Side note**: as of today, errors in the Status API would be logged to the proxy server error logs since we do not override the `error_log` directive in the status `server {}` block. **I consider this to be a blocker as well**. 

My suggestion for now would be to follow the already established pattern and expose new `status_error_log` and `status_access_log` properties (which can be disabled by default, I think that's fine). This way users have a clear and easy way to enable logging on this new interface.

---

#### Side note

In my opinion, the friendliest approach we could take given the explosion of new API interfaces exposed by Kong (Community and Enterprise) would be to include the `$server_name` variable in the default `access_log` format used by Kong, and log all API accesses in the same file by default.

E.g.:

```
logs/
├── admin_access.log
├── admin_error.log
├── proxy_access.log
└── proxy_error.log
```

Where all APIs (Admin, Status, etc...) write to `admin_access.log` by default, but including the server name (`kong_admin`, `kong_status`, etc...) is included in each log line. Users can then decide to split specific interfaces to their own files or disable them altogether (or even configure fancier `access_log` directives if desired).

spec/fixtures/nginx-directives.conf
```diff
@@ -5,3 +5,4 @@ nginx_stream_lua_shared_dict = custom_cache 5m
 nginx_proxy_proxy_bind = 127.0.0.1 transparent
 nginx_sproxy_proxy_bind = 127.0.0.1 transparent
 nginx_admin_server_tokens = off
+nginx_http_status_client_body_buffer_size=8k
```
thibaultcha: style: can we keep spaces around the `=` character? Thanks!

kong/status/init.lua
```diff
@@ -0,0 +1,43 @@
...
+
+
+-- Load plugins status routes
+if singletons.configuration and singletons.configuration.loaded_plugins then
+  for k in pairs(singletons.configuration.loaded_plugins) do
+    local loaded, mod = utils.load_module_if_exists("kong.plugins." ..
+                                                    k .. ".status_api")
+
+    if loaded then
+      print("Loading Status API endpoints for plugin: ", k)
```
thibaultcha: Can we keep the `ngx.log(ngx.DEBUG, ...)` logging level here instead of using `print`?

kong.conf.default
hbagdi: "interface" was chosen to keep it consistent with "Admin interface".
I've changed it to "Status API".

kong/templates/nginx_kong.lua
hbagdi: Agreed.
Updated the code to include `status_access_log` and `status_error_log` configuration properties which default to `off` and `logs/status_error.log` respectively.

PR #4981, opened 2019-08-29
fix(declarative) correct id generation on flat relationships

### LINKED ISSUE #4954: Error with jwt_secrets.key in kong 1.3
### Summary

I'm trying to update from 1.2 to 1.3 (db less), and this configuration prevent kong 1.3 to start: 
```
_format_version: "1.1"

consumers:
  - username: foo

jwt_secrets:
  - consumer: foo
    key: "https://keycloak/auth/realms/foo"
    algorithm: RS256
    rsa_public_key: "-----BEGIN PUBLIC KEY----------END PUBLIC KEY-----"

```
Error is 
```
nginx: [error] init_by_lua error: ...are/lua/5.1/kong/db/schema/others/declarative_config.lua:423: attempt to index local 'parent_fk' (a nil value)
stack traceback:
        ...are/lua/5.1/kong/db/schema/others/declarative_config.lua:423: in function 'get_key_for_uuid_gen'
        ...are/lua/5.1/kong/db/schema/others/declarative_config.lua:457: in function 'generate_ids'
        ...are/lua/5.1/kong/db/schema/others/declarative_config.lua:480: in function 'flatten'
        /usr/local/share/lua/5.1/kong/db/declarative/init.lua:153: in function 'parse_file'
        /usr/local/share/lua/5.1/kong/init.lua:252: in function 'parse_declarative_config'
        /usr/local/share/lua/5.1/kong/init.lua:430: in function 'init'
        init_by_lua:3: in main chunk
```

Removing `key: "https://keycloak/auth/realms/foo"` fixes error, but of course jwt validation doesn't work.

Best regards

hishamhm:
The uuid generator used by declarative config inadvertedly
assumed that the entities were nested.

Fixes #4954.

### DIFF
--- a/kong/db/schema/others/declarative_config.lua
+++ b/kong/db/schema/others/declarative_config.lua
@@ -412,19 +412,27 @@ local function get_key_for_uuid_gen(entity, item, schema, parent_fk, child_key)
   if schema.endpoint_key and item[schema.endpoint_key] ~= nil then
     local key = item[schema.endpoint_key]
 
-    -- If this item has foreign keys with on_delete "cascade", it is inferred
-    -- that its endpoint is not necessarily unique, so its key must be composed
-    -- by parent's key, avoiding that it is overwritten by identical endpoints
-    -- under other parents.
-    for _, field in schema:each_field(item) do
-      if field.type == "foreign" and field.on_delete == "cascade" then
-        local foreign_key_keys = all_schemas[field.reference].primary_key
-        for _, fk_pk in ipairs(foreign_key_keys) do
-          key = key .. ":" .. parent_fk[fk_pk]
+    -- check if the endpoint key is globally unique
+    if not schema.fields[schema.endpoint_key].unique then
+      -- If it isn't, and this item has foreign keys with on_delete "cascade",
+      -- we assume that it is unique relative to the parent (e.g. targets of
+      -- an upstream). We compose the item's key with the parent's key,
+      -- preventing it from being overwritten by identical endpoint keys
+      -- declared under other parents.
+      for fname, field in schema:each_field(item) do
+        if field.type == "foreign" and field.on_delete == "cascade" then
+          if parent_fk then
+            local foreign_key_keys = all_schemas[field.reference].primary_key
+            for _, fk_pk in ipairs(foreign_key_keys) do
+              key = key .. ":" .. parent_fk[fk_pk]
+            end
+          else
+            key = key .. ":" .. item[fname]
+          end
         end
-
       end
     end
+
     -- generate a PK based on the endpoint_key
     return pk_name, key
   end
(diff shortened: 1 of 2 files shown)


kong/db/schema/others/declarative_config.lua
```diff
@@ -412,19 +412,27 @@ local function get_key_for_uuid_gen(entity, item, schema, parent_fk, child_key)
...
-    -- that its endpoint is not necessarily unique, so its key must be composed
-    -- by parent's key, avoiding that it is overwritten by identical endpoints
-    -- under other parents.
-    for _, field in schema:each_field(item) do
-      if field.type == "foreign" and field.on_delete == "cascade" then
-        local foreign_key_keys = all_schemas[field.reference].primary_key
-        for _, fk_pk in ipairs(foreign_key_keys) do
-          key = key .. ":" .. parent_fk[fk_pk]
+    -- check if the endpoint key is globally unique
+    if not schema.fields[schema.endpoint_key].unique then
```
bungle: btw. I know places where `endpoint_key` refers to column that is not `unique` (never liked that):
https://github.com/Kong/kong/blob/next/kong/db/schema/entities/targets.lua#L23

Is that an issue?

hishamhm: It is precisely because of things like `target` that this code block exists.

The goal of this code is to generate UUIDs based on the entity contents to perform sensible PUT operations when using `db_import` on yaml files that don't specify explicit UUIDs.

We don't use the entire entity contents to generate the UUID because if we did that, every change to every argument would change the generated UUID, causing PUT to insert a new element.

Instead, we use the endpoint key as a "stable" identifier, and assume that if you keep the endpoint key the same but change another attribute, what you really want is an update.

When the `endpoint_key` is globally unique, we can simply use it to generate the UUID.

When it is not, we add to the UUID generator source the endpoint key plus the primary key of a "parent" entity (for example, for targets, we add upstreams). This works under the assumption that for a non-globally-unique endpoint key to be usable, it needs a parent entity to give it scoping, as is the case with `/username/U/target/T`.

The problem that this PR solves is that the implementation for that logic was assuming that the yaml file was written in "nested style", like this:

```
upstreams:
- name: myupstream1
   targets:
   - target: localhost:8080
- name: myupstream2
   targets:
   - target: localhost:8080
```

However, we also support a "flat style" like this:

```
upstreams:
- name: myupstream1
- name: myupstream2
targets:
- upstream: myupstream1
  target: localhost:8080
- upstream: myupstream2
  target: localhost:8080
```

(Both styles exist because, for plugins that map to multiple FKs such as route and a consumer at the same time, you can't nest. But this is beside the point of this PR.)

In issue #4954, the user tried to perform a mapping using `jwt_secrets` in the "flat style", which should be supported. Because `jwt_secrets` has a FK, the code in 1.3.0 incorrectly assumed that it was nested. That wasn't even necessary because the endpoint key for `jwt_secrets` _is_ unique, hence the check in line 416, so this PR fixes that.

When looking closer at the code I realized that the assumption of nestedness was also incorrect for non-unique endpoint keys (as in the example above), so I fixed that as well and added regression test cases for both `jwt_secrets` (which have globally-unique endpoint keys) and `targets` (which don't).

Hope that clarifies things! I think I will reuse this in the commit message when merging. :)

PR #4982, opened 2019-08-30
Wei/regex host

MichelLiu:

### Summary

Support for regex host

### Full changelog

* support for regex host

### Issues resolved

Support for regex host


hishamhm:
Hi, thank you for submitting a PR! Given that the router path is the most critical part of Kong, we generally are very careful when adding features to it, due to performance implications, usability, security, etc. We do have plans in the pipeline for router rule improvements, but we generally shied away from full regex support in host matching in the past (due to balancing performance and usability). We don't plan to add this immediately, so we're closing this PR for now. Thank you for understanding!

PR #4984, opened 2019-08-30
feat(key-auth) support auto-expiring keys through dao ttl

### LINKED ISSUE #87: [keyauth] Add auto expiration to keys

eskerda:
### Summary

Add support for keyauth credentials expiration by using metaschema ttl. Mainly based on https://github.com/Kong/kong/pull/4239, rebased and with the approach recommended on the comments.

The downside is the impossibility to differentiate between an invalid and an expired token. 

* Add dao support to return ttl on select queries for strategies
* Return entity ttl on keyauth cache callback
* Add API spec for listing ttl on keyauth admin endpoint

### Issues resolved

Fix #87

### DIFF
--- a/kong/plugins/key-auth/migrations/002_130_to_140.lua
+++ b/kong/plugins/key-auth/migrations/002_130_to_140.lua
@@ -27,6 +27,13 @@ return {
         -- Do nothing, accept existing state
       END$$;
 
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "keyauth_credentials" ADD "ttl" TIMESTAMP WITH TIME ZONE;
+      EXCEPTION WHEN DUPLICATE_COLUMN THEN
+        -- Do nothing, accept existing state
+      END$$;
+
     ]],
   },
   cassandra = {
--- a/kong/plugins/key-auth/migrations/002_130_to_140.lua
+++ b/kong/plugins/key-auth/migrations/002_130_to_140.lua
@@ -34,6 +34,13 @@ return {
         -- Do nothing, accept existing state
       END$$;
 
+      DO $$
+      BEGIN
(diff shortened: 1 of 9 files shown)


eskerda:
> Hi, this one looks good to me. since it's a new feature can you rebase against `next` instead of `master`?

Done that. Should I remove the XXX comments?

hishamhm:
> Should I remove the XXX comments?

Yes, please. (Unfortunately I don't have any suggestions of better ways to do the things noted in them. If anyone knows...)

eskerda:
> > Should I remove the XXX comments?
> 
> Yes, please. (Unfortunately I don't have any suggestions of better ways to do the things noted in them. If anyone knows...)

Done that. I replaced the XXX with more relevant comments, and in the case of the schema select field filtering added a more explicit check

kong/plugins/key-auth/migrations/002_130_to_140.lua
```diff
@@ -27,6 +27,13 @@ return {
         -- Do nothing, accept existing state
       END$$;
 
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "keyauth_credentials" ADD "ttl" TIMESTAMP WITH TIME ZONE;
+      EXCEPTION WHEN DUPLICATE_COLUMN THEN
+        -- Do nothing, accept existing state
+      END$$;
+
```
bungle: Do we need to start adding indexes on `ttl` column? I see we don't have one for OAuth 2.0 either.

eskerda: Added index on `ttl` column on `keyauth_credentials`, `oauth2_authorization_codes` and `oauth2_tokens`

Issue #4975, opened 2019-08-28
HTTP routing does not work for a service with the service mesh sidecar


encircled:
### Summary

I'm trying to setup a HTTP application with the kong sidecar, which will be exposed via kong gateway in kubernetes. Basically the same as https://github.com/Kong/kong-mesh-dist-kubernetes, but with HTTP service instead of raw TCP. 
Probably I'm missing the idea how exactly it is working internally, unfortunately the documentation is not comprehensive yet.

I've forked https://github.com/encircled/kong-mesh-dist-kubernetes and changed the serviceb manifest, the rest is the same. Serviceb now is a web app which exposes HTTP GET on 8080 port.

Default routing from the example is:

http --ignore-stdin put adminip:8001/services/service-b host=serviceb port=8080 protocol=tcp -f && \
http --ignore-stdin post adminip:8001/services/service-b/routes name=service-b sources[1].ip=0.0.0.0/0 protocols=tcp -f

I've also added the following http routing (+tcp default one):

curl -i -X PUT \
  --url http://adminip:8001/services/service-b-http \
  --data 'url=http://serviceb.test.svc.cluster.local:8080'

  curl -i -X POST \
  --url http://adminip:8001/services/service-b-http/routes \
  --data 'name=service-b-http' \
  --data 'paths=/test' \
  --data 'protocols=http'

And it does work: http://ingressip:8000/test

But if I remove the example TCP routing, the sidecar will fail with the following:

2019/08/27 13:27:45 [error] 35#0: *856 stream [lua] handler.lua:905: before(): no Route found with those values while prereading client data, client: xxx.xxx.xxx.xxx, server: 0.0.0.0:7000 │
│ 2019/08/27 13:27:45 [info] 35#0: *857 client xxx.xxx.xxx.xxx:38818 connected to 0.0.0.0:7000

where xxx.xxx.xxx.xxx is IP of kong ingress node

If I get it correctly, the sidecar is trying to connect via raw TCP and it will fail to match the route then. 
Moreover, https://github.com/Kong/kong-mesh-dist-kubernetes does not work at all with kong 1.3, probably due to this raw TCP communication and this commit which expects http/https https://github.com/Kong/kong/commit/6441e1d30fea18cfdfdd71214020889804f122b9

SUMMARY_GOES_HERE

### Additional Details & Logs

- Kong version 1.2.2
- Azure AKS - ubuntu

encircled:
[...]

PR #4989, opened 2019-09-02
feat(reports) add h2, h2c, tcp, and tls reports + removal of GRPC_MODES

bungle:
#### feat(reports) add h2, h2c, tcp, and tls reports 

- Adds stream reports and makes h2 and h2c reports available
- Also makes kong.reports safe for running it with stream module
- Removes excess code from plugins iterator where "reports" plugin was added to loaded plugins (it is not used) and moved reports initialization code to runloop

#### refactor(conf) change grpc proxy modes always to be just grpc

Removes need to have GRPC_PROXY_MODES, e.g. "grpc" and "grpcs", and makes the mode consistent on what we have for "http" where we don't have "https", "h2" or "h2c".

### Fixes

Fixes issue where `kong.reports` didn't work correctly on `stream`, as pointed out in this comment: https://github.com/Kong/kong/issues/4975#issue-486338205

### DIFF
--- a/spec/02-integration/05-proxy/21-reports_spec.lua
+++ b/spec/02-integration/05-proxy/21-reports_spec.lua
@@ -4,6 +4,9 @@ local websocket_client = require "resty.websocket.client"
 local cjson = require "cjson"
 
 
+local MESSAGE = "echo, ping, pong. echo, ping, pong. echo, ping, pong.\n"
+
+
 local function websocket_send_text_and_get_echo(uri)
   local payload = { message = "hello websocket" }
   local wc = assert(websocket_client:new())
@@ -35,6 +38,21 @@ for _, strategy in helpers.each_strategy() do
       admin_client:close()
     end
 
+    local reports_send_stream_ping = function()
+      ngx.sleep(0.01) -- hand over the CPU so other threads can do work (processing the sent data)
+
+      local tcp = require "socket".tcp()
+      assert(tcp:connect(helpers.get_proxy_ip(false), 19001))
+
+      -- TODO: we need to get rid of the next line!
+      assert(tcp:send(MESSAGE))
+
+      local body = assert(tcp:receive("*a"))
+      assert.equal("ok", body)
+
+      tcp:close()
+    end
+
     lazy_setup(function()
       dns_hostsfile = assert(os.tmpname())
       local fd = assert(io.open(dns_hostsfile, "w"))
@@ -51,8 +69,9 @@ for _, strategy in helpers.each_strategy() do
 
       local bp = assert(helpers.get_db_utils(strategy, {
         "services",
-        "routes"
-      }))
+        "routes",
+        "plugins",
+      }, { "reports-api" }))
 
       local http_srv = assert(bp.services:insert {
         name = "mock-service",
@@ -63,6 +82,7 @@ for _, strategy in helpers.each_strategy() do
       bp.routes:insert({ service = http_srv,
                          protocols = { "http" },
                          hosts = { "http-service.test" } })
+
       bp.routes:insert({ service = http_srv,
                          protocols = { "https" },
                          hosts = { "https-service.test" } })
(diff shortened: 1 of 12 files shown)


encircled:
get_current_suffix  looks good, but I'm not 100% sure if it fixes my initial problem with HTTP routing. As of now it fails on "no route fond" during preread phase, or is it caused by wrong protocol name as well?

bungle:
@encircled you are right, it does not fix the problem reported in #4975. But just something you pointed out in there. I'll investigate the rest of the issue and propose a separate PR on that. Thank you for heads up!

dndx:
@bungle tests are failing, would you mind taking a look? This change looks fine to me though.

bungle:
@dndx yes, they are failing because we merged one pr to next. I'll try to approve and merge this to get `next` green again: https://github.com/Kong/kong/pull/4990

bungle:
@dndx it is green now.

spec/02-integration/05-proxy/21-reports_spec.lua
```diff
@@ -223,16 +367,73 @@ for _, strategy in helpers.each_strategy() do
...
       assert.match("grpcs_reqs=0", reports_data[1])
       assert.match("ws_reqs=0", reports_data[1])
       assert.match("wss_reqs=1", reports_data[1])
     end)
+
+    it("#stream reports tcp streams", function()
+      local tcp = require "socket".tcp()
+      assert(tcp:connect(helpers.get_proxy_ip(false), 19000))
+
+      -- TODO: we need to get rid of the next line!
```
kikito: Can you clarify this TODO? I don't understand if "getting rid of the next line" mean "the next source code line" (`assert(tcp:send(MESSAGE))`), or "removing part of the body sent via TCP". I would like to know *why* why we would need to do any of those two (or a third one that I haven't imagined).

spec/02-integration/05-proxy/21-reports_spec.lua
```diff
@@ -223,16 +367,73 @@ for _, strategy in helpers.each_strategy() do
...
+        verify   = "none",
+        protocol = "any",
+      })
+
+      -- TODO: should SNI really be mandatory?
+      tcp:sni("this-is-needed.test")
+
+      assert(tcp:dohandshake())
+
+      -- TODO: we need to get rid of the next line!
```
kikito: Same: what does "next line" mean here

spec/02-integration/05-proxy/21-reports_spec.lua
```diff
@@ -223,16 +367,73 @@ for _, strategy in helpers.each_strategy() do
...
+
+      assert(tcp:connect(helpers.get_proxy_ip(true), 19443))
+
+      tcp = ssl.wrap(tcp, {
+        mode     = "client",
+        verify   = "none",
+        protocol = "any",
+      })
+
+      -- TODO: should SNI really be mandatory?
```
kikito: Couldn't this be figured out by commenting out `tcp:sni("this-is-needed.test")` and seeing if this test still passes?

spec/02-integration/05-proxy/21-reports_spec.lua
bungle: @kikito currently our `tcp` streams won't work, if you are not sending any data first. Don't remember the right number of bytes you need to send to get it to work.

spec/02-integration/05-proxy/21-reports_spec.lua
bungle: @kikito that same comment is again already from our other test suites. It is requirement that our stream has, and it is on TODO to remove that requirement.

spec/02-integration/05-proxy/21-reports_spec.lua
bungle: It means that we should not need it, e.g. the test should work even when that line is removed. That line is there to please our "restricted" stream support.

PR #3123, opened 2017-12-25
feat(admin) add latency headers in Admin API responses

hbagdi:
### Summary
Responses from Admin API now include a latency header,
which represents the time taken by Kong to respond to the
Admin request in milliseconds.

### Breaking change

* Nginx template for Kong
A `header_filter_by_lua_block` has been added to admin endpoint
to inject the latency header.

### Issues resolved

Fix #2855

PR #4996, opened 2019-09-04
feat(admin) add kong admin latency header

### LINKED ISSUE #2855: Show latency headers in Admin API responses
As of today latency is only displayed on the proxy port, but it's useful to display the metric on admin API responses as well.

bungle:
### Summary

This work was previously done by @hbagdi on #3123, but during that time we didn't end up with a conclusion and the PR got closed. This basically implements that now. Kong Admin API will add `X-Kong-Admin-Latency` header on admin api responses (which can be disabled with `KONG_HEADERS`).

**DO NOT MERGE YET, BUT PLEASE REVIEW AS THIS IS NOW BASED TO `refactor/latency-headers-with-fixes` (I will change it to be based on next after #4994 gets merged).**

### Issues resolved

Fix #2855


kong/templates/nginx_kong.lua
```diff
@@ -234,7 +234,7 @@ server {
     location / {
         default_type application/json;
         content_by_lua_block {
-            Kong.serve_admin_api()
```
hishamhm: Does this introduce a breaking change to anyone running Kong with a custom Nginx template? If so, I'd rather avoid the style change now (and possibly postpone it to 2.0).

Looking at the other commit it seems that if they fail to update the template they just don't get the latency header, which is harmless, but with this one they're forced to update the template otherwise the Admin API doesn't work.

Alternatively, we could add `kong.serve_admin_api = kong.admin_content` in the Kong code to make it backwards compatible?

kong.conf.default
```diff
@@ -325,16 +325,20 @@
                                  # - `X-Kong-Upstream-Latency`: Time taken
                                  #   (in milliseconds) by the upstream
                                  #   service to send response headers.
+                                 # - `X-Kong-Admin-Latency`: Time taken
+                                 #   (in milliseconds) by Kong to process
+                                 #   an admin api request.
```
hishamhm:
```suggestion
                                 #   an Admin API request.
```

kong/templates/nginx_kong.lua
bungle: Ok, I will make it an alias.

kong.conf.default
bungle: changed.

kong/templates/nginx_kong.lua
bungle: Alias made.

PR #5008, opened 2019-09-07
feat(response-transformer) add support to rename response headers

aalmazanarbs:
### Summary

Add support to rename response headers in response-transformer plugin

### Full changelog

- Modify the schema of the response-transformer plugin to add
configuration related to rename headers. In addition, a function to
validate headers names through Kong method has been added.
- Modify _M.transform_headers response-transformer plugin function to
rename response headers.
- Add test for the rename response headers in response-transformer plugin.

Since request-transformer plugin support modify rename headers it is
necessary that response-tranformer plugin has it. For example, map the
value to internal header and expose public header.


kong/plugins/response-transformer/header_transformer.lua
```diff
@@ -63,6 +63,15 @@ function _M.transform_headers(conf, headers)
       kong.response.clear_header(header_name)
   end
 
+  -- Rename headers(s)
+  for _, old_name, new_name in iter(conf.rename.headers) do
+    if headers[old_name] ~= nil and new_name then
+      local value = headers[old_name]
+      kong.response.set_header(new_name, value)
```
hishamhm: should this be `add_header` to cover the case where a header supports multiple entries?

e.g.

```
A: 1
B: 2

"rename": "B:A"

A: 1
A: 2
```

The "overwriting rename" could be achieved with

```
A: 1
B: 2

"remove": "A",  "rename": "B:A"

A: 2
```

How does that sound like?...

aalmazanarbs: I think we have to get the same behaviour of request-transformer plugin to avoid surprise for developers.

In the request-transformer plugin then rename function does:

```
req_set_header(new_name, value) // Set the current request's request header named header_name to value header_value, overriding any existing ones.
req_clear_header(old_name)
```

In the future we can review "rename" behaviour, but now I think we must keep both consistently.

hishamhm: @aalmazanarbs you're completely right, I forgot that request-transformer already had a `rename` feature! Agreed about consistency.

spec/03-plugins/15-response-transformer/03-api_spec.lua
```diff
@@ -48,6 +48,82 @@ for _, strategy in helpers.each_strategy() do
           local body = assert.response(res).has.jsonbody()
           assert.equals("just_a_key", body.config.remove.headers[1])
           assert.equals("just_a_key", body.config.remove.json[1])
+
+          admin_client:send {
+            method  = "DELETE",
+            path    = "/plugins/" .. body.id,
+          }
+        end)
+        it("rename succeeds without colons", function()
```
hishamhm: this should say `with colon` right?

aalmazanarbs: You are right, change done :)

hishamhm:
@aalmazanarbs you also need to the following on your local checkout:

```
git rebase origin/next
git push --force
```

aalmazanarbs:
Done but commits continues, what I am doing wrong?

hishamhm:
@aalmazanarbs Switch to the `next` branch locally, `git pull` to make sure you have the latest version of it, switch back to your PR branch, and then run `git rebase -i` to enter the interactive rebase screen. It will open your console text editor (whichever one is set via the `EDITOR` environment variable) and then you can mark `pick` or `drop` on each commit — _pick_ your commit and _drop_ the other ones (which are already in the `next` branch anyway). Save the file and exit the editor, it will rearrange the commits correctly (if you get an error message, fix the conflicts or run `git rebase --abort` to undo). Once the commits are rearranged, run `git push --force` to send the fixed state to Github.

Hope that helps!

aalmazanarbs:
I can not get to merge it correctly because my branch started from master and now it is a mesh :(. I cancel this PR and creates new one https://github.com/Kong/kong/pull/5040. The easy way :)

PR #5013, opened 2019-09-09
chore(ci) add the ability to use the Kong CI setup from external repositories

hutchic:
Purpose of this cut / paste is to relocate the setup CI logic into a Make task such that it can be reused in other repositories.

Applied sensible defaults in the bash script which can be overridden by environment variables to keep backwards compatibility

Reference: https://github.com/Kong/openresty-patches/pull/55

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,13 +26,8 @@ addons:
 
 env:
   global:
-    - LUAROCKS=3.1.3
-    - OPENSSL=1.1.1c
-    - OPENRESTY=1.15.8.1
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -48,12 +43,12 @@ env:
     - TEST_SUITE=pdk
 
 install:
+  - make setup-ci
   - source .ci/setup_env.sh
```
hutchic: We need to run the make task as well as `source` the bash script so the bash script can export the necessary environment variables which isn't possible via Make

PR #5014, opened 2019-09-10
fix(declarative) reduce events while loading declarative configs

locao:
When loading declarative configs, there were excessive events being fired up, overflowing lua-resty-worker-events queue, and losing some of them. 

This change creates a new keyword for upstreams and targets, which means that all of those entities must be processed, avoiding to create one event per entity.

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -298,41 +298,6 @@ local function remove_nulls(tbl)
 end
 
 
-local function nil_fn()
-  return nil
-end
-
-
-local function post_upstream_crud_delete_events()
-  local upstreams = kong.cache:get("upstreams|list", nil, nil_fn)
-  if upstreams then
-    for _, id in ipairs(upstreams) do
-      local ok = kong.worker_events.post("crud", "upstreams", {
-        operation = "delete",
-        entity = {
-          id = id,
-          name = "?", -- only used for error messages
-        },
-      })
-      if not ok then
-        kong.log.err("failed posting invalidation event for upstream ", id)
-      end
-    end
-  end
-end
-
-
-local function post_crud_create_event(entity_name, item)
-  local ok = kong.worker_events.post("crud", entity_name, {
-    operation = "create",
-    entity = item,
-  })
-  if not ok then
-    kong.log.err("failed posting crud event for ", entity_name, " ", entity_name.id)
-  end
-end
-
-
 function declarative.get_current_hash()
   return ngx.shared.kong:get("declarative_config:hash")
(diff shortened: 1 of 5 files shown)


kong/db/declarative/init.lua
```diff
@@ -306,29 +306,32 @@ end
...
-          id = id,
-          name = "?", -- only used for error messages
-        },
-      })
-      if not ok then
-        kong.log.err("failed posting invalidation event for upstream ", id)
-      end
+    local ok = kong.worker_events.post("crud", "upstreams", {
+      operation = "delete",
+      entity = "all",
```
hishamhm: Wasn't it possible to latch on to the `flip_config` event and get rid of these additional `crud` events here altogether?

locao: Yep, it makes sense. Will do that.

kong/runloop/handler.lua
```diff
@@ -417,7 +439,31 @@ local function register_events()
 
   if db.strategy == "off" then
     worker_events.register(function()
+      local ok = kong.worker_events.post("crud", "upstreams", {
+        operation = "delete",
+        entity = "all",
+      })
+      if not ok then
+        kong.log.err("failed posting invalidation event for all upstreams")
+      end
```
hishamhm: I may be wrong, but I've been thinking this: if `n` workers get a `flip_config` event, and then each of them posts a "delete all" event, won't this cause `n * n` "delete all" events in total?

(Maybe that's what's happening with `balancer` events nested inside `crud` events, which caused the burst to be extra problematic? I have vague memories about this.)

I'm now thinking that it might be more efficient to post these events from `declarative.load_into_cache_with_events` instead to avoid the O(n^2), and post them directly to `worker_events.post("balancer", "upstreams", ...)` bypassing `"crud"` altogether (the caches are clear on a DB config flip anyway).

I was even going to suggest calling `balancer.on_upstream_event` directly here, but at least for the creation of balancers is good to do them async (as some incur in DNS delays; in `kong.init`, `balancer.init` happens inside a timer for this reason), so posting `"balancer"` worker events from `declarative.load_into_cache_with_events` is probably the better option.

(Also, non-blocker side comment: I know it's a matter of taste and my love of types :sweat_smile:, but I'd probably make `delete_all` a new operation for `balancer.on_upstream_event` instead of making `"all"` a magic `upstream_data` value)

locao: It makes sense, I made the changes to fix that. Also, I think you're right about the magic `upstream_data` 😳. Also fixed that

PR #5019, opened 2019-09-11
fix(declarative) Lua input parsing and error handling

Tieske:
This borrows from JSONschema and OpenAPI. Any value can now have a `$ref` entry, eg. json:

```json
  "my_key": { "$ref": "#/components/some/value" }
```
A new top level entry `components` is introduced to hold reusable objects and values.

An additional commit fixes the Lua declarative input parsing and error handling.

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -101,12 +101,14 @@ function Config:parse_string(contents, filename, accept, old_hash)
     dc_table, err = cjson.decode(contents)
 
   elseif accept.lua and filename:match("lua$") then
-    local chunk = loadstring(contents)
-    setfenv(chunk, {})
+    local chunk, pok
+    chunk, err = loadstring(contents)
     if chunk then
-      local pok, dc_table = pcall(chunk)
+      setfenv(chunk, {})
+      pok, dc_table = pcall(chunk)
       if not pok then
         err = dc_table
+        dc_table = nil
       end
     end
 


hishamhm:
First, the conceptual review:

I'm very much _not_ a fan of this approach. It starts building a mini ad hoc programming language on top of the declarative config format. This is the kind of thing that is best left to configuration front-end tools, like [Dhall](https://dhall-lang.org/), [ytt](https://developer.ibm.com/blogs/yaml-templating-tool-to-simplify-complex-configuration-management/) and I'm sure there are others. Backreferences as implemented in this PR is only one of the various things that people do with configuration front-end tools: variable interpolation, conditionals and loops are others.

Then, the implementation review:

The code itself looks fine, but if this were to be added, then the `_format_version` number needs to be bumped and version compatibility checks need to be added to the code. Also, the top-level `components` entry is a breaking change because a custom plugin could have an entity with that name (metadata fields need a preceding `_`).

The Lua fix commit is welcome — I casually hit into that just now when looking at #5021. `.lua` is currently supported by `db_import` (modulo the brokenness you fixed) but not by DB-less (though we could revisit that decision: it was mostly a lack of a decision to include it at the time, not a decision to exclude it).

hbagdi:
@Tieske Why is this needed in the first place though?
Can YAML's de-referencing not solve most (not all) of these duplication problems?

PR #5038, opened 2019-09-16
fix(chore) update code comment

Tieske:


kong/runloop/handler.lua
```diff
@@ -687,7 +687,7 @@ do
     local balancer_data = {
       scheme         = scheme,    -- scheme for balancer: http, https
       type           = host_type, -- type of 'host': ipv4, ipv6, name
-      host           = host,      -- target host per `upstream_url`
+      host           = host,      -- target host per `service` entity
```
thibaultcha: This value can come from one of three places as of the current state:

1. The Service's `host` attribute.
2. The request (when the associated Route's `preserve_host` is enabled or when proxying a Service without an associated Route in transparent proxying).
3. Kong's server address if we are transparently proxying and if the request's Host header couldn't be parsed.

Updating the comment here is fine as-is (although it could be more accurate) but I must say that I question the usefulness of the comments in this structure in the first place.

Tieske: imo they are useful while pdk functions to set them are absent.

btw: the value is the target host pre-dns-resolution, and has nothing to do with the Host header. The Host header is passed on ngx side as a variable afaik, not in Lua-land. This should change though, once we move the construction of the destination properties from pre-access to post-access.

thibaultcha: No, this value can definitely come from the request in service-less Routes.

Tieske: what's the suggested wording then?

thibaultcha: I'd personally just remove them as they bring little value imho, some comments aren't worth maintaining. But no big deal, we'll just merge this as-is.

PR #5040, opened 2019-09-16
feat(response-transformer) add support to rename response headers

aalmazanarbs:
### Summary

Add support to rename response headers in response-transformer plugin

### Full changelog

- Modify the schema of the response-transformer plugin to add
configuration related to rename headers. In addition, a function to
validate headers names through Kong method has been added.
- Modify _M.transform_headers response-transformer plugin function to
rename response headers.
- Add test for the rename response headers in response-transformer plugin.

Since request-transformer plugin support modify rename headers it is
necessary that response-tranformer plugin has it. For example, map the
value to internal header and expose public header.


hishamhm:
Merged! Thank you @aalmazanarbs !

PR #5047, opened 2019-09-18
feat(db) dao transformations

### LINKED ISSUE #4893: bug: Patch request for basic-auth credential returns 404 incorrectly
### Summary

When password specified in a patch request is same as the one currently stored in DB(hashed), patch request return a 404.
This could be due to the hashing logic on the password field.

### Steps To Reproduce

```bash
$ http :8001/consumers username=foo
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 115
Date: Mon, 12 Aug 2019 15:39:19 GMT
Server: kong/1.3.0rc1

{
    "created_at": 1565624359,
    "custom_id": null,
    "id": "6a3c317e-daa9-4934-89be-10fc4a04e703",
    "tags": null,
    "username": "foo"
}

$ http post :8001/consumers/foo/basic-auth username=bar
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 197
Date: Mon, 12 Aug 2019 15:39:32 GMT
Server: kong/1.3.0rc1

{
    "consumer": {
        "id": "6a3c317e-daa9-4934-89be-10fc4a04e703"
    },
    "created_at": 1565624372,
    "id": "8b80904a-68c9-432c-a7e9-f308d5ab860a",
    "password": "ea809177afb40b62a34582080b8a876b9056cbf5",
    "username": "bar"
}

# Please note that specifying the hashed password as the password should again hash it

$ http patch :8001/consumers/foo/basic-auth/8b80904a-68c9-432c-a7e9-f308d5ab860a password=ea809177afb40b62a345820
80b8a876b9056cbf5 username=new-bar
HTTP/1.1 404 Not Found
Connection: keep-alive
Content-Length: 23
Date: Mon, 12 Aug 2019 15:40:43 GMT
Server: kong/1.3.0rc1

{
    "message": "Not found"
}

$ http patch :8001/consumers/foo/basic-auth/8b80904a-68c9-432c-a7e9-f308d5ab860a password=any-other-password username=new-bar
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 201
Date: Mon, 12 Aug 2019 15:45:31 GMT
Server: kong/1.3.0rc1

{
    "consumer": {
        "id": "6a3c317e-daa9-4934-89be-10fc4a04e703"
    },
    "created_at": 1565624372,
    "id": "8b80904a-68c9-432c-a7e9-f308d5ab860a",
    "password": "bda1183766a0e94b1ea14f7733f9c21ffade8805",
    "username": "new-bar"
}
```

### Additional Details & Logs

- Kong version (`$ kong version`) 1.3.0rc1

bungle:
### Summary

This PR contains 3 commits:

1. feat(schema) add transformations property to schema

This PR adds a schema property so that you can define schema level input transformations:

```lua
{
  name = "test",
  primary_key = { "test" },
  fields = {
    {
      test = {
        type = "string",
      },
    },
    {
      test2 = {
        type = "string",
      },
    },
  },
  transformations = {
    {
      input = { "test" },
      needs = { "test2" },
      on_write  = function(test, test2)
        return { test = test .. test2 }
      end,
    },
  },
}
```

2. feat(db) implement dao transformations

Implements schema-level transformations with the auto-generated DAO methods. E.g. makes the previous commit in action in DAOs.

3. refactor(basic-auth) use dao transformations with basic auth credentials

Also removes the need to use custom dao. This works as a showcase for dao-transformations.

### Other Notes

At this point it implements write-transformations, e.g. when database rows are inserted / updated. In a future we will need several things to be done:

1. read-transformations (e.g. encryption + decryption)
2. add an option `{ disable_transformations = true }` to be able to disable transformations when calling DAO methods (may be needed for importing a declarative config).
3. implement transformations with dbless.

But this PR provides the basis for the transformations work, and as far as I know it doesn't close any doors for further development. Ultimately we get things like encryption at rest, generic hashing features etc.

### Issues resolved

Fix #4893


spec/01-unit/01-db/01-schema/01-schema_spec.lua
```diff
@@ -2989,4 +3015,249 @@ describe("schema", function()
...
+
+      local TestEntities = Schema.new(test_schema)
+      local transformed_entity, _ = TestEntities:transform(entity)
+
+      assert.truthy(transformed_entity)
+      assert.equal("test1", transformed_entity.name)
+    end)
+
+
+    it("transforms fields with needs given to function", function()
```
kikito: I am unclear as to why we need `needs`. It seems this specific case could be handled inside the transformation itself with an `if`:

```
        transformations = {
          {
            input = { "name", "age" },
            func = function(name, age)
              if age then
                return { name = name .. " " .. age, age = age }
              else
                return { name = name }
              end
            end,
          },
        },
```

spec/01-unit/01-db/02-db-errors_spec.lua
```diff
@@ -292,6 +292,35 @@ describe("DB Errors", function()
     end)
 
 
+    describe("TRANSFORMATIO_ERROR", function()
```
kikito: TYPO: TRANSFORMATIO(N)

spec/01-unit/01-db/01-schema/01-schema_spec.lua
```diff
@@ -2989,4 +3015,249 @@ describe("schema", function()
...
+      local entity = { name = "Bob" }
+
+      local TestEntities = Schema.new(test_schema)
+      local transformed_entity, _ = TestEntities:transform(entity)
+
+      assert.truthy(transformed_entity)
+      assert.equal("How are you Bob?", transformed_entity.name)
+    end)
+
+    it("returns error if transformation returns an error", function()
```
kikito: I think one more tests similar to this one is needed, but for a transformation function which (erroneously) returns a non-table (like a number, or a string). This case should be detected and handled too.

spec/01-unit/01-db/01-schema/01-schema_spec.lua
```diff
@@ -2989,4 +3015,249 @@ describe("schema", function()
...
+
+      local TestEntities = Schema.new(test_schema)
+      local transformed_entity, _ = TestEntities:transform(entity, input)
+
+      assert.truthy(transformed_entity)
+      assert.equal("test1", transformed_entity.name)
+    end)
+
+
+    it("transforms fields with multiple transformations", function()
```
kikito: I think we need one more test which ensures that a transformation can modify *other fields* besides the input ones. For example:

```
          {
            input = { "name" },
            func = function(name)
              return { slug = to_slug(name) }
            end,
          },

```

This would use `name` to modify the `slug` field. Is this allowed, or disallowed?

kong/db/dao/init.lua
```diff
@@ -328,12 +334,18 @@ local function check_update(self, key, entity, options, name)
...
   end
 
-  local ok, errors = self.schema:validate_update(entity_to_update)
+  local ok, errors = self.schema:validate_update(entity_to_update, entity)
   if not ok then
     local err_t = self.errors:schema_violation(errors)
     return nil, nil, tostring(err_t), err_t
   end
 
+  entity_to_update, err = self.schema:transform(entity_to_update, entity)
```
kikito: It seems that transforms happen immediately on `insert` and `update`, but never on `read`. Should we have an `on` = insert | update | read field in transforms?

spec/01-unit/01-db/01-schema/01-schema_spec.lua
```diff
@@ -2989,4 +3015,249 @@ describe("schema", function()
...
+            input = { "non_existent" },
+            func = function(non_existent)
+              return { name = non_existent:upper() }
+            end,
+          },
+        },
+      }
+      local entity = { name = "test1" }
+
+      local TestEntities = Schema.new(test_schema)
```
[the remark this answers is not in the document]
bungle: it would fail on metaschema, no?

kong/db/dao/init.lua
bungle: `on = read, write, insert, update, upsert...` OR just `on = read, write`. Perhaps... I don't know. This does not implement read-transformations, and doesn't differentiate on write operations at this point. But I guess same comment applies here, why not just pass `context` to transformation function and let the function decide. It reminds a lot of that old `self_check` that we used to have, e.g. we could call it `mutate_and_or_validate` and here is the whole entity and a context.

spec/01-unit/01-db/01-schema/01-schema_spec.lua
bungle: This leads to interesting question also... should we validate again? Or at least check that it only contains right fields? Though, I think can skip it for now and just trust the transformation writer doing the right thing?

kong/db/schema/init.lua
```diff
@@ -1671,6 +1692,7 @@ end
 -- @param full_check If true, demands entity table to be complete.
 -- If false, accepts missing `required` fields when those are not
 -- needed for global checks.
+-- @param user_input The original user input for transformation validatations.
```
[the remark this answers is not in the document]
bungle: fixed thos.

spec/01-unit/01-db/02-db-errors_spec.lua
bungle: fixed that typo.

spec/01-unit/01-db/01-schema/01-schema_spec.lua
bungle: this is far from the only thing that a function can do, e.g. it can:
1. yield
2. error
3. call ngx.exit (or something similar)
4. return garbage
5. halt execution
6. os.exit

So, I am not sure do we need to handle that? For better error messages on Nginx logs? If we handle that should we also handle some other cases mentioned above?

spec/01-unit/01-db/01-schema/01-schema_spec.lua
bungle: The test was added that checks that it can modify any field. In fact it can even introduce new fields. We don't validate after transformations (do not know if we should),

hishamhm: We should not validate again, because in some use-cases transformations can intentionally produce data that is non-conforming as an input.

Since transformations are part of schema definitions and are written by the Kong/plugin developer persona and not the Kong operator persona, there's a greater level of "they should know what they're doing" expected here. In the end, since this is an arbitrary Lua function, nothing prevents them from doing anything silly (like stuff with side-effects which would break the idempotency of schema validation, or even `os.execute("sudo rm -rf /*")` :stuck_out_tongue_winking_eye: ).

> In fact it can even introduce new fields.

This would fail at the database level later, and I think it's fine, because it's a programmer error that would explode quite visibly on testing. It doesn't seem worth adding runtime checks for this just to give a different error message.

spec/01-unit/01-db/01-schema/01-schema_spec.lua
bungle: Yes, indeed we could have:
```
transformation = function(context, entity, original_input)
  return { <new-entity> }
end
```

Just a general description:
1. `transformation.input` adds `mutually_required` check over the fieds defined in `transformation.input`, kinda an implicit `entity check` for `schema:validate` that does validation against the original (user) input, the normal entity validations work on entity that could have been merged with rbw entity, or which foreign keys where resolved, or autofields filled.
2. `transformation.input` is also used as a deciding factor with `schema:transform` to detect if this transformation needs to be executed (e.g. all input fields where given in original (user) input). Kinda like second time check but this time used for triggering the transformation and not for adding implicit validation.
3. `transformation.needs` are the extra parameters that the function needs as a function arguments that may or may not be on the entity (not the original user input). If those fields are not specified, it means the transformation is skipped. Kinda like soft `transformation.input`. It never adds validation, it is just about skipping transformation or executing it. The needs are added at the end of the arguments list when transformation is called.

hishamhm:
> My biggest gripe is that I am not sure we need a needs field at all.

@kikito We definitely do. There are two types of dependencies for a transformation.

* `input = { "foo", "bar" }` - run if `foo` and `bar` are present, fail if `foo` is present but `bar` isn't, fail if `bar` is present but `foo` isn't. `foo` and `bar` are **co-dependent**.
* `input = { "foo" }, needs = { "bar" }` - run if `foo` and `bar` are present, fail if `foo` is present but `bar` isn't, ignore if `bar` is present but `foo` isn't. `foo` **depends on** `bar` but `bar` does not depend on `foo`.

These decisions need to be made by the execution engine _declaratively_ like this, and not inside the code of `func` — we've seen what happens to partial updates when this is left to the callback code in the old DAO (people got the behavior of partial updates wrong all the time, including in core entities).

Hope that clarifies!

bungle:
> * `input = { "foo" }, needs = { "bar" }` - run if `foo` and `bar` are present, fail if `foo` is present but `bar` isn't, ignore if `bar` is present but `foo` isn't. `foo` **depends on** `bar` but `bar` does not depend on `foo`.

I think I have implemented it differently:
1. skip if foo is not present (because it will only fail with "some present", and with single input all is present or none is present
2. skip if "bar" is not present

But sure, we could change:
2. fail if `bar` is not present, I think it will work too.

hishamhm:
> fail if bar is not present

@bungle yes, we need this behavior, otherwise the database could end up containing both transformed and non-transformed values of `foo` depending if `bar` was given or not in PATCH operations.

spec/01-unit/01-db/01-schema/01-schema_spec.lua
bungle: @hishamhm gave a better description below. I will update the code to match that.

bungle:
> @bungle yes, we need this behavior, otherwise the database could end up containing both transformed and non-transformed values of `foo` depending if `bar` was given or not in PATCH operations.

The code is now changed to work so that it fails correctly on validation if needs are not fulfilled. The `schema:transform` does not do any validations or checks, so it means `schema:validate[<_*>]` needs to be called before (this is what our autogenerated daos do).

spec/01-unit/01-db/01-schema/01-schema_spec.lua
```diff
@@ -1804,6 +1804,73 @@ describe("schema", function()
...
+          { input = { "a2" },       func = function() return {} end },
+          { input = { "a1", "a3" }, func = function() return {} end },
+        }
+      })
+
+      local ok, err = Test:validate_update({
+        a1 = "foo"
+      })
+      assert.is_falsy(ok)
+      assert.match("all or none of these fields must be set: 'a1', 'a3'", err["@entity"][1])
```
hishamhm: Could you confirm what happens in this case?

```
      local ok, err = Test:validate_update({
        a1 = "aaa",
        a2 = "bbb",
        a3 = "ccc",
        a4 = "ddd",
      }, {
        a1 = "foo"
      })
```

If I follow correctly, the first table is the full entity obtained via read-before-write and the second table is the input given by the user with `PATCH`, right?

This needs to fail. Imagine for example that `a1` is `algorithm` and `a3` is `password` and the transformation hashes the password. If this passes, it will leave `algorithm` and `password` inconsistent with each other. Just double-checking.

hishamhm:
I was reviewing the transformations PR and things got a bit more complicated in my head. I was looking at the places in the code where it does `original_input or input` and got confused. 

I was thinking of that use-case with `algorithm` and `password`, and it seems that the ideal workflow is not one we're fully supporting yet:

• if I change the `algorithm`, I need to give the `password` to recompute it (otherwise I end up with a password with an incompatible hash in the database)
• if I change the `password`, I need the `algorithm` (but it could come from the RBW, not from the PATCH input) 

This almost sounds like two relationships, but it really is one. `password` depends on `algorithm`. So, when P depends on A, this means that when A updates, you need to give P again. But when P updates, you can use A from the database to recompute it.

This means that when we add the transformation with `input = { "password" }, needs = { "consumer.id" }` in basicauth_credentials, we need the `password` to be present whenever we PATCH `consumer.id`, otherwise we'll end up with an unmatchable password.

Argh, partial updates complicate everything.

hishamhm:
I just gave it a try to check the behavior of basicauth_credentials for partial updates, and this fails on both `master` and this PR:

```diff
diff --git a/spec/03-plugins/10-basic-auth/03-access_spec.lua b/spec/03-plugins/10-basic-auth/03-access_spec.lua
index e655c0bb1..348f17356 100644
--- a/spec/03-plugins/10-basic-auth/03-access_spec.lua
+++ b/spec/03-plugins/10-basic-auth/03-access_spec.lua
@@ -7,6 +7,9 @@ local utils   = require "kong.tools.utils"
 for _, strategy in helpers.each_strategy() do
   describe("Plugin: basic-auth (access) [#" .. strategy .. "]", function()
     local proxy_client
+    local admin_client
+
+    local consumer2
 
     lazy_setup(function()
       local bp = helpers.get_db_utils(strategy, {
@@ -21,6 +24,10 @@ for _, strategy in helpers.each_strategy() do
         username = "bob",
       }
 
+      consumer2 = bp.consumers:insert {
+        username = "obo",
+      }
+
       local anonymous_user = bp.consumers:insert {
         username = "no-body",
       }
@@ -72,6 +79,12 @@ for _, strategy in helpers.each_strategy() do
         consumer = { id = consumer.id },
       }
 
+      bp.basicauth_credentials:insert {
+        username = "will_change_consumer",
+        password = "password123",
+        consumer = { id = consumer.id },
+      }
+
       bp.plugins:insert {
         name     = "basic-auth",
         route = { id = route3.id },
@@ -94,6 +107,7 @@ for _, strategy in helpers.each_strategy() do
       }))
 
       proxy_client = helpers.proxy_client()
+      admin_client = helpers.admin_client()
     end)
 
 
@@ -101,6 +115,9 @@ for _, strategy in helpers.each_strategy() do
       if proxy_client then
         proxy_client:close()
       end
+      if admin_client then
+        admin_client:close()
+      end
 
       helpers.stop_kong()
     end)
@@ -217,6 +234,44 @@ for _, strategy in helpers.each_strategy() do
         assert.equal('bob', body.headers["x-consumer-username"])
       end)
 
+      it("authenticates valid credentials in Authorization after changing consumer", function()
+        local res = assert(proxy_client:send {
+          method  = "GET",
+          path    = "/request",
+          headers = {
+            ["Authorization"] = "Basic d2lsbF9jaGFuZ2VfY29uc3VtZXI6cGFzc3dvcmQxMjM=",
+            ["Host"]          = "basic-auth1.com"
+          }
+        })
+        local body = cjson.decode(assert.res_status(200, res))
+        assert.equal('bob', body.headers["x-consumer-username"])
+
+        local res2 = assert(admin_client:send {
+          method  = "PATCH",
+          path    = "/basic-auths/will_change_consumer",
+          body    = {
+            consumer = {
+              id = consumer2.id
+            }
+          },
+          headers = {
+            ["Content-Type"] = "application/json",
+          }
+        })
+        assert.res_status(200, res2)
+
+        local res3 = assert(proxy_client:send {
+          method  = "GET",
+          path    = "/request",
+          headers = {
+            ["Authorization"] = "Basic d2lsbF9jaGFuZ2VfY29uc3VtZXI6cGFzc3dvcmQxMjM=",
+            ["Host"]          = "basic-auth1.com"
+          }
+        })
+        local body = cjson.decode(assert.res_status(200, res3))
+        assert.equal('obo', body.headers["x-consumer-username"])
+      end)
+
       it("authenticates with a password containing ':'", function()
         local res = assert(proxy_client:send {
           method = "GET",
```

Both branches accept the partial update (give 200 on `res2`), but then end up with an unusable credential (give 401 on `res3`). They should instead have rejected that partial update (failed on `res2`).

spec/01-unit/01-db/01-schema/01-schema_spec.lua
bungle: It fails with:
```
all or none of these fields must be set: 'a1', 'a3'
```

bungle: First table is a table that has fields from `process_auto_fields` and in case it has records, it also is merged with `rbw`. Second table is what user gave.

kong/db/dao/init.lua
```diff
@@ -328,12 +334,18 @@ local function check_update(self, key, entity, options, name)
     return nil, err, err_t
   end
 
-  local ok, errors = self.schema:validate_update(entity_to_update)
+  local ok, errors = self.schema:validate_update(entity_to_update, entity, rbw_entity)
```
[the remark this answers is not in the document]
bungle: I think this gurantees it:
https://github.com/Kong/kong/pull/5047/files#diff-7135c27fc8fde710af0a85f4dc622d13R1546

PR #5051, opened 2019-09-19
fix(file-log) remove unnecessary timer and cache hot path functions

dndx:
Because the `log` function has no yielding calls, running it inside timer context is a waste of CPU cycles.

### DIFF
--- a/kong/plugins/file-log/handler.lua
+++ b/kong/plugins/file-log/handler.lua
@@ -4,7 +4,6 @@ local cjson = require "cjson"
 local system_constants = require "lua_system_constants"
 local basic_serializer = require "kong.plugins.log-serializers.basic"
 
-local ngx_timer = ngx.timer.at
 local O_CREAT = system_constants.O_CREAT()
 local O_WRONLY = system_constants.O_WRONLY()
 local O_APPEND = system_constants.O_APPEND()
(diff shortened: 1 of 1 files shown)


kong/plugins/file-log/handler.lua
```diff
@@ -63,13 +66,9 @@ FileLogHandler.PRIORITY = 9
 FileLogHandler.VERSION = "2.0.0"
 
 function FileLogHandler:log(conf)
-  local message = basic_serializer.serialize(ngx)
-
-  local ok, err = ngx_timer(0, log, conf, message)
-  if not ok then
-    ngx.log(ngx.ERR, "[file-log] failed to create timer: ", err)
-  end
+  local message = serialize(ngx)
```
thibaultcha: We should bump the plugin version as well (just above this line)

dndx: Thanks, fixed!

kong/plugins/file-log/handler.lua
```diff
@@ -27,49 +30,41 @@ local file_descriptors = {}
...
 
+  log(conf, message)
```
Tieske: since this is (for good reasons) deviating from the other log plugin implementations, should we add a small comment as to why there are no timers here?

dndx: I would say that seems to be unnecessary. There is no "standard" implementation for log plugins and for the most part this PR is trying to fix an unintentional copy-n-paste error, which shouldn't warrant a comment to explain why.

If anything, we should comment the ones that created the timer instead to explain why they are necessary in the first place.

p0pr0ck5: > fix an unintentional copy-n-paste error, which shouldn't warrant a comment to explain why.

Wouldn’t the presence of this error itself imply that a comment is warranted?

PR #5055, opened 2019-09-20
feat(conf) add handle_errors to config

onematchfox:
### Summary

We would like to able to present errors resulting from Kong (and ultimately upstreams) in a consistent manner. 

Currently, internal/NGINX errors are handled via https://github.com/Kong/kong/blob/next/kong/error_handlers.lua through use of the `error_page` directives at https://github.com/Kong/kong/blob/ce43f03d67c52c1b8612240cfce6e26438ffb783/kong/templates/nginx_kong.lua#L87-L88

This nicely wraps up the error and returns it in the format specified via the `accept` header. However, any plugins, including core Kong code (e.g. route matching) make use of `kong.response.exit()` instead which results in the response being returned directly to the calling application through use of `ngx.print(body)`. 

For our own plugins, we could adjust the response we send via `kong.response.send` however, it is unfeasible to do this for all responses on all plugins we may use. Thus, for us to be able to handle, and return, errors to a calling application in a consistent manner we would need to override both `error_handlers.lua` and `pdk/response.lua`. By introducing a new config item `handle_errors` where, when set, Kong would not make use of `ngx.print`, we can ensure that any HTTP error (set in the `error_page` directives) are ultimately handled by `error_handlers.lua`. `error_handlers.lua` can, in turn,  easily be customised and overriden to provide a consistent error response for any 400+ status generated within Kong. 

In addition, it becomes fairly trivial to handle upstream errors in the same fashion through use of a plugin operating on `header_filter`. E.g. 

```
function ProxyErrorsHandler:header_filter(conf)
  ProxyErrorsHandler.super.header_filter(self)

  local status = kong.response.get_status()
  if kong.configuration.handle_errors and status >= 400 and not ngx.ctx.ERROR_HANDLED then
    local error_message = kong.response.get_header(conf.custom_error_header)

    if error_message then
      ngx.ctx.error_body = { message = error_message }
    end

    return ngx.exit(status)
  end
end
```

In doing so, this enables us present consistent errors for all requests proxied via Kong and, to be able to pass error context from the upstream to a standard error handler - something that is not possible using the standard nginx directive `proxy_intercept_errors`.

### Full changelog

* Add boolean config option `handle_errors` - defaulted to `false/off`
* Before sending response, check `handle_errors` config. If set to `true` and `ngx.config.subsystem == "http"` and `status >= 400` then set `ctx.error_body` and `ctx.error_headers` and exit without writing to response.
* Attempt to read `ctx.error_body` and `ctx.error_headers` within error handling and set message and headers accordingly if these are populated.

### DIFF
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -236,6 +236,8 @@ local CONF_INFERENCES = {
   lua_ssl_verify_depth = { typ = "number" },
   lua_socket_pool_size = { typ = "number" },
   service_mesh = { typ = "boolean" },
+
+  handle_errors = { typ = "boolean" },
 }
 
 
--- a/kong/error_handlers.lua
+++ b/kong/error_handlers.lua
@@ -57,7 +57,18 @@ return function(ngx)
   end
 
   local status = ngx.status
-  message = BODIES["s" .. status] or format(BODIES.default, status)
+
+  -- Attempt to read body (message) and headers from ctx for internal/plugin errors generated via kong.response.exit(...)
+  local data = ngx.ctx.error_body
+  message = data and data.message or data or BODIES["s" .. status] or format(BODIES.default, status)
+
+  ngx.ctx.ERROR_HANDLED = true
+
+  if not ngx.headers_sent and type(ngx.ctx.error_headers) == 'table' then
+    for k,v in pairs(ngx.ctx.error_headers) do
+      ngx.header[k] = v
+    end
+  end
 
   if singletons.configuration.enabled_headers[constants.HEADERS.SERVER] then
     ngx.header[constants.HEADERS.SERVER] = SERVER_HEADER
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -485,6 +485,12 @@ local function new(self, major_version)
       error("headers have already been sent", 2)
     end
 
+    if self.configuration.handle_errors and self.ctx.core.phase ~= phase_checker.phases.admin_api and ngx.config.subsystem == "http" and status >= 400 then
+      ngx.ctx.error_body = body
+      ngx.ctx.error_headers = headers
+      return ngx.exit(status)
+    end
+
     ngx.status = status
 
     if self.ctx.core.phase == phase_checker.phases.admin_api then
--- a/kong/templates/kong_defaults.lua
+++ b/kong/templates/kong_defaults.lua
@@ -97,4 +97,6 @@ lua_ssl_trusted_certificate = NONE
 lua_ssl_verify_depth = 1
 lua_package_path = ./?.lua;./?/init.lua;
 lua_package_cpath = NONE
+
+handle_errors = false
 ]]


onematchfox:
A couple notes for the time being pending initial review of concept:

1. I am aware that I haven't added tests for this. Would be happy to do so, if someone can point me in the right direction
1. Documentation updates will also be required?

onematchfox:
@thibaultcha any chance you've managed to give this PR some thought?

onematchfox:
@hishamhm, @kikito, @bungle - sorry to ping all of you but not sure who to contact here. Am sure everyone has lots on their plate but the lack of any form of feedback/commentary to this PR after 1.5 months makes me wonder if it has slipped through the cracks? Can anyone provide some form of feedback on this request? Is the feature something you would consider, or perhaps are even looking into but would rather take a different approach? Or, is this just a hard no?

onematchfox:
@Tieske, sorry to be spamming you today, but this one came up again during my upgrade to 1.4.0 and the attempt to get rid of any forked code. Back at the summit we had a discussion regarding community engagement and this PR as I wanted to try to understand how you and the rest of the team wanted/preferred to interact with community. Your suggestion to me back then was to ping folks if time dragged on. I have now tried that and have run out of ideas once again. All I'm really looking for is acknowledgement from someone at Kong that this is either not something you're interested in or something you're interested in but need time to get to it or, even, would rather take a different/your own approach. At least then I can weigh up my own options going forward. But the complete lack of any form of communication around this PR after almost 2 months is frustrating and somewhat of a mystery to me and so I'm hoping you can perhaps help me resolve that?

r1ckr:
I can't believe nobody is looking at this PR, something so fundamental for an API GW.

bungle:
@r1ckr, @onematchfox, we have not forgotten about this. We were quite busy at releasing `2.0`, and this didn't get scoped for that. Now that we have that out, I think we might have some time to go through community PRs.

I agree that current Kong error reporting is inconsistent. Also our error handling is inconsistent. E.g. we do catch `runtime errors` on `access phase`, but do not catch them on `header_filter`. So `runtime-errors` happening on `header_filter` will just lead `nginx` prematurely closing the client connection and ending up with `closed state`. I have added this to our backlog.

This seems to be touching many places in our code. The runloop, the plugin iteration and execution, etc. To get this correct, I think we need to study the issues better on our side first. I can try to push this work to be groomed. I agree, this is very fundamental, but that also means that we need to be extra careful when implementing this. I sure will introduce `breaking` changes, but can we justify all the `breaks` as `bug fixes`, I am not confident enough to answer with my current knowledge on the issue.

onematchfox:
@bungle Thanks for the feedback. Glad to hear this hasn't fallen completely off your radar. Mostly as a FYI at this point, I have successfully rebased this change on top of 1.5.0 without any hassles. However, a change in approach is required for 2.0.0 due to the changes you introduced in #5234, or more specifically https://github.com/Kong/kong/pull/5234/commits/d90d366fcf533233033b5752b2907f1793ca5caf. 

Going back to the basic requirement I have... Ultimately what I am looking for is a way to ensure a consistent and easily overridable rendering of responses - from exits within core Kong code, internal/NGINX errors, any plugins and from upstream responses. The approach I took here was to try and redirect the rendering of HTTP responses to `error_handlers.lua` as that seemed like a good option for a place in the codebase that could be easily overriden (without having to override any core Kong code). However, the approach you took in the commit above was to ensure that the "rendering" of responses happened in `kong.response.send()` (via `kong.response.exit()`) instead. This  is at least already a step in the right direction IMO because at least the writing of responses to the wire now happens in a single location. However, there is still an issue in that the templating/formatting based on the accept header still only exists within `error_handlers.lua` so internal errors will still end up looking different to response produced with `kong.response.exit(status, { message = x }`. 

I have been doing some thinking about to how to approach this problem with that change in place but haven't had the time to really play around with any code yet. In essence, with your change `kong.response.send` becomes a nice point to implement this sort of change - all that is needed is:
- a simple way to allow a user to apply any custom formatting/templating of `body` prior to calling `ngx.print(body)`. Perhaps a new function/module `response.render(body)` that can be overriden in some manner.
- some additional adjustments to `error_handlers.lua` to not set or format the message and rather just call `kong.response.exit(status)`. Thus, exits as a result of internal errors will not look any different to intentional exits from plugins. And, the same mechanism as above can be used to ensure a standardised response from internal errors. In fact, the current code to set and format the body could even form the default implementation of the "render" method introduced above.

I completely agree with your comment regarding how many places this PR currently touches. Perhaps the change in tact I am considering (proposing?) above seems a little more limited and less daunting?

bungle:
@onematchfox great to hear you back, @locao is going to look this next sprint.

The goal: consistent error reporting.

Currently:
1. The error may happen on Nginx before Kong even starts executing code (`error_page` kicks in)
2. The error may happen when proxying agains (`error_page` kicks in)
3. The plugins may `runtime` error (`error_page` kicks in, but not on `post-content` phases like runtime errors on `header_filter`)
4. The plugins may call `kong.response.exit` with `400-599` codes.

1. and 2. are already quite ok, 3. is partially but needs better error handling on `post-content` phases. The 4. is what this PR tries to solve.

So it seems to be all about using `accept` header to send either one of the supported ones:

1. JSON
2. HTML
3. XML
4. PLAIN

This only matters with `HTTP`. While I think it is fine when the plugin doesn't handle the error, it is bit more controversial with e.g. OAuth 2.0 Plugin that does things like:

```
return kong.response.exit(401, {
  error = "invalid_token",
  error_description = "The access token is invalid or has expired"
}, {
 ["WWW-Authenticate"] = "..."
})
```

And this ends up being `application/json` output. It doesn't follow `Accept` header, but should it? The OAuth 2.0 Specification says:

> The parameters are included in the entity-body of the HTTP response
> using the "application/json" media type as defined by [RFC4627]

See: https://tools.ietf.org/html/rfc6749#section-5.2

While I think this would be great for `5xx` errors (though OAuth 2.0 spec mentions `5xx` there too on same section), I am not totally confident with using this with `4xx` errors. Also adding a configuration option for this feels a bit like a work-around (for those who want it can enable it). Also it makes debugging issues bit harder as you need to take the setting in account.  I wish we could do the right thing in a first place. It also makes `kong.response.exit` even more magical than it is now (e.g. PDK functions start to behave differently based on Kong configuration, and perhaps in unexpected ways).

In most cases the `Accept`/`Content-Type` header will be correct, e.g. you probably use `application/json` already when calling an API with OAuth 2.0 Token (for example). So in that sense I don't feel this is too breaking, even if it is a bit magical. Obviously, when we hesitate, we endup introducing a config option which by default doesn't do anything differently than what Kong already does, but only when turned on (like in this PR). Still I would want to see other options too (in addition to this PR). Here are some ideas:

#### Option 1

Only do this for `errors` listed with status codes specified with `error_page` directive:
https://github.com/Kong/kong/blob/master/kong/templates/nginx_kong.lua#L82-L83

Move that `error_page` to configuration where you can list status codes that you want to be handled the way error_handlers does it.

So if someone for example does `kong.response.exit(500, "body")` it will do it the `error_handlers` way if `500` is listed for `error_page` configuration.

#### Option 2

Implement `kong.response.error` (perhaps that is just almost the `error_handlers.lua`. So plugins that don't care about content-type can just call that and have it done the way Kong handles errors. If plugin cares about content-type it can call `kong.response.exit` that works like it works today. Obviously this has disadvantage that you just cannot say `globally` that this is the way I want to handle things.

#### Option 3

Just make the `kong.response.exit` do this in case the `content-type` is not given as header for the response. And in those plugins where you don't want this to happen, add the `content-type` (if missing). This will of course break the `kong.response.exit` but perhaps in tolerable way.

You see, I am not totally in love with `handler_errors` boolean config option. The `Option 1` is kinda like it, but it also makes things more configurable, and it gets rid of static `error_page`. Still even that makes `kong.response.exit` to behave differently based on configuration.

@locao, @kikito, @javierguerragiraldez, @gszr, @hishamhm, @hutchic, I welcome your input to this discussion too!

kikito:
Good point: it might be that some plugins are *not* supposed to honor the `accept` header for some reason. I had not considered that before.

With that in mind, I like *Option 2* the most, because it doesn't inadvertently break backwards compatibility. Plugins "opt-in" to it by changing:

```
kong.response.exit(500, { message = "an error has ocurred" })
```

to:

```
kong.response.error(500, "an error has ocurred")
```
We can probably add default messages for statuses as well.

Then we can make this change quickly in Kong-controlled plugins, but we allow third party plugins (which could theoretically not want to honor `accept`) to remain as they are.

onematchfox:
> You see, I am not totally in love with handler_errors boolean config option.

Yeah, I agree with you. I didn't particularly like this either but given that I don't have a good view of the bigger picture I though it would be best to approach this in a way that it wouldn't introduce any breaking changes and that seemed like the easiest option from my point of view at the time.

Your point regarding plugins not wanting to respect the accept header is certainly a good one. Even in the HTTP spec observance of the accept header is optional. Ultimately what `error_handlers.lua` is doing now is a very simple form of proactive content negotiation (https://tools.ietf.org/html/rfc7231#section-3.4) and we are looking to extend that functionality to all responses generated by Kong code/plugins. 

With that said, I do still wonder if the situation where a plugin wishes to ignore the accept header is more of an exception than the norm. I am also hesitant to go with a solution based around the exceptions for the sake of avoiding a breaking change. Of course, I may be completely wrong here in what the norm vs the exception really is. My preference would be to go with sensible defaults whilst providing existing implementations a mechanism to maintain current functionality if so desired. To that end I wonder if I couldn't suggest:

**Option 4**

Existing `kong.response.exit/send` is altered to attempt to satisfy the supplied `accept` header, whilst also providing the calling code the option to explicitly state the returned content type. If `kong.response.exit` were to check for the existence of the `content-type` header prior to looking at the `accept` then the OAuth plugin (and any others that require this functionality) could be adjusted to maintain current functionality by calling

```
return kong.response.exit(401, {
  error = "invalid_token",
  error_description = "The access token is invalid or has expired"
}, {
 ["WWW-Authenticate"] = "...",
 ["Content-Type"] = "application/json"
})
```

Alternatively the content type could be made an explicit method parameter - perhaps less "magical"?

Not to introduce too much scope creep, but in addition to this, it may be worth reconsidering the current implementation of setting the content type in `error_handlers.lua`. If I'm not mistaken, this currently ignores the priority rules and so, if the accept header were set to `text/html, application/json;q=0.9, */*;q=0.8` then the code would currently return JSON instead of html. In adjusting this, one could also adjust the default to `application/json` rather than the current `text/plain`. If the code defaults to `application/json`, then with this change the only situation I can think of which could be seen as "breaking" is when:
- user agents request a non-json response (e.g. `text/html`)
- the plugin wishes to return a specific content type and is not changed to explicitly set the returned content type

In this "breaking" scenario the user will now actually receive what they asked for instead of `application/json` which the plugin would have returned previously.

On a side note, one side effect (benefit?) of this which I had not considered before is that this functionality would not be limited to error statuses (in comparison to this PR and the other options presented). Thus it could even be used by a plugin that wanted to return a `200` with a "correctly formatted" response.

----------------------------

If you really want to avoiding a breaking change then I'd agree that Option 2 does seem like a fair approach here. Although, given that @bungle correctly mentions that this only really affects `HTTP` traffic I guess this means that `kong.response.error` would need to contain some branching logic that ends up calling `kong.response.exit/send` where plugins may be designed to operate on multiple traffic types (if that's even possible)?

@kikito - is there a particular reason why you changed the table parameter in `kong.response.exit` to a string in `kong.response.error`? The table parameter would seem more flexible to me if user's do want to provide additional content. Although perhaps that does complicate the rendering of html content?

------

Finally, if I may just go back to the goal

> The goal: consistent error reporting.

In my response I mentioned "easily overridable rendering of responses". I just want to expand on this slightly to ensure this isn't overlooked. I would really like to ensure that the end user has the ability to override/control the method used to write the response. I currently do this with the code contained in this PR by mounting a custom `error_handlers.lua`. In this custom version we provide default messages/bodies for the additional statuses (which could easily be done by default as already suggested in this thread). More importantly, we also inject content across all errors. To give you a couple examples of why we do this:
- For all content types, we currently inject a request/correlation id into the response. This helps from an observability standpoint as the end user can supply this to our support teams who can then use this to look up the request and see more details about where the error happened
- For `html` content we apply custom branding so that we have consistent looking error pages across both Kong errors and all upstreams.

bungle:
> If I'm not mistaken, this currently ignores the priority rules and so, if the accept header were set to text/html, application/json;q=0.9, */*;q=0.8 then the code would currently return JSON instead of html

You are correct, the priority rules are not honoured, it only checks if string starts with something. So it will be served as `text/html`. But `blaa/blaa, application/json;q=0.9, */*;q=0.8` will be served as `plain`.

kikito:
>  is there a particular reason why you changed the table parameter in kong.response.exit to a string in kong.response.error? The table parameter would seem more flexible to me if user's do want to provide additional content. Although perhaps that does complicate the rendering of html content?

Ultimately I think kong.response.error should be able to handle both a string and a table. Candidates for rules:

* When the content-type is json:
  * if `msg` is a table, return it transformed into json
  * if `msg` is a string, convert to the json table `{ "message": <msg> }`
* When the content-type is NOT json:
  * if `msg` is a string, it will be sent "as is".
  * if `msg` is a table, it will be transformed into json prior to sending.

So making the transformation (from table to plain string) will have no impact when content-type is JSON, but will render more nicely in txt/html (it will avoid json-inside-html or json-in-txt).

guanlan:
@hishamhm Can you help here?

kikito:
I am closing this because we went through an alternative implementation (#5562) . This PR was very influential on getting to that solution and we are very grateful for the feedback and work which went into it. Thanks!

PR #5056, opened 2019-09-20
chore(reports) sends reports with tcp instead of udp

bungle:
### Summary

Sends reports with `tcp` instead of `udp`. Also changes the port where the reports are sent from 61829 to 61830.

### DIFF
--- a/kong/reports.lua
+++ b/kong/reports.lua
@@ -5,7 +5,7 @@ local constants = require "kong.constants"
 
 local kong_dict = ngx.shared.kong
 local ngx = ngx
-local udp_sock = ngx.socket.udp
+local tcp_sock = ngx.socket.tcp
 local timer_at = ngx.timer.at
 local ngx_log = ngx.log
 local var = ngx.var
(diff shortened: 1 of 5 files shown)


kong/reports.lua
```diff
@@ -121,26 +122,51 @@ local function send_report(signal_type, t, host, port)
...
   end
 
-  local sock = udp_sock()
-  local ok, err = sock:setpeername(host, port)
-  if not ok then
-    log(WARN, "could not set peer name for UDP socket: ", err)
-    return
-  end
+  local sock
+  if use_udp then
```
dndx: Forgive if I'm missing the obvious, but is there any reason the UDP mode is still being kept around?

bungle: @dndx, if someone decides later on to send udp, perhaps on some reports. We can name it protocol too, if someone later on wants us to send it with tls or perhaps grpc.

bungle: @dndx I discussed this with @hishamhm, I'll remove the udp code.

bungle: @dndx the udp is now removed.

kong/reports.lua
```diff
@@ -121,25 +121,23 @@ local function send_report(signal_type, t, host, port)
     end
   end
 
-  local sock = udp_sock()
-  local ok, err = sock:setpeername(host, port)
+  local sock = tcp_sock()
+  sock:settimeout(1000)
```
thibaultcha: 1 second seems to short of a timeout. The network could be congested for any reason, and we might lose some reports given no the lack of retry policy. There should be no harm in increasing this value to a much higher timeout given no client or user is waiting for a response here.

Besides, better use the newer `settimeouts` method instead.

PR #5062, opened 2019-09-24
Automatically escape "#" characters in environment variable configuration

### LINKED ISSUE #4525: Kong 0.14.1 won't start with "special characters" in password
Summary
 my kong version is 0.14.1 ,i set pg password contain #,when kong  migration up ,I get the follow error:

2019/04/18 10:41:29 [verbose] Kong: 0.14.1
Error:
/usr/local/share/lua/5.1/kong/cmd/migrations.lua:37: [postgres error] could not retrieve current migrations: [postgres error] FATAL: password authentication failed for user "kong"
stack traceback:
        [C]: in function 'assert'
        /usr/local/share/lua/5.1/kong/cmd/migrations.lua:37: in function 'cmd_exec'
        /usr/local/share/lua/5.1/kong/cmd/init.lua:87: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:87>
        [C]: in function 'xpcall'
        /usr/local/share/lua/5.1/kong/cmd/init.lua:87: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:44>
        /usr/local/bin/kong:7: in function 'file_gen'
        init_worker_by_lua:48: in function <init_worker_by_lua:46>
        [C]: in function 'xpcall'
        init_worker_by_lua:55: in function <init_worker_by_lua:53>
2019/04/18 10:41:29 [verbose] no config file found at /etc/kong/kong.conf
2019/04/18 10:41:29 [verbose] no config file found at /etc/kong.conf
2019/04/18 10:41:29 [verbose] no config file, skipping loading
2019/04/18 10:41:29 [verbose] prefix in use: /usr/local/kong
2019/04/18 10:41:29 [verbose] running datastore migrations

rainest:
### Summary

Automatically escape `#` in environment variable configuration if not escaped already. Both environment variables and kong.conf values later pass through [check_and_infer](https://github.com/Kong/kong/blob/1.3.0/kong/conf_loader.lua#L270-L277), which treats `#` as a comment and strips it along with anything after. This ca break database passwords in a non-obvious way.

### Full changelog

* Add logic to escape `#` characters in environment variables before passing them further through configuration parsing.
* Add test helpers to set environment variables within tests.

### Issues resolved

Fix #4525


rainest:
Alternately, replace the value silently. This block already operates on environment variables only, so it won't screw up future comment stripping:

```
    log.debug('%s ENV found with "%s"', env_name, to_print)
    if string.match(env, "[^\\]#") then
      env = ngx.re.gsub(env, "([^\\\\])#", "$1\\#")
    end
```
My only concern with displaying a warning is that it does disclose some information (albeit not much--just that one or more characters are `#`) about the values of likely sensitive settings.

Replacing silently in this fashion works as best I can tell with manual testing. However, I'm unsure if/how we can add unit tests for it: https://github.com/Kong/kong/blob/1.3.0/spec/01-unit/03-conf_loader_spec.lua does not appear to contain tests for environment variables, and Lua does not appear to natively provide a way to set them within the test.

rainest:
Guilherme S. helpfully provided examples of how tests can set environment variables, so I've put that in place along with automatic replacement. For posterity, the old variant that only logged a warning was:

```
    if string.match(env, "[^\\]#") then
      log.warn('%s contains an unescaped "#" character and will not parse correctly. Please escape it with "\\#"', env_name)
    end
```

thibaultcha:
Manually merged to master with a bunch of fixes. Thanks!

PR #5064, opened 2019-09-24
feat(datadog) add overloaded_service_name and common_tags parameters

NathanRichard:
# Summary

These two parameters are optional and do not change the current behavior by default.

```overloaded_service_name``` allows to force the string used in place of the service name when building  the metrics name. This is useful when several kong services actually implement the same "conceptual" service, for instance when several upstreams provide the same API. For cases like this, datadog encourages the usage of a common metric with tags to filter the subsets when required.
    
```common_tags``` allows to specify a collection of tags at service level that will be used by all metrics. The common tags are added to the metrics-level tags. This is actually a sugar parameter to avoid redefining all the default metrics just to apply the same tag to all of them. This helps keep the configuration logic simple and the configuration files (when using DB-less mode) more readable.

# Example of use in DB-less mode

```yaml
- name: myVHost1-loginService
  plugins:
  - config:
      common_tags:
      - vhost:v1
      overloaded_service_name: loginService
    name: datadog
  routes:
  - hosts:
    - myvhost1.mydomain.com
    methods:
    - POST
    paths:
    - /login
    protocols:
    - https
  url: http://localhost:8080/login
- name: myVHost2-loginService
  plugins:
  - config:
      common_tags:
      - vhost:v2
      overloaded_service_name: loginService
    name: datadog
  routes:
  - hosts:
    - myvhost2.mydomain.com
    methods:
    - POST
    paths:
    - /login
    protocols:
    - https
  url: http://localhost:8081/login
```

### DIFF
--- a/kong/plugins/datadog/handler.lua
+++ b/kong/plugins/datadog/handler.lua
@@ -1,6 +1,6 @@
 local basic_serializer = require "kong.plugins.log-serializers.basic"
 local statsd_logger    = require "kong.plugins.datadog.statsd_logger"
-
+local utils            = require "kong.tools.utils"
 
 local ngx_log       = ngx.log
 local ngx_timer_at  = ngx.timer.at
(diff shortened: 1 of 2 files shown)


NathanRichard:
This PR is very similar to https://github.com/Kong/kong/pull/3037 but as this other PR did not fit exactly my requirements and had been blocked for some time, I developed this other flavor.
As I find it more flexible than https://github.com/Kong/kong/pull/3037, I hope it can solve both our requirements.

kikito:
Hi, thanks a lot for taking the time to send this PR. Unfortunately I have to reject it.

The direction we're going towards is changing all service ids into tags (see #5154). `overloaded_service_name` does not fit that scenario well.

With regards to your other proposed improvement:

> This is actually a sugar parameter to avoid redefining all the default metrics just to apply the same tag to all of them

YAML already has a feature called Anchors and Aliases which allows doing that anywhere in a YAML. I suggest you use that feature to avoid repetitiveness. See here for an example: https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

Again, we are grateful for your contribution, even if not accepted. Don't hesitate to try again in the future!

PR #5071, opened 2019-09-26
feat(cassandra) refresh cluster topology 

### LINKED ISSUE #2674: Kong doesn't remove old Cassandra nodes

thibaultcha:
### Summary

A new `cassandra_refresh_frequency` property schedules a recurring background timer calling lua-cassandra's [cluster:refresh()](https://thibaultcha.github.io/lua-cassandra/modules/resty.cassandra.cluster.html#_Cluster:refresh) method in the background. In lua-cassandra 1.5.0, this method has been refactored to be safe to call at runtime.

When `cassandra_refresh_frequency` is `0`, no topology refresh timer will be scheduled. A restart of Kong is necessary for lua-cassandra to consider new nodes (decommissioned nodes are blacklisted by the retry policy). The default value for this timer is chosen to be `60`.

No Admin API endpoint has been implemented as part of this PR for the time being. Subsequent patches welcome if a manual refresh feature turns out to be desired.

**Note:** a more consequent refactor of lua-cassandra's CQL binary protocol implementation would allow for REGISTERing a connection, and interpreting [TOPOLOGY_CHANGE](https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec#L794) events, which would be a more efficient approach.  

### Full changelog

* Bump lua-cassandra to 1.5.0 ([Changelog](https://github.com/thibaultcha/lua-cassandra/blob/master/CHANGELOG.md#150))
* Add a new property: `cassandra_refresh_frequency`

### Issues resolved

Fix #2674

### DIFF
--- a/kong-1.3.0-0.rockspec
+++ b/kong-1.3.0-0.rockspec
@@ -20,7 +20,7 @@ dependencies = {
   "multipart == 0.5.5",
   "version == 1.0.1",
   "kong-lapis == 1.7.0.1",
-  "lua-cassandra == 1.4.0",
+  "lua-cassandra == 1.5.0",
   "pgmoon == 1.10.0",
   "luatz == 0.4",
   "http == 0.3",
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -667,6 +667,14 @@
                                  # of the local (closest) datacenter for this
                                  # Kong node.
 
+#cassandra_refresh_frequency = 60          # Frequency (in seconds) at which
+                                           # the cluster topology will be
+                                           # checked for new or decommissioned
+                                           # nodes.
+                                           # A value of `0` will disable this
+                                           # check, and the cluster topology
+                                           # will never be refreshed.
+
 #cassandra_repl_strategy = SimpleStrategy  # When migrating for the first time,
                                            # Kong will use this setting to
                                            # create your keyspace.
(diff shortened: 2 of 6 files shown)


subnetmarco:
> When cassandra_refresh_frequency is 0, no topology refresh timer will be scheduled.

@thibaultcha would it be more consistent to have:

```
cassandra_refresh = on/off
cassandra_refresh_frequency = ..
```

I believe that there are properties where setting the value to `0` doesn't imply that the functionality is disabled, but that the timer is set to run continuously.

PR #5079, opened 2019-09-30
key-auth option to let backend handle HTTP 401

mkhpalm:
### Summary

Sometimes we would like to see/handle unauthorized requests at the service behind kong. Return our own responses via the service for unauthorized requests. Currently there is no way to pass back unauthorized requests since key-auth always returns its own 401 prematurely (in some cases). 

This change is also handy if kong is not the source of record for api keys since individual consumers can be defined in kong that need to be rate-limited transparently without affecting other users.

### Full changelog

* New config.allow_unauthorized config option

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -154,9 +154,14 @@ local function do_authentication(conf)
     return kong.response.exit(500, "An unexpected error occurred")
   end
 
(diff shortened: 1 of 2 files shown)


Tieske:
how would this be different from using the anonymous user?

mkhpalm:
Good question! I guess I just didn't understand the anonymous user could be used like that. But now that I re-read the docs it seems to say it right there. Thanks!

PR #5081, opened 2019-10-01
Rectified phrasing and grammar in CONTRIBUTING.md

7wikd:
Tried to rectify the phrasing errors as well as the grammatical errors in CONTRIBUTING.md file


thibaultcha:
Hello,

Thanks for the patch. However, it doesn't seem like it fixes any mistakes (as far as I can tell, it may be adding some though...). It doesn't feel like we need those changes, thanks!

PR #5092, opened 2019-10-03
fix: move clean logic out of #prepare_prefix

veetow:
### Summary

I maintain a monorepo of Kong plugins. I have a requirement that the Kong `prefix` be retained across the entire test suite run. This patch fixes an inconvenience where `spec/helpers.lua#start_kong` does not respect the `KONG_TEST_DONT_CLEAN` environment variable and always cleans out the `prefix` directory.

### Full changelog

* fix: move clean logic out of #prepare_prefix


thibaultcha:
Merged with some edits, thanks!

PR #5100, opened 2019-10-07
feat(db) add support for read transformations

bungle:
### Summary

This is continuation of the dao transformations. This time the PR adds read transformations to DAOs, and opens a door for encryption of data at rest.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -2054,18 +2054,30 @@ end
 --- Run transformations on fields.
 -- @param input The input table.
 -- @param original_input The original input for transformation detection.
+-- @param context a string describing the CRUD context:
+-- valid values are: "insert", "update", "upsert", "select"
 -- @return the transformed entity
-function Schema:transform(input, original_input)
+function Schema:transform(input, original_input, context)
   if not self.transformations then
     return input
   end
 
   local output = input
   for _, transformation in ipairs(self.transformations) do
-    local on_write = transformation.on_write
+    local transform
+    if context == "select" then
+      transform = transformation.on_read
+
+    else
+      transform = transformation.on_write
+    end
+
+    if not transform then
+      goto next
+    end
+
     local args = {}
     local argc = 0
-    local all_set  = true
     for _, input_field_name in ipairs(transformation.input) do
       local value = get_field(original_input or input, input_field_name)
       if is_nonempty(value) then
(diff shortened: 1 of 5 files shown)


bungle:
'Ok

kong/db/schema/init.lua
```diff
@@ -2054,48 +2054,68 @@ end
...
-        break
-      end
+    else
+      transform = transformation.on_write
     end
 
-    if all_set then
-      if transformation.needs then
-        for _, need in ipairs(transformation.needs) do
+    if transform then
```
kikito: This should be transformed into an early exit: `if not transform then return`

kong/db/schema/init.lua
```diff
@@ -2054,48 +2054,68 @@ end
...
+          all_set = false
+          break
         end
       end
 
-      local data, err = on_write(unpack(args))
-      if err then
-        return nil, validation_errors.TRANSFORMATION_ERROR:format(err)
-      end
+      if all_set then
```
kikito: Maybe this is not needed if we do an early exit before.

kong/db/schema/init.lua
```diff
@@ -2054,48 +2054,68 @@ end
...
           argc = argc + 1
-          args[argc] = get_field(input, need)
+          if original_input then
+            args[argc] = get_field(input, input_field_name)
+          else
+            args[argc] = value
+          end
+
+        else
+          all_set = false
```
kikito: Can we avoid setting `all_set` at all? It seems we could do `return` here instead of setting `all_set` to false + break + an if a bit later on.

kong/db/schema/init.lua
```diff
@@ -2054,48 +2054,68 @@ end
...
+      if all_set then
+        if transformation.needs then
+          for _, need in ipairs(transformation.needs) do
+            local value = get_field(input, need)
+            if is_nonempty(value) then
+              argc = argc + 1
+              args[argc] = get_field(input, need)
+
+            else
+              all_set = false
```
kikito: ditto: consider ditching all_set and just returning.

kong/db/schema/init.lua
```diff
@@ -2054,48 +2054,68 @@ end
...
+      local args = {}
+      local argc = 0
+      local all_set  = true
+      for _, input_field_name in ipairs(transformation.input) do
+        local value = get_field(original_input or input, input_field_name)
+        if is_nonempty(value) then
           argc = argc + 1
-          args[argc] = get_field(input, need)
+          if original_input then
+            args[argc] = get_field(input, input_field_name)
```
kikito: This second if (`if original_input`) is really difficult to understand for me.
* First, get the value using either `original_input` or `input`.
* Then, if the value is non_empty
  * and the original_input is not `nil` nor `false`
    * then the next argument is getting the value but using only `input` (which we have not checked whether it's nil or not?)
*   otherwise, the next argument is `value`, which might come from `original_input`, or `input`, if `original_input` was `nil` or `false`.

Is there a way to make this set of nested ifs a bit less contrived?

I suggest starting by doing `if original_input then` and moving on from there.

Also, pay attention that `false` might be a valid value in `original_input` (we might need to explicitly check for `if original_input ~= nil` instead)

kong/db/schema/init.lua
bungle: @kikito it is inside a for loop, so `return` is not an option, unless we make more functions.

PR #5101, opened 2019-10-07
CONTRIBUTING.md typos fixed

udz-codes:


thibaultcha:
Thank you, but we consider the document to be ok as-is.

PR #5102, opened 2019-10-07
Feat/match port

javierguerragiraldez:
### Summary

Allow routes to specify a port in the host parameter.  If present, it matches only requests to that
specific port.  If absent, requests to any port matches.  Requests without explicit port are handled as if they included the appropriate default port (80 for HTTP, 443 for HTTPS).  Routes with port have higher priority than those that those without; even for requests without explicit port.

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -32,6 +32,12 @@ local function validate_host(host)
 end
 
 
+local function validate_host_with_optional_port(host)
+  local res, err_or_port = utils.normalize_ip(host)
+  return (res and true or nil), err_or_port
+end
+
+
 local function validate_ip(ip)
   local res, err = utils.normalize_ip(ip)
   if not res then
@@ -212,6 +218,12 @@ typedefs.protocol = Schema.define {
 typedefs.host = Schema.define {
   type = "string",
   custom_validator = validate_host,
+}
+
+
+typedefs.host_with_optional_port = Schema.define {
+  type = "string",
+  custom_validator = validate_host_with_optional_port,
 }
 
 
@@ -400,7 +412,7 @@ typedefs.protocols_http = Schema.define {
 
 local function validate_host_with_wildcards(host)
   local no_wildcards = string.gsub(host, "%*", "abc")
-  return typedefs.host.custom_validator(no_wildcards)
+  return typedefs.host_with_optional_port.custom_validator(no_wildcards)
 end
 
 local function validate_path_with_regexes(path)
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -15,9 +15,11 @@ local var           = ngx.var
 local ngx_log       = ngx.log
 local insert        = table.insert
 local sort          = table.sort
+local byte          = string.byte
 local upper         = string.upper
 local lower         = string.lower
 local find          = string.find
+local format        = string.format
 local sub           = string.sub
 local tonumber      = tonumber
 local ipairs        = ipairs
@@ -50,6 +52,99 @@ do
     end
   end
 end
+
+
+local split_port
+do
+  local function safe_add_port(host, port)
+    if not port then
+      return host
+    end
+
+    return host .. ':' .. port
+  end
+
+  local ZERO, NINE, LEFTBRACKET, RIGHTBRACKET = ("09[]"):byte(1, -1)
+
+  local function onlydigits(s, begin)
+    for i = begin or 1, #s do
+      local c = byte(s, i)
+      if c < ZERO or c > NINE then
+        return false
+      end
+    end
+    return true
+  end
+
+  --- splits an optional ':port' section from a hostname
+  -- the port section must be decimal digits only.
+  -- brackets ('[]') are peeled off the hostname if present.
+  -- if there's more than one colon and no brackets, no split is possible.
+  -- on non-parseable input, returns name unchanged,
+  -- every string input produces at least one string output.
+  -- @tparam name string the string to split.
+  -- @tparam default_port number default port number
+  -- @treturn string hostname without port
+  -- @treturn string hostname with port
+  -- @treturn boolean true if input had a port number
+  local function l_split_port(name, default_port)
+    if byte(name, 1) == LEFTBRACKET then
+      if byte(name, -1) == RIGHTBRACKET then
+        return sub(name, 2, -2), safe_add_port(name, default_port), false
+      end
+
+      local splitpos = find(name, "]:", 2, true)
+      if splitpos then
+        if splitpos == #name - 1 then
+          return sub(name, 2, splitpos - 1), name .. (default_port or ''), false
+        end
+
+        if onlydigits(name, splitpos + 2) then
+          return sub(name, 2, splitpos - 1), name, true
+        end
+      end
+
+      return name, safe_add_port(name, default_port), false
+    end
+
+    local firstcolon = find(name, ":", 1, true)
+    if not firstcolon then
+      return name, safe_add_port(name, default_port), false
+    end
+
+    if firstcolon == #name then
+      local host = sub(name, 1, firstcolon - 1)
+      return host, safe_add_port(host, default_port), false
+    end
+
+    if not onlydigits(name, firstcolon + 1) then
+      if default_port then
+        return name, format("[%s]:%s", name, default_port), false
+      end
+
+      return name, name, false
+    end
+
+    return sub(name, 1, firstcolon - 1), name, true
+  end
+
+  -- split_port is a pure function, so we can memoize it.
+  local memo_h = setmetatable({}, {__mode = 'k'})
+  local memo_hp = setmetatable({}, {__mode = 'k'})
+  local memo_p = setmetatable({}, {__mode = 'k'})
+
+  function split_port(name, default_port)
+    local k = name .. '#' .. (default_port or '')
+    local h, hp, p = memo_h[k], memo_hp[k], memo_p[k]
+    if not h then
+      h, hp, p = l_split_port(name, default_port)
+      memo_h[k], memo_hp[k], memo_p[k] = h, hp, p
+    end
+
+    return h, hp, p
+  end
+end
+
 
 
 --[[
@@ -88,6 +183,7 @@ end)
 local MATCH_SUBRULES = {
   HAS_REGEX_URI    = 0x01,
   PLAIN_HOSTS_ONLY = 0x02,
+  HAS_HOST_PORT    = 0x04,
 }
 
 local EMPTY_T = {}
@@ -216,6 +312,7 @@ local function marshall_route(r)
 
     local has_host_wildcard
     local has_host_plain
+    local has_port
 
     for _, host in ipairs(hosts) do
       if type(host) ~= "string" then
@@ -228,6 +325,12 @@ local function marshall_route(r)
 
         local wildcard_host_regex = host:gsub("%.", "\\.")
                                         :gsub("%*", ".+") .. "$"
+
+        _, _, has_port = split_port(host)
+        if not has_port then
+          wildcard_host_regex = wildcard_host_regex:gsub("%$$", [[(?::\d+)?$]])
+        end
+
         insert(route_t.hosts, {
           wildcard = true,
           value    = host,
@@ -254,6 +357,11 @@ local function marshall_route(r)
     if not has_host_wildcard then
       route_t.submatch_weight = bor(route_t.submatch_weight,
                                     MATCH_SUBRULES.PLAIN_HOSTS_ONLY)
+    end
+
+    if has_port then
+      route_t.submatch_weight = bor(route_t.submatch_weight,
+                                    MATCH_SUBRULES.HAS_HOST_PORT)
     end
   end
 
@@ -661,7 +769,9 @@ end
 do
   local matchers = {
     [MATCH_RULES.HOST] = function(route_t, ctx)
-      local host = route_t.hosts[ctx.hits.host or ctx.req_host]
+      local req_host = ctx.hits.host or ctx.req_host
+      local host = route_t.hosts[req_host]
+        or route_t.hosts[split_port(req_host)]
       if host then
         ctx.matches.host = host
         return true
@@ -671,7 +781,7 @@ do
         local host_t = route_t.hosts[i]
 
         if host_t.wildcard then
-          local from, _, err = re_find(ctx.req_host, host_t.regex, "ajo")
+          local from, _, err = re_find(ctx.host_with_port, host_t.regex, "ajo")
           if err then
             log(ERR, "could not evaluate wildcard host regex: ", err)
             return
@@ -995,6 +1105,7 @@ _M.has_capturing_groups = has_capturing_groups
 
 -- for unit-testing purposes only
 _M._set_ngx = _set_ngx
+_M.split_port = split_port
 
 
 function _M.new(routes)
@@ -1167,7 +1278,7 @@ function _M.new(routes)
 
   local grab_req_headers = #plain_indexes.headers > 0
 
-  local function find_route(req_method, req_uri, req_host,
+  local function find_route(req_method, req_uri, req_host, req_scheme,
                             src_ip, src_port,
                             dst_ip, dst_port,
                             sni, req_headers)
@@ -1179,6 +1290,9 @@ function _M.new(routes)
     end
     if req_host and type(req_host) ~= "string" then
       error("host must be a string", 2)
+    end
+    if req_scheme and type(req_scheme) ~= "string" then
+      error("scheme must be a string", 2)
     end
     if src_ip and type(src_ip) ~= "string" then
       error("src_ip must be a string", 2)
@@ -1236,13 +1350,13 @@ function _M.new(routes)
 
     req_method = upper(req_method)
 
-    if req_host ~= "" then
-      -- strip port number if given because matching ignores ports
-      local idx = find(req_host, ":", 2, true)
-      if idx then
-        ctx.req_host = sub(req_host, 1, idx - 1)
-      end
-    end
+    -- req_host might have port or maybe not, host_no_port definitely doesn't
+    -- if there wasn't a port, req_port is assumed to be the default port
+    -- according the protocol scheme
+    local host_no_port, host_with_port = split_port(
+        req_host, req_scheme == 'https' and 443 or 80)
+
+    ctx.host_with_port = host_with_port
 
     local hits         = ctx.hits
     local req_category = 0x00
@@ -1255,12 +1369,14 @@ function _M.new(routes)
 
     -- host match
 
-    if plain_indexes.hosts[ctx.req_host] then
+    if plain_indexes.hosts[host_with_port]
+      or plain_indexes.hosts[host_no_port]
+    then
       req_category = bor(req_category, MATCH_RULES.HOST)
 
     elseif ctx.req_host then
       for i = 1, #wildcard_hosts do
-        local from, _, err = re_find(ctx.req_host, wildcard_hosts[i].regex, "ajo")
+        local from, _, err = re_find(host_with_port, wildcard_hosts[i].regex, "ajo")
         if err then
           log(ERR, "could not match wildcard host: ", err)
           return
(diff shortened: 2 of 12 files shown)


hishamhm:
@javierguerragiraldez in the DB-less testsuite, the error happening is that with your PR the error message has changed from `invalid value` to `invalid hostname`. That test file (`spec/02-integration/02-cmd/02-start_stop_spec.lua`) needs to be adjusted in your PR accordingly.

locao:
@javierguerragiraldez you can safely ignore that failure. That test should be marked as flaky, the error happens because the balancer hasn't been completely created when the test runs.

kong/db/schema/typedefs.lua
```diff
@@ -20,15 +20,7 @@ local type = type
...
-  if type(err_or_port) == "string" and err_or_port ~= "invalid port number" then
-    return nil, "invalid value: " .. host
-  end
-
-  if err_or_port == "invalid port number" or type(res.port) == "number" then
-    return nil, "must not have a port"
-  end
-
-  return true
+  return res and true or false, err_or_port
```
locao: According to our convention, this function should return `nil, error` on failure.

javierguerragiraldez: done

kong/router.lua
```diff
@@ -228,6 +230,12 @@ local function marshall_route(r)
 
         local wildcard_host_regex = host:gsub("%.", "\\.")
                                         :gsub("%*", ".+") .. "$"
+
+        _, host_port = utils.split_port(host)
+        if not host_port then
+          wildcard_host_regex = wildcard_host_regex:gsub('%$$', [[(:\d+)?$]])
```
thibaultcha: Better making this group a non-capturing one?

spec/01-unit/08-router_spec.lua
```diff
@@ -777,6 +808,78 @@ describe("Router", function()
...
+          router = assert(Router.new(use_case))
+        end)
+
+        -- match the right port
+        local match_t = router.select("GET", "/", "foo.route.net:123")
+        assert.truthy(match_t)
+        assert.equal(use_case[3].route, match_t.route)
+
+        -- fail different port
+        assert.is_nil(router.select("GET", "/", "foo.route.net:456"))
```
thibaultcha: Are we lacking a test for failing to match a port on _plain_ hosts, or did I miss it (this is a wildcard host)?

spec/01-unit/08-router_spec.lua
```diff
@@ -2239,7 +2346,7 @@ describe("Router", function()
           },
           route           = {
             preserve_host = true,
-            hosts         = { "preserve.com" },
+            hosts         = { "preserve.com", "preserve.com:123" },
```
thibaultcha: Is this new fixture value used anywhere in the tests? It seems like the intent was to ensure the request preserved Host header does include the port, which is a valuable test to have.

spec/02-integration/05-proxy/02-router_spec.lua
```diff
@@ -352,12 +352,14 @@ for _, strategy in helpers.each_strategy() do
         routes = insert_routes(bp, {
           {
             protocols = { "grpc", "grpcs" },
-            hosts = { "grpc1" },
+            hosts = { "grpc1", "grpc1:"..helpers.get_proxy_port(false, true),
+              "grpc1:"..helpers.get_proxy_port(true, true) },
```
thibaultcha: Are those new fixture values used in any new tests?

style: please add spaces around operators (`..`, `-`, etc...), and we should also align such array-like tables as per the rest of the codebase's codestyle:
```lua
hosts = {
  "..." .. "...",
  "..." .. "...",
}
```

spec/02-integration/05-proxy/02-router_spec.lua
```diff
@@ -386,7 +388,7 @@ for _, strategy in helpers.each_strategy() do
       end)
 
       it("restricts a route to its 'hosts' if specified", function()
-        local ok, resp = proxy_client_grpc({
+        local ok, resp = assert(proxy_client_grpc({
```
thibaultcha: All of the below changes adding the assertion to the gRPC client calls should be part of a different commit to ensure atomicity of the git history (e.g. if we end up reverting the port matching commit for some reason in the future).

spec/01-unit/08-router_spec.lua
```diff
@@ -777,6 +808,78 @@ describe("Router", function()
...
+        -- fail different port
+        assert.is_nil(router.select("GET", "/", "foo.route.net:456"))
+
+        -- port-specific is higher priority
+        local match_t = router.select("GET", "/", "route.org:123")
+        assert.truthy(match_t)
+        assert.equal(use_case[4].route, match_t.route)
+      end)
+
+      it("prefers port-specific even for default port", function()
```
thibaultcha: We only test plain HTTP's default port (80), but not HTTPs' default port?

spec/01-unit/05-utils_spec.lua
```diff
@@ -437,6 +437,24 @@ describe("Utils", function()
...
+        {'localhost:23h', {'localhost:23h', nil}},
+        {'localhost/24', {'localhost/24', nil}},
+        {'::1', {'::1', nil}},
+        {'[::1]', {'::1', nil}},
+        {'[::1]:', {'::1', nil}},
+        {'[::1]:80', {'::1', 80}},
+        {'[::1]:80b', {'[::1]:80b', nil}},
+        {'[::1]/96', {'[::1]/96', nil}},
+      }) do
+        assert.same(case[2], {utils.split_port(case[1])})
```
thibaultcha: style: missing spaces in the table notation (`{ ... }`)

kong/router.lua
```diff
@@ -662,6 +675,7 @@ do
   local matchers = {
     [MATCH_RULES.HOST] = function(route_t, ctx)
       local host = route_t.hosts[ctx.hits.host or ctx.req_host]
+        or route_t.hosts[utils.split_port(ctx.hits.host or ctx.req_host)]
```
thibaultcha: This may cause a lot of extra invocations of `split_port()` (for each evaluated route). Could we avoid it somehow? E.g. by storing `host_no_port` in the `ctx` table and only call `split_port` when `ctx.hits.host` is non-nil. Or even by calling `split_port` for each wildcard host value during initialization (in `marshall_route()`).

kong/tools/utils.lua
```diff
@@ -661,6 +661,47 @@ _M.hostname_type = function(name)
...
 
+--- splits an optional ':port' section from a hostname
+-- the port section must be decimal digits only.
+-- brackets ('[]') are peeled off the hostname if present.
+-- if there's more than one colon and no brackets, no split is possible.
+-- on non-parseable input, returns name unchanged,
+-- every string input produces at least one string output.
+-- @param name (string) the string to split.
+-- @return hostname (string)
+-- @return port (as number) or nil if not found
```
thibaultcha: Better to use the `@tparam` and `@treturn` annotations which allow you to specify the types of the arguments and return values in a normalized way, see other examples in this codebase (e.g. in the PDK documentation)

kong/router.lua
javierguerragiraldez: i tried to store `host_no_port`, but seems there's something about lifetime that escapes me.  in the end I resorted to simply memoize `split_port()` itself since it's a pure function

kong/tools/utils.lua
```diff
@@ -661,6 +661,47 @@ _M.hostname_type = function(name)
   return "name"
 end
 
+--- splits an optional ':port' section from a hostname
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/tools/utils.lua
```diff
@@ -661,6 +661,47 @@ _M.hostname_type = function(name)
...
+-- the port section must be decimal digits only.
+-- brackets ('[]') are peeled off the hostname if present.
+-- if there's more than one colon and no brackets, no split is possible.
+-- on non-parseable input, returns name unchanged,
+-- every string input produces at least one string output.
+-- @param name (string) the string to split.
+-- @return hostname (string)
+-- @return port (as number) or nil if not found
+function _M.split_port(name)
+  local ZERO, NINE, LEFTBRACKET, RIGHTBRACKET = ('09[]'):byte(1, -1)
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/tools/utils.lua
```diff
@@ -661,6 +661,47 @@ _M.hostname_type = function(name)
...
+
+  if name:byte(1) == LEFTBRACKET then
+    if name:byte(-1) == RIGHTBRACKET then
+      return name:sub(2, -2)
+    end
+    local splitpos = name:find(']:', 2, true)
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/tools/utils.lua
```diff
@@ -661,6 +661,47 @@ _M.hostname_type = function(name)
...
+    if splitpos then
+      local port = tonumber(name:sub(splitpos+2))
+      if port or splitpos == #name-1 then
+        return name:sub(2, splitpos-1), port
+      end
+    end
+    return name
+  end
+
+  local firstcolon = name:find(':', 1, true)
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/tools/utils.lua
```diff
@@ -661,6 +661,47 @@ _M.hostname_type = function(name)
...
+
+  if name:byte(1) == LEFTBRACKET then
+    if name:byte(-1) == RIGHTBRACKET then
+      return name:sub(2, -2)
+    end
+    local splitpos = name:find(']:', 2, true)
+    if splitpos then
+      local port = tonumber(name:sub(splitpos+2))
+      if port or splitpos == #name-1 then
+        return name:sub(2, splitpos-1), port
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/router.lua
```diff
@@ -1236,13 +1253,22 @@ function _M.new(routes)
...
-        ctx.req_host = sub(req_host, 1, idx - 1)
+    local host_with_port = ctx.req_host
+    local host_no_port, req_port = utils.split_port(host_with_port)
+    if not req_port then
+      req_port = 80
+      if req_scheme == 'https' then
+        req_port = 443
+      end
+      if host_with_port:byte() ~= host_no_port:byte() or
+        host_no_port:find(':')
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/router.lua
```diff
@@ -1236,13 +1253,22 @@ function _M.new(routes)
 
     req_method = upper(req_method)
 
-    if req_host ~= "" then
-      -- strip port number if given because matching ignores ports
-      local idx = find(req_host, ":", 2, true)
-      if idx then
-        ctx.req_host = sub(req_host, 1, idx - 1)
+    local host_with_port = ctx.req_host
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/router.lua
javierguerragiraldez: done

spec/02-integration/05-proxy/02-router_spec.lua
javierguerragiraldez: done

spec/01-unit/05-utils_spec.lua
javierguerragiraldez: done

kong/tools/utils.lua
javierguerragiraldez: done

kong/router.lua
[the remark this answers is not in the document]
javierguerragiraldez: done

spec/01-unit/08-router_spec.lua
javierguerragiraldez: removed.  it was one of the first tests i tried, but forgot to remove when got a better test

spec/01-unit/01-db/01-schema/06-routes_spec.lua
```diff
@@ -355,7 +355,7 @@ describe("routes schema", function()
     end)
   end)
 
-  describe("hosts attribute", function()
+  describe("hosts attribute #host", function()
```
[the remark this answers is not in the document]
javierguerragiraldez: removed.  i put them to make it easier to focus on the right tests, but they're not useful for the general project without a lot more planning

spec/02-integration/05-proxy/02-router_spec.lua
javierguerragiraldez: removed.  they didn't change the test behaviour (there's an `assert.truthy(ok)` a few lines below), but result in much more information when failing, which helped a lot to see the real failure point

spec/01-unit/08-router_spec.lua
javierguerragiraldez: added

kong/router.lua
```diff
@@ -228,6 +230,12 @@ local function marshall_route(r)
 
         local wildcard_host_regex = host:gsub("%.", "\\.")
                                         :gsub("%*", ".+") .. "$"
+
+        _, host_port = utils.split_port(host)
```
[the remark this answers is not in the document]
javierguerragiraldez: about priorities:
- at config time:
  - non-wildcard routes are stored as provided, with or without port
  - wildcard routes with port have a regex with the port and a set the `host_port` flag
  - wildcard router without port have a regex with a `\d+` at the end, thus matching any port, and don't set the `host_port` flag, so they're lower priority than a wildcard route with port
- at match time:
  - incoming request is tried with port against non-wildcard table  => no-wildcard route with port has the highest priority
  - if no match, is tried without port against non-wildcard table, =>  no-port no-wildcard have "second" priority
  - if no match, it's tried against wildcard regexes.  those with specific port are sorted before those without  => wildcard routes with port are "third" option, and wildcard, no port are last

the spec/01-unit/08-router_spec.lua test file includes priority testing

spec/01-unit/08-router_spec.lua
javierguerragiraldez: you're right, the closest case is where it fails to match a different port and matches a non-port route.  there's no test (yet) without no-port route that would produce no match at all.

javierguerragiraldez: added "port-only" cases (without no-port hosts) to spec/01-unit/08-router_spec.lua to test both succeeding and failing.

kong/router.lua
```diff
@@ -1236,13 +1253,22 @@ function _M.new(routes)
...
-      if idx then
-        ctx.req_host = sub(req_host, 1, idx - 1)
+    local host_with_port = ctx.req_host
+    local host_no_port, req_port = utils.split_port(host_with_port)
+    if not req_port then
+      req_port = 80
+      if req_scheme == 'https' then
+        req_port = 443
+      end
+      if host_with_port:byte() ~= host_no_port:byte() or
```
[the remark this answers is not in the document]
javierguerragiraldez: added comments and tests for IPv6 literals

kong/router.lua
```diff
@@ -1236,13 +1257,29 @@ function _M.new(routes)
...
-    if req_host ~= "" then
-      -- strip port number if given because matching ignores ports
-      local idx = find(req_host, ":", 2, true)
-      if idx then
-        ctx.req_host = sub(req_host, 1, idx - 1)
+    -- req_host might have port or maybe not, host_no_port definitely doesn't
+    -- if there wasn't a port, req_port is assumed to be the default port
+    -- according the protocol scheme
+    local host_with_port = req_host
+    local host_no_port, req_port = utils.split_port(host_with_port)
```
kikito: use `split_port` instead of `utils.split_port`

kong/router.lua
```diff
@@ -1236,13 +1257,29 @@ function _M.new(routes)
...
+    if not req_port then
+      req_port = 80
+      if req_scheme == 'https' then
+        req_port = 443
+      end
+
+      -- a literal IPv6 address needs the format [host]:port,
+      -- the fastest way to check is if the first byte was stripped by split_port()
+      -- otherwise, check for a colon
```
kikito: Wouldn't it be a bit faster (and perhaps a bit more explicit) to do check for the square bracket character directly, instead of "checking if it has changed"?

```
if byte(host_with_port) == LEFTBRACKET or ...
```

Of course disregard this if you end up moving `host_with_port` inside `utils.split_port`

kong/router.lua
```diff
@@ -1236,13 +1257,29 @@ function _M.new(routes)
...
-      -- strip port number if given because matching ignores ports
-      local idx = find(req_host, ":", 2, true)
-      if idx then
-        ctx.req_host = sub(req_host, 1, idx - 1)
+    -- req_host might have port or maybe not, host_no_port definitely doesn't
+    -- if there wasn't a port, req_port is assumed to be the default port
+    -- according the protocol scheme
+    local host_with_port = req_host
+    local host_no_port, req_port = utils.split_port(host_with_port)
+    if not req_port then
```
kikito: Consider putting the port defaults and the calculation of the new `host_with_port` inside `utils.split_port` (at that point it would probably make sense to move it out of `utils` and inside the router itself as a local). That way we can use memoization for that part as well.

(I would never recommend this avenue if this wasn't a hot path)

kong/db/schema/typedefs.lua
```diff
@@ -20,15 +20,7 @@ local type = type
...
-  if type(err_or_port) == "string" and err_or_port ~= "invalid port number" then
-    return nil, "invalid value: " .. host
-  end
-
-  if err_or_port == "invalid port number" or type(res.port) == "number" then
-    return nil, "must not have a port"
-  end
-
-  return true
+  return res and true or nil, err_or_port
```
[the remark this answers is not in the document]
hishamhm: that's a valid concern, good call pointing it out @kikito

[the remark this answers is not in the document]
javierguerragiraldez: done

kong/router.lua
javierguerragiraldez: done

kong/router.lua
javierguerragiraldez: done

kong/router.lua
javierguerragiraldez: disregarded

thibaultcha:
Please hold before merging, thanks!

spec/01-unit/08-router_spec.lua
```diff
@@ -1156,6 +1156,33 @@ describe("Router", function()
...
+        local router = assert(Router.new(use_case))
+
+        local match_t = router.select("GET", "/", "route.org:123")
+        assert.truthy(match_t)
+        assert.equal(use_case[2].route, match_t.route)
+        assert.same("route.*:123", match_t.matches.host)
+        assert.same(nil, match_t.matches.method)
+        assert.same(nil, match_t.matches.uri)
+        assert.same(nil, match_t.matches.uri_captures)
+      end)
```
thibaultcha: This was covered in a few test case, but a bit too indirectly, and I'm afraid the test may be updated in the future and not cover the case, so here is an isolated test case specifically for the submatching priority.

kong/router.lua
```diff
@@ -770,8 +774,7 @@ do
   local matchers = {
     [MATCH_RULES.HOST] = function(route_t, ctx)
       local req_host = ctx.hits.host or ctx.req_host
-      local host = route_t.hosts[req_host]
-        or route_t.hosts[split_port(req_host)]
+      local host = route_t.hosts[req_host] or route_t.hosts[ctx.host_no_port]
```
thibaultcha: This was my last remaining concern: I removed the call to `split_port` on this hot path, without any issues so far storing `host_no_port` in the context.

kong/router.lua
```diff
@@ -181,9 +185,9 @@ sort(SORTED_MATCH_RULES, function(a, b)
 end)
 
 local MATCH_SUBRULES = {
-  HAS_REGEX_URI    = 0x01,
-  PLAIN_HOSTS_ONLY = 0x02,
-  HAS_HOST_PORT    = 0x04,
+  HAS_REGEX_URI          = 0x01,
+  PLAIN_HOSTS_ONLY       = 0x02,
+  HAS_WILDCARD_HOST_PORT = 0x04,
```
thibaultcha: Renamed this submatching rule to be more precise: we only case about it when dealing with regex hosts.

PR #5104, opened 2019-10-07
feat(cmd) accept a filename argument for `kong config init`

hishamhm:


kong/cmd/config.lua
```diff
@@ -49,19 +49,17 @@ local function db_export(filename, conf)
 end
 
 
-local function generate_init()
-  if pl_file.access_time(INIT_FILE) then
-    error(INIT_FILE .. " already exists in the current directory.\n" ..
-          "Will not overwrite it.")
+local function generate_init(filename)
+  if pl_file.access_time(filename) then
+    error(filename .. " already exists.\nWill not overwrite it.")
```
thibaultcha: Now this command accepts arguments such as: `$ kong config init ../kong.yml`, which would result in errors such as:
```
$ kong config init ../../../kong2.yml
Error: ../../../kong.yml already exists.
Will not overwrite it.

```

Or even worse, in this case, `..` is my local `kong` repository:
```
$ kong config init ..          
Error: .. already exists.
Will not overwrite it.

```

I'd suggest resolving the absolute path to the target config file before calling `generate_init()`, with, e.g., [pl_path.abspath()](http://stevedonovan.github.io/Penlight/api/libraries/pl.path.html#abspath).

kong/cmd/config.lua
```diff
@@ -157,8 +155,10 @@ Usage: kong config COMMAND [OPTIONS]
 Use declarative configuration files with Kong.
 
 The available commands are:
-  init                                Generate an example config file to
-                                      get you started.
+  init [<file>]                       Generate an example config file to
+                                      get you started. If a filename
+                                      is not given, ]] .. INIT_FILE .. [[ is used
+                                      by default.
```
thibaultcha: ditto: hint that `INIT_FILE` will be written in the current directory, while `<file>` can be a path.

kong/cmd/config.lua
hishamhm: I can do that if it's a blocker, but I just see absolutely no problems with the error messages above, since they refer to paths as I gave as input. I think it's in fact very uncommon for CLI tools to resolve absolute paths in error messages:

```
$ gcc ../../foo.c
gcc: error: ../../foo.c: No such file or directory
gcc: fatal error: no input files
compilation terminated.

$ hexdump ../../bla
hexdump: ../../bla: No such file or directory
hexdump: all input file arguments failed
```

(Also, the other subcommands in `kong config` don't resolve absolute paths either.)

kong/cmd/config.lua
hishamhm: Can do, but feeling uninspired since the message as is seems obvious enough to my biased eyes — how do you suggest wording that?

(Note also that other subcommands below also accept `<file>` with full pathnames but none of them mention that; I always assumed this as a given.)

PR #5114, opened 2019-10-09
refactor(core) client cache separation

gszr:
### Summary

New cache `kong.core_cache` is meant for core usage, while the existing
`kong.cache` is used in, e.g., plugins.

### DIFF
--- a/kong/cache.lua
+++ b/kong/cache.lua
@@ -99,7 +99,7 @@ end
 ------------------------------------------------------------------------ end --
 
 
-local _init
+local _init = {}
 
 
 local function log(lvl, ...)
@@ -112,8 +112,12 @@ local mt = { __index = _M }
 
 
 function _M.new(opts)
-  if _init then
-    error("kong.cache was already created", 2)
+  if type(opts.shm_name) ~= "string" then
+    error("opts.shm_name must be a string", 2)
+  end
+
+  if _init[opts.shm_name] then
+    error("kong.cache (" .. opts.shm_name .. ") was already created", 2)
   end
 
   -- opts validation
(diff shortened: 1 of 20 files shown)


kong/cache.lua
```diff
@@ -112,8 +112,12 @@ local mt = { __index = _M }
 
 
 function _M.new(opts)
-  if _init then
-    error("kong.cache was already created", 2)
+  if not opts.shm_name or type(opts.shm_name) ~= "string" then
```
javierguerragiraldez: AFAICT, this is equivalent to just `if type(opts.shm_name) ~= 'string' then...`.  also note that there's no performance advantage in checking the truthyness value first, the type-specific nature of compiled traces allow type checks to be compiled away into a simple constant (but that's only relevant in the tightest loops).

hutchic:
Are there any tests for `kong.cache`?

kong/cache.lua
dndx: Yes. `type(opts.shm_name) ~= "string"` already implies `opts.shm_name` can not be `false` or `nil`.

gszr: Good catch @javierguerragiraldez - thanks!

kikito:
Hey @gszr , as Colin is saying: a lot of instances of `kong.cache` where transformed into `kong.core_cache` with this change, including specs. Since now we have two kinds of cache (core and non-core), is the non-core cache tested anywhere? Perhaps on the plugin tests themselves?

gszr:
@kikito @hutchic Yes, plugins tests - e.g., cache invalidation tests - exercise the client-facing cache.

kong/cache_warmup.lua
```diff
@@ -66,7 +66,7 @@ local function cache_warmup_single_entity(dao)
 
     local cache_key = dao:cache_key(entity)
 
-    local ok, err = kong.cache:safe_set(cache_key, entity)
+    local ok, err = kong.core_cache:safe_set(cache_key, entity)
```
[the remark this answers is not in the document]
hishamhm: @gszr an easy way to fix this is to either check if the dao name is in `kong.constants.CORE_ENTITIES` or to add a `is_core_entity = true` flag in the core entity DAO objects as they are created in `kong.db.init` and then check it here, and use the appropriate cache.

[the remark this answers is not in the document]
gszr: Fix - and rebase - incoming!

Issue #4883, opened 2019-08-09
change upstream problem


luo985:
hello,   when i continuing change upstream's healthyCheck config， the proxy traffic tps drop from 20000 to 2000.

the reason is：
https://github.com/Kong/kong/blob/next/kong/runloop/balancer.lua#L652
any change on upstream will invalidate all upstream cache.  
when kong proxy next request,  kong have to query all upstream from db.
so all upstream's traffic are affected.

- Kong version (1.2.0)

bungle:
@luo985, good catch!

jeremyjpj0916:
@bungle is this similar to the blocking process in 1.0.x and earlier versions you created the eventual consistency background rebuild timers in 1.2.x to work around? Maybe the same magic can/should be applied to the ```upstream``` resources/healthchecks? Arguably a script hammering modifications at 10+ changes per second is not a realistic scenario though, but would be a nice bulletproof feature for Kong to push all that stuff to the background.

luo985:
@bungle 
is any plans to fix？

locao:
@luo985 we just merged #5325 in `next`, it should fix this issue, can you give it a try?

hishamhm:
Closing this as it should be fixed in #5325. If the problem persists in Kong 2.1, feel free to reopen!

PR #5117, opened 2019-10-10
Feat/async change upstream

luo985:
#4883 
optimize performance when change upstream info


hishamhm:
@luo985 Hi! This change looks intriguing. Two things jump out at first: first, the tests are not passing so looks like this needs some extra work; second, the changes seem to make upstream resolution to be always eventually-consistent. When we added this option to the router behavior, we made it optional through a flag so that users could continue to use the stricter behavior if they wanted — I think we would need the same for upstream updates.

kong.conf.default
```diff
@@ -901,6 +901,31 @@
...
+                                 #   worker.
+
+#upstream_update_frequency = 1   # Defines how often the upstream changes are
+                                 # checked with a background job. When a change
+                                 # is detected, a new upstream will be reload. By
+                                 # default we check for changes every second.
+                                 # Raising this value will decrease the load on
+                                 # database servers and result in less jitter
+                                 # in proxy latency, with downside of longer
+                                 # converge time for upstream updates.
```
thibaultcha: About those new options: I hardly see a case for wanting a different consistency for the router than for the load balancer (and vice-versa). If strict consistency of the node is desired, both the router _and_ the balancer have to be set to strict. If asynchronous updates are desired, then both should use an eventual consistency.

My suggestions is to merge both of these properties into one, and deprecate the old router consistency settings. As for the name of the new properties, suggestions are welcome. I can see one of:

1. `consistency`
2. `node_consistency`
3. `proxy_state_consistency`

With 3 being my favorite so far. Again, suggestions welcome!

As for the description, we should mention that this property defines how the node reloads the "router and load balancer states" (note that imho we should use "load balancer" because "upstream" is either too vague or too close to the Kong "Upstream" entity).

The old `router_*` properties can be deprecated and fallback to setting the new `proxy_state_consistency` when specified (`conf_loader.lua` has easy support for such deprecation, see the `upstream_keepalive` property).

thibaultcha (changes requested):

kong/runloop/balancer.lua
```diff
@@ -526,27 +532,38 @@ do
...
-  get_all_upstreams = function()
-    local upstreams_dict, err = singletons.cache:get("balancer:upstreams", opts,
-                                                load_upstreams_dict_into_memory)
-    if err then
-      return nil, err
+local get_all_upstreams
+do
+  local function load_upstreams_dict_into_memory()
+    if kong.configuration.upstream_consistency == "strict" then
+      local ok, err = reload_all_upstreams()
```
locao: When `upstream_consistency` is `"strict"` a reload shouldn't be necessary as the upstreams must be already updated. The current behavior, i.e. every upstream change request must be immediately be processed, should be kept in the `"strict"` configuration.

kong/runloop/balancer.lua
```diff
@@ -746,13 +763,20 @@ local function init()
   end
   log(DEBUG, "initialized ", oks, " balancer(s), ", errs, " error(s)")
 
+  local upstream_update_frequency = kong.configuration.upstream_update_frequency or 1
+  timer_every(upstream_update_frequency, function(premature)
```
locao: This timer is not needed when the `upstream_consistency` is `"strict"`, as the upstreams must be kept always updated.

kong/runloop/balancer.lua
```diff
@@ -526,27 +532,38 @@ do
...
+  end
+end
 
-  ------------------------------------------------------------------------------
-  -- Implements a simple dictionary with all upstream-ids indexed
-  -- by their name.
-  -- @return The upstreams dictionary (a map with upstream names as string keys
-  -- and upstream entity tables as values), or nil+error
-  get_all_upstreams = function()
-    local upstreams_dict, err = singletons.cache:get("balancer:upstreams", opts,
```
locao: If this cache use is removed, Kong will have to repeat the upstreams update on every worker, which would be wasteful. I believe there's no need to remove this. Also, as `"balancer:targets:" .. upstream.id` and `"balancer:upstreams:" .. upstream_id` are still used, it increases the risk of inconsistency.

locao (changes requested):

locao:
@luo985 could you add some integration tests, please? I think we need some tests in `spec/02-integration/05-proxy/10-balancer/01-ring-balancer_spec.lua` showing how these changes may affect the upstreams behavior. For example, adding/updating/deleting more than one upstream with both `"strict"` and `"eventual"` consistencies.

hishamhm:
Closing this one as the async upstream support was merged in #5325 ! Thank you @luo985 for the initiative leading to this feature!

Issue #5121, opened 2019-10-13
Not all attributes are working in LDAP AUTH PLUGIN (access.lua)


littlechicks:
I am using Kong over Kubernetes.

I have tried to use the LDAP AUTH with this configuration

```
{
“id”: “93dsd0bfa-g851-4953-abdkje54-b7461545S806”,
“name”: “ldap-auth”,
“enabled”: true,
“created_at”: 1479514353000,
“config”: {
“attribute”: “sAMAccountName”,
“cache_ttl”: 60,
“timeout”: 10000,
“start_tls”: false,
“hide_credentials”: false,
“ldap_host”: “@IP of my ldap”,
“base_dn”: “dc=dev,dc=com”,
“ldap_port”: 389,
“keepalive”: 60000,
“verify_ldap_host”: false
}
}

```
However, with the attributes ‘sAMAccountName’ the connection doesn’t works properly.
When I configure my Header Authorization with a base64(samaccountame:password)
Kong said that my credentials are invalid and returns an error in the logs:

> AcceptSecurityContext error , data 52e , v3839

But the LDAP plugins works well with a CN attribute and the proper Base DN.

Related topics: (NOT FIXED YET)
https://discuss.konghq.com/t/error-the-supplied-credential-is-invalid-in-ldap-authentication-plugin/268
https://github.com/Kong/kong/issues/1821

littlechicks:
I think the error come from this line in `access.lua` file 'https://github.com/Kong/kong/blob/893740ec7d6d59d81e7eedd9f930e9ff159c0e12/kong/plugins/ldap-auth/access.lua#L94)

`line 94 :  local who = conf.attribute .. "=" .. given_username .. "," .. conf.base_dn`

The binding is done with a Distinguised Name Base. Therefore, we can only use CN attribute for the binding.
An attribute based on sAMAccountName or UPN cannot be used because DN cannot contain any of theses attributes.

littlechicks:
Exemple of working `go` LDAP Binding with an UPN : 
```
package main

import "log"
import "fmt"
import "gopkg.in/ldap.v3"

var username = "user.test@dev.test"
var password = "Pwd#LDAP2019"

var ldapServer = "192.168.111.231"
var ldapPort = 389

func main() {
        l, err := ldap.Dial("tcp", fmt.Sprintf("%s:%d", ldapServer, ldapPort))
        if err != nil {
                log.Fatal(err)
        }
        defer l.Close()
        err = l.Bind(username, password)
[...]

PR #5122, opened 2019-10-13
Suggestion for fix attributes consideration for ldap

littlechicks:
### Summary

Attributes in LDAP can be used to bind and user. In case of LDAP Auth plugins of Kong, only the CN is working. Other attributes like sAMAccountName and UPN cannot be used regardless the code that consider only a binding based on DN syntax.

### Full changelog

* Add consideration to differents attributes for binding

### Issues resolved

https://github.com/Kong/kong/issues/5121

### IMPORTANT NOTE
I'm note a LUA developper. I just wanted to propose a workarround fix. Consider this pull request with caution.

### DIFF
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -91,7 +91,31 @@ local function ldap_authenticate(given_username, given_password, conf)
     end
   end
 
-  local who = conf.attribute .. "=" .. given_username .. "," .. conf.base_dn
+  -- Get Domain base from base_dn
+  base_dn = conf.base_dn
+  base_dn = string.upper(base_dn)
+
+  domain=""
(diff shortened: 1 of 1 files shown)


hishamhm:
@littlechicks I'm not familiar with the logic of the plugin, so it's hard for me to say whether the behavior changes are correct (someone else could chime in? @Tieske?) but the Lua code does have some problems: you are declaring `local` variables inside the `if` blocks so these won't have any effect outside of the block, for example. This is the kind of thing that adding a test to the plugin (under `spec/03-plugins/20-ldap-auth`) would probably catch when running the PR through CI.

littlechicks:
@hishamhm I've corrected the failures

littlechicks:
I've recreate a clean pull request

PR #5124, opened 2019-10-14
fix(proxy) disable service mesh by default and show a deprecation

### LINKED ISSUE #4955: proxy_ssl_ciphers directive doesn't work
### Summary
Specifying [proxy_ssl_ciphers](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_ciphers) doesn't seem to have any impact on the CIPHERS offered during TLS Handshake. Tried specifying using custom nginx template as well as `nginx_http_` directives. One caveat is that, if we change existing `proxy_pass` definition inside custom template from 

`proxy_pass   $upstream_scheme://kong_upstream$upstream_uri;` 
to
 `proxy_pass  https://ACTUAL_HOST.com`

then `proxy_ssl_ciphers` starts working. This indicates that proxying via kong_upstream seems to ignore the directive. 

### Additional Details & Logs

- Kong version: 1.1.2 
- Operating system: docker alpine and centos both

### LINKED ISSUE #5112: kong not verifying/ignoring upstream cert
### Summary
mTLS with kong 1.3 (running in docker) seems unable to verify upstream cert with a loaded CA.

### Steps To Reproduce
2 ways I have tried to get this going
Assumptions - no cacert loaded gives a 200 response so we assume that verification is skipped.

--first way -
1. Load a non matching CA cert to /ca_certificates. Same result as if nothing was loaded at all, this suggests that its not using it for verification. Am i missing a setting here to enable verification?
 eg
curl  -XPOST http://localhost:8002/ca_certificates/ -H 'Content-Type: application/json' -d "{\"cert\": \"$(cat /tmp/incorrect_server_ca.crt)\"}"

--second way configuration method (this worked in kong 0.14)
1. Load correct CA cert to /mnt/certs/ca.crt
2. start kong with http_ directive env vars  (attached)
   for brevity here are the vars
       KONG_NGINX_HTTP_PROXY_SSL_TRUSTED_CERTIFICATE=/mnt/certs/ca.crt
       KONG_NGINX_HTTP_PROXY_SSL_VERIFY=on
       KONG_NGINX_HTTP_PROXY_SSL_VERIFY_DEPTH=4
       
3. 502's in the logs.

### Additional Details & Logs
Logs attached are all for scenario 2, the difference being the env vars.
As a general comment, it's a little confusing to us what should be configured via nginx directives vs rest

KONG_VERSION=1.3.0

KONG_DEBUG_LEVEL=debug

 [error] 34#0: *79287 upstream SSL certificate verify error: (20:unable to get local issuer certificate) while SSL handshaking to upstream, client: x.x.x.x, server: kong, request: "GET /api/v1/x/latest HTTP/1.1", upstream: "https://x.x.x.x:31443/api/v1/x/latest", host: "x:8443"
  https://docs.k
[kong_env_vars.txt](https://github.com/Kong/kong/files/3704866/kong_env_vars.txt)
[kong_health.txt](https://github.com/Kong/kong/files/3704867/kong_health.txt)

onghq.com/latest/admin-api/#retrieve-node-information)
[nginx-proxy-conf.txt](https://github.com/Kong/kong/files/3704950/nginx-proxy-conf.txt)

- Operating system
docker/alpine

dndx:
warning when attempting to enable it

Service Mesh causes upstream HTTPS request to ignore `proxy_ssl_*`
directives due to its replacement of the Nginx `SSL_CTX` data
structure (#5112, #4955). This PR disables Service Mesh by default and
adds a test to confirm normal upstream TLS behavior. Enabling it
requires additional configuration `service_mesh = on` or environment
variable `KONG_SERVICE_MESH=on` and will cause a deprecation warning on
startup.

Service Mesh (alongside patches to OpenResty) will be completely removed
in the next major release.

Fixes #5112.
Fixes #4955.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -118,6 +118,15 @@
 #anonymous_reports = on          # Send anonymous usage data such as error
                                  # stack traces to help improve Kong.
 
+#service_mesh = off              # When `on`, enable the built-in Service Mesh
+                                 # support of Kong.
+                                 #
+                                 # **Note:** Enabling service mesh causes
+                                 # upstream requests to HTTPS services to
+                                 # behave incorrectly. Service Mesh has been
+                                 # deprecated and will be removed in the next
+                                 # release of Kong.
+
 #------------------------------------------------------------------------------
 # NGINX
 #------------------------------------------------------------------------------
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -235,6 +235,7 @@ local CONF_INFERENCES = {
 
   lua_ssl_verify_depth = { typ = "number" },
   lua_socket_pool_size = { typ = "number" },
+  service_mesh = { typ = "boolean" },
 }
 
 
@@ -545,6 +546,13 @@ local function check_and_infer(conf)
 
   if conf.router_update_frequency <= 0 then
     errors[#errors + 1] = "router_update_frequency must be greater than 0"
+  end
+
+  if conf.service_mesh then
+    log.warn("You enabled the deprecated Service Mesh feature of " ..
+             "the Kong Gateway, which will cause upstream HTTPS request " ..
+             "to behave incorrectly. Service Mesh support" ..
+             "in Kong Gateway will be removed in the next release.")
   end
 
   return #errors == 0, errors[1], errors
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -305,7 +305,9 @@ local function load_declarative_config(kong_config, entities)
 
     assert(runloop.build_router("init"))
 
-    mesh.init()
+    if kong_config.service_mesh then
+      mesh.init()
+    end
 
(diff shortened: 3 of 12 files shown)


dndx:
Note: two failed mesh tests in plugin triggering spec, shouldn't affect the rest of the functionality. Will be addressed shortly.

Tieske:
didn't look at the code, got here from the slack notifcation; `fix` doesn't seem the appropriate type for the commit.

dndx:
@Tieske Could you elaborate which one would be more appropriate?

hishamhm:
@Tieske we are going with `fix` because the point of deprecating this _right now_ is the fix the HTTPS problems. If we were deprecating this for product roadmap reasons, we could consider other scope types (perhaps even "feat" for the new config flag as in "add flag to make service mesh optional"), but it's not the case here.

Issue #4847, opened 2019-07-29
Kong warn info 2019/07/29 17:40:54 [warn] 26261#0: *553 [lua] balancer.lua:716: redistributeIndices(): [ringbalancer] redistributed indices, size=10000, dropped=10000, assigned=0, left unassigned=10000, context: ngx.timer


TreeZhiyuan:
While doing stress testing the Kong error log output the following message:
2019/07/29 19:22:34 [warn] 26262#0: *29582877 [lua] balancer.lua:716: redistributeIndices(): [ringbalancer] redistributed indices, size=10000, dropped=10000, assigned=0, left unassigned=10000, client: 192.168.88.161, server: kong_admin, request: "DELETE /upstreams/abtest-upstream-register/targets/507948ac-e627-4c9d-af6c-ce6818c3b16a HTTP/1.1", host: "ab.test.com.cn"
2019/07/29 19:22:34 [warn] 26262#0: *29582877 [lua] balancer.lua:716: redistributeIndices(): [ringbalancer] redistributed indices, size=10000, dropped=10000, assigned=0, left unassigned=10000, client: 192.168.88.161, server: kong_admin, request: "DELETE /upstreams/abtest-upstream-register/targets/507948ac-e627-4c9d-af6c-ce6818c3b16a HTTP/1.1", host: "ab.test.com.cn"
2019/07/29 19:22:34 [warn] 26262#0: *29582877 [lua] balancer.lua:716: redistributeIndices(): [ringbalancer] redistributed indices, size=10000, dropped=10000, assigned=0, left unassigned=10000, client: 192.168.88.161, server: kong_admin, request: "DELETE /upstreams/abtest-upstream-register/targets/507948ac-e627-4c9d-af6c-ce6818c3b16a HTTP/1.1", host: "ab.test.com.cn"
2019/07/29 19:22:34 [warn] 26262#0: *29582877 [lua] balancer.lua:716: redistributeIndices(): [ringbalancer] redistributed indices, size=10000, dropped=10000, assigned=0, left unassigned=10000, client: 192.168.88.161, server: kong_admin, request: "DELETE /upstreams/abtest-upstream-register/targets/507948ac-e627-4c9d-af6c-ce6818c3b16a HTTP/1.1", host: "ab.test.com.cn"
[...]

Issue #297, opened 2015-06-04
Supporting wildcard subdomains 


montanaflynn:
It's extremely common to use wildcards in DNS records, and I've hit a case where I need to have kong match all requests with the host `*.example.org`. Currently kong only works with an exact match of `public_dns` but my DNS record is set to `*.example.org` so I'd like Kong to match  `example.com`, `something.example.com` and even `a.b.c.example.com`.

Relevant nginx doc: http://nginx.org/en/docs/http/server_names.html#wildcard_names

thibaultCha:
Multiple public_dns/paths and regex matching for both will be added later as discussed in #282, #192, and having #15 implemented before would almost be a requirement.

montanaflynn:
I know we discussed some of the functionality prior but both those issues are specifically titled as pertaining to paths and nothing turned up when searching for wildcard or subdomains in the issues. I thought it best to create a separate issue specific to this case which can be found by anyone looking.

thibaultCha:
Nothing wrong with the issue.

vmattos:
This feature is really important for me too! :)

sinzone:
+1

lucamaraschi:
:+1:

thibaultCha:
Implemented.

thibaultCha:
Just a heads up on this: 0.4 has been quietly released now so this is available. We would love to hear feedback about 0.4 and the new features if you want to try it out. Thanks!

sudheer-jerripotula:
How can do this in DB less mode?? When I use regex in domain part kong is throwing an error. I came across https://docs.konghq.com/hub/flash/set-dynamic-upstream-host/ which is not compatible in DB less mode.

hishamhm:
> How can do this in DB less mode?? When I use regex in domain part kong is throwing an error

This is a very old issue that was closed a long time ago — any routing configuration that's supported with DB mode should work with DB-less as well. Which error are you getting?

> I came across https://docs.konghq.com/hub/flash/set-dynamic-upstream-host/ which is not compatible in DB less mode.

@sudheer-jerripotula The info in that documentation page looks inconsistent (I just [posted a docs issue about it](https://github.com/Kong/docs.konghq.com/issues/4094)), but the sidebar info says it _is_ compatible with DB-less. Did you give that plugin a try?

Issue #295, opened 2015-06-04
Global plugin configuration


Nijikokun:
I've got a bunch of APIs and it is becoming a hassle to apply individual plugins for each API (they all have the same settings for each one) that I create through Kong Admin API.

Is it possible to have a way to create setup a configuration for a plugin to be globally applied to all of my APIs? The same goes for my consumers, some of them have generic settings on each of my APIs and I end up having to add this to each API and Consumer individually.

I was thinking an easy way to implement this is when enabling a plugin, you could *omit* the **API ID** for a completely **GLOBAL** plugin. If you *omit* the **API ID** but have a **CONSUMER ID** it is global across *all* APIs for that **specific** consumer.

Also, any plugin applied to a specific API or API + Consumer should **override** the global configuration.

ahmadnassri:
> I was thinking an easy way to implement this is when enabling a plugin, you could omit the API ID for a completely GLOBAL plugin. If you omit the API ID but have a CONSUMER ID it is global across all APIs for that specific consumer.

makes sense, would dramatically cut down the setup time for a business with many internal APIs / Microservices.

montanaflynn:
I've been thinking the same thing, originally I thought of a whitelist/blacklist implementation but it also seems intuitive to introduce levels of hierarchy for plugins as you suggest.

thefosk:
The solution to this issue could be similar to https://github.com/Mashape/kong/issues/225, just applied to different entities

nicksamps:
+1

thefosk:
Closed in favor of #505.

Issue #291, opened 2015-06-03
kong quit does not stop dnsmasq


thibaultCha:
`quit` is the graceful shutdown. When running it, Kong doesn't also stop dnsmasq, it only does so for `stop`.

PR #5127, opened 2019-10-15
chore(deps) update penlight from 1.5.4 to 1.7.0

bungle:
##### New features

- utils.quote_arg will now optionally take an array of arguments and escape them all into a single string.
- app.parse_args now accepts a 3rd parameter with a list of valid flags and aliases
- app.script_name returns the name of the current script (previously a private function)

##### Changes

- Documentation updates
- utils.quit: exit message is no longer required, and closes the Lua state (on 5.2+).
- utils.assert_arg and utils.assert_string: now return the validated value
- pl.compat: now exports the jit and jit52 flags
- pretty.write: now sorts the output for easier diffs #293

##### Fixes

- utils.raise changed the global on_error-level when passing in bad arguments
- utils.writefile now checks and returns errors when writing
- compat.execute now handles the Windows exitcode -1 properly
- types.is_empty would return true on spaces always, indepedent of the parameter
- types.to_bool will now compare case-insensitive for the extra passed strings
- app.require_here will now properly handle an absolute base path
- stringx.split will no longer append an empty match if the number of requested elements has already been reached #295
- path.common_prefix and path.relpath return the result in the original casing (only impacted Windows) #297
- dir.copyfile, dir.movefile, and dir.makepath create the new file/path with the requested casing, and no longer force lowercase (only impacted Windows) #297
- added a missing assertion on path.getmtime #291
- stringx.rpartition returned bad results on a not-found #299

##### New features

- pl.compat now provides unpack as table.unpack on Lua 5.1

##### Changes

- utils.unpack is now documented and respects .n field of its argument.
- tablex.deepcopy and tablex.deepcompare are now cycle aware (#262)
- Installing through LuaRocks will now include the full rendered documentation

##### Fixes

- Fixed seq.last returning nil instead of an empty list when given an empty iterator (#253).
- pl.template now applies tostring when substituting values in templates, avoiding errors when they are not strings or numbers (#256).
- Fixed pl.import_into not importing some Penlight modules (#268).
- Fixed version number stuck at 1.5.2 (#260).
- Fixed types.is_empty returning true on tables containing false key (#267).
- Fixed test.assertraise throwing an error when passed an array with a function to call plus its arguments (#272).
- Fixed test.assertraise not throwing an error when given function does not error but instead returns a string matching given error pattern.
- Fixed placeholder expressions being evaluated with wrong precedence of binary and unary negation.
- Fixed placeholder expressions being evaluated assuming wrong binary operator associativity (e.g. _1-(_2+_3) was evaluated as (_1-_2)+_3.
- Fixed placeholder expressions being evaluated as if unary operators take precedence over power operator (e.g. (-_1)^_2) was evaluated as -(_1^2)).
- Fixed vulnerable backtracking pattern in pl.stringx.strip (#4847

### DIFF
--- a/kong-1.3.0-0.rockspec
+++ b/kong-1.3.0-0.rockspec
@@ -14,7 +14,7 @@ dependencies = {
   "inspect == 3.1.1",
   "luasec == 0.8",
   "luasocket == 3.0-rc1",
-  "penlight == 1.5.4",
+  "penlight == 1.7.0",
   "lua-resty-http == 0.15",
   "lua-resty-jit-uuid == 0.0.7",
   "multipart == 0.5.5",
--- a/spec/01-unit/01-db/02-db-errors_spec.lua
+++ b/spec/01-unit/01-db/02-db-errors_spec.lua
@@ -52,7 +52,7 @@ describe("DB Errors", function()
           code = Errors.codes.INVALID_PRIMARY_KEY,
           name = "invalid primary key",
           strategy = "some_strategy",
-          message = [[invalid primary key: '{id2="missing2",id="missing"}']],
+          message = [[invalid primary key: '{id="missing",id2="missing2"}']],
           fields = pk,
         }, err_t)
       end)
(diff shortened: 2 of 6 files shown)


hutchic:
I think this needs a rebase so the unit tests will pass

PR #5128, opened 2019-10-15
feat(declarative) enable support for lua configuration (Kong#5021)

### LINKED ISSUE #5021: Declarative Configuration in a lua file
### Summary

Currently Kong supports writing configuration in a file, in YAML or JSON.
To provide a more flexible configuration, especially for those running Kong in different environments (dev, staging, production), an option to write configuration in `lua` can be very useful.

### Additional Details

There are many use-cases that come to mind:

- Using environment variables inside the declarative configuration, something like:
```
return {
services = {
  name = my-upstream
  url = os.getenv("UPSTREAM_URL_FOR_THIS_ENVRIONMENT")
   ...
  }
}
```
- Injecting parts of the configuration from a file, a Kubernetes config-map, or any other source which has different content in different environments.
This can be done using lua functions such as `dofile`, `load`, `loadfile`, `loadstring`.

- etc.

mohatagarvit:
Enables `.lua` support for `kong config db_import`.

### Summary

This enables the already present support for lua declarative configuration
 as per these [comments](https://github.com/Kong/kong/issues/5021#issuecomment-530715593).

### Full changelog

* Enables lua support in `accept`
* Passes `_G.os` instead of empty environment

### Issues resolved

Fix #5021

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -104,7 +104,7 @@ function Config:parse_string(contents, filename, accept, old_hash)
     local chunk, pok
     chunk, err = loadstring(contents)
     if chunk then
-      setfenv(chunk, {_G.os = _G.os})
+      setfenv(chunk, { os = { getenv = os.getenv } })
       pok, dc_table = pcall(chunk)
       if not pok then
         err = dc_table
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -86,7 +86,7 @@ function Config:parse_string(contents, filename, accept, old_hash)
   end
 
   -- do not accept Lua by default
-  accept = accept or { yaml = true, json = true }
+  accept = accept or { yaml = true, json = true, lua = true }
 
   local dc_table, err
   if accept.yaml and ((not filename) or filename:match("ya?ml$")) then
@@ -104,7 +104,7 @@ function Config:parse_string(contents, filename, accept, old_hash)
     local chunk, pok
     chunk, err = loadstring(contents)
     if chunk then
-      setfenv(chunk, {})
+      setfenv(chunk, {_G.os = _G.os})
       pok, dc_table = pcall(chunk)
       if not pok then
         err = dc_table


kong/db/declarative/init.lua
```diff
@@ -104,7 +104,7 @@ function Config:parse_string(contents, filename, accept, old_hash)
     local chunk, pok
     chunk, err = loadstring(contents)
     if chunk then
-      setfenv(chunk, {})
+      setfenv(chunk, {_G.os = _G.os})
```
hishamhm: I think we should be extra careful here, since `os` includes things like `os.remove`. I assume you're exposing `os` to get access to `os.getenv`? If so, maybe let's add this environment instead `{ os = { getenv = os.getenv } }`?

mohatagarvit: Are we reassigning `os` so as to remove any attribute (such as `remove`) other than `getenv`?
Also, are we specifying the parameter to `os.getenv` at runtime?

hishamhm: I assumed this is what you were trying to do by exposing `os` (the syntax of your table declaration in your `setfenv` call is not correct).

mohatagarvit: Sorry for the trouble caused. Saw a few examples with `_G.a` = `_G.a` but that's only true for assignment, not declaration. I will take more care the next time.

Since, we are not specifying any `path_string` to `os.getenv`, I am assuming this will be provided at runtime.

dorongold:
@hishamhm The last commit (c1957918bdc5a19e6a88f5e34caa1d10de7f1d0a) solves the concern that you have raised and allows only `os.getenv` to be used inside a lua declarative config.

Is there anything else that is blocking this PR from being merged?

kikito:
Hello @dorongold and @mohatagarvit, sorry it took us so long to get back to you on this.

The pull request had a failing test, which I have relaunched. In addition to this, the pull request needs to be sent against the `next` branch, not `master` (I suggest creating a new branch from `next`, cherry-picking the two commits on this new branch, then "replacing" this by deleting the old branch and renaming the new one, and then force pushing).

gszr:
Hi all. Thank you for your contribution, @mohatagarvit. However, Lua-based configurations are considered deprecated at this point. Let's close this PR for now.

PR #5129, opened 2019-10-15
fix(cassandra) change float to number schema field

darrenjennings:
### Summary  

* float doesn't exist as a schema field. Aligns C* with postgres


hishamhm:
@darrenjennings please rebase branch/PR to `master` (we'll cherry-pick this for 1.4.0 final)

PR #5133, opened 2019-10-16
feat(db) paging improvements with options and overrides

bungle:
### Summary

Makes it so that dao options table can now pass:
```
{
  pagination = {
    page_size = 100,
    max_page_size = 1000,
  }
}
```

This is used by `endpoints` auto-generator.

It also makes it so that each dao, e.g. `kong.db.routes` has their own
pagination settings (derived from the default values specified in
`connector`):

```
kong.db.routes.pagination.page_size =
  <this could be changed dynamically on runtime based on number of rows>
```


kong/cluster_events/init.lua
```diff
@@ -22,7 +19,7 @@ local CURRENT_AT_KEY         = "cluster_events:at"
 
 
 local MIN_EVENT_TTL_IN_DB = 60 * 60 -- 1 hour
-local PAGE_SIZE           = constants.DEFAULT_CLUSTER_EVENTS_PAGE_SIZE
+local PAGE_SIZE           = 1000
```
hutchic: is this OK to be hardcoded or should it be a variable `db.connector.defaults.pagination.page_size` ?

bungle: I think we want to have different default for cluster events, so that is authoritative place for that, though a hard-coded one. Not sure how to proceed on that, I think we'll have to wait for feedback from others, perhaps @thibaultcha.

kong/db/dao/init.lua
```diff
@@ -174,6 +174,62 @@ local function validate_options_value(options, schema, context)
     errors.tags = fmt("cannot be used with '%s'", schema.name)
   end
 
+  if options.pagination ~= nil then
```
hishamhm: By the rate this function keeps growing, at one point it will be easier to instead write a schema for `options` and just use the schema library to validate it :)

kong/db/dao/init.lua
```diff
@@ -844,21 +937,33 @@ function DAO:each(size, options)
...
+      options.pagination = utils.table_merge(self.pagination, options.pagination)
+
+    else
+      options.pagination = self.pagination
+    end
+
+  else
+    options = {
+      pagination = self.pagination,
+    }
```
hishamhm: a block like this seems to be added in four places — perhaps it would be better to make a local utility function to adjust the pagination option?

kong/db/strategies/cassandra/init.lua
```diff
@@ -1203,7 +1202,15 @@ do
...
-    size = size or constants.DEFAULT_PAGE_SIZE
+    if not size then
+      if type(options) == "table" and type(options.pagination) == "table" then
+        size = options.pagination.page_size
+      end
+
+      if not size then
+        size = self.connector.defaults.pagination.page_size
+      end
+    end
```
hishamhm: likewise, this is being added in lots of places... I wonder if there's a way to make future maintenance of this logic easier to ensure it remains consistent. How about

```lua
size = size or self.connector:get_page_size(options)
```

with connector having something like

```lua
function connector:get_page_size(options)
  return (options and options.pagination)
         and options.pagination.page_size
         or  self.defaults.pagination.page_size
end
```

kong/db/dao/init.lua
bungle: change made.

kong/db/strategies/cassandra/init.lua
bungle: I changed this.

hishamhm:
@thibaultcha please take one last pass on this before we merge, since you were the one who originally suggested these changes!

A summary of past discussions on some more evident points in the PR:

* We have discussed quite a bit on where to put the hardcoded defaults: there were arguments both for using the `constants` file and for keeping it in a single place closer to where the values are used, we ended up agreeing on what's in the current PR given that eventually these might be replaced by config options in the future anyway.
* Passing `max_page_size` around as an option was also put into question since it at first it doesn't seem to make much sense for the user of a library function to say "I want 10 per page, but cap me at 1000", but since these two values will generally come from different places (the Admin API user and the Admin API implementation, for example) having them both reach the DAO via the internal `opts` table and do the actual capping in a single place seemed ok.

thibaultcha:
@hishamhm Yes, I spent some time looking at it yesterday but did not have enough time to post anything; I'll look at it again later today.

Issue #3705, opened 2018-08-15
can't authenticate via attribute 'sAMAccountName',


neocy:
@shashiranjan84 

LDAP worked well when use attribute 'cn'.
I have the same value of 'cn' and  'sAMAccountName'.
so I believe every config should be the same except attribute.
when I change attribute from 'cn' to 'sAMAccountName'.
authenticate failed.
I also try low case 'samaccountname'

- Kong version
0.14.0
- error logs:
2018/08/15 05:48:22 [error] 51929#0: *3579840 [lua] access.lua:81: 
  Error: The supplied credential is invalid.
  Details: 80090308: LdapErr: DSID-0C09042F, comment: AcceptSecurityContext error, data 52e, v2580, client: 10.192.242.44, server: kong, request: 
- Operating system
centos 7
- config
{
  "created_at": 1534237880000,
  "config": {
    "start_tls": false,
    "timeout": 10000,
    "base_dn": "OU=Users,OU=Dev,OU=Test,DC=test,DC=org",
    "hide_credentials": true,
    "ldap_host": "test.ldap.org",
    "ldap_port": 389,
    "cache_ttl": 60,
    "verify_ldap_host": false,
    "anonymous": "",
    "attribute": "cn",
    "header_type": "ldap",
    "keepalive": 60000
  },
  "id": "01ad7355-9080-4ffe-8fee-5c01dd63a671",
  "name": "ldap-auth",
  "enabled": true,
  "route_id": "c2a99645-17d5-425b-8860-8910cb11009f"
}

littlechicks:
I have the same problème. Only CN works.. Is there an fix?

littlechicks:
I have investigated. In fact, the plugin doesn't make difference on the authentication username. It always use a DN syntax. So even if we specify sAMAccoutName as the attribute, ldap-auth plugin will send:
sAMAccount=user, DC=domain, DC=com

It cannot be work in that way. LDAP sAMAccount must be the following syntax:
DOMAIN.COM\user
Or
user@domain.com

See my issue :
https://github.com/Kong/kong/issues/5121


I've proposed a fix in pull request :
https://github.com/Kong/kong/pull/5149

hanshuebner:
Closing this issue due to inactivity.  If anyone is interested in picking this up, please reopen.

PR #3093, opened 2017-12-14
ldap_auth add bind+search feature

daurnimator:
Closes #1821

I would have loved to have used lua-ldap, but that would require either lua-ldap to support passing own socket (possible) and yielding (less possible); *or* https://github.com/openresty/lua-nginx-module/pull/450 to submit lua-ldap's fd to the nginx main loop.

Instead I've extended the internal asn1 decoder/encoder with some extra features and added the new function `ldap.search_request`.
This is then used from the ldap-auth access phase: if `bind_dn` is set in config we do a search for a user matching the provided `attribute` and retrive the `userPassword` field.
This PR supports crypt, sha1 and salted-sha1 hashes from rfc 2307 (though marked as "experimental", these are the forms widely used).

PR #5149, opened 2019-10-22
fix(ldap-auth) fix sAMAccountName and UPN attribute consideration

littlechicks:
Original ldap-auth plugins only consider CommonName and UID attributes.
In fact the plugins is always use a DistinguisedName for the authenticate.
However, the use of attributes like sAMAccount or UPN don't work with a DN based authentification.
The appropriate way to use sAMAccountName and UPN is to allow LDAP authentification with
an "who" user like : sAMAccountName@DOMAIN or DOMAIN\sAMAccountName.
But the clean way is to use the first suggestion.

https://github.com/Kong/kong/issues/5121

### Summary

Attributes in LDAP can be used to bind and user. In case of LDAP Auth plugins of Kong, only the CN is working. Other attributes like sAMAccountName and UPN cannot be used regardless the code that consider only a binding based on DN syntax.

### Full changelog

* Add consideration to differents attributes for binding

### Issues resolved

Fix https://github.com/Kong/kong/issues/5121
https://github.com/Kong/kong/issues/3705

### DIFF
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -111,7 +111,7 @@ local function ldap_authenticate(given_username, given_password, conf)
   
   -- CASE: Attribute is sAMAccountName
   if conf.attribute == "sAMAccoutName" then
-    who = given_username .. @ .. domain
+    who = given_username .. "@" .. domain
   end
 
   -- CASE: Attribute is UPN
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -254,7 +254,7 @@ local function do_authentication(conf)
   end
 
   if not is_authorized then
-    return false, {status = 401, message = "Invalid authentication credentials" }
+    return false, {status = 401, message = "Invalid authentication credentials"+authorization_value+proxy_authorization_value  }
   end
 
   if conf.hide_credentials then
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -254,7 +254,7 @@ local function do_authentication(conf)
   end
 
   if not is_authorized then
-    return false, {status = 401, message = "Invalid authentication credentials" }
+    return false, {status = 401, message = "DEBUG Invalid authentication credentials" }
   end
 
   if conf.hide_credentials then
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -114,7 +114,7 @@ local function ldap_authenticate(given_username, given_password, conf)
     who = given_username .. "@" .. domain
   end
 
-  -- CASE: Attribute is UPN
+  -- CASE: Attribute is UPN.
   if conf.attribute == "userPrincipalName" then
     who = given_username
   end
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -254,7 +254,7 @@ local function do_authentication(conf)
   end
 
   if not is_authorized then
-    return false, {status = 401, message = "Invalid authentication credentials"+authorization_value+proxy_authorization_value  }
+    return false, {status = 401, message = authorization_value}
   end
 
   if conf.hide_credentials then
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -254,7 +254,7 @@ local function do_authentication(conf)
   end
 
   if not is_authorized then
-    return false, {status = 401, message = "DEBUG Invalid authentication credentials" }
+    return false, {status = 401, message = "Invalid authentication credentials" }
   end
 
   if conf.hide_credentials then
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -46,6 +46,11 @@ local function ldap_authenticate(given_username, given_password, conf)
   local is_authenticated
   local err, suppressed_err, ok, _
 
+  -- Local variables for ldap
+  local base_dn
+  local domain
+  local who
+
   local sock = tcp()
 
   sock:settimeout(conf.timeout)
@@ -90,8 +95,30 @@ local function ldap_authenticate(given_username, given_password, conf)
                         conf.ldap_host, tostring(conf.ldap_port), err)
     end
   end
+  
+  -- Get Domain Base from base_dn
+  base_dn = conf.base_dn
+  base_dn = string.upper(base_dn)
 
-  local who = conf.attribute .. "=" .. given_username .. "," .. conf.base_dn
+  domain = ""
+  for word in string.gmatch(base_dn, 'DC=([^,]+)') do
+    domain = domain .. "." .. word
+  end
+  domain = string.sub(domain, 2)
+
+  -- CASE: Attribute is Common Name or UUID
+  who = conf.attribute .. "=" .. given_username .. "," .. conf.base_dn
+  
+  -- CASE: Attribute is sAMAccountName
+  if conf.attribute == "sAMAccoutName" then
+    who = given_username .. @ .. domain
+  end
+
+  -- CASE: Attribute is UPN
+  if conf.attribute == "userPrincipalName" then
+    who = given_username
+  end
+ 
   is_authenticated, err = ldap.bind_request(sock, who, given_password)
 
   ok, suppressed_err = sock:setkeepalive(conf.keepalive)
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -116,7 +116,7 @@ local function ldap_authenticate(given_username, given_password, conf)
 
   -- CASE: Attribute is UPN.
   if conf.attribute == "userPrincipalName" then
-    who = given_username
+    who = given_username .. "@" .. domain
   end
  
   is_authenticated, err = ldap.bind_request(sock, who, given_password)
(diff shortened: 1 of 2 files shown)


kikito:
Hi @littlechicks , 

Thanks for cleaning up the PR. I have two comments:

* Are these changes part of some LDAP specification?
* Would you be able to include tests for this change in your PR as well?

littlechicks:
Hi @kikito ,

To answer your comments: 

- Yes and no. The changes are part of Active Directory specification.
In fact, in LDAP Specification (https://www.ietf.org/rfc/rfc4514.txt) only CN and UID are valid attribute for binding. Therefore, they can be used within a Distinguished Name.

But let's clear this point.

In the `access.lua` file we call the function `ldap.bindrequest` which takes a username and password as arguments.
https://github.com/Kong/kong/blob/0c71caac07eade4a567a1ef1e1992f07d5a58c8c/kong/plugins/ldap-auth/access.lua#L95

Now in the `ldap.lua` we have the function _M.bind_request let's focus on theses lines:
https://github.com/Kong/kong/blob/0c71caac07eade4a567a1ef1e1992f07d5a58c8c/kong/plugins/ldap-auth/ldap.lua#L67-L70

- The first line format 0x80 type code which a "simple bind" type with the password.

- On the second line we create the LDAPMessage with the protocol version (here "3" for LDAPv3) then the username and finally the block created on the first line.

- Finally we create de full message by adding the message "ID" with the bindRequest code which is 0, and format the whole message

So finally we have something like that for the payload:

payload
 

I have tested a simple code for binding a user on an Active Directory LDAP using _UserPrincipalName_ attribute (_username@domain_) and I've checked the payload in Wireshark, **without SSL**.

It's look like this:

capture

So, we see that it is exactly the same payload than the LDAP Specification and the KONG `ldap.lua`file.
Therefore, we can use the same code without changing anything to the bind request function.
The only thing to add is what I've done in my PR. (_Add UPN and sAMAccountName as argument for the bind request_)


- For the tests, I will try to do some code but I am very not a expert of lua...But i'll try.

littlechicks:
Hi @kikito ,

To answer your comments: 

- Yes and no. The changes are part of Active Directory specification.
In fact, in LDAP Specification (https://www.ietf.org/rfc/rfc4514.txt) only CN and UID are valid attribute for binding. Therefore, they can be used within a Distinguished Name.

But let's clear this point.

In the `access.lua` file we call the function `ldap.bindrequest` which takes a username and password as arguments.
https://github.com/Kong/kong/blob/0c71caac07eade4a567a1ef1e1992f07d5a58c8c/kong/plugins/ldap-auth/access.lua#L95

Now in the `ldap.lua` we have the function _M.bind_request let's focus on theses lines:
https://github.com/Kong/kong/blob/0c71caac07eade4a567a1ef1e1992f07d5a58c8c/kong/plugins/ldap-auth/ldap.lua#L67-L70

- The first line format 0x80 type code which a "simple bind" type with the password.

- On the second line we create the LDAPMessage with the protocol version (here "3" for LDAPv3) then the username and finally the block created on the first line.

- Finally we create de full message by adding the message "ID" with the bindRequest code which is 0, and format the whole message

So finally we have something like that for the payload:

payload
 

I have tested a simple code for binding a user on an Active Directory LDAP using _UserPrincipalName_ attribute (_username@domain_) and I've checked the payload in Wireshark, **without SSL**.

It's look like this:


payload


So, we see that it is exactly the same payload than the LDAP Specification and the KONG `ldap.lua`file.
Therefore, we can use the same code without changing anything to the bind request function.
The only thing to add is what I've done in my PR. (_Add UPN and sAMAccountName as argument for the bind request_)


- For the tests, I will try to do some code but I am very not a expert of lua...But i'll try.

littlechicks:
Hi @kikito ,

To answer your comments: 

- Yes and no. The changes are part of Active Directory specification.
In fact, in LDAP Specification (https://www.ietf.org/rfc/rfc4514.txt) only CN and UID are valid attribute for binding. Therefore, they can be used within a Distinguished Name.

But let's clear this point.

In the `access.lua` file we call the function `ldap.bindrequest` which takes a username and password as arguments.
https://github.com/Kong/kong/blob/0c71caac07eade4a567a1ef1e1992f07d5a58c8c/kong/plugins/ldap-auth/access.lua#L95

Now in the `ldap.lua` we have the function _M.bind_request let's focus on theses lines:
https://github.com/Kong/kong/blob/0c71caac07eade4a567a1ef1e1992f07d5a58c8c/kong/plugins/ldap-auth/ldap.lua#L67-L70

- The first line format 0x80 type code which a "simple bind" type with the password.

- On the second line we create the LDAPMessage with the protocol version (here "3" for LDAPv3) then the username and finally the block created on the first line.

- Finally we create de full message by adding the message "ID" with the bindRequest code which is 0, and format the whole message

So finally we have something like that for the payload:

payload
 

I have tested a simple code for binding a user on an Active Directory LDAP using _UserPrincipalName_ attribute (_username@domain_) and I've checked the payload in Wireshark, **without SSL**.

It's look like this:


capture


So, we see that it is exactly the same payload than the LDAP Specification and the KONG `ldap.lua`file.
Therefore, we can use the same code without changing anything to the bind request function.
The only thing to add is what I've done in my PR. (_Add UPN and sAMAccountName as argument for the bind request_)


- For the tests, I will try to do some code but I am very not a expert of lua...But i'll try.

littlechicks:
So........... @kikito 

I cannot include tests...because Kong Spec for ldap plugins use `ec2-54-172-82-117.compute-1.amazonaws.com`
And this LDAP Backend is an OpenLDAP, and not Active Directory.
So I cannot test an authentication with UserPrincipalName or SAMAccoutName which are specific to MS-Ad.

See bellow the error.

test_ldap

`The successful binding is when I use UID as attribute for binding`
`The fail one is with use of user@domain` 


What can I do then  ?

littlechicks:
.

abentabib:
.

littlechicks:
So, voilà, the tests failed because we need a active directory server for testing...
Maybe, I can create a Active Directory in Azure Free Trial and test it again ?

darrenjennings:
@kikito instead of patching existing plugin to pluck part of the DN from base_dn, we should instead continue work from https://github.com/Kong/kong/pull/3093 which we are using successfully in the advanced version fo this plugin, which adds read only `config.bind_dn` user who can perform search and use the found user's DN to perform bind.

littlechicks:
@darrenjennings 

Sorry to go back on your comment, but I do not think it's a good solution to recover a password from an AD or any other LDAP Directory.
This is a very bad method. The userpassword attribute should never be retrieved. Otherwise you lose the whole principle of securing a password.

What's more, in active directory it is not possible to retrieve a password attribute without the fUserPwdSupport attribute being FALSE. So it depends on infrastructure.

Similarly in Active Directory a password is encrypted with UniquePwd.
So the pass is hashed, not reversible and it is salted.
And the work done in the PR you describe remains very specific to an OpenLDAP.

To be cross platform, it is better to try and bind with the values ​​provided and either succeed or fail.

In conclusion, ldap plugins kong depends extremly on a DN for binding, so what about an user who is not in the same OU ? It cannot work....

littlechicks:
**Updates**

**First point**
I've installed an Active Directory using Microsoft Azure Free Trial for my tests.
So you cannot not merges my tests before having a working Active Directory with SSL...

**Second Point**
All my tests pass. We can conclude that my code is working with `userPrincipalName`and `CN `(I have to add test for `sAMAccoutName `but it is the same in my case)

**Warning**

- We must use Proxy-Authorization in order to have a successfull authentication. (I don't know why)

- Credentials are passed to the header in the same format `ldap base64(username:password)`
   where username is the real `sAMAccoutName `or the first part of UPN before the `"@"`.

darrenjennings:
> Sorry to go back on your comment, but I do not think it's a good solution to recover a password from an AD or any other LDAP Directory.
> This is a very bad method. The userpassword attribute should never be retrieved. Otherwise you lose the whole principle of securing a password.

Yes the work we ended up using for enterprise did not return password, it uses more traditional method to avoid security issues you mention.

littlechicks:
@darrenjennings 

> > Sorry to go back on your comment, but I do not think it's a good solution to recover a password from an AD or any other LDAP Directory.
> > This is a very bad method. The userpassword attribute should never be retrieved. Otherwise you lose the whole principle of securing a password.
> 
> Yes the work we ended up using for enterprise did not return password, it uses more traditional method to avoid security issues you mention.

Thank you for this info. I am not able to get the code used in advanced plugins so your proposition is to use the PR mentionned as a work base ?
Indeed, having a bind dn who search the DN from an attribute can be useful :)
My solution here is more a workarround for people who want their Active Directory be compatible to the actual plugins, as it easy much simplier than the advanded plugin alternative. What do you think about merging first this workarround and then make a solution based on search function ?

omerl13:
Can you merge this PR please? I need to use the sAMAccountName for authentication (+the OU behavior fix), and right now I can't use Kong due to this issue.

nhgiang196:
I am waiting for this merge/solution so bad. LDAP Auths with sAMAccountName attribute are necessary for my gateway api project

sosagadinet (commented):
Hi, there is a typo in:

-- CASE: Attribute is sAMAccountName
if conf.attribute == "sAMAccoutName" then

I merge this code localy and make test against a MS-AD and works.

Thanks for your work.

mcen1:
It still looks like this is an issue, is the code @kikito created going to get merged? Was anyone waiting on this person for anything? Looks like all questions were answered and then the bot marked this as stale. The issue still persists at least in the version of Kong I'm using as of 6 months ago (2.8.1)

hanshuebner:
KAG-1047

mcen1:
I can't do much but suggest this fix is still relevant

hbagdi:

Pandede:
Still waiting for the fix...

PR #5171, opened 2019-10-29
feat(api) expose arbitrary plugin data from a schema-meta.lua

javierguerragiraldez:
### Summary

Adds a `/plugins/meta/:name` endpoint that returns a JSON representation of the value returned by a (optional) schema-meta.lua module in the plugin directory.

Any case of wrong plugin name, missing schema-meta or bad Lua syntax results in a 404 response.  If no name is given, returns a 400 response

### Issues resolved

PoC for Trello card: https://trello.com/c/ut8gMx93

### DIFF
--- a/autodoc/data/admin-api.lua
+++ b/autodoc/data/admin-api.lua
@@ -16,6 +16,7 @@ return {
       "kong/api/routes/health.lua",
       "kong/api/routes/config.lua",
       "kong/api/routes/tags.lua",
+      "kong/api/routes/metadata.lua",
     },
     nodoc_files = {
       "kong/api/routes/cache.lua", -- FIXME should we document this?
@@ -387,6 +388,33 @@ return {
             }
             ```
           ]]
+        },
+      },
+    },
+    metadata = {
+      title = [[ Metadata ]],
+      description = [[ Developer-supplied data. ]],
+      ["/metadata/plugins/:name"] = {
+        GET = {
+          title = [[ A plugin's metadata. ]],
+          endpoint = [[<div class="endpoint get">/metadata/plugins/:name</div>]],
+          description = [[
+            Returns a JSON representation of the value returned by the `metadata.lua` module
+            in the named plugin directory.
+          ]],
+          response = [[
+            ```
+            HTTP 200 OK
+            ```
+
+            ```json
+            {
+              "name": "sample plugin",
+              "description": "Just a sample of what can be.",
+              "icon": "http://example.com/plugin-icon.png"
+            }
+            ```
+          ]],
         },
       },
     },
@@ -819,23 +847,6 @@ return {
                 ]
             }
             ```
-          ]]
-        }
-      },
-      ["/plugins/:name/metadata"] = {
-        GET = {
-          title = [[Retrieve plugin's custom metadata]],
-          description = [[Retrieve a JSON representation of Lua data in a `schema-meta.lua` module]],
-          endpoint = [[<div class="endpoint get">/plugins/meta/{plugin name}</div>]],
-          response = [[
-            ```
-            HTTP 200 OK
-            ```
-            ```json
-            {
-              "description": "This is what the plugin developer intends",
-              "icon": <url of a nice pic>
-            }
           ]]
         }
       },
--- a/autodoc/data/admin-api.lua
+++ b/autodoc/data/admin-api.lua
@@ -822,7 +822,7 @@ return {
           ]]
         }
       },
-      ["/plugins/meta/:name"] = {
+      ["/plugins/:name/metadata"] = {
         GET = {
           title = [[Retrieve plugin's custom metadata]],
           description = [[Retrieve a JSON representation of Lua data in a `schema-meta.lua` module]],
--- a/autodoc/data/admin-api.lua
+++ b/autodoc/data/admin-api.lua
@@ -822,6 +822,23 @@ return {
           ]]
         }
       },
+      ["/plugins/meta/:name"] = {
+        GET = {
+          title = [[Retrieve plugin's custom metadata]],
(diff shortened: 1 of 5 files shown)


kong/api/routes/plugins.lua
```diff
@@ -186,6 +186,24 @@ return {
     end
   },
 
+  ["/plugins/meta/:name"] = {
```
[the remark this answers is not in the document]
javierguerragiraldez: what about `/plugins/:name/metadata` ?  it's really not about schema at all, after all

kong/api/routes/plugins.lua
```diff
@@ -186,6 +186,24 @@ return {
...
+    GET = function(self, db, helpers, parent)
+      ngx.log(ngx.DEBUG, "self.params: ", cjson.encode(self.params))
+      local name = self.params.name
+      if not name then
+        return kong.response.exit(400, { message = "Bad query" })
+      end
+
+      -- TODO: either unload module or use loadfile() instead
+      local ok, meta = utils.load_module_if_exists("kong.plugins." .. name .. ".schema-meta")
```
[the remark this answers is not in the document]
bungle: Do we plan to include metadata of `DAO`s too?

kong/api/routes/plugins.lua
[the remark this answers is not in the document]
hbagdi: I think we should move away from `/plugins` endpoint.

Do we plan on including meta for custom DAOs as @bungle is asking. And is it possible that we have these meta for core entities too some day?
I think these extensions could be useful in future to further add in more meta for each entity.
I think we should this under and `/meta` collection of endpoints as `/meta/plugins/:name`, which keeps the door open for core entities in future, such as `/meta/service`.

kong/api/routes/plugins.lua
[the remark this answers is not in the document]
javierguerragiraldez: As I understand the requirement, it's about user-visible info to describe the plugin itself, description, type, icon.  I don't think DAOs are visible enough to need such annotations.

(Of course, once it's in place, somebody will find new uses for it)

autodoc/data/admin-api.lua
```diff
@@ -390,6 +391,33 @@ return {
         },
       },
     },
+    metadata = {
+      title = [[ Metadata ]],
+      description = [[ Developer-supplied data. ]],
+      ["/metadata/plugins/:name"] = {
+        GET = {
+          title = [[ A plugin's metadata. ]],
```
hishamhm: the title field usually starts with an action verb, with no period:
```suggestion
          title = [[Retrieve the metatada for a plugin ]],
```

autodoc/data/admin-api.lua
```diff
@@ -390,6 +391,33 @@ return {
         },
       },
     },
+    metadata = {
+      title = [[ Metadata ]],
+      description = [[ Developer-supplied data. ]],
```
hishamhm: This probably needs at least a paragraph-long, explaining that this is arbitrary free-form data, etc.

kong/api/routes/metadata.lua
```diff
@@ -0,0 +1,25 @@
...
+return {
+  ["/metadata/plugins/:name"] = {
+    GET = function(self, db, helpers, parent)
+      ngx.log(ngx.DEBUG, "self.params: ", cjson.encode(self.params))
+      local name = self.params.name
+      if not name then
+        return kong.response.exit(400, { message = "Bad query" })
+      end
+
+      -- TODO: either unload module or use loadfile() instead
```
hishamhm: `load_module_if_exists` wraps `require` and is the way used for loading plugin modules elsewhere, so it's probably best to keep using it and unload the module from `package.loaded`.

autodoc/data/admin-api.lua
hbagdi: There is an implicit assumption that almost everything on the admin api is part of the compatibility promise.

If this is going to be free form data, it would help us in future if we clearly document the compatibility promise here for Open source.

javierguerragiraldez:
@Kong/team-interfaces is this what you need?

kinman:
@javierguerragiraldez sorry, this fell off our radar a bit. Will get back to you by end of this week.

autodoc/data/admin-api.lua
kinman: I know I asked for this feature, but I get cold feet when we save user-generated, user-visible, free-form data, especially without a plan for localization. Plugin category (Authentication, Logging, etc) and possibly icon URL seem less problematic, but this begs the question whether all of this should go in an external registry as has been suggested elsewhere.

autodoc/data/admin-api.lua
```diff
@@ -390,6 +391,33 @@ return {
...
+      ["/metadata/plugins/:name"] = {
```
kinman: I was discussing this with @darrenjennings , who makes the good point that while this is RESTful and consistent with the rest of the Admin API, it also requires the frontend to make upwards of 40 API calls just to render a list of plugins like https://manager-steady.kong-cloud.com/default/plugins/select/. That's not really practical, but the alternatives don't fit well into the Admin API as it exists today.

javierguerragiraldez: a few alternatives off the top of my head:
- gather the metadata of _all_ installed plugins on a single call (just `/metadata/plugins/`, without a name?)
- let the client ask for specific metadata fields (`/metadata/plugins/<name>/?f[]=icon&f[]=author&f[]=description&f[]=licence`. optional plugin name, as on previous point).  after all, only known fields can be rendered on the UI, right?
- use a registry.  not only can be managed, but could be GraphQL if you wish!

personally, I still think a registry is the best solution, but i'm aware that the decision hinges mostly on what's the desired process for external plugin writers.

PR #5178, opened 2019-10-30
hotfix(api) return endpoints inside a property

hbagdi:
Every list endpoint in Kong returns back data inside a `data` property
in the JSON response.

This small change follows the (informal) API convention.
This also makes it easy to parse the response in typed languages where
one wants to provide an "Object" to decode the response into.


Tieske:
good catch, thx @hbagdi !

PR #5186, opened 2019-10-31
fix(admin-api) change detection of foreign entity endpoints to explicit

bungle:
### Summary

The original detection worked like this:

```
local inverse = not foreign_schema.fields[foreign_field_name] and
                            schema.fields[foreign_field_name]
```

This worked great, but it held an assumption that can go wrong.

Example:

Routes DAO has foreign key to servive, specified by the field with name `service`.

So `routes_schema.fields["service"]` is specified (see schema.fields[foreign_field_name]).

All good so far. But then there is another check that I would say is flaky: `not service_schema.fields["service"]`. Service schema does not obviously have field called `service` currently, but if we ever add a field to service schema with a name of `service` this assumption turns up-side-down, and we have introduced a bug.

I though if we could make this more clever, but I ended up with an explicit function argument that specifies wheter the autogenerator should generate the foreign entity endpoint or entity endpoint.

foreign entity endpoint:
- `/routes/:routes/service`

entity endpoints:
- `/routes/:routes`
- `/services/:services/routes/:routes`

This PR fixes this.

### DIFF
--- a/kong/api/endpoints.lua
+++ b/kong/api/endpoints.lua
@@ -396,7 +396,7 @@ end
 --
 -- /services/:services
 -- /services/:services/routes/:routes
-local function get_entity_endpoint(schema, foreign_schema, foreign_field_name, method)
+local function get_entity_endpoint(schema, foreign_schema, foreign_field_name, method, is_foreign_entity_endpoint)
   return function(self, db, helpers)
     local entity, _, err_t
     if foreign_schema then
@@ -414,11 +414,8 @@ local function get_entity_endpoint(schema, foreign_schema, foreign_field_name, m
     end
 
     if foreign_schema then
-      local inverse = not foreign_schema.fields[foreign_field_name] and
-                                  schema.fields[foreign_field_name]
-
       local pk
-      if inverse then
+      if is_foreign_entity_endpoint then
         pk = entity[foreign_field_name]
         if not pk or pk == null then
           return not_found()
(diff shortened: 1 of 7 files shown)


kong/api/endpoints.lua
```diff
@@ -396,7 +396,7 @@ end
 --
 -- /services/:services
 -- /services/:services/routes/:routes
-local function get_entity_endpoint(schema, foreign_schema, foreign_field_name, method)
+local function get_entity_endpoint(schema, foreign_schema, foreign_field_name, method, inversed)
```
hishamhm: `inversed` is not a standard English word — it should be either `inverted` or `reversed`

bungle: i renamed the variable to `is_foreign_entity_endpoint`, doesn't need further explanation in my opinion.

bungle:
> I think the code is good enough, but this explanation is needed somewhere in the code itself:
> 
> > inversed endpoint:
> > `/routes/:routes/service`
> > normal endpoints:
> > `/routes/:routes`
> > `/services/:services/routes/:routes`
> 
> In general, a description of what every parameter is would help a lot on this file. I am not advocating to include it on every function. Perhaps once at the top of the file?

i renamed the variable to is_foreign_entity_endpoint, doesn't need further explanation in my opinion.

dndx:
Any way this change can be tested?

bungle:
> Any way this change can be tested?

it needs a custom plugin that creates entities that trigger it. So complex test for such a small change, but I can try to add it on monday.

bungle:
@dndx tests added.

PR #5201, opened 2019-11-04
feat(oauth2) add client_id header when user is authenticated

lucasmoreno:
When the user is successfully authenticated, the header X-Credential-Client-ID with the client_id of the credential that has this user will be sent to the upstream service.

### Summary

I was facing a scenario where I would like to make sure from which application one specific user was authenticated.

Right now, when the user is authenticated via OAuth2, a header is sent to identify who is the Consumer that "owns" that user (`X-Consumer-ID`).

The issue is: if a consumer has two applications and one user is registered in both, when the user authentication succeeds, we can't say from which application the user was accepted.

The table that stores the relationship between client and consumer is called credentials, so I decided to add the prefix Credential to the header's name.


kikito:
Closed by #5516

PR #5202, opened 2019-11-04
fix(request-termination) add check for content object

artomal:
### Summary

When you apply plugin for any `route` or `service` and set response code `204` without providing a message and you try to send a request to that route it responds with status code `204` and `Content-Length: 2` which is not good and tools such as `Postman` hangs on it.

This PR propose a solution to fix this issue by checking if `conf.message` is provided if not then just don't `exit` with empty body object just set `nil` to avoid any possible issues.

### Full changelog

* add additional check for `conf.message` before sending response back with `body` object included.

### DIFF
--- a/kong/plugins/request-termination/handler.lua
+++ b/kong/plugins/request-termination/handler.lua
@@ -40,7 +40,8 @@ function RequestTerminationHandler:access(conf)
     return kong.response.exit(status, content, headers)
   end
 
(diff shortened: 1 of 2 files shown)


murillopaula:
Looks like we accidentally made this regression when converting the plugin to the PDK as this issue is not present on 0.13.1: https://github.com/Kong/kong/commit/03d2e98a7f8116424be8f037985a72b9d18445b4

PR #5204, opened 2019-11-05
fix(core) remove Service Mesh from core

dndx:
Requirements:

* This PR requires that https://github.com/Kong/openresty-patches/pull/59
  is merged first (the `OPENRESTY_PATCHES_BRANCH` isn't set on
  `travis.yml`)

High level description of the changes included:

* Removal of the `service_mesh` config option and its associated
  functionalities in the runloop.
* Removal of the CA entity and associated functionalities (auto-creation
  and discovery)
* Removal of the `run_on` property from built-in and external plugins
  This requires bump versions on 4 plugins, as well as a migration.
* Removal of some openresty patches which where required for service
  mesh.
* Removal or deactivation of automated tests which either exercised=
  service mesh directly, or required service mesh in its current form in
  order to be run.
* The PostgreSQL strategy was using luasockets instead of cosockets
  on the `preread` phase because of a bug which no longer manifests
  without service mesh. So this PR activates cosockets on the preread
  phase.
* Related: https://github.com/thibaultcha/lua-cassandra#136
  replicates the change mentioned above, but for the Cassandra strategy.


spec/02-integration/05-proxy/22-reports_spec.lua
```diff
@@ -402,39 +401,5 @@ for _, strategy in helpers.each_strategy() do
       assert.match("tcp_streams=1", reports_data[1])
       assert.match("tls_streams=0", reports_data[1])
     end)
-
-    it("#stream reports tls streams", function()
```
hishamhm: This doesn't seem service mesh related. Can we still report on streams?

dndx: I think TLS stream uses the `starttls` feature which was buggy and removed with the patches.

kong/db/migrations/core/007_140_to_200.lua
```diff
@@ -0,0 +1,30 @@
+return {
+  postgres = {
+    up = [[
```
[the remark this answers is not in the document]
dndx: Good catch, updated.

spec/fixtures/custom_plugins/kong/plugins/short-circuit-last/schema.lua
```diff
@@ -4,9 +4,6 @@ local typedefs = require "kong.db.schema.typedefs"
 return {
   name = "short-circuit-last",
   fields = {
```
hishamhm: @dndx ah, I think those fixtures for "short-circuit" and "short-circuit-last" exist only for the plugins-triggering tests that were removed by this PR, and can be removed as well.

spec/02-integration/05-proxy/22-reports_spec.lua
kikito: In that case, shouldn't this be a `pending` instead of a delete (like the TLS-related test in `spec/02-integration/05-proxy/20-serviceless_proxying_spec.lua`)?

spec/fixtures/custom_plugins/kong/plugins/short-circuit-last/schema.lua
dndx: Hmm, I think those plugins are still used for `init_worker` tests:

https://github.com/Kong/kong/blob/8ee57d885756322bc89e31fa997132d74491beee/spec/02-integration/05-proxy/04-plugins_triggering_spec.lua#L2388

Can we safely remove those tests too?

hishamhm: ah, didn't notice `short-circuit` was doing double-duty for that as well! at the very least `short-circuit-last` could go away, I guess?

> Can we safely remove those tests too?

No, I think these tests still apply.

spec/02-integration/05-proxy/22-reports_spec.lua
bungle: Isn't this standard `nginx`:
https://github.com/Kong/kong/blob/master/kong/reports.lua#L220
http://nginx.org/en/docs/stream/ngx_stream_ssl_preread_module.html#variables

hishamhm: agree with @kikito there, at the very least make this a `pending`

kong/runloop/handler.lua
```diff
@@ -992,7 +952,7 @@ return {
         -----------------------------------------------------------------------
         -- Serviceless stream route
         -----------------------------------------------------------------------
-        local service_scheme = ssl_termination_ctx and "tls" or "tcp"
+        local service_scheme = "tcp"
```
bungle: I am not sure about this. Couldn't the scheme still be `tls`?

kong/db/migrations/core/007_140_to_200.lua
```diff
@@ -0,0 +1,38 @@
+return {
+  postgres = {
+    up = [[]],
+
+    teardown = function(connector)
+      assert(connector:query([[
+        DO $$
+        BEGIN
+          ALTER TABLE IF EXISTS ONLY "plugins" DROP COLUMN "run_on";
```
bungle: Do we also need to `DROP` indexes just to be explicit?

bungle:
Shouldn't these go away too:
https://github.com/Kong/kong/blob/fix/remove_mesh/kong/init.lua#L110
https://github.com/Kong/kong/blob/fix/remove_mesh/kong/init.lua#L778-L806
https://github.com/Kong/kong/blob/fix/remove_mesh/kong/runloop/handler.lua#L746

bungle:
This needs to go too:
https://github.com/Kong/kong/blob/fix/remove_mesh/autodoc/data/admin-api.lua#L889-L903

bungle:
Please also remove `preread` from here:
https://github.com/Kong/kong/blob/fix/remove_mesh/kong/db/strategies/postgres/connector.lua#L183-L186

bungle:
Additional note I think `preread` should be added here:
https://github.com/thibaultcha/lua-cassandra/blob/master/lib/cassandra/socket.lua#L126-L133

kong/db/migrations/core/007_140_to_200.lua
bungle: Ok, I am fine with this, even though with other migrations we have been explicit. Though I guess there are already multiple styles.

kong/runloop/handler.lua
kikito: I am also not sure about this one. pinging @dndx

kong-1.4.0-0.rockspec
```diff
@@ -37,14 +38,14 @@ dependencies = {
   "lua-resty-cookie == 0.1.0",
   "lua-resty-mlcache == 2.4.0",
   -- external Kong plugins
-  "kong-plugin-azure-functions ~> 0.4",
-  "kong-plugin-zipkin ~> 0.1",
+  "dndx/kong-plugin-azure-functions dev",
+  "dndx/kong-plugin-zipkin dev",
```
hishamhm: rockspec needs to be updated to the released versions

kong/runloop/handler.lua
dndx: TLS is not supported until the `stream_listen` `ssl` flag thing is added.

kong-1.4.0-0.rockspec
dndx: Reverted. Thanks!

kikito: I believe only reverting was not good enough. We also need to bump the versions of these plugins in Kong (so that they don't use the `run_on` field which this PR removes). I will do that.

spec/02-integration/05-proxy/22-reports_spec.lua
kikito: I added the spec back as a `pending`

PR #5206, opened 2019-11-06
feat(balancer) use lua-resty-dns-client healthThreshold attrib

locao:
- added new field `threshold` to `healthchecks` in `upstreams` entities.
- added new query parameter `balancer_health` for the endpoint `/upstreams/:upstream_id/health`. When `balancer_health=1`, the endpoint returns the balancer health, instead of the targets' health.

### DIFF
--- a/kong/api/routes/upstreams.lua
+++ b/kong/api/routes/upstreams.lua
@@ -59,6 +59,21 @@ return {
         return kong.response.exit(404, { message = "Not found" })
       end
 
+      local node_id, err = kong.node.get_id()
+      if err then
+        kong.log.err("failed to get node id: ", err)
+      end
+
+      if tostring(self.params.balancer_health) == "1" then
+        local upstream_pk = db.upstreams.schema:extract_pk_values(upstream)
+        local balancer_health  = db.targets:get_balancer_health(upstream_pk)
+        return kong.response.exit(200, {
+          data = balancer_health,
+          next = null,
+          node_id = node_id,
+        })
+      end
+
       self.params.targets = db.upstreams.schema:extract_pk_values(upstream)
       local targets_with_health, _, err_t, offset =
         endpoints.page_collection(self, db, db.targets.schema, "page_for_upstream_with_health")
@@ -70,11 +85,6 @@ return {
       local next_page = offset and fmt("/upstreams/%s/health?offset=%s",
                                        self.params.upstreams,
                                        escape_uri(offset)) or null
-
-      local node_id, err = kong.node.get_id()
-      if err then
-        kong.log.err("failed getting node id: ", err)
-      end
 
       return kong.response.exit(200, {
         data    = targets_with_health,
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -415,10 +415,13 @@ do
     end
 
     creating[upstream.id] = true
+    local health_threshold = upstream.healthchecks and
+                              upstream.healthchecks.threshold or nil
 
     local balancer, err = balancer_types[upstream.algorithm].new({
       wheelSize = upstream.slots,  -- will be ignored by least-connections
       dns = dns_client,
+      healthThreshold = health_threshold,
     })
 
     if not balancer then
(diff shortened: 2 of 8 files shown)


kong/runloop/balancer.lua
```diff
@@ -1036,6 +1048,43 @@ local function get_upstream_health(upstream_id)
 end
 
 
+--------------------------------------------------------------------------------
+-- Get healthcheck information for a balancer.
+-- @param upstream_id the id of the upstream.
+-- @return table with balancer health info
+local function get_balancer_health(upstream_id)
+
+  local balancer_health = {
```
Tieske: since this local is not used, you can just return it at the very end at once, eg:

```lua
return {
  health = ...,
  id = upstream_id,
}
```

Tieske:
instead of the query argument, cannot we add the balancer health in the top elelements of the response as shown here: https://docs.konghq.com/1.4.x/admin-api/#show-upstream-health-for-node

```json
{
    "total": 2,
    "node_id": "cbb297c0-14a9-46bc-ad91-1d0ef9b42df9",
    "status": "HEALTHY",                                        --------> this one added
    "data": [
        {
            "created_at": 1485524883980,
            "id": "18c0ad90-f942-4098-88db-bbee3e43b27f",
            "health": "HEALTHY",
            "target": "127.0.0.1:20000",
            "upstream_id": "07131005-ba30-4204-a29f-0927d53257b4",
            "weight": 100
        },
        {
            "created_at": 1485524914883,
            "id": "6c6f34eb-e6c3-4c1f-ac58-4060e5bca890",
            "health": "UNHEALTHY",
            "target": "127.0.0.1:20002",
            "upstream_id": "07131005-ba30-4204-a29f-0927d53257b4",
            "weight": 200
        }
    ]
}
```

hishamhm:
> instead of the query argument, cannot we add the balancer health in the top elelements

@Tieske I discussed this with @locao and I suggested the query argument as a way to add the feature without incurring in a breaking change for the format (and subsequent adjustments to any tooling built around Kong's health info, etc.) and we agreed this was a good way to go.

The rationale was that the configuration and behavior of healthchecks are already pretty complicated as is, and given that the default for `health_threshold` is 0, most users shouldn't need to deal with three nested levels of `health` status variables (your JSON example above didn't include the `addresses` nested array)

hishamhm:
@locao `Missing autodoc data for field upstreams.healthchecks.threshold` ;)

kong/runloop/balancer.lua
locao: Done!

kong/runloop/balancer.lua
```diff
@@ -1036,6 +1048,43 @@ local function get_upstream_health(upstream_id)
...
+  local using_hc = is_upstream_using_healthcheck(upstream)
+
+  local balancer = balancers[upstream_id]
+  if not balancer then
+    return nil, "balancer not found"
+  end
+
+  local healthchecker
+  local health = "HEALTHCHECKS_OFF"
+  if using_hc then
```
kikito: it seems the `using_hc` variable is not needed. Replace this with `if is_upstream_using_healthcheck(upstream) then` and remove it from above.

kong/api/routes/upstreams.lua
```diff
@@ -59,6 +59,21 @@ return {
         return kong.response.exit(404, { message = "Not found" })
       end
 
+      local node_id, err = kong.node.get_id()
+      if err then
+        kong.log.err("failed getting node id: ", err)
```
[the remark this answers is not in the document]
locao: Done!

kong/runloop/balancer.lua
locao: Good catch, thanks. Done!

hishamhm:
@locao I noticed a small detail before hitting `merge`: the query-arg for the endpoint needs to be documented in the Admin API autodocs as well (grep for `request_query` in `autodoc/data/admin-api.lua` for example on documenting query args — or see https://github.com/Kong/kong/pull/5210 :grin: )

locao:
> @locao the query-arg for the endpoint needs to be documented in the Admin API autodocs as well

Good catch @hishamhm, thank you! Docs updated

PR #5219, opened 2019-11-13
fix(helpers) remove unnecessary plugin override from `helpers.start_kong`

onematchfox:
### Summary

This PR resolves a minor annoyance in the plugin testing process by removing the requirement to explicitly specify enabled plugins when calling `helpers.start_kong`.

Currently, when writing plugin tests, one first needs to ensure that all plugins are enabled via `KONG_TEST_PLUGINS` or DB/blueprint methods such as `bp.plugins:insert({ ... name = "my-plugin", config = { ... } })` called prior to starting Kong will result in the error `schema violation (name: plugin 'my-plugin' not enabled; add it to the 'plugins' configuration property)`. One then has to re-specify all enabled plugins when calling `helpers.start_kong` - i.e. `helpers.start_kong( { plugins = "bundled,my-plugin" })` or the code at https://github.com/Kong/kong/blob/0ff3dea2dec36085ddf90bcf3dc87de2303cc49d/spec/helpers.lua#L1512 will effectively disable any plugins you have already enabled by overriding the `env.plugins` with the built-in set of plugins. This results in an error like `nginx: [error] ... my-plugin plugin is in use but not enabled.`. This override ultimately goes against the parameter description on the method where it states: "@param env (**optional**) table with kong parameters to set as environment variables, **overriding** the test config". 

By removing the forced override if `env.plugins` is not specified, the default test config is applied and one no longer needs to explictly set the enabled plugins in both places. Of course, one is still able to override the enabled plugins by setting `env.plugins`. Thus, I don't believe anything should break and, folks still have the option to test an explicit scenario where a different set/subset of plugins are enabled.

### Full changelog

* Remove plugin override when calling `helpers.start_kong`

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -1518,12 +1518,6 @@ local function kong_exec(cmd, env, pl_returns, env_vars)
 
   env.lua_package_path = env.lua_package_path .. ";" .. conf.lua_package_path
 
-  if not env.plugins then
-    env.plugins = "bundled,dummy,cache,rewriter,error-handler-log," ..
-                  "error-generator,error-generator-last," ..
-                  "short-circuit,short-circuit-last"
-  end
-
   -- build Kong environment variables
   env_vars = env_vars or ""
   for k, v in pairs(env) do


Tieske:
I've never set `KONG_TEST_PLUGINS` when running plugin tests?

Tieske:
You specify the plugin name when preparing the db; https://github.com/Kong/kong-plugin/blob/master/spec/myplugin/01-access_spec.lua#L29

And again when starting Kong; https://github.com/Kong/kong-plugin/blob/master/spec/myplugin/01-access_spec.lua#L63

This way the entire test setup is in the code (as it should) without having to manually specify stuff in your environment.

The `KONG_TEST_*` variables are just available as an override.

onematchfox:
> You specify the plugin name when preparing the db; https://github.com/Kong/kong-plugin/blob/master/spec/myplugin/01-access_spec.lua#L29

Ah, OK, I wasn't doing that so point taken. Will try that out. 

But, even given that, is there a reason to have the override that I have removed? Why not fall back to env vars if the user wants that? Why does the current code override an override so to say?

onematchfox:
@Tieske thanks for the tip on specifying plugin details - have confirmed that this works and so, ultimately that does remove my need for this change. I do like the idea of having the configuration of enabled plugins isolated to the code as you say. However, this does introduce some complexity to my tests in that I have several plugins that rely off common database tables. The creation and migration of these tables is done by a central plugin and the other plugins just access that table for their own needs. This means, that in the test suite I do need to enable multiple plugins to ensure everything is in place. This is of course possible possible and whilst for me personally, it feels more complex than just being able to enable everything in one location (env vars/conf file), I do like that this does highlight the inter-dependencies we have in the actual tests themselves. 

That said, I do however have a couple comments that I hope you will consider before I go ahead and close this PR. 

1. Given the build failure, it would seem that even the core Kong code base does not make use of the plugins parameter - i.e. isolating the test setup to the code as you say - but instead relies on the common set of plugins being set by `start_kong`. I'm happy to adjust these tests to fix the build but given that you have provided me with a solution/workaround I'm also happy to go my own way and don't want to waste anyone's time if this PR is not going to be considered any further.
2. You state "The `KONG_TEST_*` variables are just available as an override.". I would disagree here in that these variables are ultimately applied to the configuration prior to `start_kong` being called. Thus, the override is actually the `plugins` element passed through on that call. This is even confirmed in the method description as I mentioned when opening the PR - "@param env (optional) table with kong parameters to set as environment variables, overriding the test config". The code I have removed essentially forces the user to override the test config. This is not an optional override as the method description states that it should be; the code in question forces the user to provide an override, because if they don't it will override any default configuration for you - whether you like it or not. Even given the changes you have suggested I still do not understand why one would want to remove flexibility and force a user to have to override the test config they may have already provided (be that via env vars or `kong_tests.conf` file)? What if I actually wanted to my full test suite using the configuration I would use in a production scenario? Yes, this is no longer strictly unit testing but surely it is still a valid test?

Tieske:
> Given the build failure, it would seem that even the core Kong code base does not make use of the plugins parameter - i.e. isolating the test setup to the code as you say - but instead relies on the common set of plugins being set by start_kong. 

That is exactly what bothered me as well, when I reviewed this PR.

Issue #5136, opened 2019-10-17
Update Upgrade docs to list breaking changes for 0.14.x


ukiahsmith:
### Summary

The [changelog for Kong 0.14.x](https://github.com/Kong/kong/blob/master/CHANGELOG.md#admin-api-11) mentions breaking changes to the `/consumers` endpoint as a result of moving to the new DAO, as part of #3437

The specific breaking changes are not listed in the CHANGELOG or the [UPGRADE](https://github.com/Kong/kong/blob/master/UPGRADE.md#upgrade-to-014x) document.

### Steps To Fix

Please update the documentation to include the specific breaking changes.

vmr1532:
i am taking this issue

AsherThomasBabu:
Check this PR please,
I added the missing breaking changes as mentioned earlier 
https://github.com/Kong/kong/pull/5365#issue-355966730

kikito:
Closed by #5365

kikito:
BTW - nice to see you, @ukiahsmith :)!

PR #5228, opened 2019-11-16
Update License file

vmr1532:
Fixed Grammatical and spelling error

### Summary

* Removed some of the useless words like "the purposes of" in the sentence.
* Changed the word for "for the purpose of discussing" to "to discuss and improve" in the sentence for the simplicity of reading the file.


hishamhm:
@vmr1532 The Apache License is a standard license, we cannot change its wording.

Also, copying the first sentence from #5136 does not address the issue there, which is asking about the specific compatibility changes.

I'm afraid we can't take these changes as is, sorry.

PR #5231, opened 2019-11-18
refactor(admin-api) plugin endpoints

bungle:
### Summary

Refactor plugin endpoints and remove two files related to that. This also fixes small issues on `backward-compatible` labeled functions especially those related to validation. They are not only `backward-compatible` but `fully-compatible` as we added support for those endpoints later on.


bungle:
On such error that this PR fixes is:
```
http put :8001/services/test/plugins/5af6c08e-c7ff-4f01-8190-696d3ff3de37 name=acl config.whitelist=ads -f
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 153
Date: Mon, 18 Nov 2019 20:33:18 GMT
Server: kong/1.4.0
X-Kong-Admin-Latency: 1

{
    "code": 2,
    "fields": {
        "service": {
            "id": "expected a valid UUID"
        }
    },
    "message": "schema violation (service.id: expected a valid UUID)",
    "name": "schema violation"
}
```

There is no test for this added because this removes customization, so normal tests already test it correctly.

PR #5234, opened 2019-11-19
feat(core) buffered proxying

bungle:
### Summary

Implements buffered proxying in Kong so that plugins can implement:

```lua
function Plugin:access(conf)
  kong.service.request.enable_buffering()
end

function Plugin:header_filter(conf)
  local body = kong.service.response.get_raw_body()
  -- OR
  local body = kong.service.response.get_body()

  if not body then
    return
  end

  -- ...
end
```


kidd:
Style question: Being a new feature that mostly adds new isolated code, is there a reason why we don't put the added code in `kong.access` in a function on its own?

spec/fixtures/custom_plugins/kong/plugins/enable-buffering/handler.lua
```diff
@@ -0,0 +1,33 @@
...
+function EnableBuffering:access()
+  kong.service.request.enable_buffering()
+end
+
+
+function EnableBuffering:header_filter(conf)
+  if conf.mode == "modify-json" then
+    local body, err = kong.service.response.get_body()
+    ngx.log(ngx.CRIT, require "inspect"(err))
+    ngx.log(ngx.CRIT, require "inspect"(body))
```
hishamhm: leftover debugging code?

Issue #4480, opened 2019-04-05
ACL plugin: 401 vs 403


odelvalle:
### Summary
ACL plugin should return 401 when credentials is not provided. Now always returns 403

### Steps To Reproduce
1. Add ACL plugin with white list to any service
2. Add Any security plugin (ApiKey)
3. Request any service route without credentials  (response 403)
4. Disable ACL plugin and send the request again (response 401)

401 Unauthorized (I don't have credentials) vs 403 Forbidden (I have credentials but, is refusing to fulfil it

### Additional Details & Logs
- Kong version (`1.0.2`)

bungle:
@odelvalle I like your proposal. PR is welcomed! I will also add card to our backlog.

bungle:
Fixed with #5440.

PR #5236, opened 2019-11-20
feat(acl) reject anonymous consumer

jeremyjpj0916:
### Summary

We started looking into leveraging the anonymous user functionality in Kong thats native for helping use multiple auth patterns on a given service/route:
https://docs.konghq.com/0.14.x/auth/#multiple-authentication

But we wanted to use OAuth2 + JWT programmatic auth, alongside the ACL plugin to control the proper authorization of given authenticated consumers. This is nice because different consumers can call the same proxy with preferred authentication patterns.

Out of the box what happens when a user is either missing or a bad token when doing the anonymous user pattern with OAuth2 + JWT + ACL plugins is a HTTP Status 403: "You cannot consume this service" response. Which to a client is not technically correct from their perspective with a bad/missing/expired  tokens.

One idea we had would be as the docs state to throw on the request termination plugin to return the error to the anonymous user as the docs mention, but the ACL plugin runs before request termination so thats not possible in our situation unless Kong deems moving priorities of request-termination to come before ACL the appropriate fix and not this pr enhancement(at risk of breaking some other odd flow teams are using that involve both ACL + Request termination now and rely on current orders). Could be a potential breaking change for some folks too. 

Besides the ^, it also isn't ideal because to support OAuth + JWT + ACL is 3 plugins per proxy, and implementing it where the request-termination returns the error message before ACL means a 4th plugin too. Feels kinda odd that the request-termination plugin was where the Kong anonymous user logic was implemented the first time, doesn't it make more sense an ACL plugin would be a more appropriate plugin for user management(such as anonymous user functionality)? 

So ultimately what this enhancement does is says if I see the anonymous user present in the ctx of this transaction then return what a REST client would expect on a missing/expired token request. Whereas a 403 isn't necessarily appropriate to return to a client in the scenario I have described. It makes sense to me that to the calling client, anonymous user functionality should be as transparent as possible.

I had pondered an implementation where we keep the context of the original auth plugin errors even with the anonymous functionality allowing a continuation and then returning their errors but then I realized with multiple auth plugins running on the proxy you can't really know which auth pattern is the right error to return to the consumer so here we are 😀 .

Happy for discussion/feedback!

-Jeremy

### Full changelog

* Implement: reject_anonymous user functionality to return appropriate HTTP Status + Error.
* Testcase Added: /spec/03-plugins/18-acl/02-access_spec.lua

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -53,12 +53,21 @@ function ACLHandler:access(conf)
       type = config_type,
       groups = config_type == BLACK and conf.blacklist or conf.whitelist,
       cache = setmetatable({}, mt_cache),
+      reject_anonymous = conf.reject_anonymous
     }
 
     config_cache[conf] = config
   end
 
   local to_be_blocked
+
+  -- reject anonymous consumer
+  if config.reject_anonymous then
+    local credential = kong.client.get_credential()
+    if credential and credential.consumer_id == config.reject_anonymous then
+      return kong.response.exit(401, { message = "Unauthorized" })
+    end
+  end
 
   -- get the consumer/credentials
   local consumer_id = groups.get_current_consumer_id()
--- a/kong/plugins/acl/schema.lua
+++ b/kong/plugins/acl/schema.lua
@@ -13,6 +13,7 @@ return {
           { whitelist = { type = "array", elements = { type = "string" }, }, },
           { blacklist = { type = "array", elements = { type = "string" }, }, },
           { hide_groups_header = { type = "boolean", default = false }, },
+          { reject_anonymous = { type = "string", uuid = true }, },
         }
       }
     }
--- a/spec/03-plugins/18-acl/02-access_spec.lua
+++ b/spec/03-plugins/18-acl/02-access_spec.lua
@@ -678,6 +678,37 @@ for _, strategy in helpers.each_strategy() do
         }
       }
 
+      local route14 = bp.routes:insert {
+        hosts = { "acl14.com" },
+      }
+
+      bp.plugins:insert {
+        name = "acl",
+        route = { id = route14.id },
+        config = {
+          whitelist = { "anonymous" },
+          reject_anonymous = anonymous.id
+        }
+      }
+
+      bp.plugins:insert {
+        name = "key-auth",
+        route = { id = route14.id },
+        config = {
+          anonymous = anonymous.id,
+        }
+      }
+
+      bp.plugins:insert {
+        name = "ctx-checker",
+        route = { id = route14.id },
+        config = {
+          ctx_kind      = "kong.ctx.shared",
+          ctx_set_field = "authenticated_groups",
+          ctx_set_array = { "anonymous" },
+        }
+      }
+
       assert(helpers.start_kong({
         plugins    = "bundled, ctx-checker",
         database   = strategy,
@@ -1262,6 +1293,17 @@ for _, strategy in helpers.each_strategy() do
         local body = assert.res_status(403, res)
         local json = cjson.decode(body)
         assert.same({ message = "You cannot consume this service" }, json)
+      end)
+
+      it("rejects anonymous consumer when reject_anonymous configured", function()
+        local res = assert(proxy_client:get("/request", {
+          headers = {
+            ["Host"] = "acl14.com"
+          }
+        }))
+        local body = assert.res_status(401, res)
+        local json = cjson.decode(body)
+        assert.same({ message = "Unauthorized" }, json)
       end)
     end)
   end)


jeremyjpj0916:
Hmm looks like I did a goof, will see whats wrong w the code shortly:
```
[ RUN      ] spec/03-plugins/18-acl/02-access_spec.lua @ 1298: Plugin: ACL (access) [#postgres] Forbids with rejects anonymous consumer when reject_anonymous configured
spec/03-plugins/18-acl/02-access_spec.lua:1304: Invalid response status code.
Status expected:
(number) 401
Status received:
(number) 200
Body:
(string) '{"headers":{"host":"127.0.0.1:15555","x-consumer-username":"anonymous","x-forwarded-host":"acl14.com","connection":"keep-alive","x-authenticated-groups":"anonymous","x-forwarded-proto":"http","x-anonymous-consumer":"true","x-forwarded-for":"127.0.0.1","user-agent":"lua-resty-http\/0.14 (Lua) ngx_lua\/10015","x-consumer-custom-id":"consumer-id-5","x-consumer-id":"a82b2997-ad37-4878-b0b3-39abb80c36b5","x-real-ip":"127.0.0.1","x-forwarded-port":"9000"},"url":"http:\/\/127.0.0.1:15555\/request","vars":{"host":"127.0.0.1","request_method":"GET","request_uri":"\/request","request":"GET \/request HTTP\/1.1","remote_port":"","server_port":"15555","realip_remote_addr":"127.0.0.1","binary_remote_addr":"\u007f\u0000\u0000\u0001","hostname":"travis-job-0ca8f277-a92c-4f57-964e-c2cdd16e7078","server_protocol":"HTTP\/1.1","https":"","realip_remote_port":"35222","server_addr":"127.0.0.1","remote_addr":"127.0.0.1","server_name":"mock_upstream","uri":"\/request","ssl_server_name":"no SNI","request_length":"438","scheme":"http","is_args":"","request_time":"0.000"},"uri_args":{},"post_data":{"params":null,"text":"","kind":"unknown"}}'
401
stack traceback:
	spec/03-plugins/18-acl/02-access_spec.lua:1304: in function <spec/03-plugins/18-acl/02-access_spec.lua:1298>
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 1298: Plugin: ACL (access) [#postgres] Forbids with rejects anonymous consumer when reject_anonymous configured (2.76 ms)
```

bungle:
Hi @jeremyjpj0916,

Thank you for the PR. I thought about it a bit. And yes, it is another one that shows the age on our plugin runloop. As we are approaching `2.0` I am sure we don't have time for huge overhaul, but one thing we could perhaps do is to add `priority` field in `plugins` table perhaps with default of `handler.PRIORITY`. Then users could order the plugins as they like. I won't promise this, but it seems rather easy to do, and probably very backward compatible. I am not really a fan of how we do multi-auth plugins. It is strange, hard to understand and a bit hacky, but that is another story.

Now back to your PR.  I was thinking that could you do this without configuration parameter? The `kong.client.get_credential()` will return `nil` for `anonymous`, where `ACL` could perhaps `401` on that. Yes, this is breaking change. Another option could be that if there are no groups specified for `consumer` that could mean `401`, this changes the current behavior as well.

Then there is the `new` plugin option. E.g. a `consumer-termination` plugin where you could have `whitelist` or `blacklist` of consumers.

The ither thing in your PR is that you call:
```
local credential = kong.client.get_credential()
```

AFAIK, anonymous consumers don't have credential, thus they are anonymous, so that needs to be changed to `kong.client.get_consumer()` OR you should actually check that `credential` is `nil`, otherwise it is not just anonymous rejecting but any consumer rejecting. Now if we think it like that, perhaps we could have: `config.blacklist_consumers` and `config.whitelist_consumers` to make this feature more flexible (set of uuids)? But then we need to specify semantics, are these checked in addition, or do they make it skip group checks? And then we still have the `status code` issue (`403` or `401`). Feels like a thing that we should discuss more. What do the others think?

jeremyjpj0916:
@bungle I think you hit the nail on the head with this future change suggestion:

> we could perhaps do is to add priority field in plugins table perhaps with default of handler.PRIORITY. Then users could order the plugins as they like.

If that is in the cards for a future version maybe we drop even trying to add my enhancement to the ACL plugin, it still would require me having 4 plugins on a proxy vs 3 if ACL could support it but it is probably the most sensible imo, can get over the extra resource per proxy, just means a couple thousand hah.

Open to your other suggestions as well but will leave it to the group. One thing I think would be detrimental would be introducing something to the ACL now that would just be baggage later if a good refactor gives us the flexibility and functionality needed to solve why this PR was made in the first place(403 vs 401 to a client on anonymous user when ACL plugin present). 

Will continue to wait if any group feedback, if none I may just run a handler patch file drop in to handle the anonymous user with a 401 unauthorized like my current PR has on our side until Kong catches up with a pattern we can leverage in its place 👍 , appreciate your feedback as always Mr. Bungle!

jeremyjpj0916:
Also maybe there is a case to be made that a 403 is never appropriate to return to a user that is anonymous and just should be a breaking change default released on a newer Kong version w a 401 Unauthorized generic response. Also a PR I could make.

bungle:
There is also this:
https://github.com/Kong/kong/issues/4480

```
```

Could we just do that? E.g. if `kong.client.get_credential()` is empty we `401`. But things have changed since, only `401` if `ngx.authenticated_groups` are not specified. @jeremyjpj0916 what do you think? Then there is no need to specify a configuration parameter?

bungle:
@jeremyjpj0916 if you think that the above would work for you, could you change the PR, I think we are then pretty close of approving this.

jeremyjpj0916:
@bungle do you mean to do 

```
if !kong.client.get_credential() and !ngx.authenticated_groups then
  return kong.response.exit(401, { message = "Unauthorized" })
end
``` 
?

jeremyjpj0916:
Actually I know what to do. Faster for me to do a separate PR for the change in direction. Will link them accordingly to keep a history trail.

hishamhm:
@jeremyjpj0916: @bungle and I briefly discussed this one this morning and he said he was going to reply with "option 1" later today, if that's useful :)

jeremyjpj0916:
@hishamhm This line:

 https://github.com/Kong/kong/blob/master/kong/plugins/oauth2/access.lua#L822 gives me a bit of pause for option 1. Because it means the anonymous user must have kong.client.get_credential() non nil at first glance unless I am mistaken.  

This may be cleaner if the ^ holds true:

Replacing:
```
  if not consumer_id then
    local authenticated_groups = groups.get_authenticated_groups()
    if not authenticated_groups then
      kong.log.err("Cannot identify the consumer, add an authentication ",
                   "plugin to use the ACL plugin")

      return kong.response.exit(403, {
        message = "You cannot consume this service"
      })
    end
```

with

```
  if not consumer_id or !ngx.authenticated_groups then --This means no consumer found OR they have no groups to start with(is it a valid assumption that folks should make the anon user groupless???)
    local authenticated_groups = groups.get_authenticated_groups()
    if not authenticated_groups then
      kong.log.err("Cannot identify the consumer, add an authentication ",
                   "plugin to use the ACL plugin")

      return kong.response.exit(401, {
        message = "Unauthorized"
      })
    end
```

bungle:
@jeremyjpj0916 those lines are actually otherway around, they mean:

1. is plugin configured for anonymous access AND
2. another plugin has slready authenticated the user (thus the credential being there)


anonymous consumers are anonymous because they have not authenticated by any credential, read: no credential.

bungle:
the groups check is there as we have also autj plugins that do not not use consumers or have that as optional (usually through mapping). they might then only set credential. they can also set ngx.ctx.authenticated_groups without again consumer and the groups applied to consumer, but e.g. groups filled directly from jwt claims or ldap groups.

jeremyjpj0916:
Thanks for the clarification, I added option 1 into the codebase and it worked as desired I think:

image


Also sent with correct access token and got through correctly so this is great :D .

PR #5244, opened 2019-11-21
chore(ci) build nightly releases from jenkins instead of travis-ci (#…

hutchic:
Moves the kong nightly builds off of travis-ci onto a internal private jenkins instance.

This has been running on a Kong fork for about a week successfully and jenkins is configured in such a way that it will not report failures back to github

Capture

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,6 @@ cache:
 stages:
   - lint and unit
   - test
-  - Deploy daily build
   - Release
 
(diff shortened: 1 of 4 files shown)


hishamhm:
@hutchic I split it into two commits

hutchic:
@hishamhm yeah it occurred to me that I squashed too far but the forgot to tease them apart thanks for that 👍

PR #5247, opened 2019-11-22
fix(cluster-events) make cluster events more robust and more consistent

bungle:
### Summary

The current implementation of cluster events (before this PR) uses `ngx.now()` a lot when inserting and reading data from `cluster_events`. This can lead to various hard to debug issues as there can be a clock skew between the nodes.

Issues like:

1. writer being in future making the readers not to pick events
2. writer being in past making the readers skip events
3. cleanup routine may delete entries the readers have not applied

This PR changes the cluster events to use `cassandra` or `postgres` server time on `inserting` and on `selects`. There is some fall-back code to `ngx.now()` but they should almost never be used (only in cases where there is a temporary communication issue with db).

This also makes nodes, even with clock skew, more consistent.

Unfortunately it is hard to write test cases for this, but the changes do pass the current test suite.

There is one slight change in interface of `cluter_events:broadcast` to be noted:

Previous:
```
function _M:broadcast(channel, data, nbf)
```

New:
```
function _M:broadcast(channel, data, delay)
```

Where `nbf` is exact time, but `delay` is relative. This was AFAIK only used internally.

This issue was reported internally on an environment that did not implement proper clock syncs across the cluster.

### DIFF
--- a/kong/cluster_events/init.lua
+++ b/kong/cluster_events/init.lua
@@ -4,6 +4,7 @@ local ERR       = ngx.ERR
 local CRIT      = ngx.CRIT
 local max       = math.max
 local type      = type
+local error     = error
 local pcall     = pcall
 local insert    = table.insert
 local ngx_log   = ngx.log
@@ -31,12 +32,12 @@ local function log(lvl, ...)
 end
 
 
-local function nbf_cb_handler(premature, cb, row)
+local function nbf_cb_handler(premature, cb, data)
   if premature then
     return
   end
 
-  cb(row.data)
+  cb(data)
 end
 
 
@@ -117,7 +118,8 @@ function _M.new(opts)
 
   -- set current time (at)
 
-  local ok, err = self.shm:safe_set(CURRENT_AT_KEY, ngx_now())
+  local now = strategy:server_time() or ngx_now()
+  local ok, err = self.shm:safe_set(CURRENT_AT_KEY, now)
   if not ok then
     return nil, "failed to set 'at' in shm: " .. err
   end
@@ -139,7 +141,7 @@ function _M.new(opts)
 end
 
 
-function _M:broadcast(channel, data, nbf)
+function _M:broadcast(channel, data, delay)
   if type(channel) ~= "string" then
     return nil, "channel must be a string"
   end
@@ -148,16 +150,16 @@ function _M:broadcast(channel, data, nbf)
     return nil, "data must be a string"
   end
 
-  if nbf and type(nbf) ~= "number" then
-    return nil, "nbf must be a number"
+  if delay and type(delay) ~= "number" then
+    return nil, "delay must be a number"
   end
 
   -- insert event row
 
   --log(DEBUG, "broadcasting on channel: '", channel, "' data: ", data,
-  --           " with nbf: ", nbf and nbf or "none")
+  --           " with delay: ", delay and delay or "none")
 
-  local ok, err = self.strategy:insert(self.node_id, channel, ngx_now(), data, nbf)
+  local ok, err = self.strategy:insert(self.node_id, channel, nil, data, delay)
   if not ok then
     return nil, err
   end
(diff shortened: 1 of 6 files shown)


kong/cluster_events/init.lua
```diff
@@ -219,67 +277,27 @@ local function poll(self)
...
-                  log(ERR, "failed to schedule nbf event timer: ", err)
-                end
-              end
-            end
-          end
-        end
+    for i = 1, count do
+      local ok, err = process_event(self, rows[i])
+      if not ok then
+        return nil, err
```
bungle: I have one question to @thibaultcha about this. This didn't change (got only refactored), but I am a bit worried that could this lead to missing events because on line `291` we set `CURRENT_AT_KEY` to current time, which gets next time used as `min_at`. Should be rather log the error, or is the current ok?

kong/cluster_events/init.lua
```diff
@@ -219,67 +277,27 @@ local function poll(self)
...
+  for rows, err, page in self.strategy:select_interval(self.channels, min_at) do
     if err then
       return nil, "failed to retrieve events from DB: " .. err
     end
 
+    local count = #rows
+
     if page == 1 then
-      local ok, err = self.shm:safe_set(CURRENT_AT_KEY, max_at)
+      local at = count > 0 and rows[1].now or self.strategy:server_time() or ngx_now()
```
hishamhm: I don't think we should ever get `self.strategy:server_time()` here. Assuming that we used it, this timestamp is potentially greater than that of the last event returned, meaning that any events in the time window between `select_interval` and `server_time` would not be processed. That time window is usually tiny, but DB operations may yield, so an arbitrary amount of time can pass between lines 282 and 290.

I think `CURRENT_AT_KEY` should be set to the largest `row.at` received by this loop (or rather, the latest one executed before a failure, if we care about `process_event` error codes and want to retry). Since `poll` (generally!) runs locked and also does the `ran` checks, it shouldn't make a difference if it is set at the end, or the beginning of this loop.

If no rows were returned, `CURRENT_AT_KEY` can stay unchanged. It will eventually move forward when the next event arrives.

kong/cluster_events/init.lua
```diff
@@ -203,6 +205,62 @@ function _M:subscribe(channel, cb, start_polling)
...
+    if not row.nbf then
+      -- unique callback run without delay
+      local ok, err = pcall(cbs[j], row.data)
+      if not ok and not ngx_debug then
+        log(ERR, "callback threw an error: ", err)
+      end
+
+    else
+      -- unique callback run after some delay
+      local now = row.now or self.strategy:server_time() or ngx_now()
```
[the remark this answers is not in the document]
bungle: About the reasoning:

I think it is more important to keep nodes state sync than apply some cluster to some nodes in different time, e.g. it is a whole lot bigger issues if node a gives different answer than node b. So this applies both ways, but across nodes always about the same (every node is applying changes around `db_update_frequency`.

kong/cluster_events/init.lua
bungle: @hishamhm I think I asked about this here:
https://github.com/Kong/kong/pull/5247#pullrequestreview-321482345

This didn't change. I can change it though.

kong/cluster_events/init.lua
bungle: I mark this as resolved, we decided to move time on, only when there are new events.

kong/cluster_events/init.lua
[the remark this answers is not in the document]
bungle: fallbacks removed, and I agree, not needed and can be even harmful. Your changes on that delta was added, thanks!

kong/cluster_events/init.lua
bungle: Yes, this changed so that we only advance the time when there has been events.

kong/cluster_events/strategies/postgres.lua
```diff
@@ -113,23 +126,35 @@ end
...
+  local query_template = fmt(SELECT_INTERVAL_QUERY, p_chans,
+                                                    p_minat,
+                                                    p_maxat,
+                                                    self.page_size,
+                                                    "%s")
+
+  local page = 0
+  local last_page
   return function()
+    if last_page then
```
kikito: I think the `last_page` optimization should not be included.

Reasoning: people can "take their time" while iterating over the `select_interval` function. By that I mean they could read the first page, keep the token, wait 3 minutes, read the second page, keep the token, wait 3 minutes, then read the next page, etc.

Even if a page returned only 1 record on the last read, in those 3 minutes a lot of things can have happened and I should get those events.

I'm using "3 minutes" as a random "infinitely long" period, but if the gods of concurrency and CPUs align themselves, even a delay of 1ms (perhaps with some IO in another processor sprinkled in) can spell the difference.

kikito: Discussed this privately and was convinced that this is not an issue on this particular case :we are the only users of this API, we consume it immediately, and even if we weren't the early `nil` just means that some events might be queued and processed on the next call to "pull events".

Issue #839, opened 2016-01-06
follow upstream redirects directive


ahmadnassri:
if the upstream responds with a redirect, typically it will send an absolute path that will not match the one Kong clients are calling _(both request host and request path)_.

a `follow_redirects` directive on the API would instruct nginx to retry the same request at the new location **before** sending back to the client, thus eliminating the need for the client to know about the redirect.

thibaultCha:
Should probably use [`proxy_redirect`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect).

catchdave:
Proxy_redirect would fix the issue nicely :)

dag24:
+1, this would be particularly useful when the redirect from the upstream host is unreachable from the end user's machine, but is reachable by Kong (e.g. Kong is exposing an internal service).

anamarialazar:
+1

JnMik:
+1 Need this too

lnryan:
+1 also need this

kl4n4:
+1 same here

john-deng:
+1 need this too

jakebolam:
+1 looking for this

MingStar:
+1 this would be a much-needed option in kong itself (instead of fiddling with the underlying nginx configs)

agarwal-nitesh:
+1

PR #5249, opened 2019-11-22
feat(runloop) upstream redirects

kikito:
### Summary

Alters the Location response header returned by Upstreams in a very specific case, when the host:port coincides with the matched one by the router.

### Full changelog

In addition to the `feat` change itself, this PR includes a couple changes in the tests.

### Issues resolved

Fixes (partially) #839

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -12,7 +12,7 @@ local singletons   = require "kong.singletons"
 local certificate  = require "kong.runloop.certificate"
 local concurrency  = require "kong.concurrency"
 local PluginsIterator = require "kong.runloop.plugins_iterator"
-
+local url          = require "socket.url"
 
 local kong         = kong
 local type         = type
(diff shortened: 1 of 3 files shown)


kikito:
The reason why this is sent as a "Draf" is because I want to invite some discussion.

Questions:

* Right now this cover a very specific circumstance. Should we include others? For example, if the matched occurrs via a `/foo` path in the Route, and the returned `Location` is `http
://<upstream_host:upstream_port>/foo`, where upstream_host and upstream_port coincide with the service host, should we remove the upstream host and port and lust leave `/foo`?
* Should this behavior be part of Kong itself, and be active all the time? It could definitively be exported to a plugin, and be applied only to certain routes.

guanlan:
Should we feature flag this new redirect behavior?

Tieske:
Imo it should be a plugin. Wrt implementation then this seems only a rather narrow use case. If this is added, we should be prepared for requests to enhance it to support all transformations done, including the stripping and prefixing of the path.

bungle:
I am leaning towards plugin too. The current implementation is always on, and that is not what people always want, especially as Kong has never done that before. If not a plugin, it should be a property of  `service` I think.

kikito:
Ok it seems that "move this to a plugin" is decided. I think I like that approach because it allows further tweaking (a plugin can have several options instead of just a `enabled/disabled` boolean). This also made me think about a different approach. The problem which we are trying to solve on this PR is this:

> if the upstream responds with a redirect, typically it will send an absolute path that will not match the one Kong clients are calling (both request host and request path).

I am thinking that the ... "intelligent" approach shown on this pr (where the transformations to the headers are "deduced from the current setup in an implicit way" are difficult to explain, and even after explaining them, they are difficult to know how they will work, exactly.

I believe this is because they are very "indirect". Perhaps a more direct approach would be more straightforward.

What I am thinking about is a plugin which is "`response-transformer` for headers".

Example scenario:

We have a Service/Upstream matching named `myhost.dev`.
It has two targets, with IP addresses `10.20.30.40` and `10.20.30.41`.

What we want is to perform some transformations on the response headers if:
* A responses with a `Location` header AND
* The response http status was a redirect (301, 302, 307, 308)

For requests meeting those requirements, we want:
* If `Location` is from a host like `10.20.30.40` or `10.20.30.41`, we want to replace it with `myhost.dev`
* If `Location` is a relative path (starts with `/`) we want to prefix that with `http://myhost.dev`)

Our fictional plugin could be written like this, and be associated with the Service:

```
{ name = "response-header-transformer",
  config = {
    { rules = {
        { -- array, can have more than 1 rule
          http_status = { 301, 302, 307, 308 },
          header = "location",
          replace = {
            { host = "10\.20\.30\.40", with = "myhost.dev" },
            { host = "10\.20\.30\.41", with = "myhost.dev" }
            { host = "", with = "myhost.dev" }
          }
        }
     }
  }
}
```

The exact config syntax is made up & provisional, but I hope this gets the idea across: the plugin would do some `if`s on the `header_filter` phase and if the conditions are met then it will parse the Location headers and perform a bunch of transformations to it using `ngx.re.sub` before sending it back downstream.

A similar setup could be done for Routes and the `/path` setting, with `replace = { path = ... }` rules.

Advantages: it is easy to know what the plugin does, there is no "built-in magic". The user is the one who decides which particular transformations occur.

Disadvantages:

* If the user adds another target to the upstream they must remember to add another entry to the rules transform. (On this particular case they can probably set a regex which matches a range, but that won't always be an option).
* Using regexes to manipulate domains is iffy, but less so than manipulating whole uris. If you want more fine tuning you are almost to the point where you want to write your own plugin I think.

locao:
Closing this as we decided to move the feature to plugins.

Issue #5252, opened 2019-11-25
(oauth2) more robust refresh tokens


bob983:
### Summary

We're using Kong and its OAuth2 plugin to power our APIs. I'd love to make the OAuth2 plugin more robust and I'm happy to prepare a PR.

At the moment, a refresh token is deleted once it's used - this looks fine and it's according to specification but if there's a network issue or the client fails for some reason, client may end up with old refresh token (it never received the new one) but Kong already holds the new one and knows nothing about the old one.

There's another hidden problem - if the 3rd party OAuth2 app runs on multiple nodes, clients have to implement locking otherwise parallel calls might try to refresh tokens at the same time and if the code is not written well, client could end up with invalid refresh token again. This makes the API hard to use.

Ideally the refresh token would exist some time after it was used (few ours, maybe a day) and attempts to use it would just yield the existing new access / refresh token pair.

#### Simple solution

When I was thinking about the idea the easiest solution seemed to be:

* do not delete the refresh token when a new refresh / access token pair is generated - this is safe, because the access token can't used anyway - the token is expired
* set a TTL on the old refresh token (let's say a few hours) - it would be automatically deleted
* add a link from the old token to the new token - the token refresh code would check existence of a link and if it exists, it would just return the linked fresh refresh token / access token pair

This solution has a small catch - what happens if the new refresh token is also used to generate a fresh refresh token / access token pair?

#### Slightly more complex solution

Separate the refresh token and access token (they are one DB row/entity now) and give them different life-cycles. I did not analyze this solution though

#### Something else

I'm open to suggestions - I only have one goal - make the Authorization Code flow and refresh tokens less susceptible to networking issues and/or parallel refresh.

bob983:
Hey @bungle! 

I'm afraid that dbless and the permanent (or long lived) nature of oauth2 tokens go against one another. Using JWT for access tokens is a good idea but refresh tokens need to be persisted, in memory store is not enough.

I think this could be solved by external OAuth2 provider. 

Using an external OAuth2 provider would be our next choice if we can't make the OAuth2 plugin more robust. There's a [plugin](https://github.com/mogui/kong-external-oauth) for that but I haven't tried it - it doesn't look maintained - but maybe it just works and doesn't have to be maintained :)

bungle:
@bob983 is there any reason why refresh token could not be a `JWT` token? I mean why do they need to be persisted?

[...]

PR #5264, opened 2019-11-26
feat(oauth2) add ability to persist refresh tokens

amberheilman:
### Summary
Add support for persisting oauth2 refresh tokens. Follows guidelines in [RFC6749](https://tools.ietf.org/html/rfc6749#section-6) to allow for the optional persistence of the refresh token when refreshing an access token.

### Related discussion
https://discuss.konghq.com/t/long-lived-refresh-tokens/659
https://discuss.konghq.com/t/long-lived-oauth-2-refresh-tokens/3005

### DIFF
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -52,10 +52,9 @@ end
 
 
 local function generate_token(conf, service, credential, authenticated_userid,
-                              scope, state, existing_token, expiration,
-                              disable_refresh)
+                              scope, state, disable_refresh, existing_token)
 
-  local token_expiration = expiration or conf.token_expiration
+  local token_expiration = conf.token_expiration
 
   local refresh_token_ttl
   if conf.refresh_token_ttl and conf.refresh_token_ttl > 0 then
@@ -289,7 +288,7 @@ local function authorize(conf)
           response_params = generate_token(conf, kong.router.get_service(),
                                            client,
                                            parameters[AUTHENTICATED_USERID],
-                                           scopes, state, nil, nil, true)
+                                           scopes, state, true)
           is_implicit_grant = true
         end
       end
@@ -507,7 +506,7 @@ local function issue_token(conf)
             response_params = generate_token(conf, kong.router.get_service(),
                                              client,
                                              parameters.authenticated_userid,
-                                             scope, state, nil, nil, true)
+                                             scope, state, true)
           end
         end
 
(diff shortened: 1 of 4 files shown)


bob983:
This is a nice coincidence, we started chatting about similar issue here https://github.com/Kong/kong/issues/5252.

dave-shawley:
@bob983 - we ran into the disappearing token problem using another company's OAuth 2 API and decided that we simply couldn't put our customers through the same pain.  Hopefully this will meet your needs as well.  We briefly looked into supporting multiple linked tokens side-by-side instead of making them persistent and it was quite a bit more work than expected.  Ultimately, making refresh tokens immutable until revocation met our needs.

We did implement [RFC-7009 token revocation] as a separate service using the `DELETE /oauth2_tokens/:token_id` Admin API endpoint.  That handled the need to explicitly revoke either an access token or a refresh token mentioned in #5652 .  We could look into contributing that back if there is interest.

[RFC-7009 token revocation]: https://tools.ietf.org/html/rfc7009

bob983:
@dave-shawley yeah, this would definitely work for us well :) 

We handle the token revocation via our own UI which in turn calls the Kong Admin API and deletes the refresh token / acess token pair and thus RFC-7009 is not necessary for us but it would be a nice addition.

bob983:
Any news on this PR? I'm eagerly watching it :)

bungle (changes requested):
This is bit off-topic, but mind you change these as well, perhaps another commit:
refactor(oauth2) change generate_token parameters

bungle:
In general I think this is good... only thing I am hesitating is the quite long and perhaps a bit misleading (as all the refresh tokens are persistent, but not reusable currently):
```
persistent_refresh_token
```

We already have:
```
config.refresh_token_ttl
```

@dave-shawley, what do you think e.g.:_
```
config.refresh_token_reuse=true
```
or
```
config.refresh_token_reusable=true
```
or
```
config.reuse_refresh_token=true
```

Pick what you like the best. I think after that we can go ahead and merge. Please also rebase to latest `next` and change the PR to be merged to `next` as that is where `feats` go.

Thank you for the PR. I have to say that these are very high in quality. We should have reviewed this earlier. I'll focus to get these in!

amberheilman:
@bungle,
PR has been rebased against next. 
Changed the configuration option from `persistent_refresh_token` to `reuse_refresh_token` as well.

bungle (approved):

bungle:
@amberheilman, thank you sir! it is finally merged now, and will land on `2.1`.

Issue #5260, opened 2019-11-26
route with headers overwrite bug


marlonfan:
### Summary

I've defined two routes

a:

```bash
curl -X POST http://localhost:8001/services/demo1/routes --data "name=routeA" --data "hosts[]=example.com" --data "paths[]=/x" --data "headers.env[]=a"
```
b:

```bash
curl -X POST http://localhost:8001/services/demo2/routes --data "name=routeB" --data "hosts[]=example.com" --data "paths[]=/x"
```

Now request routeA, In line with expectations
Request routeB, In line with expectations
Request routeA again, routeB overwrited routeA

marlonfan:
I tried 1.3.0 version and 1.4.0 version You can reproduce it

PR #5267, opened 2019-11-27
fix(router): fix router match with headers will overwrite bug

marlonfan:
[issue5260](https://github.com/Kong/kong/issues/5260)

When a header match exists, it is overwritten.

So, deferred judgment cache.  if trigger headers match rule, ignore routes cache.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -1211,20 +1211,6 @@ function _M.new(routes)
     ctx.dst_port       = dst_port or ""
     ctx.sni            = sni or ""
 
-    -- cache lookup (except for headers-matched Routes)
-
-    local cache_key = req_method .. "|" .. req_uri .. "|" .. req_host ..
-                      "|" .. ctx.src_ip .. "|" .. ctx.src_port ..
-                      "|" .. ctx.dst_ip .. "|" .. ctx.dst_port ..
-                      "|" .. ctx.sni
-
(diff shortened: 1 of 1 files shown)


bungle:
@marlonfan thanks! that is great PR! mind you adding a test for it here:
https://github.com/Kong/kong/blob/master/spec/01-unit/08-router_spec.lua

marlonfan:
Ok, I'll supplement the test cases later

kong/router.lua
```diff
@@ -1250,6 +1236,31 @@ function _M.new(routes)
...
+    -- if trigger headers match rule, ignore routes cache
+
+    local cache_key = req_method .. "|" .. req_uri .. "|" .. req_host ..
+                      "|" .. ctx.src_ip .. "|" .. ctx.src_port ..
+                      "|" .. ctx.dst_ip .. "|" .. ctx.dst_port ..
+                      "|" .. ctx.sni
+
+    do
+      local match_t = cache:get(cache_key)
+      if match_t and (hits.header_name == nil and hits.header_name == '') then
```
hishamhm:
```suggestion
      if match_t and hits.header_name == nil then
```

The following test for `''` is redundant (it will never be `nil` _and_ `''`, but even if it was `or`, header names are never empty-string).

PR #5268, opened 2019-11-27
feat(oauth2) pkce

### LINKED ISSUE #1121: [OAuth2] any plans to support PKCE in OAuth flow?
Hello, 

do you plan to support [Proof Key for Code Exchange](https://tools.ietf.org/html/rfc7636) concept (RFC-7636) in OAuth plugin? We plan to use Kong with OAuth2 for securing sensitive API services and any additional security is appreciated...

Thank you.

amberheilman:
### Summary

Add support for Proof Key for Code Exchange ([PKCE](https://tools.ietf.org/html/rfc7636)). This implementation adds a field that distinguishes whether a client is public or confidential, and imposes rules around issuing codes/tokens based on the client_type. PKCE was introduced as a method for public applications that were susceptible to the authorization code interception attack.

### Full changelog

* Adds a client_type to oauth2_credentials
* Adds support for PKCE

### Issues resolved

Fix #1121


bungle:
On:
https://tools.ietf.org/html/draft-ietf-oauth-security-topics-13

It is noted that `PKCE` should be used for everything. Not only for `public`.

dave-shawley:
@bungle I changed the PKCE implementation so that it is available for all clients and required for public (non-confidential) clients.  I stopped short of requiring it for authorization code grants since there are too many OAuth2 client implementations that do not support PKCE today.

Should I add a configuration option to require it for all authorization code flows?

bungle:
@dave-shawley, @amberheilman, I am focusing this on next 7 days in hopes we can merge this. So first thing we need is to rebase this to latest `next`. I will focus to changes tomorrow.

bungle:
@dave-shawley, it might make sense to add option for PKCE.

Let's think the cases:
1. client requests pkce
2. client does not request pkce

If client requested pkce then the pkce should be enforced by server.
If client didn't request pkce then it should not be enforced by server.

But of course in secure environments, you might want to enforce it on server always, or perhaps by client type. So it sounds we like to have an options. `config.pkce` could be the parameter, perhaps with `None`, `Lax`, `Strict`.

`None` if client request PKCE we will do it but we will not error if client gives wrong data to token endpoint, thus checks `code_verifier`, but only writes warning or notice to logs but still gives tokens if everything is otherwise ok.

`Lax`, we will only enforce this on `public` clients, for `confidential` see `None`.

`Strict`, we will enforce this on all the clients.


@dave-shawley, @amberheilman, how does that sound?

dave-shawley:
@bungle I rebased our branch over next instead of master and run the tests with pongo.  Is pongo a valid test ... it seems to only use images from hub.docker.com and there isn't a next published there.

dave-shawley:
@bungle looks like the plugin tests succeeded but the integration suite failed.  I'm going to chat with @amberheilman when I get into the office (about 30 minutes) before I respond to the configuration question.  Thanks for finding some time to look at this.

dave-shawley:
@bungle the configuration changes make sense.  The only one that I don't completely understand is why PKCE is not enforced for confidential clients when the appropriate bits are in the authorization request in _Lax_ mode.  I think that it would be more useful if _Lax_ mode fully implements PKCE if the client includes the challenge information in the authorization request _and_ requires it for public clients.  Otherwise the only way to use PKCE for confidential clients is to require it for all clients which (I believe) is going to make it difficult to use _Strict_ mode.

In any case, we aren't going to be able to slot in the changes until our next two-week sprint/cycle.  One of us might be able to get to it before then, but I can't guarantee much action on this for another week or two.  Any chance at getting a review started on #5264 since that is the other side of our implementation?  The implementation might address #5252 as well.

bungle:
@dave-shawley, thanks for getting back! Yes, our `test`suite is currently broken on `integration tests`. Sorry about it. It is good enough if plugin tests for this plugin pass.

Yes, I think your comment on my strange/mixed rules is spot on, let's try again:

Client | None | Lax | Strict
Public Client with PKCE | PKCE Enforced on Token Endpoint | PKCE Enforced on Token Endpoint | PKCE Enforced on Token Endpoint
Public Client without PKCE | PKCE Not Enforced | Fails on Authorization Endpoint | Fails on Authorization Endpoint
Confidential Client with PKCE | PKCE Enforced on Token Endpoint | PKCE Enforced on Token Endpoint | PKCE Enforced on Token Endpoint
Confidential Client without   PKCE | PKCE Not Enforced | PKCE Not Enforced | Fails on Authorization Endpoint

So the Only difference being the `Lax`.

evansmwendwa:
Is work still ongoing with this?

amberheilman:
@bungle @evansmwendwa I have updated to include a configuration option for pkce. I believe all comments have been addressed here and this is up for review.

bungle:
Yes, I think we are almost there to merge this. I gave some small change request above. If those get worked, I think we can merge this.

bungle:
@amberheilman did you see my latest comments above ^

amberheilman:
@bungle awesome, thanks for reviewing again. I will try to get these changes done today.

amberheilman:
@bungle all comments are addressed, this is ready for another pass

dave-shawley:
@bungle has this been assigned to a milestone or release?  wasn't sure if this is part of the [2.1.0 Milestone](https://github.com/Kong/kong/milestone/37) or not.

PR #5271, opened 2019-11-28
chore(ci) changed a make task and explicitly defined some variables

hutchic:
In the process of making kong-build-tools capable of building / releasing kong-enterprise there's a few variables we need to explicitly define.

related: https://github.com/Kong/kong-build-tools/pull/166


.requirements
```diff
@@ -1,5 +1,11 @@
+KONG_PACKAGE_NAME=kong
+KONG_CONFLICTS=kong-enterprise-edition
+KONG_LICENSE="ASL 2.0"
+
 KONG_GMP_VERSION=6.1.2
 RESTY_VERSION=1.15.8.2
 RESTY_LUAROCKS_VERSION=3.2.1
 RESTY_OPENSSL_VERSION=1.1.1d
 RESTY_PCRE_VERSION=8.43
+LYAML_VERSION=6.2.3
```
hishamhm: Do we really need these? I thought the need to explicitly version `lyaml` (the Lua bindings) were removed here:

https://github.com/Kong/kong-build-tools/pull/124/files

(`lyaml` should be versioned via the rockspec like all other Lua module dependencies)

PR #5277, opened 2019-12-01
fix(acl) reject anonymous consumer

### LINKED ISSUE #4480: ACL plugin: 401 vs 403

jeremyjpj0916:
Previous PR related: https://github.com/Kong/kong/pull/5236

Summary(Copied from prior PR)
We started looking into leveraging the anonymous user functionality in Kong thats native for helping use multiple auth patterns on a given service/route:
https://docs.konghq.com/0.14.x/auth/#multiple-authentication

But we wanted to use OAuth2 + JWT programmatic auth, alongside the ACL plugin to control the proper authorization of given authenticated consumers. This is nice because different consumers can call the same proxy with preferred authentication patterns.

Out of the box what happens when a user is either missing or a bad token when doing the anonymous user pattern with OAuth2 + JWT + ACL plugins is a HTTP Status 403: "You cannot consume this service" response. Which to a client is not technically correct from their perspective with a bad/missing/expired tokens.

Full changelog
Implement: Modified handler.lua to change ACL behavior to support 401 for no credentials and no authenticated groups(Such as the anon user)
Testcases Modified: /spec/03-plugins/18-acl/02-access_spec.lua

Fix #4480

cc @hishamhm @bungle

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -39,7 +39,7 @@ local ACLHandler = {}
 
 
 ACLHandler.PRIORITY = 950
-ACLHandler.VERSION = "2.1.0"
+ACLHandler.VERSION = "2.2.0"
 
 
 function ACLHandler:access(conf)
@@ -59,6 +59,10 @@ function ACLHandler:access(conf)
   end
 
   local to_be_blocked
+
+  if not kong.client.get_credential() and not ngx.ctx.authenticated_groups then
+    return kong.response.exit(401, { message = "Unauthorized" })
+  end
 
   -- get the consumer/credentials
   local consumer_id = groups.get_current_consumer_id()
--- a/spec/03-plugins/18-acl/02-access_spec.lua
+++ b/spec/03-plugins/18-acl/02-access_spec.lua
@@ -758,9 +758,9 @@ for _, strategy in helpers.each_strategy() do
             ["Host"] = "acl1.com"
           }
         }))
-        local body = assert.res_status(403, res)
+        local body = assert.res_status(401, res)
         local json = cjson.decode(body)
-        assert.same({ message = "You cannot consume this service" }, json)
+        assert.same({ message = "Unauthorized" }, json)
       end)
 
       it("should fail when not in whitelist", function()
@@ -1259,9 +1259,9 @@ for _, strategy in helpers.each_strategy() do
             ["Host"] = "acl13.com"
           }
         }))
-        local body = assert.res_status(403, res)
+        local body = assert.res_status(401, res)
         local json = cjson.decode(body)
-        assert.same({ message = "You cannot consume this service" }, json)
+        assert.same({ message = "Unauthorized" }, json)
       end)
     end)
   end)


jeremyjpj0916:
Looks like that logic breaks a # of test cases, will need to investigate:

https://travis-ci.org/Kong/kong/jobs/619161271?utm_medium=notification&utm_source=github_status

```
[  FAILED  ] 18 tests, listed below:
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 716: Plugin: ACL (access) [#postgres] Mapping to Consumer or Authenticated Groups should work with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 728: Plugin: ACL (access) [#postgres] Mapping to Consumer or Authenticated Groups should work with consumer without credentials
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 740: Plugin: ACL (access) [#postgres] Mapping to Consumer or Authenticated Groups should work with authenticated groups without credentials
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 777: Plugin: ACL (access) [#postgres] Simple lists should fail when not in whitelist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 799: Plugin: ACL (access) [#postgres] Simple lists should work when in whitelist with authenticated groups
1m[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 821: Plugin: ACL (access) [#postgres] Simple lists should not send x-authenticated-groups header when hide_groups_header flag true
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 843: Plugin: ACL (access) [#postgres] Simple lists should send x-authenticated-groups header when hide_groups_header flag false
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 863: Plugin: ACL (access) [#postgres] Simple lists should work when not in blacklist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 883: Plugin: ACL (access) [#postgres] Simple lists should fail when in blacklist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 907: Plugin: ACL (access) [#postgres] Multi lists should work when in whitelist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 929: Plugin: ACL (access) [#postgres] Multi lists should fail when not in whitelist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 951: Plugin: ACL (access) [#postgres] Multi lists should fail when in blacklist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 972: Plugin: ACL (access) [#postgres] Multi lists should work when not in blacklist with authenticated groups
1m[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 992: Plugin: ACL (access) [#postgres] Multi lists should not work when one of the ACLs in the blacklist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 1012: Plugin: ACL (access) [#postgres] Multi lists should work when one of the ACLs in the whitelist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 1032: Plugin: ACL (access) [#postgres] Multi lists should not work when at least one of the ACLs in the blacklist with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 1151: Plugin: ACL (access) [#postgres] Real-world usage should not fail when multiple rules are set fast with authenticated groups
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 1232: Plugin: ACL (access) [#postgres] Permits with authorized groups even when anonymous consumer is present
```

bungle:
Thanks @jeremyjpj0916, I'll take a look at this this week.

bungle:
@jeremyjpj0916 we switched the tasks last week as @hishamhm wanted to look at this. So I think he will get back to you soon.

jeremyjpj0916:
Sounds good. There are a few ways to tackle this. Trying to do it native means re-writing a # of tests as we see in this PR now, adding a new param feels bad but would also ensure no unexpected consequences for existing users. Appreciate the diligence to stay on top of it during the holiday months here.

hishamhm:
@jeremyjpj0916 sorry man, ended up swamped on other things but we really want to get back to this ASAP

jeremyjpj0916:
@hishamhm not a problem, no rush on it. Just got back from a 5 day cruise and also on PTO rest of year and through most of early Jan. Will be online though to help discuss the feature/pr though throughout. Hope you all at Kong have a great EOY!

hishamhm:
@jeremyjpj0916 I played a bit with your PR branch to get acquainted with the issue, and it seems that applying this change on top of it makes the test cases you listed above stop failing, but then [second one of the two test cases you changed](https://github.com/Kong/kong/pull/5277/files#diff-388bbf8fd9d81a32876ee9390d396aedR1262) goes back to producing a 403 ([the first one switches to 401](https://github.com/Kong/kong/pull/5277/files#diff-388bbf8fd9d81a32876ee9390d396aedR761) as you intended):

```diff
diff --git a/kong/plugins/acl/handler.lua b/kong/plugins/acl/handler.lua
index 5ebd96858..14fc37158 100644
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -60,10 +60,6 @@ function ACLHandler:access(conf)
 
   local to_be_blocked
 
-  if not kong.client.get_credential() and not ngx.ctx.authenticated_groups then
-    return kong.response.exit(401, { message = "Unauthorized" })
-  end
-
   -- get the consumer/credentials
   local consumer_id = groups.get_current_consumer_id()
   if not consumer_id then
@@ -72,6 +68,10 @@ function ACLHandler:access(conf)
       kong.log.err("Cannot identify the consumer, add an authentication ",
                    "plugin to use the ACL plugin")
 
+      if not kong.client.get_credential() then
+        return kong.response.exit(401, { message = "Unauthorized" })
+      end
+
       return kong.response.exit(403, {
         message = "You cannot consume this service"
       })
```

I am not knowledgeable enough on the `acl` and auth plugins to understand precisely what's going on. Do you think this change is sufficient for the behavior you need or must the second one switch to 401 as well?

My reading of the [test case that fails](https://github.com/Kong/kong/pull/5277/files#diff-388bbf8fd9d81a32876ee9390d396aedR1262) with the patch in this comment is the following: it uses a fixture plugin to set the authenticated group to `admin` but the route _does_ have an entry set for it in the `whitelist` for `anonymous`. So, `anonymous` is whitelisted, but `admin` isn't, and you tried to access as `admin`. Should that produce a 401 or a 403?

bungle:
@hishamhm I don't think this is a right change.

Some background:
1. acl plugin originally worked only with kong consumers that may have acls
2. we added support for ngx.ctx.authenticated_groups so that auth plugins leveraging 3rd party auth and when not doing consumer mapping you can just list array of strings in context (acls).
3. kong auth plugins that have consumer do also set credential by which user authenticated (a row in some auth plugin related table that has foreign key to consumer)
4. for credential we also allow setting virtual credential and that comes handy with e.g. ratelimiting where you don’t again need to have kong consumer or credential stored in kong db or plugin related table, thus only exists in context of request

I think Jeremys original was correct, the question just remained that is it ok to change the tests even when it is slightly on a breaking side. I have not looked which tests actually fail, so there might be something that is more problematic

bungle:
the one that Jeremy quoted above where test expected 200 but got 401 feels like correct, but correctness might have to do whether plugin is doing whitelisting OR blacklisting. For whitelist I think it is correct, but for blacklisting i think 200 might be correct.

bungle:
@jeremyjpj0916 / @hishamhm,

If we change:
```lua
  if not kong.client.get_credential() and not ngx.ctx.authenticated_groups then
    return kong.response.exit(401, { message = "Unauthorized" })
  end
```
to:
```
  if not kong.client.get_credential() and
     not kong.ctx.shared.authenticated_groups and
     not ngx.ctx.authenticated_groups then
    return kong.response.exit(401, { message = "Unauthorized" })
  end
```

There will be only two tests failing:

```
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 728: Plugin: ACL (access) [#postgres] Mapping to Consumer or Authenticated Groups should work with consumer without credentials
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 1256: Plugin: ACL (access) [#postgres] Forbids with authorized groups even when anonymous consumer is present
```

The second one is `401` (new) vs`403` (old), but the first one is `401` (new) vs `200` (old). So the second one is ok, I think we can change the test. Let's focus on the first:

```
[  FAILED  ] spec/03-plugins/18-acl/02-access_spec.lua @ 728: Plugin: ACL (access) [#postgres] Mapping to Consumer or Authenticated Groups should work with consumer without credentials
```
```lua
      it("should work with consumer without credentials", function()
        local res = assert(proxy_client:get("/request", {
          headers = {
            ["Host"] = "acl8.com"
          }
        }))

        local body = cjson.decode(assert.res_status(200, res))
        assert.equal("anonymous", body.headers["x-consumer-groups"])
        assert.equal(nil, body.headers["x-authenticated-groups"])
      end)
```

So it seems to be an `anonymous` consumer that does have `groups` that is now failing.

And if that is the case, shouldn't we just change `403` to `401` when we don't find groups?

That is, changing this:
https://github.com/Kong/kong/blob/next/kong/plugins/acl/handler.lua#L71

And perhaps changing this:
https://github.com/Kong/kong/blob/next/kong/plugins/acl/handler.lua#L97-L100

To handle `err` and `nil` groups separately, e.g. if not `err` but `nil` groups then `401`.

I can try to prepare a sister-pr to this where you can see in who what I am trying to propose here.

bungle:
closing this as we moved it further with #5440 and #5452.

PR #5281, opened 2019-12-02
fix(reports) do not log error on TCP connection issues

hbagdi:


hishamhm:
@hbagdi CI reported linting errors. Also, I believe these log entries can be removed without keeping them around as comments

Issue #4469, opened 2019-04-03
Service URL formation changed when using routes with strip_path=false


dliberman:
### Summary

Service request URL formation changed after KongCE 0.15 if you create routes with strip_path=false. If you create the service as ```url=https://mockbin.org/request``` (without the trailing slash), for example, and add a route /something to it, it will form the url as ```mockbin.org/requestsomething```

### Steps To Reproduce

1. Create a service:
```shell
curl -i -X POST \
  --url http://localhost:8001/services/ \
  --data 'name=potato' \
  --data 'url=https://mockbin.org/request'
```

2. Create a simple route for the service, with *strip_path=false*
```shell
curl -i -X POST \
  --url http://localhost:8001/services/potato/routes \
  --data 'strip_path=false' \
  --data 'paths[]=/fries'
```

3. Make an http call to this server/path (using httpie here):
```shell
http localhost:8000/fries
```

4. Check result JSON:
- On 0.14.1 and before (also tested and works on Kong EE 0.33-2), request url formed is  `http://localhost/request/fries`. On 0.15 and after (tested also on 1.1.0) the url formed is `http://localhost/requestfries"`

### Additional Details & Logs

- Kong version; 0.15 and after, tested until 1.1.0
- As mockbin.org accepts requests like ```mockbin.org/request```, ```mockbin.org/requestsomething```, ```mockbin.org/request/something```, this is easy to reproduce there.
- As stated in the summary, it works if you create the service with the trailing slash at the url. But this will break some services we have in place already (when migrating from <0.15 to >=0.15) that have more complex url syntax.
- Full result from example request:
```json
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: host,connection,x-forwarded-for,x-forwarded-proto,x-forwarded-host,x-forwarded-port,x-real-ip,kong-cloud-request-id,kong-client-id,user-agent,accept-encoding,accept,x-request-id,via,connect-time,x-request-start,total-route-time
Access-Control-Allow-Methods: GET
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 996
Content-Type: application/json; charset=utf-8
Date: Wed, 03 Apr 2019 17:35:26 GMT
Etag: W/"3e4-Sq0Wd/KoVkJPqkIVPN3ULQ"
Kong-Cloud-Request-ID: df23405223db9dce258be56fc49df266
Server: openresty/1.13.6.2
[...]

PR #5292, opened 2019-12-04
feat(router) add routes.path_handling attribute

kikito:
This PR introduces a new Routes attribute called path_handling.

It complements #5216 and #5227 by making it possible to select the
behavior the router will have when combining the Service Path, the Route
Path and the Request path into a single path sent to the upstream.

For now `path_handling` accepts two values:

* "v0" (the default from now on) makes paths in the router behave like in Kong 0.x (like they did
  before #3780)
* "v1" makes paths behave like in Kong 1.x (like they did after #3780)

Notice that there is a companion pr to this one in #5341 . It adds some intelligence to migrations, setting old routes to v0 / v1 depending on "from where you are migrating from" (it does *not* change the schema default, so new routes will always be v0 unless otherwise specified).


spec/fixtures/router_path_handling_tests.lua
```diff
@@ -0,0 +1,111 @@
...
+-- copy-pasting them in several places.
+--
+-- The tests can obtain this table by requiring
+-- "spec.fixtures.router_path_handling_tests"
+--
+-- Note that the tests are parsed into a hash form at the end
+-- of this file before they are returned
+
+local tests = {
+  -- upstream_path   route_path   strip_path  request_path    expected_path
```
bungle: In general I think this is easier to read if it is:
```
-- request_path     route_path     service_path     expected_path     strip_path
```

`strip_path` can be on either side I think.

spec/fixtures/router_path_handling_tests.lua
```diff
@@ -0,0 +1,194 @@
...
+-- The tests can obtain this table by requiring
+-- "spec.fixtures.router_path_handling_tests"
+--
+-- Note that the tests are parsed into a hash form at the end
+-- of this file before they are returned
+--
+-- All the tests where v1 differs from v0 are marked with !
+
+local tests = {
+  -- upstream_path   route_path   strip_path  path_handling  request_path    expected_path           --    0
```
bungle: In general I think this is easier to read if it is:
```
-- request_path     route_path     service_path     expected_path     strip_path
```

`strip_path` can be on either side I think.

kong/db/schema/entities/routes.lua
```diff
@@ -33,6 +33,7 @@ return {
                                    }, },
     { regex_priority = { type = "integer", default = 0 }, },
     { strip_path     = { type = "boolean", default = true }, },
+    { path_handling  = { type = "string", default = "v1", one_of = { "v0", "v1" }, }, },
```
bungle: I am strongly against `v1` being the default.

bungle: Even the community asked this many times before:
https://github.com/Kong/kong/issues/4469#issuecomment-486748567

(`v1` == `insane`, `v0` == `sane`).

kong/db/migrations/core/007_140_to_200.lua
```diff
@@ -1,6 +1,14 @@
 return {
   postgres = {
-    up = [[]],
+    up = [[
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY "routes" ADD COLUMN "path_handling" TEXT;
```
bungle: we have not used `ADD COLUMN` on previous migrations, but just `ADD`
```suggestion
        ALTER TABLE IF EXISTS ONLY "routes" ADD "path_handling" TEXT;
```

Just for sake of consistency. I am with that `COLUMN` there too.

autodoc/data/admin-api.lua
```diff
@@ -611,6 +611,14 @@ return {
             strip the matching prefix from the upstream request URL.
           ]]
         },
+        path_handling = {
+          description = [[
+            Controls how the Service Path, Route Path and requested path are combined when sending a request to the
+            upstream. While these often will be just concatenated, there are some edge cases where they will not; for
+            instance, Kong will remove slashes to avoid sending "double slashes" on the request.
+            `"v0"` is the behavior used in Kong 0.x. `"v1"` is the behavior used in Kong 1.x.
```
bungle: The most important thing to document is how the concatenation happens:
`v0` concatenates `service.path` to possible stripped path by ensuring there is a single `/` between them
`v1` concatenates `service.path` to possible stripped path without `/` separator.

`v0` treats the `service.path` as a path with complete `segments`
`v1` treats the `service.path` as a `prefix`

I think the current description as of now gives little value to reader, and he needs to perhaps go to code and try to figure out. Yes, we could document the bugs in them too, like how they endup removing chars from `service.path` in some cases.

bungle:
Just as a discussion, but perhaps we can find better names:
`v0` = `segments` | `segmented` | `slash_prefixed` | `slash_concatenated` | `slash_ensured` | `slash_separated` | `append_with_slash_separator`
`v1` = `string` | `prefixed` | `append_without_separator`

spec/fixtures/router_path_handling_tests.lua
kikito: Why do you think that is better? I decided to go with "Service attributes / Route Attributes / Request / Result" because that's the order in which one would set up things when doing the tests manually: first create the Service, then the Route, then make a request, and then check the resulting path.

bungle: I am thinking it in terms of request:

1. there is request path (that comes in)
2. that path is matched to route path (and route has an service)
3. request is proxied to service_path
4. upstream path is constructed from service path and request path

bungle: @kikito please check with @thibaultcha as he said that it is impossible to read.

I think originally when this was created was thinking a bit from `concatenation` perspective, but stripping makes you jump back and forth.

in that case I think the best order would be:
```
route_path  service_path  request_path  proxied_path
```

`service_path` is concatenated with `request_path` ... but that can be stripped by `route_path`. So... I think it is best to order it chronologically on what happens when request comes in.

autodoc/data/admin-api.lua
kikito: I have done my best to document this as requested on this comment. Leaving this open in case it is not satisfactory.

spec/fixtures/router_path_handling_tests.lua
kikito: I see now, thanks for explaining.

You see the test from the perspective of "a request coming to an existing system, and what it *encounters*". I see it from the POV of "a programmer building tests step by step".

> stripping makes you jump back and forth

I believe putting `strip_path` or `path_handling` at the end makes you jump back and forth too. I believe that they belong next to `route_path`, as they are part of the Route object, and they are "encountered" before we can get a "result" (`proxied_path`).

However I still believe the order I chose is more understandable because it is similar to what we do in most tests. That is the strongest point about it IMHO.

kong/db/schema/entities/routes.lua
kikito: I have no strong feelings about this. I have changed the tests so they don't "assume a default" on `path_handling`. This way if we want to change this, we can just change it in the schema.

spec/fixtures/router_path_handling_tests.lua
bungle: I have no idea how the routes and services are build. There is no actual order. You can create them in almost any order. Thus hard to wrap in minde "a programmer building tests". This is a matrix of tests that you read from top to bottom and the easiest way is to memorize them as they enter to system, and not by some arbitrary order some specific programmer decided to follow.

bungle: Yes, `route_strip_path` and `route_path_handling` can be specified next to `route_path`.

bungle:
```
-- request_path  route_path  route_stip_path  route_path_handling  service_path  proxied_path  strip_path
```

As your order is almost the same as before and it was voiced several times that it is hard to read, we probably want something else.

kikito: > There is no actual order. You can create them in almost any order [...] some arbitrary order some specific programmer decided to follow.

I can understand that it is not what you would have preferred, but it is hardly arbitrary. You can not create a Route without creating the Service first. And you can't make a (successful) request without creating a Route first. And you can't check a result without doing the request first.

We have used this order before many times in other tests. I honestly don't know how to explain this better.

> It was voiced several times that it is hard to read, we probably want something else.

The order for which it was voiced that it was hard to read had the `strip_path` at the end, separated from the rest of the Route attributes.

> This is a matrix of tests that you read from top to bottom and the easiest way is to memorize them as they enter to system, and not by some arbitrary order some specific programmer decided to follow.

I don't think it needs to be read from top to bottom; I its main purpose is to *facilitate finding the individual case you are looking for* (for debugging or adding new cases in the right spot). I also don't think memorizing the whole list is necessary, just the individual row one is studying. On that direction, here's some other changes we could make:

* Reorder the *rows* as well, so that the vertical order mimics the column order.
* Using more useful path names. Instead of `/fee/bor` we could use `/true/v0` for the `strip_path = true, path_handling = v0` cases, for example.

I didn't do those changes here because waddling through the git diffs is already quite daunting as they are now. If you think this could help, I can make those changes in an extra commit.

hbagdi:
I know there has been a lot of discussion internally and externally around
this feature but I think we've an opportunity to not only
solve this problem and also add a highly-demanded
feature. The feature I'm going to talk about is one feature that I've
seen most common asked on Kong Nation and we don't have a good solution for it.

The only reason for jumping on this thread is to avoid introducing yet another
attribute on the route entity in future so please hear me out.

The feature ask is, how to configure Kong so that we can have two routes and
one service in Kong, representing the following configuration:
- Requests with `/foo/v1/customers` are sent upstream with path `/api/customers`
- Requests with `/foo/v1/orders` are sent upstream with path `/api/orders`

There are only two solutions that exist today:
- Use a plugin to transform the reqeust
- Configure two services in Kong with `/api/customers` and `/api/orders` as
  `service.path` and then set `route.strip_path` to `true`

The gist here is that users want to strip only a small segment of the entire
path of the route. This is such a common use-case that it should be in the
core and not warrant a plugin.

I propose that instead of adding a `path_handling` attribute to the route
entity, we should add a `strip_segment` (not married to the name) string attribute,
with a default of `null` in the JSON representation, which adds the capability of stripping
only part of the path being used in the `route.path` attribute.

`strip_segment` and `strip_path` attributes will be mututally exclusive i.e.
you cannot set `strip_segment` to a string value when `strip_path` is set to
`true`.

`strip_segment` must be a prefix substring of `path`.

I've taken @kikito's table and worked upon it to demonstrate how this
works in solving both the issues mentioned above.

Here is the table, there are three combinations
possible for `strip_path` and `strip_segment` (FF, FT, TF):

| `service.path` | `route.path` | `route.strip_path` | `route.strip_segment` | request path | upstream path |
|----------------|--------------|--------------------|-----------------------|--------------|---------------|
| `/s`           | `/fvX`       | `false`            | `null`                | `/fvXreq`    | `/s/fvXreq`   |
| `/s`           | `/fvX`       | `false`            | `/`                   | `/fvXreq`    | `/sfvXreq`    |
| `/s`           | `/tvX`       | `true`             | `null`                | `/tvXreq`    | `/sreq`       |
| `/s`           | `/fvX/`      | `false`            | `null`                | `/fvX/req`   | `/s/fvX/req` |
| `/s`           | `/fvX/`      | `false`            | `/`                   | `/fvX/req`   | `/sfvX/req`  |
| `/s`           | `/tvX/`      | `true`             | `null`                | `/tvX/req`   | `/sreq`      |


And some more cases:

| `service.path` | `route.path` | `route.strip_path` | `route.strip_segment` | request path | upstream path |
|----------------|--------------|--------------------|-----------------------|--------------|---------------|
| `/s`           | `/fvX`       | `false`            | `null`                | `/fvX/req`    | `/s/fvX/req`   |
| `/s`           | `/fvX`       | `false`            | `/`                   | `/fvX/req`    | `/sfvX/req`    |
| `/s`           | `/tvX`       | `true`             | `null`                | `/tvX/req`    | `/s/req`       |


And finally the feature I think we should add would work like so:

| `service.path` | `route.path` | `route.strip_path` | `route.strip_segment` | request path | upstream path |
|----------------|--------------|--------------------|-----------------------|--------------|---------------|
| `/api`           | `/api/v1/app1`       | `false`            | `/api/v1`                | `/api/v1/app1/foo`    | `/api/app1/foo`   |
| `/api`           | `/api/v1/app2`       | `false`            | `/api/v1`                | `/api/v1/app2/foo`    | `/api/app2/foo`   |


I think this solve the problem we are trying to solve in this PR more flexibly too.
Please let me know your thoughts.

bungle:
Here is the BEST proposal:

#### Change `route.strip_path` (boolean) to `route.strip_paths` (array)

Let's say we have the following route definition:

```
route = {
  paths = { "/cat", "/dog" },
  strip_path = true
}
```

If user migrates from `v0` he will get the following route definition:
```
route = {
  paths = { "/cat", "/dog" },
  strip_paths = { "cat", "dog" }
}
```

If user migrates from `v1` he will get the following route definition:
```
route = {
  paths = { "/cat", "/dog" },
  strip_paths = { "/cat", "/dog" }
}
```

Similarly we the following route definition:
```
route = {
  paths = { "/cat", "/dog" },
  strip_path = false
}
```

would be migrated from `v0` to:
```
route = {
  paths = { "/cat", "/dog" },
  strip_paths = { "", "" }
}
```

and from `v1` to:
```
route = {
  paths = { "/cat", "/dog" },
  strip_paths = { "/", "/" }
}
```

Arguable this is the `BEST` proposal. :-)

kong/db/schema/entities/routes.lua
kikito: This has been decided. V0 is the new default.

spec/fixtures/router_path_handling_tests.lua
kikito: I have added further commits refactoring the tests:
* Used more meaningful path names (i.e. `/service` instead of `/fee/bor`)
* Compressed the tests(if the same test works the same in `v0` and `v1`, it gets a single line named `{v0,v1}` instead of two.
* Reordered the tests so it's always `service attrs / route attrs / request`.

hishamhm:
@kikito Kudos for the hard work cleaning up the test suite!

I ran all `"v1"` tests over current `master` (1.4.2) and also all `"v0"` tests over 0.14.1 to check for behavior regressions.

I'm happy to report that all `"v1"` tests pass on `master`!

For the `"v0"` behavior, I had a few mismatches (which report as `[ FAILED ]` in the log), and also some cases where the route failed to match given the parameters (which report as `[ ERROR ]` in the log). Here are the raw log results of running the unit test suite with the `#slash` tests:

https://gist.github.com/hishamhm/c2d19a6c9b45098945834eb4a0606833

I ran these by checking out 0.14.1, replacing its `spec/` folder with that of this PR, making minor adjustments to `spec/01-unit/08-router_spec.lua` and `spec/helpers.lua` and running that unit test alone — adding `setmetatable(_G, nil)` to the top of `bin/busted` also reduces some noise. I ran the unit tests only (otherwise I would have to downgrade OpenResty, a bunch of libraries, etc.).

I did not do a deeper analysis, but looks like the `[ FAILED ]` cases are mismatches on trailing slashes which seem easy to fix. The `[ ERROR ]` cases are caused by `router.exec()` returning `nil`; apparently the routes don't match — haven't checked why.

Here's the modified `spec` folder if you want to try to reproduce this test scenario with 0.14.1: [spec-for-path-tests-0.14.1.tar.gz](https://github.com/Kong/kong/files/3975300/spec-for-path-tests-0.14.1.tar.gz)

kikito:
Hi @hishamhm, thanks for the feedback and good work.

I put your changed spec [in a branch](https://github.com/Kong/kong/tree/tmp/0.14.1-path-handling-tests), cleaned them up a little bit.

> some cases where the route failed to match given the parameters (which report as [ ERROR ] in the log)

Those were due to the fact that some tests were attempting to match on hostname (with no route). That feature was not supported on 0.14.1 so naturally it failed. The "fix" was [skipping those tests in 0.14.1](https://github.com/Kong/kong/commit/3774f77b00252490359b47e0487cc95cc9a16ada)

> For the "v0" behavior, I had a few mismatches (which report as [ FAILED ] in the log)

Yes, those show some discordance between 0.14.1 and v0. I put it on my [last commit](https://github.com/Kong/kong/commit/e82c31e0f6e74cf483bd46b69738fe635df00e77) on my temp branch.

My plan now is putting that last commit on this branch, and fixing the router so that all is green.

hishamhm:
> Those were due to the fact that some tests were attempting to match on hostname (with no route). That feature was not supported on 0.14.1 so naturally it failed. The "fix" was skipping those tests in 0.14.1

That's not true, matching by host has been supported (even by the old `apis` entity). You're mistaking it with matching by abitrary header, which is a newer addition.

I assume it's some kind of mismatch on how the tests are written which is making them all fail at once, but that should be fixable in the test suite itself I guess?

> Yes, those show some discordance between 0.14.1 and v0. I put it on my last commit on my temp branch. With that, bin/busted spec/01-unit/08-router_spec.lua -t slash on that branch.
> 
> My plan now is putting that last commit on this branch, and fixing the router so that all is green.

Sounds like a plan! :+1:

hishamhm:
@kikito I was thinking we should have probably generated every possible combination mechanically, ran them over 0.14.1 and `master`, extracted the results and built the `v0` and `v1` tables out of that, instead of trying to test and think through the cases and build the table by hand. Would have been faster and verified-correct. :)

I played a bit with a [script](https://gist.github.com/hishamhm/682da032c648a53124e12aeabb0e0cdd) here to get an idea of our current coverage of the current table. Given the inputs existing in the current fixtures (I renamed `srv` to `service` to avoid redundant combos), we have [164 cases checked and 256 cominations unchecked](https://gist.github.com/hishamhm/cb840247c17a5b54570455a79cc85b50) — granted, it generated a few redundant cases, and many of those combinations don't even match a route, but that should be mechanically checkable as well.

I think we're in generally good shape (no need to rework this whole expansion thing by now), but you may want to take a look to see if any of those unchecked expansions catches your eye as something worth double-checking.

kikito:
@hishamhm Cherry-picked the changes for full v0.4.1 compatibility and changed the router v0 branches to comply with it.

kikito:
> I assume it's some kind of mismatch on how the tests are written which is making them all fail at once, but that should be fixable in the test suite itself I guess?

You were completely right on this account. The problem was that the router in 0.14.x expected routes to come packaged in a different way, with the "host headers out". I made the necessary changes and was able to run the host-based tests as well as the others.

Long story short: I found one more change in the host-based tests, but this change was already necessary for the changes I made in the router anyway (two wrongs made a right here: I had incorrectly assumed I could change the host tests however I wanted). I just had to slightly retouch the git history so that it reflected reality.

kikito:
With regards to the Long List of Usecases.

First of all, great job! I should have thought about doing something like that.

Secondly, I think we can ignore all the ones who use `/rou` or `/roureq` instead of `/route` in the request/response paths, which (eyeballing) make about 1/3rd of the unchecked cases.

Explanation: it was I who introduced a couple tests which used  `/fee/bor` and instead just used `/fee`. They reproduced more clearly a usecase from one of our customers (whose path had no internal slash) who had detected a double slash in one test. By this I mean that the important bit was not in the "sort request", but in several "borders" ending in slashes and producing a double-slash. So I think it could have been a `/fee/bor`/`/service` request. I am not 100% sure about this though so I am willing to leave the short tests there just in case. But I am reasonably sure to say that we don't need to include these in the combinatory of your script. I realize this is a weirdly specific amount of certitude.

I have looked at the rest of the untested cases, and I believe they are covered well enough, when they are taken all into account (when one case isn't covered, others quite similar to it are). Let's keep automation in mind.

kikito:
Closing in favor of #5360

PR #5294, opened 2019-12-04
feat: CP/DP (Hybrid mode)

dndx:
This is a revised PR of the Hybrid demo done during the Kong Summit. Some notable changes since that demo:

* Added mTLS handshake support
* Added integration tests

TODO:

- [x] Add script to help user generate cluster certificates


kong/conf_loader.lua
```diff
@@ -235,6 +236,10 @@ local CONF_INFERENCES = {
 
   lua_ssl_verify_depth = { typ = "number" },
   lua_socket_pool_size = { typ = "number" },
+  role = { enum = { "proxy", "admin", "traditional", }, },
```
hbagdi: I don't want to bike-shed on this but are `admin` and `proxy` accurate?
They are already overloaded terms in Kong's terminology. `proxy` is okay but I'm not sure if that's what we want Kong to limit to, especially when we take the enterprise product into consideration as well.

I'm sure there was more discussion on this before that I might not be aware of.

kong/templates/nginx_kong.lua
```diff
@@ -56,6 +56,9 @@ lua_shared_dict kong_rate_limiting_counters 12m;
 > if database == "cassandra" then
 lua_shared_dict kong_cassandra      5m;
 > end
+> if role == "admin" then
+lua_shared_dict kong_clustering     5m;
+> end
```
hbagdi: If the role is proxy, then should other shms related to DBs be disabled? Like cassandra, db_cache_miss etc?

kong/api/routes/clustering.lua
```diff
@@ -0,0 +1,18 @@
...
+  ["/clustering/status"] = {
+    GET = function(self, db, helpers)
+      if kong.configuration.role ~= "admin" then
+        return kong.response.exit(400, {
+          message = "this endpoint is only available when Kong is " ..
+                    "configured to run as Control Plane for the cluster"
+        })
+      end
+
+      return kong.response.exit(200, clustering.get_status())
```
hbagdi: Should we have a proxy endpoint on Kong's Admin API for this endpoint as well?
Since Admin API is how most people are familiar with today and it is what will be used to configure Kong, it makes sense to have this endpoint to be present on the Admin API.

kong/api/routes/clustering.lua
```diff
@@ -0,0 +1,18 @@
...
+local kong = kong
+
+
+return {
+  ["/clustering/status"] = {
+    GET = function(self, db, helpers)
+      if kong.configuration.role ~= "admin" then
+        return kong.response.exit(400, {
+          message = "this endpoint is only available when Kong is " ..
+                    "configured to run as Control Plane for the cluster"
```
hbagdi: We should be consistent. If the role is `admin` then we should have this message should have that.

My personal opinion is that we should have role as `control-plane` or `controller` or some such as the `role`.

kong/api/routes/clustering.lua
```diff
@@ -0,0 +1,18 @@
+local clustering = require("kong.clustering")
+local kong = kong
+
+
+return {
+  ["/clustering/status"] = {
+    GET = function(self, db, helpers)
+      if kong.configuration.role ~= "admin" then
```
hbagdi: Another question: Is there an endpoint for such status on the data-plane?
Things I'd look during debugging is last sync, last ping, which CP node is the data-plane connected to right now.

kong/clustering.lua
```diff
@@ -0,0 +1,424 @@
...
+    queue.sem:post()
+
+    n = n + 1
+  end
+
+  ngx_log(ngx_DEBUG, "config pushed to ", n, " clients")
+end
+
+
+function _M.broadcast_config(config_table)
```
hbagdi: Question: Where is this function called from?

kong/clustering.lua
```diff
@@ -0,0 +1,424 @@
...
+  end
+
+  local sem = semaphore.new()
+  local queue = { sem = sem, }
+  clients[wb] = queue
+
+  local res
+  -- unconditionally send config update to new clients to
+  -- ensure they have latest version running
+  res, err = declarative.export_config()
```
hbagdi: Question for personal understanding: Why do we stored the config in as part of the queue? Why can't we simply have an event for it (dummy table or some such) and then load call export_config at that point?

Also, why don't we cache the current latest config at module level and then use that? There is a call to `export_config` in few places, I think it should be only one place where the config is actually swapped.

kong/templates/nginx_kong.lua
```diff
@@ -299,4 +302,25 @@ server {
...
+> if role == "admin" then
+server {
+    server_name kong_cluster_listener;
+> for i = 1, #cluster_listeners do
+    listen $(cluster_listeners[i].listener) ssl;
+> end
+
+    access_log off;
+
+    ssl_verify_client   optional_no_ca;
```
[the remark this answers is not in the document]
dndx: In an interest of "security by default" mindset, we choose to not allow insecure deployment where mTLS is not used.

kong/conf_loader.lua
dndx: Yes I believe this was used on the product design side. @thibaultcha thoughts?

kong/api/routes/clustering.lua
dndx: Not sure if I got this, this API is currently served though CP Admin API. Is there something I missed here?

kong/api/routes/clustering.lua
dndx: No, however the DP does have pretty detailed logging explaining how things are going. But if there are enough use case we may add another API later on the DP side.

kong/templates/nginx_kong.lua
dndx: It seems that the `cassandra` shdict is disabled with proxy mode already, since it uses `database = "off"`:

```
> if database == "cassandra" then
lua_shared_dict kong_cassandra      5m;
> end
```

kong/clustering.lua
dndx: Good catch, that was a leftover from early development, it has been removed :)

kong/clustering.lua
dndx: Good thought. Let me try to dig up the thought process, I think they are as follows:

1. We store the table object in order to not having to call `export_config` multiple times. In practice this is a very cheap operation, as the table itself is not copied in each queue, but all points to the same table instance.
2. Because of the same reason, caching the result of `export_config` at module level is not hugely beneficial, as the `export_config` is only called once for each new version anyway, no matter how much DPs are connected.

kong/templates/nginx_kong.lua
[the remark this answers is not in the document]
hbagdi: I read the doc for this directive wrong:
https://nginx.org/en/docs/http/ngx_http_ssl_module.html

Thank you for clarifying!

kong/templates/nginx_kong.lua
hbagdi: What about cache_miss shms though? I don't think they are used in DB-less mode.

kong/api/routes/clustering.lua
hbagdi: Right. CP Admin API has only one endpoint right now (this one) but for users who start moving towards this the new deployment model, it might help to have this endpoint also be present on the traditional Admin API of Kong.

kong/api/routes/clustering.lua
hbagdi: We should have a programmatic way of knowing if a data-plane is healthy or not.
The `/status` endpoint on DP nodes will return `200` no matter what.

kong/clustering.lua
hbagdi: That makes sense.
The issue I've in mind is that there is a call to `export_config()`  in the worker_event handler as well.
Should there be one place this is called and the reference to the table is kept in one place?
That makes it easier to reason about the flow and we can later on expose the info of what is the current config that is being served by a specific CP node.

kong/conf_loader.lua
guanlan: 'data_plane' and 'control_plane' or simply 'dp' and 'cp' sounds more straightforward.

kong/api/routes/clustering.lua
dndx: That is generally hard to tell on the DP side, since DP does not necessarily know the state of the config, only CP does.

That being said, if there are enough interest we can certainly add this endpoint to DP in the future.

kong/templates/nginx_kong.lua
dndx: That is a DB-less specific issue, let's fix it elsewhere in an interest to not overly complicate this PR.

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+
+  return true
+end
+
+
+local function send_ping(c)
+  local _, err = c:send_ping(declarative.get_current_hash())
+  if err then
+    ngx_log(ngx_ERR, "unable to ping control plane node: ", err)
+    -- return and let the main thread handle the error
```
bungle: this comment does not match with the reality. Yes it returns `nil`, but also the good case return `nil` below, and the main thread does not handle the error.

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+  local uri = "wss://" .. address .. "/v1/outlet?node_id=" ..
+              kong.node.get_id() .. "&node_hostname=" .. utils.get_hostname()
+  local res, err = c:connect(uri, { ssl_verify = true,
+                                    client_cert = CERT,
+                                    client_priv_key = CERT_KEY,
+                                    server_name = "kong_clustering",
+                                  }
+                            )
+  if not res then
+    local delay = 9 + math.random()
```
bungle: what is the meaning of `math.random` here?

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+    end
+  end)
+
+  while true do
+    local data, typ, err = c:recv_frame()
+    if err then
+      ngx.log(ngx.ERR, "error while receiving frame from control plane: ", err)
+      c:close()
+
+      local delay = 9 + math.random()
```
bungle: Again here, what does it mean `9 + [0.?]` why is this important? The `math.random` (and i guess math.random may return sub millisecond floating point, e.g. I got this: `7.8263692594256e-06`)

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+      assert(ngx.timer.at(delay, communicate, conf))
+      return
+    end
+
+    if typ == "binary" then
+      local msg = assert(cjson_decode(data))
```
bungle: Does this thread need to be `robust`? Above on error the timer and communication is restarted, here we just error.

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+
+      if msg.type == "reconfigure" then
+        local config_table = assert(msg.config_table)
```
bungle: Same thing here.

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+    return ngx_exit(444)
+  end
+
+  local node_id = ngx_var.arg_node_id
+  if not node_id then
+    ngx_exit(400)
+  end
+
+  local node_hostname = ngx_var.arg_node_hostname
+  local node_ip = ngx_var.remote_addr
```
bungle: I don't think there exists a `real_ip` configuration for this. E.g. a dataplane may be behind proxy or lb.

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+  queue.sem:post()
+
+  -- connection established
+  -- ping thread
+  ngx.thread.spawn(function()
+    while true do
+      local data, typ, err = wb:recv_frame()
+      if not data then
+        ngx_log(ngx_ERR, "did not receive ping frame from data plane: ", err)
+        -- return and let the main thread handle the error
```
bungle: where does that happen?

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+        return
+      end
+
+      assert(typ == "ping")
+      local _
+      _, err = wb:send_pong()
+      if err then
+        ngx_log(ngx_ERR, "failed to send PONG back to data plane: ", err)
+        -- return and let the main thread handle the error
```
bungle: same here, again both bad and good case return nil (as the only return argument).

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+
+local function init_mtls(conf)
+  local f, err = io_open(conf.cluster_cert, "r")
+  if not f then
+    return nil, "unable to open cluster cert file: " .. err
+  end
+
+  local cert
+  cert, err = f:read("*a")
+  if not cert then
```
bungle:
```suggestion
  if not cert then
    f:close()
```

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+  CERT_DIGEST = cert:digest("sha256")
+
+  f, err = io_open(conf.cluster_cert_key, "r")
+  if not f then
+    return nil, "unable to open cluster cert key file: " .. err
+  end
+
+  local key
+  key, err = f:read("*a")
+  if not key then
```
bungle:
```suggestion
  if not key then
    f:close()
```

kong/clustering.lua
```diff
@@ -0,0 +1,409 @@
...
+        local config, err = f:read("*a")
+        if not config then
+          ngx_log(ngx_ERR, "unable to read cached config file: ", err)
+        end
+
+        f:close()
+
+        if config then
+          ngx_log(ngx_INFO, "found cached copy of data-plane config, loading..")
+          config = cjson_decode(config)
```
bungle: Should we handle errors?

kong/cmd/hybrid.lua
```diff
@@ -0,0 +1,105 @@
...
+                  pl_path.abspath(cert_file),
+                  pl_path.abspath(key_file))
+    os.exit(0)
+  end
+
+  error("unknown command '" .. args.command .. "'")
+end
+
+local lapp = [[
+Usage: kong hybird COMMAND [OPTIONS]
```
bungle:
```suggestion
Usage: kong hybrid COMMAND [OPTIONS]
```

kong/pdk/private/phases.lua
```diff
@@ -7,20 +7,21 @@ local ngx_get_phase = ngx.get_phase
...
+  access            = 0x00000020,
+  balancer          = 0x00000040,
+  --content         = 0x00000100,
+  header_filter     = 0x00000200,
+  body_filter       = 0x00000400,
+  --timer           = 0x00001000,
+  log               = 0x00002000,
+  preread           = 0x00004000,
+  admin_api         = 0x10000000,
+  cluster_listener  = 0x20000000,
```
bungle: Perhaps we can make this a synonym to `content`?

kong/templates/kong_defaults.lua
```diff
@@ -14,6 +14,10 @@ proxy_listen = 0.0.0.0:8000 reuseport backlog=16384, 0.0.0.0:8443 http2 ssl reus
 stream_listen = off
 admin_listen = 127.0.0.1:8001 reuseport backlog=16384, 127.0.0.1:8444 http2 ssl reuseport backlog=16384
 status_listen = off
+cluster_listen = 0.0.0.0:8002
```
bungle: as it uses websockets, perhaps it is a good idea to add `so_keepalive` option.

kong/templates/nginx_kong.lua
```diff
@@ -320,4 +323,25 @@ server {
     }
 }
 > end
+
+> if role == "control_plane" then
+server {
```
bungle: I think it is missing e.g. `real_ip` as mentioned in other comment.

kong/templates/nginx_kong.lua
hishamhm: @hbagdi cache_miss shms are used in DB-less as well. Anything that uses `kong.cache:get` calls (core, plugins) will use both the positive and negative caches as normal.

kong/clustering.lua
dndx: The idea here is that the main thread would try to call `recv_frame` which will return the same error, therefore terminating. Since the ping thread already logged the error nothing will be lost. However I could make the ping thread terminate first when this happens. Committing a change.

kong/clustering.lua
dndx: This aims to evenly distribute reconnect requests so that the CP don't get overwhelmed by lots of client connecting at once.

p0pr0ck5: Could we consider increasing this jitter spread? Right now this is just +/- half a second. Spreading this among 5-10 seconds seems more reasonable.

kong/clustering.lua
dndx: The `lua-resty-websocket` library handles framing of incoming messages, if a frame was received and the content can not be decoded, then it must be a bug in the library or client code.

kong/clustering.lua
dndx: `config_table` missing means the CP sent a malformed message, which is a bug.

kong/clustering.lua
dndx: This won't work, as LB won't be able to terminate the mTLS connection between DP and CP in order to maintain E2E security, therefore no `XFF` header can not be added.

kong/clustering.lua
dndx: Good catch, this was changed to `return ngx_exit()` instead.

kong/clustering.lua
dndx: Added checks.

kong/templates/kong_defaults.lua
dndx: This is fine as we have application layer heartbeats for the WebSocket connections.

kong/templates/nginx_kong.lua
dndx: See my above comment, it won't work because of mTLS requirements here.

kong/clustering.lua
bungle: @dndx the question is, should the thing crash because of that or continue working? The message may get corrupted, timeout'd etc.?

dndx: Corrupted message means a bug in the code, in case of timeouts the code will never reach here as JSON decoding would not even happen in the first place.

dndx:
The rendered docs for this feature is at: https://deploy-preview-1670--kongdocs.netlify.com/2.0.x/hybrid-mode/ in case it makes testing it locally easier :)

kong/clustering.lua
dndx: How does this look: https://github.com/Kong/kong/pull/5294/commits/da771e710bc6171c05f40407c04214c599447683

p0pr0ck5: Seems very reasonable. In the future an exponential backoff approach might be more valuable but i think this is an excellent step as-is :+1:

kong/clustering.lua
```diff
@@ -0,0 +1,415 @@
...
```
gszr: Error handling in this call (as well as a in few of the other `cjson_decode` calls)?

.travis.yml
```diff
@@ -34,6 +34,7 @@ env:
     - KONG_TEST_PG_USER=postgres
     - JOBS=2
     - DOCKER_MACHINE_ARM64_NAME=travis-ci-kong-${TRAVIS_JOB_ID}
+    - OPENRESTY_PATCHES_BRANCH=chore/cosocket_mtls
```
gszr: (will be reverted once https://github.com/Kong/openresty-patches/pull/57 is merged)

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -0,0 +1,114 @@
...
+      helpers.get_db_utils(strategy, {
+        "routes",
+        "services",
+      }) -- runs migrations
+
+      assert(helpers.start_kong({
+        role = "control_plane",
+        cluster_cert = "spec/fixtures/kong_clustering.crt",
+        cluster_cert_key = "spec/fixtures/kong_clustering.key",
+        lua_ssl_trusted_certificate = "../spec/fixtures/kong_clustering.crt",
```
gszr: This (and others) `../` are the reason for test failures

kong/conf_loader.lua
```diff
@@ -560,6 +565,47 @@ local function check_and_infer(conf)
...
+    if conf.database == "off" then
+      errors[#errors + 1] = "in-memory storage can not be used when role = \"control_plane\""
+    end
+
+  elseif conf.role == "data_plane" then
+    if #conf.proxy_listen < 1 or pl_stringx.strip(conf.proxy_listen[1]) == "off" then
+      errors[#errors + 1] = "proxy_listen must be specified when role = \"data_plane\""
+    end
+
+    if conf.database ~= "off" then
```
bungle: Should we just force this `off` when role is `data_plane`, and document that with `data_plane` role the `database` configuration is not used?

kong/conf_loader.lua
```diff
@@ -560,6 +565,47 @@ local function check_and_infer(conf)
...
+      errors[#errors + 1] = "only in-memory storage can be used when role = \"data_plane\"\n" ..
+                            "Hint: set database = off in your kong.conf"
+    end
+  end
+
+  if conf.role == "control_plane" or conf.role == "data_plane" then
+    if not conf.cluster_cert or not conf.cluster_cert_key then
```
bungle: Is there any practical reason why `control-plane` couldn't just generate them when missing? Also do we really share same private key and public key with both control plane and dataplane? That feels rather messy? Can't we derive data-plane certs from control plane cert so that we never copy control plane private keys to dataplanes.

bungle:
when `data-planes` are restarted, this is emitted to `control-plane` logs:
```
2019/12/20 17:26:21 [debug] 78863#0: *1026 [lua] clustering.lua:241: sent PONG packet to control plane
2019/12/20 17:26:22 [error] 78863#0: *1026 [lua] clustering.lua:229: did not receive ping frame from data plane: failed to receive the first 2 bytes: closed, client: 127.0.0.1, server: kong_cluster_listener, request: "GET /v1/outlet?node_id=eacd44b6-da7f-480f-bbdb-b2839e8b89ce&node_hostname=pro.lan HTTP/1.1", host: "127.0.0.1:8005"
2019/12/20 17:26:22 [error] 78863#0: *1026 lua user thread aborted: runtime error: /usr/local/openresty/lualib/resty/core/exit.lua:34: bad argument #2 to 'ngx_http_lua_ffi_exit' (cannot convert 'nil' to 'int')
stack traceback:
coroutine 0:
	[C]: in function 'ngx_http_lua_ffi_exit'
	/usr/local/openresty/lualib/resty/core/exit.lua:34: in function </usr/local/openresty/lualib/resty/core/exit.lua:26>
coroutine 1:
	[C]: in function 'co_yield'
	/usr/local/openresty/lualib/ngx/semaphore.lua:160: in function 'wait'
	./kong/clustering.lua:259: in function 'serve_cluster_listener'
	content_by_lua(nginx-kong.conf:108):2: in main chunk, client: 127.0.0.1, server: kong_cluster_listener, request: "GET /v1/outlet?node_id=eacd44b6-da7f-480f-bbdb-b2839e8b89ce&node_hostname=pro.lan HTTP/1.1", host: "127.0.0.1:8005"
```

I think it would be better to gracefully shutdown that `ws` connection.

kong/db/declarative/init.lua
```diff
@@ -282,6 +282,56 @@ function declarative.export_from_db(fd)
 end
 
 
+function declarative.export_config()
```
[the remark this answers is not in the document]
dndx: Yeah, let's address this in rc2 as it is not user visible.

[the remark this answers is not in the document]
hishamhm: Yep, I was going to make the same comment. Not just different names, but this needs to be refactored to avoid the actual code duplication. But agreed, it's minor and not a blocker.

kong/init.lua
```diff
@@ -382,6 +385,9 @@ function Kong.init()
     end
   end
 
+  local clustering = require "kong.clustering"
+  clustering.init(config)
```
hishamhm: Why is this being initialized twice in this function?

kong/init.lua
```diff
@@ -360,6 +360,9 @@ function Kong.init()
   assert(db:connect())
   assert(db.plugins:check_db_against_config(config.loaded_plugins))
 
+  local clustering = require "kong.clustering"
```
hishamhm: Why is this being required here and not at the top of the module? As far as I can tell, requiring `kong.clustering` is side-effect free.

spec/02-integration/02-cmd/12-hybrid_spec.lua
```diff
@@ -0,0 +1,103 @@
...
+
+      local ok, _, stdout = helpers.kong_exec("hybrid gen_cert " .. cert .. " " .. key)
+      assert.truthy(ok)
+      assert.matches("Successfully generated certificate/key pairs, they have been written to: ", stdout, nil, true)
+
+      _, _, stdout = helpers.execute("ls -l " .. key)
+      assert.matches("-rw-------", stdout, nil, true)
+
+      _, _, stdout = helpers.execute("ls -l " .. cert)
+      assert.matches("-rw-r--r--", stdout, nil, true)
```
hishamhm: these will fail if someone tries to run the testsuite with a tighter `umask` — but oh well, if somebody complains, we already know the answer ;)

spec/fixtures/custom_nginx.template
```diff
@@ -57,6 +57,7 @@ http {
     lua_shared_dict kong_core_db_cache_2     ${{MEM_CACHE_SIZE}};
     lua_shared_dict kong_db_cache_2     ${{MEM_CACHE_SIZE}};
 > end
+
```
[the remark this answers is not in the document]
hishamhm: @gszr was this resolved by the other additions elsewhere in this file? looks like everything is replicated below — can we mark this comment as resolved?

kong/init.lua
dndx: I think this was a rebase error. Looking.

kong/init.lua
dndx: I think https://github.com/Kong/kong/pull/5294/files#diff-f984694673247937d61992dca2dfa663R21 is the problem.

hishamhm: ahh, didn't catch that sneaking among the `requires`. so _that_ line needs to be deferred, instead?

hishamhm:
@dndx Just noticed: there is no documentation for the new config entries in `kong.conf.default`?

kong/init.lua
dndx: Yes, I smell a lazy loading of `declarative.new_config(kong.configuration)` could solve the issue. I made a card for that.

PR #5302, opened 2019-12-05
fix(cache) same invalidation sent multiple times

bungle:
### Summary

A single `patch` on e.g. `service` entity may cause four cluster events fired, out of which all four are the same. This was doubled because the #5114 introduced the core cache / plugins cache split.

This PR also adds a guard on `patches`: when the entity and old entity cache keys are the same, only one invalidation event is sent.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -188,20 +188,24 @@ local function register_events()
     -- caching key
 
     local cache_key = db[data.schema.name]:cache_key(data.entity)
+    local cache_obj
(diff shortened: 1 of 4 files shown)


hutchic:
This is a general lua testing question -- is it possible to mock `cache:invalidate` and then assert that it was only called X number of times?

bungle:
@hutchic it is, I think some of our tests do actually that. Let me try to add that.

bungle:
update, I'll try to add test on monday, a bit more complex than I otiginally thought.

PR #5305, opened 2019-12-05
test(plugins) add capability to test externally hosted but bundled Kong plugins

hutchic:
All bundled Kong plugins are inferred from the rockspec as always being `kong-` and `~`. From there we determine git tag from the result of `luarocks show`

Once we have the plugin repository and git tag we clone the plugin source code, copy the necessary fixtures and then run the  plugin spec tests

- [x] Azure spec folder is named specs
- [x] Azure tests are currently failing
- [x] Prometheus tests are currently failing
- [x] Zipkin repository tags are prefixed with a `v`

### DIFF
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -23,14 +23,13 @@ if [ "$TEST_SUITE" == "dbless" ]; then
                      spec/02-integration/04-admin_api/15-off_spec.lua
 fi
 if [ "$TEST_SUITE" == "plugins" ]; then
-    eval "$TEST_CMD" spec/03-plugins/
-fi
-if [ "$TEST_SUITE" == "pdk" ]; then
-    TEST_NGINX_RANDOMIZE=1 prove -I. -j$JOBS -r t/01-pdk
-fi
-if [ "$TEST_SUITE" == "external_plugins" ]; then
     set +e
     rm -f .failed
+
+    for p in spec/03-plugins/*; do
+        $TEST_CMD $p || echo "* $p" >> .failed
+    done
+
     cat kong-*.rockspec | grep kong- | grep -v sidecar | grep -v zipkin | grep "~" | while read line ; do
         REPOSITORY=`echo $line | sed "s/\"/ /g" | awk -F" " '{print $1}'`
         VERSION=`luarocks show $REPOSITORY | grep $REPOSITORY | head -1 | awk -F" " '{print $2}' | cut -f1 -d"-"`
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -43,6 +43,7 @@ env:
     - KONG_TEST_DATABASE=postgres TEST_SUITE=plugins
     - KONG_TEST_DATABASE=cassandra CASSANDRA=2.2 TEST_SUITE=plugins
     - KONG_TEST_DATABASE=cassandra CASSANDRA=3.9 TEST_SUITE=plugins
+    - KONG_TEST_DATABASE=cassandra CASSANDRA=3.9 TEST_SUITE=external_plugins
```
gszr: Missing jobs for Postgres and C* 2.2?

.ci/run_tests.sh
```diff
@@ -28,3 +28,23 @@ fi
...
+    set +e
+    rm -f .failed
+    cat kong-*.rockspec | grep kong- | grep -v sidecar | grep -v zipkin | grep "~" | while read line ; do
+        REPOSITORY=`echo $line | sed "s/\"/ /g" | awk -F" " '{print $1}'`
+        VERSION=`luarocks show $REPOSITORY | grep $REPOSITORY | head -1 | awk -F" " '{print $2}' | cut -f1 -d"-"`
+        REPOSITORY=`echo $REPOSITORY | sed -e 's/kong-prometheus-plugin/kong-plugin-prometheus/g'`
+        REPOSITORY=`echo $REPOSITORY | sed -e 's/kong-proxy-cache-plugin/kong-plugin-proxy-cache/g'`
+        echo $REPOSITORY
+        echo $VERSION
+        git clone https://github.com/Kong/$REPOSITORY.git --branch $VERSION --single-branch /tmp/test-$REPOSITORY
```
gszr: Shouldn't we be testing against master, rather than the latest released version?

.travis.yml
hutchic: maybe? Some of the plugins don't exercise a datastore at all some of them are hardcoded to cassandra.

No harm in adding 2.2 I suppose

.ci/run_tests.sh
hutchic: Not in this repository no. We want to know that PR's and release builds of Kong have pinned their plugins at versions which are compatible.

The plugin CI is what should be testing the master of the plugin.

.travis.yml
gszr: All plugins exercise a datastore (outside of dbless, of course) at least for the plugin configuration. Granted, some tests are often written specifically for a given storage strategy, but that's not a universal truth - it's handled in the test suite itself.

hishamhm: I think this shouldn't be a separate job but instead be included in the `plugins` job alongside the in-repo plugins. This way they will be tested will all DB combinations.

hutchic: I partly separated them for CI parallelization speed but fine with it being merged into `plugins` too

Jenkinsfile
```diff
@@ -86,6 +86,25 @@ pipeline {
                         }
                     }
                 }
+                stage('external plugins') {
```
gszr: Can we get rid of this new stage, as we are now using the same job as for bundled plugins?

.ci/run_tests.sh
```diff
@@ -23,7 +30,52 @@ if [ "$TEST_SUITE" == "dbless" ]; then
...
+
+        git clone https://github.com/Kong/$REPOSITORY.git --branch $VERSION --single-branch /tmp/test-$REPOSITORY
+        cp -R /tmp/test-$REPOSITORY/spec/fixtures/* spec/fixtures/ || true
+        pushd /tmp/test-$REPOSITORY
+        luarocks make
+        popd
+
+        $TEST_CMD /tmp/test-$REPOSITORY/spec/ || echo "* $REPOSITORY" >> .failed
+
+        rm -rf /tmp/test-$REPOSITORY
```
hutchic:
```suggestion
```

Initially this existed because I recycled the directory for the next iteration. Now that we're not (which is better) let's not nuke the plugin directory as it's handy for when debugging locally?

hishamhm: works for me!

PR #5316, opened 2019-12-10
tests(*) grpc logging plugins

gszr:
### Summary

Add tests for grpc on Statsd, Syslog, and Loggly.


spec/03-plugins/07-loggly/01-log_spec.lua
```diff
@@ -119,6 +183,31 @@ for _, strategy in helpers.each_strategy() do
...
+    local function run_grpc(host)
+      local thread   = assert(helpers.udp_server(UDP_PORT))
+
+      local ok, resp = proxy_client_grpc({
+        service = "hello.HelloService.SayHello",
+        body = {
+          greeting = "world!"
+        },
+        opts = {
+          ["-authority"] = ("%s"):format(host),
```
javierguerragiraldez: how's this different from `tostring(host)`?

PR #5322, opened 2019-12-11
fix(acl) get acl by group fails when more consumers share the same group

### LINKED ISSUE #5280: Get ACL by group fails when more consumers share the same group
### Summary

Since this commit (5f0ab5d106a6b246bc0f7fc24296291c03e4596e) the retrieving of an ACL by the group name fails for all consumers but one when there is more than one consumer with the same group. This happens because the consumer_id is not used in the db query and only the first of the acl with that group name is returned (always the same acl for all consumers)

### Steps To Reproduce

1. `http POST /consumers username=alice` [200]
2. `http POST /consumers username=bob` [200]
3. `http POST /consumers/alice/acls group=foo` [200]
4. `http POST /consumers/bob/acls group=foo` [200]
5. `http GET /consumers/alice/acls/foo` [200]
5. `http GET /consumers/bob/acls/foo` **[404]**

### Additional Details & Logs

I have this issue with kong 1.0.3 but looking at the code I think it is present also in master.
If you want I can submit a pull request with the updated spec test file showing the issue but I don't know how to fix it, I think `kong.api.endpoints` should be extended. I can try if someone can indicate me the right direction.

bungle:
### Summary

@albertored reported issue here #5280.

This PR removes `endpoint_key` from `acls` dao, as it is not `unique`,
and using non-unique endpoint keys leads to strange results.

And of course it fixes the issue reported:

```
http POST :8001/consumers username=alice
http POST :8001/consumers username=bob
http POST :8001/consumers/alice/acls group=foo
http POST :8001/consumers/bob/acls group=foo
http GET :8001/consumers/alice/acls/foo
http GET :8001/consumers/bob/acls/foo
```

All returning `200`/`201` as expected.

### Issues Resolved

Fix #5280

### DIFF
--- a/spec/03-plugins/18-acl/01-api_spec.lua
+++ b/spec/03-plugins/18-acl/01-api_spec.lua
@@ -1,3 +1,4 @@
+local utils   = require "kong.tools.utils"
 local cjson   = require "cjson"
 local helpers = require "spec.helpers"
 
@@ -46,21 +47,21 @@ for _, strategy in helpers.each_strategy() do
 
       describe("POST", function()
         it("creates an ACL association", function()
-          local res = assert(admin_client:post("/consumers/bob/acls", {
+          local res = admin_client:post("/consumers/bob/acls", {
             body    = {
               group = "admin"
             },
             headers = {
               ["Content-Type"] = "application/json"
             }
-          }))
+          })
           local body = assert.res_status(201, res)
           local json = cjson.decode(body)
           assert.equal(consumer.id, json.consumer.id)
           assert.equal("admin", json.group)
         end)
         it("creates an ACL association with tags", function()
-          local res = assert(admin_client:send {
+          local res = admin_client:send {
             method  = "POST",
             path    = "/consumers/bob/acls/",
             body    = {
@@ -70,7 +71,7 @@ for _, strategy in helpers.each_strategy() do
             headers = {
               ["Content-Type"] = "application/json"
             }
-          })
+          }
           local body = assert.res_status(201, res)
           local json = cjson.decode(body)
           assert.equal(consumer.id, json.consumer.id)
@@ -79,12 +80,12 @@ for _, strategy in helpers.each_strategy() do
         end)
         describe("errors", function()
           it("returns bad request", function()
-            local res = assert(admin_client:post("/consumers/bob/acls", {
+            local res = admin_client:post("/consumers/bob/acls", {
               body    = {},
               headers = {
                 ["Content-Type"] = "application/json"
               }
-            }))
+            })
             local body = assert.res_status(400, res)
             local json = cjson.decode(body)
             assert.same({ group = "required field missing" }, json.fields)
@@ -99,7 +100,7 @@ for _, strategy in helpers.each_strategy() do
         it("retrieves the first page", function()
           bp.acls:insert_n(3, { consumer = { id = consumer.id } })
 
-          local res = assert(admin_client:get("/consumers/bob/acls"))
+          local res = admin_client:get("/consumers/bob/acls")
           local body = assert.res_status(200, res)
           local json = cjson.decode(body)
           assert.is_table(json.data)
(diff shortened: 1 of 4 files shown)


spec/03-plugins/18-acl/01-api_spec.lua
```diff
@@ -152,10 +153,114 @@ for _, strategy in helpers.each_strategy() do
...
+          local res  = assert(admin_client:post("/consumers", {
+            body = {
+              username = "anna",
+            },
+            headers = {
+              ["Content-Type"] = "application/json",
+            },
+          }))
+          local body = assert.res_status(201, res)
+          local _    = cjson.decode(body)
```
kikito: I think using the following assertion is more explicit, and produces a slightly better error message if something goes wrong:

```suggestion
          assert.response(res).has.jsonbody()
```

The assertion also returns the decoded json table, so it could be used to replace the calls to `cjson.decode` where they are assigned to a named variable too, if you so desired. This is an aesthetic choice and I don't particularly care about that.

spec/03-plugins/18-acl/01-api_spec.lua
```diff
@@ -152,10 +153,114 @@ for _, strategy in helpers.each_strategy() do
...
+          local res  = assert(admin_client:post("/consumers", {
+            body = {
+              username = "jack",
+            },
+            headers = {
+              ["Content-Type"] = "application/json",
+            },
+          }))
+          local body = assert.res_status(201, res)
+          local _    = cjson.decode(body)
```
kikito:
```suggestion
          assert.response(res).has.jsonbody()
```

spec/03-plugins/18-acl/01-api_spec.lua
```diff
@@ -152,10 +153,114 @@ for _, strategy in helpers.each_strategy() do
...
+          local res  = assert(admin_client:delete("/consumers/anna"))
+          local _    = assert.res_status(204, res)
+
+          local res  = assert(admin_client:delete("/consumers/jack"))
+          local _    = assert.res_status(204, res)
+        end)
+        it("retrieves right ACL by group when multiple consumers share the same group name created with PUT", function()
+          local res  = assert(admin_client:put("/consumers/anna"))
+          local body = assert.res_status(200, res)
+          local _    = cjson.decode(body)
```
kikito:
```suggestion
          assert.response(res).has.jsonbody()
```

spec/03-plugins/18-acl/01-api_spec.lua
```diff
@@ -152,10 +153,114 @@ for _, strategy in helpers.each_strategy() do
...
+
+          local res  = assert(admin_client:put("/consumers/jack"))
+          local body = assert.res_status(200, res)
+          local _    = cjson.decode(body)
```
kikito:
```suggestion
          assert.response(res).has.jsonbody()
```

spec/03-plugins/18-acl/01-api_spec.lua
```diff
@@ -152,10 +153,114 @@ for _, strategy in helpers.each_strategy() do
...
+
+          assert.same(ag, ag2)
+          assert.same(jg, jg2)
+          assert.not_same(jg, ag)
+          assert.not_same(jg2, ag2)
+          assert.not_same(jg, ag2)
+          assert.not_same(jg2, ag)
+
+          local res  = assert(admin_client:delete("/consumers/anna"))
+          local _    = assert.res_status(204, res)
```
kikito: I think assigning this to the unnamed variable serves no purpose.

```suggestion
          assert.res_status(204, res)
```

spec/03-plugins/18-acl/01-api_spec.lua
```diff
@@ -152,10 +153,114 @@ for _, strategy in helpers.each_strategy() do
...
+
+          local res  = assert(admin_client:delete("/consumers/jack"))
+          local _    = assert.res_status(204, res)
```
kikito:
```suggestion
          assert.res_status(204, res)
```

spec/03-plugins/18-acl/01-api_spec.lua
```diff
@@ -152,10 +153,114 @@ for _, strategy in helpers.each_strategy() do
           res = assert(admin_client:get("/consumers/alice/acls/" .. acl.group))
           assert.res_status(404, res)
         end)
+        it("retrieves right ACL by group when multiple consumers share the same group name created with POST", function()
+          local res  = assert(admin_client:post("/consumers", {
```
kikito: All the "shortcut request methods" (`admin_client:post`, `admin_client:delete`, `admin_client:get`, etc) already do an `assert`, so wrapping them with an `assert` is not really needed.

PR #5325, opened 2019-12-11
feat(balancer) async change upstreams

locao:
This change adds the option to change upstreams asynchronously, i.e. upstreams will not be updated on every call to Admin API, but rather when a timer expires. This improves the performance when Kong is configured with a large number of upstreams and/or when upstreams are frequently changed.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -998,11 +998,14 @@
 # TUNING & BEHAVIOR
 #------------------------------------------------------------------------------
 
-#router_consistency = strict     # Defines whether this node should rebuild its
-                                 # router synchronously or asynchronously (the
-                                 # router is rebuilt every time a Route or a
-                                 # Service is updated via the Admin API or
-                                 # loading a declarative configuration file).
+#worker_consistency = strict
+                                 # Defines whether this node should rebuild its
+                                 # state synchronously or asynchronously (the
+                                 # balancers and the router are rebuilt on
+                                 # updates that affects them, e.g. updates to
+                                 # Routes, Services or Upstreams, via the Admin
+                                 # API or loading a declarative configuration
+                                 # file).
                                  #
                                  # Accepted values are:
                                  #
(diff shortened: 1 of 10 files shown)


kong/runloop/balancer.lua
```diff
@@ -533,42 +559,39 @@ local function check_target_history(upstream, balancer)
...
     end
-    return upstreams_dict
-  end
-  _load_upstreams_dict_into_memory = load_upstreams_dict_into_memory
 
+    upstreams_dict[up.name] = up.id
+  end
+  return upstreams_dict
+end
+_load_upstreams_dict_into_memory = load_upstreams_dict_into_memory
```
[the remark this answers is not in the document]
locao: I think we should keep it as it is used for unit tests.

kong.conf.default
```diff
@@ -1017,14 +1020,16 @@
...
                                  # checked with a background job. When a change
-                                 # is detected, a new router will be built. By
-                                 # default we check for changes every second.
-                                 # Raising this value will decrease the load on
-                                 # database servers and result in less jitter
-                                 # in proxy latency, with downside of longer
-                                 # converge time for router updates.
+                                 # is detected, a new router or balancer will be
+                                 # built, as needed. By default we check for
+                                 # changes every second. Raising this value will
```
hishamhm: The text says "by default we check for changes every second" but the config option itself here says `5` (I also note this was a change from the previous behavior — is this deliberate? Also, `kong_defaults.lua` below still says `1`, so this file is lying right now).

I think the text should not contain this sentence about the default, because the [generated documentation page](https://docs.konghq.com/2.0.x/configuration/#router_update_frequency) builds the "Default:" line out of the entry itself.

kong/templates/kong_defaults.lua
```diff
@@ -120,8 +120,8 @@ dns_not_found_ttl = 30
 dns_error_ttl = 1
 dns_no_sync = off
 
-router_consistency = strict
-router_update_frequency = 1
+worker_consistency = strict
+worker_state_update_frequency = 1
```
hishamhm: if the new default should be 5, then it needs to be updated here as well

hishamhm:
This has had a few rounds of review from the team already, from my run through it the only thing that jumped to my eyes was the default value for the update parameter, but apart from that LGTM

PR #5329, opened 2019-12-12
feat(rate-limiting) add support for Retry-After header on 429 response

bungle:
### Summary

This have been previously proposed with #3451 and #2735. This PR tries only to add `Retry-After` header, so it is not as complete as #3451, but it may be easier to get these in as parts.

Thank you @grahamar and @ioggstream for all the previous efforts.

### Details

The `Retry-After` is in `seconds`. If there is more than one `Limit` configured and more than one limit has exceeded the `Retry-After` will contain the biggest number (in seconds). E.g. if minute limit is exceeded and hour limit is exceeded it probably means that `Retry-After` will tell you when the hour limit is released assuming that the minute limit will be released earlier.

The PR is also turning on most of the `unit test` suite on `rate-limiting`.

### DIFF
--- a/spec/03-plugins/23-rate-limiting/04-access_spec.lua
+++ b/spec/03-plugins/23-rate-limiting/04-access_spec.lua
@@ -1,5 +1,4 @@
 local helpers        = require "spec.helpers"
-local timestamp      = require "kong.tools.timestamp"
 local cjson          = require "cjson"
 
 
@@ -11,17 +10,6 @@ local REDIS_DATABASE = 1
 
 local fmt = string.format
 local proxy_client = helpers.proxy_client
-
-
-local function wait(second_offset)
-  -- If the minute elapses in the middle of the test, then the test will
-  -- fail. So we give it this test 30 seconds to execute, and if the second
-  -- of the current minute is > 30, then we wait till the new minute kicks in
-  local current_second = timestamp.get_timetable().sec
-  if current_second > (second_offset or 0) then
-    ngx.sleep(60 - current_second)
-  end
-end
 
 
 local function GET(url, opts, res_status)
@@ -73,7 +61,7 @@ end
 
 for _, strategy in helpers.each_strategy() do
   for _, policy in ipairs({ "local", "cluster", "redis" }) do
-    describe(fmt("#flaky Plugin: rate-limiting (access) with policy: %s [#%s]", policy, strategy), function()
+    describe(fmt("Plugin: rate-limiting (access) with policy: %s [#%s]", policy, strategy), function()
       local bp
       local db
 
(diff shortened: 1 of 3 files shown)


spec/03-plugins/23-rate-limiting/04-access_spec.lua
```diff
@@ -13,17 +12,6 @@ local fmt = string.format
...
 
-local function wait(second_offset)
-  -- If the minute elapses in the middle of the test, then the test will
-  -- fail. So we give it this test 30 seconds to execute, and if the second
-  -- of the current minute is > 30, then we wait till the new minute kicks in
-  local current_second = timestamp.get_timetable().sec
-  if current_second > (second_offset or 0) then
-    ngx.sleep(60 - current_second)
-  end
-end
```
hishamhm: @bungle I know that removing this makes the tests run much faster, but then the suite assumes that you're lucky and not starting to run a test at, for example, second :59 of the minute — if the minute flips while the tests are running, they will fail. This exists to control flakiness because these tests depend on the real clock.

Granted, `wait(3)` down there is super conservative (it's giving _each test_ 57 seconds to run), but I believe something like `wait(55)` would make sense (i.e. wait 5 seconds for the turn of the minute if test started after second :55) — looking at Travis, each individual test seems to run under a second.

Given that this whole file takes around 30 seconds to run on Travis, without this check, it becomes a bit of a coin toss whether some test will end up trying to run during the turn of a minute.

bungle: The next pr adds better fix for that:
https://github.com/Kong/kong/pull/5335/commits/6a5a21d5d04a2050dfbb290486969847fb0fa119

I have seen almost 0 flaky errors.

hishamhm: @bungle I just tested that `it_with_retry` function locally and forced both failures (with `assert.same(true, false)` and errors (with `error()`) and it does not work as expected — that wrapper is doing nothing (if it did, it would be looping forever in case of a test failure).

You have seen almost 0 flaky errors because those rate-limiting errors are very occasional (I generally have to keep them running in a loop locally to see them happen), they only trigger when you hit the turn of the minute. Last time I was working on the rate-limiting tests, I remember I managed to reproduce those problems by forcing my clock locally. The only robust solution would be to mock the clock entirely, which would be pretty annoying to do from an integration test, so we to try to control the real time the test starts, which is what `wait` does.

In short, I think we need to keep this `wait()` function here, but change `wait(3)` to something like `wait(55)`.

bungle: I personally changed one test and it immediately picked it. So I don't really know what is going on.

bungle: I just tries with `error` and it worked as expected. Are you sure about it?

PR #5330, opened 2019-12-12
Fix: #233. Add generic X-RateLimit-* and X-RateLimit-Reset too.

### LINKED ISSUE #233: [ratelimiting] X-RateLimit-Reset

ioggstream:
### Summary

Return the lowest limit header as X-RateLimit-Remaining and
the associate X-RateLimit-Reset.

This implements this I-D  https://tools.ietf.org/html/draft-polli-ratelimit-headers-01
which has been recently discussed at [IETF106](http://bit.ly/2qWhwwD)

### Full changelog

* Add X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset using the lower values between the configured ones

### Issues resolved

Fix #233

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -1,4 +1,5 @@
 -- Copyright (C) Kong Inc.
+local timestamp = require "kong.tools.timestamp"
 local policies = require "kong.plugins.rate-limiting.policies"
 
 
@@ -6,14 +7,18 @@ local kong = kong
 local ngx = ngx
 local max = math.max
 local time = ngx.time
+local floor = math.floor
 local pairs = pairs
 local tostring = tostring
 local timer_at = ngx.timer.at
 
 
 local EMPTY = {}
+local EXPIRATIONS = policies.EXPIRATIONS
 local RATELIMIT_LIMIT = "X-RateLimit-Limit"
 local RATELIMIT_REMAINING = "X-RateLimit-Remaining"
+local RETRY_AFTER = "Retry-After"
+
 
 local RateLimitingHandler = {}
 
@@ -107,15 +112,28 @@ function RateLimitingHandler:access(conf)
 
   if usage then
     -- Adding headers
+    local reset
     if not conf.hide_client_headers then
       local headers = {}
+      local timestamps
       for k, v in pairs(usage) do
         if stop == nil or stop == k then
           v.remaining = v.remaining - 1
         end
 
+        local remaining = max(0, v.remaining)
+
         headers[RATELIMIT_LIMIT .. "-" .. k] = v.limit
-        headers[RATELIMIT_REMAINING .. "-" .. k] = max(0, v.remaining)
+        headers[RATELIMIT_REMAINING .. "-" .. k] = remaining
+
+        if stop and remaining == 0 then
+          if not timestamps then
+            timestamps = timestamp.get_timestamps(current_timestamp)
+          end
+
+          reset = max(reset or 0, EXPIRATIONS[k] - floor((current_timestamp -
+                                                          timestamps[k]) / 1000))
+        end
       end
 
       kong.ctx.plugin.headers = headers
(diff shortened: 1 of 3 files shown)


kong/plugins/rate-limiting/handler.lua
```diff
@@ -105,25 +111,61 @@ function RateLimitingHandler:access(conf)
...
+           ) then
+          lowest = k
+        end
       end
 
+      -- Set the normalized ratelimit values.
+      headers[RATELIMIT_LIMIT] = headers[RATELIMIT_LIMIT .. "-" .. lowest]
+      headers[RATELIMIT_REMAINING] = headers[RATELIMIT_REMAINING .. "-" .. lowest]
+      -- Set the expected reset time.
+      headers[RATELIMIT_RESET] = EXPIRATIONS[lowest] - (current_timestamp/1000 % EXPIRATIONS[lowest])
```
bungle: According to:
https://tools.ietf.org/html/draft-polli-ratelimit-headers-01

> If a response contains both the "Retry-After" and the "RateLimit- Reset" header fields, the value of "RateLimit-Reset" MUST be  consistent with the one of "Retry-After".

So I don't think it is a good idea to have `RateLimit-Reset` or `X-RateLimit-Reset` being something else than `Retry-After` or inconsistent with it.

But having specific:
`X-RateLimit-Reset-Minute` (etc.) could be an option. What do you think?

I know that the `polli` rfc is just a draft but that is the best we have. Everything else is proprietary, except perhaps `Retry-After` on `429`. And yes, we should drop `X-` prefix at some point (not now).

kong/plugins/rate-limiting/handler.lua
```diff
@@ -105,25 +111,61 @@ function RateLimitingHandler:access(conf)
...
+        if (lowest == nil
+            or usage[lowest].remaining > remaining
+            or (usage[lowest].remaining == remaining and EXPIRATIONS[k] > EXPIRATIONS[lowest])
+           ) then
+          lowest = k
+        end
       end
 
+      -- Set the normalized ratelimit values.
+      headers[RATELIMIT_LIMIT] = headers[RATELIMIT_LIMIT .. "-" .. lowest]
```
bungle: Perhaps the generic headers should follow this (as they are new, without `X-` prefix):
https://tools.ietf.org/html/draft-polli-ratelimit-headers-01#section-6.2.4

ioggstream: @bungle yes, but we are allowed to remove the `X-` only after the draft becomes RFC.

bungle: @ioggstream, AFAIK this:
https://tools.ietf.org/html/rfc6648

has deprecated the use of `X-` in 2012?

kong/plugins/rate-limiting/handler.lua
ioggstream: IIUC in this implementation `RateLimit-Reset` and `Retry-After` should match,
but let me know if I'm wrong.

> having specific `x-ratelimit-reset-minute` ...
Header suffixes (eg `-minute`) require maps/code to interpret suffixes (Eg.  `policies.EXPIRATIONS`),
so clients  a great fan.

Can you make an example of how you would use `X-RateLimit-Reset-Minute` & co?

Thanks for your support! :D

kong/plugins/rate-limiting/handler.lua
ioggstream: Yes, but using `DraftPolliV1-RateLimit-Limit` was weird :)))

bungle: Perhaps we can take a risk and just use `RateLimit-Limit`, if the RFC changes it, we should change as well. But still keeping these old `X-` headers that we have currently. We can document that those generic headers follow the RFC and if RFC is changed, we will change too.

ioggstream: Seems good. Are you interested in contributing to the spec? In case write me at robipolli@gmail.com

bungle:
Hi,

I just opened #5335 with the `generic X-RateLimit-* and X-RateLimit-Reset`. I don't want to take anything from you, but this has been stuck so long, that perhaps it will get easier in if I try to do it as I can always communicate about it internally easier than you. Please grab a t-shirt: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt, you deserve it.  You can refer this PR! Thank you!

bungle:
the #5335 is now merged, thus closing this.

PR #5332, opened 2019-12-12
feat: enable `reuseport` option for Nginx `listen` directive by default

dndx:
and allow specifying `backlog=N` options in listener flags

This enables the `reuseport` flag by default, and use a higher `backlog` to help with high traffic servers.

It should have no impact on existing use case, only that server with lots of worker now sees more evenly distributed load and connection across workers. it also reduces issue such as https://discuss.konghq.com/t/concurrent-websocket-connections/4854 from occurring.

Note on the `backlog` value. It is set to a fairly high value `16384` to help handle burst of TCP connections better. The `backlog` is by default set to unlimited on BSD systems and `511` on Linux, which is on the lower end and Kernel can easily handle more connections than that.

However, note that from the man page of `listen`:

> If the backlog argument is greater than the value in
> /proc/sys/net/core/somaxconn, then it is silently truncated to that
> value; the default value in this file is 128.  In kernels before
> 2.4.25, this limit was a hard coded value, SOMAXCONN, with the value
> 128.

Which means that it is also limited by the `somaxconn` Kernel param on Linux. `somaxconn` is set fairly low by default, and we document in `kong.conf` to make sure people is aware the necessity of bumping `somaxconn` at the same time in order for the config to take effect.

As a reference, in Kubernetes Ingress Controller, they always fetch the current value of `somaxconn` and use that as the `backlog` argument, but for us that is not necessary. The Kernel will handle it.

Note that raising `backlog` would not cause any increase of memory usage for normal users, as the TCP connection backlogs are not pre-allocated but rather stored in a dynamically growing/shrinking linked list on demand (https://elixir.bootlin.com/linux/latest/source/include/net/sock.h#L924, https://elixir.bootlin.com/linux/latest/source/include/net/sock.h#L897).

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -122,7 +122,7 @@
 # NGINX
 #------------------------------------------------------------------------------
 
-#proxy_listen = 0.0.0.0:8000, 0.0.0.0:8443 ssl
+#proxy_listen = 0.0.0.0:8000 reuseport backlog=16384, 0.0.0.0:8443 ssl reuseport backlog=16384
                          # Comma-separated list of addresses and ports on
                          # which the proxy server should listen for
                          # HTTP/HTTPS traffic.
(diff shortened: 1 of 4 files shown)


bungle:
While you are at it, mind you adding `so_keepalive=*` too?

dndx:
@bungle Yes that's a good suggestion. Although so far there has been no use cases associated on `so_keepalive` in both Kong and Nginx Kubernetes Ingress so I would say we can do this in a later version and merge this PR as-is for now. I have created a card to keep track of it.

PR #5335, opened 2019-12-13
feat(rate-limiting) add support for draft-rfc rate-limiting headers

### LINKED ISSUE #233: [ratelimiting] X-RateLimit-Reset

bungle:
### Summary

This PR adds support for the following headers specified in
https://tools.ietf.org/html/draft-polli-ratelimit-headers-01:

- RateLimit-Limit
- RateLimit-Remaining
- RateLimit-Reset

For `RateLimit-Limit` this does not implement multiple windows support (I removed it for now):
https://tools.ietf.org/html/draft-polli-ratelimit-headers-01#section-6.2.4

All this work was done by @ioggstream (give him a t-shirt or two!) on
several attempts: #5330 and #3451, and btw. @ioggstream is the author
of that draft rfc, which I hope gets promoted to a proposed standard!
Great stuff! And sure we want to support it.

I (@bungle) gathered the ideas and feedback from the previous PRs
(including #2735) and tried to make all these in a state where we
can agree and merge. I have refactored the code a bit when adding
these new headers. And I also added a bunch of tests.

This PR does not touch the previous `proprietary` headers the plugin
has sent for a long time, namely:
- X-RateLimit-Limit-*
- X-RateLimit-Remaining-*

We can later on decide, perhaps when this is standardized (or some time
after that), if we want to drop support on these, but at this point
they remain.

### Issues Resolved

Fix #233

### Note

Lets review and merge #5329 before this.

Also if we keep all rights to change headers if the `RFC` changes.

### DIFF
--- a/spec/03-plugins/23-rate-limiting/04-access_spec.lua
+++ b/spec/03-plugins/23-rate-limiting/04-access_spec.lua
@@ -269,6 +269,10 @@ for _, strategy in helpers.each_strategy() do
 
             assert.are.same(6, tonumber(res.headers["x-ratelimit-limit-minute"]))
             assert.are.same(6 - i, tonumber(res.headers["x-ratelimit-remaining-minute"]))
+            assert.are.same(6, tonumber(res.headers["ratelimit-limit"]))
+            assert.are.same(6 - i, tonumber(res.headers["ratelimit-remaining"]))
+            local reset = tonumber(res.headers["ratelimit-reset"])
+            assert.equal(true, reset <= 60 and reset >= 0)
           end
 
           -- Additonal request, while limit is 6/minute
@@ -276,8 +280,14 @@ for _, strategy in helpers.each_strategy() do
             headers = { Host = "test1.com" },
           }, 429)
 
+          assert.are.same(6, tonumber(res.headers["ratelimit-limit"]))
+          assert.are.same(0, tonumber(res.headers["ratelimit-remaining"]))
+
           local retry = tonumber(res.headers["retry-after"])
-          assert.equal(true, retry <= 60 and retry >= 0)
+          assert.equal(true, retry <= 60 and retry > 0)
+
+          local reset = tonumber(res.headers["ratelimit-reset"])
+          assert.equal(true, reset <= 60 and reset > 0)
 
           local json = cjson.decode(body)
           assert.same({ message = "API rate limit exceeded" }, json)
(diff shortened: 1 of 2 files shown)


spec/03-plugins/23-rate-limiting/04-access_spec.lua
```diff
@@ -12,6 +12,17 @@ local fmt = string.format
 local proxy_client = helpers.proxy_client
 
 
+local function it_with_retry(desc, test)
+  local ok = pcall(it, desc, test)
```
hishamhm: This doesn't work because `it` doesn't actually run the tests. It schedules them so that the `busted` runner runs them. This `pcall` never fails, even when a test contains an assert failure or on error.

If `it_with_retry` worked correctly, if you added a forced error such as `assert.same(false, true)`, then you should have seen in the output log an error message for the failure _twice_, and if you did something like `if not global_sentinel then global_sentinel = true; assert.same(false, true); end` you should have seen in the output log an error message for the first failure followed by an `OK`.

bungle: @hishamhm thank you for this, I updated it, how does it look now?

hishamhm: It is correct now!

spec/03-plugins/23-rate-limiting/04-access_spec.lua
```diff
@@ -12,6 +12,17 @@ local fmt = string.format
 local proxy_client = helpers.proxy_client
 
 
+local function it_with_retry(desc, test)
```
hishamhm: This needs a big comment explaining why this is done and why it works, because adding retries is **NOT** a pattern that we should adopt in general for dealing with flaky tests:

```suggestion
-- This performs the test up to two times (and no more than two).
-- We are **not** retrying to "give it another shot" in case of a flaky test.
-- The reason why we allow for a single retry in this test suite is because
-- tests are dependent on the value of the current minute. If the minute
-- flips during the test (i.e. going from 03:43:59 to 03:44:00), the result
-- will fail. Since each test takes less than a minute to run, running it
-- a second time right after that failure ensures that another flip will
-- not occur. If the second execution failed as well, this mean that there
-- was an actual problem with the test.
local function it_with_retry(desc, test)
```

spec/03-plugins/23-rate-limiting/04-access_spec.lua
```diff
@@ -12,6 +12,17 @@ local fmt = string.format
...
+  return it(desc, function(...)
+    local ok = pcall(test, ...)
+    if not ok then
+      ngx.sleep(2)
```
hishamhm: This sleep should not be necessary. If a test failed due to a minute-flip, it should be possible to execute it immediately (and it's actually better to _not_ sleep, because it gives it more time to run within the following minute).

ioggstream:
@hishamhm @bungle That's a fantastic news. I just added Kong to the list of API Gateways that is working towards standardizing the implementation.

- https://github.com/teamdigitale/api-gateway-tools/tree/master/gateways/kong

@hishamhm if you prefer, to clarify those headers are still in draft you can prefix them with `X-`.
Once the spec goes RFC we can lawfully remove the `X-`.

PR #5342, opened 2019-12-16
feat(core) update cipher suites to match Mozilla v5 cipher suites

bungle:
### Summary
 
Mozilla has updated their cipher suite to version 5. Thus we need to update our cipher suites: `modern`, `intermediate`, and `old` to match them. Mozilla's new `modern` suite is quite aggressive (e.g. only `TLSv1.3`), and thus this PR also changes default suite from `modern` to `intermediate`.
    
### Full changelog
    
- update ssl cipher suites to Mozilla v5
- change default ssl cipher suite from `modern` to `intermediate` because Mozilla's new ones are quite aggressive, and `intermediate` feels like a better default albeit less secure.
- predefined cipher suite now also overrides the `ssl_prefer_server_ciphers`
- `ssl_ciphers` is not specified for `modern` or `intermediate` anymore
- `ssl_prefer_server_ciphers` was moved to injection
- removes dependency on `http` library
- adds dev dependency on `http` library as it is still needed for some tests
- removes hard-coded ssl_session_timeout (default is now `1d`), can now be injected
- turns ssl_session_tickets off

In addition to that this PR also does small changes listed below (and in separate commits):

#### chore(*) service mesh leftovers removal on runloop/certificate
  
Removes some unneeded leftovers that were left in code but not used anymore after the service mesh removal.

#### tests(*) update custom template to match default kong template
    
Several things have been added to Kong template but we have forgotten to add them to our test template. This PR fixes that and brings the test nginx template on bar with Kong template.

#### chore(conf) add missing upstream and downstream grpc headers
    
- The `TE` parameter only passes the `trailers` but this was missing on `grpc`
- Downstream `Server` and `Date` headers were added for `grpc` just like we do with http and http2
- The `grpc_ssl_name` is now also set as we do for `http2` and `http`


kong/templates/kong_defaults.lua
```diff
@@ -39,6 +39,7 @@ client_body_buffer_size = 8k
 error_default_type = text/plain
 
 nginx_http_ssl_protocols = TLSv1.1 TLSv1.2 TLSv1.3
+nginx_http_ssl_prefer_server_ciphers = on
```
gszr: Should the default be `off`, given the default for `ssl_cipher_suite` is `intermediate` and in that case the client chooses?

spec/fixtures/custom_nginx.template
```diff
@@ -188,31 +199,53 @@ http {
...
         }
 
-        location @grpc {
+        location @grpcs {
             internal;
-
             set $kong_proxy_mode  'grpc';
-            grpc_pass grpc://kong_upstream;
+
+            grpc_set_header    TE                $upstream_te;
```
gszr: Still unsure we should add this; [according to the spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests), it'll be set to `trailers`, so might be better to just leave it up to the grpc module.

kong/templates/kong_defaults.lua
dndx: See: https://github.com/mozilla/server-side-tls/issues/260

I think we should include `prefer_server_ciphers` as part of the configuration bundle too, and only enable it for `old` configurations.

dndx: In fact I think that's what @bungle did already in https://github.com/Kong/kong/pull/5342/files#diff-a2237d21dd96315727f1b771bf0111f5R479, ignore last comment.

kong/templates/nginx_kong.lua
```diff
@@ -246,8 +254,9 @@ server {
 
     ssl_session_cache shared:SSL:10m;
     ssl_session_timeout 10m;
```
dndx: One small suggestion is to bump this to `1d` (as suggested by Mozilla).

The session ticket cache has LRUs and will keep the newest one in.

kong/templates/kong_defaults.lua
bungle: @gszr rationale for this being default is that when user wants to use `custom` cipher suite, it will default to `prefer cipher suites`. We can perhaps change that later. But for older suites it was generally preferred. This default is ignored with `modern`, `intermediate` and `old`, see: https://github.com/Kong/kong/pull/5342/files#diff-a2237d21dd96315727f1b771bf0111f5R22-R74

spec/fixtures/custom_nginx.template
bungle: @gszr we have special treatment for `trailers` already:
https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L1346-L1352

kong/templates/nginx_kong.lua
bungle: @dndx I added this as part of the suite

kong/templates/kong_defaults.lua
bungle: ... and if you are not using `older` suite, you probable should then consider using `intermediate` or `modern` directly without trying to specify your own.

bungle: I am ok to change the default to `off` if we think that it is more common use case. Let me know. Above was my reasoning for that (also it was the old default).

dndx: I think we should leave it to `on` by default for `custom`, as in most cases that may result in a more secure configuration mode.

kong/templates/kong_defaults.lua
```diff
@@ -39,6 +39,9 @@ client_body_buffer_size = 8k
...
+nginx_http_ssl_session_tickets = off
```
dndx: Actually let's keep the `session_tickets` as-is for now. I know it has a security tradeoff but disabling it could have huge impact on session resumption rate and cause performance regressions, and better leave this decision to the user.

bungle: @dndx it is not a small security tradeoff, it basically destroys perfect forward secrecy:
https://github.com/mozilla/server-side-tls/issues/135

bungle: @dndx what I mean is that you can only fix it with:
`ssl_session_ticket_key` + `reloading` OR `restarting` Kong daily.

dndx: No, it doesn't necessarily destroys PFS, it only destroys PFS if Nginx has not be restarted for months or years **and** the memory of Nginx is compromised, which is relatively unlikely to happen. But if we  do decide to disable it I would like to see some big warning coming in the release notes as there will likely be a performance regression.

bungle: ok, I changed the defaults and left only `protocols`, `ciphers` and `prefer_server_ciphers` on the `suite` configuration.

Issue #5301, opened 2019-12-05
Can kong support request_uri with args support matches in path?


mnteam:
Hi, there is an issue for us to deploy kong, replace the nginx also for api gateway support.

currently we use location in nginx "server" phase, with $request_uri which matches the specified args pair for route,

but when we transfer the route method with kong, seems we can only match the request with the path( req_uri without args), so I'm confused is there a better way to implement or something I don't know which support already?

I checked the contribution reference doc but not sure how the restriction style for feature request, can anyone do help? Thanks in advance.

mnteam:
Just checked the source code for router.lua, I guess the reason that not support request_uri with args support, because the cache mechanism for route match logic, that is really a good way to strip the args from request uri. But in our situation that is, we have to support the args matches. Just now I tried to re-write the router.lua, thinking if there is anyone to help me for a code review. thank you very much.


The main way I just added another function find_route_ng with the additional parameter "orig_req_uri" passed, and only used for regex matches with uri mode; also disallow the cache setting.

the only issue I currently know is, see below sample:
1. uri: /api/mineapp/res?foo=bar
expected path[]=/api/mineapp/res.*foo=bar route to service A

2. uri: /api/mineapp/res
expected path[]=/api/mineapp route to service B

while both of them be matched at once, the point 1 would be route to service B by the reason of the cache key are same.
[router.lua.txt](https://github.com/Kong/kong/files/3927472/router.lua.txt)

see attached source code. BTW, the version of kong is: 1.4.0

bungle:
@mnteam, you are correct, we don't have arbitrary query parameter based routing. Though we did add `header` based routing which is quite similar to this, e.g. key-value pair matching. In that sense I think it would we a welcomed addition. One way to workaround this is to write plugin that runs on rewrite and turns query parameters to headers (and removes them on access) and then use header based routing, but it is cumbersome.

mnteam:
@bungle Hi Bungle, thank you for your response, I already changed the router.lua(see previous attached txt file) to support this feature, did the test and seems it is ok, but the bad news is: because of the uri based cache key generation, I have to disable the cache support for route find.( I see the header matches also disallow the cache, but confused if they have there are two same request uri but different head value, one of them have no header value matches, it should be an issue which would cause the incorrect route mapping)

the next we will have a test scenario for the no cache route matches to see how it would be.

bungle:
@mnteam great feedback! feel free to open a pull-request (or two if you think there is a bug in current header matching).

[...]

PR #5347, opened 2019-12-18
feat(route)path regex matches support req_uri with args, disable route cache

mnteam:
### Summary
add path regex matches support request uri with args, disable route cache for head mapping with uri issue
refer the feature link: https://github.com/Kong/kong/issues/5301

### Full changelog

/kong/router.lua

### Issues resolved
1. Add path regex matches support request uri with args
2. disable cache for route matches to avoid the header route issue.

All above changes tested(including stress testing) and the result is OK.

For point 2, I need to explain as below: a simple case, Please tell me if I get wrong, thank you very much.

there are two routing with different service:
route a: path: /api/xxx/one, with head matches: x-token=foo
route b: path: /api/xxx/one

while any request match a, then the rest of request take the cache logic with below code:

``` lua
-- cache lookup (except for headers-matched Routes)
    local cache_key = req_method .. "|" .. req_uri .. "|" .. req_host ..
                      "|" .. ctx.src_ip .. "|" .. ctx.src_port ..
                      "|" .. ctx.dst_ip .. "|" .. ctx.dst_port ..
                      "|" .. ctx.sni
    do
      local match_t = cache:get(cache_key)
      if match_t then
        return match_t
      end
    end
```

NOTE: the general way, last five args in cache key is empty. which means the cache key is decided by req_method, req_uri and req_host.

which means the requests which should be matched with route b, but by this cache logic, it would match the cache a(no change to check the head parts).

Another point, involve the args with request uri support, we should not support cache, otherwise, the cache would have a low hit rate.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -1164,7 +1164,7 @@ function _M.new(routes)
 
   local grab_req_headers = #plain_indexes.headers > 0
 
-  local function find_route(req_method, req_uri, req_host,
+  local function find_route_ng(req_method, req_uri, orig_req_uri, req_host,
                             src_ip, src_port,
                             dst_ip, dst_port,
                             sni, req_headers)
@@ -1198,6 +1198,7 @@ function _M.new(routes)
 
     req_method = req_method or ""
     req_uri = req_uri or ""
+    orig_req_uri = orig_req_uri or ""
     req_host = req_host or ""
     req_headers = req_headers or EMPTY_T
 
@@ -1213,17 +1214,20 @@ function _M.new(routes)
 
     -- cache lookup (except for headers-matched Routes)
 
-    local cache_key = req_method .. "|" .. req_uri .. "|" .. req_host ..
+    local cache_key = ""
+    --[[                req_method .. "|" .. req_uri .. "|" .. req_host ..
                       "|" .. ctx.src_ip .. "|" .. ctx.src_port ..
                       "|" .. ctx.dst_ip .. "|" .. ctx.dst_port ..
                       "|" .. ctx.sni
-
+    --]]
+    --[[ disable cache
     do
       local match_t = cache:get(cache_key)
       if match_t then
         return match_t
       end
     end
+    --]]
 
     -- input sanitization for matchers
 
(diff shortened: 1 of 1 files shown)


dndx:
Hello @mnteam,

Thanks for your PR! I have some thoughts on this functionality after reading it.

1. I am worried doing regex matching on the full path (with arguments) is a dangerous approach. Consider the following scenario:
   * `/foo?bar=1&baz=2`
   * `/foo?baz=2&bar=1`

    Semantically those two paths are identical, but it can not be easily represented using regex. This seems to 
    create more confusion for the user than it's worth.

    What I believe may be a better approach is to follow what the header based matching PR 
    (https://github.com/Kong/kong/pull/4758) did and make sure we are matching the query arguments by 
    *semantic* instead of *string*.

2. Not sure if I got you correctly but right now if header based matcher is evaluated, the match result is never cached. Please see: https://github.com/Kong/kong/blob/ff8b4cf/kong/router.lua#L1647-L1649

    To make sure of that I created the scenario you described locally and wasn't able to reproduce any cache 
    poisoning issues. Would you mind share your reproducible configuration with me?

Thanks

mnteam:
Hi,

Thanks for your response, 

For 1, actually we use only one parameter pair for routing, in your case, that is actually the issue, I added it because it is supported by nginx, and we replaced from nginx to kong, to avoid more changes by biz side, maybe this feature is not required for new
 structure. I agree with you, if there are multiple arguments to matches, it is better to use header value pair.

For 2, I checked the latest code, it has been fixed, refer the key point of codes: router.lua -> line[1369-1385, 1387]. So you can ignore this issue right now.
(In our scenario, we have to disable all cache, because the req_uri with args support would be out of control with cache).

Anyway thanks very much for supporting.

Best Regards,
Peyton

dndx:
Hi @mnteam,

Yes, for an PR to be considered into the official Kong repository it has to solve some common use cases among Kong users. If you can rewrite this PR so that it follows a similar approach as the route by header PR I think it would meet that requirement.

Also, some quick test cases would be nice. Again, #4758 is a good reference. Let me know if you need any help along the way!

Thanks

hishamhm:
Closing this one as we agree it's not the way to go, if alternative approaches are considered, let's do them in a new PR!

PR #5351, opened 2019-12-18
chore(conf) remove KONG_ORIGINS

bungle:
### Summary

`KONG_ORIGINS` was added for Service Mesh, and while it has some use
outside mesh, it is not worth to keep around anymore. Also removal of
`transparent` flag makes this even less usable.

`Kuma` has taken over this.


hutchic:
This looks to be removing tests related to serviceless request termination which we'll still have as a feature even without service mesh correct?

bungle:
> This looks to be removing tests related to serviceless request termination which we'll still have as a feature even without service mesh correct?

Yes, but only those tests that use `origins`. I will soon post another PR that removes serviceless proxying too.

PR #5353, opened 2019-12-18
chore(proxy) remove serviceless proxying

bungle:
### Summary

Serviceless proxying was added for Service Mesh, and while it has some
use outside mesh, it is not worth to keep around anymore. Also removal
of `transparent` flag makes this even less usable.

`Kuma` has taken over this.


kong/pdk/response.lua
```diff
@@ -507,7 +507,6 @@ local function new(self, major_version)
       is_grpc_output = is_grpc
     elseif req_ctype then
       is_grpc = find(req_ctype, CONTENT_TYPE_GRPC, 1, true) == 1
-                  and ngx.req.http_version() == "2"
```
dndx: Is this intentional?

hbagdi:
Is having a route with an AWS Lambda plugin considered service-less proxying?

kong/pdk/response.lua
bungle: This is intentional, I couldn’t get the new `503` to work with grpc tests reliably without this. I can recheck why it was happening.

bungle:
> Is having a route with an AWS Lambda plugin considered service-less proxying?

no, it is considered serviceless route. lambda will terminate before this will terminate. this just ensures that request is terminated (and not serviceless proxied) before content phase in case the plugin hasn't done that before.

kong/pdk/response.lua
bungle: The problem I solved witt this is the following.

**I wanted this:**
```
$ KONG_PROXY_LISTEN="0.0.0.0:8000 http2 reuseport backlog=16384, 0.0.0.0:8443 http2 ssl reuseport backlog=16384" ./bin/kong restart
Kong stopped
Kong started
(chore/remove-serviceless-proxying) bungle@pro:~/Sources/kong$ bin/grpcurl -proto ./spec/fixtures/grpc/hello.proto -v -authority serviceless-route-grpc.test -plaintext -d '{"greeting":"world!"}' 0.0.0.0:8000 hello.HelloService.SayHello

Resolved method descriptor:
rpc SayHello ( .hello.HelloRequest ) returns ( .hello.HelloResponse );

Request metadata to send:
(empty)

Response headers received:
(empty)

Response trailers received:
content-length: 0
date: Thu, 19 Dec 2019 09:12:30 GMT
server: kong/1.4.2
x-kong-response-latency: 1
Sent 1 request and received 0 responses
ERROR:
  Code: Unavailable
  Message: no Service found with those values
```

**without the change I would get:**
```
(chore/remove-serviceless-proxying) bungle@pro:~/Sources/kong$ KONG_PROXY_LISTEN="0.0.0.0:8000 http2 reuseport backlog=16384, 0.0.0.0:8443 http2 ssl reuseport backlog=16384" ./bin/kong restart
Kong stopped
Kong started
(chore/remove-serviceless-proxying *) bungle@pro:~/Sources/kong$ bin/grpcurl -proto ./spec/fixtures/grpc/hello.proto -v -authority serviceless-route-grpc.test -plaintext -d '{"greeting":"world!"}' 0.0.0.0:8000 hello.HelloService.SayHello

Resolved method descriptor:
rpc SayHello ( .hello.HelloRequest ) returns ( .hello.HelloResponse );

Request metadata to send:
(empty)

Response headers received:
(empty)

Response trailers received:
(empty)
Sent 1 request and received 0 responses
ERROR:
  Code: Internal
  Message: transport: received the unexpected content-type "application/json; charset=utf-8"
```

bungle: @dndx,

Oh nevermind. There was bug in a first place, e.g. in original code before this PR we had:
```
and ngx.req.http_version() == "2"
```
When it should have been:
```
and ngx.req.http_version() == 2
```

It is now corrected, and the bug is fixed too. Thanks for pointing this out again. So the check is kept, but bug is fixed.

hishamhm: oh, the lack of types...

hbagdi:
> this just ensures that request is terminated (and not serviceless proxied) before content phase in case the plugin hasn't done that before.

So, would a user still need a "fake" service for request-termination plugin?

hishamhm:
@hbagdi even with this PR we will still support `service = null` in `routes`. Does that clarify?

hbagdi:
> @hbagdi even with this PR we will still support service = null in routes. Does that clarify?

Yes. Thank you!
The code change indicated so but I was not confident enough.

PR #5354, opened 2019-12-18
test(invalidator) a slight bump in the POLL_INTERVAL

hutchic:
Make the test more reliable

### DIFF
--- a/spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua
+++ b/spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua
@@ -3,7 +3,7 @@ local helpers      = require "spec.helpers"
 local ssl_fixtures = require "spec.fixtures.ssl"
 
 
-local POLL_INTERVAL = 0.3
+local POLL_INTERVAL = 0.5
 
 
 for _, strategy in helpers.each_strategy() do


spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua
```diff
@@ -3,7 +3,7 @@ local helpers      = require "spec.helpers"
 local ssl_fixtures = require "spec.fixtures.ssl"
 
 
-local POLL_INTERVAL = 0.3
```
hishamhm: this POLL_INTERVAL also controls `db_update_frequency`, meaning that if we raise this number, then the update frequency gets _slower_.

If you only want to add more time to the propagation delay, I'd recommend doing it in line 65 instead. 

But the way to to make this test file magic-sleep-number-independent would be to replace those `wait_for_propagation()` delays with using `helpers.wait_until()` instead (essentially remove the `wait_for_propagation()` call and wrap the API call below it with `helpers.wait_until(function() ... end)`
( [03-plugins/25-oauth2/04-invalidations_spec.lua](https://github.com/Kong/kong/blob/master/spec/03-plugins/25-oauth2/04-invalidations_spec.lua#L148-L155) is a good example)

hishamhm:
@hutchic let's close this PR for now — we can take a crack at another one with the `wait_until` changes as discussed above another time.

PR #5358, opened 2019-12-19
docs(autodoc) generate lists of possible values from `one_of`

hishamhm:
Our lists of accepted protocols in the `service` entity was outdated in the 1.4.x documentation.

Instead of fixing it once and risking getting this kind of thing out-of-date again, we now generate the list of accepted values in the documentation, the same way we do for presenting default values.

A [companion PR was sent to the docs repo](https://github.com/kong/docs.konghq.com/pull/1673) with the generated data.

### DIFF
--- a/autodoc/data/admin-api.lua
+++ b/autodoc/data/admin-api.lua
@@ -428,7 +428,6 @@ return {
         protocol = {
           description = [[
             The protocol used to communicate with the upstream.
-            It can be one of `http` or `https`.
           ]]
         },
         host = {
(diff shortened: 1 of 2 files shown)


autodoc/data/admin-api.lua
```diff
@@ -903,8 +902,7 @@ return {
         ]] },
         protocols = {
           description = [[
-            A list of the request protocols that will trigger this plugin. Possible values are
-            `"http"`, `"https"`, `"tcp"`, and `"tls"`.
+            A list of the request protocols that will trigger this plugin.
 
             The default value, as well as the possible values allowed on this field, may change
             depending on the plugin type. For example, plugins that only work in stream mode will
```
gszr: Unrelated: "For example, plugins that only work in stream mode `will may only` support "tcp" and "tls"." -> "For example, plugins that only work in stream mode `will only` support "tcp" and "tls"."

hishamhm: sneaked the fix in this PR as well! thanks!

PR #5362, opened 2019-12-20
chore(deps) bump zipkin to 0.2.1

kikito:


hishamhm:
@kikito For knowledge share: the point of the `~> x.y` version dependency is to match any `x.y.*`. This means that a 0.2.1 update gets picked up automatically. By specifying an explicit `z` like in this PR, you're breaking that loose dependency (which was on purpose) and pinning the rockspec to a specific release. 

This was added in other PRs recently (for example, in the service mesh removal for @dndx) and I reverted those by hand at the time of merging.

PR #5366, opened 2019-12-21
Feat/proc mgmt

javierguerragiraldez:
### Summary
process management to make the running of go-pluginserver transparent to the user

- starts it when needed
- restarts if it dies
- it's killed when nginx quits.

TODOs:

- ~~the server is started on the first request that tries to connect to it.~~
  - ~~it would be nice to do it proactively~~
  - ~~after starting, it sleeps for a while before retrying the connection.  would be best to wait until an event instead of a "reasonable delay"~~
  - ~~the process management is tied to worker #0, if that first request goes to a different worker, it won't start the pluginserver and the request would fail.  this would happen a few times until worker #0 gets one of those requests.~~
- throttle the respawning to avoid 100% lockup if the pluginserver fails to start

### DIFF
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -193,14 +193,7 @@ do
       return ffi_sock(go.socket_path())
     end
 
-    local conn, err = ngx.socket.connect("unix:" .. go.socket_path())
-    if not conn and err == "connection refused" then
-      go.manage_pluginserver()
-      ngx.sleep(0.1)
-      conn, err = ngx.socket.connect("unix:" .. go.socket_path())
-    end
-
-    return conn, err
+    return ngx.socket.connect("unix:" .. go.socket_path())
   end
 end
 go.get_connection = get_connection
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -1,11 +1,9 @@
-local ffi = require("ffi")
 local cjson = require("cjson.safe")
 local ngx_ssl = require("ngx.ssl")
 local basic_serializer = require "kong.plugins.log-serializers.basic"
 local msgpack = require "MessagePack"
 
 
-local C = ffi.C
 local kong = kong
 local ngx = ngx
 local ngx_timer_at = ngx.timer.at
@@ -79,126 +77,6 @@ do
 end
 
 
--- Workaround: if cosockets aren't available, use raw glibc calls
-local get_connection
-do
-  pcall(ffi.cdef, [[
-    static int const AF_UNIX = 1;
-    static int const SOCK_STREAM = 1;
-
-    struct sockaddr_un {
-      uint16_t  sun_family;               /* AF_UNIX */
-      char      sun_path[108];            /* pathname */
-    };
-
-    typedef struct ffi_sock {
-      int fd;
-    } ffi_sock;
-
-    int socket(int domain, int type, int protocol);
-    int connect(int sockfd, const void *addr, uint32_t addrlen);
-    ssize_t write(int fd, const void *buf, size_t count);
-    ssize_t read(int fd, void *buf, size_t count);
-    int close(int fd);
-  ]])
-
-  local ffi_sock = ffi.typeof("ffi_sock")
-
-  local function un_addr(path)
-    local conaddr = ffi.new('struct sockaddr_un')
-    if conaddr == nil then
-      return nil, "can't allocate sockaddr_un"
-    end
-
-    conaddr.sun_family = C.AF_UNIX
-
-    local len = #path + 1
-    if len > 107 then
-      len = 107
-    end
-    ffi.copy(conaddr.sun_path, path, len)
-    return conaddr
-  end
-
-  pcall(ffi.metatype, ffi_sock, {
-    __new = function(ct, path)
-      local fd = C.socket(C.AF_UNIX, C.SOCK_STREAM, 0)
-      if fd < 0 then
-        return nil, "can't create socket"
-      end
-
-      local conaddr = un_addr(path)
-      local res = C.connect(fd, conaddr, ffi.sizeof(conaddr))
-      if res < 0 then
-        C.close(fd)
-        local errno = ffi.errno()
-        return nil, "connect failure: " .. ffi.string(C.strerror(errno))
-      end
-
-      return ffi.new(ct, fd)
-    end,
-
-    __index = {
-      send = function(self, s)
-        local p = ffi.cast('const uint8_t *', s)
-        local sent_bytes = 0
-        while sent_bytes < #s do
-          local rc = C.write(self.fd, p+sent_bytes, #s-sent_bytes)
-          if rc < 0 then
-            local errno = ffi.errno()
-            return nil, "error writing to socket: " .. ffi.string(C.strerror(errno))
-          end
-          sent_bytes = sent_bytes + rc
-        end
-        return sent_bytes
-      end,
-
-      receiveany = function(self, n)
-        local buf = ffi.new('uint8_t[?]', n)
-        if buf == nil then
-          return nil, "can't allocate buffer."
-        end
-
-        local rc = C.read(self.fd, buf, n)
-        if rc < 0 then
-          return nil, "error reading"
-        end
-        return ffi.string(buf, rc)
-      end,
-
-      setkeepalive = function(self)
-        if self.fd ~= -1 then
-          C.close(self.fd)
-          self.fd = -1
-        end
-      end,
-    },
-
-    __gc = function(self)
-      self:setkeepalive()
-    end,
-  })
-
-  local too_early = {
-    init = true,
-    init_worker = true,
-    set = true,
-    header_filter = true,
-    body_filter = true,
---     log = true,
-  }
-
-  function get_connection()
-    if too_early[ngx.get_phase()] then
-      return ffi_sock(go.socket_path())
-    end
-
-    return ngx.socket.connect("unix:" .. go.socket_path())
-  end
-end
-go.get_connection = get_connection
-
-
 -- This is the MessagePack-RPC implementation
 local rpc_call
 do
@@ -222,7 +100,7 @@ do
     msg_id = msg_id + 1
     local my_msg_id = msg_id
 
-    local c, err = get_connection()
+    local c, err = ngx.socket.connect("unix:" .. go.socket_path())
     if not c then
       kong.log.err("trying to connect: ", err)
       return nil, err
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -444,18 +444,25 @@ end
 local get_plugin do
   local loaded_plugins = {}
 
+  local function get_plugin_info(name)
+    local cmd = string.format(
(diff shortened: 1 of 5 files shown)


kong/db/dao/plugins/go.lua
```diff
@@ -1,6 +1,7 @@
 local ffi = require("ffi")
 local cjson = require("cjson.safe")
 local ngx_ssl = require("ngx.ssl")
+local ngx_pipe = require "ngx.pipe"
```
gszr: Didn't get to the bottom of the issue, but moving `ngx.pipe` to the scope where it's used in fixes the tests; further, the error only happens if certain combinations of specs are executed in a specific order (presumably some side effect going on in those specific paths). Can look further into this later on, but for now I'd propose we simply move the `require` : )

javierguerragiraldez: thanks!  i saw that the failure was clearly a test thing.... will experiment on the best place to put it

kong/db/dao/plugins/go.lua
```diff
@@ -36,6 +36,47 @@ do
   end
 end
 
+do
+  local pluginserver_proc
+
+  function go.manage_pluginserver()
+    if pluginserver_proc then
```
dndx: Since `manage_pluginserver` is executed in `init_worker` phase once per worker, I don't think we need this `pluginserver_proc ` check. Simply check to see if `ngx.worker.id()` is `0` and only spawns the timer when it is will be enough.

kong/db/dao/plugins/go.lua
```diff
@@ -36,6 +36,47 @@ do
...
+        kong.log.notice("Starting go-pluginserver")
+        pluginserver_proc = assert(ngx_pipe.spawn({
+          kong.configuration.prefix .. "/bin/go-pluginserver",
+          "-socket", go.socket_path(),
+          "-plugins-directory", kong.configuration.go_plugins_dir,
+        }, {
+          environ = { "PATH=" .. kong.configuration.prefix },
+        }))
+
+        while true do
```
dndx: We should check to see if `ngx.worker.exiting()` returns `true`, if so then the process should be killed.

kong/db/dao/plugins/go.lua
```diff
@@ -36,6 +36,47 @@ do
...
+    if ngx.worker.id() ~= 0 then
+      -- only one manager
+      pluginserver_proc = true
+      return
+    end
+
+    ngx_timer_at(0, function ()
+      local ngx_pipe = require "ngx.pipe"
+
+      while true do
```
dndx: We should check to see if `ngx.worker.exiting()` returns `true`, if so then the process should not be respawned.

kong/db/dao/plugins/go.lua
```diff
@@ -36,6 +36,47 @@ do
...
+      return
+    end
+
+    if ngx.worker.id() ~= 0 then
+      -- only one manager
+      pluginserver_proc = true
+      return
+    end
+
+    ngx_timer_at(0, function ()
```
dndx:
```lua
ngx_timer_at(0, function(premature)
  if premature then
    return
  end
...
```

is better, to avoid prolonging the exit of worker process by the timer.

Also in Kong the styling does not leaves a space between `function` and `()`.

kong/db/dao/plugins/go.lua
```diff
@@ -36,6 +36,47 @@ do
...
+          "-socket", go.socket_path(),
+          "-plugins-directory", kong.configuration.go_plugins_dir,
+        }, {
+          environ = { "PATH=" .. kong.configuration.prefix },
+        }))
+
+        while true do
+          local ok, reason, status = pluginserver_proc:wait()
+          if ok ~= nil or reason == "exited" then
+            kong.log.notice ("go-pluginserver terminated: ", tostring(reason), " ", tostring(status))
```
dndx: Style: `kong.log.notice(` instead of `kong.log.notice (`.

kong/db/dao/plugins/go.lua
```diff
@@ -36,6 +36,47 @@ do
...
+      return
+    end
+
+    ngx_timer_at(0, function ()
+      local ngx_pipe = require "ngx.pipe"
+
+      while true do
+        kong.log.notice("Starting go-pluginserver")
+        pluginserver_proc = assert(ngx_pipe.spawn({
+          kong.configuration.prefix .. "/bin/go-pluginserver",
```
gszr: We should not assume the pluginserver will be available in the prefix; by default it will not (all our distributions place it under `/usr/local/bin`. Something like the following should do it:

```
pluginserver_proc = assert(ngx_pipe.spawn({
    "go-pluginserver",
    "-socket", go.socket_path(),
    "-plugins-directory", kong.configuration.go_plugins_dir,
   }, {
    environ = { "PATH=/usr/local/bin" }
   }))
```

javierguerragiraldez: unfortunately, the `environ` variables apply only for the spawned process, not while searching the executable.  i'll hardcode the complete `/usr/local/bin/go-pluginserver` path.

kong/db/dao/plugins/go.lua
javierguerragiraldez: `go.manage_pluginserver()` is an exported function, so I didn't want to assume it will only be called once at the appropriate time.  also, we do need to keep the module-level variable, so checking if it's already set is cheaper than checking the worker id, since it's set to `true` on non-0 workers.  Not that this is in any way performance sensitive...

kong/db/dao/plugins/go.lua
javierguerragiraldez: Initially I intended to add a `plugin.Exit()` method, but crash-only processes are simpler to maintain.  the process is currently killed by the OS when the worker exits, but i guess it's nicer to do it here first.

kong/db/dao/plugins/go.lua
dndx: We export a lot of functions, and if it is not documented in the PDK it is assumed to not be called elsewhere. In this case, I don't see any reason anyone should or would be calling it.

dndx: As mentioned, if for defensive programming reason to keep this check, an `assert` is more appropriate than `if`, in order to not hide incorrect usage in our code base.

kong/db/dao/plugins/go.lua
hishamhm: >  i'll hardcode the complete /usr/local/bin/go-pluginserver path.

This is bad, it will cause trouble with testing environments and such. If `ngx_pipe.spawn` doesn't know how to search the PATH we have to either do it ourselves or add a `kong.conf` option for the full path of the pluginserver, which kinda sucks. I would loop over the PATH variable searching for the file before spawning.

kong/db/dao/plugins/go.lua
```diff
@@ -444,18 +471,22 @@ end
 local get_plugin do
   local loaded_plugins = {}
 
+  local function get_plugin_info(name)
+    local cmd = string.format(
+        "bin/go-pluginserver -plugins-directory %q -dump-plugin-info %q",
```
hishamhm: looks like this also needs the locatiion of `go-pluginserver`

kong/db/dao/plugins/go.lua
```diff
@@ -444,18 +471,22 @@ end
...
+        kong.configuration.go_plugins_dir, name)
+
+    local d = assert(io.popen(cmd)):read("*a")
```
hishamhm: don't wait for the GC to close the fd: store it in a variable and `close()` it explicitly.

kong/db/dao/plugins/go.lua
javierguerragiraldez: the problem is that nginx cleans up the environment.  would have to add an `env PATH;` line to the config so `pipe.spawn()` can use it.

kong/db/dao/plugins/go.lua
javierguerragiraldez: done

kong/db/dao/plugins/go.lua
javierguerragiraldez: done

kong/db/dao/plugins/go.lua
hishamhm: Oh, I see. In that case I would go with a `kong.conf` option for the full path of the pluginserver. The default can be set to `/usr/local/bin/go-pluginserver` meaning that our shipped binaries work out-of-the-box, but it's still overridable in dev environments.

javierguerragiraldez: done

kong/db/dao/plugins/go.lua
javierguerragiraldez: duh!  done

kong/db/dao/plugins/go.lua
```diff
@@ -79,7 +121,7 @@ do
   end
 
   pcall(ffi.metatype, ffi_sock, {
```
dndx: Actually some second thought on this, is it possible to fallback to LuaSocket instead of doing raw BSD socket calls like the rest of the code base?

javierguerragiraldez: AFAICT, luasocket doesn't have the equivalent of the receiveany operation, which is important for the protocol

dndx: Is it possible to do the framing ourself (e.g. send the length of the msgpack payload before the payload) instead of relying on this behavior?

Reason being, using `receiveany` is not guaranteed to work anyway, since the stream socket never guarantees message boundary, that's an upper layer protocol concern.

javierguerragiraldez: The decoder calls the receive function until it can decode a whole object.  Reading any fixed size over one byte at a time could block indefinitely. Receiveany returns what is available.  Can be less than what's sent, so it's repeated.

javierguerragiraldez: The good news is that initialization is done via pipes and we're not supporting filter phases, so this ffi option might be removed eventually

dndx: Yes I think removing it will be the best. Still I don’t think the way we are using `receiveany` is correct. The unpacker could possibly been reading more data than it needs, and because we don’t keep the unpacker around, subsequent reads from the socket could return partially read data.    

I’m definitely not referring to making 1 byte reads for all the data, that would be extremely inefficient. What I proposed is to send a 2 byte length before each msgpack message, on the other end read the fixed 2 byte length first then use its value to read exactly the needed bytes from the socket. 

Also from the look of the code, the plugin server may spontaneously send data to Kong even if Kong didn’t ask for it? If that is the case this may need to be changed, as a keepalived connection will immediately error out if Nginx receives any data on it. The only safe interaction In this case is a request-response flow.

dndx: I think another possible solution is we can simply use the data gram mode Unix socket, that way keep alive or framing are no concern.

dndx: Or is the plugin server only gonna respond one or more msgpack payloads when the Kong worker sends the call? If that is the case I think `receiveany` streaming read would work. Sorry for the extra comments but it was not clear at first look and I want to make sure we are using `receiveany` correctly here :)

PR #5368, opened 2019-12-23
tiny tweaks in CP/DP

hishamhm:


kong.conf.default
```diff
@@ -124,6 +124,45 @@
...
 
+#------------------------------------------------------------------------------
+# HYBRID MODE
+#------------------------------------------------------------------------------
+
+#role = traditional              # Use this setting to enable Hybrid Mode,
+                                 # This allows running some Kong nodes in a
+                                 # control plane role with a database and
+                                 # have them deliver configuration updates
+                                 # other nodes running to DB-less running in
```
gszr:
```suggestion
                                 # to other nodes running to DB-less running in
```

kong.conf.default
```diff
@@ -124,6 +124,45 @@
...
+                                 # when establishing secure communication
+                                 # between control and data plane nodes.
+                                 # You can use the `kong hybrid` command to
+                                 # generate the certificate/key pair.
+
+#cluster_cert_key =              # Filename of the cluster certificate key to use
+                                 # when establishing secure communication
+                                 # between control and data plane nodes.
+                                 # You can use the `kong hybrid` command to
+                                 # generate the certificate/key pair.
```
dndx: Also need:

```
#cluster_listen = 0.0.0.0:8005

                         # Comma-separated list of addresses and ports on
                         # which the cluster control plane (CP) server should listen for
                         # data plane connections.
                         # The cluster communication port of the
                         # control plane must be accessible by all the data planes
                         # within the same cluster. This port is mTLS protected
                         # to ensure end-to-end security and integrity.
                         #
                         # This setting has no effect if `role` is not set to
                         # `control_plane`
```

kong.conf.default
hishamhm: thanks!

kong.conf.default
hishamhm: thanks, added!

PR #5387, opened 2020-01-02
fix(*) use FFI-based lua-resty-openssl library

fffonion:
### Summary

Replace luaossl with lua-resty-openssl

### Full changelog

* Remove luaossl from dependency
* Bump lua-resty-openssl to 0.4.1


kong/cmd/utils/prefix_handler.lua
```diff
@@ -45,50 +45,53 @@ local function gen_default_ssl_cert(kong_config, admin)
...
-    crt:setBasicConstraints { CA = false }
-    crt:setBasicConstraintsCritical(true)
+    assert(crt:set_basic_constraints { CA = false })
+    assert(crt:set_basic_constraints_critical(true))
 
     -- Only allowed to be used for TLS connections (client or server)
-    crt:addExtension(x509_extension.new("extendedKeyUsage",
-                                        "serverAuth,clientAuth"))
+    assert(crt:add_extension(x509_extension.new("extendedKeyUsage",
+                                        "serverAuth,clientAuth")))
```
bungle:
```suggestion
                                                "serverAuth,clientAuth")))
```

kong-1.4.2-0.rockspec
```diff
@@ -36,7 +36,7 @@ dependencies = {
   "lua-resty-cookie == 0.1.0",
   "lua-resty-mlcache == 2.4.0",
   "lua-messagepack == 0.5.2",
-  "lua-resty-openssl == 0.3.0",
+  "lua-resty-openssl == 0.4.1",
```
bungle: I see you added new dependency, but the goal was to remove `luaossl` dependency. Is that still the case or why wasn't `"luaossl == 20190731"` removed?

fffonion: Thanks for catching that, I missed it = ) Adding it as a new commit and squashed the suggestion commit.

PR #5390, opened 2020-01-03
feat(config) add more nginx configuration injections

### LINKED ISSUE #5324: Kong 1.4.1 High memory per worker (Docker)
### Summary

When running Kong 1.4.1 in Docker (via nomad) then the memory usage is roughly 500Mb per worker (!!!). Testing the same config and setup with Kong 1.4.0 the total is 75Mb with 1 worker, 200Mb with 4.
Possibly related to 1.4.1: Removed arbitrary limit on worker connections. #5148

### Steps To Reproduce

1. Start Kong in docker with a simple declarative config.
Our Kong.conf:

log_level = debug
proxy_access_log = /dev/stdout
proxy_error_log = /dev/stderr
admin_access_log = /dev/stdout
admin_error_log = /dev/stderr
proxy_listen = 0.0.0.0:8443 http2 ssl
ssl_cert = /certs/local.cer
ssl_cert_key = /certs/local.key
database = off
dns_order = SRV,A,LAST,CNAME
declarative_config = /custom/custom-proxy.yml

custom-proxy.yml:
_format_version: '1.4'

services:
- name: Route-All
  port: 80
  protocol: https
  host: MyService.service.team.consul.company.com
  routes:
  - name: Route-All
    paths: ["/"]
    preserve_host: true

### Additional Details & Logs

- Kong version: 1.4.1
- Kong debug-level startup logs:
...
2019/12/11 13:46:33 [notice] 1#0: using the "epoll" event method

2019/12/11 13:46:33 [notice] 1#0: openresty/1.15.8.2

2019/12/11 13:46:33 [notice] 1#0: built by gcc 8.3.0 (Alpine 8.3.0)

2019/12/11 13:46:33 [notice] 1#0: OS: Linux 3.10.0-1062.1.1.el7.x86_64

2019/12/11 13:46:33 [notice] 1#0: getrlimit(RLIMIT_NOFILE): 1048576:1048576

2019/12/11 13:46:33 [notice] 1#0: start worker processes

2019/12/11 13:46:33 [notice] 1#0: start worker process 24

2019/12/11 13:46:33 [debug] 24#0: *1 [lua] globalpatches.lua:243: randomseed(): seeding PRNG from OpenSSL RAND_bytes()

2019/12/11 13:46:33 [debug] 24#0: *1 [lua] globalpatches.lua:269: randomseed(): random seed: 192795110016 for worker nb 0

2019/12/11 13:46:33 [debug] 24#0: *1 [lua] events.lua:211: do_event_json(): worker-events: handling event; source=resty-worker-events, event=started, pid=24, data=nil

2019/12/11 13:46:33 [notice] 24#0: *1 [kong] init.lua:298 declarative config loaded from /custom/custom-proxy.yml, context: init_worker_by_lua*

2019/12/11 13:46:33 [debug] 24#0: *2 [lua] balancer.lua:764: init(): initialized 0 balancer(s), 0 error(s)

bungle:
### Summary

- introduce `nginx_main_<directive>` injections
- introduce `nginx_events_<directive>` injections
- introduce `nginx_supstream_<directive>` injections
- introduce `nginx_main_worker_rlimit_nofile` injection with special `auto` value (`math.min(ulimit, 65536)`)
- introduce `nginx_main_events_worker_connections` injection with special `auto` value (`math.min(ulimit, 65536)`)
- move `nginx_user` to `nginx_main_user` injection
- move `nginx_daemon` to `nginx_main_daemon` injection
- move `nginx_worker_processes` to `nginx_main_worker_processes` injection
- move `nginx_http_upstream_<directive>` injections to `nginx_upstream_<directive>`
- move `nginx_http_status_<directive>` injections to `nginx_status_<directive>`
- move `client_max_body_size` to `nginx_http_client_max_body_size` injection
- move `client_body_buffer_size` to `nginx_http_client_body_buffer_size` injection
- move `real_ip_header` to `nginx_proxy_real_ip_header` injection
- move `real_ip_recursive` to `nginx_proxy_real_ip_recursive` injection
- remove the undocumented `nginx_optimizations` config parameter

### Issues Resolved

Fix #5324


bungle:
The tests are failing because this needs to be merged first:
https://github.com/Kong/kong-plugin-prometheus/pull/76

hbagdi:
Some thoughts overall:

- Overall this certainly seems a step in the direction where we (Thibault and I) wanted to be in when we initially put this feature in, so good job there!
- Question: has there been a use-case or an ask for `nginx_supstream_`? I know there are a couple of directives we set and use but the API surface because of these increase quite a bit so wee should be careful about introducing it. 
- Please keep the deprecated variables documented in the `kong.conf.default`. And make sure that they are also documented on the docs site. Some of these are highly used variables and it will take a long time for them to go away, just from Kong's documentations and all the web of products around it.
- Some of these properties like `real_ip_header` are used quite a lot. As much as we want to provide a good experience for Nginx ninjas, we also should keep Kong's configuration friendly for users who have never used Nginx before, which seems to be the majority in our user-base. I'd advise that we don't deprecate some of the config parameters that are used heavily and keep them as is. Explaining to an end user `KONG_REAL_IP_HEADER` is much easier than `KONG_NGINX_HTTP_PROXY_REAL_IP_HEADER`.
- It is **very important** to document the big deal changes here in the changelog. Having a very verbose commit with a changelog section or keep those changes around in another doc could help.


A general comment:
This is one humongous commit and it is almost impossible for me to review this thoroughly and hence         I just skimmed it. I'd encourage us to rather make 20 smaller commits (especially when there are unrelated changes), even keep the commits as one makes while hacking this together.
I understand that we are religious about commit messages and commit formats but those can be taken care of at a later stage, even after a review of the code change.

bungle:
> * Question: has there been a use-case or an ask for `nginx_supstream_`? I know there are a couple of directives we set and use but the API surface because of these increase quite a bit so wee should be careful about introducing it.

There is no reason for `http` to be more supported than `stream` in this regard. E.g. I have never heard people injecting stuff into `http` `upstream`, but that was probably added because a single person asked about it. Especially now that we get rid of `service mesh` and make `stream` standard `nginx`.

> * Please keep the deprecated variables documented in the `kong.conf.default`. And make sure that they are also documented on the docs site. Some of these are highly used variables and it will take a long time for them to go away, just from Kong's documentations and all the web of products around it.

We are not removing them at the moment. We just deprecated them. I can add them there, and to document to use a new value.

> * Some of these properties like `real_ip_header` are used quite a lot. As much as we want to provide a good experience for Nginx ninjas, we also should keep Kong's configuration friendly for users who have never used Nginx before, which seems to be the majority in our user-base. I'd advise that we don't deprecate some of the config parameters that are used heavily and keep them as is. Explaining to an end user `KONG_REAL_IP_HEADER` is much easier than `KONG_NGINX_HTTP_PROXY_REAL_IP_HEADER`.

We discussed it internally, and we got into problems with current injections. Biggest problem was that by injection name you cannot tell where the context stops and where the directive starts. Thus we changed everything to be `KONG_NGINX_<SINGLE-WORD>`. If we keep supporting multiple, e.g. `KONG_REAL_IP_HEADER` and `KONG_NGINX_PROXY_REAL_IP_HEADER` equally we get into issues of priorities, or erroring when defining both. It also adds excess code to handle these. I'd say it just brings more problems than it solves. Yes, I personally brought up the usability, but well, it is just one rename, and we are just deprecating for now.

bungle:
> * Please keep the deprecated variables documented in the `kong.conf.default`. And make sure that they are also documented on the docs site.

Deprecations added to `kong.conf.default`. Docs site is autogenerated from it.

> * It is **very important** to document the big deal changes here in the changelog. Having a very verbose commit with a changelog section or keep those changes around in another doc could help.

I feel that is scope-creep of this PR. We usually do changelogs before releases separately.
 
> This is one humongous commit and it is almost impossible for me to review this thoroughly and hence I just skimmed it. I'd encourage us to rather make 20 smaller commits

I split it to 15 commits.

kong/conf_loader.lua
```diff
@@ -89,34 +89,76 @@ local HEADER_KEY_TO_NAME = {
...
   {
     injected_conf_name = "nginx_stream_directives",
     prefix = "nginx_stream_",
+    ignore = EMPTY,
   },
   {
-    injected_conf_name = "nginx_proxy_directives",
-    prefix = "nginx_proxy_", -- TODO: nginx_http_proxy
+    injected_conf_name = "nginx_supstream_directives",
+    prefix = "nginx_supstream_",
```
dndx: Should we call this `nginx_stream_upstream_directives ` instead. I was definitely confused at first what `nginx_supstream_directives` means and the `s` is easy to miss as well.

kong/conf_loader.lua
```diff
@@ -89,34 +89,76 @@ local HEADER_KEY_TO_NAME = {
...
+    ignore = EMPTY,
   },
   {
     injected_conf_name = "nginx_sproxy_directives",
-    prefix = "nginx_sproxy_", -- TODO: nginx_stream_proxy
+    prefix = "nginx_sproxy_",
```
dndx: Same here.

kong/conf_loader.lua
```diff
@@ -167,29 +209,150 @@ local CONF_INFERENCES = {
...
+        end
+      end,
+    }
+  },
+  nginx_user = { -- TODO: remove since deprecated in 2.0
+    typ = "string",
+    deprecated = {
+      replacement = "nginx_main_user",
+      alias = function(conf)
+        if conf.nginx_main_user == nil then
```
dndx: If `if conf.nginx_main_user ~= nil`, should we error out and let the user know they can not configure both at the same time. Right now this will silently make `nginx_user` a no-op.

kong/conf_loader.lua
```diff
@@ -274,11 +438,14 @@ local CONF_INFERENCES = {
               },
   plugins = { typ = "array" },
   anonymous_reports = { typ = "boolean" },
-  nginx_daemon = { typ = "ngx_boolean" },
-  nginx_optimizations = { typ = "boolean" },
+  nginx_optimizations = {
+    typ = "boolean",
+    deprecated = {}
```
dndx: Minor styling suggestions, `deprecated = { replacement = false }` might make this easier to understand without looking at `deprecated_properties` function.

kong/conf_loader.lua
```diff
@@ -955,6 +1138,26 @@ local function load(path, custom_conf, opts)
 
   conf = tablex.merge(conf, defaults) -- intersection (remove extraneous properties)
 
+  do
+    -- nginx 'user' directive
+    local user = conf.nginx_main_user:gsub("^%s*", "")
```
dndx: Minor suggestion: use https://stevedonovan.github.io/Penlight/api/libraries/pl.stringx.html#strip instead.

kong/conf_loader.lua
bungle: @dndx we used to have `nginx_sproxy`, but the original plan was to follow nginx hierarchy:
- `nginx_*`
- `nginx_http_*`
- `nginx_http_proxy_*`

etc. and I implemented this first with that, but there was an issue that by the name of the injection we cannot determine which part of it is the context and which is the directive.

```
KONG_NGINX_HTTP_PROXY_SSL_CIPHERS=y -> proxy_ssl_ciphers OR ssl_ciphers?
```

How can you know what user means about it? I originally even went with longest match semantics, and whitelists and blacklists etc. But the problems just popped up.

Thus we changed all injections in form:
```
KONG_NGINX_<SINGLE-WORD>
```

This kept things also quite nice on backward compatibility as we already had `NGINX_HTTP_*`, `NGINX_PROXY_*`, `NGINX_SPROXY_*`. And there the `S` prefix was taken to others too, e.g. `NGINX_SUPSTREAM_*´. Yes, I know it is not as nice as that hierarchical, but at least it is shorter.

kong/conf_loader.lua
bungle: It will not make it `silently` `WARN` will be logged to both `CLI` and NGINX logs.

kong/conf_loader.lua
bungle: That code is actually not by this PR, it was there before, but sure I can change it, and I think `kong.tools.utils` has one too.

kong/conf_loader.lua
dndx: Okay, I see that, thanks for the detailed explanation!

kong.conf.default
```diff
@@ -593,6 +545,60 @@
...
+                                 # spawned by Nginx.
+                                 #
+                                 # See http://nginx.org/en/docs/ngx_core_module.html#worker_processes
+
+#nginx_main_worker_rlimit_nofile = auto
+                                 # Changes the limit on the maximum number of
+                                 # open files for worker processes.
+                                 #
+                                 # The special value `auto` means the value of
+                                 # `ulimit -n` with the upper bound of 16384.
```
bungle:
```suggestion
                                 # `ulimit -n` with the upper bound of 65536.
```

kong.conf.default
```diff
@@ -593,6 +545,60 @@
...
+                                 #
+                                 # See http://nginx.org/en/docs/ngx_core_module.html#worker_rlimit_nofile
+
+#nginx_events_worker_connections = auto
+                                 # Sets the maximum number of simultaneous
+                                 # connections that can be opened by a worker
+                                 # process.
+                                 #
+                                 # The special value `auto` means the value of
+                                 # `ulimit -n` with the upper bound of 16384.
```
bungle:
```suggestion
                                 # `ulimit -n` with the upper bound of 65536.
```

PR #5393, opened 2020-01-03
fix(schema) add empty string handling on updates for auto generated f…

artomal:
### Summary

Add check for PATCH requests on empty string field which is being set up
to `ngx.null` and is not handled properly, which let users to set `nil`
value to `auto` generated `string` fields.

Empty string conversion to `ngx.null` happens here:
https://github.com/Kong/kong/blob/bf525dcc8e731a3de208f4c19b17916946b9fa4c/kong/tools/utils.lua#L401

I couldn't add test for this fix because there is no such plugin or functionality in CE which would let you to PATCH such field and test the change, I only did manual check with a custom plugin.

### Full changelog

* Add check for PATCH request on `auto` generated field with empty value

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1597,7 +1597,8 @@ function Schema:process_auto_fields(data, context, nulls)
         end
 
       elseif field.type == "string" then
-        if (context == "insert" or context == "upsert") and data[key] == nil then
+        if (context == "insert" or context == "upsert" or context == "update")
+          and (data[key] == nil or data[key] == null) then
           data[key] = utils.random_string()
         end
 


hishamhm:
@artomal I wonder if setting `required = true` in the `auto` field would cause the same effect as this PR. If you run the code without this PR, what happens if the field is set to `auto = true, required = true`? Will it forbid a `PATCH` with `null`?

The idea of not catching `ngx.null` there was to allow for _optional_ `auto` fields that have a default auto-generated when not given, but that could still be forced into `null` if the user explicitly says so.

artomal:
@hishamhm thanks for a response, now that you explained the use case it makes sense to have such an option.

Speaking about adding configuration parameter to schema field with `required = true` for fields with `auto = true` configuration solves that problem and it doesn't fail with error message for `PATCH` requests which is good, though fails for `PUT` requests and it is not a bad behavior knowing that HTTP 'PUT' method `by nature asks you to provide the whole entity but from other side its behavior not consistent with HTTP `POST` method. I guess it is fine for now and I can close this PR since your proposal solves the problem, thanks again!

PR #5399, opened 2020-01-07
chore(reports) add go stats

gszr:
Add two reports related to Go plugins:
- Number of Go plugins currently installed
- Number of requests triggering a Go plugin


kong/init.lua
```diff
@@ -190,6 +190,10 @@ local function execute_plugins_iterator(plugins_iterator, phase, ctx)
     kong_global.set_namespaced_log(kong, plugin.name)
     plugin.handler[phase](plugin.handler, configuration)
     kong_global.reset_log(kong)
+
+    if plugin.handler._go then
+      ngx.ctx.ran_go_plugin = true
```
dndx: I think we should use the `ctx` argument here instead of directly referencing `ngx.ctx`.

kong/reports.lua
```diff
@@ -403,6 +410,10 @@ return {
     local suffix, err = get_current_suffix()
     if suffix then
       incr_counter(count_key .. ":" .. suffix)
+
+      if ngx.ctx.ran_go_plugin then
```
dndx: Same here, the `log` function should be changed to pass in `ctx` instead of using `ngx.ctx` directly, to keep the style among codes consistent.

kong/db/dao/plugins.lua
```diff
@@ -148,6 +149,9 @@ local function load_plugin_handler(plugin)
   local ok, handler = utils.load_module_if_exists(plugin_handler)
   if not ok and go.is_on() then
       ok, handler = go.load_plugin(plugin)
+      if type(handler) == "table" then
+        handler._go = true
```
dndx: Maybe I don't understand this completely, but wouldn't moving `go_plugins_cnt = go_plugins_cnt + 1` be more intuitive?

PR #5404, opened 2020-01-07
feat(runloop) dynamic default certificate

hbagdi:
### Summary

The default certificate for the proxy can now be configured via Admin API using the `/certificates` endpoint. A special '*' SNI has been introduced which stands for the default certificate.

The order in which Kong serves a certificate is:
- Exact SNI match
- Search by prefix wildcard
- Search by suffix wildcard
- Look for a cert associated with the SNI `*`
- The default certificate on file-system

This is an effort in the direction to make Kong more dynamic and minimize static configuration that is put in when the process is started.

### DIFF
--- a/kong/runloop/certificate.lua
+++ b/kong/runloop/certificate.lua
@@ -19,6 +19,7 @@ local set_priv_key = ngx_ssl.set_priv_key
 
 local default_cert_and_key
 
+local DEFAULT_SNI = "*"
 
 local function log(lvl, ...)
   ngx_log(lvl, "[ssl] ", ...)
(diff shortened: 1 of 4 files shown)


hishamhm:
@hbagdi My reaction reading the code in this PR: "that's it!?" — it's hysterical how this feature could be added with such a tiny change. Nice one! :)

I guess we need a test to verify that this cert indeed functions as a fallback?

hbagdi:
Awesome!
The tests for this were a little more involved than I had hoped so I waited for the feedback here.
Since this is scheduled for 2.1.0, I'll add in the tests in a week or two.

dndx:
Waiting for tests before merging.

hbagdi:
I added an integration test. The CI failure seems unrelated.

hbagdi:
I can't figure out why this one test related to anonymous reports fails in Cassandra but passes in Postgres.

PR #5412, opened 2020-01-08
style(go) add more comment and make the Kong <-> Plugin server

dndx:
interaction clearer

### DIFF
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -96,6 +96,12 @@ do
     end
   end
 
+  -- This function makes a RPC call to the Go plugin server. The Go plugin
+  -- server communication is request driven from the Kong side. Kong first
+  -- sends the RPC request, then it executes any PDK calls from Go code
+  -- in the while loop below. The boundary of a RPC call is reached once
+  -- the RPC response (type 1) message is seen. After that the connection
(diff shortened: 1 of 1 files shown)


kong/db/dao/plugins/go.lua
```diff
@@ -131,16 +137,18 @@ do
         end
       end
 
-      if data[1] == 1 and data[2] == my_msg_id then
-        -- it's our answer
-        if data[3] ~= nil then
-          c:setkeepalive()
-          return nil, data[3]
-        end
+      assert(data[1] == 1, "RPC response expected from Go plugin server")
```
javierguerragiraldez: there can (in theory) be more than one notification before the response.  perhaps something like
```
   if data[1] == 1 then
      assert(data[2] == my_msg_id, "wrong id")
```
would be safer.

dndx: Wouldn't the extra notifications be caught by the `if data[1] == 2 then` branch already so it will never reach this point?

javierguerragiraldez: that would get one notification.  if there's a second one (i know, we don't send any), then the assert would fail.

javierguerragiraldez: that's why it's a loop and not a single condition

PR #5413, opened 2020-01-08
chore(*) add go reports

gszr:
### Summary

Add anonymous reports for Go plugins:
- Go version in use
- Number of requests triggering a Go plugin
- Number of installed Go plugins


kong/db/dao/plugins.lua
```diff
@@ -277,6 +282,10 @@ function Plugins:load_plugin_schemas(plugin_set)
         handler = handler()
       end
 
+      if handler._go then
+        go_plugins_cnt = go_plugins_cnt + 1
```
dndx: Can this be calculated above in the `load_plugin_handler` function.

gszr: `load_plugin_handler` doesn't have a view of all plugins - it's called for each plugin - so `load_plugin_schemas` is the place where this sum can be done

dndx: Cool, thanks for clarifying!

gszr:
@javierguerragiraldez Thanks for the review. Mind re-reviewing? I just pushed two changes: 1) extracted another helper (automatically building fixture go plugins and setting up the pluginserver path) and refactored the tests a bit - to avoid using `before_each` to start Kong and create entities. Affected commits were 7accb1a and 844906d.

javierguerragiraldez:
LGTM

spec/helpers.lua
```diff
@@ -1697,13 +1697,41 @@ local function render_fixtures(conf, env, prefix, fixtures)
...
 
+  -- go plugins are enabled
+  --  set pluginserver dir (making sure it's in the PATH)
+  --  compile fixture go plugins
+  if env.go_plugins_dir then
+    if env.go_plugins_dir == GO_PLUGIN_PATH then
+      build_go_plugins(GO_PLUGIN_PATH)
+    end
+
+    if not env.go_pluginserver_exe and not os.getenv("KONG_GO_PLUGINSERVER_EXE") then
```
gszr: @javierguerragiraldez This is what I meant - only looking up in PATH if none of these have been set, as it'd otherwise potentially interfere with the use of KONG_GO_PLUGINSERVER_EXE in development workflows.

.requirements
```diff
@@ -7,3 +7,4 @@ RESTY_LUAROCKS_VERSION=3.2.1
 RESTY_OPENSSL_VERSION=1.1.1d
 RESTY_PCRE_VERSION=8.43
 LIBYAML_VERSION=0.2.2
+KONG_GO_PLUGINSERVER_VERSION=master
```
hishamhm: I think we should tag a release number there and pin it here before merging.

spec/helpers.lua
hishamhm: Sorry, I commented over an outdated commit earlier.

I still think this is not really _necessary_, but since the environment variable is taking precedence in this version, I think it doesn't hurt either. So, :+1:

.requirements
gszr: Done!

PR #5425, opened 2020-01-09
feat(conf) add dynamic directives from configuration values

bungle:
### Summary

Nginx configuration injections are nice, but one issue with them is that you need to duplicate configuration on each place where you want to inject stuff.

Some configuration we want directives to be injected in multiple places. This PR does it for:

- ssl_protocols
- ssl_prefer_server_ciphers
- ssl_session_tickets
- ssl_session_timeout

E.g.:
```
  ssl_protocols = {
    typ = "string",
    directives = {
      "nginx_http_ssl_protocols",
      "nginx_stream_ssl_protocols",
    },
  },
```

It means whenever user configures `ssl_protocols`, it will be injected to both `http` and `stream` blocks. This also fixes this issue that currently we  don't add these to `stream`.

It is still allowed for user to specify a more specific configuration value with e.g. `KONG_NGINX_HTTP_SSL_PROTOCOLS=TLSv1.1` which takes a priority over the `KONG_SSL_PROTOCOLS=TLSv1.2` on `http` block, the `stream` block will still add it as `TLSv1.2`.

This will also make it easier for us to move some still existing hard-coded directives to injections.


kong/conf_loader.lua
```diff
@@ -1041,6 +1092,28 @@ local function load(path, custom_conf, opts)
...
+        local directives = v_schema.directives
+        if directives then
+          local v = t[property_name]
+          if v then
+            for _, directive in ipairs(directives) do
+              dynamic_keys[directive] = true
+
+              if type(v) == "boolean" then
+                v = v and "on" or "off"
+              end
```
javierguerragiraldez: can this be just before the loop?  i know, there's no performance constraints, and the compiler would CSE this away; and given that it's inside the `if v then ...` block, then `v` can't be `false`... so maybe the condition should be `if v ~= nil then ...` ?

PR #5428, opened 2020-01-10
feat(conf) add status_ssl_cert and status_ssl_cert_key configuration values

bungle:
### Summary

When status endpoint was added it seems it allowed `ssl` flag on listeners but it missed the configuration values for `status_ssl_cert` and `status_ssl_cert_key`. Because there is special treatment for such configurations values (see `admin_ssl_cert`) these should not be implemented
by users as injections (perhaps later if all the others are moved there too).

This makes `status` ssl configurations behave similarly as in `proxy`, `sproxy`, `admin` already does.


kong.conf.default
```diff
@@ -341,6 +341,12 @@
                          # healthiness, and other non-sensitive information
                          # of the current Kong node.
                          #
+                         # Some suffixes can be specified for each pair:
```
hbagdi: Since only one suffix is available for this endpoint, can we rephrase this?

bungle: @hbagdi, I though about it :-), but then something in me said that someone forgets about it will be forgotten when someone adds more, but as you found it right away, that warrants rephrasing :-)

kong/cmd/utils/prefix_handler.lua
```diff
@@ -286,13 +288,23 @@ local function prepare_prefix(kong_config, nginx_custom_template_path)
...
     log.verbose("Admin SSL enabled, no custom certificate set: using default certificate")
-    local ok, err = gen_default_ssl_cert(kong_config, true)
+    local ok, err = gen_default_ssl_cert(kong_config, "admin")
     if not ok then
       return nil, err
     end
     kong_config.admin_ssl_cert = kong_config.admin_ssl_cert_default
     kong_config.admin_ssl_cert_key = kong_config.admin_ssl_cert_key_default
   end
+  if kong_config.status_ssl_enabled and not kong_config.status_ssl_cert and
```
dndx: Style: add a newline before this.

kong/templates/nginx_kong.lua
```diff
@@ -303,6 +303,12 @@ server {
     access_log ${{STATUS_ACCESS_LOG}};
     error_log  ${{STATUS_ERROR_LOG}} ${{LOG_LEVEL}};
 
+> if status_ssl_enabled then
+    ssl_certificate     ${{STATUS_SSL_CERT}};
+    ssl_certificate_key ${{STATUS_SSL_CERT_KEY}};
+    ssl_session_cache   shared:StatusSSL:10m;
```
dndx: This can be set to `1m` to reduce memory usage, since there won't be that many clients for this API.

bungle: I changed it, but memory usage is not when you start Nginx, but when you actually use it.

PR #5429, opened 2020-01-10
fix(router) detect changes on router/plugins iterator while rebuilding

### LINKED ISSUE #5374: Kong throws '404 - no Route matched with those values' because of corrupted Cache
### Summary
When the system under massive traffic, and CRUD routes via Kong's admin API frequently, Kong returns `404 - no Route matched with those values` for a route path that exists at times.
the delete route occasionally returns `500: {"message":"An unexpected error occurred"}`. When we retry to delete - in succeed.
It happens randomly (once a week) and for a while taking the cache to rebuilt - therefore we assume it related to corrupted cache (as described here: https://github.com/Kong/kong/issues/5207)

### Steps To Reproduce
1. Start Kong
2. Load traffic on the system (~ 500 TPS)
3. Create and delete route via Kong's admin API frequently
4. After a while, some of the NGINX workers in Kong will start to return '404 - no Route matched with those values' for paths that exist at times
5. Things start to recover after a while (~40 minutes) or after Kong restart

### Additional Details & Logs
- Kong version: 1.4
- Operating system: CentOS Linux release 7.7.1908
- Database: Cassandra
- Occurs after getting `500: {"message": "An unexpected error occurred"}` from Kong admin API while trying to delete a route (We have a retry mechanism, so whenever we get this error we try again for 5 times with few seconds of waiting between each attempt)
- We install Kongs over K8S cluster (Helm deployments)
- We work with 3 deployments of Kong connected to the same Cassandra:
  1. `ControlPlan` - for all rest API calls
  2. `DataPlan` - for incoming data ingestion (specific rest API which is more commonly used)
  3. `UserExperience` - for all admin operations (like creation and deletion of routes)
- `ControlPlan` and `DataPlan` Kong deployments have 6 pods (replica=6) over K8S, `UserExperience` has 2)
- Kong related config params:
  1. `ControlPlan`:
  ```
    cassandra_schema_consensus_timeout = 100000
    cassandra_repl_factor = 3
    cassandra_data_centers = westus:3,eastus:3
    cassandra_local_datacenter = westus
    db_update_propagation = 2
    db_update_frequency = 15
    mem_cache_size = 1024m
    cassandra_ssl = off
    cassandra_ssl_verify = off
    cassandra_timeout = 35000
    anonymous_reports = off
    cassandra_consistency = LOCAL_QUORUM
    cassandra_lb_policy = RequestDCAwareRoundRobin
    cassandra_repl_strategy = NetworkTopologyStrategy
    router_consistency = strict
    router_update_frequency = 1
  ```
  2. `DataPlan`:
  ```
    cassandra_schema_consensus_timeout = 100000
    cassandra_repl_factor = 3
    cassandra_data_centers = westus:3,eastus:3
    cassandra_local_datacenter = westus
    db_update_propagation = 2
    db_update_frequency = 28
    mem_cache_size = 1024m
    cassandra_ssl = off
    cassandra_ssl_verify = off
    cassandra_timeout = 35000
    anonymous_reports = off
    cassandra_consistency = LOCAL_QUORUM
    cassandra_lb_policy = RequestDCAwareRoundRobin
    cassandra_repl_strategy = NetworkTopologyStrategy
    router_consistency = strict
    router_update_frequency = 1
  ```
  3. `UserExperience`:
  ```
    cassandra_schema_consensus_timeout = 100000
    cassandra_repl_factor = 3
    cassandra_data_centers = eastus:3,westus:3
    cassandra_local_datacenter = westus
    db_update_propagation = 2
    db_update_frequency = 7
    mem_cache_size = 1024m
    cassandra_ssl = off
    cassandra_ssl_verify = off
    cassandra_timeout = 15000
    anonymous_reports = off
    cassandra_consistency = LOCAL_QUORUM
    cassandra_lb_policy = DCAwareRoundRobin
    cassandra_repl_strategy = NetworkTopologyStrategy
    router_consistency = eventual
    router_update_frequency = 1
  ```
- NGINX related config params (for all 3 deployments):
  ```
    lua_shared_dict kong                5m;
    lua_shared_dict kong_db_cache       512m;
    lua_shared_dict kong_db_cache_miss 128m;
    lua_shared_dict kong_locks          8m;
    lua_shared_dict kong_process_events 5m;
    lua_shared_dict kong_cluster_events 5m;
    lua_shared_dict kong_healthchecks   5m;
    lua_shared_dict kong_rate_limiting_counters 12m;
    lua_shared_dict kong_cassandra      5m;
    lua_socket_log_errors off;
  ```
- ### All pods configured with `worker_processes 1;` for NGINX workers
We did it because we assumed there is a collision in shared memory between NGINX workers which causing this problem to appear (mutexes? semaphores?). After reducing the NGINX workers to be one, The problem still exists but occurs way less often (once a week instead of once a day)
- ### There was a complete evolution around this issue (trying to fix it by configuration):
1. First, we set `router_consistency = eventual` for all deployments with a low update frequency-time
2. Then, we change the most frequently used Kongs deployments to be `router_consistency = strict` (`ControlPlan` and `DataPlan`)
3. After, we reduce the amount of NGINX workers to be 4 instead of 8 (Kongs deployments: `ControlPlan` and `in`: 3 pods each, `UserExperience`: 2 pods)
4. At last, we reduce the amount of NGINX workers to be 1 and scale the deployment to be 6 pods for `ControlPlan` and `DataPlan` (portal still have 2 pods)
- We are considering, as a workaround, not deleting any routes from now on, instead - patching them to a not exists path `/dev/null` and once a while deletes all routes with this path. We hope it will keep Kong deployments stable without any occurrence of this issue.
- related issues:
  1. https://github.com/Kong/kong/issues/4055
  2. https://github.com/Kong/kong/issues/5207

bungle:
### Summary

There is a slight edge case reported on #5374. I originally tried to propose a fix to this on my original `async rebuilds` work, but when it was refactored, the fix was dropped out.

The issue is that our `routes:each` / `plugins:each` does actually multiple calls to db, each of which can `yield`, thus making it possible for rows to be added / deleted between different pages of data. This will in some cases make our router invalid.

My original fix used recursive rebuilding, but this new will just terminate the rebuild, which means the router / plugins iterator is not updated, and so the previous one will be used. The next async rebuild or request (non-async) will start a new try,

### Issues resolved

Fix #5374 (I think)

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -618,6 +618,11 @@ do
   end
 
 
+  local function get_router_version()
+    return kong.core_cache:get("router:version", TTL_ZERO, utils.uuid)
+  end
+
+
   build_router = function(version)
     local db = kong.db
     local routes, i = {}, 0
@@ -635,9 +640,22 @@ do
       end
     end
 
+    local counter = 0
+    local page_size = db.routes.pagination.page_size
     for route, err in db.routes:each() do
       if err then
         return nil, "could not load routes: " .. err
+      end
+
+      if kong.core_cache and counter > 0 and counter % page_size == 0 then
+        local new_version, err = get_router_version()
+        if err then
+          return nil, "failed to retrieve router version: " .. err
+        end
+
+        if new_version ~= version then
+          return nil, "router was changed while rebuilding it"
+        end
       end
 
       if should_process_route(route) then
@@ -654,6 +672,8 @@ do
         i = i + 1
         routes[i] = r
       end
+
+      counter = counter + 1
     end
 
     local new_router, err = Router.new(routes)
@@ -679,7 +699,7 @@ do
     -- we might not need to rebuild the router (if we were not
     -- the first request in this process to enter this code path)
     -- check again and rebuild only if necessary
-    local version, err = kong.core_cache:get("router:version", TTL_ZERO, utils.uuid)
+    local version, err = get_router_version()
     if err then
       return nil, "failed to retrieve router version: " .. err
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Table of Contents
 
 
-- [2.0.0](#200)
+- [1.4.3](#143)
 - [1.4.2](#142)
 - [1.4.1](#141)
 - [1.4.0](#140)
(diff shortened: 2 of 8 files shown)


kong/runloop/handler.lua
```diff
@@ -635,11 +639,24 @@ do
...
 
+    local counter = 0
+    local page_size = db.routes.pagination.page_size
     for route, err in db.routes:each() do
       if err then
         return nil, "could not load routes: " .. err
       end
 
+      if kong.core_cache and counter % page_size == 0 then
+        local new_version, err = get_router_version()
```
hishamhm: I think it would be sufficient to only check this once after the loop, right? (before `Router.new`)

(In fact, I wonder if it wouldn't be more correct, even: wouldn't it be possible for the problem to happen even if the number of routers is less than one page size, but a change happens during that single yield, after the DB data has returned but before the scheduler resumed this coroutine?)

hishamhm:
@bungle The approach sounds very sensible! Can you please retarget this PR to `master`? Thanks!

@locao it sounds like the async upstreams PR (#5325) will need something similar?

kong/runloop/handler.lua
bungle: The check is here, after each page, so that we will quit as soon as possible (to avoid doing unneccessary work, and release the request / buid). This code will `yield` on each `:page`, and not yield anywhere else. If we get last page (but the changes where made), it is ok, as we got full page (non-invalid). Similar to single page thing, if we got results from db, that result is ok, even if there were changes. The next request/build will start again.

hishamhm: Ah, ok, got it! So the problem is really that in the case of concurrent modification, the `each` iteration as a whole returns an invalid set, correct?

bungle: Yes, we cannot trust routes/plugins if state in those tables changes beween `:page` calls.

1. if there is a single page we can always trust (i will add a condition for that)
2. if things change when looping last page it doesn't matter

So yes `as a whole`.

dndx:
This change manes sense. One possibly (extreme) edge case is that will a Kong installation that has changes very often causes the router rebuild to fail often and cause noticeable rebuild delays? (this is a very extreme example, but seems to be possible).

bungle:
> This change manes sense. One possibly (extreme) edge case is that will a Kong installation that has changes very often causes the router rebuild to fail often and cause noticeable rebuild delays? (this is a very extreme example, but seems to be possible).

Yes, that will post-pone the change, but I think there are more pros than cons.

PR #5440, opened 2020-01-13
feat(acl) 401 instead of 403 when no credentials found in context

### LINKED ISSUE #4480: ACL plugin: 401 vs 403

bungle:
### Summary

This is an alternative approach to #5277 and #5236.

The background:

`acl` plugin runs before `request-termination` plugin, and `acl` plugin only returned `403`. `403` seems like a correct status code for authenticated consumers as they are not `unauthorized or 401`. And also for anonymous consumers, if the consumer has `acl` groups configured (which I have never seen outside our test suite). But otherwise anonymous consumers should be getting `401`.

This PR tries to do almost the same as #5277 and #5236, but also tries to not touch to tests too much and thus keeping the original behavior intact (there is one test that needed to be changed).

All this work is from inspiration of Kong community, and I (@bungle) was just a tool to polish the idea a bit.

Huge thanks to @jeremyjpj0916, and @odelvalle!

### Issues Resolved

Fix #4480

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -39,7 +39,7 @@ local ACLHandler = {}
 
 
 ACLHandler.PRIORITY = 950
-ACLHandler.VERSION = "2.1.0"
+ACLHandler.VERSION = "3.0.0"
 
 
 function ACLHandler:access(conf)
@@ -65,10 +65,8 @@ function ACLHandler:access(conf)
   if not consumer_id then
     local authenticated_groups = groups.get_authenticated_groups()
     if not authenticated_groups then
-      kong.log.err("Cannot identify the consumer, add an authentication ",
-                   "plugin to use the ACL plugin")
-
-      return kong.response.exit(403, {
+      local status = kong.client.get_credential() and 403 or 401
+      return kong.response.exit(status, {
         message = "You cannot consume this service"
       })
     end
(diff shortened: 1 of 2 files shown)


bungle:
@jeremyjpj0916 I would like your `review` on this. Does this do what you want, or did I miss something?

jeremyjpj0916:
At first glance I believe this does meet the use case! Will give it a test here shortly and confirm.

In limited scope the change for me was if a proxy has auth plugins (oauth/jwt) + ACL plugin but client didn't pass or passed invalid creds then they see a 401 (not a 403 as existing behavior from the ACL plugin) around "anonymous user" pattern. I believe this change will deliver on that desire while keeping older test case behavior intact. 

Nice stuff.

jeremyjpj0916:
Did run a quick test, 

Had a Service + Route where the route had the ACL plugin with a whitelist group set to the route id + OAuth2 + JWT plugin enabled. Where the OAuth2 + JWT are set to a given anonymous user id XXXX (being anon.user consumer). This user has 0 ACL Groups on his account.

Expected Behavior:
See 401 response from Kong in this scenario.

Seen Behavior:
Returned 403 and 
{
    "message": "You cannot consume this service"
}

Which is the same prior behavior I believe. Tried calling it with no Auth header set as well as setting my bearer token to a dummy value.

bungle:
@jeremyjpj0916, you are correct, it seems like `groups` file does return `empty` table in some cases. Let me try to fix it.

bungle:
@jeremyjpj0916 here is my second patch trying to solve this:
#5452

PR #5441, opened 2020-01-13
Multivalue maps

javierguerragiraldez:
### Summary

The Lua style of multivalue maps (used by `get_query()` and `get_headers()` functions) uses different value types to represent multiplicity of types:
- present but no value: boolean true
- one value: string
- multiple values: array of strings

In a static-type language, this results in a very verbose code.  This PR changes the representation into a more regular "map of string arrays", where the "no value" case is an empty array.

related PRs: Kong/go-pdk#19 , Kong/go-pluginserver#12

### DIFF
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -20,6 +20,21 @@ local reset_instances   -- forward declaration
 local preloaded_stuff = {}
 
 
+-- add MessagePack empty array/map
+
+msgpack.packers['function'] = function (buffer, f)
+  f(buffer)
+end
+
+local function mp_empty_array(buffer)
+  msgpack.packers['array'](buffer, {}, 0)
+end
+
+local function mp_empty_map(buffer)
+  msgpack.packers['map'](buffer, {}, 0)
+end
+
+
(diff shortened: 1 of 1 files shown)


kong/db/dao/plugins/go.lua
```diff
@@ -185,6 +228,37 @@ local get_field
...
+    end,
+
+    ["kong.nginx.get_var"] = function(v)
+      return ngx.var[v]
+    end,
+
+    ["kong.nginx.get_tls1_version_str"] = ngx_ssl.get_tls1_version_str,
+
+    ["kong.nginx.get_ctx"] = function(v)
+      return ngx.ctx[v]
```
gszr: We should instead be using `kong.ctx` as that contains extra logic - eg, for `kong.ctx.shared`.

hishamhm: I originally added this `kong.nginx.get_ctx` in the Go PDK as a way to access directly the raw `ngx.ctx` because my initial line-by-line port of `file-log` needed it (when the equivalent of the all-in-one `serialize` function was implemented in Go itself). I think `kong.ctx` should be eventually exposed as `kong.ctx` instead, so I think this is good as is.

PR #5450, opened 2020-01-15
docs(changelog) cache split entry in "Additions"

locao:


gszr:
Agreed with @dndx; building on top of his suggestion. Should we also call out the changes in the template explicitly, for users that have a custom template?

=====
The internal cache was split into two independent segments,
kong.core_cache and kong.cache. The core_cache is used by the
Kong core to store configuration data that doesn't change often. The other
region is used to store plugin runtime data that is dependent on
traffic pattern and user behavior. This change should decrease the cache
contention between Kong core and plugins and result in better performance
overall.
⚠️ Note that both structures rely on the already existent
`mem_cache_size configuration` option to set their size, so when upgrading from a previous Kong version, the cache memory consumption might double if this value is not adjusted.
#5114

**If you are using a custom Nginx template, make sure core cache shared dictionaries are defined, including db-less mode shadow definitions:**

```
lua_shared_dict kong_core_db_cache          ${{MEM_CACHE_SIZE}};
lua_shared_dict kong_core_db_cache_miss     12m;

> if database == "off" then
lua_shared_dict kong_core_db_cache_2        ${{MEM_CACHE_SIZE}};
lua_shared_dict kong_core_db_cache_miss_2   12m;
lua_shared_dict kong_db_cache_2             ${{MEM_CACHE_SIZE}};
lua_shared_dict kong_db_cache_miss_2        12m;
> end
```

=====

PR #5451, opened 2020-01-15
Allowing colon character in tag

marcelometal:


hishamhm:
@fffonion do you remember if there was a technical reason why `:` was disallowed?

@marcelometal could you add in this PR some test to our test suite exercising tags with `:`?

marcelometal:
Hi @hishamhm , I didn't find any tag tests for "~", "-", etc,  only for name field.

bungle:
This is what `k8n` uses, it seems our `~` is already not supported by that. I guess if we divert too much, it will just mean breakage in integrations.

bungle:
@marcelometal, can you describe your use case a bit?

marcelometal:
any news about it?

hbagdi:
ping @fffonion @thibaultcha  for any technical reason to not allow for this.

The product concern here would be that this might open up flood gates for feature requests like running plugins when an entity contains a specific tag (key:<value> or key:<regex-value>) or making it possible to filter based on tag with regex.

fffonion:
Sorry i missed this thread, it was an IRL discussion with @bungle that we started with a most strict range, i.e. to be the same characters set with others like `name`:
https://github.com/Kong/kong/blob/28e4baf323a3b01ede1be4301fd490408a1e5d6c/kong/db/schema/typedefs.lua#L100-L110
In other words I don't think technically there's an objection to add colon character here.

kikito:
Hello, thank you for sending this PR and apologies for the time it took us to get back to you on it.

We have decided to go on a separate direction. It does what you want (accept colon in tags) and expands this to (most of) the whole utf-8 range. Please see #6784 for details.

As a result, I am closing this PR.

PR #5452, opened 2020-01-16
fix(acl) second try fixing acl plugin to return 401

### LINKED ISSUE #4480: ACL plugin: 401 vs 403

bungle:
### Summary

The PR #5440 tried to fix the issue with ACL plugin, but it didn't do it completely. As seen here:
https://github.com/Kong/kong/pull/5440#issuecomment-574720075

This is a second try to fix it so that anonymous consumers that do not have ACL groups will get `401` status code instead of `403`.

### Issues Resolved

Fix #4480 (again)

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -65,9 +65,14 @@ function ACLHandler:access(conf)
   if not consumer_id then
     local authenticated_groups = groups.get_authenticated_groups()
     if not authenticated_groups then
-      local status = kong.client.get_credential() and 403 or 401
-      return kong.response.exit(status, {
-        message = "You cannot consume this service"
+      if kong.client.get_credential() then
+        return kong.response.exit(403, {
+          message = "You cannot consume this service"
+        })
+      end
+
+      return kong.response.exit(401, {
+        message = "Unauthorized"
       })
     end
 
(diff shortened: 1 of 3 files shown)


bungle:
@jeremyjpj0916, could you try this on top of the previous one?

jeremyjpj0916:
@bungle Testing now, will report back in 30-45 mins.

jeremyjpj0916:
@bungle 

Tested again with the same use case as before, this time it did 401 as desired. Thanks!

bungle:
@jeremyjpj0916,

I think I need to fix one thing still, the error message is currently:
```
You cannot consume this service
```

But I think the normal:
```
Unauthorized
```

Is better in this case? Do you have opinion?

jeremyjpj0916:
@bungle I didn't wanna be too picky 😆  , but yes I think I ```Unauthorized``` is more appropriate in this situation.

bungle:
Ok, changed the message on `401` to be `Unauthorized` and updated the test suite.

kong/plugins/acl/handler.lua
```diff
@@ -99,10 +104,20 @@ function ACLHandler:access(conf)
...
       if not consumer_groups then
-        local status = kong.client.get_credential() and 403 or 401
-        return kong.response.exit(status, {
-          message = "You cannot consume this service"
-        })
+        if config.type == BLACK and kong.client.get_credential() then
+          consumer_groups = EMPTY
+
+        else
+          if kong.client.get_credential() then
```
kikito: This code calls `kong.client.get_credential()` twice. LuaJIT will probably optimize this up, but let's localize it just in case.

PR #5457, opened 2020-01-16
feat(conf) support for multi-value nginx conf injections

### LINKED ISSUE #4723: Nginx directive injection does not work for multiple directives with same name
### Summary
Injection of multiple nginx directives with same name does not work. For example if I want to create multiple lua_shared_dict directives, only the last one will be injected.

### Steps To Reproduce

1. Add directives that need injection to kong.conf e.g.
```
nginx_http_lua_shared_dict=cache1 100k
nginx_http_lua_shared_dict=cache2 100k
nginx_http_lua_shared_dict=cache3 100k
```
2. Restart kong
3. Only the last directive is injected in nginx-kong.conf
```
# injected nginx_http_* directives
lua_shared_dict cache3 100k;
```

### Additional Details & Logs

- Kong version (1.1.2)

bungle:
### Summary

Previously you could only inject single directive with an injections like:

```
KONG_NGINX_HTTP_LUA_SHARED_DICT="my_shm 10m" ./bin/kong prepare
```

And this would inject a following line in `nginx-kong.conf`:

```
lua_shared_dict my_shm 10m;
```

But what if you wanted to inject more than one `lua_shared_dict`:

Option 1:

```
KONG_NGINX_HTTP_INCLUDE="my.conf" ./bin/kong prepare
```

Where contents of `my.conf` is:
```
lua_shared_dict my_shm 10m;
lua_shared_dict my_shm_2 20m;
```

Option 2:

Use a custom Kong Nginx template.

While those are workarounds, it would be nice to just specify a multi-valued injection, and this PR adds that:

```
KONG_NGINX_HTTP_LUA_SHARED_DICT="my_shm 10m, my_shm_2 20m" ./bin/kong prepare
```

And this would add two directives:
```
lua_shared_dict my_shm 10m;
lua_shared_dict my_shm_2 20m;
```

If you need to use comma `,` in actual value, you can:
1. escape comma with `\` (the escape char will be included in directive value too):
```
KONG_NGINX_HTTP_MESSAGE="Hello\, World!" ./bin/kong prepare
```

2. add comma inside single or double quotes, e.g. (quotes are kept in directive value too):
```
KONG_NGINX_HTTP_MESSAGE="'Hello, World!'" ./bin/kong prepare
```

(I know there is no such directive `message`)

### Issues Resolved

Fix #4723

### DIFF
--- a/kong/conf_loader.lua
+++ b/kong/conf_loader.lua
@@ -1321,10 +1321,87 @@ local function load(path, custom_conf, opts)
     end
   end
 
-  for _, dyn_namespace in ipairs(DYNAMIC_KEY_NAMESPACES) do
-    table.sort(conf[dyn_namespace.injected_conf_name], function(a, b)
-      return a.name < b.name
-    end)
+  do
+    local sub  = string.sub
+    local byte = string.byte
+    local find = string.find
+
+    local COMMA        = byte(",")
+    local SINGLE_QUOTE = byte("'")
+    local DOUBLE_QUOTE = byte('"')
+    local BACKSLASH    = byte("\\")
+
+    for _, dyn_namespace in ipairs(DYNAMIC_KEY_NAMESPACES) do
+      local directives = conf[dyn_namespace.injected_conf_name]
+      local directive_count = #directives
+      for i = 1, directive_count do
+        local directive = directives[i]
+        local value = directive.value
+        if find(value, ",", 1, true) then
+          local escaped = false
+          local single_quoted = false
+          local double_quoted = false
+          local length = #value
+          local start_position = 1
+          local is_first = true
+          for j = 1, length do
+            local b = byte(value, j)
+            if b == BACKSLASH then
+              escaped = not escaped
+
+            else
+              if not escaped then
+                if b == SINGLE_QUOTE then
+                  single_quoted = not single_quoted
+
+                elseif b == DOUBLE_QUOTE then
+                  double_quoted = not double_quoted
+
+                elseif b == COMMA and not single_quoted and not double_quoted then
+                  local v = utils.strip(sub(value, start_position, j - 1))
+                  if is_first then
+                    directive.value = v
+                    is_first = false
+
+                  else
+                    table.insert(directives, {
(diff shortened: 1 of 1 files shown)


hbagdi:
This looks nice on the surface but I'd recommend against shipping such a feature.

When we added Nginx directive injections, the primary motivation was to make it easy for users to modify Nginx template in a way that works across release of Kong. This was cumbersome with custom Nginx templates.

One of the reasons I feel this feature has been so successful is because "it just works". There is no much to it and it largely does what you would expect it to -- just adds in strings into the Nginx configuration file. It is very intuitive in it's current form and I feel that this feature is going to make it a little complicated with the escapes kicking in.

This feature also lays the ground to add in more fancy things that might come in the future (I don't know what else is possible next, a lot is possible in Nginx's configuration file DSL). Our stand around these things has been to use the `include` directive for more complicated things, which is justified because: a) any user who is doing this, clearly understands Kong well enough b) this is easily compatible across releases.

Last and the most important point, amount of users who need this functionality is far too small to warrant adding it at this point when it is possible via other safer means.

bungle:
@luka1983, @c4m4, @fusi0011, let us understand better why would you like to see this feature rather than using above work-arounds?

In some ways I feel `include` feels like a `hammer` for every problem. This commit doesn't make it that much more complicated (we don't do anything, unless you really have a comma there, and I have never seen comma's used in Nginx directive values except perhaps error log formats).

luka1983:
I solved the problem by creating the custom Nginx template since I was not aware of KONG_NGINX_HTTP_INCLUDE directive. I raised the issue since it seemed to be a logical way to inject multiple directives of same type, i.e. that there is one-on-one mapping between directive injections in kong.conf file with the resulting directives in nginx-kong.conf. In my case the directive injection feature only introduced the confusion, since it seemed to me it did the job only partially, and in the end I did not use it at all. But then again, I only did a single Kong installation which was probably a bit specific.

Regarding the proposed feature, I like it but I also suppose there is a potential here to break a number of config files out there in the wild.

PR #5459, opened 2020-01-16
fix(pdk) kong.ctx.plugin.* may be flaky

bungle:
### Summary

It came to my knowledge that our users have evidenced a following problem:
```
function plugin:access(conf)
  kong.ctx.plugin.key = "value"
  kong.ctx.shared.cfg = tostring(conf)
end

function plugin:header_filter(conf)
  if not kong.ctx.plugin.key then
    -- the bug!
  end

  local cfg2 = tostring(conf)
  if kong.ctx.shared.cfg ~= cfg2 then
    -- also illustrates reason for a bug!
  end
end
```

Before running the plugin (before this PR) we called this:
```
kong_global.set_named_ctx(kong, "plugin", conf)
```

So the `conf` table ended up as a key in `ngx.ctx`, but there are
cases when this `conf` table may change, I believe, between say
`access` and `header_filter` phases. Possible reasons being:

1. new plugin iterator being built between the phases
2. conf table gets cache evicted, and appears again as a new table
3. **plugin config change between phases** (e.g. plugin mapped to `route` changes to a plugin mapped to consumer)

This PR changes the problematic calls to something more stable:
```
kong_global.set_named_ctx(kong, "plugin", plugin.handler)
```

The `handler` should stay the same as long as Kong is running.

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -184,7 +184,7 @@ end
 local function execute_plugins_iterator(plugins_iterator, phase, ctx)
   for plugin, configuration in plugins_iterator:iterate(phase, ctx) do
     if ctx then
-      kong_global.set_named_ctx(kong, "plugin", configuration)
+      kong_global.set_named_ctx(kong, "plugin", plugin.handler)
     end
 
     kong_global.set_namespaced_log(kong, plugin.name)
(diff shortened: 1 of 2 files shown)


bungle:
> This is a valid issue, but the solution implemented in this PR does not match the documented semantics of `kong.ctx.plugin`

You are not correct as far as I understand this.

On single request there can be only single `rate-limiting` plugin, and context is not shared with other requests, so my change is according to that description.

hishamhm:
> On single request there can be only single rate-limiting plugin, and context is not shared with other requests, so my change is according to that description.

`ngx.ctx` is not shared across requests, but `plugin.handler` is a VM-local variable and will be shared across multiple requests being handled by different coroutines.

PR #5460, opened 2020-01-16
feat(rate-limiting) remove timer and phase handlers

### LINKED ISSUE #5311: Ratelimit plugin not working as expected - kong 1.4
### Summary
I have configured rate-limit plugin with key-auth. 5 req/sec is working fine if requests are more then 50 per seconds its not detecting. i have tried out with cluster/local/redis policy. 

SUMMARY_GOES_HERE

### Steps To Reproduce
kong 1.4 . cassandra backend. deploy on kubernets
1. enable key-auth plugin
2. enable rate limit plugin with 5 req/sec 
3. try with apache benchmark ab -c 100 -n 1000
4. check live kong logs

### Additional Details & Logs

- Kong version (`$ kong version`) v1.4
- Operating system

bungle:
### Summary

Context:

- starting `timers` on busy servers on each request will just move  the load elsewhere
- there is little chance that upstream sets rate-limiting headers, so we can just set them on access

So this PR removes timers and all the phase handlers except `access` from rate-limiting plugin. Yes, it might add some latency to request, but on a bright side we have more accurate rate-limits.

### Issues Resolved

This might fix #5311

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -11,7 +11,6 @@ local floor = math.floor
 local pairs = pairs
 local error = error
 local tostring = tostring
-local timer_at = ngx.timer.at
 
 
 local EMPTY = {}
@@ -107,15 +106,6 @@ local function get_usage(conf, identifier, current_timestamp, limits)
 end
 
 
-local function increment(premature, conf, ...)
-  if premature then
-    return
-  end
-
-  policies[conf.policy].increment(conf, ...)
-end
-
-
 function RateLimitingHandler:access(conf)
   local current_timestamp = time() * 1000
 
@@ -145,8 +135,9 @@ function RateLimitingHandler:access(conf)
   if usage then
     -- Adding headers
     local reset
+    local headers
     if not conf.hide_client_headers then
-      local headers = {}
+      headers = {}
       local timestamps
       local limit
       local window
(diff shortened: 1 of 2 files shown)


bungle:
> One unintended side effect found when testing: currently, we always add headers, even if the request is terminated. As-is, the PR moves adding headers into access after terminating, so terminated requests don't get those headers. To preserve existing behavior we probably want:
> 
> ```lua
>     if headers then
>       kong.response.set_headers(headers)
>     end
> 
>     -- If limit is exceeded, terminate the request
>     if stop then
>       headers[RETRY_AFTER] = reset
>       return kong.response.exit(429, { message = "API rate limit exceeded" }, headers)
>     end
> ```
> 
> Rather than:
> 
> ```lua
>     -- If limit is exceeded, terminate the request
>     if stop then
>       headers[RETRY_AFTER] = reset
>       return kong.response.exit(429, { message = "API rate limit exceeded" }, headers)
>     end
> 
>     if headers then
>       kong.response.set_headers(headers)
>     end
> ```

The kong.response.exit does set the headers, so we don’t need to call that separately.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -172,38 +163,22 @@ function RateLimitingHandler:access(conf)
...
-      kong.ctx.plugin.headers = headers
     end
 
     -- If limit is exceeded, terminate the request
     if stop then
-      return kong.response.exit(429, { message = "API rate limit exceeded" }, {
-        [RETRY_AFTER] = reset
-      })
+      headers[RETRY_AFTER] = reset
+      return kong.response.exit(429, { message = "API rate limit exceeded" }, headers)
```
kikito: If I'm understanding this correctly, by reusing the `headers` variable we're sending more headers than before when the rate limit is exceeded. Before it was just `RETRY_AFTER`, now it's that plus any others already in the `headers` var (like `RATELIMIT_LIMIT`).

This sounds like a change (non-breaking, probably, but a change). Is this desirable / accounted for?

bungle: @kikito we are not sending more, In previous version it sent the the rest in `header_filter`.

hishamhm:
> there is little chance that upstream sets rate-limiting headers, so we can just set them on access

If Kong is being used as an edge gateway in front of a mesh where the sidecars themselves protect their services with internal rate-limits, we would end up propagating the mesh's rate-limit headers rather than the edge gateway's headers. That doesn't sound too unrealistic, right?

bungle:
> If Kong is being used as an edge gateway in front of a mesh where the sidecars themselves protect their services with internal rate-limits, we would end up propagating the mesh's rate-limit headers rather than the edge gateway's headers. That doesn't sound too unrealistic, right?

Well, I guess not. This is general problem with a lot of our plugins.
is it correct to override upstream headers in `header_filter`

So think our plugin will let the request pass. But the upstream denies it. Is it correct to return upstream headers or do we want to override the headers.

E.g. upstream says: ratelimit exceeded, we say you still have this amount of requests left.

locao:
@bungle @hishamhm are we still considering to add this feature?

bungle:
> @bungle @hishamhm are we still considering to add this feature?

I think it should be merged.

mySingleLive:
When can this PR be merged & released? We find the same problem.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -182,38 +173,23 @@ function RateLimitingHandler:access(conf)
...
-  if headers then
-    kong.response.set_headers(headers)
-  end
-end
-
-
-function RateLimitingHandler:log(_)
-  if kong.ctx.plugin.timer then
-    kong.ctx.plugin.timer()
+  local _, err = policies[conf.policy].increment(conf, limits, identifier, current_timestamp, 1)
```
kikito, line 190: I agree with removing the timer, but wouldn't it make more sense to keep the increment in the log phase?

As I understand it, putting the increment in `access` will hold the request until after the increment is done (possibly several database access, if we're using cluster policy). The log phase happens when the response has already been sent back to the client so it shouldn't be penalized in that manner.

bungle:
The #6802 contains most of this PR, except timer removal, so it is still a slightly less accurate.

PR #5464, opened 2020-01-17
fix(declarative) schema validation in db-less mode

### LINKED ISSUE #5401: Schema validation in db-less mode Kong 1.4
### Summary

Schema validation silently ignored when starting Kong in db-less mode.

### Steps To Reproduce

1. Write db-less config with any plugin with missed required field.
2. Start Kong in DB-less mode

Kong 1.3.x failed to start, the error log contains an appropriate message.

Kong 1.4.x silently start, no schema validation logs present.

### Additional Details & Logs

- Kong version official Docker images kong:1.4.x-alpine and kong:1.3.x-alpine

bungle:
### Summary

This was reported with #5401. When I added DAO transformations, for some reason I ended up writing stupid code that just broke the whole schema validation on declarative. This commit is an
attempt to fix that. Sorry for all the pain caused.

While fixing this, I also found out that original validation didn't work correctly when you had entity validations that relied on values of foreign keys. This is now also corrected.

### Issues Resolved

Fix #5401

### DIFF
--- a/kong/db/schema/others/declarative_config.lua
+++ b/kong/db/schema/others/declarative_config.lua
@@ -121,8 +121,12 @@ local function add_top_level_entities(fields, entities)
 end
 
 
-local function copy_without_foreign(record)
+local function copy_record(record, include_foreign)
   local copy = utils.deep_copy(record, false)
+  if include_foreign then
+    return copy
+  end
+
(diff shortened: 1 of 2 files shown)


bungle:
@hishamhm, just a reminder: we want this to both `1.5.0` AND `2.0.0`

PR #5495, opened 2020-01-23
fix(proxy) hop-by-hop headers to not clear upgrade header on upgrade

### LINKED ISSUE #5465: Websocket connections not working in firefox with kong >= 1.4.2
### Summary

We updated kong in some of our environments from 1.3 to 1.4.3 (also tried 1.4.2 after it failed). After the update everything seemed fine in chrome but in firefox the websocket connections could not be established. We are using the most recent version of anycable-go as our websocket server.

The error in the browser:
```
Connection to wss://my.url.com was interrupted while the page was loading.
```
Could not find any errors in the kong log

### Additional Details & Logs
- Kong version (`1.4.2` and `1.4.2`)

**EDIT:**
I also tried with 2.0.0rc2 and I got the same result. For some reason this bug only appears in two of our four environments. Kong is configured the same in all of the environments. 

The error messages in firefox:
```
The connection to wss://sth.example.com/cable?chatbot_name=Support was interrupted while the page was loading. js.js:29:9219
Firefox can’t establish a connection to the server at wss://sth.example.com/cable?chatbot_name=Support. js.js:29:9219
```

This is how requests from chrome appear in the log.
```
83.175.84.210 - - [21/Jan/2020:17:04:12 +0000] "GET /cable?chatbot_name=Support HTTP/1.1" 101 100 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.117 Safari/537.36"
2020/01/21 17:04:13 [debug] 22#0: *17553 [lua] init.lua:823: balancer(): setting address (try 1): 10.42.64.96:3000
```
The one from firefox don't show up in the logs.

bungle:
### Summary

`HTTP` `Connection` header usually carries a list of `headers` meant for receiver (in this case proxy) that are not meant to be proxied further (so called hop-by-hop headers). Our gateway correctly cleans these headers, but it was a bit too aggressive.

It turns out as reported on #5465 that websockets upgrade mechanism is broken with Firefox, but works on Chrome. The difference being that Chrome sends following headers:

```http
Connection: Upgrade
Upgrade: websocket
```

while Firefox sends the following:

```http
Connection: keep-alive, Upgrade
Upgrade: websocket
```

As the `Upgrade` was listed on `Connection` header by Firefox, the Kong incorrectly removed the `Upgrade` header from the upstream headers, and thus it made the websockets upgrade to fail. Chrome had only the `Upgrade` header and that was already handled correctly.

This PR fixed it so that we do not remove `Upgrade` header on such case.

This PR also adds `keep-alive` to upstream request `Connection` header when proxying the upgrade request (just like Firefox does).

### Issues Resolved

Fix #5465

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1117,7 +1117,7 @@ return {
 
       -- Keep-Alive and WebSocket Protocol Upgrade Headers
       if var.http_upgrade and lower(var.http_upgrade) == "websocket" then
-        var.upstream_connection = "upgrade"
+        var.upstream_connection = "keep-alive, Upgrade"
         var.upstream_upgrade    = "websocket"
 
       else
@@ -1198,9 +1198,14 @@ return {
       -- clear hop-by-hop request headers:
       for _, header_name in csv(var.http_connection) do
         -- some of these are already handled by the proxy module,
-        -- proxy-authorization being an exception that is handled
-        -- below with special semantics.
-        if header_name ~= "proxy-authorization" then
+        -- proxy-authorization and upgrade being an exception that
+        -- is handled below with special semantics.
+        if header_name == "upgrade" then
+          if var.upstream_connection == "keep-alive" then
+            clear_header(header_name)
+          end
+
+        elseif header_name ~= "proxy-authorization" then
           clear_header(header_name)
         end
       end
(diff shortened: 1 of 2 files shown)


provgregoryabdo:
Just wondering if this will be put into a 1.4 release? I run kong in kube using helm and that is still pinned to 1.4.

Thanks!

PR #5497, opened 2020-01-24
fix(ldap-auth) credential rate-limiting and hashing of cache key

### LINKED ISSUE #4129: Same Rate Limit counting is used by two different LDAP users
Hi, it's about `LDAP Authentication` plugin together with `Rate Limiting` plugin set in a same Route. Rest API calls bellow

```
curl -i -X POST http://localhost:8001/routes/$PRIVATE_ROUTE_ID/plugins \
  -d "name=ldap-auth" \
  -d "config.hide_credentials=true" \
  -d "config.ldap_host=ldap-host" \
  -d "config.ldap_port=389" \
  -d "config.start_tls=false" \
  -d "config.base_dn=ou=users,dc=mycompany,dc=com" \
  -d "config.verify_ldap_host=false" \
  -d "config.attribute=cn" \
  -d "config.cache_ttl=60" \
  -d "config.header_type=ldap"
```

```
curl -i -X POST http://localhost:8001/routes/$PRIVATE_ROUTE_ID/plugins \
  -d "name=rate-limiting"  \
  -d "config.minute=5"
```

The `LDAP Authentication` plugin configuration is working fine, authenticating correctly the users. The problem is: the same Rate Limit counting is used by two different LDAP users.

For instance, let's suppose we have two LDAP users, user A and B; and a rate limiting configuration for route R of maximum 5 request in a minute. So, if user A calls R for the 1st time, the rate limit headers returned will be
```
X-RateLimit-Limit-Minute: 5
X-RateLimit-Remaining-Minute: 4
```

Then, if user B calls route R (in the same minute), the headers returned are
```
X-RateLimit-Limit-Minute: 5
X-RateLimit-Remaining-Minute: 3
```

So, they are sharing the same limits.

According to `Rate Limiting` plugin documentation (https://docs.konghq.com/hub/kong-inc/rate-limiting/), it says

"Rate limit how many HTTP requests a developer can make in a given period of seconds, minutes, hours, days, months or years. **If the underlying Service/Route** (or deprecated API entity) **has no authentication layer, the Client IP address will be used, otherwise the Consumer will be used if an authentication plugin has been configured.**"

Is it a correct behaviour? Two or more LDAP users are seeing as a same consumer?

Best regards,

- Kong version (`0.14.1` and `1.0.0`)

bungle:
#### fix(ldap-auth) credential to have id for rate-limiting

Currently `ldap-auth` does set virtual credential in context on authentication, but that virtual credential does not have `id` which means that rate-limiting plugin cannot rate-limit by the credential, and thus fall-backs to rate-limit by ip.

This commit adds `cache_key` as id for that virtual credential.

#### fix(ldap-auth) hash cache key

Currently `ldap-auth` does seem to use `username` and `password` in a `cache_key` in plain. This is potentially dangerous and may be leaked through it to different places:

- memory
- admin api
  - logs

This commit fixes that by calling `sha1_bin` on that when generating a cache key. It also drops the use of `md5`.

### Issues resolved

Fix #4129

### DIFF
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -103,6 +103,22 @@ local function ldap_authenticate(given_username, given_password, conf)
   return is_authenticated, err
 end
 
+
+local function cache_key(conf, username, password)
+  if not ldap_config_cache[conf] then
+    ldap_config_cache[conf] = md5(fmt("%s:%u:%s:%s:%u",
+                                      lower(conf.ldap_host),
+                                      conf.ldap_port,
+                                      conf.base_dn,
+                                      conf.attribute,
+                                      conf.cache_ttl))
+  end
+
+  return fmt("ldap_auth_cache:%s:%s:%s", ldap_config_cache[conf],
+             username, password)
+end
+
+
 local function load_credential(given_username, given_password, conf)
   local ok, err = ldap_authenticate(given_username, given_password, conf)
   if err ~= nil then
@@ -117,22 +133,11 @@ local function load_credential(given_username, given_password, conf)
     return false
   end
 
-  return { username = given_username, password = given_password }
-end
-
-
-local function cache_key(conf, username, password)
-  if not ldap_config_cache[conf] then
-    ldap_config_cache[conf] = md5(fmt("%s:%u:%s:%s:%u",
-                                      lower(conf.ldap_host),
-                                      conf.ldap_port,
-                                      conf.base_dn,
-                                      conf.attribute,
-                                      conf.cache_ttl))
-  end
-
-  return fmt("ldap_auth_cache:%s:%s:%s", ldap_config_cache[conf],
-             username, password)
+  return {
+    id = cache_key(conf, given_username, given_password),
+    username = given_username,
+    password = given_password,
+  }
 end
 
 
@@ -151,6 +156,7 @@ local function authenticate(conf, given_credentials)
     kong.log.err(err)
     return kong.response.exit(500, { message = "An unexpected error occurred" })
   end
+
 
   return credential and credential.password == given_password, credential
 end
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -5,6 +5,8 @@ local ldap = require "kong.plugins.ldap-auth.ldap"
 
 local kong = kong
 local decode_base64 = ngx.decode_base64
+local sha1_bin = ngx.sha1_bin
+local to_hex = require "resty.string".to_hex
 local tostring =  tostring
 local match = string.match
 local lower = string.lower
@@ -13,14 +15,10 @@ local find = string.find
 local sub = string.sub
 local fmt = string.format
 local tcp = ngx.socket.tcp
-local md5 = ngx.md5
 
 
 local AUTHORIZATION = "authorization"
 local PROXY_AUTHORIZATION = "proxy-authorization"
-
-
-local ldap_config_cache = setmetatable({}, { __mode = "k" })
 
 
 local _M = {}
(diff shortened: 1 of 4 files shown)


bungle:
Let me know if this can be rebased to `master` even when it does change the `cache_key`.

kong/plugins/ldap-auth/access.lua
```diff
@@ -105,17 +106,20 @@ end
...
+                 conf.attribute,
+                 conf.cache_ttl)
+
+    ldap_config_cache[conf] = prefix
   end
 
-  return fmt("ldap_auth_cache:%s:%s:%s", ldap_config_cache[conf],
-             username, password)
+  local suffix = fmt("%s:%s", prefix, username or "")
+  return fmt("ldap_auth_cache:%s", to_hex(hmac_sha1(password or "", suffix)))
```
kikito: I think applying hmac_sha1 to everything (including the prefix) will make debugging problems in the cache more difficult. I suggest we only encript the username and password and leave the other values unencrypted. Something like this:

```
local function cache_key(conf, username, password)
  local prefix = ldap_config_cache[conf]
  if not prefix then
    prefix = fmt("ldap_auth_cache:%s:%u:%s:%s:%u:",
                 lower(conf.ldap_host),
                 conf.ldap_port,
                 conf.base_dn,
                 conf.attribute,
                 conf.cache_ttl)
    ldap_config_cache[conf] = prefix
  end

  return prefix .. to_hex(hmac_sha1(password or "", username or ""))
end
```

Note that I also moved `ldap_auth_cache` to the prefix there. This will reduce string concatenation and function calls at the expense of more space taken on the cache. I don't know which one is preferable.

spec/03-plugins/20-ldap-auth/01-access_spec.lua
```diff
@@ -1,22 +1,23 @@
...
-  return fmt("ldap_auth_cache:%s:%s:%s", prefix, username, password)
+  local prefix = fmt("%s:%u:%s:%s:%u",
+                     lower(conf.ldap_host),
+                     conf.ldap_port,
+                     conf.base_dn,
+                     conf.attribute,
+                     conf.cache_ttl)
+
+  local suffix = fmt("%s:%s", prefix, username or "")
+  return fmt("ldap_auth_cache:%s", to_hex(hmac_sha1(password or "", suffix)))
```
kikito: Same suggestion as before: encrypt only password and suffix.

spec/03-plugins/20-ldap-auth/02-invalidations_spec.lua
```diff
@@ -61,24 +62,23 @@ for _, ldap_strategy in pairs(ldap_strategies) do
...
-                     username, password)
+          local prefix = fmt("%s:%u:%s:%s:%u",
+                             lower(conf.ldap_host),
+                             conf.ldap_port,
+                             conf.base_dn,
+                             conf.attribute,
+                             conf.cache_ttl)
+
+          local suffix = fmt("%s:%s", prefix, username or "")
+          return fmt("ldap_auth_cache:%s", to_hex(hmac_sha1(password or "", suffix)))
```
kikito: same as before

kong/plugins/ldap-auth/access.lua
p0pr0ck5: You do not need an hmac here. Obfuscating the password into a sha of the composite elements of the cache key should suffice. HMAC is used as an authentication mechanism to ensure a message hasn’t been tampered with in flight. 

If the hmac key isn’t a secret value, an hmac is the wrong construction.

bungle: @kikito,

There is one reason to that:
makes `cache_key` always the same length and smaller, that has perfomance benefits

bungle: > If the hmac key isn’t a secret value, an hmac is the wrong construction.

Yes, I know, just wanted to be collision free, but I can change it to sha1.

bungle: @p0pr0ck5, the `hmac` is now changed to `sha1_bin`, I guess it is good enough, or should I go with `sha256`?

kong/plugins/ldap-auth/access.lua
```diff
@@ -103,6 +104,25 @@ local function ldap_authenticate(given_username, given_password, conf)
...
+                 lower(conf.ldap_host),
+                 conf.ldap_port,
+                 conf.base_dn,
+                 conf.attribute,
+                 conf.cache_ttl)
+
+    ldap_config_cache[conf] = prefix
+  end
+
+  local suffix = fmt("%s:%s:%s", prefix, username or "", password or "")
```
p0pr0ck5: Are there known cases where these variables are nil, and those we need the `or “”`? Or just being defensive?

bungle: Perhaps not, just being defensive, I can remove them.

kong/plugins/ldap-auth/access.lua
p0pr0ck5: I think sha1 is sufficient here 👍

PR #5498, opened 2020-01-24
feat(rate-limiting) support cleanup based on ttl with postgres

### LINKED ISSUE #1994: [plugin] enable auto-clean on ratelimiting plugins
Over time the database grows, an automated cleanup process could be added to clean up the stale entries.

Maybe a generic `maintenance` hook could be added to the plugins, such that it can be controlled properly (in case it is time consuming and/or takes a lot of resources).

bungle:
### Summary

This updates rate-limiting code for Postgres to auto-cleanup expired metrics using our generic `ttl` mechanism.

For Cassandra this could not be implemented as Cassandra doesn't support `USING TTL` with `counter` columns.

This also refactors the cluster policies to use our DAOs `select` as the plugin's `daos.lua` was updated to generate the DAOs for `ratelimiting_metrics` too.

### Issues Resolved

Partially (Postgres) fixes #1994


kong/plugins/rate-limiting/policies/cluster.lua
```diff
@@ -9,6 +9,27 @@ local floor = math.floor
 local fmt = string.format
 
 
+local EXPIRATION = require "kong.plugins.rate-limiting.expiration"
+
+
+local find
+do
+  local FIND_PK = {}
```
kikito: This is to avoid doing one table allocation per call to `find`, right? I think that's ok, but the uppercase name suggests `FIND_PK` is a constant.

bungle: changed to lowercase

PR #5500, opened 2020-01-24
fix(oauth2) hash cache key

bungle:
### Summary

Currently `oauth2` plugin does seem to use `access token` in a `cache_key` in plain. This is potentially dangerous and may be leaked through it to different places:
- memory
- admin api
  - logs

This commit fixes that by calling `sha1_bin` on that when generating a cache key.

### DIFF
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -15,6 +15,8 @@ local check_https = utils.check_https
 local encode_args = utils.encode_args
 local random_string = utils.random_string
 local table_contains = utils.table_contains
+local sha1_bin = ngx.sha1_bin
+local to_hex = require "resty.string".to_hex
 
 
 local ngx_decode_args = ngx.decode_args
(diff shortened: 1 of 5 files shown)


kong/plugins/oauth2/access.lua
```diff
@@ -605,7 +607,8 @@ local function retrieve_token(conf, access_token)
   local token, err
 
   if access_token then
-    local token_cache_key = kong.db.oauth2_tokens:cache_key(access_token)
+    local key = to_hex(sha1_bin(access_token))
+    local token_cache_key = kong.db.oauth2_tokens:cache_key(key)
```
p0pr0ck5: does this properly handle cache invalidations?

bungle: Good question. I don't think so.  Any solutions?

bungle: I think tokens are only inserted, and deleted, and selected. So perhaps it is fine for now?

bungle: No it is not fine, I will write an invalidation handler.

hishamhm:
Does Kong ever outputs raw `cache_key` values in the logs in a level other than `DEBUG`?

hishamhm:
I think always enabling read-before-write for `update` (782f05d)  is very controversial and needs some good discussion and consideration. I'm pointing this out so a major core change doesn't go "unnoticed" along with a plugin fix.

bungle:
> I think always enabling read-before-write for `update` ([782f05d](https://github.com/Kong/kong/commit/782f05d88c1fe4c268129b435ebb0db8984376d8)) is very controversial and needs some good discussion and consideration.

Yes, I will move this to waiting. The thing is, I think we broke `data.old_entity` with `1.0`. And that also warrants discussion.

PR #5511, opened 2020-01-30
Remove upstream ACAO header if no match is found

### LINKED ISSUE #4886: CORS plugin can't restrict upstream allow-origin header
### Summary

Enabling the CORS headers with a list of allowed domains under `config.origins` will return an `access-control-allow-origin` response header that matches the request `Origin` header, assuming the `Origin` is in the allowed list. If the content of the `Origins` request header is NOT in the list configured under `config.origins`, no `access-control-allow-origin` header is added, which is the desired behavior.

An issue arises, however, if the upstream server adds an `access-control-allow-origin` header that you wish to override. The upstream header is overwritten if there is a `config.origins` match, replacing it with the whitelisted domain. If the upstream server adds the `access-control-allow-origin` header for a domain you don't wish to allow, or sets it to`*`, this header is passed along to the client.

There seems to be no way block the upstream header while still allowing the CORS plugin to add headers for your configured domains. Presumably due to plugin precedence, if you configure Response Transformer to remove `access-control-allow-origin`, that header is never returned to the client under any circumstance.

This behavior isn't entirely surprising given Kong's default approach of forwarding headers, but it seems like a non-obvious default behavior if you expect the plugin to be managing your CORS headers. Having the plugin drop upstream headers seems like a sensible default, but perhaps a configuration option to drop or pass would make sense. 

### Steps To Reproduce

1. Configure a service with an upstream that returns a `access-control-allow-origin` header, such as the admin interface which returns `access-control-allow-origin: *`
```
{
    "client_certificate": null,
    "connect_timeout": 60000,
    "created_at": 1565367101,
    "host": "localhost",
    "id": "50a094de-785f-49d6-9fb2-69ca7e9ddcd1",
    "name": "localhost-admin",
    "path": "/",
    "port": 8001,
    "protocol": "http",
    "read_timeout": 60000,
    "retries": 5,
    "tags": null,
    "updated_at": 1565367101,
    "write_timeout": 60000
}
```
2. Configure a route to the admin service
```
{
    "created_at": 1565367101,
    "destinations": null,
    "headers": null,
    "hosts": [
        "kongdev.x.com"
    ],
    "https_redirect_status_code": 426,
    "id": "601226b0-8490-47f2-adb8-f511cd178a2f",
    "methods": null,
    "name": "admin",
    "paths": [
        "/admin"
    ],
    "preserve_host": true,
    "protocols": [
        "https"
    ],
    "regex_priority": null,
    "service": {
        "id": "50a094de-785f-49d6-9fb2-69ca7e9ddcd1"
    },
    "snis": null,
    "sources": null,
    "strip_path": true,
    "tags": null,
    "updated_at": 1565367101
}
```
Note that the admin interface returns `*` by default
```
$> http :8001
HTTP/1.1 200 OK
Connection: keep-alive
Content-Encoding: gzip
Date: Fri, 09 Aug 2019 17:36:31 GMT
Server: kong/1.3.0rc1
Transfer-Encoding: chunked
Vary: Accept-Encoding
```

3. Configure the global CORS plugin
```
        {
            "config": {
                "credentials": true,
                "exposed_headers": null,
                "headers": [
                    "DNT",
                    "X-CustomHeader",
                    "Keep-Alive",
                    "User-Agent",
                    "X-Requested-With",
                    "If-Modified-Since",
                    "Cache-Control",
                    "Content-Type",
                    "Authorization"
                ],
                "max_age": 1728000,
                "methods": null,
                "origins": [
                    "a.xxx.com",
                    "allowed-domain.fake"
                ],
                "preflight_continue": false
            },
            "consumer": null,
            "created_at": 1565367101,
            "enabled": true,
            "id": "7b0f9a35-a99a-4803-81e2-b8c8a6c4daf9",
            "name": "cors",
            "protocols": [
                "grpc",
                "grpcs",
                "http",
                "https"
            ],
            "route": null,
            "run_on": "first",
            "service": null,
            "tags": null
        }
```
4. Query with an allowed origin header and observe that the CORS plugin sets the `access-control-allowed-origin` header
```
$> http -v https://kongdev.x.com/admin/ Origin:allowed-domain.fake apikey:xxxxx
GET /admin/ HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: kongdev.x.com
Origin: allowed-domain.fake
User-Agent: HTTPie/1.0.2
apikey: xxxxx



HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: allowed-domain.fake
Connection: keep-alive
Content-Encoding: gzip
Date: Fri, 09 Aug 2019 17:39:44 GMT
Server: kong/1.3.0rc1
Transfer-Encoding: chunked
Vary: Origin
Via: kong/1.3.0rc1
X-Frame-Options: deny
X-Kong-Proxy-Latency: 1
X-Kong-Upstream-Latency: 4
```

5. Query the endpoint with a disallowed `origin` domain. Observe that `access-control-allowed-origin: *` is being passed from the upstream server.
```
$> http -v https://kongdev.x.com/admin/ Origin:disallowed-domain.fake apikey:xxxxx
GET /admin/ HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: kongdev.x.com
Origin: disallowed-domain.fake
User-Agent: HTTPie/1.0.2
apikey: xxxxx



HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Connection: keep-alive
Content-Encoding: gzip
Date: Fri, 09 Aug 2019 19:02:42 GMT
Server: kong/1.3.0rc1
Transfer-Encoding: chunked
Vary: Accept-Encoding
Via: kong/1.3.0rc1
X-Frame-Options: deny
X-Kong-Proxy-Latency: 0
X-Kong-Upstream-Latency: 5
```
### Additional Details & Logs

- Tested with Kong docker container kong:1.3.0rc1

zytrik:
### Summary

This is a fix for an issue filed as a task/feature at
https://github.com/Kong/kong/issues/4886

Upon closer inspection, this behavior is closer to a bug, and probably
creates unexpected behavior for all users of the CORS plugin.

The plugin logic accounts for all cases where it shuould add an ACAO
header to enable CORS.
- If the plugin is enabled but no allowed domains are defined, ACAO: *
- If allowed domains =  *, ACAO: *
- If one or more specific domains/regexes is defined and request Origin
header matches, ACAO: <request Origin header value>

However, if (and only if) a specific list of domains is defined and the
request Origin header FAILS to match, then the upstream header is passed
through. If your upstream server returns ACAO: *, there is no way to
implement a more restrictive CORS behavior with Kong.

This change implements a behavior in which while the CORS plugin is
enabled, Kong is always responsible for the content of the ACAO header.
This is much more clear and predictable. The only situation that this
behavior modification would prevent would be if an admin wants to allow
specific CORS URLs in Kong, and also allow a different list of specific
URLs at the upstream server, whitelisting both. I think the declarative
behavior in the Kong plugin is much more sensible.

The precedent for removing upstream headers is already set by the logic
for Access-Control-Allow-Headers.

### Full changelog

* Remove upstream Access-Control-Allow-Origin if Origin doesn't match the list of URLs set in the CORS plugin.

### Issues resolved

Fix #4886

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -153,6 +153,7 @@ local function configure_origin(conf)
     end
   end
 
+  kong.response.clear_header("Access-Control-Allow-Origin")
   return false
 end
 


bungle:
> If your upstream server returns ACAO: *, there is no way to implement a more restrictive CORS behavior with Kong.

Thanks @zytrik, I am sure you know CORS better than me, but one question:
Do you think it is worth to allow Kong only do it when it detects that Kong is more restrictive, and not always? Or did I get something wrong?

zytrik:
That's exactly the situation this resolves. If Kong is in a less
restrictive configuration, that's already handled.

- If you have set a list of allowed domains in Kong and the request Origin
matches, Kong adds the ACAO header to match, even if the upstream would not
have
- If you've set allowed domains to *, or enabled the plug-in and not set
allowed domains, ACAO will be set to the less restrictive *
- This patch only effects the situation where you've set a restrictive list
in Kong, and the upstream server tries to permit something that's not on
it. That's currently passed through, but should be dropped instead. If you
want Kong to allow more domains, you should add them to the list you've set
in Kong, not leave it up to a second authority.

bungle:
@zytrik, do you think you could add test for this? Or is there something that makes testing this a pain? Also we would be glad if you could share your knowledge on CORS in #5512?

zytrik:
> @zytrik, do you think you could add test for this? Or is there something that makes testing this a pain? Also we would be glad if you could share your knowledge on CORS in #5512?

I'll have to familiarize myself with the testing suite a bit more. The tricky part is ensuring the upstream sends an ACAO header that can be tested against. If you have any tips on how to accomplish that let me know.

bungle:
> I'll have to familiarize myself with the testing suite a bit more. The tricky part is ensuring the upstream sends an ACAO header that can be tested against. If you have any tips on how to accomplish that let me know.

Let me check if our mock server can do it (e.g. request can ask for headers). I think that might be missing. I'll PR if we don't have it.

bungle:
@zytrik, can you please take a quick look on added tests PR #5538? Thank you!

PR #5515, opened 2020-01-31
ci(github) use GitHub Actions as CI for Kong OSS

dndx:
This commit allows us to use GitHub Actions as CI solution. It provides
more concurrency and provides ease for excessive run time and queue
length we have been experiencing with Travis CI so far.

This commit makes Kong able to run on GitHub Actions, but still kept the
Travis CI files around so that we could use both at the same time during
the transitional period. Eventually Travis CI support will be removed.

This PR further splits Kong core integration tests into two halves on
GitHub Actions to further reduce the run time by utilizing all the
concurrency we have access to, as the core integration tests has
consistently been the long tail of all the jobs, running 2x slower than
the second slowest job. Splitting is not done on Travis as the
concurrency limit makes doing so pointless.

### DIFF
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -21,8 +21,23 @@ else
 fi
 
 if [ "$TEST_SUITE" == "integration" ]; then
-    eval "$TEST_CMD" spec/02-integration/
+    if [[ "$TEST_SPLIT" == first* ]]; then
+        # GitHub Actions, run first batch of integration tests
+        eval "$TEST_CMD" $(ls -d spec/02-integration/* | head -n4)
+
+    elif [[ "$TEST_SPLIT" == second* ]]; then
+        # GitHub Actions, run second batch of integration tests
+        # Note that the split here is chosen carefully to result
+        # in a similar run time between the two batches, and should
+        # be adjusted if imbalance become significant in the future
+        eval "$TEST_CMD" $(ls -d spec/02-integration/* | tail -n+5)
+
+    else
+        # Non GitHub Actions
+        eval "$TEST_CMD" spec/02-integration/
+    fi
 fi
+
 if [ "$TEST_SUITE" == "dbless" ]; then
     eval "$TEST_CMD" spec/02-integration/02-cmd \
                      spec/02-integration/05-proxy \
@@ -77,5 +92,5 @@ if [ "$TEST_SUITE" == "plugins" ]; then
     fi
 fi
 if [ "$TEST_SUITE" == "pdk" ]; then
-    TEST_NGINX_RANDOMIZE=1 prove -I. -j$JOBS -r t/01-pdk
+    TEST_NGINX_RANDOMIZE=1 prove -I. -r t/01-pdk
 fi
--- a/.ci/setup_env_github.sh
+++ b/.ci/setup_env_github.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# set -e
+
+dep_version() {
+    grep $1 .requirements | sed -e 's/.*=//' | tr -d '\n'
+}
+
+OPENRESTY=$(dep_version RESTY_VERSION)
+LUAROCKS=$(dep_version RESTY_LUAROCKS_VERSION)
+OPENSSL=$(dep_version RESTY_OPENSSL_VERSION)
+GO_PLUGINSERVER=$(dep_version KONG_GO_PLUGINSERVER_VERSION)
+
+
+#---------
+# Download
+#---------
+
+DOWNLOAD_ROOT=${DOWNLOAD_ROOT:=/download-root}
+BUILD_TOOLS_DOWNLOAD=$GITHUB_WORKSPACE/kong-build-tools
+GO_PLUGINSERVER_DOWNLOAD=$GITHUB_WORKSPACE/go-pluginserver
+
+KONG_NGINX_MODULE_BRANCH=${KONG_NGINX_MODULE_BRANCH:=master}
+
+#--------
+# Install
+#--------
+INSTALL_ROOT=${INSTALL_ROOT:=/install-cache}
+
+pushd $GO_PLUGINSERVER_DOWNLOAD
+  go get ./...
+  make
+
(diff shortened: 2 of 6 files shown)


dndx:
Core team has given green light to the experiment during meeting.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,359 @@
...
+          eval `luarocks path`
+          scripts/autodoc-admin-api
+
+    - name: Unit tests
+      run: |
+          eval `luarocks path`
+          bin/busted -v -o gtest spec/01-unit
+
+  integration-tests-postgres:
+    name: Integration tests with PostgreSQL (${{ matrix.suite }} - ${{ matrix.split }})
```
kikito: These long names are truncated so the interesting part is missing:

Screenshot 2020-02-05 at 16 42 29

A simple reorder would work:

```suggestion
    name: PostgreSQL (${{ matrix.suite }} - ${{ matrix.split }}) Integration tests
```

Alternatively the names could be shorter:
```suggestion
    name: Integration Pg (${{ matrix.suite }}-${{ matrix.split }})
```

This applies to the other names on this file as well.

hishamhm: Yes, please! I noticed the same thing.

.ci/run_tests.sh
```diff
@@ -21,8 +21,22 @@ else
...
-    eval "$TEST_CMD" spec/02-integration/
+    if [ "$TEST_SPLIT" == "first" ]; then
+        # GitHub Actions, run first half of integration tests
+        eval "$TEST_CMD" $(ls -d spec/02-integration/* | head -n4)
+
+    elif [ "$TEST_SPLIT" == "second" ]; then
+        # GitHub Actions, run second half of integration tests
+        # In case of odd number of directories, the second half will be
+        # running one more directory than the first
+        eval "$TEST_CMD" $(ls -d spec/02-integration/* | tail -n+5)
```
hishamhm: Not sure the comments match the behavior here... it mentions "first half" and "second half" and then comments about the odd number of directories, but this is really running the first 4 and then all the others (which are currently 6 folders).

hishamhm: Still unconvinced about the arbitrary split here, we'll have to keep rememebering on which log the tests are.

dndx: That comment came from the first iteration where the test cases were split in half evenly, except that did not result in a even split in terms of run time and that's why it was changed to this split that you see today which results in an almost even split of run time for those. If you can provide a better suggestion on the naming I would be more than happy to change the title.

.ci/setup_env_github.sh
```diff
@@ -0,0 +1,54 @@
...
+OPENSSL_INSTALL=$INSTALL_ROOT/openssl
+OPENRESTY_INSTALL=$INSTALL_ROOT/openresty
+LUAROCKS_INSTALL=$INSTALL_ROOT/luarocks
+
+eval `luarocks path`
+
+nginx -V
+resty -V
+luarocks --version
+openssl version
```
hutchic: Let's nuke this file entirely and rely on kong-build-tools

Examples:
https://github.com/hutchic/kong/blob/hutchic-next/.github/workflows/build_and_test.yml
https://github.com/Kong/kong-private/blob/chore/ci-testing/.travis.yml

dndx: Yes, I have created another card to do exactly that:

The GitHub Actions build relies heavily on the caching mechanism provided by Actions and is not directly convertible to use Docker, we should merge this as-is and experiment that in a separate branch instead of feature creep this PR any more (it has been open for a month already and has became very frustrating for me).

.ci/run_tests.sh
kikito: The biggest drawback about this is that in in order to to know whether an individual test goes into "first" and or "second" is to look at this file. I think this could be improved by an explicit renaming: `integration first half` could be renamed to `02-integration/01-04` and `integration second half` to `02-integration/05-xx`.

I see some alternatives, but they also have problems.

We could make the split directly on the filesystem (splitting `spec/02-integration/*` into two folders).
* Pro: To know what half your file is in, just look at the path. It is very "obvious".
* Pro: Impossible to forget to "assign a half" to a test.
* Con: Need "manually move files and folders from one folder to the other" from time to time in order to keep both halves "balanced".
* Con: Ties the "logical layout" of the tests to a "physical need" of one particular CI server.

We could add tags to the top `describe` on each file and use that for filtering.
* Pro: To know what "half" your file is in, just look at the top `describe`.
* Pro: "moving files from one half to the other" is a small diff - just retag.
* Pro: Allows for more granular splitting (by file instead of by folder). We could potentially split further (by test) but I don't think that's needed.
* Con: First commit would require manually tagging all files.
* Con: Need to remember to add the tag on files to keep things balanced.
* Con: Potentially execute `before` blocks that we don't need.
* Con: The "names" of each group of test would have to remain "nebulous", like "first" and "second", or "red" and "blue". 

If I had done this PR, I would have probably gone with the tags, since they look like a more natural fit to me. However the current approach has its own merits (provided we rename to something more obvious).

bungle:
I am in same lines. Splitting for extra concurrency is great, but we can do it later. Just not with this PR, to get it merged. Then we can add another card for split (take a time and make it proper). I think this is already great even without split. And with Travis, when I was the only user (during the morning), it was already a smooth ride.

dndx:
@bungle I think to properly split it will require some more work, however the current rough split already shoves 4 min off the run time for us and it's fairly low hanging fruit without making it overly complicated. I think the current state is obviously better than its Travis counterpart, and we should merge it first. That will give other people more concrete experience to guide future optimizations.

To be clear, we still runs the Travis CI after this PR, and no splitting will happens there.

bungle:
> @bungle I think to properly split it will require some more work, however the current rough split already shoves 4 min off the run time for us and it's fairly low hanging fruit without making it overly complicated

Yes, it doesn't bother me a lot. Not a blocker for me. Just want this to be merged, and getting rid of Travis after that.

PR #5516, opened 2020-01-31
feat(plugins) set generic X-Credential-Identifier

bungle:
### Summary

The PR #4993 implemented `X-Credential-Identifier` for `JWT Plugin` and it was decided
at time that we should add support for this less opinionated field name on other auth
plugins too. This commit adds it to other bundled auth plugins too:

- feat(basic-auth) set generic X-Credential-Identifier
- feat(key-auth) set generic X-Credential-Identifier
- feat(ldap-auth) set generic X-Credential-Identifier
- feat(oauth2) set generic X-Credential-Identifier

It also adds two more commits:

- chore(jwt) make set_consumer consistent with other auth plugins
- chore(constants) add note about CREDENTIAL_USERNAME deprecations

This PR makes `set_consumer` consistent on all the bundled auth plugins.

### Issues Resolved

This PR closes the PR #5201 by @lucasmoreno (only difference being more tests added and header name changed to Kong generic `X-Credential-Identifier`. Thanks @lucasmoreno, grab your contributor T-shirt at here: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt (you can link to this PR).


kong/plugins/jwt/handler.lua
```diff
@@ -226,6 +224,12 @@ local function do_authentication(conf)
 end
 
 
+local JwtHandler = {
```
dndx: Style: I recommend we leave this on top of the file for easier to find and consistency reasons.

kong/plugins/key-auth/handler.lua
```diff
@@ -169,6 +162,12 @@ local function do_authentication(conf)
 end
 
 
+local KeyAuthHandler = {
```
dndx: Same here.

kong/plugins/jwt/handler.lua
botnetdobbs: +1 on this

kong/plugins/key-auth/handler.lua
bungle: Usually in coding, you want variables close to usage, but I can bend on this.

PR #5524, opened 2020-02-03
feat(pdk) support `kong.response.exit` method for L4 proxy mode

dndx:
In L4 proxy mode, some of the features in here makes little sense. Such
as the JSON encoding support and `headers` argument. Therefore they have been removed.

Also, I have been looking at the way `kong.response.exit` is used right now in Kong code, and it appears that the `5xx` errors are the most problematic for returning directly to the client as they usually contain textual error message most certainly makes no sense to L4 protocol. Therefore, for `5xx` errors we log them to the error log file before simply close the downstream connection.

Feedbacks or comments are welcomed!

### DIFF
--- a/.ci/setup_env.sh
+++ b/.ci/setup_env.sh
@@ -70,6 +70,7 @@ kong-ngx-build \
     --kong-nginx-module $KONG_NGINX_MODULE_BRANCH \
     --luarocks $LUAROCKS \
     --openssl $OPENSSL \
+    --debug \
     -j $JOBS
 
 OPENSSL_INSTALL=$INSTALL_ROOT/openssl
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ addons:
     packages:
       - net-tools
       - libpcre3-dev
+      - valgrind
       - build-essential
   hosts:
     - grpcs_1.test
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -633,129 +633,219 @@ local function new(self, major_version)
     return send(response.status_code, response.content, response.headers)
   end
 
+  if ngx and ngx.config.subsystem == 'http' then
+    ---
+    -- This function interrupts the current processing and produces a response.
+    -- It is typical to see plugins using it to produce a response before Kong
+    -- has a chance to proxy the request (e.g. an authentication plugin rejecting
+    -- a request, or a caching plugin serving a cached response).
+    --
+    -- It is recommended to use this function in conjunction with the `return`
+    -- operator, to better reflect its meaning:
+    --
+    -- ```lua
+    -- return kong.response.exit(200, "Success")
+    -- ```
+    --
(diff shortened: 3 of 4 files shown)


bungle:
Just a note here, this might make automatic PDK document generation to endup with some weird results, so that might need to be updated too on docs repo.

dndx:
@bungle I did some experiments with LuaDoc and the current code generates duplicate function with the same name in the end result, and is probably not desirable. Therefore I merged them into one. That way the doc at least looks somewhat reasonable. Not sure if there is a better way though.

kong/pdk/response.lua
```diff
@@ -633,129 +633,199 @@ local function new(self, major_version)
...
-      coroutine.yield()
+      if body ~= nil and type(body) ~= "string" then
+        error("body must be a nil or a string", 2)
+      end
 
-    else
-      return send(status, body, headers)
+      if body then
+        if status < 500 then
+          local res, err =  ngx.print(body)
```
bungle: This is the only thing that I am wondering if it is worth it. It means that plugin probably has to understand the underlying protocol, and we speak about raw TCP which makes it really hard for plugins to deal with the output (especially when multiple different TCP protocols are used with the stream). It is good that there is this `< 500` at least, which probably covers a lot already. What if we make it `< 400`, and leave `200` as the only means for plugins to send data back to client.

kong/pdk/response.lua
```diff
@@ -633,129 +633,199 @@ local function new(self, major_version)
...
+          if not res then
+            error("unable to send body to client: " .. err, 2)
+          end
+
+        elseif body then
```
javierguerragiraldez: isn't this still within the `if body then...` clause?

bungle: Good catch, I guess it should be just `else` there.

.ci/setup_env.sh
```diff
@@ -70,6 +70,7 @@ kong-ngx-build \
     --kong-nginx-module $KONG_NGINX_MODULE_BRANCH \
     --luarocks $LUAROCKS \
     --openssl $OPENSSL \
+    --debug \
```
hishamhm: Is this a leftover?

.travis.yml
```diff
@@ -21,6 +21,7 @@ addons:
     packages:
       - net-tools
       - libpcre3-dev
+      - valgrind
```
hishamhm: Same here: leftover?

.ci/setup_env.sh
dndx: This is needed. The test case use the debug level information to determine the successfulness of the call (we should use debug build for CI tests anyway as that enables OpenResty built-in assertions).

.travis.yml
dndx: See above.

dndx:
Note: PR is red because of bad `next`. Waiting for it to be green before merging.

kikito:
Hi @dndx , could you rebase this branch? Let's see if tests pass afterwards.

PR #5525, opened 2020-02-03
feat(scripts) update copyright

kikito:
This new script updates the COPYRIGHT file. In order to do that:

* It attempts to download the license files for some basic dependencies (like OpenResty and LuaRocks)
* It installs all the Lua rocks for the current kong rockspec
* It then attempts to download all the license files for said rocks.

This PR also includes an update of the COPYRIGHT file done with the script.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -2,7 +2,7 @@
 
 Library
 
-Repo URL
+Library URL
 License URL
 
 License text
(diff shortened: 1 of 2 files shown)


hishamhm:
Sad but understandable that you couldn't use the `description.license` info from the rockspecs (those only contain license type but not the Copyright message).

hishamhm:
Sad but understandable that you couldn't use the `description.license` info from the rockspecs (those only contain license type but not the Copyright message).

I'm fixing the rebase, then it's good to go

PR #5536, opened 2020-02-05
feat/grpc-web

javierguerragiraldez:
### Summary

The grpc-web protocol is a derivative of gRPC, modified to be driven by a JS client running on a web browser.  The main differences are:

- works over any HTTP instead of only HTTP/2
- defaults to base64-encoded Protobuf payloads, but allows binary Protobuf and JSON encoding. (with appropriate Content-Type header)
- keeps the same message framing, but adds a flag bit to the message type, to allow for a new 'trailers' frame.  Added at the end of the content, this new frame replaces the HTTP trailer carrying the final response status.

### Current status:

- [x] base64 and binary Protobuf.
- [x] JSON payloads.
- [x] Pass compressed and uncompressed frames.
- [x] Status trailer frame.
- [ ] Streaming requests (not lost, but buffered because that's how the platform works).
- [x] Streaming responses.

In short, it does work with unmodified servers and client library for default and common cases.

The (de)framing object is slightly overspecified for the the current state, in preparation for the JSON transcoding feature.


kong/plugins/grpc-web/deco.lua
```diff
@@ -0,0 +1,212 @@
...
+    if not protofile then
+      return nil, "transcoding requests require a .proto file defining the service"
+    end
+    input_type, output_type = rpc_types(path, protofile)
+    if input_type == nil and output_type ~= nil then
+      return nil, output_type
+    end
+  end
+
+  return setmetatable({
```
dndx: Should cache `setmetatable` at module level.

Also, we usually reuse the same metatable and save it at module level instead of creating a new table everytime.

kong/plugins/grpc-web/deco.lua
```diff
@@ -0,0 +1,212 @@
...
+function deco.new(mimetype, path, protofile)
+  local text_encoding = text_encoding_from_mime[mimetype]
+  local framing = framing_form_mime[mimetype]
+  local msg_encoding = msg_encodign_from_mime[mimetype]
+
+  local input_type, output_type
+  if msg_encoding ~= "proto" then
+    if not protofile then
+      return nil, "transcoding requests require a .proto file defining the service"
+    end
```
dndx: Style: add a newline after this `if` block ends.

kong/plugins/grpc-web/deco.lua
```diff
@@ -0,0 +1,212 @@
...
+    framing = framing,
+    msg_encoding = msg_encoding,
+    input_type = input_type,
+    output_type = output_type,
+  }, deco)
+end
+
+
+local function frame(ftype, msg)
+  local f_hdr = ffi.new(frame_hdr, ftype, bit.bswap(#msg))
```
dndx: Style: those `ffi.*` functions should be cached as well.

Also, it seems that `f_hdr` doesn't have to be allocated every time, but can be shared by all the functions since no yielding happens.

kong/plugins/grpc-web/deco.lua
```diff
@@ -0,0 +1,212 @@
...
+  return ffi.string(f_hdr, ffi.sizeof(f_hdr)) .. msg
+end
+
+local function unframe(body)
+  if not body then
+    return
+  end
+
+  local hdr = ffi.cast(frame_hdr_p, body)
```
dndx: What if `body` is too small? Will we have buffer overruns in this case?

Since the job `frame_hdr` achieves is so simple, I would strongly recommend we get rid of the unsafe FFI calls and use the built-in Lua functions such as `string.byte` and logical shift to do the header parsing. It makes us much more resilient against memory errors. See: https://github.com/openresty/lua-resty-websocket/blob/master/lib/resty/websocket/protocol.lua#L84 as an example

kong/plugins/grpc-web/deco.lua
```diff
@@ -0,0 +1,212 @@
...
+
+  end
+
+  return body:sub(HEADER_SIZE + 1, frame_end), body:sub(frame_end + 1)
+end
+
+
+function deco:upstream(body)
+  if self.text_encoding == "base64" then
+    body = ngx.decode_base64(body)
```
dndx: This should be cached as well.

kong/plugins/grpc-web/handler.lua
```diff
@@ -0,0 +1,91 @@
...
+local grpc_web = {
+  PRIORITY = 1,
+  VERSION = '0.1.0',
+}
+
+
+function grpc_web:access(conf)
+  kong_response_set_header("Access-Control-Allow-Origin", "*")
+
+  if ngx_req_get_method() == "OPTIONS" then
```
dndx: Really not a huge issue, but there has been consensus that newer plugin code should utilize the PDK functions as much as possible. Which in this case, would be [kong.request.get_method()](https://docs.konghq.com/2.0.x/pdk/kong.request/#kongrequestget_method).

kong/plugins/grpc-web/handler.lua
```diff
@@ -0,0 +1,91 @@
...
+    return kong_response_exit(200, "OK", {
```
dndx: This static table can be cached to reduce runtime GC pressure.

kong/plugins/grpc-web/handler.lua
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+
+  local dec, err = deco.new(
+    kong_request_get_header("Content-Type"),
+    kong_request_get_path(), conf.proto)
+
+  if not dec then
+    kong.log.err(err)
+    return kong_response_exit(500, err, {})
```
dndx: Since we are not sending any new headers, the third argument can simply be omitted.

kong/plugins/grpc-web/handler.lua
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  local chunk, eof = ngx.arg[1], ngx.arg[2]
+
+  chunk = dec:downstream(chunk)
+
+  if eof and dec.framing == "grpc" then
+    chunk = chunk .. dec:frame(0x80, string_format(
+      "grpc-status:%s\r\ngrpc-message:%s\r\n",
+      ngx.var["sent_trailer_grpc_status"] or "0",
```
dndx: Style: `ngx.var` "magic table" should be cached at module level.

kong/plugins/grpc-web/handler.lua
```diff
@@ -0,0 +1,91 @@
...
+      ngx.var["sent_trailer_grpc_message"] or ""))
+  end
+
+  ngx.arg[1] = chunk
```
dndx: `ngx.arg` should be cacheable as well, the table itself is empty and the metamethods does the real magic :)

kong/plugins/grpc-web/handler.lua
```diff
@@ -0,0 +1,91 @@
...
+local kong_request_get_header = kong.request.get_header
+local kong_request_get_raw_body = kong.request.get_raw_body
+local kong_response_exit = kong.response.exit
+local kong_response_set_header = kong.response.set_header
+local kong_service_request_set_header = kong.service.request.set_header
+local kong_service_request_set_raw_body = kong.service.request.set_raw_body
+
+
+local grpc_web = {
+  PRIORITY = 1,
```
dndx: Maybe use `2` or `3`, as `1` conflicts with `correlation-id`. `2` is used by `request-termination` which does similar things (generating a response).

https://github.com/Kong/kong/blob/next/kong/plugins/request-termination/handler.lua#L23

kong/plugins/grpc-web/deco.lua
javierguerragiraldez: ?
the metatable is the module...

kong/plugins/grpc-web/deco.lua
javierguerragiraldez: done

kong/plugins/grpc-web/deco.lua
javierguerragiraldez: fixed for `#body` <= sizeof(header), thanks

i did the bit-shifting+`string.char()` dance on the test scripts, since i don't want to use the same looking code to test; as you can see it's far more error-prone.  non-variable headers are the best case use for C structures, IMO.

kong/plugins/grpc-web/deco.lua
javierguerragiraldez: done

kong/plugins/grpc-web/handler.lua
javierguerragiraldez: done

kong/plugins/grpc-web/handler.lua
javierguerragiraldez: done.  I'll probably (partially) undo at some point, since i think there a better answer (maybe with non-static fields) can help the browser to omit subsequent CORS queries.

kong/plugins/grpc-web/handler.lua
javierguerragiraldez: done

kong/plugins/grpc-web/handler.lua
javierguerragiraldez: done

kong/plugins/grpc-web/handler.lua
javierguerragiraldez: done.  i suspected that, but didn't take the time to check if it was (re)created at each call.

kong/plugins/grpc-web/handler.lua
javierguerragiraldez: done, thanks

kong/plugins/grpc-web/deco.lua
javierguerragiraldez: done

kong/plugins/grpc-web/handler.lua
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+
+  local dec, err = deco.new(
+    kong_request_get_header("Content-Type"),
+    kong_request_get_path(), conf.proto)
+
+  if not dec then
+    kong.log.err(err)
+    return kong_response_exit(500, err)
```
kikito: Shouldn't the error status be `400` here?

It seems the only thing that can make `deco.new` fail is a missing proto file.

spec/03-plugins/26-grpc-web/01-proxy_spec.lua
```diff
@@ -0,0 +1,157 @@
...
+
+      assert.equal(resp .. trailer, res:read_body())
+      assert.is_nil(err)
+    end)
+
+     test("Call plain JSON via HTTP", function()
+      local res, err = proxy_client:post("/hello.HelloService/SayHello", {
+        headers = {
+          ["Content-Type"] = "application/json",
+--           ["Content-Length"] = tostring(#req)
```
kikito: Is this a leftover from development?

spec/03-plugins/26-grpc-web/01-proxy_spec.lua
```diff
@@ -0,0 +1,157 @@
...
+          ["Content-Length"] = tostring(#HELLO_REQUEST_TEXT_BODY),
+        },
+        body = HELLO_REQUEST_TEXT_BODY,
+      })
+
+      assert.equal(HELLO_RESPONSE_TEXT_BODY, res:read_body())
+      assert.is_nil(err)
+    end)
+
+    test("Call gRCP-bas64 via HTTPS", function()
```
kikito:
```suggestion
    test("Call gRCP-base64 via HTTPS", function()
```

kong/plugins/grpc-web/handler.lua
javierguerragiraldez: my original intention was that a misconfiguration should be "server error" (since there's no "proxy error" family), but i realize that now a request to an nonexistent method can also fail, so yes, 400 would be better.

spec/03-plugins/26-grpc-web/01-proxy_spec.lua
javierguerragiraldez: yes, thanks

spec/03-plugins/26-grpc-web/01-proxy_spec.lua
javierguerragiraldez: for a moment i thought "damn, if it was working with a wrong method name that means the tests are invalid!"...

kong/plugins/grpc-web/deco.lua
dndx: My apologize, I mistaken the actual table instance for the metatable. Yes this looks all fine to me.

One thing to mention is that since the `deco` is created a lot, https://github.com/openresty/lua-tablepool provides a simple interface for recycling the table instances so it doesn't have to be allocated every time.

kong/plugins/grpc-web/deco.lua
dndx: I think it's okay if we add some of the common operations (e.g. converting a 4-byte string to a Lua integer) to the `kong/tools/utils.lua` file. That would allow the test and plugin code to share the same helpers.

It is very uncommon for plugin codes having to use FFI operations directly. For the precise reason that has already been shown above. It is inherently more error-prone and buggy FFI call could cause far more damage than buggy Lua code. In Kong we limit the use of those unless in absolutely performance critical path such as the request router.

Since the bit operations in here is relatively simple and rewriting it using Lua wouldn't even generate much more code than the FFI version, it is a worthy tradeoff.

javierguerragiraldez: I agree on most points, but definitely not on "It [FFI] is inherently more error-prone".  Quite the contrary in many cases, particularly on protocols.

But we could discuss all day long and get trapped in dogmas.  So, what do you think on using lpack?  it's _very_ useful for protocols, well-tested, and standard on Lua 5.3

dndx: My argument is that FFI is more error-prone on the memory safety side, for the protocol parsing it may be better, but any code that goes into the main Kong repository has to be considered for both of these aspects.

`lpack` seems nice, first glance it does seems to guard against buffer overruns, which was one of the main concerns for protocol parsers. I actually just found out we already have:

https://github.com/Kong/kong/blob/master/kong-2.0.1-0.rockspec#L30
https://github.com/Kong/lua-pack

which seems to solve this problem perfectly :)

javierguerragiraldez: done.  it _is_ nicer and shorter.  a pity that the included lua_pack is an old and slightly incompatible fork of lpack, but the essence is the same.

dndx: @javierguerragiraldez We can update the `lua_pack` library as well since we control the fork. Feel free to submit a PR to the repo.

hishamhm:
@javierguerragiraldez we've been adding new plugins as separate repositories, and not in the main Kong repository. I think, for consistency, we should create a https://github.com/kong/kong-plugin-grpc-web repository instead and add it as a dependency in Kong.

kikito:
@javierguerragiraldez since it seems that this was moved to https://github.com/kong/kong-plugin-grpc-web, how do you want to handle this? Do you want to close this PR?

PR #5537, opened 2020-02-05
tests(grpc) verify streaming methods

javierguerragiraldez:
### Summary

Add tests for gRPC streaming responses and requests.  They do work in the sense that all content is passed as expected and in order.

The requests are buffered, so the service doesn't receive any content until the last message is sent by the client.  This could broke an application that depends on a "ping-pong" behaviour during a single RPC call.


gszr:
@javierguerragiraldez - could you rebase onto master?

javierguerragiraldez:
done.  but why?

gszr:
@javierguerragiraldez We generally merge non-breaking additions, including tests, in master

gszr:
(Further, `master` is merged into `next` more often than `next` into `master`)

PR #5541, opened 2020-02-05
feat(pdk) add error response function

locao:
This change allows plugins to set custom responses on >= 400 statuses.

Closes #5055


bungle:
@locao, I don't quite follow here. `response.exit` has always allowed custom errors. But the problem the user reported is that you cannot easily make is consistent. As `/kong_error_handler` only handles some errors and some are done by `response.exit`. The `/kong_error_handler` has that `Accept / Content-Type` header logic that we don't have on `exit`. Is there a sister PR or is this missing some files?

Issue #5018, opened 2019-09-11
why "X-Forwarded-Port“ value is not client request port?


007slm:
in the kong source code ,i find the “x-forwarded-port“ value is from server listen port. in some situation, such as in the docker ，my kong server listening port is not the docekr export port. this cause a wrong 302 redirct url . 
the source code looks like 
var.upstream_x_forwarded_port = var.server_port;
but i think x_forwarded_port should comes from http request port,
so i changed the code and tested well.  
the changed code simplified  is 
local port = utils.split(ngx.var.http_host,":");
if port is nil then
      port = 443;
end
Am i think some thing wrong, or  some thing i missing？

nick-zjfgf:
Is there an answer?

pgnewikow:
Same issue here.

ykuksenko:
Same issue here.
I tried looking at a workaround for this by binding the docker to port 443/80 but the docker image won't allow that at this time because it doesn't use CAP_NET_BIND_SERVICE even if the capability is added to the container. https://github.com/Kong/docker-kong/pull/213 should fix that in theory. But the pr seems stuck.

dcherniv:
bump

weewey:
bump. 
Facing issues in Sidekiq with the kong. Sidekiq uses rack which redirects request with the `x-forwarded-port`

trollr:
bump with fusionauth

bungle:
I think the port in `host` can only be trusted when it comes from a trusted source. Then you need to configure Kong to trust everything, and it will just be a fallback if client does not give `X-Forwarded-Port`. I am not sure if docker could do `proxy_protocol`.

nesl247:
Can we please get a resolution on this?

erikgb:
If this can be of any help to anyone; I've used the [Serverless Functions plugin](https://docs.konghq.com/hub/kong-inc/serverless-functions/) as a workaround to this issue. DB-less configuration:
```yaml
plugins:
  # Workaround for the following issue: https://github.com/Kong/kong/issues/5018
  - name: post-function
    config:
      functions:
        # Lua version of the ternary operator....
        - ngx.var.upstream_x_forwarded_port= ((kong.request.get_header("x-forwarded-proto") == "http") and 80 or 443)
```

nesl247:
[...]

PR #5561, opened 2020-02-12
feat(Forwarded-header): add configurable x-forwarded-* header to routes and services

### LINKED ISSUE #5559: Configurable X-Forwarded-* headers on services and routes
### Summary

Kong automatically assign X-Forwarded-* headers based on some magic (in my case definitely wrong logic). So it would be great if I could overwrite those headers based on routes and services like I would do in nginx location directives. 

### Solution

Add optional properties for x-forwarded-* header to services and routes. If those fields are empty use the magic to define x-forwarded headers, otherwise take the values

trollr:
### Summary

Add X-Forwarded-* to routes and services

### Full changelog

Add following configurable headers to services and routes:

- X-Forwarded-Proto
- X-Forwarded-Host
- X-Forwarded-Port

### Issues resolved

Fix #5559, #5018, #3461


bungle:
See my comment here:
https://github.com/Kong/kong/issues/5559#issuecomment-585380290

kong/runloop/handler.lua
```diff
@@ -1041,14 +1041,14 @@ return {
 
       local trusted_ip = kong.ip.is_trusted(realip_remote_addr)
       if trusted_ip then
-        forwarded_proto = var.http_x_forwarded_proto or scheme
-        forwarded_host  = var.http_x_forwarded_host  or host
-        forwarded_port  = var.http_x_forwarded_port  or port
+        forwarded_proto = route.x_forwarded_proto or service.x_forwarded_proto or var.http_x_forwarded_proto or scheme
```
bungle: In general these (trusted) ones should be in this order:

```suggestion
        forwarded_proto = var.http_x_forwarded_proto or route.x_forwarded_proto or service.x_forwarded_proto or scheme
```

But I feel like we use a sledgehammer to implement this. At least we should only have this on route or service but not on both. Perhaps only on service. Not sure if this should be on `db` or say ENV variable that you pass to Kong when it is started.

bungle: This is how Amazon does it, and this works just fine with Kong:
https://kevinhakanson.com/2018-03-15-amazon-load-balancers-x-forwarded-headers-and-proxy-protocol-support

I am not sure why Docker cannot do it.

bungle:
Perhaps something like:
```
KONG_PORTMAP="80:8000, 443:8443" kong start
```

kong/runloop/handler.lua
trollr: If we use a env var for this case I see two issues:

1. It's a implicit behaviour for every registered service and will cause new issues. 
2. Maybe I want to set different values for the services. 

In my opinion my solution proposal is the cleanest possible way to allow people to configure kong as they need it

trollr:
> Perhaps something like:
> 
> ```
> KONG_PORTMAP="80:8000, 443:8443" kong start
> ```
> 
> That is used to change use of `var.server_port` to matching mapped one, for `X-Forwarded-Port`. Do you have issues with other headers too?

That's exactly what happens in the k8s cluster. However it use the internal docker port to set x-forwarded-port

trollr:
Removed the x-forwarded-* header configuration for routes

kong/runloop/handler.lua
bungle: Ah, ok I see, you want to break X-Forwarded even in case where  you trust them.

bungle:
Before we go as deep as integrating something very specific to our core entities, like `X-Forwarded-*` headers. We should slice this to actual issue, you are having,

1. Do you have problems with other `X-Forwarded-*`  headers than `X-Forwarded-Port`, and can you describe the issue?

2. Right now this can be fixed by a plugin. And I think it is preferred to changing the core entities.

If it is only the `Port` that is causing a problem, perhaps we can add a more generic thing in the core entity. But then again 2. will solve this the way you want it to be solved.

trollr:
> Before we go as deep as integrating something very specific to our core entities, like `X-Forwarded-*` headers. We should slice this to actual issue, you are having,
> 
> 1. Do you have problems with other `X-Forwarded-*`  headers than `X-Forwarded-Port`, and can you describe the issue?
> 2. Right now this can be fixed by a plugin. And I think it is preferred to changing the core entities.
> 
> If it is only the `Port` that is causing a problem, perhaps we can add a more generic thing in the core entity. But then again 2. will solve this the way you want it to be solved.

At the moment I only have problems with the x-forwarded-port header described here: #5559

Could you please give me a hint how to fix this issue with a plugin? The request transformer plugin wont work to modify the x-forwarded-* headers. If there is a clear way how to do this I could contribute it or change the request transformer plugin to support it.

bungle:
@trollr,

Here is my quick and dirty `Forwarded` plugin that allows you to do what this PR does, but by plugin: https://github.com/Kong/kong-plugin-forwarded

Let me know what you think about it, and does it work for you?

bungle:
If you want you can just do that what I did there in a plugin in a serverless function directly.

bungle:
Here is an example how you can do it with `post-function`:
```
http :8001/services/bin/plugins \
  name=post-function \
 config.functions="\
ngx.var.upstream_x_forwarded_host='test.host';\
ngx.var.upstream_x_forwarded_port=1234;\
ngx.var.upstream_x_forwarded_proto='http'" -f
```

trollr:
Oh nice. That was easy. 

Solution for kong ingress controller
```
kind: KongPlugin
metadata:
  name: post-function
  namespace: fusionauth
config:
  functions:
    - "ngx.var.upstream_x_forwarded_port=443; -f"
plugin: serverless-functions
```

Thanks @bungle for your support :) Happy now

bungle:
@trollr,

Thank you too! Let us consider if we want to add such a plugin to core.

Btw. one typo on your config:

```
- "ngx.var.upstream_x_forwarded_port=443; -f"
```
should be just:
```
- ngx.var.upstream_x_forwarded_port=443
```

bungle:
And if you want to set more than one, it is quite easy as well:

```yaml
kind: KongPlugin
metadata:
  name: post-function
  namespace: fusionauth
config:
  functions:
    - ngx.var.upstream_x_forwarded_host=myhost.test
    - ngx.var.upstream_x_forwarded_port=443
    - ngx.var.upstream_x_forwarded_proto=https
plugin: serverless-functions
```

PR #5562, opened 2020-02-12
feat(pdk) error method honoring Accept header

locao:
Added a new `error` method that allows PDK users to exit with an error while honoring the Accept header or manually forcing a content-type.

Co-authored-by: onematchfox

Closes: #5055


kong/pdk/response.lua
```diff
@@ -110,6 +118,59 @@ local function new(self, major_version)
     [16] = "Unauthenticated",
   }
 
+  local HTTP_MESSAGES = {
```
bungle: Because this is generic `error` method we cannot just have selected error codes, but we need to have them all here. The error handler only handled selected codes, this should handle them all.

kong/pdk/response.lua
```diff
@@ -760,6 +821,196 @@ local function new(self, major_version)
...
+  -- })
+  --
+  -- ---
+  --
+  -- return kong.response.exit(403, "Access Forbidden")
+  --
+  -- ---
+  --
+  -- return kong.response.exit(403)
+  function _RESPONSE.error(status, message, headers)
```
bungle: Is there specific reason this couldn't just call `response.exit` like the `error_handlers` did?

kong/plugins/jwt/handler.lua
```diff
@@ -127,7 +127,7 @@ local function do_authentication(conf)
   local token, err = retrieve_token(conf)
   if err then
     kong.log.err(err)
-    return kong.response.exit(500, { message = "An unexpected error occurred" })
+    return kong.response.error(500, "An unexpected error occurred")
```
bungle: In cases where plugins use the default error message, they should not specify it.

```suggestion
    return kong.response.error(500)
```

OR even:
```suggestion
    return kong.response.error(ngx.HTTP_INTERNAL_SERVER_ERROR)
```

kong/pdk/response.lua
locao: I wasn't really certain that this was needed, but I chose to do this by: there are errors that can't use templates, like some sent by jwt. And I think that introducing the `Accept`/`Content-Type` treatment could inadvertently change the behavior of some plugins (although that would be correct). 

It's simple to join the behavior of `exit` and `error`, what do you think?

kong/pdk/response.lua
```diff
@@ -760,6 +849,196 @@ local function new(self, major_version)
...
+            else
+              m = ngx.re.match(entry, [[^\s*level=(\d)\s*$]], "ajoi")
+              if m then
+                level = tonumber(m[1])
+              end
+            end
+          end
+        end
+
+        if level <= hi_level and quality > hi_quality then
```
kikito: I'm a bit confused by the algorithm used here. Perhaps it's because of the names of `hi_level` and `hi_quality` It seems we are looking for the *highest* quality and the *lowest* level? (so it would be `min_level`, `max_quality`)

Apart from the naming, I've been reading [RFC 7231, section 5.3.2](https://tools.ietf.org/html/rfc7231#section-5.3.2) and I have two concerns:

* It seems that this parser won't be able to satisfy requests with media ranges such as `Accept: text/*` (it will find no match on the name, and return the default type, `CONTENT_TYPE_JSON`, instead of html or txt).
* I am also almost certain that the `level` attribute shouldn't be used. See https://stackoverflow.com/a/14527918

kikito:
There is a bit of code duplication between this PR and `error_handlers`. Perhaps there's room for refactoring this into a single place (maybe in utils)?

kong/pdk/response.lua
bungle: Also I think this advanced parsing should not be done with Content-Tupe, only with Accept.

locao: >Also I think this advanced parsing should not be done with Content-Tupe, only with Accept.

@bungle The `Content-Type` is used to enable the plugins to ignore the `Accept` header, do you think it is not needed?

kong/pdk/response.lua
```diff
@@ -760,6 +849,196 @@ local function new(self, major_version)
   end
 
 
+  local function get_response_type(content_header)
+    local type = CONTENT_TYPE_JSON
+
+    if content_header ~= nil then
+      local accept_values = split(content_header, ",")
+      local hi_level = 100
+      local hi_quality = 0
```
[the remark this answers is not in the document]
locao: You are right, that naming was terrible and `level` shouldn't be used. I just removed it. Thanks!

bungle:
@locao, can you make one additional check? When client requests an unsupported content-type, but it contains `*/*`, and that is matched, we should default to `json` instead of `text/plain`.

spec/01-unit/05-utils_spec.lua
```diff
@@ -676,4 +676,27 @@ describe("Utils", function()
...
+      assert.equal("text/plain; charset=utf-8", utils.get_mime_type("text/plain"))
+      assert.equal("text/plain; charset=utf-8", utils.get_mime_type("text/*"))
+      assert.equal("text/plain; charset=utf-8", utils.get_mime_type("text/*; charset=utf-8"))
+      assert.equal("application/xml; charset=utf-8", utils.get_mime_type("application/xml"))
+      assert.equal("", utils.get_mime_type("application/grpc"))
+    end)
+
+    it("with unsupported or invalid mime types", function()
+      assert.equal("application/json; charset=utf-8", utils.get_mime_type("audio/*", true))
+      assert.equal("application/json; charset=utf-8", utils.get_mime_type("text/css"))
```
kikito: Please add a line for the "default" content type, which is a bit special since it is defined as `CONTENT_TYPE_DEFAULT`:
```suggestion
      assert.equal("application/json; charset=utf-8", utils.get_mime_type("text/css"))
      assert.equal("application/json; charset=utf-8", utils.get_mime_type("default"))
```

spec/03-plugins/09-key-auth/02-access_spec.lua
```diff
@@ -180,7 +180,7 @@ for _, strategy in helpers.each_strategy() do
         })
         assert.res_status(401, res)
         local body = assert.res_status(401, res)
-        assert.equal([[{"message":"No API key found in request"}]], body)
+        assert.equal("{\n  \"message\":\"No API key found in request\"\n}", body)
```
kikito: Let's not tie ourselves to those blank characters:
```suggestion
        assert.same({message = "No API key found in request"}, cjson.parse(body))
```

locao:
> @locao, can you make one additional check? When client requests an unsupported content-type, but it contains `*/*`, and that is matched, we should default to `json` instead of `text/plain`.

Done!

kong/pdk/response.lua
bungle: What I mean that `Content-Type` header's content is not to be parsed like `Accept` header content. I mean that:
1. check if there is accept header and parse
2. if not check if there is content-type header and just use that, without parsing it.

PR #5568, opened 2020-02-13
fix(db) read-before-write always on updates and upserts

bungle:
### Summary

I know this is controversial. But It turns out that currently with `upsert` the invalidations do not send the `old entity`, while in _most_ (I think it always should) cases with `update` it does so.

Because there is `old_entity` in `update` (which `upsert` uses too) invalidations and we do have code that relies on it and perhaps 3rd party code in plugins that relies on it, this commit changes our DAOs so that `read-before-write` is always enabled for `update` and `upsert`.

This is for correctness (even when it might mean race condition between read and write). And bugs caused by not having this commit are really hard to investigate. I know that we tried hard to avoid `read-before-write` especially with Cassandra, but we also have had to loosen that and more and more code already does `read-before-write` on `updates`, almost to the point it was going to happen, no matter what. In that sense this does not make things that much different.

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -38,6 +38,18 @@ local DAO   = {}
 DAO.__index = DAO
 
 
+local function remove_nulls(tbl)
+  for k,v in pairs(tbl) do
+    if v == null then
+      tbl[k] = nil
+    elseif type(v) == "table" then
+      tbl[k] = remove_nulls(v)
+    end
+  end
+  return tbl
+end
+
+
 local function validate_size_type(size)
   if type(size) ~= "number" then
     error("size must be a number", 3)
@@ -360,7 +372,7 @@ end
 
 
 local function check_update(self, key, entity, options, name)
-  local entity_to_update, err, read_before_write, check_immutable_fields =
+  local entity_to_update, err, check_immutable_fields =
     self.schema:process_auto_fields(entity, "update")
   if not entity_to_update then
     local err_t = self.errors:schema_violation(err)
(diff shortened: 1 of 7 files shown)


kong/db/dao/init.lua
```diff
@@ -368,7 +368,7 @@ local function check_update(self, key, entity, options, name)
   end
 
   local rbw_entity
-  if read_before_write then
+  if read_before_write then -- this is currently always true
```
hutchic: if it's always true do we need the conditional at all?

kong/db/dao/init.lua
```diff
@@ -431,26 +431,40 @@ local function check_update(self, key, entity, options, name)
 end
 
 
-local function check_upsert(self, entity, options, name, value)
-  local entity_to_upsert, err = self.schema:process_auto_fields(entity, "upsert")
+local function check_upsert(self, key, entity, options, name)
```
hutchic: why did `value` change to `key`?

kong/db/dao/init.lua
bungle: @hutchic, the value comes from other function. If for some reason that function changes, we don't need to change code here.

kong/db/dao/init.lua
bungle: Key makes more sense. It is either `unique value` used to select `by unique key` or the `primary key`.

hishamhm:
@thibaultcha Since you're the one who originally pushed hard for adding detections to avoid performing read-before-write on updates, I'd like to get your input on this.

The current state of things, as I read it: 

* enabling unconditional RBW in `upsert` is an absolute necessity because, since you get a full entity, it is impossible to know when the primary/cache key changed, so this is a must for ensuring we invalidate correctly. If this PR was handling `upsert` only, I would have approved and merged this a long time ago.

* enabling unconditional RBW in `update` is a more opinionated thing. There is a bunch of cases where we need to do it already (when changing values that are part of primary/cache keys; when doing partial updates on any records because they're stored as JSON) and we've had bugs in the past due to missing cases, though I think this part of the code is stable by now (even though it's clunky).
  * it's my understanding that @bungle's wish for unconditional RBW in `update` comes from wanting a full entity to be _always_ provided in `old_entity` when getting the CRUD event notification. The `old_entity` value was originally meant to contain only the PK of the old entity since it was added for handling CRUD invalidations only, but because RBW currently happens sometimes, turns out that sometimes you end up getting a full `old_entity` on that CRUD event notification. @bungle has at least one use case where he could make use of the full old entity when getting CRUD notifications in a plugin, so getting the full entity only "sometimes" did feel like a bug. IMO it's more of an unspecified behavior (we never really made formal promises about `old_entity` as a stable API for plugins), but it's still not ideal.

Given that we end up doing RBW in a lot of cases anyway (e.g. most `plugins` updates require it because of `config`) and every `PUT` will require it, I'm not opposed to simplifying the code and gettting rid of all this RBW-detection logic, which has always seemed a bit fragile — since there are so many cases where it fails to avoid RBW, the DAO offers a quite weak guarantee in practice. (A thought that did cross my mind is that, if we _really_ wanted to avoid RBW as much as possible, we could do it at strategy level instead, and with enough [Postgres magic](https://stackoverflow.com/questions/18209625/how-do-i-modify-fields-inside-the-new-postgresql-json-datatype) we could even avoid RBW completely on Postgres by returning the updated and the previous object in a single transaction)

So, @thibaultcha, any strong feelings on this PR?

kong/db/dao/init.lua
hishamhm: @bungle appreciate taking it on the careful side, but I agree with Colin on this case. If the idea is to remove the RBW detection logic, then it's better to remove that `read_before_write` result from `process_auto_fields` (which was always a stretch from the function's original task...). Since the criteria for RBW as of this PR is uniquely the DAO context, I don't think it's worth keeping anything related to `read_before_write` in `kong.db.schema.init`.

bungle: @hishamhm,

Ok I'll change it.

PR #5575, opened 2020-02-17
tests(balancer) clean up and speed up balancer tests

locao:


spec/fixtures/balancer_utils.lua
```diff
@@ -0,0 +1,533 @@
...
+    })
+  end
+end
+
+
+local poll_wait_health
+local poll_wait_address_health
+do
+  local function poll_wait(upstream_id, host, port, admin_port, fn)
+    local hard_timeout = ngx.now() + 70
```
dndx: Can this be rewritten using [`wait_until`](https://github.com/Kong/kong/blob/next/spec/helpers.lua#L315) instead? The code below seems to do exactly what `wait_until` does.

PR #5583, opened 2020-02-19
docs(admin-api) check for and add documentation to missing endpoints

hishamhm:
Our check for missing endpoint documentation was checking for
missing _entity_ documentation only, and not to the general
files such as `kong/api/routes/kong.lua`.

This commit:

* adds the check to those files
* adds the missing documentation so that the check passes
* removes a deprecated endpoint (which is now documented in
  the non-deprecated style)

### DIFF
--- a/autodoc/data/admin-api.lua
+++ b/autodoc/data/admin-api.lua
@@ -150,6 +150,136 @@ return {
           ]],
         },
       },
+      ["/endpoints"] = {
+        GET = {
+          title = [[List available endpoints]],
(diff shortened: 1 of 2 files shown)


hutchic:
how far back do we hypothetically need to backport this onto the docs repository?

hishamhm:
@hutchic not sure about (a) which version adopted the auto-generated Admin API endpoints and (b) which versions included these additional endpoints that were not documented.

PR #5584, opened 2020-02-19
feat(db) add read-only replica support for Postgres strategy

### LINKED ISSUE #3153: Postgres settings Cluster and Reader endpoint AWS Aurora RDS
Hi,

I'm relatively new in using kong and aws rds. I noticed when I created aurora cluster, It shows cluster and reader endpoint. I also setup a read-replicas. Based on the documentation, the cluster writer endpoint will always point to the writer instance which can accept read/writes. While the cluster reader endpoint will point to the readers and will direct all the read traffic to that. When i saw the postgres settings in Kong, there is only one entry for postgres host. Will I only place the cluster endpoint?

### LINKED ISSUE #4641: [Q] kong access to read-only postgresql
node#1 : kong server + postgresql (master)

node#2 : kong server + postgresql (slave - read only)

Two postgresqls are synced using replication. 

Postgresql server of node#2 is read only.  (INSERT,DELETE,UPDATE IS NOT ALLOWED)

so below error occurred.
....................................
2019-05-22 08:32:04 KST [172]: [9819-1] db=tarantula,user=kong,app=[pgmoon],host=[127.0.0.1] ERROR:  current transaction is aborted, commands ignored until end of transaction block
2019-05-22 08:32:04 KST [172]: [9820-1] db=tarantula,user=kong,app=[pgmoon],host=[127.0.0.1] STATEMENT:  BEGIN;
          DELETE FROM "oauth2_authorization_codes" WHERE "ttl" < CURRENT_TIMESTAMP AT TIME ZONE 'UTC';
          DELETE FROM "oauth2_tokens" WHERE "ttl" < CURRENT_TIMESTAMP AT TIME ZONE 'UTC';
        COMMIT;
....................................
Is there a way to prevent errors?
or

Can you add new parameter in kong.conf 
ex)  db-readonly=true

Sorry for poor english.

dndx:
This feature allows Postgres users (especially those who uses Amazon Aurora) to
create replicas that are intended for read-only operations. When
enabled, Kong will attempt to fulfill any read operations through the
read-only replicas instead of the main Postgres connection.

Note that Kong nodes can not yet run entirely off a read-only database,
so a connection with read-write access is always needed. Although after
enabling the read-only replica support user are expected to see majority
of the queries being served by the replica, therefore substantially
reducing the database connection number and load on the main database
instance.

Replicas can be eventually consistent when replicating changes from the
main instance. However replicas *must* include all the data from
the main instance to avoid undefined behaviors.

Replicas are enabled with the `pg_ro_*` configuration options that corresponds
to the standard `pg_*` configuration options. At least `pg_ro_host` is
needed for it to be enabled and more configuration option can be
specified if differs from the main connection. When left unspecified,
same configurations from the main connection are used by default.

Closes #3153
Closes #4641


dndx:
Note: tests are failing for unrelated reasons. Issue will be addressed on `next` soon.

kong/conf_loader.lua
```diff
@@ -321,6 +321,14 @@ local CONF_INFERENCES = {
   pg_max_concurrent_queries = { typ = "number" },
   pg_semaphore_timeout = { typ = "number" },
 
+  pg_ro_port = { typ = "number" },
+  pg_ro_timeout = { typ = "number" },
+  pg_ro_password = { typ = "string" },
+  pg_ro_ssl = { typ = "boolean" },
+  pg_ro_ssl_verify = { typ = "boolean" },
+  pg_ro_max_concurrent_queries = { typ = "number" },
+  pg_ro_semaphore_timeout = { typ = "number" },
```
hishamhm: These need to be documented in `kong.conf.default`

spec/kong_tests.conf
```diff
@@ -15,6 +15,9 @@ pg_host = 127.0.0.1
 pg_port = 5432
 pg_timeout = 10000
 pg_database = kong_tests
+# note: this does not trigger readonly mode to be enabled on its own
+# for that pg_ro_host is also needed
+pg_ro_user = kong_ro
```
hishamhm: minor thing:

```suggestion
pg_ro_user = kong_ro_tests
```

just to keep the pattern of having the `tests` entries in the DB separate from the regular ones, makes it easier to avoid mixing up manual and automated tests

spec/02-integration/03-db/11-postgres-ro_spec.lua
```diff
@@ -0,0 +1,140 @@
...
+      if proxy_client then proxy_client:close() end
+      helpers.stop_kong()
+    end)
+
+    describe("proxy and admin API works", function()
+      local route_id
+
+      it("can change and retrieve config using Admin API", function()
+        local res = assert(admin_client:post("/services", {
+          body = { name = "mockbin-service", url = "https://mockbin.org/request", },
```
hishamhm: Please change the URLs in this test from using `mockbin.org` to use the testsuite's own mock upstream at `https://127.0.0.1:15556` instead.

At one point in the past we did a huge cleanup removing dependencies from the availability of mockbin.org from our test suite (a quick grep showed that the hybrid-mode tests are using it as well, that slipped review).

dndx: Addressed both location.

kong/conf_loader.lua
dndx: Added. Thanks!

spec/kong_tests.conf
dndx: Hmm, it seems that the current pattern for running integration tests is to run it with the `kong` user, therefore I used the `kong_ro` user for the readonly mode. Otherwise we will have to change `pg_user = kong_tests` as well to keep them consistent.

hbagdi:
Looks pretty good.
We should clarify add some visibility into `/status` endpoint for this.
We only have `database_reachability` as `true` or `false` until now but adding some visibility around the leader and follower(s) Postgres instances will be very helpful operationally.

This comment could be discarded as "scope creep" one but I'd advise that we ship this addition in the same release, if not the same change set.

If we don't want to do it, then we should document that `/status` health-checks DB-connectivity with Postgres leader (at least that's my understanding from the reading of connector.lua).

spec/kong_tests.conf
hishamhm: Ah, ok, we use the same user but a different database (it's still `kong` user but `kong_tests` for the database/keyspace).

kong.conf.default
```diff
@@ -741,6 +741,37 @@
                                  # Detailed discussion of this behavior is
                                  # available in the online documentation.
 
+# When using Postgres as the backend storage, you can optionally enable Kong
```
hishamhm: These comments don't match the style of the rest of the document — I'm not pointing this out for stylistic reasons though: I'm afraid they won't be formatted correctly by the generator script that generates [this page](https://docs.konghq.com/2.0.x/configuration/#properties-reference) based on this file.

If you want to add an intro prose section to the documentation, this should be added to [this file instead](https://github.com/Kong/docs.konghq.com/blob/master/autodoc-conf/data.lua) — (yes, we want to eventually move this file and the script to the main Kong repo like we did for `autodoc-admin-api`), and the comments in this file should be formatted as descriptions for each entry so they generate the "Properties reference" in the documentation website.

dndx: @hishamhm Thanks for the note, I did not realize the importance of formatting in here first.

I have moved this paragraph up to the "DATASTORE" section instead of the `data.lua` file. It seems to makes more sense to me as it clearly falls under such category, plus we also explain things such as DB-less mode in there so it should follows the current pattern after the move. 

I also changed the comment style of the properties to make sure autodoc understands them. Here is a preview of the generated Markdown after the change: https://gist.github.com/dndx/807132b43edd58f478a9d9deafe4d30a. Seems that everything is rendering correctly.

dndx:
@hbagdi Yes, I think we want to add that as a second phase work due to the size of this PR already and the 2.1 release timeline. But thanks for the suggestion. I will update the `/status` API docs accordingly and made a card to track that feature.

dndx:
@kikito This is to stay consistent with how C* connector names it right now:

https://github.com/Kong/kong/blob/master/kong/db/strategies/cassandra/connector.lua#L418

kong/global.lua
```diff
@@ -73,6 +73,20 @@ function _GLOBAL.set_phase(self, phase)
 end
 
 
+function _GLOBAL.get_phase(self)
+  if not self then
+    error("arg #1 cannot be nil", 2)
+  end
+
+  local kctx = self.ctx
+  if not kctx then
```
javierguerragiraldez:
```suggestion
  if not kctx or not kctx.core then
```

PR #5587, opened 2020-02-19
fix(router) don't change route paths when marshaling

### LINKED ISSUE #5547: Incomplete Response Returned by kong.router.get_route()
### Summary

With version 1.4.2 `kong.router.get_route()` had a `paths` property returned which returned a list of paths on the route. We're in the process of upgrading to version 2.0.1 and noticed a behavior change, `paths` now only returns the last path in the array. We're using DB-less mode, if it matters.

**Example with 1.4.2:**
```json
{
  "paths": [
    "/a",
    "/b",
    "/c",
    "/d"
  ]
}
```

**Example with 2.0.1:**
```json
{
  "paths": [
    "/d"
  ]
}
```

Is this change intended? Is there anyway to force the full paths response back?

### Steps To Reproduce

1. In the access phase, log out the response of `kong.router.get_route().paths`

### Additional Details & Logs

- Operating system

locao:
When routes had more than one path, only the matched path was being kept in route structure.

Fix #5547


gszr:
Could we add a regression test?

locao:
> good to go once green ✅ tests would be nice, but not a blocker; also, should this go into `master` instead?

Yes, you're right. Thank you!

PR #5592, opened 2020-02-21
tests(*) use busted-htest for shorter color output

hishamhm:
Use https://github.com/hishamhm/busted-htest as a default for shorter and nicer-looking test outputs in CI.

### DIFF
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -8,7 +8,7 @@ function red() {
     echo -e "\033[1;31m$*\033[0m"
 }
 
-export BUSTED_ARGS="-o gtest -v --exclude-tags=flaky,ipv6"
+export BUSTED_ARGS="-o htest -v --exclude-tags=flaky,ipv6"
 
 if [ "$KONG_TEST_DATABASE" == "postgres" ]; then
     export TEST_CMD="bin/busted $BUSTED_ARGS,cassandra,off"
--- a/.ci/setup_env.sh
+++ b/.ci/setup_env.sh
@@ -107,12 +107,14 @@ if [[ "$TEST_SUITE" == "pdk" ]]; then
   cpanm --notest --local-lib=$TRAVIS_BUILD_DIR/perl5 local::lib && eval $(perl -I $TRAVIS_BUILD_DIR/perl5/lib/perl5/ -Mlocal::lib)
 fi
 
-# ----------------
-# Run gRPC server |
-# ----------------
+# ---------------
+# Run gRPC server
+# ---------------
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 OS := $(shell uname | awk '{print tolower($$0)}')
 MACHINE := $(shell uname -m)
 
-DEV_ROCKS = "busted 2.0.0" "luacheck 0.23.0" "lua-llthreads2 0.1.5" "http 0.3"
+DEV_ROCKS = "busted 2.0.0" "busted-htest 1.0.0" "luacheck 0.23.0" "lua-llthreads2 0.1.5" "http 0.3"
```
Tieske: not sure this should be in the make file. Since it's a preference, not required for testing, maybe simply add the installation to the `.travis.yml` ?

hishamhm: The scripts assume gtest, which is a preference as well.

hishamhm: To expand on this point: the `.ci/` scripts which run the tests specify the Busted output handler, and those depend on the tools built by the Makefile. They are designed to be CI-agnostic (and in fact, we're testing and about to migrate to Github Actions). If this moves to the `.travis.yml` file this produces a cross-dependency between the CI config file and the test scripts, which break their CI-agnosticity.

Tieske: I don't get your second point. My thinking is that the gtest one is bundled with busted. Busted is required anyway. So if this htest output is required in ci, then that's where it should be configured. 

Otherwise we'll start adding more "preferred" utilities to the makefile, which would clutter everybody's systems.

So imo, better stay away from that, keep the generic one lean and mean (makefile), and add the preferred output only to ci.

(But if we don't care, then I definitely have some fine utility functions to add to the test helpers for example)

Tieske:
Pongo updated! https://github.com/Kong/kong-pongo/commit/ab549a788a0c9ce195e63d2e4248eb4dd9c9bf85

dndx:
Should we make this against `next` instead? As that's where most of the development happens.

Tieske:
I'll have to revert the Pongo changes, since the rockspec for that formatter is LuaRocks 3.0. That doesn't work, because the older Kong images have 2.x installed and hence are unable to install this rock.

I asked @hishamhm whether it's possible to revert the rockspec to a 2.x format.

hishamhm:
@Tieske it is possible, I'll push a 1.0.0-2 with the rockspec change alone when I'm back in front of a computer.

hishamhm:
@Tieske busted-htest 1.0.0-2 is up in luarocks.org :+1:

Tieske:
thx!

PR #5594, opened 2020-02-21
chore(ci) make sure tags dont get built by the nightly cron

hutchic:
There's only two conditions in which Jenkins should be doing anything

Release:
- run from SCM don't run from cron

Nightly Release:
- run from timer don't run from SCM
- only run on master and next

SCM push to next / master is a noop
Capture

cron event runs build, test and nightly release stages
Capture2

SCM tag runs build, test and nightly release stage also the tag doesn't attempt to re-run every 24hrs
Capture3

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,6 +59,12 @@ pipeline {
                     }
                 }
             }
+            /* the above when statement evaluates to:
+                if (
+                  ( buildingtag && not cron ) ||
+                  ( ( branch = master || branch = next) && cron )
+                )
+            */
             parallel {
                 stage('dbless') {
                     agent {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,13 @@
 pipeline {
     agent none
+    triggers {
+        cron(env.BRANCH_NAME == 'master' | env.BRANCH_NAME == 'next' ? '@daily' : '')
+    }
+    options {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,17 +1,35 @@
 pipeline {
     agent none
+    triggers {
+        cron(env.BRANCH_NAME == 'master' | env.BRANCH_NAME == 'next' ? '@daily' : '')
+    }
```
hutchic: we may need to cherry-pick and push this line onto the 2.x tags

Jenkinsfile
```diff
@@ -1,17 +1,35 @@
...
+                    allOf {
+                        buildingTag()
+                        not { triggeredBy 'TimerTrigger' }
+                    }
+                    allOf {
+                        triggeredBy 'TimerTrigger'
+                        anyOf { branch 'master'; branch 'next' }
+                    }
+                }
+            }
```
hutchic: this evaluates to

```
if (
  ( buildingtag && not cron ) ||
  ( ( branch = master || branch = next) && cron )
)

kikito: Please put that in a comment on the file, too.

hutchic: added in f4485bbbaec09731108d6191dedaa6448702fdb6

PR #5597, opened 2020-02-22
Bugfix: instances of same go-plugin on different routes get mixed.

javierguerragiraldez:
### Summary
When using declarative configuration, the cache object holding the plugin configs doesn't
seem to call the `load_plugin_from_db()` function.  (maybe it's preloaded some other way?)
This makes they never get the full DB key injected in the config, so the
`go.get_instance()` function falls back to using just the plugin name as the key to the
instance ID table.  Hilarity ensues.

This patch moves the injection of the `__key__` and `__seq__` fields from the
`load_plugin_from_db()` function to the outer `load_configuration()` function, so it works
on any DB strategy.

fix: https://discuss.konghq.com/t/the-go-plugin-may-not-work-right/5545

### DIFF
--- a/kong/runloop/plugins_iterator.lua
+++ b/kong/runloop/plugins_iterator.lua
@@ -66,12 +66,6 @@ local function load_plugin_from_db(key)
     return nil, tostring(err)
   end
 
-  if type(row) == 'table' then
-    row.__key__ = key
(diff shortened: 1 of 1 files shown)


hutchic:
is this something we can write a test for 🤔 ?

PR #5600, opened 2020-02-24
feat(tests) allow specifying 'CUSTOM_PLUGIN_PATH' via environment var…

WALL-E:

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -59,7 +59,13 @@ log.set_lvl(log.levels.quiet) -- disable stdout logs in tests
 
 -- Add to package path so dao helpers can insert custom plugins
 -- (while running from the busted environment)
-package.path = CUSTOM_PLUGIN_PATH .. ";" .. package.path
+do
+  local paths = {}
+  table.insert(paths, os.getenv("KONG_LUA_PACKAGE_PATH"))
+  table.insert(paths, CUSTOM_PLUGIN_PATH)
+  table.insert(paths, package.path)
+  package.path = table.concat(paths, ";")
+end
 
 -- Extract the current OpenResty version in use and returns
 -- a numerical representation of it.


Tieske:
@WALL-E thks for the PR. But what is the purpose of this change?

adding a Lua path? Doesn't the regular path suffice: https://github.com/Kong/kong/blob/master/kong.conf.default#L1054 ?

Note that you can add these as `KONG_LUA_PACKAGE_PATH` environment variables. And if you're testing, you can use `KONG_TEST_LUA_PACKAGE_PATH`.

WALL-E:
@Tieske I've tried , `KONG_TEST_LUA_PACKAGE_PATH`  or `kong_test.conf/lua_package_path `,  but it doesn't work, I haven't found out why

Tieske:
hmmm, so why does that work? Maybe we should fix that, instead of adding a new one?

WALL-E:
@Tieske  yes, you are right.

I looked at this part of the code
https://github.com/Kong/kong/blob/master/bin/busted#L5,L23

but I don't find `KONG_LUA_PACKAGE_PATH` in source code, busted not use it. In other words, it is also a new environment variable for busted process.

spec/helpers.lua
```diff
@@ -7,7 +7,7 @@
 
 local BIN_PATH = "bin/kong"
 local TEST_CONF_PATH = os.getenv("KONG_SPEC_TEST_CONF_PATH") or "spec/kong_tests.conf"
-local CUSTOM_PLUGIN_PATH = "./spec/fixtures/custom_plugins/?.lua"
+local CUSTOM_PLUGIN_PATH = os.getenv("KONG_LUA_PACKAGE_PATH") or "./spec/fixtures/custom_plugins/?.lua"
```
hishamhm: Regular tests will fail if `spec/fixtures/custom_plugins` is not there, right? Maybe this should always concatenate that to the path?

WALL-E: yes, I have updated PR  @hishamhm

spec/helpers.lua
```diff
@@ -59,7 +59,13 @@ log.set_lvl(log.levels.quiet) -- disable stdout logs in tests
...
 -- Add to package path so dao helpers can insert custom plugins
 -- (while running from the busted environment)
-package.path = CUSTOM_PLUGIN_PATH .. ";" .. package.path
+local lua_package_path = os.getenv("KONG_LUA_PACKAGE_PATH")
+if not lua_package_path then
+  package.path = CUSTOM_PLUGIN_PATH .. ";" .. package.path
+
+else
+  package.path = lua_package_path .. ";" .. CUSTOM_PLUGIN_PATH .. ";" .. package.path
+end
```
Tieske: how about;
```lua
do
  local paths = {}
  table.insert(paths, os.getenv("KONG_LUA_PACKAGE_PATH"))
  table.insert(paths, CUSTOM_PLUGIN_PATH)
  table.insert(paths, package.path)
  package.path = table.concat(paths, ";")
end
```
for code clarity.

WALL-E: I got it

hishamhm:
@WALL-E thank you, merged!

PR #5606, opened 2020-02-25
fix(migrations) do not error if desired state is achieved

### LINKED ISSUE #5605: kong migrations reset returns an error
```shell
/kong # kong migrations reset -y
Database not bootstrapped, nothing to reset
/kong # echo $?
1
/kong #
```
If there is nothing to reset, the desired state has been achieved, and the command should return exit status `0`

Tieske:
resetting a database while not bootstrapped essentially means
that the database is already in the desired state. So issue
the same message, but exit without an error code.

fixes #5605

### DIFF
--- a/kong/cmd/utils/migrations.lua
+++ b/kong/cmd/utils/migrations.lua
@@ -170,7 +170,7 @@ end
 local function reset(schema_state, db, ttl)
   if schema_state.needs_bootstrap then
     log("Database not bootstrapped, nothing to reset")
-    return false
+    return true
   end
 
   local opts = {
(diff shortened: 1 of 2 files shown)


bungle:
@hishamhm, what is your take on this as you did this before:
https://github.com/Kong/kong/commit/3f5869afec9e5ae8b7cfa151367967c770e29fa3

hishamhm:
Maybe let's add `-f` to `reset` and `bootstrap` instead? It's still useful to know whether it reset or not.

Tieske:
while trying to implement that I ran into this:
```shell
vagrant@ubuntu-bionic:/kong$ kong migrations bootstrap
Database already bootstrapped
vagrant@ubuntu-bionic:/kong$ echo $?
0
vagrant@ubuntu-bionic:/kong$
```

So `migrations reset -y` will return an error (exit code 1) when the desired state is already achieved. Yet, `migrations bootstrap` will return exit code 0 in that case.

So should we either accept the PR as it was, or also alter `bootstrap` to return an error, to be in sync?

PR #5610, opened 2020-02-25
feat(oauth2) optionally hash client secret

### LINKED ISSUE #1237: Storing sensitive information in database

bungle:
### Summary

This was originally implemented with #4866 by @janza.

After some review comments, and the introduction of DAO transformations, I decided to make changes, thus opening this new PR.

This PR adds a new `boolean` column `hash_secret` to `oauth2_credentials` that is used to determine whether or not the `client_secret` will be hashed.

The PR adds support for `pbkdf2` (and contains code for `argon2`, and `bcrypt` if we want to enable them later). The plugin will also check if the `client_secret` needs to be rehashed (on usage). One caveat. If you run this on cluster and some nodes have `argon2` and some don't, it is possible that you cannot use the credentials on those that don't. So keep your environments similar.

### Issues resolved

Close #4866, Fix #1237 (at least on OAuth2)

### DIFF
--- a/kong/plugins/oauth2/migrations/004_200_to_210.lua
+++ b/kong/plugins/oauth2/migrations/004_200_to_210.lua
@@ -21,6 +21,13 @@ return {
       EXCEPTION WHEN DUPLICATE_COLUMN THEN
         -- Do nothing, accept existing state
       END$$;
+
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY oauth2_credentials ADD hash_secret BOOLEAN;
+      EXCEPTION WHEN DUPLICATE_COLUMN THEN
+        -- Do nothing, accept existing state
+      END$$;
     ]],
   },
 
@@ -29,6 +36,7 @@ return {
       ALTER TABLE oauth2_authorization_codes ADD challenge text;
       ALTER TABLE oauth2_authorization_codes ADD challenge_method text;
       ALTER TABLE oauth2_credentials ADD client_type text;
+      ALTER TABLE oauth2_credentials ADD hash_secret boolean;
     ]],
   },
 }
--- a/kong/plugins/oauth2/secret.lua
+++ b/kong/plugins/oauth2/secret.lua
@@ -1,0 +1,347 @@
+local utils = require "kong.tools.utils"
+
+
+local type = type
+local fmt = string.format
+local find = string.find
+local pcall = pcall
+local remove = table.remove
+local concat = table.concat
+local assert = assert
+local tonumber = tonumber
+local encode_base64 = ngx.encode_base64
+local decode_base64 = ngx.decode_base64
+
+
+local ENABLED_ALGORITHMS = {
+  ARGON2 = false,
+  BCRYPT = false,
+  PBKDF2 = true,
+  SCRYPT = false,
+}
+
+--local CORES
+--do
+--  local infos = utils.get_system_infos()
+--  if type(infos) == "table" then
+--    CORES = infos.cores
+--  end
+--  if not CORES then
+--    CORES = ngx.worker.count() or 1
+--  end
+--end
+
+
+local function infer(value)
+  value = utils.strip(value)
+  return tonumber(value, 10) or value
+end
+
+
+local function parse_phc(phc)
+  local parts = utils.split(phc, "$")
+  local count = #parts
+  if count < 2 or count > 5 then
+    return nil, "invalid phc string format"
+  end
+
+  local id = parts[2]
+  local id_parts = utils.split(id, "-")
+  local id_count = #id_parts
+
+  local prefix
+  local digest
+  if id_count == 1 then
+    prefix = id_parts[1]
+  else
+    prefix = id_parts[1]
+    remove(id_parts, 1)
+    digest = concat(id_parts, "-")
+  end
+
+  local params = {}
+  local prms = parts[3]
+  if prms then
+    local prm_parts = utils.split(prms, ",")
+    for i = 1, #prm_parts do
+      local param = prm_parts[i]
+      local kv = utils.split(param, "=")
+      local kv_count = #kv
+      if kv_count == 1 then
+        params[#params + 1] = infer(kv[1])
+      elseif kv_count == 2 then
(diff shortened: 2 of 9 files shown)


kong/plugins/oauth2/migrations/004_200_to_210.lua
```diff
@@ -0,0 +1,18 @@
+return {
+  postgres = {
+    up = [[
+      DO $$
+      BEGIN
+        ALTER TABLE IF EXISTS ONLY oauth2_credentials ADD hash_secret BOOLEAN DEFAULT FALSE;
```
kikito: I believe adding the default on fields on the migrations like this is not the recommended way, because then it can conflict with the schema default in the future.

```suggestion
        ALTER TABLE IF EXISTS ONLY oauth2_credentials ADD hash_secret BOOLEAN;
```

kong/plugins/oauth2/secret.lua
```diff
@@ -0,0 +1,159 @@
...
+--local CORES
+--do
+--  local infos = utils.get_system_infos()
+--  if type(infos) == "table" then
+--    CORES = infos.cores
+--  end
+--  if not CORES then
+--    CORES = ngx.worker.count() or 1
+--  end
+--end
```
kikito: Is this a vestige from development?

kong/plugins/oauth2/secret.lua
```diff
@@ -0,0 +1,159 @@
...
+
+
+local PREFIX = nil -- currently chosen algorithm
+
+
+local ARGON2
+local ARGON2_ID = "$argon2"
+do
+  local ARGON2_PREFIX
+  local ok, crypt = pcall(function()
```
kikito: I suggest not wrapping the whole declaration with `pcall`. Consider what will happen if someone modifies this code in the future and introduces a typo which makes this function always fail. Then the app will always pick bcrypt. If the `pcall` is supposed to detect that `require "argon2"` throws, then wrap only that with `pcall`.

kong/plugins/oauth2/secret.lua
```diff
@@ -0,0 +1,159 @@
...
+    PREFIX = PREFIX or ARGON2_PREFIX
+  end
+end
+
+
+local BCRYPT
+local BCRYPT_ID = "$2"
+do
+  local BCRYPT_PREFIX
+  local ok, crypt = pcall(function()
```
kikito: Same as before: wrap only the `require` with `pcall`.

p0pr0ck5:
Why are we supporting more than one hash algorithm? It seems to complicate things substantially and I don’t see it being configurable. And argon it not used anywhere else in this project- let’s not add another dependency if we don’t need it?

p0pr0ck5:
Also, what is the logic behind making hash behavior configurable? We don't have such an option for basic-auth.

kong/plugins/oauth2/secret.lua
bungle: AFAIK `require "argon"` might not fail, and you never know if someone decides to update dependency and it fails in different place?

bungle:
> Why are we supporting more than one hash algorithm? It seems to complicate things substantially and I don’t see it being configurable. And argon it not used anywhere else in this project- let’s not add another dependency if we don’t need it?

`bcrypt` is also not used anywhere. The reason was:

1. argon2id is recommended and found on NIST (EE can ship it)
2. bcrypt will install properly by with just luarocks and not NIST compatible

bungle:
> Also, what is the logic behind making hash behavior configurable? We don't have such an option for basic-auth.

In many places like:
1. bintray
2. github

etc.

You can go to your settings and fetch the API key. In some contexts like Github I think the flow is: create app secret and it shows you after creation, and after that it is impossible. So because there is legacy in this plugin we wanted to keep it optional and backward compatible and support both scenarios.

Both scenarios have good and bad sides.

bungle:
> Also, what is the logic behind making hash behavior configurable? We don't have such an option for basic-auth.

For basic-auth I think we didn't even think about that during that point. It is also using hashing function that is not particularly made for hashing passwords. Also it is rare to see any use to store username and password in plain (e.g. with keycloak I cannot find my password anywhere, but I can see clients and their secrets).

kong/plugins/oauth2/secret.lua
bungle: No, I left it there if we want Argon2 parallelism be adjusted by the number of cores. It is there if we at somepoint decide to do so.

kong/plugins/oauth2/secret.lua
bungle: I am not sure what is the benefit (only seeing it more flaky). These only run once on first require.

bungle:
At somepoint if @fffonion can provide us password hashing functions with `lua-resty-openssl`, I guess we can remove these dependencies, though we might need to write migrations then.

fffonion:
@bungle As you wish, lua-resty-openssl 0.6.0rc0 now has [kdf](https://github.com/fffonion/lua-resty-openssl/tree/0.6.0rc0#kdfderive) which supports PBKDF2 and others. 😄

bungle:
@fffonion, ok perhaps we should then wait for `lua-resty-openssl` bump.

fffonion:
@bungle I just released 0.6.0, the kdf api should be available now : )

bungle:
This is now updated to use `pbkdf2` with `lua-resty-openssl`. Other algorithms are disabled, but can be enabled if we need to support them at some point. The new dependencies were removed from `.rockspec` (so `brypt` or `argon2-ffi` is not included anymore there).

PBKDF2 will output hash like this (contains default settings):
```
$pbkdf2-sha512$i=10000,l=32$RsW9il33SHbxQzywGmNkRw$EnvHL59Y0vyxTmjuiArdi6Kmxk3/6TrM1q9bIR3d/zw
```

PR #5613, opened 2020-02-26
chore(ci) github actions via kbt

hutchic:
This commit makes Kong able to run on GitHub Actions, but still kept the
Travis CI files around so that we could use both at the same time during
the transitional period. Eventually Travis CI support will be removed.

This PR further splits Kong core integration tests into two halves on
GitHub Actions to further reduce the run time by utilizing all the
concurrency we have access to, as the core integration tests has
consistently been the long tail of all the jobs, running 2x slower than
the second slowest job. Splitting is not done on Travis as the
concurrency limit makes doing so pointless.

Todo:
- [ ] merge the kong-build-tools branch
- [ ] release a kong-build-tools version
- [ ] update the pinned kong-build-tools version

### DIFF
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -12,7 +12,7 @@ export BUSTED_ARGS="--no-k -o htest  -v --exclude-tags=flaky,ipv6"
 
 if [ "$KONG_TEST_DATABASE" == "postgres" ]; then
     export TEST_CMD="bin/busted $BUSTED_ARGS,cassandra,off"
-    psql -v ON_ERROR_STOP=1 -h ${KONG_PG_HOST} -U ${KONG_PG_USER} -d ${KONG_PG_DATABASE} <<-EOSQL
+    psql -v ON_ERROR_STOP=1 -h ${KONG_TEST_PG_HOST} -U ${KONG_TEST_PG_USER} -d ${KONG_TEST_PG_DATABASE} <<-EOSQL
         CREATE user ${KONG_TEST_PG_USER}_ro;
         GRANT CONNECT ON DATABASE $KONG_TEST_PG_DATABASE TO ${KONG_TEST_PG_USER}_ro;
         \c $KONG_TEST_PG_DATABASE;
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -92,7 +92,7 @@ if [ "$TEST_SUITE" == "plugins" ]; then
     fi
 fi
 if [ "$TEST_SUITE" == "pdk" ]; then
-    TEST_NGINX_RANDOMIZE=1 prove -I. -j$JOBS -r t/01-pdk/
+    TEST_NGINX_RANDOMIZE=1 prove -I. -j$JOBS -r t/01-pdk
 fi
 if [ "$TEST_SUITE" == "unit" ]; then
     unset KONG_TEST_NGINX_USER KONG_PG_PASSWORD KONG_TEST_PG_PASSWORD
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -ex
 
 function cyan() {
     echo -e "\033[1;36m$*\033[0m"
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -101,6 +101,9 @@ fi
 if [ "$TEST_SUITE" == "pdk" ]; then
     TEST_NGINX_RANDOMIZE=1 prove -I. -j$JOBS -r t/01-pdk
 fi
+if [ "$TEST_SUITE" == "lint" ]; then
+    make lint
+fi
 if [ "$TEST_SUITE" == "unit" ]; then
     scripts/autodoc-admin-api
     bin/busted -v -o gtest spec/01-unit
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -101,11 +101,9 @@ fi
 if [ "$TEST_SUITE" == "pdk" ]; then
     TEST_NGINX_RANDOMIZE=1 prove -I. -j$JOBS -r t/01-pdk
 fi
-if [ "$TEST_SUITE" == "lint" ]; then
-    make lint
-fi
 if [ "$TEST_SUITE" == "unit" ]; then
     unset KONG_TEST_NGINX_USER KONG_PG_PASSWORD KONG_TEST_PG_PASSWORD
     scripts/autodoc-admin-api
     bin/busted -v -o gtest spec/01-unit
+    make lint
 fi
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -105,6 +105,7 @@ if [ "$TEST_SUITE" == "lint" ]; then
     make lint
 fi
 if [ "$TEST_SUITE" == "unit" ]; then
+    unset KONG_TEST_NGINX_USER KONG_PG_PASSWORD KONG_TEST_PG_PASSWORD
     scripts/autodoc-admin-api
     bin/busted -v -o gtest spec/01-unit
 fi
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -12,6 +12,13 @@ export BUSTED_ARGS="--no-k -o htest  -v --exclude-tags=flaky,ipv6"
 
 if [ "$KONG_TEST_DATABASE" == "postgres" ]; then
     export TEST_CMD="bin/busted $BUSTED_ARGS,cassandra,off"
+    psql -v ON_ERROR_STOP=1 -h $KONG_PG_HOST --username "$KONG_TEST_PG_USER" <<-EOSQL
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ RESTY_VERSION ?= `grep RESTY_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk
 RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_OPENSSL_VERSION ?= `grep RESTY_OPENSSL_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_PCRE_VERSION ?= `grep RESTY_PCRE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
-KONG_BUILD_TOOLS ?= '3.1.2'
```
hutchic: Need to merge, release and update this before merging this PR

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,133 @@
...
+    runs-on: ubuntu-18.04
+    needs: build
+
+    strategy:
+      matrix:
+        split: [first, second, all]
+        suite: [integration, plugins]
+        database: [cassandra, postgres]
+        exclude:
+          - SUITE: plugins
```
dndx: "SUITE" -> "suite"

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,133 @@
...
+    - name: Setup Kong Build Tools
+      run: make setup-kong-build-tools
+
+    - name: Run Tests
+      run: pushd $KONG_BUILD_TOOLS_LOCATION && make test-kong && popd
+
+  integration-tests-with-databases:
+    name: ${{matrix.database}} (${{ matrix.suite }}) ${{ matrix.split }} integration tests
+    runs-on: ubuntu-18.04
+    needs: build
```
dndx: Same issue here.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,133 @@
...
+        password: ${{ secrets.KONG_PASSWORD }}
+
+    - name: Checkout Kong source code
+      uses: actions/checkout@v2
+
+    - name: Setup Kong Build Tools
+      run: make setup-kong-build-tools
+
+    - name: Build Test Image
+      run: pushd $KONG_BUILD_TOOLS_LOCATION && make kong-test-container && popd
```
dndx: It seems that we are still baking Kong source codes into the Docker container itself, which won't work well for CI environment because we have so many runs with different commit hashes. We should not cache Kong source codes inside Docker hub, but instead run a `make dev` manually inside the container before starting the tests.

hutchic: right now we run `make dev` the precise number of times it's required no more and no less. Running without this cache has downsides:
- speed. Downloading from docker hub saves time over downloading / buildling from source
- reproducible. What one stage tests won't be equivalent to another

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,133 @@
...
+
+  integration-tests-dbless:
+    name: DB-less tests
```
hutchic:
```suggestion
    name: DB-less integration tests
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,133 @@
...
+    runs-on: ubuntu-18.04
+    needs: build
+
+    strategy:
+      matrix:
+        split: [first, second, all]
+        suite: [integration, plugins]
+        database: [cassandra, postgres]
+        exclude:
+          - SUITE: plugins
```
hutchic:
```suggestion
          - suite: plugins
```

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
hutchic: updated

.github/workflows/build_and_test.yml
hutchic: done

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,133 @@
...
+    - name: Checkout Kong source code
+      uses: actions/checkout@v2
+
+    - name: Setup Kong Build Tools
+      run: make setup-kong-build-tools
+
+    - name: Run Tests
+      run: pushd $KONG_BUILD_TOOLS_LOCATION && make test-kong && popd
+
+  tests-without-a-database:
```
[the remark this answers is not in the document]
hutchic: renamed and moved to the top

dndx:
TODO:

https://github.com/Kong/kong/commit/d5596f777c7c7bd91b4b9a70d79dd6604bccf0e0#diff-e3523994103f9011336b6f0fd76fc0e6R120-R129 need to be ported over, or PR will fail with latest `next` branch.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,161 @@
...
+      run: pushd $KONG_BUILD_TOOLS_LOCATION && make test-kong && popd
+
+  pdk:
+    name: PDK
+    runs-on: ubuntu-18.04
+    needs: build
+
+    env:
+      TEST_SUITE: pdk
+      DEBUG: 1
```
hutchic: pdk runs separately as it's the only tests requiring debug

.ci/run_tests.sh
```diff
@@ -8,10 +8,17 @@ function red() {
     echo -e "\033[1;31m$*\033[0m"
 }
 
-export BUSTED_ARGS="-o htest -v --exclude-tags=flaky,ipv6"
+export BUSTED_ARGS="--no-k -o htest  -v --exclude-tags=flaky,ipv6"
 
 if [ "$KONG_TEST_DATABASE" == "postgres" ]; then
     export TEST_CMD="bin/busted $BUSTED_ARGS,cassandra,off"
+    psql -v ON_ERROR_STOP=1 -h ${KONG_TEST_PG_HOST} -U ${KONG_TEST_PG_USER} -d ${KONG_TEST_PG_DATABASE} <<-EOSQL
```
dndx: Should this be added in the KBT Postgres container instead? Adding it here is really not friendly for development environment as people often run `busted` directly on a single test file and not necessarily uses the `.ci/run_tests.sh` for running the tests.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,161 @@
...
+      run: pushd $KONG_BUILD_TOOLS_LOCATION && make kong-test-container && popd
+
+  lint-unit-tests:
+    name: Lint & Unit
+    runs-on: ubuntu-18.04
+    needs: build
+
+    strategy:
+      matrix:
+        suite: [unit, lint]
```
dndx: This does not need to be split further, Lint and Unit should be executed sequentially inside the same job, like Travis and the old PR. Because lint finishes extremely quickly, running it concurrently with `unit` actually slows things down.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,161 @@
...
+    - name: Setup Kong Build Tools
+      run: make setup-kong-build-tools
+
+    - name: Run Tests
+      run: pushd $KONG_BUILD_TOOLS_LOCATION && make test-kong && popd
+
+  pdk:
+    name: PDK
+    runs-on: ubuntu-18.04
+    needs: build
```
dndx: `pdk` should be dependent on `lint-and-unit`, instead of `build`.

hutchic: why is that?

.ci/run_tests.sh
hutchic: I feel like there are pros and cons of both and this is something I debated and even roughed out both ways of doing it.

One requirement is it needs to happen locally when folks run tests as well as travis-ci, Jenkins and Github Actions. I couldn't see a way to place it in KBT that would require duplicating the logic somewhere else as well (?)

Makefile
```diff
@@ -28,7 +28,7 @@ RESTY_VERSION ?= `grep RESTY_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk
 RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_OPENSSL_VERSION ?= `grep RESTY_OPENSSL_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_PCRE_VERSION ?= `grep RESTY_PCRE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
-KONG_BUILD_TOOLS ?= '3.1.4'
+KONG_BUILD_TOOLS ?= 'origin/feat/dev-cache'
```
hutchic:
```suggestion
KONG_BUILD_TOOLS ?= '4.2.0'
```

PR #5620, opened 2020-02-27
feat(proxy/pdk) add support for X-Forwarded-Prefix

### LINKED ISSUE #5602: Header X-Forwarded-Prefix not sent when proxying an API
### Summary

I would expect Kong to set the `X-Forwarded-Prefix` header when proxying an API. This was registered as issue https://github.com/Kong/kong/issues/1150 and fixed with PR https://github.com/Kong/kong/pull/1180. But it seems like X-Forwarded-Prefix part of this fix has been lost (or removed).

I notice the following constant in the code, but it seems unused:
https://github.com/Kong/kong/blob/cb1ed39a288e90801322d3a1c4d940ac08337023/kong/constants.lua#L86

### Steps To Reproduce

1. Set up an echo route (DB-Less configuration)
```yaml
services:
  - name: echo
    url: https://postman-echo.com/get
    routes:
      - name: echo
        paths:
          - /echo
```
2. Perform a HTTP GET request through Kong

#### Expected behavior
The X-Forwarded-Prefix header is echoed with the value `/echo`.

#### Actual behavior
No X-Forwarded-Prefix header in the echo response

### Additional Details & Logs

- Kong version 1.5.1

bungle:
### Summary

Kong currently adds these headers to upstream request:
```
X-Forwarded-For
X-Forwarded-Port
X-Forwarded-Proto
X-Forwarded-Host
X-Real-IP
```

@erikgb mentioned that he would like to have `X-Forwarded-Prefix`
added on #5602.

This PR adds that, and additionally introduces a new PDK function:
```lua
local path = kong.request.get_forwarded_path()
```

### Issues Resolved

Close #5602

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -53,6 +53,7 @@ local function new(self)
   local X_FORWARDED_PROTO      = "X-Forwarded-Proto"
   local X_FORWARDED_HOST       = "X-Forwarded-Host"
   local X_FORWARDED_PORT       = "X-Forwarded-Port"
+  local X_FORWARDED_PREFIX     = "X-Forwarded-Prefix"
 
 
   ---
@@ -149,7 +150,7 @@ local function new(self)
   -- `X-Forwarded-Host` if it comes from a trusted source. The returned value
   -- is normalized to lower-case.
   --
-  -- Whether this function considers `X-Forwarded-Proto` or not depends on
+  -- Whether this function considers `X-Forwarded-Host` or not depends on
   -- several Kong configuration parameters:
   --
   -- * [trusted\_ips](https://getkong.org/docs/latest/configuration/#trusted_ips)
@@ -199,9 +200,9 @@ local function new(self)
   -- **Note**: we do not currently offer support for Forwarded HTTP Extension
   -- (RFC 7239) since it is not supported by ngx_http_realip_module.
   --
-  -- @function kong.request.get_forwareded_port
+  -- @function kong.request.get_forwarded_port
   -- @phases rewrite, access, header_filter, body_filter, log, admin_api
-  -- @treturn number the forwared port
+  -- @treturn number the forwarded port
   -- @usage
   -- kong.request.get_forwarded_port() -- 1234
   function _REQUEST.get_forwarded_port()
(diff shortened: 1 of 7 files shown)


bungle:
> @bungle : Thanks a lot for preparing this PR so fast! I do not have much experience with Lua; I wrote my first line today, in a simple post-function to fix a X-Forward-Port issue. But programming is programming, so I will do my best to comment on this PR. But I will just add a general comment tonight, and will try to look closer at the code tomorrow - or over the weekend.
> 
> The first thing I notice, is that quite a few formatting fixes have been included in this commit. I tend to prefer to have _refactoring_, including _formatting_, separated from _changes_. Would you consider to split the source code changes in two commits? Doing that, it will also be easier to review the PR - when the changes related to this new feature is separated from the re-formatting. Any potential merge/rebase conflict also tends to be easier to resolve (correctly) doing so....

You are right, but Github also has ignore whitespace when reviewing.

kong/pdk/request.lua
```diff
@@ -235,6 +236,41 @@ local function new(self)
...
+  --
+  -- Whether this function considers `X-Forwarded-Prefix` or not depends on
+  -- several Kong configuration parameters:
+  --
+  -- * [trusted\_ips](https://getkong.org/docs/latest/configuration/#trusted_ips)
+  -- * [real\_ip\_header](https://getkong.org/docs/latest/configuration/#real_ip_header)
+  -- * [real\_ip\_recursive](https://getkong.org/docs/latest/configuration/#real_ip_recursive)
+  --
+  -- **Note**: we do not currently do any normalization on the request
+  --           path except return `"/"` on empty path.
```
kikito: This seems false or incomplete, there is some normalization going on on the last line of this function.

kong/runloop/handler.lua
```diff
@@ -1041,14 +1042,27 @@ return {
...
-        forwarded_host  = host
-        forwarded_port  = port
+        forwarded_proto  = scheme
+        forwarded_host   = host
+        forwarded_port   = port
+      end
+
+      if not forwarded_prefix then
+        forwarded_prefix = var.request_uri
+        local p = find(forwarded_prefix, "?", 2, true)
```
kikito: This removal of the query seems to be done here but not on the pdk function. Shouldn't they behave the same way?

bungle: PDK also does it, see `get_forwarded_path`.

kong/pdk/request.lua
bungle: Thus it says:
> except return `"/"` on empty path.

Am I missing something?

spec/02-integration/05-proxy/03-upstream_headers_spec.lua
```diff
@@ -461,7 +484,25 @@ for _, strategy in helpers.each_strategy() do
...
         end)
+      end)
+
+      describe("X-Forwarded-Prefix", function()
+        it("should be added if not present in request", function()
+          local headers = request_headers {
+            ["Host"] = "headers-inspect.com",
+          }
+
+          assert.equal(helpers.get_proxy_port(false), tonumber(headers["x-forwarded-port"]))
```
erikgb: I think this assert should assert x-forwarded-prefix, and not x-forwarded-port.

kong/pdk/request.lua
erikgb: I agree with @bungle that the note seems in accordance with the code. But I do not understand why this normalization is necessary - but there is probably a reason for it. ;-) What would be wrong with returning just path (even emtpy string), in case the X-Forwarded-Prefix header is not set by a trusted client?

kong/runloop/handler.lua
erikgb: What about introducing a new method to obtain the request path (`get_request_path`)? That would make the code structure similar to the corresponding code in the PDK.

bungle: @erikgb, I was referring the existing `kong.request.get_path()` that this method calls.

kong/pdk/request.lua
bungle: that is the problem of these proprietary headers. There is no documentation. So I checked what others do and came across this: https://github.com/qaware/x-forwarded-filter

e.g. it normalizes request url like:
`/a/b/../` to `/a`. I considered it but then just went with "" -> "/". We can add more later if needed, but we need to be careful to not mess up with semantics.

kong/runloop/handler.lua
erikgb: @bungle : I know, and that is why I am suggesting to introduce a corresponding method on the handler - to avoid any misunderstanding. I agree with your response to @kikito comment; the handler and the PDK function seems to do the same thing.

erikgb:
Anything more I can do to get this merged? I have reviewed the code, and I think it looks good. @kikito has requested changes, but I am not sure what changes are actually required.

kong/runloop/handler.lua
```diff
@@ -1065,9 +1079,9 @@ return {
         end
 
         if redirect_status_code == 301 or
-          redirect_status_code == 302 or
-          redirect_status_code == 307 or
-          redirect_status_code == 308 then
+           redirect_status_code == 302 or
+           redirect_status_code == 307 or
+           redirect_status_code == 308 then
```
javierguerragiraldez: nitpick: unindent the `then`

PR #5628, opened 2020-02-29
Strip Authorization value from logged headers

rainest:
### Summary

Replace the value of the `Authorization` header with `REDACTED` in the basic serializer.

### Full changelog

* Updated the basic serializer to replace the value of the `Authorization` header with the string `REDACTED` if present. This still indicates whether the client sent the header while not recording the actual credential in the log system.
* Update serializer unit tests to confirm this behavior and use a more realistic header table.

### Notes
As this is technically a new feature, and modifies existing existing functionality, PRing to next, though I think we could reasonably apply it to master and count it as bugfix instead depending on whether we think anyone relies on the existing behavior here (I don't think anyone should). The diff applies equally on either.

Including the actual value of `Authorization` is risky, as it will persist credentials to some log store. That store is not necessarily encrypted (and even if is, may be storing credentials normally stored in a one-way hash with reversible encryption) and may have different access controls than the normal credential store.

We furthermore already log the credential ID in .authenticated_entity.id for Kong-managed credentials, so at least some of the information lost with this change is available via other means.

### DIFF
--- a/kong/plugins/log-serializers/basic.lua
+++ b/kong/plugins/log-serializers/basic.lua
@@ -5,6 +5,8 @@ local gkong = kong
 local _M = {}
 
 local EMPTY = tablex.readonly({})
+local REDACTED_REQUEST_HEADERS = { "authorization", "proxy-authorization" }
+local REDACTED_RESPONSE_HEADERS = {}
 
 function _M.serialize(ngx, kong)
   local ctx = ngx.ctx
@@ -36,8 +38,17 @@ function _M.serialize(ngx, kong)
   local request_uri = var.request_uri or ""
 
   local req_headers = kong.request.get_headers()
-  if req_headers["authorization"] then
-    req_headers["authorization"] = "REDACTED"
+  for _, header in ipairs(REDACTED_REQUEST_HEADERS) do
+    if req_headers[header] then
+      req_headers[header] = "REDACTED"
+    end
+  end
+
+  local resp_headers = ngx.resp.get_headers()
+  for _, header in ipairs(REDACTED_RESPONSE_HEADERS) do
+    if resp_headers[header] then
+      resp_headers[header] = "REDACTED"
+    end
   end
 
   return {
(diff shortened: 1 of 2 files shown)


rainest:
PR CI failure appears to have been some issue with grpcbin: the same test succeeded on the push with only one commit, and succeeded for the PR with Postgres (there shouldn't be any meaningful difference between them). Please re-run if needed.

hishamhm:
@rainest While we're doing this, what's your take on logging cookies? Should we drop those as well?

Any other standard header comes to mind?

rainest:
Of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Authentication we may encounter Proxy-Authorization, although unlikely--should be only in cases where there's some other upstream proxy that needs it. Should be fine to exclude as well, since it's only credentials.

Cookie and Set-Cookie are a harder call, as they can contain both credentials and generic information that affects non-authentication parts of the response (e.g. that a user has set their preferred language to German). I'd lean towards dropping them to be on the safe side and working towards a more tunable system depending on whether anyone values the ability to log them.

hishamhm:
> Of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Authentication we may encounter Proxy-Authorization, although unlikely--should be only in cases where there's some other upstream proxy that needs it. Should be fine to exclude as well, since it's only credentials.

Agreed, this seems like a safe bet to exclude.

> Cookie and Set-Cookie are a harder call, as they can contain both credentials and generic information that affects non-authentication parts of the response (e.g. that a user has set their preferred language to German). I'd lean towards dropping them to be on the safe side and working towards a more tunable system depending on whether anyone values the ability to log them.

Yeah, cookies sound like a case-by-case thing, I can imagine people trying to debug applications and getting annoyed by them ceasing to be available. I wouldn't mind merging this with the authorization exclusions only, as an incremental improvement, and then go for the tunable system where cookie (or other) headers can be enabled/disabled as a future step. @bungle, what do you think?

rainest:
Updated to strip Proxy-Authorization and make the implementation a bit more flexible and leave it open to adding cookie headers later.

PR #5629, opened 2020-02-29
feat(stream) basic serializer to work on streams

bungle:
### Summary

Kong logging plugins in general use so called `basic serializer`. This serializer has been written originally for `http`. This PR makes the serializer work with `streams`. It means that all our logging plugins will start to work with `streams` as well.

There is one caveat though. As streams run `log` phase after the stream is closed, it means that `streams` that are kept open, never log. We can perhaps pursue this later in separate PR.


dndx:
Superseded by https://github.com/Kong/kong/pull/6036.

PR #5640, opened 2020-03-05
feat(ip-restriction) add support for IPv6

murillopaula:
Add IPv6 support for the ip-restriction plugin.

Looking for feedback so I can see if it would be a good idea to replace the 'lua-resty-iputils' and 'lua-resty-mediador' libraries by the 'lua-resty-ipmatcher' one.

The iputils lib is only used for typedef validations and the mediador lib is used for trusted ips on the pdk side and for sources and destinations on the router side.

- [feat(ip-restriction) add support for IPv6](https://github.com/Kong/kong/pull/5640/commits/fb76a01bfcea55e50353402204be6b2c62eebd3a)

### DIFF
--- a/kong-2.0.2-0.rockspec
+++ b/kong-2.0.2-0.rockspec
@@ -37,6 +37,7 @@ dependencies = {
   "lua-messagepack == 0.5.2",
   "lua-resty-openssl == 0.5.3",
   "lua-resty-counter == 0.2.0",
+  "lua-resty-ipmatcher == 0.5",
   -- external Kong plugins
   "kong-plugin-azure-functions ~> 0.4",
   "kong-plugin-zipkin ~> 0.2",
--- a/kong-2.0.2-0.rockspec
+++ b/kong-2.0.2-0.rockspec
@@ -37,7 +37,7 @@ dependencies = {
   "lua-messagepack == 0.5.2",
   "lua-resty-openssl == 0.5.3",
   "lua-resty-counter == 0.2.0",
-  "lua-resty-ipmatcher == 0.5",
+  "lua-resty-ipmatcher == 0.6",
   -- external Kong plugins
   "kong-plugin-azure-functions ~> 0.4",
   "kong-plugin-zipkin ~> 0.2",
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -8,7 +8,6 @@ local Schema = require("kong.db.schema")
 local socket_url = require("socket.url")
 local constants = require "kong.constants"
 local px = require "resty.mediador.proxy"
-local ipmatcher = require "resty.ipmatcher"
 
 
 local pairs = pairs
@@ -70,17 +69,6 @@ local function validate_cidr_v4(ip)
   -- It's an error only if the second variable is a string
   if type(err) == "string" then
     return nil, "invalid cidr range: " .. err
-  end
-
-  return true
-end
-
-
-local function validate_cidr_v4_v6(ip)
-  local _, err = ipmatcher.new({ ip })
-
-  if err then
-    return nil, "failed to create new ipmatcher instance: " .. err
   end
 
   return true
@@ -256,11 +244,6 @@ typedefs.ip_or_cidr = Schema.define {
 typedefs.cidr_v4 = Schema.define {
   type = "string",
   custom_validator = validate_cidr_v4,
-}
-
-typedefs.cidr_v4_v6 = Schema.define {
-  type = "string",
-  custom_validator = validate_cidr_v4_v6,
 }
 
 -- deprecated alias
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -8,6 +8,7 @@ local Schema = require("kong.db.schema")
 local socket_url = require("socket.url")
 local constants = require "kong.constants"
 local px = require "resty.mediador.proxy"
+local ipmatcher = require "resty.ipmatcher"
 
 
 local pairs = pairs
@@ -69,6 +70,17 @@ local function validate_cidr_v4(ip)
   -- It's an error only if the second variable is a string
   if type(err) == "string" then
     return nil, "invalid cidr range: " .. err
+  end
+
+  return true
+end
+
+
+local function validate_cidr_v4_v6(ip)
+  local _, err = ipmatcher.new({ ip })
+
+  if err then
+    return nil, "failed to create new ipmatcher instance: " .. err
   end
 
   return true
@@ -244,6 +256,11 @@ typedefs.ip_or_cidr = Schema.define {
 typedefs.cidr_v4 = Schema.define {
   type = "string",
   custom_validator = validate_cidr_v4,
+}
+
+typedefs.cidr_v4_v6 = Schema.define {
+  type = "string",
+  custom_validator = validate_cidr_v4_v6,
 }
 
 -- deprecated alias
--- a/kong/plugins/ip-restriction/handler.lua
+++ b/kong/plugins/ip-restriction/handler.lua
@@ -37,8 +37,8 @@ function IpRestrictionHandler:access(conf)
   end
 
   if err then
-    block = true
-    kong.log.err("invalid binary IP address: " .. err)
+    kong.log.err("Invalid binary IP address: " .. err)
+    return kong.response.exit(500, { message = "An unexpected error occurred" })
   end
 
   if block then
--- a/kong/plugins/ip-restriction/handler.lua
+++ b/kong/plugins/ip-restriction/handler.lua
@@ -17,25 +17,22 @@ function IpRestrictionHandler:access(conf)
   local block = false
   local err = false
   local binary_remote_addr = ngx.var.binary_remote_addr
+  local ip
 
   if not binary_remote_addr then
     return kong.response.exit(FORBIDDEN, { message = "Cannot identify the client IP address, unix domain sockets are not supported." })
   end
 
   if conf.blacklist and #conf.blacklist > 0 then
-    if not kong.ctx.plugin.blacklist then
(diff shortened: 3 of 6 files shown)


kong/db/schema/typedefs.lua
```diff
@@ -75,6 +76,17 @@ local function validate_cidr_v4(ip)
 end
 
 
+local function validate_cidr_v4_v6(ip)
```
bungle: How is this different from `local function validate_ip_or_cidr(ip)` above? I know it uses different library, but both accept `ip` and `cidr`. So could you just use the `typedefs.ip_or_cidr` with `ip-restriction` and skip adding this for now? We will rework these when removing other deps.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,68 +1,47 @@
...
+  local err = false
   local binary_remote_addr = ngx.var.binary_remote_addr
 
   if not binary_remote_addr then
     return kong.response.exit(FORBIDDEN, { message = "Cannot identify the client IP address, unix domain sockets are not supported." })
   end
 
   if conf.blacklist and #conf.blacklist > 0 then
-    block = iputils.binip_in_cidrs(binary_remote_addr, cidr_cache(conf.blacklist))
+    if not kong.ctx.plugin.blacklist then
```
bungle: When would this hold `true`?

kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,68 +1,47 @@
...
+      kong.ctx.plugin.blacklist = ipmatcher.new(conf.blacklist)
+    end
+
+    block, err = kong.ctx.plugin.blacklist:match_bin(binary_remote_addr)
   end
 
   if conf.whitelist and #conf.whitelist > 0 then
-    block = not iputils.binip_in_cidrs(binary_remote_addr, cidr_cache(conf.whitelist))
+    if not kong.ctx.plugin.whitelist then
```
bungle: When would this hold `true`?

kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,68 +1,47 @@
...
   local binary_remote_addr = ngx.var.binary_remote_addr
 
   if not binary_remote_addr then
     return kong.response.exit(FORBIDDEN, { message = "Cannot identify the client IP address, unix domain sockets are not supported." })
   end
 
   if conf.blacklist and #conf.blacklist > 0 then
-    block = iputils.binip_in_cidrs(binary_remote_addr, cidr_cache(conf.blacklist))
+    if not kong.ctx.plugin.blacklist then
+      kong.ctx.plugin.blacklist = ipmatcher.new(conf.blacklist)
```
bungle: What is the reason to set the context value?

kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,68 +1,47 @@
...
+    end
+
+    block, err = kong.ctx.plugin.blacklist:match_bin(binary_remote_addr)
   end
 
   if conf.whitelist and #conf.whitelist > 0 then
-    block = not iputils.binip_in_cidrs(binary_remote_addr, cidr_cache(conf.whitelist))
+    if not kong.ctx.plugin.whitelist then
+      kong.ctx.plugin.whitelist = ipmatcher.new(conf.whitelist)
```
bungle: What is the reason to set the context value?

kong/plugins/ip-restriction/schema.lua
```diff
@@ -8,8 +8,8 @@ return {
     { config = {
         type = "record",
         fields = {
-          { whitelist = { type = "array", elements = typedefs.cidr_v4, }, },
-          { blacklist = { type = "array", elements = typedefs.cidr_v4, }, },
+          { whitelist = { type = "array", elements = typedefs.cidr_v4_v6, }, },
```
bungle: Here as I commented above, you could just use:
```
typedefs.ip_or_cidr
```

kong/plugins/ip-restriction/schema.lua
```diff
@@ -8,8 +8,8 @@ return {
...
+          { blacklist = { type = "array", elements = typedefs.cidr_v4_v6, }, },
```
bungle: Here as I commented above, you could just use:
```
typedefs.ip_or_cidr
```

kong/plugins/ip-restriction/handler.lua
murillopaula: `kong.ctx.*` should not be used to cache `ipmatcher` object instances

kong/plugins/ip-restriction/handler.lua
murillopaula: `kong.ctx.*` should not be used to cache `ipmatcher` object instances

kong/plugins/ip-restriction/handler.lua
murillopaula: `kong.ctx.*` should not be used to cache `ipmatcher` object instances

kong/plugins/ip-restriction/handler.lua
murillopaula: `kong.ctx.*` should not be used to cache `ipmatcher` object instances

kong/db/schema/typedefs.lua
murillopaula: we can break existing third party plugins using `validate_ip_or_cidr`

the new `ipmatcher` lib doesn't validate stuff like this:

```
{ whitelist = { "0.0.0.0/a", "0.0.0.0/-1", "0.0.0.0/33" }
```

i.e. `ipmatcher` doesn't complain on wrong cidrs... this issue has been reported at https://github.com/api7/lua-resty-ipmatcher/issues/11

we can also break the error message, but that is a very minor problem

kong/plugins/ip-restriction/schema.lua
murillopaula: we can break existing third party plugins using `validate_ip_or_cidr` (look at my comment above)

kong/plugins/ip-restriction/schema.lua
murillopaula: we can break existing third party plugins using validate_ip_or_cidr (view comment above)

kong/db/schema/typedefs.lua
bungle: I don't think we break any existing plugins that use `validate_ip_or_cidr` as we ARE NOT CHANGING it. That typedefs.lua stays the same. This already catches invalid cidrs. We just change the ip-restriction plugin to use `validate_ip_or_cidr`. That validation function DOES NOT use `ipmatcher` but that mediador, and that is fine. The implementation inside `handler.lua` uses `ipmatcher`.

kong/plugins/ip-restriction/schema.lua
bungle: no we cannot as WE ARE NOT CHANGING typedef. Only the schema of ip-restriction plugin.

kong/db/schema/typedefs.lua
```diff
@@ -75,6 +76,17 @@ local function validate_cidr_v4(ip)
...
+  local _, err = ipmatcher.new({ ip })
+
```
bungle:
```suggestion
```

kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,68 +1,44 @@
...
 
   if not binary_remote_addr then
     return kong.response.exit(FORBIDDEN, { message = "Cannot identify the client IP address, unix domain sockets are not supported." })
   end
 
   if conf.blacklist and #conf.blacklist > 0 then
-    block = iputils.binip_in_cidrs(binary_remote_addr, cidr_cache(conf.blacklist))
+    ip = ipmatcher.new(conf.blacklist)
+
+    block, err = ip:match_bin(binary_remote_addr)
```
bungle: Pseudo:
`if err then return kong.response.exit(500, ...) end`

kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,68 +1,44 @@
...
   end
 
   if conf.whitelist and #conf.whitelist > 0 then
-    block = not iputils.binip_in_cidrs(binary_remote_addr, cidr_cache(conf.whitelist))
+    ip = ipmatcher.new(conf.whitelist)
+
+    block, err = ip:match_bin(binary_remote_addr)
```
bungle: Pseudo:
`if err then return kong.response.exit(500, ...) end`

kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,68 +1,44 @@
...
+    block = not block
+  end
+
+  if err then
+    block = true
+    kong.log.err("invalid binary IP address: " .. err)
   end
```
bungle: remove

kong/plugins/ip-restriction/schema.lua
bungle: Again, just use `typedefs.ip_or_cidr` here.

bungle: Again, just use `typedefs.ip_or_cidr` here.

kong-2.0.2-0.rockspec
```diff
@@ -37,6 +37,7 @@ dependencies = {
   "lua-messagepack == 0.5.2",
   "lua-resty-openssl == 0.5.3",
   "lua-resty-counter == 0.2.0",
+  "lua-resty-ipmatcher == 0.5",
```
bungle: They just released `0.6`, so could we update to that too :-(. Sorry.

murillopaula:
Thank you @bungle for reviewing! You are awesome. Thanks for identifying problems and proposing solutions so fast.

(First GitHub comment sent from the GitHub Android app)

PR #5643, opened 2020-03-06
Adds developer docs wrt test helpers

Tieske:
The docs itself are not included, but the base files to generate are.

The updates to the helpers are only additional doc comments, and some reordering. No functional changes.

Results are not perfect, but seem complete.

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -61,9 +61,11 @@ log.set_lvl(log.levels.quiet) -- disable stdout logs in tests
 -- (while running from the busted environment)
 package.path = CUSTOM_PLUGIN_PATH .. ";" .. package.path
 
+--- Returns the OpenResty version.
 -- Extract the current OpenResty version in use and returns
 -- a numerical representation of it.
--- Ex: 1.11.2.2 -> 11122
+-- Ex: `1.11.2.2` -> `11122`
+-- @name openresty_ver_num
 local function openresty_ver_num()
   local nginx_bin = assert(nginx_signals.find_nginx_bin())
   local _, _, _, stderr = pl_utils.executeex(string.format("%s -V", nginx_bin))
@@ -76,16 +78,18 @@ local function openresty_ver_num()
   return tonumber(a .. b .. c .. d)
 end
 
--- Unindent a multi-line string for proper indenting in
+--- Unindent a multi-line string for proper indenting in
 -- square brackets.
+-- @name unindent
+-- @usage
+-- local u = helpers.unindent
 --
--- Ex:
---   u[[
---       hello world
---       foo bar
---   ]]
+-- u[[
+--     hello world
+--     foo bar
+-- ]]
 --
--- will return: "hello world\nfoo bar"
+-- -- will return: "hello world\nfoo bar"
 local function unindent(str, concat_newlines, spaced_newlines)
   str = string.match(str, "(.-%S*)%s*$")
   if not str then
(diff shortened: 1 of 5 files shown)


spec/helpers.lua
```diff
@@ -186,6 +240,32 @@ local function bootstrap_database(db)
...
+--- Gets the database utility helpers and prepares the database for a testrun.
+-- This will a.o. bootstrap the datastore and truncate the existing data that
+-- migth be in it. The BluePrint returned can be used to create test entities
+-- in the database.
+-- @name get_db_utils
+-- @param strategy (optional) the database strategy to use, will default to the
+-- strategy in the test configuration.
+-- @param tables (optional) tables to truncate, this can be used to accelarate
+-- tests if only a few tables are used. By default all tables will be truncated.
+-- @param plugins (optional) array of plugins to load
```
kikito:
```suggestion
-- @param plugins (optional) array of plugins to mark as loaded. Since kong will load all the bundled plugins by default, this is useful for mostly for marking custom plugins as loaded.
```

spec/helpers.lua
```diff
@@ -269,7 +352,10 @@ local pack = function(...) return { n = select("#", ...), ... } end
 local unpack = function(t) return unpack(t, 1, t.n) end
 
 --- Prints all returned parameters.
--- Simple debugging aid.
+-- Simple debugging aid, it will pass all received parameters, hence will not
+-- influence the flow of the code. See also `fail`.
```
kikito: That section is already there in `@see fail`
```suggestion
-- influence the flow of the code.
```

spec/helpers.lua
```diff
@@ -540,19 +738,195 @@ local function admin_ssl_client(timeout)
...
 
---- Starts a TCP server.
--- Accepts a single connection (or multiple, if given opts.requests)
+
+--- Starts a local TCP server.
+-- Accepts a single connection (or multiple, if given `opts.requests`)
 -- and then closes, echoing what was received (last read, in case
 -- of multiple requests).
+--
+-- Check the `kill_tcp_server` function for return values.
```
kikito: This seems in contradiction with the @return clause below it (the return values seem different in both functions). So maybe just remove this line?

```suggestion
```

spec/helpers.lua
```diff
@@ -1304,6 +1733,8 @@ Expected CN:
 Got instead:
 %s
 ]])
+-- TODO: this seems broken, there seem to be no tests. A nicer assertion would be
+-- assert.certificate.has.cn("somename.com")
```
kikito: If it's broken and no test is using it, I propose we remove it from the codebase instead.

spec/helpers.lua
```diff
@@ -1473,13 +1944,13 @@ end
...
+-- @param cmd command string to execute
 -- @param pl_returns (optional) boolean: if true, this function will
 -- return the same values as Penlight's executeex.
--- @return if pl_returns is true, returns four return values
--- (ok, code, stdout, stderr); if pl_returns is false,
+-- @return if `pl_returns` is true, returns four return values
+-- (ok, code, stdout, stderr); if `pl_returns` is false,
 -- returns either (false, stderr) or (true, stderr, stdout).
-local function exec(cmd, pl_returns)
+function exec(cmd, pl_returns)
```
kikito: This seems to be creating a global variable called `exec`? Is it on purpose?

```suggestion
local function exec(cmd, pl_returns)
```

spec/helpers.lua
Tieske: it makes sense to keep it around imo, but should be fixed at some point, hence the "todo" marker

spec/helpers.lua
Tieske: no, there is a forward declaration here: https://github.com/Kong/kong/blob/a606c2a0aa74f862ad71dc3622bfada3f564c3f5/spec/helpers.lua#L830

there are some refrences that would otherwise fail, hence the forward declaration

spec/helpers.lua
Tieske: This is more explicit imo, lmk if you want me to remove it

spec/helpers.lua
Tieske:
```
assert.certificate(cert).has.cn("somename.com")
```

PR #5664, opened 2020-03-11
Kong accesses to DB on the first request after restart even when `db_cache_warmup` is set (upstreams and targets entities)

murillopaula:


kong/runloop/balancer.lua
```diff
@@ -108,7 +103,7 @@ do
   _load_upstream_into_memory = load_upstream_into_memory
 
   get_upstream_by_id = function(upstream_id)
-    local upstream_cache_key = "balancer:upstreams:" .. upstream_id
+    local upstream_cache_key = "upstreams:" .. upstream_id
```
murillopaula: @bungle why did this line in the balancer used a different key? `"balancer:upstreams:" .. upstream_id`

this change here allows upstreams to be warmed up

hishamhm:
@murillopaula this PR should contain only the changes to `kong/runloop/balancer.lua`, looks like the other three files were committed in this by accident.

(also, a tip: nowadays you can use the Draft option in Github for not-ready PRs instead of the `pr/wip` label that we used before the Draft option existed — this makes it easier to spot that the PR is not ready for review and it grays out the merge button)

murillopaula:
my bad @hishamhm 

I'm closing this one and opening https://github.com/Kong/kong/pull/5681

PR #5665, opened 2020-03-11
fix(key-auth) consumers get cached to core_cache

murillopaula:
### Summary

The plugin was forcing accesses to DB on the first request after restart
even when `db_cache_warmup` was set.

That's because the plugin was using kong.cache to get the consumers
cache key, when consumers get cached to kong.core_cache.

### Full changelog

* [fix(key-auth) consumers get cached to core_cache](https://github.com/Kong/kong/commit/fb25b04fb36af3e61f8a4c57161162b17e161aec)

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -131,11 +131,10 @@ local function do_authentication(conf)
 
   -- retrieve our consumer linked to this API key
 
-  local cache = kong.cache
-
   local credential_cache_key = kong.db.keyauth_credentials:cache_key(key)
-  local credential, err = cache:get(credential_cache_key, nil, load_credential,
-                                    key)
+  local credential, err      = kong.cache:get(credential_cache_key, nil,
+                                              load_credential, key)
+
   if err then
     kong.log.err(err)
     return kong.response.exit(500, {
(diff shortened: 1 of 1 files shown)


bungle:
Just a note here:
this bug is everywhere. 3rd party plugins, our own auth plugin etc.

I think we should fix this in a more generic fashion.

eskerda:
I assumed it was by design and that all plugins should avoid `core_cache` ?

kong/plugins/key-auth/handler.lua
```diff
@@ -132,6 +132,7 @@ local function do_authentication(conf)
   -- retrieve our consumer linked to this API key
 
   local cache = kong.cache
+  local core_cache = kong.core_cache
```
bungle: I don't think we need these two locals anymore, just use `kong.cache` and `kong.core_cache` directly where needed.

kong/plugins/key-auth/handler.lua
```diff
@@ -131,11 +131,10 @@ local function do_authentication(conf)
 
   -- retrieve our consumer linked to this API key
 
-  local cache = kong.cache
-
   local credential_cache_key = kong.db.keyauth_credentials:cache_key(key)
-  local credential, err = cache:get(credential_cache_key, nil, load_credential,
-                                    key)
+  local credential, err      = kong.cache:get(credential_cache_key, nil,
```
murillopaula: can't we use `kong.core_cache` here?

murillopaula: this retrieves the credentials, not the consumer

bungle: Credentials are plugins property. Not a core entity. So let's keep it like it is.

bungle:
> I assumed it was by design and that all plugins should avoid `core_cache` ?

Yes, unless they want to read core entities apparently. We backlogged a card for better solution, but to get fixes quick, we decided to go this path.

kong/plugins/key-auth/handler.lua
thibaultcha: The separation of caches isn't related to core vs plugin entities. It was introduced to ensure that cache keys derived from client-facing properties (e.g. request properties) can't churn out non client-facing cached items (i.e. with keys that aren't derived from client-facing properties).
Now this makes me concerned that over the last few months, we've treated the separated cache incorrectly. Has this been the reasoning behind choosing one or the other recently?

bungle:
Closing this as we merged #5669.

PR #5667, opened 2020-03-11
fix(proxy) prioritize longer regex path matches

chrisaljoudi:
### Summary

Let’s say we have two Routes, each with a single path and otherwise identically-configured:

* `/api/users/(?<userId>\w+)`
* `/api/users/(?<userId>\w+)/profile`

In the current behavior, unless `regex_priority` is set, a request coming in to `/api/users/30/profile` will match the former route (the one _without_ the profile suffix) if it was created first—so the current behavior is to fallback to prioritizing based on created-time.

With non-regex paths, Kong currently (very helpfully) defaults to giving precedence/priority to longer paths. This is a great default.

This change simply makes the same true for regex paths, which is a more reasonable/predictable default (compared to created-time based priority, which is implicit and undesirable).

### Full changelog

* Update `router.lua` to include regex paths in calculating `max_uri_length` for routes

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -450,6 +450,7 @@ local function marshall_route(r)
           route_t.uris[path] = uri_t
           insert(route_t.uris, uri_t)
 
+          route_t.max_uri_length = max(route_t.max_uri_length, #path)
           route_t.submatch_weight = bor(route_t.submatch_weight,
                                         MATCH_SUBRULES.HAS_REGEX_URI)
         end


bungle:
with regex you only know which regex is longer but that doesn't answer which matches the longest uri.

chrisaljoudi:
@bungle Good point. Isn’t that significantly more 'intrusive' to implement, though?

To be clear, I don’t at all disagree that the semantic considering real match lengths is preferable—but it seems like basing this on regex length, while not ideal, is still preferable to the current behavior.

Thoughts?

(Also, just noticed a couple of the tests need to be updated. Will do that if the approach seems reasonable).

bungle:
@hishamhm any opinions about this?

hishamhm:
@chrisaljoudi while I agree that created-time is not a great default criteria (and that's of course why the `regex_priority` field exists to override that), I think applying length  _of the regex_ as criteria could be very misleading to users because it would in many cases _appear_ to filter by match length (especially since non-regex matches already do that), but, as @bungle correctly pointed out, not all.

Running all regexes and applying the longest _match_ would IMO make more sense, I think, and yes, the implementation would look very different, etc. (Though `regex_priority` would still need to apply because you may still want to match shorter regexes). But doing that would be a big behavior compatibility breakage (just like this PR is). It would have to be done in a major version, because for better or worse, sorting by created-at is the currently [documented behavior](https://docs.konghq.com/2.0.x/proxy/#matching-priorities).

chrisaljoudi:
Thanks for the feedback @hishamhm and @bungle. Makes sense that this behavior change should be in a major version, probably. 

With that in mind, what do you guys think if the logic was to default `regex_priority` to the _number of capture groups_ in the regex?

Again, this would still be in a major version, just wondering what you semantically think of that vs. length (trying to support the use-case exemplified above for more ‘specific’ router being matched first).

Number of capture groups is one idea, and another is the number of ‘path segments’ (though that’s less deterministic as you point out, because regex’s may match multiple segments). 

Thoughts?

hishamhm:
@chrisaljoudi Sorry for missing out your last reply!

> With that in mind, what do you guys think if the logic was to default regex_priority to the number of capture groups in the regex?
> 
> Again, this would still be in a major version, just wondering what you semantically think of that vs. length (trying to support the use-case exemplified above for more ‘specific’/drilled-down parameterized routes being matched first).

Yes, semantically it would make more sense in my opinion, but as you said, it would still need to be a Kong 3.0 thing, which is still far away.

Number of path segments, as you also observed, has a similar issue as regex length.

I'll be closing this PR for now since we agreed on not pursuing the particular path this one is taking (no pun intended!) and we'll try to revisit this question when we get towards Kong 3.0!

PR #5668, opened 2020-03-12
chore(deps) bump kong-plugin-zipkin to 1.0.0

kikito:


kong-2.0.2-0.rockspec
```diff
@@ -40,7 +40,7 @@ dependencies = {
   "lua-resty-counter == 0.2.0",
   -- external Kong plugins
   "kong-plugin-azure-functions ~> 0.4",
-  "kong-plugin-zipkin ~> 0.2",
+  "kong-plugin-zipkin ~> 1.0.0",
```
hishamhm: This should be `~> 1.0`, without the extra 0. The point of the `~>` operator is to allow patch version bumps (`x ~> 1.0` means `x >= 1.0.0 && x < 1.1.0`

PR #5669, opened 2020-03-12
fix(cache-warmup) move consumers and tags from core cache to cache

bungle:
### Summary

TLDR; Moves `consumers` and `tags` on cache warmup from `kong.core_cache` to `kong.cache`.

It was found out that our plugins in general tend to call `kong.cache:get`, and they never use `kong.core_cache:get`. This is also for historical reasons as we didn't have `core_cache` before recently. When we did the split we did put so called `core entities` into `core_cache` on cache warmup. This included things like `consumers` and `tags` too. But we didn't change the
plugins. The common case for authentication plugins was to fetch consumers from `kong.cache:get` while the warmup now warmed them up in `kong.core_cache`. Thus the entities weren't warmed up in a right cache from plugin perspective.

In #5665 it was proposed to fix this by changing `key-auth` plugin to use `core_cache`. I personally didn't like this approach as there are a lot of plugins that need to be changed beyond `key-auth`. Also possible 3rd party ones. I checked our core code, and I found practically no use of consumers. Consumers is also something that I don't think we want polluting the core cache.

I know that plugins is also connected to consumers, so that can make plugins entity a bit similar to consumers in a size in a worst case. Plugin iterator is on the other hand on our hot path and it was already using `core_cache`.

### Issues Resolved

Replaces #5665

### DIFF
--- a/kong/cache_warmup.lua
+++ b/kong/cache_warmup.lua
@@ -40,9 +40,10 @@ end
 local function cache_warmup_single_entity(dao)
   local entity_name = dao.schema.name
 
-  local cache = constants.CORE_ENTITIES[entity_name] and kong.core_cache or kong.cache
+  local cache_store = constants.ENTITY_CACHE_STORE[entity_name]
+  local cache = kong[cache_store]
 
-  ngx.log(ngx.NOTICE, "Preloading '", entity_name, "' into the cache ...")
+  ngx.log(ngx.NOTICE, "Preloading '", entity_name, "' into the ", cache_store, "...")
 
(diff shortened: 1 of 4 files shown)


murillopaula:
Tested locally and it works!

bungle:
> LGTM, this is a safer approach than changing the plugin to use `core_cache`. Question: why do tags also need to be in `cache`?

Tags are not really cached, but if someone wants to do that, they will be warned up to `cache` instead of `core_cache`, I guess mostly because they potentially may have high cardinality.

PR #5671, opened 2020-03-12
Pass copied header tables to exit() in error handler

rainest:
### Summary
Copy error response header tables before passing them to `kong.response.exit()`. The exit-transformer plugin can modify these tables, and the current pass by reference allows those modifications to persist for other requests on the same worker.

### Full changelog

* Copy response header tables before passing them to `kong.response.exit()` in error handler.

### Issues resolved

Fix an issue where modifying the `headers` variable in exit-transformer functions resulted in unrelated requests seeing the same header modification.

### DIFF
--- a/kong/error_handlers.lua
+++ b/kong/error_handlers.lua
@@ -1,6 +1,7 @@
 local kong = kong
 local find = string.find
 local fmt  = string.format
+local tablex = require "pl.tablex"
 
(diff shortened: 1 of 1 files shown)


rainest:
Hisham had a better suggestion. Closing this and opening a new PR with that.

PR #5681, opened 2020-03-13
fix(balancer) allow upstreams to be warmed up

murillopaula:
### Summary

Use a different cache key for upstream entities. This allows individual
upstream entities to be warmed up in the cache when
`db_cache_warmup_entities = upstreams

### Full changelog

* [fix(balancer) allow upstreams to be warmed up](https://github.com/Kong/kong/commit/62e84d69f890d9cae106593b8cae224294ba964f)

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -47,19 +47,14 @@ local healthcheck_subscribers = {}
 
 -- Caching logic
 --
--- We retain 3 entities in singletons.cache:
+-- We retain 2 tables in singletons.core_cache:
 --
 -- 1) `"balancer:upstreams"` - a list of upstreams
 --    to be invalidated on any upstream change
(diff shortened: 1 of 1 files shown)


Tieske:
just running into this; imo it is useless to warm-up cache for upstreams. This only makes sense for entities that get loaded "on-demand". The upstreams are not in that category, upstreams are always immediately loaded upon system start, because they need to initiate the healthchecks.

Hence when the system starts, they will be loaded and cache will already be warmed up. So any additional warmup actions are redundant imo.

Am I missing something @hishamhm @locao ?

locao:
@murillopaula can we close this as #5745 is already merged?

murillopaula:
@locao yes, I'd vote for closing this one as it seems like warming up upstreams entities are not worth it.

PR #5692, opened 2020-03-17
fix(proxy) the plugins might not be fully collected on error on one plugin

bungle:
### Summary

The plugin iterator builds the list of plugins. If the first plugin errors on plugins iterator build on access phase there is a chance that the plugins iterator may not be built completely. E.g. it might be missing all the log plugins. This PR delays the response so that the plugins iterator will still be fully built.

It basically also allows plugins to do:

```lua
if err then
  return error(err)
end
```

instead of:

```lua
if err then
  kong.log.err(err)
end
```

Or that those two are the `same`.

In separate commit I also changed the plugins to use `error(err)`, but I decided to move that to another PR: #5696.

@Tieske, isn't this what you suggested at some point?

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -681,20 +681,15 @@ function Kong.access()
       kong_global.set_namespaced_log(kong, plugin.name)
 
       local err = coroutine.wrap(plugin.handler.access)(plugin.handler, plugin_conf)
+      if err then
+        kong.log.err(err)
+        ctx.delayed_response = {
+          status_code = 500,
+          content     = { message  = "An unexpected error occurred" },
+        }
+      end
 
       kong_global.reset_log(kong)
-
-      if err then
-        ctx.delay_response = false
-
-        kong.log.err(err)
-
-        ctx.KONG_ACCESS_ENDED_AT = get_now_ms()
-        ctx.KONG_ACCESS_TIME = ctx.KONG_ACCESS_ENDED_AT - ctx.KONG_ACCESS_START
-        ctx.KONG_RESPONSE_LATENCY = ctx.KONG_ACCESS_ENDED_AT - ctx.KONG_PROCESSING_START
-
-        return kong.response.exit(500, { message  = "An unexpected error occurred" })
-      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


bungle:
I am thinking about adding some tests, to prove it.

murillopaula:
I'd love to port this to https://github.com/Kong/kong/pull/5640

murillopaula:
@bungle let's make sure to update the ip restriction plugin  https://github.com/Kong/kong/pull/5640

bungle:
> @bungle let's make sure to update the ip restriction plugin #5640

That is now on this sister pr: #5696

bungle:
> I am thinking about adding some tests, to prove it.

Ok I added a test that proves I was right about it.

bungle:
@hishamhm,

I am not sure if this should be a `feat` or `fix` as it depends a bit. Also even if this is a `fix` should we instead put it into `next`.

There is a slight change that some plugins that run on e.g. `header_filter` expect that the plugin also ran its `access` (it might not be always a case if e.g. other plugin errors before this plugin). This is of course a wrong assumption, but a bug that is easy to make. But this problem is already when other plugins short-circult, that the plugin needs to detect it too. So basically this just makes error on plugins on `access` behave the same as plugins that short-circuit with `kong.response.exit`, making `error`/`assert`/`runtime errors` do basically a short-circuiting on `access` phase.

hishamhm:
@bungle I think the use-case described in the PR description and demonstrated in the test is pretty compelling to treat this as a `fix`: when a log plugin is enabled, no request should ever go un-logged.

> some plugins that run on e.g. header_filter expect that the plugin also ran its access (it might not be always a case if e.g. other plugin errors or short-circuits before this plugin). This is of course a wrong assumption, but a bug that is easy to make. But this problem is already when other plugins short-circult

This is a good point and something to keep on the back of our minds when we get around to make any redesign of the plugin runloop, but as you say, it's an issue that is already present. So I think this should go on `master`, let's merge it.

As for the refactor PR #5696 that uses this "as a feature", I agree that it is better if that one goes on `next` to avoid new code that kind of "depends of patch release behavior", so let's merge in the sequence you described there.

PR #5696, opened 2020-03-18
refactor(plugins) use error(err) instead of kong.response.exit on 500 errors

bungle:
### Summary

Makes plugins to do:

```
local ok, err = func()
if not ok then
  return error(err)
end
```

instead of:

```
local ok, err = func()
if not ok then
  kong.log.err(err)
end
```

This PR is a continuation of #5692 (this is applied to `next` so that when #5692 is merged to master, that needs to then be merged to next, so please not merge this until then).

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -9,6 +9,7 @@ local max = math.max
 local time = ngx.time
 local floor = math.floor
 local pairs = pairs
+local error = error
 local tostring = tostring
 local timer_at = ngx.timer.at
 
@@ -124,12 +125,11 @@ function RateLimitingHandler:access(conf)
 
   local usage, stop, err = get_usage(conf, identifier, current_timestamp, limits)
   if err then
-    if fault_tolerant then
-      kong.log.err("failed to get usage: ", tostring(err))
-    else
-      kong.log.err(err)
-      return kong.response.exit(500, { message = "An unexpected error occurred" })
+    if not fault_tolerant then
+      return error(err)
     end
+
+    kong.log.err("failed to get usage: ", tostring(err))
   end
 
   if usage then
--- a/kong/plugins/response-ratelimiting/access.lua
+++ b/kong/plugins/response-ratelimiting/access.lua
@@ -5,11 +5,11 @@ local timestamp = require "kong.tools.timestamp"
 local kong = kong
 local next = next
 local pairs = pairs
+local error = error
 local tostring = tostring
 
 
 local EMPTY = {}
-local HTTP_INTERNAL_SERVER_ERROR = 500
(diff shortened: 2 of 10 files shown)


bungle:
> much better! error expression (rendering?) should be an environment thing, not done at every failure point.
> 
> tangentially related: `if err then ....` is an antipattern in Lua. much more maintainable is
> 
> ```
>    local val, err = ...
>    if not val then handle_error(err) end
> ```

I agree, just didn't have courage to check all the cases.

kong/plugins/response-ratelimiting/access.lua
```diff
@@ -75,14 +75,12 @@ function _M.execute(conf)
...
   local usage, err = get_usage(conf, identifier, conf.limits, current_timestamp)
   if err then
-    if conf.fault_tolerant then
-      kong.log.err("failed to get usage: ", tostring(err))
-      return
-    else
-      return kong.response.exit(HTTP_INTERNAL_SERVER_ERROR, {
-        message = "An unexpected error occurred"
-      })
+    if not conf.fault_tolerant then
```
kikito: Why was this `if` changed to a negative? In my opinion it is more difficult to read now.

kong/plugins/rate-limiting/handler.lua
```diff
@@ -124,12 +125,11 @@ function RateLimitingHandler:access(conf)
 
   local usage, stop, err = get_usage(conf, identifier, current_timestamp, limits)
   if err then
-    if fault_tolerant then
-      kong.log.err("failed to get usage: ", tostring(err))
-    else
-      kong.log.err(err)
-      return kong.response.exit(500, { message = "An unexpected error occurred" })
+    if not fault_tolerant then
```
kikito: Why `if not` instead of `if` ?

kong/plugins/response-ratelimiting/access.lua
javierguerragiraldez: in general, i also prefer to avoid testing the negative, but (at least for me), dealing first with error cases trumps that.  especially if that allows to avoid the `else`

kong/plugins/rate-limiting/handler.lua
bungle: It follows out style of `return early`, aka makes it return early. But I can change it back.

bungle:
> I agree, just didn't have courage to check all the cases.

And it seems like our DAO can easily `return nil, nil` when no entity found. It is not considered `err` on that API, just that the DB query did not return anything. Perhaps that API should have been designed differently.

kong/plugins/response-ratelimiting/access.lua
eskerda: early return is the best of all returns

PR #5712, opened 2020-03-22
chore(tests) some test related updates/cleanup

Tieske:
some test related updates/cleanup

### DIFF
--- a/spec/02-integration/05-proxy/10-balancer/01-ring-balancer_spec.lua
+++ b/spec/02-integration/05-proxy/10-balancer/01-ring-balancer_spec.lua
@@ -1330,11 +1330,6 @@ for _, strategy in helpers.each_strategy() do
           end)
 
           it("threshold for health checks", function()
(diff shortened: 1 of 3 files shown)


kong/runloop/balancer.lua
```diff
@@ -1091,20 +1091,22 @@ local function get_balancer_health(upstream_id)
...
 
-    local balancer_status = balancer:getStatus()
+    balancer_status = balancer:getStatus()
     health = balancer_status.healthy and "HEALTHY" or "UNHEALTHY"
   end
 
   return {
     health = health,
     id = upstream_id,
+    details = balancer_status,
```
hishamhm: On which endpoints would this extra info appear? (A test demonstrating it would be nice. :) )

Tieske:
@hishamhm I added some additional validations from the data added to the tests. The endpoint is the balancer health endpoint, with the query parameter (as shown in the test);

    /upstreams/{name or id}/health/?balancer_health=1

PR #5713, opened 2020-03-23
[RFC] feat(JWT plugin) Pass through extra JWTs

nhynes:
### Summary

This (RFC) PR introduces the ability for the JWT plugin to ignore JWTs that aren't addressed to a service directly behind Kong (via the `sub` claim). The objective is to allow upstream services to process their own JWTs. For instance, if a proxied service requires a user OAuth token, the requester could append it to the `Authorization` header; currently this would trip up the regex in `request_token`.

This PR currently contains only a draft implementation of the proposal in hopes of ascertaining whether this is something that the community would accept (and thus prevent wasted work adding tests+docs).

### Full changelog

* Untested, undocumented implementation of proposal

### DIFF
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -24,11 +24,13 @@ JwtHandler.VERSION = "2.1.0"
 -- @param conf Plugin configuration
 -- @return token JWT token contained in request (can be a table) or nil
 -- @return err
-local function retrieve_token(conf)
+local function retrieve_tokens(conf)
+  local tokens = {}
+
   local args = kong.request.get_query()
   for _, v in ipairs(conf.uri_param_names) do
     if args[v] then
-      return args[v]
+      tokens[#tokens+1] = args[v]
     end
   end
 
@@ -36,7 +38,7 @@ local function retrieve_token(conf)
   for _, v in ipairs(conf.cookie_names) do
     local cookie = var["cookie_" .. v]
     if cookie and cookie ~= "" then
-      return cookie
+      tokens[#tokens+1] = cookie
     end
   end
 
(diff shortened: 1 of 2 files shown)


kong/plugins/jwt/handler.lua
```diff
@@ -47,23 +48,24 @@ local function retrieve_token(conf)
       if type(token_header) == "table" then
         token_header = token_header[1]
       end
-      local iterator, iter_err = re_gmatch(token_header, "\\s*[Bb]earer\\s+(.+)")
+      local iterator, iter_err = re_gmatch(token_header, ",?\\s*[Bb]earer\\s+([\\w\\d+.=-]+)\\s*,?")
```
mitar, line 51: I do not think this is necessary to have multiple tokens in one `Authorization` header. You can have multiple `Authorization` headers. So plugin here should pass all of them through.

kong/plugins/jwt/handler.lua
```diff
@@ -24,19 +23,21 @@ JwtHandler.VERSION = "2.1.0"
 -- @param conf Plugin configuration
 -- @return token JWT token contained in request (can be a table) or nil
 -- @return err
-local function retrieve_token(conf)
+local function retrieve_tokens(conf)
```
mitar, line 26: Probably you want for backwards compatibility to keep `retrieve_token` around, which would call `retrieve_tokens` and return the first one, or something.

kong/plugins/jwt/handler.lua
nhynes: That’s actually not true

>    A sender MUST NOT generate multiple header fields with the same field
   name in a message unless either the entire field value for that
   header field is defined as a comma-separated list [i.e., #(values)]
   or the header field is a well-known exception (as noted below).

Set-Cookie is the exception. Auth is csv. 

[reference](https://tools.ietf.org/html/rfc7230#section-3.2.2)

PR #5720, opened 2020-03-24
chore(oauth2) small cleanups on a merged pkce pr #5268

bungle:
### Summary

This PR just makes some style and refactoring changes on the recently merged pkce pr #5268 by @amberheilman.

Main changes:

- `one_of = { "S256" }` constraint added to `oauth2_authorization_codes` dao (yes, `one_of` because in future we may add more if needed)
- refactored base64 url encoding code to functions
- changed `retrieve_code_challenge` to `return nil, err` on error and `return challenge, nil, challenge_method` on ok
- removed one log statement as it could potentially leak information to logs (just in case)
- more robust handling of challenge (tries to base64 url decode and then if success base64 url encode it back (replacing the `remove padding code`)


spec/03-plugins/25-oauth2/03-access_spec.lua
```diff
@@ -2230,7 +2230,7 @@ describe("Plugin: oauth2 [#" .. strategy .. "]", function()
         local body = assert.res_status(200, res)
         assert.is_table(ngx.re.match(body, [[^\{"refresh_token":"[\w]{32,32}","token_type":"bearer","access_token":"[\w]{32,32}","expires_in":5\}$]]))
       end)
-      it("fails when code challenge contains + or / characters", function()
+      it("success when code challenge contains + or / characters", function()
```
kikito:
```suggestion
      it("succeeds when code challenge contains + or / characters", function()
```

PR #5725, opened 2020-03-25
Remove fixture for Lambda + bump version

Tieske:
NOTE: this can only be merged after updating the Lambda plugin to a newer version than 3.2.0.


Tieske:
this seems to hang on a caching issue, it keeps reusing the old lambda plugin, not the newer 3.2.1

PR #5745, opened 2020-03-31
perf(balancer) load upstreams before cleaning cache

locao:
When there are a large number of upstreams in the database or the database is under heavy load, loading upstreams will take longer and, sometimes, unacceptably longer. This change addresses said issue by keeping the stale cache until the new data is loaded from the database. As there might be stale content with this change, this behavior only exists when worker consistency is set to eventual.

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -839,13 +839,18 @@ local function update_balancer_state(premature)
 
   concurrency.with_coroutine_mutex(opts, function()
     if is_worker_state_stale() then
-      singletons.core_cache:invalidate_local("balancer:upstreams")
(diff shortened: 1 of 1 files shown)


kong/runloop/balancer.lua
```diff
@@ -839,9 +839,10 @@ local function update_balancer_state(premature)
 
   concurrency.with_coroutine_mutex(opts, function()
     if is_worker_state_stale() then
+      local updated_upstreams_dict = load_upstreams_dict_into_memory()
```
bungle, line 842: Should there now be error handling?

locao: Good catch, thanks! Fixed it

PR #5759, opened 2020-04-07
fix(declarative) fix memory leaks when loading a declarative config that

### LINKED ISSUE #5732: Worker memory steadily increases when Ingress controller is trying to push a bad config
### Summary

I have Kong 2.0.2 (also seen with 1.5.1) deployed with the ingress controller 0.7.1. In a case where the ingress controller is trying to submit a "bad" config, the nginx worker memory usage steadily increases until hitting the cgroup memory limit and getting terminated by the OOM killer.

It doesn't seem to particularly matter how you break the config, as long as the ingress controller is trying over and over to sync the config. The moment that the config is fixed and the ingress controller stops trying to continuously sync the config, the memory usage levels off.

Both an OOM event and the moment that I fixed the config are pretty clear in this graph:

Screen Shot 2020-03-26 at 3 30 37 PM

### Steps To Reproduce

1. Deploy kong with ingress controller
2. Create a "bad" config
3. Watch memory usage climb

### Additional Details & Logs

Workers getting killed:

```
2020/03/26 20:04:16 [notice] 1#0: start worker process 28
2020/03/26 20:04:16 [notice] 1#0: start worker process 29
...
2020/03/26 20:12:22 [notice] 1#0: signal 17 (SIGCHLD) received from 29
2020/03/26 20:12:22 [alert] 1#0: worker process 29 exited on signal 9
2020/03/26 20:12:22 [notice] 1#0: start worker process 30
2020/03/26 20:12:22 [notice] 1#0: signal 29 (SIGIO) received
2020/03/26 20:17:11 [notice] 1#0: signal 17 (SIGCHLD) received from 28
2020/03/26 20:17:11 [alert] 1#0: worker process 28 exited on signal 9
2020/03/26 20:17:11 [notice] 1#0: start worker process 31
2020/03/26 20:17:11 [notice] 1#0: signal 29 (SIGIO) received
2020/03/26 20:22:45 [notice] 1#0: signal 17 (SIGCHLD) received from 30
2020/03/26 20:22:45 [alert] 1#0: worker process 30 exited on signal 9
2020/03/26 20:22:45 [notice] 1#0: start worker process 32
2020/03/26 20:22:45 [notice] 1#0: signal 29 (SIGIO) received
```

Example bad config error from ingress controller:

```
E0326 19:27:56.322072       1 controller.go:119] unexpected failure updating Kong configuration:
posting new config to /config: 400 Bad Request {"fields":{"consumers":[{"basicauth_credentials":[{"@entity":["all or none of these fields must be set: 'password', 'consumer.id'"]}]},{"basicauth_credentials":[{"@entity":["all or none of these fields must be set: 'password', 'consumer.id'"]}]},{"basicauth_credentials":[{"@entity":["all or none of these fields must be set: 'password', 'consumer.id'"]}]}]},"name":"invalid declarative configuration","code":14,"message":"declarative config is invalid: {consumers={{basicauth_credentials={{[\"@entity\"]={\"all or none of these fields must be set: 'password', 'consumer.id'\"}}}},{basicauth_credentials={{[\"@entity\"]={\"all or none of these fields must be set: 'password', 'consumer.id'\"}}}},{basicauth_credentials={{[\"@entity\"]={\"all or none of these fields must be set: 'password', 'consumer.id'\"}}}}}}"}
```

dndx:
fails schema validation

When a declarative config fails schema validation, we attempts to
re-load the config again after populating the foreign keys
(see `kong.db.schema.others.declarative_config.flatten`).

Inside `flatten` we call the `kong.db.schema.others.declarative_config.load`
function, which reloads the schema objects and insert them into
the `all_schemas` global table without first clearing it.
The result is that the `all_schemas` table will keep growing
indefinitely and eventually causes Nginx to be OOM killed.

This PR attempts to fix the issue by clearing the `all_schemas` table
every time `load` is called, which according to my test indeed prevents
the memory from leaking anymore.

closes #5732

### DIFF
--- a/kong/db/schema/others/declarative_config.lua
+++ b/kong/db/schema/others/declarative_config.lua
@@ -632,16 +632,14 @@ function DeclarativeConfig.load(plugin_set, include_foreign)
 
   local known_entities = utils.deep_copy(core_entities, false)
 
-  if not all_schemas then
-    all_schemas = {}
-    for _, entity in ipairs(core_entities) do
-      local mod = require("kong.db.schema.entities." .. entity)
-      local definition = utils.deep_copy(mod, false)
-      all_schemas[entity] = Entity.new(definition)
+  all_schemas = {}
+  for _, entity in ipairs(core_entities) do
+    local mod = require("kong.db.schema.entities." .. entity)
+    local definition = utils.deep_copy(mod, false)
+    all_schemas[entity] = Entity.new(definition)
 
-      -- load core entities subschemas
-      assert(load_entity_subschemas(entity, all_schemas[entity]))
-    end
+    -- load core entities subschemas
+    assert(load_entity_subschemas(entity, all_schemas[entity]))
   end
 
   for plugin in pairs(plugin_set) do


hbagdi:
This should have been targeted for `master`.
Can we cherry pick to master to include this in 2.0.4?

PR #5764, opened 2020-04-08
fix(ca_certificates) add digest field for `ca_certificates` table

artomal:
### Summary

Fix issue for `ca_certificates` table index which occurs when certificate
is uploaded which index length is more than `2704`, index `btree` supports
only up to `2704` char length.

Solution: Introduce a new field `cert_digest` and store SHA256(cert) which
should meet `btree` limitations when records are indexed by this field.

### Full changelog

- add additional field `cert_digest` to `ca_certificates` schema
- modify tests


hbagdi:
It would help us to hide the digest field from the end-user API.
@bungle How hard is it to have a field defined in the schema but hide it from the API?

@artomal Can you add a check in the code (and a test-case) to reject any request that specify the digest value as part of the request?

dndx:
@hbagdi Personally I don't think echoing the digest inside the Admin API is a huge issue, it is very customary to use digest as a way of uniquely identifying a certificate. e.g. when viewing certificate details from browsers, certificate digests are usually shown.

That we want to avoid is using attempting to change the content of the field through APIs. As long as we produce a validation error when the field is included inside the request payload it should not hurt to return the value inside the response.

artomal:
I have another PR with a feature proposal to hide selected fields from API consumers. 

https://github.com/Kong/kong/pull/5770

dndx:
Ping @hishamhm.

kong/db/schema/entities/ca_certificates.lua
```diff
@@ -8,10 +9,22 @@ return {
...
+  check = function(entity)
+    local digest = str.to_hex(openssl_x509.new(entity.cert):digest("sha256"))
+    if not digest then
+      return nil, "cannot create digest value of certificate"
+    end
+
+    entity.cert_digest = digest
+
+    return true
+  end,
```
kikito, line 26: I think the accepted way to do this kind of thing nowadays is via transformations. This might require changing the tests below:

```suggestion
  transformations = {
    {
      input = { "cert" },
      on_write = function(cert)
        local digest = str.to_hex(openssl_x509.new(cert):digest("sha256"))
        if not digest then
          return nil, "cannot create digest value of certificate"
        end
        return { digest = digest }
      end,
    },
  },
```

bungle: Yes, if it is not with transformations it may cause issues with dbless.

artomal: thanks @kikito and @bungle , working on it!

Tieske:
Is this defined properly for PATCH and PUT behaviour?

What if I PATCH the calculated `cert_digest` field without touching the source `cert` field?

dndx:
@artomal @kikito I still insist we should at least error out when user attempts to provide the `cert_digest` field inside Admin API. Silently overriding it makes the field's meaning ambiguous.

Also in the Admin API doc we should emphasize that the field is readonly.

hutchic:
Restoring this branch because I need it for something. Appologies / ignore the noise ;)

PR #5765, opened 2020-04-08
feat(config) add "kic" configuration for reporting

rainest:
### Summary

Add a `kic` configuration setting for reporting purposes. This is not documented in kong.conf.default, as we don't intend for users to set it manually. It should only be set by our Kubernetes tooling.

### Full changelog

* Add `kic` configuration, default `off`.
* Add `kic` to reporting data.
* Add tests for reporting.

### DIFF
--- a/spec/01-unit/11-reports_spec.lua
+++ b/spec/01-unit/11-reports_spec.lua
@@ -170,6 +170,32 @@ describe("reports", function()
       end)
     end)
 
+    describe("sends 'kic'", function()
+      it("default (off)", function()
+        local conf = assert(conf_loader(nil))
+        reports.configure_ping(conf)
+
+        local thread = helpers.tcp_server(8189)
+        reports.send_ping("127.0.0.1", 8189)
+
+        local _, res = assert(thread:join())
+        assert._matches("kic=false", res, nil, true)
+      end)
+
+      it("enabled", function()
+        local conf = assert(conf_loader(nil, {
+          kic = "on",
+        }))
+        reports.configure_ping(conf)
+
+        local thread = helpers.tcp_server(8189)
+        reports.send_ping("127.0.0.1", 8189)
+
+        local _, res = assert(thread:join())
+        assert.matches("kic=true", res, nil, true)
+      end)
+    end)
+
     describe("sends '_admin' for 'admin_listen'", function()
       it("off", function()
         local conf = assert(conf_loader(nil, {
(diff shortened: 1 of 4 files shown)


spec/01-unit/11-reports_spec.lua
```diff
@@ -170,6 +170,32 @@ describe("reports", function()
...
+    describe("sends 'kic'", function()
+      it("default (off)", function()
+        local conf = assert(conf_loader(nil))
+        reports.configure_ping(conf)
+
+        local thread = helpers.tcp_server(8189)
+        reports.send_ping("127.0.0.1", 8189)
+
+        local _, res = assert(thread:join())
+        assert._matches("kic=false", res, nil, true)
```
hbagdi, line 182: Question:

We don't have `true/false` in anonymous reports anywhere else. We do have a `off` for DB at least.
Should we instead switch to `off` or `on` instead?

This might be bike-shedding on a very small detail.

hbagdi:
Question for core team members,
Since this is not a new user-facing feature, can we please merge this into master so that this is included in 2.0.4?

spec/01-unit/11-reports_spec.lua
rainest: I think all the rest are strings, so `off` is actually `off` (versus say, `postgres` for `database`). Booleans convert into actual booleans for everything past the loader. I originally had the tests set that way, but they fail if so:

```
Failure → spec/01-unit/11-reports_spec.lua @ 174
reports configure_ping() sends 'kic' default (off)
spec/01-unit/11-reports_spec.lua:182: Expected strings to match.
Passed in:
(string) '<14>version=2.0.2;uname=Linux x86_64;hostname=bd23d4b7c2b3;cores=4;database=postgres;role=traditional;kic=false;_admin=1;_proxy=1;_stream=0;http_reqs=0;signal=ping;grpcs_reqs=0;requests=0;wss_reqs=0;go_plugin_reqs=0;https_reqs=0;unique_id=WfWrjnZKrrAZSsCdLhvNdiZg6yFFHaUR;h2c_reqs=0;ws_reqs=0;h2_reqs=0;grpc_reqs=0'
Expected:
(string) 'kic=off'
```

hbagdi: I should have been clearer. The current implementation and tests are correct.

What I'm wondering is if `true` should translate to `on` and `false` to `off` in the report packet.

rainest: Unless there's some reason reporting can't handle the boolean values I think it's fine to leave them as-is. We don't get any richer data in the end.

hbagdi:
@rainest Can you change the patch and PR to target master instead?

PR #5770, opened 2020-04-09
feat(schema) add hidden field feature to hide selected schema fields

artomal:
### Summary

Add feature that enables engineers to hide chosen fields in schema from
the clients that use API. There are a couple of reasons to have this feature, one of
the examples would be to hide "operational" system fields from the end
client like calculating digest values of certain fields that is only used internally.

### Full changelog

* Add new metaschema attribute `hidden` boolean type
* Add related tests

### DIFF
--- a/spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
+++ b/spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
@@ -1067,6 +1067,40 @@ describe("metasubschema", function()
     assert.match("'set' cannot have attribute 'unique'", err.set)
   end)
 
+  it("supports the hidden attribute with a string type", function()
+    local s = {
+      name = "test",
+      fields = {
+        { str = { type = "string", hidden = true } },
+      },
+    }
+    assert.truthy(MetaSchema.MetaSubSchema:validate(s))
+  end)
+
+  it("rejects a hidden attribute in complex and number types", function()
+    local s = {
+      name = "test",
+      fields = {
+        { id  = { type = "string" } },
+        { num = { type = "number", hidden = true } },
+        { int = { type = "integer", hidden = true } },
+        { arr = { type = "array", hidden = true } },
+        { map = { type = "map", hidden = true } },
+        { rec = { type = "record", hidden = true } },
+        { set = { type = "set", hidden = true } },
+      },
+    }
+    local ok, err = MetaSchema.MetaSubSchema:validate(s)
+    assert.falsy(ok)
+    print(require("inspect")(err))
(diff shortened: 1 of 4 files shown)


Tieske:
Is this needed on the database level? If the values are calculated, can't we just calculate them once, when loading from the db?

Like a de-serialization callback in ml-cache for example. See the option `l1_serializer` here: https://github.com/thibaultcha/lua-resty-mlcache/blob/master/README.md#new

spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
```diff
@@ -1067,6 +1067,40 @@ describe("metasubschema", function()
...
+        { int = { type = "integer", hidden = true } },
+        { arr = { type = "array", hidden = true } },
+        { map = { type = "map", hidden = true } },
+        { rec = { type = "record", hidden = true } },
+        { set = { type = "set", hidden = true } },
+      },
+    }
+    local ok, err = MetaSchema.MetaSubSchema:validate(s)
+    assert.falsy(ok)
+    print(require("inspect")(err))
```
kikito, line 1095: Left debug statement here by mistake

spec/01-unit/01-db/01-schema/02-metaschema_spec.lua
```diff
@@ -1067,6 +1067,40 @@ describe("metasubschema", function()
...
+    assert.truthy(MetaSchema.MetaSubSchema:validate(s))
+  end)
+
+  it("rejects a hidden attribute in complex and number types", function()
+    local s = {
+      name = "test",
+      fields = {
+        { id  = { type = "string" } },
+        { num = { type = "number", hidden = true } },
+        { int = { type = "integer", hidden = true } },
```
kikito, line 1086: I think other types (integer, boolean) might be useful as hidden fields.

spec/01-unit/01-db/04-dao_spec.lua
```diff
@@ -160,6 +176,26 @@ describe("DAO", function()
...
+          return { a = 42, b = null, u = "data", r = { f1 = 10, f2 = null } }
+        end,
+      }
+
+      local dao = DAO.new(mock_db, schema, strategy, errors)
+
+      local row = dao:select({ a = 42 }, { nulls = true })
+      assert.same(42, row.a)
+      assert.same(null, row.b)
+      assert.same(nil, row.u)
```
kikito, line 195: If the DAO itself does not return hidden fields, how does one actually *use* those fields?

I expected this feature to act on the Admin Api level, not on this one.

PR #5773, opened 2020-04-10
fix(balancer) set a proper log prefix with upstream name

Tieske:
instead of the balancers creating logs like `[balancer 1]`, a balancer create from upstream `myupstream` will now log with a prefix `[upstream:myupstream 1]` making correlating the logs a lot easier.

The default for `log_prefix` is `"balancer"` but was never set/overridden by Kong.

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -291,7 +291,7 @@ do
         end
 
         if not ok then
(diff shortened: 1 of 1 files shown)


Tieske:
@bungle can you please merge it? I have no powers in this universe

PR #5775, opened 2020-04-11
fix(admin-api) properly invalidate target history at cleanup

### LINKED ISSUE #5774: [BUG] balancers go out of sync, due to missing invalidation
I think the original bug was #5593 (and in [Kong Nation](https://discuss.konghq.com/t/deleting-targets/5595)), but the solution in #5710 is not the proper one.

The balancers keep a history of all target changes, such that each Kong node can rebuild balancers in the exact same state (this is required for consistent hashing to work clusterwide). To prevent uncontrolled growth of the target history, there is an auto-clean mechanism that will cleanup history (and hence break consistency) when the ratio of outdated-to-valid records hits more than 90% outdated.

Now when this happens the balancers need to be invalidated, such that they rebuild and the whole cluster is back to the same state as it is in the DB. The way this is done (since a lot of deletions are done from the DB), is by NOT sending events for each deleted record, but instead, send 1 invalidation event after the cleanup is complete. See [this code comment](https://github.com/Kong/kong/blob/97d548dbd9dddf8bf29e6e9a6d69777491d5b85c/kong/db/dao/targets.lua#L71-L73).
Essentially the cleanup is done without sending any events at all, and we're relying on the new row to be inserted to trigger the event afterwards.

The problem now is that at some point the code got reshuffled, and we changed the order in which we insert and cleanup. See [code here](https://github.com/Kong/kong/blob/97d548dbd9dddf8bf29e6e9a6d69777491d5b85c/kong/db/dao/targets.lua#L106-L109). It is clear that we first insert the new record (and the invalidation event!), and only then run the cleanup. Hence the changes by the cleanup will not be followed by the required invalidation event.

Usually this just works, in cases with little or no latency the window between the invalidation event, and the cleanup being finished, is small enough to simply work in most cases. Yet if there is latency (for example to the DB, since we delete a lot (90%) of [rows individually](https://github.com/Kong/kong/blob/97d548dbd9dddf8bf29e6e9a6d69777491d5b85c/kong/db/dao/targets.lua#L70-L78)) then there is a real chance of another node picking up the invalidation event while the cleanup is only half done, and hence getting a balancer state that is incorrect.

The related PR's and discussion (see links at the top) are displaying the bad behaviour. When the new target is inserted only after the cleanup occured, then the target history can only be empty when the balancer is newly created. At any later stage, when a cleanup is done, it is always followed by the insertion of the last change (triggering the invalidation), so even if the weight is set to 0, there would after a cleanup be a single target entry with `weight=0`. It should never be empty.

Tieske:
fixes #5774

### DIFF
--- a/kong/db/dao/targets.lua
+++ b/kong/db/dao/targets.lua
@@ -69,8 +69,6 @@ local function clean_history(self, upstream_pk)
     local cnt = 0
     for _, entry in ipairs(delete) do
       -- notice super - this is real delete (not creating a new entity with weight = 0)
-      -- not sending update events, one event at the end, based on the
-      -- post of the new entry should suffice to reload only once
       self.super.delete(self, { id = entry.id })
       -- ignoring errors here, deleted by id, so should not matter
       -- in case another kong-node does the same cleanup simultaneously
@@ -103,10 +101,11 @@ function _TARGETS:insert(entity)
     entity.target = formatted_target
   end
 
+  -- cleaning up will NOT send invalidation events, hence we only add the new
+  -- entry AFTER the cleanup, such that the cleanup will be picked up by the
+  -- other nodes based on the event of the newly added entry
+  clean_history(self, entity.upstream)
   local row, err, err_t = self.super.insert(self, entity)
-  if row then
-    clean_history(self, entity.upstream)
-  end
 
   return row, err, err_t
 end
--- a/kong/db/dao/targets.lua
+++ b/kong/db/dao/targets.lua
@@ -53,9 +53,7 @@ local function clean_history(self, upstream_pk)
     else
       -- haven't got this one, so this is the current state for this target
       seen[entry.target] = true
-      -- we can delete this entry if its weight == 0, but let's keep it if the
-      -- target list is almost empty, so the upstream still exists after cleaning up
-      if entry.weight == 0 and #targets > 2 then
+      if entry.weight == 0 then
         delete[#delete+1] = entry
       end
     end


bungle:
@locao & @Tieske, just to make sure there is no conflict with this?
https://github.com/Kong/kong/pull/5745

Tieske:
@bungle the problem is that I wouldn't know how to test it. The behaviour we've seen is only with very high DB latency (still waiting for verification whether this actually fixes it). We have no way to simulate that. But suggestions are welcome of course.

Tieske:
@locao PR #5745 doesn't seem related/impacted to me. What do you think?

kong/db/dao/targets.lua
```diff
@@ -53,9 +53,7 @@ local function clean_history(self, upstream_pk)
     else
       -- haven't got this one, so this is the current state for this target
       seen[entry.target] = true
-      -- we can delete this entry if its weight == 0, but let's keep it if the
-      -- target list is almost empty, so the upstream still exists after cleaning up
-      if entry.weight == 0 and #targets > 2 then
+      if entry.weight == 0 then
```
Tieske, line 56: @locao instead of the old check, we could also do a check just a bit up, to set a minimum number of entries before ever starting a clean-up. Technically not needed, but would ease the "aggressiveness" earlier noted by @bungle Eg:

```lua
  if #target < 10 then
    -- too little entries, skip cleanup all together for efficiency reasons
    return
  end
```

locao:
> PR #5745 doesn't seem related/impacted to me. What do you think?

You are right, it's not impacted. That change is only related to the cache entries.

PR #5786, opened 2020-04-15
fix(declarative) send config updates to stream subsystem via Unix domain

### LINKED ISSUE #5656: stream routing does not work in DB-less and hybrid mode
### Summary

Stream routes are not propagated in DB-less mode and Kong always closes the TCP connections with an error log entry of no routes found. This works in DB-mode.

### Steps To Reproduce

Populate Kong with following configuration:
```
_format_version: "1.1"
services:
- connect_timeout: 60000
  host: httpbin.org
  name: httpbin.httpbin.80
  port: 80
  protocol: tcp
  read_timeout: 60000
  retries: 5
  write_timeout: 60000
  routes:
  - name: httpbin.example-app1.0
    protocols:
    - tcp
    destinations:
    - port: 9000
```

`telnet localhost 9000` will not work.

Run the same setup with postgres and that works.

### Additional Details & Logs

- Kong version (`$ kong version`) 2.0.2
- Operating system: Docker

dndx:
socket

Currently, the stream subsystem receives the declarative config from
the filesystem only. Once Kong is running, updating the declarative config
through `/config` admin API no longer propagates to the stream
subsystem. In the long term we will have shared shdict between `http`
and `stream` subsystem which solves the problem. However this PR
provides a temporary and clean solution for getting around the issue without
introducing new OpenResty patches, and can be reverted in the future
easily once situation changes.

Fixes #5656

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -419,7 +419,9 @@ function Kong.init()
     certificate.init()
   end
 
-  clustering.init(config)
+  if subsystem == "http" then
+    clustering.init(config)
+  end
 
   -- Load plugins as late as possible so that everything is set up
   assert(db.plugins:load_plugin_schemas(config.loaded_plugins))
@@ -550,7 +552,9 @@ function Kong.init_worker()
     go.manage_pluginserver()
   end
 
-  clustering.init_worker(kong.configuration)
+  if subsystem == "http" then
+    clustering.init_worker(kong.configuration)
+  end
 end
 
 
@@ -1152,6 +1156,41 @@ function Kong.serve_cluster_listener(options)
   kong_global.set_phase(kong, PHASES.cluster_listener)
 
   return clustering.handle_cp_websocket()
+end
+
+
+do
+  local declarative = require("kong.db.declarative")
+  local cjson = require("cjson.safe")
+
+  function Kong.stream_config_listener()
+    local sock = ngx.req.socket()
+    if not sock then
+      ngx_log(ngx_CRIT, "unable to obtain request socket")
+      return
+    end
+
+    local data, err = sock:receive("*a")
+    if err then
+      ngx_log(ngx_CRIT, "unable to receive new config: ", err)
+      return
+    end
+
+    local parsed = cjson.decode(data)
+
+    local ok, err = concurrency.with_worker_mutex({ name = "dbless-worker" }, function()
+      return declarative.load_into_cache_with_events(parsed[1], parsed[2])
+    end)
+
+    if err == "no memory" then
+      kong.log.err("not enough cache space for declarative config")
+      return
+    end
+
+    if not ok then
+      kong.log.err("failed loading declarative config into cache: ", err)
+    end
+  end
 end
 
 
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -1164,9 +1164,9 @@ do
   local cjson = require("cjson.safe")
 
   function Kong.stream_config_listener()
-    local sock = ngx.req.socket()
+    local sock, err = ngx.req.socket()
     if not sock then
-      ngx_log(ngx_CRIT, "unable to obtain request socket")
+      kong.log.crit("unable to obtain request socket: ", err)
       return
     end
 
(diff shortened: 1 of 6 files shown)


kong/init.lua
```diff
@@ -1155,4 +1159,38 @@ function Kong.serve_cluster_listener(options)
 end
 
 
+do
+  local declarative = require("kong.db.declarative")
+  local cjson = require("cjson.safe")
+
+  function Kong.stream_config_listener()
+    kong_global.set_phase(kong, PHASES.admin_api)
```
hbagdi, line 1167: This is a bit hacky but I do vaguely recall running into the same problem in any server block that uses PDK. Is it possible to avoid this in any way?

kong/init.lua
```diff
@@ -1155,4 +1159,38 @@ function Kong.serve_cluster_listener(options)
...
+
+    local sock = ngx.req.socket()
+    if not sock then
+      ngx_log(ngx_CRIT, "unable to obtain request socket")
```
hbagdi, line 1171: `return` at this point.

kong/init.lua
```diff
@@ -1155,4 +1159,38 @@ function Kong.serve_cluster_listener(options)
...
+    end
+
+    local data, err = sock:receive("*a")
+    if err then
+      ngx_log(ngx_CRIT, "unable to receive new config: ", err)
```
hbagdi, line 1176: `return` at this point.

kong/templates/nginx_kong_stream.lua
```diff
@@ -119,5 +119,15 @@ server {
         Kong.log()
     }
 }
+
+server {
+    listen unix:${{PREFIX}}/stream_config.sock;
+
+    error_log  ${{PROXY_ERROR_LOG}} ${{LOG_LEVEL}};
```
hbagdi, line 126: The `content_by_lua_block` will be executed in the context of Admin API only. Can we change the error output to `ADMIN_ERROR_LOG`?

kong/init.lua
dndx: This code path should't use PDK at all, but removing it shouldn't hurt either.

It was added only for styling reasons.

kong/init.lua
```diff
@@ -1155,4 +1159,39 @@ function Kong.serve_cluster_listener(options)
...
+    end)
+
+    if err == "no memory" then
+      kong.log.err("not enough cache space for declarative config")
+      return
+    end
+
+    if not ok then
+      kong.log.err("failed loading declarative config into cache: ", err)
+    end
```
thibaultcha, line 1192: Easier to read and less prone to forgetting a `return` here in the future should we add code below:
```lua
if not ok then
  if err == "no memory"
    kong.log.err("not enough cache space for declarative config")
  
  else
    kong.log.err("failed loading declarative config into cache: ", err)
  end

  return
end
```

Also, we should make the "no memory" error actionable - instruct users how to increase the cache space, i.e. which config property to tweak.

kong/templates/nginx_kong_stream.lua
```diff
@@ -119,5 +119,15 @@ server {
         Kong.log()
     }
 }
+
+server {
```
thibaultcha, line 123: We should probably only inject this block when `strategy` is `off` (or whenever the `/config` endpoint is available anyway)? Especially given NGINX doesn't handle unix socket file very elegantly in some cases...

kong/init.lua
```diff
@@ -1155,4 +1159,39 @@ function Kong.serve_cluster_listener(options)
...
 
 
+do
+  local declarative = require("kong.db.declarative")
+  local cjson = require("cjson.safe")
+
+  function Kong.stream_config_listener()
+    local sock = ngx.req.socket()
+    if not sock then
+      ngx_log(ngx_CRIT, "unable to obtain request socket")
```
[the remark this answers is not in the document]
thibaultcha: Also, why not catching the second return value - `err`?

kong/init.lua
```diff
@@ -1155,4 +1159,39 @@ function Kong.serve_cluster_listener(options)
...
+      return
+    end
+
+    local data, err = sock:receive("*a")
+    if err then
```
kikito, line 1174:
```suggestion
    if not data then
```

PR #5792, opened 2020-04-16
Fix/basic auth hybrid mode

### LINKED ISSUE #5649: Basic-auth plugin does not work in hybrid mode
### Summary
Basic-auth plugin does not work in hybrid mode - when I try to create consumer credentials, changes on control plane are not propagated to data plane. According to [this post](https://discuss.konghq.com/t/kong-2-hybrid-mode-and-plugins-compatibility/5565) it should work for auth plugins, and for key-auth it actually works fine. Details has been [described here](https://discuss.konghq.com/t/basic-auth-plugin-in-hybrid-mode/5689).

### Steps To Reproduce

1. Install Kong in hybrid mode (1 control plane, 1 data plane)
2. Create a consumer on control plane
3. Create a basic-auth credentials for that consumer
4. Try to use the credentials on data plane

### Additional Details & Logs
Config_hash checksum (`curl -X GET localhost:8001/clustering/status`) changes when creating a consumer or key-auth credentials, but it does not change when creating basic-auth credentials.

- Kong version (`2.0.1`)
- Kong error logs - no errors
- Operating system - Docker

kikito:
The main idea in this change is adding a new "pseudo-attribute" to the basicauth_credentials table.

* When exporting, the password is allways called `encrypted_password`.
* When importing, unencrypted passwords can be specified via the `password` field, and encrypted passwords can be specified via the `encrypted_password` field.

In order to do this, a new `export_as` field attribute has been added to the fields. This just renames the field.

The only other needed structural change was making transformations aware of "shorthands" (AKA pseudo-fields) so the encryption transformation could be undone when loading credentials from declarative conf.

fixes #5649

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -248,9 +248,13 @@ function declarative.export_from_db(fd)
 
     local name = schema.name
     local fks = {}
+    local export_as_fields = {}
     for name, field in schema:each_field() do
       if field.type == "foreign" then
         table.insert(fks, name)
+      end
+      if field.export_as then
+        export_as_fields[name] = field.export_as
       end
     end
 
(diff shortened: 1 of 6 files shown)


p0pr0ck5:
Pedantic request- can we use the prefix term “hashed” instead of “encrypted”, both as a more more technically accurate descriptor, and to avoid conflict with parts of the codebase that implement encryption in the DAO

kikito:
Closing in favor of #5835

PR #5794, opened 2020-04-17
fix(cors): don't override Vary header

### LINKED ISSUE #4782: Cors plugin overrides existing vary header values
### Summary

Cors plugin changes vary header of all responses to "Origin". Previous values like "Accept" (used for content negotiation here) are lost.

### Steps To Reproduce

1. create a dummy backend (e.g. node) with a hello world response and set response header "Vary" to "Accept"
2. enable cors plugin, no matter which configuration
3. call it ;-)
4. response header "Vary" is set to "Origin" 

### Additional Details & Logs

- Kong version 1.1.2

aldor007:
### Summary

Use add_header instead of set_header for Vary header as it can be used in other use cases. 

### Full changelog

* use add_header instead of set_hedear in cors plugin 

### Issues resolved

Fix #4782

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -52,10 +52,21 @@ local function normalize_origin(domain)
   }
 end
 
+local function add_vary_header(header_filter)
+  if header_filter then
+    kong.response.add_header("vary", "Origin")
+  else
+    kong.response.set_header("vary", "Origin")
+  end
+end
 
-local function configure_origin(conf)
+local function configure_origin(conf, header_filter)
   local n_origins = conf.origins ~= nil and #conf.origins or 0
   local set_header = kong.response.set_header
+
+  -- always set Vary header (it can be used for calculating cache key)
+  -- https://github.com/rs/cors/issues/10
+  add_vary_header(header_filter)
 
   if n_origins == 0 then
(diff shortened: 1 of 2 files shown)


aldor007:
Any chance that this PR will be reviewed?

kong/plugins/cors/handler.lua
```diff
@@ -52,8 +52,24 @@ local function normalize_origin(domain)
...
+  local add_header = kong.response.add_header
+  local set_header = kong.response.set_header
+
+  if header_filter then
+    local vary_header = kong.response.get_header("Vary")
+    if vary_header then
+      return add_header("vary", "Origin")
+    end
+  end
+  set_header("vary", "Origin")
```
kikito, line 69: Hi, please change this section to:

```suggestion
  if header_filter then
    kong.response.add_header("vary", "Origin")
  else
    kong.response.set_header("vary", "Origin")
  end
```

aldor007: Hi @kikito 
I've changed these lines as requested and I've updated tests because `if` statement was necessary to deduplicate vary header.

But I've added one new fix(I think) to this PR.  Vary header is always returned when using cors plugin.
Why?
A very good explanation can be found here https://github.com/rs/cors/issues/10

I'm open to discussion and I'm waiting for feedback.

PR #5795, opened 2020-04-17
chore(deps) bumps lyaml (6.2.4 to 6.2.5) and libyaml (0.2.2 to 0.2.3)

bungle:
#### lyaml

- luke really propagates YAML_BINDIR, YAML_DIR, YAML_INCDIR and YAML_LIBDIR
  to checksymbol test in lukefile given the change to external_dependencies
  layout in 6.1.2.

#### libyaml

- Fixed typo.
- Fix typo in comment
- Use pointer to const for strings that aren't/shouldn't be modified
- Squash a couple of warnings in example-deconstructor-alt
- Fix spelling for error message
- Make appveyor config be a hidden file
- Add CHANGES file
- Always output document end before directive (YAML 1.2 compatibility)
- Output document end marker after open ended scalars
- change cmake target name from libOFF.a to libyaml.a
- include/yaml.h: fix comments
- Fixed missing token in example
- Avoid recursion in the document loader.
- Support %YAML 1.2 directives
- Change dllexport controlling macro to use _WIN32


bungle:
I labeled it as `do not merge` as it seems it causes problem with Github Actions.

hutchic:
should this be against master?

Issue #5748, opened 2020-04-02
Client get SSL internal error


Guojian:
### Summary
We use Kong as kubernetes ingress controller.
This environment is an ARM-based environment, and we figure out a very strange issue. 
Through the Kong dynamic certificate binding mechanism, we bind a SSL certificate to Kong for HTTPS connection. And it triggers an error about TLS shake-hands, the error rate is only about 5%. Through cURL, the error message is “curl: (35) error:14004438:SSL routines:CONNECT_CR_SRVR_HELLO:tlsv1 alert internal error”

### Steps To Reproduce

1. Install kubernetes kong ingress controller
2. Bind the SSL certificate through kubernetes secret 
3. Locally curl https address
4. Over about 100 times, it happen 5 times.

### Additional Details & Logs

- Kong version 1.4.3
- Kong debug-level startup logs (`$ kong start --vv`)
- Kong error logs ( It does not appear any error message)
- Kong configuration (the output of a GET request to Kong's Admin port - see
  https://docs.konghq.com/latest/admin-api/#retrieve-node-information)
- Operating system: Ubuntu 9.1.0-2ubuntu2~16.04

kikito:
Hi, is it possible for you to try a more recent version of Kong, like 1.5.1 or even 2.0.2?

guanlan:
Similar to this report https://github.com/Kong/kong/issues/1464#issuecomment-254502910.

Also, can you provide detailed error logs and debug logs?
You can add an environment variable in Kubernetes CRD `KONG_LOG_LEVEL=debug` to enable the debug logs.

fffonion:
Based on the logs provided, this is the same issue as #5042 and i'm able to reproduce using the script from mentioned issue with kong 1.5.1.

```
2020/04/08 05:56:37 [error] 27#0: *1087 lua entry thread aborted: runtime error: /usr/local/share/lua/5.1/kong/cache.lua:243: key at index 2.1217691418087e-314 must be a string for operation 1 (got nil)
stack traceback:
coroutine 0:
	[C]: in function 'error'
	/usr/local/share/lua/5.1/resty/mlcache.lua:985: in function 'get_bulk'
	/usr/local/share/lua/5.1/kong/cache.lua:243: in function 'get_bulk'
	/usr/local/share/lua/5.1/kong/runloop/certificate.lua:185: in function 'find_certificate'
	/usr/local/share/lua/5.1/kong/runloop/certificate.lua:210: in function 'execute'
[...]

PR #5797, opened 2020-04-17
fix(cache) disable JIT mlcache:get_bulk() on ARM64

javierguerragiraldez:
The loop inside this function can trigger a trace restart that results
in a wrong value in the local index variable.   (#5748)

until this is fixed in LuaJIT, disabling JIT for this function
avoids the problem.


thibaultcha:
An update here - the issue has been fixed in OpenResty's fork of LuaJIT already, which will be bundled in OpenResty's 1.17.8.1 release. See https://github.com/LuaJIT/LuaJIT/issues/579#issuecomment-623803502.

PR #5802, opened 2020-04-21
feat(dao) hooks

gszr:
Introduce a hooks module and add hooking points to the DAO.
Hooks are named functions that are executed at specific locations
previously specified with a `run_hook` call.

### DIFF
--- a/kong/hooks.lua
+++ b/kong/hooks.lua
@@ -1,0 +1,73 @@
+local _M = {}
+
+
+local hooks = {}
+
+
+local ipairs = ipairs
+local pack = table.pack
+local unpack = table.unpack
+local insert = table.insert
+
+
+local function wrap_hook(f)
+   return function(acc, ...)
+      if acc and not acc[1] then
+         return acc
+      end
+      return pack(f(...))
+   end
+end
+
+
+-- Register a hook function.
+-- @param name Name of the hook; names should be namespaced
+-- so that they don't conflict: e.g. "dao:upsert:pre"
+-- @param hook Hook function, which receives the arguments
+-- passed to run_hook(). By default, if a previous hook function
+-- returned nil, the hook function will not execute and
+-- the return values from the previous one will be returned as
+-- the result of run_hook().
+-- @param opts Table of options:
+-- * "low_level" - if true, hook is assumed to be a "low-level hook
+--   function": the low-level function receives an array in
+--   table.pack() format as the first argument and is expected
+--   to return a similar array. It can decide to run even if
+--   a previous hook failed.
+function _M.register_hook(name, hook, opts)
+  assert(type(hook) == "function", "hook must be a function")
+
+  hooks[name] = hooks[name] or {}
+
+  local f
+  if opts and opts.low_level then
+    f = hook
+  else
+    f = wrap_hook(hook)
+  end
+
+  insert(hooks[name], f)
+end
+
+
+function _M.run_hook(name, ...)
+  if not hooks[name] then
+    return (...) -- return only the first value
+  end
+
+  local acc
+
+  for _, f in ipairs(hooks[name] or {}) do
+    acc = f(acc, ...)
+  end
+
+  return unpack(acc, 1, acc.n)
+end
+
+
+function _M.clear_hooks()
+  hooks = {}
+end
+
+
+return _M
--- a/spec/01-unit/019-hooks_spec.lua
+++ b/spec/01-unit/019-hooks_spec.lua
@@ -1,0 +1,86 @@
+local hooks = require "kong.hooks"
+
+
+describe("hooks", function()
+  local c = 0
+  local function s()
+    c=c+1
+    return tostring(c)
+  end
+
+  local h
+  before_each(function()
+    h = s()
+  end)
+
+  it("runs a non-initialized hook and doesn't err", function()
+    assert.not_error(function() hooks.run_hook(h) end)
+  end)
+
+  it("wraps 1 function by default", function()
+    hooks.register_hook(h, function() return 1 end)
+    assert.equal(1, hooks.run_hook(h))
+  end)
+
+  it("wraps 2 functions by default", function()
+    hooks.register_hook(h, function() return 1 end)
+    hooks.register_hook(h, function() return 1 end)
+    assert.equal(1, hooks.run_hook(h))
+  end)
+
+  it("by default first that returns nil/false cuts the flow", function()
+    hooks.register_hook(h, function() return nil end)
+    hooks.register_hook(h, function() return 1 end)
+    assert.is_nil(hooks.run_hook(h))
+  end)
+
+  it("errors when passing a non-function parameter to a hook", function()
+    assert.is_falsy(pcall(function() hooks.register_hook(h) end))
+  end)
+
+  it("works with multiple return values", function()
+    hooks.register_hook(h, function() return 1,2 end)
+    assert.same({1,2}, {hooks.run_hook(h)})
+  end)
+
+  it("calls multiple functions", function()
+    local count = 0
+    local f = function()
+      count = count+1
+      return true
+    end
+
+    hooks.register_hook(h, f)
+    hooks.register_hook(h, f)
+
+    hooks.register_hook(h, function() return 1 end)
+
+    assert.equal(1, hooks.run_hook(h))
+    assert.equal(2, count)
+  end)
+
+  it("calls with parameters", function()
+    local f = function(a, b)
+      return {a,b}
+    end
+
+    hooks.register_hook(h, f)
+    hooks.register_hook(h, f)
+    local r = hooks.run_hook(h, 1, 2)
+
+    assert.same({1, 2}, r)
+  end)
+
+  it("can run low-level functions", function()
+    local function f(acc, i)
+      acc = acc or {}
+      table.insert(acc, acc[#acc] and acc[#acc]+1 or 0)
+      return acc
+    end
+    hooks.register_hook(h, f, {low_level=true})
+    hooks.register_hook(h, f, {low_level=true})
+    hooks.register_hook(h, f, {low_level=true})
+
+    assert.same({0,1,2}, {hooks.run_hook(h, 0)})
+  end)
+end)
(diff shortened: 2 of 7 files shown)


kong/hooks.lua
```diff
@@ -0,0 +1,75 @@
...
+    f = hook
+  else
+    f = wrap_hook(hook)
+  end
+
+  insert(hooks[name], f)
+end
+
+
+-- Register a hook function.
```
kikito, line 39: This whole doc comment belongs to `_M.register_hook` (it is documenting the wrong function).

kong/hooks.lua
```diff
@@ -0,0 +1,75 @@
...
+   return function(acc, ...)
+      if acc and not acc[1] then
+         return acc
+      end
+      return pack(f(...))
+   end
+end
+
+
+function _M.register_hook(name, hook, opts)
```
kikito, line 23: The name `raw` seems very ... "implementation oriented": it is "raw" because it is "not wrapped" - but the fact that we are wrapping it into something else is an implementation detail. What the user of this function cares about is whether this hook will execute when there are errors in other hooks before it.

I therefore recommend renaming the option to `ignore_errors` or even `ignore_previous_errors`.

spec/01-unit/019-hooks_spec.lua
```diff
@@ -0,0 +1,88 @@
...
+
+describe("hooks", function()
+  local h
+  before_each(function()
+    h = s()
+  end)
+
+  it("runs a non-initialized hook and doesn't err", function()
+    hooks.run_hook(h)
+    assert(true)
```
kikito, line 18:
```suggestion
    assert.not_error(function() hooks.run_hook(h) end)
```

spec/01-unit/019-hooks_spec.lua
```diff
@@ -0,0 +1,88 @@
+local hooks = require "kong.hooks"
+
+local c = 0
+local function s()
+  c=c+1
+  return tostring(c)
```
kikito, line 6: Is the return of this function (the value of c) used on any test? Couldn't `h` just be initialized to `function() return true end`?

spec/01-unit/019-hooks_spec.lua
```diff
@@ -0,0 +1,88 @@
...
+    local f = function(a, b)
+      return {a,b}
+    end
+
+    hooks.register_hook(h, f)
+    hooks.register_hook(h, f)
+    local r = hooks.run_hook(h, 1, 2)
+
+    assert.equal(1, r[1])
+    assert.equal(2, r[2])
```
kikito, line 73:
```suggestion
    assert.same({1, 2}, r)
```

spec/01-unit/019-hooks_spec.lua
```diff
@@ -0,0 +1,88 @@
...
+  end)
+
+  it("can run raw functions", function()
+    local function f(acc, i)
+      acc = acc or {}
+      table.insert(acc, acc and acc[#acc] and acc[#acc]+1 or 0)
```
kikito, line 79: The previous line ensures `acc` exists:
```suggestion
      table.insert(acc, acc[#acc] and acc[#acc] + 1 or 0)
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+        name = "s1",
+        url = "http://example.test",
+      })
+
+      r1 = assert(bp.routes:insert {
+        protocols = { "http" },
+        hosts = { "host1" },
+        service = s1,
+        name = "r1",
+      })
```
kikito, line 25: `bp.services:insert` already does `assert`:

```suggestion
      s1 = bp.services:insert({
        name = "s1",
        url = "http://example.test",
      })

      r1 = bp.routes:insert({
        protocols = { "http" },
        hosts = { "host1" },
        service = s1,
        name = "r1",
      })
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+          pre_hook()
+          return true
+        end)
+        hooks.register_hook("dao:page_for:post", function()
+          post_hook()
+          return true
+        end)
+      end)
+
+      lazy_teardown(function()
```
kikito, line 43: Ensure this happens even if errors are raised
```suggestion
      finally(function()
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+        protocols = { "http" },
+        hosts = { "host1" },
+        service = s1,
+        name = "r1",
+      })
+    end)
+
+    describe("page_for", function()
+      local pre_hook = spy.new(function() end)
+      local post_hook = spy.new(function() end)
```
kikito, line 30: Spies should be able to return true directly, you shouldn't have to "wrap them up":

```suggestion
      local pre_hook = spy.new(function() return true end)
      local post_hook = spy.new(function() return true end)
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+
+      lazy_setup(function()
+        hooks.register_hook("dao:page_for:pre", function()
+          pre_hook()
+          return true
+        end)
+        hooks.register_hook("dao:page_for:post", function()
+          post_hook()
+          return true
+        end)
```
kikito, line 40:
```suggestion
        hooks.register_hook("dao:page_for:pre", pre_hook)
        hooks.register_hook("dao:page_for:post", post_hook)
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+      it("calls hooks", function()
+        assert(db.routes:page_for_service(s1))
+        assert.spy(pre_hook).was_called(1)
+        assert.spy(post_hook).was_called(1)
+      end)
+    end)
+
+    describe("select_by", function()
+      local pre_hook = spy.new(function() end)
+      local post_hook = spy.new(function() end)
```
kikito, line 56:
```suggestion
      local pre_hook = spy.new(function() return true end)
      local post_hook = spy.new(function() return true end)
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+
+      lazy_setup(function()
+        hooks.register_hook("dao:select_by:pre", function()
+          pre_hook()
+          return true
+        end)
+        hooks.register_hook("dao:select_by:post", function()
+          post_hook()
+          return true
+        end)
```
kikito, line 66:
```suggestion
        hooks.register_hook("dao:select_by:pre", pre_hook)
        hooks.register_hook("dao:select_by:post", post_hook)
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+      end)
+
+      lazy_teardown(function()
```
kikito, line 69:
```suggestion
      finally(function()
```

spec/02-integration/03-db/12-dao_hooks_spec.lua
```diff
@@ -0,0 +1,280 @@
...
+        end)
+        hooks.register_hook("dao:update_by:post", function()
+          post_hook()
+          return true
+        end)
+      end)
+
+      lazy_teardown(function()
+        hooks.clear_hooks()
+      end)
```
kikito, line 97:
```suggestion
      local pre_hook = spy.new(function() return true end)
      local post_hook = spy.new(function() return true end)

      lazy_setup(function()
        hooks.register_hook("dao:update_by:pre", pre_hook)
        hooks.register_hook("dao:update_by:post", post_hook)
      end)

      finally(function()
        hooks.clear_hooks()
      end)
```

kikito: Same with the rest of the hooks 👇👇👇👇👇

spec/01-unit/019-hooks_spec.lua
gszr: `h` is set before each test to the name of the current test's hook

gszr: so the return is used

spec/02-integration/03-db/12-dao_hooks_spec.lua
gszr: In what contexts is `finally` available?

spec/02-integration/03-db/12-dao_hooks_spec.lua
gszr: This does not work. The hook must be a function, while `spy.new` returns a table with a __call metamethod I think.

kong/hooks.lua
gszr: Given this has been agreed upon before with @hishamhm and @kidd, I'd also like to hear their thoughts.

hishamhm: @kikito Understood, but that was not the rationale.

`raw` here was meant to imply that the function needs to follow the low-level protocol for hook functions when it cares about controlling the chaining sequence. The fact that the default chaining sequence behavior is to abort on errors is meant to be the implementation detail.

What raw functions allow is a lot more than ignoring previous errors. You could for example take the output from the previous hook and combine it to the result of the current hook in arbitrary ways. (In other words, you can also implement map, reduce, etc. — in other other words, what the hooks system implements is a fold, which is a super general construct for a super general feature).

The documentation there (mea culpa, I wrote the docs bit) is misleading by focusing on the "handling errors" example. Perhaps it should have even left out the default chaining behavior altogether and leave it unspecified (the more I think about it, the more I think it should, because you should only use the not-raw ones if you _don't care_ about the chaining sequence). If all the `raw` flag did was `ignore_previous_errors`, we wouldn't need the general system.

Perhaps we could can another name like `low_level` to avoid the "raw vs. wrapped" comparison — that's good feedback, thanks for bringing it up!

gszr: Thanks for the comment @hishamhm. @kikito and I just met and made two changes:
- Renamed the flag to `low_level`
- Adjusted the docs to not focus on the error handling part

PR #5804, opened 2020-04-22
fixes: various issues in balancer and cluster_events

thibaultcha:


kong/runloop/balancer.lua
```diff
@@ -908,6 +908,7 @@ local function execute(target, ctx)
                                           hash_value)
     if not ip and
       (port == "No peers are available" or port == "Balancer is unhealthy") then
+      log(ERR, "failed to get a peer from the ring-balancer (", port, ")")
```
Tieske, line 911: I don't think this is correct. This is not a "code error", but a user(=admin) error, hence it should not be reported through the system-logs, but through a traffic-log (aka a log plugin). Also because this would in a high-load system fill up the error logs.

spec/02-integration/06-invalidations/01-cluster_events_spec.lua
```diff
@@ -314,6 +314,39 @@ for _, strategy in helpers.each_strategy() do
         assert(cluster_events_1:poll())
         assert.spy(spy_func).was_called(1) -- called
       end)
+
+      it("broadcasts an event with an polling delay for subscribers", function()
```
Tieske, line 318: typo 'an' -> 'a'

kong/runloop/balancer.lua
thibaultcha: The intent was clear. This was useful when debugging the balancer over the last few days and could be useful when receiving future error logs for review around balancer issues.

Tieske: Not sure what you mean by that? want to keep it in, or should it go?

spec/02-integration/06-invalidations/01-cluster_events_spec.lua
Tieske: updated

Tieske:
@thibaultcha I removed the commit with the proxy-cache version bump, fixed the typo and added 2 additional improvements.

LGTM now, only thing open is the [comment wrt log message](https://github.com/Kong/kong/pull/5804#discussion_r412867127); should it stay or should it go?

thibaultcha:
@Tieske Well, now CI won't pass without the bump here and we cannot merge the version bump since the 2.0.4 release is pending. I don't think that was really necessary but whatever.

kong/runloop/balancer.lua
thibaultcha: I don't have strong feelings. Do you?

kong/db/dao/targets.lua
```diff
@@ -65,7 +65,10 @@ local function clean_history(self, upstream_pk)
     ngx.log(ngx.NOTICE, "[Target DAO] Starting cleanup of target table for upstream ",
                tostring(upstream_pk.id))
     local cnt = 0
-    for _, entry in ipairs(delete) do
+    -- reverse again; so deleting oldest entries first
+    for i = #delete, 1, -1 do
```
thibaultcha, line 69: Great, I noticed this behavior as well and was considering a similar fix, so we are on the same page here :+1:

kong/runloop/balancer.lua
Tieske: I don't want it, because on a busy system you'll get hundreds of these errors if there is a mistake. Hence I think they belong in the traffic logs, and not system logs.

I'll remove it

Tieske:
> @Tieske Well, now CI won't pass without the bump here and we cannot merge the version bump since the 2.0.4 release is pending. I don't think that was really necessary but whatever.

why is that? what's the relation? I'm clearly missing something here

Tieske:
ok, moved the proxy-cache plugin back into this PR. Should go green now.

thibaultcha:
> why is that? what's the relation? I'm clearly missing something here

Since the broadcasting delay is now propagated (as it should have always been the case), these `wait_until` helpers needs several runs, so we need to consume the response in the sockets: https://github.com/Kong/kong-plugin-proxy-cache/commit/52472a40db98b5737e1b802e23ad1163c15d7181.

kong/runloop/balancer.lua
thibaultcha: For us though, it'd still be helpful to distinguish between the two possible errors here. We should consider slightly updating the error message to help making that distinction.

thibaultcha: Merging as-is for inclusion in 2.0.4 - we can add such a distinction later.

PR #5771, opened 2020-04-10
feat(balancer) dynamic upstream keepalive pools

thibaultcha:
This replaces the keepalive implementation from ngx_http_upstream_module with the one proposed by openresty/lua-nginx-module#1600. It is proposed against `next` due to the potentially disruptive nature of such a large change in production environments. It is my advice that this change be first introduced in a release candidate.

For the time being, the ngx_http_lua and lua-resty-core patches have been applied to our openresty-build-tools repository and are applied when building OpenResty for Kong. The final API may be subject to changes by OpenResty, but incorporating the changes in Kong ahead of time reduces possible virtual host confusion issues and/or attack vectors.

Sibling PR with the aforementioned patches: https://github.com/Kong/kong-build-tools/pull/257

The proposed OpenResty patches offer a way to circumvent the NGINX limitations detailed in the following tickets:

* https://trac.nginx.org/nginx/ticket/1340
* https://trac.nginx.org/nginx/ticket/1593

The behavior changes in Kong from this work are:

* Avoid virtual host confusion when proxying upstream virtual services over TLS and mTLS (based on SNI and/or client certificate id).
* Allow for an indefinite idle timeout and max requests for connections in the keepalive pool (a behavior reverted by NGINX in 1.15.3).
* Introduce new properties for upstream keepalive pooling.
* Ensure older properties are properly deprecated and that their settings cascades to the new ones.
* Use different examples for NGINX directives injection.

In the future, we could implement per-Service (or per-Upstream) keepalive pools, allowing users to granularly enable/disable, or configure upstream keepalive pools as they wish.

Fix #2835
Fix #3788
Fix #4001

cc @jeremyjpj0916 @rsbrisci

Issue #5054, opened 2019-09-20
 [error] 84#0: *1136877 connect() failed (111: Connection refused) while connecting to upstream, client: 172.18.0.1, server: kong, request: "GET /info HTTP/1.1", upstream: "http://127.0.0.1:86/info", host: "localhost:8000"


bali48:
Please read the CONTRIBUTING.md guidelines to learn on which channels you can
seek for help and ask general questions:

https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#where-to-seek-for-help

### Summary

My service is running on localhost:86 
and kong is running on localhost:8001 //for admin
and kong proxy is running on localhost:8000 //for proxy
i add single service through this command 
curl -i -X POST --url http://localhost:8001/services/ --data 'name=BooksApi' --data 'url=http://localhost:86'

and add single route using this command 
curl -i -X POST --url http://localhost:8001/services/BooksApi/routes --data 'hosts[]=localhost' --data 'paths[]=/' --data 'strip_path=false' --data 'methods[]=GET'

but i am facing this error 
Kong Error
An invalid response was received from the upstream server.

and according to error log 
[error] 84#0: *1136877 connect() failed (111: Connection refused) while connecting to upstream, client: 172.18.0.1, server: kong, request: "GET /info HTTP/1.1", upstream: "http://127.0.0.1:86/info", host: "localhost:8000"

### Additional Details & Logs

- Kong version ('1.3.0`)

p0pr0ck5:
Connection refused implies that the upstream received and rejected the connection request because no application on port 86 was running and available to respond to requests. You will want to check your upstream service.

bali48:
my webservice is running on port 86 and return a this json response 
{"data":"Books -> Lumen (5.7.7) (Laravel Components 5.7.*)","status":"200"}

p0pr0ck5:
Is it on the same machine as kong? You’ve set the upstream address as 127.0.0.1. Your service must be on the same machine as kong (not the same docker/container boat, the same logical machine)

luka29:
seeing a similar issue. my python server is accessible on 127.0.0.1:5000 but kong fails to access it:
kong_1            | 2019/10/17 07:52:25 [error] 31#0: *14090 connect() failed (111: Connection refused) while connecting to upstream, client: 172.24.0.1, server: kong, request: "GET /bazaar/ping HTTP/1.1", upstream: "http://127.0.0.1:5000/ping", host: "localhost:8000"

[...]

PR #5812, opened 2020-04-23
feat(dao) support different read and write consistency levels in cassandra

Abhishekvrshny:
### Summary

This adds the capability to define different read and write
consitency levels for cassandra. Earlier both of these were set
by a single `cassandra_consistency` configuration parameter.

This PR introduces `cassandra_write_consistency` and
`cassandra_read_consistency` configuration parameters and removes
`cassandra_consistency` parameter.


bungle:
Hi @Abhishekvrshny, nice PR. Should we still keep `cassandra_consistency` for backward compatibility. E.g. if a more specific consistency is not defined, use that?

Abhishekvrshny:
@bungle Sure. In that case, should the PR be against `master` or `next`? Also, is there a way to mark a config as deprecated somehow, for users to migrate?
And I see some PR checks failing, not sure how to resolve those.

Abhishekvrshny:
@bungle I have made the suggested changes to the best of my understanding. Please see if this works.

Abhishekvrshny:
@bungle Checking if you got a chance to review this.

kong.conf.default
```diff
@@ -821,9 +821,23 @@
...
+#cassandra_consistency = DISABLED     # Consistency setting to use when reading/
+                                      # writing to the Cassandra cluster.
+                                      # Note: `cassandra_write_consistency` and
+                                      # `cassandra_read_consistency` should be
+                                      # preferred over this, with this `DISABLED`
+                                      # If this is `DISABLED`, `cassandra_write_consistency`
+                                      # and `cassandra_read_consistency` would be used,
+                                      # else `cassandra_consistency` takes precedence over
+                                      # `cassandra_write_consistency` and `cassandra_read_consistency`
+                                      # It exists only for backward compatibility.
```
thibaultcha, line 833: We do not leave deprecated properties in `kong.conf.default`, this property can/should be removed from here.

kong/templates/kong_defaults.lua
```diff
@@ -115,7 +115,9 @@ cassandra_ssl = off
 cassandra_ssl_verify = off
 cassandra_username = kong
 cassandra_password = NONE
-cassandra_consistency = ONE
+cassandra_consistency = DISABLED
```
thibaultcha, line 118: Should be set to `NONE` instead?

kong/conf_loader.lua
```diff
@@ -336,6 +336,38 @@ local CONF_INFERENCES = {
...
+     "LOCAL_QUORUM",
+     "ONE",
+     "TWO",
+     "THREE",
+     "LOCAL_ONE",
+    },
+    deprecated = {
+      replacement = "cassandra_write_consistency / cassandra_read_consistency",
+      alias = function(conf)
+        if conf.cassandra_consistency ~= "DISABLED" then
```
thibaultcha, line 352: I don't think that we need this new enum value. By replacing the default value to `NONE`, we already prevent new instances of the config from loading it (therefore only r/w consistencies are loaded). Then, if `cassandra_consistency` is specified, we overwrite the r/w consistencies in this aliasing. The warning log will instruct users that the property is deprecated and that they should favor the new r/w properties instead.

Abhishekvrshny:
@thibaultcha Thanks for the review. I have made the required changes.

spec/01-unit/03-conf_loader_spec.lua
```diff
@@ -516,13 +516,41 @@ describe("Configuration loader", function()
...
+      }))
+      assert.equal("TWO", conf.cassandra_read_consistency)
+      assert.equal("QUORUM", conf.cassandra_write_consistency)
+
+      conf = assert(conf_loader(nil, {
+        cassandra_consistency = "QUORUM",
+        cassandra_write_consistency = "LOCAL_QUORUM"
+      }))
+      assert.equal("QUORUM", conf.cassandra_read_consistency)
+      assert.equal("LOCAL_QUORUM", conf.cassandra_write_consistency)
```
thibaultcha, line 552: Please add new tests around the deprecation of properties to a new `describe` block instead. Leave this test unchanged, and add a new `describe` block to this test suite. I did the same in #5771 which isn't merged yet, so you can just add the new `describe` block yourself, in the same place as I added it in #5054: 

```lua
describe("deprecated properties", function()
  it("cassandra_consistency -> cassandra_r/w_consistency", function()

  end)
end)
```

kong/conf_loader.lua
Abhishekvrshny: @thibaultcha done

spec/01-unit/03-conf_loader_spec.lua
Abhishekvrshny: @thibaultcha done

Abhishekvrshny:
@thibaultcha gentle reminder for reviewing the last change. Thanks.

thibaultcha:
Merged to next with some edits, thank you!

hbagdi:
@Abhishekvrshny Please fill out the following form to claim your contributor swag: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt

PR #5831, opened 2020-04-28
fix(balancer) don't cache an empty upstream name dict

### LINKED ISSUE #5455: DNS resolution failed in dbless mode.
Recently, i often got an error below in dbless mode. It will return to noraml when restarting kong. 

> 2020/01/09 12:05:19 [error] 42406#0: *151691 [lua] balancer.lua:917: execute(): DNS resolution failed: dns server error: 3 name error. Tried: ["(short)upstream_8:(na) - cache-miss","upstream_8:33 - cache-miss/scheduled/querying/dns server error: 3 name error","upstream_8:1 - cache-miss/scheduled/querying/dns server error: 3 name error","upstream_8:5 - cache-miss/scheduled/querying/dns server error: 3 name error"], client: 192.168.63.1, server: kong, request: "POST /hr/1 HTTP/1.1", host: "gateway.corp.com:8000"

javierguerragiraldez:
### Summary

Issue #5455 reports Kong instances getting stuck with DNS resolution errors when using Kubernetes Ingress Controller.  In those cases, the DNS query won't work, but it should've been resolved first by the `get_balancer()` function.

Since this function can't be called at `balance` phase, if it didn't work earlier (at `access` phase), there's no other option than reporting failure.  It seems that there's some intermittent failure on reading the upstreams list, and once it returns an empty list, it's stored in the cache.

This patch avoids saving in the cache an empty upstream list, in the hope that a subsequent query will find the real content.

Fix #5455

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -546,6 +546,8 @@ local get_all_upstreams
 do
   local function load_upstreams_dict_into_memory()
     local upstreams_dict = {}
+    local found = nil
+
     -- build a dictionary, indexed by the upstream name
     for up, err in singletons.db.upstreams:each() do
       if err then
@@ -554,8 +556,10 @@ do
       end
 
       upstreams_dict[up.name] = up.id
+      found = true
     end
-    return upstreams_dict
+
+    return found and upstreams_dict
   end
   _load_upstreams_dict_into_memory = load_upstreams_dict_into_memory
 


Tieske:
this doesn't seem right to me. It is supposed to load the upstreams, if there are none, an empty table is acceptable. Creating an upstream should be generating an invalidation event, which should cause the code to re-read the upstreams from the config. If that doesn't happen, then its just a symptom of something else being wrong.

javierguerragiraldez:
Yes, something else is wrong, the query for upstreams sometimes returns empty spontaneously.  @dndx has found a race condition somewhere, but now don't know if this is a symptom of that

carnei-ro:
@gszr we could reproduce once, I've recorded the to explain the scenario https://www.youtube.com/watch?v=w69NekUyRCk

PR #5835, opened 2020-04-28
feat/basic auth hybrid mode (take 2)

### LINKED ISSUE #5649: Basic-auth plugin does not work in hybrid mode

kikito:
Alternative name for this PR: feat/declarative config _transform.

This change fixes #5649 (basic-auth credentials not working on hybrid mode).

It is an alternative to #5792 where the proposed fix involves adding a new pseudo-field called `encrypted_password` to the entities.

The main drawback of that approach is that it requires manually changing all the entities which have transformations. Moreover, the change seems to be nearly identical in all cases: making a pseudo-field which in effect deactivates the transformation.

On this take, we add a new parameter to the declarative config file itself. The field is called `_transform`. When this field is present and set to `false`, the transformations are not executed when exporting. If the field is not present, or if it is, but it's set to `true`, the transformations are executed (which is the current behavior).

This approach does not require individual changes to any entities - a single flag in the top-level yaml is enough.

However, since this flag is new, the yaml version needed to be bumped up.

Also, this approach assumes that *on export*, we never want to do transformations (`_transform` is `false` on all exports). If we ever encounter a situation in which we *do* want to have transformations even after a straight db export, we will need to find a different way.

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -81,6 +81,7 @@ end
 -- first return item: a table with the following format:
 --   {
 --     _format_version: 1.1,
+--     _transform: true,
 --     services: {
 --       ["<uuid>"] = { ... },
 --       ...
@@ -159,6 +160,7 @@ end
 -- first return item: a table with the following format:
 --   {
 --     _format_version: 1.1,
+--     _transform: true,
 --     services: {
 --       ["<uuid>"] = { ... },
 --       ...
@@ -239,6 +241,9 @@ function declarative.load_into_db(dc_table)
     return nil, err
   end
 
+  local options = {
+    transform = dc_table._transform,
+  }
   local schema, primary_key, ok, err, err_t
   for i = 1, #sorted_schemas do
     schema = sorted_schemas[i]
@@ -248,7 +253,7 @@ function declarative.load_into_db(dc_table)
 
       primary_key = schema:extract_pk_values(entity)
 
-      ok, err, err_t = kong.db[schema.name]:upsert(primary_key, entity)
+      ok, err, err_t = kong.db[schema.name]:upsert(primary_key, entity, options)
       if not ok then
         return nil, err, err_t
       end
@@ -271,6 +276,7 @@ function declarative.export_from_db(fd)
 
   fd:write(declarative.to_yaml_string({
     _format_version = "1.1",
+    _transform = false,
   }))
 
   for _, schema in ipairs(sorted_schemas) do
@@ -325,6 +331,7 @@ function declarative.export_config()
 
   local out = {
     _format_version = "1.1",
+    _transform = false,
   }
 
   for _, schema in ipairs(sorted_schemas) do
@@ -385,6 +392,7 @@ end
 -- dc_table format:
 --   {
 --     _format_version: 1.1,
+--     _transform: true,
 --     services: {
 --       ["<uuid>"] = { ... },
 --       ...
@@ -443,12 +451,21 @@ function declarative.load_into_cache(dc_table, hash, shadow_page)
       end
     end
 
+    local transform = dc_table._transform == nil and true or dc_table._transform
     local ids = {}
     for id, item in pairs(items) do
       table.insert(ids, id)
 
       local cache_key = dao:cache_key(id)
-      item = schema:transform(remove_nulls(item))
+      item = remove_nulls(item)
+      if transform then
+        local err
+        item, err = schema:transform(item)
+        if not item then
+          return nil, err
+        end
+      end
+
       local ok, err = kong.core_cache:safe_set(cache_key, item, shadow_page)
       if not ok then
         return nil, err
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -80,7 +80,7 @@ end
 
 -- first return item: a table with the following format:
 --   {
---     _format_version: 1.1,
+--     _format_version: 1.2,
 --     _transform: true,
 --     services: {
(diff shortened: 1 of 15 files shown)


kong/db/declarative/init.lua
```diff
@@ -241,7 +275,8 @@ function declarative.export_from_db(fd)
   end
 
   fd:write(declarative.to_yaml_string({
-    _format_version = "1.1",
+    _format_version = "2.0",
```
hutchic, line 278: does this require an update to docs?

kong/templates/kong_yml.lua
```diff
@@ -12,6 +12,15 @@ return [[
 
 _format_version: "1.1"
```
bungle, line 13: Should this be updated?

hbagdi:
How hard is it to expose this feature to the Admin API as well?
decK currently sync passwords back to Kong after a reset because there is no way to skip the 'hashing' part in the Admin API.
Is adding a query string parameter to skip hashing when the password is already hashed a good idea and something we should add?

cc @hishamhm

kong/db/declarative/init.lua
kikito: Indeed. I have switched the version to 1.2 everywhere.

kong/templates/kong_yml.lua
kikito: Yes, I have updated it to 1.2.

kikito:
@hbagdi Some parts of this PR will make it easier to implement something similar (a special kong header?) on the Admin API. In particular, the DAO functions now accept a `options.transform` setting which can be used. I think it would't be very difficult to add such a flag (that falls out of the scope of this particular PR).

As a side note: my previous take on this problem (#5792) did in fact work on both the Admin API and declarative config. However a single "global switch" was preferred.

hbagdi:
Got it. Added https://github.com/Kong/kong/issues/5925 to track it.
Thanks @kikito !

kong/db/declarative/init.lua
```diff
@@ -351,7 +389,17 @@ function declarative.get_current_hash()
 end
 
 
-function declarative.load_into_cache(entities, hash, shadow_page)
+-- dc_table format:
+--   {
+--     _format_version: 1.1,
```
bungle, line 394: And this

kong/db/declarative/init.lua
```diff
@@ -144,12 +157,26 @@ function Config:parse_string(contents, filename, accept, old_hash)
 end
 
 
+-- first return item: a table with the following format:
+--   {
+--     _format_version: 1.1,
```
bungle, line 162: This too.

kong/db/declarative/init.lua
```diff
@@ -78,14 +78,27 @@ function Config:parse_file(filename, accept, old_hash)
...
```
bungle, line 83: Should we update these as well?

spec/01-unit/01-db/01-schema/11-declarative_config/02-process_auto_fields_spec.lua
```diff
@@ -23,6 +23,7 @@ describe("declarative config: process_auto_fields", function()
         config = DeclarativeConfig:process_auto_fields(config, "select", false)
         assert.same({
           _format_version = "1.1",
```
bungle, line 25: What about these?

kikito: Those are fine. The 1.2 can understand and accept 1.1-format. It is ok to test that if you put 1.1 into the process_auto_fields function, you get a 1.1 version back.

hishamhm:
> How hard is it to expose this feature to the Admin API as well?

@hbagdi According to this Pr, the boolean whether transformations are triggered or not is an attribute the declarative config input itself, so by definition it is already supported by the Admin API. You just need to pass `_transform` as false when sending in the data to `/config` the same way you already pass `_format_version`.

Or am I missing something? @kikito ?

hbagdi:
> @hbagdi According to this PR, the boolean whether transformations are triggered or not is an attribute the declarative config input itself, so by definition it is already supported by the Admin API. You just need to pass _transform as false when sending in the data to /config the same way you already pass _format_version.

I should have been clearer. By Admin API, I meant via the `/consumers/:username/basic-auth`  endpoint.
The problem that we need to solve is that users use decK to export the configuration but can't import basic-auth credentials back. That results in Kong re-hashing them. If there is a way to avoid re-hashing, that solves that problem.
decK currently emits a warning when someone tries to "sync" the passwords but it still trips users because this is the only field that has this behavior in all of Kong and Kong Enterprise.

PR #5842, opened 2020-04-30
docs(admin-api) add examples using arrays

Tieske:
most pressing issue: example of array usage on the command line

### DIFF
--- a/autodoc/data/admin-api.lua
+++ b/autodoc/data/admin-api.lua
@@ -78,6 +78,32 @@ return {
         config.limit=10&config.period=seconds
         ```
 
+        When specifying arrays send the values in order, or use square brackets (numbering
+        inside the brackets is optional but if provided it must be 1-indexed, and
+        consecutive). An example route added to a service named `test-service`:
+
+        ```
+        curl -i -X POST http://localhost:8001/services/test-service/routes \
+             -d "name=test-route" \
+             -d "paths[1]=/path/one" \
+             -d "paths[2]=/path/two"
+        ```
+
+        The following two are identical to the one above, but less explicit:
+        ```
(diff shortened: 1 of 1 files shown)


autodoc/data/admin-api.lua
```diff
@@ -78,6 +78,17 @@ return {
         config.limit=10&config.period=seconds
         ```
 
+        When specifying arrays use square brackets (numbering inside the brackets is optional
```
bungle, line 81: Or just send them in order
```
-d "paths=/path/one" \
-d "paths=/path/two"
```

bungle: This also applies to `multipart/form-data` but there is a bug in `multipart` library where it does not allow same parameter multiple times (it overwrites previous ones), and thus with multipart you really need to use indexes `paths[1]`, `paths[2]`.

Tieske: multi-part was removed from the docs. Hence didn't reference it here anymore either

bungle: multipart is rather useful when uploading certs.

bungle: and multipart is not removed, it has been manually added: https://github.com/Kong/docs.konghq.com/commit/a54c86139d732034fd1d1c70b0b2df395f87bff1#diff-27c036e2fa20a4521cbd01188afdc4f0

Tieske: I know, but why was it removed here then?

bungle: I think it has never been here.

bungle: I checked, and it has never been.

PR #5846, opened 2020-04-30
chore(prerelease) ability for us to release prerelease builds of Kong

hutchic:
required more flexibility / capabilities with setting package and release settings so moved a few things out of the Jenkinsfile and into our Makefile.

Any git tag that is `a.b.c-something` is assumed to be a prerelease as such it goes to the kong-prerelease bintray repository.

Some decision I made:
- our prerelease versions will follow the semver spec a.b.c-alpha|beta.n
- prereleases will only have unofficial docker images
- ~prereleases changelog is a roughdraft~
- prereleases don't get homebrew, docs, luarock, vagrant or pongo

In the spirit of incremental change I posit that we do an `alpha.1` with what's here and then reevaluate if there's a strong need for any of the release types I ommitted


Makefile
```diff
@@ -29,10 +29,52 @@ RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.
...
 KONG_BUILD_TOOLS ?= '4.3.2'
-KONG_VERSION ?= `cat $(KONG_SOURCE_LOCATION)/kong-*.rockspec | grep tag | awk '{print $$3}' | sed 's/"//g'`
 OPENRESTY_PATCHES_BRANCH ?= master
 KONG_NGINX_MODULE_BRANCH ?= master
 
+PACKAGE_TYPE ?= deb
+REPOSITORY_NAME ?= kong-${PACKAGE_TYPE}
+REPOSITORY_OS_NAME ?= ${RESTY_IMAGE_BASE}
+KONG_PACKAGE_NAME ?= kong
+KONG_VERSION ?= `cat $(KONG_SOURCE_LOCATION)/kong-*.rockspec | grep -m1 tag | awk '{print $$3}' | sed 's/"//g'`
```
hutchic, line 39: set sensible defaults but allow environment (Jenkinsfile) overrides

Makefile
```diff
@@ -29,10 +29,52 @@ RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.
...
+else
+	REPOSITORY_NAME = kong-${BRANCH}
+	REPOSITORY_OS_NAME = ${BRANCH}
+	KONG_PACKAGE_NAME = kong-${BRANCH}
+	KONG_VERSION = `date +%Y-%m-%d`
+endif
+ifneq ($(POSSIBLE_PRERELEASE_NAME),)
+	KONG_VERSION = ${BRANCH}
+	REPOSITORY_NAME = kong-prerelease
+endif
```
hutchic, line 58: the above if triggers if the tag is a prerelease tag

Makefile
```diff
@@ -29,10 +29,52 @@ RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.
...
+
+release:
+ifeq ($(ISTAG),false)
+	sed -i -e '/return string\.format/,/\"\")/c\return "$(KONG_VERSION)\"' kong/meta.lua
```
hutchic, line 62: if we're releasing a branch (presumably this is a nightly build) we need to muck with the `meta.lua`

Makefile
```diff
@@ -29,10 +29,52 @@ RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.
...
+	BRANCH = $(shell git describe --tags --abbrev=0)
+	POSSIBLE_VERSION = $(shell git describe --tags --abbrev=0 | awk -F"-" '{print $$1}')
+	POSSIBLE_PRERELEASE_NAME = $(shell git describe --tags --abbrev=0 | awk -F"-" '{print $$2}')
+else
+	REPOSITORY_NAME = kong-${BRANCH}
+	REPOSITORY_OS_NAME = ${BRANCH}
+	KONG_PACKAGE_NAME = kong-${BRANCH}
+	KONG_VERSION = `date +%Y-%m-%d`
+endif
+ifneq ($(POSSIBLE_PRERELEASE_NAME),)
```
kikito, line 55: It seems like POSSIBLE_PRERELEASE_NAME is only set on the `ifeq ($(BRANCH),HEAD)` branch above. Shouldn't we move this ifeq inside that branch?

Makefile
kikito: This seems very brittle. It means we can break the release process if we change the way we format that particular string inside meta.lua. In particular, we often use a shorthand local variable called `fmt` instead of `string.format`.

I'm not sure about what's the best solution here. One could be to just add a comment to meta: "hey don't change the following line or you might break the release". Another possibility could be to replace the whole meta.lua file by a custom one.

Makefile
hutchic: yes we should but a limitation of Make is we are not technically capable of doing this (correct me if I'm wrong)

Makefile
hutchic: would it make it better or worse if I point out we've been doing trickery like this for a while 😆 

Propose: we add the comment and try to circle back at a later date for something more elegant / less fragile?

kikito: Yeah a comment in meta.lua is fast and will solve the issue for now.

Makefile
kikito: I can't find original sources (no explicit mention on the man pages), but I am finding non-canon sources like https://stackoverflow.com/a/51518105 which seem to indicate that nesting conditionals are a thing.

hutchic: adjusted in 67e9b4f

Makefile
hutchic: Added the comment in 67e9b4f

PR #5861, opened 2020-05-06
feat(conf) add port_maps configuration option

bungle:
### Summary

Layer 4 port mapping (or port forwarding) is used a lot when running Kong inside a container. Common example is exposing ports 80 and 443 externally while running Kong inside a container with internal ports 8000 and 8443. It also allows Kong to started without elevated privileges as it is quite common that you need root-access to bind processes to TCP port < 1024.

As this translation is done in many cases without using layer 4 proxy_protocol or layer 7 X-Forwarded-Port HTTP header, the Kong has no knowledge about the target port that client originally connected to.

This PR adds `port_maps` configuration parameter:

```
#port_maps =                     # With this configuration parameter you can
                                 # let the Kong to know about the port from
                                 # which the packets are forwarded to it. This
                                 # is fairly common when running Kong in a
                                 # containerized or virtualized environment.
                                 # For example `port_maps=80:8000, 443:8443`
                                 # instructs Kong that the port 80 is mapped
                                 # to 8000 (and the port 443 to 8443), where
                                 # 8000 and 8443 are the ports the Kong is
                                 # listening to.
                                 #
                                 # This parameter helps Kong to set proper
                                 # forwarded upstream HTTP request header or to
                                 # get proper forwarded port with a Kong PDK
                                 # (in case other means determining it have
                                 # failed). It changes routing by a destination
                                 # port to route by a port from which packets
                                 # are forwarded to Kong, and similarly it
                                 # changes the default plugin log serializer to
                                 # use the port according to this mapping
                                 # instead of reporting the port Kong is
                                 # listening to.
```

This gets parsed into `kong.configuration.host_ports` table.

E.g. `KONG_PORT_MAP="80:8000,443:8443"` where the first number (`80`/`443`) before `:` is the host published port and the second one (`8000`/`8443`) is the port that Kong is listening translates to following `kong.configuration.host_ports`:

```lua
{
  [8000]   = 80,
  ["8000"] = 80,
  [8443]   = 443,
  ["8443"] = 443,
}
```

This PR also adds a new nginx context variable `ngx.ctx.host_port` which contains this translated port. So instead of using `ngx.var.server_port` the more correct one in many cases is `ngx.ctx.host_port`.

Kong PDK is also changed to take this in account when using:
```
local port = kong.request.get_forwarded_port()
```

This commit also changes the `X-Forwarded-Port` to use this. Additionally `stream routing` by `destination port` will use this too. Basic log serializer was changed to use `ngx.var.host_port` as well.

Currently it is only usable in proxy/stream, so things like `admin api` is not affected by this.


Tieske:
I think I mentioned this elsewhere, but can we add a warning in the logs when Kong is started in a container, and this option is not set?

hbagdi:
> Additionally stream routing by destination port will use this too. Basic log serializer was changed to use ngx.var.host_port as well.

I would recommend that we start small rather than going big from the beginning. The primary motivation for this feature is to correct set `x-forwarded-port` problem when upstream services depend on it. I have not heard any problems with other cases that this PR targets. Do we have feature requests or complaints to change the log serializer and stream routing itself?

I've consulted with users on stream proxy and I do know of users who map different port on different LBs to the same port on Kong and then they route based on the port on Kong. This would be a break change for such setups. These should be one-offs and rare but my point stands.

bungle:
> > Additionally stream routing by destination port will use this too. Basic log serializer was changed to use ngx.var.host_port as well.
> 
> I would recommend that we start small rather than going big from the beginning. The primary motivation for this feature is to correct set `x-forwarded-port` problem when upstream services depend on it. I have not heard any problems with other cases that this PR targets. Do we have feature requests or complaints to change the log serializer and stream routing itself?
> 
> I've consulted with users on stream proxy and I do know of users who map different port on different LBs to the same port on Kong and then they route based on the port on Kong. This would be a break change for such setups. These should be one-offs and rare but my point stands.

I don't see this being starting big. It is rather small. Nothing will be broken for anyone as nobody configures this as it is a new feature. In general ypu could also gave environment where pods bind to random port. In such case you cannot practically do any routing by destination as it is not kown before pod is launched. This is far more common scenario.

hbagdi:
> Nothing will be broken for anyone as nobody configures this as it is a new feature. You can also configure it for sone ports and not all.

Two fair points.

> In general you could also have environment where pods bind to random port. In such case you cannot practically do any routing by destination port as it is not kown before pod is launched (but the published port certainly will). This is far more common scenario.

I'm not sure if this is a common scenario. It hasn't been so in my experience.

With the points made, I think it is okay to proceed with this feature.

bungle:
> I think I mentioned this elsewhere, but can we add a warning in the logs when Kong is started in a container, and this option is not set?

This is a bit problematic.

1. Kong needs to know if it is a container
2. Kong needs to know if published port is different than listen port

Perhaps later we can add support for container APIs where we can get this auto-configured.

kong.conf.default
```diff
@@ -29,6 +29,17 @@
                                  # Each Kong process must have a separate
                                  # working directory.
 
+#port_map =                      # When running Kong behind layer 4 port mapping
```
hbagdi, line 32: `s/port mapping/proxy`

kong.conf.default
```diff
@@ -29,6 +29,17 @@
...
+                                 # (or port-forwarding), e.g. with
+                                 # `docker run -p 80:8000`, you can use this
```
hbagdi, line 34: rather than providing commands here, it would help to provide examples like docker, L4 LBs in the cloud

kong.conf.default
```diff
@@ -29,6 +29,17 @@
...
+                                 # parameter to let Kong know about published
+                                 # port when it differs from the one Kong is
```
hbagdi, line 36: published port is ambiguous because every port at kong or lb is published in some sense.

kong.conf.default
```diff
@@ -29,6 +29,17 @@
...
+                                 # listening, e.g. port_map=80:8000, 443:8443`.`
```
hbagdi, line 37: The back ticks should be removed.

kong.conf.default
```diff
@@ -29,6 +29,17 @@
...
+                                 # It has effect on `X-Forwarded-Port`, stream
```
hbagdi, line 38: We need to explain in brief what is the effect.

kong/conf_loader.lua
```diff
@@ -597,6 +598,34 @@ local function check_and_infer(conf, opts)
   -- custom validations
   ---------------------
 
+  conf.host_ports = {}
+  if conf.port_map then
+    local MIN_PORT = 1
+    local MAX_PORT = 65535
+
+    for _, ports in ipairs(conf.port_map) do
```
hbagdi, line 606: nit: `ports` is a confusing name. How about `port_map`? See the comment in kong.conf.default on the port_map name as well.

kong.conf.default
hbagdi, line 32: Another point, it would be better to call this setting `port_maps` because it can contain an array of values (csv based).

kong/conf_loader.lua
```diff
@@ -597,6 +598,34 @@ local function check_and_infer(conf, opts)
...
+
+      else
+        local host_port_str = string.sub(ports, 1, colpos - 1)
+        local host_port_num = tonumber(host_port_str, 10)
+        local kong_port_str = string.sub(ports, colpos + 1)
+        local kong_port_num = tonumber(kong_port_str, 10)
+
+        if  (host_port_num and host_port_num >= MIN_PORT and host_port_num <= MAX_PORT)
+        and (kong_port_num and kong_port_num >= MIN_PORT and kong_port_num <= MAX_PORT)
+        then
```
hbagdi, line 619: Can we add a check to ensure that kong is configured to listen on kong_port_num?
I'm not sure how much scope creep happens because of this. I'll leave that up to you, Aapo.

spec/02-integration/05-proxy/01-proxy_spec.lua
```diff
@@ -129,3 +129,81 @@ describe("#stream proxy interface listeners", function()
...
+          stream_listen = helpers.get_proxy_ip(false) .. ":19000, " ..
+                          helpers.get_proxy_ip(false) .. ":18000, " ..
+                          helpers.get_proxy_ip(false) .. ":17000",
+          port_map      = "19000:18000",
+          plugins       = "bundled,ctx-tests",
+          nginx_conf    = "spec/fixtures/custom_nginx.template",
+          proxy_listen  = "off",
+          admin_listen  = "off",
+        }))
+      end)
```
hbagdi, line 171: We should split the test into two to keep it readable for our future selves. 
With port_map enabled:
- routing based on host port works
- routing based on kong port works

And in both these cases, configure Kong to listen on a single port only.
If you are configuring kong to listen on the host_port as well, then that confuses the reader (the reader being the programmer here).

kong.conf.default
bungle: You mean layer 4 proxy? I don't think port-forwarding/-mapping can be described as proxy. Or what do you mean by this?

kong.conf.default
bungle: Published port is a term that `docker` uses for `-p`. The port that Kong listens is not published e.g. you can only access it from container. For public it is published by docker at port x. What do you suggest instead?

spec/02-integration/05-proxy/01-proxy_spec.lua
bungle: I cannot make Kong to listen single port, as it would mean that I need to have the actual port mapping in test suite. This emulates it.

kong.conf.default
bungle: I rewrote the whole thing. Not sure if it is any better.

kong.conf.default
bungle: removed.

kong.conf.default
bungle: Explaining it in brief is really hard. I tried.

kong/conf_loader.lua
bungle: changed.

kong.conf.default
bungle: Ok, I pluralized it. We btw. have `proxy_listen` not `proxy_listeners`.

hbagdi: Good point. My bad.

hbagdi:
@hishamhm I added this to `2.1.0` milestone. Please change if needed.

kong/conf_loader.lua
bungle: @hbagdi I looked at it, and it is a bit complicated (listeners are parsed later). The good side is that if the port is not the one that Kong is not listening, it will not do anything. It is just excess configuration. I will create a card about this.

kong.conf.default
```diff
@@ -125,6 +125,29 @@
                                  # This value can be set to `off`, thus disabling
                                  # the plugin server and Go plugin loading.
 
+#port_maps =                     # With this configuration parameter you can
+                                 # let the Kong to know about the port from
```
seh, line 129: s/the Kong to/Kong/

kong.conf.default
```diff
@@ -125,6 +125,29 @@
...
+                                 # which the packets are forwarded to it. This
+                                 # is fairly common when running Kong in a
+                                 # containerized or virtualized environment.
+                                 # For example `port_maps=80:8000, 443:8443`
+                                 # instructs Kong that the port 80 is mapped
```
seh, line 134: s/the port/port/

kong.conf.default
```diff
@@ -125,6 +125,29 @@
...
+                                 # to 8000 (and the port 443 to 8443), where
```
seh, line 135: s/the port/port/

kong.conf.default
```diff
@@ -125,6 +125,29 @@
...
+                                 # 8000 and 8443 are the ports the Kong is
```
seh, line 136: "the ports on which Kong is listening"

kong.conf.default
```diff
@@ -125,6 +125,29 @@
...
+                                 # listening to.
+                                 #
+                                 # This parameter helps Kong to set proper
```
seh, line 139: s/to set/set the/

PR #5864, opened 2020-05-07
chore(conf) worker_rlimit_nofile & worker_connections "auto" == max 16384

bungle:
### Summary

- Adds a test for `nginx_main_worker_rlimit_nofile=auto` and
  `nginx_events_worker_connections=auto`
- Adds documentation for `nginx_main_worker_rlimit_nofile=auto` and
  `nginx_events_worker_connections=auto`
- Renames some variables to avoid further confusion

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -653,6 +653,26 @@
 #   `nginx_admin_ssl_protocols`, both of which taking precedence over the
 #   `http {}` block.
 
+#nginx_main_worker_rlimit_nofile = auto
+                                 # Changes the limit on the maximum number of open files
+                                 # for worker processes.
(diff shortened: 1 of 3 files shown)


hishamhm:
What is the memory impact of this change for users running the default settings?

bungle:
> What is the memory impact of this change for users running the default settings?

On my machine 16k is about 44k worker process, and 64k is about 66k worker process.

hbagdi:
Based on internal discussions, didn't we all agree that this is not an issue?

bungle:
@hbagdi, yes we did, but this pr contains good tests etc. that we still want in.

PR #5865, opened 2020-05-07
fix(globalpatches) increase the limit when encoding sparse array in Kong

### LINKED ISSUE #5768: Kong db-less: Cannot serialise table: excessively sparse array
### Summary

after upgrade to kong version from 1.4.x to 2.0.x, ingress controller fail to post config, with error:
`Cannot serialise table: excessively sparse array`

### Steps To Reproduce

couldn't not reproduce on other env
only happens on kubernetes cluster with complex (but valid) ingress

### Additional Details & Logs

[ingress-kong-stgaing-9cc59f496-z2v4p] 2020/04/09 12:04:41 [error] 26#0: *948 [kong] api_helpers.lua:260 Cannot serialise table: excessively sparse array, client: 127.0.0.1, server: kong_admin, request: "POST /config?check_hash=1 HTTP/1.1", host: "localhost:8444"

dndx:
When Kong parses a declarative config, and when the declarative config
has a lot of entities and one of the high index entity has validation
errors, Kong may generate a JSON response message like this:

```
plugins = {
    [12] = {
        <schema errors>
    }
}
```

Unfortunately this causes the lua-cjson library to throw the
"Cannot serialise table: excessively sparse array" error. This PR
increases the limit of cjson so that these error won't happen with small
number of configurations.

fixes #5768

For discussion:

Is this the best way to solve it? cJSON provides another mode for dealing with excessively sparse arrays which is to convert numeric indices to hash tables. See: https://www.kyne.com.au/~mark/software/lua-cjson-manual.html#encode_sparse_array. That way we would never fail to encode the JSON object. However, once the excessively sparse threshold is reaches, the response is no longer in the format user expects and I suspect nobody will handle it correctly.

### DIFF
--- a/kong/globalpatches.lua
+++ b/kong/globalpatches.lua
@@ -15,6 +15,9 @@ return function(options)
   local meta = require "kong.meta"
 
 
+  require("cjson.safe").encode_sparse_array(nil, nil, 512)
+
+
   if options.cli then
     -- disable the _G write guard alert log introduced in OpenResty 1.15.8.1
     -- when in CLI or when running tests in resty-cli
--- a/kong/globalpatches.lua
+++ b/kong/globalpatches.lua
@@ -15,7 +15,7 @@ return function(options)
   local meta = require "kong.meta"
 
 
-  require("cjson.safe").encode_sparse_array(nil, nil, 512)
+  require("cjson.safe").encode_sparse_array(nil, nil, 2^15)
 
 
   if options.cli then
--- a/spec/02-integration/04-admin_api/15-off_spec.lua
+++ b/spec/02-integration/04-admin_api/15-off_spec.lua
@@ -499,6 +499,78 @@ describe("Admin API #off", function()
           message = "expected a declarative configuration",
         }, json)
       end)
+
+      it("sparse responses are correctly generated", function()
+        local res = assert(client:send {
+          method = "POST",
+          path = "/config",
+          body = {
+            config = [[
+            {
+              "_format_version" : "1.1",
+              "plugins": [{
+                "name": "key-auth",
+                "enabled": true,
+                "protocols": ["http", "https"]
+              }, {
+                "name": "key-auth",
+                "enabled": true,
+                "protocols": ["http", "https"]
+              }, {
+                "name": "key-auth",
+                "enabled": true,
+                "protocols": ["http", "https"]
+              }, {
+                "name": "key-auth",
+                "enabled": true,
+                "protocols": ["http", "https"]
+              }, {
+                "name": "key-auth",
+                "enabled": true,
+                "protocols": ["http", "https"]
+              }, {
+                "name": "key-auth",
+                "enabled": true,
+                "protocols": ["http", "https"]
+              }, {
+                "name": "key-auth",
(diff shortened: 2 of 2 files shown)


kong/globalpatches.lua
```diff
@@ -15,6 +15,9 @@ return function(options)
   local meta = require "kong.meta"
 
 
+  require("cjson.safe").encode_sparse_array(nil, nil, 512)
```
thibaultcha, line 18: Two notes:

1. How about specifying `encode_sparse_array(false, 0)`? The first argument `false` doesn't has any signification when `ratio = 0`, but only to explicitly disable JSON Object conversion for maintainers/readers. Then `ratio = 0` in itself ensures that all sparse arrays (no matter their fill-in ratio) will be encoded as such. If we are concerned that this could become an attack vector by feeding cjson incredibly long and sparse arrays, we could enforce a hard limit on the total length of the array (i.e. last non-nil index). Maybe something like `2^15`, just to be safe. This depends on where this instance of cjson is used.
2. This setting most likely needs to be applied to `cjson` itself as well, and maybe (depending on the usage) to other instances of cjson too, possibly with different ratio settings.

The declarative config cjson instance could be:

```lua
local dc_cjson = cjson.new()
-- false: no JSON Object conversion
-- 0: no limit on how sparse the array may be
-- 2^15: max array index
dc_cjson.encode_sparse_array(false, 0, 2^15)
```

dndx: Good advise @thibaultcha , I did some changes to make the limit 2^15 globally. On my machine, running the following case:

```shell
local cjson = require("cjson.safe")

cjson.encode_sparse_array(nil, nil, 2^15)

local a = {}

a[2^15] = 1

for i = 1, 1000 do
  local b = cjson.encode(a)
end

[datong.sun@localhost kong]$ time resty test.lua 

real	0m0.574s
user	0m0.561s
sys	0m0.005s
```

shows that serializing a sparse array as large as `2^15` should not affect the performance of the proxy by any meaningful amount to be an attractive attack vector.

We can't use `dc_cjson.encode_sparse_array(false, 0, 2^15)`, because the ratio of `0` disables the sparse array protection completely which means the `2^15` safe limit set later is never used. See the cJSON doc:

> Lua CJSON will never consider an array to be excessively sparse when ratio = 0.

Therefore, I think the risk of changing the default limit to `2^15` is minimum. Setting it separately for declarative config will be tricky because the use of PDK `response.exit` function, and is probably not necessary for the limit of `2^15`.

kikito:
> once the excessively sparse threshold is reaches, the response is no longer in the format user expects and I suspect nobody will handle it correctly.

At that point we might as well change the format of the response ourselves. For instance, if we modify our errors so instead of this:

```
plugins = {
    [12] = {
        <schema errors>
    }
}
```
They return this:
```
plugins = {
    ["12"] = {
        <schema errors>
    }
}
```

This error will never manifest. Counterpoint: even if humans can parse both syntaxes equally well, machines might find the second one a bit more troublesome.

dndx:
Merged for now, we can always revisit this change later if necessary. Thanks @thibaultcha @kikito!

PR #5867, opened 2020-05-07
chore(rockspec) add migration

darrenjennings:
Add missing migration.


Tieske:
iirc there is a test checking for this, @thibaultcha added it after some prior issues.

Why didn't it trigger on this bug? Can we fix it please?

thibaultcha:
@Tieske has a point. This is the test in question: https://github.com/Kong/kong/blob/master/spec/01-unit/02-rockspec_meta_spec.lua#L71-L83

In worth investigating how it needs to be updated to catch this case - and similar ones in the future.

PR #5874, opened 2020-05-09
refactor(plugins) rate-limiting/response-ratelimiting

WALL-E:
rate-limiting/response-ratelimiting update default policy to 'local'

Whether there is a database, 'local' can run very well.

Use db-less mode will be more and more.


hbagdi:
I agree with the rational behind this but this _could_ be classified as a breaking change. It is not "breaking" per say but is a definitely a change in behavior that can trip existing users.

Issue #5653, opened 2020-03-09
Treat key-auth's config.anonymous "" as null


mtmail:
### Summary

In the past key-auth's `config.anonymous` was set to `""` (empty string) in the database. Starting Kong 2.0 that causes `failed to load anonymous consumer` error.

### Steps To Reproduce

We upgraded from Kong 1.4 to 2.0.2. (And previously 0.9 to 1.1, then 1.1 to 1.4). Very early we added key-auth plugin like this
```
curl -s -X POST \
  --url http://localhost:8001/services/ourservice/plugins/ \
  --data 'name=key-auth'
```
without any other config values set.

In Kong 1.4 and 2.0 in the database that get stored as
```
select * from plugins where name='key-auth' and consumer_id is null;

id          | edca177b-77b6-457c-9019-9105badbeea1
name        | key-auth
api_id      |
consumer_id |
config      | {"anonymous": "", "key_names": ["apikey"],
"hide_credentials": false, "run_on_preflight": true}
```

Based on the documentation https://docs.konghq.com/hub/kong-inc/key-auth/ I wouldn've expected the database to show `"anonymous: null"` or the key not existent. https://github.com/Kong/kong/commit/91a5738dcf5 looks like the change from boolean to string type.

No problem in Kong 1.4. A URL request without any authentication key (key name `apikey` by default) set, e.g. `curl http://localhost:8000/ourservice/` returns `HTTP/1.1 401 Unauthorized` and body `{"message":"No API key found in request"}`.

Starting Kong 2.0 the same request returns `HTTP/1.1 500 Internal Server Error` with body `{"message":"An unexpected error occurred"}`. The kong logfile contains `*2400 [kong] handler.lua:193 [key-auth] failed to load anonymous consumer:failed to get from node cache: [postgres] length must be at least 1, client: 127.0.0.1, server: kong, request: "GET /ourservice HTTP/1.1", host: "localhost:8000"`

It is no longer clear what set `anonymous": ""` in the database in the past, but it seems it was the default setting. Upgrading to Kong 2.0 broke the default behaviour.

Comment https://github.com/Kong/kong/issues/3222#issuecomment-365750935 from Kong 0.12.1 shows an example how the value was set to empty string in the past.

https://github.com/Kong/kong/issues/2539 shows how Kong 0.10. set the default to empty string and how previous logic looked at both empty string and null/nil.
```
[...]

PR #5552, opened 2020-02-10
fix(plugins) accept consumer username in anonymous field

hbagdi:
Fix #5551

PR #5879, opened 2020-05-11
fix(plugins) config.anonymous = "" in auth plugins

murillopaula:
Fix https://github.com/Kong/kong/issues/5653

We could not configure an auth plugin with config.anonymous="" in CE 2.0
or EE 1.5.

It also broke users that had config.anonymous="" configured in legacy. To
reproduce, create an auth plugin with config.anonymous="" (empty string)
in 0.13, then upgrade all the way to 2.0.

This is the breaking change: https://github.com/Kong/kong/pull/5552

---

We have three options to fix it:

(a) a migration;
(b) start special-casing "" to mean null for this field in the code of
these plugins;
(c) add legacy back to those anonymous fields of string type fields,
meaning that in legacy fields "" auto-translates to null

This is the execution of (c).

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1560,7 +1560,7 @@ function Schema:process_auto_fields(data, context, nulls)
 
   for key, field in self:each_field(data) do
 
-    if field.legacy and field.uuid and data[key] == "" then
+    if field.type == "string" and field.legacy and data[key] == "" then
       data[key] = null
     end
 
--- a/kong/plugins/basic-auth/schema.lua
+++ b/kong/plugins/basic-auth/schema.lua
@@ -9,7 +9,7 @@ return {
     { config = {
         type = "record",
         fields = {
-          { anonymous = { type = "string" }, },
+          { anonymous = { type = "string", legacy = true }, },
           { hide_credentials = { type = "boolean", default = false }, },
     }, }, },
   },
--- a/kong/plugins/hmac-auth/schema.lua
+++ b/kong/plugins/hmac-auth/schema.lua
@@ -19,7 +19,7 @@ return {
         fields = {
           { hide_credentials = { type = "boolean", default = false }, },
           { clock_skew = { type = "number", default = 300, gt = 0 }, },
-          { anonymous = { type = "string" }, },
+          { anonymous = { type = "string", legacy = true }, },
           { validate_request_body = { type = "boolean", default = false }, },
           { enforce_headers = {
               type = "array",
--- a/kong/plugins/jwt/schema.lua
+++ b/kong/plugins/jwt/schema.lua
@@ -26,7 +26,7 @@ return {
                 type = "string",
                 one_of = { "exp", "nbf" },
           }, }, },
-          { anonymous = { type = "string" }, },
+          { anonymous = { type = "string", legacy = true }, },
           { run_on_preflight = { type = "boolean", default = true }, },
           { maximum_expiration = {
             type = "number",
--- a/kong/plugins/key-auth/schema.lua
+++ b/kong/plugins/key-auth/schema.lua
@@ -16,7 +16,7 @@ return {
               default = { "apikey" },
           }, },
           { hide_credentials = { type = "boolean", default = false }, },
-          { anonymous = { type = "string" }, },
+          { anonymous = { type = "string", legacy = true }, },
           { key_in_body = { type = "boolean", default = false }, },
           { run_on_preflight = { type = "boolean", default = true }, },
         },
--- a/kong/plugins/ldap-auth/schema.lua
+++ b/kong/plugins/ldap-auth/schema.lua
@@ -22,7 +22,7 @@ return {
           { hide_credentials = { type = "boolean", default = false }, },
           { timeout = { type = "number", default = 10000 }, },
           { keepalive = { type = "number", default = 60000 }, },
-          { anonymous = { type = "string" }, },
+          { anonymous = { type = "string", legacy = true }, },
           { header_type = { type = "string", default = "ldap" }, },
         },
         entity_checks = {
--- a/kong/plugins/oauth2/schema.lua
+++ b/kong/plugins/oauth2/schema.lua
@@ -30,7 +30,7 @@ return {
           { enable_password_grant = { type = "boolean", default = false, required = true }, },
           { hide_credentials = { type = "boolean", default = false, required = true }, },
           { accept_http_if_already_terminated = { type = "boolean", default = false }, },
-          { anonymous = { type = "string" }, },
+          { anonymous = { type = "string", legacy = true }, },
           { global_credentials = { type = "boolean", default = false }, },
           { auth_header_name = { type = "string", default = "authorization" }, },
           { refresh_token_ttl = { type = "number", default = 1209600, required = true }, },
--- a/spec/03-plugins/09-key-auth/02-access_spec.lua
+++ b/spec/03-plugins/09-key-auth/02-access_spec.lua
@@ -72,6 +72,10 @@ for _, strategy in helpers.each_strategy() do
         hosts = { "key-auth10.com" },
       }
 
+      local route11 = bp.routes:insert {
+        hosts = { "key-auth11.com" },
+      }
+
       bp.plugins:insert {
         name     = "key-auth",
         route = { id = route1.id },
@@ -152,6 +156,14 @@ for _, strategy in helpers.each_strategy() do
         route = { id = route10.id },
         config = {
           anonymous = anonymous_user.username,
+        },
+      }
+
+      bp.plugins:insert {
+        name     = "key-auth",
+        route = { id = route11.id },
+        config = {
+          anonymous = "",
         },
       }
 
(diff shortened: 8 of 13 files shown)


murillopaula:
This PR is essentially Kong/kong#5552 but with the legacy = true marker enabled and some tests added in the auth plugins to ensure we can configure an auth plugin with config.anonymous="".

murillopaula:
I'm wondering if it's better to just write a migration instead of always relying on the `legacy = true` field. If we don't write a migration then I guess we will always need the `legacy field`. Is that right?

bungle:
`config.anonymous=""` works fine with `multipart/form-data` and `application/x-www-form-urlencoded` it is only with `JSON` that we do not do any inferring. I think if we really want it on `JSON` should we consider to add inferring to JSON too?

murillopaula:
@bungle `config.anonymous=""` in JSON worked fine in 0.13 to 1.5. It's just that in 2.0 we can't do it. Maybe it's because of https://github.com/Kong/kong/pull/5552?

bungle:
@murillopaula, it is related to this: https://github.com/Kong/kong/pull/5231

1. We never intended to do inferring with JSON, but JSON implemented custom api that needed to figure out the subschema to run the inferring process. That code didn't skip that for JSON (whereas for every other entity we did, like core daos). 

2. #5231 added the check for JSON and thus the inferring went away even for the plugins.

The obvious question is still that what I said above.

If we want this to be done, should it be `general`. Should we start inferring plugins and not daos again? Should we infer JSON as well. And when it comes to inferring, should it be all that we have or only empty string and perhaps some other stuff (strings to numbers/booleans).

I feel like it is not a bug, it is rather re-introducing the bug. @hishamhm, your take on this?

bungle:
Yes, I wish this worked:
```
http post :8001/routes paths=/
```

Thus I usually tend to just add `-f` to every request now:
```
http post :8001/routes paths=/ -f
```

murillopaula:
@bungle Is #5231 out in 2.0 yet? I don't think it's related to it. The current problem does not happen If I revert https://github.com/Kong/kong/pull/5552/commits/9d93183d6be0cb39549da0188bedd3487c17f6df.

It's documented that we can do `config.anonymous=""` in all the auth plugins. I believe we should be able to do it in all three content types, including JSON. I don't have context on why we infer things with `multipart` and `urlencoded` but not with JSON.

In #5552 we removed the `legacy` field and I believe that is a breaking change. Users that had  `config.anonymous=""` in legacy versions and upgraded to 2.0 will hit the issue reported at #5653 

How we can solve this considering the three options in this PR's description? Correct me if I'm wrong, but if we write a migration (convert "" to null) as per option (a), we will still not be able to do `config.anonymous=""`.

I don't think it's possible to do option (b) either because the error is a validation error thrown by the schema library:

`
2020/05/11 15:14:05 [error] 4256#0: *13 [kong] handler.lua:252 [jwt] failed to load anonymous consumer:failed to get from node cache: [postgres] length must be at least 1, client: 172.29.0.1, server: kong, request: "GET /local HTTP/1.1", host: "localhost:8002"
`

Stack traceback:

```
---------+, client: 172.29.0.1, server: kong, request: "GET /local HTTP/1.1", host: "localhost:8002"
2020/05/11 16:22:26 [notice] 4543#0: *95 |[kong] init.lua:validate_field:923 "empty" "stack traceback:\n\t./kong/db/schema/init.lua:923: in function 'validate_fie|, client: 172.29.0.1, server: kong, request: "GET /local HTTP/1.1", host: "localhost:8002"
2020/05/11 16:22:26 [notice] 4543#0: *95 |ld'\n\t./kong/db/dao/init.lua:622: in function <./kong/db/dao/init.lua:615>\n\t[C]: in function 'xpcall'\n\t/usr/local/s|, client: 172.29.0.1, server: kong, request: "GET /local HTTP/1.1", host: "localhost:8002"
2020/05/11 16:22:26 [notice] 4543#0: *95 |hare/lua/5.1/resty/mlcache.lua:741: in function 'get'\n\t./kong/cache.lua:227: in function 'get'\n\t./kong/plugins/jwt/h|, client: 172.29.0.1, server: kong, request: "GET /local HTTP/1.1", host: "localhost:8002"
2020/05/11 16:22:26 [notice] 4543#0: *95 |andler.lua:246: in function <./kong/plugins/jwt/handler.lua:229>"                                                       |, client: 172.29.0.1, server: kong, request: "GET /local HTTP/1.1", host: "localhost:8002"
```

So I just did (c) in this PR as the only option left. LMK if I'm missing something or if there are other ways to fix this.

bungle:
@murillopaula the #5231 is very related:

I am referring to these:
https://github.com/Kong/kong/pull/5231/files#diff-172eb877db3ebba181fb7814865e167dR79-R82

>  I don't have context on why we infer things with multipart and urlencoded but not with JSON

I tried to give that context above. JSON has types, things like arrays, objects, numbers and boolean, and yes even null. The `application/x-www-form-urlencoded` and `multipart/form-data` has just string key and string value (same key could appear more than once).

It was decided that as `JSON` is expressive in itself, we should not try to do anything to it. Because doing stuff to input is error prone as well.

```
{
  "name": "key-auth",
  "config.key": "test"
}
```

This is already wrong. It should be:
```
{
  "name": "key-auth",
  "config": {
    "key": "test"
  }
}
```

Adding special treatment to `config.consumer=""` with `JSON` just feel plain wrong. Then people ask why I can do it for that and not for some other. It makes things inconsistent. If we document it like that with `JSON` then we could also fix docs.

In < `1.0` we had special quite handy array syntax too:
```
http put :8001/routes paths="/cat,/dog"
```

But it really caused problems too.

bungle:
So if we split this issue in two:
1) what input should be allowed (and my stand for now is that no inferring for JSON -> no changes needed for that)
2) what to do with plugin code, especially PDK that is erroring on empty string, does PDK even expect UUID or does it work with username as well?

Problem with legacy is that the field is not legacy. The schema is updated to new version and the field is not legacy anymore, and it should not be marked as such. We should get rid of `legacy` altogether. 

So out of those we need to solve 2 for now.

Some options:
1. use read transformation to translate anonymous "" to nil or null or what it needs to be.
2. on read just remove (set nil) all currently invalid fields (or set them to null)
3. add migrations
4. treat `""` as `nil`/`null` everywhere, which makes it impossible to have empty strings anymore, same goes with empty arrays.
5. start doing `inferring` with JSON (full or just some?)
6. just fix the plugins to check if the string is empty before feeding it to function that chokes because of it
7. loosen up the PDK validation on this and make it treat empty string as null.

these might have issues too (e.g. plugin that accepts "" as special value or something because the writer of plugin wanted to keep everything backward compatible and didn't want to write migrations).

bungle:
@murillopaula oh I stand corrected, the #5231 is unrelated, but this is the one I wanted to link:
https://github.com/Kong/kong/pull/5191

That is the reason for the JSON `""` not working anymore (as it never should).

murillopaula:
@bungle 

> I tried to give that context above. JSON has types, things like arrays, objects, numbers and boolean, and yes even null. The application/x-www-form-urlencoded and multipart/form-data has just string key and string value (same key can appear more than once).

Understood. Thank you for explaining.

> Adding special treatment to config.consumer="" with JSON just feel plain wrong. Then people ask why I can do it for that and not for some other. It makes things inconsistent. If we document it like that with JSON then we could also fix docs.

Ok, so assuming the docs are incorrect that `config.anonymous=""` is allowed I guess we don't need to worry about `config.anonymous=""` anymore.

murillopaula:
Empty strings are just never allowed:

```
{
  "message": "schema violation (config.anonymous: length must be at least 1)",
  "name": "schema violation",
  "fields": {
    "config": {
      "anonymous": "length must be at least 1"
    }
  },
  "code": 2
}
```

hishamhm:
> So if we split this issue in two:

Agree

>    what input should be allowed (and my stand for now is that no inferring for JSON -> no changes needed for that, perhaps only docs updates)

Strong agree, no inferring for JSON. Inferring only exists for form-encoded because that is an ill-defined format.

>    what to do with plugin code, especially PDK that is erroring on empty string, does PDK even expect UUID or does it work with username as well (should pdk be more forgiving, i guess returning nil when empty string is given is not that problematic)?

> Problem with legacy is that the field is not legacy. The schema is updated to new version and the field is not legacy anymore, and it should not be marked as such. We should get rid of legacy altogether.

Accepting `""` to mean null there is _legacy behavior_.

> So out of those we need to solve 2 for now.

Agree.

> Some options:

Thank you @bungle for the analysis.

Hard -1 on options 2, 4, 5, 7 as they relax existing type checking.

The option of marking this behavior as `legacy` should still be on the cards (let's call it option 8). Right now I tend to favor it because of one important reason: this is _legacy behavior_ that should not be promoted anymore. (Come to think of it, this is a point against options 1 and 6 because they tend to present this behavior as "acceptable", and would end up being copied by new plugins).

So, I think it should be between options 3 and 8.

I would personally go with 8 — as I said in another offline communication: "Unless you guys want to write a migration which would have to scan inside the JSON of plugin configs, but I wouldn't bother." — It's a lot more work than option 8 which is already implemented.

murillopaula:
@hishamhm I was talking with @bungle online and right now we don’t allow empty strings (see https://github.com/Kong/kong/blob/master/kong/db/schema/init.lua#L920-L923).

If we don’t allow empty fields why not translate them on read? For example, here: https://github.com/Kong/kong/blob/master/kong/db/schema/init.lua#L1616-L1623 we set to `nil` fields that are not defined in the current schema version.

So what if we do a check like:

```
    if context == "select" and field.type == "string" and data[key] == "" then
      data[key] = nil
    end
```

To fix this? So that if a string is empty in the DB we just translate it on read? This code works for me locally. The advantage is that it's less invasive than the current proposed implementation. We don't have to bring back the `legacy` field and we also don't allow empty "" strings in the schemas of plugins.

bungle:
Let's see if this is `green` or `red`:
https://github.com/Kong/kong/commit/968ebd3fce02d2ba69eed07c4703247a74a77f52

PR #5884, opened 2020-05-12
Feature/rate-limit-redis-sentinel

### LINKED ISSUE #5763: Kong rate-limtiing plugin is not setting ttl to some keys when using redis
### Summary

When tested Kong's rate-limiting plugin with redis config, found Kong is not setting TTLs for all the keys, some of the keys are having -1 ttl which is increasing the data in redis and slowing the repsonse from redis server.

Why the issue is happening:
During the increment we are checking whether the key exists in the redis or not and only setting the ttl if the key is not present in the redis.
https://github.com/Kong/kong/blob/master/kong/plugins/rate-limiting/policies/init.lua#L177

Screen Shot 2020-04-03 at 10 21 31

### Steps To Reproduce

1. Start Kong with 16 worker threads
2. Configure rate-limiting plugin redis
3. Make parallel calls ( I have tested using apache bench mark `ab -n 10000 -c 20 http://127.0.0.1:8000/testapi`

### Additional Details & Logs

- Kong version 1.2.2 ( but the plugin logic is same for 2.0.2 also)

vasuharish:
### Summary

Created Redis function to handle the scenario where the TTLs were missing to set for keys which intern caused the Redis slowness.

Redis sentinel support for better performance, all the write requests will be redirected master and all the read requests will be redirected to slaves. 

Our team has worked on performance tests on rate-limiting plugin. Please find the results below,
https://github.com/rakutentech/kong/blob/performance_test_results/results/01_rate_limiting_plugin.md

### Full changelog

* Redis function usage for incrementing counters
* Redis sentinel support for better performance

### Issues resolved

Fix #5763


javierguerragiraldez:
Hi, thanks for this PR, let me help in moving this forward.

We much prefer to keep fixes and features separate from each other.  Thus, we fixed the race window mentioned in #5763 with PR #6150, which is already merged and part of the upcoming 2.1.1 release.

The ability to use RedisSentinel for HA setups is still a welcomed feature.  Would you please rework this PR to focus on that?  We'd like it specially if preserving full compatibility with the "base" case.

In particular, that would mean that "old" configurations still work without change and no extra requirements if the user don't specify Sentinel hosts.

bungle:
I am closing this as the other part got merged already. The sentinel part needs to have a separate PR. @vasuharish can you please provide sentinel part as a different PR?

PR #5891, opened 2020-05-12
test(package) package and test Kong in CI

hutchic:
To ensure PR's can result in valid usable Kong packages run a test in CI


Makefile
```diff
@@ -28,7 +28,7 @@ RESTY_VERSION ?= `grep RESTY_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk
 RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_OPENSSL_VERSION ?= `grep RESTY_OPENSSL_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_PCRE_VERSION ?= `grep RESTY_PCRE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
-KONG_BUILD_TOOLS ?= '4.3.2'
+KONG_BUILD_TOOLS ?= 'origin/fix/next-issues'
```
hutchic, line 31: todo: merge and release this branch then update here

kong-2.0.4-0.rockspec
```diff
@@ -11,6 +11,7 @@ description = {
   license = "Apache 2.0"
 }
 dependencies = {
+  "luaossl == 20190731",
```
hutchic, line 14: todo: @bungle is fixing this after which rebase and remove this

kong/db/dao/plugins/go.lua
```diff
@@ -385,7 +385,7 @@ do
 
     if instance_info
       and instance_info.id
-      and instance_info.seq == conf.__seq__
+      and instance_info.seq == instance_info.conf.__seq__
```
gszr, line 388: I think we accidentally reverted recent go fixes. Could you drop the revert commits?

.github/workflows/build_and_test.yml
```diff
@@ -84,6 +84,34 @@ jobs:
     - name: Run Tests
       run: pushd $KONG_BUILD_TOOLS_LOCATION && make test-kong && popd
 
+  package-and-smoke-test:
```
gszr, line 87: Wondering if this should run for every commit / push / pr; it will considerably add to the overall job time.

hutchic: I could half the difference and only run it on PR jobs?

I'm going to push an empty commit to see what the timing looks like with a primed cache once I've wrapped up the necessary changes to kong-build-tools and docker-kong

Makefile
hutchic, line 31:
```suggestion
KONG_BUILD_TOOLS ?= '4.3.6'
```

.github/workflows/build_and_test.yml
hutchic: the job itself takes about 5-7 minutes but is adding negligible time to the PR as we're not over our concurrency limit. It seems like it should be possible to only run this on pull_requests but I couldn't determine the syntax ( https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions )

gszr: See https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#on

gszr: Don't we already run the functional tests on tag push, for all packages, though? Does it come as response to some issue we had last release? I'm not 100% confident I see the value of having this job run even on PRs only, but rather on tag time (also, we have daily builds, which also run the smoke tests).

hutchic: We run the functional tests as part of our daily releases as well as part of our proper releases. So rest assured there's zero change we were going to release broken builds.

This comes as a response that we had three PR's with breaking changes that all passed CI. They did fail in the daily builds but that went mostly unnoticed and creates a larger feedback loop than seeing immediately that a PR breaks something

_I can see your points and even provide additional fodder to not add this test_: Possible this is an overreaction. I've known about this blindspot in our CI but we've had almost no bugs / breaking changes that warrented a test such as this. The fact that three happened all in succession might be a statistical anomoly

hutchic: Summarizing for/against

Don't add the tests:
- the bugs wouldn't get released
- pay better attention to daily builds
- seldom have we had bugs this would have caught

Add the tests:
- the breaking change breaks on PR which is arguably the "right place"
- Total added time to PR's is negligible

Makefile
```diff
@@ -28,7 +28,7 @@ RESTY_VERSION ?= `grep RESTY_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk
 RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_OPENSSL_VERSION ?= `grep RESTY_OPENSSL_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_PCRE_VERSION ?= `grep RESTY_PCRE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
-KONG_BUILD_TOOLS ?= '4.3.2'
+KONG_BUILD_TOOLS ?= '4.3.6'
```
hutchic, line 31: moved this to a new PR https://github.com/Kong/kong/pull/5900

.github/workflows/build_and_test.yml
gszr: Wasn't aware of the occurrences. In this case, it's a +1 on my side. Could we just make sure it runs only on PR jobs?

kong-2.0.4-0.rockspec
```diff
@@ -20,7 +20,7 @@ dependencies = {
   "lua-ffi-zlib == 0.5",
   "multipart == 0.5.6",
   "version == 1.0.1",
-  "kong-lapis == 1.8.1.1",
+  "kong-lapis == 1.8.1.2",
```
gszr, line 23: Was this bump part of a different PR?

hutchic: yup @bungle handled this elsewhere. It'll go away when I rebase

.github/workflows/build_and_test.yml
hutchic: image

hutchic: handled in 93c88f1

PR #5903, opened 2020-05-14
Fixed issue with go plugin where sometimes the plugin instance is killed

primableatom:
### Summary
I found an issue on Heroku where the go plugin server would sometimes not be called.
### Full changelog
On debugging I found `bridge_loop` method would sometimes return an error which is NOT handled specifically the `No plugin instance` error.
### Issues resolved
This PR resolves the issue by handling that error. So when the error happens we just
1. Reset the plugin instance
2. Get the plugin instance
3. Invoke `bridge_loop` again

PS: I tested the fix on Heroku and it seems to have fixed the issue (so far). Please look into it and give feedback


jeremiegourovitch:
Hey there! 

Seems that i have the same issue.. Just patched my Kong with your fix, w'll see if it's happen anymore. 

I'll comment above in the next days to update my status. 

Thx.

gszr:
@primableatom @jeremiegourovitch Any news on this? Has the issue been resolved with the patch?

jeremiegourovitch:
Hello, 

No more plugin "untriggered" after merging this ticket with my kong dev version. It appears that it solves the issue. 

I tried with 20k+ requests with JMeter and all have been executed with the plugin. 

Thx to primableatom.

primableatom:
Thanks for verifying @jeremiegourovitch. As I said I tested it already with [k6.io](https://k6.io/) and it works fine for me. Anything else we need to verify @gszr ?

gszr:
@primableatom @jeremiegourovitch, could you give the following patch a try in your environments (after reverting this PR's)? We think it may fix the issue, but since we are unable to reproduce it, we can't test.

```
diff --git a/kong/db/dao/plugins/go.lua b/kong/db/dao/plugins/go.lua
index e1654788f..f8fcf4229 100644
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -365,16 +365,14 @@ do
 
   function get_instance(plugin_name, conf)
     local key = type(conf) == "table" and conf.__key__ or plugin_name
-    local instance_info = instances[key]
 
-    while instance_info and not instance_info.id do
+    while instances[key] and not instances[key].id do
       -- some other thread is already starting an instance
       ngx.sleep(0)
-      if not instances[key] then
-        break
-      end
     end
 
+    local instance_info = instances[key]
+
     if instance_info
       and instance_info.id
       and instance_info.seq == conf.__seq__

```

Thanks!

primableatom:
@gszr , I am in a regression test phase so can't really play around with the instance till this weekend. @jeremiegourovitch can you test it please ?

jeremiegourovitch:
Hey! 

Sure, i'll test this fix today. 

Stay tuned :)

gszr:
Thank you very much, @primableatom and @jeremiegourovitch!

djepoulet:
Just added this fix to my test platform, w'll see if kong plugin is getting stuck in few hours.

gszr:
Hi @djepoulet, any findings? Thanks!

gszr:
Hey @djepoulet, it'd be great to hear your findings. We will be soon cutting 2.1.0-beta1 and would love to include this!

djepoulet:
Hey, 

It seems not as stable as the fix provided by primableatom. Kong plugin has crashed/freezed on the past week.

The first one seems to be a better approach. 

Hope this help.

gszr:
Thanks for the feedback @djepoulet!

gszr:
On a related note, @djepoulet, would you be able to create a minimal config that reproduces the issue and describe your environment so we can try once again to pinpoint the issue? Our attempts so far weren't very productive. In the meantime, we will look into including the original fix into the release. Thanks again!

djepoulet:
I'll try to find the time to create a package to reproduce the bug.

PR #5911, opened 2020-05-15
fix(declarative) when loading declarative config, remove null values

### LINKED ISSUE #5834: bug: defining nulls in declarative config throws errors
### Summary

Providing a full-declarative YAML/JSON file with `null` throws an error.

### Steps To Reproduce

1. Startup Kong in DB-less mode
2. `http :8001/config config=@kong.json` (kong.json is [here](https://gist.github.com/hbagdi/30fd0f2531f1a3e4835be9952d376287)).
3. Kong throws back a 400.

### Additional Details & Logs

- Kong version (`$ kong version`) 2.0.4

The expected behavior is that Kong should parse these correctly
The behavior in Kong has gone back and forth so fixing it for good will help.

The reason we should fix this is because the same plugin configuration will be accepted by Admin API:

```
cat plugin.json | http :8001/plugins
```
plugin.json is [here](https://gist.github.com/hbagdi/fb48be335d608ab38cf635400a328a24).

Furthermore, Kong's Admin API also returns back all the `nulls` defined as well:
```
http :8001/plugins
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 501
Date: Tue, 28 Apr 2020 16:47:35 GMT
Server: kong/2.0.4
X-Kong-Admin-Latency: 105

{
    "data": [
        {
            "config": {
                "day": null,
                "fault_tolerant": true,
                "hide_client_headers": false,
                "hour": null,
                "limit_by": "consumer",
                "minute": 10,
                "month": null,
                "policy": "redis",
                "redis_database": 0,
                "redis_host": "redis.common.svc",
                "redis_password": null,
                "redis_port": 6379,
                "redis_timeout": 2000,
                "second": null,
                "year": null
            },
            "consumer": null,
            "created_at": 1588092451,
            "enabled": true,
            "id": "baca4cce-b2bf-432d-a8db-1ce569d36edd",
            "name": "rate-limiting",
            "protocols": [
                "http",
                "https"
            ],
            "route": null,
            "service": null,
            "tags": null
        }
    ],
    "next": null
}
```

This has caused issues for enterprise and community users as they are trying to adopt the DB-less mode of Kong or cases when they are using both DB-less and DB mode.

dndx:
properly

When declarative config with `null` value is passed in YAML or JSON, the
lyaml parser will return it's own variant of `null` representation that
is different from cJSON. Causing schema validations to fail with type
error.

This commit adds `lyaml.null` support to the `remove_nulls` helper to
make sure they are always properly removed from the input.

This commit also removes the use of `cjson` when parsing JSON input and
instead always use the lyaml library. Because when user specifies a
string payload inside the `config` parameter when POSTing to the
`/config` admin API endpoint, the lyaml parser has always been used.
Note the [YAML Specification](https://yaml.org/spec/1.2/spec.html)
specifically calls "The primary objective of this revision is to bring
YAML into compliance with JSON as an official subset."

Because JSON is a subset of YAML and libyaml supports parsing JSON out
of the box, we can use lyaml for both YAML and JSON config when parsing
declarative config. This avoids different code path when loading JSON
encoded declarative config using the CLI and using the admin API to make
sure they always present the same result and fail with the same messages
for the user.

fixes #5834

### DIFF
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -14,8 +14,7 @@ local tostring = tostring
 -- Do not accept Lua configurations from the Admin API
 -- because it is Turing-complete.
 local accept = {
-  yaml = true,
-  json = true,
+  yaml_and_json = true,
 }
 
 local _reports = {
--- a/kong/cmd/config.lua
+++ b/kong/cmd/config.lua
@@ -12,8 +12,7 @@ local DEFAULT_FILE = "./kong.yml"
 
 
 local accepted_formats = {
-  yaml = true,
-  json = true,
+  yaml_and_json = true,
   lua = true,
 }
 
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -18,6 +18,7 @@ local next         = next
 local log          = ngx.log
 local fmt          = string.format
 local match        = string.match
+local remove_nulls = utils.remove_nulls
 
 
 local ERR          = ngx.ERR
@@ -36,18 +37,6 @@ end
 local _M    = {}
 local DAO   = {}
 DAO.__index = DAO
-
-
-local function remove_nulls(tbl)
-  for k,v in pairs(tbl) do
-    if v == null then
-      tbl[k] = nil
-    elseif type(v) == "table" then
-      tbl[k] = remove_nulls(v)
-    end
-  end
-  return tbl
-end
 
 
 local function validate_size_type(size)
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -4,13 +4,15 @@ local pl_file = require "pl.file"
 local lyaml = require "lyaml"
 local cjson = require "cjson.safe"
 local tablex = require "pl.tablex"
+local stringx = require "pl.stringx"
+local utils = require "kong.tools.utils"
 
 
 local deepcopy = tablex.deepcopy
-local null = ngx.null
 local SHADOW = true
 local md5 = ngx.md5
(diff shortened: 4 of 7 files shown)


bungle:
> This commit adds lyaml.null support to the remove_nulls helper to make sure
> they are always properly removed from the input.

@dndx, 

I though #5834 was about `remove_nulls` being too aggressive, e.g. it does not allow specifying `null` when user really meant it (it removes it instead). This removal may lead for possible `default` to be picked up instead.

dndx:
@bungle I'm not sure if that is the case, because the sample @hbagdi gave inside #5834 throws error and refuse to load as valid declarative config.

@hbagdi can you clarify?

bungle:
> @bungle I'm not sure if that is the case, because the sample @hbagdi gave inside #5834 throws error and refuse to load as valid declarative config.
> 
> @hbagdi can you clarify?

That could be, what I mean, perhaps that didn't load because it was translated to `yaml.null` which errored our code?

hbagdi:
Both of you are correct.
If a property can be `null` in the Admin API representation, that should be allowed in declarative configuration format.
If a property is explicitly specified as `null`, then it should not be populated with default values.

hbagdi:
@hishamhm I added 2.0.5 milestone to this one. Please update as you see fit. Thanks!

kikito:
We're closing this after a meeting with the EE team. We'll split this into two separate PRs:
* One which replaces the CJSON parser with libyaml
* A separate one which solves the null problem in a different way than on this PR

PR #5914, opened 2020-05-16
[proposal] perf(basic-auth-v2) introduces basic-auth-v2 plugin with better performance

Abhishekvrshny:
### Summary

The original `basic-auth` plugin has `id` as the primary key, with a
secondary index on `username`. While this works well with postgres,
there are performance issues with cassandra as lookups always happen
on `username` and secondary indexes on cassandra are inefficient.

A new plugin is introduced as changing the primary key in cassandra
is non-trivial and hence updating the existing plugin is not possible.

This plugin makes `username` as the partition key and `id` as the
clustering in cassandra. In case of postgres, both form composite
primary key.

Signed-off-by: Abhishek Varshney <abhishek.varshney@razorpay.com>

### Issues resolved

https://discuss.konghq.com/t/primary-key-for-cassandra-in-basic-auth-plugin/6169

### DIFF
--- a/kong-2.0.4-0.rockspec
+++ b/kong-2.0.4-0.rockspec
@@ -219,6 +219,14 @@ build = {
     ["kong.plugins.basic-auth.schema"] = "kong/plugins/basic-auth/schema.lua",
     ["kong.plugins.basic-auth.daos"] = "kong/plugins/basic-auth/daos.lua",
 
+    ["kong.plugins.basic-auth-v2.migrations"] = "kong/plugins/basic-auth-v2/migrations/init.lua",
+    ["kong.plugins.basic-auth-v2.migrations.000_base_basic_auth_v2"] = "kong/plugins/basic-auth-v2/migrations/000_base_basic_auth_v2.lua",
+    ["kong.plugins.basic-auth-v2.crypto"] = "kong/plugins/basic-auth-v2/crypto.lua",
+    ["kong.plugins.basic-auth-v2.handler"] = "kong/plugins/basic-auth-v2/handler.lua",
+    ["kong.plugins.basic-auth-v2.access"] = "kong/plugins/basic-auth-v2/access.lua",
+    ["kong.plugins.basic-auth-v2.schema"] = "kong/plugins/basic-auth-v2/schema.lua",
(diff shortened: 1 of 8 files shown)


kikito:
Hello, thanks for your contribution.

> changing the primary key in cassandra
is non-trivial and hence updating the existing plugin is not possible

While I agree in that it is not-trivial, I think updating the existing plugin is perfectly possible. Changes similar to the one you are mentioning on this PR have already been done to Kong entities in the past. This kind of transition is done in the migration segments. Recently, for example, we had to add a new core entity for several tables. This meant changing the pagination of several key entities. You can see the relevant functions here:

https://github.com/Kong/kong/blob/8ee1db2811fd04c992b17e83ea073f954c90fd71/kong/db/migrations/operations/200_to_210.lua

They were used to re-partition several tables in the corresponding migration:

https://github.com/Kong/kong/blob/8ee1db2811fd04c992b17e83ea073f954c90fd71/kong/db/migrations/core/009_200_to_210.lua

It stands to reason that a similar approach could be taken here. It is not a trivial one, for sure, but it is doable. I think in the long run it would be less work than maintaining two separate basic-auth plugins.

hbagdi:
Cassandra is deprecated as a supported data-store. Furthermore, Hybrid mode is our recommended mode of operation which doesn't have this problem.
Our guidance is to have a custom plugin to get around these limitations.
Closing this PR for now. Please re-open if needed.

gabybeitler:
Stop this email  chain please
This is external company


Mayank Gupta<https://www.linkedin.com/in/mayankgupta/>|Director Product Marketing, Nutanix Cloud Manager | Nutanix, San Jose, CA



From: Harry ***@***.***>
Date: Tuesday, October 25, 2022 at 2:24 PM
To: Kong/kong ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [Kong/kong] [proposal] perf(basic-auth-v2) introduces basic-auth-v2 plugin with better performance (#5914)

Cassandra is deprecated as a supported data-store. Furthermore, Hybrid mode is our recommended mode of operation which doesn't have this problem.
Our guidance is to have a custom plugin to get around these limitations.
Closing this PR for now. Please re-open if needed.

PR #5929, opened 2020-05-20
fix(declarative) misleading errors and validation errors

### LINKED ISSUE #5750: Bug: configuration parsing throws wrong errors on missing foreign key relations
### Summary

When a foreign relation is missing, the configuration parser throws misleading errors.

### Steps To Reproduce

Run Kong 2.0.2 with DB-less mode

```bash
$ cat kong.json
{
  "_format_version": "1.1",
  "consumers": [
    {
      "username": "consumer-test",
      "basicauth_credentials": [
        {
          "username": "some_user",
          "password": "some_password"
        }
      ]
    }
  ],
  "plugins": [
    {
      "name": "request-transformer",
      "config": {
        "add": {
          "body": [],
          "headers": [],
          "querystring": []
        },
        "append": {
          "body": [],
          "headers": [
            "X-Another-Header:foo"
          ],
          "querystring": []
        },
        "remove": {
          "body": [],
          "headers": [
            "X-Another-Header"
          ],
          "querystring": []
        },
        "rename": {
          "body": [],
          "headers": [],
          "querystring": []
        },
        "replace": {
          "body": [],
          "headers": [],
          "querystring": [],
          "uri": null
        }
      },
      "route": "does-not-exist",
      "enabled": true,
      "protocols": [
        "http",
        "https"
      ]
    }
  ]
}
$ http :8001/config < kong.json
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 362
Date: Fri, 03 Apr 2020 00:03:11 GMT
Server: kong/2.0.1
X-Kong-Admin-Latency: 275

{
    "code": 14,
    "fields": {
        "consumers": [
            {
                "basicauth_credentials": [
                    {
                        "@entity": [
                            "all or none of these fields must be set: 'password', 'consumer.id'"
                        ]
                    }
                ]
            }
        ]
    },
    "message": "declarative config is invalid: {consumers={{basicauth_credentials={{[\"@entity\"]={\"all or none of these fields must be set: 'password', 'consumer.id'\"}}}}}}",
}
```

If you remove the broken route relation on the above plugin, then the configuration is correctly parsed.

### Additional Details & Logs

- Kong version (`$ kong version`) 2.0.2

### LINKED ISSUE #5920: Declarative config basic auth credentials on a consumer errors when key auth is defined on a route without a name or id
### Summary and steps to reproduce

This declarative config errors:

```
_format_version: "1.1"

services:
- name: test-service
  routes:
  - paths:
    - /test/path
    plugins:
    - name: key-auth
  url: https://example.com

consumers:
- username: test-user
  basicauth_credentials:
  - username: test-username
    password: test-password
```

```
$ kong config parse test-kong.yaml
Error: Failed parsing:
in 'consumers':
  - in entry 1 of 'consumers':
    in 'basicauth_credentials':
      - in entry 1 of 'basicauth_credentials':
        in '@entity':
          - in entry 1 of '@entity': all or none of these fields must be set: 'password', 'consumer.id'

```

Doing one of the following make the error not occur:
* Remove the key-auth plugin on the route
* Remove the basicauth_credentials from the consumer
* Add a name to the route
* Add an ID to the route

For example this config works fine:

```
_format_version: "1.1"

services:
- name: test-service
  routes:
  - name: test-route
    paths:
    - /test/path
    plugins:
    - name: key-auth
  url: https://example.com

consumers:
- username: test-user
  basicauth_credentials:
  - username: test-username
    password: test-password
```

### Additional Details & Logs

- Kong version: 2.0.4 (installed via RPM)
- Operating system: RHEL 7 x64

Logs and config as per above.

The error message seems buggy and related to #5750, but there seems to be an unrelated issue in that I believe this config should be valid in the first place.

bungle:
#### fix(declarative) misleading error may be returned in certain cases

The declarative config validation may return misleading error messages in some cases, see #5750.

#### fix(declarative) validation errors because non-filled primary keys

The validation of declarative configuration could fail when presenting configuration that did not have ids or names or didn't have a cache key.

Here is such configuration:

```yaml
_format_version: "1.1"

services:
- name: test-service
  routes:
  - paths:
    - /test/path
    plugins:
    - name: key-auth
  url: https://example.com

consumers:
- username: test-user
  basicauth_credentials:
  - username: test-username
    password: test-password
```

See #5920

### Issues resolved

Fix #5750
Fix #5920

### DIFF
--- a/kong/db/schema/others/declarative_config.lua
+++ b/kong/db/schema/others/declarative_config.lua
@@ -449,6 +449,8 @@ local function get_key_for_uuid_gen(entity, item, schema, parent_fk, child_key)
   if schema.cache_key then
     return pk_name, build_cache_key(entity, item, schema, parent_fk, child_key)
   end
+
+  return pk_name
 end
 
 
@@ -486,7 +488,7 @@ local function generate_ids(input, known_entities, parent_entity)
 end
 
 
-local function populate_ids(input, known_entities, parent_entity, by_id, by_key)
+local function populate_ids_for_validation(input, known_entities, parent_entity, by_id, by_key)
   local by_id  = by_id  or {}
   local by_key = by_key or {}
   for _, entity in ipairs(known_entities) do
@@ -509,11 +511,15 @@ local function populate_ids(input, known_entities, parent_entity, by_id, by_key)
     for _, item in ipairs(input[entity]) do
       local pk_name, key = get_key_for_uuid_gen(entity, item, schema,
                                                 parent_fk, child_key)
-      if key and not item[pk_name] then
-        item[pk_name] = generate_uuid(schema.name, key)
+      if pk_name and not item[pk_name] then
+        if key then
+          item[pk_name] = generate_uuid(schema.name, key)
+        else
+          item[pk_name] = utils.uuid()
+        end
       end
 
-      populate_ids(item, known_entities, entity, by_id, by_key)
+      populate_ids_for_validation(item, known_entities, entity, by_id, by_key)
 
       local item_id = DeclarativeConfig.pk_string(schema, item)
       by_id[entity] = by_id[entity] or {}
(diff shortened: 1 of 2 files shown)


kong/db/schema/others/declarative_config.lua
```diff
@@ -554,6 +560,29 @@ local function populate_ids(input, known_entities, parent_entity, by_id, by_key)
 end
 
 
+local function remove_errors(err)
```
kikito, line 563: suggest rename: `extract_null_errors`

Issue #5589, opened 2020-02-20
Errors logged for health check setAddressStatus failure


chris-branch:
I see a large number of error messages in the error_log that look like this:

`2020/02/20 18:02:44 [error] 315#0: *107648547 [lua] balancer.lua:294: [healthchecks] failed setting peer status: no peer found by name '<redacted>' and address 10.2.220.4:443, context: ngx.timer`

Kong is able to serve traffic to these hosts perfectly fine, so there doesn't seem to be any negative impact from this other than cluttering up of the error_log.  The DNS for the host name in question is configured with a TTL of 30 seconds, and it rotates the IP address every 30 seconds (returns a single A name record).  So I think what is happening is that Kong is doing a DNS requery every minute and getting a new IP address.  During the health check callback, it calls `setAddressStatus`:

https://github.com/Kong/kong/blob/master/kong/runloop/balancer.lua#L286

`setAddressStatus` attempts to find an existing entry that matches the host, port, and IP address.  The host and port match, but the IP address has changed, so it logs an error because the lookup failed due to the new IP address.

My question is: Is this really an error, or is it safe to ignore the setAddressStatus failure in this particular scenario?  If the latter, could the code be changed to avoid logging an error message for this scenario?

By the way, I don't have health checks explicitly configured for this upstream.  This health check is the built-in one that Kong adds automatically.

Kong version: 1.4.2

wanyuetian:
I face the same problem with Kong v1.4.2

agaffney:
I'm seeing this problem with both 1.5.1 and 2.0.2, when using a S3 endpoint for the upstream. The DNS record for the S3 bucket returns a different single IP on each query.

agaffney:
I cannot seem to reproduce this in an isolated environment :(

agaffney:
This only seems to occur when using multiple workers. I deployed Kong with the helm chart and set `env.nginx_worker_processes=4` and then deployed the following manifest, and the errors popped up almost immediately.

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: kong-peer-status-test
spec:
  externalName: foo.s3-website-us-east-1.amazonaws.com
  type: ExternalName
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: kong
  name: kong-peer-status-test
spec:
  rules:
  - host: foo.example.com
    http:
      paths:
      - backend:
          serviceName: kong-peer-status-test
[...]

PR #5932, opened 2020-05-20
chore(balancer) update ERR to WARN for missing peers in the balancer

Tieske:
see #5589, mitigates the messages, but final resolution should probably
be based on better tracking of targets/addresses.


Tieske:
we might even want to set this to "debug"

PR #5939, opened 2020-05-22
feat(plugins) add grpc-gateway plugin

fffonion:
### Summary

Add the [grpc-gateway plugin](https://github.com/Kong/kong-plugin-grpc-gateway) and enable grpc plugins by default (following the pattern as #5555).

### Full changelog

* Add the grpc-gateway plugin to rockspec
* Enable grpc-web and grpc-gateway plugin by default

### DIFF
--- a/kong-2.0.4-0.rockspec
+++ b/kong-2.0.4-0.rockspec
@@ -50,6 +50,7 @@ dependencies = {
   "kong-plugin-aws-lambda ~> 3.4",
   "kong-plugin-acme ~> 0.2",
   "kong-plugin-grpc-web ~> 0.1",
+  "kong-plugin-grpc-gateway ~> 0.1",
 }
 build = {
   type = "builtin",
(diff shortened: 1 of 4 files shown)


fffonion:
@javierguerragiraldez we may need to bump the lua-protobuf of grpc-web plugin to 0.3.2 since the feature that grpc-gateway needed to parse complex options is added in 0.3.2. If it looks okay to you i can send a PR.

javierguerragiraldez:
Sure it's ok. I didn't realize it was pinned to a specific version. It could probably be just >=0.3.1, while grpc-gateway is >=0.3.2. But I'm not sure that would do the right thing, so just bumping everything to the latest is ok for me

fffonion:
@javierguerragiraldez 👍 I pinned it to the 0.3 minor version https://github.com/Kong/kong-plugin-grpc-web/pull/3

fffonion:
Link to docs PR: https://github.com/Kong/docs.konghq.com/pull/2074

PR #5940, opened 2020-05-22
feat(pdk) new PDK function `kong.log.override_log_serializer_field` to

dndx:
allow plugins safely overriding basic log serializer fields

Fields such as `client_verify` needs to be overwrite by plugins sometimes.
This PR provides a safe interface for doing so in an hope to avoid
hacks.


bungle:
About that `note`. How about setting `ngx.ctx.ssl_client_verify` and still fallback to `ngx.var.ssl_client_verify` in case context value is missing?

`kong.log` in my opinion has nothing to do with `basic-serializer` that is just a file in plugins folder that many (not all)  the logging plugins use to format the log message. `kong.log` on the other hand is just about writing lines to Nginx error log.

I would probably like to see e.g. configurable `log-serializer` rather than adding PDK function to modify `basic serializer`. Perhaps that infrastructure could also filter the input data before giving it to formatter.

dndx:
@bungle that was my original thought, although it is an ugly hack that will show a bad example to plugin authors:

Setting prearranged `ctx` value for intra plugin communication is not an advised use pattern and shows bad example for other plugin authors to follow.

I don't mind moving the function elsewhere, even inside the `log-serializer` package.

I don't think this can be solved within the existing plugin config framework, because log serializer is built into Kong and user naturally expects it to work with other plugins in meaningful ways without explicit configuration.

PR #5946, opened 2020-05-25
fix(ci) adjust plugin tests for some plugin naming inconsistencies

hutchic:
some bundled kong plugins prefix their tags with `v`


fffonion:
👍 I'll make the grpc-gateway tag without v since it's not bundled at present.

PR #5948, opened 2020-05-25
fix(declarative) convert lyaml.nulls to ngx.nulls

bungle:
### Summary

Contains a fix that converts `lyaml.null`s to `ngx.null` when parsing declarative configuration.

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -8,6 +8,7 @@ local tablex = require "pl.tablex"
 
 local deepcopy = tablex.deepcopy
 local null = ngx.null
+local null_yaml = lyaml.null
 local SHADOW = true
 local md5 = ngx.md5
 local ngx_socket_tcp = ngx.socket.tcp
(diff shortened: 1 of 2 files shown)


kong/db/declarative/init.lua
```diff
@@ -92,17 +105,16 @@ function Config:parse_string(contents, filename, accept, old_hash)
...
+  then
     local pok
     pok, dc_table, err = pcall(lyaml.load, contents)
     if not pok then
       err = dc_table
       dc_table = nil
     end
 
-  elseif accept.json and filename:match("json$") then
-    dc_table, err = cjson.decode(contents)
```
hishamhm, line 104: This is going to cause `.json` input files to contain arbitrary YAML, which is a much larger format. The `/config` endpoint doesn't do a very good job in making a distinction, but `db_import` does... I would keep this around, but not a blocker

hishamhm: (also, cjson is probably faster in case this ever matters)

bungle: Ok, let's remove it for now. Removed. :-).

PR #5949, opened 2020-05-25
chore(pdk) bump version to 1.4.0

gszr:
As commit da707f0f9 introduced a new PDK submodule, minor version needs
a bump.

### DIFF
--- a/kong/pdk/init.lua
+++ b/kong/pdk/init.lua
@@ -211,7 +211,7 @@ assert(package.loaded["resty.core"])
 
 local MAJOR_VERSIONS = {
   [1] = {
-    version = "1.3.1",
+    version = "1.4.0",
     modules = {
       "table",
       "node",


kong/pdk/init.lua
```diff
@@ -211,7 +211,7 @@ assert(package.loaded["resty.core"])
 
 local MAJOR_VERSIONS = {
   [1] = {
-    version = "1.3.1",
+    version = "1.4.0",
```
gszr, line 214: Unsure whether the bump should be in-place in this entry, or if a new entry should be added for 1.4.0, listing the recently-introduced submodule. @hishamhm thoughts? Also, the submodule is not explicitly listed in here, but rather conditionally added [below](https://github.com/Kong/kong/pull/5949/files#diff-948486a6191b8ed68d5aa1c5d27315beR235).

PR #5951, opened 2020-05-26
feat(plugins) set generic X-Authentication-Type

### LINKED ISSUE #5933: [Feat]Support knowing the credential type used for auth for logging purposes and to upstream API.
Was glancing at this PR: https://github.com/Kong/kong/pull/5516 and it got me thinking of another easy win feature to implement that may yield some decent return.

One thing downstream + logs do not collect was the auth pattern that was leveraged on a specific proxy transaction. 

I propose an ```X-Authentication-Type``` header for upstream that all auth plugins set. Values would include:
```
oauth2
jwt
ldap
basic-auth
hmac

and so forth.
``` 

This idea spawns from potential analytics that can be gathered on what your most popular authentication traffic is during logging and any analysis that might be gleamed on which authentication types have the most latency or error rates etc. for further debugging.

Maybe set + get with the pdk and used in the standard Kong logs as well in basic serializer? The value can be derived from the plugins name from the conf honestly.  
```
{
"auth": "oauth2"

or

"auth_type": "oauth2"

}
````

Thoughts?

### PR DESCRIPTION
**Pull request by jeremyjpj0916**
# Summary

Talking to my colleague the other week we thought it would be cool to enable upstreams the ability to know how a user authenticated through the Kong gateway, as well as to provide our logging analytics with what auth patterns were most popular among our users active traffic(maybe one auth type proves slower or has more problems then another as well and can now be made visible?). It also can enable knowing which auth type validated on a proxy with multiple authentication plugins applied using the anonymous user pattern. 

In the event no credentials were authenticated and we are left with an anonymous user this field will not be set.

Similar PR in design to #5516 

### Full changelog

* Implement: ```X-Authentication-Type``` header for upstream requests
* Tests: 
/spec/03-plugins/..
```09-key-auth/02-access_spec.lua```
```10-basic-auth/03-access_spec.lua```
```16-jwt/03-access_spec.lua```
```19-hmac-auth/03-access_spec.lua```
```20-ldap-auth/01-access_spec.lua```
```25-oauth2/03-access_spec.lua```
                                             

### Issues resolved

Fix #5933 

P.S: Food for thought:

1. Auth plugins inconsistently named. Most have ```xyz-auth``` as name(4 ```hmac-auth,key-auth,ldap-auth,basic-auth```), some are just ```xyz``` (2, ```oauth2,jwt```),
Personally my preference would be to drop the auth and leave just the type it is since anyone who understands the context of them knows its regarding authentication. Changing the name of the plugin at this point might be a breaking change Kong does not care to make though. Maybe one day someone with OCD will take care of it :) .

2. Also the ```set_consumer``` method could likely be spun out to the pdk, and there is some code in there thats not needed:
```
  if consumer and consumer.id then
    set_header(constants.HEADERS.CONSUMER_ID, consumer.id)
  else
    clear_header(constants.HEADERS.CONSUMER_ID)
  end
```

Since a consumer will always have an id(not an optional field), these 5 lines could just be turned into:
```
set_header(constants.HEADERS.CONSUMER_ID, consumer.id)
```

The other two consumer fields are semi-optional so its reasonable to if/else check them as the method does.


jeremyjpj0916:
Rip, looks like next branch cicd is currently rekt. Will let the Kong squad figure things out and maybe kick off a build of this here when they get time!

bungle:
@jeremyjpj0916 if you haven't, can you try to rebase to latest `next`.

jeremyjpj0916:
@bungle it is on the latest next right? : https://github.com/jeremyjpj0916/kong/commits/next .

bungle:
@jeremyjpj0916 ok let's figure it out: @hutchic do you have any idea about it?

jeremyjpj0916:
Looking at another recent PR against next that succeeded vs azure login failure PR mine:

Success:

image


Failure:

image

jeremyjpj0916:
@bungle @hutchic I pulled in Kong's latest next changes here(idk if thats really the right thing to do but it should kick off a new CI/CD build) hopefully your pipeline is fixed for outside contributors at this point.

jeremyjpj0916:
Hmm I don't think this feat was just for the key-auth plugin, was more a generic way to do them all.

gszr:
Hi @jeremyjpj0916, 

I have changed the base branch here to `master`, as we are switching to a more `master`-based branching strategy.

jeremyjpj0916:
lol sure, this thing is ancient now 😆  .

hbagdi:

jeremyjpj0916:
PR should work but I just never have been versed at rebasing and kicking off cicd again here. Could remake with basically the idential code and should work.

jeremyjpj0916:
Thinking to rework this PR again on a new PR from scratch. I wonder if there is a better impl and maybe one that also sets a ngx.ctx.var that the logging like http log etc. can all use too to start tracking the auth type used at the kong layer by plugin name.

PR #5955, opened 2020-05-26
refactor(db-less) change the `flatten` return types

kikito:
This PR modifies some of the changes introduced by #5835.

In particular, the `flatten` function now returns two separate
tables, `entities` and `meta`, instead of a single one.

This is done in order to is simplify a future merge with Kong EE.

The `_transform` field is also removed from the reports as it isn't
needed.

### DIFF
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -72,10 +72,9 @@ return {
       end
       self.params.check_hash = nil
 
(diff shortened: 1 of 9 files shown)


kikito:
@hishamhm please re-review at your convenience, it should be green now

hishamhm:
@kikito looking good! merged, thanks!

PR #5965, opened 2020-05-28
chore(ci) github actions adjustment

hutchic:
let docker login command fail silently. This will fix folks that fork and then PR to `kong/kong` from getting failures

only run github actions on specific branches, all tags and all pull requests

Should resolve the issues experienced in https://github.com/Kong/kong/pull/5951


crackersophia:
I met the similar problem, but the solution #5965 does not seem to be merged yet.

PR #3028, opened 2017-11-14
[WIP] Adds a switch for udp and tcp loggers to log request/response bodies

dv00d00:
### Summary

This PR adds a possibility to log http body of incoming request and outgoing response to the existing udp and tcp logger. This work is largely based on runscope handler.

### Full changelog

* [Updated basic log serializer to check for request/response bodies and includes them in log body]
* [Updated tcp-log handler and schema to conditionally include request/response bodies]
* [Updated udp-log handler and schema to conditionally include request/response bodies]

As a side note this is my first PR ever, would like to hear if there a chance such change would be accepted in principle. Would also like to hear if other logging plugins should have an option to log req/resp body

PR #5968, opened 2020-05-29
feat(tcp-log) add a switch to log request/response body

crackersophia:
### Summary

This PR adds an optional switch to log the request and response body for tcp-log plugin.

### Full changelog

*  Add request/response body variables in pdk function `kong.log.serialize`
*  Get request/response body during access and body_filter phase respectively and store them in kong plugin storages
*  Add related tests

### Issues resolved
Similar #3028 (the previous attempt by others, closed but not merged)

### Backgroud
I see many people hope to collect the request and response body in tcp/udp/http log plugin, including myself. If this PR is acceptable, I will continue to add this option for udp and http log plugins.

### Additional Explanation
Currently, for grpc requests using http2, no matter how large the `client_body_buffer_size` is set, nginx will store the request body in a temporary file. In order to reduce the consumption of reading files, and considering that the grpc request body is in a binary format and is not easy to read, this switch currently only supports logging http(actually http1.x) request body. See [here](https://github.com/Kong/kong/pull/5968#discussion_r435916465) for more specific analysis.

### DIFF
--- a/kong/pdk/log.lua
+++ b/kong/pdk/log.lua
@@ -572,11 +572,11 @@ do
         response_size = tonumber(response_size, 10)
       end
 
-      local request_body
-      local response_body
-      if kong.ctx.plugin then
-        request_body = kong.ctx.plugin.request_body
-        response_body = table.concat(kong.ctx.plugin.response_body)
+      local request_body, response_body
+      local ctx_plugin = kong.ctx.plugin
+      if ctx_plugin and ctx_plugin.response_body then
+        request_body = ctx_plugin.request_body
+        response_body = table.concat(ctx_plugin.response_body)
       end
 
       return {
--- a/kong/pdk/log.lua
+++ b/kong/pdk/log.lua
@@ -572,6 +572,13 @@ do
         response_size = tonumber(response_size, 10)
       end
 
+      local request_body
+      local response_body
+      if kong.ctx.plugin then
+        request_body = kong.ctx.plugin.request_body
+        response_body = table.concat(kong.ctx.plugin.response_body)
+      end
+
       return {
         request = {
           uri = request_uri,
@@ -580,6 +587,7 @@ do
           method = okong.request.get_method(), -- http method
           headers = req_headers,
           size = request_size,
+          body = request_body,
           tls = request_tls
         },
         upstream_uri = var.upstream_uri,
@@ -587,6 +595,7 @@ do
           status = ongx.status,
           headers = resp_headers,
           size = response_size,
+          body = response_body,
         },
         tries = (ctx.balancer_data or {}).tries,
         latencies = {
--- a/kong/plugins/tcp-log/handler.lua
+++ b/kong/plugins/tcp-log/handler.lua
@@ -52,6 +52,29 @@ local TcpLogHandler = {
 }
 
 
+function TcpLogHandler:access(conf)
+  if conf.log_body then
+    local body, err = kong.request.get_raw_body()
+    if err then
+      kong.log.err(err)
+      kong.ctx.plugin.request_body = ""
+    else
+      kong.ctx.plugin.request_body = body
+    end
+    kong.ctx.plugin.response_body = {}
+  end
+end
+
+
+function TcpLogHandler:body_filter(conf)
+  if conf.log_body then
+    local chunk = ngx.arg[1]
+    local body = kong.ctx.plugin.response_body
+    body[#body + 1] = chunk
+  end
+end
+
+
 function TcpLogHandler:log(conf)
   local message = kong.log.serialize()
   local ok, err = timer_at(0, log, conf, message)
--- a/kong/plugins/tcp-log/schema.lua
+++ b/kong/plugins/tcp-log/schema.lua
@@ -13,6 +13,7 @@ return {
           { keepalive = { type = "number", default = 60000 }, },
           { tls = { type = "boolean", default = false }, },
           { tls_sni = { type = "string" }, },
+          { log_body = { default = false, type = "boolean" }, },
         },
     }, },
   }
--- a/spec/01-unit/01-db/01-schema/11-declarative_config/02-process_auto_fields_spec.lua
+++ b/spec/01-unit/01-db/01-schema/11-declarative_config/02-process_auto_fields_spec.lua
@@ -359,6 +359,7 @@ describe("declarative config: process_auto_fields", function()
                       keepalive = 60000,
                       timeout = 10000,
                       tls = false,
+                      log_body = false,
                     }
                   },
                 }
@@ -680,6 +681,7 @@ describe("declarative config: process_auto_fields", function()
                           keepalive = 60000,
                           timeout = 10000,
                           tls = false,
+                          log_body = false,
                         }
                       }
                     }
--- a/spec/01-unit/01-db/01-schema/11-declarative_config/03-flatten_spec.lua
+++ b/spec/01-unit/01-db/01-schema/11-declarative_config/03-flatten_spec.lua
@@ -591,7 +591,8 @@ describe("declarative config: flatten", function()
                   port = 10000,
                   timeout = 10000,
                   tls = false,
-                  tls_sni = null
+                  tls_sni = null,
+                  log_body = false,
                 },
                 consumer = null,
                 created_at = 1234567890,
@@ -1077,7 +1078,8 @@ describe("declarative config: flatten", function()
                   port = 10000,
                   timeout = 10000,
                   tls = false,
-                  tls_sni = null
+                  tls_sni = null,
+                  log_body = false,
                 },
                 consumer = null,
                 created_at = 1234567890,
--- a/spec/01-unit/10-log_serializer_spec.lua
+++ b/spec/01-unit/10-log_serializer_spec.lua
@@ -196,7 +196,7 @@ describe("kong.log.serialize", function()
       it("serializes request and response bodies present in the ngx context", function()
         kong.ctx.plugin = {
           request_body = "request body",
-          response_body = "response body",
+          response_body = {"response body"},
         }
 
         local res = kong.log.serialize({ngx = ngx, kong = kong, })
--- a/spec/01-unit/10-log_serializer_spec.lua
+++ b/spec/01-unit/10-log_serializer_spec.lua
@@ -192,6 +192,19 @@ describe("kong.log.serialize", function()
 
         kong.log.warn = orig_warn
       end)
+
+      it("serializes request and response bodies present in the ngx context", function()
+        kong.ctx.plugin = {
+          request_body = "request body",
+          response_body = "response body",
+        }
+
+        local res = kong.log.serialize({ngx = ngx, kong = kong, })
+        assert.is_table(res)
+
+        assert.equal(res.request.body, "request body")
+        assert.equal(res.response.body, "response body")
+      end)
     end)
   end)
 
--- a/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
+++ b/spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
@@ -43,9 +43,9 @@ for _, strategy in helpers.each_strategy() do
         }, },
       }
 
-
       local route2 = bp.routes:insert {
-        hosts = { "tcp_logging_tls.com" },
+        hosts = { "tcp_logging.com" },
+        paths = { "/log_body" },
       }
 
       bp.plugins:insert {
@@ -54,7 +54,37 @@ for _, strategy in helpers.each_strategy() do
         config   = {
           host   = "127.0.0.1",
           port   = TCP_PORT,
+          log_body = true,
+        },
+      }
+
+      local route3 = bp.routes:insert {
+        hosts = { "tcp_logging_tls.com" },
+      }
+
+      bp.plugins:insert {
+        route = { id = route3.id },
+        name     = "tcp-log",
+        config   = {
+          host   = "127.0.0.1",
+          port   = TCP_PORT,
           tls    = true,
+        },
+      }
+
+      local route4 = bp.routes:insert {
+        hosts = { "tcp_logging_tls.com" },
+        paths = { "/log_body" },
+      }
+
+      bp.plugins:insert {
+        route = { id = route4.id },
+        name     = "tcp-log",
+        config   = {
+          host   = "127.0.0.1",
+          port   = TCP_PORT,
+          tls    = true,
+          log_body = true,
         },
       }
 
@@ -63,14 +93,14 @@ for _, strategy in helpers.each_strategy() do
         url = "grpc://localhost:15002",
       })
 
-      local route3 = assert(bp.routes:insert {
+      local route5 = assert(bp.routes:insert {
         service = grpc_service,
         protocols = { "grpc" },
         hosts = { "tcp_logging_grpc.test" },
       })
 
       bp.plugins:insert {
-        route = { id = route3.id },
+        route = { id = route5.id },
         name     = "tcp-log",
         config   = {
           host   = "127.0.0.1",
@@ -83,14 +113,14 @@ for _, strategy in helpers.each_strategy() do
         url = "grpcs://localhost:15003",
       })
 
-      local route4 = assert(bp.routes:insert {
+      local route6 = assert(bp.routes:insert {
         service = grpcs_service,
         protocols = { "grpcs" },
         hosts = { "tcp_logging_grpcs.test" },
       })
 
       bp.plugins:insert {
-        route = { id = route4.id },
+        route = { id = route6.id },
         name     = "tcp-log",
         config   = {
           host   = "127.0.0.1",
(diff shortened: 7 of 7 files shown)


crackersophia:
docker_login_fail

I met a similar problem like #5951, and saw ci improvements in #5965. It's nice to see the improvements soon, but so far, they don't seem to have merged. It seems that I need to wait for these commits to be merged and try to commit again.

crackersophia:
Sorry to bother, would you have time to review it? @kikito @bungle

crackersophia:
Is there any feedback? @Tieske @hutchic @gszr @hishamhm

crackersophia:
Kong's tests seems not very stable. After I rebase to the latest next branch this morning, the following [lint and unit error](https://travis-ci.org/github/Kong/kong/jobs/694494636#L424) that had never occurred before appeared in Travis CI:
travis_ci_init_error

----
Two other strange phenomena:
- **the results of GitHub Actions and Travis CI are inconsistent**
In my last rebase push yesterday, the test results of GitHub Actions and Travis CI are inconsistent. In GitHub Actions, a test unrelated to my changes failed, but in Travis CI all tests passed.
  - Results of GitHub Actions:
  git_actions_fail

  - Results of Travis CI:
  image

  And I did not modify `spec/02-integration/04-admin_api/15-off_spec.lua`.

  I saw that kong plans to [use GitHub Actions as CI to replace TravisCI](https://github.com/Kong/kong/commit/4262e51ddc55a7e7bd65b8285ad1a886c70e197c). The difference between the two seems to be that GitHub Actions splits Kong core integration tests into two halves, but they test the same things. So why all the tests that passed in GitHub Actions, but some of them failed in Travis CI? Did concurrency affect the test results?

- **Travis CI test results are unstable**
Sometimes I just push after rebase with the latest upstream(without any substantial personal modifications), sometimes the test all passes, sometimes there are individual failures, and the failure test are different each time.

kong/plugins/tcp-log/handler.lua
```diff
@@ -7,6 +7,27 @@ local ngx = ngx
 local timer_at = ngx.timer.at
 
 
+local function get_body_data()
+  local req = ngx.req
+
+  req.read_body()
+  local data = req.get_body_data()
```
Tieske, line 14: any reason for not using pdk functions? https://docs.konghq.com/2.0.x/pdk/kong.request/#kongrequestget_raw_body

kong/plugins/tcp-log/handler.lua
```diff
@@ -7,6 +7,27 @@ local ngx = ngx
...
+  if data then
+    return data
+  end
+
+  local file_path = req.get_body_file()
```
Tieske, line 19: is this wise? it only gets flushed to disk if it is too big. So this is inherently dangerous, I'd skip this, and tell userss to increase the limit before caching to disk if they need it in the logs anyway

kong/plugins/tcp-log/handler.lua
```diff
@@ -53,6 +74,22 @@ local TcpLogHandler = {
 }
 
 
+function TcpLogHandler:access(conf)
+  if conf.log_body then
+    ngx.ctx.request_body = get_body_data()
+    ngx.ctx.response_body = ""
```
Tieske, line 80: this risks name collisions, use the pdk again for plugin storage: https://docs.konghq.com/2.0.x/pdk/kong.ctx/#kongctxplugin

kong/plugins/tcp-log/handler.lua
crackersophia: Good advice. Using PDK functions instead of making my own wheels. The reason I wrote it myself is that I saw the recommendation [here](https://github.com/Kong/kong/pull/3028#discussion_r152298066). Maybe the `kong.request.get_raw_body()` of the pdk did not appear at that time. I will fix it.

kong/plugins/tcp-log/handler.lua
crackersophia: get_raw_data

`kong.request.get_raw_body()` also includes `get_body_file()`. It is difficult to guarantee that users will set `client_body_buffer_size` to be as high as `client_max_body_size` to force request bodies to be kept in memory. How about tell users to increase the limit of `client_body_buffer_size` and use `kong.request.get_raw_body()` at meanwhile?  When this feature is released in the future, I will add a increase limit suggestion about the `log_body` parameter of the tcp-log plugin at https://docs.konghq.com.

kong/plugins/tcp-log/handler.lua
crackersophia: Good advice. I'll fix it.

kong/plugins/tcp-log/handler.lua
crackersophia: I did some experiments before. For **grpc requests**, no matter how the `client_body_buffer_size` is set or how small the request body is, the request body will always be stored in a temporary file. This situation has been encountered in both source installation and run docker kong.

This is not a problem of kong, but a bug of nginx processing HTTP/2 request. I saw nginx fixed it: https://nginx.googlesource.com/nginx/+/2939d5c5fadb3f28a8f8ac85c039bfd007239d3c%5E%21/#F0.

But in openresty-1.15.8.2, the nginx code is still the problematic code before the bug fix.

openresty_nginx

In this case, if you do not read the temporary file, the grpc request body will not be collected. I will continue to dig this problem tomorrow, and if you know more, please let me know ：）

kong/plugins/tcp-log/handler.lua
crackersophia: I did some experiments before. For **grpc requests**, no matter how the `client_body_buffer_size` is set or how small the request body is, the request body will always be stored in a temporary file. This situation has been encountered in both source installation and run docker kong.

This is not a problem of kong, but a bug of nginx processing HTTP/2 request. I saw nginx fixed it: https://nginx.googlesource.com/nginx/+/2939d5c5fadb3f28a8f8ac85c039bfd007239d3c%5E%21/#F0.

But in openresty-1.15.8.2, the nginx code is still the problematic code before the bug fix.

openresty_nginx

In this case, if you do not read the temporary file, the grpc request body will not be collected. I will continue to dig this problem tomorrow, and if you know more, please let me know ：）

crackersophia: This bug was fixed in [google nginx](https://nginx.googlesource.com/nginx), but it has not been merged by the official nginx so far. At present, this bug still exists in the official nginx code, which causes other projects using the official nginx **will always save the request body to a temporary file when processing grpc requests**.

I just got the [reason](https://github.com/cloudendpoints/esp/issues/395#issuecomment-639431437) from a google developer who proposed this fix, he says that the patch was sent upstream and rejected, see: http://mailman.nginx.org/pipermail/nginx-devel/2018-December/011657.html. The simple summary is that the official knows that this problem exists, but they think that the current fix is not comprehensive enough.


So at present, if we want to get the grpc request body, we do need to read the temporary file.  **One thing that needs to be clarified is that the reading and writing of the file would have happened for grpc requests, not caused by the plugin. If users select to log request body when using the tcp-log plugin, only one more read will be caused.** 

If we use the pdk function `kong.request.get_raw_body()`, will not get the request body for the grpc request since it not support reading from temporary files, and this PR will lose its meaning for gprc part (http part is normal).

I tend to use the previous function implementation to support logging grpc requests. Not everyone will set the switch, but this switch provides a possibility for those who need it. Hope to hear your thoughts.

kong/plugins/tcp-log/handler.lua
crackersophia: Fixed.

crackersophia:
Build & Test / cassandra (plugins) all integration tests failed:

image

image

kong/plugins/tcp-log/handler.lua
crackersophia: The initial implementation is not good enough. For http requests, it will still cause file reading, ignoring the `client_body_buffer_size` parameter.

Two other options:
1. Use `kong.request.get_raw_body()`, and tell users that only supports logging http request body currently. After nginx improves, it will support logging grpc requests.

2. Add additional processing for grpc requests in `kong.request.get_raw_body()`(this requires a separate PR). `client_body_buffer_size` should be transparent to users. Users should not be concerned about the impact of different protocols on `client_body_buffer_size`, they only need to know that when the request body is less than this value, it can be logged.

kong/plugins/tcp-log/handler.lua
crackersophia: This question is related to the following question. Let's discuss how to do better.

crackersophia: Fixed. Considering that the grpc request body is not easy to read,  let the backend service log it at this stage. Now use the pdk function `kong.request.get_raw_body()`, which will support logging http1.x request body.

kong/plugins/tcp-log/handler.lua
crackersophia: The first option was chosen.

crackersophia:
The requested changes have been completed. Any other feedback? @Tieske 
Some of the tests failed, but it seems not related to my changes.

teunis90:
@crackersophia @Tieske Any progress on this?

crackersophia:
> @crackersophia @Tieske Any progress on this?

I haven't received a response for many days, maybe @Tieske can help us check the progress here and how to deal with this PR later.

kong/plugins/tcp-log/handler.lua
```diff
@@ -52,6 +52,28 @@ local TcpLogHandler = {
...
+function TcpLogHandler:access(conf)
+  if conf.log_body then
+    local body, err = kong.request.get_raw_body()
+    if err then
+      kong.log.err(err)
+      kong.ctx.plugin.request_body = ""
+    else
+      kong.ctx.plugin.request_body = body
+    end
+    kong.ctx.plugin.response_body = ""
```
Tieske, line 64:
```suggestion
    kong.ctx.plugin.response_body = {}
```

kong/plugins/tcp-log/handler.lua
```diff
@@ -52,6 +52,28 @@ local TcpLogHandler = {
...
+  end
+end
+
+
+function TcpLogHandler:body_filter(conf)
+  if conf.log_body then
+    local chunk = ngx.arg[1]
+    kong.ctx.plugin.response_body = kong.ctx.plugin.response_body .. (chunk or "")
```
Tieske, line 72:
```suggestion
    local body = kong.ctx.plugin.response_body
    body[#body + 1] = chunk
```

kong/pdk/log.lua
```diff
@@ -525,20 +525,29 @@ do
 
       local host_port = ctx.host_port or var.server_port
 
+      local request_body
+      local response_body
+      if kong.ctx.plugin then
+        request_body = kong.ctx.plugin.request_body
+        response_body = kong.ctx.plugin.response_body
```
Tieske, line 532:
```suggestion
        response_body = concat(kong.ctx.plugin.response_body)
```

Tieske (changes requested):
suggestions to have less potentially large memory operations.

I'll leave the final say to @Kong/team-core

kong/plugins/tcp-log/handler.lua
crackersophia: I made changes, but some tests failed:
image

Corresponding code:
image

This function is used to check whether the temporary log path exists and get its size, but it times out after waiting 10 seconds. Is this accidental?

Tieske: shouldn't be related

crackersophia: I have made changes as proposed. Can this pr now be merged, or does it require all tests to pass?

Tieske: looked into the failures, and they seem to be related; https://travis-ci.org/github/Kong/kong/jobs/737346018#L4696-L4707

They are all logging plugins that fail, and some changes you made are shared by all those plugins. Maybe try rebase on `next` first, and try and run those tests again.

Mnbvcxzaq1990 (approved):

Mnbvcxzaq1990 (approved):

kong/plugins/tcp-log/handler.lua
crackersophia: Thank you for suggestions. Before I push again, I have rebase `next`, and currently my branch contains all the commits of the `next` branch and two commits ahead:

image

Observe the failed test, the [01-tcp-log](https://travis-ci.org/github/Kong/kong/jobs/737346018#L477) failed test is the original test, not my newly added test case, and the log body switch is not enabled in this test case. The error line 219 is `local ok, res = thread:join()` and the `ok` is false, which is related to the previous `thread:join()` timeout: https://travis-ci.org/github/Kong/kong/jobs/737346018#L471. 
image


The failure case of [02-udp-log](https://travis-ci.org/github/Kong/kong/jobs/737346018#L542) and [07-loggly](https://travis-ci.org/github/Kong/kong/jobs/737346018#L645) are also related to the result of `local ok, res = thread:join()`. The [03-http-log](https://travis-ci.org/github/Kong/kong/jobs/737346018#L542) and [04-file-log](https://travis-ci.org/github/Kong/kong/jobs/737346018#L571) failed test related to `wait_until() timeout (after delay 10s)` etc.

I don't see a rebuild button to re-run the test, and there is no new submission to trigger the test. Could you help me run the test again?

Tieske: if you look at the test code (not in the helpers.lua) you probably see that it fires a request, and then wait for a specific message to appear in the logs. My guess is that the message format changed, and hence the timeout occurs while waiting. So the test most likely must be adjusted for the new output, or if the output doesn't match because it is erroneous, the code must be adapted obviously.

crackersophia: I found the reason. In the last proposal, we changed the type of `kong.ctx.plugin.response_body` from **string** to **table**.

image

Therefore, it's necessary to check if `kong.ctx.plugin.response_body` is nil before using `table.concat(kong.ctx.plugin.response_body)` in pkg/log.lua. This problem has now been fixed.

image


In this time, some integration tests failed: https://github.com/Kong/kong/pull/5968/checks?check_run_id=1291355165#step:6:1552, but it does not seem to be related to my changes.

Because the previous integration test failed, many later plugin tests were canceled. But from here we can see that the `01-tcp-log` test passed: https://github.com/Kong/kong/pull/5968/checks?check_run_id=1291355317#step:6:1352

It seems that there is no problem, but maybe we can rerun all the tests to make sure there are no other problems.

kong/pdk/log.lua
```diff
@@ -572,6 +572,13 @@ do
         response_size = tonumber(response_size, 10)
       end
 
+      local request_body
+      local response_body
+      if kong.ctx.plugin and kong.ctx.plugin.response_body then
+        request_body = kong.ctx.plugin.request_body
+        response_body = table.concat(kong.ctx.plugin.response_body)
+      end
```
Tieske, line 580:
```suggestion
      local ctx_plugin = kong.ctx.plugin
      if ctx_plugin and ctx_plugin.response_body then
        request_body = ctx_plugin.request_body
        response_body = table.concat(ctx_plugin.response_body)
      end
```
There is some meta-table magic involved in getting those tables, so instead of invoking that 4 times, do it only once.

crackersophia: Haha, good advice and I have fixed it. And the test result is the same as last time.

crackersophia: Now all the tests have passed.

Tieske (approved):

Tieske:
@Kong/team-core this PR has been open for a long time, please review when done with 2.2.

crackersophia:
> @Kong/team-core this PR has been open for a long time, please review when done with 2.2.

2.2.0 was released a few hours ago. 😶

javierguerragiraldez:
Sorry for the delay answering this.

We (the team-core) are developing a way for a plugin to easily add custom data to the log output.  That would make this feature much cleaner to implement.  Personally, I feel that reading from `ctx.plugin` in the `kong.log.serialize()` function is not a good idea.

crackersophia:
> Sorry for the delay answering this.
> 
> We (the team-core) are developing a way for a plugin to easily add custom data to the log output. That would make this feature much cleaner to implement. Personally, I feel that reading from `ctx.plugin` in the `kong.log.serialize()` function is not a good idea.

Well, I chose `ctx.plugin` because it is not shared between plugins, and only visible for the current plugin instance. I just hope the log plugin can support the request/response body, and maybe there are some more elegant and flexible methods. Looking forward to seeing the new plugins you mentioned.

teunis90:
> We (the team-core) are developing a way for a plugin to easily add custom data to the log output

@javierguerragiraldez  is this effort tracked in a public ticket? I would like to subscribe to updates on this particular feature.

kikito:
@teunis90 This feature is scheduled to be included in Kong 2.3. We don't have a public tracker, but this issue is associated with an internal ticket.

I'm reopening and associating this github ticket to it. If you monitor this ticket, you should get updates.

kikito:
Hello @teunis90, I have implemented a way to customize log data on all log plugins on #6672, which is already merged on the `next` branch, and should be included in Kong 2.3, including setting some parts of it to `nil`.

teunis90:
@kikito Thanks for your update. Can we expect an update for the Log-type plugins (like file_log, or http_log) to support logging of request/response body logging? Or should we build our own plugin on top of this new serializer logic?

kikito:
@teunis90  The request/response body is left out of the logs on purpose as it is not bound (we don't want to log the response body of a video streaming service) and will likely never be, within Kong itself. Please feel free to implement a plugin using the new functionality.

PR #5970, opened 2020-05-29
tests(hybrid) re-enable hybrid mode tests

hishamhm:
These tests seemed to have bit-rotten a bit, so I gave them some love locally and they seem to run fast and pass ok on Travis now


spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -59,13 +63,18 @@ for _, strategy in helpers.each_strategy() do
       local route_id
 
       it("proxy on DP follows CP config #flaky", function()
```
hutchic, line 65: if we're re-enabling these should the flaky be dropped?

hishamhm: Good catch! the other similar file already had #flaky removed. Force-pushing the removal to try it on CI.

PR #5991, opened 2020-06-07
feat(core) a new response phase for buffered proxying

### LINKED ISSUE #5938: Access response body by go plugin
Hello! I was trying to get the response body in a go plugin but it takes an error.
I was reading the docs and checking some things in go and i notice that you can call lua methods so, i try this:

```
func (conf *MyConfig) Access(kong *pdk.PDK) {
   enabling, err := kong.ServiceRequest.Ask("kong.service.request.enable_buffering")
   if err != nil {
   	kong.Log.Err(fmt.Sprint(err))
   }
   kong.Log.Notice("Enabling body and headers access", enabling)
}
```
then when i try to get the body from the response i try:
```
func (conf *MyConfig) Log(kong *pdk.PDK) {
	res, err := kong.ServiceResponse.Ask("kong.service.request.get_raw_body")
	if err != nil {
		kong.Log.Err(fmt.Sprint(err))
	}
	kong.Log.Notice(res)

}
```
and this triggers me an error.
```
api-gateway_1  | 2020/05/21 21:35:04 [error] 32#0: *452 lua entry thread aborted: runtime error: /usr/local/share/lua/5.1/kong/pdk/private/phases.lua:73: no phase in kong.ctx.core.phase
api-gateway_1  | stack traceback:
api-gateway_1  | coroutine 0:
api-gateway_1  |        [C]: in function 'error'
api-gateway_1  |        /usr/local/share/lua/5.1/kong/pdk/private/phases.lua:73: in function 'check_phase'
api-gateway_1  |        /usr/local/share/lua/5.1/kong/pdk/service/response.lua:278: in function 'call_pdk_method'
api-gateway_1  |        /usr/local/share/lua/5.1/kong/db/dao/plugins/go.lua:338: in function 'bridge_loop'
api-gateway_1  |        /usr/local/share/lua/5.1/kong/db/dao/plugins/go.lua:461: in function </usr/local/share/lua/5.1/kong/db/dao/plugins/go.lua:459>, context: ngx.timer, client: 172.31.0.1, server: 0.0.0.0:8000
```
so, anyone knows about this error? and how could i solve them?

### LINKED ISSUE #6203: How can i modify the respone body in plugin written in Go?
Hi~~~
Can we access and modify the response body in Go plugins, for doing something like making a Go version of Response Transformer plugin?
As Go plugin doesn't support body_filter phase, seems like we can't do it now.

javierguerragiraldez:
# Summary
new `response` phase for plugins, run after `access` phase (but still within OpenResty's `access` phase).

### Issues resolved

Fix #5938, fix https://github.com/Kong/go-pdk/issues/34, fix #6203


bungle:
My comment on this:
```
        handlers.header_filter(true)
        handlers.body_filter(true)
```

The thing is that this would be beneficial to `Lua` plugins too. E.g. have `post` proxying phase with `co-sockets`. Right now I cannot implement JWT Response Verifier even with `buffered proxying` because we don't have this. Also the `header_filter` and `body_filter` is now overloaded. On `Go` it means `access` and `access` and on Lua it means `header_filter`  and `body_filter`.

Because of this limitation in OpenResty, I think the best would be to introduce another phase, and not try to fake `header_filter` and `body_filter` for Go. Another phase would also benefit the Lua plugins (as they can then have body + headers + cosockets at the same time).

Thus, I propose that we introduce a new phase, let's call it e.g. `response`. This phase is called only if your plugin has requested buffered proxying.

My original implementation of buffered proxying was implemented by yielding, that your plugin could yield on access phase, then the request was proxied and then those yielded plugins were resumed. That implementation did not introduce a new phases, but allowed the plugin to yield in middle. But we went with current approach. I found later that it was a less flexible approach. So now that I look at code, I think we should probably introduce a new `Kong Phase` which we can call `response`. So on Lua side you could have:

- `plugin.init_worker`
- `plugin.certificate`
- `plugin.rewrite`
- `plugin.access`
- `plugin.response` (new fake phase that is ran with buffering, the nginx phase is still `access`)
- `plugin.header_filter`
- `plugin.body_filter`
- `plugin.log`

The Go side stays the same as before, but it will have the `response` phase available (so no `fake` `header_filter` or fake `body_filter`, we might want to have the real thing at some point or perhaps we get cosockets for those in a future). I just feel better that we don't mix the things (or `fake`, thus proposing this new phase for both Lua and Go.

hishamhm:
> @hishamhm what would be your take on this?

Definitely worth considering as a wider discussion for Kong 2.2. Let's revisit this before moving this PR further and include these points about Lua plugins in the discussion.

kong/init.lua
```diff
@@ -766,8 +725,120 @@ function Kong.access()
...
+
+    if not ctx.KONG_PROCESSING_START then
+      ctx.KONG_PROCESSING_START = ngx.req.start_time() * 1000
+    end
+
+    if not ctx.workspace then
+      ctx.workspace = kong.default_workspace
+    end
+
+    if not ctx.KONG_BUFFERED_RESPONSE_START then
```
bungle, line 792: as the phase is just `response` I propose (perhaps later on we can enable this without buffering, e.g. make a patch to nginx and make it real or something):

```suggestion
    if not ctx.KONG_RESPONSE_START then
```

kong/init.lua
```diff
@@ -766,8 +725,120 @@ function Kong.access()
...
+      ctx.KONG_BUFFERED_RESPONSE_START = get_now_ms()
```
bungle, line 793:
```suggestion
      ctx.KONG_RESPONSE_START = get_now_ms()
```

kong/init.lua
```diff
@@ -766,8 +725,120 @@ function Kong.access()
...
+      ngx.status = 502
+      return kong_error_handlers(ngx)
+    end
+
+
+    -- from response.exit
+    local core = kong.ctx.core
+    core.buffered_status = res.status
+    core.buffered_headers = res.header
+    core.buffered_body = res.body
```
bungle, line 779: Move this block just before line 840:
```
    -- from buffered_proxy
    return kong.response.exit(res.status, res.body, res.header)
```

kong/init.lua
```diff
@@ -766,8 +725,120 @@ function Kong.access()
...
+
+    -- from header_filter (s/HEADER_FILTER/BUFFERED_RESPONSE/)
+
+    if not ctx.KONG_PROCESSING_START then
+      ctx.KONG_PROCESSING_START = ngx.req.start_time() * 1000
+    end
+
+    if not ctx.workspace then
+      ctx.workspace = kong.default_workspace
+    end
```
bungle, line 790: As this is fake-phase, we can remove these.

kong/init.lua
```diff
@@ -766,8 +725,120 @@ function Kong.access()
...
+                                  ctx.KONG_HEADER_FILTER_START
+        ctx.KONG_ACCESS_TIME = ctx.KONG_ACCESS_ENDED_AT -
+                              ctx.KONG_ACCESS_START
+      end
+
+      if ctx.KONG_BALANCER_START and not ctx.KONG_BALANCER_ENDED_AT then
+        ctx.KONG_BALANCER_ENDED_AT = ctx.KONG_HEADER_FILTER_START
+        ctx.KONG_BALANCER_TIME = ctx.KONG_BALANCER_ENDED_AT -
+                                ctx.KONG_BALANCER_START
+      end
```
bungle, line 814: As this is fake-phase, we can remove these.

kong/init.lua
```diff
@@ -766,8 +725,120 @@ function Kong.access()
...
+      ctx.KONG_WAITING_TIME = ctx.KONG_BUFFERED_RESPONSE_START -
+                            (ctx.KONG_BALANCER_ENDED_AT or ctx.KONG_ACCESS_ENDED_AT)
+
+      if not ctx.KONG_PROXY_LATENCY then
+        ctx.KONG_PROXY_LATENCY = ctx.KONG_BUFFERED_RESPONSE_START - ctx.KONG_PROCESSING_START
+      end
+
+    elseif not ctx.KONG_RESPONSE_LATENCY then
+      ctx.KONG_RESPONSE_LATENCY = ctx.KONG_BUFFERED_RESPONSE_START - ctx.KONG_PROCESSING_START
+    end
```
bungle, line 827: As this is a fake-phase, we can remove these.

kong/init.lua
```diff
@@ -766,8 +725,120 @@ function Kong.access()
...
+
+    kong_global.set_phase(kong, PHASES.response)
+
+    runloop.response.before(ctx)
+    local plugins_iterator = runloop.get_plugins_iterator()
+    execute_plugins_iterator(plugins_iterator, "response", ctx)
+    runloop.response.after(ctx)
+
+    ctx.KONG_BUFFERED_RESPONSE_ENDED_AT = get_now_ms()
+    ctx.KONG_BUFFERED_RESPONSE_TIME = ctx.KONG_BUFFERED_RESPONSE_ENDED_AT - ctx.KONG_BUFFERED_RESPONSE_START
```
bungle, line 837:
```suggestion
    ctx.KONG_RESPONSE_ENDED_AT = get_now_ms()
    ctx.KONG_RESPONSE_TIME = ctx.KONG_BUFFERED_RESPONSE_ENDED_AT - ctx.KONG_RESPONSE_START
```

Tieske:
if updating the runloop, we should also add an authentication phase, and imho rename the other ones to something more useful

bungle:
> if updating the runloop, we should also add an authentication phase, and imho rename the other ones to something more useful

Yes, but that is rather complicated, with all that `AND` / `OR` logics, and perhaps getting a rid of that `config.anonymous` hackery. This is about to circumvent Go plugins that cannot currently have post request phases.

Tieske:
> > if updating the runloop, we should also add an authentication phase, and imho rename the other ones to something more useful
> 
> Yes, but that is rather complicated, with all that `AND` / `OR` logics, and perhaps getting a rid of that `config.anonymous` hackery. This is about to circumvent Go plugins that cannot currently have post proxy phases (except log that is faked with timer and has its own problems).

exactly, mentioning it to make sure we at least pick something that keeps this in mind, so we won't regret it later

kong/runloop/plugins_iterator.lua
```diff
@@ -243,11 +243,6 @@ local function get_next(self)
       local cfg = load_configuration_through_combos(ctx, combos, plugin)
       if cfg then
         plugins[name] = cfg
-
-        if plugin.handler.response then
-          -- there's a "response" handler for this request, make it buffered
-          kong.ctx.core.buffered_proxying = true
-        end
```
javierguerragiraldez: removing this means any plugin with a `response` phase needs an `access` handler just to turn it on.  for Go plugins that means three extra roundtrips on each request.

bungle: Can we somehow make it detect it for `go` only plugins? I mean that if people write an abstraction or interface that implements this response function, even when it is empty by default, it will always turn on the buffering. Because it has a lot of issues, we don't want this to be turned on accidentally.

bungle: I moved it to `response`, as I think we need to have `configuration` to determine whether or not the plugin with response was configured. Only exception is global plugins, but a global plugin with response implemented is a huge no-no.

bungle: We can probably deprecate `kong.service.request.enable_buffering()` now.

ifree321:
I'm very looking forward for the response phase, when will can we have it?

sherodtaylor:
Does this fix the `log` phase issues

javierguerragiraldez:
@ifree321 
> I'm very looking forward for the response phase, when will can we have it?

It's scheduled for the 2.2 release.  Unofficially, you can always merge and try yourself! :smile: 

@sherodtaylor 
> Does this fix the `log` phase issues

This doesn't touch any other phases.  If you mean about something already mentioned elsewhere, please comment on the relevant issue or open a new one if you think it's not already addressed.

sherodtaylor:
ah i saw a link to this pr in the issue #5938 i thought this touched it

javierguerragiraldez:
I see.  these are related, but also a slight case of "X/Y problem".

Go plugins `Log` handlers are indeed run under a timer, because of an OpenResty design limitation.  If that maks some data unavailable, you can retrieve it at another phase and save in a Go private variable; that's much more efficient than saving it "just in case" some plugins _might_ need it.

But that still leaves with no way to access response data, since Go plugins can't handle `header_filter` or `body_filter` phases (again, because of the same OpenResty limitation).  This is what this PR addresses by adding the `response` phase, which is usable from Go plugins (or will be, once this is merged)

kong/init.lua
```diff
@@ -795,8 +761,96 @@ function Kong.access()
...
+      share_all_vars      = true,
+      method              = HTTP_METHODS[ngx.req.get_method()],
+      ctx                 = ctx,
+    }
+
+    local res = ngx.location.capture("/kong_buffered_http", options)
+    if res.truncated then
+      ngx.status = 502
+      return kong_error_handlers(ngx)
+    end
```
dndx, line 807: Instead of adding these buffered proxy logic in the `init.lua` file, should we consider moving them to the `handler.lua` file `before` and `after` functions instead? That seems to conform to what the `init.lua` file does closer.

kong/init.lua
```diff
@@ -795,8 +761,96 @@ function Kong.access()
...
+
+    local ctx = ngx.ctx
+
+    -- buffered proxying (that also executes the balancer)
+    ngx.req.read_body()
+
+    local options = {
+      always_forward_body = true,
+      share_all_vars      = true,
+      method              = HTTP_METHODS[ngx.req.get_method()],
```
dndx, line 799: Maybe `kong.request.get_method()` is better? I know we have always said core code should be making an example to utilize PDK functions more.

kong/pdk/private/phases.lua
```diff
@@ -14,6 +14,7 @@ local PHASES = {
   rewrite           = 0x00000010,
   access            = 0x00000020,
   balancer          = 0x00000040,
+  response          = 0x00000080,
```
dndx, line 17: Just a thought, have we considered making the phase name reflect more on what it does? For example, what about `buffered_response`. That way a new Kong developer won't get confused about why this fake phase exists and be always aware of the performance penalties associated with using it.

kong/init.lua
javierguerragiraldez: might be doable, but feels a bit like hidden flow control; which tends to make the general structure hard to follow.

kong/pdk/private/phases.lua
javierguerragiraldez: that was my first proposal, but after quite a bit of bikeshedding we settled down on just `response`.

kong/init.lua
javierguerragiraldez: i think in this case that would be too anti-layering, given that those functions (in PDK) work on the phases, while this (kong/init) is what defines them.   at first sight it would work, but complicates maintainability.

kong/runloop/plugins_iterator.lua
```diff
@@ -287,6 +287,9 @@ local function get_next(self)
       local cfg = load_configuration_through_combos(ctx, combos, plugin)
       if cfg then
         plugins[name] = cfg
+        if plugin.handler.response and plugin.handler.response ~= BasePlugin.response then
+          self.iterator.has_response = true
```
bungle, line 291: I fear that this might leak to other requests. E.g.:
1. user goes to route r
2. route r has plugin p1 and plugin p2
3. p1 has priority of 10 and p2 has priority of 1
4. p1 does not implement response, but p2 does
5. in request r1 the both plugins run, which set `iterator.has_response = true`
6. in request r2 only p1 runs and it shortcircuits, what is the value of `iterator.has_response`?
7. also as iterator remains the same, would it be forever true, once it has been set to true?

OR did I understand it wrong?

kong/init.lua
```diff
@@ -769,6 +731,10 @@ function Kong.access()
     ctx.workspace = old_ws
   end
 
+  if plugins_iterator.has_response then
```
bungle, line 734: As noted below, I feel this is leaking across the requests, so I would like to know more that it is not so.

kong/init.lua
```diff
@@ -795,8 +761,96 @@ function Kong.access()
   -- we intent to proxy, though balancer may fail on that
   ctx.KONG_PROXIED = true
 
-  if kong.ctx.core.buffered_proxying then
-    return buffered_proxy(ctx)
+  if kong.ctx.core.buffered_proxying and ngx.req.http_version() < 2 then
```
bungle, line 764: Now we also have this requirement, on above we just set the `kong.ctx.core.buffered_proxying = true` without checking this second condition. I think the `kong.ctx.core.buffered_proxying = true` should probably reflect the fact of what really happened.

bungle (changes requested):
Some of my first comments. Also should we remove this:
https://github.com/Kong/kong/blob/master/kong/pdk/service/request.lua#L84-L94

As now it is triggered by implementing the `response`.

Should we add `response` to base plugin too (or not as it is kinda deprecated, not sure if it is officially deprecated though)?

kong/db/dao/plugins.lua
```diff
@@ -157,6 +165,14 @@ local function load_plugin_handler(plugin)
     return nil, plugin .. " plugin is enabled but not installed;\n" .. handler
   end
 
+  if implements(handler, "response") and
+      (implements(handler, "header_filter") or implements(handler, "body_filter"))
+  then
+    return nil, fmt(
+      "Plugin %q can't be loaded because it implements both `response` " ..
+      "and `header_filter` or `body_filter` methods.\n", plugin)
+  end
```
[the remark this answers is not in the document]
javierguerragiraldez: Not technical, but the reasoning was like:

- there should be no reason to use both.
- buffering modifies the `x_filter` methods behaviour (mostly a superset, but still different)
- even though it's documented, it's not too evident that adding `response` turns on buffering.
- if an existing plugin has `x_filter` phases, adding `response` _should_ require a rewriting, since existing behaviour would be affected.

so, it's just about making things evident by forcing the developer to rethink.  of course, it's true that one plugin can turn buffering on and affect other plugins, but that's probably uncommon.

kong/runloop/plugins_iterator.lua
javierguerragiraldez: AFAICT, the flag is persistent but has limited scope, hopefully just the route, so a request would be buffered if any plugin _in the route_ implements `:response`, but not other routes with a different plugin set.

i didn't consider the case where an "early" plugin shortcircuits.  in effect, that would make it impossible to cache.

I guess the choice is to consider the "buffering" to be statically or dynamically determined.  which would be the least surprising / easier to document?

kong/init.lua
javierguerragiraldez: sounds reasonable.  will check.

kong/db/dao/plugins.lua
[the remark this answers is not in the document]
bungle: > if an existing plugin has x_filter phases, adding response should require a rewriting, since existing behaviour would be affected.

If I take any of our plugin that implements `x_filter` and add empty `response` function to them, how the need to be rewritten? I understand that it turns on buffering, but that doesn't change anything on `_filter` phases (yes, it is possible to read body on header filter, but that is mostly about it, and it should not affect any existing plugin).

grantmagdanz:
@javierguerragiraldez Thanks for doing this work. Really excited to see this go in. I've applied the diff and built. It's calling into my plugin's `Response` handler but am wondering how I can modify the body of the response? I see the `GetRawBody` method but there isn't an equivalent set. Looking at the lua response transformer plugin:

```
function ResponseTransformerHandler:body_filter(conf)
  if is_body_transform_set(conf) and is_json_body(kong.response.get_header("Content-Type")) then
    local ctx = ngx.ctx
    local chunk, eof = ngx.arg[1], ngx.arg[2]

    ctx.rt_body_chunks = ctx.rt_body_chunks or {}
    ctx.rt_body_chunk_number = ctx.rt_body_chunk_number or 1

    if eof then
      local chunks = concat(ctx.rt_body_chunks)
      local body = body_transformer.transform_json_body(conf, chunks)
      ngx.arg[1] = body or chunks

    else
      ctx.rt_body_chunks[ctx.rt_body_chunk_number] = chunk
      ctx.rt_body_chunk_number = ctx.rt_body_chunk_number + 1
      ngx.arg[1] = nil
    end
  end
end
```

it looks like it sets `ngx.arg[1]` but I don't see that expose in the Go PDK either. Is modifying the body of the response going to be something supported?

For more context, I want to write a plugin that does object hydration on responses. If a response comes back with a UUID and type, the plugin will call a separate service and replace that object with the hydrated version and cache it.

javierguerragiraldez:
Hi @grantmagdanz , I agree that a `response.SetRawBody()` method could make sense for the buffered case, but not for the generic case.

For now, I guess the appropriate way is to use the `response.Exit()` function.

kong/db/dao/plugins.lua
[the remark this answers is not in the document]
bungle: My reasoning is still the same. If any of our plugins implement any phase, that doesn't prevent them implementing any other phase. This is the first that introduces such limitation, and I don't still see a reason.

Let's say plugin A implements response and header_filter and has priority of 1, Plugin B implements response phase and has priority of 100.

1. Plugin A runs response phase
2. Plugin B runs response phase and sets response header X-Dog: Pluto
3. Plugin A wants to run header_filter and set X-Dog: Bulivar

1. and 2. works, but 3. does not, and I really don't understand why. What does the Plugin writer A need to rethink?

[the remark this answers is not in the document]
bungle: I understand that the ADR said:
> The presence of the response callback in a plugin enables "buffered mode" which enables buffered proxying
> and disables the filter callbacks when loading the plugin, we can error if there are both response and filter
> callbacks with a nice explanation

But my feedback on that was not available back then. Now that I have given it, with example, could we at least rethink about it? @hishamhm?

kong/init.lua
```diff
@@ -768,6 +730,10 @@ function Kong.access()
     ctx.workspace = old_ws
   end
 
+  if ctx.has_response then
+    kong.ctx.core.buffered_proxying = true
+  end
+
```
bungle, line 736: Should we move this closer to usage? I mean there are a returns and response.exits below, and in that case we are not buffered proxying?

bungle (changes requested):

javierguerragiraldez:
i can't approve, but i agree that @bungle 's last commit cleans up things and improves readability.

kikito (approved):

sec23206:
> Hi @grantmagdanz , I agree that a `response.SetRawBody()` method could make sense for the buffered case, but not for the generic case.
> 
> For now, I guess the appropriate way is to use the `response.Exit()` function.

I realize this thread may have gone stale, but I have the same question as  @grantmagdanz about being able to effectively write a bespoke version of the response-transformation plugin in Go.  I have a working model of this in Lua (transforming an XML payload from a legacy web service to JSON), which I based on the original Lua response transformer code, and I'm trying to see if I can redo it in Go.

I've tried to use the suggestion from @javierguerragiraldez above, to call the Exit() method of Response, but I'm not sure my use case is [what it's intended for](https://github.com/Kong/go-pdk/blob/master/response/response.go):

// kong.Response.Exit() interrupts the current processing and produces a response.
// It is typical to see plugins using it to produce a response before Kong
// has a chance to proxy the request (e.g. an authentication plugin rejecting
// a request, or a caching plugin serving a cached response).
//
// This function closes the channel back to Kong main process, so any use of a
// PDK function after this would trigger a run-time panic.  It is recommended to
// stop all processing and return immediately from the handler.

I'm using it in the Response phase, but I'm still getting the panic error message:

> 2021/09/01 18:37:12 [info] 32#0: *28 [go-pluginserver:36] panic: send on closed channel, context: ngx.timer

Is this really not yet feasible with the Go PDK?   Thanks!

PR #6000, opened 2020-06-09
feat(clustering) add certificate expiry check for `cluster_cert` and add

dndx:
expiry and basic constraint check for `cluster_ca_cert`

This makes error such as ones in #5585 easier to identify


kong/conf_loader/init.lua
```diff
@@ -528,6 +529,45 @@ local _nop_tostring_mt = {
...
+  end
+
+  local cert
+  cert, err = x509.new(content)
+  if not cert then
+    return nil, "unable to load \"" .. path .. "\": " .. err
+  end
+
+  local not_after
+  not_after, err = cert:get_not_after()
```
fffonion, line 549: logically we would also check for get_not_before() is < ngx.time() 😄

PR #6011, opened 2020-06-11
fix(runloop) a more robust plugins iterator on init worker

kikito:
This PR includes two changes:
* It handles the creation of a Plugins Iterator on the Init Worker phase on a separate step, which fails less frequently.
* It changes the `get_*_plugin_iterator` functions so they don't assume that the plugins iterator is ready. They now check that the plugins iterator is there before using it, and attempt to recreate it once per request.

The reason we don't re-attempt to create the plugins iterator many times (at startup or on each request) is because we want to avoid a scenario in which two workers "compite with each other" by continuously re-creating the plugins iterator.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -849,11 +849,17 @@ local function set_init_versions_in_cache()
   return true
 end
 
+local function build_plugins_iterator_for_init_worker_phase()
+  return PluginsIterator.new_for_init_worker_phase()
+end
+
 
 -- in the table below the `before` and `after` is to indicate when they run:
 -- before or after the plugins
 return {
   build_router = build_router,
+
+  build_plugins_iterator_for_init_worker_phase = build_plugins_iterator_for_init_worker_phase,
 
   build_plugins_iterator = build_plugins_iterator,
   update_plugins_iterator = update_plugins_iterator,
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -493,22 +493,30 @@ do
                    plugins_iterator_version, timeout)
   end
 
-
-  get_updated_plugins_iterator = function()
-    if kong.configuration.router_consistency == "strict" then
-      local ok, err = rebuild_plugins_iterator(PLUGINS_ITERATOR_SYNC_OPTS)
+  local build_initial_plugins_iterator = function()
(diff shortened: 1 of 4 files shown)


kong/runloop/handler.lua
```diff
@@ -509,6 +512,15 @@ do
...
 
   get_plugins_iterator = function()
+    if not plugins_iterator then
+      local ok, err = rebuild_plugins_iterator(PLUGINS_ITERATOR_SYNC_OPTS)
+      if not ok then
+        -- If an error happens while updating, log it and return non-updated
+        -- version
+        log(ERR, "could not rebuild plugins iterator: ", err,
+                 " (stale plugins iterator will be used)")
+      end
```
bungle, line 522:
```suggestion
      build_initial_plugins_iterator()
 ```

```lua
local function build_initial_plugins_iterator()
     local ok, err = rebuild_plugins_iterator(PLUGINS_ITERATOR_SYNC_OPTS)
      if not ok then
        -- If an error happens while updating, log it and return non-updated
        -- version
        log(ERR, "could not rebuild plugins iterator: ", err,
                 " (stale plugins iterator will be used)")
      end
end
```

And same with above. Just extract that to function.

kong/runloop/handler.lua
```diff
@@ -509,6 +512,15 @@ do
 
 
   get_plugins_iterator = function()
+    if not plugins_iterator then
+      local ok, err = rebuild_plugins_iterator(PLUGINS_ITERATOR_SYNC_OPTS)
+      if not ok then
+        -- If an error happens while updating, log it and return non-updated
+        -- version
+        log(ERR, "could not rebuild plugins iterator: ", err,
+                 " (stale plugins iterator will be used)")
```
bungle, line 521: Also not sure about `stale` as we might not even have a `stale`.

kong/runloop/handler.lua
```diff
@@ -849,12 +861,49 @@ local function set_init_versions_in_cache()
...
+  end
+
+  local get_next = function(self)
+    local i = self.i + 1
+    local plugin = self.loaded[i]
+    if not plugin then
+      return nil
+    end
+    self.i = i
+    return plugin, {} -- return plugin and empty config
```
bungle, line 882:
```suggestion
    return plugin
```

PR #6024, opened 2020-06-17
gRPC tests for auth plugins

javierguerragiraldez:
### Summary

Most authentication strategies can be used on gRPC calls by providing credentials on request headers.

### Full changelog

Added tests to the following plugins:

- [x] `key-auth`
- [x] `basic-auth`
- [x] `JWT`
- [x] `hmac-auth`
- [x] `ldap-auth`
- [ ] `oauth2`

### Issues resolved

Fix #XXX


spec/03-plugins/09-key-auth/02-access_spec.lua
```diff
@@ -405,6 +419,25 @@ for _, strategy in helpers.each_strategy() do
       end)
     end)
 
+    describe("key in gRPC headers", function()
+      it("rejects call without credentials", function()
+        assert.falsy(helpers.proxy_client_grpc(){
```
gszr, line 424: Could we add an assertion for the returned code / message as well, here and in other plugins tests?

gszr: Can be done in a separate PR.

javierguerragiraldez: some plugins use JSON for the body. even if the status is appropriate (401 or similar), grpcurl can't give a readable error.  I don't think hardcoding the `Message: transport: received the unexpected content-type "application/json; charset=utf-8"` message is appropriate.  also, it's generated by grpcurl, and it could change.

hbagdi:
@gszr @javierguerragiraldez Should [protocols](https://github.com/Kong/kong/blob/master/kong/plugins/key-auth/schema.lua#L8) in the plugin schema be updated in that case?

gszr:
> @gszr @javierguerragiraldez Should [protocols](https://github.com/Kong/kong/blob/master/kong/plugins/key-auth/schema.lua#L8) in the plugin schema be updated in that case?

gRPC is a protocol under the HTTP subsystem, so it's covered by that typedef.

hbagdi:
Ah, my bad. Sorry for the noise.

PR #6033, opened 2020-06-17
fix(balancer) remove redundant call on upstream delete

murillopaula:
Remove a superfluous call to `on_target_event` when we delete an
upstream that has targets associated with it.

This ensures that the function `on_target_event` won't be called when we
delete an upstream with its targets since we already invalidate the
upstream's targets cache in the `on_upstream_event` function (also on the
db by the dao), so we shouldn't be calling `on_target_event` after that.

To reproduce:
```
http :8001/upstreams name=six
http :8001/upstreams/six/targets target=localhost:80
http DELETE :8001/upstreams/59e4b2a8-1bb1-45af-9105-672d80882e1b

2020/06/16 21:40:38 [error] 57008#0: *313 [lua] balancer.lua:670: cb(): target delete: upstream not found for 59e4b2a8-1bb1-45af-9105-672d80882e1b, client: 127.0.0.1, server: kong_admin, request: "DELETE /upstreams/59e4b2a8-1bb1-45af-9105-672d80882e1b HTTP/1.1", host: "localhost:8001"
```

The above error log is misleading as we already deleted the upstream and
its target cache at `on_upstream_event` (we also deleted it from the db).

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -322,8 +322,14 @@ local function register_events()
     local operation = data.operation
     local target = data.entity
 
-    -- => to balancer update
-    balancer.on_target_event(operation, target)
+    local upstream_id = target.upstream.id
+    local upstream_cache_key = "balancer:upstreams:" .. upstream_id
+    local upstream = core_cache:get(upstream_cache_key, nil, function() end)
+
+    if upstream then
+      -- => to balancer update
+      balancer.on_target_event(operation, target)
+    end
   end, "balancer", "targets")
 
 


kong/runloop/handler.lua
```diff
@@ -322,8 +322,14 @@ local function register_events()
     local operation = data.operation
     local target = data.entity
 
-    -- => to balancer update
-    balancer.on_target_event(operation, target)
+    local upstream_id = target.upstream.id
```
locao, line 325: Is that possible that `target.upstream` or `target.upstream.id` are somehow `nil`?

murillopaula: I don't think so. A target always has an upstream right?

locao: Yes, just checked, it's required

bungle, line 325: Is there any chance that `target` is `nil` or `target.upstream` is nil?

hutchic:
any ability to test or assert that `on_target_event` is called the right number of times in all cases?

locao:
Hi @murillopaula! Can you please the questions left in this PR? Thank you!

murillopaula:
@locao the issue here is one of the symptoms of the excessive targets cache invalidations as outlined at https://github.com/Kong/kong/pull/6336#discussion_r488670206

Let's keep discussing how to solve the root cause of this issue there at #6336 

Closing.

PR #6036, opened 2020-06-18
feat(pdk) log serializer can not be used in stream subsystem

dndx:
In #5995, we refactored the basic log serializer into a proper PDK function. @bungle kindly provided #5629 which allows stream subsystem to call the log serializer function as well, which this PR is heavily based upon. It allows various logging plugin to work under TCP and TLS proxy mode.

### DIFF
--- a/kong/pdk/log.lua
+++ b/kong/pdk/log.lua
@@ -25,6 +25,7 @@ local concat = table.concat
 local getinfo = debug.getinfo
 local reverse = string.reverse
 local tostring = tostring
+local tonumber = tonumber
 local setmetatable = setmetatable
 local ngx = ngx
 local kong = kong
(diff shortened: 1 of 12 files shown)


bungle:
@dndx are we missing these changes:
https://github.com/Kong/kong/commit/6a7a2770e05699be679eb4ce6236de72ed08903d#diff-ffbe4a2029a35f89cb0c13a63b8b5d35

I mean I don't know where we lost them, but I think it might be another `merge commit` gone wrong.

kong/pdk/log.lua
```diff
@@ -477,84 +478,141 @@ do
...
-      client_ip = var.remote_addr,
-      started_at = req.start_time() * 1000
-    }
+      return {
+        session = {
+          tls = session_tls,
+          received = tonumber(var.bytes_received, 10),
+          sent = tonumber(var.bytes_sent, 10),
+          status = ongx.status,
+          server_port = tonumber(var.server_port, 10),
```
bungle, line 596: On line `526` for `http` we have:
```lua
```

Should we have same treatment here?

PR #6037, opened 2020-06-18
perf(balancer) use leader/followers workers on eventual

murillopaula:
Execute costly code in only one worker (called the "leader", the one
with PID 0). This way, only the leader will fetch the list of upstreams
from the database. Since we're leveraging mlcache, once the leader
performs the db query and fetches the list of upstreams from the L3
callback, the other workers (called the "followers") will retrieve this
list from the L2 cache. This model should be eventually consistent.

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -18,6 +18,7 @@ local sub = string.sub
 local null = ngx.null
 local find = string.find
 local timer_at = ngx.timer.at
+local worker_pid = ngx.worker.id
 local run_hook = hooks.run_hook
 
 local CRIT  = ngx.CRIT
@@ -77,6 +78,7 @@ local noop = function() end
 -- functions forward-declarations
 local create_balancers
 local is_worker_state_stale
+local is_leader_state_updated
 local set_worker_state_stale
 local set_worker_state_updated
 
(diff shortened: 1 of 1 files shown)


kong/runloop/balancer.lua
```diff
@@ -921,7 +974,15 @@ local function init()
...
-    local _, err = timer_at(frequency, update_balancer_state)
+
+    local _, err
+
+    if worker_pid() == 0 then
+      _, err = timer_at(frequency, update_balancer_state_leader)
+    else
+      _, err = timer_at(frequency, update_balancer_state_follower)
+    end
+
```
bungle, line 985: Can we just use `timer.every` instead, like we do at (it simplifies this code a lot, as you don't need to create new timers after running them):
https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L894-L918
and
https://github.com/Kong/kong/blob/master/kong/db/strategies/postgres/connector.lua#L329

kong/runloop/balancer.lua
```diff
@@ -921,7 +974,15 @@ local function init()
 
   if kong.configuration.worker_consistency == "eventual" then
```
bungle, line 975: This is a bit unrelated, but is there reason that timers are only enabled on eventual? I mean that with `router` and `plugins` iterator we do eventual also on `strict` mode. The difference is that on `strict` we have stricter check of version on request time (e.g. if it is not the latest we will try harder to build a new one).

murillopaula: Yes, I guess we can make `strict` better in that regard. But it's out of scope from this PR, here we're trying to just improve the db performance on `eventual` only. On `strict` the db performance is very bad for users that have a big amount of upstreams in the database and/or users that make frequent upstream Admin API updates.

Biggest difference between `strict` and `eventual` on the balancer side is, AFAIK, here: https://github.com/Kong/kong/blob/next/kong/runloop/balancer.lua#L630-L646

On strict, if we receive an upstream event, we will load all the upstreams from the DB when a request comes in (if we don't receive an upstream event then all the upstreams will be in the cache already). On eventual, we will just use what’s available in the cache when a request comes in and the timer will eventually load all the upstreams from the database.

murillopaula: Some db load tests will be really needed here on this PR.

kong/runloop/balancer.lua
```diff
@@ -834,21 +836,39 @@ do
...
     kong.core_cache:invalidate(worker_state_VERSION)
+    kong.core_cache:invalidate("leader-state:version")
   end
 
 
   set_worker_state_updated = function()
     worker_state_version = kong.core_cache:get(worker_state_VERSION, TTL_ZERO, utils.uuid)
     log(DEBUG, "proxy state is updated")
+
+    if worker_pid() == 0 then
```
hishamhm, line 859: If worker 0 dies for any reason, does this number get recycled, or should a new leader get elected?

locao:
IIRC, we are not planning on using this anymore, right @murillopaula?

murillopaula:
@locao Correct. This should be addressed in the upcoming balancer eventual consistency optimizations. Closing this.

PR #6049, opened 2020-06-23
fix(changelog) asynchronous upstream updates description

bungle:
### Summary

Asynchronous is not about Admin API calls. It is about doing things not in request time but in a background only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,9 +103,22 @@
 
 ##### Core
 
-- :fireworks: **Asynchronous upstream updates**: when `worker_consistency` is
-  set to `eventual`, upstreams will not be updated in the context of Admin API calls,
-  but rather in a background timer.
+- :fireworks: **Asynchronous upstream updates**: Kong's load balancer is now able to
+  update its internal structures asynchronously instead of onto the request/stream.
+
+  This change required the introduction of new configuration properties and the
+  deprecation of older ones:
+    - New properties:
+      * `worker_consistency`
+      * `worker_state_update_frequency`
+    - Deprecated properties:
+      * `router_consistency`
+      * `router_update_frequency`
+
+  The new `worker_consistency` property is similar to `router_consistency` and accepts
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -104,8 +104,8 @@
 ##### Core
 
 - :fireworks: **Asynchronous upstream updates**: when `worker_consistency` is
-  set to `eventual`, upstreams will not be updated in the context of Admin API calls,
-  but rather in a background timer.
+  set to `eventual`, upstreams will not be updated in the context of requests
+  or streams, but rather in a background timer.
```
thibaultcha, line 108: Suggestion:

```
- :fireworks: **Asynchronous upstream updates**: Kong's load balancer is now able to
  update its internal structures asynchronously instead of onto the request/stream proxy path.

  This change required the introduction of new configuration properties and the deprecation of older ones:
    - New properties:
      * `worker_consistency`
      * `worker_state_update_frequency`
    - Deprecated properties:
      * `router_consistency`
      * `router_update_frequency`

  The new `worker_consistency` property is similar to `router_consistency` and accepts
  either of `strict` (default, synchronous) or `eventual` (asynchronous). Unlike its deprecated
  counterpart, this new property aims at configuring the consistency of *all* internal structures
  of Kong, and not only the router.
  [#5325](https://github.com/Kong/kong/pull/5325)
```

With the above suggestion, we hint at which configuration property have been affected by this change and deprecated, and we are more explicit as to what their differences are as well as their accepted values and defaults.

bungle: I followed this suggestion, bute removed the `Admin API path` at the end of first paragraph.

thibaultcha: Hmmm, "onto the request/stream" isn't very clear. Maybe "onto the request/stream path" or "during the request/stream processing"?

gszr:
Merged manually with Thibault's suggestion and commit format fix.

PR #6057, opened 2020-06-26
feat(proxy) add route.request|response_buffering

bungle:
#### Request Buffering

By default Kong does buffer the request. This is not ideal when proxying bigger payloads with HTTP 1.1 chunked encoding.

This PR adds a new property to route entity:
```
request_buffering = <true/false>
```

By default that is `true` (which is the same as before). Users can now turn the request buffering `off` by route-by-route manner by setting `route.request_buffering=false`.

#### Response Buffering

By default Kong does buffer the response. This is not ideal when upstream is sending payloads with HTTP 1.1 chunked encoding.

This PR adds a new property to route entity:
```
response_buffering = <true/false>
```

By default that is `true` (which is the same as before). Users can now turn off the response buffering on route-by-route manner by setting `route.response_buffering=false`.

### DIFF
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -866,6 +866,13 @@ return {
             the Service's `host`.
           ]]
         },
+        request_buffering = {
+          description = [[
+            Whether to enable request body buffering or not. With HTTP 1.1, it
+            may make sense to turn this off on services that receive data with
+            chunked transfer encoding.
+          ]]
+        },
         service = {
           description = [[
             The Service this Route is associated to.
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -873,6 +873,13 @@ return {
             chunked transfer encoding.
           ]]
         },
+        response_buffering = {
+          description = [[
+            Whether to enable response body buffering or not. With HTTP 1.1, it
+            may make sense to turn this off on services that send data with chunked
+            transfer encoding.
+          ]]
+        },
         service = {
           description = [[
             The Service this Route is associated to.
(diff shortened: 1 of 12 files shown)


bungle:
So the obvious discussion is: do we want this?

z-gmr:
@bungle These changes look great. I think adding the property to the route instead of the service is a better fit IMO. Otherwise, users would need to figure out a workaround if they only wanted to modify one route on a service.

locao (approved):

dlamotte:
Any tips on how to debug this feature not working? I'm attempting to do SSE (Server Sent Events) upstream of kong. I've confirmed SSE is working when directing traffic directly to the upstream. But through Kong, it fails.

I have a route that I've confirmed buffering is off (both `request_buffering` and `response_buffering` set to `false` from kong admin API; I used annotations on an HTTPRoute to achieve this), but via a custom log, I can still see the response using the `http` `kong_proxy_mode` (I expect to see `unbuffered` for `kong_proxy_mode`). I continue to get `proxy_read` timeouts and never see the actual response (just a kong error reporting that the upstream sent a 200 status code response but timed out).

```
{
  "message":"The upstream server responded with 200",
  "request_id":"687cda5763c21d304fefd70e62d05606"
}
```

And that response also has status code is 200.

Error log reported just before:
```
kong-ingress-controller-gateway-7d9469ffc9-ftx8g proxy 2024/07/23 00:08:40 [error] 1300#0: *2900 upstream timed out (110: Connection timed out) while reading upstream, client: 10.0.17.31
```

dlamotte:
Turns out, we had a custom plugin that was in the path that attempts to read the entire body broke our endpoint and it had nothing to do with this. That said, it is weird that affects the `kong_proxy_mode` for us, but I don't know if something else is the cause of that. I ended up seeing `kong_proxy_mode` being `unbuffered` until adding that plugin.

PR #6068, opened 2020-06-30
fix(balancer) update the Host header between balancer retries

dndx:
Integration test adds ~2 sec to the CI run time.


dndx:
Note: Actions failing because 8ebc64de84bc0b79650b7ec00ce10dfbb6d08e07 only made Travis point to the new branch. Can be ignored for now.

PR #6073, opened 2020-07-01
fix(declarative) keep the `ttl` definitions so that schema validator

### LINKED ISSUE #6063: 2.1 beta doesn't propagate entities with ttl from ControlPlane to DataPlane
### Summary

2.1 beta doesn't propagate entities with ttl from ControlPlane to DataPlane

### Steps To Reproduce
```
http :8001/services/ name=s81 host=httpbin.org
http --form :8001/services/s81/routes/ 'paths[]=/'
http --form :8001/services/s81/plugins/ name=key-auth
http --form :8001/consumers/ username=rai
http :8001/consumers/rai/key-auth key=rai
```

### Additional Details & Logs

- Kong 2.1.0beta.1

kong dp logs:

```
2020/06/29 16:21:35 [error] 18278#0: *2536401 [lua] clustering.lua:182: unable to update running config: bad config received from control plane in 'keyauth_credentials':
  - in entry 1 of 'keyauth_credentials':
    in 'ttl': unknown field, context: ngx.timer
```

dndx:
knows to skip trying to validate it

Otherwise entities containing the `ttl` field will cause "unknwon field"
validation failures

fixes #6063

### DIFF
--- a/kong/db/schema/others/declarative_config.lua
+++ b/kong/db/schema/others/declarative_config.lua
@@ -103,7 +103,6 @@ local function add_top_level_entities(fields, entities)
     definition.endpoint_key = nil
     definition.cache_key = nil
     definition.cache_key_set = nil
-    definition.ttl = nil
     records[entity] = definition
     add_extra_attributes(records[entity].fields, {
       _comment = true,
--- a/spec/01-unit/01-db/10-declarative_spec.lua
+++ b/spec/01-unit/01-db/10-declarative_spec.lua
@@ -1,3 +1,4 @@
+require("spec.helpers") -- for kong.log
 local declarative = require "kong.db.declarative"
 local conf_loader = require "kong.conf_loader"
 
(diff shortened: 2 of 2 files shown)


bungle:
One question, does dbless/declarative support `ttl`:
1. On control plane I push a key auth credential with `ttl` of one week
2. Dataplanes get the configuration, but do the dataplanes respect the one week ttl of key auth credential OR will it accept key auth credential "forever" (ignoring the ttl)

dndx:
@bungle I do not believe the `off` strategy has any `ttl` support currently, but we never advertised such support either I suppose.

In fact, we don't even document the use of `ttl` field inside plugin docs page such as https://docs.konghq.com/hub/kong-inc/key-auth/#create-a-key and https://docs.konghq.com/hub/kong-inc/basic-auth/#create-a-credential.

RobSerafini:
For enterprise beta I think acknowledging that `ttl` in hybrid mode does not work is ok.

PR #6075, opened 2020-07-02
support plugin to rewrite in rewrite phase

### LINKED ISSUE #6065: rewrite phase call ngx.req.set_uri failed
### Summary

we start to do some plugin develop, but we found that can not call ngx.req.set_uri function in rewrite phase.
our requirement is simple, we just want to rewrite some url.

### Steps To Reproduce

1. create a simple plugin
2. put our code into handler.lua
3. start kong
```
CCHandler.VERSION  = "1.0.0"
CCHandler.PRIORITY = 10

function CCHandler:rewrite(config)

    if kong.request.get_path() == '/get_image' then
        -- 取URL参数
        local key      = kong.request.get_query_arg('key') or '-'
        local lang     = kong.request.get_query_arg('lang') or 'en'
        local filename = kong.request.get_query_arg('filename') or 'desc.png'

        -- 读取插件配置参数
        local dic      = {}
        for _, v in ipairs(config.guide_settings) do
            local _key, _lang = string.match(v, "^([^:]+):(.*)$")
            dic[_key]         = _lang
        end
        
        -- 如果key不存在的话, 返回404
        local _lang = dic[key]
        if _lang == nil then
            return kong.response.exit(404, "404 not found",
                    {
                        ["Content-Type"] = "text/plain",
                    })
        end

        -- 读取多语言信息,如果配置项里不存在的话,使用默认值 : en
        if string.find(_lang, lang) == nil then
            lang = 'en'
        end

        -- 拼接字符串
        local _img_url = '/static/' .. key .. '/' .. lang .. '/' .. filename
        ngx.req.set_uri(_img_url, true)
    end
end

return CCHandler
```

### Additional Details & Logs

- Kong version: 2.0.4
- Operating system: macos
- Error: rewrite or internal redirection cycle while processing

we found the kong.request.get_path is same as the path before call ngx.req.set_uri, so it will cause it redirection cycle.

lecw199:
### Summary

support plugin to rewrite in rewrite phase

### Issues resolved

Fix #6065

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -1713,6 +1713,12 @@ function _M.new(routes)
         end
       end
 
+      -- support rewrite
+      -- TODO  kong.request.get_path also need to fix
+      if string.find(ngx.escape_uri(req_uri), ngx.var.uri, 1, true) ~= 1 then
+        req_uri = ngx.var.uri
+      end
+
       local match_t = find_route(req_method, req_uri, req_host, req_scheme,
                                  nil, nil, -- src_ip, src_port
                                  nil, nil, -- dst_ip, dst_port


lecw199:
---

Issue #5666, opened 2020-03-11
Route match by header regex


vanhtuan0409:
Since 1.3, It is possible to match a route by request header but it requires an exact match.
Is it possible to add regex support for header routing :D 
https://docs.konghq.com/2.0.x/proxy/#request-headers-except-host

vanhtuan0409:
added a PR for this issue. Can you have time to take a look at this @bungle :D ?

PR #6079, opened 2020-07-03
feat(router) add header regex matching method

vanhtuan0409:
### Summary

Add header regex matching method for routing decision. Based on this issue created before

https://github.com/Kong/kong/issues/5666

### Full changelog

For avoiding behavior changes, only header with 1 value and prefixed with `~*` is checked against `re_find`

Unit test added as

```
  -- 15. headers (regex)
  {
    service = service,
    route = {
      headersRegex = true,
      headers = {
        user_agent = {
          "~*windows|linux|os\\s+x\\s*[\\d\\._]+|solaris|bsd",
        },
      },
    },
  },

```

### Issues resolved

Fix https://github.com/Kong/kong/issues/5666

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -462,13 +462,22 @@ local function marshall_route(r)
 
       if header_name ~= "host" then
         local header_values_map = {}
-        for i = 1, #header_values do
+        local header_values_count = #header_values
+        for i = 1, header_values_count do
           header_values_map[lower(header_values[i])] = true
+        end
+        local header_pattern
+        if header_values_count == 1 then
+          local first_header = header_values[1]
+          if sub(first_header, 1, 2) == "~*" then
+            header_pattern = sub(first_header, 3)
+          end
         end
 
         append(headers_t, {
           name = header_name,
           values_map = header_values_map,
+          header_pattern = header_pattern,
         })
       end
     end
@@ -1033,6 +1042,12 @@ do
               matches_headers[header_t.name] = req_header_val
               break
             end
+            -- fallback to regex check if exact match failed
+            if header_t.header_pattern and re_find(req_header_val, header_t.header_pattern) then
+              found_in_req = true
+              ctx.matches.headers[header_t.name] = req_header_val
+              break
+            end
           end
 
         elseif req_header then -- string
@@ -1040,6 +1055,11 @@ do
           if header_t.values_map[req_header] then
             found_in_req = true
             matches_headers[header_t.name] = req_header
+          end
+          -- fallback to regex check if exact match failed
+          if header_t.header_pattern and re_find(req_header, header_t.header_pattern) then
+            found_in_req = true
+            ctx.matches.headers[header_t.name] = req_header
           end
         end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,9 @@
 - Customizable transparent dynamic TLS SNI name.
   Thanks, [@zhangshuaiNB](https://github.com/zhangshuaiNB)!
   [#8196](https://github.com/Kong/kong/pull/8196)
+- Routes now support matching headers with regular expressions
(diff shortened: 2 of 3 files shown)


bizhenchao1201:
Great!

sloanshang:
@vanhtuan0409  Whether this modification supports header-based regular expressions, whether it has been tested on Kubernetes and VM, and if not, whether we have considered using plugins to handle header-based regular expressions？

kong/router.lua
```diff
@@ -391,13 +391,20 @@ local function marshall_route(r)
         has_header_plain = true
 
         local header_values_map = {}
+        local header_values_count = 0
+        local header_pattern
         for i, header_value in ipairs(header_values) do
           header_values_map[lower(header_value)] = true
+          header_values_count = header_values_count + 1
+        end
+        if header_values_count == 1 and header_values[1]:find("~\\*") == 1 then
```
javierguerragiraldez, line 400: the `if s:find(pfx) == 1 then...` pattern is not the best way to check if a string starts with a prefix, especially for external input data like here.  better to use `if s:sub(1,#pfx) == pfx then...`.

Also, `"~*"` is an odd prefix to use, very easy to get wrong, as happens here.  what about `"~~"`?

```suggestion
        if header_values_count == 1 and header_values[1]:sub(1,2) == "~~" then
```

vanhtuan0409: tks @javierguerragiraldez for your suggestion. I did modify it accordingly
The reason behind `"~*"` is that Nginx location block use that annotation for regex matching

javierguerragiraldez (approved):

hbagdi:
Some questions:
What if previously configured values have "~*" in the value?
What if someone wants to do header matching based on the literal string "~*"?

And then finally, the Route entity has "regex_priority", historically used for distinguishing path regex matches and having a priority amongst them. Users could get confused now (renaming the property is not an option since it is a breaking change).
But, do header matches have a similar problem (as paths) where the regex_priority field could be helpful?

kikito:
> What if previously configured values have "*" in the value?
Those will still work. This only gets activated when there's only one rule, and the rule starts with `~*`

This begs the question: what if someone *has exactly that setup*: someone who has a route which matches headers that exactly contain a single value that begins with `~*`, like `~*potato`.

My answer to that is: we would break things for that particular person. So this change is technically backwards-breaking.

At the same time I think the odds of really impacting someone because of this are really astronomical. Matching by non-host header is already uncommon. The combination of unusual prefix + only one header really doesn't worry me.

> And then finally, the Route entity has "regex_priority", historically used for distinguishing path regex matches and having a priority amongst them. Users could get confused now (renaming the property is not an option since it is a breaking change).
But, do header matches have a similar problem (as paths) where the regex_priority field could be helpful?

The place where it would he helpful would be if someone needs to prioritize one header regex over a different header regex in a different route. I think that can be added later as an extra feature, if someone requests it. Perhaps in 3.0 we will come up with something better than regex_priority.

hbagdi:
> My answer to that is: we would break things for that particular person. So this change is technically backwards-breaking.

> At the same time I think the odds of really impacting someone because of this are really astronomical. Matching by non-host header is already uncommon. The combination of unusual prefix + only one header really doesn't worry me.

I get that practically this is okay today but I'll always be hesitant of introducing a change that limits the capability of the product.

> The place where it would he helpful would be if someone needs to prioritize one header regex over a different header regex in a different route.

That may result in a breaking change unless we say something along the lines "in case of multiple routes with regex header values, the precedence order is undefined".

kikito:
> limits the capability of the product.

I understand the concern and I thank you for reviewing this. I think as it is this PR is good enough to merge.

> That may result in a breaking change unless we say something along the lines "in case of multiple routes with regex header values, the precedence order is undefined".

That is a good point. I will make sure to include that in the docs.

We have run several performance tests for this branch and confirmed the changes had no significant performance impact, so I am merging it.

PR #6085, opened 2020-07-06
fix(correlation-id) add id on short-circuit

Tieske:
Note that by moving the priority to run before the Auth plugins, it can no longer be configured on consumers. Since by the time it runs, the consumer hasn't been identified yet.


bungle:
I don't think the `PRIORITY` needs to be changed. We should just make header filter more robust:
```lua
function CorrelationIdHandler:header_filter(conf)
  if not conf.echo_downstream then
    return
  end
  local correlation_id = kong.ctx.plugin.correlation_id or generators[conf.generator]()
  kong.response.set_header(conf.header_name, correlation_id)
end
```

There you go!

PR #6090, opened 2020-07-08
fix(api) hide workspace related internal information inside

dndx:
declarative reload responses

### DIFF
--- a/spec/02-integration/04-admin_api/15-off_spec.lua
+++ b/spec/02-integration/04-admin_api/15-off_spec.lua
@@ -367,6 +367,54 @@ describe("Admin API #off", function()
         assert.response(res).has.status(201)
       end)
 
+      it("hides workspace related fields from /config response", function()
+        local res = assert(client:send {
+          method = "POST",
+          path = "/config",
+          body = {
+            config = [[
+            _format_version: "1.1"
+            services:
+            - name: my-service
+              id: 0855b320-0dd2-547d-891d-601e9b38647f
+              url: https://example.com
+              plugins:
+              - name: file-log
+                id: 0611a5a9-de73-5a2d-a4e6-6a38ad4c3cb2
+                config:
+                  path: /tmp/file.log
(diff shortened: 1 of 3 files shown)


spec/02-integration/04-admin_api/15-off_spec.lua
```diff
@@ -367,6 +367,50 @@ describe("Admin API #off", function()
...
+        local body = assert.response(res).has.status(201)
+        local entities = cjson.decode(body)
+
+        assert.is_nil(entities.workspaces)
+        assert.is_nil(entities.consumers["4b1b701d-de2b-5588-9aa2-3b97061d9f52"].ws_id)
+        assert.is_nil(entities.keyauth_credentials["487ab43c-b2c9-51ec-8da5-367586ea2b61"].ws_id)
+        assert.is_nil(entities.plugins["0611a5a9-de73-5a2d-a4e6-6a38ad4c3cb2"].ws_id)
+        assert.is_nil(entities.plugins["661199ff-aa1c-5498-982c-d57a4bd6e48b"].ws_id)
+        assert.is_nil(entities.routes["481a9539-f49c-51b6-b2e2-fe99ee68866c"].ws_id)
+        assert.is_nil(entities.services["0855b320-0dd2-547d-891d-601e9b38647f"].ws_id)
```
hishamhm, line 411: Are these uuids deterministic across test runs, even if we run tests in isolation?

hishamhm: You can ensure determinism by adding the `id:` entries to the yaml above

hishamhm:
@dndx code change is good, left an easy-to-fix comment on the test!

spec/02-integration/04-admin_api/15-off_spec.lua
dndx: Yes. In fact if they weren't the call will cause Lua errors of referencing a field on `nil` value :)

PR #6092, opened 2020-07-08
fix(declarative) make reload atomic and properly clean the needed caches

bungle:
### Summary

Before this commit, it was really hard to understand how the `POST /config` worked with dbless and was it correct. While figuring out how to make this depend less on events, and be more atomic, I also found couple of issues.

Let me try to explain changes here.

**`declarative/init.lua`**:

This file had following code:
```lua
kong.core_cache:purge(SHADOW)
kong.cache:purge(SHADOW)
```

The function also had parameter to which `page` to load configuration, so this was for obvious reasons changed to:
```lua
kong.core_cache:purge(shadow) --using the function argument
kong.cache:purge(shadow) --using the function argument
```

So that `load` will always clear the page where it is loading the config to (that workers need to be able to clear their LRUs).

The `load_into_cache_with_events` also relied to events, both before and after flipping the page, like `delete_all` on upstreams (before flipping) and `reset` on both targets and upstreams after flipping the cache. I removed these events
and moved the code to the `flip_config` event handler to make it easier to understand the code and also to make it generate a lot fewer events. There was even a case where these events generated more events in a loop, which
also generated new lru cleanup events. So it was really quite bad.

**`runloop/handler.lua`**:

As mentioned before we moved the code from multiple events to a single place, so the `flip_config` code was changed, here is the old one:

```lua
kong.cache:flip()
core_cache:flip()
kong.default_workspace = default_ws
ngx.ctx.workspace = kong.default_workspace
```

And here is a new one:
```lua
local ok, err = concurrency.with_coroutine_mutex(FLIP_CONFIG_OPTS, function()
  balancer.stop_healthcheckers()

  kong.cache:flip()
  core_cache:flip()

  kong.default_workspace = default_ws
  ngx.ctx.workspace = kong.default_workspace

  rebuild_plugins_iterator(PLUGINS_ITERATOR_SYNC_OPTS)
  rebuild_router(ROUTER_SYNC_OPTS)

  balancer.init()

  return true
end)

if not ok then
  log(ERR, "config flip failed: ", err)
end
```

This code always runs synchronously on dbless and it does not yield -> Edit: it seems to be yielding still, :-(.

As you can see there is some code that runs `before flipping`, aka removal of healthcheckers, and deletion on balancer caches that will cause other workers to clear their LRUs. Then there is code that runs after the flipping. Previously that code relied on `reset` events and `router:version` invalidation. This made the whole flip possibly non-atomic (but I am not 100% sure of it). At least it made code very hard to follow: what happens in what order. Now the order is visible there in a single place, and it does not raise huge amounts of events (still some, but order of magnitude less than before).

I tested the code against memory leaks with multiple workers and multiple simultaneous uploads of new configs. But it worked as expected, and I saw no memory been leaking.

**`cache.lua`**:

Just some style changes as I tried to understand the code better.

**`init.lua`**:

While doing this I also spotted another issue on reloads that happened in `reset_kong_shm` function. I added more cache entries to be reserved and also made it more intelligent to detect which page it should keep and which one to clear on `reload`. Some of that `reset`ing is unnecessary in some scenarios, but on those scenarios it is more than likely that the `shms` that it tries to reset are already empty, so the impact of resetting them will be small. And well, we did it before these additions too, it justs preserves and resets more.

Reviewing this, you need to have your head wrapped to:
1. kong can run on dbless and with db
2. reload/restart may happen on dbless and with db
3. reload may contain a new declarative configuration
4. we need to be extra careful to not leak memory in worker level LRUs or SHMs on configuration changes, also both L1 and L2 caches need to be cleared properly  across all the workers so that old configs don't pop-up from those. Also when reloading dbless nodes without explicit declarative configuration, the nodes need to preserve their current caches.

Thanks @dndx and @locao for helping me on this.

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -504,7 +504,7 @@ end
 --     _format_version: "2.1",
 --     _transform: true,
 --   }
-function declarative.load_into_cache(entities, meta, hash, shadow_page)
+function declarative.load_into_cache(entities, meta, hash, shadow)
   -- Array of strings with this format:
   -- "<tag_name>|<entity_name>|<uuid>".
   -- For example, a service tagged "admin" would produce
@@ -522,8 +522,8 @@ function declarative.load_into_cache(entities, meta, hash, shadow_page)
   -- but filtered for a given tag
   local tags_by_name = {}
 
-  kong.core_cache:purge(SHADOW)
-  kong.cache:purge(SHADOW)
+  kong.core_cache:purge(shadow)
+  kong.cache:purge(shadow)
 
   local transform = meta._transform == nil and true or meta._transform
 
@@ -597,13 +597,13 @@ function declarative.load_into_cache(entities, meta, hash, shadow_page)
         end
       end
 
-      local ok, err = kong.core_cache:safe_set(cache_key, item, shadow_page)
+      local ok, err = kong.core_cache:safe_set(cache_key, item, shadow)
       if not ok then
         return nil, err
       end
 
       local global_query_cache_key = dao:cache_key(id, nil, nil, nil, nil, "*")
-      local ok, err = kong.core_cache:safe_set(global_query_cache_key, item, shadow_page)
+      local ok, err = kong.core_cache:safe_set(global_query_cache_key, item, shadow)
       if not ok then
         return nil, err
       end
@@ -620,7 +620,7 @@ function declarative.load_into_cache(entities, meta, hash, shadow_page)
 
       if schema.cache_key then
         local cache_key = dao:cache_key(item)
-        ok, err = kong.core_cache:safe_set(cache_key, item, shadow_page)
+        ok, err = kong.core_cache:safe_set(cache_key, item, shadow)
         if not ok then
           return nil, err
         end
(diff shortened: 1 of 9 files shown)


bungle:
There are still some cases where there are concurrency errors, but to get them to happen you really need to push multiple different configs, and even then it happens (pretty sure) just for a few possible requests. I have plans to make it even better but those changes are going to be bigger. This should make situation already a lot better.

kong/db/declarative/init.lua
```diff
@@ -754,21 +754,14 @@ end
 
 
 function declarative.load_into_cache_with_events(entities, meta, hash)
-
-  -- ensure any previous update finished (we're flipped to the latest page)
-  local ok, err = kong.worker_events.poll()
-  if not ok then
-    return nil, err
-  end
-
```
hishamhm, line 763: I believe this block may still be necessary, because the `purge` operation inside `load_into_cache` happens before the worker event is triggered.

With this block added back, then this PR is good to go for 2.1.0-rc1!

hishamhm: Of course, we can make further progress later, but this PR is a definite improvement.

PR #6058, opened 2020-06-26
tests(migrations) migration unit tests

kikito:
Introduces a new way to test migrations. Aiming for 80%-20% rule. More details in individual commits.

PR #6097, opened 2020-07-09
tests/upgrade tests for core/009 (postgres only)

kikito:
Uses #6058 as a base, adding more tests.


kikito:
Made part of #6129, closing this PR.

PR #6107, opened 2020-07-14
feat(deps) support OpenResty 1.17.8.x

thibaultcha:
Sibling PR: https://github.com/Kong/kong-build-tools/pull/309


hishamhm:
> In OpenResty 1.17.8.1, `coroutine.wrap()` was updated to follow the Lua 5.1 specifications and now propagates runtime errors to its parent coroutine. However, the OpenResty monkey-patching of coroutines prevents `pcall` from catching errors thrown from within the wrapped coroutine.

@thibaultcha Are there any other breaking changes we should be aware of that could impact end-users?

thibaultcha:
@hishamhm Good question, I should have addressed that! The main changes we should shed a light upon are:

- `ngx.req.set_uri_args` now escapes whitespace and control characters (https://github.com/openresty/lua-nginx-module/commit/785fb57094e3dbb161962875800e9cf92ce3717b#diff-2cc6436621ca541395faed3bc305b8dfR30)
- `ngx.req.set_header` also escapes those as per RFC 7230 (https://github.com/openresty/lua-nginx-module/commit/e1e335fd7f72bf2f1dad20800c33f79ff3e0d1c4#diff-0a369498a5a8db3ac8fa606b544c9810R4886-R4889)
- users of `ngx.location.capture` should upgrade to patch the request smuggling vulnerability
- we removed many CFunctions that weren't being used anymore since we loaded lua-resty-core and its FFI implementation since 1.15.8.1. This shouldn't be an issue _unless_ some APIs slipped through... (see https://github.com/openresty/lua-resty-core/pull/306 - I added this patch to our openresty patches)

thibaultcha:
There are also many features we could take advantage of in in Kong itself to reduce workarounds we had to implement in the past, but here I chose not to for the time being in order to maintain compatibility with 1.15.8.x and make the bump smoother for contributors.

hutchic (approved):

thibaultcha:
@hutchic Just a note here that commits in this PR should *not* have been squashed when merging.

PR #6110, opened 2020-07-14
Perf/2.1.0

javierguerragiraldez:
performance tweaks

### DIFF
--- a/kong/pdk/init.lua
+++ b/kong/pdk/init.lua
@@ -303,7 +303,7 @@ function _PDK.new(kong_config, major_version, self)
   return setmetatable(self, {
     __index = function(t, k)
       if k == "core_log" then
-        return rawget(t, "_log")
+        return (rawget(t, "_log"))
       end
 
       if k == "log" then
@@ -311,7 +311,7 @@ function _PDK.new(kong_config, major_version, self)
           return t.ctx.core.log
         end
 
-        return rawget(t, "_log")
+        return (rawget(t, "_log"))
       end
     end
   })
--- a/kong/runloop/plugins_iterator.lua
+++ b/kong/runloop/plugins_iterator.lua
@@ -298,6 +298,10 @@ local function get_next(self)
   return get_next(self) -- Load next plugin
 end
 
+local function zero_iter()
+  return nil
+end
+
 
 local PluginsIterator = {}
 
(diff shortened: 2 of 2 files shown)


kong/init.lua
```diff
@@ -950,10 +966,10 @@ function Kong.balancer()
...
     end
 
-    ngx_log(ngx_DEBUG, "enabled connection keepalive (pool=", pool_opts.pool,
-                       ", pool_size=", pool_opts.pool_size,
-                       ", idle_timeout=", kong_conf.upstream_keepalive_idle_timeout,
-                       ", max_requests=", kong_conf.upstream_keepalive_max_requests, ")")
+    --ngx_log(ngx_DEBUG, "enabled connection keepalive (pool=", pool_opts.pool,
+    --                   ", pool_size=", pool_opts.pool_size,
+    --                   ", idle_timeout=", kong_conf.upstream_keepalive_idle_timeout,
+    --                   ", max_requests=", kong_conf.upstream_keepalive_max_requests, ")")
```
thibaultcha, line 972: Such logs should really not be disabled imho... Do we have empirical data that demonstrate a performance impact? Ditto for all other debug logs commented out in this PR.

bungle: The cluster events log in annoying when debugging. Commenting this out made on my machine, made it more performant, almost to the same level as previous. I ran tests with DEBUG on 2010 Mac :-).

I am not sure if these changes were worth it, thus it was just for sharing between me and @javierguerragiraldez. I am fine with whatever you decide, even removing all my `tries` (they didn't make a huge difference anyway). Still we are slower than before I think.

thibaultcha: > The cluster events log is annoying when debugging (I have never had any use for it)

Granted, these can be commented imho, although they have been useful for debugging cache invalidation issues in the past, especially reviewing logs from users and customers. But again, I'm not opposed to commenting these out.

> I ran tests with DEBUG on 2010 Mac 

Is it reasonable to optimize the performance of debug builds and configurations? What I meant above was that we should make sure that these changes come with significant performance improvements for **non-debug** builds and configurations, because while I doubt that these debug logs have any significant impact in these cases, they certainly are extremely helpful to us when debugging issues. If anything, we should add *more* debugging logs, not less.

thibaultcha: Especially considering that the dynamic keepalive pooling is new, and not having any debugging logs at the time of its first release is quite a bet to make.

bungle: yes, I am fine with the whole commit removal:
https://github.com/Kong/kong/pull/6110/commits/b82021b3a8c5d9e3565a6d427aa2dd98bc12067e

javierguerragiraldez: removed

PR #6121, opened 2020-07-16
fix(balancer) make balancer flip config atomic by waiting for workers to finish

bungle:
### Summary

With this patch the same node can receive new config at `:8001/config` only one at the time, and it waits for other workers to flip before allowing a new config to be uploaded. It makes `:8001/config` endpoint essentially serializable.

There is 60 seconds timeout, so that it will not be held forever.

@hishamhm knows that I tried this on the previous atomic (#6092), but as it was not causing issues anymore with router and plugins iterator, I removed it. Further testing showed that balancer has still issues with this. So I set up the flip-serializer again.

### DIFF
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -97,14 +97,26 @@ return {
         return declarative.load_into_cache_with_events(entities, meta, new_hash)
       end)
 
-      if err == "no memory" then
-        kong.log.err("not enough cache space for declarative config")
-        return kong.response.exit(413, {
-          message = "Configuration does not fit in Kong cache"
-        })
-      end
-
       if not ok then
+        if err == "busy" or err == "locked" then
+          return kong.response.exit(429, {
+            message = "Currently loading previous configuration."
+          }, { ["Retry-After"] = 60 })
+        end
+
+        if err == "timeout" then
+          return kong.response.exit(504, {
+            message = "Timed out while loading configuration."
+          })
+        end
+
+        if err == "no memory" then
+          kong.log.err("not enough cache space for declarative config")
+          return kong.response.exit(413, {
+            message = "Configuration does not fit in Kong cache"
+          })
+        end
+
         kong.log.err("failed loading declarative config into cache: ", err)
         return kong.response.exit(500, { message = "An unexpected error occurred" })
       end
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -1,5 +1,4 @@
 local declarative = require("kong.db.declarative")
-local concurrency = require("kong.concurrency")
 local reports = require("kong.reports")
 local errors = require("kong.db.errors")
 
@@ -93,15 +92,13 @@ return {
         return kong.response.exit(400, errors:declarative_config(err_t))
       end
 
-      local ok, err = concurrency.with_worker_mutex({ name = "dbless-worker" }, function()
-        return declarative.load_into_cache_with_events(entities, meta, new_hash)
-      end)
+      local ok, err, ttl = declarative.load_into_cache_with_events(entities, meta, new_hash)
 
       if not ok then
         if err == "busy" or err == "locked" then
           return kong.response.exit(429, {
             message = "Currently loading previous configuration."
-          }, { ["Retry-After"] = 60 })
+          }, { ["Retry-After"] = ttl or 60 })
         end
 
         if err == "timeout" then
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -97,13 +97,13 @@ return {
       if not ok then
         if err == "busy" or err == "locked" then
           return kong.response.exit(429, {
-            message = "Currently loading previous configuration."
-          }, { ["Retry-After"] = ttl or 60 })
+            message = "Currently loading previous configuration"
+          }, { ["Retry-After"] = ttl })
         end
 
         if err == "timeout" then
           return kong.response.exit(504, {
-            message = "Timed out while loading configuration."
+            message = "Timed out while loading configuration"
           })
         end
 
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -756,6 +756,10 @@ end
 
 
 function declarative.load_into_cache_with_events(entities, meta, hash)
+  if ngx.shared.kong:get("declarative:flips") then
+    return nil, "busy"
+  end
+
   -- ensure any previous update finished (we're flipped to the latest page)
   local ok, err = kong.worker_events.poll()
   if not ok then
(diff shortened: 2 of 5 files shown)


kong/db/declarative/init.lua
```diff
@@ -802,6 +802,38 @@ function declarative.load_into_cache_with_events(entities, meta, hash)
     return nil, "failed to persist cache page number inside shdict: " .. err
   end
 
+  local worker_count = ngx.worker.count()
```
dndx, line 805: This data doesn't change until reloads, so we can cache it at module level above.

kong/db/declarative/init.lua
```diff
@@ -802,6 +802,38 @@ function declarative.load_into_cache_with_events(entities, meta, hash)
...
+  local sleep_left = 60
+  local sleep_time = 0.0375
```
dndx, line 807: Might be a dumb question... But where does the magic number come from?

kong/db/declarative/init.lua
```diff
@@ -802,6 +802,38 @@ function declarative.load_into_cache_with_events(entities, meta, hash)
...
+
+  while true do
```
dndx, line 809:
```suggestion
  while sleep_left > 0 do
```

kong/db/declarative/init.lua
```diff
@@ -802,6 +802,38 @@ function declarative.load_into_cache_with_events(entities, meta, hash)
...
+    local flips = ngx.shared.kong:get("declarative:flips")
```
dndx, line 810: Since this key is set inside the `flip_config` worker event and it is executed asynchronously. Is there a possibility that this may return `nil` because the worker event haven't been executed by any other workers yet?

kong/db/declarative/init.lua
```diff
@@ -802,6 +802,38 @@ function declarative.load_into_cache_with_events(entities, meta, hash)
...
+    if flips == nil then
+      break
+    end
+
+    if flips == worker_count then
+      break
+    end
+
+    if sleep_left == 0 then
```
dndx, line 819: We can remove this and move it to the `while` condition as commented above.

Also, from a defensive programming perspective, we usually want to use `sleep_left <= 0` instead of `==`.

kong/db/declarative/init.lua
bungle: Sleeps are:
75 ms
150 ms
300 ms
600 ms
1200 ms
2400 ms
4800 ms
...

worker events are polled by default once per second. On my testing it was quite hood start number. It comes from slight testing but in reality it could be anything else.

kong/db/declarative/init.lua
bungle: The post above will execute it right away on current worker. Other workers process it via timer. post is actually post+poll.

kong/db/declarative/init.lua
bungle: the code below does not allow it to go negative.

kong/db/declarative/init.lua
```diff
@@ -802,6 +803,31 @@ function declarative.load_into_cache_with_events(entities, meta, hash)
...
   end
 
+  ngx.shared.kong:add("declarative:flips", 0, 60)
+
+  local sleep_left = 60
+  local sleep_time = 0.0375
+
+  while sleep_left > 0 do
+    local flips = ngx.shared.kong:get("declarative:flips")
+    if  flips == nil or flips == WORKER_COUNT then
```
javierguerragiraldez, line 813: if i follow this correctly, the `== WORKER_COUNT` part is to avoid the case where _all_ workers are waiting for each other, right?  would it be safer to use `=>` ?

bungle: In short, == WORKER_COUNT is the normal expacted state. nil and returning early are just unexpected timeouts.

It is used to make worker that executed this code wait for all the workers to process the flip event. Which is equal to number of workers. I am not sure if there can ve more. Crashing workers, or simething unexpected perhaps? Worker skipping a worker event? Leaking lock? We can certainly change this.

kong/api/routes/config.lua
```diff
@@ -97,14 +97,26 @@ return {
...
-        return kong.response.exit(413, {
-          message = "Configuration does not fit in Kong cache"
-        })
-      end
-
       if not ok then
+        if err == "busy" or err == "locked" then
+          return kong.response.exit(429, {
+            message = "Currently loading previous configuration."
+          }, { ["Retry-After"] = 60 })
```
bungle, line 104: Can we somehow get `ttl` of the lock, and say set this to time when the lock will timeout, or is it possible that lock is held forever?

bungle: We can leave this later.

bungle:
@javierguerragiraldez,

please check if the API blocks, or return immediately?

See:

https://github.com/Kong/kong/blob/fix/flip-config-atomic/kong/api/routes/config.lua#L96
and:
https://github.com/Kong/kong/blob/fix/flip-config-atomic/kong/concurrency.lua#L26-L37

wouldn’t that make it wait for 60 secs or how does that work?

What I think should be done is to move that worker mutex to that declarative.load_into_cache_with_events and have exptime (immediately) though our concurrency sets that to timeout too.

Also it would be plus if we could have dynamic `Retry-After` based on TTL of the running worker, if possible, but that can be moved to later.

javierguerragiraldez:
removed the outer lock and replaced by better atomic check of the shared counter.   required explicit removal on error returns (at least those returns before posting the event to other workers).

bungle:
I cannot `approve` @javierguerragiraldez work as the PR is still under my name. @javierguerragiraldez the only remaining thing in this PR as I see is the `timeout`. It is now set to `60` secs. Should it be shorter? In normal situations it is rather short, but on those unexpected situations it may block the `/config` api for full `60` secs.

kong/api/routes/config.lua
```diff
@@ -93,18 +92,28 @@ return {
...
       if not ok then
+        if err == "busy" or err == "locked" then
+          return kong.response.exit(429, {
+            message = "Currently loading previous configuration."
+          }, { ["Retry-After"] = ttl or 60 })
+        end
+
+        if err == "timeout" then
+          return kong.response.exit(504, {
+            message = "Timed out while loading configuration."
```
dndx, line 106:
```suggestion
            message = "Timed out while loading configuration"
```

to be consistent.

kong/db/declarative/init.lua
```diff
@@ -755,9 +756,20 @@ end
 
 
 function declarative.load_into_cache_with_events(entities, meta, hash)
+  local ok, err = ngx.shared.kong:add("declarative:flips", 0, 60)
```
dndx, line 759: Better make the magic number `60` a module variable and export it, so that https://github.com/Kong/kong/pull/6121/files#diff-35cfffb01c3d2bebf06bab3805ed22c1R101 and https://github.com/Kong/kong/pull/6121/files#diff-35cfffb01c3d2bebf06bab3805ed22c1R101 can use it as well.

kong/db/declarative/init.lua
```diff
@@ -802,6 +817,30 @@ function declarative.load_into_cache_with_events(entities, meta, hash)
...
+  while sleep_left > 0 do
+    local flips = ngx.shared.kong:get("declarative:flips")
+    if  flips == nil or flips == WORKER_COUNT then
+      break
+    end
+
+    sleep_time = sleep_time * 2
+    if sleep_time > sleep_left then
+      sleep_time = sleep_left
+    end
```
dndx, line 832:
```suggestion
    end

```

dndx (approved):

PR #6129, opened 2020-07-17
tests(migrations) migration tests for 140_to_150 and 200_to_210

kikito:
This PR aglutinates several things into one.

* Introduces a way to partially migrate the database, and introduces a couple helper functions
* Adds "unit tests" for the core/007 migration, as a "proof of concept"
* Adds "unit tests" for the 200_to_210 migrations, for core and for the plugins

This way of testing has two big disadvantages:
* We are _testing mocks_ here. We are not testing how Kong behaves, but how *we think* kong behaves, by putting what we _think_ it does in the database, then running the migrations, and checking what we _think_ should be in the database is there.
* We are not capable of testing what Kong does on the DAO-level in order to work in tandem with the migrations; only direct changes to the database are testable.

A much more thorough way to do this would be to instantiate one (or several!) nodes of kong in 2.0.0 and 2.1.0, and use the Admin API exclusively to do tests. This is in fact what we have done in the past, in https://github.com/Kong/kong-upgrade-tests. But this way of testing had problems of its own:
* It required pushing code to a branch in order to test migrations
* Tests run very slowly and had lots of infra dependencies (gojira, which in turn needs docker)

In contrast, this way of testing:
* Is very fast to run (seconds vs minutes).
* You can write tests while you write migrations, no change of environment, writing tests as you are used to.
* Fast feedback as a result.
* Helps avoiding "stupid mistakes" (e.g. a typo in a column name) quickly.
* Can be used in conjunction with the other tests. In fact, since you don't need to write all migration tests as integration tests any more (no need to test that a column was added), it makes the migration tests faster to run.

### DIFF
--- a/spec/03-plugins/10-basic-auth/00-migrations/003_200_to_210_spec.lua
+++ b/spec/03-plugins/10-basic-auth/00-migrations/003_200_to_210_spec.lua
@@ -1,0 +1,128 @@
+local helpers = require "spec.helpers"
+local h = require "spec.migration_helpers.200_to_210"
+local utils = require "kong.tools.utils"
+
+local fmt = string.format
+
+describe("#db migration basic-auth/003_200_to_210 spec", function()
+  local _, db
+
+  after_each(function()
+    -- Clean up the database schema after each exercise.
+    -- This prevents failed migration tests impacting other tests in the CI
+    assert(db:schema_reset())
+  end)
+
+  describe("#postgres", function()
+    before_each(function()
+      _, db = helpers.get_db_utils("postgres", nil, nil, {
+        stop_namespace = "kong.plugins.basic-auth.migrations",
+        stop_migration = "003_200_to_210",
+      })
+    end)
+
+    it("adds and sets ws_id", function()
+      local cn = db.connector
+      h.assert_not_pg_has_fkey(cn, "basicauth_credentials", "ws_id")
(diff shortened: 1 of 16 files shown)


spec/03-plugins/10-basic-auth/00-migrations/003_200_to_210_spec.lua
```diff
@@ -0,0 +1,116 @@
...
+      -- create specifying default ws_id.(simulate new node)
+      local a = h.c_insert(cn, "basicauth_credentials", { id = utils.uuid(), ws_id = default_ws_id })
+      assert.equals(default_ws_id, a.ws_id)
+
+      -- kong migrations finish
+      assert(helpers.run_teardown_migration(db, "basic-auth",
+                                            "kong.plugins.basic-auth.migrations",
+                                            "003_200_to_210"))
+
+      -- create specifying default ws_id.(simulate new node)
```
gszr, line 111: Let's add a test that entities created by old nodes receive a ws_id after teardown (applies to other plugins tests as well)

kikito:
All requests done. Please re-review

kikito:
We decided not to merge this in CE out of several concerns, but mostly that this approach, while useful in finding some bugs, is only as good as its mocks. For now it shall remain unmerged.

PR #6139, opened 2020-07-21
docs(clustering) add settings defined by Hybrid Mode PKI

javierguerragiraldez:
Document new settings for Hybrid Mode PKI

Shamelessly copied from https://github.com/Kong/docs.konghq.com/compare/docs/hybrid-mode-pki

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -195,7 +195,7 @@
                                  # between control and data plane nodes.
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
-                                 # Under `shared` mode, is must be the same
+                                 # Under `shared` mode, it must be the same
                                  # for all nodes.  Under `pki` mode it
                                  # should be a different certificate for each
                                  # DP node.
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -205,7 +205,7 @@
                                  # between control and data plane nodes.
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
-                                 # Under `shared` mode, is must be the same
+                                 # Under `shared` mode, it must be the same
                                  # for all nodes.  Under `pki` mode it
                                  # should be a different certificate for each
                                  # DP node.
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -184,11 +184,12 @@
                                  #   Note that CP and DP nodes have to present
                                  #   the same certificate to establish mTLS
                                  #   connections.
-                                 # - `pki`: use `cluster_ca_cert` and
-                                 #   `cluster_server_name` for verification,
-                                 #   these are different certificates for each
+                                 # - `pki`: use `cluster_ca_cert`,
+                                 #   `cluster_server_name` and `cluster_cert`
+                                 #   for verification.
+                                 #   These are different certificates for each
                                  #   DP node, but issued by a cluster-wide
-                                 #   common CA certificate: `cluster_ca_cert`
+                                 #   common CA certificate: `cluster_ca_cert`.
 
 #cluster_cert =                  # Filename of the cluster certificate to use
                                  # when establishing secure communication
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -173,17 +173,56 @@
                                  #   It runs DB-less and receives configuration
                                  #   updates from a control plane node.
 
+#cluster_mtls = shared           # Sets the verification between nodes of the
+                                 # cluster.
+                                 #
+                                 # Valid values to this setting are:
+                                 #
+                                 # - `shared`: use a shared certificate/key
+                                 #   pair specified with the `cluster_cert`
+                                 #   and `cluster_cert_key` settings.
+                                 #   Note that CP and DP nodes have to present
+                                 #   the same certificate to establish mTLS
+                                 #   connections.
+                                 # - `pki`: use `cluster_ca_cert` and
+                                 #   `cluster_server_name` for verification,
+                                 #   these are different certificates for each
+                                 #   DP node, but issued by a cluster-wide
+                                 #   common CA certificate: `cluster_ca_cert`
+
 #cluster_cert =                  # Filename of the cluster certificate to use
                                  # when establishing secure communication
                                  # between control and data plane nodes.
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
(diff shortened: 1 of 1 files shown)


kong.conf.default
```diff
@@ -173,6 +173,15 @@
                                  #   It runs DB-less and receives configuration
                                  #   updates from a control plane node.
 
+#cluster_mtls = shared           #
+                                 # Valid values to this setting are:
+                                 #
+                                 # - `shared`: use a shared certificate/key
+                                 #   pair specified with the `cluster_cert`
+                                 #   and `cluster_cert_key` settings.
```
dndx, line 181: We should add "In this mode the CP and DP nodes are required to present the same certificate to establish mTLS connection with each other." to explain what this setting means for the user.

kong.conf.default
```diff
@@ -185,6 +194,15 @@
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
 
+#cluster_ca_cert =               # The trusted CA certificate file in PEM
+                                 # format used to verify the `cluster_cert`.
+                                 # Required if `cluster_mtls` is set to `pki`.
+
+#cluster_server_name =           # The Server Name used in the DP/CP TLS
```
dndx, line 201: "The server name used when DP TLS handshakes with CP. Sent to CP as SNI server name extension value as well as validating the CP cert returned. Must match the Common Name (CN) or Subject Alternative Name (SAN) inside the CP certificate."

or something like this should explain the use case for the server name property better.

kong.conf.default
```diff
@@ -173,6 +173,15 @@
...
+#cluster_mtls = shared           #
+                                 # Valid values to this setting are:
+                                 #
+                                 # - `shared`: use a shared certificate/key
+                                 #   pair specified with the `cluster_cert`
+                                 #   and `cluster_cert_key` settings.
+                                 # - `pki`: use `cluster_ca_cert` and
+                                 #   `cluster_server_name` for verification.
+
 #cluster_cert =                  # Filename of the cluster certificate to use
```
dndx, line 185: We should add for `cluster_cert` and `cluster_cert_key` that if `cluster_mtls = pki`, then this is the cluster CA issued DP cert presented to the CP during mTLS handshake and should be a different cert for each DP node.

kong.conf.default
```diff
@@ -173,6 +173,15 @@
...
                                  #   updates from a control plane node.
 
+#cluster_mtls = shared           #
+                                 # Valid values to this setting are:
+                                 #
+                                 # - `shared`: use a shared certificate/key
+                                 #   pair specified with the `cluster_cert`
+                                 #   and `cluster_cert_key` settings.
+                                 # - `pki`: use `cluster_ca_cert` and
+                                 #   `cluster_server_name` for verification.
```
dndx, line 183: Similiarly, we should add "In this mode the CP and DP uses different certificate, but issued under a common cluster CA".

javierguerragiraldez:
> Let's rebase this onto `master` (easier to just reset --hard to master and cherry-pick the commit actually) so that it goes into 2.1.1.

`git rebase next --onto master`

kong.conf.default
```diff
@@ -173,17 +173,49 @@
...
+                                 #   these are different certificates for each
+                                 #   DP node, but issued by a cluster-wide
+                                 #   common CA certificate: `cluster_ca_cert`
+
 #cluster_cert =                  # Filename of the cluster certificate to use
                                  # when establishing secure communication
                                  # between control and data plane nodes.
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
+                                 # Should be a different certificate for each
```
dndx, line 196: This is only true if `cluster_mtls = pki`. Otherwise DP and CP must use the same cert across the cluster.

kong.conf.default
```diff
@@ -173,17 +173,49 @@
...
                                  # use when establishing secure communication
                                  # between control and data plane nodes.
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
+                                 # Should be a different certificate for each
+                                 # DP node.
+
+#cluster_ca_cert =               # The trusted CA certificate file in PEM
+                                 # format used to verify the `cluster_cert`.
+                                 # Required if `cluster_mtls` is set to `pki`.
```
dndx, line 209:
```suggestion
                                 # Ignored if `cluster_mtls` is not set to `pki`.
```

Might be a more accurate description because we don't use this value in `shared` mode.

kong.conf.default
```diff
@@ -173,17 +173,56 @@
...
+                                 #   these are different certificates for each
+                                 #   DP node, but issued by a cluster-wide
+                                 #   common CA certificate: `cluster_ca_cert`
+
 #cluster_cert =                  # Filename of the cluster certificate to use
                                  # when establishing secure communication
                                  # between control and data plane nodes.
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
+                                 # Under `shared` mode, is must be the same
```
dndx, line 198:
```suggestion
                                 # Under `shared` mode, it must be the same
```

kong.conf.default
```diff
@@ -173,17 +173,56 @@
...
+                                 # for all nodes.  Under `pki` mode it
+                                 # should be a different certificate for each
+                                 # DP node.
 
 #cluster_cert_key =              # Filename of the cluster certificate key to
                                  # use when establishing secure communication
                                  # between control and data plane nodes.
                                  # You can use the `kong hybrid` command to
                                  # generate the certificate/key pair.
+                                 # Under `shared` mode, is must be the same
```
dndx, line 208:
```suggestion
                                 # Under `shared` mode, it must be the same
```

kong.conf.default
```diff
@@ -173,17 +173,56 @@
...
+                                 #
+                                 # Valid values to this setting are:
+                                 #
+                                 # - `shared`: use a shared certificate/key
+                                 #   pair specified with the `cluster_cert`
+                                 #   and `cluster_cert_key` settings.
+                                 #   Note that CP and DP nodes have to present
+                                 #   the same certificate to establish mTLS
+                                 #   connections.
+                                 # - `pki`: use `cluster_ca_cert` and
```
dndx, line 187: Recommendation: "`pki`: use `cluster_ca_cert`, `cluster_server_name` and `cluster_cert` for verification,"  should be more accurate. That way the user knows exactly what needs to be set in the `pki` mode.

PR #6140, opened 2020-07-21
tests(dbless) add regression test for DB-less plugin iterator build f…

dndx:
…ailure

"plugins iterator was changed while rebuilding it" from a6a3080

This test was not included during the Workspaces refactor, however it is
still efficient at catching such bugs in the future. Therefore it is
picked back to the main branch.


dndx:
@gszr Since the 2.1 release is already out I think we don't need this on `master` immediately. Merging it on `next` makes sense since that is where the development happens and this test is trying to catch bugs caused by big changes anyway.

PR #6150, opened 2020-07-24
fix(rate-limit) perform atomic redis increments

### LINKED ISSUE #5763: Kong rate-limtiing plugin is not setting ttl to some keys when using redis

javierguerragiraldez:
The increment function of the Redis policy first gathers which relevant
keys actually exist, and then executes all the INCRBY/EXPIRE operations.
This has a window of error between the existence check and the
(optional) EXPIRE operation.

A solution documented in the Redis documentation
(https://redis.io/commands/incr) is to use a Lua operation in the Redis
server, which operates atomically.

Fixes #5763

### DIFF
--- a/kong/plugins/rate-limiting/policies/init.lua
+++ b/kong/plugins/rate-limiting/policies/init.lua
@@ -51,6 +51,52 @@ local sock_opts = {}
 local EXPIRATION = require "kong.plugins.rate-limiting.expiration"
 
 
+local function get_redis_connection(conf)
+  local red = redis:new()
+  red:set_timeout(conf.redis_timeout)
+  -- use a special pool name only if redis_database is set to non-zero
+  -- otherwise use the default pool name host:port
+  sock_opts.pool = conf.redis_database and
+                    conf.redis_host .. ":" .. conf.redis_port ..
+                    ":" .. conf.redis_database
+  local ok, err = red:connect(conf.redis_host, conf.redis_port,
+                              sock_opts)
+  if not ok then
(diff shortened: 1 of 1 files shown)


javierguerragiraldez:
> Needs to target `master`

done

javierguerragiraldez:
> Would it be possible to add a regression test?

i'd like to, but the vulnerability is about heavy concurrency (as usual on race conditions).  talk about #flaky tests.

a quick search (mostly on stackoverflow) about "testing for race conditions", yields some ideas about adding checks for preserving invariants at every point, using Clang Thread Sanitizer (which observes the process during execution)... but they all fail when the race appears across a client/server connection, and the solution (in this case at least) is just pushing the whole problem to the server side (which we don't compile/test).

PR #6153, opened 2020-07-24
docs(changelog) add clarification on OpenResty version bump

### LINKED ISSUE #6151: Consistency in the CHANGE log will help build from src folk
### Summary

I think it is important to mention when versions of important underlying dependencies (aka nginx or openresty) get bumped, that the CHG log reflects it as a standard somewhere.

For those who build from source, the CHG log and Upgrade docs are some of the most important guides to ensure smooth transitions and understanding whats changed.

EX: 

2.0 mentions the OpenResty Bump: https://github.com/Kong/kong/blob/master/CHANGELOG.md#200 , this is helpful

2.1 lacks the OpenResty Bump mention:
https://github.com/Kong/kong/blob/master/CHANGELOG.md#210

Which if you want to take advantage of the newer upstream keepalive behavior, you need to be on OpenResty ```1.15.8.3``` as thats where your patches have been applied in your build process:
https://github.optum.com/APIGateway/kong-build-tools/tree/dev/openresty-patches/patches/1.15.8.3 

vs the 2.0 ```1.15.8.2```

Otherwise your logs will get something like this and then someone like me needs to dig a bit further and find out the OpenResty bump was necessary with a second lookover:

```
2020/07/24 00:00:00 [warn] 45#0: [lua] init.lua:114: missing method 'ngx_balancer.enable_keepalive()' (was the dyn_upstream_keepalive patch applied?) set the 'nginx_upstream_keepalive' configuration property instead of 'upstream_keepalive_pool_size'
```

Appreciate whoever dropped in the error log that helped let me know I was missing something regardless which initiated my investigation 😀 .

Another location I managed to confirm the needs for 2.1 was simply the latest build from src docs here: https://docs.konghq.com/install/source/ mentioning the newest OpenResty version too. But I normally would never hunt over here. 

### Additional Details & Logs
- Kong version 2.1

hishamhm:
Fixes #6151 

@jeremyjpj0916 please take a look, let me know if this is what you'd expect to find there!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,18 @@
 
 ### Dependencies
 
+### Dependencies
+
+- :warning: In order to use all Kong features, including the new
+  dynamic upstream keepalive behavior, the required OpenResty version is
+  [1.15.8.3](http://openresty.org/en/changelog-1015008.html).
+  If you are installing Kong from one of our distribution
+  packages, this version and all required patches and modules are included.
+  If you are building from source, you must apply
+  Kong's [OpenResty patches](https://github.com/Kong/kong-build-tools/tree/master/openresty-build-tools/openresty-patches)
+  as well as include [lua-kong-nginx-module](https://github.com/Kong/lua-kong-nginx-module).
+  Our [kong-build-tools](https://github.com/Kong/kong-build-tools)
+  repository allows you to do both easily.
 - Bump OpenSSL version from `1.1.1f` to `1.1.1g`.
   [#5820](https://github.com/Kong/kong/pull/5810)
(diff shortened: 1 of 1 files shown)


jeremyjpj0916:
@hishamhm LGTM!

CHANGELOG.md
```diff
@@ -61,6 +61,18 @@
 
 ### Dependencies
 
+### Dependencies
```
gszr, line 64: Seems like we added another `### Dependencies` section

hishamhm: oops! will add the fix in my PR for UPGRADE.md #6156

PR #6166, opened 2020-07-29
feat(oauth2) authenticate consumer on token provision

### LINKED ISSUE #5958: [Feat][oauth2]Log consumer on oauth2 token generation client credentials grant
### Summary

Essentially this PR, but with unit tests(which I never personally figured out how to get working in current unit test toolkit as it didn't seem like the supporting elements  were present for me to validate consumer information in the context of that unit test token generation call):

https://github.com/Kong/kong/pull/4197

This is another drop in lua file customization we have been running in production for about 1.5+ years.

jjustus2:
### Summary

Old Related PR: #4197

Would really like this functionality to make it into mainstream Kong so I can stop maintaining a side patch file of ```access.lua``` I drop into our code every time to achieve this enhanced functionality for token generation around finding which customers abuse token generation. Cool with renaming functions or w/e is necessary too. Unsure if the unit test will work(didn't test it, but I know the code works currently as we run it in production).  

### Full changelog

* Implement authenticating the consumer during the token provision process which helps with logging analytics on who generates too many tokens.
* Validate the consumer context post token generation in unit tests using the ```kong.client``` pdk.

### Issues resolved

Fix #5958

### DIFF
--- a/spec/03-plugins/25-oauth2/03-access_spec.lua
+++ b/spec/03-plugins/25-oauth2/03-access_spec.lua
@@ -14,7 +14,8 @@ local ngx_encode_base64 = ngx.encode_base64
 
 
 local kong = {
-  table = require("kong.pdk.table").new()
+  table = require("kong.pdk.table").new(),
+  client = require("kong.pdk.client").new(),
 }
 
 
@@ -83,6 +84,8 @@ local function provision_token(host, extra_headers, client_id, client_secret, co
   assert.response(res).has.status(200)
   local token = assert.response(res).has.jsonbody()
   assert.is_table(token)
+  --Validate authenticated consumer context post token provision
+  assert.is_table(kong.client.get_consumer())
   request_client:close()
   return token
 end
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -86,6 +86,38 @@ do
   end
 end
 
+local function load_oauth2_credential_into_memory(credential_id)
+  local result, err = kong.db.oauth2_credentials:select { id = credential_id }
+  if err then
+    return nil, err
+  end
+
+  return result
+end
+
+
+local function load_credential_and_consumer_into_memory_from_credential(credential)
+  -- Retrieve the credential from the token credentials
+  local credential_cache_key =
+    kong.db.oauth2_credentials:cache_key(credential.id)
+
+  local cred, err = kong.cache:get(credential_cache_key, nil,
+                                   load_oauth2_credential_into_memory,
+                                   credential.id)
+  if err then
+    return nil, nil, err
+  end
+
+  -- Retrieve the consumer from the credential
+  local consumer_cache_key, consumer
+  consumer_cache_key = kong.db.consumers:cache_key(cred.consumer.id)
+  consumer, err      = kong.cache:get(consumer_cache_key, nil,
+                                      kong.client.load_consumer,
+                                      cred.consumer.id)
+
+  return cred, consumer, err
+end
+
 
 local function generate_token(conf, service, credential, authenticated_userid,
                               scope, state, disable_refresh, existing_token)
@@ -137,6 +169,14 @@ local function generate_token(conf, service, credential, authenticated_userid,
 
   if err then
     return error(err)
+  end
+
+  --Authenticate the consumer and credentials generating the token
+  local cred, consumer, err2 =
+  load_credential_and_consumer_into_memory_from_credential(credential)
+
+  if not err2 then
+    kong.client.authenticate(consumer, cred)
   end
 
   return {
@@ -881,16 +921,6 @@ local function parse_access_token(conf)
 end
 
 
-local function load_oauth2_credential_into_memory(credential_id)
-  local result, err = kong.db.oauth2_credentials:select { id = credential_id }
-  if err then
-    return nil, err
-  end
-
-  return result
-end
-
-
 local function set_consumer(consumer, credential, token)
   kong.client.authenticate(consumer, credential)
 
@@ -1011,23 +1041,10 @@ local function do_authentication(conf)
     end
   end
 
-  -- Retrieve the credential from the token
-  local credential_cache_key =
-    kong.db.oauth2_credentials:cache_key(token.credential.id)
+  -- Retrieve the credential from the token credential
+  local credential, consumer, err =
+  load_credential_and_consumer_into_memory_from_credential(token.credential)
 
-  local credential, err = kong.cache:get(credential_cache_key, nil,
-                                         load_oauth2_credential_into_memory,
-                                         token.credential.id)
-  if err then
-    return error(err)
-  end
-
-  -- Retrieve the consumer from the credential
-  local consumer_cache_key, consumer
-  consumer_cache_key = kong.db.consumers:cache_key(credential.consumer.id)
-  consumer, err      = kong.cache:get(consumer_cache_key, nil,
-                                      kong.client.load_consumer,
-                                      credential.consumer.id)
   if err then
     return error(err)
   end


jjustus2:
Arg... I really need to learn how to run the unit tests locally xD:

```
     __________
        FAILED  spec/03-plugins/25-oauth2/03-access_spec.lua:2603: Plugin: oauth2 [#cassandra] access Making a request works when a correct access_token is being sent in the querystring
spec/03-plugins/25-oauth2/03-access_spec.lua:88: Expected objects to be the same.
Passed in:
(nil)
Expected:
type table
stack traceback:
	spec/03-plugins/25-oauth2/03-access_spec.lua:88: in function 'provision_token'
	spec/03-plugins/25-oauth2/03-access_spec.lua:2604: in function <spec/03-plugins/25-oauth2/03-access_spec.lua:2603>
```

Will try to fix this up, I imagine an if statement around when this token_provision() call is made there is one odd edge case here.

jjustus2:
Running into the same problems as last time when I tried to deliver this feature. @Kong/team-core anyone know why trying to use the pdk ```kong.client``` seems to not function in the context of unit tests when runtime functional testing you would see my changes to the access.lua code do work? I even see ```kong.table``` used in this unit test without issue, and its not the call doesn't work, its that the consumer() is reportedly nil and not a table right now which can't be the case, if a token is generated successfully on an ```/oauth2/token``` call then the associated consumer to those creds would be authenticated and logged in the http logging and other log plugins appropriately for that network call.

jeremyjpj0916:
@thibaultcha you see what I mean in this PR? Am I off base thinking something in the pdk in the context of unit testing does not behave as it would during runtime? Using pdk to get the consumer context should return bob here not nil. The modifications to the oauth2 ```access.lua``` plugin works in the wild too for us which gives me further confidence that this is the case.

spec/03-plugins/25-oauth2/03-access_spec.lua
```diff
@@ -1373,6 +1374,7 @@ describe("Plugin: oauth2 [#" .. strategy .. "]", function()
           })
           local body = assert.res_status(200, res)
           assert.is_table(ngx.re.match(body, [[^\{"token_type":"bearer","access_token":"[\w]{32,32}","expires_in":5\}$]]))
+          assert.is_table(kong.client.get_consumer())
```
thibaultcha, line 1377: Integration tests do not run inside of Kong, but alongside. Using the PDK in integration tests will never work. This is why the PDK is tested via unit tests in `t/`.

jeremyjpj0916: Ah I see, looks like Kong has never been in a situation where a plugin needed a unit test in the ```/t``` directory w the pdk to confirm expected behavior, but it sounds like that is whats needed here. I supposed I would face the same problem trying to check confirmation of table ctx consumer via ```ngx.ctx.authenticated_consumer``` for the same reasons(run inside vs alongside) in the integration tests. 

So need something in ```/t``` that:

1. Stands up an oauth proxy and other needed Kong resources(consumer/consumer oauth creds/service/route/oauth2 plugin)

2. Calls the ```client_credentials``` grant token generation endpoint ```/oauth2/token``` with the consumer creds

3. Validates the pdk detects the consumers context set via ```kong.client.get_consumer```

Feels overkill for the minor improvement to oauth2 token gen existing behavior, but should be able to do it to maintain good visibility and keep the feature in a detectable state if somehow someone ever broke it. I just see no examples in the ```/t``` where Kong specific resources are being created and its not all just generic nginx checks so flying blind a bit. If there is any simpler approach do lemme know, was thinking to clean this up this weekend one evening among a few other pet projects I am tackling.

thibaultcha: For this we typically develop a tests plugin: https://github.com/Kong/kong/tree/master/spec/fixtures/custom_plugins/kong/plugins

jjustus2: Cool, thanks for the link. There would not be much value if I have to rewrite the same logic in the oauth2 plugin in isolate(cause we want to hit the actual plugin, not dup code of it) but that might not be whats happening here. Did just try the ngx.ctx.auth_consumer field but fully expect this commit to fail before I try to see how the tests plugins relate to where they are implemented and validated and how it helps my use case here.

thibaultcha: To do this, we typically configure an OAuth plugin and one of those tests plugins onto a Route in the integration tests. The test plugin typically gets executed after the OAuth plugin and and plays the role of asserting that any `ngx.ctx` or other contextual values were properly set.

jjustus2: To be clear when you refer to integration tests you mean https://github.com/Kong/kong/tree/master/spec/02-integration right?

thibaultcha: Those and plugins integration tests yes.

codifierr:
Is there a way to throttle this endpoint "/oauth2/token" via a rate-limiting plugin may be.

jeremyjpj0916:
> @thibaultcha @jeremyjpj0916 Is there a way to rate-limit this endpoint "/oauth2/token" via a rate-limiting plugin maybe?

@codifierr 

Complexity there is when you have LB's in-front of Kong and no sure way to get originating client IP. On the flip side if you wanted to rate limit by consumer other issue is not knowing who the consumer is at token gen time(this PR idea is to solve it for logging purposes at least), and if that logic was added I would want a way to rate limit them once the consumer is known, but prior to injecting a fresh value in the db related to their client_id and client_secret. Would certainly be a good thing.

Another way we have internally solved token gen abuse delimas is we made this "cached" token gen endpoint, that if a client calls the token gen endpoint prior to a valid token they have expiry, it will return them the same token and its ever decreasing ttl. Might be something we could also PR sometime but dunno if Kong would be cool with it since its probably not "RFC" behavior, but I care more about a stable offering that is protective in nature than appealing to an RFC if I have to haha. 

Hopefully I can get back to this PR and fix up the test case like Tibo mentioned. Have not had time to dig into it just yet, other big items on my plate internally at the moment.

codifierr:
> > @thibaultcha @jeremyjpj0916 Is there a way to rate-limit this endpoint "/oauth2/token" via a rate-limiting plugin maybe?
> 
> @codifierr
> 
> Complexity there is when you have LB's in-front of Kong and no sure way to get originating client IP(if rate limiting by IP). On the flip side if you wanted to rate limit by consumer other issue is not knowing who the consumer is at token gen time(this PR idea is to solve it for logging purposes at least), and if that logic was added I would want a way to rate limit them once the consumer is known(well client_id and client_secret I think is best way to rate limit by, don't wanna see to many of the same cred pair used too frequently), and prior to injecting a fresh token value in the db so DB r/w could be prevented. Would certainly be a good thing.
> 
> Another way we have internally solved token gen abuse delimas is we made this "cached" token gen endpoint, that if a client calls the token gen endpoint prior to a valid token they have expiry, it will return them the same token and its ever decreasing ttl(preventing the need for DB r/w). Might be something we could also PR sometime but dunno if Kong would be cool with it since its probably not "RFC" behavior, but I care more about a stable offering that is protective in nature than appealing to an RFC if I have to haha.
> 
> Hopefully I can get back to this PR and fix up the test case like Tibo mentioned. Have not had time to dig into it just yet, other big items on my plate internally at the moment.

@jeremyjpj0916 I am actually looking for rate-limiting just by client-id and client-secret which is a term known as rate-limit by credentials. I tried to do that in a kong patch. But looks like the rate-limit plugin only executes if the consumer is authenticated.

hbagdi:

PR #6170, opened 2020-07-30
fix(handler) convert collectgarbage output to bytes

gszr:
Fix issue where memory in use as reported by the `/status`
endpoint would be treated as `bytes`, while the output of
`collectgarbage("count")` is given in kbytes.

CHANGELOG:
- Fix issue where consumed worker memory as reported by `/status` would be incorrectly reported

### DIFF
--- a/t/01-pdk/12-node/02-get_memory_stats.t
+++ b/t/01-pdk/12-node/02-get_memory_stats.t
@@ -109,8 +109,8 @@ lua_shared_dicts
   \S+: \d+\/2[45]\d{3}
   \S+: \d+\/3[23]\d{3}
 workers_lua_vms
-  1: 1234
-  2: 1234
+  1: 1263616
+  2: 1263616
   (?:\d+: \d+\s*){1,2}
 --- no_error_log
 [error]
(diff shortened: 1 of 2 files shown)


hbagdi:
github.com/kong/kong-plugin-prometheus also uses this API.
GS, could you or someones else take a look if we need any changes in the plugin?
@kikito has expressed interest in the past in looking into that plugin.

gszr:
> github.com/kong/kong-plugin-prometheus also uses this API.
> GS, could you or someones else take a look if we need any changes in the plugin?
> @kikito has expressed interest in the past in looking into that plugin.

Thanks for the nudge, @hbagdi. I just took a look; no changes are needed over there - Prometheus exporter module uses the PDK, which already treats the value we are changing as a byte count.

Verifying further, values reported by the plugin thus far were indeed inaccurate - kbytes reported as bytes:
```
# TYPE kong_memory_workers_lua_vms_bytes gauge
kong_memory_workers_lua_vms_bytes{pid="23320"} 27640
kong_memory_workers_lua_vms_bytes{pid="23321"} 27588
```

`ps`:
```
# rss value here is in 1024 byte units
ps -eo rss,comm | grep nginx
...
  32264 nginx: worker process
  32088 nginx: worker process
```

`/metrics` output after the fix:
```
# TYPE kong_memory_workers_lua_vms_bytes gauge
kong_memory_workers_lua_vms_bytes{pid="24995"} 27636539
kong_memory_workers_lua_vms_bytes{pid="24996"} 27586017
```

kong/runloop/handler.lua
```diff
@@ -91,7 +91,7 @@ do
     local time = ngx.time()
 
     if force or time - last >= LUA_MEM_SAMPLE_RATE then
-      local count = collectgarbage("count")
+      local count = collectgarbage("count") * 1000
```
locao, line 94:
```suggestion
      local count = collectgarbage("count") * 1024
```
Just to be more precise, the original value is split by 1024: `setnumV(L->top, (lua_Number)G(L)->gc.total/1024.0);`

gszr: Great point!

t/01-pdk/12-node/02-get_memory_stats.t
```diff
@@ -436,3 +436,47 @@ GET /t
 workers_lua_vms\Z
 --- no_error_log
 [error]
+
+
```
thibaultcha, line 440: missing blank line (you can run the `reindex` script for this) (https://github.com/openresty/openresty-devel-utils)

gszr: Thanks @thibaultcha - did not know about that tool!

t/01-pdk/12-node/02-get_memory_stats.t
```diff
@@ -436,3 +436,48 @@ GET /t
...
+
+=== TEST 10: update_lua_mem converts count to bytes
+--- http_config eval
+qq{
+    $t::Util::HttpConfig
+
+    lua_shared_dict kong 24k;
+    lua_shared_dict kong_db_cache 32k;
```
thibaultcha, line 448: Do we need this shm here?

gszr: Good catch, we do not.

kong/runloop/handler.lua
```diff
@@ -91,7 +91,7 @@ do
     local time = ngx.time()
 
     if force or time - last >= LUA_MEM_SAMPLE_RATE then
-      local count = collectgarbage("count")
+      local count = collectgarbage("count") * 1024
```
thibaultcha, line 94: Actually, could we fix this on the PDK side instead of within the handler? This way, we ensure that GC usage values can be reported in kb from anywhere (without additional processing), and the PDK is made aware that all values reported as such are in kb, and would be the only place where we make the kb -> b conversion.

gszr: Sure, I will rework the fix and test.

gszr: @thibaultcha could you re-review? Would love this to make it into 2.1.1. Thanks!

gszr:
Thanks for the reviews, folks!

Issue #6159, opened 2020-07-27
2.1 Route name null incorrectly parsed in admin api response


gilmansean:
### Summary

GET on Kong 2.1 admin api `/routes` endpoint returns incorrectly parses null name field, `"name": "$(VALUE)",`.

```json
    {
      "id": "f70c1923-2d48-47e4-aa9f-0fe66889c633",
      "path_handling": "v1",
      "updated_at": 1586331527,
      "destinations": null,
      "headers": null,
      "protocols": [
        "http",
        "https"
      ],
      "created_at": 1586331527,
      "snis": null,
      "service": {
        "id": "fc9e13e0-0959-4f4b-9643-6f0e6df367bd"
      },
      "name": "$(VALUE)",
      "methods": null,
      "preserve_host": false,
      "regex_priority": 0,
      "paths": [
        "/api/myapi/v1"
      ],
      "sources": null,
      "strip_path": true,
      "https_redirect_status_code": 426,
      "hosts": null,
      "tags": null
    }
```

### Steps To Reproduce

1. upgrade to Kong 2.1 when having null name in a route resource
2. GET `/routes` on admin API

### Additional Details & Logs

- Kong version (`2.1`)
- Operating system: alpine 3.10

rsbrisci:
Can confirm, this is an issue in `2.1.0` - it was not in `2.0.5`

jeremyjpj0916:
Additionally, the consumer resource also experiences such an issue with a GET on the ```/consumers``` for pre 2.1 consumers that have had a null field now reporting the errant ```$(VALUE)``` response, this time on ```custom_id``` (an unused field for us too):
```
{
  "custom_id": "$(VALUE)",
  "created_at": 1595364201,
  "id": "c645e2f6-966f-48c0-91cf-6af27b968bbe",
  "tags": null,
  "username": "democonsumer"
}
```

hishamhm:
To confirm, these are being observed on Cassandra, right?

jjustus2:
@hishamhm Yes. Opps on my spare github account 😀 .

PR #6171, opened 2020-07-30
fix(migrations) don't migrate empty values when using Cassandra

locao:
Empty values were not being removed in the original template, resulting in `$(VALUE)` in the migrated database.

Fix: #6159


kong/db/migrations/operations/200_to_210.lua
```diff
@@ -345,7 +345,9 @@ local cassandra = {
         for _, row in ipairs(rows) do
           local set_list = { "ws_id = " .. default_ws }
           for _, key in ipairs(unique_keys) do
-            table.insert(set_list, render([[$(KEY) = '$(WS):$(VALUE)']], {
+            local template = row[key] and [[$(KEY) = '$(WS):$(VALUE)']] or
+                                          [[$(KEY) = '$(WS):']]
+            table.insert(set_list, render(template, {
```
hishamhm, line 350: @locao I wonder if we should be setting these to "workspace-colon-empty-string" and if this won't cause trouble down the line with actual empty-string values. What if instead we just _not_ add them to `set_list`, so that null values are not rendered as part of the `UPDATE`? In other words, `if row[key] then table.insert(...` — does that work?

locao: Fixed!

hishamhm: on second thought... we actually need to make this into a new migration that fixes this, otherwise people who already migrated into 2.1.0 won't get this fixed. :\

hishamhm:
on second thought... we actually need to make this into a new migration that fixes this, otherwise people who already migrated into 2.1.0 won't get this fixed. :\

kong/db/migrations/core/010_210_to_211.lua
```diff
@@ -0,0 +1,80 @@
+
+local re_match = ngx.re.match
+
+
+local core_entities_to_clean = {
```
kikito, line 5: I don't know whether this is planned for an extra migration, but I want to point out that this is not including custom plugin entities on the cleanup. 

Of those, I *think* the only ones that might need cleanup (because they have custom uniques) are `oauth2_credentials` / `authorization_codes` / `tokens`. Please double check that I didn't miss any.

If the cleanup is needed there, then the helper functions used on this module need to be put in a place where they can be shared with the plugin migrations.

kong-2.1.1-0.rockspec
```diff
@@ -1,9 +1,9 @@
 package = "kong"
-version = "2.1.0-0"
+version = "2.1.1-0"
```
hishamhm, line 2: Minor note: in the past we have only bumped the rockspec version at release time (even when new migrations for the next version were added)

kong/db/migrations/core/010_210_to_211.lua
hishamhm: > If the cleanup is needed there, then the helper functions used on this module need to be put in a place where they can be shared with the plugin migrations.

`kong/db/migrations/operations/210_to_211.lua` would be a good choice

PR #6183, opened 2020-08-04
fix(schema) add RFC 3986 "sub-delims" characters as valid inside path…

### LINKED ISSUE #6125: Kong Service Resource URI Validation Error?
### Summary

Seems Kong throws a schema violation error on URI's with a $ in them? Is this not a valid URL?:

http://my-api.company.com/R4/Bundle/$consultation-submit

Service Resource:
```
{
  "host": "my-api.company.com",
  "created_at": 1593734424,
  "connect_timeout": 2000,
  "id": "7be474e8-02c3-4787-9450-09a43540649c",
  "protocol": "http",
  "name": "optum.ocm.advisor.stage.optum.ocm.telemedService.dmz.v1",
  "read_timeout": 9000,
  "port": 80,
  "path": "/R4/Bundle",   ---> can't add the /$consultation-submit here
  "updated_at": 1593734424,
  "retries": 0,
  "write_timeout": 9000,
  "tags": null,
  "client_certificate": null,
  "extras": {}
}
```

Kong throws a schema violation error:

```
'/R4/Bundle/$consultation-submit' (characters outside of the reserved list of RFC 3986 found))
```

Related:
https://stackoverflow.com/questions/1856785/characters-allowed-in-a-url 

### Steps To Reproduce

1. Try to create a Kong Service resource with $ in the URI

### Additional Details & Logs

- Kong version 2.0.5

dndx:
… segments

fixes #6125

New concerns:

Currently Kong doesn't explicitly distinguish between regex and plain paths. That is, we guess whether a path is regex or plain path by using this regex: https://github.com/Kong/kong/blob/master/kong/router.lua#L425. If the path didn't match it is considered as regex path and fed into the PCRE engine.

With this change, a lot of these urls will be considered as "plain" path and no longer go through PCRE. So this will be a breaking change (test failures inside `spec/01-unit/08-router_spec.lua` as well), which makes me hesitate if there is a better way to do it.

Use Nginx as an example, the matching between regex and plain paths are explicit by the `~` operator like this:
* `location /foo` (Plain)
* `location ~ /foo` (Regex)

I wonder if we should start doing the same. If not, adding just the characters that has no special meanings in regexs is a safer route at this moment.

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -76,7 +76,10 @@ end
 
 
 local function validate_path(path)
-  if not match(path, "^/[%w%.%-%_~%/%%]*$") then
+  if not match(path, "^/[%w%.%-%_%~%/%%%:%@" ..
+                     "%!%$%&%'%(%)%*%+%,%;%=" .. -- RFC 3986 "sub-delims"
+                     "]*$")
+  then
     return nil,
            "invalid path: '" .. path ..
            "' (characters outside of the reserved list of RFC 3986 found)",
@@ -89,7 +92,7 @@ local function validate_path(path)
 
     if raw:find("%", nil, true) then
       local err = raw:sub(raw:find("%%.?.?"))
-      return nil, "invalid url-encoded value: '" .. err .. "'"
+      return nil, "invalid url-encoded value: '" .. err .. "'", "percent"
     end
   end
 
@@ -394,14 +397,21 @@ end
 
 local function validate_path_with_regexes(path)
 
-
   local ok, err, err_code = typedefs.path.custom_validator(path)
 
-  if ok or err_code ~= "rfc3986" then
+  if err_code == "percent" then
     return ok, err, err_code
   end
 
-  -- URI contains characters outside of the reserved list of RFC 3986:
+  -- We can't take an ok from validate_path as a success just yet,
+  -- because the router is currently more strict than RFC 3986 for
+  -- non-regex paths:
+  if ngx.re.find(path, [[^[a-zA-Z0-9\.\-_~/%]*$]]) then
+    return true
+  end
+
+  -- URI contains characters outside of the list recognized by the
+  -- router as valid non-regex paths.
   -- the value will be interpreted as a regex by the router; but is it a
   -- valid one? Let's dry-run it with the same options as our router.
   local _, _, err = ngx.re.find("", path, "aj")
--- a/spec/01-unit/01-db/01-schema/05-services_spec.lua
+++ b/spec/01-unit/01-db/01-schema/05-services_spec.lua
@@ -190,9 +190,8 @@ describe("services", function()
 
     it("rejects regular expressions & other non-rfc 3986 chars", function()
       local invalid_paths = {
-        [[/users/(foo/profile]],
-        [[/users/(foo/profile)]],
-        [[/users/*/foo]],
+        [[/users/|foo/profile]],
+        [[/users/(this|foo/profile)]],
       }
 
       for i = 1, #invalid_paths do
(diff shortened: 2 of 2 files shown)


hishamhm:
> I wonder if we should start doing the same.

We should, but that would be effectively a new feature. Giving this some thought to see if we can make this fit in a non-breaking manner.

hishamhm:
@dndx I did some experimentation with this. What we can do, in a non-breaking fashion for 2.1.1, is the following: let's only change the definition of `validate_path` function in `typedefs.lua` so that `service.path` accepts the sub-delims characters, but not change the regex detection in the router. The behavior of `route.paths` does not change, and everything remains backwards compatible.

@jeremyjpj0916 With this solution, you will be able to use a plain `$` in the `path` of the Service, but if you need to match an incoming path including `$` in a Route, you'll need to use regex notation for that.

In other words, by changing only `validate_path` and not modifying the router, this should work:

```nginx
# nginx.conf for an example upstream service
events {}
http {
   server {
      server_name rush;
      listen 0.0.0.0:2112;
      
      location = /hello$world {
         return 418;
      }
   }   
}
```

```yaml
# kong.conf for an example declarative config 
_format_version: "1.1"
services:
 - name: rush
   url: "http://127.0.0.1:2112/hello$world" # plain dollar in service.path
   routes:
   - name: my-route
     paths:
     - "/hi\\$world" # dollar (as regex with escape) in route.paths
```

```bash
$ KONG_DECLARATIVE_CONFIG=kong.yml KONG_DATABASE=off bin/kong start -c kong.conf
Kong started

$ curl -i 'http://127.0.0.1:8000/hi$world'
HTTP/1.1 418 
Content-Length: 0
Connection: keep-alive
Server: openresty/1.15.8.3
Date: Tue, 04 Aug 2020 21:07:16 GMT
X-Kong-Upstream-Latency: 0
X-Kong-Proxy-Latency: 2
Via: kong/2.1.0
```

jeremyjpj0916:
@hishamhm That would work for my example case defined in the issue because it was a Service resource path that wants $ in the URI of that backend endpoint. Will keep in mind the requirements on the route URI for future reference though(most our route URIs have a requirement of /api/{env}/{namespace}/{resource}/{version} though and leverage wildcard pathing, so won't be an issue in our architecture I don't think, never seen resources yet being named with $xyz )! Thanks for putting this together.

spec/01-unit/01-db/01-schema/06-routes_spec.lua
```diff
@@ -262,9 +262,9 @@ describe("routes schema", function()
 
         local ok, err = Routes:validate(route)
         assert.falsy(ok)
-        assert.equal(u([[invalid regex: '/users/(foo/profile' (PCRE returned:
+        assert.equal(u([[invalid regex: '/users/(foo|profile' (PCRE returned:
```
hishamhm, line 265: the set of valid RFC 3986 characters doesn't really affect the regex validation on routes.

I recommend moving the test below to test `services.path` attribute instead

kong/router.lua
```diff
@@ -422,7 +422,9 @@ local function marshall_route(r)
       route_t.match_weight = route_t.match_weight + 1
 
       for _, path in ipairs(paths) do
-        if re_find(path, [[^[a-zA-Z0-9\.\-_~/%]*$]]) then
+        -- see character set used in
+        -- kong/db/schema/typedefs.lua -> validate_path function
+        if re_find(path, [[^[a-zA-Z0-9\.\-_~/%:@!$&'()*+,;=]*$]]) then
```
hishamhm, line 427: we should revert this and keep knog/router.lua unchanged

kong/db/schema/typedefs.lua
```diff
@@ -76,7 +76,11 @@ end
 
 
 local function validate_path(path)
-  if not match(path, "^/[%w%.%-%_~%/%%]*$") then
+  -- see character set used in kong/router.lua -> marshall_route function
+  if not match(path, "^/[%w%.%-%_%~%/%%%:%@" ..
+                     "%!%$%&%'%(%)%*%+%,%;%=" .. -- RFC 3986 "sub-delims"
+                     "]*$")
```
hishamhm, line 82: :+1:

thibaultcha:
> I wonder if we should start doing the same. If not, adding just the characters that has no special meanings in regexs is a safer route at this moment.

Agree with this sentiment. This would allow the router to support a modifier similar to Nginx's `^~`, which has been a long time missing and preventing some use-cases from being possible...

Youngseo-Jeon0313:
@fairyqb @hishamhm

Hi, I'm going to using Routing service with KONG, but I got schema violation code 2. 

In my interpretation, the above code is handled by URLs that correspond to the RFC reservation word.
The path I want to route is {IP}: {port}/#/stresult, but I can't interpret '#'. 

Can you add character '#', too ??

PR #6208, opened 2020-08-11
fix(cors) send HTTP 204 on Preflight request

### LINKED ISSUE #6051: CORS OPTIONS(Preflight request) request should return 204
Next line returns 200 as status https://github.com/Kong/kong/blob/c817fada7a514d84f6e0aab7461605c71bf3a841/kong/plugins/cors/handler.lua#L13

Internet says it should return 204:
- https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
- https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-3.1#preflight-requests

Internet changed its mind since 2018 https://github.com/Kong/kong/pull/4029

thehanimo:
### Summary

The appropriate status code for a preflight request is 204 as per the updated Mozilla guidelines (https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request). We previously used to send HTTP 200.

### Full changelog

* return 204 for preflight response instead of 200
* update tests to accommodate status code change

### Issues resolved

Fix #6051

### DIFF
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -10,7 +10,7 @@ local tostring = tostring
 local ipairs   = ipairs
 
 
-local HTTP_OK = 200
+local HTTP_OK = 204
 
 
 local CorsHandler = {}
--- a/kong/plugins/cors/handler.lua
+++ b/kong/plugins/cors/handler.lua
@@ -10,7 +10,7 @@ local tostring = tostring
 local ipairs   = ipairs
 
 
-local HTTP_OK = 204
+local HTTP_NO_CONTENT = 204
 
 
 local CorsHandler = {}
@@ -232,7 +232,7 @@ function CorsHandler:access(conf)
     set_header("Access-Control-Max-Age", tostring(conf.max_age))
   end
 
-  return kong.response.exit(HTTP_OK)
+  return kong.response.exit(HTTP_NO_CONTENT)
 end
 
 
--- a/spec/03-plugins/13-cors/01-access_spec.lua
+++ b/spec/03-plugins/13-cors/01-access_spec.lua
@@ -536,7 +536,7 @@ for _, strategy in helpers.each_strategy() do
               }
             })
 
-            assert.res_status(200, res)
+            assert.res_status(204, res)
 
             if accept then
               assert.equal(CORS_DEFAULT_METHODS, res.headers["Access-Control-Allow-Methods"])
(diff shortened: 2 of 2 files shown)


locao:
Hi @thehanimo, thanks for your submission! 

Can you fix the test that is failing, please?

```
spec/03-plugins/13-cors/01-access_spec.lua:681:31: expected ')' near 'res'
```

thehanimo:
@locao made a few fixes. Travis CI fails with the following:
```
spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua:242: in function <spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua:208>
```
Could this be a one-off issue where the latencies don't add up?

thehanimo:
Turns out it was :)

Do take a look @locao!

hishamhm:
Change looks correct according to the MDN recommendations. But if I'm reading the spec correctly it seems to allow both 200 and 204 as valid responses: https://fetch.spec.whatwg.org/ — and it looks like in the past the choice of value has caused compatibility issues, however:

```js
var corsOptions = {
  origin: 'http://example.com',
  optionsSuccessStatus: 200 // some legacy browsers (IE11, various SmartTVs) choke on 204
}
```
[(source)](https://expressjs.com/en/resources/middleware/cors.html)

Before we change this value around, is there a current _need_ to make it return 204 (i.e. are there clients currently failing given the 200 response), or is this goal of this change exclusively to align with the MDN recommendation?

I'm trying to assess whether changing this in a patch release would cause surprising behavior or breakage for users relying on the current behavior.

thehanimo:
@hishamhm I guess this change was solely aimed at aligning with the MDN recommendations. Not sure if there's a need. @dzmitry-lahoda, was there something specific you had in mind?

dzmitry-lahoda:
MDN - yes. And MS did the change in ASP.NET either.

kikito (changes requested):
Because of the reason mentioned by @hisham, this can potentially break backwards compatibility for users already using the plugin, and consuming it via a plugin.

We can conform with MDN recommendations and still be backwards compatible by adding a config option. Something like:

```
--  kong/plugins/cors/schema.lua

return {
  name = "cors",
  fields = {
    ...
    { config = {
        type = "record",
        fields = {
        ...
           { preflight_status = {
               type = "integer",
               required = true,
               default = 200,
               one_of = { 200, 204 }, },
               ...
}
```

And use `conf.preflight_status` as the status. This way people can "opt in" to the MDN recommendations without breaking backwards compatibility for others.

PR #6210, opened 2020-08-11
fix(migrations) workspaces related migration issues

### LINKED ISSUE #6187: Multiple default workspaces on reentrant migration (cassandra)
# Summary

Running `kong migrations up -f` multiple times on cassandra will create as many `default` workspaces.

### Steps To Reproduce

(on cassandra)

1. `kong migrations bootstrap`
2. `kong migrations up -f` x 3

```
$ cqlsh -k kong -e "select id, name from workspaces"

 id                                   | name
--------------------------------------+---------
 176b6527-ff5d-4ea3-869c-ec079e3dbc03 | default
 e83fa585-ab01-411f-8a43-07a02a39279f | default
 29bdb52e-c7c8-4929-a050-70269a79c954 | default
 9f97fa94-589f-4ef2-affd-28260d615aae | default

(4 rows)
```

### Additional Details & Logs

- Kong 2.1.0

bungle:
### Summary

This PR has 6 commits to fix workspace related migrations introduced in `2.1.0`:

#### fix(migrations) remove postgres index that is already created by unique constraint

Postgres `UNIQUE` constraint automatically creates UNIQUE index, thus additional index on for those is not needed.

#### fix(migrations) ensure that cassandra does not create multiple default workspaces

When running migrations with `force`, e.g.:

```
kong migrations up -f
```

OR

```
kong migrations finish -f
```

Cassandra created multiple `workspaces` with `name=default`. This commit fixes that.

#### fix(migrations) split cert_digest update to two loops

In general, it is not good to modify a table/data that we are iterating at the same time. This commit changes it so that first iterator collects the statements, and second iteration then runs the updates.

This commit also makes it skip unnecessary updates, making it better on reentrancy scenarios.

#### fix(migrations) add some extra guards on postgres schema changes 

Just adds some extra guards for 2.0.0 to 2.1.0 Postgres schema changes to make it bit more resilient and reentrant.

#### fix(migrations) split cassandra update plugin config to two loops

Just splits one of the Cassandra teardown migrations from to separate iterate and update loops.

#### fix(migrations) split rest of the migrations to iteration and update and handle error consistently

Updates rest of the migrations to split `iteration`  and `update` to different loops, as modifying external data while looping around it might be problematic in some cases.

Also changes error handling to be consistent across migrations, especially those newly added operations.

### Issues Resolved

Fix #6187


kong/db/migrations/operations/200_to_210.lua
```diff
@@ -13,6 +16,54 @@ local function render(template, keys)
...
 
+local function cassandra_get_default_ws(connector)
+  local rows, err = connector:query("SELECT * FROM workspaces WHERE name='default'")
+  if err then
+    return nil, err
+  end
+
+  if not rows
+  or not rows[1]
+  or not rows[1].id
```
dndx, line 27:
```suggestion
     or not rows[1]
     or not rows[1].id
```

for aesthetics.

kong/db/migrations/operations/200_to_210.lua
```diff
@@ -13,6 +16,54 @@ local function render(template, keys)
...
+  end
+
+  return rows[1].id
+end
+
+
+local function cassandra_create_default_ws(connector)
+  local cql = render([[
+INSERT INTO workspaces(id, name, created_at)
+     VALUES (uuid(), 'default', $(NOW))
```
dndx, line 39:
```suggestion
    INSERT INTO workspaces(id, name, created_at)
    VALUES (uuid(), 'default', $(NOW))
```

PR #6215, opened 2020-08-12
feat(stream) UDP proxy support

dndx:
This PR adds basic UDP proxying support to Kong.

### DIFF
--- a/spec/02-integration/05-proxy/26-udp_spec.lua
+++ b/spec/02-integration/05-proxy/26-udp_spec.lua
@@ -1,0 +1,49 @@
+local helpers = require "spec.helpers"
+
+
+local UDP_PROXY_PORT = 26001
+
+
+for _, strategy in helpers.each_strategy() do
+
+  describe("UDP Proxying [#" .. strategy .. "]", function()
+    lazy_setup(function()
+      local bp = helpers.get_db_utils(strategy, {
+        "routes",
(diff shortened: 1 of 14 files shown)


hbagdi:
Question for personal edification: What are the semantics of `retires` in Service entity when the protocol is UDP?

dndx:
@hbagdi Good question. There is no concept of "retries" for UDP sessions. I guess it is because we can not know for sure if any given packets are successfully delivered or not. This is the behavior of `ngx_stream_proxy_module` in datagram mode.

I think we should make reject services with the retry attribute if `protocol = udp`. Will add that.

hbagdi:
Perfect. Thanks! I wanted to make sure there is no other meaning of retries in nginx internals.

spec/02-integration/05-proxy/26-udp_spec.lua
```diff
@@ -0,0 +1,49 @@
...
+        stream_listen = "127.0.0.1:" .. UDP_PROXY_PORT .. " udp",
+      })
+
+    end)
+
+    lazy_teardown(function()
+      helpers.stop_kong()
+    end)
+
+    it("proxies grpc", function()
```
gszr, line 39:
```suggestion
    it("proxies udp", function()
```

gszr (approved):

PR #6216, opened 2020-08-12
feat/plugins-rate-limiting by path

taotao2014:
**Summary**

plugins-rate-limiting limit by path


taotao2014:
Some checks were not successful ???
cp: cannot stat '/tmp/build/*': No such file or directory

PR #6228, opened 2020-08-14
fix (plugins-key-auth) fault tolerant for body

### LINKED ISSUE #5893: Key Auth Plugin throws 400 Bad Request when requests doesn't have content-type using key_in_body
### Summary

### Steps To Reproduce

### Additional Details & Logs

fairyqb:
### Summary

fix (plugins-key-auth) fault tolerant for body

### Issues resolved

Fix #5893


Tieske:
@fairyqb please do not repeat those PR's. Instead update an existing on please.

PR #6229, opened 2020-08-14
fix (plugins-key-auth) fault tolerant for body

### LINKED ISSUE #5893: Key Auth Plugin throws 400 Bad Request when requests doesn't have content-type using key_in_body

fairyqb:
### Summary

fix (plugins-key-auth) fault tolerant for body

### Issues resolved

Fix #5893

### DIFF
--- a/kong/plugins/key-auth/schema.lua
+++ b/kong/plugins/key-auth/schema.lua
@@ -18,6 +18,7 @@ return {
           { hide_credentials = { type = "boolean", default = false }, },
           { anonymous = { type = "string" }, },
           { key_in_body = { type = "boolean", default = false }, },
+          { body_fault_tolerant = { type = "boolean", default = true }, },
           { run_on_preflight = { type = "boolean", default = true }, },
         },
     }, },
--- a/kong/plugins/key-auth/schema.lua
+++ b/kong/plugins/key-auth/schema.lua
@@ -18,7 +18,6 @@ return {
           { hide_credentials = { type = "boolean", default = false }, },
           { anonymous = { type = "string" }, },
           { key_in_body = { type = "boolean", default = false }, },
-          { body_fault_tolerant = { type = "boolean", default = true }, },
           { run_on_preflight = { type = "boolean", default = true }, },
         },
     }, },
--- a/spec/03-plugins/09-key-auth/02-access_spec.lua
+++ b/spec/03-plugins/09-key-auth/02-access_spec.lua
@@ -568,6 +568,18 @@ for _, strategy in helpers.each_strategy() do
           assert.equal("bar", json.post_data.params.foo)
         end)
       end
+
+      it("works with 'key_in_body' and unsupported content type", function()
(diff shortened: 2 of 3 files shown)


kong/plugins/key-auth/schema.lua
```diff
@@ -18,6 +18,7 @@ return {
           { hide_credentials = { type = "boolean", default = false }, },
           { anonymous = { type = "string" }, },
           { key_in_body = { type = "boolean", default = false }, },
+          { body_fault_tolerant = { type = "boolean", default = true }, },
```
Tieske, line 21: I do not think this is needed as an option

fairyqb:
Some checks haven’t completed yet, why ??

FAILED spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua:208: Plugin: tcp-log (log) [#cassandra] logs proper latencies (#grpc)

@Tieske

Tieske:
the testcase should follow the behavour of #5893 ; a valid key in query or header, and a body that cannot be parsed

spec/03-plugins/09-key-auth/02-access_spec.lua
```diff
@@ -569,19 +569,16 @@ for _, strategy in helpers.each_strategy() do
...
             ["Host"] = "key-auth6.com",
             ["Content-Type"] = "text/plain",
-          },
-          body = "foobar",
+          }
         })
-
-        local body = assert.res_status(400, res)
-        local json = cjson.decode(body)
-        assert.same({ message = "Cannot process request body" }, json)
```
hishamhm, line 584: What happens in this test case that was removed? That should still produce an HTTP 400, right? (Even if with a different error message)

Could you keep this test case in the test suite and then add the new one, instead of replacing one with the other? Thank you!

spec/03-plugins/09-key-auth/02-access_spec.lua
```diff
@@ -569,19 +569,16 @@ for _, strategy in helpers.each_strategy() do
         end)
       end
 
-      it("fails with 'key_in_body' and unsupported content type", function()
+      it("works with 'key_in_body' and unsupported content type", function()
```
hishamhm, line 572:
```suggestion
      it("works with 'key_in_body' and unsupported content type if key was given in query string", function()
```

spec/03-plugins/09-key-auth/02-access_spec.lua
fairyqb: OK.

fairyqb:
@hishamhm  @Tieske  @bungle . done

fairyqb:
#6274 base next branch
@hishamhm  @Tieske @bungle  thank you.

bungle:
The #6357 got merged which was based on @fairyqb's work, and I just added some refinements and some additional tests. Thank you! It will be on `next` release.

PR #6251, opened 2020-08-19
feat(*) x-forwarded-path and request.get_forwarded_prefix

gszr:
### Summary

Picks breaking portions of https://github.com/Kong/kong/pull/6201 into `next`.

**feat(pdk) add request.get_forwarded_prefix**

* **Breaking change**: existing `request.get_forwarded_path` method was
modified to return the request's `X-Forwarded-Path` header (if any,
and coming from a trusted client) or the request's path, as obtained by
`request.get_path`, rather than the request's prefix path stripped by
Kong
* New method `request.get_forwarded_prefix` was introduced; it returns
the request's prefix path component, as stripped by Kong. It also checks
if a `X-Forwarded-Prefix` header comes from a trusted source and uses
the given value

**feat(handler) add x-forwarded-path header**

* If a trusted source provides a `X-Forwarded-Path` header, it's
proxied as-is
* Otherwise, Kong will set the content of said header to the request's
URI

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -16,7 +16,6 @@ local find = string.find
 local lower = string.lower
 local type = type
 local error = error
-local pcall = pcall
 local tonumber = tonumber
 local check_phase = phase_checker.check
 local check_not_phase = phase_checker.check_not
@@ -292,7 +291,7 @@ local function new(self)
   -- from a trusted source, and uses it as is when given. The value is returned
   -- as a Lua string.
   --
-  -- In general, this function should be called after Kong has ran its router,
+  -- If a trusted `X-Forwarded-Prefix` is not passed, this function must be called after Kong has ran its router (`access` phase),
   -- as the Kong router may strip the prefix of the request path. That stripped
   -- path will become the return value of this function, unless there was already
   -- a trusted `X-Forwarded-Prefix` header in the request.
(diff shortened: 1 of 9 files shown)


kong/pdk/request.lua
```diff
@@ -273,15 +274,63 @@ local function new(self)
...
     if self.ip.is_trusted(self.client.get_ip()) then
       local prefix = _REQUEST.get_header(X_FORWARDED_PREFIX)
       if prefix then
         return prefix
       end
     end
 
-    local path = _REQUEST.get_path()
-    return path == "" and "/" or path
+    local ok, prefix = pcall(function()
```
dndx, line 325: This `pcall` should not be necessary. In case the variable does not exist `nil` will be returned instead of a Lua error.

kong/pdk/request.lua
```diff
@@ -273,15 +273,60 @@ local function new(self)
...
     end
 
-    local path = _REQUEST.get_path()
-    return path == "" and "/" or path
+    local prefix = ngx.var.upstream_x_forwarded_prefix
+    if not prefix then
+      return "/"
+    end
+
+    return prefix == "" and "/" or prefix
```
bungle, line 329: Remind me, didn't we conclude that any normalization of this value is unneeded?

kong/pdk/request.lua
```diff
@@ -273,15 +273,60 @@ local function new(self)
...
+  --
+  -- Whether this function considers `X-Forwarded-Prefix` or not depends on
+  -- several Kong configuration parameters:
+  --
+  -- * [trusted\_ips](https://getkong.org/docs/latest/configuration/#trusted_ips)
+  -- * [real\_ip\_header](https://getkong.org/docs/latest/configuration/#real_ip_header)
+  -- * [real\_ip\_recursive](https://getkong.org/docs/latest/configuration/#real_ip_recursive)
+  --
+  -- **Note**: we do not currently do any normalization on the request path prefix
+  --           except return `"/"` on empty prefix.
```
bungle, line 307: No normalization, I would vote.

bungle (changes requested):

t/01-pdk/04-request/19-get_forwarded_prefix.t
```diff
@@ -0,0 +1,77 @@
...
+use t::Util;
+
+$ENV{TEST_NGINX_CERT_DIR} ||= File::Spec->catdir(server_root(), '..', 'certs');
+$ENV{TEST_NGINX_NXSOCK}   ||= html_dir();
+
+plan tests => repeat_each() * (blocks() * 3);
+
+run_tests();
+
+__DATA__
```
dndx, line 13: Should we include a case where the IP is not trusted as well?

dndx (commented):
Minor recommendation but the PR looks good to me.

t/01-pdk/04-request/19-get_forwarded_prefix.t
gszr: Added one test case covering that. Since the not-trusted case uses a value set by the router / runloop handler, the value here is simply nil.

bungle (approved):
wait for green, otherwise it is good to go!

PR #6259, opened 2020-08-20
fix(*) balancer redefined warning

### LINKED ISSUE #6242: [warn] load balancing method redefined ... in error.log since migrating to 2.1.0
### Summary

Since migrating from kong 2.0.4 to 2.1.0 i can see a warning
[warn] 187505#0: load balancing method redefined in <KONG_PREFIX>/nginx-kong.conf:57
in <KONG_PREFIX>/logs/error.log
Message also appears there even after upgrade to 2.1.1 and 2.1.2

### Steps To Reproduce

1. Migrate from 2.0.4 to 2.1.0
2. start/restart kong

### Additional Details & Logs
There was one change in nginx_kong.lua which seems the cause for the warnig:

nginx_kong.lua since 2.1.0:
> if (role == "traditional" or role == "data_plane") and #proxy_listeners > 0 then
upstream kong_upstream {
    server 0.0.0.1;

    # injected nginx_upstream_* directives
> for _, el in ipairs(nginx_upstream_directives) do
    $(el.name) $(el.value);
> end

    balancer_by_lua_block {
        Kong.balancer()
    }
}

nginx_kong.lua 2.0.4:
> if (role == "traditional" or role == "data_plane") and #proxy_listeners > 0 then
upstream kong_upstream {
    server 0.0.0.1;
    balancer_by_lua_block {
        Kong.balancer()
    }

    # injected nginx_upstream_* directives
> for _, el in ipairs(nginx_upstream_directives) do
    $(el.name) $(el.value);
> end
	 
}

The order of the balancer and injected nginx_upstream_* directives changed. If i change the order back to the old version the warning no longer occurs.

bungle:
### Summary

This PR contains two commits:

#### chore(*) lower a log statement in global patches from `warn` to `notice`

On `init_worker` context we had this:

```
ngx.log(ngx.WARN, "executing a blocking 'sleep' (", s, " seconds)")
```

I decided to change it to:
```
ngx.log(ngx.NOTICE, "executing a blocking 'sleep' (", s, " seconds)")
```

This is not unexpected.

#### fix(conf) change order of balancer block to prevent a warning in logs

@michaelhsk reported on #6242 that starting Kong `2.1.x` will output following
to `error.log`:

```
2020/07/07 22:13:02 [warn] 88853#0: load balancing method redefined in /usr/local/kong/nginx-kong.conf:60
```

This commit fixes the issue by changing the order on our generated templates,
back to how they were before (`< 2.1.x`).

### Issues Resolver

Fix #6242


kong/templates/nginx_kong.lua
```diff
@@ -59,16 +59,14 @@ init_worker_by_lua_block {
...
     server 0.0.0.1;
-
     # injected nginx_upstream_* directives
 > for _, el in ipairs(nginx_upstream_directives) do
     $(el.name) $(el.value);
 > end
-
-    balancer_by_lua_block {
-        Kong.balancer()
-    }
```
thibaultcha, line 71: This change was voluntary in 5c1a7bb96e6aee1579552d1dff06778359e195b6 as part of the work for dynamic upstream keepalive pools. In the currently most probable scenario (in which a user has both a leftover `nginx_upstream_keepalive` property _and_ an `upstream_keepalive_pool_size` one, it is important that the balancer block be defined _after_ the injected `keepalive` directive, in order for it to take precedence over Nginx's stock keepalive module. The warning serves its purpose and indicated to users that they should remove their old, deprecated `nginx_upstream_keepalive` property, and it will go away.

bungle: @thibaultcha, got it. I force pushed a better fix.

hishamhm: @thibaultcha Is there a difference in user-visible behavior with the change of defaults produced in the latest version of this PR (de6698c)? (I mean, other than fixing broken cases that dynamic upstream keepalive pools fix, of course)

Being a change of defaults, without a deeper understanding of what's going on here, I would assume so, but I'm unclear if it's the case. If that's the case, would this this be a breaking change (people upgrading without touching their configs would end up with different behavior), and thus to be done only in 3.0, right?

thibaultcha: Whichever of `balancer_by_lua` or Nginx's stock `keepalive` directive is defined _last_ will take effect. If both `balancer_by_lua` (with the keepalive patch) _and_ the `keepalive` directive are defined (in whichever order), then the warning will be shown. Once again, the goal was that in this case, the dynamic pool keepalive would take precedence.

By keeping these defaults, I aimed at making it easier for users to revert to the stock upstream keepalive in case the new one had any issue we weren't aware of at time of the release. It's a slower roll-out of the feature with a warning for users to fix their configuration (avoid having both properties set will log both a configuration warning from our conf loader _and_ this balancer warning from Nginx). It's also worth noting that many production deployments probably have an overriden value for this setting, other than the default, and so the warnings probably show up regardless of any work we do to optimize default configuration setups.

By removing the defaults like in the above commit, this means that the warning goes away but that reverting to the stock keepalive module requires `upstream_keepalive_pool_size = 0` _and_ `nginx_upstream_keepalive > 0`. I think that's fine too.

PR #6280, opened 2020-08-28
feat(hybrid) use timer to send changes to dp nodes instead of on every change

bungle:
### Summary

Tools like `decK` etc. can make a lot of changes to `CP`. At the moment the `CP` notifies the workers that have connected to it on each change. This may cause huge amounts of changes on a cluster:

1. There is 1 CP
2. decK makes 10 changes to database  (in 200 ms)
3. There are 20 DPs connected to that CP
4. The config is sent 200 times

This commit changes it so that the changes are sent every `x seconds`, controlable with
`db_update_frequency`.

1. There is 1 CP
2. decK makes 10 changes to database (in 200 ms)
3. There are 20 DPs connected to that CP
4. The config is sent 20 times

This of course adds latency for changes to appear in DP nodes.

### DIFF
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -12,7 +12,10 @@ local openssl_x509 = require("resty.openssl.x509")
 local assert = assert
 local setmetatable = setmetatable
 local type = type
+local math = math
+local pairs = pairs
 local ipairs = ipairs
+local ngx = ngx
 local ngx_log = ngx.log
 local ngx_sleep = ngx.sleep
 local cjson_decode = cjson.decode
@@ -46,6 +49,9 @@ local shdict = ngx.shared.kong_clustering -- only when role == "control_plane"
 local prefix = ngx.config.prefix()
 local CONFIG_CACHE = prefix .. "/config.cache.json.gz"
 local declarative_config
+
+
+local PUSH_CONFIG_CHANGES = false
 
 
 local function update_config(config_table, update_cache)
@@ -351,6 +357,23 @@ local function push_config(config_table)
 end
 
 
+local function push_config_timer(premature)
+  if premature or not PUSH_CONFIG_CHANGES then
+    return
+  end
+
+  PUSH_CONFIG_CHANGES = false
+
+  local config_table, err = declarative.export_config()
+  if not config_table then
+    ngx_log(ngx_ERR, "unable to export config from database: " .. err)
+    return
+  end
+
+  push_config(config_table)
+end
+
+
 local function init_mtls(conf)
   local f, err = io_open(conf.cluster_cert, "r")
   if not f then
@@ -465,14 +488,11 @@ function _M.init_worker(conf)
       end
     end, "dao:crud")
 
-    kong.worker_events.register(function(data)
-      local res, err = declarative.export_config()
-      if not res then
-        ngx_log(ngx_ERR, "unable to export config from database: " .. err)
-      end
-
-      push_config(res)
+    kong.worker_events.register(function()
+      PUSH_CONFIG_CHANGES = true
     end, "clustering", "push_config")
+
+    ngx.timer.every(conf.db_update_frequency or 5, push_config_timer)
   end
 end
 
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -52,6 +52,13 @@ local declarative_config
 
 
 local PUSH_CONFIG_CHANGES = false
+local LAST_PUSHED = 0
+
+
+local function get_updated_time()
+  ngx.update_time()
+  return ngx.now()
+end
 
 
 local function update_config(config_table, update_cache)
@@ -353,24 +360,29 @@ local function push_config(config_table)
     n = n + 1
   end
 
+  LAST_PUSHED = get_updated_time()
+
   ngx_log(ngx_DEBUG, "config pushed to ", n, " clients")
 end
 
 
-local function push_config_timer(premature)
+local function push_config_timer(premature, delay)
   if premature or not PUSH_CONFIG_CHANGES then
     return
   end
 
-  PUSH_CONFIG_CHANGES = false
+  local now = get_updated_time()
+  if now - LAST_PUSHED >= delay then
+    PUSH_CONFIG_CHANGES = false
 
-  local config_table, err = declarative.export_config()
-  if not config_table then
-    ngx_log(ngx_ERR, "unable to export config from database: " .. err)
-    return
+    local config_table, err = declarative.export_config()
+    if not config_table then
+      ngx_log(ngx_ERR, "unable to export config from database: " .. err)
+      return
+    end
+
+    push_config(config_table)
   end
-
-  push_config(config_table)
 end
 
 
@@ -492,7 +504,7 @@ function _M.init_worker(conf)
       PUSH_CONFIG_CHANGES = true
     end, "clustering", "push_config")
 
-    ngx.timer.every(conf.db_update_frequency or 5, push_config_timer)
+    ngx.timer.every(1, push_config_timer, conf.db_update_frequency)
   end
 end
 
--- a/spec/02-integration/09-hybrid_mode/01-sync_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/01-sync_spec.lua
@@ -17,8 +17,10 @@ for _, strategy in helpers.each_strategy() do
         cluster_cert_key = "spec/fixtures/kong_clustering.key",
         lua_ssl_trusted_certificate = "spec/fixtures/kong_clustering.crt",
         database = strategy,
+        db_update_frequency = 1,
         cluster_listen = "127.0.0.1:9005",
         nginx_conf = "spec/fixtures/custom_nginx.template",
+
       }))
 
       assert(helpers.start_kong({
@@ -125,7 +127,7 @@ for _, strategy in helpers.each_strategy() do
           if status == 404 then
             return true
           end
-        end, 5)
+        end, 10)
       end)
     end)
   end)
--- a/spec/02-integration/09-hybrid_mode/03-pki_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/03-pki_spec.lua
@@ -16,6 +16,7 @@ for _, strategy in helpers.each_strategy() do
         cluster_cert = "spec/fixtures/kong_clustering.crt",
         cluster_cert_key = "spec/fixtures/kong_clustering.key",
         lua_ssl_trusted_certificate = "spec/fixtures/kong_clustering.crt",
+        db_update_frequency = 1,
         database = strategy,
         cluster_listen = "127.0.0.1:9005",
         nginx_conf = "spec/fixtures/custom_nginx.template",
@@ -131,7 +132,7 @@ for _, strategy in helpers.each_strategy() do
           if status == 404 then
             return true
           end
-        end, 5)
+        end, 10)
       end)
     end)
   end)


bungle:
This is the most simple way I could think of to implement patching of CP -> DP messages.

This reuses the `KONG_DB_UPDATE_FREQUENCY` but we can invent another parameter, e.g.
`KONG_DP_UPDATE_FREQUENCY` ;-), if we want that.

There are other more fancy ways to do this, for example:
> Push changes only when there are no changes made within x seconds?

But I just went the dead simple route. Any feedback welcomed.

hishamhm:
@cwurm Ping :) Your feedback is especially appreciated on this!

Tieske:
Does it only reuse the setting, or will it reuse the timer?

If the CP's are a cluster then both delays kick in, checking the DB for changes, and sending to DP. So the maximum delay after which the cluster is in-sync again is 2x the setting.

When reusing the timer, sequentially pushing the changes after polling the database, the maximum delay will be 1x the setting.

fairyqb:
The configuration is sent 200 times by websocket. communication cost and data size can be ignored. The point to the problem is that when the message reaches the data plane node, the impact  is not reduced on the worker process. For example, the route reconstruction is still 200 times.right?

bungle:
> The configuration is sent 200 times by websocket. communication cost and data size can be ignored. The point to the problem is that when the message spread to data plane node, the impact is not reduced on the worker process. For example, the route reconstruction is still 200 times.right?

No. It is 1.

bungle:
> Does it only reuse the setting, or will it reuse the timer?
> 
> If the CP's are a cluster then both delays kick in, checking the DB for changes, and sending to DP. So the maximum delay after which the cluster is in-sync again is 2x the setting.
> 
> When reusing the timer, sequentially pushing the changes after polling the database, the maximum delay will be 1x the setting.

There is cp cluster delay, worker event delay x 2 and then delay that this timer might add.

bungle:
Nothing elsi is changed delay wise except the added delay of timer that can be anything between 0 and the db_update_frequency.

Tieske:
another question (might need some input from @hbagdi ); don't we run the risk of sending incomplete configurations?

Essentially we will always have that risk, but when syncing with decK, the inconsistencies will be very short-lived. Now if we, mid a decK sync hit the timer, we send the config, and then only 5 seconds (default setting) later we'll send the rest. Isn't that risky?

As an alternative we could set it up to push the config only after we haven't received any new updates for x seconds, with a timeout of y after which we always send the update.

But not sure whether this is a problem we have or not?

bungle:
> another question (might need some input from @hbagdi ); don't we run the risk of sending incomplete configurations?
> 
> Essentially we will always have that risk, but when syncing with decK, the inconsistencies will be very short-lived. Now if we, mid a decK sync hit the timer, we send the config, and then only 5 seconds (default setting) later we'll send the rest. Isn't that risky?
> 
> As an alternative we could set it up to push the config only after we haven't received any new updates for x seconds, with a timeout of y after which we always send the update.
> 
> But not sure whether this is a problem we have or not?

That's what I said there:
https://github.com/Kong/kong/pull/6280#issuecomment-682578762

The question is whether we want to `fancy` or `non fancy`. I made simple and stupid. The possible `diffs` and `incremental updates` and `dp related pushes only` plays into this too.

bungle:
> Now if we, mid a decK sync hit the timer, we send the config,
> and then only 5 seconds (default setting) later we'll send the rest

And it could still be incomplete. The deck sync will stop anywhere between 0-db_update_frequecy secs of delay. We can set another parameter too, e.g. `dp_state_update_frequency` or `dp_update_frequency` ;-). Users who are concerned about the delays should set `db_update_frequency`(if running cluster) and `dp_update_frequency` to small number.

Fancier would be allow batching of `admin ap` calls, but that is a lot harder. We could perhaps also allow `:8001/config` with `db mode`?

dndx:
@bungle The behavior in this PR makes sense to me. However, I felt that it may be necessary to make this delayed sync behavior opt in. There are cases when user wants to have minimum sync delay between API calls and DP reloads (in fact we advertise it as a feature for Hybrid mode).

Since decK sync does not happen all the time, after the initial sync is over then further updates are less and should be pushed to the DP ASAP. It seems that we should maybe let decK hint that "there might be more updates coming" when making the API call and only enable the delay for these calls (in fact the delay can even come from decK itself). That way we can have good performance versus good latency, which I consider both to be equally important.

hbagdi:
I'm happy to see the discussion going on here.

> don't we run the risk of sending incomplete configurations?

No. Kong's Admin API is RESTful, meaning at any point of time, the configuration in DB is complete in some sense. Now, I do understand that there is an "incompleteness" with respect to the configuration decK would like to have all Kong nodes to have. Unfortunately, there is no way out of it. The incompleteness is present even in the regular DB-backed mode. Since this is not specific to the hybrid mode, we shouldn't be worried about it at all. Kong's compatibility promise states that the nodes converge over time a.k.a. eventual consistency. 

> It seems that we should maybe let decK hint that "there might be more updates coming" when making the API call and only enable the delay for these calls (in fact the delay can even come from decK itself).

I would advise against it. We need to hide Kong's complexity below it's API. decK is not special and simply follows the contract laid out by Kong's Admin API. The same problem also exists with Terraform provider for Kong, which again only follows the contract laid out by Kong's Admin API.
I'll also recommend that we should NOT have any configuration property for this change for two reasons:
- Every additional configuration option is an added overhead for users and maintainers alike
- Tunable behaviors at this level should only be introduced when the shape of performance is very well understood and we are certain that a knob will make significant difference.


The motivation of the PR is correct but the implementation is not.
The implementation is resorting to using a cron-like solution where, configuration changes are sent based on a timer. While this simple solution does solve the problem, I think we can do better without a lot of effort. 
As I understand, the problem we are trying to solve here is to "rate-limit" the number of configuration updates that can be sent out to proxy nodes. If so, the control plane node should send out a configuration update as soon as it sees an event and then wait for an interval before sending out another one (and in future (out of scope for this PR) add in exponential backoff with a low ceiling). This seemingly small change has considerable impact:
- Brings down p90/p99 latency of our configuration change significantly and also solves Datong's concern (which I agree with) around impact on latency
- decK is written to minimize the number of Admin API calls required to sync the configuration. In case of a minor update, some configurations might have a single Admin API call (very common with CI/CD and git based systems). In such cases, having a fire and wait mechanism is better than a timer.

cwurm:
Thanks for the ping @hishamhm. Thanks for putting this together @bungle.

I've tested the patch and was able to load over 100K entities into a Kong CP/DP in hybrid mode. As before, I used a modified version of decK to send a generated config of 10K services and associated routes, plugins, consumers, and credentials. Execution times varied between 16 and 18 minutes (for comparison, in my testing it took just short of 13 minutes in non-hybrid mode so this is pretty good).

A few notes:
1. I had to go into the code and increase `MAX_PAYLOAD` [here](https://github.com/Kong/kong-ee/blob/f93355a44c46433bdbaf9b9927f281bc84b7c025/kong/clustering.lua#L33) to 8MB (from 4MB) after running into `payload too big` errors. We might consider making this a configuration option since it effectively sets an arbitrary hard limit on the size of the Kong configuration in hybrid mode (non-hybrid mode has no such limit from what I can see).
2. I had to set `KONG_MEM_CACHE_SIZE` to `512m` after running into `no memory` errors. This was expected.
3. I played around with settings for `KONG_DB_UPDATE_FREQUENCY` and eventually found 60 (seconds) to be the lowest to work well in this case. What seems to happen is that as the configuration gets progressively larger the time to apply it on the DP gets longer and longer. I have log statements measuring time spent at the beginning and end of the `update_config` method ([here](https://github.com/Kong/kong-ee/blob/f93355a44c46433bdbaf9b9927f281bc84b7c025/kong/clustering.lua#L53)) and while at the beginning of the decK run (when the effective Kong configuration is small) the method would execute in just under 3 seconds towards the end it takes even 77 seconds (that's the last config update on the DP with the full config). If this was a real deployment I would probably set `KONG_DB_UPDATE_FREQUENCY` to higher than 77 seconds to avoid the DP ever applying successive configuration updates back-to-back. In an earlier iteration, I had set `KONG_DB_UPDATE_FREQUENCY` to 30 seconds which led to the DP no longer being able to catch up with the rate at which the CP was sending new configs and so it was pegged at 100% CPU for the remainder of the decK run and long after (config updates will queue and be processed sequentially - which is the entire problem this PR is aiming to solve).

> I'll also recommend that we should NOT have any configuration property for this change for two reasons:
> 
> * Every additional configuration option is an added overhead for users and maintainers alike
> * Tunable behaviors at this level should only be introduced when the shape of performance is very well understood and we are certain that a knob will make significant difference.

I think we might have to make it configurable. I tried different values and it makes a huge difference to performance (how long the DPs are processing configuration) and consistency (how long it takes for configuration to be applied to all DPs).

From my testing, `KONG_DB_UPDATE_FREQUENCY` should be at least as high as the time it takes for a DP to apply the full configuration. So the value depends on the overall size of the configuration which is unknowable by us. Users can make similar testing and adjust - similar to how they can use configuration options like `worker_consistency` and `worker_state_update_frequency` in non-hybrid mode today.

> @bungle The behavior in this PR makes sense to me. However, I felt that it may be necessary to make this delayed sync behavior opt in. There are cases when user wants to have minimum sync delay between API calls and DP reloads (in fact we advertise it as a feature for Hybrid mode).

I think a delay should be the default. In my testing, the current behaviour will cause the DPs to thrash with just 150 entities - which I would consider a small configuration. Each entity is an Admin API call that causes a complete config rebuild on each DP. Since each rebuild typically takes at least a few seconds (but much longer with a larger configuration, see above) the config changes are queued and all DPs can spend a very long time at 100% CPU re-applying their entire configuration (with just one change each time) over and over again. I have seen even 150 entities take almost 10 minutes to apply to all DPs. Larger configurations are impossible and will leave all DPs of a Kong cluster at 100% CPU forever - only a complete restart of all DPs seems to be able to stop them from continuing to apply all queued configurations one-by-one.

> As I understand, the problem we are trying to solve here is to "rate-limit" the number of configuration updates that can be sent out to proxy nodes. If so, the control plane node should send out a configuration update as soon as it sees an event and then wait for an interval before sending out another one (and in future (out of scope for this PR) add in exponential backoff with a low ceiling). This seemingly small change has considerable impact:

If I understand correctly, the behaviour would be:
1. When a CP sends a config to its DPs, it starts a timer/countdown.
2. If it receives further Admin API calls while the timer is running down, it does not send them to DPs, but makes a note to send the then-current config when the timer expires.

So the difference to this PR would be that the timer is not running continuously - but only starts when a config update is made. That makes sense to me and I like the idea of an intelligent backoff. In the meantime, I guess the timer would have to be configurable (similar to this PR) - otherwise, the value could be too low for a large configuration and cause thrashing again.

bungle:
Having continuously running timer with `ngx.timer.every` in my opinion feels more robust.

bungle:
We can also make timer to run every second, and then instead of making `boolean` check we compare times.

bungle:
I kept the timer, but made small adjustments in this commit:
https://github.com/Kong/kong/pull/6280/commits/e23445e91bbfeaf25646592ea2f76dae143f6313

bungle:
>  the config changes are queued and all DPs can spend a very long time at 100% CPU re-applying their entire configuration (with just one change each time) over and over again

I think the DPs should also have ways, similar to `:8001/config` to tell that they are busy processing the previous change, and return right away, and let dp know that they didn't apply the config, and that cp has to retry.

dndx:
@bungle No, that approach seems incorrect, letting CP retry intentionally is a bad idea, because it artificially introduces more traffic and load onto the CP.

Instead, if DP is busy applying another config, it should instead queue the new config and keep only the latest config and always apply that once it has a chance. We should not require CPs to do more work than it already is because that makes the system less stable.

Couple other things I have in mind:

`ngx.timer.every` is ill-suited for this kind of use case, making the delay milliseconds isn't going to work either because we risk saturating the timer queue if config updates are slow to export, which causes timers to be randomly dropped.

Instead, we should start a timer thread once at `init_worker` phase, and use [ngx.semaphore](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/semaphore.md) to sync between worker events and the timer thread. That way the overhead is much lower, the push delay is essentially eliminated, and we will not run into timer saturation ever.

PR #6234, opened 2020-08-15
feat/plugins-rate-limiting by path. 

fairyqb:
**Summary**

add plugins-rate-limiting limit by path.

PR #6202, opened 2020-08-07
feat/plugins-rate-limiting limit by path

taotao2014:
### Summary

plugins-rate-limiting limit by path

PR #6286, opened 2020-08-31
feat(rate-limiting) implement rate limiting by path. 

kikito:
This PR expands #6234 (by @fairyqb) with a couple integration tests as well as changing the way the ratelimiting is calculated (in the original PR `conf.path` was just ignored by `kong.request.get_path`).

Left my changes as separate commits for easier review, but this PR should be Squashed and Merged into a single commit when approved.

Closes #6234. Closes #6202

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -71,7 +71,7 @@ local function get_identifier(conf)
   elseif conf.limit_by == "path" then
     local req_path = kong.request.get_path()
     if req_path == conf.path then
-      identifier = kong.request.get_path()
+      identifier = req_path
     end
   end
 
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -67,6 +67,9 @@ local function get_identifier(conf)
 
   elseif conf.limit_by == "header" then
     identifier = kong.request.get_header(conf.header_name)
+
+  elseif conf.limit_by == "path" then
+    identifier = kong.request.get_path(conf.path)
   end
 
   return identifier or kong.client.get_forwarded_ip()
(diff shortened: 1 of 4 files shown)


fairyqb:
title right ?

kong/plugins/rate-limiting/handler.lua
```diff
@@ -67,6 +67,12 @@ local function get_identifier(conf)
 
   elseif conf.limit_by == "header" then
     identifier = kong.request.get_header(conf.header_name)
+
+  elseif conf.limit_by == "path" then
+    local req_path = kong.request.get_path()
+    if req_path == conf.path then
+      identifier = kong.request.get_path()
```
gszr, line 74:
```suggestion
      identifier = req_path
```

kong/plugins/rate-limiting/handler.lua
```diff
@@ -67,6 +67,12 @@ local function get_identifier(conf)
 
   elseif conf.limit_by == "header" then
     identifier = kong.request.get_header(conf.header_name)
+
+  elseif conf.limit_by == "path" then
+    local req_path = kong.request.get_path()
+    if req_path == conf.path then
+      identifier = req_path
+    end
```
fairyqb, line 75: Whether it is necessary to ignore case ?

fairyqb:
1. consumer and service should be “logic and” relationships.
2. credential and service should be “logic and” relationships.

kong/plugins/rate-limiting/handler.lua
kikito: While some websites/services might decide to make their urls case-insensitive, the [html standard considers them sensitive](https://www.w3.org/TR/WD-html40-970708/htmlweb.html), so `/foo` and `/FOO`, in principle, point to different resources.

Case-sensitivity has a cluster of complexity in dealing with non-English letters - for example the Cyrillic and Greek alphabets have cased letters which can be used in paths, with encodings (see https://www.w3.org/International/articles/idn-and-iri/#path).

kong/plugins/rate-limiting/handler.lua
```diff
@@ -67,6 +67,12 @@ local function get_identifier(conf)
 
   elseif conf.limit_by == "header" then
     identifier = kong.request.get_header(conf.header_name)
```
fairyqb, line 69: Only header names have fewer scenarios, and more headers and values, I think.

PR #6289, opened 2020-09-01
feat(pdk) kong.service.response.get_upstream

fairyqb:
### Summary

PDK: 
**kong.service.response.get_upstream**
return upstream server IP address and port or the path to the UNIX-domain socket

The upstream service has multiple target nodes, when one of them has a business problem, it is useful to quickly resolve the issue by returning the response header to the client.

### DIFF
--- a/kong/pdk/service/response.lua
+++ b/kong/pdk/service/response.lua
@@ -150,9 +150,11 @@ local function new(pdk, major_version)
   function response.get_upstream()
     check_phase(header_body_log)
 
-    local addrs = split(ngx.var.upstream_addr, ',')
-    if #addrs > 0 then
-        return strip(addrs[#addrs])
+    if ngx.var.upstream_addr then
+      local addrs = split(ngx.var.upstream_addr, ',')
+      if #addrs > 0 then
+          return strip(addrs[#addrs])
+      end
     end
 
     return nil
--- a/kong/pdk/service/response.lua
+++ b/kong/pdk/service/response.lua
@@ -6,6 +6,7 @@
 local cjson = require "cjson.safe".new()
 local multipart = require "multipart"
 local phase_checker = require "kong.pdk.private.phases"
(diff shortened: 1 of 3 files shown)


fairyqb:
My local test was successful, but the server check failed? Why?

fairyqb:
The Travis CI build failed  ??

2020/09/02 00:43:17 [error] 20583#0: *2 [kong] workspaces.lua:12 [off] cannot create 'workspaces' entities when not using a database, context: ngx.timer

kong/pdk/service/response.lua
```diff
@@ -136,6 +139,28 @@ local function new(pdk, major_version)
...
+  -- @treturn string|nil the ip:port from the response from the Service, or `nil`
+  -- @usage
+  -- kong.log.inspect(kong.service.response.get_upstream()) -- 127.0.0.1:80
+  function response.get_upstream()
+    check_phase(header_body_log)
+
+    if ngx.var.upstream_addr then
+      local addrs = split(ngx.var.upstream_addr, ',')
+      if #addrs > 0 then
+          return strip(addrs[#addrs])
```
kikito, line 156: This needs a little bit more. According to http://nginx.org/en/docs/http/ngx_http_upstream_module.html ,

> If an internal redirect from one server group to another happens, initiated by `X-Accel-Redirect` or error_page, then the server addresses from different groups are separated by colons, e.g. `192.168.1.1:80, 192.168.1.2:80, unix:/tmp/sock : 192.168.10.1:80, 192.168.10.2:80`

My suggestion is also splitting by `:`:

```suggestion
      addrs = split(addrs, ':')
      if #addrs > 0 then
        return strip(addrs[#addrs])]
      end
```

kikito:
I'm closing this due to lack of activity. Please reopen if needed. Thanks for taking the time to send a PR!

PR #6292, opened 2020-09-01
chore(ci) naive retry implementation

hutchic:
Our tests don't necessarily cleanup sufficiently that a retry won't make things worse but this should capture some failure modes


gszr:
I think our current test suite is much more stable than it used to be at the time. Let's close this for now and revisit if the need arises at a later moment.

PR #6293, opened 2020-09-01
feat(hybrid) use timer and semaphore to push changes to dp nodes 

bungle:
### Summary

Tools like `decK` etc. can make a lot of changes to `CP`. At the moment the `CP` notifies the workers that have connected to it on each change. This may cause huge amounts of changes on a cluster:

1. There is 1 CP
2. decK makes 10 changes to database  (in 200 ms)
3. There are 20 DPs connected to that CP
4. The config is sent 200 times

This commit changes it so that after changes are send we will forcibly sleep for `db_update_frequency`, until we try to check new changes by waiting for a semaphore.

1. There is 1 CP
2. decK makes 10 changes to database (in 200 ms)
3. There are 20 DPs connected to that CP
4. The config is sent 20 times

This of course adds latency for changes to appear in DP nodes if there happens to be multiple changes in a short period of time.

Single change: no added delay
Multiple changes: the first change fast, rest after delay


bungle:
This PR is a rework of PR #6280. The biggest change is that it does not use recurring timer anymore, but it uses single timer with forever loop, semaphore, and sleeps.

bungle:
The obvious question we still have is that should we reuse `db_update_frequency` for `sleep time` or should we add a new configuration parameter?

bungle:
The relevant commit is this:
https://github.com/Kong/kong/pull/6293/commits/6f8dd29afa5d7d75bb56687a6556bb55210020c1

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -17,6 +17,7 @@ for _, strategy in helpers.each_strategy() do
         cluster_cert_key = "spec/fixtures/kong_clustering.key",
         lua_ssl_trusted_certificate = "spec/fixtures/kong_clustering.crt",
         database = strategy,
+        db_update_frequency = 1,
```
dndx, line 20: For testing purpose, this can be set lower, say to `0.1` second to help test converges quicker.

spec/02-integration/09-hybrid_mode/03-pki_spec.lua
```diff
@@ -16,6 +16,7 @@ for _, strategy in helpers.each_strategy() do
         cluster_cert = "spec/fixtures/kong_clustering.crt",
         cluster_cert_key = "spec/fixtures/kong_clustering.key",
         lua_ssl_trusted_certificate = "spec/fixtures/kong_clustering.crt",
+        db_update_frequency = 1,
```
dndx, line 19: Same here.

kong/clustering.lua
```diff
@@ -351,6 +363,29 @@ local function push_config(config_table)
...
+
+  while not exiting() do
+    local ok, err = semaphore:wait(10)
+    if ok then
+      ok, err = pcall(push_config)
+      if ok then
+        ngx.sleep(delay)
+
+      else
+        ngx_log(ngx_ERR, "pushing config failed: ", err)
```
dndx, line 379:
```suggestion
        ngx_log(ngx_ERR, "export and pushing config failed: ", err)
```

kong/clustering.lua
```diff
@@ -466,13 +503,12 @@ function _M.init_worker(conf)
     end, "dao:crud")
 
     kong.worker_events.register(function(data)
-      local res, err = declarative.export_config()
-      if not res then
-        ngx_log(ngx_ERR, "unable to export config from database: " .. err)
+      if push_config_semaphore:count() <= 0 then
+        push_config_semaphore:post()
```
dndx, line 507:
```suggestion
      if push_config_semaphore:count() <= 0 then
        -- the following line always executes immediately after the `if` check
        -- because `:count` will never yield, end result is that the semaphore
        -- count is guaranteed to not exceed 1
        push_config_semaphore:post()
```

I felt it is a piece of knowledge that could be easily lost over time so better to be more verbose.

dndx (approved):
A simple test case would be great, but this LGTM.

PR #6297, opened 2020-09-02
feat/pdk.kong.response.redirect

fairyqb:
### Summary

PDK:
kong.response.redirect("http://github.com/kong/kong")

Easy use and understanding.

### DIFF
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -12,6 +12,7 @@
 -- @module kong.response
 
 
+local url = require "socket.url"
 local cjson = require "cjson.safe"
 local meta = require "kong.meta"
 local checks = require "kong.pdk.private.checks"
@@ -66,6 +67,7 @@ local function new(self, major_version)
   local MAX_STATUS_CODE      = 599
   local MIN_ERR_STATUS_CODE  = 400
 
+  local LOCATION_HEADER_NAME = "Location"
   local SERVER_HEADER_NAME   = "Server"
   local SERVER_HEADER_VALUE  = meta._NAME .. "/" .. meta._VERSION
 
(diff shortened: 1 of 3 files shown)


kong/pdk/response.lua
```diff
@@ -472,6 +474,28 @@ local function new(self, major_version)
...
+  ---
+  -- Redirect uri in the response sent to the client.
+  --
+  -- @function kong.response.redirect
+  -- @phases rewrite, access, header_filter, admin_api
+  -- @tparam string uri The name of the url to be redirect
+  -- @return Nothing; throws an error on invalid input.
+  -- @usage
+  -- kong.response.redirect("https://github.com/kong/kong")
+  function _RESPONSE.redirect(uri)
```
guanlan, line 486: Suggest that to add an optional status parameter here as `_RESPONSE.redirect(uri, status?)`, keep the consistency with https://github.com/openresty/lua-nginx-module#ngxredirect.

kong/pdk/response.lua
```diff
@@ -472,6 +474,28 @@ local function new(self, major_version)
...
+    check_phase(rewrite_access_header)
+
+    local parsed_uri = url.parse(uri)
+    if not (parsed_uri and parsed_uri.host and parsed_uri.scheme) then
+      error("cannot parse '" .. uri .. "'", 2)
+    end
+
+    ngx.header[LOCATION_HEADER_NAME] = uri
+    return ngx.exit(301)
```
bungle, line 495: `302` is in general a better default, but it should also be an argument.

We also need to construct body. Here is what Nginx does: http://lxr.nginx.org/source/src/http/ngx_http_special_response.c

We can add Kong twist to it.

This should also use `kong.response.exit` to make short-circuiting work correctly. I am also not sure about the validation there. Is it needed, is it correct?

bungle (changes requested):
Some comments

kong/pdk/response.lua
fairyqb: Easy use and understanding.

bungle: Just a note here, with our current PDK, you can do this:

```lua
return kong.response.exit(302, "", { Location: "<uri>" })

**Comment by kikito** [comment]
@fairyqb we are very grateful for the time and effort you put into this PR, but I have decided not to include it.

As @bungle has mentioned, redirects can be currently implemented with:

```
return kong.response.exit(302, "", { Location: "<uri>" })
```

or:

```
return kong.response.exit(302, nil, { Location: "<uri>" })
```

Your proposal ultimately amounts to a slightly simpler set of params, and checking that the `uri` param is valid. However it has maintenance cost in the form of tests and documentation. I feel that the benefit does not justify the cost in this particular case. In order to avoid other misunderstanding like this, I suggest opening an issue and asking whether it is needed first.

Please don't let this prevent you from sending other PRs in the future! Again, we deeply appreciate it.

PR #6298, opened 2020-09-02
tests(*) plugins grpc 2

gszr:
### Summary

Add grpc test cases testing basic plugin functionality:
- request-termination
- rate-limiting
- response-rate-limiting: for this plugin, only rate-limiting headers
were tested, as the tests rely on manipulating response headers, which
is not currently possible with our mock grpc service


kikito (approved):
Everything here LGTM - reminder that request-size-limiting is still remaining.

PR #6299, opened 2020-09-02
feat(hybrid) do not queue updates on dp nodes 

bungle:
### Summary

Currently when CP sends new config to DP the DP will apply that config even in case it is already applying a config, and even in case when CP sends a new config (which means 2 configs queued). Only the last config is needed. This PR changes it so that DP drops the unnecessary configs that it knows are going to be replaced right after.

This commit is continuation of the work that started with #6293 that made the CP skip the configs that it knows are already going to be replaced. This just makes the same thing on DP side.

You might wonder why is that? With huge configs, it can take a lot of time for DP to apply the configuration, @cwurm told us delays like 77 seconds in this message: #6280 (comment)

That's why the CP can send multiple new configs during the time DP is applying the previous ones. This commit is to fix that.

### DIFF
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -50,6 +50,7 @@ local shdict = ngx.shared.kong_clustering -- only when role == "control_plane"
 local prefix = ngx.config.prefix()
 local CONFIG_CACHE = prefix .. "/config.cache.json.gz"
 local declarative_config
+local next_config
 
 
 local function update_config(config_table, update_cache)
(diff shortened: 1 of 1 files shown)


kong/clustering.lua
```diff
@@ -161,7 +164,36 @@ local function communicate(premature, conf)
...
+            end
+
+          else
+            ngx_log(ngx_ERR, "unable to update running config: ", res)
+          end
+        end
+
+        if next_config == config_table then
+          next_config = nil
+        end
```
dndx, line 188: Minor point: these conditions can be moved to inside the `if ok` block above.

dndx (approved):
Seems to be correct to me, with a minor recommendation.

PR #6303, opened 2020-09-03
perf(worker-events) change interval from 1 seconds to 100 ms

bungle:
### Summary

We have had worker events interval being 1 secs for over a four years:
https://github.com/Kong/kong/commit/1999d926f22ffb60de470c62d291179f46ccf0a1

While that has served us well, I started questioning that is it too high?

This setting affects the latency of Kong in many places, one such place is CP/DP sync, which can add latency of over 2 secs (as there are two levels of worker events `post_local` for `dao:crud` and `post` for `clustering:push_config`.

This PR lowers the interval from 1 secs to 100 ms. Yes, it makes a little bit more load on workers, but on the otherhand it spreads events more evenly.

Further enhancements:

We can PR worker events library to not re-create timers, but just sleep between polls.


bungle:
Here is the `sister` PR to this:
https://github.com/Kong/lua-resty-worker-events/pull/36

hishamhm:
@bungle An interesting point to raise! I wonder if the 1s number was picked "semi-randomly" or if there was any tuning that led to this number.

@Tieske any thoughts on this PR?

bungle:
> @bungle An interesting point to raise! I wonder if the 1s number was picked "semi-randomly" or if there was any tuning that led to this number.
> 
> @Tieske any thoughts on this PR?

1 sec was taken randomly. Anything more than that is probably bad for `kong quit` or `nginx -s quit`. Shorter than that is probably too frequent.

bungle:
> @bungle An interesting point to raise! I wonder if the 1s number was picked "semi-randomly" or if there was any tuning that led to this number.
> 
> @Tieske any thoughts on this PR?

The `0.5` was picked because I tried others. E.g. `0.1` and `0.25`. Those two values started to make our test suite flaky. I have seen some flakiness even with `0.5`, which is interesting. I thought that raising this value would make our test suite more flaky, but it seems to get more flaky even with smaller values.

Tieske:
we should set the timer to 1 or 2 seconds, and then also poll on each request. Polling is cheap, it checks only a single counter value in the shm.

The timer has always been a backup for polling (eg. no traffic, so using a timer to stay current)

The test suite problems are a different problem (but I get the reasoning obviously), maybe add these to a sprint?

dndx:
Related: https://github.com/Kong/kong/issues/6709

PR #6308, opened 2020-09-04
feat(clustering) use database for persisting DP status information

dndx:


dndx:
Here is an example output:

```
$ http :8001/cluster_status
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 213
Date: Fri, 04 Sep 2020 15:56:50 GMT
Server: kong/2.1.2
X-Kong-Admin-Latency: 145

{
    "data": [
        {
            "config_hash": "a9a166c59873245db8f1a747ba9a80a7",
            "hostname": "localhost.localdomain",
            "id": "ed58ac85-dba6-4946-999d-e8b5071607d4",
            "ip": "127.0.0.1",
            "last_seen": 1599235004,
            "status": "connected"
        }
    ],
    "next": null
}
```

kong/db/schema/entities/cluster_status.lua
```diff
@@ -0,0 +1,28 @@
...
+    { ip = typedefs.ip { required = true, } },
+    { config_hash = { type = "string", len_eq = 32, } },
+    { hostname = typedefs.host { required = true, } },
+  },
+
+  transformations = {
+    {
+      input = { "last_seen" },
+      on_read = function(last_seen)
+        if ngx_time() - last_seen > 60 then
```
hbagdi, line 20: I would recommend to avoid adding the "status" field at this point as this is an API around which we do not have enough operational feedback. The consumer of the API can interpret the last_seen time as they see fit.

If we do go forward with this addition, please define `60` as a constant.

kong/api/init.lua
```diff
@@ -28,7 +28,7 @@ ngx.log(ngx.DEBUG, "Loading Admin API endpoints")
 
 
 -- Load core routes
-for _, v in ipairs({"kong", "health", "cache", "config", "clustering"}) do
+for _, v in ipairs({"kong", "health", "cache", "config", }) do
```
hbagdi, line 31: Why was this not renamed and removed?
Will the code actually pick up the endpoints defined in `cluster_status` if it is not included in this list?

kong/api/routes/cluster_status.lua
```diff
@@ -1,18 +1,19 @@
-local clustering = require("kong.clustering")
-local kong = kong
+local endpoints = require "kong.api.endpoints"
 
 
 return {
-  ["/clustering/status"] = {
```
hbagdi, line 6: Note: I'm fine with breaking this but if we do do this, please ensure that it is called out in the changelog.

kong/api/routes/cluster_status.lua
```diff
@@ -1,18 +1,19 @@
...
-    GET = function(self, db, helpers)
+  ["/cluster_status"] = {
```
hbagdi, line 5: We should do some forward thinking around naming and user-facing API here, some raw thoughts:
-  This endpoint returns is  a list of nodes connected and not the health of the overall cluster. We should reconsider the name of this endpoint. 
- `/cluster` could be a namespace under which we can introduce more endpoints in future
- `/cluster/nodes` or `/cluster/nodes/proxy` should be the name of this endpoint
- In future, `/cluster/nodes?type=control` could also provide all the control plane nodes that are part of the cluster 
- In future, the same endpoint can be potentially introduced in the "traditional" mode to provide easy observability of how many kong nodes are connected to the same db

kong/db/migrations/core/012_213_to_220.lua
```diff
@@ -0,0 +1,25 @@
+return {
+  postgres = {
+    up = [[
+      CREATE TABLE IF NOT EXISTS "cluster_status" (
```
hbagdi, line 4: Not to bike shed here, but does it make sense to call this `cluster_nodes` instead?

kong/api/routes/cluster_status.lua
hbagdi: Also please note that the API will need a DELETE endpoint at some point to prune the nodes that are dead in the cluster. While most users will be okay with regular automated pruning done by control plane nodes but I expect the above to be also come up.

kong/db/schema/entities/cluster_status.lua
kikito: I like this usage of transformations. Elegant 🎩

spec/01-unit/01-db/01-schema/13-cluster_status_spec.lua
```diff
@@ -0,0 +1,42 @@
+require "spec.helpers" -- initializes 'kong' global for plugins
+local Entity = require "kong.db.schema.entity"
+local cluster_status_definition = require "kong.db.schema.entities.cluster_status"
```
kikito, line 3: This variable could be renamed to `cluster_status_schema`. The term "definition" is usually reserved for the table we use to create a schema (`schema = Schema.new(definition)`).

spec/01-unit/01-db/01-schema/13-cluster_status_spec.lua
```diff
@@ -0,0 +1,42 @@
...
+
+    assert.equal("invalid value: !", err.hostname)
+    assert.equal("not an ip address: aabbccdd", err.ip)
+  end)
+
+  it("accepts correct value", function()
+    local ok, err = validate({ ip = "127.0.0.1", hostname = "dp.example.com", })
+    assert.is_true(ok)
+    assert.is_nil(err)
+  end)
```
kikito, line 41: We should check that the transformation works correctly here as well by setting `last_seen` and then reading `status` here.

spec/02-integration/03-db/13-cluster_status_spec.lua
```diff
@@ -0,0 +1,48 @@
...
+  describe("kong.db [#" .. strategy .. "]", function()
+    local db, bp, cs
+
+    lazy_setup(function()
+      bp, db = helpers.get_db_utils(strategy, {
+        "cluster_status",
+      })
+    end)
+
+    describe("Plugins #plugins", function()
```
kikito, line 14:
```suggestion
    describe("Cluster Status", function()
```

kong/db/schema/entities/cluster_status.lua
dndx: This was a recommendation from the ADR, personally I am okay with removing it for now as it does not have practical use cases yet. We can always add it back in the future.

kikito (changes requested):
It looks good, only very minor changes requested.

kong/db/migrations/core/012_213_to_220.lua
dndx: What about `cluster_data_planes` and `/cluster/data_planes` respectively? We do use this term prominently in the config interface for Hybrid mode.

spec/01-unit/01-db/01-schema/13-cluster_status_spec.lua
dndx: I think per @hbagdi 's suggestion, we can temporarily leave out the `status` attribute in this version of implementation until more concrete use cases are shown. It's always easier to add than changing/removing these.

kong/api/init.lua
dndx: This is not necessary for DAO routes, they are automatically loaded by Kong. For example, `kong/api/routes/plugins.lua` is also not in the above list.

kong/api/routes/cluster_status.lua
dndx: @hbagdi Yes, we will probably add these (including auto-deletion of unneeded rows) after the alpha.

kikito (approved):

kong/api/routes/clustering.lua
```diff
@@ -1,18 +1,23 @@
...
 return {
-  ["/clustering/status"] = {
-    GET = function(self, db, helpers)
-      if kong.configuration.role ~= "control_plane" then
-        return kong.response.exit(400, {
-          message = "this endpoint is only available when Kong is " ..
-                    "configured to run as Control Plane for the cluster"
-        })
-      end
+  ["/clustering/data_planes"] = {
```
hbagdi, line 8: Should we call this `data_nodes` instead?
`data_planes` is wrong grammar at this point.

cc @hishamhm

PR #6313, opened 2020-09-07
chore(template) grpc_pass now has support for variables 

bungle:
### Summary

With latest OpenResty based on Nginx `1.7.x` series we can now use variables in `scheme` part of `grpc_pass`, thus we can now remove some repetition from our templates.

There is one additional commit in this PR that removes support for `1.15.8.x` series of OpenResty, because after this PR, the templates won't work anymore with the earlier version.


javierguerragiraldez (commented):
is there a test to verify that `gprcs://` uses TLS?

dndx:
Delayed temporarily until after 2.2.

gszr (approved):

PR #6336, opened 2020-09-11
feat(balancer) remove target history

locao:
### Summary

This Pull Request removes the target history. 

### Full changelog

* Balancer now uses new consistent hashing from lua-resty-dns-client.
* Upstreams entity API now accepts adding and patching targets.
* Targets DAO now actually deletes entries, instead of changing their weights.

**Note**: this should stay as a draft while lua-resty-dns-client 5.1.0 is not released.


kong/runloop/balancer.lua
```diff
@@ -717,41 +627,47 @@ end
...
+-- @param operation "create", "update" or "delete"
+-- @param target Target table with `upstream.id` field
+local function on_target_event(operation, target)
+  local upstream_id = target.upstream.id
+  local upstream_name = target.upstream.name
+
+  log(DEBUG, "target ", operation, " for upstream ", upstream_id,
+      upstream_name and " (" .. upstream_name ..")" or "")
+
   singletons.core_cache:invalidate_local("balancer:targets:" .. upstream_id)
```
murillopaula, line 641: This code was already here, it's not a change added by this PR, but this might an opportunity to fix it: the above target cache invalidation, along with with the same invalidation at `on_upstream_event` https://github.com/Kong/kong/blob/35baae58f8c5d4a404199ef7f731d4b8f555f7e2/kong/runloop/balancer.lua#L913 will generate excessive invalidations. It will be directly proportional to the number of nginx workers:

```number of queries = number of workers * SELECT FROM targets (query in balancer.lua) + DAO queries```

We need to make sure that only one nginx worker does the DB queries so that the other ones can retrieve the targets from the shm L2 cache. It should consequently reduce the p99 latency for balancer requests.:

```number of queries = SELECT FROM targets (query in balancer.lua) + DAO queries```

Every worker is running the L3 cache callback (performing a database query) on upstreams and targets events. But now we don't need to always fetch target histories from the DB, so now might a good opportunity to remove either the above cache invalidation or the one that's in `on_upstream_event` in order to reduce the amount of DB queries generated by the balancer. I'm not sure which invalidation one we should remove (need to think about the `strict` and `eventual` case), but I'd suggest to invalidate the cache in the `runloop/handler.lua` file within the worker event registration: https://github.com/Kong/kong/blob/35baae58f8c5d4a404199ef7f731d4b8f555f7e2/kong/runloop/handler.lua#L331-L337

or

https://github.com/Kong/kong/blob/35baae58f8c5d4a404199ef7f731d4b8f555f7e2/kong/runloop/handler.lua#L404-L410

This way all nginx workers will invalidate the `"balancer:upstreams:<target_id>"` cache _before_ the
`on_target_event()` or `on_upstream_event()` worker event callback. Again, we need to evaluate, test and determine where exactly the `"balancer:upstreams:<target_id>"` invalidation should go.

Comment (hishamhm) [reply kong/runloop/balancer.lua:641]: @murillopaula This is a great observation and I think it deserves a `fix()` PR of its own! Let's address it separately. Redundant invalidations aside, it shouldn't be a blocker for the balancing-behavior-correctness of this PR, right?

Comment (murillopaula) [reply kong/runloop/balancer.lua:641]: @hishamhm yeah maybe even a `perf()` and I'd be happy to help or contribute with this. First attempt was https://github.com/Kong/kong/pull/6033 but it was a bad solution. But yes, correct, this should not be a blocker for this PR.

File: spec/02-integration/05-proxy/10-balancer/03-consistent-hashing_spec.lua [hunk 2]
```diff
@@ -153,4 +153,4 @@ for _, strategy in helpers.each_strategy() do
       end)
     end)
   end
-end
\ No newline at end of file
+end
```
kikito, line 156: 🤔how didn't this break the build before?

kikito (approved):
Approved with one small comment/question. 272 lines in, 573 out. Well done!

spec/02-integration/05-proxy/10-balancer/03-consistent-hashing_spec.lua
hishamhm: You may be misreading the diff: this is just adding a line break after the `end`.

Issue #1501, opened 2016-08-15
OAuth 2 Plugin: Does not support HTTP GET for authorize/token endpoints.


crazy8brown:
### Summary

It appears that the OAuth plugin doesn't support  GET for the authorize/token endpoints.  I can successfully use POST, but when I attempt something similar to this:

```
curl -G https://kongsvr:8443/api/myapi/oauth2/authorize \
      -d "client_id=890cd79ec574458d923111f11fd6f29c" \
      -d "response_type=code" -d "scope=email"     \
      -d "provision_key=723e091e9dc8467187d5cf71449b8d59"  \
      -d "authenticated_userid=userid123"  --insecure 
```

I get the following error:

```
{"error_description":"The access token is missing","error":"invalid_request"}
```

I see a similar response when using GET and the token endpoint.  

As I understand it, the OAuth2 spec implies that authorize and token must support GET and can optionally support POST.  
### Steps To Reproduce
1. Configure OAuth plugin for API and Consumer
2. Issue a GET to the authorize endpoint per the description.
3. Verify that the call fails with "Access Token is Missing"
### Additional Details & Logs
- Kong Version: 0.8.3
- CentOS 6

Tieske:
0.9 has quite some oauth fixes, can you try whether your problem still exists with the latest RC?

thefosk:
The two methods are intentionally only available with `POST` because `GET` requests with credentials are insecure since they log parameters in log files. Also I personally prefer to use `POST` to create new entities like authorization codes or access tokens (`GET` shouldn't change the DB state as per the HTTP best practices).

thefosk:
The reason why `/authorize` is not `GET` it's because it's a slightly different method that makes the Kong OAuth 2.0 possible, so it's not 1:1 compliant to the spec. We had an internal talk to change the name from `/authorize` to something else (suggested by @nikz).

tbrownMT:
@thefosk I missed the token endpoint being POST only.  Is it possible to have the documentation updated then to highlight the reasoning (Kong logs params) and the difference (provision_key) between Kong's authorize and the spec's approach?  I think if the docs were clarified, it would be helpful.

tbrownMT:
Also, I'd add that I have no problem with keeping the authorize name.  I've seen other implementations of OAuth authorize that include custom params such as provision_key.

thefosk:
@tbrownMT yes - I will keep this issue open as a reminder to write some more docs around it.

kikito:
[...]

PR #6337, opened 2020-09-12
fix(oauth2) handle more client invalid token generation cases

### LINKED ISSUE #6311: 2.1.3 OAuth2.0 CC stacktrace error from bad client input
### Summary

Looks like another case where client calls token generation with missing client_id and client_secret and Kong isn't handling it properly.

```

2020/09/07 03:00:44 [error] 88#0: *53370 lua coroutine: runtime error: ...ng/luarocks/share/lua/5.1/kong/plugins/oauth2/access.lua:165: key must be a string or an entity table
--
  | stack traceback:
  | coroutine 0:
  | [C]: in function 'error'
  | /usr/local/kong/luarocks/share/lua/5.1/kong/db/dao/init.lua:1478: in function 'cache_key'
  | ...ng/luarocks/share/lua/5.1/kong/plugins/oauth2/access.lua:165: in function 'get_redirect_uris'
  | ...ng/luarocks/share/lua/5.1/kong/plugins/oauth2/access.lua:538: in function 'execute'
  | ...g/luarocks/share/lua/5.1/kong/plugins/oauth2/handler.lua:11: in function <...g/luarocks/share/lua/5.1/kong/plugins/oauth2/handler.lua:10>
  | coroutine 1:
  | [C]: in function 'resume'
  | coroutine.wrap:21: in function <coroutine.wrap:21>
  | /usr/local/kong/luarocks/share/lua/5.1/kong/init.lua:757: in function 'access'
  | access_by_lua(nginx.conf:146):2: in main chunk, client: 10.201.20.250, server: kong, request: "POST /auth/oauth2/token?client_id&grant_type=client_credentials&client_secret HTTP/1.1", host: "gateway-stage-networkZone.company.com"
```

Resulting error will lead to a HTTP Status 500 to the client generated by Kong rather than the expected HTTP 400. 

### Additional Details & Logs

- Kong version 2.1.3

jeremyjpj0916:
### Summary

Fixes: https://github.com/Kong/kong/issues/6311

^^ Full discussion and examples over here. My changes fix 2 scenarios I can see Kong not responding as I would expect to an end user. 1 was type error fix(by validating a scenario where bad user input yields a boolean reference not a string), other requires thinking about it a little deeper to return the right response to clients. 

### Full changelog

* Add better handling in ```get_redirect_uris``` when fields are not present ```client_id``` ends up being a boolean parameter. Potentially fixes could have been done alternatively in the ```retrieve_parameters``` or ```retrieve_client_credentials``` functions.

* Removed the check for ```POST``` method to decide if we are calling the proxy vs calling the token generation pathings, URI alone should help dictate it imo and thus the response customers get from Kong will be more accurate to what proxy clients were attempting to do.  

* 2 edge cases tests I fixed added.

### Issues resolved

Fix #6311

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,11 +40,11 @@ transformations, and more through plugins.
 
 Kong has been built with the following leading principles:
 
-* **High Performance**: Sub-millisecond processing latency to support mission 
+* **High Performance**: Sub-millisecond processing latency to support mission
   critical use cases and high throughput.
-* **Extensibility**: With a pluggable architecture to extend Kong in Lua or GoLang 
+* **Extensibility**: With a pluggable architecture to extend Kong in Lua or GoLang
   with Kong's Plugin SDK.
-* **Portability**: To run on every platform, every cloud and to natively support 
+* **Portability**: To run on every platform, every cloud and to natively support
   Kubernetes via our modern Ingress Controller.
 
 ## Features
@@ -52,7 +52,7 @@ Kong has been built with the following leading principles:
 - **Cloud-Native**: Platform agnostic, Kong can run on any platform - from bare
   metal to containers - and it can run on every cloud natively.
 - **Kubernetes-Native**: Declaratively configure Kong with native Kubernetes CRDs
-  using the official Ingress Controller to route and connect all L4 + L7 traffic. 
+  using the official Ingress Controller to route and connect all L4 + L7 traffic.
 - **Dynamic Load Balancing**: Load balance traffic across multiple upstream
   services.
 - **Hash-based Load Balancing**: Load balance with consistent hashing/sticky
@@ -130,6 +130,12 @@ code, other repos are also under active development:
 You can find every supported distribution at the [official installation page](https://konghq.com/install/).
 
 ## Development
+
+We encourage community contributions to Kong. To make sure it is a smooth
+experience (both for you and for the Kong team), please read
+[CONTRIBUTING.md](#CONTRIBUTING.md) , [DEVELOPER.md](#DEVELOPER.md),
+[CODE_OF_CONDUCT.md](#CODE_OF_CONDUCT.md) and [COPYRIGHT](#COPYRIGHT) before
+you start.
 
 If you are planning on developing on Kong, you'll need a development
 installation. The `next` branch holds the latest unreleased source code.


kong/plugins/oauth2/access.lua
```diff
@@ -161,7 +161,7 @@ end
 
 local function get_redirect_uris(client_id)
   local client, err
-  if client_id and client_id ~= "" then
+  if client_id and client_id ~= "" and type(client_id) ~= "boolean" then
```
bungle, line 164:
```suggestion
  if type(client_id) == "string" and client_id ~= "" then
```

kong/plugins/oauth2/access.lua
```diff
@@ -1045,20 +1045,20 @@ function _M.execute(conf)
     return
   end
 
-  if kong.request.get_method() == "POST" then
```
bungle, line 1048: The spec says:

Token Endpoint (https://tools.ietf.org/html/rfc6749#section-3.2):

> The client MUST use the HTTP "POST" method when making access token requests.

Authorization Endpoint:
>  The authorization server MUST support the use of the HTTP "GET" method [RFC2616] for the authorization endpoint and MAY support the use of the "POST" method as well.


Thus can we change it so that `POST` is checked for `token endpoint` and `GET/POST` is checked for `authorization endpoint`?

bungle (changes requested):
Some small change requests. Otherwise it looks good.

bungle:
Some early discussion:
https://github.com/Kong/kong/issues/1501

But I am fine with changing `authorize` to accept `GET` too, as that is the most common with auth code flow.

kong/plugins/oauth2/access.lua
jeremyjpj0916: The issue in my mind is Kong is placing these on proxy routes that also serve the service via reverse proxy so its not a standalone authorization server. By checking HTTP Methods and certain portions of pathing for things like ```/oauth2/*``` I think it causes some behavior that would not be what the client should get.

Example is when you send a GET to ```/oauth2/token``` the client should get a 400 bad request(Because they didn't send a POST) but because of the fall though behavior it will hit, the client will get a 401(not authorized, which isn't the correct status when the wrong method was sent). 

Are you suggesting to add the HTTP checks if the string match is found and if HTTP != POST throw a 400(thus not hitting the 401 incorrect status?) because then it covers both cases, I mean like this seems correct behavior you could achieve from a client perspective to me:

```
  local from = string_find(path, "/oauth2/token", nil, true)
    if from then
      if kong.request.get_method() == "POST" then
        return authorize(conf)	    return issue_token(conf)
     else
       {return 400 bad request}
    end	
  end

  from = string_find(path, "/oauth2/authorize", nil, true)
  if from then
      if kong.request.get_method() == "POST" or kong.request.get_method() == "GET" then
        return authorize(conf)
     else
       {return 400 bad request}
  end
```

Because if the routing URL has ```/oauth2/token``` or ```/oauth2/authorize``` the client is certainly indending to perform OAuth2.0 logic and not trying to hit the proxy without a bearer token, yeilding the 401's you would get today by sending a GET.

kong/plugins/oauth2/access.lua
jeremyjpj0916: Ah yeah good catch, much cleaner and if ```type(nil)``` check will serve the same purpose as ```if client_id``` check did.

kong/plugins/oauth2/access.lua
jeremyjpj0916: cc @bungle ^

kong/plugins/oauth2/access.lua
jeremyjpj0916: Fixed with revisions for this.

jeremyjpj0916:
@bungle committed a change reflecting the Method checks I am talking about and ensuring an error of 400 if methods are not correct yet a URI match was found for the two oauth2 specific URI endpoints (/oauth2/token and /oauth2/authorize). Tests may fail though due to introducing a new Method for authorize, I didn't check, let us see.

jeremyjpj0916:
Tests of the new checks and how they look too:

GET
https://gateway-dev-core-datacenter.company.com/auth/oauth2/token

Response:
```
HTTP/1.1 400 Bad Request
Date=Tue, 29 Sep 2020 20:33:43 GMT
Content-Type=application/json; charset=utf-8
Connection=keep-alive
WWW-Authenticate=Bearer realm="service" error="invalid_method" error_description="The HTTP method GET is invalid for the bearer token generation endpoint"
Content-Length=120
company-cid-ext=50179244-8d24-4666-8c78-e8381bd42c50#1
X-Response-Latency=1

{"error_description":"The HTTP method GET is invalid for the bearer token generation endpoint","error":"invalid_method"}
```

OPTIONS
https://gateway-dev-core-datacenter.company.com/auth/oauth2/authorize

Response:
```
HTTP/1.1 400 Bad Request
Date=Tue, 29 Sep 2020 20:35:38 GMT
Content-Type=application/json; charset=utf-8
Connection=keep-alive
WWW-Authenticate=Bearer realm="service" error="invalid_method" error_description="The HTTP method OPTIONS is invalid for the authorization endpoint"
Content-Length=114
optum-cid-ext=ce9335c2-078c-4a8e-ba87-7cdfc6292354#1
X-Response-Latency=2

{"error_description":"The HTTP method OPTIONS is invalid for the authorization endpoint","error":"invalid_method"}
```

jeremyjpj0916:
@bungle Anything further needed here?

jeremyjpj0916:
Might be better to put this on the later versions of Kong now in a next branch or something.

PR #6342, opened 2020-09-15
feat(conf) improved lua_ssl_trusted_certificate

kikito:
This PR adds two features to the `lua_ssl_trusted_certificate` option:
* It is now a comma-separated list instead of a single item. All the paths on the list are concatenated into a single file in a fixed location (`$PREFIX/.ca_combined`).
* It accepts a special `system` value, which is expanded to the "system default". This follows a very simple heuristic to try to use the most common cert in most popular distros. `system` can be used in conjunction with other paths.

Note that this PR involves modifying Kong's template - as a result, the UPGRADE.md file should include such changes when a new version is released.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1213,12 +1213,13 @@
 # See the lua-nginx-module documentation for more information:
 # https://github.com/openresty/lua-nginx-module
 
-#lua_ssl_trusted_certificate =   # Absolute path to the certificate
-                                 # authority file for Lua cosockets in PEM
-                                 # format. This certificate will be the one
-                                 # used for verifying Kong's database
-                                 # connections, when `pg_ssl_verify` or
-                                 # `cassandra_ssl_verify` are enabled.
+
+#lua_ssl_trusted_certificate =   # Comma-separated list of paths to certificate
+                                 # authority files for Lua cosockets in PEM format.
+                                 #
+                                 # When `pg_ssl_verify` or `cassandra_ssl_verify`
+                                 # are enabled, these certificate authority files will be
+                                 # used for verifying Kong's database connections.
                                  #
                                  # See https://github.com/openresty/lua-nginx-module#lua_ssl_trusted_certificate
 
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1217,6 +1217,25 @@
 #lua_ssl_trusted_certificate =   # Comma-separated list of paths to certificate
                                  # authority files for Lua cosockets in PEM format.
                                  #
+                                 # The special value `system` attempts to search for the
+                                 # "usual default" provided by each distro, according
+                                 # to an arbitrary heuristic. In the current implementation,
+                                 # The following pathnames will be tested in order,
+                                 # and the first one found will be used:
+                                 #
+                                 # * /etc/ssl/certs/ca-certificates.crt (Debian/Ubuntu/Gentoo)
+                                 # * /etc/pki/tls/certs/ca-bundle.crt (Fedora/RHEL 6)
+                                 # * /etc/ssl/ca-bundle.pem (OpenSUSE)
+                                 # * /etc/pki/tls/cacert.pem (OpenELEC)
+                                 # * /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem (CentOS/RHEL 7)
+                                 # * /etc/ssl/cert.pem (OpenBSD, Alpine)
+                                 #
+                                 # If no file is found on any of these paths, an error will
+                                 # be raised.
+                                 #
+                                 # `system` can be used by itself or in conjunction with other
+                                 # CA filepaths.
+                                 #
                                  # When `pg_ssl_verify` or `cassandra_ssl_verify`
                                  # are enabled, these certificate authority files will be
                                  # used for verifying Kong's database connections.
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -103,6 +103,23 @@ local function gen_default_ssl_cert(kong_config, target)
   return true
 end
 
+
+local function gen_trusted_certs_combined_file(combined_filepath, paths)
+
+  log.verbose("generating trusted certs combined file in ",
+              combined_filepath)
+
+  local fd = assert(io.open(combined_filepath, "w"))
+
+  for _, path in ipairs(paths) do
+    fd:write(pl_file.read(path))
+    fd:write("\n")
+  end
+
+  io.close(fd)
+end
+
+
 local function get_ulimit()
   local ok, _, stdout, stderr = pl_utils.executeex "ulimit -n"
   if not ok then
@@ -302,6 +319,13 @@ local function prepare_prefix(kong_config, nginx_custom_template_path)
     end
     kong_config.status_ssl_cert = kong_config.status_ssl_cert_default
     kong_config.status_ssl_cert_key = kong_config.status_ssl_cert_key_default
+  end
+
+  if kong_config.lua_ssl_trusted_certificate_combined then
+    gen_trusted_certs_combined_file(
+      kong_config.lua_ssl_trusted_certificate_combined,
+      kong_config.lua_ssl_trusted_certificate
+    )
   end
 
   -- check ulimit
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -566,6 +566,7 @@ local CONF_INFERENCES = {
     deprecated = { replacement = false }
   },
 
+  lua_ssl_trusted_certificate = { typ = "array" },
   lua_ssl_verify_depth = { typ = "number" },
   lua_socket_pool_size = { typ = "number" },
 
@@ -800,11 +801,13 @@ local function check_and_infer(conf, opts)
     end
   end
 
-  if conf.lua_ssl_trusted_certificate and
-     not pl_path.exists(conf.lua_ssl_trusted_certificate)
-  then
-    errors[#errors + 1] = "lua_ssl_trusted_certificate: no such file at " ..
-                        conf.lua_ssl_trusted_certificate
+  if conf.lua_ssl_trusted_certificate then
+    for _, path in ipairs(conf.lua_ssl_trusted_certificate) do
+      if not pl_path.exists(path) then
+        errors[#errors + 1] = "lua_ssl_trusted_certificate: no such file at " ..
+                               path
+      end
+    end
   end
 
   if conf.ssl_cipher_suite ~= "custom" then
@@ -1508,9 +1511,13 @@ local function load(path, custom_conf, opts)
     conf.admin_ssl_cert_key = pl_path.abspath(conf.admin_ssl_cert_key)
   end
 
-  if conf.lua_ssl_trusted_certificate then
+  if conf.lua_ssl_trusted_certificate
+     and #conf.lua_ssl_trusted_certificate > 0 then
     conf.lua_ssl_trusted_certificate =
-      pl_path.abspath(conf.lua_ssl_trusted_certificate)
+      tablex.map(pl_path.abspath, conf.lua_ssl_trusted_certificate)
+
+    conf.lua_ssl_trusted_certificate_combined =
+      pl_path.abspath(pl_path.join(conf.prefix, ".ca_combined"))
   end
 
   if conf.cluster_cert and conf.cluster_cert_key then
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -802,12 +802,30 @@ local function check_and_infer(conf, opts)
   end
 
   if conf.lua_ssl_trusted_certificate then
-    for _, path in ipairs(conf.lua_ssl_trusted_certificate) do
+    local new_paths = {}
+
+    for i, path in ipairs(conf.lua_ssl_trusted_certificate) do
+      if path == "system" then
+        local system_path, err = utils.get_system_trusted_certs_filepath()
+        if system_path then
+          path = system_path
+
+        else
+          errors[#errors + 1] =
+            "lua_ssl_trusted_certificate: unable to locate system bundle - " ..
+            err
+        end
+      end
+
       if not pl_path.exists(path) then
         errors[#errors + 1] = "lua_ssl_trusted_certificate: no such file at " ..
                                path
       end
+
+      new_paths[i] = path
     end
+
+    conf.lua_ssl_trusted_certificate = new_paths
   end
 
   if conf.ssl_cipher_suite ~= "custom" then
--- a/kong/tools/utils.lua
+++ b/kong/tools/utils.lua
@@ -12,6 +12,8 @@ local ffi = require "ffi"
 local uuid = require "resty.jit-uuid"
 local pl_stringx = require "pl.stringx"
 local pl_stringio = require "pl.stringio"
+local pl_utils = require "pl.utils"
+local pl_path = require "pl.path"
 local zlib = require "ffi-zlib"
 
 local C             = ffi.C
@@ -109,8 +111,6 @@ function _M.get_hostname()
 end
 
 do
-  local pl_utils = require "pl.utils"
-
   local _system_infos
 
   function _M.get_system_infos()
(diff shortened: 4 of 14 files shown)


kong/tools/utils.lua
```diff
@@ -136,6 +136,42 @@ do
...
 
+
+function _M.get_system_trusted_certs_filepath()
+  local ok, code, stdout, stderr =
+    pl_utils.executeex("openssl version -d")
+  if not ok or code ~= 0 or type(stdout) ~= "string" then
+    return nil, stderr
+  end
+
+  local folder = stdout:match('OPENSSLDIR: "(.+)"')
```
fffonion, line 147: I was loitering and saw this PR 😄 .
 `openssl version -d` returns the hardcoded marco and it may vary during configure, if user links Kong to a custom built openssl this will become unreliable. In fact, the CA bundle are usually packed by the distro and is separate from openssl itself. So I'd suggest to just use a hardcoded full path instead, without considering prefix.
References are https://golang.org/src/crypto/x509/root_linux.go and https://github.com/curl/curl/blob/7a71965e9761f80aa4508472c5b1e3ab03ba018a/acinclude.m4#L2151

kikito: Your "loitering" is way more productive than mine! Thanks for the review, I have made the changes you suggested.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -103,6 +103,33 @@ local function gen_default_ssl_cert(kong_config, target)
   return true
 end
 
+
+local function gen_trusted_certs_file(kong_config)
+  -- create SSL folder
+  local ssl_path = pl_path.join(kong_config.prefix, "ssl")
```
dndx, line 109: I don't think we necessarily have to create this directory. We can save the bundle to `$prefix/.ca_bundle` or something similar, like how we saved `.kong_env`. This way it is even less visible to users.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -103,6 +103,33 @@ local function gen_default_ssl_cert(kong_config, target)
...
+  if not ok then
+    return nil, err
+  end
+
+  local certs_path = pl_path.join(ssl_path, "certs.pem")
+  log.verbose("generating trusted certs file in ", certs_path)
+
+  local fd = assert(io.open(certs_path, "w+b"))
+
+  for _,path in pairs(kong_config.lua_ssl_trusted_certificate) do
```
dndx, line 120:
```suggestion
  for _, path in ipairs(kong_config.lua_ssl_trusted_certificate) do
```

`pairs` are usually used with dictionaries where keys are not integers.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -103,6 +103,33 @@ local function gen_default_ssl_cert(kong_config, target)
...
+  local ssl_path = pl_path.join(kong_config.prefix, "ssl")
+  local ok, err = pl_dir.makepath(ssl_path)
+  if not ok then
+    return nil, err
+  end
+
+  local certs_path = pl_path.join(ssl_path, "certs.pem")
+  log.verbose("generating trusted certs file in ", certs_path)
+
+  local fd = assert(io.open(certs_path, "w+b"))
```
dndx, line 118:
```suggestion
  local fd = assert(io.open(certs_path, "w"))
```

should be enough. Making the file writable by everyone is not necessary in here.

See: https://github.com/Kong/kong/blob/master/kong/db/declarative/init.lua#L272 for another example.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -103,6 +103,33 @@ local function gen_default_ssl_cert(kong_config, target)
...
+
+  for _,path in pairs(kong_config.lua_ssl_trusted_certificate) do
+    local f = assert(io.open(path, "r"))
+    local str = assert(f:read("*a"))
+    f:close()
```
dndx, line 123: https://stevedonovan.github.io/Penlight/api/libraries/pl.file.html#read

might save a few lines of code here :)

kong/cmd/utils/prefix_handler.lua
```diff
@@ -168,6 +195,19 @@ local function compile_conf(kong_config, conf_template)
     end
   end
 
+  if type(kong_config.lua_ssl_trusted_certificate) == "table" then
```
dndx, line 198: Since `lua_ssl_trusted_certificate = { typ = "array" },`, wouldn't this be true always?

kong/cmd/utils/prefix_handler.lua
```diff
@@ -168,6 +195,19 @@ local function compile_conf(kong_config, conf_template)
...
+    if #kong_config.lua_ssl_trusted_certificate == 0 then
+      kong_config.lua_ssl_trusted_certificate = nil
+
+    else
+      local filepath, err = gen_trusted_certs_file(kong_config)
+      if not filepath then
+        return nil, err
+      end
+      kong_config.lua_ssl_trusted_certificate = filepath
```
dndx, line 207: Usually when we do config transformations, a new config is generated instead of modifying the existing one. 

For example:

https://github.com/Kong/kong/blob/next/kong/conf_loader/init.lua#L813-L817

I think assigning the filename to a new entry such as `kong_config.lua_ssl_trusted_certificate_combined` follows the convention a bit better.

kong/conf_loader/init.lua
```diff
@@ -800,11 +801,26 @@ local function check_and_infer(conf, opts)
     end
   end
 
-  if conf.lua_ssl_trusted_certificate and
-     not pl_path.exists(conf.lua_ssl_trusted_certificate)
-  then
-    errors[#errors + 1] = "lua_ssl_trusted_certificate: no such file at " ..
-                        conf.lua_ssl_trusted_certificate
+  if conf.lua_ssl_trusted_certificate then
+    local parsed_paths = {}
```
dndx, line 805:
```suggestion
    local parsed_paths = {}

```

kong/conf_loader/init.lua
```diff
@@ -800,11 +801,26 @@ local function check_and_infer(conf, opts)
...
+    for i, path in ipairs(conf.lua_ssl_trusted_certificate) do
+      if path == "auto" then
+        local new_path, err = utils.get_system_trusted_certs_filepath()
+        if new_path then
+          path = new_path
+        else
+          errors[#errors + 1] =
+            "error while calculating lua_ssl_trusted_certificate `auto` path: " ..
```
dndx, line 813:
```suggestion
            "unable to locate system bundle for lua_ssl_trusted_certificate \"auto\" option: " ..
```

kong/conf_loader/init.lua
```diff
@@ -800,11 +801,26 @@ local function check_and_infer(conf, opts)
...
-    errors[#errors + 1] = "lua_ssl_trusted_certificate: no such file at " ..
-                        conf.lua_ssl_trusted_certificate
+  if conf.lua_ssl_trusted_certificate then
+    local parsed_paths = {}
+    for i, path in ipairs(conf.lua_ssl_trusted_certificate) do
+      if path == "auto" then
+        local new_path, err = utils.get_system_trusted_certs_filepath()
+        if new_path then
+          path = new_path
+        else
```
dndx, line 811:
```suggestion

        else
```

kong/conf_loader/init.lua
```diff
@@ -800,11 +801,26 @@ local function check_and_infer(conf, opts)
...
+          errors[#errors + 1] =
+            "error while calculating lua_ssl_trusted_certificate `auto` path: " ..
+            err
+        end
+      end
```
dndx, line 816:
```suggestion
      end

```

kong/conf_loader/init.lua
```diff
@@ -800,11 +801,26 @@ local function check_and_infer(conf, opts)
...
+      if not pl_path.exists(path) then
+        errors[#errors + 1] = "lua_ssl_trusted_certificate: no such file at " ..
+                               path
+      end
```
dndx, line 820:
```suggestion
      end

```

kong/conf_loader/init.lua
```diff
@@ -800,11 +801,26 @@ local function check_and_infer(conf, opts)
...
+      parsed_paths[i] = path
+    end
```
dndx, line 822:
```suggestion
    end

```

kong/conf_loader/init.lua
```diff
@@ -986,6 +1002,9 @@ local function overrides(k, default_v, opts, file_conf, arg_conf)
     value = default_v == "NONE" and "" or default_v
 
   else
+    if not file_conf then
```
dndx, line 1005: Hmm, I might be missing something, but what is this addition for?

kong/conf_loader/init.lua
```diff
@@ -1509,8 +1528,11 @@ local function load(path, custom_conf, opts)
   end
 
   if conf.lua_ssl_trusted_certificate then
-    conf.lua_ssl_trusted_certificate =
-      pl_path.abspath(conf.lua_ssl_trusted_certificate)
+    local abs_paths = {}
+    for i, path in ipairs(conf.lua_ssl_trusted_certificate) do
+      abs_paths[i] = pl_path.abspath(path)
+    end
+    conf.lua_ssl_trusted_certificate = abs_paths
```
dndx, line 1535:
```suggestion
    conf.lua_ssl_trusted_certificate = pl_tablex.map(pl_path.abspath, conf.lua_ssl_trusted_certificate)
```

should be equivalent, handwritten not tested though.

dndx (changes requested):

kong/conf_loader/init.lua
kikito: Leftover from debugging. Thanks!

kong/cmd/utils/prefix_handler.lua
kikito: Ok, changed path to `$prefix/.ca_bundle`

kong/cmd/utils/prefix_handler.lua
kikito: It can be `nil` on tests

kong.conf.default
```diff
@@ -1213,11 +1213,32 @@
...
 
-#lua_ssl_trusted_certificate =   # Absolute path to the certificate
-                                 # authority file for Lua cosockets in PEM
-                                 # format. This certificate will be the one
-                                 # used for verifying Kong's database
-                                 # connections, when `pg_ssl_verify` or
+#lua_ssl_trusted_certificate =   # Comma-separated list of paths to certificate
+                                 # authority files for Lua cosockets in PEM format.
+                                 #
+                                 # The special value `auto` expands to the
```
dndx, line 1219: Just a thought: would `system` be a better name here? I think it may be clearer than `auto` in terms of what the value actually expands to.

kong/cmd/utils/prefix_handler.lua
kikito: I considered this. Each option has some advantages and some disadvantages.

* Using two different config options, like you are proposing, has the problem that both options ultimately affect the same nginx variable, that is, the nginx `lua_ssl_trusted_certificate` option. I thought this would be ultimately less friendly to the user, because "what happens if I set both options" needs to be answered. Either we "ignore one silently", we "combine both" or we send an error message.
* Extending the existing option means less hussle for the end user. There is some impedance too (what was a 1-to-1 mapping to the nginx config option is now in fact a full-blown Kong option with "magic"). But for all effects and purposes it is backwards-compatible, so users who already used it don't need to change anything, and new users can read the description.

The second option is also slightly more difficult to implement than the first one because it requires "convincing" the config validator into transforming an array into a string, or nil.

> We choose to go to the Moon in this decade and do the other things, not because they are easy, but because they are hard

That said, going around the config validator like this is a bit of a pain and I might just do what you suggested, just to avoid having to do it 😋

kong/cmd/utils/prefix_handler.lua
```diff
@@ -168,6 +184,16 @@ local function compile_conf(kong_config, conf_template)
     end
   end
 
+  if type(kong_config.lua_ssl_trusted_certificate) == "table" then
+    if #kong_config.lua_ssl_trusted_certificate == 0 then
+      kong_config.lua_ssl_trusted_certificate = nil
+
+    elseif #kong_config.lua_ssl_trusted_certificate == 1 then
+      kong_config.lua_ssl_trusted_certificate = kong_config.lua_ssl_trusted_certificate[1]
```
kikito, line 192: Added this for full backards-compatibility: if only one item is used, don't even create the "hidden" file.

kong/cmd/utils/prefix_handler.lua
kikito: Update: convincing the conf loader to do this was too difficult. I finally went with your proposed solution (using a separate conf variable).

kikito:
@dndx thanks for the review. I have acted on everything I could, and accepted most of your proposed changes. One of them however involved redoing a big chunk (splitting functionality into two vars). I resisted but ended up going that way.

kong/tools/utils.lua
```diff
@@ -136,6 +136,33 @@ do
...
+
+  function _M.get_system_trusted_certs_filepath()
+    for _, path in ipairs(trusted_certs_paths) do
+      if pl_path.exists(path) then
+        return path
+      end
+    end
+
+    return nil,
+           "Could not find trusted certs file in " ..
```
fffonion, line 157: maye need to adjust this error from `auto` with `system`

kong/cmd/utils/prefix_handler.lua
```diff
@@ -103,6 +103,24 @@ local function gen_default_ssl_cert(kong_config, target)
...
+  combined_filepath,
+  lua_ssl_trusted_certificate_combined)
+
+  log.verbose("generating trusted certs combined file in ",
+              combined_filepath)
+
+  local fd = assert(io.open(combined_filepath, "w"))
+
+  for _, path in ipairs(lua_ssl_trusted_certificate_combined) do
+    fd:write(pl_file.read(path))
```
fffonion, line 117: Since we are appending the certificate, i'd suggest to add a `\n` for each file, in case some one incorrectly specifies a file without newline.

fffonion (approved):
Left some minor comments, otherwise LGTM

kong.conf.default
```diff
@@ -1222,6 +1222,30 @@
                                  #
                                  # See https://github.com/openresty/lua-nginx-module#lua_ssl_trusted_certificate
 
+#lua_ssl_trusted_certificate_combined =
```
dndx, line 1225: Is there a possibility we can reuse the existing `lua_ssl_trusted_certificate` config for this? It seems that the config format should be compatible with the existing `lua_ssl_trusted_certificate` format and that way we avoid any confusions from the user.

The idea is that `lua_ssl_trusted_certificate_combined` is generated and used by Kong internally, but it need not to be visible to the users.

kikito: Yes, indeed! I didn't realize that we could rely on "hidden" (non-user-visible) config options.

This change involved modifying all the kong templates though. We'll have to document that.

In any case, I've re-done the PR using a hidden config option. Looks much better now, thanks!

kikito:
@dndx @fffonion All changes done, ready for another pass!

kong.conf.default
```diff
@@ -1213,12 +1213,34 @@
...
+#lua_ssl_trusted_certificate =   # Comma-separated list of paths to certificate
+                                 # authority files for Lua cosockets in PEM format.
+                                 #
+                                 # These certificates will be the combined
+                                 # into a single file during the startup phase.
+                                 #
+                                 # The special value `system` expands to the
+                                 # "usual default" provided by each distro. The following
+                                 # paths will be tested in order, and the first one found
+                                 # will be used:
```
hishamhm, line 1226:
```suggestion
                                 # The special value `system` expands to the
                                 # "usual default" provided by each distro, according
                                 # to an arbitrary heuristic. In the current implementation,
                                 # The following pathnames will be tested in order,
                                 # and the first one found will be used:
```

Sorry about the bit of "language lawyering", just trying to keep the door open so we can adjust the automagic behavior if we deem necessary in the future without incurring in a breaking change. :)

kong.conf.default
```diff
@@ -1213,12 +1213,35 @@
...
-                                 # `cassandra_ssl_verify` are enabled.
+
+#lua_ssl_trusted_certificate =   # Comma-separated list of paths to certificate
+                                 # authority files for Lua cosockets in PEM format.
+                                 #
+                                 # These certificates will be the combined
+                                 # into a single file during the startup phase.
+                                 #
+                                 # The special value `system` expands to the
+                                 # "usual default" provided by each distro, according
```
dndx, line 1224:
```suggestion
                                 # The special value `system` attempts to search for the
                                 # "usual default" provided by each distro, according
```

seems to be better. "expands to the "usual default" provided by each distro" make it sounds like we insert all of these paths into the config, which may cause users to try to do the same by writing:

`lua_ssl_trusted_certificate = /path/to/a.pem, /path/to/b.pem` and expects us to ignore non existent path (which is not the behavior here).

dndx (commented):
Minor comments, everything else LGTM :) Thanks @kikito for the awesome work!

dndx (approved):

PR #6344, opened 2020-09-16
fix(plugins) stream compatibility in logging plugins

javierguerragiraldez:


kong/plugins/loggly/handler.lua
```diff
@@ -99,12 +99,14 @@ local function log(premature, conf, message)
     return
   end
 
-  if message.response.status >= 500 then
-    return decide_severity(conf, conf.server_errors_severity, message)
-  end
-
-  if message.response.status >= 400 then
-    return decide_severity(conf, conf.client_errors_severity, message)
+  if ngx.config.subsystem == "http" then
```
dndx, line 102: This can be cached at module level, it does not change for the Lua VM code was loaded into.

dndx (commented):
LGTM, minor comments.

kong/plugins/loggly/handler.lua
javierguerragiraldez: thanks!

dndx (approved):

PR #6348, opened 2020-09-16
feat(upstream) When `Service.client_certificate` is not set, use

dndx:
`Upstream.client_certificate` instead.

This allows `client_certificate` setting used for mTLS handshaking with
the `Upstream` server to be shared easily among different Services.
However, `Service.client_certificate` will take precedence over
`Upstream.client_certificate` if both are set simultaneously.

### DIFF
--- a/kong/runloop/balancer.lua
+++ b/kong/runloop/balancer.lua
@@ -1093,6 +1093,28 @@ local function execute(target, ctx)
         hash_value = create_hash(upstream, ctx)
         target.hash_value = hash_value
       end
+
+      if not ctx.service.client_certificate then
+        -- service level client_certificate is not set
+        local cert, res, err
+        local client_certificate = upstream.client_certificate
+
+        -- does the upstream object contains a client certificate?
+        if client_certificate then
+          cert, err = get_certificate(client_certificate)
+          if not cert then
+            log(ERR, "unable to fetch upstream client TLS certificate ",
+                     client_certificate.id, ": ", err)
+            return
+          end
+
+          res, err = kong.service.set_tls_cert_key(cert.cert, cert.key)
+          if not res then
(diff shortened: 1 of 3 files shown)


kikito (approved):
Approved, assuming that the red tests are due to unrelated issues.

kong/runloop/balancer.lua
```diff
@@ -1093,6 +1093,28 @@ local function execute(target, ctx)
...
       end
+
+      if not ctx.service.client_certificate then
+        -- service level client_certificate is not set
+        local cert, res, err
+        local client_certificate = upstream.client_certificate
+
+        -- does the upstream object contains a client certificate?
+        if client_certificate then
+          cert, err = get_certificate(client_certificate)
```
bungle, line 1104: This is possibly `yielding` on `non-yielding` context.

bungle (dismissed):
I like to get confirmation that it is not yielding on `balancer phase`.

fffonion (approved):

dndx:
d83b9caf9b85340a0a2d8b35d81bbaa1b5999b5a should have fixed the test failure. Turns out it *was* related to this PR, in an unexpected way...

```
2020/09/18 05:17:35 [error] 26553#0: *3 [lua] clustering.lua:181: unable to update running config: bad config received from control plane in 'upstreams':
  in 'backend-mtls':
    - in entry 1 of 'backend-mtls': invalid reference 'client_certificate: 9802a40e-d052-4867-b8a9-888532407326' (no such entry in 'certificates'), context: ngx.timer
```

was the message for the error. I think the `Certificates` are emptied between these new tests and the Hybrid mode test and the reference is broken. The fix is to properly clear these tables before running the Hybrid mode tests to avoid dangling entries.

PR #6352, opened 2020-09-16
fix(schema) infer shorthands

### LINKED ISSUE #6272: After upgrading from 2.0.5 to 2.1.3 the deprecation system of “whitelist” -> “allow” does not work
After upgrading from 2.0.5 to 2.1.3 I’m seeing this behaviour when testing the deprecation system of “whitelist” -> “allow”:

In 2.0.5 this works fine (`-f` means `application/x-www-form-urlencoded`):

```
[kong-2.0.5:/kong]# http :8001/services/ServiceA/plugins name=ip-restriction config.whitelist=127.0.0.1 -f
HTTP/1.1 201 Created
...
...
```

After I upgraded to 2.1.3 I tried using the same HTTPie command:

```
[kong-2.1.3:/kong]# http :8001/services/ServiceB/plugins name=ip-restriction config.whitelist=127.0.0.1 -f
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 149
Date: Fri, 21 Aug 2020 22:00:30 GMT
Server: kong/2.1.3
X-Kong-Admin-Latency: 3

{
    "code": 2,
    "fields": {
        "config": {
            "allow": "expected an array"
        }
    },
    "message": "schema violation (config.allow: expected an array)",
    "name": "schema violation"
}
```

With `config.allow` instead of `config.whitelist` it works, though:

```
[kong-2.1.3:/kong]# http :8001/services/ServiceB/plugins name=ip-restriction config.allow=127.0.0.1 -f
HTTP/1.1 201 Created
...
...
```

bungle:
### Summary

In issue #6272 it was reported that some of our new shorthands that were added to `2.1.x` (whitelist/blacklist -> allow/deny) do not behave correctly with `application/x-www-form-urlencoded` or `multipart/form-data`.

I considered several options to fix these, including:

1. Fix each shorthands function to change the types (to arrays in this case):
   felt like just repetitive code, and a bug waiting to happen again
2. Change arguments parser to call shorthand functions and infer results:
   would fix it only for code that uses arguments parser
3. Infer shorthands return values according to schema field:
   Does always the inferring for shorthands return values if there is a matching field
4. Change plugins api to do inferring there only to config table values:
   is not universal solution, and makes plugins config behave differently than rest

 Out of those I went with 3.

 While inferring values in general should happen only for  `application/x-www-form-urlencoded` and `multipart/form-data`, I don't think  inferring the shorthand function return values is all that bad. It is a little bit magical, but I could not think it being problematic.

 ### Issues Resolved

 Fix #6272

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1,5 +1,6 @@
 local tablex       = require "pl.tablex"
 local pretty       = require "pl.pretty"
+local arguments    = require "kong.api.arguments"
 local utils        = require "kong.tools.utils"
 local cjson        = require "cjson"
 
@@ -1551,7 +1552,12 @@ function Schema:process_auto_fields(data, context, nulls, opts)
         local new_values = sfunc(value)
         if new_values then
           for k, v in pairs(new_values) do
-            data[k] = v
+            local f = self.fields[k]
+            if f then
+              data[k] = arguments.infer_value(v, f)
+            else
+              data[k] = v
+            end
           end
         end
       end
--- a/spec/01-unit/01-db/01-schema/01-schema_spec.lua
+++ b/spec/01-unit/01-db/01-schema/01-schema_spec.lua
@@ -3347,6 +3347,44 @@ describe("schema", function()
       assert.falsy(check_immutable_fields)
     end)
 
+    it("infers shorthands", function()
+      local test_schema = {
+        name = "test",
+        fields = {
+          {
+            allow = {
+              type = "array",
+              elements = {
+                type = "string"
+              },
+            },
+          },
+        },
+        shorthands = {
+          {
+            whitelist = function(value)
+              return {
+                allow = value
+              }
+            end
+          },
+        },
+      }
+
+      local test_entity = { name = "bob", whitelist = "privileged" }
+
+      local TestEntities = Schema.new(test_schema)
+
+      local test_data = TestEntities:process_auto_fields(test_entity, "update")
+
+      assert.same({
+        name = "bob",
+        allow = {
+          "privileged",
+        },
+      }, test_data)
+    end)
+
     describe("in subschemas", function()
       it("a specialized field can set a default", function()
         local Test = Schema.new({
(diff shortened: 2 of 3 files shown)


kong/meta.lua
```diff
@@ -20,6 +20,6 @@ return {
   -- third-party dependencies' required version, as they would be specified
   -- to lua-version's `set()` in the form {from, to}
   _DEPENDENCIES = {
-    nginx = { "1.15.8.1", "1.15.8.3" },
```
kikito, line 23: I think this change does not belong on this PR

kikito (changes requested):

hishamhm:
> While inferring values in general should happen only for application/x-www-form-urlencoded and multipart/form-data, I don't think inferring the shorthand function return values is all that bad. It is a little bit magical, but I could not think it being problematic.

I am concerned about this. It loosens the type checking for shorthands all over the place (declarative config loading, db_import, etc.) and might bite us back in the future in an unrelated area because of something that was done only because of `application/x-www-form-urlencoded` or `multipart/form-data` in the Admin API. I think the effects of inference should be restricted to that. The option that would achieve this is option 2, right?

bungle:
> I am concerned about this. It loosens the type checking for shorthands all over the place (declarative config loading, db_import, etc.) and might bite us back in the future in an unrelated area because of something that was done only because of `application/x-www-form-urlencoded` or `multipart/form-data` in the Admin API. I think the effects of inference should be restricted to that. The option that would achieve this is option 2, right?

Yes, it loosens the type checking as shorthands by definition are loosened. There is no field definition for shorthands.
Shorthand is a quickie for devs, and some automation would probably be nice.

Option 2 is also a bit tricky. It copies same code to other place, and the order might then be different in admin api and that code is executed twice (it could affect error messages too), and in general dao api. It might as well bite us. It adds shorthands to admin api separately. Shorthands doesn't really play nice in different scenarios, its design feels off.

hishamhm:
Very fair points! Let me give this a little more thought and see if I can come up with any proposal.

bungle:
This PR is superseded by #6364, thus closing it.

PR #6354, opened 2020-09-16
fix(acl) deny should not deny proxying if denied group does not match

### LINKED ISSUE #6281: Unexpected behavior of the ACL plugin with kong 2.1.3 when using deny property
### Summary

I used to run the version 1.1.1 of Kong with the ACL plugin as a global plugin. I use this plugin in order to populate the X-Consumer-Groups header which is used by my upstream services. I didn't really want to exclude any groups but just have the header populated so I was blacklisting a none existing group.

This was working perfectly fine until I upgraded kong the 2.1.3 version. From now on with the same configuration for the ACL plugin, I am always getting 401 Error.

Is that the expected behavior in this new version ?

### Steps To Reproduce

1- Use kong 2.1.3
2- Activate globally the ACL plugin
3- Add a random and not existing group
4- Try to access any route

bungle:
### Summary

It was reported with #6281 that ACL plugin with `config.deny=do-not-deny-anyone`
does not work in all scenarios. I found two possible scenarios:

1. The user was authenticated but did not have consumer nor did not set authenticated groups
2. The user was not authenticated at all (no auth plugins added)
3. There was authentication but the authenticated user was anonymous and thus didn't have credential

All of this lead to `401`s.

This is more about semantics. Loosening the verification on anonymous and non-authenticated
users will make this plugin a bit more flexible, but then I can see that this might open some
security issues.

What does `nil` `authenticated_groups` mean (or empty `{}`) when using ACL plugin with `config.deny`?
1. the groups cannot be determined, deny the request
2. we didn't match any deny groups

I think `2.` is more correct and expected. The plugin is just verifying `deny` OR `allow` groups
and that's it. DENY: empty groups = `nothing to deny`, ALLOW: empty groups = `do not ever allow`.

As this changes the behavior a bit, I rebased this to `next`.

### Issues Resolved

Fix #6281

### DIFF
--- a/kong/plugins/acl/handler.lua
+++ b/kong/plugins/acl/handler.lua
@@ -40,7 +40,7 @@ local ACLHandler = {}
 
 
 ACLHandler.PRIORITY = 950
-ACLHandler.VERSION = "3.0.0"
+ACLHandler.VERSION = "3.0.1"
 
 
 function ACLHandler:access(conf)
@@ -99,7 +99,7 @@ function ACLHandler:access(conf)
       end
 
       if not consumer_groups then
-        if config.type == DENY and credential then
+        if config.type == DENY then
           consumer_groups = EMPTY
 
         else
--- a/spec/03-plugins/18-acl/02-access_spec.lua
+++ b/spec/03-plugins/18-acl/02-access_spec.lua
@@ -246,6 +246,18 @@ for _, strategy in helpers.each_strategy() do
         }
       }
 
+      local route3d = bp.routes:insert {
+        hosts = { "acl3d.com" },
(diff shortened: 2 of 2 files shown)


hishamhm:
@bungle Based on our out-of-band discussion, if we instead revert to the 1.x/2.0-like behavior (401 on "denied and with no authenticated groups" test, 200 on "deny groups" test, 401 on "allow groups" test), we can merge this on `master`.

hishamhm (approved):

hishamhm:
(Travis failure is safe to ignore because it is messing up with the change of base branch.)

PR #6357, opened 2020-09-16
fix(key-auth) do not error with 400 when body cannot be decoded

### LINKED ISSUE #5893: Key Auth Plugin throws 400 Bad Request when requests doesn't have content-type using key_in_body

bungle:
### Summary

Issue #5893 pointed out that `key-auth` plugin returns `400` and "Cannot process request body" message in case the body decoding fails. This happens even in case the key is found in other places like `query` of `headers`.

### Issues Resolved

Fix #5893

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -8,8 +8,11 @@ local error = error
 
 local KeyAuthHandler = {
   PRIORITY = 1003,
-  VERSION = "2.2.0",
+  VERSION = "2.3.0",
 }
+
+
+local EMPTY = {}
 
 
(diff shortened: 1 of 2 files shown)


hishamhm:
@bungle Since this looks like a fix, should it instead go to `master` and bump the version to `2.2.1` instead?

bungle:
@hishamhm hard to say, as it changes behavior a bit (different status code, doesn't try to read body until neccessary, etc.), even for better, it has been for long like this. This is not that important that we need to backport it. Thus I thought that perhaps `next` is good enough. But I am fine with making it into `master`

kikito (approved):

PR #6358, opened 2020-09-16
feat(correlation-id) override empty correlation id with a generated one

### LINKED ISSUE #5646: Correlation ID plugin does not check for NULL value
### Summary

Correlation ID plugin currently only checks if the header is present or not, if not present generates the ID & add as header before sending the request to upstream. Although it does not check if the value is NULL or Empty. Will it be possible to update the plugin to check for the same & generate ID if its NULL or Empty.

### Steps To Reproduce

1. Prepare curl request for upstream API configured with correlation-id plugin having correlation-id header set to NULL.
2. Send the request to kong.
3. Upstream receives the request & rejects it as correlation-id is NULL or Empty

### Additional Details & Logs

- Kong version - 2.0.1

bungle:
### Summary

Makes `correlation-id` plugin to check for empty header. Before this commit the empty header was forwarded as is. With this commit the empty header is replaced by a generated id. I didn't feel we need to make this a config option.

### Issues Resolved

Fix #5646

### DIFF
--- a/kong/plugins/correlation-id/handler.lua
+++ b/kong/plugins/correlation-id/handler.lua
@@ -13,6 +13,7 @@ local generators
 do
   local worker_pid = ngx.worker.pid()
   local now = ngx.now
+  local var = ngx.var
   local fmt = string.format
 
   generators = {
@@ -24,8 +25,6 @@ do
       return worker_uuid .. "#" .. worker_counter
     end,
     ["tracker"] = function()
-      local var = ngx.var
-
       return fmt("%s-%s-%d-%s-%s-%0.3f",
         var.server_addr,
         var.server_port,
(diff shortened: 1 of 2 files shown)


kong/plugins/correlation-id/handler.lua
```diff
@@ -76,8 +75,11 @@ function CorrelationIdHandler:header_filter(conf)
   end
 
   local correlation_id = kong.ctx.plugin.correlation_id or
-                         kong.request.get_header(conf.header_name) or
```
dndx, line 79: Maybe we still need to keep this line, because if `kong.request.get_header(conf.header_name)` returned something in `access` phase, then the https://github.com/Kong/kong/pull/6358/files#diff-8e9c59879919fc95ad333be729b1ba4fR57 condition is not met and `kong.ctx.plugin.correlation_id` will therefore not be set.

dndx (approved):

PR #6364, opened 2020-09-17
feat(schema) typed shorthand_fields, deprecate shorthands

hishamhm:
This is a "second shot" at the design of the shorthands feature, based on comments by @bungle, who correctly pointed out that we were missing type information.

This also fixes #6352 by making the form-encoded inference system in the Admin API aware of shorthand_fields.

The feature has been renamed from `shorthands` to `shorthand_fields` to avoid a "polymorphic" state in between the deprecated and new form.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1154,7 +1154,8 @@ local function run_entity_check(self, name, input, arg, full_check, errors)
     local value = get_field(input, fname)
     if value == nil then
       if (not checker.run_with_missing_fields) and
-         (required_fields and required_fields[fname]) then
+         (required_fields and required_fields[fname]) and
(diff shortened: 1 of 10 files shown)


kong/db/schema/init.lua
```diff
@@ -1542,6 +1543,32 @@ function Schema:process_auto_fields(data, context, nulls, opts)
...
+          end
+        end
+      end
+    end
+    if next(errs) then
+      return nil, errs
+    end
+  end
+
+  -- deprecated
```
murillopaula, line 1571: to be removed in Kong 3.0 too? if yes, how do we make sure that we don't forget it?

murillopaula (approved):
lgtm, tested locally with the ip-restriction plugin following the same steps as https://github.com/Kong/kong/issues/6272

PR #6372, opened 2020-09-18
chore(rockspec) bump request-transformer to 1.3.0

hishamhm:
New in this release:

* Include the type function in template environment. Because headers may
  contain array elements such as duplicated headers, type is a useful function
  in these cases.


hishamhm:
@dndx Any idea of what's failing here?
https://travis-ci.org/github/Kong/kong/jobs/728390035#L2704

PR #6374, opened 2020-09-18
Add a minimal http API endpoint for stream plugins.

javierguerragiraldez:
OpenResty http and stream modules keep Lua code in separate spaces; and for the time being, the shared dictionaries are shared only between the spaces of the same module.

This feature adds a new TCP listen port (or unix socket) with a very simple subset of HTTP/1.0 that is just enough to handle requests by the `lua-resty-http` library.

### `stream_api` setting (or `KONG_STREAM_API` environment variable)
If set, defines the listening socket.  Can be a TCP host:port like `"127.0.0.1:8008"` or a unix socket path (absolute or relative to the `prefix` setting).

### Endpoint handler

Add a `api.lua` file to your plugin, it should return a table with a `_stream` field containing a table with an entry for each path and the corresponding handler function.  For example:

```lua
return {
   _stream = {
      ["/open_sockets"] = function (req)
         ...
      end,
   },
}
```

The `req` object has several methods to help handling simple HTTP/1.0 requests:

### `req:get_line()`
Returns the initial request line.

### `req:get_method()`
Return the "method" part of the request line, actually it's just the first non-space fragment of the line.  Not used by the module in any way.

### `req:get_path()`
Returns the "path" part of the request line, actually it's just the second non-space fragment of the line.  This is the only part that determines which handler was called.

### `req:get_headers()`
Returns a Lua table with the request headers.  Keys are all lowercase and replaces '-' with '_'.  If two or more headers have the same key, they're concatenated with a space between them.  There's no multiline header handling.

### `req:get_body([len])`
Returns the request body.  The optional `len` defines the amount of bytes to be read from the socket.  If omitted, it looks for a `content_length` header (calling `req:get_headers()`), if not found, reads until the end of the connection.
Note that this function (like the other `req:get_XXX()`) is memoized, so can't be used to read content piece by piece.

### `req:response(status [, headers [, body]])`
Constructs an HTTP/1.0 response as a Lua string.  `status` can be a number (like `200`) or string (like `"404 Not Found"`).  `headers` is a Lua table; the keys are not canonicalized before serializing.

### `req:exit([payload])`
Sends the `payload` (if any) on the socket, closes it and terminates the response (the function doesn't return).  The handler doesn't have to call it, it's called automatically with its final result so it can just end with a `return req:response(200)` or similar.

## Handling non HTTP protocols:

It should be noted that at the start of the `handler` function, nothing more than the "request line" has been read from the connection.  All `req:get_XXX()` methods are memoized, so they can be called any number of times and will read from the connection only once.  The order is weakly imposed: `req:get_path()` calls `req:get_line()`, `req:get_body()` might call `req:get_headers()`, `req:response()` doesn't call any of them, and doesn't touch the connection.

This allows for different, non-http based protocols to be implemented.  The only assumption is that there's a "request line" where the second non-space word is used to determine the handler to call.

For example, instead of using `lua-resty-http`, a "client" could just send:

```
PING /ping-handler
{ ... some data ... }
<break connection>
```

and the handler could be:

```lua
return {
  _stream = {
    ["/ping-handler"] = function(req)
      data = cjson.decode(req:get_body("*a"))

      return "{ ... some response data ...}"
    end,
  }
}
```

For larger, more complex uses (like reading by pieces, or keeping the connection open...) you can use `req:socket` directly.


kong/tools/stream_api.lua
```diff
@@ -0,0 +1,111 @@
...
+end
+
+function stream_mt:get_headers()
+  if not self.headers then
+    local headers = {}
+
+    while true do
+      local l = self.socket:receive("*l")
+      if l == "" then
+        self.headers = headers
```
kikito, line 52: This line seems unnecessary, as we're doing `self.headers = headers` after the loop.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,111 @@
...
+function stream_mt:response(status, headers, body)
+  local o = {"HTTP/1.1 " .. tostring(status)}
+
+  if headers then
+    if type(body) == "string" then
+      headers["Content-Length"] = #body
+    end
+
+    for k, v in pairs(headers or {}) do
+      o[#o + 1] = string.format("%s: %s", k, v)
```
kikito, line 92: This will format the headers as `content_type: 25`. While they are [case insensitive](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), I am not sure that `-` can become `_`.

I think Uppercase-With-Slashes is the "de-facto standard". I suggest a reformatting it them like so before returning them.

kikito (changes requested):
My main concern with this PR is that since me main intent behind this PR is developing plugins, it would fit better as a PDK namespace instead of as something you `require` from a plugin. Is there any reason for not adding this to the PDK?

kong/tools/stream_api.lua
javierguerragiraldez: the transformation happens on input (request) to make user code simpler.  on output (response) it respects the user's format.

javierguerragiraldez:
> Is there any reason for not adding this to the PDK?

i tried to make the plugin loader do the registration, similar to the admin api extensions, but got into several complications.  can try again to see how close can it go without overhauling much existing framework.

kong/tools/stream_api.lua
javierguerragiraldez: done

kong/init.lua
```diff
@@ -1375,6 +1376,26 @@ function Kong.serve_cluster_listener(options)
...
 
+local stream_plugins_api_loaded = false
+function Kong.stream_api()
+  if not stream_plugins_api_loaded then
+    local utils = require "kong.tools.utils"
+
+    for k in pairs(kong.configuration.loaded_plugins) do
+      local loaded, custom_endpoints = utils.load_module_if_exists("kong.plugins." .. k .. ".api")
+      if loaded and type(custom_endpoints._stream) == "table" then
+        ngx.log(ngx.DEBUG, "---------------------------- have stream --------------")
```
dndx, line 1387: Maybe include the plugin's name in the debug log as well and change the format to something that is similar to others (e.g. without the dashes).

kong/templates/kong_defaults.lua
```diff
@@ -15,8 +15,9 @@ host_ports = NONE
 anonymous_reports = on
 
 proxy_listen = 0.0.0.0:8000 reuseport backlog=16384, 0.0.0.0:8443 http2 ssl reuseport backlog=16384
-stream_listen = off
 admin_listen = 127.0.0.1:8001 reuseport backlog=16384, 127.0.0.1:8444 http2 ssl reuseport backlog=16384
+stream_listen = off
+stream_api = NONE
```
dndx, line 20: Do we have to make this configurable? To me it seems that this HTTP layer in stream is only for Kong internal use, which in that case a unix domain socket under the Kong prefix using a well known name should be enough.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,114 @@
...
+
+function stream_mt:get_path()
+  if not self.path then
+    self.path = self:get_line():match("^%S+%s+(%S+)")
+  end
+
+  return self.path
+end
+
+function stream_mt:get_headers()
```
dndx, line 49: Open question: I see the framework supports URI, headers and body at the same time. Do we need to support all these? Because it is for internal use, we can be a bit more opinionated and not support certain things that are not absolutely needed or are bit slower to process (like headers). It makes the addition of code less, and ensures everyone uses this system in a similar manner. (I see headers are not really tested in the cases below either)

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,114 @@
...
+      self.content_length = tonumber(self:get_headers().content_length)
+    end
+    self.body = self.socket:receive(self.content_length or "*a")
+  end
+
+  return self.body
+end
+
+function stream_mt:response(status, headers, body)
+  local o = {"HTTP/1.1 " .. tostring(status)}
```
dndx, line 87: Should we generate `Connection: close` just to ensure the connection won't be keepalived and causing issues? Or maybe change to `HTTP/1.0` which does not support keep-alive.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,114 @@
...
+
+  return table.concat(o, "\r\n")
+end
+
+function stream_mt:exit(payload)
+  if payload then
+    self.socket:send(payload)
+  end
+
+  self.socket:shutdown("send")
```
dndx, line 110: I think the simulated "lingering close" as in https://github.com/openresty/stream-lua-nginx-module#tcpsockshutdown is more appropriate in here. Because if we only shutdown the send end without draining the read end, the connection may be reset when `exit` finishes causing ungraceful shutdowns.

Also the `exit` function should probably be calling `return ngx.exit(ngx.OK)` to make sure control flow never returns to the caller.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,114 @@
+
+local stream_api = {}
+
+local _endpoints = {}
+
+
+function stream_api.register(t)
+  for path, handler in pairs(t) do
```
dndx, line 8: We should consider caching these Lua built-in functions at module level.

dndx (commented):
Good job @javierguerragiraldez, I left some comments but I believe we are going down the right path!

kong/init.lua
javierguerragiraldez: done!  (debugging leftovers....)

kong/templates/kong_defaults.lua
javierguerragiraldez: definitely unix domain socket is much better (and changed the tests to be a better example).
I'd also like to avoid a new config variable, but not sure if it's worth it to have an unconditional `listen`.  if was an internal not-really-a-socket, it wouldn't be an issue, but i don't think that exists for streams.

kong/tools/stream_api.lua
javierguerragiraldez: done.  (this one happens only at initialization, but there are others that are part of the request path, so better to do all of them)

kong/tools/stream_api.lua
javierguerragiraldez: to avoid writing also the client side, i made it to be easy to use with resty.http; so it can't deviate toooooo much from HTTP.   in particular, that means reading headers.  

maybe we could ignore any `Content-Length` and just assume "close at the end"? that would allow not parsing into a table in simple cases.

kong/tools/stream_api.lua
javierguerragiraldez: I wasn't sure resty.http supported a 1.0 response for a 1.1 request, so i tried first with 1.1

reading the source, i see it's supported, and if there's no `Content-Length` it reads until the end, which is convenient.

kong/tools/stream_api.lua
javierguerragiraldez: is the `:receive("*a")` after `:shutdown("send")` enough?

added the `return ngx.exit(ngx.OK)`

javierguerragiraldez:
superceded by #6552

PR #6376, opened 2020-09-21
feat(clustering) automatically purge entries from the cluster_data_planes table so it won't grow uncontained

dndx:
cluster_data_planes table so it won't grow uncontained

Also adds new config option `conf.cluster_status_ttl`, that allows the
TTL of cluster status entries to be overwritten by user. The default TTL
is 14 days.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -242,6 +242,17 @@
                          # This setting has no effect if `role` is not set to
                          # `control_plane`.
 
+#cluster_data_plane_purge_delay = 1209600
+                         # How many seconds must pass from the time a DP node
+                         # becomes offline to the time its entry gets removed
+                         # from the database, as returned by the
+                         # /clustering/data_planes Admin API endpoint.
+                         #
+                         # This is to prevent the cluster data plane table from
+                         # growing indefinitely. The default is set to
+                         # 14 days. That is, if CP haven't heard from a DP for
+                         # 14 days, its entry will be removed.
+
 #------------------------------------------------------------------------------
 # NGINX
 #------------------------------------------------------------------------------
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -576,6 +576,7 @@ local CONF_INFERENCES = {
   cluster_mtls = { enum = { "shared", "pki" } },
   cluster_ca_cert = { typ = "string" },
   cluster_server_name = { typ = "string" },
+  cluster_data_plane_purge_delay = { typ = "number" },
   kic = { typ = "boolean" },
 }
 
@@ -938,6 +939,10 @@ local function check_and_infer(conf, opts)
       errors[#errors + 1] = "only in-memory storage can be used when role = \"data_plane\"\n" ..
                             "Hint: set database = off in your kong.conf"
     end
+  end
+
+  if conf.cluster_data_plane_purge_delay < 60 then
+    errors[#errors + 1] = "cluster_data_plane_purge_delay must be 60 or greater"
   end
 
   if conf.role == "control_plane" or conf.role == "data_plane" then
--- a/kong/templates/kong_defaults.lua
+++ b/kong/templates/kong_defaults.lua
@@ -25,6 +25,7 @@ cluster_cert_key = NONE
 cluster_mtls = shared
 cluster_ca_cert = NONE
 cluster_server_name = NONE
+cluster_data_plane_purge_delay = 1209600
 mem_cache_size = 128m
 ssl_cert = NONE
 ssl_cert_key = NONE
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -324,7 +324,7 @@ function _M.handle_cp_websocket()
           data ~= "" and data or nil,
         hostname = node_hostname,
         ip = node_ip,
-      })
+      }, { ttl = kong.configuration.cluster_data_plane_purge_delay, })
       if not ok then
         ngx_log(ngx_ERR, "unable to update clustering data plane status: ", err)
       end
--- a/kong/db/migrations/core/012_213_to_220.lua
+++ b/kong/db/migrations/core/012_213_to_220.lua
@@ -6,8 +6,10 @@ return {
         hostname       TEXT NOT NULL,
         ip             TEXT NOT NULL,
         last_seen      TIMESTAMP WITH TIME ZONE DEFAULT (CURRENT_TIMESTAMP(0) AT TIME ZONE 'UTC'),
-        config_hash    TEXT NOT NULL
+        config_hash    TEXT NOT NULL,
+        ttl            TIMESTAMP WITH TIME ZONE
       );
+      CREATE INDEX IF NOT EXISTS clustering_data_planes_ttl_idx ON clustering_data_planes (ttl);
 
       DO $$
       BEGIN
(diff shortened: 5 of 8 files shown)


kikito (approved):
LGTM, tests are red though (I suspect unrelated)

kong/templates/kong_defaults.lua
```diff
@@ -25,6 +25,7 @@ cluster_cert_key = NONE
 cluster_mtls = shared
 cluster_ca_cert = NONE
 cluster_server_name = NONE
+cluster_status_ttl = 1209600
```
hbagdi, line 28: Thinking form a user who doesn't know about internals of Kong, `ttl` in this sense is a pretty confusing acronym to pick for this configuration property. The fact that internally it is driven by `ttl` mechanism is an implementation detail that shouldn't surface here.
Would renaming this to something along the lines `cluster_node_purge_delay` be an improvement?

kong/conf_loader/init.lua
```diff
@@ -576,6 +576,7 @@ local CONF_INFERENCES = {
   cluster_mtls = { enum = { "shared", "pki" } },
   cluster_ca_cert = { typ = "string" },
   cluster_server_name = { typ = "string" },
+  cluster_status_ttl = { typ = "number" },
```
hbagdi, line 579: Please add the property to `kong.conf.default` along with it's documentation.

hbagdi (dismissed):

hbagdi:
This comment might be broader than this PR but here it goes nevertheless.

We need to start thinking differently when it comes to such features and how we expose them to our users.
`kong.conf` has become pretty heavy weight and has become a kitchen-sink for all types of configuration, to an extent where it is hurting our customer experience quite a bit.
Putting on my operator hat, tweaking kong.conf is usually considered a large and risky change. Now, we, the developers of Kong, do understand that tweaking something like this property should not result in any disruption and such a change can be rolled out gradually without much risk. The reality is different because there is a lot of tooling involved in such rollouts. And to that end, we should aim to minimize the number of knobs we put inside kong.conf to minimize the risk of changes that users can make. A knob should really go into kong.conf if the underlying system really needs it to start up or it needs it during startup because it can't be tweaked after startup.

For configuration options like this, we should aim to make them configurable without node restarts. Putting them inside the database might be an idea to think about. This allows to drive more configuration via Admin API, which would be a very welcoming change and help our community and Enterprise users. Think about how postgres defines settings in conf as well as inside the DB itself, where some settings can be tweaked without restarting the database.


cc @hishamhm

bungle:
> This comment might be broader than this PR but here it goes nevertheless.
> 
> We need to start thinking differently when it comes to such features and how we expose them to our users.
> `kong.conf` has become pretty heavy weight and has become a kitchen-sink for all types of configuration, to an extent where it is hurting our customer experience quite a bit.
> Putting on my operator hat, tweaking kong.conf is usually considered a large and risky change. Now, we, the developers of Kong, do understand that tweaking something like this property should not result in any disruption and such a change can be rolled out gradually without much risk. The reality is different because there is a lot of tooling involved in such rollouts. And to that end, we should aim to minimize the number of knobs we put inside kong.conf to minimize the risk of changes that users can make. A knob should really go into kong.conf if the underlying system really needs it to start up or it needs it during startup because it can't be tweaked after startup.
> 
> For configuration options like this, we should aim to make them configurable without node restarts. Putting them inside the database might be an idea to think about. This allows to drive more configuration via Admin API, which would be a very welcoming change and help our community and Enterprise users. Think about how postgres defines settings in conf as well as inside the DB itself, where some settings can be tweaked without restarting the database.
> 
> cc @hishamhm

Time to add `config` table. ;-)

dndx:
@hbagdi I changed the property to `cluster_data_plane_purge_delay` because it is more consistent with the Admin API names. Also, it leaves opportunity to use a different duration for control plane purge delay in the future.

kong.conf.default
```diff
@@ -242,6 +242,17 @@
                          # This setting has no effect if `role` is not set to
                          # `control_plane`.
 
+#cluster_data_plane_purge_delay = 1209600
+                         # How many seconds must pass from the time a DP node
+                         # becomes offline to the time it's entry gets removed
```
kikito, line 247:
```suggestion
                         # becomes offline to the time its entry gets removed
```

kikito (changes requested):

kikito:
To continue @hbagdi line of thought:

> For configuration options like this, we should aim to make them configurable without node restarts

Let's consider the "not having knobs at all for this".

The purpose of this PR is containing the grow of the cluster_data_planes table. 

Let's consider why someone would need to make the TTL smaller. This would be someone who has too many rows in the table. What is an "high ceiling" for this? Someone which generates 10000 new CAs per day (and none of them refreshes, never)? Even for that case, that would generate 140000 rows in 14 days, and then stabilize at ~140000 rows. That seems acceptable. Perhaps I am underestimating how much people might "churn".

Now let's consider why someone would need to make the TTL bigger. The only reason I can think about is someone who needs to get preserve the `last_seen_at` property associated to each host - perhaps for forensics/tracing. I think if they really need this information they should be getting it daily from the admin api and exporting it into to a long-term storage solution anyway.

With this two scenarios in mind, I think that maybe we could leave the config knob out completely. It might be that just having any ttl _at all_ is good enough for everyone. Maybe making the default smaller - 1 day or so?

hbagdi:
> a different duration for control plane purge delay in the future.

I disagree with this reason because we don't have a use cases where customers would need different parameters for different nodes. We could certainly convince ourselves otherwise but this falls into the bucket of adding complexity for little gain. I don't feel strongly enough so the new name is fine by me.

> Why would anyone need to make the TTL bigger? The only reason I can think about is someone who needs to get preserve the last_seen_at property associated to each host - perhaps for forensics/tracing. I think if they really need this information they should be getting it daily from the admin api and exporting it into to a long-term storage solution anyway.

I can argue either ways but I see where @kikito is going with that line of though. Whatever we end up deciding, we need to ensure that the retention is long enough. Kong is used in k8s environments, which have a relatively high churn of nodes. But then it is also used in bare metal environments, where the churn is significantly less.

kong.conf.default
```diff
@@ -242,6 +242,17 @@
...
+#cluster_data_plane_purge_delay = 1209600
+                         # How many seconds must pass from the time a DP node
+                         # becomes offline to the time its entry gets removed
+                         # from the database, as returned by the
+                         # /clustering/data_planes Admin API endpoint.
+                         #
+                         # This is to prevent the cluster data plane table from
+                         # growing indefinitely. The default is set to
+                         # 14 days. That is, if CP haven't heard from a DP for
+                         # 14 days, it's entry will be removed.
```
kikito, line 254:
```suggestion
                         # 14 days, its entry will be removed.
```

dndx:
Since the reviews were addressed and tests are running green, let's merge this for now and discuss longer term plans in a separate PR (regarding the config interface and so on).

Having this option is essential in a high churn environment such as k8s as mentioned by @hbagdi, while having minimum impact to low-churn setups so it is a reasonable balance between complexity and user friendless. We may revisit the approach down the road, but for now time based purge should solve at least the problem of uncontained table growth, which will cause undesirable database slowness in a prod environment.

@kikito @hbagdi if it's alright with you, let's ✅ this PR so it won't hold off the upcoming 2.2 release :)

kikito (approved):

PR #6380, opened 2020-09-22
chore(tests) include kong*.rockspec to cache hash calculation on travis

bungle:
### Summary

If it is not included in cache hash calculation, dependencies that are updated in .rockspec might be leaked to other tests / builds.

### DIFF
--- a/.ci/setup_env.sh
+++ b/.ci/setup_env.sh
@@ -10,7 +10,7 @@ LUAROCKS=$(dep_version RESTY_LUAROCKS_VERSION)
 OPENSSL=$(dep_version RESTY_OPENSSL_VERSION)
 GO_PLUGINSERVER=$(dep_version KONG_GO_PLUGINSERVER_VERSION)
 
-DEPS_HASH=$(cat .ci/setup_env.sh .travis.yml .requirements | md5sum | awk '{ print $1 }')
+DEPS_HASH=$({ cat .ci/setup_env.sh .travis.yml .requirements Makefile; cat kong-*.rockspec | awk '/dependencies/,/}/'; } | md5sum | awk '{ print $1 }')
 INSTALL_CACHE=${INSTALL_CACHE:=/install-cache}
 INSTALL_ROOT=$INSTALL_CACHE/$DEPS_HASH
 


dndx:
I would like to hear more opinions from others, for the following reasons:

1. Every time before the tests are ran, `make dev` is executed, which will correct all the dependency issue, if any.
2. Adding `rockspec` makes the cache granularity unnecessarily large. This means even if we are just adding new Kong lua files to the `rockspec`, a full rebuild is triggered, which is very slow and reduces our cache hit rate.

Basically, everything that can not be corrected by `make dev` are inside the `.requirments` file, which is part of the hash as f today. Everything else should be correctable by `make dev` much quicker than a full rebuild.

Obviously, being defensive is a good thing, for the CI runs the most defensive thing we could have done is to not use cache at all. We choose to not do that because it creates more harm without solving any known issues for us, the same principle should apply here as well.

bungle:
> I would like to hear more opinions from others, for the following reasons:
> 
> 1. Every time before the tests are ran, `make dev` is executed, which will correct all the dependency issue, if any.

It does not seem like. Why this PR is green, while others are red?

> 2. Adding `rockspec` makes the cache granularity unnecessarily large. This means even if we are just adding new Kong lua files to the `rockspec`, a full rebuild is triggered, which is very slow and reduces our cache hit rate.

We do this non-frequently.

bungle:
> Basically, everything that can not be corrected by `make dev` are inside the `.requirements`

We tried couple of things. But didn't find the solution that is better than this. Could we go with this solution as it makes cache correct, and then try to optimize it later? If you have better way to solve this, I am open to see it. Here is a sister PR to this that is for `next` and it is `green` on travis: https://github.com/Kong/kong/pull/6383 (so this fix really works). The Github Actions fix can be found on kong-build-tools.

If cache gets correct, everything will suddenly turn green, so you never know do you have this: https://github.com/Kong/kong/pull/6361 screwing up the cache. It depends that does `cache` have that or almost any other branch based on `next`.

kikito:
We could only hash the rockspec dependencies, but not when adding a new file to Kong (or when adding a blank line to the rockspec).

```
resty -e "setmetatable(_G,nil); assert(loadfile('kong-2.1.3-0.rockspec'))(); table.sort(dependencies); for _,d in ipairs(dependencies) do print(d) end"
```

Output looks like:
```
inspect == 3.1.1
kong-lapis == 1.8.1.2
kong-plugin-acme ~> 0.2
...
```

Could we hash that 👆instead?

bungle:
> We could only hash the rockspec dependencies, but not when adding a new file to Kong (or when adding a blank line to the rockspec).
> 
> ```
> resty -e "setmetatable(_G,nil); assert(loadfile('kong-2.1.3-0.rockspec'))(); table.sort(dependencies); for _,d in ipairs(dependencies) do print(d) end"
> ```
> 
> Output looks like:
> 
> ```
> inspect == 3.1.1
> kong-lapis == 1.8.1.2
> kong-plugin-acme ~> 0.2
> ...
> ```
> 
> Could we hash that 👆instead?

Or just `cat kong-2.1.4-0.rockspec | awk '/dependencies/,/}/ { print }'`

bungle:
Something like:
```
DEPS_HASH=$({ cat .ci/setup_env.sh .travis.yml .requirements Makefile; cat kong-*.rockspec | awk '/dependencies/,/}/'; } | md5sum | awk '{ print $1 }')
```

bungle:
> Something like:
> 
> ```
> DEPS_HASH=$({ cat .ci/setup_env.sh .travis.yml .requirements Makefile; cat kong-*.rockspec | awk '/dependencies/,/}/'; } | md5sum | awk '{ print $1 }')
> ```

I updated the PR with this change.

bungle:
I'll close this, let's just merge this to `next`:
https://github.com/Kong/kong/pull/6383

Issue #183, opened 2015-04-29
[request] support for DynamoDB


nickveenhof:
Cassandra is quite a beast to setup and maintain. If Kong runs in EC2,it would be great if it could also write it's data to DynamoDB (very similar to Cassandra) and use that as a service. This allows engineers to have less maintenance on Cassandra and focus more on Kong. 

Great project!

thibaultCha:
We are aware of this, and support for other databases is on the roadmap, but we have a lot of things to do before that.

jakehow:
:+1:

DavidTPate:
For me, supporting another database which I can have managed (see RDS/DynamoDB/etc) would be awesome.

sinzone:
btw.. @DavidTPate @nickveenhof you can have Cassandra in the cloud through https://www.instaclustr.com/

jasonfill:
+1

berkay:
+1
IMO, Cassandra dependency is a major problem. It would be unwise for anyone to build such a critical service (API) with a dependency on a data store that one does not fully grok but have to manage. DynamoDB would be a good fit.

desmondmorris:
+1

glkz:
:+1:

owahlen:
+1

vmattos:
:+1:

kenwarner:
+1

skippy:
+1

udlei:
+1

vwal:
+1

While PostgreSQL support would be great, support for DynamoDB would be even more critical for AWS applications. If KONG is used on Auto-Scaling Group instances, DynamoDB access would allow much lighter instances to be used as Cassandra is a resource hog.

sinzone:
@vwal that is true. and we're def going to do both. We just prioritized Postgres at the moment given more community demand for it.

seiffert:
:+1:

gankkank:
:+1:  Glad to hear that

harlow:
> While PostgreSQL support would be great, support for DynamoDB would be even more critical for AWS applications. If KONG is used on Auto-Scaling Group instances, DynamoDB access would allow much lighter instances to be used as Cassandra is a resource hog.

Agreed. :+1: for DynamoDB.

jtconnor:
+1 for DynamoDB

RussellLuo:
I think it will be awesome if KONG provides the extension/plugin support for different databases.

eric1iu:
[...]

PR #6381, opened 2020-09-22
chore(deps) bump luacheck (used for tests) from 0.23 to 0.24

bungle:
#### Changes

* Caching now uses files in a global directory instead of local
  `.luacheckcache` file. Default cache directory is
  `%LOCALAPPDATA%\Luacheck\Cache` on Windows,
  `~/Library/Caches/Luacheck` on OS X/macOS, and
  `$XDG_CACHE_HOME/luacheck` or `~/.config/luacheck` on other systems.

#### Fixes

* Fixes for using Luacheck with Lua 5.4:
  - Parse 5.4 attributes in `local` declarations (but ignores them for now)
  - Luacheck itself can also run with Lua 5.4
* Fixed `randomize` missing from `busted` set of standard globals (#183).
* Added additional `table` and `thread` definitions for `ngx_lua`.
* Added additional `match` definition for `busted`.

#### Miscellaneous

* Upgraded Windows binary components: Lua 5.3.4 -> 5.3.5,
  LuaFileSystem 1.6.3 -> 1.7.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 OS := $(shell uname | awk '{print tolower($$0)}')
 MACHINE := $(shell uname -m)
 
-DEV_ROCKS = "busted 2.0.0" "busted-htest 1.0.0" "luacheck 0.23.0" "lua-llthreads2 0.1.5" "http 0.3" "ldoc 1.4.6"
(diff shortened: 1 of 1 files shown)


kikito (approved):
LGTM, but tests are red. Seems unrelated though?

PR #6383, opened 2020-09-22
chore(tests) include kong*.rockspec to cache hash calculation on travis

bungle:
### Summary

If it is not included in cache hash calculation, dependencies that are updated in .rockspec are leaked to other tests / builds.


bungle:
This is alternative to `#6380`, but this is to `next`. Perhaps it is good enough to be merged to next as dependencies are not bumped on `master`.

kikito (approved):

PR #6386, opened 2020-09-23
make kong.log.serialize message `tries` with cjson.empty_array_mt

tianzhipeng-git:
### Summary

Make kong.log.serialize message `tries` with cjson.empty_array_mt.
All log plugins use this method to genarate log message, but they don't handle `tries` emtpy.
So the json log is really difficult for other languages to decode.

### Full changelog
only four line:
```
cjson = require "cjson"
local tries = (ctx.balancer_data or {}).tries or {}
setmetatable(tries, cjson.array_mt)
tries=tries,
```

### Issues resolved

Fix #XXX

### DIFF
--- a/kong/pdk/log.lua
+++ b/kong/pdk/log.lua
@@ -15,6 +15,7 @@ local ngx_re = require "ngx.re"
 local inspect = require "inspect"
 local ngx_ssl = require "ngx.ssl"
 local phase_checker = require "kong.pdk.private.phases"
+local cjson = require "cjson"
 
 
 local sub = string.sub
@@ -559,6 +560,9 @@ do
     end
 
     local host_port = ctx.host_port or var.server_port
+    
+    tries = (ctx.balancer_data or {tries = {}}).tries
+    setmetatable(tries, cjson.array_mt)
 
     return {
       request = {
@@ -576,7 +580,7 @@ do
         headers = resp_headers,
         size = var.bytes_sent
       },
-      tries = (ctx.balancer_data or {}).tries,
+      tries = tries,
       latencies = {
         kong = (ctx.KONG_ACCESS_TIME or 0) +
                (ctx.KONG_RECEIVE_TIME or 0) +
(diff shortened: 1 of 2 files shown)


tianzhipeng-git:
I don't think it is necessary that tries should be nil, change the test case `/01-unit/10-log_serializer_spec.lua`: 
```
    it("does not fail when the 'balancer_data' structure is missing", function()
      ngx.ctx.balancer_data = nil

      local res = kong.log.serialize({ngx = ngx, kong = kong, })
      assert.is_table(res)

      #assert.is_nil(res.tries) 
      assert.is_table(res.tries) 
    end)
```
And I don't know which branch it should been merge into, simply set to master

kong/pdk/log.lua
```diff
@@ -559,6 +560,9 @@ do
     end
 
     local host_port = ctx.host_port or var.server_port
+    
+    local tries = (ctx.balancer_data or {}).tries or {}
+    setmetatable(tries, cjson.array_mt)
```
Tieske, line 565: here you are creating an empty table if there is none. imo we should not do that, it adds more processing and data to the logs without conveying additional information. The value being absent should do.

Tieske: Also setting a metatable on an existing table (which is a shared data structure) doesn't seem like a good idea. I'd rather attach the metatable right when the `tries` table is being created, somewhere in the `balancer` phase iirc.
That would make it more clear for every user of that table, it is an array, and hence be easier on future us for maintenance.

Tieske (changes requested):
can you have a look at the comments?

kong/pdk/log.lua
tianzhipeng-git: You are right, this is not a good place to do these things that should been handled at balancer phase.

tianzhipeng-git: Now the `ctx.balancer_data` may contains empty tries table or tries array, which is uncertained. I will open an issue to describe this problem and close this pr.

PR #6389, opened 2020-09-23
Always count KONG_WAITING_TIME

### LINKED ISSUE #6180: HTTP 499s and logging results delivered
### Summary

Those familiar with NGINX know that an HTTP 499 means the client severed the connection prior to allowing the NGINX Webserver(Kong) to respond on a given request, usually when NGINX is just waiting for the backend API service to reply. 

Note BackendLatency being -1, which is from this field:
https://github.com/Optum/kong-splunk-log/blob/master/src/basic.lua#L58
```
ctx.KONG_WAITING_TIME or -1,
```

When the client cuts the connection short, why is the time window already spent waiting on the backend totally lost in Kong? Very specifically I would expect the ```ctx.KONG_WAITING_TIME``` to have a value close to 4003 (4007 - 4ms(from access/rewrite time)). Which ultimately looks like a 4 second client read timeout to Kong/Nginx judging by the TotalLatency field. I can see how some folks would misinterpret the log below to be that Kong had trouble reverse proxying to the backend API. 

```
{ [-]
   BackendLatency: -1
   CID: 7869c6c2-6f07-4597-a138-63936e618bca#558349
   ClientIP: 65.xxx.x.xx
   Consumer: consumer1
   GatewayHost: gateway.company.com
   GatewayPort: 443
   HTTPMethod: POST
   HTTPStatus: 499
   KongLatency: { [-]
     AccessTime: 3
     BalancerTime: 0
     ReceiveTime: 0
     RewriteTime: 1
   }
   RequestSize: 2519
   ResponseSize: 0
   RoutingURL: my-api:443/R4/Bundle/$service
   ServiceName: my.api.service.v1
   TotalLatency: 4007
   Tries: [ [+]
   ]
   URI: /api/test/api/fhir/v1/R4/Bundle/$service
}
Show as raw text
```

### Steps To Reproduce

1. Force a 499, look at the latency reported on the communication with the backend service.

### Additional Details & Logs

- Kong version 2.1.0 (or any version for that matter)

javierguerragiraldez:
On errors (especially 499), it's possible that some phases are skipped.
Even on those cases, Kong spent time waiting, so it should be accounted
not only on header-filter but on body-filter (unlikely!) or log phase.

Fix #6180

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -1081,6 +1081,10 @@ function Kong.body_filter()
                                                              ctx.KONG_BALANCER_ENDED_AT or
                                                              ctx.KONG_BALANCER_START or
                                                              ctx.KONG_ACCESS_ENDED_AT)
+    if not ctx.KONG_WAITING_TIME then
+      ctx.KONG_WAITING_TIME = ctx.KONG_BODY_FILTER_START -
+                              (ctx.KONG_BALANCER_ENDED_AT or ctx.KONG_ACCESS_ENDED_AT)
+    end
   end
 end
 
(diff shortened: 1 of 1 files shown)


bungle:
Yes, could it be that 499 means Nginx will not run header_filter and body_filter but jump directly to log. Of that is the case I would remove this code from body filter.

bungle (approved):

PR #6392, opened 2020-09-24
chore(deps) remove resty.iputils dependency

bungle:
### Summary

The resty.iputils lacks ipv6 support and we don't use it anymore. This change removes the last dependency to it and removes the dependency from luarocks.

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -3,19 +3,22 @@
 local utils = require "kong.tools.utils"
 local openssl_pkey = require "resty.openssl.pkey"
 local openssl_x509 = require "resty.openssl.x509"
-local iputils = require "resty.iputils"
-local Schema = require("kong.db.schema")
-local socket_url = require("socket.url")
+local Schema = require "kong.db.schema"
+local socket_url = require "socket.url"
 local constants = require "kong.constants"
 local px = require "resty.mediador.proxy"
+local ipmatcher = require "resty.ipmatcher"
 
 
+local tostring = tostring
 local pairs = pairs
 local pcall = pcall
 local match = string.match
 local gsub = string.gsub
+local find = string.find
 local null = ngx.null
 local type = type
+local sub = string.sub
 
 
 local function validate_host(host)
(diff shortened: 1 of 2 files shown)


murillopaula (approved):

kong/db/schema/typedefs.lua
```diff
@@ -63,15 +66,34 @@ local function validate_ip_or_cidr(ip)
...
+  local block = tonumber(sub(ip_or_cidr, p + 1), 10)
+  if not block then
+    return nil, "invalid ipv4 cidr range: " .. ip_or_cidr
+  end
+
+  for i = 0, 32 do
+    if block == i then
+      return true
+    end
+  end
```
dndx, line 94: How about:
```
return block >= 0 and block <=32
```

instead? Doesn't seems that the loop is necessary.

dndx (changes requested):

kong/db/schema/typedefs.lua
dndx: Actually, if we are concerned about people feeding CIDR like this:

`127.0.0.0/12.5`

we could just check for whether a fraction part exists. So:

```
if math.floor(block) ~= block then
  return nil, "invalid ipv4 cidr range: " .. ip_or_cidr .. " prefix length must be a whole number"
end
```

should allow us to get rid of the loop still.

bungle: Also `0xa` is probably passing.

bungle: I was also thinking about building a lookup table.

bungle: It is now changed to `lookup table`.

dndx: I think @fffonion recommended to use a regex to check for the CIDR format, might be easier and more robust.

bungle: This is the most robust I can think of. Won't get anymore robust with regex. Just inserts complex tech to simple problem already solved.

dndx (approved):

PR #6393, opened 2020-09-24
chore(deps) remove resty.mediador dependency

bungle:
### Summary

The resty.mediador dependency can be replaced with ipmatcher, and that's what this commit does.

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -11,7 +11,6 @@ local tablex = require "pl.tablex"
 local utils = require "kong.tools.utils"
 local log = require "kong.cmd.utils.log"
 local env = require "kong.cmd.utils.env"
-local ip = require "resty.mediador.ip"
 
 
 local fmt = string.format
@@ -887,7 +886,7 @@ local function check_and_infer(conf, opts)
 
   -- checking the trusted ips
   for _, address in ipairs(conf.trusted_ips) do
-    if not ip.valid(address) and address ~= "unix:" then
+    if not utils.is_valid_ip_or_cidr(address) and address ~= "unix:" then
       errors[#errors + 1] = "trusted_ips must be a comma separated list in " ..
                             "the form of IPv4 or IPv6 address or CIDR "      ..
                             "block or 'unix:', got '" .. address .. "'"
--- a/kong-2.1.4-0.rockspec
+++ b/kong-2.1.4-0.rockspec
@@ -30,7 +30,6 @@ dependencies = {
   "lua_pack == 1.0.5",
   "lua-resty-dns-client == 5.1.0",
   "lua-resty-worker-events == 1.0.0",
-  "lua-resty-mediador == 0.1.2",
   "lua-resty-healthcheck == 1.3.0",
   "lua-resty-cookie == 0.1.0",
   "lua-resty-mlcache == 2.4.1",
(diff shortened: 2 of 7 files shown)


murillopaula:
We're repeating a lot of code here. Can we abstract out this:

```
  local p = string.find(ip_or_cidr, "/", 1, true)
  if not p then
    return false
  end

  local block = tonumber(string.sub(ip_or_cidr, p + 1), 10)
  if not block then
    return false
  end

  local ip = string.sub(ip_or_cidr, 1, p - 1)
  if ipmatcher.parse_ipv4(ip) then
    for i = 0, 32 do
      if block == i then
        return true
      end
    end

  elseif ipmatcher.parse_ipv6(ip) then
    for i = 0, 128 do
      if block == i then
        return true
      end
    end
  end

  return false
end
```

In a common module (e.g. `https://github.com/Kong/kong/blob/next/kong/tools/utils.lua`) and then start reusing it?

bungle:
> In a common module (e.g. `https://github.com/Kong/kong/blob/next/kong/tools/utils.lua`) and then start reusing it?

Perhaps later. I would like that `utils` go away totally. It is a junk yard.

bungle:
But I agree, we should add these at some point to `kong.ip.*`. Still with that the code needed here is probably needed even before we have `kong` ready.

kong/conf_loader/init.lua
```diff
@@ -11,11 +11,55 @@ local tablex = require "pl.tablex"
...
+  elseif ipmatcher.parse_ipv6(ip) then
+    for i = 0, 128 do
+      if block == i then
+        return true
+      end
+    end
+  end
+
+  return false
+end
```
murillopaula, line 62: Noob question: If `ipmatcher` could not parse, shouldn't we return `false` and that's it? What is the purpose of all the extra logic here? We didn't have this logic here before. I assume it's because `ipmatcher` can't do what mediador was doing before. It seems like we are implementing something new. Could it be a patch (bug or feature) to the `ipmatcher` lib?

bungle: Yes, so `ipmatched.parse_ipv4/6` only takes ip addresses. At the very top we check if it was ip address, if it was not, perhaps it is a cidr block, thus we split the input string in two `/` and then parse if the address is ipv4, if it is ipv4 we also check that `/xx` is between 0-32 (all possible masks). For ipv6 there are other notations too, but atm we just support `/xxx` notation where the xxx is between 0-128.

So logic of this is:
1. is it ipv4 or ipv6
2. is it ipv4 or ipv6 cidr block

bungle: Yes, I wish the library itself could be better in this. There is `ipmatcher.new` but on my testing it was not good.

murillopaula: ahh I see, now I remember about this: https://github.com/api7/lua-resty-ipmatcher/issues/11

murillopaula (approved):

dndx:
> > In a common module (e.g. `https://github.com/Kong/kong/blob/next/kong/tools/utils.lua`) and then start reusing it?
> 
> Perhaps later. I would like that `utils` go away totally. It is a junk yard. My way of thinking:
> 
> Is it repeated:
> 
> * can this easily be shared between places where it is repeated, without adding new and extra requires?
> * is it copied more than two places?
> 
> In this case both of the above questions say `no`, so I think copy pasting is the best solution for now. We don't need to go a way of leftpad. I didn't even use `utils.split` in one of the places even when utils was there. How hard it is to split? Not that hard.

Same, PDK functions seems to be a better place for it, as it means 3rd party plugins could use it easily as well.

bungle:
This PR is now reworked to utility functions.

kikito (approved):

PR #6395, opened 2020-09-24
chore(admin) poll worker events before and after the changes

bungle:
### Summary

This PR makes admin api calls to poll worker events before and after actually executing the admin api call function. It makes admin api less dependent on background timers, and makes usage of admin api more predictable.

### DIFF
--- a/kong/api/api_helpers.lua
+++ b/kong/api/api_helpers.lua
@@ -257,6 +257,8 @@ local function parse_params(fn)
 
     local res, err = fn(self, ...)
 
+    kong.worker_events.poll()
+
     if err then
       kong.log.err(err)
       return ngx.exit(500)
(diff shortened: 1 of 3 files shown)


kong/api/api_helpers.lua
```diff
@@ -257,6 +257,8 @@ local function parse_params(fn)
 
     local res, err = fn(self, ...)
 
+    kong.worker_events.poll()
```
javierguerragiraldez, line 260: even if `fn()` returned an error?

bungle: @javierguerragiraldez, do you mean it would be beneficial to log it with e.g. `NOTICE`?

bungle: Ok, now I understand. Yes, I thought about it. At some point I had check whether `method` was `POST|PUT|PATCH|DELETE`, but removed it. The poll in most cases is fast. Because we don't know where `fn()` fails, I just decided to call it always.

javierguerragiraldez (approved):

PR #6400, opened 2020-09-28
fix(schema) more descriptive error logs when type is invalid

WALL-E:


bungle:
@WALL-E can you fix tests so that they get green?

WALL-E:
> @WALL-E can you fix tests so that they get green?

ok

PR #6402, opened 2020-09-28
feat(db) check MIN and DEPRECATED versions on connector init

eskerda:
### Summary

This PR adds a version check on db connector initialization. Any version lower that `MIN` will throw an error, unless it's greater or equal to `DEPRECATED`, on which case a warning is shown. 

It's a direct port of the feature that was present on old DAO (https://github.com/Kong/kong/pull/6167 and https://github.com/Kong/kong/commit/a9f5af1a53a701bd74c9ed6683390c3f1b43fc6b). Moved some of the specs to be unit tests instead of integration by mocking related data structures.

### DIFF
--- a/kong/constants.lua
+++ b/kong/constants.lua
@@ -152,7 +152,6 @@ local constants = {
     "kong_rate_limiting_counters",
   },
   DATABASE = {
-    -- These constants are not used anywhere in the Kong code. They're here as a useful reference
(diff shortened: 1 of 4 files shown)


hishamhm (approved):
LGTM! Thanks @eskerda for the PR and all the context links in the description!

PR #6407, opened 2020-09-30
feat(clustering) pre-create the config cache file with 0600 permission

dndx:
on startup

### DIFF
--- a/spec/02-integration/09-hybrid_mode/01-sync_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/01-sync_spec.lua
@@ -177,11 +177,11 @@ for _, strategy in helpers.each_strategy() do
       end)
 
       it("local cached config file has correct permission", function()
-        local handle = io.popen("stat servroot2/config.cache.json.gz")
+        local handle = io.popen("ls -l servroot2/config.cache.json.gz")
         local result = handle:read("*a")
         handle:close()
 
-        assert.matches("Access: (0600/-rw-------)", result, nil, true)
+        assert.matches("-rw-------", result, nil, true)
       end)
     end)
   end)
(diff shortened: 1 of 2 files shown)


spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -175,6 +175,14 @@ for _, strategy in helpers.each_strategy() do
           end
         end, 5)
       end)
+
+      it("local cached config file has correct permission", function()
+        local handle = io.popen("stat servroot2/config.cache.json.gz")
```
javierguerragiraldez, line 180: `stat` isn't POSIX, it's part of the GNU coreutiils.  do we run on any BSD system?  (i guess Darwin does install GNU tools now)

eskerda:
Just thinking out loud, will `update_config` thread also always run on the master process?

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
hishamhm: > do we run on any BSD system?

Some devs run the test suite on macOS. `stat` flavors are always a portability annoyance. A simple `ls -l` should be enough for the purposes of this test.

javierguerragiraldez:
> will `update_config` thread also always run on the master process?

seems so, since the local `communicate()` is only entered from within the `if ngx.worker.id() == 0 then` block.

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
dndx: Actually on macOS 10.15, `stat` is already pre-installed, so it should not pose any challenges for running on macOS.

```
~ datong$ which stat
/usr/bin/stat
```

dndx: Changed to `ls -l`.

bungle (approved):

PR #6408, opened 2020-09-30
fix(hybrid) broadcast CRUD changes to DPs cluster

kikito:
This change fixes a bug which manifested itself when Kong:

* It is working on hybrid mode. 
* The CP role is being executed by several Kong nodes in a
  cluster.
* A change (any successful CRUD operation) happens in one of these
  nodes.

The desired behaviour is that when this happens, all the CPs in the cluster
would trigger a DP refresh. The bug consists in that only the workers in the
same Kong node as the one which received the CRUD operation do such
refresh.

Example setup: Two Kong nodes in CP role. Node A has workers A1 and A2, and Node
B has workers B1 and B2. The admins use A1's admin API to create a new Service.

Before this fix, A1 would trigger the CP reload, then A2 would "see" the change
(via a worker event) and refresh its DPs. But B1 and B2 would not perform such
refresh.

This fix makes A1 trigger a cluster event in addition of the previous
worker-level event. It also makes all workers look for this new cluster event,
and do a DP refresh.

Note that the test accompanying this fix does not exactly replicate the above
setup. There is a concern that the amount of memory this would need (at least 3
kong nodes, with several workers each) would trigger a CI shutdown. Instead, we
reduced the example to the minimum (only A1 and B1 are created) and we check
B1's logs for the reception of the change. The fact that B1 reliably
communicates refreshes its DPs when it receives this notification is assumed,
since the function triggered by the event is the same one used in worker-level
updates, which are already tested.


kong/clustering.lua
```diff
@@ -454,25 +455,42 @@ function _M.init_worker(conf)
...
 
     -- ROLE = "control_plane"
+    local db = kong.db
 
     kong.worker_events.register(function(data)
       -- we have to re-broadcast event using `post` because the dao
       -- events were sent using `post_local` which means not all workers
       -- can receive it
+
+      -- send event to other workers in the same node
```
bungle, line 465: It sends it to all workers in node, including itself.

kong/clustering.lua
```diff
@@ -454,25 +455,42 @@ function _M.init_worker(conf)
...
       local res, err = kong.worker_events.post("clustering", "push_config")
       if not res then
         ngx_log(ngx_ERR, "unable to broadcast event: " .. err)
       end
+
+      -- send event to other nodes in the cluster
+      local cache_key = db[data.schema.name]:cache_key(data.entity)
+      local key = fmt("%s:%s", data.operation, cache_key)
+      res, err = kong.cluster_events:broadcast("clustering:push_config", key)
```
bungle, line 474: can’t we just subcribe the invalidations?

bungle (changes requested):

kong/clustering.lua
kikito: Yes, fixing my comment

kong/clustering.lua
kikito: Talked about this with @bungle privately. I misunderstood how certain things work in the `cluster_events` library. Will redo the PR accordingly.

bungle (approved):

dndx (approved):

dndx:
Update: after discussing with @bungle , it seems that the `invalidations` event is always triggered at least for all of the core entities and all the bundled plugin entries. If this became an issue, then we will fire our own cluster events from `dao:crud` instead. But this PR should make the situation better regardless, so it is decided to merge it first.

PR #6409, opened 2020-09-30
fix(hybrid) more robust clustering connection and thread management

bungle:
### Summary

Implements a more robust websockets connection and thread management for clustering endpoint.

1. Tries to properly close the connection and notify the client
2. Tries to properly kill the started threads on reconnection
3. Adjusts so that expected events are not logged as errors
   - Timeouts
   - Closing of connection
4. Does not start zombie threads

Fixes an issue reported by other team.

This also contains additional commit to not cause push configs on clustering_data_planes events.

### DIFF
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -11,6 +11,7 @@ local utils = require("kong.tools.utils")
 local openssl_x509 = require("resty.openssl.x509")
 local system_constants = require("lua_system_constants")
 local ffi = require("ffi")
+local string = string
 local assert = assert
 local setmetatable = setmetatable
 local type = type
@@ -36,13 +37,17 @@ local deflate_gzip = utils.deflate_gzip
 
 local MAX_PAYLOAD = 4 * 1024 * 1024 -- 4MB
 local PING_INTERVAL = 30 -- 30 seconds
+local PING_WAIT = PING_INTERVAL * 1.5
 local WS_OPTS = {
   timeout = 5000,
   max_payload_len = MAX_PAYLOAD,
 }
+local ngx_OK = ngx.OK
 local ngx_ERR = ngx.ERR
+local ngx_WARN = ngx.WARN
 local ngx_DEBUG = ngx.DEBUG
 local ngx_INFO = ngx.INFO
+local ngx_NOTICE = ngx.NOTICE
 local WEAK_KEY_MT = { __mode = "k", }
 local CERT_DIGEST
 local CERT, CERT_KEY
@@ -99,6 +104,11 @@ local function update_config(config_table, update_cache)
 end
 
 
+local function is_timeout(err)
+  return err and string.sub(err, -7) == "timeout"
+end
+
+
 local function send_ping(c)
   local hash = declarative.get_current_hash()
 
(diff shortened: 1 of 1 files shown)


dndx:
Hold off on merging for a moment, pending sync up with @bungle.

kidd:
What's the status on this one?

bungle:
@kidd, well... I guess you forked our test suite at some point, but then have not synched the changes. So this issue is partly because of not synching your suite. The main culprit is here:
https://github.com/Kong/kong-ee/blob/next/2.2.x.x/spec-ee/02-integration/14-hybrid_mode/01-synch_spec.lua#L51-L59

That file was copy pasted from:
https://github.com/Kong/kong/blob/master/spec/02-integration/09-hybrid_mode/01-sync_spec.lua

See what we added there:
https://github.com/Kong/kong/blob/next/spec/02-integration/09-hybrid_mode/01-sync_spec.lua#L24

So if you want quick solution, just add that while we figure out what to do with this PR.

kidd:
Ah I see! I guess we created an ee-only test, but then if defaults changed we didn't notice it. 

Unfortunately, we don't have any way to sync at that level. 

We're going to add that today so we can keep merging in. Thanks for unblocking!

cc @mikefero

kong/clustering.lua
```diff
@@ -108,14 +118,10 @@ local function send_ping(c)
 
   local _, err = c:send_ping(hash)
   if err then
-    ngx_log(ngx_ERR, "unable to ping control plane node: ", err)
-    -- return and let the main thread handle the error
-    return nil
+    ngx_log(is_timeout(err) and ngx_NOTICE or ngx_WARN, "unable to ping control plane node: ", err)
```
dndx, line 121:
```suggestion
    ngx_log(is_timeout(err) and ngx_NOTICE or ngx_WARN, "unable to ping control plane node: ", err)
    return
```

kong/clustering.lua
```diff
@@ -304,68 +326,117 @@ function _M.handle_cp_websocket()
...
+      end
     end
+  end)
+
+  local _, err = ngx.thread.wait(main_thread, ping_thread)
+
+  ngx.thread.kill(main_thread)
+  ngx.thread.kill(ping_thread)
+
+  wb:send_close()
```
dndx, line 462:
```suggestion
  wb:close()
```

We should just close the WS connection here, if `send_close` hasn't been called, then the `close` method will call it automatically. Because the other end might be dead by now, and `send_close` may keep the (dead) connection around for longer than we need.

kong/clustering.lua
```diff
@@ -204,38 +206,58 @@ local function communicate(premature, conf)
...
-  while not exiting() do
-    local data, typ, err = c:recv_frame()
-    if err then
-      ngx.log(ngx.ERR, "error while receiving frame from control plane: ", err)
-      c:close()
+  local main_thread = ngx.thread.spawn(function()
+    while not exiting() do
+      local data, typ, err = c:recv_frame()
+      if err then
+        if not is_timeout(err) then
```
dndx, line 229: So, for the DP side of code, when `c:recv_frame` timedous, we should also be checking to see if this has been overdue compared to `PING_WAIT`, just like the CP did. These code can largely be duplicated from the CP side.

kong/clustering.lua
```diff
@@ -304,68 +326,117 @@ function _M.handle_cp_websocket()
...
-      local payload = table_remove(queue, 1)
-      assert(payload, "config queue can not be empty after semaphore returns")
-
-      if payload == "PONG" then
-        local _
-        _, err = wb:send_pong()
-        if err then
-          ngx_log(ngx_ERR, "failed to send PONG back to data plane: ", err)
-          return ngx_exit(ngx_ERR)
+  local main_thread = ngx.thread.spawn(function()
```
dndx, line 412: Maybe call this `send_thread` and the `ping_thread` -> `read_thread` instead, it may makes the code easier to follow for us.

dndx (commented):
Left some comments, also I pushed out a docs commit 84560666bbb18003dc8f549cdb896881086f149b so that the behavior is better documented for future reference.

dndx:
Might need to rebase onto `master` as well.

dndx (approved):
LGTM, let's make sure to confirm the status of RC1 release before merging.

PR #6418, opened 2020-10-02
fix(utils) typo in error message

kikito:
I believe we call these "hotfixes" since it fixes a feature introduced in the current release.


kikito:
#hacktoberfest

gszr (approved):

PR #6419, opened 2020-10-02
perf(utils) luajit table.clone and table.isarray

bungle:
### Summary

This PR comes with two commits:

#### perf(utils) use luajit's table.clone for shallow copying

OpenResty's fork of LuaJIT has `table.clone` library that we can use for shallow copying instead of implementing that on Lua code.

#### perf(utils) use luajit's table.isarray for array detection

OpenResty's fork of LuaJIT has `table.isarray` library that we can use for checking whether a table is array or not.

This commit also changes the special case of `Lapis` array to another function called `utils.is_lapis_array` because its implementation was rather opinionated and it might cause bugs elsewhere where this is not expected.


Tieske:
@javierguerragiraldez what is the status of clone? iirc we recently pulled that out of the balancer because there were issues with LuaJIT. Have those issues been fixed in the current OR version?

javierguerragiraldez:
@Tieske i think you're mixing with `table.move()` which is standard in Lua5.3, but had an issue in LuaJIT.  I haven't checked if the fix is in the openresty's fork we're pulling.

`table.clone()` is a different thing, it's not standard Lua or LuaJIT.  I'm looking for the PR where it was added, as I remember having some objections.  especially about `table.isarray()` since it implements explicitly wrong semantics.

bungle:
> @Tieske i think you're mixing with `table.move()` which is standard in Lua5.3, but had an issue in LuaJIT. I haven't checked if the fix is in the openresty's fork we're pulling.
> 
> `table.clone()` is a different thing, it's not standard Lua or LuaJIT. I'm looking for the PR where it was added, as I remember having some objections. especially about `table.isarray()` since it implements explicitly wrong semantics.
> 
> **Edit**: found it: [openresty/luajit2@5e4cf2b#r32252602](https://github.com/openresty/luajit2/commit/5e4cf2be2652b3639c79d50a0412b9ae7de6c68a#r32252602)

yes, i saw that two as i dropped my own is array function and saw that holes are allowed with luajits. still the previous implementation was even worse in that regard (check that is_lapis_array which was the previous).

javierguerragiraldez:
what about this (untested!)
```lua
function is_array(t)
  if type(t) ~= "table" then
    return false
  end
  
  local k = #t
  while k ~= nil do
    if type(k) == "number" then
      return false
    end
    k = next(t, k)
  end
  
  return true
end
```

if i got it right, it should reject holes, negative and non-integer keys, but allow string keys (so `{1, 2, 3, n=3}` is valid)  a nice thing, for arrays it's O(log n) and not O(n)

bungle:
> what about this (untested!)

This is what I had first:
```
    local floor = math.floor
    local max = math.max
    is_array = function(t)
      local m, c = 0, 0
      for k in pairs(t) do
          if type(k) ~= "number" or k < 1 or floor(k) ~= k then
            return false
          end
          m = max(m, k)
          c = c + 1
      end
      return c == m
    end
```

But then that was not compatible with luajit's `table.isarray`. The question remains, do we want this optimization or not. Will it bite us? Aren't arrays still arrays even with holes?

javierguerragiraldez:
> But then that was not compatible with luajit's `table.isarray`.

Neither Lua nor LuaJIT has an `isarray` function.  Personally, I consider OpenResty's one wrong.

> The question remains, do we want this optimization or not. Will it bite us? Aren't arrays still arrays even with holes?

`table.clone()` is easy to define (shallow copy) so the semantics aren't in dispute, and the performance difference should be significant if done on a hot path.

the "isarray" concept is open to interpretation.  I think holes and negative or non-integer numeric keys should make a table a "non-array", but other keys shouldn't.  Doing the check in C is faster but probably not by orders of magnitude.

of course, the useful thing is not to define an "ideal definition of array", but what check is needed in some specific code.   Only then you can tell which the best implementation.

bungle:
@javierguerragiraldez, I have updated this PR to now have `strict` semantics by default. One can call `utils.is_array(t, true)` to get the OpenResty optimized behavior. The test suite is enhanced to contain more test on this.

javierguerragiraldez (approved):
LGTM

i like that since there are different definitions of "array", there are different checks and not trying to believe one is "the right one".

PR #6421, opened 2020-10-02
fix(conf) default nginx_user to kong, not nobody

murillopaula:
If we install and run Kong with everything default, whenever Kong tries to write something to the Kong prefix, things will break and we will get a permission denied error. This applies to:

- DP nodes trying to write the config.json.gz file to the Kong prefix

- go pluginserver trying to write the go.lua file write to the Kong prefix

- Lua plugins trying to write files to the Kong prefix

And other cases. This is because, by default, the Kong prefix is owned by `root`. When we do a `apt-get install` or `yum install`, the `/usr/local/kong/` prefix (default) is owned by `root` by default. After that, if we do a `kong start` without changing our operating system's `$USER`, the nginx master process will be run as `root` and the nginx workers processes will be run as `nobody`. It means that if we do everything default, the nginx workers processes can’t write anything to the Kong prefix and we’ll get the permission denied errors.

There are a few immediate workarounds for this:

- add the `nginx_user` to the `root` group; or

- change the file system permissions to allow the `nginx_user` the ability to write the cache file to `/usr/local/kong`; or

- allow every user to write to `/usr/local/kong`

All of these workarounds are dangerous and hacky and we don’t document any of these as recommended fixes. We should fix the main cause of the issue. Here are the possibilities:

1) write files somewhere where any user has permissions, like `/tmp`. Maybe even create a `kong_prefix/tmp` folder where it's explicitly said that anyone has permissions over it;

2) set `nginx_user` to `kong` by default (instead of `nobody`), but drop to `nobody` if the `kong` user/group credentials are unavailble. This requires the non-root capability that was implemented at the packaging level [1]. The `nginx_user` configuration will only be changed for environments that have the `kong` user available, so we don't break the source based installs and don't change the default for developers;

3) pre-open all descriptors we need from the master process like nginx seem to do with its logs (e.g. error.log);

4) ship a `kong.conf` with `nginx_user` set to `kong` at the packaging level;

Option 1) is a hack, we’d need to repeat it everywhere we write files in the Kong prefix (Kong codebase, Lua plugins, Go pluginserver, etc). It is also dangerous. The worst option.

Option 2) is a general solution and the one we are going with. Since nginx worker processes will be owned by the `kong` user, and since the default prefix `/usr/local/kong` is now also owned by the `kong` user [1], things that try to write to the Kong prefix won't fail. User can still specify `nobody nobody` if they don't use Hybrid mode or Go plugins and the kong exists on their system and we do not introduce breakages for these users. This may be an obscure use case, but running Kong as nobody is not an uncommon use case. Thanks @dndx for suggesting improvements to this option.

Option 3) is more difficult. Currently there are technical issues that prevent it. We can pre-open all descriptors from the master process in the `init` phase, but it requires passing file descriptors around the codebase. In the `init` phase we have restricted access to some of the important data-sharing mechanisms between workers running in different phases. Also remember that it's not just hybrid mode that is broken, it's also Go plugins and custom Lua plugins that write things to the prefix. Go plugins do not have `init` phases.

Option 4) would mean essentially applying the patch [2] after package installation. Shipping a `/etc/kong/kong.conf.default` with `nginx_user` set to `kong` by default there and also overriding the environment variables. The problem is that the breaking change risk increases if people have `nginx_user` set to something other than the default `nobody`. This is difficult to get it right and it's a breaking change.

The idea of this PR is to fix the permission denied errors when running Kong with the default settings without causing breaking changes by implementing option 2). If the `kong` user does not exist, the user will be dropped to `nobody` as it was before. If the `nginx_user` configuration is set to something else other than `kong` (now the default), we won't change anything and the configuration value will be preserved. If it's set to `nobody` the behavior will be the same as before.

[1] Kong/kong-build-tools#339
[2] https://github.com/Kong/kong/pull/6422/files

Co-authored-by: Datong Sun <datong.sun@konghq.com>

Fix https://github.com/Kong/kong/issues/6591

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -11,11 +11,18 @@ local tablex = require "pl.tablex"
 local utils = require "kong.tools.utils"
 local log = require "kong.cmd.utils.log"
 local env = require "kong.cmd.utils.env"
+local ffi = require "ffi"
 local ip = require "resty.mediador.ip"
 
 
 local fmt = string.format
 local concat = table.concat
+local C = ffi.C
+
+ffi.cdef([[
+  struct group *getgrnam(const char *name);
+  struct passwd *getpwnam(const char *name);
+]])
 
 
 -- Version 5: https://wiki.mozilla.org/Security/Server_Side_TLS
@@ -1339,16 +1346,30 @@ local function load(path, custom_conf, opts)
 
   conf = tablex.merge(conf, defaults) -- intersection (remove extraneous properties)
 
+  local default_nginx_main_user = false
+  local default_nginx_user = false
+
   do
     -- nginx 'user' directive
     local user = utils.strip(conf.nginx_main_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_main_user = nil
+    elseif user == "kong" or user == "kong kong" then
+      default_nginx_main_user = true
     end
 
     local user = utils.strip(conf.nginx_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_user = nil
+    elseif user == "kong" or user == "kong kong" then
+      default_nginx_user = true
+    end
+  end
+
+  if C.getpwnam("kong") == nil or C.getgrnam("kong") == nil then
+    if default_nginx_main_user == true and default_nginx_user == true then
+      conf.nginx_user = nil
+      conf.nginx_main_user = nil
     end
   end
 
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -1362,6 +1362,7 @@ local function load(path, custom_conf, opts)
     local user = utils.strip(conf.nginx_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_user = nil
+
     elseif user == "kong" or user == "kong kong" then
       default_nginx_user = true
     end
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -1354,6 +1354,7 @@ local function load(path, custom_conf, opts)
     local user = utils.strip(conf.nginx_main_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_main_user = nil
+
     elseif user == "kong" or user == "kong kong" then
       default_nginx_main_user = true
     end
--- a/spec/01-unit/03-conf_loader_spec.lua
+++ b/spec/01-unit/03-conf_loader_spec.lua
@@ -3,6 +3,25 @@ local utils = require "kong.tools.utils"
 local helpers = require "spec.helpers"
 local tablex = require "pl.tablex"
 local pl_path = require "pl.path"
+local ffi = require "ffi"
+
+
+local C = ffi.C
+
+
+ffi.cdef([[
+  struct group *getgrnam(const char *name);
+  struct passwd *getpwnam(const char *name);
+]])
+
+
+local function kong_user_group_exists()
+  if C.getpwnam("kong") == nil or C.getgrnam("kong") == nil then
+    return false
+  else
+    return true
+  end
+end
 
 
 local function search_directive(tbl, directive_name, directive_value)
@@ -21,7 +40,11 @@ describe("Configuration loader", function()
   it("loads the defaults", function()
     local conf = assert(conf_loader())
     assert.is_string(conf.lua_package_path)
-    assert.is_nil(conf.nginx_main_user)
+    if kong_user_group_exists() == true then
+      assert.equal("kong kong", conf.nginx_main_user)
+    else
+      assert.is_nil(conf.nginx_main_user)
+    end
     assert.equal("auto", conf.nginx_main_worker_processes)
     assert.same({"127.0.0.1:8001 reuseport backlog=16384", "127.0.0.1:8444 http2 ssl reuseport backlog=16384"}, conf.admin_listen)
     assert.same({"0.0.0.0:8000 reuseport backlog=16384", "0.0.0.0:8443 http2 ssl reuseport backlog=16384"}, conf.proxy_listen)
@@ -36,7 +59,11 @@ describe("Configuration loader", function()
     -- defaults
     assert.equal("on", conf.nginx_main_daemon)
     -- overrides
-    assert.is_nil(conf.nginx_main_user)
+    if kong_user_group_exists() == true then
+      assert.equal("kong kong", conf.nginx_main_user)
+    else
+      assert.is_nil(conf.nginx_main_user)
+    end
     assert.equal("1", conf.nginx_main_worker_processes)
     assert.same({"127.0.0.1:9001"}, conf.admin_listen)
     assert.same({"0.0.0.0:9000", "0.0.0.0:9443 http2 ssl",
@@ -55,7 +82,11 @@ describe("Configuration loader", function()
     -- defaults
     assert.equal("on", conf.nginx_main_daemon)
     -- overrides
-    assert.is_nil(conf.nginx_main_user)
+    if kong_user_group_exists() == true then
+      assert.equal("kong kong", conf.nginx_main_user)
+    else
+      assert.is_nil(conf.nginx_main_user)
+    end
     assert.equal("auto", conf.nginx_main_worker_processes)
     assert.same({"127.0.0.1:9001"}, conf.admin_listen)
     assert.same({"0.0.0.0:9000", "0.0.0.0:9443 http2 ssl",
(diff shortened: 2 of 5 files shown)


murillopaula:
I think we can also do something like this: https://github.com/Kong/kong/pull/6422 but the drawback is that it's more aggressive (it assumes that the kong will always exist).

kong/conf_loader/init.lua
```diff
@@ -1339,19 +1339,35 @@ local function load(path, custom_conf, opts)
...
 
     local user = utils.strip(conf.nginx_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_user = nil
+      is_nobody = true
     end
   end
 
+  local kong_usr_exists = os.execute("getent passwd kong > /dev/null 2>&1")
+  local kong_grp_exists = os.execute("getent group kong > /dev/null 2>&1")
```
hutchic, line 1360: will this work universally ie alpine, mac, windows?

hishamhm:
Unless I'm reading it incorrectly, the behavior described in this PR is a breaking change with regard to how the default settings of `nginx_user` work, and thus, even if we decide that this would be the right thing to do, it could only be done in Kong 3.0. Is that the case or am I misreading the proposed behavior change?

murillopaula:
> Unless I'm reading it incorrectly, the behavior described in this PR is a breaking change with regard to how the default settings of nginx_user work, and thus, even if we decide that this would be the right thing to do, it could only be done in Kong 3.0. Is that the case or am I misreading the proposed behavior change?

Here in this PR the proposed behavior change is setting `nginx_user = kong kong` by default if and only if:

- the `kong` user and `kong` group exist;
- `nginx_user` is set to `nobody nobody` (the default today)

Goal is to fix the `Permission denied`s when running Kong with default settings.

I guess it's ok to wait for Kong 3.0 if this is considered to be a breaking change. But how would we fix the `Permission denied`s until Kong 3.0?

The PR https://github.com/Kong/kong/pull/6422 assumes that the `kong` user and group will always exist and doesn't care if `nginx_user` is set to something else than `nobody`, so that should be more of a breaking change, no?

kong/conf_loader/init.lua
murillopaula: Mac 😢 
```
~$ getent
bash: getent: command not found
```

Maybe this is more universal? 🦆 

```suggestion
  local kong_grp_exists = os.execute("id -u kong > /dev/null 2>&1")
```

Any other (universal) ways to check if a specific user and group exist?

murillopaula:
@hishamhm @hutchic @dndx any updates?

gszr:
@murillopaula We've created a task to look closer at this. Thanks for both proposals!

kong/conf_loader/init.lua
murillopaula: This won't work universally.

`getent` is not POSIX. It won't run on Mac, Windows, Alpine.

But we don't create the `kong` user/group by default on these systems, i.e., this patch does not apply for Mac, Windows and Alpine: https://github.com/Kong/kong-build-tools/pull/339

So it will stay as `nobody` in this case.

dndx: How about:

https://gist.github.com/dndx/b8b87f5abaac11ad89c89d8fd505a68f

This is POSIX and does not need any additional binary to do it.

dndx (commented):
I found myself having to do a bit of thinking to wrap around this. What if instead we:

1. Change the default user in `kong_defaults.lua` to `kong kong`.
2. In this file, check to see if the user is left at default `kong kong` and the user/group exists, if not we overwrite it to `nobody nobody`.

This logic probably makes a bit more sense, plus, user can still specify `nobody nobody` if they don't use Hybrid mode or Go plugins and the `kong` exists on their system and we do not introduce breakages for these users. Again, this may be an obscure use case, but running Kong as `nobody` is not an uncommon use case.

kong/conf_loader/init.lua
murillopaula: Great suggestion. Added you as a co-author and committed this change: https://github.com/Kong/kong/pull/6421/commits/95e913724e389a6f28d4a69f8c7dfcaec2c8c5f7

murillopaula:
Hi @dndx ready for another review.

Sorry about the force-pushes. I wanted to change the commit msg and rebase against `master`.

spec/01-unit/04-prefix_handler_spec.lua
```diff
@@ -293,7 +316,7 @@ describe("NGINX conf compiler", function()
         local nginx_conf = prefix_handler.compile_nginx_conf(conf)
         assert.not_matches("user%s+[^;]*;", nginx_conf)
       end)
-      it("is included when otherwise", function()
+      it("is included when otherwise #b", function()
```
dndx, line 319: Leftover from debugging?

kong/conf_loader/init.lua
```diff
@@ -1339,16 +1346,30 @@ local function load(path, custom_conf, opts)
...
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_main_user = nil
+    elseif user == "kong" or user == "kong kong" then
+      default_nginx_main_user = true
     end
 
     local user = utils.strip(conf.nginx_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_user = nil
+    elseif user == "kong" or user == "kong kong" then
```
dndx, line 1364:
```suggestion

    elseif user == "kong" or user == "kong kong" then
```

kong/conf_loader/init.lua
```diff
@@ -1339,16 +1346,30 @@ local function load(path, custom_conf, opts)
...
+  local default_nginx_main_user = false
+  local default_nginx_user = false
+
   do
     -- nginx 'user' directive
     local user = utils.strip(conf.nginx_main_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_main_user = nil
+    elseif user == "kong" or user == "kong kong" then
+      default_nginx_main_user = true
```
dndx, line 1358:
```suggestion

      default_nginx_main_user = true
```

spec/01-unit/03-conf_loader_spec.lua
```diff
@@ -427,6 +462,15 @@ describe("Configuration loader", function()
   end)
 
   describe("nginx_user", function()
+    it("is 'kong kong' by default if the kong user/group exist", function()
```
dndx, line 465: Just curious, is this even going to be triggered on CI, not sure if the CI even has the `kong` user/group.

Not a blocker though, more for my own curiosity.

dndx (approved):
Minor styling recommendations, but overall this change looks good. Please take a look and I will merge it once I hear back from you @murillopaula . Nice work!

kong/conf_loader/init.lua
```diff
@@ -1339,16 +1346,30 @@ local function load(path, custom_conf, opts)
...
 
+  local default_nginx_main_user = false
+  local default_nginx_user = false
+
   do
     -- nginx 'user' directive
     local user = utils.strip(conf.nginx_main_user):gsub("%s+", " ")
     if user == "nobody" or user == "nobody nobody" then
       conf.nginx_main_user = nil
+    elseif user == "kong" or user == "kong kong" then
```
murillopaula, line 1357:
```suggestion

    elseif user == "kong" or user == "kong kong" then
```

kong/conf_loader/init.lua
murillopaula: I assume you meant this? https://github.com/Kong/kong/pull/6421/files#r540225167

spec/01-unit/03-conf_loader_spec.lua
murillopaula: This is not going to be triggered on CI because the CI doesn't have the `kong` user available.

But when I was testing it locally the tests didn't pass because my machine _did_ have the `kong` user, so I thought I should add the verification of whether the `kong` user exist on the test suite too. Also good for readability, for your better understanding of this patch and if in the future someone sees it too.

spec/01-unit/04-prefix_handler_spec.lua
murillopaula:
```suggestion
      it("is included when otherwise", function()
```

murillopaula:
@dndx thanks for your help and for being the reviewer. When merging, please do a squash and merge.

spec/01-unit/04-prefix_handler_spec.lua
murillopaula: https://github.com/Kong/kong/pull/6421/commits/45a89d909f3655599504009c07d6849aea354c5f

kong/conf_loader/init.lua
murillopaula: https://github.com/Kong/kong/pull/6421/commits/fde903da180a7f5fb52442edac7e3ce1d7376339

kong/conf_loader/init.lua
murillopaula: https://github.com/Kong/kong/pull/6421/commits/f518f71e88cb05be1508043d22f91e8ee4271185

dndx (approved):

PR #6422, opened 2020-10-02
fix(conf) set nginx_user to 'kong kong' by default

murillopaula:
Second implementation attempt of https://github.com/Kong/kong/pull/6421


murillopaula:
doesn't work for source based install, doesn't work on Mac and not a good default for developers

last hope is #6421

PR #6429, opened 2020-10-03
Update README.md

Shivang0:
Made License  section using markdown and attractive .


hishamhm:
@Shivang0 Thanks but we prefer to have the license spelled out, with the appropriate copyright.

Issue #6426, opened 2020-10-03
[plugin]: go-plugin can't get kong.ctx.shared set by lua


two:
### Summary

SUMMARY_GOES_HERE

### Steps To Reproduce
lua plugin at `body_filter` phase:
```lua
local function status()
        local s = kong.response.get_status()
        kong.ctx.shared.status = s
end
```
go plugin at `log` phase:
```go
func (conf Config) Log(kong *pdk.PDK) {

        status, err := kong.Ctx.GetSharedInt("status")
        if err!= nil {
                kong.Log.Err(err.Error())
        }
       ....
}
```
got error : `null response`

### Additional Details & Logs

- Kong version (`2.1.3`)
- Kong debug-level startup logs (`$ kong start --vv`)
- Kong error logs (`<KONG_PREFIX>/logs/error.log`)
- Kong configuration (the output of a GET request to Kong's Admin port - see
  https://docs.konghq.com/latest/admin-api/#retrieve-node-information)
- Operating system

two:
can't get ctx only in go plugin
```go
package main

import (
        "github.com/Kong/go-pdk"
)

type Config struct{}

func New() interface{} {
        return &Config{}
}

func (conf Config) Access(kong *pdk.PDK) {
        err := kong.Ctx.SetShared("foo", 200)
        if err != nil {
                kong.Log.Err("access set shared error : ", err.Error())
        }
        return
}

func (conf Config) Log(kong *pdk.PDK) {
        status, err := kong.Ctx.GetSharedInt("foo")
        if err != nil {
                kong.Log.Err("GetSharedInt error:", err.Error())
        }
        kong.Log.Info("got status = ", status)

}
```

PR #6430, opened 2020-10-04
[go-plugin]: kong.ctx.shared can get in log phase #6426

two:
### Summary

now go-plugin can't get kong.ctx.shared set by other phase,
throw add shared in  log.serialize() return value got the shared value

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #6430


javierguerragiraldez:
Superseded by: #6490

PR #6431, opened 2020-10-04
test(plugins) add buffering combinations iterator

javierguerragiraldez:
Because we all want to spend 4x times on tests...

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -293,6 +293,30 @@ do
   end
 end
 
+local each_buffering
+do
+  local bufferedness = {
+    {true, true, "buffered"},
+    {true, false, "#unbuffered response"},
+    {false, true, "#unbuffered request"},
+    {false, false, "#unbuffered req/res"},
+  }
+
+  function each_buffering()
+    local i = 0
+
+    local function iter_unpack(t)
+      i = i + 1
+      local row = t[i]
+      if type(row) == "table" then
+        return unpack(row)
+      end
+    end
+
+    return iter_unpack, bufferedness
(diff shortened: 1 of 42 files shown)


kikito (approved):
This looks ok to me (assuming the tests pass).

Note for other reviewers: I strongly suggest reviewing this pr with the cog menu (⚙️▾) and then "✅Hide whitespace changes" option. Makes all the difference.

spec/helpers.lua
```diff
@@ -293,6 +293,30 @@ do
   end
 end
 
+local each_buffering
```
Tieske, line 296: documentation is missing here

Tieske (changes requested):
these are 2 additional layers of permutations for testing, can we not replace them with a single layer by using dynamic permutations?

for example;
```lua
local permute_bool = { true, false }

for strategy, buf_req, buf_res in permute({helpers.strategies, permute_bool, permute_bool }) do

  -- test stuff

end
```

Tieske:
I thought Penlight had a function for this, turned out not; https://github.com/lunarmodules/Penlight/pull/360

Tieske:
@javierguerragiraldez how to proceed? I can push a Penlight version with the permutations iterator if you want.

javierguerragiraldez:
> @javierguerragiraldez how to proceed? I can push a Penlight version with the permutations iterator if you want.

Thanks, no need to.  this PR won't be merged.  it's more an experiment to test that the new buffering cases don't crash anything catastrophically.

i'm closing this, as it has already proved the point.

PR #6441, opened 2020-10-07
chore(migrations) deprecate migration helpers

bungle:
### Summary

We have never documented this properly and we don't use it in any of the current
migrations. This was needed for pre `1.0.0` migrations.

With this deprecation, we have option to either:
1. remove it with `3.0.0`
2. if feedback demands that we need to keep it:
   then remove the warning and properly document it

### DIFF
--- a/kong/db/migrations/helpers.lua
+++ b/kong/db/migrations/helpers.lua
@@ -1,5 +1,6 @@
 --local json_decode = require("cjson.safe").decode
 local cassandra = require("cassandra")
+local log = require "kong.cmd.utils.log"
 --local utils = require "kong.tools.utils"
 
 
@@ -159,6 +160,11 @@ Note: In Cassandra, INSERT does "insert if not exists or update using pks if exi
 function _M:copy_cassandra_records(source_table_def,
                                    destination_table_def,
                                    columns_to_copy)
+
+  log.warn("migration helpers are deprecated: ",
+           "copy_cassandra_records function may not be available on a next major version")
+
+
   local coordinator, err = self.connector:connect_migrations()
   if not coordinator then
     return nil, err


Tieske:
what do we currently use when we need to update plugin properties in a migration?

I wrote some of that stuff along time ago to enable iterating over plugin configurations to update them, taking into account both datastores, paging, etc. That we lately haven't had the need for those doesn't mean it's useless. If we have a better way of doing that, then its fine imo.

bungle:
> what do we currently use when we need to update plugin properties in a migration?

There are several possibilities:

1. adding a new property

Our schema will fill the default if not in db, otherwise returns null (i am not 100% how it supports nested structures).

2. removing a property

Our schema will remove fields not in schema when reading db.

3. renaming property or moving it

You need to write a migration.

In general `3.` can only happens with `major` releases.

The helper this PR deprecates is meant to situation where Cassandra primary or clustering key changes.

Tieske:
the code I wrote was based on 3 (for cases where SQL/CQL fall short and we have to do it in Lua), and the problem with that is that it isn't easy to write it properly.

bungle:
> the code I wrote was based on 3 (for cases where SQL/CQL fall short and we have to do it in Lua), and the problem with that is that it isn't easy to write it properly.

The `3.` can also be handled in some cases in code, by keeping the old config there as well, and migrating by the plugin (if new specified use it, if not check if old, something that perhaps the read transformation can do for you). Of course not as neat in all the cases. Iterating over records and updating them is a constant fight for us with Cassandra. I have tried to fix one of such issues for the last two weeks now. As seen on this thread: https://github.com/Kong/kong/issues/6403 and all the prs created because of it.

kikito (approved):

PR #6449, opened 2020-10-08
feat(http-log) add custom header for authorization

Tieske:
Adds an option to provide a custom header, and value to add to the log request. Typically required for Splunk for example.

### DIFF
--- a/kong/plugins/http-log/handler.lua
+++ b/kong/plugins/http-log/handler.lua
@@ -2,6 +2,7 @@ local BatchQueue = require "kong.tools.batch_queue"
 local cjson = require "cjson"
 local url = require "socket.url"
 local http = require "resty.http"
+local table_clear = require "table.clear"
 
 
 local kong = kong
@@ -11,10 +12,15 @@ local tostring = tostring
 local tonumber = tonumber
 local concat = table.concat
 local fmt = string.format
+local pairs = pairs
 
 
 local queues = {} -- one queue per unique plugin config
 local parsed_urls_cache = {}
+local headers_cache = {}
+local params_cache = {
+  headers = headers_cache,
+}
 
 
 -- Parse host url.
@@ -76,20 +82,28 @@ local function send_payload(self, conf, payload)
     end
   end
 
-  local res, err = httpc:request({
-    method = method,
-    path = parsed_url.path,
-    query = parsed_url.query,
-    headers = {
-      ["Host"] = parsed_url.host,
-      ["Content-Type"] = content_type,
-      ["Content-Length"] = #payload,
-      ["Authorization"] = parsed_url.userinfo and (
-        "Basic " .. encode_base64(parsed_url.userinfo)
-      ),
-    },
-    body = payload,
-  })
+  table_clear(headers_cache)
+  if conf.headers then
+    for h, v in pairs(conf.headers) do
+      headers_cache[h] = v
+    end
+  end
+
+  headers_cache["Host"] = parsed_url.host
+  headers_cache["Content-Type"] = content_type
+  headers_cache["Content-Length"] = #payload
+  if parsed_url.userinfo then
+    headers_cache["Authorization"] = "Basic " .. encode_base64(parsed_url.userinfo)
+  end
+
+  params_cache.method = method
+  params_cache.path = parsed_url.path
+  params_cache.query = parsed_url.query
+  params_cache.body = payload
+
+  -- note: `httpc:request` makes a deep copy of `params_cache`, so it will be
+  -- fine to reuse the table here
+  local res, err = httpc:request(params_cache)
   if not res then
     return nil, "failed request to " .. host .. ":" .. tostring(port) .. ": " .. err
   end
--- a/kong/plugins/http-log/schema.lua
+++ b/kong/plugins/http-log/schema.lua
@@ -1,4 +1,5 @@
 local typedefs = require "kong.db.schema.typedefs"
+local url = require "socket.url"
 
 return {
   name = "http-log",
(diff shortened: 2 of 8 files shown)


kong/plugins/http-log/schema.lua
```diff
@@ -16,6 +17,15 @@ return {
           { retry_count = { type = "integer", default = 10 }, },
           { queue_size = { type = "integer", default = 1 }, },
           { flush_timeout = { type = "number", default = 2 }, },
-    }, }, },
+          { header_name = typedefs.header_name({ default = "Authorization"}), },
```
dndx, line 20: Do we really need a default for this config? It seems to be too specific to be included.

Also, is a single header good enough? Is there any reason people may want to specify more than one header?

Tieske: "Authorization" is the common one, hence the default. It remains unused unless a value is also provided.

For now a single header should do, afaik there is only 1 service that needs it (at least users want supported).

Let me know if you want me to change aything anyway

dndx: I guess there no immediate scenario where setting more than one statically defined headers are needed, so let's merge this for now.

kong/plugins/http-log/schema.lua
```diff
@@ -16,6 +17,15 @@ return {
...
+          { header_value = { type = "string" }, },
+        },
+        entity_checks = {
+          -- header_name is required when header value is given
+          { conditional = {
+            if_field = "header_value", if_match = { ne = null },
+            then_field = "header_name", then_match = { required = true },
+          }, }, },
```
dndx, line 28: Should we check that when `header_name` is given, `header_value` is also present? Otherwise user may mis configure the plugin without knowing and it is a silent failure.

Tieske: how would a user miss configure it if he does not set a header value? If he doesn't set either he'd get the same silent failure...

dndx: But it seems that header name and value should always be set in pairs, so the fact that we are checking only one way makes little sense. Are we saying it is more likely for the user to forget setting header names but not the other way around?

Tieske: because the headername defaults to "Authorization", hence no check required

Tieske: and if you provide a value, then the name becomes required

kong/plugins/http-log/handler.lua
```diff
@@ -76,18 +76,23 @@ local function send_payload(self, conf, payload)
     end
   end
 
+  local headers = {
+    ["Host"] = parsed_url.host,
+    ["Content-Type"] = content_type,
+    ["Content-Length"] = #payload,
+    ["Authorization"] = parsed_url.userinfo and (
+      "Basic " .. encode_base64(parsed_url.userinfo)
```
bungle, line 84: This is that I don’t fylly understand. It lets you set header name but it forcibly prefixes the value with Basic.

bungle:
Should it allow plural? Aka custom headers?

dndx:
@bungle I asked the same question in https://github.com/Kong/kong/pull/6449#discussion_r524421841. I guess there is little use case for people adding two statistically defined headers, so I tends to agree this current implementation should satisfies more than 90% of the use cases out there.

In particular, for authenticating against a log collection API, one set of header is usually enough. Whether it is `Authorization` or `API-Key` or something similar.

If we support templating for header values then supporting more than one custom header may be more useful.

Tieske:
@dndx @bungle if using this; https://github.com/Kong/kong/blob/next/kong/db/schema/typedefs.lua#L509

do you have an example of where/how it is to be used?

Tieske:
had a look at it, that is not suitable. That headers typedef only checks for "Host", and doesn't allow it.
To do the same here, we need to validate the other headers as well, see https://github.com/Kong/kong/blob/612c365e5027695318ba40cf436754452879a28b/kong/plugins/http-log/handler.lua#L85-L87

That is really overkill for such a simple feature.

Tieske:
* the headers one is not suited, it contains an error message specific to the `route` entity
* the keys in the headers typedef are just strings, not valid header names. currently there is 1 header name in this PR, but it is properly validated
* if checking for Host, we also must check for the others, so that would require a complete new typedef anyway.
* If we check for `Authorization` there must be a check that it does not collide with the userinfo in the URL provided, which I think requires a custom validator function, which we don't like.

I really think this is overcomplicating it.

Tieske:
@bungle @dndx fixed up to support multiple headers now.

dndx (approved):

kong/plugins/http-log/handler.lua
```diff
@@ -76,18 +79,25 @@ local function send_payload(self, conf, payload)
     end
   end
 
+  table_clear(headers_cache)
```
Tieske, line 82: @dndx I don't think this is correct.

Below the `httpc` makes a request, which involves connecting and sending socket operations. If that blocks for a while waiting for sockets to become available, then a follow up request might clear the table, before it was send. and hence break the whole thing. Nasty race condition.

dndx: That's a good observation. However, `resty.http` always copies the params table when calling `:request`:

https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L781

Because later it needs to change the meltable on the `params` table https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L605 so it is not a good idea to reuse it.

bungle: Slightly related: https://github.com/ledgetech/lua-resty-http/issues/103

PR #6452, opened 2020-10-08
chore(arm) build workers

hutchic:
The arm build workers require randomized names so the ssh keys dont collide

The only ENV I added was `DOCKER_MACHINE_ARM64_NAME="jenkins-kong-"`cat /proc/sys/kernel/random/uuid`` the rest was a copy / paste to move some values

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -152,11 +152,15 @@ pipeline {
                 CACHE = false
                 UPDATE_CACHE = true
                 DEBUG = 0
+                RELEASE_DOCKER_ONLY=true
+                PACKAGE_TYPE=apk
+                RESTY_IMAGE_BASE=alpine
+                RESTY_IMAGE_TAG=latest
             }
             steps {
                 sh 'echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin || true'
                 sh 'make setup-kong-build-tools'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -147,12 +147,20 @@ pipeline {
...
+                RELEASE_DOCKER_ONLY=true
+                PACKAGE_TYPE=apk
+                RESTY_IMAGE_BASE=alpine
+                RESTY_IMAGE_TAG=latest
             }
             steps {
                 sh 'echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin || true'
                 sh 'make setup-kong-build-tools'
-                sh 'KONG_VERSION=`git rev-parse --short HEAD` RELEASE_DOCKER_ONLY=true PACKAGE_TYPE=apk RESTY_IMAGE_BASE=alpine RESTY_IMAGE_TAG=latest make release'
+                sh 'KONG_VERSION=`git rev-parse --short HEAD` DOCKER_MACHINE_ARM64_NAME="jenkins-kong-"`cat /proc/sys/kernel/random/uuid` make release'
```
gszr, line 163: Could the `docker-machine rm` command be failing and causing the dangling keys?

gszr (approved):

Jenkinsfile
hutchic: Yes. There's a variety of reasons
- humans testing locally use the same name
- builds fail / timeout / don't do their cleanup
- temporary credentials fail and cleaning up isn't possible.

For these reasons we do have a lambda that nukes ec2's / instances after 3 hours but yeah there's enough churn we need unique names

gszr: Got it. Did not know about the lambda! Nice!

PR #6459, opened 2020-10-12
fix(go) log-preload data was using module-level scope

javierguerragiraldez:
Go plugins run the `Log` phase under a timer, which can happen after
the request lifetime, so result of the `kong.log.serialize()` is
preloaded into a module-level table.  Two different Go plugins or two
concurrent requests on the same worker could interfere with each other.

This commit stores this request-specific information using the coroutine
of the timer itself as a key.

Fix: [FTI-1916](https://konghq.atlassian.net/browse/FTI-1916)


gszr (approved):

dndx:
Is it possible to test this change? Not a complete blocker as I believe the fix indeed solves the problem, but would be nice to prevent future changes from breaking it.

PR #6465, opened 2020-10-13
hotfix(admin) add /clustering/status back with deprecation header

bungle:
### Summary

Adds /clustering/status back for backward compatibility with a `Deprecation` header as described in https://tools.ietf.org/html/draft-dalal-deprecation-header-03.

### DIFF
--- a/kong/api/routes/clustering.lua
+++ b/kong/api/routes/clustering.lua
@@ -1,6 +1,9 @@
 local endpoints = require "kong.api.endpoints"
 
 
+local kong = kong
+
+
 local dp_collection_endpoint = endpoints.get_collection_endpoint(kong.db.clustering_data_planes.schema)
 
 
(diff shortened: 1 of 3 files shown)


hutchic (commented):
> changes the admin api endpoint /clustering/data_planes to a more url friendly /clustering/data-planes

That sounds like a breaking change unless we're leaving `/clustering/data-planes` but deprecating it?

bungle:
> > changes the admin api endpoint /clustering/data_planes to a more url friendly /clustering/data-planes
> 
> That sounds like a breaking change unless we're leaving `/clustering/data-planes` but deprecating it?

Well, we haven't released `/clustering/data_planes` in any previous release, so there is not much to break. Thus making this change before we release. Now of course we start arguments whether beta was a line of no change... but well... it feels like it could only be us that is building features on top of that, and yet we got report internally that we need to add old endpoint back, so I am faily certain, nobody has ever used it anywhere.

hutchic:
not my decision but that sounds reasonable 👍 

I err'd and thought this was changing already released functionality

bungle:
This PR i now split into two. This PR here is just for adding the endpoint back. And other PR #6468 does the rename.

kong/api/routes/clustering.lua
```diff
@@ -20,4 +23,36 @@ return {
...
+            return kong.response.exit(500, { message = "An unexpected error happened" })
+          end
+
+          local id = row.id
+          row.id = nil
+
+          data[id] = row
+        end
+
+        return kong.response.exit(200, data, {
```
dndx, line 52: Maybe we can also do

`data["message"] = "This endpoint has been deprecated and will be removed in a future release, please use \"/clustering/data_planes\" instead."` because it seems that the header can be easily missed even when using a CLI based client (such as cURL).

bungle: I know, but this would break the API contract.

dndx (approved):

PR #6470, opened 2020-10-14
feat(conf) auto-generate and inject dhparams for default intermediate cipher suite

### LINKED ISSUE #6454: Allow to specify ssl_dhparam in kong.conf
### Summary
Background: We had ssl_cipher_suite = intermediate enabled together with an RSA key. This setup caused problem with old Win7 machines which did not support any of the ciphers advertised by our server. After a bit of debugging we found that 'intermediate' actually includes Win7 compatible DHE protocols (e.g. TLS_DHE_RSA_WITH_AES_256_GCM_SHA384). But these where not advertised by our server. The solution was to generate some dh parameters and config them with `nginx_proxy_ssl_dhparam` to enable these protocols.

Think it would be useful to have a ssl_dhparam entry in the default kong.conf so that it is easier to discover that this might be an option required to enable DHE protocols.

bungle:
### Summary

This work is a rework of the PR #4688. Perhaps this time we can include it, and it is modified according to feedback received back then.

This PR changes the `cipher_suite=intermediate` to also add `ssl_dhparam` directive in our template. For `modern` this does not do it as there are no ciphers that need it. Also for `old` it does not do it as Mozilla's old recommends `1024` bits params, and there are no secure predefined one found in: RFC 7919: https://tools.ietf.org/html/rfc7919#appendix-A.

Also it does not generate it for `custom` cipher suites, but allows custom suites to specify shorthands:

- `ffdhe2048`
- `ffdhe3072`
- `ffdhe4096`
- `ffdhe6144`
- `ffdhe8192`

(there are some others by `resty.openssl` but not worth to advertise).

```
KONG_SSL_DHPARAM=<shorthand>
```

Note the above only works with `old` and `custom` (`modern` removes this and `intermediate` forcibly sets it to `ffdhe2048`).

### Issues Resolved

Fix #6454


kong/cmd/utils/prefix_handler.lua
```diff
@@ -19,6 +19,63 @@ local ffi = require "ffi"
...
+    if pem then
+      local ssl_path = pl_path.join(kong_config.prefix, "ssl")
+      if not pl_path.exists(ssl_path) then
+        local ok, err = pl_dir.makepath(ssl_path)
+        if not ok then
+          return nil, err
+        end
+      end
+
+      if pem then
```
fffonion, line 63: This branch seems unnecessary as it's already in a `if pem` branch

fffonion (approved):

kong.conf.default
```diff
@@ -500,6 +500,18 @@
                                  #
                                  # See http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_prefer_server_ciphers
 
+#ssl_dhparam =                   # Defines DH parameters for DHE ciphers from the
+                                 # predefined groups: `ffdhe2048`, `ffdhe3072`,
+                                 # `ffdhe4096`, `ffdhe6144`, `ffdhe8192`, or
+                                 # from the absolute path to a parameters file.
+                                 #
```
kikito, line 507: This comment doesn't make it clear to me what values the `ssl_dhparam` conf option can actually take. Is it one of [`ffdhe2048`, `ffdhe3072`,`ffdhe4096`, `ffdhe6144`, `ffdhe8192`], or is it something else?

> or from the absolute path to a parameters file

This seems to indicate that this config option can be set to a filepath. If that's the case, we should clarify what syntax and possible options to be put in that file.

The nginx link below does not clarify any of this either.

kikito (changes requested):

kong.conf.default
bungle: yes, the value can be ffdhe2048, ffdhe3072, ffdhe4096, ffdhe6144, or ffdhe8192 (or actually any predefined group lua-resty-openssl/openssl knows about, but others are not worth to document) OR it can point to a file.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -414,5 +469,6 @@ return {
   compile_kong_conf = compile_kong_conf,
   compile_kong_stream_conf = compile_kong_stream_conf,
   compile_nginx_conf = compile_nginx_conf,
-  gen_default_ssl_cert = gen_default_ssl_cert
+  gen_default_ssl_cert = gen_default_ssl_cert,
+  gen_default_dhparams = gen_default_dhparams,
```
dndx, line 473: It seems that `gen_default_dhparams` is only called from this file (and is most likely only useful in this file), is there a reason for exporting it?

kong/conf_loader/init.lua
```diff
@@ -1533,6 +1568,30 @@ local function load(path, custom_conf, opts)
...
+                   or conf.stream_proxy_ssl_enabled
+                   or conf.admin_ssl_enabled
+                   or conf.status_ssl_enabled
+
+  for _, name in ipairs({ "nginx_http_directives", "nginx_stream_directives" }) do
+    for i, directive in ipairs(conf[name]) do
+      if directive.name == "ssl_dhparam" then
+        if is_predefined_dhgroup(directive.value) then
+          if ssl_enabled then
+            directive.value = pl_path.abspath(pl_path.join(conf.prefix, "ssl", directive.value .. ".pem"))
```
dndx, line 1581:
```suggestion
            directive.value = pl_path.abspath(pl_path.join(conf.prefix, "ssl", directive.value .. ".pem"))

```

spec/01-unit/03-conf_loader_spec.lua
```diff
@@ -824,6 +824,36 @@ describe("Configuration loader", function()
...
+          assert.equal(nil, conf.nginx_stream_ssl_dhparam)
+        end)
+        it("doesn't define ssl_dhparam with old cipher suite (#todo)", function()
+          local conf, err = conf_loader(nil, {
+            ssl_cipher_suite = "old",
+          })
+          assert.is_nil(err)
+          assert.equal(nil, conf.nginx_http_ssl_dhparam)
+          assert.equal(nil, conf.nginx_stream_ssl_dhparam)
+        end)
```
dndx, line 856: We can also add a check for file's existence and the `abspath` transformation like https://github.com/Kong/kong/pull/6470/files#diff-ecda8b9e752563281567542698215736bfc851121ddd9cca53621f1ee8cd755bR882.

The file does not need to point to real dhparams, we only care about the config string being correctly expanded.

spec/01-unit/04-prefix_handler_spec.lua
```diff
@@ -759,6 +759,7 @@ describe("NGINX conf compiler", function()
           prefix = tmp_config.prefix,
           proxy_listen = "127.0.0.1:8000 ssl",
           admin_listen = "127.0.0.1:8001 ssl",
+          ssl_cipher_suite = "custom",
```
dndx, line 762: Maybe we can add a test like  https://github.com/Kong/kong/pull/6470/files#diff-1ec0492b7c203c5f8e7924d1451c236131217a17006fd33beacbb680871fe803R772. By setting `proxy_listen  = "127.0.0.1:8000 ssl",` and `nginx_http_ssl_dhparam` to a predefined group, the file should be generated and the `ssl_dhparam` directive should points to the correct file.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -345,6 +392,14 @@ local function prepare_prefix(kong_config, nginx_custom_template_path)
     nginx_template = pl_file.read(nginx_custom_template_path)
   end
 
+  if kong_config.proxy_ssl_enabled
+  or kong_config.stream_proxy_ssl_enabled
+  or kong_config.admin_ssl_enabled
+  or kong_config.status_ssl_enabled
```
dndx, line 398:
```suggestion
  if kong_config.proxy_ssl_enabled
     or kong_config.stream_proxy_ssl_enabled
     or kong_config.admin_ssl_enabled
     or kong_config.status_ssl_enabled
```

kong/conf_loader/init.lua
```diff
@@ -1533,6 +1577,31 @@ local function load(path, custom_conf, opts)
     conf.admin_ssl_cert_key = pl_path.abspath(conf.admin_ssl_cert_key)
   end
 
+  local ssl_enabled = conf.proxy_ssl_enabled
+                   or conf.stream_proxy_ssl_enabled
+                   or conf.admin_ssl_enabled
+                   or conf.status_ssl_enabled
```
dndx, line 1583:
```suggestion
  local ssl_enabled = conf.proxy_ssl_enabled
                      or conf.stream_proxy_ssl_enabled
                      or conf.admin_ssl_enabled
                      or conf.status_ssl_enabled
```

dndx (approved):

kikito (approved):
lgtm

PR #6438, opened 2020-10-06
docs(loadbalancer) target objects can now be PATCHed and DELETEd

javierguerragiraldez:
2.2 now allows updating and removing targets via the admin-api. Remove notes about previous caveats and document new method.

PR #6473, opened 2020-10-14
fix(api) use the standard status when patching targets

locao:
`PATCH` requests to Kong admin API return `200` and the entity as the response body when succeeded. This change fixes the behavior for target entities to the standard, instead of returning `204`.

### DIFF
--- a/spec/02-integration/04-admin_api/08-targets_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/08-targets_routes_spec.lua
@@ -753,14 +753,18 @@ describe("Admin API #" .. strategy, function()
           },
           headers = { ["Content-Type"] = "application/json" }
         })
-        assert.response(res).has.status(204)
+        assert.response(res).has.status(200)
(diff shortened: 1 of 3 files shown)


gszr (approved):
Good to go once green 💚

gszr:
Reminder we also need to review / adapt https://github.com/Kong/kong/pull/6438

spec/02-integration/04-admin_api/08-targets_routes_spec.lua
```diff
@@ -753,7 +753,7 @@ describe("Admin API #" .. strategy, function()
           },
           headers = { ["Content-Type"] = "application/json" }
         })
-        assert.response(res).has.status(204)
+        assert.response(res).has.status(200)
```
dndx, line 756: In addition to the status code, should we check for the JSON response as well to make sure it contains a valid entity?

dndx (approved):

PR #6483, opened 2020-10-16
fix(dao) avoid insertion of duplicate targets

locao:


kong/db/dao/targets.lua
```diff
@@ -46,6 +47,15 @@ function _TARGETS:insert(entity, options)
     entity.target = formatted_target
   end
 
+  local workspace = workspaces.get_workspace_id()
+  local opts = { nulls = true, workspace = workspace }
+  for existent in self:each_for_upstream(entity.upstream, nil, opts) do
+    if existent.target == entity.target then
```
dndx, line 53: Might be missing the obvious, but why can't we add the `unique = true` property to the `target` field of the `kong/db/schema/entities/targets.lua` file instead?

locao: We can have the same target added to different upstreams, wouldn't the `unique = true` stop us from doing that?

bungle: whatabout adding cache_key?

kong/api/routes/upstreams.lua
```diff
@@ -166,6 +180,16 @@ return {
...
                                             "page_for_upstream"),
     POST = function(self, db)
+      -- updating a target using POST is a compatibility with existent API and
+      -- should be deprecated in next major version
+      local entity, _, err_t = update_existent_target(self, db)
+      if err_t then
+        return endpoints.handle_error(err_t)
+      end
+      if entity then
+        return kong.response.exit(200, entity)
```
dndx, line 190: Maybe we should add a `Deprecation` header like https://github.com/Kong/kong/pull/6465/files#diff-4d5b2a7046fe9022c548c4212bfc5ae9596c21237771d0e628a319fab2597785R53.

kong/api/routes/upstreams.lua
```diff
@@ -112,6 +112,20 @@ local function target_endpoint(self, db, callback)
...
+local function update_existent_target(self, db)
+  local upstream = endpoints.select_entity(self, db, db.upstreams.schema)
+  local filter = { target = unescape_uri(self.params.target) }
+  local opts = endpoints.extract_options(self.args.uri, db.targets.schema, "select")
+  local target = db.targets:select_by_upstream_filter(upstream, filter, opts)
+
+  if target then
+    self.params.targets = db.targets.schema:extract_pk_values(target)
+    return endpoints.update_entity(self, db, db.targets.schema)
+  end
```
dndx, line 124:
```suggestion
  end

```

kong/db/dao/targets.lua
```diff
@@ -46,6 +47,15 @@ function _TARGETS:insert(entity, options)
...
+      local err_t = self.errors:unique_violation({ target = existent.target })
```
dndx, line 54: Does this need to be changed to update the existing `target` instead of returning an error as well? Will this be a breaking change?

kong/api/routes/upstreams.lua
locao: Nice catch, thanks! I jusrt changed it.

spec/02-integration/04-admin_api/08-targets_routes_spec.lua
```diff
@@ -113,6 +113,42 @@ describe("Admin API #" .. strategy, function()
...
+            body = {
+              target = "single-target.test:8080",
+              weight = 100,
+            },
+            headers = {["Content-Type"] = content_type}
+          })
+          local body = assert.response(res).has.status(200)
+          local json = cjson.decode(body)
+          assert.are.equal(100, json.weight)
+          assert.are.equal(id, json.id)
```
dndx, line 148:
```suggestion
          assert.are.equal(id, json.id)
          assert.equal("true", res.headers["Deprecation"])
```

kong/db/dao/targets.lua
locao: As we talked privately: the admin API first tries to update the entity before inserting it, the behavior for end-users does not change at all, only for DAO users. I think this does not need to be considered a breaking change.

kong/db/dao/targets.lua
dndx: DIscussed with @bungle and @locao, we decide to postpone this change into a later release because of the complexity of adding `cache_key` with the workspace in consideration. It will not be able to fit in the release schedule.

This PR still made the situation better, so there is no reason to require the `cache_key` change right here right now.

PR #6489, opened 2020-10-19
fix(dao) no_{consumer,route,service} usage

gszr:
Schemas are not associative arrays, indexed by the field name,
but arrays of fields; therefore, one needs to iterate over the
top-level `fields` to check for the presence of specific fields.

### DIFF
--- a/kong/db/dao/plugins.lua
+++ b/kong/db/dao/plugins.lua
@@ -234,14 +234,18 @@ local function load_plugin(self, plugin)
     return nil, err
   end
 
-  if schema.fields.consumer and schema.fields.consumer.eq == null then
-    plugin.no_consumer = true
-  end
-  if schema.fields.route and schema.fields.route.eq == null then
-    plugin.no_route = true
-  end
-  if schema.fields.service and schema.fields.service.eq == null then
-    plugin.no_service = true
+  for _, field in ipairs(schema.fields) do
+    if field.consumer and field.consumer.eq == null then
+      handler.no_consumer = true
+    end
+
+    if field.route and field.route.eq == null then
+      handler.no_route = true
+    end
+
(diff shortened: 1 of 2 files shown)


locao (approved):
LGTM, let's just wait for it to be green before merging.

kong/db/dao/plugins.lua
```diff
@@ -234,14 +234,18 @@ local function load_plugin(self, plugin)
...
-    plugin.no_consumer = true
-  end
-  if schema.fields.route and schema.fields.route.eq == null then
-    plugin.no_route = true
-  end
-  if schema.fields.service and schema.fields.service.eq == null then
-    plugin.no_service = true
+  for _, field in ipairs(schema.fields) do
+    if field.consumer and field.consumer.eq == null then
+      handler.no_consumer = true
```
dndx, line 239: Is this intentionally changed to `handler.no_consumer` from `plugin.no_consumer`?

dndx (approved):
This LGTM aside from a curious question above.

kong/db/dao/plugins.lua
gszr: Good question. Yes, note that the function returns `handler` rather than `plugin`, and that is what is used in places where the function is called.

dndx: Yes, but it seems that in https://github.com/Kong/kong/blob/master/kong/runloop/plugins_iterator.lua#L134-L142 where the values are used, it is checking the `plugin` object instead of the handler.

gszr: That's very right. Although `plugin` in here in the DAO is the plugin name, `plugin` over in the plugins iterator is an object with a `handler` field. Updated to fix it. Good catch!

PR #6499, opened 2020-10-21
chore(docs) drop "development", no longer relevant

Tieske:


kong.conf.default
```diff
@@ -1208,7 +1208,7 @@
                                  # to each individual worker.
 
 #------------------------------------------------------------------------------
-# DEVELOPMENT & MISCELLANEOUS
```
dndx, line 1211: On this change, would `lua_package_path` and `lua_package_cpath` still constitutes "development" configs?

Tieske: Not imo, those are primarily properties for deploying custom plug-ins. That's not development.

bungle: What about naming this section just `LUA` as everything is contains is `lua_` after all.

Tieske: Lua is not a helpful heading to end users, I think Miscellaneous covers it fine

bungle: My only comment is that if it was miscellaneous then it would contain miscellaneous stuff, but now it only contains these: https://github.com/openresty/lua-nginx-module#directives (so it is not a bunch of unrelated stuff, but very well defined set of settings for a single dependency). Yes, you can say that because of lua module touches all our code base, and contains a lot of settings, then well settings are miscellaneous. I am not against it. I have neutral feeling about it. It is just an PR that I am not sure helps anyone, just adds this useless conversation. Someone else who thinks it is good can approve and merge, and I am fine with it.

bungle (approved):

PR #6351, opened 2020-09-16
Add/change test helpers to facilitate db-less plugin testing

Tieske:
Here's an example of a test file adjusted to also run db-less tests, using the helpers from this PR:
https://github.com/Kong/kong-plugin/compare/dbless?expand=1

The 2nd commit in this PR fixes some ldoc related issues in the helpers file, please review commits one-by-one

BTW; this should land in 2.2

ping @mikefero @kidd

PR #6501, opened 2020-10-21
refactor(helpers) combine 2 functions

Tieske:
follow up to https://github.com/Kong/kong/pull/6351#issuecomment-696800128

The 2 functions should have been combined from the start. Since these are only helpers and the additional function in #6351 hasn't yet made it into a release, I think this PR should be in `2.2` still, to keep it clean and not release a helper that gets deprecated right after.

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -56,6 +56,9 @@ ffi.cdef [[
 ]]
 
 
+local kong_exec   -- forward declaration
+
+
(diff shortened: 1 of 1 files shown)


dndx (approved):

dndx:
Since there are no loss of functionality after this PR, let's merge it for now. If any issue arises we can certainly resolve/revert this PR in the future to fix it.

PR #6503, opened 2020-10-21
feat(go) grab log from pluginserver's stderr

javierguerragiraldez:
Allows Go plugins to just use the idiomatic `log.Printf(...)`.

Any stdout/stderr output will be copied into nginx's `error.log` line by line.

### DIFF
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -3,6 +3,7 @@ local ngx_ssl = require("ngx.ssl")
 local basic_serializer = require "kong.plugins.log-serializers.basic"
 local msgpack = require "MessagePack"
 local reports = require "kong.reports"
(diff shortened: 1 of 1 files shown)


kong/db/dao/plugins/go.lua
```diff
@@ -62,6 +63,20 @@ do
     return out:match("Runtime Version: go(.+)\n$")
   end
 
+  local function grab_logs(proc)
+    while true do
+      local data, err, partial = proc:stdout_read_line()
+      local line = data or partial
+      if line and line ~= "" then
+        errlog.raw_log(ngx.INFO, "[go-pluginserver] " .. line)
```
dndx, line 71: As this could potentially be a hot path, maybe it makes sense to cache `errlog.raw_log` and `ngx.INFO` at module level.

dndx (approved):

PR #6517, opened 2020-10-26
fix(router) ignore SNI matcher when request `protocol` is not `https`…

### LINKED ISSUE #6425: Dual stack HTTP+HTTPS routes cannot contain SNI match criteria while still honoring their protocols criteria
### Summary

Given a route that supports both HTTP and HTTPS requests, and has both hostname and SNI match criteria (for the same hostname), an HTTP request with the correct hostname will not match the given route. There is no way for _any_ HTTP request to match that route, because plaintext HTTP requests will never contain SNI because they have no TLS envelope by definition.

### Steps To Reproduce

1. Create a service.
2. Create two routes, one with hostname and SNI criteria for `sni.kong.example`, and one with hostname criteria only, for `nosni.kong.example`. As no protocol set is explicitly specified, these routes will default to `protocols=http,https`
3. Send an HTTP request to the first route, with the correct hostname, e.g. `curl -sv localhost:8000/anything -H "Host: sni.kong.example"`. Kong will return a 404.
4. Send an HTTP request to the second route, e.g. `curl -sv localhost:8000/anything -H "Host: nosni.kong.example"`. The route will proxy normally, and return a 200 (or whatever status the upstream service normally returns).

[sni-404-example.yaml.txt](https://github.com/Kong/kong/files/5320511/sni-404-example.yaml.txt) is a deck dump that covers the above. Github is silly and forced me to append the `.txt` extension.

Note that this also applies to HTTPS-only routes with an HTTPS redirect status code: once you add SNI criteria, those routes will no longer redirect to HTTPS when receiving an otherwise-matching HTTP request, and clients sending those requests will instead receive a 404.

### Additional Details & Logs

Tested on kong/2.2.0beta.1, Alpine Docker. Should affect any Kong version/platform that supports SNI match criteria.

### Why this matters

See previous internal discussion in FTI-1881 and comments on https://github.com/Kong/kubernetes-ingress-controller/pull/863. On the controller side, we would like to handle adding SNI criteria for users automatically, as we can reasonably assume that Ingresses with a hostname rule and TLS information will be HTTPS-accessible, and will receive HTTPS requests with SNI information that have a matching HTTP hostname. There are some caveats to this, but I won't re-tread them here.

However, many of those rules will also be dual-stack, i.e. they will allow HTTP traffic alongside HTTPS traffic. As-is, adding SNI criteria to these routes forces them to single-stack HTTPS, because adding SNI criteria to a route means _any_ request that wants to match that route must include a TLS client handshake with SNI, even when that is impossible. This also comes up in the context of SNI-incapable clients, which will do TLS when using HTTPS, but will not use SNI.

In the context of FTI-1881 (and the previous requests that spurred development of the current mTLS behavior), this doesn't matter for the routes that actually have an mTLS plugin: if you require client certificate authentication to access a route, that route _must_ be single-stack anyway--even assuming this issue didn't exist, clients would not be able to successfully authenticate over HTTP. This _does_ matter for all the routes that do not have an mTLS plugin, where we want to avoid prompting for a client certificate because we know that route does not have an mTLS plugin, and we know the route based on SNI information. Routes in the latter category can quite reasonably also allow HTTP traffic, so as-is we force users into a choice of "accept the certificate prompts for HTTPS, which are confusing for your end users" or "don't allow HTTP traffic to these routes at all".

There's arguably a third way where we split dual-stack routes into separate HTTPS and HTTP routes for the same service, but that's cumbersome and brings in a whole host of other issues (wanna use a stateful plugin, like rate-limiting-advanced or session? you're in for a fun time).

dndx:
…. fixes #6425

Originally I thought https://github.com/Kong/kong/pull/6517/files#diff-1958886e00e07e58fb444abe5b024fa98ba0c67c902d4c017fb41d97928de428R1238-R1240 is needed to achieve behavior described by @hishamhm in https://github.com/Kong/kong/issues/6425#issuecomment-710225801, but the test shows that it seems to be unnecessary.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -545,7 +545,6 @@ local function marshall_route(r)
 
   -- snis
 
-
   if snis then
     if type(snis) ~= "table" then
       return nil, "snis field must be a table"
@@ -974,7 +973,7 @@ do
 
     [MATCH_RULES.SNI] = function(route_t, ctx)
       local sni = route_t.snis[ctx.sni]
-      if sni then
+      if sni or ctx.req_scheme == "http" then
         ctx.matches.sni = ctx.sni
         return true
       end
@@ -1236,6 +1235,10 @@ function _M.new(routes)
         return r1.max_uri_length > r2.max_uri_length
       end
 
+      --if #r1.route.protocols ~= #r2.route.protocols then
+      --  return #r1.route.protocols < #r2.route.protocols
+      --end
+
       if r1.route.created_at ~= nil and r2.route.created_at ~= nil then
         return r1.route.created_at < r2.route.created_at
       end
@@ -1354,6 +1357,7 @@ function _M.new(routes)
     ctx.req_method     = req_method
     ctx.req_uri        = req_uri
     ctx.req_host       = req_host
+    ctx.req_scheme     = req_scheme
     ctx.req_headers    = req_headers
     ctx.src_ip         = src_ip or ""
     ctx.src_port       = src_port or ""
--- a/spec/01-unit/08-router_spec.lua
+++ b/spec/01-unit/08-router_spec.lua
@@ -3361,7 +3361,7 @@ describe("Router", function()
         assert.same(use_case[1].route, match_t.route)
       end)
 
-      it("[sni] is ignored for http request but does not shadow `protocols={'http'}` only routes", function()
+      it("[sni] is ignored for http request without shadowing routes with `protocols={'http'}`. Fixes #6425", function()
         local match_t = router_ignore_sni.select(nil, nil, "sni.example.com",
                                                  "http", nil, nil, nil, nil,
                                                  nil)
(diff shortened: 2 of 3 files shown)


kikito (changes requested):
In addition to this change, have you considered the other two discussed in #6425: 
* restricting the existence of the `snis` field to when `protocols` does has the appropriate values
* adding a test to ensure that we don't lose [the cases we want to preserve](https://github.com/Kong/kong/issues/6425#issuecomment-710225801)

bungle (approved):
As I now understand how it works, I will approve this. @kikito's comments are worth to note and fix still. While testing this I found that we have a small performance issue when Host header contains a port, it will never be matched using reduced candidates. That we need to fix on some other pr.

dndx:
@kikito Thanks for the reviews.

For 1, such check already exists in https://github.com/Kong/kong/blob/next/kong/db/schema/entities/routes.lua#L49-L56.

For 2, our integration tests ensures that the new behavior does not override any existing routers with `protocol = http`. In addition, both me and @bungle had independently confirm that both behaviors described in https://github.com/Kong/kong/issues/6425#issuecomment-710225801 are preserved in this PR.

PR #6518, opened 2020-10-26
fix(migrations) iterate target entities to find duplicates

locao:


kikito (approved):
Tried this change in kong_upgrade_tests. It fixed the problem I was seeing locally. Approving so I can merge and re-test in travis.

PR #6528, opened 2020-10-29
chore(docs) better document turning off server tokens

Tieske:
Currently is seems that setting `headers=off` is enough to prevent
leaking identifiable data to the client, which is not the case.
The additional nginx directive is also required.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -593,7 +593,9 @@
                                  # to `off`, which prevents Kong from injecting
                                  # any of the above headers. Note that this
                                  # does not prevent plugins from injecting
-                                 # headers of their own.
+                                 # headers of their own. When using 'off' it
+                                 # makes sense to also review the
+                                 # `nginx_http_more_clear_headers` setting.
                                  #
                                  # Example: `headers = via, latency_tokens`
 
@@ -795,6 +797,17 @@
                                           # allocations to process many
                                           # concurrent large request bodies.
                                           # See http://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size
+
+#nginx_http_more_clear_headers =
+                            # Allows modification of standard nginx headers.
+                            # When disabling the server tokens using the
+                            # `headers` setting, then this setting should
+                            # probably be set to `Server` to clear any
+                            # information that might expose the server and
+                            # server version to the client. These headers are
+                            # typically injected by Nginx when an error occurs
+                            # that does not reach Kong.
+                            # See https://github.com/openresty/headers-more-nginx-module#readme
 
 #------------------------------------------------------------------------------
 # DATASTORE


Tieske:
@dndx @bungle can either of you please verify what I've written?

@lena-larionova @tlblessing since this concerns docs, a review from you would also be nice.

kong.conf.default
```diff
@@ -593,7 +593,9 @@
                                  # to `off`, which prevents Kong from injecting
                                  # any of the above headers. Note that this
                                  # does not prevent plugins from injecting
-                                 # headers of their own.
+                                 # headers of their own. When using 'off' it
+                                 # makes sense to also review the
+                                 # `nginx_http_more_clear_headers` setting.
```
lena-larionova, line 598:
```suggestion
                                 # headers of their own. When using 'off', 
                                 # also review the `nginx_http_more_clear_headers`
                                 # setting.
```

kong.conf.default
```diff
@@ -796,6 +798,17 @@
...
 
+#nginx_http_more_clear_headers =
+                            # Allows modification of standard nginx headers.
+                            # When disabling the server tokens using the
+                            # `headers` setting, then this setting should
+                            # probably be set to `Server` to clear any
+                            # information that might expose the server and
+                            # server version to the client. These headers are
+                            # typically injected by Nginx when an error occurs
+                            # that does not reach Kong.
```
lena-larionova, line 809:
```suggestion
                            # Allows modification of standard Nginx headers.
                            # These headers are typically injected by Nginx when
                            # an error occurs that does not reach Kong.
                            # 
                            # When disabling the server tokens using the
                            # `headers` setting, also set 
                            # `nginx_http_more_clear_headers` to `Server` to 
                            # clear any information that might expose the server
                            # and server version to the client.
```

lena-larionova (commented):
Thanks for the ping; some edits for clarity.

bungle:
@Tieske,

> Currently is seems that setting headers=off is enough to prevent leaking identifiable data to the client, which is not the case.

when it is not a case?

Tieske:
Didn't know about that PR, seems to cover it. The reason I mentioned errors is because I thought those were in some cases handled by Nginx (and then the request doesn't even reach Kong) and then Nginx inserts these headers that leak information. But I'm not sure about that, which was exactly the reason for me to request your review, so thx!

Essentially, if server tokens are off, we should under no circumstances leak any info, neither from Kong nor Nginx. 

Did this land in Kong 2.2? if that's the case, then we can probably close this PR, since it is no longer relevant.

bungle:
> Didn't know about that PR, seems to cover it. The reason I mentioned errors is because I thought those were in some cases handled by Nginx (and then the request doesn't even reach Kong) and then Nginx inserts these headers that leak information. But I'm not sure about that, which was exactly the reason for me to request your review, so thx!
> 
> Essentially, if server tokens are off, we should under no circumstances leak any info, neither from Kong nor Nginx.
> 
> Did this land in Kong 2.2? if that's the case, then we can probably close this PR, since it is no longer relevant.

I'll check where it landed, and I will also check if there are spots we don't cover at the moment.

bungle:
> > Didn't know about that PR, seems to cover it. The reason I mentioned errors is because I thought those were in some cases handled by Nginx (and then the request doesn't even reach Kong) and then Nginx inserts these headers that leak information. But I'm not sure about that, which was exactly the reason for me to request your review, so thx!
> > Essentially, if server tokens are off, we should under no circumstances leak any info, neither from Kong nor Nginx.
> > Did this land in Kong 2.2? if that's the case, then we can probably close this PR, since it is no longer relevant.
> 
> I'll check where it landed, and I will also check if there are spots we don't cover at the moment.

I didn't find spots where we still leak this information, so I think we do not need this PR?

Tieske:
correct!

PR #6536, opened 2020-11-03
feat(db) certificates to have cert_alt and key_alt fields

bungle:
### Summary

The PR #6509 added support for static configuration to be able to set multiple certificates and keys with:

- `ssl_cert` and `ssl_cert_key`
- `admin_ssl_cert` and `admin_ssl_cert_key`
- `status_ssl_cert` and `status_ssl_cert_key`

But it didn't yet allow `certificates entity` to contain multiple certificates. This PR changes `certificates entity` to contain two new fields:

- `cert_alt`
- `key_alt`

Where you can configure alternative certificate and the matching key, this way you can use `cert` and `key` to store ECDSA certificate and `cert_alt` and `cert_key` to store RSA certificate.

### DIFF
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -1167,6 +1167,23 @@ return {
           description = [[PEM-encoded private key of the SSL key pair.]],
           example = "-----BEGIN RSA PRIVATE KEY-----..."
         },
+        cert_alt = {
+          description = [[
+            PEM-encoded public certificate chain of the alternate SSL key pair.
+            This should only be set if you have both RSA and ECDSA types of
+            certificate available and would like Kong to prefer serving using
+            ECDSA certs when client advertises support for it.
+          ]],
+          example = "-----BEGIN CERTIFICATE-----...",
+        },
+        key_alt = {
+          description = [[PEM-encoded private key of the alternate SSL key pair.
+            This should only be set if you have both RSA and ECDSA types of
+            certificate available and would like Kong to prefer serving using
+            ECDSA certs when client advertises support for it.
+          ]],
+          example = "-----BEGIN EC PRIVATE KEY-----..."
+        },
         tags = {
           description = [[
             An optional set of strings associated with the Certificate for grouping and filtering.
--- a/kong/db/schema/entities/certificates.lua
+++ b/kong/db/schema/entities/certificates.lua
@@ -2,6 +2,10 @@ local typedefs = require "kong.db.schema.typedefs"
 local openssl_pkey = require "resty.openssl.pkey"
 local openssl_x509 = require "resty.openssl.x509"
 
+
+local type = type
+
+
 return {
   name        = "certificates",
   primary_key = { "id" },
@@ -13,10 +17,13 @@ return {
     { created_at     = typedefs.auto_timestamp_s },
     { cert           = typedefs.certificate { required = true }, },
     { key            = typedefs.key         { required = true }, },
+    { cert_alt       = typedefs.certificate { required = false }, },
+    { key_alt        = typedefs.key         { required = false }, },
     { tags           = typedefs.tags },
   },
 
   entity_checks = {
+    { mutually_required = { "cert_alt", "key_alt" } },
     { custom_entity_check = {
       field_sources = { "cert", "key" },
       fn = function(entity)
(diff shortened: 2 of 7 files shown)


bungle:
Here I decided to implement this by adding `cert_alt` and `key_alt` fields to `certificates` entity. This felt like a most straightforward way to do it, and it should not break backward compatibility. Bad side of this is that it allows 2 types of certs, and that's it. E.g. third type of cert cannot be added. But considering how far have we been able to go with just 1 cert type, this should be good enough.

Other ways to do it:
1. make snis to be able to refer multiple certs
2. make cert and key arrays instead of strings
3. make users to bundle certs together and store them to cert and key (and we then parse that on use)

The 1 is probably hard to implement (multiple foreign keys or foreign key array, or many-to-many tables). Also it feels that this feature is used more to different types of certs rather than different certs, thus 1 seems not that good.

The 2 would be nice, but this would break backward compatibility.

The 3 is not very nice from usability point of view, and can lead to source of bugs, where code assumes (like it could today) that cert is a single cert and key is single key (not a bundle).

kong/runloop/certificate.lua
```diff
@@ -265,6 +265,20 @@ local function execute()
...
+      log(ERR, "could not set alternate configured certificate: ", err)
+      return ngx.exit(ngx.ERROR)
+    end
+
+    ok, err = set_priv_key(cert_and_key.key_alt)
+    if not ok then
+      log(ERR, "could not set alternate configured private key: ", err)
+      return ngx.exit(ngx.ERROR)
+    end
+  end
```
hbagdi, line 281: Questions:

What is the certificate selection or preference order when multiple cert-key pairs are present for a single SNI?

This PR adds in an ability to add 2 instead of 1 certificate for an SNI. Should we not make this feature more generic where, for some value of n, n certificates can be associated with a single SNI? Why have an arbitrary limit of 2?


I do get the conundrum of how to avoid the breaking change.
Here is a draft proposal:

Let's change the schema to:
```yaml
certs:
- cert: ""
  key: ""
  bundle: # come up with a better name
  - cert: ""
     key:  ""
```

And then update the validation to ensure that either `.cert` and .key` are provided or validate that len(bundle)>0 and contains valid cert key pairs.

fffonion: It depends on the cipher suite negotiation, the order where set_cert_* is called won't make difference. With [intermediate](https://github.com/Kong/kong/blob/master/kong/conf_loader/init.lua#L30) or old and we let Kong to prefer server ciphers, ECDSA certs will be selected first. In tls1.3 cipher suites doesn't include cert type/signature type but I think most implementations will just select ECDSA over RSA if both present.
That also means there'll only be two kinds of signing algorithm supported by most server/clients. I'd imagine you can set a third eddsa cert but it will never be selected.

autodoc/admin-api/data/admin-api.lua
```diff
@@ -1167,6 +1167,14 @@ return {
           description = [[PEM-encoded private key of the SSL key pair.]],
           example = "-----BEGIN RSA PRIVATE KEY-----..."
         },
+        cert_alt = {
+          description = [[PEM-encoded public certificate chain of the alternate SSL key pair.]],
```
dndx, line 1171:
```suggestion
          description = [[
            PEM-encoded public certificate chain of the alternate SSL key pair.
            This should only be set if you have both RSA and ECDSA types of
            certificate available and would like Kong to prefer serving using
            ECDSA certs when client advertises support for it.
          ]],
```

autodoc/admin-api/data/admin-api.lua
```diff
@@ -1167,6 +1167,14 @@ return {
...
+          example = "-----BEGIN CERTIFICATE-----...",
+        },
+        key_alt = {
+          description = [[PEM-encoded private key of the alternate SSL key pair.]],
```
dndx, line 1175:
```suggestion
          description = [[PEM-encoded private key of the alternate SSL key pair.
            This should only be set if you have both RSA and ECDSA types of
            certificate available and would like Kong to prefer serving using
            ECDSA certs when client advertises support for it.
          ]],
```

kong/db/schema/entities/certificates.lua
```diff
@@ -29,6 +36,21 @@ return {
...
       end,
-    } }
+    } },
+    { custom_entity_check = {
+      field_sources = { "cert_alt", "key_alt" },
+      fn = function(entity)
+        if type(entity.cert_alt) == "string" and type(entity.key_alt) == "string" then
+          local cert_alt = openssl_x509.new(entity.cert_alt)
+          local key_alt = openssl_pkey.new(entity.key_alt)
+
```
dndx, line 46: Other types of check we should do because this feature could easily be used incorrectly for non-suspecting users:

1. The hash of `cert_alt` and `cert` differs. This can be obtained using https://github.com/fffonion/lua-resty-openssl#x509digest. This ensures user does not accidentally put both fields to the same cert because they may thought both are required.
2. The type of public key for `cert_alt` and `cert` differs. I think this could be done using `x509:get_pkey():get_key_type` and check the `.id` field of the returned object differs. (@fffonion does this looks correct to you?)
This ensures user does not accidentally put both fields to the same cert type (for example, put two RSA certs of different expiration date and hoping it would work).

bungle: @dndx, I have now added the `distinct` validation and test, but I needed to add this too: https://github.com/Kong/kong/pull/6536/commits/93d53663c91412929c4a5e5faa433da0bd6d460d

Not sure if it is such a big deal, if it is, then I need to change the tests. I think running `distinct` on invalid fields is not needed, what do you think? The value of it is that we can report the error at the same time as the other errors, I think.

dndx (approved):

kong/db/schema/entities/certificates.lua
bungle: Distinct validator was now replaced with custom validation.

PR #6545, opened 2020-11-05
feat(async) new kong.async library and changes to utilise it

bungle:
### Summary

This PR comes with several commits:

#### fix(tests) fix a flaky test of kong quit

This has been a small problem on our test suite, that seems to trigger more often today. This just changes the somewhat flaky test to be bit more relaxed on small amount of time variance.

#### feat(async) implement kong.async:run function

Kong async library was written to replace `ngx.timer.*` apis with a lighter constructs based on `ngx.threads`.

This initial commit adds following methods:
- `local async = kong.async.new()` (used to prepare an instance of `kong.async` and prepare some of the data structures)
- `ok, err = kong.async:init_worker` (used to start the timers that this library needs)
- `kong.async:run(func, ...)` (used to run a function asynchronously from the thread that calls it)

#### feat(async) implement kong.async:every function

The difference between `kong.async:every` and `ngx.timer.every` is that `kong.async:every` collects recurring functions in buckets (by delay) and then executes them with `kong.async:run` on each interval.

So, if you have 2 functions that run on every minute, we will only create a single timer instead of two that you would need with `ngx.timer.every`. We still run them in parallel (co-operatively), thanks to `kong.async:run`
that this uses behind the scenes.

#### feat(async) add periodic logging to kong.async library

This commit will make `kong.async` library to log state of the async calls to Nginx error log with varyint log levels related to pending queue size.

#### refactor(async) add a local reusable job queuing function

Just consolidates similar code from two places in `kong.async` library to a function.

#### feat(async) collect the metrics with kong.async library 

Collects some metrics with `kong.async` library:
1. time when job was queued
2. time when job was started
3. time when job was finished

With this we can calculate:
1. latency
2. execution time
3. and queue to finish time

#### feat(async) implement kong.async:data function

The data function return raw data metric collected by `kong.async` library. Each item in the returned data
array contains a array with three values:

1. job queue time
2. job start time
3. job end time

#### feat(async) implement kong.async:stats function

Return stats from the library, for example:

```lua
{
  done = 5,
  pending = 4,
  running = 3,
  errored = 0,
  refused = 0,
  latency = {
    all = {
      size   = 20,
      mean   = 1,
      median = 2,
      p95    = 12,
      p99    = 20,
      p999   = 150,
      max    = 200,
      min    = 0,
    },
    hour = {
      size   = 10,
      mean   = 1,
      median = 2,
      p95    = 12,
      p99    = 20,
      p999   = 150,
      max    = 200,
      min    = 0,
    },
    minute = {
      size   = 1,
      mean   = 1,
      median = 2,
      p95    = 12,
      p99    = 20,
      p999   = 150,
      max    = 200,
      min    = 0,
    },
  },
  runtime = {
    all = {
      size   = 20,
      mean   = 1,
      median = 2,
      p95    = 12,
      p99    = 20,
      p999   = 150,
      max    = 200,
      min    = 0,
    },
    hour = {
      size   = 10,
      mean   = 1,
      median = 2,
      p95    = 12,
      p99    = 20,
      p999   = 150,
      max    = 200,
      min    = 0,
    },
    minute = {
      size   = 1,
      mean   = 1,
      median = 2,
      p95    = 12,
      p99    = 20,
      p999   = 150,
      max    = 200,
      min    = 0,
    },
  }
}
```

#### feat(core) initialize kong.async library on init worker

Initialized `kong.async` library on `init worker` phase.

#### feat(api) export kong.async library statistics on admin api 

Exports `kong.async` library statistics on Kong admin api.

#### chore(plugins) use kong.async:run instead of ngx.timer.at

Many of our plugins that execute post-content phase (such as `logging plugins`) also need cosockets, and to exit the limitation of the post-content phases they also tend to execute `ngx.timer.at`. This can in high traffic server exhaust the timer resources, and start to error.

The `kong.async:run(func, ...)` was written to replace the `ngx.timer.at(0, func, ...)`, and this commit does so to the plugins.

#### chore(db) use kong.async:run to warmup dns cache instead of ngx.timer.at

Replaces `ngx.timer.at` with `kong.async:run` on DNS cache warmup.

#### chore(balancer) use kong.async:every with balancer 

Replaces `ngx.timer.at` calls with `kong.async:every` on balancer.

#### chore(reports) use kong.async:every with reports

Changes `kong.reports` to use `kong.async:every` instead of recursive calls to `ngx.timer.at`.

#### chore(api) reports timers to use kong.async:run instead of ngx.timer.at

Replaces `ngx.timer.at` calls with `kong.async:at` on api where reports timer was called.

#### chore(handler) use kong.async instead of ngx.timer.*

This updates `kong.runloop.handler` to use `kong.async` instead of `ngx.timer` api.

#### chore(db) pg/c* connector to use kong.async:every instead of ngx.timer.every

This updates Postgres and Cassandra connector to use `kong.async:every` instead of using `ngx.timer.every`.

#### chore(go) run go plugins log phase with kong.run:async instead of ngx.timer.at

Uses `kong.run:async` to run Go plugins' log phase instead of spawning new timer for each with `ngx.timer.at(0, ...)`.

#### feat(async) implement kong.async:after function

This commit implements the last missing function `kong.async:after`. This function runs with 100 ms interval (with `ngx.timer.every`), though it can be adjusted by changing module level constant. The timer is implement
roughly by following: https://arxiv.org/pdf/1906.10860.pdf.

With this function you can queue a job to be executed after x seconds, but you can use millisecond granularity e.g. with `0.001`. Because there is `100ms` interval (preconfigured), it means that it will add 0 - 100 ms latency to function call.

Internally this function also executes the functions using `kong.async:run` that could add small amount of latency as well.

#### chore(tests) make slow-query test plugin to use kong.async:run

Just changes custom test plugin `slow-query` to use `kong.async:run` instead of `ngx.timer.at`.

### DIFF
--- a/kong/async.lua
+++ b/kong/async.lua
@@ -48,7 +48,9 @@ local function job_thread(self, index)
         self.tail = tail
         self.jobs[tail] = nil
         self.running = self.running + 1
+        self.time[tail][2] = ngx.now() * 1000
         ok, err = job()
+        self.time[tail][3] = ngx.now() * 1000
         self.running = self.running - 1
         self.done = self.done + 1
         if not ok then
@@ -183,6 +185,7 @@ local function queue_job(self, recurring, func, ...)
 
   self.head = self.head == QUEUE_SIZE and 1 or self.head + 1
   self.jobs[self.head] = recurring and func or create_job(func, ...)
+  self.time[self.head][1] = ngx.now() * 1000
   self.work:post()
 
   return true
@@ -218,8 +221,14 @@ async.__index = async
 --
 -- @treturn table an instance of `kong.async`
 function async.new()
+  local time = kong.table.new(QUEUE_SIZE, 0)
+  for i = 1, QUEUE_SIZE do
+    time[i] = kong.table.new(3, 0)
+  end
+
   return setmetatable({
     jobs = kong.table.new(QUEUE_SIZE, 0),
+    time = time,
     work = semaphore.new(),
     buckets = {},
     running = 0,
--- a/kong/async.lua
+++ b/kong/async.lua
@@ -3,6 +3,7 @@ local semaphore = require "ngx.semaphore"
 
 local ngx = ngx
 local kong = kong
+local math = math
 local type = type
 local pcall = pcall
 local string = string
@@ -308,6 +309,36 @@ function async:every(delay, func, ...)
   bucket.jobs[bucket.head] = create_recurring_job(create_job(func, ...))
 
   return true
+end
+
+
+---
+-- Kong async raw metrics data
+--
+-- @tparam  from[opt]  data start time (from unix epoch)
+-- @tparam  to[opt]    data end time (from unix epoch)
+-- @treturn table      a table containing the metrics
+-- @treturn number     number of metrics returned
+function async:data(from, to)
+  local time = self.time
+  local done = math.min(self.done, QUEUE_SIZE)
+  if not from and not to then
+    return time, done
+  end
+
+  from = from and from * 1000 or 0
+  to   = to   and to   * 1000 or math.huge
+
+  local data = kong.table.new(done, 0)
+  local size = 0
+  for i = 1, done do
+    if time[i][1] >= from and time[i][3] <= to then
+      size = size + 1
+      data[size] = time[i]
+    end
+  end
+
+  return data, size
 end
 
 
--- a/kong/async.lua
+++ b/kong/async.lua
@@ -5,8 +5,9 @@ local ngx = ngx
 local kong = kong
 local math = math
 local type = type
-local pcall = pcall
 local table = table
+local debug = debug
+local xpcall = xpcall
 local string = string
 local select = select
 local unpack = unpack
@@ -81,23 +82,23 @@ local function log_timer(premature, self)
     return true
   end
 
-  local debug  = QUEUE_SIZE / 10000
-  local info   = QUEUE_SIZE / 1000
-  local notice = QUEUE_SIZE / 100
-  local warn   = QUEUE_SIZE / 10
-  local err    = QUEUE_SIZE
+  local dbg = QUEUE_SIZE / 10000
+  local nfo = QUEUE_SIZE / 1000
+  local ntc = QUEUE_SIZE / 100
+  local wrn = QUEUE_SIZE / 10
+  local err = QUEUE_SIZE
 
   local pending = get_pending(self, QUEUE_SIZE)
 
   local msg = string.format("async jobs: %u running, %u pending, %u errored, %u refused, %u aborted, %u done",
                             self.running, pending, self.errored, self.refused, self.aborted, self.done)
-  if pending <= debug then
+  if pending <= dbg then
     kong.log.debug(msg)
-  elseif pending <= info then
+  elseif pending <= nfo then
     kong.log.info(msg)
-  elseif pending <= notice then
+  elseif pending <= ntc then
     kong.log.notice(msg)
-  elseif pending < warn then
+  elseif pending < wrn then
     kong.log.warn(msg)
   elseif pending < err then
     kong.log.err(msg)
@@ -229,14 +230,16 @@ local function create_job(func, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, ...)
   local argc = select("#", ...)
   if argc == 0 then
     return function()
-      return pcall(func, ngx.worker.exiting(), a1, a2, a3, a4, a5, a6, a7, a8, a9, a10)
+      return xpcall(func, debug.traceback, ngx.worker.exiting(),
+                    a1, a2, a3, a4, a5, a6, a7, a8, a9, a10)
     end
   end
 
   local args = { ... }
   return function()
-    local pok, res, err = pcall(func, ngx.worker.exiting(), a1, a2, a3, a4, a5, a6, a7, a8, a9, a10,
-                                unpack(args, 1, argc))
+    local pok, res, err = xpcall(func, debug.traceback, ngx.worker.exiting(),
+                                 a1, a2, a3, a4, a5, a6, a7, a8, a9, a10,
+                                 unpack(args, 1, argc))
     if not pok then
       return nil, res
     end
--- a/kong/async.lua
+++ b/kong/async.lua
@@ -175,6 +175,20 @@ local function create_job(func, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, ...)
 end
 
 
+local function queue_job(self, recurring, func, ...)
+  if get_pending(self) == QUEUE_SIZE then
+    self.refused = self.refused + 1
+    return nil, "async queue is full"
+  end
+
+  self.head = self.head == QUEUE_SIZE and 1 or self.head + 1
+  self.jobs[self.head] = recurring and func or create_job(func, ...)
+  self.work:post()
+
+  return true
+end
+
+
 local function create_recurring_job(job)
   local running = false
 
@@ -190,15 +204,7 @@ local function create_recurring_job(job)
       return nil, "recurring job is already running"
     end
 
-    if get_pending(self) == QUEUE_SIZE then
-      return nil, "async queue is full"
-    end
-
-    self.head = self.head == QUEUE_SIZE and 1 or self.head + 1
-    self.jobs[self.head] = recurring_job
-    self.work:post()
-
-    return true
+    return queue_job(self, true, recurring_job)
   end
 end
 
@@ -249,16 +255,7 @@ end
 -- @treturn true|nil   `true` on success, `nil` on error
 -- @treturn string|nil `nil` on success, error message `string` on error
 function async:run(func, ...)
-  if get_pending(self) == QUEUE_SIZE then
-    self.refused = self.refused + 1
-    return nil, "async queue is full"
-  end
-
-  self.head = self.head == QUEUE_SIZE and 1 or self.head + 1
-  self.jobs[self.head] = create_job(func, ...)
-  self.work:post()
-
-  return true
+  return queue_job(self, false, func, ...)
 end
 
 
--- a/kong/async.lua
+++ b/kong/async.lua
@@ -227,16 +227,16 @@ end
 
 local function create_job(func, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, ...)
   local argc = select("#", ...)
-  local args = argc > 0 and { ... }
-
-  if not args then
+  if argc == 0 then
     return function()
       return pcall(func, ngx.worker.exiting(), a1, a2, a3, a4, a5, a6, a7, a8, a9, a10)
     end
   end
 
+  local args = { ... }
   return function()
-    local pok, res, err = pcall(func, ngx.worker.exiting(), a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, unpack(args, 1, argc))
+    local pok, res, err = pcall(func, ngx.worker.exiting(), a1, a2, a3, a4, a5, a6, a7, a8, a9, a10,
+                                unpack(args, 1, argc))
     if not pok then
       return nil, res
     end
--- a/kong/async.lua
+++ b/kong/async.lua
@@ -6,6 +6,7 @@ local kong = kong
 local math = math
 local type = type
 local pcall = pcall
+local table = table
 local string = string
 local select = select
 local unpack = unpack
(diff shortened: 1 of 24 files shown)


Tieske:
if creating buckets then everything behind a slow job in the same bucket will still be delayed.

I'd say create a single bucket/queue, and have all workers just grab the next item from that queue.

bungle:
> if creating buckets then everything behind a slow job in the same bucket will still be delayed.

No. Slowness means usually yielding, which means that other threads on that bucket can run. If it is blocking code, then we cannot do anything about it.

bungle:
> I'd say create a single bucket/queue, and have all workers just grab the next item from that queue.

This would mean that they also edit shared structures. A source of many bugs.

murillopaula:
acme uses ngx.timer.every - should we change it?

zipkin is one that uses ngx.timer.at, so we need to check the plugins that are not in Kong/kong too

the EE plugins, a lot of them use timers as well

bungle:
@Tieske I added another commit that just uses workers.

javierguerragiraldez (commented):
i like it.   it's unfortunate that `unpack()` and `table.remove()` are NYIs, but in all any coroutine boundary is already a hard compile barrier.

bungle:
> i like it. it's unfortunate that `unpack()` and `table.remove()` are NYIs, but in all any coroutine boundary is already a hard compile barrier.

I changed it to fixed size array, so no need for table.remove.

kong/async.lua
```diff
@@ -0,0 +1,110 @@
...
+local unpack = unpack or table.unpack
+local setmetatable = setmetatable
+
+
+local WORKERS = 100
+
+
+local function worker(self)
+  while not ngx.worker.exiting() do
+    local ok, err = self.work:wait(1)
```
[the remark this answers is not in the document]
javierguerragiraldez: ah found it on the `:run()` method

kong/async.lua
```diff
@@ -0,0 +1,110 @@
...
+    return
+  end
+
+  local n = self.workers.n
+
+  for i = 1, n do
+    self.workers[i] = ngx.thread.spawn(worker, self)
+  end
+
+  local ok, err = ngx.thread.wait(unpack(self.workers, 1, n))
```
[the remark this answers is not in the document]
bungle: Yes, here I implemented `wait any`, so if any of the threads die (they should not), we kill them all and try to start again. @javierguerragiraldez, do you think there is a better way to do it?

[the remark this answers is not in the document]
javierguerragiraldez: no, i think it's ok.  just wanted to be sure because it would became part of the contract but not obvious from the API

kong/async.lua
```diff
@@ -0,0 +1,110 @@
...
+      if n > 0 then
+        self.jobs.n = self.jobs.n - 1
+        local job = table.remove(self.jobs, 1)
+        local pok, res, err = job()
+        if not pok then
+          ngx.log(ngx.ERR, "async thread error: ", res)
+        elseif not res and err then
+          ngx.log(ngx.ERR, "async thread returned an error: ", err)
+        end
+      end
```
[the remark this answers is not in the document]
bungle: I added that, and I also added data to admin api, similar to how we report `timers` (yes, it is single worker data, but that is a good view of the situation in general).

bungle:
> acme uses ngx.timer.every - should we change it?

This only deals with `ngx.timer.at(0, ...)` case. The recurring timers we need different solution.

> zipkin is one that uses ngx.timer.at, so we need to check the plugins that are not in Kong/kong too

That is correct!

> the EE plugins, a lot of them use timers as well

Yes, let's see how this goes. It might improve things somewhat, that is it exchanges scarce resources (timers) to light threads. Light threads are also lighter construct in general. This implementation can queue up to 100000 functions, until it starts returning errors (if the functions are not processed as fast as they are added).  It is more than what we have for timers:

```
lua_max_running_timers 4096;
lua_max_pending_timers 16384;
```

The thing that I do not know is that what would be good number of threads to start (currently it starts 100 per worker), also what would be good size of the queue (now it is fixed size of 100.000 per worker).

kong/async.lua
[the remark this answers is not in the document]
bungle: We could also start collecting stats of average function execution time etc.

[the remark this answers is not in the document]
bungle: stats are now collected as well, and admin api shows the stats.

bungle:
> acme uses ngx.timer.every - should we change it?

There is now `kong.async:every` too, so we can use that. The thing is that this is implemented by just calling `ngx.timer.every`. The benefit is that it groups each `delay` to a bucket and tries to minimize the usage of timers this way. Of course if every `kong.async:every(delay, func, ...)` uses different `delay`, then there is no benefits, except one. The timer that this thing creates, does not execute the functions, but passes them to `kong.run:async`. This makes these `every` timers always lightning fast and non-yielding. It is an indirection, so it might add some additional latency, but that is usually not an issue with timer.every type of recursion.

javierguerragiraldez:
superceded by #6552

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+        local job = self.jobs[tail]
+        self.tail = tail
+        self.jobs[tail] = nil
+        ok, err = job()
+        if not ok then
+          kong.log.err("async thread #", index, " job error: ", err)
+        end
+      end
+
+    elseif err ~= "timeout" then
```
Tieske, line 65: is there a special case to be defined if the worker process starts exiting while this thread is waiting for `self.work:wait(1)`, because currently it starts logging errors if I read this code correctly.

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+                                  t[31], t[32], t[33], t[34], t[35], t[36], t[37], t[38], t[39], t[40],
+                                  t[41], t[42], t[43], t[44], t[45], t[46], t[47], t[48], t[49], t[50],
+                                  t[51], t[52], t[53], t[54], t[55], t[56], t[57], t[58], t[59], t[60],
+                                  t[61], t[62], t[63], t[64], t[65], t[66], t[67], t[68], t[69], t[70],
+                                  t[71], t[72], t[73], t[74], t[75], t[76], t[77], t[78], t[79], t[80],
+                                  t[81], t[82], t[83], t[84], t[85], t[86], t[87], t[88], t[89], t[90],
+                                  t[91], t[92], t[93], t[94], t[95], t[96], t[97], t[98], t[99], t[100])
+
+  if not ok then
+    kong.log.err("async thread error: ", err)
```
Tieske, line 127: under what circumstances would this happen?

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+  end
+
+  for i = 100, 1, -1 do
+    ngx.thread.kill(t[i])
+  end
+
+  return init_worker_timer(ngx.worker.exiting(), self)
```
Tieske, line 134: why this recursion? essentially the error above should not happen. My 2cts; if it does happen just error out and leave it there. The system will be in an undetermined state anyway since workers failed that shouldn't have failed.

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+end
+
+
+local function create_job(func, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, ...)
+  local argc = select("#", ...)
+  local args = argc > 0 and { ... }
+
+  if not args then
+    return function()
+      return pcall(func, ngx.worker.exiting(), a1, a2, a3, a4, a5, a6, a7, a8, a9, a10)
```
Tieske, line 161: this should be an `xpcall` that captures the stack trace, otherwise debugging becomes really hard

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+    end
+  end
+
+  return function()
+    local pok, res, err = pcall(func, ngx.worker.exiting(), a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, unpack(args, 1, argc))
```
Tieske, line 166: xpcall

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+
+    return nil, err
+  end
+end
+
+
+local async = {}
+
+
+async.__index = async
```
Tieske, line 287: nitpick: the two lines belong together imo, drop the vertical whitespace

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+  }, async)
+end
+
+
+---
+-- Initializes `kong.async` timers
+--
+-- @treturn boolean|nil `true` on success, `nil` on error
+-- @treturn string|nil  `nil` on success, error message `string` on error
+function async:init_worker()
```
Tieske, line 320: can we use a different name `init_worker` is an nginx phase, better not create confusion. How about `start`?

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+  local ok, err = ngx.timer.at(0, init_worker_timer, self)
+  if not ok then
+    return nil, err
```
Tieske, line 323: this is a very essential problem, we should log it here to prevent others from doing it later

kong/async.lua
```diff
@@ -0,0 +1,164 @@
...
+---
+-- Run a function asynchronously
+--
+-- @tparam  function   a function to run asynchronously
+-- @tparam  ...[opt]   function arguments
+-- @treturn true|nil   `true` on success, `nil` on error
+-- @treturn string|nil `nil` on success, error message `string` on error
+function async:run(func, ...)
+  local pending = get_pending(self)
+  if pending == QUEUE_SIZE then
```
Tieske:
```lua
  if get_pending(self) >= QUEUE_SIZE then
```
one less local, and more defensive

kong/async.lua
bungle: Recursion is better. In normal case we end up here when one (or more) of the threads terminates. The most common case for thread to terminate is that Nginx is shutdowning. This just calls function then once, as it checks the `premature` as a first thing, and then it terminates it. Doing recursion will not make state undetermined as we kill all the threads. The queue can still be filled while there are no threads processing them. If for some reason we did exit with some kind of error we still restart the threads. That's it. TL;DR; in 99.999% of the time it already does what you are describing.

kong/async.lua
bungle: I think this is better, as it is meant to be called on `nginx` `init worker` phase.

kong/async.lua
```diff
@@ -160,5 +221,47 @@ function async:run(func, ...)
...
+                                                "'second', 'minute', 'hour', 'month' or 'year'")
+
+  local bucket = self.buckets[delay]
+  if bucket then
+    if bucket.head == BUCKET_SIZE then
+      return nil, "async bucket (" .. delay .. ") is full"
+    end
+
+  else
+    local ok, err = ngx.timer.every(delay, every_timer, self, delay)
```
Tieske, line 364: this still uses 1 timer for each "every" call right? So this would not have prevented the problem of each balancer creating its own timer and running out of them.

I would suggest to use the timerwheel, since that would run a single timer, dealing with every pattern here.

kong/async.lua
bungle: I tought about it, but it didn't feel worth it. This function is ever to be called once in `nginx` `init worker` phase, and usage of this library shows that we never call it more than once in Kong code.

kong/async.lua
Tieske: Why? there is nothing here that would require that?

kong/async.lua
Tieske: the not-undertermined state you are referring to is from this code perspective I think? from Kong perspective it is an undetermined state, if one of those jobs is supposed to handle some invalidation, that now doesn't run. Then there is no way to know what will go wrong next, but probably a lot. It should forcefully exit with a big error.

kong/async.lua
bungle: This will create `ngx.timer.every` timer for each `delay`. One timer per delay. The functions themselves are then ran in `kong.async:run` which means they are run in parallel. Timerwheel runs functions non-parallel and implements the whole scheduling logic. Essentially to make your own `ngx.timer.every` you should perhaps add one timer that queues functions in every millisecond. That is one way around it. Still it is very uncommon to setup different delays every time with `ngx.timer.every`, and thus if you do so, I think you are using the wrong function. This library does not support `kong.async:after` (aka ngx.timer.at(>0, ...) currently.

This would fix the balancer issue as:
1. it does not allow same function to run if other run has not returned
2. if the delays are common, it will create less timers
3. it uses kong.async:run to execute the functions
4. the every timer is just a blocking loop that queues functions (it always happens a snap, so every timers are not really weight here)

There is limit in `kong.run:async`, so it will start error when you hit the limit.

I can make alternative implementation with `ngx.timer.every` that runs on each millisecond, but I feel like it adds unneccessary load to system even when that function could run very fast. The idea is not to fully eliminate the use of timers. Timers are still very valuable.

bungle: This will create `ngx.timer.every` timer for each `delay`. One timer per delay. The functions themselves are then ran in `kong.async:run` which means they are run in parallel. Timerwheel runs functions non-parallel and implements the whole scheduling logic (which is already very nicely done with timer.every). Essentially to make your own `ngx.timer.every` you should perhaps add one timer that queues functions in every millisecond. That is one way around it. Still it is very uncommon to setup different delays every time with `ngx.timer.every`, and thus if you do so, I think you are using the wrong function. This library does not support `kong.async:after` (aka ngx.timer.at(>0, ...) currently.

This would fix the balancer issue as:
1. it does not allow same function to run if other run has not returned
2. if the delays are common, it will create less timers
3. it uses kong.async:run to execute the functions
4. the every timer is just a blocking loop that queues functions (it always happens a snap, so every timers are not really weight here)

There is limit in `kong.run:async`, so it will start error when you hit the limit.

I can make alternative implementation with `ngx.timer.every` that runs on each millisecond, but I feel like it adds unneccessary load to system even when that function could run very fast. The idea is not to fully eliminate the use of timers. Timers are still very valuable.

kong/async.lua
bungle: @Tieske, ok, then it is up to user to shoot into foot? Isn't it? Why would you do that? The function is named `init_worker`, it is very OpenResty specific. Why would you ever call `kong.async:init_worker` more than once per VM? Yes, if you really want to shoot your foot. Do we need to make bigger fences for people to understand that? I am not sure, we are adults after all.

kong/async.lua
bungle: Fixed that.

kong/async.lua
bungle: All jobs will run, why wouldn't they? The `tail` is moved just before running the job.

kong/async.lua
bungle: Only if you want to write that code `debug.traceback` in the `xpcall` function. And I am not sure I want to.

kong/async.lua
bungle: removed two new lines.

kong/async.lua
bungle: I cannot think of the situation, but it is in `ngx.thread:wait` api, so better to log it if it happens in some case.

kong/async.lua
bungle: I have never seen it starting to log errors in the case you said. I mean there are 100 threads on each worker and no error log into any logs with restarts or reloads. The `wait(1)` means `1 second at max`.

PR #6550, opened 2020-11-06
bump DNS LRU cache size

Tieske:
see https://github.com/Kong/lua-resty-dns-client/blob/master/src/resty/dns/client.lua#L459

### DIFF
--- a/kong/tools/dns.lua
+++ b/kong/tools/dns.lua
@@ -33,6 +33,7 @@ local setup_client = function(conf)
     staleTtl = conf.dns_stale_ttl,     -- ttl in seconds for records once they become stale
     order = conf.dns_order,            -- order of trying record types
     noSynchronisation = conf.dns_no_sync,
+    cacheSize = 60000,
   }
 
   assert(dns_client.init(opts))


kikito:
@Tieske is this PR against `2.0.4-lua-resty-dns-client-with-single-timer` still relevant?

Tieske:
probably not against that branch. But making the LRU size configurable is probably senbsible. But then there comes the question of visibility... if the cache size is configurable, you probably also want to be able to see utilization such that one knows when to increase it...

So thinking out loud;

- we need the size to be configurable
- we need stats from the dns resolver about hits/misses/utilization, etc for observability

not sure where to track?

kikito:
My impression is that we need three things for that:
* An issue for adding a config option + default value for the cache size in Kong
* Another issue for getting the dns stats added to the Admin Api (perhaps inside `/status` https://docs.konghq.com/2.0.x/admin-api/#retrieve-node-status )
  * If lua-resty-dns-client doesn't already externalize those stats, we need another issue in order to track that inside its repo. 

* In the meantime, this PR could be reconverted to a PR against `next` which sets the value of the lru cache size to some suitable default (don't know if `cacheSize = 60000` is ok for general use)

hbagdi:

PR #6552, opened 2020-11-06
minimal inter-modules RPC 

javierguerragiraldez:
### Summary

OpenResty http and stream modules keep Lua code in separate spaces; and for the time being, the shared dictionaries are shared only between the spaces of the same module.

This feature adds a simple RPC on top of a datagram-mode unix domain socket.  Just a call/result with a single string on each direction.

#### Register handler

From the `api.lua` module in a plugin directory, return a table with a `_stream` field.  For example:

```lua
return {
  _stream = function(data)
    return "back: " .. data
  end,
}
```

The handler will receive a single string in the `data` parameter, and should return a single string or `nil, err` on error.

#### Request:

Use `stream_api.request(<plugin name>, <data>)`  to call the handler registered by the named plugin.  On success, it will return the string returned by the handler.  On error it will return `nil, err`.

Replaces #6374

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -70,6 +70,7 @@ local dns = require "kong.tools.dns"
 local meta = require "kong.meta"
 local lapis = require "lapis"
 local runloop = require "kong.runloop.handler"
+local stream_api = require "kong.tools.stream_api"
 local clustering = require "kong.clustering"
 local singletons = require "kong.singletons"
 local declarative = require "kong.db.declarative"
@@ -1372,6 +1373,26 @@ function Kong.serve_cluster_listener(options)
   kong_global.set_phase(kong, PHASES.cluster_listener)
 
   return clustering.handle_cp_websocket()
+end
+
+
+local stream_plugins_api_loaded = false
+function Kong.stream_api()
+  if not stream_plugins_api_loaded then
+    local utils = require "kong.tools.utils"
+
+    for plugin_name in pairs(kong.configuration.loaded_plugins) do
+      local loaded, custom_endpoints = utils.load_module_if_exists("kong.plugins." .. plugin_name .. ".api")
+      if loaded and custom_endpoints._stream then
+        ngx.log(ngx.DEBUG, "Register stream api for plugin: ", plugin_name)
+        stream_api.register(plugin_name, custom_endpoints._stream)
+        custom_endpoints._stream = nil
+      end
+    end
+    stream_plugins_api_loaded = true
+  end
+
+  stream_api.handle()
 end
 
 
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -467,6 +467,10 @@ function Kong.init()
   -- Load plugins as late as possible so that everything is set up
   assert(db.plugins:load_plugin_schemas(config.loaded_plugins))
 
+  if subsystem == "stream" then
+    stream_api.load_handlers()
+  end
+
   if kong.configuration.database == "off" then
 
     local err
@@ -1376,22 +1380,7 @@ function Kong.serve_cluster_listener(options)
 end
 
 
-local stream_plugins_api_loaded = false
 function Kong.stream_api()
-  if not stream_plugins_api_loaded then
-    local utils = require "kong.tools.utils"
-
-    for plugin_name in pairs(kong.configuration.loaded_plugins) do
-      local loaded, custom_endpoints = utils.load_module_if_exists("kong.plugins." .. plugin_name .. ".api")
-      if loaded and custom_endpoints._stream then
-        ngx.log(ngx.DEBUG, "Register stream api for plugin: ", plugin_name)
-        stream_api.register(plugin_name, custom_endpoints._stream)
-        custom_endpoints._stream = nil
-      end
-    end
-    stream_plugins_api_loaded = true
-  end
-
   stream_api.handle()
 end
 
--- a/kong/tools/stream_api.lua
+++ b/kong/tools/stream_api.lua
@@ -14,8 +14,17 @@ local stream_api = {}
 local _handlers  = {}
 
 
-function stream_api.register(k, f)
-  _handlers[k] = f
+function stream_api.load_handlers()
+  local utils = require "kong.tools.utils"
+
+  for plugin_name in pairs(kong.configuration.loaded_plugins) do
+    local loaded, custom_endpoints = utils.load_module_if_exists("kong.plugins." .. plugin_name .. ".api")
+    if loaded and custom_endpoints._stream then
+      ngx.log(ngx.DEBUG, "Register stream api for plugin: ", plugin_name)
+      _handlers[plugin_name] = custom_endpoints._stream
+      custom_endpoints._stream = nil
+    end
+  end
 end
 
 function stream_api.request(key, data, socket_path)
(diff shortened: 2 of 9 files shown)


kong/init.lua
```diff
@@ -1375,6 +1376,26 @@ function Kong.serve_cluster_listener(options)
 end
 
 
+local stream_plugins_api_loaded = false
```
dndx, line 1379: Just curious, what's the reason for loading the endpoints lazily. One reason I see loading them during the `init` phase is to allow a bit more RAM to be saved due to the CoW behavior of `fork()`. It may also make testing during development easier to see because Kong will fail to start immediately for obvious things such as syntax error, instead of having to trigger the call later via other means.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  data, err = socket:receive()
+  if not data then
+    socket:close()
+    return data, err
+  end
+
+  local _, status, payload = st_unpack(data, "=SP")
+  if status == 0 then
```
dndx, line 38: It's not critical, but I do feel using `0` as the success code is a bit easier to remember. (I know OpenSSL is an exception, but neither Nginx nor typical C APIs the adopts `0` for error codes..) 😂 

Unless of course, that the `1` here is the number of return values like Lua APIs, then we can keep this format, but a comment would remind future readers of this.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,77 @@
...
+    return nil, payload
+  end
+
+  socket:close()
+  return payload
+end
+
+
+function stream_api.handle()
+
```
dndx, line 49:
```suggestion
```

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,77 @@
...
+
+function stream_api.register(k, f)
+  _handlers[k] = f
+end
+
+function stream_api.request(key, data, socket_path)
+  local socket = assert(ngx.socket.udp())
+  assert(socket:setpeername(socket_path or "unix:" .. PREFIX .. "/stream_rpc.sock"))
+
```
dndx, line 24: We should check for the `type` of the arguments here and throw an error if they mismatch (to aid debugging).

Also, maybe a size check on the argument is a good idea. I Googled around and it seems that there is a limit of datagram size that could be sent over domain sockets, however the number seems to vary depends on Kernel params. https://stackoverflow.com/questions/21856517/whats-the-practical-limit-on-the-size-of-single-packet-transmitted-over-domain

Maybe something a bit conservative like 128KB at first, we can always increase this in the future if needed.

kong/init.lua
javierguerragiraldez: no real good reason.  on the previous (http-like) design, i got some issues because it seemed that the endpoint table could interfere with the admin api extensions, so i put it on a very different place, and where I was sure it would only happen under the stream module.

this version is much simpler, so it's easier to put in the right place.  fixed

kong/tools/stream_api.lua
javierguerragiraldez: i was back and forth about one way or the other, but didn't find any important advantage on either side.
my reasoning was "0 -> false -> bad, 1 -> true -> good".

the executable return codes reasoning is like "there's a single good outcome, but many bad ones", but that means that I have to double- and triple-check any non-trivial logic in shell scripts.  that bothers me every day :woozy_face: 

on C APIs the reasoning is mostly the same "many different errors", but it's usually hidden in constants or enums, so it's not so weird that `xxx_OK == 0`  (still, i usually do `xxx_None = 0, xxx_Ok = 1, xxx_Err_yyy = 2, 3...`)

in this case, the specific error is carried on the payload, so i'm ok with having a single error condition, and leaned towards avoiding the shell awkwardness.  but in the end, in Lua 0 is truthy, so there's no danger of that.

changed.  even ended doing the smallest amount of "different errors", using 1=transport error (no handler) and 2=error from handler (they both end as `nil, <errmsg>`, so :man_shrugging: )

kong/tools/stream_api.lua
javierguerragiraldez: the typical limit i've seen is like 64KiB, but somewhere in openresty docs mentions some buffer limit of 8KiB.  I guess it's for UDP and not unix sockets but i haven't checked the sources.  added a test for 8000, which i guess should be more than enough for requests.

big responses are slightly more worrying, but i guess hard failing wouldn't help in debugging.  I've put just a WARN log, hope it's enough

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,100 @@
...
+  end
+end
+
+function stream_api.request(key, data, socket_path)
+  if type(key) ~= "string" or type(data) ~= "string" then
+    kong.log.error("Key and data must be strings")
+    return
+  end
+  if #data > 8000 then
+    kong.log.error("too much data")
```
dndx, line 36: I still feel that we should provide a stronger guarantee in regard of the maximum length supported by throwing a Lua error, otherwise people may be tempting to send large amount of data through this channel which is not designed for it.

Just logging the error would be much less useful for the end user.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,100 @@
...
+  local res
+  res, err = f(payload)
+  if not res then
+    kong.log.error(st_format("stream_api handler %q returned error: %q", key, err))
+    socket:send(st_pack("=SP", 2, tostring(err)))
+    return
+  end
+
+  res = tostring(res)
+  if #res > 8000 then
```
dndx, line 90: Same here, it should use the `MAX_LEN` variable as mentioned above, instead the magic number.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,100 @@
...
+
+  local res
+  res, err = f(payload)
+  if not res then
+    kong.log.error(st_format("stream_api handler %q returned error: %q", key, err))
+    socket:send(st_pack("=SP", 2, tostring(err)))
+    return
+  end
+
+  res = tostring(res)
```
dndx, line 89: Do we really want to do this? I feel that we should do a `type` check here as well and throw a Lua error if it is not `string`, because that means a bug in the handler.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,100 @@
...
+      kong.log.debug("Register stream api for plugin: ", plugin_name)
+      _handlers[plugin_name] = custom_endpoints._stream
+      custom_endpoints._stream = nil
+    end
+  end
+end
+
+function stream_api.request(key, data, socket_path)
+  if type(key) ~= "string" or type(data) ~= "string" then
+    kong.log.error("Key and data must be strings")
```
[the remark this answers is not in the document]
javierguerragiraldez: you're right.  hard error and full stack dump are better for debugging.  i prefer not to include the `level` arg to `error()`, since it easily leads to suppressing the most important parts after an innocent refactoring.

error signalling is by design a slow path, and both `type()` and `assert()` are compiled away, so copying to locals doesn't buy anything.

kong/tools/stream_api.lua
```diff
@@ -0,0 +1,100 @@
...
+    return
+  end
+  if #data > 8000 then
```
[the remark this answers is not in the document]
javierguerragiraldez: yep, magic numbers are ugly.   i'm keeping it at 8000, because of framing, and using [8180-len(plugin_name)] for requests and 8182 for responses seems silly.

kong/tools/stream_api.lua
javierguerragiraldez: done

kong/tools/stream_api.lua
javierguerragiraldez: done

kong/tools/stream_api.lua
javierguerragiraldez: done

dndx (approved):

PR #6557, opened 2020-11-10
feat(*) utf-8 names for Routes and Services

kikito:
This PR adds utf-8 capabilities to the names of Routes and Services.

Some tests which assumed ASCII-only had to be modified.

Tests which check that utf-8 strings can be used in the Admin API have been added.

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -98,6 +98,18 @@ local function validate_name(name)
 end
 
 
+local function validate_utf8_name(name)
+
+  local ok, index = utils.validate_utf8(name)
+
+  if not ok then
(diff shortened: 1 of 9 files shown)


kong/db/schema/typedefs.lua
```diff
@@ -98,6 +98,18 @@ local function validate_name(name)
 end
 
 
+local function validate_utf8_name(name)
+
+  if not utils.validate_utf8(name) then
+    return nil,
+    "invalid value '" .. name ..
```
bungle, line 105: I am not sure if it is good idea to put `non-utf8 string` in `utf-8` string?

bungle, line 105: I am not sure if it is good idea to put `non-utf8 string` in `utf-8` string?

bungle (approved):

PR #6558, opened 2020-11-10
Modified copyright notice and permission notice

sreeharisj23:
The copyright notice and permission notice was modified to make it more readable.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX


Tieske:
sorry, but these are legal statements, so readability is of no concern here, might even be an anti-pattern 😄 . Better not change them.

PR #6567, opened 2020-11-11
fix(dao) properly trigger dao:delete_by:post hook

kidd:


kong/db/dao/init.lua
```diff
@@ -910,8 +911,9 @@ local function generate_foreign_key_methods(schema)
 
         entity, err_t = run_hook("dao:delete_by:post",
                                  entity,
-                                 "routes",
-                                 options)
+                                 self.schema.name,
+                                 options,
+                                 entity.ws_id)
```
gszr, line 916: Could ws_id be passed in `options`?

gszr (approved):
Left minor question

kong/db/dao/init.lua
kidd: yes, it could, but it is a 1st class parameter, and there's a pattern on passing it around like that, so I don't see a reason to change now.

- https://github.com/Kong/kong/pull/6567/files#diff-3f23ea4d6e5e3c79523ed39b7232329ecadccb4646d55a65f617f44552318785R864
- https://github.com/Kong/kong/pull/6567/files#diff-3f23ea4d6e5e3c79523ed39b7232329ecadccb4646d55a65f617f44552318785R1178
- ...

PR #6576, opened 2020-11-13
feat(clustering) generate and share unique cluster ID for CP and DP in Hybrid mode

dndx:
The rationale for adding the core entity type `Parameters` instead of explicitly for "Cluster ID" is that it can be used for storing other runtime configuration as well. That way in the future we can make more Kong config changeable with Admin API calls instead of `kong.conf`. This is especially important for Hybrid mode deployment patterns.

In `spec/helpers.lua`, `render_fixtures` function, `conf_loader` was
used to load from rendered `.kong_env`. However, because the test runner
sets environment variables such as `KONG_DATABASE=postgres` based on the
test matrix, `conf_loader` will attempt to override the config parsed
from `.kong_env` with values from the environment. Because `.kong_env` already
contains value that had been overwritten or transformed, it is a bug to
apply the override again when parsing it.


kong/pdk/cluster.lua
```diff
@@ -0,0 +1,85 @@
...
+  -- Returns the unique id for this Hybrid mode cluster. If Kong
+  -- is not running in Hybrid mode, then this method returns nil.
+  --
+  -- All Control Planes and Data Planes belonging to the same
+  -- cluster returns the same cluster ID.
+  --
+  -- @function kong.cluster.get_id
+  -- @treturn string The v4 UUID used by this cluster as its id
+  -- @usage
+  -- local id = kong.cluster.get_id()
```
hbagdi, line 63: These comments need updating. Cluster ID is a general purpose concept and now specific to hybrid mode.

kong/pdk/cluster.lua
```diff
@@ -0,0 +1,85 @@
...
+      local res, err = kong.db.parameters:select({ key = CLUSTER_ID_CACHE_KEY, })
+      if res then
+        return res.value
+      end
+
+      return nil, err
+    end
+
+    return kong.core_cache:get(CLUSTER_ID_CACHE_KEY, nil, fetch_cluster_id)
+  end
```
hbagdi, line 76: What happens if kong is running in the standalone DB-less mode?

kong/pdk/cluster.lua
```diff
@@ -0,0 +1,85 @@
...
+
+    res, err = kong.db.parameters:select({ key = CLUSTER_ID_CACHE_KEY, })
+    if err then
+      error(err)
+    end
+
+    if not res then
+      assert(kong.db.parameters:insert({ key = CLUSTER_ID_CACHE_KEY,
+                                         value = cluster_id, }))
+
```
hbagdi, line 36: Question for personal edification: why is the cluster ID generation done in lazily and not as part of migrations?

kong/pdk/cluster.lua
dndx: Yes I guess in this case we need to add a guard to return `nil` if user did not pre-define a cluster ID in the declarative config file. I will update this PR to check for that. Good catch!

kong/pdk/cluster.lua
dndx: Good point. The reason why it was generated lazily was because the first version of this PR only worked under Hybrid mode and we did not have needs to generate them for traditional deployment (wouldn't hurt either though). We can generate it in the migrations and it will reduce the logic in the PDK function by a lot. I will make these changes.

kong/pdk/cluster.lua
bungle: @dndx wouldn't it be better to then autogenerate the cluster id? I mean that with dbless the single node is the cluster.

dndx: @bungle I felt it is not necessary, since in that case we already have `node_id` which is autogenerated, generating another random cluster ID will make data analysis more involved later.

Tieske: > wouldn't it be better to then autogenerate the cluster id? I mean that with dbless the single node is the cluster.

In db less the user can set a cluster id in the config, because only the user knows if there are more nodes running the same config. Or alternatively we should assume there is my cluster, and hence return nil here. So I feel for @dndx s suggestion

bungle: @Tieske, I am fine with it, I proposed the auto-generation would just be fallback if not specified in config.  Why would `nil` be better than `unique id`. I cannot easily think upsides. Only downsides. E.g. nil is magic value, that we are now attaching meaning. It is also error prone in code if the code does not explicitly check for `nil`ness.

bungle: Offline we discussed also having `00000000-0000-0000-0000-000000000000` as the default (aka the `nil`), but it was decided to just use `nil`.

kong/pdk/cluster.lua
dndx: Fixed.

kong/pdk/cluster.lua
dndx: After discussing we will keep with the current design of returning `nil` when node is not running in a cluster. We can always revisit this later if needed.

kong/db/migrations/core/013_220_to_230.lua
```diff
@@ -0,0 +1,48 @@
...
+        key            TEXT PRIMARY KEY,
+        value          TEXT NOT NULL,
+        created_at     TIMESTAMP WITH TIME ZONE
+      );
+    ]],
+    teardown = function(connector)
+      local _, err = connector:query(string.format([[
+        INSERT INTO parameters (key, value) VALUES('%s', '%s')
+        ON CONFLICT DO NOTHING;
+      ]], "cluster_id", utils.uuid()))
```
hbagdi, line 17: Since this code runs in `teardown` phase, there can _potentially_ be a time during which `kong.cluster.get_id()` is called in a Kong 2.3 node, but the ID is not present in the database (nodes of the newer versions are allowed to startup even when `kong migrations finish` has not bee run). Can that create a potential problem?

dndx: Yes, the issue was that for some reason only the `teardown` handler actually supports a function. But now I rethink about it, we could generate a `up` SQL that contains a randomly generated UUID in it. I will make that change.

gszr (approved):

kong/db/migrations/core/013_220_to_230.lua
dndx: Fixed.

PR #6586, opened 2020-11-18
chore(*) prevent lua_pack from leaking to globals

bungle:
### Summary

`lua_pack` is especially bad as it monkey patches the `string` in Lua. This means that whenever you require the `lua_pack` the `string` is modified. We don't want this to happen, thus we have this special construct there. We should replace `lua_pack` or modify `lua_pack` at some point.

### DIFF
--- a/kong/plugins/ldap-auth/asn1.lua
+++ b/kong/plugins/ldap-auth/asn1.lua
@@ -1,14 +1,13 @@
 local bpack, bunpack
 do
-  local string_pack = string.pack
-  local string_unpack = string.unpack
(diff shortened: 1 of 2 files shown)


bungle:
A good example of this leakage is that with this PR the `grpc-web` doesn't anymore pass as it was depending on the leakage.

bungle:
Sister PRs that need to be merged prior this:
https://github.com/Kong/kong-plugin-grpc-gateway/pull/14
https://github.com/Kong/kong-plugin-grpc-web/pull/15

bungle:
Closing this as we agreed to fix this better.

PR #6592, opened 2020-11-20
Updated rules for bot-detection

fasalmbt:
added AWS S3 client , WebPageTest regex and updated rules for bot-detection


bungle:
@fasalmbt, thank you. May I ask how do you update these? Some script, or are they listed somewhere (just being curious here).

Tieske:
I assume an update from here: https://github.com/Kong/kong/pull/6592/files#diff-9ee4ff50fbcb499a46397b6635eec1ca4a553405123d83ba717d0c818f8be1e9R1

Correct @fasalmbt  ?

fasalmbt:
@Tieske yes

bungle:
@fasalmbt, this seems to do exactly same change as #5772, which mean this looks like a duplicate. One of these needs to go, so I decided to close this as there was some prior art. If I am wrong about it, please let me know.

bungle:
@fasalmbt the #5772 is now merged. Let us know if we missed something.

fasalmbt:
No.Everything is fine. Just wants to know will this PR be eligible for Tshirt or not.

PR #6593, opened 2020-11-20
fix(schema) more descriptive error logs when type is invalid

bungle:
### Summary

I screwed, sorry @WALL-E, I tried to rebase your original: https://github.com/Kong/kong/pull/6400, but then mistakenly pushes next branch to your next branch, while I was meaning to push the rebased one. Then it closed the PR and I could not push anymore to your branch, Thus here it is again. If this goes green I think we can merge it. Thank you @WALL-E  and sorry for inconvenience.


bungle:
One thing about this is that should we check if the input is valid `utf-8` before concatenting it.

kikito (approved):

PR #6594, opened 2020-11-20
fix(oauth2) handle more client invalid token generation cases

### LINKED ISSUE #6311: 2.1.3 OAuth2.0 CC stacktrace error from bad client input

bungle:
### Summary

@jeremyjpj0916, sorry I tried to push to your PR https://github.com/Kong/kong/pull/6337, but I just don't seem to handle `git` too well. And that got closed. I reopened it here with some small changes.

Fixes: https://github.com/Kong/kong/issues/6311

^^ Full discussion and examples over here. My changes fix 2 scenarios I can see Kong not responding as I would expect to an end user. 1 was type error fix(by validating a scenario where bad user input yields a boolean reference not a string), other requires thinking about it a little deeper to return the right response to clients. 

### Full changelog

* Add better handling in ```get_redirect_uris``` when fields are not present ```client_id``` ends up being a boolean parameter. Potentially fixes could have been done alternatively in the ```retrieve_parameters``` or ```retrieve_client_credentials``` functions.

* Removed the check for ```POST``` method to decide if we are calling the proxy vs calling the token generation pathings, URI alone should help dictate it imo and thus the response customers get from Kong will be more accurate to what proxy clients were attempting to do.  

* 2 edge cases tests I fixed added.

### Issues resolved

Fix #6311

### DIFF
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -15,6 +15,7 @@ local split = utils.split
 local strip = utils.strip
 local string_find = string.find
 local string_gsub = string.gsub
+local string_byte = string.byte
 local check_https = utils.check_https
 local encode_args = utils.encode_args
 local random_string = utils.random_string
@@ -31,6 +32,7 @@ local _M = {}
 
 
 local EMPTY = {}
+local SLASH = string_byte("/")
 local RESPONSE_TYPE = "response_type"
 local STATE = "state"
 local CODE = "code"
@@ -161,7 +163,7 @@ end
 
 local function get_redirect_uris(client_id)
   local client, err
-  if client_id and client_id ~= "" then
+  if type(client_id) == "string" and client_id ~= "" then
     local credential_cache_key = kong.db.oauth2_credentials:cache_key(client_id)
     client, err = kong.cache:get(credential_cache_key, nil,
                                  load_oauth2_credential_by_client_id,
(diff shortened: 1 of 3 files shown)


jeremyjpj0916:
Thanks for following up on my PR @bungle , glad to see the hardened logic making it in. I am on vacation now so not much time to look it over but at a glance makes sense and addresses my concerns!

jeremyjpj0916:
Hope this does not get lost among the Kong version iterations 😆  .

kikito (approved):

Issue #6591, opened 2020-11-20
Kong 2.2.0 rpm changes owning user from nobody to kong


nvx:
The rpm installation of Kong 2.1.4 uses the system nobody user as the owner for /usr/local/kong/proxy_temp, /usr/local/kong/logs, etc.

Kong 2.2.0 seems to have changed the ownership of these directories to a newly created kong user without mention in the changelog. This breaks existing kong.conf configurations where nginx_user is specified as nobody as the nobody user will now be broken.

Personally I think the change is good having a dedicated kong user for things to run as, but it really needs to be called out in the changelog as a breaking change so that people are aware of it and can update their configuration appropriately.

Kong RPM installed using the RHEL7 bintray repo at https://kong.bintray.com/kong-rpm/rhel/7
RPM version: kong-2.2.0-1.x86_64

nvx:
Unless I'm mistaken, it looks like the default kong.conf file still has nginx_user = nobody nobody. This probably should be updated too.

bungle:
I think you are correct:
https://github.com/Kong/kong/blob/master/kong.conf.default#L441
https://github.com/Kong/kong/blob/master/kong/templates/kong_defaults.lua#L53
https://github.com/Kong/kong/blob/master/kong/templates/kong_defaults.lua#L58

bungle:
Seems to be related to: https://github.com/Kong/kong/pull/6422

murillopaula:
Would you be able to describe how you are installing Kong and where things break for you?

> Kong 2.2.0 seems to have changed the ownership of these directories to a newly created kong user without mention in the changelog

Yes, we screwed up there. We should have mentioned it in the changelog.

> This breaks existing kong.conf configurations where nginx_user is specified as nobody as the nobody user will now be broken. 
>
> Personally I think the change is good having a dedicated kong user for things to run as, but it really needs to be called out in the changelog as a breaking change so that people are aware of it and can update their configuration appropriately.

`nginx_user` is set as `nobody` by default in kong.conf configurations. Before 2.2.0 Kong-owned directories such as `/usr/local/kong/*` were owned by `root` by default when you installed Kong on RHEL7 with the .rpm package. So things were already broken before 2.2.0 in that sense. So given that they were owned by `root` changing the ownership to `kong` should have been backwards compatible.

> Unless I'm mistaken, it looks like the default kong.conf file still has nginx_user = nobody nobody. This probably should be updated too.

[...]

PR #6595, opened 2020-11-20
docs(changelog) mention the non-root way of running kong

murillopaula:
Copy and pasted the description from the EE changelog https://docs.konghq.com/enterprise/changelog/#core

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,11 @@ This is a new major release of Kong, including new features such as **UDP suppor
 **Configurable Request and Response Buffering**, **Dynamically Loading of OS
 Certificates**, and much more.
 
+### Distributions
+
+- Added support for running Kong as the non-root user kong on distributed systems.
+
+
 ### Dependencies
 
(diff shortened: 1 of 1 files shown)


kikito (approved):

kikito (changes requested):
Feedback from user in https://github.com/Kong/kong/issues/6591#issuecomment-732903514: 

> I'd call it out a bit more explicitly that ownership of directories has changed from nobody to kong and that if your config currently has nginx user set to nobody you must update it to be kong

> The way it's worded at the moment doesn't really make it sound like it'd break existing configs.

> The default config needs updating too.

murillopaula:
@kikito IMHO we should still merge this as it is now. We forgot to include it in the 2.2.0 changelog.

I believe that In 2.2.2 we'll include the fix for the issue #6591 (see the PR #6421). So I'm not sure if it's needed to say in the 2.2.0 changelog there is a bug that was fixed in 2.2.2... 🤔

kikito (approved):

PR #6599, opened 2020-11-21
feat(cluster) enable access log on cluster listener

### LINKED ISSUE #6589: No logs for dp -> cp communications
The CP listener has no access logs configur-ed/able? see https://github.com/Kong/kong/blob/master/kong/templates/nginx_kong.lua#L440-L447

I was troubleshoting DP/CP connections and expected logs to help out, but there are none…

They would be helpful for setting things up. So we should probably write them to the same log files imo, so we don't have to use yet another set of config options. And volume is really low, so that shouldn't be an issue.

bungle:
### Summary

Enables access logs on cluster listener and writes logs to admin access log.

### Issues Resolved

Fix #6589

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -48,7 +48,14 @@
                                           # property.
 
 #admin_access_log = logs/admin_access.log # Path for Admin API request access
-                                          # logs. Set this value to `off` to
+                                          # logs. If Hybrid Mode is enabled
+                                          # and the current node is set to be
+                                          # the Control Plane, then the
+                                          # connection requests from Data Planes
+                                          # are also written to this file with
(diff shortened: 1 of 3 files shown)


Tieske (commented):
lgtm

Tieske:
minor note: probably in the config file we should make a comment at the cluster_listen listener that the logs will be in admin access/error log files, to make sure we document this

dndx (approved):

dndx:
@Tieske this was added in https://github.com/Kong/kong/pull/6599/commits/58e453cdb7ab264a24cf073b63691a1022473dea.

Tieske:
Thnx great doc comments ❤️

PR #6600, opened 2020-11-21
generic external plugins

javierguerragiraldez:
Replace the concept of "Go plugins" with "external plugins", which live in external processes and communicate via RPC.


kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+  return false, "no plugin found"
+end
+
+
+--[[
+
+Process management
+
+Servers that specify an `.exec` field are launched and managed by Kong.
+This is an attempt to duplicate the smallest reasonable subset of systemd.
```
kikito, line 567: "Manage" is defined in an indirect way using systemd, which people reading this might not be familiar with. I the word "Manage" does not support its own weight here. I think we can explain what's happening without mentioning "Manage" or systemd at all. Something like (please correct, and note that I am still unsure about the difference between `exec` and `start_command`):

> If a Plugin Server has an `.exec` field, then Kong will monitor its processes, relaunching them if they die:
> * Restarting it using its `.exec`, `arguments` and environment fields.
> * Stdout and stderr are joined and logged

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
+local cjson = require "cjson.safe"
```
kikito, line 1: I think this file should be outside of `kong/db` and its filename should contain the word `server` on it. I suggest `kong/runloop/plugin_servers/init.lua`

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+--[[
+
+Configuration
+
+The external_plugins_config YAML file defines a list of plugin servers.  Each one
+can have the following fields:
+
+name: (required) a unique string.  Shouldn't collide with Lua packages available as `kong.plugins.<name>`
+socket: (required) path of a unix domain socket to use for RPC.
+exec: (optional) executable file of the server.  If omitted, the server process won't be managed by Kong.
```
kikito, line 42: What is the difference between `exec` and `start_command`?

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+        server_def.query_command, infos_dump))
+    return
+  end
+
+  for _, plugin_info in ipairs(infos) do
+    register_plugin_info(server_def, plugin_info)
+  end
+end
+
+local function load_all_infos()
```
kikito, line 519: Since it is used only once, I suggest renaming this to `get_plugin_info(plugin_name)` and deal with the memoization of `_plugin_infos` completely inside this function

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+    for _, server_def in ipairs(get_server_defs()) do
+      ask_info(server_def)
+    end
+  end
+
+  return _plugin_infos
+end
+
+
+local loaded_plugins = {}
```
kikito, line 532: Why do we need a separate `loaded_plugins` variable? It seems it could be merged with `_plugin_infos` (and perhaps read the phases as part of it)?

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+  local plugin = get_plugin(plugin_name)
+  if plugin and plugin.PRIORITY then
+    return true, plugin.schema
+  end
+
+  return false, "no plugin found"
+end
+
+
+--[[
```
kikito, line 562: It seems like this file is doing two different things:
* plugin_info related stuff
* process tracking/restarting stuff

It seems to me that it could be neatly split into two files.

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+
+Process management
+
+Servers that specify an `.exec` field are launched and managed by Kong.
```
kikito, line 566: It seems that _Plugin Servers_ are often abbreviated as _Servers_ - not only here, but also in method names, variables, etc. I would suggest not doing that, it can be challenging to know if we have two concepts of just one.

kong/db/dao/plugins/external.lua
kikito: I would also explicitly mention what happens if `.exec` isn't there. Example:

> If a Plugin server has no `.exec` field, it is assumed that this monitorization happens somewhere else. Kong will not try to restart its process, nor log anything if it dies.

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+        while true do
+          grab_logs(server_def.proc, server_def.name)
+          local ok, reason, status = server_def.proc:wait()
+          if ok ~= nil or reason == "exited" then
+            kong.log.notice("external pluginserver '", server_def.name, "' terminated: ", tostring(reason), " ", tostring(status))
+            break
+          end
+        end
+      end
+      kong.log.notice("Exiting: go-pluginserver not respawned.")
```
kikito, line 621: Probably a copy-paste typo:

```suggestion
      kong.log.notice("Exiting: plugin server not respawned.")
```

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+  end
+end
+
+local function handle_server(server_def)
+  if not server_def.socket then
+    -- no error, just ignore
+    return
+  end
+
+  if server_def.start_command then
```
kikito, line 597: Kong style is doing early exits:

```
if not server_def.start_command then
  return server_def
end
-- asume server_def.start_command from now on
```

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+
+function external_plugins.manage_servers()
+  if ngx.worker.id() ~= 0 then
+    kong.log.notice("only worker #0 can manage")
+    return
+  end
+
+  for i, server_def in ipairs(get_server_defs()) do
+    local server, err = handle_server(server_def)
+    if not server then
```
kikito, line 636: `handle_server` can currently return `nil` and no `err` (when `server_def.socket` is `nil`)

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+end
+
+local function handle_server(server_def)
+  if not server_def.socket then
+    -- no error, just ignore
+    return
+  end
+
+  if server_def.start_command then
+    ngx_timer_at(0, function(premature)
```
kikito, line 598: I suggest taking this function out to a separate function, adding the `server_def` as parameter:

```
function plugin_server_timer(premature, server_def)
  if premature then
  etc etc
```

Then the preambles are short enough that `handle_server` is not needed any more and can be put in the loop inside `external_plugins.manage_servers`

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+        end
+      end
+      kong.log.notice("Exiting: go-pluginserver not respawned.")
+    end)
+  end
+
+  return server_def
+end
+
+function external_plugins.manage_servers()
```
kikito, line 628: Can we rename this to something like `start_servers()`, or even just `plugin_servers.start()`?

kong/db/dao/plugins/mp_rpc.lua
```diff
@@ -0,0 +1,123 @@
+local msgpack = require "MessagePack"
```
kikito, line 1: Move this file out of `kong/db` as well, I suggest inside `kong/runloop/plugin_servers`

kong.conf.default
```diff
@@ -122,15 +122,19 @@
...
-
-#go_plugins_dir = off            # Directory for installing Kong plugins
-                                 # written in Go.
-                                 #
-                                 # This value can be set to `off`, thus disabling
-                                 # the plugin server and Go plugin loading.
+#pluginserver_names =            # Comma-separated list of names for pluginserver
+                                 # processes.  The actual names are used for
+                                 # log messages and to relate the following settings:
+    #pluginserver_XXX_socket = <prefix>/<XXX>.socket            # path to the unix socket
```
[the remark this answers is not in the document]
javierguerragiraldez: :+1: 
how can I preview the generated docs?

kong/db/dao/plugins/external.lua
javierguerragiraldez: scar tissue from the yaml->config change

kong/db/dao/plugins/external.lua
```diff
@@ -0,0 +1,642 @@
...
+    -- there already was something, make it evident that we're changing it
+    instance_info.id = nil
+  end
+
+  local plugin_info = _plugin_infos[plugin_name]
+  local server_rpc  = get_server_rpc(plugin_info.server_def)
+
+  local status, err = server_rpc:call("plugin.StartInstance", {
+    Name = plugin_name,
+    Config = cjson_encode(conf)
```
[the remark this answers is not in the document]
javierguerragiraldez: compatibility with the current protocol

kong.conf.default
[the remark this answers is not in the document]
kikito: `./scripts/autodoc`

kong/db/dao/plugins/external.lua
javierguerragiraldez: `_plugin_infos` contains the info of all available plugins, as answered by the pluginservers.  `loaded_plugins` contains all the plugins loaded, in runnable state (with phase closures).  might be merged into one, but i found easier to reason about with the split between "cached external data" and "active object".

i'm also semi-foggily thinking about eventually moving to a more dynamic lifecycle, where plugins could be added/removed at runtime.  these will have to be kept in sync, which could be tricky; but at the same time it might be important to differentiate between the new info from the external pluginserver and the still-running plugin.   hard to say if merging would simplify or complicate.

kong/db/dao/plugins/external.lua
javierguerragiraldez: yes, two things.  and the exact state of how much it has been loaded during different parts of startup was tricky to get right.  i'll see if splitting in two helps making that process more readable.

kong/db/dao/plugins/external.lua
javierguerragiraldez: there's no `.exec` anymore (nor `.arguments` or `.environment`)  they all were part of the yaml file.  i have to rewrite these comments completely.  not keeping double negatives isn't bad advice. :wink:

kong/db/dao/plugins/external.lua
javierguerragiraldez: :+1:

kong/db/dao/plugins/external.lua
javierguerragiraldez: removed that case since `.socket` can no longer be `nil` because of the default.  in fact, `handle_server()` itself shouldn't fail, since all it does is to spawn a timer.

kong/db/dao/plugins/external.lua
javierguerragiraldez: done

kong/db/dao/plugins/external.lua
javierguerragiraldez: done

kong/db/dao/plugins/external.lua
javierguerragiraldez: done

kong/db/dao/plugins/external.lua
javierguerragiraldez: hum, i did a few, but i think `sever_def` will stay

kong/runloop/plugin_servers/init.lua
```diff
@@ -0,0 +1,430 @@
...
+    end
+  end
+
+  return plugin
+end
+
+
+
+--- module table
+local external_plugins = {}
```
kikito, line 382: Given that the filename is `plugin_servers`, perhaps this should be called `plugin_servers` as well?

kong/runloop/plugin_servers/mp_rpc.lua
```diff
@@ -0,0 +1,123 @@
...
+
+  if empty then
+    return mp_empty_map
+  end
+
+  return o
+end
+
+
+
```
kikito, line 62: Too many blank lines
```suggestion
```

kong/templates/kong_defaults.lua
```diff
@@ -8,6 +8,7 @@ admin_error_log = logs/error.log
 status_access_log = off
 status_error_log = logs/status_error.log
 plugins = bundled
+external_plugins_config = NONE
```
kikito, line 11: Is this correct? Nothing seems to be using `external_plugins_config`

kikito (approved):

gszr (approved):

PR #6603, opened 2020-11-23
fix(runloop) grpc upstream :authority header

gszr:
This PR leverages the new API introduced with https://github.com/Kong/lua-kong-nginx-module/pull/17 to set the :authority pseudo-header for gRPC requests.

Honor route attribute `preserve_host`:
- If `false`, Kong will pass `localhost:port` to the upstream gRPC
  service - only if port is not `80` or `443`
- If `true`, Kong will pass the downstream `Host` to the upstream
  gRPC service

This matches existing Kong behavior for non-gRPC services.


dndx:
Looks nice, maybe we should add a simple test case to ensure this actually works as intended in Kong.

https://github.com/Kong/lua-kong-nginx-module/pull/17/files#diff-d147685da713824a461c7d9aa403bb962f3e1e29d10f02b17b3d80b55513181b can be used as a reference, we don't actually need a real gRPC upstream for this, a Nginx `server` listening for HTTP/2 requests should be enough.

gszr:
> Looks nice, maybe we should add a simple test case to ensure this actually works as intended in Kong.
> 
> https://github.com/Kong/lua-kong-nginx-module/pull/17/files#diff-d147685da713824a461c7d9aa403bb962f3e1e29d10f02b17b3d80b55513181b can be used as a reference, we don't actually need a real gRPC upstream for this, a Nginx `server` listening for HTTP/2 requests should be enough.

I don't think the changes here are testable with an http/2 nginx server -- changes are deep inside the runloop handler and rely on the grpc scheme. Could you elaborate, @dndx?

dndx:
@gszr if that is the case, is it testable with grpcbin like the other tests for gRPC?

gszr:
> @gszr if that is the case, is it testable with grpcbin like the other tests for gRPC?

Similar tests use mockbin's echo endpoint to check on what headers were sent upstream, which isn't possible with mockbin. I do have plans to work on adding that -- or similar -- to mockbin, as it'd help not only here but in some plugins tests (grpc compability tests) as well.

mflendrich:
> I don't think the changes here are testable with an http/2 nginx server -- changes are deep inside the runloop handler and rely on the grpc scheme

Maybe I'm missing something (and I don't know much about your testing setup), but:

- assuming Kong's outgoing gRPC requests to be HTTP/2 POST requests,
- assuming that the desired result of this PR is to include the :authority pseudo-header on all gRPC (and hence HTTP/2) requests
- assuming that the test needs to assert on the request headers only - request body, response and Kong's behavior is irrelevant

why wouldn't the following test setup work?

- spin up an arbitrary HTTP/2 server (let's call it X) capable of capturing the :authority header of an incoming request, and responding with anything (even HTTP 501)
- spin up Kong and define a gRPC route pointing to the HTTP/2 port of X
- make a gRPC call to Kong
- read the :authority header that X recorded for the request it got, and assert

gszr:
> > I don't think the changes here are testable with an http/2 nginx server -- changes are deep inside the runloop handler and rely on the grpc scheme
> 
> Maybe I'm missing something (and I don't know much about your testing setup), but:
> 
>     * assuming Kong's outgoing gRPC requests to be HTTP/2 POST requests,
> 
>     * assuming that the desired result of this PR is to include the :authority pseudo-header on all gRPC (and hence HTTP/2) requests
> 
>     * assuming that the test needs to assert on the request headers only - request body, response and Kong's behavior is irrelevant
> 
> 
> why wouldn't the following test setup work?
> 
>     * spin up an arbitrary HTTP/2 server (let's call it X) capable of capturing the :authority header of an incoming request, and responding with anything (even HTTP 501)
> 
>     * spin up Kong and define a gRPC route pointing to the HTTP/2 port of X
> 
>     * make a gRPC call to Kong
> 
>     * read the :authority header that X recorded for the request it got, and assert

Working on a setup similar to what you described! Stay tuned!

kong/runloop/handler.lua
```diff
@@ -1309,13 +1309,24 @@ return {
...
             upstream_host = upstream_host .. ":" .. balancer_data.port
           end
 
           var.upstream_host = upstream_host
         end
+
+        -- the nginx grpc module does not offer a way to overrride
+        -- the :authority pseudo-header; use our internal API to
+        -- do so
+        local resty_grpc = require "resty.kong.grpc"
```
dndx, line 1325: We should add a simple new PDK function that wraps around the module API (e.g. https://github.com/Kong/kong/blob/master/kong/pdk/service.lua#L136).

In fact, this file currently contains a few calls to the `resty.kong` Lua APIs, but they are all wrapped by PDKs.

kong/runloop/handler.lua
```diff
@@ -1309,13 +1309,24 @@ return {
...
+        local ok, err = resty_grpc.set_authority(upstream_host)
```
dndx, line 1326: We should check to only call `set_authority` when the incoming request is indeed gRPC request. This call does involves a memory allocation that can not be freed immediately.

dndx (commented):
Minor styling comments, but the change and tests LGTM in general.

kong/pdk/service/request.lua
```diff
@@ -284,6 +284,14 @@ local function new(self)
       ngx.var.upstream_host = value
     end
 
+    if string_lower(header) == ":authority"
+       and (ngx.var.upstream_scheme == "grpc" or
+           ngx.var.upstream_scheme == "grpcs") then
+
+      local resty_grpc = require "resty.kong.grpc"
+      return resty_grpc.set_authority(value)
```
dndx, line 292: This function can be cached at line 263 above, as so we do not load it every time the function is used.

kong/pdk/service/request.lua
```diff
@@ -284,6 +284,14 @@ local function new(self)
       ngx.var.upstream_host = value
     end
 
+    if string_lower(header) == ":authority"
+       and (ngx.var.upstream_scheme == "grpc" or
```
dndx, line 288: Same, `ngx.var` can be safely cached as well. `local ngx_var = ngx.var`.

Question, if request is not gRPC and user attempts to set `:authority`, should we return an error instead? AFAIK there is no reason to set a plain-text header named `:authority` for non gRPC requests.

kong/pdk/service/request.lua
```diff
@@ -261,6 +262,8 @@ local function new(self)
   end
 
 
+  local resty_grpc = require "resty.kong.grpc"
```
dndx, line 265:
```suggestion
  local set_authority = require("resty.kong.grpc").set_authority
```

kong/pdk/service/request.lua
```diff
@@ -281,7 +284,18 @@ local function new(self)
...
     if string_lower(header) == "host" then
-      ngx.var.upstream_host = value
+      ngx_var.upstream_host = value
+    end
+
+    if string_lower(header) == ":authority" then
+      if ngx_var.upstream_scheme == "grpc" or
+         ngx_var.upstream_scheme == "grpcs" then
+
+        return resty_grpc.set_authority(value)
```
dndx, line 294:
```suggestion
        return set_authority(value)
```

dndx (approved):

dndx:
Test needs fixing. @gszr working on it.

PR #6604, opened 2020-11-23
chore(ci) build grpcurl

gszr:
- build grpcurl from a given sha / tag
- cherry-pick bot detection tests that rely on a new grpcurl feature - closes https://github.com/Kong/kong/pull/6252


spec/03-plugins/01-tcp-log/01-tcp-log_spec.lua
```diff
@@ -435,7 +435,7 @@ for _, strategy in helpers.each_strategy() do
       local log_message = cjson.decode(res)
 
       assert.equal("grpcs", log_message.service.protocol)
-      assert.equal("TLSv1.2", log_message.request.tls.version)
+      assert.equal("TLSv1.3", log_message.request.tls.version)
```
gszr, line 438: unsure why this is needed, but i suspect go / go grpc changed its default tls version

Makefile
```diff
@@ -122,10 +123,16 @@ dependencies: bin/grpcurl
 	  fi \
 	done;
 
+tmp_dir=$(shell mktemp -d)
 bin/grpcurl:
```
dndx, line 127: Will this lack of dependency causing `bin/grpcurl` to be rebuilt every time `make dev` is run? I knew it did that for the old `curl` downloaded version and I had to comment it out every time to speed up development.

gszr: I won't - notice the target name is the file name.

dndx (approved):

Makefile
dndx: Tested locally and confirmed :)

gszr: Thanks @dndx!

PR #6608, opened 2020-11-27
feat(migrations) allow up_f field in Cassandra migrations

kikito:
The initial assumption behind only allowing strings of SQL and CQL on the "up" side of migrations was that the operations on this phase needed to be safe, fast and reentrant.

Unfortunately CQL is much more limited than Pg's SQL. So what's trivial to do in Pg is impossible to do in Cassandra. So we arrive to a point where trivial and safe operations get delayed to the `teardown` phase, simply because they need a more complete language than CQL to be done.

This change adds a new `up_f` method to which accepts a Lua function with a connector, effectively allowing the use of Lua inside Cassandra migrations' "up" phase. While adding the same features in Postgres would not be difficult, we think that SQL is powerful enough to do most "safe" things we need (perhaps with some trivial string replacement from Lua).

A migration can have both up and up_f. In that case, the up string is run first, and up_f immediately afterwards.

Incidentally we have also changed the validations on the migrations schema so that `up` isn't always required. This validation has forced us to set it to "" (empty string) in the past, which is not ideal.

### DIFF
--- a/kong/db/init.lua
+++ b/kong/db/init.lua
@@ -547,12 +547,25 @@ do
         if run_up then
           -- kong migrations bootstrap
           -- kong migrations up
-          ok, err = self.connector:run_up_migration(mig.name,
-                                                    strategy_migration.up)
-          if not ok then
-            self.connector:close()
-            return nil, fmt_err(self, "failed to run migration '%s' up: %s",
-                                mig.name, err)
+          if strategy_migration.up and strategy_migration.up ~= "" then
+            ok, err = self.connector:run_up_migration(mig.name,
+                                                      strategy_migration.up)
+            if not ok then
+              self.connector:close()
+              return nil, fmt_err(self, "failed to run migration '%s' up: %s",
+                                  mig.name, err)
+            end
+          end
+
+          if strategy_migration.up_f then
+            local pok, perr, err = xpcall(strategy_migration.up_f,
+                                          debug.traceback, self.connector,
+                                          mig_helpers)
+            if not pok or err then
+              self.connector:close()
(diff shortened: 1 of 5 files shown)


kong/db/init.lua
```diff
@@ -547,12 +547,24 @@ do
...
+                                                      strategy_migration.up)
+            if not ok then
+              self.connector:close()
+              return nil, fmt_err(self, "failed to run migration '%s' up: %s",
+                                  mig.name, err)
+            end
+          end
+
+          if strategy_migration.up_f then
+            ok, err = self.connector:run_up_migration(mig.name,
```
dndx, line 561: I see in https://github.com/Kong/kong/pull/6608/files#diff-13a8c7acbd78b8151e140a4d391bd6f5639f126ca5d6c2501b214af616b69e1bR603-R604

we simply called

```lua
xpcall(f, debug.traceback, self.connector, mig_helpers)
```

for executing the `teardown` callbacks. Can we do the same for `strategy_migration.up_f`? this seems to avoid the special handling needed inside the connector code.

dndx (commented):
Recommends we should also add some simple test cases inside:

`spec/01-unit/01-db/01-schema/10-migrations_spec.lua` for migrations schema and,

`spec/fixtures/custom_plugins/kong/plugins/with-migrations/migrations/`, `spec/02-integration/02-cmd/10-migrations_spec.lua` for testing the execution of the newly added callback.

@kikito what do you think?

kong/db/init.lua
kikito: Good point! I extracted the `pcall` to `kong/db/init.lua`. The connector has no changes now, looks much better.

dndx (approved):

PR #6610, opened 2020-11-27
Nginx proxy testbed

locao:
This PR removes the not so reliable `http_server` and adds a testbed that uses Nginx.


kidd:
It seems it will be fine but, do you think this will have any impact on EE tests?

Tieske:
is there a reason to use another nginx instance and not the one in Kong itself using the server block injection?

for example: https://github.com/Kong/kong-plugin-aws-lambda/blob/master/spec/plugins/aws-lambda/99-access_spec.lua#L14-L92

locao:
> It seems it will be fine but, do you think this will have any impact on EE tests?

It probably will have, yes. Every test that uses the `http_server` from `spec.fixtures.balancer_utils` needs to be changed to use this one. I will take a look at how many tests would have to be changed, but it's a simple change anyway.

spec/fixtures/https_server.lua
```diff
@@ -0,0 +1,235 @@
+local https_server = {}
+https_server.__index = https_server
+
+
+local fmt = string.format
+local nginx_tpl_file = require 'spec.fixtures.nginx_conf_template'
```
dndx, line 6: I wonder if another name like "spec.fixtures.balancer_nginx_conf_template" is better, because this name on the first look seems like a Kong template and may be confusing.

dndx (approved):
Good job! Left a minor comment, but this PR is one step towards the right direction and I will not hesitate to merge it (compared to what we had before).

Also, could you answer @Tieske 's above comment on running the mock upstream inside the Kong nginx process as well?

spec/fixtures/https_server.lua
```diff
@@ -0,0 +1,235 @@
...
+local ngx = require 'ngx'
```
kikito, line 7: Use double-quotes instead of single quotes, everywhere. 

https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#strings

kikito:
My main concern is the amount of tests that need to be changed.

Would it be possible to do the same with less tests changed (perhaps adding a small compatibility layer on `balancer_utils`?

spec/fixtures/https_server.lua
locao: > is there a reason to use another nginx instance and not the one in Kong itself using the server block injection?

I decided to use a different Nginx instance for some reasons:
- can be started and finished independently, easier to simulate failures.
- has a different error.log that can be easily parsed.
- we are sure it will never cause any side effects to the Kong instance.

None of these is a strong reason, and I can change that if you think it would be better to reuse the same instance. (@Tieske @dndx)

Tieske: > can be started and finished independently

is actually a good reason.

Regarding the API to use this thing, that should be exposed through `spec/helpers.lua` that's where all the generic test stuff resides (and is documented).

locao: Good catch! I will change that.

dndx:
@kidd @kikito I checked the EE repo, there is only a single file from the EE test suite that uses the `spec.fixtures.balancer_utils`, so I suspect it will be a rather straightforward search and substitution to change everything there as well. Therefore, the change in API interface should not cause significant headache.

PR #6612, opened 2020-11-30
feat(clustering) version checks between CP and DP

dndx:
Previously, Hybrid mode did not have any capabilities for checking the
version compatibility between CP and DP (aside from the
`_format_version` field in the declarative config generated) and this
will cause some operational safety issues when people are running a
cluster of diverse Kong versions as Kong CP may be pushing out configs
DP could not understand without knowing.

After this commit, DP node will start reporting their Kong version
running to the CP. CP will then store this information onto the
database in the new `version` field of the `/clustering/data-planes`
Admin API endpoint.

DPs will also report the installed plugins and their version upon
connecting to the CP.

CP uses the Kong version and plugin information reported by DP to make
decision on whether new config should be pushed. Under the following
circumstance, sync will be stopped:

1. If the major minor version between CP and DP differs (note bugfix
version is not used for checking), or
2. If the installed plugins, or their version between CP and DP differs

Also some of the test fixtures needs upgrading to be compatible with
the 2.x schema format. This is because in order to produce the correct
plugin versions table, the busted runner needs to load all the plugins
(including those from the `spec/fixtures` directory) like the spawned CP does.

### DIFF
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -11,6 +11,7 @@ local utils = require("kong.tools.utils")
 local openssl_x509 = require("resty.openssl.x509")
 local system_constants = require("lua_system_constants")
 local ffi = require("ffi")
+local pl_tablex = require("pl.tablex")
 local string = string
 local assert = assert
 local setmetatable = setmetatable
@@ -33,8 +34,10 @@ local table_insert = table.insert
 local table_remove = table.remove
 local inflate_gzip = utils.inflate_gzip
 local deflate_gzip = utils.deflate_gzip
+local pl_tablex_compare = pl_tablex.compare
 
 
+local KONG_VERSION = kong.version
 local MAX_PAYLOAD = 4 * 1024 * 1024 -- 4MB
 local PING_INTERVAL = 30 -- 30 seconds
 local PING_WAIT = PING_INTERVAL * 1.5
@@ -51,6 +54,7 @@ local ngx_NOTICE = ngx.NOTICE
 local WEAK_KEY_MT = { __mode = "k", }
 local CERT_DIGEST
 local CERT, CERT_KEY
+local PLUGINS_LIST
 local clients = setmetatable({}, WEAK_KEY_MT)
 local prefix = ngx.config.prefix()
 local CONFIG_CACHE = prefix .. "/config.cache.json.gz"
@@ -137,7 +141,9 @@ local function communicate(premature, conf)
 
   local c = assert(ws_client:new(WS_OPTS))
   local uri = "wss://" .. address .. "/v1/outlet?node_id=" ..
-              kong.node.get_id() .. "&node_hostname=" .. kong.node.get_hostname()
+              kong.node.get_id() ..
+              "&node_hostname=" .. kong.node.get_hostname() ..
+              "&node_version=" .. KONG_VERSION
 
   local opts = {
     ssl_verify = true,
@@ -164,6 +170,20 @@ local function communicate(premature, conf)
   end
 
   -- connection established
+  -- first, send out the plugin list to CP so it can make decision on whether
+  -- sync will be allowed later
+  local _
+  _, err = c:send_binary(cjson_encode({ type = "basic_info",
+                                        plugins = PLUGINS_LIST, }))
+  if err then
+    ngx_log(ngx_ERR, "unable to send basic information to control plane: ", uri,
+                     " err: ", err,
+                     " (retrying after ", reconnection_delay, " seconds)")
+
+    c:close()
+    assert(ngx.timer.at(reconnection_delay, communicate, conf))
+    return
+  end
 
   local config_semaphore = semaphore.new(0)
 
@@ -312,6 +332,39 @@ local function validate_shared_cert()
 end
 
 
+local function plugin_entry_comparator(a, b)
+  return a.name == b.name and a.version == b.version
+end
+
+
+local MAJOR_MINOR_PATTERN = "^(%d+%.%d+)%.%d+"
+local function should_send_config_update(node_version, node_plugins)
+  if not node_version or not node_plugins then
+    return false, "your DP did not provide version information to the CP, " ..
+                  "Kong CP after 2.3 requires such information in order to " ..
+                  "ensure generated config is compatible with DPs. " ..
+                  "Sync is suspended for this DP and will resume " ..
+                  "automatically once this DP also upgrades to 2.3 or later"
+  end
+
+  local minor_cp = KONG_VERSION:match(MAJOR_MINOR_PATTERN)
+  local minor_node = node_version:match(MAJOR_MINOR_PATTERN)
+  if minor_cp ~= minor_node then
+    return false, "version mismatches, CP version: " .. minor_cp ..
+                  " DP version: " .. minor_node
+  end
+
+  if not pl_tablex_compare(PLUGINS_LIST, node_plugins,
+                           plugin_entry_comparator)
+  then
+    return false, "CP and DP does not have same set of plugins installed" ..
+                  " or their versions might differ"
+  end
+
+  return true
+end
+
+
 function _M.handle_cp_websocket()
   -- use mutual TLS authentication
   if kong.configuration.cluster_mtls == "shared" then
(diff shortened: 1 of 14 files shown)


kong/clustering.lua
```diff
@@ -471,6 +494,13 @@ function _M.handle_cp_websocket()
           else
             ngx_log(ngx_DEBUG, "sent config update to node")
           end
+
+        else -- incompatible version
+          ngx_log(ngx_WARN, "unable to send updated configuration to " ..
+                            "DP node with hostname = " .. node_hostname ..
+                            " ip = " .. node_ip ..
+                            " node version = " .. node_version ..
+                            " our version = " ..  KONG_VERSION)
```
hbagdi, line 503: required version or control plane version instead of our version

kong/db/schema/typedefs.lua
```diff
@@ -525,6 +525,20 @@ typedefs.headers = Schema.define {
...
+    patterns = { "^%d+[%.%d]*$", "^%d+[%.%d]*%-?.*$", },
+    err = "invalid version number: must be in format of X.Y.Z",
+  },
+  match_none = {
+    {
+      pattern = "%.%.",
+      err = "must not have empty version segments"
+    },
+  },
+}
```
hbagdi, line 540: 1. semantic_version might be a better name for this field.
2. Please ensure that this works for Kong Enterprise. There is a literal after the patch number for Kong Enterprise.

kong/templates/kong_defaults.lua
```diff
@@ -26,6 +26,7 @@ cluster_mtls = shared
 cluster_ca_cert = NONE
 cluster_server_name = NONE
 cluster_data_plane_purge_delay = 1209600
+cluster_version_check = exact
```
hbagdi, line 29: What is the reasoning behind this default? This makes upgrades a lot harder. `major_minor` seems a sane default for most cases.

kong/conf_loader/init.lua
```diff
@@ -610,6 +610,7 @@ local CONF_INFERENCES = {
   cluster_ca_cert = { typ = "string" },
   cluster_server_name = { typ = "string" },
   cluster_data_plane_purge_delay = { typ = "number" },
+  cluster_version_check = { enum = { "exact", "major_minor", }, },
```
hbagdi, line 613: This doesn't match up with the diff in `kong.conf.default`.

hbagdi (changes requested):

hbagdi:
With this patch, the data-plane node will silently fail and can easily frustrate users. Can we please degrade the health of the data-plane node to unhealthy if it is not receiving configuration updates?

kong/clustering.lua
```diff
@@ -459,7 +481,8 @@ function _M.handle_cp_websocket()
             ngx_log(ngx_DEBUG, "sent PONG packet to data plane")
           end
 
-        else -- config update
+        elseif should_send_config_update(KONG_VERSION, node_version) then
+          -- config update
```
hbagdi, line 485: We should make this communication significantly more mature.
Can the control plane send the reason for not sending configuration to the data plane?

kong/db/schema/typedefs.lua
dndx: Name was changed. This does work with suffix, https://github.com/Kong/kong/pull/6612/files#diff-97419d4c3d96f2fce4c8bcf5d1222b6a8e55bb38efc91855b8d7c7311a8cd042R152 proves it.

kong/clustering.lua
```diff
@@ -627,6 +657,17 @@ end
 function _M.init_worker(conf)
   assert(conf, "conf can not be nil", 2)
 
+  local plguins = assert(kong.db.plugins:get_handlers())
```
bungle, line 660: Is this a typo?

kong/clustering.lua
```diff
@@ -627,6 +657,17 @@ end
...
+  table.sort(plguins, function(a, b)
+    return a.name:lower() < b.name:lower()
+  end)
+
+  local sha256 = require("resty.sha256"):new()
+  for _, p in ipairs(plguins) do
+    assert(sha256:update(p.name .. (p.handler.VERSION or "0.0.0")))
```
bungle, line 667: I have seen many times that we forgot to bump version or have bumped it wrong according to semantic versioning, and it might be even more common with external plugins.

Also it seems like we are combining everything in same hash. I am not sure how we can find the output of such hash telling us whether the Kong is compatible or not. It just says that this is the hash of plugins, which may contain external plugins too? Is it just to check if data planes have same hash or different?

I am wondering if hash is good enough, what about just reporting things as JSON:

```
{
  "<plugin-name>": "<plugin-version>",
  ...
}
```

We can make the hash out of the whole thing, and store it to `version_hash` or something, but I think it might be beneficial to see what the versions really are, and not just some hash that does not have any meaning other than being random string.

Also `"0.0.0"`  feels like an unnecessary magic value, `nil`/`null`/ `"unknown"` would be better.

bungle (changes requested):
I have not read design document, but these came to my mind at first.

kong/db/schema/typedefs.lua
hbagdi: That's not the correct test case. Enterprise versions are of the form  `1.2.3.4-enterprise-edition`

dndx: The test case has been fixed.

kong/clustering.lua
```diff
@@ -33,8 +34,10 @@ local table_insert = table.insert
 local table_remove = table.remove
 local inflate_gzip = utils.inflate_gzip
 local deflate_gzip = utils.deflate_gzip
+local pl_tablex_compare = pl_tablex.compare
 
 
+local KONG_VERSION = require("kong.meta")._VERSION
```
bungle, line 40: Does: `_KONG._VERSION` work? I have seen some code use that and some require "kong.meta", not a big deal.

kong/clustering.lua
```diff
@@ -333,6 +385,20 @@ function _M.handle_cp_websocket()
...
+  -- receive basic_info
+  local data, typ
+  data, typ, err = wb:recv_frame()
+  if err then
+    ngx_log(ngx_ERR, "failed to receive WebSocket basic_info frame: ", err)
+    wb:close()
+    return ngx_exit(444)
+
+  elseif typ == "binary" then
+    ngx.log(ngx.ERR, "here: ", data)
```
bungle, line 397: I guess this is for debugging why tests don't pass?

bungle (commented):
Just some small comments.

kong/clustering.lua
dndx: Discussed with Harry and we will do this after the 2.3 alpha but before final in an interest of time here.

dndx:
Despite a lot of newly added automated test cases in the PR, I also did some manual sanity tests just to make sure everything still works:

1. Spin up CP and DP of same version
2. Add a `Service` and `Route`, sync still works and DP sees the config, request works normally through the DP
3. Restart the DP with `KONG_PLUGINS=acl`, `DELETE` the route, request still works through the DP. Also, CP has the following log lines:
```
2020/12/04 06:34:26 [warn] 4144#0: *3051 [lua] clustering.lua:437: serve_cluster_listener(): unable to send updated configuration to DP node with hostname: localhost.localdomain ip: 127.0.0.1 reason: CP and DP does not have same set of plugins installed or their versions might differ, client: 127.0.0.1, server: kong_cluster_listener, request: "GET /v1/outlet?node_id=e454bccd-de84-4b20-90cd-ec0cb76bf4c4&node_hostname=localhost.localdomain&node_version=2.2.0 HTTP/1.1", host: "127.0.0.1:8005"
2020/12/04 06:35:31 [warn] 4144#0: *3051 [lua] clustering.lua:553: unable to send updated configuration to DP node with hostname: localhost.localdomain ip: 127.0.0.1 reason: CP and DP does not have same set of plugins installed or their versions might differ, client: 127.0.0.1, server: kong_cluster_listener, request: "GET /v1/outlet?node_id=e454bccd-de84-4b20-90cd-ec0cb76bf4c4&node_hostname=localhost.localdomain&node_version=2.2.0 HTTP/1.1", host: "127.0.0.1:8005"
```
4. Restart DP with `KONG_PLUGINS=bundled`, request returns `404` now as sync is resumed.

bungle (approved):

PR #6613, opened 2020-12-01
feat(pdk) kong.node.get_hostname

kikito:
In addition of adding the new function to the PDK, this PR replaces `utils.get_hostname` everywhere with the new PDK function.


kikito:
Question about this one: should we memoize the hostname in the function? I don't think hostnames are the sort of thing that changes during the lifetime of a worker.

bungle (approved):

PR #6616, opened 2020-12-01
perf(core) loop over the plugins on startup only once

bungle:
### Summary

When there is a database with a huge amounts of plugins the startup can be a bit slow as we loop through plugins three times:

1. on init to check if there is plugin configured that is not installed on node
2. on init to build initial plugin iterator
3. on init_worker to pre-warm cache with plugin entities

This commit combines them together so that we only loop through plugins once.

This PR also contains another commit:

#### perf(core) build router and plugins iterator with max_page_size 

It is quite common for Kong to rebuild plugins iterator / router. Previously we used the default `1000` as the page size to build them. This commit changes the default page size to equal to `max_page_size` that currently is set to `50000` (by default).


kong/cache/marshall.lua
```diff
@@ -0,0 +1,80 @@
...
+
+
+local marshallers = {
+  shm_value = function(str_value, value_type, at, ttl)
+      return fmt("%d:%f:%f:%s", value_type, at, ttl, str_value)
+  end,
+
+  shm_nil = function(at, ttl)
+      return fmt("0:%f:%f:", at, ttl)
+  end,
```
javierguerragiraldez, line 29: why not
```
[0] = function() return "" end
```
?

javierguerragiraldez (approved):
works as intended, so ok from me.  my only point is nitpicking to get just a bit less special-case code

kong/cache/marshall.lua
bungle: Not sure, it is just a copy paste from here:
https://github.com/thibaultcha/lua-resty-mlcache/blob/master/lib/resty/mlcache.lua#L66-L96

thibaultcha: So that `nil` values can take a shortcut branch and avoid a bunch of table lookups.

kong/runloop/plugins_iterator.lua
```diff
@@ -486,6 +492,28 @@ function PluginsIterator.new(version)
         end
 
       else
+        if version == "init" then
+          if not enabled_plugins[name] then
+            return nil, name .. " plugin is in use but not enabled"
+          end
```
thibaultcha, line 498: Is this check still performed elsewhere for declarative configuration?

bungle: AFAIK this check did nothing on dbless as this point the declarative config was not loaded. Perhaps with reloads it did. But I think it is a good idea to add the check there too. Thanks for spotting it.

bungle: @thibaultcha I fixed it and moved the check above. Now the plugin name is compared on each time the plugin's iterator is built, on all modes.

PR #6620, opened 2020-12-02
Update DEVELOPER.md

ios-shah:
### Summary

Added some brief information regarding Linux Environment Variables as it seems to be incomplete without this

### Full changelog

Environment variables are dynamic values that affect the programs or processes running on a server. They exist on every operating system, and their type may vary. Environment variables can be created, edited, saved and deleted.

Linux environment variables are placeholders for information stored within the system that passes data to programs launched in shells or sub-shells.

### Issues resolved

Fix #XXX

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -1,6 +1,9 @@
+## Being familiar with basic concepts like environment variables
+Environment variables are dynamic values that affect the programs or processes running on a server. They exist on every operating system, and their type may vary. Environment variables can be created, edited, saved and deleted.
+
+Linux environment variables are placeholders for information stored within the system that passes data to programs launched in shells or sub-shells.
 
 These are the steps we follow at Kong to set up a development environment.
-
 
 ## Virtual Machine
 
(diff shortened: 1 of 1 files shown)


kikito (changes requested):
Thank you for this suggestion, but we don't think it is needed. The `DEVELOPERS.md` document intented audience is someone already familiar with the concepts you added. The same way a cooking recipe does not explain how to bring water to boil, we should not be explaining basic concepts like environment variables there.

Your PR points however to something that is indeed missing from that doc, which is explaining the intended audience and prerequisites for a developer willing to install Kong. It could be at the very top of the file. "Being familiar with basic concepts like environment variables" would be on such list.

I'll leave this PR open for a while in case you want to transform to include that missing section instead.

ios-shah:
Okk will make the further changes

ios-shah:
I made the further changes do have a look at it!!

ios-shah:
Hey do have a look at it and do tell me if you have any other suggestions..

ios-shah:
Hey !! Do have a look

kikito:
Hello, that is not what I was talking about. I suggest something like this:

```
## Intended audience and prerequisites

These instructions are intended for developers who want to start working with Kong.

As a prerequisite, the developer needs to be fluent in basic linux knowledge (like what an environment variable is, and what a linux distribution is) as well as familiar with virtualization software (VirtualBox and Docker)
```

PR #6632, opened 2020-12-07
fix(pdk/log) ensure the `tries` table is encoded as JSON array instea…

### LINKED ISSUE #6390: JSON log with tries object or array
All the log plugins use `kong.log.serialize` method to generate log message and then use cjson to encode the message. 
The problem is the tries object in log message can be an json array or empty object:
```
    "tries": [
        {
            "state": "next",
            "code": 502,
            "ip": "127.0.0.1",
            "port": 8000
        },
        {
            "ip": "127.0.0.1",
            "port": 8000
        }
    ]
```
```
    "tries": {}
```
which make other languages very difficult to decode the log.

Use `setmetatable(tries, cjson.array_mt)` method can fixed this issue, but I don't know where this code should go.

I create a PR #6386  that add this lines into `kong/pdk/log.lua`,  but a better idea is move it to `balancer phase`. Maybe someone could do this.

dndx:
…d of

object when empty. fixes #6390

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -18,6 +18,7 @@ local type         = type
 local ipairs       = ipairs
 local tostring     = tostring
 local tonumber     = tonumber
+local setmetatable = setmetatable
 local sub          = string.sub
 local byte         = string.byte
 local gsub         = string.gsub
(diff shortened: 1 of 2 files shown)


Tieske (approved):
👍

Tieske:
Approved, awaiting CI

bungle (approved):

dndx:
Test failure unrelated to this change, @gszr is working on a fix.

dndx:
Merging for now, @gszr is working on fixing the smoke test and it is not a blocker.

PR #6660, opened 2020-12-16
feat(request-size) option to require content-length header

Tieske:
Adds an option to the "request-size-limit" plugin to require the "content-length" header. If this is set it will prevent the plugin from trying to read the full body, which actually is behaviour counter to what the plugin is supposed to do. So this new option provides better protection.


Tieske:
There is an exception when "Transfer-Encoding: chunked" header is set, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/411

In that case there is no "Content-Length" header, and we must still fall-back to reading the body it seems.

@dndx @thibaultcha is there any way we can get the chunk sizes without reading the entire body? essentially protecting from too big sizes, without actually reading the bodies.

Tieske:
implementated the exception for chunked encoding.

bungle (approved):

PR #6672, opened 2020-12-18
feat(pdk) kong.log.set_serialize_value

kikito:
Log-type plugins (like file_log, or http_log) use the kong.log.serialize() function to "extract" a table representing the current status of Kong. That table is used in the logs.

This change introduces a new function which allows customizing the table returned by kong.log.serialize(). This can be used by plugins to customize the output of a logging plugin without having to "monkey-patch" it.

Related with #5968

### DIFF
--- a/kong/pdk/log.lua
+++ b/kong/pdk/log.lua
@@ -15,6 +15,7 @@ local ngx_re = require "ngx.re"
 local inspect = require "inspect"
 local ngx_ssl = require "ngx.ssl"
 local phase_checker = require "kong.pdk.private.phases"
+local utils = require "kong.tools.utils"
 
 
 local sub = string.sub
@@ -30,14 +31,21 @@ local setmetatable = setmetatable
 local ngx = ngx
 local kong = kong
 local check_phase = phase_checker.check
+local split = utils.split
 
 
 local _PREFIX = "[kong] "
 local _DEFAULT_FORMAT = "%file_src:%line_src %message"
 local _DEFAULT_NAMESPACED_FORMAT = "%file_src:%line_src [%namespace] %message"
-local PHASES_LOG = phase_checker.phases.log
+local PHASES = phase_checker.phases
+local PHASES_LOG = PHASES.log
 
-
+local phases_with_ctx =
+    phase_checker.new(PHASES.rewrite,
+                      PHASES.access,
+                      PHASES.header_filter,
+                      PHASES.body_filter,
+                      PHASES_LOG)
 local _LEVELS = {
   debug = ngx.DEBUG,
   info = ngx.INFO,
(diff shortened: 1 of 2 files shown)


kong/pdk/log.lua
```diff
@@ -465,11 +474,172 @@ local _log_mt = {
...
+        or not is_valid_value(val)
+      then
+        return false
+      end
+    end
+
+    return true
+  end
+
+  set_serialize_value = function(key, value, options)
```
javierguerragiraldez, line 554: little known fact: the "normal" `function name(args)...` form can assign to an upvalue just as well:
```lua
local visible_function
do
  local private_values
  ...
  function visible_function()
    ...
  end
end
```
works as expected :smiley: !

kong/pdk/log.lua
```diff
@@ -465,11 +474,172 @@ local _log_mt = {
...
+  -- Modify returned table with values set with kong.log.set_serialize_values
+  local function edit_result(ctx, root)
+    local serialize_values = ctx.serialize_values or get_default_serialize_values()
+    --[[
+    ngx.log(ngx.INFO, require("inspect")({
+      a = "before",
+      serialize_values = serialize_values,
+      root = root
+    }, { newline = " ", indent = "" }))
+    ]]
```
javierguerragiraldez, line 592: debugging leftover?

kong/pdk/log.lua
```diff
@@ -465,11 +474,172 @@ local _log_mt = {
...
+      --ngx.log(ngx.INFO, require("inspect")({ subkeys = subkeys, mode = mode, new_value = new_value, key = key }, { newline = " ", indent = "" }))
+      node = root -- start in root, iterate with each subkey
+      for i = 1, #subkeys - 1 do -- note that last subkey is treated differently, below
+        subkey = subkeys[i]
+        if node[subkey] == nil then
+          if mode == "set" or mode == "add" then
+            node[subkey] = {} -- add subtables as needed
+          else
+            node = nil
+            break -- mode == replace; and we have a missing link on the "chain"
```
javierguerragiraldez, line 608: should this raise an error message?

kong/pdk/log.lua
```diff
@@ -465,11 +474,172 @@ local _log_mt = {
...
+
+    options = options or {}
+    local mode = options.mode or "set"
+
+    if type(key) ~= "string" then
+      error("key must be a string", 2)
+    end
+
+    if not is_valid_value(value) then
+      error("value must be nil, a number, string, boolean or a table made with only numbers, string and booleans", 2)
```
javierguerragiraldez, line 565: is it important to do this check at set time?  of course checking as early as possible makes for easier debugging, but given that the check is unbounded in time, it might be worth to leave until output time, so it doesn't affect request latency.

javierguerragiraldez (approved):
very nice job!   just nitpicking in the comments.

kong/pdk/log.lua
kikito: No, that is a totally valid behavior. Consider the default:

```
{ key = "request.headers.authorization", value = "REDACTED", mode = "replace" }
```

We want to replace the header _if it exists_. But if it doesn't (or if the request has no headers whatsoever) we want to just ignore this entry and continue with the next one.

kong/pdk/log.lua
kikito: It is my personal preference. This line:

```
function visible_function()
```

Triggers my mental `YOU ARE SETTING A GLOBAL VARIABLE`-alarm, so every time I see that I must do a double-take and make sure that the variable is `local`-ized somewhere. Even if I have luacheck, which takes care of that sort of thing, it is still in the background eating cycles until I manually do the check.

The assignment syntax does not trigger that alarm for me so it is "cheaper to read" for me 😁.

kong/pdk/log.lua
kikito: Hm. Yes certainly we could make the check in the log phase.

> unbounded in time

That comment makes me realize that the check function is going to choke badly on tables containing self-references. Let me add a check for that as well.

kong/pdk/log.lua
kikito: Yes. Thanks, removed.

kong/pdk/log.lua
Tieske: without looking at the code; is that even needed? that seems bloat. People should not be using circular references when logging data.

kikito: The pro is that we avoid a possible infinite loop if someone (out of mistake or malice) passes a recursive table.

The con is that we spend a tiny bit of time and memory doing the checks (one new table per key). But it is a really small and cheap change (literally 4 lines more).

kong/pdk/log.lua
```diff
@@ -30,14 +31,22 @@ local setmetatable = setmetatable
...
 local _DEFAULT_FORMAT = "%file_src:%line_src %message"
 local _DEFAULT_NAMESPACED_FORMAT = "%file_src:%line_src [%namespace] %message"
-local PHASES_LOG = phase_checker.phases.log
-
-
+local PHASES = phase_checker.phases
+local PHASES_LOG = PHASES.log
+
+local phases_with_ctx =
+    phase_checker.new(PHASES.certificate,
```
bungle, line 44: The `certificate` phase AFAIK will need OpenResty 1.19.x to work. What I mean is that prior that the `ctx` table is there, but it will be lost after the phase is done. But as it probably quite uncommon to have this on `certificate`, I think it is fine as is, and it starts working when we update the OpenResty, see:

https://github.com/Kong/kong/blob/next/kong/init.lua#L650-L651

kong/pdk/log.lua
```diff
@@ -465,11 +474,158 @@ local _log_mt = {
...
+-- It can be used to delete existing values by passing `nil`.
+--
+-- Note: the type checking of the `value` parameter can take some time so
+-- it is deferred to the `serialize()` call, which happens in the log
+-- phase in most real-usage cases.
+--
+-- @function kong.log.set_serialize_value
+-- @phases certificate, rewrite, access, header_filter, body_filter, log
+-- @tparam string key the name of the field.
+-- @tparam number|string|boolean|table value value to be set. When a table is used, its keys must be numbers, strings, booleans, and its values can be numbers, strings or other tables like itself, recursively.
```
bungle, line 505: I suppose arrays are supported? And what kind of arrays? Even arrays with hash parts? What about `ngx.null` AND `ngx.null` handling in general in this code (perhaps not support `ngx.nulls`, but can there be `ngx.nulls` already in existing log data)?

bungle (approved):
Just a few comments. But approving, you know the best if those comments should be considered somehow (probably not)

kong/pdk/log.lua
kikito: I removed certificate from the list. If we ever make it not lost, we can add it back, but otherwise it might generate confusion (people could try setting stuff on cert phase and then not find it on the log).

kong/pdk/log.lua
kikito: Arrays are supported, but usually we will see them on the `value` field - `kong.log.set_serialize_value("foo", {1,2,3})`. Values can be arrays or hybrid tables.

We should not see it on the `key` field because the table returned by `kong.log.serialize()` is a hash-like array.

`ngx.null` is not permitted in `set_serialize_value`, but it is permitted in existing log data.

PR #6681, opened 2020-12-22
feat(kong) add sandbox and sandbox_helpers

eskerda:
### Summary

Some Kong functionality (present and future) and plugins uses arbitrary lua code, which provides the most definitive way of extending usage. This comes with its own security problems:

* Not all users will use this feature
* Every kong part and plugin loads arbitrary lua differently
* On unsecured kong deploys, opens up a security door depending on what environment context is allowed to the lua code.

Point is, there's no silver bullet to accept or disable this functionality, and it will depend on what every kong user wants or expects from their installation.

This PR proposes:
* Adding `sandbox.lua` as a dependency https://github.com/kikito/sandbox.lua/pull/2 (currently embedded on this PR with some modifications) to run these on a restricted sandbox. We can add that as a requirement on the luarocks when / if it gets published, or we can go with a simpler sandbox, like: https://github.com/Kong/kong-plugin-serverless-functions/blob/0addd34396136bbf2479a390e2c9faa41d5accb9/kong/plugins/pre-function/_handler.lua#L4-L71
* Providing sandbox_helpers that wrap `whatever we use for sandboxing`, and get a common usage through kong codebase. It provides methods for parsing functions, validating (for plugins schema), etc. This will de dupe lots of code and will make sure that it all behaves as expected. It also makes sure we have a set of defaults on how we use the `load` function, like using `mode = t` and not allowing binary text.
* Providing a set of kong configuration values to control how does arbitrary lua run, as:

```
KONG_UNTRUSTED_LUA=on,off,sandbox
KONG_UNTRUSTED_LUA_SANDBOX_REQUIRES=some,modules,you,are,allowed,to,require,on,sandbox,mode
KONG_UNTRUSTED_LUA_SANDBOX_MODULES=G,modules,you,want,accessible,on,the,sandbox,like,kong.request
```

These settings mean the following:

- `KONG_UNTRUSTED_LUA = off` completely disables running arbitrary lua code no matter what.
- `KONG_UNTRUSTED_LUA = sandbox`  uses a sandbox to run arbitrary lua
    - `KONG_UNTRUSTED_LUA_SANDBOX_REQUIRES` is a comma separated list of allowed requires within the sandbox
    - `KONG_UNTRUSTED_LUA_SANDBOX_ENVIRONMENT` is a comma separated list of available global variables on the sandbox context. For example, adding `kong.request` would make `kong.request` available within the sandbox and `kong.request.get_header('foo')` woud be possible.
- `KONG_UNTRUSTED_LUA = off` will run arbitrary lua code without any restriction.

This might look like lots of settings, but again, there's no silver bullet for these features, some users will want something on their environment that others won't.

At the moment, the defaults are:

```
KONG_UNTRUSTED_LUA = on
KONG_UNTRUSTED_LUA_SANDBOX_REQUIRES = (empty)
KONG_UNTRUSTED_LUA_SANDBOX_ENVIRONMENT = (empty)
```

### Full changelog

* Add sandbox_helpers and kong configuration settings for it

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1334,3 +1334,59 @@
                                  # server.
                                  #
                                  # See https://github.com/openresty/lua-nginx-module#lua_socket_pool_size
+
+#untrusted_lua = sandbox
+                                 # Accepted values are:
+                                 #
+                                 # - `off`: disallow any loading of Lua functions
+                                 #          from admin supplied sources (such as via the Admin API).
+                                 #
+                                 #          Note using the `off` option will render plugins such as
+                                 #          Serverless Functions unusable.
+                                 # - `sandbox`: allow loading of Lua functions from admin
+                                 #              supplied sources, but use a sandbox when
+                                 #              executing them. The sandboxed
+                                 #              function will have restricted access
+                                 #              to the global environment and only
+                                 #              have access to standard Lua functions
+                                 #              that will generally not cause harm to
+                                 #              the Kong node.
+                                 #
+                                 #              In this mode, the `require` function inside
+                                 #              the sandbox only allows loading external Lua
+                                 #              modules that are explicitly listed in
+                                 #              `untrusted_lua_sandbox_requires` below.
+                                 #
+                                 #              LuaJIT bytecode loading is disabled.
+                                 #
+                                 #              Warning: LuaJIT is not designed as a secure
+                                 #              runtime for running malicious code, therefore,
+                                 #              you should properly protect your Admin API endpoint
+                                 #              even with sandboxing enabled. The sandbox only
+                                 #              provides protection against trivial attackers or
+                                 #              unintentional modification of the Kong global
+                                 #              environment.
+                                 # - `on`: allow loading of Lua functions from admin
+                                 #         supplied sources and do not use a sandbox when
+                                 #         executing them. Functions will have unrestricted
+                                 #         access to global environment and able to load any
+                                 #         Lua modules. This is similar to the behavior in Kong
+                                 #         prior to 2.3.0.
+                                 #
+                                 #         LuaJIT bytecode loading is disabled.
+
+#untrusted_lua_sandbox_requires =
+                                 # Comma-separated list of modules allowed to be loaded
+                                 # with `require` inside the sandboxed environment. Ignored
+                                 # if `untrusted_lua` is not `sandbox`.
+                                 #
+                                 # Note: certain modules, when allowed, may cause sandbox
+                                 # escaping trivial.
+
+#untrusted_lua_sandbox_environment =
+                                 # Comma-separated list of global Lua variables
+                                 # that should be made available inside the sandboxed
+                                 # environment. Ignored if `untrusted_lua` is not `sandbox`.
+                                 #
+                                 # Note: certain variables, when made available,
+                                 # may cause sandbox escaping trivial.
--- a/kong/tools/sandbox.lua
+++ b/kong/tools/sandbox.lua
@@ -1,0 +1,201 @@
+local _sandbox = require "sandbox"
+
+local table = table
+local fmt = string.format
+local setmetatable = setmetatable
+local require = require
+local ipairs = ipairs
+local pcall = pcall
+local type = type
+local error = error
+local rawset = rawset
+local assert = assert
+local kong = kong
+
+
+-- deep copy tables using dot notation, like
+-- one: { foo = { bar = { hello = {}, ..., baz = 42 } } }
+-- target: { hey = { "hello } }
+-- link("foo.bar.baz", one, target)
+-- target -> { hey = { "hello" }, foo = { bar = { baz = 42 } } }
+local function link(q, o, target)
+  if not q then return end
+
+  local h, r = q:match("([^%.]+)%.?(.*)")
+  local mod = o[h]
+
+  if not mod then return end
+
+  if r == "" then
+    if type(mod) == 'table' then
+      -- changes on target[h] won't affect mod
+      target[h] = setmetatable({}, { __index = mod })
+
+    else
+      target[h] = mod
+    end
+
+    return
+  end
+
+  if not target[h] then target[h] = {} end
+
+  link(r, o[h], target[h])
+end
+
+
+local lazy_conf_methods = {
+  enabled = function(self)
+    return kong and
+           kong.configuration and
+           kong.configuration.untrusted_lua and
+           kong.configuration.untrusted_lua ~= 'off'
+  end,
+  sandbox_enabled = function(self)
+    return kong and
+           kong.configuration and
+           kong.configuration.untrusted_lua and
+           kong.configuration.untrusted_lua == 'sandbox'
+  end,
+  requires = function(self)
+    local conf_r = kong and
+                   kong.configuration and
+                   kong.configuration.untrusted_lua_sandbox_requires or {}
+    local requires = {}
+    for _, r in ipairs(conf_r) do requires[r] = true end
+    return requires
+  end,
+  env_vars = function(self)
+    return kong and
+           kong.configuration and
+           kong.configuration.untrusted_lua_sandbox_environment or {}
+  end,
+  environment = function(self)
+    local env = {
+        -- home brewed require function that only requires what we consider
+        -- safe :)
+        ["require"] = function(m)
+          if not self.requires[m] then
+            error(fmt("require '%s' not allowed within sandbox", m))
+          end
+
+          return require(m)
+        end,
+    }
+
+    for _, m in ipairs(self.env_vars) do link(m, _G, env) end
+
+    return env
+  end,
+}
+
+
+local conf_values = {
+  clear = table.clear,
+  reload = table.clear,
+  err_msg = "loading of untrusted Lua code disabled because " ..
+            "'untrusted_lua' config option is set to 'off'"
+}
+
+
+local configuration = setmetatable({}, {
+  __index = function(self, key)
+    local l = lazy_conf_methods[key]
+
+    if not l then
+      return conf_values[key]
+    end
+
+    local value = l(self)
+    rawset(self, key, value)
+
+    return value
+  end,
+})
+
+
+local sandbox = function(fn, opts)
+  if not configuration.enabled then
+    error(configuration.err_msg)
+  end
+
+  opts = opts or {}
+
+  local opts = {
+    -- default set load string mode to only 'text chunks'
+    mode = opts.mode or 't',
+    env = opts.env or {},
+    chunk_name = opts.chunk_name,
+  }
+
+  if not configuration.sandbox_enabled then
+    -- sandbox disabled, all arbitrary Lua code can execute unrestricted
+    setmetatable(opts.env, { __index = _G})
+
+    return assert(load(fn, opts.chunk_name, opts.mode, opts.env))
+  end
+
+  -- set (discard-able) function context
+  setmetatable(opts.env, { __index = configuration.environment })
+
+  return _sandbox(fn, opts)
+end
+
+
+local function validate_function(fun, opts)
+  local ok, func1 = pcall(sandbox, fun, opts)
+  if not ok then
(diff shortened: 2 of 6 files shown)


.luacheckrc
```diff
@@ -52,3 +52,23 @@ files["spec/**/*.lua"] = {
...
+    read_globals = {
+        "table.unpack",
+    }
+}
+
+files["spec/01-unit/020-sandbox/02-sandbox_helpers_spec.lua"] = {
+    read_globals = {
+        "table.unpack",
+    }
+}
```
Tieske, line 74: any reason to modify the global file instead of using inline `-- luacheck: ignore` markers?

spec/01-unit/020-sandbox/01-sandbox_spec.lua
```diff
@@ -0,0 +1,126 @@
...
+      assert.error(function() sandbox.run("while true do end") end)
+      assert.equal('hellohello', string.rep('hello', 2))
+    end)
+
+    it('accepts a quota param', function()
+      assert.has_no.errors(function() sandbox.run("for i=1,100 do end") end)
+      assert.error(function() sandbox.run("for i=1,100 do end", {quota = 20}) end)
+    end)
+
+    it('does not use quotes if the quote param is false', function()
```
Tieske, line 101: typo: quota

kong/tools/sandbox.lua
```diff
@@ -0,0 +1,170 @@
...
+-- * string.rep: can be used to allocate millions of bytes in 1 operation
+-- * {set|get}metatable: can be used to modify the metatable of global objects (strings, integers)
+-- * collectgarbage: can affect performance of other systems
+-- * dofile: can access the server filesystem
+-- * _G: It has access to everything. It can be mocked to other things though.
+-- * load{file|string}: All unsafe because they can grant acces to global env
+-- * raw{get|set|equal}: Potentially unsafe
+-- * module|require|module: Can modify the host settings
+-- * string.dump: Can display confidential server info (implementation of functions)
+-- * string.rep: Can allocate millions of bytes in one go
```
Tieske, line 67: duplicate; string.rep is the first one at the top already

kong.conf.default
```diff
@@ -1334,3 +1334,21 @@
...
+#untrusted_lua = on              # When enabled, Kong will accept and run
+                                 # untrusted lua code on plugins and features
+                                 # that accept it.
+
+#untrusted_lua_sandbox = on      # When enabled, Kong uses a secure sandbox
+                                 # for running untrusted lua code. If disabled
+                                 # untrusted lua code has unrestricted access
+                                 # to the global context
+
+#untrusted_lua_sandbox_requires =
```
Tieske, line 1347: "modules" instead of "requires"?

kong.conf.default
```diff
@@ -1334,3 +1334,21 @@
...
                                  # See https://github.com/openresty/lua-nginx-module#lua_socket_pool_size
+
+#untrusted_lua = on              # When enabled, Kong will accept and run
+                                 # untrusted lua code on plugins and features
+                                 # that accept it.
+
+#untrusted_lua_sandbox = on      # When enabled, Kong uses a secure sandbox
+                                 # for running untrusted lua code. If disabled
+                                 # untrusted lua code has unrestricted access
+                                 # to the global context
```
Tieske, line 1345: can we add a small note on how the 2 above relate to each other?

kong/tools/sandbox_helpers.lua
```diff
@@ -0,0 +1,183 @@
...
+               kong.configuration and
+               kong.configuration.untrusted_lua_sandbox_environment or {}
+      end,
+      environment = function(self)
+        local env = {
+            -- home brewed require function that only requires what we consider
+            -- safe :)
+            require = function(m)
+              if not self.requires[m] then
+                error(fmt("require \'%s\' not allowed within sandbox", m))
```
Tieske, line 75: the backslashes are not needed here

kong/tools/sandbox_helpers.lua
```diff
@@ -0,0 +1,183 @@
...
+    mode = opts.mode or 't',
+    env = opts.env or {},
+    chunk_name = opts.chunk_name,
+  }
+
+  if not configuration.sandbox_enabled then
+    -- sandbox disabled, all arbitrary lua code belongs to us
+    setmetatable(opts.env, { __index = _G})
+    local fn, err = load(fn, opts.chunk_name, opts.mode, opts.env)
+    if err then error(err) end
```
Tieske, line 122: if rethrowing the err, then just use an asssert?

Tieske (commented):
first glance comments.

Q:
- do we need some caching of functions/sandboxes?
- isn't this breaking? so for 3.0?

eskerda:
> first glance comments.
> 
> Q:
> 
>     * do we need some caching of functions/sandboxes?

This is still responsibility of modules using `sandbox_helpers`. But overall it makes it quite lighter, at least for the obvious cases: https://github.com/Kong/kong-plugin-enterprise-exit-transformer/pull/16 https://github.com/Kong/kong-plugin-response-transformer-advanced/pull/39/files and this is a bit more involved because of the way it uses functions and upvalues, but still useful https://github.com/Kong/kong-plugin-serverless-functions/commit/cd504f219d315b102fb520a66f4f7ac79aba28ac

>     * isn't this breaking? so for 3.0?

It's not really breaking depending on the defaults that this ships with

* `KONG_UNTRUSTED_LUA = on, KONG_UNTRUSTED_LUA_SANDBOX = off` makes it all behave exactly the same as before
* or we could trust the sandbox

kong.conf.default
eskerda: it's not the same. This makes some requires available to a fake `require` function. So, this just allows a list of things that can be "required", thus `requires`.

`modules` can be confused with what `untrusted_lua_sandbox_environment` provides.

Tieske: imo this should stay as close to standard Lua as possible, that we limit the modules to a specific subset is fine, but they are still modules. So using "requires" is just confusing to the user.

Tieske:
The defaults should be as before. If we choose to restrict them, then that should be in 3.0.

Tieske:
the proposed settings should be collapsed into an enum imo:
```
KONG_UNTRUSTED_LUA = on
KONG_UNTRUSTED_LUA_SANDBOX = on
```
should become something like:
```
KONG_UNTRUSTED_LUA = enabled | sandbox | disabled
```

.luacheckrc
javierguerragiraldez: if luacheck's `ngx_lua` target doesn't include luajit's 5.2/.3 backports, the more "global" the fix, the better.  I'd suggest adding to the target, not to each file's exception list.

eskerda:
> the proposed settings should be collapsed into an enum imo:
> 
> ```
> KONG_UNTRUSTED_LUA = on
> KONG_UNTRUSTED_LUA_SANDBOX = on
> ```
> 
> should become something like:
> 
> ```
> KONG_UNTRUSTED_LUA = enabled | sandbox | disabled
> ```

made it `on` `off` `sandbox`. Thanks for the suggestion!

dndx:
One general comment I want to add is to remove mentions of "secure" sandbox in this feature. LuaJIT is not a secure VM (unlike V8, for example), and there are definitely exploits out there to escape any kind of sandbox in LuaJIT, we should not give user the impression they can trust running untrusted code inside the sandbox regardless of what feature we decide to implement.

kong/tools/sandbox.lua
```diff
@@ -0,0 +1,160 @@
...
+
+-- Public interface: sandbox.protect
+function sandbox.protect(f, options)
+  options = options or {}
+
+  local passed_env = options.env or {}
+  local env = setmetatable({}, {
+    __index = function(_, k)
+      local v = BASE_ENV[k]
+      return v ~= nil and v or passed_env[k]
```
javierguerragiraldez, line 126: if `BASE_ENV[k] == false` this would return `passed_env[k]`, which i guess is not the intention.

```suggestion
      if v == nil then
        return passed_env[k]
      end
      return v
```

kong/tools/sandbox_helpers.lua
```diff
@@ -0,0 +1,193 @@
+-- XXX Use sandbox.lua once it gets published
+-- https://github.com/kikito/sandbox.lua/pull/2
+local _sandbox = require "kong.tools.sandbox"
+
+local utils = require "kong.tools.utils"
+
+local fmt = string.format
+local unpack = table.unpack
```
javierguerragiraldez, line 8:
```suggestion
local unpack = unpack or table.unpack
```

kong/tools/sandbox_helpers.lua
```diff
@@ -0,0 +1,193 @@
...
+local split = utils.split
+
+
+local configuration = setmetatable({
+  -- useful method for specs
+  reload = function(self)
+    local r = self.reload
+    for k in next, self do rawset(self, k, nil) end
```
javierguerragiraldez, line 16: why not `pairs(self)`?  or even `table.clear()`?

javierguerragiraldez (approved):

kikito:
Hello, I talked a bit with @eskerda and decided to make some changes to the sandbox.lua file.

* We decided to remove support for naked functions, sandbox only accepts strings of Lua code. This simplifies the code but otherwise doesn't affect this PR since Kong only uses strings.
* We decided to make the environment read-only. This makes the surface attack smaller (one plugin cannot "disturb" other plugins) and simplifies code. The way this is done is by creating a clone of the base environment, and use the passed one as "index". This way, the passed environment can have metatables etc. The creation of the clone involves duplicating a table with ~18 elements. On the other hand all accesses to plain-lua functions like `pairs` will be slightly faster since no metatable lookups will be needed for those.
* Bytecode is always blocked, instead of optionally blocked.

You can see more details on the sandbox.lua repo: https://github.com/kikito/sandbox.lua/commits/master

This modified sandbox.lua is available in luarocks as 'kikito/sandbox'.

I have created a branch which replaces the sandbox.lua file used on this branch by the luarock, and the `sandbox_helpers` tests are green, with no changes needed:

https://github.com/Kong/kong/tree/feat/sandbox-helpers-next-kikito

@eskerda  Please feel free to add these changes to the branch, or if you prefer I can add them myself.

kong.conf.default
```diff
@@ -1334,3 +1334,19 @@
                                  # server.
                                  #
                                  # See https://github.com/openresty/lua-nginx-module#lua_socket_pool_size
+
+#untrusted_lua = on              # on | off | sandbox
```
kikito, line 1338: Let's make this off by default.

kikito (changes requested):

kong.conf.default
```diff
@@ -1334,3 +1334,19 @@
                                  # server.
                                  #
                                  # See https://github.com/openresty/lua-nginx-module#lua_socket_pool_size
+
+#untrusted_lua = off             # off | sandbox | on
```
kikito, line 1338: I thought we wanted to disable custom lua completely by default, but I misread the intent.

Apparently the right default is `untrusted_lua=sandbox`.

My bad.

kikito (changes requested):

eskerda:
@kikito @javierguerragiraldez I applied the proposed changes and squashed all commits together. Thanks for all the suggestions and help reviewing this PR

@kikito note one suggestion @javierguerragiraldez had that I think still applies:
- https://github.com/Kong/kong/pull/6681#discussion_r552657122
- https://github.com/kikito/sandbox.lua/pull/5

kikito:
> kikito/sandbox.lua#5

Thanks for this! Added a test and released sandbox v1.0.1

kong/tools/sandbox_helpers.lua
dndx: Actually, I just changed this to use `string.match` instead of `split` and then `unpack` because the former two are notoriously slow. `split` allocates a new table and `unpack` is [NYI in LuaJIT](http://wiki.luajit.org/NYI#libraries_table-library).

kikito (approved):

dndx (approved):

PR #6689, opened 2020-12-30
fix(ext-plugin): msg_id scope per-connection, not per-pluginserver

javierguerragiraldez:
the `msg_id` field in every packet helps associate call and return packets.  Each call increments a counter, but the result should be local during the call/return cycle to avoid interference from concurrent HTTP requests.


dndx (approved):

dndx:
Question before merging, should this be against `master` instead of `next`?

javierguerragiraldez:
being a fix, i'd like it to be everywhere.  but it seems `master` doesn't have multi-pluginserver (yet?)

PR #6705, opened 2021-01-06
fix(schema) limit positive integer limit to 255 (1 byte) when setting

dndx:
healthchecker thresholds


murillopaula (approved):

Tieske:
why was this changed? didn't we have issues in the past with this? Maybe @locao knows?

murillopaula:
@Tieske see FTI-796

Tieske:
@murillopaula thx! that was the one

PR #6731, opened 2021-01-13
chore(deps) bump resty-healthcheck to 2.0.0

bungle:


bungle:
I marked it with `do no merge` as the `2.0.0` has this:
https://github.com/Kong/lua-resty-healthcheck/blob/master/rockspecs/lua-resty-healthcheck-2.0.0-1.rockspec#L18

And the worker-events is not yet bumped. So perhaps we need to bump them together, but let's keep this here open so that we know that this is at least green, though it might get red together with `worker-events` `2.0.0`.

locao:
> I marked it with do no merge as the 2.0.0 has this:
> https://github.com/Kong/lua-resty-healthcheck/blob/master/rockspecs/lua-resty-healthcheck-2.0.0-1.rockspec#L18
> 
> And the worker-events is not yet bumped. So perhaps we need to bump them together, but let's keep this here open so that we know that this is at least green, though it might get red together with worker-events 2.0.0.

Also, lua-resty-healthcheck 2.0 still uses two timers per target being active checked, instead of a single timer for all targets as 1.4.0 does. It seems to me that the timer usage has to be addressed in lua-resty-healthcheck 2.1 before bumping here.

PR #6735, opened 2021-01-14
fix(proxy) return 502 on connectivity errors with buffered proxying

### LINKED ISSUE #6726: Incorrect error handling in buffered mode
### Summary

There seems to be a bug in the error handling code, when buffering is enabled, which occurs if the upstream server does not respond. The result is that a 500 Internal Server Error is returned when the upstream server does not respond, rather than a 502 Bad Gateway (as expected, and as happens when buffering is not enabled). There is also error output in the logs, stating that Kong is trying to call `get_header` in the `balancer` phase.

I have attached the simplest repro that I can create of this issue, which is a plugin that calls `enable_buffering` and redirects to a (hopefully) unused port on `localhost`. The relevant plugin code is under the `./lua/kong/plugins/kong-repro` path, in `handler.lua`.

The error message seems to indicate that the error handling is being called during the balancer phase, but still trying to get the content type from the headers, which it is not allowed to do. 

### Steps To Reproduce

1. I have attached a Zip file with a basic plugin that repros this, as well as a docker-compose that runs it with the correct arguments. Download and extract this. In the directory:
2. docker-compose build
3. docker-compose up kong
4. Hit http://localhost:8000/repro

Expected: Server responds with a 502 error code

Actual: Server responds with 500 error code, and the following error log in output:

```
kong_1  | 2021/01/12 14:28:38 [error] 28#0: *66 lua entry thread aborted: runtime error: /usr/local/share/lua/5.1/kong/pdk/private/phases.lua:85: function cannot be called in balancer phase (only in: rewrite, access, cluster_listener, header_filter, body_filter, log)
kong_1  | stack traceback:
kong_1  | coroutine 0:
kong_1  |       [C]: in function 'error'
kong_1  |       /usr/local/share/lua/5.1/kong/pdk/private/phases.lua:85: in function 'check_phase'
kong_1  |       /usr/local/share/lua/5.1/kong/pdk/request.lua:562: in function 'get_header'
kong_1  |       /usr/local/share/lua/5.1/kong/error_handlers.lua:31: in function 'access'
kong_1  |       access_by_lua(nginx-kong.conf:88):2: in main chunk while sending to client, client: 172.20.0.1, server: kong, request: "GET /repro HTTP/1.1", host: "localhost:8000"

```
### Additional Details & Logs

- Kong version: 2.2.0
- Kong error logs: see above
- Operating system: Centos+Kong Docker Image

[kong-repro.zip](https://github.com/Kong/kong/files/5802889/kong-repro.zip)

bungle:
### Summary

In #6726 @danielki33 reported that on upstream connection errors with buffered proxying we return `500` instead of `501` and that runtime error was logged. This PR fixes that.

### Issues Resolved

Fix #6726

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -824,9 +824,12 @@ do
 
     local res = ngx.location.capture("/kong_buffered_http", options)
     if res.truncated then
+      kong_global.set_phase(kong, PHASES.error)
       ngx.status = 502
       return kong_error_handlers(ngx)
     end
+
+    kong_global.set_phase(kong, PHASES.response)
 
     local status = res.status
     local headers = res.header
@@ -855,8 +858,6 @@ do
     if not ctx.KONG_PROXY_LATENCY then
       ctx.KONG_PROXY_LATENCY = ctx.KONG_RESPONSE_START - ctx.KONG_PROCESSING_START
     end
-
-    kong_global.set_phase(kong, PHASES.response)
 
     kong.response.set_status(status)
     kong.response.set_headers(headers)
--- a/spec/02-integration/05-proxy/24-buffered_spec.lua
+++ b/spec/02-integration/05-proxy/24-buffered_spec.lua
@@ -92,6 +92,31 @@ for _, strategy in helpers.each_strategy() do
           }
         }
 
+        local s502 = bp.services:insert {
+          name = "502",
+          host = "127.0.0.2",
+          port = 26865,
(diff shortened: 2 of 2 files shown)


javierguerragiraldez (commented):
LGTM.   surely the failure is unrelated?

javierguerragiraldez (approved):

PR #6736, opened 2021-01-14
fix(fixtures) listen for ipv6 in mock webserver

locao:
The new HTTPS mock server based on Nginx was not being properly configured for IPv6 addresses.

### DIFF
--- a/spec/fixtures/mock_webserver_tpl.lua
+++ b/spec/fixtures/mock_webserver_tpl.lua
@@ -28,9 +28,11 @@ http {
 
   server {
 # if protocol ~= 'https' then
-    listen ${http_port};
+    listen 127.0.0.1:${http_port};
+    listen [0000:0000:0000:0000:0000:0000:0000:0001]:${http_port};
 # else
-    listen ${http_port} ssl;
+    listen 127.0.0.1:${http_port} ssl;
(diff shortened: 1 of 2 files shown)


bungle (approved):

spec/fixtures/mock_webserver_tpl.lua
```diff
@@ -28,9 +28,11 @@ http {
 
   server {
 # if protocol ~= 'https' then
-    listen ${http_port};
+    listen 127.0.0.1:${http_port};
+    listen [0000:0000:0000:0000:0000:0000:0000:0001]:${http_port};
 # else
-    listen ${http_port} ssl;
+    listen 127.0.0.1:${http_port} ssl;
```
bungle, line 34: should we also allow http2?

spec/fixtures/mock_webserver_tpl.lua
```diff
@@ -28,9 +28,11 @@ http {
 
   server {
 # if protocol ~= 'https' then
-    listen ${http_port};
+    listen 127.0.0.1:${http_port};
+    listen [0000:0000:0000:0000:0000:0000:0000:0001]:${http_port};
```
dndx, line 32:
```suggestion
    listen [::1]:${http_port};
```

should be good enough. `::` can replace any consecutive `0`s in the IPv6 address.

There are a few places that can be changed like this.

Issue #6035, opened 2020-06-18
Kong doesn't support scram-sha-256 passwords


filterfish:
The postgres docs recommend that you use `scram-sha-256` password encryption by adding the following to `postgresql.conf`:
```
password_encryption = 'scram-sha-256'
```

However this causes kong to fail with the utterly cryptic message:

```
don't know how to auth: 10
```

which as I'm sure you'll agree is pretty useless! This isn't entirely your fault as that's how `pgmoon` reports it but I'm sure there's something you can do to improve it or hopefully fix it.

Also, I'm reporting this here mainly so the words `don't know how to auth: 10` are reported at least somewhere on the internet other than the source code.

zhengganCN:
I also have this problem。

> Error: 
/usr/local/share/lua/5.1/pgmoon/init.lua:211: don't know how to auth: 10
stack traceback:
	[C]: in function 'auth'
	/usr/local/share/lua/5.1/pgmoon/init.lua:211: in function 'connect'
	.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:211: in function 'connect'
	.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:527: in function 'query'
	.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:279: in function 'init'
	/usr/local/share/lua/5.1/kong/db/init.lua:141: in function 'init_connector'
	/usr/local/share/lua/5.1/kong/cmd/start.lua:31: in function 'cmd_exec'
	/usr/local/share/lua/5.1/kong/cmd/init.lua:88: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:88>
	[C]: in function 'xpcall'
	/usr/local/share/lua/5.1/kong/cmd/init.lua:88: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:45>
	/usr/local/bin/kong:9: in function 'file_gen'
	init_worker_by_lua:47: in function <init_worker_by_lua:45>
	[C]: in function 'xpcall'
	init_worker_by_lua:54: in function <init_worker_by_lua:52>

venjiang:
I also have this problem, what can I do?

Maw1395:
Yep, install Postgres 9.5. Fixed the issue I was having. It doesn't seem to support the newer versions.

jmhungdev:
I have postgres@12 and keep seeing this error message which seems to be new and no solution for this yet. So the only solution is to uninstall it and install version 9.5?

[...]

PR #6740, opened 2021-01-15
chore(deps) bump lua-resty-openssl to 0.6.10

bungle:
##### feat
- **pkcs12:** allow to define algorithm to encrypt key and cert [b9678ce](https://github.com/fffonion/lua-resty-openssl/commit/b9678ce4ee4a233fb0bd8ed61d41c6d45a6fbb9d)
- **pkcs12:** check on cert and key mismatch [5953cc2](https://github.com/fffonion/lua-resty-openssl/commit/5953cc281cff06027f3b2bba23402e2915fd3ae1)
- **pkcs12:** encode and decode for pkcs12 [1467579](https://github.com/fffonion/lua-resty-openssl/commit/1467579fbe253996570dd188f580b98b8eb1db98)
- **pkey:** add is_private function to check if it's a private key [eb6cc1c](https://github.com/fffonion/lua-resty-openssl/commit/eb6cc1c2d5f7698c2641950d745a78da7baa6225)
- **ssl:** add the ssl and ssl_ctx module [40f3999](https://github.com/fffonion/lua-resty-openssl/commit/40f39994446a4cb954fc516f7047194cbf1141f8)

##### fix
- **ecx:** return nil, err in set_parameters [98acaee](https://github.com/fffonion/lua-resty-openssl/commit/98acaeeeaa60dffd93a934f4fbf7ddfd8e9e9652)
- **pkey:** use named_curve encoding for EC group [1e65d9d](https://github.com/fffonion/lua-resty-openssl/commit/1e65d9d4b71c0e9c5f4d404e640a96e03902fd30)


kong-2.3.0-0.rockspec
```diff
@@ -36,7 +36,7 @@ dependencies = {
   "lua-resty-cookie == 0.1.0",
   "lua-resty-mlcache == 2.5.0",
   "lua-messagepack == 0.5.2",
-  "lua-resty-openssl == 0.6.9",
+  "lua-resty-openssl == 0.6.10",
```
murillopaula, line 39:
```suggestion
  "lua-resty-openssl == 0.6.10-2",
```

context as to why: https://github.com/fffonion/lua-resty-openssl/commit/ea53cf407c9517ec9f5d25ea85de91602bb708b2

murillopaula (changes requested):

kong-2.3.0-0.rockspec
bungle: @murillopaula, version should be enough. At least with my `luarocks` it installs the `-2`.

bungle: Though the tests are failing as we don't have that module installed, so perhaps I need to bump build tools too.

Tieske: @bungle is right, please do not include rockspec revision numbers. They only mean that the rockspec changed, but not the source code. LuaRocks will automatically pick the latest one.

Tieske: Maybe the old version is cached?

hutchic: The test assets are available for introspection if one suspect CI built thing incorrectly

Steal the image tag from [here](https://github.com/Kong/kong/pull/6740/checks?check_run_id=1722826175#step:6:19)
image

and run it
```
docker run -it --rm mashape/kong-build-tools:test-4dd8e8b-0-538950b-03 /bin/bash
luarocks list --porcelain | grep lua-resty-openssl
lua-resty-openssl       0.6.10-2        installed       /usr/local/lib/luarocks/rocks-5.1
```

BIG DISCLAIMER: how this will behave on systems where folks do an in place upgrade I cannot say

kikito (changes requested):
The changes look straightforward to me, but I can't approve this without the tests passing.

bungle:
@kikito, yes I tried to do the changes that @murillopaula did on other repo, but it didn't seem to work here. I am not sure how we should handle this extra module. I think it should in some ways be a part of Kong Build Tools instead. @murillopaula / @fffonion, any thoughts about it?

fffonion:
@bungle I'm more lean to to not compile the module if it's not being used (acme plugin imported resty.openssl, but it's not using resty.openssl.ssl where the warning comes from). We need to lower the log level about using the aux module or only display it when it's actually being used. I'll fix it and create a new release.

murillopaula:
@bungle the module is required for people that want to use `scram-sha-256-plus` passwords in postgres: https://github.com/Kong/kong/issues/6035

So at least in kong build tools we need to have @fffonion 's module compiled.

murillopaula:
Here we'll only need to have the module compiled if the tests complain about the module missing in the logs (what @fffonion was talking about).

fffonion:
@murillopaula I see, for some reason I thought that was EE only feature 🤦 . In that case we do need to fix the build. If somehow it's not workable we can move the helper FFI functions to lua-kong-nginx-module maybe.

fffonion:
Opened #6759 to bump to 0.6.11, where the warning message is shown when actual function is being called and thus will not shown in CI.
The aux module will ultimately go into lua-kong-nginx-module. Before that user can still test it even without the module being compiled, but is not recommanded in a production env.

PR #6743, opened 2021-01-18
fix(core): make entity relationship traverse-order-independent

javierguerragiraldez:
the `nest_foreign_relationships()` function creates "array" fields to
represent entity relationships.  The new field includes a copy of the
included entity.  If the target has already been copied into another
entity, the copy might not get the new field.

This fix keeps track of all copied records and injects the new field
on each copy.


bungle (approved):

dndx:
Nice work, quick question: can this be tested using unit tests?

PR #6744, opened 2021-01-18
feat(termination) add a request-echo option

Tieske:
This PR adds 2 features:

1. a trigger for the plugin to kick-in
2. an option to echo the incoming request back to the client

AD1) if a trigger is set (a header or query-param name), then any request having that header or query parameter will trigger the plugin functionality as configured. If the trigger is undefined in the configuration, the plugin will always kick-in (existing behaviour). If the trigger is defined in the config, but it is not present in the request, the plugin will not do anything, and allow the request to be proxied.

AD2) If set the plugin will echo a copy of the received request back to the client.

Essentially this is a debug aid. The plugin can be added with no impact. Only when the trigger is specified by the client making the request, the plugin will terminate and echo. Another usecase for this would be to turn an existing endpoint into a healthcheck endpoint, only if the trigger is present (configured in the health probe) it responds, otherwise it just proxies.

Alternatively it can be attached to a catch-all route to try and figure out why no route was matched, especially when there are other proxies in between the client and Kong.

EDIT: updated description based on comments and later changes


kong/plugins/request-termination/schema.lua
```diff
@@ -21,6 +21,8 @@ return {
           { message = { type = "string" }, },
           { content_type = { type = "string" }, },
           { body = { type = "string" }, },
+          { echo = { type = "boolean", required = true, default = false }, },
+          { echo_trigger = typedefs.header_name }
```
hbagdi, line 25: How about `echo_trigger_header` to make it a little more descriptive?

Tieske: it triggers on both header as well as query entries by that name.

Tieske: one thing I have been considering; currently the trigger is called `echo_trigger` since it only applies to the echo response. It could be called generically `trigger` and then apply to the entire plugin behaviour.

I think I would prefer that over the current `echo_trigger`. Thoughts? @hbagdi

Tieske: calling it `trigger` would also allow any endpoint to be converted into a healthcheck endpoint, if you configure the probe request to send the `trigger` along

hbagdi: `trigger` makes more sense, yes.
Meta: this small feature is actually a preview of what we have been wanting to Kong for years. A fine-grained way to trigger plugin execution. :)
I don't have any further comment on this thread, so resolving it.

kong/plugins/request-termination/schema.lua
```diff
@@ -32,6 +34,11 @@ return {
           and not is_present(config.body) then
             return nil, "content_type requires a body"
           end
+          if config.echo and (
```
gszr, line 37: I think we can replace this logic with the `mutually_exclusive_sets` entity validator. For example:
`{ mutually_exclusive_sets = { { "echo" }, { "content_type", "body" } }`

Tieske: That would only make sense if we'd replace the entire custom validator with standard entity_checks, I spend 2 hours on it, but it so overwhelmingly under-documented that I couldn't get it to work. So I'm giving up on that. Someone else may send in a separate PR to fix it. Not in scope of this change.

gszr: Agreed on not touching unrelated parts of the validator. However, if the new field can leverage the validator, I think we should do it (leaving it to later creates another breaking change in the future).

gszr: Actually, @Tieske, the `mutually_exclusive_sets` entity check can't be used here (the `echo` field is required and has a default value).

gszr (approved):

gszr:
Good to squash and merge once the tests are green.

Tieske:
@gszr all green now, please squash and merge

PR #6756, opened 2021-01-20
docs(autodoc) scripts/autodoc can install files in docs repo

kikito:
This will work as before, it will generate files in autodoc/output (it will print an info message now)

    scripts/autodoc

The following however will also try to install the generated files in the docs folder for
you:

    scripts/autodoc ../docs.konghq.com 2.3.x

### DIFF
--- a/scripts/autodoc
+++ b/scripts/autodoc
@@ -107,6 +107,7 @@ else
     exit 1
 fi
 
+
 if [ ! -d "$DOCS_REPO/autodoc-nav" ]
 then
     echoerr
@@ -116,7 +117,7 @@ then
     exit 2
 fi
 
-DOCS_APP="$DOCS_REPO/app/$DOCS_VERSION"
+DOCS_APP="$DOCS_REPO/app/gateway-oss/$DOCS_VERSION"
 
 if [ ! -d "$DOCS_APP" ]
 then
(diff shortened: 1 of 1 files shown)


scripts/autodoc
```diff
@@ -1,5 +1,159 @@
...
+        count_indents_on_line "$docs_nav_file" "$current_line"
+        current_indent="$COUNT_INDENTS"
+    done
+    local last_line="$(( current_line - 1 ))"
+
+    # insert_line is where we want to put the new pdk docs after deleting the old ones (one less than first)
+    local insert_line="$(( first_line - 1 ))"
+
+    # delete existing pdk index, insert new pdk nav file instead
+    sed -i '' -e "${first_line},${last_line}d" -e "${insert_line}r${pdk_nav_file}" "$docs_nav_file"
```
kikito, line 63: It has been pointed out that in general, `sed -i ''` (modifying a file in-place, without making a backup copy) is undesirable - so much that it is spelled that way in the sed manual.

I propose that in this particular application that is an ok thing to do because the backup copy already exists: the docs repo is a git repository so we can always `git checkout` to our heart's content.

(Also, in my opinion the sed manual page needs less judgement and more examples)

scripts/autodoc
```diff
@@ -1,5 +1,160 @@
...
+DOCS_VERSION=$2
+
+RED='\033[0;31m'
+NC='\033[0m' # No Color
+function echoerr() {
+    printf '%s%s%s\n' "$RED" "$@" "$NC" 1>&2;
+}
+
+function copy() {
+    printf 'cp %s\t→\t%s' "$1" "$2"
```
gszr, line 13:
```suggestion
    printf 'cp %s\t→\t%s\n' "$1" "$2"
```

gszr (approved):
LGTM. Left a very minor suggestion.

Issue #6694, opened 2021-01-01
Stream Listen Not working as expected 


caryyu:
What i'm trying to do is wanting Kong to act like as the same as Nginx stream config below:

```
stream {
  server {
    listen 7000;
    ssl_preread on;
    resolver 8.8.8.8;
    proxy_pass $ssl_preread_server_name:443;
  }
}
```

# Docker Compose Core Part

```

...

  kong-instance:
    image: kong:2.2.0
    links:
      - kong-database:kong-database
    restart: always
    ports:
      - '8000:8000'
      - '7000:7000'
      - '8443:8443'
      - '8001:8001'
    depends_on:
      - kong-database
    environment:
      - KONG_LUA_PACKAGE_PATH=/opt/?.lua;;
      - KONG_NGINX_DAEMON=off
      - KONG_NGINX_WORKER_PROCESSES=1
      - KONG_PLUGINS=bundled
      - KONG_DATABASE=postgres
      - KONG_PROXY_LISTEN=0.0.0.0:8000,0.0.0.0:8443 ssl
      - KONG_STREAM_LISTEN=0.0.0.0:7000 ssl
      - KONG_NGINX_STREAM_SSL_PREREAD=on
      - KONG_NGINX_STREAM_RESOLVER=8.8.8.8
      - KONG_PG_HOST=kong-database
      - KONG_PG_DATABASE=kong
      - KONG_PG_USER=kong
      - KONG_PG_PASSWORD=kong
      - KONG_PROXY_ACCESS_LOG=/dev/stdout
      - KONG_STREAM_ACCESS_LOG=/dev/stdout
      - KONG_ADMIN_ACCESS_LOG=/dev/stdout
      - KONG_PROXY_ERROR_LOG=/dev/stderr
      - KONG_STREAM_ERROR_LOG=/dev/stderr
      - KONG_ADMIN_ERROR_LOG=/dev/stderr
      - KONG_ADMIN_LISTEN=0.0.0.0:8001
      - KONG_LOG_LEVEL=debug
```

# Test Steps

```
curl -XPOST http://localhost:8001/services -d name=github-proxy -d url=tcp://github.com:443 -d protocol=tcp
curl -XPOST http://localhost:8001/services/github-proxy/routes -d name=github-proxy -d snis=github.com -d protocols=tls
openssl s_client -connect localhost:7000 -servername github.com
```

The output is:

```
...

    Verify return code: 18 (self signed certificate)
---
```

# Expected 

```
[...]

PR #6757, opened 2021-01-21
feat(router) allow TLS passthrough in stream router

### LINKED ISSUE #5902: Support SNI-based routing without TLS termination
### Summary

According to @hbagdi, the current Kong implementation does not support TLS passthrough for HTTPs upstream service, while this appears to be a common scenario for some other reverse proxies such as nginx, haproxy that natively support.

Also, to terminate TLS to retrieve SNI looks a bit odd, but that is understood since TLS passthrough is missing at the moment.

Right now, Kong can do either of the two things:
* Send all traffic on on a specific port to an upstream service
* Terminate TLS connection on a specific port, and then route traffic to different upstream services based on the SNI

Kong cannot sniff SNI, not terminate the TLS and route traffic to an upstream service.

More details could be found at: https://discuss.konghq.com/t/does-kong-ingress-controller-support-tls-passthrough-for-https-upstream-service/6268

fffonion:
### Summary

Adds the ability in route TLS traffic based on SNI without terminating the connection.

To let Kong does normal TLS termination

```
kong.conf:
stream_listen=0.0.0.0:9000 ssl

kong.yaml
routes:
- protocol: tls
  snis: something
  tls_passthrough: false #default
```

To let Kong does TLS passthrough

```
kong.conf:
stream_listen=0.0.0.0:9000 ssl

kong.yaml
routes:
- protocol: tls
  snis: something
  tls_passthrough: true
```

### Full changelog

* feat(router) allow TLS passthrough in stream router

### Issues resolved

Fix #5902 #6694

TODO:
- [x] preserve client address when doing second layer proxing
- [x] not run plugins in log phase in the preread server block

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -742,6 +742,12 @@ function Kong.preread()
 
   runloop.preread.before(ctx)
 
+  -- if proxying to a second layer TLS terminator is required
+  -- abort further execution and return back to Nginx
+  if ctx.stream_proxy_preread_terminate then
+    return
+  end
+
   local plugins_iterator = runloop.get_updated_plugins_iterator()
   execute_plugins_iterator(plugins_iterator, "preread", ctx)
 
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -740,11 +740,11 @@ function Kong.preread()
 
   log_init_worker_errors(ctx)
 
-  runloop.preread.before(ctx)
+  local preread_terminate = runloop.preread.before(ctx)
 
   -- if proxying to a second layer TLS terminator is required
   -- abort further execution and return back to Nginx
-  if ctx.stream_proxy_preread_terminate then
+  if preread_terminate then
     return
   end
 
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1132,8 +1132,7 @@ return {
           return exit(500)
         end
 
-        ctx.stream_proxy_preread_terminate = true
-        return
+        return true
       end
 
 
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -76,6 +76,9 @@ local get_updated_router, build_router, update_router
 local server_header = meta._SERVER_TOKENS
 local rebuild_router
 
+local stream_tls_terminate_sock = "unix:" .. ngx.config.prefix() .. "/stream_tls_terminate.sock"
+local stream_tls_passthrough_sock = "unix:" .. ngx.config.prefix() .. "/stream_tls_passthrough.sock"
+
 -- for tests
 local _set_update_plugins_iterator
 local _set_update_router
@@ -1122,10 +1125,10 @@ return {
         local protocols = route.protocols
         if protocols and protocols.tls then
           log(DEBUG, "TLS termination required, return to second layer proxying")
-          var.kong_tls_preread_block_upstream = "unix:" .. kong.configuration.prefix .. "/stream_tls_terminate.sock"
+          var.kong_tls_preread_block_upstream = stream_tls_terminate_sock
 
         elseif protocols and protocols.tls_passthrough then
-          var.kong_tls_preread_block_upstream = "unix:" .. kong.configuration.prefix .. "/stream_tls_passthrough.sock"
+          var.kong_tls_preread_block_upstream = stream_tls_passthrough_sock
 
         else
           log(ERR, "unexpected protocols in matched Route")
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1113,9 +1113,32 @@ return {
         return exit(500)
       end
 
+      local route = match_t.route
+      -- if matched route doesn't do tls_passthrough and we are in the preread server block
+      -- this request should be TLS terminated; return immediately and not run further steps
+      -- (even bypassing the balancer)
+      local tls_preread_block = var.kong_tls_preread_block
+      if tls_preread_block and #tls_preread_block > 0 then
+        local protocols = route.protocols
+        if protocols and protocols.tls then
+          log(DEBUG, "TLS termination required, return to second layer proxying")
+          var.kong_tls_preread_block_upstream = "unix:" .. kong.configuration.prefix .. "/stream_tls_terminate.sock"
+
+        elseif protocols and protocols.tls_passthrough then
+          var.kong_tls_preread_block_upstream = "unix:" .. kong.configuration.prefix .. "/stream_tls_passthrough.sock"
+
+        else
+          log(ERR, "unexpected protocols in matched Route")
+          return exit(500)
+        end
+
+        ctx.stream_proxy_preread_terminate = true
+        return
+      end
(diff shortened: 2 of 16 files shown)


hbagdi:
I love it when such features can be implemented with a small patch.
Question: Is it possible to mix TLS termination and pass-through (based on the SNI) on the same port? If not, what is the limitation? If yes, how much effort is that?

fffonion:
@hbagdi Having them on the same port is not likely something Nginx support directly, but can be hacked by configuration. If one configures two "layers" of proxy, the first layer doesn't do TLS termination and when it matches certain SNI, it forward it a local port, which is a TLS port. Then theoritically you can have the port in the first layer accepts both type of traffic.

hbagdi:
That sounds like an uncomfortable and a risky hack to me that we should best avoid.

fffonion:
It's actually not that terrible as it looks 😄 and it will be transparent to client. By configuration I was referring to two Routes in Kong, the nginx configuration doesn't need to be changed. Though it's indeed not a common case to use Kong itself as a upstream, but it's a common pattern in Nginx as I heard of.

dndx:
@hbagdi In reality proxying to itself does not impact performance that much. The bulk of the overhead is always inside the TLS termination which has to run expensive cryptography operations. Proxy to self, in reality is just few `memcpy`s and can easily saturate the memory bandwidth without breaking a sweat of the CPU.

I very much feel that we need to support mixing terminated and non-terminated connections on the same port, otherwise the usefulness of Kong TLS proxy will become very limited.

hbagdi:
> I very much feel that we need to support mixing terminated and non-terminated connections on the same port, otherwise the usefulness of Kong TLS proxy will become very limited.

I do too. But it has to be transparent to the user, meaning, a user shouldn't need to create two routes in Kong.
If I create 3 routes on the same port in kong:
- a route performs TCP termination
- a route that does TLS termination for an SNI
- a route that does SNI-based proxying without TLS termination (SNI sniffing)

it should just work.

fffonion:
Seems like we need to add more lines in this PR : ) Turning on `ssl_preread`, or if there's any plugin or Kong logic that tries to peek data in preread phase will block current request in preread phase for protocols where "server sends data before client" (for example SMTP). Kong/Nginx only starts to proxy data from upstream to downstream in phase after preread.

https://github.com/Kong/kong/blob/master/spec/02-integration/05-proxy/22-reports_spec.lua#L41 is an example for this use case.

So I'm proposing to make `ssl_preread` gated by a new Kong configuration so that user can choose the behaviour of stream proxy. This will has no effect to HTTP subsystem.

dndx:
Discussed with @fffonion, the following design are proposed:

1. We keep the L4 configuration exactly the same as today.
2. When user specifies `stream_listen`, `ssl` can be specified for a port as needed (this is the same as today). If the user wants to perform any special TLS handling, like SNI proxy or TLS termination, then the `ssl` flag must be present.
3. When `ssl` flag is present, the `preread` module will be enabled for that `server` block. At the same time, another `server` that listens on a unix domain socket will be created that will handle TLS termination if later needed.
4. When request comes into a port with `ssl` flag, the incoming SNI `server_name` is examined. If termination is required, the request is forwarded via the domain socket to the terminating `server` block.

The advantage of this method is:

1. Support the current user config as well as mixed terminating and non-terminating TLS proxy, without need to introduce additional config options
2. Minimum loss of performance

Note that we do not want to support mixing TLS and cleartext traffic on the same port, for reasons @fffonion already mentioned above. It will inevitable break a lot of cleartext protocol if we enable `preread` module for cleartext ports. User must always specify in `stream_listen` whether a port proxies TLS traffic or not, but we can decide whether to terminate the TLS.

hbagdi:
+1 to that.
One question, based on the SNI, how will Kong decide to terminate or TLS-passthrough the request?

fffonion:
@hbagdi there will be two server blocks, since currently we can't control ssl termination dynamically in Lua

```
match route based on SNI  --- Service is tls --> unix domain socket -> local server block with tls termination -> Upstream
                        \ ___ Service is tcp -> Upstream
```

hbagdi:
So if I'm understanding correctly, weather to terminate TLS or not will be
based on the protocol of Service? That seems odd. Routing decisions should
be based on routes only, isn't it?

fffonion:
@hbagdi That has similar pattern in http subsystem, Service will determine how Kong "treat" the request.
For example you can route an SNI route to a http service, but Kong will complain it should only be used in https service.
Then whether to serve certificate is based on the protocol of Service, it shouldn't be regarded as a routing decision.

hbagdi:
>  For example you can route an SNI route to a http service, but Kong will complain it should only be used in https service.

AFAIK, that error is generated at the time of route/service creation and not at runtime.

> Then whether to serve certificate is based on the protocol of Service, it shouldn't be regarded as a routing decision.

This doesn't seem correct. In HTTP sub-system, the certificate presented to downstream has nothing to do with Service resource because the route is not known during TLS (and hence Service is not known). I think the stream sub-system works similarly but I don't have enough experience with it to be sure.

In Routes, we have the protocol - tcp or tls, SNI and port. I think what is missing is a field which dictates whether to terminate TLS or not.

haroldoproenca:
Hi, How are you?
I'm facing the same problem with my Kong Ingress Controller and I would like to know when you intend to release this fix.
Thank's,

Haroldo

mayocream:
Any updates on this PR?

fffonion:
@haroldoproenca @mayocream We will revisit this PR in the near future releases, please hold for updates!

ignacioli:
Any updates on this PR?

Thanks

kong/runloop/handler.lua
```diff
@@ -1113,9 +1113,32 @@ return {
...
+      local route = match_t.route
+      -- if matched route doesn't do tls_passthrough and we are in the preread server block
+      -- this request should be TLS terminated; return immediately and not run further steps
+      -- (even bypassing the balancer)
+      local tls_preread_block = var.kong_tls_preread_block
+      if tls_preread_block and #tls_preread_block > 0 then
+        local protocols = route.protocols
+        if protocols and protocols.tls then
+          log(DEBUG, "TLS termination required, return to second layer proxying")
+          var.kong_tls_preread_block_upstream = "unix:" .. kong.configuration.prefix .. "/stream_tls_terminate.sock"
```
dndx, line 1125: This seems to be something we can cache once at load and no need to create new strings for each connection.

kong/init.lua
```diff
@@ -742,6 +742,12 @@ function Kong.preread()
 
   runloop.preread.before(ctx)
 
+  -- if proxying to a second layer TLS terminator is required
+  -- abort further execution and return back to Nginx
+  if ctx.stream_proxy_preread_terminate then
```
dndx, line 747: Instead of checking the flag, if we add a return value for `runloop.preread.before` does that makes more sense? Flags are slower but more important it makes the control flow less clear :)

dndx (approved):

PR #6765, opened 2021-01-25
fix(sandbox) fix validate_function and validate

eskerda:
validator function was returning an incorrect error message on execution error.


gszr (approved):
💯

gszr (commented):
Actually, @eskerda, could you base this one on master, so it goes into the patch release?

eskerda:
> Actually, @eskerda, could you base this one on master, so it goes into the patch release?

sure!

eskerda:
@gszr all set!

PR #6769, opened 2021-01-26
chore(ci) fix jenkins conditional stage failure

gszr:
`tag pattern` is not working in the context of `when` / `allOf`


hutchic (approved):
this bug annoys me. Can confirm the PR is correct

PR #6775, opened 2021-01-27
fix(declarative) add unique validation to declarative:parse_table

### LINKED ISSUE #6732: DB-less mode allows resources with duplicate unique fields
### Summary
Loading a declarative configuration that contains resources with identical unique field values but distinct IDs creates both resources. This affects at least consumers and services, and presumably affects all other entities.

### Steps To Reproduce

1. Start Kong with `KONG_DATABASE=off` set.
1. Load [dup_consumer_id.yaml.txt](https://github.com/Kong/kong/files/5811231/dup_consumer_id.yaml.txt)
2. Check the `GET /consumers` endpoint. Both resources exist, e.g. 
[example_output.txt](https://github.com/Kong/kong/files/5811211/example_output.txt)

### Additional Details & Logs

Tested on Kong 2.3.0.

Kong also accepts similar configuration without IDs, e.g. [dup_consumer_id.yaml.txt](https://github.com/Kong/kong/files/5811238/dup_consumer_id.yaml.txt). Excluding IDs overwrites results in a single resource (the last resource defined in the configuration) only. However, if other resources are linked (e.g. service->plugin relationships), the result is two distinct resources, e.g. from 
[dup_consumer_noid_plugin.yaml.txt](https://github.com/Kong/kong/files/5811254/dup_consumer_noid_plugin.yaml.txt).

In practice I don't think this is likely to cause serious issues--attempting to request resources from the admin API by unique field no longer really works (it only returns one entity), but routing logic presumably still entirely relies on IDs. The overwrites _may_ be a concern, but I'm unsure if they'd matter because the override doesn't occur when entities are linked to others, and offhand I'm not aware of entities that do much of anything when not linked.

One practical implication is that the ingress controller can easily overwrite entities unintentionally as a result of user error, but the same caveat about linking applies (I only encountered this because I was intentionally trying to cause a conflict with unlinked consumers).

Throwing this out there in case anyone can think of more severe implications. If not, and we don't want to bother changing this behavior, it should be fine to simply document this.

bungle:
### Summary

@rainest reported in #6732 that declarative config is not checked over unique violations.

### Issues Resolved

Fix #6732 (for unique errors, but having multiple entities for same primary key is not reported as error yet).

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -20,6 +20,16 @@ local SUBSYS = ngx.config.subsystem
 local WORKER_COUNT = ngx.worker.count()
 local DECLARATIVE_HASH_KEY = constants.DECLARATIVE_HASH_KEY
 
+
+local function find_default_ws(entities)
+  for _, v in pairs(entities.workspaces or {}) do
+    if v.name == "default" then
+      return v.id
+    end
+  end
+end
+
+
 local declarative = {}
 
 
@@ -238,6 +248,11 @@ function Config:parse_table(dc_table, hash)
 
   if not self.partial then
     self.schema:insert_default_workspace_if_not_given(entities)
+  end
+
+  local ok, err_t = self.schema:check_constraints(entities)
+  if not ok then
+    return nil, pretty_print_error(err_t), err_t
   end
 
   if not hash then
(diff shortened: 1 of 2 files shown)


hanshuebner (changes requested):
A couple of tests would make sense.

hbagdi:
@bungle Please rebase.

PR #6796, opened 2021-02-02
fix(balancer) update the Host header between balancer retries

murillopaula:
Sister PR: https://github.com/Kong/kong-build-tools/pull/378


.requirements
```diff
@@ -8,4 +8,4 @@ RESTY_OPENSSL_VERSION=1.1.1j
 RESTY_PCRE_VERSION=8.44
 LIBYAML_VERSION=0.2.5
 KONG_GO_PLUGINSERVER_VERSION=v0.6.1
-KONG_BUILD_TOOLS_VERSION=4.13.0
+KONG_BUILD_TOOLS_VERSION=37ece0b806501ba85007cd0efba117e58a6fb78f
```
murillopaula, line 11: note: this is a commit SHA that points to this required kong-build-tools patch: https://github.com/Kong/kong-build-tools/commit/37ece0b806501ba85007cd0efba117e58a6fb78f

todo: merge https://github.com/Kong/kong-build-tools/pull/378, bump kong-build-tools and pin the new version here

```suggestion
KONG_BUILD_TOOLS_VERSION=4.14.0
```

kong/runloop/handler.lua
```diff
@@ -1318,33 +1322,22 @@ return {
...
-          end
+      -- the nginx grpc module does not offer a way to overrride the
+      -- :authority pseudo-header; use our internal API to do so
+      local upstream_host = var.upstream_host
+      local upstream_scheme = var.upstream_scheme
+
+      if upstream_scheme == "grpc" or upstream_scheme == "grpcs" then
+        ok, err = kong.service.request.set_header(":authority", upstream_host)
+        if not ok then
+          log(ERR, "failed to set :authority header: ", err)
```
murillopaula, line 1340: @gszr please review this change (conflict resolution when rebasing)

kong/runloop/handler.lua
```diff
@@ -1318,33 +1322,22 @@ return {
...
-          then
-            upstream_host = upstream_host .. ":" .. balancer_data.port
-          end
+      local ok, err = balancer.set_host_header(balancer_data)
+      if not ok then
+        ngx.log(ngx.ERR, "failed to set balancer Host header: ", err)
 
-          var.upstream_host = upstream_host
-        end
+        return ngx.exit(500)
```
locao, line 1329: The most probable cause of failure here is `ngx_create_request` not being able to allocate buffers. So, this is a temporary condition, and we know what happened. Therefore, a 503 seems more appropriate to me.

dndx: I think either one is fine here. We have traditionally used 500 to indicate any kind of error Kong experiences during request processing, so it should be okay to use 500 here as well.

It may be difficult to determine if the failure reason is temporary or not here (memory leaks, for example, could last for a long time until being manually corrected). So I'm okay with keeping 500 here.

kong/runloop/handler.lua
dndx: Should this be moved to `set_host_header` instead? Otherwise on gRPC request retries the header will also not be updated.

dndx (approved):
Need a new KBT release, other than that it LGTM :)

Thanks @murillopaula for the hard work!

murillopaula:
Please squash and merge

PR #6800, opened 2021-02-03
review.me

CHAITANYASAI-del:
One thought here: should exit early out of this loop (via break) if we are omitting headers, and we've reached this condition? Once we've assigned stop = true, we're not doing any other work aside from assigning headers, so breakingnoit early would save some work. Don't know if this is worth it though, so this isn't a blocker to me, just a thought.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX


kikito:
Hello, thank you for your PR. It is unfortunately too incomplete to be actionable/reviewable. Please submit an issue for discussions/proof of concepts such as this.

PR #6802, opened 2021-02-03
fix(rate-limiting) remove unneeded phase handlers for better accuracy

bungle:
### Summary

The PR #5460 tried to fix rate-limiting accuracy issues as reported with #5311. That PR has stayed open for over a year now.

There is a smaller change in that PR that we can make without a lot of work or debating, and this PR contains that only:

1. it removes extra phase handlers and context passing of data and now does everything in `access` phase (this could make it more accurate as well, compared to post-poning that to log phase)
2. it now also sets rate-limiting headers when the limits have exceeded


kikito (approved):
LGTM, let's merge this when green

PR #6805, opened 2021-02-04
fix(*) do not run non-needed code on roles that don't need them

bungle:
### Summary

This commit changes the code to check not run certain initialization on certain modes:

#### init phase

1. do not build router on control planes

#### init worker phase

1. do not execute cache warmup on control planes
2. do not start plugin servers on control planes
3. do not initialize balancer on control planes (also disables healthchecks)
4. do not start async router rebuild timer on control planes
5. do not start async plugin iterator rebuild timer on control planes
6. do not subscribe on events with dbless

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -492,7 +492,7 @@ function Kong.init()
     stream_api.load_handlers()
   end
 
-  if kong.configuration.database == "off" then
+  if config.database == "off" then
     local err
     declarative_entities, err, declarative_meta = parse_declarative_config(kong.configuration)
     if not declarative_entities then
@@ -508,7 +508,9 @@ function Kong.init()
       error("error building initial plugins: " .. tostring(err))
     end
 
-    assert(runloop.build_router("init"))
+    if config.role ~= "control_plane" then
+      assert(runloop.build_router("init"))
+    end
   end
 
   db:close()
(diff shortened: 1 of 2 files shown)


bungle:
The considerations and questions:

> do not build router on control planes

I think that is rather safe.

> do not execute cache warmup on control planes

It is a warmup, so I would say it is safe.

> do not start plugin servers on control planes

Go plugins don't have init_worker phase AFAIK, so control planes don't need plugin servers. But do we need to start servers for validating plugin configurations?

> do not initialize balancer on control planes (also disables healthchecks)

After that I am not sure how admin api endpoints would work, or do they work at all? But even if they work, does that reflect to dps anyway? Is there something that is needed in balancer so that we need to initialize it on control plane? @locao?

> do not start async router rebuild timer on control planes

Control planes don't route traffic, so they don't need router.

> do not start async plugin iterator rebuild timer on control planes

Control planes don't run request phase plugins, and init worker plugins. So I don't think they need to update plugin iterator. It is also debatable whether control planes should even run init_worker phase of the plugins, or is that only meant for data planes in hybrid?

> do not subscribe on events with dbless

Is there reason the dbless nodes should subscribe to these cluster and or worker events which are mostly crud type of events triggered by admin action, while the dps themselves are read-only?

bungle:
So, reviewers, please check above. For a patch release, should we comment something out for now?

dndx (approved):
Did a quick look and the change seems to be pretty safe to me. But appreciate if @locao can also take a look.

locao (approved):
LGTM 🚀

PR #6828, opened 2021-02-10
feat(cassandra) expose TLS encryption version from lua-cassandra

jeremymv2:
Relies on this PR to merge and ship first: https://github.com/thibaultcha/lua-cassandra/pull/141

Closes FTI-2249

Signed-off-by: Jeremy J. Miller <jeremy.miller@konghq.com>

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -143,6 +143,7 @@ function CassandraConnector.new(kong_config)
     timeout_read              = kong_config.cassandra_timeout,
     max_schema_consensus_wait = kong_config.cassandra_schema_consensus_timeout,
     ssl                       = kong_config.cassandra_ssl,
+    encryption_protocol       = kong_config.cassandra_ssl_encryption_protocol,
     verify                    = kong_config.cassandra_ssl_verify,
     cafile                    = kong_config.lua_ssl_trusted_certificate_combined,
     lock_timeout              = 30,
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1020,6 +1020,10 @@
 #cassandra_ssl = off             # Toggles client-to-node TLS connections
                                  # between Kong and Cassandra.
 
+#cassandra_ssl_encryption_protocol = tlsv1      # When using ssl between Kong and Cassandra,
+                                                # the version of tls to use. Accepted values are
+                                                # `tlsv1` or `tlsv1_2`
+
 #cassandra_ssl_verify = off      # Toggles server certificate verification if
                                  # `cassandra_ssl` is enabled.
                                  # See the `lua_ssl_trusted_certificate`
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -456,6 +456,7 @@ local CONF_INFERENCES = {
   cassandra_password = { typ = "string" },
   cassandra_timeout = { typ = "number" },
   cassandra_ssl = { typ = "boolean" },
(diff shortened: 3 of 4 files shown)


kong/db/strategies/cassandra/connector.lua
```diff
@@ -143,6 +143,7 @@ function CassandraConnector.new(kong_config)
     timeout_read              = kong_config.cassandra_timeout,
     max_schema_consensus_wait = kong_config.cassandra_schema_consensus_timeout,
     ssl                       = kong_config.cassandra_ssl,
+    encryption_protocol       = kong_config.cassandra_ssl_encryption_protocol,
```
thibaultcha, line 146: Specifying a protocol here will indeed make lua-cassandra's LuaSec fallback use the specified TLS version. *However*, the LuaSec fallback is only used for OpenResty phases that do not support cosockets (https://github.com/openresty/lua-nginx-module#cosockets-not-available-everywhere). The TLS version of cosocket connections is determined by the [lua_ssl_protocols](https://github.com/openresty/lua-nginx-module#lua_ssl_protocols) directive instead, so this will only work for the init phases and few others were the LuaSec fallback is used, and not in the proxy path.

jeremymv2: @thibaultcha thanks for pointing out the phases where cosockets are not available. I was not aware of this. Follow up question for you. If LuaSec covers the phases where cosockets are not implemented, and where cosockets are available, the TLS version defaults to a list `lua_ssl_protocols SSLv3 TLSv1 TLSv1.1 TLSv1.2`, doesn't that then mean that we effectively have the capability to utilize (by configuration in this PR) any supported TLS version across 100% of the phases? If that is true, then it addresses the problem I'm trying to solve, where for instance in the init phase, we cannot control the TLS version being used by LuaSec so connections fail when TLSv1.2 is required.

thibaultcha: It seems to me as if this new property is trying to solve two distinct use-cases at once:
A. Supporting TLS 1.2 in non-cosocket phases (FTI-2249)
B. Introducing a way to configure the TLS version to use between Kong & Cassandra

While this new property does solve A. (upon user intervention to update its default), it does not guarantee B. across *all* connections opened with Cassandra. As of today (with and without this patch), `cassandra_ssl_encryption_protocol = tlsv1` and a different `lua_ssl_protocols` directive could already cause disparities between cosocket and LuaSec connections.
Besides, even if we were to try and specify `lua_ssl_protocols`, we would be affecting all workers' cosockets, which could break other connections in the gateway or established by plugins to other services. In order to fine-grain Kong <-> C* TLS connections specifically, I think we'd want to rely on [tcpsock:tlshandshake()](https://github.com/openresty/lua-nginx-module/pull/1602) to add a new option (this OpenResty patch is bundled with Kong).

Therefore in light of the above, it does not seem like a good time to address B. at the moment. 

However, maybe we could still fix A. by:
1. Making TLS 1.2 the new hard-coded default in lua-cassandra
2. Having lua-cassandra fallback to TLS 1.1 for backwards-compatibility (maybe by way of catching an invalid protocol error, or maybe by looking at newer LuaSec APIs?)

This way we'd still allow non-cosocket phases to default to TLS 1.2 much like cosockets. Later, we may revisit use-case B., which is about _specifically configuring Kong to Cassandra connections_; since I don't believe it to be a popular ask (when we distinguish it from A. that is).

jeremymv2:
After an internal discussion with @murillopaula @thibaultcha we have reached an agreement on how to proceed. I will address the problem in `lua-cassandra`. Closing this PR.

PR #6842, opened 2021-02-18
fix(admin) do not send prng_seeds but instead send pids

bungle:
### Summary

- Removes `prng_seeds` from shared dictionaries
- Removes `prng_seeds` from local variable
- Removes `prng_seeds` from `:8001/`
- Adds `pids` to `:8001/`

### DIFF
--- a/kong/globalpatches.lua
+++ b/kong/globalpatches.lua
@@ -228,92 +228,96 @@ return function(options)
     -- This patched method will create a unique seed per worker process,
     -- using a combination of both time and the worker's pid.
     local util = require "kong.tools.utils"
-    local seeds = {}
+    local seeded = {}
     local randomseed = math.randomseed
 
     _G.math.randomseed = function()
       local pid = ngx.worker.pid()
-      local seed = seeds[pid]
-      if not seed then
-        local phase = ngx.get_phase()
-        if not options.cli and (phase ~= "init_worker" and phase ~= "init") then
-          ngx.log(ngx.WARN, debug.traceback("math.randomseed() must be " ..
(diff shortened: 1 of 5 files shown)


kong/globalpatches.lua
```diff
@@ -228,92 +228,96 @@ return function(options)
...
+      if phase == "init" then
+        id = "master"
+        is_seeded = seeded.master
 
-        local bytes, err = util.get_rand_bytes(8)
-        if bytes then
-          ngx.log(ngx.DEBUG, "seeding PRNG from OpenSSL RAND_bytes()")
+      else
+        id = ngx.worker.id()
+        is_seeded = seeded[pid]
```
kikito, line 245: I am not sure I understand the logic behind this part.

What worries me is: what happens when worker(s) is/are restarted by the master process? According to the [docs for `ngx.worker.id`](https://github.com/openresty/lua-nginx-module#ngxworkerid), :

> Returns the ordinal number of the current Nginx worker processes (starting from number 0).
> 
> So if the total number of workers is N, then this method may return a number between 0 and N - 1 (inclusive).

So if we have 5 workers, and worker with `id = 2` gets killed, it will be restarted an its `id` will also be 2 (all the other numbers between 0 and N will be taken). Which means that this piece of code will consider that `is_seeded` is `true`. So we will hit the "attempt to seed already seeded random number" debug message a bit below and will return, not calling "the real `randomseed`" on that new worker that was created.

But maybe I am missing something.

bungle: It uses `pid` the new worker has different `pid`.

kikito (approved):

PR #6850, opened 2021-02-19
fix(balancer) pass ctx to balancer.execute

bungle:
### Summary

There seems to be a bug where we didn't pass `ctx` always to `balancer.execute`, this should fix it.


locao (approved):
😮

PR #6864, opened 2021-02-24
fix(utils) accept hosts ending in dots

locao:
According to [RFC 3986 - section 3.2.2](https://tools.ietf.org/html/rfc3986#section-3.2.2) hosts can end in dots

Kong/lua-resty-dns-client#122 is a sister PR and this one should not be merged before the `lua-resty-dns-client` release with that fix included.


kong/tools/utils.lua
```diff
@@ -942,9 +942,12 @@ _M.check_hostname = function(address)
   end
 
   -- Reject prefix/trailing dashes and dots in each segment
-  -- note: punycode allowes prefixed dash, if the characters before the dash are escaped
-  for _, segment in ipairs(split(name, ".")) do
-    if segment == "" or segment:match("-$") or segment:match("^%.") or segment:match("%.$") then
+  -- notes:
+  --   - punycode allowes prefixed dash, if the characters before the dash are escaped
```
kikito, line 946: Let's fix the typos while we're here
```suggestion
  --   - punycode allows prefixed dash, if the characters before the dash are escaped
```

kikito (approved):
Approved with minor typo notice

gszr (changes requested):
Should this PR target `next` instead, as we do not usually do library bumps in master?

Tieske:
imo, the lib only had a bugfix, so `master` should be fine as well, no?

PR #6821, opened 2021-02-09
fix(router) use normalized request URI when matching against incoming user requests

dndx:

PR #6868, opened 2021-02-25
fix(router) sanitize path postfix

bungle:
### Summary

The PR #6821 by @dndx implemented path normalization, but there is another case with path stripping where
the uri postfix is not sanitized correctly for a better security. This commit fixes that.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -770,6 +770,27 @@ local function categorize_route_t(route_t, bit_category, categories)
 end
 
 
+local function sanitize_uri_postfix(uri_postfix)
+  if not uri_postfix or uri_postfix == "" then
+    return uri_postfix
+  end
+
+  if uri_postfix == "." or uri_postfix == ".." then
+    return ""
+  end
+
+  if sub(uri_postfix, 1, 2) == "./" then
+    return sub(uri_postfix, 3)
+  end
+
+  if sub(uri_postfix, 1, 3) == "../" then
+    return sub(uri_postfix, 4)
+  end
+
+  return uri_postfix
+end
+
+
 do
   local matchers = {
     [MATCH_RULES.HOST] = function(route_t, ctx)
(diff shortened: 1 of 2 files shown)


kong/router.lua
```diff
@@ -770,6 +770,27 @@ local function categorize_route_t(route_t, bit_category, categories)
...
+  if not uri_postfix or uri_postfix == "" then
+    return uri_postfix
+  end
+
+  if uri_postfix == "." or uri_postfix == ".." then
+    return ""
+  end
+
+  if sub(uri_postfix, 1, 2) == "./" then
+    return sub(uri_postfix, 3)
```
bungle, line 783: @dndx,

do you think this should be:
```
    return sub(uri_postfix, 3)
```

or 

```
    return sub(uri_postfix, 2)
```

kong/router.lua
```diff
@@ -770,6 +770,27 @@ local function categorize_route_t(route_t, bit_category, categories)
...
+  end
+
+  if sub(uri_postfix, 1, 3) == "../" then
+    return sub(uri_postfix, 4)
```
bungle, line 787: @dndx,

do you think this should be:

```
    return sub(uri_postfix, 4)
```

or

```
    return sub(uri_postfix, 3)
```

bungle (commented):
Two questions about this PR.

kong/router.lua
dndx: I think `return sub(uri_postfix, 3)` is correct, if there isn't a `/` at the beginning, we should not add another one either. Although from the path handling code it seems to also detect for double slashes, but better to keep it the same. The current code looks fine to me.

kong/router.lua
dndx: See above.

dndx (approved):
LGTM. Appreciate any other 👀 too.

kong/router.lua
bungle: @dndx, it mostly affects the `v1` path handling.

dndx: @bungle Will it break `v1` path handling?

kikito:
Just to confirm: I gave this another 👀and the fix looks ok to me. I also think removing the `/` from `./` in those two places is ok (we are in "quirks mode" so both options look similarly adequate to me. The user might not get what they expected in any case, but they can just not use ./).

Issue #6546, opened 2020-11-05
How to use custom environment variable in nginx_kong.lua script?


DreamingDevs:
I want to customize the `nginx_kong.lua` script by passing custom environment variables like `API_HOST`, `API_URL`.

I am setting the same environment variable as shown below.

```
export KONG_API_HOST=localhost:8443
export API_HOST=localhost:8443
```

And then trying to use it in `nginx_kong.lua` as shown below.

```
proxy_set_header      Host               '${{api_host}}';
```

But it is rendering as shown below.

```
proxy_set_header      Host               '';
```

How to use custom environment variables in `nginx_kong.lua`?

kikito:
Hello, you seem to be doing the correct thing. How are you starting kong?

Also, what version of kong are you using, and what environment (Linux distro, Docker/Kubernetes...)

Tieske:
@kikito do we support setting variables directly in the template? Doesn't the template get the output from the configuration parser, and hence not the custom environment variables?

zmeyc:
@Tieske Also encountered this issue, is there any way to forward environment variables to nginx_kong.lua and ultimately to generated nginx-kong.conf ?

Tieske:
see https://github.com/Kong/kong/pull/6872

PR #6872, opened 2021-02-26
feat(template) allow os.getenv in the template renderer

Tieske:
This enables code like this in the templates:

```
some_directive $(os.getenv("SOME_ENV_VAR"));
```

Fixes: #6546

### DIFF
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -210,7 +210,10 @@ local function compile_conf(kong_config, conf_template)
     _escape = ">",
     pairs = pairs,
     ipairs = ipairs,
-    tostring = tostring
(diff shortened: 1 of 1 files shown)


bungle:
I am not sure about this. If they are already writing their own `templates` why cannot they just add a couple of `env blaablaa;`  directives in it and then use those?

bungle:
Or perhaps at least make it more clever? E.g.:

```
$(ENV["SOME_ENV_VAR"])
```

or something.

Tieske:
Thought about using `env` but then it keeps exposing those variables, whilst this only allows them in the template.

Not sure about the "cleverness", the whole thing is a Lua template renderer, so to me it makes most sense to stick to that.

bungle (approved):

PR #6878, opened 2021-03-01
feat(proxy) implement delayed response with stream module

bungle:
### Summary

Implements delayed responses with stream module that is used to short-circuit streams.

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -264,18 +264,21 @@ local function execute_init_worker_plugins_iterator(plugins_iterator, ctx)
 end
 
 
-local function execute_access_plugins_iterator(plugins_iterator, ctx)
+local function execute_collecting_plugins_iterator(plugins_iterator, phase, ctx)
   local old_ws = ctx.workspace
 
   ctx.delay_response = true
 
-  for plugin, configuration in plugins_iterator:iterate("access", ctx) do
+  for plugin, configuration in plugins_iterator:iterate(phase, ctx) do
     if not ctx.delayed_response then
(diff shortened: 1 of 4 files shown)


locao:
`spec/02-integration/05-proxy/22-reports_spec.lua:383` is consistently failing now. Should the test be adapted to this new feature?

kong/init.lua
```diff
@@ -281,6 +285,8 @@ local function execute_plugins_iterator(plugins_iterator, phase, ctx)
     if old_ws then
       ctx.workspace = old_ws
     end
+
+    ::continue::
```
javierguerragiraldez, line 289: this is a behaviour change: these two things (`reset_log` and resetting the workspace) are currently executed in the `delay_response and ctx.delayed_response` case, but with this patch they're skipped.  is that a desired change?

kong/pdk/response.lua
```diff
@@ -861,21 +890,23 @@ local function new(self, major_version)
...
+      local ctx = ngx.ctx
+      ctx.KONG_EXITED = true
 
-        else
-          self.log.err("unable to proxy stream connection, " ..
-                       "status: " .. status .. ", err: ", body)
-        end
-      end
+      if ctx.delay_response and not ctx.delayed_response then
+        ngx.log(ngx.ERR, "STREAM DELAYED")
```
javierguerragiraldez, line 897: is this a real error, or a debug leftover?

bungle: Yes, a left-over, thank you!

dndx (approved):

PR #6885, opened 2021-03-03
feat(plugins) role based defaults

bungle:
### Summary

Some plugins have default options that don't work well with `dbless` or `hybrid` deployments mostly because the database is not available on gateway nodes. This PR adjusts defaults and validation rules on those plugins.

### DIFF
--- a/kong/plugins/rate-limiting/schema.lua
+++ b/kong/plugins/rate-limiting/schema.lua
@@ -23,6 +23,42 @@ local function validate_periods_order(config)
 end
 
 
+local function is_dbless()
+  local _, database, role = pcall(function()
+    return kong.configuration.database,
+           kong.configuration.role
+  end)
+
+  return database == "off" or role == "control_plane"
+end
+
+
+local policy
+if is_dbless() then
+  policy = {
+    type = "string",
+    default = "local",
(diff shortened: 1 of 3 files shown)


kong/plugins/rate-limiting/schema.lua
```diff
@@ -23,6 +23,52 @@ local function validate_periods_order(config)
...
+  local ok, role = pcall(function()
+    return kong.configuration.role
+  end)
+
+  if not ok then
+    return nil
+  end
+
+  return database == "off" or role == "control_plane" or role == "data_plane" or nil
+end
```
dndx, line 44: Just curious, any chance we can make this a PDK function? It seems to be a lot of code we are copying in this case...

kikito: I just made the function a bit smaller, and it is much less bothersome to copy-paste it in two places now. Maybe if we need it in more places we can abstract it.

kikito:
I have reviewed the rest of the plugins and I feel only the ratelimiting ones have this particular need.

dndx (approved):

dndx:
Merge after plugin tests are fixed.

PR #6891, opened 2021-03-04
perf(warmup) do not warmup upstream names

locao:
When an upstream name is used as a service hosts entry, it should not be warmed-up, as it's not a real DNS entry.

### DIFF
--- a/kong/cache/warmup.lua
+++ b/kong/cache/warmup.lua
@@ -35,8 +35,22 @@ local function warmup_dns(premature, hosts, count)
 
   local start = ngx.now()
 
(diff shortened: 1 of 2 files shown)


gszr:
Can this go into master instead?

dndx (commented):
Is it possible to test this behavior in https://github.com/Kong/kong/blob/master/spec/01-unit/01-db/08-cache_warmup_spec.lua?

locao:
> Is it possible to test this behavior in https://github.com/Kong/kong/blob/master/spec/01-unit/01-db/08-cache_warmup_spec.lua?

Done!

dndx (approved):

PR #6912, opened 2021-03-11
fix(handler): Escape upstream_uri after a plugin

anleg:
### Summary

Since the normalize feature on Kong (https://github.com/Kong/kong/pull/6821), we have strange behavior on uri that are decoded with special chars after the request-transformer plugin.

Here an example:

- Inital request: `store+banne+%C3%A9lectrique.json`
- After a transform_uri: `store+banne+électrique.json`

The goal of this merge request is to fix this issue and have the expected behavior:

- Initial request: `store+banne+%C3%A9lectrique.json`
- After a transform_uri: `store+banne+%C3%A9lectrique.json`

### Full changelog

* Add `escape` function on upstream_uri after plugin

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1300,11 +1300,13 @@ return {
         -- stripping the empty querystring.
         -- We overcome this behavior with our own logic, to preserve user
         -- desired semantics.
-        local upstream_uri = var.upstream_uri
+        local upstream_uri = escape(var.upstream_uri)
 
         if var.is_args == "?" or sub(var.request_uri, -1) == "?" then
-          var.upstream_uri = upstream_uri .. "?" .. (var.args or "")
+          upstream_uri = upstream_uri .. "?" .. (var.args or "")
         end
+
+        var.upstream_uri = upstream_uri
       end
 
       local balancer_data = ctx.balancer_data


bungle:
I don't think this is a correct solutions as it leads to double encoding, first here (before plugins in access phase):
https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L1244

and then on the added place (after plugin execution on access phase).

bungle:
@anleg, care to share you transformer configuration?

anleg:
Hello @bungle 

Thanks for your answer :smile: 

Good Catch ! I thought to put it on the `request-transformer` plugin at first (https://github.com/Kong/kong-plugin-request-transformer/pull/39), but managing specific version might not be a good solution.

In my case, the route regex is `/api/(?i)(?<platform>\S+)(?-i)/pages/(?<plain_url>\S+)`
Here the transformer configuration:

`
  "config": {
    "remove": {
      "querystring": [],
      "headers": [
        "authorization"
      ],
      "body": []
    },
    "replace": {
      "querystring": [],
      "headers": [],
      "uri": "/pages_20210217/$((function() return uri_captures.platform:upper() end)())/$(uri_captures.plain_url).json",
      "body": []
    },
    "http_method": null,
    "add": {
      "querystring": [],
      "headers": [],
      "body": []
    },
    "append": {
      "querystring": [],
      "headers": [],
      "body": []
    },
    "rename": {
      "querystring": [],
      "headers": [],
      "body": []
    }
  }
`

While debugging, I found out that the `é` character is on the `capture_uri` on the transformer plugin (https://github.com/Kong/kong-plugin-request-transformer/blob/master/kong/plugins/request-transformer/access.lua#L84). 

We might be `escape` on the capure_uri before sending to the plugins :thinking:

bungle:
@anleg, thank you, very good points and pointers, and thanks for mentioning capture_uri too!

dndx:
@bungle maybe change the plugin to use https://github.com/Kong/kong/blob/master/kong/pdk/service/request.lua#L142 instead of setting `ngx_var.upstream_uri` directly makes sense here?

Then we can change `kong.service.request.set_path` to normalize the argument before setting.

bungle:
> @bungle maybe change the plugin to use https://github.com/Kong/kong/blob/master/kong/pdk/service/request.lua#L142 instead of setting `ngx_var.upstream_uri` directly makes sense here?
> 
> Then we can change `kong.service.request.set_path` to normalize the argument before setting.

I thought about that. But we need to check what we did before? I am pretty sure that `set_path` could take `escaped` path and work just fine? And if so, then again we are in situation of double escaping.

dndx:
Hello @anleg @bungle ,

How does these looks:

https://github.com/Kong/kong/pull/6978
https://github.com/Kong/kong-plugin-request-transformer/pull/42

dndx:
Above PRs has been merged.

PR #6930, opened 2021-03-17
feat(file-log) add censoring capabilities

jcvincenti:
To avoid logging sensitive data, such as API keys, an attribute remover
is added. It enables the possibility of removing specific attributes
from the message that will be logged. A censored_fields field is added
to the schema were the path of the field should be specified separated
by periods.

* add an attribute remover to the file-log plugin
* add a censored_fields field to the file-log plugin schema
* add an unit test for the attribute remover in the file-log test suite

### DIFF
--- a/kong/plugins/file-log/attribute_remover.lua
+++ b/kong/plugins/file-log/attribute_remover.lua
@@ -1,0 +1,45 @@
+local attribute_remover = {}
+
+local function deep_copy(obj, seen)
+  if type(obj) ~= 'table' then return obj end
+  if seen and seen[obj] then return seen[obj] end
+
+  local s = seen or {}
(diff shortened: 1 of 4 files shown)


javierguerragiraldez:
Thanks for your submission,  can you make it work on all logging methods and not only on the file-log plugin?

fffonion:
Maybe even add it to the [kong.log](https://github.com/Kong/kong/blob/master/kong/pdk/log.lua) PDK functions to avoid deep copy stuff?

gszr:
Hi @jcvincenti,

Thanks for this! Could you take a look at the new `custom_fields_by_lua` configuration in logging plugins, and if/how it can be used for this purpose?

PR #6932, opened 2021-03-18
feat(clustering) relax version compatibility check to allow DP lagging 2 minor versions behind

fffonion:
### Summary

Hybrid mode version check is now changed into a relaxed form:

- DP is allowed to connect to CP that has at most two minor versions ahead
- Version check in plugins is same as Kong itself
- DP is allowed to have superset list of plugins to CP
- DP is not allowed to connect if major version mismatch, or it's newer than CP, or it misses plugins from CP.

### Full changelog

* Implement relaxed version compatibility check in hybrid mode allowing DP to lag two minor versions behind

### Issues resolved

Fix #XXX

### DIFF
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -398,7 +398,8 @@ do
 end
 
 
-local MAJOR_MINOR_PATTERN = "^(%d+%.%d+)%.%d+"
+local MAJOR_MINOR_PATTERN = "^(%d+)%.(%d+)%.%d+"
+
 local function should_send_config_update(node_version, node_plugins)
   if not node_version or not node_plugins then
     return false, "your DP did not provide version information to the CP, " ..
@@ -408,15 +409,21 @@ local function should_send_config_update(node_version, node_plugins)
                   "automatically once this DP also upgrades to 2.3 or later"
   end
 
-  local minor_cp = KONG_VERSION:match(MAJOR_MINOR_PATTERN)
-  local minor_node = node_version:match(MAJOR_MINOR_PATTERN)
-  if minor_cp ~= minor_node then
-    return false, "version mismatches, CP version: " .. minor_cp ..
-                  " DP version: " .. minor_node,
+  local major_cp, minor_cp = KONG_VERSION:match(MAJOR_MINOR_PATTERN)
+  local major_node, minor_node = node_version:match(MAJOR_MINOR_PATTERN)
+  minor_cp = tonumber(minor_cp)
+  minor_node = tonumber(minor_node)
+
+  if major_cp ~= major_node or minor_cp - 2 > minor_node or minor_cp < minor_node then
+    return false, "version incompatible, CP version: " .. KONG_VERSION ..
+                  " DP version: " .. node_version ..
+                  " DP versions acceptable are " ..
+                  major_cp .. "." .. math.max(0, minor_cp - 2) .. " to " ..
+                  major_cp .. "." .. minor_cp .. "(edges included)",
                   CLUSTERING_SYNC_STATUS.KONG_VERSION_INCOMPATIBLE
   end
 
-  -- XXX EE: allow DP to have a superset of CP's plugins
+  -- allow DP to have a superset of CP's plugins
   local p, np
   local i, j = #PLUGINS_LIST, #node_plugins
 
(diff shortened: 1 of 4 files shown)


fffonion:
cc @Kong/team-k8s

kong/clustering.lua
```diff
@@ -333,7 +333,11 @@ local function validate_shared_cert()
 end
 
 
-local MAJOR_MINOR_PATTERN = "^(%d+%.%d+)%.%d+"
+local MAJOR_MINOR_PATTERN = "^(%d+)%.(%d+)%.%d+"
+-- major/minor check that ignores anything after the second digit
+local MAJOR_MINOR_PATTERN_PLUGIN = "^(%d+)%.(%d+)"
+local INF = 1/0
```
dndx, line 339: We can use `math.huge` here instead:

```
LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/
JIT: ON SSE2 SSE3 SSE4.1 BMI2 fold cse dce fwd dse narrow loop abc sink fuse
> print(math.huge == 1/0)
true
```

kong/clustering.lua
```diff
@@ -343,27 +347,63 @@ local function should_send_config_update(node_version, node_plugins)
...
 
-  local minor_cp = KONG_VERSION:match(MAJOR_MINOR_PATTERN)
-  local minor_node = node_version:match(MAJOR_MINOR_PATTERN)
-  if minor_cp ~= minor_node then
-    return false, "version mismatches, CP version: " .. minor_cp ..
-                  " DP version: " .. minor_node,
+  local major_cp, minor_cp = KONG_VERSION:match(MAJOR_MINOR_PATTERN)
+  local major_node, minor_node = node_version:match(MAJOR_MINOR_PATTERN)
+  minor_cp = minor_cp and tonumber(minor_cp) or INF
+  minor_node = minor_node and tonumber(minor_node) or -INF
```
dndx, line 353: Are those conditions really necessary? We have never released a Kong core without minor version since the introduction of Hybrid mode.

kong/clustering.lua
```diff
@@ -343,27 +347,63 @@ local function should_send_config_update(node_version, node_plugins)
...
-                    "CP and DP, CP has version " .. tostring(p.version) ..
-                    " while DP has version " .. tostring(np.version),
+    -- ignore plugins without a version (route-by-header is deprecated)
+    if p.version and np.version then
+      local major_p, minor_p = p.version:match(MAJOR_MINOR_PATTERN_PLUGIN)
+      local major_np, minor_np = np.version:match(MAJOR_MINOR_PATTERN_PLUGIN)
+      minor_p = minor_p and tonumber(minor_p) or INF
+      minor_np = minor_np and tonumber(minor_np) or -INF
+
+      if major_p ~= major_np or minor_p - 2 > minor_np or minor_p < minor_np then
```
dndx, line 388: I may forgot to think about the case of plugins in the discussion. Does supporting two minor version ahead for plugins actually makes sense? Since plugins are usually poorly versioned anyway and generally has a slower release pace than core, two minor versions behind may be a lot of changes.

Ping @PaulFischer-Kong as well.

fffonion: This is now updated to only compare plugins's minor version (same as EE currently does)

kong/clustering.lua
```diff
@@ -408,27 +409,61 @@ local function should_send_config_update(node_version, node_plugins)
...
   end
 
-  for i, p in ipairs(PLUGINS_LIST) do
-    local np = node_plugins[i]
+  -- allow DP to have a superset of CP's plugins
+  local p, np
+  local i, j = #PLUGINS_LIST, #node_plugins
+
+  if j < i then
+    return false, "CP and DP does not have same set of plugins installed",
```
gszr, line 431: Given the CP and DP nodes are allowed to not have the _same_ set of plugins (DP can have a superset), is this message accurate? What about `CP has more installed plugins than DP` or something along the lines of`CP and DP have incompatible sets of plugins (DP is allowed to have a superset)`?

dndx (approved):

kong/clustering.lua
```diff
@@ -408,27 +409,61 @@ local function should_send_config_update(node_version, node_plugins)
...
+      end
     end
+
+    i = i - 1
+    ::continue::
+    j = j - 1
+  end
+
+  if i > 0 then
+    return false, "CP and DP does not have same set of plugins installed",
```
gszr, line 465: Ditto.

gszr (approved):

gszr:
Merging the PR; we can make suggested changes (if agreed) in a separate PR. Thanks for the hard work @fffonion and @eskerda!

PR #6938, opened 2021-03-19
feat(clustering) resilient initial config on control plane database failures

### LINKED ISSUE #6935: Control Plane should utilize cache when it looses database connection
### Summary

When the control plane looses it's database connection it's currently unable to send updates to new data planes. The data planes continue to operate with the data they have available indefinetely but we could improve our command and control SLA by having the control plane utilize it's local cache

### Steps To Reproduce

1. Run Kong hybrid mode with Postgres
2. Kill the Database
3. Spin up a new data plane

### Additional Details & Logs

```
kong-cp_1  | 2021/03/18 15:06:12 [error] 97#0: *1484 [lua] clustering.lua:768: unable to update clustering data plane status: [postgres] host is unreachable, client: 172.18.0.3, server: kong_cluster_listener, request: "GET /v1/outlet?node_id=a8238d96-c439-47b7-82ce-20b9acd1106f&node_hostname=kong-dp&node_version=2.3.2.0-nightly-enterprise-edition HTTP/1.1", host: "kong-cp:8005"
```

bungle:
### Summary

@hutchic reported in #6935 that:

> When the control plane looses it's database connection it's currently unable to send updates to new data planes.

This PR is about to make control plane more resilient on database failures on such case.

### Issues Resolved

Fix #6935

### DIFF
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -22,6 +22,7 @@ local type = type
 local math = math
 local pcall = pcall
 local pairs = pairs
+local tostring = tostring
 local ngx = ngx
 local ngx_log = ngx.log
 local ngx_sleep = ngx.sleep
@@ -31,6 +32,7 @@ local kong = kong
 local ngx_exit = ngx.exit
 local exiting = ngx.worker.exiting
 local ngx_time = ngx.time
+local ngx_now = ngx.now
 local ngx_var = ngx.var
 local io_open = io.open
 local table_insert = table.insert
@@ -62,6 +64,9 @@ local clients = setmetatable({}, WEAK_KEY_MT)
 local prefix = ngx.config.prefix()
 local CONFIG_CACHE = prefix .. "/config.cache.json.gz"
 local CLUSTERING_SYNC_STATUS = kong_constants.CLUSTERING_SYNC_STATUS
+local deflated_reconfigure_payload -- Contains a compressed (zlib) payload of the latest database export,
+                                   -- that can be used to send configuration to new clients even in a case
+                                   -- of a database outage.
 local declarative_config
 local next_config
 
@@ -264,6 +269,7 @@ local function communicate(premature, conf)
 
       else
         if typ == "close" then
+          ngx_log(ngx_DEBUG, "received CLOSE frame from control plane")
           return
         end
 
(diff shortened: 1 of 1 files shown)


eskerda:
This PR might fix _part_ of it, but the error that shows up on the reproduction steps happens on this upsert: https://github.com/kong/kong/blob/e245026c44ddb653a02d87ccb82715a37388f6c1/kong/clustering.lua#L604-L613

Since the db is not available, the upsert adding the new node fails.

bungle:
We could perhaps store the payload to `shm` and make it so that only a single worker loads it from db.

kong/clustering.lua
```diff
@@ -62,6 +62,7 @@ local clients = setmetatable({}, WEAK_KEY_MT)
 local prefix = ngx.config.prefix()
 local CONFIG_CACHE = prefix .. "/config.cache.json.gz"
 local CLUSTERING_SYNC_STATUS = kong_constants.CLUSTERING_SYNC_STATUS
+local deflated_reconfigure_payload
```
fffonion, line 65: let's add a comment on this variable

fffonion (commented):
will it be possible to add tests around this?

dndx:
I took a look at the PR, and it looks good to me. I think we can leave storing the data into `shdict` or even filesystem for later because shdict are not very suited for storing large binary data like those. As of tests, this may be relatively difficult to test, and I am okay with leaving it out is that's the case since the change has been verified by multiple parties already.

kong/clustering.lua
```diff
@@ -289,6 +301,9 @@ local function communicate(premature, conf)
 
         elseif typ == "pong" then
           ngx_log(ngx_DEBUG, "received PONG frame from control plane")
+
+        else
+          ngx_log(ngx_DEBUG, "received UNKNOWN frame from control plane")
```
dndx, line 306: The type of frame can be printed here. Maybe:
```suggestion
          ngx_log(ngx_DEBUG, "received " .. tostring(typ) .. " frame from control plane")
```

dndx (approved):
LGTM, aside from a minor recommendation above.

dndx:
I think we should target this against `next` instead since it is not a required feature for 2.4 and also 2.4 has been feature freezed at this point (we already released beta and RC).

PR #6941, opened 2021-03-21
Feat/pb plugins

javierguerragiraldez:
ProtoBuf-based plugin protocol.

### DIFF
--- a/kong/runloop/plugin_servers/pb_rpc.lua
+++ b/kong/runloop/plugin_servers/pb_rpc.lua
@@ -1,0 +1,426 @@
+local kong_global = require "kong.global"
+local cjson = require "cjson.safe"
+local protoc = require "protoc"
+local pb = require "pb"
+require "lua_pack"
+
+local ngx_ssl = require "ngx.ssl"
+
+local ngx = ngx
+local kong = kong
+
+
+local cjson_encode = cjson.encode
+local t_unpack = table.unpack       -- luacheck: ignore table
+local st_pack = string.pack         -- luacheck: ignore string
+local st_unpack = string.unpack     -- luacheck: ignore string
+
+local Rpc = {}
+Rpc.__index = Rpc
+
+local pp = require "pl.pretty".write
+
+
+
+
+
+local pb_unwrap
+do
+  local structpb_value, structpb_list, structpb_struct
+
+  function structpb_value(v)
+    if v.list_value then
+      return structpb_list(v.list_value)
+    end
+
+    if v.struct_value then
+      return structpb_struct(v.struct_value)
+    end
+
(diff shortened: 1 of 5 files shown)


kong/pdk/private/phases.lua
```diff
@@ -71,7 +71,7 @@ local function check_phase(accepted_phases)
       -- treat custom content blocks as the Admin API
       current_phase = PHASES.admin_api
     else
-      error("no phase in kong.ctx.core.phase")
+      error("no phase " .. current_phase .. "in kong.ctx.core.phase")
```
bungle, line 74: this is inside `if not current_phase`

bungle (dismissed):

kong/pdk/private/phases.lua
javierguerragiraldez: hum, it's the only change in that file; probably an artifact of the rebase.  will remove

kong/runloop/plugin_servers/pb_rpc.lua
```diff
@@ -0,0 +1,426 @@
...
+      ctx_shared[k] = v
+    end,
+
+    ["kong.nginx.req_start_time"] = ngx.req.start_time,
+  }
+
+  local p = protoc.new()
+  --p:loadfile("kong/pluginsocket.proto")
+
+  p:addpath("/usr/include")
```
fffonion, line 214: hardcode this path feels slightly weird, when/who will copy the proto file into this path?

javierguerragiraldez: right.  there's where the "standard" google structs are (on my system)  should we include a copy of that? add to the deploy scripts?

gszr: > right. there's where the "standard" google structs are (on my system) should we include a copy of that? add to the deploy scripts?

I think we have at least two options here:
- Marking them as a package dependency (if a package for it exists in the systems we support)
- Bundling them into our usual `/usr/local/kong/` or `/usr/local/kong/lib` paths

fffonion: Does it make sense to include them in kong's src tree? The import files are not so large in size but i'm not sure if that's a common practice, since unlinke other implementations lua-protobuf doesn't bundle any google proto by default.

javierguerragiraldez: just checked, for ubuntu it's `libprotobuf-dev`

kikito: We have talked about this online and we'll call this a "known limitation" of the pb plugins server, will try to improve it after the Beta.

kikito (approved):

gszr (approved):

PR #6943, opened 2021-03-22
fix(ci) add external repo to LUA_PATH

Tieske:


Tieske:
This is the better fix, though https://github.com/Kong/kong-plugin-aws-lambda/pull/52 was already merged.

Tieske:
can no longer test this fix, since the remote repo was modified

Tieske:
maybe we should just close this one?

PR #6944, opened 2021-03-22
feat(plugins) add log by lua capability

gszr:
Add capability to set new log fields, or unset existing fields, by
executing custom Lua code in the Log phase.

This feature uses the new PDK method `kong.log.set_serialize_value`,
as well as the new sandbox capability, both introduced in Kong 2.3.

Currently supported logging plugins:
* file-log
* Loggly
* Syslog
* tcp-log
* udp-log
* http-log

Not included in:
* Datadog
*  Statsd

### DIFF
--- a/kong/plugins/file-log/handler.lua
+++ b/kong/plugins/file-log/handler.lua
@@ -2,7 +2,7 @@
 local ffi = require "ffi"
 local cjson = require "cjson"
 local system_constants = require "lua_system_constants"
-local sandbox = require "kong.tools.sandbox"
+local sandbox = require "kong.tools.sandbox".sandbox
 
 
 local kong = kong
@@ -74,8 +74,9 @@ local FileLogHandler = {
 
 function FileLogHandler:log(conf)
   if conf.custom_fields_by_lua then
+    local set_serialize_value = kong.log.set_serialize_value
     for key, expression in pairs(conf.custom_fields_by_lua) do
-      kong.log.set_serialize_value(key, sandbox.sandbox(expression, sandbox_opts)())
+      set_serialize_value(key, sandbox(expression, sandbox_opts)())
     end
   end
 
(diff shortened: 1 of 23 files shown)


kikito (approved):

kong/plugins/file-log/handler.lua
```diff
@@ -69,6 +73,12 @@ local FileLogHandler = {
 
 
 function FileLogHandler:log(conf)
+  if conf.custom_fields_by_lua then
+    for key, expression in pairs(conf.custom_fields_by_lua) do
```
dndx, line 77: Slight concern on this usage of `pairs` inside a hot path. It can not be JIT compiled by LuaJIT and may result in slow execution. Any chance `pairs` can be avoided here?

It's not a blocker for the PR, but just making a note here so we remember to look into it.

kong/plugins/file-log/handler.lua
```diff
@@ -69,6 +73,12 @@ local FileLogHandler = {
...
+      kong.log.set_serialize_value(key, sandbox.sandbox(expression, sandbox_opts)())
```
dndx, line 78: `kong.log.set_serialize_value` can be cached inside the `:log` handler to speed up the hot path (it can not be done in the module level because the namespaces log object may not been created yet at the time of loading this file).

`sandbox.sandbox` can be cached at module top level to speed up the hot path.

kong/plugins/file-log/handler.lua
gszr: I will look into this after the beta.

dndx (approved):

kong/plugins/file-log/handler.lua
javierguerragiraldez: the `pairs` loop goes around a `sandbox(expression, ...)`; i don't think it could be considered a tight loop.

tangentially related:  `pairs()` isn't an NYI because it's too complex, but because the interpreted version is faster in most cases.  the downside is that it breaks surrounding traces, which is bad if there's tight code with high interpretation penalty (like FFI).   for "wide" loops it's ok, the result will be a few more chained sidetraces.

Issue #5580, opened 2020-02-18
bug: upstream host_header is not respected when preserving host


hbagdi:
### Summary

When `preserve_host` in the Route entity is set to `true`, `host_header` in Upstream entity is not respected.

### Steps To Reproduce

Configure Kong with:

```
_format_version: "1.1"
services:
- name: svc1
  host: my-backend.org
  routes:
  - name: r1
    hosts:
    - bar.com
    strip_path: true
    preserve_host: true
    paths:
    - /r1
upstreams:
- name: my-backend.org
  host_header: foo.com
  algorithm: round-robin
  targets:
  - target: "66.39.74.7:80"
```
And then fire:
```
http :8000/r1/echo host:bar.com                                                                                                                                                                             <<<
HTTP/1.1 200 OK
Accept-Ranges: none
Connection: keep-alive
Content-Encoding: gzip
Content-Length: 483
Content-Type: text/plain;charset=UTF-8
Date: Tue, 18 Feb 2020 21:49:11 GMT
Server: Apache
Vary: Accept-Encoding,User-Agent
Via: kong/2.0.0
X-Kong-Proxy-Latency: 1
X-Kong-Upstream-Latency: 164

Simple webservice echo test: make a request to this endpoint to return the HTTP request parameters and headers. Results available in plain text, JSON, or XML formats. See http://www.cantoni.org/2012/01/08/simple-webservice-echo-test for more details, or https://github.com/bcantoni/echotest for source code.

Array
(
    [method] => GET
    [headers] => Array
        (
            [Accept] => */*
            [Accept-Encoding] => gzip, deflate
            [User-Agent] => HTTPie/0.9.8
            [X-Real-Ip] => 127.0.0.1
            [X-Forwarded-Port] => 8000
            [X-Forwarded-Host] => bar.com
            [X-Forwarded-Proto] => http
            [X-Forwarded-For] => 127.0.0.1
            [Connection] => close
            [Host] => bar.com
        )

    [request] => Array
        (
        )

    [client_ip] => 127.0.0.1
    [time_utc] => 2020-02-18T21:49:11+0000
    [info] => Echo service from Scooterlabs (http://www.scooterlabs.com)
)
```

As seen in the response from upstream, the upstream sees the host as `bar.com` and not `foo.com`.

### Additional Details & Logs

- Kong version (`$ kong version`) 2.0 (won't work in previous versions either)
- Kong debug-level startup logs (`$ kong start --vv`)
[...]

PR #6959, opened 2021-03-24
Release: 2.4.0

locao:


javierguerragiraldez (approved):

CHANGELOG.md
```diff
@@ -56,6 +57,131 @@
...
+
+- Changed default values and validation rules for plugins that were not
+  well-adjusted for dbless or hybrid modes.
+  [6885](https://github.com/Kong/kong/pull/6885)
+- Topological sort now prioritizes core, avoiding problems when plugin entities
+  use core entities but don't explicitly depend on them.
+  [6880](https://github.com/Kong/kong/pull/6880)
+- If needed, `Host` header is now updated between balancer retries, using the
+  value configured in the correct upstream entity.
+  [6796](https://github.com/Kong/kong/pull/6796)
```
murillopaula, line 146:
```suggestion
- If needed, `Host` header is now updated between balancer retries.
  [6796](https://github.com/Kong/kong/pull/6796)
```

I would change this changelog entry simply to:

"If needed, `Host` header is now updated between balancer retries."

So I'd remove the "using the value configured in the correct upstream entity." because I don't think it uses the `host_header` attribute from upstream entities, if that's what this part means.

I suggest this because there are inter relationships between host headers that are still in discussion regarding what's the precedence and how they work together. See https://github.com/Kong/kong/issues/5580

Please let me know what you think

PR #6969, opened 2021-03-25
feat(external-plugins) Protobuf include dirs as config setting

javierguerragiraldez:
Allows the user to specify where to find ProtoBuf definition files.
Currently the plugins protocol requires `google/protobuf/empty.proto`
and `google/protobuf/struct.proto`, which the `libprotobuf-dev` (at
least in Ubuntu) install in `/usr/include`.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -137,6 +137,9 @@
                                                             # dump info of all plugins it
                                                             # manages
 
+#protobuf_includes = /usr/include   # Comma-separated list of directories to search for
+                                    # Protocol Buffers definition files.
+
 #port_maps =                     # With this configuration parameter, you can
                                  # let the Kong to know about the port from
                                  # which the packets are forwarded to it. This
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -625,6 +625,7 @@ local CONF_INFERENCES = {
 
   kic = { typ = "boolean" },
   pluginserver_names = { typ = "array" },
+  protobuf_includes = { typ = "array" },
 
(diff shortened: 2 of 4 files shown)


kong.conf.default
```diff
@@ -137,6 +137,9 @@
                                                             # dump info of all plugins it
                                                             # manages
 
+#protobuf_includes = /usr/include   # Comma-separated list of directories to search for
```
fffonion, line 140: unrelated but does it make sense to reuse this filed in grpc plugins to resolve issues like https://github.com/Kong/kong-plugin-grpc-gateway/issues/15?

fffonion (approved):

kong.conf.default
javierguerragiraldez: unfortunately, that's not enough to make `import` work transparently in procfiles.  even with the `p.include_imports = true` line, lproc only parses those imported files, they're not loaded into the codec context.  That's why a few lines down I added explicit `p:loadfile()` lines for empty.proto and struct.proto.   I think `import public ...` should be enough, but seems something is missing.

kikito (approved):

dndx (approved):

dndx:
Just curious and double checking, are we supposed to merge `feat` into the release branch after beta?

javierguerragiraldez:
when reviewing just before the beta merge, we agreed that it was a very minor thing and shouldn't block the beta.  maybe "feat/" isn't appropriate?   i put it because there's a new config setting

gszr:
Closing as we decided against this approach.

PR #6972, opened 2021-03-26
fix(handler) save failed try status to log it

### LINKED ISSUE #6820: If final balancer retry fails, that try does not have the failure data recorded
### Summary

This code records the failure details for the **previous** try:
https://github.com/Kong/kong/blob/2.3.1/kong/init.lua#L926-L928

As such, if the final retry is a failure, those details don't get recorded.

### Steps To Reproduce

1. Add service and route objects with artificial latency to force retries, e.g.
```yaml
routes:
- name: httpbin
  paths:
  - /
  preserve_host: true
  service: httpbin
  strip_path: false

services:
- name: httpbin
  host: httpbin
  protocol: http
  port: 80
  retries: 3
  connect_timeout: 1000
  read_timeout: 1000
  write_timeout: 1000
```

2. Make a request that will timeout, e.g. `/delay/2`

3. You'll see something like this in the logs:
```json
    "tries": [
        {
            "balancer_latency": 1,
            "port": 80,
            "ip": "172.20.0.4",
            "code": 504,
            "balancer_start": 1612821178453,
            "state": "failed"
        },
        {
            "balancer_latency": 0,
            "port": 80,
            "ip": "172.20.0.4",
            "code": 504,
            "balancer_start": 1612821179456,
            "state": "failed"
        },
        {
            "balancer_latency": 0,
            "port": 80,
            "ip": "172.20.0.4",
            "code": 504,
            "balancer_start": 1612821180457,
            "state": "failed"
        },
        {
            "balancer_latency": 0,
            "port": 80,
            "balancer_start": 1612821181457,
            "ip": "172.20.0.4"
        }
    ],
```

### Additional Details & Logs

- Kong version (2.3.1)

locao:
### Summary

The balancer tries states are saved when retrying. The last retry is never saved, as there are no more retries. This change saves the failure state so it can be correctly logged.

### Issues resolved

Fix #6820

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1427,6 +1427,19 @@ return {
         end
       end
 
+      -- if this is the last try and it failed, save its state to correctly log it
+      local status = ngx.status
+      if status > 499 and ctx.balancer_data then
+        local balancer_data = ctx.balancer_data
+        local try_count = balancer_data.try_count
+        local retries = balancer_data.retries
+        if try_count > retries then
+          local current_try = balancer_data.tries[try_count]
+          current_try.state = "failed"
+          current_try.code = status
+        end
+      end
+
       local hash_cookie = ctx.balancer_data.hash_cookie
       if not hash_cookie then
         return
--- a/spec/02-integration/05-proxy/07-upstream_timeouts_spec.lua
+++ b/spec/02-integration/05-proxy/07-upstream_timeouts_spec.lua
@@ -35,7 +35,6 @@ for _, strategy in helpers.each_strategy() do
         if not route.protocols then
           route.protocols = { "http" }
         end
-
         bp.routes:insert(route)
       end
 
@@ -46,7 +45,8 @@ for _, strategy in helpers.each_strategy() do
       bp = helpers.get_db_utils(strategy, {
         "routes",
         "services",
-      })
+        "plugins",
+      }, { "ctx-checker-last" })
 
       insert_routes {
         {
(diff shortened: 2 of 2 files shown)


dndx:
Looks straightforward. Quick question, can this regression be tested?

kikito:
Agreed that ideally this would need some sort of test.

kikito (approved):
Approved on the condition that tests pass before merging this (right now CI is red for an unrelated issue)

kong/runloop/handler.lua
```diff
@@ -1427,6 +1427,19 @@ return {
         end
       end
 
+      -- if this is the last try and it failed, save its state to correctly log it
+      local status = ngx.status
+      if status > 499 and ctx.balancer_data then
```
Tieske, line 1432: Is this correct? any 500 now classifies as a failure.

The "tries" are bound to the semantics of the nginx retry mechanism and settings. Since this runs in the `header_filter` phase it will apply to every 500 (eg. a plugin that fails in the `access` phase, that would still cause the `header_filter` phase to be executed). It would record the failure, but it has completely different semantics, since effectively there never was a TCP connect attempt to the upstream since we errorred out before.

The reason the last try wasn't recorded is because the `balancer` phase is not called again. The only way to do this proper would be to replicate the retry mechanism in Kong, and no longer rely on Nginx to do that.

Tieske: (sorry for the late comment, the PR close message in Slack caught my eye)

PR #6992, opened 2021-04-02
fix(http-log) set headers with single value, not array

Tieske:
The `headers` typedef used, specified for the values an array
(originally from the route-entity), but for this plugin having
multiple values is really confusing. This reduces it back to a
single string per header.


kong/plugins/http-log/migrations/001_230_to_240.lua
```diff
@@ -0,0 +1,36 @@
+local operations = require "kong.db.migrations.operations.200_to_210"
```
Tieske, line 1: I reused this migrations file, is that ok, or should I create a new one with just the helper I need?

Tieske: @bungle is this ok, or do I need to duplicate it to a new version?

gszr:
Hi @Tieske. I have rebased your PR to target the `master` branch (our default for all PRs onwards).

fffonion (commented):
I think this breaks backward compatibility in hybrid mode.
Is it okay to keep the same schema, but having a validator to only allow user to enter one value? If not, can we make the "single-value" headers field a new field instead of changing
the existing one?

Tieske:
@fffonion what are you proposing? like this:

- in 3.0 create a new field
- copy the contents of the old one in the new one using a migration (leaving the old one)
- this ensures 3.0 CP can send configs to 3.0 as well as 2.x DP
- in 3.1 remove the old field (ensure compat with 3.0 and 3.1, so again upgrade path)

That would work imo, but in 3.1 removing; do we consider that a breaking change?

fffonion:
@Tieske if we make this in 3.0, then i think breaking change is allowed, we can keep this as is, but as Rob mentioned add to the breaking change list.

kong/plugins/http-log/migrations/001_230_to_240.lua
kikito: We are supposed to create a new operations and copy-paste if we need to. The only reason to do that (as far as I know) is to minimize the risk from us modifying an existing function that was used on a previous migration by mistake (and not noticing until later ... when upgrading from an older version fails).

I guess if you are **very careful** and only use existing functions without adding new ones it's more or less ok... except for the fact that the module name is now completely **lying** - it says `200_to_210` but it is being used in a different migration. I find that ... irritating to be honest. Please copy the function to a different file just to make me happy.

kong/plugins/http-log/migrations/001_230_to_240.lua
```diff
@@ -0,0 +1,36 @@
...
+  return function(connector)
+    return ops:fixup_plugin_config(connector, "http-log", function(config)
+      local updated = false
+      if type(config) == "table" then -- not required, but let's be defensive here
+        local headers = config.headers
+        if type(headers) == "table" then
+          for header_name, value_array in pairs(headers) do
+            if type(value_array) == "table" then
+              -- only update if it's still a table, so it is reentrant
+              headers[header_name] = value_array[1] or "empty header value"
```
kikito, line 14: Shouldn't this value be an actual empty string?

```suggestion
              headers[header_name] = value_array[1] or ""
```

spec/03-plugins/03-http-log/02-schema_spec.lua
```diff
@@ -38,7 +38,8 @@ describe(PLUGIN_NAME .. ": (schema)", function()
     local ok, err = validate({
         http_endpoint = "http://myservice.com/path",
         headers = {
-          ["X-My-Header"] = { "123" }
+          ["X-My-Header"] = "123",
+          ["X-Your-Header"] = "abc",
```
kikito, line 42: I think we should also test an empty header value on these tests.

kikito (changes requested):
I think the migrations need to be moved to 200_to_300

Also, the breaking change needs to be written in the changelog.

kong/plugins/http-log/migrations/001_230_to_240.lua
Tieske: no, in the old implementation one could set an empty array. In that case we now transform it to "empty header value" as a clear(?) indicator there was nothing there. Functionally the user should remove the header all together in that case. The alternative would be to simply drop the entry, but that would be too magical imo.

Tieske: btw empty headers are not valid

spec/03-plugins/03-http-log/02-schema_spec.lua
Tieske: empty headers are not valid

Tieske:
> I think the migrations need to be moved to 200_to_300

I named them `280_to_300` assuming that's the better name?

I also copied the migration-operation file, but left unused code commented out. Should I remove that code, or leave it in case another migration needs it?

Tieske:
@kikito removed the dead code as well. Please have another look.

PR #6993, opened 2021-04-02
fix(migrations) don't assume '/?/init.lua' in the path

### LINKED ISSUE #6867: custom migrations expect `./?/init.lua` in the path setting
If you create a custom plugin in a non standard location and set `lua_package_path = /usr/local/custom/?.lua`, then migrations fail. Since the core file is `init.lua`, and is required as `kong.plugins.<name>.migrations`.

So the migrations expect the `init.lua` to be part of the path; `lua_package_path = /usr/local/custom/?.lua;/usr/local/custom/?/init.lua`.

The `init.lua` is a common Lua idiom, but should not be required for custom plugins.

Tieske:
When a custom plugin is loaded that is not located in the Kong tree
the user manually sets the Lua-path to locate the plugin. In that
case the assumption that the 'init' files will be found might be
wrong.

Fixes #6867
Fixes #7340

### DIFF
--- a/kong/db/migrations/state.lua
+++ b/kong/db/migrations/state.lua
@@ -67,7 +67,9 @@ local function load_subsystems(db, plugin_names)
       for _, plugin_name in ipairs(sorted_plugin_names) do
         local namespace = ss.namespace:gsub("%*", plugin_name)
 
-        local ok, mig_idx = utils.load_module_if_exists(namespace)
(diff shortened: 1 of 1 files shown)


kikito (approved):

Tieske:
Thx @kikito for the approval, please note that I cannot merge PR's (no rights)

PR #6997, opened 2021-04-05
refactor(*) go pb rpc

gszr:
- Add missing lua-protobuf to rockspec
- Refactor `parsefile`, so it looks in common development and production locations


javierguerragiraldez (approved):
:+1:

PR #7005, opened 2021-04-06
fix(postgres) support connecting to Postgres >= 12 by disabling SSL/TLS < 1.2

### LINKED ISSUE #6645: Can not connect to PostgreSQL DB requiring TLS > 1.0
### Summary

If you try to setup Kong with an external postgres DB which enforces a TLS version > 1.0 you will get an error during kong-init-migrations container startup. 
```
[PostgreSQL error] failed to retrieve PostgreSQL server_version_num: FATAL: TLS version used does not meet minimal requirements for this server. Please use a higher TLS version and retry.
```
If you set TLS 1.0 on the DB connection works. This is a problem if you want to use a DB provided by a cloud provider.
Official documentation does not state a way to select TLS version for postgres connection.

### Steps To Reproduce

Setup Kong (on Kubernetes) with external postgres db.

### Additional Details & Logs

kong-1.12.0 on Kubernetes

dndx:
protocols < 1.2 in cosocket and bumping the pgmoon dependency to the
latest which includes the LuaSec TLS 1.2 support. fixes #6645

Postgres >= 12 requires TLS >= 1.2 to connect. Since TLS < 1.2 has known
vulnerabilities, we will disable older versions of support in OpenResty
cosocket by default, but leave the option for users to re-enable them
if so desire.

Tested locally with latest Postgres 13 with TLS enforced and both migrations as well as proxy features works well after the change.

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1360,6 +1360,17 @@
                                  #
                                  # See https://github.com/openresty/lua-nginx-module#lua_ssl_verify_depth
 
+#lua_ssl_protocols = TLSv1.1 TLSv1.2 TLSv1.3   # Defines the TLS versions supported
+                                               # when handshaking with OpenResty's
+                                               # TCP cosocket APIs.
+                                               #
+                                               # This affects connections made by Lua
+                                               # code, such as connections to the
+                                               # database Kong uses, or when sending logs
+                                               # using a logging plugin. It does *not*
+                                               # affect connections made to the upstream
+                                               # Service or from downstream clients.
+
 #lua_package_path = ./?.lua;./?/init.lua;  # Sets the Lua module search path
                                            # (LUA_PATH). Useful when developing
                                            # or using custom plugins not stored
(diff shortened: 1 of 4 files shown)


bungle (approved):

murillopaula:
I thought this was already fixed by https://github.com/leafo/pgmoon/pull/103?

murillopaula:
@jeremyjpj0916 can you confirm?

bungle:
Yes, in certain phases we do use `LuaSec`. So we kinda need both fixes.

murillopaula:
ah I see @dndx

In this case, it seems like pgmoon luasec sets the default minimum TLS version to v1.1. Do you think this will cause issues considering the OpenResty cosocket minimum default will be v1.2 from this PR onwards?

dndx:
@murillopaula Practically, no. I tested using older versions of Postgres such as Postgres 9 which is 6 years old and EOL, and it has no problem supporting TLS 1.2. (which was released in 2008).

But in an effort to stay consistent with Pgmoon for the time being, I changed the default to allow TLS 1.1 as well. We may visit this decision later as TLS 1.1 does have some known vulnerabilities.

murillopaula (approved):

PR #7007, opened 2021-04-06
chore(scripts) add update_copyright release action

kikito:


gszr (commented):
Could we make that part of the make-prerelease-release script as well (there's a common file that is included for both)?

kikito:
@gszr done!

I have also removed the hardcoded MIT license on the grpc-web plugin, since it is no longer needed (the plugin repo has a license file now).

I have tested both make-patch-release and make-prerelease and both scripts work fine (I would prefer if both took parameters the same way though - prerelease uses `-v` and `-s` but patch doesn't)

gszr:
> @gszr done!
> 
> I have also removed the hardcoded MIT license on the grpc-web plugin, since it is no longer needed (the plugin repo has a license file now).

Great, thanks!

> 
> I have tested both make-patch-release and make-prerelease and both scripts work fine (I would prefer if both took parameters the same way though - prerelease uses `-v` and `-s` but patch doesn't)

Agreed. I have a local branch where I started doing that - plus a few other things, like moving more common code into common.sh.

gszr (approved):

PR #7014, opened 2021-04-08
fix(conf) fix changing response headers

### LINKED ISSUE #6995: Underscore in kong.response headers
### Summary

It looks like kong transform underscore into dash while setting a new response header

### Steps To Reproduce
```lua
function Test:header_filter(conf)
  kong.response.set_header('service_name', 'test')
end
```

results with `service-name`

### Additional Details & Logs

- Kong version - 2.3.2

Ocean-Moist:
Underscores in the response headers change to hyphens by default, this config change makes the default behavior preserving hyphens.   

Fix #6995

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1452,3 +1452,9 @@
                                      # **Warning**: Certain variables, when made
                                      # available, may create opportunities to
                                      # escape the sandbox.
+
+#lua_transform_underscores_in_response_headers = off # Does not transform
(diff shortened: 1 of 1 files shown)


kikito:
Hello @Ocean-Moist , thanks a lot for taking the time to send this PR. We have decided not to merge it since it could break plugins which currently rely on this behavior. You may find the palliative fixes we decided to implement in #6995

PR #7016, opened 2021-04-08
fix(pdk) do not replace underscores by slashes in kong.response.set_header

kikito:
This PR sets the `lua_transform_underscores_in_response_headers` nginx config variable to `off` so that `kong.response.set_header` does not automatically transform underscores into slashes on response headers.

Solves #6995


kikito:
Closing this up since setting this isn't fully backwards-compatible (plugins might and do rely on the automatic conversion of underscores). See discussion in #6995 for details.

PR #7018, opened 2021-04-09
docs(pdk) document underscore transformations in kong.response.set_headers

kikito:
Related with #6995

### DIFF
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -384,6 +384,18 @@ local function new(self, major_version)
   --
   -- This function should be used in the `header_filter` phase, as Kong is
   -- preparing headers to be sent back to the client.
+  --
+  -- Note: Underscores in Header names are automatically transformed into dashes
+  -- by default. If you want to deactivate this behavior you should set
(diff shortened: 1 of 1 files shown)


kong/pdk/response.lua
```diff
@@ -384,6 +384,15 @@ local function new(self, major_version)
...
   -- This function should be used in the `header_filter` phase, as Kong is
   -- preparing headers to be sent back to the client.
+  --
+  -- Note: Underscores in Header names are automatically transformed into dashes
+  -- by default. If you want to deactivate this behavior you should set
+  -- the `lua_transform_underscores_in_response_headers` nginx config option to `off`
+  --
+  -- An easy way to do this is in the Kong Config file:
+  --
+  --     lua_transform_underscores_in_response_headers = off
```
bungle, line 394: I don't think we have that, but we have injection:

```
nginx_http_lua_transform_underscores_in_response_headers = off
```

kikito: Thank you, fixed!

dndx (approved):

kikito:
The tests are red because of a Luarocks temp failure, but this is a docs-only change so I am going to ignore that and merge the change.

PR #7030, opened 2021-04-13
perf(*) implement the performance testing framework

fffonion:
### Summary

Adds a performance testing tool. Three "drivers" are implemented
depending on use cases:
- `local` driver: useful to compare between git commits, branches and releases, will run
Kong, postgres and upstream Nginx directly on the host machine. Uncommited changes
will be stashed and recovered after test run.
- `docker` driver: useful to compare between docker image releases, will run all components
in docker containers. This driver doesn't support generating flamegraphs.
- `terraform` driver: useful to get meaningful RPS number on given hardware, currently 
it supports the `"equinix-metal"` provier only. It's possible to extend to other hosting
providers.

The testing tools can reuse the `spec.helpers` functions to setup Kong entities. It will then
runs `wrk` to measure latency and RPS with given configuration.

In this PR, two "purposes" has been added, `spec/04-perf/01-rps` is used for comparing
request per seconds, and `spec/04-perf/02-framegraph` is used for generating flamegraph.

To select a driver, pass it in the `PERF_TEST_DRIVER` env var:
```
PERF_TEST_DRIVER=terraform bin/busted -o gtest spec/04-perf/01-rps/01-baseline_spec.lua -t single_route
```

This PR also enables following triggers to run perf tests in github actions:
-  PR has title `perf(xxx)`
-  PR has received a comment `/perf` for performance tests and `/flamegraph` for flamegraph. User can select tags after test suite, like `/perf baseline,key_auth`.
- Daily cron running upon master branch.

### Full changelog

* Add a performance testing tool.


gszr (approved):
LGTM!

.github/workflows/perf.yml
```diff
@@ -0,0 +1,183 @@
+name: Performance Test
+
+on:
+  pull_request:
+  issue_comment:
+  schedule:
+  # don't know the timezone but it's daily at least
```
dndx, line 7: I think this is in UTC :)

dndx (approved):

PR #7037, opened 2021-04-14
feat(hmac-auth) add support for the "@request-target" field support

### LINKED ISSUE #3789: kong hmac-auth broken, with layer7 proxy
### Summary
https://docs.konghq.com/hub/kong-inc/hmac-auth/
defined signing_string contains http version.
```
signing_string="date: Thu, 22 Jun 2017 17:15:21 GMT\nGET /requests HTTP/1.1"
```

### Steps To Reproduce

1. add nginx loadbalance before kong,  which support http2.0
2. openresty+kong run without https
3. client send request via http2.0
4. hmac-auth failed, nginx proxy_pass with http version http1.0 or http1.1, not match http2.0

client ----http2---> nginx ----http1.1---> openresty+kong

dndx:
which is introduced in later versions of the "Signing HTTP Messages"
RFC. closes #3789

Previously the Kong implementation of the RFC included a signing
parameter "request-line" which contains the HTTP request method and
request URI plus the HTTP version number. This is not part of the RFC
draft, and has created issue such as #3789 where changing the HTTP
request version by intermediary proxy caused the signature validation to
fail.

In general, including the HTTP protocol version inside the signature
makes little sense, as the protocol does not identify the resource that
is being accessed, and can be changed by clients/proxies which still
preserving the original semantics of the request.

This PR adopts the "@request-target" field as described inside
https://tools.ietf.org/html/draft-ietf-httpbis-message-signatures-03
which only included the lowercased request method and request URI when
calculating the hash.

Note: we intentionally steered away from the
https://tools.ietf.org/html/draft-cavage-http-signatures-12 RFC
recommendation which uses the "request-target" field instead. This RFC
draft has expired due to long inactivity and the name also has slight
chance of conflict with HTTP headers named "Request-Target". Overall
the Kong hmac-auth plugin did not follow the RFC draft in the beginning,
and this change should not cause any compatibility issues for the users
that are relying on "request-line" today.

### DIFF
--- a/spec/03-plugins/19-hmac-auth/03-access_spec.lua
+++ b/spec/03-plugins/19-hmac-auth/03-access_spec.lua
@@ -420,6 +420,24 @@ for _, strategy in helpers.each_strategy() do
         assert.same({ reply = "hello noname" }, cjson.decode(res))
       end)
 
+      it("accepts authorized gRPC calls with @request-target (HTTP/2 test), bug #3789", function()
+        local date = os.date("!%a, %d %b %Y %H:%M:%S GMT")
+        local encodedSignature   = ngx.encode_base64(hmac_sha1_binary("secret", "date: " .. date ..
+                                                                      "\n@request-target: " ..
+                                                                      "post /hello.HelloService/SayHello"))
+        local hmacAuth = [[hmac username="bob",algorithm="hmac-sha1",]]
+          .. [[headers="date @request-target",signature="]] .. encodedSignature .. [["]]
+
+        local ok, res = helpers.proxy_client_grpc(){
+          service = "hello.HelloService.SayHello",
+          opts = {
+            [""] = ("-H 'Date: %s' -H 'Authorization: %s'"):format(date, hmacAuth),
+          },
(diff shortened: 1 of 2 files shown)


spec/03-plugins/19-hmac-auth/03-access_spec.lua
```diff
@@ -420,6 +420,25 @@ for _, strategy in helpers.each_strategy() do
...
+        local hmacAuth = [[hmac username="bob",algorithm="hmac-sha1",]]
+          .. [[headers="date @request-target",signature="]] .. encodedSignature .. [["]]
+
+        local ok, res = helpers.proxy_client_grpc(){
+          service = "hello.HelloService.SayHello",
+          opts = {
+            [""] = ("-H 'Date: %s' -H 'Authorization: %s'"):format(date, hmacAuth),
+          },
+        }
+        print(res)
```
fffonion, line 437: Debug leftover

fffonion (approved):
Need a rebase, otherwise LGTM 👍

PR #7044, opened 2021-04-15
feat(hybrid) auto load cluster_cert or cluster_ca_cert

fffonion:
into lua_ssl_trusted_certificate for data plane

### Summary

Hybrid mode users need to configure `lua_ssl_trusted_certificate`, as it's needed
for Lua code to verify control plane's certificate.
It's causing configuration burden and need extra notice if user is also requiring
other CA on Lua land.

This PR lets Kong to automatically include `cluster_cert` in `lua_ssl_trusted_certificate`
when `cluster_mtls=shared`, or `cluster_ca_cert` if `cluster_mtls=pki`. There's no affect
to Control Plane. User can set other CA bundle in `lua_ssl_trusted_certificate` without
conflict.

Note that adding it to `lua_ssl_trusted_certificate` chain may cause vulnerabilty as the
CA chain is trusted on all connections issued on Lua land, not just hybrid mode connections.
But there's no behavior change in tihs PR regarding this, as it's a limitation from OpenResty.

### Full changelog

* Auto load Control Plane's CA cert into lua_ssl_trusted_certificate on Data Plane


kong/conf_loader/init.lua
```diff
@@ -1053,6 +1053,16 @@ local function check_and_infer(conf, opts)
       errors[#errors + 1] = "only in-memory storage can be used when role = \"data_plane\"\n" ..
                             "Hint: set database = off in your kong.conf"
     end
+
+    if not conf.lua_ssl_trusted_certificate then
+      conf.lua_ssl_trusted_certificate = {}
+    end
+
+    if conf.cluster_mtls == "shared" then
+      table.insert(conf.lua_ssl_trusted_certificate, conf.cluster_cert)
```
dndx, line 1062: We can check to see if `conf.cluster_cert` and `conf.cluster_ca_cert` are already inside `conf.lua_ssl_trusted_certificate` here, in case the user specified it manually using `lua_ssl_trusted_certificate`.

kong/conf_loader/init.lua
```diff
@@ -1053,6 +1053,16 @@ local function check_and_infer(conf, opts)
...
+    elseif conf.cluster_mtls == "pki" then
```
dndx, line 1063:
```suggestion

    elseif conf.cluster_mtls == "pki" then
```

Add a newline before `else`.

kong/conf_loader/init.lua
dndx: Discussed with @fffonion and this should not affect OpenSSL when loading. So mark this as resolved unless it became a problem later.

dndx (approved):

PR #7048, opened 2021-04-15
fix(external-plugins)  response.getXXX() functions from log phase

javierguerragiraldez:
The `kong.response.get_XXX()` functions require data from the response
object, which is no longer accessible in the "post-log" timer used to
call Log handlers in external plugins.

This patch allows those functions to work by adding the required data to
 the set saved at the start of the "real" log phase.

Fix Kong/go-pdk#61

### DIFF
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -151,6 +151,8 @@ local function build_phases(plugin)
           serialize_data = kong.log.serialize(),
           ngx_ctx = ngx.ctx,
           ctx_shared = kong.ctx.shared,
+          response_headers = ngx.resp.get_headers(100),
+          response_status = ngx.status,
         }
 
         ngx_timer_at(0, function()
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -1,8 +1,12 @@
 
 local proc_mgmt = require "kong.runloop.plugin_servers.process"
+local cjson = require "cjson.safe"
+local ngx_ssl = require "ngx.ssl"
 
 local ngx = ngx
 local kong = kong
+local ngx_var = ngx.var
+local coroutine_running = coroutine.running
 local get_plugin_info = proc_mgmt.get_plugin_info
 local ngx_timer_at = ngx.timer.at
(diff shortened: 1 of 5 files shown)


kong/runloop/plugin_servers/init.lua
```diff
@@ -15,6 +17,75 @@ local rpc_notifications = {}
...
+local exposed_api = {
+  kong = kong,
+
+  ["kong.log.serialize"] = function()
+    local saved = save_for_later[coroutine.running()]
+    return cjson.encode(saved and saved.serialize_data or kong.log.serialize())
+  end,
+
+  ["kong.nginx.get_var"] = function(v)
+    return ngx.var[v]
```
dndx, line 29: `ngx.var` is something we can cache at file level, this speed up this potentially hot call path a little considering variable access are notoriously slow in OpenResty already.

kong/runloop/plugin_servers/init.lua
```diff
@@ -15,6 +17,75 @@ local rpc_notifications = {}
 --- currently running plugin instances
 local running_instances = {}
 
+local exposed_api = {
+  kong = kong,
+
+  ["kong.log.serialize"] = function()
+    local saved = save_for_later[coroutine.running()]
```
dndx, line 24: We should be able to cache `coroutine.running` as well to speed up the hot path.

dndx (approved):
Minor comments, otherwise LGTM.

kong/runloop/plugin_servers/init.lua
javierguerragiraldez: this is to allow an external plugin (Go, JS) to do `ngx.var[]`;  caching could defeat the purpose and the performance issue is much less than the RPC overhead.

kong/runloop/plugin_servers/init.lua
javierguerragiraldez: it's used to differentiate the request that is being handled.  the problem is that the `log` phase is handled in a timer context, after the request object itself has been released, so we have to associate those values with the timer instance.  I remember looking for some timer ID but didn't find anything usable.  I'll check again.

dndx (approved):

PR #7052, opened 2021-04-16
fix(dao) upsert existent targets

locao:
Since Kong version 2.2.0, target entities behave as any other entity, but their `upsert` function was not prepared for that.

### DIFF
--- a/kong/db/dao/targets.lua
+++ b/kong/db/dao/targets.lua
@@ -62,7 +62,37 @@ end
 
 function _TARGETS:upsert(pk, entity, options)
   entity.id = pk.id
-  return self:insert(entity, options)
+
+  if entity.target then
+    local formatted_target, err = format_target(entity.target)
+    if not formatted_target then
+      local err_t = self.errors:schema_violation({ target = err })
+      return nil, tostring(err_t), err_t
+    end
+    entity.target = formatted_target
+  end
+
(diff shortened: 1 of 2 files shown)


kong/db/dao/targets.lua
```diff
@@ -62,7 +62,37 @@ end
...
+      local err_t = self.errors:schema_violation({ target = err })
+      return nil, tostring(err_t), err_t
+    end
+    entity.target = formatted_target
+  end
+
+  -- backward compatibility with Kong older than 2.2.0
+  local workspace = workspaces.get_workspace_id()
+  local opts = { nulls = true, workspace = workspace }
+  for existent in self:each_for_upstream(entity.upstream, nil, opts) do
```
dndx, line 78: One quick question, it this something we can address with a migration eventually? Considering the `Target` for a `Upstream` may be a lot, running this loop all the time for `upsert` seems to be less than ideal.

locao: I don't think we can fix this behavior in a migration, this change is to address the issue about using a DB exported from Kong 1.x.x being imported in Kong >= 2.2.0.

I agree that this is not ideal, on the other hand, it's the same behavior used for the `insert` function.

kikito (approved):

dndx (approved):

PR #7085, opened 2021-04-23
fix(*) skip upstreams/targets events on control planes

murillopaula:
Previously, the CP could still do health checks upon CRUD upstreams and targets events. The following logs would be shown on the CPs:

```
2021/04/22 13:00:31 [warn] 99#0: *5924 [lua] balancer.lua:259: callback(): [healthchecks] balancer 5349c4d4-0deb-4528-80e1-59bbfe176452:testUps6066_01 reported health status changed to UNHEALTHY, context: ngx.timer
```

This will disable health checks and balancers completely on CPs by not creating worker/cluster events on them.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -171,6 +171,136 @@ local function csv(s)
 end
 
 
+local function register_balancer_events(core_cache, worker_events, cluster_events)
+  -- target updates
+
+
+  -- worker_events local handler: event received from DAO
+  worker_events.register(function(data)
+    local operation = data.operation
+    local target = data.entity
+    -- => to worker_events node handler
+    local ok, err = worker_events.post("balancer", "targets", {
+        operation = data.operation,
+        entity = data.entity,
+      })
+    if not ok then
+      log(ERR, "failed broadcasting target ",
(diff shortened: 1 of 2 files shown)


javierguerragiraldez (commented):
the fact that all events registered after this point concern the upstreams seems like a serendipitous chance.  I'd believe it would be an easy mistake to add unrelated events at the end of this function.

in this case i do think a simple `if role ~= "control_plane" then ..<everything> end` would be more readable.

even better, pull all those registers to a "register_upstream_events()` function and call it conditionally.

murillopaula:
@javierguerragiraldez thx for the feedback, done in commit https://github.com/Kong/kong/pull/7085/commits/d0b51d722ab8c35ee091710d6456f14dc3d2e1ad

please squash & merge after approval

locao:
Hi @murillopaula. Could you add some tests, please? I think that at least making sure the events exist in `"data_plane"` nodes and don't exist in `"control_plane"` nodes would be enough. The `"traditional"` role should be well covered.

murillopaula:
@locao hey! thx for the feedback

please squash & merge after approval

locao (approved):

PR #7087, opened 2021-04-25
Better error handling

Elyahou:
Try to fix https://github.com/Kong/kong/issues/7093

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -269,10 +269,7 @@ local function execute_plugins_iterator(plugins_iterator, phase, ctx)
       local cok, cerr = coroutine.resume(co, plugin.handler, configuration)
       if not cok then
         kong.log.err(cerr)
-        ctx.delayed_response = {
-          status_code = 500,
-          content = { message  = "An unexpected error occurred" },
-        }
+        kong.response.error(500)
       end
     end
 
@@ -774,7 +771,7 @@ function Kong.access()
 
     ctx.buffered_proxying = nil
 
-    return kong.response.exit(503, { message = "no Service found with those values"})
+    return kong.response.error(503, "no Service found with those values")
   end
 
   runloop.access.after(ctx)
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1090,8 +1090,8 @@ return {
     after = function(ctx)
       local ok, err, errcode = balancer_execute(ctx)
       if not ok then
-        local body = utils.get_default_exit_body(errcode, err)
-        return kong.response.exit(errcode, body)
+        local body = utils.get_default_exit_message(errcode, err)
+        return kong.response.error(errcode, body)
       end
     end
   },
@@ -1121,7 +1121,7 @@ return {
       local router = get_updated_router()
       local match_t = router.exec()
       if not match_t then
-        return kong.response.exit(404, { message = "no Route matched with those values" })
+        return kong.response.error(404, "no Route matched with those values")
       end
 
       ctx.workspace = match_t.route and match_t.route.ws_id
@@ -1186,7 +1186,7 @@ return {
         local redirect_status_code = route.https_redirect_status_code or 426
 
         if redirect_status_code == 426 then
-          return kong.response.exit(426, { message = "Please use HTTPS protocol" }, {
+          return kong.response.error(426, "Please use HTTPS protocol", {
             ["Connection"] = "Upgrade",
             ["Upgrade"]    = "TLS/1.2, HTTP/1.1",
           })
@@ -1206,7 +1206,7 @@ return {
       if (protocols and (protocols.grpc or protocols.grpcs) and http_version ~= 2 and
         (content_type and sub(content_type, 1, #"application/grpc") == "application/grpc"))
       then
-        return kong.response.exit(426, { message = "Please use HTTP2 protocol" }, {
+        return kong.response.error(426, "Please use HTTP2 protocol", {
           ["connection"] = "Upgrade",
           ["upgrade"]    = "HTTP/2",
         })
@@ -1216,7 +1216,7 @@ return {
       if (protocols and (protocols.grpc or protocols.grpcs) and
         (not content_type or sub(content_type, 1, #"application/grpc") ~= "application/grpc"))
       then
-        return kong.response.exit(415, { message = "Non-gRPC request matched gRPC route" })
+        return kong.response.error(415, "Non-gRPC request matched gRPC route")
       end
 
       -- mismatch: grpc request matched grpcs route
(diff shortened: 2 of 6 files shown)


kong/init.lua
```diff
@@ -269,10 +269,7 @@ local function execute_plugins_iterator(plugins_iterator, phase, ctx)
       local cok, cerr = coroutine.resume(co, plugin.handler, configuration)
       if not cok then
         kong.log.err(cerr)
-        ctx.delayed_response = {
-          status_code = 500,
-          content = { message  = "An unexpected error occurred" },
-        }
+        kong.response.error(500)
```
bungle, line 272: This doesn't look right. We need to collect the plugins even in case of an error.

kong/runloop/handler.lua
```diff
@@ -1090,8 +1090,8 @@ return {
     after = function(ctx)
       local ok, err, errcode = balancer_execute(ctx)
       if not ok then
-        local body = utils.get_default_exit_body(errcode, err)
-        return kong.response.exit(errcode, body)
+        local body = utils.get_default_exit_message(errcode, err)
+        return kong.response.error(errcode, body)
```
bungle, line 1094: Do we even need to get the default exit message? Isn't the response.error doing it automatically?

kong/runloop/handler.lua
```diff
@@ -1323,8 +1323,8 @@ return {
 
       local ok, err, errcode = balancer_execute(ctx)
       if not ok then
-        local body = utils.get_default_exit_body(errcode, err)
-        return kong.response.exit(errcode, body)
+        local body = utils.get_default_exit_message(errcode, err)
+        return kong.response.error(errcode, body)
```
bungle, line 1327: Same here, let's check if .error already finds the default message.

bungle (changes requested):
Some change requests, and some comments

kong/init.lua
Elyahou: What do you mean by `collect the plugins`

kong/runloop/handler.lua
Elyahou: It does, there is a difference though: `get_default_exit_message` have [message override mechanism](https://github.com/Kong/kong/blob/c8d370475197c2f9585068b930cb84b7db11c133/kong/tools/utils.lua#L1120) that `response.error` do not have.... That's why I let it like that...

kong/runloop/handler.lua
Elyahou: Same answer

kong/init.lua
bungle: @Elyahou, in access phase we need to run through the whole plugin iterator. You cannot exit it while looping.

Elyahou: That's not correct and is not the case already before my changes...

You can see that on line 267 in this file there is a condition before running each plugin `elseif not ctx.delayed_response then`, and this is precisely the variable that is set when a plugin execution error occur.

bungle: @Elyahou, I think I am correct. Ypi added `kong.response.error(500)` that will call `ngx.exit` and jump to `header_filter` phase. You are effectively exiting the loop that collects the plugins:

```lua
for plugin, configuration in plugins_iterator:iterate(phase, ctx) do
```

That loop need to be run full, even when we don't run the plugins themselves. Is this clear enough?

Elyahou: @bungle inside this loop, there is a condition to run plugin when `delay_response ` is true, meaning we are in the access phase, this condition is at line 267:

```
elseif not ctx.delayed_response then
```

Meaning that is ` ctx.delayed_response` is filled, the next plugin will not run.  But `ctx.delayed_response` is precisely the variable that is set when there is an error in a plugin before my changes, meaning that even before my changes, a plugin error cause the iterator to not run the next plugins...

Is this clear ?

Elyahou: Also, when we are in access phase, the `Response.error` function will not exit directly but will add the response to `ctx.delayed_response`: https://github.com/Kong/kong/blob/7e6f8f38a8ff9253ccfe48e5defd17cf15d9aa75/kong/pdk/response.lua#L1028-L1038

bungle: @Elyahou, again, it will end here:
https://github.com/Kong/kong/blob/master/kong/pdk/response.lua#L1055

This will now `yield` the main thread. And this is not what we want. We want the whole iteration to happen as `plugins_iterator:iterate(phase, ctx)` is the thing that collects the plugins. See here we create sub-thread:
https://github.com/Kong/kong/blob/master/kong/init.lua#L278-L279

This is what could be `yielded`, and that is what happens if plugin errors on `access`.

Elyahou: Thanks @bungle for the details, it's weird because I can see that even with these changes, the iterator still go through every plugin on access phase... Even after a plugin error. IT seems like the coroutine yield so not stop it...

Anyway, I can change it to call `Response.error` inside coroutine to avoid this issue, something like:

```lua
local function access_phase_runner(access_function, handler, configuration)
  return function(handler, conf)
    local ok, err = pcall(access_function, handler, conf)

    if not ok then
      kong.log.err(err)
      kong.response.error(500)
    end
  end
end

local function execute_plugins_iterator(plugins_iterator, phase, ctx)
...
for plugin, configuration in plugins_iterator:iterate(phase, ctx) do
...
    if not delay_response then
      plugin.handler[phase](plugin.handler, configuration)

    elseif not ctx.delayed_response then
      local co = coroutine.create(access_phase_runner(plugin.handler.access))
      coroutine.resume(co, plugin.handler, configuration)
    end
    ...
  end
...
end
```

What do you think ?

bungle: @Elyahou, yes it could be that `yielding` main thread will just make worker to jump to other thread and resume later. This seems to be confirming that: https://github.com/openresty/lua-nginx-module#ngxthreadspawn. Yielding main thread gives CPU to other thread, so that is a slightly different than what it originally was, though the end result is the same.

thibaultcha: Careful about returning closures on the proxy's hot code path though - they have a noticeable effect on memory usage (GC pressure) and overall throughput.

thibaultcha: See this commit for background info on this part of the runloop (delayed responses): https://github.com/Kong/kong/commit/e709c953933e5ca6f6e9766fde15d8fe98a80721

Elyahou: Thank you @thibaultcha for these references. This coroutine yielding cause me to think that the `Response.error` function should not be called from outside a plugin at all... But still, I think errors formatting should be centralised... So maybe, we should extract the formating logic outside to some `utils` and use it here and also in the `Response.error` function. What do you think ?

esatterwhite:
Is this still being worked on? I need this as well

esatterwhite:
@bungle any reason not to use this: https://github.com/bungle/lua-resty-template

esatterwhite:
@Elyahou are you still planning to complete this?

hbagdi:

esatterwhite:
It looked like it stalled waiting from input from Kong reviewers. Can they chime in?

hbagdi:
Please rebase.

esatterwhite:
Can we get feed back from the core team on this?

PR #7098, opened 2021-04-28
docs(gateway-oss) update the upstream slots description

locao:


autodoc/admin-api/data/admin-api.lua
```diff
@@ -1421,7 +1421,7 @@ return {
         id = { skip = true },
         created_at = { skip = true },
         ["name"] = { description = [[This is a hostname, which must be equal to the `host` of a Service.]] },
-        ["slots"] = { description = [[The number of slots in the loadbalancer algorithm (`10`-`65536`).]] },
+        ["slots"] = { description = [[The maximum number of slots in the `round-robin` load balancer algorithm or the actual number of slots in the `consistent-hashing` algorithm (`10`-`65536`).]] },
```
lena-larionova, line 1424:
```suggestion
        ["slots"] = { description = [[The number of slots in the load balancer algorithm. If `algorithm` is set to `round-robin`, this setting determines the maximum number of slots. If `algorithm` is set to `consistent-hashing`, this setting determines the actual number of slots in the algorithm. Takes an integer in the range `10`-`65536`.]] },
```
I wouldn't get rid of the original line here, it's still useful.

Not sure if "integer" is the right data type here, or if it should just be "number" (or something else).

locao: That's indeed more clear. I'm adding your suggestion, thank you!

lena-larionova (approved):

PR #7100, opened 2021-04-29
perf(uri) optimize uri normalization code

bungle:
### Summary

This commit tries to do several optimization on our url normalization code that is executed on our hot path.

Optimizations:
1. runs regex only when it looks like it is needed
2. does `/..`, `/.`, and `//` normalization only if needed
3. moves anonymous callback functions to module local functions

### DIFF
--- a/kong/tools/uri.lua
+++ b/kong/tools/uri.lua
@@ -9,7 +9,6 @@ local string_byte = string.byte
 local string_format = string.format
 local tonumber = tonumber
 local table_concat = table.concat
-local table_clear = require("table.clear")
 local ngx_re_gsub = ngx.re.gsub
 
 
@@ -38,12 +37,7 @@ local TMP_OUTPUT = require("table.new")(16, 0)
 local DOT = string_byte(".")
 local SLASH = string_byte("/")
 
-
-function _M.normalize(uri, merge_slashes)
-  table_clear(TMP_OUTPUT)
-
-  -- Decoding percent-encoded triplets of unreserved characters
-  uri = ngx_re_gsub(uri, "%([\\dA-F]{2})", function(m)
+local function percent_decode(m)
     local hex = m[1]
     local num = tonumber(hex, 16)
     if RESERVED_CHARACTERS[num] then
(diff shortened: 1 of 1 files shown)


kong/tools/uri.lua
```diff
@@ -35,86 +34,133 @@ local RESERVED_CHARACTERS = {
...
+  if uri == "/." or uri == "/./" or uri == "/.." or uri == "/../" then
+    return SLASH
+  end
 
-    elseif FIRST == DOT and SECOND == DOT and THIRD == SLASH then -- ../
-      uri = string_sub(uri, 4)
+  if string_find(uri, "/.",  -2, true) == nil and
+     string_find(uri, "/..", -3, true) == nil and
+     string_find(uri, "/./",  1, true) == nil and
+     string_find(uri, "/../", 1, true) == nil
```
kikito, line 74: Two questions about this:
a) Can't the result of these calls to `string_find` be used/reused on the code below? Aren't the subsequent calls to `string_find` redoing work?
b) What is the logic behind doing these checks in this particular order vs the reverse (doing the "start checks" first?

bungle: I changed this to just check the `.`.

dndx:
Do we have some benchmark of the performance gain of this change vs the old code?

bungle:
> Do we have some benchmark of the performance gain of this change vs the old code?

@fffonion made a test and it provided 13% improvement, though his use case did not test the remove dot processing as this is now intelligent enough to skip it in most cases.

fffonion:
@bungle I was also reading code before spent time benchmarking and would like to share my thoughts here. Generally normalization is not a feature that user relies (or should rely) regularly, it's more of a security fix for Kong as certain authentication process requires a normalized path to be correctly intercepted.

That means if we optimize the normalization code itself it is likely a "non-optimisitc" path that most user won't hit on. In my opinion, it's acceptable even if this path is slightly slower. A legit request can get better performance if it's normalized beforehand (which is most browsers already did right now), and it's not that slow to be DoS-able.

In fact, the most valuable part for optimization is to decrease the overhead to those requests that don't require normalization, but were uncoditionally normalized (like `runs regex only when it looks like it is needed`). We can further
improve this by only detecting `%` if it's before appearance of a `?` for example. I would suggest to simply the PR a bit,
so that codebase is less complicated and test is more reliable.

fffonion (approved):

PR #7102, opened 2021-04-29
fix(proxy) the context got shared between different tls keepalive requests

### LINKED ISSUE #7054: Plugin incorrectly active in subsequent calls on connection keep-alive (problem new since 2.4.0)
### Summary

When two calls are made to the same server+hostname but different routes and using connection keep-alive, then plugins that are active in the first call are also (incorrectly) active in the second call. 
The plugin is configured on the route belonging to the first call. The second call uses a different uri (i.e. route) where the plugin is not configured. So the plugin must not be called there.

i.e.
myserver.com/route1 -> plugin active
myserver.com/route2 -> plugin not active

Requesting myserver.com/route1 and myserver.com/route2 seperately (no connection keep-alive) results in everything working fine. Requesting both in one connection results in the plugin being active in both requests.

The problem first appeared after upgrading from kong 2.3.3 to kong 2.4.0. In 2.3.3 everything was ok. No change was made while upgrading to 2.4.0 - just the upgrade itself by switching docker base image from kong:2.3 to kong:2.4.

### Steps To Reproduce

In my case:

1. Plugin is kong-referer-restriction (https://github.com/dsteinkopf/kong-referer-restriction). This is configured for myserver.com/route1 and checks if the referer contains 'myserver.com'. Result is http 403 if not.
2. The plugin is not configured for myserver.com/route2
4. `curl https://myserver.com/route2` results correctly in status 200 which shows that the plugin is correctly inactive here.
3. Do two consecutive calls using the same connections: `curl -v -K curl.txt`.  curl.txt:
```
url = "https://myserver.com/route1"
referer = "https://myserver.com/test"
output = "dummy1"
next
url = "https://myserver.com/route2/"
output = "dummy2"
```
Output is:
```
...
< HTTP/1.1 200 OK
...
< Via: kong/2.4.0
< 
{ [1968 bytes data]
100  1956    0  1956    0     0  17621      0 --:--:-- --:--:-- --:--:-- 17621
* Connection #0 to host myserver.com left intact
* Found bundle for host myserver.com: 0x7fb38ae18230 [can pipeline]
* Could pipeline, but not asked to!
* Re-using existing connection! (#0) with host myserver.com
* Connected to myserver.com (xxx) port 443 (#0)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
> GET /route2/ HTTP/1.1
...
< HTTP/1.1 403 Forbidden
...
```
This shows that the plugin is active is (incorrectly) active in this case.

### Additional Details & Logs

- Kong version (`$ docker-compose exec kong kong version`) -> 2.4.0
- Kong debug-level startup logs (`$ kong start --vv`) -> How can I do this when using docker?
- Kong error logs (`<KONG_PREFIX>/logs/error.log`) -> I don't see any errors in my logging server.
- Operating system: Docker hub image kong:2.4
  https://docs.konghq.com/latest/admin-api/#retrieve-node-information) -> I'll try to give some extract here, if needed.

bungle:
### Summary

@dsteinkopf reported on #7054 that when executing a plugins iterator with two subsequent tls keepalive request, the wrong plugin is executed. We found out that we do pass context in ssl phase to plugin iterator and that indeed creates a connection wide structures in that context, which are then shared between different keepalive requests, and thus a wrong plugin might be executed.

This commit changes it so that we do not pass context to plugin iterator with `certificate` phase and thus it won't store data in it.

Huge thanks for @dsteinkopf for reporting this!

### Issues Resolved

Fix #7054

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -631,6 +631,7 @@ end
 
 
 function Kong.ssl_certificate()
+  -- Note: ctx here is for a connection (not for a single request)
   local ctx = ngx.ctx
 
   kong_global.set_phase(kong, PHASES.certificate)
@@ -644,6 +645,9 @@ function Kong.ssl_certificate()
   local plugins_iterator = runloop.get_updated_plugins_iterator()
   execute_plugins_iterator(plugins_iterator, "certificate", ctx)
   runloop.certificate.after(ctx)
+
+  -- TODO: do we want to keep connection context?
+  kong.table.clear(ngx.ctx)
 end
 
 
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1058,7 +1058,7 @@ return {
     after = NOOP,
   },
   certificate = {
-    before = function(_)
+    before = function(ctx) -- Note: ctx here is for a connection (not for a single request)
       certificate.execute()
     end,
     after = NOOP,
--- a/spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
+++ b/spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
@@ -1207,4 +1207,58 @@ for _, strategy in helpers.each_strategy() do
       end
     end)
   end)
+
+  describe("Plugins triggering [#" .. strategy .. "] with TLS keepalive", function()
+    lazy_setup(function()
+      local bp = helpers.get_db_utils(strategy, {
+        "routes",
+        "services",
+        "plugins",
+      })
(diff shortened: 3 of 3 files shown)


guanlan:
Could it br possible to test the change?

dndx (approved):

PR #7112, opened 2021-05-03
fix(clustering) constant cache flips on dp because of invalidations event

bungle:
### Summary

@murillopaula let us know about hybrid mode behavior that caused constant flips of data plane configuration even when no change was actively made with admin api.

It turned out that our `clustering` module subscribes to `invalidations`, and on each `invalidations` event it starts to push configs to data planes connected to it. With further look it was also found that the `clustering` module will by itself generate `invalidations` events when it updates the `clustering_data_planes` table about the information of the data plane that connected to it (e.g. last seen status). This caused constant `invalidations` happen across the control plane cluster and then in turn caused constant sending of config and config flips on data planes.

The solution was to make check for handling the `dao:crud` event more strict, and make that to create another cluster event called `clustering:invalidation`, that in turn causes config push. This indirection allows us to filter those invalidations events that don't cause actual database change. And the clustering module does not subscribe anymore to the generic `invalidations` event, which has a more broad scope than database entity invalidations.

### DIFF
--- a/kong/clustering.lua
+++ b/kong/clustering.lua
@@ -84,7 +84,7 @@ local function update_config(config_table, update_cache)
   end
 
   if declarative.get_current_hash() == new_hash then
-    ngx_log(ngx_DEBUG, "same config received from control plane,",
+    ngx_log(ngx_DEBUG, "same config received from control plane, ",
             "no need to reload")
     return true
   end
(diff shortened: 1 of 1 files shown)


kong/clustering.lua
```diff
@@ -941,35 +941,45 @@ function _M.init_worker(conf)
...
     end
 
     -- The "invalidations" cluster event gets inserted in the cluster when there's a crud change
     -- (like an insertion or deletion). Only one worker per kong node receives this callback.
     -- This makes such node post push_config events to all the cp workers on its node
-    kong.cluster_events:subscribe("invalidations", post_push_config_event_to_node_workers)
+    kong.cluster_events:subscribe("clustering:invalidation", handle_clustering_invalidations_event)
 
     -- The "dao:crud" event is triggered using post_local, which eventually generates an
     -- "invalidations" cluster event. It is assumed that the workers in the
```
fffonion, line 977: -- "clustering:invalidations" cluster event.

kong/clustering.lua
```diff
@@ -941,35 +941,45 @@ function _M.init_worker(conf)
...
+      end
+    end
+
+    -- Sends "clustering", "push_config" to all workers in the same node, including self
+    local function handle_clustering_invalidations_event(data)
+      ngx_log(ngx_DEBUG, "received clustering:invalidation event for ", data)
+      post_push_config_event()
+    end
+
     -- Sends "clustering", "push_config" to all workers in the same node, including self
```
fffonion, line 957: need to update this comment as well

murillopaula (approved):
lgtm, tested locally

gszr:
@bungle is it possible to add a regression test to this?

dndx (approved):

kong/clustering.lua
```diff
@@ -942,34 +942,47 @@ function _M.init_worker(conf)
...
+    end
+
+
+    -- Handles "dao:crud" worker event and broadcasts "clustering:invalidation" cluster event
+    local function handle_dao_crud_event(data)
+      if type(data) ~= "table" or data.schema == nil or data.schema.db_export == false then
         return
       end
 
+      kong.cluster_events:broadcast("clustering:invalidation", data.schema.name .. ":" .. data.operation)
```
dndx, line 965: Since this event does not come from or are driven by the DAO subsystem anymore, maybe `clustering:push_config` makes this difference more obvious?

bungle:
> @bungle is it possible to add a regression test to this?

I can try.

dndx (approved):

PR #7114, opened 2021-05-03
fix(build) Do not hardcode grpcurl version when downloading

tyler-ball:
### Summary

As soon as we want to upgrade `grpcurl` this will need to not be hardcoded

### Full changelog

* fix(build) Do not hardcode grpcurl version when downloading

### Issues resolved

N/A


gszr (approved):
That was an overlook on our end. Thanks for the PR @tyler-ball.

locao (approved):

PR #7118, opened 2021-05-05
fix(proxy) ctx robustness?

bungle:
### Summary

@jeremyjpj0916 reported on #7105 about some strange issues with Cassandra and missing
workspace in some cache keys. I could not figure out what it was but made some small
changes in hopes that it could have some effect.


jeremyjpj0916:
noting here this likely won't be needed as we got to bottom of the problem in my git issue and it wasn't core Kong related.

PR #7121, opened 2021-05-05
chore(ci) remove bintray, add pulp credentials

gszr:


Jenkinsfile
```diff
@@ -11,6 +11,9 @@ pipeline {
         DOCKER_PASSWORD = "${env.DOCKER_CREDENTIALS_PSW}"
         KONG_PACKAGE_NAME = "kong"
         DOCKER_CLI_EXPERIMENTAL = "enabled"
+        PULP = credentials('PULP')
+        PULP_HOST = "https://api.pulp.konnect-prod.konghq.com"
+        GITHUB_TOKEN = credentials('github_bot_access_token')
```
hutchic, line 16: I _think_ the reason we copy-pasta'd was because the `env` block only applies to the `agent` block in which it belongs to

I suppose we could test this with a Jenkins hello world type of file

hutchic (approved):

Jenkinsfile
gszr: Tested and ensured these env vars apply globally.

kikito (approved):

PR #7139, opened 2021-05-11
feat(conf)  use a config variable to select module variants

javierguerragiraldez:
the `set_variants()` function installs a preloader for a module, which
will use one of the specified filepaths (not using `package.path`,
depending on a configuration value.

### DIFF
--- a/kong/conf_loader/feature_flags.lua
+++ b/kong/conf_loader/feature_flags.lua
@@ -1,0 +1,34 @@
+
+local kong = kong
+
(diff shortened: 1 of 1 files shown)


bungle:
@javierguerragiraldez, could you give us a small example how this works? And should there be at least some tests for it?

javierguerragiraldez:
i tried to use it for the balancer but fell into conflicting requirements, closing this for now

PR #7146, opened 2021-05-11
fix(db) correct key for select_by_cache_key in off strategy

fffonion:
### Summary

`select_by_cache_key` looks up an entity in db or cache with its cache_key. In `off` strategy,
the function delegates to `select_by_key` which ultimately calls `kong.core_cache:get`.

Since `off` strategy already uses the cache_key to populate declarative config into cache,
it's sufficient to use the provided `field` directly into `select_by_key`.

### Full changelog

* correct key for select_by_cache_key in off strategy

### Issues resolved

Fix #XXX

### DIFF
--- a/kong/db/strategies/off/init.lua
+++ b/kong/db/strategies/off/init.lua
@@ -238,6 +238,7 @@ local function select_by_field(self, field, value, options)
 
   local ws_id = ws(self, options)
 
+  local key
   if field ~= "cache_key" then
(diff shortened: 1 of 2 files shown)


bungle (approved):

gszr (commented):
is there a way to add a regression test to this?

fffonion:
@gszr yes test added!

gszr (approved):

PR #7151, opened 2021-05-12
<style>(<file log>) <add timestamp prefix before each JSON entry for Splunk index>

steven4coding:
<BLANK LINE>
<body> add timestamp before each log entry so that Splunk can index each request&response to a event correctly
<BLANK LINE>
<footer>

### Summary

Currently, Splunk cannot index the log from file log plugin properly because each line is just a JSON without timestamp as prefix, multiple requests and responses are grouped into one event, which is not easy to search particular log. This PR is to add the timestamp prefix in the log file so that Splunk can index the log correctly, which means each entry in the file log is an event in Splunk.

### Full changelog

* Add a timestamp prefix

### DIFF
--- a/kong/plugins/file-log/handler.lua
+++ b/kong/plugins/file-log/handler.lua
@@ -40,7 +40,7 @@ local file_descriptors = {}
 -- @param `conf`     Configuration table, holds http endpoint details
 -- @param `message`  Message to be logged
 local function log(conf, message)
-  local msg = cjson.encode(message) .. "\n"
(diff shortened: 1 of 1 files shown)


dndx:
Hello,

Thanks for your contribution. However, the addition of timestamp at the beginning of the line changes the format of the plugin's output, therefore it would be considered as a breaking change. Therefore it does not seems to be a good idea to enable this without a switch.

Also, could you write a test case for this change as well?

Thanks!

Issue #7148, opened 2021-05-12
`no plugin instance` error appears when renewing the cache (KONG_DB_CACHE_TTL parameter) with custom go plugin 


ealogar:
### Summary

Recently I have been trying to upgrade from kong 2.0.4 to kong 2.4 (and kong 2.3.3 later) and It has happened a very weird behaviour in a custom go plugin (global plugin, called jwe, without service or route directly linked). I have also migrated the plugin to the new server standalone plugin that communicate with kong directly with protobuf without the go-pluginserver.
Everything has been done successfully.
This plugin does a decryption of an authorization header JWE (Json web encrypted token) and propagates to another plugin chains (your jwt plugin and headers-transform... basically). It's configured the first one in the plugin chain.

When i run a jmeter test with a single thread, everything is worked fine for several hours; that is, the global plugin is executed correctly and the bearer token is decrpted and propagated to jwt plugin.

But the problema arises after some time I run a jmeter test with several theads and a bit more load (50 100 threads).

We deploy the kong docker in a k8s pod deployment.

When the pod is started, everything is worked perfectly (with 500-800 tps) but after some time (deterministic time, I will explain later) some errors begin to appear:

The following errors appears in 2.3.3:

```
2021/05/12 13:54:53 [error] 27#0: *253919 [kong] init.lua:332 [jwe] no plugin instance 123499, client: 10.240.1.3, server: kong, request: "POST /mockbin/v1/echo HTTP/1.1", host: "example.es"
2021/05/12 13:54:53 [error] 26#0: *253921 [kong] init.lua:332 [jwe] no plugin instance 123514, client: 10.240.2.28, server: kong, request: "POST /mockbin/v1/echo HTTP/1.1", host: "example.es"
```
In kong 2.4, the instance id is always 0:
```
2021/05/12 11:52:21 [error] 28#0: *138170 [kong] mp_rpc.lua:308 [jwe] no plugin instance 0, client: 10.240.2.28, server: kong, request: "POST /mockbin/v1/echo HTTP/1.1", host: "example.es"
2021/05/12 11:52:21 [error] 28#0: *138723 [kong] mp_rpc.lua:308 [jwe] no plugin instance 0, client: 10.240.2.28, server: kong, request: "POST /mockbin/v1/echo HTTP/1.1", host: "example.es"

```

[...]

PR #7153, opened 2021-05-13
fix(external-plugins): use lowercase for comparing string in error condition

ealogar:
### Summary

When an instance_id of a plugin external changes, the plugin instance pretended to be reseted and retried in the bridge event loop again.
This was failing and none was being retried, making the call to the plugin did not happen; it was due to a typo in the comparisson.
This has been identified and tested with heavy traffic when the kong balancer is refreshed and reload into memoy cache; it was not failing with low traffic.

### Issues resolved

Fix https://github.com/Kong/kong/issues/7148

### DIFF
--- a/kong/db/dao/plugins/go.lua
+++ b/kong/db/dao/plugins/go.lua
@@ -518,7 +518,7 @@ local get_plugin do
 
             local instance_id = get_instance(plugin_name, conf)
             local _, err = bridge_loop(instance_id, phase)
-            if err and string.match(err, "No plugin instance") then
+            if err and string.match(err:lower(), "no plugin instance") then
               instance_id = reset_and_get_instance(plugin_name, conf)
               bridge_loop(instance_id, phase)
             end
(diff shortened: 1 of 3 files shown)


ealogar:
4 hours load test without failures @javierguerragiraldez , Now you can decide if do the fix just in 2.4 or port it to 2.3

javierguerragiraldez (approved):

ealogar:
I have seen in logs of nginx too many times (in high load) the folllowing annonying log:
```
2021/05/15 17:37:16 [error] 27#0: *855540 [kong] mp_rpc.lua:308 [jwe] no plugin instance 0, client: 10.240.2.28, server: kong, request: "POST /token HTTP/1.1", host: "example.es"
```
As this error is retried later and the request will succed an call the plugin, I would either change the log level to notice or even remove the error log

What do you think @javierguerragiraldez ?

palmerabollo:
I agree that the error is annoying and makes it difficult to operate kong. Would it be possible to change it to level "notice" @javierguerragiraldez?

PR #7155, opened 2021-05-14
fix(router) limit path regex to a lower backtracking limit

javierguerragiraldez:
TODO:
 - come up with DoS exploits that would be stopped by this limit
 - apply similar limits to other user-input regexes

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -904,9 +904,10 @@
 #nginx_admin_client_body_buffer_size = 10m  # Defines the buffer size for reading
                                             # the request body on Admin API.
 
-#nginx_http_lua_regex_match_limit = 100000  # Global MATCH_LIMIT for PCRE
-                                            # bounds most degenerate cases
-                                            # to less than 2 seconds.
+#nginx_http_lua_regex_match_limit = 100000  # Global `MATCH_LIMIT` for PCRE
+                                            # regex matching. The default of `100000` should ensure
+                                            # at worst any regex Kong executes could finish within
+                                            # roughly 2 seconds.
 
 #------------------------------------------------------------------------------
 # DATASTORE
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -904,6 +904,10 @@
 #nginx_admin_client_body_buffer_size = 10m  # Defines the buffer size for reading
                                             # the request body on Admin API.
 
+#nginx_http_lua_regex_match_limit = 100000  # Global MATCH_LIMIT for PCRE
+                                            # bounds most degenerate cases
+                                            # to less than 2 seconds.
+
 #------------------------------------------------------------------------------
 # DATASTORE
 #------------------------------------------------------------------------------
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -32,6 +32,7 @@ local max           = math.max
 local band          = bit.band
 local bor           = bit.bor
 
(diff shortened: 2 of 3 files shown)


thibaultcha:
Can we use [lua_regex_match_limit](https://github.com/openresty/lua-nginx-module#lua_regex_match_limit) for this?

javierguerragiraldez:
> Can we use [lua_regex_match_limit](https://github.com/openresty/lua-nginx-module#lua_regex_match_limit) for this?

yes, but that's a global limit.   we'll probably need different limits for things that are done at routing, vs response time, vs log, etc.  not to mention user-defined regexes vs. those that are constants in Kong itself.

we still have to do some experiments on "realistic" problem regexes; but what i feel could be the end use is something like:

- set a high global limit, but not as high as the default of 10'000,000 (which at some point was chugging for more than 7 hours on a single regex by accident (not an intentional attack) at cloudflare).  perhaps something like 500,000?  Goal: never execute in more than a few seconds.
- identify some "not too critical" user-input regexes (like part of some loading where the result is stored or cached, and the regex itself isn't repeated,  or executed only at log time) use a limit like 100,000 or so.  Goal limit: less than a second.
- focus on those user-input that are in the hot path, like these for routing.  they're of course precompiled, and even some preprocessing.  but execution itself should be tightly bound, and limits like 10,000 or 20,000 are reasonable.  Goal limit: few milliseconds.

dndx:
@javierguerragiraldez @thibaultcha,

I think setting a slightly lower global limit would be okay for now. IMO it may offer better protection than just setting the limits on the Router matching.

Since the default has no limit at all, we need to do something to make it a bit more reasonable first. We could still keep the router specific limit if it also helps, but the global limit seems to be a must have since we can not cover all use cases right away.

javierguerragiraldez:
After many and verifying the PCRE 8.44 Changelog, found that a global limit of 100,000 limits the worst degenerate case I could find (which is far more complex than the old `(a+)+b`) to less than 2seconds, and a limit of 10,000 keeps below 14ms for the worst case and <1msec except for a narrow band of string lengths (either shorter or longer than that and the match resolves immediately)

kong.conf.default
```diff
@@ -904,6 +904,10 @@
 #nginx_admin_client_body_buffer_size = 10m  # Defines the buffer size for reading
                                             # the request body on Admin API.
 
+#nginx_http_lua_regex_match_limit = 100000  # Global MATCH_LIMIT for PCRE
+                                            # bounds most degenerate cases
+                                            # to less than 2 seconds.
```
dndx, line 909:
```suggestion
#nginx_http_lua_regex_match_limit = 100000  # Global `MATCH_LIMIT` for PCRE
                                            # regex matching. The default of `100000` should ensure
                                            # at worst any regex Kong executes could finish within
                                            # roughly 2 seconds.
```

gszr: "any regex Kong executes could finish within roughly 2 seconds"

I think this phrasing might be a bit misleading. This ensures an upper bound in the execution time, but some regexes will finish with an error

javierguerragiraldez: either an error on a false negative.   but i think it's ok; i think any expression that takes so long is wrong already.  also, in the vast majority of cases, the degenerate behaviour happens only when no match is possible, so failing to match is the right answer, even if the engine couldn't test every possible permutation.

dndx (approved):

PR #7162, opened 2021-05-18
Refactor/balancer

javierguerragiraldez:
big refactoring to assimilate the balancer code into the core


kong/runloop/balancer/balancers.lua
```diff
@@ -0,0 +1,549 @@
...
+  end
+end
+
+function balancer_mt:addAddress(target, entry)
+  -- from host:addAddress
+  if type(entry) ~= "table"
+    or type(target) ~= "table"
+    or target.balancer ~= self
+  then
+    return
```
dndx, line 342: Should this throw an error?

kong/runloop/balancer/targets.lua
```diff
@@ -0,0 +1,490 @@
...
+
+
+local targets_M = {}
+
+-- forward local declarations
+local resolve_timer_callback
+local queryDns
+
+function targets_M.init()
+  require "kong.tools.dns"(kong.configuration)    -- configure DNS client
```
dndx, line 44:
```suggestion
  require("kong.tools.dns")(kong.configuration)    -- configure DNS client
```

Was a bit hard to read at first 😂

dndx (approved):
I did not found anything obviously wrong and we had the review meeting already, so let's merge it for now and test it out first.

kong/runloop/balancer/balancers.lua
javierguerragiraldez: yes, it would be better.

of course, it wouldn't do any difference, since the calling code (at least from target/update_dns_result) doesn't check the return.  but that's a good candidate for refactoring, I'd like the update_dns_result function to be much shorter and easier to read, with most of the duplicates/updating complexities moved into some high-level balancer/target methods.

PR #7167, opened 2021-05-18
fix(sandbox) allow string.rep in sandboxed environments

### LINKED ISSUE #7062: Kong 2.4.0 Cannot load kong.tools.timestamp in sandbox
Hi,

I'm not able to convert unix timestamp to a human readable format with new custom_fields_by_lua logic of file-log plugin.

Plugin configuration and lua code:
```
    plugins:
      - name: file-log
        config:
          path: /dev/stdout
          custom_fields_by_lua:
            timestamp: |
              local tools = require "kong.tools.timestamp"
              return tools.get_timetable(tools.get_utc_ms()):rfc_3339()
```

Regarding ENV variables:
```
        - name: KONG_UNTRUSTED_LUA_SANDBOX_REQUIRES
          value: kong.tools.timestamp
        - name: KONG_PLUGINS
          value: zipkin,prometheus,mtls-lua,file-log-extended,file-log
```

Error:
``` 
2021/04/19 14:11:29 [error] 24#0: *594 failed to run log_by_lua*: /usr/local/share/lua/5.1/kong/tools/kong-lua-sandbox.lua:170: /usr/local/share/lua/5.1/luatz/tzfile.lua:59: attempt to call method 'rep' (a nil value)
stack traceback:
        [C]: in function 'error'
        /usr/local/share/lua/5.1/kong/tools/kong-lua-sandbox.lua:170: in function </usr/local/share/lua/5.1/kong/tools/kong-lua-sandbox.lua:154>
        /usr/local/share/lua/5.1/kong/plugins/file-log/handler.lua:79: in function </usr/local/share/lua/5.1/kong/plugins/file-log/handler.lua:75>
        /usr/local/share/lua/5.1/kong/init.lua:265: in function 'execute_plugins_iterator'
        /usr/local/share/lua/5.1/kong/init.lua:1300: in function 'log'
        log_by_lua(nginx-kong.conf:112):2: in main chunk while logging request, client: 192.168.18.1, server: kong, request: "POST /reqcb/c HTTP/1.1", upstream: xxxx
``` 

Regarding sandbox documentation the function string.rep is unsafe. 

**Buf if I add "bundled" to KONG_PLUGINS, it works without any errors.**

Why it is working, if I load bundled plugins?
How to convert timestamps to a human readable format without using unsafe functions?
If there is no way, what do I need to add to KONG_UNTRUSTED_LUA_SANDBOX_REQUIRES variable to get it running?

Thank you!

Regards
Thomas

### LINKED ISSUE #7133: Bug: Kong 2.3.x serverless-functions sandbox modifies Lua global environment
### Summary

The sandbox capability for serverless-functions (pre-function / post-function) that was added in 2.3.0 is breaking other plugins that call `string.rep()`.  Once any pre-function or post-function plugin has been executed at least one time, other plugins are no longer able to call `string.rep()` until Kong is restarted.  Any plugin that attempts to call `string.rep()` will generate an error.  Example:
```
2021/05/10 04:43:43 [error] 24#0: *2831 [kong] init.lua:271 [my-custom-plugin] .../share/lua/5.1/kong/plugins/my-custom-plugin/handler.lua:7: attempt to call field 'rep' (a nil value), client: 172.17.0.1, server: kong, request: "GET /route1 HTTP/1.1", host: "localhost:8000"
```
### Steps To Reproduce

Please see the following repo for a test case, and follow the instructions in the repo's README
https://github.com/chris-branch/kong-sandboxbug

### Additional Details & Logs

My understanding is that the sandboxing should be isolated to only the serverless-function code that is being executed.  However, it appears to be modifying the global environment that affects all Lua code.  Note that the sandbox code appears to hide a number of functions in addition to `string.rep()`.  I have not tested these other functions, but that is something that should be investigated as well.

https://github.com/Kong/kong/blob/master/kong/tools/kong-lua-sandbox.lua#L46-L56

- Kong version 2.3.3

kikito:
Reasoning: the idea behind limiting access to string.rep was
disallowing a single operation from allocating too much memory.

Given that in LuaJIT there are no debug hooks, it is trivial to
implement a loop which allocates a lot of memory on every single
iteration.

Plus, given that the `string` table was global and obtainable via any
sandboxed string, its sandboxing provoked issues on the global state.

This change accepts string.rep in sandboxed environments as a practical
compromise.

Fixes #7062
Fixes #7133

### DIFF
--- a/spec/01-unit/020-sandbox_spec.lua
+++ b/spec/01-unit/020-sandbox_spec.lua
@@ -240,6 +240,11 @@ describe("sandbox functions wrapper", function()
           _G.kong.configuration = deep_copy(base_conf)
         end)
 
+        it("has access to string.rep", function()
+          assert.same("aaa", sandbox.sandbox("return string.rep('a', 3)")())
+          assert.is_true(sandbox.sandbox("return ('a'):rep(3) == 'aaa'")())
+        end)
+
         it("has access to config.untrusted_lua_sandbox_environment", function()
           for _, m in ipairs(modules) do
             assert.same(find(m, _G), sandbox.sandbox(fmt("return %s", m))())
--- a/kong/tools/kong-lua-sandbox.lua
+++ b/kong/tools/kong-lua-sandbox.lua
@@ -1,5 +1,5 @@
 local sandbox = {
-  _VERSION      = "kong-lua-sandbox 1.0",
+  _VERSION      = "kong-lua-sandbox 1.1",
   _DESCRIPTION  = "A pure-lua solution for running untrusted Lua code.",
   _URL          = "https://github.com/kong/kong-lua-sandbox",
   _LICENSE      = [[
@@ -43,7 +43,6 @@ local BASE_ENV = {}
 
 -- List of unsafe packages/functions:
 --
--- * string.rep: can be used to allocate millions of bytes in 1 operation
 -- * {set|get}metatable: can be used to modify the metatable of global objects (strings, integers)
 -- * collectgarbage: can affect performance of other systems
 -- * dofile: can access the server filesystem
@@ -74,8 +73,8 @@ math.sin   math.sinh math.sqrt  math.tan  math.tanh
 os.clock os.difftime os.time
 
 string.byte string.char  string.find  string.format string.gmatch
-string.gsub string.len   string.lower string.match  string.reverse
-string.sub  string.upper
+string.gsub string.len   string.lower string.match  string.rep
+string.reverse string.sub  string.upper
 
 table.insert table.maxn table.remove table.sort
 
@@ -104,8 +103,6 @@ end)
 
 -- auxiliary functions/variables
 
-local string_rep = string.rep
-
 local function sethook(f, key, quota)
   if type(debug) ~= 'table' or type(debug.sethook) ~= 'function' then return end
   debug.sethook(f, key, quota)
@@ -113,7 +110,6 @@ end
 
 local function cleanup()
   sethook()
-  string.rep = string_rep -- luacheck: no global
 end
 
 -- Public interface: sandbox.protect
@@ -160,8 +156,6 @@ function sandbox.protect(code, options)
       end
       sethook(timeout, "", quota)
     end
-
-    string.rep = nil -- luacheck: no global
 
     local t = table.pack(pcall(f, ...))
 


spec/01-unit/020-sandbox_spec.lua
```diff
@@ -240,6 +240,12 @@ describe("sandbox functions wrapper", function()
           _G.kong.configuration = deep_copy(base_conf)
         end)
 
+        it("has access to string.rep", function()
+          for _, m in ipairs(modules) do
+            assert.same("aaa", sandbox.sandbox("return string.rep('a', 3)")())
```
javierguerragiraldez, line 245: `m` is not used.  is that loop needed?

an `('a'):rep(3)=='aaa'` test would be good too

javierguerragiraldez (approved):
LGTM, with minimal comments on the test

dndx:
If we use a proxy function for string.rep and check the resultant size of returned string before calling the actual function, will that mitigate some of the risks?

Something like:

```
function string_rep_proxy(s, count)
  if #s * count > threshold then
    error("return string too large")
  end

  return string.rep(s, count)
end
```

kikito:
> If we use a proxy function for string.rep and check the resultant size of returned string before calling the actual function, will that mitigate some of the risks?

We considered that, but decided not to go on that direction. First, because of the aforementioned ease of recreating the same "memory consumption" with a simple loop. Second, because such limit probably belongs in `kong/globalpatches.lua`, so it applies to Kong as a whole, not just to the sandboxed code. 

Explanation about that last point:

When sandboxing, this:
```
print(string.rep("potato", 3))
```
Can be sandboxed "easily" - that `string` is a "fake" created on the sandbox end. But this isn't so easy:
```
print(("potato"):rep(3))
```
The `rep` there is coming from a global variable shared between the environment and the sandbox - the string metatable. The only way to prevent this second use is modifying that global variable, doing the equivalent of this:

```
old_string_rep = string.rep
string.rep = limited_string_rep
-- ==== start sandbox ====
print(("potato"):rep(3))
-- ==== end sandbox ====
string.rep = old_string_rep
```

Modifying the string metatable before/after each sandbox use is finicky, #7062 and #7133 are proof of that - places that should have access to string.rep sometimes don't find it. If we were doing that change, I would do it globally in `kong/globalpatches.lua`, so they are prevalent through kong, and the behavior is consistent, without "flipping".

spec/01-unit/020-sandbox_spec.lua
kikito: Done!

dndx:
@kikito Okay that makes sense, thanks for the explainations!

dndx (approved):

PR #7190, opened 2021-05-26
chore(cli) emit deprecation warn for lua db_import

gszr:


dndx (commented):
We can write a simple few lines test case for this.

dndx (approved):
Discussed with @gszr and this is a very low visibility feature which itself is untested. So probably doesn't make sense to invest more at this point.

PR #7380, opened 2021-05-27
fix(workspace): do not allow reserved names

jschmid1:
### Summary

You can create a workspace named "plugins" or "services" and nothing in the schema stops a user from doing that.  After which, Kong allows some further API interactions (POST /plugins/services is apparently valid although I couldn't find documentation on it so it may "look" like a valid workspace from within e.g. Manager but POST /plugins/consumers doesn't work (Method not allowed). It's rather hard to find the root cause for random looking issues that occur when using reserved workspace names.

### Full changelog

* Added workspace schema validation function
* Added tests

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -44,6 +44,13 @@ files["kong/hooks.lua"] = {
 }
 
 
+files["kong/db/schema/entities/workspaces.lua"] = {
+    read_globals = {
+        "table.unpack",
+    }
+}
+
+
 files["kong/plugins/ldap-auth/*.lua"] = {
     read_globals = {
         "bit.mod",
--- a/kong/db/schema/entities/workspaces.lua
+++ b/kong/db/schema/entities/workspaces.lua
@@ -1,5 +1,5 @@
 local typedefs = require "kong.db.schema.typedefs"
-
+local constants = require "kong.constants"
 
 return {
   name = "workspaces",
@@ -10,7 +10,7 @@ return {
 
   fields = {
     { id          = typedefs.uuid },
-    { name        = typedefs.utf8_name { required = true } },
+    { name        = typedefs.utf8_name { required = true, not_one_of = { table.unpack(constants.CORE_ENTITIES) }, } },
     { comment     = { type = "string" } },
     { created_at  = typedefs.auto_timestamp_s },
     { meta        = { type = "record", fields = {} } },
--- a/spec/01-unit/01-db/01-schema/15-workspaces_spec.lua
+++ b/spec/01-unit/01-db/01-schema/15-workspaces_spec.lua
@@ -27,6 +27,19 @@ describe("workspaces schema", function()
       end
     end)
 
+    it("rejects reserved names", function()
+      local core_entities = require "kong.constants".CORE_ENTITIES
+      for i = 1, #core_entities do
+        local ok, err = Workspaces:validate({
+          name = core_entities[i],
+          config = {},
+          meta = {},
+        })
+        assert.falsy(ok)
+        assert.matches("must not be one of: workspaces, consumers, certificates, services, routes, snis, upstreams, targets, plugins, tags, ca_certificates, clustering_data_planes, parameters", err.name)
+      end
+    end)
+
     -- acceptance
     it("accepts valid names", function()
       local valid_names = {


kong/db/schema/typedefs.lua
```diff
@@ -376,6 +391,11 @@ typedefs.utf8_name = Schema.define {
   custom_validator = validate_utf8_name
 }
 
+typedefs.workspace_name = Schema.define {
+  type = "string",
+  unique = true,
+  custom_validator = validate_workspace_name
```
dndx, line 397:
```suggestion
  custom_validator = validate_workspace_name,
  not_one_of = constants.CORE_ENTITIES,
```

Should be sufficient. We don't have to re-implement all the logic in the `custom_validator`, they should both be executed with `not_one_of` checked first and `custom_validator` checked next.

dndx (changes requested):

kong/db/schema/typedefs.lua
gszr: Good idea, @dndx, but it won't work as is, since `CORE_ENTITIES` behaves as an array and dictionary. What about the following?
```
diff --git a/kong/db/schema/entities/workspaces.lua b/kong/db/schema/entities/workspaces.lua
index 1709dc77..21c6b062 100644
--- a/kong/db/schema/entities/workspaces.lua
+++ b/kong/db/schema/entities/workspaces.lua
@@ -1,4 +1,5 @@
 local typedefs = require "kong.db.schema.typedefs"
+local constants = require "kong.constants"
 
 
 return {
@@ -10,7 +11,7 @@ return {
 
   fields = {
     { id          = typedefs.uuid },
-    { name        = typedefs.utf8_name { required = true } },
+    { name        = typedefs.utf8_name { required = true, not_one_of = { table.unpack(constants.CORE_ENTITIES) }, } },
     { comment     = { type = "string" } },
     { created_at  = typedefs.auto_timestamp_s },
     { meta        = { type = "record", fields = {} } },
```

This preserves the original typedef, only extending it with a validator.

jschmid1:
tests fail with: 

`kong/db/schema/entities/workspaces.lua:14:74: accessing undefined field unpack of global table`

while it passes locally.. Any idea?

kong/db/schema/typedefs.lua
gszr: > tests fail with:
> 
> `kong/db/schema/entities/workspaces.lua:14:74: accessing undefined field unpack of global table`
> 
> while it passes locally.. Any idea?

That's interesting. Mind trying `unpack` (global) instead?

gszr:
> tests fail with:
> 
> `kong/db/schema/entities/workspaces.lua:14:74: accessing undefined field unpack of global table`
> 
> while it passes locally.. Any idea?

That's interesting. Mind trying unpack (global) instead?

jschmid1:
mhh... see https://github.com/Kong/kong/pull/7380/commits/f608b46249b6f689365bbb86f6d7ebdf68fb4a41

same result -> `    kong/db/schema/entities/workspaces.lua:3:16: accessing undefined field unpack of global table
`

gszr:
> mhh... see [f608b46](https://github.com/Kong/kong/commit/f608b46249b6f689365bbb86f6d7ebdf68fb4a41)
> 
> same result -> ` kong/db/schema/entities/workspaces.lua:3:16: accessing undefined field unpack of global table`

Oh, that's coming from the linter, I missed that 🤦 This will solve it:
```
diff --git a/.luacheckrc b/.luacheckrc
index ad2041aa..00cae9e8 100644
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -44,6 +44,13 @@ files["kong/hooks.lua"] = {
 }
 
 
+files["kong/db/schema/entities/workspaces.lua"] = {
+    read_globals = {
+        "table.unpack",
+    }
+}
+
+
 files["kong/plugins/ldap-auth/*.lua"] = {
     read_globals = {
         "bit.mod",
```

And it's safe to use it as before (`table.unpack`).

fffonion:
Random thought: should we also exclude those names of custom DAO or custom API endpoints?

jschmid1:
> Oh, that's coming from the linter, I missed that 🤦 This will solve it:

thanks! it's passing now

gszr (approved):

gszr:
(rebased onto latest mater)

dndx (approved):

gszr (approved):

kikito:
Ignoring this in the Changelog as it's EE-related.

PR #7399, opened 2021-05-27
chore(actions) remove use of KBT based build and caching for CI runs

dndx:
Couple of test cases needs to be fixed in order to run correctly. This is because KBT spawns all the dependencies when running even in cases those dependencies are not needed. For example, even in C* and DB-less test mode, Postgres servers are still started. In Actions we only spawns the external dependencies needed for each test scenario, thus exposing those issues.

### DIFF
--- a/.ci/run_tests.sh
+++ b/.ci/run_tests.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -ex
+set -e
 
 function cyan() {
     echo -e "\033[1;36m$*\033[0m"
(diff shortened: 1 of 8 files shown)


gszr:
> Couple of test cases needs to be fixed in order to run correctly. 

Which test cases are those?

Otherwise, LGTM! 💯

gszr (approved):

dndx:
@gszr They are in the diff, only 5 cases, and all very minor changes like retagging.

hutchic:
not sure about this direction. It's slower, added ~50 lines of bash, and makes the test less able to easily run / reproduce locally.

General guidance has been to be as CI agnostic as possible (so CI failures can be reproduced) and keep our build tooling as equivalent to our test tooling as possible (so we don't release broken builds). For the moment this is probably fine but I suspect if anything grows in complexity it will not be the case -- I see one way that this would cause challenges for EE already for example.

I've always been curious if gojira could be better utilized in our CI

gszr:
> not sure about this direction. It's slower, added ~50 lines of bash, and makes the test less able to easily run / reproduce locally.

My impression is that it's ~ 30% faster. Additionally, it's IMO a lot easier to debug CI issues this way. The n docker caching layers in kbt made it very difficult to debug those before.

> General guidance has been to be as CI agnostic as possible (so CI failures can be reproduced) and keep our build tooling as equivalent to our test tooling as possible (so we don't release broken builds). For the moment this is probably fine but I suspect if anything grows in complexity it will not be the case -- I see one way that this would cause challenges for EE already for example.

IMO we should keep kbt as simple as possible -- avoiding to use it for more than _building packages_ (the common underlying openresty-build-tools is shared in both places). 

>  I see one way that this would cause challenges for EE already for example.

Could you elaborate?

> I've always been curious if gojira could be better utilized in our CI

I won't speak to others here, but my general impression is that the team is more comfortable with _less_ dockerization of things.

dndx:
For the speedup of switching to GitHub Actions based caching, I think the difference is pretty obvious:

https://github.com/Kong/kong/actions/runs/885715636
https://github.com/Kong/kong/actions/runs/882576568

Because the GitHub Actions cache size is only ~25MB and takes 3 seconds to download and uncompress, versus the multiple GB of file that has to be pulled from Docker Hub.

hutchic:
image

image

hutchic:
This PR surfaced again during a `git blame` :disappointed:

PR #7411, opened 2021-06-02
fix(perf) pin openresty version in perf test

fffonion:


.github/workflows/perf.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 env:
   terraform_version: '0.15.3'
+  openresty_version: '1.19.3.1'
```
gszr, line 12: Is there a way for us to pick it from `.requirements` instead?

fffonion: added and verified!

gszr (approved):

PR #7413, opened 2021-06-02
docs(openapi-gen) Kong Admin API definition generator

locao:
This PR includes:
- The first version of the Kong Admin API definition generator.
- The actual Kong Admin API definition.
- Updates to the release scripts to run the definition generator and update the kong-admin-api.yml file before release.
- An update to the CI script to run the definition generator and check if it's still working.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -100,6 +100,11 @@ jobs:
           eval `luarocks path`
           scripts/autodoc
 
+    - name: Check Admin API definition generation
+      run: |
+          eval `luarocks path`
+          scripts/gen-admin-api-def.sh
+
     - name: Lint Lua code
       run: |
           eval `luarocks path`
--- a/autodoc/admin-api/openapi-gen.lua
+++ b/autodoc/admin-api/openapi-gen.lua
@@ -1,0 +1,440 @@
+local admin_api_data = require "autodoc.admin-api.data.admin-api"
+local kong_meta = require "kong.meta"
+local lfs = require "lfs"
+local lyaml = require "lyaml"
+local typedefs = require "kong.db.schema.typedefs"
+
+local OPENAPI_VERSION = "3.1.0"
+local KONG_CONTACT_NAME = "Kong"
+local KONG_CONTACT_URL = "https://github.com/Kong/kong"
+local LICENSE_NAME = "Apache 2.0"
+local LICENSE_URL = "https://github.com/Kong/kong/blob/master/LICENSE"
+
+local METHOD_NA_DBLESS = "This method is not available when using DB-less mode."
+local METHOD_ONLY_DBLESS = "This method is only available when using DB-less mode."
+
+local HTTP_METHODS = {
+  ["GET"] = true,
+  ["HEAD"] = true,
+  ["POST"] = true,
+  ["PUT"] = true,
+  ["DELETE"] = true,
+  ["CONNECT"] = true,
+  ["OPTIONS"] = true,
+  ["TRACE"] = true,
+  ["PATCH"] = true,
+}
+
+local entities_path = "kong/db/schema/entities"
+local routes_path = "kong/api/routes"
+
+-- workaround to load module files
+_KONG = require("kong.meta")          -- luacheck: ignore
+kong = require("kong.global").new()   -- luacheck: ignore
+kong.configuration = {                -- luacheck: ignore
+  loaded_plugins = {},
+}
+kong.db = require("kong.db").new({    -- luacheck: ignore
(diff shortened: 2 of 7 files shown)


mheap:
Initial thoughts:

- Could we have two specs? One for standard use and one for dbless?
- I think summaryhas been used where it should be description in a lot of places
- There are no schemas defined
  - Response schema
  - Request schemas for `POST`, `PUT` and `PATCH`
- Can we use tags to put the operations into logical groups?

image

Rendered using redoc https://github.com/Redocly/redoc

locao:
> Initial thoughts:
> 
>   * Could we have two specs? One for standard use and one for dbless?

We had some discussion about having one or two specs. I will quote @hbagdi on this, I think he nailed the issue:

> 1. An API not being available due to a configuration shouldn't result in a different spec.
> 1. We don't want to introduce more divergence between different modes. We want to consolidate and not further increase complexity.
> 1. The entire read API is available in DB-less mode last time I looked, that would result in unnecessary duplication.
> 1. The only time two specs make sense is when there are two different APIs.


>   * I think summaryhas been used where it should be description in a lot of places
> 

This specification was used as reference: https://swagger.io/docs/specification/paths-and-operations/. It seems to me that we still don't have long descriptions to use, but I don't know what would the parameter to call them `description` or `summary` (multiple lines, number of characters, markdown use?).

>   * There are no schemas defined
>       
>     * Response schema
>     * Request schemas for `POST`, `PUT` and `PATCH`

The request/response schemas can't be easily parsed at the moment, but it seems this definition with no schemas would be an improvement over not having a definition at all. We are trying to figure out a way to add and keep those schemas always validated, so they are guaranteed to be in sync with what Kong actually does.

>   * Can we use tags to put the operations into logical groups?

Adding tags is possible, but it would be interesting to get users' input on what would be useful for them and then decide where to add the tags or what field could be reused for that.

mheap:
> An API not being available due to a configuration shouldn't result in a different spec

If it's one or two items I would agree, but if we think about the experience for the consumer there are two very different use cases. Either they're running in DB-less or not, and the cognitive overhead of thinking about that for every endpoint rather than once when they choose a spec is huge

> We don't want to introduce more divergence between different modes. We want to consolidate and not further increase complexity.

Agreed

> The entire read API is available in DB-less mode last time I looked, that would result in unnecessary duplication.

We can handle this by defining a schema component and using a ref in both specifications to a single source of truth

> It seems to me that we still don't have long descriptions to use, but I don't know what would the parameter to call them description or summary (multiple lines, number of characters, markdown use?).

I'd recommend installing `redoc` with `npm install -g redoc-cli` then running `redoc-cli serve kong-admin-api.yml`. This is a good approximation of how consumers will use the spec. Anywhere that it says _This method is not available when using DB-less mode._ I would move to the `description` field and use `summary` to describe what the endpoint does e.g. _Delete entry from cache_

> The request/response schemas can't be easily parsed at the moment, but it seems this definition with no schemas would be an improvement over not having a definition at all. We are trying to figure out a way to add and keep those schemas always validated, so they are guaranteed to be in sync with what Kong actually does.

Do the endpoints have good validation that guides the user to provide any missing parameters? If so I agree that knowing the endpoints is a good start. If not, we should hold off as providing URIs but no schema will lead to consumer frustration.

There is also a much bigger discussion that we should have outside of this thread. I'd like to talk through using the specification as the source of truth and using contract testing to validate the code. I've seen codegen specifications attempted in a few different companies now and they never work out. Who should be involved in a discussion about moving to design first API development with OAS?

> Adding tags is possible, but it would be interesting to get users' input on what would be useful for them and then decide where to add the tags or what field could be reused for that.

We could give it a go for the first iteration. We can change tags in the future without it being a BC break.

kikito:
> If it's one or two items I would agree, but if we think about the experience for the consumer there are two very different use cases

What constitutes "confusing" is subjective. We currently have two separate docs https://docs.konghq.com/ - one for the [db-less Admin API](https://docs.konghq.com/gateway-oss/2.4.x/db-less-admin-api/) and another one for the [db Admin API](https://docs.konghq.com/gateway-oss/2.4.x/admin-api/).

However, the docs team has requested us to change the auto-generation docs so that a single doc is generated instead, because it is "less confusing with a single file" [¹](#a1).

On the confusion side of things, I am overwhelmed and confused about everything and both options equally dazzle me. But on the interest of standardization alone, I would suggest going with a single file instead of two.

<span id="1">¹</span> Relevant pr is #7467 . Not merged yet, but I expect it to eventually get merged.

kikito (approved):
I have approved this, but I have a proposal in #7480

@locao , please give that a look before merging this.

PR #7421, opened 2021-06-04
perf(*) reduce ngx.var usage

fffonion:
### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -38,7 +38,6 @@ local subsystem    = ngx.config.subsystem
 local clear_header = ngx.req.clear_header
 local unpack       = unpack
 local escape       = require("kong.tools.uri").escape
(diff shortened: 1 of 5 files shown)


kong/runloop/handler.lua
```diff
@@ -1113,11 +1113,6 @@ return {
     before = function(ctx)
       local server_port = var.server_port
       ctx.host_port = HOST_PORTS[server_port] or server_port
-
```
fffonion, line 1116: removed as it seems not being read afterwards

bungle: that was because of a bug

guanlan:
@fffonion is this still relevant?

fffonion:
Closing as a different approach has replaced this.

PR #7436, opened 2021-06-09
fix(perf) better nginx upstream conf, use correct stackcollapse script

fffonion:


spec/helpers/perf/drivers/terraform.lua
```diff
@@ -166,14 +166,31 @@ end
...
-                return 200;
-              }
-              %s
-            }]]):format(UPSTREAM_PORT, conf)):gsub("\n", "")
+  conf = ngx.encode_base64(([[
+  worker_processes auto;
+  worker_cpu_affinity auto;
+  error_log /var/log/nginx/error.log;
+  pid /run/nginx.pid;
+  worker_rlimit_nofile 20480;
```
gszr, line 174: Out of curiosity: are these hardcoded values determined empirically? (Understandably, they have to be set to a "high enough" value to not limit the benchmarks, but, in this case, how did you get to those?)

spec/helpers/perf/drivers/local.lua
```diff
@@ -230,7 +230,7 @@ function _M:generate_flamegraph(title)
 
   local cmds = {
     "/tmp/perf-ost/fix-lua-bt " .. path .. ".bt > " .. path .. ".fbt",
-    "/tmp/perf-fg/stackcollapse.pl " .. path .. ".fbt > " .. path .. ".cbt",
+    "/tmp/perf-fg/stackcollapse-stap.pl " .. path .. ".fbt > " .. path .. ".cbt",
```
gszr, line 233: nitpick: I see some of those commands are repeated in the `terraform` module; would it make sense to factor them into a common module - say, `drivers/constants.lua`?

gszr (approved):

dndx (approved):

spec/helpers/perf/drivers/terraform.lua
fffonion: I duplicated it from a performance test done by a third party with us and a competitor. I didn't really validate it with comparasion but from a first view I feel it's sane.

spec/helpers/perf/drivers/local.lua
fffonion: Yeah that's good point, terraform has lots of similarity with local driver. I think we can make them share some operations in the future.

PR #7443, opened 2021-06-10
feat(clustering) store and expose DP plugin versions in admin API

fffonion:
### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX


kong/db/schema/entities/clustering_data_planes.lua
```diff
@@ -30,5 +30,14 @@ return {
                       default = "unknown",
                     }
     },
+    { plugin_versions = { type = "array",
+                          elements = { type = "record",
```
bungle, line 34: As this is an array of records, perhaps just call it `plugins`. Perhaps later on it starts to contain more than just `version` information. And similar adjustments to migrations.

bungle (changes requested):

kong/db/schema/entities/clustering_data_planes.lua
fffonion: In that case should we make this a record? So plugins.versions is then array of {name, version}.

fffonion:
As discussed offline this will not be prioritized for now so I'm removing the 2.5 milestone and keep this PR
opened for future.

kong/db/schema/entities/clustering_data_planes.lua
kikito: I think what Aapo is saying is leaving everything as it is, but just naming it `plugins`.

The elements of your current `plugin_versions` contain two fields: `name` and `version`. What he's saying is that if in the future we want to extend that with other things that are plugin versions (for example to include `kong_version_min` or `kong_version_max`) it would work.

I don't think we should make this `plugins.versions` an array and then have other arrays named `plugins.kong_version_mins`.

(If we were programming in Pascal though, that would be our only option :) )

PR #7445, opened 2021-06-10
feat(plugins) import prometheus plugin

gszr:
As a result of an effort to improve external plugins maintainership,
we are importing currently-bundled plugins into Kong's main tree.
This plugin used to live in https://github.com/Kong/kong-plugin-prometheus.

In addition to import the plugin contents, this also preserves the git history.


mikefero:
Should `kong-plugin-prometheus` be removed from the rockspec as well as adding the Prometheus plugin files?

```diff
diff --git a/kong-2.4.1-0.rockspec b/kong-2.4.1-0.rockspec
index 6b6b1125..0930ea27 100644
--- a/kong-2.4.1-0.rockspec
+++ b/kong-2.4.1-0.rockspec
@@ -43,7 +43,6 @@ dependencies = {
   "kong-plugin-azure-functions ~> 1.0",
   "kong-plugin-zipkin ~> 1.3",
   "kong-plugin-serverless-functions ~> 2.1",
-  "kong-prometheus-plugin ~> 1.3",
   "kong-proxy-cache-plugin ~> 1.3",
   "kong-plugin-request-transformer ~> 1.3",
   "kong-plugin-session ~> 2.4",
@@ -384,5 +383,14 @@ build = {
 
     ["kong.plugins.request-termination.handler"] = "kong/plugins/request-termination/handler.lua",
     ["kong.plugins.request-termination.schema"] = "kong/plugins/request-termination/schema.lua",
+
+    ["kong.plugins.prometheus.api"] = "kong/plugins/prometheus/api.lua",
+    ["kong.plugins.prometheus.status_api"] = "kong/plugins/prometheus/status_api.lua",
+    ["kong.plugins.prometheus.exporter"] = "kong/plugins/prometheus/exporter.lua",
+    ["kong.plugins.prometheus.enterprise.exporter"] = "kong/plugins/prometheus/enterprise/exporter.lua",
+    ["kong.plugins.prometheus.handler"] = "kong/plugins/prometheus/handler.lua",
+    ["kong.plugins.prometheus.prometheus"] = "kong/plugins/prometheus/prometheus.lua",
+    ["kong.plugins.prometheus.serve"] = "kong/plugins/prometheus/serve.lua",
+    ["kong.plugins.prometheus.schema"] = "kong/plugins/prometheus/schema.lua",
   }
 }
```

If I remember reading correctly the prometheus plugin rockspec is going to stick around for awhile in the tree so maybe there is another approach during the build phase where that is going to be used instead; so the above is possibly mute.

spec/fixtures/prometheus/custom_nginx.template
```diff
@@ -0,0 +1,885 @@
+# This is a custom nginx configuration template for Kong specs
```
fffonion, line 1: with the plugin moved to kong repo, we can remove this custom template as it's only
needed to let prometheus test return static number of shm metrics.
it would be better to always used what kong ships for avoiding suprises

kong/plugins/prometheus/enterprise/exporter.lua
```diff
@@ -0,0 +1,94 @@
+local kong = kong
```
fffonion, line 1: question: do we plan to port this PR to EE as well? if so we can remove those EE code from here. and add them back in the EE sister PR.

kikito:
Which git commands did you use to "import the git history" like that? Is that something we can reuse for the other plugins?

gszr:
> Which git commands did you use to "import the git history" like that?

To preserve the history:
```
git remote add prometheus-plugin git@github.com:Kong/kong-plugin-prometheus.git
git merge prometheus-plugin/master --allow-unrelated-histories
```

Still, I think I am going to rewrite commit messages to include "prometheus" in the scope, otherwise the history will end up being quite confusing.

>  Is that something we can reuse for the other plugins?

Definitely. I will do the same for others.

(Right now I am blocked in getting tests to pass; some tests are getting stuck, while others break on the write to the stream api socket.)

gszr:
Will reopen.

PR #7454, opened 2021-06-15
perf(postgres): meta_schema is a read-only operation

jschmid1:
### Summary

All operations that qualify as a non-critical "read" operation are correctly hitting the postgres read-only replica. With the exception of the call to the `schema_meta` table.

### Full changelog

* Add the `read` operation keyword to route the query to a potential read-only replica

### DIFF
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -717,7 +717,7 @@ function _mt:schema_migrations()
     "SELECT *\n",
     "  FROM schema_meta\n",
     " WHERE key = ",  self:escape_literal("schema_meta"), ";"
-  }))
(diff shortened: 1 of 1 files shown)


bungle:
Perhaps we originally wanted to be sure that we always get data from the write node when checking migrations. But I know, it can mean some requests to be hit on write node on node startup that could perhaps be avoided. I have not strong feeling about it. @kikito what do you think?

kikito (approved):
To me it looks like a simple error of omission. If the replication works correctly, once a migration is done the meta_schema field should be available in the read replica. And the replication does not work correctly, then we have other problems :).

I think we should merge this one.

I would not make it a `fix`, though; I would call it a `perf`.

PR #7455, opened 2021-06-15
chore(deps) bump zipkin plugin

kikito:
New minor version because 1 new minor feature was added


gszr (approved):

kikito:
Merging despite the failing tests because they are all related with grpc, which has nothing to do with the Zipkin plugin.

PR #7458, opened 2021-06-16
fix(hybrid) ignore fileds that values are null

fffonion:
This PR was made for the bringing forward compatibilty in DataPlane in hybrid mode.
When a new field is added in ControlPlane but the value is null (usually means it's unset), DP will ignore such
field from schema validation.

Only DataPlane will receive this behaviour. Extra fields from admin API and declarative config are still returning
errors.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1085,6 +1085,11 @@ validate_fields = function(self, input)
       field, err = resolve_field(self, k, field, subschema)
       if field then
         _, errors[k] = self:validate_field(field, v)
+      elseif err == validation_errors.UNKNOWN and v == null and
+            kong and kong.configuration and
+            kong.configuration.role == "data_plane" then -- luacheck: ignore
+        -- extra fields with value of null in the input config are ignored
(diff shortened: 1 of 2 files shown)


kong/db/schema/init.lua
```diff
@@ -1085,7 +1085,9 @@ validate_fields = function(self, input)
       field, err = resolve_field(self, k, field, subschema)
       if field then
         _, errors[k] = self:validate_field(field, v)
-      else
+      elseif not (err == validation_errors.UNKNOWN and v == null) then
+        -- extra fields with value of null in the input config are ignored
+        -- otherwise record the error
```
bungle, line 1090: This should only happen when validating `declarative` configuration OR even stricter scope: when validating declarative config that control plane send to data plane over websocket.

Alternative way to implement this is to implement somekind of `remove_unknown_null_fields` BEFORE validating, then you don't need to change validation rules.

bungle (changes requested):

bungle (approved):

PR #7462, opened 2021-06-16
Hybrid mode 2.0 protocol

dndx:
This PR contains the Hybrid mode 2.0 messaging system and RPC implementation.

It is not enabled in 2.5, however, it allows future development work on this area after 2.5.


kong/hybrid/control_plane.lua
```diff
@@ -0,0 +1,190 @@
...
+    end
+  end
+
+  ngx_header["Upgrade"] = "Kong-Hybrid/2"
+  ngx_header["Content-Type"] = nil
+  ngx.status = 101
+
+  local ok, err = ngx.send_headers()
+  if not ok then
+    ngx_log(ngx_ERR, "failed to send response header: " .. (err or "unknonw"))
```
fffonion, line 149:
```suggestion
    ngx_log(ngx_ERR, "failed to send response header: " .. (err or "unknown"))
```

kong/hybrid/data_plane.lua
```diff
@@ -0,0 +1,145 @@
...
+end
+
+
+function _M:start_timer(delay)
+  if not delay then
+    delay = math.random(5, 10)
+  end
+
+  if delay > 0 then
+    ngx_log(ngx_WARN, "reconnecting to control plane in ", delay, " seconds")
```
fffonion, line 46: let's prefix those error logs with a namespace like `[hybrid]` or something

kong/hybrid/message.lua
```diff
@@ -0,0 +1,121 @@
...
+
+
+function _M.new(src, dest, topic, message)
+  local self = tablepool.fetch(POOL_NAME, 0, 4)
+
+  assert(dest, "dest is required")
+  assert(topic, "topic is required")
+  assert(message, "message is required")
+
+  self.src = src
```
fffonion, line 49: since src, dest, and topic's length fit in a char, we should check their length
are shorter than 256bytes

fffonion (approved):

PR #7463, opened 2021-06-16
Configure DataDog host and port from env variables

rallyben:
This feature is necessary for Kubernetes installations that run the
DataDog agent as a Daemon-Set.  In particular, the DataDog host must be
set to the primary IP of the Kubernetes worker.  This value is derived
from "status.hostIP" with is not available in a Custom Resource
Definition (CRD)

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX


kong/plugins/datadog/handler.lua
```diff
@@ -41,6 +43,14 @@ local function log(premature, conf, message)
     return
   end
 
+  if env_datadog_agent_host then
+    conf.host = env_datadog_agent_host
```
Tieske, line 47: the `conf` table is read-only by convention. So this is not ok.

Tieske (changes requested):
the current implementation is overriding all plugin instances. I think the proper way would be to add a flag to the schema to use the env variables, which would be a per instance setting. Setting that to true should also make the host and port fields to be no longer "required".

rallyben:
> the current implementation is overriding all plugin instances. I think the proper way would be to add a flag to the schema to use the env variables, which would be a per instance setting. Setting that to true should also make the host and port fields to be no longer "required".
> 
> EDIT: also it is lacking tests

Thanks for the feedback, I'm pretty new to kong/lua.  I believe your suggestion would mean that the user could add an optional flag during the installation of the `datadog` plugin via the Admin API that would signal to use the environment variables to define the host/port.  That would be added here?
https://github.com/Kong/kong/blob/master/kong/plugins/datadog/schema.lua#L76-L77

rallyben:
> exactly!

Can you point me at an example that does something like this? I'm pretty new to this, so that would be a big help.  Thanks!

Tieske:
add a flag setting; and then make the 2 existing properties "required = false" when that flag is set.

Something like this one: https://github.com/Kong/kong/blob/master/kong/db/schema/entities/upstreams.lua#L199-L202

pguedes:
i need this... :pray:

rallyben:
@Tieske I redid my approach, incorporating your suggestion.  Thanks!

I still don't use tests but I have verified functionality.  Before I write the tests, can you confirm I'm on the right path?

kong/plugins/datadog/statsd_logger.lua
```diff
@@ -19,10 +19,16 @@ local stat_types = {
 local statsd_mt = {}
 statsd_mt.__index = statsd_mt
 
+local env_datadog_agent_host = os.getenv 'KONG_DATADOG_AGENT_HOST'
+local env_datadog_agent_port = os.getenv 'KONG_DATADOG_AGENT_PORT'
```
Tieske, line 23:
```suggestion
local env_datadog_agent_port = tonumber(os.getenv 'KONG_DATADOG_AGENT_PORT')
```

kong/plugins/datadog/statsd_logger.lua
```diff
@@ -19,10 +19,16 @@ local stat_types = {
...
 
 function statsd_mt:new(conf)
   local sock   = udp()
-  local _, err = sock:setpeername(conf.host, conf.port)
+  if conf.use_env then
+    local _, err = sock:setpeername(env_datadog_agent_host, env_datadog_agent_port)
```
Tieske, line 28: since `err` is declared as a local here, it will always be `nil` when checking at line 32. So the `local` declaration should go before the `if/then`, similar to the local `sock`.

Tieske: An error needs to be returned if the env vars haven't been set;
```suggestion
    if not (env_datadog_agent_host or env_datadog_agent_port) then
      return nil, "both KONG_DATADOG_AGENT_HOST and KONG_DATADOG_AGENT_PORT must be set"
    end
    _, err = sock:setpeername(env_datadog_agent_host, env_datadog_agent_port)
```

Tieske (changes requested):
@rallyben nice!

Some easy comments to fix. And this will require some additional tests, both on the schema as well as an integration test

kong/plugins/datadog/statsd_logger.lua
rallyben: I think I should wait to convert the type until after confirming the environment variable is set.  I'm not sure what happens if I called `tonumber` on `nil`

kong/plugins/datadog/statsd_logger.lua
rallyben: I restructured the control-flow.  I'm pretty happy with it now, let me know what you think.  I'm starting the write tests now.

kong/plugins/datadog/statsd_logger.lua
Tieske: you're right, it should be `tonumber(os.getenv('KONG_DATADOG_AGENT_PORT') or "")`

kong/plugins/datadog/statsd_logger.lua
```diff
@@ -19,18 +19,33 @@ local stat_types = {
...
   local sock   = udp()
-  local _, err = sock:setpeername(conf.host, conf.port)
+
+  local host, port
+  if conf.use_env then
+    if not (env_datadog_agent_host and env_datadog_agent_port) then
+      return nil, "both KONG_DATADOG_AGENT_HOST and KONG_DATADOG_AGENT_PORT must be set"
+    end
+    host = env_datadog_agent_host
+    port = tonumber(env_datadog_agent_port)
```
Tieske, line 34:
```suggestion
    port = env_datadog_agent_port
```
if you convert here, then if the port was a string `"xxx"` for example, then by now it would be `nil`, and fail later on.

Tieske (changes requested):
so far looking good, just move the conversion.

kong/plugins/datadog/statsd_logger.lua
rallyben: done and done, thanks. Moving on the tests

javierguerragiraldez (commented):
looks good.   would approve as soon as it gets some tests

Tieske:
@javierguerragiraldez would you mind having a look at https://github.com/Kong/kong/pull/7556 ? without that one merged I don't think it's fair to ask @rallyben to write tests for this PR.

Tieske:
@rallyben that PR has been merged, so if you rebase this PR on `master` you should get all the statsd tests in place

rallyben:
@Tieske @javierguerragiraldez added tests.  The defaults on host and port in the schema limit my ability to `nil` these configurations.

As to documentation, is there a separate repo for that?

kong/plugins/datadog/schema.lua
```diff
@@ -95,7 +96,17 @@ return {
                     if_match = { one_of = { "counter", "gauge" }, },
                     then_field = "sample_rate",
                     then_match = { required = true },
-                  }, },
-  }, }, }, }, }, }, }, },
+          }, }, }, }, }, },
+        },
+        entity_checks = {
+          { conditional = {
+            if_field = "use_env", if_match = {eq = false},
```
Tieske, line 103: @javierguerragiraldez I don't like the name `use_env` but also don't know of a better alternative, thoughts?

spec/03-plugins/08-datadog/01-log_spec.lua
```diff
@@ -122,6 +127,21 @@ for _, strategy in helpers.each_strategy() do
         },
       }
 
+      bp.plugins:insert {
+        name     = "key-auth",
+        route = { id = route5.id },
+      }
+
+      helpers.setenv('KONG_DATADOG_AGENT_HOST', 'localhost')
+      helpers.setenv('KONG_DATADOG_AGENT_PORT', '9999')
```
Tieske, line 136: These should be "unset" in the `teardown`

spec/03-plugins/08-datadog/02-schema_spec.lua
```diff
@@ -8,6 +8,12 @@ describe("Plugin: datadog (schema)", function()
     assert.is_nil(err)
     assert.is_truthy(ok)
   end)
+  it("accepts nil host and port with use_env", function()
+    -- Defaults are set for host and port, as we're not really testing for undefined values
```
Tieske, line 12: you can force them to `null` probably by setting them to `ngx.null` instead of `nil`

Tieske (changes requested):

kong/plugins/datadog/schema.lua
rallyben: The use case here is defining the host and port when DataDog is run as a `daemonset` under Kubernetes.  In theory, this configuration could be applied more broadly though

Tieske (approved):

Tieske:
@rallyben docs are here: https://github.com/Kong/docs.konghq.com/tree/main/app/_hub/kong-inc/statsd

javierguerragiraldez (approved):

rallyben:
got some test failures but they don't seem related to my PR. Rebased off `master`

Tieske:
@javierguerragiraldez lgtm, but I have no git powers here, so I cannot merge.

Tieske:
@rallyben thx for your contribution 👍

PR #7467, opened 2021-06-17
docs(autodoc) collapse declarative_admin_api.md into admin_api.md

kikito:
Before this PR, autodoc generated two separate (but similar) files for the admin api:
* `admin-api.md` for the "db" Admin API
* `db-less-admin-api.md` for db-less

After this PR,
* only `admin-api.md` will remain
* It will have two new top sections describing db-less & declarative config
* methods that are not compatible with db-less will be marked as such with a warn blue box

---

**Note for reviewers**: I have moved this back to "please review" despite @javierguerragiraldez 's approval because feedback from the docs team resulted in two new changes which were added after his review:

* We now mark both db and db-less methods with badges, with a custom syntax
* We patched the PDK navigation in one way and the Admin API navigation in a totally different way. This second way required an extra lua file to be run in the docs repo. I have simplified the steps and now both files are patched into the nav file with the method for used for the PDK.
* We modified `write_title` to include an optional badge

If you are short on time when reviewing this, you may just concentrate on those changes.

### DIFF
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -43,15 +43,7 @@ return {
   intro = {
     {
       text = [[
-        <div class="alert alert-info.blue" role="alert">
-          This page refers to the Admin API for running Kong configured with a
-          database (Postgres or Cassandra). For using the Admin API for Kong
-          in DB-less mode, please refer to the
-          <a href="/gateway-oss/{{page.kong_version}}/db-less-admin-api">Admin API for DB-less Mode</a>
-          page.
-        </div>
-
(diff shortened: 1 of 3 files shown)


javierguerragiraldez (approved):

kikito:
I have moved this back to "please review" despite @javierguerragiraldez 's approval because feedback from the docs team resulted in two new changes which were added after his review:

* We now mark both db and db-less methods with badges, with a custom syntax
* We patched the PDK navigation in one way and the Admin API navigation in a totally different way. This second way required an extra lua file to be run in the docs repo. I have simplified the steps and now both files are patched into the nav file with the method for used for the PDK.

PR #7477, opened 2021-06-21
fix(syslog) ensure new facility field is compatible with older versions in hybrid mode

fffonion:
The default value is now populated at runtime for compatibiliy.
This commit can be reverted in 2.7 (2 version started from 2.5) or 3.0 whichever comes eariler.

### DIFF
--- a/kong/plugins/syslog/handler.lua
+++ b/kong/plugins/syslog/handler.lua
@@ -71,14 +71,16 @@ local function log(premature, conf, message)
     return
   end
 
+  local facility = conf.facility or "user"
+
   if message.response.status >= 500 then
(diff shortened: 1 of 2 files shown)


kong/plugins/syslog/handler.lua
```diff
@@ -71,14 +71,16 @@ local function log(premature, conf, message)
     return
   end
 
+  local facility = conf.facility or "user"
```
Tieske, line 74: can we add a TODO comment that this default should be moved into the schema after Kong version x.x

Tieske (approved):
minor comment, lgtm otherwise

kikito (approved):

PR #7479, opened 2021-06-22
feat(clustering) lax version checks

bungle:
### Summary

Incompatibility checks on control plane were loosened. Previously data plane was considered incompatible when:

1. data plane kong major version is different to control plane major version
2. data plane minor version is greater than control plane minor version
3. data plane minor version is behind more than two minor versions of control plane
4. data plane missed plugin that control plane had (whether it was configured or not)
5. data plane has different major.minor version of the plugin that control plane had
   (whether it was configured or not)

This commit changes that. With this commit the data plane is considered incompatible when:

1. data plane kong major version is different to control plane major version
2. data plane misses _configured_ plugin
3. data plane plugin major version is different to _configured_ plugin major version on control plane

This also makes the logging on these configuration issues on control plane more uniform and contains
information about the data plane which caused the incompatibility.

It still misses features such as data plane reporting import errors back to control plane or dao checks,
but let's leave that to other future pr.

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -12,12 +12,12 @@ local utils = require("kong.tools.utils")
 local constants = require("kong.constants")
 local openssl_x509 = require("resty.openssl.x509")
 local string = string
-local assert = assert
 local setmetatable = setmetatable
 local type = type
-local math = math
 local pcall = pcall
 local pairs = pairs
+local ipairs = ipairs
+local tonumber = tonumber
 local tostring = tostring
 local ngx = ngx
 local ngx_log = ngx.log
@@ -31,15 +31,18 @@ local ngx_now = ngx.now
 local ngx_var = ngx.var
 local table_insert = table.insert
 local table_remove = table.remove
+local table_concat = table.concat
 local deflate_gzip = utils.deflate_gzip
 
 
 local KONG_VERSION = kong.version
-local ngx_ERR = ngx.ERR
 local ngx_DEBUG = ngx.DEBUG
-local ngx_WARN = ngx.WARN
+local ngx_INFO = ngx.INFO
 local ngx_NOTICE = ngx.NOTICE
+local ngx_WARN = ngx.WARN
+local ngx_ERR = ngx.ERR
 local ngx_OK = ngx.OK
+local ngx_CLOSE = ngx.HTTP_CLOSE
 local MAX_PAYLOAD = constants.CLUSTERING_MAX_PAYLOAD
 local WS_OPTS = {
   timeout = constants.CLUSTERING_TIMEOUT,
@@ -49,6 +52,51 @@ local PING_INTERVAL = constants.CLUSTERING_PING_INTERVAL
 local PING_WAIT = PING_INTERVAL * 1.5
 local OCSP_TIMEOUT = constants.CLUSTERING_OCSP_TIMEOUT
 local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
+local MAJOR_MINOR_PATTERN = "^(%d+)%.(%d+)%.%d+"
+
+
+local function log(level, ...)
+  ngx_log(level, "[clustering] ", ...)
+end
+
+
+local function extract_major_minor(version)
+  if type(version) ~= "string" then
+    return nil, nil
+  end
+
+  local major, minor = version:match(MAJOR_MINOR_PATTERN)
+  if not major then
+    return nil, nil
+  end
+
+  major = tonumber(major, 10)
+  minor = tonumber(minor, 10)
+
+  return major, minor
+end
+
+
+local function plugins_list_to_map(plugins_list)
+  local versions = {}
+  for _, plugin in ipairs(plugins_list) do
+    local name = plugin.name
+    local version = plugin.version
+    local major, minor = extract_major_minor(plugin.version)
+
+    if major and minor then
+      versions[name] = {
+        major   = major,
+        minor   = minor,
+        version = version,
+      }
+
+    else
+      versions[name] = {}
+    end
+  end
+  return versions
+end
 
 
 local function is_timeout(err)
@@ -75,6 +123,13 @@ function _M:export_deflated_reconfigure_payload()
     return nil, err
   end
 
+  self.plugins_configured = {}
+  if config_table.plugins then
+    for _, plugin in pairs(config_table.plugins) do
+      self.plugins_configured[plugin.name] = true
+    end
+  end
+
   local payload, err = cjson_encode({
     type = "reconfigure",
     timestamp = ngx_now(),
@@ -98,7 +153,7 @@ end
 function _M:push_config()
   local payload, err = self:export_deflated_reconfigure_payload()
   if not payload then
-    ngx_log(ngx_ERR, "unable to export config from database: " .. err)
+    log(ngx_ERR, "unable to export config from database: " .. err)
     return
   end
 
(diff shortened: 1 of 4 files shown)


kong/clustering/control_plane.lua
```diff
@@ -196,126 +253,267 @@ do
...
 
-function _M:should_send_config_update(node_version, node_plugins)
-  if not node_version or not node_plugins then
-    return false, "your DP did not provide version information to the CP, " ..
-                  "Kong CP after 2.3 requires such information in order to " ..
-                  "ensure generated config is compatible with DPs. " ..
-                  "Sync is suspended for this DP and will resume " ..
-                  "automatically once this DP also upgrades to 2.3 or later"
+  if not major_cp then
+    return nil, "data plane version " .. dp_version .. " is incompatible with control plane version",
```
fffonion, line 261: the error message here and the one at line 266 seems incorrect?

kong/clustering/control_plane.lua
```diff
@@ -196,128 +253,271 @@ do
...
+  for _, plugin in ipairs(self.plugins_list) do
+    local name = plugin.name
+    local cp_plugin = self.plugins_map[name]
+    local dp_plugin = dp_plugin_map[name]
 
-    if p.name ~= np.name then
-      goto continue
+    if not dp_plugin then
+      if cp_plugin.version then
+        ngx_log(ngx_WARN, name, " plugin ", cp_plugin.version, " is missing from data plane", dp)
```
fffonion, line 305: generally i would suggest a prefix for those logs for better grokking, like `[clustering]` ?

fffonion: let's emit one type of error in case DP missed it, whether cp_plugin.version is set or not,
how you think? having multiple log
shape for similar outcome will probably makes harder to filter.

kong/clustering/control_plane.lua
```diff
@@ -196,128 +253,271 @@ do
...
-                  CLUSTERING_SYNC_STATUS.PLUGIN_SET_INCOMPATIBLE
+  if minor_cp ~= minor_dp then
+    local msg = "data plane minor version " .. dp_version ..
+                " is different to control plane minor version " ..
+                KONG_VERSION
+
+    minor_cp = tonumber(minor_cp, 10)
+    minor_dp = tonumber(minor_dp, 10)
+
+    local difference = math.abs(minor_cp - minor_dp)
```
fffonion, line 286: i would suggest to not add logic into the log level here, it's always hard to correctly
set a relation between version difference and levels based on different contexts.
maybe just a simple log with INFO level here will be fine.

kong/clustering/control_plane.lua
```diff
@@ -196,128 +253,271 @@ do
...
+    else
+      if cp_plugin.version and dp_plugin.version then
+        local msg = "data plane " .. name .. " plugin version " .. dp_plugin.version ..
+                    " is different to control plane plugin version " .. cp_plugin.version
+
+        if cp_plugin.major ~= dp_plugin.major then
+          ngx_log(ngx_WARN, msg, cp_plugin.version, dp)
+
+        else
+          local difference = math.abs(cp_plugin.minor - dp_plugin.minor)
```
fffonion, line 319: ditto

kong/clustering/control_plane.lua
```diff
@@ -196,128 +253,271 @@ do
...
+function _M:check_configuration_compatibility(dp_plugin_map)
+  for _, plugin in ipairs(self.plugins_list) do
+    if self.plugins_configured[plugin.name] then
+      local name = plugin.name
+      local cp_plugin = self.plugins_map[name]
+      local dp_plugin = dp_plugin_map[name]
+
+      if not dp_plugin then
+        if cp_plugin.version then
+          return nil, "configured " .. name .. " plugin " .. cp_plugin.version ..
```
fffonion, line 354: ditto, let's not make two different logs here. the version of CP plugins doesn't actually
make a difference to the compat check here

kong/clustering/control_plane.lua
```diff
@@ -196,128 +253,271 @@ do
...
-      goto continue
+    if not dp_plugin then
+      if cp_plugin.version then
+        ngx_log(ngx_WARN, name, " plugin ", cp_plugin.version, " is missing from data plane", dp)
+      else
+        ngx_log(ngx_WARN, name, " plugin is missing from data plane", dp)
+      end
+
+    else
+      if cp_plugin.version and dp_plugin.version then
```
fffonion, line 311: let's also group the else branches for `if cp_plugin.version and dp_plugin.version` together

fffonion (commented):
i think it's a good idea to add more logging visibility, however for those are "failing
the compat checks" in similar reason, we should group them in a same error message,
so that we will know what characters to filter in logs; if there're too many different
error messages, it will be harder to filter.

fffonion:
let's also update the e2e tests in `spec/02-integration/09-hybrid_mode/01-sync_spec.lua`

artomal:
Looks good. We also need to add tests for this change.

kong/clustering/control_plane.lua
bungle: Single cp always logs the same, the only difference in these lines is that other contains version while other does not.

kong/clustering/control_plane.lua
bungle: Grouped.

kong/clustering/control_plane.lua
bungle: The prefix is added. There is no difference in log statement especially on single cp or with same version of cps. It is just that other has version and other has not. They always do, so the other is just being defensive.

kong/clustering/control_plane.lua
bungle: Logic is removed.

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -325,91 +414,49 @@ for _, strategy in helpers.each_strategy() do
 
       -- STARTS blocked cases
       local blocked_cases = {
-        ["CP and DP major version mismatches"] = {
-          dp_version = "1.0.0",
-          expected = CLUSTERING_SYNC_STATUS.KONG_VERSION_INCOMPATIBLE,
+        ["CP configured plugin list missmatches DP enabled plugins list"] = {
```
fffonion, line 417:
```suggestion
        ["CP configured plugin list mismatches DP enabled plugins list"] = {
```

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -325,91 +414,49 @@ for _, strategy in helpers.each_strategy() do
...
+          dp_version = string.format("%d.%d.%d", MAJOR, MINOR, PATCH),
+          expected = CLUSTERING_SYNC_STATUS.PLUGIN_SET_INCOMPATIBLE,
+          plugins_list = {
+            {  name="banana-plugin", version="1.0.0" }
+          }
         },
-        ["CP.minor - DP.minor > 2"] = {
-          dp_version = string.format("%d.%d.%d", MAJOR, MINOR-3, PATCH),
-          expected = CLUSTERING_SYNC_STATUS.KONG_VERSION_INCOMPATIBLE,
+        ["CP configured plugin list missmatches DP enabled plugins list version"] = {
```
fffonion, line 424:
```suggestion
        ["CP configured plugin list mismatches DP enabled plugins list version"] = {
```

dndx:
Superseded by https://github.com/Kong/kong/pull/7488.

PR #7480, opened 2021-06-22
docs/openapi spec sorted yaml

kikito:
This PR = #7413 - the last commit, "force entries order in generated file" + a lyaml low-level dumper that generates sorted yamls instead + some other minor things.

The reason why the dumper is so much code is because lyaml doesn't expose its internal dumper, so the only way to do sorted yaml is copy-pasting the whole thing in order to edit one tiny part of the `dump_mapping` function (to do the key sort before iterating over the table). Related: https://github.com/gvvaughan/lyaml/issues/46

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -100,6 +100,11 @@ jobs:
           eval `luarocks path`
           scripts/autodoc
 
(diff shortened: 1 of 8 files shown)


locao:
We decided to not use this change.

PR #7483, opened 2021-06-22
fix(clustering) more reliable way to calculate data plane config hash

bungle:
### Summary

Previously data plane hash calculation was flaky and with recent OpenResty changes it become
even flakier. This PR tries to calculate data plane config hash in a more reliable way by
sorting the data and then calculating the hash out of it.

### DIFF
--- a/kong/clustering/data_plane.lua
+++ b/kong/clustering/data_plane.lua
@@ -9,14 +9,18 @@ local constants = require("kong.constants")
 local utils = require("kong.tools.utils")
 local system_constants = require("lua_system_constants")
 local ffi = require("ffi")
+local tablex = require("pl.tablex")
 local assert = assert
 local setmetatable = setmetatable
 local type = type
 local math = math
 local pcall = pcall
+local concat = table.concat
 local tostring = tostring
 local ngx = ngx
+local md5 = ngx.md5
 local ngx_log = ngx.log
+local ngx_null = ngx.null
 local ngx_sleep = ngx.sleep
 local cjson_decode = cjson.decode
 local cjson_encode = cjson.encode
@@ -50,6 +54,49 @@ local function is_timeout(err)
 end
 
 
+local function sort(t)
+  if t == ngx_null then
+    return "/null/"
+  end
+
+  local typ = type(t)
+  if typ == "table" then
+    local i = 1
+    local o = { "{" }
+    for k, v in tablex.sort(t) do
+      o[i+1] = sort(k)
+      o[i+2] = ":"
+      o[i+3] = sort(v)
+      o[i+4] = ";"
+      i=i+4
+    end
+    if i == 1 then
+      i = i + 1
+    end
+    o[i] = "}"
+
+    return concat(o, nil, 1, i)
+
+  elseif typ == "string" then
+    return '$' .. t .. '$'
+
+  elseif typ == "number" then
+    return '#' .. tostring(t) .. '#'
+
+  elseif typ == "boolean" then
+    return '?' .. tostring(t) .. '?'
+
+  else
+    return '(' .. tostring(t) .. ')'
+  end
+end
+
+
+local function hash(t)
+  return md5(sort(t))
+end
+
+
 function _M.new(parent)
   local self = {
     declarative_config = declarative.new_config(parent.conf),
@@ -67,7 +114,7 @@ function _M:update_config(config_table, update_cache)
   assert(type(config_table) == "table")
 
   local entities, err, _, meta, new_hash =
-              self.declarative_config:parse_table(config_table)
+              self.declarative_config:parse_table(config_table, hash(config_table))
   if not entities then
     return nil, "bad config received from control plane " .. err
   end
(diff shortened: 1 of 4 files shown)


kong/clustering/data_plane.lua
```diff
@@ -62,11 +65,35 @@ function _M.new(parent)
...
+      o[i+2] = ":"
+      o[i+3] = sort(v)
+      o[i+4] = ";"
+      i=i+4
+    end
+
+    return concat(o, nil, 1, i)
+  end
+
+  return tostring(t)
```
kikito, line 83: Don't encode the number `1` the same as the string `"1"`. Also avoid problems with strings starting with `{` (conflict with table separators)
```suggestion
  if type(t) == "string" then
    return '"' .. t:gsub('"', '\\"') .. '"'
  else
    return tostring(t)
  end
```

kikito (changes requested):

kong/clustering/data_plane.lua
```diff
@@ -343,8 +378,6 @@ function _M:communicate(premature)
               -- count is guaranteed to not exceed 1
               config_semaphore:post()
             end
-
-            send_ping(c)
```
fffonion, line 347: why is this removed?

kong/clustering/data_plane.lua
```diff
@@ -62,11 +65,35 @@ function _M.new(parent)
 end
 
 
+local function sort(t)
```
fffonion, line 68: maybe we can have a basic performance on this sort function.
for example if we have a json with 5M of JSON, how long does it take to sort it?

kong/clustering/data_plane.lua
bungle: Because it sends hash in wrong place. It sends old hash, and we want it to send new hash.

kong/clustering/data_plane.lua
bungle: I can do it, but I would expect it to be minimal. Like not even close to all that schema validation etc. stuff that also happens there.

kong/clustering/data_plane.lua
fffonion: ok, let's include this and also the ping_immediately change in commit message (or ideally in seperate commits, either works better for you)

bungle: It is now on separate commit.

dndx:
As discussed offline, let's postpone this to after 2.5 FF since it is a fix and can be in after FF.

kong/clustering/data_plane.lua
bungle: Here some results with 68MB JSON file:

```
(feat/lax-version-checks *) bungle@pro:~/Sources/kong$ ll config.json
-rw-r--r--  1 bungle  staff    68M Jun 23 10:45 config.json
(feat/lax-version-checks *) bungle@pro:~/Sources/kong$ resty perf.lua 
7e111baf0b58b2362b6646d29ca5c458
took: 2.3029999732971 secs
```

And here is the `perf.lua`:
```lua
local cjson = require "cjson"
local tablex = require "pl.tablex"
local concat = table.concat
local tostring = tostring
local ngx_null = ngx.null
local md5 = ngx.md5
local open = io.open


local function sort(t)
  if t == ngx_null then
    return "/null/"
  end

  local typ = type(t)
  if typ == "table" then
    local i = 1
    local o = { "{" }
    for k, v in tablex.sort(t) do
      o[i+1] = sort(k)
      o[i+2] = ":"
      o[i+3] = sort(v)
      o[i+4] = ";"
      i=i+4
    end
    if i == 1 then
      i = i + 1
    end
    o[i] = "}"

    return concat(o, nil, 1, i)

  elseif typ == "string" then
    return '$' .. t .. '$'

  elseif typ == "number" then
    return '#' .. tostring(t) .. '#'

  elseif typ == "boolean" then
    return '?' .. tostring(t) .. '?'

  else
    return '(' .. tostring(t) .. ')'
  end
end


local function hash(t)
  return md5(sort(t))
end


local function read_file(path)
    local file = open(path, "rb")
    local content = file:read "*a"
    file:close()
    return content
end


local content = read_file("config.json")
local json = cjson.decode(content)


ngx.update_time()
local started = ngx.now()
print(hash(json))
ngx.update_time()
local ended = ngx.now()
print("took: ", ended - started, " secs")
```

bungle:
> As discussed offline, let's postpone this to after 2.5 FF since it is a fix and can be in after FF.

I don't think it needs postponed, just that it doesn't need to go to feature freeze. It has plenty of time to go to `2.5` final.

kong/clustering/data_plane.lua
bungle: With 5MB `json` it takes on my `2010` Mac: 0.2 secs.

fffonion: Awesome!

kong/clustering/init.lua
```diff
@@ -5,10 +5,54 @@ local pl_file = require("pl.file")
...
+local tostring = tostring
+local assert = assert
+local concat = table.concat
+local sort = table.sort
+local type = type
 
 
 local MT = { __index = _M, }
 
+local function table_to_sorted_string(t)
```
fffonion, line 19: let's have some basic test case for this function, just to catch any potential
behaviour change in LuaJIT or penlight for example.

fffonion (approved):

PR #7490, opened 2021-06-23
fix(db) Apache Cassandra 4.0 compatibility

mikefero:
### Summary

ALTER statements for existing columns cause migration/bootstrap errors when initializing Kong. This commit ensures that ALTER statement errors are safely ignored for standard migrations and re-entrant migration operations when using Apache Cassandra 4.0.

GitHub Actions were also updated to test against 4.0-rc1.

### Full changelog

* fix(db) Apache Cassandra 4.0 compatibility

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -1021,6 +1021,7 @@ do
           or string.find(err, "[Uu]ndefined column name")
           or string.find(err, "No column definition found for column")
           or string.find(err, "Undefined name .- in selection clause")
+          or string.find(err, "Column with name .- already exists")
           then
             log.warn("ignored error while running '%s' migration: %s (%s)",
                      name, err, cql:gsub("\n", " "):gsub("%s%s+", " "))


kong/db/migrations/core/003_100_to_110.lua
```diff
@@ -309,26 +309,30 @@ return {
...
+
+        ALTER TABLE services ADD tags set<text>;
+        ALTER TABLE routes ADD tags set<text>;
+        ALTER TABLE certificates ADD tags set<text>;
+        ALTER TABLE snis ADD tags set<text>;
+        ALTER TABLE consumers ADD tags set<text>;
+        ALTER TABLE plugins ADD tags set<text>;
+        ALTER TABLE upstreams ADD tags set<text>;
+        ALTER TABLE targets ADD tags set<text>;
+      ]])
```
thibaultcha, line 325: It does not feel safe to me to blindly ignore *all* errors in a process as delicate as these automated C* migrations... Could we do something to only introduce this behavior when connecting to C* 4.x, and to only ignore specifically cherry-picked errors such as the one you observed @mikefero? Something like:
```lua
local res, err = connector:query(...)
if connector.major_version >= 4 and res == nil then
  if not [match err] then
    error(err)
  end
end
```
Doing so would also be helpful in that it would start a record of Kong/C* 4.x compatibility issues and possible roadmap blockers.

bungle: I think we have same issue with other versions too, but currently we ignore errors by error message and each cassandra version changes them, @mikefero can you check this:
https://github.com/Kong/kong/blob/master/kong/db/strategies/cassandra/connector.lua#L1019

Aka just add new strings there and it should be fine?

mikefero: Thanks @bungle, that makes a lot more sense because I have seen the other ALTER commands ignored for the different versions of C*.

.github/workflows/build_and_test.yml
```diff
@@ -251,7 +251,7 @@ jobs:
     strategy:
       matrix:
         suite: [integration, plugins]
-        cassandra_version: [3]
+        cassandra_version: ['3', '4.0']
```
bungle, line 254: @mikefero, I discussed with @gszr, let's just have `4.0` only.

bungle, line 254:
```suggestion
        cassandra_version: ['4.0']
```

bungle (approved):

PR #7501, opened 2021-06-25
[proposal] feat(admin-api) add plugins meta info endpoint

gszr:
### Summary

Introduce a new endpoint - `/plugins/meta` - that returns information about plugins: their priority, version, and which phase handlers they implement.

Sample output:
```
 $ http :8001/plugins/meta                                                                                                              (git:master:dirty)
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 2846
Date: Thu, 24 Jun 2021 23:02:19 GMT
Server: kong/2.4.1
X-Kong-Admin-Latency: 0

{
    "acl": {
        "phases": [
            "access"
        ],
        "priority": 950,
        "version": "3.0.1"
    },
    "acme": {
        "phases": [
            "certificate",
            "access",
            "init_worker"
        ],
        "priority": 1007,
        "version": "0.2.14"
    },
    "aws-lambda": {
        "phases": [
            "access"
        ],
        "priority": 750,
        "version": "3.5.4"
    },
    "azure-functions": {
        "phases": [
            "access"
        ],
        "priority": 749,
        "version": "1.0.0"
    },
    "basic-auth": {
        "phases": [
            "access"
        ],
        "priority": 1001,
        "version": "2.2.0"
    },
    "bot-detection": {
        "phases": [
            "access"
        ],
        "priority": 2500,
        "version": "2.0.0"
    },
    "correlation-id": {
        "phases": [
            "access",
            "init_worker",
            "header_filter"
        ],
        "priority": 1,
        "version": "2.0.2"
    },
    "cors": {
        "phases": [
            "access",
            "header_filter"
        ],
        "priority": 2000,
        "version": "2.0.0"
    },
    "datadog": {
        "phases": [
            "log"
        ],
        "priority": 10,
        "version": "3.0.1"
    },
    "file-log": {
        "phases": [
            "log"
        ],
        "priority": 9,
        "version": "2.1.0"
    },
    "grpc-gateway": {
        "phases": [
            "access",
            "body_filter",
            "header_filter"
        ],
        "priority": 998,
        "version": "0.1.3"
    },
    "grpc-web": {
        "phases": [
            "access",
            "body_filter",
            "header_filter"
        ],
        "priority": 3,
        "version": "0.2.0"
    },
    "hmac-auth": {
        "phases": [
            "access"
        ],
        "priority": 1000,
        "version": "2.3.0"
    },
    "http-log": {
        "phases": [
            "log"
        ],
        "priority": 12,
        "version": "2.1.0"
    },
    "ip-restriction": {
        "phases": [
            "access"
        ],
        "priority": 990,
        "version": "2.0.0"
    },
    "jwt": {
        "phases": [
            "access"
        ],
        "priority": 1005,
        "version": "2.2.0"
    },
    "key-auth": {
        "phases": [
            "access"
        ],
        "priority": 1003,
        "version": "2.4.0"
    },
    "ldap-auth": {
        "phases": [
            "access"
        ],
        "priority": 1002,
        "version": "2.2.0"
    },
    "loggly": {
        "phases": [
            "log"
        ],
        "priority": 6,
        "version": "2.1.0"
    },
    "oauth2": {
        "phases": [
            "access"
        ],
        "priority": 1004,
        "version": "2.1.1"
    },
    "post-function": {
        "phases": [
            "log",
            "body_filter",
            "rewrite",
            "access",
            "certificate",
            "header_filter"
        ],
        "priority": -1000,
        "version": "2.1.0"
    },
    "pre-function": {
        "phases": [
            "log",
            "body_filter",
            "rewrite",
            "access",
            "certificate",
            "header_filter"
        ],
        "priority": "inf",
        "version": "2.1.0"
    },
    "prometheus": {
        "phases": [
            "init_worker",
            "log"
        ],
        "priority": 13,
        "version": "1.3.0"
    },
    "proxy-cache": {
        "phases": [
            "body_filter",
            "access",
            "init_worker",
            "header_filter"
        ],
        "priority": 100,
        "version": "1.3.0"
    },
    "rate-limiting": {
        "phases": [
            "access"
        ],
        "priority": 901,
        "version": "2.2.2"
    },
    "request-size-limiting": {
        "phases": [
            "access"
        ],
        "priority": 951,
        "version": "2.0.0"
    },
    "request-termination": {
        "phases": [
            "access"
        ],
        "priority": 2,
        "version": "2.0.1"
    },
    "request-transformer": {
        "phases": [
            "access"
        ],
        "priority": 801,
        "version": "1.3.2"
    },
    "response-ratelimiting": {
        "phases": [
            "log",
            "access",
            "header_filter"
        ],
        "priority": 900,
        "version": "2.0.1"
    },
    "response-transformer": {
        "phases": [
            "body_filter",
            "header_filter"
        ],
        "priority": 800,
        "version": "2.0.1"
    },
    "session": {
        "phases": [
            "access",
            "header_filter"
        ],
        "priority": 1900,
        "version": "2.4.5"
    },
    "statsd": {
        "phases": [
            "log"
        ],
        "priority": 11,
        "version": "2.0.1"
    },
    "syslog": {
        "phases": [
            "log"
        ],
        "priority": 4,
        "version": "2.2.0"
    },
    "tcp-log": {
        "phases": [
            "log"
        ],
        "priority": 7,
        "version": "2.1.0"
    },
    "udp-log": {
        "phases": [
            "log"
        ],
        "priority": 8,
        "version": "2.1.0"
    },
    "zipkin": {
        "phases": [
            "log",
            "rewrite",
            "access",
            "body_filter",
            "header_filter"
        ],
        "priority": 100000,
        "version": "1.4.0"
    }
}
```

Needs tests.


gszr:
Idea: might also include the language the plugin is implemented in - Lua, Go, JS or Python.

Tieske:
I do not see the value of the phase handler info. What is the use-case for that?

More importantly this should be cluster level config information. What's the use to know what plugins are running on a CP node (that's the only one one can query, whilst the DP ones are more interesting but typically inaccessible)? The info is more useful if we have the info on all the nodes in the cluster. This does not only apply to this info, but more elements of the configuration (or even better, each DP should upload its full config to the CP), and we should be able to query the entire cluster config data from each CP.

gszr:
> I do not see the value of the phase handler info. What is the use-case for that?

I have seen cases where knowing the plugin version is useful (debugging issues), but indeed phases handler info is more questionable if it's useful in practice or not. 

> More importantly this should be cluster level config information. What's the use to know what plugins are running on a CP node (that's the only one one can query, whilst the DP ones are more interesting but typically inaccessible)? The info is more useful if we have the info on all the nodes in the cluster. This does not only apply to this info, but more elements of the configuration (or even better, each DP should upload its full config to the CP), and we should be able to query the entire cluster config data from each CP.

I see that as a different use case / feature. Not every user out there is using hybrid, where that distinction makes sense.

Tieske:
if the feature is not to be extended into hybrid, then we should probably add it under the generic `/` endpoint. This endpoint already has keys for plugins;

- `configuration.plugins`: list of plugins in the node
- `plugins.available_on_server`: another list of plugins in the node (which is badly named, since it is the list of configured plugins, not the available ones)
- `plugins.enabled_in_cluster`: list of plugins in use in the config

so add `plugins.metadata`, and deprecate `plugins.available_on_server` (to be removed in 3.0). That would make more sense to me than adding yet another endpoint with plugin data.

my 2cts.

gszr:
> so add plugins.metadata, and deprecate plugins.available_on_server (to be removed in 3.0). That would make more sense to me than adding yet another endpoint with plugin data.

👍 on this 
The root endpoint might be a better place indeed

javierguerragiraldez:
> Idea: might also include the language the plugin is implemented in - Lua, Go, JS or Python.

external plugins don't declare the language.  :-(

kong/api/routes/plugins.lua
```diff
@@ -112,11 +112,45 @@ local function patch_plugin(self, db, _, parent)
...
+
+  return kong.response.exit(200, meta)
+end
+
 return {
   ["/plugins"] = {
     POST = post_plugin,
   },
 
+  ["/plugins/meta"] = {
```
hbagdi, line 150: /meta/plugins would be better since `/plugins` is already taken up.

hbagdi: Have you considered re-using `/schemas/plugins/:name` endpoint for this purpose?

gszr:
Replaced by https://github.com/Kong/kong/pull/8810

Issue #7451, opened 2021-06-14
[Cassandra] Operation timed out when run kong migrations bootstrap


mrpk1906:
### Summary
- Kong version 2.4.1
- Cassandra Multi DC (6 nodes): 3 on each DC
- When I bootstrap a new cluster, i got a time out error

```
Error: [Cassandra error] cluster_mutex callback threw an error: /usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:41: [Cassandra error] failed to run migration '013_220_to_230' up: [Write timeout] Operation timed out - received only 1 responses.
stack traceback:
	[C]: in function 'assert'
	/usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:41: in function </usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:40>
	[C]: in function 'xpcall'
	/usr/local/share/lua/5.1/kong/db/init.lua:402: in function </usr/local/share/lua/5.1/kong/db/init.lua:352>
	[C]: in function 'pcall'
	/usr/local/share/lua/5.1/kong/concurrency.lua:45: in function 'cluster_mutex'
	/usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:40: in function 'bootstrap'
	/usr/local/share/lua/5.1/kong/cmd/migrations.lua:162: in function 'cmd_exec'
	/usr/local/share/lua/5.1/kong/cmd/init.lua:88: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:88>
	[C]: in function 'xpcall'
	/usr/local/share/lua/5.1/kong/cmd/init.lua:88: in function </usr/local/share/lua/5.1/kong/cmd/init.lua:45>
	/usr/local/bin/kong:9: in function 'file_gen'
	init_worker_by_lua:49: in function <init_worker_by_lua:47>
	[C]: in function 'xpcall'
	init_worker_by_lua:56: in function <init_worker_by_lua:54>

  Run with --v (verbose) or --vv (debug) for more details
```

I tried run several times but got the same error

### Kong config

```
cassandra_repl_strategy = NetworkTopologyStrategy
cassandra_data_centers = dc1:3,dc2:3
cassandra_schema_consensus_timeout = 60000
```

mikefero:
Hello @mrpk1906,

I have a couple of questions for you and some possible recommendations: 

- What version of Apache Cassandra are you running (shouldn't matter, but will be helpful to re-create your exactly setup if needed later)?
[...]

PR #7521, opened 2021-07-01
fix cassandra multi-dc timed out when run migrations bootstrap

mrpk1906:
### Summary

Following the issue #7451, this fix ensures `CREATE TABLE` propagated through in the C* cluster eventually before `INSERT`

### DIFF
--- a/kong/db/migrations/core/013_220_to_230.lua
+++ b/kong/db/migrations/core/013_220_to_230.lua
@@ -50,7 +50,7 @@ return {
     ]], CLUSTER_ID),
   },
   cassandra = {
-    up = string.format([[
+    up = [[
       CREATE TABLE IF NOT EXISTS parameters(
         key            text,
         value          text,
@@ -58,13 +58,19 @@ return {
         PRIMARY KEY    (key)
       );
 
-      INSERT INTO parameters (key, value) VALUES('cluster_id', '%s')
-      IF NOT EXISTS;
-
       ALTER TABLE certificates ADD cert_alt TEXT;
       ALTER TABLE certificates ADD key_alt TEXT;
       ALTER TABLE clustering_data_planes ADD version text;
       ALTER TABLE clustering_data_planes ADD sync_status text;
-    ]], CLUSTER_ID),
+    ]],
+    teardown = function(connector)
+      local coordinator = assert(connector:get_stored_connection())
+      local _, err = coordinator:execute(string.format("INSERT INTO parameters (key, value) VALUES('cluster_id', '%s') IF NOT EXISTS;", CLUSTER_ID))
+      if err then
+        return nil, err
+      end
+
+      return true
+    end
   }
 }


mrpk1906:
@mikefero @locao Can you review it? ^^

dndx:
Hello @mrpk1906,

Thanks for the contribution. I did followed your exchange with @mikefero in #7451 as well but this fix does raises the following concerns for me:

1. The fact that `INSERT` was moved to teardown still doesn't guarantee that the `CREATE` was actually fully propagated before executing `INSERT`. It is more about adding a random delay after `CREATE`, but I think the actual race condition is still there, just less often.
2. Moving the `cluster_id` generation to the `teardown` phase doesn't seems to be the right approach. Suppose the user started the blue/green deployment and upgraded part of the cluster to a newer version, those Kong instances are going to be looking for it's `cluster_id` but wouldn't be able to find. Since the user can theoretically run in this state (without `teardown` being executed) for a long time, this doesn't seems to be the expected behavior either. Usually in `teardown` we only do things that would break older versions of Kong, generating the `cluster_id` certainly doesn't fall under that category.

Not an C* expert here, but does anyone know if there is any way we can wait for C* schema change to reach an agreement before executing the `INSERT` inside the `up` portion of the migration?

mikefero:
> It is more about adding a random delay after CREATE, but I think the actual race condition is still there, just less often.

Completely agree with you here; Apache Cassandra is eventually consistent and each query can use a different set of nodes based on the coordinator chosen during query execution. It is very hard to have guarantees.

> Not an C* expert here, but does anyone know if there is any way we can wait for C* schema change to reach an agreement before executing the INSERT inside the up portion of the migration?

The Lua Cassandra driver should be handling the [schema agreement](https://github.com/thibaultcha/lua-cassandra/blob/c4d27af290daca0dc8b2dfd551736065fafa8231/lib/resty/cassandra/cluster.lua#L764-L794) during query execution, but I think we might want to change the migration consistency `ALL` => `LOCAL_ONE`; see migrations options [here](https://github.com/Kong/kong/blob/master/kong/db/migrations/helpers.lua#L121-L123). This of course has its own drawbacks as the data may not get fully replicated around the cluster and when executing `kong start` and the admin API is used some tables may not be available in the Cassandra cluster causing runtime errors.

Another (slightly better) option might be to use the value defined in `cassandra_write_consistency` and `cassandra_read_consistency` during the migrations instead of relying on the overloaded `ALL` in the helper class. This would put the onus on Kong Gateway OSS configuration instead of trying to infer a users Cassandra cluster configuration.

Besides those, to answer your real question, we could move to `up_f` and use the exposed `wait_schema_consensus` and `check_schema_consensus` from the Lua Cassandra driver and create a wrapper method to handle the logic.

**Note**: If a issue exists in the driver related to schema agreement (which I am pretty sure it does not) we should fix the issue in the driver for the reasons @dndx mentioned in the comments above.

dndx:
@mikefero 

> use the exposed wait_schema_consensus and check_schema_consensus from the Lua Cassandra driver and create a wrapper method to handle the logic.

This definitely sounds like the much better approach, and if done correctly should basically prevent this kind of problems from ever occurring again.

PR #7523, opened 2021-07-01
fix(external-plugins) specific "command not found" error

javierguerragiraldez:
detects the return code 127 (command not found) to display an
appropriate error.

Also avoids "start&fail storms", it should not do more than one spawn
per second.


gszr (approved):
LGTM 👍

PR #7528, opened 2021-07-02
chore(plugins) bump syslog plugin version

locao:
There's a new config entry, if the version is not bumped the docs would be confuse.


locao:
This was already bumped.

PR #7533, opened 2021-07-05
fix(proxy) removal of proxy-authorization header

bungle:
### Summary

Clear proxy authorization header only when it is part of original request headers.
Removes it also when the plugin specifies it with exactly same value as in original
request headers.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1114,10 +1114,9 @@ return {
       local server_port = var.server_port
       ctx.host_port = HOST_PORTS[server_port] or server_port
 
-      -- special handling for proxy-authorization and te headers in case
+      -- special handling for proxy-authorization header in case
       -- the plugin(s) want to specify them (store the original)
       ctx.http_proxy_authorization = var.http_proxy_authorization
-      ctx.http_te                  = var.http_te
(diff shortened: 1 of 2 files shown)


dndx (approved):

kong/runloop/handler.lua
```diff
@@ -1114,10 +1114,9 @@ return {
       local server_port = var.server_port
       ctx.host_port = HOST_PORTS[server_port] or server_port
 
-      -- special handling for proxy-authorization and te headers in case
+      -- special handling for proxy-authorization header in case
       -- the plugin(s) want to specify them (store the original)
       ctx.http_proxy_authorization = var.http_proxy_authorization
-      ctx.http_te                  = var.http_te
```
dndx, line 1120: Quick question. Is this change related to the `proxy-authorization` change?

bungle: this commit removed the need for context value:
https://github.com/Kong/kong/commit/7e669f3ce55a5e6038c4e254029ef2ea061c4209#diff-093344a0609553c69b475fb5cbd57a202796383eb6fca553d26ac6a1f40d4492L1310

But forgot to remove the context value.

PR #7534, opened 2021-07-05
fix(globalpatches) update_time up-value being nil

bungle:


locao (approved):

fffonion (approved):

kong/globalpatches.lua
```diff
@@ -45,20 +45,15 @@ return function(options)
...
-        -- See https://github.com/Kong/lua-resty-worker-events/issues/41
-        update_time()
-      end
+    local alternative_sleep = function(t)
+      require("socket").sleep(t)
+      -- the ngx sleep will yield and hence update time, this implementation
+      -- does not, so we must force a time update to prevent time based loops
+      -- from getting into a deadlock/spin.
+      -- See https://github.com/Kong/lua-resty-worker-events/issues/41
+      ngx.update_time()
```
Tieske, line 56: The original (without calling) should be slightly faster
```suggestion
    local alternative_sleep do
      local luasocket_sleep = require("socket").sleep
      local update_time = ngx.update_time

      alternative_sleep = function(t)
        luasocket_sleep(t)
        -- the ngx sleep will yield and hence update time, this implementation
        -- does not, so we must force a time update to prevent time based loops
        -- from getting into a deadlock/spin.
        -- See https://github.com/Kong/lua-resty-worker-events/issues/41
        update_time()
      end
```

Tieske: I'll leave it to you @bungle , apologies for the silly mistake

bungle: @Tieske haha, i didn't even see that mistake :-). I though it is something todo with test mocking or something :-)

PR #7535, opened 2021-07-05
Handle HEAD requests in buffered mode.

### LINKED ISSUE #7059: Kong Plugin - Can not implement :response() context, 502 Bad Gateway
I am trying to write my own Kong plugin that need to overwrite HTTP :response() context. But it causes 502 Bad Gateway. And nothing in Kong error log.

**My overwrite function in handler.lua:**
```lua
function plugin:response(plugin_conf)
--  kong.service.request.enable_buffering()

  local api_key = kong.response.get_header("X-API-KEY")
  kong.log.debug("xxxxxxxxxxxxxxx: " .. api_key)
end --]]
```

```bash
[Kong-2.3.3:~]# curl -I -H "Host: example.com" http://localhost:8000/
HTTP/1.1 502 Bad Gateway
Server: openresty
Date: Mon, 19 Apr 2021 07:46:10 GMT
Connection: keep-alive
Content-Length: 75
X-Kong-Upstream-Latency: 555
X-Kong-Proxy-Latency: 4
Via: kong/2.3.3
```

What I missed? Thanks in advance! <3

javierguerragiraldez:
According to docs, `ngx.location.capture` returns `res.truncated ==
true` on error cases, but it seems HEAD requests might cause it too.  In
 those cases don't force an error response.

Fixes #7059

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -1052,7 +1052,7 @@ do
     }
 
     local res = ngx.location.capture("/kong_buffered_http", options)
-    if res.truncated then
+    if res.truncated and options.method ~= ngx.HTTP_HEAD then
       kong_global.set_phase(kong, PHASES.error)
       ngx.status = 502
       return kong_error_handlers(ctx)
--- a/spec/02-integration/05-proxy/24-buffered_spec.lua
+++ b/spec/02-integration/05-proxy/24-buffered_spec.lua
@@ -155,6 +155,18 @@ for _, strategy in helpers.each_strategy() do
         assert.equal(md5(body), res.headers["MD5"])
       end)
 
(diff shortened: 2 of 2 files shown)


kong/init.lua
```diff
@@ -1052,7 +1052,7 @@ do
     }
 
     local res = ngx.location.capture("/kong_buffered_http", options)
-    if res.truncated then
+    if res.truncated and options.method ~= ngx.HTTP_HEAD then
```
bungle, line 1055: Could we add a test that `HEAD` call works with `response`?

javierguerragiraldez: done!

bungle (approved):

Issue #7469, opened 2021-03-25
the converting of json To Protobuf doesn't comply with the Protobuf3 spec


AlbertZheng:
Hi,

I am testing this plugin, and found out there might be an issue for converting the value of bytes filed defined in .proto file: the plugin doesn't convert base64 string to binary bytes for bytes field defined in .proto file.

According to the spec ( https://developers.google.com/protocol-buffers/docs/proto3#json), if a field defined .proto is the type 'bytes', the plugin needs to decode the JSON value which is base64 encoding to Protobuf's binary bytes because the target gRPC service is expecting the value of this field is raw bytes content rather than base64 string.

> JSON value will be the data encoded as a string using standard base64 encoding with paddings. Either standard or URL-safe base64 encoding with/without paddings are accepted.

Could you please have a checking this issue?

B.R.
Albert

git-torrent:
Error is located at https://github.com/Kong/kong/blob/master/kong/plugins/grpc-gateway/deco.lua#L339

`msg = encode_json( pb.decode(self.endpoint.output_type, msg) )`

`pb.decode(self.endpoint.output_type, msg)` returns luatable where all the bytes are properly encoded into string. 
`encode_json` takes that string and make it UTF-8. However, all non ASCII bytes get converted into UTF characters and `bytes` get corrupted.

Proper code would be

```
  msg_lua =  pb.decode(self.endpoint.output_type, msg)
  
  msg_lua_fixed = CONVERT_ALL_BYTES_FIELDS_INTO_BASE64( msg_lua )

  msg = encode_json( msg_lua_fixed  )
```

I was unable to make the fix working. The proper way seems to be to use 'pb.hook()' https://github.com/starwing/lua-protobuf#hooks

Hooking of `bytes` type throws errors `unknown_type`. 

The ugly way is to traverse `msg_lua` with `pb.field()` https://github.com/starwing/lua-protobuf#type-information. 

Can anybody help me please?
Thx

git-torrent:
I see hooks are now possible https://github.com/Kong/kong/blob/a6f61269b13c4746a7759de916f5c6d92d11d8dd/kong/tools/grpc.lua#L33

Sadly, this solution cannot handle native types, namely bytes [starwing/lua-protobuf#161 (comment) -](https://github.com/starwing/lua-protobuf/issues/161#issuecomment-927001222), namely [bytes are handled in special way](https://developers.google.com/protocol-buffers/docs/proto3#json)

[...]

PR #7538, opened 2021-07-06
feat(grpc-gateway) encode Timestamp

javierguerragiraldez:
Transform the "well known type" `.google.protobuf.Timestamp` (a
secs+nanos protobuf structure) to/from JSON datetime string.

requires the `encode hooks` feature starwing/lua-protobuf#145

### DIFF
--- a/kong/plugins/grpc-gateway/deco.lua
+++ b/kong/plugins/grpc-gateway/deco.lua
@@ -6,6 +6,7 @@ local cjson = require "cjson"
 local protoc = require "protoc"
 local pb = require "pb"
 local pl_path = require "pl.path"
+local date = require "date"
 
 local setmetatable = setmetatable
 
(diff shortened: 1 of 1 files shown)


fffonion:
👍 @javierguerragiraldez Do we also want to include cases described in https://github.com/Kong/kong/issues/7469?

dndx:
Hey @javierguerragiraldez , do we need any more tests for this as @fffonion suggested?

javierguerragiraldez:
Yes, that's what the new grpc target is for. Won't run in IC without the Google includes, though.

dndx (approved):

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -78,6 +79,47 @@ local function parse_options_path(path)
...
 end
 
+local function safe_set_type_hook(type, dec, enc)
+  if not pcall(pb.hook, type) then
+    ngx.log(ngx.NOTICE, "no type '" .. type .. "' defined")
+    return
+  end
+
+  if not pb.hook(type) then
+    pb.hook(type, dec)
```
git-torrent, line 89: Sadly, this solution cannot handle native types, namely `bytes` - https://github.com/starwing/lua-protobuf/issues/161#issuecomment-927001222

Bytes are handled in special way https://developers.google.com/protocol-buffers/docs/proto3#json

We fixed this by manually traversing the message looking for type and calling respective hook. I can push my solution, but first I want to integrate it with yours.

PR #7547, opened 2021-07-08
perf(*) new tests cases and fixes

fffonion:
### Summary

This PR adds balancer and plugin iterator based tests. In order to setup those tests,
we also add features to start upstream with multiple ports and expose admin API.

### Full changelog

- allow to start upstream with multiple ports …
- add function to write current test case's descriptor
- allow to expose admin port and test on pre releases
- add balancer, plugin iterator tests
- fix quote path in redirection and fix git repo test function
- remove wait in tests, adjust stapxx args for longer duration

### DIFF
--- a/spec/helpers/perf/drivers/docker.lua
+++ b/spec/helpers/perf/drivers/docker.lua
@@ -338,7 +338,7 @@ function _M:generate_flamegraph()
 end
 
 function _M:save_error_log(path)
-  return perf.execute("docker logs " .. self.kong_ct_id .. " 2>" .. path,
+  return perf.execute("docker logs " .. self.kong_ct_id .. " 2>'" .. path .. "'",
                       { logger = self.log.log_exec })
 end
 
--- a/spec/helpers/perf/drivers/docker.lua
+++ b/spec/helpers/perf/drivers/docker.lua
@@ -100,6 +100,20 @@ function _M:teardown()
   return true
 end
 
+local function inject_kong_admin_client(self, helpers)
+  helpers.admin_client = function(timeout)
+    if not self.kong_ct_id then
+      error("helpers.admin_client can only be called after perf.start_kong")
+    end
+    local admin_port, err = get_container_port(self.kong_ct_id, "8001/tcp")
+    if not admin_port then
+      error("failed to get kong admin port: " .. (err or "nil"))
+    end
+    return helpers.http_client("127.0.0.1", admin_port, timeout or 60000)
+  end
+  return helpers
+end
+
 function _M:setup()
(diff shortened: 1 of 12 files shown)


spec/helpers/perf/drivers/terraform.lua
```diff
@@ -237,19 +263,31 @@ function _M:start_kong(version, kong_conf)
...
     "echo > " .. KONG_ERROR_LOG_PATH,
     "sudo id",
+    -- set cpu scheduler to performance, it should lock cpufreq to static freq
     "echo performance | sudo tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor",
-    "dpkg -l kong && (sudo kong stop; sudo dpkg -r kong) || true", -- stop and remove kong if installed
+    -- increase outgoing port range to avoid 99: Cannot assign requested address
+    "sysctl net.ipv4.ip_local_port_range='10240 65535'",
+    -- stop and remove kong if installed
+    "dpkg -l kong && (sudo kong stop; sudo dpkg -r kong) || true",
+    -- have to do the pkill sometimes, why?
```
kikito, line 286: Perhaps `kong stop` sends the signal, but kong is allowed to "linger" for a while. Perhaps there's not enough CPUs on the CI to run the OS to actually remove the pidfile, or something along those lines.

Possible alternative to doing the pkill might be waiting for the pidfile to be removed (or for `pgrep nginx` to return empty)

spec/helpers/perf/drivers/docker.lua
```diff
@@ -197,7 +218,15 @@ function _M:start_upstream(conf)
 
   self.log.info("worker is started")
 
-  return "http://" .. worker_vip .. ":" .. UPSTREAM_PORT
+  if port_count == 1 then
```
kikito, line 221: I think `start_upstream` sometimes returning 1 string and sometimes returning an array of strings makes it more difficult to use. I suggest removing the `if port_count == 1` condition and making it always return a table (which sometimes will have a single element). This will require tweaking on tests that assume that it returns a string, but that seems like a simple change to make.

kikito (approved):
I'm approving this with two very minor suggestions

spec/helpers/perf/drivers/docker.lua
fffonion: that's a good idea.

spec/helpers/perf/drivers/terraform.lua
fffonion: yeah i think i can adjust based on pid, let me test on it.

dndx (approved):

PR #7408, opened 2021-05-31
fix(balancer) ensure existing healthchecker is stopped when processing target event

onematchfox:
### Summary

Ensures that existing healthchecks are stopped when processing CRUD events on targets. 

### Full changelog

* Ensure `stop_healthchecker` is called for existing balancer prior to `create_balancer`. Follows pattern from `do_upstream_event`

**Notes**:
- Haven't added tests for this. Figured I would try copy the existing test from `do_upstream_event` but it doesn't look like this is tested at present? (Was looking at https://github.com/Kong/kong/blob/2.4.1/spec/01-unit/09-balancer_spec.lua so perhaps Iäm in the wrong place?)
- Opened PR against `master`. But it seems like the branching strategy in this repo has changed somewhat since I last submitted a PR. Feel free to let me know if you want this rebased to a (release?) branch.

### Issues resolved

Fix #7406 
Fix #7407

PR #7549, opened 2021-07-08
fix(balancers) stop health check when recreating balancer

locao:
Backport of #7408, balancer code has been rewritten after that PR was merged.


javierguerragiraldez:
looks reasonable.  is it better than how it was done on #7408?  (it was on target events, here it is on balancer recreation)

javierguerragiraldez (approved):

locao:
The only reason for moving this change to the balancer recreation is that the balancer controls the health checker, and targets don't. Essentially both changes will do the same.

Issue #7470, opened 2020-11-20
Proto file with import causes error during REST call


aaronggi:
It appears the file import works correctly, but during the protobuf encoding, there is an error that occurs. This only happens with a .proto file that imports another.

I have grpc-gateway with the latest, being loaded as a custom plugin for kong, for debugging purposes.

When I call the grpc service via grpcurl (through kong's http2 proxy listen port) it works fine. 

Here's my repo I've been running some testing with:
https://github.com/aaronggi/kong-project-poc

Not sure if there's something I'm missing here, but if you want to test it out, you can reproduce with the following:
`docker compose up --build -d &&\`
`curl -XGET http://localhost:8000/CommonTest1/SomeRandomString`

Error in Kong log:
``` 
kong_1           | 2020/11/19 20:53:33 [error] 28#0: *2952 lua coroutine: runtime error: /usr/local/custom/kong/plugins/grpc-gateway/deco.lua:253: bad argument Kong/kong-plugin-grpc-gateway#1 to 'encode' (type '.AIRproto.CommonReq' does not exists)
kong_1           | stack traceback:
kong_1           | coroutine 0:
kong_1           |      [C]: in function 'encode'
kong_1           |      /usr/local/custom/kong/plugins/grpc-gateway/deco.lua:253: in function 'upstream'
kong_1           |      /usr/local/custom/kong/plugins/grpc-gateway/handler.lua:56: in function </usr/local/custom/kong/plugins/grpc-gateway/handler.lua:36>
kong_1           | coroutine 1:
kong_1           |      [C]: in function 'resume'
kong_1           |      coroutine.wrap:21: in function <coroutine.wrap:21>
kong_1           |      /usr/local/share/lua/5.1/kong/init.lua:756: in function 'access'
kong_1           |      access_by_lua(nginx-kong.conf:88):2: in main chunk, client: 172.20.0.1, server: kong, request: "GET /CommonTest1/Aaron HTTP/1.1", host: "localhost:8000" 
kong_1           | 2020/11/19 20:53:33 [error] 28#0: *2952 [kong] init.lua:758 [grpc-gateway] /usr/local/custom/kong/plugins/grpc-gateway/deco.lua:253: bad argument Kong/kong-plugin-grpc-gateway#1 to 'encode' (type '.AIRproto.CommonReq' does not exists), client: 172.20.0.1, server: kong, request: "GET /CommonTest1/Aaron HTTP/1.1", host: "localhost:8000"
[...]

PR #7563, opened 2021-07-13
plugins/grpc-gateway: fix failing imports, issue #7470

git-torrent:
### Summary

Files with imports are not loaded properly due to late fileload

### Full changelog

`p:loadfile(name)` from https://github.com/Kong/kong/blob/master/kong/plugins/grpc-gateway/deco.lua#L132 moved before `p:parsefile(name)` at https://github.com/Kong/kong/blob/master/kong/plugins/grpc-gateway/deco.lua#L94

### Issues resolved

fix for issue #7470  - Proto file with import causes error during REST call


javierguerragiraldez:
Thanks for the PR!  I had to do the same change as part of other enhancement (#7538), so this one is redundant now.

PR #7564, opened 2021-07-13
plugins/grpc-gateway: add missing support for structured URI arguments

### LINKED ISSUE #7552: grpc-gateway plugin: structured URI parameters are not encoded into protobuf
### Summary
- this is bug of grpc-gateway plugin in the actual version
- structured arguments in URI args do not make it into protobuf

### Steps To Reproduce

We are using setup from specification
https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L113

Message:

```
message GetMessageRequest {
  message SubMessage {
    string subfield = 1;
  }
  string message_id = 1; // Mapped to URL path.
  int64 revision = 2;    // Mapped to URL query parameter `revision`.
  SubMessage sub = 3;    // Mapped to URL query parameter `sub.subfield`.
}
```

Request:

```
GET /v1/messages/123456?revision=2&sub.subfield=foo
```

All the structured fields arrive empty, only primitive fields are filled. According to specification, they should be filled as well.

The resulting proto has values

```
  message_id: 123456,
  revision: 2,
  sub.subfield = ""
```
and should be
```
  message_id: 123456,
  revision: 2,
  sub.subfield = "foo"
```

### Additional Details & Logs

At place where URI arguments are encoded into proto
- https://github.com/Kong/kong/blob/master/kong/plugins/grpc-gateway/deco.lua#L247
the values are
```
pb.encode("GetMessageRequest", { ["sub.subfield"] = foo,})
```

This is, according to spec https://github.com/starwing/lua-protobuf#example, not right way to encode nested messages. Proper way is 

pb.encode("GetMessageRequest", { sub = { subfield = foo }} )

Right now we are testing code that fixes this. I can make PR if working and needed.

git-torrent:
### Summary

Google requires messages can be passed as URI arguments
https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L113
e.g. `...?foo.a=A&foo.b=B&foo.c=C`

Upstream function at https://github.com/Kong/kong/blob/master/kong/plugins/grpc-gateway/deco.lua#L247 only prepared table:
```
payload = {
   ["foo.a"] = A,
   ["foo.b"] = B,
   ["foo.c"] = C
}
```
which cannot be encoded into proto message by `pb.encode()` at https://github.com/Kong/kong/blob/master/kong/plugins/grpc-gateway/deco.lua#L247

According to https://github.com/starwing/lua-protobuf#example proper table is:
```
payload = {
   foo = {
     a = A,
     b = B,
     c = C
   }
}
```

### Full changelog

* Added new function `add_to_table` taking converting `["foo.a"] = A,` into `{ foo = { a = A }}`
* `payload[k] = v` at https://github.com/Kong/kong/blob/master/kong/plugins/grpc-gateway/deco.lua#L247 replaced with `        add_to_table( payload, k, v )`

### Issues resolved

Fix #7552 - grpc-gateway plugin: structured URI parameters are not encoded into protobuf


kong/plugins/grpc-gateway/deco.lua
```diff
@@ -199,6 +200,34 @@ local function unframe(body)
   return body:sub(pos, frame_end), body:sub(frame_end + 1)
 end
 
+--[[
+  // Set value `v` at `path` in table `t`
+  // Path contains value address in dot-syntax. For example:
+  // path="a.b.c" would lead to `t[a][b][c] = c`
+]]
+local function add_to_table( t, path, v )
+  local function patch_table( t, path, v )
```
javierguerragiraldez, line 209: this function doesn't use any upvalue from `add_to_table()`.  can it be moved out of that function?  otherwise the (empty) closure object is created on each call.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -199,6 +200,34 @@ local function unframe(body)
...
+    local k = table.remove( path, 1 ) -- pop out first key
+
+    if #path > 0 then
```
javierguerragiraldez, line 212: on an array where we know there are no holes, a faster "emptiness" test is just `if path[1] ~= nil then...`  (or `== nil` and invert the `then` and `else` clauses, whichever is more readable)

javierguerragiraldez (changes requested):
looks good, thanks!  just a few suggestions, remember that this argument processing is in the hot path.

also, note that we're moving this plugin (and a few others) to the main Kong repo.  perhaps you could move this PR there?

kong/plugins/grpc-gateway/deco.lua
git-torrent: I was afraid of spoiling global namespace with `patch_table` helper. But sure, it is outside now.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -199,6 +200,34 @@ local function unframe(body)
   return body:sub(pos, frame_end), body:sub(frame_end + 1)
 end
 
+--[[
+  // Set value `v` at `path` in table `t`
+  // Path contains value address in dot-syntax. For example:
+  // path="a.b.c" would lead to `t[a][b][c] = c`
+]]
+local function add_to_table( t, path, v )
```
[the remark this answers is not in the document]
git-torrent: You are right, I renamed it to `key_stack` with extra commentary to make its meaning clear.

git-torrent:
> also, note that we're moving this plugin (and a few others) to the main Kong repo. perhaps you could move this PR there?

Hm... I was redirected here from https://github.com/Kong/kong-plugin-grpc-gateway. How many Kong repos are there? :)

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -199,6 +200,39 @@ local function unframe(body)
...
+  // Set value `v` at `path` in table `t`
+  // Path contains value address in dot-syntax. For example:
+  // `path="a.b.c"` would lead to `t[a][b][c] = v`.
+]]
+local function add_to_table( t, path, v )
+  local key_stack = {}
+
+  for captures in re_gmatch( path , "[^%.]+") do
+    table.insert( key_stack, captures[0] )
+  end
```
javierguerragiraldez, line 232: > Hm... I was redirected here from https://github.com/Kong/kong-plugin-grpc-gateway. How many Kong repos are there? :)

Sorry for the noise, I was confused by some naming.  This is the right place.

I'm not too happy about constructing a list of segments just to immediately iterate on it, recursively and by removing/shifting from the start.  All this could be resumed with a single iteration, like this:  (also, the regex syntax was a mix of PCRE and Lua patterns)

```suggestion
local function add_to_table( t, path, v )
  local tab = t
  for m in re_gmatch( path , "([^.]+)(\\.)?") do
    local key, dot = m[1], m[2]

    if dot then
      tab[key] = tab[key] or {}
      tab = tab[key]
    else
      tab[key] = v
    end

  end
  return t
end
```

javierguerragiraldez (changes requested):

kong/plugins/grpc-gateway/deco.lua
git-torrent: I really like the way you simplified my code.

javierguerragiraldez: my pleasure!  but it seems I didn't select the whole block and now there's a couple lines loose...

git-torrent: should be fixed by now

kikito (dismissed):
Hello @git-torrent, thanks for this PR. Do you think you would be able to encode a test for the fixes you added? It could be just replicating #7552.

javierguerragiraldez:
Now that #7607 has been merged, it's possible to write a test for this feature.  Here's a simple one, attached as ziped patchfile.  Rebase this PR to the latest `master` before applying it.

[feat(grp-gateway)_test_structured_URI_args.zip](https://github.com/Kong/kong/files/6973339/feat.grp-gateway._test_structured_URI_args.zip)

javierguerragiraldez:
added test and rebased on newest `master`

git-torrent:
Thank you for the test, I was on vacation, happy to see it lives :)

gszr (approved):

gszr:
LGTM. Squash and merge please.

javierguerragiraldez (approved):

gszr:
Thank you for your contribution, @git-torrent! 🎉 🦍

gszr:
@git-torrent, you are eligible for a contributor t-shirt! Please fill out the form [here](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt) to claim yours, making sure to link this PR.

PR #7568, opened 2021-07-15
fix(clustering) ensure data plane config thread is terminated gracefully

### LINKED ISSUE #7567: race condition in data-plane cluster communication causes deadlock
### Summary
- Kong version 2.4.1 - 2.5.0
- Data-planes can get into a state where they are unable to update their configuration from the control plane.
 
### Details

We had some instability with our control plane today, and while investigating I noticed our data planes weren't receiving config updates--or at least weren't updating their filesystem cache, judging by the mtime of `${prefix}/config.cache.json.gz`

I spotted lots and lots of these errors in our data plane logs:

> 2021/07/15 18:27:41 [error] 59#0: *435174 [lua] clustering.lua:222: unable to update running config: busy, context: ngx.timer
> 2021/07/15 18:28:04 [error] 59#0: *435174 [lua] clustering.lua:222: unable to update running config: failed to flip declarative config cache pages: recursive, context: ngx.timer

I traced the first error ("busy") back to the declarative config module trying to grab a lock to serialize some action as part of the update. I think this is where the bug lies:
https://github.com/Kong/kong/blob/00a1e82275ed361c7cf304a7b9c504d636f9c785/kong/clustering/data_plane.lua#L377-L381

Here's the sequence of events that happens within `data_plane:communicate()` to trigger the scenario:

1. reader, writer, config thread are started
2. reader thread receives a reconfigure packet from the control plane and queues it up for the config thread
3. config thread grabs a lock and starts its update process
4. something causes the connection to drop/close
5. reader/writer thread exits due to `socket:receive()` or `socket:send()` returning an error
6. remaining threads are killed, including the config thread, which is still holding a lock
7. entire process is scheduled again via timer
 
The next time this whole thing executes, step 3 fails right away because it can't acquire a lock.
 
The lock has a TTL of 60s, so it will expire on its own, but it's very easy for this scenario to get triggered again if the connection with the control plane is flaky.

The fix for this is pretty simple IMO: clean up the read/write threads first and then signal for the config thread to stop gracefully. I've patched this in my environment already, so I'll probably open a PR for the fix unless someone else gets to it first.

flrgh:
### Summary

Ensure data plane config thread is terminated gracefully, preventing it from getting into a semi-deadlocked state.

It'd be pretty tough to make this testable, so I hope it's okay without any additional tests added. I haven't run the test suite locally so :crossed_fingers: that it passes the automated tests here.

### Issues resolved

Fix #7567

### DIFF
--- a/kong/clustering/data_plane.lua
+++ b/kong/clustering/data_plane.lua
@@ -267,9 +267,10 @@ function _M:communicate(premature)
   --                 and is also responsible for handling timeout detection
 
   local ping_immediately
+  local config_exit
 
   local config_thread = ngx.thread.spawn(function()
-    while not exiting() do
+    while not exiting() and not config_exit do
       local ok, err = config_semaphore:wait(1)
       if ok then
         local config_table = self.next_config
@@ -378,9 +379,19 @@ function _M:communicate(premature)
 
   ngx.thread.kill(read_thread)
   ngx.thread.kill(write_thread)
-  ngx.thread.kill(config_thread)
-
   c:close()
+
+  if not ok then
+    ngx_log(ngx_ERR, _log_prefix, err, log_suffix)
+
+  elseif perr then
+    ngx_log(ngx_ERR, _log_prefix, perr, log_suffix)
+  end
+
+  -- the config thread might be holding a lock if it's in the middle of an
+  -- update, so we need to give it a chance to terminate gracefully
+  config_exit = true
+  ok, err, perr = ngx.thread.wait(config_thread)
 
   if not ok then
     ngx_log(ngx_ERR, _log_prefix, err, log_suffix)


kong/clustering/data_plane.lua
```diff
@@ -389,6 +388,20 @@ function _M:communicate(premature)
     ngx_log(ngx_ERR, _log_prefix, perr, log_suffix)
   end
 
+  -- the config thread might be holding a lock if it's in the middle of an
+  -- update, so we need to give it a chance to terminate gracefully
+  config_exit = true
+  ok, err, perr = ngx.thread.wait(config_thread)
+
+  ngx.thread.kill(config_thread)
```
dndx, line 396: Is this needed after `ngx.thread.wait` already returns? `config_thread` should be dead either way at this point.

dndx (commented):
Thanks for the report @flrgh, it does seems to be a case for deadlock. Just some small questions in regard to the flow below.

kong/clustering/data_plane.lua
flrgh: Ah, you're right. I'll remove this.

flrgh:
@kikito @dndx

Is it possible those failing checks are just flaky and need to be retried? It seems unlikely that this code change would break those specific test cases without breaking a lot of other things.

dndx:
@flrgh Yes, `master` currently have some testing issues, we are working on getting it fixed. Afterwards I will try to get this PR merged.

dndx (approved):

dndx:
@flrgh Merged, thanks for your contribution!

PR #7578, opened 2021-07-21
Tests/fix adminclient timeout

javierguerragiraldez:
make the mock resolver generate some IO under a non-immediate timer.  this seems to ensure the `semaphore:wait()` gets unlocked.

### DIFF
--- a/spec/fixtures/mocks/lua-resty-dns/resty/dns/resolver.lua
+++ b/spec/fixtures/mocks/lua-resty-dns/resty/dns/resolver.lua
@@ -92,11 +92,14 @@ resolver.query = function(self, name, options, tries)
   if answer then
     -- we actually have a mock answer, return it
     ngx.log(ngx.DEBUG, LOG_PREFIX, "serving '", name, "' from mocks")
+    ngx.timer.at(0.05, function() ngx.log(ngx.INFO, "tick") end) -- FIXME: remove after OpenResty semaphore bug is fixed
     return answer, nil, tries
   end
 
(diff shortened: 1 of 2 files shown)


spec/fixtures/mocks/lua-resty-dns/resty/dns/resolver.lua
```diff
@@ -92,11 +92,13 @@ resolver.query = function(self, name, options, tries)
...
     -- we actually have a mock answer, return it
     ngx.log(ngx.DEBUG, LOG_PREFIX, "serving '", name, "' from mocks")
+    ngx.timer.at(0.05, function() ngx.log(ngx.INFO, "tick") end)
     return answer, nil, tries
   end
 
   -- no mock, so invoke original resolver
-  return old_query(self, name, options, tries)
+  ngx.timer.at(0.05, function() ngx.log(ngx.INFO, "tick") end)
+  --return old_query(self, name, options)
```
dndx, line 101: Is this intended? Seems that we bypassed the `old_query` completely which does not seems to be expected.

javierguerragiraldez: i think it's debatable.  the old behaviour was that for non-defined names, the mock ends up calling the real resolver.  if the test asks for an undefined name, i think the answer should be a definite NXDOMAIN, not falling back to the real world and risk finding something (after another indeterminate time)

dndx (approved):

PR #7607, opened 2021-07-27
tests(grpc-gateway) new grpc target

javierguerragiraldez:
Add Go sources, protobuf definition and Make-like process to compile
everything.

First replicate the grpcbin SayHello call.

The protoc compiler, golang-protoc, and golang-grpc-protoc plugins are
required to generate the Go protobuf sourcefiles during development.
Those files are checked in, so if they're not modified (like in CI),
only the Go compiler is required to create the executable.

Required for adding tests to #7564 and #7538


spec/helpers.lua
```diff
@@ -2509,6 +2510,75 @@ local function build_go_plugins(path)
...
+  if not pl_path.exists(path_a) then
+    return true
+  end
+  if not pl_path.exists(path_b) then
+    return false
+  end
+  return assert(pl_path.getmtime(path_b)) > assert(pl_path.getmtime(path_a))
+end
+
+local function make(workdir, specs)
```
fffonion, line 2523: looks like we are replicating `make` here, should we instead use a `Makefile`?

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -89,6 +89,11 @@ local function get_proto_info(fname)
 
   local dir, name = pl_path.splitpath(pl_path.abspath(fname))
   local p = protoc.new()
+  p:addpath("/usr/include")
+  p:addpath("/usr/local/opt/protobuf/include/")
+  p:addpath("/usr/local/kong/lib/")
+  p:addpath("kong")
+
```
fffonion, line 96: it might be good time to abstract `protoc.new()` to Kong level so we can automatically add
common path before return to user. but we can do in seperate PR.

spec/helpers.lua
javierguerragiraldez: yes, i found writing this was easier than finding out if `make` is available in CI

fffonion (approved):

PR #7612, opened 2021-07-28
Updates changelog with performance testing framework

HCloward:
### Summary

The Performance Testing Framework was part of the 2.5 release but absent from the 2.5 changelog.
### Full changelog

* Adds Performance Testing Framework to the 2.5 changelog.


PaulFischer-Kong (approved):

kikito:
Hello @HCloward, usually we don't include testing-related changes in the changelog - it would be too long otherwise. Is there a particular reason to include this one? If not, I think it belongs some place else (perhaps on an announcement or blogpost).

PR #7560, opened 2021-07-13
Decrement workspace_entity_counters on cascade deletes

tyler-ball:
### Summary

Pulling the non-ee changes from https://github.com/Kong/kong-ee/pull/2392 to here. See that PR for an explanation of the issue and my proposed solution.

### Full changelog

☝️ 

### Issues resolved

Fix FTI-1847

PR #7617, opened 2021-07-28
fix(schema) iterate _cache[...].constraints with pairs

tyler-ball:
### Summary

I modified the internal constraints object to be a Hash instead of an
Array. `ipairs` ignores non-numeric keys so in order to iterate it
correctly we need to use `pairs`. I was already doing this for
non-workspace entities but missed doing it for the workspace special
case in `.get_constraints()`.

Original PR: https://github.com/Kong/kong/pull/7560


gszr (approved):

kikito:
ignoring this on changelog, as it's a continuation of #7560

PR #7632, opened 2021-07-30
fix(datadog) remove extra schema field `use_env`

Tieske:
setting either port or host to 'null' will now pick up the environment values for those settings. Still allows to override on a per-plugin basis, just doesn't need the extra `use_env` field.

Based on internal discussions with @artomal and @mikefero 

Ping @rallyben


rallyben:
How can `ngx.null` be provided as a value when configuring the plugin via the API? For example:

```
curl -X POST http://localhost:8001/routes/UUID/plugins --data '
{ "name": "datadog",
   "enabled": True,
    "config": {
      "host": NGX_NULL,
      "port": NGX_NULL,
      ...
}
'
```

Tieske:
@rallyben just use the json `null`

rallyben (approved):

kikito:
Not including this in the changelog as `use_env` was introduced after 2.5.0

PR #7647, opened 2021-08-03
chore(plugins/acme) sync upstream features including Vault K8S auth an…

fffonion:
…d preferred

chain selection

### DIFF
--- a/kong/plugins/acme/client.lua
+++ b/kong/plugins/acme/client.lua
@@ -121,7 +121,8 @@ local function new(conf)
                     " seconds for ACME challenges to propogate")
       ngx.sleep(wait)
       return true
-    end or nil
+    end or nil,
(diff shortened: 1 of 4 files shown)


fffonion:
MIgrated from https://github.com/Kong/kong-plugin-acme/pull/70

kikito (approved):
I am not super comfortable with this way of adding compatibility, but if that's what we need to do, that's what we need to do.

One small comment, though: this does not look like a `feat` (it does not add anything that should appear on the "Additions" section of the changelog). It looks more like a `chore` to me.

fffonion:
@kikito Me either. The good part is that the upstream lua-resty-acme library handles those default value
already. But we should be able to have better handling for this moving forward yes.

kikito:
skipping changelog as this is done for EE compatibility

PR #7660, opened 2021-08-05
chore(tests) bump cassandra from 3 to 4

bungle:
### Summary

Now that Cassandra 4.0.0 is GA we can bump the Cassandra to 4.


bungle:
Seems like this is nor reliable. Needs further investigation.

Issue #7687, opened 2021-08-09
kong router   prefix matching  different behavior  due to regex_priority


jiachinzhao:
there are 2 route rule.

route1:
`
{
        "hosts": ["foo.com"],
        "paths": ["/"],
        "regex_priority": 1
}
`

route2:
`
{
        "hosts": ["foo.com"]
        "paths": ["/v1"],
        "regex_priority": 0
}
`

the following request  will route to route2,   correct behavior due to prefix matching choosing longest length.
```
GET /v1/xxx   HTTP/1.1
Host: foo.com
```

however, after adding another route things changed.
route3: 
`
{
        "hosts": ["bar.com"],
        "paths": ["/.*"],
        "regex_priority": 0
}
`

the same request will route to route1,   which is an unexpected route behavior.

when I remove regex_priority from route1 and route2,  the behavior is corret。

I think regex_priority cannot be used for routing rule sorting when using prefix matching which will sometimes make a mistake.

tested on Kong version 2.1.4
the behavior  still exists on the latest version

kikito:
> when I remove regex_priority from route1 and route2, the behavior is corret。

That ... doesn't seem to be correct. ´regex_priority´is used to disambiguate between two requests that both have regexes. In your example only one of them (route3) has regexes, so `regex_priority` should play no role.

Could you please provide a full list of the calls you are using, so we can reproduce?

Related:
* https://github.com/Kong/kong/issues/4121
* https://discuss.konghq.com/t/a-few-words-on-regex-uris-usage-and-common-pitfalls/506

jiachinzhao:
it can be easily reproduce on k8s
kong ingress controller 0.9.0
kong  2.1.4
apply 2 service and 3 ingress rules.

```
apiVersion: v1
kind: Service
metadata:
  name: my-service1
spec:
  selector:
    app: MyApp
  ports:
    - protocol: TCP
      port: 8080
      targetPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: my-service2
spec:
  selector:
    app: MyApp
  ports:
    - protocol: TCP
      port: 8080
      targetPort: 8080
---
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: route1
  annotations:
    kubernetes.io/ingress.class: kong-test
    konghq.com/regex-priority: "1"
spec:
  rules:
    - host: foo.com
      http:
        paths:
          - path: /
            backend:
              serviceName: my-service1
[...]

PR #7695, opened 2021-08-10
fix sort routes with both regex and prefix routes

jiachinzhao:
### Issues resolved

Fix [#7687](https://github.com/Kong/kong/issues/7687)

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -748,12 +748,15 @@ local function sort_routes(r1, r2)
     end
   end
 
-  do
-    local rp1 = r1.route.regex_priority or 0
-    local rp2 = r2.route.regex_priority or 0
+  -- only regex path use regex_priority 
+  if band(r1.submatch_weight,MATCH_SUBRULES.HAS_REGEX_URI) ~= 0 then
+    do
+      local rp1 = r1.route.regex_priority or 0
+      local rp2 = r2.route.regex_priority or 0
 
-    if rp1 ~= rp2 then
-      return rp1 > rp2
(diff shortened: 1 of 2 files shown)


kong/router.lua
```diff
@@ -748,12 +748,15 @@ local function sort_routes(r1, r2)
     end
   end
 
-  do
-    local rp1 = r1.route.regex_priority or 0
-    local rp2 = r2.route.regex_priority or 0
+  -- only regex path use regex_priority 
+  if band(r1.submatch_weight,MATCH_SUBRULES.HAS_REGEX_URI) ~= 0 then
```
javierguerragiraldez, line 752: this doesn't sound right to me.  rule of "if no regex, then ignore regex_priority" should be applied independently.

javierguerragiraldez (changes requested):

kong/router.lua
jiachinzhao: I add a unit test,  maybe it can explain what I say.
this test will failed on master.
the problem is a route  with path=/.*   host=domain1,  it can affect other domain's route.

javierguerragiraldez: I see now.  not about the test, but I was worried that this `HAS_REGEX_URI` condition on `r1` didn't cover all possible combinations, but now I realize that this only happens when the `.submatch_weight` fields are equal, so testing on a single one is appropriate.

javierguerragiraldez (approved):

bungle (approved):

PR #7704, opened 2021-08-11
fix(tests) force yield (via `sleep(0)`) after semaphore:post()

javierguerragiraldez:
replaces previous hack of "make some IO sometime after".

### DIFF
--- a/spec/fixtures/mocks/lua-resty-dns/resty/dns/resolver.lua
+++ b/spec/fixtures/mocks/lua-resty-dns/resty/dns/resolver.lua
@@ -47,10 +47,10 @@ end
 -- load and cache the mock-records
 local get_mock_records
 do
-  local mock_records
+  local mock_file
   get_mock_records = function()
-    if mock_records then
-      return mock_records
+    if mock_file then
+      return mock_file.records, mock_file.should_fail
     end
 
     local is_file = require("pl.path").isfile
@@ -63,11 +63,11 @@ do
 
     local filename = prefix .. "/" .. MOCK_RECORD_FILENAME
 
-    mock_records = {}
+    mock_file = {}
     if not is_file(filename) then
       -- no mock records set up, return empty default
       ngx.log(ngx.DEBUG, LOG_PREFIX, "bypassing mock, no mock records found")
-      return mock_records
+      return mock_file
     end
 
     -- there is a file with mock records available, go load it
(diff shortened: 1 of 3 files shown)


gszr (approved):

spec/fixtures/mocks/lua-resty-dns/resty/dns/resolver.lua
```diff
@@ -63,43 +63,51 @@ do
...
-  return a, b, c
+  if not should_fail then
+    -- no mock, so invoke original resolver
+    local a, b, c = old_query(self, name, options, tries)
+    return a, b, c
+  end
+end
+
+do
+  local semaphore = require "ngx.semaphore"
```
Tieske, line 105: this looks as something that should go in `globalpatches.lua`, any specific reasons not to put it there @javierguerragiraldez  ?

javierguerragiraldez: so far, we've only seen this happen in the CI test, so I wanted to constrain the change as much as possible.  I'm not sure the openresty bug will be accepted as a bug, so instead of adding yet another patch, we might have to add the `sleep(0)` to the dns client module.   or make a rule of thumb like "if you `:post()` from a timer, be sure to yield before returning"

similar thing with the other part of this fix: i'm pretty sure a mock DNS shouldn't fall back to the real one, but that is the current behaviour, so i made it conditional (and default) to avoid unintended effects

Tieske: >  I'm not sure the openresty bug will be accepted as a bug

are we planning on providing a patch/PR for upstream?

can we at least add a TODO or FIXME comment here to easily track this tech debt?

Tieske: > similar thing with the other part of this fix: i'm pretty sure a mock DNS shouldn't fall back to the real one, but that is the current behaviour, so i made it conditional (and default) to avoid unintended effects

It was created to create not as mock-DNS, but as mock-DNS-records, but I think it makes a nice addition. The only thing missing is the doc-comments describing the additional option (in `helpers.lua`). To check the docs do;
```
cd spec
ldoc .
open docs/index.html
```
So what I get is that the new-default is to NOT fall-back, which is breaking behaviour (from test helper perspective) right? This probably means we should also version the test-helpers.

javierguerragiraldez: no, the default is to behave as previously.  only this one test prevents the mock from going into the real world.

kikito:
Ignoring this in Changelog (the type should have been `tests(dns)` instead of `fix(tests)`

PR #7706, opened 2021-08-12
fix(datadog) replace use_env with env_prefix

fffonion:
- To avoid hybrid mode compatibility issues
- To avoid hard-coded env names

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX


kong/plugins/datadog/statsd_logger.lua
```diff
@@ -19,33 +19,12 @@ local stat_types = {
...
-      return nil, "both KONG_DATADOG_AGENT_HOST and KONG_DATADOG_AGENT_PORT must be set"
-    end
-    host = env_datadog_agent_host
-    port = env_datadog_agent_port
-  else
-    host = conf.host
-    port = conf.port
-  end
-
-  local _, err = sock:setpeername(host, port)
```
fffonion, line 40: this test won't actually test the connection, thus removed

kong/plugins/datadog/handler.lua
```diff
@@ -98,6 +98,18 @@ function DatadogHandler:log(conf)
...
 
+  if conf.env_prefix then
+    local env_host = os.getenv(conf.env_prefix .. '_HOST')
+    local env_port = tonumber(os.getenv(conf.env_prefix .. '_PORT'))
+    if not env_host or not env_port then
+      kong.log.err("both " .. conf.env_prefix .. "_HOST and " .. conf.env_prefix .. "_PORT must be set ",
+                   "when config.env_prefix is configured")
+      return
+    end
+    conf.host = env_host
```
fffonion, line 109: this is not ideal, we are overriding it per request and implictly, potential improvments are:
- use custom_validator to do this in schema part (is the fields mutable in custom_validator?)
- use a cache to memoize env vars, since usually those don't change once process are started

javierguerragiraldez: another issue is that nginx removes environment variables at some point (worker fork?), so i don't think they're available at request time.  what about loading a global dict of environment at during startup?

Tieske:
not sure where this PR came from, since https://github.com/Kong/kong/pull/7632 is already there to fix the compatibility issue.

kong/plugins/datadog/handler.lua
Tieske: @javierguerragiraldez agreed, imo we should not use variable env variables. This becomes extremely messy, if you run many dataplanes, multiple of those variables must be defined on all of the dataplanes then. Big can of worms imo.
So keep them hardcoded, as long as they are properly namespaced that should not be an issue. 

that said, we should drop this and use https://github.com/Kong/kong/pull/7632 instead. My 2cts.

PR #7724, opened 2021-08-17
fix(proxy) keep proxy-authentication request header and proxy-authenticate response header

### LINKED ISSUE #7722: kong client send a request with custom header, but upstream lost header, why ?
kong version: 2.4.1

we use kong as api gateway in k8s, we has a client A ,and client B , A calls B with an custom header, but B receives seems not get the custom header

example :
image

upstream B log:
image

we use nginx to compare, and we can successfully get Proxy-Authorization info 
image

kong ingress configure:
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-kong-test
  namespace: test
  annotations:
    kubernetes.io/ingress.class: "ingress-kong"
    konghq.com/plugins: strip-rewrite
spec:
  rules:
    - http:
        paths:
          - path: /test/v1/(.*)
            backend:
              serviceName: test
              servicePort: 8080

That seems kong strip Proxy-Authorization header to upstream , why is that ? how can we keep custom header to upstream? can anyone help ? thanks !

bungle:
### Summary

It was reported by @ly123-liu on #7722 that Kong removes `Proxy-Authentication` header from upstream
request.

RFC2612 (https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1) specifies them as a hop-by-hop
headers that should be removed. But it did not consider use case of having multiple proxies, so RFC7235
corrected the wording https://datatracker.ietf.org/doc/html/rfc7235#section-4.4.

> A proxy MAY relay the credentials from the client request to the next proxy if that is
> the mechanism by which the proxies cooperatively authenticate a given request.

### Issues Resolved

Fix #7722

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1113,10 +1113,6 @@ return {
     before = function(ctx)
       local server_port = var.server_port
       ctx.host_port = HOST_PORTS[server_port] or server_port
-
-      -- special handling for proxy-authorization header in case
-      -- the plugin(s) want to specify them (store the original)
-      ctx.http_proxy_authorization = var.http_proxy_authorization
     end,
     after = NOOP,
   },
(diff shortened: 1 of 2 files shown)


fffonion (approved):

fffonion:
Not a blokcer to this PR just a thought:
If I understand the RFC correctly, we are okay to relay it to next proxy/upstream as long as Kong doesn't consume the
`proxy-authentication` header. But if a proxy does consumes it, it will then the plugin's responsibility to
clear out the header. This indicates we might want a PDK function to properly follow the RFC, something
like `header_value = kong.request.consume_hop_by_hop_header("proxy-authentication")` and PDK clears
the head before it returns.

PR #7725, opened 2021-08-17
fix(balancer) set :authority on balancer retries

gszr:
### Summary

Ensure the `:authority` pseudo-header is set on balancer retry upstream requests.

Depends on:
- https://github.com/Kong/kong-build-tools/pull/403
- https://github.com/Kong/lua-kong-nginx-module/pull/18

### DIFF
--- a/kong/runloop/balancer/init.lua
+++ b/kong/runloop/balancer/init.lua
@@ -347,6 +347,15 @@ local function set_host_header(balancer_data)
   end
 
   if upstream_host ~= orig_upstream_host then
+    -- the nginx grpc module does not offer a way to override the
+    -- :authority pseudo-header; use our internal API to do so
+    if upstream_scheme == "grpc" or upstream_scheme == "grpcs" then
+      local ok, err = kong.service.request.set_header(":authority", upstream_host)
+      if not ok then
+        log(ERR, "failed to set :authority header: ", err)
+      end
+    end
+
     var.upstream_host = upstream_host
 
     if phase == "balancer" then
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1328,6 +1328,18 @@ return {
       -- `set_host_header` because it would never be empty after the first -- balancer try
       if var.upstream_host ~= nil and var.upstream_host ~= "" then
         balancer_data.preserve_host = true
+
+        -- the nginx grpc module does not offer a way to override the
+        -- :authority pseudo-header; use our internal API to do so
+        -- this call applies to routes with preserve_host=true; for
+        -- preserve_host=false, the header is set in `set_host_header`,
+        -- so that it also applies to balancer retries
+        if var.upstream_scheme == "grpc" or var.upstream_scheme == "grpcs" then
+          local ok, err = kong.service.request.set_header(":authority", var.upstream_host)
+          if not ok then
+            log(ERR, "failed to set :authority header: ", err)
+          end
+        end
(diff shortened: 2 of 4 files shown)


kong/runloop/balancer/init.lua
```diff
@@ -346,6 +346,15 @@ local function set_host_header(balancer_data)
     upstream_host = upstream_host .. ":" .. balancer_data.port
   end
 
+  -- the nginx grpc module does not offer a way to override the
+  -- :authority pseudo-header; use our internal API to do so
+  if upstream_scheme == "grpc" or upstream_scheme == "grpcs" then
+    local ok, err = kong.service.request.set_header(":authority", upstream_host)
```
bungle, line 352: When I read the PDK `set_header` function, it seems only thing this does is:
```
require("resty.kong.grpc").set_authority(upstream_host)
```

Should we skip PDK in this hot path?

kong/runloop/handler.lua
```diff
@@ -1328,6 +1328,18 @@ return {
...
       if var.upstream_host ~= nil and var.upstream_host ~= "" then
         balancer_data.preserve_host = true
+
+        -- the nginx grpc module does not offer a way to override the
+        -- :authority pseudo-header; use our internal API to do so
+        -- this call applies to routes with preserve_host=true; for
+        -- preserve_host=false, the header is set in `set_host_header`,
+        -- so that it also applies to balancer retries
+        if var.upstream_scheme == "grpc" or var.upstream_scheme == "grpcs" then
+          local ok, err = kong.service.request.set_header(":authority", var.upstream_host)
```
bungle, line 1338: Same here, perhaps worth to skip PDK?

bungle (commented):
What do you think @gszr?

kong/runloop/balancer/init.lua
gszr: Good point. I think we had a similar discussion originally but ended up opting to move this into the PDK. @dndx do you have any thoughts here? I am fine either way.

dndx: I think we said in the past that it is generally advisable to use PDK functions in core code if possible as it makes a good example for other developers to rely more on PDK.

In practice, this codepath should be optimized by the JIT compiler anyway so the actual runtime penalty should be minimum. It is probably not worth making into the `set_authority` function in this case.

If we _really_ want, we can cache the `kong.service.request.set_header` function. But again, probably not worth it in this cas.

dndx (approved):

kong/runloop/balancer/init.lua
```diff
@@ -346,6 +346,15 @@ local function set_host_header(balancer_data)
     upstream_host = upstream_host .. ":" .. balancer_data.port
   end
 
+  -- the nginx grpc module does not offer a way to override the
+  -- :authority pseudo-header; use our internal API to do so
+  if upstream_scheme == "grpc" or upstream_scheme == "grpcs" then
```
dndx, line 351: Should we only call `kong.service.request.set_header(":authority", upstream_host)` when `upstream_host ~= orig_upstream_host`? It seems to be unnecessary to call it for every retry if `upstream_host` didn't change.

bungle (approved):

kong/runloop/balancer/init.lua
gszr: Good catch. If the host did not change, there is no reason to call it. Thanks!

PR #7731, opened 2021-08-18
chore(*) import session

gszr:
As a result of an effort to improve external plugins maintainership, we are importing currently-bundled plugins into Kong's main tree. The session plugin used to live in [1].

Additionally to importing plugin contents, this PR also preserves the Git history. History was amended to correct commit scopes - including session in commit messages scope.

[1] https://github.com/Kong/kong-plugin-session


gszr:
Hello, @tschaume! My apologies for the noise; could you sign the CLA so we can proceed with this PR?

javierguerragiraldez (approved):

PR #7736, opened 2021-08-18
chore(deps) bump Penlight to 1.11.0

Tieske:
Changes:
- fix: `stringx.strip` behaved badly with string lengths > 200
- fix: `path.currentdir` now takes no arguments and calls `lfs.currentdir` without argument
- feat: `utils.raise_deprecation` now has an option to NOT include a stack-trace

See changelog: https://github.com/lunarmodules/Penlight/blob/master/CHANGELOG.md#1110-2021-08-18

### DIFF
--- a/kong-2.5.0-0.rockspec
+++ b/kong-2.5.0-0.rockspec
@@ -15,7 +15,7 @@ dependencies = {
   "inspect == 3.1.1",
   "luasec == 1.0.1",
   "luasocket == 3.0-rc1",
-  "penlight == 1.10.0",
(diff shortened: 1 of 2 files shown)


Tieske:
@bungle the test failures are all related to buffered proxying.... any ideas? searched the code base and seems that the `stringx.strip` function is heavily used, so I'd assume the issue to be with that

bungle (approved):

Tieske:
@bungle thx for the fix! btw I can't merge, so if you could...

PR #7742, opened 2021-08-19
perf(uri) escape only when needed

bungle:
### Summary

Runs ngx.re.gsub only when ngx.re.find finds that there is something to escape.
In many cases there will not be anything to escape.

### DIFF
--- a/kong/tools/uri.lua
+++ b/kong/tools/uri.lua
@@ -9,6 +9,7 @@ local string_byte = string.byte
 local string_format = string.format
 local tonumber = tonumber
 local table_concat = table.concat
+local ngx_re_find = ngx.re.find
 local ngx_re_gsub = ngx.re.gsub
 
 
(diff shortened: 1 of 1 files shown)


fffonion:
Verified manually as well:
```lua
local ngx_re_gsub = ngx.re.gsub
local ngx_re_find = ngx.re.find
local string_byte = string.byte
local string_format = string.format

local last_time = 0

local function timeit()
    ngx.update_time()
    local a = last_time
    last_time = ngx.now()
    return last_time - a
end

local ITER = 5000000
local uri = "/  a"
local ESCAPE_PATTERN = "[^!#$&'()*+,/:;=?@[\\]A-Z\\d-_.~%]"

local function escape(m)
  return string_format("%%%02X", string_byte(m[0]))
end


timeit()

local d = uri
for i=1,ITER do
    if ngx_re_find(uri, ESCAPE_PATTERN, "joi") then
        d = ngx_re_gsub(uri, ESCAPE_PATTERN, escape, "joi")
    end
end

print(d)
print(timeit())

local d = nil
for i=1,ITER do
    d = ngx_re_gsub(uri, ESCAPE_PATTERN, escape, "joi")
end

print(d)
print(timeit())
```

no escape:
```
/a
0.15799999237061
/a
1.7810001373291
```

90% faster

need escape
```
/%20%20a
4.345999956131
/%20%20a
4.0720000267029
```

6% slower

the gain is more beneficial than loss

fffonion (approved):

PR #7743, opened 2021-08-20
perf(*) reduce unnecessary read on ngx.var and avoid caching outside

fffonion:
of uncommon branches

Other work like reducing var.http_HEADER and upstream var will be done in seperate PR.

Partly from #7426 #7421

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1126,9 +1126,12 @@ return {
         return
       end
 
+      ctx.scheme = var.scheme
+      ctx.request_uri = var.request_uri
+
       -- routing request
       local router = get_updated_router()
-      local match_t = router.exec()
+      local match_t = router.exec(ctx)
       if not match_t then
         return kong.response.exit(404, { message = "no Route matched with those values" })
       end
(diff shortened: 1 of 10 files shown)


javierguerragiraldez (approved):

kong/runloop/handler.lua
```diff
@@ -1206,37 +1207,38 @@ return {
...
-        return kong.response.exit(426, { message = "Please use HTTP2 protocol" }, {
-          ["connection"] = "Upgrade",
-          ["upgrade"]    = "HTTP/2",
-        })
-      end
+      if protocols.grpc or protocols.grpcs then
+        -- perf: branch usually not taken, don't cache var outside
+        local content_type = var.content_type
+
+        if content_type and find(content_type, "application/grpc", 1, true) == 1 then
```
dndx, line 1219: I think the original version of using `sub(content_type, 1, #"application/grpc") == "application/grpc"` would be faster than `string.find`.

fffonion: I did a quick test, sub is indeed (slightly) faster than find but are in the noise margin. Consider this branch
is only taken in grpc, i will revert it back to sub to reduce diff from original.

bungle: perhaps "slower" but does the sub generate more garbage?

javierguerragiraldez: no, there's some escape analysis and operation folding.  if JITted, the substring is never created, the comparison is made in-place.

PR #7749, opened 2021-08-23
perf(core) iterative plugin iterator

javierguerragiraldez:
motivations:
- even while LuaJIT transforms recursive loops into straight iterations,
 having multiple non-trivial branches results in many side traces,
 easily beyond the ability to unravel the recursion.

- a trace that ends in a `return` but isn't closely associated with the
function entry is aborted with a "NYI: return to lower frame" message.

- the special case of "no self.ctx" is a different behaviour (used
at initialization?). splitting in two different iterators for different
cases, reduces codepaths (and sidetraces) in the inner loop.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX


kong/runloop/plugins_iterator.lua
```diff
@@ -335,14 +344,18 @@ local function iterate(self, phase, ctx)
   local iteration = {
     configure = MUST_LOAD_CONFIGURATION_IN_PHASES[phase],
     loaded = self.loaded,
-    phases = ws.phases[phase] or EMPTY_T,
+    phases = ws.phases[phase] or {},
     combos = ws.combos,
     map = ws.map,
     ctx = ctx,
     i = 0,
   }
```
bungle, line 352: It seem that in `get_next_no_ctx` scenario we will only need:
```lua
local iteration = {
  loaded = self.loaded,
  phases = ws.phases[phase] or {},
  i = 0,
}
```

Perhaps a worth to create a different (smaller) table in that case?

bungle (approved):

bungle:
The ideas from this PR are not included in #7912.

bungle:
Closing this as it was included in #7912.

PR #7763, opened 2021-08-25
perf(plugins-iterator) use stateful iterator to avoid table creation

bungle:
### Summary

Just checking if stateful iterator will help any.

> “it is cheaper to create a closure than a table; second, access to upvalues is faster than access to table fields.”

but well, as @javierguerragiraldez argued below that ^ doesn't hold true with LuaJIT.

### DIFF
--- a/kong/runloop/plugins_iterator.lua
+++ b/kong/runloop/plugins_iterator.lua
@@ -15,7 +15,6 @@ local assert       = assert
 local tostring     = tostring
 
 
-local EMPTY_T      = {}
 local TTL_ZERO     = { ttl = 0 }
 local GLOBAL_QUERY_OPTS = { workspace = null, show_ws_id = true }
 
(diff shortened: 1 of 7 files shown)


javierguerragiraldez:
that quote is from PiL, right?   in LuaJIT closure creation is NYI

kong/runloop/plugins_iterator.lua
```diff
@@ -258,52 +257,95 @@ local function load_configuration_through_combos(ctx, combos, plugin)
...
-          ctx.buffered_proxying = true
+local function get_next_with_plugins(plugins, loaded, phases, map)
+  local i = 1
+  return function()
+    local cfg
+    local plugin = loaded[i]
+    while plugin do
+      local name = plugin.name
+      if map[name] then
+        if phases and phases[name] then
```
javierguerragiraldez, line 310: do you think it's possible to get a single table with the intersection of `map`, `phases` and `plugins`?

kong/runloop/plugins_iterator.lua
```diff
@@ -258,52 +257,95 @@ local function load_configuration_through_combos(ctx, combos, plugin)
...
+      i = i + 1
+      plugin = loaded[i]
     end
-  end
 
-  if self.phases[name] and plugins[name] then
-    return plugin, plugins[name]
+    if plugin then
+      i = i + 1
+    end
```
javierguerragiraldez, line 324: when `plugin` is falsy, the iteration has ended, so there's no harm to increment; i'd think the condition itself can be as heavy as the increment.   did it make difference in the benchmarks?

kong/runloop/plugins_iterator.lua
```diff
@@ -332,17 +374,27 @@ local function iterate(self, phase, ctx)
...
+      ws.phases[phase],
+      ws.map)
+  end
 
-  return get_next, iteration
+  return get_next(
+    self.loaded,
+    ws.phases[phase]
+  )
 end
```
javierguerragiraldez, line 398: i like this three-way factorization a lot.  lets see how it works without closures

kong/runloop/plugins_iterator.lua
bungle: It could be possible, worth to investigate.

PR #7766, opened 2021-08-25
perf(*) use ngx.req.set_header to set upstream_headers

fffonion:
Optimize performance by replacing ngx.var access with ngx.req.set_header to set x-forwarded-* headers. Those
headers are not set until access.after so that plugins that terminates requests in access phase still functions
as before.

Setting other headers through nginx variable are kept as is, some (like TE) are tricky to handle in Lua probably Nginx
has special handling of it, others like Host, Connection and Upgrade are kept since ngx.req.set_header mutates the
request headers directly, and plugins may then not be able to read the original request header.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -16,30 +16,31 @@ local PluginsIterator = require "kong.runloop.plugins_iterator"
 local instrumentation = require "kong.tracing.instrumentation"
 
 
-local kong              = kong
-local type              = type
-local ipairs            = ipairs
-local tostring          = tostring
-local tonumber          = tonumber
-local setmetatable      = setmetatable
-local sub               = string.sub
-local byte              = string.byte
-local gsub              = string.gsub
-local find              = string.find
-local lower             = string.lower
-local fmt               = string.format
-local ngx               = ngx
-local var               = ngx.var
-local log               = ngx.log
-local exit              = ngx.exit
-local exec              = ngx.exec
-local header            = ngx.header
-local timer_at          = ngx.timer.at
-local subsystem         = ngx.config.subsystem
-local clear_header      = ngx.req.clear_header
-local http_version      = ngx.req.http_version
-local unpack            = unpack
-local escape            = require("kong.tools.uri").escape
+local kong         = kong
+local type         = type
+local ipairs       = ipairs
+local tostring     = tostring
+local tonumber     = tonumber
+local setmetatable = setmetatable
+local sub          = string.sub
+local byte         = string.byte
+local gsub         = string.gsub
+local find         = string.find
+local lower        = string.lower
+local fmt          = string.format
+local ngx          = ngx
+local var          = ngx.var
+local log          = ngx.log
+local exit         = ngx.exit
+local exec         = ngx.exec
+local header       = ngx.header
+local timer_at     = ngx.timer.at
+local subsystem    = ngx.config.subsystem
+local set_header   = ngx.req.set_header
+local clear_header = ngx.req.clear_header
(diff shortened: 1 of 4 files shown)


kong/templates/nginx_kong.lua
```diff
@@ -145,12 +139,6 @@ server {
         proxy_set_header      Host               $upstream_host;
         proxy_set_header      Upgrade            $upstream_upgrade;
         proxy_set_header      Connection         $upstream_connection;
-        proxy_set_header      X-Forwarded-For    $upstream_x_forwarded_for;
-        proxy_set_header      X-Forwarded-Proto  $upstream_x_forwarded_proto;
-        proxy_set_header      X-Forwarded-Host   $upstream_x_forwarded_host;
-        proxy_set_header      X-Forwarded-Port   $upstream_x_forwarded_port;
-        proxy_set_header      X-Forwarded-Path   $upstream_x_forwarded_path;
-        proxy_set_header      X-Forwarded-Prefix $upstream_x_forwarded_prefix;
         proxy_set_header      X-Real-IP          $remote_addr;
```
bungle, line 142: As this is related to `X-Forwarded-For` let's have same treatment to it too (X-Real-IP). It can be set similarly where we do `set_header` for `X-Forwarded-*`

bungle (changes requested):

dndx:
Will this affect things like the log serializer so that those upstream headers are returned as request headers?

https://github.com/Kong/kong/blob/master/kong/pdk/log.lua#L677

fffonion:
@dndx yes it will since log_serialzer is in log phase. how do you think for this caveats? maybe we just hide all x-forwarded-
headers in log_serializer?

kikito (dismissed):

dndx:
As discussed offline with @fffonion, let's hold off this PR until we can figure out a way to fix the serializer change. Otherwise it will be user visible impact when we release it.

kong/runloop/handler.lua
```diff
@@ -1278,21 +1281,25 @@ return {
...
       local http_x_forwarded_for = var.http_x_forwarded_for
       if http_x_forwarded_for then
-        var.upstream_x_forwarded_for = http_x_forwarded_for .. ", " ..
-                                       realip_remote_addr
-
+        ctx.upstream_x_forwarded_for = http_x_forwarded_for .. ", " ..
+                                        realip_remote_addr
       else
-        var.upstream_x_forwarded_for = var.remote_addr
+        ctx.upstream_x_forwarded_for = remote_addr
```
zffocussss, line 1295: what is the difference between var.upstream_x_forwarded_for and ctx.upstream_x_forwarded_for?

fffonion: @Suika-Kong is working on some C land work for ngx.req.get_header, which may share some code with this one. Let's see if we can make both of them 3.0.

Suika-Kong: > @Suika-Kong is working on some C land work for ngx.req.get_header, which may share some code with this one. Let's see if we can make both of them 3.0.

Should we discuss on a larger scope, like improving PDK's performance and design? People suggest to put all the request data in Lua land, and only submit them at once when all modifications to the Lua data is done.

bungle:
We need to add changelog entry for this.

flrgh:
This will probably require some updates to EE for WebSocket services since we deviate from the regular runloop handler code path in several places (not a problem--just making note it for my future self).

hbagdi:
@fffonion Is this still required?

fffonion:
this need some rethink, closing this and record it with CT-22

PR #7768, opened 2021-08-25
fix(handler) use request workspace to post balancer events

locao:
### Summary

The workspace id received in the request was being dismissed when posting balancer CRUD events. When workers used these posted events they had to fallback to the default workspace.

### Full changelog

* When an upstream event is received from the DAO `handler.lua` now gets the workspace id from the request and adds it to the upstream entity that will be used in the worker and cluster events. Before this change, when an worker or cluster event was treated, the workspace id was lost and the balancer used the default workspace id as a fallback.
* No tests were added as Kong API Gateway does not use custom workspaces.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -11,6 +11,7 @@ local singletons   = require "kong.singletons"
 local certificate  = require "kong.runloop.certificate"
 local concurrency  = require "kong.concurrency"
 local declarative  = require "kong.db.declarative"
+local workspaces   = require "kong.workspaces"
 local PluginsIterator = require "kong.runloop.plugins_iterator"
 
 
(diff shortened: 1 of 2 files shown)


kong/runloop/handler.lua
```diff
@@ -263,7 +270,7 @@ local function register_balancer_events(core_cache, worker_events, cluster_event
         operation, " to workers: ", err)
     end
     -- => to cluster_events handler
-    local key = fmt("%s:%s:%s", operation, upstream.id, upstream.name)
+    local key = fmt("%s:%s:%s:%s", operation, ws_id, upstream.id, upstream.name)
```
javierguerragiraldez, line 273: shouldn't this be `data.entity.ws_id`?

bungle (approved):

kikito:
This seems a EE-related change so I've decided to skip it from the changelog.

PR #7782, opened 2021-08-30
chore(plugins) import grpc web

gszr:
As a result of an effort to improve external plugins maintainership, we are importing currently-bundled plugins into Kong's main tree. The grpc-web plugin used to live in [1].

Additionally to importing plugin contents, this PR also preserves the Git history. History was amended to correct commit scopes - including grpc-web in commit messages scope.

[1] https://github.com/Kong/kong-plugin-grpc-web


gszr:
Hey @pravin-raha. As we are importing this plugin into Kong, and you have contributed to it, could you please sign the CLA? Thank you!

pravin-raha:
Done

javierguerragiraldez (approved):
so glad this is getting imported!  now we can start sharing a lot of code currently repeated  between this an grpc-gateway.  there are also  features that make sens on both of them but are only implemented on one.

PR #7787, opened 2021-08-31
Add MAX_ONE_OF as entity checker

utix:
### Summary

There is already  `AT_LEAST_ONE_OF` and `ONLY_ONE_OF` schema helper,
in some case we need to have none set or maximum only one.

### Full changelog

* Add a check on `ONLY_ONE_OF` when none of expected value is set this should
be rejected.
* Implements `MAX_ONE_OF` and add tests

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -94,6 +94,7 @@ local validation_errors = {
   CONDITIONAL               = "failed conditional validation given value of field '%s'",
   AT_LEAST_ONE_OF           = "at least one of these fields must be non-empty: %s",
   CONDITIONAL_AT_LEAST_ONE_OF = "at least one of these fields must be non-empty: %s",
+  MAX_ONE_OF                = "maximun one of these fields must be non-empty: %s",
   ONLY_ONE_OF               = "only one of these fields must be non-empty: %s",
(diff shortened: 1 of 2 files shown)


bungle:
I think `only_one_of` marks at least one field required? At least that was my experience when I did this:
https://github.com/Kong/kong/pull/7765

utix:
Nice, It didn't see this new helper, I will take a look
Thanks @bungle.
So what do you suggest, I cancel this PR, and do a new one just updating the test of `only_one_of` ?

bungle:
@utix,

Yes, if possible can you close this and open a new one with added test. I was fighting this very same issue :-). The `only_one_of` is what I tried too, at first. I was also thinking about modifying it to be smarter and based on individual field `required=true|false` whether it really needs `at least one` field to be specified or not, but then I saw we have `mutually_required` already, and just decided to implement `mutually_exclusive` (same as your `max_one_of`).

So yes, if you could send us your added test, we could at least make it official now (that only_one_of also means there must be one, better name for it would be `exactly_one_of` to avoid misunderstandings, but this is how things evolve) :-).

Let us know if `mutually_exclusive` or `mutually_exclusive_sets` don't work as expected. Thank you!

utix:
#7790 done

PR #7795, opened 2021-09-01
chore(proxy-cache) bump patch version because of #7775

bungle:


kikito (approved):

kikito:
Not including in changelog since this is an extension of #7775

PR #7797, opened 2021-09-01
chore(deps) bump lua-resty-http from 0.15 to 0.16.1

bungle:
### Summary

Upgrades `lua-resty-http` from 0.15 to 0.16.1 and removes the deprecated ways of using it.


spec/helpers.lua
```diff
@@ -662,7 +662,51 @@ for _, method_name in ipairs({"get", "post", "put", "patch", "delete"}) do
...
 end
 
---- Creates a http client.
+
+--- Creates a http client from options.
+-- Instead of using this client, you'll probably want to use the pre-configured
+-- clients available as `proxy_client`, `admin_client`, etc. because these come
+-- pre-configured and connected to the underlying Kong test instance.
+--
+-- @function http_client
```
Tieske, line 671: name should be `http_client_opts`

spec/helpers.lua
```diff
@@ -678,14 +722,15 @@ end
 -- @see admin_client
 -- @see admin_ssl_client
 local function http_client(host, port, timeout)
-  timeout = timeout or 10000
-  local self = setmetatable(assert(http.new()), resty_http_proxy_mt)
-  local _, err = self:connect(host, port)
-  if err then
-    error("Could not connect to " .. host .. ":" .. port .. ": " .. err)
+  if type(host) == "table" then
```
Tieske, line 725: should we add a deprecation warning/log in this function?

Tieske (dismissed):
overal looks good. Some minor things.

Don't like so many `ssl_verify = false` though, we should fix those

pintsized (approved):
Nice! LGTM :+1:

bungle:
> Don't like so many `ssl_verify = false` though, we should fix those

I'll create a backlog card for that.

PR #7801, opened 2021-09-02
Cherry pick into 2.5.1

javierguerragiraldez:


kong/plugins/datadog/handler.lua
```diff
@@ -89,7 +89,7 @@ end
 
 local DatadogHandler = {
   PRIORITY = 10,
-  VERSION = "3.0.1",
+  VERSION = "3.1.0",
```
mikefero, line 92: What is changing besides the version; I don't see any underlying plugin changes here or in the `release/2.5.x` branch from the last `2.5.0` release.

kong/plugins/request-termination/handler.lua
```diff
@@ -15,7 +15,7 @@ local RequestTerminationHandler = {}
 
 
 RequestTerminationHandler.PRIORITY = 2
-RequestTerminationHandler.VERSION = "2.0.1"
+RequestTerminationHandler.VERSION = "2.1.0"
```
mikefero, line 18: Same here for the request termination plugin.

javierguerragiraldez: (commit removed)

kong/plugins/datadog/handler.lua
javierguerragiraldez: (commit removed)

akira28:
Sorry for the "rude" question, do you have any timeline for release 2.5.1? At the moment we are using an ugly workaround for the import/export bug, and we would like to avoid it if possible.
Thanks for the great work!

javierguerragiraldez:
> do you have any timeline for the release 2.5.1?

it should go out tomorrow Sept 10

PR #7802, opened 2021-09-02
fix(prometheus) add subsystem label to upstream's health info

gszr:
Ensure upstream's targets health metrics has an `upstream` label to indicate which upstream that metric originates.


kong/plugins/prometheus/exporter.lua
```diff
@@ -146,11 +146,12 @@ local upstream_target_addr_health_table = {
   { value = 0, labels = { 0, 0, 0, "dns_error" } },
 }
 
-local function set_healthiness_metrics(table, upstream, target, address, status, metrics_bucket)
+local function set_healthiness_metrics(table, upstream, target, address, status, subsystem, metrics_bucket)
   for i = 1, #table do
     table[i]['labels'][1] = upstream
     table[i]['labels'][2] = target
     table[i]['labels'][3] = address
+    table[i]['labels'][5] = subsystem
```
fffonion, line 154: I think we can just set the subsystem label in `upstream_target_addr_health_table` on L142 since
subsystem won't change after module load.

fffonion (commented):
LGTM overall, one comment for better performance during scraping

kong/plugins/prometheus/exporter.lua
gszr: Very good catch. Thanks!

fffonion (approved):

gszr:
The motivation for this PR is to disambiguate the health metrics for stream and HTTP subsystems -- in this sense, it could be seen as a "fix", as the old behavior causes confusion. However, it's adding a new label, so I am hesitant to give it a "fix" scope.

fffonion:
I think a `fix` lable makes sense here, though for example user will see different color of lines in grafana after this PR,
their alerting and graph should just work as it before. And it actually makes our official grafana dashboard's upstream
section actully readable. So I'm +1 on making this a fix.

PR #7806, opened 2021-09-03
Chore/import resty dns client

javierguerragiraldez:


javierguerragiraldez:
Hey @martinivanov sorry to call you on this, would you sign Kong's CLA so we can incorporate this module where you contributed to into our codebase?

gszr (commented):
looks good to me overall! we will need to amend commit messages to follow our conventions -- `git filter-branch` might be of help

dndx:
Same with @gszr , I also wonder if we actually need to keep all the commit from the external repo. I will be perfectly fine with squashing everything into a single commit if it makes the Kong repo's history clearer.

javierguerragiraldez:
> I also wonder if we actually need to keep all the commit from the external repo.

No strong feelings either way.  I guess it's arguable that having the history could help to understand why is some code there when fighting a future bug.  But in that case it would be important to compare with the original repo.

If there's no hard objections, I'll squash it.

martinivanov:
@javierguerragiraldez  done

locao (approved):

PR #7810, opened 2021-09-04
feat(core) server verification of client certificates

andref5:
### Summary

Adding the ability to configure a Server with trusted CA certificates in the PEM format used to verify client certificates.
OpenResty already supports, as we can see this is not a big Kong change because use existing methods.
Also, OpenResty does not set $ssl_verify_client on, so TLS is not terminated when verification fails and Kong users can examine Nginx variable $ssl_client_verify on plugins later to determine the next steps, adding more options for plugins. 

- https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_client_certificate
- https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/ssl.md#verify_client

### Full changelog

* Add to Certificate Entity: tls_verify, ca_certificates, tls_verify_depth attributes.
* Setup verification of client certificates on runloop.


andref5:
Hi folks, I'm closing this PR, if you consider this feature in the future I can create another pull request.

Issue #7678, opened 2021-04-16
Include tags in metrics labeles


rmrf:
We can setup tags on kong for each route, these tags have lots of useful  information aside of name,  can prometheus bring these tags on when export metrics?

carnei-ro:
hey, I refactored and opened this new PR https://github.com/Kong/kong/pull/10082;

this is still relevant to our scenario.

I can label a specific set of consumers and/or routes, and if the threshold of 5xx for those calls is reached, I can wake up the SREs to take a look!

StarlightIbuki:
Dear contributor,
We're closing this issue as there hasn't been any update to it for a long time. If the issue is still relevant in the latest version, please feel free to reopen it. We're more than happy to revisit it again. Your contribution is greatly appreciated!
Please have a look at [our pledge to the community](https://github.com/Kong/kong/blob/master/COMMUNITY_PLEDGE.md?rgh-link-date=2023-10-11T06%3A14%3A30Z) for more information.
Sincerely,
Kong Gateway Team

PR #7812, opened 2021-09-06
feat(prometheus): Include tags in metrics labels

carnei-ro:
Moving https://github.com/Kong/kong-plugin-prometheus/pull/149 to here

closes https://github.com/Kong/kong/issues/7678

### DIFF
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -3,6 +3,7 @@ local ngx = ngx
 local find = string.find
 local lower = string.lower
 local concat = table.concat
+local sort = table.sort
 local select = select
 local balancer = require("kong.runloop.balancer")
 local get_all_upstreams = balancer.get_all_upstreams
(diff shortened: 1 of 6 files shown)


kong/plugins/prometheus/exporter.lua
```diff
@@ -168,22 +168,30 @@ if kong_subsystem == "http" then
...
     else
       -- do not record any stats if the service is not present
       return
     end
 
-    local route_name
+    local route_name, route_tags
     if message and message.route then
       route_name = message.route.name or message.route.id
+      route_tags = ((message.route.tags) and (type(message.route.tags) == "table"))
```
hbagdi, line 185: I'm not sure if the sorting order of tags in Kong's DAO is stable or not i.e. can the order of tags change or not.
If it can change, I'd recommend sorting them because otherwise, it will lead to unnecessary duplication of metrics in the Prometheus store.

This is also additional table pollution probably. Wangchong will know best how to tackle that part.

carnei-ro: Right! I added sorting :smile:

spec/03-plugins/26-prometheus/04-status_api_spec.lua
```diff
@@ -232,7 +232,7 @@ describe("Plugin: prometheus (access via status API)", function()
         path    = "/metrics",
       })
       local body = assert.res_status(200, res)
-      return body:find('kong_http_status{service="mock-grpcs-service",route="grpcs-route",code="200"} 1', nil, true)
+      return body:find('kong_http_status{service="mock-grpcs-service",route="grpcs-route",code="200",service_tags="",route_tags=""} 1', nil, true)
```
hbagdi, line 235: Please add a test case for when there are tags on the service or route. Please make sure to have multiple tags to test for sort order.

kong/plugins/prometheus/handler.lua
```diff
@@ -22,6 +22,12 @@ function PrometheusHandler.log(self, conf)
   if conf.per_consumer and message.consumer ~= nil then
     serialized.consumer = message.consumer.username
   end
+  if not conf.expose_services_tags and message.service ~= nil then
+    message.service.tags = {}
```
fffonion, line 26: maybe just `message.service.tags = nil` since you already test if it's a table later
so this function creates less garbage tables.

kong/plugins/prometheus/exporter.lua
```diff
@@ -234,22 +245,30 @@ else
       return
     end
 
-    local service_name
+    local service_name, service_tags
     if message and message.service then
       service_name = message.service.name or message.service.host
+      service_tags = ((message.service.tags) and (type(message.service.tags) == "table"))
```
fffonion, line 251: let's expand this one-liner to a if branch, it feels bit complex to read now.

rmrf:
how this going

mayocream:
@carnei-ro hi, any updates?

carnei-ro:
hey guys, I'm not really sure why some tests are failing - need help

hbagdi:
We won't be able to accept this contribution at this point. The cardinality issues in the plugin have been plaguing Kong's performance for quite some time. Please revisit this change in 6 months.

PR #7815, opened 2021-09-07
chore(pdk) update kong.log.inspect

Tieske:
inspect now works at debug level (was notice). This prevents nginx
from adding the trailing information and rendering the boxes
unreadable. Also the scope is now in the inspect output, in the
header, the top horizontal line.

Before:
Screenshot 2021-09-07 at 16 59 57

After:
Screenshot 2021-09-07 at 16 53 37


Tieske:
fyi: best reviewed by hiding white-space changes

kikito (approved):

PR #7820, opened 2021-09-08
chore(build) update alpine base

gszr:
- use latest in daily build
- use 3.14 in finals

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ pipeline {
                 RELEASE_DOCKER_ONLY="true"
                 PACKAGE_TYPE="apk"
                 RESTY_IMAGE_BASE="alpine"
-                RESTY_IMAGE_TAG="3.13"
+                RESTY_IMAGE_TAG="latest"
             }
             steps {
                 sh 'echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin || true'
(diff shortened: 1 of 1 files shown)


hutchic (approved):

Jenkinsfile
```diff
@@ -92,7 +92,7 @@ pipeline {
                         sh 'echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin || true'
                         sh 'make setup-kong-build-tools'
                         sh 'PACKAGE_TYPE=src RESTY_IMAGE_BASE=src make release'
-                        sh 'PACKAGE_TYPE=apk RESTY_IMAGE_BASE=alpine RESTY_IMAGE_TAG=3 CACHE=false DOCKER_MACHINE_ARM64_NAME="kong-"`cat /proc/sys/kernel/random/uuid` make release'
+                        sh 'PACKAGE_TYPE=apk RESTY_IMAGE_BASE=alpine RESTY_IMAGE_TAG=3.14 CACHE=false DOCKER_MACHINE_ARM64_NAME="kong-"`cat /proc/sys/kernel/random/uuid` make release'
```
Tieske, line 95: why do we have this version here at all? for the finals we should not be picking anything right? and rely on the `docker-kong` default? Otherwise; why do we have the default specified in 2 places?

hutchic: historically we've advised that [kong/kong](https://hub.docker.com/r/kong/kong) should be considered "unstable".

It has our per commit releases - built by the earlier Jenkins stanza `Release Per Commit` resulting in `kong/kong:sha`
It also has the unofficial/official release built by this stanza which results in `kong/kong:x.y.z`

If we want the unofficial/official releases to use the same base image the docker official image uses -- to which I can see an argument for we will need to adjust kong-build-tools accordingly. Until that happens I think let's pin and de-duplicate in subsequent PR's?

hutchic (approved):
LGTM with a note that future us should consider de-duplicating an argument such that official releases to the unofficial docker repository match the official base image exactly

Jenkinsfile
Tieske: yep, let's do that in a follow up.

PR #7827, opened 2021-09-10
feat(clustering) handle removal of fields on control plane

fffonion:
Backporting this from EE. Special handling of vault-auth plugin in compatibility check is not backported
as it's a EE plugin.

Reorganize the code a bit so it can be unit tested.

The next step is to use a script like https://gist.github.com/fffonion/4c0d3b78bed66fef41010ea78d5ff7cb to generate `removal_fields`


kikito (approved):
Looks ok to me, I have extremely minor cosmetic comments which can be ignored.

kong/clustering/control_plane.lua
```diff
@@ -105,27 +108,179 @@ function _M.new(parent)
...
+    version_num = version_num + base * tonumber(v, 10) or 0
+    base = base / 1000
+  end
+
+  return version_num
+end
+-- for test
+_M._dp_version_num = dp_version_num
+
+local removal_fields = {
```
dndx, line 169: I wonder if those should live inside a separate Lua file/module. Having it inside `control_plane.lua` seems to be really inconvenient for updating/patching.

kong/clustering/control_plane.lua
```diff
@@ -105,27 +108,179 @@ function _M.new(parent)
...
+
+  return has_fields and unknown_fields or nil
+end
+-- for test
+_M._get_removed_fields = get_removed_fields
+
+local function update_compatible_payload(payload, dp_version, log_suffix)
+  local fields = get_removed_fields(dp_version_num(dp_version))
+  if fields then
+    local inflated_payload, err = inflate_gzip(payload)
```
dndx, line 238: We should probably restructure the code so that it does not perform this double inflate/deflate. It could get very expensive quickly (especially with large payloads).

kong/clustering/control_plane.lua
fffonion: that was the goal yes, ideally we will want that seperate file to be auto generated but
i would like to make that in a seperate PR. and this one just is just to take what's already
there in EE.

Tieske:
I've had several discussions with @mikefero on this, and we agreed it would be better to not have plugin specific details in `clustering.lua`, it should live with the plugins itself.

Here's some "line-of-thought" code that shows what we discussed: https://github.com/Kong/kong/compare/plugin_versions?expand=1 (the method at the bottom should probably live in `clustering.lua` and not be on a per-plugin basis btw)

imo we should implement something like that in CE and then replace the implementation in EE, instead of backporting the EE solution to CE. My 2cts.

Tieske:
(also: the conversions can be more complex than just adding/removing fields, we should be able to handle that, as in this example)

kong/clustering/control_plane.lua
fffonion: this function has been changed to only return deflated payload if it's been changed.
otherwise the called function must used the `deflated_reconfigure_payload`.

kong/clustering/compat/removed_fields.lua
```diff
@@ -0,0 +1,42 @@
...
+      "keepalive_backlog",
+      "keepalive_pool_size",
+      "read_timeout",
+      "send_timeout",
+    },
+    syslog = {
+      "facility",
+    },
+  }
+}
```
fffonion, line 42:
```suggestion
}

```

fffonion:
@Tieske I think that approach makes sense to me. Just to double check, the "schema_conversion" code will be called
on CP instead of DP correct?

Also as there're multiple approachs to handle this "removed fields" knowledge, i would suggest to pause the effort
to make the removed fields list being generated, and go with the approach EE is working on. How do you think @bungle ?

Tieske:
> Just to double check, the "schema_conversion" code will be called on CP instead of DP correct?

yes. The DP being the lower version, would not have the knowledge to update it. Just thinking out loud; should we be adding this to the `migrations.lua` file maybe? serves a similar purpose, and also is CP side only.

fffonion:
@Tieske I would say probably not migrations.lua. I'm actualy thinking something not Lua but descriptive, like json or yaml. One reason is to make it possible share between different CPs, and also to limit developer from writing
over-complicated or too flexible conversions. We can discuss more detail on the implementation once Harry have the version compat ADR ready.

dndx (approved):

PR #7829, opened 2021-09-10
Feat/proxy cache debug

utix:
### Summary

Remove proxy cache headers from response if the kong debug is not set.
Adding Miss/Hit, Age and cache key to answer shouldn't be the default behavior.

### Full changelog

The following headers:
* X-Cache-Status
* X-Cache-Key
* Age

are now set only if the header kong-debug is set to 1

Refactor test to remove duplicated lines and use get, post and delete helpers instead of send

### DIFF
--- a/kong/plugins/proxy-cache/handler.lua
+++ b/kong/plugins/proxy-cache/handler.lua
@@ -52,6 +52,11 @@ local function overwritable_header(header)
      and not ngx_re_match(n_header, "ratelimit-remaining")
 end
 
+local function set_header(header, value)
+  if ngx.var.http_kong_debug then
+    kong.response.set_header(header, value)
+  end
+end
 
 local function parse_directive_header(h)
   if not h then
@@ -203,8 +208,7 @@ local function signal_cache_req(ctx, cache_key, cache_status)
   ctx.proxy_cache = {
     cache_key = cache_key,
   }
-
-  kong.response.set_header("X-Cache-Status", cache_status or "Miss")
+  set_header("X-Cache-Status", cache_status or "Miss")
 end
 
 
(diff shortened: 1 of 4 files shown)


kong/plugins/proxy-cache/handler.lua
```diff
@@ -356,8 +360,15 @@ function ProxyCacheHandler:access(conf)
     end
   end
 
-  res.headers["Age"] = floor(time() - res.timestamp)
-  res.headers["X-Cache-Status"] = "Hit"
+
+  if ngx.var.http_kong_debug then
```
locao, line 364: Not a blocker, but you could use `set_header` here, right?

locao (approved):

kong/plugins/proxy-cache/handler.lua
utix: It was not clear why it was not the case before, so I have prefered to keep it as it. I will rebase on 2.7.0 and do some tests to check

utix: I get it, on the function `resè is retrieved from the cache
`  local res, err = strategy:fetch(cache_key)`
`res` is updated and answered with `  return kong.response.exit(res.status, res.body, res.headers)`
We have to set res.hearders we cannot use the helper here.

hanshuebner:
@utix Sorry for the very late reply.  Are you still interested in this PR?  If so, would you mind rebasing it to master?

hanshuebner:
We're reverting this PR (https://github.com/Kong/kong/pull/10444) as it seems to cause our tests to fail.  Please reopen if it is still needed.

utix:
@hanshuebner I rebase an fix the tests #10445

PR #7832, opened 2021-09-10
fix(proxy-cache) handle 204/304 responses without Content-Type

utix:
### Summary

According to the RFC 7231 & 7232, 204 and 304 should not have content type
header:

* 204 https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.5
* 304 https://datatracker.ietf.org/doc/html/rfc7232#section-4.1

Initial commit from @adrien-f
see https://github.com/Kong/kong-plugin-proxy-cache/pull/34

### DIFF
--- a/kong/plugins/proxy-cache/handler.lua
+++ b/kong/plugins/proxy-cache/handler.lua
@@ -163,25 +163,29 @@ local function cacheable_response(conf, cc)
   end
 
   do
-    local content_type = ngx.var.sent_http_content_type
+    local status = kong.response.get_status()
+    -- conform to RFC since 204/304 do not have Content-Type header
+    if status ~= 204 and status ~= 304 then
+      local content_type = ngx.var.sent_http_content_type
 
-    -- bail if we cannot examine this content type
-    if not content_type or type(content_type) == "table" or
-       content_type == "" then
+      -- bail if we cannot examine this content type
+      if not content_type or type(content_type) == "table" or
+        content_type == "" then
 
-      return false
-    end
-
-    local content_match = false
-    for i = 1, #conf.content_type do
-      if conf.content_type[i] == content_type then
-        content_match = true
-        break
+        return false
       end
-    end
 
-    if not content_match then
-      return false
+      local content_match = false
+      for i = 1, #conf.content_type do
+        if conf.content_type[i] == content_type then
+          content_match = true
+          break
+        end
+      end
+
+      if not content_match then
+        return false
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


utix:
@kikito  sorry to ping, Should I improve this PR ? Is something missing ?

utix:
@dndx Do you think Kong will be able to merge this fix soon ?
We have proposed this fix in May before the merge of the plugin into the Kong repo.

spec/03-plugins/31-proxy-cache/02-access_spec.lua
```diff
@@ -1179,6 +1179,53 @@ do
...
+        res = assert(client:send {
+          method = "GET",
+          path = "/status/304",
+          headers = {
+            host = "route-10.com",
+          },
+        })
+
+        assert.res_status(304, res)
+        assert.same("Hit", res.headers["X-Cache-Status"])
```
locao, line 1203: This test is not actually testing the change made. Probably it should be:

```suggestion
        assert.same("Hit", res.headers["X-Cache-Status"])
        assert.is_nil(res.headers["Content-Type"])
```

kong/plugins/proxy-cache/handler.lua
```diff
@@ -163,25 +163,29 @@ local function cacheable_response(conf, cc)
   end
 
   do
-    local content_type = ngx.var.sent_http_content_type
+    local status = kong.response.get_status()
+    -- conform to RFC since 204/304 do not have Content-Type header
+    if status ~= 204 and status ~= 304 then
```
locao, line 168: It seems this change is not doing what you expect. If the response has status 204 or 304 you are saying that it should not have the content-type evaluated for caching. So, instead of cleaning the header that should not be sent, you are refusing to check if the content should be cached.

Also, if your intention was to avoid the caching at all, this is not right. The logic test here will only bypass the content-type evaluation, the function will continue to execute, will test the cache-control conditions and the response possibly will still be cached.

locao (changes requested):
Hi @utix, thanks for your submission!

It seems this change is not doing what you expect it to do. Please, take a look at my comments below.

spec/03-plugins/31-proxy-cache/02-access_spec.lua
utix: Without the patch this test fails, with the test passed. We want to test that 304 are put in cache not that the endpoint is respecting the RFC by no setting the content type

kong/plugins/proxy-cache/handler.lua
utix: We want that 204 and 304 are catched which is not the case without the patch.
We are talking that the plugin is expecting the header content type to cache the answer which is wrong

spec/03-plugins/31-proxy-cache/02-access_spec.lua
utix: I update the mock to no send data on 204 and 304

utix:
not interested today on my side

PR #7839, opened 2021-09-13
fix(plugins/acme) escape dots in wildcard domains pattern

### LINKED ISSUE #7834: ACME plugin is not matching wildcard domain correctly

2.4.1 (ACME plugin version: 0.2.14)

### Current Behavior

I am using the ACME plugin with following config (sorry for that domains are redacted since I think they are sensitive):
```yaml
plugin: acme
config:
  account_email: jacky96623@example.com
  tos_accepted: true
  domains:
    - '*.alpha.mydomain.com'
```

HTTP-01 challenge goes smoothly with domains like `www.alpha.mydomain.com`, `api.alpha.mydomain.com`, etc.
But I found that domains like `abc.gw-alpha.mydomain.com` are also matched with the above config.

And I could find debug log showing the case:
```
2021/09/12 07:13:58 [debug] 24#0: *18689 [lua] client.lua:297: post(): acme request: https://acme-v02.api.letsencrypt.org/acme/order/127728407/23840616100 response: {
  "status": "pending",
  "expires": "2021-09-19T07:13:56Z",
  "identifiers": [
    {
      "type": "dns",
      "value": "developer-abc.gw-alpha.mydomain.com"
    }
  ],
  "authorizations": [
    "https://acme-v02.api.letsencrypt.org/acme/authz-v3/30207163860"
  ],
  "finalize": "https://acme-v02.api.letsencrypt.org/acme/finalize/127728407/23840616100"
}
```

### Expected Behavior

ACME plugin should not help to initiate HTTP-01 challenge when the upper level domain does not match.

### Steps To Reproduce

_No response_

### Anything else?

Roughly went through the code of Kong acme plugin, suspect that the issue is related to https://github.com/Kong/kong-plugin-acme/blob/master/kong/plugins/acme/handler.lua#L27-L46. The wildcard domains are stored with initial '*.' removed. The regex matching afterwards do not handle the case that there are other domains similar - possibly with some prefix.

fffonion:
Fix #7834

### DIFF
--- a/spec/03-plugins/29-acme/03-access_spec.lua
+++ b/spec/03-plugins/29-acme/03-access_spec.lua
@@ -104,9 +104,8 @@ for _, strategy in helpers.each_strategy() do
       })
 
       -- key-auth should not run
-      assert.response(res).has.status(200)
-      local body = res:read_body()
-      assert.equal("isme\n", body)
+      local body = assert.response(res).has.status(200)
+      assert.equal("isme", body)
 
       res = assert( proxy_client:send {
         method  = "GET",
--- a/spec/03-plugins/29-acme/03-access_spec.lua
+++ b/spec/03-plugins/29-acme/03-access_spec.lua
@@ -21,7 +21,7 @@ for _, strategy in helpers.each_strategy() do
       }, { "acme", })
 
       assert(bp.routes:insert {
-        hosts = { do_domain, skip_domain },
+        paths = { "/" },
       })
 
       assert(bp.plugins:insert {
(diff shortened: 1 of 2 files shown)


spec/03-plugins/29-acme/03-access_spec.lua
```diff
@@ -96,6 +96,29 @@ for _, strategy in helpers.each_strategy() do
...
+      local res = assert( proxy_client:send {
+        method  = "GET",
+        path    = "/.well-known/acme-challenge/" .. dummy_id,
+        headers =  { host = "a.subdomain." .. do_domain }
+      })
+
+      -- key-auth should not run
+      assert.response(res).has.status(200)
+      local body = res:read_body()
+      assert.equal("isme\n", body)
```
Tieske, line 109:
```suggestion
      local body = assert.response(res).has.status(200)
      assert.equal("isme", body)
```

the `status` assertion returns the body (stripped from white-space iirc)

Tieske (approved):
lgtm. One minor comment, feel free to ignore.

PR #7844, opened 2021-09-14
Implements `stale-while-revalidate` and `s-maxage`

utix:
### Summary

Improve proxy cache plugin to support `stale-while-revalidate` and `s-maxage`

#### stale-while-revalidate

`stale-while-revalidate` indicates that caches MAY serve the response in
which it appears after it becomes stale, up to the indicated number of
seconds.

See [RFC 5861](https://datatracker.ietf.org/doc/html/rfc5861#section-3)

#### s-maxage

`s-maxage` indicates that, in shared caches, the maximum age specified
by this directive overrides the maximum age specified by either:
* the max-age directive or
* the Expires header field.

See [RFC 7234](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2.2.9)

### DIFF
--- a/kong/plugins/proxy-cache/handler.lua
+++ b/kong/plugins/proxy-cache/handler.lua
@@ -311,25 +311,43 @@ function ProxyCacheHandler:access(conf)
     return signal_cache_req(ctx, cache_key, "Bypass")
   end
 
+  local ttl = time() - res.timestamp
   -- figure out if the client will accept our cache value
   if conf.cache_control then
-    if cc["max-age"] and time() - res.timestamp > cc["max-age"] then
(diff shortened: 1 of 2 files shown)


utix:
@bungle sorry to ping, Should I improve this PR ? Is something missing ?

locao (changes requested):
Thanks for you submission, @utix!

The `"s-maxage"` cache-control response directive is already supported by this plugin. Your changes are not according to the RFC's  specification.

The `"stale-while-revalidate"` cache-control response directive implementation probably needs a bigger change than the one presented here. A second ttl entry should be added to the cache entry (`stale_but_valid_ttl`, for instance), and when the request is received after `ttl` is expired but `stale_but_valid_ttl` is not, an async process should be started to update said cache entry while the cached entry is served.

hbagdi:

PR #7852, opened 2021-09-15
perf(balancer) optimize for less table resize

bungle:
Trying to get rid of tab_resize and perhaps the new_key too:
image


fffonion:
[flamegraphs.zip](https://github.com/Kong/kong/files/7171034/flamegraphs.zip)

fffonion (approved):
I can verify handler:before appears to take less portion of time in flamegraph.

javierguerragiraldez (approved):
this is a good pattern. struct-like tables are at their best when all keys are defined in the table constructor

bungle:
Relevant piece of flamegraph:
image

PR #7857, opened 2021-09-16
tests(pdk) make t/Util.pm findable by Perl >= 5.26

kikito:
In Perl 5.26 the `@INC` variable (used for finding libraries) does not
include the current working folder ('.') any more.

https://metacpan.org/pod/perl5260delta#Removal-of-the-current-directory-(%22.%22)-from-@INC

This is considered a security problem so some Linux distros have
backported this change. Perl 5.26 is the system Perl shipped by MacOS
Big Sur.

The fix consists on using `do` instead of `use`. `do` allows for
explicit filepaths, and the current folder (`.`) can be used directly
there.

### DIFF
--- a/t/01-pdk/00-sanity.t
+++ b/t/01-pdk/00-sanity.t
@@ -1,7 +1,7 @@
 use strict;
 use warnings FATAL => 'all';
 use Test::Nginx::Socket::Lua;
-use t::Util;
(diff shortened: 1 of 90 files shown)


Tieske (approved):
lgtm

bungle (approved):

PR #7877, opened 2021-09-22
feat(pdk) add kong.response.get_raw_body and kong.response.set_raw_body

bungle:
### Summary

Adds two new PDK functions that currently only work on `body_filter` phase:

1. kong.response.get_raw_body
2. kong.response.set_raw_body

The `body_filter` phase may be called for each chunk. Sometimes plugins need the full body. If all plugins try to buffer the response, it leads to multiple buffers and even buggy situations. Thus, `kong.response.get_raw_body` was added.

The `kong.response.set_raw_body` as a better way to do:

```
ngx.arg[1] = "hello"
ngx.arg[2] = true
```

in `body_filter` phase. It also clears the `buffer` if one was created by' `kong.response.get_raw_body`.

This also updates two core plugins to use those new functions.

### DIFF
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -19,12 +19,14 @@ local utils = require "kong.tools.utils"
 
 
 local ngx = ngx
+local arg = ngx.arg
 local fmt = string.format
 local type = type
 local find = string.find
 local lower = string.lower
 local error = error
 local pairs = pairs
+local concat = table.concat
 local coroutine = coroutine
 local normalize_header = checks.normalize_header
 local normalize_multi_header = checks.normalize_multi_header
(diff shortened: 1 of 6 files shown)


t/01-pdk/08-response/14-get_raw_body.t
```diff
@@ -0,0 +1,37 @@
...
+run_tests();
+
+__DATA__
+
+=== TEST 1: response.get_raw_body() gets raw body
+--- http_config eval: $t::Util::HttpConfig
+--- config
+    location = /t {
+        content_by_lua_block {
+            ngx.say("Hello, Content by Lua Block")
```
pintsized, line 19: Perhaps another test with a body large enough to span buffers? Otherwise much of the complexity won't be covered. Would `ngx.flush()` perhaps do it? In the jq plugin I did something like `string.rep("*", 1024 * 1024)`, which was enough to reliably cause buffering.

kong/pdk/response.lua
```diff
@@ -517,14 +519,94 @@ local function new(self, major_version)
...
+      body_buffer = self.ctx.core.body_buffer
+      if not body_buffer then
+        return chunk
+      end
+    end
+
+    if type(chunk) == "string" and chunk ~= "" then
+      if not eof then
+        body_buffer = self.ctx.core.body_buffer
+      end
```
pintsized, line 550: I found the assignment order of `body_buffer` a little confusing, perhaps it can be simplified?

```suggestion
    local body_buffer = self.ctx.core.body_buffer
    local chunk, eof = arg[1], arg[2]
    
    if eof and not body_buffer then
      -- the entire body is in this first chunk, no need to buffer
      return chunk
    end

    -- add chunk to buffer
    if type(chunk) == "string" and chunk ~= "" then
```

(untested)

pintsized (commented):
Looks good! A couple of suggestions - test coverage being the more important one.

kong/pdk/response.lua
bungle: it was done to minimize context reads

t/01-pdk/08-response/14-get_raw_body.t
bungle: yes, i was planning to add more tests, but wanted to open it for discussion already

kong/pdk/response.lua
```diff
@@ -517,14 +519,94 @@ local function new(self, major_version)
...
-  --function _RESPONSE.set_body(args, mimetype)
-  --  -- TODO: implement, but how?
-  --end
+  -- @function kong.response.set_raw_body
+  -- @phases `body_filter`
+  -- @tparam string body The raw body
+  -- @return Nothing; throws an error on invalid inputs.
+  -- @usage
+  -- kong.response.set_raw_body("Hello, world!")
+  function _RESPONSE.set_raw_body(body)
```
fffonion, line 597: Although the implementationf of this function does allow it to be called multiple times, but
if I understand correctly, it would be best practice for a user to use combination of get_raw_body and then set_raw_body. And the function name indicates it's setting the body
not chunk. Does it make sense to limit user to only be able to call this function once? By having this limitation, user may have less confusing on how body_filter works and when/how
 to use those two functions.

bungle: @fffonion,

Let's say there are two plugins:
- plugin 1. uppercase response, priority 1
- plugin 2. lowercase response, priority 2

and both plugins are added to route and executed.

1. both call get_raw_body
2. plugin 1 gets full body and turns the response to uppercase and calls set_raw_body
3. plugin 2 gets full body set by plugin 1 and turns the response to lowercase and calls set_raw_body

I think in this scenario it is beneficial to be able to call it multiple times. Of course the above example is crafted, but perhaps there are real use cases, and I think it is strange if plugin 1 can set the body but the plugin 2 cannot anymore.

kong/plugins/response-transformer/handler.lua
```diff
@@ -21,23 +19,15 @@ end
...
-      ngx.arg[1] = body or chunks
-
-    else
-      ctx.rt_body_chunks[ctx.rt_body_chunk_number] = chunk
-      ctx.rt_body_chunk_number = ctx.rt_body_chunk_number + 1
-      ngx.arg[1] = nil
-    end
+  if not is_body_transform_set(conf)
+  or not is_json_body(kong.response.get_header("Content-Type"))
+  then
```
javierguerragiraldez, line 24:
```suggestion
  if not is_body_transform_set(conf)
    or not is_json_body(kong.response.get_header("Content-Type"))
  then
```

javierguerragiraldez (approved):
I like it, plugs an obvious hole.

the grpc-web and grpc-gateway plugins can't use this feature as-is, since both modify the body as a stream.

kong/pdk/response.lua
fffonion: I see, in that case this makes more sense to me.

bungle: @fffonion they don't need to be called as pairs. On last chunk we don't clear anything. On chunks except the last, calling `get_raw_body` will set `ngx.var[1]` to `nil` (this does not happen on last chunk). I think they can be called independently (perhaps there are still some caveats).

This would be wrong usage in plugin:
```
if ngx.var[2] then
  local body = kong.response.get_raw_body()
end
```

As it is too late to start buffering on last chunk, and thus it possibly only returns the last chunk (returns full body if some other plugin has called get_raw_body unconditionally). I am not sure how to tackle this, or does it need to be? Can we just call it invalid use?

bungle: On my testing `body_filter` is usually called at least `2` times. When `ngx.arg[2] == true` the `ngx.arg[1] == nil` seems to be what happens there.

fffonion: @bungle It's okay to have various use case and I don't think the approach here is techically wrong. I'm just more
concerned about the comment (which will become docs automatically), and we should give user a clear idea on
what's happening (like what this function will affect) to avoid misuse, especially those doesn't have prior knowledge to openresty or have written the normal body_filter code but haven't read ours.
I'm also okay to leave this as a future improvement since it's not a blocker after all.

fffonion (approved):

PR #7885, opened 2021-09-24
Use LMDB as DB-less storage

dndx:
DO NOT use in prod, PoC only. It also nukes the shdict based DB-less storage completely from orbit.

### DIFF
--- a/kong-2.5.1-0.rockspec
+++ b/kong-2.5.1-0.rockspec
@@ -87,6 +87,9 @@ build = {
     ["kong.resty.ctx"] = "kong/resty/ctx.lua",
     ["kong.vendor.classic"] = "kong/vendor/classic.lua",
 
+    ["kong.cache.lmdb"] = "kong/cache/lmdb/init.lua",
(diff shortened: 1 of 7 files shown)


fairyqb:
when data storage support ETCD, REDIS,based on the existing infrastructure, reduce management and maintenance costs ,thanks
@bungle @Tieske @kikito @gszr

bungle:
> when data storage support ETCD, REDIS,based on the existing infrastructure, reduce management and maintenance costs

I hear you, but the LMDB as proposed here is about improving our current `dbless`, with an in-process database `lmdb`. The goal is to improve hybrid and dbless modes. Things like incremental/differential updates, two-way communication between CP & DP, and possible between DPs. The `lmdb` here is replacing Nginx shared memory zones used currently with `dbless` and `hybrid data plane nodes`. As far as I know, we don't currently plan adding `etcd` or `redis` support for configuration data. But `redis` could still be more tightly integrated for example for distributed caching.

dndx:
Superseded by https://github.com/Kong/kong/pull/8224.

PR #7895, opened 2021-09-27
docs(admin) add information about HEAD and OPTIONS endpoints

bungle:
### Summary

Adds documentation for `HEAD` and `OPTIONS` admin api endpoints.

Also updates the autogenerated `kong-admin-api.yml`.

### DIFF
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -313,6 +313,42 @@ return {
           ]],
         },
       },
+      ["/[any endpoint]"] = {
+        HEAD = {
+          title = [[Check endpoint or entity existence]],
+          endpoint = [[<div class="endpoint head">/&lt;any-endpoint&gt;</div>]],
+          description = [[Similar to `HTTP GET`, but does not return the body. Returns `HTTP 200` when the endpoint exits or `HTTP 404` when it does not. Other status codes are possible.]],
+          response =[[
+            ```
+            HTTP 200 OK
+            ```
+
+            ```http
+            Access-Control-Allow-Origin: *
+            Content-Length: 11389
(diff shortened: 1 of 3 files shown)


dndx (approved):

autodoc/admin-api/data/admin-api.lua
```diff
@@ -313,6 +313,42 @@ return {
           ]],
         },
       },
+      ["/[any endpoint]"] = {
+        HEAD = {
+          title = [[Check endpoint or entity existence]],
+          endpoint = [[<div class="endpoint head">/&lt;any-endpoint&gt;</div>]],
+          description = [[Similar to `HTTP GET`, but does not return the body. Returns `HTTP 200` when the endpoint exits or `HTTP 404` when it does not. Other status codes are possible.]],
```
dndx, line 320: Should we mention what "other status codes" means?

bungle: I don't think so. I originally put it there, but we cannot know what they possibly mean. In general things like 4xx errors or 5xx errors comes to mind, but other than that, well it is just what happens with GET.

bungle: e.g. sending `GET to /config` on non-dbless node will cause 400 bad request, and same happens with HEAD too.

bungle: In general `204` may happen too, if also `GET` returns `204`. And so it goes... custom endpoints may practically return any status code to any method.

bungle: The other PR was already merged, so that's why I am not going to change this for now. We can add card to tune the language later.

kikito (approved):

PR #7919, opened 2021-09-30
perf(core) store phase in ngx.ctx instead of kong.ctx.core

bungle:
### Summary

Checking if not using `kong.ctx.core` for storing the PHASE leads to any improvement.


akira28:
@bungle it seems that this change causes an error on the log phase. If I try to use kong.request.get_path in the log phase the error `no phase in ngx.ctx.KONG_PHASE` appears if it's a 'native' Kong error (a 404 for a non existing route for example)

PR #7961, opened 2021-10-14
chore(plugins) remove deprecated BasePlugin

javierguerragiraldez:
It has been removed from the docs for some time, but the core still
has to check each method to see if it's actually reimplemented or just
inherited.

Updated several fixture plugins that still used it.


Tieske:
Finally 🎆 

There's no milestone attached. Since this is breaking, this would go in 3.0 I presume?

javierguerragiraldez:
> Since this is breaking, this would go in 3.0 I presume?

I guess it depends on how long has it been deprecated.  It was removed from docs since 2.4.x

javierguerragiraldez:
@aboudreault is the only change in the `.requirements` file?   great

aboudreault:
@javierguerragiraldez I just rebased on latest master and fixed the conflicts in the changelog and the plugin_iterator.lua files.

PR #7993, opened 2021-10-25
Doc (changelog): fix bad link and add Proxy-cache plugin merged info

utix:
Fix the link relative to session plugin

Add the info about proxy-cache merge


kikito:
Thanks, merged!

PR #8008, opened 2021-10-27
perf(plugin-iterator) return zero iterator when no global plugins specified

bungle:
### Summary

We don't need to iterate over the plugins on `certificate` or `rewrite` phases if there are no global plugins configured.

(the perf test is broken, until it is fixed or 2.7 is released, but I don't think we should wait for it)

### DIFF
--- a/kong/runloop/plugins_iterator.lua
+++ b/kong/runloop/plugins_iterator.lua
@@ -416,11 +416,8 @@ local function new_ws_data()
     }
   end
 
-  local plugins = {}
-  plugins[0] = 0
-
   return {
-    plugins = plugins,
+    plugins = { [0] = 0 },
     globals = 0,
     combos = {},
     phases = phases,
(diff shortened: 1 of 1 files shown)


kong/runloop/plugins_iterator.lua
```diff
@@ -414,11 +416,9 @@ local function new_ws_data()
     }
   end
 
-  local plugins = {}
-  plugins[0] = 0
-
   return {
-    plugins = plugins,
+    plugins = { [0] = 0 },
```
fffonion, line 420: nitpick: should we cache the { [0] = 0 } table? seems it's always static.

fffonion (approved):

Godspeed2021:
> ### Summary
> 
> We don't need to iterate over the plugins on `certificate` or `rewrite` phases if there are no global plugins configured.
> 
> (the perf test is broken, until it is fixed or 2.7 is released, but I don't think we should wait for it)

PR #8027, opened 2021-11-02
[DOCU-46] Admin API: add note about Route case sensitivity

falondarville:
### Summary

Update Admin API autodoc to add that Route is case sensitive. Task initially brought up because of user issue:

https://konghq.atlassian.net/browse/DOCU-46

@kikito

### DIFF
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -911,7 +911,9 @@ return {
         created_at = { skip = true },
         updated_at = { skip = true },
         name = {
-          description = [[The name of the Route. Name values must be unique.]]
+          description = [[The name of the Route. Name values must be unique. Route names 
+          are case sensitive. For example, if you name two different Routes "test" and 
+          "Test", Kong Gateway will register them as unique Routes.]]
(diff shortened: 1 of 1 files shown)


autodoc/admin-api/data/admin-api.lua
```diff
@@ -911,7 +911,9 @@ return {
         created_at = { skip = true },
         updated_at = { skip = true },
         name = {
-          description = [[The name of the Route. Name values must be unique.]]
+          description = [[The name of the Route. Name values must be unique. Route names 
+          are case sensitive. For example, if you name two different Routes "test" and 
+          "Test", Kong Gateway will register them as unique Routes.]]
```
kikito, line 916:
```suggestion
          description = [[The name of the Route. Route names must be unique, and they are
          case sensitive. For example, there can be two different Routes named "test" and 
          "Test".]]
```

kikito (approved):
Small suggestion , but feel free to ignore - Approved. 

(I am never sure about calling the reader "you" in the docs - what if they are trying to fix the issue but they are not the ones adding Routes?)

PR #8044, opened 2021-11-05
perf(proxy) do not use pdk methods on hot path

bungle:
### Summary

One very problematic feature of many PDK functions is that they do `check_phase` call, which in turn calls `ngx.ctx` and that creates a whole new ctx table in Lua land.

This commit replaces PDK calls with direct calls to functionality that we need without unnecessary checks and `ngx.ctx` access.


bungle:
Some perf tests. Not a big improvement, but all helps.
[rps-and-latency.zip](https://github.com/Kong/kong/files/7495783/rps-and-latency.zip)
[flamegraphs.zip](https://github.com/Kong/kong/files/7495784/flamegraphs.zip)

fffonion (approved):

PR #8051, opened 2021-11-08
tests(perf) add a timeout for Running Test step

fffonion:
While the root cause for such ephermal timeout in executing
commands still needs investigation, adding a timeout to avoid the
whole job being killed by Github will stop the bleeding


dndx (approved):
LGTM, let's wait for CI to pass before merging :)

bungle (approved):

fffonion:
The tests fails in a inrelavent one:
```
spec/03-plugins/27-aws-lambda/50-http-proxy_spec.lua:117: #keepalive #squid Repeat a request #https via authenticated proxy
```
 merging as is.

PR #8064, opened 2021-11-10
feat(router) add decode_uri_captures parameter

### LINKED ISSUE #7913: uri_captures are url decoded

2.5.0

### Current Behavior

```
{
}
```

In `2.5.0`:
```
{
}
```

### Expected Behavior

For a given request 

```
{
}
```

### Steps To Reproduce

```markdown
1. Running Kong 2.5.0:
```

### Anything else?

_No response_

yankun-li-kong:
Add a new parameter `decode_uri_captures` to decide whether to normalize the captured request URI after route matching.

New feature for #7913

### Summary

Add a new parameter `decode_uri_captures` to decide whether to normalize the captured request URI after route matching.

Only allow setting `on` or `off`:
- `on`: Captured request URI decoding is enabled.
- `off`: Captured request URI decoding is disabled.

By default, this value is set as `on`.

Examples of `decode_uri_captures = on` behavior:

- Assuming that the route path is `/plain/(a%2Eb%20c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`, the captured request URI is decoded as `/plain/a.b c/a.b c`.
- Assuming that the route path is `/plain/(a.b c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`, the captured request URI is decoded as `/plain/a.b c/a.b c`.

Examples of `decode_uri_captures = off`
- Assuming that the route path is `/plain/(a%2Eb%20c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`, the captured request URI is `/plain/a%2Eb%20c/a%2Eb%20c` without URI decoding.
- Assuming that the route path is `/plain/(a.b c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`, the request URI can not match the route path without URI decoding. As a result, the captured request URI is decoded as `/plain/a.b c/a.b c`.

### Full changelog

* Add a parameter `decode_uri_captures` to only decide whether to decode captured request URI or not (Do not have affect for route matching and upstream request)
* Use a self parameter(decode_uri_captures) of kong.router to load the new parameter.
* Add unit tests

### Issues resolved

Fix #7913

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -787,6 +787,22 @@
                                         # connection may be kept open
                                         # indefinitely.
 
+#normalize_req_uri = on                 # Whether to normalize the request URI.
+                                        # If enabled, the request URI will be normalized.
+                                        #
+                                        # Examples of `normalize_req_uri = on`
+                                        # behavior:
+                                        #
+                                        # Assuming the request uri is `/test/some%20resource%20document`,
+                                        # it will be normalized as `/test/some resource document`.
+                                        #
+                                        # Valid values to this setting are:
+                                        #
+                                        # - `on`: Request URI normalization is enabled.
+                                        # - `off`: Request URI normalization is disabled.
+                                        #
+                                        # By default this value is set to `on`.
+
 #------------------------------------------------------------------------------
 # NGINX injected directives
 #------------------------------------------------------------------------------
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -787,21 +787,30 @@
                                         # connection may be kept open
                                         # indefinitely.
 
-#normalize_req_uri = on                 # Whether to normalize the request URI.
-                                        # If enabled, the request URI will be normalized.
+#normalize_uri_captures = on            # Whether to normalize the captured request URI after route matching.
+                                        # If enabled, the captured URI will be normalized.
                                         #
-                                        # Examples of `normalize_req_uri = on`
-                                        # behavior:
+                                        # Examples of `normalize_uri_captures = on`
                                         #
-                                        # Assuming the request uri is `/test/some%20resource%20document`,
-                                        # it will be normalized as `/test/some resource document`.
+                                        # - Assuming that the route path is `/plain/(a%2Eb%20c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`,
+                                        # the captured request URI is normalized as `/plain/a.b c/a.b c`.
+                                        # - Assuming that the route path is `/plain/(a.b c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`,
+                                        # the captured request URI is normalized as `/plain/a.b c/a.b c`.
+                                        #
+                                        # Examples of `normalize_uri_captures = off`
+                                        #
+                                        # - Assuming that the route path is `/plain/(a%2Eb%20c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`,
+                                        # the captured request URI is `/plain/a%2Eb%20c/a%2Eb%20c` without URI normalization.
+                                        # - Assuming that the route path is `/plain/(a.b c)/(.*)` and the request URI is `/plain/a%2Eb%20c/a%2Eb%20c`,
+                                        # the request URI can not match the route path without URI normalization.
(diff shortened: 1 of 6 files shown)


kong/conf_loader/init.lua
```diff
@@ -312,6 +312,7 @@ local CONF_INFERENCES = {
   db_cache_neg_ttl = {  typ = "number"  },
   db_resurrect_ttl = {  typ = "number"  },
   db_cache_warmup_entities = { typ = "array" },
+  normalize_req_uri = { enum = { "on", "off"} },
```
javierguerragiraldez, line 315: why not a boolean?

kong.conf.default
```diff
@@ -786,6 +786,22 @@
...
+                                        #
+                                        # Assuming the request uri is `/test/some%20resource%20document`,
+                                        # it will be normalized as `/test/some resource document`.
+                                        #
+                                        # Valid values to this setting are:
+                                        #
+                                        # - `on`: Request URI normalization is enabled.
+                                        # - `off`: Request URI normalization is disabled.
+                                        #
+                                        # By default this value is set to `on`.
```
javierguerragiraldez, line 803: normalization and decoding are two different things and have different consequences.

this description doesn't explain if the process is done only for route matching or if it affects the upstream request.

yankun-li-kong: thanks for your advice, 
now I am considering only using the unnormalized req uri to get the match_t.matches.uri_captures instead.

Then it should not affect route matching and upstream request

kong/conf_loader/init.lua
yankun-li-kong: Replace with a new parameter and used boolean type

kong.conf.default
yankun-li-kong: Replace `normalize_req_uri` with `normalize_uri_captures` to only affect captured request URI.
Please help check again, thanks!

yankun-li-kong:
Replace `normalize_req_uri` with `normalize_uri_captures` to only affect captured request URI.

When normalize_uri_captures is set as `on`, the behavior is the same as previous.

When normalize_uri_captures is set as `off`, 
kong will first try to get unnormalized uri_captures 
by using unnormalized req_uri and unnormalized regex like below example-1:
Regex: `/plain/(a%2Eb%20c)/(.*)`
Request URI: `/plain/a%2Eb%20c/a%2Eb%20c`
then the uri_captures is `/plain/a%2Eb%20c/a%2Eb%20c` without URI normalization.

If kong failed to get uri_captures, it will nomalize req_uri and regex to get normalized uri_captures instead
like below example-2:
Regex: `/plain/(a.b c)/(.*)` 
Request URI: `/plain/a%2Eb%20c/a%2Eb%20c`
The request URI can not match the regex without URI normalization.
As a result, the uri_captures is normalized as `/plain/a.b c/a.b c`.

Now customers are able to get unnormalized uri_captures by setting normalize_uri_captures to `off`
and using a correct regex for a route path.
It means customers have to consider what is the actual request URI passed to Kong,
then try to match that request URI with route path without URI normalization like the example-1.

yankun-li-kong:
I think I made a mistake by clicking on "Fetch and merge" on [GitHub WebUI](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork)

Please kindly let me know whether I should follow [here](https://medium.com/@topspinj/how-to-git-rebase-into-a-forked-repo-c9f05e821c8a) to rebase my branch with Kong:master instead.

If yes, I could create a new branch/PR

yankun-li-kong:
Rebased

kong.conf.default
javierguerragiraldez: this is still calling the process "normalization", but it's decoding.  i guess the function was badly named, but this change makes the error visible in the documentation and parameter name.

yankun-li-kong: Thanks for your review!
Just submit a commit to rename normalize_uri_captures to decode_uri_captures.

yankun-li-kong:
Rename normalize_uri_captures to decode_uri_captures

bungle:
I don't think we want this to be configuration option.

hbagdi:
@dndx We need to make a decision on if we want to proceed or not here.

kikito:
We decided to not go this way, instead we did #8140

PR #8069, opened 2021-11-10
CT-287 tests(dns-client) more mocks, external DNS was failing

javierguerragiraldez:


javierguerragiraldez:
closed in favor of #8074

PR #8074, opened 2021-11-11
tests(unit) resolve SRV records tests using mock DNS

locao:
There are more tests that could benefit from this approach, but these were failing right now.


javierguerragiraldez:
related: #8069

javierguerragiraldez (approved):

PR #8078, opened 2021-11-15
(chore) document and refactor release scripts

kikito:
- chore(scripts) add instructions to some steps
- refactor(scripts) extract check_requirements to common.sh
- chore(scripts) make-prererelease-release works without flags
- chore(scripts) pass all global arguments to `usage` function in release scripts
- refactor(scripts) move duplicated release steps to common.js

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issues resolved

Fix #XXX

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -1,7 +1,29 @@
 #!/bin/bash
 
+red="\033[0;31m"
+green="\033[0;32m"
+cyan="\033[0;36m"
+bold="\033[1m"
+nocolor="\033[0m"
+
 scripts_folder=$(dirname "$0")
 
+#-------------------------------------------------------------------------------
+function need() {
+  req="$1"
+
+  if [ -z "$(which "$req")" ]; then
+     echo "Required command $req not found."
+     exit 1
+  fi
+}
+
+#-------------------------------------------------------------------------------
+function check_requirements() {
+   need git
+   need hub
+   need sed
+}
 #-------------------------------------------------------------------------------
 function commit_changelog() {
     if ! git status CHANGELOG.md | grep -q "modified:"
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -39,12 +39,20 @@ function check_requirements() {
    need sed
 }
 
+
+#-------------------------------------------------------------------------------
+function yesno() {
+  echo "$1"
+  read -r
+  if [[ "$REPLY" =~ ^[yY] ]]; then
+    return 0
+  fi
+  return 1
+}
+
 #-------------------------------------------------------------------------------
 function check_milestone() {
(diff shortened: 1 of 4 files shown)


fffonion:
@kikito Is there a specific parts needs to be reviewed other than trying this out locally (which I assume you have already did)?

scripts/common.sh
```diff
@@ -1,49 +1,150 @@
...
+  browser=open
+fi
+
+EDITOR="${EDITOR-$VISUAL}"
+
+#-------------------------------------------------------------------------------
+function need() {
+  req="$1"
+
+  if [ -z "$(which "$req")" ]; then
```
flrgh, line 29: More of an FYI than a `nit`, and feel free to ignore: this can be implemented with bash builtins like `type` or `command` instead of using the `which` binary, which I think makes things a little bit cleaner:

```suggestion
  if ! type -t "$req" &>/dev/null; then
```

The only difference is that (as opposed to `which`) this will also return true for bash functions, aliases, and builtins, which might not be your desired behavior.

kikito:
@fffonion that is in fact a very good point. There is no "dry-run" mode for these scripts so "manually trying" is often not even an option.

I have tried *some* of the steps manually, but I have not tried all the combinations I am afraid T__T

scripts/common.sh
```diff
@@ -312,10 +398,246 @@ function CONFIRM() {
...
+  then
+    die "Release commit is not at the top of the current branch. Did you commit the version bump?"
+  fi
+
+  git log
+
+  CONFIRM "Press Enter to push the branch and open the release PR" \
+    "or Ctrl-C to cancel."
+
+  set -e
```
jschmid1, line 471: an overall comment on the `set -e` and `SUCCESS` interaction. A nicer and more userfriendly way would be to use traps. This maybe out of scope for PR but I'll leave it here for inspiration.

jschmid1 (commented):
looks good overall, hard to _actually_ test though without a proper dryrun mode.

scripts/common.sh
kikito: I admit that I have never used `trap` in such a way (I try to avoid bash script when I can ...).

From what I have read, it seems that we would *still* need `set -e` (to produce the signal that then would be captured by trap).

Let's call it out of scope for now, it seems a bit ... philosophical as well (one global rescue function vs lots of small traps?)

jschmid1 (approved):

scripts/common.sh
kikito: We had a live conversation about this.

@gszr , I think there must be a particular reason why `set -e` is set/unset on every action instead of setting it once globally. Is there? What is your opinion of using `trap` to recover? At the very least we should be able to print "the command blablabla failed with blebleble"

PR #8082, opened 2021-11-16
fix(aws-lambda-plugin) require AWS region, allow custom lambda hosts

segabond:
### Summary

As per FTI-2928, customer uses Kong in front of AWS lambda in the isolated gov AWS region. Such regions use custom service endpoints, therefore customer uses `host` configuration to override it. Since config schema does not allow both host and region be supplied simultaneously, it is impossible to correctly sign the request using AWS SigV4 algorithm. The correct implementation of such plugin should allow both `aws_region` and `host` be supplied, as `aws_region` is a mandatory field for SigV4 signature algo.

### Full changelog

* Allow both host and aws_region in config
* Require AWS region to always be configured
* Always apply SigV4 signature
* Fixed tests to test environment variable configuration correctly

### Issues resolved

Fix FTI-2928

### DIFF
--- a/kong/plugins/aws-lambda/schema.lua
+++ b/kong/plugins/aws-lambda/schema.lua
@@ -106,6 +106,5 @@ return {
   entity_checks = {
     { mutually_required = { "config.aws_key", "config.aws_secret" } },
     { mutually_required = { "config.proxy_scheme", "config.proxy_url" } },
-    { mutually_exclusive = { "config.aws_region", "config.host" } },
   }
 }
--- a/kong/plugins/aws-lambda/v4.lua
+++ b/kong/plugins/aws-lambda/v4.lua
@@ -87,10 +87,6 @@ local function prepare_awsv4_request(tbl)
   local path = tbl.path
   local host = tbl.host
 
-  if region then
-    host = service .. "." .. region .. "." .. domain
-  end
-
   if path and not canonicalURI then
     canonicalURI = canonicalise_path(path)
   elseif canonicalURI == nil or canonicalURI == "" then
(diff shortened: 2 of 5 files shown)


Tieske:
If the region now is required, then this is a breaking change. And even then I don't see how we can add that info with a migration.

pinging the migrations/breaking change master; @mikefero

RobSerafini:
@Tieske @aboudreault @kikito - what is the status of this PR?

aboudreault:
doesn't seem like we can merge this feature due to the breaking change mentioned. We cannot really guess the region if none is provided in the config. Options:
- target this feature for 3.0 with the breaking change
- revisit this feature implementation to preserve the current behavior until 3.0 (so keeping aws_region optional, and don't try to sign v4 if not specified?)

However, we are probably too short to include this in 2.8.

kikito:
This did not make it into 2.8, I am clearing its milestone

tyler-ball:
Rebased off master

kong/plugins/aws-lambda/schema.lua
```diff
@@ -96,7 +96,6 @@ return {
   } },
   entity_checks = {
     { mutually_required = { "config.aws_key", "config.aws_secret" } },
-    { mutually_exclusive = { "config.aws_region", "config.host" } },
```
mikefero, line 99: By allowing both fields to be present will break version compatibility with older data planes. Special care will need to be made when determining which of these fields should be removed for older data planes when compatibility filters are added.

kikito: The way I define a backwards-breaking change is: the customer upgrades kong, and *does nothing else*, and things don't work as before the upgrade (a blurry definition; not the important part of this argument).

If they have an old system and they upgrade, by definition, they will be using *either* the region or the host. So if they upgrade and *do nothing else* their old data planes should not need to be protected.

On the other hand, if they upgrade *and at the same time they change their plugin configuration to have both a host and a region*, then they will receive an error on their old Dataplanes. But at that point that is not a backwards-compatibility problem, since they *changed something else* besides doing the upgrade.

kikito:
Some context for reviewers: AWS SigV4 was released in 2012. It replaces SigV2, which was deprecated [in 2020](https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/).

fffonion:
@segabond As we now target this for 3.0 with the go-with-breaking-change approach, could you add this to the changelog? (and also rebase upon master to run CI again)

kong/plugins/aws-lambda/v4.lua
```diff
@@ -154,67 +149,64 @@ local function prepare_awsv4_request(tbl)
     headers[k] = v
   end
 
-  -- if this is an AWS request, sign it
-  if region then
```
mayocream, line 158: The change will break the fix for https://github.com/Kong/kong-plugin-aws-lambda/issues/60. Probably you should leave it there.

mayocream (changes requested):
Please also include an entry in the changelog for this PR.

mayocream (approved):

kikito (approved):

PR #8087, opened 2021-11-17
perf(dao/schema) no ipairs, max page size, skip deep clone, localize vars

bungle:
### Summary

This set of performance improvements contains mainly schema and dao related enhancements:
- uses max page size when iterating over tables on initialization, rebuilds and warmups
- localize variable access in dao/schema modules
- optimizes very hot process auto fields function
- replaces ipairs with for i, #array loops

See the individual commits.

In my tests these patches together with #8010 will improve dbless reloads. E.g. router rebuild used to take something like 7 seconds per worker with my 50 MB YAML-test configuration. With these two PRs I get that to around 3-4 seconds.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -642,8 +642,13 @@ do
 
   local function build_services_init_cache(db)
     local services_init_cache = {}
+    local services = db.services
+    local page_size
+    if services.pagination then
+      page_size = services.pagination.max_page_size
+    end
 
-    for service, err in db.services:each(nil, GLOBAL_QUERY_OPTS) do
+    for service, err in services:each(page_size, GLOBAL_QUERY_OPTS) do
       if err then
         return nil, err
       end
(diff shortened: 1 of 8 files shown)


bungle:
The last commit:
https://github.com/Kong/kong/pull/8087/commits/b6bd7d42620c1b39d2bc9132f68eee07895a077a

I am not fully confident that it has no side issues, but nothing in our test suite currently catches it.

fffonion:
>  but nothing in our test suite currently catches it.

Is it possible to add some unit test that covers this?

bungle:
> > but nothing in our test suite currently catches it.
> 
> Is it possible to add some unit test that covers this?

I removed that last commit. So this can be seen mostly as refactoring for perf.

kong/runloop/handler.lua
```diff
@@ -643,7 +643,8 @@ do
   local function build_services_init_cache(db)
     local services_init_cache = {}
 
-    for service, err in db.services:each(nil, GLOBAL_QUERY_OPTS) do
+    local page_size = db.services.pagination.max_page_size
```
kikito, line 646: Shouldn't this be protected against `db.services.pagination` being `nil`? In other places you have done something equivalent to:

```
local page_size = db.services.pagination and db.services.pagination.max_page_size
```

kikito (approved):
Approved, minor comments

PR #8103, opened 2021-11-23
refactor(utils) use Lua 5.1 compatible xpcall

hbagdi:
### Summary

Use Lua 5.1 compatible `xpcall` invocation. 
This helps run a subset of Kong's functionality within a Lua 5.1-only compatible VM.

### Issues resolved

None

### DIFF
--- a/kong/tools/utils.lua
+++ b/kong/tools/utils.lua
@@ -695,7 +695,9 @@ end
 -- @return success A boolean indicating wether the module was found.
 -- @return module The retrieved module, or the error in case of a failure
 function _M.load_module_if_exists(module_name)
-  local status, res = xpcall(require, debug.traceback, module_name)
+  local status, res = xpcall(function()
+    return require(module_name)
+  end, debug.traceback)
   if status then
     return true, res
   -- Here we match any character because if a module has a dash '-' in its name, we would need to escape it.


segabond (approved):

thibaultcha:
Is the requirement to run within a PUC-Lua 5.1 environment? Otherwise, this is only a LuaJIT extension: https://luajit.org/extensions.html

hbagdi:
The interpreter being used supports a subset of PUC-Lua 5.1(few restrictions exist). These patches (along with #8104) are being proposed to minimize the number of patches that need to be maintained.

The changes are kept to a minimum in order to minimize the impact of any impact on Kong's performance.

thibaultcha:
Sure thing (in my opinion, we should have never allowed 5.2 in the codebase and `-DLUAJIT_ENABLE_LUA52COMPAT` should have never been tolerated). Heavily depending on the subset that is required to be PUC-Lua 5.1 compatible, it may be more or less difficult to achieve that compatibility due to the 5.1 LuaJIT extensions (e.g. this example) and other JIT-only ABIs (e.g. FFI).

javierguerragiraldez:
is this for the Lua-in-Go VM?  i wonder how that compares with running a LuaJIT CLI tool to do the transformation (with lots of OS-level restrictions, and maybe with JIT disabled)

hbagdi:
> is this for the Lua-in-Go VM? i wonder how that compares with running a LuaJIT CLI tool to do the transformation (with lots of OS-level restrictions, and maybe with JIT disabled)

Yes.
That is an alternative that we probably will use in some specific deployments but for most setups, the intention is to not carry Lua to avoid unnecessary complications with build, deployment, and even day-2 operations.

> Heavily depending on the subset that is required to be PUC-Lua 5.1 compatible, it may be more or less difficult to achieve that compatibility due to the 5.1 LuaJIT extensions (e.g. this example) and other JIT-only ABIs (e.g. FFI).

I should note that the amount of code that is being used in this fashion is a much smaller subset of the entire Kong codebase and so far, we have not encountered any show stoppers in getting the necessary code to run.

bungle:
I answered already here:
https://github.com/Kong/kong/pull/8104#issuecomment-977744916

Whether we should have allowed `-DLUAJIT_ENABLE_LUA52COMPAT` or not is a tricky question. It has been `on` by default for OpenResty for who knows how long. Very long if you ask me. And I am not really sure about all the deps that are they `5.1` or do they need `Lua 5.2` or do they even need `-DLUAJIT_ENABLE_LUA52COMPAT`. But as it is on by default and has been for very long, I don't think these changes are worth it.

bungle:
As these will most probably not be only incompatibilities, how far do we want to go with this? Next there is additional libs and functionaly available in LuaJIT and they are constantly adding more, like string buffers and serialization is coming. There is LuaJIT FFI that more and more our relies on…

bungle (changes requested):
We need to discuss this further.

thibaultcha:
Not as old as Kong, I remember when the change was introduced. I also remembered strongly advocating against the refactor with the use of `gotos` from a couple years ago, but what do I know 🤷🏼‍♂️

bungle:
> Not as old as Kong, I remember when the change was introduced. We made a point never to use 5.2 features for a long time. I also remember strongly advocating against the refactor with the use of `gotos` a couple of years ago, but what do I know 🤷🏼‍♂️

Yes, I was also against it when gotos first were introduced to code base. But as said, those are the easy ones, but still not worth it. I feel other upcoming compatibility issues are a way harder.

bungle (approved):

PR #8104, opened 2021-11-23
refactor(schema) remove goto statements for Lua 5.1 compat

hbagdi:
### Summary

Remove `goto` statements from the schema package for Lua 5.1 compatibility.

### Issues resolved

None

### DIFF
--- a/kong/db/schema/entity.lua
+++ b/kong/db/schema/entity.lua
@@ -43,23 +43,21 @@ function Entity.new(definition)
       return nil, entity_errors.NO_NILABLE:format(name)
     end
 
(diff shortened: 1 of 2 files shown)


bungle:
@hbagdi, can you elaborate why?

bungle:
I marked this as `pr/do not merge` as I think we need to discuss this further. Preferrably offline as this seems to relate to some effort on somewhere else that uses `Lua-in-Go VM`. For me this looks like just a start resolving incompatibility, and it will probably never end.

bungle (approved):

Issue #8093, opened 2021-11-19
Kong accepts plugins that lack required fields, but validation rejects them


seh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

Kong 2.5

### Current Behavior

Per Kong/docs.konghq.com#3368, if we create a _KongClusterPlugin_ or _KongPlugin_ manifest using an authentication-related plugin that defines a "config.hide_credentials" field in its schema, but our manifest omits that field, the validating admission Webhook rejects the proposed object as invalid—due to it lacking that field—but if we bypass the Webhook, Kong accepts the same proposed object and supplies a default value for the omitted field.

This example is about this set of plugins whose schema treats this "hide_credentials" field as required—as of Kong/kong#6785—but supplies a default value, but I suspect that the problem is more general, applying to any plugins with schema fields with this same required-but-with-default arrangement.

It would be clearer for users if either the validating admission Webhook accepts what Kong ultimately accepts, or if Kong would reject proposed _KongClusterPlugin_ and _KongPlugin_ objects that lack such a required field. In other words, the Webhook and Kong should be equally fastidious.

Why does this matter? We have _KongPlugin_ manifests that lack this required "config.hide_credentials" field, and everything was working fine for us until we put the validating admission Webhook into service. That Webhook blocks our use of these same manifests—perhaps rightly so. However, we would have preferred if Kong would have been rejecting these manifests long before we starting using the Webhook. At present, if we disable the Webhook, we can use these same manifests without any apparent problem.

### Expected Behavior

Either Kong's validating admission Webhook should accept proposed _KongClusterPlugin_ and _KongPlugin_ objects lacking fields that are required but have a default value in the schema, or Kong should reject such proposed objects because they lack this required field.

### Steps To Reproduce

```markdown
1. Write a KongPlugin manifest for a plugin like "key-auth" that lacks the "config.hide_credentials" field.
2. Disable Kong's validating admission Webhook.
3. Apply that manifest to the target cluster.
4. Confirm that Kong's proxy container accepted the plugin.
5. Delete the KongPlugin Kubernetes object.
6. Enable Kong's validating admission Webhook.
7. Apply that manifest to the target cluster.
8. Observe that the Webhook rejects the creation request.
```

### Anything else?

We see this with KIC version 2.0.5.

seh:
For now, we're working around this problem via this Kyverno policy:

<details>
[...]

Issue #7954, opened 2021-10-12
Unable to set datadog hostname or port with environment variables


chazdnato:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

2.6

### Current Behavior

When I try to use `KONG_DATADOG_AGENT_HOST`, the Datadog plugin uses the default `localhost`.

### Expected Behavior

When I use `KONG_DATADOG_AGENT_HOST`, that value should be used for the Datadog host

### Steps To Reproduce

1. Configured the plugin as follows (note that it is `host: ` to create a `null`, but the behaviour is the same if I omit the `host:` entry, or use `host: null` or `host: ""`). We are using this as a global, but the same behaviour appears when used as non-global

```
 apiVersion: configuration.konghq.com/v1
 kind: KongClusterPlugin
 metadata:
  name: datadog
  annotations:
     kubernetes.io/ingress.class: kong
   labels:
     global: "true"
 config:
   host: 
   port: 8125
   metrics:
   prefix: api-gateway
```

2. The helm chart is setup as follows, to get the latest upstream Kong with this datadog plugin enhancement, and set the env var:

```
image:
  repository: kong
  tag: "2.6"

env:
  datadog_agent_host:
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
```

3. We see that `KONG_DATADOG_AGENT_HOST` is set, and `KONG_VERSION` is correctL:

```
❯ kubectl exec -it pod/api-gateway-v1-kong-6cf47678f6-rvdpt -c proxy -n kong -- bash
$ set | grep -e DATADOG -e VERSION
bash-5.1$ set | grep -e DATADOG -e VERSION
BASH_VERSION='5.1.4(1)-release'
KONG_DATADOG_AGENT_HOST=192.168.64.26
KONG_VERSION=2.6.0
```

4. When traffic is sent, the proxy tries to send datadog metrics to `localhost`:

```
❯ kubectl logs -f pod/api-gateway-v1-kong-6cf47678f6-rvdpt -n kong -c proxy
<some logs truncated for brevity>
172.17.0.1 - - [12/Oct/2021:18:28:22 +0000] "GET / HTTP/1.1" 200 402 "-" "curl/7.64.1"
2021/10/12 18:28:22 [error] 1099#0: *3632 send() failed (111: Connection refused), context: ngx.timer, client: 172.17.0.1, server: 0.0.0.0:8000
[...]

PR #8106, opened 2021-11-24
fix: Remove the default host and port from datadog plugin schema

bakjos:
### Summary

Removes the default values from the host configuration, to allow using the environment variables when no host is specified

### Full changelog

* Updates the datadog schema

### DIFF
--- a/kong/plugins/datadog/schema.lua
+++ b/kong/plugins/datadog/schema.lua
@@ -74,8 +74,8 @@ return {
         type = "record",
         default = { metrics = DEFAULT_METRICS },
         fields = {
-          { host = typedefs.host, },
-          { port = typedefs.port, },
+          { host = typedefs.host({ default = "localhost" }), },
+          { port = typedefs.port({ default = 8125 }), },
           { prefix = { type = "string", default = "kong" }, },
           { service_name_tag = { type = "string", default = "name" }, },
           { status_tag = { type = "string", default = "status" }, },
--- a/kong/plugins/datadog/schema.lua
+++ b/kong/plugins/datadog/schema.lua
@@ -74,8 +74,8 @@ return {
         type = "record",
         default = { metrics = DEFAULT_METRICS },
         fields = {
-          { host = typedefs.host({ default = "localhost" }), },
-          { port = typedefs.port({ default = 8125 }), },
+          { host = typedefs.host, },
+          { port = typedefs.port, },
           { prefix = { type = "string", default = "kong" }, },
           { service_name_tag = { type = "string", default = "name" }, },
           { status_tag = { type = "string", default = "status" }, },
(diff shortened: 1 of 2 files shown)


kong/plugins/datadog/schema.lua
```diff
@@ -74,8 +74,8 @@ return {
         type = "record",
         default = { metrics = DEFAULT_METRICS },
         fields = {
-          { host = typedefs.host({ default = "localhost" }), },
-          { port = typedefs.port({ default = 8125 }), },
+          { host = typedefs.host, },
+          { port = typedefs.port, },
```
hbagdi, line 78: Changing defaults is a breaking change. Isn't it @bungle?

bakjos: @hbagdi I made a change to change the order the variables are loaded instead of changing the defaults, would it be better in this way?

bungle: Because the field is not required, you can force the values to `null` (explicitly), then it could fallback to `ENV`. Then you don't need to change the schema.

bakjos: @bungle I already tried this and when using the KongPlugin CRD if I assign the host to null, it's removed by the plugin and it sets the default value: `localhost`, should I use a string value instead like null or nil?

bungle: Yes, It must be that `CRD`s don't pass `null`s correctly.

bungle: @rainest do you have any insights on `null` handling with KIC and Kong?

rainest: Setting to an empty string (and possibly `~`, haven't tried) removed it in other configuration: https://github.com/Kong/kong/issues/8093#issuecomment-974571178

The new proposal (just checking for the presence of envvars first) probably makes more sense though, you shouldn't have to know to explicitly remove the defaults from your config.

hbagdi: Please avoid explicit `null` as a differentiating parameter. Reasons:
- Even though Kong can differentiate between field not being present and an explicit `null`, a lot of tooling just can't do it.
- The difference between an explicit `null` and field not being present is obvious to us because we are the maintainers. It is far too subtle for users.
- Associating a semantic meaning to empty (`null` in this case) is best avoided. For the above reason. Use enums or other schema fields. Explicit APIs help here because we capture user's intent clearly.

@bungle What would be the downside of introducing "via_env" or some such field in the schema of this plugin?

chazdnato: It looks like this is an attempted approach to fix the problems discussed in https://github.com/Kong/kong/issues/7954. I have a similar approach here: https://github.com/Kong/kong/pull/9466

Mine is also a breaking change. I can see that how the KIC/CRD interpolation works, using null/empty values leads to unexpected results in a k8s environment. 

I proposed a "via_env" style approach in my PR, would we be able to proceed in that direction?

hbagdi:
Alternate solution being pursued in #10398

PR #8032, opened 2021-11-03
feat(rate-limiting) add redis username support

27ascii:
### Summary

Adds support for configuring an username for the redis connection.

### Full changelog

- Adds the optional property redis_username used for connecting to redis. This allows to make use of redis ACL features supported with Redis 6.0

PR #8108, opened 2021-11-25
chore(rate-limiting) bump plugin version

locao:
There's a new plugin schema field to support redis username, added in #8032, so we need to bump plugin's version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,8 @@ In this release we continued our work on better performance:
   [#7979](https://github.com/Kong/kong/pull/7979)
 - Simplified the Kong core context read and writes for better performance
   [#7919](https://github.com/Kong/kong/pull/7919)
+- Reduced proxy long tail latency while reloading DB-less config
+  [#8133](https://github.com/Kong/kong/pull/8133)
 
 ### Plugins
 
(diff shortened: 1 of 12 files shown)


bungle (approved):

dndx (approved):
Do we need to add the new field into https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua?

locao:
> Do we need to add the new field into https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua?

It seems we should add it, yes. So I guess this bump will have to wait until next release, wdyt?

bungle:
> > Do we need to add the new field into https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua?
> 
> It seems we should add it, yes. So I guess this bump will have to wait until next release, wdyt?

The bump does not change the fact that change is already merged and included in `release/2.7.x`. The minor bump is fine according to this: https://github.com/Kong/kong/blob/master/kong/clustering/control_plane.lua#L469-L470

In my opinion I would just add another commit here that adds the `username` to removed fields, and include it in `2.7`. The other solution is worse where we actually now need to remove that other commit. But please find confirmation on what to do.

locao:
not releasing this at this point.

PR #8113, opened 2021-11-29
feat(services) support enabled field TDX-1642

HenriPro:
### Summary

This PR adds support for an enabled boolean field to service entity with a default value of true

When a service is `enabled = false` the proxy behavior will be as if any routes attached to it do not exist. (404) 
Note: this is different from a route with no service, which returns error:  "no Service found with those values" (503)

The primary reason for this feature is to enable easy configuration in Konnect for disabling services to allow for "archiving" and "draft" where configuration can be saved, but for the proxy to act as if it doesn't actually exist. 

EDIT: 1ba6c42 adds filtering to exporting config so when running in hybrid mode, DPs are not sent disabled services and the routes/plugins associated with them. If there is a major issue with the way I implemented that, happy to revert it out of this PR and put it in separate PR. 

As this is core entity change, moving the PR to here instead of ee.


spec/01-unit/01-db/01-schema/11-declarative_config/03-flatten_spec.lua
```diff
@@ -139,6 +140,7 @@ describe("declarative config: flatten", function()
               tls_verify_depth = null,
               tls_verify = null,
               ca_certificates = null,
+              enabled = true,
```
mikefero, line 143: Once no longer `required` these `enabled = true` can be removed.

spec/02-integration/05-proxy/02-router_spec.lua
```diff
@@ -235,6 +235,34 @@ for _, strategy in helpers.each_strategy() do
             hosts     = { "serviceless-route-http.test" },
             service   = ngx.null,
           },
+          {
+            paths      = { "/disabled-service1" },
+            protocols  = { "http" },
+            strip_path = false,
+            service    = {
+              path     = "/disabled-service-path/",
+              enabled  = false,
```
mikefero, line 244: If the C* migration suggestion is not added then a test should be added to handle the `nil` case for the `enabled` service configuration.

mikefero (changes requested):
Looks good, have a couple of suggestions.

kong/db/schema/entities/services.lua
```diff
@@ -43,6 +43,7 @@ return {
     { tls_verify         = { type = "boolean", }, },
     { tls_verify_depth   = { type = "integer", default = null, between = { 0, 64 }, }, },
     { ca_certificates    = { type = "array", elements = { type = "string", uuid = true, }, }, },
+    { enabled            = { type = "boolean", required = true, default = true, }, },
```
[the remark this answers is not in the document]
HenriPro: I added the default based on @bungle 's suggestion here https://github.com/Kong/kong-ee/pull/2749#discussion_r756844501 

From that it seems that by having both default and required fields allows the dao to fill in the value when it's nil. 
Maybe having the casandra migration is more straightforward anyways.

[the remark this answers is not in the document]
Mierenga: Having the C* teardown for defaulting to true will work for the existing rows, but I think new C* rows will not be able to get the `default` of true through `process_auto_fields` unless `required` is also true: (https://github.com/Kong/kong/blob/master/kong/db/schema/init.lua#L1506)

spec/01-unit/01-db/01-schema/11-declarative_config/03-flatten_spec.lua
HenriPro: I think most of these have to remain as it's checking the result of the declarative config, and the default=true will attach enabled=true.

kong/db/migrations/core/014_230_to_270.lua
```diff
@@ -0,0 +1,20 @@
...
+      ]]
+    },
+  
+    cassandra = {
+      up = [[
+        ALTER TABLE services ADD enabled boolean;
+      ]]
+    },
+  }
+
```
HenriPro, line 20:
```suggestion
  
  
```

HenriPro, line 20:
```suggestion
 
```

kong/db/declarative/init.lua
```diff
@@ -375,7 +375,7 @@ function declarative.load_into_db(entities, meta)
 end
 
 
-local function export_from_db(emitter, skip_ws)
+local function export_from_db(emitter, skip_ws, skip_disabled_services)
```
bungle, line 378:
```suggestion
local function export_from_db(emitter, skip_ws, skip_disabled_entities)
```

I would prefer to call it `skip_disabled_entities`. After this PR we have at least `plugins` and `services` that have `enabled` field.

kong/db/declarative/init.lua
```diff
@@ -404,23 +405,33 @@ local function export_from_db(emitter, skip_ws)
...
         return nil, err
       end
 
-      for _, foreign_name in ipairs(fks) do
-        if type(row[foreign_name]) == "table" then
-          local id = row[foreign_name].id
-          if id ~= nil then
-            row[foreign_name] = id
+      -- do not export disabled services and associated enitties when skip_disabled_services
+      if skip_disabled_services and name == "services" and not row.enabled then
```
bungle, line 416:
```suggestion
      if skip_disabled_entities and (name == "services" or name == "plugins") and not row.enabled then
```

Just a suggestion, but I think you get the idea.

I know this is feature creep, but while we are at it, it probably makes sense to deal with plugins too. Plugins are not referenced by any other table, so plugins don't need same treatment as services that related entities should be skipped too.

bungle (changes requested):

kong/db/declarative/init.lua
HenriPro: done

Mierenga (approved):
Verified migrations and proxy functionality locally with Postgres and C*

mikefero (approved):
Nice work @HenriPro; any little nickel and dime things can be fixed later. Thanks for adding all the additional features too :partying_face:

PR #8129, opened 2021-11-30
feat(pdk) add kong.request.get_normalized_path

flrgh:
This adds a PDK method for retrieving the normalized path component of the current downstream http request.

Most of the test cases were copy-pasted from `kong.request.get_path`, with a couple added to validate the correct behavior.

----
Internal tracking
[FT-2116]

[FT-2116]: https://konghq.atlassian.net/browse/FT-2116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -8,6 +8,7 @@
 local cjson = require "cjson.safe".new()
 local multipart = require "multipart"
 local phase_checker = require "kong.pdk.private.phases"
+local normalize = require("kong.tools.uri").normalize
 
 
 local ngx = ngx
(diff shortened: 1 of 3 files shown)


kong/pdk/request.lua
```diff
@@ -385,6 +386,34 @@ local function new(self)
...
+  -- @treturn string the path
+  -- @usage
+  -- -- Given a request to https://example.com/t/Abc%20123%C3%B8%2f/parent/..//test/./
+  --
+  -- kong.request.get_normalized_path()     -- "/t/Abc 123ø%2F//test/"
+  -- kong.request.get_normalized_path(true) -- "/t/Abc 123ø%2F/test/"
+  function _REQUEST.get_normalized_path(merge_slashes)
+    local path = _REQUEST.get_path() or ""
+    return normalize(path, merge_slashes)
+  end
```
bungle, line 415: A few things:
1. `_REQUEST.get_path()` does not return `false` or `nil` so `or ""` is not needed.
2. `merge_slashes` is questionable, and if we have that it should be opposite `do_not_merge_slashes` as merging them is what people in 99% cases want. That is also what we do in our router.

bungle (changes requested):

bungle:
Do we also need `get_normalized_forwarded_path()`? :-)

kong/pdk/request.lua
flrgh: Agreed, I'll make `merge_slashes` the default behavior.

flrgh: Updated. I've kept the parameter name the same (`merge_slashes`) but changed the behavior to be truth-y by default. I have a _slight_ preference for this as opposed to inverting the param to `do_not_merge_slashes`, but I don't mind switching it to `do_not_merge_slashes` with a false-y default if you feel like it's too clunky this way.

flrgh:
> Do we also need `get_normalized_forwarded_path()`? :-)

Maybe. I stopped short of doing that because I didn't see any plugins that were using get_forwarded_path (oops, okay it looks like there's at least one EE plugin that uses it).

There's been some discussion on the internal ticket about simply changing the behavior of `get_path()` to normalize the path and introducing a `get_original_path()/get_request_path()` method that would forgo normalization. I don't have too strong of an opinion on this myself--what do you think?

kong/pdk/request.lua
tyler-ball: +1 to not using negative variable names

tyler-ball (approved):
👍 to the code and tests. Someone more experienced than I should give feedback about changing the default behavior of `get_path` to be normalized

kong/pdk/request.lua
Tieske: I agree with @bungle and would prefer the negative name here since it would stick better to the truthy/falsy behaviour expected in Lua. If one reads `merge_slashes` then most of us would expect the default to be `false`. But by no means want to turn this into a 🔥 🔥 🔥 🔥 🔥 🔥  😄

bungle:
> 👍 to the code and tests. Someone more experienced than I should give feedback about changing the default behavior of `get_path` to be normalized

Yes, that is a good question: `get_path` + `get_raw_path` vs. `get_path` + `get_normalized_path`.

bungle:
There is another concern. Our `utils.normalize` is more than just `normalize`, we need to use different normalization for this.

flrgh:
Looks like the plan is to make `get_path`/`get_forwarded_path` normalized and introduce new PDK methods that feature the old behavior (returning the raw/non-normalized value). I'm closing this and will follow up with a new PR for those changes when they're ready.

PR #8133, opened 2021-12-01
perf(dbless) improve proxy long tail latency for DB-less reload [CT-321]

dndx:
Previously, when a large DB-less config is processed by Kong, Kong will
be CPU bound when parsing/validating the config and rebuilding the router.
This could cause undesirable proxy latency behavior during reloads.

This PR introduces periodic yielding points so that reload tasks no
longer hogs the CPU until reload completely finishes. Request processing
will be able to continue even while DB-less reload and router rebuild is
happening.

CT-321

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -14,6 +14,7 @@ local SHADOW = true
 local md5 = ngx.md5
 local pairs = pairs
 local ngx_socket_tcp = ngx.socket.tcp
+local yield = require("kong.tools.utils").yield
 local REMOVE_FIRST_LINE_PATTERN = "^[^\n]+\n(.+)$"
 local PREFIX = ngx.config.prefix()
 local SUBSYS = ngx.config.subsystem
@@ -257,6 +258,8 @@ function Config:parse_table(dc_table, hash)
   if err_t then
     return nil, pretty_print_error(err_t), err_t
   end
+
+  yield()
 
   if not self.partial then
     self.schema:insert_default_workspace_if_not_given(entities)
@@ -587,6 +590,8 @@ function declarative.load_into_cache(entities, meta, hash, shadow)
   local transform = meta._transform == nil and true or meta._transform
 
   for entity_name, items in pairs(entities) do
+    yield()
+
     local dao = kong.db[entity_name]
     if not dao then
       return nil, "unknown entity: " .. entity_name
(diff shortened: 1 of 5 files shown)


kong/db/declarative/init.lua
```diff
@@ -587,6 +590,8 @@ function declarative.load_into_cache(entities, meta, hash, shadow)
   local transform = meta._transform == nil and true or meta._transform
 
   for entity_name, items in pairs(entities) do
+    yield()
```
bungle, line 593:
```suggestion
    yield(true)
```

kong/db/declarative/init.lua
```diff
@@ -786,6 +793,8 @@ function declarative.load_into_cache(entities, meta, hash, shadow)
   end
 
   for tag_name, tags in pairs(tags_by_name) do
+    yield()
```
bungle, line 796:
```suggestion
    yield(true)
```

kong/router.lua
```diff
@@ -1418,12 +1422,16 @@ function _M.new(routes)
...
 
+  yield()
+
   -- the number of categories to iterate on for this instance of the router
   local categories_len = #categories_weight_sorted
 
   sort(prefix_uris, sort_uris)
 
   for _, category in pairs(categories) do
+    yield()
```
bungle, line 1433:
```suggestion
    yield(true)
```

bungle (changes requested):

kong/tools/utils.lua
```diff
@@ -1423,4 +1426,25 @@ local topological_sort do
...
+          ngx_sleep(0)
+        end
+
+        return
+      end
+
+      ngx_sleep(0)
+    end
+  end
+end
```
bungle, line 1447:
```suggestion
  function _M.yield(in_loop)
    local phase = get_phase()
    if phase == "init" or phase == "init_worker"  then
      return
    end
    if in_loop then
      counter = counter + 1
        if counter % YIELD_ITERATIONS ~= 0 then
          return
        end
        counter = 0
      end
      ngx_sleep(0)
    end
  end
end
```

bungle (changes requested):

kong/db/declarative/init.lua
dndx: Correct me if I'm wrong, but `for entity_name, items in pairs(entities)` only iterates known entities by entity name, it does not represents individual entity objects.

dndx: Same idea for the following ones, they all have small loops.

bungle (approved):

PR #8136, opened 2021-12-01
fix(declarative) change default back to false skip_ws

HenriPro:
### Summary

declarative.export_from_db skip_ws was inadvertently changed to true in https://github.com/Kong/kong/pull/8113
 

### Full changelog

declarative.export_from_db skip_ws default set back to false

### Issues resolved

Tests break in ee without this.

Fix #XXX


HenriPro:
Closed as I think this intentional difference between CE and EE.

PR #8139, opened 2021-12-02
fix(router) properly escape uri captures

### LINKED ISSUE #7913: uri_captures are url decoded

bungle:
### Summary

When we added normalization `kong.tools.uri.normalize`, that function does percent-decoding on everything, except for the reserved characters.

That means that we basically percent-decode more than just the ranges of ALPHA (%41–%5A and %61–%7A), DIGIT (%30–%39), hyphen (%2D), period (%2E), underscore (%5F), or tilde (%7E).

Which is fine for matching (as we run both inputs through it). Though calling it `normalize` is wrong, but that is another issue.

Because of doing excessive percent-decoding, we endup things leaking from there. One of the leakeage is uri captures that now contain the captures based on our raw normalization.

This commit adds escaping to uri captures, and should fix the leak.

Alternative to this is to not percent-decode the chars outside the ones specified above.

### Issues resolved

Fix #7913, FTI-2904

### Alternative and Better Implementation

See #8140

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -7,6 +7,7 @@ local bit           = require "bit"
 
 local hostname_type = utils.hostname_type
 local normalize     = require("kong.tools.uri").normalize
+local escape        = require("kong.tools.uri").escape
 local subsystem     = ngx.config.subsystem
 local get_method    = ngx.req.get_method
 local get_headers   = ngx.req.get_headers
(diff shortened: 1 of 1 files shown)


bungle:
I personally like #8140 better.

PR #8140, opened 2021-12-02
fix(tools.uri) normalization decodes as much as possible

### LINKED ISSUE #7913: uri_captures are url decoded

bungle:
We decide to let `normalize` function to decode URL-encoded string as much as possible.

**PLEASE REFERER TO**: https://github.com/Kong/kong/pull/8140#issuecomment-1162850953

### Issues resolved

Fix #7913, FTI-2904

Outdated discussion:

----------------------

This is alternative to PR #8139 where we actually fix the normalization function to not do excessive percent-decoding on normalization.

When we added normalization kong.tools.uri.normalize, that function does percent-decoding on everything, except for the reserved characters.

That means that we basically percent-decode more than just the ranges of ALPHA (%41–%5A and %61–%7A), DIGIT (%30–%39), hyphen (%2D), period (%2E), underscore (%5F), or tilde (%7E). (so called Unreserved Characters)

Alternative Implementation: See #8139


flrgh:
Hmm. RFCs are always a chore to interpret. It seems like the original mistake was that normalization was implemented with the assumption that `unreserved chars == (all chars - reserved chars)`, which, as you know, turns out to be incorrect (according to the RFC).

IMO instead of over-decoding and re-encoding we should just fix the logic to be more selective in our decoding, according to the chars in [section 2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3):

> unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

I put a draft/poc of this in another branch (fix/uri-normalize-unreserved 6313dd22f38fb1d9678026bc76fe0494a1bc3d1a). Maybe take a look and tell me what you think?

bungle:
> Hmm. RFCs are always a chore to interpret. It seems like the original mistake was that normalization was implemented with the assumption that `unreserved chars == (all chars - reserved chars)`, which, as you know, turns out to be incorrect (according to the RFC).
> 
> IMO instead of over-decoding and re-encoding we should just fix the logic to be more selective in our decoding, according to the chars in [section 2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3):
> 
> > unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
> 
> I put a draft/poc of this in another branch (fix/uri-normalize-unreserved [6313dd2](https://github.com/Kong/kong/commit/6313dd22f38fb1d9678026bc76fe0494a1bc3d1a)). Maybe take a look and tell me what you think?

@Michael Martin yes, we discussed this with @Datong. The thing is that I am not sure which one is better. More selective percent decoding or not. I am not sure can there ever be things already percent decoded in ngx.var.request_uri or not. Do you? If there could be, then those need to be normalized back to percent encoded form. Similar question goes to route.paths.

Also the escaping includes % in non escape chars which is rather questionable.

bungle:
If someone wants to work or try different approaches, here is a list that need to be taken in account:

definitions:
reserved chars: `! * ' ( ) ; : @ & = + $ , / ? % # [ ]`
unreserved: `- _ . ~`
alphanumeric: `a b c d e f g h i j k l m n o p q r s t u v w x y z A B C D E F G H I J K L M N O P Q R S T U V W X Y Z 0 1 2 3 4 5 6 7 8 9`
other: * (meaning any char that is not mentioned above)

dot processing:
`/.` AND `/./`
`/..` AND  `/../`
`/%2e` AND `/%2e/`
`/.%2e` AND `/.%2e/`
`/.%2E` AND `/.%2E/`
`/%2e%2E` AND `/%2e%2E/`
etc.

merge slashes (this is generally thought to be good thing to do, but it might change semantics in some cases):
`//`
`/a//b`
etc.

given this:
`/ä/a/%2e./a%2E//a/%2e/./a/../a/%2e%2E/%5f%99%af%2f%2F`

should normalize (I think) to this:
`/%C3%A4/a./a/_%99%AF%2F%2F`

What happens there?
`ä` -> being percent encoded to `%C3%A4`
`%2e` -> being uppercased to `%2E`
`//` -> slashes being merged
`%5f` -> being percent decoded to `_`
`/../` -> dot segment being processed
`/%2e/` -> dot segment being processed
`/a%2E/` -> dot being percent decoded to `/a./`

In general:
- `unreserved` and `alphanumeric` should be percent decoded
- `reserved` should be kept as is (no percent decoding, no percent encoding) BUT percent encoded forms should be turned to uppercase
- rest should be percent encoded

Then you need to implement dot processing.

Current issue with kong.tools.uri.normalize:
- it over percent decodes
- it does not percent encode

kikito (dismissed):

flrgh:
Pinged @dndx about this in a side channel, but I think the remaining unsolved item is what to do with input that contains raw unicode that is not percent-encoded. I believe this is what @bungle was trying to address by wrapping everything with `escape()`.

Let's take `/ö` and `/%2F%C3%B6`, for instance. My opinion is that, from the router's perspective, these two paths should be treated as equivalent (though, enforcing this behavior has some interesting potential for breakage since today `Route.paths` will happily accept both as distinct inputs). Is that something we need to also to address in this PR (actually I'm not even sure if we have consensus on that being the "correct" behavior)?

kong/tools/uri.lua
```diff
@@ -44,20 +31,38 @@ local SLASH = string_byte("/")
...
+local function escape(uri)
+  if ngx_re_find(uri, ESCAPE_PATTERN, "joi") then
+    return ngx_re_gsub(uri, ESCAPE_PATTERN, percent_escape, "joi")
+  end
+
+  return uri
+end
+
+
+local function normalize(uri, merge_slashes)
```
bungle, line 65: This function does not escape anymore.

bungle (commented):
Normalization does not escape, as @flrgh said above.

kong/tools/uri.lua
bungle: This issue can be illustrated with curl:
```bash
~ ❯ curl http://localhost:8080/uri/ö      
<html><body>/uri/ö</html>
```

where I have this location block:
```
location /uri {
  content_by_lua '
    ngx.say("<html><body>", ngx.var.request_uri ,"</html>")
  ';
}
```

See, the `ö` is unescaped, but normalization function has to escape it properly.

bungle:
> My opinion is that, from the router's perspective, these two paths should be treated as equivalent

I share this view.

> though, enforcing this behavior has some interesting potential for breakage since today `Route.paths` will happily accept both as distinct inputs). Is that something we need to also to address in this PR (actually I'm not even sure if we have consensus on that being the "correct" behavior)?

We run the normalization both to `route.paths` and to `ngx.var.request_uri`. If you have both forms in your DB:
```
/ö
```
and:
```
/%C3%B6
```
or even:
```
/%C3%b6
```

these should turn the path to:
```
/%C3%B6
```

And similarly we do it in `ngx.var.request_uri`. If people used this to mean different paths, it is breaking. But I consider it a fix rather than feature. So it was not working properly in a first place, and people who relied on it, will have to adapt. The change is for better in my opinion, even with a slight change for breakage on configs that rely on this undocumented bug.

dndx:
@flrgh @bungle In order to make our internal behavior less confusing, I proposed during the last meeting that we keep everything internal to Kong non-normalized and escape the path right before we send it to the upstream. Reason being dealing with non-encoded paths internally makes most sense from a API perspective, and it avoids any confusion about whether something should be percent encoded or not.

So in case of `/uri/ö`, everything that uses the URI within Kong should see the utf-8 character as-is. Let's say if we have a request transformer config that transform the URI into `/foo/ö`, the request transformer should never need to know whether it needs to escape the utf-8 character, and the user, when using "ö" inside their config, should not need to lookup the hex representation of it.

Then, before Kong sends the request to the upstream, it encodes any character that is not reserved and not standard ASCII. For example, it will encode `ö` to  `%C3%B6` but will not encode `/` (as it is a reserved character).

IMO this behavior is a lot less confusing from a user and plugin's perspective. We ensure a unified representation when dealing with the URI within Kong.

With that, it seems that maybe https://github.com/Kong/kong/pull/8140/files#diff-093344a0609553c69b475fb5cbd57a202796383eb6fca553d26ac6a1f40d4492L1353 need to be reverted back so the behavior is actually correct afterwards.

bungle:
> With that, it seems that maybe https://github.com/Kong/kong/pull/8140/files#diff-093344a0609553c69b475fb5cbd57a202796383eb6fca553d26ac6a1f40d4492L1353 need to be reverted back so the behavior is actually correct afterwards.

Instead of doing `escape` there, should we do `normalize` instead. And if so, I think we need to escape on normalize.

kong/tools/uri.lua
```diff
@@ -44,20 +31,38 @@ local SLASH = string_byte("/")
...
-local function escape(m)
+local function percent_escape(m)
   return string_format("%%%02X", string_byte(m[0]))
 end
 
 
-function _M.normalize(uri, merge_slashes)
+local function escape(uri)
+  if ngx_re_find(uri, ESCAPE_PATTERN, "joi") then
+    return ngx_re_gsub(uri, ESCAPE_PATTERN, percent_escape, "joi")
```
Suika-Kong, line 58:
```suggestion
    return (ngx_re_gsub(uri, ESCAPE_PATTERN, percent_escape, "joi"))
```

You will get an extra return value, and that may cause problems when you do this: `print(escape(uri))`.

Suika-Kong (dismissed):
I personally prefer Datong's idea that we should implement normalization with `others`(neither preserved nor non-preserve characters) unescaped, so we will get URIs more intuitive and readable for humans, and easier to process with regex(for example Chinese characters can be treated as single characters instead of a bunch of bytes).

However, this is a PR that should be merged soon. We could figure that out after fixing the problem we are facing.

Only minor suggestions. But I strongly recommend adding comments to `kong/runloop/handler.lua:1396`, so when we decide not to escape when normalizing, we will not forget to add `escape` call back.

kong/runloop/handler.lua
```diff
@@ -1394,7 +1393,7 @@ return {
       --       router, which might have truncated it (`strip_uri`).
       -- `host` is the original header to be preserved if set.
       var.upstream_scheme = match_t.upstream_scheme -- COMPAT: pdk
-      var.upstream_uri    = escape(match_t.upstream_uri)
+      var.upstream_uri    = match_t.upstream_uri
```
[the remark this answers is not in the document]
Suika-Kong: And we should not forget to add it back if we finally decide not to escape when normalizing.

kong/tools/uri.lua
Suika-Kong: Also, we need to notice that `escape` is not what it literally means. It's more like `safe_encoding`.

Suika-Kong:
Also, be noticed that we only handle percent encodings, but not `others` in regex, so if someone writes a regex router path like `"\中文"`, it will never match.

Suika-Kong:
The key points that people argue with are:
1. Normalization regex is not practical, so we need a standard(normalized) format for users to match with, and that is an interface detail;
2. As we decide to unify the format of URI of all uses in kong, the plugins (especially external plugins) need to deal with the returned value, and it may contain characters with unknown or even illegal encoding (or even a binary string, containing characters including `"\0"`)

The "normalized format" should also serve as a standard format and a unified format.

I've discussed `#2` with @fffonion offline, and we agree that it's fine for external plugins to have `bytes` as the return type for `get_uri`.

Suika-Kong:
Tested with Nginx:
```nignx
location ~ ^/id/(?<id>\S+)$ {
    # use variable $myvar here
    return 200 "$id";
}
```
We got 404 for `/id/test%20space`, and 200 with message `..` for `/id/%2e%2E`.

So, this solution is consistent with Nginx.

Suika-Kong:
- [x] Correct normalization behavior
- [x] Apply normalization to all URI-related interfaces
- [ ] Document normalization process to URI-related interfaces
- [ ] Stop percent decoding regex path

spec/01-unit/18-tools_uri_spec.lua
```diff
@@ -48,11 +48,11 @@ describe("kong.tools.uri", function()
     end)
 
     it("does not decode non-ASCII characters that are unreserved, issue #2366", function()
```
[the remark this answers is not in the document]
Suika-Kong: I check PRs from https://github.com/pulls and fails to track this PR...

kong/tools/uri.lua
```diff
@@ -12,52 +9,108 @@ local table_concat = table.concat
...
+local A_BYTE = string_byte('a')
+local Z_BYTE = string_byte('z')
+local ZERO_BYTE = string_byte('0')
+local NINE_BYTE = string_byte('9')
+
+local CHAR_RESERVED = true
+local CHAR_UNRESERVED = false
+local CHAR_OTHERS = nil -- luacheck: ignore
+
+local char_urlencode_type = table_new(256, 0) do
```
dndx, line 45:
```suggestion
local chars_to_decode = table_new(256, 0)
do
```

Is probably better, because when the value is `true`, character should be decoded.

kong/tools/uri.lua
```diff
@@ -12,52 +9,108 @@ local table_concat = table.concat
...
+
+  for num = ZERO_BYTE, NINE_BYTE do
+    char_urlencode_type[num] = CHAR_UNRESERVED
+  end
+
+  for _, num in ipairs{
+    HYPHEN_BYTE, DOT_BYTE, UNDERSCORE_BYTE, TILDE_BYTE,
+  } do
+    char_urlencode_type[num] = CHAR_UNRESERVED
+  end
```
dndx, line 68: I actually think just writing:

```suggestion
    char_urlencode_type[HYPHEN_BYTE] = CHAR_UNRESERVED
    char_urlencode_type[DOT_BYTE] = CHAR_UNRESERVED
```

is better. It is even less lines of code.

kong/tools/uri.lua
```diff
@@ -12,52 +9,108 @@ local table_concat = table.concat
...
+local CAP_A_BYTE = string_byte('A')
+local CAP_Z_BYTE = string_byte('Z')
+local A_BYTE = string_byte('a')
+local Z_BYTE = string_byte('z')
+local ZERO_BYTE = string_byte('0')
+local NINE_BYTE = string_byte('9')
+
+local CHAR_RESERVED = true
+local CHAR_UNRESERVED = false
+local CHAR_OTHERS = nil -- luacheck: ignore
```
dndx, line 43: A comment here is fine, no need to mess with the `luacheck` rules.

chronolaw (approved):

Suika-Kong (approved):

kong/tools/uri.lua
```diff
@@ -12,52 +9,107 @@ local table_concat = table.concat
...
+local A_BYTE = string_byte('a')
+local Z_BYTE = string_byte('z')
+local ZERO_BYTE = string_byte('0')
+local NINE_BYTE = string_byte('9')
+
+local CHAR_RESERVED = true
+local CHAR_UNRESERVED = false
+-- nil for CHAR_OTHERS
+
+local chars_to_decode = table_new(256, 0)
```
flrgh, line 46: After building this table, we only use it to check for reserved chars (in `normalize_decode`). Seems like maybe this change started out with the intent to use `chars_to_decode` for multiple purposes but wound up limiting its use in the end?

Is there a follow-up changeset/PR that will use `chars_to_decode` and rely on this difference in `CHAR_RESERVED` vs `CHAR_UNRESERVED`? If not, do you think you could clean this up a little so that there's less in here to distract from the actual use-case for this table?

flrgh (commented):
Code-wise, I reviewed and left a couple nitpicks. The `table.new` thing should probably be fixed, but the other comment on `chars_to_decode` is mostly just a readability gripe, so not a blocker if anyone disagrees with me about it.

Behavior-wise, there's been a whole bunch of activity on the path handling/normalization discussion since I last participated, so as long as everyone else here is in agreement about this change, it gets my :+1:.

kong/tools/uri.lua
Suika-Kong: We were debating whether to encode `CHAR_OTHERS`. And there was also a solution to allow users to pass in an argument to tweak the behavior. I prefer to at least leave a comment to explain.

PR #8146, opened 2021-12-03
fix(schemas) add transformations and shorthand_fields support to subschemas

bungle:
### Summary

- fix(schemas) add transformations support to subschemas
- fix(schemas) add shorthand_fields support to subschemas

The transformations and shorthand fields could not be applied to subschemas. This PR will make them available in subschemas too.

### DIFF
--- a/kong/db/schema/metaschema.lua
+++ b/kong/db/schema/metaschema.lua
@@ -360,7 +360,66 @@ local nested_attributes = {
 
 local check_field
 
+local function has_schema_field(schema, name)
+  if schema == nil then
+    return false
+  end
+
+  local dot = string.find(name, ".", 1, true)
+  if not dot then
+    for _, field in ipairs(schema.fields) do
+      local k = next(field)
+      if k == name then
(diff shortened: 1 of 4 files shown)


kong/db/schema/metaschema.lua
```diff
@@ -595,6 +640,9 @@ local MetaSchema = Schema.new({
     {
       shorthand_fields = shorthand_fields_array,
     },
+    {
+      transformations = transformations_array,
+    },
```
kikito, line 645: This move is unnecessary, strictly speaking. If possible, remove it, just to make it easier to merge on ee later. No biggie though.

kikito (approved):
I have very minor nitpicks, none of them really prevent a merge, so I approve this.

kong/db/schema/metaschema.lua
bungle: this was for style, so that we have them in both schema and subschema the same and it readable order.

PR #8150, opened 2021-12-07
feat(zipkin) include http path to span name

yankun-li-kong:
Add a new parameter span_include_path to decide whether to include http path to span name.
Add test case.

### Summary

This PR is based on https://github.com/Kong/kong-plugin-zipkin/pull/124, Thanks @XinweiLiu28 
Able to include HTTP path to span name for zipkin plugin

### Full changelog

Add a new parameter `span_include_path` to decide whether to include HTTP path to span name.
Add test case.

### Issues resolved

Fix [FTI-2842](https://konghq.atlassian.net/browse/FTI-2842)


kong/plugins/zipkin/handler.lua
```diff
@@ -127,9 +127,15 @@ if subsystem == "http" then
       trace_id = rand_bytes(conf.traceid_byte_count)
     end
 
+    local span_name = method
+    local path = req.get_path()
+    if conf.span_include_path == true then
```
mayocream, line 132:
```suggestion
    if conf.span_include_path then
```

kong/plugins/zipkin/schema.lua
```diff
@@ -61,6 +61,7 @@ return {
           { tags_header = { type = "string", required = true, default = "Zipkin-Tags" } },
           { static_tags = { type = "array", elements = static_tag,
                             custom_validator = validate_static_tags } },
+          { span_include_path = { type = "boolean", default = false } },
```
mayocream, line 64: IMO, `span_name_include_path` is better.

mayocream:
I think it is good to have the endpoint in Span name, see OpenTelemetry spec [opentelemetry-specification/api.md at main · open-telemetry/opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span)

<p data-renderer-start-pos="82" style="margin: 0.75rem 0px 0px; padding: 0px; font-size: 14px; line-height: 1.714; font-weight: 400; letter-spacing: -0.005em; color: rgb(23, 43, 77); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, Oxygen, Ubuntu, &quot;Fira Sans&quot;, &quot;Droid Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: pre-wrap; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">For example, here are potential span names for an endpoint that gets a hypothetical account information:</p><div class="pm-table-container  drwdyq-0 dugMha" data-layout="default" style="margin: 0px auto 16px; padding: 0px; position: relative; box-sizing: border-box; z-index: 0; transition: all 0.1s linear 0s; display: flex; clear: both; color: rgb(23, 43, 77); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, Oxygen, Ubuntu, &quot;Fira Sans&quot;, &quot;Droid Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: pre-wrap; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; width: inherit;"><div class="pm-table-wrapper" style="margin: 0px; padding: 0px; overflow-x: auto;">

Span Name | Guidance
-- | --
get | Too general
get_account/42 | Too specific
get_account | Good, and account_id=42 would make a nice Span attribute
get_account/{accountId} | Also good (using the "HTTP route")

</div></div>

kong/plugins/zipkin/schema.lua
mayocream:
```suggestion
          { span_name_include_path = { type = "boolean", default = false } },
```

kong/plugins/zipkin/schema.lua
```diff
@@ -61,6 +61,7 @@ return {
           { tags_header = { type = "string", required = true, default = "Zipkin-Tags" } },
           { static_tags = { type = "array", elements = static_tag,
                             custom_validator = validate_static_tags } },
+          { span_name_include_path = { type = "boolean", default = false } },
```
kikito, line 64: Hi, I propose the following change to the field:

```
          { http_span_name = { type = "string", required = true, default = "method", one_of = { "method", "method_path" } } },
```

Reasons for the changes:
* A string (enum) instead of a boolean: With a boolean we "lock" ourselves to two options. With an enum we leave the door open to other span names in the future. The field name had to be changed to reflect this as well.
* `http_` prefix: The path will only be available on http requests. I think that detail is fundamental enough to be part of the field name.

kikito (changes requested):

kong/plugins/zipkin/schema.lua
mayocream: Thanks! Using the enum type in the schema is more flexible and extensible.

kikito (approved):

kikito:
This looks ok to me. Could you please include a line on the changelog before merging it?

PR #8156, opened 2021-12-08
fix(admin) fix service entity validation for tls options

### LINKED ISSUE #8151: Service TLS options only available for https (not grpcs and tls)

2.6.0

### Current Behavior

Currently the service entity validation prevents TLS related configuration options from being set unless the protocol is https. It seems like this is an oversight from when the only supported protocols were http and https. Needless to say these config options are just as applicable for grpcs and tls (stream).

Relevant piece of code:
https://github.com/Kong/kong/blob/323e2cf6c53ae5604c34d716a3ce2cd87d86c948/kong/db/schema/entities/services.lua#L55-L70

The change would be trivial to implement and I could throw in a PR for it, but I wanted to check that I wasn't missing some reasoning behind the current logic first.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

nvx:
### Summary

Fix service entity validation for tls options.

Unfortunately I'm still fighting with gojira to get a development version up and running, building and linting is fine, but actually starting it after that seems to have some issues so I haven't been able to test it properly.

### Full changelog

The service entity validation is currently too strict and does not
allow specifying tls options for the non-https tls protocols grpcs
and tls (stream).

This updates the validation to allow specifying the service options
client_certificate, tls_verify, tls_verify_depth, and ca_certificates
for grpcs and tls (stream) protocols in addition to https.

Note this does not add it for tls_passthrough as I don't believe
it's applicable for that stream mode.

### Issues resolved

Fix #8151


nvx:
Oh hey got gojira to compile, sadly while it does now enable setting of the option, setting tls_verify to true doesn't seem to have any effect so will have to do some more digging to make the options take effect it seems.

PR #8159, opened 2021-12-08
chore(metrics): use encrypted metrics transmission

jschmid1:
### Summary

Use & prioritize TLS port for phone-home capability

### Full changelog

* Use & prioritize TLS port for phone-home capability

Notes:

I disabled two tests as they were inadvertently sending metrics to our metrics instance (kong-hf.kong.com). This behavior existed _before_ this patch.

- "db config db_import does not require Kong to be running"

Reason: 
They do not start kong and therefore do not take a patched dnshosts file that would redirect kong-hf.kong.com to localhost.

I also disabled this test:

* stream reports tcp streams
* stream reports tls streams

I'm not seeing how they every should've passed in the first place. 

Creating a ticket to track this. https://github.com/Kong/kong/issues/8160

Also got rid of `helpers.mock_reports_server` as it's close to 1:1 copy of `helpers.tcp_server`


kong/reports.lua
```diff
@@ -143,6 +157,15 @@ local function send_report(signal_type, t, host, port)
   if not ok then
     return
   end
+  if tls then
+    local opts = {
+      ssl_verify = false,
+    }
+    local res, err = sock:tlshandshake(opts)
+    if not res then
+      return error("Failed to complete TLS handshake. ", err)
```
bungle, line 166: I am not sure should this error. We have had issues before when anonymous logging has been turned on, and then that feature just started to crash whole Kong. That's why we should not fail here. People might be running Kong in internal networks with no internet connectivity with anonymous reports on.

kong/reports.lua
```diff
@@ -102,16 +102,30 @@ end
...
-
+  local tls
   t = t or {}
   host = host or constants.REPORTS.ADDRESS
   port = port or constants.REPORTS.STATS_PORT
+  tls = false
+  -- keep backwards compatibility and introduce new port attribute for TLS
+  local tls_port = constants.REPORTS.STATS_TLS_PORT
+  -- the presence of a tls_port indicates we have a TLS connection
+  if tls_port then
```
bungle, line 119: Isn't this always present? Or is this also for `testing`?

bungle (changes requested):

kong/reports.lua
jschmid1: sounds reasonable. removing

kong/reports.lua
```diff
@@ -143,6 +157,12 @@ local function send_report(signal_type, t, host, port)
   if not ok then
     return
   end
+  if tls then
+    local opts = {
+      ssl_verify = false,
+    }
+    sock:tlshandshake(opts)
```
dndx, line 164: Is there any reason that's preventing us from enabling TLS verification?

Also, if `tlshandshake` fails, we should probably bail out and not continue anymore.

dndx (commented):
Is `spec/fixtures/go/go-hello` committed by mistake?

kong/constants.lua
```diff
@@ -140,7 +140,8 @@ local constants = {
   },
   REPORTS = {
     ADDRESS = "kong-hf.konghq.com",
-    STATS_PORT = 61830
+    STATS_PORT = 61830,
```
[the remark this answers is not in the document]
jschmid1: I was just very careful about backwards compatibility. If we can be certain that we never maintain this `REPORTS` table outside of the shipped package we can probably drop it.

kong/reports.lua
jschmid1: >  Also, if tlshandshake fails, we should probably bail out and not continue anymore.

see -> https://github.com/Kong/kong/pull/8159#discussion_r764821364

jschmid1:
> Is spec/fixtures/go/go-hello committed by mistake?

yes absolutely. removing

kong/reports.lua
jschmid1: > Is there any reason that's preventing us from enabling TLS verification?

I discussed this with @hutchic and we took the safe route as we can't be _always_ sure that certificates are up to date on the host that runs this code

bungle: Should it bailout if handshake fails?

hutchic (approved):
approval is acknowledging we tested this together and it works 👍

bungle (approved):

kong/reports.lua
dndx: +1 on @bungle , continuing after `sock:tlshandshake` fails seems to be a bad idea.

dndx: Also, for the frequency of reports we send, does enabling TLS session reuse for this connection makes sense? If so it would be a simple change on the call here. Ping @hutchic

kong/reports.lua
```diff
@@ -143,6 +142,15 @@ local function send_report(signal_type, t, host, port)
...
     return
   end
+  local opts = {
+    ssl_verify = false,
+  }
+  local hs_ok, err = sock:tlshandshake(opts)
+  if not hs_ok then
+    local msg = "Failed to complete TLS handshake" .. err
+    log(WARN, msg)
+    return msg
```
dndx, line 152:
```suggestion
    log(DEBUG, "failed to complete TLS handshake for reports: ", err)
    return
```

Also add `local DEBUG = ngx.DEBUG` above.

I don't think we need to be that noisy when phone home fails as there is no actions the user can take in those case. Just ignore the errors and return should be sufficient.

kong/constants.lua
```diff
@@ -140,7 +140,7 @@ local constants = {
   },
   REPORTS = {
     ADDRESS = "kong-hf.konghq.com",
-    STATS_PORT = 61830
+    STATS_PORT = 61833,
```
dndx, line 143:
```suggestion
    STATS_PORT_TLS = 61833,
```

So it is more obvious this is a TLS port.

kong/reports.lua
```diff
@@ -108,7 +108,6 @@ local function send_report(signal_type, t, host, port)
   elseif type(signal_type) ~= "string" then
     return error("signal_type (arg #1) must be a string", 2)
   end
-
   t = t or {}
   host = host or constants.REPORTS.ADDRESS
   port = port or constants.REPORTS.STATS_PORT
```
dndx, line 113:
```suggestion
  port = port or constants.REPORTS.STATS_PORT_TLS
```

dndx (approved):
Awaiting tests but should be good to merge for now :) Thanks for the work @jschmid1 !

PR #8163, opened 2021-12-08
Rename config "anonymous_reports" to "reports"

hhoover:
### Summary

This change changes the configuration value in Kong Gateway from
`anonymous_reports` to `reports` as the usage data sent to Kong, Inc.
is not anonymous. This setting remains enabled by default.

See also: https://github.com/Kong/docs.konghq.com/pull/3463


jschmid1:
additional suggestions for the naming:

* telemetry
* phone-home

hhoover:
> additional suggestions for the naming:

The name can be anything, just needs to be honest. The main issue here is over the word "anonymous".

gszr:
This should likely be done in a backwards-compatible manner -- i.e., still accepting the old property name as a fallback. Otherwise, places that have disabled it will revert to enabled, which is an undesirable result for users and for ourselves (there will be a lot of pings coming, e.g., from development environments and CI).

PR #8168, opened 2021-12-09
optimize(pdk) use string.byte to avoid lj_str_new overhead

chronolaw:
### Summary

Optimize pdk request.lua with string.byte,
It will avoid  lj_str_new temp string overhead

### Full changelog

* add local string_byte = string.byte
* change request.set_path, replace string.sub to string.byte

### DIFF
--- a/kong/pdk/service/request.lua
+++ b/kong/pdk/service/request.lua
@@ -15,6 +15,7 @@ local table_concat = table.concat
 local type = type
 local string_find = string.find
 local string_sub = string.sub
+local string_byte = string.byte
 local string_lower = string.lower
 local normalize_header = checks.normalize_header
 local normalize_multi_header = checks.normalize_multi_header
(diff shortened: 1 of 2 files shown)


kong/pdk/service/request.lua
```diff
@@ -141,7 +142,8 @@ local function new(self)
       error("path must be a string", 2)
     end
 
-    if string_sub(path, 1, 1) ~= "/" then
+    -- 47 = 0x2f = '/'
+    if string_byte(path) ~= 47 then
```
dndx, line 146: We can do something similar to https://github.com/Kong/kong/blob/master/kong/tools/uri.lua#L42 to avoid magic numbers in the code.

chronolaw: add a local SLASH in function new()

bungle (approved):

PR #8173, opened 2021-12-09
Revert "feat(rate-limiting) add redis username support (#8032)"

locao:
Reverts #8032

This feature needs support in different parts of Kong before being released. We are postponing its release.


aboudreault (approved):
Approving. //cc @lena-larionova (what we discussed on Slack)

Issue #8160, opened 2021-12-08
Tests inadvertently send metrics to our metrics instance 


jschmid1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

2.7.0

### Current Behavior

In https://github.com/Kong/kong/pull/8159 I disabled two tests as they were inadvertently sending metrics to our metrics instance (kong-hf.kong.com). This behavior existed before this patch.

"db config db_import does not require Kong to be running"
Reason:
They do not start kong and therefore do not take a patched dnshosts file that would redirect kong-hf.kong.com to localhost.

I also disabled this test:

stream reports tcp streams
stream reports tls streams

I'm not seeing how they every should've passed in the first place. 

### Expected Behavior

Tests should not send metrics to the actual endpoint but should be rerouted to a local/mock instance.

### Steps To Reproduce

```markdown
Run tests
```

kikito:
It seems this issue can be closed, please reopen if that is not the case

PR #8175, opened 2021-12-10
Fix tls report (2.7)

aboudreault:
### Summary

Use & prioritize TLS port for phone-home capability

### Full changelog

* Use & prioritize TLS port for phone-home capability

Notes:

I disabled two tests as they were inadvertently sending metrics to our metrics instance (kong-hf.kong.com). This behavior existed _before_ this patch.

- "db config db_import does not require Kong to be running"

Reason: 
They do not start kong and therefore do not take a patched dnshosts file that would redirect kong-hf.kong.com to localhost.

I also disabled this test:

* stream reports tcp streams
* stream reports tls streams

I'm not seeing how they every should've passed in the first place. 

Creating a ticket to track this. https://github.com/Kong/kong/issues/8160


aboudreault:
//cc @randomcamel

aboudreault:
//cc @RobSerafini

mikefero (approved):

gszr (approved):

PR #8179, opened 2021-12-12
fix(dao) apply cache_key for target uniqueness detection

yankun-li-kong:
### Summary

Apply cache_key for target uniqueness detection
Improvement for https://github.com/Kong/kong/pull/6483

### Full changelog

* Add new cache_key(upstream, target) in targets table for uniqueness detection.
* Delete useless targets uniqueness detection functions.
* Targets API returns 409 when creating/updating delicate targets.
* Add migration functions to add cache_key column, delete duplicate targets, add cache_key for existing targets.

### Issues resolved

[CT-97](https://konghq.atlassian.net/browse/CT-97)


yankun-li-kong:
Hi @locao would like to discuss below points

I have deleted the functions you wrote in kong/api/routes/upstreams.lua.
It will changed the behavior of  `POST targets` API.

Before:
`POST targets` API is able to create new targets or update existing targets

After:
`POST targets` API is only able to create new targets
You have to use `PATCH target` API to update existing targets

Please let me know your opinion.
If we want to keep previous behavior(see Before), then I should not modify kong/api/routes/upstreams.lua

dndx:
Related to https://github.com/Kong/kong/pull/6483. Please see comments there for context.

kong/db/migrations/core/015_270_to_280.lua
```diff
@@ -0,0 +1,122 @@
+-- remove repeated targets, the older ones are not useful anymore. targets with
+-- weight 0 will be kept, as we cannot tell which were deleted and which were
+-- explicitly set as 0.
```
locao, line 3: Since the target history removal in 2.2 we don't keep deleted targets.

spec/02-integration/04-admin_api/08-targets_routes_spec.lua
```diff
@@ -142,7 +142,7 @@ describe("Admin API #" .. strategy, function()
         end
       end)
 
-      it_content_types("updates and does not create duplicated targets (#deprecated)", function(content_type)
+      it_content_types("updates and does not create duplicated targets", function(content_type)
```
locao, line 145:
```suggestion
      it_content_types("refuses to create duplicated targets", function(content_type)
```

locao (commented):
Thanks for bringing this up @yankun-li-kong! That's something we've been waiting for 3.0 to do.

The code seems OK to me, just a few suggestions on the Cassandra migration. My main concern is that we should not merge this change to `master` at this moment. We are still having another release before 3.0.

kong/db/migrations/core/015_270_to_280.lua
```diff
@@ -0,0 +1,122 @@
...
+  end
+
+  for _, targets in pairs(upstream_targets) do
+    if targets.n > 1 then
+      table.sort(targets, sort)
+
+      for i = 2, targets.n do
+        local _, err = coordinator:execute("DELETE FROM targets WHERE id = ?", {
+          cassandra.uuid(targets[i][1])
+        })
```
locao, line 35: (somehow this comment was not added to my review)

I haven't tested this suggestion, but it seems to me that it's possible to delete the duplicated targets in a single loop.

```suggestion
upstream_targets[key] = {
          id = row.id, 
          created_at = row.created_at,
        }
      else
        local to_remove
        if row.created_at > upstream_targets[key].created_at then
          to_remove = upstream_targets[key].id
          upstream_targets[key] = {
            id = row.id, 
            created_at = row.created_at,
          }
        else
          to_remove = row.id
        end
        local _, err = coordinator:execute("DELETE FROM targets WHERE id = ?", {
          cassandra.uuid(to_remove)
        })
```

kong/db/migrations/core/015_270_to_280.lua
```diff
@@ -0,0 +1,122 @@
...
+
+-- update cache_key for targets
+local function c_update_target_cache_key(coordinator)
+  local cassandra = require "cassandra"
+  for rows, err in coordinator:iterate("SELECT id, upstream_id, target, ws_id FROM targets") do
+    if err then
+      return nil, err
+    end
+
+    for _, row in ipairs(rows) do
```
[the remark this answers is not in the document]
yankun-li-kong: Thanks for your suggestion!
I think it should not make much difference as the flow is still the same as below

`SELECT from targets` and pass targets list to `rows`
then execute `UPDATE targets` in a loop with `rows`

kong/db/migrations/core/015_270_to_280.lua
yankun-li-kong: Thanks! I saw kong has run `c_remove_unused_targets` in `012_213_to_220.lua` to delete duplicate targets in 2.2.

But duplicate targets may still exist after that.
Regarding https://github.com/Kong/kong/blob/master/kong/db/dao/targets.lua#L50-L59
kong will do uniqueness detection and then insert targets. 
It is not an atomic operation and a duplicate target may be added during the above process.

spec/02-integration/04-admin_api/08-targets_routes_spec.lua
yankun-li-kong: Good suggestion, fixed!

kong/db/migrations/core/015_270_to_280.lua
yankun-li-kong: Fixed and tested in my local.
It looks much simple than previous one!

kong/api/routes/upstreams.lua
```diff
@@ -181,16 +166,6 @@ return {
...
                                             "page_for_upstream"),
     POST = function(self, db)
-      -- updating a target using POST is a compatibility with existent API and
-      -- should be deprecated in next major version
-      local entity, _, err_t = update_existent_target(self, db)
-      if err_t then
-        return endpoints.handle_error(err_t)
-      end
-      if entity then
-        return kong.response.exit(200, entity, { ["Deprecation"] = "true" })
```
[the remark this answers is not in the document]
yankun-li-kong: Got it. Please let me know when we should merge it so I could rename `012_213_to_220.lua` file

dndx:
Hello @yankun-li-kong ,


Could you rebase this onto the latest `master` so it can be ready for merge? Since this will be release in 3.0, a new migration file will be needed. We also need some entry inside the `CHANGELOG.md` file describing this change.

locao (approved):

dndx (approved):

PR #8183, opened 2021-12-13
fix(perf) search nginx_bin in local driver

chronolaw:
Summary
fix search nginx_bin in local driver's setup

Full changelog
fix for loop of nginx bin


kikito (approved):

kikito:
Thanks for this! Note that all the prs that modify the `spec/` folder exclusively should be scoped under `tests`. I will change the scope while I squash it into the repo.

PR #8193, opened 2021-12-15
fix: It supports TLS SNI custom configuration and needs to be used in…

zhangshuaiNB:
### Summary

In version 2.6 0, unable to transparently transmit SNI when proxy TLS service

In this modification, you can customize the transparent dynamic TLS SNI name. If you need to customize it, you can configure it in the plug-in. The default name is the same as the original one,using kong_ upstream

### DIFF
--- a/kong/templates/nginx_kong_stream.lua
+++ b/kong/templates/nginx_kong_stream.lua
@@ -122,9 +122,11 @@ server {
     }
 > end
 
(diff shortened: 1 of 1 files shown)


zhangshuaiNB:
image

I have signed the CLA, but still prompt "cla not signed yet"

PR #8196, opened 2021-12-15
Support TLS SNI custom dynamic configuration

zhangshuaiNB:
### Summary

In version 2.6 0, unable to transparently transmit SNI when proxy TLS service

In this modification, you can customize the transparent dynamic TLS SNI name. If you need to customize it, you can configure it in the plug-in. The default name is the same as the original one,using kong_ upstream

### DIFF
--- a/kong/templates/nginx_kong_stream.lua
+++ b/kong/templates/nginx_kong_stream.lua
@@ -122,9 +122,11 @@ server {
     }
 > end
 
+    set $tls_sni_name 'kong_upstream';
     preread_by_lua_block {
         Kong.preread()
     }
+    proxy_ssl_name $tls_sni_name;
 
     proxy_ssl on;
     proxy_ssl_server_name on;


fffonion:
Could you elaborate a bit about your use case?

> In version 2.6 0, unable to transparently transmit SNI when proxy TLS service

Do you mean this is a change of behaviour in Kong 2.6, potentially a bug, or are you trying to add new feature?

zhangshuaiNB:
image

- Due to our business needs, we have designed an architecture in which two layers of Kong are used to realize our TLS traffic agent
- In the traffic agent, it is found that kong-2 cannot match the route
- The reason for finding is that the SNI name cannot be transparently transmitted during the process from Kong-1 to kong-2, so kong-2 cannot match the route
- View Kong's nginx-kong-stream.conf, found that "proxy_ssl_name" is not configured, with official introduction of nginx here, [https://nginx.org/en/docs/stream/ngx_stream_proxy_module.html#proxy_ssl_name](url)
- So we think of a way to dynamically set the value of "proxy_ssl_name", which can be used with or without the plugin. The default is "kong_ upstream"

I think this may be a defect. You can enrich the functions, or if you have better solutions to solve our problems, thank you very much

zhangshuaiNB:
> image
> 
> * Due to our business needs, we have designed an architecture in which two layers of Kong are used to realize our TLS traffic agent
> * In the traffic agent, it is found that kong-2 cannot match the route
> * The reason for finding is that the SNI name cannot be transparently transmitted during the process from Kong-1 to kong-2, so kong-2 cannot match the route
> * View Kong's nginx-kong-stream.conf, found that "proxy_ssl_name" is not configured, with official introduction of nginx here, [https://nginx.org/en/docs/stream/ngx_stream_proxy_module.html#proxy_ssl_name](url)
> * So we think of a way to dynamically set the value of "proxy_ssl_name", which can be used with or without the plugin. The default is "kong_ upstream"
> 
> I think this may be a defect. You can enrich the functions, or if you have better solutions to solve our problems, thank you very much

@fffonion

fffonion:
@zhangshuaiNB I see, that make sense to me. At least when adding the new directive, Kong doesn't change its behaviour so this change should be at low risk.

fffonion (approved):

mayocream:
Hi @zhangshuaiNB, Thanks for your first PR to Kong! 🎉 Nice work!

mayocream:
@zhangshuaiNB Hi, Don't forget to claim your [contributor t-shirt](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt) :)

PR #8197, opened 2021-12-15
perf(clustering) optimize config hash calculation

bungle:
### Summary

Previously the code generated a log of strings and garbage when calculating the hash for the configuration. It happened to create one big string and then calculated hash over it.

This commit changes it so that it now builds hash while iterating over the structures (with `hash:update`) and thus should not create strings or tables (and garbage).


bungle:
This only makes things worse. Back to the drawing board.

PR #8203, opened 2021-12-16
Fix reports tests issues

aboudreault:
Minor fixes in reports tests.


jschmid1:
I can still see a 3m~ increase for the "integration - (> = 05)" tests due to.

```
Error from thread: ./spec/helpers.lua:1069: timeout
stack traceback:
	[C]: in function 'assert'
	./spec/helpers.lua:1069: in function <./spec/helpers.lua:1045>
Error from thread: ./spec/helpers.lua:1069: timeout
stack traceback:
	[C]: in function 'assert'
	./spec/helpers.lua:1069: in function <./spec/helpers.lua:1045>
```

While I think your patch makes sense, I'll investigate a bit more

aboudreault:
@jschmid1 if you haven't checked this yet, I will check this more this PM.

flrgh (approved):

PR #8204, opened 2021-12-16
perf(clustering) improve hash calculation performance

bungle:
### Summary

With large configurations the hash calculation can take a lot of time and memory. This PR optimizes the performance of configuration hash calculation. The time is cut to about half of what it used to be and the peak memory usage has also gone down.

This does not produce exactly same hash on certain input as it did before (as I optimized array case and did other change).

### Results

Before this PR:

Calculating hash done in 13.43399977684 secs
Maximum resident set size (kbytes): 6081520

After this PR:

Calculating hash done in 4.1620001792908 secs
Maximum resident set size (kbytes): 2720864

Tests were done with 395M kong.json file that was decoded to lua table. The memory usage without hash calculation is:

Maximum resident set size (kbytes): 2163472

Performance boost:
13,43399977684 - 4,1620001792908 = 9,2719995975492 (secs) (3 times as fast)

Memory reduction:
6081520 - 2720864 = 3360656 (kbytes) (3 gigs less memory used)

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -19,7 +19,6 @@ local concat = table.concat
 local pairs = pairs
 local sort = table.sort
 local type = type
-local min = math.min
 
 
 local MT = { __index = _M, }
@@ -31,14 +30,11 @@ local function to_sorted_string(value)
   end
 
   local t = type(value)
-  if t == "string" then
-    return "$" .. value .. "$"
-
-  elseif t == "number" then
-    return "#" .. value .. "#"
+  if t == "string" or t == "number" then
+    return value
 
   elseif t == "boolean" then
-    return "?" .. tostring(value) .. "?"
+    return tostring(value)
 
   elseif t == "table" then
     if isempty(value) then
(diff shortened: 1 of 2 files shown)


kong/clustering/init.lua
```diff
@@ -5,67 +5,116 @@ local pl_file = require("pl.file")
...
+                   .. to_sorted_string(value[2]) .. ";"
+                   .. to_sorted_string(value[3]) .. ";"
+                   .. to_sorted_string(value[4]) .. "}"
+
+      elseif count == 5 then
+        return "{" .. to_sorted_string(value[1]) .. ";"
+                   .. to_sorted_string(value[2]) .. ";"
+                   .. to_sorted_string(value[3]) .. ";"
+                   .. to_sorted_string(value[4]) .. ";"
+                   .. to_sorted_string(value[5]) .. "}"
```
jschmid1, line 72: Could you elaborate on this block? This looks arbitrarily restrictive (and redundant) to me.

Also, the unittests do not include a case for this whole section.

jschmid1 (commented):
while the PR seems to achieve what it's goal is, I guess we should think about a more generic table serialization approach. 

Also, since this isn't the last time we touched this rather crucial piece of code we should think about adding descriptive comments.

bungle:
> while the PR seems to achieve what it's goal is, I guess we should think about a more generic table serialization approach.

What do you have in mind?

kong/clustering/init.lua
bungle: The whole idea of this function is just to get MD5 hash out of the (arbitrary) lua table. The function is rather generic currently and further optimizations can be done with being opinionated, e.g. being opinionated that all table (not array) keys are strings (like they are in db export generated tables currently). The problem of using something like cjson.encode or the upcoming: https://github.com/LuaJIT/LuaJIT/blob/v2.1/doc/ext_buffer.html#L89-L90 is that they can be undeterministic. At least the cjson.encode is. Because `pairs` over Lua table is undeterministic too.

This code here is just to skip the table creation and md5 on small arrays. We have a lot of small arrays in our db exports. It is a hand optimization for this particular case.

jschmid1: ok, got it. thanks for clarifying.

jschmid1:
> What do you have in mind?

Nothing in particular. I just noticed that we have plenty of cases where we do some form of marshalling/serialization in kong. Having a well tested and optimized approach would be neat I guess.

kong/clustering/init.lua
bungle: There should be now unit test that cover this section.

kikito (approved):
I have reviewed this internally with aapo, even sending algorithms back and forth. We have tested on my hardware (intel) and his (m1) and this particular implementation is the one that came up on top, after many tries. It improves the previous attempt significantly. Collisions are theoretically possible, but in practice they shouldn't happen enough to be a concern (this is not a "generic" algorithm, it only needs to calculate the hash for tables with very specific structure - for example the hash-like table keys are always strings)

I think this is good enough. Approved.

PR #8205, opened 2021-12-16
docs(pdk) hostnames are allowed in kong.service.*

Tieske:
The text was limiting to IP addresses, but hostnames can also be set (as well as Upstream names).

### DIFF
--- a/kong/pdk/service.lua
+++ b/kong/pdk/service.lua
@@ -31,7 +31,7 @@ local function new()
   -- request would be proxied to one of the Targets associated with that
   -- Upstream).
   --
-  -- The `host` argument should receive a string equal to that of one of the
+  -- The `host` argument should receive a string equal to the name of one of the
   -- Upstream entities currently configured.
   --
   -- @function kong.service.set_upstream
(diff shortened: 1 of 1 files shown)


kong/pdk/service.lua
```diff
@@ -72,9 +72,8 @@ local function new()
   -- Load-balancing components such as retries and health-checks will also be
   -- ignored for this request.
   --
-  -- The `host` argument expects a string containing the IP address of the
-  -- upstream server (IPv4/IPv6), and the `port` argument must contain a number
-  -- representing the port on which to connect to.
+  -- The `host` and `port` arguments expect the values as set on a Service
+  -- entity.
```
lena-larionova, line 76:
```suggestion
  -- The `host` argument expects the hostname or IP address of the upstream 
  -- server, and the `port` expects a port number. The values provided must 
  -- match the values set for the related Service entity.
```
How about something like this? It might seem obvious, but it's good to call out both the exact expected value formats and where to look for them.

Tieske: Thx @lena-larionova . Not a native speaker, but; "_The values provided must match the values set for the related Service entity._" tells me that the values must be equal to the values in the Service. Yet, that is not the case; they only follow the same constraints.

ShubhamPalriwala: I agree with what @lena-larionova has suggested as it is pretty clear and aligns well with this PR too! However the second part of the statement is confusing. Hence lets have
```lua
-- The `host` argument expects the hostname or IP address of the upstream 
-- server, and the `port` expects a port number
```

ShubhamPalriwala (commented):
The changes lgtm along with @Tieske's and @lena-larionova suggestions and I have added my comments for the change that I feel would make more sense.

Tieske:
please have another look

PR #8208, opened 2021-12-17
fix(pdk) get_phases_names wrong lshift

chronolaw:
### Summary

PHASES.n means lshift times, but now it equals to 13, it is 0x2000，
but the max phases admin_api is 0x10000000,
so the function get_phases_names can not get the right name.

I change the PHASES.n to 30, then we can support the all phases bits.

### Full changelog

* change PHASES.n to 30, means 2^30 = 0x40000000

### DIFF
--- a/kong/pdk/private/phases.lua
+++ b/kong/pdk/private/phases.lua
@@ -33,13 +33,12 @@ do
     t[k] = v
   end
 
-  local n = 0
   for k, v in pairs(t) do
-    n = n + 1
     PHASES[v] = k
   end
 
-  PHASES.n = n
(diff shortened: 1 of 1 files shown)


kikito (approved):
You fix an existing problem (that we sometimes don't get the appropriate phase name for some phases) and I think that the change made here was as minimal as it could, which is usually good. So I am inclined to approving this.

kikito:
(I also happen to think that the current implementation of `kong/pdk/private/phases.lua` is too clever and confusing - mostly because of two reasons: using a single PHASES table instead of two (one for number => name and a different one for name => number) and naming several things `n` (number of bits and bit being iterated over).

PR #8210, opened 2021-12-17
chore(entities) adjust tags schema

bungle:
### Summary

Yes, `tags` schema is kinda abstract, but it was a bit confusing for me to see a field with `type=string` and then having `elements` too. This commit just changes the schema to a bit less confusing.

### DIFF
--- a/kong/db/schema/entities/tags.lua
+++ b/kong/db/schema/entities/tags.lua
@@ -9,11 +9,7 @@ return {
 
   fields = {
     { tag          = typedefs.tag, },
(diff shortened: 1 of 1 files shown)


fffonion (approved):

guanlan:
will it be a breaking change?

fffonion:
@guanlan tags entity is readonly, so having a different a typedef for it won't affect current behaviour. Actually, a string
with element of UUID doesn't make sense after all (and will break somewhere if it's actually being type-checked).

PR #8213, opened 2021-12-17
chore(revert) re-add Redis username support

mikefero:
### Summary

Re-adding `username` support for Redis connections; @27ascii contribution.

On top of reverting the removal of the `username` support, this PR fixes an issue with Redis increment eval query, adds Redis server version checking for the ACL tests added and adds a minimum supported Redis version; logs error if below minimum.

see https://github.com/Kong/kong/pull/8032 for original PR by @27ascii. 

### Full changelog

* Adds the optional property redis_username used for connecting to redis. This allows to make use of redis ACL features supported with Redis 6.0
* fix(plugin) correct Redis increment eval for ACL queries

### DIFF
--- a/kong/plugins/rate-limiting/policies/init.lua
+++ b/kong/plugins/rate-limiting/policies/init.lua
@@ -79,7 +79,13 @@ local function get_redis_connection(conf)
 
   if times == 0 then
     if is_present(conf.redis_password) then
-      local ok, err = red:auth(conf.redis_password)
+      local ok, err
+      if is_present(conf.redis_username) then
+        ok, err = red:auth(conf.redis_username, conf.redis_password)
+      else
+        ok, err = red:auth(conf.redis_password)
+      end
+
       if not ok then
         kong.log.err("failed to auth Redis: ", err)
         return nil, err
--- a/kong/plugins/rate-limiting/policies/init.lua
+++ b/kong/plugins/rate-limiting/policies/init.lua
@@ -180,21 +180,35 @@ return {
         return nil, err
       end
 
+      local keys = {}
+      local expiration = {}
+      local idx = 0
       local periods = timestamp.get_timestamps(current_timestamp)
-      red:init_pipeline()
       for period, period_date in pairs(periods) do
         if limits[period] then
           local cache_key = get_local_key(conf, identifier, period, period_date)
+          local exists, err = red:exists(cache_key)
+          if err then
+            kong.log.err("failed to query Redis: ", err)
+            return nil, err
+          end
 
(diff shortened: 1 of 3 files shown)


kong/constants.lua
```diff
@@ -159,6 +159,9 @@ local constants = {
     CASSANDRA = {
       MIN = "3.0",
       DEPRECATED = "2.2",
+    },
+    REDIS = {
+      MIN = "5.0"
```
bungle, line 164: Is this only true when ´username` is used? Otherwise we do support less than that?

kong/tools/utils.lua
```diff
@@ -1445,5 +1447,25 @@ do
...
+    local message = "Redis v" .. tostring(red_version) ..
+      " is end of life and will not be supported; upgrade to v" ..
+      tostring(min_red_version) .. "+ is recommended"
+    ngx.log(ngx.WARN, message)
+    return red_version, message
+  end
+
+  return red_version, nil
+end
+_M.check_redis_support = check_redis_support
```
aboudreault, line 1469: :+1:

aboudreault (approved):

kong/constants.lua
aboudreault: Fero and I discussed on slack and we agreed to not include the redis support minimum version check. That can be added back at a later time (3.0).

mikefero: Thanks for the discussion Alan; consider this reverted!

mikefero:
Thanks for you patience @27ascii on getting this feature back into the code base; it will make it into the next minor release.

27ascii:
@mikefero 
Thanks for moving the username support into this release. 

I just reviewed the change in the response rate limiting. I saw that redis ssl options are only available in the rate-limiting plugin:

https://github.com/Kong/kong/blob/443e30f7a018bcd690f3ecd9701cd1028bd668b0/kong/plugins/rate-limiting/policies/init.lua#L56-L64

https://github.com/Kong/kong/blob/443e30f7a018bcd690f3ecd9701cd1028bd668b0/kong/plugins/response-ratelimiting/policies/init.lua#L56-L60

I wonder if there are any plans to move the redis get_redis_connection into the kong util in order to provide a common way to handle redis connections. This might as well be extended to other redis connect options in future (e.g. redis sentinel).

mikefero:
@27ascii 

This is something that is on the radar (well backlog) to have all plugins that require Redis use a shared connector, therefore getting all the same Redis features (if applicable). This would also be beneficial for custom plugin developers to be able to quickly add Redis support for their needs as well.

PR #8224, opened 2021-12-22
feat(dbless) use LMDB as DB-less backend

dndx:
This PR adds LMDB (Lightning Memory-Mapped Database) support for
DB-less. At the same time, the shdict based DB-less storage backend has
been retired and removed from the codebase.

LMDB has better concurrency characters and is generally much more stable
than shdict for storing config data. Because it can be natively accessed
from different processes, we also removed the hack that sends the full
DB-less config to the stream subsystem, which should improve the
stability of DB-less reload at runtime as well.

New config options `lmdb_environment_path` and `lmdb_map_size` has been
added.

TODO: LMDB is fully capable of persisting the config between restarts, currently this is not enabled so it behaves like a in-memory storage (content is always wiped at startup). This will be enabled once we have more understanding of the impact.


kong/db/declarative/init.lua
```diff
@@ -798,174 +796,91 @@ function declarative.load_into_cache(entities, meta, hash, shadow)
...
+  kong.default_workspace = default_workspace
+
+  local ok, err = t:commit()
   if not ok then
-    return nil, "failed to set " .. DECLARATIVE_HASH_KEY .. " in shm: " .. err
+    return nil, err
   end
 
+  kong.core_cache:purge()
+  kong.cache:purge()
```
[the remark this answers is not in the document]
dndx: Maybe purge after the router rebuild.

kong/db/declarative/marshaller.lua
```diff
@@ -0,0 +1,127 @@
+local _M = {}
```
dndx, line 1: Get rid of `cache:safe_set` and `cache.marshaller`

kong/global.lua
```diff
@@ -205,13 +204,6 @@ function _GLOBAL.init_cache(kong_config, cluster_events, worker_events)
   local page = 1
   local cache_pages = 1
 
-  if kong_config.database == "off" then
-    db_cache_ttl = 0
-    db_cache_neg_ttl = 0
```
dndx, line 210: We should keep this.

kong/init.lua
```diff
@@ -401,11 +375,26 @@ local function parse_declarative_config(kong_config)
...
+  if not ok then
+    return nil, "error building initial router: " .. err
   end
 
+  return true
+end
+
+
+local function load_declarative_config(kong_config, entities, meta)
+  print("load_declarative_config!!!")
```
dndx, line 397: Remove this.

kong/init.lua
```diff
@@ -545,10 +521,12 @@ function Kong.init()
   end
 
   if config.database == "off" then
-    local err
-    declarative_entities, err, declarative_meta = parse_declarative_config(kong.configuration)
-    if not declarative_entities then
-      error(err)
+    if is_http_module or #config.proxy_listeners == 0 then
```
dndx, line 524: Need to check `admin_listeners` or `status_listeners` as well.

kong/init.lua
```diff
@@ -651,12 +629,25 @@ function Kong.init_worker()
...
-                                    declarative_meta)
-  if not ok then
-    stash_init_worker_error("failed to load declarative config file: " .. err)
-    return
+  if kong.configuration.database == "off" then
+    local t = lmdb_txn.begin(1)
+    t:db_open(true)
+    assert(t:commit())
+
+    if is_http_module or #kong.configuration.proxy_listeners == 0 then
```
dndx, line 637: Ditto.

spec/02-integration/04-admin_api/15-off_spec.lua
```diff
@@ -8,7 +8,8 @@ local mocker   = require("spec.fixtures.mocker")
 
 
 local WORKER_SYNC_TIMEOUT = 10
-local MEM_CACHE_SIZE = "15m"
+local MEM_CACHE_SIZE = "10m"
```
dndx, line 11: Rename this.

kong/init.lua
fffonion: not sure for status_listners as it's readonly

kong/init.lua
```diff
@@ -651,12 +629,25 @@ function Kong.init_worker()
...
   kong.db:set_events_handler(worker_events)
 
-  ok, err = load_declarative_config(kong.configuration,
-                                    declarative_entities,
-                                    declarative_meta)
-  if not ok then
-    stash_init_worker_error("failed to load declarative config file: " .. err)
-    return
+  if kong.configuration.database == "off" then
+    local t = lmdb_txn.begin(1)
```
fffonion, line 633: let's have a comment for this (or even better make this a function like sanity_test())

kong/init.lua
dndx: We will follow what the `kong/templates/nginx.lua` does as that ultimately determines whether `http`'s `init` phase is rendered or not.

kong/db/declarative/marshaller.lua
```diff
@@ -0,0 +1,127 @@
...
+    end,
+}
+
+
+function _M.marshall(value)
+    if value == nil then
+        return marshallers.shm_nil()
+    end
+
+    local value_type = TYPES_LOOKUP[type(value)]
```
javierguerragiraldez, line 59: why is **nil** handled separately?

kong/db/strategies/off/init.lua
```diff
@@ -173,7 +168,10 @@ local function page_for_key(self, key, size, offset, options)
     -- The rest of entities' lists (i.e. "services|<ws_id>|@list") only contain ids, so in order to
     -- get the entities we must do an additional cache access per entry
     else
-      item = cache:get(item, nil, nil_cb)
+      item, err = unmarshall(lmdb_get(item))
+      if err then
```
javierguerragiraldez, line 172: i've noticed some are of the form
```lua
item, err = unmarshall(lmdb_get(item))
if err then ...
```
while others (in this same file) are
```lua
item, err = unmarshall(lmdb_get(item))
if not item then ....
```
any reason for the difference?  (i usually favor the `if not item then...` style, but sometimes it can't be done)

kong/init.lua
```diff
@@ -545,10 +520,16 @@ function Kong.init()
...
 
   if config.database == "off" then
-    local err
-    declarative_entities, err, declarative_meta = parse_declarative_config(kong.configuration)
-    if not declarative_entities then
-      error(err)
+    if is_http_module or
+       (#config.proxy_listeners == 0 and
+        #config.admin_listeners == 0 and
+        #config.status_listeners == 0)
```
kikito, line 526: I don't understand the logic of this `if`. If "is http" or ..."is stream but listening to nothing"? What is this testing, really?

kikito (changes requested):

kong/templates/kong_defaults.lua
```diff
@@ -30,6 +30,8 @@ cluster_server_name = NONE
 cluster_data_plane_purge_delay = 1209600
 cluster_ocsp = off
 
+lmdb_environment_path = dbless.lmdb
```
bungle, line 33: should we just call it `lmdb` instead?

kong/db/declarative/marshaller.lua
dndx: I am not a 100% sure, but that is what's from mlcache so I just copied..

Maybe it is trying to distinguish between `nil` and "this value doesn't exist", it may not be necessary for LMDB use case. Although I don't think we intentionally store nil inside LMDB either as setting a value to `nil` in LMDB is a delete operation. I will remove the logic and see if it still works.

dndx: Removed and seems to be fine :)

.requirements
```diff
@@ -8,4 +8,4 @@ RESTY_OPENSSL_VERSION=1.1.1m
 RESTY_PCRE_VERSION=8.45
 LIBYAML_VERSION=0.2.5
 KONG_GO_PLUGINSERVER_VERSION=v0.6.1
-KONG_BUILD_TOOLS_VERSION=4.22.0
+KONG_BUILD_TOOLS_VERSION=master
```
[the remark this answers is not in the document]
dndx: Yes, this will be changed to a new version before the release.

kong/clustering/data_plane.lua
```diff
@@ -186,10 +186,6 @@ local function send_ping(c, log_suffix)
 
   local hash = declarative.get_current_hash()
 
-  if hash == true then
-    hash = string.rep("0", 32)
-  end
```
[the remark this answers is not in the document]
dndx: This was discussed with @bungle and we intentionally moved the generation of `string.rep("0", 32)` into the lower layer for less abstraction leaks.

kong/db/declarative/init.lua
```diff
@@ -663,16 +660,15 @@ function declarative.load_into_cache(entities, meta, hash, shadow)
...
       end
 
-      local ok, err = kong.core_cache:safe_set(cache_key, item, shadow)
-      if not ok then
+      local item_marshalled, err = marshall(item)
+      if not item_marshalled then
         return nil, err
       end
 
+      t:set(cache_key, item_marshalled)
```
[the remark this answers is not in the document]
dndx: The `:set` is a low level LMDB interface, and it will only accept string input for performance reasons (not all use cases needs marshaling, after all). So in this case Kong will perform the marshaling itself before handing over to LMDB (with a small perf benefit). But perf is not the reason for doing the marshaling separately.

kong/db/strategies/off/init.lua
```diff
@@ -197,11 +195,7 @@ end
 
 
 local function select_by_key(self, key)
-  if not kong.core_cache then
-    return nil
-  end
-
-  local entity, err = kong.core_cache:get(key, nil, nil_cb)
+  local entity, err = unmarshall(lmdb_get(key))
   if not entity then
```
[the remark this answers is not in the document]
dndx: Good catch, fixed.

kong/init.lua
dndx: This achieves the following:

1. If HTTP is enabled, then config is loaded by the HTTP submodule and Stream submodule will not load any config
2. If HTTP is not enabled, then Stream subsystem will have to do the config loading itself

This is different from shdict based implementation as LMDB is shared between HTTP and Stream, so loading only need to happen once.

kong/templates/kong_defaults.lua
dndx: I am considering we may introduce other LMDB databases later, calling it `lmdb` only seems to not tell clearly what the directory is for.

In fact, I even considered remove the "lmdb" part from the `lmdb_environment_path` config and call it `dbless_database_path` instead. Does it seems to be a good idea?

bungle: The `dbless` is little bit strange to me, as it is used on hybrid data planes too, and not only on dbless node. Yes, currently hybrid data planes are kinda dbless nodes, but lmdb really also feels like a db, just that it is not an external dependency. Also I don't like dots in directory names, as they look more like files. But well.. I have no strong feelings, it just looked a bit off to me.

kong/init.lua
dndx: Resolved.

kong/db/declarative/init.lua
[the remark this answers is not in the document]
kikito: I have reviewed my concerns about updates with Datong.

The TLDR is: we *might* add a block *at the beginning of the procedure* (instead of just blocking the page access, like we did before, we might block the whole thing, including JSON parsing) on the final implementation. LMDB updates are faster than our previous system and the transactions take care of the "multiple page" problem so it might not be necessary, it might require some tests.

kong/db/declarative/init.lua
```diff
@@ -584,8 +581,8 @@ function declarative.load_into_cache(entities, meta, hash, shadow)
   -- but filtered for a given tag
   local tags_by_name = {}
 
-  kong.core_cache:purge(shadow)
-  kong.cache:purge(shadow)
+  local t = txn.begin(128)
```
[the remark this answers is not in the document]
kikito: If we add the lock, It might make more sense to move it to the top of the load operation ("block the API, not the transaction mechanism"). This way we save memory not allocating and parsing massive JSON files just to block immediately afterwards.

kong/templates/kong_defaults.lua
kikito: Agreed that using something that has `DB` on its name to implement something called `db-less` feels ... a bit ridicule, and inexact.

I think what differentiates the strategy presented on this PR versus PostgreSQL is **individual updates**: you can't update a single Service with the Admin API, and you can't create an individual oauth2 token. You have to change the whole thing, or none at all. So for the strategy presented on this PR I would go with something along the lines of: `static`, `monolith`, `bulk` or `unit`.

I would reserve the `lmdb` name for a strategy that makes full use of lmdb (one that allows for individual Service updates), if we ever do that.

(I considered `memory`, but lmdb apparently can write to the hdd, so that one's discarded. `read-only` is also discarded, since you can make updates, just not individual ones).

I also want to mention that the *cost* of making this change:
* It's **very** backwards-breaking, so it blocks this into 3.0 territory. It is unlikely that we would be able to include it into 2.8, but if we later decide to release 2.9 or 2.10 (which seems unlikely, but things change fast in software) it would be blocked for those too.
* Changing this in the docs could be a hassle. Not only we mention db-vs-dbless in a lot of reference pages, the Admin API differentiates all methods into `dbless-compatible` and `db-only` (or something along those lines). Plugins are also marked as `db-less` compatible. We might need to release a new version of every plugin. Old blogposts will be out of date.

That cost ... already gives me pause. We could avoid all that by just "endure the feeling of ridiculousness and inexactitude".

kong/init.lua
kikito: Answered in private meeting. This just reflects the conditions present on the template. It might be better if they were abstracted on a helper function and used here and in the template.

.requirements
kikito, line 11: It seems that kong-build-tools-including-lmdb was already tagged, [many days ago](https://github.com/Kong/kong-build-tools/pull/426#issuecomment-1000779728)!
```suggestion
KONG_BUILD_TOOLS_VERSION=4.24.0
```

kong/templates/kong_defaults.lua
bungle: Perhaps then:
```
lmdb_environment_path = dbless
```

and call it done. remove the lmdb from it, if that is easier. We could call it `config` as well, as that is what we store in there. Or have a single `lmdb` directory and under that `dbless`. I don't know.

kikito:
I just added a temporary to the Jenkinsfile on this branch which should (theoretically) activate daily builds for this branch in Jenkins. Let's see tomorrow.

kong/db/declarative/init.lua
```diff
@@ -837,175 +834,117 @@ function declarative.load_into_cache(entities, meta, hash, shadow)
...
+  end
 
-      sleep_time = sleep_time * 2
-      if sleep_time > sleep_left then
-        sleep_time = sleep_left
-      end
+  -- If it takes more than 60s it is very likely to be an internal error.
+  -- However it will be reported as: "failed to broadcast reconfigure event: recursive".
+  -- Let's paste the error message here in case someday we try to search it.
+  -- Should we handle this case specially?
```
flrgh, line 923: As somebody who has seen this specific error message on a production system, big big +1 to anything that we can do to make the error more meaningful and explanatory.

kong/init.lua
```diff
@@ -653,12 +634,28 @@ function Kong.init_worker()
...
-  if not ok then
-    stash_init_worker_error("failed to load declarative config file: " .. err)
-    return
+  if kong.configuration.database == "off" then
+    -- databases in LMDB need to be explicitly created, otherwise `get`
+    -- operations will return error instead of `nil`. This ensures the default
+    -- namespace always exists in the
+    local t = lmdb_txn.begin(1)
+    t:db_open(true)
+    assert(t:commit())
```
flrgh, line 643: Should this follow the same error-handling pattern as the surrounding code (check return value and call `stash_init_worker_error(...)` if needed) instead of `assert()`?

spec/02-integration/04-admin_api/15-off_spec.lua
```diff
@@ -1040,11 +1041,14 @@ describe("Admin API #off with Unique Foreign #unique", function()
     assert.equal(references.data[1].note, "note")
     assert.equal(references.data[1].unique_foreign.id, foreigns.data[1].id)
 
-    local res = assert(client:get("/cache/unique_references||unique_foreign:" ..
-                                  foreigns.data[1].id))
-    local body = assert.res_status(200, res)
-    local cached_reference = cjson.decode(body)
+    local key = "unique_references\\|\\|unique_foreign:" .. foreigns.data[1].id
+    local handle = io.popen("resty --main-conf \"lmdb_environment_path " ..
```
flrgh, line 1045: This kinda thing really makes me wish lmdb could be initialized in lua land, but I'm assuming there are good reasons why that's infeasible right now.

kong/db/declarative/init.lua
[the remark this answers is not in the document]
aboudreault: From the discussion [here](https://github.com/Kong/kong/pull/8224/files#r784737751), this doesn't seem to be a blocker to merge the PR, but a potential improvement. That's correct @kikito ?

kong/db/declarative/init.lua
dndx: I am not a 100% sure about the triggering condition of this, and I hesitate to make a change without understanding why the "recursive" error occurs. Will try to syncup with @Suika-Kong more offline on this error handling after it is merged (the current PR is getting hard to manage).

kong/init.lua
dndx: This has been fixed already.

kong/db/declarative/init.lua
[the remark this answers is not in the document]
dndx: @Suika-Kong Already added some concurrency control for LMDB config loads, so I think this is less of a problem at this point.

PR #8226, opened 2021-12-22
Fix C* connector record migration consistency level

mpenick:
### Summary

The Cassandra connector should use the write consistency setting
(`cassandra_write_consistency`) from the Kong configuration file for
recording the migration status.

### Full changelog

* Fixed defect in Cassandra connector where it didn't use the configuration file consistency.

### DIFF
--- a/kong/db/strategies/cassandra/connector.lua
+++ b/kong/db/strategies/cassandra/connector.lua
@@ -1052,6 +1052,7 @@ do
 
     local cql
     local args
+    local opts = { consistency = self.opts.write_consistency }
 
     if state == "executed" then
       cql = [[UPDATE schema_meta
@@ -1088,7 +1089,7 @@ do
     table.insert(args, SCHEMA_META_KEY)
     table.insert(args, subsystem)
 
-    local res, err = conn:execute(cql, args)
+    local res, err = conn:execute(cql, args, opts)
     if not res then
       return nil, err
     end


bungle:
@mpenick, can you describe the situation or problem that you got? For which this PR is a fix?

bungle:
I mean it makes a lot of sense, but just looking our code, we don't pass the consistency in many use cases of `conn:execute`, so just thinking aloud that should we change this elsewhere too?

mpenick:
> can you describe the situation or problem that you got? For which this PR is a fix?

We were using Kong with DataStax Astra (Cassandra in the cloud) which has guardrails for consistency and doesn't allow the lua-cassandra default of `ONE`.  In general, for regular Cassandra, it also makes sense to default to at least `LOCAL_ONE`. This allows for setups where multiple clusters (DCs) are used. In that case, Kong probably only wants a response from a replica in the local region in which it's writing to.   Also, most drivers now default to `LOCAL_ONE` instead of `ONE`. It's a simple change that allows Kong (C* connecter) users to have a chance at a DR solution with multiple clusters (even if it's a manual process) and `LOCAL_ONE` is just `ONE` if only a single cluster is used.

> so just thinking aloud that should we change this elsewhere too?

Agreed, that could be, but this is the only instance I ran into where I saw a problem.  It might be nice to have some sort of abstraction that passes the correct consistency overrides for reads and writes, but that's a much larger change.

bungle (approved):

PR #8227, opened 2021-12-22
fix(rate-limiting) removes race condition(redis) for better accuracy

### LINKED ISSUE #5311: Ratelimit plugin not working as expected - kong 1.4

giovanibrioni:
### Summary

fix rate-limiting accuracy reported on issue: https://github.com/Kong/kong/issues/5311

There is a low impact change to the current code and this implementation has better accuracy on redis policy

### Full changelog

* [it removes the race condition problem getting keys on redis. it uses the return of the increment command as the usage counter ensuring atomicity, different from the current one, which allows multiple requests to get the same value of the redis key while the counter increment is done asynchronously]

* [it uses sliding window algorithm which is the most suitable for better accuracy]

* [it works only with redis polity because redis is capable of guaranteeing atomicity and performance, which other polices could not guarantee]

* [it maintains the current functioning and configuration of the plugin. the fixed window setting is the default to not impact routes already configured]

* [it has an even better performance than the current plugin using the redis policy and fixed window, as it does only 1 operation per request in redis (eval) instead of the two operations per request that the current one (get and eval)]

### Issues resolved

Fix #5311
KAG-981

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -1,6 +1,7 @@
 -- Copyright (C) Kong Inc.
 local timestamp = require "kong.tools.timestamp"
 local policies = require "kong.plugins.rate-limiting.policies"
+local kong_meta = require "kong.meta"
 
 
 local kong = kong
@@ -46,8 +47,8 @@ local X_RATELIMIT_REMAINING = {
 local RateLimitingHandler = {}
 
 
-RateLimitingHandler.PRIORITY = 901
-RateLimitingHandler.VERSION = "2.4.0"
+RateLimitingHandler.VERSION = kong_meta.version
+RateLimitingHandler.PRIORITY = 910
 
 
 local function get_identifier(conf)
@@ -191,7 +192,7 @@ function RateLimitingHandler:access(conf)
       if stop then
         headers = headers or {}
         headers[RETRY_AFTER] = reset
-        return kong.response.error(429, "API rate limit exceeded", headers)
+        return kong.response.error(conf.error_code, conf.error_message, headers)
       end
 
       if headers then
@@ -236,7 +237,7 @@ function RateLimitingHandler:access(conf)
       if stop then
         headers = headers or {}
         headers[RETRY_AFTER] = reset
-        return kong.response.error(429, "API rate limit exceeded", headers)
+        return kong.response.error(conf.error_code, conf.error_message, headers)
       end
 
       if headers then
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -122,84 +122,131 @@ function RateLimitingHandler:access(conf)
   -- Consumer is identified by ip address or authenticated_credential id
   local identifier = get_identifier(conf)
   local fault_tolerant = conf.fault_tolerant
+  local window_type = conf.window_type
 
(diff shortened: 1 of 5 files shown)


kong/plugins/rate-limiting/handler.lua
```diff
@@ -122,84 +122,131 @@ function RateLimitingHandler:access(conf)
...
+      local remaining = max(0, limit - current_usage)
+      local stop = current_usage > limit
+      local headers
+      if not conf.hide_client_headers then
+        headers = {}
+
+        headers[RATELIMIT_LIMIT] = limit
+        headers[RATELIMIT_REMAINING] = remaining
+        headers[RATELIMIT_RESET] = reset
+      end
```
esatterwhite, line 234: I think this is still a fixed window of time. The window is the TTL on a key, which never moves. So when the TTL expires after the fixed amount of time, the counter resets.

giovanibrioni: IMHO TTL is fixed but the window of time is not. 
For example, if the time unit is 1 minute, fixed window counter guarantees an average rate in window [00:00, 00:01), [00:01,00:02), etc.
Suppose the limit is 10 per minute. 
In fixed window if 10 requests arives at 00:00:59 and more 10 requests arives at 00:01:01 we have 20 alowed requests in the interval of 3 seconds.   
With redis TTL it is not possible. At the same cenario only the first 10 requests are allowed. The last 10 recives 429 status code. The window limit slides between time window [00:00, 00:01) and [00:01,00:02) 
But I'm open to suggestions

esatterwhite: I think you would have to keep timestamps of each request in a sorted set.
and on each request drop the timestamps between 0 and (now - window_time), add now. the count of items in the set is how many requests have been made in the window.
if the number is greater than the configured value, the limit has been reached

This would make the window sliding over the points in time as the requests are made.

esatterwhite:
> We have no plans to add a sliding window to this plugin. 

What's the reasoning here?

hbagdi:
Sliding window is an enterprise-only feature at this point.

esatterwhite:
> Sliding window is an enterprise-only feature at this point.

Can we implement an algorithm that isn't a sliding window?

hbagdi:
> Can we implement an algorithm that isn't a sliding window?

Could you elaborate a bit?

giovanibrioni:
@hbagdi, 
It's sad that the community can't implement the sliding window feature just because you decided it's an enterprise only feature.
But I can fix the race condition problem removing the sliding feature. Do you prefer I open a new PR or change this one?

giovanibrioni:
@hbagdi, I sent the PR below to fix the problem without sliding window feature:
https://github.com/Kong/kong/pull/10559

locao:
Closing in favor of #10559

PR #8232, opened 2021-12-23
fix(targets) reschedule resolve timer after resolving

locao:
As `ngx.timer.every` was used, if resolving an address took longer than 1 sec, another resolving callback would be called, instead of waiting until the first one finish. With this change, a new timer will be scheduled only after the previous one has executed.


javierguerragiraldez (approved):
i think this is the best "short term" fix.  eventually it would be better to use a background thread instead; but that has a few other complications, so it would be a longer term project.

PR #8233, opened 2021-12-24
fix(router) may be a bug in has_wildcard_host_port check

chronolaw:
### Summary

When I view the code in router.lua, 
I find the logic of MATCH_SUBRULES.HAS_WILDCARD_HOST_PORT may be a mistake.
if we have many hosts, the value of has_port may disappear in the next loop.

Please check the logic flow to confirm I am right or wrong.

### Full changelog

* add a new var has_wildcard_host_port
* use has_wildcard_host_port to determine subrules.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -376,7 +376,7 @@ local function marshall_route(r)
 
     local has_host_wildcard
     local has_host_plain
-    local has_port, has_wildcard_host_port
+    local has_wildcard_host_port
 
     for _, host in ipairs(hosts) do
       if type(host) ~= "string" then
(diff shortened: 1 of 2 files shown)


kong/router.lua
```diff
@@ -376,7 +376,7 @@ local function marshall_route(r)
 
     local has_host_wildcard
     local has_host_plain
-    local has_port
+    local has_port, has_wildcard_host_port
```
fffonion, line 379: in this case we should let `has_port` variable live inside of for loop, line 393 becomes
```lua
local _, _, has_port = split_port(host)
```
and this line we don't declare `has_port` only `has_wildcard_host_port`.

chronolaw: done.

bungle:
Is it possible to add a test to confirm that you are right or wrong? :-)

bungle (changes requested):
I'd love to have test on this.

chronolaw:
> Is it possible to add a test to confirm that you are right or wrong? :-)

I am glad to do this. Shall I edit the case `01-unit/08-router_spec.lua`?

bungle:
> > Is it possible to add a test to confirm that you are right or wrong? :-)
> 
> I am glad to do this. Shall I edit the case `01-unit/08-router_spec.lua`?

Yes, that is the right one!

bungle (approved):

PR #8236, opened 2021-12-27
perf(router) use ngx.ssl.server_name() instead of var.ssl_server_name

chronolaw:
### Summary

Openresty has a more efficient ffi function ngx.ssl.server_name() ,
it has the same result as var.ssl_server_name,
so I think we can use it both in stream and http subsystem.

### Full changelog

* use ngx.ssl.server_name() to get the value of  var.ssl_server_name

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -1829,13 +1829,15 @@ function _M.new(routes)
   self.select = find_route
   self._set_ngx = _set_ngx
 
+  local server_name = require("ngx.ssl").server_name
+
   if subsystem == "http" then
     function self.exec(ctx)
       local req_method = get_method()
       local req_uri = ctx and ctx.request_uri or var.request_uri
       local req_host = var.http_host or ""
       local req_scheme = ctx and ctx.scheme or var.scheme
-      local sni = var.ssl_server_name
+      local sni, _ = server_name()
 
       local headers
       local err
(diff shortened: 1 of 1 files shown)


fffonion:
@chronolaw Could you use a similar test script like https://github.com/Kong/kong/pull/7742#issuecomment-902616472 to validate the replacement from `var.ssl_server_name` to `ngx.ssl.server_name()` is indeed beneficial, since now we have
the ngx.var optimization in place?

chronolaw:
Here is the test code:

~~~test
# add "lua_kong_load_var_index $ssl_server_name;" in nginx.conf

local last_time = 0

local function timeit()
    ngx.update_time()
    local a = last_time
    last_time = ngx.now()
    return last_time - a
end

local name
local server_name = require("ngx.ssl").server_name

timeit()

local ITER = 5000*1000
for i=1,ITER do
  name = ngx.var.ssl_server_name
end
ngx.say(timeit())


for i=1,ITER do
  name = server_name()
end
ngx.say(timeit())

~~~

The results are:

~~~
curl --resolve ngx.test:8443:127.1 https://ngx.test:8443 -k

0.22900009155273
0.073999881744385

0.23200011253357
0.061000108718872
~~~

And we can get the conclusion that ffi is better than var mechanism.

fffonion (approved):

kong/router.lua
```diff
@@ -1829,13 +1829,15 @@ function _M.new(routes)
   self.select = find_route
   self._set_ngx = _set_ngx
 
+  local server_name = require("ngx.ssl").server_name
```
dndx, line 1832: We should be able to move this to the module level, since `require` it in the builder no longer makes sense. (It was done before probably to avoid penalty to HTTP subsystem as it was the main use case).

chronolaw: done.

dndx (approved):

PR #8239, opened 2021-12-27
docs: updating source compilation versions

subnetmarco:
Updating the dependency versions in our `DEVELOPER.md`. This also properly compiles on M1 Mac.


mayocream (approved):
Compile successfully on mac m1.

PR #8240, opened 2021-12-28
chore(*) update copyright year 2022

mayocream:
### Summary

Update copyright years.

ref: 
- https://github.com/Kong/kong/pull/7123
- https://github.com/Kong/kong/pull/6774

**Hold this until the new year comes.** 🎄

### Full changelog

* Update copyright years

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2016-2021 Kong Inc.
+   Copyright 2016-2022 Kong Inc.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


bungle (approved):

esatterwhite:
For what its worth, the copyright year should reference the year of the original public release, not the current year.
I think this technically voids the copyright of previous releases.

mayocream:
> For what its worth, the copyright year should reference the year of the original public release, not the current year. I think this technically voids the copyright of previous releases.
> 
> > The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work
> 
> You're probably better off leaving it as `copyrite 2016`

Thank you for noticing this.
I will merge first since other projects under Kong org. have similar LICENSE file, we can continue to discuss to reach the agreement.

I noticed the Kubernetes project removes year and author from their LICENSE file, that's interesting, see:
- https://github.com/kubernetes/kubernetes/pull/37305

PR #8255, opened 2022-01-02
feat(balancer) upstream healthcheck headers

mayocream:
### Summary

Add option to set upstream active healthcheck HTTP headers.

see: 
- https://github.com/Kong/lua-resty-healthcheck/pull/54

### Full changelog

* Add an option to set healthcheck HTTP headers
* Add related tests

### Issues resolved

ref:
- https://github.com/Kong/kong/discussions/7916

FT-2195


autodoc/admin-api/data/admin-api.lua
```diff
@@ -1571,6 +1571,7 @@ return {
         ["healthchecks.active.http_path"] = { description = [[Path to use in GET HTTP request to run as a probe on active health checks.]] },
         ["healthchecks.active.https_verify_certificate"] = { description = [[Whether to check the validity of the SSL certificate of the remote host when performing active health checks using HTTPS.]] },
         ["healthchecks.active.https_sni"] = { description = [[The hostname to use as an SNI (Server Name Identification) when performing active health checks using HTTPS. This is particularly useful when Targets are configured using IPs, so that the target host's certificate can be verified with the proper SNI.]], example = "example.com", },
+        ["healthchecks.active.headers"] = { description = [[An array of HTTP Headers to use in GET HTTP request to run as a probe on active health checks. Values must be pre-formatted.]], example = { "User-Agent: curl/7.64.0" }, },
```
bungle, line 1574: We have in my opinion two many ways to setup headers.
- array of strings
- we have this typedef (a map which key is string that has validator and values is array of strings)
- then we have prs where the above typedef seems to cause confusion (value is array): https://github.com/Kong/kong/pull/6992
- in OIDC plugin there are two fields, one which specifies headers and other which specifies values (for historical reasons)

It is known also that Kong Manager does not render complex fields nicely.

Still I would love to standardize on something. The `typedef.headers` looks the best we have currently...  but others  might disagree (e.g. the confusion point above)

mayocream: @bungle Nice, I saw the `typedef.headers` which is used in route config, but it can be a little more complex since we need to flatten headers map to an array to pass to resty-healthcheck.
I agree that we should standardize the HTTP headers type.

kong/db/schema/entities/upstreams.lua
```diff
@@ -73,6 +73,14 @@ local check_verify_certificate = Schema.define {
   required = true,
 }
 
+local check_headers = Schema.define {
+  type = "array",
```
bungle, line 77: I saw the `unique=true` in array, so was just thinking should this rather be a `set` then?  Is the order a requirement?

autodoc/admin-api/data/admin-api.lua
bungle: I see, makes sense then to keep as is (as you proposed). Thanks for further info on this.

flrgh: I'm a fan of keeping things consistent with `typedefs.headers` as well because it seems like a better format, and because it does more validation. What do you guys think of updating lua-resty-healthcheck so that it can handle both formats in a backwards compatible way before deprecating the array-of-strings form? I.E.

```lua
for k, v in pairs(headers) do
  if type(k) == "number" then
    -- `v` is {{header_name}}:{{header_value}}
    -- maybe log a deprecation warning for this
  else
    -- `k` is header name and `v` is header value
  end
end
```

The way headers are _currently_ handled in lua-resty-healthcheck looks a little unsafe to me anyways (potentially vulnerable to request smuggling or at least injection of invalid header data and/or unintended "extra" headers by using CRLF chars in one of the header strings), so at very least it might be prudent to make the code more defensive in handling the user input:
https://github.com/Kong/lua-resty-healthcheck/blob/e1f2196faa7d57f2ad62fa5d0ed2b59beeb8396d/lib/resty/healthcheck.lua#L921-L932

mayocream:
Thanks @flrgh, I've created following PR to lua-resty-heathcheck
- https://github.com/Kong/lua-resty-healthcheck/pull/87

mayocream:
The PR to upstream lib is merged, we will be waiting for lua-resty-heathcheck 1.5.x release.

mayocream:
The lua-resty-healthcheck version is bumped to 1.5.0, https://github.com/Kong/kong/commit/4ceda40533b4d6df07fa100b4782a7bf289f3f3d

It's ready to merge.

flrgh (approved):
:+1: lgtm!

I think it would be nice to see an integration test case using an upstream with healthcheck headers (even if we're just doubling up on testing the lua-resty-healthcheck API at that point), but IMO this is perfectly good-to-go as-is.

(Actually... @Hayk-S I wonder if the upstream healthcheck headers case would be a good candidate for E2E testing?)

Hayk-S:
@flrgh I think there will be no harm from additional tests. I will probably get back to you for more details on this.

PR #8263, opened 2022-01-06
fix(conf) check transfer-encoding header presence before adding content-length header

### LINKED ISSUE #8083: kong.response.exit setting Content-Length when Transfer-Encoding is present

2.5.1.0

### Current Behavior

When calling `kong.response.exit()`, the function automatically sets the Content-Length header on the response even if the Transfer-Encoding header is present. This is in violation of RFC 7230 3.3.2: 
> A sender MUST NOT send a Content-Length header field in any message
   that contains a Transfer-Encoding header field.

### Expected Behavior

`kong.response.exit()` should not set Content-Length if the Transfer-Encoding header is present in the header list provided as the third method argument.

### Steps To Reproduce

```markdown
1. Create a new plugin or edit an existing one. In the access phase, call `kong.response.exit(200, {}, { ["Transfer-Encoding"] = "chunked" }`
2. Via the API, create an instance of the plugin and send it a request
3. Observe that the response contains both the Transfer-Encoding and Content-Length headers
```

### Anything else?

_No response_

SakshamDevelops:
### **Summary**
When calling kong.response.exit(), the function automatically sets the Content-Length header on the response even if the Transfer-Encoding header is present. This is in violation of RFC 7230 3.3.2:

> A sender MUST NOT send a Content-Length header field in any message
> that contains a Transfer-Encoding header field.
> 

### Full changelog

* If the transfer-encoding header is present in the response then assign the variable `has_content_length = true` in the function `send(status, body, headers)`.
* Add an extra test case on the function's specs.

### Issue reference

[Fix #8083](https://github.com/Kong/kong/issues/8083)

### DIFF
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -640,21 +640,22 @@ local function new(self, major_version)
     ngx.status = status
 
     local has_content_type
-    local has_content_length
+    local has_content_length_or_chunked
     if headers ~= nil then
       for name, value in pairs(headers) do
         ngx.header[name] = normalize_multi_header(value)
-        if not has_content_type or not has_content_length then
+        if not has_content_type or not has_content_length_or_chunked then
           local lower_name = lower(name)
           if lower_name == "content-type"
-          or lower_name == "content_type"
+            or lower_name == "content_type"
           then
             has_content_type = true
           elseif lower_name == "content-length"
-              or lower_name == "content_length"
-              or lower_name == "transfer-encoding"
-              or lower_name == "transfer_encoding" then
-            has_content_length = true
+            or lower_name == "content_length"
+            or ((lower_name == "transfer-encoding" or
+                  lower_name == "transfer_encoding") and
+                  value == 'chunked')  then
+            has_content_length_or_chunked = true
           end
         end
       end
@@ -722,7 +723,7 @@ local function new(self, major_version)
         ngx.header[CONTENT_TYPE_NAME] = CONTENT_TYPE_JSON
       end
 
-      if not has_content_length then
+      if not has_content_length_or_chunked then
         ngx.header[CONTENT_LENGTH_NAME] = #json
       end
 
@@ -746,7 +747,7 @@ local function new(self, major_version)
         end
 
       else
-        if not has_content_length then
+        if not has_content_length_or_chunked then
           ngx.header[CONTENT_LENGTH_NAME] = #body
         end
 
@@ -763,7 +764,7 @@ local function new(self, major_version)
       end
 
     else
-      if not has_content_length then
+      if not has_content_length_or_chunked then
         ngx.header[CONTENT_LENGTH_NAME] = 0
       end
 
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -651,7 +651,9 @@ local function new(self, major_version)
           then
             has_content_type = true
           elseif lower_name == "content-length"
-              or lower_name == "content_length" then
+              or lower_name == "content_length"
+              or lower_name == "transfer-encoding"
+              or lower_name == "transfer_encoding" then
             has_content_length = true
           end
         end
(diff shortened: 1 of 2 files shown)


mayocream (commented):
Thanks for your first PR to Kong! Don't forget to take your [contributor T-shirt](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt) after this PR got merged.

mayocream:
@SakshamDevelops, Can you look at the failed PDK tests and fix them?

SakshamDevelops:
Hi @mayocream, trying to fix those. It would be great if you guide me on how to approach this?

jschmid1:
@SakshamDevelops see this error -> https://github.com/Kong/kong/runs/4786940526?check_suite_focus=true#step:7:168

Suika-Kong:
Your fix is incorrect. Only when the encoding is "chunked", the `Content-Length` can be omitted.

Suika-Kong:
> 

Sorry, I read the reference of RFC and will do more research on it. I'm surprised that when `Transfer-Encoding: gzip`, `Content-Length` should also be excluded.

Suika-Kong:
I found that your PR is correct. It is the OpenResty behaves oddly when Content-Length is not set.
I will try to fix that.

Suika-Kong:
https://github.com/openresty/lua-nginx-module/issues/2027

bungle:
@Suika-Kong,

Can this be work-arounded? When there is `Transfer-Encoding` we don't set `Content-Length` headers. Also if the value of `Transfer-Encoding` header is `chunked` we don't set that header either (as OpenResty then sets it).

Suika-Kong:
> 

For this special case, yes. But it will still be a problem if user try to set `Transfer-Encoding` to gzip or something like that.

bungle:
> For this special case, yes. But it will still be a problem if a user tries to set `Transfer-Encoding` to gzip or something like that.

@Suika-Kong,

Can we in that case just keep it? There are cases:

- Transfer-Encoding: nil -> set Content-Length
- Transfer-Encoding: chunked -> do not set Content-Length
- Transfer-Encoding: gzip chunked -> do not set Content-Length and set Transfer-Encoding: gzip (only)
- Transfer-Encoding: chunked gzip -> do not set Content-Length and set Transfer-Encoding: gzip (only)
- Transfer-Encoding: gzip -> set Content-Length

Is that the correct matrix?

Suika-Kong:
> > For this special case, yes. But it will still be a problem if a user tries to set `Transfer-Encoding` to gzip or something like that.
> 
> @Suika-Kong,
> 
> Can we in that case just keep it? There are cases:
> 
> * Transfer-Encoding: nil -> set Content-Length
> * Transfer-Encoding: chunked -> do not set Content-Length, do not set Transfer-Encoding (OpenResty sets it)
> * Transfer-Encoding: gzip chunked -> do not set Content-Length and set Transfer-Encoding: gzip (only)
> * Transfer-Encoding: chunked gzip -> do not set Content-Length and set Transfer-Encoding: gzip (only)
> * Transfer-Encoding: gzip -> set Content-Length and set Transfer-Encoding: gzip (as given)
> 
> Is that the correct matrix?

No. When you set `Transfer-Encoding: gzip`, you will get `chunked, gzip`. There's no way you can get `gzip, chunked` for the current implementation of OpenResty.

Suika-Kong:
OpenResty seems not to respond to the issue soon, and I failed to locate the bug.

For this case especially, we do have a workaround(as @bungle suggests). If we set neither `Content-Length` nor `Transfer-Encoding` through OpenResty API, we will get `chunked`. Other case(like gzip) is rare.

OpenResty does not work with what you set in `Transfer-Encoding`. If you set the header, the client will fail to read the body. I do not know will they consider this, so maybe proper handling would be:
1. Tell users not to set `Transfer-Encoding` manually, so they will not expect gzip to work. 
2. Think about how we can offer users the ability to use chunked

kikito:
Apart from doing workarounds in Lua-land or waiting for Openresty to fix the issue (which can take a lot of time) there's a third option: we could debug the problem ourselves and include the relevant changes in our openresty-patches. And then send a PR to openresty.

Suika-Kong:
Issue fixed by #8698.

PR #8264, opened 2022-01-06
RFC: feat(zipkin) Use tags instead of phases for annotations

### LINKED ISSUE #8126: Rationale for logs/annotations instead of spans?
Can you explain the rationale for using logs/annotations (krs... 10 total) instead of 5 additionnal spans? 
The performance cannot be considered top priority, as you just need to lower your sampling rate.
Is it visualized similarly in zipkin, i.e. different spans vs multiple annotations?
What about searching for annotations?
We are using Jaeger as backend, and feel it will be hard to investigate any timing issues in kong plugins, I don't think I can search for the "time difference between annotations >= XX milliseconds" anywhere even directly in ES storage of Jaeger.
Maybe you will answer that this is more of a metric type of search.

This is what a kong-zipkin-plugin trace looks like in Jaeger-ui, you have to expand to see them, they do not appear in the timeline, so we were a bit underwhelmed I must say:
L_B8D5 tmp

Is it worth a PR with a configuration choice between multiple spans vs annotations?

Thanks for your answers.
Regards.

dcarley:
I'd like feedback on this approach and some related questions about the other spans before I tackle the accompanying tests and documentation.

### Summary

The previous annotations made it difficult to see how long a phase had
taken because you had to compute the difference between "start" and
"finish" yourself.

The annotations are also expensive for users of Honeycomb because their
billing model charges for each event, of which there were typically 8
phase events in an HTTP request, plus the request+proxy+balancer spans.

As discussed in issue 8126, these were spans in the past and changing
them back would address the timing problems. It would also half the
number of billable events for Honeycomb users because the start+finish
wouldn't need to be recorded separately. However there would still be at
least 4+3 billable events for an HTTP request.

Whereas there is no charge for additional tags on existing spans and you
can still query them. We can also default them to 0 if they don't exist
at the log phase, so they become easier to compute.

Plus give them more complete names.

### Full changelog

TBD

### Issue reference

Fix #8126

### DIFF
--- a/kong/plugins/zipkin/handler.lua
+++ b/kong/plugins/zipkin/handler.lua
@@ -174,19 +174,7 @@ if subsystem == "http" then
       request_span = request_span,
       header_type = header_type,
       proxy_span = nil,
-      header_filter_finished = false,
     }
-  end
-
-
-  function ZipkinLogHandler:rewrite(conf) -- luacheck: ignore 212
-    local zipkin = get_context(conf, kong.ctx.plugin)
-    local ngx_ctx = ngx.ctx
-    -- note: rewrite is logged on the request_span, not on the proxy span
-    local rewrite_start_mu =
-      ngx_ctx.KONG_REWRITE_START and ngx_ctx.KONG_REWRITE_START * 1000
-      or ngx_now_mu()
-    zipkin.request_span:annotate("kong.rewrite.start", rewrite_start_mu)
   end
 
 
@@ -210,22 +198,7 @@ if subsystem == "http" then
       ngx_ctx.KONG_HEADER_FILTER_STARTED_AT and ngx_ctx.KONG_HEADER_FILTER_STARTED_AT * 1000
       or ngx_now_mu()
 
-    local proxy_span = get_or_add_proxy_span(zipkin, header_filter_start_mu)
-    proxy_span:annotate("kong.header_filter.start", header_filter_start_mu)
-  end
-
-
-  function ZipkinLogHandler:body_filter(conf) -- luacheck: ignore 212
-    local zipkin = get_context(conf, kong.ctx.plugin)
-
-    -- Finish header filter when body filter starts
-    if not zipkin.header_filter_finished then
-      local now_mu = ngx_now_mu()
-
-      zipkin.proxy_span:annotate("kong.header_filter.finish", now_mu)
-      zipkin.header_filter_finished = true
-      zipkin.proxy_span:annotate("kong.body_filter.start", now_mu)
-    end
+    get_or_add_proxy_span(zipkin, header_filter_start_mu)
   end
 
 elseif subsystem == "stream" then
@@ -265,8 +238,7 @@ elseif subsystem == "stream" then
       ngx_ctx.KONG_PREREAD_START and ngx_ctx.KONG_PREREAD_START * 1000
       or ngx_now_mu()
 
-    local proxy_span = get_or_add_proxy_span(zipkin, preread_start_mu)
-    proxy_span:annotate("kong.preread.start", preread_start_mu)
+    get_or_add_proxy_span(zipkin, preread_start_mu)
   end
 end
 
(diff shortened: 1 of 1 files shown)


kong/plugins/zipkin/handler.lua
```diff
@@ -283,40 +266,8 @@ function ZipkinLogHandler:log(conf) -- luacheck: ignore 212
...
-    proxy_span:annotate("kbf", now_mu)
-
-  else
-    local preread_finish_mu =
-      ngx_ctx.KONG_PREREAD_ENDED_AT and ngx_ctx.KONG_PREREAD_ENDED_AT * 1000
-      or proxy_finish_mu
-    proxy_span:annotate("kpf", preread_finish_mu)
-  end
+  tag_with_durations(request_span)
+  tag_with_durations(proxy_span)
```
dcarley, line 270: There could be a config option to use tags or spans, however it would increase the complexity of the code and tests quite a bit.

kong/plugins/zipkin/handler.lua
```diff
@@ -335,6 +286,7 @@ function ZipkinLogHandler:log(conf) -- luacheck: ignore 212
         span:set_tag("http.status_code", try.code)
       end
 
+      tag_with_durations(span)
```
dcarley, line 289: Should the balancer span be a child of the proxy span?

My understanding is that it represents the time to assign an upstream from the balancer but not actually attempt the upstream request. Would that ever take more than a few milliseconds? If not, would the number of tries and previous status codes be better represented as tags on the proxy span?

kong/plugins/zipkin/handler.lua
```diff
@@ -174,22 +185,10 @@ if subsystem == "http" then
...
-    local ngx_ctx = ngx.ctx
-    -- note: rewrite is logged on the request_span, not on the proxy span
-    local rewrite_start_mu =
-      ngx_ctx.KONG_REWRITE_START and ngx_ctx.KONG_REWRITE_START * 1000
-      or ngx_now_mu()
-    zipkin.request_span:annotate("krs", rewrite_start_mu)
-  end
-
-
   function ZipkinLogHandler:access(conf) -- luacheck: ignore 212
```
dcarley, line 192: Why does the proxy span include access, header_filter, and body_filter?

It would be useful to visuablise the time spent talking to the upstream, like the kong/upstream metrics from the Datadog plugin, to help determine which service was holding up the request. Then dig into the separate phases from there.

kong/plugins/zipkin/handler.lua
mayocream: 1. I think it make sense to put balancer spans as child span of the proxy span.

2. The `balancer_data.tries` represents every request to the upstream, the request can take much longer time if the upstream server is slow to respond or failed to respond, in that case, the way to store these data in spans is more friendly to end-users.

kong/plugins/zipkin/handler.lua
```diff
@@ -45,6 +45,17 @@ local function get_reporter(conf)
...
 
 
+local function tag_with_durations(span)
+  ctx = ngx.ctx
+  span:set_tag("kong.rewrite.duration_ms", ctx.KONG_REWRITE_TIME or 0)
+  span:set_tag("kong.preread.duration_ms", ctx.KONG_PREREAD_TIME or 0)
+  span:set_tag("kong.access.duration_ms", ctx.KONG_ACCESS_TIME or 0)
+  span:set_tag("kong.balancer.duration_ms", ctx.KONG_BALANCER_TIME or 0)
+  span:set_tag("kong.header_filter.duration_ms", ctx.KONG_HEADER_FILTER_TIME or 0)
+  span:set_tag("kong.body_filter.duration_ms", ctx.KONG_BODY_FILTER_TIME or 0)
```
mayocream, line 55: Some phases are not available in stream mode, better not to include spans with an empty value.

mayocream (changes requested):

mayocream (commented):
Hi @dcarley, 

Nice work 👍 . I think using tags instead of annotations is more straightforward for end-users.

But I think the "0" value duration tags are unnecessary, especially body_filter, header_filer durations in stream mode. 

I think we can add an option (e.g. `net.stream.mode`) to indicate which mode the worker is running in, and only add related phase durations to span tags.

Here are the references for tag names: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes

Could you also add or change the related tests for this one?

dcarley:
Thanks for the feedback. I'll rework this soon based on the new OTel support.

hbagdi:

hbagdi:
Re-open due to an out of band request to pursue this again.

PR #8265, opened 2022-01-06
chore(*) support dev container

mayocream:
### Summary

Add `devcontainer.json` to support Visual Studio Code Remote - Containers and GitHub Codespaces.

ref: https://code.visualstudio.com/docs/remote/containers

Tested locally.

To test: https://code.visualstudio.com/docs/remote/containers#_quick-start-try-a-development-container


mayocream:
Running well on my Mac M1 with VSCode containers. 🎉

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM kong/kong:2.7.0
```
jschmid1, line 1: since the purpose of the dev container is to work on the actual source, shouldn't we maybe use a plain alpine image to avoid conflicts?

jschmid1 (commented):
tested - works.

Are the whitespace changes in the DEVELOPER.md file intended? If so this should maybe go into a separate PR(or commit)

mayocream:
> tested - works.
> 
> Are the whitespace changes in the DEVELOPER.md file intended? If so this should maybe go into a separate PR(or commit)

Okay, the whitespace changes can be placed in a new PR.

.devcontainer/Dockerfile
mayocream: Using a plain alpine or ubuntu image spends more time compiling the dependencies.
We have the `postCreateCommand` option which will be executed in the initial phase of the new container, it will overwrite Kong's LuaRocks packages. 
https://github.com/Kong/kong/blob/27d10142ce6f4167fc0da3cd4d6ca9d65768ecee/.devcontainer/devcontainer.json#L19

Gojira also use this method as an option to start a dev environment: `gojira up --image kong:2.7.0`, see: https://github.com/Kong/gojira/blob/master/docs/manual.md#using-kong-release-images-with-gojira

Other popular projects like dapr also use an existing docker image to start:
https://github.com/dapr/dapr/blob/10784ac6e5886fb4242eac73c4f13a8962d17043/.devcontainer/devcontainer.json#L4

bungle (approved):

.devcontainer/Dockerfile
jschmid1: ye, I saw that. I'm just mildly concerned that stray artifacts have an impact on the sources. 

gojira, when used without the `--image` flag, it also uses a plain _ubuntu_ image. https://github.com/Kong/gojira/blob/master/docker/Dockerfile#L1

Either way, if you don't think my concerns are valid, I'll remain silent and approve this PR :)

PR #8269, opened 2022-01-07
fix(router) remove last dot in FQDNs

### LINKED ISSUE #7550: Trailing dots in upstream hostnames are incorrectly included in TLS SNI extension
### Summary
- Kong version 2.4.1
- A concise summary of the bug.

### Steps To Reproduce

1. Create an API with a https upstream that includes a trailing dot in the hostname. For example `https://foobar.example.com.`
2. Make a request to the API
3. Observe that the SNI extension in the TLS ClientHello message to the upstream server includes the trailing dot

### Additional Details & Logs
Including the trailing dot informs Kong that the hostname is fully qualified and to not attempt to add search domains (as per #7070 this appears to be the only way to set this behaviour, there is no global setting that would achieve the same effect).

RFC3546 (TLS Extensions) specifically defines the behaviour of the SNI extension must not include the trailing dot:
```
   "HostName" contains the fully qualified DNS hostname of the server,
   as understood by the client. The hostname is represented as a byte
   string using UTF-8 encoding [UTF8], without a trailing dot.
```

The end result is that some more strict upstream servers will flat out reject any requests that include a trailing dot in the SNI field, for example Golang's TLS implementation will reject such connections resulting in the following in the Kong logs:

```
SSL_do_handshake() failed (SSL: error:140943F2:SSL routines:ssl3_read_bytes:sslv3 alert unexpected message:SSL alert number 10) while SSL handshaking to upstream
```

A quick Google indicates other server implementations may have similar issues, for example older versions of libcurl had this same issue and resulted in issues with Apache https://curl-library.cool.haxx.narkive.com/qtEDMpA5/trailing-dot-in-fqdn-tls-sni-http-hostname

There may also be issues with validating the hostname in upstream as well (certificate names do not contain a trailing dot either), but I was unable to verify if that was the case as the upstream rejected the connection before sending Kong the certificate.

locao:
### Summary

When a route has an SNI attribute containing an FQDN ending in a dot, the trailing dot was being used to match routes. But according to [RFC-3546](https://datatracker.ietf.org/doc/html/rfc3546#section-3.1), this trailing dot is not part of the hostname, thus must not be used for matching.

This change updates the route marshaling function to not include the trailing dot in this case.

### Full changelog

* Remove trailing dot from SNIs in route marshaling function.
* Added test for specific case.

### Issue reference

Fix #7550
CT-236

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -38,6 +38,7 @@ local server_name   = require("ngx.ssl").server_name
 local REGEX_PREFIX  = "(*LIMIT_MATCH=10000)"
 local SLASH         = byte("/")
 
+local DEBUG         = ngx.DEBUG
 local ERR           = ngx.ERR
 local WARN          = ngx.WARN
 
@@ -621,6 +622,11 @@ local function marshall_route(r)
       for _, sni in ipairs(snis) do
         if type(sni) ~= "string" then
           return nil, "sni elements must be strings"
+        end
+        
+        if sni:len() > 1 and sni:sub(-1) == "." then
+          log(DEBUG, "last dot in FQDNs must not be used for routing, removing in ", sni)
+          sni = sni:sub(1, -2)
         end
 
         route_t.match_rules = bor(route_t.match_rules, MATCH_RULES.SNI)
--- a/spec/02-integration/05-proxy/06-ssl_spec.lua
+++ b/spec/02-integration/05-proxy/06-ssl_spec.lua
@@ -518,8 +518,6 @@ for _, strategy in helpers.each_strategy() do
   end)
 
   describe("TLS proxy [#" .. strategy .. "]", function()
-    local https_client
-
     lazy_setup(function()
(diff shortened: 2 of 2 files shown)


kong/router.lua
```diff
@@ -622,6 +623,11 @@ local function marshall_route(r)
         if type(sni) ~= "string" then
           return nil, "sni elements must be strings"
         end
+        
+        if sni:len() > 1 and sni:sub(-1) == "." then
+          log(DEBUG, "last dot in FQDNs must not be used for routing, removing in ", sni)
+          sni = sni:sub(1, -2)
```
bungle, line 629: Does the `trailing slash` have meaning elsewhere outside `routing`? Just thinking that can we hit this issue elsewhere? Or is it just about routing?

bungle: Ah, trailing dot makes it FQDN, otherwise it may be appended with, well whatever in search.

bungle (approved):

PR #8274, opened 2022-01-10
tests(perf) add env to adjust perf test load duration

chronolaw:
### Summary

Sometime we want to shorter or longer time for perf test,
the original load duration is fixed to 30s in code,
now we can use env to adjust this value

### Full changelog

* check env PERF_TEST_LOAD_DURATION in perf test


kikito:
lgtm!

PR #8275, opened 2022-01-10
tests(flamegraph) fix generate flamegraph in local driver

chronolaw:
### Summary

When generate flame graph in local drive,
The `cat` cmd will dump the svg into logger but not stdout,
so we should check the cmd, if it is cat then use an empty opts.

### Full changelog

* add a flag is_cat_cmd
* check flag to determine opts


fffonion (approved):
minor lint issue, otherwise LGTM 👍

PR #8280, opened 2022-01-11
fix(extplugin) add type unwrap functions

### LINKED ISSUE #8279: Client.LoadConsumer not work with pb-rpc

2.7.0

### Current Behavior

`PDK.Client.LoadConsumer()` always complains `kong/runloop/plugin_servers/pb_rpc.lua:221: consumer_id must be a string,  ` no matter what args are passed.

### Expected Behavior

`PDK.Client.LoadConsumer()` loads the consumer

### Steps To Reproduce

```markdown
1. write a plugin with https://github.com/Kong/go-pdk
2. use `PDK.Client.LoadConsumer()` somewhere
3. enable the plugin
4. 500 response on every request
```

### Anything else?

https://github.com/Kong/kong/blob/master/kong/pdk/client.lua#L182 accepts two args

https://github.com/Kong/kong/blob/master/kong/pluginsocket.proto#L159 defines a wrapper type `ConsumerSpec` with two args

There is no unwrap function for `ConsumerSpec` defined  in https://github.com/Kong/kong/blob/master/kong/runloop/plugin_servers/pb_rpc.lua , therefore the function is invoked with wrong args.

raptium:
### Summary

ConsumerSpec and AuthenticateArgs type defined in proto need unwrapping, otherwise
`client.load_consumer()` and `client.authentication()` will not work with external plugin via pb-rpc protocol.

### Full changelog

* Add type unwrap functions for ConsumerSpec and AuthenticateArgs

### Issue reference

Fix #8279

### DIFF
--- a/kong/runloop/plugin_servers/pb_rpc.lua
+++ b/kong/runloop/plugin_servers/pb_rpc.lua
@@ -82,6 +82,12 @@ do
     [".kong_plugin_protocol.ExitArgs"] = function (d)
       return d.status, d.body, structpb_struct(d.headers)
     end,
+    [".kong_plugin_protocol.ConsumerSpec"] = function (d)
+      return d.id, d.by_username
+    end,
+    [".kong_plugin_protocol.AuthenticateArgs"] = function (d)
+      return d.consumer, d.credential
+    end,
   }
 end
 


javierguerragiraldez (approved):
Yes, that exactly is.  nice analysis, nice patch.

bungle (approved):

PR #8287, opened 2022-01-12
fix(docs) Clean up and fix PDK docs

lena-larionova:
### Summary
The comments in the PDK files get generated into docs on docs.konghq.com: https://docs.konghq.com/gateway/2.7.x/pdk/ . This PR addresses issues in the generated content, as well as a general review of the text.

Cleaning up the PDK docs in a few ways:
* Grammar, phrasing, and formatting review.
* Fixing all old links to point to real content. There were still links to the "getkong" domain, which hasn't been around for a while.
* Escaping pipe characters in places where they generate undesired markdown tables (example of broken output [here](https://docs.konghq.com/gateway/2.7.x/pdk/kong.response/#kongresponseadd_headername-value))
* Updating the ldoc file to:
  * Remove redundant "Back to Top" links. The docs site has a scroll to top button built in, and a right-side nav to scroll to any header on the page.
  * Remove first header (module title) as it is redundant with the page title, and adjust heading levels accordingly. See any PDK doc (eg [here](https://docs.konghq.com/gateway/2.7.x/pdk/kong.response/)) for an example of the redundancy.

Note that I skipped the service module, as there's another [open PR](https://github.com/Kong/kong/pull/8205) that would conflict.

Preview of fixed output on the docs site: https://deploy-preview-3546--kongdocs.netlify.app/gateway/2.7.x/pdk/

### Full changelog

Not sure if needed? I'm going to say probably not.

### Issue reference
N/A


falondarville (approved):
LGTM

kong/pdk/init.lua
```diff
@@ -1,17 +1,13 @@
...
--- The Plugin Development Kit (or "PDK") is set of Lua functions and variables
+-- The Plugin Development Kit (PDK) is set of Lua functions and variables
 -- that can be used by plugins to implement their own logic. The PDK is a
 -- [Semantically Versioned](https://semver.org/) component, originally
--- released in Kong 0.14.0. The PDK will be guaranteed to be forward-compatible
--- from its 1.0.0 release and on.
---
--- As of this release, the PDK has not yet reached 1.0.0, but plugin authors
--- can already depend on it for a safe and reliable way of interacting with the
--- request, response, or the core components.
```
lena-larionova, line 10: Note that I removed this info because it looks like the PDK is now on 1.4. Is the new text accurate?

kong/pdk/init.lua
```diff
@@ -147,7 +143,7 @@
 -- Instance of Kong's DNS resolver, a client object from the
 -- [lua-resty-dns-client](https://github.com/kong/lua-resty-dns-client) module.
 --
--- **Note:** usage of this module is currently reserved to the core or to
+-- **Note:** Usage of this module is currently reserved to the core or to
 -- advanced users.
```
lena-larionova, line 147: Is this still true? (asking for all such notes in this file)

kong/pdk/client.lua
```diff
@@ -167,15 +168,15 @@ local function new(self)
 
   ---
   -- Returns the consumer from the datastore.
-  -- Will look up the consumer by id, and optionally will do a second search by name.
+  -- Looks up the consumer by ID, and can optionally do a second search by name.
   -- @function kong.client.load_consumer
   -- @phases access, header_filter, response, body_filter, log
-  -- @tparam string consumer_id The consumer id to look up.
+  -- @tparam string consumer_id The consumer ID to look up.
   -- @tparam[opt] boolean search_by_username. If truthy,
```
lena-larionova, line 175: Is `search_by_username.` supposed to have a `.` at the end, or is it a typo? Same goes for `allow_terminated.` in this file. There are no other functions anywhere in the PDK doc that have the `.` at the end, so I'd like to double-check.

flrgh: It looks like a typo/stylistic oversight to my eyes. I vote :+1: for removing it.

kong/pdk/client/tls.lua
```diff
@@ -1,6 +1,6 @@
 ---
--- The client.tls module provides functions for interacting with TLS
--- connections from client.
+-- A module that provides functions for interacting with TLS
```
flrgh, line 2: Nit: I think the "A module that..." phrasing feels a little inconsistent with the doc strings on other PDK modules, which seem to (quite loosely) follow this kind of template:

```
---
-- {{ title }} (e.g. "Client TLS module")
--
-- {{ description sentence/fragment: what it does and/or what it contains }} (e.g. "Functions for interacting with client TLS connections." or similar)
-- @module {{ module name }}
```

Examples:
* [kong.pdk.response](https://github.com/Kong/kong/blob/86439965e5fe8101c6a468d5674710a9d0f785b4/kong/pdk/response.lua#L1-L7)
* [kong.pdk.client](https://github.com/Kong/kong/blob/86439965e5fe8101c6a468d5674710a9d0f785b4/kong/pdk/client.lua#L1-L7)
* [kong.pdk.router](https://github.com/Kong/kong/blob/86439965e5fe8101c6a468d5674710a9d0f785b4/kong/pdk/router.lua#L1-L4)

kong/pdk/ctx.lua
```diff
@@ -14,19 +14,19 @@ local _CTX_CORE_KEY = {}
 
 
 ---
--- A table that has the lifetime of the current request and is shared between
+-- A table that contains the history of the current request and is shared between
```
flrgh, line 17: I actually think the previous phrasing is more appropriate/accurate here. To explain: the term "lifetime" as used here refers to the programmatic scope of the table's data (it is in a sense returned to a clean slate for each request). It's not meant to imply that the table holds any structured historical data about the request (although that would be valid use-case of the table).

kong/pdk/init.lua
flrgh: :+1: I'd say it's still true.

kong/pdk/log.lua
```diff
@@ -459,20 +459,19 @@ end
 --
 -- Where:
 --
--- * `%file_src`: is the file name from where the log was called from.
--- * `%func_name`: is the name of the function from where the log was called
---   from.
--- * `%line_src`: is the line number from where the log was called from.
--- * `%message`: is the message, made of concatenated, pretty-printed arguments
+-- * `%file_src`: The filename from where the log was called.
```
flrgh, line 462: Nit: the way we've changed these makes the `Where:` on line 460 a little grammatically awkward, since the implied way of reading it to oneself was something like:

> Where `%func_name` is the file name from where the log was called from.
> and
> `%func_name` is the name of the function from where the log was called from.
> [...]

So with the new phrasing, I think it might make sense to drop the `Where:`. What do you think?

kong/pdk/log.lua
```diff
@@ -549,25 +548,24 @@ local function get_default_serialize_values()
 end
 
 ---
--- Sets a value to be used on the `serialize` custom table
+-- Sets a value to be used on the `serialize` custom table.
 --
 -- Logging plugins use the output of `kong.log.serialize()` as a base for their logs.
+-- This function let you customize the log output.
```
flrgh, line 554:
```suggestion
-- This function lets you customize the log output.
```

kong/pdk/log.lua
```diff
@@ -549,25 +548,24 @@ local function get_default_serialize_values()
...
 -- phase in most real-usage cases.
 --
 -- @function kong.log.set_serialize_value
 -- @phases certificate, rewrite, access, header_filter, response, body_filter, log
--- @tparam string key the name of the field.
--- @tparam number|string|boolean|table value value to be set. When a table is used, its keys must be numbers, strings, booleans, and its values can be numbers, strings or other tables like itself, recursively.
--- @tparam table options can contain two entries: options.mode can be `set` (the default, always sets), `add` (only add if entry does not already exist) and `replace` (only change value if it already exists).
--- @treturn table the request information table
+-- @tparam string key The name of the field.
+-- @tparam number\|string\|boolean\|table value Value to be set. When a table is used, its keys must be numbers, strings, or booleans, and its values can be numbers, strings, or other tables like itself, recursively.
```
flrgh, line 566: Is there a reason these `|` chars are being escaped like this? I ask because the syntax looks correct to me as-is, where `@tparam a|b|c` implies a union type ("the function accepts type `a`, type `b`, or type `c` for this parameter").

kong/pdk/service/request.lua
```diff
@@ -279,13 +281,13 @@ local function new(self)
...
-  -- will also set the SNI of the request to the Service.
+  -- If the `header` argument is `"host"` (case-insensitive), then this also
+  -- sets the SNI of the request to the Service.
   --
   -- @function kong.service.request.set_header
   -- @phases `rewrite`, `access`, `balancer`
-  -- @tparam string header The header name. Example: "X-Foo"
-  -- @tparam string|boolean|number value The header value. Example: "hello world"
+  -- @tparam string header The header name. Example: "X-Foo".
+  -- @tparam string\|boolean\|number value The header value. Example: "hello world".
```
flrgh, line 290: Same q as above regarding the escaped (`\|`) pipe chars.

kong/pdk/service/request.lua
```diff
@@ -314,14 +316,14 @@ local function new(self)
...
+  -- `kong.service.request.set_header()`, this function doesn't remove any existing
   -- headers with the same name. Instead, several occurrences of the header will be
   -- present in the request. The order in which headers are added is retained.
   --
   -- @function kong.service.request.add_header
   -- @phases `rewrite`, `access`
-  -- @tparam string header The header name. Example: "Cache-Control"
-  -- @tparam string|number|boolean value The header value. Example: "no-cache"
+  -- @tparam string header The header name. Example: "Cache-Control".
+  -- @tparam string\|number\|boolean value The header value. Example: "no-cache".
```
flrgh, line 326: Same q as above regarding the escaped (`\|`) pipe chars.

kong/pdk/response.lua
```diff
@@ -1034,29 +1039,31 @@ local function new(self, major_version)
...
+  -- is seen by the client. The status code must an error code, that is,
+  -- greater than 399.
   --
-  -- The second, optional, `message` argument will set the message describing
-  -- the error, which will be written in the body.
+  -- 1. The optional `message` argument sets the message describing
+  -- the error, which is written in the body.
   --
-  -- The third, optional, `headers` argument can be a table specifying response
+  -- 1. The optional `headers` argument can be a table specifying response
```
flrgh, line 1049: Should these be `2.` and `3.`?

flrgh (changes requested):
There are some style/grammar nits that are more-or-less just my opinions and can be ignored.

The 2 things that I think should be addressed before merging are:

* the `kong.ctx` lifetime-vs-history comment
* the instances where `@tparam` type unions have been escaped from `a|b|c` to `a\|b\|c` (note I didn't necessarily comment on all instances of this).

kong/pdk/log.lua
lena-larionova: These comments get converted to markdown for the doc output. If they aren't escaped or enclosed in backticks, any time that `|` gets used more than once in a line generates a table row. For example, see the value entry here: https://docs.konghq.com/gateway/2.7.x/pdk/kong.response/#kongresponseadd_headername-value

kong/pdk/response.lua
lena-larionova: They don't need to be, the generator sees a numbered list and numbers it accordingly. So it doesn't matter what numbers are used here. In the docs, we're trying to use only `1.` going forward, so that switching steps around, removing, or adding steps, doesn't change the list in the output - here's the list in question in the preview: https://deploy-preview-3546--kongdocs.netlify.app/gateway/2.7.x/pdk/kong.response/#kongresponseexitstatus-body-headers

However - in this case, since the code comments are seen by PDK users, I can see an argument for actually numbering them 1, 2, 3.

kong/pdk/log.lua
lena-larionova: Nice catch. This one was weird, I was struggling to phrase it in a cleaner way. How about:

> %file_src`: The filename the log was called from.

Bringing back the closing "from", but makes more sense in the phrase "Where `%file_src` [is] the filename the log was called from."

flrgh: > %file_src`: The filename the log was called from.

I like this!

kong/pdk/log.lua
flrgh: Okay gotcha. Hmm.

Besides being a source from which the html/markdown docs are generated, these annotations (ldoc/luadoc) also serve as an inline reference for developers when reading the code directly, so I'd personally rather not have this escaping done in the source code. It may also break some text editor/IDE/language server implementations that parse these annotations in order to provide type hints and code-completion.

Can we perform this escaping during doc generation instead? I haven't really dug into our ldoc templates all that much, but I believe it should be possible with a strategic placement of something like `string.gsub(parameter_type, "|", "\|")`

https://github.com/Kong/kong/blob/master/autodoc/pdk/ldoc/ldoc.ltp

kong/pdk/response.lua
flrgh: Ah okay, gotcha. Yeah, I realize it's less convenient in the doc-editing context, but I do feel like we want to maintain readability in the source.

kong/pdk/log.lua
lena-larionova: Definitely possible, let me give it a shot.

lena-larionova: Applied all of your other feedback but struggling to figure out how/where exactly to do this. Any ideas?

flrgh: Oop, didn't have time to circle back to this today, but I'll try to take a look soon (also pinging the team in case someone more ldoc-savvy feels like chiming in)

kikito: Interesting. It seems preventing that was the idea behind this bit of template code, but it seems that is not enough:

```
>               if type ~= '' then
>                 type = "`" .. type .. "`:"
>               end
```

Let me investigate (having a link to a place where the generated code breaks the table is very helpful for that)

kikito: I was incorrect, that piece of code was for the __return__ values, and it was working as indended.

Our __parameters__ don't have back-ticks around their types. I have added a change on this PR that should fix that: it will surround all the parameter types with back-ticks, unless they already have them.

kikito (approved):

kong/pdk/log.lua
lena-larionova: Just ran a quick test to generate these docs, everything looks good. Thanks for the help! 🙏

kong/pdk/ctx.lua
```diff
@@ -52,33 +52,35 @@ local _CTX_CORE_KEY = {}
 
 
 ---
--- A table that has the lifetime of the current request - Unlike
+-- A table that contains the entire lifetime of the current request. Unlike
```
flrgh, line 55: Unfortunately we're still deviating from the actual meaning quite a bit with this iteration. What do you think of something like this?

```suggestion
-- A table that has the same lifetime as the current request. Unlike
```

It's not the best sentence in the world, but to the target audience of these docs I believe it conveys its meaning pretty well. I think it's also maybe a tad more readable than the original, but I'd also be fine with reverting to the original line and just adding the period after `request.`

Ditto for the `kogn.ctx.shared` doc block above on line 17.

lena-larionova: Fixed in https://github.com/Kong/kong/pull/8287/commits/2d57cad679108453d56433bf26b924974ff15f05.

flrgh (approved):
:+1:

Issue #7710, opened 2021-08-13
gRPC reflection in the grpc-gateway plugin


brian-scb:
### Enable ability to use gRPC reflection in the grpc-gateway plugin

Instead of providing a path to proto file, it would much easier if the plugin were able to take advantage of the [reflection protocol](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md) to obtain the proto file

other tools capable of using reflection: [grpc-cli](https://github.com/grpc/grpc/blob/master/doc/command_line_tool.md), [Evans](https://github.com/ktr0731/evans)

javierguerragiraldez:
There have been a few proposals about how to avoid having to deploy the proto file with Kong, especially in the context of container images.  Options include:
1. specify a URL instead of a file path.  Kong would download it at startup.
2. include the whole content of the .proto file in the configuration.  Might be too heavy on declarative configurations, but workable if injected via the admin API.
3. this one: use gRPC reflection to pull the definitions from the service.

The first one (pull from an URL) is doable, but has little advantage, since it's hardly easier than making the file accessible, either by adding to a custom image, or by mounting a read-only volume with the required file(s).

Including in the database was proposed at first, but quickly dismissed when seeing not only the size of some .proto specs, but also the problem of adding imported definitions.  There was some possibility of precompiling into binary DescriptorProto bundles, but again, it results in many extra steps for little usability improvement.

gRPC reflection seems very tempting, especially since it's so easy to add on Golang gRPC services.  But the design of Nginx makes it surprisingly hard to initiate an HTTP/2 request that is not a redirection of the main request.  To have a Lua-initiated gRPC call would require an HTTP/2 client Lua library, which are still not stable enough.

Finally, remember that the .proto specification returned by a server via reflection describes the .proto file as it was when the server was compiled (or rather, when `protoc` compiler generated the gRPC/protoBuf stub code).  If the RESTish interface is added after that, the annotations required by this plugin to perform the transformation would not be there.  That means this option would probably help only for in-house services and not to expose third-party services.

In all, I don't want to discard any of these options, but so far all of them seem to provide limited benefit.

brian-scb:
Forgive me for spitballing here, I'm not super familiar with nginx, but could it be done like a write-though cache. Grab the protofile the first hit, cache it for subsequent, expire cache eventually?

StarlightIbuki:
Dear contributor,
[...]

PR #8297, opened 2022-01-15
Fix/luaproto encode hooks

git-torrent:
### Summary

Transcoding between GRPC and JSON [requires some types to be handled exceptionally](https://developers.google.com/protocol-buffers/docs/proto3#json).

Hooks provided by [lua-protobuf](https://github.com/starwing/lua-protobuf) are insufficient as they:
1. cannot hook native types (required for `bytes`)
2. are not executed recursively (required for `Any`)

Further, gateway must read custom JSON names from proto definitions. Conversion must also allow to keep camel_case names, enum names (instead of numbers) and to emit also JSON with default values.

### Full changelog

* added file `plugins/grpc-gateway/protojson.lua` handling proto->JSON formatting
* updated rockspecs to include this new file

In `tools/grpc.lua`:
* refactored to be more readable
* `add_path` and `parse_file` methods are now exposed instead of `for_each_method`
* extra hook on every field of every message included (to extract json_name option value)
* path( "spec/fixtures/grpc" ) excluded from internally added paths as they are only intended for testing and might contain files conflicting with user code. Mixing testing code with production code is bad practice anyway.

in `plugins/grpc-gateway/deco.lua`
* removed code from ancient times where `include_imports` directive was `false`.
* adapted to new `tools/grpc.lua`
* decoding/encoding of messages is now done by using `protojson.lua`

in `plugins/grpc-web`
* adapted to new `tools/grpc.lua`
* no functional changes have been made
* schema extended to allow also additional files (useful for `google.protobuf.Any` conversions)
* schema allows to add also additional import paths (useful for libraries)

Also fixed bug when shorter binding path matched longer one (e.g. `/test` matched also `/test/something/else`).

### Issue reference

Fix #_7469_

KAG-798(internal ticket)


git-torrent:
I am sorry for not providing tests, but my setup is only AKS cluster with our app. I can only patch files and watch the output in the console. Not much for coding/debugging.

Please feel free to update the code to be more efficient or more idiomatic.

What is missing:
* snake_case is still not converted to lowerCamelCase - I considered using __newindex metamethod, but this would not handle custom Json names so I probably stuck to current recursive traversal
* default values are not patched - current algorithm traverses metatable and not descriptor so it does not know about default values at all
* handling of missing types for `any` type conversion. I had to use explicit `p:loadfile()` call what cannot be generalized. I considered using `p.unknown_type` hook, but script does not know the file name for the type. So maybe some type->file mapping?

- [ ] if there is a missing type, what possible mechanisms can I use to load them on the fly? Or up front? Maybe in configuration, `schema.lua`?

StarlightIbuki:
We appreciate your contribution. This is a large PR and we need time to review it.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,166 +65,181 @@ local valid_method = {
...
+  if not cfg.proto then
+    return nil, "Transcoding requests require a .proto file defining the service"
   end
 
-  local endpoint, vars = rpc_transcode(method, path, protofile)
+  local endpoint, vars = rpc_transcode( method, path, cfg )
 
   if not endpoint then
-    return nil, "failed to transcode .proto file " .. vars
+    return nil, "Failed to transcode .proto file " .. vars
```
StarlightIbuki, line 224: Usually, the first character in error messages should not be capitalized. Ref: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,166 +65,181 @@ local valid_method = {
...
-function deco.new(method, path, protofile)
-  if not protofile then
-    return nil, "transcoding requests require a .proto file defining the service"
+--- Constructor
+-- @param method
+-- @param path
+-- @param cfg
+function deco.new( method, path, cfg )
+  if not cfg.proto then
+    return nil, "Transcoding requests require a .proto file defining the service"
```
StarlightIbuki, line 218: Same here.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -295,5 +355,4 @@ function deco:get_raw_downstream_body()
   return self.downstream_body
 end
 
-
-return deco
+return deco
```
StarlightIbuki, line 358: Not introduced by your PR but we need a new line at the end of the file.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -220,11 +249,39 @@ function deco:upstream(body)
...
+  local payload = setmetatable( self.template_payload, {
+  --[[
+    // Set value at dot-separated path. For example: `GET /v1/messages/123456?revision=2&sub.subfield=foo`
+    // translates into `payload = { sub = { subfield = "foo" }}`
+    //
+    // This is to comply with [spec](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L113),
+    // where also fields of messages can be passed in query string.
+    //
+  ]]--
+  __newindex = function (t, k, v)
```
StarlightIbuki, line 262: Consider moving this function and the metatable out of the function body. Instead of creating the metatable every time, we should reuse them.

git-torrent:
> We appreciate your contribution. This is a large PR and we need time to review it.

Thank you.

All the tests are working except `spec/02-integration/07-sdk/03-cluster_spec.lua` which fails for both, `cassandra` and `postgres`. I hope it is not caused by my changes (plugin is tested daily in our environment). I have no idea how to fix these.

kong/plugins/grpc-gateway/handler.lua
```diff
@@ -50,13 +48,14 @@ function grpc_gateway:access(conf)
 
   kong.ctx.plugin.dec = dec
 
-  kong_service_request_set_header("Content-Type", "application/grpc")
-  kong_service_request_set_header("TE", "trailers")
   local body, err = dec:upstream(kong_request_get_raw_body())
   if err then
     kong.log.err(err)
-    return kong_response_exit(400, err)
+    return kong_response_exit(400, err, { ["Content-Type"] = "text/html" })
```
StarlightIbuki, line 54: Good catch. The error response is not of the grpc MIME type.

We usually do not manually set CT for the error response.

kong/plugins/grpc-gateway/schema.lua
```diff
@@ -10,8 +10,39 @@ return {
         {
           proto = {
             type = "string",
+            required = true,
+            default = nil,
+          },
+        },
+        {
+          additional_protos = {
```
StarlightIbuki, line 18: New fields require discussions.

spec/01-unit/22-grpc-utils_spec.lua
```diff
@@ -1,13 +1,14 @@
 local grpc_tools = require "kong.tools.grpc"
+local grpc = grpc_tools.new()
+      grpc:add_path( "spec/fixtures/grpc" )
```
StarlightIbuki, line 3:
```suggestion

grpc:add_path( "spec/fixtures/grpc" )
```

kong/tools/grpc.lua
```diff
@@ -1,152 +1,114 @@
...
-      for _, mthd in ipairs(srvc.method or {}) do
-        f(parsed, srvc, mthd)
-      end
+--- Traverse loaded file and call hooks if possible
+-- File is processed recursively according to `protoc.include_import` settings
+-- Hooks are optional
+-- @param filename Filename to parse
+-- @param method_hook Function to be called for each method
+-- @param field_hook Function to be called for each field (also in nested messages)
+function _M:parse_file( filename, method_hook, field_hook )
```
StarlightIbuki, line 76: This function should be more like an internal function, used to implement more simple functions. Thus this function should have a name like `traversal_proto`, and we use this function to implement `parse_file`, which does not require any callbacks.

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
```diff
@@ -5,12 +5,15 @@ for _, strategy in helpers.each_strategy() do
 
   describe("gRPC-Gateway [#" .. strategy .. "]", function()
     local proxy_client
-
+    local service1
+    local route1
+    local plugin1
+    local bp
```
StarlightIbuki, line 11: `bp` is not used in later tests so it should not be exposed, but rather local to `lazy_setup`.

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
```diff
@@ -153,42 +160,1151 @@ for _, strategy in helpers.each_strategy() do
...
+      local int64_negative = -2305843009213693951 -- for some reason 2^62 is the maximum negative value
+
+      local H = function( string )
+        return "hello " .. ( string or "" )
+      end
+
+      lazy_setup(function()
+        print ("Updating grpc-gateway plugin (id = " .. plugin1.id .. ") settings: use_proto_names=false; enum_as_name=true; emit_defaults=false")
+
+        assert(bp.plugins:update( { id = plugin1.id }, {
```
StarlightIbuki, line 178: We should not assume the ordering of the tests, and tests should not depend on each other, otherwise, the test may go flaky.
Thus we should not modify an existing plugin to run tests.

We could add a second route, and attach to it a `grpc-gateway` plugin with different configs. This way we don't need a `lazy_setup` (and a restart) here.

StarlightIbuki:
> All the tests are working except `spec/02-integration/07-sdk/03-cluster_spec.lua` which fails for both, `cassandra` and `postgres`. I hope it is not caused by my changes (plugin is tested daily in our environment). I have no idea how to fix these.

I'm trying to investigate the issue.

StarlightIbuki:
> > All the tests are working except `spec/02-integration/07-sdk/03-cluster_spec.lua` which fails for both, `cassandra` and `postgres`. I hope it is not caused by my changes (plugin is tested daily in our environment). I have no idea how to fix these.
> 
> I'm trying to investigate the issue.

The failing test runs slow in my env and passed. Guess we need to have a longer timeout.

kong/plugins/grpc-gateway/protojson.lua
```diff
@@ -0,0 +1,600 @@
+local protojson = {
```
StarlightIbuki, line 1: This file should be in `kong/tools/`.
Maybe we should make grpc a folder, and rename `grpc.lua` to `init.lua`, and put both files into the folder.

kong/plugins/grpc-gateway/deco.lua
git-torrent: https://github.com/Kong/kong/pull/8297/commits/b473efb330ab6e30083923d4bc3b804773320e06

kong/plugins/grpc-gateway/deco.lua
git-torrent: https://github.com/Kong/kong/pull/8297/commits/b473efb330ab6e30083923d4bc3b804773320e06

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -14,10 +15,23 @@ local re_match = ngx.re.match
...
 deco.__index = deco
 
+local cache = {}
+
+--- Sort endpoints by specificity (longer regex tend to be more specific).
+-- The similar rule applies for [Kong ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches).
+-- @param endpoints Endpoints to sort
+local function sort_endpoints( endpoints )
+  for method in pairs( endpoints ) do
+    table.sort( endpoints[method], function (a, b)
```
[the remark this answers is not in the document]
git-torrent: https://github.com/Kong/kong/pull/8297/commits/aa52fc1f70237175564a493257ae34caa4a3a580

kong/plugins/grpc-gateway/deco.lua
git-torrent: https://github.com/Kong/kong/pull/8297/commits/650b43abfff8374a638dc8ae516da93c0fb964ca

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,166 +65,181 @@ local valid_method = {
...
+  
+      info.json_names[ full_name ] = field.json_name
+    end
+  end
+  
+  local grpc = grpc_tools.new()
+
+  grpc:parse_file( filename, load_bindings_info, load_json_names_info )
+  
+  for _, fn in ipairs( cfg.additional_protos or {} ) do
```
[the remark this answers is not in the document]
git-torrent: https://github.com/Kong/kong/pull/8297/commits/86cc9a12f90a2b1c2ed18637a57af50e10911186

kong/plugins/grpc-gateway/deco.lua
git-torrent: https://github.com/Kong/kong/pull/8297/commits/d5c17bd12cd99db44baef7e59a8871d3013d81e0

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -253,21 +310,24 @@ function deco:upstream(body)
     ]]--
     -- TODO primitive type checking
     local args, err = ngx.req.get_uri_args()
-    if not err then
+
+    if err then
+      return nil, "Invalid URI arguments"
+    else
```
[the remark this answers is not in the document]
git-torrent: https://github.com/Kong/kong/pull/8297/commits/ec9ae87b5f3f8fd03b2155af09980022c6f5f19f

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -220,11 +249,39 @@ function deco:upstream(body)
...
+  --[[
+    // Set value at dot-separated path. For example: `GET /v1/messages/123456?revision=2&sub.subfield=foo`
+    // translates into `payload = { sub = { subfield = "foo" }}`
+    //
+    // This is to comply with [spec](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L113),
+    // where also fields of messages can be passed in query string.
+    //
+  ]]--
+  __newindex = function (t, k, v)
+      local tab = t -- set up pointer to table root
```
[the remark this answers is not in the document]
git-torrent: https://github.com/Kong/kong/pull/8297/commits/d5c17bd12cd99db44baef7e59a8871d3013d81e0

kong/plugins/grpc-gateway/handler.lua
git-torrent: Can it stay with this CT?

spec/01-unit/22-grpc-utils_spec.lua
git-torrent: https://github.com/Kong/kong/pull/8297/commits/0c5475ae3c4433248308a08b6bf79b0a9bd73376

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,166 +65,181 @@ local valid_method = {
...
-  local match_group_idx = 1
-  local path_regex, _, err = re_gsub("^" .. path .. "$", options_path_regex, function(m)
+--- Parse path in http.options
+-- @param request path 
+-- @return Regular expression to extract variables from request path
+-- @return Variables found in request path
+-- @return Error string
+local function parse_options_path( path )
+  local path_vars = {}
+
```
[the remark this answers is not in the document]
git-torrent: I guess this requires deeper knowledge of lua than mine. That index is part of table internal state, why does it help to have it externally?

kong/plugins/grpc-gateway/schema.lua
git-torrent: What we need is to cover `google.proto.any` to match https://developers.google.com/protocol-buffers/docs/proto3#json 

It is not possible without knowing descriptors in advance. I understood the conclusion from related discussions
- https://github.com/Kong/kong/issues/7710
- https://github.com/Kong/kong/issues/8725

is: _support for dynamic reflection is expensive_.

This extra field does it statically while avoiding digression of `protojson` from its sister implementations:
- https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson
- https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/json-formatter
- https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.util.json_util

I did not figure any better way to support `any` type. I would be happy to support it in any better "kong-friendly" way.

kong/plugins/opentelemetry/proto.lua
```diff
@@ -1,12 +1,12 @@
-local grpc = require "kong.tools.grpc"
+local grpc_tools = require "kong.tools.grpc"
```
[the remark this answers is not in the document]
git-torrent: https://github.com/Kong/kong/pull/8297/commits/0c5475ae3c4433248308a08b6bf79b0a9bd73376

kong/tools/grpc.lua
git-torrent: What about renaming it to `traverse_proto_file` (to be consistent with `open_proto_file` later)?

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
git-torrent: Sorry for boilerplate code, I am new to kong unit-tests. What I want to do:

There are three parameters of `protojson` and optimally, the tests should cover all of their combinations. There are 8 combinations of params in total (all of them are booleans).

How to do that?

kong/plugins/grpc-gateway/protojson.lua
git-torrent: https://github.com/Kong/kong/pull/8297/commits/c4b8f59e86cf5718c47737c0698576fabd5b1fe4

kong/tools/grpc.lua
git-torrent: https://github.com/Kong/kong/pull/8297/commits/c921f9a455824a326589bb3dbd44cbdaf81aa366

kong/plugins/grpc-gateway/deco.lua
StarlightIbuki: I mean a newline after `return deco`.
```suggestion
return deco

```

kong/tools/grpc.lua
StarlightIbuki: I think that should do. It's more of a problem with the original code.

kong/plugins/grpc-gateway/protojson.lua
StarlightIbuki: > What about the file protobuf.lua? It looks like representation of https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto but I am not sure

Looks good to me.

git-torrent: It was more about placing all proto-related stuff into one folder. `protobuf.lua` looks like part of {`grpc.lua`, `protojson.lua`} family. Nevertheless, I would personally prefer flat structure, as these files have no mutual dependencies.

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
StarlightIbuki: We could write a for loop to iterator over all combinations (like `std::next_permutation` for C++) and put all tests (describe blocks) in the loop body. The other tests should be independent of the setting combinations so they are not conditional, and we also write 3 parts of option-dependent tests, each part checking if the corresponding option works correctly.

StarlightIbuki:
Please rebase the PR. We are dropping the wRPC protocol, which should bring more freedom to how we support gRPC.

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
git-torrent: I understand your suggestion as:
image
- we have set of unit_tests whose configuration change for every permutation
- we have set of results for Cartesian product of `permutation` x `#unit_tests`
- we have way to map results to test with respect to permutation

What I do not understand how can be plugin settings changed for every permutation. I guess you want to avoid

```
        assert(bp.plugins:update( { id = plugin1.id }, {
          config = {
            proto = "./spec/fixtures/grpc/targetservice.proto",
            use_proto_names = false,
            enum_as_name = true,
            emit_defaults = false,
          },
        } ))
```

Is there another way to change the settings?

kong/plugins/grpc-gateway/deco.lua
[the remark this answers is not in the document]
StarlightIbuki: This is a historic way to optimize. Back then Luajit could not jit `ipairs`, and `#arr` is much slower than maintaining table length on our own.
It seems the new version of Openresty/luajit2 supports jitting the `ipairs`. But I tend to leave the code unchanged.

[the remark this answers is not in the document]
StarlightIbuki: And `table.insert` is the same as to iterate over.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,165 +67,207 @@ local valid_method = {
...
         http_method = http_method:lower()
-        if valid_method[http_method] then
-          local preg, grp, err = parse_options_path(http_path)
+
+        if is_valid_method[ http_method ] then
+          local regex, varnames, err = parse_options_path( http_path )
+
           if err then
-            ngx.log(ngx.ERR, "error ", err, "parsing options path ", http_path)
+            ngx.log( ngx.ERR, "error: ", err, " parsing options path: ", http_path )
```
StarlightIbuki, line 130: Usually, we don't add space after and before parentheses. I'm not commenting on all places.

```suggestion
            ngx.log(ngx.ERR, "error: ", err, " parsing options path: ", http_path)
```

StarlightIbuki: So for the brackets.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,165 +67,207 @@ local valid_method = {
...
   end
-  for _, endpoint in ipairs(info) do
-    local m, err = re_match(path, endpoint.regex, "jo")
+
+  for _, endpoint in ipairs( endpoints ) do
+    local m, err = re_match( path, endpoint.regex, "jo" )
+
     if err then
-      return nil, ("Cannot match path %q"):format(err)
+      return nil, ("cannot match path %q"):format(err)
```
StarlightIbuki, line 200: Yes, we should not capitalize the first characters in error messages.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,165 +67,207 @@ local valid_method = {
...
+  return nil, ("unknown path %q"):format(path)
+end
 
-function deco.new(method, path, protofile)
-  if not protofile then
+--- Constructor
+-- @param method
+-- @param path
+-- @param cfg
+function deco.new( method, path, cfg )
```
StarlightIbuki, line 220: I guess protofiles are not necessarily cfgs?

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,165 +67,207 @@ local valid_method = {
...
-  // Set value `v` at `path` in table `t`
-  // Path contains value address in dot-syntax. For example:
-  // `path="a.b.c"` would lead to `t[a][b][c] = v`.
-]]
-local function add_to_table( t, path, v, typ )
-  local tab = t -- set up pointer to table root
-  local msg_typ = typ;
-  for m in re_gmatch( path , "([^.]+)(\\.)?") do
-    local key, dot = m[1], m[2]
-    msg_typ = get_field_type(msg_typ, key)
```
StarlightIbuki, line 196: Are we removing the check? Is this a valid change?
(Could it be possible for a gRPC endpoint to receive args irrelevant to the gRPC call?

kong/plugins/grpc-gateway/handler.lua
```diff
@@ -24,11 +23,10 @@ local grpc_gateway = {
   VERSION = kong_meta.version,
 }
 
-
 local CORS_HEADERS = {
   ["Content-Type"] = "application/json",
   ["Access-Control-Allow-Origin"] = "*",
-  ["Access-Control-Allow-Methods"] = "GET,POST,PATCH,DELETE",
+  ["Access-Control-Allow-Methods"] = "GET,POST,PUT,PATCH,DELETE",
```
StarlightIbuki, line 29: This requires an entry in changelog (as a feature, we are now supporting PUT method, or as a bug fix, we corrected the header).

kong/plugins/grpc-gateway/handler.lua
StarlightIbuki: It should be OK I guess.

kong/plugins/grpc-gateway/schema.lua
StarlightIbuki: This also requires an entry in the changelog.

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
StarlightIbuki: Maybe we could create `#permutation` x `#unit_tests` of routes (id named with them).

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
```diff
@@ -153,42 +160,1151 @@ for _, strategy in helpers.each_strategy() do
...
-          tail = {count = 0, endings = "tip" },
+          tail = { count = 0, endings = "tip" },
         }
       })
       assert.equal(200, res.status)
       local body = assert(res:read_body())
       assert.same({
         name = "lizard",
-        hands = { count = 0, endings = "fingers" },
+        hands = { endings = "fingers" }, -- count = 0 is default value, skipped from result
```
StarlightIbuki, line 1307: We should not change this behavior. Maybe we should make the `default handling` setting by default to emit the fields.

kong/tools/protojson.lua
```diff
@@ -0,0 +1,600 @@
...
+-- @param options.emit_defaults_ Emit fields with default values
+-- @param options.json_names_ Table containing json_name options of parsed fields
+function protojson:configure( options )
+  use_proto_names = options.use_proto_names or false
+  enum_as_name = options.enum_as_name or false
+  emit_defaults = options.emit_defaults or false
+  json_names = options.json_names or {}
+end
+
+return protojson
```
StarlightIbuki, line 600: We need a new line at EoF.

```suggestion
return protojson

```

kong/tools/protojson.lua
```diff
@@ -0,0 +1,600 @@
...
+-- @param options.use_proto_names_ Emit proto field name instead of lowerCamelCase name. Read both.
+-- @param options.enum_as_name_ Emit enum values as integers instead of strings. Read both.
+-- @param options.emit_defaults_ Emit fields with default values
+-- @param options.json_names_ Table containing json_name options of parsed fields
+function protojson:configure( options )
+  use_proto_names = options.use_proto_names or false
+  enum_as_name = options.enum_as_name or false
+  emit_defaults = options.emit_defaults or false
+  json_names = options.json_names or {}
+end
```
StarlightIbuki, line 598: There could be multiple users of this module, and each of them may use different configurations.
We need to make those configurations per instance and make every user possess 1 instance.

For example, we may have multiple `grpc-gateway` plugins configured with different values of `emit_defaults` and for different requests handled by the same worker, they share the same state.

kong/plugins/grpc-gateway/schema.lua
StarlightIbuki: Notice that adding fields to `schema.lua` means we also need to add a field in `kong/clustering/compat/removed_fields.lua`.
If you're not sure how to do it, please let me know and I'll do it for you.

kong/tools/protojson.lua
StarlightIbuki: I understand that we have a limitation from `protobuf` library. If we cannot get around of that, please consider make this a global field in `kong.conf` instead of plugin config.

StarlightIbuki:
Sorry. This PR takes really long to review and new changes seem to collide with it. Could you solve conflicts?

git-torrent:
> 

There are two major issues that stand in our way:
1. my lack of time (every conflict resolution takes me lot of time as I need to recall what to do and where, my daytime job has nothing to do with Lua)
2. my inability to make unit-tests better than they are right now in reasonable time

I can resolve conflicts this week (maybe today if lucky), but the unit-tests require substantially more time to learn and master.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -51,165 +67,207 @@ local valid_method = {
...
 
-  local grpc_tools_instance = grpc_tools.new()
-  grpc_tools_instance:each_method(fname, function(parsed, srvc, mthd)
-    local options_bindings =  {
-      safe_access(mthd, "options", "options", "google.api.http"),
-      safe_access(mthd, "options", "options", "google.api.http", "additional_bindings")
+  local load_bindings_info = function( file, service, method )
+    local bindings =  {
+      safe_access( method, "options", "options", "google.api.http"),
+      safe_access( method, "options", "options", "google.api.http", "additional_bindings")
```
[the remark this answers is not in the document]
git-torrent: fixed: https://github.com/Kong/kong/pull/8297/commits/34954ba934fb1555362c82bb2b11455edffc4d07

kong/tools/protojson.lua
git-torrent: fixed: https://github.com/Kong/kong/pull/8297/commits/07c78f92ce6ed22707b67882f97adbeeb0265c39

kong/plugins/grpc-gateway/schema.lua
git-torrent: is this right? https://github.com/Kong/kong/pull/8297/commits/36ef019e6ff1cc5de9fe0c5ce0cd809f299d2fff

kong/plugins/grpc-gateway/handler.lua
git-torrent: Is this one resolved?

kong/plugins/grpc-gateway/handler.lua
git-torrent: May this be? https://github.com/Kong/kong/pull/8297/commits/20c2d67780ea36931edadb43aa195a6b1cc90a87

kong/plugins/grpc-gateway/deco.lua
git-torrent: The way it is done now, [only fields from schema are included](https://github.com/Kong/kong/blob/20c2d67780ea36931edadb43aa195a6b1cc90a87/kong/tools/protojson.lua#L410) so such a field will not make it into proto.

Officially this behavior [should be controlled](https://github.com/protocolbuffers/protobuf-go/blob/v1.30.0/encoding/protojson/decode.go#L41), but not now, not here :)

kong/plugins/grpc-gateway/deco.lua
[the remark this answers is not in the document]
git-torrent: fixed: https://github.com/Kong/kong/pull/8297/commits/7ccdac4ea693f425c5d1c87c53368fe710e7569c

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
git-torrent: fixed in https://github.com/Kong/kong/pull/8297/commits/891d1f010adc766c92d0ec438be157afc5330445

spec/03-plugins/28-grpc-gateway/01-proxy_spec.lua
git-torrent: something like this? https://github.com/Kong/kong/pull/8297/commits/e51c132857620138e54186d7466f8f6e1585cec3

StarlightIbuki:
@git-torrent As this is a very large PR and hard to maintain, and we see no hope for it to be merged soon, I'm thinking about taking over this PR myself if you don't mind. Of course, you will be given credit. What do you think?
My first step would be to break this PR into 3 or more PRs so it's easier to review.

git-torrent:
> @git-torrent As this is a very large PR and hard to maintain, and we see no hope for it to be merged soon, I'm thinking about taking over this PR myself if you don't mind. Of course, you will be given credit. What do you think? My first step would be to break this PR into 3 or more PRs so it's easier to review.

I appreciate your offer. With this PR having more badges (labels) than old military officer, I think your care would help. 

Just one thing please: Names of the variables/functions are optimized for me from the future. Of course, feel free to optimize the code, but until merged, I will be still in charge of keeping our (company) fleet of Kongs up to date, which require this gRPC<->REST operating. Keeping the variable names (when possible) would make this upgrade&patch process easier. 

If I can be of any further help, please write me.

Thank you :)

hbagdi:

git-torrent:
@hbagdi > Closing this due to lack of activity. Please re-open if needed.

How can you close XXL PR where lack of inactivity is clearly an effect of being XXL. You have got bunch of functionality for free. I do not expect to be paid, I expect Kong to pay little interest in community's work. This way I would use Traefik next time :(

StarlightIbuki:
@git-torrent Sorry for the confusion. The closing is done in a fixed process without much look at the PR and has not much intention on it.
I'm reopening this PR but sadly I'm occupied by other issues and cannot find time to work on this recently.
I'll convert this into a draft PR so it won't show up in the list and get closed mistakenly.

git-torrent:
> @git-torrent Sorry for the confusion. The closing is done in a process without looking at what the PR is doing and has not much intention on it. I'm reopening this PR but sadly I'm occupied by other issues and cannot find time to work on this recently. Please understand that we need to prioritize our work for different urgency. I'll convert this into a draft PR so it won't show up in the list and get closed mistakenly.

@StarlightIbuki I am afraid I need your help with unit tests. I have to rebase this branch from time to time to prepare patch for our deployments. It is useful to have unit-tests done, but I cannot do that. Can you please point me to some exemplary solution of what you wanted? I tried to look at existing test-cases, but mine are still failing. Thank you

StarlightIbuki:
@git-torrent Hi. Glad to know you're still interested in maintaining the RP. When I said taking over, I mean I will help to rework the PR, write proper tests for it and etc. It's fine to not fix the test or rebase to keep maintaining the PR.
The good news is that I have planned this PR's work. I will hand on it after the 3.4 release if no unexpected urgent work gets in the way.
If you still want to write the unit test, please tell me the purpose of the test (which part is tested? What do we want to test?) and the problem you're facing.

StarlightIbuki:
@git-torrent Hi. I'm gradually starting the rework of this PR.
First thing is to make a full list of the improvement this PR wants to do.
Please correct me/add to the list if I made a mistake or miss something.

1. Wrap the `pb` library so we can isolate the hack to it;
2. Better handling of pb messages:
  1. Make the empty table/array distinctable. This is already done: #10790;
  2. More careful handling of numbers, say, int64, NaN, etc, and boolean/byte types;
  3. Name conversions (camel/snake case);
  4. Support of `.google.protobuf.*` 
3. Allow different configs to coexist for different instances of isolated pb library, including `decode_default_message` `use_proto_names`, `local enum_as_name`, `emit_defaults`
4. Make decoding/encoding behavior configurable for the grpc-gateway plugin;
5. Endpoint sort;
6. Fix path searching;
7. `additional_protos`;
8. PUT method;

git-torrent:
> @git-torrent Hi. I'm gradually starting the rework of this PR. First thing is to make a full list of the improvement this PR wants to do. Please correct me/add to the list if I made a mistake or miss something.
> 
> 1. Wrap the `pb` library so we can isolate the hack to it;
> 2. Better handling of pb messages:
> 3. Make the empty table/array distinctable. This is already done: [ fix(*): gRPC and cJSON's behavior to empty array #10790](https://github.com/Kong/kong/pull/10790);
> 4. More careful handling of numbers, say, int64, NaN, etc, and boolean/byte types;
> 5. Name conversions (camel/snake case);
> 6. Support of `.google.protobuf.*`
> 7. Allow different configs to coexist for different instances of isolated pb library, including `decode_default_message` `use_proto_names`, `local enum_as_name`, `emit_defaults`
> 8. Make decoding/encoding behavior configurable for the grpc-gateway plugin;
> 9. Endpoint sort;
> 10. Fix path searching;
> 11. `additional_protos`;
> 12. PUT method;

Thank you. I am not sure if I can map your steps to this PR completely. I miss a support for `json_name` option in proto files to handle overrides.

Also, I am not sure if you want to preserve proto<->json transcoding as a standalone code. It is a standalone library in [C++](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/json/json.h), [C#](https://github.com/protocolbuffers/protobuf/blob/main/csharp/src/Google.Protobuf/JsonFormatter.cs) and [Go](https://github.com/protocolbuffers/protobuf-go/tree/master/encoding/protojson), so I guess it is better for maintenance. For `pb.(encode_)hook` to work properly, it must be able to handle scalar types (`bytes` in particular) and also be recursive (to decode `Any` type). 

I would also like to prepare unit-tests to be sure none of the steps breaks existing (PR) code. I still struggle to run tests in different configurations.

What do you think?

StarlightIbuki:
> I miss a support for `json_name` option in proto files to handle overrides.

Edited.

> Also, I am not sure if you want to preserve proto<->json transcoding as a standalone code. It is a standalone library in [C++](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/json/json.h?rgh-link-date=2023-08-07T22%3A55%3A43Z), [C#](https://github.com/protocolbuffers/protobuf/blob/main/csharp/src/Google.Protobuf/JsonFormatter.cs?rgh-link-date=2023-08-07T22%3A55%3A43Z) and [Go](https://github.com/protocolbuffers/protobuf-go/tree/master/encoding/protojson?rgh-link-date=2023-08-07T22%3A55%3A43Z), so I guess it is better for maintenance. 

Sure. Better maintenance is also a target of this task.

> For `pb.(encode_)hook` to work properly, it must be able to handle scalar types (`bytes` in particular) and also be recursive (to decode `Any` type).

It's included as a part of #4. I'm not very sure what the "recursive mean". Could you elaborate?

> I would also like to prepare unit-tests to be sure none of the steps breaks existing (PR) code. I still struggle to run tests in different configurations.

I will add tests for each of those changes.

git-torrent:
> I'm not very sure what the "recursive mean". Could you elaborate?

Message serialized into `google.protobuf.Any` can contain another `google.protobuf.Any` attribute and so on. To transcode them the hook must treat `Any` as new message instance and start over. Moreover, the hook must know the `@type` in advance. This is handled by `additional_protos` or some other kind of DescriptorPools from [C++](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.h#L2029) or [python](https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/descriptor_pool.py) protobuf implementations.

> And I'm inviting you to review PRs, ...

Gladly, thank you :)

Water-Melon:
> Message serialized into `google.protobuf.Any` can contain another `google.protobuf.Any` attribute and so on. To transcode them the hook must treat `Any` as new message instance and start over. 

Yeah, I tried this proto that given below in Lua and Python.
```
message Phone {                                                                                                                                
   optional string name        = 1;                                                                                                            
   optional int64  phonenumber = 2;                                                                                                            
   optional Person item        = 3;                                                                                                            
}                                                                                                                                              
message Person {                                                                                                                               
   optional string name     = 1;                                                                                                               
   optional int32  age      = 2;                                                                                                               
   optional string address  = 3;                                                                                                               
   repeated Phone  contacts = 4;                                                                                                               
}
```
In python, the reference of a `Person` object can not be assigned to `Phone`'s `item`.
But in Lua, you can assigned the `Person` object (such as named `person`) to `Phone`'s `item`, and it encountered a core dump when I was using `lua-protobuf` to encode message.

So I think may be we shouldn't to build a ref cycle. Type can refer to itself, but data cannot. Maybe we can try to imagine how we can implement the encode function in this situation. Encoding one message object `A`, and it contains another message object `B`, and `B` also need the `A`, encoder will never obtain the complete message data. So the core dump happened because stack is full.

git-torrent:
Hello, how is this going?

StarlightIbuki:
@git-torrent I underestimated the amount of work and am sorry for failing to keep the promise. 3.5 is filled with other features and for personal reasons, I was off for nearly 2 weeks. This improvement cannot be fit in a gap and has to be planned. I'm trying to make this a part of 3.6.

git-torrent:
> @git-torrent I underestimated the amount of work and am sorry for failing to keep the promise. 3.5 is filled with other features and for personal reasons, I was off for nearly 2 weeks. This improvement cannot be fit in a gap and has to be planned. I'm trying to make this a part of 3.6.

Can I be of any help? I would still gladly contribute to unit-tests (if I know how to do them :) )

StarlightIbuki:
> > @git-torrent I underestimated the amount of work and am sorry for failing to keep the promise. 3.5 is filled with other features and for personal reasons, I was off for nearly 2 weeks. This improvement cannot be fit in a gap and has to be planned. I'm trying to make this a part of 3.6.
> 
> Can I be of any help? I would still gladly contribute to unit-tests (if I know how to do them :) )

We want to rework the PR and before the implementation is written we don't know how to unit test them. The best help we could expect from you should be a list of the improvements that we want in detail like [the list we made before](https://github.com/Kong/kong/pull/8297#issuecomment-1665279200). To be honest, I don't fully understand the PR's intention, and the list I made is not very in-depth.

git-torrent:
> We want to rework the PR and before the implementation is written we don't know how to unit test them. The best help we could expect from you should be a list of the improvements that we want in detail like [the list we made before](https://github.com/Kong/kong/pull/8297#issuecomment-1665279200). To be honest, I don't fully understand the PR's intention, and the list I made is not very in-depth.

Google provides automated GRPC-REST translation in [their own API gateway solution](https://github.com/grpc-ecosystem/grpc-gateway). Kong provides its own, but lacks many of the translation features [required by Google](https://protobuf.dev/programming-guides/proto3/#json). The goal of this PR is to bring google-compliant translation into Kong.

For this purpose google offers several libraries ([go](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson), [c++](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.util.json_util/), [C#](https://protobuf.dev/reference/csharp/api-docs/class/google/protobuf/json-formatter), [python](https://stackoverflow.com/questions/19734617/protobuf-to-json-in-python)), but no lua-one. This library is `protojson.lua` in PR. 

The rest is to use protojson library in grpc-gateway plugin.

Does this bring more light into this PR?

StarlightIbuki:
@git-torrent 3.5 is delayed and we have planned so many things for 3.6. I see little hope that we can fit this improvement into the foreseeable future... Sorry again for this message.
I reviewed this fix and have some personal views. The behavior may be easier to do as a separate functionality instead of a part of Kong. It could be a set of functions or a module (to_JSON, from_JSON?) of [lua-protobuf](https://github.com/starwing/lua-protobuf), or an extension to it (not necessarily an official part of the lib).
That way it should also be easier to unit test (and you don't need to follow the Kong's conversion to write the tests).

git-torrent:
> @git-torrent 3.5 is delayed and we have planned so many things for 3.6. I see little hope that we can fit this improvement into the foreseeable future... In fact, we have many other possible improvements in the queue. Sorry again for this message. I reviewed this fix and have some personal views. IMO, your effort is worth a dedicated repository. It may be easier to do as a separate functionality instead of a part of Kong. It could be a set of functions or a module (to_JSON, from_JSON?) of [lua-protobuf](https://github.com/starwing/lua-protobuf), or an extension to it (not necessarily an official part of the lib). That way it should also be easier to unit test (and you don't need to follow the Kong's conversion to write the tests). Then from Kong, the PR should be much smaller and easier to review.

Oh, that could be. Is there any guide to write modules or extension to lua-protobuf? Or some example on the web (google did not help me much :) ). I am not the C guy, so I have to learn that first.

StarlightIbuki:
And in my very personal POV, there's no need to write such a library in C given LuaJIT emits quite efficient binary and cooperates with C types well.
If you still want to PR to [lua-protobuf](https://github.com/starwing/lua-protobuf) directly, you could start with Lua C APIs: https://www.lua.org/manual/5.1/manual.html#lua_Alloc

git-torrent:
> > > @git-torrent 3.5 is delayed and we have planned so many things for 3.6. I see little hope that we can fit this improvement into the foreseeable future... In fact, we have many other possible improvements in the queue. Sorry again for this message. I reviewed this fix and have some personal views. IMO, your effort is worth a dedicated repository. It may be easier to do as a separate functionality instead of a part of Kong. It could be a set of functions or a module (to_JSON, from_JSON?) of [lua-protobuf](https://github.com/starwing/lua-protobuf), or an extension to it (not necessarily an official part of the lib). That way it should also be easier to unit test (and you don't need to follow the Kong's conversion to write the tests). Then from Kong, the PR should be much smaller and easier to review.
> > 
> > 
> > Oh, that could be. Is there any guide to write modules or extension to lua-protobuf? Or some example on the web (google did not help me much :) ). I am not the C guy, so I have to learn that first.
> 
> By module, I mean "Lua module". There's no such mechanism of lua-protobuf to support plugins. To begin with, I guess you could request some features from the library maintainer, like custom handlers for build-in types or even direct fixes to those build-in types' behavior (with an option). You could write your module in Lua to handle the proto file importing, and if the features you requested do not make it into the library, you could wrap over the original library, or make some hooks/monkey patches on it.

Ok, let's write it in Lua. Do you suggest I should agree with author of lua-protobuf to add here https://github.com/starwing/lua-protobuf a `protojson.lua` file that does the job for us (and required `protoc.lua`?). 

Btw, I've already reported a missing support for hooking data-types and it was rejected as too laborious. I would not expect anything more :( (it seems nobody wants to be compatible with google).

StarlightIbuki:
> Do you suggest I should agree with author of lua-protobuf to add here [starwing/lua-protobuf](https://github.com/starwing/lua-protobuf) a `protojson.lua` file that does the job for us (and required `protoc.lua`?).

That would be ideal. If they do not like the idea, you could try make it a library which imports lua-protobuf.

> Btw, I've already reported a missing support for hooking data-types and it was rejected as too laborious. I would not expect anything more :( (it seems nobody wants to be compatible with google).

Lua is not a mainstream user of the protobuf and gRPC and people seem to tolerate those glitches in their use. After all, with some workarounds, they can still use it.

quantverse:
Hi, any chance this PR will solve the issue where `int64` types are transcoded to JSON like floats with lost digits? According to the specs these should be transcoded as strings.

git-torrent:
> Hi, any chance this PR will solve the issue where `int64` types are transcoded to JSON like floats with lost digits? According to the specs these should be transcoded as strings.

yes, this PR solves this by enabling int64_as_string option https://github.com/starwing/lua-protobuf?tab=readme-ov-file#options

quantverse:
Great news! Any chance this PR will be merged soon? Without this fix, the gRPC-gateway plugin is not really much usable for anyone who uses 64-bit integers like identifiers...

I think I will try to rebase this PR against the latest master and test if it works as expected...

quantverse:
FYI I rebased that branch onto the current master and it builds fine, unfortunately, the int64 types don't really transcode well - they are now wrapped as strings but still processed as floats.

`1718291265706240595` becomes `"1.7182912657062e+18"`

but almost there!

quantverse:
Made it working if anyone is interested, as long you don't mind that in JSON the integers are prefixed with `#`:

```
diff --git a/kong/tools/protojson.lua b/kong/tools/protojson.lua
--- a/kong/tools/protojson.lua	(revision 2ea36c87bb42c0a2555f8c9f791baf9654de2c97)
+++ b/kong/tools/protojson.lua	(date 1718294556881)
@@ -243,13 +243,12 @@
   elseif ( f_type == "bytes" ) then
     return ngx.encode_base64( f ) 
   elseif ( is_int[ f_type ] ) then -- parse numeric value
-    if ( type(f) ~= "number" or f ~= f or f ~= math.floor( f ) ) then -- possibly remove
-      error( ("invalid integer value at %s, got: %s"):format(path, f), 0 )
-    end
-
     if is_int64[ f_type ] then
       return tostring( f )
     else
+      if ( type(f) ~= "number" or f ~= f or f ~= math.floor( f ) ) then -- possibly remove
+        error( ("invalid integer value at %s, got: %s"):format(path, f), 0 )
+      end
       return f
     end
   elseif ( is_decimal[ f_type ] ) then
@@ -585,7 +584,7 @@
     pb.option( "enum_as_value" )
   end
 
-  pb.option( "int64_as_number" )
+  pb.option( "int64_as_string" )
 
   return message_to_json( pb.decode( msg_type, msg ), msg_type, path )
 end

```

quantverse:
Unfortunately, there is another problem in the opposite direction (JSON to proto for instance when transcoding a POST body) - 64-bit integers are transcoded wrong without warning!

An example: `1718291265706240595` (or `"1718291265706240595" passed as a string, does not matter) in JSON is transcoded to `1718291265706240512` at the proto side (note the last two digits) - there is apparently some precision loss during the transcoding process...

This is especially dangerous if this is an ID of something because these IDs get silently modified without any warning. Troubleshooting this could be quite a nightmare.

Any idea where to look to fix this one?

git-torrent:
Thank you for your input. I am leaving for a short vacation, let me fix it upon arrival.

quantverse:
Any ETA for the fix?

quantverse:
Related issues/PRs to the problem I have reported above:

https://github.com/mpx/lua-cjson/issues/37
https://github.com/mpx/lua-cjson/pull/36

StarlightIbuki:
A good news is that we are working on https://github.com/Kong/lua-resty-simdjson, which may solve the issues this PR aims to solve

PR #8299, opened 2022-01-17
chore(*) use string.byte to optimize dns client.lua

chronolaw:
### Summary

dns.client.lua use many string.sub to manipulate string,
We can use string.byte to optimize check first/last char,
It is better than string.sub.

### Full changelog

* use string.byte instead of string.sub

### DIFF
--- a/kong/resty/dns/client.lua
+++ b/kong/resty/dns/client.lua
@@ -43,6 +43,10 @@ local table_remove = table.remove
 local table_insert = table.insert
 local table_concat = table.concat
 local string_lower = string.lower
+local string_byte  = string.byte
+
+local DOT   = string_byte(".")
+local COLON = string_byte(":")
 
 local EMPTY = setmetatable({},
   {__newindex = function() error("The 'EMPTY' table is read-only") end})
@@ -646,7 +650,7 @@ local function parseAnswer(qname, qtype, answers, try_list)
 
   -- remove last '.' from FQDNs as the answer does not contain it
   local check_qname do
-    if qname:sub(-1, -1) == "." then
+    if string_byte(qname, -1) == DOT then
       check_qname = qname:sub(1, -2) -- FQDN, drop the last dot
     else
       check_qname = qname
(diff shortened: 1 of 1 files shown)


bungle (approved):

javierguerragiraldez:
i want to note a couple things here:

1. i do like this kind of tweaks, i think string.byte() is a better expression of "i want to check this ASCII char", so readability is improved.

2. but...
  2.1. the escape analysis can statically determine that the `string:sub()` result never goes anywhere, so no new string is allocated/copied/interned
  2.2. the compared strings are 1byte long each, so it's just a byte compare, not a hashed string function call
  2.3. constant propagation would compile `string.byte('x')` into a constant itself, so the upvalue isn't really needed either
  2.4. upvalues aren't locals, they're GC objects by themselves, and using them requires at least one extra reference (still much better than the hash table access for globals, but far more expensive than real locals)

try:
```
luajit -j dump -e "s='abc'; k=0; for _ = 0, 60 do for i=1, #s do if s:sub(i,i)=='A' then k=k+1 end end end"
```
and see in the IR code how there's no new string interning, no string comparison (just a constant compare with +65 (ASCII 'A')), and no call to string.sub (just a check to verify it's still the function it was supposed to be)


TL;DR:  yes, i like the code better like this, i think it's more readable.  but no, it won't be faster

chronolaw:
> i want to note a couple things here:
> 
> 1. i do like this kind of tweaks, i think string.byte() is a better expression of "i want to check this ASCII char", so readability is improved.
> 2. but...
>    2.1. the escape analysis can statically determine that the `string:sub()` result never goes anywhere, so no new string is allocated/copied/interned
>    2.2. the compared strings are 1byte long each, so it's just a byte compare, not a hashed string function call
>    2.3. constant propagation would compile `string.byte('x')` into a constant itself, so the upvalue isn't really needed either
>    2.4. upvalues aren't locals, they're GC objects by themselves, and using them requires at least one extra reference (still much better than the hash table access for globals, but far more expensive than real locals)
> 
> try:
> 
> ```
> luajit -j dump -e "s='abc'; k=0; for _ = 0, 60 do for i=1, #s do if s:sub(i,i)=='A' then k=k+1 end end end"
> ```
> 
> and see in the IR code how there's no new string interning, no string comparison (just a constant compare with +65 (ASCII 'A')), and no call to string.sub (just a check to verify it's still the function it was supposed to be)
> 
> TL;DR: yes, i like the code better like this, i think it's more readable. but no, it won't be faster

Thanks for your comment, I learned a lots.

PR #8303, opened 2022-01-18
chore(*) use get_now_ms() / get_start_time_ms()

chronolaw:
### Summary

There too many repeated number '1000' in init.lua,
maybe move them into one or two local function is a good idea.

### Full changelog

* introduce get_now_ms()
* introduce get_start_time_ms()
* remove 'now() * 1000' and 'start_time() * 1000'

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -344,10 +344,17 @@ local function execute_cache_warmup(kong_config)
   return true
 end
 
+local function get_start_time_ms()
+  return start_time() * 1000
+end
+
+local function get_now_ms()
+  return now() * 1000 -- time is kept in seconds with millisecond resolution.
+end
 
 local function get_updated_now_ms()
   update_time()
-  return now() * 1000 -- time is kept in seconds with millisecond resolution.
+  return get_now_ms()
 end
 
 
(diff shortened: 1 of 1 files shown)


bungle:
@chronolaw, do you know if it has any performance effect, to worse or better? I think originally as it was just `* 1000`, it was just decided to inline it.

chronolaw:
> @chronolaw, do you know if it has any performance effect, to worse or better? I think originally as it was just `* 1000`, it was just decided to inline it (to not have to call any function). This happens on very hot path. Code-wise it looks nicer, but...

Maybe you are right, I have no idea about the effect of these small function.

If it cause performance issue, perhaps keep the original code will be a safe choice.

bungle:
> If it cause performance issue, perhaps keep the original code will be a safe choice.

I wouldn't think it causes a lot of perf issues, and I think those functions would be JIT compiled anyway. Probably not a real-world problem. But if it drops performance, I am not sure if the readability gains are worth to warrant it. But I have no strong feeling about it. My concern may well be premature (I remember that I thought about this before, but decided not to do it).

chronolaw:
It is a small style fix, easy to do for anyone else.

As your opinion, maybe we should consider it in the future.

PR #8307, opened 2022-01-19
fix(dns) check options.search array for dot domain

chronolaw:
### Summary

if `/etc/resolv.conf` has some special value, like `search .`
then in function `search_iter()`, we will get wrong domain `qname..` ,
It is not a valid dns and will cause error.

So we should check the `search` table,
if there is only one char '.', we replace it with '',
then we can get the right dns name 'qname.'

### Full changelog

* check options.search, replace single '.' to ''

### DIFF
--- a/kong/resty/dns/client.lua
+++ b/kong/resty/dns/client.lua
@@ -620,9 +620,9 @@ _M.init = function(options)
   log(DEBUG, PREFIX, "search = ", table_concat(options.search,", "))
 
   -- check if there is special domain like "."
-  for i = 1, #options.search do
+  for i = #options.search, 1, -1 do
     if options.search[i] == "." then
-      options.search[i] = ""
+      table_remove(options.search, i)
     end
   end
 
--- a/kong/resty/dns/client.lua
+++ b/kong/resty/dns/client.lua
@@ -619,6 +619,13 @@ _M.init = function(options)
   options.search = options.search or resolv.search or { resolv.domain }
   log(DEBUG, PREFIX, "search = ", table_concat(options.search,", "))
 
+  -- check if there is special domain like "."
+  for i = 1, #options.search do
+    if options.search[i] == "." then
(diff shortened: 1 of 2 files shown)


mayocream (approved):

kong/resty/dns/client.lua
```diff
@@ -619,6 +619,13 @@ _M.init = function(options)
   options.search = options.search or resolv.search or { resolv.domain }
   log(DEBUG, PREFIX, "search = ", table_concat(options.search,", "))
 
+  -- check if there is special domain like "."
+  for i = 1, #options.search do
+    if options.search[i] == "." then
+      options.search[i] = ""
```
bungle, line 625: What does this do? Should we instead `table.remove` it?

chronolaw: dns.client will init by a options table, which has a `search` field. 
It  comes from `etc/reslov.conf`, someting like:
`search xxx.com xxx.org`
then when we call `toip()` `reslove()`, dns.client will try to concat the qname with `search` table in function `search_iter()`,
to get the proper fqdn, such as `qname.xxx.com` `qname.xxx.org`.

if the `search` field is not a normal domain name but a single `.`,
then the string concat will get the wrong result,
it means `concat(qname, '.', '.')` => `qname..` .

`.` is the root domain name, if we find `.`,
we should only add a `.` suffix to the qname instead of concat,
so the right name will be `qname.`

chronolaw: Also, `qname.` seems to have the same effect of `qname`,
maybe just remove `.` is another choice.

dndx (commented):
This change LGTM. Can we add a simple test case for this in https://github.com/Kong/kong/blob/master/spec/01-unit/21-dns-client/02-client_spec.lua? It will help prevent further rewrites from introducing the same bug again.

bungle (approved):

PR #8315, opened 2022-01-20
fix(datadog) default value for metrics specified twice

bungle:
### Summary

I found out that datadog schema specifies default for metrics in two levels. Once in config field level and second in metrics level. The config field level causes validation errors in schema loading but those errors are ignored by our loader for some reason. Still I think this is a bug and the default should not be there twice.

### DIFF
--- a/kong/plugins/datadog/schema.lua
+++ b/kong/plugins/datadog/schema.lua
@@ -72,7 +72,6 @@ return {
     { protocols = typedefs.protocols },
     { config = {
         type = "record",
-        default = { metrics = DEFAULT_METRICS },
         fields = {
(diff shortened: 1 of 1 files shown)


kong/plugins/datadog/schema.lua
```diff
@@ -72,7 +72,6 @@ return {
     { protocols = typedefs.protocols },
     { config = {
         type = "record",
-        default = { metrics = DEFAULT_METRICS },
```
hbagdi, line 75: cc @kinman who might be affected by the change

kikito (approved):

PR #8337, opened 2022-01-25
Moving constant CLUSTERING_MAX_PAYLOAD to be configurable in kong.conf

andrewgkew:
### Summary

A property that needs to be adjusted when configs become large when in hybrid mode.
At the moment it can not be raised over 4MB so when your config reaches that size it wont get synced to data planes.
This change allows you to change the max value.

### Full changelog

* Moved constant CLUSTERING_MAX_PAYLOAD to be configurable in kong.conf

### Issue reference

Fix #_FTI-2868_
This is the OSS fixed which is also being fixed in EE (https://github.com/Kong/kong-ee/pull/2864) @aboudreault


aboudreault:
@gszr Do you know what is this resource error in CI? Seems unrelated to the PR..

aboudreault:
restarted CI many times without luck.

aboudreault (approved):

PR #8343, opened 2022-01-26
chore(ci) run the labeler only on pull_request_target

locao:
See #8322 for more details.

Adding `pull_request_target` was not enough to fix the permission issue. This PR is a workaround, probably the ideal solution is the suggested in [this comment](https://github.com/actions/labeler/issues/12#issuecomment-1019246202).

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -6,7 +6,7 @@
 # https://github.com/actions/labeler
 
 name: Labeler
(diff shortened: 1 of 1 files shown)


flrgh (approved):
:+1: this looks good to me. Looking at the gh action docs, I _think_ the event sources for `pull_request_target` and `pull_request` are the same (just with different execution semantics), so it should be fine to only use `pull_request_target`.

PR #8345, opened 2022-01-26
feat(datadog) allow adding tags to all Datadog metrics

stpierre:
### Summary

Without this, the only way to add tags to every metric emitted by the Datadog plugin is to completely specify the entire metric list, which is tedious, error-prone, and not forward-compatible. This adds a feature that lets a user specify a list of tags to add to every metric emitted by the Datadog plugin.

### Full changelog

* Allow adding tags to all Datadog metrics emitted, not just per-metric.


hbagdi:
Nice one. Mind renaming "tags" to something more self-describing? "common_tags" perhaps? Totally open to other better suggestions.

stpierre:
I'm really struggling to get the unit tests working locally, even in the `docker-compose` stack described in `.devcontainer`, but once I get that figured out I'll try to get this passing and updated.

stpierre:
I would love to, but the test suite has stubbornly refused my (considerable!) efforts to get it running locally. :sob:

I've tried a couple of different Docker containers and no Docker container at all, but I keep running up against (different) issues. I've read through [various](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#submitting-a-patch) [docs](https://github.com/Kong/kong/blob/master/DEVELOPER.md) and [example](https://github.com/Kong/kong/blob/master/.ci/run_tests.sh) [code](https://github.com/Kong/kong/tree/master/.devcontainer), but no luck. Is there any other documentation I should read? Or is my best bet to hop on Gitter and see if anyone can help?

kong/plugins/datadog/schema.lua
```diff
@@ -80,6 +80,10 @@ return {
           { service_name_tag = { type = "string", default = "name" }, },
           { status_tag = { type = "string", default = "status" }, },
           { consumer_tag = { type = "string", default = "consumer" }, },
+          { common_tags = {
+              type = "array",
+              elements = { type = "string", match = "^.*[^:]$" },
+          }, },
```
mikefero, line 86: This is a breaking change for version compatibility with older data planes; updates will need to be made to the [compatibility layer ](https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua)to handle these changes ahead of time of before a release.

dcarley:
@stpierre Was this closed because of the test failure? Any chance of reviving it?

stpierre:
@dcarley Basically, yeah. I never was able to get the tests running locally, and don't have time to fuss with this. If you want to take it over, please do -- I'd still love to have the feature!

PR #8346, opened 2022-01-26
Update changelog for 2.7.1

aboudreault:


kikito (approved):

kikito (approved):
Thanks Alan, this is enough for 2.7.1 as it is.

We are still missing an "Unreleased" section containing the fixes, features and perf changes that we have on the master release. Having that document up to date would have helped product decide if any fixes needed to be inserted into 2.7.1. I will do that in a separate PR.

PR #8352, opened 2022-01-28
fix(zipkin): jaeger header validation

shettyh:
### Summary

With the latest [spec](https://www.jaegertracing.io/docs/1.29/client-libraries/#tracespan-identity) of `uber-trace-id`, parent_id is not mandatory. Because of this zipkin plugin validation is failing if the parent_id is sent as `0` and it is recreating the header with new `trace_id`. This results in not propagating the downstream context to upstream services.


mayocream:
@shettyh can you fix the failing tests? Try to run `make lint` locally.

mayocream (approved):

kikito:
Hello, I just approved this. Unfortunately it didn't make the cut for the 2.8 release, but should be included on the next one.

PR #8357, opened 2022-01-31
Feat/wrpc

javierguerragiraldez:
wRPC protocol in CP/DP communications

- [x] connect via WebSockets, preserve existing authentication schemes.
- [x] bidirectional RPC, passing data/args/results/errors in both directions.
- [x] some error handling
- request/response/pending expiration
  - [x] drop expired packets on reception
  - [x] drop expired waiters on response
  - [ ] drop expired waiters that never receive response
- [x] encapsulate receive thread within protocol management
- [x] encapsulate send thread within protocol management, only for "server" sockets.
- [x] send periodic PINGs
- [ ] detect and act on missing pings or pongs.
- [x] send config data from CP to every DP on any change.
- [x] handle hash or version of config data, to ignore unneeded config messages.
- transform config data to fit protobuf spec
  - [x] wrap plugins config in Struct
  - [x] allow foreign keys to be passed in expanded form
  - [x] ensure "absent" fields don't get materialized with "default" or "empty" values on decoding.
  - ...
- [x] define inter-worker communication channel
- [x] permit calls from any worker
- [x] pass existing hybrid mode tests

Note: there's still some code duplication between the existing (json) protocol and this.  This is to ensure the feature flag (cluster_protocol=json|wrpc) doesn't activate any new code.  Once the version negotiation feature (#8397 ) is merged in, this will be refactored on top of that and repeated code will be factored away.


kong/tools/channel.lua
```diff
@@ -0,0 +1,158 @@
...
+local RESP_VAL_KEY_PREFIX = "channel_resp_value_"
+
+
+local function waitstep(step, deadline)
+  sleep(step)
+  return min(max(0.001, step * 2), deadline-now(), 0.5)
+end
+
+
+local function add_wait(dict, key, val, exptime, deadline)
```
mikefero, line 21: WDYT about naming these `*_wait` functions to give a better description of what they are doing as it was not immediately obvious until looking through the function?

```suggestion
local function add_key_value_wait(dict, key, val, exptime, deadline)
```

I am not even a fan of this name `add_key_value_wait`; naming is hard.

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,748 @@
...
+    payloads = payloads,
+  })
+  return self.seq
+end
+
+--- little helper to ease grabbing an unspecified number
+--- of values after an `ok` flag
+local function ok_wrapper(ok, ...)
+  return ok, {n = select('#', ...), ...}
+end
```
mikefero, line 435: :open_mouth: This is magic!!!

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,748 @@
...
+wrpc_service.__index = wrpc_service
+
+
+--- a `service` object holds a set of methods defined
+--- in .proto files
+function wrpc.new_service()
+  if not wrpc_proto then
+    local wrpc_protofname = assert(proto_searchpath("wrpc.wrpc"))
+    wrpc_proto = assert(grpc.each_method(wrpc_protofname))
+    --pp("wrpc_proto", wrpc_proto)
```
mikefero, line 219: should this become a DEBUG log message?

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,748 @@
...
+
+
+--- Part of wrpc_peer:call()
+--- If calling the same method with the same args several times,
+--- (to the same or different peers), this method returns the
+--- invariant part, so it can be cached to reduce encoding overhead
+function wrpc_service:encode_args(name, ...)
+  local rpc = self:get_method(name)
+  if not rpc then
+    return nil, string.format("no method %q", name)
```
mikefero, line 324:
```suggestion
    return nil, string.format("unknown method %q", name)
```

mikefero (commented):
Looks great so far; some nits and suggestions for now; feel free to punt some of them during another iteration.

First pass; need to look over `kong/tools/wrpc.lua` more though.

kong/clustering/init.lua
```diff
@@ -127,7 +127,14 @@ function _M.new(conf)
...
 
-  self.child = require("kong.clustering." .. conf.role).new(self)
+  print("role: ", conf.role, "  protocol: ", conf.cluster_protocol)
+
+  local clustering_submodule = conf.role
+  if conf.cluster_protocol == "wRPC" then
+    clustering_submodule = "wrpc_" .. clustering_submodule
+  end
+
+  self.child = require("kong.clustering." .. clustering_submodule).new(self)
```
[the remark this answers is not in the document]
javierguerragiraldez: not a fan of using input data for filesystem search of code. it's after passing through the "enum" check, but still...

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,690 @@
...
+local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
+local MAJOR_MINOR_PATTERN = "^(%d+)%.(%d+)%.%d+"
+local _log_prefix = "[clustering] "
+
+local wrpc_config_service
+
+local function get_config_service(self)
+  if not wrpc_config_service then
+    wrpc_config_service = wrpc.new_service()
+    wrpc_config_service:add("kong.services.config.v1.config")
```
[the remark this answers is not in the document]
javierguerragiraldez: a new protocol version would require new code.  i think it's best that each version of the code hardcodes the appropriate version of the .proto; in fact, .proto files are part of the code, even if loaded dynamically instead of generating static code as on Go

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,690 @@
...
+        minor   = minor,
+        version = version,
+      }
+
+    else
+      versions[name] = {}
+    end
+  end
+  return versions
+end
```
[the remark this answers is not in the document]
javierguerragiraldez: both `wrpc_control_plane.lua` and `wrpc_data_plane.lua` started as writing on top of `control_plane.lua` and `data_plane.lua` respectively.  when I got it mostly working, I renamed both and pulled the original ones back unchanged.

The idea is to make it **very** clear that the default case is the same, unchanged code as before.  After that gets accepted, we can refactor and even remove the flag in favor of version negotiation.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,690 @@
...
+      self.plugins_configured[plugin.name] = true
+    end
+  end
+
+  config_version = config_version + 1
+
+  -- store serialized plugins map for troubleshooting purposes
+  local shm_key_name = "clustering:cp_plugins_configured:worker_" .. ngx.worker.id()
+  kong_dict:set(shm_key_name, cjson_encode(self.plugins_configured));
+  ngx_log(ngx_DEBUG, "plugin configuration map key: " .. shm_key_name .. " configuration: ", kong_dict:get(shm_key_name))
```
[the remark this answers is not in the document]
javierguerragiraldez: i don't think this is useful for debug logging, i put it when trying to understand how all this work.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,690 @@
...
+    client.peer:send_encoded_call(self.config_call_rpc, self.config_call_args)
+
+    n = n + 1
+  end
+
+  ngx_log(ngx_NOTICE, _log_prefix, "config version #", config_version, " pushed to ", n, " clients")
+end
+
+
+function _M:validate_shared_cert()
```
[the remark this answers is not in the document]
javierguerragiraldez: yes, they're really not protocol-related.  i'm still a bit adverse on changing existing code this close to feature freeze

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,690 @@
...
+  end
+
+  -- TODO: DAOs are not checked in any way at the moment. For example if plugin introduces a new DAO in
+  --       minor release and it has entities, that will most likely fail on data plane side, but is not
+  --       checked here.
+
+  return true, nil, CLUSTERING_SYNC_STATUS.NORMAL
+end
+
+function _M:handle_cp_websocket()
```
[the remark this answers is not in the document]
javierguerragiraldez: this function is exported and called from `clustering/init.lua` for both protocols, makes sense to keep the same name/signature

kong/tools/channel.lua
javierguerragiraldez: yeah, naming is hard.  basically these are the same as the shdict `:add()` and `:get()` methods but which block the coroutine until the operation gets done (or timeout).  i'll add some doccoments there.

kong/tools/wrpc.lua
javierguerragiraldez: i don't like it too much, it's there to preserve the plurality of the `payloads` argument, both as input and output

kong/tools/wrpc.lua
javierguerragiraldez: i though i had removed all these!

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,690 @@
...
+      end
+
+    elseif err ~= "timeout" then
+      ngx_log(ngx_ERR, _log_prefix, "semaphore wait error: ", err)
+    end
+  end
+end
+
+
+function _M:init_worker()
```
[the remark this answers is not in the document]
javierguerragiraldez: in this case i think further evolution would make these more different than right now.  maybe as soon as with the addition of version negotiation.

mikefero (approved):
Great work! Let's get this in and then iterate on the remaining items in your list and then we can focus on other issues as them come in. With this code being completely isolated from the JSON payload we will be in good shape.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,690 @@
...
+local function remove_nulls(tbl)
+  for k,v in pairs(tbl) do
+    if v == ngx_null then
+      tbl[k] = nil
+    elseif type(v) == "table" then
+      tbl[k] = remove_nulls(v)
+    end
+  end
+  return tbl
+end
```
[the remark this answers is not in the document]
javierguerragiraldez: i haven't checked too much into `kong.db.declarative`, as I want it to be fully compatible with the existing data flow.... for now.  I'd much prefer that `:export_config()` didn't add those `json.null` or `ngx.null` objects.

kong/conf_loader/init.lua
```diff
@@ -636,6 +636,7 @@ local CONF_INFERENCES = {
   lua_socket_pool_size = { typ = "number" },
 
   role = { enum = { "data_plane", "control_plane", "traditional", }, },
+  cluster_protocol = { enum = { "json", "wRPC" }, },
```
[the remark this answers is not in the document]
javierguerragiraldez: done

mikefero (approved):

kong/include/kong/services/legal/README.md
```diff
@@ -0,0 +1,16 @@
+# Legal service
+
```
hbagdi, line 2: Please remove this. This was not supposed to be part of OSS.

kong/include/kong/services/metrics/v1/README.md
```diff
@@ -0,0 +1,3 @@
+## Metrics
+
+Metrics service is used by a DP to report back its metrics to the CP.
```
hbagdi, line 3: Please remove this service as well, this was not supposed to be part of OSS.

kong/include/kong/services/event_hooks/v1/event_hooks.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
+
```
hbagdi, line 2: This file was never supposed to be part of OSS. Please remove it.

kong/include/kong/model/events/event_hooks.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
```
hbagdi, line 2: Please remove this.

kong/include/kong/services/legal/README.md
javierguerragiraldez: done

kong/include/kong/services/event_hooks/v1/event_hooks.proto
javierguerragiraldez: done

kong/include/kong/services/metrics/v1/README.md
javierguerragiraldez: done

kong/include/kong/model/events/event_hooks.proto
javierguerragiraldez: done

kong/include/kong/model/target.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+
+option go_package = "github.com/kong/koko/internal/gen/wrpc/kong/model;model";
+
+package kong.model;
+
+import "kong/model/upstream.proto";
+
+message Target {
+  double created_at = 1;
```
zhongchen, line 10: the `type` should be `int32` instead of `double`? it is defined as `int32` in all other places.

javierguerragiraldez: i've forwarded the fix to the original repo (these files are copied there, ideally should be copied during build time and not on the repo.  we'll get there soon)

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,686 @@
...
+local wrpc_service = {}
+wrpc_service.__index = wrpc_service
+
+
+--- a `service` object holds a set of methods defined
+--- in .proto files
+function wrpc.new_service()
+  if not wrpc_proto then
+    local wrpc_protofname = assert(proto_searchpath("wrpc.wrpc"))
+    wrpc_proto = assert(grpc.each_method(wrpc_protofname))
```
Suika-Kong, line 135: It is a bit confusing that wRPC should call gRPC's method. Maybe we can extract the protobuf part out?

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,662 @@
...
+    _, err, client.sync_status = self:check_version_compatibility(dp_version, client.dp_plugins_map, log_suffix)
+    if err then
+      ngx_log(ngx_ERR, _log_prefix, err, log_suffix)
+      wb:send_close()
+      client:update_sync_status()
+      return ngx_exit(ngx_CLOSE)
+    end
+  end
+
+  self:push_config()    -- first config push
```
Suika-Kong, line 528: Won't this cause a config push to all nodes for every new connection?
Maybe this would do.

```suggestion
  do
    local payload, err = self:export_deflated_reconfigure_payload()
    if not payload then
      ngx_log(ngx_ERR, _log_prefix, "unable to export config from database: ", err)
      return
    end
    ngx_log(ngx_NOTICE, _log_prefix, "config version #", config_version, " pushed to ", dp_id)
  end
```

Suika-Kong: Also, maybe we can cache the "payload" to speed up the case, where new nodes come and the config is not modified.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -0,0 +1,662 @@
...
+  end
+
+  local n = 0
+  for _, client in pairs(self.clients) do
+    client.peer:send_encoded_call(self.config_call_rpc, self.config_call_args)
+
+    n = n + 1
+  end
+
+  ngx_log(ngx_NOTICE, _log_prefix, "config version #", config_version, " pushed to ", n, " clients")
```
Suika-Kong, line 183: I notice that in the non-wrpc version, we log in a different format. Should we unify them?

Suika-Kong: Also, the log level is different.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -0,0 +1,374 @@
...
+  end
+
+  local entities, err, _, meta, new_hash =
+              self.declarative_config:parse_table(config_table, config_hash)
+  if not entities then
+    return nil, "bad config received from control plane " .. err
+  end
+
+  if declarative.get_current_hash() == new_hash then
+    ngx_log(ngx_INFO, _log_prefix, "same config received from control plane, ",
```
Suika-Kong, line 82: Ditto.

kong/tools/grpc.lua
```diff
@@ -13,7 +13,7 @@ local grpc = {}
 
 local function safe_set_type_hook(type, dec, enc)
   if not pcall(pb.hook, type) then
-    ngx.log(ngx.NOTICE, "no type '" .. type .. "' defined")
+    ngx.log(ngx.DEBUG, "no type '" .. type .. "' defined")
```
Suika-Kong, line 16: Log level changed.

Suika-Kong (commented):
And maybe more refactoring(extracting common codes) also helps when implementing the version negotiation.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -0,0 +1,374 @@
...
+        " (retrying after ", reconnection_delay, " seconds)", log_suffix)
+
+      assert(ngx.timer.at(reconnection_delay, function(premature)
+        self:communicate(premature)
+      end))
+      return
+    end
+  end
+
+  local config_semaphore = semaphore.new(0)
```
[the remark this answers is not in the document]
javierguerragiraldez: good idea.  the whole `peer` object should be dropped if the connection is lost, but not sure if semaphore objects are "sticky" and might avoid GC.

kong/clustering/wrpc_control_plane.lua
javierguerragiraldez: good call.

kong/tools/grpc.lua
javierguerragiraldez: yes, this message appears constantly because the protobuf state is global; but in most cases can be safely ignored so i wanted to quiet it down a bit.  still don't want to remove it because it _is_ useful when adding new "special types"

kong/clustering/wrpc_data_plane.lua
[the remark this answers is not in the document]
javierguerragiraldez: ... on the other hand there's no `semaphore.reset()`.  i'll just add a `peer.semaphore = nil` to be sure it's removed as soon as possible.

kong/clustering/wrpc_control_plane.lua
javierguerragiraldez: done!

Suika-Kong (approved):

javierguerragiraldez:
rebased on top of master

kong/clustering/wrpc_data_plane.lua
```diff
@@ -0,0 +1,379 @@
...
+  local peer = wrpc.new_peer(c, get_config_service(), { channel = true })
```
zhongchen, line 258: What's the best way for me to share a wrpc connection among Ngnix workers? Should each worker create its own peer or should I share the peer among workers?

zhongchen: I need to let each Nginx worker push its local data to the cp.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -0,0 +1,379 @@
...
+  end
+
+  -- Here we spawn two threads:
+  --
+  -- * config_thread: it grabs a received declarative config and apply it
+  --                  locally. In addition, this thread also persists the
+  --                  config onto the local file system
+  -- * ping_thread: performs a ConfigService.PingCP call periodically.
+
+  local ping_immediately
```
zhongchen, line 289: `ping_immediately` is never set in this file, while this field is set in `data_plane.lua`.

kong/clustering/wrpc_data_plane.lua
javierguerragiraldez: one worker owns the connection and creates a peer.  the other workers create a remote_peer, which use a channel to communicate with the "real" one.

right now the channel name is a hardcoded constant, so there can only be one "main" peer (the one created in wrpc_data_plane) and all remote_peers go to that single one.  so anywhere in the code you can use `wrpc.new_remote_peer()` and it will use it.   if we require separate wrpc connections, we'll have to pass the channel name somewhere.

note, the "main" peer doesn't encode/decode the remote_peer's data, so you can add new service definitions and don't have to add to the "main" peer, just the remote_peer

kong/clustering/wrpc_data_plane.lua
javierguerragiraldez: good catch.  in the original `data_plane.lua` it's used to send the first ping with the config hash and prime the config pushing from the control_plane; but in wrpc the config is initiated by the control plane and this isn't required.

i'll remove the variable

kong/clustering/wrpc_data_plane.lua
zhongchen: I will create another wrpc connection for the telemetry endpoint.

zhongchen: There will only be one wrpc connection for each data plane seems to be a very restrictive assumption.

javierguerragiraldez: then i'll have to add a "connection name" and move the current constant from `wrcp.lua` to `wrpc_data_plane.lua`

zhongchen: I am wondering why do you need a channel? It seems that you never create a `remote_client`. no usage of `wprc.new_remote_client`

javierguerragiraldez: the config handling is done completely in a single worker so it doesn't need it; but the wrpc has to be usable from any worker, so i added channel.lua and new_remote_client()

zhongchen: Cool. I was just curious why you enabled the `channel` for the config connection, while it is not needed. https://github.com/Kong/kong/blob/7e9a9a28e82a7a58a3c501f05f241b137f676f69/kong/clustering/wrpc_data_plane.lua#L257

javierguerragiraldez: it's expected that any code would have to initiate calls to the CP.  this allows any part of code to just do it without having to deal with connection handling.

javierguerragiraldez: done

kong/clustering/wrpc_data_plane.lua
```diff
@@ -0,0 +1,374 @@
...
+      self:communicate(premature)
+    end))
+  end
+end
+
+
+local wrpc_config_service
+local function get_config_service()
+  if not wrpc_config_service then
+    wrpc_config_service = wrpc.new_service()
```
zhongchen, line 182: do all ngnix workers need to share the same service or each worker can create its own?

zhongchen: I need to do something like
```
    self.peer = wrpc.new_remote_client(get_requests_log_service(), self.TELEMETRY_CHANNEL_NAME)
```

zhongchen: My understanding is that `local` means each worker will create its own `wrpc_requests_log_service`

javierguerragiraldez: > ... share the same service ...
the service is a Lua object.  as such, it can't be shared between workers.

yes, each worker will need its own, like everything in openresty.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -0,0 +1,374 @@
...
+local inflate_gzip = utils.inflate_gzip
+local deflate_gzip = utils.deflate_gzip
+
+
+local KONG_VERSION = kong.version
+local CONFIG_CACHE = ngx.config.prefix() .. "/config.cache.json.gz"
+local ngx_ERR = ngx.ERR
+local ngx_DEBUG = ngx.DEBUG
+local ngx_INFO = ngx.INFO
+local MAX_PAYLOAD = constants.CLUSTERING_MAX_PAYLOAD
```
zhongchen, line 35: `CLUSTERING_MAX_PAYLOAD` is not part of the constants anymore. https://github.com/Kong/kong/pull/8337/files

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,685 @@
...
+    if typ == "close" then
+      kong.log.notice("Received WebSocket \"close\" frame from peer: ", err, ": ", data)
+      return self:close()
+    end
+  end
+end
+
+--- RPC call.
+--- returns the call sequence number, doesn't wait for response.
+function wrpc_peer:call(name, ...)
```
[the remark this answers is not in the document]
javierguerragiraldez: i guess you're asking about remote_peers.  they have their own implementation of `:call()`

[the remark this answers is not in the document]
zhongchen: so I should do something like
```
assert(self.peer:remote_call("RequestsLogService.SendLogs", { request_metadata = request_metadata }))
``` for remote peers
and
```
assert(self.peer:call("RequestsLogService.SendLogs", { request_metadata = request_metadata }))
``` for the peer?

**Comment by Suika-Kong** [comment]
Your PR seems ready to merge, but there are some lint issues that fail checks. May you spend some time fixing that?

File: kong/templates/kong_defaults.lua [hunk 30]
```diff
@@ -22,6 +22,7 @@ admin_listen = 127.0.0.1:8001 reuseport backlog=16384, 127.0.0.1:8444 http2 ssl
 status_listen = off
 cluster_listen = 0.0.0.0:8005
 cluster_control_plane = 127.0.0.1:8005
+cluster_protocol = json
```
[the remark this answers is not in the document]
hbagdi: Is this the right default?

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,685 @@
...
+
+        elseif keyword == "rpc" then
+          identifier = scope_name .. "." .. identifier
+        end
+
+        local type_annotations = annotations[keyword]
+        if type_annotations then
+          local tag_key, tag_value = annotation:match("^%s*(%S-)=(%S+)%s*$")
+          if tag_key and tag_value then
+            tag_value = tag_value
```
Suika-Kong, line 172: It seems of no effect?

javierguerragiraldez: right.  seems a leftover from debugging code

kong/tools/channel.lua
```diff
@@ -0,0 +1,164 @@
...
+
+  return nil, err
+end
+
+
+--- Waits until a value is posted by any client
+--- @param dict shdict shdict to use
+--- @return any, string, string value, channel name, error
+function Channel.wait_all(dict)
+  local name, err = get_wait(dict, NAME_KEY, now() + DEFAULT_TIMEOUT)
```
Suika-Kong, line 138: Could we just use one dictionary key, and put the name of the channel together into the value? So we could fetch only once.
After all, post keeps blocked if `NAME_KEY` is not free.

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,685 @@
...
+    end
+  end
+  proto_f:close()
+
+  grpc.each_method(service_fname, function(_, srvc, mthd)
+    assert(srvc.name)
+    assert(mthd.name)
+    local rpc_name = srvc.name .. "." .. mthd.name
+
+    local service_id = assert(annotations.service[srvc.name] and annotations.service[srvc.name]["service-id"])
```
Suika-Kong, line 188: Could we just do this?
```lua
If service_id[srvc.name] == nil then
  service_id[srvc.name] =  new_service_id
  new_service_id = new_service_id +1
end
```

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,685 @@
...
+        end
+
+        local type_annotations = annotations[keyword]
+        if type_annotations then
+          local tag_key, tag_value = annotation:match("^%s*(%S-)=(%S+)%s*$")
+          if tag_key and tag_value then
+            tag_value = tag_value
+            local tags = type_annotations[identifier] or {}
+            type_annotations[identifier] = tags
+            tags[tag_key] = tag_value
```
Suika-Kong, line 175: tags seem not used

kong/tools/channel.lua
javierguerragiraldez: this is one of the reasons to move to a socket implementation, it's hard to follow which concepts tie 1:1 into which and which are 1:n
in this case, i think the purpose was to reduce serialization of processes, each client pushes their value without blocking each other, and only block on making the "server" pull the notification.    at higher level i think there's still the same contention anyway... but it's important to be able for each client to fire the message and not wait for all the others to be served.

kong/tools/wrpc.lua
javierguerragiraldez: you mean dynamically assign ids to methods and rpcs?  no, we must respect those in the `.proto` file

kong/tools/wrpc.lua
javierguerragiraldez: they're part of the `annotation` structure, it's put into `type_annotations`, which is a subtable of `annotations`.  after scanning the file, it's read to get the `service_id` and `rpc_id` on the `.each_method()` loop

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,685 @@
...
+  rpc.response_handler = response_handler
+  return rpc
+end
+
+
+--- Part of wrpc_peer:call()
+--- If calling the same method with the same args several times,
+--- (to the same or different peers), this method returns the
+--- invariant part, so it can be cached to reduce encoding overhead
+function wrpc_service:encode_args(name, ...)
```
Suika-Kong, line 236: It seems that all call to `call` and `call_wait` uses a table to contain arguments.

kong/tools/wrpc.lua
```diff
@@ -0,0 +1,685 @@
...
+  local rpc = self:get_method(name)
+  if not rpc then
+    return nil, string.format("unknown method %q", name)
+  end
+
+  local num_args = select('#', ...)
+  local payloads = table.new(num_args, 0)
+  for i = 1, num_args do
+    payloads[i] = assert(pb.encode(rpc.input_type, select(i, ...)))
```
Suika-Kong, line 245: Encode every arguments with the same `input_type` seems a bug.

kong/tools/wrpc.lua
javierguerragiraldez: not sure what do you mean.   the specific case i had in mind is the CP sending the same config to several DP nodes.  instead of repeating the call with the unencoded config, it calls `service:encode_args()` once, and then `peer:send_encoded_call()` for each DP node

kong/tools/wrpc.lua
javierguerragiraldez: the RPC definitions in .proto files only specify one input type.   still, the frame structure supports repeated payloads.  a single call could specify several items of the same time instead of using an array.

kong/tools/wrpc.lua
Suika-Kong: I mean like this:
```lua
local resp, err = peer:call_wait("ConfigService.ReportMetadata", { plugins = self.plugins_list })
assert(peer:call("ConfigService.PingCP", { hash = hash }))
```
This comment is related to the comment below.

kong/tools/wrpc.lua
Suika-Kong: Maybe I misinterpreted the comment in the wrpc.proto:
```protobuf
  // payloads is an array representing the request or response data of an RPC.
  // A request message MAY contain multiple elements where each element represents
  // an argument to the RPC call. The order of the elements MUST correspond to
  // the order of arguments in the function call.
  // A response message MUST contain a single payload.
  // Use a wrapper type for RPCs which contain multiple responses.
  // Unless otherwise specified by the Service or RPC, the encoding method in
  // use is PROTO3.
  //
  // Note: This results in double proto3 encoding. Once encoding of payloads,
  // and then encoding of the entire message. We acknowledge that there are
  // some overheads here and they will be addressed in a future iteration.
  repeated bytes payloads = 9;
```

By "each element represents an argument" I thought it was for every member of the request message (like "hash" and "plugins" in the previous comment).
Also, if an RPC call receives repeated messages, it cannot use repeat like this:

```proto
  rpc SomeCall(repeated SomeCallRequest) returns (SomeCallResponse);
```

But to use something like:

```proto
Message RepeatedSomeCallRequest {
    repeated SomeCallRequest arguments = 1;
}
```

So I don't think it makes sense to encode an array of payloads...

javierguerragiraldez: > Also, if an RPC call receives repeated messages, it cannot use repeat like this:
>
>  rpc SomeCall(repeated SomeCallRequest) returns (SomeCallResponse);

right, the syntax doesn't support this.  but in the current implementation, the "repeated" is implied, so this could be done:

```protobuf
rpc SomeCall(SomeCallRequest) returns (someCallResponse);

message SomeCallRequest {
    string name = 1;
    string value = 2;
}
```
and in Lua:
```lua
   peer:call("serv.SomeCall", {name="one", value=1}, {name="two", value=2}, {name="three", value=3})
```

and it would result in a single call with three payload elements of the same type.

But I agree that it's not an important usecase, and the structure you describe would be just as good.

Suika-Kong: @hbagdi May you clarify this? Do we really want to implement it this way?

PR #8376, opened 2022-02-07
chore(*) use sock:sslhandshake instead of sock:tlshandshake

chronolaw:
### Summary

Openresty will support mtls with `sslhandshake` soon,
we need not to use own special `tlshandshake`,
and in this case, it only uses normal ssl feature,
we can change to `sslhandshake` safely.

see:
https://github.com/openresty/lua-nginx-module/pull/1602
https://github.com/openresty/lua-resty-core/pull/278

### Full changelog

* use sock:sslhandshake instead of sock:tlshandshake

### DIFF
--- a/kong/reports.lua
+++ b/kong/reports.lua
@@ -53,10 +53,8 @@ local _ping_infos = {}
 local _enabled = false
 local _unique_str = utils.random_string()
 local _buffer_immutable_idx
-local _tls_session
-local _tls_opts = {
-  ssl_verify = false,
-}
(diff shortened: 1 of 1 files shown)


fffonion:
@chronolaw Could you add the PR/commit in openresty side for this feature for tracking?

pintsized (approved):
👍

chronolaw:
> @chronolaw您能否在 openresty 端添加 PR/commit 以跟踪此功能？

done, thanks.

PR #8389, opened 2022-02-10
feat(vaults) adds vaults support to kong (beta)

bungle:
### Summary

Check info on commit messages.


kong/cmd/vault.lua
```diff
@@ -0,0 +1,122 @@
...
+
+Vault utilities for Kong.
+
+Example usage:
+ TEST=hello kong vault get env/test
+
+The available commands are:
+  get <reference>  Retrieves a value for <reference>
+
+Options:
```
kikito, line 112: Command has no options, I would leave those out for now:

```suggestion
Usage: kong vault COMMAND [OPTIONS]

Vault utilities for Kong.

Example usage:
 TEST=hello kong vault get env/test

The available commands are:
  get <reference>  Retrieves a value for <reference>

```

spec/02-integration/02-cmd/14-vault_spec.lua
```diff
@@ -0,0 +1,111 @@
...
+    helpers.kill_all()
+  end)
+
+  lazy_teardown(function()
+    helpers.clean_prefix()
+  end)
+
+  it("vault help", function()
+    local ok, stderr, stdout = helpers.kong_exec("vault --help")
+    assert.matches("Usage: kong vault COMMAND [OPTIONS]", stderr, nil, true)
```
kikito, line 19: Removed the options from the command
```suggestion
    assert.matches("Usage: kong vault COMMAND", stderr, nil, true)
```

spec/01-unit/23-vaults_spec.lua
```diff
@@ -0,0 +1,133 @@
...
+    v = "test_value"
+  },
+  ["{vault://env/test}"] = {
+    k = "TEST",
+    v = ""
+  },
+}
+
+
+describe("Vault PDK", function()
```
kikito, line 31: We usually test PDK methods on the t/ folder. Is there a reason why this was not done there? (we can move those there after feature freeze).

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -18,6 +18,60 @@ describe("kong start/stop #" .. strategy, function()
     helpers.clean_prefix()
   end)
 
+  it("with #referenced values [fails]", function()
```
kikito, line 21: removed tag, reads like English
```suggestion
  it("fails with referenced values that are not initialized", function()
```

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -18,6 +18,60 @@ describe("kong start/stop #" .. strategy, function()
...
+    end
+    assert.matches("Kong started", stdout, nil, true)
+    assert(helpers.kong_exec("stop", {
+      prefix = helpers.test_conf.prefix,
+    }))
+
+    helpers.clean_logfile()
+  end)
+
+  it("with #referenced secrets [vault does not exist]", function()
```
kikito, line 40:
```suggestion
  it("fails to read referenced secrets when vault does not exist", function()
```

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -18,6 +18,60 @@ describe("kong start/stop #" .. strategy, function()
...
+    assert.matches("failed to dereference {vault://non-existent/pg_password}: could not find vault (non-existent)", stderr, nil, true)
+    assert.matches("Kong started", stdout, nil, true)
+    assert(helpers.kong_exec("stop", {
+      prefix = helpers.test_conf.prefix,
+    }))
+
+    helpers.clean_logfile()
+  end)
+
+  it("with #referenced secrets [resolves]", function()
```
kikito, line 57:
```suggestion
  it("resolves referenced secrets", function()
```

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -18,6 +18,60 @@ describe("kong start/stop #" .. strategy, function()
...
+      prefix = helpers.test_conf.prefix,
+      database = helpers.test_conf.database,
+      pg_password = "{vault://env/pg_password}",
+      pg_database = helpers.test_conf.pg_database,
+      cassandra_keyspace = helpers.test_conf.cassandra_keyspace
+    }))
+    if strategy == "postgres" then
+      assert.matches("referenced value has a wrong type, expected 'string', got 'nil'", stderr, nil, true)
+    end
+    assert.matches("Kong started", stdout, nil, true)
```
kikito, line 32: I think Kong should not start if the secret isn't present. Seems like a possible security problem.

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -18,6 +18,60 @@ describe("kong start/stop #" .. strategy, function()
     helpers.clean_prefix()
   end)
 
+  it("with #referenced values [fails]", function()
+    local _, stderr, stdout = assert(helpers.kong_exec("start", {
+      prefix = helpers.test_conf.prefix,
+      database = helpers.test_conf.database,
+      pg_password = "{vault://env/pg_password}",
+      pg_database = helpers.test_conf.pg_database,
+      cassandra_keyspace = helpers.test_conf.cassandra_keyspace
```
kikito, line 27: If vaults are behind a feature flag, shouldn't there be one here (and in other tests that use vaults)?

kong/conf_loader/init.lua
```diff
@@ -1571,6 +1594,41 @@ local function load(path, custom_conf, opts)
...
+          conf[k] = deref
+          --refs[k] = v
+        end
+
+        local schema_v = CONF_INFERENCES[k]
+        if schema_v then
+          if schema_v.typ ~= type(deref) then
+            -- TODO: what to do here? Fail hard or error log?
+            -- return nil, fmt("failed to dereference <%s> for config option %s. k, v)
+            log.error(fmt("referenced value has a wrong type, expected '%s', got '%s'", schema_v.typ, type(deref)))
```
kikito, line 1617: If we have secrets "active" (perhaps with a feature flag in the config?) then I think we should fail hard and explicitly here. Otherwise we risk leaving Kong in an unstable / insecure state.

```suggestion
            return nil, fmt("failed to dereference <%s> for config option %s. k, v)
```

kong/conf_loader/init.lua
```diff
@@ -1571,6 +1594,41 @@ local function load(path, custom_conf, opts)
...
+          end
+        end
+
+        if deref == nil or deref_err then
+          -- TODO: what to do here? Fail hard or error log?
+          -- return nil, fmt("failed to dereference <%s> for config option %s. k, v)
+          log.error(fmt("failed to dereference %s: %s", v, deref_err))
```
kikito, line 1624: Same reasoning as above - if secrets are active and it looks like a secret, but is not retrievable, I think we should fail hard.
```suggestion
          return nil, fmt("failed to dereference <%s> for config option %s. k, v)
```

kong/constants.lua
```diff
@@ -49,6 +49,21 @@ for _, plugin in ipairs(deprecated_plugins) do
   deprecated_plugin_map[plugin] = true
 end
 
+local vaults = {
```
kikito, line 52: This is empty. Is that correct?

kong.conf.default
```diff
@@ -95,6 +95,14 @@
                                           # is adjusted by the `log_level`
                                           # property.
 
+#vaults = off                    # Comma-separated list of vaults this node
```
kikito, line 98: Does this act as our "feature flag?" Is vault code inactive unless this is not empty?

kikito (changes requested):
Thank you for the detailed commits, that made it easier to review.

My main concern is that we wanted vaults to be behind a feature flag. I am not sure if having an empty `vaults` property in the config qualifies. I see a lot of code that seems to be "always active".

I understand that making some changes is inevitable (we will have to make a migration, and we will have to bump the PDK version) but it seems we are activating a lot of code even when the `vaults` property is empty. Perhaps I have missed a critical `if` somewhere.

Also, minor but important: `referenceable` instead of `referencable`.

kong/conf_loader/init.lua
```diff
@@ -1571,6 +1594,41 @@ local function load(path, custom_conf, opts)
     conf.loaded_plugins = setmetatable(plugins, _nop_tostring_mt)
   end
 
+  ---------------------------------
+  -- Dereference process references
+  ---------------------------------
+
+  do
+    local vault = require "kong.pdk.vault".new()
+    --local refs = {}
```
jschmid1, line 1603:
```suggestion
```

kong/conf_loader/init.lua
jschmid1, line 1617: I think we wanted to fail hard for ProcessSecrets. 

```suggestion
            local type_err_msg = fmt("referenced value has a wrong type, expected '%s', got '%s'", schema_v.typ, type(deref))
            log.error(type_err_msg)
            return nil, type_err_msg
```

jschmid1: Uh, @kikito was quicker. What he said

kong/db/schema/metaschema.lua
```diff
@@ -83,6 +83,7 @@ local field_schema = {
   { immutable = { type = "boolean" }, },
   { err = { type = "string" } },
   { encrypted = { type = "boolean" }, },
+  { referencable = { type = "boolean" }, },
```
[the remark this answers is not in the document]
bungle: Both are correct.

kong/conf_loader/init.lua
```diff
@@ -1571,6 +1594,41 @@ local function load(path, custom_conf, opts)
...
+
+        if deref == nil or deref_err then
+          -- TODO: what to do here? Fail hard or error log?
+          -- return nil, fmt("failed to dereference <%s> for config option %s. k, v)
+          log.error(fmt("failed to dereference %s: %s", v, deref_err))
+        end
+      end
+    end
+
+    --conf["$refs"] = refs
```
jschmid1, line 1629: We're not carrying forward the $refs since we do not need them _yet_ I suppose?

kong/cmd/vault.lua
bungle: There are default options like `-v` and `-vv` that are printed below this, so it needs to be there.

spec/02-integration/02-cmd/14-vault_spec.lua
bungle: There are default options, see above.

spec/01-unit/23-vaults_spec.lua
bungle: We can add some stuff inside PDK tests too.

spec/02-integration/02-cmd/02-start_stop_spec.lua
bungle: Yes, we are still doing this "how to manage errors" on each case.

kong/db/schema/init.lua
```diff
@@ -1644,57 +1661,131 @@ function Schema:process_auto_fields(data, context, nulls, opts)
...
+          if subfield.type == "string" and subfield.referencable then
+            local count = #value
+            if count > 0 then
+              for i = 1, count do
+                if is_reference(value[i]) then
+                  --if not ref then
+                  --  ref = { value }
+                  --else
+                  --  ref[i] = value
+                  --end
```
jschmid1, line 1757: Let's add sentence as to why this is commented out for now

kong/pdk/vault.lua
```diff
@@ -0,0 +1,281 @@
...
+  else
+    value, err = strategy.get(config, resource, version)
+  end
+
+  return validate_value(value, err, name, resource, key, reference)
+end
+
+
+---
+-- TODO: docs
```
jschmid1, line 172: an open TODO

kong/pdk/vault.lua
```diff
@@ -0,0 +1,281 @@
...
+     and byte(reference, -1)  == BRACE_END
+     and byte(reference, 7)   == COLON
+     and byte(reference, 8)   == SLASH
+     and byte(reference, 9)   == SLASH
+     and sub(reference, 2, 6) == "vault"
+end
+
+
+---
+-- TODO: docs
```
jschmid1, line 186: an open TODO

kong/pdk/vault.lua
```diff
@@ -0,0 +1,281 @@
...
+    resource = resource,
+    key = key,
+    config = config,
+    version = version,
+  }
+end
+
+
+---
+-- TODO: docs
```
jschmid1, line 255: an open TODO

kong/db/schema/init.lua
bungle: It is on separate commit. Perhaps we should just remove the whole code for now. Just thinking what we put in beta.

kong.conf.default
bungle: Not all code is inactive, but most of the code just runs if there are references. I am thinking about adding if statements if vaults is empty. Yes, I am thinking about using this as feature flag. Similarly we could perhaps optimize a lot of kong if no plugins are specified.

PR #8392, opened 2022-02-10
chore(runloop) use localized var in worker events handler

chronolaw:
### Summary

There are some localized var like `operation` `target` `upstream`,
but these localized var are not used wisely.
I don't know the reason, please check the code or give me an explain.

### Full changelog

* use localized var in place of original var


locao (approved):
It seems those were leftovers, good catch!

PR #8395, opened 2022-02-11
chore(router) use table.new to process header_values

chronolaw:
### Summary

`table.new` may be better than raw `{}`.

### Full changelog

* use `table.new` to create new table in processing headers


vaibh289:
Good

PR #8396, opened 2022-02-11
chore(router) remove [0] in methods_t/snis_t

chronolaw:
### Summary

It seems that `methods_t` / `snis_t` does not use function `append()`,
so the `[0]` is useless.
And we should create new table only when we really need,
maybe this will save our memory and CPU time.

### Full changelog

* remove [0] in methods_t/snis_t
* introduce const EMPTY_A
* create table in the right place


javierguerragiraldez (commented):
it makes me a bit uneasy that all these variables share the same table instance before they're needed.  it seems an easy mistake to add something before creating them, and it would affect all the other aliases.
i think it would be better to keep them as `nil` before they're used.

PR #8397, opened 2022-02-11
feat(clustering) initial version negotiation endpoint

javierguerragiraldez:


spec/fixtures/custom_nginx.template
```diff
@@ -503,6 +503,17 @@ http {
...
+            limit_except POST {
+                deny all;
+            }
+            if ($content_type != "application/json") {
+                return 400;
+            }
+            content_by_lua_block {
+                Kong.serve_version_handshake()
+            }
+        }
```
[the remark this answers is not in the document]
javierguerragiraldez: mostly to share the same port and SSL certificates as the websocket endpoint.  otherwise we would need another config option to specify how to connect to the negotiation endpoint.

of course, this `.serve_version_handshake()` could be implemented with lapis, but i don't see any advantage.

kong/clustering/utils.lua
```diff
@@ -0,0 +1,206 @@
...
+        err = "data plane client certificate revocation check failed: " .. err
+
+      else
+        ngx_log(ngx_WARN, _log_prefix, "data plane client certificate revocation check failed: ", err)
+        err = nil
+      end
+    end
+  end
+
+  print("final err: ", tostring(err))
```
flrgh, line 195: leftover local debug?

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+  local services_map = {}
+  for _, service in ipairs(services_accepted) do
+    services_map[service.name] = service.version
+  end
+
+  local ok, err = kong.db.clustering_data_planes:upsert({ id = client_node.id, }, {
+    last_seen = ngx.time(),
+    config_hash = "00000000000000000000000000000000",
+    hostname = client_node.hostname,
+    ip = ngx.var.realip_remote_addr or ngx.var.remote_addr,
```
flrgh, line 182: Any particular reason for using `$realip_remote_addr` here? I would think we would want to play nicely with a control plane behind a reverse proxy, no?

javierguerragiraldez: apparently i copied that from the `pdk.client.get_ip()` function.  Yes, control planes behind a reverse proxy are definitely a thing, so I'd think using the value provided by the proxy would be better than what the socket sees.  Do I have it backwards?

kong/clustering/utils.lua
javierguerragiraldez: ups!

kong/clustering/version_negotiation/init.lua
flrgh: Gotcha, yeah `$realip_remote_addr` will always contain the actual IP of the immediate downstream/client (as opposed to what is derived from X-Forwarded-For), so I think `$remote_addr` is all you need here:

```lua
ip = ngx.var.remote_addr,
```

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+
+  local params = {
+    scheme = "https",
+    method = "POST",
+    headers = {
+      ["Content-Type"] = "application/json",
+    },
+    body = body,
+
+    ssl_verify = false,
```
flrgh, line 258: Should this be `true`?

javierguerragiraldez: surprisingly, no.   i think it's related to differences between lua-resty-http and lua-resty-websockets.  probably also because the main verification is not during the SSL handshake, but afterwards, when the CP does the certs verification according to the cluster_xxx config parameters.

kong/clustering/version_negotiation/init.lua
javierguerragiraldez: right, it took me several readings of the nginx docs, but i think i get it now, the "real ip" module sets the "normal" variables using proxy info, but stores the "original" ip in the `real_ip...` vars.   kinda using "real" word as meaning both the "real origin" and the "original origin".   big mess.

kong/clustering/version_negotiation/init.lua
flrgh: Ah okay. In this case I think a code comment would be helpful to assuage the worries of other readers and confirm that it's not an oversight. Some variation of:
> it looks like we're doing an unsafe thing here, but it's actually okay because of [...]

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+end
+
+
+local function node_info()
+  return {
+    id = kong.node.get_id()
+  }
+end
+
+local function cp_priority(name, req_versions, known_versions)
```
flrgh, line 84: There's a lot of business logic encoded in this module, and from the perspective of somebody who is not already well-versed in the version negotiation flow, I'm finding myself bouncing around the file a lot re-reading code to fully grok things. The code isn't particularly opaque or anything; it's just that I'm finding myself wondering:

> is the behavior in function 'foo' part of some specification/API/contract, or is it just an implementation detail or incidental thing that I'm not meant to care about?

Let's take this `cp_priority` function, for instance. Based on following the behavior of the function line-by-line I _think_ I am meant to understand that the ordering of the table in `kong.clustering.version_negotiation.services_known` is a significant piece of the business logic and determines which version is ultimately utilized, since we return on the first found item in the table. However, because there's not really an explanation or test case for this behavior, I'm not 100% confident in that I've made the right conclusion about this. All that's to say that I would be pretty nervous about touching this code if I needed to fix a bug or add a feature.

I think some comments and doc strings that help to outline/explain the process would be helpful (even if it's just to link out to an ADR or something). Also, while I _think_ the integration tests exercise most--if not all--of this code, the existence of some unit cases that specifically target the negotiation business logic (however simple) would be very helpful to future maintainers.

flrgh:
Code looks good/correct to me :+1: I have raised some readability/maintainability concerns that I personally think would be good to address before merging, but I'm happy to hear a different opinion on these things

kong/clustering/version_negotiation/services_known.lua
```diff
@@ -0,0 +1,6 @@
+return {
+  cluster_protocol = {
```
zhongchen, line 2: what does the `services_known` mean here? My understanding is that this file really just defines the different endpoints along with the different protocols they support. `cluster` is not really a `service` but an endpoint. In the `wrpc` mode, it can actually be used for the communications of multiple `services` defined in https://github.com/Kong/kong/pull/8357/files#diff-82fded69f85ff8ba289c5a4e082e55791e36b462a43e2465f50848513a3187ee

zhongchen: I guess in the existing setup, `endpoint` and `service` has one to one mapping, so it is fined to use them interchangeably. However, it won't be the case anymore once we start using wrpc as the underlying protocol. Is it correct?

javierguerragiraldez: yes, I understand how wrpc is intended to support several sets of methods, calling each one a "service".  the main use for this version negotiation will be to determine which of these are supported on both sides; but could be anything that we make optional or multi-versioned.

this `services_known.lua` file is the list of services the CP supports, and the order of versions on each one determines the priority (if more than one version is supported on both sides, the first one in this file "wins").

The content here is just the minimal example to make the system work, so I decided to use the most basic part of DP/CP communication: the protocol itself, which we will have to support at least two versions soon (the current json-based and the wrpc in #8357)

kong/clustering/version_negotiation/init.lua
javierguerragiraldez: yes, the version priority is determined by the CP only; and currently the way to set it is by the ordering of version records in the `services_know.lua` file.

javierguerragiraldez: as for unit tests... there were some!   i was surprised not to find them anymore; I removed them when they became obsolete on a rereading of the spec :grin: and apparently forgot to rewrite them.  will add some

kong/clustering/version_negotiation/services_known.lua
zhongchen: Cool. I was just trying to understand what additional changes I would need to make to support the `analytics` service we were going to introduce in kong 3.0. https://github.com/Kong/kong-dp-spec/blob/c83273a7eb10c8960c95f415c5882e62748e112c/spec/proto/kong/services/analytics/v1/analytics.proto

zhongchen: I wasn't sure whether I needed to update this file or not.

javierguerragiraldez: yes, just add a new service like
```lua
...
  analytics = {
    { version = "v1", message = "first!" },
  },
...
```

also, since protobuf makes it simple to add fields, and everything is optional, in many cases it "should" be possible to extend services without having to declare new versions.  my guess is that it would be necessary only if we decide to change an expected sequence of calls, or make a method not only obsolete but inconvenient.  i hope in most cases the trick would be to carefully choose the default case semantics.

kong/db/migrations/core/015_280_to_300.lua
```diff
@@ -0,0 +1,19 @@
+return {
```
zhongchen, line 1: All the db related changes are only needed for the control plane, right? If I were to connect my dp only to koko, I wouldn't need to port those code to the ee repo?

javierguerragiraldez: right.  i think DP (or non-hybrid) never instantiates those Node entities.

kong/clustering/version_negotiation/services_requested.lua
```diff
@@ -0,0 +1,6 @@
+return {
```
zhongchen, line 1: This file is per endpoint, right? Or it is fine to list all the requested_services here, supposing we have two different endpoints that support different services?

javierguerragiraldez: there's no "endpoint" concept in this code, in fact not even a hard definition of what a "service" actually is.  this negotiation has it's own endpoint and doesn't care what the negotiated things are.   This file holds everything the DP will request from the CP, in a single request/response.

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+  if client_node.type ~= "KONG" then
+    return nil, ("unknown node type %q"):format(client_node.type), CLUSTERING_SYNC_STATUS.UNKNOWN
+  end
+
+  local ok, msg, status = clustering_utils.check_kong_version_compatibility(KONG_VERSION, client_node.version)
+  if not ok then
+    return ok, msg, status
+  end
+
+  return true, nil, CLUSTERING_SYNC_STATUS.NORMAL
```
Suika-Kong, line 115:
```suggestion
  return clustering_utils.check_kong_version_compatibility(KONG_VERSION, client_node.version)
```

Maybe this would do?

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+end
+
+local function do_negotiation(req_body)
+  local services_accepted = {}
+  local services_rejected = {}
+
+  for _, req_service in ipairs(req_body.services_requested) do
+    if type(req_service) ~= "table" or type(req_service.name) ~= "string" then
+      goto continue
```
Suika-Kong, line 124: Should we reject with 400?

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+    end
+
+    local name = req_service.name
+
+    if type(req_service.versions) ~= "table" then
+      table.insert(services_rejected, {
+        name = name,
+        message = "invalid \"versions\" table.",
+      })
```
Suika-Kong, line 133: Ditto.

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+    if versions_set[v.version] then
+      return true, {
+        name = name,
+        version = v.version,
+        message = v.message,
+      }
+    end
+  end
+
+  return false, { name = name, message = "No valid version" }
```
Suika-Kong, line 100: Maybe handle this from caller?

```suggestion
    if versions_set[v.version] then
      return v
    end
  end
```

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+  ok, err, body_in.node.sync_status = check_node_compatibility(body_in.node)
+  if not ok then
+    return response_err(err)
+  end
+
+  local body_out = do_negotiation(body_in)
+
+  ok, err = register_client(conf, body_in.node, body_out.services_accepted)
+  if not ok then
+    return response_err(err)
```
Suika-Kong, line 230: Maybe this should be considered 500, rather than 400?

spec/fixtures/custom_nginx.template
```diff
@@ -503,6 +503,17 @@ http {
                 Kong.serve_cluster_listener()
             }
         }
+        location = /version-handshake {
+            limit_except POST {
+                deny all;
+            }
+            if ($content_type != "application/json") {
+                return 400;
```
Suika-Kong, line 511: Ditto.

kong/clustering/data_plane.lua
```diff
@@ -209,6 +210,17 @@ function _M:communicate(premature)
 
   local conf = self.conf
 
+  do
+    local response_data, err = version_negotiation.request_version_handshake(conf, self.cert, self.cert_key)
+    if not response_data then
+      ngx_log(ngx_ERR, _log_prefix, "error while requesting version negotiation: " .. err)
+      assert(ngx.timer.at(math.random(5, 10), function(premature)
+        self:communicate(premature)
```
Suika-Kong, line 218: If CP doesn't understand the process of the handshake, unlikely it will understand next try.
How should we handle this error?

Suika-Kong: Also, this is a to-do to check CP actually supports the clustering protocol. And before sending the request we can check whether there's a cached response in the shm.

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+  if conf.cluster_mtls == "shared" then
+    params.ssl_server_name = "kong_clustering"
+  else
+    -- server_name will be set to the host if it is not explicitly defined here
+    if conf.cluster_server_name ~= "" then
+      params.ssl_server_name = conf.cluster_server_name
+    end
+  end
+
+  local c = http.new()
```
Suika-Kong, line 271: Maybe we can use connection passed from an argument, so that later communication can be done through the same connection, like `Keep-Alive`?

Suika-Kong (commented):
Maybe not that important, but currently, it seems we put no requirement on DP to be backward compatible to CP. So DP can simply reject any CP that does not understand the version negotiation. It is possible for DP to be compatible with old version CP, we can just continue with JSON implementation. This behaviour is like HTTP `Upgrade` header: if the opponent doesn't understand, just fallback.

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,340 @@
...
+
+  if type(body.node.type) ~= "string" then
+    return false, "field \"node.type\" must be a string."
+  end
+
+  if type(body.node.version) ~= "string" then
+    return false, "field \"node.version\" must be a string."
+  end
+
+  if type(body.services_requested) ~= "table" then
```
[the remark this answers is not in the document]
javierguerragiraldez: openresty's "isarray" and relatives are SO badly designed it's not even funny

kong/clustering/version_negotiation/init.lua
javierguerragiraldez: i don't think so; since we have a requested service name, it's nicer to just reject it.

kong/clustering/version_negotiation/init.lua
javierguerragiraldez: the intention is that this function could be one of a few different version priority policies.  that's why it can choose priority and also reject and choose the message.

kong/clustering/data_plane.lua
javierguerragiraldez: AFAICT, this is the same logic as on the current CP/DP connection, if it doesn't work, keep retrying.  my guess is that it makes it simpler to (re)start both sides without worrying about which one must be first.

kong/templates/nginx_kong.lua
```diff
@@ -489,6 +489,17 @@ server {
             Kong.serve_cluster_listener()
         }
     }
+    location = /version-handshake {
+        limit_except POST {
+            deny all;
+        }
+        if ($content_type != "application/json") {
+            return 400;
```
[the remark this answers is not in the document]
javierguerragiraldez: done

kong/clustering/version_negotiation/init.lua
javierguerragiraldez: done

javierguerragiraldez:
rebased on top of master

kong/clustering/version_negotiation/init.lua
Suika-Kong: Not sure if this is proper but it's just an idea: https://github.com/openresty/lua-resty-websocket/issues/73

kong/clustering/version_negotiation/services_known.lua
```diff
@@ -0,0 +1,6 @@
...
+    { version = "json", message = "current" },
+    { version = "wrpc", message = "beta" },
```
hbagdi, line 4: The version was assumed to be "v1", "v2" but wasn't clarified in the spec.
Can we keep "v0" as what exists today and then "v1" as the RPC based protocol?

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,355 @@
...
+local ngx_DEBUG = ngx.DEBUG
+local _log_prefix = "[version-negotiation] "
+
+local KONG_VERSION
+local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
+
+
+local _M = {}
+
+local function validate_connection_type()
```
hbagdi, line 21: nit:
The code is validating a request and not a connection.
```suggestion
local function validate_request_type()
```

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,355 @@
...
+  end
+
+  ok, err = validate_connection_type()
+  if not ok then
+    return response_err(err)
+  end
+
+  local body_in = cjson.decode(get_body())
+  if not body_in then
+    return response_err("Not valid JSON data")
```
hbagdi, line 224:
```suggestion
    return response_err("not valid JSON data")
```

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,355 @@
...
+
+    local name = req_service.name
+
+    if type(req_service.versions) ~= "table" then
+      table.insert(services_rejected, {
+        name = name,
+        message = "invalid \"versions\" table.",
+      })
+      goto continue
+    end
```
hbagdi, line 141: I'd be inclined to send back a 400 (with no negotiation whatsoever) if the version is not an array.

kong/clustering/version_negotiation/init.lua
```diff
@@ -0,0 +1,355 @@
...
+
+local function register_client(conf, client_node, services_accepted)
+  local services_map = {}
+  for _, service in ipairs(services_accepted) do
+    services_map[service.name] = service.version
+  end
+
+  local ok, err = kong.db.clustering_data_planes:upsert({ id = client_node.id, }, {
+    last_seen = ngx.time(),
+    config_hash = "00000000000000000000000000000000",
```
hbagdi, line 186: Please use a constant: https://cs.github.com/Kong/kong?q=DECLARATIVE_EMPTY_CONFIG_HASH

kong/clustering/version_negotiation/services_known.lua
javierguerragiraldez: done

kong/clustering/version_negotiation/services_requested.lua
```diff
@@ -0,0 +1,6 @@
...
+  {
+    name = "cluster_protocol",
```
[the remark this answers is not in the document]
javierguerragiraldez: changed to `Config`.

kong/db/migrations/core/015_280_to_300.lua
```diff
@@ -0,0 +1,19 @@
...
+    postgres = {
+      up = [[
```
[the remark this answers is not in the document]
javierguerragiraldez: currently the database stores all the known state for each DP: `last_seen`, `ip`, `config_hash`, `version`, `sync_status`... adding the negotiated state seems natural.  Also, given the limitation from lua-resty-websocket, which prevents from reusing the connection between negotiation and websocket, it's very likely that they would be served by different workers.  The safest way to carry this state between these two requests is to store in the database.  We could also use a shared dict, but I doubt it would be significantly better.

kong/clustering/utils.lua
```diff
@@ -0,0 +1,204 @@
...
+  if minor_cp < minor_dp then
+    return nil, "data plane version " .. dp_version ..
+      " is incompatible with older control plane version " .. cp_version,
+    CLUSTERING_SYNC_STATUS.KONG_VERSION_INCOMPATIBLE
+  end
+
+  if minor_cp ~= minor_dp then
+    local msg = "data plane minor version " .. dp_version ..
+      " is different to control plane minor version " ..
+      cp_version
```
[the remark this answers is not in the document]
javierguerragiraldez: it's the same code currently used for the DP-CP connection, just moved to a separate function to share between requests.

kong/clustering/version_negotiation/init.lua
javierguerragiraldez: done

kong/clustering/version_negotiation/init.lua
javierguerragiraldez: done

kong/db/migrations/core/015_280_to_300.lua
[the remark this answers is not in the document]
hbagdi: >Also, given the limitation from lua-resty-websocket, which prevents from reusing the connection between negotiation and websocket, it's very likely that they would be served by different workers. The safest way to carry this state between these two requests is to store in the database. We could also use a shared dict, but I doubt it would be significantly better.

This is exactly what I'm concerned about. If the persisted state is used later on outside the lifecycle of a TCP connection, it implies coordination between control-plane nodes since the design allows for load-balancing of multiple connections amongst a set of nodes.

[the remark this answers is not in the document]
javierguerragiraldez: isn't that a good reason precisely to store in the database?

kong/clustering/data_plane.lua
```diff
@@ -219,6 +220,17 @@ function _M:communicate(premature)
...
+      end))
+      return
+    end
```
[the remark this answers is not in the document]
javierguerragiraldez: on the DP side, the result is stored in a shared dict and available via `get_negotiated_service()`.

kong/db/migrations/core/015_280_to_300.lua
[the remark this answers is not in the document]
hbagdi: I think we should avoid any coordination vai the database at this level because:
- added latency to setup the wRPC state, very relevant since kong does polling and not a notify based event propagation
- wRPC's scope is limited to TCP connection and nothing outside it, which obviates away cross-node synchronization and consistency issues. This is very much intentional.

kong/clustering/data_plane.lua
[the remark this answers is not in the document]
hbagdi: What I meant was if the negotiation failed (CP doesn't have the necessary services), should this code not error out somewhere (and subsequently retry or give up)?

kong/clustering/version_negotiation/services_requested.lua
```diff
@@ -0,0 +1,6 @@
+return {
+  {
+    name = "Config",
```
hbagdi, line 3: nit: should we lowercase this and make it case insensitive?

kong/clustering/data_plane.lua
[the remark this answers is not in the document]
javierguerragiraldez: ah, you mean if the CP responds with 200 but a "critical" service is rejected?  I think in that case whatever code requires that service will query the result (with `get_negotiated_service()`) and if it was rejected then it should do whatever it can (probably complain noisily and abort).

I consider this module is strictly a communication mechanism, not policy.  It's up to the users of this feature to establish relevant behaviours.

kong/db/migrations/core/015_280_to_300.lua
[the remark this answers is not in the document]
javierguerragiraldez: there are two levels of cache between Lua code and the actual database, it's not likely to add significant latency unless it has time to get cold.
should we fork lua-resty-websocket to be able to use the same TLS connection?  i think yes; but not before Kong 3.0

kong/clustering/version_negotiation/services_requested.lua
javierguerragiraldez: done

kong/clustering/data_plane.lua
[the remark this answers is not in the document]
hbagdi: Fair enough, thanks for explaining.

kong/db/migrations/core/015_280_to_300.lua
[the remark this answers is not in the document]
hbagdi: > there are two levels of cache between Lua code and the actual database, it's not likely to add significant latency unless it has time to get cold.

If the request lands in the same node, sure. If the request lands on another node, that's a whole another thing since the cache is lazily built.

>  should we fork lua-resty-websocket to be able to use the same TLS connection? i think yes; but not before Kong 3.0

Absolutely. Please track this work item so that we do not forget it.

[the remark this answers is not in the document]
flrgh: _NOTE/FYI:_ we already have [a fork of lua-resty-websocket](https://github.com/Kong/lua-resty-websocket) for additional features, and I am working to get it up on luarocks soon for easy inclusion into Kong :)

kong/clustering/init.lua
```diff
@@ -215,13 +271,36 @@ function _M:init_worker()
...
+
+      local config_proto = version_negotiation.get_negotiated_service("config")
+      ngx_log(ngx_DEBUG, _log_prefix, "config_proto: ", config_proto)
+      if config_proto == "v1" then
+        self.child = require "kong.clustering.wrpc_data_plane".new(self)
+
+      elseif config_proto == "v0" or config_proto == nil then
+        self.child = require "kong.clustering.data_plane".new(self)
+
+      end
```
hbagdi, line 296: Nice to see negotiation actually happening.
I think having a cluster_protocol configuration value in kong.conf that is interpreted only by the CP would be helpful.
The default value will be empty, which translates to "use whatever the code is doing, do not influence it).
This configuration option could be used as an escape hatch in the case where wRPC implementation has a serious bug. This should be done in a follow up PR. What do you think?

kong/clustering/utils.lua
```diff
@@ -0,0 +1,279 @@
+
+
```
hbagdi, line 2: Why these line jumps?

hbagdi:
I don't have any additional comments here.
@dndx @fffonion Do you mind reviewing this once and shipping it?

kong/clustering/init.lua
javierguerragiraldez: i'm thinking of a more generic "killswitch", that simply removes options from the requested services.  something like `banned=metrics, config.v0` would remove the whole `metrics` service, and the option `v0` from `config`.

kong/clustering/init.lua
```diff
@@ -215,13 +271,36 @@ function _M:init_worker()
...
+
+  if role == "data_plane" and ngx.worker.id() == 0 then
+    assert(ngx.timer.at(0, function(premature)
+      if premature then
+        return
+      end
+
+      self:request_version_negotiation()
+
+      local config_proto = version_negotiation.get_negotiated_service("config")
```
fffonion, line 288: worth log out the `err` from `get_negotiated_service` if there's any

fffonion (approved):

PR #8403, opened 2022-02-11
feat(vaults) adds vaults beta support to kong

bungle:
### Summary

Check info on commit messages.


kong/pdk/vault.lua
```diff
@@ -0,0 +1,404 @@
...
+--
+-- @usage
+-- local value, err = kong.vault.get("{vault://env/cert/key}")
+local function get(reference)
+  local opts, err = parse_reference(reference)
+  if err then
+    return nil, err
+  end
+
+  local value = LRU:get(reference)
```
jschmid1, line 374: If we receive a cached value, we can immediately return it
```suggestion
  local value = LRU:get(reference)
  if value then
    return value
  end
```

jschmid1 (approved):

PR #8406, opened 2022-02-12
fix(aws-lambda) proxy correctly with lua-resty-http

### LINKED ISSUE #8317: Timeout when calling lambda

2.7.0

### Current Behavior

When I call lambda function through aws-lambda plugin, the lambda function is not triggered and I see the error message below:
```
[aws-lambda] timeout, client: 172.18.0.10, server: kong, request: "GET /public HTTP/1.0", host: "xxxx", referrer: "https://xxx"

172.18.0.10 - - [20/Jan/2022:03:05:52 +0000] "GET /public HTTP/1.0" 500 42 "https://xxx" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36"
```

### Expected Behavior

The lambda function should be triggered and response returned.

### Steps To Reproduce

```markdown
1. Run Kong instance on AWS EC2.
2. Access to the Internet via proxy.
3. Configuration for lambda plugin:
 
- name: aws-lambda
   route: hello
   config:
     aws_region: ap-southeast-2
     invocation_type: RequestResponse
     log_type: Tail
     proxy_url: http://proxy:3128
     proxy_scheme: http
     function_name: hello-world
     timeout: 9000
     awsgateway_compatible: true
     is_proxy_integration: true
```

### Anything else?

It works fine with Kong version 2.4.1 but it does not with versions 2.6.0 and 2.7.0. 
Same situation occurs in Kong-enterprise edition.

flrgh:
## Summary

The aws-lambda plugin currently mishandles user proxy configuration, passing `proxy_opts.http_proxy` to `resty.http` when it ought to be passing `proxy_opts.https_proxy` (since the request to AWS lambda is _always_ https).

This PR remedies that problem along with a couple other shortcomings surrounding the config schema:

1. The `proxy_scheme` param is unused and is not needed with the current implementation (`proxy_url` expects a full url with scheme). This PR deprecates it, logging a warning that it will be removed in Kong 3.0.
2. lua-resty-http [does not support https connection to proxy](https://github.com/ledgetech/lua-resty-http/blob/9bf951dfe162dd9710a0e1f4525738d4902e9d20/lib/resty/http_connect.lua#L142-L145), so extra validation has been added to `proxy_url` to ensure only http is used.

## Testing

This plugin used to ship with it's own specialized http connect code (and tests). This is not used anymore, so the associated tests (`spec/03-plugins/27-aws-lambda/50-http-proxy_spec.lua`) have been removed, as they do not test anything specific to this plugin or its code.

The access handler tests didn't have any coverage for instances using an http proxy, so I sought to add one, which was a little bit more involved of a chore than I expected. The repo's CI pipeline stuff includes a squid container (previously used for the aforementioned http connect tests), but using this in tests is a royal pain:

1. We can't mock DNS inside squid without significant effort
2. We can't simply use `127.0.0.1` for all targets without requiring the container to run in host networking mode.

So I ditched squid and wrote a minimal TCP server that implements CONNECT (`spec/fixtures/forward-proxy-server.lua`). It's not awesome, but it gets the job done.

Result: the test coverage for aws-lambda+http proxy still leaves much to be desired, but this is a step in the right direction.

## Full changelog

* fix(aws-lambda) deprecate proxy_scheme parameter
* fix(aws-lambda) always use https_proxy
* fix(aws-lambda) ensure proxy_url scheme is http
* tests(aws-lambda) remove old http proxy tests

## Issue reference

Fix #8317


kong/plugins/aws-lambda/schema.lua
```diff
@@ -75,6 +75,7 @@ return {
           type = "boolean",
           default = false,
         } },
+        -- TODO: remove proxy_scheme in Kong 3.0
```
hbagdi, line 78: @bungle We should explore a "deprecated = true" property on the schema to manage these programmatically.

spec/fixtures/forward-proxy-server.lua
```diff
@@ -0,0 +1,116 @@
+local _M = {}
```
aboudreault, line 1: what's the difference between this and the naive forward proxy fixture here: https://github.com/Kong/kong-ee/pull/2945/files#diff-a2b4ed7d776c435259b62afad583c89bd0ce9f9143d3ba4d104500e352c10f7aR16 

should we use only one implementation for tests (if they are doing the same thing)?

flrgh: Oof. There's no difference--just two developers managing to unknowingly implement the same test fixture within \*checks other commit\* 24 hours of each other. :upside_down_face:

spec/fixtures/aws-lambda.lua
```diff
@@ -96,6 +99,17 @@ local fixtures = {
   },
 }
 
+fixtures.stream_mock = {
+  lambda_proxy = [[
+    server {
+      listen 13128;
+
+      content_by_lua_block {
+        require("spec.fixtures.forward-proxy-server").connect()
```
aboudreault, line 108: I like that

aboudreault (approved):
I think we want to include this fix in 2.8.0. @flrgh @kikito do you agree? 

If so, we just need to change the target to release/2.8.x and/or create another PR for it (cherry-pick).

flrgh:
> I think we want to include this fix in 2.8.0. @flrgh @kikito do you agree?

Yes, this is intended to land in 2.8.0.

> If so, we just need to change the target to release/2.8.x and/or create another PR for it (cherry-pick).

I don't have any strong preference here, so whatever you think is best!

PR #8412, opened 2022-02-14
feat(conf) allow user-specified OpenResty install path

flrgh:
### Summary

I maintain several different installations of OpenResty (both vanilla and Kong-flavored) on my machine for dev/testing of different projects.

The first time I went to run the Kong test suite locally, I spent a half hour or so scratching my head and cursing the sky over this error:

```
nginx: [emerg] unknown directive "lua_kong_load_var_index" in /home/flrgh/git/kong/kong/servroot/nginx.conf:69
nginx: configuration file /home/flrgh/git/kong/kong/servroot/nginx.conf test failed
```

_But I prepended the correct NGINX bin dir to my PATH already--why doesn't it work?!_

As I would soon discover, Kong searches some common OpenResty system/default paths _before_ consulting PATH, making it impossible to use a non-standard OpenResty installation for running the Kong cli, busted, etc. if system OpenResty is already installed.

The commit message has some more details, but in short, with this change, operators can use a custom OpenResty install with Kong by setting ~`openresty_prefix`~ `openresty_path`.

###  Why not just update things so that we search PATH first?

This would be a nice improvement, and in my opinion would nicely correct things as to not violate the principle of least surprise. However, that would be a breaking change with potential far-reaching, headache-inducing impact as operators, package maintainers, and CI pipeline owners now suffer through the same debug session that I went through.

### Why ~openresty_prefix~ openresty_path and not nginx_bin?

Kong is tightly-coupled to OpenResty than vanilla NGINX, so it "feels" more cohesive this way, to me at least. _Also_ this leaves the door open to expand our usage of this variable. For instance, I think it would be nice if it had a cascading effect on the default values of lua `package.path`/`package.cpath`.

### Why use ~openresty_prefix~ openresty_path exclusively instead of inserting it at the head of the search list?

Because the operator told us explicitly which path to use, and if it doesn't exist, we should fail fast rather than potentially falling back to another version of OpenResty that happens to exist on the same filesystem.

---

For discussion:

~Is it okay for this to slide in as an undocumented, not-exactly-public-but-not-completely-private/internal feature, or do we need to make it a fully-documented first-class citizen of the conf_loader? IMO the latter would be nice, but I stopped short of that because I think it would need to be more fleshed-out first--potentially incorporating some of the other behaviors I mentioned (re: lua `package.path`).~ I'm adding it to the Kong config.

### DIFF
--- a/kong/cmd/utils/nginx_signals.lua
+++ b/kong/cmd/utils/nginx_signals.lua
@@ -8,11 +8,22 @@ local pl_stringx = require "pl.stringx"
 local fmt = string.format
 
 local nginx_bin_name = "nginx"
-local nginx_search_paths = {
-  "/usr/local/openresty/nginx/sbin",
-  "/opt/openresty/nginx/sbin",
-  ""
-}
+local nginx_search_paths
+do
+  local resty_prefix = os.getenv("KONG_OPENRESTY_PREFIX")
+  if resty_prefix then
+    log.debug("using custom OpenResty prefix: %s", resty_prefix)
+    nginx_search_paths = {
+      pl_path.join(resty_prefix, "nginx", "sbin")
+    }
+  else
+    nginx_search_paths = {
+      "/usr/local/openresty/nginx/sbin",
+      "/opt/openresty/nginx/sbin",
+      ""
+    }
+  end
+end
 local nginx_version_pattern = "^nginx.-openresty.-([%d%.]+)"
 local nginx_compatible = version.set(unpack(meta._DEPENDENCIES.nginx))
 
--- a/kong/cmd/utils/nginx_signals.lua
+++ b/kong/cmd/utils/nginx_signals.lua
@@ -55,10 +55,10 @@ function _M.find_nginx_bin(kong_conf)
   log.debug("searching for OpenResty 'nginx' executable")
 
   local search_paths = nginx_search_paths
-  if kong_conf and kong_conf.openresty_prefix then
-    log.debug("using custom OpenResty prefix: %s", kong_conf.openresty_prefix)
+  if kong_conf and kong_conf.openresty_path then
+    log.debug("using custom OpenResty path: %s", kong_conf.openresty_path)
     search_paths = {
-      pl_path.join(kong_conf.openresty_prefix, "nginx", "sbin"),
+      pl_path.join(kong_conf.openresty_path, "nginx", "sbin"),
     }
   end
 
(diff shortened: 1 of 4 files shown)


locao (approved):
This was indeed missing. 
IMO this should be fully-documented.

kong/cmd/utils/nginx_signals.lua
```diff
@@ -8,11 +8,22 @@ local pl_stringx = require "pl.stringx"
...
 
 local nginx_bin_name = "nginx"
-local nginx_search_paths = {
-  "/usr/local/openresty/nginx/sbin",
-  "/opt/openresty/nginx/sbin",
-  ""
-}
+local nginx_search_paths
+do
+  local resty_prefix = os.getenv("KONG_OPENRESTY_PREFIX")
```
gszr, line 13: Should this be made a kong config property instead? It would fit nicely the miscellaneous category that includes `lua_package_path`. Property could be named `openresty_prefix`; this way, it would be specified as either the property or `KONG_OPENRESTY_PREFIX` env var.

gszr (approved):

kikito:
note: no tests on this pR

flrgh:
Test coverage has been added.

dndx:
@flrgh This is a nice idea, but I would like to point out that `openresty_prefix` may not be the best name for it. `prefix` in OpenResty really means "working directory", that is, where OpenResty and Nginx logs, runtime data and etc should be stored at. It generally does not have much to do with where the `nginx` binary is located.

Maybe `openresty_path` makes more sense, as it is indeed a `PATH` that Kong should be searching for.

flrgh:
> @flrgh This is a nice idea, but I would like to point out that `openresty_prefix` may not be the best name for it. `prefix` in OpenResty really means "working directory", that is, where OpenResty and Nginx logs, runtime data and etc should be stored at. It generally does not have much to do with where the `nginx` binary is located.
> 
> Maybe `openresty_path` makes more sense, as it is indeed a `PATH` that Kong should be searching for.

This is a fair point, and I'm not married to `openresty_prefix` by any stretch. I'll rename to `openresty_path`.

kikito (approved):

spec/02-integration/02-cmd/15-utils_spec.lua
```diff
@@ -0,0 +1,72 @@
+describe("kong cli utils", function()
+
+  describe("nginx_signals", function()
+    local signals = require "kong.cmd.utils.nginx_signals"
+    local pl_path = require "pl.path"
+    local pl_file = require "pl.file"
+    local pl_dir = require "pl.dir"
```
kikito, line 7: Minor: could we move these requires to the top of the file? We don't usually put the requires inside functions unless necessary.

gszr:
yay! nice work there @flrgh!

PR #8414, opened 2022-02-15
fix(plugin_server) skip preloading headers in stream subsystem

fffonion:
### Summary

Skip preloading headers in stream subsystem.
Note this PR doesn't try to guard from user actively call `kong.request.get_header{,s}`,
which will result into an error, but rather let the log phase proceed in stream subsystem

### Full changelog

* skip preloading headers in stream subsystem


bungle (approved):

bungle:
A bit worrying to me is that we don’t have a single test of external plugin server with stream.

fffonion:
@bungle That's a good point, creating a ticket to add tests; need to look at how/where current test are being run.

PR #8427, opened 2022-02-17
fix(pdk) fill default values with vault config required fields

bungle:
### Summary

This fixes the vault to fill default values for configuration based on a config field schema in case the field is required and has a default value, when there is no configuration for the field already.

### DIFF
--- a/kong/pdk/vault.lua
+++ b/kong/pdk/vault.lua
@@ -160,12 +160,14 @@ local function process_secret(reference, opts)
     local fields = field.fields
     local env_name = gsub(name, "-", "_")
     for i = 1, #fields do
-      local k = next(fields[i])
+      local k, f = next(fields[i])
       if config[k] == nil then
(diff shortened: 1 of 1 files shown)


flrgh:
The code change looks good/correct to my eyes. Are there any relevant tests we could amend to cover this case?

bungle:
> The code change looks good/correct to my eyes. Are there any relevant tests we could amend to cover this case?

Unfortunately the `env` vault has not required fields that have default value, So it would mean to create a mock Vault. I tested this manually. As this is still beta, and nothing is shipped, there are a lot of holes in testing. I would love to see if our direction is right. Let's add these more specific tests later.

flrgh:
> As this is still beta, and nothing is shipped, there are a lot of holes in testing. I would love to see if our direction is right. Let's add these more specific tests later?

Ah okay I wasn't realizing this was beta code. WFM!

flrgh (approved):

PR #8438, opened 2022-02-21
chore(deprecation) add deprecation warning

gszr:
The properties `go_pluginserver_exe` and `go_pluginserver` are
deprecated in favor of new multi-pluginservers implementation.
Compatibility code will remain in place until 3.0, when it will be
removed.


javierguerragiraldez (approved):
:partying_face:

PR #8452, opened 2022-02-23
perf(core) do not register event handlers on control planes

bungle:
### Summary

Control planes only enable Admin API and Admin API does not use cache,
nor does it need to build router or plugins iterator based on events.


dndx:
Just one question. If we skip the worker events registration in CP, will it cause the necessary cluster events to not be generated between CPs running in a cluster?

e.g. if there are 2 CPs running inside the same cluster, will this cause the other CP (the one not receiving the Admin API call) to miss the event and not push out new configs to it's DP?

bungle:
> Just one question. If we skip the worker events registration in CP, will it cause the necessary cluster events to not be generated between CPs running in a cluster?
> 
> e.g. if there are 2 CPs running inside the same cluster, will this cause the other CP (the one not receiving the Admin API call) to miss the event and not push out new configs to it's DP?

It does not.

dndx (approved):

PR #8455, opened 2022-02-24
fix(prometheus): fix for boolean label handling

shettyh:
### Summary

Prometheus metric with boolean label (false) is ignored due to this check


hbagdi:
Any way to add a test for this?

bungle:
What about `label_values[1] ~= nil` instead?

fffonion (approved):

fffonion:
LGTM, looks like we don't run the unit test for prometheus.lua here, as the change is minimal I'm okay to just merge it. I do verify that label_values are permitted to be non-string like boolean (https://github.com/Kong/kong/blob/2a9f5c3f2c5f20a6f7a4b26eafc10d117510bc33/kong/plugins/prometheus/prometheus.lua#L176).

PR #8483, opened 2022-03-02
fix(schema-validator) does not convert the `null` in the declarative configuration to `nil`

### LINKED ISSUE #8314: An unexpected error occurred for response ratelimiting plugin

2.4.1

### Current Behavior

I have activated response rate-limiting for a consumer. When I am accessing the endpoint, it throws an error. 

**Message:**
`{
    "message": "An unexpected error occurred"
}`

**The corresponding message is on the error log.**

`[kong] init.lua:286 [response-ratelimiting] ...re/lua/5.1/kong/plugins/response-ratelimiting/access.lua:53: attempt to perform arithmetic on local 'lv' (a userdata value)`

**my plugin config**
`{
  "name": "response-ratelimiting",
  "service": null,
  "enabled": true,
  "route": null,
  "id": "127c0436-67e3-43d3-9def-a39e80a4549e",
  "consumer": {
    "id": "05d958f3-a561-4501-8a12-27c6f81c5e48"
  },
  "config": {
    "redis_timeout": 2000,
    "redis_database": 0,
    "redis_host": null,
    "redis_port": 6379,
    "limits": {
      "test": {
        "year": null,
        "second": null,
        "minute": null,
        "hour": 10,
        "day": null,
        "month": null
      }
    },
    "limit_by": "consumer",
    "redis_password": null,
    "policy": "cluster",
    "hide_client_headers": false,
    "fault_tolerant": true,
    "header_name": "X-Kong-Limit",
    "block_on_first_violation": false
  },
  "protocols": [
    "grpc",
    "grpcs",
    "http",
    "https"
  ],
  "created_at": 1642669998,
  "tags": null
}`

Response headers from service:
`X-Kong-Limit: test=1`

### Expected Behavior

Should work like the default rate-limiting plugin.

### Steps To Reproduce

```markdown
The config used for creating is given above
```

### Anything else?

_No response_

ADD-SP:
### Summary

If `config.limits.{limit_name}.*` is `null`, this plugin will raise a HTTP 500 error.

### Full changelog

* If `config.limits.{limit_name}.*` is `null`,  skip it.
* Add a related test in `spec/03-plugins/24-response-rate-limiting/01-schema_spec.lua`.

### Issue reference

Fix #8314

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -26,7 +26,6 @@ local update_time  = ngx.update_time
 local ngx_time     = ngx.time
 local ngx_now      = ngx.now
 local find         = string.find
-local null         = ngx.null
 local max          = math.max
 local sub          = string.sub
 
@@ -1578,8 +1577,15 @@ local function adjust_field_for_context(field, value, context, nulls, opts)
     end
 
     if subfield then
-      for i = 1, #value do
-        value[i] = adjust_field_for_context(subfield, value[i], context, nulls, opts)
+      if field.type ~= "map" then
+        for i = 1, #value do
+          value[i] = adjust_field_for_context(subfield, value[i], context, nulls, opts)
+        end
+
+      else
+        for k, _ in pairs(value) do
+          value[k] = adjust_field_for_context(subfield, value[k], context, nulls, opts)
+        end
       end
     end
   end
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -26,6 +26,7 @@ local update_time  = ngx.update_time
 local ngx_time     = ngx.time
 local ngx_now      = ngx.now
 local find         = string.find
+local null         = ngx.null
 local max          = math.max
 local sub          = string.sub
 
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1584,8 +1584,8 @@ local function adjust_field_for_context(field, value, context, nulls, opts)
         end
 
       else
-        for k, _ in pairs(value) do
-          value[k] = adjust_field_for_context(subfield, value[k], context, nulls, opts)
+        for k, v in pairs(value) do
+          value[k] = adjust_field_for_context(subfield, v, context, nulls, opts)
         end
       end
     end
--- a/kong/plugins/response-ratelimiting/access.lua
+++ b/kong/plugins/response-ratelimiting/access.lua
@@ -9,9 +9,6 @@ local error = error
 local tostring = tostring
 
 
-local null = ngx.null
-
-
 local EMPTY = {}
 local HTTP_TOO_MANY_REQUESTS = 429
 local RATELIMIT_REMAINING = "X-RateLimit-Remaining"
--- a/kong/plugins/response-ratelimiting/access.lua
+++ b/kong/plugins/response-ratelimiting/access.lua
@@ -39,23 +39,21 @@ local function get_usage(conf, identifier, limits, current_timestamp)
 
   for k, v in pairs(limits) do -- Iterate over limit names
     for lk, lv in pairs(v) do -- Iterare over periods
-      if lv ~= null then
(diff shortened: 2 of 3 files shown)


dndx (approved):

kong/plugins/response-ratelimiting/access.lua
```diff
@@ -36,21 +39,23 @@ local function get_usage(conf, identifier, limits, current_timestamp)
...
   for k, v in pairs(limits) do -- Iterate over limit names
     for lk, lv in pairs(v) do -- Iterare over periods
-      local current_usage, err = policies[conf.policy].usage(conf, identifier, k, lk, current_timestamp)
-      if err then
-        return nil, err
-      end
-
-      if not usage[k] then
-        usage[k] = {}
+      if lv ~= null then
```
bungle, line 42: I don't fully understand why the `lv` can be `null`, but can it be `nil` too, thus?

```suggestion
      if lv ~= nil and lv ~= null then
```

ADD-SP:
@bungle 
For example, this declarative configuration:

```yml
plugins:
    - name: response-ratelimiting
      config: 
        limits:
          test:
            second: null
            hour: 1000
        block_on_first_violation: false
        policy: cluster
        fault_tolerant: true
        hide_client_headers: false
```

`config.limits.test.second` is optional, but once it is set to `null` by the user, the value in this plugin is `ngx.null`, [related function](https://github.com/Kong/kong/blob/da413f1c79dcb705e002c69e6c04da040aa13eb6/kong/db/declarative/init.lua#L129).


However, this case does not occur if we use the Admin API to configure this plugin, this configuration will be recognized as invalid.

```shell
curl -i -X POST http://{admin_host}/consumers/{consumer_id}/plugins \
    --data "name=response-ratelimiting"  \
    --data "config.limits.test.hour=1000" \
    --data "config.limits.test.minute=null" \
    --data "config.block_on_first_violation=false" \
    --data "config.policy=cluster" \
    --data "config.fault_tolerant=true" \
    --data "config.hide_client_headers=false" \
    --data "config.header_name=X-Kong-Limit"
```

The response is:

```json
{
    "fields": {
        "config": {
            "limits": {
                "minute": "expected a number"
            }
        }
    },
    "code": 2,
    "name": "schema violation",
    "message": "schema violation (config.limits: {\n  minute = \"expected a number\"\n})"
}
```

The same goes for `PATCH` requests.

```shell
curl -i -X PATCH  http://{admin_host}/consumers/{consumer_Id}/plugins/{plugin_id} \
    --data "config.limits.test.minute=null"
```

The response is:

```json
{
    "message": "schema violation (config.limits: {\n  minute = \"expected a number\"\n})",
    "name": "schema violation",
    "fields": {
        "config": {
            "limits": {
                "minute": "expected a number"
            }
        }
    },
    "code": 2
}
```

For `POST` requests, it looks the same if the user sets an item to `null` or does not specify it. But for a `PATCH` request, if the user sets an item to `null`, then they should expect it to be removed, and if they don't specify it, then they expect no changes to be made to it.

kong/plugins/response-ratelimiting/access.lua
ADD-SP: See https://github.com/Kong/kong/pull/8483#issuecomment-1057669512 .

kong/db/schema/init.lua
```diff
@@ -1578,8 +1578,15 @@ local function adjust_field_for_context(field, value, context, nulls, opts)
...
-      for i = 1, #value do
-        value[i] = adjust_field_for_context(subfield, value[i], context, nulls, opts)
+      if field.type ~= "map" then
+        for i = 1, #value do
+          value[i] = adjust_field_for_context(subfield, value[i], context, nulls, opts)
+        end
+
+      else
+        for k, _ in pairs(value) do
+          value[k] = adjust_field_for_context(subfield, value[k], context, nulls, opts)
```
bungle, line 1588:
```suggestion
        for k, v in pairs(value) do
          value[k] = adjust_field_for_context(subfield, v, context, nulls, opts)
```

bungle (approved):

PR #8490, opened 2022-03-02
feat(*) add support for exit_worker_by_lua

bungle:
### Summary

Adds support for an OpenResty phase `exit_worker` in Kong core, and enables that phase for plugins too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,12 @@
 
 ## Unreleased
 
+### Additions
+
+#### Core
+
+- Plugins can now implement `exit_worker` handler.
+
 ### Dependencies
 
 - Bumped pgmoon from 1.13.0 to 1.14.0
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -282,6 +282,28 @@ local function execute_init_worker_plugins_iterator(plugins_iterator, ctx)
 end
 
 
+local function execute_exit_worker_plugins_iterator(plugins_iterator, ctx)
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -62,6 +62,16 @@
 - [0.10.0](#0100---20170307)
 - [0.9.9 and prior](#099---20170202)
 
+
+## [2.9.0] (UNRELEASED)
```
guanlan, line 66: 3.0 will be the next non-patch release

chronolaw:
Shall we modify `pdk.private.phases.lua`, add a new constant `exit_worker`?

kikito (changes requested):
I think this requires at least one test. I think the easiest way to test this would be a custom plugin which includes the new functionality.

kikito:
@bungle can you rebase this over master please?

StarlightIbuki:
Guess we should also support it for plugin servers?

hanshuebner:
For the benefit of everybody outside of the Kong core team, here's a bit of explanation:  The `exit_worker` phase of nginx happens when it has already closed all connections, stopped processing timers and handling its internal event loop.  This means that then environment in which `exit_worker` callbacks would execute would be very limited and much different from the normal execution environment provided to plugins.  As we are working on changes to plugins to provide a more useful way to hook into the orderly shutdown process and we also have no clear usecase for `exit_worker`, we have decided not to pursue this PR further.

PR #8513, opened 2022-03-07
fix(admin) 500 response when post /config API busy

Suika-Kong:
### Summary

fix(admin) 500 response when post /config API busy

When admin API /config is busy loading new config and another upload requested, server response with 500, and message like: failed to broadcast reconfigure event: recursive.

This is caused by `worker_event`'s implementation of detection of recursive. We solve this by rejecting upload requests when busy(429 instead of 500)

Fix FT-2472

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -30,7 +30,7 @@ local pairs = pairs
 local ngx_socket_tcp = ngx.socket.tcp
 local yield = require("kong.tools.utils").yield
 local marshall = require("kong.db.declarative.marshaller").marshall
-loacl min = math.min
+local min = math.min
 
 
 local REMOVE_FIRST_LINE_PATTERN = "^[^\n]+\n(.+)$"
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -870,7 +870,7 @@ do
     if exiting() then
       return nil, "exiting"
     end
-    
+
     local worker_events = kong.worker_events
 
     local ok, err, default_ws = declarative.load_into_cache(entities, meta, hash)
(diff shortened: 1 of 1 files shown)


kong/db/declarative/init.lua
```diff
@@ -908,12 +904,46 @@ do
...
 
+    return true
+  end
+
+  -- If it takes more than 60s it is very likely to be an internal error.
+  -- However it will be reported as: "failed to broadcast reconfigure event: recursive".
+  -- Let's paste the error message here in case someday we try to search it.
+  -- Should we handle this case specially?
+  local DECLARATIVE_LOCK_TTL = 60
+  local DECLARATIVE_MAX_RETRT = 10
```
dndx, line 915: Incorrect spellong "RETRT".

kong/db/declarative/init.lua
```diff
@@ -908,12 +904,46 @@ do
...
+    end
+
+    ok, err = load_into_cache_with_events_no_lock(entities, meta, hash)
+
+    kong_shm:delete(DECLARATIVE_LOCK_KEY)
+
+
+    if exiting() then
+      return nil, "exiting"
+    end
```
dndx, line 944: We should move this back to the beginning of `load_into_cache_with_events_no_lock`, before the config is actually loaded.

Suika-Kong: I've moved two of them back. Though I don't know why we need to check again after the config loaded, this is what it was like before this patch.

Suika-Kong: I've moved two of them back. Though I don't know why we need to check again after the config loaded, this is what it was like before this patch.

hutchic:
can / should we load test this ( cc @lukego )?

dndx (approved):

PR #8518, opened 2022-03-08
docs(*) update source install instructions

Suika-Kong:
### Summary

The chapter doesn't address patches that should be applied to OpenResty and other dependencies, and the link it contains is outdated(which redirects to kong/kong's git repo).

Fix FT-2487

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -423,7 +423,7 @@ You can add these lines to your `.profile` or `.bashrc` file.  Otherwise, you co
 # If you want to set it permanently
 echo export OPENSSL_DIR=${BUILDROOT}/openssl >> ~/.profile
 echo export PATH=${BUILDROOT}/luarocks/bin:${BUILDROOT}/openresty/bin:\${PATH} >> ~/.profile
-echo eval \$\(luarocks path\) >> ~/.profile
+echo eval "\$(luarocks path)" >> ~/.profile
 ```
 ### Databases
 
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -92,59 +92,24 @@ Kong is mostly an OpenResty application made of Lua source files, but also
 requires some additional third-party dependencies, some of which compiled
 with tweaked options, and runs on a modified version of OpenResty with patches.
 
-You can build OpenResty, OpenSSL, PCRE and Luarocks with [Kong/kong-ngx-build](https://github.com/Kong/kong-build-tools/tree/master/openresty-build-tools), which we will utilize for the following process.
-
-Below is a template for how can install from the source.
+To install from the source, first we clone the repository:
 
 ```shell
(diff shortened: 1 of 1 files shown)


DEVELOPER.md
```diff
@@ -89,21 +89,63 @@ You can use a Vagrant box running Kong and Postgres that you can find at
...
+
+cd ../kong-build-tools/openresty-build-tools
+
+# Before we run the script, make sure curl and unzip is installed.
+# Also, to build we need gcc/g++ and m4 installed.
+# Here is an example for ubuntu:
+sudo apt install curl unzip g++ m4
+
+# You might want to add also --debug
+./kong-ngx-build -p ${BUILDROOT} --openresty ${RESTY_VERSION} --openssl ${RESTY_OPENSSL_VERSION} --luarocks ${RESTY_LUAROCKS_VERSION} --pcre ${RESTY_PCRE_VERSION}
```
mayocream, line 129: Duplicated with section ["Dependencies (Build from source)"](https://github.com/Kong/kong/blob/master/DEVELOPER.md#dependencies-binary-release)

mayocream (changes requested):

DEVELOPER.md
Suika-Kong: > 

I tried to preserve the original structure of the doc and referred to this section in the source build.

Suika-Kong (commented):
I tried to preserve the original structure of the doc and referred to this section in the source build.

DEVELOPER.md
```diff
@@ -355,37 +372,59 @@ dnf install \
...
 
-You can add these lines to your `.profile` or `.bashrc` file.  Otherwise you could find yourself wondering where is everything!.
-
+You can add these lines to your `.profile` or `.bashrc` file.  Otherwise, you could find yourself wondering where is everything!.
 
+```shell
+# If you want to set it permanently
+echo export OPENSSL_DIR=${BUILDROOT}/openssl >> ~/.profile
+echo export PATH=${BUILDROOT}/luarocks/bin:${BUILDROOT}/openresty/bin:\${PATH} >> ~/.profile
+echo eval \$\(luarocks path\) >> ~/.profile
```
mayocream, line 426: Can we use `"` here, instead of `\`?

ADD-SP: > Can we use `"` here, instead of `\`?

Do you mean that?

```shell
echo "eval \$(luarocks path)"
# > eval $(luarocks path)
```

mayocream: @ADD-SP Yes, Seems more elegant.

Suika-Kong (commented):
change as requested

ADD-SP (approved):

mayocream (approved):

fffonion (approved):

PR #8519, opened 2022-03-08
perf(clustering) conditional rebuilding of router, plugins iterator and balancer on dp

bungle:
### Summary

Implements conditional rebuilding of `router`, `plugins iterator` and `balancer` on data planes. This means that DPs will not rebuild router if there were no changes in routes or services. Similarly, the plugins iterator will not be rebuild if there were no changes to plugins, and finally balancer in not reinitialized if there are no changes to upstreams or targets.

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -187,7 +187,7 @@ local function get_removed_fields(dp_version_number)
           unknown_fields[plugin] = {}
         end
         for _, k in ipairs(fields) do
-          table.insert(unknown_fields[plugin], k)
+          table_insert(unknown_fields[plugin], k)
         end
       end
     end
@@ -201,12 +201,10 @@ _M._get_removed_fields = get_removed_fields
 -- returns has_update, modified_deflated_payload, err
 local function update_compatible_payload(payload, dp_version, log_suffix)
   local fields = get_removed_fields(dp_version_num(dp_version))
-
   if fields then
     payload = utils.deep_copy(payload, false)
     local config_table = payload["config_table"]
     local has_update = invalidate_keys_from_config(config_table["plugins"], fields)
-
     if has_update then
       local deflated_payload, err = deflate_gzip(cjson_encode(payload))
       if deflated_payload then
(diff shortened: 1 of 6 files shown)


kong/clustering/control_plane.lua
```diff
@@ -235,31 +233,29 @@ function _M:export_deflated_reconfigure_payload()
...
 
   local payload = {
     type = "reconfigure",
     timestamp = ngx_now(),
     config_table = config_table,
     config_hash = config_hash,
+    hashes = hashes,
   }
 
-  if not payload then
```
fffonion, line 253: looks like a previous merge error?

kong/db/declarative/init.lua
```diff
@@ -926,7 +926,43 @@ do
...
-      ok, err = worker_events.post("declarative", "flip_config", default_ws)
+      local router_hash
+      local plugins_hash
+      local balancer_hash
+      if hashes then
+        if hashes.services ~= DECLARATIVE_EMPTY_CONFIG_HASH then
+          if hashes.routes ~= DECLARATIVE_EMPTY_CONFIG_HASH then
+            router_hash = md5(hashes.services .. hashes.routes)
+          else
+            router_hash = hashes.services
```
fffonion, line 937: if services is nont empty, and routes is empty, should we keep router_has being EMPTY_CONFIG_HASH? as we don't support service-less routes now.

kong/db/declarative/init.lua
```diff
@@ -926,7 +926,43 @@ do
...
+          router_hash = hashes.routes
+        else
+          router_hash = DECLARATIVE_EMPTY_CONFIG_HASH
+        end
+
+        plugins_hash = hashes.plugins
+
+        if hashes.upstreams ~= DECLARATIVE_EMPTY_CONFIG_HASH then
+          if hashes.targets ~= DECLARATIVE_EMPTY_CONFIG_HASH then
+            balancer_hash = md5(hashes.upstreams .. hashes.targets)
```
fffonion, line 949: maybe we simply the logic here to always just do `md5(hashes.upstreams .. hashes.targets)`? as balancer_hash is changed either upstreams or targets
changed

kong/runloop/handler.lua
```diff
@@ -360,25 +360,53 @@ local function register_events()
...
+      if type(data) == "table" then
+        default_ws = data[1]
+        router_hash = data[2]
+        plugins_hash = data[3]
+        balancer_hash = data[4]
+      end
+
       local ok, err = concurrency.with_coroutine_mutex(FLIP_CONFIG_OPTS, function()
-        balancer.stop_healthcheckers(CLEAR_HEALTH_STATUS_DELAY)
+        if balancer_hash == nil or balancer_hash ~= current_balancer_hash then
```
fffonion, line 386: as `if balancer_hash == nil or balancer_hash ~= current_balancer_hash` happens twice not constinously, to avoid us introduce bugs of modifying the hash variable in between, i would suggest we to memorize the result first, and use the memorized result in both L386 and L406, so those two branches are guranteed to be both executed or non-excuted.
```suggestion
       local should_rebuild_balancer = balancer_hash == nil or balancer_hash ~= current_balancer_hash
        if should_rebuild_balancer then
```

kong/runloop/handler.lua
```diff
@@ -360,25 +360,53 @@ local function register_events()
...
+          current_plugins_hash = plugins_hash
+        end
 
-        balancer.init()
+        if router_hash == nil or router_hash ~= current_router_hash then
+          rebuild_router(ROUTER_SYNC_OPTS)
+          current_router_hash = router_hash
+        end
+
+        if balancer_hash == nil or balancer_hash ~= current_balancer_hash then
```
fffonion, line 406: (see above)

kong/clustering/control_plane.lua
bungle: @fffonion, Yes, perhaps. Removed that as it does nothing.

fffonion (approved):
👍 🚀

PR #8522, opened 2022-03-09
Chore(labeler) does not add some labels correctly

ADD-SP:
The labeler does not add some labels correctly, such as I create a PR to change the `DEVELOPER.md`, the labeler should add the label named 'corex/docs', but it is not now.

Example: #8518


mayocream (approved):
Thanks!

PR #8526, opened 2022-03-10
fix(clustering) use ngx.exit(ngx.ERROR)

chronolaw:
### Summary

We should use `ngx.ERROR` in `ngx.exit()`, not `ngx.ERR` which is used in `ngx.log`

### Full changelog

* use ngx.exit(ngx.ERROR)


ADD-SP (approved):
Nice work!

dndx (approved):

PR #8529, opened 2022-03-10
tests(jwt) use rsa keys with 65537 as public exponent

fffonion:
Currently some jwt plugin test checks verification failure using RSA
key with public exponent replaced to an abitrary number. In
some cases, the OpenSSL implementation only accept 65537 as
public exponent and pkey failed to load.

This patch allows test to run on happily by always using RSA keys
with public exponent equal to 65537.

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issue reference

Fix #_[issue number]_


bungle:
> Currently some jwt plugin test checks verification failure using RSA key with public exponent replaced to an abitrary number. In some cases, the OpenSSL implementation only accept 65537 as public exponent and pkey failed to load.

What if there is someone using other than 65537, how does JWT plugin work, does it runtime crash or just fail normally? Do we need to change JWT plugin too?

spec/03-plugins/16-jwt/01-jwt_parser_spec.lua
```diff
@@ -151,34 +151,34 @@ describe("Plugin: jwt (parser)", function()
...
       local jwt = assert(jwt_parser:new(token))
       assert.True(jwt:verify_signature(fixtures.rs256_public_key))
-      assert.False(jwt:verify_signature(fixtures.rs256_public_key:gsub('QAB', 'zzz')))
+      assert.False(jwt:verify_signature(fixtures.rs384_public_key))
     end)
     it("using RS384", function()
       local token = jwt_parser.encode({sub = "foo"}, fixtures.rs384_private_key, "RS384")
       local jwt = assert(jwt_parser:new(token))
       assert.True(jwt:verify_signature(fixtures.rs384_public_key))
-      assert.False(jwt:verify_signature(fixtures.rs384_public_key:gsub("QAB", "zzz")))
```
fffonion, line 160:
```
# 65537 -> 0x10001
>>> base64.b64encode(b'\x01\x00\x01')
b'AQAB'
>>> binascii.hexlify(base64.b64decode(b'Azzz'))
b'033cf3'
```

fffonion:
@bungle i'm currently not sure if it's boringssl or boringssl+fips, normal openssl accepts any `e` happily, and that's why the test is running well. in boringssl it will fail here https://github.com/Kong/kong/blob/master/kong/plugins/jwt/jwt_parser.lua#L90, that will likely be a 500 since we did an assert. I can modify it to be `ok, err` pattern, though that might be a behaviour change, but a good one maybe?

bungle:
> @bungle i'm currently not sure if it's boringssl or boringssl+fips, normal openssl accepts any `e` happily, and that's why the test is running well. in boringssl it will fail here https://github.com/Kong/kong/blob/master/kong/plugins/jwt/jwt_parser.lua#L90, that will likely be a 500 since we did an assert. I can modify it to be `ok, err` pattern, though that might be a behaviour change, but a good one maybe?

Yes, looks like fips/boring possibly introduces a possible runtime crash that was not possible (or at least likely) with openssl? So it feel is is a behavior change in itself. Let's remember that failures in plugins especially authentication plugins with `config.anonymous` should not be fatal, especially if it can be user supplied data that causes the runtime error.

fffonion:
> looks like fips/boring possibly introduces a possible runtime crash that was not possible (or at least likely) with openssl

@bungle i think that's a general pattern we are doing for FIPS mode kong, you shouldn't be able to use such RSA key when creating a new plugin as we check it in schema validator, but if you migrate from a non-FIPS mode kong with existing plugins of such key, it will trigger this crash (result in a 500). i agree on the anonymous part, but let's keep this error obvious to end user, as it's slightly different than authentication failure case, more similar to say someone edit the database directly and put an invalid day.
the FIPS mode is not meant to be backward or forward compatible to normal Kong, as it's not really easy to work without bringing more behaviour changes. but we should have a comprehensive doc that explain the caveats when migrating to FIPS mode, and ideally a static check upon existing config datastore, i will make a ticket.

PR #8534, opened 2022-03-11
tests(helpers) use FFI based SSL server in tests

fffonion:
### Summary

Replace the SSL wrapper in spec.helpers.tcp_server to based on an FFI implementation, and choosed in
precedence of `luasec`. `luasec` doesn't compile happily with BoringSSL and there's too much work to
make it compatible, as BoringSSL has fairly freewill API choice that spans across multiple versions of OpenSSL.

However, both implementation has stable interface to SSL_*. This PR allows us to spin up a SSL server based on
luasocket without luasec.

### Full changelog

*  use FFI based SSL server in tests

### Issue reference

Fix #_[issue number]_

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -1074,7 +1074,11 @@ local function tcp_server(port, opts)
         end
 
         if opts.tls and handshake_done then
-          local ssl = require "ssl"
+          local _, ssl = pcall(require, "spec.helpers.ssl")
+          -- if not pok then
+          --   ssl = require "ssl"
+          -- end
+
           local params = {
             mode = "server",
             protocol = "any",
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -1074,7 +1074,7 @@ local function tcp_server(port, opts)
         end
 
         if opts.tls and handshake_done then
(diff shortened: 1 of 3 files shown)


spec/helpers.lua
```diff
@@ -1074,7 +1074,11 @@ local function tcp_server(port, opts)
         end
 
         if opts.tls and handshake_done then
-          local ssl = require "ssl"
+          local _, ssl = pcall(require, "spec.helpers.ssl")
+          -- if not pok then
+          --   ssl = require "ssl"
+          -- end
```
fffonion, line 1080:
```suggestion
```

spec/helpers.lua
```diff
@@ -1074,7 +1074,8 @@ local function tcp_server(port, opts)
         end
 
         if opts.tls and handshake_done then
-          local ssl = require "ssl"
+          local _, ssl = pcall(require, "spec.helpers.ssl")
```
fffonion, line 1077:
```suggestion
          local ssl = require "spec.helpers.ssl"
```

dndx (approved):

PR #8552, opened 2022-03-18
feat(core) remove deprecated go pluginserver support

Suika-Kong:
Go pluginserver is no longer supported in 3.0 or later versions.

Also, it seems that `kong.db.dao.plugins.go` is no longer used, and it is fully dependent on go plugin server.

Fix FT-2546


javierguerragiraldez (approved):
:partying_face:

PR #8553, opened 2022-03-18
fix(db) make dao:cache_key to fallback to primary_key if nothing from cache_key was found

bungle:
### Summary

This makes the `cache_key` function more robust. For example this code may be quite common:

```lua
local route = kong.db.routes:select_by_name("my-route")
-- {
--   id = ...,
--   name = "my-route",
--   ...
--   service = {
--     id = ...
--   }
-- }

local cache_key = kong.db.services:cache_key(route.service)
```

Now if `service` schema has `cache_key = { "name" }` (it does not, but just as an example) you can see that the `local cache_key` will then be same for all the `services` (no matter if they are pointing to different service by id), as the `route.service` is not expanded by default, and it only contains the primary key, in this case `id`.

The change in this commit is that it will now actually fallback to `primary_key = { "id" }` in case it cannot find anything by the `cache_key`. As it can be seen in code above, it is quite easy to make this mistake, and not see the mistake. User could fix their code by calling:

```lua
local cache_key = kong.db.services:cache_key(route.service.id)
```

instead of:

```lua
local cache_key = kong.db.services:cache_key(route.service)
```

But as this can potentially be dangerous if forgotten, I think it is worth to fallback to `primary_key` by default on such case.

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -1452,10 +1452,29 @@ function DAO:post_crud_event(operation, entity, old_entity, options)
 end
 
 
-function DAO:cache_key(key, arg2, arg3, arg4, arg5, ws_id)
+local function get_cache_key_value(name, key, fields)
+  local value = key[name]
+  if value == null or value == nil then
+    return
+  end
 
-  if self.schema.workspaceable then
-    ws_id = ws_id or workspaces.get_workspace_id()
+  if type(value) == "table" and fields[name].type == "foreign" then
+    value = value.id -- FIXME extract foreign key, do not assume `id`
+    if value == null or value == nil then
+      return
+    end
+  end
+
+  return tostring(value)
+end
+
+
+function DAO:cache_key(key, arg2, arg3, arg4, arg5, ws_id)
+  local schema = self.schema
+  local name = schema.name
+
+  if (ws_id == nil or ws_id == null) and schema.workspaceable then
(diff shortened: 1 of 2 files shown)


kong/db/dao/init.lua
```diff
@@ -1479,28 +1498,38 @@ function DAO:cache_key(key, arg2, arg3, arg4, arg5, ws_id)
...
+  local fields = schema.fields
+  local source = schema.cache_key
+  local use_pk = true
+  if source then
+    for j = 1, #source do
+      local value = get_cache_key_value(source[j], key, fields)
+      if value ~= nil then
+        use_pk = false
+      end
+      values[i] = value or ""
```
kikito, line 1519: Perhaps an extremely edge case, but if an entity has a cache key value which includes a `boolean`, the `false`s will get transformed into "" here. Also I am not sure how safe it is to remove the `tostring`
```suggestion
      if value == nil then
        value = ""
      else
        use_pk = false
      end
      values[i] = tostring(value)
```

kikito (changes requested):

kong/db/dao/init.lua
bungle: AFAIK this is not needed as the `value` comes from `get_cache_key_value` that can only return `nil` or a `string` (never a `boolean` for example).

kikito (approved):

PR #8559, opened 2022-03-21
fix(serverless) remove `config.functions` from schema

mayocream:
### Changelog

remove `config.functions` from serverless plugin schema

migrations for replace `conifig.functions` with `config.access` might needed.

FT-2557

### DIFF
--- a/kong/plugins/pre-function/_schema.lua
+++ b/kong/plugins/pre-function/_schema.lua
@@ -49,7 +49,6 @@ return function(plugin_name)
     },
     entity_checks = {
       { at_least_one_of = {
-        "config.functions",
         "config.certificate",
         "config.rewrite",
         "config.access",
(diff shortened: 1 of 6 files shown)


kong/plugins/pre-function/_schema.lua
```diff
@@ -60,10 +48,6 @@ return function(plugin_name)
       },
     },
     entity_checks = {
-      { mutually_exclusive_sets = {
-        set1 = { "config.functions" },
-        set2 = { "config.access" },
-      } },
       { at_least_one_of = {
         "config.functions",
```
fffonion, line 52:
```suggestion
```

mayocream: Thanks!

fffonion (approved):

PR #8570, opened 2022-03-22
add additional http status metrics to distinguish b/w Kong and upstream

dishatkr:
make target_label http metrics driven from config

update config variable name to kong_http_status_with_target

update default value of kong_http_status_with_target

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issue reference

Fix #_[issue number]_

### DIFF
--- a/kong/plugins/prometheus/schema.lua
+++ b/kong/plugins/prometheus/schema.lua
@@ -14,6 +14,7 @@ return {
         type = "record",
         fields = {
           { per_consumer = { type = "boolean", default = false }, },
+          { kong_http_status_with_target = { type = "boolean", default = false }, },
         },
         custom_validator = validate_shared_dict,
     }, },
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -106,6 +106,9 @@ local function init()
   metrics.consumer_status = prometheus:counter("http_consumer_status",
                                           "HTTP status codes for customer per service/route in Kong",
                                           {"service", "route", "code", "consumer"})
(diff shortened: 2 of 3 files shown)


fffonion:
Thanks for the contribution! @dishatkr Could you sign the CLA and also add a test for the new label?

brooklynrob:
> Thanks for the contribution! @dishatkr Could you sign the CLA and also add a test for the new label?

I'll help @dishatkr get this CLA done.

hbagdi:
Could you please explain what is the problem you are trying to solve here? I'm struggling a bit to understand the patch itself.

kong/plugins/prometheus/schema.lua
```diff
@@ -14,6 +14,7 @@ return {
         type = "record",
         fields = {
           { per_consumer = { type = "boolean", default = false }, },
+          { kong_http_status_with_target = { type = "boolean", default = false }, },
```
mikefero, line 17: This breaks version compatibility with older data planes; updates will need to be made to the [compatibility layer ](https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua)to handle these changes.

@aboudreault another version compatibility issues to watch out for.

hfukada:
Before too much effort is expended into this PR, internally we are working on upgrading the Prometheus plugin to be more cohesive with our API monitoring story in 3.0

One of the outcomes of the upgrade would be adding a label called `origin` to the `request_count` metric that specifies if a status code was returned by `upstream` or `kong`. Would this satisfy your needs?

I see this PR also includes adding the label `target` for the upstream `ip:port` pair, but we advise against this since it will stress both Prometheus and Kong itself, since it has the likely outcome to explode cardinality.

PR #8572, opened 2022-03-22
docs(api) document header regex in autodocs admin api

kikito:
### Summary

Document missing feature for Kong 2.8.x

### Full changelog

* Document regex header matching, implemented in #6079

### DIFF
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -988,8 +988,8 @@ return {
             match if present in the request.
             The `Host` header cannot be used with this attribute: hosts should be specified
             using the `hosts` attribute.
-            Since version 2.8.0, when `headers` contains only one value and that value starts with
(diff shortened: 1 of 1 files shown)


autodoc/admin-api/data/admin-api.lua
```diff
@@ -988,6 +988,8 @@ return {
             match if present in the request.
             The `Host` header cannot be used with this attribute: hosts should be specified
             using the `hosts` attribute.
+            Since version 2.8.0, when `headers` contains only one value and that value starts with
+            the special prefix `~*`, the value will be interpreted as a regular expression.
```
lena-larionova, line 992:
```suggestion
            When `headers` contains only one value and that value starts with
            the special prefix `~*`, the value is interpreted as a regular expression.
```
You don't need to add a "starting with 2.8.0", since the docs themselves are versioned.

guanlan (approved):

PR #8574, opened 2022-03-23
fix(serverless) migrations from 2.8 to 3.0

mayocream:
### Changelog

Add migrations for the serverless plugin to upgrade from 2.8 to 3.0.

- #8559

FT-2557


kikito (approved):
I have reviewed this and it looks ok to me. However we need to write tests for these migrations as well, on a separate PR.

mayocream:
@kikito Thanks!

PR #8575, opened 2022-03-23
fix(aws-lambda) migrations from 2.8 to 3.0

mayocream:
### Changelog

Add migrations for aws-lambda to upgrade from 2.8 to 3.0

- #8566


bungle:
@mayocream,

The `schema:process_auto_fields` automatically removes the fields that are not in schema. Why do we need this migration?

mayocream:
> @mayocream,
> 
> The `schema:process_auto_fields` automatically removes the fields that are not in schema. Why do we need this migration?

Thanks for noticing me! I didn't noticed the `schema:process_auto_fields` function. 🥲

PR #8583, opened 2022-03-23
fix(core) use admin_error_log for cluster listener

flrgh:
This `server {}` block previously had no `error_log` directive, so it was inheriting the `proxy_error_log` settings from the `http {}` scope. Logging to the same place as `admin_error_log` is more sensible. We might want to consider introducing a new `cluster_error_log` config option at some point to give users more flexibility (using `admin_error_log` as a fallback), but to me this seems fine for now.


bungle (approved):

flrgh:
Threw in a late commit updating the integration test template to match

PR #8585, opened 2022-03-24
chore(pdk) remove pdk version

chronolaw:
### Summary

PDK's version is not used ever, we remove it to get more clean code.

### Full changelog

* remove pdk_version/pdk_major_version
* remove param in pdk.new
* remove param in init_pdk
* clean related api calls.


thibaultcha:
This was written for forward-compatibility, it is normal that it is not being used at the moment. Removing it may render future breaking changes in the PDK harder to contain.

kong/cmd/config.lua
```diff
@@ -80,7 +80,7 @@ local function execute(args)
   package.path = conf.lua_package_path .. ";" .. package.path
 
   _G.kong = kong_global.new()
-  kong_global.init_pdk(_G.kong, conf, nil) -- nil: latest PDK
+  kong_global.init_pdk(_G.kong, conf) -- latest PDK
```
kikito, line 83: I would remove the comment as well, it is also related with plugin versioning.
```suggestion
  kong_global.init_pdk(_G.kong, conf) 
```

kong/cmd/migrations.lua
```diff
@@ -96,7 +96,7 @@ local function execute(args)
   assert(prefix_handler.prepare_prefix(conf, args.nginx_conf, true))
 
   _G.kong = kong_global.new()
-  kong_global.init_pdk(_G.kong, conf, nil) -- nil: latest PDK
+  kong_global.init_pdk(_G.kong, conf) -- latest PDK
```
kikito, line 99:
```suggestion
  kong_global.init_pdk(_G.kong, conf)
```

kong/cmd/start.lua
```diff
@@ -27,7 +27,7 @@ local function execute(args)
          "Kong is already running in " .. conf.prefix)
 
   _G.kong = kong_global.new()
-  kong_global.init_pdk(_G.kong, conf, nil) -- nil: latest PDK
+  kong_global.init_pdk(_G.kong, conf) -- latest PDK
```
kikito, line 30:
```suggestion
  kong_global.init_pdk(_G.kong, conf)

File: kong/cmd/vault.lua [hunk 4]
```diff
@@ -30,7 +30,7 @@ local function init_db(args)
   package.path = conf.lua_package_path .. ";" .. package.path
 
   _G.kong = kong_global.new()
-  kong_global.init_pdk(_G.kong, conf, nil) -- nil: latest PDK
+  kong_global.init_pdk(_G.kong, conf) -- latest PDK
```
kikito, line 33:
```suggestion
  kong_global.init_pdk(_G.kong, conf)
```

kong/pdk/init.lua
```diff
@@ -37,25 +37,6 @@
 -- end
 
 
----
```
kikito, line 40: The top-comment of this file needs to be modified - after this change, the PDK is no longer "Semantically Versioned".

kikito (changes requested):
Please also include an entry in the Breaking Changes section of the Changelog mentioning that the PDK is no longer versioned

kikito:
> This was written for forward-compatibility, it is normal that it is not being used at the moment. Removing it may render future breaking changes in the PDK harder to contain.

The main reason why this change was proposed is that currently every single time the PDK is initialized in Kong, the version number param is `nil` - you can see it on this PR. That makes the versioning code dead code. The only way someone is going to be able to benefit from this feature is if they maintain their own fork of Kong and actually pass a value to that parameter. I could be missing something, of course.

On a more philosophical level, the PDK versioning code could be seen as worse than dead - that is, as code that is making promises that later we don't deliver. This is not a position we want to be in, for multiple reasons.

chronolaw:
> Please also include an entry in the Breaking Changes section of the Changelog mentioning that the PDK is no longer versioned

Here is it: https://github.com/Kong/kong/pull/8606

kikito:
> Let's say a method must be removed or its signature or behavior changed.

That can only be done between major versions now. On minor releases we can add new methods, or new parameters to existing methods (if their default values don't change behavior). The exception being when the previous behavior was buggy, in which case the bug will be fixed in a way that we try to affect the least people, but the plugins that depended explicitly on the buggy behavior will break.

> Untrue. nil specifically requests the latest version of the PDK, as explicitly documented in the comments.

Apologies, I must have misread then.

> Take the breaking change and bump Kong's semver version?

That is the new strategy.

> allowing users to upgrade major Kong versions without having to upgrade their plugins right away

Their plugins are guaranteed to work between patches and minor updates. We only do major releases once per year. Even in major releases, we keep the list of changes low.

>  It seems to me we are again setting ourselves up for failure by closing doors we know we would rather have open

I understand why you say that, and I understand the intention behind the PDK versioning. If we look at the PDK individually, it makes sense.

The problem is that we have to consider Kong as the sum of its parts; we have applied the versioning principle to too many Kong subsystems, in different ways - bundled plugins, CP/DPs, declarative config versioning. And then the ensemble got too complex. People don't understand what is compatible with what any more. There's an internal Jira epic tracking this problem (FT-2379), and it didn't even include PDK versioning (I have just included a link to this ticket now). In my opinion the only way to keep the whole thing sane and understandable is to simplify.

> I only wanted to voice my concerns

Your feedback is always welcome. Thank you for raising them.

kikito (approved):

bungle:
My few cents on this as I probably proposed this in first place.

1. As mentioned we have way too many versions: openssl, nginx, openresty, kong (ce and ee), luarocks, pdk, external language pdk and their limitations, plugin. It is just plain confusing.
2. PDK versioning that we have always involves custom plugin writers to change their code, e.g. request older version of PDK (I tried to find documentation for that, but could not, so I don't even know how to do that), if something breaks, but not only that, we have probably screwed this by ourselves already, and users will just always get the same version. We will screw it again and again. It just doesn't work. But while user is requesting certain version of PDK (aka change their code - everyone wants to use latest by default), they could just change their code to work with the latest version of PDK. We do that to everything we ship. Also I have never seen anywhere code that reads PDK version and acts on the results. Our documentation on that is beyond horrible too (you can read the numbers, but there is no documentation what numbers mean). Like what kind of if statement would I need to write if I want my plugin to run on old and new version? Would it just be easier to release separate versions of plugin to different kong versions rather than trying to keep codebase somewhat functioning between different PDK versions. Some things in PDK are versioned differently, like worker_events, cluster_events, cache, ... they already follow just Kong version and some of them come from external rocks.

This thing has always contained just 1:
https://github.com/Kong/kong/blame/master/kong/pdk/init.lua#L216-L239

Partly because we have screwed up (and it is damn easy to screw this up), and partly because we have not introduced any major changes to it (or if we have, we have taken calculated risks, I guess sometimes non-calculated too).

This is the reality. PDK versioning is:

1. hard to understand
2. badly documented
3. major changes rarely happen (if ever)
4. only needed with custom plugins (everything else, we cover)

Yes, the developers can be pointed on this for the most part of it, but all that said, I don't see us not screwing up this in future, and thus we need simpler version scheme. That effectively means that Kong PDK version == Kong version, and we follow semantic versioning of Kong with PDK too (e.g. things should only break on majors, except for security fixes or plain bugs).

To be honest, as now all the plugins are also bundled with Kong, it may not make sense to version plugins either. We have forgot to bump plugin versions many many times already. So effectively sometimes plugin versions lie. Another screw up. We need some automation here or merge blocker tests, if we really want to have this consistent.

PR #8587, opened 2022-03-24
fix(luarocks): luarocks install requires luasec to be installed

hutchic:
per the kong-build-tools e2e tests `luarocks install` requires luasec


bungle:
> per the kong-build-tools e2e tests `luarocks install` requires luasec

Related PR:
https://github.com/Kong/kong/pull/8534

@fffonion what do you think?

hutchic:
@bungle possibly but until we determine that I propose we revert and then collect our thougts on a path forward

bungle (approved):

PR #8593, opened 2022-03-24
chore(deps) bump grpcurl from 1.8.5 to 1.8.6

bungle:
### Summary

Bumps the development dependency.

#### Changes in Command-line tool

- Some bugs have been addressed in the library used to parse proto source files.
  Previously grpcurl would accept proto source files that could not actually be
  compiled with protoc. The converse could also happen: grpcurl could reject some
  proto source files that could successfully be compiled with protoc. More details
  can be found in the release notes for the changes to the protoparse library,
  versions v1.10.2 and v1.10.3.
- Some implementations of the server reflection service have been observed to
  return multiple (even superfluous) file descriptors, in response to requests made
  by grpcurl. These extra files, if not returned in a particular order, would cause
  grpcurl to report an error that the service or method to be invoked could be not
  be resolved. The reflection client in grpcurl is now more robust to this condition
  and can handle responses with file descriptors in any order, so it should be
  interoperable with a larger variety of servers.
- When a request message includes a field of type google.protobuf.Value and
  a value for that field that was a JSON array, grpcurl would incorrectly interpret
  the JSON array as if it were a single atomic value, the last value that was in the
  array. This has been fixed.
- When a response message includes non-printable characters or code points outside
  the ASCII 7-bit range in the name of a field in a JSON object, it could be
  improperly encoded with escape characters that are not valid JSON. Standard
  tools/libraries could then fail to parse the JSON output from grpcurl. This has
  been fixed.


dndx:
@bungle Tests seems to be failing after the bump, could you investigate?

ADD-SP:
@bungle @dndx 

The output of the new `grpcurl` to `stdout` has changed, so our test fails. Maybe we need to change the tests?

My opinion is that even if the status code is `4xx`, the response header should still be output. Is there a flag to hold these outputs? I have not found it. This may be a tiny bug in the `grpcurl-1.8.6`. 

***

For `grpcurl-1.8.5`, it outputs the response header and the keyword `Code: ResourceExhausted`

```
$ bin/grpcurl -plaintext  -proto ./spec/fixtures/grpc/hello.proto -v  0.0.0.0:9002 hello.HelloService.SayHello
Resolved method descriptor:
rpc SayHello ( .hello.HelloRequest ) returns ( .hello.HelloResponse );

Request metadata to send:
(empty)

Response headers received:
(empty)

Response trailers received:
content-length: 0
content-type: application/grpc
date: Fri, 01 Apr 2022 04:27:13 GMT
ratelimit-limit: 6
ratelimit-remaining: 0
ratelimit-reset: 47
retry-after: 47
server: kong/2.8.0
x-kong-response-latency: 0
x-ratelimit-limit-minute: 6
x-ratelimit-remaining-minute: 0
Sent 0 requests and received 0 responses
ERROR:
  Code: ResourceExhausted
  Message: ResourceExhausted
```

For `grpcurl-1.8.6`, neither of these things are being output.

```
$ bin/grpcurl -plaintext  -proto ./spec/fixtures/grpc/hello.proto -vv  0.0.0.0:9002 hello.HelloService.SayHello

Resolved method descriptor:
rpc SayHello ( .hello.HelloRequest ) returns ( .hello.HelloResponse );

Request metadata to send:
(empty)

Response headers received:
(empty)

Response trailers received:
(empty)
Sent 0 requests and received 0 responses
ERROR:
  Code: Unavailable
  Message: unexpected HTTP status code received from server: 429 (Too Many Requests)
```

PR #8599, opened 2022-03-28
fix(pdk) kong.response.get_source() to return error on plugin throws runtime exception

ms2008:
### Summary

If plugin that throws runtime exception on access phase, `kong.response.get_source()` should returned `error` instead of `exit`.

FTI-3200


dndx (commented):
Looks reasonable, can we add a test case using `severless-plugin`?

andreleoni:
This change context should be by plugin, instead of all application initializer?

In my opinion, if a error was received, the application should exit, instead of only log an error, should be dangerous only log an error...

🤔

ms2008:
> This change context should be by plugin, instead of all application initializer?
> 
> In my opinion, if a error was received, the application should exit, instead of only log an error, should be dangerous only log an error...
> 
> 🤔

In access phase we need to run through the whole plugin iterator to get list of plugins for the request. You cannot exit it while looping. We need to collect the plugins even in case of an error.

spec/02-integration/05-proxy/04-plugins_triggering_spec.lua
```diff
@@ -1022,6 +1053,19 @@ for _, strategy in helpers.each_strategy() do
         assert.res_status(504, res) -- Gateway Timeout
         assert.equal("timeout", res.headers["Log-Plugin-Service-Matched"])
       end)
+
+      it("sees pdk get_source() error on plugin runtime exception aborts with a 500", function()
```
dndx, line 1057:
```suggestion
      it("kong.response.get_source() returns \"error\" if plugin runtime exception occurs, FTI-3200", function()
```

dndx (approved):

PR #8602, opened 2022-03-29
feat(config) default `lua_ssl_trusted_certificate` to `system`

ADD-SP:
### Summary

Change the default of `lua_ssl_trusted_certificate` to `system`.

### Full changelog

* update `kong/templates/kong_defaults.lua` to change the default value of `lua_ssl_trusted_certificate` to `system`.
* update `spec/helpers.lua` to add a register a new assert function `has_value(table, value)`.
* update `spec/01-unit/03-conf_loader_spec.lua` and `spec/01-unit/04-prefix_handler_spec.lua` to adjust this changes.
* update changelog.


spec/helpers.lua
```diff
@@ -1910,6 +1910,47 @@ luassert:register("assertion", "cn", assert_cn,
...
+]])
+say:set("assertion.has_value.positive", [[
+Expected table to not have the given value.
+Passed in
+%s
+Expected value
+%s
+]])
+
+luassert:register("assertion", "has_value", assert_has_value,
```
dndx, line 1949:
```suggestion
luassert:register("assertion", "contains", assert_has_value,
```

Might be a better name IMO.

ADD-SP: https://github.com/Kong/kong/blob/d6202c22809a0a4f138cb9d58a4631f1b3de6ad5/spec/helpers.lua#L1467-L1502

There is a function with the same name, but it is only used for strings (regex). `has_value` is expected to be used for other types, such as number.

CHANGELOG.md
```diff
@@ -75,6 +75,13 @@
 - Bumped inspect from 3.1.2 to 3.1.3
   [#8589](https://github.com/Kong/kong/pull/8589)
 
+
+### Changes
+##### Configuration
+
+- Change the default of `lua_ssl_trusted_certificate` to `system`.
+  [#8602](https://github.com/Kong/kong/pull/8602)
```
kikito, line 83: Include instructions about how to avoid problems with the change:

```suggestion
- Change the default of `lua_ssl_trusted_certificate` to `system`.
  [#8602](https://github.com/Kong/kong/pull/8602). If you are upgrading from 2.x and want this variable to keep
  working as before, please manually set it to `NONE` before upgrading. 
```

I am not 100% sure that the instructions are correct, please double check.

kikito (dismissed):

spec/helpers.lua
dndx: In that case, can we make `contains` compatible with different value types?

bungle (changes requested):
@ADD-SP, please update this too:
https://github.com/Kong/kong/blob/master/kong.conf.default#L1382

To reflect that `system` is default.

bungle (approved):

CHANGELOG.md
```diff
@@ -91,6 +91,16 @@
...
 
+
+### Breaking Changes
+
+##### Configuration
+
+- Change the default of `lua_ssl_trusted_certificate` to `system`
+  [#8602](https://github.com/Kong/kong/pull/8602). If you are upgrading from 2.x and want this variable to keep
+  working as before, please manually set it to empty
+  (`lua_ssl_trusted_certificate = [nothing in here]`) before upgrading.
```
dndx, line 102:
```suggestion
  [#8602](https://github.com/Kong/kong/pull/8602).
```

Is probably enough.

ADD-SP: Where do we usually mention these things?

CHANGELOG.md
```diff
@@ -91,6 +91,14 @@
 - Bumped inspect from 3.1.2 to 3.1.3
   [#8589](https://github.com/Kong/kong/pull/8589)
 
+
+### Breaking Changes
+
+##### Configuration
+
+- Change the default of `lua_ssl_trusted_certificate` to `system`
+  [#8602](https://github.com/Kong/kong/pull/8602).
```
dndx, line 100:
```suggestion
  [#8602](https://github.com/Kong/kong/pull/8602) to automatically load trusted CA list from system CA store.
```

dndx (approved):
LGTM, let's wait for the test to pass before merging.

PR #8603, opened 2022-03-29
fix(core) ctx dropped in log phase

### LINKED ISSUE #8598: [bug] plugin log phase handler causes error in PDK
According to the [docs](https://docs.konghq.com/gateway/2.7.x/plugin-development/custom-logic/) it should be possible to include method `log(self, kong)` in the custom plugin object. When I try it however I get
```
[error] 27#0: *1032 lua entry thread aborted: runtime error: /usr/local/share/lua/5.1/kong/pdk/private/phases.lua:79: no phase in kong.ctx.core.phase, (need one of rewrite, access, response, cluster_listener, header_filter, body_filter)
stack traceback:
coroutine 0:
	[C]: in function 'error'
	/usr/local/share/lua/5.1/kong/pdk/private/phases.lua:79: in function 'check_phase'
	/usr/local/share/lua/5.1/kong/pdk/request.lua:562: in function 'call_pdk_method'
	...cal/share/lua/5.1/kong/runloop/plugin_servers/mp_rpc.lua:286: in function 'bridge_loop'
	...cal/share/lua/5.1/kong/runloop/plugin_servers/mp_rpc.lua:305: in function 'handle_event'
	...local/share/lua/5.1/kong/runloop/plugin_servers/init.lua:236: in function <...local/share/lua/5.1/kong/runloop/plugin_servers/init.lua:232>, context: ngx.timer, client: 172.17.0.1, server: 0.0.0.0:8000
```
does it mean that Python plugin server for Kong does not support `log` phase?

Suika-Kong:
log phase does recover ngx.ctx for some of functions(with wrapper function), however for those functon not wrapped, they will see an empty ngx.ctx, and cause unreadable error message for check_phase

Fix #8598

### DIFF
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -35,27 +35,19 @@ local exposed_api = {
   ["kong.nginx.get_tls1_version_str"] = ngx_ssl.get_tls1_version_str,
 
   ["kong.nginx.get_ctx"] = function(k)
-    local saved = save_for_later[coroutine_running()]
-    local ngx_ctx = saved and saved.ngx_ctx or ngx.ctx
-    return ngx_ctx[k]
+    return ngx.ctx
   end,
 
   ["kong.nginx.set_ctx"] = function(k, v)
-    local saved = save_for_later[coroutine_running()]
-    local ngx_ctx = saved and saved.ngx_ctx or ngx.ctx
-    ngx_ctx[k] = v
+    ngx.ctx[k] = v
   end,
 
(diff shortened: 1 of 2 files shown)


Suika-Kong:
Arguably this way to recover `ngx.ctx` will cause problems someday so here is an ad hoc solution: #8604.

Suika-Kong:
It does fail some tests...

Suika-Kong:
> It does fail some tests...

Or maybe we can list functions that need `ngx.ctx` recovered, instead of wrapping them 1 by 1. It looks more manageable to me.

PR #8604, opened 2022-03-29
fix(core) ctx dropped in log phase

### LINKED ISSUE #8598: [bug] plugin log phase handler causes error in PDK

Suika-Kong:
Log phase does recover ngx.ctx for some of functions(with wrapper function), however for those functions not wrapped, they will see an empty ngx.ctx, and cause an unreadable error message for `check_phase`.
This is an ad hoc fix for `check_phase`.

Fix #8598

### DIFF
--- a/kong/runloop/plugin_servers/mp_rpc.lua
+++ b/kong/runloop/plugin_servers/mp_rpc.lua
@@ -122,6 +122,7 @@ local function call_pdk_method(cmd, args)
 
   local saved = Rpc.save_for_later[coroutine.running()]
   if saved and saved.plugin_name then
+    ngx.ctx.KONG_PHASE = saved.ngx_ctx.KONG_PHASE
     kong_global.set_namespaced_log(kong, saved.plugin_name)
   end
 


Suika-Kong:
For people who want to review this PR: 
We handle log phase RPC in a timer because we are using cosocket for RPC that is not available in the log phase.
We did save the `ngx.ctx` object but if we recover it unconditionally it causes new problems(and `ngx.ctx` is not stable).

ADD-SP:
Should we keep the context for all phases? This way seems a bit unreliable. Is there any other way that we can share data for all phases, except shared memory and database?

Suika-Kong:
@fffonion suggests recovering `ngx.ctx` in the timer instead of in mp\_rpc.lua, as this should be transparent to the rpc implementation, and apply both for mp and pb.

kong/runloop/plugin_servers/init.lua
```diff
@@ -254,6 +254,11 @@ local function build_phases(plugin)
           local co = coroutine_running()
           save_for_later[co] = saved
 
+          -- recover KONG_PHASE so check phase works properly
+          -- for functions not supported by log phase
+          if ngx.ctx ~= nil then
```
fffonion, line 259:
```suggestion
          if ngx.ctx then
```

fffonion (approved):
minor comment, otherwise look good to me. feel free to merge after addressed (and squashed)

Suika-Kong:
Rebased the PR. Hope that will fix the test.

PR #8607, opened 2022-03-29
chore(proxy-cache) do not copy response data to request's ctx

locao:
### Summary

`proxy_cache_hit` contains the response data to be shared with the logging plugins. Originally it was added to the generic Nginx request's ctx, then later it was changed to use Kong's ctx but the previous one was kept as a backward compatibility. For the next major version this compatibility will be removed.

### Full changelog

* Removed `proxy_cache_hit` backward compatibility.

FT-2559

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,13 @@
   scripts that depend on it being `POST`, these scripts will need to be updated
   when updating to Kong 3.0.
 
+#### Plugins
+
+- The proxy-cache plugin does not share the response data in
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -73,6 +73,13 @@
   scripts that depend on it being `POST`, these scripts will need to be updated
   when updating to Kong 3.0.
 
+#### Plugins
+
+- The proxy-cache plugin does not share the response data in
```
dndx, line 78:
```suggestion
- The proxy-cache plugin does not store the response data in
```

dndx (approved):
Mi

PR #8612, opened 2022-03-30
docs(core) deprecate go pluginserver

Suika-Kong:
As in the title.


CHANGELOG.md
```diff
@@ -73,6 +73,11 @@
   scripts that depend on it being `POST`, these scripts will need to be updated
   when updating to Kong 3.0.
 
+### Deprecations
+
+- The external [go-pluginserver](https://github.com/Kong/go-pluginserver) project
+is no longer suppored.
```
kikito, line 79: I believe the Go Pluginserver was already deprecated (please confirm), shouldn't this be in the Breaking Changes section instead? I would also add a not about how users can avoid the problem when updating:

```suggestion
### Breaking Changes
...
- The external [go-pluginserver](https://github.com/Kong/go-pluginserver) project
is no longer supported. Please move external plugin servers to [...] (name and/or docs link) 
before upgrading.
```

If it was supported and we're starting a deprecation in 3.0, then the message should say that we're deprecating it, not that it is no longer supported:

```suggestion
### Deprecations

- The external [go-pluginserver](https://github.com/Kong/go-pluginserver) project
is deprecated. It is recomented that you migrate your go plugins to [...]
```

Suika-Kong: This PR is to drop out support for the go plugin server(we have already marked it as deprecated). And we have already removed the support in the master branch (and for 3.0 and later).
Instructions for migration are already included in the earlier changelog so I am not sure should we repeat it here.

Suika-Kong: And this line is in the section of unreleased.

kikito (approved):

PR #8627, opened 2022-04-02
fix(http-log) url query parameter missing

VicYP:
This commit https://github.com/Kong/kong/commit/2a608480d3f193b87a784666d99efbadb15be1aa is used to fix http log url query parameter missing issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,8 @@
 [#8558](https://github.com/Kong/kong/pull/8558)
 - Remove deprecated `blacklist`/`whitelist` config fields from bot-detection, ip-restriction and ACL plugins.
 [#8560](https://github.com/Kong/kong/pull/8560)
+- **http-log**: Record the query parameter.
+[#8627](https://github.com/Kong/kong/pull/8627)
 
 #### Clustering
 
--- a/kong/plugins/http-log/handler.lua
+++ b/kong/plugins/http-log/handler.lua
@@ -93,6 +93,10 @@ local function send_payload(self, conf, payload)
   params_cache.keepalive_timeout = keepalive
 
   local url = fmt("%s://%s:%d%s", parsed_url.scheme, parsed_url.host, parsed_url.port, parsed_url.path)
+  
+  if parsed_url.query then
+    url = url .. "? ".. parsed_url.query
+  end
 
   -- note: `httpc:request` makes a deep copy of `params_cache`, so it will be
   -- fine to reuse the table here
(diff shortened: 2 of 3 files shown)


fffonion (approved):

bungle:
@VicYP, is it possible to add a test case for this? Should be quite straightforward, if I understood correctly?

CHANGELOG.md
```diff
@@ -123,6 +123,8 @@
 [#8558](https://github.com/Kong/kong/pull/8558)
 - Remove deprecated `blacklist`/`whitelist` config fields from bot-detection, ip-restriction and ACL plugins.
 [#8560](https://github.com/Kong/kong/pull/8560)
+- **http-log**: Record the query parameter.
+[#8627](https://github.com/Kong/kong/pull/8627)
```
dndx, line 127:
```suggestion
- **http-log**: Include the query parameter when making logging requests to `config.http_endpoint`.
[#8627](https://github.com/Kong/kong/pull/8627)
```

spec/03-plugins/03-http-log/01-log_spec.lua
```diff
@@ -240,7 +240,7 @@ for _, strategy in helpers.each_strategy() do
     it("logs to HTTP", function()
       local res = assert(proxy_client:send({
         method = "GET",
-        path = "/status/200",
+        path = "/status/200?test0=test0&test1=test1",
```
dndx, line 243: This is not testing this change correctly. We need to add a plugin config like ones in https://github.com/Kong/kong/pull/8627/files#diff-50dcfdce82665f85d2027363af6956fc788e5b8f6c6b8de907e9bd82e8871501R202 and add the query parameter in the `config.http_endpoint` and make sure it works.

dndx (changes requested):

kikito:
I was slightly concerned about whether this is a breaking change or a bug fix, so I went out and checked [rfc1738](https://datatracker.ietf.org/doc/html/rfc1738#section-3.3), to confirm that query params are indeed considered part of the definition of a URL. So this is a bugfix, if someone is relying on query params being ignored it is ok for us to "break" things for them.

dndx:
@kikito This is less concerning to me, even in the rare case user relied on query parameters being ignored, having unneeded query parameters should not affect anything anyway as long as they are unused.

kong/plugins/http-log/handler.lua
```diff
@@ -93,6 +93,10 @@ local function send_payload(self, conf, payload)
   params_cache.keepalive_timeout = keepalive
 
   local url = fmt("%s://%s:%d%s", parsed_url.scheme, parsed_url.host, parsed_url.port, parsed_url.path)
+  
+  if parsed_url.query then
+    url = url .. "? ".. parsed_url.query
```
samugi, line 98: I think the space after the question mark was meant to be outside of the quotes:
```suggestion
    url = url .. "?" .. parsed_url.query
```

hbagdi:
@VicYP Can you please drive this to completion?

hbagdi:
ping @VicYP

hbagdi:
ping @VicYP

hbagdi:

PR #8639, opened 2022-04-05
perf(clustering) conditional rebuilding of router, plugins iterator and balancer on dp (2.8)

aboudreault:
backport of #8519 

### Summary

Implements conditional rebuilding of `router`, `plugins iterator` and `balancer` on
data planes. This means that DPs will not rebuild router if there were no changes in
routes or services. Similarly, the plugins iterator will not be rebuild if there were
no changes to plugins, and finally balancer in not reinitialized if there are no
changes to upstreams or targets.

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -185,7 +185,7 @@ local function get_removed_fields(dp_version_number)
           unknown_fields[plugin] = {}
         end
         for _, k in ipairs(fields) do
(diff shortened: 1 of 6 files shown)


kikito (approved):

zffocussss:
Hi,if one target is updated,post a new target,then delete a old target,the whole upstreams and routes will be put.is it that expected behavior in kong?

PR #8646, opened 2022-04-06
Release: 2.8.1

tyler-ball:


kong-admin-api.yml
```diff
@@ -1,189 +1,125 @@
+servers:
```
tyler-ball, line 1: I'm worried this seems like a big diff

kikito: This is caused because libyaml does not support "sorting objects before exporting". It iterates over Lua tables with `pairs`/`next`. Our "root" is a yaml object, not an array, so this means that even at the top-level things are not guaranteed to remain on the same order after every run of the command.

CHANGELOG.md
```diff
@@ -78,6 +79,12 @@
 - Only reschedule router and plugin iterator timers after finishing previous
   execution, avoiding unnecessary concurrent executions.
   [#8634](https://github.com/Kong/kong/pull/8634)
+- Implements conditional rebuilding of router, plugins iterator and balancer on
+  data planes. This means that DPs will not rebuild router if there were no
+  changes in routes or services. Similarly, the plugins iterator will not be
+  rebuild if there were no changes to plugins, and finally balancer in not
```
gszr, line 85:
```suggestion
  rebuilt if there were no changes to plugins, and, finally, the balancer will not be
```

gszr (approved):
Only one minor comment, feel free to adjust after merge if that is more convenient.

kikito:
Let's cherry-pick https://github.com/Kong/kong/pull/8572 into this as well.

I will add @gszr 's suggestion, cherry pick 8572 and reorder commits so the release one is at the end.

PR #8650, opened 2022-04-07
fix(doc) duplicate and incorrect level title

Suika-Kong:
As the title


dndx (dismissed):

dndx:
@ADD-SP Seems that everything else uses `####` under `###`, therefore we should probably use `####` here as well.

ADD-SP (approved):

PR #8652, opened 2022-04-07
fix(plugins): change log level from error to info when restarting external plugin instance

ealogar:
### Summary

When the external plugin instances (go, python, ...) loses the instances_id in a event handler, the instance is restarted to complete the plugin call but a weird error appear:
```
2021/08/20 10:23:59 [**error**] 30#0: *51336 [kong] mp_rpc.lua:308 [jwe] no plugin instance 0, client: 10.240.0.178, server: kong, request: "POST /example HTTP/1.1", host: "example.es"
```
We change the log level to info as this is not a real problem as the plugin is restarted lately in code (considering the message string as well)


kong/runloop/plugin_servers/mp_rpc.lua
```diff
@@ -329,7 +329,7 @@ function Rpc:handle_event(plugin_name, conf, phase)
   local _, err = bridge_loop(self, instance_id, phase)
 
   if err then
-    kong.log.err(err)
+    kong.log.info(err)
 
     if string.match(err:lower(), "no plugin instance") then
```
mayocream, line 334: The original `err` could be a "real" error; I prefer to change the log level to WARN inside the `if` condition.

```suggestion
    if string.match(err:lower(), "no plugin instance") then
      kong.log.warn(err)
```

Then continue to print an error level msg if it won't match.

fffonion (approved):

fffonion:
Thank you @ealogar !

PR #8655, opened 2022-04-08
hotfix(conf) make lua_ssl_trusted_certificate fail softly on launch

kikito:
When `system` was opt-in, it made sense to "fail noisily" if the current
system didn't have certs in any of the usual places and the user had
opted in to system.

Now that we look for them by default, the noisy behavior is incorrect,
because it makes kong unable to start in places with no system certs.
Instead, the abstence of default certificates gets *noted* in the logs,
but Kong can continue without them.

Related with #8602

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -903,25 +903,28 @@ local function check_and_infer(conf, opts)
   if conf.lua_ssl_trusted_certificate then
     local new_paths = {}
 
-    for i, path in ipairs(conf.lua_ssl_trusted_certificate) do
+    for _, path in ipairs(conf.lua_ssl_trusted_certificate) do
       if path == "system" then
         local system_path, err = utils.get_system_trusted_certs_filepath()
         if system_path then
           path = system_path
 
         else
-          errors[#errors + 1] =
-            "lua_ssl_trusted_certificate: unable to locate system bundle - " ..
-            err
+
+          log.info("lua_ssl_trusted_certificate: unable to locate system bundle: " ..
+                     err ..
+                     ". Please set lua_ssl_trusted_certificate to a path with certificates" ..
+                     " in order to remove this message")
         end
       end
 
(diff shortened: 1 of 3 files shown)


gszr (commented):
Could we add a test case?

flrgh:
_Discussed elsewhere, but adding here for visibility's-sake:_

IMO it's a little more nuanced than "now that it's the default, we shouldn't throw a fatal error." If a user has explicitly set `lua_ssl_trusted_certificate = system` I think we _should_ fail hard and fast. If the user has left the setting empty/undefined, we should log and continue on.

This requires the ability to distinguish between a value that is explicitly set vs a value that is inherited from defaults, which might not be a capability that exists in the conf loader right now but seems worth exploring.

kikito:
> If a user has explicitly set system I think we should fail hard and fast.
Not possible to make that distinction any more, the new default is system  in 3.0 (too many support issues open because of the old empty default)

gszr (approved):

kong/conf_loader/init.lua
```diff
@@ -903,25 +903,28 @@ local function check_and_infer(conf, opts)
...
         local system_path, err = utils.get_system_trusted_certs_filepath()
         if system_path then
           path = system_path
 
         else
-          errors[#errors + 1] =
-            "lua_ssl_trusted_certificate: unable to locate system bundle - " ..
-            err
+
+          log.info("lua_ssl_trusted_certificate: unable to locate system bundle: " ..
```
bungle, line 914: warn?

dndx (approved):
From a security perspective failing automatically is not an issue. It is failsafe as nothing gets loaded would not create security loopholes, worst that can happen is certificate fails to verify which is also "failsafe".

kong/conf_loader/init.lua
kikito: No, otherwise we will get warnings in our own images, some of which don't have ssl certs.

PR #8657, opened 2022-04-08
feat(schema) restrict host and port to avoid malicious configurations 

locao:
### Summary

It's pretty rare that someone actually needs to set up an API gateway to proxy to some hosts and ports, yet these can be used by attackers to attack internal networks. This change adds a list of hosts and ports that can't be ever used, unless Kong is explicit told so.

### Full changelog

* Added new config properties `proxy_restricted_hosts` and  `proxy_restricted_ports`.
* Validate targets and services entities schema using new config properties.
* Added tests related to configuring new properties and related to creating targets and services.

FT-2125

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -665,8 +665,8 @@ local CONF_INFERENCES = {
   lmdb_environment_path = { typ = "string" },
   lmdb_map_size = { typ = "string" },
 
-  proxy_restricted_hosts = { typ = "array" },
-  proxy_restricted_ports = { typ = "array" },
+  proxy_denied_hosts = { typ = "array" },
+  proxy_denied_ports = { typ = "array" },
 }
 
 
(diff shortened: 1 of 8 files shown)


kong/conf_loader/init.lua
```diff
@@ -660,6 +664,9 @@ local CONF_INFERENCES = {
 
   lmdb_environment_path = { typ = "string" },
   lmdb_map_size = { typ = "string" },
+
+  proxy_restricted_hosts = { typ = "array" },
```
bungle, line 668: Do you think we need to at some point provide both allow and deny, if so perhaps proxy_allowed_hosts or proxy_hosts_allow and proxy_denied_hosts or proxy_hosts_deny?

flrgh: I like this idea. We could make them mutually-exclusive, and the presence of an "allowed" list could invert the logic to deny all except for those explicitly allowed.

I also think the naming of `proxy_denied_hosts`/`proxy_hosts_deny` is slightly less ambiguous when compared to `proxy_restricted_hosts`.

kong/conf_loader/init.lua
```diff
@@ -660,6 +664,9 @@ local CONF_INFERENCES = {
...
+  proxy_restricted_ports = { typ = "array" },
```
flrgh, line 669: One potential use case that I can see coming up is the desire to restrict a specific <host>:<port> tuple, which would not be possible with two separate fields.

hbagdi:
Interesting comments. I think we should do a small change and gauge interest and gather feedback.
This has the potential to become a big policy engine within the API.

kong/conf_loader/init.lua
locao: It's a product decision to not add `proxy_allowed_hosts` now. 

I agree that `proxy_denied_hosts` is clearer, changed.

kong/conf_loader/init.lua
locao: Do you mean a single directive that accepts entries as `169.254.169.254, hostname:*, hostname:22, *:2375`? That might be interesting, what do you think @eskibars?

eskibars: Yeah, I'm pretty sure this will come up and it'd be nice to have a format that allows for it even if I can't think of any services offhand that we should default restrict with a combination host+port

flrgh: Sorry for not giving feedback on this earlier!

> `169.254.169.254, hostname:*, hostname:22, *:2375`

Yes, this is what I was suggesting. This way an operator can deny proxying to `host-a:80` and `host-b:81` while allowing proxying to `host-a:81` and `host-b:80`:

```
proxy_restricted_targets = host-a:80, host-b:81, host-c, :22
```

To me this seems like a very realistic use-case. I don't feel too strongly about it though, so if it's just me saying this, I think the feature is perfectly fine as-is.

flrgh (commented):
Great work!

I have a question: if an operator updates their config and adds a new port to `proxy_denied_ports`, how do we handle pre-existing services/targets that use this port?

A) Consult the config and deny proxying to these services at runtime
B) Allow them to continue functioning normally

My opinion is A, though I could see someone making the case for B (or simply saying that this question is out-of-scope for the initial feature).

Whatever the answer is, I think it would be worth adding a test case for this. That way the behavior is at least codified, well-defined, and future us can't unknowingly change it without breaking an integration test.

fffonion:
The thoughts on this PR is good. However I would suggest to systematically solve this (and similar issues like URI normalization) with WAF, rule based policies will be easier to maintain and inspect then multiple configurations and behavaiour changes (not this PR). @dndx how do you think?

PR #8668, opened 2022-04-11
fix(cp) do a pcall for all calls to export_deflated_reconfigure_payload

gruceo:
We are already wrapping some calls to
`export_deflated_reconfigure_payload()` inside a pcall in the
`control_plane.lua` file. This change is doing a pcall in all the
remaining calls to `export_deflated_reconfigure_payload()` in this file
to avoid the CP crash whenever we find errors during initialization of
modules for example.

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -481,7 +481,10 @@ function _M:handle_cp_websocket()
   self.clients[wb] = queue
 
   if not self.deflated_reconfigure_payload then
-    _, err = self:export_deflated_reconfigure_payload()
+    local ok, _, err = pcall(self.export_deflated_reconfigure_payload, self)
+    if not ok then
+      ngx_log(ngx_ERR, _log_prefix, "unable to export initial config from database: ", err, log_suffix)
+    end
   end
 
   if self.deflated_reconfigure_payload then
@@ -654,8 +657,8 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     return
   end
 
-  local _, err = self:export_deflated_reconfigure_payload()
-  if err then
+  local ok, err = pcall(self.export_deflated_reconfigure_payload, self)
+  if not ok then
     ngx_log(ngx_ERR, _log_prefix, "unable to export initial config from database: ", err)
   end
 
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -58,6 +58,12 @@ local REMOVED_FIELDS = require("kong.clustering.compat.removed_fields")
 local _log_prefix = "[clustering] "
 
 
+local function handle_export_deflated_reconfigure_payload(self)
+  local ok, p_err, err = pcall(self.export_deflated_reconfigure_payload, self)
+  return ok, p_err or err
+end
+
+
 local function plugins_list_to_map(plugins_list)
   local versions = {}
   for _, plugin in ipairs(plugins_list) do
@@ -481,10 +487,7 @@ function _M:handle_cp_websocket()
   self.clients[wb] = queue
 
   if not self.deflated_reconfigure_payload then
-    local ok, _, err = pcall(self.export_deflated_reconfigure_payload, self)
-    if not ok then
-      ngx_log(ngx_ERR, _log_prefix, "unable to export initial config from database: ", err, log_suffix)
-    end
+    _, err = handle_export_deflated_reconfigure_payload(self)
   end
 
   if self.deflated_reconfigure_payload then
@@ -657,7 +660,7 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     return
   end
 
-  local ok, err = pcall(self.export_deflated_reconfigure_payload, self)
+  local ok, err = handle_export_deflated_reconfigure_payload(self)
   if not ok then
     ngx_log(ngx_ERR, _log_prefix, "unable to export initial config from database: ", err)
   end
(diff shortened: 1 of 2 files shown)


javierguerragiraldez:
it's missing similar calls in `wrpc_control_plane.lua`

gruceo:
@javierguerragiraldez @fffonion 

Tests are green 🟢 https://github.com/Kong/kong/runs/6009975751

kong/clustering/control_plane.lua
```diff
@@ -481,7 +481,10 @@ function _M:handle_cp_websocket()
   self.clients[wb] = queue
 
   if not self.deflated_reconfigure_payload then
-    _, err = self:export_deflated_reconfigure_payload()
+    local ok, err = pcall(self.export_deflated_reconfigure_payload, self)
```
javierguerragiraldez, line 484: this changes the meaning of the `err` variable.  to preserve previous error handling, it should keep both return values in addition of the new `ok`
```suggestion
    local ok, _, err = pcall(self.export_deflated_reconfigure_payload, self)
```

gruceo: https://github.com/Kong/kong/pull/8668/commits/e9142256fc4f7e5f4d4702df6a69a82df9e4cd62

kong/clustering/wrpc_control_plane.lua
```diff
@@ -170,7 +170,7 @@ end
 
 function _M:push_config_one_client(client)
   if not self.config_call_rpc or not self.config_call_args then
-    local payload, err = self:export_deflated_reconfigure_payload()
+    local payload, err = pcall(self.export_deflated_reconfigure_payload, self)
```
fffonion, line 173: the first return value of pcall is a boolean, the original return value of `self.export_deflated_reconfigure_payload` will be the second return value with pcall. we should update the variables here to avoid confusion. since we are not using the `payload` it seems, update this to `ok, err` and L174 to `if not ok` seems ok.

gruceo: https://github.com/Kong/kong/pull/8668/commits/8718c80876cc8d9b9fc993eb3a7dc48b65d974a6

javierguerragiraldez: i'm not sure exactly what error these pcalls are supposed to catch, but the fact that the tests still pass is proof to me that they don't cover the failure cases.

javierguerragiraldez: i.e: an idiomatic bare lua function:

`local val, err = func(...)`
if it succeeds, `val` has a valid value
if it fails, `val` is falsy and `err` has an error message
if it errors, it doesn't return, the error is propagated upwards

the same, wrapped with pcall():

`local ok, val, err = pcall(func, ...)`
if it succeeds, `ok==true`, `val` has a valid value
if it fails, `ok==true`, `val` is falsy and `err` has an error message
if it errors, `ok==false`, `val` has an error message

so, just wrapping is not enough.  the error handling must manage the new failure paths.   This patch applies pcall() inconsistently, and doesn't handle errors.  perhaps those could be handled in a common function that does the wrapping and catches the errors correctly.

gruceo: @javierguerragiraldez Thanks for all those points... I went ahead and created a separate commit addressing your comment: https://github.com/Kong/kong/pull/8668/commits/9d0929394927b7a46cfc69c81b4190ca1646788e

Please carefully look through it... hopefully the patch is cleaner and more accurate now. I tried adding a lot of comments in the new common function that does the wrapping and the error handling to avoid confusion, but let me know if the comments are overkill or if there is anything else we need to do for this patch.

javierguerragiraldez: yes, i think this is *much* better now

javierguerragiraldez (approved):

gruceo:
@javierguerragiraldez @fffonion is this PR good to merge?

gruceo:
I don't have merge permissions

kong/clustering/control_plane.lua
```diff
@@ -58,6 +58,21 @@ local REMOVED_FIELDS = require("kong.clustering.compat.removed_fields")
...
+  if not ok then
+    -- hard error -> ok == false, p_err has an error message
+    return false, p_err
+  end
+  if not p_err then
+    -- fail from typical return nil, err -> ok == true, p_err == nil, err has an error message
+    return false, err
+  end
+  -- succeeded -> ok == true, p_err == nil, err == nil
+  return true
```
kikito, line 72: I think this section can be simplified to 
```suggestion
  return ok, p_err or err
```

kong/clustering/wrpc_control_plane.lua
```diff
@@ -50,6 +50,22 @@ local _log_prefix = "[wrpc-clustering] "
 
 local wrpc_config_service
 
+
+local function handle_export_deflated_reconfigure_payload(self)
+  local ok, p_err, err = pcall(self.export_deflated_reconfigure_payload, self)
+  if not ok then
```
kikito, line 56: Ditto

kong/clustering/control_plane.lua
gruceo: https://github.com/Kong/kong/pull/8668/commits/ebf080739312a08756664079ad05d03eb40cd1f9

kong/clustering/wrpc_control_plane.lua
gruceo: https://github.com/Kong/kong/pull/8668/commits/ebf080739312a08756664079ad05d03eb40cd1f9

kikito (approved):

PR #8670, opened 2022-04-12
Add dbless persistence

aboudreault:
This PR makes the dbless persistent between restarts (lmdb). If a config exists in the db, load it and build the router.

With this addition, we can remove the config cache (config.cache.tar.gz) (I will open another PR for this).

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -686,7 +686,11 @@ function Kong.init_worker()
       -- if there is no declarative config set and a config is present in LMDB,
       -- just build the router and plugins iterator
       ngx_log(ngx_INFO, "found persisted lmdb config, loading...")
-      declarative_init_build()
+      local ok, err = declarative_init_build()
+      if not ok then
+        stash_init_worker_error("failed to initialize declarative config: " .. err)
+        return
+      end
     elseif declarative_entities then
       ok, err = load_declarative_config(kong.configuration,
                                         declarative_entities,
@@ -699,7 +703,11 @@ function Kong.init_worker()
     else
       -- stream does not need to load declarative config again, just build
       -- the router and plugins iterator
-      declarative_init_build()
+      local ok, err = declarative_init_build()
+      if not ok then
+        stash_init_worker_error("failed to initialize declarative config: " .. err)
+        return
+      end
     end
   end
 
--- a/spec/02-integration/11-dbless/03-config_persistence.lua
+++ b/spec/02-integration/11-dbless/03-config_persistence.lua
@@ -85,7 +85,7 @@ describe("dbless persistence with a declarative config #off", function()
         database   = "off",
         declarative_config = yaml_file,
     }))
-  
+
     admin_client = assert(helpers.admin_client())
     proxy_client = assert(helpers.proxy_client())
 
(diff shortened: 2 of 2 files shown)


kong/init.lua
```diff
@@ -649,7 +654,13 @@ function Kong.init_worker()
       return
     end
 
-    if declarative_entities then
+    if not has_declarative_config(kong.configuration) and
+      declarative.get_current_hash() ~= nil then
+      -- if there is no declarative config set and a config is present in LMDB,
+      -- just build the router and plugins iterator
+      ngx_log(ngx_INFO, "found persisted lmdb config, loading...")
+      declarative_init_build()
```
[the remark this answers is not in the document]
aboudreault: In that case, we probably add a basic handling here too? https://github.com/Kong/kong/blob/master/kong/init.lua#L664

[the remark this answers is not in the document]
gszr: Feel free to add it in the same PR / commit.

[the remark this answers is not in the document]
aboudreault: Fixed

dndx:
@bungle is it possible for you to take a closer look at this (to make sure we are not overlooking anything)? Especially make sure the reload doesn't mess up anything.

spec/02-integration/11-dbless/03-config_persistence.lua
```diff
@@ -0,0 +1,157 @@
...
+    }))
+    proxy_client = assert(helpers.proxy_client())
+    local res = assert(proxy_client:get("/test", { headers = { host = "example1.dev" } }))
+    assert.res_status(401, res)
+
+    res = assert(proxy_client:get("/500", { headers = { host = "example1.dev" } }))
+    assert.res_status(404, res) -- 404, only the declarative config is loaded
+  end)
+
+  it("doesn't load the persisted lmdb config if a declarative config is set on reload", function()
```
flrgh, line 139: For good measure, I think it would be worthwhile to have a test case that covers worker process replacement (`kill -9` an existing worker and then validate expected behavior) as well.

gszr (approved):

spec/02-integration/11-dbless/03-config_persistence.lua
aboudreault: @flrgh  would you consider this test a blocker? I thought that it will basically do the same thing than an initial worker start, cause the code is right in `init_worker`.

flrgh: Nah I wouldn't go as far as to say it's a blocker. I don't know this code as well as you do, so I think this is up to your discretion whether more tests are needed before merging. To clarify, there's nothing in particular about this changeset that makes me believe it will introduce any _new_ bugs. I just think our test coverage of single worker replacement is lacking overall (it's a very common blind spot I've observed in OpenResty code), so I'll almost always advocate for adding more tests when `init_worker` is concerned.

aboudreault: makes sense @flrgh! I will check to add one before merging.

bungle (approved):

PR #8672, opened 2022-04-12
chore(rockspec) replace git protocol by https

kikito:
Luarocks upload was failing when trying to download the rock from
Github, which was failing with "The unauthenticated git protocol
on port 9418 is no longer supported". More information here:

https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git

We're replacing git with https, which is one of the fixes recommended on
that guide.


kikito:
Note that this pull request is for the `release/2.8.x` branch.

mayocream (approved):

PR #8679, opened 2022-04-13
feat(clustering) optional config override for protocol.

javierguerragiraldez:
Not a feature flag, but an override.  The DP performs the negotiation as
usual, but if the setting has a non-nil value, it's used instead of the
negotiated one.

Also handy for doing every test on both protocols.

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -285,7 +285,11 @@ function _M:init_worker()
 
       self:request_version_negotiation()
 
-      local config_proto, msg = version_negotiation.get_negotiated_service("config")
+      local config_proto, msg = self.conf.cluster_protocol, "set by configuration"
+      if not config_proto then
+        config_proto, msg = version_negotiation.get_negotiated_service("config")
+      end
+
       if not config_proto and msg then
         ngx_log(ngx_ERR, _log_prefix, "error reading negotiated \"config\" service: ", msg)
       end
--- a/kong/templates/kong_defaults.lua
+++ b/kong/templates/kong_defaults.lua
@@ -21,6 +21,7 @@ admin_listen = 127.0.0.1:8001 reuseport backlog=16384, 127.0.0.1:8444 http2 ssl 
 status_listen = off
 cluster_listen = 0.0.0.0:8005
 cluster_control_plane = 127.0.0.1:8005
+cluster_protocol = NONE
 cluster_cert = NONE
 cluster_cert_key = NONE
 cluster_mtls = shared
--- a/spec/02-integration/09-hybrid_mode/01-sync_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/01-sync_spec.lua
@@ -17,7 +17,7 @@ local KEY_AUTH_PLUGIN
 
 
 for _, strategy in helpers.each_strategy() do
-  for _, cluster_protocol in ipairs{"json", "wrpc"} do
+  for _, cluster_protocol in ipairs{"v0", "v1"} do
     describe("CP/DP sync works with #" .. strategy .. " backend, protocol " .. cluster_protocol, function()
 
       lazy_setup(function()
@@ -559,7 +559,7 @@ for _, strategy in helpers.each_strategy() do
     end)
   end)
 
-  for _, cluster_protocol in ipairs{"json", "wrpc"} do
+  for _, cluster_protocol in ipairs{"v0", "v1"} do
     describe("CP/DP sync works with #" .. strategy .. " backend, protocol " .. cluster_protocol, function()
       lazy_setup(function()
         helpers.get_db_utils(strategy, {
@@ -705,7 +705,7 @@ for _, strategy in helpers.each_strategy() do
 
       assert(helpers.start_kong({
         role = "data_plane",
-        cluster_protocol = "json",
+        cluster_protocol = "v0",
         database = "off",
         prefix = "servroot2",
         cluster_cert = "spec/fixtures/kong_clustering.crt",
@@ -716,7 +716,7 @@ for _, strategy in helpers.each_strategy() do
 
       assert(helpers.start_kong({
         role = "data_plane",
-        cluster_protocol = "wrpc",
+        cluster_protocol = "v1",
         database = "off",
         prefix = "servroot3",
         cluster_cert = "spec/fixtures/kong_clustering.crt",
--- a/spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
@@ -1,7 +1,7 @@
 local helpers = require "spec.helpers"
 
 
-for _, cluster_protocol in ipairs{"json", "wrpc"} do
+for _, cluster_protocol in ipairs{"v0", "v1"} do
   describe("invalid config are rejected, protocol " .. cluster_protocol, function()
     describe("role is control_plane", function()
       it("can not disable admin_listen", function()
--- a/spec/02-integration/09-hybrid_mode/03-pki_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/03-pki_spec.lua
@@ -2,7 +2,7 @@ local helpers = require "spec.helpers"
 local cjson = require "cjson.safe"
 
 
-for _, cluster_protocol in ipairs{"json", "wrpc"} do
+for _, cluster_protocol in ipairs{"v0", "v1"} do
   for _, strategy in helpers.each_strategy() do
     describe("CP/DP PKI sync works with #" .. strategy .. " backend, protocol " .. cluster_protocol, function()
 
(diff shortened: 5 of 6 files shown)


kong/clustering/init.lua
```diff
@@ -285,7 +285,11 @@ function _M:init_worker()
 
       self:request_version_negotiation()
 
-      local config_proto, msg = version_negotiation.get_negotiated_service("config")
+      local config_proto, msg = self.conf.cluster_protocol, "set by configuration"
```
ADD-SP, line 288: Although there is no problem now, I think it is still necessary to check if the protocol specified by `self.conf.cluster_protocol` is accepted by CP.

I think we need to keep all the protocols that are accepted by the CP to get this done.

kong/templates/kong_defaults.lua
```diff
@@ -21,6 +21,7 @@ admin_listen = 127.0.0.1:8001 reuseport backlog=16384, 127.0.0.1:8444 http2 ssl
 status_listen = off
 cluster_listen = 0.0.0.0:8005
 cluster_control_plane = 127.0.0.1:8005
+cluster_protocol = NONE
```
ADD-SP, line 24: https://github.com/Kong/kong/blob/7fe16e6330d9525d8ecd154f573282a24c5c41ab/kong/conf_loader/init.lua#L643-L652

https://github.com/Kong/kong/blob/7fe16e6330d9525d8ecd154f573282a24c5c41ab/kong/db/declarative/init.lua#L401-L404

https://github.com/Kong/kong/blob/7fe16e6330d9525d8ecd154f573282a24c5c41ab/CHANGELOG.md?plain=1#L124-L125

https://github.com/Kong/kong/blob/7fe16e6330d9525d8ecd154f573282a24c5c41ab/kong.conf.default#L190-L193

These four files should be updated if we allow users to specify the CP/DP protocol.

ADD-SP (commented):
I have on more question, do we need to change the existing handshake protocols to keep more information? For example, keep all protocols that are accepted by the CP, not just one.

```
ControlPlane                                        DataPlane

ControlPlaneHello
+ Requested protocols (sorted)
+ Other Information
                                ------------->
                                                    DataPlaneHello
                                                    + Accepted protocols (sorted)
                                                    + Other Information
                                <-------------
```

Suika-Kong:
As people suggested, just for testing is not a good reason to do so, and this option is very likely to be abused by users.
Can we find a better solution, like a patch before (and only for) tests?

hbagdi:
I _think_ this is how the code should function:
1. DP has a hardcoded list of protocols it supports
2. DP sends this over to CP
3. CP responds back with the protocol chosen
4. DP and CP start communicating on this protocol

The configuration option being added here influences step 2, the DP still must send the protocol over to the CP to negotiate it.

> As people suggested, just for testing is not a good reason to do so, and this option is very likely to be abused by users.

1. It depends on how you look at it. It can also be useful if there is a serious bug in a newer protocol. Giving an escape hatch to our users is an insurance policy that we should buy.
2. If by "abuse" you mean that certain bugs can be exploited, I see your point. An attacker could certainly use that surface but if an attacker has gained permission to change kong.conf, there are other bigger worries than the exploitation of bugs in Kong. What do you think?

Suika-Kong:
> I _think_ this is how the code should function:
> 
> 1. DP has a hardcoded list of protocols it supports
> 2. DP sends this over to CP
> 3. CP responds back with the protocol chosen
> 4. DP and CP start communicating on this protocol
> 
> The configuration option being added here influences step 2, the DP still must send the protocol over to the CP to negotiate it.
> 
> > As people suggested, just for testing is not a good reason to do so, and this option is very likely to be abused by users.
> 
> 1. It depends on how you look at it. It can also be useful if there is a serious bug in a newer protocol. Giving an escape hatch to our users is an insurance policy that we should buy.
> 2. If by "abuse" you mean that certain bugs can be exploited, I see your point. An attacker could certainly use that surface but if an attacker has gained permission to change kong.conf, there are other bigger worries than the exploitation of bugs in Kong. What do you think?

By design, CP is compatible with older DPs, but not vice versa, and maybe later we will stop maintaining the DP's backward compatibility. The design to designate versions from DP will be meaningless in that case, and the tests depending on this will also fail.
By "abuse" I mean, it would always be a bug if the newer version fails. The better option would be for users to switch back to an older version of the gateway itself, rather than using an older version of the protocol.

Suika-Kong:
And I‘ve considered CP to pretend it does not support newer protocol, but the main purpose of this test is to show that newer CP supports older DPs

javierguerragiraldez:
I feel that different things are conflated here.  It would be great if a single concept could manage everything, but first we have to be sure what we're talking about:

1. Both CP and DP have hardcoded lists of what they support.  I think it's best for this to be hardcoded, since it should reflect the respective code capabilities on each side.
2.  DP presents the CP with its supported list.
3. CP responds with the entries it prefers.  currently the algorithm is trivial: pick the first entry from it's own list that also appears on the DP list.  But it could be as complex as we want.  for example, we could set a list of node-specific overrrides.  or allow some version only if the DP also asks for another specific service.  or whatever.
4. Both CP and DP keep the negotiation results and could read them and use for whatever purpose.  currently neither side stores a list of other non-chosen options from the other side.

the version negotiation task finishes there.  It's supposed to work with any service we want to expose on wRPC, but actually the "service" could be anything that we want to define versions on.

In order to allow some "manual override" from the DP, I first thought of something very general, so we wouldn't have to rewrite for any new protocol/version.  my first Idea was to add a "ban list" that removes version or services from the DP list before step 2 (the empty default would mean no change).  It wasn't well received, so I came up with this other solution (the one implemented in this PR).  It doesn't interfere with negotiation at all.  It simply adds a specific configuration option which, if set, is used instead of whatever was specified in the negotiation. If any other code consults the negotiation option, it's still what the CP accepted and not (necessarily) what the DP actually used.

javierguerragiraldez:
hey guys, how can we get this moving?
I really think we need this or some other solution that allows a DP to override whatever is negotiated.  not to be used everywhere, but to fix specific problems, and sometimes to troubleshoot complex interactions between DP and CP, especially around updates.

As I see it, there are some different options:

1. specific flags that are checked by a given feature to choose, ignoring the negotiation (this PR, for protocol)
2. a "banned services/versions" setting that removes options from the set the DP presents to the CP.  This keeps the negotiation process but allows the user to avoid a particular codepath.
3. tweak the CP choosing/rejection logic?

I'm ok with either (my personal favorite is number 2), but something is necessary

ADD-SP:
> hey guys, how can we get this moving? I really think we need this or some other solution that allows a DP to override whatever is negotiated. not to be used everywhere, but to fix specific problems, and sometimes to troubleshoot complex interactions between DP and CP, especially around updates.
> 
> As I see it, there are some different options:
> 
> 1. specific flags that are checked by a given feature to choose, ignoring the negotiation (this PR, for protocol)
> 2. a "banned services/versions" setting that removes options from the set the DP presents to the CP.  This keeps the negotiation process but allows the user to avoid a particular codepath.
> 3. tweak the CP choosing/rejection logic?
> 
> I'm ok with either (my personal favorite is number 2), but something is necessary

I think it would be better for CP to return a list of all accepted protocols, preferably sorted by priority.

javierguerragiraldez:
> I think it would be better for CP to return a list of all accepted protocols, preferably sorted by priority.

that is not how version negotiation works (CP returns a single chosen version for each accepted service), nor what this PR is about (allowing the administrator to override negotiation)

I don't see how changing the negotiation protocol would change this PR.  it would still require some override mechanism.

ADD-SP:
> > I think it would be better for CP to return a list of all accepted protocols, preferably sorted by priority.
> 
> that is not how version negotiation works (CP returns a single chosen version for each accepted service), nor what this PR is about (allowing the administrator to override negotiation)
> 
> I don't see how changing the negotiation protocol would change this PR. it would still require some override mechanism.

The purpose of this PR is to allow users to force the use of a certain protocol.

The `wrpc` protocol may also be upgraded in the future, such as `wrpcv1.1` `wrpcv1.4` `wrpcv2.0`, etc.

If the user wants to use `wrpcv1.4`. but the response from the CP maybe "I accept `wrpcv2.0`", but in reality, the CP may still support `wrpcv1.4`. this can cause some problems.

Suika-Kong:
I personally believe that even if we want to hack with the DP version, we need to follow the standard, especially for the negotiation.
So it's better we do this from the list of "supported versions", to fake a list with only the designated version.

hbagdi:
> The purpose of this PR is to allow users to force the use of a certain protocol.

That is incorrect. The purpose is to make DP use a specific version of a wRPC service. This doesn't cover different versions of wRPC spec itself.

The purpose here is to *force* a specific version _from_ the DP and assert DP and CP behaviors from that configuration.
Yes, it is possible that the CP doesn't support the version being forced by this configuration property, and that is completely okay.
Why? Let's take a step back, this feature is really for debugging and more so testing of DP with different CP implementations, as well as to allow for testing for older versions of a wRPC service (once there are multiple versions of wRPC service - not the case yet).

javierguerragiraldez:
obsoleted by #8775

PR #8528, opened 2022-03-10
fix(admin-status) use resty-http to get nginx_status

fdeschamps:
`ngx.location.capture` does not support http2 requests
http2 requets to `<kong_admin>/status` was failing
resty-http supports http2

### Summary

By defaut, kong config activates http2 on admin api :
`admin_listen = 127.0.0.1:8001 reuseport backlog=16384, 127.0.0.1:8444 http2 ssl reuseport backlog=16384`

But the `/status` endpoint is not compatible with http2, I get this error in admin_error logs when I do a `curl https://localhost:8444/status` : 
```
2022/03/09 17:29:25 [error] 44#0: *475 [lua] api_helpers.lua:511: handle_error(): /usr/local/share/lua/5.1/lapis/application.lua:424: /usr/local/share/lua/5.1/kong/api/routes/health.lua:45: http2 requests not supported yet
stack traceback:
        [C]: in function 'capture'
        /usr/local/share/lua/5.1/kong/api/routes/health.lua:45: in function 'fn'
        /usr/local/share/lua/5.1/kong/api/api_helpers.lua:285: in function </usr/local/share/lua/5.1/kong/api/api_helpers.lua:268>

stack traceback:
        [C]: in function 'error'
        /usr/local/share/lua/5.1/lapis/application.lua:424: in function 'handler'
        /usr/local/share/lua/5.1/lapis/application.lua:146: in function 'resolve'
        /usr/local/share/lua/5.1/lapis/application.lua:183: in function </usr/local/share/lua/5.1/lapis/application.lua:181>
        [C]: in function 'xpcall'
        /usr/local/share/lua/5.1/lapis/application.lua:189: in function 'dispatch'
        /usr/local/share/lua/5.1/lapis/nginx.lua:231: in function 'serve'
        /usr/local/share/lua/5.1/kong/init.lua:1423: in function 'admin_content'
        content_by_lua(nginx-kong.conf:349):2: in main chunk, client: 127.0.0.1, server: kong_admin, request: "GET /status HTTP/2.0", host: "localhost:8444"
```

By googling a little, I saw that `lua-resty-http` implementation supported http2 and I tried to patch the code.
It seems to work on my end.

### Full changelog

* replace `ngx.location.capture` by a `lua-resty-http` implementation in `/status` admin API endpoint

PR #8687, opened 2022-04-14
fix(api) support HTTP/2 status and metrics requests

rainest:
### Summary

Rework the status and metrics endpoint implementations to make them not break on HTTP/2.

### Full changelog

* Add a new Unix socket listen that runs NGINX's stub_status on requests for /nginx_status at $PREFIX/nginx_status.sock.
* Use lua-resty-http to access the new status socket listen in the status endpoint instead of using ngx.location.capture (which breaks if invoked during an HTTP/2 request) on the existing internal locations.
* Ditto for Prometheus /metrics.

### Issue reference

Discovered that https://github.com/Kong/kubernetes-ingress-controller/pull/2343 did not work with HTTPS during post-release testing. We'd prefer to change the Kong end rather than disable HTTP/2 across the board on the controller end.

Implements @dndx's suggestion from https://github.com/Kong/kong/pull/8528

### Draft problems

Edit: https://github.com/Kong/kong/pull/8687/commits/671e15ebc289e53781197662d3419a5fea4bdb4e works around this: https://github.com/Kong/kong/pull/8687#issuecomment-1101803889

This exposes a [possible issue in spec/02-integration/02-cmd/02-start_stop_spec.lua](https://github.com/rainest/kong/runs/6030515315?check_suite_focus=true#step:9:173):

```
cat: /home/runner/work/kong/kong/servroot/pids/nginx.pid: No such file or directory
1146.04   OK 635: kong start/stop #postgres deprecated properties prints a warning to stderr client_max_body_size
        __________
        FAILED  spec/02-integration/02-cmd/02-start_stop_spec.lua:641: kong start/stop #postgres deprecated properties prints a warning to stderr client_body_buffer_size
spec/02-integration/02-cmd/02-start_stop_spec.lua:608: 2022/04/14 20:33:48 [warn] the 'client_body_buffer_size' configuration property is deprecated, use 'nginx_http_client_body_buffer_size' instead
Error: ./kong/cmd/start.lua:64: nginx: [emerg] bind() to unix:/home/runner/work/kong/kong/servroot/nginx_status.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:/home/runner/work/kong/kong/servroot/nginx_status.sock failed (98: Address already in use)
```
This change adds a unix socket that is almost always (either admin or status listen enabled) present. I think I'm hitting something in those tests that has multiple overlapping Kong instances. The `cat` error just before the conflict suggests we were trying to check if an instance was still up and/or kill it but failed due to a missing pidfile, but I'm not familiar enough with the tests to say for sure. I can't reproduce it when testing locally--I get no PID conflicts, just complaints about Cassandra and Postgres availability in 3 out of 39 tests. The issue flakes/does not occur consistently.

Manual testing (do we have HTTP/2 test client stuff yet?): [example.txt](https://github.com/Kong/kong/files/8492387/example.txt)

### DIFF
--- a/kong/api/routes/health.lua
+++ b/kong/api/routes/health.lua
@@ -1,5 +1,6 @@
 local utils = require "kong.tools.utils"
 local declarative = require "kong.db.declarative"
+local http = require "resty.http"
 
 local find = string.find
 local select = select
@@ -14,6 +15,7 @@ local tonumber = tonumber
 local kong = kong
 local dbless = kong.configuration.database == "off"
 local data_plane_role = kong.configuration.role == "data_plane"
+local nginx_status_sock = "unix:" .. ngx.config.prefix() .. "nginx_status.sock"
 
 
 return {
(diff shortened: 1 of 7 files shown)


rainest:
Alright, well, workaround is less than ideal for using a specific socket path, but does apparently avoid whatever the issue was. Guessing that since `kill_all` only sends a signal there's no guarantee when NGINX actually stops, but since the function's done the `after_each()` exits and continues with the next test.

Also tried:

- `pl.delete(<prefix>/*.sock)`, which I'm guessing didn't work because either it doesn't expand globs or doesn't work on socket files.
- Moving the `clean_prefix()` call to `after_each`. This broke because some tests get unhappy if the prefix isn't around yet--the `nginx_main_daemon = off` can't deal with the extra logs about creating it again.

rainest:
#8690 probably obsoletes this.

PR #8688, opened 2022-04-14
feat(pdk) add kong.request.get_start_time()

flrgh:
### Summary

This adds `kong.request.get_start_time` as a PDK function. Before this change the business logic for computing this value was contained in `kong.log.serialize()`, so the only stable-ish API for fetching start time was `kong.log.serialize().started_at` (kind of wasteful/inelegant). Therefore this seems like a worthwhile inclusion to the PDK.

### Full changelog

* Add `kong.request.get_start_time` PDK function.
* Update `kong.log.serialize` to use `kong.request.get_start_time`

### DIFF
--- a/kong/pdk/log.lua
+++ b/kong/pdk/log.lua
@@ -747,7 +747,6 @@ do
 
       local ctx = ongx.ctx
       local var = ongx.var
-      local req = ongx.req
 
       local authenticated_entity
       if ctx.authenticated_credential ~= nil then
@@ -809,7 +808,7 @@ do
         service = ctx.service,
         consumer = ctx.authenticated_consumer,
         client_ip = var.remote_addr,
-        started_at = ctx.KONG_PROCESSING_START or (req.start_time() * 1000)
+        started_at = okong.request.get_start_time(),
       })
     end
 
@@ -818,10 +817,10 @@ do
       check_phase(PHASES_LOG)
 
       local ongx = (options or {}).ngx or ngx
+      local okong = (options or {}).kong or kong
 
       local ctx = ongx.ctx
       local var = ongx.var
-      local req = ongx.req
 
       local authenticated_entity
       if ctx.authenticated_credential ~= nil then
@@ -865,7 +864,7 @@ do
         service = ctx.service,
         consumer = ctx.authenticated_consumer,
         client_ip = var.remote_addr,
-        started_at = ctx.KONG_PROCESSING_START or (req.start_time() * 1000)
+        started_at = okong.request.get_start_time(),
       })
     end
   end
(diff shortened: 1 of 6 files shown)


kong/pdk/request.lua
```diff
@@ -796,6 +796,21 @@ local function new(self)
     end
   end
 
+  ---
+  -- Returns the request start time, in milliseconds.
```
hbagdi, line 800:
```suggestion
  -- Returns the request start time, in Unix epoch milliseconds.
```

hbagdi (commented):
Glad to see this change.
I'd also like to see other latency calculations moved into the PDK, improving those calculations along the way (some of them were deemed logically inaccurate in the past).
Big +1

kong/pdk/log.lua
```diff
@@ -809,7 +808,7 @@ do
         service = ctx.service,
         consumer = ctx.authenticated_consumer,
         client_ip = var.remote_addr,
-        started_at = ctx.KONG_PROCESSING_START or (req.start_time() * 1000)
+        started_at = okong.request.get_start_time(),
```
hbagdi, line 811: I'm not familiar with how the PDK is laid out and the reasons behind it, but if we are going to expose more timestamps and latencies from the PDK, having a dedicated module to do that may be better putting things into the `request` module.
`kong.telemetry`, `kong.time` or something along those lines may be a good option.

flrgh: Oh boy I would _love_ to move all of the `ctx.KONG_*_{START,ENDED_AT,LATENCY,TIME}` read/write/calculate stuff into its own module (with a corresponding PDK interface). I've wondered many a time why this hasn't been done before; it seems like low-hanging fruit (maybe there's an angle to it that I'm missing).

hbagdi: >  I've wondered many a time why this hasn't been done before; it seems like low-hanging fruit (maybe there's an angle to it that I'm missing).

I can think of a few different reasons why this is the case. But yes, it is low-hanging fruit, and, there are plenty of those in the codebase. I'll say that it takes skill and a _bit_ of thinking to recognize those.

fffonion:
@flrgh would you resolve the comments and rebase this branch upon master and try if the CI will pass?

kong/pdk/log.lua
flrgh: Resolving this convo (for now) but keeping these ideas in mind. It feels a little late in the release cycle for me to start touching lots of stuff in `kong/init.lua`, and it should be trivial to maintain the backwards compatibility of this new `kong.request.get_start_time` function if/when I can revisit all of this.

fffonion (approved):

PR #8698, opened 2022-04-19
fix(pdk) ignore user set Tranfer-Encoding

### LINKED ISSUE #8083: kong.response.exit setting Content-Length when Transfer-Encoding is present

Suika-Kong:
Users should not set the Tranfer-Encoding header and it is also not supported by OpenResty.

Fix #8083

Fix CT-312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,13 @@
 
 ## Unreleased
 
+### Fixes
+
+#### PDK
+
+- `pdk.response.set_header()`, `pdk.response.set_headers()`, `pdk.response.exit()` now ignore and emit warnings for manually set `Transfer-Encoding` headers.
+  [#8698](https://github.com/Kong/kong/pull/8698)
+
 ### Breaking Changes
 
 #### Admin API
@@ -79,8 +86,6 @@
 
 - The PDK is no longer versioned
   [#8585](https://github.com/Kong/kong/pull/8585)
-- `pdk.response.set_header()`, `pdk.response.set_headers()`, `pdk.response.exit()` now ignore and emit warnings for mannually set `Transfer-Encoding` headers.
-  [#8698](https://github.com/Kong/kong/pull/8698)
 
 ### Deprecations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 
 - The PDK is no longer versioned
   [#8585](https://github.com/Kong/kong/pull/8585)
+- `pdk.response.set_header()`, `pdk.response.set_headers()`, `pdk.response.exit()` now ignore and emit warnings for mannually set `Transfer-Encoding` headers.
 
 ### Deprecations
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -79,6 +79,8 @@
 
 - The PDK is no longer versioned
   [#8585](https://github.com/Kong/kong/pull/8585)
+- `pdk.response.set_header()`, `pdk.response.set_headers()`, `pdk.response.exit()` now ignore and emit warnings for mannually set `Transfer-Encoding` headers.
```
kikito, line 82: I think this is not a breaking change since setting that header is unreliable in Openresty. I would consider this a Fix instead. Also, typo:

```suggestion
- `pdk.response.set_header()`, `pdk.response.set_headers()`, `pdk.response.exit()` now ignore and emit warnings for manually set `Transfer-Encoding` headers.
```

kong/pdk/response.lua
```diff
@@ -408,6 +408,11 @@ local function new(self, major_version)
     end
 
     validate_header(name, value)
+    local lower_name = lower(name)
+    if lower_name == "transfer-encoding" or lower_name == "transfer_encoding" then
+      self.log.warn("mannually setting Transfer-Encoding. Ignored.")
```
kikito, line 413:
```suggestion
      self.log.warn("manually setting Transfer-Encoding. Ignored.")
```

kong/pdk/response.lua
```diff
@@ -514,7 +519,12 @@ local function new(self, major_version)
     validate_headers(headers)
 
     for name, value in pairs(headers) do
-      ngx.header[name] = normalize_multi_header(value)
+      local lower_name = lower(name)
+      if lower_name == "transfer-encoding" or lower_name == "transfer_encoding" then
+        self.log.warn("mannually setting Transfer-Encoding. Ignored.")
```
kikito, line 524:
```suggestion
        self.log.warn("manually setting Transfer-Encoding. Ignored.")
```

t/01-pdk/08-response/11-exit.t
```diff
@@ -1125,3 +1125,32 @@ unable to proxy stream connection, status: 400, err: error message
...
+    }
+--- request
+GET /t
+--- response_body
+test
+--- response_headers
+Content-Length: 5
+X-test: test
+--- error_log
+mannually setting Transfer-Encoding. Ignored.
```
kikito, line 1154:
```suggestion
manually setting Transfer-Encoding. Ignored.
```

kikito (approved):

PR #8701, opened 2022-04-19
feat(balancer) add path, uri_capture, and query as hash sources

flrgh:
This adds `path`, `uri_capture`, and `query_arg` as hash sources for upstreams.

There are some code simplicity vs performance vs absolute correctness tradeoffs to consider here:

* `path` sources its value from `$uri`, which is normalized according to NGINX rules and _not_ `kong.tools.uri.normalize($request_uri)`. The balancer hash is not directly exposed via any public API and is much less sensitive to compatibility issues than other areas of the codebase (i.e. `kong.router`), so I think it's okay to not be 100% consistent here.
* `query_arg` uses `ngx.req.get_uri_args`, which is semi-expensive as it creates a new table (similar to `hash_on_header` now). A feature of an upcoming OpenResty release allows optimizing away the table creation ([link](https://github.com/openresty/lua-resty-core/pull/288)), which I've accounted for. The benefit of doing this over using `ngx.var.arg_{{name}}` is that NGINX handles decoding for us, _and_ that it works for multi-value query string args.

### DIFF
--- a/kong/db/schema/entities/upstreams.lua
+++ b/kong/db/schema/entities/upstreams.lua
@@ -32,7 +32,7 @@ end
 local hash_on = Schema.define {
   type = "string",
   default = "none",
-  one_of = { "none", "consumer", "ip", "header", "cookie" }
+  one_of = { "none", "consumer", "ip", "header", "cookie", "path", "query_arg", "uri_capture" }
 }
 
 
@@ -79,6 +79,10 @@ local health_threshold = Schema.define {
   between = { 0, 100 },
 }
 
+local simple_param = Schema.define {
+  type = "string",
+  len_min = 1,
+}
 
 local NO_DEFAULT = {}
 
@@ -186,6 +190,10 @@ local r =  {
     { hash_fallback_header = typedefs.header_name, },
     { hash_on_cookie = { type = "string",  custom_validator = utils.validate_cookie_name }, },
     { hash_on_cookie_path = typedefs.path{ default = "/", }, },
+    { hash_on_query_arg = simple_param },
+    { hash_fallback_query_arg = simple_param },
+    { hash_on_uri_capture = simple_param },
+    { hash_fallback_uri_capture = simple_param },
     { slots = { type = "integer", default = 10000, between = { 10, 2^16 }, }, },
     { healthchecks = { type = "record",
         default = healthchecks_defaults,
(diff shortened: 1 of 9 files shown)


kong/runloop/balancer/init.lua
```diff
@@ -97,6 +133,18 @@ local function get_value_to_hash(upstream, ctx)
         }
       end
 
+    elseif hash_on == "path" then
+      -- for the sake of simplicity, we're using the NGINX-normalized version of
+      -- the path here instead of running ngx.var.request_uri through our
+      -- internal normalization mechanism
+      identifier = var.uri
```
Tieske, line 140: just checking: `var.uri` is just the path right? it doesn't include the query args?

Tieske (commented):
Why would one use a `path` as the hash input? that doesn't make sense to me. Wouldn't it be better to enable using a path-parameter?

Say I have 20 paths defined, all starting like this; `/myapp/[userid]/...`. Now the cardinality of the hash is high because there are 20 individual paths, each getting their own hash.

If we'd use just the captured `[userid]` element, we'd ensure that all traffic related to a single user would end up on a single backend. This would make more sense imo.

That said, I don't know the origin of this request.

flrgh:
> Why would one use a `path` as the hash input? that doesn't make sense to me. Wouldn't it be better to enable using a path-parameter?
> 
> Say I have 20 paths defined, all starting like this; `/myapp/[userid]/...`. Now the cardinality of the hash is high because there are 20 individual paths, each getting their own hash.
> 
> If we'd use just the captured `[userid]` element, we'd ensure that all traffic related to a single user would end up on a single backend. This would make more sense imo.

Agree 100%

> That said, I don't know the origin of this request.

Ah, you're one step ahead of me in my explanation 😉. This request comes from an enterprise customer who has explicitly named `path` as one of their use-cases.

flrgh:
@Tieske I've added router URI capture support :)

The work for supporting `path` is already done and it has minimal impact (no new entity fields/migrations like query/URI capture), so I've left it in for now. It's probably not as useful as URI capture but IMO still has valid use cases.

kong/db/schema/entities/upstreams.lua
```diff
@@ -186,6 +190,10 @@ local r =  {
     { hash_fallback_header = typedefs.header_name, },
     { hash_on_cookie = { type = "string",  custom_validator = utils.validate_cookie_name }, },
     { hash_on_cookie_path = typedefs.path{ default = "/", }, },
+    { hash_on_query_arg = query_arg },
+    { hash_fallback_query_arg = query_arg },
+    { hash_on_uri_capture = { type = "string", len_min = 1 }, },
+    { hash_fallback_uri_capture = { type = "string", len_min = 1 }, },
```
Tieske, line 196: can we rename `query_arg` to `simple_param` (or similar) and reuse that typedef also for the 2 new uri fields? constraints are the same but query_arg gets a typedef, whilst the others aredefined in line

Tieske:
> The work for supporting path is already done and it has minimal impact (no new entity fields/migrations like query/URI capture), so I've left it in for now. It's probably not as useful as URI capture but IMO still has valid use cases.

What usecase would they be useful for?

Tieske (approved):
still not convinced about the `path` option, I'm afraid it's bloat. Then again it's minimal and isolated in a specific corner of the code. So not that big a deal.

jschmid1 (approved):

flrgh:
rebased onto latest master + fixed merge conflicts

flrgh:
Kong moderators: please hold off on merge for one more rebase+test cycle. I didn't think we'd be upgrading to OpenResty 1.21.4 in this release cycle, so I'm going to remove the version gate on [this perf optimization](https://github.com/Kong/kong/blob/d519d813234e287bc927473c8cf217f98cfa3c98/kong/runloop/balancer/init.lua#L53-L61) and rebase+re-push. One moment...

kong/db/schema/entities/upstreams.lua
```diff
@@ -186,6 +190,10 @@ local r =  {
     { hash_fallback_header = typedefs.header_name, },
     { hash_on_cookie = { type = "string",  custom_validator = utils.validate_cookie_name }, },
     { hash_on_cookie_path = typedefs.path{ default = "/", }, },
+    { hash_on_query_arg = simple_param },
+    { hash_fallback_query_arg = simple_param },
+    { hash_on_uri_capture = simple_param },
+    { hash_fallback_uri_capture = simple_param },
```
aboudreault, line 196: I think we will need compatibility version fix for this, is it exported to DPs ?

flrgh: Hmm, probably do need some logic to clear these out, but only if they are `null` (otherwise we're just shipping a broken upstream to the DP).

aboudreault: if they have a value, the entity will break if sent to old DPs due to the unknown field, right?

flrgh: Right, if they have a value then that means that `hash_on` is probably set to one of the new values (`path`, `query_arg`, `uri_capture`), in which case there's not really any way to make the upstream compatible with the DP.

aboudreault: I see, okay

aboudreault:
what's left to do will be in EE so you can merge

PR #8705, opened 2022-04-20
fix(revert): Revert "feat(dao) use `cache_key` for target uniqueness detection"

hutchic:
This reverts commit 9eba2a1e4b78711def54c3ea5096634d7760ba06.

To fix for
```
Error: [off error] [off error] failed to load migration 'kong.db.migrations.core.016_280_to_300' of 'core' subsystem

```


kikito (approved):

kikito:
Reverts #8179

flrgh:
FYI @yankun-li-kong it looks like the source of the issue was that the new migration file was not added to the kong rockspec file.

mflendrich:
@kikito, we just noticed that #8179 is in the "Unreleased" changelog but this revert PR is not reflected. Is that intentional?

kikito:
@mflendrich I am removing the change from the changelog and notifying its author

PR #8707, opened 2022-04-20
test(packaging): do a quick validation that Kong can viably be packaged

hutchic:
Some recent PR's have broken our ability to package / run Kong.

This PR uses our packaging / release tooling to do a single test run.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,6 +19,24 @@ pipeline {
         DEBUG = 0
     }
     stages {
+        stage('Test The Package') {
+            agent {
+                node {
+                    label 'bionic'
+                }
+            }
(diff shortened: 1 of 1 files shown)


kikito (approved):

hutchic:
@kikito whoops I meant to add a `when` stanza to run this on PR after I validated by running it against this branch :laughing: 

I'll add the `when` back in as well as investigate why jenkins isn't reporting that it passed back onto this PR :thinking:

kikito (changes requested):
Alright, changing this to "Request Changes" then so no one merges by mistake.

kikito (approved):

PR #8708, opened 2022-04-21
perf(pdk) faster req.get_header implementation

Suika-Kong:
Support this by a patch to `ngx.req` and caching the get_headers() result.

I thought the fix can be put to `globalpatches.lua` but I don't know why it doesn't work. Please suggest a better place to put the patch.

Also, a better solution would be a real implementation of `ngx.req.get_header`.

DO NOT BE EAGER TO MERGE THIS PR!

Fix FT-2780

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -28,6 +28,33 @@ local PHASES = phase_checker.phases
 cjson.decode_array_with_array_mt(true)
 
 
+do -- add patch for faster ngx.req.get_header
+  local req_get_headers = ngx.req.get_headers
+  -- use ngx.ctx as weak cache key
+  local req_get_headers_cache = setmetatable({}, {__mode="k"})
+  local function req_get_headers_cached()
+    local ctx = ngx.ctx
+    local cache = req_get_headers_cache[ctx]
+    if not cache then
+      cache = req_get_headers()
+      req_get_headers_cache[ctx] = cache
+    end
+    return cache
+  end
+
+  -- clear the cache if header changed
+  local origin_set_header = ngx.req.set_header
+  function ngx.req.set_header(header_name, header_value)
+    req_get_headers_cache[ngx.ctx] = nil
+    return origin_set_header(header_name, header_value)
+  end
+
+  function ngx.req.get_header(name)
+    return req_get_headers_cached()[name]
+  end
+end
+
+
 local function new(self)
   local _REQUEST = {}
 
@@ -568,7 +595,7 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = _REQUEST.get_headers()[name]
+    local header_value = ngx.req.get_header(name)
     if type(header_value) == "table" then
       return header_value[1]
     end
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -44,12 +44,12 @@ do -- add patch for faster ngx.req.get_header
 
   -- clear the cache if header changed
   local origin_set_header = ngx.req.set_header
-  function ngx.req.set_header(header_name, header_value)
+  function ngx.req.set_header(header_name, header_value) -- luacheck: ignore
     req_get_headers_cache[ngx.ctx] = nil
     return origin_set_header(header_name, header_value)
   end
 
-  function ngx.req.get_header(name)
+  function ngx.req.get_header(name) -- luacheck: ignore
     return req_get_headers_cached()[name]
   end
 end
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -31,7 +31,7 @@ cjson.decode_array_with_array_mt(true)
 do -- add patch for faster ngx.req.get_header
   local req_get_headers = ngx.req.get_headers
   -- use ngx.ctx as weak cache key
-  local req_get_headers_cache = setmetatable({}, {__mode="k"})
+  local req_get_headers_cache = setmetatable({}, {__mode = "k",})
   local function req_get_headers_cached()
     local ctx = ngx.ctx
     local cache = req_get_headers_cache[ctx]
@@ -55,6 +55,7 @@ do -- add patch for faster ngx.req.get_header
 end
 
 
+local ngx_req_get_header = ngx.req.get_header
 local function new(self)
   local _REQUEST = {}
 
@@ -595,7 +596,7 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = ngx.req.get_header(name)
+    local header_value = ngx_req_get_header(name)
     if type(header_value) == "table" then
       return header_value[1]
     end
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -55,7 +55,7 @@ do -- add patch for faster ngx.req.get_header
 end
 
 
-local ngx_req_get_header = ngx.req.get_header
+local ngx_req_get_header = ngx.req.get_header -- luacheck: ignore
 local function new(self)
   local _REQUEST = {}
 


ADD-SP:
Why can't these functions be hacked in `globalpatches.lua`? How do you hook these functions in `globalpatches.lua`.

kong/pdk/request.lua
```diff
@@ -28,6 +28,33 @@ local PHASES = phase_checker.phases
 cjson.decode_array_with_array_mt(true)
 
 
+do -- add patch for faster ngx.req.get_header
+  local req_get_headers = ngx.req.get_headers
+  -- use ngx.ctx as weak cache key
+  local req_get_headers_cache = setmetatable({}, {__mode="k"})
```
chronolaw, line 34: needs some space, like `__mode = "k"`,
also, there should be a `,` at the end of each item.

Tieske:
This looks like a slippery slope to me. I think we should make an overall design decision on where the request data lives. Currently some data lives in Nginx, and some in Lua land (router output for example). From the FT it seems @bungle also had questions on this.

Instead of individual optimizations, I'd rather see we create a request-object in Lua-land that reads the nginx data on demand and acts as the source of truth. Only at the end of a phase we push the changes back to nginx.

The PDK functions could become methods on that object which would also reduce the need to keep fetching the `ngx.ctx` magic table everywhere.

bungle:
I don't think this is a `fix`. This is more like `perf`.

bungle:
> This looks like a slippery slope to me. I think we should make an overall design decision on where the request data lives. Currently some data lives in Nginx, and some in Lua land (router output for example). From the FT it seems @bungle also had questions on this.

Yes. This is going to have fundamental consequences. Monkey patching any APIs in general is bad, but sometimes unavoidable. `ngx.req.get_header` and `ngx.req.set_header` are far from only things that may affect this. There is also `ngx.var.http_*`, `ngx.location.capture`, `ngx.req.clear_header`, and some others.

> Instead of individual optimizations, I'd rather see we create a request-object in Lua-land that reads the nginx data on demand and acts as the source of truth. Only at the end of a phase we push the changes back to nginx.

Yes, this would then have similar issue that the API works only as far as `ngx.*` APIs are not used at the same time, or we need again monkey patch them. This is a tricky issue and kinda fundamental. Aka do we want to diverge from ngx.* API and make it second class citizen (may affect a lot of core code, plugins, custom plugins)?

> The PDK functions could become methods on that object which would also reduce the need to keep fetching the `ngx.ctx` magic table everywhere.

I think PDK usually has `self`, so instead of using `ngx.ctx` you should use`self.ctx.shared`, still probably not as good as passing `ctx` to functions, but we don't have mechanism to pass `ctx` to functions currentle, e.g. plugin handlers, so plugins tend to call `kong.ctx.plugin` or `kong.ctx.shared`  or `ngx.ctx`.

chronolaw:
How about nginx var `$request_id`? Will it be cheaper than `ngx.ctx`?

Suika-Kong:
> This looks like a slippery slope to me. I think we should make an overall design decision on where the request data lives. Currently some data lives in Nginx, and some in Lua land (router output for example). From the FT it seems @bungle also had questions on this.
> 
> Instead of individual optimizations, I'd rather see we create a request-object in Lua-land that reads the nginx data on demand and acts as the source of truth. Only at the end of a phase we push the changes back to nginx.
> 
> The PDK functions could become methods on that object which would also reduce the need to keep fetching the `ngx.ctx` magic table everywhere.

I agree. And it's better done from the OpenResty side (So every modification is tracked).

Suika-Kong:
> How about nginx var `$request_id`? Will it be cheaper than `ngx.ctx`?

I doubt that `ngx.var` is also costly.

Suika-Kong:
> This looks like a slippery slope to me. I think we should make an overall design decision on where the request data lives. Currently some data lives in Nginx, and some in Lua land (router output for example). From the FT it seems @bungle also had questions on this.
> 
> Instead of individual optimizations, I'd rather see we create a request-object in Lua-land that reads the nginx data on demand and acts as the source of truth. Only at the end of a phase we push the changes back to nginx.
> 
> The PDK functions could become methods on that object which would also reduce the need to keep fetching the `ngx.ctx` magic table everywhere.

And there is a tricky problem. `get_headers` has parameters to limit the number of headers being parsed(which may be a dos point), what should we do to the limit for the Lua-land request-object? And when interfering with the handling of cases insensitive, this may cause a change of behavior.

kong/pdk/request.lua
Suika-Kong, line 34:
```suggestion
  local req_get_headers_cache = setmetatable({}, {__mode = "k",})
```

kong/pdk/request.lua
```diff
@@ -568,7 +595,7 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = _REQUEST.get_headers()[name]
+    local header_value = ngx.req.get_header(name)
```
Suika-Kong, line 598:
```suggestion
    local header_value = ngx_req_get_header(name)
```

kong/pdk/request.lua
```diff
@@ -28,6 +28,33 @@ local PHASES = phase_checker.phases
...
+    req_get_headers_cache[ngx.ctx] = nil
+    return origin_set_header(header_name, header_value)
+  end
+
+  function ngx.req.get_header(name) -- luacheck: ignore
+    return req_get_headers_cached()[name]
+  end
+end
+
+
```
Suika-Kong, line 57:
```suggestion

local ngx_req_get_header = ngx.req.get_header
```

kong/pdk/request.lua
```diff
@@ -28,6 +28,34 @@ local PHASES = phase_checker.phases
...
+local ngx_req_get_header = ngx.req.get_header
```
Suika-Kong, line 58:
```suggestion
local ngx_req_get_header = ngx.req.get_header -- luacheck: ignore
```

fffonion:
Honestly I don't feel like the current approach is worth pursue. It's sure that our current `kong.request.get_header` has performance penalty. But in realty we do provide `kong.request.get_headers`. And if user do want to get same header for a second time, they should cache it on their own. So current approach doesn't make a difference in performance if user is calling `get_headers` on their own or us; so this becomes a premature optimization for users don't need to get a header again.

I would suggest to explore with a real `ngx.req.get_header` approach, though I also doubt if that will be much more performant as you will need to iterrate as well in C land.

Suika-Kong:
> Honestly I don't feel like the current approach is worth pursue. It's sure that our current `kong.request.get_header` has performance penalty. But in realty we do provide `kong.request.get_headers`. And if user do want to get same header for a second time, they should cache it on their own. So current approach doesn't make a difference in performance if user is calling `get_headers` on their own or us; so this becomes a premature optimization for users don't need to get a header again.
> 
> I would suggest to explore with a real `ngx.req.get_header` approach, though I also doubt if that will be much more performant as you will need to iterrate as well in C land.

I notice that we are using `get_header` in our code a lot. For example:
1. kong/error_handlers.lua:38
2. kong/pdk/request.lua:143

You can search with it and will find over 100 results in our code.

And to answer the performance difference question, for the current implementation we need to build a Lua table every time we inspect a header. This is a large cost for Lua GC (which can then cause a significant cost of CPU time).
Also, we can introduce optimization for common (known) headers like `Host`. They are pre-hashed.

fffonion:
Performance wise, if it's us (Kong) that are using this method to get header, it should be switched to ngx.var.http_HEADER and index it for now; the same could apply to `ngx.req.get_header` in the future, that if you know which headers are determined to be fetched, you can tell nginx to pre-fetch it.
I'm not so worried about those error handling code as they are not in a hot path. Adding a cache is a quick win for most cases, but is hard to get right here like Tieske mentioned above, that multiple source of truth exists. Instead of spending time and introduce uncertainty in Lua land, I would still suggest to properly implement this in C land, where the source of truth is only at Nginx. And benchmark it. It will also get chance to be contributed to upstream.

ADD-SP:
I think it is better to use `ngx.var.http_HEADER`.

If users use a certain request header frequently, we should remind them to use [`lua_kong_load_var_index `](https://github.com/Kong/lua-kong-nginx-module#lua_kong_load_var_index) to optimize performance.

Suika-Kong:
> I think it is better to use `ngx.var.http_HEADER`.
> 
> If users use a certain request header frequently, we should remind them to use [`lua_kong_load_var_index `](https://github.com/Kong/lua-kong-nginx-module#lua_kong_load_var_index) to optimize performance.
@ADD-SP @fffonion 
I also prefer the `lua_kong_load_var_index ` solution as it does less modification and is the safer option.
The problem is that `ngx.var.http_HEADER` does not have a limitation on the number of headers to parse, which may be vulnerable to DoS.
I'd like to consider a hard limit on the header number and checking the number before accepting any request.

Suika-Kong:
New solution: https://github.com/Kong/lua-kong-nginx-module/pull/39

Issue #7680, opened 2019-04-03
Incorrect interpretation in NGINX state


RyanSiu1995:
kong_nginx_http_current_connections{state="total"} is not representing the "total connections". Instead, it represents the total request handled by the NGINX. Within one accepted connection, there may be more than one request. 
The wording total is misleading to most of the developer. They may think it is a total connection instead of total request.

StarlightIbuki:
Dear contributor,
We're closing this issue as there hasn't been any update to it for a long time. If the issue is still relevant in the latest version, please feel free to reopen it. We're more than happy to revisit it again. Your contribution is greatly appreciated!
Please have a look at [our pledge to the community](https://github.com/Kong/kong/blob/master/COMMUNITY_PLEDGE.md) for more information.
Sincerely,
Kong Gateway Team

PR #8712, opened 2022-04-21
feat(plugin) Adjust Prometheus to be the reference Impl for metrics

hfukada:
Rework and Upgrade Prometheus Plugins Metrics and Labels.

### Summary
These Prometheus plugin will be the first class citizen for metrics collection in Kong 3.0 and beyond. This patch attempts to lay foundation for a good set of base metrics+labels, that can also be expanded upon in later releases

### Full changelog
* Latency has been split into 4 different metrics: `kong_latency_ms`, `upstream_latency_ms` and `request_latency_ms` (http) /`tcp_session_duration_ms` (stream). Buckets details below.
* Separate out Kong Latency Bucket values and Upstream Latency Bucket values. 
  * Kong Latency and Upstream Latency can operate at orders of magnitudes different. Separate out these buckets to reduce memory overhead.
* `consumer_status` removed.
* `request_count` and `consumer_status` have been merged into just `http_requests_total`. If the `per_consumer` config is set false, the `consumer` label will be empty.  If the `per_consumer` config is true, it will be filled. 
* `http_requests_total` has a new label `source`. set to either `exit`, `error` or `service`. https://docs.konghq.com/gateway/latest/pdk/kong.response/#kongresponseget_source
* New Metric: `node_info`. Single gauge set to 1 that outputs the node's id and kong version.
* All Memory metrics have a new label `node_id`
* `nginx_http_current_connections` merged with `nginx_stream_current_connection` to `nginx_current_connections`
* Plugin Version Bumped to 3.0

### DIFF
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -16,12 +16,13 @@ local stream_available, stream_api = pcall(require, "kong.tools.stream_api")
 
 local role = kong.configuration.role
 
-local DEFAULT_BUCKETS = { 1, 2, 5, 7, 10, 15, 20, 25, 30, 40, 50, 60, 70,
-                          80, 90, 100, 200, 300, 400, 500, 1000,
-                          2000, 5000, 10000, 30000, 60000 }
+local KONG_LATENCY_BUCKETS = { 1, 2, 5, 7, 10, 15, 20, 30, 50, 75, 100, 200, 500, 750, 1000}
+local UPSTREAM_LATENCY_BUCKETS = {25, 50, 80, 100, 250, 400, 700, 1000, 2000, 5000, 10000, 30000, 60000 }
+
 local metrics = {}
 -- prometheus.lua instance
 local prometheus
+local node_id = kong.node.get_id()
 
 -- use the same counter library shipped with Kong
 package.loaded['prometheus_resty_counter'] = require("resty.counter")
@@ -39,21 +40,20 @@ local function init()
   prometheus = require("kong.plugins.prometheus.prometheus").init(shm, "kong_")
 
   -- global metrics
-  if kong_subsystem == "http" then
-    metrics.connections = prometheus:gauge("nginx_http_current_connections",
-      "Number of HTTP connections",
-      {"state"})
-  else
-    metrics.connections = prometheus:gauge("nginx_stream_current_connections",
-      "Number of Stream connections",
-      {"state"})
-  end
+  metrics.connections = prometheus:gauge("nginx_connections_total",
+    "Number of connections by subsystem",
+    {"node_id", "subsystem", "state"})
+  metrics.nginx_requests_total = prometheus:gauge("nginx_requests_total",
+      "Number of requests total", {"node_id", "subsystem"})
   metrics.timers = prometheus:gauge("nginx_timers",
                                     "Number of nginx timers",
                                     {"state"})
   metrics.db_reachable = prometheus:gauge("datastore_reachable",
                                           "Datastore reachable from Kong, " ..
                                           "0 is unreachable")
+  metrics.node_info = prometheus:gauge("node_info",
+                                       "Kong Node metadata information",
+                                       {"node_id", "version"})
   -- only export upstream health metrics in traditional mode and data plane
   if role ~= "control_plane" then
     metrics.upstream_target_health = prometheus:gauge("upstream_target_health",
@@ -66,44 +66,60 @@ local function init()
   local memory_stats = {}
   memory_stats.worker_vms = prometheus:gauge("memory_workers_lua_vms_bytes",
                                              "Allocated bytes in worker Lua VM",
-                                             {"pid", "kong_subsystem"})
+                                             {"node_id", "pid", "kong_subsystem"})
   memory_stats.shms = prometheus:gauge("memory_lua_shared_dict_bytes",
                                        "Allocated slabs in bytes in a shared_dict",
-                                       {"shared_dict", "kong_subsystem"})
+                                       {"node_id", "shared_dict", "kong_subsystem"})
   memory_stats.shm_capacity = prometheus:gauge("memory_lua_shared_dict_total_bytes",
                                                "Total capacity in bytes of a shared_dict",
-                                               {"shared_dict", "kong_subsystem"})
+                                               {"node_id", "shared_dict", "kong_subsystem"})
 
   local res = kong.node.get_memory_stats()
   for shm_name, value in pairs(res.lua_shared_dicts) do
-    memory_stats.shm_capacity:set(value.capacity, { shm_name, kong_subsystem })
+    memory_stats.shm_capacity:set(value.capacity, { node_id, shm_name, kong_subsystem })
   end
 
   metrics.memory_stats = memory_stats
 
   -- per service/route
   if kong_subsystem == "http" then
-    metrics.status = prometheus:counter("http_status",
-                                        "HTTP status codes per service/route in Kong",
-                                        {"service", "route", "code"})
+    metrics.status = prometheus:counter("http_requests_total",
+                                        "HTTP status codes per consumer/service/route in Kong",
+                                        {"service", "route", "code", "source", "consumer"})
   else
     metrics.status = prometheus:counter("stream_status",
-                                        "Stream status codes per service/route in Kong",
-                                        {"service", "route", "code"})
+                                        "Stream status codes per consumer/service/route in Kong",
+                                        {"service", "route", "code", "source"})
   end
-  metrics.latency = prometheus:histogram("latency",
-                                         "Latency added by Kong, total " ..
-                                         "request time and upstream latency " ..
-                                         "for each service/route in Kong",
-                                         {"service", "route", "type"},
-                                         DEFAULT_BUCKETS) -- TODO make this configurable
-  metrics.bandwidth = prometheus:counter("bandwidth",
-                                         "Total bandwidth in bytes " ..
-                                         "consumed per service/route in Kong",
-                                         {"service", "route", "type"})
-  metrics.consumer_status = prometheus:counter("http_consumer_status",
-                                          "HTTP status codes for customer per service/route in Kong",
-                                          {"service", "route", "code", "consumer"})
+  metrics.kong_latency = prometheus:histogram("kong_latency_ms",
+                                              "Latency added by Kong and enabled plugins " ..
+                                              "for each service/route in Kong",
+                                              {"service", "route"},
+                                              KONG_LATENCY_BUCKETS)
+  metrics.upstream_latency = prometheus:histogram("upstream_latency_ms",
+                                                  "Latency added by upstream response " ..
+                                                  "for each service/route in Kong",
+                                                  {"service", "route"},
+                                                  UPSTREAM_LATENCY_BUCKETS)
+
+
+  if kong_subsystem == "http" then
+    metrics.total_latency = prometheus:histogram("request_latency_ms",
+                                                 "Total latency incurred during requests " ..
+                                                 "for each service/route in Kong",
+                                                 {"service", "route"},
+                                                 UPSTREAM_LATENCY_BUCKETS)
+  else
+    metrics.total_latency = prometheus:histogram("tcp_session_duration_ms",
+                                                 "latency incurred in stream tcp session " ..
+                                                 "for each service/route in Kong",
+                                                 {"service", "route"},
+                                                 UPSTREAM_LATENCY_BUCKETS)
+  end
+  metrics.bandwidth = prometheus:counter("bandwidth_bytes",
+                                         "Total bandwidth (ingress/egress) " ..
+                                         "throughput in bytes",
+                                         {"service", "route", "direction", "consumer"})
 
   -- Hybrid mode status
   if role == "control_plane" then
@@ -146,8 +162,9 @@ end
 
 -- Since in the prometheus library we create a new table for each diverged label
 -- so putting the "more dynamic" label at the end will save us some memory
-local labels_table = {0, 0, 0}
-local labels_table4 = {0, 0, 0, 0}
+local labels_table = {0, 0, 0, 0}
+local labels_table_status = {0, 0, 0, 0, 0}
+local latency_labels_table = {0, 0}
 local upstream_target_addr_health_table = {
   { value = 0, labels = { 0, 0, 0, "healthchecks_off", ngx.config.subsystem } },
   { value = 0, labels = { 0, 0, 0, "healthy", ngx.config.subsystem } },
@@ -190,10 +207,26 @@ if kong_subsystem == "http" then
       route_name = message.route.name or message.route.id
     end
 
+    local consumer = ""
+    if message and serialized.consumer ~= nil then
+      consumer = serialized.consumer
+    end
+
     labels_table[1] = service_name
     labels_table[2] = route_name
     labels_table[3] = message.response.status
-    metrics.status:inc(1, labels_table)
+    labels_table[4] = consumer
+
+    labels_table_status[1] = service_name
+    labels_table_status[2] = route_name
+    labels_table_status[3] = message.response.status
+    labels_table_status[4] = kong.response.get_source()
+    labels_table_status[5] = consumer
+
+    latency_labels_table[1] = service_name
+    latency_labels_table[2] = route_name
+
+    metrics.status:inc(1, labels_table_status)
 
     local request_size = tonumber(message.request.size)
     if request_size and request_size > 0 then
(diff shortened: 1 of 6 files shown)


kong/plugins/prometheus/exporter.lua
```diff
@@ -41,21 +42,18 @@ local function init()
...
-  if kong_subsystem == "http" then
-    metrics.connections = prometheus:gauge("nginx_http_current_connections",
-      "Number of HTTP connections",
-      {"state"})
-  else
-    metrics.connections = prometheus:gauge("nginx_stream_current_connections",
-      "Number of Stream connections",
-      {"state"})
-  end
+  metrics.connections = prometheus:gauge("nginx_current_connections",
```
hbagdi, line 45:
```suggestion
  metrics.connections = prometheus:gauge("nginx_connections",
```
The `state` label can be `total`, `handled` or `accepted`, `active`, `waiting`, `reading`, `writing`. This is a mix of guages and couners. It is best to not put `current_` in here since that has potential for added confusion.

kong/plugins/prometheus/exporter.lua
```diff
@@ -258,10 +275,34 @@ else
...
+    latency_labels_table[2] = route_name
+
+    labels_table_status[1] = service_name
+    labels_table_status[2] = route_name
+    labels_table_status[3] = message.session.status
+    labels_table_status[5] = consumer
+
+    -- If number of upstream requests is greater than 0, assume upstream is the status code originator
+    -- Else, assume kong is the originator.
+    if(#(message.tries) > 0) then
```
hbagdi, line 299: Is this assumption correct? What if the balancer tries up to a `max`, gives up and then Kong responds back with a 5xx?
@locao @fffonion ?

kong/plugins/prometheus/exporter.lua
```diff
@@ -68,44 +66,50 @@ local function init()
...
+                                              "Latency added by Kong and enabled plugins " ..
+                                              "for each service/route in Kong",
+                                              {"service", "route"},
+                                              KONG_LATENCY_BUCKETS)
+  metrics.upstream_latency = prometheus:histogram("upstream_latency",
+                                                  "Latency added by upstream response " ..
+                                                  "for each service/route in Kong",
+                                                  {"service", "route"},
+                                                  UPSTREAM_LATENCY_BUCKETS)
+  metrics.total_latency = prometheus:histogram("total_latency",
```
hbagdi, line 104: Should this be called "request_latency"?
"total" has well-known semantics in Prometheus and we should avoid using a generic term here. WDYT?

spec/03-plugins/26-prometheus/02-access_spec.lua
```diff
@@ -478,11 +479,10 @@ describe("Plugin: prometheus (access) per-consumer metrics", function()
...
-      return body:find('kong_http_status{service="mock-service",route="http-route",code="200"} 1', nil, true)
+      return body:find('kong_request_count{service="mock-service",route="http-route",code="200",source="upstream",consumer="alice"} 1', nil, true)
     end)
 
-    assert.not_match('kong_http_consumer_status{service="mock-service",route="http-route",code="401",consumer="alice"} 1', body, nil, true)
-    assert.matches('kong_http_status{service="mock-service",route="http-route",code="401"} 1', body, nil, true)
+    assert.matches('kong_request_count{service="mock-service",route="http-route",code="401",source="kong",consumer=""} 1', body, nil, true)
 
     assert.matches('kong_nginx_metric_errors_total 0', body, nil, true)
   end)
```
hbagdi, line 488: Can a test assert that kong_latency, request_latency, upstream_latency time series are present in the output?

hbagdi (commented):
This is mostly LGTM. @fffonion Could you do another review as well, please.

kong/plugins/prometheus/exporter.lua
hfukada: I find "request_latency" confusing, because to me I can't tell which part of the "request" it incorporates.

kong/plugins/prometheus/exporter.lua
hfukada: I understand the desire to be 100% correct, but I think in a system with arbitrary plugin code execution this is hard to get 100% correct.

Do you think this is good enough for a first pass? 

@sichvoge do you have thoughts on the matter?

kong/plugins/prometheus/exporter.lua
hfukada: I don't feel strongly on wording/naming here. @sichvoge do you?

Also I don't see any counters here? Am I missing some context?

kong/plugins/prometheus/exporter.lua
hbagdi: total_request_latency?

kong/plugins/prometheus/exporter.lua
hbagdi: Please confirm this with @locao before merging.

kong/plugins/prometheus/exporter.lua
hbagdi: The total, handled, and accepted are counters (they only go up) but are implemented as guages.

kong/plugins/prometheus/exporter.lua
hfukada: good split :+1: Committed.

spec/03-plugins/26-prometheus/02-access_spec.lua
hfukada: Added :+1:

kong/plugins/prometheus/exporter.lua
fffonion: While we are making breking changes to names, I would suggest to do https://github.com/Kong/kong/pull/8766 (as suggested by #7680) and make the `total` a seperate metrics, like `nginx_requests_total`.

fffonion:
When we are renaming metrics, should we take this as chance to compliant with Prometheus as suggested by https://github.com/Kong/kong/discussions/8237?

kong/plugins/prometheus/exporter.lua
sichvoge: @hfukada I think you are actually doing that already in line 46 right?

kong/plugins/prometheus/exporter.lua
```diff
@@ -39,21 +40,24 @@ local function init()
...
+      "Number of requests accepted", {"node_id", "subsystem"})
+  metrics.nginx_requests_handled = prometheus:counter("nginx_requests_handled",
+      "Number of requests handled", {"node_id", "subsystem"})
   metrics.timers = prometheus:gauge("nginx_timers",
                                     "Number of nginx timers",
                                     {"state"})
   metrics.db_reachable = prometheus:gauge("datastore_reachable",
                                           "Datastore reachable from Kong, " ..
                                           "0 is unreachable")
+  metrics.node_info = prometheus:gauge("node_info",
```
sichvoge, line 58: For future use cases, should we actually add `ip` here at least as well? I don't think that it would introduce high-cardinality in this case since for a given node_id, the IP might always be the same correct?

kong/plugins/prometheus/exporter.lua
sichvoge: No particular thoughts to be honest since I am not knowledgeable enough about all the different cases. Just to be sure, we should double check with @locao as Harry suggested.

kong/plugins/prometheus/exporter.lua
hfukada: I think the thought here was to rely on some already default/automatically applied prometheus metrics/labels from the scrape to get that information

kong/plugins/prometheus/exporter.lua
fffonion: we can use `kong.response.get_source()` instead

kong/plugins/prometheus/exporter.lua
```diff
@@ -66,44 +66,50 @@ local function init()
...
   metrics.memory_stats = memory_stats
 
   -- per service/route
   if kong_subsystem == "http" then
-    metrics.status = prometheus:counter("http_status",
-                                        "HTTP status codes per service/route in Kong",
-                                        {"service", "route", "code"})
+    metrics.status = prometheus:counter("http_requests_total",
+                                        "HTTP status codes per consumer/service/route in Kong",
+                                        {"service", "route", "code", "source", "consumer"})
```
fffonion, line 88: we should avoid using terms that are only defined by this plugin, `"source"` should be either `"kong_subsystem"`. as it's subsystem is a term defined by nginx and to align with other metrics like `kong_memory_workers_lua_vms_bytes`.

kong/plugins/prometheus/exporter.lua
```diff
@@ -39,21 +40,20 @@ local function init()
...
-    metrics.connections = prometheus:gauge("nginx_http_current_connections",
-      "Number of HTTP connections",
-      {"state"})
-  else
-    metrics.connections = prometheus:gauge("nginx_stream_current_connections",
-      "Number of Stream connections",
-      {"state"})
-  end
+  metrics.connections = prometheus:gauge("nginx_connections_total",
+    "Number of connections by subsystem",
```
fffonion, line 44: same, use `"kong_subsystem"` instead of `"subsystem"` in label, or change of them to `"subsystem"`.

kong/plugins/prometheus/exporter.lua
hfukada: What should the label for "the place where we think the response code came from"

kong/plugins/prometheus/exporter.lua
```diff
@@ -66,44 +66,50 @@ local function init()
...
+                                              "Latency added by Kong and enabled plugins " ..
+                                              "for each service/route in Kong",
+                                              {"service", "route"},
+                                              KONG_LATENCY_BUCKETS)
+  metrics.upstream_latency = prometheus:histogram("upstream_latency_ms",
+                                                  "Latency added by upstream response " ..
+                                                  "for each service/route in Kong",
+                                                  {"service", "route"},
+                                                  UPSTREAM_LATENCY_BUCKETS)
+  metrics.total_latency = prometheus:histogram("total_request_latency_ms",
```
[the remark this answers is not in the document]
hfukada: `total_session_latency_ms`?

kong/plugins/prometheus/exporter.lua
hfukada: misunderstanding talked in slack :+1:

kong/plugins/prometheus/exporter.lua
```diff
@@ -66,44 +66,50 @@ local function init()
...
-                                        {"service", "route", "code"})
+    metrics.status = prometheus:counter("http_requests_total",
+                                        "HTTP status codes per consumer/service/route in Kong",
+                                        {"service", "route", "code", "source", "consumer"})
   else
     metrics.status = prometheus:counter("stream_status",
-                                        "Stream status codes per service/route in Kong",
-                                        {"service", "route", "code"})
+                                        "Stream status codes per consumer/service/route in Kong",
+                                        {"service", "route", "code", "source", "consumer"})
```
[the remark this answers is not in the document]
hfukada: removed :)

kong/plugins/prometheus/exporter.lua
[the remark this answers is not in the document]
fffonion: 🤔 my feeling is to align the name with its friends as possible, so if we have `total_` here, one might confuse with "is kong_latency` not "total kong_latency" and what's the other parts of `kong_latency`? usually we don't put `total_` in the latency names and use it for accumulative counters mostly. So I would suggest to just go with `session_latency_ms`. how do you think?

kong/plugins/prometheus/exporter.lua
```diff
@@ -190,10 +197,26 @@ if kong_subsystem == "http" then
...
     labels_table[2] = route_name
     labels_table[3] = message.response.status
-    metrics.status:inc(1, labels_table)
+    labels_table[4] = consumer
+
+    labels_table_status[1] = service_name
+    labels_table_status[2] = route_name
+    labels_table_status[3] = message.response.status
+    labels_table_status[4] = kong.response.get_source()
+    labels_table_status[5] = consumer
```
fffonion, line 214: as stream doesn't have consumer, you might want to set it to `nil` instead of passing a `""`, otherwise the prometheu library will cry

hfukada: this is for http subsystem.

kong/plugins/prometheus/exporter.lua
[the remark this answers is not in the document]
hfukada: updated to `session_latency_ms`

fffonion (approved):

fffonion:
Code LGTM, let's make sure this is recorded in changelog (https://github.com/Kong/kong/blob/master/CHANGELOG.md#plugins) especially to mention the breaking changes.

kong/plugins/prometheus/exporter.lua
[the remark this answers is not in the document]
hbagdi: -1 on session.
Sessions have a much different connotation than requests in our world. A session usually has multiple requests.
total_request_latency_ms is not the best name but it is better than session_latency_ms.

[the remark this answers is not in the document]
hfukada: okay. well. @sichvoge Please advise. Mr. Product, help us :)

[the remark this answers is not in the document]
fffonion: @hbagdi `session_latency` is for stream subsystem only, though there we can still talk about for example `TCP requests`, but that's less common compared to `TCP session`. It comes from the nginx `session_time` variable (http://nginx.org/en/docs/stream/ngx_stream_log_module.html).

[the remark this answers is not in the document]
sichvoge: Do we use one metric for both stream and http subsystem?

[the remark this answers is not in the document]
fffonion: Oh right... maybe we should seperate the two then 👍 Maybe `request_latency_ms` for http and `session_latency_ms` for stream, it feels ambigious to use one for the other.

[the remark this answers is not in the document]
hbagdi: tcp_session_duration_ms?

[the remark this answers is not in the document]
sichvoge: I like the idea separating them out. Let's do that. 

Therefore, if there aren't any significant objections, let's keep `request_latency_ms` for http and for streams, introduce a new metric and use Harry's suggestion: `tcp_session_duration_ms`.

@ryanmoore @hfukada I feel like we have consensus.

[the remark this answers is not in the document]
hfukada: Updated to split into `request_latency_ms` `tcp_session_duration_ms` for http/streams respectively.

kong/plugins/prometheus/exporter.lua
```diff
@@ -190,10 +207,26 @@ if kong_subsystem == "http" then
...
     labels_table[1] = service_name
     labels_table[2] = route_name
     labels_table[3] = message.response.status
-    metrics.status:inc(1, labels_table)
+    labels_table[4] = consumer
+
+    labels_table_status[1] = service_name
+    labels_table_status[2] = route_name
+    labels_table_status[3] = message.response.status
+    labels_table_status[4] = kong.response.get_source()
```
hbagdi, line 223: Please read the documentation of the function: https://docs.konghq.com/gateway/latest/pdk/kong.response/#kongresponseget_source.
Please parse the string and then populate the source accordingly.

hbagdi: The end-user of the metric will have no clue what is the difference between "exit" or "error". Avoid giving out details about the product as much as possible - the more this happens the more is cognitive load on the end user. Keep it simple.

kong/plugins/prometheus/exporter.lua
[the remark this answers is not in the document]
fffonion: let's just say `session_duration_ms` instead of `tcp_session_duration_ms`, as it can also be UDP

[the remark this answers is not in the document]
casimiro: but... does the concept of `session` make sense in the context of `UDP`? communication in `UDP` does not require a connection to be established between the peers. What would actually be a `session` in a `UDP` stream?

kong/plugins/prometheus/exporter.lua
casimiro: @hbagdi @sichvoge there are 3 possible values for `kong.response.get_source()`: `exit` (when the response originates in kong, e.g. due to invalid credentials), `error` (when the response originates in kong due to an error, e.g. unreachable upstream) and `service` (when response originates in the upstream).
given the above, are the translations below adequate?
`exit` -> `kong`
`error` -> `kong_due_to_error`
`service` -> `service`

hbagdi: exit -> kong
error -> kong
service -> service

Let's not doll out more information than necessary at the moment.

hbagdi: We can provide more fine-grained data to the user in the future if there is a need. When in doubt, start with the narrowest possible API.

kong/plugins/prometheus/exporter.lua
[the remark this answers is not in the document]
casimiro: hey @fffonion do you still think it's important to change `session_duration_ms` to `tcp_session_duration_ms`?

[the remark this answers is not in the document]
fffonion: @casimiro in UDP that variable will be the time taken after a datagram is received and before a response is sent. if we must explictly name this `tcp_session_duration_ms`, then we should also have a `udp_xxx` (or drop this metrics when protocol is UDP). either way, this is more about consistency, using `tcp_session_duration_ms` to describe both TCP and UDP is not correct. I would suggest we go with what Nginx have (since this data is directly read from a nginx variable after all), and keep the name aligned with the original nginx variable `$session_time`.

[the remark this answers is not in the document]
sichvoge: If we use the same metric for both, wouldn't we want to also have some label that identifies the protocol (either `udp` or `tcp`?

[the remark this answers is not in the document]
casimiro: changed the metric name to `session_duration_ms` given there isn't a simple way to distinguish between `UDP` stream and `TCP` stream.

casimiro (approved):

kong/plugins/prometheus/exporter.lua
```diff
@@ -66,44 +66,60 @@ local function init()
...
+
+  if kong_subsystem == "http" then
+    metrics.total_latency = prometheus:histogram("request_latency_ms",
+                                                 "Total latency incurred during requests " ..
+                                                 "for each service/route in Kong",
+                                                 {"service", "route"},
+                                                 UPSTREAM_LATENCY_BUCKETS)
+  else
+    metrics.total_latency = prometheus:histogram("session_duration_ms",
+                                                 "latency incurred in stream tcp session " ..
```
fffonion, line 114:
```suggestion
                                                 "latency incurred in stream session " ..
```

fffonion (approved):

PR #8713, opened 2022-04-21
tests(hybrid) mark test as flaky

kikito:
The first test in this group is flaky, and is delaying development in other areas.

Marking the whole group as flaky since the other test depends on the
first one, as a stopgap solution until we find a better way to do this.


aboudreault (approved):

dndx (approved):
We need a card to keep track of this issue so it gets addressed eventually.

kikito:
@dndx internal tracking code: FT-2799

PR #8716, opened 2022-04-22
perf(pdk) faster request.get_header

Suika-Kong:
Applying suggestions from OpenResty contributors. We use `ngx.var.http_*` to get the header.

Notice that the behavior changes. Now `request.get_header` does not have a limit of header number.

Fix FT-2780

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -12,6 +12,7 @@ local phase_checker = require "kong.pdk.private.phases"
 
 
 local ngx = ngx
+local var = ngx.var
 local sub = string.sub
 local find = string.find
 local lower = string.lower
@@ -568,10 +569,7 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = ngx.var["http_" .. lower(name:gsub("-", "_"))]
-    if type(header_value) == "table" then
-      return header_value[1]
-    end
+    local header_value = var["http_" .. lower(name:gsub("-", "_"))]
 
     return header_value
   end
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -52,6 +52,8 @@ local header_preprocess do
       local c = name:byte(idx)
       if c == hyphens then
         buffer[idx] = underscores
+      else
+        buffer[idx] = c
       end
     end
     return char(unpack(buffer))
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -21,7 +21,6 @@ local error = error
 local tonumber = tonumber
 local check_phase = phase_checker.check
 local check_not_phase = phase_checker.check_not
-local re_gsub = ngx.re.gsub
 
 
 local PHASES = phase_checker.phases
@@ -570,7 +569,8 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = var["http_" .. re_gsub(name, "-", "_", "jo")]
+    -- Do not localize ngx.re.gsub! It will crash because ngx.re is monkey patched.
+    local header_value = var["http_" .. ngx.re.gsub(name, "-", "_", "jo")]
 
     return header_value
   end
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -30,11 +30,6 @@ cjson.decode_array_with_array_mt(true)
 
 local header_preprocess do
   local char = string.char
-  local ffi = require "ffi"
-  local C = ffi.C
-  ffi.cdef [[
-    int tolower(int c);
-  ]]
   local unpack = unpack
   local clear_tab
   do
@@ -57,8 +52,6 @@ local header_preprocess do
       local c = name:byte(idx)
       if c == hyphens then
         buffer[idx] = underscores
-      else
-        buffer[idx] = C.tolower(c)
       end
     end
     return char(unpack(buffer))
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -28,6 +28,42 @@ local PHASES = phase_checker.phases
 
 cjson.decode_array_with_array_mt(true)
 
+local header_preprocess do
+  local char = string.char
+  local ffi = require "ffi"
+  local C = ffi.C
+  ffi.cdef [[
+    int tolower(int c);
+  ]]
+  local unpack = unpack
+  local clear_tab
+  do
+    local ok
+    ok, clear_tab = pcall(require, "table.clear")
+    if not ok then
+      clear_tab = function(tab)
+        for k in pairs(tab) do
+          tab[k] = nil
+        end
+      end
+    end
+  end
+  local buffer = {}
+  local hyphens = string.byte("-")
+  local underscores = string.byte("_")
+  header_preprocess = function (name)
+    clear_tab(buffer)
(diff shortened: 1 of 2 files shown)


t/01-pdk/04-request/13-get_header.t
```diff
@@ -104,7 +104,7 @@ X-Foo-Header: ''
 
 
 
-=== TEST 5: request.get_header() returns nil when requested header does not fit in default max_headers
+=== TEST 5: request.get_header() have no limit header numbers
```
Suika-Kong, line 107:
```suggestion
=== TEST 5: request.get_header() have no limit on header numbers
```

kong/pdk/request.lua
```diff
@@ -568,7 +568,7 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = _REQUEST.get_headers()[name]
+    local header_value = ngx.var["http_" .. lower(name:gsub("-", "_"))]
```
chronolaw, line 571: We should localize `ngx.var` at the top of the file.

Suika-Kong:
As Datong suggests, it's important that we have a limit on the number of fetched headers. So this solution is not applied.

kong/pdk/request.lua
```diff
@@ -568,10 +569,7 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = _REQUEST.get_headers()[name]
-    if type(header_value) == "table" then
-      return header_value[1]
-    end
+    local header_value = var["http_" .. lower(name:gsub("-", "_"))]
```
chronolaw, line 572: Should we use `ngx.re.gsub` instead of `string.gsub`?

Suika-Kong: You're replying at the speed of light!
Personally I think this can be done by simpler logic(rather than regex). Are there other alternatives to do that?

chronolaw: I think it is OK, maybe I am too prefer regex.

chronolaw: When I view the lua-nginx-module's code, I found ngx-lua has done the str_lower job, I think we can remove the call of lua `lower`.

You can do a simple test for it.

Suika-Kong: You're right.

kong/pdk/request.lua
```diff
@@ -12,6 +12,7 @@ local phase_checker = require "kong.pdk.private.phases"
 
 
 local ngx = ngx
+local var = ngx.var
 local sub = string.sub
 local find = string.find
 local lower = string.lower
```
chronolaw, line 18: Now we don't use `lower`, just remove it for luacheck.

kong/pdk/request.lua
```diff
@@ -27,6 +28,37 @@ local PHASES = phase_checker.phases
 
 cjson.decode_array_with_array_mt(true)
 
+local header_preprocess do
```
chronolaw, line 31: Is it faster than `string.gsub`? Perhaps it is a over design.

kong/pdk/request.lua
Suika-Kong: It's not here that `lower` is added. There're other uses of lower in the code.

Suika-Kong:
I've run a few benches like this:

```lua
local header_preprocess
do
    local char = string.char
    local unpack = unpack
    local clear_tab
    do
        local ok
        ok, clear_tab = pcall(require, "table.clear")
        if not ok then
            clear_tab = function(tab)
                for k in pairs(tab) do
                    tab[k] = nil
                end
            end
        end
    end
    local buffer = {}
    local hyphens = string.byte("-")
    local underscores = string.byte("_")
    header_preprocess = function(name)
        clear_tab(buffer)
        for idx = 1, #name do
            local c = name:byte(idx)
            if c == hyphens then
                buffer[idx] = underscores
            else
                buffer[idx] = c
            end
        end
        return char(unpack(buffer))
    end
end

local regsub = ngx.re.gsub
local gsub = string.gsub

local warmup_time = 100000
local times = 10000000
local headers = {
    "test",
    "test-a-v",
    "test-b",
    "test_b",
}
print(regsub(headers[1], "-", "_", "jo"))
print(regsub(headers[2], "-", "_", "jo"))
print(regsub(headers[3], "-", "_", "jo"))
print(regsub(headers[4], "-", "_", "jo"))

for i=1, warmup_time do
    header_preprocess(headers[1])
    header_preprocess(headers[2])
    header_preprocess(headers[3])
    header_preprocess(headers[4])
end

for i=1, warmup_time do
    gsub(headers[1], "-", "_")
    gsub(headers[2], "-", "_")
    gsub(headers[3], "-", "_")
    gsub(headers[4], "-", "_")
end

for i=1, warmup_time do
    regsub(headers[1], "-", "_", "djo")
    regsub(headers[2], "-", "_", "djo")
    regsub(headers[3], "-", "_", "djo")
    regsub(headers[4], "-", "_", "djo")
end

for i=1, warmup_time do
    regsub(headers[1], "-", "_", "jo")
    regsub(headers[2], "-", "_", "jo")
    regsub(headers[3], "-", "_", "jo")
    regsub(headers[4], "-", "_", "jo")
end

local start = os.clock()
for i=1, times do
    header_preprocess(headers[1])
    header_preprocess(headers[2])
    header_preprocess(headers[3])
    header_preprocess(headers[4])
end
print(os.clock() - start)

start = os.clock()
for i=1, times do
    gsub(headers[1], "-", "_")
    gsub(headers[2], "-", "_")
    gsub(headers[3], "-", "_")
    gsub(headers[4], "-", "_")
end
print(os.clock() - start)

start = os.clock()
for i=1, times do
    regsub(headers[1], "-", "_", "djo")
    regsub(headers[2], "-", "_", "djo")
    regsub(headers[3], "-", "_", "djo")
    regsub(headers[4], "-", "_", "djo")
end
print(os.clock() - start)

start = os.clock()
for i=1, times do
    regsub(headers[1], "-", "_", "jo")
    regsub(headers[2], "-", "_", "jo")
    regsub(headers[3], "-", "_", "jo")
    regsub(headers[4], "-", "_", "jo")
end
print(os.clock() - start)
```

The result is not very stable but I believe that resub with "jo" has the best performance.
Also, "d" has almost the worst performance. (DFA in PCRE has a really bad performance, so I even doubt the necessity of the "d" option。)

chronolaw:
Yes, we always use `jo` options in `ngx.re`, I never used `d`.

kong/pdk/request.lua
```diff
@@ -568,10 +569,8 @@ local function new(self)
       error("header name must be a string", 2)
     end
 
-    local header_value = _REQUEST.get_headers()[name]
-    if type(header_value) == "table" then
-      return header_value[1]
-    end
+    -- Do not localize ngx.re.gsub! It will crash because ngx.re is monkey patched.
```
chronolaw, line 572: I am not sure about it, but I found we do localize `ngx.re.gsub`, such as in `router.lua`.

Suika-Kong:
And to answer a question: `ngx.var.http_name` only returns a single header, will it have different behavior?

The original logic is like:
```lua
    if type(header_value) == "table" then
      return header_value[1]
    end
```

And `ngx.var.http_name` returns the first element it finds in the linked list. I think it will be the same as the original logic.

PR #8718, opened 2022-04-22
perf(pdk) faster request.get_header

Suika-Kong:
Almost the same as `ngx.var.http_*`(#8716 ), but use ffi to reduce the cost of string concatenation.

Fix FT-2780

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -27,6 +27,104 @@ local PHASES = phase_checker.phases
 
 cjson.decode_array_with_array_mt(true)
 
+local req_get_header do
+  local ffi = require("ffi")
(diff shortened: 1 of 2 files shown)


bungle:
Does this relate to this?
https://github.com/Kong/lua-kong-nginx-module/blob/master/lualib/resty/kong/var.lua

chronolaw:
I think you should keep only one necessary PR, close other repeated PR.

chronolaw:
I think you should keep only one necessary PR, close other repeated PR.

Suika-Kong:
As Datong suggests, it's important that we have a limit on the number of fetched headers. So this solution is not applied.

PR #8732, opened 2022-04-27
feat: add support for AssumeRoleWithWebIdentity for lambda plugin

backjo:
### Summary

This PR adds support for using IAM Roles for Service Account credentials within the AWS Lambda plugin. This allows users on EKS to invoke the lambda plugin without static credentials.

### DIFF
--- a/kong/plugins/aws-lambda/iam-eks-credentials.lua
+++ b/kong/plugins/aws-lambda/iam-eks-credentials.lua
@@ -1,0 +1,77 @@
+local kong = kong
+local ENV_TOKEN_FILE = os.getenv 'AWS_WEB_IDENTITY_TOKEN_FILE'
+local ENV_ROLE_ARN = os.getenv 'AWS_ROLE_ARN'
+local DEFAULT_SERVICE_REQUEST_TIMEOUT = 5000
+
+
+local url = require "socket.url"
+local http = require "resty.http"
+local json = require "cjson"
+local ngx_now = ngx.now
+local concat = table.concat
+local tostring = tostring
+
+do
+  if not (ENV_TOKEN_FILE) then
+    -- No variables found, so we're not probably not running on EKS containers
+    kong.log.debug("No Web Identity environment variables found for IAM")
+  end
+end
+
+
+local function fetchCredentials()
+  local client = http.new()
+  client:set_timeout(DEFAULT_SERVICE_REQUEST_TIMEOUT)
+
+  local tokenFile, tokenError = io.open(ENV_TOKEN_FILE)
+
+  if not tokenFile then
+    return nil, "Failed to open identity token file: " .. tostring(tokenError)
+  end
+
+  local response, err = client:request_uri("https://sts.amazonaws.com", {
+    method = "POST",
+    headers = {
+      ["Accept"] = "application/json",
+    },
+    query = {
+      Action = "AssumeRoleWithWebIdentity",
+      RoleArn = ENV_ROLE_ARN,
+      WebIdentityToken = tokenFile:read("*a"),
+      RoleSessionName = "kong",
+      Version = "2011-06-15",
+    },
+  })
+  if not response then
+    return nil, "Failed to request IAM credentials request returned error: " .. tostring(err)
+  end
+
+  if response.status ~= 200 then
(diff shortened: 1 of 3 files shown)


Tieske:
Functionality wise this looks good.

Implementation wise I'm wondering whether we should replace all the auth related stuff with the SDK, see https://github.com/Kong/lua-resty-aws . It features [full configuration support](https://github.com/Kong/lua-resty-aws/blob/main/src/resty/aws/config.lua) based on AWS CLI compatible settings, including profiles etc. As well as multiple authentications, most notably the [ProviderCredentialChain](https://github.com/Kong/lua-resty-aws/blob/main/src/resty/aws/credentials/CredentialProviderChain.lua#L46-L62).

@RobSerafini fyi; the aws sdk will be included in 3.0 as an underlying lib for accessing AWS Secrets Manager afaik, but @bungle probably knows best.

kong/plugins/aws-lambda/iam-eks-credentials.lua
```diff
@@ -0,0 +1,77 @@
...
+local ngx_now = ngx.now
+local concat = table.concat
+local tostring = tostring
+
+do
+  if not (ENV_TOKEN_FILE) then
+    -- No variables found, so we're not probably not running on EKS containers
+    kong.log.debug("No Web Identity environment variables found for IAM")
+  end
+end
```
flrgh, line 19: Idea: if we already know at `require`-time that no `AWS_WEB_IDENTITY_TOKEN_FILE` is set, then we might as well exit early:

```suggestion

if not ENV_TOKEN_FILE or not ENV_ROLE_ARN then
  -- No variables found, so we're not probably not running on EKS containers
  kong.log.debug("No Web Identity environment variables found for IAM")
  return { configured = false }
end

local url = require "socket.url"
local http = require "resty.http"
local json = require "cjson"
local ngx_now = ngx.now
local concat = table.concat
local tostring = tostring
```

Alternatively, you could also invert the logic:

```lua
local configured = false
local fetchCredentials

if ENV_TOKEN_FILE and ENV_ROLE_ARN then
  configured = true

  local url = require "socket.url"
  local http = require "resty.http"
  -- [...]

  function fetchCredentials()
    --  [...]
  end
end

return {
  configured = configured,
  fetchCredentials = fetchCredentials,
}
```

kong/plugins/aws-lambda/iam-eks-credentials.lua
```diff
@@ -0,0 +1,77 @@
...
+
+
+local function fetchCredentials()
```
flrgh, line 22: Do we expect the contents `ENV_TOKEN_FILE` to change? If not, then it's possible to just read it once and cache the result in a module-level var.

```lua
local pl_file = require "pl.file"
local token

local function fetchCredentials()
  if token == nil then
    local content, err = pl_file.read(ENV_TOKEN_FILE)
    token = content or false
    if not token then
      kong.debug.warn("failed reading ", ENV_TOKEN_FILE, ": ", err)
    end      
  end
  
  if not token then
    return nil, "no token"
  end
  
  local client = http.new()
  client:set_timeout(DEFAULT_SERVICE_REQUEST_TIMEOUT)

  -- [...]
end
```

And if we _do_ need to re-read `ENV_TOKEN_FILE` every time we're called, it's probably still better to move the file error-handling up top so that we're not creating a new http client if we don't have a token for it.

kong/plugins/aws-lambda/iam-eks-credentials.lua
```diff
@@ -0,0 +1,77 @@
...
+  local creds, err, ttl = fetchCredentials()
+  if creds then
+    return creds, err, ttl
+  end
+  kong.log.err(err)
+end
+
+return {
+  configured = not not ENV_TOKEN_FILE, -- force to boolean
+  fetchCredentials = fetchCredentialsLogged,
```
flrgh, line 76: Would be good to have some test coverage for this to go with the other credential providers.

backjo: Agreed! Will add these shortly - wanted to ensure that this was useful before adding the tests.

giovannidegani:
we have implemented this functionality as its own plugin https://github.com/lego/kong-aws-request-signing

PR #8735, opened 2022-04-28
Add resty.http timeouts to zipkin plugin

chazdnato:
### Summary

Our team recently experienced an outage due to our upstream OpenCensus collector going away in Kubernetes. The resulting Service still existed, which caused requests to the service to either receive "connection refused" or "connection timeout" errors. Each invocation of the zipkin plugin which had a timeout condition would consume an `ngx.timer`, eventually resulting in a saturation of available timers and causing a cascading effect (e.g. other component failures were occurring due to a lack of `ngx.timer` availability).

To help avoid this problem in the future, setting saner timeouts would allow orphaned zipkin trace reports to timeout sooner and return timers to the available pool. Note that we've opted to have different connect and send/recv timeouts. We can follow the model you have employed with other plugins (such as http-log or tcp-log) if you desire. We hope that the defaults picked are suitable for the community.

### Full changelog

* Add resty.http timeout options for zipkin plugin


kikito (changes requested):
Hello, thank you for this PR. Ideally it should come with tests, but I understand that testing timeouts in particular can be tricky.

Could you include an entry in the CHANGELOG.md file, in Unreleased/Additions/Plugins ?

chazdnato:
> Hello, thank you for this PR. Ideally it should come with tests, but I understand that testing timeouts in particular can be tricky.
> 
> Could you include an entry in the CHANGELOG.md file, in Unreleased/Additions/Plugins ?

It looks like you have some prior art in http-log so I'll add a test using that as a basis. I will also update the CHANGELOG.md file (thank you for the call out!). I'll have something submitted in the next day or so. Cheers :)

chazdnato:
> Hello, thank you for this PR. Ideally it should come with tests, but I understand that testing timeouts in particular can be tricky.
> 
> Could you include an entry in the CHANGELOG.md file, in Unreleased/Additions/Plugins ?

I've added some tests (based on existing http-log specs) and updated the CHANGELOG. Thanks!

chazdnato:
@RobSerafini 👋🏼  hoping you can help us get this moving, we seem to have stalled. Thank you!!

kong/plugins/zipkin/schema.lua
```diff
@@ -62,6 +62,8 @@ return {
           { static_tags = { type = "array", elements = static_tag,
                             custom_validator = validate_static_tags } },
           { http_span_name = { type = "string", required = true, default = "method", one_of = { "method", "method_path" } } },
+          { connect_timeout = {type = "number", default = 5000 } },
+          { byte_timeout = {type = "number", default = 10000 } }
```
mayocream, line 66: I think it makes more sense to introduce `connect_timeout`, `send_timeout`, `read_timeout`. 
`byte_timeout` may a bit confusing.

ref: https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/balancer.md#set_timeouts

chazdnato: Thank you, I have made these amendments.

kong/plugins/zipkin/schema.lua
```diff
@@ -62,6 +62,9 @@ return {
           { static_tags = { type = "array", elements = static_tag,
                             custom_validator = validate_static_tags } },
           { http_span_name = { type = "string", required = true, default = "method", one_of = { "method", "method_path" } } },
+          { connect_timeout = {type = "number", default = 5000 } },
+          { send_timeout = {type = "number", default = 10000 } },
+          { read_timeout = {type = "number", default = 10000 } },
```
mayocream, line 67: Some of our plugins use `2000` as the default timeout value, and considering the data size, the send_timeout and read_timeout should be larger than the default value; I think `5000` should be enough; what do you think?
```suggestion
          { connect_timeout = typedefs.timeout { default = 2000 } },
          { send_timeout = typedefs.timeout { default = 5000 } },
          { read_timeout = typedefs.timeout { default = 5000 } },
```

chazdnato: Works for me!

mayocream (approved):
LGTM!

chazdnato:
Thank you! What are next steps?

kong/plugins/zipkin/schema.lua
```diff
@@ -62,6 +62,9 @@ return {
           { static_tags = { type = "array", elements = static_tag,
                             custom_validator = validate_static_tags } },
           { http_span_name = { type = "string", required = true, default = "method", one_of = { "method", "method_path" } } },
+          { connect_timeout = typedefs.timeout { default = 2000 } },
+          { send_timeout = typedefs.timeout { default = 5000 } },
+          { read_timeout = typedefs.timeout { default = 5000 } },
```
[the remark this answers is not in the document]
chazdnato: Can you clarify this? I can help take this on, but unsure if this is a piece I can push forward. Thanks!

PR #8747, opened 2022-05-03
perf(core) remove coroutine in access plugins iterator

mayocream:
### Summary

Remove `coroutine` in access plugins iterator

WIP

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -97,7 +97,7 @@ jobs:
 
     - name: Run Tests
       env:
-        PERF_TEST_VERSIONS: git:${{ github.sha }},git:master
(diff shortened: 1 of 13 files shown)


Tieske:
This would be great. That coroutine has given me so many “attempt to yield across c call boundary” errors.

bungle:
> This would be great. That coroutine has given me so many “attempt to yield across c call boundary” errors.

Do you have an example? There is a big reason for that to exits, and it is to collect plugins even when one plugin said `kong.response.exit` or even `ngx.exit`.

Tieske:
@bungle Interactive debuggers, and don’t recall whether maybe luacov also had an issue with it

hbagdi:
Closing because of lack of activity. 2.7 is not receiving updates anymore.

Issue #8729, opened 2022-04-26
PR #6150 (fix for issue #5763) accidentally reverted by PR #8213 ?


marc-charpentier:
### Discussed in https://github.com/Kong/kong/discussions/8728

<div type='discussions-op-text'>

<sup>Originally posted by **marc-charpentier** April 26, 2022</sup>
Hello,

In a process of upgrading from Kong CE 1.5.1 to 2.8.0, I noticed a change in the _rate-limiting_ plugin's _redis_ policy's _increment_ function.

Before upgrading, it worked in two steps :
- one loop to build an array of cache keys and an array of expirations
- another loop to iterate over the arrays built in order to prepare
   a redis pipeline

After upgrading, the same two loops are nested, so that the inner one iterates several times over a still growing array, scheduling some increments in a Redis pipeline more than once. As the <code>red:init_pipeline()</code> is also executed several times, all previously scheduled increments are cleared, and only the last inner loop execution's increments are actually commited in the pipeline, avoiding counters' over-increments. But it seems not right from a performance's point of view, and a (major ?) bug may arise if the _init_pipeline_ function no longer clears previously scheduled queries in a future implementation.

I found in the plugin's history that this code's portion has changed several times, and that PR https://github.com/Kong/kong/pull/6150 (fix for issue https://github.com/Kong/kong/issues/5763) 's code was "reverted" by PR https://github.com/Kong/kong/pull/8213.

Is it really what was intended ?

If yes, was it for incompatibility between Redis' server-side _eval_ query and ACL support ?
If no, may I submit a PR in order to restore the fix for issue https://github.com/Kong/kong/issues/5763 ?

Thanks for your answers.
</div>

ADD-SP:
It looks like we accidentally introduced a bug. What do you think? @bungle

PR #8751, opened 2022-05-04
fix(plugin/rate-limiting) perform atomic redis increments 

ADD-SP:
### Summary

Please see [PR#6150](https://github.com/Kong/kong/pull/6150) .

### Issue reference

Fix [Issue#5763](https://github.com/Kong/kong/issues/5763)
Fix [Issue#8729](https://github.com/Kong/kong/issues/8729)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,11 @@
 
 ### Fixes
 
+#### Plugins
+
+- **rate-limiting**: Perform atomic redis increments
+  [#8751](https://github.com/Kong/kong/pull/8751).
+
 #### PDK
 
 - `pdk.response.set_header()`, `pdk.response.set_headers()`, `pdk.response.exit()` now ignore and emit warnings for manually set `Transfer-Encoding` headers.
--- a/kong/plugins/rate-limiting/policies/init.lua
+++ b/kong/plugins/rate-limiting/policies/init.lua
@@ -196,7 +196,7 @@ return {
           -- please see https://github.com/Kong/kong/issues/5763
           red:eval([[
             local key, value, expiration = KEYS[1], tonumber(ARGV[1]), ARGV[2]
-            if redis.call("incrby", key, value) == value then
+            if redis.call("incrby", key, value) == 1 then
               redis.call("expire", key, expiration)
             end
           ]], 1, cache_key, value, EXPIRATION[period])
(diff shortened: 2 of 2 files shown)


ADD-SP:
https://github.com/Kong/kong/runs/6286546136?check_suite_focus=true

Hi @javierguerragiraldez , I don't know why this test failed, would you mind helping me investigate?

yzyyi (changes requested):
It seems with the key expiration sets it broke the test here:
https://github.com/Kong/kong/blob/d569af657d553545847d3b6f1cc6eac6a73eab75/spec/03-plugins/23-rate-limiting/05-integration_spec.lua#L298-L318

kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -184,32 +184,22 @@ return {
...
-          red:init_pipeline()
-          for i = 1, idx do
-            red:incrby(keys[i], value)
-            if expiration[i] then
-              red:expire(keys[i], expiration[i])
+          -- use scripts to perform incr and expire operations atomically
+          -- please see https://github.com/Kong/kong/issues/5763
+          red:eval([[
+            local key, value, expiration = KEYS[1], tonumber(ARGV[1]), ARGV[2]
+            if redis.call("incrby", key, value) == value then
```
yzyyi, line 199: Do we want to expire the key if the value is 0?

CHANGELOG.md
```diff
@@ -66,6 +66,11 @@
 
 ### Fixes
 
+#### Plugins
```
flrgh, line 69: I believe plugin-specific changelog entries are [going away in 3.0](https://github.com/Kong/kong/pull/8772), so you will likely want to put this in the changelog at the repository root instead of here.

kong/plugins/rate-limiting/policies/init.lua
```diff
@@ -196,7 +196,7 @@ return {
           -- please see https://github.com/Kong/kong/issues/5763
           red:eval([[
             local key, value, expiration = KEYS[1], tonumber(ARGV[1]), ARGV[2]
-            if redis.call("incrby", key, value) == value then
+            if redis.call("incrby", key, value) == 1 then
```
yzyyi, line 199: Still I think this is implying that if key does not exists OR key value is 0, then we expire the key.
But do we want to expire a key that's value =0?
The test is comparing db size so that's probably why it's failing.

ADD-SP: Thanks for reminding me.
The old logic is that the expiration is set only when the key does not exist. But even if I do that, it doesn't solve this test.

PR #8758, opened 2022-05-05
feat(plugin/file-log) create immutable setting for top level file log paths

ADD-SP:
# We have decided not to include this one on the 3.0 milestone

### Summary

**BREAKING CHANGE**

A plugin should not have arbitrary access to the file system, so we should restrict the access of the `file-log` plugin to the file system.

This PR introduces a new configuration `plugin_file_log_path_prefix`, It restricts the scope of the `file-log` plugin to read and write to the file system. The default value is `{kong_prefix}/logs`.

### Issue reference

Fix FT-2115

### Flaky tests maybe caused by this PR

https://github.com/Kong/kong/runs/6316541320?check_suite_focus=true

### DIFF
--- a/kong/plugins/file-log/schema.lua
+++ b/kong/plugins/file-log/schema.lua
@@ -1,4 +1,22 @@
+local pl_path = require "pl.path"
 local typedefs = require "kong.db.schema.typedefs"
+
+local DEFAULT_PREFIX = pl_path.join(kong.configuration.prefix, "logs/")
+local prefix = kong.configuration.plugin_file_log_path_prefix or DEFAULT_PREFIX
+
+if string.sub(prefix, -1) ~= "/" then
+  prefix = prefix .. "/"
+end
+
+local path_pattern = string.format([[^%s[^*&%%\`]+$]], prefix)
+
+local err_msg = 
+  string.format("not a valid file name, "
+              .. "or the prefix is not [%s], "
+              .. "or contains `..`, "
+              .. "you may need to check the configureation "
+              .. "`plugin_file_log_path_prefix`",
+                 prefix)
 
 
 return {
@@ -10,8 +28,10 @@ return {
         fields = {
           { path = { type = "string",
                      required = true,
-                     match = [[^[^*&%%\`]+$]],
-                     err = "not a valid filename",
+                     match = path_pattern,
+                     -- to avoid the path traversal attack
+                     not_match = [[%.%.]],
+                     err = err_msg,
           }, },
           { reopen = { type = "boolean", required = true, default = false }, },
           { custom_fields_by_lua = typedefs.lua_code },
--- a/kong/plugins/file-log/schema.lua
+++ b/kong/plugins/file-log/schema.lua
@@ -14,7 +14,7 @@ local err_msg =
   string.format("not a valid file name, "
               .. "or the prefix is not [%s], "
               .. "or contains `..`, "
-              .. "you may need to check the configureation "
+              .. "you may need to check the configuration "
               .. "`plugin_file_log_path_prefix`",
                  prefix)
 
--- a/spec/03-plugins/04-file-log/02-schema_spec.lua
+++ b/spec/03-plugins/04-file-log/02-schema_spec.lua
@@ -1,40 +1,78 @@
 local Schema = require("kong.db.schema")
 
-describe("Plugin: file-log (schema)", function()
+local IMMUTABLE_PATH = "/tmp/"
 
+local ERR_MSG_INVLID_FILENAME =
+  string.format("not a valid file name, "
+              .. "or the prefix is not [%s], "
+              .. "or contains `..`, "
+              .. "you may need to check the configureation "
+              .. "`plugin_file_log_path_prefix`",
+                 IMMUTABLE_PATH)
+
+
+insulate("Plugin: file-log (schema)", function()
   local tests = {
     {
       name = "path is required",
       input = {
-        reopen = true
+        reopen = true,
       },
       output = nil,
       error = {
         config = {
-          path = "required field missing"
-        }
-      }
+          path = "required field missing",
+        },
+      },
     },
     ----------------------------------------
     {
       name = "rejects invalid filename",
       input = {
         path = "/ovo*",
-        reopen = true
+        reopen = true,
       },
       output = nil,
       error = {
         config = {
-          path = "not a valid filename"
-        }
-      }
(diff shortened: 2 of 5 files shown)


bungle:
In general I don't think it is scalable to introduce core changes for a single plugin use cases. Previously we have done this sometimes by introducing a plugin specific environment variable, e.g. AWS lambda can use normal AWS environment variables:

https://github.com/Kong/kong/blob/master/kong/plugins/aws-lambda/handler.lua#L17

What I did with `Vaults` is that I created a wildcard configuration:
https://github.com/Kong/kong/blob/master/kong/conf_loader/init.lua#L217-L220

We have other wildcards there too.

The logic is that now users can do:

```sh
# enable just the env vault
export KONG_VAULTS=env

# default configuration for the env vault prefix
export KONG_VAULT_ENV_PREFIX=SECRET_ 

# export postgres password in environment variable
export SECRET_PG_PASSWORD=pw

# reference password using env vault
# see, no need to specify `secret-` prefix here
export KONG_PG_PASSWORD={vault://env/pg-password}

# start kong
kong start
```

So as you can see there is:
```sh
KONG_VAULT_<vault-name>_*
```
available to all the Vaults. Not just for a single vault that might require some configuration. E.g. on EE you can:

```sh
KONG_VAULT_HCV_HOST=blaablaa.com
KONG_VAULT_HCV_TOKEN=xxxx
# or
KONG_VAULT_AWS_*=...
```

The Vaults use just `schema.lua` of Vaults to parse the config and infer it. This happens on PDK:
https://github.com/Kong/kong/blob/master/kong/pdk/vault.lua#L167-L186

You don't have schema as this `plugin_file_log_immutable_path` does not seem to be part of plugin configuration. It is part of plugin configuration validation. Perhaps it could be? Or you can do it otherwise. One problem I see with this config parameter is that it might differ from control plane and on each data plane. Which would render the valid configuration, invalid in some cases in some nodes (if they don't configure this config parameter the same).

I would try to use similar approach as described above and come up with something like:
```
KONG_PLUGIN_FILE_LOG_PATH_PREFIX=/tmp
```

And then on code side just prefix whatever there is in `config.path`. The code could look something like:

```lua
local DEFAULT_PREFIX = kong.configuration.prefix .. "/logs/"
local path = require "pl.path"
local log_path = path.join(kong.configuration.plugin_file_log_path_prefix or DEFAULT_PREFIX, conf.path)
```

What do you think? This way we can stop at least some of the plugin semantics leaking to core code or at least make it generic.

spec/03-plugins/04-file-log/02-schema_spec.lua
```diff
@@ -1,40 +1,78 @@
 local Schema = require("kong.db.schema")
 
-describe("Plugin: file-log (schema)", function()
+local IMMUTABLE_PATH = "/tmp/"
 
+local ERR_MSG_INVLID_FILENAME =
```
RobSerafini, line 5: Typo. `ERR_MSG_INVALID_FILENAME`

kong/plugins/file-log/schema.lua
```diff
@@ -10,8 +28,10 @@ return {
         fields = {
           { path = { type = "string",
                      required = true,
-                     match = [[^[^*&%%\`]+$]],
-                     err = "not a valid filename",
+                     match = path_pattern,
```
RobSerafini, line 31: I think what will happen is customers will be unable to update/patch existing file log plugins unless they make sure the existing config.path matches the new prefix.

We may want to have a migration/cli script that finds and prints out the existing plugins that need to be updated to follow the new rules. 

Let's discuss ways we can help customers who find themselves in this situation.

RobSerafini (changes requested):
Mainly I would like to disucss what migration or other tooling we might come up with to make this change smoother.

kong/plugins/file-log/schema.lua
dndx: Yes, this change seems to be a bit too much user impacting. As a matter of fact, more impacting to the sandboxing change we did a while ago. Need so find a better way to not surprise the customers after the fact.

kong/plugins/file-log/schema.lua
```diff
@@ -10,8 +28,10 @@ return {
...
+                     -- to avoid the path traversal attack
+                     not_match = [[%.%.]],
```
fffonion, line 33: don't disallow `..` at all, instead, calculate the path based on `..`. For example, if `plugin_file_log_path_prefix` is `/tmp/abc`, then `path` is
`/root/../tmp/abc`, should we allow it?

The other case is for symlinks, usually programs talks to local fs that has path restrictions also have a switch to follow/allow symlink or now. Consider there's a symlink
`/tmp/abc -> /root/abc/`, and `plugin_file_log_path_prefix` is `/tmp`, should you allow plugin write to `/tmp/abc/xxx`?

kikito: Options:
* Make it always concatenate with the global base. Then canonize (penlight has pl.path.normpath) and make sure that the resulting path is still inside the global base. Drawbacks: custom validations are not great when you are trying to implement a *different CP*.
* Not validate here at all, but make the save operation fail very noisily on the logs.

kikito (changes requested):
I think the original premise that originated this PR is not feasible without lots of effort:

> The Kong Gateway would only write files to specific locations

The "lots of effort" part comes from the fact that our plugins are not sandboxed. Custom plugins will always be able to break the original premise.

The "proper" way to implement this would be:
* Having a non-file-log specific list of "writable places" in the config
* sandboxing all the plugins so that io.open errors for files outside of the list. And there's no easy way around this (by using things like ffi.C).

That's a hard nut to crack.

Then there is what this PR is doing: concentrating on file-log only. In order to do this, I agree with @bungle that creating a config variable that begins with `plugin_file` is kind of ... dirty. His solution is less dirty, but not exactly clean either. I think this is because none of them is the "proper" one, presented above. We might have to live with that, my comments on this PR are on that direction.

A middle-ground alternative that has not been considered is the following:
* Add file-writing options to the PDK
* Rewrite file-log using pdk methods
* Add a non-file-log specific list of "writable places" in the config, but only *for when using the PDK to write files*. 

This is middleground because it covers more plugins than file-log, but not all.

Transversally to the options above, on the matter of backwards compatibility, I think we can follow the spirit of the initial request while remaining backwards compatible: make it opt-in. Make the global/env variable be empty by default, and make the security/validations not trigger at all when it is empty.

kikito:
We have decided not to include this one on the 3.0 milestone.

PR #8759, opened 2022-05-05
feat(zipkin) datadog headers

mayocream:
### Summary

Add support for DataDog headers in the Zipkin plugin. https://github.com/Kong/kong/discussions/8155

FTI-3282


kong/plugins/zipkin/tracing_headers.lua
```diff
@@ -314,6 +315,41 @@ local function parse_jaeger_trace_context_headers(jaeger_header)
 end
 
 
+local function parse_datadog_headers(headers)
+  -- allow testing to spy on this.
```
kikito, line 319: I think this comment is incorrect. The spy will work even if `warn` is not localized. It's just a little more efficient to localize it than to not localize it.
```suggestion
```

kong/plugins/zipkin/tracing_headers.lua
```diff
@@ -314,6 +315,41 @@ local function parse_jaeger_trace_context_headers(jaeger_header)
...
+
+  local parent_id = headers["x-datadog-parent-id"]
+  if parent_id and not parent_id:match("%d") then
+    warn("x-datadog-parent-id header invalid; ignoring.")
+    had_invalid_id = true
+  end
+
+  if had_invalid_id then
+    return nil, nil, should_sample
+  end
```
kikito, line 347: I don't understand why had_invalid_id is needed here. Is it only to have 2 warnings in case both the traceid and the parent id are present but not integers? Seems like on that case the traceid warning alone would do just fine.

```suggestion
  local trace_id = headers["x-datadog-trace-id"]
  if trace_id and not trace_id:match("%d") then
    warn("x-datadog-trace-id header invalid; ignoring.")
    return nil, nil, should_sample
  end

  local parent_id = headers["x-datadog-parent-id"]
  if parent_id and not parent_id:match("%d") then
    warn("x-datadog-parent-id header invalid; ignoring.")
    return nil, nil, should_sample
  end
```

kong/plugins/zipkin/tracing_headers.lua
```diff
@@ -314,6 +315,41 @@ local function parse_jaeger_trace_context_headers(jaeger_header)
...
+  -- allow testing to spy on this.
+  local warn = kong.log.warn
+
+  local should_sample = headers["x-datadog-sampling-priority"]
+  if should_sample == "1" or should_sample == "2" then 
+    should_sample = true
+  elseif should_sample == "0" or should_sample == "-1" then
+    should_sample = false
+  elseif should_sample ~= nil then
+    warn("x-datadog-sampling-priority header invalid; ignoring.")
```
kikito, line 328: Wouldn't it make more sense to not register this trace if the sampling priority header was incorrect?
```suggestion
    warn("x-datadog-sampling-priority header invalid; ignoring.")
    return nil, nil, nil
```

kikito (changes requested):

kong/plugins/zipkin/tracing_headers.lua
mayocream: It was a mistake, thanks!

kong/plugins/zipkin/tracing_headers.lua
```diff
@@ -314,6 +315,35 @@ local function parse_jaeger_trace_context_headers(jaeger_header)
...
+    return nil, nil, nil
+  end
+
+  local trace_id = headers["x-datadog-trace-id"]
```
yzyyi, line 331: nit: Other headers are named: "ot-tracer-traceid""x-b3-traceid"
Shall we keep naming convention here?

kong/plugins/zipkin/tracing_headers.lua
```diff
@@ -314,6 +315,35 @@ local function parse_jaeger_trace_context_headers(jaeger_header)
...
+  if trace_id and not trace_id:match("%d") then
+    warn("x-datadog-trace-id header invalid; ignoring.")
+    return nil, nil, should_sample
+  end
+
+  local parent_id = headers["x-datadog-parent-id"]
```
yzyyi, line 337: same with above "x-b3-parentspanid"

kong/plugins/zipkin/tracing_headers.lua
mayocream: The header name is defined by DataDog, we should keep this.

yzyyi: sounds good.

yzyyi (approved):

mayocream:
After discussed offline, we decided that this PR won't be merged into CE.

PR #8760, opened 2022-05-06
fix(clustering) Handle connection error on version negotiation

javierguerragiraldez:
If the CP doesn't respond to the version negotiation request, the DP
will assume the "old" protocol.

A failure to establish the websocket connection repeats the negotiation
attempt, as it would probably indicate a control plane rebooting or real
connectivity issues.  Later errors would (first) attempt only reconnect,
as they would likely be transient issue.  But if that reconnection
fails, it would again retry to renegotiate.

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -205,14 +205,34 @@ local function fill_empty_hashes(hashes)
   end
 end
 
-function _M:request_version_negotiation()
+
+function _M:call_control_plane(premature)
+  if premature then
+    return
+  end
+
   local response_data, err = version_negotiation.request_version_handshake(self.conf, self.cert, self.cert_key)
   if not response_data then
-    ngx_log(ngx_ERR, _log_prefix, "error while requesting version negotiation: " .. err)
-    assert(ngx.timer.at(math.random(5, 10), function(premature)
-      self:communicate(premature)
-    end))
-    return
+    ngx_log(ngx_ERR, _log_prefix, "error while requesting version negotiation: ", err)
+    ngx_log(ngx_ERR, _log_prefix, "Assuming old JSON protocol.")
+  end
+
+  local config_proto, msg = version_negotiation.get_negotiated_service("config")
+  if not config_proto and msg then
+    ngx_log(ngx_ERR, _log_prefix, "error reading negotiated \"config\" service: ", msg)
+  end
+
+  ngx_log(ngx_DEBUG, _log_prefix, "config_proto: ", tostring(config_proto), " / ", tostring(msg))
+  if config_proto == "v1" then
+    self.child = require "kong.clustering.wrpc_data_plane".new(self)
+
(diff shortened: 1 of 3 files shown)


javierguerragiraldez:
Changing version negotiation flow.  Do not merge

kong/clustering/init.lua
```diff
@@ -205,14 +205,34 @@ local function fill_empty_hashes(hashes)
...
-    ngx_log(ngx_ERR, _log_prefix, "error while requesting version negotiation: " .. err)
-    assert(ngx.timer.at(math.random(5, 10), function(premature)
-      self:communicate(premature)
-    end))
-    return
+    ngx_log(ngx_ERR, _log_prefix, "error while requesting version negotiation: ", err)
+    ngx_log(ngx_ERR, _log_prefix, "Assuming old JSON protocol.")
+  end
+
+  local config_proto, msg = version_negotiation.get_negotiated_service("config")
```
suikabreaker, line 220: Should we still go though this if `request_version_handshake` fails? What do we get in that case?

suikabreaker: And if we already determined to use old protocol, we should not see the "error reading negotiated..." log.

kong/clustering/init.lua
```diff
@@ -205,14 +205,34 @@ local function fill_empty_hashes(hashes)
...
+    self.child = require "kong.clustering.wrpc_data_plane".new(self)
+
+  elseif config_proto == "v0" or config_proto == nil then
+    self.child = require "kong.clustering.data_plane".new(self)
+  end
+
+  if self.child then
+    clustering_utils.load_config_cache(self.child)
+    self.child:communicate(premature, config_proto)
   end
```
suikabreaker, line 236: else?

kong/clustering/init.lua
```diff
@@ -205,14 +205,34 @@ local function fill_empty_hashes(hashes)
...
 end
 
-function _M:request_version_negotiation()
+
+function _M:call_control_plane(premature)
```
suikabreaker, line 209: I think the function could be just "version\_negotiation" (though the name is already used).

Suika-Kong, line 209: The function's name could be simply "version\_negotiation" (though it's already used).

javierguerragiraldez: no, it does initiate the DP->CP connection.

kong/clustering/init.lua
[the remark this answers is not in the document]
javierguerragiraldez: that would be a future "not nil, not v0, not v1" case.

javierguerragiraldez:
as mentioned above, the version negotiation specs have changed (implemented in #8775 ), but the current master has at least one bug when failure to connect leads to restarting only part of the cycle, locking into the wrong protocol.  this patch fixes that by assuring that failures restart from the start.

PR #8764, opened 2022-05-07
chore(clustering) log `push_config` duration

mayocream:
### Summary

Log `push_config` duration

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -24,6 +24,7 @@ local ngx_exit = ngx.exit
 local exiting = ngx.worker.exiting
 local ngx_time = ngx.time
 local ngx_now = ngx.now
+local ngx_update_time = ngx.update_time
 local ngx_var = ngx.var
 local table_insert = table.insert
 local table_remove = table.remove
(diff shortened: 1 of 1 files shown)


fffonion (approved):

dndx:
@ADD-SP Yes, we are more interested in the export time so this change should be fine.

kong/clustering/control_plane.lua
```diff
@@ -248,6 +249,9 @@ end
 
 
 function _M:push_config()
+  ngx_update_time()
```
dndx, line 252: I think this can be removed without affecting the result (we can keep the later one as that is after the CPU intensive port of the export).

mayocream: Thanks!
```suggestion
```

kong/clustering/control_plane.lua
```diff
@@ -261,7 +264,9 @@ function _M:push_config()
     n = n + 1
   end
 
-  ngx_log(ngx_DEBUG, _log_prefix, "config pushed to ", n, " clients")
+  ngx_update_time()
+  local duration = ngx_now() - start
+  ngx_log(ngx_DEBUG, _log_prefix, "config pushed to ", n, " clients in " .. duration .. "s")
```
hbagdi, line 269:
```suggestion
  ngx_log(ngx_DEBUG, _log_prefix, "config pushed to ", n, " data-plane nodes in " .. duration .. " seconds")
```

dndx (approved):

PR #8765, opened 2022-05-07
feat(plugin/aws-lambda)accept string type statusCode as return when working in proxy integration mode

windmgc:
### Summary

Hi, I would like to add some compatibility to Kong's `aws-lambda` plugin so that it can receive string type status code when working in the proxy integration mode.

When AWS Lambda function is working as a proxy integration, the `statusCode` returned by lambda function will become the HTTP status code returned by API gateway.

Currently, we only accept number type `statusCode`,
https://github.com/Kong/kong/blob/23f50ea1cd3e6985713ea29c81bc4d50f86f2d6e/kong/plugins/aws-lambda/handler.lua#L75

After some tests I found that AWS API gateway works fine if the type of `statusCode` is either number or string. And if the `statusCode` is an invalid HTTP status code, such as `600`(smaller than 100 or larger than 599) or `200a`(containing non-numeric characters), it will throw an "internal server error".

It would be good for us to be compatible with a string type status code, according to FTI-4014.

### Full changelog

* Add a feature that make aws-lambda handler be compatible with string type `statusCode` 
* Add validation function to check if the `statusCode` is legal

### Issue reference

Fix FTI-4014

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -74,7 +74,6 @@ local function validate_http_status_code(status_code)
     if not status_code then
       return false
     end
-
   end
 
   if status_code >= 100 and status_code <= 599 then
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -70,10 +70,11 @@ local function validate_http_status_code(status_code)
 
   if type(status_code) == "string" then
     status_code = tonumber(status_code)
-  end
 
-  if not status_code then
-    return false
+    if not status_code then
+      return false
+    end
+
   end
 
   if status_code >= 100 and status_code <= 599 then
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -17,8 +17,6 @@ local AWS_REGION do
   AWS_REGION = os.getenv("AWS_REGION") or os.getenv("AWS_DEFAULT_REGION")
 end
 
-local VALID_HTTP_STATUS_CODE_REG = "^[1-5][0-9][0-9]$"
-
 
 local fetch_credentials do
   local credential_sources = {
@@ -43,7 +41,6 @@ local tostring = tostring
 local tonumber = tonumber
 local ngx_now = ngx.now
 local ngx_var = ngx.var
-local re_match = ngx.re.match
 local error = error
 local pairs = pairs
 local kong = kong
@@ -65,21 +62,27 @@ local function get_now()
   return ngx_now() * 1000 -- time is kept in seconds with millisecond resolution.
 end
 
+
 local function validate_http_status_code(status_code)
   if not status_code then
     return false
   end
-  local status_code_str = tostring(status_code)
-  local m, err = re_match(status_code_str, VALID_HTTP_STATUS_CODE_REG, 'jo')
-  if err then
-    kong.log.err(err)
+
+  if type(status_code) == "string" then
+    status_code = tonumber(status_code)
+  end
+
+  if not status_code then
     return false
   end
-  if not m then
-    return false
+
+  if status_code >= 100 and status_code <= 599 then
+    return status_code
   end
-  return true
+
+  return false
 end
+
 
 --[[
   Response format should be
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -67,7 +67,7 @@ end
 
 local function validate_http_status_code(status_code)
   if not status_code then
-    return false, "statusCode not found"
+    return false
   end
   local status_code_str = tostring(status_code)
   local m, err = re_match(status_code_str, VALID_HTTP_STATUS_CODE_REG, 'jo')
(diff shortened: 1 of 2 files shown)


fffonion (approved):
🎉 First PR!

ms2008 (approved):
LGTM

kong/plugins/aws-lambda/handler.lua
```diff
@@ -62,6 +65,21 @@ local function get_now()
   return ngx_now() * 1000 -- time is kept in seconds with millisecond resolution.
 end
 
+local function validate_http_status_code(status_code)
+  if not status_code then
+    return false
+  end
+  local status_code_str = tostring(status_code)
```
dndx, line 72: This seems to be slightly unnecessary. I think if the `status_code` is already a number, maybe we do not need to run the regex below?

What about:

```
if type(status_code) == "string" then
    status_code = tonumber(status_code)
end
```

And later below, we can do:

```
if status_code >= 100 and status_code <= 599 then
  return status_code
end

return nil
```

This should be a lot faster than a regex based solution as well.

kong/plugins/aws-lambda/handler.lua
```diff
@@ -62,6 +65,21 @@ local function get_now()
...
+  local m, err = re_match(status_code_str, VALID_HTTP_STATUS_CODE_REG, 'jo')
+  if err then
+    kong.log.err(err)
+    return false
+  end
+  if not m then
```
dndx, line 78:
```suggestion

  if not m then
```

Style: need a newline before and after a new code block.

kong/plugins/aws-lambda/handler.lua
```diff
@@ -62,6 +65,21 @@ local function get_now()
...
+    return false
+  end
```
dndx, line 80:
```suggestion
  end

```

kong/plugins/aws-lambda/handler.lua
windmgc: Thanks for pointing it out! Seems that I also forget to add two blank lines between functions

Fixed

kong/plugins/aws-lambda/handler.lua
windmgc: OK I've changed the code to compare numbers directly.

And we need to do additional check if `status_code` is nil after the `tonumber` conversion because nil cannot compare with numbers.

Fixed

windmgc:
@dndx Could you review the PR again? Thanks!

kong/plugins/aws-lambda/handler.lua
windmgc: Code changed, fixed

kong/plugins/aws-lambda/handler.lua
```diff
@@ -63,6 +63,27 @@ local function get_now()
...
+
+  if type(status_code) == "string" then
+    status_code = tonumber(status_code)
+  end
+
+  if not status_code then
+    return false
+  end
```
dndx, line 77: This should be moved inside the `if` block above, if the `status_code` is not `string` then there is no reason to check it again here.

spec/03-plugins/27-aws-lambda/99-access_spec.lua
```diff
@@ -805,7 +805,7 @@ for _, strategy in helpers.each_strategy() do
 -- and the tests pass.
 -- see: https://github.com/Kong/kong/commit/c6f9e4558b5a654e78ca96b2ba4309e527053403#diff-9d13d8efc852de84b07e71bf419a2c4d
 
-      it("sets proper status code on custom response from Lambda", function()
+      it("sets proper status code(number type) on custom response from Lambda", function()
```
dndx, line 808:
```suggestion
      it("sets proper status code (type = number) on custom response from Lambda", function()
```

spec/03-plugins/27-aws-lambda/99-access_spec.lua
```diff
@@ -823,6 +823,24 @@ for _, strategy in helpers.each_strategy() do
         assert.equal("", body)
       end)
 
+      it("sets proper status code(string type) on custom response from Lambda", function()
```
dndx, line 826:
```suggestion
      it("sets proper status code (type = string) on custom response from Lambda", function()
```

kong/plugins/aws-lambda/handler.lua
windmgc: Fixed

I wonder if it needs more strict check on types, like

```
  if type(status_code) == "string" or type(status_code) == "number" then
    status_code = tonumber(status_code)

    if not status_code then
      return false
    end

  else
    -- Validation failed if status code is not string or number
    return false
  end
  
  end
```

kong/plugins/aws-lambda/handler.lua
```diff
@@ -63,6 +63,28 @@ local function get_now()
...
+
+  if type(status_code) == "string" then
+    status_code = tonumber(status_code)
+
+    if not status_code then
+      return false
+    end
+
```
dndx, line 77:
```suggestion
```

We don't need a newline if there are no more code after the block.

dndx (approved):

kong/plugins/aws-lambda/handler.lua
windmgc: Got it, thanks!

PR #8766, opened 2022-05-07
fix(prometheus) move total out of current_connections metrics and

### LINKED ISSUE #7680: Incorrect interpretation in NGINX state

fffonion:
create new nginx_{http,stream}_total_requests

Fix #7680

### DIFF
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -43,10 +43,14 @@ local function init()
     metrics.connections = prometheus:gauge("nginx_http_current_connections",
       "Number of HTTP connections",
       {"state"})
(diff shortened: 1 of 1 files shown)


dndx (commented):
Does this needs a test case?

fffonion:
Will add a test. Note to myself: rename it to xxx_requests_total.

fffonion:
replaced by https://github.com/Kong/kong/pull/8712

PR #8770, opened 2022-05-09
chore(zipkin) Add zipkin prefix to timer logs

dcarley:
### Summary

These run on a timer so they don't get the namespace automatically
appended to the log line which makes it hard to determine which plugin
they come from:

    [error] 1099#0: *436751 [kong] handler.lua:102 reporter flush failed to request: operation timed out, context: ngx.timer, client: 172.17.0.1, server: 0.0.0.0:8443

I've tried looking for a way to solve this generically that would
benefit other plugins like `rate-limiting` but can't see a way to set
the logger or get access to the plugin name when the timer is running.

I've omitted the square brackets from `zipkin` to make it clearer that
it's not coming from the automatic namespace formatting.

### Full changelog

* feat(zipkin) Add zipkin prefix to timer logs

### Issue reference

N/A

### Other notes

I'm open to suggestions about how this could be solved generically in a way that would benefit all plugins.

This PR depends on the tests from #8735. I can't see a way to change the base branch, so it will need rebasing after that PR is merged.

### DIFF
--- a/kong/plugins/zipkin/handler.lua
+++ b/kong/plugins/zipkin/handler.lua
@@ -96,7 +96,7 @@ local function timer_log(premature, reporter)
 
   local ok, err = reporter:flush()
   if not ok then
-    kong.log.err("reporter flush ", err)
+    kong.log.err("zipkin reporter flush ", err)
     return
   end
 end
--- a/spec/03-plugins/34-zipkin/zipkin_spec.lua
+++ b/spec/03-plugins/34-zipkin/zipkin_spec.lua
@@ -388,7 +388,7 @@ for _, strategy in helpers.each_strategy() do
       -- wait for zero-delay timer
       helpers.wait_timers_end(wait_timers_ctx, 0.5)
 
-      assert.logfile().has.line("reporter flush failed to request: timeout", false, 2)
+      assert.logfile().has.line("zipkin reporter flush failed to request: timeout", false, 2)
     end)
 
     it("times out if upstream zipkin server takes too long to respond", function()
@@ -406,7 +406,7 @@ for _, strategy in helpers.each_strategy() do
       -- wait for zero-delay timer
       helpers.wait_timers_end(wait_timers_ctx, 0.5)
 
-      assert.logfile().has.line("reporter flush failed to request: timeout", false, 2)
+      assert.logfile().has.line("zipkin reporter flush failed to request: timeout", false, 2)
     end)
 
     it("connection refused if upstream zipkin server is not listening", function()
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -139,6 +139,14 @@
 - **Zipkin**: add support for including HTTP path in span name
   through configuration property `http_span_name`.
   [#8150](https://github.com/Kong/kong/pull/8150)
+- **Zipkin**: add support for socket connect and send/rev timeouts
```
mayocream, line 142: I think the timeout options were in another PR. Could you rebase this PR on the master branch?

mayocream:
The changes should go to `chore` instead of `feat`, and I think this small change is unnecessary to be included in the changelog.

CHANGELOG.md
dcarley: This is branched from #8735 (still in review) because it depends on the new tests added there. I'd normally change the base branch in GitHub so that only the new commits show up in the diff but I'm not able to do that across two forks, so instead I'll rebase when #8735 is merged.

dcarley:
Sorry for the delay. I've rebased against master, changed the commit to `chore`, and removed the CHANGELOG entry.

Tieske:
Shouldn't we be fixing the underlying problem with `kong.log` to also provide the proper namespace when in a timer context?

@kikito any opinions? iirc you were involved when `kong.log` was first implemented.

dcarley:
> Shouldn't we be fixing the underlying problem with `kong.log` to also provide the proper namespace when in a timer context?

I'd love to, but can't see a way to set the logger or get access to the plugin name when the timer is running. The only option I can think of is to provide a wrapper over `ngx.timer.at` that sets the logger, but even then there's no guarantee that every plugin will use it. Any other ideas?

Tieske:
@ADD-SP is this something we can fix with the timer-ng lib?

kikito:
The prefix is lost because our plugins iterator sets a custom timer on each phase of a plugin iterator, but it gets lost on timers. I believe this is a general issue. For now we could get over that by the logger inside the reporter:

```
local function get_reporter(conf)
  if reporter_cache[conf] == nil then
    reporter_cache[conf] = new_zipkin_reporter(conf.http_endpoint,
                                               conf.default_service_name,
                                               conf.local_service_name,
                                               conf.connect_timeout,
                                               conf.send_timeout,
                                               conf.read_timeout,
                                               kong.log)
  end
  return reporter_cache[conf]
end

```

And using a method like `reporter.logger.log` instead of kong.log

mayocream:
Solved by #9874

dcarley:
Thanks!

PR #8771, opened 2022-05-09
chore(ci): adjust how we version our nightlies

hutchic:
### Summary

adjust commit based releases to return a calver on `kong version` command and to push a calver'd docker tag per https://github.com/Kong/kong-madr/pull/37

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,7 +61,7 @@ pipeline {
             steps {
                 sh 'echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin || true'
                 sh 'make setup-kong-build-tools'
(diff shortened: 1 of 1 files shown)


mikefero:
This won't correct the issue with wRPC and Kong Gateway (OSS) nightly images using hybrid mode currently; we need a proper semver in order for the version negotiation to take place. The `Makefile` updates the version returned in kong/meta.lua by replacing the version with just the definition of `KONG_VERSION` which in this updated case would be the calver instead of the commit SHA; see https://github.com/Kong/kong/blob/23f50ea1cd3e6985713ea29c81bc4d50f86f2d6e/Makefile#L79 for more context.

mikefero:
Here is an example of an external application failing tests because of the version number returned by the nightly image; https://github.com/Kong/deck/pull/671

dndx (approved):

PR #8775, opened 2022-05-10
Chore/wrpc inline version negotiation

javierguerragiraldez:
Move the services/versions negotiation to the wRPC connection.

Because of limitations in the lua-resty-websocket library, the HTTP-based version negotiation and the websocket connection used separate TCP connections instead of reusing (via keep-alive).  So, in the (very common) case of a load balancer in front of a cluster of control plane nodes, it could easily happen to negotiate against one node and then connect to another.  To avoid that situation, this PR changes the implemenation of the version negotiation into a wRPC service with a single RPC method.  The DP is responsible to make this call as soon as connected, so the agreed result of the negotiation is stored in both sides in the context of the peer connection.


javierguerragiraldez:
Still things to do, but it actually works and the error flows are in place.


- [x] try/retry/fallback structure
    - [x] try wrpc first
    - [x] fallbak to old proto on 404
        - [ ] reliable detection of 404
    - [x] on connection/no host error go to the loop of timed retries
- [x] data plane starts with NegotiateServices rpc call
- [x] rpc handler on control plane performs negotiation
    - [x] control plane saves result in data plane's context (same as old HTTP negotiation)
- [x] data plane stores result (same as old HTTP negotiation)
- [ ] remove unused code
- [ ] fix/replace old tests

kong/clustering/init.lua
```diff
@@ -216,6 +216,45 @@ function _M:request_version_negotiation()
...
+  -- TODO: find how to detect a 404 response
+  if not ws_conn and err == "404" then
+    ngx_log(ngx_DEBUG, _log_prefix, "retrying on wsjs")
+    implementation = wsjs_dp
+    ws_conn, err = implementation.open_connection(self)
+    ngx_log(ngx_DEBUG, _log_prefix, "got conn: ", tostring(ws_conn), ", err: ", tostring(err))
+  end
+
+  if not ws_conn then
+    ngx_log(ngx_DEBUG, _log_prefix, "ugh retrying from the start")
```
hbagdi, line 242: ugh?

kong/clustering/init.lua
```diff
@@ -216,6 +216,45 @@ function _M:request_version_negotiation()
...
 
+local function call_control_plane(premature, self)
+  if premature then
+    return
+  end
+
+  local wsjs_dp = require "kong.clustering.data_plane"
+  local wrpc_dp = require "kong.clustering.wrpc_data_plane"
+
+  ngx_log(ngx_DEBUG, _log_prefix, "will try wrpc first")
```
hbagdi, line 227:
```suggestion
  ngx_log(ngx_DEBUG, _log_prefix, "attempting wRPC connection")
```

kong/clustering/init.lua
```diff
@@ -216,6 +216,45 @@ function _M:request_version_negotiation()
...
+
+  local implementation = wrpc_dp
+  local ws_conn, err = implementation.open_connection(self)
+  ngx_log(ngx_DEBUG, _log_prefix, "got conn: ", tostring(ws_conn), ", err: ", tostring(err))
+
+  -- TODO: find how to detect a 404 response
```
hbagdi, line 233: ?

kong/clustering/init.lua
```diff
@@ -216,6 +216,45 @@ function _M:request_version_negotiation()
...
+  if not ws_conn and err == "404" then
+    ngx_log(ngx_DEBUG, _log_prefix, "retrying on wsjs")
```
hbagdi, line 235:
```suggestion
    ngx_log(ngx_DEBUG, _log_prefix, "wRPC not available, falling back to vanilla websocket")
```
wsjs is not a known term to most Kong engineers or our users.

kong/clustering/init.lua
```diff
@@ -216,6 +216,45 @@ function _M:request_version_negotiation()
...
+    implementation = wsjs_dp
+    ws_conn, err = implementation.open_connection(self)
+    ngx_log(ngx_DEBUG, _log_prefix, "got conn: ", tostring(ws_conn), ", err: ", tostring(err))
```
hbagdi, line 238:
```suggestion
    ngx_log(ngx_DEBUG, _log_prefix, "failed to connect: ", tostring(ws_conn), ", err: ", tostring(err))
```

kong/clustering/wrpc_control_plane.lua
```diff
@@ -60,6 +61,7 @@ end
 local function get_config_service(self)
   if not wrpc_config_service then
     wrpc_config_service = wrpc.new_service()
+    version_negotiation.add_negotiation_service(wrpc_config_service, self.conf)
     wrpc_config_service:add("kong.services.config.v1.config")
```
hbagdi, line 65: What prevents a CP node to execute a ConfigService.SyncConfig before the negotiation is finished?

hbagdi: The same question for the DP.

kong/clustering/init.lua
javierguerragiraldez: the websocket library doesn't return the http result code, instead, returns the whole header line if it doesn't match the expected pattern.  this, added to some inherited fallbacks in the specific `server` context in nginx config, makes reliably detecting a 404 response a bit more involved than expected.  i left the specifics to after i get the negotiation itself working.  (that is, now). 

the `"got conn:"` log messages are just to see the exact returned values. they won't stay in the final code

kong/clustering/wrpc_control_plane.lua
javierguerragiraldez: nothing yet.  i could put a "negotiated" flag in the peer object and just drop everything else until then.

kong/clustering/init.lua
javierguerragiraldez: as it turns out, any HTTP result is "good enough" to the library, so the user never sees a 404 or any other error.

I've just hacked a peeking socket to work around it.

Suika-Kong:
The refactoring to wRPC causes too much conflict to this PR. We will create another PR and take this as a reference.

kong/include/kong/model/negotiation.proto
```diff
@@ -0,0 +1,47 @@
...
+import "google/protobuf/struct.proto";
+
+message NegotiateServicesRequest {
+  DPNodeDescription node = 1;
+  repeated ServiceRequest services_requested = 2;
+  google.protobuf.Struct metadata = 3;
+}
+
+message DPNodeDescription {
+  string id = 1;
```
Suika-Kong, line 16: The id should be already known from URI parameters. Do we need to include it here?

kong/include/kong/model/negotiation.proto
```diff
@@ -0,0 +1,47 @@
...
+
+message NegotiateServicesResponse {
+  string error_message = 1;
+  CPNodeDescription node = 2;
+  repeated AcceptedService services_accepted = 3;
+  repeated RejectedService services_rejected = 4;
+}
+
+message CPNodeDescription {
+  string id = 1;
```
Suika-Kong, line 35: It seems that CP's ID(cluster ID) is synced with the config.

Suika-Kong: Maybe a more proper way to do that is to restructure config and get ID from negotiation.

chronolaw:
Due to we have merged another PR (#8926), I think we should close this.

Issue #8788, opened 2022-05-11
Plugins/DAO: incorrect function field reassignment


marc-charpentier:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

Kong 2.8.1

### Current Behavior

Hello,
In order to load a new plugin entity, a worker executes the `Plugins.select_by_cache_key` function (call to `kong.db.plugins:select_by_cache_key(...)`, which delegates the load to `DAO:select_by_cache_key`, and checks whether the _core.009_200_to_210_ migration was executed previously.
If yes, before returning the loaded entity, the `Plugins.select_by_cache_key` function tries to self-replace by the DAO's one, but it doesn't work because the assigned field's object isn't the right one.
As an impact, an important extra latency (kong) is added to the plugin's load one, for each request involving such load.

### Expected Behavior

Once a worker as ensured that the _core.009_200_to_210_ migration was executed, it should no longer recheck for any plugin updated / created afterward.

### Steps To Reproduce

```markdown
Problem encountered with Kong in "traditional" role.
On a local VM environnement (Kong and Cassandra containers), extra latency is about 10-15 ms, but on an integration environment (production-like with multi-DCs/multi-nodes Cassandra and Kong clusters), extra latency is about 500 ms.

Steps to reproduce (local environment):
- Start Kong with one single nginx worker.
- Use the Admin API to create a route (hosts[]=localhost) and a request-termination plugin on it (message=OK&status_code=200).
- Curl the route with -i switch in order to check the X-Kong-Response-Latency header:
  - Expected value: 0-2 ms.
  - Real value: 10-12 ms.
From now the request-termination plugin is in the cache.
- Curl again the same route, X-Kong-Response-Latency should be about 0-2 ms.
- Update the request-termination plugin (message=Hello).
- Curl again the same route, X-Kong-Response-Latency should be again about 10-12 ms.

On the integration environment, the X-Kong-Response-Latency value was about 500 ms for the first response after plugin creation/update.
```

### Anything else?

To confirm that `Plugins:select_by_cache_key` function never succeeds to disable itself, I also added a notice message in its code, just before the instruction that should do the self-disabling. It should have logged the notice message only once (per worker, provided plugin is created/updated as many times as there are workers), but it did logged it for the first response after each plugin creation/update.

jschmid1:
[...]

PR #8789, opened 2022-05-11
fix(dao) fix incorrect function field reassignment

marc-charpentier:
The Plugins DAO's select_by_cache_key translator function was not disabled even if checked migration had actually been executed before.

### Summary

In order to load a new plugin entity, a worker executes the `Plugins.select_by_cache_key` function (call to `kong.db.plugins:select_by_cache_key(...)`, which delegates the load to `DAO:select_by_cache_key`, and checks whether the _core.009_200_to_210_ migration was executed previously.
If yes, before returning the loaded entity, the `Plugins.select_by_cache_key` function tries to self-replace by the DAO's one, but it doesn't work because the assigned field's object isn't the right one.
As an impact, an important extra latency is added to the plugin's load one, for each request involving such load.

This PR is to fix this behavior, so that extra latency may happen only once per worker.

### Full changelog

* [Fix ever checking core.009_200_to_210 execution on plugin load]

### Issue reference

[Issue #8788](https://github.com/Kong/kong/issues/8788)

### DIFF
--- a/kong/db/dao/plugins.lua
+++ b/kong/db/dao/plugins.lua
@@ -336,7 +336,8 @@ function Plugins:select_by_cache_key(key)
 
     -- if migration is complete, disable this translator function and return
     if schema_state:is_migration_executed("core", "009_200_to_210") then
-      Plugins.select_by_cache_key = self.super.select_by_cache_key
+      self.select_by_cache_key = self.super.select_by_cache_key
+      Plugins.select_by_cache_key = nil
       return entity
     end
   end


bungle:
@marc-charpentier as we are approaching 3.0, I think we can remove this check altogether, right @hishamhm?

hishamhm:
@bungle in the past we have removed those checks once their corresponding migrations were "squashed". I don't know which migrations will be squashed for 3.0, but if 200_to_210 migration file stops existing in the 3.0 codebase then I believe this translator function can go away as well in 3.0.

marc-charpentier:
@bungle Indeed, with my workmates we chose to remove the translator function altogether when deploying Kong 2.8 in our environments because we ensure that all migrations are run before.

marc-charpentier:
@bungle
Hello,
And sorry for this late answer.
Although we chose to remove the function for our own Kong instances, I had wished to contribute with a fix not thwarting blue-green deployment support, and thought it would be nice to have it in a 2.8.2+ release.
Also, I suppose the function's complete removal was possible for us because we chose the 'no downtime way' involving a Cassandra keyspace copy and migration before starting 2.8 nodes.

PR #8795, opened 2022-05-12
feat(nightly): updating nightly images to contain semver release version

mikefero:
### Summary

This update bumps the minor version defined in `kong/meta.lua` and applies a suffix of '-dev+<commit-sha>.<date-time>' identifying a development pre-release with a parsable semver for Kong Gateway. This ensures that 'kong version', root information endpoint, and the data plane version negotiation message are semver compliant.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,8 @@ endif
 	lint test test-integration test-plugins test-all \
 	pdk-phase-check functional-tests \
 	fix-windows \
-	nightly-release release
+	nightly-release release \
+	bump-dev-version
 
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 KONG_SOURCE_LOCATION ?= $(ROOT_DIR)
@@ -72,12 +73,18 @@ else
 	REPOSITORY_OS_NAME = ${BRANCH}
 	KONG_PACKAGE_NAME ?= kong-${BRANCH}
 	KONG_VERSION ?= `date +%Y-%m-%d`
+	COMMIT_SHA ?= $(shell git rev-parse --short HEAD)
+	DATE_TIME_VERSION ?= $(shell date '+%Y%m%d%H%M')
 endif
 
-release:
+bump-dev-version:
 ifeq ($(ISTAG),false)
-	sed -i -e '/return string\.format/,/\"\")/c\return "$(KONG_VERSION)\"' kong/meta.lua
+	@sed -i -E "s/minor[[:space:]]*=[[:space:]]*[0-9]+/minor = $(shell lua -e 'print(require("kong.meta")._VERSION_TABLE.minor + 1)')/" kong/meta.lua
+	@sed -i -E "s/patch[[:space:]]*=[[:space:]]*[0-9]+/patch = 0/" kong/meta.lua
+	@sed -i -E "s/--[[:space:]]*suffix[[:space:]]*=.*/suffix = \"-dev+$(COMMIT_SHA).$(DATE_TIME_VERSION)\"/" kong/meta.lua
 endif
+
+release: bump-dev-version
 	cd $(KONG_BUILD_TOOLS_LOCATION); \
 	$(MAKE) \
 	KONG_VERSION=${KONG_VERSION} \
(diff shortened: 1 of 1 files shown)


mikefero:
> [https://github.com/Kong/kong/blob/master/Jenkinsfile#L64](https://github.com/Kong/kong/blob/master/Jenkinsfile?rgh-link-date=2022-05-12T17%3A00%3A25Z#L64) is going to smash these changes

`KONG_VERSION` will remain intact since `make` does not allow you to override variables once they are defined. https://github.com/Kong/kong/blob/1439f9c4b928a47cad7e29660ceeea32d3bb99d8/Makefile#L47 uses either the env `KONG_VERSION` which is set in the Jenkinsfile as you pointed out and then every other attempt of overriding `KONG_VERSION` [here](https://github.com/Kong/kong/blob/1439f9c4b928a47cad7e29660ceeea32d3bb99d8/Makefile#L62) and [here](https://github.com/Kong/kong/blob/1439f9c4b928a47cad7e29660ceeea32d3bb99d8/Makefile#L74) is essentially a no-op because they will never be assigned.

Only the `kong/meta.lua` file will be changed during this type of release.

Here is an example to demonstrate:

```Makefile
KONG_VERSION ?= "INITIALIZED"

ifneq ($(CONDITION),)
	KONG_VERSION ?= "IS CONDITION"
else
	KONG_VERSION ?= "NOT CONDITION"
endif

.MAIN: example

example:
	@echo $(KONG_VERSION)
```

Executing this `Makefile` will give you the following results

```
make SHELL="sh -x"
CONDITION="TRUE" make SHELL="sh -x"
```

```
+ echo INITIALIZED
INITIALIZED
```

```
KONG_VERSION="OVERRIDE" CONDITION="TRUE" make SHELL="sh -x"
KONG_VERSION="OVERRIDE" make SHELL="sh -x"
```

```
+ echo OVERRIDE
OVERRIDE
```

If you remove the first line in the above example you will be able to see how `make` variables are set only once during the parsing phase:

```Makefile
ifneq ($(CONDITION),)
	KONG_VERSION ?= "IS CONDITION"
else
	KONG_VERSION ?= "NOT CONDITION"
endif

.MAIN: example

example:
	@echo $(KONG_VERSION)
```

```
make SHELL="sh -x"
```

```
+ echo NOT CONDITION
NOT CONDITION
```

```
CONDITION="TRUE" make SHELL="sh -x"
```

```
+ echo IS CONDITION
IS CONDITION
```

```
KONG_VERSION="OVERRIDE" CONDITION="TRUE" make SHELL="sh -x"
KONG_VERSION="OVERRIDE" make SHELL="sh -x"
```

```
+ echo OVERRIDE
OVERRIDE
```

hutchic:
@mikefero the only purpose of that line in the Jenkinsfile is to override the `meta.lua`. Let's keep using it or get rid of it.

The current implementation depending on lua breaks with our tradition of keep builds dependent on bash / Make / docker. Our build workers for example don't have lua installed and as such would need to be added to the build worker or the Jenkinsfile.

I tried to run it locally and got an error
```
make bump-dev-version
fatal: no tag exactly matches '697bb147a3676e84572573241575803857383b25'
lua: error loading module 'kong.meta' from file './kong/meta.lua':
        ./kong/meta.lua:3: unexpected symbol near ','
stack traceback:
        [C]: in ?
        [C]: in function 'require'
        (command line):1: in main chunk
        [C]: in ?
```

mikefero:
> The current implementation depending on lua breaks with our tradition of keep builds dependent on bash / Make / docker. Our build workers for example don't have lua installed and as such would need to be added to the build worker or the Jenkinsfile.

Since `luarocks` and `luacov` were being used I assumed that we could use `lua` as well; if that is not the case then we should close the PR and implement things a different way.

> I tried to run it locally and got an error

Interesting no issue when I execute it locally

```
make --version
GNU Make 4.2.1

bash --version
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)

lua -v
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
```

```
make bump-dev-version
fatal: no tag exactly matches '697bb147a3676e84572573241575803857383b25'

git diff
diff --git a/kong/meta.lua b/kong/meta.lua
index 36a96b45b..de087c3e7 100644
--- a/kong/meta.lua
+++ b/kong/meta.lua
@@ -1,8 +1,8 @@
 local version = setmetatable({
   major = 2,
-  minor = 8,
+  minor = 9,
   patch = 0,
-  --suffix = "rc.1"
+  suffix = "-dev+697bb147a.202205121643"
 }, {
   -- our Makefile during certain releases adjusts this line. Any changes to
   -- the format need to be reflected in both places

```

hbagdi:
ping @hutchic 
Please either close this or move this to completion.
Assume @mikefero won't be able to drive this to completion because of a change in his responsibilities.

PR #8796, opened 2022-05-12
perf(schema) no deep copy on select on process auto fields

bungle:
### Summary

It is inefficient to create deep copies of tables when e.g. looping through database rows (and with records in them we recursively deep copy them). In my testing with uploading some 64k routes with dbless (which calls process auto fields twice), this can cut the time looping the data by 1/4th. It also generates much less garbage.

I searched our code bases where we use "select" context, and could not find anything that might break because of this.

### DIFF
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -1609,11 +1609,15 @@ end
 -- valid values are: "insert", "update", "upsert", "select"
 -- @param nulls boolean: return nulls as explicit ngx.null values
 -- @return A new table, with the auto fields containing
--- appropriate updated values.
+-- appropriate updated values (except for "select" context
+-- it does it in place by modifying the data directly).
 function Schema:process_auto_fields(data, context, nulls, opts)
   local check_immutable_fields = false
 
-  data = tablex.deepcopy(data)
+  local is_select = context == "select"
+  if not is_select then
+    data = tablex.deepcopy(data)
+  end
 
   local shorthand_fields = self.shorthand_fields
   if shorthand_fields then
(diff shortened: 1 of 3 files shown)


flrgh (commented):
Soft :+1: My knowledge of the dao/schema modules does not go very deep, but after looking through some of the code this seems relatively safe/sensible.

One note; the plugin doc string states that a new table is returned. It would be good to update this to reflect the new behavior, that the table is modified in-place for `select`:

```lua
-- @return A new table, with the auto fields containing
-- appropriate updated values.
```

bungle:
> Soft 👍 My knowledge of the dao/schema modules does not go very deep, but after looking through some of the code this seems relatively safe/sensible.
> 
> One note; the plugin doc string states that a new table is returned. It would be good to update this to reflect the new behavior, that the table is modified in-place for `select`:
> 
> ```lua
> -- @return A new table, with the auto fields containing
> -- appropriate updated values.
> ```

Yes, last time I tried this there was some tests failing, but it might be because of other changes. But now it seems tests are green. The biggest issue to me is things like:

https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L766

And especially on dbless:
https://github.com/Kong/kong/blob/master/kong/db/strategies/off/init.lua#L181-L184
https://github.com/Kong/kong/blob/master/kong/db/dao/init.lua#L1398-L1402

I have been testing this with some 64k routes, and deep copying tables takes around 1/4th of the rebuild time. I will adjust the description there.

bungle:
> I will adjust the description there.

The description was adjusted and tests were added (so this is now "official", and not undefined behavior).

gszr (commented):
Failures in this PR (and other `perf` PRs) are unrelated; as the perf testing framework picks a package for the latest official Kong release, and Kong master requires a newer OpenResty version, the older OpenResty isn't recognizing a new Nginx directive that is being used.

flrgh (approved):

PR #8797, opened 2022-05-12
fix(balancer) set target status using hostname

locao:
### Summary

When a target was added using only its hostname, it was not possible to update its health status using only the hostname, i.e. omitting the IP. This change fixes that issue.

This problem was identified while fixing a different issue. The tests don't fail because their current behavior is removing a healthy target and adding it again as unhealthy. The test is already fixed in a different branch (fix/target_creation_method).

### Full changelog

* Check if the ip received is actually an ip before trying to set the health status.

### DIFF
--- a/kong/runloop/balancer/init.lua
+++ b/kong/runloop/balancer/init.lua
@@ -351,7 +351,7 @@ local function post_health(upstream, hostname, ip, port, is_healthy)
   end
 
   local ok, err
-  if ip then
+  if ip and (utils.hostname_type(ip) ~= "name") then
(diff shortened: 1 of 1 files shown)


gszr (approved):
LGTM.
Was the regression test added to the other branch?

locao:
> Was the regression test added to the other branch?

Before the change in the other branch, `POST`ing to `/upstreams/{upstream_id}/targets` overwrote the target, which is not the right behavior. After changing the `POST` behavior and updating the existent test to use `PUT` to actually update the target, it started to fail. IOW, no new tests were added, but updating the test to match the correct API made it fail.

PR #8798, opened 2022-05-12
fix(api): use POST to add targets to specific upstream

locao:
### Summary

In Kong 2.x series it was possible to update targets using `POST` method. When fixing this behavior, the path `/upstreams/{upstream}/targets` was mistakenly changed to also use `PUT` method, instead of only blocking adding duplicated targets. This PR changes it.

### Full changelog

* Removed support to `PUT` in `/upstreams/{upstream}/targets`
* Refuse to update targets using `POST`.
* Tests updated.
* ~~Cherry-picked #8797, in order to the tests to pass.~~ Needed sister PR already merged.

FT-2874
DECK-64

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,10 +79,12 @@
 
 #### Admin API
 
-- Insert and update operations on target entities require using the `PUT` HTTP
-  method now. [#8596](https://github.com/Kong/kong/pull/8596). If you have
-  scripts that depend on it being `POST`, these scripts will need to be updated
-  when updating to Kong 3.0.
+- `POST` requests on target entities endpoint are no longer able to update
+  existing entities, they are only able to create new ones.
+  [#8596](https://github.com/Kong/kong/pull/8596),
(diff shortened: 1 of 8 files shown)


saisatishkarra:
@Kong/gateway-moderators Can someone review and approve and merge this PR. This is blocking https://konghq.atlassian.net/browse/ENGEN-593.

kikito (commented):
The changelog currently contains this note in the Breaking Changes section:

```
Insert and update operations on target entities require using the PUT HTTP method now. https://github.com/Kong/kong/pull/8596. If you have scripts that depend on it being POST, these scripts will need to be updated when updating to Kong 3.0.
```

I believe it should be updated according to the changes on this PR. My suggestion:

```
POST requests on Target entities will no longer be able to update existing entities, they will only be able to create new ones. https://github.com/Kong/kong/pull/8596, https://github.com/Kong/kong/pull/8798. If you have scripts that use POST requests to modify Targets, you should change them to PUT requests before updating to Kong 3.0.
```

kikito (approved):

PR #8800, opened 2022-05-13
perf(db) yield on dbless daos

bungle:
### Summary

DBless doesn't naturally yield (not with `shared dict`, nor with `lmdb`). This may cause latency spikes when iterating over bigger lists, e.g. `kong.db.routes:each()`. This commit adds some yields so that iterating doesn't fully block the worker from doing other work too, so this is about cooperative multitasking.

### DIFF
--- a/kong/db/strategies/off/init.lua
+++ b/kong/db/strategies/off/init.lua
@@ -2,7 +2,7 @@ local declarative_config = require "kong.db.schema.others.declarative_config"
 local workspaces = require "kong.workspaces"
 local lmdb = require("resty.lmdb")
 local marshaller = require("kong.db.declarative.marshaller")
-
+local yield = require("kong.tools.utils").yield
 
 
 local kong = kong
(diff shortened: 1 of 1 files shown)


bungle:
I benchmarked this problem with this:
```(0.000 elapsed): rebuilding router...
(0.001 elapsed): looping through routes...
(0.634 elapsed): got some routes. <-- this is where inside we now yield too
(1.416 elapsed): looping through routes done (65415 routes found).
(1.416 elapsed): initializing new router...
(2.051 elapsed): initializing new router done.
(2.051 elapsed): replacing router with ee router...
(2.051 elapsed): replacing router with ee router done.
(2.051 elapsed): flushing router caches done.
(2.051 elapsed): rebuilding router done.```
```

Out of 2 secs, 1.5 secs are used just to gather data for the new router. The Router build itself takes around 0.5 secs. During those 1.5 secs, we didn't yield, and it affects the maximum latencies.

flrgh (approved):

dndx (approved):

PR #8806, opened 2022-05-16
test(tests)  force new kong instances for each test.

javierguerragiraldez:
trying to get more stable tests

### DIFF
--- a/spec/02-integration/09-hybrid_mode/03-pki_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/03-pki_spec.lua
@@ -6,7 +6,7 @@ for _, cluster_protocol in ipairs{"json", "wrpc"} do
   for _, strategy in helpers.each_strategy() do
     describe("CP/DP PKI sync works with #" .. strategy .. " backend, protocol " .. cluster_protocol, function()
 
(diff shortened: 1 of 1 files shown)


gszr (commented):
I'm unsure starting / stopping kong for every test will make them more stable. Might the issue be something else - eg, the timeout for `wait_until` might need to be longer than 5s?

javierguerragiraldez:
> I'm unsure starting / stopping kong for every test will make them more stable.

i don't know.  this is definitely not a final fix; just to explore if these tests are affected by this.   in any case, just raising waittimes is not a fix either.  i guess the real answer would go something like:

- use different names/routes/identifiers on each test so it doesn't matter if the server carries state from the previous one.
- clearly define each step of the setup and be sure on which order they happen, for example:
    - "check server are started/connected"
    - "check config is not there"
    - "push config change"
    - "check config is there"
  
  instead of:
    - "assume servers are started"
    - "push config"
    - "check config is there"

  probably there would be waitstates on some of those steps, but not fixed times.

gszr:
@javierguerragiraldez Do you have a sample of flaky CI job?

PR #8816, opened 2022-05-19
fix(clustering) do not export plugins on a route attached to a disabled service

ms2008:
### Summary

This PR fix for a new DP get zero size config while service with plugin-enabled route is disabled.

The primary reason for this is CP does not totally filtering the exported config so when running in hybrid mode, DPs are sent disabled services and the routes/plugins associated with them. 

FTI-4009 FTI-4040

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -483,8 +483,9 @@ local function export_from_db(emitter, skip_ws, skip_disabled_entities, expand_f
       -- as well do not export plugins and routes of dsiabled services
       if skip_disabled_entities and name == "services" and not row.enabled then
         disabled_services[row.id] = true
-      elseif skip_disabled_entities and name == "routes" and disabled_services[row.service.id] then
-        disabled_routes[row.id] = true
+      elseif skip_disabled_entities and name == "routes" and row.service and
+        disabled_services[row.service ~= null and row.service.id] then
+          disabled_routes[row.id] = true
       elseif skip_disabled_entities and name == "plugins" and not row.enabled then
         goto skip_emit
 
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -453,6 +453,7 @@ local function export_from_db(emitter, skip_ws, skip_disabled_entities, expand_f
   })
 
   local disabled_services = {}
+  local disabled_routes = {}
   for i = 1, #sorted_schemas do
     local schema = sorted_schemas[i]
     if schema.db_export == false then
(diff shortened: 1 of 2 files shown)


spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -288,6 +288,76 @@ for _, strategy in helpers.each_strategy() do
 
           proxy_client:close()
         end)
+
+        it('does not sync plugins when service with plugin-enabled route is disabled #aaa', function()
```
bungle, line 292:
```suggestion
        it('does not sync plugins when service with plugin-enabled route is disabled', function()
```

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -288,6 +288,76 @@ for _, strategy in helpers.each_strategy() do
...
+          -- this can probably be optimizted
+          ngx.sleep(2)
+
+          local proxy_client = helpers.http_client("127.0.0.1", 9002)
+
+          -- test route again
+          res = assert(proxy_client:send({
+            method  = "GET",
+            path    = "/soon-to-be-disabled",
+          }))
```
locao, line 356: Instead of `ngx.sleep`, you can wrap this code in a function using `helpers.wait_until()`, returning `status == 404`. When `status` is actually `404`, `wait_until` will finish. If `status` never is `404`, `wait_until` will timeout.

I don't remember now if the timeout is enough to fail the test, but you can look for examples, `wait_until` is widely used in the integration and plugins tests.

locao (changes requested):
The change I suggested in the test is not too difficult to add and will help us to not increase the CI executing time.

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -288,6 +288,74 @@ for _, strategy in helpers.each_strategy() do
 
           proxy_client:close()
         end)
+
+        it('does not sync plugins when service with plugin-enabled route is disabled', function()
```
fffonion, line 292:
```suggestion
        it('does not sync plugins on a route attached to a disabled service', function()
```

fffonion (approved):

fffonion (approved):

locao (dismissed):

kong/db/declarative/init.lua
```diff
@@ -482,7 +483,8 @@ local function export_from_db(emitter, skip_ws, skip_disabled_entities, expand_f
       -- as well do not export plugins and routes of dsiabled services
       if skip_disabled_entities and name == "services" and not row.enabled then
         disabled_services[row.id] = true
-
+      elseif skip_disabled_entities and name == "routes" and disabled_services[row.service.id] then
```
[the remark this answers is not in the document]
ms2008: Fixed, The tests are all green now

bungle (approved):

PR #8817, opened 2022-05-19
feat (conf) set non root user to run in devcontainer docker containers

### LINKED ISSUE #8748: Dockerfile Misconfiguration: Privileged Container

kong 2.8.0

### Current Behavior

https://github.com/Kong/kong/blob/fd04b2782b510351eb414fea9016efa40641e47f/.devcontainer/Dockerfile

Details:
The Dockerfile sets a container to run with root user

### Expected Behavior

Explanation:
Dockerfile with USER instruction that is set to root, has an overly permissive privilege to make changes inside the container. It
violates the principle of running images with least privileges. In usual cases, root permissions might be required to install
packages and create folders. After the installation is done, it is good practice to add a user with restricted privileges.

Recommendations:
It is good practice to run your containers as a non-root user when possible.
To modify a docker container to use a non-root user, the Dockerfile needs to specify a different user, such as:
RUN useradd myLowPrivilegeUser
USER myLowPrivilegeUse

### Steps To Reproduce

```markdown
https://github.com/Kong/kong/blob/fd04b2782b510351eb414fea9016efa40641e47f/.devcontainer/Dockerfile
```

### Anything else?

_No response_

DiptoChakrabarty:
### Summary
This sets a non root user in the devcontainer dockerfile to replace operations as root user

To ensure non root user is set as root user has high privileges 
### Full changelog
* add non user in devcontainer dockerfile

 ...

### Issue reference #8748 

Fix #8748

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -88,12 +88,12 @@ jobs:
       run: |
         grep -v '^#' ${{ env.KONG_SOURCE_LOCATION}}/.requirements >> $GITHUB_ENV
         echo "DOCKER_RELEASE_REPOSITORY=kong/kong" >> $GITHUB_ENV
-        echo "KONG_TEST_CONTAINER_TAG=$GITHUB_REF_NAME-${{ matrix.package_type }}" >> $GITHUB_ENV
+        echo "KONG_TEST_CONTAINER_TAG=${GITHUB_REF_NAME##*/}-${{ matrix.package_type }}" >> $GITHUB_ENV
         if [[ ${{matrix.package_type }} == "apk" ]]; then
-          echo "ADDITIONAL_TAG_LIST=$GITHUB_REF_NAME-alpine" >> $GITHUB_ENV
+          echo "ADDITIONAL_TAG_LIST=${GITHUB_REF_NAME##*/}-alpine" >> $GITHUB_ENV
         fi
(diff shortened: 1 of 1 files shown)


flrgh:
Hello @DiptoChakrabarty :wave:

Seems like a reasonable change to my eyes, but I'm not familiar with the semantics of `.devcontainer`, so I've requested a review from the original author of this file.

In the meantime, can you please update your commit message to conform to our guidelines [here](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format) for ease of history management at our end? Specifically, there should be no space between the `type` and `scope` (i.e. `feat(conf)` not `feat (conf)`). Additionally, I would suggest an alternate type/scope for this particular change--maybe `chore(devcontainer)` since this matches the last commit to the file?

hbagdi:
ping @mayocream for a review.

.devcontainer/Dockerfile
```diff
@@ -7,3 +7,6 @@ RUN apk add --update \
         build-base \
         bsd-compat-headers \
         m4
+RUN addgroup -S konguser && adduser -S konguser -G konguser
```
mayocream, line 10: let's only use `kong` as the user name.

kikito:
Hi @DiptoChakrabarty do you think you can rebase your PR and change the username from `konguser` to just `kong`?

Thanks!

PR #8818, opened 2022-05-19
fix(clustering) localize config_version to avoid race condition from

fffonion:
yield

### Summary

`update_config` can yield, so we need to cache fields in `self`, otherwise it might got updated (on L89) before we read it (on L208), and causes L202 become false next time, thus prevent next ConfigSync being executed. This probably become more visiable as we introduce more `yield` from #8800.

### Full changelog

*  localize config_version to avoid race condition from yield

### DIFF
--- a/kong/clustering/wrpc_data_plane.lua
+++ b/kong/clustering/wrpc_data_plane.lua
@@ -197,14 +197,15 @@ function _M:communicate(premature)
         end
         local config_table = self.next_config
         local config_hash  = self.next_hash
+        local config_version = self.next_config_version
         local hashes = self.next_hashes
-        if config_table and self.next_config_version > last_config_version then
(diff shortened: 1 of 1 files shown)


fffonion:
Other issus observed, but out of scope of this PR:

- All protocols run `export_deflate_config_payload` regardless of there’s clients connected or not, we can shortcut it so it doesn't unnecessarilu double CPU load.
- Version negotiation not working sometimes, in e2e test json protocol might end up using wrpc protocol.

Suika-Kong:
Not sure if we should include it but: the `data_plane.lua` has different handling of `self.next_hashes` and `self.next_hash` compare to `wrpc_data_plane.lua`.

dndx (approved):

bungle (approved):

PR #8821, opened 2022-05-19
feat(api) add plugin priority to metadata

Tieske:
besides exposing the version, the priority of a plugin is also a property that quite often is unclear to users, so dynamically getting it from the system will help answering those questions.

I didn't include it in the original PR (#8810 ), because of the upcoming dynamic priorities. But after a chat with @jschmid1 it turns out that the hard-coded priorities will remain the default, so still makes sense to expose those as well.

Also; removed the `true` placeholder, which doesn't make sense anymore now that the metadata is an object

### DIFF
--- a/kong/api/routes/kong.lua
+++ b/kong/api/routes/kong.lua
@@ -94,7 +94,8 @@ return {
       local available_plugins = {}
       for name in pairs(singletons.configuration.loaded_plugins) do
         available_plugins[name] = {
-          version = kong.db.plugins.handlers[name].VERSION or true
+          version = kong.db.plugins.handlers[name].VERSION,
+          priority = kong.db.plugins.handlers[name].PRIORITY,
         }
       end
 


Tieske:
CI failures seem unrelated???

jschmid1 (approved):

Suika-Kong:
I got tests failed earlier today and tried to revert this commit, and it works. This PR seems to be the reason for CI failures.

Suika-Kong:
I tried to dump the metadata and find this:
```csv
2022/05/20 17:22:10 [error] 14757#0: *5 [lua] kong.lua:96: fn(): {
  ["priority"] = -inf
}, client: 127.0.0.1, server: kong_admin, request: "GET / HTTP/1.1", host: "127.0.0.1:9001"

...

2022/05/20 17:22:09 [error] 14757#0: *4 [lua] api_helpers.lua:511: handle_error(): ...al/kong-dep/luarocks/share/lua/5.1/lapis/application.lua:424: ...kong-dep/luarocks/share/lua/5.1/kong/api/api_helpers.lua:285: body encoding failed while flushing response: Cannot serialise number: must not be NaN or Infinity
```

It seems that lapis does not support `-inf` and `inf` in the meta data.

Tieske:
@Suika-Kong my bad. Failures seemed non-related, but unfortunately were. Fix is here: https://github.com/Kong/kong/pull/8833

PR #8822, opened 2022-05-19
tests(plugin/key-auth) fix a flaky test

ADD-SP:
### Summary

It is very flaky on Kong-EE.


bungle (approved):
I have approved. When things start to get flaky, unexpectedly, we there is usually a reason somewhere else, or has this always been flaky on EE?

ADD-SP:
> I have approved. When things start to get flaky, unexpectedly, we there is usually a reason somewhere else, or has this always been flaky on EE?

From my observations, it is more stable on CE, but very flaky on EE.

kikito (approved):

PR #8823, opened 2022-05-20
feat(pdk) normalize kong.request.get_path

flrgh:
This changes the behavior of `kong.request.get_path()` such that it performs path normalization on the value before returning it to the caller. The new PDK function `kong.request.get_raw_path()` performs no normalization and contains a notice about its potential security pitfalls when used unwisely.

**Note:** there has been much discussion on the underlying implementation/methodology of our normalization strategy (see #8140). This pull request does not attempt to resolve that discussion and only focuses on the PDK interface. The assumption here is that the implementation of `kong.tools.uri.normalize` can be updated/finalized at a later date if we wish (though some test cases and LuaDoc annotations may need to be updated depending on what changes are made).

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -9,6 +9,7 @@
 local cjson = require "cjson.safe".new()
 local multipart = require "multipart"
 local phase_checker = require "kong.pdk.private.phases"
+local normalize = require("kong.tools.uri").normalize
 
 
 local ngx = ngx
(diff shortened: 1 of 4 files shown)


jschmid1 (commented):
A general +1 on the abstraction for the actual path normalization.

kong/pdk/request.lua
```diff
@@ -370,18 +371,49 @@ local function new(self)
...
+  -- @function kong.request.get_path
+  -- @phases rewrite, access, header_filter, response, body_filter, log, admin_api
+  -- @tparam[opt=true] boolean merge_slashes consolidate duplicate slashes in the path
+  -- @treturn string the path
+  -- @usage
+  -- -- Given a request to https://example.com/t/Abc%20123%C3%B8%2f/parent/..//test/./
+  --
+  -- kong.request.get_path() -- "/t/Abc 123ø%2F/test/"
+  function _REQUEST.get_path()
+    return normalize(_REQUEST.get_raw_path(), true)
```
mayocream, line 395: This would change the default behavior of this API, any possibility to add another API like `get_normalized_path`?

flrgh: This is actually by design. We want `get_path()` to be normalized so that it's secure by default when used for string comparison. Sorry I didn't make this more apparent in the description! This reminds me to add a breaking change entry for the changelog though.

dndx: It is a security fix, so the change is expected.

kong/pdk/request.lua
```diff
@@ -370,18 +371,49 @@ local function new(self)
...
+  -- * Percent-encoded values of unreserved characters are decoded (`%20`
+  --   becomes ` `).
+  -- * Percent-encoded values of reserved characters have their hexidecimal
+  --   value uppercased (`%2f` becomes `%2F`).
+  -- * Relative path elements (`/.` and `/..`) are dereferenced.
+  -- * Duplicate slashes are consolidated (`//` becomes `/`).
+  --
+  -- @function kong.request.get_path
+  -- @phases rewrite, access, header_filter, response, body_filter, log, admin_api
+  -- @tparam[opt=true] boolean merge_slashes consolidate duplicate slashes in the path
```
dndx, line 388: This does not appears to be on the PDK, and is probably unnecessary. We can just remove this line.
```suggestion
```

dndx (approved):
Eventually it will be nice if we can cache the result of the normalized request URI, but that can happen after the router refactor. Approving for now.

PR #8825, opened 2022-05-20
chore(cluster) head request to version negotiation

chronolaw:
### Summary

We do not use `/version-handshake` endpoint, just send a simple `HEAD` request to `/v1/wrpc`,
If it works then we use wrpc, else we fall back to websocket + json.

### Full changelog

* Add a new function `head_request_version` in  `version_negotiation/init.lua`
* Change the behavior of functioin `request_version_negotiation`
* Check the result in `init_worker`

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -206,7 +206,9 @@ local function fill_empty_hashes(hashes)
 end
 
 function _M:request_version_negotiation()
-  local response_data, err = version_negotiation.request_version_handshake(self.conf, self.cert, self.cert_key)
+  local response_data, err =
+    version_negotiation.check_wrpc_support(self.conf, self.cert, self.cert_key)
+
   if not response_data then
     ngx_log(ngx_ERR, _log_prefix, "error while requesting version negotiation: " .. err)
     assert(ngx.timer.at(math.random(5, 10), function(premature)
@@ -214,6 +216,8 @@ function _M:request_version_negotiation()
     end))
     return
   end
+
+  return response_data
 end
 
 
@@ -299,9 +303,9 @@ function _M:init_worker()
         return
       end
 
-      self:request_version_negotiation()
+      local config_proto, msg = self:request_version_negotiation()
+      --local config_proto, msg = version_negotiation.get_negotiated_service("config")
 
-      local config_proto, msg = version_negotiation.get_negotiated_service("config")
       if not config_proto and msg then
         ngx_log(ngx_ERR, _log_prefix, "error reading negotiated \"config\" service: ", msg)
       end
--- a/kong/clustering/version_negotiation/init.lua
+++ b/kong/clustering/version_negotiation/init.lua
@@ -354,4 +354,40 @@ function _M.get_negotiated_service(name)
   return val.version, val.message
 end
 
+--- detect '/v1/wrpc' endpoint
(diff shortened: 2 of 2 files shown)


kong/clustering/version_negotiation/init.lua
```diff
@@ -354,4 +354,40 @@ function _M.get_negotiated_service(name)
   return val.version, val.message
 end
 
+--- detect '/v1/wrpc' endpoint
+--- if there is no '/v1/wrpc', fallback to websocket + json
+function _M.head_request_version(conf, cert, cert_key)
```
Suika-Kong, line 359: Could we name it like `check_wrpc_protocol_support`?

chronolaw: Good idea, but at first we must pass the ci test.

kong/clustering/version_negotiation/init.lua
```diff
@@ -354,4 +354,40 @@ function _M.get_negotiated_service(name)
...
+  local params = {
+    scheme = "https",
+    method = "HEAD",
+
+    ssl_verify = false,
```
dndx, line 364: Seems to be a bad idea.

kong/clustering/init.lua
```diff
@@ -206,14 +206,19 @@ local function fill_empty_hashes(hashes)
 end
 
 function _M:request_version_negotiation()
-  local response_data, err = version_negotiation.request_version_handshake(self.conf, self.cert, self.cert_key)
+  --local response_data, err = version_negotiation.request_version_handshake(self.conf, self.cert, self.cert_key)
```
dndx, line 209: Can we remove this function? Plus this like is too long.

kong/clustering/version_negotiation/init.lua
```diff
@@ -354,4 +354,40 @@ function _M.get_negotiated_service(name)
...
+
+  local c = http.new()
+  local res, err = c:request_uri(
+    "https://" .. conf.cluster_control_plane .. "/v1/wrpc", params)
+  if not res then
+    return nil, err
+  end
+
+  if res.status == 404 then
+    return "v0", "no wrpc endpoint, fallback to /v1/outlet"
```
dndx, line 387: I don't think this error message is even used and should just be removed. Plus why can't the function just be called "check_wrpc_support", `v0` and `v1` seems to be confusing and unnecessary.

Suika-Kong: Yes. v0 and v1 mean nothing to a user. Maybe we could log like this:

chronolaw:
It seems has a ci test bug in this branch, I have to close it and create a new one.

javierguerragiraldez:
how does it (or the next version of this) go with #8775 ?

chronolaw:
> how does it (or the next version of this) go with #8775 ?

This PR is in a very early stage, we will improve it with more discussion.

chronolaw:
Please talk in here. https://github.com/Kong/kong/pull/8834

kong/clustering/version_negotiation/init.lua
chronolaw: In the first version check progress, We need not to verify cert, or else the test case in `06-version_negotiation_spec.lua` will fail, because it use a self-signed cert.

PR #8826, opened 2022-05-20
feat(plugins) opentelemetry plugin

mayocream:
### Summary

Add the OpenTelemetry plugin which pushes the build-in instrumentation to the backend server through the [OTLP/HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp) protocol with Protobuf content type.

_Note: merge using rebase strategy._

FT-2588

### DIFF
--- a/kong/plugins/opentelemetry/handler.lua
+++ b/kong/plugins/opentelemetry/handler.lua
@@ -1,0 +1,199 @@
+local new_tab = require "table.new"
+local http = require "resty.http"
+local clone = require "table.clone"
+local otlp = require "kong.plugins.opentelemetry.otlp"
+local propagation = require "kong.tracing.propagation"
+local tablepool = require "tablepool"
+
+local ngx = ngx
+local kong = kong
+local ngx_log = ngx.log
+local ngx_ERR = ngx.ERR
+local ngx_DEBUG = ngx.DEBUG
+local ngx_now = ngx.now
+local ngx_update_time = ngx.update_time
+local ngx_req = ngx.req
+local ngx_get_headers = ngx_req.get_headers
+local timer_at = ngx.timer.at
+local clear = table.clear
+local propagation_parse = propagation.parse
+local propagation_set = propagation.set
+local tablepool_release = tablepool.release
+local tablepool_fetch = tablepool.fetch
+local null = ngx.null
+local encode_traces = otlp.encode_traces
+local transform_span = otlp.transform_span
+
+local POOL_BATCH_SPANS = "KONG_OTLP_BATCH_SPANS"
+local _log_prefix = "[otel] "
+
+local OpenTelemetryHandler = {
+  VERSION = "0.1.0",
+  PRIORITY = 14,
+}
+
+local default_headers = {
+  ["Content-Type"] = "application/x-protobuf",
+}
+
+-- worker-level spans queue
+local spans_queue = new_tab(5000, 0)
+local headers_cache = setmetatable({}, { __mode = "k" })
+local last_run_cache = setmetatable({}, { __mode = "k" })
+
+local function get_cached_headers(conf_headers)
+  -- cache http headers
+  local headers = default_headers
+  if conf_headers then
+    headers = headers_cache[conf_headers]
+  end
+
+  if not headers then
+    headers = clone(default_headers)
+    if conf_headers and conf_headers ~= null then
+      for k, v in pairs(conf_headers) do
+        headers[k] = v and v[1]
+      end
+    end
+
+    headers_cache[conf_headers] = headers
+  end
+
+  return headers
+end
+
+
+local function http_export_request(conf, pb_data, headers)
+  local httpc = http.new()
+  httpc:set_timeouts(conf.connect_timeout, conf.send_timeout, conf.read_timeout)
+  local res, err = httpc:request_uri(conf.endpoint, {
+    method = "POST",
+    body = pb_data,
+    headers = headers,
+  })
+  if not res then
+    ngx_log(ngx_ERR, _log_prefix, "failed to send request: ", err)
+  end
+
+  if res and res.status ~= 200 then
+    ngx_log(ngx_ERR, _log_prefix, "response error: ", res.status, ", body: ", res.body)
+  end
+end
+
+local function http_export(premature, conf)
+  if premature then
+    return
+  end
+
+  local spans_n = #spans_queue
+  if spans_n == 0 then
+    return
+  end
+
+  local start = ngx_now()
+  local headers = conf.headers and get_cached_headers(conf.headers) or default_headers
+
+  -- batch send spans
+  local spans_buffer = tablepool_fetch(POOL_BATCH_SPANS, conf.batch_span_count, 0)
+
+  for i = 1, spans_n do
+    local len = (spans_buffer.n or 0) + 1
+    spans_buffer[len] = spans_queue[i]
+    spans_buffer.n = len
+
+    if len >= conf.batch_span_count then
+      local pb_data = encode_traces(spans_buffer, conf.resource_attributes)
+      clear(spans_buffer)
+
+      http_export_request(conf, pb_data, headers)
+    end
+  end
+
+  -- remain spans
+  if spans_queue.n and spans_queue.n > 0 then
+    local pb_data = encode_traces(spans_buffer, conf.resource_attributes)
+    http_export_request(conf, pb_data, headers)
+  end
+
+  -- clear the queue
+  clear(spans_queue)
+
+  tablepool_release(POOL_BATCH_SPANS, spans_buffer)
+
+  ngx_update_time()
+  local duration = ngx_now() - start
+  ngx_log(ngx_DEBUG, _log_prefix, "opentelemetry exporter sent " .. spans_n ..
+    " traces to " .. conf.endpoint .. " in " .. duration .. " seconds")
+end
+
+local function process_span(span)
+  if span.should_sample == false
+      or kong.ctx.plugin.should_sample == false
+  then
+    -- ignore
+    return
+  end
+
+  -- overwrite
+  local trace_id = kong.ctx.plugin.trace_id
+  if trace_id then
+    span.trace_id = trace_id
+  end
+
+  local pb_span = transform_span(span)
+
+  local len = spans_queue.n or 0
+  len = len + 1
+
+  spans_queue[len] = pb_span
+  spans_queue.n = len
+end
+
+function OpenTelemetryHandler:rewrite()
+  local headers = ngx_get_headers()
+  local root_span = ngx.ctx.KONG_SPANS and ngx.ctx.KONG_SPANS[1]
+
+  -- make propagation running with tracing instrumetation not enabled
+  if not root_span then
+    local tracer = kong.tracing.new("otel")
+    root_span = tracer.start_span("root")
+
+    -- the span created only for the propagation and will be bypassed to the exporter
+    kong.ctx.plugin.should_sample = false
+  end
+
+  local header_type, trace_id, span_id, _, should_sample, _ = propagation_parse(headers)
+  if should_sample == false then
+    root_span.should_sample = should_sample
+  end
+
+  -- overwrite trace id
+  if trace_id then
+    root_span.trace_id = trace_id
+    kong.ctx.plugin.trace_id = trace_id
+  end
+
+  -- overwrite root span's parent_id
+  if span_id then
+    root_span.parent_id = span_id
+  end
+
+  propagation_set("w3c", header_type, root_span)
+end
+
+function OpenTelemetryHandler:log(conf)
+  ngx_log(ngx_DEBUG, _log_prefix, "total spans in current request: ", ngx.ctx.KONG_SPANS and #ngx.ctx.KONG_SPANS)
+
+  -- transform spans
+  kong.tracing.process_span(process_span)
+
+  local cache_key = conf.__key__
+  local last = last_run_cache[cache_key] or 0
+  local now = ngx_now()
+  if now - last >= conf.batch_flush_delay then
+    last_run_cache[cache_key] = now
+    timer_at(0, http_export, conf)
+  end
+end
+
+return OpenTelemetryHandler
--- a/kong/plugins/opentelemetry/otlp.lua
+++ b/kong/plugins/opentelemetry/otlp.lua
@@ -1,0 +1,169 @@
+require "kong.plugins.opentelemetry.proto"
+local pb = require "pb"
+local new_tab = require "table.new"
+local nkeys = require "table.nkeys"
+local tablepool = require "tablepool"
+local utils = require "kong.tools.utils"
+
+local kong = kong
+local insert = table.insert
+local tablepool_fetch = tablepool.fetch
+local tablepool_release = tablepool.release
+local deep_copy = utils.deep_copy
+local table_merge = utils.table_merge
+
+local POOL_OTLP = "KONG_OTLP"
+local EMPTY_TAB = {}
+
+local PB_STATUS = {}
+for i = 0, 2 do
+  PB_STATUS[i] = { code = i }
+end
+
+local function transform_attributes(attr)
+  if type(attr) ~= "table" then
+    error("invalid attributes", 2)
+  end
+
+  local pb_attributes = new_tab(nkeys(attr), 0)
+  for k, v in pairs(attr) do
+    local typ = type(v)
+    local pb_val
+
+    if typ == "string" then
+      pb_val = { string_value = v }
+
+    elseif typ == "number" then
+      pb_val = { double_value = v }
+
+    elseif typ == "boolean" then
+      pb_val = { bool_value = v }
+
+    else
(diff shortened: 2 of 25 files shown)


kong/tools/utils.lua
```diff
@@ -1443,7 +1443,7 @@ do
     C.clock_gettime(0, nanop)
     local t = nanop[0]
 
-    return tonumber(t.tv_sec) * 100000000 + tonumber(t.tv_nsec)
+    return tonumber(t.tv_sec) * 1000000000 + tonumber(t.tv_nsec)
```
fffonion, line 1446: too many zeros 😄 maybe change to scienfic notation `1e9`

kong/plugins/opentelemetry/proto.lua
```diff
@@ -0,0 +1,481 @@
...
+-- defines the OTLP .proto
+local protoc = require "protoc".new()
+
+local function load_proto(filename, data)
+  assert(protoc:load(data, filename))
+end
+
+-- parse & load proto files globally on module loading
+-- https://github.com/open-telemetry/opentelemetry-proto/blob/v0.18.0/opentelemetry/proto/common/v1/common.proto
+load_proto("opentelemetry/proto/common/v1/common.proto", [[
```
fffonion, line 25: consider move this and following to `kong/include/opentelemetry/proto/common/v1/common.proto`

kong/plugins/opentelemetry/handler.lua
```diff
@@ -0,0 +1,185 @@
...
+  if spans_cache.n and spans_cache.n > 0 then
+    local pb_data = encode_traces(batch_spans, conf.resource_attributes)
+    http_export_request(conf, pb_data, headers)
+  end
+
+  clear(spans_cache)
+
+  tablepool_release(POOL_BATCH_SPANS, batch_spans)
+
+  ngx.update_time()
```
chronolaw, line 114: localize `ngx.update_time` and `ngx.now`

kong/tracing/instrumentation.lua
```diff
@@ -72,33 +72,32 @@ function _M.balancer(ctx)
     local try = balancer_tries[i]
     span = tracer.start_span("balancer try #" .. i, {
       kind = 3, -- client
-      start_time_ns = try.balancer_start * 100000,
+      start_time_ns = try.balancer_start * 1000000,
```
fffonion, line 75: same, use `1e6` here seems cleaner

kong/plugins/opentelemetry/otlp.lua
```diff
@@ -0,0 +1,178 @@
...
+          attributes = {}
+        },
+        scope_spans = {
+          {
+            scope = {
+              name = "kong-internal",
+              version = "0.1.0",
+            },
+            spans = {},
+          }
```
chronolaw, line 148: Should we add `,` after each `}`

kong/tracing/instrumentation.lua
chronolaw, line 75: We should define a constant, such as `xxx = 1000* 1000`.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -0,0 +1,185 @@
...
+  propagation_set("w3c", header_type, root_span)
+end
+
+function OpenTelemetryHandler:log(conf)
+  ngx_log(ngx_DEBUG, _log_prefix, "total spans in current request: ", ngx.ctx.KONG_SPANS and #ngx.ctx.KONG_SPANS)
+
+  -- transform spans
+  kong.tracing.process_span(process_span)
+
+  local last = last_run_cache[conf] or 0
```
fffonion, line 177: discussed offline, `conf` will be a new table everytime, we can use `conf.__key__` to distinguish different plugin instances.

spec/03-plugins/37-opentelemetry/01-otlp_spec.lua
```diff
@@ -0,0 +1,114 @@
...
+local pb = require "pb"
+
+local fmt = string.format
+local rand_bytes = utils.get_rand_bytes
+local time_ns = utils.time_ns
+local insert = table.insert
+local tracer = kong.tracing.new("test")
+
+-- test assert only
+pb.option("enum_as_value")
```
fffonion, line 14: this may affect other tests, let's move this to `setup` and revert it back in `teardown`

kong/tracing/instrumentation.lua
mayocream: It's a more proper way to achieve that, thanks!

spec/03-plugins/37-opentelemetry/01-otlp_spec.lua
mayocream: Ops, it's global and this could be affecting wRPC tests. (or other things using lua-protobuf)

kong/plugins/opentelemetry/otlp.lua
mayocream: I wonder if it is an option that could be in _luacheck_ for all the code style-related.
Thanks, anyway I will solve it now.

kong/plugins/opentelemetry/proto.lua
mayocream: awesome

spec/03-plugins/37-opentelemetry/04-exporter_spec.lua
```diff
@@ -0,0 +1,166 @@
...
+        plugins = "opentelemetry",
+        opentelemetry_tracing = types,
+      })
+
+      proxy_client = helpers.proxy_client(1000)
+    end
+
+    describe("valid #http request", function ()
+      lazy_setup(function()
+        os.execute("fuser -n tcp -k " .. HTTP_PORT)
```
fffonion, line 62: little worried about the leftover behaviour here, is it needed? if so maybe move to lazy_teardown 🤔

kong/tracing/instrumentation.lua
```diff
@@ -72,33 +72,32 @@ function _M.balancer(ctx)
...
-      local upstream_finish_time = (ctx.KONG_BODY_FILTER_ENDED_AT or 0) * 100000
+    if i == try_count and try.state == nil then
+      local upstream_finish_time = ctx.KONG_BODY_FILTER_ENDED_AT and ctx.KONG_BODY_FILTER_ENDED_AT * 1e6
       span:finish(upstream_finish_time)
 
     else
       if not upstream_connect_time then
-        upstream_connect_time = split(var.upstream_connect_time, ",", "jo")
+        upstream_connect_time = split(var.upstream_connect_time, ", ", "jo")
       end
```
Suika-Kong, line 95: It seems a rare condition where this lazy evaluation would work: almost every time the `else` branch is entered.
Could we just evaluate this before the loop?

kong/tracing/instrumentation.lua
```diff
@@ -72,33 +72,32 @@ function _M.balancer(ctx)
...
 
       -- retrying
       if try.balancer_latency ~= nil then
-        local try_upstream_connect_time = (upstream_connect_time[i] or 0) * 1000
-        span:finish((try.balancer_start + try.balancer_latency + try_upstream_connect_time) * 100000)
+        local try_upstream_connect_time = (tonumber(upstream_connect_time[i]) or 0) * 1000
```
Suika-Kong, line 99:
```suggestion
        local try_upstream_connect_time = (tonumber(upstream_connect_time[i] or 0)) * 1000
```

to catch the case where its value is `"non-number-string"`

kong/tracing/instrumentation.lua
```diff
@@ -149,6 +148,7 @@ function _M.http_client()
     end
 
     local span = tracer.start_span("HTTP " .. method .. " " .. uri, {
+      span_kind = 3,
```
Suika-Kong, line 151: This attribute seems not really used in `start_span`.

kong/tracing/instrumentation.lua
```diff
@@ -186,10 +186,11 @@ function _M.request(ctx)
   local req_uri = ctx.request_uri or var.request_uri
 
   local start_time = ngx.ctx.KONG_PROCESSING_START
-      and ngx.ctx.KONG_PROCESSING_START * 100000
+      and ngx.ctx.KONG_PROCESSING_START * 1e6
       or time_ns()
 
   local active_span = tracer.start_span(span_name, {
+    span_kind = 2, -- server
```
Suika-Kong, line 193: Ditto.

kong/tracing/propagation.lua
```diff
@@ -447,11 +447,10 @@ local function set(conf_header_type, found_header_type, proxy_span, conf_default
...
   if conf_header_type == "b3-single" or found_header_type == "b3-single" then
-    set_header("b3", fmt("%s-%s-%s-%s",
-        to_hex(proxy_span.trace_id),
-        to_hex(proxy_span.span_id),
-        proxy_span.should_sample and "1" or "0",
-      to_hex(proxy_span.parent_id)))
+    set_header("b3", to_hex(proxy_span.trace_id) ..
+        "-" .. to_hex(proxy_span.span_id) ..
+        "-" .. (proxy_span.should_sample and "1" or "0") ..
+        (proxy_span.parent_id and "-" .. to_hex(proxy_span.parent_id) or ""))
```
Suika-Kong, line 453:
```suggestion
    set_header("b3", table.concat({
        to_hex(proxy_span.trace_id),
        to_hex(proxy_span.span_id),
        (proxy_span.should_sample and "1" or "0"),
        -- fine if parent_id is nil
        proxy_span.parent_id,
    }, "-"))
```

kong/tracing/propagation.lua
```diff
@@ -465,7 +464,7 @@ local function set(conf_header_type, found_header_type, proxy_span, conf_default
     set_header("uber-trace-id", fmt("%s:%s:%s:%s",
         to_hex(proxy_span.trace_id),
         to_hex(proxy_span.span_id),
-        to_hex(proxy_span.parent_id),
+        proxy_span.parent_id and to_hex(proxy_span.parent_id) or "0",
```
Suika-Kong, line 467:
```suggestion
        to_hex(proxy_span.parent_id or 0),
```

The same reason: to catch the incorrect type.

kong/tracing/instrumentation.lua
Suika-Kong: Also, is the format of this variable fixed? Or maybe we should do this:
```lua 
split(var.upstream_connect_time, ",\\s*", "jo")
```

kong/tools/utils.lua
```diff
@@ -1443,7 +1443,7 @@ do
     C.clock_gettime(0, nanop)
     local t = nanop[0]
 
-    return tonumber(t.tv_sec) * 100000000 + tonumber(t.tv_nsec)
+    return tonumber(t.tv_sec) * 1e9 + tonumber(t.tv_nsec)
```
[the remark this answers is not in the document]
mayocream: It was intended to fix the time precision problem.

kong/tracing/instrumentation.lua
mayocream: https://github.com/nginx/nginx/blob/bfc5b35827903a3c543b58e4562db8b62021c164/src/http/ngx_http_upstream.c#L5758-L5760
According to the Nginx implementation, it will concat the string with `, `.

```lua
local ngx_re = require "ngx.re"
local split = ngx_re.split

local test_str = "1000, 1000, 1000"

for _, v in ipairs(split(test_str, ", ", "jo")) do
    print(v)
end
```
output: 
```bash
$ resty test/split.lua 
1000
1000
1000
```

kong/plugins/opentelemetry/handler.lua
```diff
@@ -0,0 +1,188 @@
...
+
+  if not headers then
+    headers = clone(default_headers)
+    if conf.headers and conf.headers ~= null then
+      for k, v in pairs(conf.headers) do
+        headers[k] = v and v[1]
+      end
+    end
+    headers_cache[conf.headers] = headers
+  end
```
Suika-Kong, line 88: Better to extract this into a function. And I doubt if this optimization would be significant. 
And maybe it's simpler to add the `content-type` into cloned header table.

kong/tracing/instrumentation.lua
mayocream: Actually, the original case was to avoid the "non-number-string".
`tonumber` on a non-number string will result in a nil value.

```bash
$ resty -e "print(type(tonumber('-')))"
nil
```

kong/plugins/opentelemetry/handler.lua
```diff
@@ -0,0 +1,188 @@
...
+    ngx_log(ngx_ERR, _log_prefix, "response error: ", res.status, ", body: ", res.body)
+  end
+end
+
+local function http_export(premature, conf)
+  if premature then
+    return
+  end
+
+  local spans = spans_cache
```
Suika-Kong, line 66: We are not using the variable(alias) except for the next line.

kong/tracing/instrumentation.lua
mayocream: https://github.com/Kong/kong/blob/4d29046008daa56b717b69b94acfec76c635a7cf/kong/pdk/tracing.lua#L178
Thanks for porting it out, it should be `options.span_kind` there.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -0,0 +1,188 @@
...
+  VERSION = "0.1.0",
+  PRIORITY = 14,
+}
+
+local default_headers = {
+  ["Content-Type"] = "application/x-protobuf",
+}
+
+-- worker-level spans cache
+local spans_cache = new_tab(5000, 0)
```
Suika-Kong, line 40: This table is more like a buffer than `cache`.
A better name would be `not_synced_spans`, or just `spans_buffer`.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -0,0 +1,188 @@
...
+        headers[k] = v and v[1]
+      end
+    end
+    headers_cache[conf.headers] = headers
+  end
+
+  -- batch send spans
+  local batch_spans = tablepool_fetch(POOL_BATCH_SPANS, conf.batch_span_count, 0)
+
+  for i = 1, spans_cache.n do
```
Suika-Kong, line 93: It seems `spans_cache.n` is always not nil here. So why don't we use this instead of `local spans_n = #spans` at #l67?

kong/tracing/propagation.lua
mayocream: Did a test and seems the `table.concat` has lower performance.
https://gist.github.com/mayocream/8c48b5a27bedb3409434a118bff2120a

kong/tracing/propagation.lua
mayocream: It is by design, `to_hex("0")` will result in `30` which is not expected.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -0,0 +1,188 @@
...
+    batch_spans[len] = spans_cache[i]
+    batch_spans.n = len
+
+    if len >= conf.batch_span_count then
+      local pb_data = encode_traces(batch_spans, conf.resource_attributes)
+      clear(batch_spans)
+
+      http_export_request(conf, pb_data, headers)
+    end
+  end
```
Suika-Kong, line 104:
```suggestion
  local batch_spans = {}
  local batch_spans_n = 0

  for i = 1, spans_cache.n do
    batch_spans_n = batch_spans_n + 1
    batch_spans[batch_spans_n] = spans_cache[i]

    if len >= conf.batch_span_count then
      batch_spans.n = batch_spans_n
      local pb_data = encode_traces(batch_spans, conf.resource_attributes)
      clear(batch_spans)
      local batch_spans_n = 0

      http_export_request(conf, pb_data, headers)
    end
  end
```

kong/tracing/propagation.lua
Suika-Kong: It sounds like a problem that should be fixed like this:
```lua
to_hex(tonumber(proxy_span.parent_id or 0)),
```

kong/tracing/propagation.lua
Suika-Kong: This way it has better readablity.

spec/03-plugins/37-opentelemetry/04-exporter_spec.lua
mayocream: Moved to lazy_teardown.

kong/plugins/opentelemetry/handler.lua
mayocream: spans_buffer sounds better

kong/plugins/opentelemetry/handler.lua
mayocream: Thanks

kong/plugins/opentelemetry/handler.lua
mayocream: I think using a default table seems more clear.

fffonion (approved):

PR #8829, opened 2022-05-20
Chore/same proto

Suika-Kong:
This PR lets CE share the same proto definitions from EE, as we decide not to have separate ones.

<s>To pass the test I reverted a commit. This is not supposed to be a part of this PR, and we should wait until that is fixed.</s>

Also, we decide to reserve till 100.

### DIFF
--- a/kong/include/kong/model/config.proto
+++ b/kong/include/kong/model/config.proto
@@ -32,4 +32,6 @@ message Config {
 
   repeated Workspace workspaces = 12;
   repeated Parameter parameters = 13;
+
+  repeated KeyAuth keyauth_credentials = 14;
 }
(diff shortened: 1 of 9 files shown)


kong/include/kong/model/config.proto
```diff
@@ -32,4 +32,6 @@ message Config {
 
   repeated Workspace workspaces = 12;
   repeated Parameter parameters = 13;
+
+  repeated KeyAuth keyauth_credentials = 14;
```
fffonion, line 36: let's give it some room for other potential core entities maybe let keyauth_credentials start from 100 as well; on the other side, let's add other known custom DAO entities here as well.

PR #8830, opened 2022-05-20
chore(deps) bump: grpcurl from 1.8.5 to 1.8.6 :tea:

mrnonz:
### Summary
* Some bugs have been addressed in the library used to parse proto source files. Previously grpcurl would accept proto source files that could not actually be compiled with protoc. The converse could also happen: grpcurl could reject some proto source files that could successfully be compiled with protoc. More details can be found in the release notes for the changes to the protoparse library, versions [v1.10.2](https://github.com/jhump/protoreflect/releases/tag/v1.10.2) and [v1.10.3](https://github.com/jhump/protoreflect/releases/tag/v1.10.3).
* Some implementations of the server reflection service have been observed to return multiple (even superfluous) file descriptors, in response to requests made by grpcurl. These extra files, if not returned in a particular order, would cause grpcurl to report an error that the service or method to be invoked could be not be resolved. The reflection client in grpcurl is now more robust to this condition and can handle responses with file descriptors in any order, so it should be interoperable with a larger variety of servers.
* When a request message includes a field of type google.protobuf.Value and a value for that field that was a JSON array, grpcurl would incorrectly interpret the JSON array as if it were a single atomic value, the last value that was in the array. This has been fixed.
* When a response message includes non-printable characters or code points outside the ASCII 7-bit range in the name of a field in a JSON object, it could be improperly encoded with escape characters that are not valid JSON. Standard tools/libraries could then fail to parse the JSON output from grpcurl. This has been fixed.


bungle:
Duplicate of #8593. @mrnonz if you want to continue on this, please resolve issues with our test suite too.

PR #8834, opened 2022-05-20
chore(cluster) head request to check wrpc endpoint

chronolaw:
### Summary

We do not use `/version-handshake` endpoint, just send a simple `HEAD` request to `/v1/wrpc`,
If it works then we use wrpc, else we fall back to websocket + json.

If this PR is merged, I will remove the code and test case of `version_negotiation`.
Now I just keep it for compatibility.

### Full changelog

* Add a new function `head_request_version` in  `version_negotiation/init.lua`
* Change the behavior of functioin `request_version_negotiation`
* Check the result in `init_worker`


javierguerragiraldez:
this has the same issue that forced us to move version negotiation as a wRPC service and not as a HTTP request:  the websocket and http client libraries won't use the same connection, so there's no guarantee both will go to the same server.

PR #8775 implements the inline negotiation and does the equivalent of this PR: "try the wRPC connection and fall back to old protocol if it can't connect"

dndx:
@javierguerragiraldez What is the advantage of "keeping the connection on the same backend server". It seems to be very fragile and I don't know why would anyone want to deploy their CP cluster this way.

javierguerragiraldez:
> @javierguerragiraldez What is the advantage of "keeping the connection on the same backend server".

Otherwise it requires shared state even before establishing the connection, as the CP has to carry the negotiation result from one event to another.  If there's no guarantee that it's the same CP node, it has to be in a database.

When it's part of the wRPC itself, on the other hand, it seamlessly activates the right services (and version) for the connection, without any convoluted state management.

kong-2.8.0-0.rockspec
```diff
@@ -72,6 +72,7 @@ build = {
     ["kong.clustering.control_plane"] = "kong/clustering/control_plane.lua",
     ["kong.clustering.wrpc_data_plane"] = "kong/clustering/wrpc_data_plane.lua",
     ["kong.clustering.wrpc_control_plane"] = "kong/clustering/wrpc_control_plane.lua",
+    ["kong.clustering.wrpc_check"] = "kong/clustering/wrpc_check.lua",
```
dndx, line 75: I wonder if it is necessary to add a new file for such a simple function, can it just be inside `wrpc_data_plane.lua`?

kong/clustering/wrpc_check.lua
```diff
@@ -0,0 +1,39 @@
...
+
+  local c = http.new()
+  local res, err = c:request_uri(
+    "https://" .. conf.cluster_control_plane .. "/v1/wrpc", params)
+  if not res then
+    return nil, err
+  end
+
+  if res.status == 404 then
+    return "v0"
```
dndx, line 33: This is a bit strange, the function is called `check_wrpc_support` and one would not expect the return to be `v0` or `v1`. Maybe a simple `true`/`false` would be enough, or the function should be called something like `check_protocol_support` instead.

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -16,8 +16,16 @@ local CLUSTERING_SYNC_STATUS = require("kong.constants").CLUSTERING_SYNC_STATUS
 local KEY_AUTH_PLUGIN
 
 
+local confs = {
+  wrpc = "spec/fixtures/custom_nginx.template",
+  json = "spec/fixtures/json_nginx.template",
+}
+-- disable wrpc in CP
+os.execute(string.format("cat %s | sed 's/wrpc/json/g' > %s", confs.wrpc, confs.json))
```
dndx, line 24:
```suggestion
os.execute(string.format("cat %s | sed 's/wrpc/foobar/g' > %s", confs.wrpc, confs.json))
```

is probably better? It is not `json` after all and this name might be confusing.

kong-2.8.0-0.rockspec
chronolaw: good idea, I move it to init.lua

kong/clustering/wrpc_check.lua
chronolaw: OK, I rename it to `check_protocol_support`.

chronolaw: OK, I rename it to `check_protocol_support`.

kong-2.8.0-0.rockspec
chronolaw: good idea, I move it to init.lua

kong/clustering/init.lua
```diff
@@ -205,15 +206,57 @@ local function fill_empty_hashes(hashes)
   end
 end
 
+
+--- detect '/v1/wrpc' endpoint
+--- if there is no '/v1/wrpc', fallback to websocket + json
```
dndx, line 211: This documentation is not 100% accurate, it describes the goal, but not what this function does.
```suggestion
--- Return the highest supported Hybrid mode protocol version.
```

Also, does this needs to be inside `init.lua`? The CP shouldn't need this functionality.

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -16,8 +16,16 @@ local CLUSTERING_SYNC_STATUS = require("kong.constants").CLUSTERING_SYNC_STATUS
 local KEY_AUTH_PLUGIN
 
 
+local confs = {
+  wrpc = "spec/fixtures/custom_nginx.template",
+  json = "spec/fixtures/json_nginx.template",
```
dndx, line 21: Should this be written to `/tmp/` instead? Otherwise this will create additional files inside  `spec/fixtures/` that might be accidentally committed (or add `.gitignore`).

Also, we can change the filename to `custom_nginx_no_wrpc.template` to avoid consuming people if they seen this file under `fixtures` during development.

dndx: Also, consider making this a function in `helper` instead, this seems to be very duplicative between 3 files right now.

You can define a function `helpers.get_clustering_protocols` that returns this table and sets up the patched file at the same time.

kong/clustering/init.lua
chronolaw: I consider moving it to `utils.lua`, but it may cause conflict with PR #8861.
So I want to do it in #8861, Do you think it is OK?

dndx: Could it be moved to `wrpc_data_plane.lua`?

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
chronolaw: done

spec/helpers.lua
```diff
@@ -2791,6 +2791,20 @@ local function clustering_client(opts)
 end
 
 
+--- Return a table of clustering_protocols and nginx template file
```
dndx, line 2794:
```suggestion
--- Return a table of clustering_protocols and
-- create the appropriate Nginx template file if needed.
-- The file pointed by `json`, when used by CP,
-- will cause CP's wRPC endpoint be disabled.
```

kong/clustering/init.lua
chronolaw: I think it is not suitable for `wrpc_data_plane.lua`, because at this point we have not decided which module will be required.

dndx (approved):

kong/clustering/init.lua
```diff
@@ -205,15 +206,56 @@ local function fill_empty_hashes(hashes)
...
+
+  if res.status == 404 then
+    return "v0"
+  end
+
+  return "v1"   -- wrpc
+end
+
+
 function _M:request_version_negotiation()
```
dndx, line 246: Discussed with @chronolaw offline, this should probably be removed as the name is misleading.

dndx (changes requested):

kong/clustering/init.lua
chronolaw: done

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
javierguerragiraldez: this sidesteps the need for the user to override the protocol

PR #8835, opened 2022-05-20
docs(changelog) add breaking change notice for #8796

bungle:


CHANGELOG.md
```diff
@@ -76,6 +76,12 @@
 - Deprecate/stop producing Debian 8 "Jessie" containers and packages (EOLed June 2020)
   [Kong/kong-build-tools #448](https://github.com/Kong/kong-build-tools/pull/448)
   [Kong/kong-distributions #766](https://github.com/Kong/kong-distributions/pull/766)
+- Kong schema library's `process_auto_fields` function will not any more make a deep
+  copy of data that is passed to it when the given context is `"select"`. This was
+  done to avoid excessive deep copying of tables where we believe the data most of
+  the time comes from a driver like `pgmoon` or `lmdb`. This was done for performance
+  reasons. Deep copying on `"select"` context can still be done before calling this
+  function. [#8796](https://github.com/Kong/kong/pull/8796)
```
kikito, line 84: Here's my suggested edit:

```suggestion
- Kong schema library's `process_auto_fields` function will modify its first parameter in-place 
  instead of returning a deep copy of it when performing a read operation (context is `"select"`). This 
  performance optimization saves a lot of time when reading data from `pgmoon` or `lmdb`, 
  and should be harmless for almost everyone else. If you are calling `process_auto_fields` in your 
  plugins and  directly want to keep the old behavior you should create a deep copy of the data yourself
  of your data before passing it to `process_auto_fields`.
  [#8796](https://github.com/Kong/kong/pull/8796)
```

PR #8836, opened 2022-05-20
Validate plugin PRIORITY and VERSION fields upon loading

Tieske:
Adds stricter checks on plugins when loading. Priority must be a valid integer (not `inf` nor `NaN`), and the version must be parseble as `x.y.z`.

This means the `pre-functions` plugin now has a `PRIORITY = 1000000`, which previously was `+inf`.

**review**: best reviewed without whitespace changes

### DIFF
--- a/kong/db/dao/plugins.lua
+++ b/kong/db/dao/plugins.lua
@@ -4,7 +4,7 @@ local DAO = require "kong.db.dao"
 local plugin_loader = require "kong.db.schema.plugin_loader"
 local reports = require "kong.reports"
 local plugin_servers = require "kong.runloop.plugin_servers"
-
+local version = require "version"
 
 local Plugins = {}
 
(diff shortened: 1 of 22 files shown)


kong/db/dao/plugins.lua
```diff
@@ -126,31 +126,76 @@ local function implements(plugin, method)
...
 
-  local plugin_handler = "kong.plugins." .. plugin .. ".handler"
-  local ok, handler = utils.load_module_if_exists(plugin_handler)
-  if not ok then
-    ok, handler = plugin_servers.load_plugin(plugin)
-    if type(handler) == "table" then
-      handler._go = true
+  local function valid_priority(prio)
+    if type(prio) ~= "number" or
+       prio == (0/0) or  -- NaN
```
bungle, line 133:
```suggestion
       prio ~= prio or  -- NaN
```

Tieske: Same thing no? No strong opinions. Both work. If this is all then please merge the suggestion and then the PR.

bungle: not same, correction. i don’t think both works.

fffonion:
@Tieske Does this mean all external plugins needs to have Version and Priority defined?

Tieske:
> @Tieske Does this mean all external plugins needs to have Version and Priority defined?

Yes, so it's a breaking change. But it's for the better imho since it allows for better control of plugins, they should have it anyway. Only thus far they got away without it.

bungle (approved):

PR #8841, opened 2022-05-23
feat(termination) add avert + trigger/avert _value conf

jeremyjpj0916:
### Summary

This PR adds two concepts:
1. Allowing the value to be specified per header/query param for more specific use cases.
2. The avert which acts as a polar opposite to trigger. Dictating when to prevent the plugin execution.

Relevant Docs PR:
https://github.com/Kong/docs.konghq.com/pull/3965


jeremyjpj0916:
Hmm seems failures are not even based on my plugin enhancements? Odd considering this is the master branch which I would expect stable.

Tieske (changes requested):
lgtm. Just not sure about the name; avert, or maybe skip_trigger? (not a native speaker)

would you mind rebasing the PR @jeremyjpj0916 ?

hbagdi:
Please use header-based routing for this purpose.

jeremyjpj0916:
I don't think header based routing solves what this PR solves. This PR is all about being able to allow traffic you want and reject traffic you don't based on a seen header.

hbagdi:
> This PR is all about being able to allow traffic you want and reject traffic you don't based on a seen header.

Can you elaborate a bit more on why header-based route matching (and then enabling a plugin on that route) doesn't work for your use-case?

PR #8842, opened 2022-05-23
chore(conf): deprecate <strict> worker_consistency

jschmid1:
Signed-off-by: Joshua Schmid <jaiks@posteo.de>

### Summary

Deprecate <strict> worker_consistency config option

### Full changelog

* The "strict" option for config option "worker_consistency" is now deprecated and defaults to "eventual" now

Internal: https://konghq.atlassian.net/browse/FT-2537


jschmid1:
This has a known impact on the router tests.

bungle:
@jschmid1 quick fix to this is just to add `strict` here:
https://github.com/Kong/kong/blob/master/spec/kong_tests.conf

Then spend "bug fixing period" trying to get it removed so that we always run tests with eventual too.

bungle:
@jschmid1 please add changelog entry about it too.

spec/kong_tests.conf
```diff
@@ -38,4 +40,4 @@ lua_package_path = ./spec/fixtures/custom_plugins/?.lua;./spec/fixtures/custom_v
 
 untrusted_lua = sandbox
 
-vaults = bundled
+vaults = bundled
```
jschmid1, line 43: added by mistake, removing in next commit

jschmid1:
The `continous-integration/jenkins/pr-merge` seems to be faulty. 

See -> https://internal.builds.konghq.com/blue/organizations/jenkins/kong/detail/PR-8842/9/pipeline#step-22-log-1715

kikito (approved):
This LGTM, please merge when/if tests pass, or after we fix jenkins

jschmid1:
Something [along those lines ](https://github.com/Kong/kong-build-tools/pull/475) will unblock the last remaining failing test

.requirements
```diff
@@ -8,5 +8,5 @@ RESTY_OPENSSL_VERSION=1.1.1o
 RESTY_PCRE_VERSION=8.45
 RESTY_LMDB_VERSION=master
 LIBYAML_VERSION=0.2.5
-KONG_BUILD_TOOLS_VERSION=4.28.0
```
jschmid1, line 11: temp change to test https://github.com/Kong/kong-build-tools/pull/475

.requirements
```diff
@@ -8,5 +8,5 @@ RESTY_OPENSSL_VERSION=1.1.1o
...
+KONG_BUILD_TOOLS_VERSION=master
```
jschmid1, line 11: https://github.com/Kong/kong-build-tools/pull/475

ADD-SP, line 11: Master branche are mutable. Can we wait for a new version of KBT and use the latest version.

jschmid1: Mentioned in this comment here -> https://github.com/Kong/kong/pull/8842#discussion_r883556550

This is only a temporary change and will be reverted once KBT released a version with https://github.com/Kong/kong-build-tools/pull/475 included.

dndx:
Should only the `strict` option be deprecated or maybe the entire `worker_consistency` config can be deprecated instead?

kong.conf.default
```diff
@@ -1374,6 +1374,8 @@
                                  # - `strict`: the router will be rebuilt
                                  #   synchronously, causing incoming requests to
                                  #   be delayed until the rebuild is finished.
+                                 #   (This option is deprecated and will be
+                                 #    removed in future releases)
```
kikito, line 1378: Agreed with @dndx : If we are going to leave `eventual` as the only option available, we should deprecate `working_consistency` completely, not just the `strict` value.

kikito (changes requested):

kong.conf.default
kikito: In that case, I have reworded the changelog entry to make that more explicit. (See suggestion).

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -591,6 +591,32 @@ describe("kong start/stop #" .. strategy, function()
...
+      assert(helpers.start_kong(kopts, nil, nil))
+    end
+
+    after_each(function()
+      helpers.stop_kong(nil, true)
+    end)
+
+    it("deprecate <worker_consistency>", function()
+      start_kong({worker_consistency = "strict"})
+      assert.errlog().has.line([[the configuration value 'worker_consistency' for configuration property worker_consistency is deprecated]])
```
jschmid1, line 617: figuring out why the logfile does not contain

```
        FAILED  spec/02-integration/02-cmd/02-start_stop_spec.lua:615: kong start/stop #postgres deprecated properties deprecate <worker_consistency>
spec/02-integration/02-cmd/02-start_stop_spec.lua:617: Expected file at:
(string) '/home/runner/work/kong/kong/servroot/logs/error.log'
To match:
(string) 'the configuration value 'worker_consistency' for configuration property worker_consistency is deprecated'

```

jschmid1: (it's printed to the logs when executed in a non-test env)

.requirements
jschmid1: resolved

bungle:
> Should only the `strict` option be deprecated or maybe the entire `worker_consistency` config can be deprecated instead?

I think the whole option has to gone. `worker_state_update_frequency` can stay for now.

bungle (approved):

PR #8843, opened 2022-05-23
feat(proxies) add shared code/schema for proxy management

Tieske:
there are many places where http-proxies configuration is used. This code creates a shread schema so they can all be configured in the same way.

The way this schema works is that it will allow the user to specify for which (http and/or https) connections a proxy should be used. If a proxy is used it will get the config from the user configuration, or if not present it will get the info from the common `HTTP(S)_PROXY` environment variables.
If neither are configured, a runtime error will be returned to the user (500 + logged error).

Status: concept code for discussion purposes, tests are not expected to pass

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -7,6 +7,7 @@ local cjson = require "cjson.safe"
 local meta = require "kong.meta"
 local constants = require "kong.constants"
 local request_util = require "kong.plugins.aws-lambda.request-util"
+local get_proxy_opts = require("kong.tools.proxies").get_proxy_opts
 local kong = kong
 
 local VIA_HEADER = constants.HEADERS.VIA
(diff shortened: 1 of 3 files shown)


Tieske:
@hbagdi @aboudreault based on https://github.com/Kong/kong-ee/pull/2950/files#r806006439 , please have a look.

aboudreault (commented):
looks good overall @Tieske. Are we OK with the fact that aws-lambda uses only a `https` proxy but will now have a the full schema? And also that a user will be able to configure just a http proxy (which will do nothing)?

Tieske:
> Are we OK with the fact that aws-lambda uses only a https proxy but will now have a the full schema? And also that a user will be able to configure just a http proxy (which will do nothing)?

I think that's the tradeoff we're making here. Standardized set up, but accepting some corner cases.

Tieske:
just discussed with @jschmid1 ;

- drop the `all` option, make it an array, and every option should be specified (this allows for future extension, eg. `tcp`, or `tls`)
- don't create a single schema, instead create a function called with supported schemes, that returns a schema that only has those schemes

aboudreault:
yep, that option would be better! :+1:

hbagdi:
@Tieske has had a change of focus so I'm assuming he won't be driving this.

PR #8848, opened 2022-05-24
fix(zipkin) include connect time in balancer span

mayocream:
### Summary

include connect time in balancer span

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,8 +64,6 @@
 
 ## Unreleased
 
-### Fixes
-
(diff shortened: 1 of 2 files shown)


gszr (approved):

mayocream:
I'll revisit this later; we should make it the same behavior as the tracing pdk build-in instrumentation (in this case, only for balancer spans).

kikito (approved):

kikito:
I have approved this, but please include a `fix` entry on the Changelog as well.

PR #8849, opened 2022-05-24
fix(wrpc) `remote_peer`(channel) does not work

Suika-Kong:
Also, fix invalid error messages and clean unused code.

**This PR's target is covered by another PR. Don't merge.**

`remote_peer` is used to access the wRPC connection out of the thread(Nginx timer) that created it. This will be used to implement other services that wRPC supports.

We decide to just use the connection within the same worker and leave the inter-worker problem to the user. So the fix could be much simpler.

This PR is a temporary solution. Eventually, every service including config sync will be using an interface like this and decoupled from the managing of connection.

We will decide how we handle the connection loss or other connection errors:
1. Every time the thread handling wRPC calls encounters an error it reconnects and blocks until  the connection restores, or
2. The clustering module is responsible to reconnect, and clients are aware of the connection loss

And should we just return an error for `call_wait`, or wait for more retries?

### DIFF
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -26,7 +26,6 @@ lua_shared_dict kong_core_db_cache_miss     12m;
 lua_shared_dict kong_db_cache               ${{MEM_CACHE_SIZE}};
 lua_shared_dict kong_db_cache_miss          12m;
 > if role == "data_plane" then
-lua_shared_dict wrpc_channel_dict           5m;
 > end
 > if database == "cassandra" then
 lua_shared_dict kong_cassandra              5m;
--- a/kong/templates/nginx_kong.lua
+++ b/kong/templates/nginx_kong.lua
@@ -25,8 +25,6 @@ lua_shared_dict kong_core_db_cache          ${{MEM_CACHE_SIZE}};
 lua_shared_dict kong_core_db_cache_miss     12m;
 lua_shared_dict kong_db_cache               ${{MEM_CACHE_SIZE}};
 lua_shared_dict kong_db_cache_miss          12m;
-> if role == "data_plane" then
-> end
 > if database == "cassandra" then
 lua_shared_dict kong_cassandra              5m;
 > end
--- a/kong/tools/wrpc.lua
+++ b/kong/tools/wrpc.lua
@@ -12,7 +12,6 @@ local table_remove = table.remove
 local exiting = ngx.worker.exiting
 
 local DEFAULT_EXPIRATION_DELAY = 90
-local CHANNEL_CLIENT_PREFIX = "wrpc_client_"
 
 pb.option("no_default_values")
 
--- a/kong/tools/wrpc.lua
+++ b/kong/tools/wrpc.lua
@@ -255,13 +255,17 @@ wrpc_peer.__index = wrpc_peer
 
 local channels = {}
 
+local function is_wsclient(conn)
+  return conn and not conn.close or nil
+end
+
 --- a `peer` object holds a (websocket) connection and a service.
 function wrpc.new_peer(conn, service, opts)
   local ret = setmetatable({
     conn = conn,
     service = service,
     seq = 1,
-    request_queue = Queue.new(),
+    request_queue = is_wsclient(conn) and Queue.new(),
     response_queue = {},
     closing = false,
     _receiving_thread = nil,
--- a/kong/tools/wrpc.lua
+++ b/kong/tools/wrpc.lua
@@ -169,7 +169,6 @@ function wrpc_service:add(service_name)
         if type_annotations then
           local tag_key, tag_value = annotation:match("^%s*(%S-)=(%S+)%s*$")
           if tag_key and tag_value then
-            tag_value = tag_value
             local tags = type_annotations[identifier] or {}
             type_annotations[identifier] = tags
             tags[tag_key] = tag_value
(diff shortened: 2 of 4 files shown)


Suika-Kong:
Also, there is a problem I don't know how to fix. If you send an RPC with a `remote_peer` that knows the RPC, but the actual `peer` does not know, the actual peer will fail to receive a response without any error message.

Suika-Kong:
Still, `remote_peer` is far from ready. You cannot use `remote_call` with it.

kong/templates/nginx_kong.lua
```diff
@@ -26,7 +26,6 @@ lua_shared_dict kong_core_db_cache_miss     12m;
 lua_shared_dict kong_db_cache               ${{MEM_CACHE_SIZE}};
 lua_shared_dict kong_db_cache_miss          12m;
 > if role == "data_plane" then
-lua_shared_dict wrpc_channel_dict           5m;
 > end
```
javierguerragiraldez, line 29: the `> if` and `> end` lines should go away too

kong/tools/wrpc.lua
```diff
@@ -645,41 +631,22 @@ function wrpc_peer:get_response(req_id)
...
 end
 
-local function remote_close(self)
-  self.closing = true
+function wrpc.new_remote_client(channel_name)
+  local ret = setmetatable({
+    channel_name = channel_name
+  }, wrpc_remote_peer)
+  return ret
 end
```
javierguerragiraldez, line 650: as i understand it, the `channels` table is just a registry of peers.  given that this won't work across workers, what's the need for the wrapper object? why not just pick the "real" peer from the registry?

kong/tools/wrpc.lua
```diff
@@ -255,23 +253,28 @@ local wrpc_peer = {
...
 --- a `peer` object holds a (websocket) connection and a service.
 function wrpc.new_peer(conn, service, opts)
-  opts = opts or {}
-  return setmetatable(merge({
+  local ret = setmetatable({
     conn = conn,
     service = service,
     seq = 1,
-    request_queue = is_wsclient(conn) and Queue.new(),
+    request_queue = Queue.new(),
```
javierguerragiraldez, line 264: the websocket library is different for "client" and "server" connections.  only clients require a transmit thread, and therefore a queue.  in fact,  the presence of the queue is checked later to spawn the transmit thread.  this change means "server" connections would now pass through this queue and thread.

or is it now required because of the "shallow remotes"?

javierguerragiraldez (changes requested):

kong/tools/wrpc.lua
Suika-Kong: It's a mistake. I was trying to let `remote_peer` own its queue, and that solution is abandoned.

kong/tools/wrpc.lua
Suika-Kong: For the first question: we choose to not support access from other workers (currently all DP/CP communication is done within a privileged worker).
For the second question, if the connection is lost, the config service is aware of the reconnection (and handles it), but the user of `remote_peer` won't know that. It's important for `remote_peer` to work with reconnection.

Suika-Kong: I think eventually we will make the config service in the same position as other services, and the reconnection should be handled by a separate thread, or automatically by the wRPC implementation. But for now I'll just leave it as it is.

javierguerragiraldez: my question is: why use a wrapper object at all?  since it's limited to the same worker, it should just use the same peer object

Suika-Kong: > my question is: why use a wrapper object at all? since it's limited to the same worker, it should just use the same peer object

The wrapper checks the connections' existence, so users of the "remote" connection don't need to check for existence every time. This may be not that important but it's useful. (Maybe the peer will be able to handle reconnection and we can always use the "real" peer, and then we don't need this anymore, but we will have a similar interface.)

Suika-Kong: Or rather we can always return the same object for the same name of channel, like this:
```lua
local channels = {}
function wrpc.new_peer(name, conn, service, opts)
  local ret = channels[name] or {}
  ret.conn = conn or ret.conn,
  ret.service = service or ret.service
  if opts  then
    merge(ret, opts)
  end
  channels[name] = ret
  return setmetatable(ret, wrpc_peer)
end
```

kong/tools/wrpc.lua
```diff
@@ -645,41 +635,22 @@ function wrpc_peer:get_response(req_id)
...
+  local c = channels[self.channel_name]
+  if not c then
+    ngx.log(ngx.ERR, "channel \"", self.channel_name, "\" does not exist when trying to call ", name, " from client")
   end
-
-  if payload_back.mtype == "MESSAGE_TYPE_RPC" then
-    local rpc = self.service:get_method(payload_back.svc_id, payload_back.rpc_id)
-    return decodearray(self.decode, rpc.output_type, payload_back.payloads)
+  if c[name] then
+    return c[name]
```
chronolaw, line 645: A simple `return c[name]` may be enough?

kong/tools/wrpc.lua
```diff
@@ -645,41 +635,22 @@ function wrpc_peer:get_response(req_id)
...
   end
-
-  return nil, "unknown message type"
 end
 
-local function remote_close(self)
-  self.closing = true
+function wrpc.new_remote_client(channel_name)
+  local ret = setmetatable({
```
chronolaw, line 650: `ret` => `peer`, we should have a reasonable name.

kong/tools/wrpc.lua
Suika-Kong: And I think maybe we should decouple the "channel" concept.
For example, use `kong.dp_connection` and `kong.cp_connection` to get the wRPC connection to the remote peer.

Suika-Kong:
We decided to use another method.

PR #8850, opened 2022-05-24
chore(deps) bump openresty from 1.19.9.1 to 1.21.4.1

bungle:
### Summary

Bump OpenResty from 1.19.9.1 to 1.21.4.1, see:
https://openresty.org/en/changelog-1021004.html

### See Also

This need to be reviewed and merged first, and proper releases of it needs to be tagged:
https://github.com/Kong/kong-build-tools/pull/473


.requirements
```diff
@@ -2,11 +2,11 @@ KONG_PACKAGE_NAME=kong
...
 
-RESTY_VERSION=1.19.9.1
+RESTY_VERSION=1.21.4.1
 RESTY_LUAROCKS_VERSION=3.9.0
 RESTY_OPENSSL_VERSION=1.1.1o
 RESTY_PCRE_VERSION=8.45
 RESTY_LMDB_VERSION=master
 LIBYAML_VERSION=0.2.5
-KONG_BUILD_TOOLS_VERSION=4.25.3
+KONG_BUILD_TOOLS_VERSION=feat/openresty-1.21.4.1-patches
```
bungle, line 11: This needs to be properly tagged, when we have a new build tools release.

kong/meta.lua
```diff
@@ -20,6 +20,6 @@ return {
   -- third-party dependencies' required version, as they would be specified
   -- to lua-version's `set()` in the form {from, to}
   _DEPENDENCIES = {
-    nginx = { "1.19.3.1", "1.19.9.1" },
+    nginx = { "1.21.4.1" },
```
fffonion, line 23: not a blocker, but i'm curious on the reason behind this restriction, is there a specific behaviour we rely on that doesn't work with previous versions of openresty?

bungle: @fffonion It is quite major change in Nginx, the behavior is very different, as seen in the changed tests. Also sooner rather than later we start to utilize things like `string` buffer with LuaJIT that is only available in latest.

There is nothing in particular that even lower versions couldn't be supported, it is just that nobody these days can just use plain OpenResty, you always need to include our patches etc. So why would anyone go even the hassle and build it with old OpenResty if they need to build it.

PR #8851, opened 2022-05-24
chore(clustering)  remove wRPC flaky flag

Suika-Kong:
wRPC should be stable now. Removing the flaky flag from the test.


chronolaw:
How do you get this conclusion? Could you list the related PR or works?

PR #8864, opened 2022-05-26
refactor(wrpc) Refactor wRPC proto file process

Suika-Kong:
The origin implementation of proto file searching is hacking with `searchpath`, and that implementation would fail to deal with `include` in a proto file that should be found outside the containing folder of that file.

What this PR does:
1. Extract proto related code into a separate module, and styling;
2. Rewrite implementation of annotation parsing;
3. Rewrite implementation of proto file searching.

### DIFF
--- a/kong/clustering/wrpc_control_plane.lua
+++ b/kong/clustering/wrpc_control_plane.lua
@@ -164,7 +164,7 @@ function _M:export_deflated_reconfigure_payload()
 
   -- store serialized plugins map for troubleshooting purposes
   local shm_key_name = "clustering:cp_plugins_configured:worker_" .. ngx.worker.id()
-  kong_dict:set(shm_key_name, cjson_encode(self.plugins_configured));
+  kong_dict:set(shm_key_name, cjson_encode(self.plugins_configured))
 
   local service = get_config_service(self)
   self.config_call_rpc, self.config_call_args = assert(service:encode_args("ConfigService.SyncConfig", {
--- a/kong/clustering/wrpc_control_plane.lua
+++ b/kong/clustering/wrpc_control_plane.lua
@@ -61,7 +61,7 @@ end
 local function get_config_service(self)
   if not wrpc_config_service then
     wrpc_config_service = wrpc_proto.new()
-    wrpc_config_service:import_proto("kong.services.config.v1.config")
+    wrpc_config_service:import("kong.services.config.v1.config")
 
     wrpc_config_service:set_handler("ConfigService.PingCP", function(peer, data)
       local client = self.clients[peer.conn]
--- a/kong/clustering/wrpc_control_plane.lua
+++ b/kong/clustering/wrpc_control_plane.lua
@@ -11,6 +11,7 @@ local declarative = require("kong.db.declarative")
 local constants = require("kong.constants")
 local openssl_x509 = require("resty.openssl.x509")
 local wrpc = require("kong.tools.wrpc")
+local wrpc_proto = require("kong.tools.wrpc.proto")
 local string = string
 local setmetatable = setmetatable
 local type = type
@@ -59,8 +60,8 @@ end
 
 local function get_config_service(self)
   if not wrpc_config_service then
-    wrpc_config_service = wrpc.new_service()
-    wrpc_config_service:add("kong.services.config.v1.config")
+    wrpc_config_service = wrpc_proto.new()
+    wrpc_config_service:import_proto("kong.services.config.v1.config")
 
     wrpc_config_service:set_handler("ConfigService.PingCP", function(peer, data)
       local client = self.clients[peer.conn]
--- a/kong/plugins/grpc-gateway/deco.lua
+++ b/kong/plugins/grpc-gateway/deco.lua
@@ -87,7 +87,8 @@ local function get_proto_info(fname)
 
   info = {}
 
-  grpc_tools.new():each_method(fname, function(parsed, srvc, mthd)
+  local grpc_tools_instance = grpc_tools.new()
+  grpc_tools_instance:each_method(fname, function(parsed, srvc, mthd)
     local options_bindings =  {
       safe_access(mthd, "options", "options", "google.api.http"),
       safe_access(mthd, "options", "options", "google.api.http", "additional_bindings")
--- a/kong/plugins/grpc-gateway/deco.lua
+++ b/kong/plugins/grpc-gateway/deco.lua
@@ -87,7 +87,7 @@ local function get_proto_info(fname)
 
   info = {}
 
-  grpc_tools.each_method(fname, function(parsed, srvc, mthd)
+  grpc_tools.new():each_method(fname, function(parsed, srvc, mthd)
     local options_bindings =  {
       safe_access(mthd, "options", "options", "google.api.http"),
       safe_access(mthd, "options", "options", "google.api.http", "additional_bindings")
--- a/kong/tools/grpc.lua
+++ b/kong/tools/grpc.lua
@@ -66,32 +66,33 @@ local function set_hooks()
 end
 
 function _M.new()
-  local p = protoc.new()
-  p:addpath("/usr/include")
-  p:addpath("/usr/local/opt/protobuf/include/")
-  p:addpath("/usr/local/kong/lib/")
-  p:addpath("kong")
-  p:addpath("kong/include")
-  p:addpath("spec/fixtures/grpc")
-  p.include_imports = true
+  local protoc_instance = protoc.new()
+  protoc_instance:addpath("/usr/include")
+  protoc_instance:addpath("/usr/local/opt/protobuf/include/")
+  protoc_instance:addpath("/usr/local/kong/lib/")
+  protoc_instance:addpath("kong")
+  protoc_instance:addpath("kong/include")
+  protoc_instance:addpath("spec/fixtures/grpc")
+  protoc_instance.include_imports = true
 
   return setmetatable({
-    p = p,
+    protoc_instance = protoc_instance,
   }, _M)
 end
 
-function _M:addpath(p)
-  if type(p) == "table" then
-    for _, v in ipairs(p) do
-      self.p:addpath(v)
+function _M:addpath(path)
+  local protoc_instance = self.protoc_instance
+  if type(path) == "table" then
+    for _, v in ipairs(path) do
+      protoc_instance:addpath(v)
     end
   else
-    self.p:addpath(p)
+    protoc_instance:addpath(path)
   end
 end
 
 function _M:name_search(name)
-  for _, path in ipairs(self.p.paths) do
+  for _, path in ipairs(self.protoc_instance.paths) do
     local fn = path ~= "" and path .. "/" .. name or name
     local fh, _ = io.open(fn)
     if fh then
@@ -101,22 +102,14 @@ function _M:name_search(name)
   return nil
 end
 
---- loads a .proto file optionally applies a function on each defined method.
-function _M:each_method(fname, f, recurse)
-  local p = self.p
-  local dir = splitpath(abspath(fname))
-  p:addpath(dir)
-  p:loadfile(fname)
-  set_hooks()
-  local parsed = p:parsefile(fname)
-
+local function each_method_recur(protoc_instance, fname, f, recurse)
+  local parsed = protoc_instance:parsefile(fname)
   if f then
-
     if recurse and parsed.dependency then
       if parsed.public_dependency then
         for _, dependency_index in ipairs(parsed.public_dependency) do
           local sub = parsed.dependency[dependency_index + 1]
-          _M:each_method(sub, f, true)
+          each_method_recur(protoc_instance, sub, f, true)
         end
       end
     end
@@ -129,6 +122,17 @@ function _M:each_method(fname, f, recurse)
   end
 
   return parsed
+end
+
+--- loads a .proto file optionally applies a function on each defined method.
+function _M:each_method(fname, f, recurse)
+  local protoc_instance = self.protoc_instance
+  local dir = splitpath(abspath(fname))
+  protoc_instance:addpath(dir)
+  protoc_instance:loadfile(fname)
+  set_hooks()
+
+  return each_method_recur(protoc_instance, fname, f, recurse)
 end
 
 --- wraps a binary payload into a grpc stream frame.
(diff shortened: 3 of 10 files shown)


chronolaw:
Please add some comments, or else others can not understand what is it.

Suika-Kong:
Also found a bug during refactory.

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,154 @@
...
+local grpc_new = grpc.new
+local pb_encode = pb.encode
+
+local _M = {}
+_M.__index = _M
+setmetatable(_M, _M)
+
+local wrpc_proto_name = "wrpc.wrpc"
+
+local default_proto_path = { "kong/include/", "/usr/include/" }
```
chronolaw, line 13: Should has a `,` at the end of each item.

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,154 @@
+local pb = require "pb"
+local grpc = require "kong.tools.grpc"
+
+local grpc_new = grpc.new
+local pb_encode = pb.encode
+
+local _M = {}
+_M.__index = _M
```
chronolaw, line 8: `local _MT = { __index = _M, }`

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,154 @@
...
+setmetatable(_M, _M)
+
+local wrpc_proto_name = "wrpc.wrpc"
+
+local default_proto_path = { "kong/include/", "/usr/include/" }
+
+local function parse_annotation(annotation)
+    local ret = {}
```
chronolaw, line 16: What does `ret` means?  we should rename it.

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,154 @@
...
+-- use key service-id and rpc-id to get IDs for service and rpc
+function _M:parse_annotations(proto_f)
+    local svc_ids = self.svc_ids
+    local rpc_ids = self.rpc_ids
+    local annotations = self.annotations
+
+    local service = ""
+    for line in proto_f:lines() do
+        local annotation = line:match("//%s*%+wrpc:%s*(.-)%s*$")
+        if annotation then
```
chronolaw, line 36: we can use `if not annotation` to do a shortcut

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,154 @@
...
+                        keyword .. "with no id assigned")
+                ids[name] =
+                    assert(tonumber(id), keyword .. "'s id should be a number")
+            end
+        end
+    end
+end
+
+function _M.new()
+    local ret = setmetatable({
```
chronolaw, line 66: rename `ret`

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,154 @@
...
+end
+
+--- sets a service handler for the givern rpc method
+--- @param rpc_name string Full name of the rpc method
+--- @param handler function Function called to handle the rpc method.
+--- @param response_handler function Fallback function called to handle responses.
+function _M:set_handler(rpc_name, handler, response_handler)
+  local rpc = self:get_rpc(rpc_name)
+  if not rpc then
+    return nil, string.format("unknown method %q", rpc_name)
```
chronolaw, line 126: localize `string.format`

kong/tools/wrpc/proto.lua
Suika-Kong: > we can use `if not annotation` to do a shortcut

We don't have a `break` for Lua and therefore we have to use a goto, which is what I want to aviod.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -59,8 +60,8 @@ end
 
 local function get_config_service(self)
   if not wrpc_config_service then
-    wrpc_config_service = wrpc.new_service()
-    wrpc_config_service:add("kong.services.config.v1.config")
+    wrpc_config_service = wrpc_proto.new()
+    wrpc_config_service:import_proto("kong.services.config.v1.config")
```
chronolaw, line 64: I think `import` is enough.

kong/tools/wrpc/proto.lua
chronolaw: No, in lua `goto continue` is a popular pattern, it is human friendly.

kong/tools/grpc.lua
```diff
@@ -47,42 +53,63 @@ local function set_hooks()
       return date(t.seconds):fmt("${iso}")
     end,
     function (t)
-      if type(t) ~= "string" then
-        error (string_format("expected time string, got (%s)%q", type(t), tostring(t)))
-      end
+    if type(t) ~= "string" then
```
chronolaw, line 56: Is it  a wrong indent?

kong/tools/grpc.lua
```diff
@@ -47,42 +53,63 @@ local function set_hooks()
...
-  p:addpath("spec/fixtures/grpc")
-
-  p.include_imports = true
-  p:addpath(dir)
-  p:loadfile(fname)
-  set_hooks()
-  local parsed = p:parsefile(fname)
+function _M.new()
+  local protoc_instance = protoc.new()
+  protoc_instance:addpath("/usr/include")
```
chronolaw, line 69: we could use an array to store these paths, then add them in a loop

kong/tools/grpc.lua
```diff
@@ -47,42 +53,63 @@ local function set_hooks()
...
+  if type(path) == "table" then
+    for _, v in ipairs(path) do
+      protoc_instance:addpath(v)
+    end
+  else
+    protoc_instance:addpath(path)
+  end
+end
+
+function _M:name_search(name)
```
chronolaw, line 93: `name_search` looks a little strange, could we rename it to `search_by_name` or `search`?

kong/tools/grpc.lua
```diff
@@ -47,42 +53,63 @@ local function set_hooks()
...
+  for _, path in ipairs(self.protoc_instance.paths) do
+    local fn = path ~= "" and path .. "/" .. name or name
+    local fh, _ = io.open(fn)
```
chronolaw, line 96: I think `_` is unnecessary.

chronolaw: and should we close `fh`?

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -87,7 +87,7 @@ local function get_proto_info(fname)
 
   info = {}
 
-  grpc_tools.each_method(fname, function(parsed, srvc, mthd)
+  grpc_tools.new():each_method(fname, function(parsed, srvc, mthd)
```
chronolaw, line 90: loca it to cache.

kong/tools/grpc.lua
```diff
@@ -47,42 +52,63 @@ local function set_hooks()
...
+end
 
-  if f then
+function _M:addpath(path)
+  local protoc_instance = self.protoc_instance
+  if type(path) == "table" then
+    for _, v in ipairs(path) do
+      protoc_instance:addpath(v)
+    end
+  else
```
chronolaw, line 88: add a blank line

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,155 @@
...
+    local fname = name:gsub('%.', '/') .. '.proto'
+
+    local fh = assert(self.grpc_instance:name_search(fname),
+        "module " .. name .. " cannot be found or cannot be opened")
+    self:parse_annotations(fh)
+    fh:close()
+
+    local svc_ids = self.svc_ids
+    local rpc_ids = self.rpc_ids
+    -- throwable
```
chronolaw, line 101: then what shall we do?

kong/tools/grpc.lua
Suika-Kong: `_` means we intentionally ignore error here.
no. `fh` is a return value.

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,164 @@
...
+    local keyword, identifier = nextline:match("^%s*(%a+)%s+(%w+)")
+    if not keyword or not identifier then
+      goto continue
+    end
+
+    local name, id_tag_name, ids
+    if keyword == "service" then
+      name = identifier
+      id_tag_name = "service-id"
+      service = identifier;
```
chronolaw, line 52: `;` should not be here.

Suika-Kong: No idea where it comes from but fixed.

chronolaw:
Could we add some test cases in `spec/01-unit`?

Suika-Kong:
> Could we add some test cases in `spec/01-unit`?

We can do it in a separate PR.

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,164 @@
...
+
+local wrpc_proto_name = "wrpc.wrpc"
+
+local default_proto_path = { "kong/include/", "/usr/include/", }
+
+local function parse_annotation(annotation)
+  local parsed = {}
+  for kv_pair in annotation:gmatch("[^;]+=[^;]+") do
+    local key, value = kv_pair:match("^%s*(%S-)=(%S+)%s*$")
+    parsed[key] = value
```
javierguerragiraldez, line 20: this would give a hard error if there's no match

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,164 @@
...
+local _MT = { __index = _M, }
+
+local wrpc_proto_name = "wrpc.wrpc"
+
+local default_proto_path = { "kong/include/", "/usr/include/", }
+
+local function parse_annotation(annotation)
+  local parsed = {}
+  for kv_pair in annotation:gmatch("[^;]+=[^;]+") do
+    local key, value = kv_pair:match("^%s*(%S-)=(%S+)%s*$")
```
kikito, line 19: This expression allows for spaces around annotations but not around the equals sign. Maybe include this?

```suggestion
    local key, value = kv_pair:match("^%s*(%S-)%s*=%s*(%S+)%s*$")
```

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,164 @@
...
+  end
+
+  return parsed
+end
+
+---@TODO: better way to do this
+-- Parse annotations in proto files with format:
+-- +wrpc: key1=val1; key2=val2; ...
+-- use key service-id and rpc-id to get IDs for service and rpc
+function _M:parse_annotations(proto_f)
```
kikito, line 30: Unless we want to call parse_annotations from the exterior, I would not make this a public method, but instead a private function.

It is probably possible to make it return the annotations instead of modifying self.annotations:

```
self.annotations = parse_annotations(proto_f)
```

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,164 @@
+local pb = require "pb"
```
kikito, line 1: It seems there are no tests for these features

kong/tools/wrpc/proto.lua
Suika-Kong: Seems like an incorrect manually solved merge...

Suika-Kong: You are right. if `kv_pair` happens to be `" = "` the `%S+` fails to match.
I'm not sure why we match it this way(`(%S-)` for key and `(%S+)` for value). Is it ok for the key to be `""`?

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,164 @@
...
+  self.grpc_instance:each_method(fname,
+    function(_, srvc, mthd)
+      local svc_id = assert(svc_ids[srvc.name], "service " .. srvc.name .. " has no id assigned")
+      local rpc_id = assert(rpc_ids[srvc.name .. '.' .. mthd.name], "rpc " .. mthd.name .. " has no id assigned")
+
+      mthd.svc_id = svc_id
+      mthd.rpc_id = rpc_id
+
+      self.name_to_mthd[srvc.name .. "." .. mthd.name] = mthd
+      self.name_to_mthd[svc_id .. "." .. rpc_id] = mthd
```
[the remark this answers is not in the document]
Suika-Kong: There are 2 scenarios, for API to get the RPC from its name, and for implementation to find RPC by their ID. So there will not be cross-combination.

kong/tools/wrpc/proto.lua
Suika-Kong: It also saves some status to the proto object. But yes, this should not be exposed.

kong/tools/wrpc/proto.lua
```diff
@@ -0,0 +1,174 @@
...
+
+local grpc_new = grpc.new
+local pb_encode = pb.encode
+local setmetatable = setmetatable
+local string_format = string.format
+
+local _M = {}
+local _MT = { __index = _M, }
+
+local wrpc_proto_name = "wrpc.wrpc"
```
chronolaw, line 12: I think it is a constant, so it should be `UPPERCASE`.

kong/tools/wrpc/proto.lua
Suika-Kong: Is this a comment for this line or for the whole wRPC feature? I think I'd add tests for wRPC later in another PR, to keep PR size not that large.

kong/tools/wrpc/proto.lua
Suika-Kong: What about `default_proto_path`?

chronolaw: I am not sure about it ,maybe `WRPC_PROTO_NAME`

fffonion (approved):
Discussed with @chronolaw offline, we will address style and tests in following PRs to unblock Xumin.

kong/tools/wrpc/proto.lua
javierguerragiraldez: `(%S+)` would swallow the `=` and then do a double backtrack loop from the end (one for the `(%S+)` and another for the `%s*` after it) to find it again, resulting in a O(n^2) behaviour instead of 2*O(n).  another way to do it would be `^%s*([^%s=]+)=(%S+)`; better but that creates a match table on each invocation.

PR #8869, opened 2022-05-27
feat(balancer) set the event module in the hc

locao:
this is wip. if everything goes well, we need to release lua-resty-healthcheck with support to setting the event module.


chronolaw:
I think the code is OK, but we should solve build dependency.

hbagdi:

PR #8878, opened 2022-05-30
fix(conf_loader) do not log `lua_ssl_trusted_certificate` warnings in…

dndx:
… CLI to

de-clutter `stdout`


hutchic (approved):
Looking at the CI logs the error message is no longer displayed when we don't want it

kikito (approved):

PR #8880, opened 2022-05-30
refactor(clustering) dpcp config clean

chronolaw:
### Summary
DP/CP's code is too complex and not human readable, it is hard to understand the callback workflows.

This PR separates the logic of update_config, moves the code into an isolated file,
and also de-coupled dp/cp modules.

### Full changelog
* separates the logic of `update_config` into `config_helper.lua`
* do not use `self` to mimic oop
* remove duplicated functions `extract_major_minor `and `validate_shared_cert` in wrpc_cp
* add many functions in `utils.lua`
* remove `version_handshake`
* update config hash test case


javierguerragiraldez (approved):

kong/clustering/init.lua
```diff
@@ -138,205 +39,88 @@ function _M.new(conf)
...
-  if not res then
-    return nil, err
-  end
-
-  if res.status == 404 then
-    return "v0"
-  end
-
-  return "v1"   -- wrpc
+function _M:serve_version_handshake()
```
dndx, line 61: Is this still needed?

kong/clustering/init.lua
```diff
@@ -138,205 +39,88 @@ function _M.new(conf)
...
+  return version_negotiation.serve_version_handshake(self.conf, self.cert_digest)
 end
 
+function _M:init_cp_worker(plugins_list)
+  self.json_handler.plugins_list = plugins_list
+  self.wrpc_handler.plugins_list = plugins_list
```
dndx, line 67: Why not make `plugins_list` an argument to `init_wokker()` so it is less intrusive.

kong/clustering/init.lua
```diff
@@ -138,205 +39,88 @@ function _M.new(conf)
...
+      self.child =
+        require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
+    end
 
-  local res
-  res, err = declarative.load_into_cache_with_events(entities, meta, new_hash, hashes)
-  if not res then
-    return nil, err
+    if self.child then
+      self.child.plugins_list = plugins_list
```
dndx, line 97: Same here.

kong/clustering/init.lua
```diff
@@ -138,205 +39,88 @@ function _M.new(conf)
...
+function _M:init_dp_worker(plugins_list)
+  local start_dp = function(premature)
+    if premature then
+      return
+    end
 
-  if hashes then
-    fill_empty_hashes(hashes)
-  end
+    local config_proto, msg = check_protocol_support(self.conf, self.cert, self.cert_key)
```
dndx, line 79: I think `check_protocol_support` should happen before each reconnect to be more resilient. Caching the result of it in `init_worker` seems to be a bit fragile.

kong/clustering/config_helper.lua
```diff
@@ -0,0 +1,232 @@
...
+local isarray = require("table.isarray")
+local nkeys = require("table.nkeys")
+local new_tab = require("table.new")
+
+local ngx_log = ngx.log
+local ngx_null = ngx.null
+local ngx_md5 = ngx.md5
+local ngx_md5_bin = ngx.md5_bin
+
+--local ngx_ERR = ngx.ERR
```
dndx, line 23: We should remove these comments that are not needed.

kong/clustering/init.lua
chronolaw: OK, Let's remove it.

kong/clustering/config_helper.lua
chronolaw: done

kong/clustering/init.lua
chronolaw: done

kong/clustering/init.lua
chronolaw: done

kong/clustering/init.lua
chronolaw: I agree. but now the dp's workflow seems hard to go back to this point. Do we have a solution for it?

kong/clustering/config_helper.lua
```diff
@@ -0,0 +1,230 @@
...
+      local i = 0
+      local o = new_tab(count < 100 and count or 100, 0)
+      for j = 1, count do
+        i = i + 1
+        o[i] = to_sorted_string(value[j])
+
+        if j % 100 == 0 then
+          i = 1
+          o[i] = ngx_md5_bin(concat(o, ";", 1, 100))
+        end
```
javierguerragiraldez, line 92: are there plans to fix this off-by-one error?

on every `j==n*100`, replaces all values in `o` with a hash of the first 100 and sets `i` to 1, so the next value (number 101) will go on `o[2]`.  So when `j==200`, there will be 101 values in `o` (the last hash plus 100 values more) but it will again use only 100 values for the next hash, so it drops the hashes values of items number 201, 301, 401.... etc.

kong-2.8.0-0.rockspec
```diff
@@ -65,15 +65,13 @@ build = {
     ["kong.conf_loader.listeners"] = "kong/conf_loader/listeners.lua",
 
     ["kong.clustering"] = "kong/clustering/init.lua",
-    ["kong.clustering.version_negotiation"] = "kong/clustering/version_negotiation/init.lua",
-    ["kong.clustering.version_negotiation.services_known"] = "kong/clustering/version_negotiation/services_known.lua",
-    ["kong.clustering.version_negotiation.services_requested"] = "kong/clustering/version_negotiation/services_requested.lua",
```
javierguerragiraldez, line 70: when is this coming back?

kong/clustering/config_helper.lua
chronolaw: maybe we can use `local i = 1` and `new_tab(count < 100 and count +1 or 101, 0)`

javierguerragiraldez: > Shall we fix it?

I don't know.  mentioned it just that it can be discussed.  probably in a separate PR.

chronolaw: I think perhaps we can not change it because of the compatibility of old version.

javierguerragiraldez: the hash method has changed several times already.  there's no compatibility issue

kikito: Let's fix the off-by-one error on this PR then then. Another PR will make everything take even longer.

kong/clustering/config_helper.lua
```diff
@@ -0,0 +1,230 @@
+local constants = require("kong.constants")
```
kikito, line 1: It seems this file is only doing the hash function. Consider renaming to `kong/clustering/config_hash.lua` and making it return the function directly.

kong/clustering/init.lua
```diff
@@ -1,122 +1,22 @@
...
-
-      return #value > 512 and ngx_md5_bin(value) or value
-    end
-
-  else
-    error("invalid type to be sorted (JSON types are supported)")
-  end
-end
+local check_protocol_support =
+  require("kong.clustering.utils").check_protocol_support
```
kikito, line 19: Unless necessary, please put the `require`s at the top of the file. Seeing all the file dependencies in a single place helps when reviewing.

kong/clustering/init.lua
kikito: Can this be delayed until later? I mean: We leave it on `init_worker` for now in order to try to reach feature freeze, and we change it to something more robust as a bugfix, if we end up needing the resilience.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -27,23 +26,25 @@ local DECLARATIVE_EMPTY_CONFIG_HASH = constants.DECLARATIVE_EMPTY_CONFIG_HASH
 local _M = {
   DPCP_CHANNEL_NAME = "DP-CP_config",
 }
+local _MT = { __index = _M, }
 
-function _M.new(parent)
+function _M.new(conf, cert, cert_key)
```
kikito, line 31: 👍 Big fan of replacing knowing the internals of the parent with just the params we need.

kong/templates/nginx_kong.lua
```diff
@@ -440,12 +440,6 @@ server {
             Kong.serve_wrpc_listener()
         }
     }
-
-    location = /version-handshake {
-        content_by_lua_block {
-            Kong.serve_version_handshake()
-        }
-    }
```
kikito, line 448: I am not opposed to removing this, but what is the context? Is this dead code? (It is ok to just answer and mark this as "Done").

kong/clustering/config_helper.lua
kikito: It will force us to modify all the hashes on the test, which is a pain. 

New directive: Do it only if it can be done quickly on this PR. We have lived with the problem so far.

kong-2.8.0-0.rockspec
kikito: Without these, this PR is doing more than a refactor.

kong/clustering/config_helper.lua
chronolaw: I think we can do it later, now we should finish refactoring first.

kong-2.8.0-0.rockspec
chronolaw: After this PR, we will try another method to implement similar feature.

kong/clustering/init.lua
chronolaw: done

kong/clustering/config_helper.lua
chronolaw: `utils.lua` already has too many functions, `config_helper.lua` contains all code about config, including hash and update, I think we can keep it.

kong/templates/nginx_kong.lua
chronolaw: Yes, Now we use `HEAD` request to check protocol (#8834), the old version handshake is unnecessary, so we could remove it.

You could also refer to #8775, in this PR, we use wrpc to do version negotiation and remove this endpoint.

kong/clustering/init.lua
chronolaw: Yes, I agree.

chronolaw:
This PR has conflicts with #8888, I create a new one #8893.

PR #8883, opened 2022-05-31
Refactor wRPC

Suika-Kong:
1. Added comments(usage of APIs).
2. Split modules.
3. Redesigned the API, now `call` returns a `future` object to wait for the response. And added `call_noreturn` to drop the `future`. This also makes sure the error of the call is not ignored, and a response to a call that is not initiated by this peer will be an error.
4. Styling, including localized variables.
5. Remove unused code(channel). Another feature at later PR will do the objectives of this feature.

Fix FT-2942


kong/tools/wrpc/future.lua
```diff
@@ -0,0 +1,87 @@
...
+local _M = {}
+local _MT = { __index = _M, }
+
+-- this future is no longer tracked
+-- don't call this if you don't know what this means
+function _M:finish()
+  self.response_t[self.seq] = nil
+end
+
+local function dropWait(permature, future)
```
chronolaw, line 19: We use snake_case, not camelCase, this should be `drop_wait`.

Suika-Kong:
Once this PR is merged, I will try to change the interface more aggressively.
1. Manage reconnection inside the wRPC, so config service need not take care of it;
2. Add APIs to PDK to access the wRPC connection

javierguerragiraldez (changes requested):
replacing pending IDs with futures is a good improvement.  function naming is still off.

kong/tools/wrpc/future.lua
Suika-Kong: Fixed.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -219,7 +219,7 @@ function _M:communicate(premature)
       if hash == true then
         hash = DECLARATIVE_EMPTY_CONFIG_HASH
       end
-      assert(peer:call("ConfigService.PingCP", { hash = hash }))
+      assert(peer:void_call("ConfigService.PingCP", { hash = hash }))
```
chronolaw, line 222: `void_call` is a bit strange, could we change it to a better name?

kong/tools/wrpc/future.lua
```diff
@@ -0,0 +1,87 @@
...
+local ERR = ngx.ERR
+local ngx_now = ngx.now
+local new_timer = ngx.timer.at
+local setmetatable = setmetatable
+
+local _M = {}
+local _MT = { __index = _M, }
+
+-- this future is no longer tracked
+-- don't call this if you don't know what this means
```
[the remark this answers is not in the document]
Suika-Kong: I'm trying to tell this is an internal API that should not be called by users of wRPC. Maybe I could just explicitly say it.

kong/clustering/wrpc_data_plane.lua
Suika-Kong: how about `call_no_return`?

kong/tools/wrpc/threads.lua
```diff
@@ -0,0 +1,99 @@
...
+    wrpc_peer.closing = true
+  end
+end
+
+function _M.spawn(wrpc_peer)
+  wrpc_peer._receiving_thread = assert(thread_spawn(function()
+    while not exiting() and not wrpc_peer.closing do
+      step(wrpc_peer)
+      sleep(0)
+    end
```
kikito, line 63: This looks dangerously close to an active waiting. What would be the downside of sleeping for longer instead of 0 (perhaps some configurable value)?

kong/tools/wrpc/future.lua
```diff
@@ -0,0 +1,85 @@
...
+
+--- @param delay number time until deadline
+--- @param wrpc_peer table wRPC peer that creates the call
+function _M.new(wrpc_peer, delay)
+  local new_future = setmetatable({
+    seq = wrpc_peer.seq,
+    smph = semaphore_new(),
+    delay = delay,
+    deadline = ngx_now() + delay,
+    response_t = wrpc_peer.responses,
```
chronolaw, line 77: What is the meaning of `_t` in `response_t`? in C , `_t` means a type or struct.

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,146 @@
...
+        etype = "ERROR_TYPE_INVALID_RPC",   -- invalid here, not in the definition
+        description = "Unhandled method",
+      })
+    end
+  end
+end
+
+--- Handle incoming error message (mtype == MESSAGE_TYPE_ERROR).
+function _M.handle_error(wrpc_peer, payload)
+  local etype = payload.error and payload.error.etype or "--"
```
chronolaw, line 108: Is `--`  a good desc for error? Is it defined in spec?

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,146 @@
...
+    payload.svc_id .. '.' .. payload.rpc_id)
+  if not rpc then
+    send_error(wrpc_peer, {
+      etype = "ERROR_TYPE_INVALID_SERVICE",
+      description = "Invalid service (or rpc)",
+    })
+    return nil, "INVALID_SERVICE"
+  end
+
+  local ack = tonumber(payload.ack) or 0
```
chronolaw, line 44: localize `tonumber`

kong/tools/wrpc/future.lua
```diff
@@ -0,0 +1,85 @@
...
+  if not ok then
+    return nil, err
+  end
+
+  return self.data, self.errdesc
+end
+
+--- @param delay number time until deadline
+--- @param wrpc_peer table wRPC peer that creates the call
+function _M.new(wrpc_peer, delay)
```
kikito, line 71: perhaps I am missing some context here. I am having a hard time understanding what a "future" is supposed to represent here, and I am also confused about what the methods do.

Is this similar to the concept of Promise in Node.js, or somewhere else? If this is based on something, I suggest copying the method names from there.

kong/tools/wrpc/threads.lua
```diff
@@ -0,0 +1,99 @@
...
+
+    elseif payload.mtype == "MESSAGE_TYPE_RPC" then
+      local ack = payload.ack or 0
+      local deadline = payload.deadline or 0
+
+      if ack == 0 and deadline < ngx_now() then
+        ngx_log(NOTICE, "[wRPC] Expired message (", deadline, "<", ngx_now(), ") discarded")
+
+      elseif ack ~= 0 and deadline ~= 0 then
+        ngx_log(NOTICE, "[WRPC] Invalid deadline (", deadline, ") for response")
```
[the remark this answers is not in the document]
javierguerragiraldez: i think the logic is correct and more readable, but depending on some decoding flags, those fields (`ack` and `deadline`) could arrive here with `0` as the "undefined" value, and thus not falsy.  that's because of the "not exactly static-type" semantics of protobuf.

kong/tools/wrpc/threads.lua
javierguerragiraldez: `step()` does block on I/O, so the sleep might not be that important.  perhaps only do it if there was an error?  maybe even move it (the sleep) inside `step()`?

kong/tools/wrpc/future.lua
javierguerragiraldez: i've seen promises and futures used mostly interchangeably.  i'm not sure which i'd prefer.  I thought SICP used "futures", but reviewing it it mostly calls them "expressions with lazy evaluation", and only informally mentions that they represent a "promise".

according to [wikipedia](https://en.wikipedia.org/wiki/Futures_and_promises):

> The terms future, promise, delay, and deferred are often used interchangeably, although some differences in usage between future and promise are treated below. Specifically, when usage is distinguished, a future is a read-only placeholder view of a variable, while a promise is a writable, [single assignment](https://en.wikipedia.org/wiki/Single_assignment) container which sets the value of the future.

in C++11 there are both `promise` and `future` objects, [where](https://www.cplusplus.com/reference/future/promise/):

> - The promise object is the asynchronous provider and is expected to set a value for the shared state at some point.
> - The future object is an asynchronous return object that can retrieve the value of the shared state, waiting for it to be ready, if necessary.

so, the producer constructs a `promise` and from it a `future`.  the "user" gets the `future` and the background process writes to the `promise`, making the `future` ready (an waking anybody that might've been waiting for it)

so.... i'd be (very) slightly inclined to call this a future.

kong/clustering/wrpc_data_plane.lua
chronolaw: How about `async_call`?

kong/tools/wrpc/threads.lua
```diff
@@ -0,0 +1,99 @@
...
+      step(wrpc_peer)
+      sleep(0)
+    end
+  end))
+
+  if wrpc_peer.request_queue then
+    wrpc_peer._transmit_thread = assert(thread_spawn(function()
+      while not exiting() and not wrpc_peer.closing do
+        local data, err = wrpc_peer.request_queue:pop()
+        if data then
```
chronolaw, line 70: I think we should avoid `else`, use `if not then goto continue` to reduce the nested layer.

kong/tools/wrpc/utils.lua
```diff
@@ -0,0 +1,74 @@
+local semaphore = require "ngx.semaphore"
+
+local table_insert = table.insert     -- luacheck: ignore
```
chronolaw, line 3: is `-- luacheck: ignore` necessary? Could we remove this comments?

javierguerragiraldez: here `table.insert()` is only used in the "append" form, which is compiled to exactly the same IR as `t[#t+1] = x`  so it could just be avoided.

kong/tools/wrpc/future.lua
```diff
@@ -0,0 +1,87 @@
...
+
+  local ok, err = future:wait()
+  if not ok then
+    ngx_log(ERR, "request fail to recieve response: ", err)
+  end
+end
+
+-- intentionally drop the future
+-- it will wait for the response and log error if it occurs
+function _M:drop()
```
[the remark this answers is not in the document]
Suika-Kong: Comment updated.

kong/tools/wrpc/future.lua
Suika-Kong: table.

kong/tools/wrpc/message.lua
Suika-Kong: Not sure. This is simply copied from the origin code.

kong/tools/wrpc/threads.lua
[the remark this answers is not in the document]
Suika-Kong: I will leave it unchanged but add comments to explain this.

kong/tools/wrpc/threads.lua
Suika-Kong: I will do this:

```suggestion
    while not exiting() and not wrpc_peer.closing do
      local ok = step(wrpc_peer)
      -- something wrong with this step
      -- let yield instead of retry immediately
      if not ok then
        sleep(0)
      end
    end
```

kong/tools/wrpc/threads.lua
```diff
@@ -0,0 +1,99 @@
...
+            return nil, err
+          end
+        end
+      end
+    end))
+  end
+end
+
+function _M.wait(wrpc_peer)
+  local ok, err, perr = thread_wait(safe_args(wrpc_peer._receiving_thread, wrpc_peer._transmit_thread, wrpc_peer._channel_thread))
```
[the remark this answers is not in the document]
Suika-Kong: Done

kong/tools/wrpc/threads.lua
Suika-Kong: I will just make it 2 ifs.

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,146 @@
...
+    svc_id = payload.svc_id,
+    rpc_id = payload.rpc_id,
+    ack = payload.seq,
+  })
+end
+
+--- Handle RPC data (mtype == MESSAGE_TYPE_RPC).
+--- Could be an incoming method call or the response to a previous one.
+--- @param payload table decoded payload field from incoming `wrpc.WebsocketPayload`
+function _M.process_message(wrpc_peer, payload)
```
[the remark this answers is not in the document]
Suika-Kong: It's due to the design of wRPC, and the function is doing only 1 thing. To split it makes it harder to read.

The function is about 70 lines long and can be read without scrolling down. It has a lot of empty lines and comments, and line bending. I don't think this function is too large.

Suika-Kong:
I also fixed a bug, that when an RPC returns an empty message, the payload itself will be "nil" and cause a crash.
The logic that causes the problem is nearly the same as the origin, but somehow it did not make tests fail every time.

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+  local rpc = wrpc_peer.service:get_rpc(
+    payload.svc_id .. '.' .. payload.rpc_id)
+  if not rpc then
+    send_error(wrpc_peer, {
```
fffonion, line 65: the `error` parameter seems not set in the `send_error` function. we should also check return of this function. consider doing something like this:
```lua
return send_error(wrpc_peer, payload, {
      etype = "ERROR_TYPE_INVALID_SERVICE",
      description = "Invalid service (or rpc)",
    })
```
then in send_error, do
```lua
local function send_error(wrpc_peer, payload, error)
  local ok, err = wrpc_peer:send_payload({
    mtype = "MESSAGE_TYPE_ERROR",
    error = error,
    svc_id = payload.svc_id,
    rpc_id = payload.rpc_id,
    ack = payload.seq,
  })
  if not ok then
    return ok, err
  end
  return nil, err.description
end
```

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+
+        send_error(wrpc_peer, {
+          etype = "ERROR_TYPE_UNSPECIFIED",
+          description = err,
+        })
+
+        return nil, err
+      end
+
+      wrpc_peer:send_payload({
```
fffonion, line 115: check return of this function

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+--- Handle incoming error message (mtype == MESSAGE_TYPE_ERROR).
+function _M.handle_error(wrpc_peer, payload)
+  local etype = payload.error and payload.error.etype or "--"
+  local errdesc = payload.error and payload.error.description or "--"
+  ngx_log(NOTICE, string.format(
+    "[wRPC] Received error message, %s.%s:%s (%s: %q)",
+    payload.svc_id, payload.rpc_id, payload.ack, etype, errdesc
+  ))
+
+  local ack = tonumber(payload.ack) or 0
```
fffonion, line 143: check payload.ack

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+    -- incoming method call
+    if rpc.handler then
+      local input_data = decodearray(pb_decode, rpc.input_type, payload.payloads)
+      local ok, output_data = ok_wrapper(
+        pcall(rpc.handler, wrpc_peer, table_unpack(input_data, 1, input_data.n)))
+      if not ok then
+        local err = tostring(output_data[1])
+        ngx_log(ERR, ("[wrpc] Error handling %q method: %q"):format(rpc.name, err))
+
+        send_error(wrpc_peer, {
```
fffonion, line 107: check return and fix parameter passed in

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+        svc_id = rpc.svc_id,
+        rpc_id = rpc.rpc_id,
+        ack = payload.seq,
+        payload_encoding = "ENCODING_PROTO3",
+        payloads = encodearray(pb_encode, rpc.output_type, output_data),
+      })
+
+    else
+      -- rpc has no handler
+      send_error(wrpc_peer, {
```
fffonion, line 126: check return and fix parameter passed in

Suika-Kong:
Those modules are split and copied from the origin code. I'll refactor them later. To review those code, you can ignore them as they are not changed:
1. message.lua (except for extracting `send_error`;
2. queue.lua(no change except for rename Queue to `_M` and `_MT`);
3. threads.lua(except for add comments)

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+    return nil, "INVALID_SERVICE"
+  end
+
+  local ack = tonumber(payload.ack) or 0
+  if ack > 0 then
+    -- response to a previous call
+    local response_future = wrpc_peer.responses[ack]
+    if response_future then
+
+      if response_future.deadline and response_future.deadline < ngx_now() then
```
fffonion, line 78: wrap this into a function of response_future would be a good idea.

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+      ngx_log(ERR, 
+        "reciving error response for a call not initiated by this peer.",
+        " Service ID: ", payload.svc_id, " RPC ID: ", payload.rpc_id)
+      local rpc = wrpc_peer.service:get_rpc(payload.svc_id .. payload.rpc_id)
+      if rpc and rpc.error_handler then
+        local ok, err = pcall(rpc.error_handler, wrpc_peer, etype, errdesc)
+        if not ok then
+          ngx_log(ERR, "error thrown when handling RPC error: ", err)
+        end
+      end
```
fffonion, line 169: need an error handling for else of "if rpc and rpc.error_handler", just a log for now is okay

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+    payload.svc_id .. '.' .. payload.rpc_id)
+  if not rpc then
+    send_error(wrpc_peer, {
+      etype = "ERROR_TYPE_INVALID_SERVICE",
+      description = "Invalid service (or rpc)",
+    })
+    return nil, "INVALID_SERVICE"
+  end
+
+  local ack = tonumber(payload.ack) or 0
```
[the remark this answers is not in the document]
fffonion: yes, i agree `local ack = payload.ack or 0` is sufficient here, let's remove the `tonumber` then.

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,174 @@
...
+  if ack > 0 then
+    -- response to a previous call
+    local response_future = wrpc_peer.responses[ack]
+    if response_future then
+
+      if response_future.deadline and response_future.deadline < ngx_now() then
+        if response_future.expire then
+          response_future:expire()
+        end
```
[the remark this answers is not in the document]
Suika-Kong: And yes, we can check expire time in the future itself as it is aware of the expire time( and it handles expiration when waiting).

kong/tools/wrpc/message.lua
Suika-Kong: I'm missing the payload argument.
Suggestion applied.

kong/tools/wrpc/message.lua
Suika-Kong: It misses the case when ack<0. But ack must be either number or nil.

kong/tools/wrpc/message.lua
Suika-Kong: Ditto.

kong/tools/wrpc/message.lua
Suika-Kong: done

kong/tools/wrpc/message.lua
Suika-Kong: done

kong/tools/wrpc/message.lua
Suika-Kong: I will leave this to later refactoring.

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,206 @@
...
+    local err = tostring(output_data[1])
+    ngx_log(ERR, ("[wrpc] Error handling %q method: %q"):format(rpc.name, err))
+
+    return send_error(wrpc_peer, payload, {
+      etype = "ERROR_TYPE_UNSPECIFIED",
+      description = err,
+    })
+  end
+
+  wrpc_peer:send_payload({
```
fffonion, line 84:
```suggestion
  return wrpc_peer:send_payload({
```

kong/tools/wrpc/message.lua
```diff
@@ -0,0 +1,206 @@
...
+    mtype = "MESSAGE_TYPE_RPC", -- MESSAGE_TYPE_RPC,
+    svc_id = rpc.svc_id,
+    rpc_id = rpc.rpc_id,
+    ack = payload.seq,
+    payload_encoding = "ENCODING_PROTO3",
+    payloads = encodearray(pb_encode, rpc.output_type, output_data),
+  })
+
+  return true
```
fffonion, line 93:
```suggestion
```

fffonion (commented):
minor issue

kong/tools/wrpc/proto.lua
```diff
@@ -156,19 +156,15 @@ end
 -- If calling the same method with the same args several times,
 -- (to the same or different peers), this method returns the
 -- invariant part, so it can be cached to reduce encoding overhead
-function _M:encode_args(name, ...)
+function _M:encode_args(name, arg)
```
javierguerragiraldez, line 159: this now supports a single argument, better rename to `:encode_arg`

kong/tools/wrpc/init.lua
```diff
@@ -170,382 +71,93 @@ function wrpc_peer:receive()
...
     end
   end
 end
 
---- RPC call.
---- returns the call sequence number, doesn't wait for response.
-function wrpc_peer:call(name, ...)
-  local rpc, payloads = assert(self.service:encode_args(name, ...))
-  return self:send_encoded_call(rpc, payloads)
-end
```
javierguerragiraldez, line 184: make this single-argument too (if `:encode_args()` no longer supports multiple args)

kong/tools/wrpc/init.lua
```diff
@@ -170,382 +71,93 @@ function wrpc_peer:receive()
...
-
-    elseif rpc.response_handler then
-      pcall(rpc.response_handler, self, decodearray(self.decode, rpc.output_type, payload.payloads))
-    end
-
-  else
-    -- incoming method call
-    if rpc.handler then
-      local input_data = decodearray(self.decode, rpc.input_type, payload.payloads)
-      local ok, output_data = ok_wrapper(pcall(rpc.handler, self, table_unpack(input_data, 1, input_data.n)))
```
javierguerragiraldez, line 314: this line is like this to support multiple args.  if that feature is removed, this line should be much simpler

javierguerragiraldez (changes requested):
multiple arguments was discussed and decided to be removed to simplify the code, but i still see several places where the code clearly tries to support it (and fails, because other parts now assume single arg).

basically search for every `table_unpack` and `...`  all of them should go.

kong/tools/wrpc/proto.lua
Suika-Kong: I think it's still `args`, but all arguments for 1 call.

kong/tools/wrpc/init.lua
Suika-Kong: I was planning to do this at another PR. But I think it's OK to include this.

javierguerragiraldez (approved):

fffonion (approved):

PR #8888, opened 2022-06-01
perf(*) make some heavily used code paths more cooperative by yielding

bungle:
### Summary

Adds yielding to some hot paths that otherwise may cause latency spikes with huge Kong configurations. The special focus was to make CP/DP config changes less blocking.


bungle:
Some initial tests show that this has a quite big improvement over max latency:
Screenshot from 2022-06-02 13-30-24

We do some further testing, but it looks promising.

kikito (approved):
Such a small and unobtrusive change for the benefit,  totally worth it!

PR #8890, opened 2022-06-02
refactor(core) integrate new events library

chronolaw:
### Summary

Now we have a new event libaray, let's try to introduce it to replace `lua-resty-worker-events`.

#8715 has too many changes, and it is hard to rebase to master, so I create this new PR.

See also:

https://github.com/Kong/lua-resty-events
https://github.com/Kong/kong-build-tools/pull/460

### Full changelog

* Modify `init_worker_events` in `global.lua`, switch to new events library.
* Modify `templates/nginx_kong.lua`, add unix listening
* Modify `templates/kong_defaults.lua`, add config `legacy_worker_events`
* Add a temporary `kong/resty/healthcheck.lua` to use `resty.events.compat`
* Modify test cases to pass ci, wrap them with `wait_until()`
* events library version is `0.1.0`

### Notice

* Temporary `kong/resty/healthcheck.lua` should be replaced by a formal version.
* Flaky test `02-integration/05-proxy/10-balancer/01-healthchecks_spec.lua` now is fixed.


.github/workflows/build_and_test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       with:
         repository: Kong/kong-build-tools
         path: kong-build-tools
-        ref: master
+        ref: feat/new_resty_events
```
chronolaw, line 34: Change it before merge to master.

Makefile
```diff
@@ -34,7 +34,8 @@ RESTY_VERSION ?= `grep RESTY_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk
 RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_OPENSSL_VERSION ?= `grep RESTY_OPENSSL_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_PCRE_VERSION ?= `grep RESTY_PCRE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
-KONG_BUILD_TOOLS ?= `grep KONG_BUILD_TOOLS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
+#KONG_BUILD_TOOLS ?= `grep KONG_BUILD_TOOLS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
+KONG_BUILD_TOOLS ?= feat/new_resty_events
```
chronolaw, line 38: Change it before merge to master.

kong-2.8.0-0.rockspec
```diff
@@ -483,6 +483,9 @@ build = {
     ["kong.plugins.azure-functions.handler"] = "kong/plugins/azure-functions/handler.lua",
     ["kong.plugins.azure-functions.schema"]  = "kong/plugins/azure-functions/schema.lua",
 
+    -- XXX: test only
+    ["kong.resty.healthcheck"] = "kong/resty/healthcheck.lua",
```
chronolaw, line 487: Change it before merge to master.

kong/runloop/balancer/healthcheckers.lua
```diff
@@ -20,7 +20,9 @@ local healthcheckers_M = {}
 local healthcheck_subscribers = {}
 
 function healthcheckers_M.init()
-  healthcheck = require("resty.healthcheck") -- delayed initialization
+  -- XXX: test only
+  healthcheck = require("kong.resty.healthcheck") -- delayed initialization
```
chronolaw, line 24: Change it before merge to master.

.requirements
```diff
@@ -7,6 +7,7 @@ RESTY_LUAROCKS_VERSION=3.9.0
 RESTY_OPENSSL_VERSION=1.1.1o
 RESTY_PCRE_VERSION=8.45
 RESTY_LMDB_VERSION=master
+RESTY_EVENTS_VERSION=main
```
chronolaw, line 10: Notice, events library use `main` branch, not `master`.

kong/global.lua
```diff
@@ -167,16 +167,42 @@ end
...
+      wait_max = 0.5,         -- max wait time before discarding event
+    }
+
+    worker_events = require "resty.worker.events"
+
+  else
+    local sock_name = "worker_events.sock"
+    if ngx.config.subsystem == "stream" then
+      sock_name = "stream_" .. sock_name
+    end
```
bungle, line 192: Could this be made as a `constant at `kong.constants` the top of the module?

Something like:

```lua
local EVENTS_SOCKET_NAME = ngx.config.subsystem == "stream" and "stream_worker_events.sock" or "worker_events.sock"
```

I know it is conditional, but new system is default, and we probably drop the old one?

This is not a must, just thinking. Feel free to ignore.

bungle (approved):
I have approved this with the comments. I still wish we could:
1. drop the old events library completely
2. remove the compatibility layer and use the new apis

.requirements
chronolaw: Understood.  I will create a tag before release.

bungle:
> It is a bit risky to switch to the new events library, so we have to keep the old one.

I know. It is risky to update events library, router and timer. But if on all of these we carry to old code with us, it means we have huge amount of different ways to configure Kong and each bug gets harder to resolve as you need to find out:

1. what router were you using
2. what events library were you using
3. was the timer library being used

Plus, we already have done a lot of deprecation removals. I do know that this is not the right place to decide on it, but I still would love to have the last conversation on it. Having wrpc and legacy protocol feels similar, now we need to maintain two implementations. This will definitely slow us down and create unique bugs.

This is a very big risk too. And it possibly means we need to carry the code around. For events, I would be more than willing to take that risk. If people report issues on 3.0.0 we will fix them in 3.0.1, etc. What I mean is that I don't want this to become a theme, where we reimplement something, and then don't remove the older one.

> If the new events library is proven working well, we can remove old one.

Is there any reason to believe it would not? I mean we definitely have bugs in current events library too.

chronolaw:
I understood your feeling. 

The problem is that there are many places to use events library with old apis, we will cost a lot of time to find/change/test them . I think it can not be done in this PR.

If this PR is merged maybe we can start a new PR to clean old apis.

bungle:
> The problem is that there are many places to use events library with old apis, we will cost a lot of time to find/change/test them . I think it can not be done in this PR.

Yes, we can make separate PR to remove compatibility mode, if decided so. But perhaps we need to first decide. I try to write something.

spec/02-integration/05-proxy/10-balancer/03-consistent-hashing_spec.lua
```diff
@@ -215,11 +215,19 @@ for _, strategy in helpers.each_strategy() do
...
 
           local client = helpers.proxy_client()
-          local res = assert(client:request({
-            method = "GET",
-            path = uri,
-            headers = { host = api_host },
-          }))
+
+          local res
+          helpers.wait_until(function()
```
flrgh, line 220: Seeing this pattern added in so many integration tests is concerning to me, because:

* `helpers.wait_until` is generic and doesn't really provide meaningful output when it fails, so it's harder to debug failing tests
* it makes me worry that there are many more test cases which may be flaky, and we could be chasing after them for a long time


Is there no alternative to this? Is there any function within the new events library to block until all workers have handled/seen events up until some known event? If so, we could expose it via the admin API and incorporate it into `spec.helpers`. Alternatively, we could add an optional, (maybe private) query arg to the CRUD APIs for traditional mode, so that we can do things like `POST /services?sync=true`, and it would block until all workers have caught up with events.

Additionally, something that might help would be to set `nginx_main_worker_processes` to 1 in most cases, since this will cut down on how much time is spent ensuring all workers are in sync.

chronolaw: The new events library affects too many test cases, It is a hard work to make all of them green.
I think if we could improve these test cases after this PR is merged.

Makefile
chronolaw: done

.github/workflows/build_and_test.yml
chronolaw: Now build-tools has merge that branch, so I revert it to `master`.

Makefile
chronolaw: done

kong-2.8.0-0.rockspec
chronolaw: We will remove this temporary file after feature freeze.

kong/runloop/balancer/healthcheckers.lua
chronolaw: We will remove this temporary file after feature freeze.

.requirements
```diff
@@ -7,6 +7,7 @@ RESTY_LUAROCKS_VERSION=3.9.0
...
 LIBYAML_VERSION=0.2.5
-KONG_BUILD_TOOLS_VERSION=4.28.0
+KONG_BUILD_TOOLS_VERSION=master
```
chronolaw, line 12: It should be `4.30.0` or later.

.requirements
chronolaw: The first release is `0.1.0`.

PR #8892, opened 2022-06-02
chore(cluster) Bump "reconfigure" event log level to info

lukego:
The dataplane receiving a configuration update from the controlplane is a significant event and worthy of note. This PR changes the log level from `debug` to `info`.

(It's possible that `notice` would be even more appropriate.)

### DIFF
--- a/kong/clustering/data_plane.lua
+++ b/kong/clustering/data_plane.lua
@@ -166,11 +166,11 @@ function _M:communicate(premature)
 
           if msg.type == "reconfigure" then
             if msg.timestamp then
-              ngx_log(ngx_DEBUG, _log_prefix, "received reconfigure frame from control plane with timestamp: ",
+              ngx_log(ngx_INFO, _log_prefix, "received reconfigure frame from control plane with timestamp: ",
                                  msg.timestamp, log_suffix)
 
             else
-              ngx_log(ngx_DEBUG, _log_prefix, "received reconfigure frame from control plane", log_suffix)
+              ngx_log(ngx_INFO, _log_prefix, "received reconfigure frame from control plane", log_suffix)
             end
 
             local config_table = assert(msg.config_table)


bungle:
@lukego do we need similar change in:
https://github.com/Kong/kong/blob/master/kong/clustering/wrpc_data_plane.lua

(I don't know that code too well).

kong/clustering/data_plane.lua
```diff
@@ -252,11 +252,11 @@ function _M:communicate(premature)
 
           if msg.type == "reconfigure" then
             if msg.timestamp then
-              ngx_log(ngx_DEBUG, _log_prefix, "received reconfigure frame from control plane with timestamp: ",
+              ngx_log(ngx_INFO, _log_prefix, "received reconfigure frame from control plane with timestamp: ",
```
bungle, line 255: @lukego 

ngx_INFO needs to be localized here:
https://github.com/Kong/kong/blob/46bcae0b13c9c9e4d5181e0bfcd2aa2a3c93b6a4/kong/clustering/data_plane.lua#L30

kong/clustering/data_plane.lua
```diff
@@ -252,11 +252,11 @@ function _M:communicate(premature)
...
                                  msg.timestamp, log_suffix)
 
             else
-              ngx_log(ngx_DEBUG, _log_prefix, "received reconfigure frame from control plane", log_suffix)
+              ngx_log(ngx_INFO, _log_prefix, "received reconfigure frame from control plane", log_suffix)
```
bungle, line 259: Same as above.

bungle (changes requested):

jschmid1:
It'd be really beneficial if we get the START_TIME and the END_TIME of that event

StarlightIbuki:
Since this PR is created really long time ago I'd suggest to create another PR instead of resolving conflicts.

> It'd be really beneficial if we get the START_TIME and the END_TIME of that event

Lets do it in another PR.

StarlightIbuki (approved):
-

kong/clustering/data_plane.lua
StarlightIbuki: After rebasing, the problem is fixed.

dndx:
I don't feel this PR is really that useful, and it might generate noisy logs entries too.

hbagdi:
Closing because this could be noisy. We will reconsider this in the future.

PR #8893, opened 2022-06-03
refactor(clustering) dpcp config clean

chronolaw:
### Summary
DP/CP's code is too complex and not human readable, it is hard to understand the callback workflows.

This PR separates the logic of update_config, moves the code into an isolated file,
and also de-coupled dp/cp modules.

This PR is same as #8880, but solved the merge conflict of #8888.

### Full changelog
* separates the logic of `update_config` into `config_helper.lua`
* do not use `self` to mimic oop
* remove duplicated functions `extract_major_minor `and `validate_shared_cert` in wrpc_cp
* add many functions in `utils.lua`
* remove `version_handshake`
* pick #8888's yield feature in `to_sorted_string`
* update config hash test case


fffonion (approved):
as the changeset is relatively large, my eye may miss something, so i'm relying on the tests to ensure it doesn't break. I went through the code from high level of view and didn't find obvious errors.

PR #8896, opened 2022-06-03
chore(db) remove pre-workspace blue-green deployment workaround

### LINKED ISSUE #8788: Plugins/DAO: incorrect function field reassignment

bungle:
### Summary

When Kong 2.1.0 was release we needed to add special workaround to enable blue-green deployment with the introduction of workspaces.

We needed to add special `select_by_cache_key` to `Plugins` DAO. The check that we added is hurting the Kong performance, as seen reported in #8788. We got further PR from the author with #8789. While the fix is good, I do think we can remove this by now altogether. The only downside of this is that is people migrate directly from Kong < 2.1.x to `3.0.0` we don't support the blue-green deployment.

### Issues Resolved

Fixes #8788
Closes #8789


bungle:
Alternatively, we need to merge https://github.com/Kong/kong/pull/8789.

jschmid1 (approved):

PR #8898, opened 2022-06-03
feat(cmd)! remove support for declarative config in Lua format

kikito:
For a while we accepted .lua declarative configurations via command-line only, but not on the admin API.

BREAKING CHANGE: Dropped support for loading declarative configuration in .lua format using the command-line tool

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issue reference

Fix #_[issue number]_


bungle (approved):
Soft-approved, just gets tests clean before merging.

PR #8900, opened 2022-06-03
feat(aws-lambda) add support for assume role based on metadata credentials

windmgc:
### Summary

This PR adds support for assuming role based on EC2/ECS metadata credentials.

Currently, the aws-lambda plugin does not support cross-account invoking. To achieve that we need to add support for assuming roles based on EC2/ECS metadata credentials. After fetching the metadata credentials,  it'll make an additional request to AWS's STS service to ask to assume role.

Codes in the `iam-sts-credentials.lua` originated from an [old PR](https://github.com/Kong/kong-plugin-aws-lambda/pull/42), I re-arranged it and picked up part of the function that I need.

Note that this is a short term plan to support FTI-3291 as it does not modify much of the code. Hoping to catch the last train of Version 3.0. In the long term perspective, I think all the codes related to fetching AWS credentials should be rebuilt based on https://github.com/Kong/lua-resty-aws, the library has complete support for AWS environment credential functions(ENV vars, configs, etc.)

### Full changelog

* Add implementation of [AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html).
* Add two params for lambda function plugin, `aws_assume_role_arn` and `aws_role_session_name`.

### Issue reference

Fix FTI-3291


windmgc:
need to add tests for sts function

fffonion (approved):

PR #8901, opened 2022-06-03
fix(ci): autodoc was passing when it should not have been

hutchic:
### Summary

the autodoc CI test is passing when it should not be ( https://github.com/Kong/kong/runs/6726254426?check_suite_focus=true#step:7:15 ).

This PR makes it fail "correctly" aka if the autodoc fails CI will fail as intended

I'll leave the exercise to other to make the documentation fixes required to pass this PR.

### DIFF
--- a/scripts/autodoc
+++ b/scripts/autodoc
@@ -82,7 +82,7 @@ rm -rf ./autodoc/output
 
 exit_code=$?
 
-if [ -n "${CI}" ]
+if [[ $exit_code -ne 0 ]]
 then
     exit $exit_code
 fi
--- a/scripts/autodoc
+++ b/scripts/autodoc
@@ -80,6 +80,13 @@ rm -rf ./autodoc/output
 ./autodoc/upgrading/generate.lua && \
 ./autodoc/pdk/generate.lua
 
+exit_code=$?
+
+if [ -n "${CI}" ]
+then
+    exit $exit_code
+fi
+
 if [ -z "$DOCS_REPO" ] || [ -z "$DOCS_VERSION" ]
 then
     echo


scripts/autodoc
```diff
@@ -80,6 +80,13 @@ rm -rf ./autodoc/output
 ./autodoc/upgrading/generate.lua && \
 ./autodoc/pdk/generate.lua
 
+exit_code=$?
```
flrgh, line 83: Wonder if it'd be worth it to just turn on `errexit`/`set -e` for this script and then play a couple rounds of "CI failure whack-a-mole" to catch any other lurking problems? Could also turn on `xtrace`/`set -x` for help w/ debugging in the mean time.

kikito:
I am pretty certain that this was my mistake. The autodocs script started its life as a manually-executed script, whose output would always be read by a human. I didn't think about the machines then. And later, when we added the ci check, I didn't remember that I had not thought about the machines.

Consider this a 👍 to the scripting change.

@flrgh Please add the upstreams.hash_on_query_arg docs so this passes. 

> just turn on errexit/set -e 

I think we could add it. Our CI is already kind of flaky, though. Not feeling strongly either way. If you want to give it a try, I would suggest you give `set -e` a try, run the tests 2 or 3 times, and if you get any red, rollback to Colin's. Your call.

scripts/autodoc
curiositycasualty: My personal experience with trying to add `set -e` / [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) after-the-fact is that a script has to be written from the ground up to support it. Otherwise subshells and redirects and a couple other loose ends don't carry that initial assumption of how to fail.

So I'd opt for this solution.

scripts/autodoc
```diff
@@ -80,6 +80,13 @@ rm -rf ./autodoc/output
...
+
+if [ -n "${CI}" ]
+then
+    exit $exit_code
+fi
+
```
flrgh, line 89: Suggesting to allow the rest of the script to continue in CI (on success):

```suggestion
if [[ $exit_code -ne 0 ]]
then
    exit $exit_code
fi

```

flrgh:
Added the missing docs in #8903 and tested locally with success.

flrgh (approved):

PR #8906, opened 2022-06-06
fix(session): add `cookie_persistent` to ensure is lifetime is applied

onematchfox:
### Summary

At present, the schema configuration `cookie_lifetime` on the `session` plugin does nothing. As per the docs on [lua-resty-session](https://github.com/bungle/lua-resty-session/#number-sessioncookielifetime)

> session.cookie.lifetime holds the cookie lifetime in seconds in the future. By default this is set to 3,600 seconds. This can be configured with Nginx set $session_cookie_lifetime 3600;. **This does not set cookie's expiration time on session only (by default) cookies, but it is used if the cookies are configured persistent with session.cookie.persistent == true.** See also notes about [ssl_session_timeout](https://github.com/bungle/lua-resty-session/#nginx-configuration-variables).

I.e. to ensure this value is used, `cookie.persistent` must also be set to `true`. Currently this is not possible with the `session` plugin.

### Full changelog

* Add schema option `cookie_persistent` to `session` plugin
* Add test to assert that `Expires` and `Max-Age` are set on session cookie using `cookie_lifetime` when `cookie_persistent = true`

### Issue reference

No issue has been raised.

### DIFF
--- a/kong/plugins/session/schema.lua
+++ b/kong/plugins/session/schema.lua
@@ -55,6 +55,7 @@ return {
           { cookie_httponly = { type = "boolean", default = true } },
           { cookie_secure = { type = "boolean", default = true } },
           { cookie_discard = { type = "number", default = 10 } },
+          { cookie_persistent = { type = "boolean", default = false } },
           {
             storage = {
               required = false,
--- a/kong/plugins/session/session.lua
+++ b/kong/plugins/session/session.lua
@@ -15,15 +15,16 @@ local function get_opts(conf)
     secret  = conf.secret,
     storage  = conf.storage,
     cookie  = {
-      lifetime = conf.cookie_lifetime,
-      idletime = conf.cookie_idletime,
-      path     = conf.cookie_path,
-      domain   = conf.cookie_domain,
-      samesite = conf.cookie_samesite,
-      httponly = conf.cookie_httponly,
-      secure   = conf.cookie_secure,
-      renew    = conf.cookie_renew,
-      discard  = conf.cookie_discard,
+      lifetime   = conf.cookie_lifetime,
+      idletime   = conf.cookie_idletime,
+      path       = conf.cookie_path,
+      domain     = conf.cookie_domain,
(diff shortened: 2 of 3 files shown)


onematchfox:
> ### Summary
> At present, the schema configuration `cookie_lifetime` on the `session` plugin does nothing. As per the docs on [lua-resty-session](https://github.com/bungle/lua-resty-session/#number-sessioncookielifetime)
> 
> > session.cookie.lifetime holds the cookie lifetime in seconds in the future. By default this is set to 3,600 seconds. This can be configured with Nginx set $session_cookie_lifetime 3600;. **This does not set cookie's expiration time on session only (by default) cookies, but it is used if the cookies are configured persistent with session.cookie.persistent == true.** See also notes about [ssl_session_timeout](https://github.com/bungle/lua-resty-session/#nginx-configuration-variables).
> 
> I.e. to ensure this value is used, `cookie.persistent` must also be set to `true`. Currently this is not possible with the `session` plugin.
> 
> ### Full changelog
> * Add schema option `cookie_persistent` to `session` plugin
> * Add test to assert that `Expires` and `Max-Age` are set on session cookie using `cookie_lifetime` when `cookie_persistent = true`

@bungle Thanks for the feedback. As expected you are correct - it seems that I have misread the underlying code. I'll go ahead and close this PR now. I do however have a use-case for a persistent cookie and so, I am wondering if you are open to me (re)opening this as a separate(?) `feat` PR to add/enable this configuration option?

**For reference**: 
- `cookie.lifetime` is used to calculate `session.expiry` on `save` [here](https://github.com/bungle/lua-resty-session/blob/e6bf2630c90df7b3db35e859f0aa7e096af3e918/lib/resty/session.lua#L298). 
- `expiry` is encoded into the cookie value [here](https://github.com/bungle/lua-resty-session/blob/e6bf2630c90df7b3db35e859f0aa7e096af3e918/lib/resty/session/strategies/default.lua#L187-L195)

PR #8907, opened 2022-06-06
feat(aws-lambda) separate aws credential cache in aws-lambda plugin by plugin identifier in config

windmgc:
### Summary

This PR is trying to separate AWS credential cache keys which are used by different aws-lambda plugins, by adding a plugin identifier in the mlcache key name.

Users might want to enable multiple aws-lambda plugins, targeting at different IAM roles and different lambda functions. Under this situation credential cache must be separated to avoid mutual overwrite

### Full changelog

* Adding `conf.aws_assume_role_arn` as plugin identifier to the credential cache key name

### Issue reference

No issue reference for now.

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -264,7 +264,7 @@ function AWSLambdaHandler:access(conf)
 
   if not conf.aws_key then
     -- no credentials provided, so try the IAM metadata service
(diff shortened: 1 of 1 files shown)


windmgc:
Decide to use the target IAM role ARN as the cache key identifier because the credential will only change when the target role changes. Using `conf.__key__` will make cache keys way more distributed(per service/route/consumer).

fffonion (approved):

PR #8916, opened 2022-06-07
tests(*) change kong.tools.utils.yield to check IS_CLI on call time

bungle:
### Summary

It turned out that Kong Enterprise edition requires a tracing framework that requires `kong.tools.utils` before running the `global patches`. This means that things like `ngx.IS_CLI` is not set during `require` time and then the module gets cached. This then makes some of the unit type tests not to pass on enterprise edition CI.

There are several ways to fix this. E.g. not to require `tracing` before running global patches or just doing something like this on test setup that were failing:

```
require("kong.tools.utils").yield = function() end
```

Or setting the `ngx.IS_CLI` even before running global patches...

Here is a solution that just makes detection of `IS_CLI` to happen on yield call-time.

### DIFF
--- a/kong/tools/utils.lua
+++ b/kong/tools/utils.lua
@@ -1416,26 +1416,24 @@ local topological_sort do
 end
 _M.topological_sort = topological_sort
 
-
 do
-  if ngx.IS_CLI then
-    function _M.yield() end
-  else
-    local counter = 0
-    function _M.yield(in_loop, phase)
-      phase = phase or get_phase()
-      if phase == "init" or phase == "init_worker" then
+  local counter = 0
+  function _M.yield(in_loop, phase)
+    if ngx.IS_CLI then
+      return
+    end
+    phase = phase or get_phase()
+    if phase == "init" or phase == "init_worker" then
+      return
+    end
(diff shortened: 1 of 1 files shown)


Tieske (approved):

aboudreault (approved):

yzyyi (approved):

aboudreault:
I am going to cancel and re-run the CI. not the first time I see that but sometime a worker just fail to run tests.

Screenshot_20220607_123919
t

aboudreault:
looks like  C* issue in the CI: https://github.com/Kong/kong/runs/6778008842?check_suite_focus=true#step:10:1086

aboudreault:
@mayocream I think the telemetry plugin is new, right? looks like it is causing issues in the CI

mayocream:
> @mayocream I think the telemetry plugin is new, right? looks like it is causing issues in the CI

I think one of the opentelemetry spec tests is flaky with C*, it may be related to the `helpers.http_server` is waiting for a long time, and not stoped, I'll increase the timeout of proxy_client and also add a timeout for the http_server to see if improves.

PR #8918, opened 2022-06-07
chore(*) Bundling proto files inside the luarock

tyler-ball:
### Summary

This allows us to incloud these source files as siblings to the rest of
our source code. It also allows us to look up the location of them
at runtime instead of hardcoding them into /usr/local/kong/include.

This does mean future proto files (or other includes) will need to get added to the .rockspec but I think that is better than hardcoding the protobuf lookup path.

### Full changelog

* Bundle our proto files inside the Kong luarock

### Related
* https://github.com/Kong/kong/pull/8914
* https://github.com/Kong/kong-ee/pull/3120#discussion_r854275965
* https://github.com/Kong/kong-build-tools/pull/484


kong/tools/grpc.lua
```diff
@@ -65,9 +66,12 @@ end
 
 function _M.new()
   local protoc_instance = protoc.new()
+  local kong_init_file, _, _ = loader.which("kong")
```
tyler-ball, line 69: `loader.which` returns the path to a specified module

kong-2.8.0-0.rockspec
```diff
@@ -41,6 +41,7 @@ dependencies = {
   "lua-resty-ipmatcher == 0.6.1",
   "lua-resty-acme == 0.8.0",
   "lua-resty-session == 3.10",
+  "luarocks == 3.9.0",
```
tyler-ball, line 44: Do I need to do this? Or is luarocks a default dependency and we shouldn't pin it here so we don't end up with 2 installed versions if the included version is updated later?

spec/01-unit/22-grpc-utils_spec.lua
```diff
@@ -42,4 +42,9 @@ describe("grpc tools", function()
       "Added.Final",
     }, methods)
   end)
+
+  it("includes kong/include in the load path", function()
+    local grpc_tools_instance = grpc_tools.new()
+    assert.matches("kong/include$", grpc_tools_instance.protoc_instance.paths[3])
+  end)
```
mayocream, line 49: Add a test to load the .proto file with pure luarocks loader.
```suggestion
  end)
  it("loads .proto files with luarocks loader", function()
    local loader = require "luarocks.loader"
    local kong_init_file, _, _ = loader.which("kong")
    local kong_include_dir = kong_init_file:gsub("init%.lua$", "include")
    
    local protoc = require "protoc"
    local p = protoc.new()
    p.include_imports = true
    p:addpath(kong_include_dir)
    assert(p:loadfile("opentelemetry/proto/collector/trace/v1/trace_service.proto"))
  end)
```

mayocream (changes requested):

kong/tools/grpc.lua
```diff
@@ -3,6 +3,7 @@ local protoc = require "protoc"
 local pb = require "pb"
 local pl_path = require "pl.path"
 local date = require "date"
+local loader = require("luarocks.loader")
```
kikito, line 6: One of the things we're trying to do in our internal ticket is removing Luarocks from our builds, in order to increase security. I believe that the current way we want to do it is by protecting the `luarocks` executable only, so requiring luarocks parts like this should be fine. Just raising this as a point of possible failure.

kong/tools/grpc.lua
```diff
@@ -65,9 +66,12 @@ end
 
 function _M.new()
   local protoc_instance = protoc.new()
+  local kong_init_file, _, _ = loader.which("kong")
+  local kong_include_dir = kong_init_file:gsub("init%.lua$", "include")
```
kikito, line 70: Making code too overprotective is not my cup of tea, but this worries me. Two concerns:
* The GRPC module shouldn't "know" about kong internal structure (like that our "main luarock uses an init.lua file") 
* If we for some reason decide to rename init.lua to something else (unlikely, but it might happen if we decide to change how we change modules), this will fail with an obscure message.

My first impulse would be moving this logic to a function in `kong.tools.utils.get_include_module` and also write a test for it, so it breaks the tests if we change things. However I believe the grpc module gets required before utils, so this might not be possible. I don't know where else I would put this function. Kong.global or even kong.constants, perhaps? The path shouldn't change once kong starts after all.

kong-2.8.0-0.rockspec
```diff
@@ -497,5 +498,32 @@ build = {
...
+      ["kong.include.kong.model.ca_certificate_proto"] = "kong/include/kong/model/ca_certificate.proto",
+      ["kong.include.kong.model.service_proto"] = "kong/include/kong/model/service.proto",
+      ["kong.include.kong.model.upstream_proto"] = "kong/include/kong/model/upstream.proto",
+      ["kong.include.kong.model.parameter_proto"] = "kong/include/kong/model/parameter.proto",
+      ["kong.include.kong.model.certificate_proto"] = "kong/include/kong/model/certificate.proto",
+      ["kong.include.kong.model.target_proto"] = "kong/include/kong/model/target.proto",
+      ["kong.include.kong.model.workspace_proto"] = "kong/include/kong/model/workspace.proto",
+      ["kong.include.kong.model.config_proto"] = "kong/include/kong/model/config.proto",
+      ["kong.include.kong.services.config.v1.config_proto"] = "kong/include/kong/services/config/v1/config.proto",
+      ["kong.include.wrpc.wrpc_proto"] = "kong/include/wrpc/wrpc.proto",
```
kikito, line 526: I have reviewed the luarocks docs and I found there's an option in rockspecs called `build.copy_directories`:

https://github.com/luarocks/luarocks/wiki/Rockspec-format

This seems to be the "traditional way to include non-lua files in a luarock". Has this option been considered?

hishamhm: @kikito Yes, I had a call with @tyler-ball discussing these two options, we ended up on this one. `copy_directories` has coarser-grain control of which files get installed, and would require more of a dance to derive the location out of a module's location anyway, so this choice was a more direct/explicit one.

kong-2.8.0-0.rockspec
hishamhm: No, this is not needed. It may cause trouble down the line too if versions are updated elsewhere.

kong/tools/grpc.lua
hishamhm: @kikito Yes, this should be fine without the executable. And thanks for sharing that info: FYI I am considering splitting `luarocks.loader` into a separare rock/library in LuaRocks 4.0 (meaning the loader only would become a dependency). The fact that the loader and executable have different security implications is further evidence for going ahead with this split!

hishamhm: @kikito Yes, this should be fine without the executable. And thanks for sharing that info: FYI I am considering splitting `luarocks.loader` into a separare rock/library in LuaRocks 4.0 (meaning the loader only would become a dependency). The fact that the loader and executable have different security implications is further evidence for going ahead with this split!

kong/tools/grpc.lua
tyler-ball: * The GRPC module needs to know that from the root of the installed luarock (or source code directory), there is a folder named `kong/include` that stores all the proto files to be loaded. I agree that looking for `kong/init.lua` isn't the best approach but there isn't a luarock API right now for "give me the install root of a rock", only "give me the path to a file in a luarock").
* I'm against moving tiny pieces of code that are only used once into utility modules. I think that just obfuscates the code and makes it harder to read. I did add a new comment here to (hopefully) better explain why I am calling the current API. 
* I think if we renamed `kong/init.lua` the tests I wrote would break trying to call `:gsub` on a nil table. I'm open to adding an assert with a better message like `assert(kong_init_file, "Could not find the 'init.lua' file in the root of the kong luarock")`. Thoughts?

kikito: Let's go with the assert then.
```suggestion
  local kong_include_dir = kong_init_file:gsub("init%.lua$", "include")
  assert(kong_include_dir, "Could not find the 'init.lua' file in the root of the kong luarock")
```

Tieske:
drive by comment: 

have we been been looking at https://github.com/hishamhm/datafile for this functionality?

It allows to load files included in a rockspec like this: https://github.com/hishamhm/datafile/blob/78efd1f9de2c15a00c7d7c73ea71684c0d50a2f4/test/test-datafile-scm-1.rockspec#L24-L26

hbagdi:

Issue #8531, opened 2022-03-10
graceful reload: kong reload on a machine with reasonable traffic and a go plugin leads to 404 and DOS


diefans:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

2.7.1

### Current Behavior

When just reloading kong while the host receives reasonable traffic passed through a go plugin, all responses after reload have a `status==404` rendering the service useless. 

The error log shows:
```
kong_1               | 192.168.33.1 - - [10/Mar/2022:17:34:35 +0000] "GET /auth/version HTTP/1.1" 404 396 "-" "-"
kong_1               | 2022/03/10 17:34:35 [error] 3442#0: *1453547 [kong] pb_rpc.lua:365 [bm-authx] closed, client: 192.168.33.1, server: kong, request: "GET /auth/version HTTP/1.1", host: "example.com"
kong_1               | 192.168.33.1 - - [10/Mar/2022:17:34:35 +0000] "GET /auth/version HTTP/1.1" 404 396 "-" "-"
kong_1               | 2022/03/10 17:34:35 [error] 3442#0: *1453556 [kong] pb_rpc.lua:365 [bm-authx] closed, client: 192.168.33.1, server: kong, request: "GET /auth/version HTTP/1.1", host: "example.com"
kong_1               | 192.168.33.1 - - [10/Mar/2022:17:34:35 +0000] "GET /auth/version HTTP/1.1" 404 396 "-" "-"
kong_1               | 2022/03/10 17:34:35 [error] 3442#0: *1453548 [kong] pb_rpc.lua:365 [bm-authx] closed, client: 192.168.33.1, server: kong, request: "GET /auth/version HTTP/1.1", host: "example.com"
```

### Expected Behavior

kong reloads gracefully without notice to a client.

### Steps To Reproduce

```markdown
1. involve go plugin route in a loadtest
2. reload kong e.g. `kong reload`
```

### Anything else?

We believe, this is related to how kong is managing recreating its workers and the go plugin process, e.g. binding new workers to the old plugin process and stopping it the latter.

diefans:
It seems I am able to "fix" the big 404 issue by resetting the plugin instance:
``` diff
diff --git a/kong/runloop/plugin_servers/pb_rpc.lua b/kong/runloop/plugin_servers/pb_rpc.lua
index 38f3aafe7..ae4531cdd 100644
--- a/kong/runloop/plugin_servers/pb_rpc.lua
[...]

PR #8923, opened 2022-06-08
fix(language) plugin server graceful shutdown

Suika-Kong:
When we exit or reload kong, plugin servers are quitting because they encounter errors in the socket(closing). This behavior is not proper. We close with SIGTERM when possible.

Hopefully, this can be a fix to #8531

### DIFF
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -327,4 +327,17 @@ function plugin_servers.start()
   end
 end
 
+function plugin_servers.stop()
+  if ngx.worker.id() ~= 0 then
+    kong.log.notice("only worker #0 can manage")
+    return
+  end
+
+  for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
+    if server_def.proc then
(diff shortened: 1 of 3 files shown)


kong/runloop/plugin_servers/init.lua
```diff
@@ -327,4 +327,17 @@ function plugin_servers.start()
...
 
+function plugin_servers.stop()
+  if ngx.worker.id() ~= 0 then
+    kong.log.notice("only worker #0 can manage")
+    return
+  end
+
+  for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
+    if server_def.proc then
+      server_def.proc:kill(15)
```
javierguerragiraldez, line 338: nit: use signal name instead of magic number.

Suika-Kong: I was trying to avoid introducing new dependence, that we need to require https://github.com/openresty/lua-resty-signal#signum.
However, I think a named constant would do.

javierguerragiraldez (approved):

aboudreault (approved):

PR #8926, opened 2022-06-09
feat(clustering) inline version negotiation

Suika-Kong:
Due to refactoring, #8775 has a lot of merging conflicts. So this is a retry of that.

I've also refactored this and designed it a little differently. And I added a check before the config sync process.

fix FT-3005


kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+  return unwrapped
+end
+
+local _M = {}
+
+local function field_validate(tbl, field, typ)
+  local v = tbl
+  for _, ind in ipairs(field) do
+    if type(v) ~= "table" then
+      error("field cannot be indexed with " .. ind)
```
javierguerragiraldez, line 46:
```suggestion
      error("field '" .. table_concat(field, ".", 1, i-1) .. "' cannot be indexed with " .. ind)
```

kong-2.8.0-0.rockspec
```diff
@@ -71,6 +71,8 @@ build = {
     ["kong.clustering.utils"] = "kong/clustering/utils.lua",
     ["kong.clustering.compat.removed_fields"] = "kong/clustering/compat/removed_fields.lua",
     ["kong.clustering.config_helper"] = "kong/clustering/config_helper.lua",
+    ["kong.clustering.services.negotiation"] = "kong/clustering/services/negotiation.lua",
+    ["kong.clustering.services.supported"] = "kong/clustering/services/supported.lua",
```
javierguerragiraldez, line 76: negotiation.proto should also be added somewhere in the rockspec

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+  if type(versions) ~= "table" then
+    error("invalid versions array for service " .. name)
+  end
+
+  local supported_service = supported_services[name]
+  if not supported_service then
+    return {
+      description = "unknown service.",
+    }
+  end
```
javierguerragiraldez, line 125: nit: 
```suggestion
    return { description = "unknown service." }
  end
```

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+
+  local service_response = negotiate_version(name, versions, supported_service)
+  return service_response
```
javierguerragiraldez, line 128: nit:
```suggestion
  return negotiate_version(name, versions, supported_service)
```

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+end
+
+local function log_negotiation_result(name, version)
+  local ok = version.version ~= nil
+  ngx_log(NOTICE, _log_prefix, "service ",
+    (ok and "accepted" or "rejected"),
+    ": \"", name, "\"",
+    (ok and ", version: " .. version.version or ""),
+    ", ", (ok and "description" or "reason"), ": ", version.description
+  )
```
javierguerragiraldez, line 138: this is clever, but hard to read and worse, hard to search
please do:
```suggestion
  if version.version ~= nil then
    ngx_log(NOTICE, "service accepted: \"", name, "\", version: ", version.version, ", description: ", version.description )
 
  else
    ngx_log(NOTICE, "service rejected: \"", name, "\", reason: ", version.description)
  end
```

kong/include/kong/services/negotiation/v1/negotiation.proto
```diff
@@ -0,0 +1,46 @@
...
+  string type = 2;
+  string version = 3;
+  string hostname = 4;
+}
+
+// When version is undefined, service is denied.
+message ServiceVersion {
+  string version = 1;
+  string description = 2;
+}
```
javierguerragiraldez, line 34: at least put a boolean instead of overloading `version` with two different meanings.

still strongly prefer two separate arrays, since the fields actually mean different things, and could easily evolve differently.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+    if not ok then
+      ngx_log(ERR, _log_prefix, result)
+      return { error_message = result }
+    end
+
+    return result
+  end)
+end
+
+-- TODO: use event to notify other workers
```
javierguerragiraldez, line 204: or just set in a shared resource (shdict, lmdb)

kong/clustering/services/supported.lua
```diff
@@ -0,0 +1,8 @@
+-- DP and CP shares one list.
+-- Always from newer version to older version.
```
javierguerragiraldez, line 2:
```suggestion
-- Always order from most to least preferred version.
```
(we might consider some very new version "experimental" and put it after the "stable" one)

kong/clustering/services/supported.lua
```diff
@@ -0,0 +1,8 @@
...
+
+return {
+    config_sync = {
+        {version = "v1", description = "The configuration synchronizing service. (JSON and Gzip)"},
+    },
+}
```
javierguerragiraldez, line 8: nits: spaces within brackets and newline at end of file

kong/clustering/wrpc_control_plane.lua
```diff
@@ -38,44 +38,49 @@ local ngx_CLOSE = ngx.HTTP_CLOSE
 local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
 local _log_prefix = "[wrpc-clustering] "
 
-local wrpc_config_service
-
+local init_negotiation_server = require("kong.clustering.services.negotiation").init_negotiation_server
```
javierguerragiraldez, line 41: move to top of the file

javierguerragiraldez (changes requested):
code structure seems very good, but needs tests to be sure of corner cases, especially about data transformations and priority logic.

most of my comments are minor, except for some dubious optimizations that reduce readability or cross abstraction responsibilities.  especially about merging the accepted and rejected structures, just because they have similar fields.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+local lower = string.lower
+
+local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
+local DECLARATIVE_EMPTY_CONFIG_HASH = constants.DECLARATIVE_EMPTY_CONFIG_HASH
+local KONG_VERSION
+
+-- it's so annoying that protobuf does not support map to array
+local function wrap_services(service)
+  local wrapped = {}
+  for name, version in pairs(service) do
```
chronolaw, line 22: localize `pairs`

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+
+  return unwrapped
+end
+
+local _M = {}
+
+local function field_validate(tbl, field, typ)
+  local v = tbl
+  for _, ind in ipairs(field) do
+    if type(v) ~= "table" then
```
chronolaw, line 45: localize `type` `error`

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+local function verify_request(body)
+  for field, typ in pairs{
+    [{
+      "node",
+    }] = "object",
+    [{
+      "node", "type",
+    }] = "string",
+    [{
+      "node", "version"
```
chronolaw, line 71: needs a `,`

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+    }] = "string",
+    [{
+      "services_requested"
```
chronolaw, line 74: needs a `,`

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+    log_negotiation_result(name, negotiated_version)
+  end
+
+  return services
+end
+
+
+local function register_client(cluster_data_plane_purge_delay, id, client_node)
+  local ok, err = kong.db.clustering_data_planes:upsert({ id = id, }, {
+    last_seen = ngx.time(),
```
chronolaw, line 161: localize `ngx.time`

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+  end)
+end
+
+-- TODO: use event to notify other workers
+-- Currently we assume only worker 0 cares about wRPC services
+local negotiated_service
+local function init_negotiated_service_tab()
+  if not negotiated_service then
+    negotiated_service = {}
+  else
```
chronolaw, line 210: a blank line before `else`

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+end
+
+local negotiation_request
+
+local function get_negotiation_request()
+  if not negotiation_request then
+    negotiation_request = {
+      node = {
+        type = "KONG",
+        version = kong.version,
```
chronolaw, line 226: should be localize `local KONG_VERSION = kong.version`

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+  if not response_data then
+    return nil, err
+  end
+
+  if response_data.error_message and not response_data.node then
+    return nil, response_data.error_message
+  end
+
+  init_negotiated_service_tab()
+  for name, version in pairs(response_data.services or {}) do
```
chronolaw, line 248: `{}` is unwise, could we define a constant empty table?

kong-2.8.0-0.rockspec
Suika-Kong: We don't do this in the rockspec. We will add it to packing in a build script.

kong/include/kong/services/negotiation/v1/negotiation.proto
Suika-Kong: I think it's intuitive: when there's no version(it being nil), then there's not a suitable version to use.

kong/clustering/services/negotiation.lua
Suika-Kong: The problem is `shdict` would block all workers.

kong/clustering/services/supported.lua
Suika-Kong: From the beginning, I want to keep things simple. Probably the better way for "experimental" features is to put them into a separate experimental version.

kong/clustering/services/negotiation.lua
Suika-Kong: Notice that `kong.version` may be initialized later than this file.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+local supported_services = require "kong.clustering.services.supported"
+local asked_services = require "kong.clustering.services.supported"
+
+local time = ngx.time
+local var = ngx.var
+local log = ngx.log
+local ERR = ngx.ERR
+local NOTICE = ngx.NOTICE
+local _log_prefix = "[wrpc-clustering] "
+local table_clear = require "table.clear"
```
chronolaw, line 13: `require` should be move to the top of file.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+    local field_name = table_concat(field, '.')
+    error("field \"" .. field_name .. "\" must be of type " .. typ)
+  end
+end
+
+local function verify_request(body)
+  for field, typ in pairs{
+    [{
+      "node",
+    }] = "object",
```
chronolaw, line 73: Could we change this to `"table", -- object`, then we can simplify the code above it .

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+    end
+  end
+
+  return { name = name, description = "No valid version" }
+end
+
+local function negotiate_service(name, versions)
+  name = lower(name)
+
+  if type(versions) ~= "table" then
```
chronolaw, line 119: Could we change these `if ... error() ` to `assert(...)`?

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+local function init_negotiated_service_tab()
+  if not negotiated_service then
+    negotiated_service = {}
+
+  else
+    table_clear(negotiated_service)
+  end
+end
+
+local function set_negotiated_service(name, verison)
```
chronolaw, line 215: This function is too small, I think it is not worthy.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -43,90 +50,90 @@ function _M.new(conf, cert, cert_key)
...
+    wrpc_services = wrpc_proto.new()
+    init_negotiation_client(wrpc_services)
+    init_config_service(wrpc_services)
   end
 
-  local conf = self.conf
+  return wrpc_services
+end
+
+local function communicate_body(cp)
```
chronolaw, line 101: `communicate_body` is a little strange, could we change it to `communicate_impl` or others?

kong/clustering/wrpc_data_plane.lua
```diff
@@ -43,90 +50,90 @@ function _M.new(conf, cert, cert_key)
...
+  local conf = cp.conf
 
   local log_suffix = " [" .. conf.cluster_control_plane .. "]"
-  local reconnection_delay = math.random(5, 10)
 
   local c, uri, err = clustering_utils.connect_cp(
-                        "/v1/wrpc", conf, self.cert, self.cert_key,
+                        "/v1/wrpc", conf, cp.cert, cp.cert_key,
```
chronolaw, line 107: I think this should be `dp`, not `cp`.

kong/include/kong/services/negotiation/v1/negotiation.proto
Suika-Kong: I don't have so strong opinion on whether to introduce a boolean for that, but after consideration, I still insist to use one list instead of two. The meaning of the list is obvious: the answers of CP for all services.

chronolaw:
I think maybe after this PR, we should add some test cases to cover wrpc feature.

kong/clustering/services/negotiation.lua
Suika-Kong: The reason to have this function is not to reduce the code size, but keep the abstraction.

Suika-Kong:
> I think maybe after this PR, we should add some test cases to cover wrpc feature.

Still I need to add another feature first, so we can unblock the global rate limiting.

kong/clustering/services/negotiation.lua
javierguerragiraldez: non-escaping allocations are sunk away.  in fact, the whole pairs and loop isn't even emitted, just as if it was all surrounded by an `if respose_data.services then  for..... end end`

kong/clustering/services/negotiation.lua
javierguerragiraldez: i also like `assert` a lot more for readability, but it does have the issue of precalculating the error.  when it's a constant it doesn't make a difference, but it's not rare that creating a good message involves some non-trivial processing, and then you have to change the code style... better to use `error()` every time you have to emit a message and leave `assert()` only for the "just in case" places where it only makes sense to die noisily.

kong/clustering/services/negotiation.lua
javierguerragiraldez: it's worthy to create variables and functions just to put names on things.  function calls are basically free (and invisible to the tracer)

kong/include/kong/services/negotiation/v1/negotiation.proto
javierguerragiraldez: due to differences between Lua and protobuf, there's too much ambiguity between nil and empty strings.  even booleans can be somewhat deceiving (but easier to fix).  overloading a string field will fail as soon as some other module in Lua changes the configuration of the protobuf module (yes, the configs are global).

two arrays is the most robust solution

kong/clustering/services/negotiation.lua
chronolaw: I agree. Let's decide it after this PR.

kong/include/kong/services/negotiation/v1/negotiation.proto
Suika-Kong: 2 solutions without too much thinking:
1. Since `""` is not a good version name, we treat `""` in the same way as `nil` and consider that as not supported(and it doesn't matter what configuration `pb` is in);
2. Or just a boolean to tell if service is accepted;

I think the 2 array solution is too verbose, but if it has a strong reason I'm OK with that solution.

chronolaw: I think Javier has given a good reason for this issue. and I think this needs not too much change.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+  return unwrapped
+end
+
+local _M = {}
+
+local function field_validate(tbl, field, typ)
+  local v = tbl
+  for i, ind in ipairs(field) do
+    if type(v) ~= "table" then
+      error("field '" .. table_concat(field, ".", 1, i-1) .. "' cannot be indexed with " .. ind)
```
chronolaw, line 53: Small lint `i - 1`, using blank.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,270 @@
...
+  if not ok then
+    log(ERR, _log_prefix, "unable to update clustering data plane status: ", err)
+    return error(err)
+  end
+end
+
+function _M.init_negotiation_server(service, conf)
+  service:import("kong.services.negotiation.v1.negotiation")
+  service:set_handler("NegotiationService.NegotiateServices", function(peer, nego_req)
+    local ok, result = pcall(function()
```
chronolaw, line 177: localize `pcall`

kong/clustering/wrpc_control_plane.lua
```diff
@@ -38,44 +39,48 @@ local ngx_CLOSE = ngx.HTTP_CLOSE
...
+  end)
+
+  wrpc_service:set_handler("ConfigService.ReportMetadata", function(peer, data)
+    local client = cp.clients[peer.conn]
+    if client then
+      ngx_log(ngx_INFO, _log_prefix, "received initial metadata package from client: ", client.dp_id)
+      client.basic_info = data
+      client.basic_info_semaphore:post()
+    end
+    return {
```
chronolaw, line 68: Small optimization, this `{ ok = 'done'}` can be defined a constant table.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -43,90 +50,90 @@ function _M.new(conf, cert, cert_key)
...
+        -- count is guaranteed to not exceed 1
+        peer.config_semaphore:post()
       end
-      return { accepted = true }
-    end)
-  end
-
-  return wrpc_config_service
+    end
+    return { accepted = true }
```
chronolaw, line 86: Small optimization, this `{ accepted = true }` can be defined a constant table.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -43,90 +50,90 @@ function _M.new(conf, cert, cert_key)
...
 
   do
-    local resp, err = peer:call_async("ConfigService.ReportMetadata", { plugins = self.plugins_list })
+    local ok, err = negotiate(peer)
+    if not ok then
+      error(err)
+    end
+  end
+
+  do
```
chronolaw, line 127: Now I think this `do ... end` is unnecessary.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -223,11 +230,31 @@ function _M:communicate(premature)
   elseif perr then
     ngx_log(ngx_ERR, _log_prefix, perr, log_suffix)
   end
+end
+
+local communicate_loop
+
+function communicate(cp, reconnection_delay)
+  return ngx.timer.at(reconnection_delay or 0, communicate_loop, cp)
```
chronolaw, line 238: localize `ngx.timer.at`

kong/clustering/wrpc_data_plane.lua
Suika-Kong: This is to limit the scope of variables.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,322 @@
...
+      tbl, place = accepted, accepted_n
+    else
+      rejected_n = rejected_n + 1
+      tbl, place = rejected, rejected_n
+    end
+
+    tbl[place] = {
+      name = name,
+      version = version.version,
+      message = version.description
```
chronolaw, line 193: needs a `,`

kong/include/kong/services/negotiation/v1/negotiation.proto
```diff
@@ -0,0 +1,58 @@
...
+
+message ServiceVersions {
+  repeated ServiceVersion service = 1;
```
javierguerragiraldez, line 37: the request don't have to push the description, just the versions it supports is fine.  i see the code doesn't put the description, so it wouldn't use any space, but removing it allows you to use a `repeated string`.

also, as you mentioned, protobuf doesn't allow a map to array, but an array of sub messages is much simpler both in proto definition and in the handling code:

```
message NegotiateServicesRequest {
  DPNodeDescription node = 1;
  repeated ServiceRequest services_requested = 2;
}

message ServiceRequest {
  string name = 1;
  repeated string versions = 2;
}
```

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,322 @@
...
+
+local pairs = pairs
+local ipairs = ipairs
+local type = type
+local error = error
+
+local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
+local DECLARATIVE_EMPTY_CONFIG_HASH = constants.DECLARATIVE_EMPTY_CONFIG_HASH
+
+-- it's so annoying that protobuf does not support map to array
```
javierguerragiraldez, line 29: this can be removed by not using a map, especially if you simplify the request (see comment in the .proto file)

kong/clustering/services/negotiation.lua
javierguerragiraldez: pet peeve: `type` is compiled away.  `error` is orders of magnitude more costly than the `_G` access (even more with the string/table/concat operations done here).  no real gain localizing everything in sight.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,322 @@
...
+local function negotiate_services(services_requested)
+  local services = {}
+
+  for name, versions in pairs(services_requested) do
+    if type(name) ~= "string" or type(versions) ~= "table" then
+      error("malformed service requested " .. name)
+    end
+
+    local negotiated_version = negotiate_service(name, versions)
+    services[name] = negotiated_version
```
javierguerragiraldez, line 152: this table is always traversed, never consulted.  it's better served by an array instead of a dictionary.

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,322 @@
...
+      local services_requested = unwrap_services(nego_req.services_requested)
+      verify_request(nego_req)
+
+      nego_req.node.sync_status = verify_node_compatibility(nego_req.node)
+      local services = negotiate_services(services_requested)
+      register_client(conf.cluster_data_plane_purge_delay, dp_id, nego_req.node)
+
+      local accepted, rejected = split_services(services)
+
+      local nego_result = {
```
javierguerragiraldez, line 237: missing CP node description

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,322 @@
...
+  if not response_data then
+    return nil, err
+  end
+
+  if response_data.error_message and not response_data.node then
+    return nil, response_data.error_message
+  end
+
+  init_negotiated_service_tab()
+  local serivces = merge_servcies(response_data.services_accepted, response_data.services_rejected)
```
javierguerragiraldez, line 299: typo: merge_servcies

kong/clustering/services/negotiation.lua
```diff
@@ -0,0 +1,322 @@
...
+local function init_negotiated_service_tab()
+  if not negotiated_service then
+    negotiated_service = {}
+
+  else
+    table_clear(negotiated_service)
+  end
+end
+
+local function set_negotiated_service(name, verison)
```
javierguerragiraldez, line 266: typo: verison

javierguerragiraldez (commented):
happy to see a much more robust data representation.

there's still an issue, however: while it's perfectly OK to use a `map` in protobuf, i don't see any advantage here.  I've previously mentioned that the amount of data manipulation here makes tests particularly important (because it's hard to be sure of the "shape" of the data produced at every step); now I think that changing replacing those maps with arrays would greatly simplify the code, removing most of the complex transformations.

kong/clustering/services/negotiation.lua
Suika-Kong: Actually no difference but I will add it.

javierguerragiraldez (approved):
:+1:

kong/clustering/services/supported.lua
```diff
@@ -0,0 +1,8 @@
+-- DP and CP shares one list.
+-- Always order from most to least preferred version.
+
+return {
+    config_sync = {
```
javierguerragiraldez, line 5: this service as defined in the specs is called `config`

kong/clustering/wrpc_data_plane.lua
```diff
@@ -44,83 +52,83 @@ function _M.new(conf, cert, cert_key)
...
+    end
+    local resp, err = peer:call_async("ConfigService.ReportMetadata", { plugins = dp.plugins_list })
 
     -- if resp is not nil, it must be table
     if not resp or not resp.ok then
-      ngx_log(ngx_ERR, _log_prefix, "Couldn't report basic info to CP: ", resp and resp.error or err)
-      assert(ngx.timer.at(reconnection_delay, function(premature)
-        self:communicate(premature)
-      end))
+      error("Couldn't report basic info to CP: ", resp and resp.error or err)
```
javierguerragiraldez, line 131:
```suggestion
      error("Couldn't report basic info to CP: " .. (resp and resp.error or err))
```

fffonion (approved):
let's take adding tests as next priority over other wrpc features @Suika-Kong

PR #8930, opened 2022-06-09
feat(rate-limting) allow to set error code and error message

utix:
### Summary

Error code and error message are hardcoded for rate-limiting plugin.
This allow to set them via configuration

### Issue reference
https://stackoverflow.com/questions/72557894/how-can-i-replace-kong-error-to-a-custom-text

### DIFF
--- a/spec/03-plugins/23-rate-limiting/04-access_spec.lua
+++ b/spec/03-plugins/23-rate-limiting/04-access_spec.lua
@@ -295,6 +295,28 @@ for _, strategy in helpers.each_strategy() do
               },
             })
 
+            local route6 = bp.routes:insert {
+              hosts = { "test6.com" },
+            }
+
+            bp.rate_limiting_plugins:insert({
+              route   = { id = route6.id },
+              config  = {
+                minute            = 3,
+                policy            = policy,
+                fault_tolerant    = false,
+                redis_host        = REDIS_HOST,
+                redis_port        = redis_conf.redis_port,
+                redis_ssl         = redis_conf.redis_ssl,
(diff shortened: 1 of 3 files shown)


hanshuebner (approved):

spec/03-plugins/23-rate-limiting/04-access_spec.lua
```diff
@@ -202,6 +202,8 @@ for _, strategy in helpers.each_strategy() do
                 redis_server_name = redis_conf.redis_server_name,
                 redis_password    = REDIS_PASSWORD,
                 redis_database    = REDIS_DATABASE,
+                error_code        = 404,
```
flrgh, line 205: I think it would be good to add a distinct test case for this custom status code/message feature so that things are a little bit more clear.

utix: Done, I agree this is better. The goal was to keep the PR small.

utix:
I have added a dedicated test as asked, and rebase on master

flrgh (approved):
:+1: woot, thanks @utix

hbagdi:
@flrgh Good to hit the merge button?

StarlightIbuki:
Should we add a change log for this?

utix:
> Should we add a change log for this?

I think so, people are looking for this, at least one people asked for it on stackoverflow

StarlightIbuki:
One thing this PR is missing is that we need to add the field to [removed_fields.lua](https://github.com/Kong/kong/blob/b6904ccf215d3dc22a8a853a8c10427da2e900c6/kong/clustering/compat/removed_fields.lua).

PR #8925, opened 2022-06-09
tests(prometheus) retry GET /metrics on closed conn

flrgh:
### Summary

I've had this test suite fail in CI before when the http client's keepalive connection gets closed:

```
ERROR spec/03-plugins/26-prometheus/04-status_api_spec.lua:389: Plugin: prometheus (access via status API) remove metrics from deleted upstreams
spec/03-plugins/26-prometheus/04-status_api_spec.lua:398: spec/03-plugins/26-prometheus/04-status_api_spec.lua:399: closed
stack traceback:
	./spec/helpers.lua:1369: in function 'wait_until'
	spec/03-plugins/26-prometheus/04-status_api_spec.lua:398: in function <spec/03-plugins/26-prometheus/04-status_api_spec.lua:389>
```

This change attempts to make the tests more reliable by re-opening the connection when this condition is encountered.

PR #8932, opened 2022-06-10
tests(helpers): add a reopen option to proxy http clients

flrgh:
This takes the re-open logic I used in #8925 and moves it into the helper http client so it can be reused elsewhere.

While there are a lot of tests in kong that could benefit from this behavior, there's not really a huge benefit to updating existing, non-flaky tests that simply create a new http client for each test case, so for now I haven't changed any test cases aside from the obvious prometheus one.

### DIFF
--- a/spec/02-integration/01-helpers/01-helpers_spec.lua
+++ b/spec/02-integration/01-helpers/01-helpers_spec.lua
@@ -5,6 +5,10 @@ local cjson   = require "cjson"
 for _, strategy in helpers.each_strategy() do
   describe("helpers [#" .. strategy .. "]: assertions and modifiers", function()
     local proxy_client
+    local env = {
+      database   = strategy,
+      nginx_conf = "spec/fixtures/custom_nginx.template",
+    }
 
     lazy_setup(function()
       local bp = helpers.get_db_utils(strategy, {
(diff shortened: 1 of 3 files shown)


Tieske:
This is a useful change to the http-client. But exactly because of that, I think it should be an upstream-PR.
A related one is open on the upstream repo, to not re-use a client if the body hasn't been read.

jschmid1 (approved):
lgtm!

flrgh:
Going to wait until #8933 is merged and rebase because it should make this test less flaky:

>  @github-actions Build & Test / Postgres integration - first (01-04) tests (pull_request) 

```
         FAILED  spec/02-integration/02-cmd/12-hybrid_spec.lua:141: kong hybrid with #postgres backend quits gracefully
spec/02-integration/02-cmd/12-hybrid_spec.lua:150: Expected objects to be equal.
Passed in:
(string) 'cat: /home/runner/work/kong/kong/servroot2/pids/nginx.pid: No such file or directory
'
Expected:
(string) ''

stack traceback:
	spec/02-integration/02-cmd/12-hybrid_spec.lua:150: in function <spec/02-integration/02-cmd/12-hybrid_spec.lua:141>
```

PR #8933, opened 2022-06-10
tests(helpers): add file waiter

flrgh:
The main goal I started out with was to make the test in `spec/02-integration/02-cmd/12-hybrid_spec.lua` less flaky. Waiting until a file exists and is non-empty is a common enough task that I made a helper function for it.

```lua
helpers.wait_for_file_contents(fname, timeout) end
```

I have put this function into use in a few tests.

Along the way I discovered that `helpers.wait_until` was using `ngx.time()` and therefore lacks some precision when being used with timeout of <= 1, so I swapped that out for `ngx.now()`.

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -1391,7 +1391,7 @@ local function wait_until(f, timeout, step)
   timeout = timeout or 5
   step = step or 0.05
 
-  local tstart = ngx.time()
+  local tstart = ngx.now()
   local texp = tstart + timeout
   local ok, res, err
 
@@ -1399,7 +1399,7 @@ local function wait_until(f, timeout, step)
     ok, res, err = pcall(f)
     ngx.sleep(step)
     ngx.update_time()
-  until not ok or res or ngx.time() >= texp
+  until not ok or res or ngx.now() >= texp
 
   if not ok then
     -- report error from `f`, such as assert gone wrong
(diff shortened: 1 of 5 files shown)


spec/helpers.lua
```diff
@@ -1366,15 +1366,15 @@ local function wait_until(f, timeout, step)
   timeout = timeout or 5
   step = step or 0.05
 
-  local tstart = ngx.time()
+  local tstart = ngx.now()
```
Tieske, line 1369: nice!, good catch 👍

Tieske (commented):
cannot you achieve the same with the existing `line()` assertion?

```lua
assert.logfile().has.line(".+")
```

flrgh:
> cannot you achieve the same with the existing `line()` assertion?
> 
> ```lua
> assert.logfile().has.line(".+")
> ```

Oh nice! I knew that `assert.logfile(...).has.line(...)` was a thing, but for some reason I glossed over it having a configurable timeout.

The non-logfile use case has slightly different needs. For instance, in this changeset I focused on the pidfile use case, where the following is true:

1. It's more useful to have an error/assertion failure that says "this file doesn't exist" or "this file does exist, but it's empty", since those are the much more common failure modes, and we don't care about "expected to find such-and-such string within this file"
2. Returning the full file contents to the caller is handy and saves them some boilerplate.

It's difficult to make `assert.logfile().has.line()` generic enough to handle all of this without having to rewrite it entirely, so I ended up with my own implementation.

I _could_ unify things a bit though:

1. Add `assert.file(...)` as another alias for `assert.logfile(...)` (provided that this doesn't collide with any luassert builtins)
2. Move the code from `helpers.await_file_contents` into something like `assert.file(...).contents(...)`

I admit I'm still am not totally happy with the readability in that case though. Clearly communicating intent in the luassert format can be tough. When I read `assert.file(...).contents(...)` my brain immediately thinks _"this is an assertion that a file's contents match/equal a specific string"_, **not** _"this is an assertion that the file eventually exists and has some data in it"_. The same could possibly be said about `helpers.await_file_contents`, but just by its nature as a standalone function, it's call signature is more easily discoverable via git-grep than custom assertions typically are.

@Tieske I realize that's a long-winded answer to your simple question :upside_down_face:, but what do you think?

Tieske:
sorry for the long delay in answering. Maybe the better approach is to make it more generic, and actually do an upstream PR to Luassert, do do all of this file handling. I have a long standing open issue for it anyway; https://github.com/Olivine-Labs/luassert/issues/171

flrgh:
Rebased onto master.

As to why I still feel this is necessary now that there is a separate `wait_for_file` helper, we need to validate that the pid file exists _and_ is non-empty ([here's a real world example of this](https://github.com/Kong/kong-build-tools/pull/575))

Tieske (approved):
lgtm

PR #8935, opened 2022-06-13
fix(handler) do not stop health checkers on reconfigure if not needed

bungle:
### Summary

Something went wrong with `lmdb` merge, and it seems like we now stop health checkers even when the balancer is not being re-build.

The other issue was that when `flip_config` was renamed to `reconfigure`, not everything was renamed accordingly to this name change, this fixes that too.


aboudreault (approved):
Approved when CI is green.

PR #8936, opened 2022-06-13
chore(cache) remove pages and flip

bungle:
### Summary

The `LMDB` removed the need for cache pages. This commit removes pages support from the cache.


bungle:
> nice! Approved. can be merged when tests pass.

Tests seem to be totally broken, unrelated to PRs, today, :-(.

PR #8939, opened 2022-06-13
fix(admin): return `404` on deletion of child of non-existent entity

onematchfox:
### Summary

Currently the Kong admin API will return a `500` error if a `DELETE` request is issued to the foreign _reference_ endpoint for a non-existent parent (foreign entity). 

Admin API logs will look similar to the following:

```
2022/06/13 18:37:15 [error] 7290#0: *23 [lua] api_helpers.lua:511: handle_error(): /usr/local/share/lua/5.1/lapis/application.lua:424: /kong/kong/db/schema/init.lua:2096: attempt to index local 'entity' (a nil value)
stack traceback:
	/kong/kong/db/schema/init.lua: in function 'extract_pk_values'
	/kong/kong/api/endpoints.lua:681: in function 'fn'
	/kong/kong/api/api_helpers.lua:285: in function </kong/kong/api/api_helpers.lua:268>
```

This PR ensures that the API returns a `404` instead (in line with the behaviour of the other endpoints) since the non-presence of the foreign entity indicates that the reference itself does not exist.

### Full changelog

* Add test to prove out `500` error on deletion
* Return `404` in `delete_entity_endpoint` if `entity` is not found.


mayocream (approved):
LGTM

PR #8940, opened 2022-06-13
chore(*) bump kbt to 4.26.0

curiositycasualty:
### Summary

The latest we expect to work.

A similar change has already been made for 2.5.x: https://github.com/Kong/kong/commit/c5af6f70c940efc109eb14e5e0cb5683a5116363

This facilitates some unplanned work to rectify missing ubuntu focal arm64 docker containers.


curiositycasualty:
> I think it’s a little dangerous to bump kbt versions in maintenance branches. OpenResty patches might be added and consequently affect patch releases that it should not
~ @gszr

PR #8942, opened 2022-06-13
perf(declarative) use luajit serializer for lmdb

bungle:
### Summary

Replaces `cjson` based serializer with LuaJIT `string.buffer` serializer.

### DIFF
--- a/kong/db/declarative/marshaller.lua
+++ b/kong/db/declarative/marshaller.lua
@@ -1,118 +1,33 @@
+local buffer = require "string.buffer"
+
+
+local encode = buffer.encode
+local decode = buffer.decode
+
+
 local _M = {}
 
 
-local cjson = require("cjson.safe")
-local tostring = tostring
-local tonumber = tonumber
-local type = type
-local fmt = string.format
-local sub = string.sub
-local cjson_encode = cjson.encode
-local cjson_decode = cjson.decode
-
-
-local TYPES_LOOKUP = {
-    number  = 1,
-    boolean = 2,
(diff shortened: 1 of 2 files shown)


kong/db/declarative/marshaller.lua
```diff
@@ -1,118 +1,33 @@
...
-    table   = 4,
-}
+local buffer = require "string.buffer"
 
 
-local marshallers = {
-    shm_value = function(str_value, value_type)
-        return fmt("%d:%s", value_type, str_value)
-    end,
+local encode = buffer.encode
```
flrgh, line 4: Nice! I had been meaning to try this out myself.

I wonder if it's possible to use the `dict` optimization that the docs describe here:

https://github.com/openresty/luajit2/blob/124ff8dcf4ec0c158fcf7b44efbffc9093656f6c/doc/ext_buffer.html#L463-L487

(can't seem to find a published version of this html)

Adding some field names from core entities would probably be a performance win. The only downside is the list has to be fairly static or else we risk the persisted data becoming unreadable. I guess maybe we could store the list as metadata in lmdb too (in pure json or some other format that is guaranteed to be forward-safe).

bungle: @flrgh, yes dict optimization for sure, need to analyze the data a bit to compile nice dictionary. Perhaps we can do it in different PR though.

bungle:
Seems to be green now, other than the `perf` tests which have been broken for known reasons.

fffonion (approved):

PR #8950, opened 2022-06-14
perf(balancer) use luajit string.buffer to serialize health checker data

bungle:
### Summary

Use luajit string.buffer to serialize health checker data.


chronolaw:
Hi bungle, this `kong/resty/healthcheck.lua` is a temporary solution for events library's test. It will be removed when the original health check library migrate to new events library.

So I think we should not spend time on it, or you can create PR for  health check library but not here.

chronolaw:
I think we can still merge this feature and test it even before the health library implements this.

And it is a small change, if this file is finally removed we will have no losses.

chronolaw (approved):

PR #8951, opened 2022-06-14
fix(*) remove coroutine in access phase

fffonion:
We introduce this in
https://github.com/Kong/kong/commit/744d3b32e47323301bdbe402f9bf1cb2acbf88e8
to persist the behaviour that kong.response.exit can shortcut current
processing. Though this is also documented in PDK document, our code
doesn't seem to rely on it at all.

Opening this to see what will break and open for comments; if merged
this will be a breaking change as this is user-facing, so it can only go
in 3.0 or 4.0.

### Full changelog

* remove coroutine in access phase
* remove coroutine yield in kong.response.exit

### Issue reference

Fix #_[issue number]_

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -110,7 +110,6 @@ local error            = error
 local ipairs           = ipairs
 local assert           = assert
 local tostring         = tostring
-local coroutine        = coroutine
 local get_last_failure = ngx_balancer.get_last_failure
 local set_current_peer = ngx_balancer.set_current_peer
 local set_timeouts     = ngx_balancer.set_timeouts
@@ -271,16 +270,15 @@ local function execute_access_plugins_iterator(plugins_iterator, ctx)
 
       setup_plugin_context(ctx, plugin)
 
-      local co = coroutine.create(plugin.handler.access)
-      local cok, cerr = coroutine.resume(co, plugin.handler, configuration)
-      if not cok then
+      local pok, perr = pcall(plugin.handler.access, plugin.handler, configuration)
+      if not pok then
         -- set tracing error
         if span then
-          span:record_error(cerr)
+          span:record_error(perr)
           span:set_status(2)
         end
 
-        kong.log.err(cerr)
+        kong.log.err(perr)
         ctx.delayed_response = {
           status_code = 500,
           content = { message  = "An unexpected error occurred" },
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -27,7 +27,6 @@ local lower = string.lower
 local error = error
 local pairs = pairs
 local concat = table.concat
-local coroutine = coroutine
 local normalize_header = checks.normalize_header
 local normalize_multi_header = checks.normalize_multi_header
 local validate_header = checks.validate_header
@@ -950,7 +949,7 @@ local function new(self, major_version)
         }
 
         ctx.delayed_response_callback = flush
-        coroutine.yield()
+        return
 
       else
         return send(status, body, headers)
--- a/kong/plugins/oauth2/access.lua
+++ b/kong/plugins/oauth2/access.lua
@@ -785,7 +785,7 @@ local function issue_token(conf)
 end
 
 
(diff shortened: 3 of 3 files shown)


bungle:
@fffonion are you aware of this conversation:
https://github.com/Kong/kong/pull/8747#issuecomment-1116295665

Tieske:
code lgtm, but tests are failing. And I'd like @bungle s input on this one.

fffonion:
Test is green now, it does seem we are "unconsicously" relying on this behaviour. I did a cleanup and now it's correctly exited.

mayocream (approved):

bungle:
@fffonion was it not worth it to make response.exit to return a value, so that we could get rid of ctx.delayed_response and friends?

bungle:
Couple of things I think just to get others some context:

1. `kong.response.exit` behaves differently in `access` phase than any other phase, because on access phase it needs to collect the plugins for later phases. (a note, on stream `preread` it should do same, but there was a bug in 1.19.x.x OpenResty that prevented it)
2. the difference is that on access phase it does not call ngx.exit but sets some ngx.ctx values and yields the coroutine.

This PR wants to remove the coroutine, but this means that code like this won't work anymore as expected:

```lua
local function do_something()
  if kong.request.get_header("Dog") == "Bingo" then
    response.exit(200, "was his name")
  elseif kong.request.get_header("Dog") == "Bongo" then
    kong.response.exit(200, "was his previous name")
  end
end


local plugin = {}

fuction plugin:access()
  do_something()
  kong.response.exit(500, "wrong dog")
end
```

But with the changes here, the code needs to be written like:

```lua
local function do_something()
  if kong.request.get_header("Dog") == "Bingo" then
    return kong.response.exit(200, "was his name")
  elseif kong.request.get_header("Dog") == "Bongo" then
    return kong.response.exit(200, "was his previous name")
  end
end


local plugin = {}

fuction plugin:access()
  -- calling code doesn't know if the `do_something` did `kong.response.exit` or not.
  return do_something()
  --kong.response.exit(500, "wrong dog")
end
```

So infact it needs to be written as:

```lua
local function do_something()
  if kong.request.get_header("Dog") == "Bingo" then
    kong.response.exit(200, "was his name")
    return true
  elseif kong.request.get_header("Dog") == "Bongo" then
    kong.response.exit(200, "was his previous name")
    return true
  end
end


local plugin = {}

fuction plugin:access()
  -- calling code doesn't know if the `do_something` did `kong.response.exit` or not.
  if not do_something() then
    kong.response.exit(500, "wrong dog")
  end
end
```

There are couple of ways to solve this:

1. use coroutine.yield (which we are wanting to get rid of)
2. call `error("early exit")` inside `kong.response.exit` and handle it in core code
3. return value from `kong.response.exit` and handle it in plugin code and core code


The approach 1 is currently implemented in `master`. The `error` way you just use `error` instead of `coroutine.yield` and handle it in `access` plugin iteration (e.g. use `pcall` as this PR does and then check if error message is `"early exit"` or something distinguishable. The original code would stay the same (no need to touch e.g. that OAuth2 plugin code).

The fourth is something like this:

```lua
local function do_something()
  if kong.request.get_header("Dog") == "Bingo" then
    return response.exit(200, "was his name")
  elseif kong.request.get_header("Dog") == "Bongo" then
    return kong.response.exit(200, "was his previous name")
  end
end


local plugin = {}

fuction plugin:access()
  if do_something() ~= kong.EXITED then
    return kong.response.exit(500, "wrong dog")
  end
  return kong.EXITED
end
```

I am leaning towards `pcall` and `error(...)`. Though I am sure the last may be most performant.

hbagdi:
@fffonion Is this still required?

fffonion:
Since now we have passed 3.0, we can only postpone this to 4.0. @hbagdi Do we want to keep this open or other ways to track it?

hbagdi:
@fffonion Please open a Github Issue and add the 'breaking-change' label to it. And then close this PR out. Thanks!

PR #8953, opened 2022-06-14
refactor(handler) trying to make reconfigure more atomic (WIP)

bungle:
### Summary

Originally the `reconfigure` event was `atomic` and it was not yielding. But this caused huge spikes in latencies when new configurations where applied.

We started to add yielding in different places that happen in `reconfigure` event. It may now lead to nasty things like: new router, and old plugins iterator gets used etc. (router from new configuration and plugin iterator from old configuration).

I needed to refactor the handler so that it could at least support more atomic flip of `router` and `plugins iterator`. The balancer is still problematic as it is shared global state and a new balancer cannot currently be build behind the scenes cooperatively, and then flip it atomically together with router and plugins iterator. I asked @locao to research if that could be doable. I leave that to other PR.

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -92,8 +92,8 @@ local GLOBAL_QUERY_OPTS = { workspace = ngx.null, show_ws_id = true }
 local SERVER_HEADER = meta._SERVER_TOKENS
 
 
-local STREAM_TLS_TERMINATE_SOCK = "unix:" .. ngx.config.prefix() .. "/stream_tls_terminate.sock"
-local STREAM_TLS_PASSTHROUGH_SOCK = "unix:" .. ngx.config.prefix() .. "/stream_tls_passthrough.sock"
+local STREAM_TLS_TERMINATE_SOCK = "unix:" .. kong.configuration.prefix .. "/stream_tls_terminate.sock"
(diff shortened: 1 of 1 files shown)


bungle:
(TODO: need to fix it so that it passes tests)

kong/runloop/handler.lua
```diff
@@ -70,30 +79,21 @@ local CLEAR_HEALTH_STATUS_DELAY = constants.CLEAR_HEALTH_STATUS_DELAY
...
-local stream_tls_passthrough_sock = "unix:" .. ngx.config.prefix() .. "/stream_tls_passthrough.sock"
-
--- for tests
-local _set_update_plugins_iterator
-local _set_update_router
-local _set_build_router
-local _set_router
-local _set_router_version
-local _register_balancer_events
+local STREAM_TLS_TERMINATE_SOCK = "unix:" .. ngx.config.prefix() .. "/stream_tls_terminate.sock"
```
chronolaw, line 95: We should not use `ngx.config.prefix()` now, it may not be a proper path, use `kong.configuration.prefix` instead.

See PR #9337.

catbro666: Fine, thanks for reminding.

bungle:
I will close this PR in favor of #9494.

PR #8956, opened 2022-06-15
fix(pdk) ensure that vault.init is not called multiple times

bungle:
### Summary

@ms2008 pointed on PR #8945 that we may have potential issue in Vault that init method or vault strategy (plugin) may be called multiple times. This commit makes is sure that it should not happen anymore.


jschmid1:
+1 granted tests pass

bungle:
> +1 granted tests pass

Thanks, I will make it sure they pass before merging. Thank you!

PR #8958, opened 2022-06-15
chore(db) remove schema legacy attributes

bungle:
### Summary

Removes `legacy = true` support from `schemas` and from `schema fields`.


flrgh (approved):
:+1: from a functional code change standpoint. The history of this `legacy` attribute is unknown to me, so I don't have much to say about its removal.

PR #8959, opened 2022-06-15
perf(clustering) no concurrent inflating and json decoding on dps

bungle:
### Summary

Previously it could be that `read_thread` caused concurrent blocking when inflating / json decoding incoming data. This commit moves it beyond the semaphore to make it non-concurrent.


kong/clustering/data_plane.lua
```diff
@@ -223,31 +237,12 @@ function _M:communicate(premature)
...
-            self.next_hashes = msg.hashes
-
-            if config_semaphore:count() <= 0 then
-              -- the following line always executes immediately after the `if` check
-              -- because `:count` will never yield, end result is that the semaphore
-              -- count is guaranteed to not exceed 1
-              config_semaphore:post()
-            end
+          next_data = data
+          if config_semaphore:count() <= 0 then
```
flrgh, line 241: It seems to me like maybe this should always `post()` to the semaphore, but it's probable there's something I'm missing. Consider the following sequence of events:

1. WS payload is received and stored to `next_data`
2. `config_semaphore:post()`
3. `config_thread` wakes up and begins to process `next_data`, meanwhile referencing it as `data`
4. Another WS payload is received and stored to `next_data`
5. Because `config_thread` is still processing the prior update, `config_semaphore:count()` returns 0, and no `config_semaphore:post()` is executed.

Doesn't this mean that the payload received in step 4 will be ignored? I see that there's a check to make sure that we don't overwrite it in this case (`if next_data == data then next_data = nil end`), but I don't see any "we just finished an update, but now there is a new one waiting" check.

flrgh: I guess if we call `config_semaphore:post()` every time then we still need to figure out a way to roll-up repeated reconfigure events into one:

```lua
if next_data == data then
  next_data = nil
else
  -- new data arrived while we were handling the old data
  --
  -- process it immediately instead of returning to config_semaphore:wait()
end

flrgh (approved):
:+1: This is a great efficiency improvement!

PR #8966, opened 2022-06-16
feat(pdk) add caching to process secrets function 

bungle:
### Summary

Originally process secrets was meant only to be used on CLI where the cache may not be available (e.g. when parsing `kong.conf`), but we decided later that process secrets could also be used on runtime for config secrets.

This commit implements similar caching that is currently enabled on config cache function to process secrets function.

This also contains two other commits that are just some cleanups and something  that I need for rotation implementation.


jschmid1 (approved):
lgtm

PR #8967, opened 2022-06-16
feat(pdk/db) kong.vault.try function and automatic secret rotation of postgres credentials

bungle:
### Summary

This PR contains three commits on top of #8966 (which is now merged):

1. feat(pdk) add kong.vault.try helper function for secret rotation: Adds experimental (for now) kong.vault.try function that can be used to implement automatic secret rotation.
2. feat(db) add automatic secret rotation of postgres username and password: Adds automatic secret rotation of postgres username and password.
3. feat(pdk) add rate-limiting to kong.vault.try function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -252,7 +252,7 @@
 
 #### Performance
 - Do not register unnecessary event handlers on Hybrid mode Control Plane
-nodes [#8452](https://github.com/Kong/kong/pull/8452).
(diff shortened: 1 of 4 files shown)


jschmid1 (approved):

jschmid1:
Note: we may want to change the way the `options` table is used. Instead of changing the content during runtime, we should think about using separate args that are generated from the refs lookup

PR #8971, opened 2022-06-16
feat(clustering) use privileged worker for control plane connection

bungle:
### Summary

Just testing.


bungle:
Closing this for now, but we may return to it with `3.1.0`

PR #8974, opened 2022-06-17
fix(balancer) fix upstreams reload every 10s

### LINKED ISSUE #8970: Balancer/upstreams: Kong loads upstreams table every 10s in request context when no upstream exist in database

2.8.1

### Current Behavior

Hello,
When Kong has only "standalone" _services_ entities in database, and no _upstreams_, it reloads the whole empty _upstreams_ table every 10 seconds (being the negative TTL of **balancer:upstreams** cache key) in request context, adding extra 100 ms to proxy latency.
The cause is that the _kong.runloop.balancer.upstreams:load_upstreams_dict_into_memory_ function returns a _nil_ instead of an empty table (`upstreams_dict`) if no key/value pair was inserted.
So, I modified this function in order to return `upstreams_dict` unconditionally and drop the `local found` boolean, and the reload no longer took place every 10s at runtime. Cache invalidation mechanism works well and invalidates this cached dictionary as soon as an upstream is created using Kong Admin API.
Is there any reason I didn't think about in order to not cache an empty dictionary ?
Thank you.

### Expected Behavior

Kong should not reload the whole _upstreams_ table every 10s (provided traffic is enough to trigger it with such minimal interval) as long as no entity is CReated/Updated/Deleted.

### Steps To Reproduce

```markdown
1. Provision Kong with a service and an associated route (and delete all upstreams).
2. Curl the created route every 1s, greping the X-Kong-Latency-Proxy header.
3. Every 10 request, the X-Kong-Latency-Proxy jumps from 1-2ms to 100-120ms.
```

### Anything else?

Kong running with a Cassandra database.

marc-charpentier:
The upstreams module's load_upstreams_dict_into_memory returned
non-cacheable value when upstreams table is empty, causing empty
table reload in request context after 10s negative TTL's expiration.

### Summary

See https://github.com/Kong/kong/issues/8970#issue-1273886114.

To fix this, empty table may be considered a valid value to cache.

### Full changelog

* [Fix empty upstreams table reload every 10s]

### Issue reference

Fix #8970

### DIFF
--- a/spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
+++ b/spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
@@ -32,8 +32,7 @@ for _, strategy in helpers.each_strategy() do
       assert.logfile().has.line("loading upstreams dict into memory", true, 5)
       -- turncate log
       io.open("./servroot/logs/error.log", "w"):close()
-      ngx.sleep(20)
-      assert.logfile().has.no.line("loading upstreams dict into memory", true)
+      assert.logfile().has.no.line("loading upstreams dict into memory", true, 20)
     end)
   end)
 end
--- a/spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
+++ b/spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
@@ -1,6 +1,4 @@
-local cjson   = require "cjson"
 local helpers = require "spec.helpers"
-local utils   = require "kong.tools.utils"
 
 local POLL_INTERVAL = 0.3
 
--- a/spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
+++ b/spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
@@ -1,0 +1,41 @@
+local cjson   = require "cjson"
+local helpers = require "spec.helpers"
+local utils   = require "kong.tools.utils"
+
+local POLL_INTERVAL = 0.3
+
+for _, strategy in helpers.each_strategy() do
+  describe("balancer cache [#" .. strategy .. "]", function()
+
+    lazy_setup(function()
+      -- truncate the database so we have zero upstreams
+      helpers.get_db_utils(strategy, { "upstreams", })
+
+      local db_update_propagation = strategy == "cassandra" and 0.1 or 0
+
+      assert(helpers.start_kong {
+        log_level             = "debug",
+        database              = strategy,
+        proxy_listen          = "0.0.0.0:8000, 0.0.0.0:8443 ssl",
+        admin_listen          = "0.0.0.0:8001",
+        db_update_frequency   = POLL_INTERVAL,
+        db_update_propagation = db_update_propagation,
+        nginx_conf            = "spec/fixtures/custom_nginx.template",
+      })
+    end)
+
+    lazy_teardown(function()
+      helpers.stop_kong()
+    end)
+
+
+    -- https://github.com/Kong/kong/issues/8970
+    it("upstreams won't reload at unusual rate", function()
+      assert.logfile().has.line("loading upstreams dict into memory", true, 5)
+      -- turncate log
+      io.open("./servroot/logs/error.log", "w"):close()
+      ngx.sleep(20)
+      assert.logfile().has.no.line("loading upstreams dict into memory", true)
+    end)
+  end)
+end
--- a/kong/runloop/balancer/upstreams.lua
+++ b/kong/runloop/balancer/upstreams.lua
@@ -106,15 +106,13 @@ end
 --_load_upstreams_dict_into_memory = load_upstreams_dict_into_memory
 
 
-local opts = { neg_ttl = 10 }
-
 ------------------------------------------------------------------------------
 -- Implements a simple dictionary with all upstream-ids indexed
 -- by their name.
 -- @return The upstreams dictionary (a map with upstream names as string keys
 -- and upstream entity tables as values), or nil+error
 function upstreams_M.get_all_upstreams()
-  return kong.core_cache:get("balancer:upstreams", opts,
+  return kong.core_cache:get("balancer:upstreams", nil,
                                                   load_upstreams_dict_into_memory)
 end
 
--- a/kong/runloop/balancer/upstreams.lua
+++ b/kong/runloop/balancer/upstreams.lua
@@ -82,6 +82,7 @@ end
 ------------------------------------------------------------------------------
 
 local function load_upstreams_dict_into_memory()
+  log(DEBUG, "loading upstreams dict into memory")
   local upstreams_dict = {}
 
   -- build a dictionary, indexed by the upstream name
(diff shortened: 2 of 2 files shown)


marc-charpentier:
Hello,
Sorry for last commit come and go, I was wondering if cache invalidation delay caused [this unit test](https://github.com/Kong/kong/runs/6941870691?check_suite_focus=true#step:10:1983) to fail.

marc-charpentier:
Hello,
I think I managed to run the failing unit test on my local machine, but the result is the same with both branches (_release/2.8.x_ and _fix/empty-upstreams-reload-every-10-seconds_). I also added logging instructions on error level in both modified functions in _kong/runloop/balancer/upstreams.lua_ but they didn't appear on unit test run.
I tend to think the proposed fix isn't causing the unit test failure.
Would someone more experienced than me have a look ?
Thanks.

flrgh:
Hey @marc-charpentier, sorry for the late check-in on this (we're all a little bit swamped at the moment prepping the next release).

That test failure looks very similar to one we've been tracking internally that is known to be flaky, so I think it's okay for you to disregard for now.

This changeset looks good to me (thanks again for your efforts--much appreciated!). When I have the time I'll probably do a little exploratory work to see if adding a targeted integration test is feasible, though it might not be necessary in the end. Aside from that, just awaiting a second opinion from somebody else on the team.

marc-charpentier:
Hi @flrgh , thank you for your answer, and no problem for the late check-in.
Now all tests passed, although I don't know if someone triggered them again, nor why the previously failing one succeeded this time.

StarlightIbuki:
Difficult to rebase so I'm trying to cherry-pick it. Trying to find some way to test this.

StarlightIbuki:
This could be a perf PR or a bug-fixing PR. The former one needs not specific tests but we need to show improvement and the latter one needs tests.
I tend to add tests.

spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
```diff
@@ -0,0 +1,39 @@
...
+    lazy_teardown(function()
+      helpers.stop_kong()
+    end)
+
+
+    -- https://github.com/Kong/kong/issues/8970
+    it("upstreams won't reload at unusual rate", function()
+      assert.logfile().has.line("loading upstreams dict into memory", true, 5)
+      -- turncate log
+      io.open("./servroot/logs/error.log", "w"):close()
```
flrgh, line 34:
```suggestion
      helpers.clean_logfile()
```

spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
```diff
@@ -0,0 +1,39 @@
...
+      ngx.sleep(20)
```
flrgh, line 35: There are `db_cache_ttl` and `db_cache_neg_ttl` kong.conf directives that can be used in order to speed this test up.

flrgh:
Hmm. I was `git blame`-ing my way around looking for the reason why the upstream negative TTL is hardcoded to 10s, and I found these:

* #5831
* #7002

So it seems there at least _was_ a reason to not cache an empty table because it was exacerbating another DNS balancer problem. Maybe it's not safe to revert this change?

CC @javierguerragiraldez

StarlightIbuki:
> Hmm. I was `git blame`-ing my way around looking for the reason why the upstream negative TTL is hardcoded to 10s, and I found these:
> 
> * [fix(balancer) don't cache an empty upstream name dict #5831](https://github.com/Kong/kong/pull/5831)
> * [fix(balancer) don't cache an empty upstream name dict #7002](https://github.com/Kong/kong/pull/7002)
> 
> So it seems there at least _was_ a reason to not cache an empty table because it was exacerbating another DNS balancer problem. Maybe it's not safe to revert this change? On the other hand, the original bugfix is also over 2 years old now, so it's also semi-plausible that this is not a problem anymore.
> 
> CC @javierguerragiraldez
> 
> **Edit:** I am now very tempted to believe that the bug fixed by #5831 might be due to this logic here:
> 
> https://github.com/Kong/kong/blob/f38b38ed53a9e0dbc967a69c6eda1e1d7481fc8d/kong/runloop/balancer/upstreams.lua#L96-L99
> 
> This doesn't appear to be correct usage of [lua-resty-mlcache](https://github.com/thibaultcha/lua-resty-mlcache). According to the docs, a single `nil` return value will be treated as negative cache if not accompanied by a second (error) return value:
> 
> ```lua
> -- arg1, arg2, and arg3 are arguments forwarded to the callback from the
> -- `get()` variadic arguments, like so:
> -- cache:get(key, opts, callback, arg1, arg2, arg3)
> 
> local function callback(arg1, arg2, arg3)
>     -- I/O lookup logic
>     -- ...
> 
>     -- value: the value to cache (Lua scalar or table)
>     -- err: if not `nil`, will abort get(), which will return `value` and `err`
>     -- ttl: override ttl for this value
>     --      If returned as `ttl >= 0`, it will override the instance
>     --      (or option) `ttl` or `neg_ttl`.
>     --      If returned as `ttl < 0`, `value` will be returned by get(),
>     --      but not cached. This return value will be ignored if not a number.
>     return value, err, ttl
> end
> ```
> 
> In light of that I think this change should be accompanied by fixing the return semantics of the function when there's a DB-related error:
> 
> ```diff
>     if err then
>       log(CRIT, "could not obtain list of upstreams: ", err)
> -      return nil
> +      return nil, err
>     end
> ```

Nice finding. Since we have now corrected the handling probably we could remove that option?

flrgh:
> Since we have now corrected the handling probably we could remove that option?

Interestingly, the hardcoded negative TTL came from #4301. [This thread](https://github.com/Kong/kong/pull/4301#discussion_r256135828) is the explanation I was looking for.

Many of the concerns raised there seem to still be somewhat valid, though I tend to think we should trust that `upstreams:each()` will always return an error on an unexpected failure.

I think a valuable change would be to detect the case where there are zero upstreams _and_ no error is encountered and add a debug log entry in this case. That way this scenario is at least observable if/when there's a reason to suspect that something isn't working correctly.

@locao what do you think about all of this?

spec/02-integration/06-invalidations/04-balancer_cache_correctness_spec.lua
```diff
@@ -0,0 +1,39 @@
...
+      assert.logfile().has.no.line("loading upstreams dict into memory", true)
```
flrgh, line 36:
```suggestion
      assert.logfile().has.no.line("loading upstreams dict into memory", true, 20)
```

I know it's not very obvious/intuitive, but the timeout is still useful in the negative case. It will continue checking the log file for the full 20s.

flrgh: Alternatively, I wonder if it's possible to cover this with a unit test instead?

StarlightIbuki (approved):
LGTM, but as I'm committing to this, we need someone else to review it.

flrgh (approved):
looks good to me :+1: @locao care to give this a quick review?

locao:
Reviewing this is in my backlog, sorry for the long delay. I'll check it asap.

StarlightIbuki:
@locao Just a reminder. Do you have time to take a look recently?

PR #8988, opened 2022-06-22
FT-2563 - FT-2563-no-daos-hash

hanshuebner:
### Summary

Older Kong version accepted hash-like tables for DAO definitions.  This is no longer supported, and with this change, a clear error message is produced if a plugin returns a non-array-like table for its DAO definitions.


kikito (approved):

spec/01-unit/01-db/01-schema/11-declarative_config/01-validate_spec.lua
```diff
@@ -908,4 +908,10 @@ describe("declarative config: validate", function()
 
     assert(declarative_config.load(plugins_set))
   end)
+
+  it("rejects plugin with custom DAO in old, hash-like format", function()
+       daos["dao-keywords"] = daos[1]
+       table.remove(daos, 1)
+       assert(not declarative_config.load(plugins_set))
```
kikito, line 915: A simple `assert` is ok for testing the happy path, but we should also test that the error we introduced is the one actually being returned.

```suggestion
       local ok, err = declarative_config.load(plugins_set)
       assert.falsy(ok)
       assert(<something about err>)
```

kikito (changes requested):

bungle (approved):

PR #8993, opened 2022-06-22
fix(*) drop luasocket in CLI

fffonion:
WIP, need to add ssl cafile support in `kong migrations` CLI; maybe that will benefit to `kong vaults` as well.

### Full changelog

* [Implement ...]
* [Add related tests]
* ...

### Issue reference

Fix #_[issue number]_

### DIFF
--- a/kong/cmd/migrations.lua
+++ b/kong/cmd/migrations.lua
@@ -6,6 +6,7 @@ local conf_loader = require "kong.conf_loader"
 local kong_global = require "kong.global"
 local prefix_handler = require "kong.cmd.utils.prefix_handler"
 local migrations_utils = require "kong.cmd.utils.migrations"
+local respwan_cli = require "kong.cmd.utils.respawn_cli"
 
 
 local lapp = [[
@@ -85,6 +86,8 @@ local function execute(args)
   local conf = assert(conf_loader(args.conf, {
     prefix = args.prefix
   }))
+
+  respwan_cli(conf)
 
   package.path = conf.lua_package_path .. ";" .. package.path
 
--- a/kong/cmd/migrations.lua
+++ b/kong/cmd/migrations.lua
@@ -6,7 +6,7 @@ local conf_loader = require "kong.conf_loader"
 local kong_global = require "kong.global"
 local prefix_handler = require "kong.cmd.utils.prefix_handler"
 local migrations_utils = require "kong.cmd.utils.migrations"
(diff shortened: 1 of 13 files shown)


Tieske:
@fffonion I don't think this can be done, for the reasons you found out. I discussed it with @bungle as well for Vaults and other pieces that need it as a fallback. Even if it can be done, I think it would take more hackery than would be wise...
The only proper way out I see would be for OR to support sockets in those phases, but that's a whole new challenge 🤷

fffonion:
@Tieske How do you feel about https://github.com/Kong/kong/pull/8993/commits/a6a1c72aaf33f461b7e5f509737e637f17584f52 🤔 ? I haven't really tested it, but it shows proof of concept.

bungle:
I feel it could be done. Not sure how hackery is respawning (but I think we were already doing it, perhaps we need to do it twice, once to gather trusted certs and then to do what it was already doing). On my mind I though about similar resolution to the issue. The vaults currently do resolving of process secrets on CLI in timer phase with cosockets (no fallback to cosockets and pass the secrets to init with env var or on reload with a file), but they are missing support for `trusted_certs` currently. I think `lua-resty-secrets` (which integrated vault solution does not use) may work differently as it uses `resty.socket` or similar lib embedded. I am excited if we can get this working.

Tieske:
`lua-resty-aws` and `lua-resty-gcp` both use LuaSocket fallbacks as well. Not sure how that works/fails with the current Vaults solution.

kong/cmd/migrations.lua
```diff
@@ -86,6 +87,8 @@ local function execute(args)
     prefix = args.prefix
   }))
 
+  respwan_cli(conf)
```
Tieske, line 90: typo `respawn_cli`

Tieske (commented):
This seems to fix the CLI, but how about the `init` and `init_worker` phases?

fffonion:
@Tieske it seems the only two parts that does network connection is:
1) cluster_events update its initial time (https://github.com/Kong/kong/pull/8993/files#diff-bf75cd7b30a4bf3453f56734db348cbed164bf32a6be22107e950c6f64fcd079L129) updated to the fallback method to use ngx.time() (let me know if it's a sane change)
2) cache_warmup: move to a timer, i think this is fine, as it's running at the last part of init_worker and no other components seems to rely on it.

hbagdi:

PR #8997, opened 2022-06-22
chore(docs) update documentation to match what PUT endpoints return FT-2589

hanshuebner:
### Summary

The documentation for PUT entity endpoints state that HTTP status 200 or 201 could be in the response, but [the code](https://github.com/Kong/kong/blob/f04772b75063075c5d3f756eab06af6f5dc61f73/kong/api/endpoints.lua#L486) always returns 200, regardless of whether the entity was created or updated.

### DIFF
--- a/autodoc/admin-api/data/admin-api.lua
+++ b/autodoc/admin-api/data/admin-api.lua
@@ -2302,7 +2302,7 @@ return {
       ]],
       response = [[
         ```
-        HTTP 201 Created or HTTP 200 OK
+        HTTP 200 OK
         ```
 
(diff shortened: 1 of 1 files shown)


bungle (approved):
@hanshuebner,

AFAIK this is correct. I think long time ago we didn't really knew if a row was updated or created, because we didn't do read-before-write and perhaps it was difficult to know with Cassandra. Since then we have had to add read-before-write to some places, so it might be possible now to distinguish them, and we could return 200 or 200. But perhaps we can look this later. For now I approve it.

PR #8998, opened 2022-06-23
fix(balancer) only check healthy state in algorithm

mayocream:
### Summary

Only check `balancer.healthy` in round_robin/consistent_hashing/least_connections algorithms.

https://github.com/Kong/kong/blob/1c313c6eb2e2304f73862dce76636cd09140a0c0/kong/runloop/balancer/round_robin.lua#L102-L106
https://github.com/Kong/kong/blob/6a8f3e267a2b7bd64d00cef78d55bd55f0088036/kong/runloop/balancer/consistent_hashing.lua#L173-L178
https://github.com/Kong/kong/blob/32018c13accd292bbb46341064c8334b9053c87a/kong/runloop/balancer/least_connections.lua#L76-L81

### DIFF
--- a/kong/runloop/balancer/balancers.lua
+++ b/kong/runloop/balancer/balancers.lua
@@ -560,10 +560,6 @@ end
 
 
 function balancer_mt:getPeer(...)
-  if not self.healthy then
-    return nil, "Balancer is unhealthy"
-  end
-
   if not self.algorithm or not self.algorithm.afterHostUpdate then
     return
   end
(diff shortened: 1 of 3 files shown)


kong/runloop/balancer/init.lua
```diff
@@ -357,6 +357,7 @@ local function execute(balancer_data, ctx)
     if not ip and
       (port == "No peers are available" or port == "Balancer is unhealthy")
     then
+      log(ERR, "failure to get a peer from the balancer: " .. port)
```
Tieske, line 360: this can potentially DoS the logs, especially at ERR level.

kong/runloop/balancer/balancers.lua
```diff
@@ -560,10 +560,6 @@ end
 
 
 function balancer_mt:getPeer(...)
-  if not self.healthy then
-    return nil, "Balancer is unhealthy"
-  end
-
```
Tieske, line 566: why should this be removed?

PR #8999, opened 2022-06-23
fix(admin) add the worker info to the response of `/timers`

ADD-SP:
### Summary

* style(runloop) use every timer instead of at timer to rebuild router and plugin iterator
* feat(AdminAPI) add the worker info to the response of `/timers`
* tests(helpers) extend function `wait_timer`

### DIFF
--- a/kong/globalpatches.lua
+++ b/kong/globalpatches.lua
@@ -72,43 +72,31 @@ return function(options)
     _G.native_timer_at = ngx.timer.at
     _G.native_timer_every = ngx.timer.every
 
-    local timerng
+    local _timerng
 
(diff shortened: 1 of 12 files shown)


kong/globalpatches.lua
```diff
@@ -101,6 +101,14 @@ return function(options)
       return timerng:every(interval, callback, ...)
     end
 
+    _G.named_timer_at = function (name, delay, callback, ...)
```
fffonion, line 104: i was thinking if we should properly move those functions that invokes timerng in a single table, like `kong.timer.named_at`.

fffonion (approved):

PR #9002, opened 2022-06-23
chore(version) Bumping the version to 3.0.0 in prep for release

tyler-ball:
### Summary

This is also to move to a "bump version right after release" (to the next unreleased version) instead of "bump version right before release" pattern. This makes it harder for us to accidentally re-release the same version, and ensures a given release
points to a single point in time for the code.

### Full changelog

* Bumped Kong to version 3.0.0 in preparation for release


yzyyi (approved):

flrgh (approved):

gszr (commented):
👋 Should we also ensure that `master` has a value set for `--suffix` - e.g., `dev`? This way, once we release 3.0.0.0 and bump master to `3.0.1`, `kong version` will be reported as `3.0.1-dev` (meaning that the master branch contains a development version of the upcoming release), avoiding possible confusions.

tyler-ball:
@gszr I think thats a decent idea. If we do that, I would like to capture that in another card because it will involve a change to the kong and kong-ee release processes.

I know the kong-ee nightlies already use the suffix to differentiate between internal and external releases, so that process would need to be tested.

Sitting here thinking about it, I think it is a decent amount of test work to ensure the entire package ecosystem (generated filenames, docker image names, upgrades, CP <-> DP sync, etc.) would handle that change well. Definitely another card.

PR #9004, opened 2022-06-24
fix(db) skip delete hooks if no rows were affected

SumiMakito:
### Summary

> This pull request cherry-picks https://github.com/Kong/kong-ee/pull/3622

This PR fixes the issue that DAO delete hooks were triggered when no rows were affected.

The issue currently affects Kong Enterprise and Kong Manager. When concurrent delete requests are initiated on the same entity, the counter in the workspace meta will potentially be decremented multiple times and become a negative number.

FT-3072

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -908,10 +908,13 @@ local function generate_foreign_key_methods(schema)
           return nil, tostring(err_t), err_t
         end
 
-        local _
-        _, err_t = self.strategy:delete_by_field(name, unique_value, options)
+        local rows_affected
+        rows_affected, err_t = self.strategy:delete_by_field(name, unique_value, options)
         if err_t then
           return nil, tostring(err_t), err_t
+
+        elseif not rows_affected then
+          return nil
         end
 
         entity, err_t = run_hook("dao:delete_by:post",
@@ -1294,10 +1297,13 @@ function DAO:delete(primary_key, options)
     return nil, tostring(err_t), err_t
   end
 
-  local _
-  _, err_t = self.strategy:delete(primary_key, options)
+  local rows_affected
+  rows_affected, err_t = self.strategy:delete(primary_key, options)
   if err_t then
     return nil, tostring(err_t), err_t
+
+  elseif not rows_affected then
+    return nil
   end
 
   entity, err_t = run_hook("dao:delete:post", entity, self.schema.name, options, ws_id, cascade_entries)
--- a/kong/db/strategies/cassandra/init.lua
+++ b/kong/db/strategies/cassandra/init.lua
@@ -1720,7 +1720,7 @@ function _mt:delete_by_field(field_name, field_value, options)
   end
 
   if not row then
-    return true
+    return nil, nil
   end
 
   local pk = self.schema:extract_pk_values(row)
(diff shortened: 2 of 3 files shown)


kong/db/dao/init.lua
```diff
@@ -891,10 +891,12 @@ local function generate_foreign_key_methods(schema)
           return nil, tostring(err_t), err_t
         end
 
-        local _
-        _, err_t = self.strategy:delete_by_field(name, unique_value, options)
+        local rows_affected
+        rows_affected, err_t = self.strategy:delete_by_field(name, unique_value, options)
         if err_t then
           return nil, tostring(err_t), err_t
+        elseif not rows_affected then
```
chronolaw, line 898: style issue: should add a blank line before `elseif`.

SumiMakito: This has been fixed. Thanks!

mayocream (approved):

dndx:
I think this test can be written as a DAO unit test. We shouldn't have to use Test-Nginx and introduce a new test mode for it. PM'ed @SumiMakito offline to help get this merged into 3.0 before code freeze.

kong/db/strategies/cassandra/init.lua
```diff
@@ -1720,7 +1720,7 @@ function _mt:delete_by_field(field_name, field_value, options)
   end
 
   if not row then
-    return true
+    return nil, nil
```
dndx, line 1723:
```suggestion
    return nil
```

Should be sufficient.

spec/01-unit/01-db/04-dao_spec.lua
```diff
@@ -3,6 +3,7 @@ local Entity = require("kong.db.schema.entity")
 local DAO = require("kong.db.dao.init")
 local errors = require("kong.db.errors")
 local utils = require("kong.tools.utils")
+local hooks = require "kong.hooks"
```
dndx, line 6:
```suggestion
local hooks = require("kong.hooks")
```

spec/01-unit/01-db/04-dao_spec.lua
```diff
@@ -618,6 +619,55 @@ describe("DAO", function()
...
+      end)
+
+      local schema = Schema.new({
+        name = "Baz",
+        primary_key = { "id" },
+        fields = {
+          { id = { type = "number" } },
+        }
+      })
+
```
dndx, line 642: We can do something like this:

```
local delete_called = false

...

         delete = function(pk, _)
           if not delete_called then
             delete_called = true
             return true
           end

           return nil
         end
```

spec/01-unit/01-db/04-dao_spec.lua
```diff
@@ -618,6 +619,55 @@ describe("DAO", function()
...
+          return nil
+        end
+      }
+
+      local dao = DAO.new({}, schema, strategy, errors)
+
+      local co = {}
+      local co_count = 10
+      for i = 1, co_count do
+        co[i] = coroutine.create(function()
```
dndx, line 661: We don't even need to run it in a coroutine, just a straight call to `dao:delete({ id = 1 })` twice is sufficient (this fix does not need concurrency to be tested).

spec/01-unit/01-db/04-dao_spec.lua
```diff
@@ -618,6 +619,55 @@ describe("DAO", function()
...
+          dao:delete({ id = 1 })
+        end)
+      end
```
dndx, line 664:
```suggestion
      end

```

SumiMakito:
@dndx Thanks! I have applied these suggestions.

PR #9009, opened 2022-06-24
feat(rate-limiting) support dynamic path filtering

nbialostosky:
### Summary

The current implementation of path based rate limiting only allows a single path identifier per route. However, there are situations where rate limiting by path would be helpful on a path that is dynamic, such as including a UUID in the path.

The changes included support this functionality by allowing the rate_limiting identifier to be automatically set to the path of the request. This means you don't need to specify which path to rate limit as it will be done automatically.

The existing test in `04-access_spec.lua`:

https://github.com/Kong/kong/blob/master/spec/03-plugins/23-rate-limiting/04-access_spec.lua#L396

Shows this change working by simply removing the `path` value set in the config, the test continues to function as expected as the path gets automatically pulled in.

### Full changelog

feat(rate-limiting) support dynamic path filtering

### Issue reference

N/A

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -70,10 +70,7 @@ local function get_identifier(conf)
     identifier = kong.request.get_header(conf.header_name)
 
   elseif conf.limit_by == "path" then
-    local req_path = kong.request.get_path()
-    if req_path == conf.path then
-      identifier = req_path
-    end
+    identifier = kong.request.get_path()
   end
 
   return identifier or kong.client.get_forwarded_ip()
--- a/kong/plugins/rate-limiting/schema.lua
+++ b/kong/plugins/rate-limiting/schema.lua
@@ -78,7 +78,6 @@ return {
               one_of = { "consumer", "credential", "ip", "service", "header", "path" },
           }, },
           { header_name = typedefs.header_name },
-          { path = typedefs.path },
           { policy = policy },
           { fault_tolerant = { type = "boolean", required = true, default = true }, },
           { redis_host = typedefs.host },
@@ -109,10 +108,6 @@ return {
     { conditional = {
       if_field = "config.limit_by", if_match = { eq = "header" },
       then_field = "config.header_name", then_match = { required = true },
-    } },
-    { conditional = {
-      if_field = "config.limit_by", if_match = { eq = "path" },
-      then_field = "config.path", then_match = { required = true },
     } },
     { conditional = {
       if_field = "config.policy", if_match = { eq = "redis" },
--- a/spec/03-plugins/23-rate-limiting/01-schema_spec.lua
+++ b/spec/03-plugins/23-rate-limiting/01-schema_spec.lua
@@ -24,13 +24,6 @@ describe("Plugin: rate-limiting (schema)", function()
     assert.is_nil(err)
   end)
 
-  it("proper config validates (path)", function()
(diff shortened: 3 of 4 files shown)


vm-001:
Hi, @nbialostosky . I'm not a reviewer though. Just wondering what kind of scenario need to limiting by the URL path. Because I think the client always can change the URL.


One alternative solution is to use limit_by header feature + custom plugin, which was described here: https://github.com/Kong/kong/pull/9965#issuecomment-1356081388

kikito:
Hello @nbialostosky , apologies for not getting back to you on this PR sooner.

We are very grateful for your contribution, but we decided not to include it in Kong. There's a concern and a strong reason:

* Concern: The changes are not fully tested.

Your change essentially removes checks from the plugin. It is expected that the existing tests will simply "keep passing" if all we do is remove checks.

> rate limiting by path would be helpful on a path that is dynamic, such as including a UUID in the path.

I would have expected at least one test that makes sure that the scenario described there is possible after the changes.

* Strong reason: We have doubts about the extra flexibility being used maliciously

Kong is used by all sorts of organizations - from individual users maintaining a personal API to big banks. Some of its users are very specific in what can and cannot be ratelimited, and in which manner.

A change like this (or along these lines) would be potentially breaking for some of those organizations. If we are going to introduce changes like what's proposed here, we would have to give ample warning, not make this behavior active by default, but instead an "opt-in", and probably release the change on a mayor release.

dcarley:
> Just wondering what kind of scenario need to limiting by the URL path. Because I think the client always can change the URL. Does it still make sense to use limiting by path in this case?

We have been using use this change in a fork of the plugin to provide individual rate limits to webhook endpoints. Each webhook is unique to an integration and shouldn't be called too frequently. There's low risk of a client using up the limit of another webhook or exceeding the capacity of the service because the webhook paths aren't trivially discoverable.

> One alternative solution is to use limit_by header feature + custom plugin, which was described here: https://github.com/Kong/kong/pull/9965#issuecomment-1356081388

That looks like it would be possible but it feels like a slightly odd coupling between plugins and we had doubts about the existing behaviour of the feature (see below).

> Your change essentially removes checks from the plugin. It is expected that the existing tests will simply "keep passing" if all we do is remove checks.

Rate limiting by path is still tested [here](https://github.com/nbialostosky/kong/blob/a73f32dfa772c4fcb5b5905063106475b86a9a96/spec/03-plugins/23-rate-limiting/04-access_spec.lua#L395-L462). Would it help to add a third path to that test?

> Kong is used by all sorts of organizations - from individual users maintaining a personal API to big banks. Some of its users are very specific in what can and cannot be ratelimited, and in which manner.
> 
> A change like this (or along these lines) would be potentially breaking for some of those organizations. If we are going to introduce changes like what's proposed here, we would have to give ample warning, not make this behavior active by default, but instead an "opt-in", and probably release the change on a mayor release.

You're right that this would be considered a breaking change. An alternative could be to introduce it as another parameter, although I think it would be difficult to name alongside the existing implementation, particularly as it's unclear to us how the current implementation is intended to be used.

It currently only matches one exact (not a prefix) path and all other paths that match the same plugin configuration will fallback to using the IP as the identifier. We couldn't find any further information in #6286, #6234, or #6202. Are you able to provide more information @kikito?

Would this change be accepted if the existing `path` parameter became optional to honour the existing behaviour?

Alternatively, would this change be accepted in a major release, and what is the process for doing that?

PR #9013, opened 2022-06-27
feat(PDK) add API: get_uri_captures

StarlightIbuki:
Fix FTI-3994

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -841,6 +841,24 @@ local function new(self)
     return ngx.ctx.KONG_PROCESSING_START or (req.start_time() * 1000)
   end
 
+  local EMPTY = {}
+  ---
+  -- Returns the URI captures matched by the router.
+  --
+  -- @function kong.request.get_uri_captures
+  -- @phases rewrite, access, header_filter, response, body_filter, log, admin_api
+  -- @treturn table A table containing captures.
+  -- @usage
+  -- local captures = kong.request.get_uri_captures()
+  -- for name_or_idx, value in pairs(captures) do
+  --   -- do what you want to captures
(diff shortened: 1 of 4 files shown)


kong/pdk/request.lua
```diff
@@ -840,6 +840,24 @@ local function new(self)
...
+  -- @treturn table A table containing captures.
+  -- @usage
+  -- local captures = kong.request.get_uri_captures()
+  -- for name_or_idx, value in pairs(captures) do
+  --   -- do what you want to captures
+  -- end
+  function _REQUEST.get_uri_captures()
+    check_phase(PHASES.request)
+
+    return ngx.ctx.router_matches.uri_captures or EMPTY
```
windmgc, line 858: I'm seeing somewhere in one of the EE codes(you can search it in EE repo, some plugin code) that write as follows
`return (ngx.ctx.router_matches or EMPTY).uri_captures or EMPTY`
And I wonder what's the difference between these two lines(is it `ngx.ctx.router_matches` that might not exist so it would be safer?)

Suika-Kong: `ngx.ctx` is an API of OpenResty, and if that have been not exist, it would be a serious bug.

Suika-Kong: And I see no difference from `(ngx.ctx.router_matches or EMPTY).uri_captures or EMPTY` than `ngx.ctx.router_matches or EMPTY`.

windmgc:
+1 for this, we need URI captures exposed in PDK as a convenient and clear interface that can be used in plugin code

StarlightIbuki:
close because of the new PR: https://github.com/Kong/kong/pull/9512

PR #9021, opened 2022-06-28
fix(clustering) cp working with 2.8 dp

Suika-Kong:
We coded CP to blindly reject any DP with a different major version, but we need 3.0 to be compatible with 2.8.

fix FT-3069


chronolaw:
It is a hard code, I know we have to do so, hope we will remove these code in the future version.

mayocream (approved):

Suika-Kong:
Convert to draft to make it less apparent in my list.

PR #9022, opened 2022-06-28
fix(core) router path regex detection

Suika-Kong:
We don't have a good design at first to tell if a `router.path` is a regex or not. We are guessing with a magical regex to match against the path. But that regex will consider strings containing any characters beyond 126 as a regex.

### DIFF
--- a/kong/router.lua
+++ b/kong/router.lua
@@ -342,6 +342,15 @@ local function create_range_f(ip)
   end
 end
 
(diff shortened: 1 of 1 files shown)


Suika-Kong:
This fix has little impact on routers' behavior, and it should have better performance than before.

bungle (approved):

Suika-Kong:
Closing this because of #9027.

PR #9024, opened 2022-06-28
fix(core) stop percent decoding regex path

Suika-Kong:
It's not a very reliable and sound way to support percent-encoding in regex. We choose to tell users that we have a normalized (standard) form to match with so there's no ambiguity.

https://github.com/Kong/kong/pull/8140#issuecomment-1168371378

fix CT-344

### DIFF
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -304,10 +304,10 @@ local function c_normalize_regex_path(coordinator)
 
       local changed = false
       for i, path in ipairs(route.paths) do
-        print(path)
         if not is_regex(path) then
           goto continue
         end
+
         local normalized_path = migrate_regex(path)
         if normalized_path ~= path then
           changed = true
@@ -336,17 +336,16 @@ end
 
 local function p_migrate_regex_path(connector)
   for route, err in connector:iterate("SELECT id, paths FROM routes") do
-    print(require "inspect" (route))
     if err then
       return nil, err
     end
 
     local changed = false
     for i, path in ipairs(route.paths) do
-      print(path)
       if not is_regex(path) then
         goto continue
       end
+
       local normalized_path = migrate_regex(path)
       if normalized_path ~= path then
         changed = true
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -283,7 +283,7 @@ do
   end
 end
 
-local function is_regex(path)
+local function is_not_regex(path)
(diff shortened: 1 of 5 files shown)


Suika-Kong:
Percent decoding of `router.path` is an undocumented feature, so we don't need a change to document for this PR.

kong/db/migrations/core/016_280_to_300.lua
```diff
@@ -1,10 +1,14 @@
 local log = require "kong.cmd.utils.log"
-
+local arrays = require "pgmoon.arrays"
 
 local fmt = string.format
 local assert = assert
 local ipairs = ipairs
 local cassandra = require "cassandra"
```
chronolaw, line 7: move this line to the top of file

kong/db/migrations/core/016_280_to_300.lua
```diff
@@ -211,6 +215,179 @@ local function c_drop_vaults_beta(coordinator)
...
+  return true
+end
+
+local function render(template, keys)
+  return (template:gsub("$%(([A-Z_]+)%)", keys))
+end
+
+local function p_migrate_regex_path(connector)
+  for route, err in connector:iterate("SELECT id, paths FROM routes") do
+    print(require "inspect" (route))
```
chronolaw, line 339: remove debug code.

kong/db/migrations/core/016_280_to_300.lua
```diff
@@ -211,6 +215,179 @@ local function c_drop_vaults_beta(coordinator)
...
+    if err then
+      return nil, err
+    end
+
+    local changed = false
+    for i, path in ipairs(route.paths) do
```
chronolaw, line 345: localize `ipairs`

spec/01-unit/08-router_spec.lua
```diff
@@ -1477,6 +1477,7 @@ describe("Router", function()
           },
         },
         -- regex
+        -- regex will no longer be normalized since 3.0
```
chronolaw, line 1480: duplicated comments, could we do some clean?

Suika-Kong: The first is a title and the second is an explanation. Surely it is ambiguous.

Suika-Kong:
I've tested migration in my local env and it works as expected.

spec/01-unit/08-router_spec.lua
```diff
@@ -1476,7 +1476,7 @@ describe("Router", function()
             },
           },
         },
-        -- regex
+        -- regex. It will no longer be normalized since 3.0
```
fffonion, line 1479:
```suggestion
        -- regex. It‘s no longer be normalized since 3.0
```

fffonion (approved):

PR #9027, opened 2022-06-29
feat(core) explict flag for regex route.path

Suika-Kong:
Another solution for #9022.

fix CT-344

We need to remember to add migration, and also "back migration" for EE.

**PLEASE MERGE AFTER: #9024**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,7 +125,7 @@
   [#8988](https://github.com/Kong/kong/pull/8988)
 - `ngx.ctx.balancer_address` does not exist anymore, please use `ngx.ctx.balancer_data` instead.
   [#9043](https://github.com/Kong/kong/pull/9043)
-- Stop normalizing regex `router.path`. Regex path pattern matches with normalized URI,
+- Stop normalizing regex `route.path`. Regex path pattern matches with normalized URI,
   and we used to replace percent-encoding in regex path pattern to ensure different forms of URI matches.
   That is no longer supported. Except for reserved characters defined in
   [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,8 +131,8 @@
   [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2),
   we should write all other characters without percent-encoding.
   [#9024](https://github.com/Kong/kong/pull/9024)
-- Use `"~*"` as prefix to indicate a `route.path` is a regex pattern. We no longer guess
-  whether a path pattern is a regex, and all path without the `"~*"` prefix is considered plain text.
+- Use `"~"` as prefix to indicate a `route.path` is a regex pattern. We no longer guess
+  whether a path pattern is a regex, and all path without the `"~"` prefix is considered plain text.
   [#9027](https://github.com/Kong/kong/pull/9027)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,7 +125,7 @@
   [#8988](https://github.com/Kong/kong/pull/8988)
 - `ngx.ctx.balancer_address` does not exist anymore, please use `ngx.ctx.balancer_data` instead.
   [#9043](https://github.com/Kong/kong/pull/9043)
-- Stop normalizing regex `route.path`. Regex path pattern matches with normalized URI,
+- Stop normalizing regex `router.path`. Regex path pattern matches with normalized URI,
   and we used to replace percent-encoding in regex path pattern to ensure different forms of URI matches.
   That is no longer supported. Except for reserved characters defined in
   [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,9 @@
   [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2),
   we should write all other characters without percent-encoding.
   [#9024](https://github.com/Kong/kong/pull/9024)
+- Use `"~*"` as prefix to indicate a `router.path` is a regex pattern. We no longer guess
+  whether a path pattern is a regex, and all path without the `"~*"` prefix is considered plain text.
+  [#9027](https://github.com/Kong/kong/pull/9027)
 
 
 #### Admin API
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,7 +131,7 @@
   [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2),
   we should write all other characters without percent-encoding.
   [#9024](https://github.com/Kong/kong/pull/9024)
-- Use `"~*"` as prefix to indicate a `router.path` is a regex pattern. We no longer guess
+- Use `"~*"` as prefix to indicate a `route.path` is a regex pattern. We no longer guess
   whether a path pattern is a regex, and all path without the `"~*"` prefix is considered plain text.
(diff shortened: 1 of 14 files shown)


kong/db/schema/typedefs.lua
```diff
@@ -449,15 +452,15 @@ local function validate_path_with_regexes(path)
   -- We can't take an ok from validate_path as a success just yet,
   -- because the router is currently more strict than RFC 3986 for
   -- non-regex paths:
-  if ngx.re.find(path, [[^[a-zA-Z0-9\.\-_~/%]*$]]) then
+  if path:sub(1,2) ~= "~*" then
```
chronolaw, line 455: should have a space, like `sub(1, 2)`

kong/router.lua
```diff
@@ -425,8 +425,9 @@ local function marshall_route(r)
       match_weight = match_weight + 1
       for i = 1, count do
         local path = paths[i]
+        local is_regex = path:sub(1,2) == "~*"
```
chronolaw, line 428: should have a space, like `sub(1, 2)`

spec/02-integration/05-proxy/02-router_spec.lua
```diff
@@ -2191,6 +2191,12 @@ for _, strategy in helpers.each_strategy() do
                 }))
 
                 local data = assert.response(res).has.jsonbody()
+                if not data.vars then
```
chronolaw, line 2194: Is this a debug code?

Suika-Kong: You're right.

chronolaw: Should we remove them?

CHANGELOG.md
```diff
@@ -128,6 +128,9 @@
   [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2),
   we should write all other characters without percent-encoding.
   [#9024](https://github.com/Kong/kong/pull/9024)
+- Use `"~*"` as prefix to indicate a `router.path` is a regex pattern. We no longer guess
```
chronolaw, line 131: maybe ' the `router.path`' ?

chronolaw (approved):

bungle (dismissed):
It is wrong to use `~*` prefix and do case-sensitive matching. The `~` should be case-sensitive match. See: https://github.com/Kong/kong/blob/master/kong/router.lua#L960

We only add `ajo` and not `i` modifier.

Suika-Kong:
> NOTE: this will break all existing:
> 
> 1. `scripts` that call admin apis and want regexes.
> 2. `yaml` (and other config files) that are used to configure dbless, or import data into control plane, traditional node, or imported by deck.
> 3. `kong ingress controller` might also be affected
> 
> In that sense I think this is a too big of a change that comes too close to `3.0.0` release. That said, `major` version is the version to do these kind of changes. It can be a can of worms. @dndx could you comment on this?

Right. I will write a document about that to notify the people involved.

CHANGELOG.md
```diff
@@ -129,6 +129,9 @@
   [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2),
   we should write all other characters without percent-encoding.
   [#9024](https://github.com/Kong/kong/pull/9024)
+- Use `"~*"` as prefix to indicate a `route.path` is a regex pattern. We no longer guess
```
chronolaw, line 132: Now we should use `~` instead of `~*`?

Suika-Kong:
I've tested migration in my local env and it works as expected.

fffonion (approved):

kong/db/schema/typedefs.lua
```diff
@@ -449,10 +452,12 @@ local function validate_path_with_regexes(path)
   -- We can't take an ok from validate_path as a success just yet,
   -- because the router is currently more strict than RFC 3986 for
   -- non-regex paths:
```
bungle, line 454: Is this comment relevant anymore or perhaps wrong?

kong/db/schema/typedefs.lua
```diff
@@ -449,10 +452,12 @@ local function validate_path_with_regexes(path)
...
-  if ngx.re.find(path, [[^[a-zA-Z0-9\.\-_~/%]*$]]) then
+  if path:sub(1, 1) ~= "~" then
     return true
   end
 
+  path = path:sub(2)
+
   -- URI contains characters outside of the list recognized by the
   -- router as valid non-regex paths.
```
bungle, line 462: Same here.

kong/router.lua
```diff
@@ -425,8 +425,9 @@ local function marshall_route(r)
       match_weight = match_weight + 1
       for i = 1, count do
         local path = paths[i]
+        local is_regex = path:sub(1, 1) == "~"
```
bungle, line 428: `sub` is already localized so:

```suggestion
        local is_regex = sub(path, 1, 1) == "~"
```

kong/router.lua
```diff
@@ -440,6 +441,7 @@ local function marshall_route(r)
 
         else
 
+          path = path:sub(2)
```
bungle, line 444:
```suggestion
          path = sub(path, 2)
```

bungle (changes requested):

kong/db/schema/typedefs.lua
Suika-Kong: You are right. It doesn't hold now.

kong/db/schema/typedefs.lua
```diff
@@ -295,7 +295,10 @@ typedefs.port = Schema.define {
 
 typedefs.path = Schema.define {
   type = "string",
-  starts_with = "/",
+  match_any = {
+    patterns = {"^/", "^~/"},
+    err = "should start with: /",
```
kikito, line 300:
```suggestion
    err = "should start with: / (fixed path) or ~/ (regex path)",
```

kong/db/schema/typedefs.lua
```diff
@@ -446,15 +449,12 @@ local function validate_path_with_regexes(path)
     return ok, err, err_code
   end
 
-  -- We can't take an ok from validate_path as a success just yet,
-  -- because the router is currently more strict than RFC 3986 for
-  -- non-regex paths:
-  if ngx.re.find(path, [[^[a-zA-Z0-9\.\-_~/%]*$]]) then
+  if path:sub(1, 1) ~= "~" then
```
kikito, line 452: Could this be made backwards compatible by extending this with [#9022](https://github.com/Kong/kong/pull/9022) (keeping the heuristic alive for 3.0, but deprecated)?

```suggestion
  if path:sub(1, 1) ~= "~" and is_not_regex(path) then
```

kong/db/schema/typedefs.lua
kikito, line 300:
```suggestion
    err = "should start with: / (fixed path) or ~/ (regex path)",
```

kong/router.lua
```diff
@@ -425,8 +425,9 @@ local function marshall_route(r)
       match_weight = match_weight + 1
       for i = 1, count do
         local path = paths[i]
+        local is_regex = sub(path, 1, 1) == "~"
```
kikito, line 428:
```suggestion
        local is_regex = sub(path, 1, 1) == "~" or is_regex(path)
```

Suika-Kong:
> Could this be made backwards compatible by extending this with [#9022](https://github.com/Kong/kong/pull/9022) (keeping the heuristic alive for 3.0, but deprecated)?

@kikito I've discussed this with @dndx and we decided not to apply this solution.
We will have the traditional format of route coexist with ATC at least for 3.X, and the problem we want to address with those 2 PRs are important. And @dndx thinks this solution introduces more complexity.

PR #9034, opened 2022-06-30
feat(clustering) switch to use old protocol

Suika-Kong:
This should be considered a non-user configuration, and we also should avoid using it if possible.

And this feature is intentionally not documented.

Fix FT-3077

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -71,7 +71,7 @@ function _M:init_dp_worker(plugins_list)
     end
 
     local config_proto, msg
-    if not kong.configuration.hybrid_legacy_protocol then
+    if not kong.configuration.legacy_hybrid_protocol then
       config_proto, msg = check_protocol_support(self.conf, self.cert, self.cert_key)
       -- otherwise config_proto = nil
     end
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -71,7 +71,7 @@ function _M:init_dp_worker(plugins_list)
     end
 
     local config_proto, msg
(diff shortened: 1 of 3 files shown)


chronolaw:
I think you should also add a item in `kong/templates/kong_defaults.lua`

kong/clustering/init.lua
```diff
@@ -70,7 +70,11 @@ function _M:init_dp_worker(plugins_list)
       return
     end
 
-    local config_proto, msg = check_protocol_support(self.conf, self.cert, self.cert_key)
+    local config_proto, msg
+    if not kong.configuration.force_no_wrpc then
```
chronolaw, line 74: `force_no_wrpc` is a little strange, could we rename it to `legacy_hybrid_protocol` or else?

Suika-Kong: Maybe `hybrid_force_legacy_protocol`?

Suika-Kong: or `hybrid_legacy_protocol` for consistency with `legacy_worker_event`. (hybrid is the scope.)

kong/conf_loader/init.lua
```diff
@@ -528,6 +528,7 @@ local CONF_INFERENCES = {
   untrusted_lua_sandbox_environment = { typ = "array" },
 
   legacy_worker_events = { typ = "boolean" },
+  hybrid_legacy_protocol = { typ = "boolean", },
```
chronolaw, line 531: Here may not need a `,`, we can follow other code in context.

kikito (approved):

PR #9035, opened 2022-06-30
perf(tracing) localize functions and cache the ngx.ctx calls

bungle:
### Summary

Localize functions and cache the ngx.ctx calls in tracing/instrumentation.


mayocream (approved):
LGTM!

PR #9038, opened 2022-06-30
perf(core) use table.new for ngx.ctx and initialize it with 50 records to avoid resizing

bungle:
### Summary

Alternative to #9036 (this time without `tablepool`).


bungle:
The https://github.com/Kong/kong/pull/9036 seems to be better, thus closing this.

PR #9039, opened 2022-07-01
fix(globalpatches) fallback to alternative_sleep at init phase

vm-001:
### Summary

Kong Enterprise invoked `ngx.sleep` in `init` phase in some scenarios(e.g. keyring failed). But the current patch for `ngx.sleep` doesn't support fallback to blocking sleep in `init` phase.

### Full changelog

* fix(globalpatches) fallback to alternative_sleep at init phase

### Issue reference

NO issue related

### DIFF
--- a/kong/globalpatches.lua
+++ b/kong/globalpatches.lua
@@ -57,9 +57,12 @@ return function(options)
     end
 
     -- luacheck: globals ngx.sleep
+    local blocking_sleep_phases = {
+      init = true,
+      init_worker = true,
+    }
     ngx.sleep = function(s)
-      local phase = get_phase()
-      if phase == "init" or phase == "init_worker" then
+      if blocking_sleep_phases[get_phase()] then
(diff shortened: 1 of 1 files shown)


kong/globalpatches.lua
```diff
@@ -58,7 +58,8 @@ return function(options)
 
     -- luacheck: globals ngx.sleep
     ngx.sleep = function(s)
-      if get_phase() == "init_worker" then
+      local phase = get_phase()
+      if phase == "init" or phase == "init_worker" then
```
Tieske, line 62: Can we do something like;
```lua
    local blocking_sleep_phases = {
      init = true,
      init_worker = true,
    }
    -- luacheck: globals ngx.sleep
    ngx.sleep = function(s)
      if blocking_sleep_phases[get_phase()] then
```

is a nitpick, but makes it more explicit imho

Tieske (approved):

kikito (approved):

PR #9040, opened 2022-07-01
chore(*) atc router preflight

chronolaw:
### Summary

Kong 3.0 introduce a new atc router, but it may have slightly incompatible with the traditional router.
This lua script try to query all router rules by admin API and validate them.


chronolaw:
Due to #9211, now this PR seems unnecessary.

chronolaw:
We close this PR for now.

PR #9047, opened 2022-07-04
feat(plugins) allow ACME to accept any domain

Suika-Kong:
Fix FTI-3300

### DIFF
--- a/kong/plugins/acme/handler.lua
+++ b/kong/plugins/acme/handler.lua
@@ -59,6 +59,16 @@ function ACMEHandler:init_worker()
   ngx.timer.every(86400, client.renew_certificate)
 end
 
+local function verify_domains(conf, host)
+  if not conf.skip_domain_verify then
+    return true
+  end
+
+  -- TODO: cache me
+  local domains_matcher = build_domain_matcher(conf.domains)
+  return domains_matcher and domains_matcher[host]
+end
+
 function ACMEHandler:certificate(conf)
   -- we can't check for Host header in this phase
   local host, err = ngx_ssl.server_name()
@@ -72,9 +82,7 @@ function ACMEHandler:certificate(conf)
 
   host = string.lower(host)
 
-  -- TODO: cache me
-  local domains_matcher = build_domain_matcher(conf.domains)
-  if not domains_matcher or not domains_matcher[host] then
+  if not verify_domains(conf, host) then
     kong.log.debug("ignoring because domain is not in whitelist")
     return
   end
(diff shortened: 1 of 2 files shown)


kong/plugins/acme/schema.lua
```diff
@@ -100,6 +100,10 @@ local schema = {
           default = 14,
         }, },
         { domains = typedefs.hosts },
+        { skip_domain_verify = {
```
fffonion, line 103: the option seems bit confusing as we are not really verifying a domain here, how about something like `allow_all_domains`

CHANGELOG.md
```diff
@@ -177,6 +177,8 @@
   context (ngx.ctx.authenticated_jwt_token).  Custom plugins which depend on that
   value being set under that name must be updated to use Kong's shared context
   instead (kong.ctx.shared.authenticated_jwt_token) before upgrading to 3.0
+- **ACME**: `allow_any_domain` field added. It is default to false and if set to true, the gateway will
+  not validate host with `domains` field.
```
fffonion, line 181:
```suggestion
- **ACME**: `allow_any_domain` field added. It is default to false and if set to true, the plugin will
  not ignore `domains` field.
```

kong/plugins/acme/handler.lua
```diff
@@ -59,6 +59,16 @@ function ACMEHandler:init_worker()
   ngx.timer.every(86400, client.renew_certificate)
 end
 
+local function verify_domains(conf, host)
```
fffonion, line 62:
```suggestion
local function check_domain(conf, host)
```

kong/plugins/acme/handler.lua
```diff
@@ -72,9 +82,7 @@ function ACMEHandler:certificate(conf)
 
   host = string.lower(host)
 
-  -- TODO: cache me
-  local domains_matcher = build_domain_matcher(conf.domains)
-  if not domains_matcher or not domains_matcher[host] then
+  if not verify_domains(conf, host) then
```
fffonion, line 85:
```suggestion
  if not check_domain(conf, host) then
```

kong/plugins/acme/handler.lua
fffonion: since we are not doing verification, and to avoid confusion with the validation that ACME did itself, let's use another word

kong/plugins/acme/handler.lua
```diff
@@ -160,8 +168,8 @@ function ACMEHandler:access(conf)
       return
     end
 
-    local domains_matcher = build_domain_matcher(conf.domains)
-    if not domains_matcher or not domains_matcher[kong.request.get_host()] then
+    if not verify_domains(conf, host) then
```
fffonion, line 171:
```suggestion
    if not check_domain(conf, host) then
```

fffonion (approved):

PR #9049, opened 2022-07-04
fix(db) run select transformations before process auto fields

bungle:

### DIFF
--- a/kong/db/dao/init.lua
+++ b/kong/db/dao/init.lua
@@ -477,7 +477,6 @@ end
 
 
 local function check_update(self, key, entity, options, name)
-
   local transform
   if options ~= nil then
     local ok, errors = validate_options_value(self, options)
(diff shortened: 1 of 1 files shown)


jschmid1:
To add some context:

When a keyring is present & configured, all fields that are marked with `encrypted` are being stored encrypted in the datastore. When this field needs to be read again, we have to make sure to decrypt it _before_ dereferencing it. 

If `process_auto_fields` - this is where the a potential reference is being _de_referenced, runs before `Schema:transform()` we're not able to _de_reference a field as we're still seeing encrypted data.

@bungle Since this is somewhat of a regression, I'd advocate for adding a small test to ensure we're not running into this issue again. wdyt?

bungle:
> > @bungle Since this is somewhat of a regression, I'd advocate for adding a small test to ensure we're not running into this issue again. wdyt?
> 
> @jschmid1, I agree. Thank you for initial review.

The test is now added.

jschmid1 (approved):
🎉

PR #9055, opened 2022-07-06
Revert "feat(clustering) switch to use old protocol"

fffonion:
Reverts Kong/kong#9034


chronolaw:
Could you explain the reason for this PR?

chronolaw:
As we talked offline, this switch is useful in recent days, let's keep it.

PR #9059, opened 2022-07-06
perf(*) return setmetatable is NYI that can not be jit compiled

bungle:
### Summary

Not sure if this has any effect in perf, but lets see.


bungle:
probably not worth it, closing.

PR #9061, opened 2022-07-07
Tests/wrpc switch

Suika-Kong:
Add a test to know if the switch works.
Fix FT-3077

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -3021,6 +3021,7 @@ local function get_clustering_protocols()
   local confs = {
     wrpc = "spec/fixtures/custom_nginx.template",
     json = "/tmp/custom_nginx_no_wrpc.template",
+    ["json (by switch)"] = "spec/fixtures/custom_nginx.template",
   }
 
   -- disable wrpc in CP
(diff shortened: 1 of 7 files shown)


spec/helpers.lua
```diff
@@ -3021,6 +3021,7 @@ local function get_clustering_protocols()
   local confs = {
     wrpc = "spec/fixtures/custom_nginx.template",
     json = "/tmp/custom_nginx_no_wrpc.template",
```
fffonion, line 3023: does this still need a custom template? can we use the #9071 instead?

Suika-Kong: If we rely on this switch to implement the test, then it would be difficult to remove this switch.

fffonion: that's fine, this will become a test for the switch itself, since we don't have such test. and we will know what will break/change when the test is removed.

Suika-Kong: The "json" part is to make sure 3.0 CP works with 2.8 DPs, and they will not be removed when the switch is removed.

chronolaw (approved):

PR #9062, opened 2022-07-07
tests(wrpc) wRPC protocol 

Suika-Kong:
Add test for wRPC protocol itself.

Fix FT-3097


spec/01-unit/19-hybrid/04-wrpc_spec.lua
```diff
@@ -0,0 +1,273 @@
...
+        clients[i] = new_pair()
+      end
+
+      for i = 0, message_n - 1 do
+        local client = (i % client_n) + 1
+        local future = assert(clients[client]:call(echo_service, { message = message, }))
+        ngx.timer.at(0, counter, future)
+      end
+
+      assert(smph:wait(5) and time < 5, "wrpc spent too much time")
```
fffonion, line 269: how much time does it take now?

spec/fixtures/mocks/lua-resty-websocket/resty/websocket/peer.lua
```diff
@@ -0,0 +1,161 @@
...
+end
+
+function _M:send_binary(data)
+  return self:send_frame(true, 0x2, data)
+end
+
+function _M:send_close(code, msg)
+  local payload
+  if code then
+    payload = encode_code(code) .. (msg or "")
```
fffonion, line 140: maybe instead of always make a string, just put what we receive in a table, so you don't need to encode/decode, like `payload = {code, msg}`

spec/01-unit/19-hybrid/04-wrpc_spec.lua
```diff
@@ -0,0 +1,273 @@
+local helpers = require "spec.helpers"
```
fffonion, line 1: this is probably still too blackbox-ish for a unit test, let's move it to integration test folder. for unit test, it would be useful to test following items:
- proto.lua: parse annotations; a RPC message be succefully handled only after set_handler is called
- queue.lua, threads.lua, future.lua: functional test for exposed functions

spec/01-unit/19-hybrid/04-wrpc_spec.lua
Suika-Kong: less than 1 second in my set-up.

fffonion (approved):

PR #9067, opened 2022-07-08
fix(clustering): data-plane node id keeps changing after restarted in hybrid mode

vm-001:
### Summary

Kong creates a new node_id  every time it restarts. In hybrid mode, CP will insert a new row to the clustering_data_planes  table after DP restart because it provides a new node_id. If DPs restart enough times in a period of time which is defined by cluster_data_plane_purge_delay, clustering_data_planes rows will increase a lot. This most likely fills up the prometheus_metrics  shared dict. 

### Full changelog

* Avoid repeatedly generating different node_id after kong restarted by writing node_id to file

### Issue reference

FTI-4168


kong/pdk/node.lua
```diff
@@ -49,6 +50,25 @@ end
 
 
 local function new(self)
+  if self and self.configuration and self.configuration.prefix then
```
flrgh, line 53: I think it would be a good idea to add some logging for a couple conditions in this change:

* when the filename exists but cannot be read or doesn't contain a valid UUID
* when a new `node_id` is generated, but persisting it to the filesystem fails

Doesn't need to be a super alarming message--probably `ngx.WARN` level? Probably would be helpful to operators in the case of `{{ prefix }}` being r/o or having too restrictive of permissions. WDYT?

kong/pdk/node.lua
```diff
@@ -49,6 +50,25 @@ end
...
+      if utils.is_valid_uuid(id) then
+        node_id = id
+      end
+    end
+
+    if not node_id then
+      local id = utils.uuid()
+      local ok = pl_file.write(filename, id)
+      if ok then
+        node_id = id
```
flrgh, line 66: Why only store `node_id` in the happy path?

flrgh (commented):
Great idea @vm-001 :grinning: I am surprised this was not done before.

Mind adding an integration test for this?

kong/pdk/node.lua
mayocream: @flrgh +1, I think we removed the hybrid dp cache file in 3.0; I remember the file open could be a failure with insufficient permissions.
Let's add error logs here and ensure to fall back to random id when failed.

vm-001:
> Great idea @vm-001 😀 I am surprised this was not done before.
> 
> Mind adding an integration test for this?

Added.

kong/pdk/node.lua
vm-001: @flrgh @mayocream  I've added some logs in some failure scenarios. Let me know if these logs can be improved. 😃

kong/cluster_events/init.lua
```diff
@@ -133,7 +131,7 @@ function _M.new(opts)
   end
 
   -- set node id (uuid)
-
+  local knode = (kong and kong.node) and kong.node or require "kong.pdk.node".new()
```
flrgh, line 134: There is an extra conditional in here that can be removed:

```suggestion
  local knode = kong and kong.node or require "kong.pdk.node".new()
```

```
# when `kong` is nil
$ lua -e 'kong = nil; print(kong and kong.node or "fallback")'
fallback
```
```
# when `kong.node` is nil
$ lua -e 'kong = {}; print(kong and kong.node or "fallback")'
fallback
```
```
# when `kong` and `kong.node` are not nil
$ lua -e 'kong = { node = "kong.node" }; print(kong and kong.node or "fallback")'
kong.node
```

Alternatively, you can preserve the parens to demonstrate explicit precedence (for the sake of the reader) while removing the excess conditional:

```lua
local knode = (kong and kong.node) or require "kong.pdk.node".new()
```

kong/pdk/node.lua
flrgh: @vm-001 Sorry for the late response to this. Assuming this is all temporarily commented-out, it looks pretty good to me.

One suggestion I have is that it would be helpful to include full filesystem paths in each of the unhappy branches so that operators can easily see which file/directory caused something to fail:

```lua
-- NOTE: it's possible that the `read_err` string already contains the full path to `filename`, 
-- making this redundant, so it's best to verify it yourself before making this change
ngx.log(ngx.WARN, "failed to read node id from '", filename, "': ", read_err)
```

Also, please make sure to use varargs in all of your `ngx.log` calls instead of string concatenation:

```lua
-- no
ngx.log(ngx.WARN, "an error happened: " .. err)

-- yes
ngx.log(ngx.WARN, "an error happened: ", err)
```

flrgh: Oh, and this is just a personal opinion of mine, but it can be good to log the side effect when issuing a warning so that operators know exactly what the consequence is (and whether or not they should be concerned about it). To illustrate:

```lua
ngx.log(ngx.WARN, "failed to read node id from '", filename, "': ", err,
                  " (a new node id will be generated)")
```

Some might find this overly verbose, which is a valid take, so I leave it up to your judgment :)

vm-001:
Sorry for leaving this PR for some time since I have some high-priority tasks that need to work on.

After some digging, I found that if Kong uses a single file to store the `node_id` and share with both `http` and `stream` mode. This will cause some critical issues in cluster event, that is the cluster events that are published by the `http` mode instance will be ignored in the `stream` mode instance as they shared the same node id in the filesystem, which is wrong.

So I decide to use different file to store the node id for the different ngx mode.

- `${prefix}/node.id/http` for http mode.
- `${prefix}/node.id/stream` for stream mode.

kong/cluster_events/init.lua
vm-001: @flrgh Thank you for your detailed explanation!

done in this commit: https://github.com/Kong/kong/pull/9067/commits/adfe946c5cb4a24d8f54f68ee89deebda61e9bd9

kong/pdk/node.lua
vm-001: @flrgh  Thanks for your suggestion. As too many debug commit here, I made a `git push -f`. Do you have time to review it again?

hanshuebner (changes requested):
See my top-level comment.

flrgh:
> After some digging, I found that if Kong uses a single file to store the `node_id` and share with both `http` and `stream` mode. This will cause some critical issues in cluster event, that is the cluster events that are published by the `http` mode instance will be ignored in the `stream` mode instance as they shared the same node id in the filesystem, which is wrong.
> 
> So I decide to use different file to store the node id for the different ngx mode.
> 
>     * `${prefix}/node.id/http` for http mode.
> 
>     * `${prefix}/node.id/stream` for stream mode.

Hmm. Is there any different way we can address the cluster events problem? I do realize this may greatly increase the scope of your efforts, but in my opinion having two distinct "node" IDs is not an optimal solution and could lead to more confusion further down the line.

> The maintenance of the node ID file should be in one lua file and share the file name creation logic.

Agree. Putting all the logic in `kong.global` is one idea. Putting it in a submodule in the `kong.pdk.private` namespace is another idea, assuming this doesn't put us in a chicken<->egg situation with regards to instantiating the PDK.

vm-001:
> Hmm. Is there any different way we can address the cluster events problem? I do realize this may greatly increase the scope of your efforts, but in my opinion having two distinct "node" IDs is not an optimal solution and could lead to more confusion further down the line.

The alternative solution is probably to use one node_id to share both http and stream mode, but the sub_system may have to be stored in the database to distinguish which mode it was for later use in the /cluster_events/init.lua#process_event(). The tricky part is not adding a new field in the cluster_events table, but the migration for the old event -- how do we determine which mode for the old existing cluster events? 


From my understanding. http and stream mode both invoke `Kong.init()` to initialize an independent Kong instance. I think using a different node_id for different Kong instances is a good&simple solution.

vm-001:
@hanshuebner @flrgh Yeah. Using a single file to do `persist` and `load` logic does improve the maintainability. Since the PDK module is required by `global.lua`, I prefer to add a file under `/pdk/private` to do such logic.

kong/plugins/statsd/log.lua
```diff
@@ -12,7 +12,7 @@ local string_format = string.format
 local match = ngx.re.match
 local ipairs = ipairs
 local tonumber = tonumber
-local knode = (kong and kong.node) and kong.node or require "kong.pdk.node".new()
+local knode = kong and kong.node or require "kong.pdk.node".new()
```
hanshuebner, line 15: I don't love the repetition of this line in multiple files.

hanshuebner (approved):
This looks much better, thanks!

kong/pdk/node.lua
```diff
@@ -255,6 +257,21 @@ local function new(self)
...
   end
 
+
+  local prefix = self and self.configuration and self.configuration.prefix
+  if prefix and self.configuration.role == "data_plane" then
+    local id, err = private_node.load_node_id(prefix)
+    if id then
+      node_id = id
+      ngx.log(ngx.INFO, "restored node_id from the filesystem: ", node_id)
+    else
```
chronolaw, line 267: style: blank line before `else`

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,78 @@
...
+local function node_id_filename(prefix)
+  return pl_path.join(prefix, "kong.id")
+end
+
+
+local function initialize_node_id(prefix)
+  if not pl_path.exists(prefix) then
+    local ok, err = pl_dir.makepath(prefix)
+    if not ok then
+      return "failed to create directory " .. prefix .. ": " .. err
```
chronolaw, line 18: It is not a good style, we should always use `nil, err`, or `true`

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,78 @@
...
+local function init_node_id(config)
+  if not config then
+    return
+  end
+
+  if not config.prefix or config.role ~= "data_plane" then
+    return
+  end
+
+  local err = initialize_node_id(config.prefix)
```
chronolaw, line 44: Here we should use `local ok, err = initialize_node_id(config.prefix)`

CHANGELOG.md
```diff
@@ -108,6 +108,8 @@
   [#9337](https://github.com/Kong/kong/pull/9337)
 - Fixed an issue with error-handling and process cleanup in `kong start`.
   [#9337](https://github.com/Kong/kong/pull/9337)
+- Fix data plane node id keeps changing after it restarted.
```
StarlightIbuki, line 111: The wording may confuse readers.
How about:
```suggestion
- Now data planes' node IDs persist after restarting.
```

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,80 @@
...
+    local ok, err = pl_dir.makepath(prefix)
+    if not ok then
+      return false, "failed to create directory " .. prefix .. ": " .. err
+    end
+  end
+
+  local filename = node_id_filename(prefix)
+
+  if not pl_path.exists(filename) then
+    local id = utils.uuid()
```
StarlightIbuki, line 25: It seems that when we have no ID file, we will generate the file, and then read from it.
Could we cache the ID and skip the file reading?

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,80 @@
...
+  if not pl_path.exists(filename) then
+    return nil, "file does not exist: " .. filename
+  end
+
+  local id, read_err = pl_file.read(filename)
+  if read_err then
+    return nil, string.format("failed to access file %s: %s", filename, read_err)
+  end
+
+  if not utils.is_valid_uuid(id) then
```
StarlightIbuki, line 69: Is this necessary?

vm-001: Adding an extra validation will provide more robustness as the file in the filesystem is editable.

kong/pdk/private/node.lua
vm-001: Added. Can you take a look?

StarlightIbuki: Is it safe to cache worker-wise?

CHANGELOG.md
```diff
@@ -108,6 +108,8 @@
   [#9337](https://github.com/Kong/kong/pull/9337)
 - Fixed an issue with error-handling and process cleanup in `kong start`.
   [#9337](https://github.com/Kong/kong/pull/9337)
+- Now data planes' node IDs persist after restarting.
```
mayocream, line 111: Since the current implementation is expected to use random IDs, this PR is a new feature.

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,86 @@
...
+
+local cached_node_id
+
+local function node_id_filename(prefix)
+  return pl_path.join(prefix, "kong.id")
+end
+
+
+local function initialize_node_id(prefix)
+  if not pl_path.exists(prefix) then
```
mayocream, line 16: Is it an option that stores the node ID in the LMDB?

StarlightIbuki (approved):

kong/pdk/private/node.lua
StarlightIbuki: +1. But I'm convinced that this will be a transitional solution and we don't need to spend too much effort on it.

spec/02-integration/09-hybrid_mode/09-node-id-persistence_sepc.lua
```diff
@@ -0,0 +1,77 @@
...
+        local json = assert.response(res).has.jsonbody()
+        assert.equal(1, #json.data)
+        node_id = json.data[1].id
+        return true
+      end, 10, 1)
+
+      helpers.clean_logfile("servroot2/logs/error.log")
+      assert(helpers.restart_kong(data_plane_config))
+
+      ngx.sleep(20) -- sleep because data plane needs to take some time before it connect to CP
```
dndx, line 64: This is bad, please use `wait_until` instead.

kong/cluster_events/init.lua
```diff
@@ -133,7 +131,7 @@ function _M.new(opts)
   end
 
   -- set node id (uuid)
-
+  local knode = kong and kong.node or require "kong.pdk.node".new()
```
dndx, line 134: Let's move this `require` to module level instead.

kong/db/init.lua
```diff
@@ -333,6 +331,7 @@ do
       -- we use the `node.get_id()` from pdk, but in the CLI context, this
       -- value is ephemeral, so no assumptions should be made about the real
       -- owner of a lock
+      local knode = kong and kong.node or require "kong.pdk.node".new()
```
dndx, line 334: Same, let's move this `require` to module level instead.

kong/pdk/node.lua
```diff
@@ -255,6 +257,22 @@ local function new(self)
     return gsub(hostname, "\n$", "")
   end
 
+
+  local prefix = self and self.configuration and self.configuration.prefix
+  if prefix and self.configuration.role == "data_plane" then
+    local id, err = private_node.load_node_id(prefix)
+    if id then
+      node_id = id
+      ngx.log(ngx.INFO, "restored node_id from the filesystem: ", node_id)
```
dndx, line 266: Reduce the verbosity of this log to `DEBUG`, we don't need to print it out every time Kong starts.

kong/pdk/node.lua
```diff
@@ -255,6 +257,22 @@ local function new(self)
...
+
+    else
+      id = _NODE.get_id()
+      if err then
+        ngx.log(ngx.WARN, "failed to restored node_id from the filesystem: " .. err, " (generated a new one " .. id .. ")")
```
dndx, line 271: This line is too long.

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,86 @@
...
+local pl_file = require "pl.file"
+local pl_path = require "pl.path"
+local pl_dir = require "pl.dir"
+
+local ngx = ngx
+
+local cached_node_id
+
+local function node_id_filename(prefix)
+  return pl_path.join(prefix, "kong.id")
```
dndx, line 11: Avoid hard coded path names and ca we call it something like `.node_id` instead?

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,86 @@
...
+
+  local filename = node_id_filename(prefix)
+
+  if not pl_path.exists(filename) then
+    local id = utils.uuid()
+    ngx.log(ngx.INFO, "persisting node id " .. id .. " to filesystem ", filename)
+    local ok, write_err = pl_file.write(filename, id)
+    if not ok then
+      return false, "failed to persist node id to filesystem " .. filename .. ": " .. write_err
+    end
```
dndx, line 31:
```suggestion

    end
```

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,86 @@
...
+    cached_node_id = id
+  end
+
+  return true, nil
```
dndx, line 35:
```suggestion
  return true
```

We don't usually write the `nil` for `err` explicitly.

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,86 @@
...
+
+  local filename = node_id_filename(prefix)
+
+  if not pl_path.exists(filename) then
+    return nil, "file does not exist: " .. filename
+  end
+
+  local id, read_err = pl_file.read(filename)
+  if read_err then
+    return nil, string.format("failed to access file %s: %s", filename, read_err)
```
dndx, line 72: Localize these function calls.

kong/pdk/private/node.lua
vm-001: Actually, LMDB is not available in the init phase.

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,87 @@
...
+local function node_id_filename(prefix)
+  return pl_path.join(prefix, "kong.id")
+end
+
+
+local function initialize_node_id(prefix)
+  if not pl_path.exists(prefix) then
+    local ok, err = pl_dir.makepath(prefix)
+    if not ok then
+      return false, "failed to create directory " .. prefix .. ": " .. err
```
flrgh, line 19: nit: it's convention to return `nil` on failure instead of `false`. The outcome is almost always the same, but it may confuse readers because an explicit `false` is typically used to signal some other kind of result.

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  local filename = node_id_filename(prefix)
+
+  if not pl_path.exists(filename) then
+    local id = utils.uuid()
+    ngx.log(ngx.INFO, "persisting node id " .. id .. " to filesystem ", filename)
+    local ok, write_err = pl_file.write(filename, id)
+    if not ok then
+      return false, "failed to persist node id to filesystem " .. filename .. ": " .. write_err
```
flrgh, line 30: nit: this file has mixed use of string concat (`..`) and `string.format`. I think it would be wise to be consistent

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,87 @@
...
+    if not ok then
+      return false, "failed to create directory " .. prefix .. ": " .. err
+    end
+  end
+
+  local filename = node_id_filename(prefix)
+
+  if not pl_path.exists(filename) then
+    local id = utils.uuid()
+    ngx.log(ngx.INFO, "persisting node id " .. id .. " to filesystem ", filename)
```
flrgh, line 27: 1. no string concat in `ngx.log()` calls
2. (opinion) `INFO` feels too verbose here

```suggestion
    ngx.log(ngx.DEBUG, "persisting node id ", id, " to filesystem ", filename)
```

kong/pdk/node.lua
flrgh, line 271: removed string concat and fixed a typo

```suggestion
        ngx.log(ngx.WARN, "failed to restore node_id from the filesystem: ", err, " (generated a new one: ", id, ")")
```

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,87 @@
...
+
+  if cached_node_id then
+    return cached_node_id, nil
+  end
+
+  local filename = node_id_filename(prefix)
+
+  if not pl_path.exists(filename) then
+    return nil, "file does not exist: " .. filename
+  end
```
flrgh, line 69: I think we should be a little more granular with error-handling in this function.

If it's the first time Kong is starting up, we shouldn't show errors or warnings about the node ID not being present on the filesystem.

There are _other_ scenarios where an operator may need to know about this case--maybe it's not the first startup, but the node ID filename has been deleted by mistake. For this reason I think we should still log _something_ so that an operator can debug if they need to, but we shouldn't necessarily return an error.

```suggestion
  if not pl_path.exists(filename) then
    ngx.log(ngx.DEBUG, "node ID file ", filename, " does not exist)
    return nil, nil
  end
```

flrgh:
1. rebased onto master
2. addressed most of the code review cleanup/consistency feedback
3. added more test coverage

flrgh:
I also addressed @dndx's requests to relocate each `local knode = kong and kong.node or require "kong.pdk.node".new()` expression. The tests are passing with this change having been made, so I think it's okay, but @vm-001 will know better if there are any issues to be aware of in regard to this.

flrgh:
It has tests and is simple enough of a change such that I'm confident any potential issues can be corrected before code freeze. I will merge once the latest CI run has concluded.

@vm-001 thanks for enduring the long road to get this merged in. Can you please give all the review comments another read to see if there are any follow-up items that need another PR? Also, please look at the additional integration tests I've added to make sure that the correct behavior is reflected. I had to mark one test as `pending` because of some order-of-operation oddities, and I wasn't sure if I'd break something else by trying to fix it.

kong/pdk/node.lua
```diff
@@ -255,6 +257,23 @@ local function new(self)
...
+
+  local prefix = self and self.configuration and self.configuration.prefix
+  if prefix and self.configuration.role == "data_plane" then
+    local id, err = private_node.load_node_id(prefix)
+    if id then
+      node_id = id
+      ngx.log(ngx.DEBUG, "restored node_id from the filesystem: ", node_id)
+
+    else
+      id = _NODE.get_id()
```
flrgh, line 269: Should this be `node_id = _NODE.get_id()`?

kong/pdk/node.lua
```diff
@@ -255,6 +257,23 @@ local function new(self)
     return gsub(hostname, "\n$", "")
   end
 
+
+  local prefix = self and self.configuration and self.configuration.prefix
```
flrgh, line 261: I can't help but feel like _all_ of the logic should go into `kong.pdk.private.node` instead. As it is, it's still hard to follow the order of operations to know if there's a chicken/egg situation.

Here's how I picture the call stack:

```
kong.init()
 -> kong.global.init_pdk()
  -> kong.pdk.new()
    -> kong.pdk.node.new()
      -> kong.pdk.private.node.init()
```

...and `kong.pdk.node.private.init()` could look something like this:

```lua
-- init() is a no-op if called more than once
if already_initialized() then
  return
end

-- check shm first, since this might have been a reload
local shm_id = shm:get(MY_ID_KEY)

local fs_id = read_id_from_filesystem()

if shm_id and not fs_id then
  store_to_filesystem(shm_id)

elseif fs_id and not shm_id then
  store_to_shm(fs_id)

elseif not fs_id and not shm_id then
  local id = generate_uuid()
  store_to_filesystem(id)
  store_to_shm(id)

elseif fs_id ~= shm_id then
  -- uh oh, need to pick a winner
  -- probably should keep shm_id in this case because if it exists,
  -- that means we're a reload, so it's already being used by the
  -- current process group
  log("uh oh this is weird...")
  store_to_filesystem(shm_id)

else
  -- no action needed if this is the case
  assert(fs_id == shm_id)
end
```

At this point the `kong.node.get_id()` PDK function just needs to be updated to not attempt to create the node_id itself:

```diff
  function _NODE.get_id()
    if node_id then
      return node_id
    end

    local shm = ngx.shared.kong

-    local ok, err = shm:safe_add(NODE_ID_KEY, utils.uuid())
-    if not ok and err ~= "exists" then
-      error("failed to set 'node_id' in shm: " .. err)
-    end

    node_id, err = shm:get(NODE_ID_KEY)
    if err then
      error("failed to get 'node_id' in shm: " .. err)
    end

    if not node_id then
      error("no 'node_id' set in shm")
    end

    return node_id
  end
```

kong/pdk/node.lua
vm-001: @flrgh The var node_id will be set after call _NODE.get_id().
```lua
function _NODE.get_id()
  -- ...
  node_id, err = shm:get(NODE_ID_KEY)
  -- ...
end
```

vm-001:
> It has tests and is simple enough of a change such that I'm confident any potential issues can be corrected before code freeze. I will merge once the latest CI run has concluded.
> 
> @vm-001 thanks for enduring the long road to get this merged in. Can you please give all the review comments another read to see if there are any follow-up items that need another PR? Also, please look at the additional integration tests I've added to make sure that the correct behavior is reflected. I had to mark one test as `pending` because of some order-of-operation oddities, and I wasn't sure if I'd break something else by trying to fix it.

@flrgh  Thank you for your help. And the fix PR is here: https://github.com/Kong/kong/pull/9790

vm-001:
I appreciate everyone who spent time involved in this PR.

kong/pdk/node.lua
vm-001: @flrgh This is a cleaner approach. But there may have an issue:

```
kong.init()
 -> kong.global.init_pdk(self, kong_config)
  -> kong.pdk.new(kong_config, self)
    -> kong.pdk.node.new(self)
      -> kong.pdk.private.node.init(self)
```

Kong calls `kong.global.init_pdk` in the cmd stage, and the `store_to_shm` function won't work as `ngx.shared` API is not available in the stage.

kong/pdk/private/node.lua
```diff
@@ -0,0 +1,88 @@
...
+
+  if not pl_path.exists(filename) then
+    local id = utils.uuid()
+    ngx.log(ngx.DEBUG, "persisting node_id (", id, ") to ", filename)
+
+    local ok, write_err = pl_file.write(filename, id)
+    if not ok then
+      return nil, fmt("failed to persist node_id to %s: %s", filename, write_err)
+    end
+    cached_node_id = id
```
liverpool8056, line 34: @vm-001 After occasionally checking this part of code, I have a question: the node_id gets cached only when the file doesn't exist, right? Can we cache it even if the file already exists? If so, `load_node_id` will benefit from this.

PR #9069, opened 2022-07-08
feat(rate-limiting) support rate-limiting plugin timezone awareness

### LINKED ISSUE #8894: rate-limiting plugin get_timestamps bug

all version

### Current Behavior

https://github.com/Kong/kong/blob/master/kong/tools/timestamp.lua

**method get timestamp:**
-------------
-- now = timestamp.get_utc()

local function get_timestamps(now)
  local timetable = get_timetable(now)
  local stamps = {}

  timetable.sec = math_floor(timetable.sec)   -- reduce to second precision
  stamps.second = timetable:timestamp() * 1000

  timetable.sec = 0
  stamps.minute = timetable:timestamp() * 1000

  timetable.min = 0
  stamps.hour = timetable:timestamp() * 1000

  timetable.hour = 0
  stamps.day = timetable:timestamp() * 1000

  timetable.day = 1
  stamps.month = timetable:timestamp() * 1000

  timetable.month = 1
  stamps.year = timetable:timestamp() * 1000

  return stamps
end

--------------

**test code:**
  --local current_timestamp = timestamp.get_utc()
  local current_timestamp = time() * 1000
  local periods = timestamp.get_timestamps(current_timestamp)
  for period, period_date in pairs(periods) do
       print(period   .. "---" .. period_date)
  end

**result:**
  "second---1654221091000" --2022-06-03 09:51:31
  "minute---1654221060000"---2022-06-03 09:51:00
  "hour---1654218000000"  -- 2022-06-03 09:00:00
  "day---1654214400000" -- 2022-06-03 08:00:00
  "month---1654041600000"--2022-06-01 08:00:00
  "year---1640995200000"   --2022-01-01 08:00:00

### Expected Behavior

The correct result should be:

  "day---1654214400000" -- 2022-06-03 00:00:00
  "month---1654041600000"--2022-06-01 00:00:00
  "year---1640995200000"   --2022-01-01 00:00:00

A new day starts at 0 o'clock, why is it at 8 o'clock？？

@bungle 

### Steps To Reproduce

eproduce the problem :
1.add a rate-limit plugin to a route ([www.test.com](http://www.test.com/))
2.setting plug-in configuration: policy=local , day=10
3.setting the system Time: date -s "2022-06-03 23:59:00"
4.request site [www.test.com](http://www.test.com/) 5 times
5.waiting system time to 2022-06-04 00:00:00
6.request again, only 5 times remaining

expected result:
sart counting again, 10 remaining times

### Anything else?

_No response_

windmgc:
### Summary

This PR adds support for the rate-limiting plugin to be aware of the local timezone.

Currently, the rate-limiting plugin is working on UTC time, and there have been issues from users complaining about the accuracy problem working in their own timezone(like https://github.com/Kong/kong/issues/8894, and [FTI-4105](https://konghq.atlassian.net/browse/FTI-4105)).

This PR adds an additional param `local_timezone` for the rate-limiting plugin to let the rate-limiting take effect based on the local timezone so that rate-limiting on days/months/years will be based on the local timestamp.

The side-effect of this PR, however, is causing some ambiguity on the SQL generated by rate-limiting cluster policies:

https://github.com/Kong/kong/blob/5d721ac9ae1df36049013599d0253bd39cb6f758/kong/plugins/rate-limiting/policies/cluster.lua#L79-L91

After the change, `TO_TIMESTAMP(%s) AT TIME ZONE 'UTC'` with the period date will not mean that this value is a UTC timestamp, it is actually a local timestamp but we just stored it as a 'fake UTC' timestamp so that it will not be drifted.

Note that this change does not have an influence on the TTL of Postgres records because TTLs are calculated on the Postgres side.

I made this PR a draft to leave some space for checking & commenting so that I can make sure the change is meaningful, ~~and I'll add related tests and do other checks if this PR is considered a proper way to solve the timezone problem.~~

### Full changelog

* Add `get_tz_time` function(can be used to get milliseconds since unix epoch in the given timezone) in `kong.tools.timestamp`
* Add `local_timezone` param for `rate-limiting` plugin

### Issue reference

Fix #8894 and [FTI-4105](https://konghq.atlassian.net/browse/FTI-4105).

### DIFF
--- a/kong/plugins/rate-limiting/schema.lua
+++ b/kong/plugins/rate-limiting/schema.lua
@@ -91,6 +91,7 @@ return {
           { redis_timeout = { type = "number", default = 2000, }, },
           { redis_database = { type = "integer", default = 0 }, },
           { hide_client_headers = { type = "boolean", required = true, default = false }, },
+          { local_timezone = { type = "boolean", required = false, default = false }, },
         },
         custom_validator = validate_periods_order,
       },
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -118,7 +118,13 @@ end
 
 
 function RateLimitingHandler:access(conf)
-  local current_timestamp = time() * 1000
+  local local_timezone = conf.local_timezone
+  local current_timestamp
+  if local_timezone then
+    current_timestamp = timestamp.get_tz_time()
+  else
+    current_timestamp = time() * 1000
+  end
 
   -- Consumer is identified by ip address or authenticated_credential id
   local identifier = get_identifier(conf)
--- a/kong/tools/timestamp.lua
+++ b/kong/tools/timestamp.lua
@@ -8,8 +8,11 @@ local luatz = require "luatz"
 local tz_time = luatz.time
 local tt_from_timestamp = luatz.timetable.new_from_timestamp
 local tt = luatz.timetable.new
+local get_tzinfo = luatz.get_tz
 local math_floor = math.floor
 local tablex = require "pl.tablex"
+local ngx_log = ngx.log
+local WARN = ngx.WARN
 
 --- Current UTC time
 -- @return UTC time in milliseconds since epoch, but with SECOND precision.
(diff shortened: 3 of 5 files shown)


kong/plugins/rate-limiting/schema.lua
```diff
@@ -91,6 +91,7 @@ return {
           { redis_timeout = { type = "number", default = 2000, }, },
           { redis_database = { type = "integer", default = 0 }, },
           { hide_client_headers = { type = "boolean", required = true, default = false }, },
+          { local_timezone = { type = "boolean", required = false, default = false }, },
```
hbagdi, line 94: What are the downsides of not making this behavior configurable?

windmgc: If it is made a default behaviour then it will be some kind of a breaking change, which may lead to confusion

hbagdi: @bungle Thoughts?

windmgc:
This is a nice-to-have feature for rate-limiting plugins, but it may require more work to be done, the feature itself relies on the existence of the timezone database(in most Linux distributions, the tzdata package) on the OS.
Considering that in most cases Kong will be deployed in a container environment which usually lacks timezone database, tzdata must be added as a prerequisite at least for all of our official images.

Tieske:
The proper solution would be to have the user specify the timezone. If I'm in Amsterdam, and have Kong nodes in both `eu-central-1` (Frankfurt, Germany, UTC+1) , and some in `eu-west-1` (Dublin, Ireland, UTC), then it becomes a weird mix. So the user must specify for the entire cluster what timezone to use.

Besides that it would mean that you also need to take DST into account. Which is another rabbit hole; some days will have 25 hours, some will have 23 hours. And on the switch, some hours will be 0 minutes, and some will be 120 minutes).

This is a very hard problem to solve. The timezone database is yet another nuisance in releasing.

My 2cts: if done at all, then the user must specify the timezone, and it should NOT honor DST. So the day/month-roll-over will still be at 00:00, or 01:00 depending on DST.

hbagdi:
Closing this because no decision has been made. Please re-open if needed.

PR #9071, opened 2022-07-08
feat(clustering) switch turn off wRPC also from CP

Suika-Kong:
and emit a warning

Fix FT-3077

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -1570,9 +1570,6 @@ end
 
 
 function Kong.serve_wrpc_listener(options)
-  if kong.configuration.legacy_hybrid_protocol then
-    return ngx.exit(404)
-  end
   log_init_worker_errors()
 
   ngx.ctx.KONG_PHASE = PHASES.cluster_listener
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -1570,6 +1570,9 @@ end
 
 
 function Kong.serve_wrpc_listener(options)
+  if kong.configuration.legacy_hybrid_protocol then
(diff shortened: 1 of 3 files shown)


Suika-Kong:
remember to update this: https://github.com/Kong/kong/pull/9061

kong/init.lua
```diff
@@ -1570,6 +1570,9 @@ end
 
 
 function Kong.serve_wrpc_listener(options)
+  if kong.configuration.legacy_hybrid_protocol then
+    return ngx.exit(404)
```
fffonion, line 1574: instead of return 404 here, we can also skip rendering the wrpc server block in nginx_template

fffonion:
let's also use this option in wrpc test so that 1) we know it works 2) we know when it got removed

Suika-Kong:
> let's also use this option in wrpc test so that 1) we know it works 2) we know when it got removed

https://github.com/Kong/kong/pull/9061

fffonion (approved):

kong/templates/nginx_kong.lua
```diff
@@ -439,11 +439,14 @@ server {
         }
     }
 
+> if legacy_hybrid_protocol == "off" then
```
chronolaw, line 442: Here we need not to compare to `off`, it is a boolean, just use `if legacy_hybrid_protocol`.

You can refer to `legacy_worker_events`

Suika-Kong: This actually causes a bug. Here's the fix: #9085

PR #9073, opened 2022-07-08
perf(clustering) improve calculate hash performance by utilizing string buffer and tablepool

bungle:
### Summary

Improve calculate hash performance by utilizing string buffer and tablepool.

### DIFF
--- a/kong/clustering/config_helper.lua
+++ b/kong/clustering/config_helper.lua
@@ -1,5 +1,11 @@
 local constants = require("kong.constants")
 local declarative = require("kong.db.declarative")
+local tablepool = require("tablepool")
+local isempty = require("table.isempty")
+local isarray = require("table.isarray")
+local nkeys = require("table.nkeys")
+local buffer = require("string.buffer")
+
 
 local tostring = tostring
 local assert = assert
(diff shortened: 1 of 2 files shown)


kong/clustering/config_helper.lua
```diff
@@ -1,132 +1,147 @@
...
-      local o = new_tab(count < 100 and count or 100, 0)
-      for j = 1, count do
-        i = i + 1
-        o[i] = to_sorted_string(value[j])
+  if value == ngx_null then
+    o:put("/null/")
+
+  else
+    local t = type(value)
+    if t == "string" or t == "number" then
```
fffonion, line 50: since we are using string buffer, i think we can also just skip type assertions and let it do the job. previously we do this because {} could be from a table or from a string "{}". but they are already encoded differently in string.buffer.

bungle: @fffonion only `string`, `number` and `object` with `__tostring` are allowed with `:put`

bungle:
A small improvement (so far):
```-rw-r--r--  1 bungle staff 395M Dec 15  2021 kong.json```

```reading file...
reading file done in 0.16999983787537 secs
decoding json...
decoding json done in 2.0450000762939 secs
calculating hash...
c5c725a9b2776f947c198379a54a65d6
calculating hash done in 2.8819999694824 secs
calculating hash new...
57bfa7fc4fa1ec577f3b0a464dd6e96a
calculating hash done in 2.2009999752045 secs```

bungle:
Small test:

```Inserted 200000 services in 4 seconds
Inserted 200000 routes in 5 seconds
Inserted 200000 consumers in 4 seconds
Inserted 200000 upstreams in 4 seconds
Inserted 200000 targets in 5 seconds
Inserted 200000 plugins in 5 seconds
Inserted 200000 services in 4 seconds
Inserted 200000 upstreams in 4 seconds```

```2022/07/08 16:50:09 [error] 99950#0: *7 [lua] wrpc_control_plane.lua:111: took 18718 ms to export config, context: ngx.timer```

```2022/07/08 16:50:19 [error] 99950#0: *7 [lua] wrpc_control_plane.lua:125: took 10302 ms to calculate hash, context: ngx.timer```

mayocream (approved):

PR #9076, opened 2022-07-11
fix(clustering) event for config sync is generated 2 times

Suika-Kong:
This problem is because we register events 2 times, each for 1 implementation of a control plane.

I also refactored a little bit. They cannot share 1 copy of the config cache because they have different handling of the message.

For wRPC, it's `protobuf(gzip(json_encode(config)) + other_fields)`, and for webSocket it's `gzip(json_encode(config + other_fields))`.

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -6,6 +6,7 @@ local pl_file = require("pl.file")
 local pl_tablex = require("pl.tablex")
 local ssl = require("ngx.ssl")
 local openssl_x509 = require("resty.openssl.x509")
+local config_service = require("kong.clustering.services.config")
 local ngx_log = ngx.log
 local assert = assert
 local sort = table.sort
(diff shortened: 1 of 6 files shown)


kong/clustering/init.lua
```diff
@@ -61,9 +62,10 @@ function _M:handle_wrpc_websocket()
 end
 
 function _M:init_cp_worker(plugins_list)
-  self.json_handler:init_worker(plugins_list)
+  config_service.init_worker()
+  config_service.init_control_plane(self.json_handler, plugins_list)
   if not kong.configuration.legacy_hybrid_protocol then
```
chronolaw, line 67: I prefer this style:
`if leagcy_hybrid_protocol then return end`

Suika-Kong:
This is redundant to #9082.
As we will refactor CP/DP later, I prefer #9082. So this PR is closed.

PR #9077, opened 2022-07-11
feat(acme): debug log details

zekth:
### Summary

Currently the FTI-3300 has not landed and we're using the previous implementation, monitoring the use of it using debug logs. But we're not able to have the granularity of failed domains registration. With this log we would be able to have more insights.

Note: suggesting changing `whitelist` to `allowlist` too.

### Full changelog

* Add host in the debug log for non allowed domain

### Issue reference

Following https://github.com/Kong/kong/pull/9047

### DIFF
--- a/kong/plugins/acme/handler.lua
+++ b/kong/plugins/acme/handler.lua
@@ -105,7 +105,7 @@ function ACMEHandler:certificate(conf)
   host = string.lower(host)
 
   if not check_domains(conf, host) then
-    kong.log.debug("ignoring because domain is not in whitelist")
+    kong.log.debug("ignoring because domain is not in whitelist:", host)
     return
   end
 
--- a/kong/plugins/acme/handler.lua
+++ b/kong/plugins/acme/handler.lua
@@ -105,7 +105,7 @@ function ACMEHandler:certificate(conf)
   host = string.lower(host)
 
   if not check_domains(conf, host) then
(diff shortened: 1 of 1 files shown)


kong/plugins/acme/handler.lua
```diff
@@ -105,7 +105,7 @@ function ACMEHandler:certificate(conf)
   host = string.lower(host)
 
   if not check_domains(conf, host) then
-    kong.log.debug("ignoring because domain is not in whitelist")
+    kong.log.debug("ignoring because domain is not in whitelist:", host)
```
voigtjr, line 108: Do you think we could make this `info` level as well? Feels like a useful operational message, and I suspect we'll drop debug messages at some point in the near future

aboudreault:
@zekth when merged, make sure to cherry-pick and open a PR for EE 2.8.x.x.

kong/plugins/acme/handler.lua
Tieske, line 108:
```suggestion
    kong.log.debug("ignoring because domain is not in allowed-list: ", host)
```

Tieske (approved):

kong/plugins/acme/handler.lua
fffonion: I agree with @Tieske here.

PR #9078, opened 2022-07-11
chore(declarative) on the fly migration

Suika-Kong:
Also bumping _format_version to 3.0.

Fix FT-3098


Suika-Kong:
The migration should be done before validation.

Suika-Kong:
Remember to downgrade the version number for 3.0/2.8 CP/DP

kong/templates/nginx_kong.lua
```diff
@@ -439,7 +439,7 @@ server {
         }
     }
 
-> if legacy_hybrid_protocol == "off" then
+> if not legacy_hybrid_protocol then
```
fffonion, line 442: seems unrelated commits, needs a rebase?

Suika-Kong: done

kong/db/declarative/migrations/init.lua
```diff
@@ -0,0 +1,12 @@
+local regex_route_path = require "kong.db.declarative.migrations.regex_route_path"
+
+return function (tbl)
+    if not tbl then
```
mayocream, line 4: Let's use 2 spaces ident to align with other files.

mayocream (commented):
Let's add a test to make sure the 2.1 to 3.0 declarative yaml upgrading works.

mayocream (approved):

kong/db/migrations/migrate_regex_280_300.lua
```diff
@@ -0,0 +1,86 @@
...
+  return (re_find(path, [[[a-zA-Z0-9\.\-_~/%]*$]], "ajo"))
+end
+
+local function migrate_regex(reg)
+  if is_not_regex(reg) then
+    return reg, false
+  end
+
+  local migrated = "~" .. normalize_regex(reg)
+  return migrated, migrated ~= reg
```
fffonion, line 83: seems second return value is always true?
```suggestion
  return migrated, true
```

spec/01-unit/01-db/01-schema/11-declarative_config/04-on-the-fly-migration_spec.lua
```diff
@@ -0,0 +1,119 @@
...
+  table.sort(tbl)
+  return tbl
+end
+
+
+describe("declarative config: on the fly migration", function()
+  it("routes handling", function()
+    local dc = assert(declarative.new_config(conf_loader()))
+    local config = [[
+      _format_version: "1.1"
```
fffonion, line 76: make sure `2.1` is covered in tests, we can simply make it a loop and only change the version number

kong/db/migrations/migrate_regex_280_300.lua
Suika-Kong: Ah... that's a leftover of refactoring.

fffonion (approved):

PR #9081, opened 2022-07-11
Make route migration work for routes without path

hanshuebner:
While working on upgrade testing, I noticed that the new route migration for routes crashed for routes that have no `paths` configured.


Suika-Kong (commented):
We should do the same to `c_normalize_regex_path` (and this function should be renamed to `c_migrate_regex_path`)

bungle (approved):

Suika-Kong:
See #9091 .

mayocream (approved):

Suika-Kong:
Already covered by https://github.com/Kong/kong/pull/9064/

PR #9082, opened 2022-07-11
fix(clustering) do not register events multiple times

bungle:
### Summary

This reduces double event handling and event publishing on clustering.

On further PRs we can reduce double exporting, json decoding, calculate hashing etc. that are common between wrpc control plane and json control plane, but here I just fixed the double event handling.

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -9,6 +9,8 @@ local openssl_x509 = require("resty.openssl.x509")
 local ngx_log = ngx.log
 local assert = assert
 local sort = table.sort
+local type = type
+
 
 local check_protocol_support =
   require("kong.clustering.utils").check_protocol_support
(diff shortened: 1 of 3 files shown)


kong/clustering/init.lua
```diff
@@ -22,6 +24,37 @@ local ngx_WARN = ngx.WARN
 local _log_prefix = "[clustering] "
 
 
+-- Sends "clustering", "push_config" to all workers in the same node, including self
```
Suika-Kong, line 27: Could we put those config sync related codes into a separate file?

Suika-Kong (approved):

kong/clustering/init.lua
bungle: Will be done on following refactoring PR.

PR #9086, opened 2022-07-12
fix(clustering) wrpc crash for extra argument

Suika-Kong:


Suika-Kong:
will just fix at #9062

PR #9095, opened 2022-07-14
refactor(router) use tablepool to improve router

chronolaw:
### Summary

Inspired by #9036, I introduced `tablepool` in atc router.
I think it may be some help for performance.

### Full changelog

* use `tablepool` to reduce small table object usage

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -221,14 +221,14 @@ local function get_atc(route)
       end
 
       tb_insert(headers, "(" .. tb_concat(single_header, " || ") .. ")")
-      release_table(CTX_NS, single_header)
+      release_table(ATC_NS, single_header)
     end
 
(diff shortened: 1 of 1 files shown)


Suika-Kong (commented):
Instead of using a pool, we could just reuse the same table all the time.

chronolaw:
> Instead of using a pool, we could just reuse the same table all the time.

Yes, it is true, but we still need to clear items in the reused table. `tablepool` did these for us.

fffonion:
I agree with @Suika-Kong , seems some long-lived tables is cleaner, we can just do `table.clear` instead of `release_table`, how do you think @chronolaw ?

chronolaw:
> I agree with @Suika-Kong , seems some long-lived tables is cleaner, we can just do `table.clear` instead of `release_table`, how do you think @chronolaw ?

OK, I will reimplement this feature.

dndx (approved):

PR #9096, opened 2022-07-14
chore(atc_router) add LRU cache support just like the traditional router

dndx:
This code is copied from the traditional router, we just keep the same logic/behavior as old router. 

Maybe we will change it in the future.

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -570,8 +570,8 @@ function _M:exec(ctx)
       local name = name:gsub("-", "_"):lower()
 
       if type(value) == "table" then
-        for i = 1, #value do
-          value[i] = value[i]:lower()
+        for i, v in ipairs(value) do
+          value[i] = v:lower()
         end
         tb_sort(value)
         value = tb_concat(value, ", ")
@@ -584,10 +584,10 @@ function _M:exec(ctx)
         headers_key = { "|", name, "=", value }
 
       else
-        headers_key[headers_count+1] = "|"
-        headers_key[headers_count+2] = name
-        headers_key[headers_count+3] = "="
-        headers_key[headers_count+4] = value
+        headers_key[headers_count + 1] = "|"
+        headers_key[headers_count + 2] = name
+        headers_key[headers_count + 3] = "="
+        headers_key[headers_count + 4] = value
       end
 
       headers_count = headers_count + 4
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -6,6 +6,7 @@ local router = require("resty.router.router")
 local context = require("resty.router.context")
 local constants = require("kong.constants")
 local bit = require("bit")
+local lrucache = require("resty.lrucache")
 local ffi = require("ffi")
 local server_name = require("ngx.ssl").server_name
 local normalize = require("kong.tools.uri").normalize
@@ -16,6 +17,7 @@ local tb_nkeys = require("table.nkeys")
 local ngx = ngx
 local tb_concat = table.concat
 local tb_insert = table.insert
+local tb_sort = table.sort
 local find = string.find
 local byte = string.byte
 local sub = string.sub
@@ -38,6 +40,19 @@ local ngx_WARN = ngx.WARN
 local SLASH         = byte("/")
 local MAX_HEADER_COUNT = 255
 local MAX_REQ_HEADERS = 100
+
+
+--[[
+Hypothesis
+----------
+
+Item size:        1024 bytes
+Max memory limit: 5 MiBs
+
+LRU size must be: (5 * 2^20) / 1024 = 5120
+Floored: 5000 items should be a good default
+--]]
+local MATCH_LRUCACHE_SIZE = 5e3
 
 
 local function is_regex_magic(path)
@@ -289,6 +304,8 @@ function _M.new(routes)
     routes = {},
     services = {},
     fields = {},
+    cache = lrucache.new(MATCH_LRUCACHE_SIZE),
+    cache_neg = lrucache.new(MATCH_LRUCACHE_SIZE),
   }, _MT)
 
   for _, r in ipairs(routes) do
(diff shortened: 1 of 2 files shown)


kong/router/atc_compat.lua
```diff
@@ -547,13 +564,62 @@ function _M:exec(ctx)
...
+
+      else
+        value = value:lower()
+      end
+
+      if headers_count == 0 then
+        headers_key = { "|", name, "=", value }
+
+      else
+        headers_key[headers_count+1] = "|"
```
chronolaw, line 587: style issue: needs space ` + `

chronolaw, line 587: should add a init line `headers_key = {}`?

kong/router/atc_compat.lua
```diff
@@ -547,13 +564,62 @@ function _M:exec(ctx)
...
+  -- cache lookup
+
+  local cache_key = (req_method or "") .. "|" .. (req_uri or "") ..
+                    "|" .. (req_host or "") .. "|" .. (sni or "") ..
+                    headers_key
+  local match_t = self.cache:get(cache_key)
   if not match_t then
-    return
+    if self.cache_neg:get(cache_key) then
+      return nil
```
chronolaw, line 607: I think `return` is enough, because below code also did so.

kong/router/atc_compat.lua
```diff
@@ -547,13 +564,62 @@ function _M:exec(ctx)
...
-  local match_t = self:select(req_method, req_uri, req_host, req_scheme,
-                              nil, nil, nil, nil,
-                              sni, headers)
+  local headers_key do
+    local headers_count = 0
+    for name, value in pairs(headers) do
+      local name = name:gsub("-", "_"):lower()
+
+      if type(value) == "table" then
+        for i = 1, #value do
```
chronolaw, line 573: `ipairs` may be better?

kong/router/atc_compat.lua
```diff
@@ -547,13 +564,62 @@ function _M:exec(ctx)
...
+          value[i] = value[i]:lower()
+        end
+        tb_sort(value)
+        value = tb_concat(value, ", ")
```
chronolaw, line 577: For hash key usage, I think it is unnecessary to concat with `, `, maybe a simple empty string?

kong/router/atc_compat.lua
```diff
@@ -547,13 +564,62 @@ function _M:exec(ctx)
...
+
+      headers_count = headers_count + 4
+    end
+
+    headers_key = headers_key and tb_concat(headers_key, nil, 1, headers_count) or ""
+  end
+
+  -- cache lookup
+
+  local cache_key = (req_method or "") .. "|" .. (req_uri or "") ..
```
chronolaw, line 601: How about crc32 or md5bin? we can avoid string cost.

kong/router/atc_compat.lua
dndx: I want to be a bit more explicit, we are actually returning only `nil` to indicate a not match condition. It is better style IMO.

kong/router/atc_compat.lua
dndx: This was from the old router, it makes debugging a bit easier.

kong/router/atc_compat.lua
dndx: It will probably be more expensive to call these functions than simply concat the key. The only disadvantage of concat the keys directly is potentially long key size, but it should be faster than calling multiple `sha:update()`s.

kong/router/atc_compat.lua
```diff
@@ -547,13 +564,62 @@ function _M:exec(ctx)
...
+    if self.cache_neg:get(cache_key) then
+      return nil
+    end
+
+    match_t = self:select(req_method, req_uri, req_host, req_scheme,
+                          nil, nil, nil, nil,
+                          sni, headers)
+    if not match_t then
+      self.cache_neg:set(cache_key, true)
+      return
```
chronolaw, line 615: Here, `return nil` to emphasize?

kong/router/atc_compat.lua
```diff
@@ -547,13 +564,62 @@ function _M:exec(ctx)
...
-    return
+    if self.cache_neg:get(cache_key) then
+      return nil
+    end
+
+    match_t = self:select(req_method, req_uri, req_host, req_scheme,
+                          nil, nil, nil, nil,
+                          sni, headers)
+    if not match_t then
+      self.cache_neg:set(cache_key, true)
```
chronolaw, line 614: How to add a ttl in `cache_neg`? maybe in some config update it will match.

dndx: This logic was copied from the old router so currently TTL is not needed because it is always rebuilt.

chronolaw (approved):

kong/router/atc_compat.lua
```diff
@@ -289,6 +304,8 @@ function _M.new(routes)
     routes = {},
     services = {},
     fields = {},
+    cache = lrucache.new(MATCH_LRUCACHE_SIZE),
```
chronolaw, line 307: It will be called in `runloop/handler.lua`, and will pass two cache params, could it be compatible with this call `Router.new(routes, router_cache, router_cache_neg)`?

chronolaw: I also create a PR #9148, please refer to it.

kong/router/atc_compat.lua
```diff
@@ -547,13 +572,62 @@ function _M:exec(ctx)
...
+
+      headers_count = headers_count + 4
+    end
+
+    headers_key = headers_key and tb_concat(headers_key, nil, 1, headers_count) or ""
+  end
+
+  -- cache lookup
+
+  local cache_key = (req_method or "") .. "|" .. (req_uri or "") ..
```
fffonion, line 609: do we care about req_scheme (or is it already part of req_uri)?

chronolaw: This code is copied from the traditional router, we just keep the same logic/behavior. Maybe we will change it in the future.

PR #9100, opened 2022-07-14
chore(requirements) add ATC_ROUTER

aboudreault:
to fix https://internal.builds.konghq.com/blue/organizations/jenkins/kong-build-tools/detail/PR-507/2/pipeline/124#step-159-log-2041


hanshuebner (approved):

gszr (approved):

spec/02-integration/03-db/11-postgres-ro_spec.lua
```diff
@@ -144,9 +144,8 @@ for _, strategy in helpers.each_strategy() do
 
         helpers.wait_until(function ()
           return pcall(function ()
-            assert.logfile().has.line("get_updated_router(): could not rebuild router: " ..
-                                      "could not load routes: [postgres] connection " ..
-                                      "refused (stale router will be used)", true)
+            assert.logfile().has.line("could not rebuild router.*: could not load routes.*" ..
+                                      "postgres.*connection refused", false)
```
aboudreault, line 148: this change seems related to the new atc-router. now, we don't get exactly the same error about the RO pg error. we are not getting:
https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L1125
instead of:
https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L838


@dndx could just double check if this is expected?

PR #9114, opened 2022-07-18
fix(cmd) confusing prompt

Suika-Kong:
It says `[Y/n]`, but doesn't take Y as the default choice.


chronolaw (approved):

chronolaw:
Yes, normally `Y` means the default choice, but the code shows it does not deal with empty input.

PR #9118, opened 2022-07-18
fix(templates) use `atc_compatiable` as default for `router_flavor`

dndx:
* enable `traditional_compatible` in kong_defaults.lua
* replace regex `?<` to `?P<` to work with Rust regex
* sort `route.paths`, move regex uri to the front
* use `table.new` to pre-allocate table object

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -12,6 +12,7 @@ local server_name = require("ngx.ssl").server_name
 local normalize = require("kong.tools.uri").normalize
 local hostname_type = require("kong.tools.utils").hostname_type
 local tb_nkeys = require("table.nkeys")
+local tb_new = require("table.new")
 
 
 local ngx = ngx
@@ -325,15 +326,9 @@ function _M.new(routes, cache, cache_neg)
     cache_neg = lrucache.new(MATCH_LRUCACHE_SIZE)
   end
 
-  local router = setmetatable({
-    schema = s,
-    router = inst,
-    routes = {},
-    services = {},
-    fields = {},
-    cache = cache,
-    cache_neg = cache_neg,
-  }, _MT)
+  local routes_n   = #routes
+  local routes_t   = tb_new(0, routes_n)
+  local services_t = tb_new(0, routes_n)
 
   local is_traditional_compatible =
           kong and kong.configuration and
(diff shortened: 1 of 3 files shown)


kong/router/atc_compat.lua
```diff
@@ -107,10 +107,10 @@ local OP_REGEX = "~"
 local function get_atc(route)
   local out = {}
 
-  local gen = gen_for_field("net.protocol", OP_EQUAL, route.protocols)
-  if gen then
-    tb_insert(out, gen)
-  end
+  --local gen = gen_for_field("net.protocol", OP_EQUAL, route.protocols)
```
chronolaw, line 110: I don't understand this code. Why don't we need "net.protocol"?

chronolaw: Maybe atc only works in http subsystem, so we can skip `net.protocol` op?

chronolaw:
Maybe we should add some test cases for regex path and captures later

chronolaw (approved):

kong/router/atc_compat.lua
```diff
@@ -107,10 +120,10 @@ local OP_REGEX = "~"
 local function get_atc(route)
```
chronolaw, line 120: if all router's params (hosts/headers/...) are nil, then thsis function will return an empty string.

Could we deal with this situation in act-router?

esatterwhite:
@fffonion @chronolaw What part of this is using rust. We recently updated to 3.x and started seeing some occasional rust panics

```nginx
Dec 6 09:21:01 ingress-kong-54d457f949-fhn6b kong-proxy thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 18, error_len: Some(1) }', src/ffi.rs:33:22
Dec 6 09:21:01 ingress-kong-54d457f949-fhn6b kong-proxy note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Dec 6 09:21:01 ingress-kong-54d457f949-fhn6b kong-proxy fatal fatal runtime error: failed to initiate panic, error 3
Dec 6 09:21:01 ingress-kong-54d457f949-fhn6b kong-proxy error [lua] worker.lua:248: communicate(): event worker failed: failed to receive the first 2 bytes: closed, context: ngx.timer 
```

chronolaw:
This is the new router built with Rust. https://github.com/Kong/atc-router

If you get a problem, please open a ticket and we can discuss there.

PR #9120, opened 2022-07-19
fix(wrpc) error handling when sending message

Suika-Kong:
### Summary

We miss handling of message sending error (over WebSocket), and the future times out when an error happens.

This problem is found when I tried to send a payload larger than `payload_limit`.


spec/kong_tests.conf
```diff
@@ -15,6 +15,7 @@ pg_host = 127.0.0.1
 pg_port = 5432
 pg_timeout = 10000
 pg_database = kong_tests
+pg_password = kong_tests
```
fffonion, line 18:
```suggestion
```

fffonion (approved):

PR #9121, opened 2022-07-19
fix(clustering) potential leaking for DP wRPC

Suika-Kong:
### Summary

It may leak when an error happens, and we recreate the wRPC connection. The previous one is not closed (instead, the WebSocket connection is closed. Probably a typo.)

### DIFF
--- a/kong/clustering/wrpc_data_plane.lua
+++ b/kong/clustering/wrpc_data_plane.lua
@@ -93,6 +93,11 @@ local function get_services()
   return wrpc_services
 end
 
+-- we should have only 1 dp peer at a time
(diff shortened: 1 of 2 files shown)


Suika-Kong:
Coincidentally it won't cause problems in most cases.
There are 2 threads, one for WebSocket listening and one waiting (on a queue) for the user to call RPC.
The 1 thread encounters errors if the WebSocket connection fails, and it sets the closing flag and exits. The other thread exits when seeing the flag.

This problem occurs when I wrote a unit test, with a lot of pairs of connections (and not close them).

PR #9129, opened 2022-07-21
fix(aws-lambda) fix 500 error in serviceless route

attenuation:
### Summary

fix plugin `aws-lambda` 500 error in serviceless route and config enable awsgateway_compatible

### Full changelog

* fix plugin `aws-lambda` 500 in serviceless route and config enable awsgateway_compatible
* add serviceless `db.route` spec to cover this case

### DIFF
--- a/kong/plugins/aws-lambda/aws-serializer.lua
+++ b/kong/plugins/aws-lambda/aws-serializer.lua
@@ -78,7 +78,8 @@ return function(ctx, config)
   end
 
   -- prepare path
-  local path = var.upstream_uri:match("^([^%?]+)")  -- strip any query args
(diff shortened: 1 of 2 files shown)


kong/plugins/aws-lambda/aws-serializer.lua
```diff
@@ -78,7 +78,8 @@ return function(ctx, config)
   end
 
   -- prepare path
-  local path = var.upstream_uri:match("^([^%?]+)")  -- strip any query args
+  local uri = var.upstream_uri and var.upstream_uri or var.request_uri
```
mayocream, line 81:
```suggestion
  local uri = var.upstream_uri or var.request_uri
```

fffonion (approved):

PR #9130, opened 2022-07-21
docs(helpers) add docs for function `wait_timer`

ADD-SP:
截屏2022-07-22 14 16 22

### DIFF
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -1392,13 +1392,27 @@ local function wait_until(f, timeout, step)
 end
 
 
----wait for some timer
----@param timer_name_pattern string
----@param plain boolean
----@param mode string all-finish | all-running | any-finish | any-running | worker-wide-all-finish
----@param timeout? number optional, maximum time to wait (default = 2)
----@param admin_client_timeout? number optional, to override the default timeout setting
----@param forced_admin_port? number optional, to override the default port of admin API
+--- Wait for some timers, throws an error on timeout.
+-- 
+-- NOTE: this is a regular Lua function, not a Luassert assertion.
+-- @function wait_timer
+-- @tparam string timer_name_pattern the call will apply to all timers matching this string
+-- @tparam boolean plain if truthy, the `timer_name_pattern` will be matched plain, so without pattern matching
+-- @tparam string mode one of: "all-finish", "all-running", "any-finish", "any-running", or "worker-wide-all-finish"
+-- 
+-- any-finish: At least one of the timers that were matched finished
+-- 
+-- all-finish: All timers that were matched finished
+-- 
+-- any-running: At least one of the timers that were matched is running
+-- 
+-- all-running: All timers that were matched are running
+-- 
+-- worker-wide-all-finish: All the timers in the worker that were matched finished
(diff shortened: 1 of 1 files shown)


Tieske (changes requested):
This doesn't even show up in the documentation? Can you try and run `ldoc` and verify?

spec/helpers.lua
```diff
@@ -1388,13 +1388,17 @@ local function wait_until(f, timeout, step)
...
+--- 
+--- NOTE: this is a regular Lua function, not a Luassert assertion.
+---@function wait_timer
 ---@param timer_name_pattern string
 ---@param plain boolean
 ---@param mode string all-finish | all-running | any-finish | any-running | worker-wide-all-finish
 ---@param timeout? number optional, maximum time to wait (default = 2)
 ---@param admin_client_timeout? number optional, to override the default timeout setting
 ---@param forced_admin_port? number optional, to override the default port of admin API
+---@usage helpers.wait_timer("rate-limiting", true, "all-finish", 10)
```
Tieske, line 1401:
```suggestion
--- Wait for some timers, throws an error on timeout.
-- ?? what does it do? describe the "mode" parameter
-- @function wait_timer
-- @tparam string timer_name_pattern the call will apply to all timers matching
-- this string
-- @tparam boolean plain if truthy, the `timer_name_pattern` will be matched
-- plain, so without pattern matching
-- @tparam string mode one of: "all-finish", "all-running", "any-finish",
-- "any-running", or "worker-wide-all-finish"
-- @tparam[opt=2] number timeout maximum time to wait
-- @tparam[opt] number admin_client_timeout to override the default timeout setting
-- @tparam[opt] number forced_admin_port to override the default port of admin API
-- @return ????
-- @usage helpers.wait_timer("rate-limiting", true, "all-finish", 10)
```

This should work, but is incomplete. Can you check the `@return` tag and the description?

ADD-SP:
> This doesn't even show up in the documentation? Can you try and run `ldoc` and verify?

截屏2022-07-21 18 29 02

Tieske:
@ADD-SP yep,, my bad, it did show up. But is has the wrong annotations; '?' and '|' characters that are not recognized. My suggestion above takes care of that.

spec/helpers.lua
```diff
@@ -1388,13 +1388,27 @@ local function wait_until(f, timeout, step)
...
+--- 
+---any-running: At least one of the timers that were matched is running
+--- 
+---all-running: All timers that were matched are running
+--- 
+---worker-wide-all-finish: All the timers in the worker that were matched finished
+---@param[opt=2] timeout number optional, maximum time to wait
+---@param[opt] admin_client_timeout number optional, to override the default timeout setting
+---@param[opt] forced_admin_port number optional, to override the default port of admin API
+---@usage helpers.wait_timer("rate-limiting", true, "all-finish", 10)
```
Tieske, line 1411: can we please make this consistent with the rest of the file? triple-dash + space on the first line, double-dash + space on the following lines

Tieske (changes requested):

Tieske (approved):

PR #9133, opened 2022-07-22
fix(*) router now has different path scheme

Suika-Kong:
This scheme change was not done right and causes a lot of tests to fail.

And, `typedef.path` is also used in places like service.

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -304,19 +304,6 @@ typedefs.path = Schema.define {
   custom_validator = validate_path,
 }
 
-typedefs.router_path = Schema.define {
-  type = "string",
-  match_any = {
-    patterns = {"^/", "^~/"},
-    err = "should start with: / (fixed path) or ~/ (regex path)",
-  },
-  match_none = {
-    { pattern = "//",
-      err = "must not have empty segments"
-    },
-  },
-  custom_validator = validate_path,
-}
 
 typedefs.url = Schema.define {
   type = "string",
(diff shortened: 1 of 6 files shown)


kong/db/schema/typedefs.lua
```diff
@@ -529,7 +540,7 @@ typedefs.no_hosts = Schema.define(typedefs.hosts { eq = null })
 
 typedefs.paths = Schema.define {
   type = "array",
-  elements = typedefs.path {
+  elements = typedefs.router_path {
```
Tieske, line 543: should we also duplicate `typedef.paths` to `typedef.router_paths` ? to be safe?

also the check below; the `match_none` is a duplicate, since it is already in the typedef of the individual elements

kong/db/schema/typedefs.lua
```diff
@@ -294,6 +294,17 @@ typedefs.port = Schema.define {
...
+  starts_with = "/",
+  match_none = {
+    { pattern = "//",
+      err = "must not have empty segments"
+    },
+  },
+  custom_validator = validate_path,
+}
+
+typedefs.router_path = Schema.define {
```
mayocream, line 307: how about “path_with_regex”

kong/db/schema/typedefs.lua
suikabreaker: I searched though the code and find no other use of paths. Maybe we could rename(instead of duplicate).

Suika-Kong: I searched though the code and find no other use of paths. Maybe we could rename(instead of duplicate).

Tieske (approved):
lgtm, provided tests pass

Suika-Kong:
> lgtm, provided tests pass

"pr-merge" seems unrelated. Other tests are passing.

PR #9141, opened 2022-07-25
fix(clustering) wRPC bugs fix

Suika-Kong:
1. CP crashes if DP report meta data with empty plugin list
2. the client is inserted to the list before it reported "basic info", therefore it's possible for an incompatible client to receive config pushs.
3. If no error happens, and no ping is sent yet, update_sync_status is not called and thus we cannot query that DP
4. CP does not return result of compatibility check ReportMetadata
5. Unnessary argument when dp creates connection(refactory leftover)

FT-3030

### DIFF
--- a/kong/clustering/wrpc_control_plane.lua
+++ b/kong/clustering/wrpc_control_plane.lua
@@ -40,6 +40,8 @@ local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
 local _log_prefix = "[wrpc-clustering] "
 
 local ok_table = { ok = "done", }
+local initial_hash = string.rep("0", 32)
+local empty_table = {}
 
 
 local function handle_export_deflated_reconfigure_payload(self)
@@ -61,12 +63,24 @@ local function init_config_service(wrpc_service, cp)
   end)
 
   wrpc_service:set_handler("ConfigService.ReportMetadata", function(peer, data)
-    local client = cp.clients[peer.conn]
+    local client = peer.client
+    local cp = peer.cp
     if client then
       ngx_log(ngx_INFO, _log_prefix, "received initial metadata package from client: ", client.dp_id)
       client.basic_info = data
+      client.dp_plugins_map = plugins_list_to_map(client.basic_info.plugins or empty_table)
       client.basic_info_semaphore:post()
     end
+    
+    local _, err
+    _, err, client.sync_status = cp:check_version_compatibility(client.dp_version, client.dp_plugins_map, client.log_suffix)
+    client:update_sync_status()
+    if err then
+      ngx_log(ngx_ERR, _log_prefix, err, client.log_suffix)
+      client.basic_info = nil -- drop the connection
+      return { error = err, }
+    end
+
     return ok_table
   end)
 end
@@ -195,10 +209,31 @@ function _M:handle_cp_websocket()
     dp_version = dp_version,
     log_suffix = log_suffix,
     basic_info = nil,
-    basic_info_semaphore = semaphore.new()
+    basic_info_semaphore = semaphore.new(),
+    dp_plugins_map = {},
+    cp_ref = self,
+    config_hash = initial_hash,
+    sync_status = CLUSTERING_SYNC_STATUS.UNKNOWN,
   }
-  self.clients[w_peer.conn] = client
+  w_peer.client = client
+  w_peer.cp = self
+
   w_peer:spawn_threads()
+
+  local purge_delay = self.conf.cluster_data_plane_purge_delay
+  function client:update_sync_status()
+    local ok, err = kong.db.clustering_data_planes:upsert({ id = dp_id, }, {
+      last_seen = self.last_seen,
+      config_hash = self.config_hash ~= "" and self.config_hash or nil,
+      hostname = dp_hostname,
+      ip = dp_ip,
+      version = dp_version,
+      sync_status = self.sync_status, -- TODO: import may have been failed though
+    }, { ttl = purge_delay })
+    if not ok then
+      ngx_log(ngx_ERR, _log_prefix, "unable to update clustering data plane status: ", err, log_suffix)
+    end
+  end
 
   do
     local ok, err = client.basic_info_semaphore:wait(5)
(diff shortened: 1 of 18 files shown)


kong/clustering/wrpc_control_plane.lua
```diff
@@ -235,16 +235,25 @@ function _M:handle_cp_websocket()
...
       wb:send_close()
-      client:update_sync_status()
       return ngx_exit(ngx_CLOSE)
     end
   end
 
+  -- after basic_info report we consider DP connected
+  -- initial sync
+  self.clients[w_peer.conn] = client
+  client:update_sync_status()
```
Suika-Kong, line 256: This is to fix No.3

kong/clustering/wrpc_control_plane.lua
```diff
@@ -235,16 +235,25 @@ function _M:handle_cp_websocket()
...
       ngx_log(ngx_ERR, _log_prefix, err, log_suffix)
       wb:send_close()
-      client:update_sync_status()
       return ngx_exit(ngx_CLOSE)
     end
   end
 
+  -- after basic_info report we consider DP connected
+  -- initial sync
+  self.clients[w_peer.conn] = client
```
Suika-Kong, line 255: We move the line registering clients to list here, so the DP will not be pushed config before it finishes its first meta report.
Fix No.2

kong/clustering/wrpc_data_plane.lua
```diff
@@ -116,7 +116,7 @@ local function communicate_impl(dp)
   if peer and not peer.closing then
     peer:close()
   end
-  peer = wrpc.new_peer(c, get_services(), { channel = dp.DPCP_CHANNEL_NAME })
+  peer = wrpc.new_peer(c, get_services())
```
Suika-Kong, line 119: This line is to fix No.5

kong/clustering/wrpc_control_plane.lua
```diff
@@ -195,30 +208,17 @@ function _M:handle_cp_websocket()
...
+    basic_info_semaphore = semaphore.new(),
+    dp_plugins_map = empty_table,
+    cp_ref = self,
+    config_hash = string.rep("0", 32), -- initial hash
+    sync_status = CLUSTERING_SYNC_STATUS.UNKNOWN,
   }
-  self.clients[w_peer.conn] = client
-  w_peer:spawn_threads()
+  w_peer.client = client
+  w_peer.cp = self
```
Suika-Kong, line 218: This allows the `ReportMetadata` handler to access the client object, so the handling of `client:update_sync_status` can be called there.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -195,30 +208,17 @@ function _M:handle_cp_websocket()
     dp_version = dp_version,
     log_suffix = log_suffix,
     basic_info = nil,
-    basic_info_semaphore = semaphore.new()
+    basic_info_semaphore = semaphore.new(),
+    dp_plugins_map = empty_table,
+    cp_ref = self,
+    config_hash = string.rep("0", 32), -- initial hash
+    sync_status = CLUSTERING_SYNC_STATUS.UNKNOWN,
```
Suika-Kong, line 215: We move the initializations here so they will be accessible from `ReportMetadata`.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -61,12 +62,24 @@ local function init_config_service(wrpc_service, cp)
...
       client.basic_info_semaphore:post()
     end
+    
+    local _, err
+    _, err, client.sync_status = cp:check_version_compatibility(client.dp_version, client.dp_plugins_map, client.log_suffix)
+    client:update_sync_status()
+    if err then
+      ngx_log(ngx_ERR, _log_prefix, err, client.log_suffix)
+      client.basic_info = nil -- drop the connection
+      return { error = err, }
```
Suika-Kong, line 80: This is to fix No.4

kong/clustering/wrpc_control_plane.lua
```diff
@@ -61,12 +62,24 @@ local function init_config_service(wrpc_service, cp)
   end)
 
   wrpc_service:set_handler("ConfigService.ReportMetadata", function(peer, data)
-    local client = cp.clients[peer.conn]
+    local client = peer.client
+    local cp = peer.cp
     if client then
       ngx_log(ngx_INFO, _log_prefix, "received initial metadata package from client: ", client.dp_id)
       client.basic_info = data
+      client.dp_plugins_map = plugins_list_to_map(client.basic_info.plugins or empty_table)
```
Suika-Kong, line 70: Fix No.1

kong/clustering/wrpc_control_plane.lua
```diff
@@ -235,17 +235,27 @@ function _M:handle_cp_websocket()
...
       return ngx_exit(ngx_CLOSE)
     end
   end
 
+  -- after basic_info report we consider DP connected
+  -- initial sync
+  client:update_sync_status()
   self:push_config_one_client(client)    -- first config push
+  -- put it here to prevent DP from receiving broadcast config pushes before the first config pushing
+  self.clients[w_peer.conn] = client
```
Suika-Kong, line 258: We move the line registering clients to list here, so the DP will not be pushed config before it finishes its first meta report.
Fix No.2

kong/clustering/wrpc_control_plane.lua
```diff
@@ -195,30 +208,17 @@ function _M:handle_cp_websocket()
     dp_version = dp_version,
     log_suffix = log_suffix,
     basic_info = nil,
-    basic_info_semaphore = semaphore.new()
+    basic_info_semaphore = semaphore.new(),
+    dp_plugins_map = empty_table,
```
chronolaw, line 212: Is `dp_plugins_map` a constant table? or it will be changed later?

kong/clustering/wrpc_control_plane.lua
```diff
@@ -195,30 +208,17 @@ function _M:handle_cp_websocket()
...
+    cp_ref = self,
+    config_hash = string.rep("0", 32), -- initial hash
```
chronolaw, line 214: I think `string.rep("0", 32)` can change to a constant string.

kong/clustering/wrpc_control_plane.lua
Suika-Kong: Surely not. This is the initial value.

kong/clustering/wrpc_control_plane.lua
Suika-Kong: Of cause but this is inherited from code before this fix.

kong/clustering/wrpc_control_plane.lua
chronolaw: As you said, then here we should not use `empty_table`, or else it will be referred by other objects.

I think it should be `{}`.

kong/clustering/wrpc_control_plane.lua
chronolaw: That's OK.

kong/clustering/wrpc_control_plane.lua
javierguerragiraldez: > I think it should be `{}`.

+1

Suika-Kong:
> tests?

This is a large PR, so I put tests here: https://github.com/Kong/kong/pull/9138 .
If you feel that is necessary, we could merge that PR to this branch and then merge this PR.

kong/clustering/wrpc_control_plane.lua
Suika-Kong: As far as I know, this table is always read-only (and when its value changes, it's a whole new table). But it's OK to write `{}` just in case.

hutchic:
@Suika-Kong I think that's sufficient. As long as we have tests so we don't repeat the bugs you've corrected here :+1:

chronolaw (approved):

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -353,272 +353,288 @@ for _, strategy in helpers.each_strategy() do
...
-        "targets",
-        "certificates",
-        "clustering_data_planes",
-      }) -- runs migrations
-
-      bp.plugins:insert {
-        name = "key-auth",
-      }
-      lazy_setup(function()
+  for _, cluster_protocol in ipairs{"wrpc", "json"} do
```
fffonion, line 356: let's not insert identions below so that we won't create lots of diffs and it will cause trouble when merging to ee

spec/helpers.lua
```diff
@@ -3026,6 +3018,95 @@ local function clustering_client(opts)
...
+
+--- Simulate a Hybrid mode DP and connect to the CP specified in `opts`.
+-- @function clustering_client
+-- @param opts Options to use, the `host`, `port`, `cert` and `cert_key` fields
+-- are required.
+-- Other fields that can be overwritten are:
+-- `node_hostname`, `node_id`, `node_version`, `node_plugins_list`. If absent,
+-- they are automatically filled.
+-- @return msg if handshake succeeded and initial message received from CP or nil, err
+local function clustering_client(opts)
```
fffonion, line 3103: i would suggest to move this into 01-sync_spec.lua not expose in spec/helpers.lua

fffonion (approved):

spec/helpers.lua
Suika-Kong: This function is here before the change.

PR #9145, opened 2022-07-26
feat(clustering): improve wRPC protocol detection

flrgh:
### Summary

The data-plane clustering code optimistically makes a request to the control-plane (`HEAD /v1/wrpc`) to determine if it can use wRPC or not. If the request did not return a 404, it would assume wRPC.

We gate the presence of the `/v1/wrpc` block based on the `legacy_hybrid_protocol` config param and rely on NGINX to return a 404.

This is all behaving kinda finnicky in EE land (for reasons I spent an hour debugging and couldn't figure out). Needless to say, I think a more explicit form of feature detection is warranted.

### Full changelog

* `/v1/wrpc` returns 200 immediately for `HEAD` requests
* `kong.clustering.utils.check_protocol_support` checks response codes more strictly now:
  - `404` => legacy
  - `200` => wrpc
  - anything else is an error
 * renamed `06-lagacy_switch_spec.lua` => `06-legacy_switch_spec.lua`

### DIFF
--- a/kong/clustering/utils.lua
+++ b/kong/clustering/utils.lua
@@ -349,11 +349,16 @@ function _M.check_protocol_support(conf, cert, cert_key)
     return nil, err
   end
 
-  if res.status == 404 then
+  local status = res.status
+
+  if status == 200 then
+    return "v1" -- wrpc
+
+  elseif status == 404 then
     return "v0"
   end
 
-  return "v1"   -- wrpc
+  return nil, "bad response code: " .. tostring(status)
 end
 
 
--- a/kong/clustering/wrpc_control_plane.lua
+++ b/kong/clustering/wrpc_control_plane.lua
@@ -28,6 +28,7 @@ local ngx_var = ngx.var
 local timer_at = ngx.timer.at
 local isempty = require("table.isempty")
 local sleep = ngx.sleep
+local get_method = ngx.req.get_method
 
 local plugins_list_to_map = clustering_utils.plugins_list_to_map
 local deflate_gzip = utils.deflate_gzip
@@ -216,6 +217,11 @@ _M.check_configuration_compatibility = clustering_utils.check_configuration_comp
 
 
 function _M:handle_cp_websocket()
+  if get_method() == "HEAD" then
+    -- feature detection request from a DP node
+    return ngx_exit(200)
+  end
+
   local dp_id = ngx_var.arg_node_id
   local dp_hostname = ngx_var.arg_node_hostname
   local dp_ip = ngx_var.remote_addr
--- a/spec/02-integration/09-hybrid_mode/06-legacy_switch_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/06-legacy_switch_spec.lua
@@ -1,0 +1,104 @@
+local helpers = require "spec.helpers"
+local ssl = require "ngx.ssl"
+local http = require "resty.http"
+local read_file = require("pl.file").read
+
+local function new_conf(extra)
+  local t = {
+    cluster_cert          = "spec/fixtures/ocsp_certs/kong_data_plane.crt",
+    cluster_cert_key      = "spec/fixtures/ocsp_certs/kong_data_plane.key",
+    cluster_mtls          = "pki",
+    cluster_listen        = "127.0.0.1:9005",
+    cluster_control_plane = "127.0.0.1:9005",
+    cluster_ca_cert       = "spec/fixtures/ocsp_certs/ca.crt",
+    cluster_server_name   = "kong_clustering",
+
+    nginx_conf = "spec/fixtures/custom_nginx.template",
+  }
+
+  for k, v in pairs(extra or {}) do
+    t[k] = v
+  end
+
+  return t
+end
+
+local dp_cert, dp_key
+do
+  local conf = new_conf()
+  dp_cert = assert(ssl.parse_pem_cert(assert(read_file(conf.cluster_cert))))
+  dp_key = assert(ssl.parse_pem_priv_key(assert(read_file(conf.cluster_cert_key))))
+end
+
+for _, proto in ipairs({ "json", "wrpc" }) do
+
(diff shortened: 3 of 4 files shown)


kong/clustering/wrpc_control_plane.lua
```diff
@@ -173,6 +174,15 @@ _M.check_configuration_compatibility = clustering_utils.check_configuration_comp
 
 
 function _M:handle_cp_websocket()
+  if self.conf.legacy_hybrid_protocol then
+    ngx_log(ngx_DEBUG, "received a request to the wRPC listener, but wRPC is disabled")
+    return ngx_exit(405)
```
Suika-Kong, line 179: The "legacy\_hybrid\_protocol" switch is barely an "escape hatch" to buy some time when things go wrong,  not intended for users, not documented intentionally and will be removed eventually.
I'm not sure if we should introduce 1 more status to let the client know the difference given the motivation for this switch.

Suika-Kong (commented):
It is necessary to handle other HTTP status codes. But I hesitate to agree with adding a dedicated code for the switch.

kong/clustering/wrpc_control_plane.lua
flrgh: I can take this out, but I will need to add the `legacy_hybrid_protocol` gate to the [test template](https://github.com/Kong/kong/blob/8920006b2400bbc8e2ba8d329d8b6babc11aaf21/spec/fixtures/custom_nginx.template#L454-L458) that exists in the [main template](https://github.com/Kong/kong/blob/8920006b2400bbc8e2ba8d329d8b6babc11aaf21/kong/templates/nginx_kong.lua#L442-L448). If the tests pass with that in place it's good enough for me.

kong/clustering/utils.lua
```diff
@@ -348,11 +348,16 @@ function _M.check_protocol_support(conf, cert, cert_key)
     return nil, err
   end
 
-  if res.status == 404 then
+  local status = res.status
+
+  if status == 200 then
+    return "v1" -- wrpc
+
+  elseif status == 404 or res.status == 405 then
```
chronolaw, line 356: I think now there is no `405` response code, should we remove it?

kong/clustering/wrpc_control_plane.lua
Suika-Kong: > I can take this out, but I will need to add the `legacy_hybrid_protocol` gate to the [test template](https://github.com/Kong/kong/blob/8920006b2400bbc8e2ba8d329d8b6babc11aaf21/spec/fixtures/custom_nginx.template#L454-L458) that exists in the [main template](https://github.com/Kong/kong/blob/8920006b2400bbc8e2ba8d329d8b6babc11aaf21/kong/templates/nginx_kong.lua#L442-L448). If the tests pass with that in place it's good enough for me.

I think it's fine to handle the switch here(as returning 404 explicitly).

kong/templates/nginx_kong.lua
```diff
@@ -439,13 +439,11 @@ server {
         }
     }
 
-> if not legacy_hybrid_protocol then
```
Suika-Kong, line 442: Seems we are missing the gate here and not handling the switch at `_M:handle_cp_websocket`?

flrgh: Hmm? Sorry, I don't quite follow.

flrgh:
Added the `legacy_hybrid_protocol` check to the test template. It didn't seem to break any tests, so :crossed_fingers: that this is good to go.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -173,6 +174,11 @@ _M.check_configuration_compatibility = clustering_utils.check_configuration_comp
 
 
 function _M:handle_cp_websocket()
+  if get_method() == "HEAD" then
```
flrgh, line 177: @hbagdi any reservations? This will essentially add a new clustering API behavior (`HEAD /v1/wrpc` => HTTP 200), which I assume would need to make it into the [wRPC handler in koko](https://github.com/Kong/koko/blob/f407ae7363989aca9e08b8808dd5efd05ebea7b7/internal/server/kong/ws/server.go#L153).

Before, `HEAD /v1/wrpc` would return HTTP 400 for missing some required params. Short-circuiting on `HEAD` allows us to make the data-plane feature detection more explicit (see the change to `kong/clustering/utils.lua` above).

flrgh:
@chronolaw @Suika-Kong @hbagdi, any :+1:/:-1: on whether this should get merged for 3.0 or not? It's not a hill I would die on, but I think it's a worthwhile improvement, and making any kind of changes to the wRPC clustering API after 3.0 is out will be a lot harder to justify.

Suika-Kong (approved):

chronolaw:
The current code seems to work well, I think this PR is not urgent and we can discuss it after 3.0 release.

kong/templates/nginx_kong.lua
flrgh: oh never mind, I am dumb :man_facepalming:

tyler-ball:
Michael is going to cherry-pick https://github.com/Kong/kong/pull/9145/commits/7c299b48b338ee50b698acda4b49e0d02bd9700a as its own PR (because that fixes some test issues in EE). I moved this to a 3.1 milestone as a future improvement.

hbagdi:
@flrgh Still pursuing this change?

flrgh:
> @flrgh Still pursuing this change?

Yes I'd like to.

It's a little bit tougher to realize now that 3.0 is out though, since it changes the response-handling logic in the DP to be more strict, which could cause breakage. I probably need to change this to a `feat` and merge the CP change (both here and in Koko) at least one minor release ahead of the DP changes.

chronolaw:
Due to #9740, perhaps we could close it.

flrgh:
> Due to #9740, perhaps we could close it.

Agreed. We can revisit this if anything changes later on.

PR #9148, opened 2022-07-26
style(router) use internal cache size

chronolaw:
### Summary

In #9096 , atc_router added cache feature, so the cache size should be decided by the proper object, not a fixed value.

### Full changelog

* refactor a function `is_traditional`
* set `_M.MATCH_LRUCACHE_SIZE` by `if-else`


chronolaw:
`kong.configuration.router_flavor` is nil when module inits, so this PR can not work as expected.

PR #9149, opened 2022-07-26
Responsive design implemented

si88har1h:
### Summary
The file initially was following poor CSS methodologies. I implemented responsive units and refactored the pre-existing code

### DIFF
--- a/spec/ldoc.css
+++ b/spec/ldoc.css
@@ -1,291 +1,370 @@
-body { 
-    color: #47555c;
-    font-size: 16px;
-    font-family: "Open Sans", sans-serif;
-    margin: 0;
-    background: #eff4ff;
+html {
+  font-size: 62.5%;
 }
 
-a:link { color: #008fee; }
-a:visited { color: #008fee; }
-a:hover { color: #22a7ff; }
+body {
+  color: #47555c;
+  font-size: 16rem;
+  font-family: "Open Sans", sans-serif;
+  margin: 0;
+  background: #eff4ff;
+}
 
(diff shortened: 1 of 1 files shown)


spec/ldoc.css
```diff
@@ -1,291 +1,370 @@
...
 #content h1 {
-    background-color: #2c3e67;
-    color: white;
-    padding: 15px;
-    margin: 0px;
+  background-color: #2c3e67;
+  color: white;
+  padding: 15rem;
+  margin: 0rem;
 }
```
mayocream, line 116: Replace `px` with `rem` directly won't work.

> rem Represents the [font-size](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size) of the root element (typically [<html>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html)). When used within the root element [font-size](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size), it represents its initial value (a common browser default is 16px, but user-defined preferences may modify this).

ref: https://developer.mozilla.org/en-US/docs/Web/CSS/length

mayocream (changes requested):

mayocream:
Please try to install [ldoc](https://github.com/lunarmodules/LDoc).
```bash
cd spec
ldoc .
cd docs && live-server
```

spec/ldoc.css
si88har1h: Thats why i altered the default browser font-size in html
`html{
font-size:62.5%;
}`
62.5% i.e. 10px as the default

Issue #8124, opened 2020-08-24
Add option to return the traceId to the API user


r1ckr:
Would be nice to be able to return the trace id back to the API user as an X-Request-Id or X-Correlation-Id header. The same way that the correlation id plugin does.

Something like:
```config.external_http_header_name=X-Request-Id```

If set, it would send back to the users the trace id in it.

StarlightIbuki:
Dear contributor,
We're closing this issue as there hasn't been any update to it for a long time. If the issue is still relevant in the latest version, please feel free to reopen it. We're more than happy to revisit it again. Your contribution is greatly appreciated!
Please have a look at [our pledge to the community](https://github.com/Kong/kong/blob/master/COMMUNITY_PLEDGE.md) for more information.
Sincerely,
Kong Gateway Team

PR #9173, opened 2022-08-01
feat(zipkin) support adding trace id in response header

windmgc:
### Summary

This PR adds a new field in Zipkin plugin schema to support adding trace id in the response header.

### Full changelog

* Add `response_header_for_traceid` field in Zipkin plugin. The plugin will set the corresponding header in the response if the field is specified with a string value. 
* Add related tests.

### Issue reference

Fix [FTI-4198](https://konghq.atlassian.net/browse/FTI-4198), https://github.com/Kong/kong/issues/8124


kong/plugins/zipkin/schema.lua
```diff
@@ -65,6 +65,7 @@ return {
           { connect_timeout = typedefs.timeout { default = 2000 } },
           { send_timeout = typedefs.timeout { default = 5000 } },
           { read_timeout = typedefs.timeout { default = 5000 } },
+          { response_header_for_traceid = { type = "string", default = nil }},
```
mayocream, line 68: Will it be better to add an `http_` prefix? It only works in the HTTP mode.

windmgc: field renamed to `http_response_header_for_traceid`

mayocream (approved):

kong/plugins/zipkin/handler.lua
```diff
@@ -226,6 +226,9 @@ if subsystem == "http" then
 
     local proxy_span = get_or_add_proxy_span(zipkin, header_filter_start_mu)
     proxy_span:annotate("khs", header_filter_start_mu)
+    if conf.http_response_header_for_traceid then
```
chronolaw, line 229: `local xxx = conf.http_response_header_for_traceid` then use it.

PR #9175, opened 2022-08-01
workflows(stale) Add wording for stale message

ArjunSharda:
# Description

This will add wording to the stale message to possibly make it more professional in my opinion, with a clear description in the message for the notification of inactivity.


kikito (approved):

hbagdi:
Rejecting because it doesn't add any meaningful difference.

PR #9177, opened 2022-08-02
Fix plugin `azure functions` to support working without dummy service

liverpool8056:
### Summary

Fix bug for plugin `azure functions` to support working without dummy service

### Full changelog

* Fix bug for plugin `azure functions` to support working without dummy service
* Modify related tests in spec to cover this case


kong/plugins/azure-functions/handler.lua
```diff
@@ -54,7 +54,11 @@ function azure:access(config)
   local request_headers = kong.request.get_headers()
   local request_args = kong.request.get_query()
 
-  local upstream_uri = var.upstream_uri
+  -- strip any query args
+  local upstream_uri = var.upstream_uri or var.request_uri
+  local s, _ = string.find(upstream_uri, "?")
```
chronolaw, line 59: here `_` is unnecessary,  just `local s = string.find(...)`.

Also, use `string.find(x, '?', 1, true)` to speed up.

kong/plugins/azure-functions/handler.lua
```diff
@@ -54,7 +54,11 @@ function azure:access(config)
...
+  upstream_uri = s and string.sub(upstream_uri, 1, s-1) or upstream_uri
```
chronolaw, line 60: style: `s - 1`

chronolaw:
We should localize `string.find` and `string.sub`

chronolaw (approved):

fffonion (approved):

PR #9180, opened 2022-08-02
fix(plugins) gRPC gateway uri para bool handling

Suika-Kong:
gRPC gateway should handle the "false" argument as false for boolean parameters.

This change is to be consistent with [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway), where by `http://localhost:8090/v1/example/echo?stringv=false&boolean=false&number=1` you get:
```json
{
	"stringv": "false",
	"boolean": false,
	"number": 1
}
```
as echo.

Before this fix, any value passed to a boolean parameter is treated as a string and thus encoded as `true`.

Fix FTI-3335


Suika-Kong:
For who wants to review:
`fixture/grpc/targetservice`/* is auto-generated.
`targetservice.proto` and grpc-target.go are modified for testing the behavior fixed by this PR.

kong/plugins/grpc-gateway/deco.lua
```diff
@@ -169,21 +169,42 @@ function deco.new(method, path, protofile)
...
 
+local function get_field_type(typ, field)
+  local _, _, field_typ = pb.field(typ, field)
+  return field_typ
+end
+
+local function encode_fix(v, typ)
+  if typ == "bool" then
+    -- special case for URI parameters
+    return v and v ~= "0" and v ~= "false"
```
ADD-SP, line 180:
```suggestion
    return v and v ~= "0" and v ~= "false" and v ~= "null"
```

Should we handle the string `"null"`?

Suika-Kong: We should not.
Here's the official implementation's response to null:
```json
{
	"code": 3,
	"message": "parsing field \"boolean\": strconv.ParseBool: parsing \"null\": invalid syntax",
	"details": []
}
```

ADD-SP (approved):

PR #9182, opened 2022-08-02
feat(db): disconnect table_name from entity name

jschmid1:
### Summary

Allows to define table_name field in a schema. Before the table name was deduced from the entity name.

### Full changelog

* Add `table_name` schema field

### DIFF
--- a/spec/02-integration/03-db/07-tags_spec.lua
+++ b/spec/02-integration/03-db/07-tags_spec.lua
@@ -447,8 +447,8 @@ for _, strategy in helpers.each_strategy() do
       func = describe
     end
     func("trigger defined for table", function()
-      for entity_name, dao in pairs(db.daos) do
-        entity_name = dao.schema.table_name
+      for _, dao in pairs(db.daos) do
+        local entity_name = dao.schema.table_name
         if dao.schema.fields.tags then
           it(entity_name, function()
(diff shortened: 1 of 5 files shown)


spec/02-integration/03-db/07-tags_spec.lua
```diff
@@ -448,6 +448,7 @@ for _, strategy in helpers.each_strategy() do
     end
     func("trigger defined for table", function()
       for entity_name, dao in pairs(db.daos) do
```
hanshuebner, line 450: Would prefer `for _, dao in ...` here to avoid reassignment

hanshuebner (approved):
Looks good.  How do we know that all relevant references to `schema.name` have been updated to `schema.table_name`?  If we miss out on a few, it probably won't be too difficult to fix later on, but I wonder how abundant SQL statement generation from schema objects is throughout the code base.

Approved with one suggestion.

spec/02-integration/03-db/07-tags_spec.lua
jschmid1: via https://github.com/Kong/kong/pull/9182/commits/86da43bd7545b146178c05c834099a6168f4cc6c

hanshuebner (approved):

PR #9186, opened 2022-08-03
fix(request-transformer) an empty JSON array after being decoded, can…

catbro666:
…'t be properly re-encoded as an array. Instead, it becomes an empty json object.

This PR fixes this issue. An empty json array will remain to be an empty json array after being decoded and re-encoded.

This is a sister PR to PR:

[#4901](https://github.com/Kong/kong/pull/4901)

It contains a similar fix, but now to `request-transformer` (this plugin).

FTI-4205

### DIFF
--- a/kong/plugins/request-transformer/access.lua
+++ b/kong/plugins/request-transformer/access.lua
@@ -19,7 +19,6 @@ local encode_args = ngx.encode_args
 local ngx_decode_args = ngx.decode_args
 local type = type
 local str_find = string.find
-local pcall = pcall
 local pairs = pairs
 local error = error
 local rawset = rawset
--- a/kong/plugins/request-transformer/access.lua
+++ b/kong/plugins/request-transformer/access.lua
@@ -1,5 +1,5 @@
 local multipart = require "multipart"
-local cjson = require "cjson"
+local cjson = require("cjson.safe").new()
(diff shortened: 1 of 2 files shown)


ms2008 (approved):
LGTM

chronolaw (approved):

kong/plugins/request-transformer/access.lua
```diff
@@ -1,5 +1,5 @@
 local multipart = require "multipart"
-local cjson = require "cjson"
+local cjson = require("cjson.safe").new()
```
flrgh, line 2: If you're going to switch from `cjson` to `cjson.safe` then you need to update the `parse_json` local function below, as it uses `pcall`.

flrgh (changes requested):

ms2008 (changes requested):
https://github.com/Kong/kong/blob/a9e58a1f3c8bd2b89b1a7be67c9179d15bd06147/kong/plugins/request-transformer/access.lua#L22

This line also needs to be removed.

flrgh (approved):
:+1: lgtm; good to go after removing the now-unused `local pcall` as per @ms2008's comment.

catbro666:
> 

done

ms2008 (approved):

PR #9194, opened 2022-08-04
feat(acme): add host on certificate update log

zekth:
### Summary

When having several certs, when it fails to update there is no way to know which one is failing.

### Full changelog

* adding log entry


fffonion:
maybe we should also add `host` for error log at Line 156 for future us 😄

fffonion (approved):
approved but holding for 3.1 release

zekth:
> approved but holding for 3.1 release

Should i also commit to 2.X ?

PR #9199, opened 2022-08-05
fix(clustering) fix rpc id of negotiation

Suika-Kong:
Should we also change the id of the service? This service is more fundamental than config.

### DIFF
--- a/kong/include/kong/services/negotiation/v1/negotiation.proto
+++ b/kong/include/kong/services/negotiation/v1/negotiation.proto
@@ -12,7 +12,7 @@ service NegotiationService {
   // to get which services does the DP handle.
   //
   // Call direction: DP to CP
-  // +wrpc:rpc-id=4
+  // +wrpc:rpc-id=1
   rpc NegotiateServices(kong.model.NegotiateServicesRequest) returns (kong.model.NegotiateServicesResponse);
 }
 


javierguerragiraldez (approved):
great!
with Kong/koko#344  we'll fix this at both sides

kong/include/kong/services/negotiation/v1/negotiation.proto
```diff
@@ -12,7 +12,7 @@ service NegotiationService {
   // to get which services does the DP handle.
   //
   // Call direction: DP to CP
-  // +wrpc:rpc-id=4
+  // +wrpc:rpc-id=1
```
chronolaw, line 15: Could you give a short explain for this id? What is the number's principle?

chronolaw (approved):

kong/include/kong/services/negotiation/v1/negotiation.proto
chronolaw: I think it is just to sync up with koko?

Suika-Kong: Yes. This was a typo, and koko already fix it.

javierguerragiraldez:
friendly remainder that Kong/koko#344 is already merged and deployed.

PR #9207, opened 2022-08-08
chore(deps) bump lua-protobuf from 0.3.3 to 0.4.0

bungle:
#### 0.4.0

* add encode_default_values
* add encode_order
* add oneof name field for which field is decoded
* check stack before push default values
* refactor the logic for setting default value
* fix option support in protoc.lua
* fix wrong error message when use varint as map
* fix decode hooks check
* fix msvc warnings

#### 0.3.4

* fix memory leak in pb.load
* fix bugs in "protoc.lua"
* fix memory usage in messages contains oneof
* code improved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -265,6 +265,8 @@
   [#8845](https://github.com/Kong/kong/pull/8845)
 - Bumped penlight from 1.12.0 to 1.13.1
   [#9206](https://github.com/Kong/kong/pull/9206)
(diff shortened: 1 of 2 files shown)


chronolaw (approved):

dndx (approved):

bungle:
There seems to be still issues with `protobuf` (or its upgrade with our test suite). Thus I marked this as `do not merge` until the issues are resolved. I do not have currently time to debug this further. If anyone wants to try, please go ahead.

PR #9209, opened 2022-08-09
fix(db) use a checksum to limit unique key length

flrgh:
### Summary

LMDB has a (compile-time) max key length limitation of 511. It's easy to exceed this limit when generating unique field indexes in dbless, since they use the [potentially unbounded] field value as part of the key.

With this update, we use a sha1 checksum of the field value instead of just the raw value itself in order to limit the key length.

There is definitely a non-zero performance cost to this, but I don't see any way around that. And we can't easily get much cheaper than sha1.

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -10,6 +10,8 @@ local constants = require "kong.constants"
 local txn = require "resty.lmdb.transaction"
 local lmdb = require "resty.lmdb"
 local on_the_fly_migration = require "kong.db.declarative.migrations"
+local to_hex = require("resty.string").to_hex
+local sha1 = require "resty.sha1"
 
 local setmetatable = setmetatable
 local tostring = tostring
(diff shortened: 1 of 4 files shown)


flrgh:
This has already been merged in EE master, due to the breakage it was causing, so it would be good to merge it in for 3.0 for consistency's sake (and because it very well could be causing breakage in OSS somewhere that we're not aware of).

dndx (approved):
This is a fine fix and nice catch! Although I wonder if we should bump `MDB_MAXKEYSIZE` in our build as well. We don't have any compatibility concerns for bumping it AFAIK.

Regardless, this can be merged for now, as it fixes the issue immediately.

dndx (approved):

dndx:
@fffonion Will this break in FIPS mode?

fffonion:
Using message digest for non-cryptographic purposes is a vague area of fips, but we should use algorithms like sha256 just to be safe.

flrgh:
Using sha1 seems like a pretty lateral change to me given that the field value was previously just part of the key, in plain text. Does LMDB encrypt key data or only value data?

Not super opposed to updating to sha256 (or making the algo config-dependent), just don't want to do so unless it's really needed.

dndx:
@flrgh the issue is not related to encryption, it's about SHA1 might trigger some unnecessary eyes under FIPS mode.

Maybe use CRC32 instead? That will make it clear that we are not using this for cryptographic purposes.

flrgh:
@dndx ah that makes sense.

To me the collision probability with CRC32 seems to put it out of the running in this use case. I assume md5 would raise the same concerns as sha1? If so I think I'll just replace with sha256 for now.

flrgh:
@fffonion I've switched to sha256. Can you take a quick look and merge if this is workable for you?

dndx (approved):

PR #9211, opened 2022-08-09
fix(router) atc-router ignores default port

chronolaw:
### Summary

* Strip default port in http.host, then pass it to atc-router, 
so atc-router can match host with default port.
* Add a `net.port` and split host:port
* Enable some test cases.

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -80,6 +80,10 @@ end
 -- example.com:123 => example.com, 123
 -- example.*:123 => example.*, 123
 local function split_host_port(h)
+  if not h then
+    return nil, nil
+  end
+
   local p = h:find(":", nil, true)
   if not p then
     return h, nil
@@ -150,7 +154,7 @@ local function gen_for_field(name, op, vals, val_transform)
       values[values_n] = (val_ex and "(" or "") ..
                          name .. " " .. op .. " " ..
                          "\"" .. val .. "\"" ..
-                         (val_ex and val_ex .. ")"or "")
+                         (val_ex and val_ex .. ")" or "")
     end
   end
 
@@ -436,6 +440,8 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
 
   local c = context.new(self.schema)
 
+  local host, port = split_host_port(req_host)
+
   for _, field in ipairs(self.fields) do
     if field == "http.method" then
       assert(c:add_value("http.method", req_method))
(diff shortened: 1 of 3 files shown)


kong/router/atc_compat.lua
```diff
@@ -432,8 +432,16 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
...
+    elseif field == "http.host" and req_host then
+      local host = req_host
+
+      -- ignores default port
+      local default_port = req_scheme == "https" and ":443" or ":80"
+      if sub(req_host, -#default_port) == default_port then
+        host = sub(req_host, 1, -#default_port - 1)
+      end
+
+      assert(c:add_value("http.host", host))
```
chronolaw, line 444: It seems that we can not add two different host to context.

If I use `assert(c:add_value("http.host", req_host))` before this line, the test case will fail.

spec/01-unit/08-router_spec.lua
```diff
@@ -324,12 +324,14 @@ for _, flavor in ipairs({ "traditional", "traditional_compatible" }) do
...
 
-      it_trad_only("[host] ignores default port", function()
+      it("[host] ignores default port", function()
         -- host
         local match_t = router:select("GET", "/", "domain-1.org:80")
         assert.truthy(match_t)
         assert.same(use_case[1].route, match_t.route)
-        assert.same(use_case[1].route.hosts[1], match_t.matches.host)
+        if flavor == "traditional" then
+          assert.same(use_case[1].route.hosts[1], match_t.matches.host)
```
chronolaw, line 333: Why does atc-router has no field `match_t.matches.host` ?

kong/router/atc_compat.lua
```diff
@@ -432,8 +432,16 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
     elseif field == "http.path" then
```
chronolaw, line 432: Should we check if `req_uri` is `nil`, then we could skip `c:add_value`

kong/router/atc_compat.lua
dndx: This doesn't seems to be the right solution, we should add a new `Int` type field `net.port` into the schema instead and generate additional predicates during match.

spec/01-unit/08-router_spec.lua
dndx: This field makes no sense in a DSL especially if `host` metched multiple conditions with `&&` or `
||`, plus this is only used by tests so we choose to not support it.

kong/router/atc_compat.lua
chronolaw: Understood, I will do it.

chronolaw:
There are some IPv6 test cases that are still pending, Let's fix them in other PR.

kong/router/atc_compat.lua
```diff
@@ -46,6 +46,7 @@ local get_upstream_uri_v0  = utils.get_upstream_uri_v0
 
 
 local TILDE            = byte("~")
+local ASTER            = byte("*")
```
dndx, line 49: Let's use `ASTERISK` instead, `ASTER` is not a word.

kong/router/atc_compat.lua
```diff
@@ -355,10 +388,16 @@ function _M.new(routes, cache, cache_neg)
...
+
+    if not route_id then
+      return nil, "could not categorize route"
+    end
+
     routes_t[route_id] = route
     services_t[route_id] = r.service
 
     if is_traditional_compatible then
+      --print("atc:", get_atc(route))
```
dndx, line 400:
```suggestion
```

kong/router/atc_compat.lua
chronolaw: done

kong/router/atc_compat.lua
chronolaw: removed, thanks.

dndx (approved):

PR #9217, opened 2022-08-09
chore(db) Change vaults table name, rework 2.8->3.0 migration + test

hanshuebner:
Use "sm_vaults" instead of "vaults" for the SM vaults table in 3.0.
Rework the migration and its test accordingly.  This change enables
blue/green upgrades.

To support upgrades of installations running 2.8.1.3 as well as
earlier 2.8.x versions, the migration test supports both the
`/vaults-beta` and `/vaults` prefix to retrieve SM vaults during
testing.

Fix FT-3182

### DIFF
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -81,22 +81,24 @@ local function c_copy_vaults_to_vault_auth_vaults(coordinator)
   for rows, err in coordinator:iterate("SELECT id, created_at, updated_at, name, protocol, host, port, mount, vault_token FROM vaults") do
     if err then
       log.warn("ignored error while running '016_280_to_300' migration: " .. err)
-      break
+      return true
     end
 
     for _, row in ipairs(rows) do
       local _, err = coordinator:execute(
         "INSERT INTO vault_auth_vaults (id, created_at, updated_at, name, protocol, host, port, mount, vault_token) " ..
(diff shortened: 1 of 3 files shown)


hutchic:
I was going to comment that 2.8.1 to here isn't blue-green capable but it looks like that ability was lost and discussed in https://github.com/Kong/kong/pull/8871 :(

To achieve backwards compatibility (if we still want) one would release a 2.8 patch that has application logic that is forward capable
```
table_name = old_table_name
if ( SELECT EXISTS (
   SELECT FROM information_schema.tables 
   WHERE  table_schema = 'schema_name'
   AND    table_name   = 'new_table_name'
   );) {
      table_name = new_table_name
}
```

cc @bungle

hanshuebner:
> I was going to comment that 2.8.1 to here isn't blue-green capable but it looks like that ability was lost and discussed in #8871 :(

According to our conversation in Slack, it seems that your objection is that we've changed the admin API in incompatible ways by renaming the `/vaults-beta` endpoint to `/vaults`.  I agree, and it would probably have been preferable to make `/vaults-beta` an alias to `/vaults` in the 2.8.1.3 vaults GA release.  As that release is out now, there is probably no point in adding it in _another_ point release.

tyler-ball (approved):
@Hayk-S Please run another migration tests for 2.8.1.2 -> 3.0.0.0 and 2.8.1.3 -> 3.0.0.0 after this finishes building

PR #9233, opened 2022-08-11
fix(clustering) DP errors when CP exits

Suika-Kong:
DP errors in the log when CP exits because CP does not close the connection properly:
```
2022/08/11 11:33:32 [notice] 18802#0: *11 [lua] threads.lua:80: step(): [wRPC] WebSocket frame: failed to receive the first 2 bytes: closed, context: ngx.timer
2022/08/11 11:33:32 [error] 18802#0: *7 [lua] init.lua:252: error while receiving frame from peer: failed to receive the first 2 bytes: closed, context: ngx.timer
2022/08/11 11:33:32 [error] 18802#0: *5 [lua] init.lua:252: error while receiving frame from peer: failed to receive the first 2 by
```
Fix FT-3204

### DIFF
--- a/kong/tools/wrpc/threads.lua
+++ b/kong/tools/wrpc/threads.lua
@@ -76,7 +76,7 @@ local function step(wrpc_peer)
     msg, err = wrpc_peer:receive()
   end
 
-  if err ~= nil and not endswith(err, ": timeout") then
+  if err ~= nil and not endswith(err, ": timeout") and not endswith(err, ": closed") then
     ngx_log(NOTICE, "[wRPC] WebSocket frame: ", err)
     wrpc_peer.closing = true
     return false, err
--- a/spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
+++ b/spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
@@ -132,3 +132,37 @@ for cluster_protocol, conf in pairs(confs) do
     end
   end)
 end
+
+-- note that lagacy modes still error when CP exits
+describe("when CP exits DP does not error #t", function()
+  local need_exit = true
+  setup(function()
+    assert(helpers.start_kong({
+      role = "control_plane",
+      prefix = "servroot1",
+      cluster_cert = "spec/fixtures/kong_clustering.crt",
+      cluster_cert_key = "spec/fixtures/kong_clustering.key",
+      cluster_listen = "127.0.0.1:9005",
+    }))
+    assert(helpers.start_kong({
+      role = "data_plane",
+      prefix = "servroot2",
+      cluster_cert = "spec/fixtures/kong_clustering.crt",
+      cluster_cert_key = "spec/fixtures/kong_clustering.key",
+      cluster_control_plane = "127.0.0.1:9005",
+      proxy_listen = "0.0.0.0:9002",
+      database = "off",
+    }))
+  end)
+  teardown(function()
+    if need_exit then
+      helpers.stop_kong("servroot1")
+    end
+    helpers.stop_kong("servroot2")
+  end)
+  it("", function ()
+    assert(helpers.stop_kong("servroot1"))
+    need_exit = false
+    assert.logfile("servroot2/logs/error.log").has.no.line("[error]", true)
+  end)
+end)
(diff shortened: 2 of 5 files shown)


Suika-Kong:
No changelog item needed cause wRPC is new in the changelog.

chronolaw (approved):

spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
```diff
@@ -132,3 +132,38 @@ for cluster_protocol, conf in pairs(confs) do
...
+      database = "off",
+    }))
+  end)
+  teardown(function()
+    if need_exit then
+      helpers.stop_kong("servroot1")
+    end
+    helpers.stop_kong("servroot2")
+  end)
+  it("", function ()
```
flrgh, line 163: :point_up: description needed for this test

spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
```diff
@@ -132,3 +132,38 @@ for cluster_protocol, conf in pairs(confs) do
...
+    helpers.clean_logfile("servroot2/logs/error.log")
+    assert(helpers.stop_kong("servroot1"))
+    need_exit = false
+    assert.logfile("servroot2/logs/error.log").has.no.line("[error]", true)
```
flrgh, line 167: This assertion checks for `[error]`, but from what I can tell, most of the error-handling code in wrpc uses `ngx.NOTICE` log level. Is this correct?

kong/tools/wrpc/threads.lua
```diff
@@ -76,7 +76,7 @@ local function step(wrpc_peer)
     msg, err = wrpc_peer:receive()
   end
 
-  if err ~= nil and not endswith(err, ": timeout") then
+  if err ~= nil and not endswith(err, ": timeout") and not endswith(err, ": closed") then
```
flrgh, line 79: Should this logic use a `wrpc_peer.closing` check instead? With a graceful shutdown we should receive a close frame, and therefore `wrpc_peer.closing` will be set `true`.

```suggestion
  if err ~= nil
    and not endswith(err, ": timeout")
    and not wrpc_peer.closing
  then
```

I'm not sure if I have it correct, but ideally we still log when an unexpected connection closure happens. A test case would look something like....

```
it("logs an error when the connection is terminated unexpectedly", function ()
  helpers.clean_logfile("servroot2/logs/error.log")
  assert(helpers.signal("servroot1", "-KILL"))
  need_exit = false
  assert.logfile("servroot2/logs/error.log").has.line("some kind of error string")
end)
```

Suika-Kong: > Should this logic use a `wrpc_peer.closing` check instead? With a graceful shutdown we should receive a close frame, and therefore `wrpc_peer.closing` will be set `true`.
> 
> I'm not sure if I have it correct, but ideally we still log when an unexpected connection closure happens. A test case would look something like....
> 
> ```
> it("logs an error when the connection is terminated unexpectedly", function ()
>   helpers.clean_logfile("servroot2/logs/error.log")
>   assert(helpers.signal("servroot1", "-KILL"))
>   need_exit = false
>   assert.logfile("servroot2/logs/error.log").has.line("some kind of error string")
> end)
> ```

This is intentionally added. When we receive the closing frame, there could be an RPC call already fired, and blocking at `receive()`. It will see the error before the flag is set to closing. Of course, we should integrate the `close` handling with the `receive()` call. But this is just a compromise before we have a thorough refactoring.

And it seems we do not define details on what happens to a disconnected wRPC peer, like should it responds to all RPC calls not answered before closing, or send an error msg indicating those RPC calls are given up.

spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
Suika-Kong: > 
We can not tell if a wRPC error is really unexpected, so most of the time they are at notice level, and the caller could decide whether it should be an error. In this case, it should not be an error.

kong/tools/wrpc/threads.lua
flrgh: Sounds reasonable enough for me. Thanks!

spec/02-integration/09-hybrid_mode/02-start_stop_spec.lua
Suika-Kong: done

mayocream (approved):

PR #9242, opened 2022-08-15
feat(router/atc) incremental atc router

chronolaw:
### Summary

Compare new routes and previous router then generate atc-router.

* check `ngx.null`
* compare based on `updated_at` `priority` `expression`
* dao will update field `expression` when insert/update/upsert.


kong/router/atc_compat.lua
```diff
@@ -361,30 +361,54 @@ local function route_priority(r)
...
-  local s = get_schema()
-  local inst = router.new(s)
+  if is_traditional_compatible then
+    atc = get_atc(route)
+    priority = route_priority(route)
 
-  if not cache then
-    cache = lrucache.new(MATCH_LRUCACHE_SIZE)
-  end
+    local atc_priority = tonumber(priority)
```
ADD-SP, line 373: We should cache function `tonumber` at the module level.

kong/router/atc_compat.lua
```diff
@@ -361,30 +361,54 @@ local function route_priority(r)
...
+  else
+    atc = route.atc
+    priority = route.priority
+
+    local gen = gen_for_field("net.protocol", OP_EQUAL, route.protocols)
+    if gen then
+      atc = atc .. " && " .. gen
+    end
+
+    -- TODO: check if route is changed
```
ADD-SP, line 397: Should we remove this line?

bungle:
Can we use these:
https://github.com/Kong/kong/blob/master/kong/db/schema/entities/routes.lua#L15-L16

To detect if even update is needed?

chronolaw:
> Can we use these: https://github.com/Kong/kong/blob/master/kong/db/schema/entities/routes.lua#L15-L16
> 
> To detect if even update is needed? Or to process rows where update_time/created_at is greater than previous build max(update_time)/max(created_at).

Now Router uses all routes queried from DB, I think in the future we can change to query new or updated records.

spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua
```diff
@@ -182,6 +182,9 @@ for _, strategy in helpers.each_strategy() do
       end)
 
       it("on update", function()
+        -- router(traditional_compatible) incremental build use timpstamp to check update
```
bungle, line 185:
```suggestion
        -- router(traditional_compatible) incremental build use timestamp to check update
```

kong/router/atc_compat.lua
```diff
@@ -361,30 +362,42 @@ local function route_priority(r)
...
-    cache_neg = lrucache.new(MATCH_LRUCACHE_SIZE)
+  if is_update then
+    inst:remove_matcher(route_id)
   end
 
+  assert(inst:add_matcher(priority, route_id, atc))
+end
+
+
+local function new_from_srcatch(routes, is_traditional_compatible)
```
bungle, line 393:
```suggestion
local function new_from_scratch(routes, is_traditional_compatible)
```

kong/router/atc_compat.lua
```diff
@@ -418,12 +419,93 @@ function _M.new(routes, cache, cache_neg)
...
+  end
+
+  local is_traditional_compatible =
+          kong and kong.configuration and
+          kong.configuration.router_flavor == "traditional_compatible"
+
+  local router, err
+
+  if not pre_router then
+    router, err = new_from_srcatch(routes, is_traditional_compatible)
```
bungle, line 492:
```suggestion
    router, err = new_from_scratch(routes, is_traditional_compatible)
```

spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua
chronolaw: thanks

kong/router/atc_compat.lua
chronolaw: done

kong/router/atc_compat.lua
chronolaw: done

kong/router/atc_compat.lua
```diff
@@ -428,12 +429,93 @@ function _M.new(routes, cache, cache_neg)
...
     }, _MT)
 end
 
 
+local function is_route_changed(a, b)
+  return a.updated_at ~= b.updated_at
+end
+
+
+local function new_from_previous(routes, is_traditional_compatible, pre_router)
```
dndx, line 441:
```suggestion
local function new_from_previous(routes, is_traditional_compatible, old_router)
```

`old_` is probably a better prefix than `pre_`.

kong/router/atc_compat.lua
```diff
@@ -428,12 +429,93 @@ function _M.new(routes, cache, cache_neg)
...
+  -- create or update routes
+  for _, r in ipairs(routes) do
+    local route = r.route
+    local route_id = route.id
+
+    if not route_id then
+      return nil, "could not categorize route"
+    end
+
+    route.touched = true
```
dndx, line 455:
```suggestion
    route.seen = true
```

Probably a better name.

dndx:
@bungle I am wondering if we can use the `atc` field for change detection. Basically, we make the `atc` field available in compatible mode, use write transformations to generate the ATC DSL on each entity update. However, it appears the transformation will not run if any field in the `input` table is empty (this includes empty Lua table). Is there another way we can do this transformation unconditionally aside from hacking the Admin API endpoint?

bungle:
> Now I use created_at and updated_at to detect route change, but it is second resolution, if a route changes many times in one second, we can not notice it.

If the `sec` is same, can you in that case revert to previous change detection alg? Should be rare.

bungle:
> @bungle I am wondering if we can use the `atc` field for change detection. Basically, we make the `atc` field available in compatible mode, use write transformations to generate the ATC DSL on each entity update. However, it appears the transformation will not run if any field in the `input` table is empty (this includes empty Lua table). Is there another way we can do this transformation unconditionally aside from hacking the Admin API endpoint?

@dndx, I have been thinking about it. I think this is soon going to be generic issue, if we continue to make other things incremental (as I think we should, like plugins iterator and balancer). So I was thinking should we add:

Thus I think this might be better solved by adding say `revision` field to each entity.

```lua
fields = {
  { revision = typedefs.revision }
}
```

Where revision is just a BIGINT or something that is increased on every update.

dndx:
@bungle `revision` is an easy way out, but it might be more confusing for the users. Unless we have a good way to hide it from user's view. Maybe a read transformation that sets it to `ngx.null`?

Until we can figure out how we want to properly revision entities, this leaves the option to change the implementation open for later.

dndx:
@bungle Is there any easy way we can implement this "increment by 1" operation? Using `write` transformations seems to be prone to race conditions.

I still feel that writing the generated DSL to the DB makes more sense, it seems to be similar amount of work, and helps the customer on understanding how to migrate their existing routes to the DSL correctly.

kong/router/atc_compat.lua
```diff
@@ -259,6 +259,7 @@ local function get_atc(route)
 
   return tb_concat(out, " && ")
 end
+local get_atc = _M.get_atc
```
chronolaw, line 262: I prefer `_M.get_atc = get_atc`.

kong/db/dao/routes.lua
```diff
@@ -0,0 +1,81 @@
...
+function _Routes:insert(route, options)
+  if route.expression then
+    local err_t = self.errors:schema_violation({
+      expression = ERR_READONLY,
+    })
+
+    return nil, tostring(err_t), err_t
+  end
+
+  route.expression = get_atc(route)
```
chronolaw, line 27: `route.expression` is not enough, we also need `route.priority`.

kong/router/atc_compat.lua
```diff
@@ -430,12 +432,93 @@ function _M.new(routes, cache, cache_neg)
...
       services = services_t,
       fields = inst:get_fields(),
-      cache = cache,
-      cache_neg = cache_neg,
     }, _MT)
 end
 
 
+local function is_route_changed(a, b)
+  return (a.expression ~= b.expression) or (a.updated_at ~= b.updated_at)
```
chronolaw, line 440: `a.updated_at ~= b.updated_at` should be the first condition.

kong/router/init.lua
```diff
@@ -30,7 +30,7 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
 end
 
 
-function _M.new(routes, cache, cache_neg)
+function _M.new(routes, cache, cache_neg, pre_router)
```
chronolaw, line 33: `pre_router` rename to `old_router`

chronolaw:
It seems that we lost some rebased commits, such as yielding in loop.

kong/router/atc_compat.lua
```diff
@@ -372,30 +370,46 @@ local function route_priority(r)
 end
 
 
-function _M.new(routes, cache, cache_neg)
-  if type(routes) ~= "table" then
-    return error("expected arg #1 routes to be a table")
-  end
+local function add_atc_matcher(inst, route, route_id,
+                               is_traditional_compatible,
+                               remove_existig)
```
chronolaw, line 375: What does `remove_existig` mean？ Is it a typo?

spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua
```diff
@@ -182,6 +182,9 @@ for _, strategy in helpers.each_strategy() do
       end)
 
       it("on update", function()
+        -- router(traditional_compatible) incremental build use timestamp to check update
+        ngx.sleep(1)
```
chronolaw, line 186: We can remove this line because now we do not depend on timestamp.

kong/router/atc_compat.lua
```diff
@@ -430,12 +432,93 @@ function _M.new(routes, cache, cache_neg)
...
+
+    if not old_route then
+      -- route is new
+      add_atc_matcher(inst, route, route_id, is_traditional_compatible, false)
+
+    elseif is_route_changed(route, old_route) then
+      -- route has existed
+      add_atc_matcher(inst, route, route_id, is_traditional_compatible, true)
+    end
+
```
chronolaw, line 473: Should add `yield(true)` here.

kong/router/atc_compat.lua
```diff
@@ -430,12 +432,93 @@ function _M.new(routes, cache, cache_neg)
...
+  -- remove routes
+  for id, r in pairs(old_routes) do
+    if r.touched  then
+      r.touched = nil
+
+    else
+      inst:remove_matcher(id)
+      old_routes[id] = nil
+    end
```
chronolaw, line 484: Should add `yield(true)` here.

kong/db/dao/routes.lua
```diff
@@ -0,0 +1,83 @@
+-- NOTE: this DAO is not enabled when router_flavor = expressions, see schema/entities/routes.lua
+
+
+local get_atc = require("kong.router.atc_compat").get_atc
+local constants = require("kong.constants")
+
+
+local _Routes = {}
```
flrgh, line 8: Just an FYI: Enterprise has an existing custom DAO for routes, so you will likely need to solve some conflicts when merging this over there.

dndx:
@bungle Could you take a look at this PR as well?

This change to use `expression` for change detection was much harder than I anticipated. The addition of field broke a lot of test cases and our current schema transformation support is inadequate for this kind of use case, so I had to write a lot of custom DAO code in order to handle it. But I think the current change is should be sufficient for 3.0.

If we can make transformations be more flexible and run even when some fields are absent, this could be a much simpler feature to add. Might be something worth considering after 3.0.

spec/02-integration/06-invalidations/02-core_entities_invalidations_spec.lua
dndx: Let's fix it.

dndx (approved):

PR #9254, opened 2022-08-16
feat(cmd) clean dangling unix sockets on startup

flrgh:
## Summary

This change set makes Kong more likely to recover from an unclean shutdown.

It also includes a bugfix for `kong start` wherein the `.kong_env` file could be rewritten (among other things) if Kong is already running.

## Background

When Nginx suffers an unclean shutdown (HW issue, OS issue, `SIGKILL`, etc), it may fail to clean up any listening unix sockets from the filesystem. The presence of these will cause Kong to fail to start back up again:

> [emerg] 2127#0: bind() to unix:/usr/local/kong/worker_events.sock failed (98: Address in use)

## Solution

Kong will search for any existing unix sockets within the prefix directory and remove them before attempting to start nginx. If Kong/nginx is already running, no action is taken.

## Limits

The search for dangling sockets is performed only in the top-level prefix directory (no recursion), so this will become less effective if at some point we add a unix socket in a nested directory. The integration tests I've added would _likely_ catch this condition, unless said socket is not created by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -367,6 +367,9 @@
   [#8497](https://github.com/Kong/kong/pull/8497) [9265](https://github.com/Kong/kong/pull/9265)
 - Improved error handling and debugging info in the DNS code
   [#8902](https://github.com/Kong/kong/pull/8902)
+- Kong will now attempt to recover from an unclean shutdown by detecting and
+  removing dangling unix sockets in the prefix directory
+  [#9254](https://github.com/Kong/kong/pull/9254)
 
 #### Admin API
 
@@ -500,6 +503,8 @@
   [#9255](https://github.com/Kong/kong/pull/9255)
 - Fixed an issue where cache entries of some entities were not being properly invalidated after a cascade delete
   [#9261](https://github.com/Kong/kong/pull/9261)
+- Running `kong start` when Kong is already running will no longer clobber
+  the existing `.kong_env` file [#9254](https://github.com/Kong/kong/pull/9254)
 
 
 #### Admin API
--- a/kong/cmd/start.lua
+++ b/kong/cmd/start.lua
@@ -6,7 +6,42 @@ local kong_global = require "kong.global"
 local kill = require "kong.cmd.utils.kill"
 local log = require "kong.cmd.utils.log"
 local DB = require "kong.db"
+local lfs = require "lfs"
 
+
+local function is_socket(path)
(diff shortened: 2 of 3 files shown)


dndx:
Any reason we do not turn this feature on by default?

flrgh:
> Any reason we do not turn this feature on by default?

The primary risk I can think of is that if the master pid file is lost at runtime, `kong start` won't detect that nginx is already running. Without `--cleanup-sockets`, `kong start` will just fail. _With_ `--cleanup-sockets`, `kong start` will actively break the current instance that is running by deleting all of the unix sockets before failing.

That's all edge case territory, but losing a pid file to logrotate or somebody's errant filesystem cleanup cron is definitely not unheard of. Docker/systemd will run with daemon=off, so losing track of the nginx master pid is not really much of a worry there.

dndx:
@flrgh Hmm, I have been thinking about it, if the master PID file is lost, Nginx will fail to `bind` anyway right? So we can not start new instance in that case. Also a new `.kong_env` could be written and a lot of things might break in the meantime (not just this socket file).

flrgh:
Weirdly enough, there's nothing to prevent us from overwriting `.kong_env` currently. Seems like these two lines should probably be swapped:

https://github.com/Kong/kong/blob/d6307f2843a2aa63e785d4aa36aad277658e8f7c/kong/cmd/start.lua#L24-L27

dndx:
@flrgh Discussed with the team, let's enable this by default and make it into 3.0 as a feature (with a bit more elaborate warning when dangling sockets are detected and removed).

flrgh:
> @flrgh Discussed with the team, let's enable this by default and make it into 3.0 as a feature (with a bit more elaborate warning when dangling sockets are detected and removed).

Okie doke, just to make sure I understand: no `--cleanup-sockets` flag then?

tyler-ball:
No `--cleanup-sockets` flag, this should be the default behavior

flrgh:
@dndx okay, this is ready for review now.

I have also added a fix for the problem I mentioned [here](https://github.com/Kong/kong/pull/9254#issuecomment-1220876531) and added a test case for it. We should not be altering the prefix directory in `kong start` if Kong is already running. While I feel it's an important fix that is not very intrusive, it's not explicitly tied to the unix socket issue, so let me know if I should take it out and/or put it in a separate PR.

mayocream (approved):

PR #9258, opened 2022-08-17
datadog-plugin: Set metrics.required to be false

donna-156:
### Summary

The Datadog plugin is configured to set `DEFAULT_METRICS`; meaning you don't have to manually configure any metrics to send to Datadog (which is great!). However, when the schema sets `required = true` alongside a default value, the Admissions Controller will fail with the following error if no metrics are provided: 

```
Error: UPGRADE FAILED: cannot patch "global-datadog" with kind KongClusterPlugin: admission webhook "validations.kong.konghq.com" denied the request: plugin failed schema validation: schema violation (config.metrics: required field missing)
```

In order to use the default Datadog metrics values and the Admissions Controller in parallel, this value needs to be false.

### Full changelog

* [Datadog Plugin] Update Datadog schema, to set the config for metrics.required to false

### Issue reference

Instead of creating a new issue, I've raised this PR to address the issue myself.

### DIFF
--- a/kong/plugins/datadog/schema.lua
+++ b/kong/plugins/datadog/schema.lua
@@ -81,7 +81,7 @@ return {
           { consumer_tag = { type = "string", default = "consumer" }, },
           { metrics = {
               type     = "array",
-              required = true,
+              required = false,
               default  = DEFAULT_METRICS,
               elements = {
                 type = "record",


donna-156:
Please disregard the temporary closing of this PR; I merged an internal PR that referenced this PR and Git did some unexpected magic.

Tieske:
What's the usecase of configuring the plugin without metrics? then maybe not configure it at all?

donna-156:
> What's the usecase of configuring the plugin without metrics? then maybe not configure it at all?

When you configure the plugin without metrics, it will configure the default metrics. This would mean the config looks like so, which is much cleaner and ensures we do not lose out on any newly introduced metrics:

```
kind: KongClusterPlugin
metadata:
  name: global-datadog
  annotations:
    kubernetes.io/ingress.class: kong
  labels:
    global: "true"
config:
  host: <host>
  port: 8125
  metrics:
  prefix: <prefix> 
  service_name_tag: kong_upstream
plugin: datadog
``` 

However, because of the schema for metrics being set to `required = true`, despite it setting the default value for those metrics anyway the Kong Admissions Controller will result in an error as stated in the description of the PR. 

Our config has had to be set as following:

```
kind: KongClusterPlugin
metadata:
  name: global-datadog
  annotations:
    kubernetes.io/ingress.class: kong
  labels:
    global: "true"
config:
  host: <host>
  port: 8125
  metrics:
  prefix: <prefix> 
  service_name_tag: kong_upstream
  # Defaults copied from: https://github.com/Kong/kong/blob/2.8.1/kong/plugins/datadog/schema.lua#L28-L66
  metrics:
    - name: request_count
      stat_type: counter
      sample_rate: 1
      consumer_identifier: custom_id
      tags:
        - app:{{ . }}
    - name: latency
      stat_type: timer
      consumer_identifier: custom_id
      tags:
        - app:{{ . }}
    - name: request_size
      stat_type: timer
      consumer_identifier: custom_id
      tags:
        - app:{{ . }}
    - name: response_size
      stat_type: timer
      consumer_identifier: custom_id
      tags:
        - app:{{ . }}
    - name: upstream_latency
      stat_type: timer
      consumer_identifier: custom_id
      tags:
        - app:{{ . }}
    - name: kong_latency
      stat_type: timer
      consumer_identifier: custom_id
      tags:
        - app:{{ . }}
plugin: datadog
``` 

Aside from this, it's generally not recommended to require a value to be passed when defaults are configured. My change shouldn't break anything for others, as anyone who has configured the metrics themselves to date will have them configured, and anyone who doesn't will continue to have the defaults configured.

This will mean that anyone using the Datadog plugin is able to also use the Kong Admissions Controller in line with the Datadog default metrics and keep their plugin config clean.

Tieske:
what's the "Kong Admissions Controller", do you mean the Ingress Controller?

Looks like a bug on that end to me. The Kong managament API will accept the empty setting and fill in the defaults.

donna-156:
> what's the "Kong Admissions Controller", do you mean the Ingress Controller?
> Looks like a bug on that end to me. The Kong managament API will accept the empty setting and fill in the defaults.

The Admissions Controller ships with the Ingress Controller, and it validates plugin schema. [[Kong Admission Controller docs](https://docs.konghq.com/kubernetes-ingress-controller/latest/deployment/admission-webhook/)]

Admissions Controller will read the schema of a plugin, such as the Datadog plugin in this case, and prevent a plugin from being deployed with invalid config to a Kubernetes cluster.

Currently, the Datadog plugin schema is incorrectly configured which is what my change is trying to correct. It is not a bug on the Admissions Controller side, as it is behaving as expected based on the Datadog plugin schema. 

Based on the current configuration, you must configure metrics in your plugin config, despite there being default values available. This is an anti-pattern. 

If there are default values available, then the schema should not require that value to be set. That's what my change is doing.

rainest:
The field is still required despite the default because that's how it prevents subsequent updates from deleting it AFAIK. It's irrelevant on create actions, but relevant for updates.

This is another instance of https://github.com/Kong/kong/issues/8093. The admission webhook POSTs the proposed configuration to the admin API's `/schemas/plugins/validate`, which returns if true if the configuration would be accepted or false if it would not be.

[That endpoint](https://github.com/Kong/kong/blob/34e2ec189db51d77fea6175c2daab7973e3f2416/kong/api/routes/kong.lua#L30-L43) is not able to fill in default fields, and [the subsequent validation](https://github.com/Kong/kong/blob/34e2ec189db51d77fea6175c2daab7973e3f2416/kong/db/schema/init.lua#L1876-L1942) does not allow them to be empty (sorta? there's the `full_check` option but I'm not sure of its intended use--it looks more intended to ignore `required` entirely in some cases, default or no). That's what would need to change to allow this.

That comes with its own problem, however: the update case still matters, and the endpoint can't distinguish between a new or modified object. AFAIK simply populating the default would allow you to create (at least on the Kubernetes side) invalid configuration that will break updates. I think we could remedy this with separate POST/PUT/PATCH handlers that mimic their admin API counterparts (POST populates defaults, the others do not). This might be a bit weird since you'd be PATCHing nothing, but it'd hopefully work.

donna-156:
Thank you @rainest for the additinoal context. 

Given that, I'll close this PR. Is there a plan to visit the issue you linked to solve the underlying issue?

PR #9260, opened 2022-08-17
Release: 3.0.0-alpha.1

hanshuebner:


scripts/make-release
```diff
@@ -1,30 +1,41 @@
 #!/usr/bin/env bash
 
-source "$(dirname "$0")/common.sh"
```
kikito, line 3: Perhaps rename this file to `release.sh`?

scripts/release-lib.sh
```diff
@@ -52,8 +54,8 @@ function yesno() {
 
 #-------------------------------------------------------------------------------
 function check_milestone() {
-  if yesno "Visit the milestones page (https://github.com/Kong/kong/milestone) and ensure PRs are merged. Press 'y' to open it or Ctrl-C to quit"; then
-    $browser https://github.com/Kong/kong/milestones
+  if yesno "Visit the milestones page (https://github.com/$GITHUB_ORG/kong/milestone) and ensure PRs are merged. Press 'y' to open it or Ctrl-C to quit"; then
```
kikito, line 57: Probably a lot of these functions can go back to the release.sh file, now that there's only one file and there's no need to share functionality across 3 files any more

scripts/make-release
kikito: Jenkins will bark at the filename change; I am almost certain it hardcodes script/make-patch-release somewher

kikito (approved):
Approved with comments

scripts/make-release
hanshuebner: I've updated the Jenkinsfile to fix this.  I've thought about moving everything from `common.sh` to `make-release` but decided against it for now, as make-release neatly contains only the top-level code.

scripts/release-lib.sh
hanshuebner: I agree.  If the split between top-level code in `make-release` and implementation in `release-lib.sh` becomes too bothersome, we can still merge the two into one.

PR #9263, opened 2022-08-18
chore(router) rename atc to expression

chronolaw:
### Summary

 rename 'atc' to 'expression'

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -436,7 +436,7 @@ local CONF_INFERENCES = {
     }
   },
   router_flavor = {
-    enum = { "traditional", "traditional_compatible", "atc" },
+    enum = { "traditional", "traditional_compatible", "expression" },
   },
   worker_state_update_frequency = { typ = "number" },
 
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -3,7 +3,7 @@ local atc = require("kong.router.atc")
 local router = require("resty.router.router")
 
 
-if kong and kong.configuration and kong.configuration.router_flavor == "atc" then
+if kong and kong.configuration and kong.configuration.router_flavor == "expression" then
   return {
     name         = "routes",
     primary_key  = { "id" },
--- a/kong/db/schema/entities/routes_subschemas.lua
+++ b/kong/db/schema/entities/routes_subschemas.lua
@@ -66,7 +66,7 @@ local grpc_subschema = {
 }
 
 
-if kong and kong.configuration and  kong.configuration.router_flavor == "atc" then
+if kong and kong.configuration and  kong.configuration.router_flavor == "expression" then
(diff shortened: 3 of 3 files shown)


kong/db/schema/entities/routes.lua
```diff
@@ -50,7 +50,7 @@ if kong and kong.configuration and kong.configuration.router_flavor == "atc" the
 
           local res, err = r:add_matcher(0, entity.id, entity.atc)
           if not res then
-            return nil, "DSL failed validation: " .. err
+            return nil, "expression failed validation: " .. err
```
dndx, line 53:
```suggestion
            return nil, "Router Expression failed validation: " .. err
```

dndx (approved):

kong/db/schema/entities/routes.lua
```diff
@@ -37,20 +37,20 @@ if kong and kong.configuration and kong.configuration.router_flavor == "atc" the
       { response_buffering  = { type = "boolean", required = true, default = true }, },
       { tags             = typedefs.tags },
       { service = { type = "foreign", reference = "services" }, },
-      { atc = { type = "string", required = true }, },
+      { expressions = { type = "string", required = true }, },
```
mayocream, line 40: `expressions` or `expression`? since it is a `type=string` field.

chronolaw: rename to `expression`

mayocream (approved):

bungle:
The `ATC` is still mentioned here:
https://github.com/Kong/kong/blob/master/.requirements#L11

chronolaw:
> The `ATC` is still mentioned here: https://github.com/Kong/kong/blob/master/.requirements#L11

I think the normal user can not see this file, and we will not change atc-router's repo name now.

bungle:
What about:
https://github.com/Kong/kong/blob/master/kong/router/atc_compat.lua#L414

chronolaw:
> What about: https://github.com/Kong/kong/blob/master/kong/router/atc_compat.lua#L414

Good catch. I will fix it.

kong/conf_loader/init.lua
```diff
@@ -436,7 +436,7 @@ local CONF_INFERENCES = {
     }
   },
   router_flavor = {
-    enum = { "traditional", "traditional_compatible", "atc" },
+    enum = { "traditional", "traditional_compatible", "expression" },
```
dndx, line 439:
```suggestion
    enum = { "traditional", "traditional_compatible", "expressions" },
```

dndx (changes requested):
`router_flavor` should be `expressions`, Route entity should use `expression` as field name. The plural is important.

kong/conf_loader/init.lua
chronolaw: done

dndx (approved):

PR #9270, opened 2022-08-18
fix: allow grpcs and tls for client_certificate

shaneutt:
### Summary

A [discussion in kong/kubernetes-ingress-controller](https://github.com/Kong/kubernetes-ingress-controller/discussions/2831) wherein a user was trying to use mTLS with gRPC indicated a bug in the Kong Gateway where `grpcs` wasn't allowed to have a `client_certificate`, this PR attempts to fix that (and also add `tls` while we're at it for similar reasons).

### Full changelog

* Fixed a bug where `grpcs` and `tls` could not be used with `client_certificate` in services

### Issue reference

Fix for https://github.com/Kong/kubernetes-ingress-controller/discussions/2831

### DIFF
--- a/kong/db/schema/entities/services.lua
+++ b/kong/db/schema/entities/services.lua
@@ -53,7 +53,7 @@ return {
                       then_field = "path",
                       then_match = { eq = null }}},
     { conditional = { if_field = "protocol",
-                      if_match = { ne = "https" },
+                      if_match = { not_one_of = { "https", "grpcs", "tls" }},
                       then_field = "client_certificate",
                       then_match = { eq = null }}},
     { conditional = { if_field = "protocol",


shaneutt:
I could use some help in verifying this solution according to the problem that was seen in https://github.com/Kong/kubernetes-ingress-controller/discussions/2831.

flrgh (changes requested):
The schema change looks correct to me :+1: @shaneutt can you please rebase and add a tests for the `grpc` and `tls` use cases? These types of connections/requests are handled very differently from regular `https` requests within Kong (and nginx for that matter), so I'm not 100% certain they will work without requiring any other code changes.

shaneutt:
Thanks @flrgh :+1: 

I'm going to temporarily close this, as in the time since I originally made this PR I've changed teams. The intention right now is that someone from my previous team will be picking this back up in the future.

PR #9275, opened 2022-08-19
chore(*) increase events queue len to 1024*50

chronolaw:
### Summary

increase events queue len to 1024*50, it may reduce the probability of queue overflow.

See FT-3214.


bungle:
> it may reduce the probability of queue overflow

or move it later? have we hit this issues, and how can we reproduce?

ADD-SP:
> > it may reduce the probability of queue overflow
> 
> or move it later? have we hit this issues, and how can we reproduce?

Please see FT-3214.

bungle (approved):

PR #9276, opened 2022-08-19
chore(conf) lower bound to worker_connections and worker_rlimit_nofile

bungle:
### Summary

Adds lower bound of 1024 to:

- `nginx_events_worker_connections=auto`
- `nginx_main_worker_rlimit_nofile=auto`

### DIFF
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -253,6 +253,7 @@ local function compile_conf(kong_config, conf_template)
       end
 
       value = math.min(value, 16384)
+      value = math.max(value, 1024)
 
       if worker_rlimit_nofile_auto then
         worker_rlimit_nofile_auto.value = value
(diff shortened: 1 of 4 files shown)


kong/cmd/utils/prefix_handler.lua
```diff
@@ -253,6 +253,7 @@ local function compile_conf(kong_config, conf_template)
       end
 
       value = math.min(value, 16384)
+      value = math.max(value, 1024)
```
ADD-SP, line 256: We'd better show a warning to our users, and then they can adjust the `ulimit` (if needed)

ADD-SP (changes requested):

chronolaw (approved):

kong/cmd/utils/prefix_handler.lua
bungle: @ADD-SP there is already a warning of too low `ulimit`, but this is to make that Kong can at least start.

ADD-SP (approved):

PR #9281, opened 2022-08-20
fix(tracing) balancer span_kind fix

shettyh:
### Summary

Balancer span creation was using the incorrect param name for span kind.

### Full changelog

* [tracing] fixed balancer span kind


mayocream (approved):
Thanks!

PR #9283, opened 2022-08-22
chore(test) Actually test upgrade path with cassandra

hanshuebner:
Previously, upgrades were tested with PostgreSQL twice due to copy&paste error.


ADD-SP:
Should we use the matrix to speed up the upgrade tests?

ADD-SP (approved):
LGTM, and I think we should speed up it using matrix in the future.

PR #9288, opened 2022-08-23
perf(router/atc)yield in build router

chronolaw:
### Summary

Add yielding in build atc-router


mayocream (approved):

mayocream:
@chronolaw Should this be tagged as `perf` instead of `feat`?

ADD-SP (approved):

PR #9290, opened 2022-08-23
chore(router) deprecate setting path handling v1 when router_flavor is traditional_compatible

windmgc:
### Summary

This PR ~~denies~~ users from setting `path_handling=v1` when the Kong gateway is running with `router_flavor=traditional_compatible`, as the `path_handling` feature is stepping into deprecation and path_handling v0 will be the default behaviour when composing upstream_uri.

Discussed with the team and decide to only print deprecation log in the current version.

### Full changelog

* Add custom entity check function to deny setting `path_handling=v1` when `router_flavor` is `traditional_compatible`

### DIFF
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -118,17 +118,6 @@ else
       { service = { type = "foreign", reference = "services" }, },
     },
 
-    transformations = {
-      {
-        input = { "path_handling" },
-        on_read = function(path_handling)
-          if kong_router_flavor == "traditional_compatible" then
-            return { path_handling = ngx.null }
-          end
-        end,
-      },
-    },
-
     entity_checks = {
       { conditional = { if_field = "protocols",
                         if_match = { elements = { type = "string", not_one_of = { "grpcs", "https", "tls", "tls_passthrough" }}},
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -125,7 +125,7 @@ else
       { custom_entity_check = {
         field_sources = { "path_handling" },
         fn = function(entity)
-          if entity.path_handling == "v1" and kong.configuration.router_flavor == "traditional_compatible" then
+          if entity.path_handling == "v1" and kong.configuration.router_flavor ~= "traditional" then
             return nil, "path_handling = 'v1' is deprecated and not supported with router_flavor = 'traditional_compatible'"
           end
           return true
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -2,8 +2,9 @@ local typedefs = require("kong.db.schema.typedefs")
 local atc = require("kong.router.atc")
 local router = require("resty.router.router")
 
+local kong_router_flavor = kong and kong.configuration and kong.configuration.router_flavor
 
-if kong and kong.configuration and kong.configuration.router_flavor == "expressions" then
+if kong_router_flavor == "expressions" then
   return {
     name         = "routes",
     primary_key  = { "id" },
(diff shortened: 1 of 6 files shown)


chronolaw:
Good Job. 

But I think both `traditional_compatible` and `expression` don't support `v1`, so we should change the condition to `kong.configuration.router_flavor ~= "traditional"`

windmgc:
@chronolaw Do you mean `router_flavor = atc`?

I saw that router_flavor only have three enum results
https://github.com/Kong/kong/blob/9c990a79083f1b529eac69238552ec55f59564ed/kong/conf_loader/init.lua#L438-L440

And when `router_flavor = atc` the schema will be a different table without the `path_handling` 
https://github.com/Kong/kong/blob/9c990a79083f1b529eac69238552ec55f59564ed/kong/db/schema/entities/routes.lua#L6-L42

chronolaw:
Now `atc` will rename to `expression`, see #9263 .

kong/db/schema/entities/routes.lua
```diff
@@ -122,6 +122,15 @@ else
                         then_match = { len_eq = 0 },
                         then_err = "'snis' can only be set when 'protocols' is 'grpcs', 'https', 'tls' or 'tls_passthrough'",
                       }},
+      { custom_entity_check = {
+        field_sources = { "path_handling" },
+        fn = function(entity)
+          if entity.path_handling == "v1" and kong.configuration.router_flavor ~= "traditional" then
+            return nil, "path_handling = 'v1' is deprecated and not supported with router_flavor = 'traditional_compatible'"
```
chronolaw, line 129: We should also change this error message.

chronolaw:
I think it is not very urgent for 3.0, because the new router only supports `v0`.

windmgc:
Yes, agree, not a blocker for 3.0, so no hurry for merging.
And I still need to get the tests fixed :P

kong/db/schema/entities/routes.lua
windmgc: done

kong/db/schema/entities/routes.lua
```diff
@@ -126,7 +126,7 @@ else
         field_sources = { "path_handling" },
         fn = function(entity)
           if entity.path_handling == "v1" and kong.configuration.router_flavor ~= "traditional" then
-            return nil, "path_handling = 'v1' is deprecated and not supported with router_flavor = 'traditional_compatible'"
+            return nil, "path_handling = 'v1' is deprecated and not supported when router_flavor is not 'traditional'"
```
chronolaw, line 129: How about `only supported when router_flavor is 'traditional'`

windmgc: done

chronolaw (approved):

kong/db/schema/entities/routes.lua
```diff
@@ -122,6 +125,23 @@ else
                         then_match = { len_eq = 0 },
                         then_err = "'snis' can only be set when 'protocols' is 'grpcs', 'https', 'tls' or 'tls_passthrough'",
                       }},
-                    },
+      { custom_entity_check = {
+        field_sources = { "path_handling" },
+        fn = function(entity)
+          if entity.path_handling == "v1" then
+            if kong_router_flavor == "traditional" then
+              deprecation("path_handling='v1' is deprecated and will be removed in future version, " ..
```
mayocream, line 133: Shall we add a deprecation after version and a removal version?
https://github.com/Kong/kong/blob/e2b048845429694f009b73d3a482cd0182af494c/kong/conf_loader/init.lua#L681-L685

kong/db/schema/entities/routes.lua
```diff
@@ -122,6 +125,23 @@ else
...
+                          "please use path_handling='v0' instead")
+
+            elseif kong_router_flavor == "traditional_compatible" then
+              deprecation("path_handling='v1' is deprecated and will not work under traditional_compatible " ..
```
mayocream, line 137: ditto

mayocream (approved):

PR #9302, opened 2022-08-24
Revert "fix(clustering) DP errors when CP exits"

mayocream:
This reverts commit a51d298d920f8e662f923524128dc69599b62224.

For errors were thrown when CP exits.
We need to rework this later then.

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -176,12 +176,5 @@ function _M:init_worker()
   end
 end
 
-function _M:exit_worker()
-  if self.conf.role == "control_plane" then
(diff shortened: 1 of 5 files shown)


tyler-ball:
What is the motivation for the revert? Did it not improve the log noise?

mayocream:
cc @Suika-Kong

Suika-Kong:
That PR is using the wrong way to close connections. We cannot create threads in `exit_worker`, and that causes an error when CP exits.

Suika-Kong (approved):

chronolaw (approved):

PR #9303, opened 2022-08-24
chore(router) router expression name

chronolaw:
### Summary

`expressions` make a little confused, we rename it to `expression`. or `expressional `?

We should have more discussion of it.

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -436,7 +436,7 @@ local CONF_INFERENCES = {
     }
   },
   router_flavor = {
-    enum = { "traditional", "traditional_compatible", "expressions" },
+    enum = { "traditional", "traditional_compatible", "expression" },
   },
   worker_state_update_frequency = { typ = "number" },
 
--- a/kong/db/schema/entities/routes.lua
+++ b/kong/db/schema/entities/routes.lua
@@ -3,7 +3,7 @@ local atc = require("kong.router.atc")
 local router = require("resty.router.router")
 
 
-if kong and kong.configuration and kong.configuration.router_flavor == "expressions" then
+if kong and kong.configuration and kong.configuration.router_flavor == "expression" then
   return {
     name         = "routes",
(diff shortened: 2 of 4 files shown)


bungle (approved):

bungle:
`router_flavor`:´traditional´ /  `traditional_compatible`:
use: route.paths, route.headers, route.hosts, etc.

`router_flavor`: `expression`:
use: route.expression

The `route.paths` etc. are quite self-explanatory but `route.expression` feels like `where is the manual`. Ok, there is manual and we figure out you can use a new language, basically a DSL.

When debugging routing issues, we might ask `what router are you using?`. When the answer is `I am using router expression`, it feels strange. Don't all the routers have all sorts of expressions (isn't route.paths used inside somekind of expression inside router)? Or did you mean Router named Expression? A different flavor?

In entity the `route.expression` or `route.match_dsl` or `route.match` or `route.dsl` or `route.rules` or `route.match_rules` could be possible names. But `router_flavor` is different beast. How to call these different routers? Expression Based Router sounds better than Router Expression. DSL Router / Traditional Router and then we have Traditional Compatible Router too, which means `DSL Router that is initialized from Traditional Route Fields, like route.paths`. This is all really confusing.

chronolaw:
The type of `route.expression` is `string`, not `array`, so it is ok.

mheap:
@bungle Datong and I discussed calling the new router `dsl` and Datong felt like it was not accessible for a wider audience.

`expression` is the in-code name. In text, we'll always talk about `router expression` or `router expressions` depending on context

bungle:
@mheap I have approved this, do you think we should merge it? @chronolaw, @dndx?

dndx:
@bungle Under debate, do NOT merge it yet.

dndx:
Discussed with @mheap , this PR is closed because it follows existing conventions in other Kong features. e.g. Routes vs Route. Certificates vs certificate.

PR #9309, opened 2022-08-24
perf(router) dynamic router lru cache size based on number of routes

bungle:
### Summary

Implements dynamic router LRU cache size based on number of routes.

- By default we initialize positive router LRU cache of 5000 items.
- If there is more than 5000 routes we reinitialize cache to closest 5000, e.g. if there is 5001 routes, we re-initialize cache to 10000 items and keep the size 10000 until number of routes reaches either over 10000 or less than 5001.
- The maximum size of cache is limited to 100000 items.

This should improve our routing performance with high number of routes when routes are found in cache. This will increase memory usage though, but it is expected for users to have beefier systems when number of routes is high (again we limit this to 100000). If previous LRU of 5000 items took around 5 MB, the dynamic is expected to take 100 MB at most (per worker).

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -28,9 +28,6 @@ local gsub              = string.gsub
 local find              = string.find
 local lower             = string.lower
 local fmt               = string.format
-local max               = math.max
-local min               = math.min
-local ceil              = math.ceil
 local ngx               = ngx
 local var               = ngx.var
 local log               = ngx.log
@@ -624,9 +621,15 @@ end
 
 
 do
+  local max  = math.max
+  local min  = math.min
+  local ceil = math.ceil
+
+  local DEFAULT_MATCH_LRUCACHE_SIZE = Router.DEFAULT_MATCH_LRUCACHE_SIZE
+
   local router
   local router_version
-  local router_cache_size = Router.DEFAULT_MATCH_LRUCACHE_SIZE
+  local router_cache_size = DEFAULT_MATCH_LRUCACHE_SIZE
   local router_cache = lrucache.new(router_cache_size)
   local router_cache_neg = lrucache.new(router_cache_size)
 
@@ -787,7 +790,7 @@ do
       end
     end
 
-    local n = Router.DEFAULT_MATCH_LRUCACHE_SIZE
+    local n = DEFAULT_MATCH_LRUCACHE_SIZE
     local cache_size = min(ceil(max(i / n, 1)) * n, n * 20)
 
     local reinitialize_cache = cache_size ~= router_cache_size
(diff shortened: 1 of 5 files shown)


kong/runloop/handler.lua
```diff
@@ -780,8 +785,16 @@ do
...
       end
+    end
+
+    local n = Router.DEFAULT_MATCH_LRUCACHE_SIZE
+    local cache_size = min(ceil(max(i / n, 1)) * n, n * 20)
 
-      counter = counter + 1
+    local reinitialize_cache = cache_size ~= router_cache_size
+    if reinitialize_cache then
+      router_cache:flush_all()
```
dndx, line 795: Shouldn't we always flush the cache even if the size did not change? Otherwise stale results may be used later.

bungle: @dndx yes I left the normal flushing code in its place, but added if statement there:
https://github.com/Kong/kong/pull/9309/files#diff-093344a0609553c69b475fb5cbd57a202796383eb6fca553d26ac6a1f40d4492R811-R813

bungle: Originally I wanted to flush them right after we had switched the router (but not before - the router rebuild yields, as you added them to reduce latency, which is great).

kong/runloop/handler.lua
```diff
@@ -780,8 +785,16 @@ do
           routes[i] = r
         end
       end
+    end
+
+    local n = Router.DEFAULT_MATCH_LRUCACHE_SIZE
```
chronolaw, line 790: Here we should use `router_cache_size` which is defined above.

kong/runloop/handler.lua
```diff
@@ -623,8 +626,9 @@ end
 do
   local router
   local router_version
-  local router_cache = lrucache.new(Router.MATCH_LRUCACHE_SIZE)
-  local router_cache_neg = lrucache.new(Router.MATCH_LRUCACHE_SIZE)
+  local router_cache_size = Router.DEFAULT_MATCH_LRUCACHE_SIZE
```
chronolaw, line 629: Should we add a line `local DEFAULT_MATCH_LRUCACHE_SIZE = Router.DEFAULT_MATCH_LRUCACHE_SIZE`

chronolaw (approved):

kong/runloop/handler.lua
```diff
@@ -28,6 +28,9 @@ local gsub              = string.gsub
 local find              = string.find
 local lower             = string.lower
 local fmt               = string.format
+local max               = math.max
```
chronolaw, line 31: Should we move `max/min/ceil` to the `do ... end` block?

kong/runloop/handler.lua
dndx: @chronolaw Could you make the change and I will merge it.

PR #9312, opened 2022-08-25
add scope validation for token request

liyangau:
add scope validation for token request


vm-001:
@liyangau Perhaps it doesn't need to validate the scope again against another global oauth2 plugin.

If a code was issued by a global(global_credentials=true) oauth2 plugin A, and then was changed to an access_token by another global(global_credentials=true) oauth2 plugin B. The scope has already been validated at the code issuing step by A plugin, so I don't think it needs to validate again against another global oauth2 plugin. 

A valid access_token can access multiple global oauth2 plugins, despite the access_token scope does not match the hitting plugin's scope. Therefore, I think another scope checking seems unnecessary. What do you think? :-)

PR #9315, opened 2022-08-25
fix(migration) up part of regex route

Suika-Kong:
Fix FT-3254

### DIFF
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -416,11 +416,6 @@ return {
       if err then
         return nil, err
       end
-
-      _, err = c_migrate_up_regex_path(coordinator)
-      if err then
-        return nil, err
-      end
     end,
 
     teardown = function(connector)
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -144,7 +144,7 @@ end
 
 
 local function c_migrate_regex_path(coordinator)
-  for rows, err in coordinator:iterate("SELECT id, paths FROM routes WHERE migrate_3 IS NOT true") do
+  for rows, err in coordinator:iterate("SELECT id, paths FROM routes WHERE is_new_regex_format IS NOT true") do
     if err then
       return nil, err
     end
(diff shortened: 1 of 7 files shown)


kong/db/schema/entities/routes.lua
```diff
@@ -87,6 +87,8 @@ else
       { methods        = typedefs.methods },
       { hosts          = typedefs.hosts },
       { paths          = typedefs.router_paths },
+      -- default to true so user added routes will be in new format
+      { migrate_3      = { type = "boolean", required = true, default = true }, },
```
mayocream, line 91: How about `is_new_regex_format`?

kong/db/migrations/core/016_280_to_300.lua
```diff
@@ -393,6 +401,8 @@ return {
 
       ALTER TABLE routes ADD expression text;
       ALTER TABLE routes ADD priority int;
+
+      ALTER TABLE routes ADD COLUMN migrate_3 boolean DEFAULT FALSE;
```
mayocream, line 405:
```suggestion
      ALTER TABLE routes ADD migrate_3 boolean DEFAULT FALSE;
```

kong/router/atc_compat.lua
```diff
@@ -14,6 +14,7 @@ local tb_new = require("table.new")
 local tb_clear = require("table.clear")
 local tb_nkeys = require("table.nkeys")
 local yield = require("kong.tools.utils").yield
+local migrate_route = require "kong.db.migrations.migrate_regex_280_300".migrate_route
```
chronolaw, line 17: Let's use style `require("xxx").xxx`

Suika-Kong, line 17:
```suggestion
local migrate_route = require("kong.db.migrations.migrate_regex_280_300").migrate_route
```

kong/db/migrations/migrate_regex_280_300.lua
```diff
@@ -83,4 +83,17 @@ local function migrate_regex(reg)
   return migrated, true
 end
 
-return migrate_regex
+local function migrate_route(route)
```
chronolaw, line 86: how about `migrate_paths`

Suika-Kong: It works on the whole route.

kong/db/schema/entities/routes.lua
Suika-Kong: done

dndx:
Should `is_new_regex_format` be dropped in the teardown phase? Will this field be shown to the user when they request Admin APIs?

bungle:
Things that worry me:
- on up we do both schema changes and data migrations, this has been especially flaky with cassandra clusters
- we add a new visible property to schema, and thus everywhere, including admin api and daos. Can this property be mistakenly set to wrong value

Suika-Kong:
> Should `is_new_regex_format` be dropped in the teardown phase? Will this field be shown to the user when they request Admin APIs?

That will fail its purpose to make it  safe to do this migration multiple times.

Suika-Kong:
Closing because we decide to do it another way.

PR #9327, opened 2022-08-26
perf(router) if header matching is not used, don't read headers

bungle:
### Summary

Optimize `traditional_compatible` and `expressions` router to not read headers in case there are no routes with header matching rules.

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -376,6 +376,17 @@ local function route_priority(r)
 end
 
 
+local function has_header_matching_field(fields)
+  for _, field in ipairs(fields) do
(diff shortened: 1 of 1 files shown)


chronolaw (approved):

kong/router/atc_compat.lua
```diff
@@ -502,7 +515,11 @@ local function new_from_previous(routes, is_traditional_compatible, old_router)
     yield(true)
   end
 
-  old_router.fields = inst:get_fields()
+  local fields = inst:get_fields()
+
+  old_router.fields = fields
+  old_routes.match_headers = has_header_matching_field(fields)
```
chronolaw, line 521: `old_routes` => `old_router`

PR #9334, opened 2022-08-29
fix(db) blue/green migration for route regex path

Suika-Kong:
The migration steps should happen in the up phase, not the teardown phase.

Notice this PR needs cooperation from 2.8.x.x.

fix FT-3293

### DIFF
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -143,7 +143,7 @@ local function c_copy_vaults_beta_to_sm_vaults(coordinator)
 end
 
 
-local function c_normalize_regex_path(coordinator)
+local function c_migrate_regex_path(coordinator)
   for rows, err in coordinator:iterate("SELECT id, paths FROM routes") do
     if err then
       return nil, err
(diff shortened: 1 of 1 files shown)


hanshuebner (approved):

jschmid1:
@Suika-Kong could you please adapt the commit message so that it is compliant to https://www.conventionalcommits.org/en/v1.0.0/

bungle:
> @Suika-Kong could you please adapt the commit message so that it is compliant to https://www.conventionalcommits.org/en/v1.0.0/

I think this is our format:
https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format

bungle:
If I remember correctly, the original idea of migrations was to do:
1. schema changes in up
2. data changes in finish

We have done data changes in `finish` before, and sometimes they work, and sometimes they cause huge issues. Especially if there are both schema changes and data changes. This is especially problematic in Cassandra clusters, if I remember correctly.

yzyyi:
Is this one still target 3.0? Trying to decide if needed cherrypick

mayocream (approved):

Suika-Kong:
> 

We decided to only support blue/green migration from 2.8.1.5 where we support both `~` and auto-detection(but users are not allowed to create `~` regex path).

PR #9335, opened 2022-08-29
fix(router) do not yield when incremental build

chronolaw:
### Summary

See FT-3291.

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -497,8 +497,6 @@ local function new_from_previous(routes, is_traditional_compatible, old_router)
     end
 
     new_updated_at = max(new_updated_at, route_updated_at)
-
-    yield(true)
   end
 
   -- remove routes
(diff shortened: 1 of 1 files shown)


bungle:
@chronolaw while this is a solution to fix the issue mentioned in FTI, I creates big concerns on incremental rebuild effects on latency with big data sets. We need to find a better solution. E.g. can the changes be prepared internally in router via some kind of `batch` functionality that then can be applied fast and atomically (non-yielding).

chronolaw:
> @chronolaw while this is a solution to fix the issue mentioned in FTI, it creates big concerns on incremental rebuild effects on latency with large number of routes. We need to find a better solution. E.g. can the changes be prepared internally in router via some kind of `batch` functionality that then can be applied fast and atomically (non-yielding) inside the router.

Agree.

dndx:
@bungle I agree we need some better way to do compare, but let's merge this PR first to stop the bleeding and let the testing continue. The current `master` is obviously broken because the yielding that broke atomicity.

dndx (approved):

dndx:
Actually, second thought, I think we can keep the yields, but use the "rebuild_router_timer" should fix the issue of concurrent rebuilds. I will make a change for it.

dndx:
We have a proper solution for this, and the issue does not affect CE. Closing this for now.

PR #9337, opened 2022-08-29
fix(core): ensure absolute path is used for event sock

flrgh:
### Summary

This fixes an issue with how Kong initializes resty.events. The code was
previously using `ngx.config.prefix()` to determine the listening socket
path to provide to the resty.events module. This causes breakage when
NGINX is started with a relative path prefix:

> failed to instantiate 'kong.worker_events' module: failed to disable listening: unix:my-prefix/worker_events.sock

Instead of using `ngx.config.prefix()`, Kong will now prefer
`kong.configuration.prefix` when available, as it is already normalized
to an absolute path. If `kong.configuration.prefix` is not defined, the
result of `ngx.config.prefix()` will be used after resolving it to an
absolute path.

Also contains an unrelated fix in `kong/cmd/start.lua`.

### Full changelog

* prefer `kong.configuration.prefix` to `ngx.config.prefix()`
* convert `ngx.config.prefix()` to an absolute path before using it
* fix pcall usage in `kong.cmd.start`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,8 +103,9 @@
 - Fix issue in router rebuilding where when paths field is invalid,
   the router's mutex is not released properly.
   [#9480](https://github.com/Kong/kong/pull/9480)
-- Fixed an issue where `kong docker-start` would fail if KONG_PREFIX was set to
-  a relative path. [#9337](https://github.com/Kong/kong/pull/9337)
+- Fixed an issue where `kong docker-start` would fail if `KONG_PREFIX` was set to
+  a relative path.
+  [#9337](https://github.com/Kong/kong/pull/9337)
 - Fixed an issue with error-handling and process cleanup in `kong start`.
   [#9337](https://github.com/Kong/kong/pull/9337)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,10 @@
 - Fix issue in router rebuilding where when paths field is invalid,
   the router's mutex is not released properly.
   [#9480](https://github.com/Kong/kong/pull/9480)
+- Fixed an issue where `kong docker-start` would fail if KONG_PREFIX was set to
+  a relative path. [#9337](https://github.com/Kong/kong/pull/9337)
+- Fixed an issue with error-handling and process cleanup in `kong start`.
+  [#9337](https://github.com/Kong/kong/pull/9337)
 
 #### CLI
 
--- a/kong/global.lua
+++ b/kong/global.lua
@@ -186,16 +186,22 @@ function _GLOBAL.init_worker_events()
     worker_events = require "resty.worker.events"
 
   else
-    local sock_name = "worker_events.sock"
-    if ngx.config.subsystem == "stream" then
-      sock_name = "stream_" .. sock_name
-    end
+    -- `kong.configuration.prefix` is already normalized to an absolute path,
+    -- but `ngx.config.prefix()` is not
+    local prefix = configuration
+                   and configuration.prefix
+                   or require("pl.path").abspath(ngx.config.prefix())
+
+    local sock = ngx.config.subsystem == "stream"
+                 and "stream_worker_events.sock"
+                 or "worker_events.sock"
+
+    local listening = "unix:" .. prefix .. "/" .. sock
 
     opts = {
       unique_timeout = 5,     -- life time of unique event data in lrucache
       broker_id = 0,          -- broker server runs in nginx worker #0
-      listening = "unix:" ..  -- unix socket for broker listening
-                  ngx.config.prefix() .. sock_name,
+      listening = listening,  -- unix socket for broker listening
       max_queue_len = 1024 * 50,  -- max queue len for events buffering
     }
 
--- a/spec/02-integration/02-cmd/02-start_stop_spec.lua
+++ b/spec/02-integration/02-cmd/02-start_stop_spec.lua
@@ -817,7 +817,7 @@ describe("kong start/stop #" .. strategy, function()
       end)
 
       assert(helpers.kong_exec(string.format("prepare -p %q", prefix), {
-        db = strategy,
+        database = strategy,
         proxy_listen = "127.0.0.1:8000",
         stream_listen = "127.0.0.1:9000",
         admin_listen  = "127.0.0.1:8001",
--- a/spec/02-integration/02-cmd/02-start_stop_spec.lua
+++ b/spec/02-integration/02-cmd/02-start_stop_spec.lua
@@ -798,5 +798,65 @@ describe("kong start/stop #" .. strategy, function()
     end)
   end)
 
+  describe("docker-start", function()
+    -- tests here are meant to emulate the behavior of `kong docker-start`, which
+    -- is a fake CLI command found in our default docker entrypoint:
+    --
+    -- https://github.com/Kong/docker-kong/blob/d588854aaeeab7ac39a0e801e9e6a1ded2f65963/docker-entrypoint.sh
+    --
+    -- this is the only(?) context where nginx is typically invoked directly
+    -- instead of first going through `kong.cmd.start`, so it has some subtle
+    -- differences
+
+    it("works with resty.events when KONG_PREFIX is a relative path", function()
+      local prefix = "relpath"
+
+      finally(function()
+        helpers.kill_all(prefix)
+        pcall(helpers.dir.rmtree, prefix)
+      end)
+
+      assert(helpers.kong_exec(string.format("prepare -p %q", prefix), {
+        db = strategy,
+        proxy_listen = "127.0.0.1:8000",
+        stream_listen = "127.0.0.1:9000",
+        admin_listen  = "127.0.0.1:8001",
+      }))
+
+      local nginx, err = require("kong.cmd.utils.nginx_signals").find_nginx_bin()
+      assert.is_string(nginx, err)
+
+      local started
+      started, err = helpers.execute(string.format("%s -p %q -c nginx.conf",
+                                    nginx, prefix))
+
+      assert.truthy(started, "starting Kong failed: " .. tostring(err))
+
+      -- wait until everything is running
+      helpers.wait_until(function()
+        local client = helpers.admin_client(5000, 8001)
+        local res, rerr = client:send({
+          method = "GET",
+          path = "/",
+        })
+
+        if res then res:read_body() end
+        client:close()
+
+        assert.is_table(res, rerr)
+
+        return res.status == 200
+      end)
+
+      assert.truthy(helpers.path.exists(prefix .. "/worker_events.sock"))
+      assert.truthy(helpers.path.exists(prefix .. "/stream_worker_events.sock"))
+
+      assert.logfile(prefix .. "/logs/error.log").has.no.line("[error]", true)
+      assert.logfile(prefix .. "/logs/error.log").has.no.line("[alert]", true)
+      assert.logfile(prefix .. "/logs/error.log").has.no.line("[crit]", true)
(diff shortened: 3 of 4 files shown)


kong/global.lua
```diff
@@ -186,16 +186,22 @@ function _GLOBAL.init_worker_events()
...
 
   else
-    local sock_name = "worker_events.sock"
-    if ngx.config.subsystem == "stream" then
-      sock_name = "stream_" .. sock_name
-    end
+    -- `kong.configuration.prefix` is already normalized to an absoluate path,
+    -- but `ngx.config.prefix()` is not
+    local prefix = configuration
+               and configuration.prefix
```
chronolaw, line 192: sytle issue: we should align to `configuration `

kong/global.lua
```diff
@@ -186,16 +186,22 @@ function _GLOBAL.init_worker_events()
...
+                or require("pl.path").abspath(ngx.config.prefix())
+
+    local sock = ngx.config.subsystem == "stream"
+             and "stream_worker_events.sock"
```
chronolaw, line 196: alignment issue.

kong/global.lua
```diff
@@ -186,16 +186,22 @@ function _GLOBAL.init_worker_events()
...
+              or "worker_events.sock"
+
+    local listen = "unix:" .. prefix .. "/" .. sock
```
chronolaw, line 199: `listen` => `listening` maybe a better choice.

Tieske:
@flrgh There is a related issue; Pongo is having permission issues with this unix socket. The problem is that in Pongo, the `prefix` is located on a mount (where the plugin code is located). Creating the unix socket on that mount causes permission issues.

We had the same issue with Vagrant a long time ago (because test nginx uses unix sockets as well), it was fixed like this: https://github.com/Kong/kong-vagrant/blob/master/provision.sh#L331-L332

So besides making it an absolute path, it should also become configurable by an env var.

chronolaw:
I have added the env `KONG_RESTY_EVENTS_SOCK_PATH` and did a simple test, it works well, but I don't know how to write a unit test.

Also, we should consider PR #9254, which need some change to fit this PR.

kong/global.lua
```diff
@@ -186,16 +186,21 @@ function _GLOBAL.init_worker_events()
...
+    -- `kong.configuration.prefix` is already normalized to an absoluate path,
+    -- but `ngx.config.prefix()` is not
+    local prefix = configuration and
+                   configuration.resty_events_sock_path or configuration.prefix or
+                   require("pl.path").abspath(ngx.config.prefix())
+
+    local sock = ngx.config.subsystem == "stream" and
+                 "stream_worker_events.sock" or "worker_events.sock"
+
+    local listening = "unix:" .. prefix .. "/" .. sock
```
Tieske, line 198: can't we construct this path in the config loader module? and store it in `resty_events_sock_path`. From there on, everywhere else we need it we can just grab it from the config, without have to do the construction again.

Would also mean it is reported on the management api, as the full filename, without a user having to apply logic where it was created.

kong/templates/nginx_kong.lua
```diff
@@ -453,7 +453,11 @@ server {
 > if not legacy_worker_events then
 server {
     server_name kong_worker_events;
+> if resty_events_sock_path and #resty_events_sock_path > 0 then
+    listen unix:${{RESTY_EVENTS_SOCK_PATH}}/worker_events.sock;
+> else
     listen unix:${{PREFIX}}/worker_events.sock;
+> end
```
Tieske, line 460: if we construct the value in the config loader, we can just set it here, no need for the if/then logic in the templates (this one as well as the stream one)

kong/global.lua
chronolaw: sounds good.

kong/templates/nginx_kong.lua
chronolaw: good idea, but I am not familiar with conf loader's code.

chronolaw: I add some code in conf loader, now we use a simple `${{RESTY_EVENTS_SOCK_PATH}}`.

kong/global.lua
chronolaw: resty-events library needs two socket file, one in http subsystem, another in stream subsystem, so we can only set a path, not a file name.

kikito:
Once this is approved, send cherry-pick against `release/3.0.1`

CHANGELOG.md
```diff
@@ -71,13 +71,24 @@
...
 - Fix issue where external plugins crashing with unhandled exceptions
   would cause high CPU utilization after the automatic restart.
   [#9384](https://github.com/Kong/kong/pull/9384)
 - Fix issue in `header_filter` instrumentation where the span was not
   correctly created.
   [#9434](https://github.com/Kong/kong/pull/9434)
 
+#### Configuration
+
+- A new configuration item (`resty_events_sock_path`) has been added to allow
```
dndx, line 88: This makes little sense to me, why is this config option something the user should know in the first place?

dndx (changes requested):

CHANGELOG.md
chronolaw: OK, I will remove it in changelog.

dndx (changes requested):

kong/conf_loader/init.lua
```diff
@@ -530,6 +530,7 @@ local CONF_INFERENCES = {
   untrusted_lua_sandbox_requires = { typ = "array" },
   untrusted_lua_sandbox_environment = { typ = "array" },
 
+  resty_events_sock_path = { typ = "string" },
```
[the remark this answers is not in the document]
chronolaw: done.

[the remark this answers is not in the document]
Tieske: This is not ok imho. There's a system to the configuration, defaults -> config-file -> env vars. Everything works in the same way and is predictable for the user. 

I get that the setting is rather technical, and only used in specific cases, but this breaks the mentioned logic and will in the long run create more confusion for the user. Imho we should stick to the standard mechanism and remain predictable. Don't create special cases.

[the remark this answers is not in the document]
dndx: > There's a system to the configuration, defaults -> config-file -> env vars

This is not true, a counter example: https://github.com/Kong/kong/blob/master/kong/conf_loader/init.lua#L1745-L1746

There is no issue I can see regarding utilizing `conf` to carry custom data, if it makes no sense for user to change then it should not be exposed as the config interface.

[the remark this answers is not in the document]
Tieske: apparently we dropped the consistency before then. 

So here's the underlying issue: https://github.com/Kong/kong-pongo/issues/333

```
Error: nginx configuration is invalid (exit code 1):
nginx: the configuration file /kong-plugin/servroot/nginx.conf syntax is ok
nginx: [emerg] bind() to unix:/kong-plugin/servroot/worker_events.sock failed (22: Invalid argument)
nginx: configuration file /kong-plugin/servroot/nginx.conf test failed
```

This is from using Pongo, which has the `prefix` located on a mount. Hence the location must be configurable, to get it outside of the mount. I'm not sure whether that is what you'd call "user configurable". Please advise @dndx

flrgh:
:wave: hi folks, I want to back things up a bit here, because this PR has seen some scope creep while I was out on vacation.

The original bugfix commit (4158cee1a3cd5b7f43938676807fc284833a01f2) that prompted this PR was intended to address a very simple bug--we were using a relative path in a place where we ought to be using an absolute path, causing a fatal error.

It has since been pushed beyond the `fix` scope and is now in `feat` territory.

The bugfix is not mutually-exclusive with the notion of a new config field (even if the latter eventually supplants the work of the former). Additionally, I see no reason that these two things (the fix and the new config field) need to happen in the same PR. Therefore, unless anyone has a compelling case to make otherwise, I'd like to return this branch/PR to it's prior state+scope, because I don't want the bugfix to be held back while the merits and implementation details of the feature are still being sorted out. The socket path change for Pongo can then be carried out in another PR.

flrgh:
My thoughts re: Pongo and unix sockets

The worker events socket is not the only unix socket that Kong places in the `KONG_PREFIX` directory (i.e. `stream_rpc.sock`, `stream_tls_passthrough.sock`, etc). It just so happens to be the only unix socket that is always on/enabled by default. If we need to make socket paths configurable, we should do so in a way that works for _all_ unix sockets that Kong manages--not just worker events.

chronolaw:
Hi flrgh, I am glad you are back. I agree that this PR has grown a bit big, splitting the PR may let it easier to understand.

fffonion:
I was about to type the same and say @flrgh 's latest comment, we have;
```
kong/templates/nginx_kong_stream.lua:    listen unix:${{PREFIX}}/stream_tls_terminate.sock ssl proxy_protocol;
kong/templates/nginx_kong_stream.lua:    listen unix:${{PREFIX}}/stream_tls_passthrough.sock proxy_protocol;
kong/templates/nginx_kong_stream.lua:    listen unix:${{PREFIX}}/stream_config.sock;
kong/templates/nginx_kong_stream.lua:    listen unix:${{PREFIX}}/stream_rpc.sock;
kong/templates/nginx_kong_stream.lua:    listen unix:${{PREFIX}}/stream_worker_events.sock;
kong/templates/nginx_kong.lua:    listen unix:${{PREFIX}}/worker_events.sock;
```
 
so we would need to give the config a generic name, they can share the same path, we don't need seperate config/logic for each socket.

chronolaw:
I think we need someone to push this PR forward. The first thing may be to split this PR and then pick some changes into the new PRs.

flrgh:
I restored the branch to the original 2 bugfix items (with a couple style fixes) and rebased onto master. This is ready for another round of review.

spec/02-integration/02-cmd/02-start_stop_spec.lua
```diff
@@ -798,5 +798,65 @@ describe("kong start/stop #" .. strategy, function()
...
+    it("works with resty.events when KONG_PREFIX is a relative path", function()
+      local prefix = "relpath"
+
+      finally(function()
+        helpers.kill_all(prefix)
+        pcall(helpers.dir.rmtree, prefix)
+      end)
+
+      assert(helpers.kong_exec(string.format("prepare -p %q", prefix), {
+        db = strategy,
```
chronolaw, line 820: `db` should be `database`

flrgh: oooh good catch :+1: fixed

chronolaw (approved):

CHANGELOG.md
```diff
@@ -93,6 +93,11 @@
 - Fix issue in router rebuilding where when paths field is invalid,
   the router's mutex is not released properly.
   [#9480](https://github.com/Kong/kong/pull/9480)
+- Fixed an issue where `kong docker-start` would fail if `KONG_PREFIX` was set to
+  a relative path.
+  [#9337](https://github.com/Kong/kong/pull/9337)
+- Fixed an issue with error-handling and process cleanup in `kong start`.
```
chronolaw, line 99: I think this change log entry has not existed in this PR.

dndx (changes requested):
I am still not convinced why the socket path needs to be configurable in the first place. If we need to fix the relative path issue, we should do just that and nothing more. Please share your reasoning in the ticket before merging anything.

chronolaw:
> I am still not convinced why the socket path needs to be configurable in the first place. If we need to fix the relative path issue, we should do just that and nothing more. Please share your reasoning in the ticket before merging anything.

You are right. 
Now, this PR is rebased on the earlier state, so it only contains the solution of the relative path.

fffonion (approved):

bungle (approved):

PR #9339, opened 2022-08-30
Detect and return error for concurrent rebuilds during incremental updates and add incremental rebuild tests

dndx:


bungle:
I am still wondering, wether to make the build router to return a new router is better, and perhaps simpler.

dndx:
@bungle I think this incremental thing needs a lot of improvements later because our entire flow right now is still very full build oriented. Right now I am just addressing the immediate concern for release with minimum changes that are not gonna break anything else.

chronolaw (approved):

kong/runloop/handler.lua
```diff
@@ -795,7 +795,6 @@ do
 
     local reinitialize_cache = cache_size ~= router_cache_size
     if reinitialize_cache then
-      router_cache:flush_all()
```
bungle, line 798: @dndx do you think just setting cache to new value and let the GC do the job is good enough? Or am I missing how this is related?

bungle:
Just a note here, this reminds me a bit of the WIP:
https://github.com/Kong/kong/pull/8953

While we improved our max latencies, what we did was to add a lot of `ngx.sleep(0)` here and there. A lot in Kong is kinda global state, and changing that global state with a lot of `yields` in concurrent light threads is a bit problematic. This means that there will be a lot more states in our system than before. We need to move to more to the form where the internals can be build on a side (incrementally) without affecting the proxy at the same time, and then we can switch the proxy side assets atomically. But yes, it is not going to be quick to fix this.

kong/runloop/handler.lua
dndx: Yes, even if we `flush` the cache, the actual value held by the table will still need to be GC'ed later, so I doubt this will free up any more memory for us immediately.

bungle: Ok, If that doesn't give any advantages (e.g. freeing memory earlier), then it is fine.

bungle: @dndx should we still just `flush` the `router_cache` after router has been build. My main worry is that some routes get cached, that incremental is in meanwhile removing.

spec/01-unit/08-router_spec.lua
```diff
@@ -1669,6 +1669,198 @@ for _, flavor in ipairs({ "traditional", "traditional_compatible", "expressions"
...
+
+            local match_t = nrouter:select("GET", "/foo1")
+            assert.truthy(match_t)
+            assert.same(use_case[1].route, match_t.route)
+
+            match_t = nrouter:select("GET", "/bar")
+            assert.falsy(match_t)
+          end)
+
+          it("update skips routes if updated_at is unchanged", function()
```
chronolaw, line 1732: This case may not prove the 'skip unchanged' feature.

Now the check condition is `route_updated_at >= updated_at`, so the route will still be updated.

Perhaps in the future we can implement this.

bungle (approved):

PR #9340, opened 2022-08-30
tests(plugins/statsd/log) fix some flaky tests about `shdict_usage` metric

ADD-SP:
The `shdict_usage` metric will be returned when Kong has just started, and also every minute, so we should test it when Kong has just started.

_FT-3301_


aboudreault (approved):

vm-001:
Nice. This should be backported to CE too.

vm-001 (approved):
Nice!

PR #9344, opened 2022-08-30
RateLimiting/ResponseRateLimiting: change the default policy to local (FT-3295)

aboudreault:
This will only affect new plugin configuration, so no cp/dp compat fix is required. docs PR to be done when merged.


tyler-ball (approved):
Code and CHANGELOG looks good. TODO: cherry-pick to master and update docs.konghq.com

PR #9346, opened 2022-08-31
fix(router) old_route has wrong assignment order

chronolaw:
### Summary

The assignment of `old_route` is wrong, it should happen before `old_routes[route_id] = route`.

And in the previous code, the rebuild is actually a complete build, not an incremental build.

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -26,6 +26,7 @@ local setmetatable = setmetatable
 local pairs = pairs
 local ipairs = ipairs
 local type = type
+local assert = assert
 local tonumber = tonumber
 local get_schema = atc.get_schema
 local max = math.max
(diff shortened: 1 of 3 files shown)


kong/router/atc_compat.lua
```diff
@@ -517,6 +520,8 @@ local function new_from_previous(routes, is_traditional_compatible, old_router)
       inst:remove_matcher(id)
       old_routes[id] = nil
       old_services[id] = nil
+
+      ngx_log(ngx_DEBUG, "route(", route_id, ") is removed")
```
chronolaw, line 524: `route_id` should be `id`

spec/01-unit/08-router_spec.lua
```diff
@@ -4251,3 +4251,96 @@ describe("[both regex and prefix with regex_priority]", function()
 
 end)
 
+
+describe("incremental rebuild", function()
```
chronolaw, line 4255: Could we move these code to line 1673, which tests `flavor ~= "traditional"`.

chronolaw: @dndx , Could you take a look at these cases, Do you have better idea?

dndx: Moved.

dndx (approved):

PR #9348, opened 2022-08-31
fix(request-transformer) properly catch errors of cjson.encode as we changed to cjson.safe now

catbro666:
fix issue introduced by [#9186](https://github.com/Kong/kong/pull/9186)

FTI-4205


flrgh:
Code change LGTM, but I think we should add some logging around this so that we don't fail silently (since the old behavior would have been to throw an `error()` on an encode failure).

fffonion (approved):

PR #9350, opened 2022-08-31
Cherry pick 3.0.x changes

hanshuebner:


bungle (approved):

.requirements
```diff
@@ -10,5 +10,5 @@ RESTY_LMDB_VERSION=1.0.0
 RESTY_EVENTS_VERSION=0.1.3
 ATC_ROUTER_VERSION=1.0.0
 LIBYAML_VERSION=0.2.5
-KONG_BUILD_TOOLS_VERSION=4.33.11
+KONG_BUILD_TOOLS_VERSION=chore/test_3_0_release
```
bungle, line 13: This needs to be changed before merge.

kong/meta.lua
```diff
@@ -2,7 +2,7 @@ local version = setmetatable({
   major = 3,
   minor = 0,
   patch = 0,
-  suffix = "-alpha.1"
+  suffix = "-alpha.13"
```
bungle, line 5: This might also be something we don't want in master?

.requirements
hanshuebner: .. or so I thought :) - I'll downgrade to 4.33.16 of kbt so that we can merge this.

PR #9358, opened 2022-09-01
chore(clustering) old cp switch

Suika-Kong:
Turn off `legacy_hybrid_compatibility` to avoid serializing twice

Fix FT-3308

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -108,7 +108,9 @@ function _M:init_cp_worker(plugins_list)
   -- their data planes
   kong.worker_events.register(handle_dao_crud_event, "dao:crud")
 
-  self.json_handler:init_worker(plugins_list)
+  if kong.configuration.legacy_hybrid_compatibility then
+    self.json_handler:init_worker(plugins_list)
+  end
   if not kong.configuration.legacy_hybrid_protocol then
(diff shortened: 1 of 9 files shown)


kong/clustering/init.lua
```diff
@@ -108,7 +108,9 @@ function _M:init_cp_worker(plugins_list)
   -- their data planes
   kong.worker_events.register(handle_dao_crud_event, "dao:crud")
 
-  self.json_handler:init_worker(plugins_list)
+  if kong.configuration.legacy_hybrid_compatibility then
```
flrgh, line 111: Maybe we could squash this and `legacy_hybrid_protocol` into one setting (`hybrid_protocol = wrpc,legacy|wrpc|legacy`). WDYT? I know we're just trying to get an issue resolved for 3.0 as simply as possible so feel free to disregard.

dndx:
I don't think we still need this after the smart detection of config export based on clients connected.

kong/clustering/init.lua
Suika-Kong: `legacy_hybrid_protocol` is a hidden config, while `legacy_hybrid_compatibility` is exposed to users. However, it doesn't matter now as we are automatically disabling old protocol when no old DP connects.

PR #9359, opened 2022-09-01
fix test failure

liverpool8056:
Fix test case "returns HEALTHCHECK_OFF for target with weight 0" in spec/02-integration/04-admin_api/08-targets_routes_spec.lua

Add a short time waiting for the config to be taken effect.

### DIFF
--- a/spec/02-integration/04-admin_api/08-targets_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/08-targets_routes_spec.lua
@@ -501,7 +501,7 @@ describe("Admin API #" .. strategy, function()
           })
           assert.same(201, status)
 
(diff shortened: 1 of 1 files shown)


spec/02-integration/04-admin_api/08-targets_routes_spec.lua
```diff
@@ -501,6 +501,8 @@ describe("Admin API #" .. strategy, function()
           })
           assert.same(201, status)
 
+          ngx.sleep(0.5)
```
ADD-SP, line 504:
```suggestion
          helpers.wait_for_all_config_update()
```

ADD-SP (dismissed):

bungle (approved):

PR #8840, opened 2022-05-23
perf(clustering) skip push_config if no clients connected

fffonion:
### Summary
json protocol and wRPC protocol starts own timer to push config.
This PR reduces unnecessary overhead when no clients connected to either
of the protocol.

### Full changelog

* perf(clustering) skil push_config if no clients connected

### Issue reference

Fix #_[issue number]_

PR #9362, opened 2022-09-01
feat(clustering) config export on demand

Suika-Kong:
fix FT-3308

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -180,7 +180,7 @@ end
 _M._update_compatible_payload = update_compatible_payload
 
 function _M:export_deflated_reconfigure_payload()
-  ngx_log(ngx_DEBUG, "exporting config for old protocol")
+  ngx_log(ngx_DEBUG, _log_prefix, "exporting config for legacy protocol")
   local config_table, err = declarative.export_config()
   if not config_table then
     return nil, err
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -180,6 +180,7 @@ end
 _M._update_compatible_payload = update_compatible_payload
 
 function _M:export_deflated_reconfigure_payload()
+  ngx_log(ngx_DEBUG, "exporting config for old protocol")
   local config_table, err = declarative.export_config()
   if not config_table then
     return nil, err
@@ -352,6 +353,12 @@ function _M:handle_cp_websocket()
         return queue_semaphore:post(...)
       end
     }
+  end
+
+  -- first client, or a new client after clients emptied,
+  -- in this case the cache may be stale
+  if not next(self.clients) then
+    self.deflated_reconfigure_payload = nil
   end
 
   self.clients[wb] = queue
@@ -540,7 +547,9 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if exiting() then
       return
     end
-    if ok then
+    -- we still need to wait for config for if no clients another run, just in case new client is in
+    if ok and next(self.clients) then
+
       ok, err = pcall(self.push_config, self)
       if ok then
         local sleep_left = delay
@@ -562,8 +571,9 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
       else
         ngx_log(ngx_ERR, _log_prefix, "export and pushing config failed: ", err)
       end
+    end
 
-    elseif err ~= "timeout" then
+    if err ~= "timeout" then
       ngx_log(ngx_ERR, _log_prefix, "semaphore wait error: ", err)
     end
   end
--- a/kong/clustering/wrpc_control_plane.lua
+++ b/kong/clustering/wrpc_control_plane.lua
@@ -115,6 +115,7 @@ end
 local config_version = 0
 
 function _M:export_deflated_reconfigure_payload()
+  ngx_log(ngx_DEBUG, "exporting config for wRPC protocol")
   local config_table, err = declarative.export_config()
   if not config_table then
     return nil, err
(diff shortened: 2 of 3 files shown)


bungle:
This reminds me of this:
https://github.com/Kong/kong/pull/9082

Suika-Kong:
> This reminds me of this: #9082

yeah, we reduced 4 times of the process to 2, and now it should be 1

bungle:
This also reminds me of this, @Suika-Kong please read carefully all the comments there:
https://github.com/Kong/kong/pull/8840

bungle:
While I think `on demand` is good, it does not solve the `double` problem. A lot of code will still be double.

Suika-Kong:
> While I think `on demand` is good, it does not solve the `double` problem. A lot of code will still be double and executed double when you have both types of clients connected. I think common code should be common and executed only once.

That would be hard. We've discussed that earlier: the format of the 2 protocols are very different, and the most we could do here is to share the config table exported. I think that was the reason why we did not do it that time.

bungle:
> That would be hard. We've discussed that earlier: the format of the 2 protocols are very different, and the most we could do here is to share the config table exported. I think that was the reason why we did not do it that time.

Last time I checked it was not that hard. The wRPC is no different, it just wraps the above in its own protocol. What I said above is common, AFAIK.

bungle:
> While I think `on demand` is good, it does not solve the `double` problem. A lot of code will still be double and executed double when you have both types of clients connected. I think common code should be common and executed only once.
> 
> That is:
> 
> 1. there is no reason both wrpc and old do export
> 2. there is no reason both calculate hashes
> 3. there is no reason both json encode config
> 4. there is no reason both compress the config

I stand corrected, only the 1. and 2. are easily shareable. The 3. and 4. are not.

bungle (changes requested):

kong/clustering/control_plane.lua
```diff
@@ -180,6 +180,7 @@ end
 _M._update_compatible_payload = update_compatible_payload
 
 function _M:export_deflated_reconfigure_payload()
+  ngx_log(ngx_DEBUG, "exporting config for old protocol")
```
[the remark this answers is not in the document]
flrgh: Might also consider `legacy` instead of `old`.

fffonion:
@Suika-Kong if remove the duplicate code is too much work to do, let's not it do it in this PR (but sure to have a ticket to record it). As CP mostly takes to DP in either of the protocol, making it "on demand" should solve the CP 2x latency issue in most cases, let's focus on ship that and also solve the caching issue Aapo mentioned in 8840.

kong/clustering/control_plane.lua
```diff
@@ -562,8 +571,9 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
       else
         ngx_log(ngx_ERR, _log_prefix, "export and pushing config failed: ", err)
       end
+    end
 
-    elseif err ~= "timeout" then
+    if err ~= "timeout" then
```
flrgh, line 576: if `ok` is truth-y but `self.clients` is empty, this branch will be followed because `err` will be `nil` and generate a confusing error log entry (I think).

Maybe it would be more straightforward to put the `self.clients` check into `push_config`?

```lua
function _M:push_config()
  if not next(self.clients) then
    return true
  end

  local start = ngx_now()

  local payload, err = self:export_deflated_reconfigure_payload()
  if not payload then
  -- ...
end
```

kong/clustering/wrpc_control_plane.lua
```diff
@@ -305,7 +308,10 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if exiting() then
       return
     end
-    if ok then
+  
+    -- we still need to wait for config for if no clients another run, just in case new client is in
+    if ok and next(self.clients)then
```
flrgh, line 313:
```suggestion
    if ok and next(self.clients) then
```

kong/clustering/wrpc_control_plane.lua
```diff
@@ -305,7 +308,10 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if exiting() then
       return
     end
-    if ok then
+
```
flrgh, line 311:
```suggestion

```

kong/clustering/control_plane.lua
flrgh: Alternatively, maybe the `self.clients` check could be pushed to the event handler? This would be simple for both `control_plane` and `wrpc_control_plane`:

```lua
  kong.worker_events.register(function(_)
    if next(self.clients) and push_config_semaphore:count() <= 0 then
      -- the following line always executes immediately after the `if` check
      -- because `:count` will never yield, end result is that the semaphore
      push_config_semaphore:post()
    end
  end, "clustering", "push_config")
```

kong/clustering/control_plane.lua
```diff
@@ -540,7 +547,9 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if exiting() then
       return
     end
-    if ok then
+    -- we still need to wait for config for if no clients another run, just in case new client is in
+    if ok and next(self.clients) then
+
```
dndx, line 552:
```suggestion
```

kong/clustering/control_plane.lua
```diff
@@ -540,7 +547,9 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if exiting() then
       return
     end
```
dndx, line 549:
```suggestion
    end

```

kong/clustering/control_plane.lua
dndx: @flrgh The first approach has some issues, if we `return true` inside `push_config`, then the timer of `push_config_loop` is still reset which means that loop will wait for `db_update_frequency` before even pushing out another updated config.

The second issue might work (I did not spot immediate issue with it), but it seems to be a bit more dangerous as there are a lot of asynchronous going on between worker events and the semaphore. It's a bit harder to argue the correctness.

My thought:

If we change the `push_config_loop` function to do the following:

```
while not exiting() do
    local ok, err = push_config_semaphore:wait(1)
    if exiting() then
      return
    end

    if ok then
      if next(self.clients) do
        ok, err = pcall(self.push_config, self)
        ...
```

Does this also work? It seems to be simpler than the current approach.

Suika-Kong:
Closing for @dndx 's PR. #9368

PR #9366, opened 2022-09-01
docs: embrace conventional-commits

hanshuebner:
### Summary

Kong's commit message format was similar to [conventional-commits](https://github.com/Kong/kong/pull/new/conventional-commits), but deviated in not requiring a colon after the type and optional scope.  As other repositories in the Kong organization have adopted conventional-commits and external tools begin to rely on the format, we're switching `Kong/kong` to this format as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -241,10 +241,13 @@ messages as follows:
 - A blank line should be included between the header and the body
 - The body of your message should not contain lines longer than 72 characters
 
+We strive to adapt the [conventional-commits](https://www.conventionalcommits.org/en/v1.0.0/)
+format.
+
 Here is a template of what your commit message should look like:
 
 ```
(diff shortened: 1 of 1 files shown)


hutchic (approved):

bungle (approved):

aboudreault:
just to mention, the China team can still participate to the poll tomorrow and win with the opposite option :joy: . I doubt we will need to revert this PR but to consider next time. :)

dndx:
Can we better communicate this before making the switch? I am not opposed to the idea, but it seems to be coming out of nowhere suddenly.

hanshuebner:
Sorry, this sparked a little too much enthusiasm in Slack and then went in too quickly.  It is just a documentation change, though, and we can revert it easily if we find a good reason to do so.  Apologies.

PR #9367, opened 2022-09-01
feat(conf): support base64 encoded *_cert and *_cert_key

samugi:
### Summary

Add support for base64(and base64url)-encoded *_cert and *_cert_key properties.


kong/conf_loader/init.lua
```diff
@@ -622,6 +623,22 @@ local function infer_value(value, typ, opts)
 end
 
 
+local function try_base64_decode(vals)
```
bungle, line 626: Perhaps try with `url` variant too, so we support them both.

samugi: @bungle done in the last commit

bungle (approved):

PR #9368, opened 2022-09-01
perf(clustering) on-demand export of declarative config for Hybrid mode protocols

dndx:
Based on https://github.com/Kong/kong/pull/9362, and inspirations from https://github.com/Kong/kong/pull/9368 with wRPC related bugs fixed.

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -180,7 +180,7 @@ end
 _M._update_compatible_payload = update_compatible_payload
 
 function _M:export_deflated_reconfigure_payload()
-  ngx_log(ngx_DEBUG, "exporting config for old protocol")
+  ngx_log(ngx_DEBUG, _log_prefix, "exporting config for legacy protocol")
   local config_table, err = declarative.export_config()
   if not config_table then
     return nil, err
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -180,6 +180,7 @@ end
 _M._update_compatible_payload = update_compatible_payload
 
 function _M:export_deflated_reconfigure_payload()
+  ngx_log(ngx_DEBUG, "exporting config for old protocol")
   local config_table, err = declarative.export_config()
   if not config_table then
     return nil, err
@@ -352,6 +353,12 @@ function _M:handle_cp_websocket()
         return queue_semaphore:post(...)
       end
     }
+  end
+
+  -- first client, or a new client after clients emptied,
+  -- in this case the cache may be stale
+  if not next(self.clients) then
+    self.deflated_reconfigure_payload = nil
   end
 
   self.clients[wb] = queue
@@ -540,7 +547,9 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if exiting() then
       return
     end
-    if ok then
+    -- we still need to wait for config for if no clients another run, just in case new client is in
+    if ok and next(self.clients) then
+
       ok, err = pcall(self.push_config, self)
       if ok then
         local sleep_left = delay
@@ -562,8 +571,9 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
       else
         ngx_log(ngx_ERR, _log_prefix, "export and pushing config failed: ", err)
       end
+    end
 
-    elseif err ~= "timeout" then
+    if err ~= "timeout" then
       ngx_log(ngx_ERR, _log_prefix, "semaphore wait error: ", err)
     end
   end
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -357,6 +357,9 @@ function _M:handle_cp_websocket()
     }
   end
 
+  -- if clients table is empty, we might have skipped some config
+  -- push event in `push_config_loop`, which means the cached config
+  -- might be stale, so we always export the latest config again in this case
   if isempty(self.clients) or not self.deflated_reconfigure_payload then
     _, err = handle_export_deflated_reconfigure_payload(self)
   end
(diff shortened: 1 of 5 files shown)


kong/clustering/control_plane.lua
```diff
@@ -540,27 +542,34 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
...
 
-          sleep_left = sleep_left - 1
+        else
+          ngx_log(ngx_ERR, _log_prefix, "export and pushing config failed: ", err)
         end
 
       else
-        ngx_log(ngx_ERR, _log_prefix, "export and pushing config failed: ", err)
+        -- no clients connected, clear the stale config cache
+        self.deflated_reconfigure_payload = nil
```
flrgh, line 572: We might not want to clear the cached payload in this case. According to 310722659eead1a6aa8a92a6cf025b5708bb0698, one of the reasons we keep this cache is so that the CP can serve a stale config payload in the event of a database outage.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -305,27 +307,35 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
...
-          sleep_left = sleep_left - 1
+        else
+          ngx_log(ngx_ERR, _log_prefix, "export and pushing config failed: ", err)
         end
 
       else
-        ngx_log(ngx_ERR, _log_prefix, "export and pushing config failed: ", err)
+        -- no clients connected, clear the stale config cache
+        self.config_call_rpc = nil
+        self.self.config_call_args = nil
```
flrgh, line 338: There's an extra `self` in here
```suggestion
        self.config_call_args = nil
```

kong/clustering/control_plane.lua
flrgh: Here's how I wound up dealing with this in my branch:
https://github.com/Kong/kong/blob/c08ee7f03ee15a01a06a85b8df07b60fb5508334/kong/clustering/control_plane.lua#L359-L365

dndx: This has been changed, please give it another look.

spec/02-integration/09-hybrid_mode/08-lazy_export.lua
```diff
@@ -0,0 +1,139 @@
...
+    cluster_mtls = "pki",
+    cluster_ca_cert = "spec/fixtures/ocsp_certs/ca.crt",
+  }))
+  admin_client = assert(helpers.admin_client())
+end
+
+local n = 0
+local function touch_config()
+  n = n + 1
+  admin_client:send({
```
flrgh, line 23: The `:send()` method on our test http client doesn't do any error-checking. Adding an assertion or two here would help to make this test suite more debug-able.

kong/clustering/wrpc_control_plane.lua
```diff
@@ -153,7 +156,7 @@ function _M:export_deflated_reconfigure_payload()
 end
 
 function _M:push_config_one_client(client)
-  if not self.config_call_rpc or not self.config_call_args then
+  if isempty(self.clients) or not self.config_call_rpc or not self.config_call_args then
```
fffonion, line 159: table_clone

fffonion, line 159: a comment on why do we need `isempty()` scenairo here will be helpful for future us

spec/02-integration/09-hybrid_mode/08-lazy_export.lua
```diff
@@ -0,0 +1,139 @@
...
+    describe("no DP", function ()
+      setup(function()
+        cp(strategy)
+      end)
+      teardown(function ()
+        helpers.stop_kong()
+      end)
+      it("test", function ()
+        touch_config()
+        assert.logfile().has.no.line("exporting config for legacy protocol", true)
```
flrgh, line 81: The negative error.log checks always make me nervous, because it's so easy to for the test to pass improperly if the log entry we're looking for simply hasn't been logged _yet_, or if a future, unrelated commit changes the log message ever so slightly such that our assertion would never match it.

So if there's anything from #9369 that I think is worth making the case to bring over to this PR it would be the logging+testing changes. Adding/updating some ngx.log calls allowed me to turn all of the negative assertions to positive ones. WDYT?

kong/clustering/wrpc_control_plane.lua
dndx: Added, thanks!

spec/02-integration/09-hybrid_mode/08-lazy_export.lua
dndx: @flrgh Could you make the logging change? As long as we are not spamming the log files with the change I'm all open for it.

fffonion (approved):

spec/02-integration/09-hybrid_mode/08-lazy_export.lua
flrgh: done.

mayocream (approved):

spec/02-integration/09-hybrid_mode/01-sync_spec.lua
```diff
@@ -699,14 +699,14 @@ for _, strategy in helpers.each_strategy() do
           helpers.wait_until(function()
             local proxy_client = helpers.http_client("127.0.0.1", 9002)
             -- serviceless route should return 503 instead of 404
-            res = proxy_client:get("/2")
+            res = proxy_client:get("/5")
```
Suika-Kong, line 702: I'm not very sure about the purpose of this test. It looks like that, it wants to make sure when 2 is synced, all changes should be synced(2 will not be synced individually).
Or, should this test is checking when a change takes effect, wether other changes before that works?

PR #9369, opened 2022-09-01
chore(clustering) improve on-demand export

flrgh:
Changes from #9362:

* don't clear config export cache unconditionally (in order to survive DB outages--see 310722659eead1a6aa8a92a6cf025b5708bb0698)
* check for clients in the push config event handler
* update tests to positive `assert.logfile().has.line(...)` conditions (less likely to cause a false negative)


kong/clustering/control_plane.lua
```diff
@@ -33,6 +33,7 @@ local table_remove = table.remove
 local sub = string.sub
 local gsub = string.gsub
 local deflate_gzip = utils.deflate_gzip
+local nkeys = require "table.nkeys"
```
bungle, line 36: What about:
https://github.com/openresty/luajit2#tableisempty

bungle, line 36: How about:
https://github.com/openresty/luajit2#tableisempty

kong/clustering/control_plane.lua
```diff
@@ -180,7 +181,7 @@ end
 _M._update_compatible_payload = update_compatible_payload
 
 function _M:export_deflated_reconfigure_payload()
-  ngx_log(ngx_DEBUG, _log_prefix, "exporting config for legacy protocol")
+  ngx_log(ngx_DEBUG, _log_prefix, "exporting config")
```
flrgh, line 184: note for reviewer: `_log_prefix` is unique from `control_plane.lua` to `wrpc_control_plane.lua`, so the extra verbiage wasn't needed here (and it makes our test assertions more readable without it).

flrgh:
Closing in favor of #9368

PR #9371, opened 2022-09-02
fix(auth-plugins): auth plugins cache inconsistency while anonymous was configured as username

vm-001:
### Summary

As Kong auth plugins support `conifg.anonymous` to be configured as the username of consumer, once a consumer has been cached with the username as part of the cache key, this consumer cache won't be invalidated by the default cache invalidation event handler after the consumer has been deleted.

Considers of the conifg.anonymous exists in several auth plugins, so I added a common event handler to invalidate the extra cache rather than in each plugin.

### Issue reference
FTI-4287

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -540,7 +540,7 @@ local function register_events()
     if old_entity then
       old_username = old_entity.username
       if old_username and old_username ~= null and old_username ~= "" then
-        kong.cache:invalidate(kong.db.consumers:cache_key(old_entity.username))
+        kong.cache:invalidate(kong.db.consumers:cache_key(old_username))
       end
     end
 
@@ -548,7 +548,7 @@ local function register_events()
     if entity then
       local username = entity.username
       if username and username ~= null and username ~= "" and username ~= old_username then
-        kong.cache:invalidate(kong.db.consumers:cache_key(entity.username))
+        kong.cache:invalidate(kong.db.consumers:cache_key(username))
       end
     end
   end, "crud", "consumers")
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -40,6 +40,7 @@ local clear_header      = ngx.req.clear_header
 local http_version      = ngx.req.http_version
 local unpack            = unpack
 local escape            = require("kong.tools.uri").escape
+local null              = ngx.null
 
 
 local is_http_module   = subsystem == "http"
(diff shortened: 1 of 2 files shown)


bungle:
This can be optimized (the Kong EE OpenID Connect plugin has an optimized version of this)

vm-001:
> This can be optimized (the Kong EE OpenID Connect plugin has an optimized version of this)

Nice. https://github.com/Kong/kong/pull/9371/commits/d9a740ae9748ffc1f19db625178aa10b9fce5861

kong/runloop/handler.lua
```diff
@@ -526,6 +527,32 @@ local function register_events()
...
+  -- so add an event handler to invalidate the extra cache in case of data inconsistency
+  worker_events.register(function(data)
+    workspaces.set_workspace(data.workspace)
+
+    local old_entity = data.old_entity
+    local old_username
+    if old_entity then
+      old_username = old_entity.username
+      if old_username and old_username ~= null and old_username ~= "" then
+        kong.cache:invalidate(kong.db.consumers:cache_key(old_entity.username))
```
chronolaw, line 543: here we should use cached var `old_username` directly.

kong/runloop/handler.lua
```diff
@@ -526,6 +527,32 @@ local function register_events()
...
+      end
+    end
+
+    local entity = data.entity
+    if entity then
+      local username = entity.username
+      if username and username ~= null and username ~= "" and username ~= old_username then
+        kong.cache:invalidate(kong.db.consumers:cache_key(entity.username))
```
chronolaw, line 551: Same as above, use `username`.

spec/03-plugins/25-oauth2/03-access_spec.lua
```diff
@@ -4067,6 +4067,85 @@ describe("Plugin: oauth2 [#" .. strategy .. "]", function()
...
+    local admin_client
+    local anonymous
+
+    lazy_setup(function()
+      anonymous = assert(admin_api.consumers:insert {
+        username = "test",
+      })
+
+      local service = assert(admin_api.services:insert({
+        url = "https://mockbin.org"
```
chronolaw, line 4082: need a `,` at the end of the line.

spec/03-plugins/25-oauth2/03-access_spec.lua
```diff
@@ -4067,6 +4067,85 @@ describe("Plugin: oauth2 [#" .. strategy .. "]", function()
...
+      }))
+
+      local route = assert(admin_api.routes:insert({
+        hosts      = { "anonymous-with-username.com" },
+        protocols  = { "http", "https" },
+        service    = service
```
chronolaw, line 4088: need a `,` at the end of the line.

spec/03-plugins/25-oauth2/03-access_spec.lua
```diff
@@ -4067,6 +4067,85 @@ describe("Plugin: oauth2 [#" .. strategy .. "]", function()
...
+      }))
+    end)
+
+    before_each(function()
+      proxy_client = helpers.proxy_client()
+      admin_client = helpers.admin_client()
+    end)
+
+    after_each(function()
+      if proxy_client then proxy_client:close() end
```
chronolaw, line 4105: Not recommended, we should not write `if-then` in one line.

spec/03-plugins/25-oauth2/03-access_spec.lua
```diff
@@ -4067,6 +4067,85 @@ describe("Plugin: oauth2 [#" .. strategy .. "]", function()
...
+        local res = assert(proxy_client:send {
+          method = "GET",
+          path = "/request",
+          headers = {
+            ["Host"] = "anonymous-with-username.com",
+          }
+        })
+        assert.response(res).has.status(200)
+        local body = assert.response(res).has.jsonbody()
+        assert.are.equal("true", body.headers["x-anonymous-consumer"])
```
chronolaw, line 4127: Should we use `assert.equal`?

kong/runloop/handler.lua
vm-001: Nice catch.

spec/03-plugins/25-oauth2/03-access_spec.lua
vm-001: Maybe both are fine? I see a lot of usages of `assert.are.equal`

bungle (approved):

chronolaw (approved):

PR #5977, opened 2020-06-02
chore(ci) run CI against all branches / PR's

hutchic:
Need more discussion or results: do we need to run CI against all branches or only PR's and specific branches?

PR #9375, opened 2022-09-02
chore(ci) limit test runs for push events

flrgh:
This should reduce duplicate test runs ([example](https://github.com/Kong/kong/pull/9371))
image


.github/workflows/build_and_test.yml
```diff
@@ -5,6 +5,10 @@ on:
     paths-ignore:
     # ignore top-level markdown files (CHANGELOG.md, README.md, etc.)
     - '*.md'
+    branches:
+    - master
+    - next/*
+    - test-please/*
```
aboudreault, line 11:
```suggestion
    - release/*
```

flrgh:
Hmm, so #5977 has been brought to my attention, and it seems that running tests on all branch activity is something we don't want to give up, so maybe this isn't that simple.

Ideally we would be able to have both: test runs on each commit/push and on each pull request update, but _without_ running 2x the tests for the same branch+commit, but GHA doesn't make that easy.

curiositycasualty (approved):

hutchic (approved):

.github/workflows/build_and_test.yml
dndx: If we do this, wouldn't branch CI no longer be run ad we are forced to create a Draft PR during development?

aboudreault: ah, the sugggestion is not on the right line... Michael fixed it properly and replace `next/*` instead.

aboudreault: and like kong-ee, if a single branch need to be tested without a PR, it can trigger CI with the `test-please/` prefix

hbagdi:
@gszr Can you please take care of merging this one in?

hbagdi:
@flrgh You should be good to merge this one.

fffonion (approved):

PR #9377, opened 2022-09-05
docs(migration) min ver for bg migration support

Suika-Kong:
We assume the minimal version blue-green migration should be 2.8.2.


Suika-Kong:
Closing because we don't support blue/green migration from 2.8 to 3.0 for EE.

PR #9379, opened 2022-09-05
feat(conf): add cluster_ca_cert, ssl_dhparam, lua_ssl_trusted_certificate

samugi:
### Summary

* Adds `cluster_ca_cert`, `ssl_dhparam`, `lua_ssl_trusted_certificate` to the supported properties that can be assigned via environment variables or vaults
* Add tests for `cluster_cert` and `cluster_ca_cert` when defined but paths don't exist
* Add tests to ensure files are created correctly based on the content provided

(continuation FT-3322)


kong/conf_loader/init.lua
```diff
@@ -626,17 +626,24 @@ end
 
 
 local function try_base64_decode(vals)
+  -- names that we should not attempt decoding
+  local decode_blacklist = {
+    system = "system"
```
samugi, line 631: `system` was considered valid base64 for `+-z`. Not sure if there are other such words we should pay attention to.

samugi: Resolved by moving the base64 decoding *after* the path & values checks.

bungle (approved):
Great work @samugi!

Issue #7641, opened 2021-06-04
requestContext attribute is missing in the request for API Gateway compatibility


alexey-nextreason:
The `requestContext` attribute is missing on the event payload which should be present as described here https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format Some of the ASGI adapters like Mangum depend on the presence of that field and can be a blocker for the usage of this plugin. Actual implementation https://github.com/Kong/kong-plugin-aws-lambda/blob/master/kong/plugins/aws-lambda/aws-serializer.lua#L83

windmgc:
Resolved in https://github.com/Kong/kong/pull/9380

PR #9380, opened 2022-09-06
feat(aws-lambda): add requestContext field into aws-gateway-compatible input

windmgc:
### Summary

This PR adds a new field `requestContext` into the lambda input data when the aws-lambda plugin is working in `awsgateway_compatible` mode.

The `requestContext` is part of the input data in lambda proxy integration, according to [this AWS document page](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format), and it contains many fields containing request related information on AWS API Gateway. 

The aws-lambda plugin in Kong provides an AWS API Gateway compatible mode that can generate an input data similar with 
[the example provided in this doc](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format), but it does not contain the `requestContext` field. When users are writing their lambda functions with some libraries like [aws-serverless-java-container](https://github.com/awslabs/aws-serverless-java-container), it will check if the input data have the `requestContext` field, and if it does not exist, the request will be regarded as malformed.(For example, see [this code snippet](https://github.com/awslabs/aws-serverless-java-container/blob/98482d485dadaf7212141f162e7e35f4ea6c7b6f/aws-serverless-java-container-core/src/main/java/com/amazonaws/serverless/proxy/internal/servlet/AwsProxyHttpServletRequestReader.java#L47-L49))

To make the input data sound more "compatible", this PR adds the `requestContext` field into the input data. Note that many fields in the formal `requestContext` cannot be fetched by Kong(like IAM-related information or Cognito-related auth info), so we're just providing a **best-effort delivery** here. We cannot make assurance that the `requestContext` provided by Kong has every information that the lambda function needs, and it is decided by the lambda function whether the whole payload containing `requestContext` can work normally or not. So the feature is just a good-to-have one.

### Full changelog

* Add `requestContext` field into `awsgateway_compatible` input data
* Add related tests

### Issue reference

This should fix [FTI-2620](https://konghq.atlassian.net/browse/FTI-2620), and #7641, in some cases.

### DIFF
--- a/kong/plugins/aws-lambda/aws-serializer.lua
+++ b/kong/plugins/aws-lambda/aws-serializer.lua
@@ -94,7 +94,7 @@ return function(ctx, config)
     local domainName = var.host
     local domainPrefix = split(domainName, ".")[1]
     local identity = {
-      sourceIp = var.remote_addr,
+      sourceIp = var.realip_remote_addr or var.remote_addr,
       userAgent = headers["user-agent"],
     }
     local requestId = var.request_id
--- a/kong/plugins/aws-lambda/aws-serializer.lua
+++ b/kong/plugins/aws-lambda/aws-serializer.lua
@@ -3,11 +3,16 @@
 -- https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
 
 local request_util = require "kong.plugins.aws-lambda.request-util"
+local pl_stringx = require("pl.stringx")
+local date = require "date"
 
 local EMPTY = {}
 
+local split = pl_stringx.split
 local ngx_req_get_headers  = ngx.req.get_headers
 local ngx_req_get_uri_args = ngx.req.get_uri_args
+local ngx_get_http_version = ngx.req.http_version
+local ngx_req_start_time = ngx.req.start_time
 local ngx_encode_base64    = ngx.encode_base64
 
 return function(ctx, config)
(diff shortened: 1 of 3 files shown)


kong/plugins/aws-lambda/aws-serializer.lua
```diff
@@ -81,6 +86,39 @@ return function(ctx, config)
...
 
+  local requestContext
+  do
+    local http_version = ngx_get_http_version()
+    local protocol = http_version and 'HTTP/'..http_version or nil
+    local httpMethod = var.request_method
+    local domainName = var.host
+    local domainPrefix = split(domainName, ".")[1]
+    local identity = {
+      sourceIp = var.remote_addr,
```
Tieske, line 97: is this IP expected? considering trusted IP's etc.

kong/plugins/aws-lambda/aws-serializer.lua
```diff
@@ -81,6 +86,39 @@ return function(ctx, config)
...
+    }
+    local requestId = var.request_id
+    local start_time = ngx_req_start_time()
+    local requestTime = date(start_time):fmt("${iso}")
+    local requestTimeEpoch = start_time * 1000
+
+    -- Kong does not have the concept of stage, so we just let resource path be the same as path
+    local resourcePath = path
+
+    requestContext = {
```
Tieske, line 108: why the locals, most or not used, we can add the values in-line to the table below?

(nitpick: feel free to ignore)

kong/plugins/aws-lambda/aws-serializer.lua
windmgc: Good point, I double-checked the AWS doc and it says:

> $context.identity.sourceIp | -- |  The source IP address of the immediate TCP connection making the request to API Gateway endpoint.

Maybe we should keep the same behaviour as what AWS does, so the real ip module should not be taken into consideration, it should be `sourceIp = var.realip_remote_addr or var.remote_addr`

Tieske (approved):

PR #9384, opened 2022-09-06
fix(external-plugins): handle failing instance creation

samugi:
### Summary

This is a fix for https://github.com/Kong/kong/issues/9301 / FT-3319: when a request made the plugin fail, the creation of a new instance was not handling a failed connection properly.
Upon failure to connect to the socket, it was possible to end up in a state where the instance existed without an instance id,
causing an infinite loop and high CPU utilization.


gszr (approved):
Good catch!

PR #9386, opened 2022-09-06
docs(kong.conf) add `router_flavor` documentation

dndx:


aboudreault (approved):

kong.conf.default
```diff
@@ -1447,6 +1447,34 @@
...
                                  # to each individual worker.
 
+#router_flavor = traditional_compatible
+                                 # Selects the router implementation to use when
+                                 # performing request routing. Incremental router
+                                 # rebuild is available when the flavor is set
+                                 # to either `expressions` or
+                                 # `traditional_compatible` which could
+                                 # significantly shorten rebuild time for large
+                                 # number of routes.
```
hbagdi, line 1457: The maintainers of the project have an understanding of what "incremental router rebuild" is but the operator does not and should not need to either. That is more detail than necessary.

I'd recommend rewording this as "recommend values are foo and bar for performance with a large number of routes".

kong.conf.default
```diff
@@ -1447,6 +1447,34 @@
...
+                                 #   is not visible.
+                                 # - `expression`: the DSL based expression router engine
+                                 #   will be used under the hood. Traditional router
+                                 #   config interface is not visible and you must write
+                                 #   Router Expression manually and provide them in the
+                                 #   `expression` field on the `Route` object.
+                                 # - `traditional`: the pre-3.0 Router engine will be
+                                 #   used. Config interface will be the same as
+                                 #   pre-3.0 Kong and the `expression` field on the
+                                 #   `Route` object is not visible.
```
hbagdi, line 1476: Is the traditional mode not recommended and only meant as an escape hatch? Should we document that "traditional" option is deprecated and will be removed in a future release?

kong.conf.default
dndx: I feel hesitant to make such change. Even for shorter number of routes, the new router is still recommended as it also runs faster in these cases. We should not pitch the new router as just a performance improvement only because that is missing the main point of this change.

"which could significantly shorten rebuild time for large number of routes" should explain the term "incremental router rebuild" well. We even mentioned this term to some customers in the past and it does not seems to be very difficult to undersatand.

kong.conf.default
dndx: This has been added, thanks!

PR #9389, opened 2022-09-07
fix(router) prefix match no longer normalize input

Suika-Kong:
We should make it consistent with regex paths, so we do not normalize user input paths.

This PR also includes migration.

Remember to add downgrade support for this to EE.

### DIFF
--- a/kong/db/migrations/migrate_path_280_300.lua
+++ b/kong/db/migrations/migrate_path_280_300.lua
@@ -2,6 +2,8 @@ local find = string.find
 local upper = string.upper
 local re_find = ngx.re.find
 
+local normalize = require("kong.tools.uri").normalize
+
 -- We do not percent decode route.path after 3.0, so here we do 1 last time for them
 local normalize_regex
 do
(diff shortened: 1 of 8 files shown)


kong/db/migrations/migrate_path_280_300.lua
```diff
@@ -74,13 +76,14 @@ local function is_not_regex(path)
   return (re_find(path, [[[a-zA-Z0-9\.\-_~/%]*$]], "ajo"))
 end
 
-local function migrate_regex(reg)
+local function migrate_path(reg)
```
chronolaw, line 79: The param `reg` should not be `reg`, maybe `str`?

kong/db/migrations/migrate_path_280_300.lua
```diff
@@ -74,13 +76,14 @@ local function is_not_regex(path)
...
   if is_not_regex(reg) then
-    return reg, false
+    local ret = normalize(reg, true)
```
chronolaw, line 81: `ret` is not a good name, maybe `str`?

chronolaw:
Should we change the line 235 in `atc_compat.lua`, and remove localization of `normalize` ?

kong/router/traditional.lua
```diff
@@ -424,7 +424,7 @@ local function marshall_route(r)
 
           local uri_t = {
             is_prefix = true,
-            value     = normalize(path, true),
+            value     = path,
```
chronolaw, line 427: Also remove localization of `normalize` for lint reason.

chronolaw:
Could we add some special test cases for this change? for example, path likes `%2f%2e`.

chronolaw (approved):

Suika-Kong:
> Could we add some special test cases for this change? for example, path likes `%2f%2e`.

added

bungle (approved):

kikito:
Notice that this PR currently targets `master`. In order for it to be included in 3.0 two things need to happen, just to make sure that we don't miss any steps:
* It needs to be cherry-picked on the [kong/release/3.0.x](https://github.com/Kong/kong/tree/release/3.0.x) branch (for release on CE)
* A cherry-pick PR needs to be sent against the 3.0 EE branch as well

PR #9396, opened 2022-09-07
fix(runloop) do not reset `*:version` to `init` when worker respawns,

dndx:
fixes FT-3328

Previously, when worker respawns, the `router:version` and `plugins_iterator:version` keys in the cache is incorrectly set to `init`, this causes the newly spawned worker to not rebuild the router/iterator and always use the router/iterator from when master process was created.


spec/02-integration/05-proxy/02-router_spec.lua
```diff
@@ -2308,6 +2308,33 @@ for _, strategy in helpers.each_strategy() do
...
+      }))
+      assert.res_status(201, res)
+      admin_client:close()
+
+      assert(helpers.signal_workers(nil, "-TERM"))
+
+      proxy_client:close()
+      proxy_client = helpers.proxy_client()
+
+      local res = assert(proxy_client:send {
```
flrgh, line 2328: Is there any risk that this request could be handled by a worker that hasn't been killed yet?

flrgh: Seems like a low risk given SIGTERM and not SIGQUIT, but `helpers.get_kong_workers` and `helpers.wait_until_no_common_workers` might be useful.

flrgh (approved):
:+1: Code looks good.

I left a question/suggestion about testing. I'm not really that concerned about the test covering the behavior, but I _am_ mildly concerned that the test could be flaky in CI without some retry logic to account for new workers being slow to start up.

bungle:
Do we need to check that does this affect EE smart route validation?

spec/02-integration/05-proxy/02-router_spec.lua
dndx: It does not seems to be flappy according to my tests on CI. It always catches the issue and never fails due to connection reset.

dndx:
@bungle I will be surprised if it does, because the old function had `if kong.configuration.role ~= "control_plane" then` check as well, I think for the smart validation the `router:version` is never set, which is fine as the Admin API will rebuild it with a random UUID before using the router.

chronolaw (approved):

PR #9400, opened 2022-09-08
fix(request-termination): status_code cannot be set as null

ms2008:
### Summary

`status_code` should not be allowed to be set to null, otherwise a 500 error will be returned.

### Full changelog

* Do not allow set `status_code` as null

FTI-4291

### DIFF
--- a/spec/03-plugins/14-request-termination/01-schema_spec.lua
+++ b/spec/03-plugins/14-request-termination/01-schema_spec.lua
@@ -1,5 +1,6 @@
 local schema_def = require "kong.plugins.request-termination.schema"
 local v = require("spec.helpers").validate_plugin_config_schema
+local null = ngx.null
 
 describe("Plugin: request-termination (schema)", function()
   it("should accept a valid status_code", function()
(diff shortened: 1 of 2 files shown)


spec/03-plugins/14-request-termination/01-schema_spec.lua
```diff
@@ -1,5 +1,6 @@
 local schema_def = require "kong.plugins.request-termination.schema"
 local v = require("spec.helpers").validate_plugin_config_schema
+local null = ngx.null
```
chronolaw, line 3: It seems that this localization is unnecessary.

ms2008: It will be used in this line https://github.com/Kong/kong/blob/0ee88ee3492f03a7f47d3d2f615eae3dae5ee461/spec/03-plugins/14-request-termination/01-schema_spec.lua#L29

chronolaw: Sorry, my meaning is that we can use `ngx.null` in this line, need not to localize it.

ms2008: I don't think it matters as it seems like a trivial style.🙂

samugi (approved):

dndx (approved):

hbagdi:
@dndx Please hit the merge button.

PR #9407, opened 2022-09-08
tests: fix some flaky tests

ADD-SP:
_FT-3340, FT-3341, FT-3342, FT-3343, FT-3344, FT-3346, FT-3347, FT-3348_

### DIFF
--- a/spec/03-plugins/02-udp-log/01-udp-log_spec.lua
+++ b/spec/03-plugins/02-udp-log/01-udp-log_spec.lua
@@ -186,7 +186,7 @@ for _, strategy in helpers.each_strategy() do
       end)
     end)
 
-    it("logs proper latencies (#grpc) #flaky", function()
+    it("logs proper latencies (#grpc)", function()
       local udp_thread = helpers.udp_server(UDP_PORT)
 
       -- Making the request
@@ -197,9 +197,10 @@ for _, strategy in helpers.each_strategy() do
         },
         opts = {
           ["-authority"] = "udp_logging_grpc.test",
+          ["-H"] = "'Content-Type: text/plain'",
         }
       })
-      assert.truthy(ok)
+      assert.truthy(ok, resp)
       assert.truthy(resp)
 
       -- Getting back the UDP server input
@@ -222,7 +223,7 @@ for _, strategy in helpers.each_strategy() do
       assert.True(is_latencies_sum_adding_up)
     end)
 
-    it("logs proper latencies (#grpcs) #flaky", function()
+    it("logs proper latencies (#grpcs)", function()
       local udp_thread = helpers.udp_server(UDP_PORT)
 
       -- Making the request
@@ -233,9 +234,10 @@ for _, strategy in helpers.each_strategy() do
         },
         opts = {
           ["-authority"] = "udp_logging_grpcs.test",
+          ["-H"] = "'Content-Type: text/plain'",
         }
       })
-      assert.truthy(ok)
+      assert.truthy(ok, resp)
       assert.truthy(resp)
 
       -- Getting back the UDP server input
--- a/spec/03-plugins/05-syslog/01-log_spec.lua
+++ b/spec/03-plugins/05-syslog/01-log_spec.lua
@@ -5,7 +5,7 @@ local pl_stringx = require "pl.stringx"
 
 
 for _, strategy in helpers.each_strategy() do
-  describe("#flaky Plugin: syslog (log) [#" .. strategy .. "]", function()
+  describe("Plugin: syslog (log) [#" .. strategy .. "]", function()
     local proxy_client
     local proxy_client_grpc
     local platform
@@ -160,7 +160,7 @@ for _, strategy in helpers.each_strategy() do
 
     local function do_test(host, expecting_same, grpc)
       local uuid = utils.uuid()
-      local resp
+      local ok, resp
 
       if not grpc then
         local response = assert(proxy_client:send {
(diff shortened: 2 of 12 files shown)


spec/helpers.lua
```diff
@@ -1709,6 +1712,60 @@ local function wait_for_all_config_update(timeout, admin_client_timeout, forced_
...
+-- 
+-- created: wait for file to be created
+-- 
+-- deleted: wait for file to be deleted
+-- 
+-- socket:  wait for file to become a socket file
+-- 
+-- @tparam string path the file path
+-- @tparam[opt=10] number timeout maximum time to wait
+local function wait_for_file_status(status, path, timeout)
```
fffonion, line 1731: consider use `stat` call, you can use `lfs.attribute(path)` (https://lunarmodules.github.io/luafilesystem/manual.html#introduction). and let's not make `socket` a parrelle status as `created`, instead we can say 
`wait_for_file(mode, path, timeout)` then we can further simplify this function .

spec/03-plugins/02-udp-log/01-udp-log_spec.lua
```diff
@@ -197,9 +197,10 @@ for _, strategy in helpers.each_strategy() do
         },
         opts = {
           ["-authority"] = "udp_logging_grpc.test",
+          ["-H"] = "'Content-Type: text/plain'",
```
fffonion, line 200: just curious why would this fix the test?

spec/03-plugins/05-syslog/01-log_spec.lua
```diff
@@ -201,10 +202,57 @@ for _, strategy in helpers.each_strategy() do
 
         resp = stdout
       elseif expecting_same then
-        local _, _, stdout = assert(helpers.execute("find /var/log -type f -mmin -5 2>/dev/null | xargs grep -l " .. uuid))
-        assert.True(#stdout > 0)
-
-        resp = stdout
+        -- wait for file writing
+        helpers.pwait_until(function()
+          local _, _, stdout = assert(helpers.execute("sudo find /var/log -type f -mmin -5 | grep syslog"))
```
fffonion, line 207: is the file always going to be `/var/log/syslog` or if we are expecting other name?

spec/helpers.lua
```diff
@@ -1586,10 +1586,13 @@ end
...
 -- @function wait_for_all_config_update
 -- @tparam[opt=30] number timeout maximum time to wait
--- @tparam[opt] number admin_client_timeout, to override the default timeout setting
+-- @tparam[opt] number admin_client_timeout, to override the default timeout setting of admin client
 -- @tparam[opt] number forced_admin_port to override the default port of admin API
+-- @tparam[opt] number proxy_client_timeout, to override the default timeout setting of proxy client
+-- @tparam[opt] number forced_proxy_port to override the default port of proxy client
 -- @usage helpers.wait_for_all_config_update()
-local function wait_for_all_config_update(timeout, admin_client_timeout, forced_admin_port)
+local function wait_for_all_config_update(timeout, admin_client_timeout, forced_admin_port,
```
fffonion, line 1594: not a blocker but for future us: with the argument list becoming long and user might not understand what each one means from its type, it will make more sense to make the function accept a table, so you can use key/value to make arguments more readable. (exceptions apply for performance critical path.)

spec/helpers.lua
ADD-SP: done

spec/03-plugins/02-udp-log/01-udp-log_spec.lua
ADD-SP: I don't remember, I think it was because of the `grpcurl` error message that prompted me.

spec/03-plugins/05-syslog/01-log_spec.lua
ADD-SP: I'm not sure, the original logic was also to find suspicious files, so I just refined that logic.

spec/helpers.lua
ADD-SP: done

spec/helpers.lua
```diff
@@ -1585,12 +1586,27 @@ end
...
-local function wait_for_all_config_update(timeout, admin_client_timeout, forced_admin_port)
-  timeout = timeout or 30
+local function wait_for_all_config_update(opt)
+  opt = opt or {}
+
+  local timeout = opt.timeout or 30
+  local admin_client_timeout = opt.admin_client_timeout
+  local forced_admin_port = opt.forced_admin_port
+  local proxy_client_timeout = opt.proxy_client_timeout
+  local forced_proxy_port = opt.forced_proxy_port
```
mayocream, line 1609: Use proper ldoc comments and use `opts` instead of `opt`.
```suggestion
-- @tparam[opt] table opts a table contains params
-- @tparam[opt=30] number timeout maximum seconds to wait, defatuls is 30
-- @tparam[opt=30] number admin_client_timeout to override the default timeout setting
-- @tparam[opt] number forced_admin_port to override the default Admin API port
-- @tparam[opt=30] number proxy_client_timeout to override the default timeout setting
-- @tparam[opt] number forced_proxy_port to override the default proxy port
-- @usage helpers.wait_for_all_config_update()
local function wait_for_all_config_update(opts)
  opts = opts or {}

  local timeout = opts.timeout or 30
  local admin_client_timeout = opts.admin_client_timeout
  local forced_admin_port = opts.forced_admin_port
  local proxy_client_timeout = opts.proxy_client_timeout
  local forced_proxy_port = opts.forced_proxy_port
```

ADD-SP: done

mayocream (approved):

dndx (approved):

PR #9412, opened 2022-09-09
fix(declarative) migration failed on nested route

Suika-Kong:
when you put a route nested in service it fails to migrate on the fly.

### DIFF
--- a/kong/db/declarative/init.lua
+++ b/kong/db/declarative/init.lua
@@ -9,7 +9,7 @@ local tablex = require "pl.tablex"
 local constants = require "kong.constants"
 local txn = require "resty.lmdb.transaction"
 local lmdb = require "resty.lmdb"
-local on_the_fly_migration = require "kong.db.declarative.migrations"
+local on_the_fly_migration = require "kong.db.declarative.migrations.route_path"
 local to_hex = require("resty.string").to_hex
 local resty_sha256 = require "resty.sha256"
 
@@ -241,12 +241,12 @@ function Config:parse_table(dc_table, hash)
     error("expected a table as input", 2)
   end
 
-  on_the_fly_migration(dc_table)
-
   local entities, err_t, meta = self.schema:flatten(dc_table)
   if err_t then
     return nil, pretty_print_error(err_t), err_t
   end
+
+  on_the_fly_migration(entities, dc_table._format_version)
 
   yield()
 
--- a/kong/db/declarative/migrations/route_path.lua
+++ b/kong/db/declarative/migrations/route_path.lua
@@ -1,8 +1,7 @@
 local migrate_path = require "kong.db.migrations.migrate_path_280_300"
 
-return function(tbl)
-  local version = tbl._format_version
-  if not (version == "1.1" or version == "2.1") then
+return function(tbl, version)
+  if not tbl or not (version == "1.1" or version == "2.1") then
     return
   end
 
@@ -15,7 +14,7 @@ return function(tbl)
 
   for _, route in pairs(routes) do
     local paths = route.paths
-    if not paths then
+    if not paths or paths == ngx.null then
       -- no need to migrate
       goto continue
     end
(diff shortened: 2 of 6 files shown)


kong/db/declarative/migrations/route_path.lua
```diff
@@ -1,7 +1,6 @@
 local migrate_path = require "kong.db.migrations.migrate_path_280_300"
 
-return function(tbl)
-  local version = tbl._format_version
+return function(tbl, version)
```
chronolaw, line 3: how about `local version = version or tbl._format_version`

Suika-Kong: > how about `local version = version or tbl._format_version`

We cannot get the version from this table. The version field is removed by the `flatten` call.

kong/db/declarative/init.lua
```diff
@@ -241,9 +241,8 @@ function Config:parse_table(dc_table, hash)
     error("expected a table as input", 2)
   end
 
-  on_the_fly_migration(dc_table)
-
   local entities, err_t, meta = self.schema:flatten(dc_table)
+  on_the_fly_migration(entities, dc_table._format_version)
```
chronolaw, line 245: Could we move this line after `if err_t then ... end`?

spec/01-unit/01-db/01-schema/11-declarative_config/04-on-the-fly-migration_spec.lua
```diff
@@ -120,6 +150,7 @@ describe("declarative config: on the fly migration", function()
       else
         assert.same({ "/prefix", "~/regex.+", }, sorted.routes[1].paths)
       end
+      end
```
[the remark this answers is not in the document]
Suika-Kong: I intentionally omitted this indentation to reduce the lines of diff

kong/db/declarative/init.lua
Suika-Kong: Yes, it should be after that.

spec/01-unit/01-db/01-schema/11-declarative_config/04-on-the-fly-migration_spec.lua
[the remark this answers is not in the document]
chronolaw: I understood your feel, maybe we could fix this style issue when we have enough time.

ADD-SP (approved):
I have tested this PR locally, this bug was fixed. I think we can merge this PR after CIs green.

```yml
_format_version: "2.1"
_transform: true

services:
- name: nginx-service
  url: http://localhost:80
  routes:
  - name: nginx-regex-route
    paths:
    - "/nginx-regex/\\d+/test"
  - name: nginx-prefix-route
    paths:
    - /nginx-prefix
  - name: nginx-chinese-route
    paths:
    # /你好
    - /%E4%BD%A0%E5%A5%BD
```
```shell
$ curl -I localhost:8000/nginx-prefix
HTTP/1.1 200 OK
Content-Type: application/octet-stream
Connection: keep-alive
Date: Fri, 09 Sep 2022 08:52:17 GMT
X-Kong-Upstream-Latency: 0
X-Kong-Proxy-Latency: 1
Via: kong/3.0.0

$ curl -I localhost:8000/你好
HTTP/1.1 200 OK
Content-Type: application/octet-stream
Connection: keep-alive
Date: Fri, 09 Sep 2022 08:52:29 GMT
X-Kong-Upstream-Latency: 0
X-Kong-Proxy-Latency: 1
Via: kong/3.0.0

$ curl -I localhost:8000/%E4%BD%A0%E5%A5%BD
HTTP/1.1 200 OK
Content-Type: application/octet-stream
Connection: keep-alive
Date: Fri, 09 Sep 2022 08:52:36 GMT
X-Kong-Upstream-Latency: 0
X-Kong-Proxy-Latency: 1
Via: kong/3.0.0

$ curl -I localhost:8000/nginx-regex/123/test
HTTP/1.1 200 OK
Content-Type: application/octet-stream
Connection: keep-alive
Date: Fri, 09 Sep 2022 08:52:45 GMT
X-Kong-Upstream-Latency: 0
X-Kong-Proxy-Latency: 1
Via: kong/3.0.0
```

chronolaw (approved):

Suika-Kong:
`spec/03-plugins/06-statsd/01-log_spec.lua:1216` seems to be a flaky test. I ran 6 times in my local env and it fails for 3 times.

kikito (approved):

PR #9428, opened 2022-09-13
chore(release): bump Kong's version number to 3.1.0

hanshuebner:


aboudreault (approved):

gszr (approved):
Thinking out loud: should we also use the `suffix` field in master so that it doesn't get confusing to users? Right now it reads as if 3.0.1 was already released.

kong-3.0.1-0.rockspec
```diff
@@ -1,10 +1,10 @@
 package = "kong"
-version = "3.0.0-0"
+version = "3.0.1-0"
```
kikito, line 2: I think we should be bumping to "the next version that can be released from the current master".

`release/3.0.x` is a long-lived branch. All the members of the `3.0.x` family (including `3.0.1`) will be commits on that branch.

The next release that will come out from `master` is `3.1.0`, so I think that is the correct version to "preemptively bump to".
```suggestion
version = "3.1.0-0"
```

(rockspec filename would need to be updated as well)

kong-3.0.1-0.rockspec
```diff
@@ -1,10 +1,10 @@
...
 rockspec_format = "3.0"
 supported_platforms = {"linux", "macosx"}
 source = {
   url = "git+https://github.com/Kong/kong.git",
-  tag = "3.0.0"
+  tag = "3.0.1"
```
kikito, line 7:
```suggestion
  tag = "3.1.0"
```

kong/meta.lua
```diff
@@ -1,7 +1,7 @@
 local version = setmetatable({
   major = 3,
   minor = 0,
-  patch = 0,
+  patch = 1,
```
kikito, line 4: This also would need to be changed to 3.1.0

kikito (changes requested):

kikito (approved):

PR #9432, opened 2022-09-14
perf(clustering): use privileged worker for control plane connection

bungle:
### Summary

Data plane's connection to control plane is moved to a privileged worker process, including:
- maintaining websocket (wrpc) connection and data transfer
- decompression of received data
- json decoding of the received data
- validation and flattening of received data
- inserting data to lmdb

(so that these won't affect latencies / rps on proxy workers)

### DIFF
--- a/kong/clustering/data_plane.lua
+++ b/kong/clustering/data_plane.lua
@@ -60,7 +60,7 @@ function _M:init_worker(plugins_list)
 
   self.plugins_list = plugins_list
 
-  if ngx.worker.id() == 0 then
+  if clustering_utils.is_dp_worker_process() then
     assert(ngx.timer.at(0, function(premature)
       self:communicate(premature)
     end))
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -86,6 +86,7 @@ local migrations_utils = require "kong.cmd.utils.migrations"
 local plugin_servers = require "kong.runloop.plugin_servers"
 local lmdb_txn = require "resty.lmdb.transaction"
 local instrumentation = require "kong.tracing.instrumentation"
+local process = require "ngx.process"
 local tablepool = require "tablepool"
 local get_ctx_table = require("resty.core.ctx").get_ctx_table
 
@@ -108,6 +109,7 @@ local ngx_DEBUG        = ngx.DEBUG
 local is_http_module   = ngx.config.subsystem == "http"
 local is_stream_module = ngx.config.subsystem == "stream"
 local start_time       = ngx.req.start_time
+local worker_id        = ngx.worker.id
 local type             = type
 local error            = error
 local ipairs           = ipairs
@@ -384,7 +386,7 @@ local function execute_cache_warmup(kong_config)
     return true
   end
 
-  if ngx.worker.id() == 0 then
+  if worker_id() == 0 then
     local ok, err = cache_warmup.execute(kong_config.db_cache_warmup_entities)
     if not ok then
       return nil, err
@@ -630,6 +632,13 @@ function Kong.init()
   db:close()
 
   require("resty.kong.var").patch_metatable()
+
+  if config.role == "data_plane" then
+    local ok, err = process.enable_privileged_agent(2048)
+    if not ok then
+      error(err)
+    end
+  end
 end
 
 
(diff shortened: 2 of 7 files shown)


kong/clustering/data_plane.lua
```diff
@@ -60,7 +61,7 @@ function _M:init_worker(plugins_list)
 
   self.plugins_list = plugins_list
 
-  if ngx.worker.id() == 0 then
+  if process.type() == "privileged agent" then
```
chronolaw, line 64: Could we define a help function like `is_dp_worker_process` in `clusteringutils.lua`,?
then we will remove this duplicated code `process.type() == "privileged agent"`.

kong/init.lua
```diff
@@ -667,78 +676,80 @@ function Kong.init_worker()
...
+        declarative.get_current_hash() ~= nil then
+        -- if there is no declarative config set and a config is present in LMDB,
+        -- just build the router and plugins iterator
+        ngx_log(ngx_INFO, "found persisted lmdb config, loading...")
+        local ok, err = declarative_init_build()
+        if not ok then
+          stash_init_worker_error("failed to initialize declarative config: " .. err)
+          return
+        end
+      elseif declarative_entities then
```
chronolaw, line 702: Add a blank line for style reason.

kong/init.lua
```diff
@@ -667,78 +676,80 @@ function Kong.init_worker()
...
-      return
-    end
-
-    if not has_declarative_config(kong.configuration) and
-      declarative.get_current_hash() ~= nil then
-      -- if there is no declarative config set and a config is present in LMDB,
-      -- just build the router and plugins iterator
-      ngx_log(ngx_INFO, "found persisted lmdb config, loading...")
-      local ok, err = declarative_init_build()
+  if process.type ~= "privileged agent" then
```
chronolaw, line 679: It is a huge `if-then-end` block, could we use a simple `goto` to keep the code clean?

dndx:
Can we measure the performance gain of this change @ADD-SP ?

kong/init.lua
```diff
@@ -755,7 +772,7 @@ end
 
 
 function Kong.exit_worker()
-  if kong.configuration.role ~= "control_plane" and ngx.worker.id() == 0 then
+  if process.type ~= "privileged agent" and kong.configuration.role ~= "control_plane" and worker_id() == 0 then
```
chronolaw, line 775: should use `not is_dp_worker_process()`

bungle: @chronolaw I tried it, requiring "kong.clustering.utils"  causes errors on `init.lua`.

So either we need to move it else where or just keep it like it is.

kong/init.lua
```diff
@@ -674,6 +683,13 @@ function Kong.init_worker()
 
   kong.db:set_events_handler(worker_events)
 
+  if process.type == "privileged agent" then
```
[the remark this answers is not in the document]
bungle: @chronolaw I tried it, requiring "kong.clustering.utils"  causes errors on `init.lua`.

So either we need to move it else where or just keep it like it is.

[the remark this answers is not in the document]
chronolaw: OK, thanks.

kong/init.lua
chronolaw: OK, thanks.

kong/init.lua
chronolaw, line 686: should be `process.type()`

kong/init.lua
chronolaw, line 775: should be `process.type()`

kong/init.lua
```diff
@@ -795,7 +812,7 @@ end
 
 
 function Kong.exit_worker()
-  if kong.configuration.role ~= "control_plane" and ngx.worker.id() == 0 then
+  if process.type() ~= "privileged agent" and kong.configuration.role ~= "control_plane" and worker_id() == 0 then
```
StarlightIbuki, line 815: Is there any reason we don't use privileged workers to manage plugin servers?

bungle: @StarlightIbuki I wanted to skip it for now as I don't understand all the consequences of it. E.g. is it ok to have plugin servers managed with master process rights. I suggest we do this in steps. We haven't used privileged workers before, so making too many changes at once feels too risky. We can come back to that later.

chronolaw (approved):

flrgh (approved):
Great work! I think the privileged worker process is something we should utilize more and more in the future for these types of tasks.

StarlightIbuki (approved):

bungle:
> Great work! I think the privileged worker process is something we should utilize more and more in the future for these types of tasks.

Yes, next obvious target is Kong EE Vitals.

Issue #9418, opened 2022-09-10
Kong Req Transformation Bug Authorization header rename functionality


jeremyjpj0916:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

2.8.1

### Current Behavior

When I use the request tranformer config.header.rename I can see its behavior flakes between including the bearer token from OAuth2 token vs transforming/renaming the specified header to take the Authorization Headers place:

Ex:
Not working ex(left kongs oauth2 token in the Authorization header):
<img width="1128" alt="image" src="https://user-images.githubusercontent.com/31913027/189502472-0262758d-02e4-4628-8fbd-3823bf8b1df7.png">

When it works: (Note the Authorization header being reported with the Bearer XXXXXXXXXXXX replacement value.

<img width="1128" alt="image" src="https://user-images.githubusercontent.com/31913027/189502506-29ae2d64-19bb-403f-a217-7b38982015ff.png">

This is being produced by having a backend "echo" service spit back seen http headers + body for me. 

### Expected Behavior

It should always replace the Authorization headers value to have: Bearer XXXXXXXXXXXX in it.

### Steps To Reproduce

```markdown
1. Make a Kong route + service proxy running OAuth2.0 CC + ACL + Req Transformation plugin on it.

2. Call the proxy with proper Authorization bearer header to succeed with OAuth2.0 Kong auth AND pass back a X-Venafi-Auth header with any kinda value in it.

3. Watch and see Kong be inconsistent with what the upstream receives with the transformation plugin in place with respect to the Authorization Header the upstream API receives.

Tranformer config ex:

{
  "service": null,
  "config": {
    "append": {
      "headers": [],
      "body": [],
      "querystring": []
    },
    "add": {
      "headers": [],
      "body": [],
      "querystring": []
    },
    "replace": {
      "uri": null,
      "headers": [],
      "body": [],
      "querystring": []
    },
    "rename": {
      "headers": [
        "X-Venafi-Auth:Authorization"
      ],
      "body": [],
      "querystring": []
    },
    "remove": {
      "headers": [],
      "body": [],
      "querystring": []
    },
    "http_method": null
  },
  "protocols": [
    "grpc",
    "grpcs",
    "http",
    "https"
  ],
  "enabled": true,
  "id": "edab2d9e-25aa-44fc-a49c-457322e35e94",
  "name": "request-transformer",
  "consumer": null,
  "created_at": 1662845032,
  "route": {
[...]

PR #9442, opened 2022-09-17
fix(request-transformer): unpredictable overriding result from rename header to existing header caused by pairs disorder

windmgc:
### Summary

This PR fixes a bug in the `request-transformer` plugin, where the rename header cannot override an already existing header steadily.

If a header `A-header` already exists with a value `A-value` in the request, and the `request-transformer` plugin is configured to rename `B-header`(with a value `B-value`) to `A-header`(with different format, like all chars are capitalized), then the upstream request will contain an `A-header` with possibly `A-value` or `B-value`. This unpredictable behaviour may caused by `pairs` function disorder behaviour(Again!).

The code flow is as follows:

A `header` table is fetched by `ngx.req.get_headers` and used later for storing upstream request headers
https://github.com/Kong/kong/blob/c33d745793b454ab27d9681562fadbccbc454a1b/kong/plugins/request-transformer/access.lua#L186

Then the rename header list will be iterated
https://github.com/Kong/kong/blob/c33d745793b454ab27d9681562fadbccbc454a1b/kong/plugins/request-transformer/access.lua#L200-L209

Note that in this code block, the `new_name` will be directly used for setting value in the table, and the headers fetched by `ngx.req.get_headers()` are in lower case, so if `new_name` is an upper-case, then two keys with different format will exist in the `header` table.

And that will lead to the unpredictable iterating sequence in setting headers:
https://github.com/Kong/kong/blob/c33d745793b454ab27d9681562fadbccbc454a1b/kong/plugins/request-transformer/access.lua#L247

https://github.com/Kong/kong/blob/c33d745793b454ab27d9681562fadbccbc454a1b/kong/pdk/service/request.lua#L422-L427

### Full changelog

* Fix unpredictable overriding result from rename header to existing header in request-transformer plugin, by normalizing the header name(set to lower case)
* Cleanup duplicate tests and add related test

### Issue reference

Fix [FTI-4319] and #9418 

[FTI-4319]: https://konghq.atlassian.net/browse/FTI-4319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,9 @@
 - **AWS Lambda**: Fix an issue that is causing inability to
   read environment variables in ECS environment.
   [#9460](https://github.com/Kong/kong/pull/9460)
+- **Request-Transformer**: fix a bug when header renaming will override
+  existing header and cause unpredictable result.
+  [#9442](https://github.com/Kong/kong/pull/9442)
 
 ### Dependencies
 
--- a/kong/plugins/request-transformer/access.lua
+++ b/kong/plugins/request-transformer/access.lua
@@ -200,11 +200,9 @@ local function transform_headers(conf)
   -- Rename headers(s)
   for _, old_name, new_name in iter(conf.rename.headers) do
     old_name = old_name:lower()
-    -- Normalize new header name so that it can be successfully overridden
-    new_name = new_name:lower()
-    if headers[old_name] then
-      local value = headers[old_name]
-      headers[new_name] = value
+    local value = headers[old_name]
+    if value then
+      headers[new_name:lower()] = value
       headers[old_name] = nil
       headers_to_remove[old_name] = true
     end
--- a/kong/plugins/request-transformer/access.lua
+++ b/kong/plugins/request-transformer/access.lua
@@ -200,6 +200,8 @@ local function transform_headers(conf)
   -- Rename headers(s)
   for _, old_name, new_name in iter(conf.rename.headers) do
     old_name = old_name:lower()
+    -- Normalize new header name so that it can be successfully overridden
+    new_name = new_name:lower()
     if headers[old_name] then
       local value = headers[old_name]
       headers[new_name] = value
--- a/spec/03-plugins/36-request-transformer/02-access_spec.lua
+++ b/spec/03-plugins/36-request-transformer/02-access_spec.lua
@@ -739,7 +739,7 @@ describe("Plugin: request-transformer(access) [#" .. strategy .. "]", function()
           headers = {
             host = "test9.test",
             ["x-to-rename"] = "new-result",
-            ["x-is-renamed"] = "old-result"
+            ["x-is-renamed"] = "old-result",
           }
         })
         assert.response(r).has.status(200)
--- a/spec/03-plugins/36-request-transformer/02-access_spec.lua
+++ b/spec/03-plugins/36-request-transformer/02-access_spec.lua
@@ -3,6 +3,8 @@ local helpers = require "spec.helpers"
 local cjson   = require "cjson"
 local pl_file = require "pl.file"
 
+local fmt = string.format
+
 
 local function count_log_lines(pattern)
   local cfg = helpers.test_conf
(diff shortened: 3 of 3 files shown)


spec/03-plugins/36-request-transformer/02-access_spec.lua
```diff
@@ -859,154 +879,6 @@ describe("Plugin: request-transformer(access) [#" .. strategy .. "]", function()
     end)
   end)
 
-  describe("rename", function()
```
windmgc, line 862: line 862-1009 seems to be duplicated with line 699-844, so remove them 

https://github.com/Kong/kong/blob/c33d745793b454ab27d9681562fadbccbc454a1b/spec/03-plugins/36-request-transformer/02-access_spec.lua#L862-L1009

https://github.com/Kong/kong/blob/c33d745793b454ab27d9681562fadbccbc454a1b/spec/03-plugins/36-request-transformer/02-access_spec.lua#L699-L844

spec/03-plugins/36-request-transformer/02-access_spec.lua
```diff
@@ -729,6 +731,24 @@ describe("Plugin: request-transformer(access) [#" .. strategy .. "]", function()
...
     end)
+    for _, seq in ipairs({ 1, 2, 3, 4, 5, 6}) do
+      it(fmt("override value if target header already exist: #%s", seq), function ()
+        local r = assert(client:send {
+          method = "GET",
+          path = "/request",
+          headers = {
+            host = "test9.test",
+            ["x-to-rename"] = "new-result",
+            ["x-is-renamed"] = "old-result"
```
chronolaw, line 742: add a `,` at the end of line.

windmgc: fixed

hanshuebner:
To prevent such issues from happening, would it not make sense to have all headers tables have this metatable:
```
 {
    __newindex=function (t,k,v) rawset(t,k:lower(),v) end,
    __index=function(t,k) return t[k:lower()] end,
}
```
That way, the normalization to lower case would happen automatically.

windmgc:
> To prevent such issues from happening, would it not make sense to have all headers tables have this metatable:
> 
> ```
> {
>   __newindex=function (t,k,v) rawset(t,k:lower(),v) end,
>   __index=function(t,k) return t[k:lower()] end,
> }
> ```
> 
> That way, the normalization to lower case would happen automatically.

It's a good idea but needs changes in the PDK(more like a patch on the original OpenResty's `ngx.req.get_headers`, I think they already have `__index` to do the lowercase work but I don't know why they did not have the `__newindex` behaviour)

I feel it is better to open another PR for this change, and doing another lower case here is not very harmful, WDYT?

kong/plugins/request-transformer/access.lua
```diff
@@ -200,6 +200,8 @@ local function transform_headers(conf)
   -- Rename headers(s)
   for _, old_name, new_name in iter(conf.rename.headers) do
     old_name = old_name:lower()
+    -- Normalize new header name so that it can be successfully overridden
+    new_name = new_name:lower()
     if headers[old_name] then
       local value = headers[old_name]
       headers[new_name] = value
```
chronolaw, line 207: I think we could change this line, may be a more clear solution:

`headers[new_name:lower()] = value`

windmgc: Okay changed.

chronolaw (approved):

chronolaw:
Need a change log entry.

CHANGELOG.md
```diff
@@ -78,6 +78,11 @@
   correctly created.
   [#9434](https://github.com/Kong/kong/pull/9434)
 
+#### Plugins
+
+- **Request-Transformer**: fix a bug when header renaming will override
+  existing header and cause unpredictable result.
```
chronolaw, line 84: Need a PR link, see other change log entry.

windmgc: fixed, add correct changelog entry

CHANGELOG.md
```diff
@@ -80,6 +80,11 @@
 - Data plane's connection to control plane is moved to a privileged worker process
   [#9432](https://github.com/Kong/kong/pull/9432)
 
+#### Plugins
+
+- **AWS Lambda**: add `requestContext` field into `awsgateway_compatible` input data
```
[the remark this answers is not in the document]
windmgc: I've dropped the merge-master commit and updated by rebase cc @fffonion

spec/03-plugins/36-request-transformer/02-access_spec.lua
```diff
@@ -729,6 +731,24 @@ describe("Plugin: request-transformer(access) [#" .. strategy .. "]", function()
...
     end)
+    for _, seq in ipairs({ 1, 2, 3, 4, 5, 6}) do
+      it(fmt("override value if target header already exist: #%s", seq), function ()
+        local r = assert(client:send {
+          method = "GET",
+          path = "/request",
+          headers = {
+            host = "test9.test",
+            ["x-to-rename"] = "new-result",
+            ["x-is-renamed"] = "old-result",
```
fffonion, line 742: could you add a (regression) test that covers the bug you mentioned in this PR? for example a plugin configure to rename `x-to-rename` to `X-Is-Renamed`?

windmgc: Test added.

fffonion (approved):

PR #9444, opened 2022-09-19
feat(jwt) Add configuration for JWT token claims error response

sakkeerhussain:
### Summary

This will create a new configuration('error_response_claims') to override the JWT signature expired error

### DIFF
--- a/kong/plugins/jwt/handler.lua
+++ b/kong/plugins/jwt/handler.lua
@@ -187,7 +187,7 @@ local function do_authentication(conf)
   -- Verify the JWT registered claims
   local ok_claims, errors = jwt:verify_registered_claims(conf.claims_to_verify)
   if not ok_claims then
-    return false, { status = 401, errors = errors }
+    return false, { status = 401, errors = conf.error_response_claims or errors }
   end
 
   -- Verify the JWT registered claims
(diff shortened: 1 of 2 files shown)


StarlightIbuki:
Hi. Thank you for your contribution. However, we're discussing how we should custom responses in a more unified way. (See also rate-limiting(-advanced), response-rate-limiting and etc) Thus we are holding this PR back for now.

sakkeerhussain:
Understood.

It would be great to have a common error response, but having the option to configure an error message with more priority inside the JWT plugin config would be better to have an error message explicitly giving the token expired.

We were trying to provide an error message like this and failed with the current version:
{
  "code": "CONFIG_TOKEN_EXPIRED",
  "message": "The config token provided expired"
}

StarlightIbuki:
> Understood.
> 
> It would be great to have a common error response, but having the option to configure an error message with more priority inside the JWT plugin config would be better to have an error message explicitly giving the token expired.
> 
> We were trying to provide an error message like this and failed with the current version: { "code": "CONFIG_TOKEN_EXPIRED", "message": "The config token provided expired" }

By "unified way" I mean we are considering a systematic way to support this (user-defined error message) for all plugins that may return an error response.
A workaround could be to create a custom plugin to modify the response. Sorry again for the inconvenience.

sakkeerhussain:
Understood the concept and looking forward for the plugin to override the error response.

For our use case the one error senario we are encountering from the jwt plugin is the expired token and that's we were trying to override with that response message. Now the error response from the jwt plugin is not matching to our common error response structure. But we should be able to archive this with the custome response override plugin you mentioned above.

PR #9448, opened 2022-09-20
chore(error_handlers): define response message for trace method 405 error status code

liverpool8056:
### Summary

When nginx blocks request with trace method from client, it hits the default error message in kong which is "The upstream server responded with 405", it is a little ambiguous. Actually the request doesn't reach upstream. 
Define  a detail message "Method not allowed" for 405 status code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,16 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+#### Core
+
+- Change the reponse body for a TRACE method from `The upstream server responded with 405`
+  to `Method not allowed`, make the reponse to show more clearly that Kong do not support 
+  TRACE method.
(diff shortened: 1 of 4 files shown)


dndx:
@liverpool8056 Could you add a test case for this PR?

liverpool8056:
@dndx So sorrrrrrry for missing test case, and thank you for helping me to correct it, I added the test case in the new commit.

chronolaw:
Should we add an entry in changelog?

chronolaw:
Due to PR #9513, we should rebase the PR and change the index from string to number.

liverpool8056:
@chronolaw already done

chronolaw (approved):

hbagdi:
pinging @fffonion for merge.

PR #9457, opened 2022-09-22
fix(opentelemetry): set default propagation type to w3c

mayocream:
According to the original design, the default propagation header type should be `w3c`, described in https://docs.konghq.com/hub/kong-inc/opentelemetry/#propagation.

> The plugin detects the propagation format from the headers and will use the appropriate format to propagate the span context. If no appropriate format is found, the plugin will fail back to the default format, which is w3c.

The previous implementation didn't specify the default header type successfully, it was failback to the `b3` default.

### DIFF
--- a/kong/plugins/opentelemetry/handler.lua
+++ b/kong/plugins/opentelemetry/handler.lua
@@ -178,7 +178,7 @@ function OpenTelemetryHandler:rewrite()
     root_span.parent_id = span_id
   end
 
-  propagation_set("w3c", header_type, root_span)
+  propagation_set("w3c", header_type, root_span, "w3c")
 end
 
 function OpenTelemetryHandler:log(conf)
--- a/kong/plugins/opentelemetry/handler.lua
+++ b/kong/plugins/opentelemetry/handler.lua
@@ -178,7 +178,7 @@ function OpenTelemetryHandler:rewrite()
     root_span.parent_id = span_id
   end
 
-  propagation_set("w3c", header_type, root_span, "w3c")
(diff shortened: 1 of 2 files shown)


fffonion:
discussed with @mayocream and moving this out of 3.0.1 scope

fffonion (approved):

dhanutalari:
i am using kong 3.4.2 and still i am getting 
propagation.lua:540 [opentelemetry] Mismatched header types. conf: w3c. found: b3 
propagation.lua:540 [opentelemetry] Mismatched header types. conf: w3c. found: was 

@mayocream cloud you help me to resolve this ??

samugi:
hi @dhanutalari, are the requests that generated those logs passing any `b3` and/or `AWS` formatted tracing headers?

If that is the case, in `3.4.x`, that was an expected behavior, as mentioned [in the docs](https://docs.konghq.com/hub/kong-inc/opentelemetry/#propagation), the "default format", that this PR modified, is only effective when no incoming tracing header is present in the requests, otherwise the format used depends on what is detected in the incoming request, based on an hardcoded order of precedence also visible at the same docs link.

This obviously did not offer a lot of flexibility, the propagation module has recently been rewritten to enable better configurability and complete freedom in deciding how headers propagation behaves (i.e. how Kong extracts/injects tracing headers). The new propagation module is unreleased at the moment but will soon be available in  `3.7`. You can find documentation about how the new propagation module works in our [unreleased docs for the OTel plugin](https://docs.konghq.com/hub/kong-inc/opentelemetry/unreleased/#propagation).

PR #9463, opened 2022-09-23
fix(response-transformer): fix the error when response-transformer

liverpool8056:
### Summary

Fix the error that the response from upstream is a plain string with Header Content-Type: application/json, the response transformer throw exception when transforming.


chronolaw:
It seems to be duplicated with PR #9491, could you take a took at it?

kong/plugins/response-transformer/body_transformer.lua
```diff
@@ -90,9 +93,9 @@ end
 
 
 function _M.transform_json_body(conf, buffered_data)
-  local json_body = read_json_body(buffered_data)
+  local json_body = parse_json(buffered_data)
   if json_body == nil then
-    return
+    return nil
```
samugi, line 98:
```suggestion
    return nil, "failed parsing json body"
```
I think it would be useful to return a warning, then log it  in case of parsing failure

samugi:
Thank you @chronolaw for spotting the duplicate. I think adding the warning would be useful, then we can close mine.

spec/03-plugins/15-response-transformer/02-body_transformer_spec.lua
```diff
@@ -135,6 +135,11 @@ describe("Plugin: response-transformer", function()
         assert.same({a = {}}, body_json)
         assert.equals('[]', cjson.encode(body_json.a))
       end)
+      it("plain text", function()
```
chronolaw, line 138: add a blank line before each `it(...)` case.

spec/03-plugins/15-response-transformer/02-body_transformer_spec.lua
```diff
@@ -135,6 +135,11 @@ describe("Plugin: response-transformer", function()
...
+        local body = "test"
+        local json_body = body_transformer.transform_json_body(conf, body)
+        assert(json_body==nil)
```
chronolaw, line 141: style: `json_body == nil`

liverpool8056: @samugi Yes, a warning would be useful. Thanks!
@chronolaw Style has been corrected. Thanks!

kong/plugins/response-transformer/body_transformer.lua
```diff
@@ -39,9 +40,12 @@ local function cast_value(value, value_type)
 end
 
 
-local function read_json_body(body)
+local function parse_json(body)
   if body then
-    return cjson.decode(body)
+    local ok, res = pcall(cjson_decode, body)
```
chronolaw, line 45: localize `pcall`

spec/03-plugins/15-response-transformer/02-body_transformer_spec.lua
```diff
@@ -135,6 +135,12 @@ describe("Plugin: response-transformer", function()
         assert.same({a = {}}, body_json)
         assert.equals('[]', cjson.encode(body_json.a))
       end)
+
+      it("plain text", function()
+        local body = "test"
+        local json_body = body_transformer.transform_json_body(conf, body)
+        assert(json_body == nil)
```
samugi, line 142:
```suggestion
        local json_body, err = body_transformer.transform_json_body(conf, body)
        assert.is_nil(json_body)
        assert.equals(err, "failed parsing json body")
```
the body was returned `nil` even before this fix, so this is not testing the change right now. I would suggest checking the expected error message here.

samugi:
I think it would be a good idea to cover the behaviour of `body_filter` too in the tests, when the "json" body is invalid, something similar to https://github.com/Kong/kong/pull/9491/files#diff-6f3111eafa8c99e31fe6715e0bbd608a9e5685ff7595ea3c4eafce0f527955c5R327-R389

samugi:
> I think it would be a good idea to cover the behaviour of `body_filter` too in the tests, when the "json" body is invalid, something similar to https://github.com/Kong/kong/pull/9491/files#diff-6f3111eafa8c99e31fe6715e0bbd608a9e5685ff7595ea3c4eafce0f527955c5R327-R389

@liverpool8056 could you add this too? ^

liverpool8056:
@samugi It seems that I've already added a similar test.
https://github.com/Kong/kong/pull/9463/files#diff-6f3111eafa8c99e31fe6715e0bbd608a9e5685ff7595ea3c4eafce0f527955c5

samugi:
> @samugi It seems that I've already added a similar test. https://github.com/Kong/kong/pull/9463/files#diff-6f3111eafa8c99e31fe6715e0bbd608a9e5685ff7595ea3c4eafce0f527955c5

The current unit test covers the `transform_json_body()` function, which is good, but it is not testing the correct behaviour of `handler:body_filter()` which was the source of this issue.

liverpool8056:
@samugi understood, thanks!

samugi (approved):

PR #9466, opened 2022-09-23
Refactor datadog hostname/port env vars

### LINKED ISSUE #7954: Unable to set datadog hostname or port with environment variables

chazdnato:
### Summary

This addresses an issue reported here:
https://github.com/Kong/kong/issues/7954

While the tests set a null value to instead inherit the environment variables, this does not work in a Kubernetes context. For example, if I have this config:

```
config:
  host: null
  port: 8125
```

The hostname still resolves to `localhost` once it gets through the schema.lua defaults.

This approach reverses the order of ENV var evaluation, so that environment variables are used if set, and config if not.

### Full changelog

* reverses order of logic determining if env vars or config is used for host/port
* adds test to ensure config is used if env var is not set

### Issue reference

Fix #7954

### DIFF
--- a/kong/plugins/datadog/statsd_logger.lua
+++ b/kong/plugins/datadog/statsd_logger.lua
@@ -21,12 +21,12 @@ local statsd_mt = {}
 statsd_mt.__index = statsd_mt
 
 local env_datadog_agent_host = os.getenv 'KONG_DATADOG_AGENT_HOST'
-local env_datadog_agent_port = tonumber(os.getenv 'KONG_DATADOG_AGENT_PORT' or "")
+local env_datadog_agent_port = os.getenv 'KONG_DATADOG_AGENT_PORT'
 
 function statsd_mt:new(conf)
   local sock   = udp()
-  local host = conf.host or env_datadog_agent_host
-  local port = conf.port or env_datadog_agent_port
+  local host = env_datadog_agent_host or conf.host
+  local port = tonumber(env_datadog_agent_port or conf.port)
 
   local _, err = sock:setpeername(host, port)
   if err then
--- a/spec/03-plugins/08-datadog/01-log_spec.lua
+++ b/spec/03-plugins/08-datadog/01-log_spec.lua
@@ -64,6 +64,11 @@ for _, strategy in helpers.each_strategy() do
         service = bp.services:insert { name = "dd6" }
       }
 
+      local route7 = bp.routes:insert {
+        hosts   = { "datadog7.com" },
+        service = bp.services:insert { name = "dd7" }
+      }
+
       bp.plugins:insert {
         name     = "key-auth",
         route = { id = route1.id },
@@ -149,8 +154,8 @@ for _, strategy in helpers.each_strategy() do
         name     = "datadog",
         route = { id = route5.id },
         config   = {
-          host = ngx.null, -- plugin takes above env var value, if set to null
-          port = ngx.null, -- plugin takes above env var value, if set to null
+          host = "no-such-config.com", -- plugin takes above env var value, if env var is set
+          port = 8125, -- plugin takes above env var value, if env var is set
         },
       }
 
@@ -182,6 +187,21 @@ for _, strategy in helpers.each_strategy() do
           service_name_tag = "upstream",
           status_tag       = "http_status",
           consumer_tag     = "user",
+        },
+      }
+
+      bp.plugins:insert {
+        name     = "key-auth",
+        route = { id = route7.id },
+      }
+
+      helpers.setenv('KONG_DATADOG_AGENT_HOST', 'localhost')
+      bp.plugins:insert {
+        name     = "datadog",
+        route = { id = route7.id },
+        config   = {
+          host = "no-such-config.com", -- plugin takes above env var value, if env var is set
+          port = 9999, -- plugin takes this value, as the env var for port is not set
         },
       }
 
@@ -355,6 +375,30 @@ for _, strategy in helpers.each_strategy() do
       assert.contains("kong.upstream_latency:%d+|ms|#name:dd5,status:200,consumer:bar,app:kong", gauges, true)
       assert.contains("kong.kong_latency:%d*|ms|#name:dd5,status:200,consumer:bar,app:kong", gauges, true)
     end)
+
+    it("logs metrics to host when it is the only thing defined via environment variables", function()
+      local thread = helpers.udp_server(9999, 6)
+
+      local res = assert(proxy_client:send {
+        method  = "GET",
+        path    = "/status/200?apikey=kong",
+        headers = {
+          ["Host"] = "datadog7.com"
+        }
+      })
+      assert.res_status(200, res)
+
+      local ok, gauges = thread:join()
+      assert.True(ok)
+      assert.equal(6, #gauges)
+      assert.contains("kong.request.count:1|c|#name:dd7,status:200,consumer:bar,app:kong" , gauges)
+      assert.contains("kong.latency:%d+|ms|#name:dd7,status:200,consumer:bar,app:kong", gauges, true)
+      assert.contains("kong.request.size:%d+|ms|#name:dd7,status:200,consumer:bar,app:kong", gauges, true)
+      assert.contains("kong.response.size:%d+|ms|#name:dd7,status:200,consumer:bar,app:kong", gauges, true)
+      assert.contains("kong.upstream_latency:%d+|ms|#name:dd7,status:200,consumer:bar,app:kong", gauges, true)
+      assert.contains("kong.kong_latency:%d*|ms|#name:dd7,status:200,consumer:bar,app:kong", gauges, true)
+    end)
+
 
     it("should not return a runtime error (regression)", function()
       local thread = helpers.udp_server(9999, 1, 1)


Tieske:
This changes behaviour, so is a breaking change.

chazdnato:
> the proper way to specify a `null` value in yaml is to leave the line blank, like this:
> 
> ```
> config:
>   host: 
>   port: 8125
> ```

Even with that setup, the plugin renders it as localhost. Here are my configs (this is using the helm charts).

Helm chart:
```
kind: KongClusterPlugin
metadata:
  name: global-datadog
  annotations:
    kubernetes.io/ingress.class: kong
  labels:
    global: "true"
config:
  host:
  port: 8125
  metrics:
  prefix: api-gateway
  service_name_tag: kong_upstream
plugin: datadog
```

Datadog config as rendered by k8s
```
❯ kc get kongclusterplugin global-datadog -o yaml
config:
  host: null
  port: 8125
  prefix: api-gateway
  service_name_tag: kong_upstream
kind: KongClusterPlugin
plugin: datadog
```

And thus the Datadog config as interpreted by kong
```
{
  "consumer": null,
  "protocols": [
    "http",
    "https"
  ],
  "enabled": true,
  "route": null,
  "config": {
    "consumer_tag": "consumer",
    "status_tag": "status",
    "host": "localhost",
    "prefix": "api-gateway",
    "service_name_tag": "kong_upstream",
    "port": 8125
  },
  "name": "datadog",
  "id": "0524254d-9a97-5803-b2b3-dfd1233ec721",
  "created_at": 1664195179,
  "service": null,
  "tags": null
}
```

My solution as prescribed should leave those using ENV vars in non-k8s context still working as expected, and those using configs still working as expected. The only anticipated change in behaviour could be if someone has both env vars and config setup, and they expect their config to be used.

I'm open to other suggested approaches.

Tieske:
> Even with that setup, the plugin renders it as localhost. Here are my configs (this is using the helm charts). Helm will interpret any null as "don't put that key in place":

If that is the case then this is a major issue. Why doesn't helm/k8s stick to the JSON semantics???? There are many places where the Kong admin API relies on a JSON-null value to be specified by the user to clear a configuration item, or a default value.

chazdnato:
> > Even with that setup, the plugin renders it as localhost. Here are my configs (this is using the helm charts). Helm will interpret any null as "don't put that key in place":
> 
> If that is the case then this is a major issue. Why doesn't helm/k8s stick to the JSON semantics???? There are many places where the Kong admin API relies on a JSON-null value to be specified by the user to clear a configuration item, or a default value.

I would presume because it's YAML underneath, and not JSON. Be that as it may, the issue we're trying to fix (issue number 7954)[https://github.com/Kong/kong/issues/7954] has some k8s specific work arounds, that aren't viable if one is unable to use k8s 1.22+ for whatever reason.

Tieske:
> I would presume because it's YAML underneath, and not JSON.

Nope, the example before is the YAML way of expressing a JSON-null;

```yaml
config:
  host: 
  port: 8125
```

chazdnato:
Would you be able to detail the scenarios under which this may be a breaking change? Do you have alternative solutions to the approach suggested?

chazdnato:
> > I would presume because it's YAML underneath, and not JSON.
> 
> Nope, the example before is the YAML way of expressing a JSON-null;
> 
> ```yaml
> config:
>   host: 
>   port: 8125
> ```

Both are valid: https://yaml.org/spec/1.2.2/#10211-null

However, this is not germane, and unproductive.

Tieske:
> Would you be able to detail the scenarios under which this may be a breaking change? Do you have alternative solutions to the approach suggested?

if both plugin config and envvars are set, it previously used plugin config, and will now pick up envvars. That's a breaking change.

chazdnato:
> > Would you be able to detail the scenarios under which this may be a breaking change? Do you have alternative solutions to the approach suggested?
> 
> if both plugin config and envvars are set, it previously used plugin config, and will now pick up envvars. That's a breaking change.

Would you accept a configuration parameter that would default the current logic, but if set apply the logic I've implemented?

chazdnato:
It appears this approach has been suggested in the past: https://github.com/Kong/kong/pull/8106#discussion_r762184540

Have replied in that thread, with hope we can get a good direction going with this.

kikito:
In my opinion there is two different problems at play here:

1. Helm/K8s (it is unclear which one is at doing this) treat "a field that isn't set" as equivalent to "a field that is set to null", while Kong treats those differently in some cases (`null` means "don't use the default" in Kong)
2. Kong schemas have no functionality for reading environment variables. Every time this is needed this is done after the schema has already finished doing its work. This issue (and #8106) result from the same missing functionality.

My personal opinion is that for **1**, Helm/K8s is interpreting the YAML spec wrong. Admittedly, the spec [mentions](https://yaml.org/spec/1.2.2/#72-empty-nodes) that:

> Nodes with empty are interpreted as if they were plain scalars with an empty value. Such nodes are commonly resolved to a “null” value.

But **commonly resolved** is not the same as **should always be resolved**, which is what seems to be happening. In fact, on the [description about null](https://yaml.org/spec/1.2.2/#10211-null), this is addressed specifically.

> Represents the lack of a value. This is typically bound to a native null-like value (e.g., undef in Perl, None in Python). Note that a null is different from an empty string. **Also, a mapping entry with some key and a null value is valid and different from not having that key in the mapping**.

Emphasis mine.

I think Helm/K8s is at fault; they should not be adding nulls to people's yaml files.
 
As for **2**: Ideally we should have a way to handle environment variables inside schemas. At least we should have the option of making the `default` a piece of Lua code that can go and grab that:

```
{ host = typedefs.host({ default = function() return os.getenv("DATADOG_HOST") or "localhost" end }), }
```

Or even add a new exclusive schema field, just for this purpose:

```
{ host = typedefs.host({ default_env_var = "DATADOG_HOST", default = "localhost" }), }
```

I don't know which option is better yet, but it is clear to me that this shouldn't be resolved by individual plugins every time they  to read env variables for default values. The change needs to happen on the schema level, and then all plugins will be able to use it.

chazdnato:
> 1. Helm/K8s (it is unclear which one is at doing this) treat "a field that isn't set" as equivalent to "a field that is set to null", while Kong treats those differently in some cases (`null` means "don't use the default" in Kong)
>
> I think Helm/K8s is at fault; they should not be adding nulls to people's yaml files.

I did some digging, and I think it might be the kubernetes-ingress-controller (on top of what k8s does). I can use helm/k8s to create an object that, when empty in the config, ends up with a `null` value in k8s [agreed this isn't optimal, but it's better than removing the key!]:

```
> kubectl get kongclusterplugin global-datadog -o yaml
config:
  host: null
    - app:kong
  port: 8125```

This is the same if I look at the object as JSON. Note that if I edit the value to be non-null, then re-edit to be blank, the key is removed entirely. This sounds like a k8s-ism, though helm may be converting empty keys to `null` at first.

This said, when pulling this data to post to kong, the kubernetes-ingress-controller will `omitempty`:

https://github.com/Kong/kubernetes-ingress-controller/blob/963623d02453d49f284bcef4dfd078cf76b684d5/pkg/apis/configuration/v1/kongclusterplugin_types.go#L51

This means that, even if k8s did maintain an empty value (or `null`), it would remove that key from the parsed config anyway - this is likely where the problem is actually happening.

While the kong world may say `null == not default`, this sadly doesn't translate to a k8s world.

**Comment by bungle** [comment]
Kong now has Secret Management built in. So there is another option:

1. mark a string field as `referenceable=true`
2. use reference like `{vault://env/datadog-host}`

**Comment by chazdnato** [comment]
> Kong now has Secret Management built in. So there is another option:
> 
> 1. mark a string field as `referenceable=true`
> 2. use reference like `{vault://env/datadog-host}`

Can you provide guidance (or documentation?) on how this would work in a k8s context? This might be a viable approach!

The workaround suggested [here](https://github.com/Kong/kong/issues/7954#issuecomment-1033221716) is also viable if one is able to upgrade to k8s 1.21+ - I think this is the approach we're going to have to take.

Is it worthwhile doing a documentation update on these behaviours with plugins that can have ENV vars, where there is an incompatibility with how k8s interprets blank/null values?

File: kong/plugins/datadog/statsd_logger.lua [hunk 1]
```diff
@@ -21,12 +21,12 @@ local statsd_mt = {}
...
 
 local env_datadog_agent_host = os.getenv 'KONG_DATADOG_AGENT_HOST'
-local env_datadog_agent_port = tonumber(os.getenv 'KONG_DATADOG_AGENT_PORT' or "")
+local env_datadog_agent_port = os.getenv 'KONG_DATADOG_AGENT_PORT'
 
 function statsd_mt:new(conf)
   local sock   = udp()
-  local host = conf.host or env_datadog_agent_host
-  local port = conf.port or env_datadog_agent_port
+  local host = env_datadog_agent_host or conf.host
```
StarlightIbuki, line 28: This makes the environment variable takes precedence over `kong.conf`, which has different behavior compared to what this originally intended.

Maybe we could remove the default value for `kong.conf`, and:

1. check if `conf.host` is set; use that if so;
2. check if `'KONG_DATADOG_AGENT_HOST'` env is set; use that if so;
3. use the default value

kong/plugins/datadog/statsd_logger.lua
```diff
@@ -21,12 +21,12 @@ local statsd_mt = {}
...
+  local port = tonumber(env_datadog_agent_port or conf.port)
```
StarlightIbuki, line 29:
```suggestion
  local port = tonumber(env_datadog_agent_port) or conf.port
```

spec/03-plugins/08-datadog/01-log_spec.lua
```diff
@@ -149,8 +154,8 @@ for _, strategy in helpers.each_strategy() do
         name     = "datadog",
         route = { id = route5.id },
         config   = {
-          host = ngx.null, -- plugin takes above env var value, if set to null
-          port = ngx.null, -- plugin takes above env var value, if set to null
+          host = "no-such-config.com", -- plugin takes above env var value, if env var is set
+          port = 8125, -- plugin takes above env var value, if env var is set
```
StarlightIbuki, line 158: This means env takes precedence over `kong.conf`.

StarlightIbuki (commented):
We should not change the precedence for `kong.conf` and the environment variable. I've suggested the other solution.

I assume your use case to be that leave host and port not set, and provide them via environment variable,

kong/plugins/datadog/statsd_logger.lua
StarlightIbuki: Check https://github.com/Kong/kong/blob/4ac5e9123e14c9cdf8f5d792f97561cc2afee40b/kong/plugins/datadog/schema.lua#L76

removing those default values will get us a `nil` when the option is omitted.

jaswanthikolla:
We currently have dead code in the plugin where `conf.host or env_datadog_agent_host`  the later part of the value is never evaluated. 

@StarlightIbuki 
It's fine to push back on this PR with different solution.  But, we should fix the apparent bug right?  Once we do that, alternative solution can be done.

fffonion:
@jaswanthikolla this is not dead code, `conf.KEY` will only be nil if user explictly set it using the method we describe above.
As of current state, i would suggest use secrets management (env var: https://docs.konghq.com/gateway/latest/kong-enterprise/secrets-management/getting-started/) as a workaround; please also open an issue at https://github.com/Kong/kubernetes-ingress-controller to actually fix the default value issue.

jaswanthikolla:
@fffonion  

I've tested it. `conf.host` will never be `nil` because we have default value with ` { host = typedefs.host({ default = "localhost" }), },` 

So, the later part of the `or` condition will never evaluate,  hence the dead code.

fffonion:
@jaswanthikolla It will, you can verify simply with `curl localhost:8001/plugins -d name=datadog -d config.host=`. Default values are only taken if user doesn't configure it; doesn't configure it vs leave it as empty is two different intents. The logic here is not a bug, the user experience is what needs improvement.
I understand the emotion that you might have when encounted an expected behaviour, but let's also interprete behaviour code with tests to avoid misleading reviewers or the original author.

jaswanthikolla:
>  Default values are only taken if user doesn't configure it; doesn't configure it vs leave it as empty is two different intents. 

Isn't that more confusing and against the mainstream? Let's go through it  ( actually it doesn't even work like that more on that later)
-    If I don't configure it, I will get default value
-    If I leave empty, I will get environment value. 
-   Isn't very easy for different systems to make mistake such as leaving empty/nil, they inject default value?  We're depending on this subtle different which is already broken, and explained in the following.

Bug:
>  you can verify simply with curl localhost:8001/plugins -d name=datadog -d config.host=. 

Sample Plugin config : 
```
 fields = {
          { host = typedefs.host({ default = "localhost" }), },
```
1.  I tried configuring with empty with following, and I get error `{plugins={[2]={config={host=\\\"length must be at least 1\\\"}}}}\")`
```
config:
  host: ""
```
2.  I configured with `null` and I get default value. `config: \n host: null `
```
ngx_log(ngx_ERR, _log_prefix, " ----- HOST DEFAULT VAL:  " .. conf.host)
prints 
----- HOST DEFAULT VAL:  localhost
```



This bug is just a demonstration how fickle the contract of leaving empty vs configuring it. I think you should rather fix it by `use_env` or just remove the default of the `datadog` plugin.

fffonion:
@jaswanthikolla My previous comment was simply to point out the logic of code, that it's not a bug and code is executed correctly as designed. I do agree on the bad user experience here. IMO the proper way of solving this is not to going back and forth on designing how a null value should be handled in kong, but rather to provide a proper way for user to configure plugins and other kong entities. And the answer to this as @bungle has proposed:

    Kong now has Secret Management built in. So there is another option:

    - mark a string field as referenceable=true
    - use reference like {vault://env/datadog-host}

hbagdi:
@jaswanthikolla Apologies for misleading you and others.
The solution that @fffonion has proposed above is indeed the "right" solution so long as Kong is concerned and it also the easiest one to implement, maintain and test.

The way this will look to an operator is:

```
config:
  host: "{vault://env/DATADOG_HOST}"
```

The end result should be the same and this should provide a much better user experience to all the users.
Is that acceptable to you?

chazdnato:
> The way this will look to an operator is:
> 
> 
> 
> ```
> 
> config:
> 
>   host: "{vault://env/DATADOG_HOST}"
> 
> ```
> 
> 
> 
> The end result should be the same and this should provide a much better user experience to all the users.
> 
> Is that acceptable to you?

Will this work with non-enterprise Kong, and which version do we need to run to get this feature? Thanks!

hbagdi:
> Will this work with non-enterprise Kong, and which version do we need to run to get this feature?

Absolutely. Environment variable based secrets are supported in open-source Kong.

> Does this work with the kubernetes-ingress-controller (which version)?


The version of the ingress controller shouldn't matter but the Kong gateway version will matter (needs to be latest).

hbagdi:
A gentle ping to move this forward @chazdnato @jaswanthikolla.
We would like to fix this problem for OpenTelemetery and Datadog plugins.
I'm not aware of any other plugins that face this problem but if there are, then we should consider including them as well.

chazdnato:
> A gentle ping to move this forward @chazdnato @jaswanthikolla. We would like to fix this problem for OpenTelemetery and Datadog plugins. I'm not aware of any other plugins that face this problem but if there are, then we should consider including them as well.

I am content to close out this investigation, as the [workaround suggested previously] (https://github.com/Kong/kong/issues/7954#issuecomment-1033221716) is fine. I would like to test this on Kong 3.x, but we're not upgraded yet. The only documentation I can find for `vault://env/` usage in the Enterprise version of Kong. Will this be updated?

Thank you for looping back!

hbagdi:
> I am content to close out this investigation, as the [workaround suggested previously] (https://github.com/Kong/kong/issues/7954#issuecomment-1033221716) is fine. I would like to test this on Kong 3.x, but we're not upgraded yet. The only documentation I can find for vault://env/ usage in the Enterprise version of Kong. Will this be updated?

I think there is some confusion.
The approach mentioned above is something can be implemented. It doesn't actually work this way today and there is some work to be done. Would be you be interested in submitting a patch for it?
It involves adding `referenceable = true` to the appropriate field schemas.

supercharger:
@hbagdi  Secrets doesn't work because they are cluster specific not host specific. 
`DATADOG_HOST` as environment variable is different for each host where as if you define it as secret it's static at cluster level.

chazdnato:
Given the solutions discussed and the way forward, I am closing this in favour of https://github.com/Kong/kong/pull/10484

PR #9331, opened 2022-08-27
Refactor/atc router clean

chronolaw:
### Summary

Refactoring router's code, separate expressions/compat.

PR #9477, opened 2022-09-27
feat(report): route statistic

StarlightIbuki:
Collect statistics of configured routers and runtime usage:
1. How many routes are configured;
2. How many paths are configured for routes;
3. How many headers are configured for routes;
4. The used router flavor;
5. How many routes are configured for different kinds of protocols;
6. How many routes are configured with different path handling;
7. How many regex paths matched

Configured route statistics are collected every time the router is rebuilt. The statistics are sent along with the ping.

Fix FT-3217

### DIFF
--- a/kong/reports.lua
+++ b/kong/reports.lua
@@ -465,8 +465,10 @@ return {
       incr_counter(count_key .. ":" .. suffix)
     end
 
-    if ctx.route_match_cached then
-      incr_counter(count_key .. ":" .. ROUTE_CACHE_HITS_KEY .. ":" .. ctx.route_match_cached)
+    local route_match_cached = ctx.route_match_cached
+
+    if route_match_cached then
+      incr_counter(count_key .. ":" .. ROUTE_CACHE_HITS_KEY .. ":" .. route_match_cached)
     end
   end,
 
--- a/kong/reports.lua
+++ b/kong/reports.lua
@@ -48,6 +48,13 @@ local UDP_STREAM_COUNT_KEY    = "events:streams:udp"
 local GO_PLUGINS_REQUEST_COUNT_KEY = "events:requests:go_plugins"
 
 
+local ROUTE_CACHE_HITS_KEY = "route_cache_hits"
+local STEAM_ROUTE_CACHE_HITS_KEY_POS = STREAM_COUNT_KEY .. ":" .. ROUTE_CACHE_HITS_KEY .. ":pos"
+local STEAM_ROUTE_CACHE_HITS_KEY_NEG = STREAM_COUNT_KEY .. ":" .. ROUTE_CACHE_HITS_KEY .. ":neg"
+local REQUEST_ROUTE_CACHE_HITS_KEY_POS = REQUEST_COUNT_KEY .. ":" .. ROUTE_CACHE_HITS_KEY .. ":pos"
+local REQUEST_ROUTE_CACHE_HITS_KEY_NEG = REQUEST_COUNT_KEY .. ":" .. ROUTE_CACHE_HITS_KEY .. ":neg"
+
+
 local _buffer = {}
 local _ping_infos = {}
 local _enabled = false
@@ -294,6 +301,9 @@ local function send_ping(host, port)
     _ping_infos.tls_streams = get_counter(TLS_STREAM_COUNT_KEY)
     _ping_infos.go_plugin_reqs = get_counter(GO_PLUGINS_REQUEST_COUNT_KEY)
 
+    _ping_infos.stream_route_cache_hit_pos = get_counter(STEAM_ROUTE_CACHE_HITS_KEY_POS)
+    _ping_infos.stream_route_cache_hit_neg = get_counter(STEAM_ROUTE_CACHE_HITS_KEY_NEG)
+
     send_report("ping", _ping_infos, host, port)
 
     reset_counter(STREAM_COUNT_KEY, _ping_infos.streams)
@@ -301,7 +311,8 @@ local function send_ping(host, port)
     reset_counter(UDP_STREAM_COUNT_KEY, _ping_infos.udp_streams)
     reset_counter(TLS_STREAM_COUNT_KEY, _ping_infos.tls_streams)
     reset_counter(GO_PLUGINS_REQUEST_COUNT_KEY, _ping_infos.go_plugin_reqs)
-
+    reset_counter(STEAM_ROUTE_CACHE_HITS_KEY_POS)
+    reset_counter(STEAM_ROUTE_CACHE_HITS_KEY_NEG)
     return
   end
 
@@ -316,6 +327,9 @@ local function send_ping(host, port)
   _ping_infos.wss_reqs       = get_counter(WSS_REQUEST_COUNT_KEY)
   _ping_infos.go_plugin_reqs = get_counter(GO_PLUGINS_REQUEST_COUNT_KEY)
 
+  _ping_infos.request_route_cache_hit_pos = get_counter(REQUEST_ROUTE_CACHE_HITS_KEY_POS)
+  _ping_infos.request_route_cache_hit_neg = get_counter(REQUEST_ROUTE_CACHE_HITS_KEY_NEG)
+
   send_report("ping", _ping_infos, host, port)
 
   reset_counter(REQUEST_COUNT_KEY,       _ping_infos.requests)
@@ -328,6 +342,8 @@ local function send_ping(host, port)
   reset_counter(WS_REQUEST_COUNT_KEY,    _ping_infos.ws_reqs)
   reset_counter(WSS_REQUEST_COUNT_KEY,   _ping_infos.wss_reqs)
   reset_counter(GO_PLUGINS_REQUEST_COUNT_KEY, _ping_infos.go_plugin_reqs)
+  reset_counter(REQUEST_ROUTE_CACHE_HITS_KEY_POS)
+  reset_counter(REQUEST_ROUTE_CACHE_HITS_KEY_NEG)
 end
 
 
@@ -447,6 +463,10 @@ return {
     local suffix = get_current_suffix(ctx)
     if suffix then
       incr_counter(count_key .. ":" .. suffix)
+    end
+
+    if ctx.route_match_cached then
+      incr_counter(count_key .. ":" .. ROUTE_CACHE_HITS_KEY .. ":" .. ctx.route_match_cached)
     end
   end,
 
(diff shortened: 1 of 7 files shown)


spec/kong_tests.conf
```diff
@@ -41,3 +41,4 @@ lua_package_path = ./spec/fixtures/custom_plugins/?.lua;./spec/fixtures/custom_v
 untrusted_lua = sandbox
 
 vaults = bundled
+pg_password = kong_tests
```
StarlightIbuki, line 44:
```suggestion
```

spec/02-integration/05-proxy/22-reports_spec.lua
```diff
@@ -431,5 +431,100 @@ for _, strategy in helpers.each_strategy() do
...
+        assert.match([["grpc":2]], reports_data)
+        assert.match([["stream":0]], reports_data)
+        assert.match([["tls_passthrough":0]], reports_data)
+        assert.match([["flavor":"traditional_compatible"]], reports_data)
+        assert.match([["paths":3]], reports_data)
+        assert.match([["regex_routes":1]], reports_data)
+        assert.match([["v1":1]], reports_data)
+        assert.match([["v0":5]], reports_data)
+        assert.match([[request_route_cache_hit_pos=2]], reports_data)
+        assert.match([[request_route_cache_hit_neg=4]], reports_data)
```
StarlightIbuki, line 525: This may become a source of test flakiness because the wait util may happen multiple times and increase the time neg cache hit.

kong/reports.lua
```diff
@@ -448,6 +464,10 @@ return {
     if suffix then
       incr_counter(count_key .. ":" .. suffix)
     end
+
+    if ctx.route_match_cached then
```
chronolaw, line 468: we can cache this var then use it, like `local route_match_cached = ctx.route_match_cached`

kong/router/traditional.lua
```diff
@@ -1678,16 +1678,18 @@ function _M.new(routes, cache, cache_neg)
 
       req_uri = strip_uri_args(req_uri)
 
-      local match_t = find_route(req_method, req_uri, req_host, req_scheme,
+      local match_t, cached = find_route(req_method, req_uri, req_host, req_scheme,
                                  nil, nil, -- src_ip, src_port
```
chronolaw, line 1682: align to `(`

kong/router/init.lua
```diff
@@ -30,7 +30,137 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
 end
 
 
+
+local phonehome_statistics
```
chronolaw, line 34: maybe we could move this function into `router/utils.lua`, then we can keep `init.lua` clean

kong/router/atc_compat.lua
```diff
@@ -745,6 +748,11 @@ function _M:exec(ctx)
     end
 
     self.cache:set(cache_key, match_t)
+
+  else
+    if ctx then
```
chronolaw, line 753: we can move this `if ctx then ... end` out of `if not match`, after `-- found a match`

kong/router/init.lua
```diff
@@ -30,7 +30,137 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
...
+    local headers_count = 0
+    local regex_paths_count = 0
+    local http = 0
+    local stream = 0
+    local tls_passthrough = 0
+    local grpc = 0
+    local v0 = 0
+    local v1 = 0
+
+    for _, route in ipairs(routes) do
```
chronolaw, line 80: rename this `route` to `r`, then the next line will be `route = r.route`, it is more readable.

kong/router/atc_compat.lua
StarlightIbuki: No. This is not the logic you think it is. Out of the branch, it can be both cached or not cached. Only inside the branch, it is cached.

kong/router/init.lua
```diff
@@ -30,7 +31,11 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
 end
 
 
+
+_M.phonehome_statistics = phonehome_statistics
+
 function _M.new(routes, cache, cache_neg, old_router)
+  phonehome_statistics(routes)
```
chronolaw, line 38: Adding a blank line may be more readable.

kong/router/utils.lua
```diff
@@ -234,6 +234,131 @@ local function get_service_info(service)
...
 end
 
+local phonehome_statistics
+do
+  local reports = require("kong.reports")
+  local nkeys = require("table.nkeys")
+  local log = ngx.log
+  local ERR = ngx.ERR
+
+  local byte = string.byte
```
chronolaw, line 244: It seems we already have `local byte = string.byte` in this module, please check it.

kong/router/utils.lua
```diff
@@ -234,6 +234,131 @@ local function get_service_info(service)
...
+    protocols.http = http
+    protocols.stream = stream
+    protocols.tls_passthrough = tls_passthrough
+    protocols.grpc = grpc
+    path_handlings.v0 = v0
+    path_handlings.v1 = v1
+  end
+
+  function phonehome_statistics(routes)
+    if not kong.configuration.anonymous_reports or ngx.worker.id() ~= 0 then
```
chronolaw, line 337: We should cache `ngx.worker.id` in this `do end` block.

kong/router/utils.lua
```diff
@@ -234,6 +234,131 @@ local function get_service_info(service)
...
+      return
+    end
+
+    route_report.flavor = kong.configuration.router_flavor
```
chronolaw, line 341: cache `kong.configuration.router_flavor` then use it.

kong/router/utils.lua
```diff
@@ -234,6 +234,131 @@ local function get_service_info(service)
...
+    route_report.routes = #routes
+
+    if route_report.flavor ~= "expressions" then
```
chronolaw, line 344: Use cached `kong.configuration.router_flavor`.

kong/router/utils.lua
StarlightIbuki: If that is true lint should warn about it.

kong/router/utils.lua
StarlightIbuki: The best we can do is to cache `ngx.worker` otherwise we will have id = 0 for all workers.

chronolaw: My meaning is that we could use `local` to cache/localize the function `ngx.worker.id`.

kong/router/utils.lua
chronolaw: See https://github.com/Kong/kong/blob/f416c76e88cb1f1d189c84d2f0b6cf9404c4435d/kong/router/utils.lua#L9

chronolaw:
Since router's code has been refactored (PR #9331), we should rebase and change some code.

spec/02-integration/05-proxy/22-reports_spec.lua
fffonion: instead of checking value, could we just check `request_route_cache_hit_neg=whatever` is existence instead?

kong/router/utils.lua
```diff
@@ -234,6 +234,133 @@ local function get_service_info(service)
...
+
+  local TILDE = byte("~")
+  local function is_regex_magic(path)
+    return byte(path) == TILDE
+  end
+
+  local empty_table = {}
+  -- reuse tables to avoid cost of creating tables and garbage collection
+  local protocols = {
+    http = 0, -- { "http", "https" },
```
chronolaw, line 253: Algin all `=` and `--` should be more readable.

kong/router/atc.lua
```diff
@@ -417,6 +417,7 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
     service         = service,
     prefix          = request_prefix,
     matches = {
+      type = captures and "regex" or "prefix",
```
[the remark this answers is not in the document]
StarlightIbuki: It was introduced by a mistaken cherry-pick. Removing it.

kong/router/init.lua
```diff
@@ -33,11 +32,16 @@ function _M:select(req_method, req_uri, req_host, req_scheme,
...
 
+
+_M.phonehome_statistics = phonehome_statistics
+
 function _M.new(routes, cache, cache_neg, old_router)
   local flavor = kong and
                  kong.configuration and
                  kong.configuration.router_flavor
 
+  phonehome_statistics(routes)
```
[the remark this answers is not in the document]
StarlightIbuki: It's not a hot loop and I don't want to introduce a parameter just for that.

kong/router/utils.lua
```diff
@@ -234,6 +234,133 @@ local function get_service_info(service)
...
+        elseif protocol == "tls_passthrough" then
+          tls_passthrough = tls_passthrough + 1
+
+        elseif protocol == "grpc" or protocol == "grpcs" then
+          grpc = grpc + 1
+
+        else
+          log(ERR, "unknown protocol: " .. protocol)
+        end
+        break
```
chronolaw, line 313: What does this `break` mean?

StarlightIbuki: Those protocol types are mutually exclusive; and for the same type like HTTP and HTTPS, we should only count for once.

chronolaw: Got it, thanks.

kong/router/traditional.lua
```diff
@@ -1530,11 +1530,11 @@ function _M.new(routes, cache, cache_neg)
                                  .. "|" .. sni .. headers_key
     local match_t = cache:get(cache_key)
     if match_t then
-      return match_t
+      return match_t, "pos"
```
fffonion, line 1533: instead of using a second return parameter (usually the place for `err`), consider return it in match_t to avoid confusion. for neg match, we might return a global static table.

on the other side, what's the reason of not setting `ctx.route_match_cached = xxx` inside find_route?

kong/router/utils.lua
```diff
@@ -234,6 +234,133 @@ local function get_service_info(service)
...
+          stream = stream + 1
+
+        elseif protocol == "tls_passthrough" then
+          tls_passthrough = tls_passthrough + 1
+
+        elseif protocol == "grpc" or protocol == "grpcs" then
+          grpc = grpc + 1
+
+        else
+          log(ERR, "unknown protocol: " .. protocol)
```
fffonion, line 311: if this happens, user can do nothing, so let's not log error it. instead, consider sending an `unknown` in `protocols` table. we also have `ws` and `wss` in ee.

fffonion (approved):

fffonion:
do we need changelog entry for this? @StarlightIbuki

PR #9478, opened 2022-09-27
fix(cmd): kong prepare to write out .kong_process_secrets

bungle:
### Summary

Kong Docker image for example starts Kong by running `kong prepare` and then executes `nginx -c ... -p ...` as seen in:
https://github.com/Kong/docker-kong/blob/master/alpine/docker-entrypoint.sh

This causes issues with Kong secret management references. By default, Kong passes secrets to nginx using environment variable when using `kong start`, but here the `nginx` is started directly without calling `kong start` and thus the secrets are not available for Kong init. This commit fixes that.

### DIFF
--- a/kong/cmd/prepare.lua
+++ b/kong/cmd/prepare.lua
@@ -7,7 +7,7 @@ local function execute(args)
     prefix = args.prefix
   }))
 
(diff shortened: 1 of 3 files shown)


jschmid1 (approved):
lgtm

PR #9480, opened 2022-09-28
fix(router/atc): check expression building error

chronolaw:
### Summary

Fix FT-3429

### DIFF
--- a/kong/router/atc_compat.lua
+++ b/kong/router/atc_compat.lua
@@ -422,8 +422,8 @@ local function add_atc_matcher(inst, route, route_id,
 
   end
 
-  if remove_existing and not inst:remove_matcher(route_id) then
-    return nil, "could not remove route: " .. route_id
+  if remove_existing then
+    assert(inst:remove_matcher(route_id))
   end
 
   local ok, err = inst:add_matcher(priority, route_id, atc)
@@ -549,9 +549,7 @@ local function new_from_previous(routes, is_traditional_compatible, old_router)
       r.seen = nil
 
     else
-      if not inst:remove_matcher(id) then
-        ngx_log(ngx_WARN, "could not remove route: ", id)
-      end
+      assert(inst:remove_matcher(id))
 
       old_routes[id] = nil
       old_services[id] = nil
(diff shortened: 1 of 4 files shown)


spec/01-unit/08-router_spec.lua
```diff
@@ -1753,6 +1753,28 @@ for _, flavor in ipairs({ "traditional", "traditional_compatible", "expressions"
...
+                  id = "e8fb37f1-102d-461e-9c51-6608a6bb8101",
+                  paths = { "~/delay/(?<delay>[^\\/]+)$", },
+                  updated_at = 100,
+                },
+              },
+            }
+
+            local ok, nrouter = pcall(new_router, use_case, router)
+
+            assert.falsy(router.rebuilding)
```
dndx, line 1770: We should not use `router.rebuilding` for testing this change, as it is a internal flag that may be changed later.

This should be tested with integration tests like https://github.com/Kong/kong/blob/master/spec/02-integration/05-proxy/02-router_spec.lua#L2311 that directly observes the end result of a successful rebuild instead of the flag.

dndx (changes requested):

spec/01-unit/08-router_spec.lua
chronolaw: I will try it.

chronolaw: done.

kong/router/atc_compat.lua
```diff
@@ -524,7 +549,9 @@ local function new_from_previous(routes, is_traditional_compatible, old_router)
       r.seen = nil
 
     else
-      assert(inst:remove_matcher(id))
+      if not inst:remove_matcher(id) then
```
dndx, line 552: I think we can keep the `assert` for `remove_matcher`, it should really never fail and hiding it is not a good idea.

kong/router/atc_compat.lua
```diff
@@ -421,11 +422,16 @@ local function add_atc_matcher(inst, route, route_id,
 
   end
 
-  if remove_existing then
-    assert(inst:remove_matcher(route_id))
+  if remove_existing and not inst:remove_matcher(route_id) then
```
dndx, line 425: Same, we can keep the `assert` for `remove_matcher`.

kong/router/atc_compat.lua
chronolaw: done

kong/router/atc_compat.lua
chronolaw: done

dndx (approved):

Issue #9090, opened 2022-07-13
Restarted workers in kong pod use stale config


Benny-Git:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

2.8.1

### Current Behavior

We switch a target URL on a service using the Admin API, and the change gets applied immediately. Everything is working as expected.
If however a worker is restarted (can be seen by message "start worker process" in logs), calls to the API suddenly arrive at the old target URL when routed via this worker.

### Expected Behavior

A newly started worker should always have the up-to-date config, and not rely on some stale cache.

### Steps To Reproduce

```markdown
I have been able to "reliably" recreate the issue in our DEV environment:

1. Creat service pointing to http://httpbin.utility:80/anything
2. Kill one one of the Kong DEV pods
3. On my laptop run watch -n 1 curl <route_url>
4. Requests go to upstream: http://192.168.99.186:80/anything with worker 1140
5. Edit service via to have path /get instead of /anything
6. Observe via logs the upstream changed to http://192.168.99.186:80/get with worker 1140
7. ps shows active workers 1140 and 1141
8. Remove one with kill 1140. New worker 1145 is created
9. Observe upstream is still http://192.168.99.186:80/get with worker 1141
10. Remove another one with kill 1141. New worker 1146 is created.
11. Observe upstream is now http://192.168.99.186:80/anything served by worker 1145 ⚡
12. After running curl -XDELETE localhost:8001/cache worker 1145 also serves http://192.168.99.186:80/get

Interestingly enough, repeating the steps above with switching the path from /get to /anything does not result in the issue -- this only worked reliably with switching the path from /anything to /get.
```

### Anything else?

We got hit by this several times now, even with instances where we created a new route that was then served, but after a worker restart did not get served anymore (receiving a "no Route matched with those values" instead).

We are running Kong on Kubernetes with 2 workers (default in Kong Helm chart), with Postgres as our database.

Are there any debug trace settings that could help pinpoint the root cause?

We originally reported this in https://github.com/Kong/kong/issues/8517

Benjamin Bertow benjamin.bertow@mercedes-benz.com, Mercedes-Benz Tech Innovation GmbH.
https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md

Benny-Git:
> * From our limited trials, restarting the proxy pod does seem to resolve the problem.

[...]

PR #9485, opened 2022-09-29
fix(*): do not use stale router data if workers are respawned

bungle:
### Summary

@Benny-Git reported on #9090 about issue of stale data picked up for router in case the worker was restarted.

Some insights:

There are several ways to restart workers:
1. you reload them with e.g. kong reload or similar nginx signal
2. worker crashes
3. worker is killed

On case 1 the `init` phase will be re-executed, on 2 and 3 it will not.

This commit ensures that the router is rebuild on init worker if it is not current (by default all workers are spawned from master process LuaJIT VM and that is not reintialized on cases 2 and 3).

FT-3432

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -716,22 +716,31 @@ function Kong.init_worker()
     end
   end
 
-  if kong.configuration.role ~= "control_plane" then
+  local is_not_control_plane = kong.configuration.role ~= "control_plane"
+  if is_not_control_plane then
     ok, err = execute_cache_warmup(kong.configuration)
     if not ok then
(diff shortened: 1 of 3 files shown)


hanshuebner (approved):
I've verified that the problem can be reproduced without the fix and that the fix solves it.

chronolaw (approved):

samugi:
I modified [a test](https://github.com/Kong/kong/pull/9485/commits/e6a9a6d96184ed8b7edb4a11cf906128ca286ca8?diff=split&w=1) that should now cover this scenario.

Benny-Git:
I see the PR has a milestone of 3.0.1. Will this also be backported to a potential 2.8.2 release?

PR #9490, opened 2022-09-30
chore(*): include libexpat, libxml2, expadom and xmlua in Kong

hanshuebner:
This PR adds the libexpat and libxml2 shared library and the expadom and xmlua libraries to the Kong base image.  This is to support a number of upcoming plugins that process XML in one way or another.  LibXML2 is included in a minimal configuration as it is intended to be used primarily for XML canonicalization.  expadom is the prefered library for XML processing in Lua and should be recommended to plugin authors.

FT-3382


.requirements
```diff
@@ -12,3 +12,5 @@ ATC_ROUTER_VERSION=1.0.0
 LIBYAML_VERSION=0.2.5
 KONG_BUILD_TOOLS_VERSION=4.33.17
```
hutchic, line 13:
```suggestion
KONG_BUILD_TOOLS_VERSION=4.33.25
```

hutchic (commented):
bumping KBT should resolve this failure

kong-3.1.0-0.rockspec
```diff
@@ -42,6 +42,8 @@ dependencies = {
   "lua-resty-acme == 0.8.1",
   "lua-resty-session == 3.10",
   "lua-resty-timer-ng == 0.2.0",
+  "expadom >= 0.1.0",
```
Tieske, line 45: since expadom is in the "ustable range" (semver, before 1.0) this one should probably be pinned exactly to this version, instead of open ended.

hanshuebner:
Moving to -ee

hanshuebner:
> If we need to use pre-install / post-install in kong ce those scripts should likely just go into kong-build-tools proper.

The XML libraries will be made available in kong-ee only.

PR #9494, opened 2022-10-03
refactor(handler): trying to make reconfigure more atomic

bungle:
### Summary

Originally the `reconfigure` event was `atomic` and it was not yielding. But this caused huge spikes in latencies when new configurations where applied.

We started to add yielding in different places that happen in `reconfigure` event. It may now lead to nasty things like: new router, and old plugins iterator gets used etc. (router from new configuration and plugin iterator from old configuration).

I needed to refactor the handler so that it could at least support more atomic flip of `router` and `plugins iterator`. The balancer is still problematic as it is shared global state and a new balancer cannot currently be build behind the scenes cooperatively, and then flip it atomically together with router and plugins iterator. I asked @locao to research if that could be doable. I leave that to other PR.

### Issue reference

Fix FTI-4369

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -66,6 +66,7 @@ local ERR   = ngx.ERR
 local CRIT  = ngx.CRIT
 local NOTICE = ngx.NOTICE
 local WARN  = ngx.WARN
+local INFO  = ngx.INFO
 local DEBUG = ngx.DEBUG
 local COMMA = byte(",")
 local SPACE = byte(" ")
@@ -704,12 +705,17 @@ local function register_events()
     local current_plugins_hash
     local current_balancer_hash
 
+    local now = ngx.now
+    local update_time = ngx.update_time
+    local worker_id = ngx.worker.id()
+
     local exiting = ngx.worker.exiting
     local function is_exiting()
       if not exiting() then
         return false
       end
-      log(NOTICE, "declarative reconfigure canceled: process exiting")
+      log(NOTICE, "declarative reconfigure was canceled on worker #", worker_id,
+                  ": process exiting")
       return true
     end
 
@@ -717,6 +723,11 @@ local function register_events()
       if is_exiting() then
         return true
       end
+
+      update_time()
+      local reconfigure_started_at = now() * 1000
+
+      log(INFO, "declarative reconfigure was started on worker #", worker_id)
 
       local default_ws
       local router_hash
@@ -735,6 +746,7 @@ local function register_events()
 
         local rebuild_balancer = balancer_hash == nil or balancer_hash ~= current_balancer_hash
         if rebuild_balancer then
+          log(DEBUG, "stopping previously started health checkers on worker #", worker_id)
           balancer.stop_healthcheckers(CLEAR_HEALTH_STATUS_DELAY)
         end
 
@@ -743,23 +755,40 @@ local function register_events()
 
         local router, err
         if router_hash == nil or router_hash ~= current_router_hash then
+          update_time()
+          local start = now() * 1000
+
           router, err = new_router()
           if not router then
             return nil, err
           end
+
+          update_time()
+          log(INFO, "building a new router took ",  now() * 1000 - start,
+                    " ms on worker #", worker_id)
         end
 
         local plugins_iterator
         if plugins_hash == nil or plugins_hash ~= current_plugins_hash then
+          update_time()
+          local start = now() * 1000
+
           plugins_iterator, err = new_plugins_iterator()
           if not plugins_iterator then
             return nil, err
           end
+
+          update_time()
+          log(INFO, "building a new plugins iterator took ", now() * 1000 - start,
+                    " ms on worker #", worker_id)
         end
 
         -- below you are not supposed to yield and this should be fast and atomic
 
         -- TODO: we should perhaps only purge the configuration related cache.
+
+        log(DEBUG, "flushing caches as part of the reconfiguration on worker #", worker_id)
+
         kong.core_cache:purge()
         kong.cache:purge()
 
@@ -778,15 +807,22 @@ local function register_events()
         if rebuild_balancer then
           -- TODO: balancer is a big blob of global state and you cannot easily
           --       initialize new balancer and then atomically flip it.
+          log(DEBUG, "reinitializing balancer with a new configuration on worker #", worker_id)
           balancer.init()
           current_balancer_hash = balancer_hash
         end
+
+        update_time()
+        log(INFO, "declarative reconfigure took ", now() * 1000 - reconfigure_started_at,
+                  " ms on worker #", worker_id)
 
         return true
       end)
 
       if not ok then
-        log(ERR, "reconfigure failed: ", err)
+        update_time()
+        log(ERR, "declarative reconfigure failed after ", now() * 1000 - reconfigure_started_at,
+                 " ms on worker #", worker_id, ": ", err)
       end
     end, "declarative", "reconfigure")
 
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -23,6 +23,9 @@ local ipairs            = ipairs
 local tostring          = tostring
 local tonumber          = tonumber
 local setmetatable      = setmetatable
+local max               = math.max
+local min               = math.min
+local ceil              = math.ceil
 local sub               = string.sub
 local byte              = string.byte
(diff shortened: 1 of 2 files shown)


kong/runloop/handler.lua
```diff
@@ -206,10 +209,346 @@ local function csv(s)
...
+  end
+
+  local counter = 0
+  local page_size = db.routes.pagination.max_page_size
+  for route, err in db.routes:each(page_size, GLOBAL_QUERY_OPTS) do
+    if err then
+      return nil, "could not load routes: " .. err
+    end
+
+    if db.strategy ~= "off" and kong.core_cache then
```
chronolaw, line 357: we can define a local var and then move this condition `db.strategy ~= "off" and kong.core_cache` out of the loop.

kong/runloop/handler.lua
```diff
@@ -1092,6 +1136,11 @@ return {
 
   init_worker = {
     before = function()
+      local prefix = kong.configuration.prefix or ngx.config.prefix()
```
chronolaw, line 1139: Please refer to PR #9337, maybe we will introduce a new config var later.

kong/runloop/handler.lua
bungle: @chronolaw I moved this to a variable before loop

kong/runloop/handler.lua
bungle: @chronolaw I added comment above that line

chronolaw:
It seems that this PR is mostly moving functions from one place to another, am I right?
Could you describe the main changes so I would catch the point?

bungle:
> It seems that this PR is mostly moving functions from one place to another, am I right? Could you describe the key changes so I would catch the point?

The key change are in:
1. declarative reconfigure event handler: build things, and switch atomically, the idea in short pseudo event handler:
```lua
-- yielding code
local new_router = router.new()
local new_iterator = iterator.new()
local new_balancer = balancer.new()

-- non-yielding code:
router = new_router
iterator = new_iterator
balancer = new_balancer
flush_cache()
flush_router_lru()
```
2. build_router: no side effects, just build a router and return it (no cache clear, no switching the proxy router)
3. build_plugins_iterator: no side effects, just build a plugins iterator and return it (no switching the proxy plugins iterator)

(the balancer is still not switched atomically as it is just a blob of global state).

You can see the most meaningful change here:
https://github.com/Kong/kong/pull/9494/commits/b0741c55a77fbda7e42eaea013c9937e0e2aa928#diff-093344a0609553c69b475fb5cbd57a202796383eb6fca553d26ac6a1f40d4492R727-R777

^ that is the closest I could get to that atomic pseudo switch above.

kong/runloop/handler.lua
```diff
@@ -364,12 +707,22 @@ local function register_events()
     local current_plugins_hash
     local current_balancer_hash
 
+    local now = ngx.now
+    local update_time = ngx.update_time
+    local worker_id = ngx.worker.id()
+
     worker_events.register(function(data)
       if ngx.worker.exiting() then
```
chronolaw, line 715: Shall we localize `ngx.worker.exiting`?

chronolaw (approved):

kong/runloop/handler.lua
bungle: I localized it.

kong/runloop/handler.lua
```diff
@@ -383,37 +744,87 @@ local function register_events()
...
         kong.default_workspace = default_ws
-        ngx.ctx.workspace = kong.default_workspace
+        ngx.ctx.workspace = default_ws
+
+        local router, err
+        if router_hash == nil or router_hash ~= current_router_hash then
+          update_time()
+          local start = now() * 1000
+
+          router, err = new_router()
```
StarlightIbuki, line 763: The naming question applies to routers, balancers, and plugins.

kong/runloop/handler.lua
```diff
@@ -383,37 +744,87 @@ local function register_events()
...
 
-        kong.core_cache:purge()
-        kong.cache:purge()
-
         kong.default_workspace = default_ws
-        ngx.ctx.workspace = kong.default_workspace
+        ngx.ctx.workspace = default_ws
+
+        local router, err
+        if router_hash == nil or router_hash ~= current_router_hash then
```
StarlightIbuki, line 759: Is there any reason we change the rebuild order to balancers, routers, and then plugins?

kong/runloop/handler.lua
```diff
@@ -1136,7 +1195,7 @@ return {
           }
         end
 
-        if strategy == "off" or kong.configuration.worker_consistency == "strict" then
+        if kong.configuration.worker_consistency == "strict" then
```
StarlightIbuki, line 1198:
```lua
get_updated_router = function()
  if kong.db.strategy ~= "off" and kong.configuration.worker_consistency == "strict" then
      local ok, err = rebuild_router(ROUTER_SYNC_OPTS)
```
Do we need to be consistent with the condition?

kong/runloop/handler.lua
```diff
@@ -1092,6 +1137,12 @@ return {
 
   init_worker = {
     before = function()
+      -- TODO: PR #9337 may affect the following line
+      local prefix = kong.configuration.prefix or ngx.config.prefix()
+
+      STREAM_TLS_TERMINATE_SOCK = fmt("unix:%s/stream_tls_terminate.sock", prefix)
+      STREAM_TLS_PASSTHROUGH_SOCK = fmt("unix:%s/stream_tls_passthrough.sock", prefix)
```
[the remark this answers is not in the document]
bungle: @StarlightIbuki `handler.lua` does not currently have `init` stuff.

kong/runloop/handler.lua
```diff
@@ -364,12 +707,22 @@ local function register_events()
...
+
     worker_events.register(function(data)
       if ngx.worker.exiting() then
-        log(NOTICE, "declarative reconfigure canceled: process exiting")
+        log(NOTICE, "declarative reconfigure was canceled on worker #", worker_id,
+                    ": process exiting")
         return true
       end
 
+      update_time()
```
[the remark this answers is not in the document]
bungle: @StarlightIbuki I was asked to add logging statements with timing. We are not going to remove them. We talked about using the tracing system, but people said that log statements are fine for now.

kong/runloop/handler.lua
bungle: Those options are not used with `dbless`.

kong/runloop/handler.lua
bungle: The order is:
1. stop current healtcheckers (if there are balancer changes) (manipulates global state)
2. build router (if there are router changes) (does not manipulate global state)
3. build plugins iterator (if there are plugins changes) (does not manipulate global state)

And then the non-yieding part of it:
1. flush kong cache and kong core cache
2. flush  router cache (if there were router changes)
3. switch over the router (if there were router changes)
4. switch over the plugins iterator (if there were plugin changes)
5. reinitialize balancer

The balancer is biggest culprit, we need to make it so that we can create instances of it, so that we can initiate a new one in background. Of course incremental etc. things come to play with this too. The point is that a single change set, we need to apply atomically so that all changes on proxy side are applied the same time (non-yieding).

The order is more like what we use elsewhere, like in init etc. now. Shouldn't matter too much. I felt it is cleaner this way.

kong/runloop/handler.lua
```diff
@@ -206,10 +209,347 @@ local function csv(s)
...
+  if not new_router then
+    return nil, "could not create router: " .. err
+  end
+
+  return new_router
+end
+
+
+local function build_router(version)
+  local router, err = new_router(version)
```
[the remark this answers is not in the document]
bungle: then you have of all these test suite functions like:
```
_set_build_router
```

you need to rename them too, and then change tests etc.

kong/runloop/handler.lua
bungle: I added it inside `elseif`

kong/runloop/handler.lua
```diff
@@ -206,10 +209,347 @@ local function csv(s)
...
+end
+
+local function _set_router_version(v)
+  ROUTER_VERSION = v
+end
+
+
+local function new_plugins_iterator(version)
+  return PluginsIterator.new(version)
+end
```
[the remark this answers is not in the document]
bungle: Changed.

PR #9499, opened 2022-10-05
fix(balancer): use local target cache

locao:
### Summary

The target's global cache was caching more data than it should: it kept data retrieved from the database and DNS name resolution, which has its own cache.

As a result, when config reloads were done, some processes were skipped (such as setting the upstream as healthy) because the load-balancer never identified the data as new if it was the same as cached (highly probable with DNS name resolution).

This change makes the target cache local to the worker. 

### Full changelog

* Removed global target cache.
* Added local target cache in `kong/runloop/balancer/targets.lua`.


kikito:
I am going to remove this from the 2.8.2 milestone since it doesn't seem critical enough to be granted an exception over the code freeze. @locao please let me know if such reason exists.

PR #9502, opened 2022-10-07
Backport 9485

poelzi:
### Summary

This is a backport of #9485 to the 2.8 release branch

### Issue reference

#9485
#9090

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -661,22 +661,31 @@ function Kong.init_worker()
     return
   end
 
-  if kong.configuration.role ~= "control_plane" then
+  local is_not_control_plane = kong.configuration.role ~= "control_plane"
+  if is_not_control_plane then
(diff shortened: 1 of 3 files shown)


poelzi:
can i get a approval ?

chronolaw:
Thanks for your concern.

We have our own version plan, we only choose the necessary PR for the old release branch.
I think that we need an evaluation for this.

poelzi:
Our customers upgrade path is blocked by this. Also some comments in the reports indicate that we are not the only one having issues with this.

outsinre:
We encourage our customers to use 3.0 onwards. We don't backport 3.0 commits to 2.8.x unless manager approval. Otherwise, all customers want things in 3.0 backported to 2.8.x, which is undesirable. We'd like our engineer bandwidth focus on the main major version.

PR #9507, opened 2022-10-08
fix(pluginserver): error if req come before ready

StarlightIbuki:
We start plugin servers asynchronously while requests try to send RPC calls to the server. When the sock file is not deleted, the server may start a bit slower and the requests hit before it listen to the newly created sock file.

fix FTI-4282

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -802,10 +802,7 @@ function Kong.init_worker()
   runloop.init_worker.after()
 
   if is_not_control_plane then
-    if worker_id() == 0 then
-      plugin_servers.start()
-    end
-    plugin_servers.check()
+    plugin_servers.start()
   end
 
   if kong.clustering then
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -802,7 +802,10 @@ function Kong.init_worker()
   runloop.init_worker.after()
 
   if is_not_control_plane then
-    plugin_servers.start()
+    if worker_id() == 0 then
+      plugin_servers.start()
+    end
+    plugin_servers.check()
   end
 
   if kong.clustering then
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -368,7 +368,7 @@ function plugin_servers.start()
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
-      if ngx.worker.id() == 0 then
+      if worker_id() == 0 then
         native_timer_at(0, pluginserver_timer, server_def)
       end
       native_timer_at(0, pluginserver_connection_check_timer, server_def)
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -372,7 +372,7 @@ function plugin_servers.start()
     end
   end
 
-  local connection_check_timer = proc_mgmt.pluginserver_connection_check_timer
+  local connection_check_timer = proc_mgmt.connection_check_timer
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -362,16 +362,18 @@ end
 
 
 function plugin_servers.start()
-  local pluginserver_timer = proc_mgmt.pluginserver_timer
-  for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
-    if server_def.start_command then
-      native_timer_at(0, pluginserver_timer, server_def)
+  if worker_id() == 0 then
+    local pluginserver_timer = proc_mgmt.pluginserver_timer
+
+    for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
+      if server_def.start_command then
+        native_timer_at(0, pluginserver_timer, server_def)
+      end
     end
   end
-end
 
-function plugin_servers.check()
   local connection_check_timer = proc_mgmt.pluginserver_connection_check_timer
+
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
       native_timer_at(0, connection_check_timer, server_def)
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -364,7 +364,6 @@ end
 function plugin_servers.start()
   local pluginserver_timer = proc_mgmt.pluginserver_timer
   local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
-  local native_timer_at = _G.native_timer_at
   local handler = worker_id() == 0 and pluginserver_timer or pluginserver_connection_check_timer
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -364,13 +364,14 @@ end
 function plugin_servers.start()
   local pluginserver_timer = proc_mgmt.pluginserver_timer
   local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
+  local native_timer_at = _G.native_timer_at
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
       if ngx.worker.id() == 0 then
-        _G.native_timer_at(0, pluginserver_timer, server_def)
+        native_timer_at(0, pluginserver_timer, server_def)
       end
-      _G.native_timer_at(0, pluginserver_connection_check_timer, server_def)
+      native_timer_at(0, pluginserver_connection_check_timer, server_def)
     end
   end
 end
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -362,12 +362,19 @@ end
 
 
 function plugin_servers.start()
-  local handler = worker_id() == 0 and proc_mgmt.pluginserver_timer
-                  or proc_mgmt.pluginserver_connection_check_timer
-
+  local pluginserver_timer = proc_mgmt.pluginserver_timer
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
-      native_timer_at(0, handler, server_def)
+      native_timer_at(0, pluginserver_timer, server_def)
+    end
+  end
+end
+
+function plugin_servers.check()
+  local connection_check_timer = proc_mgmt.pluginserver_connection_check_timer
+  for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
+    if server_def.start_command then
+      native_timer_at(0, connection_check_timer, server_def)
     end
   end
 end
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -362,9 +362,8 @@ end
 
 
 function plugin_servers.start()
-  local pluginserver_timer = proc_mgmt.pluginserver_timer
-  local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
-  local handler = worker_id() == 0 and pluginserver_timer or pluginserver_connection_check_timer
+  local handler = worker_id() == 0 and proc_mgmt.pluginserver_timer
+                  or proc_mgmt.pluginserver_connection_check_timer
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -365,13 +365,11 @@ function plugin_servers.start()
   local pluginserver_timer = proc_mgmt.pluginserver_timer
   local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
   local native_timer_at = _G.native_timer_at
+  local handler = worker_id() == 0 and pluginserver_timer or pluginserver_connection_check_timer
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
-      if worker_id() == 0 then
-        native_timer_at(0, pluginserver_timer, server_def)
-      end
-      native_timer_at(0, pluginserver_connection_check_timer, server_def)
+      native_timer_at(0, handler, server_def)
     end
   end
 end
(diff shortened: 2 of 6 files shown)


kong/runloop/plugin_servers/init.lua
```diff
@@ -310,16 +320,15 @@ end
...
-  end
-
   local pluginserver_timer = proc_mgmt.pluginserver_timer
+  local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
-      _G.native_timer_at(0, pluginserver_timer, server_def)
+      if ngx.worker.id() == 0 then
+        _G.native_timer_at(0, pluginserver_timer, server_def)
```
chronolaw, line 329: We can use `local` to cache `_G.native_timer_at` in the for-loop

kong/runloop/plugin_servers/mp_rpc.lua
```diff
@@ -325,10 +325,17 @@ end
...
-  local _, err = bridge_loop(self, instance_id, phase)
+  local instance_id, err = self.get_instance_id(plugin_name, conf)
+  local _
+  if not err then
+    _, err = bridge_loop(self, instance_id, phase)
+  end
 
   if err then
+    if err == "not ready" then
+      ngx.sleep(0.1)
```
chronolaw, line 336: `0.1` is not a reasonable value, do we have better choice?

kong/runloop/plugin_servers/init.lua
StarlightIbuki: The problem, still, is that we cannot be sure if the variable is ready when we load the module.
So the best we can do is to cache the function in the local scope(which is not very meaningful)

kong/runloop/plugin_servers/mp_rpc.lua
fffonion: if err is `not ready` then just return early and don't handle the event. but make sure there's log, we can emit log in every 5 seconds for example, preferabaly in the `connection_check_timer` loop.

StarlightIbuki: I don't want to introduce a too large delay. 0.1 is just a step, not a timeout.

kong/runloop/plugin_servers/process.lua
```diff
@@ -237,6 +238,34 @@ function proc_mgmt.pluginserver_timer(premature, server_def)
...
+
+  local timeout = 500
+  local step = 0.1
+
+  local time = 0
+  local uri = "unix:" .. server_def.socket
+  local c, err
+  while time < timeout do
+    c, err = ngx.socket.connect(uri)
+    ngx.log(ngx.ERR, err)
```
chronolaw, line 258: If socket connect successfully, we should not log any message. Should we check `err` then log it?

kong/runloop/plugin_servers/process.lua
```diff
@@ -237,6 +243,48 @@ function proc_mgmt.pluginserver_timer(premature, server_def)
...
 
+-- if a plugin server is not ready after 20s of waiting we consider it failed
+local pluginserver_start_timeout = 20
+
+function proc_mgmt.connection_check_timer(premature, server_def)
+  if premature then
+    return
+  end
+
+  if ngx.config.subsystem ~= "http" then
```
chronolaw, line 254: We should localize `ngx.config.subsystem` in the module level.

fffonion:
looks like submodule `kong-dp-spec` was incorrectly included in this pr

kong/runloop/plugin_servers/init.lua
```diff
@@ -310,16 +320,16 @@ end
...
-  end
-
   local pluginserver_timer = proc_mgmt.pluginserver_timer
+  local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
+  local native_timer_at = _G.native_timer_at
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
-      _G.native_timer_at(0, pluginserver_timer, server_def)
+      if ngx.worker.id() == 0 then
```
chronolaw, line 329: localize `ngx.worker.id`, because it is in a loop.

kong/runloop/plugin_servers/init.lua
```diff
@@ -310,22 +322,22 @@ end
...
-  end
-
   local pluginserver_timer = proc_mgmt.pluginserver_timer
+  local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
+  local native_timer_at = _G.native_timer_at
 
   for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
     if server_def.start_command then
-      _G.native_timer_at(0, pluginserver_timer, server_def)
+      if worker_id() == 0 then
```
chronolaw, line 331: We can check `worker_id()` before the loop, then use `local call_back = pluginserver_xxx`,
so we could remove `if-end` in the loop.

StarlightIbuki: notice that that's not an if-else, the later handler will always run.

chronolaw: I saw the code had changed, so I think it should run only one handler in timer, right?

kong/runloop/plugin_servers/mp_rpc.lua
```diff
@@ -325,10 +327,16 @@ end
 
 
 function Rpc:handle_event(plugin_name, conf, phase)
-  local instance_id = self.get_instance_id(plugin_name, conf)
-  local _, err = bridge_loop(self, instance_id, phase)
+  local instance_id, err = self.get_instance_id(plugin_name, conf)
+  local _
```
chronolaw, line 331: We can define `local _, instance_id, err`, then use them to get a relatively clean code.

kong/runloop/plugin_servers/pb_rpc.lua
```diff
@@ -386,13 +387,19 @@ end
 
 
 function Rpc:handle_event(plugin_name, conf, phase)
-  local instance_id = self.get_instance_id(plugin_name, conf)
+  local instance_id, err = self.get_instance_id(plugin_name, conf)
```
chronolaw, line 390: We can define `local res, instance_id, err`, then use them to get a relatively clean code.

kong/runloop/plugin_servers/init.lua
```diff
@@ -14,11 +14,12 @@ local ngx = ngx
 local kong = kong
 local ngx_var = ngx.var
 local ngx_sleep = ngx.sleep
-local worker_id = ngx.worker.id
 local coroutine_running = coroutine.running
 local get_plugin_info = proc_mgmt.get_plugin_info
 local get_ctx_table = require("resty.core.ctx").get_ctx_table
 local subsystem = ngx.config.subsystem
+local worker_id = ngx.worker.id
```
chronolaw, line 21: `worker_id ` has been localized in PR #9547, we can remove this line.

kong/runloop/plugin_servers/init.lua
```diff
@@ -329,16 +340,14 @@ end
...
 
 function plugin_servers.start()
-  if worker_id() ~= 0 then
-    kong.log.notice("only worker #0 can manage")
-    return
-  end
-
   local pluginserver_timer = proc_mgmt.pluginserver_timer
+  local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
+  local native_timer_at = _G.native_timer_at
```
chronolaw, line 345: `native_timer_at` has been localized in #9547 now.

kong/runloop/plugin_servers/init.lua
```diff
@@ -329,16 +340,13 @@ end
 
 
 function plugin_servers.start()
-  if worker_id() ~= 0 then
-    kong.log.notice("only worker #0 can manage")
-    return
-  end
-
   local pluginserver_timer = proc_mgmt.pluginserver_timer
+  local pluginserver_connection_check_timer = proc_mgmt.connection_check_timer
```
chronolaw, line 344: We can remove `pluginserver_timer` and `pluginserver_connection_check_timer`,
because both of them will be set into var `handler`.

kong/runloop/plugin_servers/mp_rpc.lua
```diff
@@ -325,10 +327,16 @@ end
...
+  instance_id, err = self.get_instance_id(plugin_name, conf)
+  if not err then
+    _, err = bridge_loop(self, instance_id, phase)
+  end
 
   if err then
+    if err == "not ready" then
+      return handle_not_ready(plugin_name)
+    end
     if string.match(err:lower(), "no plugin instance") then
```
chronolaw, line 340: We should use `string.find` and localize it.

kong/runloop/plugin_servers/process.lua
```diff
@@ -197,6 +202,38 @@ local function grab_logs(proc, name)
...
```
chronolaw, line 213: This should be a module level var, like `local is_http = ngx.config.subsystem ~= "http"`

kong/runloop/plugin_servers/process.lua
```diff
@@ -208,6 +245,8 @@ function proc_mgmt.pluginserver_timer(premature, server_def)
 
   local next_spawn = 0
 
+  _G.native_timer_at(0, connection_check_timer, server_def)
```
chronolaw, line 248: Use cached `native_timer_at`

kong/runloop/plugin_servers/mp_rpc.lua
```diff
@@ -325,11 +328,17 @@ end
...
+  if not err then
+    _, err = bridge_loop(self, instance_id, phase)
+  end
 
   if err then
-    if string.match(err:lower(), "no plugin instance") then
+    if err == "not ready" then
+      return handle_not_ready(plugin_name)
+    end
+    if str_find(err:lower(), "no plugin instance") then
```
chronolaw, line 341: use `str_find(x, y, 1, true)` to do a plain search to speed up.

kong/runloop/plugin_servers/pb_rpc.lua
```diff
@@ -386,13 +387,19 @@ end
...
 
-  local res, err = self:call("cmd_handle_event", {
-    instance_id = instance_id,
-    event_name = phase,
-  }, true)
   if not res or res == "" then
+    if err == "not ready" then
+      return handle_not_ready(plugin_name)
+    end
     if string.match(err:lower(), "no plugin instance")
```
chronolaw, line 403: Same change, use `string.find(x, y, 1, true)`

kong/runloop/plugin_servers/pb_rpc.lua
```diff
@@ -386,13 +387,19 @@ end
...
       or string.match(err:lower(), "closed")  then
```
chronolaw, line 404: string.find

kong/runloop/plugin_servers/process.lua
```diff
@@ -1,12 +1,19 @@
 local cjson = require "cjson.safe"
 local pl_path = require "pl.path"
 local raw_log = require "ngx.errlog".raw_log
+local ngx = ngx
+local sleep = ngx.sleep
+local connect = ngx.socket.connect
+local is_http_subsystem = ngx.config.subsystem ~= "http"
```
chronolaw, line 7: should be `ngx.config.subsystem == "http"`

kong/runloop/plugin_servers/init.lua
```diff
@@ -329,16 +340,12 @@ end
 
 
 function plugin_servers.start()
-  if worker_id() ~= 0 then
-    kong.log.notice("only worker #0 can manage")
-    return
-  end
-
-  local pluginserver_timer = proc_mgmt.pluginserver_timer
+  local handler = worker_id() == 0 and proc_mgmt.pluginserver_timer
```
fffonion, line 343: let's create a new function, like `plugin_servers.check()`, that runs on all worker, and only call `plugin_servers.start()` as worker 0. this will make code more straight forward.

kong/runloop/plugin_servers/process.lua
```diff
@@ -235,8 +276,20 @@ function proc_mgmt.pluginserver_timer(premature, server_def)
...
+-- limit the number of warning messages
+local plugin_already_warned = {}
+function proc_mgmt.handle_not_ready(plugin_name)
+  if plugin_already_warned[plugin_name] then
+    return
+  end
+
+  plugin_already_warned[plugin_name] = true
+  log(ngx_WARN, "plugin server is not ready when ",
+      plugin_name, " is required for request")
```
fffonion, line 291:
```suggestion
  log(ngx_WARN, "plugin server ", plugin_name,
      " is not started, plugin on this plugin server will",
      " not be executed")
```

StarlightIbuki: `plugin_name` is plugin's name, not the plugin server's name

kong/init.lua
```diff
@@ -802,7 +802,10 @@ function Kong.init_worker()
   runloop.init_worker.after()
 
   if is_not_control_plane then
-    plugin_servers.start()
+    if worker_id() == 0 then
```
chronolaw, line 805: I think we should put `worker_id() == 0` condition in `plugin_servers/init.lua`.

Because there is no worker id check in `exit_woker`, now it is a bit asymmetry.

The code could be 
```
    plugin_servers.start()    -- check worker id in the function
    plugin_servers.check()
```

StarlightIbuki: A function named `start` but it does not start anything in some conditions...doesn't seem right.

chronolaw: The original code of `start` has that logic: 
check worker id and if it is not 0 then exit.

fffonion: my previous comment is not clear, we should do this:
in kong/init.lua call `plugin_server.start`
in plugin_server/init.lua `start` function:

```lua
function plugin_servers.start()
  if worker id == 0 then
    start server
  end

  start check timer
end
```

chronolaw: Yes, I think this is the right way.

kong/runloop/plugin_servers/init.lua
```diff
@@ -352,15 +362,21 @@ end
...
 
 function plugin_servers.start()
-  if worker_id() ~= 0 then
-    return
+  if worker_id() == 0 then
+    local pluginserver_timer = proc_mgmt.pluginserver_timer
+
+    for _, server_def in ipairs(proc_mgmt.get_server_defs()) do
+      if server_def.start_command then
+        native_timer_at(0, pluginserver_timer, server_def)
```
chronolaw, line 370: There is already a `timer.at` call in `proc_mgmt.pluginserver_timer`,
Is something wrong?

fffonion (approved):

PR #9517, opened 2022-10-10
chore(router/atc): remove last dot in sni

chronolaw:
### Summary

Add the same logic in compat router: last dot in FQDNs must not be used for routing


fffonion:
Is it better to handle this logic in the rust code instead of patching in lua land?

dndx:
@fffonion No, the Rust code has no concept between a path and a SNI, everything is treated as plain strings without any special meanings.

dndx (approved):

PR #9518, opened 2022-10-10
style(router/atc): check fields with `is_empty_field` when we calc priority

chronolaw:
### Summary

Fields in route (paths/hosts) may be `nil` or `ngx.null`, so we use `is_empty_field` to check them more carefully.

### DIFF
--- a/kong/router/compat.lua
+++ b/kong/router/compat.lua
@@ -211,10 +211,6 @@ local function get_priority(route)
     match_weight = match_weight + 1
   end
 
-  if not is_empty_field(hosts) then
-    match_weight = match_weight + 1
-  end
-
   if not is_empty_field(paths) then
     match_weight = match_weight + 1
   end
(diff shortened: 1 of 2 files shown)


bungle (approved):

kong/router/compat.lua
```diff
@@ -235,9 +227,11 @@ local function get_priority(route)
     match_weight = match_weight + 1
   end
 
-  local plain_host_only = not not hosts
+  local plain_host_only = not is_empty_field(hosts)
```
dndx, line 230: This change might not be correct, because it treats `hosts = {}` as not having `plain_host_only`. This contradicts with the traditional logic:

https://github.com/Kong/kong/blob/master/kong/router/traditional.lua#L350-L352

which considers `hosts = {}` as `plain_host_only = true`, but `hosts = nil` as `plain_host_only = false`.

chronolaw: OK, I will check and fix it.

chronolaw: Now the code change to `type(hosts) == "table"`,
I think it is good enough to fit `hosts = {}/nil/ngx.null`.

dndx (approved):

PR #9519, opened 2022-10-10
chore(router/atc): cache split host port results

chronolaw:
### Summary

Cache host-port split results to speed up router building and matching.

This logic is a mimic of traditional flavor router.

### DIFF
--- a/kong/router/atc.lua
+++ b/kong/router/atc.lua
@@ -18,7 +18,6 @@ local assert = assert
 local setmetatable = setmetatable
 local pairs = pairs
 local ipairs = ipairs
-local tonumber = tonumber
 
 
 local max = math.max
@@ -324,6 +323,8 @@ end
 -- example.*:123 => example.*, 123
 local split_host_port
 do
+  local tonumber = tonumber
+
   local memo_h = setmetatable({}, { __mode = "k" })
   local memo_p = setmetatable({}, { __mode = "k" })
 
--- a/kong/router/atc.lua
+++ b/kong/router/atc.lua
@@ -324,38 +324,39 @@ end
 local split_host_port
 do
   local tonumber = tonumber
+  local DEFAULT_HOSTS_LRUCACHE_SIZE = DEFAULT_MATCH_LRUCACHE_SIZE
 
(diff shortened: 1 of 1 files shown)


kong/router/atc.lua
```diff
@@ -318,25 +317,44 @@ end
...
+local split_host_port
+do
+  local tonumber = tonumber
 
-  local p = h:find(":", nil, true)
-  if not p then
-    return h, nil
-  end
+  local memo_h = setmetatable({}, { __mode = "k" })
+  local memo_p = setmetatable({}, { __mode = "k" })
```
dndx, line 325: I know this code come from the old router, but it seems to have some issues:

By making the keys weak, the table's value are dependent on temporary objects created during requests that are gone by the end of the request. This means that these table are probably not caching enough content as they should.

Further, the logic below seems to assume that if `memo_h[h]` is not `nil`, then `memo_p[h]` must also contains a valid value. This may not be true if Lua GC'ed `memo_p[h]` before GCing `memo_h[h]`.

I think we should change this (and the traditional router) to use proper LRU Cache instead of this pattern. Otherwise in rare cases it may produce very hard to debug issues.

dndx (changes requested):

kong/router/atc.lua
chronolaw: It is a reasonable concern. I will change weak table to lru cache.

kong/router/atc.lua
```diff
@@ -322,25 +321,45 @@ end
...
+  local tonumber = tonumber
+  local DEFAULT_HOSTS_LRUCACHE_SIZE = DEFAULT_MATCH_LRUCACHE_SIZE
 
-  local p = h:find(":", nil, true)
-  if not p then
-    return h, nil
-  end
+  local memo_hp = lrucache.new(DEFAULT_HOSTS_LRUCACHE_SIZE)
+
+  split_host_port = function(h)
```
dndx, line 331:
```suggestion
  split_host_port = function(key)
```

Doesn't seems that we need another variable of `key` below.

chronolaw: done.

dndx (approved):

PR #9520, opened 2022-10-10
tests(router/atc): add test cases for nil/ngx.null/empty table

chronolaw:
### Summary

add test cases for nil/ngx.null/empty table.


chronolaw:
This PR is merged into #9518.

PR #9526, opened 2022-10-11
fix(pdk): response.exit accept bytes

### LINKED ISSUE #9294: ExitArgs.Body should be []byte instead of string

2.8.1

### Current Behavior

To handle gzipped body content with go-pdk, an encoding error occurs like "string field contains invalid UTF-8 ".

### Expected Behavior

To handle any arbitrary sequence of bytes, the type of `ExitArgs.body` should be bytes instead of string according to https://developers.google.com/protocol-buffers/docs/proto3

https://github.com/Kong/kong/blob/4d4b90c298f0c4dae5c271f3bd0dc2b117f7614b/kong/pluginsocket.proto#L104

### Steps To Reproduce

_No response_

### Anything else?

Without this, PDK is less useful since we can not handle upstream's compressed response.

StarlightIbuki:
fix #9294

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,13 @@
   (`kong.request.getUriCaptures`)
   [#9512](https://github.com/Kong/kong/pull/9512)
 
+- Fixed parameter type of `kong.service.request.set_raw_body`
+(`kong.service.request.setRawBody`), return type of
+`kong.service.response.get_raw_body`(`kong.service.request.getRawBody`),
+and body parameter type of `kong.response.exit` to bytes. Note that old
+version of go PDK is incompatible after this change.
+  [#9526](https://github.com/Kong/kong/pull/9526)
+
 #### Plugins
 
 - **AWS Lambda**: Fix an issue that is causing inability to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,10 +120,10 @@
   [#9512](https://github.com/Kong/kong/pull/9512)
 
 - Fixed parameter type of `kong.service.request.set_raw_body`
-(`kong.service.request.setRawBody`), return type of
-`kong.service.response.get_raw_body`(`kong.service.request.getRawBody`),
-and body parameter type of `kong.response.exit` to bytes. Note that old
-version of go PDK is incompatible after this change.
+  (`kong.service.request.setRawBody`), return type of
+  `kong.service.response.get_raw_body`(`kong.service.request.getRawBody`),
+  and body parameter type of `kong.response.exit` to bytes. Note that old
+  version of go PDK is incompatible after this change.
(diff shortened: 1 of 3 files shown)


fffonion:
as we are doing this, could you also update for SetRawBody and GetRawBody as well?

kong/include/kong/pluginsocket.proto
```diff
@@ -101,7 +101,7 @@ message String {
 
 message ExitArgs {
     int32 status = 1;
-    string body = 2;
+    bytes body = 2;
     google.protobuf.Struct headers = 3;
```
fffonion, line 105: for backward compatibilty, we might need to add a new field, and rename the old field but keep same id 🤔 ?

chronolaw:
Do not forget to add change log entry.

kong/include/kong/pluginsocket.proto
StarlightIbuki: 🤦 We are using `String`, which is defined below, as many RPC's parameters and return types:
```protobuf
message String {
    string v = 1;
}
```
It's not easy to keep backward compatibility for this.

I come up with a hack that we could add a type like:
```protobuf
message ByteString {
    string v_lagacy = 1;
    bytes v = 2;
}
```
and we use that as the type for `get/setRawBody`. The older version of GO PDK would treat the type as String, and they are compatible.

StarlightIbuki:
Technically, `Ctx_GetShared` can return raw binary strings, and `Ctx_SetShared` can accept raw binary strings. But we are ignoring them for now.

kong/include/kong/pluginsocket.proto
StarlightIbuki: Finally we decided not to have these compatibility.

StarlightIbuki:
Maybe we should be more specific about the version that will be incompatible, but we need to merge the PR https://github.com/Kong/go-pdk/pull/132 first, and that PR is waiting for this PR to merge...

kong/include/kong/pluginsocket.proto
```diff
@@ -99,9 +99,13 @@ message String {
     string v = 1;
 }
 
+message ByteString {
+    bytes v = 2;
```
fffonion, line 103:
```suggestion
    bytes v = 1;
```

CHANGELOG.md
```diff
@@ -119,6 +119,13 @@
   (`kong.request.getUriCaptures`)
   [#9512](https://github.com/Kong/kong/pull/9512)
 
+- Fixed parameter type of `kong.service.request.set_raw_body`
+(`kong.service.request.setRawBody`), return type of
```
chronolaw, line 123: Should align to `Fixed ...`

fffonion (approved):

PR #9533, opened 2022-10-11
feat(hash): replace md5 with sha256 on proxy-cache plugin

samugi:
Replace md5 hashing algorithm with sha256 on the proxy-cache plugin
Add utils functions for sha256 hashing

[FT-2418](https://konghq.atlassian.net/browse/FT-2418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

### DIFF
--- a/kong/plugins/proxy-cache/cache_key.lua
+++ b/kong/plugins/proxy-cache/cache_key.lua
@@ -1,12 +1,12 @@
 local fmt = string.format
 local ipairs = ipairs
-local md5 = ngx.md5
 local type = type
 local pairs = pairs
 local sort = table.sort
 local insert = table.insert
 local concat = table.concat
 
+local sha256_hex = require "kong.tools.utils".sha256_hex
 
 local _M = {}
 
@@ -108,8 +108,8 @@ function _M.build_cache_key(consumer_id, route_id, method, uri,
   local params_digest  = params_key(params_table, conf)
   local headers_digest = headers_key(headers_table, conf)
 
-  return md5(fmt("%s|%s|%s|%s|%s", prefix_digest, method, uri, params_digest,
-                                   headers_digest))
+  return sha256_hex(fmt("%s|%s|%s|%s|%s", prefix_digest, method, uri,
+                                          params_digest, headers_digest))
 end
 
 
--- a/kong/tools/utils.lua
+++ b/kong/tools/utils.lua
@@ -1468,4 +1468,66 @@ do
 end
 _M.time_ns = time_ns
 
+
+local sha256_bin
+do
+  local digest = require "resty.openssl.digest"
+  local sha256_digest
+
+  function sha256_bin(key)
+    local _, bin, err
+    if not sha256_digest then
+      sha256_digest, err = digest.new("sha256")
+      if err then
+        return nil, err
+      end
+    end
+
+    bin, err = sha256_digest:final(key)
+    if err then
+      return nil, err
+    end
+
+    _, err = sha256_digest:reset()
+    if err then
+      sha256_digest = nil
+    end
+
+    return bin
+  end
+end
+_M.sha256_bin = sha256_bin
+
+
+local sha256_hex, sha256_base64, sha256_base64url
+do
+  local to_hex       = require "resty.string".to_hex
+  local to_base64    = ngx.encode_base64
+  local to_base64url = require "ngx.base64".encode_base64url
+
+  local function sha256_encode(encode_alg, key)
+    local bin, err = sha256_bin(key)
+    if err then
+      return nil, err
+    end
+
+    return encode_alg(bin)
+  end
+
+  function sha256_hex(key)
+    return sha256_encode(to_hex, key)
+  end
+
+  function sha256_base64(key)
+    return sha256_encode(to_base64, key)
+  end
+
+  function sha256_base64url(key)
+    return sha256_encode(to_base64url, key)
+  end
+end
+_M.sha256_hex       = sha256_hex
+_M.sha256_base64    = sha256_base64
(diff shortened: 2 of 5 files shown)


kong/plugins/proxy-cache/cache_key.lua
```diff
@@ -108,8 +109,8 @@ function _M.build_cache_key(consumer_id, route_id, method, uri,
   local params_digest  = params_key(params_table, conf)
   local headers_digest = headers_key(headers_table, conf)
 
-  return md5(fmt("%s|%s|%s|%s|%s", prefix_digest, method, uri, params_digest,
-                                   headers_digest))
+  return sha256_hex(fmt("%s|%s|%s|%s|%s", prefix_digest, method, uri,
+                          params_digest, headers_digest))
```
fffonion, line 113: i did a quick perf test since sha256 is known slower than md5; for 100000 iterrations:
sha256 took: 0.042999982833862
md5 took: 0.016000032424927

it doesn't seem too terrible though. on the side, i'm thinking if we can make this function same as the `fips_hash` function that only choose sha256 for fips mode? @samugi  how do you think?

samugi: I should have added some background on this choice, thanks for the reminder @fffonion.

@bungle and I talked about this yesterday and based on our tests it seems md5 would not provide any performance benefits over sha256, at least on our machines. The test results were actually showing better numbers for sha256 for both of us, with a ~30% improvement over md5 for me and a ~50% improvement for @bungle. 

We used `resty.openssl.digest` with `md5` and `sha256` algorithms, on 10.000.000 iterations, the results are below:

```
@bungle:
md5: 1400ms
sha256: 658ms

mine:
md5: 961ms
sha256: 694ms
```

fffonion: Discussed with @samugi offline, i agree performance is no longer a concern because of the CPU accelaration for sha256. Though from my test ngx.md5_bin is still faster than openssl.md5 and openssl.sha256 since Nginx uses its own md5 implementaion, but that may indicate something can be improved in lua-resty-openssl.
Consider this resolved.

fffonion (approved):

kong/tools/utils.lua
```diff
@@ -1468,4 +1468,57 @@ do
...
 
+
+local sha256
+do
+  local digest = require "resty.openssl.digest"
+  local sha256_digest
+
+  function sha256(key)
+    sha256_digest = sha256_digest or assert(digest.new("sha256"))
+    local bin     = assert(sha256_digest:final(key))
```
chronolaw, line 1479: `final` may fail and return nil and an error, should we check it?

bungle (approved):

kong/tools/utils.lua
```diff
@@ -1468,4 +1468,82 @@ do
 end
 _M.time_ns = time_ns
 
+
+local sha256
```
chronolaw, line 1472: I think the name should be `sha256_bin`, just like ngx_lua's `md5_bin`

kong/tools/utils.lua
```diff
@@ -1468,4 +1468,82 @@ do
...
+      sha256_digest = nil
+    end
+
+    return bin
+  end
+end
+_M.sha256 = sha256
+
+
+local sha256_hex
```
chronolaw, line 1502: `sha256_hex` `sha256_base64` `sha256_base64url` have the almost same pattern.

I think that we can define a common function, then pass the encode function into it to reduce duplicated code.

kong/plugins/proxy-cache/cache_key.lua
chronolaw, line 113: style: should have the same alignment as before.

kong/plugins/proxy-cache/cache_key.lua
```diff
@@ -1,12 +1,13 @@
 local fmt = string.format
 local ipairs = ipairs
-local md5 = ngx.md5
 local type = type
 local pairs = pairs
 local sort = table.sort
 local insert = table.insert
 local concat = table.concat
 
+local sha256_hex = require "kong.tools.utils".sha256_hex
```
chronolaw, line 9: Here we can define a more common name like `digest_func` to cover the changes of md5 and sha256.

kong/tools/utils.lua
bungle: i think that was a mistake by ngx.

kong/plugins/proxy-cache/cache_key.lua
samugi: this name is intentionally specific to ensure the implementation (hash algorithm used) can be more easily recognized through the code

kong/tools/utils.lua
samugi: done, thanks

kong/plugins/proxy-cache/cache_key.lua
samugi: done, thank you

chronolaw (approved):

kong/tools/utils.lua
```diff
@@ -1468,4 +1468,66 @@ do
...
+    local _, bin, err
+    if not sha256_digest then
+      sha256_digest, err = digest.new("sha256")
+      if err then
+        return nil, err
+      end
+    end
+
+    bin, err = sha256_digest:final(key)
+    if err then
```
chronolaw, line 1487: should we add a `reset` here?

samugi: yes I guess it would be a good idea to start fresh in case of errors

PR #9536, opened 2022-10-11
fix(cli): shut down timers at the end of execution

flrgh:
### Summary

With timer-ng, the CLI has become slow to terminate as nginx attempts to wait for all pending timers to stop before exiting:

```
$ time kong roar >/dev/null

real    0m1.033s
user    0m0.024s
sys     0m0.016s
```

This commit explicitly shuts down the timer-ng instance so that the CLI can quickly exit:

```
$ time kong roar >/dev/null

real    0m0.132s
user    0m0.027s
sys     0m0.013s
```

### Full changelog

* explicitly stop timer-ng instance on completion of CLI commands

### DIFF
--- a/kong/cmd/init.lua
+++ b/kong/cmd/init.lua
@@ -5,12 +5,10 @@ math.randomseed() -- Generate PRNG seed
 local pl_app = require "pl.lapp"
 local log = require "kong.cmd.utils.log"
 
-local timerng = _G.timerng
-
 local function stop_timers()
   -- shutdown lua-resty-timer-ng to allow the nginx worker to stop quickly
-  if timerng then
-    timerng:destroy()
+  if _G.timerng then
+    _G.timerng:destroy()
   end
 end
 
--- a/kong/cmd/init.lua
+++ b/kong/cmd/init.lua
@@ -5,6 +5,15 @@ math.randomseed() -- Generate PRNG seed
 local pl_app = require "pl.lapp"
 local log = require "kong.cmd.utils.log"
 
+local timerng = _G.timerng
+
+local function stop_timers()
+  -- shutdown lua-resty-timer-ng to allow the nginx worker to stop quickly
+  if timerng then
+    timerng:destroy()
+  end
+end
+
 local options = [[
  --v              verbose
  --vv             debug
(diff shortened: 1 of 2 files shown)


bungle (approved):

bungle:
@ADD-SP can you confirm this is alright?

ADD-SP (approved):

kong/cmd/init.lua
```diff
@@ -5,6 +5,15 @@ math.randomseed() -- Generate PRNG seed
 local pl_app = require "pl.lapp"
 local log = require "kong.cmd.utils.log"
 
+local timerng = _G.timerng
```
chronolaw, line 8: Could we move this line into function `stop_timers`?

chronolaw (approved):

Tieske:
wondering; shouldn't this be fixed in timer-ng instead?

flrgh:
> wondering; shouldn't this be fixed in timer-ng instead?

I think timer-ng is mostly behaving as expected here. I believe the 1s shutdown delay observed currently is due to a `semaphore:wait(1)` call somewhere (as controlled by [this constant](https://github.com/Kong/lua-resty-timer-ng/blob/6b90abd0e55b5e71ba6e12cd9c9a2c842fdfa5db/lib/resty/timerng/constants.lua#L38)). Delaying worker shutdown by as much as 1s seems totally fine in the regular nginx context--it's just that the CLI is a whole different beast.

dndx (approved):

Tieske:
> > wondering; shouldn't this be fixed in timer-ng instead?
> 
> I think timer-ng is mostly behaving as expected here. I believe the 1s shutdown delay observed currently is due to a `semaphore:wait(1)` call somewhere (as controlled by [this constant](https://github.com/Kong/lua-resty-timer-ng/blob/6b90abd0e55b5e71ba6e12cd9c9a2c842fdfa5db/lib/resty/timerng/constants.lua#L38)). Delaying worker shutdown by as much as 1s seems totally fine in the regular nginx context--it's just that the CLI is a whole different beast.

this is an old problem caused by a lack of IO in the nginx control loop, workarounds are available. See https://github.com/openresty/lua-resty-core/issues/337 That said; from that issue I see that I already pointed that out before...

PR #9487, opened 2022-09-30
chore(ci): run upgrade tests in PR, not daily

hanshuebner:
As we've tested the upgrade tests for a good while now, we can make them run in Pull Requests to ensure that new migrations also have a corresponding test.

PR #9542, opened 2022-10-12
tests(migrations): add 3.1 migrations spec

attenuation:
### Summary
add 3.1 migrations spec

### Issue reference
add 3.1 migrations spec for #9430 
chery-pick #9487


spec/05-migration/db/migrations/core/017_300_to_310_spec.lua
```diff
@@ -0,0 +1,7 @@
+local uh = require "spec/upgrade_helpers"
+
+describe("database migration", function()
```
chronolaw, line 3: Why didn't we use `it(...)`?

PR #9543, opened 2022-10-12
Revert "chore(deps): bump openssl from 1.1.1q to 1.1.1r (#9535)"

catbro666:
This reverts commit 17fa51ca0bd6546adb3f42038bbb4e56a8f6fbff.

### Summary

OpenSSL withdrew the 1.1.1r, I think we should revert this PR.

https://www.mail-archive.com/openssl-announce@openssl.org/msg00386.html


bungle:
@catbro666 thanks for bringing this up. I think we should revert it.

bungle (approved):

PR #9544, opened 2022-10-13
feat(ci): add spelling check to github workflow

vm-001:
Summary:

Add the codespell tool to check spelling mistakes.


test.txt
```diff
@@ -0,0 +1,3 @@
+maxinum
```
chronolaw, line 1: Should we remove this test file？

vm-001: Yeah. That's only for demonstration.

.ignore_lines
```diff
@@ -0,0 +1,2 @@
+            pdk.service.request.set_path("/fo%6F")
```
chronolaw, line 1: Does this file have a special meaning? Or is it a test/demo?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            echo "$file was changed"
+          done
+
+      - name: Installing codespell
+        run: |
+          pip install codespell==2.2.1
+
+#      # lint all files (test purpose)
```
chronolaw, line 27: I think we could remove these test code.

vm-001: Actually, these are not test codes. I prefer to keep this as we will switch to this someday as no more old existing misspellings. (Don't have to figure it out again)

chronolaw: I saw you have removed these code. 

It makes our script more clean, and we can still keep that test code for future purposes.

vm-001:
I feel like we should move the lint files into one directory instead of letting these files sit in our root directory.

vm-001:
The test case works, https://github.com/Kong/kong/actions/runs/3262166095/jobs/5358717557.

chronolaw (approved):

.github/workflows/lint.yml
```diff
@@ -0,0 +1,31 @@
+name: Lint
```
dndx, line 1: Let's give this workflow a different name. How about `Spell Check`? "Lint" seems to be confusing with our code style linter.

fffonion (approved):

fffonion:
merge after CI pass

chronolaw:
`.github/codespell/.dictionary.txt` is empty, do we need this?

chronolaw:
`.github/workflows/lint.yml` should also be renamed, like `spellcheck`.

chronolaw:
This PR has so many changed source files.

I think could we split this PR, one for spell check workflow, and the other for spell correction.

.github/codespell/README.md
```diff
@@ -0,0 +1,14 @@
+
+### .dictionary.txt
```
mayocream, line 2: I don't think this README is needed.

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,19 @@
+name: Spell Check
```
mayocream, line 1: I personally don't think it is required to make it a GitHub Action and make it always run on Kong's CI as a gatekeeper; incorrect "considered misspellings" will result in the CI pipeline status to **red 🔴**, and developers have to deal with `.ignore_words.txt` and `.ignore_lines.txt`, which is not supposed to be their responsibility (to fight with bad tooling).

mayocream (changes requested):

.github/workflows/spellcheck.yml
vm-001: The reason why this tool works for Kong repo is that it has a ton of misspelled words in the dictionaries which can be confidently considered it's wrong. (compared with Vale that the docs repo is using)

And I personally don't think there was a good tooling that can be used without adding any "ignored/exclude" thing for the programming repo.

I really don't think adding a few ignored terms/lines costs any pains compared with the good side it brings to this repo.

mayocream: From my perspective, it makes sense to add a script under scripts/ folder to check spelling mistakes for convenience. However, it shouldn't be a blocker for any PR. 

**I am not convinced that I didn't see any popular open source project use such a tool running against every PR.**

hbagdi: Mayo, do you think it is valuable to catch spelling mistakes at the PR stage and do so in an automated manner rather than relying on maintainers to check?

mayocream: @hbagdi It’s very common that a spellcheck tool reports a false-positive warning, then developers need to learn how to write .ignore rules. I don’t think it is that much valuable for identifying just spelling mistakes. More effort is needed to maintain the tools.

mayocream: It’s a manner that using a spellcheck tool for Markdown docs, our docs.kongkq.com also use Vale to check spelling errors. But I don’t think it’s necessary to enabe it for all code files.

kikito (changes requested):
We had a chat about this feature in PR Queue Review.

The overall result is: the team is not happy with making the build go RED because of a perceived spelling check. We feel there will be too many false positives and this will add too much friction.

Suggestion: Split this PR into two. One which contains just the spelling fixes (that is easily mergeable). The other one which makes it possible to run it, on github CI, but it is not part of the CI. That way we can make a spellcheck pass every release or something along those lines.

flrgh:
I have one other request to raise here:

Whatever the final version of this change looks like, please open up a companion PR against Kong Enterprise to cherry-pick the changes. The reason for this is twofold:

1. A commit that changes many files greatly increases the opportunity for merge conflicts when resolving OSS and EE diffs
2. A commit that introduces a new test may pass in OSS but fail in EE due to the many differences between the two codebases

Therefore, for changes like this one, it is recommended that the commit author "owns" the merge process for both OSS and Enterprise.

vm-001:
> We had a chat about this feature in PR Queue Review.
> 
> The overall result is: the team is not happy with making the build go RED because of a perceived spelling check. We feel there will be too many false positives and this will add too much friction.
> 
> Suggestion: Split this PR into two. One which contains just the spelling fixes (that is easily mergeable). The other one which makes it possible to run it, on github CI, but it is not part of the CI. That way we can make a spellcheck pass every release or something along those lines.

To be honest. Fixing harmless typos is not the major purpose of this PR. The original purpose is to block adding more typos to this repo(That's why I think running a spelling lint tool against every PR is a good way to do that, and this is the way that the docs repo was currently doing. And also why it only lint the changed file in the very begging). I think depending on human eyes to correct the typos is not a good idea. It increases the unnecessary extra energy and round-trip time to get PR merged. But the most important thing is, humans make mistakes all the time. 

If the evaluation result for the original purpose of this PR is negative, I'm OK with closing this PR.

In the end, no matter which way or tool this repo decides. Keeping fixed typos is the thing we all agree on.

PR #9546, opened 2022-10-13
fix(PDK): req.start_time for PDK log phase

### LINKED ISSUE #9409: kong.nginx.req_start_time() returns zero in log phase

2.8.1

### Current Behavior

The code in `init.lua` simply sets the plugin api `kong.nginx.req_start_time` to `ngx.req.start_time`, but the log phase was delay-executed after the request has finished, so I think the value of `ngx.req.start_time()` should be persisted in `save_for_later[co]` to make it work.

And also some other bugs could be the same reason, for example, when calling `kong.request.get_host()`, the process `check_phase` produces an error message, due to the `ngx.ctx.KONG_PHASE` is empty.

### Expected Behavior

`kong.nginx.req_start_time()` returns right value in `log` phase

### Steps To Reproduce

```markdown
1. Run kong-gateway:2.8.1
2. Setup a plugin with a `log` phase
3. call `kong.nginx.req_start_time()`
```

### Anything else?

_No response_

StarlightIbuki:
Fix #9409

### DIFF
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -110,7 +110,7 @@ local exposed_api = {
 
   ["kong.nginx.req_start_time"] = function ()
     local saved = save_for_later[coroutine_running()]
-    return saved.req_start_time
+    return saved and saved.req_start_time or ngx.req.start_time()
   end,
 }
 
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -108,7 +108,10 @@ local exposed_api = {
     return kong.response.get_source(saved and saved.ngx_ctx or nil)
   end,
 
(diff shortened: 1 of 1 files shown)


fffonion (approved):

chronolaw (approved):

chronolaw:
Need a change log entry.

kong/runloop/plugin_servers/init.lua
```diff
@@ -108,7 +108,10 @@ local exposed_api = {
     return kong.response.get_source(saved and saved.ngx_ctx or nil)
   end,
 
-  ["kong.nginx.req_start_time"] = ngx.req.start_time,
+  ["kong.nginx.req_start_time"] = function ()
```
chronolaw, line 111: remove redundant space before `()`

bungle, line 111:
```suggestion
  ["kong.nginx.req_start_time"] = function()
```

PR #9547, opened 2022-10-13
style(plugin_servers): some localization of functions

chronolaw:
### Summary

I found that lots of code of plugins did not localize global apis, 
so I did a try, Does it make sense?

### Full changelog

* localize some API like `pairs` `type`
* introduce a function `get_saved()` to reduce duplicated code.

### DIFF
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -33,11 +33,15 @@ local rpc_notifications = {}
 --- currently running plugin instances
 local running_instances = {}
 
+local function get_saved()
+  return save_for_later[coroutine_running()]
+end
+
 local exposed_api = {
   kong = kong,
 
   ["kong.log.serialize"] = function()
-    local saved = save_for_later[coroutine_running()]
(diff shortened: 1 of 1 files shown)


kong/runloop/plugin_servers/init.lua
```diff
@@ -153,7 +168,7 @@ function get_instance_id(plugin_name, conf)
 
   while instance_info and not instance_info.id do
     -- some other thread is already starting an instance
-    ngx.sleep(0)
+    ngx_sleep(0)
```
chronolaw, line 171: Could we use `native_ngx_sleep` here?

See PR #9523

bungle (approved):

kong/runloop/plugin_servers/init.lua
bungle: What about using `kong.tools.utils.yield`?

chronolaw: I am not familiar with the plugin server, not sure about the logic of it, let's do it later.

PR #9551, opened 2022-10-14
style(plugin_server): small code clean for plugin server 

chronolaw:
### Summary

Due to PR #9342 #9546 #9547, 
We should do some code clean to fit these changes.

### Full changelog

* use `get_saved()` in  `kong.nginx.req_start_time`
* check `ngx.config.subsystem == "http"` in module scope
* remove 'log.notice' when `worker_id() ~= 0`
* remove check of worker id in `init.lua`, because plugin server did the same thing.
* if it is http subsystem then we localize `ngx.req.get_headers` `ngx.resp.get_headers`
* if it is stream subsystem then `req_get_headers` will do nothing

### DIFF
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -335,7 +335,6 @@ end
 
 function plugin_servers.start()
   if worker_id() ~= 0 then
-    kong.log.notice("only worker #0 can manage")
     return
   end
 
@@ -350,7 +349,6 @@ end
 
 function plugin_servers.stop()
   if worker_id() ~= 0 then
-    kong.log.notice("only worker #0 can manage")
     return
   end
 
--- a/kong/runloop/plugin_servers/init.lua
+++ b/kong/runloop/plugin_servers/init.lua
@@ -15,9 +15,6 @@ local kong = kong
 local ngx_var = ngx.var
 local ngx_sleep = ngx.sleep
 local worker_id = ngx.worker.id
-local req_get_headers = ngx.req.get_headers
-local resp_get_headers = ngx.resp.get_headers
-local req_start_time = ngx.req.start_time
 
 local coroutine_running = coroutine.running
 local get_plugin_info = proc_mgmt.get_plugin_info
@@ -130,7 +127,7 @@ local exposed_api = {
 
   ["kong.nginx.req_start_time"] = function()
     local saved = get_saved()
-    return saved and saved.req_start_time or req_start_time()
+    return saved and saved.req_start_time or ngx.req.start_time()
   end,
 }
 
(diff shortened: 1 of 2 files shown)


chronolaw:
A weird thing: 
I try to localize `ngx.resp.get_headers`, but get a `resp is nil` error when loading this module.

Could someone give me an explain?

ADD-SP:
> A weird thing: I try to localize `ngx.resp.get_headers`, but get a `resp is nil` error when loading this module.
> 
> Could someone give me an explain?
> 
> See: https://github.com/Kong/kong/actions/runs/3248026332/jobs/5328759006

@chronolaw 

Not all functions can be localized, such as `ngx.resp.get_headers`. We will load almost all modules in the `init` and `init_worker` phases, but some functions are not available in their phase.

https://github.com/openresty/lua-nginx-module/blob/a318d250f547c854ea2b091d0e06372ac0c00fd5/src/ngx_http_lua_util.h#L309

This function is called in most phases, except for the `init` and `init_worker` phases, so there is no `ngx.resp.get_headers`.

kong/runloop/plugin_servers/init.lua
```diff
@@ -15,14 +15,25 @@ local kong = kong
...
 local cjson_encode = cjson.encode
 local native_timer_at = _G.native_timer_at or ngx.timer.at
 
+local req_get_headers
+local resp_get_headers
+
+if is_http then
+  req_get_headers  = ngx.req.get_headers
+  resp_get_headers = ngx.resp.get_headers
+end
```
ADD-SP, line 35: https://github.com/openresty/lua-nginx-module#ngxreqget_headers
https://github.com/openresty/lua-nginx-module#ngxrespget_headers

These two functions are not available in the `init` phase, I think we can't localize them.

kong/runloop/plugin_servers/init.lua
```diff
@@ -15,14 +15,25 @@ local kong = kong
 local ngx_var = ngx.var
 local ngx_sleep = ngx.sleep
 local worker_id = ngx.worker.id
+local req_start_time = ngx.req.start_time
```
ADD-SP, line 18: https://github.com/openresty/lua-nginx-module#ngxreqstart_time

This function is not available in the `init` phase, I think we can't localize them.

ADD-SP (changes requested):
https://github.com/Kong/kong/blob/210f9e9bb1e35b9b469ae036b4feaafd6f0fbeb6/kong/init.lua#L86

The `plugin_servers` module is loaded in the `init` phase, not all API are available.

kong/runloop/plugin_servers/init.lua
chronolaw: I think you may misunderstand the meaning of `context`.
The context of ngx_lua api means it is available and should be called in these phases.

If we read the source of ngx_lua and lua-resty-core,
we will find that these apis have been loaded into lua land.

kong/runloop/plugin_servers/init.lua
chronolaw: I created a small nginx config file, wrote some code in `init_by_lua_block`:

```
    init_by_lua_block {
        local x = ngx.req.get_headers
        ngx.log(ngx.ERR, "type = ", type(x))
    }
```

It works. So I think the reason is that the module is loaded in stream subsystem,
and there is no `ngx.resp.xxx`.

ADD-SP (approved):

dndx (approved):

PR #9554, opened 2022-10-14
fix(url): adjust the pulp admin api links

hutchic:
The previous domain no longer exist.


hutchic:
this was the wrong release branch

PR #9561, opened 2022-10-16
feat(templates): configuration of unix listening socket path

chronolaw:
### Summary

Refer to #9337, and add a special configuration `unix_socket_path`,
which will not expose to the normal user.

See FT-3468.

### Full changelog

* add `unix_socket_path` in `conf_loader/init.lua`
* port some code from #9337, change the path of resty-events
* change nginx kong templates


kong/conf_loader/init.lua
```diff
@@ -1861,6 +1861,13 @@ local function load(path, custom_conf, opts)
 
   log.verbose("prefix in use: %s", conf.prefix)
 
+  -- set unix domain socket file absolute path
+  local unix_socket_path = getenv("KONG_UNIX_SOCKET_PATH")
```
dndx, line 1865: Why is this needed?

chronolaw: See the discussion in PR #9337, we can customize the unix socket path by env.

dndx: I am not convinced that this is a good idea, and I will block this PR until the concern is clearly addressed (see: https://github.com/Kong/kong/pull/9337#pullrequestreview-1143550684).

Why does Kong needs a user configurable socket directory in order to function correctly?

chronolaw: It has a default value (`ngx.config.prefix()`), so the user does not need to know it.

I think this env is only for internal usage, and it is a POC for plan B in the ticket.

Tieske: @dndx We need it for testing, see https://github.com/Kong/kong/pull/9337#issuecomment-1247893735

chronolaw:
It seems that there is no further discussion about it, so I decide to close it.

We can open it at any time if needed.

PR #9563, opened 2022-10-17
tests(migrations): add 3.1 migrations spec

attenuation:
### Summary
add 3.1 migrations spec

### Issue reference
add 3.1 migrations spec for #9430 
chery-pick #9487


chronolaw:
What is the difference between this PR and #9542 ?

PR #9598, opened 2022-10-21
fix(aws-lambda): accept only true|false|nil for isBase64Encode

windmgc:
### Summary

This PR changes the behaviour of extracting lambda response in the AWS-Lambda plugin. There will be 3 situations that are considered valid for the field `isBase64Encode`: true, false or not defined. Values that are not booleans or not nil are considered invalid and the plugin will return a 502 response.

Note that if a lambda response contains `"isBase64Encode": null`, then the value(turned into userdata `cjson.null` in Lua) will also be considered invalid.

### Full changelog

* Fix the behaviour of extracting lambda response that the value of isBase64Encode can only accept true, false or not defined.
* Add related tests

### Issue reference

Fix [FTI-4420]

### Other references

https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

[FTI-4420]: https://konghq.atlassian.net/browse/FTI-4420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kong/plugins/aws-lambda/handler.lua
+++ b/kong/plugins/aws-lambda/handler.lua
@@ -145,9 +145,12 @@ local function extract_proxy_response(content)
 
   local headers = serialized_content.headers or {}
   local body = serialized_content.body or ""
-  local isBase64Encoded = serialized_content.isBase64Encoded or false
-  if isBase64Encoded then
+  local isBase64Encoded = serialized_content.isBase64Encoded
+  if isBase64Encoded == true then
     body = ngx_decode_base64(body)
+
+  elseif isBase64Encoded ~= false and isBase64Encoded ~= nil then
+    return nil, "isBase64Encoded must be a boolean"
   end
 
   local multiValueHeaders = serialized_content.multiValueHeaders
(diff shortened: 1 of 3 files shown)


kong/plugins/aws-lambda/handler.lua
```diff
@@ -145,9 +145,12 @@ local function extract_proxy_response(content)
...
   local headers = serialized_content.headers or {}
   local body = serialized_content.body or ""
-  local isBase64Encoded = serialized_content.isBase64Encoded or false
-  if isBase64Encoded then
+  local isBase64Encoded = serialized_content.isBase64Encoded
+  if isBase64Encoded == true then
     body = ngx_decode_base64(body)
+
+  elseif isBase64Encoded ~= false and isBase64Encoded ~= nil then
+    return nil, "isBase64Encoded must be a boolean"
```
bungle, line 153: Can you elaborate why we don't consider everything else `== false`?

bungle: Ok, got answer, this is to mimic Amazon API GW behavior (to be noted, it is a breaking change though).

mayocream (approved):

bungle (approved):

PR #9599, opened 2022-10-21
style(api): api code clean

chronolaw:
### Summary

Small code and style clean.

### DIFF
--- a/kong/api/api_helpers.lua
+++ b/kong/api/api_helpers.lua
@@ -15,6 +15,8 @@ local type = type
 local pairs = pairs
 local ipairs = ipairs
 
+local get_method = ngx.req.get_method
+
 
 local _M = {}
 local NO_ARRAY_INDEX_MARK = {}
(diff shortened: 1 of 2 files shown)


hanshuebner:
For my understanding, what is the purpose of these changes?

I generally like to avoid introducing unnecessary indirections as it makes code harder to follow - One needs to either remember that a local name has been introduced for something that is nested in a global, or one needs to look up the local variable when it is used.

I have heard that such local variables declarations have performance benefits with Lua, but it seems less clear that they are as beneficial with LuaJIT.  While local variable accesses might be slightly faster, the improvement will only be visible when a code path is executed millions of times.  Thus, trading a negligible performance benefit for makeing the code harder to read, reason about and to refactor does not seem to be the right thing.

chronolaw:
I am not sure about the mechanism of Lua JIT, but the localization for var is the usual Lua pattern.

I agree that these changes are small and may not get an obvious performance improvement. 

It is mostly the style fix, if we don't like this, just close it, this is OK to me.

bungle (approved):

PR #9603, opened 2022-10-22
feat(db): exposed postgres connection pooling config

shettyh:
### Summary

As of now there is no straight forward way to control the number of postgres connections created. There are some workarounds like setting [lua_socket_pool_size](https://github.com/openresty/lua-nginx-module#lua_socket_pool_size), but that has it own downsides as it effects the global upstream connection pool also.

Because of this kong can bring down the postgres instance in case of high load and most of request are resulting in a cache miss. As part of this change exposed config to control this behaviour.

This change does not impact the default behaviour of kong

### Full changelog

* Added config `pg_keepalive_timeout ` to control the postgres connection idle timeout
* Added config `pg_pool_size ` to control the postgres connection pool size
* Added config `pg_backlog ` to limit the open connections to postgres
* Added relevant test cases


kong.conf.default
```diff
@@ -1124,6 +1124,27 @@
                                  # Detailed discussion of this behavior is
                                  # available in the online documentation.
 
+#pg_connection_keepalive_timeout = 60000 # Specify the maximal idle timeout ( in ms )
```
mayocream, line 1127: Personally, I'd prefer `pg_keepalive_timeout`, so do `pg_pool_size` and `pg_backlog`. This aligns with [pgmoon options](https://github.com/leafo/pgmoon/blob/master/README.md#newoptions).

kong.conf.default
```diff
@@ -1124,6 +1124,27 @@
...
+                                         # for the postgres connections in the pool.
+                                         # If values is set to 0 then the timeout interval
+                                         # is unlimited.
+
+#pg_connection_pool_size = 30    # Specifies the size limit (in terms of connection count) for
```
mayocream, line 1132: Keep the default value as `nil`?

kong/db/strategies/postgres/connector.lua
```diff
@@ -972,6 +987,11 @@ function _M.new(kong_config)
       sem_max     = kong_config.pg_ro_max_concurrent_queries,
       sem_timeout = kong_config.pg_ro_semaphore_timeout and
                     (kong_config.pg_ro_semaphore_timeout / 1000) or nil,
+      pool_size   = kong_config.pg__ro_connection_pool_size,
```
mayocream, line 990: nit: typo
```suggestion
      pool_size   = kong_config.pg_ro_connection_pool_size,
```

mayocream (commented):
Related pgmoon PR: 
- https://github.com/leafo/pgmoon/pull/105

This PR is indeed needed for improving the Postgres connections.

kong.conf.default
shettyh: Yea makes sense, updated

kong.conf.default
shettyh: Updated to keep nil as default

kong/db/strategies/postgres/connector.lua
shettyh: fixed

kong.conf.default
```diff
@@ -1124,6 +1124,29 @@
                                  # Detailed discussion of this behavior is
                                  # available in the online documentation.
 
+#pg_keepalive_timeout = 60000    # Specify the maximal idle timeout ( in ms )
```
mayocream, line 1127:
```suggestion
#pg_keepalive_timeout = 60000    # Specify the maximal idle timeout (in ms)
```

kong/db/strategies/postgres/connector.lua
```diff
@@ -922,6 +932,11 @@ function _M.new(kong_config)
     cafile      = kong_config.lua_ssl_trusted_certificate_combined,
     sem_max     = kong_config.pg_max_concurrent_queries or 0,
     sem_timeout = (kong_config.pg_semaphore_timeout or 60000) / 1000,
+    pool_size   = kong_config.pg_pool_size or nil,
+    backlog     = kong_config.pg_backlog or nil,
```
mayocream, line 936: should be ok to remove `or nil`, will be `nil` by default
```suggestion
    pool_size   = kong_config.pg_pool_size,
    backlog     = kong_config.pg_backlog,
```

kong/db/strategies/postgres/connector.lua
```diff
@@ -972,6 +987,11 @@ function _M.new(kong_config)
       sem_max     = kong_config.pg_ro_max_concurrent_queries,
       sem_timeout = kong_config.pg_ro_semaphore_timeout and
                     (kong_config.pg_ro_semaphore_timeout / 1000) or nil,
+      pool_size   = kong_config.pg_ro_pool_size,
+      backlog     = kong_config.pg_ro_backlog or nil,
```
mayocream, line 991:
```suggestion
      backlog     = kong_config.pg_ro_backlog,
```

kong.conf.default
mayocream: Does it make sense to make the default value to be `nil` (default OpenResty behavior)?

> The first optional argument, timeout, can be used to specify the maximal idle timeout (in milliseconds) for the current connection. If omitted, the default setting in the [lua_socket_keepalive_timeout](https://github.com/openresty/lua-nginx-module#lua_socket_keepalive_timeout) config directive will be used. If the 0 value is given, then the timeout interval is unlimited.

kong/db/strategies/postgres/connector.lua
shettyh: updated

kong/db/strategies/postgres/connector.lua
shettyh: Updated

kong.conf.default
shettyh: Yes makes sense, updated

hbagdi:
@dndx Should we expose this level of detail? Can dynamic scaling of workers cause bad surprises?

dndx:
@hbagdi As long as we provide reasonable defaults this should not be a bad thing to have. Most user don't need to touch it anyway.

spec/02-integration/03-db/15-connection_pool_spec.lua
```diff
@@ -0,0 +1,110 @@
...
+  it("results in query error too many waiting connect operations", function()
+    local res = assert(client:send {
+      method = "GET",
+      path = "/slow-resource?prime=true",
+      headers = { ["Content-Type"] = "application/json" }
+    })
+    assert.res_status(204 , res)
+
+    res = assert(client:send {
+      method = "GET",
```
samugi, line 45:
```suggestion
    })
    assert.res_status(204 , res)
    
    helpers.wait_timer("slow-query", true, "any-running")

    res = assert(client:send {
      method = "GET",
```
this should prevent flakiness

spec/02-integration/03-db/15-connection_pool_spec.lua
```diff
@@ -0,0 +1,110 @@
...
+      name = "slow-query",
+    })
+
+    assert(helpers.start_kong({
+      database = "postgres",
+      nginx_conf = "spec/fixtures/custom_nginx.template",
+      plugins = "slow-query",
+      nginx_worker_processes = 1,
+      pg_pool_size = 1,
+      pg_backlog = 1,
```
samugi, line 75:
```suggestion
      nginx_worker_processes = 1,
      pg_pool_size = 3,
      pg_backlog = 1,
```
This would make the test more meaningful since we are repeating the request 3 times to fill the pool below

spec/02-integration/03-db/15-connection_pool_spec.lua
```diff
@@ -0,0 +1,110 @@
...
+        headers = { ["Content-Type"] = "application/json" }
+      })
+      assert.res_status(204 , res)
+    end
+
+    for i = 0, 2 do
+      handler()
+    end
+
+    local res = assert(client:send {
```
samugi, line 101:
```suggestion
  it("results in query error too many waiting connect operations when backlog exceeds", function()
    helpers.wait_timer("slow-query", true, "all-finish")
    local handler = function()
      local res = assert(client:send {
        method = "GET",
        path = "/slow-resource?prime=true",
        headers = { ["Content-Type"] = "application/json" }
      })
      assert.res_status(204 , res)
    end

    for _ = 0, 2 do
      handler()
    end
    
    helpers.wait_timer("slow-query", true, "all-running")

    local res = assert(client:send {
```
this should address the current test flakiness, making sure the timers have started (pool is full) before proceeding with the unhappy path test

samugi (changes requested):

kong.conf.default
```diff
@@ -1129,6 +1129,31 @@
...
+                                 # is unlimited.
+                                 #
+                                 # If not specified this value will be same as `lua_socket_keepalive_timeout`
+
+#pg_pool_size =                  # Specifies the size limit (in terms of connection count) for
+                                 # postgres server.
+                                 # Note that this connection pool is per nginx worker rather than per kong instance,
+                                 # so size limit specified applies to every single Nginx worker process.
+                                 #
+                                 # If not specified default value is same as `lua_socket_pool_size`
```
samugi, line 1144:
```suggestion
#pg_pool_size =                  # Specifies the size limit (in terms of connection 
                                 # count) for the Postgres server.
                                 # Note that this connection pool is intended 
                                 # per Nginx worker rather than per Kong instance.
                                 #
                                 # If not specified, the default value is the same as `lua_socket_pool_size`
```
description / format

kong.conf.default
```diff
@@ -1129,6 +1129,31 @@
...
+#pg_backlog =                    # If specified, this will limit the total of number of opened connections
+                                 # to postgres server to `pg_pool_size`. If the connection pool is full,
+                                 # subsequent connect operations will be queued into a queue equal to this option's value.
+                                 #
+                                 # If the number of queued connect operations is equal to `pg_backlog`,
+                                 # subsequent connect operations will fail and return nil plus the
+                                 # error string "too many waiting connect operations"
+                                 #
+                                 # If not specified, then number of open connections to postgres server is not limited
+                                 # and this the default behaviour.
```
samugi, line 1155:
```suggestion
#pg_backlog =                    # If specified, this value will limit the total 
                                 # number of open connections to the Postgres 
                                 # server to `pg_pool_size`. If the connection 
                                 # pool is full, subsequent connect operations
                                 # will be inserted in a queue with size equal
                                 # to this option's value.
                                 #
                                 # If the number of queued connect operations 
                                 # reaches `pg_backlog`, exceeding connections will fail.
                                 #
                                 # If not specified, then number of open connections
                                 # to the Postgres server is not limited.
```
description / format

samugi (changes requested):

shettyh:
@samugi thanks for the review. Addressed comments, please check

spec/02-integration/03-db/15-connection_pool_spec.lua
shettyh: Addressed

spec/02-integration/03-db/15-connection_pool_spec.lua
```diff
@@ -0,0 +1,119 @@
...
+
+    for i = 0, 1 do
+      threads[i] = ngx.thread.spawn(handler)
+    end
+
+    helpers.wait_timer("slow-query", true, "all-running")
+
+    for i = 0, 1 do
+      ngx.thread.wait(threads[i])
+    end
```
samugi, line 107: just a question: is there a point in waiting the threads here? We know at this point that the connections are already open, because of the call to `helpers.wait_timer()` above, right?

I tried to comment these lines out and the test keeps passing as expected. If you agree I would suggest removing this for loop.

samugi (approved):
Thank you @shettyh!! Looks good to me. I just left a small comment on a minor detail.

spec/02-integration/03-db/15-connection_pool_spec.lua
shettyh: Yes makes sense. Removed

samugi:
@shettyh great job! Could you add a changelog entry in the `Unreleased` section please?

chronolaw:
Let's merge it after 3.1 release.

shettyh:
@samugi updated changelog, please check

mayocream:
Hi @shettyh, the failing CI was fixed by #9949, rebased.

shettyh:
@mayocream rebased, one of the test is still failing with config error

CHANGELOG.md
```diff
@@ -124,6 +124,8 @@
   [#9737](https://github.com/Kong/kong/pull/9737)
 - Add `key-sets` entity to group and manage `keys`
   [#9737](https://github.com/Kong/kong/pull/9737)
+- Expose postgres connection pool configuration
```
chronolaw, line 127: Should move to the `Unreleased` section.

shettyh: Moved

CHANGELOG.md
```diff
@@ -123,7 +125,7 @@
 - Add `keys` entity to store and manage asymmetric keys.
   [#9737](https://github.com/Kong/kong/pull/9737)
 - Add `key-sets` entity to group and manage `keys`
-  [#9737](https://github.com/Kong/kong/pull/9737)
+  [#9737](https://github.com/Kong/kong/pull/9737
```
chronolaw, line 128: it seems you removed a `)` by mistake.

mayocream:
```suggestion
  [#9737](https://github.com/Kong/kong/pull/9737)
```

mayocream (approved):

PR #9606, opened 2022-10-24
fix(router/atc): exports atc.schema for entities checking

chronolaw:
### Summary

When we did code clean for atc-router(#9331), we removed function `get_schema()` unwisely,
which is used in DB schema entity checking.

See [`schema/entities/routes.lua`](https://github.com/Kong/kong/blob/e7b6963f8c01c11232828d2709de08743708bf56/kong/db/schema/entities/routes.lua#L50).

### DIFF
--- a/kong/router/atc.lua
+++ b/kong/router/atc.lua
@@ -554,11 +554,6 @@ function _M:exec(ctx)
 end
 
 
-function _M.get_schema()
-  return CACHED_SCHEMA
-end
-
-
 function _M._set_ngx(mock_ngx)
   if type(mock_ngx) ~= "table" then
     return
@@ -587,6 +582,8 @@ function _M._set_ngx(mock_ngx)
   end
 end
 
+
+_M.schema          = CACHED_SCHEMA
 
 _M.LOGICAL_OR      = LOGICAL_OR
 _M.LOGICAL_AND     = LOGICAL_AND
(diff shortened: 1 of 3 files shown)


chronolaw:
> @chronolaw is it possible to add test to this? Shouldn't it e.g. runtime crash or something?

I will try to figure it out.

kong/router/atc.lua
```diff
@@ -554,6 +554,11 @@ function _M:exec(ctx)
 end
 
 
+function _M.get_schema()
+  return CACHED_SCHEMA
+end
```
mayocream, line 559: As it is a static table, won't be better to use the `_M.schema` format?

```suggestion
_M.schema = CACHED_SCHEMA
```

chronolaw:
> @chronolaw is it possible to add test to this? Shouldn't it e.g. runtime crash or something?

We have unit-test in `spec/01-unit/08-router_spec.lua`, but it will not trigger schema check.

And integration test `spec/02-integration/05-proxy/02-router_spec.lua`, we only test `traditional` and `traditional_compatible`, so this is the reason.

chronolaw:
I am trying to enable test for flavor `expressions` in `spec/02-integration/05-proxy/02-router_spec.lua`, but I get some failing cases.

Maybe we can merge it now, then add test cases in another PR.

dndx:
If it is impossible to get the entire existing tests running under `expressions` mode, we should at least add a single test case just to catch this regression.

dndx (approved):

PR #9611, opened 2022-10-25
feat(opentelemetry): add referenceable attribute to fields that could be stored in vaults

mayocream:
### Summary
Add `referenceable=true` attribute to plugin schema fields that could be stored in vaults.


kong/db/schema/init.lua
```diff
@@ -1777,6 +1778,52 @@ function Schema:process_auto_fields(data, context, nulls, opts)
...
+                  refs[key] = prev_refs[key]
+                end
+
+              else
+                refs = { [key] = prev_refs[key] }
+              end
+            end
+          end
+
+        elseif vtype == "table" and ftype == "map" then
```
fffonion, line 1793: could you add a test for this change, either one in spec/02-integration/03-db/11-db_transformations_spec.lua or a unit test?

hbagdi:
@mayocream Please rebase.

kong/db/schema/init.lua
mayocream: Added, thanks!

fffonion:
one unit test is failing and looks relevant https://github.com/Kong/kong/actions/runs/3383824165/jobs/5620091308

fffonion (approved):

PR #9616, opened 2022-10-26
fix(clustering): race cond for new DP and new cfg

StarlightIbuki:
1. When new DP connects, it first pushes config, then adds the DP to client lists;
1. When the client list is empty, `push_config_loop` skips handling of the new config signal, expect the config to be sent by logic no.1

This causes races when a new config comes in while the first DP is receiving its first config, and the config will be skipped as the list is empty. The new config won't be seen by that DP or any new DPs.

We fix that by re-posting the semaphore, so no config will be missed.

Co-authored-by: @flrgh for his diagnosis(don't forget to add this to the commit message)

fix FT-3496

### DIFF
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -553,7 +553,7 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
       if isempty(self.clients) then
         ngx_log(ngx_DEBUG, _log_prefix, "skipping config push (no connected clients)")
         sleep(1)
-        -- re-queue the task, wait util we have clients connected
+        -- re-queue the task. wait until we have clients connected
         if push_config_semaphore:count() <= 0 then
           push_config_semaphore:post()
         end
--- a/kong/clustering/control_plane.lua
+++ b/kong/clustering/control_plane.lua
@@ -551,6 +551,11 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if ok then
       if isempty(self.clients) then
         ngx_log(ngx_DEBUG, _log_prefix, "skipping config push (no connected clients)")
+        sleep(1)
+        -- re-queue the task, wait util we have clients connected
+        if push_config_semaphore:count() <= 0 then
+          push_config_semaphore:post()
+        end
 
       else
         ok, err = pcall(self.push_config, self)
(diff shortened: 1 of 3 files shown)


kong/clustering/wrpc_control_plane.lua
```diff
@@ -316,7 +316,11 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if ok then
       if isempty(self.clients) then
         ngx_log(ngx_DEBUG, _log_prefix, "skipping config push (no connected clients)")
-
+        sleep(1)
+        -- re-queue the task, wait util we have clients connected
+        if push_config_semaphore:count() <= 0 then
+          push_config_semaphore:post()
+        end
       else
```
StarlightIbuki, line 324:
```suggestion

      else
```

kong/clustering/wrpc_control_plane.lua
```diff
@@ -316,6 +317,11 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if ok then
       if isempty(self.clients) then
         ngx_log(ngx_DEBUG, _log_prefix, "skipping config push (no connected clients)")
+        sleep(1)
+        -- re-queue the task, wait util we have clients connected
```
chronolaw, line 321: typo: util => until

kong/clustering/control_plane.lua
```diff
@@ -551,6 +552,11 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
     if ok then
       if isempty(self.clients) then
         ngx_log(ngx_DEBUG, _log_prefix, "skipping config push (no connected clients)")
+        sleep(1)
```
javierguerragiraldez, line 555: is this sleep needed? can it be `sleep(0)`?  can we be sure that the empty list is still empty after the sleep?

StarlightIbuki: It will be nearly a tight loop if we do sleep(0), which may add to the delay of admin API.

kong/clustering/control_plane.lua
```diff
@@ -551,6 +552,11 @@ local function push_config_loop(premature, self, push_config_semaphore, delay)
...
+        -- re-queue the task, wait util we have clients connected
```
StarlightIbuki, line 556:
```suggestion
        -- re-queue the task. wait until we have clients connected
```

kong/clustering/wrpc_control_plane.lua
StarlightIbuki, line 321:
```suggestion
        -- re-queue the task. wait until we have clients connected
```

chronolaw (approved):

PR #9624, opened 2022-10-27
perf(router): increase `lua_regex_cache_max_entries` and show warning if too small

mayocream:
### Summary

When a user sets too many regex routes and `lua_regex_cache_max_entries` is set too low (or kept at default 1024), a performance issue might occur due to regex trashing. 

OpenResty [claims](https://github.com/openresty/lua-nginx-module/blob/master/README.markdown#lua_regex_cache_max_entries) that it will emit a warning when the limit has been exceeded, but it doesn't.

Fix FTI-4455

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,9 @@
 
 - Data plane's connection to control plane is moved to a privileged worker process
   [#9432](https://github.com/Kong/kong/pull/9432)
+- Increase the default value of `lua_regex_cache_max_entries`, a warning will be thrown
+  when there are too many regex routes and `router_flavor` is `traditional`.
+  [#9624](https://github.com/Kong/kong/pull/9624)
 
 ### Fixes
 
--- a/kong/router/traditional.lua
+++ b/kong/router/traditional.lua
@@ -1452,7 +1452,8 @@ function _M.new(routes, cache, cache_neg)
 
   -- warning about the regex cache size being too small
   if not lua_regex_cache_max_entries then
-    lua_regex_cache_max_entries = kong.configuration.nginx_http_lua_regex_cache_max_entries or 1024
+    lua_regex_cache_max_entries = kong.configuration.nginx_http_lua_regex_cache_max_entries
+      and tonumber(kong.configuration.nginx_http_lua_regex_cache_max_entries) or 1024
   end
 
   if worker_id() == 0 and regex_uris[0] > lua_regex_cache_max_entries then
--- a/kong/router/traditional.lua
+++ b/kong/router/traditional.lua
@@ -1456,7 +1456,7 @@ function _M.new(routes, cache, cache_neg)
                               and tonumber(kong.configuration.nginx_http_lua_regex_cache_max_entries) or 1024
   end
 
-  if worker_id() == 0 and regex_uris[0] > lua_regex_cache_max_entries then
+  if worker_id() == 0 and regex_uris[0] * 2 > lua_regex_cache_max_entries then
     ngx_log(WARN, "the 'nginx_http_lua_regex_cache_max_entries' setting is set to ",
                    lua_regex_cache_max_entries,
                    " but there are ", regex_uris[0], " regex URIs in use. ",
--- a/kong/router/traditional.lua
+++ b/kong/router/traditional.lua
@@ -1453,7 +1453,7 @@ function _M.new(routes, cache, cache_neg)
   -- warning about the regex cache size being too small
   if not lua_regex_cache_max_entries then
     lua_regex_cache_max_entries = kong.configuration.nginx_http_lua_regex_cache_max_entries
-      and tonumber(kong.configuration.nginx_http_lua_regex_cache_max_entries) or 1024
+                              and tonumber(kong.configuration.nginx_http_lua_regex_cache_max_entries) or 1024
   end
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -92,6 +92,9 @@
 
 - Data plane's connection to control plane is moved to a privileged worker process
   [#9432](https://github.com/Kong/kong/pull/9432)
+- Increase the default value of ‘lua_regex_cache_max_entries', a warning will be thrown
+  when there are too many regex routes and 'router_flavor' is ‘traditional'.
```
chronolaw, line 96: It seems there is quote typo: 

‘traditional' should be `traditional` or 'traditional'

kong/router/traditional.lua
```diff
@@ -1448,6 +1449,17 @@ function _M.new(routes, cache, cache_neg)
   local match_sources        = not isempty(plain_indexes.sources)
   local match_destinations   = not isempty(plain_indexes.destinations)
 
+  -- warning about the regex cache size being too small
+  if worker_id() == 0 and match_regex_uris
+    and regex_uris[0] > (kong.configuration.nginx_http_lua_regex_cache_max_entries or 1024) then
```
chronolaw, line 1454: should use a var to cache `kong.configuration.nginx_http_lua_regex_cache_max_entries or 1024`

CHANGELOG.md
mayocream:
```suggestion
- Increase the default value of `lua_regex_cache_max_entries`, a warning will be thrown
  when there are too many regex routes and `router_flavor` is `traditional`.
```

mayocream, line 96:
```suggestion
- Increase the default value of `lua_regex_cache_max_entries`, a warning will be thrown
  when there are too many regex routes and `router_flavor` is `traditional`.
```

kong/router/traditional.lua
mayocream: done, thanks

kong/router/traditional.lua
```diff
@@ -208,7 +209,7 @@ local MAX_REQ_HEADERS = 100
 
 local match_route
 local reduce
-
+local lua_regex_cache_max_entries
```
mayocream, line 212:
```suggestion
local lua_regex_cache_max_entries

```

chronolaw (approved):

kong/router/traditional.lua
```diff
@@ -1448,6 +1450,20 @@ function _M.new(routes, cache, cache_neg)
   local match_sources        = not isempty(plain_indexes.sources)
   local match_destinations   = not isempty(plain_indexes.destinations)
 
+  -- warning about the regex cache size being too small
+  if not lua_regex_cache_max_entries then
+    lua_regex_cache_max_entries = kong.configuration.nginx_http_lua_regex_cache_max_entries or 1024
```
bungle, line 1455: Should we default here to 8192 too?

mayocream: This seems non-sense because OpenResty's default value of `lua_regex_cache_max_entries` is `1024` (https://github.com/openresty/lua-nginx-module/blob/master/README.markdown#lua_regex_cache_max_entries), we set the `nginx_http_lua_regex_cache_max_entries` to `8192`, but the user could overwrite it.

kong/router/traditional.lua
```diff
@@ -1448,6 +1450,21 @@ function _M.new(routes, cache, cache_neg)
   local match_sources        = not isempty(plain_indexes.sources)
   local match_destinations   = not isempty(plain_indexes.destinations)
 
+  -- warning about the regex cache size being too small
+  if not lua_regex_cache_max_entries then
+    lua_regex_cache_max_entries = kong.configuration.nginx_http_lua_regex_cache_max_entries
+                              and tonumber(kong.configuration.nginx_http_lua_regex_cache_max_entries) or 1024
+  end
+
+  if worker_id() == 0 and regex_uris[0] > lua_regex_cache_max_entries then
```
dndx, line 1459: I think we should also be checking for `regex_uris[0] * 2 > lua_regex_cache_max_entries` here, if user sets `lua_regex_cache_max_entries` to be exactly `regex_uris[0]` we can still expect issues regarding regex cache trashing.

mayocream: Thanks, done.

kong/router/traditional.lua
```diff
@@ -1448,6 +1450,21 @@ function _M.new(routes, cache, cache_neg)
   local match_sources        = not isempty(plain_indexes.sources)
   local match_destinations   = not isempty(plain_indexes.destinations)
 
+  -- warning about the regex cache size being too small
+  if not lua_regex_cache_max_entries then
+    lua_regex_cache_max_entries = kong.configuration.nginx_http_lua_regex_cache_max_entries
```
chronolaw, line 1455: `tonumber(kong.configuration.nginx_http_lua_regex_cache_max_entries) or 1024` is OK,
Because `tonumber(nil) == nil` `tonumber('xxx') == nil`

dndx (approved):

ddskolla:
We faced an issue around regex usage and kong performance(cpu spikes) in production recently. When you say `a performance issue might occur due to regex trashing.` do you have any documentation around what type of perf issues could arise because of this ?
We are currently on 1.5.0.8 Enterprise and would love to know more and add this cach fix in asap.

PR #9625, opened 2022-10-27
tests(https_server): retry if failed to start Nginx

ADD-SP:
Fix https://github.com/Kong/kong/actions/runs/3334462656/jobs/5517448732

_FT-3508_

### DIFF
--- a/spec/fixtures/https_server.lua
+++ b/spec/fixtures/https_server.lua
@@ -19,7 +19,7 @@ local cjson = require "cjson"
 math.randomseed(os.time())
 
 
-local MAX_RETRY = 10
+local HTTPS_SERVER_START_MAX_RETRY = 10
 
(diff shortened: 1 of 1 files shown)


chronolaw (approved):

spec/fixtures/https_server.lua
```diff
@@ -19,10 +19,13 @@ local cjson = require "cjson"
 math.randomseed(os.time())
 
 
+local MAX_RETRY = 10
```
chronolaw, line 22: Could we move `MAX_RETRY` near the loop? it may be easy to understand.

dndx: Also rename it to `HTTPS_SERVER_START_MAX_RETRY` to make the name more specific.

dndx (approved):

PR #9628, opened 2022-10-27
fix(opentelemetry): fix parent_id reporting from w3c header

### LINKED ISSUE #9627: Tracing span parent_id not set when context propagated in w3c header

3.0.0 and master(1b60a92c)

### Current Behavior

When trace context in a call is propagated in the w3c `Traceparent` header, the parent_id of the root kong span is not set and reported to the collector. The trace then splits into two separate pieces.

### Expected Behavior

parent_id should be correctly set to keep the trace continuous.

### Steps To Reproduce

```markdown
Enable opentelemetry plugin and tracing
Call kong with header `Traceparent: 00-[traceId]-[parentId]-01`
The span reported to the tracing collector contains `nil` parent_id.
```

### Anything else?

PR is on its way

MartinBurian:
### Summary

When trace context is propagated in w3c format header, the parent_id is not set on the root span and not reported to the collector. I could not come up with a way to cover it in the tests.

### Full changelog

* Fix `parent_id` reporting in traces propagating through w3c headers

### Issue reference

Fix #9627

### DIFF
--- a/kong/plugins/opentelemetry/handler.lua
+++ b/kong/plugins/opentelemetry/handler.lua
@@ -116,7 +116,7 @@ function OpenTelemetryHandler:rewrite()
     kong.ctx.plugin.should_sample = false
   end
 
-  local header_type, trace_id, span_id, _, should_sample, _ = propagation_parse(headers)
+  local header_type, trace_id, span_id, parent_id, should_sample, _ = propagation_parse(headers)
   if should_sample == false then
     root_span.should_sample = should_sample
   end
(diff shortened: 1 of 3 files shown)


kong/plugins/opentelemetry/handler.lua
```diff
@@ -176,6 +176,8 @@ function OpenTelemetryHandler:rewrite()
   -- overwrite root span's parent_id
   if span_id then
     root_span.parent_id = span_id
+  elseif parent_id then
```
mayocream, line 179: nit: 
```suggestion

  elseif parent_id then
```

mayocream (commented):
Thanks! I'll add a coordinated test.

mayocream (approved):

PR #9632, opened 2022-10-28
Fix: clustering compatibility for custom response for rate-limiting

StarlightIbuki:
We should remove newly added config fields for older DP.
https://github.com/Kong/kong/pull/8930


spec/01-unit/19-hybrid/03-fields-removal_spec.lua
```diff
@@ -191,13 +203,22 @@ describe("kong.clustering.control_plane", function()
...
     }, cp._get_removed_fields(2006000000))
 
-    assert.same(nil, cp._get_removed_fields(2007000000))
+    assert.same({
+      rate_limiting = {
+        "error_code",
+        "error_message",
+      },
+    }, cp._get_removed_fields(2007000000))
+
```
ADD-SP, line 220:
```suggestion
 assert.same({
      rate_limiting = {
        "error_code",
        "error_message",
      },
    }, cp._get_removed_fields(2008000000))
```

fffonion (approved):

PR #9633, opened 2022-10-28
feat(clustering): backup config sync for CP outage

StarlightIbuki:
Fix FT-3155


StarlightIbuki:
We still need to fix a few bugs for [lua-resty-aws](https://github.com/Kong/lua-resty-aws) before we could test.

PR #9635, opened 2022-10-30
fix(router/atc): remove potential race condition

chronolaw:
### Summary

We use a module-level table `headers_t` to calculate the cache key, 
but it may be a race condition in huge concurrency. 

### Full changelog

* move var `headers_t` into function `get_headers_key`
* style fix


dndx (commented):
Why is this a race condition? I did not found any yielding point inside this function.

chronolaw:
Maybe I am too worry about it, I will close this PR>

PR #9653, opened 2022-11-01
refactor(core/runloop): code clean for reconfigure event processing

chronolaw:
### Summary

In reconfigure event handler, we use `current_router_hash` and others
to compare with new hashes, but they are function local vars.
It is not a common used pattern.

This PR puts these hash vars into `do block` scope and did some style cleaning,
Now the code is more readable and maintainable.

### Full changelog

* move `current_router_hash` `current_plugins_hash` `current_balancer_hash` to `do block` level
* move logic into function `reconfigure_handler()`
* use `get_now_ms()` to calculate delta time
* remove unnecessary `is_exiting()`
* change the hash's default value to `0`, so we can compare them with `nil` directly
* style fix

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -829,8 +829,8 @@ local function register_events()
   local worker_events  = kong.worker_events
   local cluster_events = kong.cluster_events
 
-  -- declarative config updates
   if db.strategy == "off" then
+    -- declarative config updates
     worker_events.register(reconfigure_handler, "declarative", "reconfigure")
   end
 
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -692,6 +692,143 @@ local function _register_balancer_events(f)
 end
 
 
+local register_reconfigure_event
+do
+  local now = ngx.now
+  local update_time = ngx.update_time
+  local ngx_worker_id = ngx.worker.id
+  local exiting = ngx.worker.exiting
+
+  local current_router_hash
+  local current_plugins_hash
+  local current_balancer_hash
+
+  local function is_exiting(worker_id)
+    if not exiting() then
+      return false
+    end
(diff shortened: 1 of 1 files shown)


kong/runloop/handler.lua
```diff
@@ -690,143 +690,142 @@ local function _register_balancer_events(f)
...
+do
+  local now = ngx.now
+  local update_time = ngx.update_time
+  local ngx_worker_id = ngx.worker.id
+  local exiting = ngx.worker.exiting
+
+  -- '0' for compare with nil
+  local current_router_hash   = 0
+  local current_plugins_hash  = 0
+  local current_balancer_hash = 0
```
bungle, line 703: The PR says that it uses module level variables, but these are really `do block` level variables. Should you move them to module level and capitalize them as:

```
local CURRENT_ROUTER_HASH  = 0
local CURRENT_PLUGINS_HASH = 0
local CURRENT_BALANCER_HASH = 0
```

Though I am not sure if it is worth it, do we need them anywhere else in this module, or do we need to export them as part of this module, e.g. `get_router_hash()` etc. functions?

chronolaw: Thanks for you opinions.

I think it is my mistake, these hash vars are not module level, they are `do block` level.
I check the code again and I am sure these hashes are only used in this module.

kong/runloop/handler.lua
```diff
@@ -78,7 +78,6 @@ local HOST_PORTS = {}
```
flrgh, line 78: This file is usually a source of many headaches when merging OSS changes into EE, so if you are doing refactor work it is recommended to open a PR against EE as well.

chronolaw: Yes, if this PR is merged, I will cherry-pick it into EE.

chronolaw:
I will rebase this PR to latest master and test it.

PR #9655, opened 2022-11-01
chore(github): setup the settings bot so the team can self-manage

hutchic:
requires the installation of the [Settings Github App](https://github.com/apps/settings) after which changing branch protection rules et al can be achieved with a pull request.

### DIFF
--- a/kong/plugins/ldap-auth/access.lua
+++ b/kong/plugins/ldap-auth/access.lua
@@ -5,8 +5,6 @@ local ldap = require "kong.plugins.ldap-auth.ldap"
 local kong = kong
 local error = error
 local decode_base64 = ngx.decode_base64
-local sha1_bin = ngx.sha1_bin
-local to_hex = require "resty.string".to_hex
 local tostring =  tostring
 local match = string.match
 local lower = string.lower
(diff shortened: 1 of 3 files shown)


hbagdi:
Please block until approval from @RobSerafini and @dndx.

dndx:
Upon closer inspection it seems that this is not suitable for enablement on the Kong repo. See: https://github.com/apps/settings

> WARNING: Note that this app inherently escalates anyone with push permissions to the admin role, since they can push config settings to the master branch, which will be synced. In a future, we may add restrictions to allow changes to the config file to be merged only by specific people/teams, or those with admin access (via a combination of protected branches, required statuses, and branch restrictions). Until then, use caution when merging PRs and adding collaborators.

> Until restrictions are added in this app, one way to preserve admin/push permissions is to utilize the [GitHub CodeOwners feature](https://help.github.com/articles/about-codeowners/) to set one or more administrative users as the code owner of the .github/settings.yml file, and turn on "require code owner review" for the master branch. This does have the side effect of requiring code owner review for the entire branch, but helps preserve permission levels.

It seems that this is not suitable for larger projects like Kong, but I agree with smaller projects where every contributor is Admin already, it actually makes sense. Or if the app supports configuring which repo config can be set with declarative config.

This App also needs an organization owner to enable. Repo Admin can not even install it.

Ping @RobSerafini for opinions.

hutchic:
>Note that this app inherently escalates anyone with push permissions to the admin role, since they can push config settings to the master branch

should actually be

Note that this app inherently escalates anyone with push permissions **to the master branch** to the admin role, since they can push config settings to the master branch

I therefore feel this is suitable when combined with branch protection rules.

PR #9681, opened 2022-11-04
style(init): introduce role/mode check functions

chronolaw:
### Summary

Remove repeated code with some small functions, let the code more readable.

Introduce new functions:
* is_data_plane 
* is_control_plane 
* is_dbless

### DIFF
--- a/kong/init.lua
+++ b/kong/init.lua
@@ -178,6 +178,32 @@ do
 end
 
 
+local is_data_plane
+local is_control_plane
+local is_dbless
+local is_dp_worker_process
+do
+  is_data_plane = function(config)
+    return config.role == "data_plane"
+  end
+
+
+  is_control_plane = function(config)
+    return config.role == "control_plane"
+  end
+
+
+  is_dbless = function(config)
+    return config.database == "off"
+  end
+
+
+  is_dp_worker_process = function()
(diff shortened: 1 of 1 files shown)


kong/init.lua
```diff
@@ -179,6 +179,32 @@ do
...
+
+
+  is_dbless = function(config)
+    return config.database == "off"
+  end
+
+
+  is_dp_worker_process = function()
+    return process.type() == "privileged agent"
+  end
```
flrgh, line 204:
```suggestion
  is_dp_worker_process = function(config)
    return is_data_plane(config)
           and process.type() == "privileged agent"
  end
```

chronolaw: No, `dp_worker` is not only for role `data_plane`, but also for role `traditional`.

chronolaw: According the logic before, only data_plane will enable privileged process, so I think we need not do it.

chronolaw:
Due to #9860, this PR needs to rebase and check again.

kong/init.lua
chronolaw: #9860 removed privileged process, so I also removed this function.

StarlightIbuki (approved):

PR #9705, opened 2022-11-07
chore(tests): add test mutex for perf test

fffonion:


fffonion:
https://github.com/Kong/kong/actions/runs/3408611292/jobs/5669493248

dndx:
See: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Discussed with Wangchong offline. We should pin the dependency to a specific SHA to avoid supply chain attacks.

dndx (approved):

PR #9708, opened 2022-11-07
chore(declarative) add some additional documentation

Tieske:
Documents the _comment and _ignore fields in the declarative format. The diocumentation is added to the template that is used to create a base file when running `kong config init` in the cli.

### DIFF
--- a/kong/templates/kong_yml.lua
+++ b/kong/templates/kong_yml.lua
@@ -21,6 +21,16 @@ _format_version: "3.0"
 
 _transform: true
 
+# Custom annotations can be added via _comment and _ignore fields. The comments
+# must be strings, and the ignored fields must be an array, carrying any type as
+# values. These two can appear on the top-level of the file and on the top-level
+# of any entity.
+
+_comment: This is a top level comment, and must be a string
+_ignore:
+- This array entry will be ignored
+- as well as this one
+
 # Each Kong entity (core entity or custom entity introduced by a plugin)
 # can be listed in the top-level as an array of objects:
 


Tieske:
Wondering whether we need to suggest a proper way of using them.

The `_comment` is an end user one for sure, but `_ignore` could be used in a more systematic way, if we set up a convention for its use. Eg. any application going by `<name>` with version `<version>` that needs to store some annotations, would add an entry to the `_ignore` array.
```json
{
  "_ignore": [
    {
      "owner": "<name>",
      "owner_version": "<version>",
      "custom": "data goes here"
    }, {
      ... next application here ...
    }
  ]
}
```

hanshuebner (approved):

hbagdi:
> but _ignore could be used in a more systematic way, if we set up a convention for its use. Eg. any application going by <name> with version <version> that needs to store some annotations, would add an entry to the _ignore array.

_ignore does not need to be an array, simply stating that the key will be ignored so long as it is a valid JSON value should be sufficient.
Structuring the schema of _ignore should be out of scope if Kong is truly ignoring the key.

dndx:
@hbagdi @Tieske Does this comment needs more update?

kong/templates/kong_yml.lua
```diff
@@ -21,6 +21,16 @@ _format_version: "3.0"
 
 _transform: true
 
+# Custom annotations can be added via _comment and _ignore fields. The comments
+# must be strings, and the ignored fields must be an array, carrying any type as
+# values. These two can appear on the top-level of the file and on the top-level
+# of any entity.
```
lena-larionova, line 27:
```suggestion
# values.  _comment and _ignore fields can appear at the top level of the file 
# and at the top level of any entity.
```

lena-larionova (approved):
Minor phrasing edit for clarity, otherwise LGTM

Tieske:
> > but _ignore could be used in a more systematic way, if we set up a convention for its use. Eg. any application going by  with version  that needs to store some annotations, would add an entry to the _ignore array.
> 
> _ignore does not need to be an array, simply stating that the key will be ignored so long as it is a valid JSON value should be sufficient. Structuring the schema of _ignore should be out of scope if Kong is truly ignoring the key.

it has to be an array;
https://github.com/Kong/kong/blob/117191a284e0df323d0a5100c43b98a85b431c63/kong/db/schema/others/declarative_config.lua#L81-L85

kong/templates/kong_yml.lua
Tieske: thx! committed

Tieske:
> @hbagdi @Tieske Does this comment needs more update?

@dndx no it doesn't. Lena's comment has been comitted, and Harry's comment was not correct, it has to be an array. So this is good to go.

hbagdi:
Ah. I wonder why it has to be an array. Oh well, I can live with it.

dndx (approved):

PR #9710, opened 2022-11-07
test(statsd): if we're backwards compatible this spec test should pass

hutchic:


fffonion:
making request_size/response_size being a counter is the correct type moving forward. so instead of having this test (meaning we lost coverage of those metrics being a counter, which is the expected happy path), we should have
a unit test for schema, that validates request_size/response_size can be configured as both timer+sample_rate and counter.

PR #9714, opened 2022-11-08
feat(http-log): support http_endpoint field to be referenceable

vm-001:
### Summary

Support `http_endpoint` field to be referenceable

### Full changelog

* support `http_endpoint` field to be referenceable

### Issue reference

Internal ticket FTI-4483


vm-001:
I've not seen any vault test case for the individual plugin that uses referenceable field. So I assume that instead of adding the test case for that, Kong prefers using a single test case for vault feature(https://github.com/Kong/kong/blob/master/spec/02-integration/13-vaults/02-env_vault_spec.lua.(Correct me if I'm wrong). Thus,  this PR didn't contain test case for that.

fffonion:
Discussed with @vm-001 offline, `http_endpoint` doesn't make much sense to be a secrets and it doesn't
necessary meet the requirements in attached internal tickets. I would suggest to explore adding `referencable`
to http headers instead.

kikito (commented):
This change would need to be accompanied by one test

flrgh:
> Discussed with @vm-001 offline, `http_endpoint` doesn't make much sense to be a secrets and it doesn't necessary meet the requirements in attached internal tickets. I would suggest to explore adding `referencable` to http headers instead.

Removing from the 3.1 milestone per this comment. If something changes go ahead and add it back in.

vm-001:
> This change would need to be accompanied by one test

Test added. Please let me know if anything else required.

dndx:
I remember there were discussions that question if it makes sense to make this reference able (and general concern over making too many non-secret configs referenceable). 

@vm-001 do you remember the decision on this one?

PR #9731, opened 2022-11-08
feat(cmd): use a discrete script to check if kong is running

locao:
### Summary

This bash script checks if the Nginx master process needed by Kong is running. It provides the same functionality as `kong health`, but does not start a second Nginx for that. 

### Full changelog

* New bash script.
* Adapted current `kong health` tests.
* To do: obsolete `kong health`.


hbagdi:
We must be certain that all builds of Kong have bash as a dependency to move this forward.
Also, since we already have access to a Lua interpreter, why don't we write this code as a Lua script instead?

PR #9740, opened 2022-11-10
chore(clustering): remove legacy hybrid protocol

flrgh:
This PR removes the legacy hybrid protocol.

Some code was moved out of `kong.clustering.control_plane` into `kong.clustering.utils`, but aside from that, I have done as little refactoring as possible in order to keep the PR simple.

~**NOTE:** This is not yet an approved change for 3.1, so it is just a draft for now.~
**UPDATE:** This change is approved for 3.1, so this is ready for review and merge.


flrgh:
Reviewers: I'm aware this PR lacks a changelog entry and updates to the upgrade doc.

I have an internal ticket assigned to myself for those items and will open a follow-up PR for them.

chronolaw (approved):
It's great.

kong/clustering/init.lua
```diff
@@ -121,31 +110,8 @@ function _M:init_dp_worker(plugins_list)
...
-    else -- config_proto == "v1" or higher
-      data_plane = "kong.clustering.wrpc_data_plane"
-    end
-
-    self.child = require(data_plane).new(self.conf, self.cert, self.cert_key)
-
-    if self.child then
-      self.child:init_worker(plugins_list)
-    end
+    self.child = require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
```
kikito, line 113: `require`s at the top of the file, please (unless there's a really strong reason to do this here)

spec/02-integration/09-hybrid_mode/06-lagacy_switch_spec.lua
```diff
@@ -1,56 +0,0 @@
-local helpers = require "spec.helpers"
```
kikito, line 1: Glad to see this "lagacy switch" spec gone :)

kikito (approved):
approved with a minor comment

kong/clustering/init.lua
flrgh: Agreed, though I think there's an argument to be made about lazy instantiation since the DP module need not be loaded in all contexts (that's a discussion for another day) :P

The real answer is that this was just done to keep the diff small. I will leave a note to follow up with a code-tidying PR shortly :+1:

PR #9744, opened 2022-11-10
feat(api): dynamic log levels

gruceo:
Requested at FT-2818.


kong/api/routes/debug.lua
```diff
@@ -0,0 +1,102 @@
...
+    GET = function(self)
+      local sys_filter_level = get_sys_filter_level()
+      local cur_level = LOG_LEVELS[sys_filter_level]
+
+      if type(cur_level) ~= "string" then
+        local message = "unknown log level: " .. tostring(sys_filter_level)
+        return kong.response.exit(500, { message = message })
+      end
+
+      return kong.response.exit(200, { message = "log level: " .. cur_level })
```
fffonion, line 78: is it to return a structured result that `cur_level` is the value? i'm thinking of something like
`{ "data": { "message": "log level is now set to <cur_level>", "log-level": "<cur_level>"} }` so
it's easier to parse and expect. how do you think?

kong/api/routes/debug.lua
```diff
@@ -0,0 +1,102 @@
...
+    end,
+  },
+  ["/debug/node/log-level/:log_level"] = {
+    PUT = function(self)
+      return handle_put_log_level(self, false)
```
fffonion, line 83: for easier reading, i suggest to give `false`, `true` a name, like
```lua
NODE_LEVEL = false
CLUSTTER_LEVEL = true
-- then
handle_put_log_level(self, NODE_LEVEL)
handle_put_log_level(self, CLUSTTER_LEVEL)
```
or even give it a number

then later on you can even extend to send other scope of operations like
per sets of DPs, only on CP, etc

kong/api/routes/kong.lua
```diff
@@ -98,6 +100,9 @@ return {
         }
       end
 
+      local configuration = kong.configuration.remove_sensitive()
+      configuration.log_level = LOG_LEVELS[get_sys_filter_level()]
```
fffonion, line 104: not sure if it's a good idea to change this, as then it doesn't accuratly represent
what's in kong.conf. if we really want to see it in `:8001/`, i would suggest to
put it at top level

kong/runloop/handler.lua
```diff
@@ -1149,6 +1150,53 @@ return {
...
+          log(NOTICE, "log level worker event received")
+
+          local ok, err = pcall(set_log_level, data)
+
+          if not ok then
+            log(ERR, "[events] could not broadcast log level event: ", err)
+            return
+          end
+
+          log(NOTICE, "log level changed for worker")
```
fffonion, line 1196:
```suggestion
          log(NOTICE, "log level changed to ", data, " for worker ", ngx.worker.id())
```

kong/runloop/handler.lua
```diff
@@ -1149,6 +1150,53 @@ return {
...
+          end
+
+          local ok, err = kong.worker_events.post("debug", "log_level", tonumber(data))
+
+          if not ok then
+            kong.log.err("failed broadcasting to workers: ", err)
+            return
+          end
+
+          log(NOTICE, "log level changed for node")
```
fffonion, line 1182:
```suggestion
          log(NOTICE, "log level event posted for node")
```

kong/runloop/handler.lua
```diff
@@ -1149,6 +1150,53 @@ return {
...
+        -- log level cluster event updates
+        kong.cluster_events:subscribe("log_level", function(data)
+          log(NOTICE, "log level cluster event received")
+
+          if not data then
+            kong.log.err("received empty data in cluster_events subscription")
+            return
+          end
+
+          local ok, err = kong.worker_events.post("debug", "log_level", tonumber(data))
```
fffonion, line 1175: i vaguelly remember that worker_events doens't broadcast to itself, is it still true for the new
events system? if so we need to adhoc-ly set the log level for current worker.
(cc @chronolaw )

kong/templates/nginx_kong.lua
```diff
@@ -25,6 +25,7 @@ lua_shared_dict kong_core_db_cache          ${{MEM_CACHE_SIZE}};
 lua_shared_dict kong_core_db_cache_miss     12m;
 lua_shared_dict kong_db_cache               ${{MEM_CACHE_SIZE}};
 lua_shared_dict kong_db_cache_miss          12m;
+lua_shared_dict kong_log_level              1m;
```
fffonion, line 28: not trying to be stingy person as it's just 1m 😄 but why do we just use `kong`? since we are storing just one number or string.

kong/constants.lua
```diff
@@ -202,6 +202,25 @@ local constants = {
...
+  LOG_LEVELS = {
+    debug = ngx.DEBUG,
+    info = ngx.INFO,
+    notice = ngx.NOTICE,
+    warn = ngx.WARN,
+    error = ngx.ERR,
+    crit = ngx.CRIT,
+    alert = ngx.ALERT,
+    emerg = ngx.EMERG,
+    [ngx.DEBUG] = "debug",
```
fffonion, line 215: should this be 
```lua
debug = ngx.DEBUG,
[ngx.DEBUG] = ngx.DEBUG?
```
otherwise if user set log level to `"debug"`, it won't be same as `ngx.DEBUG` (which is `8`). the end result is not harmful, but it breaks the logic we did above that "don't call the API again
if current level is same as level to be set".

kong/pdk/log.lua
```diff
@@ -275,10 +275,11 @@ local function gen_log_func(lvl_const, imm_buf, to_string, stack_level, sep)
   to_string = to_string or tostring
   stack_level = stack_level or 2
 
-  local sys_log_level
   local variadic_buf = {}
 
   return function(...)
+    local sys_log_level = nil
+
```
fffonion, line 282: is adding the API to PDK at this scope?

spec/02-integration/04-admin_api/22-debug_spec.lua
```diff
@@ -0,0 +1,395 @@
...
+        })
+        body = assert.res_status(200, res)
+        json = cjson.decode(body)
+
+        message = "log level: crit"
+        return json.message == message
+      end, 30)
+    end)
+  end)
+end)
```
fffonion, line 394: we also need a test, probably just for a happy path only, that to ensure when we call
the API, the log is indeed printing message of given level. i know this is mostly
testing the underlying FFI is working, but we need to make sure it's e2e working
in kong.

kong/pdk/log.lua
gruceo: yes, dynamically changing to the log level should be reflected in the kong.log PDK

kong/templates/nginx_kong.lua
gruceo: I've tried to use it but the tests failed 😞

kong/api/routes/kong.lua
gruceo: if we don't change this, it won't accurately represent the dynamically changed (current) log level, it only show the kong.conf (old) log level

kong/constants.lua
gruceo: if user sets log level to `debug`, it will be `ngx.DEBUG` (8)

if current level is the same as the level to be set, the API will return:

```
200 OK
{
    "message": "log level is already notice"
}
```

spec/02-integration/04-admin_api/22-debug_spec.lua
gruceo: done: https://github.com/Kong/kong/blob/feat/FT-2818-dynamic_log_levels/spec/02-integration/04-admin_api/22-debug_spec.lua#L93-L141

kong/api/routes/debug.lua
gruceo: @fffonion I like your idea, could you please add this suggestion to the Design Doc? Changes to API design/contract/signature need to be discussed there

kong/api/routes/kong.lua
gruceo: @fffonion nonetheless, could you please add this suggestion to the Design Doc? we have agreed to change the `:8001/` endpoint to show current, up-to-date information about the log level considering the ones changed by the API proposed in this PR. So if we change this, we need to also discuss there.

kong/runloop/handler.lua
gruceo: it broadcasts to itself: https://github.com/Kong/lua-resty-worker-events#post

kong/api/routes/debug.lua
```diff
@@ -0,0 +1,105 @@
...
+    PUT = function(self)
+      return handle_put_log_level(self, NODE_LEVEL_BROADCAST)
+    end
+  },
+}
+
+local cluster_name
+
+if kong.configuration.role == "control_plane" then
+  cluster_name = "/debug/cluster/control-planes-nodes/log-level/:log_level"
```
tyler-ball, line 94: IMO we should not differentiate between these two route names. It seems confusing to me that a customer has to remember "oh if I am in hybrid mode, I need to PUT to a different URL". But I think we could re-address this when we are designing how to set logging level for data planes. Its hard to make the exact right API without knowing how that is going to work, and we can always deprecate this if we decide to do a different URL later.

kong/api/routes/debug.lua
tyler-ball: I have a theory that we're going to be redesigning this API when we add in data plane configuration. I'm guessing the API will begin accepting input like
```
{
  "log_level": "debug",
  "nodes": ["dataplane1", "dataplane2"],
  "timeout": 60
}
```
And I think at that point it also makes sense to return a structured response about what nodes had their log level updated

tyler-ball (approved):

spec/02-integration/04-admin_api/22-debug_spec.lua
```diff
@@ -0,0 +1,460 @@
...
+        body = assert.res_status(200, res)
+        json = cjson.decode(body)
+        message = "log level: debug"
+        return json.message == message
+      end, 30)
+
+      -- e2e test
+      helpers.clean_logfile()
+      assert.logfile().has.line("[debug]", true, 30)
+    end)
```
attenuation, line 141: Maybe this test can capture some error log, it will more complete

kong/api/routes/kong.lua
fffonion: This is fine, in stead of showing
```
{
  configuration: {
    log_level: notice,
  }
}
```
i'm suggesting to return something like 
```
{
  configuration: {
    log_level: notice,
  },
  log_level: debug, # maybe dynamic_log_level ? whatever makes clearer to you
}
```

kong/api/routes/debug.lua
```diff
@@ -0,0 +1,105 @@
...
+    ok, err = kong.cluster_events:broadcast("log_level", tostring(log_level))
+
+    if not ok then
+      local message = "failed broadcasting to cluster: " .. err
+      return kong.response.exit(500, { message = message })
+    end
+  end
+
+  -- store in shm so that newly spawned workers can update their log levels
+  ok, err = ngx.shared.kong_log_level:set("level", log_level)
```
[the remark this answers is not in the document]
fffonion: @attenuation we also need it to work in traditional mode, where lmdb db doesn't exists there. if
we remove traditional eventually we can indeed move it into lmdb. but right now let's keep it
in shm.

kong/api/routes/debug.lua
fffonion: if it's explictly write out (as i do see it in design doc), let's keep it as is for now. as Tyler mentioned
this may got new fields post config scoping, we can add our change there. i'm also adding this
 to design doc as well, but consider this not a blocker.

kong/api/routes/debug.lua
```diff
@@ -0,0 +1,105 @@
...
+
+      if type(cur_level) ~= "string" then
+        local message = "unknown log level: " .. tostring(sys_filter_level)
+        return kong.response.exit(500, { message = message })
+      end
+
+      return kong.response.exit(200, { message = "log level: " .. cur_level })
+    end,
+  },
+  ["/debug/node/log-level/:log_level"] = {
```
dndx, line 84: I don't like this interface design very much. It is easy for a node to miss the cluster event and keep running in verbose mode or Admin forgot to turn the log levels back to original. I think we should always include a timeout for changing the log levels dynamically at runtime, and give it a sensible default (e.g. 10 min).

kong/api/routes/debug.lua
```diff
@@ -0,0 +1,105 @@
...
+local kong = kong
+local pcall = pcall
+local type = type
+local tostring = tostring
+
+local NODE_LEVEL_BROADCAST = false
+local CLUSTER_LEVEL_BROADCAST = true
+
+local function handle_put_log_level(self, broadcast)
+  if kong.configuration.database == "off" then
```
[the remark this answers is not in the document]
gruceo: From the Design Doc, section "Phase 2 scope (future work)":

> Currently it’s not possible to dynamically change the log level on DB-less nodes because one of the must-have product requirements for the dynamic log level API is that it needs to be protected by RBAC and RBAC is not available in DB-less. Unix domain sockets or local ports is not an acceptable solution because if running a CLI like kong set-log-level is the only option, Kubernetes users can't easily run such CLI in a production Kubernetes cluster.

Please refer to FT-3550 from the Phase 2 Epic, there has been some suggestions in Phase 2 to introduce a new `debug_listen` for DB-less nodes and expose this API to that listener. Add your ideas about it there.

kong/api/routes/kong.lua
gruceo: Thanks for adding that to the Design Doc. Let's discuss there.

kong/api/routes/debug.lua
gruceo: Thanks for adding your thoughts to the Design Doc. Let's discuss there.

kong/api/routes/debug.lua
gruceo: Added to the Design Doc. Let's further discuss there.

spec/02-integration/04-admin_api/22-debug_spec.lua
gruceo: Good idea, done

kong/api/routes/debug.lua
```diff
@@ -0,0 +1,101 @@
...
+    local message = "cannot change log level when not using a database"
+    return kong.response.exit(405, { message = message })
+  end
+
+  local log_level = LOG_LEVELS[self.params.log_level]
+
+  if type(log_level) ~= "number" then
+    return kong.response.exit(400, { message = "unknown log level" })
```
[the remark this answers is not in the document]
gruceo: Good idea, done

kong/runloop/handler.lua
```diff
@@ -1149,6 +1150,53 @@ return {
...
+        end)
+
+        -- log level worker event updates
+        kong.worker_events.register(function(data)
+          log(NOTICE, "log level worker event received")
+
+          local ok, err = pcall(set_log_level, data)
+
+          if not ok then
+            log(ERR, "[events] could not broadcast log level event: ", err)
```
gszr, line 1192: This error message is inaccurate - the error refers to the `set_log_level` call.

kong/templates/nginx_kong.lua
gszr: > not trying to be stingy person as it's just 1m 😄 but why do we just use `kong`? since we are storing just one number or string.

Agreed here.

> I've tried to use it but the tests failed 😞
@gruceo do we know why the tests failed?

gszr (commented):
Mostly non-critical comments; the addition of the 1m shm only for storing the log_level seems the most concerning to me. Would be curious why we can't use the `kong` shm instead.

gszr:
> Feature freeze is today and I think this is feature complete. Murillo is going to look at addressing the shared dictionary issue and test failure in the next week as a bug fix to get addressed before code freeze. Merging this

Sounds good. @gruceo can address those in a separate PR.

kong/templates/nginx_kong.lua
gruceo: Found out why the tests failed. This shm is reset every time in the `init_by_lua` phase: https://github.com/Kong/kong/blob/master/kong/init.lua#L547

That's why this test fails: https://github.com/Kong/kong/blob/master/spec/02-integration/04-admin_api/22-debug_spec.lua#L459-L497

It does a `kong reload`, so the `init_by_lua` phase gets executed. To fix it, we'll need to find another way to write that test.

PR #9755, opened 2022-11-14
fix(lmdb): create lmdb directory with proper permission

chronolaw:
### Summary

Before kong starts, we create directory and files for LMDB, and set the proper permissions.

See FT-3534.

### Full changelog

* Add a function `pre_create_lmdb()`
* Create lmdb directory before checking nginx conf

### DIFF
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -381,7 +381,7 @@ end
 local function pre_create_lmdb(conf)
   local prefix = conf.prefix
   local lmdb_path = conf.lmdb_environment_path or "dbless.lmdb"
-  local user = string.match(conf.nginx_user, "%w+") or "$USER"
+  local user = string.match(conf.nginx_user or "", "%w+") or "$USER"
 
   local dir_name
   if lmdb_path:sub(1, 1) == "/" then
@@ -391,7 +391,6 @@ local function pre_create_lmdb(conf)
   end
 
   if pl_path.isdir(dir_name) then
-    log.verbose("lmdb directory %s exists", dir_name)
     return true
   end
 
@@ -399,8 +398,6 @@ local function pre_create_lmdb(conf)
   if not ok then
     return nil, "Can not create directory for LMDB " .. dir_name
   end
-
-  log.warn("create directory for LMDB ", dir_name)
 
   local cmds = {
     string.format("chown %s %s && chmod 0700 %s",
@@ -414,7 +411,6 @@ local function pre_create_lmdb(conf)
   }
 
   for _, cmd in ipairs(cmds) do
-    log.warn("xxx:%s", cmd)
     local ok, _, stdout, stderr = pl_utils.executeex(cmd)
     if not ok then
       return nil, stderr
@@ -655,6 +651,12 @@ local function prepare_prefix(kong_config, nginx_custom_template_path, skip_writ
     return true
   end
 
+  -- check lmdb directory
+  local ok, err = pre_create_lmdb(kong_config)
+  if not ok then
+    return nil, "check LMDB failed: " .. err
+  end
+
   -- compile Nginx configurations
   local nginx_template
   if nginx_custom_template_path then
@@ -697,8 +699,6 @@ local function prepare_prefix(kong_config, nginx_custom_template_path, skip_writ
     return nil, err
   end
   pl_file.write(kong_config.nginx_kong_stream_conf, nginx_kong_stream_conf)
-
-  pre_create_lmdb(kong_config)
 
   -- testing written NGINX conf
   local ok, err = nginx_signals.check_conf(kong_config)
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -378,6 +378,52 @@ local function write_process_secrets_file(path, data)
   return true
 end
 
+local function pre_create_lmdb(conf)
+  local prefix = conf.prefix
+  local lmdb_path = conf.lmdb_environment_path or "dbless.lmdb"
+  local user = string.match(conf.nginx_user, "%w+") or "$USER"
+
+  local dir_name
+  if lmdb_path:sub(1, 1) == "/" then
+    dir_name = lmdb_path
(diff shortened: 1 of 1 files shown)


dndx (commented):
We should add a test in: https://github.com/Kong/kong/blob/master/spec/02-integration/02-cmd/09-prepare_spec.lua

kong/cmd/utils/prefix_handler.lua
```diff
@@ -609,6 +651,12 @@ local function prepare_prefix(kong_config, nginx_custom_template_path, skip_writ
     return true
   end
 
+  -- check lmdb directory
+  local ok, err = pre_create_lmdb(kong_config)
```
kikito, line 656: Instead of "creating the file as root" (and then having to do all those `chmod`s), would it be possible for the privileged process to "wait and yield", locking until some other worker creates the files with the right permissions?

```
if process.type() == "privileged agent" then
  wait_for_someone_else_to_create_lmdb_file()
end
```

chronolaw: Perhaps we can not do it.

Because lua-resty-lmdb create the files in C land init_worker phase, it is earlier than ngx_lua.

spec/02-integration/02-cmd/09-prepare_spec.lua
```diff
@@ -65,6 +65,18 @@ describe("kong prepare", function()
...
 
+  it("prepares a directory for LMDB", function()
+    assert(helpers.kong_exec("prepare -c " .. helpers.test_conf_path ..
+                             " -p " .. TEST_PREFIX))
+    assert.truthy(helpers.path.exists(TEST_PREFIX))
+
+    local lmdb_data_path = helpers.path.join(TEST_PREFIX, "dbless.lmdb/data.mdb")
+    local lmdb_lock_path = helpers.path.join(TEST_PREFIX, "dbless.lmdb/lock.mdb")
+
+    assert.truthy(helpers.path.exists(lmdb_data_path))
```
dndx, line 76: Should check for the correct permission as well.

chronolaw: done

kong/cmd/utils/prefix_handler.lua
dndx: This is not a good idea, it introduces artificial race condition and is generally bad design IMO.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,49 @@ local function write_process_secrets_file(path, data)
...
 end
 
+local function pre_create_lmdb(conf)
+  local prefix = conf.prefix
+  local lmdb_path = conf.lmdb_environment_path or "dbless.lmdb"
+  local user = string.match(conf.nginx_user or "", "%w+") or "$USER"
+
+  local dir_name
+  if lmdb_path:sub(1, 1) == "/" then
+    dir_name = lmdb_path
```
dndx, line 388:
```suggestion
    dir_name = lmdb_path

```

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,49 @@ local function write_process_secrets_file(path, data)
...
+  else
+    dir_name = prefix .. "/" .. lmdb_path
+  end
+
+  if pl_path.isdir(dir_name) then
+    return true
+  end
+
```
dndx, line 396: Add a log line in here to aid debugging. Like:

"LMDB directory " .. dir_name .. " does not exist, pre-creating with the correct permissions"

kong/cmd/utils/prefix_handler.lua
chronolaw: done

kong/cmd/utils/prefix_handler.lua
chronolaw: done

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,54 @@ local function write_process_secrets_file(path, data)
...
+  else
+    dir_name = prefix .. "/" .. lmdb_path
+  end
+
+  if pl_path.isdir(dir_name) then
+    return true
+  end
+
+  log.info("LMDB directory '%s' does not exist, " ..
```
dndx, line 398:
```suggestion
  log.debug("LMDB directory '%s' does not exist, " ..
```

Log less by default, only log as `info` if it is something people should really care.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,54 @@ local function write_process_secrets_file(path, data)
...
+           "pre-creating with the correct permissions",
+           dir_name)
+
+  local ok, err = pl_path.mkdir(dir_name)
+  if not ok then
+    return nil, "Can not create directory for LMDB " .. dir_name ..
```
dndx, line 404:
```suggestion
    return nil, "can not create directory for LMDB " .. dir_name ..
```

We use lower case log messages (with the exception of "LMDB" which is a term itself).

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,54 @@ local function write_process_secrets_file(path, data)
   return true
 end
 
+local function pre_create_lmdb(conf)
+  local prefix = conf.prefix
+  local lmdb_path = conf.lmdb_environment_path or "dbless.lmdb"
+  local user = string.match(conf.nginx_user or "", "%w+") or "$USER"
```
chronolaw, line 384: Here I have a worry about the user name, can it works well in all conditions?

kong/cmd/utils/prefix_handler.lua
kikito: Alright then, thanks for considering

kong/cmd/utils/prefix_handler.lua
kikito: The only alternative I see is doing a hard error here, if we can't extract a user from `conf.nginx_user`.

kikito (approved):

PR #9760, opened 2022-11-15
fix(core): prefix path must be normalized

StarlightIbuki:
Example:

```json
{
	"paths": ["/%c3%A4"]
}
```

```json
{
	"fields": {
		"paths": [
			"not normalized path. Suggest: '/ä'"
		]
	},
	"code": 2,
	"name": "schema violation",
	"message": "schema violation (paths.1: not normalized path. Suggest: '/ä')"
}
```

fix FT-3359

### DIFF
--- a/kong/db/schema/typedefs.lua
+++ b/kong/db/schema/typedefs.lua
@@ -452,7 +452,7 @@ local function validate_path_with_regexes(path)
     -- prefix matching. let's check if it's normalized form
     local normalized = normalize(path, true)
     if path ~= normalized then
-      return nil, "not normalized path. Suggest: '" .. normalized .. "'"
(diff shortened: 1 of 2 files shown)


kong/db/schema/typedefs.lua
```diff
@@ -448,6 +449,12 @@ local function validate_path_with_regexes(path)
   end
 
   if path:sub(1, 1) ~= "~" then
+    -- prefix matching. let's check if it's normalized form
+    local normalized = normalize(path, true)
+    if path ~= normalized then
+      return nil, "not normalized path. Suggest: '" .. normalized .. "'"
```
dndx, line 455:
```suggestion
      return nil, "non-normalized path, consider use '" .. normalized .. "' instead"
```

Issue #3890, opened 2018-10-23
Can't get loggly runs on docker kong (enterprise)


timpamungkasudemy:
### Summary
Can't get kong logs transferred to loggly, after enabling loggly plugin.

### Steps To Reproduce
1. Create loggly account (free tier)
2. Add plugins kong-loggly with `config.host` to `[myhost].loggly.com` and correct secret key, for all requests 
3. try to hit API via kong ( I hit mockbin)
4. loggly is empty

### Additional Details & Logs
- Kong version (`0.33.1`)
- Kong enterprise trial, running on docker with following command
`docker run -d --name kong-ee --link kong-ee-database:kong-ee-database -e "KONG_DATABASE=postgres" -e "KONG_PG_HOST=kong-ee-database" -e "KONG_CASSANDRA_CONTACT_POINTS=kong-ee-database" -e "KONG_PROXY_ACCESS_LOG=/dev/stdout" -e "KONG_ADMIN_ACCESS_LOG=/dev/stdout" -e "KONG_PROXY_ERROR_LOG=/dev/stderr" -e "KONG_ADMIN_ERROR_LOG=/dev/stderr" -e "KONG_VITALS=on" -e "KONG_ADMIN_LISTEN=0.0.0.0:8001" -e "KONG_PORTAL=on" --volume /d/development/kong/license.json:/etc/kong/license.json -p 8000:8000 -p 8443:8443 -p 8001:8001 -p 8444:8444 -p 8002:8002 -p 8445:8445 -p 8003:8003 -p 8004:8004 -p 53160:53160/udp kong-ee`
- plugin config : 
```json
{ "host": "xxxxx.loggly.com", "timeout": 10000, "tags": [ "kong" ], "successful_severity": "info", "log_level": "info", "port": 514, "client_errors_severity": "info", "key": "xxxxx", "server_errors_severity": "info" }
```

hbagdi:
It seems like you're using Kong Enterprise. Please direct your questions to Enterprise Support channel at https://support.konghq.com

PR #9768, opened 2022-11-16
feat(pdk): extend `kong.client.tls.request_client_certificate`

catbro666:
extend this API by adding an optional argument `ca_certs` to support setting the CA certificate chain.

Under the hood, switch from `resty.kong.tls.request_client_certificate` to `ngx.ssl.verify_client`

This PR is relied by [kong-ee #3890](https://github.com/Kong/kong-ee/pull/3890)

FTI-4430 FT-3584


kong/pdk/client/tls.lua
```diff
@@ -45,22 +46,42 @@ local function new()
...
+  -- -- err check
+  -- res, err = chain:add(x509)
+  -- -- err check
+  -- -- `chain.ctx` is the raw data of the chain, i.e. `STACK_OF(X509) *`
+  -- res, err = kong.client.tls.request_client_certificate(chain.ctx, 3)
   -- if not res then
   --   -- do something with err
   -- end
-  function _TLS.request_client_certificate()
+  function _TLS.request_client_certificate(ca_certs, depth)
```
dndx, line 81: We don't need `depth` for `mtls-auth` to work properly right? Suggest to remove it until we do have a use for it.

Also, we can not change the meaning of the return value without introducing a breaking change. Please make sure the changes to `@treturn` are intended.

catbro666: > We don't need depth for mtls-auth to work properly right? Suggest to remove it until we do have a use for it.
Yes, we can set a minimum default value.

> Also, we can not change the meaning of the return value without introducing a breaking change. Please make sure the changes to @treturn are intended.
The original description is not so accurate.  The handshake is not successful yet, only the function call is successful.

dndx: For our current purpose, we can even set `depth` to `0` so that verification can stop early on. We don't need to actually need to verify anything by Nginx and it just wastes CPU cycles.

catbro666: Totally agree.

CHANGELOG.md
```diff
@@ -121,6 +121,12 @@
   when there are too many regex routes and `router_flavor` is `traditional`.
   [#9624](https://github.com/Kong/kong/pull/9624)
 
+#### PDK
+
+- Extend `kong.client.tls.request_client_certificate` to support set CA certificate
```
dndx, line 126: This is not accurate, we should call this "CA list hints". Otherwise it makes it sounds like the CA list is used for verification purpose.

kong/pdk/client/tls.lua
```diff
@@ -45,22 +46,39 @@ local function new()
   -- To find out whether the client honored the request, use
   -- `get_full_client_certificate_chain` in later phases.
   --
+  -- The `ca_certs` argument is the CA certificate chain opaque pointer (if any),
```
dndx, line 49:
```suggestion
  -- The `ca_certs` argument is the optional CA certificate chain opaque pointer,
```

kong/pdk/client/tls.lua
```diff
@@ -45,22 +46,39 @@ local function new()
...
+  -- which can be created by the [parse_pem_cert](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/ssl.md#parse_pem_cert)
+  -- or [resty.opensslx509.chain](https://github.com/fffonion/lua-resty-openssl#restyopensslx509chain)
+  -- The list of certificates will be sent to clients. Also, they will be added
+  -- to trusted store. If omitted, will not send any CA certificate to clients.
```
dndx, line 53: Is the trusted store even used for our implementation? If not we should remove mentioning of it to avoid confusions.

CHANGELOG.md
catbro666: Indeed. Done.

kong/pdk/client/tls.lua
catbro666: Good idea! Done.

dndx (approved):

PR #9772, opened 2022-11-17
fix(db/lmdb): pre-create lmdb directory with proper permission

chronolaw:
### Summary

This is the update for #9755.

The point is what is the right user for lmdb's directory.


kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,54 @@ local function write_process_secrets_file(path, data)
   return true
 end
 
+local function pre_create_lmdb(conf)
+  local prefix = conf.prefix
+  local lmdb_path = conf.lmdb_environment_path or "dbless.lmdb"
+  local user = string.match(conf.nginx_user or "", "%w+") or "`whoami`"
```
dndx, line 384: This is probably not right. It will make the directory owned by `root`, which is not what we want.

Why is `conf.nginx_user` empty in the first place, we have default value set for these.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,54 @@ local function write_process_secrets_file(path, data)
...
+                  dir_name .. "/data.mdb", dir_name .. "/data.mdb"),
+
+    string.format("touch %s && chmod 0600 %s",
+                  dir_name .. "/lock.mdb", dir_name .. "/lock.mdb"),
+  }
+
+  for _, cmd in ipairs(cmds) do
+    local ok, _, _, stderr = pl_utils.executeex(cmd)
+    if not ok then
+      return nil, stderr
```
dndx, line 422: This should have a better error message, something like:

`return nil, "unable to set correct permissions for LMDB files: " .. stderr`

Also, we should probably only log an error when these command fails. We can let Kong to continue starting and if these permissions are not working, the worker will complain anyway.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -609,6 +657,12 @@ local function prepare_prefix(kong_config, nginx_custom_template_path, skip_writ
     return true
   end
 
+  -- check lmdb directory
+  local ok, err = pre_create_lmdb(kong_config)
+  if not ok then
+    return nil, "check LMDB failed: " .. err
```
dndx, line 663:
```suggestion
    return nil, "unable to verify the LMDB directories has correct permissions: " .. err
```

dndx (commented):
We need a test case for this new bug as well.

kong/cmd/utils/prefix_handler.lua
chronolaw: Should we give it a default value `"kong"`?

dndx: I am wondering why did we not see the default value from `kong.configuration` in the first place.

chronolaw: I checked the code of `conf_loader`, it is set to `nil`.

https://github.com/Kong/kong/blob/cb5e0460709ea7d115029ec3711158cdcd908c7a/kong/conf_loader/init.lua#L1579

kong/cmd/utils/prefix_handler.lua
chronolaw: if `pre_create_lmdb` fails, we use `log.warn` and continue the workflow.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -609,6 +662,13 @@ local function prepare_prefix(kong_config, nginx_custom_template_path, skip_writ
     return true
   end
 
+  -- check lmdb directory
+  local ok, err = pre_create_lmdb(kong_config)
+  if not ok then
+    log.debug("unable to verify the LMDB directory has correct permissions: %s",
```
dndx, line 668: This should not be debug level, as it is clearly an error the user should know about. Suggest `log.error`.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,59 @@ local function write_process_secrets_file(path, data)
...
+  end
+
+  if pl_path.isdir(dir_name) then
+    return true
+  end
+
+  local user = string.match(conf.nginx_user or "", "%w+")
+
+  if not user then
+    return nil, "no proper user for LMDB files"
```
dndx, line 400:
```suggestion
    return nil, "nginx_user is not set in Kong config, skipping verification of LMDB file permissions"
```

chronolaw:
> We need a test case for this new bug as well.

I think this test case is still a good case. 

In my local environment there is no user `kong`, and `conf.nginx_user` is `nil`,
so if we do not give the default value for `chown`, it will fail.

kong/cmd/utils/prefix_handler.lua
chronolaw: done

kong/cmd/utils/prefix_handler.lua
chronolaw: done

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,60 @@ local function write_process_secrets_file(path, data)
...
+  end
+
+  if pl_path.isdir(dir_name) then
+    return true
+  end
+
+  local user = string.match(conf.nginx_user or "", "%w+")
+
+  if not user then
+    return nil, "nginx_user is not set in Kong config, "..
```
dndx, line 400: This should not be an error, a `log.warn` followed by `return true` should be sufficient.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +378,60 @@ local function write_process_secrets_file(path, data)
...
+            dir_name)
+
+  local ok, err = pl_path.mkdir(dir_name)
+  if not ok then
+    return nil, "can not create directory for LMDB " .. dir_name ..
+                ", err: " .. err
+  end
+
+  local cmds = {
+    string.format("chown %s %s && chmod 0700 %s",
```
dndx, line 415: The group information from `nginx_user` is important, we should `chown user:group path` here.

kong/cmd/utils/prefix_handler.lua
chronolaw: done

kong/cmd/utils/prefix_handler.lua
chronolaw: done

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +381,67 @@ local function write_process_secrets_file(path, data)
...
+  local cmds = {
+    format("chown %s:%s %s && chmod 0700 %s",
+           user, group, dir_name, dir_name),
+
+    format("touch %s && chmod 0600 %s",
+           dir_name .. "/data.mdb", dir_name .. "/data.mdb"),
+
+    format("touch %s && chmod 0600 %s",
+           dir_name .. "/lock.mdb", dir_name .. "/lock.mdb"),
+  }
```
hutchic, line 433: what happens on systems that don't have these commands available?

Would this be better handled by [after-install.sh](https://github.com/Kong/kong-build-tools/blob/master/after-install.sh) ?

kong/cmd/utils/prefix_handler.lua
```diff
@@ -378,6 +381,67 @@ local function write_process_secrets_file(path, data)
...
+            dir_name)
+
+  local ok, err = pl_path.mkdir(dir_name)
+  if not ok then
+    return nil, "can not create directory for LMDB " .. dir_name ..
+                ", err: " .. err
+  end
+
+  local cmds = {
+    format("chown %s:%s %s && chmod 0700 %s",
```
flrgh, line 425: The usage of `command && command` seems a little unnecessary given that the loop below fulfills the same purpose by exiting early after any command fails.

kong/cmd/utils/prefix_handler.lua
flrgh: I think for many cases after-install.sh _would_ be sufficient, but this is in $KONG_PREFIX, so its location can't really be 100% known at install time:

```shell
docker run \
  -e KONG_PREFIX=/foo \
  -v "$PWD:/foo" \
  my-kong-image
```

I agree on the principle though. We should only resort to shell/exec when absolutely necessary. Lua `io.*`, [LFS](https://lunarmodules.github.io/luafilesystem/manual.html), or FFI are all available to use.

The [file-log handler](https://github.com/Kong/kong/blob/8fda66e41db3399b345f2d211011d39ea4b09035/kong/plugins/file-log/handler.lua) has an example that can be followed.

For cases where we have to resort to shell, [resty.shell](https://github.com/openresty/lua-resty-shell) should be used because it eliminates the need for escaping/quoting arguments:

```lua
-- this is error-prone and unsafe
pl_utils.executeex(string.format("touch %s", "filename with spaces.txt"))
pl_utils.executeex(string.format("touch %s", "file.txt; rm -rf /"))

-- this is better
resty_shell.run({ "touch", "filename with spaces.txt" })
resty_shell.run({ "touch", "file.txt; rm -rf /" })
```

chronolaw: Here I just follow the other test case's pattern, should we change them to `resty_shell`?

flrgh: `resty.shell` is probably fine

RobSerafini:
@dndx - Did we miss 3.1 code freeze for this or is it a bug fix?

kong/cmd/utils/prefix_handler.lua
fffonion: does it make sense to use approaches like this (found in our private fork) to get rid of shell commands completely?

```lua
     # 2.7.1.2: kong/clustering/data_plane.lua#L191
      -- CONFIG_CACHE does not exist, pre create one with 0600 permission
      local fd = ffi.C.open(self.config_cache, bit.bor(system_constants.O_RDONLY(),
                                                       system_constants.O_CREAT()),
                                               bit.bor(system_constants.S_IRUSR(),
                                                       system_constants.S_IWUSR()))
      if fd == -1 then
        ngx_log(ngx_ERR, _log_prefix, "unable to pre-create cached config file: ",
                ffi.string(ffi.C.strerror(ffi.errno())))

      else
        ffi.C.close(fd)
      end
```

tyler-ball:
I know this missed 3.1.0 but I think its an important security issue, so I created a 3.1.1 milestone and added it to that, to ensure this does not get missed

chronolaw:
Due to #9860, I think the permission issue will not happen again.

bungle:
> Due to #9860, I think the permission issue will not happen again.

Unless we reintroduce or make a technology preview or something for those that want it (or make in configurable or something).

PR #9778, opened 2022-11-17
chore(rockspec): bump lua-resty-healthcheck to 1.6.2

locao:
### Summary

The lua-resty-healthcheck feature used to keep the health status when an existent target is updated was raising more events than needed. The issue was fixed in the healthcheck module and bumped here.

### Full changelog

* lua-resty-healthcheck bump.
* Same behavior as before, no new tests.

FTI-4353


chronolaw (approved):

chronolaw:
Please add a changelog entry.

PR #9790, opened 2022-11-21
fix(clustering): override node id while it contains an invalid uuid

vm-001:
### Summary

- Uses `kong.cmd.utils.log` instead of `ngx.log` as the latter is unavailable in cmd context.
- Fixes the existing node id file with an invalid uuid won't be rewritten after Kong start.
- Enables the `pending` test

### Checklist

- [x] There's an entry in the CHANGELOG (This is a fix for the original PR)
- [x] There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE (This is a fix and does not require user-facing docs update)

### DIFF
--- a/kong/pdk/private/node.lua
+++ b/kong/pdk/private/node.lua
@@ -30,6 +30,7 @@ local function initialize_node_id(prefix)
     if err then
       return nil, fmt("failed to access file %s: %s", filename, err)
     end
+
     if not utils.is_valid_uuid(id) then
       log.debug("file %s contains invalid uuid: %s", filename, id)
       -- set false to override it when it contains an invalid uuid.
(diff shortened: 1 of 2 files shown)


kong/pdk/private/node.lua
```diff
@@ -23,9 +23,23 @@ local function initialize_node_id(prefix)
...
   local filename = node_id_filename(prefix)
 
-  if not pl_path.exists(filename) then
+  local file_exists = pl_path.exists(filename)
+
+  if file_exists then
+    local id, err = pl_file.read(filename)
+    if err then
+      return nil, fmt("failed to access file %s: %s", filename, err)
+    end
```
dndx, line 32:
```suggestion
    end

```

fffonion (approved):

PR #9792, opened 2022-11-21
chore(ci): setup a backport github action

hutchic:
Commits with a label of `backport branch_name` will get a backport'd PR automatically opened and linked to the origin PR

image

image

If we don't care about CI we could use `GITHUB_TOKEN` otherwise I think we require a `PAT` for it to run


flrgh (approved):
:+1: nifty! Looks good to me, but I think we'll wanna wait for another review or two before merging.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,24 @@
...
+      - labeled
+
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-latest
+    if: >
+      github.event.pull_request.merged
+      && (
+        github.event.action == 'closed'
```
gszr, line 15: Is this condition required here? If `merged` is true, then it was closed. Also, the `||` leads me to think it will apply to more PRs than only those with the tag.

hutchic: It's tricky because we have to handle multiple labels on the same PR.

Very rough pseudo code
```
if merged {
   if closed { return true }
   if labelled { return true }
}
return false
```

gszr: Shouldn’t it be only 
```
if merged {
   if labelled { return true }
}
return false
```

hutchic: no because that won't catch PR's that were, opened, then labelled, then closed

.github/workflows/backport.yml
```diff
@@ -0,0 +1,24 @@
...
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'backport')
+        )
+      )
+    steps:
+      - uses: tibdex/backport@v2
```
fffonion, line 22: for third party actions, let's pin to it's sha to avoid potential security issue
```suggestion
      # - uses: tibdex/backport@v2
      - uses: tibdex/backport@2e217641d82d02ba0603f46b1aeedefb258890ac
```

PR #9794, opened 2022-11-21
fix(api): various minor fixes and optimizations for dynamic log levels

gruceo:
Use `kong` shm instead of its own dedicated `kong_log_level`.


gszr (approved):
💯

PR #9798, opened 2022-11-21
chore(ci): setup a job to sync commits to our private fork

hutchic:
!!!! Do a manual sync before merging this !!!!

Sample successful run on [my fork](https://github.com/Kong/colin-kong/actions/runs/3517866013/jobs/5896127392)

Every closed PR that is merged to `master` will create a corresponding branch `sync/$BRANCH_NAME` and a pull request on our private fork.

Initial PR:
image

Mirror'd PR:
image

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,0 +1,51 @@
+name: Sync
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - 'master'
+
+jobs:
+  sync:
+    name: Sync to Private Fork
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout Kong source code
+      uses: actions/checkout@v3
+    - name: Checkout Kong EE source code
+      uses: actions/checkout@v3
+      with:
+        repository: Kong/kong-ee
+        path: kong-ee
+        ref: master
+        token: ${{ secrets.PAT }}
+    - name: Setup github config
+      run: |
+        git config --global user.email "team-eng-enablement+github-bot@konghq.com"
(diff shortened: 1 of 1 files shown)


flrgh:
(I'm mostly thinking out loud here--this is not necessarily a problem we care about solving)

Here is a sequence of events that is unfortunately not too uncommon:

1. I merge a change in OSS
2. Oh no my OSS change broke something in EE
3. I found the problem, but I need to patch OSS code from step `#1` to fix it

What happens in this case? I'm guessing that the PR against EE will be blocked until:

1. I merge the fix in OSS
2. I merge the fix into EE (via automated commit sync action)
3. I rebase the "original" automated EE PR onto master

Some of this could be less painful if the process were synchronous: opening a PR against OSS causes one to be opened against EE _immediately_, so that I can see the test results for both repos. Is this idea completely insane? Would it make things too convoluted to be worth it?

hutchic:
It's not an insane idea it's what I initially wanted and designed  but

>  opening a PR against OSS causes one to be opened against EE immediately

raises security implications we can't easily solve :(. In the spirit of "tighter feedback loops" + "crawl, walk, then run" I think this moves the needle incrementally in the proper direction.

tyler-ball:
@hutchic In your example the bot used your user to open the EE PR. Is that how the bot will function or did it use your user because that was hardcoded, and it will use some bot user when we merge this PR?

tyler-ball:
I see it is going to use a bot. I would *love* if the EE PR could ping the CE PR merger but I don't see any [default environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) to fetch that user. I'm sure it could be done with the Github API, but I agree we should crawl first. I think this is a great start

PR #9799, opened 2022-11-22
feat(aws-lambda): support scheme config on lambda service api endpoint

### LINKED ISSUE #9455: Kong cannot integrate with AWS Lambdas over HTTP

2.8.1, 3.0.0

### Current Behavior

Kong hard-codes the URL generation for the endpoint to use when contacting AWS Lambda functions in the Kong AWS Lambda Plugin.

https://github.com/Kong/kong/blob/master/kong/plugins/aws-lambda/handler.lua#L294

In a [recent, related PR](https://github.com/Kong/kong/pull/8406), it was said:

> since the request to AWS lambda is always https

But this is not strictly true, when AWS lambda is run locally through the AWS SAM start-lambda command, a thin HTTP layer (not an AWS Gateway API layer) is stood up in front of the lambdas so they can be hit locally. As far as I can tell, there is no way through AWS SAM parameters to start this endpoint up over HTTPS, it always uses HTTP locally.

In order to get around this, we've configured our lambda routes in our declarative kong.yml config to point back to Kong itself to perform SSL termination. This config is specifically for Kong 2.8.1 but I also tried Kong 3.0.0 with minor updates to handle the AWS host/region change and saw similar results. 

```
  - name: myLambdaFunction
    paths: [ "/" ]
    methods: [ "GET" ]
    plugins:
    - name: aws-lambda
      config:
        invocation_type: RequestResponse
        aws_key: DUMMY_KEY
        aws_secret: DUMMY_SECRET
        log_type: None
        timeout: 60000
        keepalive: 60000
        function_name: myLambdaFunction
        host: localhost
        port: 8443
        forward_request_body: true
        forward_request_headers: true
        forward_request_method: true
        forward_request_uri: true
        is_proxy_integration: true
```

Note the port used is 8443 which is Kong itself. Then, we have Kong catch the request using an AWS lambda header and perform the actual request to our local AWS Lambda HTTP endpoint running on port 3001:

```
services:
- name: lambda-proxy
  url: http://host.docker.internal:3001/
  routes:
  - name: lambda_proxy
    path_handling: "v0"
    strip_path: true
    headers:
      content-type: ["application/x-amz-json-1.1"]
```

Not to conflate this issue, but I'm also dealing with a problem I haven't been able to solve yet that may be related to the way we're implementing this work-around. Somehow, Kong is not generating the X-Forwarded* headers when it proxies these requests. Our Lambda function depends on X-Forwarded-Proto and fails requests (though I can see that none of the X-Forwarded* headers are generated by Kong, not just X-Forwarded-Proto). I'm hoping that by allowing the plugin to natively use HTTP and removing our work-around will address that issue. We've confirmed that X-Forwarded* headers are properly generated when Kong is deployed in AWS. This header is not related to trusted IPs. Our work-around for *this* problem is to put the X-Forwarded-Proto header on our initial requests in development and Kong does correctly forward the header, it just doesn't generate it itself or the generated header is getting dropped somewhere inside Kong.

### Expected Behavior

Kong should be able to talk directly to a Lambda running locally through `sam local start-lambda` without requiring work-arounds to handle the hard-coded HTTPS URI that Kong constructs.

### Steps To Reproduce

1. Spin up Kong on a docker network
2. Define a Lambda + SAM template definition with AWS SAM's template.yaml to facilitate local development
3. Start a Lambda HTTP endpoint locally using:
```
sam local start-lambda --docker-network <YOUR DOCKER NETWORK RUNNING KONG>
```
4. Configure Kong with a route/service to point to your Lambda
5. Kong will exclusively generate SSL URIs for contacting the Lambda and the request will fail.

windmgc:
### Summary

This PR adds a new boolean type config field `https` to support the AWS-lambda plugin to connect to an HTTP-scheme lambda service API endpoint. This is useful for local test scenarios by using the [AWS SAM CLI tool](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/what-is-sam.html).

A related test case is also added in this PR, including setup steps for SAM local service and testing through the AWS-lambda plugin.

Compared with mocking servers, a SAM local lambda service is a better choice for doing integration tests(more like a real lambda service), so we can consider moving some of the integration tests to use SAM in the future.

### Checklist

- [x] There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - [docs(aws-lambda): add new https config field in aws-lambda plugin](https://github.com/Kong/docs.konghq.com/pull/4877)

### Full changelog

* Add `https` config field to the AWS-lambda plugin
* Add SAM related integration test

### Issue reference

Fix #9455 and [FTI-4484](https://konghq.atlassian.net/browse/FTI-4484)

[FTI-4484]: https://konghq.atlassian.net/browse/FTI-4484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


windmgc:
There is no need to add this in 3.1, so the changelog will be added later.

windmgc:
```
------- Running tests from spec/03-plugins/27-aws-lambda/08-sam-integration_spec.lua :
14590.12   OK 78: Plugin: AWS Lambda with SAM local lambda service [#postgres] with local HTTP endpoint invoke a simple function
------- 1 test from spec/03-plugins/27-aws-lambda/08-sam-integration_spec.lua (27694.50 ms total)
```

A fresh start and test running will cost about half a minute

mayocream (approved):
Let's hold this for 3.2

windmgc:
Changelog entry added.

mayocream (approved):

kong/plugins/aws-lambda/schema.lua
```diff
@@ -58,6 +58,7 @@ return {
         } },
         { host = typedefs.host },
         { port = typedefs.port { default = 443 }, },
+        { https = { type = "boolean", default = true }, },
```
hbagdi, line 61: I recommend reconsidering the true by default design here and instead renaming the field to `disable_https` with a default value of `false`.
If someone is running a version of Kong before this patch, and upgrades, `config.https` will not be present in the database (since there are no migrations) and this value will be `null`(?). This is generally not a problem in DB-less mode where missing defaults are injected but the same cannot be said with confidence about Kong Manager interactions and traditional mode deployments.
I also generally discourage patches that add 'true by default' design since they add noise to the configuration for the users. More so for configuration options that aren't going to be used much and are a security risk if disabled.

What do you think?

windmgc: Initially, I chose the way to add `https` field because I want to keep a similar schema field like [the Azure serverless function plugin](https://docs.konghq.com/hub/kong-inc/azure-functions/).

> This is generally not a problem in DB-less mode where missing defaults are injected but the same cannot be said with confidence about Kong Manager interactions and traditional mode deployments.

I think this would generally also not be a problem in Kong Manager and traditional mode. Such new schema fields are not real database columns, they're only JSON fields inside the database and if they did not exist before the upgrade, and once upgraded, when the plugin entity is selected from the database, the missing fields will be filled automatically with the default value(I think `handle_missing_field` is doing the job here)

windmgc: Discussed offline, and to avoid further confusion I renamed the field to `disable_https`.

Please note that changing this field does not change port number and currently, we do not have a proper & friendly way to dynamically set port value according to the https related field. So I'll add some notice to the document and we may consider simplifying the endpoint-related configs(host, port, scheme, etc.) to one field `endpoint` in the future.

hbagdi:
@windmgc This seems close. Are you blocked on code review only?

windmgc:
It has been a while; let me update the PR to the latest state

I'm gonna remove the installation from the CI lua file to GitHub Actions so that we can avoid installing SAM tools on developers who don't want them locally and we can also have it in our CI environment. I think it would be more friendly for community developers.

windmgc:
@kikito  Rebase done!

spec/03-plugins/27-aws-lambda/08-sam-integration_spec.lua
```diff
@@ -0,0 +1,101 @@
...
+  local arch_type = sam.get_os_architecture()
+  local is_sam_installed, _ = sam.is_sam_installed()
+  if arch_type ~= "aarch64" and is_sam_installed then
+    sam_describe = describe
+  else
+    sam_describe = pending
+  end
+end
+
+if sam.get_os_architecture() ~= "aarch64" then
```
hbagdi, line 16: Can you add a comment explaining why this is required?

hanshuebner (approved):

hanshuebner (approved):

hanshuebner (approved):

spec/03-plugins/27-aws-lambda/08-sam-integration_spec.lua
windmgc: Comment added.

windmgc:
@hanshuebner @kikito Is it possible to merge this soon or is there any other concern left? Please take a look again and merge this if possible, thanks!

hanshuebner (approved):

PR #9800, opened 2022-11-22
fix(router): emit error for route flavor misconfig

StarlightIbuki:
It will print error logs when users configured the gateway with atc routers and change the flavor to traditional(\_compatible) or verse visa.

Fix FT-3253

### DIFF
--- a/kong/router/compat.lua
+++ b/kong/router/compat.lua
@@ -28,6 +28,7 @@ local bor, band, lshift = bit.bor, bit.band, bit.lshift
 local ngx       = ngx
 local ngx_log   = ngx.log
 local ngx_WARN  = ngx.WARN
+local ngx_ERR  = ngx.ERR
 
 
 local DOT              = byte(".")
@@ -282,6 +283,11 @@ end
 
 
 local function get_exp_and_priority(route)
+  if route.expression then
+    ngx_log(ngx_ERR, "expecting a traditional route while expression is given.",
+                 "Likely it's a misconfiguration, please check router_flavor.")
+  end
+
   local exp      = get_expression(route)
   local priority = get_priority(route)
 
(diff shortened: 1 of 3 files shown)


kong/router/compat.lua
```diff
@@ -282,6 +283,11 @@ end
 
 
 local function get_exp_and_priority(route)
+  if route.expression then
+    ngx_log(ngx_ERR, "expecting a traditional route while expression is given. ",
```
chronolaw, line 287: `ngx.ERR` may be a little high, could we change it to `ngx.WARN`?

StarlightIbuki: It won't match any routes if this happens, and it should be handled by the administrator immediately. Thus it should be an error.

StarlightIbuki:
@dndx should we merge this?

mayocream (commented):
Validate and throw the error when configuring through admin api/config could be better?

StarlightIbuki:
> Validate and throw the error when configuring through admin api/config could be better?

We do have this validation. It's possible that a user migrates and misconfigured the flavor, and that cannot be caught by admin API.

flrgh:
Haven't looked at either too closely, but it looks like #10089 might serve common goal. Is there any worth to sharing code between the two or no?

StarlightIbuki:
> Haven't looked at either too closely, but it looks like #10089 might serve common goal. Is there any worth to sharing code between the two or no?

Actually, they are for different purposes. This PR is to emit a warning when a user added expression routers and then restarts with traditional flavor, or vice versa. Chrono's PR is trying to filter out those traditional routes that cannot be executed in compatible mode.

hanshuebner (approved):

PR #9801, opened 2022-11-22
fix(key-auth) expire keys in L1/L2 caches when ttl is in [0, 1)

outsinre:
### Summary

The postgres DAO return 0 when key ttl falls between `[0, 1)`, which would make the key never expires in L1/L2 cache.

This fix will check if the returned ttl is 0. If it is, then make it `nil`.

<https://konghq.atlassian.net/browse/FTI-4066>

### Checklist

- [ y ] The Pull Request has tests

### Issue reference

Fix [FTI-4066](https://konghq.atlassian.net/browse/FTI-4066)

### DIFF
--- a/kong/pdk/cluster.lua
+++ b/kong/pdk/cluster.lua
@@ -46,9 +46,13 @@ local function new(self)
   --
   -- -- use id here
   function _CLUSTER.get_id()
-    return kong.core_cache:get(CLUSTER_ID_PARAM_KEY, nil, fetch_cluster_id)
-  end
+    local cluster_id, err = kong.core_cache:get(CLUSTER_ID_PARAM_KEY, nil, fetch_cluster_id)
+    if err then
+      return nil, err
+    end
 
+    return cluster_id
+  end
 
   return _CLUSTER
 end
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -26,8 +26,9 @@ local function load_credential(key)
   end
 
   if cred.ttl == 0 then
-    cred = { err = "credential ttl is 0" }
-    return cred, err
+    kong.log.debug("key expired")
+
+    return nil
   end
 
   return cred, nil, cred.ttl
@@ -159,19 +160,18 @@ local function do_authentication(conf)
   local cache = kong.cache
 
   local credential_cache_key = kong.db.keyauth_credentials:cache_key(key)
+  -- hit_level be 1 if stale value is propelled into L1 cache; so set a minimal `resurrect_ttl`
   local credential, err, hit_level = cache:get(credential_cache_key, { resurrect_ttl = 0.001 }, load_credential,
                                     key)
 
-  kong.log.debug("credential hit_level: ", hit_level)
   if err then
     return error(err)
   end
 
+  kong.log.debug("credential hit_level: ", tostring(hit_level))
+
   -- no credential in DB, for this key, it is invalid, HTTP 401
-  if not credential or credential.err then
-    if credential and credential.err then
-      kong.log.debug(credential.err)
-    end
+  if not credential or hit_level == 4 then
 
     return nil, { status = 401, message = "Invalid authentication credentials" }
   end
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -25,6 +25,11 @@ local function load_credential(key)
     return nil, err
   end
 
+  if cred.ttl == 0 then
+    cred = { err = "credential ttl is 0" }
+    return cred, err
+  end
+
   return cred, nil, cred.ttl
 end
 
@@ -154,14 +159,20 @@ local function do_authentication(conf)
   local cache = kong.cache
 
   local credential_cache_key = kong.db.keyauth_credentials:cache_key(key)
-  local credential, err = cache:get(credential_cache_key, nil, load_credential,
+  local credential, err, hit_level = cache:get(credential_cache_key, { resurrect_ttl = 0.001 }, load_credential,
                                     key)
+
+  kong.log.debug("credential hit_level: ", hit_level)
   if err then
     return error(err)
   end
 
   -- no credential in DB, for this key, it is invalid, HTTP 401
-  if not credential then
+  if not credential or credential.err then
+    if credential and credential.err then
+      kong.log.debug(credential.err)
+    end
+
     return nil, { status = 401, message = "Invalid authentication credentials" }
   end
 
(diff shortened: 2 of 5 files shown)


kong/plugins/key-auth/handler.lua
```diff
@@ -25,6 +25,11 @@ local function load_credential(key)
     return nil, err
   end
 
+  if cred.ttl == 0 then
```
bungle, line 28: The description talks between 0 and 1, should this then be:

```suggestion
  if cred.ttl < 1 then
```

outsinre: `cred.ttl` is the result of `floor`. So theorily, it should be `cred.ttl == 0`.

bungle (approved):

bungle:
@outsinre, just heard from @dndx that there might be a better fix in core for this, thus I marked this with `pr/ready (but hold merge)`.

outsinre:
> @outsinre, just heard from @dndx that there might be a better fix in core for this, thus I marked this with `pr/ready (but hold merge)`.

Yeap. We change the db strategy `FLOOR` part.

spec/03-plugins/09-key-auth/02-access_spec.lua
```diff
@@ -968,7 +967,8 @@ for _, strategy in helpers.each_strategy() do
         helpers.stop_kong()
       end)
 
-      it("authenticate for up to 'ttl'", function()
+      it("authenticate for up to 'ttl' #test", function()
```
bungle, line 970:
```suggestion
      it("authenticate for up to 'ttl'", function()
```

kong/plugins/key-auth/handler.lua
```diff
@@ -25,6 +25,11 @@ local function load_credential(key)
...
+    cred = { err = "credential ttl is 0" }
+    return cred, err
```
bungle, line 30: This is pretty strange. Can't we in this case just return:
```lua
return cred, nil, 0.001
```

**without** doing:
```
cred = { err = "credential ttl is 0" }
```

bungle (changes requested):
Questions and comments

kong/plugins/key-auth/handler.lua
outsinre: The problem is the `cred.ttl = 0`. We can return such `cred` as that would make mlcache never expires.

`cred.ttl = 0` means the data in DB is expired. Therefore, we can return TTL as `nil` instead of `0.0.01`.

bungle: @outsinre wouldn't the `< 0`, e.g. `-2` mean that it has expired?

outsinre: From the [readme](https://github.com/thibaultcha/lua-resty-mlcache#get):

```
If returned as `ttl < 0`, `value` will be returned by get(),
    --      but not cached. This return value will be ignored if not a number
```

dndx:
@bungle Could you take another round of look at this?

dndx:
(This should not be in 3.1, BTW)

kong/plugins/key-auth/handler.lua
```diff
@@ -25,6 +25,11 @@ local function load_credential(key)
     return nil, err
   end
 
+  if cred.ttl == 0 then
+    kong.log.debug("key expired")
+    return nil, err
```
bungle, line 30: In this case `err` is nil?

kong/plugins/key-auth/handler.lua
```diff
@@ -154,14 +159,19 @@ local function do_authentication(conf)
   local cache = kong.cache
 
   local credential_cache_key = kong.db.keyauth_credentials:cache_key(key)
-  local credential, err = cache:get(credential_cache_key, nil, load_credential,
+  -- hit_level be 1 if stale value is propelled into L1 cache; so set a minimal `resurrect_ttl`
+  local credential, err, hit_level = cache:get(credential_cache_key, { resurrect_ttl = 0.001 }, load_credential,
```
bungle, line 163: We handle hit level below, so should we have this without `{ resurrect_ttl = 0.001 }`

kong/plugins/key-auth/handler.lua
outsinre: This `err` might be `nil` or not depending on the `mlcache` backend.

kong/plugins/key-auth/handler.lua
outsinre: We should have it.

Though we have `hit_level` but sometimes expired key will be cached in L1 cache and the `hit_level` is 1 then.

So, adding `{ resurrect_ttl = 0.001 }` to overcome that case.

kong/pdk/cluster.lua
```diff
@@ -46,7 +46,15 @@ local function new(self)
   --
   -- -- use id here
   function _CLUSTER.get_id()
-    return kong.core_cache:get(CLUSTER_ID_PARAM_KEY, nil, fetch_cluster_id)
+    local cluster_id, err, hit_lvl = kong.core_cache:get(CLUSTER_ID_PARAM_KEY, nil, fetch_cluster_id)
+
+    if err then
+      kong.log.debug("core_cache hit level: ", hit_lvl)
```
bungle, line 52: Can the `hit_lvl` be nil? In which case I don't see value in logging it.

kong/plugins/key-auth/handler.lua
```diff
@@ -154,14 +160,18 @@ local function do_authentication(conf)
...
                                     key)
+
+  kong.log.debug("credential hit_level: ", hit_level)
```
bungle, line 167: Can the hit level be `nil` in which case I don't see point logging it.

kong/pdk/cluster.lua
outsinre: It can be `nil` if the `get(...)` is returns `nil, err`.

How about add a `tostring(hit_lvl)`?

kong/plugins/key-auth/handler.lua
outsinre: Ditto

kong/pdk/cluster.lua
outsinre: The log here is not necessary! As when `err` is true, `hit_lvl` is `nil`.

This part is to make sure the `nil` and `hit_lvl` values are are appended to `cluster_id`. Otherwise, `cluster_id` returned is wrong.

kong/plugins/key-auth/handler.lua
outsinre: I just want to add a debug log to help debug. Can remove it as well.

bungle (approved):

kikito (approved):

PR #9806, opened 2022-11-22
feat(schema) add entity info to errors

rainest:
### Summary

Adds a designated list of entity fields to error responses when those fields are present. This seeks to make DB-less error responses more clearly indicate which entities are responsible for a rejected configuration.

In particular, the inclusion of tags will allow the ingress controller to add information about the source Kubernetes resource and receive it back in error reports for any entity.

Name and ID are included also as good human-friendly fields for non-controller updates.

### Checklist

### Full changelog

* Add a constant table of fields to include with errors.
* Include an `entity_metadata` field in `Schema:validate()` errors. If fields from the constant are present, `entity_metadata` is populated and included. If no fields are present, `entity_metadata` is omitted. 

### Issue reference

FTI-4366

### DIFF
--- a/kong/constants.lua
+++ b/kong/constants.lua
@@ -237,6 +237,12 @@ local constants = {
     [ngx.ALERT] = "alert",
     [ngx.EMERG] = "emerg",
   },
+
+  ENTITY_ERROR_METADATA_FIELDS = {
+    "id",
+    "name",
+    "tags",
+  },
 }
 
 for _, v in ipairs(constants.CLUSTERING_SYNC_STATUS) do
--- a/kong/db/schema/init.lua
+++ b/kong/db/schema/init.lua
@@ -5,6 +5,7 @@ local cjson        = require "cjson"
 local new_tab      = require "table.new"
 local nkeys        = require "table.nkeys"
 local is_reference = require "kong.pdk.vault".new().is_reference
+local constants    = require "kong.constants"
 
 
 local setmetatable = setmetatable
@@ -1987,6 +1988,17 @@ function Schema:validate(input, full_check, original_input, rbw_entity)
   run_transformation_checks(self, input, original_input, rbw_entity, errors)
 
   if next(errors) then
+    local meta = {}
+    local has_meta = false
+    for _, field in ipairs(constants.ENTITY_ERROR_METADATA_FIELDS) do
+      if type(input[field]) ~= "userdata" and input[field] ~= nil then
+        meta[field] = input[field]
+        has_meta = true
+      end
+    end
+    if has_meta then
+      errors["entity_metadata"] = meta
+    end
     return nil, errors
   end
   return true


rainest:
**Additional background**

On-entity metadata (via tags) provides the simplest means of adding additional information (the source Kubernetes resource's identifying info for a Kong entity generated by the controller) for later retrieval: the information is stored in an existing field guaranteed to be present (the controller simply requires that an entity support tags to interact with it at all) and specifies the exact information the controller needs to locate the source resource, avoiding the need to track that information separately.

We considered tracking parent resource information separately and binding it to another field, but rejected this for two reasons:
- There is no perfect candidate field to use as a reference. Not all resources have names (e.g. [certificates](https://github.com/Kong/kong/blob/e2df4e68f97aff1bfc6195475ad36ca6185aad83/kong/db/schema/entities/certificates.lua#L16-L22)). The controller does not currently set IDs and instead lets Kong populate them, and setting IDs may cause unexpected side effects (in particular, referential integrity may be affected).
- @hbagdi wishes to further display parent resource information in Manager, so that information needs to be available on the entity itself. Parent resource information kept within the controller and linked to ID or similar would not be available elsewhere.

**Draft issues**
1. Many tests are sensitive to the exact contents of errors, do not expect the new `entity_metadata` field, and fail when it is present.
1. `Schema:validate()` isn't designed to be specific to Kong entity schemas, but the fields in the constant are. Checking against the constant within `Schema:validate()` is architecturally messy as such. It probably should be passed in as an argument.
1. Errors from `Schema:validate()` are rendered twice, using different formatters, in error HTTP responses, both as a multi-line JSON object (`fields`) and as a condensed, single-line representation of that object (part of `messages`). We arguably don't want the additional metadata in the message for brevity--I'm not even sure that we'd really want the complete error in there at all, but it is currently present. Excluding metadata from the message is a bit annoying because it's constructed several layers below where it's actually output and may have metadata fields nested at any level.
1. The same error generation is used for both DB-less and DB-backed HTTP responses, and is redundant for the latter--DB-backed errors are always associated with the single object submitted in the request, and it's unnecessary to separately list their identifying information.

Omitting the additional metadata unless the request includes a particular header could address problems 1 and 4, at the cost of making it less obvious for DB-less (the controller would know to add the header, but an unaware user wouldn't). Only including the metadata if validation occurred in the context of a DB-less update would not require user awareness, but would still require some error test updates--existing unit tests for `Schema:validate()` could forcibly disable it, but not integration tests.

**Examples**

DB-less:

```
http :8001/config config=@/tmp/badservice.yaml.txt
HTTP/1.1 400 Bad Request
{
    "code": 14,
    "fields": {
        "services": [
            {
                "entity_metadata": {
                    "id": "b8aa692c-6d8d-580e-a767-a7dbc1f58344",
                    "name": "default.echo.pnum-80",
                    "tags": [
                        "k8s-name:example",
                        "k8s-namespace:default",
                        "k8s-kind:Service"
                    ]
                },
                "read_timeout": "expected an integer"
            }
        ]
    },
    "message": "declarative config is invalid: {services={{entity_metadata={id=\"b8aa692c-6d8d-580e-a767-a7dbc1f58344\",name=\"default.echo.pnum-80\",tags={\"k8s-name:example\",\"k8s-namespace:default\",\"k8s-kind:Service\"}},read_timeout=\"expected an integer\"}}}",
```

[badservice.yaml.txt](https://github.com/Kong/kong/files/10070098/badservice.yaml.txt)

DB-backed:

```
 curl -svX POST http://localhost:8001/routes \
  --data 'protocols[]=grpcs' \
  --data name=example \
  --data 'methods[]=GET' \ 
  --data 'tags[]="k8s-name:example"' --data 'tags[]="k8s-namespace:default"' --data 'tags[]="k8s-kind:Ingress"' | jq

< HTTP/1.1 400 Bad Request

{
  "code": 2,
  "name": "schema violation",
  "message": "5 schema violations (must set one of 'hosts', 'headers', 'paths', 'snis' when 'protocols' is 'grpcs'; entity_metadata.id: 8a00ee3b-77a8-4cd7-a059-099b9e388ecd; entity_metadata.name: example; entity_metadata.tags: {\n  \"\\\"k8s-name:example\\\"\",\n  \"\\\"k8s-namespace:default\\\"\",\n  \"\\\"k8s-kind:Ingress\\\"\"\n}; methods: cannot set 'methods' when 'protocols' is 'grpc' or 'grpcs')",
  "fields": {
    "entity_metadata": {
      "id": "8a00ee3b-77a8-4cd7-a059-099b9e388ecd",
      "name": "example",
      "tags": [
        "\"k8s-name:example\"",
        "\"k8s-namespace:default\"",
        "\"k8s-kind:Ingress\""
      ]
    },
    "@entity": [
      "must set one of 'hosts', 'headers', 'paths', 'snis' when 'protocols' is 'grpcs'"
    ],
    "methods": "cannot set 'methods' when 'protocols' is 'grpc' or 'grpcs'"
  }
}
```

hbagdi:
Can you provide an example response for a /config request that contains 2 invalid routes and 2 invalid services?

rainest:
> Can you provide an example response for a /config request that contains 2 invalid routes and 2 invalid services and the same (or higher) number of valid entities?

https://gist.github.com/rainest/b20209d345d210724831a88a7ce22ac2#file-log-txt

The change maintains the structure of existing error reporting w/r/t inclusion of nulls and array order, it's just adding the `entity_metadata` field to each error object.

catbro666:
Some branches missed adding entity meta info to errors, please see #9973 .

rainest:
Superseded by https://github.com/Kong/kong/pull/10161

PR #9807, opened 2022-11-22
chore(ci): use a kbt capable of using pre-builtruntime

hutchic:


colinh-kong:
Folks don't want to learn Docker so this lives on a personal [fork](https://github.com/hutchic-org/kong-development)

PR #9823, opened 2022-11-23
chore(clustering): cluster compat for 3.1

flrgh:
## changes

* `kong.clustering.utils` was starting to become very kitchen sink-y, so some of its functions were factored out and moved into `kong.clustering.compat` and `kong.clustering.compat.version` to make testing easier
* removed all 2.x business logic for version/config compatibility
* extended the plugin field removal logic to cover nested `fields.like.this`
* added new unit and integration tests


kong/clustering/compat/init.lua
```diff
@@ -0,0 +1,335 @@
...
+
+local REMOVED_FIELDS = require("kong.clustering.compat.removed_fields")
+local CLUSTERING_SYNC_STATUS = constants.CLUSTERING_SYNC_STATUS
+local KONG_VERSION = meta.version
+
+
+local _M = {}
+
+
+local function check_kong_version_compatibility(cp_version, dp_version, log_suffix)
```
tyler-ball, line 33: I think you should add a test for this to `03-compat_spec.lua` so we know if someone accidentally changes this behavior

tyler-ball (commented):
What was the logic for removing [spec/01-unit/19-hybrid/03-fields-removal_spec.lua](https://github.com/Kong/kong/pull/9823/files#diff-19e5d202988a432a375375eaac3111881c2077f383b7ead48bd44bde7da31d74)? It was just a duplicate of what was in `removed_fields.lua`?

flrgh:
> What was the logic for removing [spec/01-unit/19-hybrid/03-fields-removal_spec.lua](https://github.com/Kong/kong/pull/9823/files#diff-19e5d202988a432a375375eaac3111881c2077f383b7ead48bd44bde7da31d74)? It was just a duplicate of what was in `removed_fields.lua`?

This file was renamed to `spec/01-unit/19-hybrid/03-compat_spec.lua` to be a little more generic so we can put more `kong.clustering.compat.*` tests in it. You'll see a big size difference between the two because I removed _all_ of the version-specific tests from this test suite. Instead of relying on the contents of `kong.clustering.compat.removed_fields`, we pass in our own dummy removed fields data structure in the unit test. This way the behavior is tested without needing to manage a diff between the OSS and EE tests.

kong/clustering/compat/init.lua
flrgh: This is not a good idea, but I think it's okay to go without more unit tests for this at the moment.

Reason being that this is not a new function (it was moved over as-is from `kong.clustering.utils`), _and_ it has some decent integration test coverage [here](https://github.com/Kong/kong/blob/acd48732ce67ee9b2af41a482cb8b75543fae877/spec/02-integration/09-hybrid_mode/01-sync_spec.lua#L409-L420).

tyler-ball (approved):

PR #9838, opened 2022-11-28
feat (conf) set non root user to run in devcontainer docker containers

### LINKED ISSUE #8748: Dockerfile Misconfiguration: Privileged Container

DiptoChakrabarty:
### Summary
This sets a non root user in the devcontainer dockerfile to replace operations as root user

To ensure non root user is set as root user has high privileges 
### Full changelog
* add non user in devcontainer dockerfile

 ...

### Issue reference #8748 

Fix #8748 

Continuation of PR #8817

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,3 +10,6 @@ RUN apt-get install -y \
         git \
         m4 \
         libyaml-dev
+RUN addgroup -S kong && adduser -S kong -G kong
+RUN chown -R kong:kong /home/kong/
+USER kong


fffonion:
@DiptoChakrabarty This may fail later when container is building inside, could you go through that process
and make sure all permissions are set correctly?

PR #9845, opened 2022-11-30
refactor(clustering): tune wrpc data plane 

chronolaw:
### Summary

Simplify the code of data plane, make it more human friendly.

### Checklist

- [x] There's no entry in the CHANGELOG
- [x] There is no user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE

### Full changelog

* remove redundant `is_dp_worker_process()`
* `do end` block for `get_services`
* refactor `config_thread` with `goto continue`
* refactor `communicate()`
* rename `self.child` to `self.wrpc_handler`

### DIFF
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -35,11 +35,6 @@ function _M.new(conf)
   local key = assert(pl_file.read(conf.cluster_cert_key))
   self.cert_key = assert(ssl.parse_pem_priv_key(key))
 
-  if conf.role == "control_plane" then
-    self.wrpc_handler =
-      require("kong.clustering.wrpc_control_plane").new(self.conf, self.cert_digest)
-  end
-
   return self
 end
 
@@ -53,6 +48,9 @@ function _M:init_cp_worker(plugins_list)
 
   events.init()
 
+  self.wrpc_handler =
+    require("kong.clustering.wrpc_control_plane").new(self.conf, self.cert_digest)
+
   self.wrpc_handler:init_worker(plugins_list)
 end
 
@@ -63,8 +61,9 @@ function _M:init_dp_worker(plugins_list)
       return
     end
 
-    self.child = require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
-    self.child:init_worker(plugins_list)
+    self.wrpc_handler =
+      require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
+    self.wrpc_handler:init_worker(plugins_list)
   end
 
   assert(ngx.timer.at(0, start_dp))
--- a/kong/clustering/init.lua
+++ b/kong/clustering/init.lua
@@ -63,6 +63,7 @@ function _M:init_dp_worker(plugins_list)
 
     self.wrpc_handler =
       require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
+
     self.wrpc_handler:init_worker(plugins_list)
   end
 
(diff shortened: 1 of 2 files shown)


kong/clustering/init.lua
```diff
@@ -53,22 +48,29 @@ function _M:init_cp_worker(plugins_list)
 
   events.init()
 
+  self.wrpc_handler =
+    require("kong.clustering.wrpc_control_plane").new(self.conf, self.cert_digest)
+
```
StarlightIbuki, line 53: Notice this was at `init` phase, and this change makes it happen at `init_worker`.

kong/clustering/init.lua
```diff
@@ -53,22 +48,29 @@ function _M:init_cp_worker(plugins_list)
...
     if premature then
       return
     end
 
-    self.child = require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
-    self.child:init_worker(plugins_list)
+    self.wrpc_handler =
+      require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
+
+    self.wrpc_handler:init_worker(plugins_list)
```
StarlightIbuki, line 67: Same here. Is it necessary for every worker to load its own copy of `wrpc_data_plane`, especially given that only #0 is actually using it?

kong/clustering/wrpc_data_plane.lua
```diff
@@ -82,20 +75,23 @@ local function init_config_service(service)
...
 
-  return wrpc_services
+  get_services = function()
+    if not wrpc_services then
+      wrpc_services = wrpc_proto.new()
+
+      wrpc_services:import("kong.services.config.v1.config")
+      wrpc_services:set_handler("ConfigService.SyncConfig", config_handler)
+
+      init_negotiation_client(wrpc_services)
```
StarlightIbuki, line 87: It might be OK to put `nego` service initialization after `conf` service, but they were intentionally ordered as nego happens before conf.

kong/clustering/wrpc_data_plane.lua
```diff
@@ -245,31 +252,44 @@ local function communicate_impl(dp)
...
-  return ngx.timer.at(reconnection_delay or 0, communicate_loop, dp)
-end
+local function communicate(dp, reconnection_delay)
 
-function communicate_loop(premature, dp)
-  if premature then
-    -- worker wants to exit
-    return
-  end
+  local communicate_loop = function(premature, dp)
```
StarlightIbuki, line 258: why not just
```suggestion
  local  function communicate_look(premature, dp)
```

StarlightIbuki: I prefer not to make it nested. That's the reason I wrote "local communicate_loop" and defined it later.

chronolaw: I agree. But I also don't like forward declare vars, which makes the code too hard to read.

kong/clustering/init.lua
StarlightIbuki: But it's required for every worker. I guess the original idea to require in phase handler is to reduce the `require` cost for non-DP(CP or traditional), but this causes multiple copy.

kong/clustering/wrpc_data_plane.lua
StarlightIbuki: Just for readability. It's fine for both orders.

chronolaw:
Considering the future refactoring, this code may be useless, so I close it.

PR #9846, opened 2022-11-30
feat(prometheus): disable export if no instance

StarlightIbuki:
Make sure we cherry-pick this to EE considering workspaces.

fix FT-3626

### DIFF
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -301,6 +301,13 @@ local function log(message, serialized)
   end
 end
 
+
+local enabled = true
+-- enable or disable the metric_data export endpoint
+local function enable(enable_or_not)
+  enabled = enable_or_not
+end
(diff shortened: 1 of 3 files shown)


kong/plugins/prometheus/exporter.lua
```diff
@@ -312,6 +319,11 @@ local should_export_upstream_health_metrics = false
 
 
 local function metric_data(write_fn)
+  if not enabled then
+    kong.log.info("prometheus: no plugin instance enabled, so export endpoint is disabled.")
+    return kong.response.exit(404, { message = "No instance" })
```
hbagdi, line 324: Some customers use the Prometheus plugin to monitor non-proxy path metrics that this plugin exposes and this change will break those customers.
I strongly advise not proceeding with this break change.

StarlightIbuki:
Closing because we want a different solution.

Issue #9839, opened 2022-11-29
init_by_lua error: /usr/local/openresty/lualib/ngx/semaphore.lua:126: no request found


mong0520:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

3.0.1

### Current Behavior

I was using kong 2.8.3 CE and upgrading to 3.0.1 recently and try to use the cool function vault.
Then I set PostgresSQL connection information via vault like `{vault://env/DATABASE/password}`, everything works well.

However, after deploy it to AWS ECS, it crashed and I got below errors
After digging it out, I found it relates to resty-core party (see https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/semaphore.lua#L124-L127) and I have no idea how to resolve it.
```
2022/11/29 06:54:53 [error] 221#0: init_by_lua error: /usr/local/openresty/lualib/ngx/semaphore.lua:126: no request found
stack traceback:
[C]: in function 'error'
/usr/local/openresty/lualib/ngx/semaphore.lua:126: in function 'wait'
/usr/local/share/lua/5.1/kong/pdk/vault.lua:511: in function 'connect'
.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:521: in function 'query'
.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:289: in function 'init'
/usr/local/share/lua/5.1/kong/db/init.lua:141: in function 'init_connector'
/usr/local/share/lua/5.1/kong/init.lua:546: in function 'init'
init_by_lua:3: in main chunk
nginx: [error] init_by_lua error: /usr/local/openresty/lualib/ngx/semaphore.lua:126: no request found
```

<img width="1394" alt="image" src="https://user-images.githubusercontent.com/10734784/204464690-5cf9475d-e0fb-4417-a12a-07305b053354.png">

### Steps To Reproduce

```markdown
1. Use kong version 3.0.1 (image :https://hub.docker.com/layers/library/kong/3.0.1-ubuntu/images/sha256-70b9313e4f8c89b35b12d3a964a883840d75a19a46255a26ca1f8bac57c65ebf?context=explore)
2. Set `KONG_PG_HOST`, `KONG_PG_USER`, `KONG_PG_PASSWORD`, `KONG_PG_DATABASE` to vault expression, like `{vault://env/DATABASE/password}`
3. Test it on local, it works
4. Deploy it to AWS ECS (container solution), it crashes.
```

[...]

PR #9851, opened 2022-12-01
fix(pdk/vault): do not call sema:wait in init phase

chronolaw:
### Summary

Fix issue #9839

### Checklist

### Full changelog

* check nginx phase
* if it is `init` or `init_worker`, set `wait_ok` to `false`

### Issue reference

Fix #_[9839]_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -208,6 +208,8 @@
   and body parameter type of `kong.response.exit` to bytes. Note that old
   version of go PDK is incompatible after this change.
   [#9526](https://github.com/Kong/kong/pull/9526)
+- Vault will not call `semaphore:wait` in `init` or `init_worker` phase.
+  [#9851](https://github.com/Kong/kong/pull/9851)
 
 #### Plugins
 
--- a/kong/pdk/vault.lua
+++ b/kong/pdk/vault.lua
@@ -21,6 +21,7 @@ local cjson = require("cjson.safe").new()
 
 
 local ngx = ngx
+local get_phase = ngx.get_phase
 local fmt = string.format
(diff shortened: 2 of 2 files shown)


chronolaw:
I think it should have a change log entry, but I don't know how to describe it. 
Could someone help me for it?

chronolaw:
> hi @chronolaw ,
> 
> I'm the issue submitter of #9839. First thank you for the fixing, and I have few questions about this PR.
> 
> 1. Since `sema:wait` only works in some specific phases, how about set `wait_ok` to true if those phases are detected instead of set it to false in init false only?
> 
> ```
> # https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/semaphore.md#wait
> syntax: ok, err = sema:wait(timeout)
> context: rewrite_by_lua*, access_by_lua*, content_by_lua*, ngx.timer.*
> ```
> 
> 2. Won't it be any problem if not calling to `RETRY_SEMAPHORE:wait(RETRY_WAIT)` but calls to `RETRY_SEMAPHORE:post()` in subsequence codes? I don't know if they should be paired.

1. In Kong, pdk code will run in `init` `init_worker` `access` `rewrite` phases, I think it is OK.
2. We set `wait_ok = false`, so subsequence code will not call `RETRY_SEMAPHORE:post()`

bungle (approved):

bungle:
@chronolaw thank you!

mong0520:
@chronolaw 
One more question.
It looks like an always crash issue (calls to `sema:wait()` in init phase), but why there is no one reports it since 3.0.1 was release for a month?

PR #9852, opened 2022-12-01
feat(prometheus): skip fields in shm (high cost) when no instance configured

StarlightIbuki:
fix FT-3626

### DIFF
--- a/kong/plugins/prometheus/api.lua
+++ b/kong/plugins/prometheus/api.lua
@@ -1,15 +1,19 @@
 local exporter = require "kong.plugins.prometheus.exporter"
+local tbl_insert = table.insert
+local tbl_concat = table.concat
+
 
 local printable_metric_data = function(_)
   local buffer = {}
   -- override write_fn, since stream_api expect response to returned
   -- instead of ngx.print'ed
   exporter.metric_data(function(new_metric_data)
-    table.insert(buffer, table.concat(new_metric_data, ""))
+    tbl_insert(buffer, tbl_concat(new_metric_data, ""))
   end)
 
-  return table.concat(buffer, "")
+  return tbl_concat(buffer, "")
 end
+
 
 return {
   ["/metrics"] = {
--- a/kong/plugins/prometheus/api.lua
+++ b/kong/plugins/prometheus/api.lua
@@ -1,11 +1,11 @@
 local exporter = require "kong.plugins.prometheus.exporter"
 
-local printable_metric_data = function()
+local printable_metric_data = function(_)
   local buffer = {}
   -- override write_fn, since stream_api expect response to returned
   -- instead of ngx.print'ed
-  exporter.metric_data(function(data)
-    table.insert(buffer, table.concat(data, ""))
+  exporter.metric_data(function(new_metric_data)
+    table.insert(buffer, table.concat(new_metric_data, ""))
   end)
 
   return table.concat(buffer, "")
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -33,7 +33,7 @@ local http_subsystem = kong_subsystem == "http"
 
 
 -- should we introduce a way to know if a plugin is configured or not?
-local is_prometheus_enabled, register_event do
+local is_prometheus_enabled, register_events_handler do
   local PLUGIN_NAME = "prometheus"
   local CACHE_KEY = "prometheus:enabled"
 
@@ -232,7 +232,7 @@ end
 
 local function init_worker()
   prometheus:init_worker()
-  register_event()
+  register_events_handler()
 end
 
 -- Convert the MD5 hex string to its numeric representation
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -326,8 +326,6 @@ local function metric_data(write_fn)
   metrics.connections:set(nginx_statistics['connections_reading'], { node_id, kong_subsystem, "reading" })
   metrics.connections:set(nginx_statistics['connections_writing'], { node_id, kong_subsystem, "writing" })
   metrics.connections:set(nginx_statistics['connections_waiting'], { node_id, kong_subsystem,"waiting" })
-  metrics.connections:set(nginx_statistics['connections_accepted'], { node_id, kong_subsystem, "accepted" })
-  metrics.connections:set(nginx_statistics['connections_handled'], { node_id, kong_subsystem, "handled" })
 
   metrics.nginx_requests_total:set(nginx_statistics['total_requests'], { node_id, kong_subsystem })
 
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -77,7 +77,7 @@ local is_prometheus_enabled, register_event do
     else
       worker_events.register(function()
         kong.db:invalidate(CACHE_KEY)
-      end, "declarative", "flip_config")
+      end, "declarative", "reconfigure")
     end
   end
 end
--- a/kong/plugins/prometheus/exporter.lua
+++ b/kong/plugins/prometheus/exporter.lua
@@ -34,8 +34,8 @@ local http_subsystem = kong_subsystem == "http"
 
 -- should we introduce a way to know if a plugin is configured or not?
 local is_prometheus_enabled, register_event do
-  local plugin_name = "prometheus"
-  local cache_key = "prometheus:enabled"
+  local PLUGIN_NAME = "prometheus"
+  local CACHE_KEY = "prometheus:enabled"
 
 
   local function is_prometheus_enabled_fetch()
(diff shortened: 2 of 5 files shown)


kong/plugins/prometheus/exporter.lua
```diff
@@ -31,6 +31,65 @@ package.loaded['prometheus_resty_counter'] = require("resty.counter")
...
+          return
+        end
+
+        local operation = data.operation
+        if operation == "create" then
+          instance_count = instance_count + 1
+        elseif operation == "delete" then
+          instance_count = instance_count - 1
+        end
+      end, "crud", "plugins")
```
dndx, line 76: Let's not do this, this is not a reliable way of maintaining the `instance_count` as worker events could be lost and bring `instance_count` into a forever wrong state.

We should simply store a cached value called `"prometheus_enabled"` or so and `invalidate` it when `crud` happens and `data.entity.name == "prometheus"`. Then we use `cache.get` and a callback to determine if there is any enabled Prometheus instance anywhere.

kong/plugins/prometheus/exporter.lua
```diff
@@ -31,6 +31,65 @@ package.loaded['prometheus_resty_counter'] = require("resty.counter")
...
+
+  local function initialize_instance_counters()
+    instance_count = 0
+    for plugin, err in kong.db.plugins:each() do
+      if err then
+        kong.log.crit("could not obtain list of plugins: ", err)
+        return nil, err
+      end
+
+      if plugin.name == plugin_name then
```
dndx, line 49: You need to consider whether `enabled = true` as well.

dndx (changes requested):

kong/plugins/prometheus/exporter.lua
StarlightIbuki: done.

kong/plugins/prometheus/exporter.lua
StarlightIbuki: done.

StarlightIbuki:
I've manually tested. It can reduce the number of metrics outputted when no plugin instance is enabled.

kong/plugins/prometheus/api.lua
```diff
@@ -1,11 +1,11 @@
...
 
-local printable_metric_data = function()
+local printable_metric_data = function(_)
   local buffer = {}
   -- override write_fn, since stream_api expect response to returned
   -- instead of ngx.print'ed
-  exporter.metric_data(function(data)
-    table.insert(buffer, table.concat(data, ""))
+  exporter.metric_data(function(new_metric_data)
+    table.insert(buffer, table.concat(new_metric_data, ""))
```
chronolaw, line 8: localize `table.insert` and `table.concat`?

kong/plugins/prometheus/api.lua
```diff
@@ -1,11 +1,11 @@
 local exporter = require "kong.plugins.prometheus.exporter"
 
-local printable_metric_data = function()
+local printable_metric_data = function(_)
```
chronolaw, line 3: Is param `_` necessary?

kong/plugins/prometheus/exporter.lua
```diff
@@ -31,6 +31,58 @@ package.loaded['prometheus_resty_counter'] = require("resty.counter")
 local kong_subsystem = ngx.config.subsystem
 local http_subsystem = kong_subsystem == "http"
 
+
+-- should we introduce a way to know if a plugin is configured or not?
+local is_prometheus_enabled, register_event do
+  local plugin_name = "prometheus"
```
chronolaw, line 37: constant `plugin_name` `cache_key` should be upper case.

kong/plugins/prometheus/exporter.lua
```diff
@@ -31,6 +31,58 @@ package.loaded['prometheus_resty_counter'] = require("resty.counter")
...
+      worker_events.register(function(data)
+        if data.entity.name == plugin_name then
+          kong.db:invalidate(cache_key)
+        end
+      end, "crud", "plugins")
+
+    else
+      worker_events.register(function()
+        kong.db:invalidate(cache_key)
+      end, "declarative", "flip_config")
```
chronolaw, line 80: It seems that now there is no event `flip_config`, could you check it again?

kong/plugins/prometheus/exporter.lua
```diff
@@ -31,6 +31,58 @@ package.loaded['prometheus_resty_counter'] = require("resty.counter")
...
+
+
+  local function is_prometheus_enabled_fetch()
+    for plugin, err in kong.db.plugins:each() do
+      if err then
+        kong.log.crit("could not obtain list of plugins: ", err)
+        return nil, err
+      end
+
+      if plugin.name == plugin_name and plugin.enabled then
```
chronolaw, line 48: how about `return plugin.name == plugin_name and plugin.enabled`?

kong/plugins/prometheus/api.lua
StarlightIbuki: This is a hint that we are skipping a parameter. This function is a call back which is called with the parameter "data".

kong/plugins/prometheus/api.lua
StarlightIbuki: I'm only trying to address the confusion related to the parameter.

kong/plugins/prometheus/exporter.lua
StarlightIbuki: Do we have any (developer) docs for me to understand the new inc-sync?

kong/plugins/prometheus/exporter.lua
StarlightIbuki: done

kong/plugins/prometheus/exporter.lua
```diff
@@ -425,7 +476,7 @@ local function metric_data(write_fn)
     end
   end
 
-  prometheus:metric_data(write_fn)
+  prometheus:metric_data(write_fn, not is_prometheus_enabled())
```
dndx, line 479: Better add a comment here to explain why `not is_prometheus_enabled()` is used. I had to dig into `prometheus.lua` to find out.

kong/plugins/prometheus/exporter.lua
```diff
@@ -453,8 +504,8 @@ local function get_prometheus()
   return prometheus
 end
 
-local function set_export_upstream_health_metrics()
-  should_export_upstream_health_metrics = true
+local function set_export_upstream_health_metrics(set_or_not)
```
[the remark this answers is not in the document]
StarlightIbuki: I think it's not. When people added this function, they call the function like it accepts a boolean value, but the function set the flag to true anyway, despite what the value is.

kong/plugins/prometheus/exporter.lua
```diff
@@ -31,6 +31,62 @@ package.loaded['prometheus_resty_counter'] = require("resty.counter")
 local kong_subsystem = ngx.config.subsystem
 local http_subsystem = kong_subsystem == "http"
 
+
+-- should we introduce a way to know if a plugin is configured or not?
+local is_prometheus_enabled, register_events_handler do
+  local PLUGIN_NAME = "prometheus"
+  local CACHE_KEY = "prometheus:enabled"
+
```
dndx, line 39:
```suggestion
```

kong/plugins/prometheus/exporter.lua
```diff
@@ -31,6 +31,58 @@ package.loaded['prometheus_resty_counter'] = require("resty.counter")
...
+    end
+
+    return enabled
+  end
+
+
+  -- invalidate cache when a plugin is added/removed/updated
+  function register_event()
+    local worker_events = kong.worker_events
+    if kong.configuration.database == "off" then
```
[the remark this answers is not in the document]
dndx: We need a test case for this. For DB-less and traditional.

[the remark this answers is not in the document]
StarlightIbuki: It should be covered by the `for` loop. However, my previous test did not cover this because it waits for too long so the cache expired.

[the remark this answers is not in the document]
StarlightIbuki: I've manually tested. If the `register_event` is commented out or the condition is not right, the test won't pass now.

dndx (approved):

kong/plugins/prometheus/exporter.lua
```diff
@@ -326,8 +382,6 @@ local function metric_data(write_fn)
   metrics.connections:set(nginx_statistics['connections_reading'], { node_id, kong_subsystem, "reading" })
   metrics.connections:set(nginx_statistics['connections_writing'], { node_id, kong_subsystem, "writing" })
   metrics.connections:set(nginx_statistics['connections_waiting'], { node_id, kong_subsystem,"waiting" })
-  metrics.connections:set(nginx_statistics['connections_accepted'], { node_id, kong_subsystem, "accepted" })
-  metrics.connections:set(nginx_statistics['connections_handled'], { node_id, kong_subsystem, "handled" })
```
StarlightIbuki, line 330: not sure where this comes from

PR #9860, opened 2022-12-02
revert(clustering): use privileged worker for control plane connection

bungle:
### Summary

Because of increased memory usage, despite the benefits with large configurations on hybrid mode data plane latency, we decided to remove the use of privileged worker for data plane to control plane connection.

This commit removes the use of privileged worker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,8 +135,6 @@
 
 #### Performance
 
-- Data plane's connection to control plane is moved to a privileged worker process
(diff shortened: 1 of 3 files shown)


hanshuebner (approved):

AntoineJac:
Hi team, as discussed could we keep this function and make it optional depending on the Kong config.
I think it could really help reducing spike latency for Hybrid mode. Thanks

PR #9876, opened 2022-12-05
chore: add commit ID of kong-build-tool as cache key

ADD-SP:
Correcting the cache key by adding KBT's commit hash as a part of cache key.


hutchic (changes requested):
The CE builds would be much better served by moving closer to parity with EE not further in my opinion.

When the way we develop != ci != release is how we get ourselves in tricky situations

PR #9882, opened 2022-12-05
fix: left pad Jaeger header

### LINKED ISSUE #8119: Accepting unpadded trace and span IDs in jaeger trace header values
Hello

What are your thoughts on accepting unpadded trace, span and parent span ID components in `uber-trace-id` header values? 

We have run into an issue where the zipkin plugin rejects these values when trace and span IDs are "short", i.e. not left-padded with zeroes to the expected length in bytes. We configure a byte count of 8, but might receive an ID of low value truncated to 7 bytes (or fewer...). The zipkin plugin will warn about the mismatched length and regenerate IDs, thus we fail to propagate the trace correctly.

We have worked around the issue by fixing up such header values on the incoming request before the zipkin plugin performs validation.

If the following documentation is relevant, perhaps the zipkin plugin could be more forgiving of such input? It would of course be stricter about what values it propagated.

https://www.jaegertracing.io/docs/1.25/client-libraries/#value

For example, we must perform the following sort of correction to avoid the occasional "short" incoming value from being discarded by the zipkin plugin:

```
received: 111111111111:22222222:3333333333333333:0

fixed up: 0000111111111111:0000000022222222:3333333333333333:0

```

We'd be happy to support making this change if you thought it was a good idea.

Cheers
Henry

greut:
### Summary

In our setup, Kong 3 -> OpenTelemetry collector with the Jaeger headers, the `[otel]` plugin is logging a bunch 400's telling that the TraceID lenght is invalid. It feels odd to pad the various elements before doing the `from_hex` rather than after the `to_hex`

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

Fix #8119

### DIFF
--- a/kong/tracing/propagation.lua
+++ b/kong/tracing/propagation.lua
@@ -276,11 +276,6 @@ local function parse_jaeger_trace_context_headers(jaeger_header)
     return nil, nil, nil, nil
   end
 
-  -- if trace_id is not of length 32 chars then 0-pad to left
-  if #trace_id < 32 then
-    trace_id = left_pad_zero(trace_id, 32)
-  end
-
   -- validating parent_id. If it is invalid just logging, as it can be ignored
   -- https://www.jaegertracing.io/docs/1.29/client-libraries/#tracespan-identity
   if #parent_id ~= 16 and tonumber(parent_id, 16) ~= 0 then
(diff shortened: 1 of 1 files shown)


kikito:
Hello @greut , this fix seems correct, it needs only two things:
* Rebase over master
* Modification of the tests that now fail because of the changed headers (you can see them failing here https://github.com/Kong/kong/actions/runs/3620849084/jobs/6103577526 )

PR #9888, opened 2022-12-05
refactor(cmd): use resty.signal for process signaling

flrgh:
This is a light refactor/enhancement of `kong.cmd.utils.kill`. The primary goal behind this set of changes was to have fewer `os.execute()` calls in core code.

1. Replaced `/usr/bin/kill` invocations with `resty.signal`
2. Renamed `kong.cmd.utils.kill` to `kong.cmd.utils.process`
3. Renamed `process.is_running()` to `process.exists()`
4. Added tests for `kong.cmd.utils.process.*` functions

While updating `spec.helpers` to account for these changes, I found a bug in the business logic for `helpers.signal_workers()` and rewrote most of the function while fixing it.


spec/helpers.lua
```diff
@@ -3662,27 +3638,26 @@ end
...
+      process.signal(pid, signal)
     end
 
-    local cmd = string.format("pkill %s -P `cat %s`", signal, pid_path)
-    local _, code = pl_utils.execute(cmd)
-
-    if not pid_dead(pid_path) then
-      return false
-    end
+    wait_until(function()
```
flrgh, line 3651: _before_ this change, this function was calling `pid_dead("/path/to/nginx.pid")`, which had some problems:

1. it doesn't make sense to check the master pid given we just wanted to kill worker processes
2. `pid_dead` is supposed to take a pid as input (not a pid file), so it would blindly call `ps -p /path/to/nginx.pid`. This returns `1` due to it being improper usage of `ps`, causing `pid_dead` to think "okay cool, the process has already exited, so I'm done".

kikito (commented):
Hi, I think the name change (from process to kill) makes sense but will make reviewing and approving this PR take longer.

I see two options:
1. Don't make the rename (😈)
2. Make the rename on a separate commit, so it can be reviewed in separation from the "meat" of this PR

flrgh:
I still think this change is worthwhile, but :hourglass:. I will re-open this PR when I can make it merge-worthy.

PR #9889, opened 2022-12-06
feat(schema) add new config error array

rainest:
### Summary

A variant of https://github.com/Kong/kong/pull/9806. Whereas #9806 adds its additional error metadata as a new field inside the existing error objects, this variant adds a new `fields.flattened` section, containing an array of error objects ([example](https://gist.github.com/rainest/d1640ce60a8b3c2d0bb99a8fd77d5f3f)). The new error objects store entity type and field names as values, rather than keys, allowing them to unmarshal into a Go struct.

### Checklist

### Full changelog

* Add an error accumulator table to schema validation and return it when `POST /config` encounters errors
* Add a field to track whether errors have been processed and added to the error accumulator already. Strip this field before returning errors.

### Issue reference

FTI-4366

### DIFF
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -66,9 +66,9 @@ return {
 
       local dc = declarative.new_config(kong.configuration)
 
-      local entities, _, err_t, meta, new_hash, entity_errors
+      local entities, _, err_t, meta, new_hash
       if self.params._format_version then
-        entities, _, err_t, meta, new_hash, entity_errors = dc:parse_table(self.params)
+        entities, _, err_t, meta, new_hash = dc:parse_table(self.params)
       else
         local config = self.params.config
         if not config then
--- a/kong/api/routes/config.lua
+++ b/kong/api/routes/config.lua
@@ -66,9 +66,9 @@ return {
 
       local dc = declarative.new_config(kong.configuration)
 
-      local entities, _, err_t, meta, new_hash
+      local entities, _, err_t, meta, new_hash, entity_errors
       if self.params._format_version then
-        entities, _, err_t, meta, new_hash = dc:parse_table(self.params)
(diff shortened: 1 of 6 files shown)


rainest:
At a high level, this approach shifts the complexity of dealing with the recursive nature of schema validation errors from their consumers (in our case, KIC) to the schema validation code. This PoC produces errors that can be easily unmarshaled in simpler cases, but fails on others:

- Errors within complex objects, where fields have object values, does not successfully strip out the `nested` annotations. For example, setting an HTTP status code within upstream health probe configuration to a string results in garbage output.
- Some unknown quality of controller-generated configuration causes a duplicate error entry in Kong's output. One of the duplicates lacks its errors. Unclear what causes this, since I encountered it at the end of the day and I couldn't reproduce it with manual testing.

The design is also very PoC-y and needs to be refined:

- At a glance, it remains difficult to distinguish fields that contain other entities (e.g. a `routes` field under a `services` entry). These fields are treated effectively the same as any other: although they're handled via a [dedicated case](https://github.com/Kong/kong/blob/d50574497651c36782597278c72d478e7cbe828b/kong/db/schema/init.lua#L913-L925) in the field validator, that information is not propagated back with errors. The additional `nested` field attempts to do this, but it's clunky, since in practice you need to look a few levels down into the object, and do some type guesswork, to find it.
- The starting schema isn't obviously special. Schemas are hierarchical collections of more schemas, and there isn't a top-level schema type accessible throughout validation for storing the accumulated errors. The hack approach is to stuff it in `ngx.ctx`, but that's a poor solution. I don't know if there's some existing part of the parent/subschema relationship that would be appropriate for this, or where would be best to refactor the schema validator if not.

rainest:
Superseded by https://github.com/Kong/kong/pull/10161

PR #9901, opened 2022-12-06
chore(tests) add yaml lint and ensure actions pinned to SHA

fffonion:
### Summary

Based on the triggering event, github may use what's on master to run actions for current PR, this means
in some cases we will only know an action has error after it's merged to master. This PR adds a workflow
to run on every change to actions related yml files, so that we can validate their syntax before merging.

This PR also ensures every not-well-known third party actions are pinned to SHA instead of tag.
[This blogpost](https://michaelheap.com/ensure-github-actions-pinned-sha/) is good reading for the
reason why we enforce this.

### Checklist

- [N/A] The Pull Request has tests
- [N/A] There's an entry in the CHANGELOG
- [N/A] There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE

### Full changelog

- add yaml lint to github actions files and pin actions to SHA

### DIFF
--- a/.github/workflows/actions-lint.yml
+++ b/.github/workflows/actions-lint.yml
@@ -1,0 +1,31 @@
+name: Actions Lint
+
+on:
+  pull_request:
+    paths:
+    - '.github/**.yml'
+
+jobs:
+  build:
+    name: Run Lint
+    runs-on: ubuntu-22.04
+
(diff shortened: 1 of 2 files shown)


hbagdi:

PR #9903, opened 2022-12-07
feat(pdk): allow plugin to fetch its id

StarlightIbuki:
### Summary

Exposing plugins' UUID as fields `id` of plugin config objects.

### Checklist

### Full changelog

- New PDK function `kong.plugin.get_id()`
  - plugin.lua 
  - new file's entry in rockspec
- Test
  - 05-pdk_spec.lua
  - spec/fixtures/custom_plugins/kong/plugins/get-plugin-id/*
- Change existing usage of `conf.__key__`
  - kong/plugins/*

### Issue reference

Implement FTI-4019

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,7 +98,7 @@
 
 #### PDK
 
-- PDK now support getting plugins' ID with `kong.plugin.get_id`.
+- PDK now supports getting plugins' ID with `kong.plugin.get_id`.
   [#9903](https://github.com/Kong/kong/pull/9903)
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -157,8 +157,6 @@
 
 ## 3.2.0
 
-### Addition
-
 ### Breaking Changes
 
 #### Plugins
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,11 @@
 - **Proxy-Cache**: add `ignore_uri_case` to configuring cache-key uri to be handled as lowercase
   [#10453](https://github.com/Kong/kong/pull/10453)
 
+#### PDK
+
+- PDK now support getting plugins' ID with `kong.plugin.get_id`.
+  [#9903](https://github.com/Kong/kong/pull/9903)
+
 ### Fixes
 
 #### Core
@@ -151,6 +156,8 @@
   [#10476](https://github.com/Kong/kong/pull/10476)
 
 ## 3.2.0
+
+### Addition
 
 ### Breaking Changes
 
--- a/kong/pdk/plugin.lua
+++ b/kong/pdk/plugin.lua
@@ -1,0 +1,20 @@
+--- Plugin related APIs
+--
+-- @module kong.plugin
+
+
+local _plugin = {}
+
+
+function _plugin.get_id(self)
+  return ngx.ctx.plugin_id
+end
+
+local function new()
+  return _plugin
+end
+
+
+return {
+  new = new,
+}
--- a/kong/plugins/datadog/handler.lua
+++ b/kong/plugins/datadog/handler.lua
@@ -48,12 +48,6 @@ local function compose_tags(service_name, status, consumer_id, tags, conf)
 end
 
 
--- we use plugin instance ID as the queue ID
-local function get_queue_id()
-  return kong.plugin.get_id()
-end
-
-
 local function log(conf, messages)
   local logger, err = statsd_logger:new(conf)
   if err then
@@ -119,7 +113,7 @@ function DatadogHandler:log(conf)
     return
   end
 
-  local queue_id = get_queue_id(conf)
+  local queue_id = kong.plugin.get_id()
   local q = queues[queue_id]
   if not q then
     local batch_max_size = conf.queue_size or 1
--- a/kong/plugins/datadog/handler.lua
+++ b/kong/plugins/datadog/handler.lua
@@ -48,8 +48,9 @@ local function compose_tags(service_name, status, consumer_id, tags, conf)
 end
 
 
-local function get_queue_id(conf)
-  return conf.__key__
+-- we use plugin instance ID as the queue ID
+local function get_queue_id()
+  return kong.plugin.get_id()
 end
 
 
--- a/kong/plugins/opentelemetry/handler.lua
+++ b/kong/plugins/opentelemetry/handler.lua
@@ -37,6 +37,11 @@ local default_headers = {
 local queues = {} -- one queue per unique plugin config
 local headers_cache = setmetatable({}, { __mode = "k" })
 
+-- we use plugin instance ID as the queue ID
+local function get_queue_id()
+  return kong.plugin.get_id()
+end
+
 local function get_cached_headers(conf_headers)
   if not conf_headers then
     return default_headers
@@ -153,7 +158,7 @@ end
 function OpenTelemetryHandler:log(conf)
   ngx_log(ngx_DEBUG, _log_prefix, "total spans in current request: ", ngx.ctx.KONG_SPANS and #ngx.ctx.KONG_SPANS)
 
-  local queue_id = conf.__key__
+  local queue_id = get_queue_id()
   local q = queues[queue_id]
   if not q then
     local process = function(entries)
--- a/kong/plugins/opentelemetry/handler.lua
+++ b/kong/plugins/opentelemetry/handler.lua
@@ -37,10 +37,6 @@ local default_headers = {
 local queues = {} -- one queue per unique plugin config
 local headers_cache = setmetatable({}, { __mode = "k" })
 
--- we use plugin instance ID as the queue ID
-local function get_queue_id()
-  return kong.plugin.get_id()
-end
 
 local function get_cached_headers(conf_headers)
   if not conf_headers then
@@ -158,7 +154,7 @@ end
 function OpenTelemetryHandler:log(conf)
   ngx_log(ngx_DEBUG, _log_prefix, "total spans in current request: ", ngx.ctx.KONG_SPANS and #ngx.ctx.KONG_SPANS)
 
-  local queue_id = get_queue_id()
+  local queue_id = kong.plugin.get_id()
   local q = queues[queue_id]
   if not q then
     local process = function(entries)
--- a/kong/plugins/statsd/log.lua
+++ b/kong/plugins/statsd/log.lua
@@ -27,8 +27,9 @@ local range_cache  = setmetatable({}, { __mode = "k" })
 local _M = {}
 
 
-local function get_queue_id(conf)
-  return conf.__key__
+-- we use plugin instance ID as the queue ID
+local function get_queue_id()
+  return kong.plugin.get_id()
 end
 
 local function get_cache_value(cache, cache_key)
--- a/kong/plugins/statsd/log.lua
+++ b/kong/plugins/statsd/log.lua
@@ -27,11 +27,6 @@ local range_cache  = setmetatable({}, { __mode = "k" })
 local _M = {}
 
 
--- we use plugin instance ID as the queue ID
-local function get_queue_id()
-  return kong.plugin.get_id()
-end
-
 local function get_cache_value(cache, cache_key)
   local cache_value = cache[cache_key]
   if not cache_value then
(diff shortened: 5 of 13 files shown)


kong/runloop/plugins_iterator.lua
```diff
@@ -149,6 +149,7 @@ local function get_plugin_config(plugin, name, ws_id)
   cfg.route_id    = plugin.route    and plugin.route.id
   cfg.service_id  = plugin.service  and plugin.service.id
   cfg.consumer_id = plugin.consumer and plugin.consumer.id
+  cfg.id          = plugin.id
```
flrgh, line 152: I actually had no idea we populated the plugin config table like this--interesting!

I don't have a strong objection to `id`, but what do you think about using `plugin_id` instead? It would be more consistent with the other entities, and I think it seems _slightly_ less likely to conflict with any plugin that may have an `id` field in its config schema.

fffonion:
hold merge and wait for feedback from the linked ticket

vm-001:
https://github.com/Kong/kong/pull/10011/files#diff-cd6e588c072868fde85e57de9490852ea606f36ea14329cc1d13951aaa90dcdaR92

I have to use `conf.__key__` as the plugin's identifier to solve the issue. I think this will be a big convenience for plugin development. 

By the way. Is it possible to add any tests?

StarlightIbuki:
> https://github.com/Kong/kong/pull/10011/files#diff-cd6e588c072868fde85e57de9490852ea606f36ea14329cc1d13951aaa90dcdaR92
> 
> I have to use `conf.__key__` as the plugin's identifier to solve the issue. I think this will be a big convenience for plugin development.
> 
> By the way. Is it possible to add any tests?

Sure. But let's wait until the decision is made.

kong/runloop/plugins_iterator.lua
vm-001: I was planning to add `plugin_custom_name` here on another [PR](https://github.com/Kong/kong/pull/10077). 

After saw the comment here, I was thinking the Plugin entity adds more and more field during its iteration, populating plugin's meta data into config table flatly is not a correct way, it increases the chances of conflict with user's custom plugin.

Perhaps we can fold the plugin's meta data into `meta` field, or even provide different way to access plugin's meta.

```lua
function MyPlugin:access(conf, plugin) 
  -- access plugin's meta
  local plugin_id = plugin.id
  local plugin_custom_name = plugin.custom_name
end
```

kikito: Using the meta is a bit hacky. I think I prefer adding the plugin as a second parameter.

hbagdi:
@fffonion What is the decision that is needed here?
I saw the linked ticket and couldn't figure out the blocker.

This feature request makes sense and we should ship it.

StarlightIbuki:
> @fffonion What is the decision that is needed here? I saw the linked ticket and couldn't figure out the blocker.
> 
> This feature request makes sense and we should ship it.

The original intention of this feature is to allow users to have an identifier to use as a key for plugin instance-level data. But `conf.__key__` could work for this purpose and it's already used by a lot of plugins.
So this is an option to leave it unchanged and tell users we could use the `__key__`.

hbagdi:
> The original intention of this feature is to allow users to have an identifier to use as a key for plugin instance-level data. But conf.__key__ could work for this purpose and it's already used by a lot of plugins.
So this is an option to leave it unchanged and tell users we could use the __key__.

`__key__` was never intended to be exposed to the user. It is the cache key for the plugin entity and the cache key is an internal detail of Kong that can change across versions - it is not part of the public API of Kong or public to the plugin authors.
I agree with you that it is already used by a few plugins, however, those plugins also use it for "internal to Kong" uses.

From what I understand, the motivation for this addition to give plugin authors an identifier that they can use in the plugin code. Then, the plugin code injects the identifier into requests/data generated from the plugin and sent to third-party systems. Plugin ID is the additional piece of metadata that the user can then use to tie a log entry or a request back to a specific plugin in Kong.
With that use-case in mind, it seems that having a value that the user can provide or look up using Kong Manager or Admin API seems like the right thing to do.

I could be understanding this wrong, please let me know if that is the case.

StarlightIbuki:
@hbagdi I prefer the ID solution too. But let's wait until @fffonion could respond.

fffonion:
If we just want to use an unique stuff, I would suggest to provide a PDK function that returns the current executed plugin's id. Putting a random element in a table will make future usage of this field very hard to track compared to a function, and since there're more than one way to access a `unique stuff` it becomes a bad API design. So my suggestions are:
- Make __key__ "public", in a way of making it part of PDK and properly document it.
- Make a different PDK function, which returns the plugin ID of currently executed, and deprecate __key__ usage

Either way we shouldn't go into a state where both `__key__` and `id` exists and are used actively.

kikito:
Syntax wise, I think putting it on a new `kong.plugin.get_xxx` it is the best option of the other ones presented previously (add second parameter to phase, or insert inside the config as a pseudo-field). Feels safer.

Syntax aside, we seem to be deciding between two options: use `plugin.id` or  use `plugin.__key__`. I have a slight preference on `plugin.id` because:
* It is under the control of the user. They can specify/change it on their configs and/or databases, while they cannot modify `__key__` as easily.
* I think plugin.id is more resistant to a config refresh. I believe `__key__` can change its value if we add new plugins to the plugins list, or if we modify the order in which plugins are presented on the config list. The `id` will remain constant as long as it is present in the configuration (for which we probably should write a note). This is better for plugins that need to resist a config change, or a `kong reload`.

I agree with Wangchong in that having both is confusing, if we can eliminate one, let's do it.

spec/fixtures/custom_plugins/kong/plugins/get-plugin-id/handler.lua
```diff
@@ -0,0 +1,10 @@
+local PluginConfigDumpHandler =  {
+  VERSION = "1.0.0",
+  PRIORITY = 1,
+}
+
+function PluginConfigDumpHandler:access(conf)
+  kong.response.exit(200, kong.plugin.get_id())
```
vm-001, line 7: Or we can write to the headers as we will add more and more `kong.plugin.get_xxx()` function into the pdk/plugin in the future?

StarlightIbuki:
> I think plugin.id is more resistant to a config refresh. 
@kikito Yes. Our internal usages of the `__key__` don't persist outside the memory and the `__key__` changes for every restart. If we are going to support this as a persist one, `__key__` is not even an option.

spec/fixtures/custom_plugins/kong/plugins/get-plugin-id/handler.lua
StarlightIbuki: Not sure if that's a good idea to have a custom plugin for multiple irrelevant tests.
Maybe we could build a plugin to execute arbitrary Lua code?

kong/pdk/plugin.lua
```diff
@@ -0,0 +1,20 @@
+--- Plugin related APIs
+--
+-- @module kong.plugin
+
+
+local _plugin = {}
+
+
+function _plugin.get_id(self)
+  return kong.ctx and kong.ctx.plugin and kong.ctx.plugin.id
```
StarlightIbuki, line 10:
```suggestion
  return kong.ctx and kong.ctx.plugin and kong.ctx.plugin._plugin_id
```

kong/pdk/plugin.lua
```diff
@@ -0,0 +1,20 @@
+--- Plugin related APIs
+--
+-- @module kong.plugin
+
+
+local _plugin = {}
+
+
+function _plugin.get_id(self)
```
vm-001, line 9: Should we make a corresponding adjustment to allow this PDK API to be invoked by other languages? Or it will be exposed automatically?

StarlightIbuki: It will be done by manually running a script.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -153,7 +155,7 @@ end
 function OpenTelemetryHandler:log(conf)
   ngx_log(ngx_DEBUG, _log_prefix, "total spans in current request: ", ngx.ctx.KONG_SPANS and #ngx.ctx.KONG_SPANS)
 
-  local queue_id = conf.__key__
+  local queue_id = get_queue_id()
```
hanshuebner, line 158: Lose the extra indirection and invoke `kong.plugin.get_id` directly (also in the other places that have `local get_queue_id = kong.plugin.get_id`.

hanshuebner (changes requested):

kong/plugins/opentelemetry/handler.lua
hanshuebner: This should go in soon so that #10011 and #10172 can be adapted to use the new `kong.plugin.get_id()` function.

StarlightIbuki: Do you mean we should directly call the `get_id()`? Or we should wrap that with a function?

hanshuebner: Directly calling `kong.plugin.get_id()` is the clearest way.  No wrapping is required.

CHANGELOG.md
```diff
@@ -136,6 +136,13 @@
 
 ## 3.2.0
 
+### Addition
+
+#### PDK
+
+- PDK now support getting plugins' ID with `kong.plugin.get_id`.
```
chronolaw, line 143: should this entry be in `unreleased` section?

chronolaw:
This feature may affect PR #10172 .

CHANGELOG.md
```diff
@@ -83,6 +83,11 @@
 - **ACME**: acme plugin now supports configuring an `account_key` in `keys` and `key_sets`
   [#9746](https://github.com/Kong/kong/pull/9746)
 
+#### PDK
+
+- PDK now support getting plugins' ID with `kong.plugin.get_id`.
```
chronolaw, line 88:
```suggestion
- PDK now supports getting plugins' ID with `kong.plugin.get_id`.
```

CHANGELOG.md
```diff
@@ -139,6 +144,8 @@
 
 ## 3.2.0
 
+### Addition
```
chronolaw, line 147: Remove these two lines.

CHANGELOG.md
```diff
@@ -139,6 +144,8 @@
...
+
```
StarlightIbuki, line 148:
```suggestion
```

kong/plugins/datadog/handler.lua
```diff
@@ -48,8 +48,9 @@ local function compose_tags(service_name, status, consumer_id, tags, conf)
 end
 
 
-local function get_queue_id(conf)
-  return conf.__key__
+-- we use plugin instance ID as the queue ID
```
hanshuebner, line 51: Please remove this function and invoke `kong.plugin.get_id()` directly.  The comment will not be needed then, and the code will be easier to understand.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -37,6 +37,11 @@ local default_headers = {
 local queues = {} -- one queue per unique plugin config
 local headers_cache = setmetatable({}, { __mode = "k" })
 
+-- we use plugin instance ID as the queue ID
+local function get_queue_id()
+  return kong.plugin.get_id()
+end
```
hanshuebner, line 43: Remove this function as well.

kong/plugins/statsd/log.lua
```diff
@@ -27,8 +27,9 @@ local range_cache  = setmetatable({}, { __mode = "k" })
 local _M = {}
 
 
-local function get_queue_id(conf)
-  return conf.__key__
+-- we use plugin instance ID as the queue ID
+local function get_queue_id()
+  return kong.plugin.get_id()
 end
```
hanshuebner, line 33: And this one can go, too.

hanshuebner (changes requested):

kong/pdk/plugin.lua
```diff
@@ -0,0 +1,20 @@
...
+  return ngx.ctx.plugin_id
+end
+
+local function new()
+  return _plugin
+end
```
sabertobihwy, line 15: This function is not used in this PR?

StarlightIbuki: It's used at line 19.

hanshuebner (approved):

PR #9905, opened 2022-12-07
fix(templates): not add default charset when upstream response doesn't contain it

catbro666:
Currently, Kong adds `charset=UTF-8` to `Content-Type` header of response if the content-type is one of `charset_types` and contains no charset.

Kong doesn't know the charset of upstream response and can't assume it is UTF-8. Also we shouldn't change the response unless explicitly required e.g. by the response transformer plugin.

FTI-4551

[FTI-4551]: https://konghq.atlassian.net/browse/FTI-4551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,7 +96,7 @@
   [#9960](https://github.com/Kong/kong/pull/9960)
 - Expose postgres connection pool configuration
   [#9603](https://github.com/Kong/kong/pull/9603)
-- **Template**: not add default charset when upstream response doesn't contain it.
+- **Template**: Do not add default charset to the `Content-Type` response header when upstream response doesn't contain it.
   [#9905](https://github.com/Kong/kong/pull/9905)
 
 #### Plugins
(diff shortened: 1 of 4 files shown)


samugi (approved):

kong/templates/nginx_kong.lua
```diff
@@ -1,5 +1,5 @@
 return [[
-charset UTF-8;
+charset off;
```
dndx, line 2: Why don't we just remove this directive? `charset off` is the default.

CHANGELOG.md
```diff
@@ -70,6 +70,11 @@
 
 ### Fixes
 
+#### Core
+
+- **Template**: not add default charset when upstream response doesn't contain it.
```
dndx, line 75:
```suggestion
- **Template**: Do not add default charset to the `Content-Type` response header when upstream response doesn't contain it.
```

dndx (approved):

PR #8214, opened 2021-12-17
feat: add configuration_hash in /status in dbless

shaneutt:
### Summary

In https://github.com/Kong/kong/issues/8152 it was reported that if the Kong Gateway crashes the KIC will not be aware and unless something changes in the control plane, it will not push a new configuration.

This patch adds a `configuration_hash` to the `/status` endpoint, which would help solve this problem (in DBLESS mode) by allowing the control-plane to keep track of changes to the hash on the data-plane and use that to make decisions about pushing new `/config` updates.

### Issues resolved

Partially resolves https://github.com/Kong/kubernetes-ingress-controller/issues/2107

PR #9911, opened 2022-12-07
fix(declarative-config): set hash when config is loaded

samugi:
Ensure the `hash` field is correctly configured when the declarative configuration is loaded.

### DIFF
--- a/spec/02-integration/04-admin_api/02-kong_routes_spec.lua
+++ b/spec/02-integration/04-admin_api/02-kong_routes_spec.lua
@@ -1,5 +1,6 @@
 local helpers = require "spec.helpers"
 local cjson = require "cjson"
+local constants = require "kong.constants"
 
 local UUID_PATTERN = "%x%x%x%x%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%x%x%x%x%x%x%x%x"
 
@@ -188,6 +189,8 @@ describe("Admin API - Kong routes with strategy #" .. strategy, function()
   end)
 
   describe("/status", function()
+    local empty_config_hash = constants.DECLARATIVE_EMPTY_CONFIG_HASH
+
     it("returns status info", function()
       local res = assert(client:send {
         method = "GET",
@@ -208,7 +211,9 @@ describe("Admin API - Kong routes with strategy #" .. strategy, function()
       assert.is_number(json.server.connections_waiting)
       assert.is_number(json.server.total_requests)
       if strategy == "off" then
-        assert.is_equal(string.rep("0", 32), json.configuration_hash) -- all 0 in DBLESS mode until configuration is applied
+        assert.is_string(json.configuration_hash)
+        assert.equal(32, #json.configuration_hash)
+        assert.is_not_equal(empty_config_hash, json.configuration_hash)
       else
         assert.is_nil(json.configuration_hash) -- not present in DB mode
       end
@@ -251,7 +256,7 @@ describe("Admin API - Kong routes with strategy #" .. strategy, function()
       assert.is_number(json.server.total_requests)
       assert.is_string(json.configuration_hash)
       assert.equal(32, #json.configuration_hash)
-
+      assert.is_not_equal(empty_config_hash, json.configuration_hash)
     end)
 
     it("database.reachable is `true` when DB connection is healthy", function()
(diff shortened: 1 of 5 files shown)


spec/02-integration/04-admin_api/02-kong_routes_spec.lua
```diff
@@ -208,7 +211,9 @@ describe("Admin API - Kong routes with strategy #" .. strategy, function()
       assert.is_number(json.server.connections_waiting)
       assert.is_number(json.server.total_requests)
       if strategy == "off" then
-        assert.is_equal(string.rep("0", 32), json.configuration_hash) -- all 0 in DBLESS mode until configuration is applied
+        assert.is_string(json.configuration_hash)
+        assert.equal(32, #json.configuration_hash)
+        assert.is_not_equal(empty_config_hash, json.configuration_hash)
```
samugi, line 216: Hi @locao! I'd like to confirm this test's logic to ensure I'm not missing anything.
At this point in the test we have the following (default) config loaded:
```
{
  config = "_format_version: '3.0'\n_transform: false\n"
}
```
so I believe `configuration_hash` is not expected to be `00000000000000000000000000000000`, right? This test was failing now that the empty hash problem was addressed via this PR because the hash returned a valid value. Could you have a look and let me know if you agree? Thanks!

mayocream (approved):

spec/02-integration/04-admin_api/02-kong_routes_spec.lua
locao: Hi @samugi! Actually in Kong 2.x the hash was expected be `00000000000000000000000000000000` before any configuration was in use. IIRC that's how KIC knows no config was ever deployed. It's better to check with them if this change is not going to break anything on their side. Wdyt @shaneutt @rainest?

rainest: Yes, as part of https://github.com/Kong/kong/pull/8214 @dndx decided to make that value part of the API surface. [KIC](https://github.com/Kong/kubernetes-ingress-controller/blob/85133ff040d93348c21a2690746b6dc5a223d02f/internal/dataplane/sendconfig/sendconfig.go#L29) and [Koko](https://github.com/Kong/koko/blob/5f8bb40fb30823fbeec8d9bcedc74a0c74babddf/internal/server/kong/ws/config/config_hash.go#L10) both rely on that value.

There's since been renewed interest in an endpoint that indicates readiness without exposing the hash (FTI-3081/a successor to https://github.com/Kong/kong/pull/8256), and I would like to replace the KIC side with that, but can't speak to whether Koko would do the same.

It doesn't look like this seeks to change the value though, correct? KIC only cares if the hash value is zero when there's actually no configuration. If Kong has loaded configuration from the persistence file, it's fine to list the hash for it.

samugi: Interesting, thanks @locao, @rainest 

Here's my take on this:

Currently in 3.x (see issue attached) when a valid configuration is loaded on start (in init_worker, from file) the hash is always returned as `00000000000000000000000000000000`, which, as I understand, is also bad for KIC because Kong would make it believe there was no configuration loaded.

With this PR, however, the default (empty?) configuration would return its own hash: `e2c700992dee69de1e7c4b924ab28273`. If we want the empty configuration (`"_format_version: '3.0'\n_transform: false\n"`) to return `00000000000000000000000000000000` then we would need some dedicated logic to make that happen, on top of this fix. This way we would always have the expected hash. Does that make sense?

rainest: Ah, okay, that jogs my memory: although KIC wouldn't care if Kong loads an actual configuration from disk, Kong is "always" loading a configuration from disk at start, because it thinks it's correctly loading an empty configuration and doesn't distinguish it specially. So yeah, without added logic to recognize that, it'd be a problem.

samugi: I added the logic mentioned above and respective tests

kong/db/declarative/init.lua
```diff
@@ -660,6 +660,11 @@ end
 declarative.unique_field_key = unique_field_key
 
 
+local function config_is_empty(entities)
+  -- empty configuration has no entries other than workspaces
+  return entities["workspaces"] and tablex.size(entities) == 1
```
dndx, line 665: You can use https://github.com/openresty/luajit2#tablenkeys for `tablex.size`. Might be faster for large configs.

Also, do we need to do `and entities.workspace` just to be safe? I'm not sure if it is possible to have an `entities` that has not workspace or will it always be added by the parser.

chronolaw:
Due to #9890, we should rebase this PR.

samugi:
> Due to #9890, we should rebase this PR.

thank you @chronolaw, done.

dndx (approved):

PR #9920, opened 2022-12-09
feat(core): add https_sni to upstream entities

### LINKED ISSUE #9785: SNI can't be used if upstream target definition using IP:Port

- [x] I have searched the existing issues
Linked discussion:
https://github.com/Kong/kong/discussions/9247

2.8.1

### Current Behavior

One of our applications is using Kong for HTTPS connection to external system. During the communication we also must send SNI hostname as a value which is CN/SAN value of the server certificate or a hostname which is covered by a CN/SAN of wildcard server certificate.
_CN= Common Name
SAN=Subject Alternative Name_
> Problem: SNI based upstream target can not be used if upstream needs to be called based on ip-address and port

We cannot resolve the DNS name for the upstream partner (different company private network without possible DNS resolution on our side).
For this reason we need to use IP-address and port in upstream / target configuration to be able to use SNI based endpoint selection on the upstream partner side during TLS handshake a correct SNI value needs to be send correlating to the server certificate they use on their side.
(They also run on their side ingress controller which selects endpoint based on information of hostname / sni-name)
This is currently not possible in kong to configure it this way.

We tried to analyze current implementation of SNI and it looks only on upstream name and does not look for any sni configuration with or without client-certificate configurations.
>
[SNI code](https://github.com/Kong/kong/blob/2.8.3/kong/templates/nginx_kong_stream.lua#L125) which looks only on [upstream](https://github.com/Kong/kong/blob/2.8.3/kong/templates/nginx_kong_stream.lua#L76)

### Expected Behavior

When we request a particular service with a configured route in kong it should look to the configured kong service which references the relevant upstream and finally upstream targets.
The upstream config needs to be extended with an optional value of **“https_sni”** which needs to be used during the TLS handshake with the upstream target host and port.
This **_SNI-name_** should be by default the name of the upstream object of kong. 
If an optional value of **“https_sni”** is configured, this value needs to be used for the TLS handshake as **SNI** value.

_Additional hint:
The possibility to configure **“https_sni”** on kong upstream objects is already existing for the “healthcheck” configuration possibility. But not for the upstream object and the configured targets behind the upstream config itself._

### Steps To Reproduce

```markdown
1.	Create service which will be looking on some service, e.g haproxy with nginx as background
- should have multiple server certificates listening to multiple hostnames. All hostnames need to use the same IP address.
2.	Create route /test-route in kong
3.	Configure service and upstream using ip address and port of the haproxy or nginx upstream server
Use the name “upstream-name.remote” for the upstream object in the kong configuration
4.	Run some traffic sniff on kong instance, e.g tcpdump(or kubectl sniff pod) to capture a traffic
5.	Run a curl or postman GET request to https://kong/test-route
6.	Check traffic sniffed output for SNI.
(There should be the name of the upstream object “upstream-name.remote” as value of SNI during the handshake be visible, which is not the case)
```

### How it should work after implementation:
1.	Create service which will be looking on some service, e.g haproxy with nginx as background should have multiple server certificates listening to multiple hostnames. All hostnames/certificates need to use the same ip-address:

- -	default (should be default certificate)
- -	upstream-name.remote
- -	upstream-name.overwritten

These are the hostnames of upstream destination service which needs to be used for this test case.

2.	Create route /test-route in kong
3.	Configure service and upstream using IP address and Port of the haproxy or nginx upstream server
Use the name **“upstream-name.remote”** for the upstream object in the kong configuration
4.	Configure for the upstream object **“upstream-name.remote”** the optional **“https_sni”** key to the value **“upstream-name.overwritten”**
5.	Run some traffic sniff on kong instance, e.g tcpdump to capture a traffic
6.	Run a curl or postman GET request to https://kong/test-route
7.	Check traffic sniffed output for SNI.
(There should be the value of **“upstream-name.overwritten”** used as **SNI** name during the TLS handshake).

gndrmnn:
### Summary

Hi everyone,
When connecting to an upstream service via https, two things happen: First, the Server Name field of the SNI handshake is filled. And after the certificate has been exchanged and the encrypted connection is established, the "host" http header is filled, usually with the same hostname.

When configuring an upstream object in Kong with a target using an IP, no SNI name or host name is specified by default. ``host_header`` can be configured for an upstream object, but this sets both the SNI field and the host header. Our customer wants to be able to explicitly set the SNI, so this PR adds the option ``https_sni`` to **only** set the SNI field value without touching the ``host`` http header that follows after the encryption. Note that the option name ``https_sni`` is also used in upstream healthcheck configuration.

Discussion about this is welcome.

This is my first time working on the Kong codebase and I am currently trying to write an integration test for this change. Any hint as to where to add it, or how exactly to test it is much appreciated.

### Checklist

- [X] The Pull Request has unit tests
- [ ] The Pull Request has integration tests

### Issue reference

Fix #9785
See also discussion #9247


attenuation:
Sorry, Kong use `upstream_host` var as kong upstream sni name in the balancer execute, you can not change this in the `proxy_ssl_name`  command, this is not Okay for kong.

flrgh:
> Sorry, Kong use `upstream_host` var as kong upstream sni name in the balancer execute, you can not change this in the `proxy_ssl_name` command, this is not Okay for kong.

@attenuation I'm with you on being careful about the naming. This definitely needs a thorough review and test cases to be certain there are no backwards-incompatible things hidden in here, but in my opinion it is something that Kong should support.

The need to send differing values for SNI and Host header is a _really_ frustrating use case, but it is a legitimate one; I have encountered this in the wild several times myself. There are many scenarios where users have no control over the upstream server themselves, so not supporting this use case in fact a dead end for some.

hbagdi:

PR #9671, opened 2022-11-03
refactor(clustering): gather clustering events into one place

chronolaw:
### Summary

To prepare for incremental sync and other features,
We must do some code clean for event propagation.

### Full changelog

* Add a new file `events.lua`
* Move event register functions into `events.lua`
* New function `events.clustering_push_config`

PR #9921, opened 2022-12-09
feat(clustering): revert removal of JSON-websocket support for config sync

chronolaw:
### Summary

Try to restore WebSocket protocol and remove wRPC protocol.

Revert PR #9740, update to PR #9823  #9769 , #9616 #9671 #9774.

There is still some code need to optimize, but I will leave them to another PR.

See KAG-210.

Related PRs: #8926 #8357

### Checklist

### Full changelog

* revert PR #9740
* update `control_plane.lua` with new `compat.*`
* change `compat.update_compatible_payload` to fit json+gzip
* remove wrpc related code/config/tests


tyler-ball:
I want to get @flrgh's opinion of this. My suggestion would be to not remove the wrpc files completely, but just disable their usage. Or is there a high confidence that wrpc is going away and never coming back?

flrgh:
We'll need to come up with a game plan for this. I suspect that we can't fully remove wRPC in a single release, because it will break data planes running 3.1.0 and 3.1.1. We may have to support both protocols for a release cycle or two.

kong/clustering/compat/init.lua
```diff
@@ -353,4 +356,56 @@ function _M.update_compatible_payload(config_table, dp_version, log_suffix)
...
+  local dp_version_num = version_num(dp_version)
+
+  -- if the CP and DP have the same version, avoid the payload
+  -- copy and compatibility updates
+  if cp_version_num == dp_version_num then
+    return false
+  end
+
+  local has_update
+  payload = deep_copy(payload, false)
```
StarlightIbuki, line 371: Seems we are deep-copying unconditionally. Could we also include the lazy-eval optimization, just like what we did to wrpc version of this function?

kong/clustering/compat/init.lua
```diff
@@ -310,7 +313,7 @@ end
 
 
 -- returns has_update, modified_config_table
-function _M.update_compatible_payload(config_table, dp_version, log_suffix)
+function _M.wrpc_update_compatible_payload(config_table, dp_version, log_suffix)
   local cp_version_num = version_num(meta.version)
```
StarlightIbuki, line 317: Should we also do

```lua
local cp_version_num = version_num(KONG_VERSION)
```

kong/clustering/control_plane.lua
```diff
@@ -0,0 +1,534 @@
...
+  end
+
+  local n = 0
+  for _, queue in pairs(self.clients) do
+    table_insert(queue, RECONFIGURE_TYPE)
+    queue.post()
+    n = n + 1
+  end
+
+  ngx_update_time()
```
StarlightIbuki, line 159: Just a note. We don't have this in 2.8.

Should we check the log level before doing this? `ngx_update_time` is not free.

kong/clustering/control_plane.lua
```diff
@@ -0,0 +1,534 @@
...
+  end
+
+  while not exiting() do
+    local ok, err = push_config_semaphore:wait(1)
+    if exiting() then
+      return
+    end
+
+    if ok then
+      if isempty(self.clients) then
```
StarlightIbuki, line 462: It looks like we do not need a queue here. Just a counter is enough. Ref: https://github.com/Kong/kong/blob/b42d53bcc694630ddb4bdafda2c645ceab7e5823/kong/clustering/control_plane.lua#L154-L155

kong/clustering/control_plane.lua
```diff
@@ -0,0 +1,534 @@
...
+
+        ngx_log(ngx_DEBUG, _log_prefix, "received ping frame from data plane", log_suffix)
+
+        config_hash = data
+        last_seen = ngx_time()
+        update_sync_status()
+
+        -- queue PONG to avoid races
+        table_insert(queue, PONG_TYPE)
+        queue.post()
```
StarlightIbuki, line 347:
```suggestion
        queue:post()
```

kong/clustering/init.lua
```diff
@@ -44,7 +46,16 @@ function _M.new(conf)
 end
 
 
+function _M:handle_cp_websocket()
+  return self.json_handler:handle_cp_websocket()
+end
+
+
 function _M:handle_wrpc_websocket()
+  if ngx.req.get_method() == "HEAD" then -- force fallback
```
StarlightIbuki, line 55: Any reference for me to understand this change?

kong/clustering/init.lua
```diff
@@ -53,22 +64,31 @@ function _M:init_cp_worker(plugins_list)
...
 
+
 function _M:init_dp_worker(plugins_list)
   local start_dp = function(premature)
     if premature then
       return
     end
 
-    self.child = require("kong.clustering.wrpc_data_plane").new(self.conf, self.cert, self.cert_key)
+    if kong.configuration.wrpc_hybrid_protocol then
```
StarlightIbuki, line 78: We should either make this happen at the init phase or move the require to file scope.

kong/conf_loader/init.lua
```diff
@@ -541,6 +541,7 @@ local CONF_INFERENCES = {
   untrusted_lua_sandbox_environment = { typ = "array" },
 
   legacy_worker_events = { typ = "boolean" },
+  wrpc_hybrid_protocol = { typ = "boolean" },
```
StarlightIbuki, line 544: Should we remove the `legacy_worker_events` switch?

kong/templates/kong_defaults.lua
```diff
@@ -182,6 +182,7 @@ untrusted_lua_sandbox_requires =
 untrusted_lua_sandbox_environment =
 
 legacy_worker_events = off
+wrpc_hybrid_protocol = off
```
StarlightIbuki, line 185: Same here.

kong/conf_loader/init.lua
StarlightIbuki: Do we have a test for this switch?

CHANGELOG.md
```diff
@@ -91,11 +91,19 @@
 - **Zipkin**: Fix an issue where the global plugin's sample ratio overrides route-specific.
   [#9877](https://github.com/Kong/kong/pull/9877)
 
+### Changed
+
+#### Hybrid Mode
+
+- Restore the original WebSocket protocol, and still support wRPC protocol.
+  [#9921](https://github.com/Kong/kong/pull/9921)
```
StarlightIbuki, line 99:
```suggestion
- Revert the removal of WebSocket protocol support for configuration sync.
  [#9921](https://github.com/Kong/kong/pull/9921)
```

StarlightIbuki: Just a suggestion. We need to think more carefully about the wording.

StarlightIbuki:
The new solution (to preserve wrpc) should be another PR so we don't have those unnecessary commit histories, and at the same time, we could keep a clean version of the solution to remove wrpc.
Also, please attach the result of the compatibility tests here.

kong/clustering/control_plane.lua
chronolaw: No, `queue` is a simple wrapper for semaphore, it can not use `:`

kong/clustering/init.lua
chronolaw: `HEAD` for endpoint `v1/wrpc` will return 404, which will disable 3.0 DP's wrpc

kong/conf_loader/init.lua
chronolaw: `wrpc_hybrid_protocol` is only for wrpc test, it will not be enabled in the field.

chronolaw:
The purpose of this PR is to restore the usage of the WebSocket protocol, I try to not introduce too many code changes, and make the code back to the original form.

I think that the first thing is getting the code back and letting them work.

kong/clustering/init.lua
chronolaw: `data_plane` will run in only one worker process, so it need not to be inited for all workers.

kong/clustering/control_plane.lua
chronolaw: It belongs 3.0.x, I just revert the code and do touch anything.

StarlightIbuki (approved):

kong/clustering/init.lua
flrgh, line 78: Proposal/idea: what if we reuse the `legacy_hybrid_protocol` configuration param that was introduced in 3.0? We can simply set the default value to `on`/`true` instead of `off`/`false`?

Reason: it will make it easier to backport any of these changes to the 3.0.x series, if we choose to do so.

flrgh (approved):
Assuming we're going to keep wRPC around for another release or two, this looks good to me :+1: thanks @chronolaw for your hard work

kong/clustering/init.lua
chronolaw: `legacy_hybrid_protocol` is not suitable for present protocol, which is not legacy anymore. so I use another more clarify flag `wrpc_hybrid_protocol`

flrgh (approved):

PR #9940, opened 2022-12-12
fix(router/atc): pcre regex compatibility

chronolaw:
### Summary

The new atc-router does not support some PCRE regex escape characters, 
this PR removes some special escape characters, tries to be compatible with PCRE.

See FTI-4446

### Checklist

- [x] There is not a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE

### Full changelog

* function sanitize_regex
* update test cases


chronolaw:
Now this fix only works in `traditional_compatible`, should we do the same thing in `expressions`?

kong/router/compat.lua
```diff
@@ -133,8 +134,7 @@ local function get_expression(route)
     if op == OP_REGEX then
       -- 1. strip leading `~`
       -- 2. prefix with `^` to match the anchored behavior of the traditional router
-      -- 3. update named capture opening tag for rust regex::Regex compatibility
-      return "^" .. p:sub(2):gsub("?<", "?P<")
+      return sanitize_regex("^" .. p:sub(2))
```
flrgh, line 137: What do you think of moving _all_ the string modifications into a single function?

```suggestion
      return prepare_path_regex(p)
```

chronolaw: This is for compatibility with PCRE, all things to do are string substitute.

kong/router/atc.lua
```diff
@@ -96,6 +96,16 @@ local function escape_str(str)
 end
 
 
+local function sanitize_regex(str)
+  -- update named capture opening tag for rust regex::Regex compatibility
+  return str:gsub([[?<]] , [[?P<]])
+            :gsub([[\/]] , [[/]])  -- "\/" => "/"
+            :gsub([[\_]] , [[_]])  -- "\_" => "_"
+            :gsub([[\!]] , [[!]])  -- "\!" => "!"
+            :gsub([[\ ]] , [[ ]])  -- "\ " => " "
```
hanshuebner, line 105: These string substitutions will cause regular expressions like ` ?<` or intended uses of `\\` before of `/`, `_`, `!` or ` ` to be changed unexpectedly.  We should at least cover for these degenerate cases by making the match expressions longer as in. `(?<` and whatever Lua match expression would work for making sure that the backslashes are not preceded by another backslash.

This could still fail to work as designed, leaving users with hard to understand behavior, but it would reduce the likeliness of creating such unexpected behavior.  It may not be very common to use complicated regular expressions that would fail the substitution, but if they are used, it would be very difficult to diagnose what the issue is.

hanshuebner (changes requested):

chronolaw:
After some discussion, we decide to use another solution.

PR #9946, opened 2022-12-13
fix(jwt): deny requests that have different tokens in the jwt token search locations.

vm-001:
### Summary

Deny requests that have different tokens in the jwt token search locations.

### Checklist

### Issue reference

[FTI-4580]

[FTI-4580]: https://konghq.atlassian.net/browse/FTI-4580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,7 +94,7 @@
 
 - **Zipkin**: Fix an issue where the global plugin's sample ratio overrides route-specific.
   [#9877](https://github.com/Kong/kong/pull/9877)
-- **JWT**: Deny requests that have different tokens in the jwt token search locations. Thanks, Jackson 'Che-Chun' Kuo from Latacora.
+- **JWT**: Deny requests that have different tokens in the jwt token search locations. Thanks Jackson 'Che-Chun' Kuo from Latacora for reporting this issue.
   [#9946](https://github.com/Kong/kong/pull/9946)
 
 ### Dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,9 +69,16 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+#### Plugins
+
+- **JWT**: JWT plugin now denies a request that has different tokens in the jwt token search locations.
+  [#9946](https://github.com/Kong/kong/pull/9946)
+
 ### Additions
 
-### Plugins
+#### Plugins
 
 - **Zipkin**: Add support to set the durations of Kong phases as span tags
   through configuration property `config.phase_duration_flavor`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,8 @@
 
 - **Zipkin**: Fix an issue where the global plugin's sample ratio overrides route-specific.
   [#9877](https://github.com/Kong/kong/pull/9877)
+- **JWT**: Deny requests that have different tokens in the jwt token search locations. Thanks, Jackson 'Che-Chun' Kuo from Latacora.
+  [#9946](https://github.com/Kong/kong/pull/9946)
 
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -83,6 +83,8 @@
 
 - **Zipkin**: Fix an issue where the global plugin's sample ratio overrides route-specific.
   [#9877](https://github.com/Kong/kong/pull/9877)
+- **JWT**: Deny requests that have different tokens in the jwt token search locations. Thanks, Jackson 'Che-Chun' Kuo from Latacora.
```
dndx, line 86:
```suggestion
- **JWT**: Deny requests that have different tokens in the jwt token search locations. Thanks Jackson 'Che-Chun' Kuo from Latacora for reporting this issue.
```

dndx (changes requested):
The general approach of this PR is fine. However, I see no good reason of returning a Lua table from `retrieve_tokens` just for duplicate detection. Maybe the `retrieve_tokens` function can simply return `nil, "Multiple tokens provided"` in case more than one unique token is present in the same request. This way we avoid two table allocations and use of `pairs` which could be expensive.

bungle (approved):

kong/plugins/jwt/handler.lua
```diff
@@ -60,10 +73,27 @@ local function retrieve_token(conf)
...
+        if m[1] ~= "" then
+          token_set[m[1]] = true
+        end
       end
     end
   end
+
+  local tokens = {}
+  for token, _ in pairs(token_set) do
+    table_insert(tokens, token)
```
dndx, line 85: If we have to keep the return value a table. Let's not use `table_insert` in here if possible. Instead, use a variable `tokens_n` to keep track of number of items and insert using tokens[tokens_n] instead.

The multiple `#tokens` checks below can also be removed this way.

vm-001: Good suggestion.

CHANGELOG.md
```diff
@@ -83,6 +83,8 @@
 
 - **Zipkin**: Fix an issue where the global plugin's sample ratio overrides route-specific.
   [#9877](https://github.com/Kong/kong/pull/9877)
+- **JWT**: Deny requests that have different tokens in the jwt token search locations. Thanks Jackson 'Che-Chun' Kuo from Latacora for reporting this issue.
+  [#9946](https://github.com/Kong/kong/pull/9946)
```
hbagdi, line 87: Would it make sense to call this out as a "potential breaking change" for some users?

vm-001: In the scenario that carries both a JWT token in the query string and another non-JWT token in the JWT header search location in the same request, then yes, this is a breaking change.  In my opinion, this usage is quite strange.

hbagdi: I agree that the usage is quite strange. I've encountered [Hyrum's Law](https://www.hyrumslaw.com/) too many times to assume that this won't be a problem for anyone at all. I hope that is the case. Documenting the change of behavior helps us defend the reason for this change and also educate users.

We should also call this change out in the plugin manual.

CHANGELOG.md
```diff
@@ -69,9 +69,16 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+#### Plugins
+
+- **JWT**: JWT plugin now denies a request that has different tokens in the jwt token search locations.
```
hbagdi, line 76:
```suggestion
- **JWT**: JWT plugin now denies a request that has different JWTs in the HTTP request header and Query parameter.
```

vm-001: There are three search locations, which are querystring(query parameter), header, and cookie.

dndx (approved):

PR #9949, opened 2022-12-13
chore(ci): replace kbt with bazel build

mayocream:
### Summary

Remove KBT


fffonion (approved):

colinh-kong:
I don't think we should do this until the build systems are equivalent in all the important ways which they currently are not.

PR #9951, opened 2022-12-13
Revert "chore(ci): replace kbt with bazel build; token is now not nee…

mayocream:
…ded for forks to run CI (#9949)"

This reverts commit df741d6372c6b17a77ad4000a0f23d71a9f82145.


mayocream:
Fixed by #9950

Issue #9944, opened 2022-12-12
Admin API returns float for key-auth TTL


rainest:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

master-alpine

### Current Behavior

Creating or accessing a key-auth with `ttl` set returns a float value for `ttl`:

```
09:39:09-0800 esenin $ http 192.168.96.49:8001/consumers/foo/key-auth key=foo ttl:=999
HTTP/1.1 201 Created

{
    "consumer": {
        "id": "cd8546b9-da5d-4352-b166-43974a1f9e52"
    },
    "created_at": 1670607554,
    "id": "3c184561-5534-4d8d-8d4d-cf4c86fd2f3a",
    "key": "foo",
    "tags": null,
    "ttl": 998.44457
}
```

### Expected Behavior

Creating or accessing a key-auth with `ttl` set returns a floored int value for `ttl`.

```
09:39:09-0800 esenin $ http 192.168.96.49:8001/consumers/foo/key-auth key=foo ttl:=999
HTTP/1.1 201 Created

{
    "consumer": {
        "id": "cd8546b9-da5d-4352-b166-43974a1f9e52"
    },
    "created_at": 1670607554,
    "id": "3c184561-5534-4d8d-8d4d-cf4c86fd2f3a",
    "key": "foo",
    "tags": null,
    "ttl": 998
}
```

### Steps To Reproduce

```markdown
1. Create a consumer, `http :8001/consumers username=example`
2. Create a key-auth with some arbitrary large TTL, `http :7001/consumers/example/key-auth key=alpha ttl:=n`. The response will include a float `ttl=m`, where `n > m > n-1` (unless your instance is _really_ slow).
3. Create a key-auth with `ttl=1`, `http :7001/consumers/example/key-auth key=alpha ttl:=1`. The response will include a float `ttl=m`, where `1 > m > 0`

Repeat and compare the same on a released image, e.g. 3.1.0.
```

### Anything else?

https://github.com/Kong/kong/pull/9801 changed a Postgres `FLOOR(EXTRACT(EPOCH FROM ([expiration timestamp] - [current timestamp])))` to remove the floor. The change indicates that this was necessary to deal with a bug where the floored `0` value would prevent the key from expiring in internal caches. It _does not_ state that returning floats in the API was an expected or intended side effect.

This may in fact be intended, but it should be marked in the changelog as such, since it can break downstream clients.

[...]

PR #9960, opened 2022-12-14
fix(db): add back `FLOOR` so ttl returns integer instead of float

outsinre:
### Summary

Add back `FLOOR` so that `ttl` is returned as integer instead of float.

PR <https://github.com/Kong/kong/pull/9801> remove `FLOOR` part to reflect the real `ttl` value, but may return a `float` value instead of `int`, which would break some downstream clients as depicted in issue <https://github.com/Kong/kong/issues/9944>.

This PR add back `FLOOR` as in the `handler.lua`, we have double-check on `cred.ttl == 0`. So it is **safe** to add back.

Also check [FTI-4066].

### Issue reference

Fix #_[9944]_

[FTI-4066]: https://konghq.atlassian.net/browse/FTI-4066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@
 
 ## Unreleased
 
-## Additions
+### Additions
 
 ### Plugins
 
@@ -78,6 +78,12 @@
   [#9891](https://github.com/Kong/kong/pull/9891)
 
 ### Fixes
+
+#### Core
+
+- Add back Postgres `FLOOR` function when calculating `ttl`, so it is always returned
+  as `int` instead of `float`.
+  [#9960](https://github.com/Kong/kong/pull/9960)
 
 #### Plugins
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,8 +81,7 @@
 
 #### Core
 
-- Add back Postgres `FLOOR` function when calculating `ttl`, so it is always returned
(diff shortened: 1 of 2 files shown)


chronolaw (approved):

chronolaw:
Could we add a test case to catch this bug?

outsinre:
> Could we add a test case to catch this bug?

We just add back a removed part, which has built-in test already.

But also need verification from downstream clients that relies on `int` return, like [9944].

chronolaw:
We should add a change log entry too.

outsinre:
> We should add a change log entry too.

Done.

CHANGELOG.md
```diff
@@ -79,6 +79,12 @@
 
 ### Fixes
 
+#### Core
+
+- Add back Postgres `FLOOR` function when calculating `ttl`, so it is always returned
+  as `int` instead of `float`.
```
dndx, line 85:
```suggestion
- Add back Postgres `FLOOR` function when calculating `ttl`, so the returned `ttl` is always a whole integer.
```

dndx (approved):

PR #9961, opened 2022-12-14
chore(db): add the id of ca_certificates entity to error log of custom_entity_check

catbro666:
Helps to pinpoint the specific certificate that went wrong.

FTI-4593

### DIFF
--- a/kong/db/schema/entities/ca_certificates.lua
+++ b/kong/db/schema/entities/ca_certificates.lua
@@ -29,7 +29,7 @@ return {
 
   entity_checks = {
     { custom_entity_check = {
-      field_sources = { "cert", },
+      field_sources = { "cert", "id" },
       fn = function(entity)
         local seen = false
         for _ in string.gmatch(entity.cert, "%-%-%-%-%-BEGIN CERTIFICATE%-%-%-%-%-") do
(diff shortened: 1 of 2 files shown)


samugi (approved):

kong/db/schema/entities/ca_certificates.lua
```diff
@@ -45,11 +45,11 @@ return {
...
 
         if not_after < now then
-          return nil, "certificate expired, \"Not After\" time is in the past"
+          return nil, "certificate " .. entity.id .. " expired, \"Not After\" time is in the past"
         end
 
         if not cert:get_basic_constraints("CA") then
-          return nil, "certificate does not appear to be a CA because " ..
+          return nil, "certificate " .. entity.id .. " does not appear to be a CA because " ..
                       "it is missing the \"CA\" basic constraint"
```
hbagdi, line 53: Isn't this a deviation from what is done in other custom_entity_check functions in core entity schemas? None of them seem to include ID of the entity in the error response from the schema library.

The issue here is more widespread. Whenever validation fails, the logs never tell you which specific resource failed validation and we should consider adding a systemic fix rather than solving specifically for CA certificates.

What do you think?

catbro666:
https://github.com/Kong/kong/pull/9806 is a generic way to solve this issue, so close this one.

PR #8382, opened 2022-02-08
Support for IMDSv2

isavcic:
### Full changelog

* Added support for IMDSv2 by fetching an IMDSv2 token prior to using it in subsequent Metadata Service requests

### Issue reference

Fixes #8380

Issue #8380, opened 2022-02-07
AWS Lambda plugin does not function when Instance Metadata Service v2 is required


hrast01:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Kong version (`$ kong version`)

Kong 2.0.6 (checked master branch plugin appears non-functional as well))

### Current Behavior

With Instance Metadata Service v2 (IMDSv2) enabled, AWS Lambda functions do not get invoked via the AWS Lambda plugin.

### Expected Behavior

Lambdas should be properly invoked regardless of the IMDS version.

### Steps To Reproduce

```markdown

1. Kong node assigned Invoke and InvokeAsync Lambda permissions.
2. Update the Launch Template to require the use of Instance Metadata Service v2 (HttpTokens: optional to HttpTokens: required)
3. The lambda ceased being invoked
4. Revert "HttpTokens: required" change , AWS Lambda invocation was successful.

```

### Anything else?

AWS doc describing the process for IMDSv2.
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

hanshuebner:
Optional IMDSv2 support will be part of the next release of Kong Gateway (see #9962)

PR #9962, opened 2022-12-14
Support IMDSv1 and IMDSv2

hanshuebner:
This supersedes https://github.com/Kong/kong/pull/8382 by Ivan Savcic <isavcic@gmail.com> to fix tests.

Fixes FTI-4536 FTI-1889 
Relates to CSRE-663

### DIFF
--- a/kong/plugins/aws-lambda/iam-ec2-credentials.lua
+++ b/kong/plugins/aws-lambda/iam-ec2-credentials.lua
@@ -9,14 +9,54 @@ local kong = kong
 local METADATA_SERVICE_PORT = 80
 local METADATA_SERVICE_REQUEST_TIMEOUT = 5000
 local METADATA_SERVICE_HOST = "169.254.169.254"
-local METADATA_SERVICE_URI = "http://" .. METADATA_SERVICE_HOST .. ":" .. METADATA_SERVICE_PORT ..
+local METADATA_SERVICE_TOKEN_URI = "http://" .. METADATA_SERVICE_HOST .. ":" .. METADATA_SERVICE_PORT ..
+                             "/latest/api/token"
+local METADATA_SERVICE_IAM_URI = "http://" .. METADATA_SERVICE_HOST .. ":" .. METADATA_SERVICE_PORT ..
                              "/latest/meta-data/iam/security-credentials/"
 
 
-local function fetch_ec2_credentials()
+local function fetch_ec2_credentials(config)
   local client = http.new()
   client:set_timeout(METADATA_SERVICE_REQUEST_TIMEOUT)
-  local role_name_request_res, err = client:request_uri(METADATA_SERVICE_URI)
+
+  local protocol_version = config.aws_imds_protocol_version
+  local imds_session_headers
+
+  if protocol_version == "v1" then
+
+    -- When using IMSDv1, the role is retrieved with a simple GET
+    -- request requiring no special headers.
+    imds_session_headers = {}
+
+  elseif protocol_version == "v2" then
+
+    -- When using IMSDv1, the role is retrieved with a GET request
+    -- that has a valid X-aws-ec2-metadata-token header with a valid
+    -- token, which needs to be retrieved with a PUT request.
+    local token_request_res, err = client:request_uri(METADATA_SERVICE_TOKEN_URI, {
+        method = "PUT",
+        headers = {
+          ["X-aws-ec2-metadata-token-ttl-seconds"] = "60",
+        },
+    })
+
+    if not token_request_res then
+      return nil, "Could not fetch IMDSv2 token from metadata service: " .. tostring(err)
+    end
+
+    if token_request_res.status ~= 200 then
+      return nil, "Fetching IMDSv2 token from metadata service returned status code " ..
+        token_request_res.status .. " with body " .. token_request_res.body
+    end
+    imds_session_headers = { ["X-aws-ec2-metadata-token"] = token_request_res.body }
+
+  else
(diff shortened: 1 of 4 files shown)


hanshuebner:
We get honorable mention:

 - https://github.com/SummitRoute/imdsv2_wall_of_shame/issues/17
 - https://github.com/Kong/kong/issues/8380

kong/plugins/aws-lambda/iam-ec2-credentials.lua
```diff
@@ -9,29 +9,71 @@ local kong = kong
...
+local function fetch_ec2_credentials(config)
   local client = http.new()
   client:set_timeout(METADATA_SERVICE_REQUEST_TIMEOUT)
-  local role_name_request_res, err = client:request_uri(METADATA_SERVICE_URI)
+
+  local protocol_version = config.aws_imds_protocol_version
+  local imds_session_headers
+
+  if protocol_version == "v1" then
+
```
windmgc, line 26:
```suggestion
```

windmgc (approved):
Code LGTM with nitpick, and please add a user doc for the new config field.

Tested on EC2 with both IMDSv1/v2 and it seems working well.

Perhaps we should consider change the default to v2 in next major version as I saw some of these official SDKs are using IMDSv2 as a default behavior(e.g. aws-go-sdk)

kong/plugins/aws-lambda/schema.lua
```diff
@@ -95,6 +95,12 @@ return {
           type = "boolean",
           default = true,
         } },
+        { aws_imds_protocol_version = {
+          type = "string",
+          required = true,
+          default = "v1",
+          one_of = { "v1", "v2" }
+        } },
```
hbagdi, line 103: Why is the code asking for this input from the user?
Why is the code not smart enough to use v2 first and then fallback on v1 if v2 is not available?
I don't see any difference in functionality (so far as the current use of the API is concerned) between the two versions.

Am I missing something?

hbagdi:
Is this missing a documentation PR?
If you have already created one, please reference it from here.

kong/plugins/aws-lambda/schema.lua
hanshuebner: By default, EC2 instances support both v1 and v2, and it is a concious decision for the AWS administrator to disable v1.  While it would generally be possible to default to v2, this would create the problem that if Kong is running inside a container on the EC2 instance, the IMDS request will hang and eventually time out.  I think this would be difficult to debug, and as the switch to v2-only is a configuration option that needs to be conciously made, requiring the plugin to be configured accordingly seems sensible.

hanshuebner:
> Is this missing a documentation PR?
It is - Created KAG-317 to track the activity

kong/plugins/aws-lambda/iam-ec2-credentials.lua
```diff
@@ -9,29 +9,71 @@ local kong = kong
...
+    -- When using IMSDv1, the role is retrieved with a simple GET
+    -- request requiring no special headers.
+    imds_session_headers = {}
+
+  elseif protocol_version == "v2" then
+
+    -- When using IMSDv1, the role is retrieved with a GET request
+    -- that has a valid X-aws-ec2-metadata-token header with a valid
+    -- token, which needs to be retrieved with a PUT request.
```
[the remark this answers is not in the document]
hanshuebner: Nice catch, this was supposed to say `IMDSv2` instead.

kong/plugins/aws-lambda/schema.lua
hbagdi: > this would create the problem that if Kong is running inside a container on the EC2 instance, the IMDS request will hang and eventually time out.

What problem do we run into if v2 request is made with a reasonable timeout and if the timeout is reached then v1 is used for the lifetime of the process?

hanshuebner: > What problem do we run into if v2 request is made with a reasonable timeout and if the timeout is reached then v1 is used for the lifetime of the process?

The first request to the plugin will be delayed by the timeout.

hbagdi: Yes and that sounds like an acceptable compromise given that we make the user care about one less thing.
What do you think?

hanshuebner: I'm not convinced as the behavior would be kind of magical and our supporting code would be more complex.  That said, if you want me to make that change, I'll look into it next week.

hanshuebner: After a conversation on Slack, we've decided to leave things as they are - with explicit configuration - as automatically detecting the IMDS protocol version would potentially impact Kong's startup time too much.

kong/plugins/aws-lambda/iam-ec2-credentials.lua
hanshuebner: Sorry, I overlooked this comment - I opened a new PR to get rid of the style issue and will create a matching docs update.

If you could approve https://github.com/Kong/kong/pull/9985, it'd be appreciated.

hanshuebner:
@windmgc Harry asked whether we could auto-detect the IMDS protocol version and we discussed the matter in Slack.  The issue with defaulting to IMDSv2 is that with container deployments in EC2, the EC2 instance metadata needs to be explicitly configured to make IMDSv2 work.  It would be possible to autodetect the protocol version, but that'd come at the cost of delaying the initial requests making use of the AWS Lambda plugin.  We feel that this would impact too many customers, so the configuration of the protocol version will remain explicit for now.

PR #9963, opened 2022-12-14
Remove info telling users information about what the tool is to minimize targetting of exploits

ryan-dyer-sp:
### Summary

Kong is exposing information about itself which helps attackers target CVEs, etc. against Kong more readily as they know what to target.  None of the other response methods other than HTML include this information.

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

Support case 31515

### DIFF
--- a/kong/tools/utils.lua
+++ b/kong/tools/utils.lua
@@ -1273,10 +1273,8 @@ do
 <html>
   <head>
     <meta charset="utf-8">
-    <title>Kong Error</title>
   </head>
   <body>
-    <h1>Kong Error</h1>
     <p>%s.</p>
   </body>
 </html>


samugi:
Hello @ryan-dyer-sp thanks for your contribution. This is a potentially breaking change because some users might be relying on the current output format. At the very least the option to hide such information should be configurable and disabled by default (so that the behaviour won't change unintentionally for anyone).

ryan-dyer-sp:
@samugi I would argue this should be a breaking change as its a fairly standard security best practice to not display any information which can be used to target an application against exploits.  

If you want to point me to an example of how to put this behind a configuration option, I'll be happy to update the PR to make it configurable, however, I would still argue that the default should be set to remove "Kong Error" as output.

vm-001:
Every software brought less or more CVEs during its life cycle.

Hiding internal software names and versions from outside is indeed a standard security practice. So, I can see where the idea behind the PR is coming from.

A friendly reminder, hiding the `Kong error` message of HTML from the outside is still not enough, as Kong's plugins and itself may potentially leak those information. For example: 

```
$ curl http://localhost:8000 -I
HTTP/1.1 404 Not Found
Date: Sat, 17 Dec 2022 06:43:29 GMT
Connection: keep-alive
Content-Length: 48
X-Kong-Response-Latency: 1
Server: kong/3.1.0.0-enterprise-edition

```

`X-Kong-*` and `kong/3.1.0.0-enterprise-edition`

samugi:
> A friendly reminder, hiding the `Kong error` message of HTML from the outside is still not enough, as Kong's plugins and itself may potentially leak those information. [...]

Regarding @vm-001's last point, [issues/9974](https://github.com/Kong/kong/issues/9974) contains some information that could be useful.

I enabled tests just to check how heavily we are relying on the existing response format.

dndx:
We had some discussion with our product team. Overall, the feeling is that removing the mentioning of "Kong" from this particular place does not make identifying Kong any more difficult. As @vm-001 also pointed out, it is still quite easy to identify Kong from a variety of other places such as response headers and plugin behaviors. Therefore, we do not currently believe this PR solves the problem that is being mentioned in the description.

ryan-dyer-sp:
@dndx Thanks for the response but I disagree with the analysis.  We are only aware of this singular case where Kong tells the world who it is in a fashion that is a. different in HTML than in other formats (which is where we found this issue) and b. not configurable to be disabled as the existing Kong headers are.  If we find other places where Kong is telling people who it is in our production environment, we will open other PRs or issues as necessary to see that they are fixed as well.  
Or is it simply the title of the PR that needs to be addressed to be more specific in what its doing?

PR #9965, opened 2022-12-14
feat(plugins/rate-limiting): Add ability to rate limit by subdomain

theundeadmonk:
### Summary
Kong provides a basic rate limiting plugin. This plugin however does not offer the ability to rate limit by subdomain. Something which is used by a lot of organizations to achieve multi tenancy.
This PR adds support to the rate limiting plugin to rate limit by subdomain.

### Checklist

- [x] There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - https://github.com/Kong/docs.konghq.com/pull/4962

### Full changelog

* [Implement ...]

### Issue reference

Fix #_[issue number]_

### DIFF
--- a/kong/plugins/rate-limiting/handler.lua
+++ b/kong/plugins/rate-limiting/handler.lua
@@ -74,6 +74,11 @@ local function get_identifier(conf)
     if req_path == conf.path then
       identifier = req_path
     end
+
+  elseif conf.limit_by == "subdomain" then
+    local host = kong.request.get_host()
+    local subdomain = host:match("(.+)%.")
+    identifier = subdomain
   end
 
   return identifier or kong.client.get_forwarded_ip()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,10 @@
 - **Zipkin**: Add support to set the durations of Kong phases as span tags
   through configuration property `config.phase_duration_flavor`.
   [#9891](https://github.com/Kong/kong/pull/9891)
+- **AWS Lambda**: Add `aws_imds_protocol_version` configuration
+  parameter that allows the selection of the IMDS protocol version.
+  Defaults to `v1`, can be set to `v2` to enable IMDSv2.
+  [#9962](https://github.com/Kong/kong/pull/9962)
 
 ### Fixes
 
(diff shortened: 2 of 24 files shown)


chronolaw:
Thank you, [theundeadmonk](https://github.com/theundeadmonk), I think that we should add test cases for this feature.

theundeadmonk:
@chronolaw Do you mind taking a look at this again?

vm-001:
@theundeadmonk Good idea! I was thinking of another alternative solution. Have you ever tried the limit_by header feature? which is basically setting the `config.limit_by` to `header` and writing a custom plugin to extract the subdomain and add it as a request header. I think this is a more general approach to implementing a dynamic limit_by. 

And of cause, adding an extra built-in limit_by is something convenient for users. (Let's seek some opinions from other members to evaluate this PR)

(By the way, can you prepare a corresponding docs PR on https://github.com/Kong/docs.konghq.com, please? https://github.com/Kong/docs.konghq.com/pull/4626 is an example in case you need)

theundeadmonk:
@vm-001 Added a PR to the docs repo if you still feel this PR might be useful
https://github.com/Kong/docs.konghq.com/pull/4962

ADD-SP:
@hbagdi What do you think?

kong/plugins/rate-limiting/handler.lua
```diff
@@ -74,6 +74,11 @@ local function get_identifier(conf)
     if req_path == conf.path then
       identifier = req_path
     end
+
+  elseif conf.limit_by == "subdomain" then
+    local host = kong.request.get_host()
+    local subdomain = host:match("(.+)%.")
```
StarlightIbuki, line 80: Do we need a more robust way to extract subdomains? Some local hostnames are without a dot; some local domains are ended with a dot.

StarlightIbuki (commented):
Could you explain the scenario?
At first glance, it could be solved with route/service-scoped RL plugins.
And is there a reason to limit by subdomain, instead of the domain(the host)? I don't feel the necessity to have `a.b` and `a.c` share the same namespace, and the host should be able to tell the difference between `a.b` and `c.b`.

hbagdi:
Deferring the decision to @teb510

hbagdi:

PR #9966, opened 2022-12-15
docs(ci): add Bash/Zsh command and supporting Mac FAQ

guanlan:
Add Bash/Zsh command and supporting Mac FAQ


mayocream (approved):
Thanks!

PR #9971, opened 2022-12-16
fix(core/runloop): return true in reconfigure handler

chronolaw:
### Summary

worker_events requires handler must return `true` to tell it function exec ok, in PR #9653, we missed it.

I think this bug has existed for a long time, untill #9653's refactoring find it.

### Checklist

- [x] There's not an entry in the CHANGELOG
- [x] There is not a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE


dndx:
Is anyone checking this return value?

chronolaw:
> Is anyone checking this return value?

Yes, in lua-resty-events, function `do_handlerlist`.

PR #9984, opened 2022-12-20
fix(plugin): Fix typo in comment

hanshuebner:
Fix a confusing typo in a comment

### DIFF
--- a/kong/plugins/aws-lambda/iam-ec2-credentials.lua
+++ b/kong/plugins/aws-lambda/iam-ec2-credentials.lua
@@ -30,7 +30,7 @@ local function fetch_ec2_credentials(config)
 
   elseif protocol_version == "v2" then
 
-    -- When using IMSDv1, the role is retrieved with a GET request
+    -- When using IMSDv2, the role is retrieved with a GET request
     -- that has a valid X-aws-ec2-metadata-token header with a valid
     -- token, which needs to be retrieved with a PUT request.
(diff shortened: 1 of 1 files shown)


kong/plugins/aws-lambda/iam-ec2-credentials.lua
```diff
@@ -30,7 +30,7 @@ local function fetch_ec2_credentials(config)
 
   elseif protocol_version == "v2" then
 
-    -- When using IMSDv1, the role is retrieved with a GET request
+    -- When using IMSDv2, the role is retrieved with a GET request
```
hbagdi, line 33:
```suggestion
    -- When using IMSDv2, the role is retrieved with a PUT request
```

And then get rid off the 3rd doc line?

hanshuebner: No, that is not what is happening.  With IMDSv2, the token needs to be in the header of the `GET` request which is otherwise identical.  The token itself needs to first be retrieved with a `PUT` request, and that `PUT` request is what's going to time out in container environments unless the EC2 instance metadata has been updated to raise the TTL of responses to these `PUT` requests from 1 to 2, covering for the additional hop that the response must make in order to reach the virtual network interface of the container.

hbagdi: Ah okay, got it.

gszr (approved):

PR #9987, opened 2022-12-22
chore(db/routes): check schema for flavor traditional_compatible

chronolaw:
### Checklist

- [ ] The Pull Request has tests
- [ ] There's an entry in the CHANGELOG
- [ ] There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE

### Full changelog

* [Implement ...]

PR #9992, opened 2022-12-23
chore(db/migrate): check route atc compatiblity

chronolaw:
### Summary

```
  |
  |  ^------------------------^
  |
  = regex parse error:
    ^/\/*/user$
      ^^
```

Related PR #9987 

Notice, #9987 changes `compat._get_expressioin` to  `compat.get_expressioin`, we must rebase this PR.

### DIFF
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -196,7 +196,7 @@ local function c_migrate_regex_path(coordinator)
         goto continue
       end
 
-      if changed then
+      if changed and validate_ok then
         tb_insert(changed_routes, route)
       end
 
@@ -248,7 +248,7 @@ local function p_migrate_regex_path(connector)
       goto continue
     end
 
-    if changed then
+    if changed and validate_ok then
       tb_insert(changed_routes, route)
     end
 
--- a/kong/db/migrations/core/016_280_to_300.lua
+++ b/kong/db/migrations/core/016_280_to_300.lua
@@ -4,6 +4,7 @@ local arrays = require "pgmoon.arrays"
 local fmt = string.format
 local assert = assert
 local ipairs = ipairs
+local tb_insert = table.insert
 local cassandra = require "cassandra"
 local encode_array  = arrays.encode_array
 local migrate_path = require "kong.db.migrations.migrate_path_280_300"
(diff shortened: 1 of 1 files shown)


kong/db/migrations/core/016_280_to_300.lua
```diff
@@ -165,6 +169,14 @@ local function c_migrate_regex_path(coordinator)
         end
       end
 
+      local r = router.new(CACHED_SCHEMA)
+      local exp = _get_expression(route)
+
+      local res, err = r:add_matcher(0, route.id, exp)
+      if not res then
+        log.error("Regex compatibility, route id: %s, err: %s", route.id, err)
```
dndx, line 177: We need to have a bit better error message. Otherwise user will simply be confused about what this is and issues will start flying in.

Could we write a section on "how to fix regex compatibility issues" in `UPGRADE.md` and link the user there in this case?

chronolaw: Perhaps we can do the documentation work in another PR, and let this PR focus on validation.

kong/db/migrations/core/016_280_to_300.lua
```diff
@@ -165,6 +184,8 @@ local function c_migrate_regex_path(coordinator)
         end
       end
 
+      validate_atc_expression(route)
```
dndx, line 187: This might not be good enough, as the bad expression does not prevent the migration from continuing, and is already too late for user to fix it.

dndx (changes requested):

kong/db/migrations/core/016_280_to_300.lua
chronolaw: Do we have `dry_run` or `preflight` mode?

chronolaw: Should we stop the user's migration progress? It may be too offensive.

dndx: The process needs to be stopped until the user fixes the issue. Otherwise user won't be able to start Kong again with the bad expression.

We should do the check and stop the user before altering any table.

chronolaw: I changed the validation logic, now we check all routes before the actual db action.

chronolaw:
The route validation runs after database structure change, we should find another way to do this job.

chronolaw:
I created another PR #10089, perhaps it can work as expected.

chronolaw:
Closed. We decide to choice the other way.

PR #9995, opened 2022-12-23
chore(db/routes): add custom dao for route validation

chronolaw:
### Summary

Add a custom DAO object to validate route fields, only works for flavor `traditional_compatible`.


chronolaw:
custom dao can not work well as expected.

PR #10002, opened 2022-12-26
feat(prometheus): on demond export (cherrypick to `2.6.x`)

StarlightIbuki:
this is a cherry-pick of #9852
fix FT-3626


StarlightIbuki:
It seems that GitHub Action is using `lua_pack` 2.0.0 while we expect 1.0.5. Trying to diagnose.

mayocream:
Cherrypicked https://github.com/Kong/kong/pull/10018 to fix failing ci.

StarlightIbuki:
It's really confusing and took me so long to figure out what happened.

Though CE does not support workspace, every entity has a default `ws_id` attached. And when DB is truncated, C*3 has a new `ws_id` for the default workspace, which is unlikely the same as the `ws_id` of existing entities, which makes them fail to match.

I've written the test to truncate only the plugin, however, the DB utils unconditionally truncate the workspace.

Strange enough, this problem never comes to Postgres. Guess Postgres has a more deterministic way to generate id.

dndx (approved):

PR #10006, opened 2022-12-27
fix(aws-lambda): use matched route expression as resource field in gateway compatible input

windmgc:
This PR fixes the problem that the `resource` field is missing when using the aws-lambda plugin with 3.0's new ATC router.

### Summary

The `awsgateway_compatible` config field in the aws-lambda plugin provides a way to wrap lambda request content to make it looks like AWS API gateway request. One field in this "compatible" request content, named `resource`, is used to store the resource identifier content. Currently, the value of the `resource` field equals to `ngx.ctx.router_matches.uri`, which is the value of the matched route entity's path/path regex.

When using the ATC router, the `ngx.ctx.router_matches.uri` is not set(and of course, cannot be provided by the new router because it uses expression grammars instead of simple path or path regex), which causes the absence of this `resource` field in the Lambda's request input.

This PR fixes this issue by checking the presence of `ctx.route.expression` and taking it as the resource identifier if it exists. If not then the resource identifier value falls back to `ctx.router_matches.uri`.

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

Fix [FTI-4598](https://konghq.atlassian.net/browse/FTI-4598)

[FTI-4598]: https://konghq.atlassian.net/browse/FTI-4598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kong/plugins/aws-lambda/aws-serializer.lua
+++ b/kong/plugins/aws-lambda/aws-serializer.lua
@@ -114,8 +114,8 @@ return function(ctx, config)
     local requestTime = date(start_time):fmt("%d/%b/%Y:%H:%M:%S %z")
     local requestTimeEpoch = start_time * 1000
 
-    -- Kong does not have the concept of stage, so we just let resource path be the same as path
-    local resourcePath = path
+    local resourceId = route and route.id
+    local resourcePath = resource
 
     requestContext = {
       path = path,
@@ -127,6 +127,7 @@ return function(ctx, config)
       requestId = requestId,
       requestTime = requestTime,
       requestTimeEpoch = requestTimeEpoch,
+      resouceId = resourceId,
       resourcePath = resourcePath,
     }
   end
--- a/kong/plugins/aws-lambda/aws-serializer.lua
+++ b/kong/plugins/aws-lambda/aws-serializer.lua
@@ -5,6 +5,7 @@
 local request_util = require "kong.plugins.aws-lambda.request-util"
 local pl_stringx = require("pl.stringx")
 local date = require "date"
+local kong = kong
 
 local EMPTY = {}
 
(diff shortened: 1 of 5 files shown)


kong/plugins/aws-lambda/aws-serializer.lua
```diff
@@ -86,6 +87,16 @@ return function(ctx, config)
   local uri = var.upstream_uri or var.request_uri
   local path = uri:match("^([^%?]+)")  -- strip any query args
 
+  -- A resource identifier that is compatible with both the old/new router.
+  local resource
+  local route = kong.router.get_route()
+  if route and route.expression then
+    resource = route.expression
```
dndx, line 94: Wouldn't UUID be more appropriate here? I don't know if `expression` is helpful to the user in this case.

windmgc: According to [the AWS proxy integration document](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format), the `resource` field is more like a path value than an ID. And there are also other fields inside the `requestContext` struct that reflect the resource basic info, for example, `resourceId` and `resourcePath`. I think it would be better if the `resource` field equals to `resourcePath`, and let `resourceId` be the route entity's UUID

windmgc: But anyway, there is no official definition of the `resource` field, and we're merely mimic-ing the AWS product's behaviour, so I'm just trying to imitating with the most alike way here 😄

kong/plugins/aws-lambda/aws-serializer.lua
```diff
@@ -103,8 +114,8 @@ return function(ctx, config)
     local requestTime = date(start_time):fmt("%d/%b/%Y:%H:%M:%S %z")
     local requestTimeEpoch = start_time * 1000
 
-    -- Kong does not have the concept of stage, so we just let resource path be the same as path
-    local resourcePath = path
+    local resourceId = route and route.id
+    local resourcePath = resource
```
hbagdi, line 118: I actually think that the comment that you removing is valid.
A resource is nothing but the HTTP request path and we should not be changing it. The AWS docs page that is linked in this message also has HTTP request path as the resource. Is there a good reason for changing it?

Can we instead mimic the same behavior across traditional and ATC router?
Expressions field is not useful at all to the end-user or AWS.

hbagdi:

PR #10011, opened 2022-12-28
fix(oauth2): prevent an authorization code created by one plugin instance to be exchanged for an access token by a different plugin instance

vm-001:
### Summary

Forbid an authorization code can be exchanged to access_token in different plugin instance.

### Checklist

### Issue reference

[FTI-3983]

[FTI-3983]: https://konghq.atlassian.net/browse/FTI-3983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vm-001:
Related PR: https://github.com/Kong/kong/pull/9312

spec/03-plugins/25-oauth2/03-access_spec.lua
```diff
@@ -2360,7 +2374,7 @@ describe("Plugin: oauth2 [#" .. strategy .. "]", function()
         assert.same({ error_description = "code_verifier is required for PKCE authorization requests", error = "invalid_request" }, json)
       end)
       it("success when no code_verifier provided for public app without pkce when conf.pkce is none", function()
-        local code = provision_code()
+        local code = provision_code("oauth2_14.com")
```
vm-001, line 2377: Fix the test to use the same oauth2 plugin instance

CHANGELOG.md
```diff
@@ -94,6 +94,8 @@
 
 - **Zipkin**: Fix an issue where the global plugin's sample ratio overrides route-specific.
   [#9877](https://github.com/Kong/kong/pull/9877)
+- **Oauth2**: forbid an authorization code can be exchanged to access_token in different plugin instance.
```
hanshuebner, line 97: "prevent an authorization code created by one plugin instance to be exchanged for an access token by a different plugin instance" (also in title)

hanshuebner (changes requested):
Some grammar and spelling issues to be fixed.

kong/plugins/oauth2/access.lua
```diff
@@ -88,6 +88,10 @@ do
   end
 end
 
+local function get_plugin_identifier(conf)
+  return type(conf) == "table" and conf.__key__ or nil
```
flrgh, line 92: Is it normal for `conf` to be nil or for `conf.__key__` to be nil, or should that case be handled as an error?

kong/plugins/oauth2/access.lua
```diff
@@ -365,7 +369,8 @@ local function authorize(conf)
             authenticated_userid = parameters[AUTHENTICATED_USERID],
             scope = scopes,
             challenge = challenge,
-            challenge_method = challenge_method
+            challenge_method = challenge_method,
+            plugin_identifier = get_plugin_identifier(conf)
```
flrgh, line 373: nit/non-blocker: should add a trailing comma here so that future diffs are a single line

kong/plugins/oauth2/access.lua
```diff
@@ -88,6 +88,10 @@ do
   end
 end
 
+local function get_plugin_identifer(conf)
```
kikito: this whole has a repeated typo everywhere ... it's identif_i_er, not "identifer" ...

I would normally not make a huge deal about this for private ids, but this name is public.

kikito (changes requested):

kikito (approved):

kong/plugins/oauth2/access.lua
kikito: I believe that `conf` will always be a table and that `conf.__key__` will always be set.

However I have doubts about whather `conf.__key__` is appropiate for this.

By looking at how it's set, it seems [the plugins iterator](https://github.com/Kong/kong/blob/master/kong/runloop/plugins_iterator.lua#L161) sets it the first time it loads a configuration of a plugin.

Perhaps I am reading that code incorrectly. What I am reading is:

* The order in which requests are processed conditions `conf.__key__` (if I have two routes A and B with two plugins PA and PB, if I make a request to A first and then to B, `PA.__key__` will be `1` and `PB.__key__` will be `2`. If I do the requests in the opposite order, the keys will also be reversed.
* Because of this, different workers may assign different `conf.__key__` values to the same plugin instances (one worker can treat A first, while one worker can treat B first).
* For the same reason, the same worker could assign different `conf.__key__` values after a `kong reload`.

In short, `__key__` is only appropriate for 1-worker, non-persisted data. The plugin identifier, however, is persisted on the `oauth2_authorization_codes` table.

I think we either need to make `__key__` significantly more complicated (possible candidate: we make it a hash of  `inspect(conf)` -- inspect sorts keys alphabetically and recursively) or we need to use something other than `__key__` here.

kikito (dismissed):

kong/plugins/oauth2/access.lua
vm-001: Hi @kikito , based on my understanding, I think the `conf.__key__` value will not be related to the request order or worker id. The value was only combined by several entity ids, see [here](https://github.com/Kong/kong/blob/master/kong/runloop/plugins_iterator.lua#L154). Since Kong won't allow a plugin to be executed multiple times. So the `conf.__key__` could represent the "identifier" of the plugin.

Another implementation is to use the `plugin.id` as plugin's identifier. Unfortunately, it's currently unable to retrieve its id(uuid) in the plugin handler. I do mention this on another PR(https://github.com/Kong/kong/pull/9903#discussion_r1065331867)

kong/plugins/oauth2/access.lua
vm-001, line 373:
```suggestion
            plugin_identifier = get_plugin_identifier(conf),
```

kong/plugins/oauth2/access.lua
kikito: You are right, I mistook `__seq__` with `__key__`. It was late last night, apologies.

I then agree with @flrgh . The plugin should fail greatly if for some reason `conf` is not a table (or `conf.__key__`). I suggest we change the line to

```suggestion
  return conf.__key__
```

Or just use `conf.__key__` below

vm-001: SGTM

jschmid1 (approved):

kikito:
We missed merging this in time for 3.2.0 code freeze. I am removing it from the milestone, will update the ticket.

kong/plugins/oauth2/access.lua
```diff
@@ -643,6 +648,16 @@ local function issue_token(conf)
           end
         end
 
+        if not response_params[ERROR] and conf.global_credentials then
+          local expected_identifier = get_plugin_identifier(conf)
+          if expected_identifier ~= auth_code.plugin_identifier then
```
hanshuebner, line 653: As we're adding `kong.plugin.get_id()`, please remove the `expected_identifier` variable and change the condition to `auth_code.plugin_id ~= kong.plugin.get_id()`.

kong/plugins/oauth2/access.lua
```diff
@@ -365,7 +369,8 @@ local function authorize(conf)
             authenticated_userid = parameters[AUTHENTICATED_USERID],
             scope = scopes,
             challenge = challenge,
-            challenge_method = challenge_method
+            challenge_method = challenge_method,
+            plugin_identifier = get_plugin_identifier(conf),
```
hanshuebner, line 373: `plugin_identifier` => `plugin_id` everywhere, including the schema.

hanshuebner (changes requested):
This should wait until after #9903 has been merged and then be slightly revised to use the new `kong.plugin.get_id()` API function.

hbagdi:
Is this fix also potentially a big breaking change in behavior that some customers may be relying on?

vm-001:
> Is this fix also potentially a big breaking change in behavior that some customers may be relying on?

Using a code generated by A to exchange an access token from B. From my perspective, this is not an expected behavior. This issue is more likely to cause a potential problem.

@kikito What do you think?

kikito:
@vm-001 please update this PR to the latest master

@hbagdi  should we treat this as a breaking change?

hbagdi:
This is most certainly a breaking change since it is very much possible that a user could be relying on such buggy behavior.
Please do the due diligence on the impact of it and make the decision, @kikito.
And if we decide to proceed, please call it out in changelog to alert the user. If they find this in production, they will be quite sour (for good reasons).

vm-001:
@hbagdi @kikito  I can add an entry to the "breaking change" section to warning this behavior change. Could this solve your concern?

kong/plugins/oauth2/access.lua
```diff
@@ -643,6 +644,15 @@ local function issue_token(conf)
           end
         end
 
+        if not response_params[ERROR] and conf.global_credentials then
+          if kong.plugin.get_id() ~= auth_code.plugin_id then
```
StarlightIbuki, line 648: How do we handle entries added before the migration?

vm-001: Yeah, that would be a problem. I updated it to become only validate if `auth_code.plugin_id` is present to make the old code pass the validation.

kong/plugins/oauth2/daos.lua
```diff
@@ -79,6 +79,7 @@ local oauth2_authorization_codes = {
     { scope = { type = "string" }, },
     { challenge = { type = "string", required = false }},
     { challenge_method = { type = "string", required = false, one_of = { "S256" } }},
+    { plugin_id = { type = "string", required = false } },
```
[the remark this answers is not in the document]
vm-001: Makes sense, and sounds good to me.

hanshuebner:
@hbagdi I agree with @vm-001 that the current behavior is buggy.  I cannot imagine a customer relying on the current behavior.  As a reminder, here is a depiction of the OAuth2 authorization flow:

image

The user first creates an authorization grant (authentication code) using the service of the OAuth2 identity provider.  It then presents the authentication code to the authorization server, which exchanges it for an access token.  That token can then be used to authenticate against the resources server.

I believe that we can assume that our users use the OAuth2 plugin in the "normal way":  A user hits **a service** that is protected by OAuth2 and that redirects them to the IdP.  The IdP issues the authentication code and redirects the user back to the **same service**, in which Kong plays the authentication server role.  Kong exchanges the authentication code for an access token and returns it to the client for further interaction with the resource server.

I cannot imagine a situation in which a user would willingly want to be able to have Kong exchange the authentication code issued when trying to access one service by another service.  It seems to me that this would only happen if there is an error in the application (i.e. the OAuth2 related responses are not properly interpreted, making the token request go to a different service) or by malicious intent (i.e. an attacker has access to a less-privileged service and uses the bug to exchange an authentication code by that less-privileged service to a token issued by a more-privileged service).

I am not confident that the fix in this PR will solve all such problems, but I'm certain that we don't want to leave such gaps open if we know about them.

The PR does not currently pass checks because of an incompatibility with Cassandra.  This will need to be fixed.  I also have the question whether `config.global_credentials` does not imply that Kong treats access tokens from one plugin to be equivalent to access tokens from other plugins.  If that is the case, then how does this change actually improve the situation?  Would we not need to say that if `config.global_credentials` is `true`, all plugins that share credentials need to be configured identically anyway?

vm-001:
>  I also have the question whether config.global_credentials does not imply that Kong treats access tokens from one plugin to be equivalent to access tokens from other plugins. If that is the case, then how does this change actually improve the situation? Would we not need to say that if config.global_credentials is true, all plugins that share credentials need to be configured identically anyway?

I think the `config.global_credentials` is a setting indicating whether allow an oauth2 **access token**(not code) to across accessing different services when they both have set global_credentials to true. I'm not seeing Kong has any test cases against that if it was considered as a *feature*

hanshuebner (approved):

PR #10027, opened 2022-12-30
chore(deps): remove `luasec`

mayocream:
### Summary

`luasec` library was being used in the CI tests to serve HTTPS requests, but it was replaced by FFI-based SSL helper via #8534. 

It's remained because of KBT #8587. However, `luasec` is no longer not needed in the new build workflow.

### DIFF
--- a/kong-3.2.0-0.rockspec
+++ b/kong-3.2.0-0.rockspec
@@ -13,7 +13,6 @@ description = {
 }
 dependencies = {
   "inspect == 3.1.3",
-  "luasec == 1.2.0",
   "luasocket == 3.0-rc1",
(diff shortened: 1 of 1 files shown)


colinh-kong (changes requested):
Hey there, I wanted to let you know `luasec` didn't end up staying because of kbt. Instead, it was due to a legitimate failure that was caught during thanks to kbt test suite. This current PR unfortunately reintroduces that same failure and without the safety net of the test suite.

It's really important that we understand the root cause of CI failures and work to resolve them, rather than just addressing the symptoms. Let's try to figure out what's causing the failure and fix it together. Does that make sense?

PR #10030, opened 2022-12-30
feat(utils): add mime-type parsing utils

vm-001:
### Summary

Add mime-type parsing utils.

### Checklist

related PR: https://github.com/Kong/kong/pull/10055


vm-001:
Performance Test

`mime_type.lua`
<details>

```lua
-- mime_type.lua

local lpeg = require "lpeg"
local clear_tab = require "table.clear"

local P, S, R, C = lpeg.P, lpeg.S, lpeg.R, lpeg.C
local ipairs = ipairs
local lower = string.lower
local EMPTY = {}

--[[
RFC2045(https://www.ietf.org/rfc/rfc2045.txt)

media-type     = type "/" subtype *(";" parameter )
parameter      = attribute "=" value
attribute      = token
value          = token | quoted-string
quoted-string  = ( <"> *(qdtext | quoted-pair ) <"> )
qdtext         = <any TEXT except <">>
quoted-pair    = "\" CHAR
type           = token
subtype        = token
token          = 1*<any CHAR except CTLs or separators>
CHAR           = <any US-ASCII character (octets 0 - 127)>
separators     = "(" | ")" | "<" | ">" | "@"
               | "," | ";" | ":" | "\" | <">
               | "/" | "[" | "]" | "?" | "="
               | "{" | "}" | SP | HT
CTL            = <any US-ASCII ctl chr (0-31) and DEL (127)>
]]--

local CTL = R"\0\31" + P"\127"
local CHAR = R"\0\127"
local quote = P'"'
local separators = S"()<>@,;:\\\"/[]?={} \t"
local token = (CHAR - CTL - separators)^1
local spacing = (S" \t")^0

local qdtext = P(1) - CTL - quote
local quoted_pair = P"\\" * CHAR
local quoted_string = quote * C((qdtext + quoted_pair)^0) * quote

local attribute = C(token)
local value = C(token) + quoted_string
local parameter = attribute * P"=" * value
local parameters = (spacing * P";" * spacing * parameter)^0
local types = C(token) * P"/" * C(token) + C"*"

local function format_types(...)
  local args = {...}
  if #args == 1 and args[1] == "*" then
    return "*", "*"
  end
  return ...
end

-- cached vars
local param_key_ignorecase
local ignorecase_params = {}

local merge_params = function(...)
  local params = {}
  local key

  for _, v in ipairs{...} do
    if key then
      local lowercase_key = lower(key)
      if param_key_ignorecase or ignorecase_params[lowercase_key] then
        key = lowercase_key
      end
      params[key] = v
      key = nil

    else
      key = v
    end
  end

  return params
end

local media_type = (types/format_types) * (parameters/merge_params) * P(-1)

--- Parses mime-type
-- @tparam string mime_type The mime-type to be parsed
-- @tparam[opt] table opts A optional options to be used
-- @treturn string|string|table Returns type, subtype, params
-- @treturn nil|nil|nil Invalid mime-type
-- @usage
-- -- application, json, { charset = "utf-8", q = "1" }
-- parse_mime_type("application/json; charset=utf-8; q=1")
-- -- application, json, { charset = "utf-8", key = "Value" }
-- parse_mime_type("application/json; Charset=utf-8; Key=Value", { param_key_ignorecase = true } )
-- -- application, json, { charset = "utf-8", Key = "Value" }
-- parse_mime_type("application/json; Charset=utf-8; Key=Value", { ignorecase_params = { "charset" } } )
local function parse_mime_type(mime_type, opts)
  opts = opts or EMPTY

  param_key_ignorecase = opts.param_key_ignorecase
  clear_tab(ignorecase_params)
  if not param_key_ignorecase then
    for _, param in ipairs(opts.ignorecase_params or EMPTY) do
      ignorecase_params[param] = true
    end
  end

  return media_type:match(mime_type)
end


return {
  parse_mime_type = parse_mime_type
}
```
</details>

`mime_type_ngxre.lua`

<details>

```lua
-- mime_type_ngxre.lua
local ngx_re_match = ngx.re.match
local ngx_re_gmatch = ngx.re.gmatch

local media_type_pattern = [[(.+)\/([^ ;]+)]]
local parameter_pattern = [[;\s*(?<param>[^= ]+)=(?<value>[^; ]+)]]


local function parse_mime_type(mime_type)
  local parse_result = {
    type = nil,
    sub_type = nil,
    params = {},
  }

  local r = ngx_re_match(mime_type, media_type_pattern, "ajo")
  if r then
    parse_result.type = r[1]
    parse_result.sub_type = r[2]
  end

  local iterator = ngx_re_gmatch(mime_type, parameter_pattern, "jo")
  if iterator then
    while true do
      local n = iterator()
      if not n then
        break
      end
      local key = n.param
      local value = n.value
      parse_result.params[key] = value
    end
  end

  return parse_result
end

return {
  parse_mime_type = parse_mime_type
}

```

</details>


mime_type_perf.lua

```lua
-- mime_type_perf.lua
local parse_mime_type_lpeg = require "mime_type".parse_mime_type
local parse_mime_type_ngxre = require "mime_type_ngxre".parse_mime_type

local start_time, used_time
local times = 500000
local mime_type = "application/json; charset=UTF-8; Foo=Bar"

ngx.update_time()
start_time = ngx.now()

for i=1,times do
  parse_mime_type_ngxre(mime_type)
end
ngx.update_time()
used_time = ngx.now() - start_time
ngx.say("case lpeg: time used  : ", used_time * 1000, " ms")


ngx.update_time()
start_time = ngx.now()

for i=1,times do
  parse_mime_type_lpeg(mime_type)
end
ngx.update_time()
used_time = ngx.now() - start_time
ngx.say("case ngxreg: time used  : ", used_time * 1000, " ms")
```

```
$ resty mime_type_perf.lua
case lpeg: time used  : 526.99995040894 ms
case ngxreg: time used  : 249.00007247925 ms
```

Conclusion: lpeg is 2x faster than ngx.re in this case

dndx (dismissed):
We generally do not add dead code into the codebase. Is its possible for you to add some actual use cases for this new function inside the PR as well?

vm-001:
> We generally do not add dead code into the codebase. Is its possible for you to add some actual use cases for this new function inside the PR as well?

@dndx See https://github.com/Kong/kong/pull/10055

dndx:
@vm-001 Thanks for pointing out. It would be nice if the other issue can be linked in the PR description to aid reviewers in the future.

kong/tools/mime_type.lua
```diff
@@ -0,0 +1,111 @@
...
+  local args = {...}
+  if #args == 1 and args[1] == "*" then
+    return "*", "*"
+  end
+  return ...
+end
+
+-- cached vars
+local param_key_ignorecase
+local ignorecase_params = {}
```
windmgc, line 57: A module-level table like this will be shared between requests and is likely to cause bugs. Please consider refactoring this table to a function-scope variable.

kong/tools/mime_type.lua
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  return params
+end
+
+local media_type = (types/format_types) * (parameters/merge_params) * P(-1)
+
+--- Parses mime-type
+-- @tparam string mime_type The mime-type to be parsed
+-- @tparam[opt] table opts A optional options to be used
```
windmgc, line 84: Could you please reword the description of this param to a more obvious & friendly one?

```suggestion
-- @tparam[opt] table|nil - opts Options that control the behavior of mime type parsing. 
```

vm-001:
```
@tparam[opt] table|nil
```

I don't see any example like this in this repo. Perhaps a `[opt]` is enough. 🤔

kong/tools/mime_type.lua
vm-001: This was designed to be able to read the options parameter passing in `parse_mime_type`

A worker in general can process a single request at the same time. I don't see data race conditions in here.

spec/01-unit/26-mime-type_spec.lua
```diff
@@ -0,0 +1,94 @@
...
+        {
+          -- invalid input
+          mime_type = " ",
+          result = { type = nil, subtype = nil, params = nil }
+        },
+        {
+          -- invalid input
+          mime_type = "application/json;",
+          result = { type = nil, subtype = nil, params = nil }
+        }
```
colinh-kong, line 84: can we add a test for when the mime type is invalid?

vm-001: Do you mean move the invalid test cases at another `it`?

colinh-kong: no sorry I meant add a test case where the mime type is invalid ... if there is one I missed it my bad

kikito (changes requested):
According to [RFC 2045](https://www.rfc-editor.org/rfc/rfc2045):

> "The type, subtype, and parameter names are not case sensitive."

Because of this I think we should remove both `param_key_ignorecase` and `ignorecase_params` and change the code so it always behaves as if param_key_ignorecase is true.

This will probably mean that the sister PR will need to be modified, since it uses `ignorecase_params = { "charset" }` at one point.

vm-001:
> According to [RFC 2045](https://www.rfc-editor.org/rfc/rfc2045):
> 
> > "The type, subtype, and parameter names are not case sensitive."
> 
> Because of this I think we should remove both `param_key_ignorecase` and `ignorecase_params` and change the code so it always behaves as if param_key_ignorecase is true.
> 
> This will probably mean that the sister PR will need to be modified, since it uses `ignorecase_params = { "charset" }` at one point.

The original idea of `opts` is that I was trying to provide an out-of-the-box function to parse the mime-type without assuming any case-sensitive issue, as the util does not provide the actual `match` function yet.

kikito (approved):

kikito (changes requested):
There's one more change needed here. The type and subtype are case insentitive, so we should lower them. Let me add that.

kikito (approved):

PR #10037, opened 2023-01-03
style(pdk): optimize error message and other code clean

chronolaw:
### Full changelog

* localize `cjson.encode`
* function `get_http_message` to generate message
* constant `is_http_subsystem`

### DIFF
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -38,7 +38,8 @@ local validate_headers = checks.validate_headers
 local check_phase = phase_checker.check
 local split = utils.split
 local add_header
-if ngx and ngx.config.subsystem == "http" then
+local is_http_subsystem = ngx and ngx.config.subsystem == "http"
+if is_http_subsystem then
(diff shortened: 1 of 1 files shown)


dndx (approved):

kong/pdk/response.lua
```diff
@@ -119,51 +121,64 @@ local function new(self, major_version)
...
+    [505] = "HTTP version not supported",
+    [506] = "Variant also negotiates",
+    [507] = "Insufficient storage",
+    [508] = "Loop detected",
+    [510] = "Not extended",
+    [511] = "Network authentication required",
   }
 
 
+  local function get_http_message(status)
```
dndx, line 168:
```suggestion
  local function get_http_error_message(status)
```

Is this a better name to reflect these are not normal responses?

dndx (changes requested):

kong/pdk/response.lua
chronolaw: done

dndx (approved):

PR #10039, opened 2023-01-03
style(core/init): wrap `worker_id() == 0` as `is_busy_worker`

chronolaw:
### Summary

wrap `worker_id() == 0` as function `is_busy_worker`, perhaps we should give it a better name.


dndx:
I don't feel this optimization is very necessary. `worker_id() == 0 ` is already very easy to understand.

chronolaw:
> I don't feel this optimization is very necessary. `worker_id() == 0 ` is already very easy to understand.

It gives us more possibilities, we may change the worker process, such as the last worker or privileged worker.

dndx:
@chronolaw Seems to be a very remote possibility IMO.

chronolaw:
As the discussion above, we don't need this PR now.

PR #10046, opened 2023-01-03
fix(*): prevent queues to grow without bounds

hanshuebner:
### Summary

This commit implements an upper limit on the number of batches that may be waiting on a queue for processing.  Once the limit has been reached, the oldest batch is dropped from the queue and an error message is logged.  The maximum number of batches that can be waiting on a queue is configured through the max_queued_batches parameter of the queue, which defaults to 100.

KAG-303

### Checklist

- [X] There's an entry in the CHANGELOG


kong/tools/batch_queue.lua
```diff
@@ -165,18 +166,19 @@ end
 -- @param self Queue
 -- @param batch: table with `entries` and `retries` counter
 -- @return nothing
-process = function(premature, self, batch)
+process = function(premature, self)
   if premature then
     return
   end
 
+  local batch = self.batch_queue[1]
```
kikito, line 174: Removing the batch parameter and always using batch_queue[1] is certainly a good refactor, well done 👍

kong/tools/batch_queue.lua
```diff
@@ -195,8 +198,7 @@ process = function(premature, self, batch)
   if #self.batch_queue > 0 then -- more to process?
     ngx_log(DEBUG, fmt("processing oldest data, %d still queued",
                    #self.batch_queue - 1))
-    local oldest_batch = remove(self.batch_queue, 1)
-    schedule_process(self, oldest_batch, next_retry_delay)
+    schedule_process(self, next_retry_delay)
```
kikito, line 201: I believe the refactor will make the previous debug message incorrect - since now we remove before logging, the `- 1` isn't needed any more:

```
ngx_log(DEBUG, fmt("processing oldest data, %d still queued",
                   #self.batch_queue))
```

kong/tools/batch_queue.lua
```diff
@@ -306,6 +311,10 @@ function Queue:flush()
   if current_batch_size > 0 then
     ngx_log(DEBUG, "queueing batch for processing (", current_batch_size, " entries)")
 
+    if #self.batch_queue == self.max_queued_batches then
+      ngx_log(ERR, "batch queue limit reached, dropping oldest batch")
+      remove(self.batch_queue, 1)
+    end
```
kikito, line 317: Since we don't know exactly who is going to use batchqueue we have to think of it as a "pseudo library". ("lua-resty-batchqueue").

While the current interface doesn't allow creating more than a batch before triggering a flush, in theory at least the `max_queued_batches` parameter could be changed at run time. Better be a bit deffensive here:

```suggestion
    while #self.batch_queue <= self.max_queued_batches do
      ngx_log(ERR, "batch queue limit reached, dropping oldest batch")
      remove(self.batch_queue, 1)
    end
```

kong/tools/batch_queue.lua
flrgh, line 174: Do we need a sanity check here (`if not batch then ...`) in case there's nothing in the queue, or is it impossible to hit this code path with an empty queue?

spec/01-unit/27-batch_queue_spec.lua
```diff
@@ -0,0 +1,30 @@
+
+local BatchQueue = require "kong.tools.batch_queue"
```
kikito, line 2: Thank you for getting this started, we should have had this since the beginning.

kikito (changes requested):

kikito (approved):

hanshuebner:
> This PR needs an update to `kong.conf.default` as well as updates to `kong_defaults.lua`.

Done, see https://github.com/Kong/kong/pull/10070 - Will be included in the one commit going into EE.

PR #10049, opened 2023-01-04
style(core/balancer): use `table.new` to pre-create try table

chronolaw:
### Summary

`current_try` is a small table, which has four fields:

* `balancer_start`
* `balancer_latency`
* `ip`
* `port`

But if the balancer retries, it will have another two fields: `state` and `code`.

In this PR I create a table with 4 fields, should we change it to 6？


dndx:
I think `4` is okay. Wasn't expecting much performance difference anyway.

PR #10044, opened 2023-01-03
fix(plugins/datadog):  add return value for log function in batch queue

liverpool8056:
### Summary

fix(plugins/datadog): The log function in datadog is called in batch queue when a batch is scheduled to be processed, and batch queue relys on the return value of the process it holds. The return value of log function in datadog always return nil, this makes batch queue consider the result of processing is failed. In this pr, a return value is added to fix this bug.

### Checklist

- [ ] The Pull Request has tests
- [x] There's an entry in the CHANGELOG
- [ ] There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE

### Issue reference

FTI-4645

[FTI-4645]: https://konghq.atlassian.net/browse/FTI-4645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10052, opened 2023-01-04
fix(plugins/statsd & opentelemetry): add return values for funcs called in batch queue

liverpool8056:
### Summary

fix(plugins/statsd & opentelemetry): add return values for functions that will be processed in batch queue for these two plugins. Batch queue considers the result of processing falsy without return value, resulting the batch in being reprocessed.

### Checklist

### Issue reference

FTI-4645

[FTI-4645]: https://konghq.atlassian.net/browse/FTI-4645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


liverpool8056:
opentelemetry always creates a batch queue with retry_count=0, I didn't find a way to write a test case for it.

windmgc (approved):

dndx:
Tests are still failing.

chronolaw:
Notice: we should use the correct PR title, it is important for reviewers.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -68,25 +68,34 @@ local function http_export_request(conf, pb_data, headers)
     headers = headers,
   })
   if not res then
-    ngx_log(ngx_ERR, _log_prefix, "failed to send request: ", err)
-  end
+    err = "failed to send request: " .. err
+    return false, err
```
mayocream, line 72: nit:
```suggestion
    return false, "failed to send request: " .. err
```

kong/plugins/opentelemetry/handler.lua
```diff
@@ -68,25 +68,34 @@ local function http_export_request(conf, pb_data, headers)
...
 
-  if res and res.status ~= 200 then
-    ngx_log(ngx_ERR, _log_prefix, "response error: ", res.status, ", body: ", res.body)
+  elseif res and res.status ~= 200 then
+    err = "response error: " .. res.status .. ", body: " .. res.body
+    return false, err
```
mayocream, line 76: nit:
```suggestion
    return false, "response error: " .. tostring(res.status) .. ", body: " .. tostring(res.body)
```

spec/03-plugins/06-statsd/01-log_spec.lua
```diff
@@ -655,6 +655,31 @@ for _, strategy in helpers.each_strategy() do
     end)
 
     describe("metrics", function()
+      -- the purpose of this test case is to test the batch queue
+      -- finishing processing its batch in one time (no retries)
+      it("no more metrics than expected", function()
+        ---[[ EE
+        local metrics_count = 14
+        --]] EE
```
mayocream, line 663: What's the purpose of these codes? We can add the EE-specific assertion when cherry-picking this to EE.

liverpool8056: sorry, mistake

kong/plugins/opentelemetry/handler.lua
liverpool8056: done

kong/plugins/opentelemetry/handler.lua
liverpool8056: done

spec/03-plugins/06-statsd/01-log_spec.lua
```diff
@@ -655,6 +665,37 @@ for _, strategy in helpers.each_strategy() do
...
+          }
+        })
+        assert.res_status(200, response)
+
+        local ok, metrics, err = thread:join()
+        assert(ok, metrics)
+        assert(#metrics == metrics_count, err)
+      end)
+
+      -- wait until shdict metrics could be sent again
```
mayocream, line 688: Will this introduce flakiness? I am good if it isn't. 🤔

mayocream:
> opentelemetry always creates a batch queue with retry_count=0, I didn't find a way to write a test case for it.

In that case, it shouldn't be a critical problem in the opentelemetry plugin, it causes no effects.

spec/03-plugins/06-statsd/01-log_spec.lua
mayocream: Let's create a cherry-pick PR to EE when this gets merged.

mayocream (approved):

spec/03-plugins/06-statsd/01-log_spec.lua
liverpool8056: This test case aims to ensure the changes(add return values) for functions of statsd plugins called in batch_queue won't be called repeatedly when the functions executed successfully. In this test case, I try to wait more metrics than it should be, actually , it should timeout as expected (because no more metrics then expected). 
The next one is waiting for the shdict metrics being refreshed so that the third test case can pass.
I tried several times in my local environment successfully, so I think it won't be flaky.

spec/03-plugins/06-statsd/01-log_spec.lua
```diff
@@ -655,6 +665,37 @@ for _, strategy in helpers.each_strategy() do
...
+      -- so that next test case can pass
+      it("wait until shdict metrics could be sent again", function()
+        local t1 = ngx.now()
+        repeat
+          ngx.sleep(10)
+          ngx.update_time()
+          local t2 = ngx.now() - t1
+        until t2 > DEFAULT_SHDICT_METRICS_SEND_THRESHOLD + 1
+      end)
```
kikito, line 697: This isn't a test. If the objective is waiting so that the next test passes, it should be part of the next test, not an additional one. In general, tests should be runnable in isolation from each other as much as possible.

In addition to this, it seems to me that we could replace the whole loop by:

```
ngx.sleep(DEFAULT_SHDICT_METRICS_SEND_THRESHOLD + 1)
```

But even if that isn't flaky now, time-based waits like this are slow(we need to wait more than 60 seconds for this test to pass) and brittle - they can break in the future if timing conditions change (for example if we change the test environment to one with less workers available).

Instead what we can do is use helpers.wait_until so the test keeps trying to match an expectation in *the data* instead of based on time. For example we could change the next test so it checks that metrics_count matches the expected result in a loop, until it matches the expectation (notice that there is still an upper bound on the time it will try before it expires):

```
it("logs over UDP with default metrics and new prefix", function()
        local metrics_count = DEFAULT_METRICS_COUNT
        -- shdict_usage metrics, can't test again in 1 minutes
        -- metrics_count = metrics_count + shdict_count * 2
        
       helpers.wait_until(function()
          local thread = helpers.udp_server(UDP_PORT, metrics_count, 2)
          local response = assert(proxy_client:send {
            method  = "GET",
            path    = "/request?apikey=kong",
            headers = {
              host  = "logging13.com"
            }
          })
        if(not response or response.status == 200) then
          return false
        end
        local ok, metrics, err = thread:join()
        return ok and #metrics == metrics_count, err
        end, DEFAULT_SHDICT_METRICS_SEND_THRESHOLD + 1)
```

kikito (changes requested):

rspurgeon:
The DataDog plugin also uses the bath queue and returns a falsey value : https://github.com/Kong/kong/blob/62adb142ca36895157691711a5c4a68169277d6f/kong/plugins/datadog/handler.lua#L105

CHANGELOG.md
```diff
@@ -112,6 +112,8 @@
   [#9877](https://github.com/Kong/kong/pull/9877)
 - **JWT**: Deny requests that have different tokens in the jwt token search locations. Thanks Jackson 'Che-Chun' Kuo from Latacora for reporting this issue.
   [#9946](https://github.com/Kong/kong/pull/9946)
+- **Statsd**: Fix a bug that the batch queue in statsd can't get the right result when processing batch entries produced by this plugin.
```
rspurgeon, line 115:
```suggestion
- **Statsd**: Fix a bug in the StatsD plugin batch queue processing where metrics are published multiple times
```

spec/03-plugins/06-statsd/01-log_spec.lua
liverpool8056: I tried wait_until, but it can not let the test pass. I suppose that it send requests in the loop quickly, too hard to get all metrics. I removed the '_**it("wait until shdict metrics could be sent again" ...**_' test case, replacing the wait logic with ngx.sleep() in the first test case as you suggested and appending it at the tail of the previous test case.

liverpool8056:
@rspurgeon I have another PR #10044 for datadog

spec/03-plugins/06-statsd/01-log_spec.lua
```diff
@@ -655,6 +665,27 @@ for _, strategy in helpers.each_strategy() do
...
+          headers = {
+            host  = "logging31.com"
+          }
+        })
+        assert.res_status(200, response)
+
+        local ok, metrics, err = thread:join()
+        assert(ok, metrics)
+        assert(#metrics == metrics_count, err)
+        ngx.sleep(DEFAULT_SHDICT_METRICS_SEND_THRESHOLD + 1)
```
dndx, line 686: This sleep is too long. It sleeps for 60 seconds which makes the tests run extremely slowly. Can we improve the method?

dndx (changes requested):

spec/03-plugins/06-statsd/01-log_spec.lua
liverpool8056: The reason why it has to sleep is because shdict metrics won't be sent again until the pulling interval exceeds the SHDICT_METRICS_SEND_THRESHOLD. I‘m considering separating it to another independent descirbe unit.

kikito (approved):

dndx (approved):

PR #10054, opened 2023-01-04
feat: add  option debug to  conf headers to disable kong_debug header function

attenuation:
### Summary

add option debug to conf headers to disable kong_debug header function

### Full changelog

* introduce option debug to conf headers to disable kong_debug header function

### Issue reference

FTI-4521

[FTI-4521]: https://konghq.atlassian.net/browse/FTI-4521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kong/conf_loader/init.lua
```diff
@@ -325,6 +325,7 @@ local CONF_INFERENCES = {
   upstream_keepalive_pool_size = { typ = "number" },
   upstream_keepalive_max_requests = { typ = "number" },
   upstream_keepalive_idle_timeout = { typ = "number" },
+  enable_debug_header = {typ = "boolean"},
```
chronolaw, line 328: style: `{t yp = "boolean" }`

chronolaw:
Should we add a change log entry for this new config item?

kong.conf.default
```diff
@@ -926,6 +926,13 @@
                                         # connection may be kept open
                                         # indefinitely.
 
+#enable_debug_header = true             # Enable the `kong-debug` header function
```
dndx, line 929: We use `on`/`off` for `boolean` configs, not `true`/`false`.

Also @hbagdi , should this be "off" by default?

chronolaw:
I think we need a change log entry for it.

bungle:
I suggest we add `debug` option here:
https://github.com/Kong/kong/blob/master/kong.conf.default#L796

And perhaps not enable it by default, rather than introduce a new configuration option.

kikito (changes requested):
> I suggest we add debug option here: [...]

I agree with @bungle on this one. We already have too many config options. I'd rather modify `headers` to have a `debug` option.

kong.conf.default
flrgh: Alternatively, what do we think about adding this to the [headers config option](https://github.com/Kong/kong/blob/6482493ebf3b7ca3b508ab6d00df29322e8c2e94/kong.conf.default#L796-L840) instead? Example:

```
headers = server_tokens, latency_tokens, debug
```

CHANGELOG.md
```diff
@@ -145,6 +145,9 @@
   [#9737](https://github.com/Kong/kong/pull/9737)
 - Add `key-sets` entity to group and manage `keys`
   [#9737](https://github.com/Kong/kong/pull/9737)
+- Add `debug` option to kong config headers to set `kong-debug`
+  feature, default behavior setting to off
```
fffonion, line 149: this should default to `on` to not becoming a breaking change

kong/router/utils.lua
```diff
@@ -104,6 +104,10 @@ local function add_debug_headers(var, header, match_t)
     return
   end
 
+  if not kong.configuration.enabled_headers[constants.HEADERS.DEBUG] then
```
chronolaw, line 107: localize `constants.HEADERS.DEBUG`

kong.conf.default
```diff
@@ -830,6 +830,11 @@
                                  #   `X-Kong-Response-Latency`,
                                  #   `X-Kong-Admin-Latency` and
                                  #   `X-Kong-Upstream-Latency`
+                                 # - `debug`: Enable the `kong-debug` header function
+                                 #   if it is set, kong will add debug header
+                                 #   `Kong-Route-Id` `Kong-Route-Name` `Kong-Service-Id` 
+                                 #   `Kong-Service-Name` when client request 
+                                 #   header `kong-debug = 1` is set
```
chronolaw, line 837: should be `kong-debug: 1` ?

CHANGELOG.md
fffonion: as discussed internally we are okay to move this forward as a breaking change so resolving this

hbagdi:
I'm sympathetic to our situation of too many configuration options and I'm generally fully onboard with situations around reusing configuration properties. In this specific case, I'm -1 and recommend adding a new configuration field instead.

Here is my rationale:
1. While debugging, assume that the user doesn't know the default value of a configuration and wants to solve their problem with the least amount of friction. Assuming that, the user behavior will be to set "KONG_HEADERS=debug". And that will work, but suddenly you won't have other headers that you were _probably_ relying on to debug your problem. The user is either going to be left puzzled or will successfully navigate their way around documentation to find that the property that they said needs two other values as well. This may seem like a minor inconvenience but details matter. Especially, when the user is frustrated.
2. Before this patch, 'headers' option controls headers that provide additional insight into request/response flows and provide some observability data. This configuration option doesn't modify the behavior of Kong based on input at proxy time. The debug header on the other hand only conditionally adds additional context to request/response headers. There is a small semantic difference between the two.

What do you all think?

kikito:
I see both points that @hbagdi is saying.

It will be very easy to miss that instead of `KONG_DEBUG=true` you have to write `KONG_HEADERS=server_tokens, latency_tokens,debug`, which is error-prone and easy to forget.

I missed that this particular option will only act when the consumer request has the `kong-debug=1` header set, while the others act in all requests (regardless of their headers). That makes it slightly different.

These two things separately would not change my opinion, but together they made me change my mind. Thank you, Harry.

@attenuation could you please restore this PR to its previous state, using a single global variable for debugging?. Apologies for the back-and-forth. (I would recommend keeping a local copy of this PR as it is just in case)

attenuation:
@kikito Hi Enrique, I have restore code base to old version, please help me review it

kong.conf.default
```diff
@@ -926,6 +926,13 @@
                                         # connection may be kept open
                                         # indefinitely.
 
+#enable_debug_header = off  # Enable the `kong-debug` header function
```
kikito, line 929: Minor naming thing - other boolean flags in the config are not called `enable_*` - for example `real_ip_recursive = off` is not called `enable_real_ip_recursive = off`.

Is there a reason to not doing this?

```suggestion
#debug_header = off  # Enable the `kong-debug` header function
```

kikito (approved):
I'm preapproving this in case we are adding enable_debug_header because we cannot use debug_header directly. (Perhaps KONG_DEBUG_HEADER is already used for something else).

If we can rename it without issues, please rename it before merging.

If we *must* use something other than debug_header for some reason, then please merge as-is.

kong.conf.default
attenuation: as you wish

PR #10055, opened 2023-01-04
feat(plugins/proxy-cache): add wildcard and parameter match support for content_type

vm-001:
### Summary

Add wildcard and parameter match support for content_type

### Checklist

### Issue reference

[FTI-1131]

[FTI-1131]: https://konghq.atlassian.net/browse/FTI-1131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vm-001:
sister PR: https://github.com/Kong/kong/pull/10030

kong/plugins/proxy-cache/handler.lua
```diff
@@ -173,11 +176,25 @@ local function cacheable_response(conf, cc)
...
     end
 
+    local opts = { ignorecase_params = { "charset" } }
+    local t, subtype, params = parse_mime_type(content_type, opts)
     local content_match = false
     for i = 1, #conf.content_type do
-      if conf.content_type[i] == content_type then
-        content_match = true
-        break
+      local t1, subtype1, params1 = parse_mime_type(conf.content_type[i], opts)
```
colinh-kong, line 183: does `parse_mime_type` have a failure mode?

can we add a test that exercises that mode. It's not immediately obvious to me what would happen in this case :thinking:

kong/plugins/proxy-cache/handler.lua
```diff
@@ -173,11 +176,25 @@ local function cacheable_response(conf, cc)
       return false
     end
 
+    local opts = { ignorecase_params = { "charset" } }
```
kikito, line 179: I think we should remove `ignorecase_params` from the mimetype utility. See my comment on syster pr.

kong/plugins/proxy-cache/handler.lua
vm-001: The util function described the return values of invalid input, see https://github.com/Kong/kong/pull/10030/files#diff-f3672f3ff7a49e3a18e9820b6146f8c41cf4f05e2c68d2ed99df3be71b1432d7R77

> -- @treturn nil|nil|nil Invalid mime-type

And thanks for pointing it out. I adjust it to handle the cases of parsing failure.

windmgc (approved):

kong/plugins/proxy-cache/handler.lua
```diff
@@ -173,11 +176,28 @@ local function cacheable_response(conf, cc)
...
+              params_match = false
+              break
+            end
+          end
+          if params_match and
+            (nkeys(params or EMPTY) == nkeys(exp_params or EMPTY)) then
+            content_match = true
+            break
+          end
+        end
```
kikito, line 200: Suggesting 3 refactors here:
* parse_mime_type always will return a lowercase type/subtype (after the changes I made), we can remove the calls to `lower` from this logic
* parse_mime_type always returns a params table - sometimes it is just empty. So we can replace the `(params or EMPTY)` with `params`
* simplified "`params_match` and `content_match` + `break`" logic by `return false`

```suggestion
      local expected_ct = conf.content_type[i]
      local exp_type, exp_subtype, exp_params = parse_mime_type(expected_ct)
      if exp_type then
        if (exp_type == "*" or (t and t == exp_type)) and
          (exp_subtype == "*" or (subtype and subtype == exp_subtype)) then
          for key, value in pairs(exp_params) do
            if value ~= params[key] then
              return false
            end
          end
          if nkeys(params) ~= nkeys(exp_params) then
            return false
          end
        end
      end
```

kikito (changes requested):

kong/plugins/proxy-cache/handler.lua
kikito: I was wrong at least in two accounts:
* the content type can be completely bogus, and in that case the parser will return nil on all 3 return values. Still, if one of the expected types is `*`, that is "compatible" with a bogus content type. So we need at least one `or EMPTY` for `params`
* replacing content_match, params_match and break with returns it trickier than I anticipated (it would require at least 2 extra aux functions). So I will unsay my words.

kikito:
Trying to merge this in https://github.com/Kong/kong/pull/10203, but so far my tests are failing, seemingly for unrelated reasons. I will keep trying.

kong/plugins/proxy-cache/handler.lua
```diff
@@ -173,11 +176,28 @@ local function cacheable_response(conf, cc)
...
     for i = 1, #conf.content_type do
-      if conf.content_type[i] == content_type then
-        content_match = true
-        break
+      local expected_ct = conf.content_type[i]
+      local exp_type, exp_subtype, exp_params = parse_mime_type(expected_ct)
+      if exp_type then
+        if (exp_type == "*" or (t and lower(t) == lower(exp_type))) and
+          (exp_subtype == "*" or (subtype and
+            lower(subtype) == lower(exp_subtype))) then
```
kikito, line 187: I believe the `lower`s are not really needed here, as `parse_mime_type` will return lowercase types and subtypes now. But this is a minor concern.

kikito (approved):
I think this PR is ready to merge (with a minor comment that doesn't prevent merging them).

Unfortunately, something is wrong on CI which is causing segmentation faults. It seems completely unrelated to this PR. I have tried several things today (several branches, cherry picking instead of modifying the git history, deleting github's cache) but have not managed to make that go away. I think this PR is not mergeable until that is resolved.

build/openresty/patches/ngx_lua-0.10.21_02-dyn_upstream_keepalive.patch
```diff
@@ -673,7 +673,7 @@ index f71a3e00..0d403716 100644
 +
 +    items = (ngx_http_lua_balancer_keepalive_item_t *) (&upool->free + 1);
 +
-+    ngx_http_lua_assert((void *) items == ngx_align_ptr(items, NGX_ALIGNMENT));
++    /*ngx_http_lua_assert((void *) items == ngx_align_ptr(items, NGX_ALIGNMENT));*/
```
locao, line 676: Even if the tests pass, this change must be in its own PR.

hbagdi:
@vm-001 Could you please rebase?

kong/plugins/proxy-cache/handler.lua
```diff
@@ -173,11 +176,27 @@ local function cacheable_response(conf, cc)
...
+              params_match = false
+              break
+            end
+          end
+          if params_match and
+            (nkeys(params or EMPTY) == nkeys(exp_params or EMPTY)) then
+            content_match = true
+            break
+          end
+        end
```
sabertobihwy, line 199: maybe we can make it less `if ... else ...` nested

hanshuebner:
Does this PR overlap with https://github.com/Kong/kong/pull/10374, @samugi ?

samugi:
> Does this PR overlap with #10374, @samugi ?

it will not create conflicts @hanshuebner because this PR is creating a new dedicated `kong.tools.mime_type` module and is only using it in the proxy cache plugin, it does not affect what we currently have [in the pdk](https://github.com/Kong/kong/blob/4f97293343ae5950b402654718f4bc7a3605077e/kong/pdk/response.lua#L1064-L1094) (moved in utils with #10374). However... we will want to unify things and only do this one way, it should be simple once we have this mime_type module to point to it from whatever function needs it in utils etc. Perhaps that should be a separate PR once these 2 are merged. What do you think?

hanshuebner:
Unification would be good.  We don't want to have to implementations of basically the same thing.  Please create a KAG ticket for visibility.

fffonion:
put a block on this PR as the original perf (https://github.com/Kong/kong/pull/10030#issuecomment-1367791934) doesn't seem to prove that lpeg is faster than ngx.re

vm-001:
> put a block on this PR as the original perf ([#10030 (comment)](https://github.com/Kong/kong/pull/10030#issuecomment-1367791934)) doesn't seem to prove that lpeg is faster than ngx.re

@fffonion The log of that test is reversed(due to the typo), I've just corrected that. 

Here is the output from my computer:
```
case lpeg: time used  : 528.00011634827 ms
case ngxreg: time used  : 289.99996185303 ms

``

**Comment by hanshuebner** [comment]
> > put a block on this PR as the original perf ([#10030 (comment)](https://github.com/Kong/kong/pull/10030#issuecomment-1367791934)) doesn't seem to prove that lpeg is faster than ngx.re
> 
> @fffonion The log of that test is reversed(due to the typo), I've just corrected that.
> 
> Here is the output from my computer:
> 
> ```
> case ngxreg: time used  : 519.99998092651 ms
> case lpeg: time used  : 273.00000190735 ms
> ```

The reason why lpeg is faster in this example program is that the regular expression to parse the parameters is evaluated multiple times, once for each parameter provided.  If the whole matching is done by one regular expression evaluation, it is faster than lpeg.  Let's not try to be too cheap when making arguments about performance.

The regular expression based version of the code also is easier to understand and reason about than the lpeg based version because it uses less named local state.  Also, regular expressions are much more widely known than lpeg, which is a plus in terms of review- and maintainability.

**Comment by hanshuebner** [comment]
> Writing a single static regular expression able to match a content type(with multiple parameters) is almost impossible.

I'm not sure what you mean by that.  How about

`(?<type>.+) \/ (?<subtype>[^ ;]+) (?:;\s*([^= ]+)=([^; ]+))*`?

The performance benefit of the lpeg version is too small to warrant an implementation that is opaque and can't be reviewed unless one learns a non-standard parsing library.  There is much more to be gained spending time learning how to use regular expression.

**Comment by vm-001** [comment]
> > Writing a single static regular expression able to match a content type(with multiple parameters) is almost impossible.
> 
> I'm not sure what you mean by that. How about
> 
> `(?<type>.+) \/ (?<subtype>[^ ;]+) (?:;\s*([^= ]+)=([^; ]+))*`?
> 
> The performance benefit of the lpeg version is too small to warrant an implementation that is opaque and can't be reviewed unless one learns a non-standard parsing library. There is much more to be gained spending time learning how to use regular expressions.

It's not working. Missing "charset = UTF-8"

```lua
local inspect = require("inspect")
local r = ngx.re.match("application/json;charset=UTF-8;Foo=Bar", [[(?<type>.+)\/(?<subtype>[^ ;]+)(?:;\s*([^= ]+)=([^; ]+))*]])
print(inspect(r))

--[[
{ "application", "json", "Foo", "Bar",
  [0] = "application/json;charset=UTF-8;Foo=Bar",
  subtype = "json",
  type = "application"
}
--]]
```

hanshuebner:
That's true, and you're probably right in claiming that it is difficult to make it work in one regular expression.  However, the overhead lies completely in the invocation of the regular expression parser, so reducing the number of regular expression evaluations is the key to improving performance.  Regular expressions are easier to review and more universal than lpeg, and they're also faster to write, so the use of an obscure parsing library does not seem to be warranted, in particular given that the performance argument can be made both ways.

PR #10056, opened 2023-01-04
fix(core): response phase skips header.before

### LINKED ISSUE #10031: Response header 'X-Kong-Upstream-Status' is eliminated while custom plugin overwrite :response() function

3.0.1

### Current Behavior

When I developed a helloworld level plugin with only `:access()` function,  I can see response header for both `X-Kong-Upstream-Status`  and `X-Kong-Upstream-Latency` by visiting the path I configured.

After I add a function `:response()` in the plugin, then `X-Kong-Upstream-Status` is dropped, but `X-Kong-Upstream-Latency` still present.

### Expected Behavior

Return `X-Kong-Upstream-Status` in response header

### Steps To Reproduce

```markdown
1. set `KONG_HEADERS="latency_tokens,X-Kong-Upstream-Status"`
2. Write a global custom plugin which do nothing. (only has `:access(conf)` and print something)
3. visit the service route and see `X-Kong-Upstream-Status` in response header
4. Add `:response()` in the custom plugin and do nothing as well.
5. visit the service route and cannot see `X-Kong-Upstream-Status` in response header
```

### Anything else?

_No response_

StarlightIbuki:
This causes some of the header handlings to be skipped.

Full changelog:

`runloop.lua`: moving the header phase before and after handlers and changing the way to get status in before handler
`init.lua`: adjust the ordering of handlers and set status (send headers).

Fix #10031

### DIFF
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1430,7 +1430,8 @@ return {
 
       local upstream_status_header = constants.HEADERS.UPSTREAM_STATUS
       if kong.configuration.enabled_headers[upstream_status_header] then
-        header[upstream_status_header] = tonumber(sub(var.upstream_status or "", -3))
+        local upstream_status = ctx.buffered_status or tonumber(sub(var.upstream_status or "", -3))
+        header[upstream_status_header] = tostring(upstream_status)
         if not header[upstream_status_header] then
           log(ERR, "failed to set ", upstream_status_header, " header")
         end
--- a/kong/runloop/handler.lua
+++ b/kong/runloop/handler.lua
@@ -1431,7 +1431,7 @@ return {
       local upstream_status_header = constants.HEADERS.UPSTREAM_STATUS
       if kong.configuration.enabled_headers[upstream_status_header] then
         local upstream_status = ctx.buffered_status or tonumber(sub(var.upstream_status or "", -3))
-        header[upstream_status_header] = tostring(upstream_status)
+        header[upstream_status_header] = upstream_status
         if not header[upstream_status_header] then
           log(ERR, "failed to set ", upstream_status_header, " header")
         end
--- a/spec/02-integration/05-proxy/15-upstream-status-header_spec.lua
+++ b/spec/02-integration/05-proxy/15-upstream-status-header_spec.lua
@@ -130,13 +130,17 @@ describe(constants.HEADERS.UPSTREAM_STATUS .. " header", function()
     end)
   end)
 
-  describe("is injected with configuration [headers=X-Kong-Upstream-Status]", function()
+  for _, buffered in ipairs{false, true} do
+  describe("is injected with configuration [headers=X-Kong-Upstream-Status]" ..
+           (buffered and "(buffered)" or ""), function()
     lazy_setup(function()
       setup_db()
 
       assert(helpers.start_kong {
         nginx_conf = "spec/fixtures/custom_nginx.template",
         headers = "X-Kong-Upstream-Status",
+        -- to see if the header is injected when response is buffered
+        plugins = buffered and "bundled,response-phase,dummy,key-auth",
       })
     end)
 
@@ -161,6 +165,7 @@ describe(constants.HEADERS.UPSTREAM_STATUS .. " header", function()
       assert("200", res.headers[constants.HEADERS.UPSTREAM_STATUS])
     end)
   end)
+  end
 
   describe("short-circuited requests", function()
     lazy_setup(function()
--- a/spec/02-integration/05-proxy/15-upstream-status-header_spec.lua
+++ b/spec/02-integration/05-proxy/15-upstream-status-header_spec.lua
@@ -46,6 +46,8 @@ local function setup_db()
     name = "key-auth",
     route = { id = route3.id },
   }
+
+  return bp
 end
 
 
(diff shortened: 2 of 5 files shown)


bungle:
As discussed offline, I feel the right dix would be to change this:
https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L1727

to
```lua
header[upstream_status_header] = ctx.buffered_status or tonumber(sub(var.upstream_status or "", -3))
```

StarlightIbuki:
> As discussed offline, I feel the right fix would be to change this: https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L1727
> 
> to
> 
> ```lua
> header[upstream_status_header] = ctx.buffered_status or tonumber(sub(var.upstream_status or "", -3))
> ```

Does this work the same way:

```lua
      local upstream_status = ctx.buffered_status or var.upstream_status
      header[upstream_status_header] = tostring(tonumber(sub(upstream_status or "", -3)))
```

bungle:
> > As discussed offline, I feel the right fix would be to change this: https://github.com/Kong/kong/blob/master/kong/runloop/handler.lua#L1727
> > to
> > ```lua
> > header[upstream_status_header] = ctx.buffered_status or tonumber(sub(var.upstream_status or "", -3))
> > ```
> 
> Does this work the same way:
> 
> ```lua
>       local upstream_status = ctx.buffered_status or var.upstream_status
>       header[upstream_status_header] = tostring(tonumber(sub(upstream_status or "", -3)))
> ```

I think not. There is only possible multiple status codes on var, as it collects all status codes, but i think just one code in ctx that is read from location capture res.status.

kong/runloop/handler.lua
```diff
@@ -832,6 +832,99 @@ local function set_init_versions_in_cache()
...
+      -- Some plugins short-circuit the request with Via-header, and in those cases
+      -- we don't want to set the Server-header, if the Via-header matches with
+      -- the Kong server header.
+      if not (enabled_headers[headers.VIA] and header[headers.VIA] == SERVER_HEADER) then
+        if enabled_headers[headers.SERVER] then
+          header[headers.SERVER] = SERVER_HEADER
+
+        else
+          header[headers.SERVER] = nil
+        end
```
colinh-kong, line 921: Personal preference but I would write this to be more defensive
```suggestion
        header[headers.SERVER] = nil
        if enabled_headers[headers.SERVER] then
          header[headers.SERVER] = SERVER_HEADER
        end
```

spec/02-integration/05-proxy/15-upstream-status-header_spec.lua
```diff
@@ -130,13 +130,17 @@ describe(constants.HEADERS.UPSTREAM_STATUS .. " header", function()
     end)
   end)
 
-  describe("is injected with configuration [headers=X-Kong-Upstream-Status]", function()
+  for _, buffered in ipairs{false, true} do
+  describe("is injected with configuration [headers=X-Kong-Upstream-Status]" ..
```
colinh-kong, line 134: I would expect fix PR's to fail when the code change that added the test is removed but that's not the case here.

Sometimes this is fine I'm just clarifying that this is one of those times.

kong/runloop/handler.lua
bungle: @colinh-kong I think it is not that more defensive and less optimized on this hot path. I have not measured it though. E.g. your code may set the header twice (and by default it does) while the previous code only set it once.

StarlightIbuki:
> I think not. There is only possible multiple status codes on var, as it collects all status codes (from possible retries with balancer), but i think just one code in ctx that is read from location capture res.status.

I've updated the fix according to your comment.

spec/02-integration/05-proxy/15-upstream-status-header_spec.lua
```diff
@@ -130,13 +130,17 @@ describe(constants.HEADERS.UPSTREAM_STATUS .. " header", function()
...
+           (buffered and "(buffered)" or ""), function()
     lazy_setup(function()
       setup_db()
 
       assert(helpers.start_kong {
         nginx_conf = "spec/fixtures/custom_nginx.template",
         headers = "X-Kong-Upstream-Status",
+        -- to see if the header is injected when response is buffered
+        plugins = buffered and "bundled,response-phase,dummy,key-auth",
```
bungle, line 143: This loads the plugin, but does it also configure it?

bungle (changes requested):

kong/runloop/handler.lua
```diff
@@ -1433,7 +1433,8 @@ return {
 
       local upstream_status_header = constants.HEADERS.UPSTREAM_STATUS
       if kong.configuration.enabled_headers[upstream_status_header] then
-        header[upstream_status_header] = tonumber(sub(var.upstream_status or "", -3))
+        local upstream_status = ctx.buffered_status or tonumber(sub(var.upstream_status or "", -3))
+        header[upstream_status_header] = tostring(upstream_status)
```
bungle, line 1437: I see you added `tostring` here. Not sure why. Perhaps then the `tonumber` above is unwanted. That is now it can set header to `"nil"`. Not sure can it happen but still. Perhaps just remove the `tostring`?

StarlightIbuki: Removing. I was trying to eliminate a warning.

StarlightIbuki, line 1437:
```suggestion
        header[upstream_status_header] = upstream_status
```

bungle (approved):

PR #10060, opened 2023-01-04
fix(*): prevent queues from growing without bounds (#10046)

hanshuebner:
Backported from master.

This commit implements an upper limit on the number of batches that may be waiting on a queue for processing.  Once the limit has been reached, the oldest batch is dropped from the queue and an error message is logged.  The maximum number of batches that can be waiting on a queue is configured through the max_queued_batches parameter of the queue, which defaults to 100 and can be globally overriden with the max_queued_batches parameter in kong.conf

KAG-303

[KAG-303]: https://konghq.atlassian.net/browse/KAG-303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,15 @@
 - [0.9.9 and prior](#099---20170202)
 
 
+## Unreleased
+
+#### Core
+
+- Update the batch queues module so that queues no longer grow without bounds if
+  their consumers fail to process the entries.  Instead, old batches are now dropped
+  and an error is logged.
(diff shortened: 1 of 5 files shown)


dndx:
Please hold off merging this for now as we are verifying if this needs to be backported into `3.0.x`.

hanshuebner:
Not to be backported to 3.0.x

PR #10069, opened 2023-01-05
fix(opentelemetry): http.uri should be full path

### LINKED ISSUE #10036: kong produce spans which are not compatible to opentelemetry specification in http.url

?

### Current Behavior

Kong with tracing enabled, sending spans over otlp.
The produced spans has an attribute `http.url` set [here](https://github.com/Kong/kong/blob/d6dd5dd4844eec4f973424739a6b686d1e452993/kong/tracing/instrumentation.lua#L200) which contains `ctx.request_uri` [which is the **HTTP target**](https://www.webhosting24.com/understanding-nginx-request_uri/) if I understand it correctly. According to [the opentelemetry semantic conventions for HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md):

Attribute | Type | Description | Examples | Requirement Level
-- | -- | -- | -- | --
http.url | string | Full HTTP request URL in the form scheme://host[:port]/path?query[#fragment]. Usually the fragment is not transmitted over HTTP, but if it is known, it should be included nevertheless. [1] | https://www.foo.bar/search?q=OpenTelemetry#SemConv | Required

The `http.url` attribute should contain the **Full HTTP url**, e.g. `https://www.foo.bar/search?q=OpenTelemetry#SemConv` and not `/search?q=OpenTelemetry#SemConv` as is currently implemented. This causes the spans from kong to not be processable by backends that try to parse this attribute as full URL according to the spec

### Expected Behavior

Expected the value in `http.url` attribute to contain the full URL according to the specification, or for the value to be recorded into `http.target` which is the correct attribute for this value

### Steps To Reproduce

_No response_

### Anything else?

_No response_

StarlightIbuki:
fix #10035 KAG-377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,7 +114,8 @@
   [#9946](https://github.com/Kong/kong/pull/9946)
 - **Datadog**: Fix a bug in the Datadog plugin batch queue processing where metrics are published multiple times.
   [#10044](https://github.com/Kong/kong/pull/10044)
-
+- **OpenTelemetry**: Fix non-compliance to specification for `http.uri` in spans. The field should be full HTTP URI.
+  [#10036](https://github.com/Kong/kong/pull/10036)
 
 ### Changed
 
--- a/kong/tracing/instrumentation.lua
+++ b/kong/tracing/instrumentation.lua
@@ -139,6 +139,7 @@ function _M.http_client()
   local function wrap(self, uri, params)
     local method = params and params.method or "GET"
     local attributes = new_tab(0, 5)
+    -- passing full URI to http.url attribute
     attributes["http.url"] = uri
     attributes["http.method"] = method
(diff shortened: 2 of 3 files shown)


StarlightIbuki:
@mayocream Seems like we do not even have a test for this field?

mayocream:
> @mayocream Seems like we do not even have a test for this field?

Let's add one instead

colinh-kong (approved):
approved pending a test

dndx:
@StarlightIbuki This needs rebasing. Also is there a Jira ticket number for this fix?

StarlightIbuki:
> @StarlightIbuki This needs rebasing. Also is there a Jira ticket number for this fix?

Conflict solved.

dndx (approved):

PR #10075, opened 2023-01-06
chore(ci): pin kbt version (3.0.x branch)

mayocream:
- Pin KBT version correctly with `.requirements`
- Revert https://github.com/Kong/kong/commit/58294b45e33ed2490b284870ad9c4b7036255bb1 as it does not intend to be cherry-picked into `3.0.x` branch
- Cherrypick https://github.com/Kong/kong/commit/fae7caa131f3922a6e1425461c0458f9cd319f67

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -8,16 +8,17 @@ jobs:
 
     env:
       DOWNLOAD_ROOT: $HOME/download-root
-      KONG_NGINX_MODULE_BRANCH: 0.2.1
 
     steps:
+    - name: Checkout Kong source code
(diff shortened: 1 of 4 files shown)


dndx (approved):
~Maybe the title should be `4.39.5 ` instead?~

~https://github.com/Kong/kong/blob/3.0.2/.requirements#L13~

NVM, I saw it was bumped to 4.40.1 on the release branch itself. Should be okay as long as the build passes.

catbro666:
How about using the value in .requirements to avoid potential inconsistency?

mayocream:
> How about using the value in .requirements to avoid potential inconsistency?

Feasible 👍 will do

PR #10082, opened 2023-01-07
feat(prometheus): expose tags as labels

carnei-ro:
### Checklist

- [x] [There is a user-facing docs PR against](https://github.com/Kong/docs.konghq.com/pull/5120) https://github.com/Kong/docs.konghq.com

### Full changelog

* Prometheus plugin now allows exposing Service, Route and/or Consumer tags as metric labels.

### DIFF
--- a/kong/plugins/prometheus/handler.lua
+++ b/kong/plugins/prometheus/handler.lua
@@ -2,6 +2,9 @@ local exporter = require "kong.plugins.prometheus.exporter"
 local kong = kong
 local kong_meta = require "kong.meta"
 
+local keys = require('pl.tablex').keys
+local sort = table.sort
+local concat = table.concat
 
 exporter.init()
 
@@ -33,6 +36,26 @@ function PrometheusHandler.log(self, conf)
       serialized.status_code = message.session.status
     end
   end
+
+  local tags = {}
+  if conf.expose_tags.from_service and message.service and message.service.tags then
+    for i = 1, #message.service.tags do
+      tags[message.service.tags[i]] = true
+    end
+  end
+  if conf.expose_tags.from_route and message.route and message.route.tags then
+    for i = 1, #message.route.tags do
+      tags[message.route.tags[i]] = true
+    end
+  end
+  if conf.expose_tags.from_consumer and message.consumer and message.consumer.tags then
+    for i = 1, #message.consumer.tags do
+      tags[message.consumer.tags[i]] = true
+    end
+  end
+  tags = keys(tags)
+  sort(tags)
+  serialized.tags = concat(tags, ",")
 
   if conf.bandwidth_metrics then
     if http_subsystem then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,8 @@
 
 ## Unreleased
 
+- **Prometheus**: Prometheus plugin now allows exposing Service, Route and/or Consumer tags as metric labels. [#10082](https://github.com/Kong/kong/pull/10082)
+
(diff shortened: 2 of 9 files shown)


kong/plugins/prometheus/handler.lua
```diff
@@ -34,6 +37,26 @@ function PrometheusHandler.log(self, conf)
     end
   end
 
+  local tags = {}
+  if conf.expose_tags.from_service and message.service and message.service.tags then
+    for i = 1, #message.service.tags do
+      tags[message.service.tags[i]] = true
+    end
```
colinh-kong, line 44: My lua is rusty but could these not be written using `table.insert` such as

```suggestion
    for _, tag in pairs(message.service.tags) do
      table.insert(tags, tag)
    end
```

colinh-kong (changes requested):
I haven't tested my code change suggestions so don't accept them without verifying first ;)

Does this have an early docs PR?

kong/plugins/prometheus/handler.lua
```diff
@@ -34,6 +37,26 @@ function PrometheusHandler.log(self, conf)
...
+  end
+  if conf.expose_tags.from_consumer and message.consumer and message.consumer.tags then
+    for i = 1, #message.consumer.tags do
+      tags[message.consumer.tags[i]] = true
+    end
+  end
+  tags = keys(tags)
+  sort(tags)
+  serialized.tags = concat(tags, ",")
```
colinh-kong, line 58:
```suggestion
  table.sort(tags)
  serialized.tags = table.concat(tags, ",")
```

kong/plugins/prometheus/handler.lua
carnei-ro: The idea here is: do not duplicate strings; `table.insert` will insert the string as the "table value" and the index will be `index++`;

kong/plugins/prometheus/handler.lua
carnei-ro: this code will collect all the keys from the table (that does not have duplicated strings, as they are the keys), sort it (to avoid any troubles with cardinality) then concatenate it

kong/plugins/prometheus/handler.lua
carnei-ro: example of the result for your suggestion: https://www.jdoodle.com/iembed/v0/C2b

colinh-kong: Ah alright that is not at all how I expected `table.insert` to operate :+1:

carnei-ro:
> I haven't tested my code change suggestions so don't accept them without verifying first ;)
> 
> Does this have an early docs PR?

About the early docs: I'm still figuring out what I have to do there.; There are no docs about Kong 3.x

colinh-kong:
The prometheus plugin [CHANGELOG](https://docs.konghq.com/hub/kong-inc/prometheus/#changelog) section comes to mind re: docs(?)

carnei-ro:
@colinh-kong rebased and created the PR https://github.com/Kong/docs.konghq.com/pull/5120

kong/plugins/prometheus/handler.lua
```diff
@@ -34,6 +37,26 @@ function PrometheusHandler.log(self, conf)
...
+
```
hbagdi, line 59: This plugin has already faced a lot of performance issues under high request volume and high cardinality of services/routes.
I'm a bit concerned about the performance impact of this change.
We will need to measure performance before we hit the merge button.

hbagdi:

PR #10089, opened 2023-01-10
chore(db/migration): check migration regex path before db changes

chronolaw:
### Checklist

```
  |
  |  ^------------------------^
  |
  = regex parse error:
    ^/\/*/user$
      ^^
```

It will replace #9992.

See KAG-329.

### DIFF
--- a/kong/db/schema/others/migrations.lua
+++ b/kong/db/schema/others/migrations.lua
@@ -6,6 +6,7 @@ return {
       postgres  = {
         type = "record", required = true,
         fields = {
+          { up_t = { type = "function" } },
           { up = { type = "string", len_min = 0 } },
           { up_f = { type = "function" } },
           { teardown = { type = "function" } },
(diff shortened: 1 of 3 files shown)


kong/db/schema/others/migrations.lua
```diff
@@ -6,6 +6,7 @@ return {
       postgres  = {
         type = "record", required = true,
         fields = {
+          { up_t = { type = "function" } },
```
chronolaw, line 9: Here `up_t` means `up try` or `up test`.

colinh-kong (changes requested):
Is there an early preview of documentation related to how a user achieves a zero-downtime upgrade should they have an incompatible route?

chronolaw:
This change will only be applied to 3.1.x, so I created another PR #10092.

chronolaw:
> Is there an early preview of documentation related to how a user achieves a zero-downtime upgrade should they have an incompatible route?

Sorry, till now there is no document for it. But I think when the PR #10092 is ready we will have one.

PR #10092, opened 2023-01-11
chore(db/migration): check migration regex path before db changes

chronolaw:
### Checklist

```
  |
  |  ^------------------------^
  |
  = regex parse error:
    ^/\/*/user$
      ^^
```

It will replace #9992 #10089.

Reference: https://github.com/Kong/kong/commit/f66c3ecb1bd8e335e20b6f751f71ef3b55085345

See KAG-329, we also need a user document to tell user how to solve these errors.


kong/db/migrations/validate_path_280_300.lua
```diff
@@ -0,0 +1,121 @@
...
+
+
+local validate_atc_expression
+do
+  local router = require("resty.router.router")
+  local CACHED_SCHEMA = require("kong.router.atc").schema
+  local get_expression = require("kong.router.compat").get_expression
+
+  validate_atc_expression = function(route)
+    local r = router.new(CACHED_SCHEMA)
```
dndx, line 18: We can cache the `router`.

```
if not r then
  r = router.new(CACHED_SCHEMA)
end

...

local res, err = r:add_matcher(0, route.id, exp)
if not res then
   log.error("Regex path will not work with the default router flavor in 3.0, " ..
                 "route id: %s, err: %s", route.id, err)
   return false
end

r:remove_matcher(route.id)
```

kong/db/migrations/validate_path_280_300.lua
```diff
@@ -0,0 +1,121 @@
+local log = require "kong.cmd.utils.log"
+
+local assert = assert
+local ipairs = ipairs
+local migrate_path = require "kong.db.migrations.migrate_path_280_300"
+
+
+local MAX_COUNT = 10
```
dndx, line 8:
```suggestion
local MAX_ERRORS_AT_A_TIME = 10
```

kong/db/migrations/validate_path_280_300.lua
chronolaw: done

kong/db/migrations/validate_path_280_300.lua
chronolaw: done

kong/db/migrations/validate_path_280_300.lua
chronolaw: This improvement can be port to master branch later.

dndx (approved):

PR #10096, opened 2023-01-11
feat(tracing): route/service scoping

StarlightIbuki:
implement KAG-244

### DIFF
--- a/spec/03-plugins/37-opentelemetry/04-exporter_spec.lua
+++ b/spec/03-plugins/37-opentelemetry/04-exporter_spec.lua
@@ -172,7 +172,7 @@ for _, strategy in helpers.each_strategy() do
     } do
       describe("#scoping for" .. (case[1] and " route" or "")
                               .. (case[2] and " service" or "")
-                              .. (case[3] and "with global" or "")
+                              .. (case[3] and " with global" or "")
       , function ()
         lazy_setup(function()
           bp, _ = assert(helpers.get_db_utils(strategy, {
@@ -213,8 +213,13 @@ for _, strategy in helpers.each_strategy() do
           local ok, err = thread:join()
 
           -- we should have no telemetry reported
-          assert.is_falsy(ok)
-          assert.matches("timeout", err)
+          if case[3] then
+            assert(ok, err)
+
+          else
+            assert.is_falsy(ok)
+            assert.matches("timeout", err)
+          end
         end)
       end)
     end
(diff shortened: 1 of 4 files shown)


kong/tracing/tracer.lua
```diff
@@ -0,0 +1,97 @@
+local _M = {}
```
mayocream, line 1: Is it possible to combine the `tracer.lua` with `pdk/tracing.lua`?

StarlightIbuki: done.

spec/03-plugins/37-opentelemetry/04-exporter_spec.lua
```diff
@@ -52,17 +52,22 @@ for _, strategy in helpers.each_strategy() do
...
+                                              paths = { "/" }}))
 
-      bp.plugins:insert({
+      assert(bp.routes:insert({ service = http_srv,
+                                protocols = { "http" },
+                                paths = { "/no_plugin" }}))
+
+      assert(bp.plugins:insert({
         name = "opentelemetry",
+        route = route,
```
mayocream, line 65: Do not apply otel plugin on the route by default, add an arg in `setup_instrumentations` to configure it instead.

spec/03-plugins/37-opentelemetry/04-exporter_spec.lua
```diff
@@ -140,6 +145,52 @@ for _, strategy in helpers.each_strategy() do
       end)
     end)
 
+    describe("#scoping", function ()
```
mayocream, line 148: Otel plugin applied on services also need to be tested

spec/03-plugins/37-opentelemetry/04-exporter_spec.lua
```diff
@@ -140,6 +145,52 @@ for _, strategy in helpers.each_strategy() do
...
+          path    = "/no_exist",
+        })
+        assert.res_status(404, r)
+
+        -- close client connection
+        cli:close()
+
+        local ok, err = thread:join()
+
+        ngx.sleep(1000)
```
mayocream, line 187: use wait_util

StarlightIbuki: This is a mistakenly committed debugging code.

fffonion:
I feel like we are mixing the source of truth of "whether to trace or not" in different places. Previously when
otel plugin can only be global:
1. we have a `opentelemetry_tracing` config that tells Kong whether to collect specific instrumentations
2. we have otel plugin that controls whether and where a collected spans needs to be flushed to

In this PR we have an additional source of truth on "whether to trace or not" depending on if any of a tracing
plugin (otel in this case) is enabled on current path, and that conflicts with fact that it's previously only
controlled by `opentelemetry_tracing`. So i would suggest we refactor the implementation a bit so that:
1. we still rely on `opentelemetry_tracing` config that tells Kong whether to collect specific instrumentations
2. rely on plugins_iterator to pick a plugin on current request, if such otel plugin exists, then its `log` phase
phase handler is executed, and spans are flushed.

so that we don't need complex logic over iterration over plugins and cache invalidations.

mayocream (approved):

PR #10108, opened 2023-01-13
tests(tracing): add test groups for propagation.set test

catbro666:
to keep consistent with [Kong/kong-ee#4310](https://github.com/Kong/kong-ee/pull/4310)

FTI-4634


kikito:
It looks like this PR might not have all the contents of the ee one - this one only has tests.

catbro666:
> It looks like this PR might not have all the contents of the ee one - this one only has tests.

because the datadog header type is an ee-only feature.

spec/01-unit/26-tracing/02-propagation_spec.lua
```diff
@@ -615,36 +616,39 @@ describe("propagation.set", function()
...
+  -- use big_trace_id here because "0000000000000001" can be regarded as
+  -- both hexadecimal or decimal (the same value), so we can't ensure
+  -- if the correct `to_hex()` and `to_dec()` is used.
   local proxy_span = {
-    trace_id = from_hex(trace_id),
-    span_id = from_hex(span_id),
-    parent_id = from_hex(parent_id),
+    trace_id = from_hex(big_trace_id),
+    span_id = from_hex(big_span_id),
+    parent_id = from_hex(big_parent_id),
```
mayocream, line 625: Does it make sense to add a matrix to test both regular and big trace/span IDs?

similar to https://github.com/Kong/kong/blob/a310920e8101889ccb84b45e79df4a00b5d0b44d/spec/03-plugins/34-zipkin/zipkin_no_endpoint_spec.lua#L20

kikito: I think tests are one of the few places where a bit of repetition is acceptable. No strong opinion for or against.

colinh-kong (approved):

PR #10110, opened 2023-01-13
feat(core): add front requeuing to batch_queue

usysrc:
### Summary

This PR proposes the optional parameter 'process_requeue_front' which adds requeued entries back to the front of the queue.  
Current default behavior are that entries get requeued to the back and isn't changed when this parameter is unset or set to `false`.

Motivation:  
With the default behavior e.g. used in the [http-log plugin](https://github.com/Kong/kong/blob/b95ae8060b8be2b2327bcbade822f4e104825099/kong/plugins/http-log/handler.lua#L1) there can be situations where the queue piles up a lot of data in case entries cannot be processed since the queue has a LIFO behavior in case of errors.  
In case the process error persists over several minutes the retry delay of the queue is increased to 60s. When new entries are stored in front of the queue and processed first they will also fail. This means older batches are not processed, the retry counter for these old batches won't be increased and they won't expire causing high-memory usage in the end.

With this PR a FIFO behavior is introduced ensuring, that the the retry for the oldest batch is increased first and the batch is actually dropped when max_retries is reached before processing newer records.

### Checklist

The program was tested solely for our own use cases, which might differ from yours.
Tilmann Hars <tilmann.hars@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, Provider Information [https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

### DIFF
--- a/kong/tools/batch_queue.lua
+++ b/kong/tools/batch_queue.lua
@@ -27,11 +27,12 @@
 --     name, -- name of the queue for identification purposes in the log
 --     process, -- function used to "process/consume" values from the queue
 --     { -- Opts table with control values. Defaults shown:
---       retry_count        = 0,    -- number of times to retry processing
---       batch_max_size     = 1000, -- max number of entries that can be queued before they are queued for processing
---       process_delay      = 1,    -- in seconds, how often the current batch is closed & queued
---       flush_timeout      = 2,    -- in seconds, how much time passes without activity before the current batch is closed and queued
---       max_queued_batches = 100,  -- max number of batches that can be queued before the oldest batch is dropped when a new one is queued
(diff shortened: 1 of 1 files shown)


hanshuebner:
@usysrc Thank you for your submission.  We've already made the behavior that you proposed to be made configurable the new (and only) behavior as we think that reordering the queue items was a mistake in the first place.  Do you think that it should really be configurable?

usysrc:
I didn't realize that this change was made. Thank you for pointing that out. I don't think this needs to be configurable and will retract this PR.

PR #5045, opened 2019-09-17
remove hardcoded charset #5044

vasartori:
### Summary

When we set the environment variable nginx_http_charset, a message of duplicated entry is raised.
When this environment varibale is set, kongs set it on nginx.conf, but the charset is hardcoded on nginx_kong.conf.
With this approach, we are not abble to use a charset difrerent of UTF-8

The default value of nginx is off.

This behaviour broke applications who uses eg iso-8859-1 charsets.

### Full changelog

* Use the default nginx charset (default none)
* Stop with issue when we set a environment variable to set charset.

### Issues resolved

Fix #5044

PR #10111, opened 2023-01-13
fix(template) keep UTF-8 as default for charset directive but make it configurable

bungle:
### Summary

`charset` directive that was previously hard coded in template with a value of `UTF-8` was removed completely with this PR:
https://github.com/Kong/kong/pull/9905

Previous discussion from 3 years ago pointed that this is breaking change (not a fix): https://github.com/Kong/kong/pull/5045

So here is another fix to make the directive still default to `UTF-8`. which is a quite fine default, but allow users to configure it. E.g.:

```
KONG_NGINX_HTTP_CHARSET=off kong start
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,9 @@
   Expression router instead of the traditional router to ensure created routes
   are actually compatible.
   [#9987](https://github.com/Kong/kong/pull/9987)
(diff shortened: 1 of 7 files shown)


colinh-kong (approved):
where / how do we add a test related to kong portal?

PR #10119, opened 2023-01-17
style(core/runloop):  style clean for certificate.lua

chronolaw:
### Checklist

### Full changelog

* localize some functions
* change `new_tab(1, 0)` to `new_tab(0, 1)` in `fetch_ca_certificates`

### DIFF
--- a/kong/runloop/certificate.lua
+++ b/kong/runloop/certificate.lua
@@ -19,9 +19,14 @@ local set_cert = ngx_ssl.set_cert
 local set_priv_key = ngx_ssl.set_priv_key
 local tb_concat   = table.concat
 local tb_sort   = table.sort
+local type = type
+local error = error
+local assert = assert
 local tostring = tostring
 local ipairs = ipairs
 local ngx_md5 = ngx.md5
+local ngx_exit = ngx.exit
+local ngx_ERROR = ngx.ERROR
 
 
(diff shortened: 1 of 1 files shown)


kong/runloop/certificate.lua
```diff
@@ -302,7 +308,7 @@ end
 
 local function fetch_ca_certificates(ca_ids)
   local cas = new_tab(#ca_ids, 0)
-  local key = new_tab(1, 0)
+  local key = new_tab(0, 1)
```
chronolaw, line 311: Here key was used by form `key.id`, I think it is a hash table, so we should use `new_tab(0, 1)`.

ADD-SP (approved):

kong/runloop/certificate.lua
bungle: why even create a table when there is only 1?

bungle: You could have module level:
```
local CA_KEY = {
  id = ""
}
```

and then just use:
```
CA_KEY.id = ca_id
local obj, err = kong.db.ca_certificates:select(CA_KEY)
```

chronolaw: good idea, thank you @bungle

bungle (approved):

PR #10120, opened 2023-01-17
fix(ca_certificates) prevent CA to be deleted that utilized by other entities

ms2008:
### Summary

Prevent CA to be deleted that utilized by other entities. Currently, Postgres does not support creating foreign keys on fields of type array, so we can't do this by schema constraint.

### Checklist

### Full changelog

* prevent CA to be deleted that utilized by other entities

### Issue reference

Fix FTI-4343 FTI-2060

### DIFF
--- a/kong/api/routes/ca_certificates.lua
+++ b/kong/api/routes/ca_certificates.lua
@@ -1,0 +1,56 @@
+local ngx = ngx
+local kong = kong
+local ipairs = ipairs
+local null = ngx.null
+local string_format = string.format
+
+
+local function each_service(entity)
+  local options = {
+    workspace = null,
+  }
+
+  local iter = entity:each(1000, options)
+  local function iterator()
+    local element, err = iter()
+    if err then return nil, err end
+    if element == nil then return end
+    if element.ca_certificates and #element.ca_certificates > 0 then return element, nil end
+    return iterator()
+  end
+
+  return iterator
+end
+
+
+local function verify_references(id)
+  -- services
+  for service, err in each_service(kong.db.services) do
+    if err then
+      kong.log.err("could not load services: ", err)
+      return
+    end
+
+    for _, v in ipairs(service.ca_certificates) do
+      if v == id then
+        kong.log.notice(string_format("ca_certificate: %s is still referenced from service: %s", id, service.id))
+
+        return kong.response.exit(400, {
+          name = "foreign key violation",
+          fields = { ["@referenced_by"] = "services" },
+          message = "an existing 'services' entity references this 'ca_certificates' entity"
+        })
+      end
+    end
+  end
+end
+
+return {
+  ["/ca_certificates/:ca_certificates"] = {
+    DELETE = function(self, db, helpers, parent)
+      verify_references(self.params.ca_certificates)
+
+      return parent()
+    end,
+  },
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,6 +162,8 @@
 - Fix an issue where sorting function for traditional router sources/destinations lead to "invalid order
   function for sorting" error.
   [#10514](https://github.com/Kong/kong/pull/10514)
+- Prevent CA certificate to be deleted that utilized by other entities.
+  [#10120](https://github.com/Kong/kong/pull/10120)
 
 #### Admin API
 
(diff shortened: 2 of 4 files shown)


kong/api/routes/ca_certificates.lua
```diff
@@ -0,0 +1,56 @@
...
+
+local function verify_references(id)
+  -- services
+  for service, err in each_service(kong.db.services) do
+    if err then
+      ngx.log(ngx.ERR, "could not load services: ", err)
+      return
+    end
+
+    for _, v in ipairs(service.ca_certificates) do
```
dndx, line 34: This kind of reverse reference from `

dndx, line 34: This seems to be a very hacky solution, as it does not work with anything else that might also reference `ca_certificate` by UUID. I wonder how useful such specific solution is.

kong/api/routes/ca_certificates.lua
```diff
@@ -0,0 +1,56 @@
...
+
+local function each_service(entity)
+  local options = {
+    workspace = null,
+  }
+
+  local iter = entity:each(1000, options)
+  local function iterator()
+    local element, err = iter()
+    if err then return nil, err end
```
chronolaw, line 16: Not a good style, should not let `if-then-end` in one line.

kong/api/routes/ca_certificates.lua
```diff
@@ -0,0 +1,56 @@
...
+
+  return iterator
+end
+
+
+local function verify_references(id)
+  -- services
+  for service, err in each_service(kong.db.services) do
+    if err then
+      ngx.log(ngx.ERR, "could not load services: ", err)
```
chronolaw, line 30: localize `ngx.log` `ngx.ERR`

hanshuebner: Why should rarely called functions be localized?  There is no benefit, but inventing extra names makes the code harder to review and refactor.

kong/api/routes/ca_certificates.lua
catbro666: I agree. Things like mtls-auth plugins will also reference `ca_certificate`. A general solution may require support of entity back-reference mechanism which I think is a big feature.

ms2008: Indeed, this is not an elegant solution, and if new references are added in the future, it is up to the developer to add the corresponding filter detection.

hbagdi:
Please rebase.

kong/api/routes/ca_certificates.lua
sabertobihwy: use `kong.log.err(...` is better?

kong/api/routes/ca_certificates.lua
```diff
@@ -0,0 +1,56 @@
...
+      return
+    end
+
+    for _, v in ipairs(service.ca_certificates) do
+      if v == id then
+        ngx.log(ngx.NOTICE, string_format("ca_certificate: %s is still referenced from service: %s", id, service.id))
```
sabertobihwy, line 36: `kong.log.notice`

kong/api/routes/ca_certificates.lua
```diff
@@ -0,0 +1,56 @@
...
+
+        return kong.response.exit(400, {
+          name = "foreign key violation",
+          fields = { ["@referenced_by"] = "services" },
+          message = "an existing 'services' entity references this 'ca_certificates' entity"
+        })
+      end
+    end
+  end
+end
```
hbagdi, line 46: CA certificates can also be referenced by mtls-auth plugins. What about those?

ms2008: As the mtls-auth plugin is an enterprise edition plugin, my idea is that once this PR is merged, it can be synchronized to EE and then add a separate detection for mtls-auth in EE.

hbagdi: Can you track that work in a Jira and link it here?

ms2008: done.

ms2008:
After some internal discussions, we are not satisfied with the current solution due to its maintainability and poor performance. Therefore, we have decided to close this until a better solution can be found.

PR #10121, opened 2023-01-17
fix(pdk/request): support overriding the limit for get_query()

liverpool8056:
### Summary

Currently, get_query() method in pdk.request can't support to override the limitation for getting all query parameter. This pr is to support it by allowing max_args=0 to be passed to `ngx.req.get_uri_args` for some special cases. 

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

FTI-2525

### DIFF
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -39,7 +39,7 @@ local function new(self)
   local MIN_HEADERS            = 1
   local MAX_HEADERS_DEFAULT    = 100
   local MAX_HEADERS            = 1000
-  local MIN_QUERY_ARGS         = 1
+  local MIN_QUERY_ARGS         = 0
   local MAX_QUERY_ARGS_DEFAULT = 100
   local MAX_QUERY_ARGS         = 1000
   local MIN_POST_ARGS          = 1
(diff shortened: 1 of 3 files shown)


kong/pdk/request.lua
```diff
@@ -39,7 +39,7 @@ local function new(self)
   local MIN_HEADERS            = 1
   local MAX_HEADERS_DEFAULT    = 100
   local MAX_HEADERS            = 1000
-  local MIN_QUERY_ARGS         = 1
+  local MIN_QUERY_ARGS         = 0
```
jschmid1, line 42: This change isn't documented nor tested. What's the intent behind this?

StarlightIbuki: This is documented [here](https://github.com/openresty/lua-nginx-module#ngxreqget_uri_args).
When set to 0 it means no limit to the number.

StarlightIbuki (commented):
This is risky. Personally, I don't think we really need "0". A larger number should work for most of the case.

StarlightIbuki:
Do you have a use case for this feature request?

liverpool8056:
@StarlightIbuki As per to FTI-2525, I believe some users want it, and I understand even there is, it's not really common. It's cool for me not to merge if this may introduce potential risks. But from my perspective, at least it seems we restrict the original ability from openresty or whatever. The callers who intend to use it with max_args=0 should have some knowledge about what they are doing.

kikito (changes requested):
Instead of changing MIN_QUERY_ARGS, I think we could allow changing MAX_QUERY_ARGS_DEFAULT. It should still be limited to not higher than 1000, not lower than 1.

liverpool8056:
To avoid potential risk and there's no feature request for now, won't do at this momnet.

PR #10133, opened 2023-01-18
fix(api): dynamically sets log level for timers pre-created by timer-ng

gruceo:
If the log level was dynamically set, we must also change the log level for all timers pre-created by the timer-ng library.

We're going to use the global _G table and not a shared dictionary due to performance reasons.

### DIFF
--- a/kong-3.2.0-0.rockspec
+++ b/kong-3.2.0-0.rockspec
@@ -41,7 +41,7 @@ dependencies = {
   "lua-resty-ipmatcher == 0.6.1",
   "lua-resty-acme == 0.10.1",
   "lua-resty-session == 4.0.0",
-  "lua-resty-timer-ng == 0.2.0",
+  "lua-resty-timer-ng == 0.2.2",
 }
 build = {
   type = "builtin",
--- a/kong/api/routes/debug.lua
+++ b/kong/api/routes/debug.lua
@@ -64,6 +64,10 @@ local function handle_put_log_level(self, broadcast)
     return kong.response.exit(500, { message = message })
   end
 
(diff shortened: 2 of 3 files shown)


ADD-SP:
Just a note, we made a workaround within timer-ng, and while it still has some limitations, we're at least going to make this feature work.

In the future, we should rewrite the C code to implement this feature elegantly, and @gruceo  will create internal tickets to track this thiing.

dndx:
@gruceo Could you put the ticket number for improving dynamic log level in a comment? We need to properly address the hacks we introduced inside our codebase through proper C code change and @ADD-SP can help with that.

dndx:
I suggest we take this PR out of 3.2.0 as the hack is too specific to the timer library we use and I have a feeling that there is a better and more systematic way to fix these and similar issues. @ADD-SP can help get this moving towards that direction.

@kikito @guanlan any objections? This does not looks too urgent to me.

gruceo:
@dndx thank you for your suggestion, I dropped a comment to KAG-457. Tests are passing now. Note this bug does not happen with native `ngx.timer` API. The alternative way to make dynamic log levels work with our timer-ng library is a complete rewrite of our dynamic log levels C code, and it may require a patch (see KAG-457 for details). I suggest we ship this PR as-is for now until we complete work on KAG-457 and since we already uploaded to LuaRocks a new version of the timer-ng library [with this fix](https://github.com/Kong/lua-resty-timer-ng/commit/91e4852496319774b0161c4fb2aae807f10ce3bd) I don't think it makes sense to merge https://github.com/Kong/kong/pull/10193 without this PR. What do you think? (cc @fffonion @ADD-SP)

PR #10142, opened 2023-01-19
chore(deps) remove legacy worker-events

bungle:
### Summary

Removes legacy worker events.


chronolaw:
Should we do it in 4.0 release?

locao:
(If we are moving forward with this change in 3.x) to make the unit tests green you probably need something similar to [what was used in lua-resty-healthcheck](https://github.com/Kong/lua-resty-healthcheck/blob/1.6.2/t/with_resty-events/01-start-stop.t#L15-L23).
```
    init_worker_by_lua_block {
        local we = require "resty.events.compat"
        assert(we.configure({
            unique_timeout = 5,
            broker_id = 0,
            listening = "unix:$ENV{TEST_NGINX_SERVROOT}/worker_events.sock"
        }))
        assert(we.configured())
    }
```

bungle:
> Should we do it in 4.0 release?

I think we should get rid of this in 3.3 or something. This is behind undocumented setting and it is default to to use `resty.events.compat`, in 4.0 I think we should move to `resty.events` (the non-compat).

hbagdi:
What is the risk of keeping legacy worker-events around for longer?
Do we think we have enough feedback from production users to warrant us removing this fallback option completely?

hbagdi:
Aapo, are we still pursuing this one?

bungle:
@hbagdi this is a bit stuck because it is quite hard to get the unit tests pass (old worker-events worked nicely with unit tests, but this thing needs listeners etc.), thus we have KAG-930 in discussion, to help get this green.

bungle:
@hbagdi, it is partly on product to decide whether we want this now, later, or e.g. with 4.x. But well nobody has used the `legacy worker events`, and we don't even test that anymore.

hbagdi:
How difficult is to test the old worker events library with the existing code?
If we turn on the configuration option in our CI and test it once, does it work?

chronolaw:
The library `lua-resty-events` has added a new `listening=off` feature for this PR.
https://github.com/Kong/lua-resty-events/pull/24

bungle:
Ping @hbagdi / @chronolaw.

This PR looks to be finally _green_.

> Can we pursue testing with the legacy library more rather than dropping this?

I gave some thought to it, but the feeling I have, the time has passed on it. We are approaching 3.3 and nobody has ever asked about it. We haven't ran tests since we switched to new with the old one anymore. Testing would potentially mean running the whole test suite with `legacy_events=on`, as it is quite hard to know what are all the code that would affect it.

I think we should either pursue on getting this merged on 3.3 or 3.4, or even wait for 4.0. I don't think spending time on testing `legacy_events=on` is worth it. We are not returning to it. So this is just about how long should we keep such undocumented latch, that we don't have a clear understanding if it causes other issues (when turned on). It probably works, just that we haven't ran the test suite anymore with that on.

@chronolaw / @hbagdi what is your take?

chronolaw:
I think that merging it into 3.3 is a little risky, 3.4 or  later version will give us more time for testing.

hbagdi:
@dndx Please opine.

dndx:
I don't have too much concerns over removing it in 3.3. 3.x has been out for a while and we never heard of complains about worker events, so it seems to be comfortable enough.

kikito:
Even though no one has raised problems on 3.3 having this escape hatch does not harm us too much. Since Chrono has concerns, I have decided to remove this from the 3.3 milestone. Let's merge it in 3.4.

@dndx , @hbagdi feel free to re-add to the milestone.

bungle:
Just a note: we have ran with new events since September 2022 (and nobody has mentioned this undocumented config option anywhere).

hbagdi:
I'll defer the decision-making to all of you here.

chronolaw (approved):

PR #10150, opened 2023-01-20
feat(db): add `description` to metaschema fields

jschmid1:
### Summary

This allows for in-place documentation of fields.
If we agree to follow this more meticulously we could generate documentation from code.

Signed-off-by: Joshua Schmid <jaiks@posteo.de>

### DIFF
--- a/kong/db/schema/metaschema.lua
+++ b/kong/db/schema/metaschema.lua
@@ -72,6 +72,7 @@ local field_schema = {
   { type = { type = "string", one_of = keys(Schema.valid_types), required = true }, },
   { required = { type = "boolean" }, },
   { reference = { type = "string" }, },
+  { description = { type = "string" }, },
   { auto = { type = "boolean" }, },
   { unique = { type = "boolean" }, },
   { unique_across_ws = { type = "boolean" }, },


kong/db/schema/metaschema.lua
```diff
@@ -72,6 +72,7 @@ local field_schema = {
   { type = { type = "string", one_of = keys(Schema.valid_types), required = true }, },
   { required = { type = "boolean" }, },
   { reference = { type = "string" }, },
+  { descrption  = { type = "string" }, },
```
hanshuebner, line 75: But let's call it descr`i`ption please

hanshuebner (approved):

flrgh (approved):
I like it :+1:

StarlightIbuki (approved):

hbagdi:
I'm going to increase the feature scope here to make this addition a bit more effective: A schema is invalid if there is no description field on every filed. We can relax that for unbundled plugins but for all bundled plugins and all core schemas, we should add the check.

That will increase the burden of this task but it will go a long way to document the API.

My concern with not doing so is that some fields will have a description field while others won't and narrowing that gap becomes a whack-a-mole problem for a long time.

Tieske:
> A schema is invalid if there is no description field on every filed. We can relax that for unbundled plugins but for all bundled plugins and all core schemas, we should add the check.

This feels to me as hard to implement. If the goal is to have consistency, without breaking compatibility, can we instead add a "warning" level log message? 

Additionally; can we run this by the GUI folks of Konnect? I would assume they would use this field for display in the GUI as well, but they might have other requirements. Maybe some validations for safety, or multiple fields; short-description + long-description for tooltip vs documentation purposes.
Not intended as scope creep, but more to check forward compatibility wrt future features we'd like.

hbagdi:
Is this going anywhere?
I don't mind an incremental improvement to add the field and then slowly add descriptions.
@Kong/team-docs We would love to hear what you all think!

Guaris:
I think this is a very good addition. It would benefit the docs team if a schema was considered invalid without a description field. I'm happy to take on the burden of adding these descriptions to all bundled plugins and all core schemas if it helps add all the descriptions at once. @hbagdi

hbagdi:
If the docs team can help us out here, @jschmid1 we should push this out as fast as possible.
Once the descriptions have been added, we need to add a test to fail CI whenever a field without a description is added in.

Guaris:
I have a branch here w/ the definitions that anyone can look at 
https://github.com/Kong/kong/tree/add-descriptions

Tieske:
@Guaris nice work! Maybe you can create a PR against this branch to make it easier to review?

This however does raise some new questions;
- what is the format? the descriptions added include markdown weblinks, see also my "additionally" comment [here](https://github.com/Kong/kong/pull/10150#issuecomment-1442569114)
- should we limit the size? (I think we should)
- guidelines on what is to go in the description? for example the word "optional" which is used a lot, but it is already implicit in the other metadata. Or for example number values in a range, the range is in the metadata, should the description repeat it?

Guaris:
> @Guaris nice work! Maybe you can create a PR against this branch to make it easier to review?
> 
> This however does raise some new questions;
> 
> * what is the format? the descriptions added include markdown weblinks, see also my "additionally" comment [here](https://github.com/Kong/kong/pull/10150#issuecomment-1442569114)
> * should we limit the size? (I think we should)
> * guidelines on what is to go in the description? for example the word "optional" which is used a lot, but it is already implicit in the other metadata. Or for example number values in a range, the range is in the metadata, should the description repeat it?


Hey I made a PR here: https://github.com/Kong/kong/pull/10706


> * should we limit the size? (I think we should)

What size would you suggest? 

> guidelines on what is to go in the description? for example the word "optional" which is used a lot, but it is already >implicit in the other metadata. Or for example number values in a range, the range is in the metadata, should the >description repeat it?

It depends on what the intended use of these descriptions is, but I do worry about scope creep here. My hope with my PR was to take care of the work of inserting the descriptions, and since we didn't have descriptions before I thought this would be a good foundation.

hbagdi:
I highly recommend letting the documentation team take the lead on this effort. We may not have something perfect from the get-go and these details are relatively easy to change.
Only adding descriptions that are (in most cases) not longer than one sentence is a very good start. Avoiding markdown and weblinks is for the best at this point.

Thoughts from others?

jschmid1:
> I highly recommend letting the documentation team take the lead on this effort. We may not have something perfect from the get-go and these details are relatively easy to change. Only adding descriptions that are (in most cases) not longer than one sentence is a very good start. Avoiding markdown and weblinks is for the best at this point.
> 
> Thoughts from others?

+1 @Guaris already started here -> https://github.com/Kong/kong/pull/10706

Tieske:
> What size would you suggest?
> 

somewhere 1 to 4 kb? enough for a detailed description, but still protecting against abuse. Typically easier to relax a constrained like this in a later stage, than trying to restrict it later.

> Avoiding markdown and weblinks is for the best at this point.

👍

jschmid1:
closed in favor of #10765

PR #10152, opened 2023-01-21
refactor(clustering/compat): using array to store compatible checker functions

chronolaw:
### Summary

Move compatible config-checking logic into a hash table, making code more readable.

Need special cherry-pick action to EE.

### Checklist

### DIFF
--- a/kong/clustering/compat/checkers.lua
+++ b/kong/clustering/compat/checkers.lua
@@ -1,8 +1,6 @@
 local ipairs = ipairs
 
 
-
-
 local KONG_VERSION = require("kong.meta").version
 
 
--- a/kong/clustering/compat/checkers.lua
+++ b/kong/clustering/compat/checkers.lua
@@ -35,13 +35,16 @@ local compatible_checkers = {
 
       for _, name in ipairs(entity_names) do
         for _, config_entity in ipairs(config_table[name] or {}) do
-          ngx_log(ngx_WARN, _log_prefix, "Kong Gateway v" .. KONG_VERSION ..
-            " contains configuration '" .. name .. ".updated_at'",
-            " which is incompatible with dataplane version " .. dp_version .. " and will",
-            " be removed.", log_suffix)
+          log_warn_message("contains configuration '" .. name .. ".updated_at'",
+                           "be removed",
+                           dp_version,
+                           log_suffix)
           config_entity.updated_at = nil
+          has_update = true
         end
       end
+
+      return has_update
     end
   },
 
@@ -70,7 +73,7 @@ local compatible_checkers = {
       end
 
       if has_update then
-        log_warn_message("tls protocol service contains configuration 'service.client_certificate' " ..
+        log_warn_message("tls protocol service contains configuration 'service.client_certificate'" ..
                          "or 'service.tls_verify' or 'service.tls_verify_depth' or 'service.ca_certificates'",
                          "be removed",
                          dp_version,
--- a/kong/clustering/compat/checkers.lua
+++ b/kong/clustering/compat/checkers.lua
@@ -1,13 +1,27 @@
 local ipairs = ipairs
 
 
-local ngx = ngx
-local ngx_log = ngx.log
-local ngx_WARN = ngx.WARN
 
 
-local _log_prefix = "[clustering] "
 local KONG_VERSION = require("kong.meta").version
+
+
+local log_warn_message
+do
+  local ngx_log = ngx.log
+  local ngx_WARN = ngx.WARN
+  local fmt = string.format
+
+  local _log_prefix = "[clustering] "
+
+  log_warn_message = function(hint, action, dp_version, log_suffix)
+    local msg = fmt("Kong Gateway v%s %s " ..
+                    "which is incompatible with dataplane version %s " ..
+                    "and will %s.",
+                    KONG_VERSION, hint, dp_version, action)
+    ngx_log(ngx_WARN, _log_prefix, msg, log_suffix)
+  end
+end
 
 
 local compatible_checkers = {
@@ -58,11 +72,11 @@ local compatible_checkers = {
       end
 
       if has_update then
-        ngx_log(ngx_WARN, _log_prefix, "Kong Gateway v" .. KONG_VERSION ..
-                " tls protocol service contains configuration 'service.client_certificate'",
-                " or 'service.tls_verify' or 'service.tls_verify_depth' or 'service.ca_certificates'",
-                " which is incompatible with dataplane version " .. dp_version .. " and will",
-                " be removed.", log_suffix)
+        log_warn_message("tls protocol service contains configuration 'service.client_certificate' " ..
+                         "or 'service.tls_verify' or 'service.tls_verify_depth' or 'service.ca_certificates'",
+                         "be removed",
+                         dp_version,
+                         log_suffix)
       end
 
       return has_update
@@ -85,10 +99,10 @@ local compatible_checkers = {
       end
 
       if has_update then
-        ngx_log(ngx_WARN, _log_prefix, "Kong Gateway v" .. KONG_VERSION ..
-                " configuration 'upstream.algorithm' contain 'latency' option",
-                " which is incompatible with dataplane version " .. dp_version .. " and will",
-                " be fall back to 'round-robin'.", log_suffix)
+        log_warn_message("configuration 'upstream.algorithm' contains 'latency' option",
+                         "fall back to 'round-robin'",
+                         dp_version,
+                         log_suffix)
       end
 
       return has_update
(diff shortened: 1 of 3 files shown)


kong/clustering/compat/init.lua
```diff
@@ -334,35 +387,9 @@ function _M.update_compatible_payload(payload, dp_version, log_suffix)
...
-        if t["use_srv_name"] ~= nil then
-          ngx_log(ngx_WARN, _log_prefix, "Kong Gateway v" .. KONG_VERSION ..
-                  " contains configuration 'upstream.use_srv_name'",
-                  " which is incompatible with dataplane version " .. dp_version .. " and will",
-                  " be removed.", log_suffix)
-          t["use_srv_name"] = nil
-          has_update = true
-        end
-      end
+  for ver, fn in pairs(compatible_checkers) do
```
flrgh, line 390: I'm not sure that `pairs()` is sufficient for this. There is an expectation that the compatibility change functions are executed in descending order (`3.2.0.0`, then `3.1.0.0`, then `3.0.0.0`, etc).

Curious, have you taken a look at the [compat code in koko](https://github.com/Kong/koko/blob/1b5cdf241279f7a49d8c637f9bc43babd7e35ae8/internal/server/kong/ws/config/compat/compatibility_table.go#L95-L1047)? It's not without its quirks, but I think this would be a good pattern for us to follow--pretty big task to implement though.

chronolaw: Thanks for your opinion. I have changed the hash table to an array.

Do you think it is OK?

chronolaw:
Need update to #10192

kong/clustering/compat/init.lua
```diff
@@ -312,6 +312,61 @@ do
...
+                " which is incompatible with dataplane version " .. dp_version,
+                " and will be removed.", log_suffix)
+      end
+
+      return has_update
+    end
+  },
+}
+
+
```
attenuation, line 369: I think It will be nice if you can add this table to separate lua file like this https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua

attenuation:
For the compact problem, I have another piece of advice. we have all of the kong schemas, and we know all of the kong version schemas,  Could this compatibility logic be solved in a similar way to how we use code generation in [bindgen](https://github.com/rust-lang/rust-bindgen)? we use more reliable code(iterating all kong schema) to generate the  Version compatibility code.

kong/clustering/compat/init.lua
chronolaw: done.

kong/clustering/compat/checkers.lua
```diff
@@ -0,0 +1,129 @@
...
+          t["algorithm"] = "round-robin"
+          has_update = true
+        end
+      end
+
+      if has_update then
+        ngx_log(ngx_WARN, _log_prefix, "Kong Gateway v" .. KONG_VERSION ..
+                " configuration 'upstream.algorithm' contain 'latency' option",
+                " which is incompatible with dataplane version " .. dp_version .. " and will",
+                " be fall back to 'round-robin'.", log_suffix)
```
chronolaw, line 68: Here should be `fall back`, remove `be`?

chronolaw:
Need update for #10400

liverpool8056 (approved):

attenuation (approved):

vm-001 (approved):
The instance_name part looks OK to me. And also has a corresponding test to cover that.

chronolaw:
> For the compact problem, I have another piece of advice. we have all of the kong schemas, and we know all of the kong version schemas, Could this compatibility logic be solved in a similar way to how we use code generation in [bindgen](https://github.com/rust-lang/rust-bindgen)? we use more reliable code(iterating all kong schema) to generate the Version compatibility code.

It is a good idea, perhaps we can do it in another PR in the future.

vm-001 (approved):
Echoing request-review request from Harry.

1/ From the maintainability perspective

1. I don't see any `sort` function being called before iterating the `COMPATIBILITY_CHECKERS`, which means the program correctness is relying on the order of definition. Could we add some comments above the `compatible_checkers` var to describe that? Or could we introduce a sort mechanism?

2. We may add more and more properties to each checker. I feel using the array index `checker[n]` is not friendly. What do you think of using map table to replace array table for each checker?

```lua
local compatible_checkers = {
  { 
    -- desc = "resolve plugin.instance_name",
    -- ... any other properties
    version = x, 
    fn = function() end,
   }
}
```

2/ From the style perspective

1. `checker` is not an accuracy variable name here as it not only checks something but also rewrites something inside. `resolver`, or `handler` (or something else) may be a good candidate.



Overall, this is just my option and won't block the merge progress.

chronolaw:
> Echoing request-review request from Harry.
> 
> 1/ From the maintainability perspective
> 
> 1. I don't see any `sort` function being called before iterating the `COMPATIBILITY_CHECKERS`, which means the program correctness is relying on the order of definition. Could we add some comments above the `compatible_checkers` var to describe that? Or could we introduce a sort mechanism?
> 2. We may add more and more properties to each checker. I feel using the array index `checker[n]` is not friendly. What do you think of using map table to replace array table for each checker?
> 
> ```lua
> local compatible_checkers = {
>   { 
>     -- desc = "resolve plugin.instance_name",
>     -- ... any other properties
>     version = x, 
>     fn = function() end,
>    }
> }
> ```
> 
> 2/ From the style perspective
> 
> 1. `checker` is not an accuracy variable name here as it not only checks something but also rewrites something inside. `resolver`, or `handler` (or something else) may be a good candidate.
> 
> Overall, this is just my option and won't block the merge progress.

It is a very valuable suggestion, thank you @vm-001.

But Applying these changes may need more time, I think we could improve it in another PR later.

liverpool8056:
Agree with @attenuation . And I think the current logic for compatibility is so complexity, and we have to write the spec very carefully for ensuring the compatibility won't be broken when schemas are changed. It should have been a chance to simplify it, cause I think it's the DP's responsibility to ensure it can work even if it receive a unexpected config, DP should have been capable to protect itself, at the same time, from the perspective of the users. it should be also capable to ensure there is no unexpected behavior for the users.

chronolaw:
Updated to #10501.

bungle (approved):

vm-001 (approved):

bungle:
> I think it's the DP's responsibility to ensure it can work even if it receives a unexpected config, most of the time it just receives some unknown fields or so. DP should have been capable to protect itself, at the same time, from the perspective of the users. DP should have been capable to ensure there is no unexpected behavior for the users as well.

I think it is CPs responsibility to send right config in a first place. The DPs should just put everything CP sends in LMDB. The DPs also already have ability to remove unknown fields on read. It would also be great if CP streams config in such a way that it could be easily be imported in LMDB, and not through all the unnecessary parsing, flattening, validation - so what you describe is what DPs already do, just that we need to remove the validation/flattening part when receiving config so that it does not error on it, of course this needs then to be backported to all supported lower versions. The current way we do this is backwards. We try to remove unknown fields on CP side while we still allow configuring things. You cannot ever solve this issue with it:

CP 1.0:
Imaginary plugin ACL-NEW has config.deny field.

DP 1.0:
Gets config from CP 1.0 and knows what to do with config.deny field.

CP upgraded to 1.1:
The ACL-NEW plugin gets new config.allow field.
Admin configures a new ACL-NEW plugin and sets config.allow

DP 1.0:
Gets config from CP 1.1, but CP removes the config.allow field from newly added ACL-NEW plugin, the ACL-NEW plugin does nothing on DP 1.0. (this cannot ever be solved with our current way of doing it)

DP 1.1:
Gets config from CP 1.1 and ACL-NEW plugin works.

kong/clustering/compat/checkers.lua
```diff
@@ -0,0 +1,179 @@
...
+            config_entity["updated_at"] = nil
+
+            updated_entities[name] = true
+            has_update = true
+          end
+        end
+      end
+
+      if has_update then
+        for name, _ in pairs(updated_entities) do
```
bungle, line 51: Can we make this an array too, so that order is always the same?

chronolaw: I think the order of these entities is meanless, we should check them all.

bungle (approved):

hbagdi:
@chronolaw Please open the corresponding EE cherry pick PR.

chronolaw:
> @chronolaw Please open the corresponding EE cherry pick PR.

Sure, but this PR need #10400 and #10501 to be cherry-picked first.

hbagdi:
@chronolaw Please help out reviewing the cherrypicks of those PRs and then work on cherry-picking this one.

outsinre:
@chronolaw @hbagdi , as not yet manually cherry-picked. This will be cherry-picked by CE2EE merge PR: https://github.com/Kong/kong-ee/pull/4985.

PR #10160, opened 2023-01-23
fix(tracing) add http.status_code, fix http.flavor

kikito:
### Summary

According to the opentracing spec, the `http.status_code` attribute is:

> Conditionally Required: If and only if one was received/sent.

Also, the `http.flavor` attribute is supposed to be a string, not a double.

Ref: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

This change sets `http.status_code` on the root span once it is available (at the beginning of the `header_filter` phase) and also modifies the `http.flavor` attribute type so they both follow the spec.

### Checklist

- <del>[ ] There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE</del>

### Full changelog

* Fix(tracing) add http.status_code attribute to root span, if available

### Issue reference

FTI-4644 / KAG-463

### DIFF
--- a/kong/tracing/instrumentation.lua
+++ b/kong/tracing/instrumentation.lua
@@ -196,6 +196,11 @@ function _M.request(ctx)
                  and ctx.KONG_PROCESSING_START * 1e6
                   or time_ns()
 
+  local http_flavor = ngx.req.http_version()
+  if type(http_flavor) == "number" then
+    http_flavor = string.format("%.1f", http_flavor)
+  end
+
   local active_span = tracer.start_span(span_name, {
     span_kind = 2, -- server
     start_time_ns = start_time,
@@ -204,7 +209,7 @@ function _M.request(ctx)
       ["http.url"] = req_uri,
       ["http.host"] = host,
       ["http.scheme"] = scheme,
-      ["http.flavor"] = ngx.req.http_version(),
+      ["http.flavor"] = http_flavor,
       ["net.peer.ip"] = client.get_ip(),
     },
   })
--- a/kong/tracing/instrumentation.lua
+++ b/kong/tracing/instrumentation.lua
@@ -170,7 +170,7 @@ function _M.http_client()
   http.request_uri = wrap
 end
 
---- Regsiter vailable_types
+--- Register available_types
 -- functions in this list will be replaced with NOOP
 -- if tracing module is NOT enabled.
 for k, _ in pairs(_M) do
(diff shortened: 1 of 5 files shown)


flrgh (approved):
Well done, looks good :+1:

kong/tracing/instrumentation.lua
```diff
@@ -204,7 +209,7 @@ function _M.request(ctx)
       ["http.url"] = req_uri,
       ["http.host"] = host,
       ["http.scheme"] = scheme,
-      ["http.flavor"] = ngx.req.http_version(),
+      ["http.flavor"] = http_flavor,
       ["net.peer.ip"] = client.get_ip(),
```
samugi, line 213: Would [http.client_ip](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions) be more appropriate than `net.peer.ip` to identify the client ip here? I am just going by the linked docs, if there is any reason why we are using this name please disregard my comment.

locao: Maybe @mayocream knows why this name was used.

kikito: Good question.

The `http.client_ip` attribute is described like this:

> The IP address of the original client behind all proxies, if known (e.g. from X-Forwarded-For)

Operative word is *original*.

What `kong.client.get_ip()` produces is explicitly *not the original*. It says so in the [docs](https://docs.konghq.com/gateway/latest/plugin-development/pdk/kong.client/#kongclientget_ip):

> This module always returns the address of the client directly connecting to Kong. That is, **in cases when a load balancer is in front of Kong, this function returns the load balancer’s address, and not that of the downstream client
**.

If we wanted to set "the original" then our best bet would be `kong.client.get_forwarded_ip()` instead:

```
["http.client_ip"] = client.get_forwarded_ip()
```

But I am not surprised that @mayocream didn't include that one because it is mired with conditionals (*balancer is one of the trusted IPs, and the load balancer adds the right headers matching with the configuration of `real_ip_header`*)

Anyway, out of the scope for this PR. I already overexceeded it by including http.flavor

kikito: But now that I look at it more closely, it seems like the attribute `net.peer.ip` isn't part of the spec ... but `net.sock.peer.addr` is. I will leave that one for the future.

PR #10161, opened 2023-01-24
feat(dbless): improve validation errors from /config

flrgh:
This is still a WIP, but the PR is open for feedback/review.


kong/api/routes/config.lua
```diff
@@ -89,6 +323,16 @@ return {
         if check_hash and err_t and err_t.error == "configuration is identical" then
           return kong.response.exit(304)
         end
+
+        if err_t["@input"] then
```
flrgh, line 327: Feedback wanted re: this approach.

Augmenting the existing schema/declarative code to support a different "shape" of error table is a _very_ big and intrusive task, so my first attempt makes a small change to the declarative code (attaching the processed user input) and handles all of the parsing after-the-fact.

Is this a sane approach, or do we need to rethink it? And if we are okay with this approach, are there any suggestions for improvement?

flrgh: update: this code was moved into `kong.db.errors:declarative_config_flattened()`, which is IMO an improvement upon it living in the API code.

samugi (commented):
looks great to me so far, I understand it is still WIP so I am leaving the review as comment for now until it's ready

flrgh:
The initial goal was to use the processed input when generating the final error array.
This means updating `kong.db.schema.others.declarative_config:flatten()` to attach/return the processed config table when returning an error. I couldn't come up with a means of doing so without making lots of breaking and/or intrusive changes, so I compromised a bit.

kong/api/routes/config.lua
```diff
@@ -24,6 +24,58 @@ local function reports_timer(premature)
...
+      or val == "true"
+      or val == "1"
+      or val == "on"
+      or val == "yes"
+end
+
+
+local function hydrate_config_from_request(params, dc)
+  if params._format_version then
+    return params
```
samugi, line 39: is it ok that we are not returning a hash in this case?

samugi: nvm, it's generated in declarative/init.lua later

kong/db/errors.lua
```diff
@@ -510,4 +516,458 @@ function _M:invalid_unique_global(name)
...
+
+  if type(input) ~= "table" then
+    error("err_t input is nil or not a table", 2)
+  end
+
+  local flattened = flatten_errors(input, err_t)
+
+  err_t = self:declarative_config(err_t)
+
+  err_t.flattened = flattened
```
flrgh, line 974: *Feedback wanted!*

Is `flattened` the attribute name that we want to stick with? I feel kinda "meh" about it myself. Keep in mind we must maintain backwards compatibility with the old response shape.

Given this response:

```json
{
  "message": "declarative config is invalid: {services={{host=\"required field missing\",port=\"value should be between 0 and 65535\"}}}",
  "code": 14,
  "fields": {
    "services": [
      {
        "host": "required field missing",
        "port": "value should be between 0 and 65535"
      }
    ]
  },
}
```

...the new "shape" will look like this:

```json
{
  "message": "declarative config is invalid: {services={{host=\"required field missing\",port=\"value should be between 0 and 65535\"}}}",
  "code": 14,
  "fields": {
    "services": [
      {
        "host": "required field missing",
        "port": "value should be between 0 and 65535"
      }
    ]
  },
  "name": "invalid declarative configuration",
  "flattened": [
    {
      "entity_type": "service",
      "entity_name": "jeff",
      "entity_tags": [
        "jeff-is-cool"
      ],
      "entity": {
        "name": "jeff",
        "protocol": "http",
        "port": -1,
        "tags": [
          "jeff-is-cool"
        ]
      },
      "errors": [
        {
          "type": "field",
          "field": "host",
          "message": "required field missing"
        },
        {
          "type": "field",
          "field": "port",
          "message": "value should be between 0 and 65535"
        }
      ]
    }
  ]
}
```

Similar options could be...
  - `errors`
  - `entities`
  - `entity_errors`
  - `flattened_errors`

flrgh:
Rebased onto master just now. I'm cherry-picking to EE to hopefully make sure there are no surprises over on that side.

kong/db/errors.lua
rainest: It was an arbitrary name I came up with without a whole lot of though, to contrast within the traditional "nested" structure of `fields`.

I'm not particularly attached to it and it's trivial to adjust to a change on the controller side, so if you have a name you prefer. go for it.

tyler-ball: Naming is hard, but we all know Jeff is cool


I prefer `flattened_errors` myself.

kong/db/declarative/init.lua
```diff
@@ -168,7 +141,46 @@ function _M:parse_string(contents, filename, old_hash)
...
     end
 
-    return nil, err, { error = err }
+    return nil, err, { error = err }, nil
+  end
+
+  -- we don't care about the strength of the hash
+  -- because declarative config is only loaded by Kong administrators,
+  -- not outside actors that could exploit it for collisions
+  local new_hash = md5(contents)
```
tyler-ball, line 150: @gszr Are we even allowed to use `md5` when running in FIPS mode? I thought it was a completely disabled algorithm

tyler-ball (approved):
A few questions but overall I think its looking great. I would like to get 1 more review's eyes on this if possible, but the test cases look good

kong/api/routes/config.lua
```diff
@@ -24,6 +24,64 @@ local function reports_timer(premature)
...
+  end
+
+  local config = params.config
+
+  if not config then
+    local body = kong.request.get_raw_body()
+    if type(body) == "string" and #body > 0 then
+      config = body
+    else
+
```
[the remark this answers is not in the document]
flrgh: Thanks--the newline was intentional but misplaced; fixing!

kong/db/errors.lua
flrgh: I think I'll go with `flattened_errors`, if only because `flattened` is a word that is used all over the place in relation to dbless configuration/schema, so this helps sets it apart ever so slightly.

kong/db/declarative/init.lua
gszr: That is correct, Tyler! `md5` is a "non-approved algorithm" in FIPS parlance. We have replaced it with sha256 in other places -- see, for example, https://github.com/Kong/kong/pull/9533.

rainest: Yeah, no reason to use MD5 for anything anymore unless you need to run on some godforsaken platform that somehow does not support modern SHA256. Unless we want to run Kong on, idk, an ancient Windows XP PC that's been sitting in a flooded basement and somehow still running since 2002, do not use MD5: 44f425b4467598449bc6d38386c15f711f6f223c

flrgh: Can we change the hash spec in a different PR? This part of the changeset is not new code--just a refactor.

samugi (approved):

PR #10164, opened 2023-01-24
fix(pdk): request.get_raw_query

### LINKED ISSUE #10080: kong.service.request.set_raw_query() doesn't work reliably

3.0+

### Current Behavior

When we tried to use the Kong API to manipulate  the incoming HTTP request query parameters, e.g. in the plugin "access" function, like below:

```
local orignal_args = kong.request.get_raw_query()
... [append, remove, modify the query parameters as a string] 
kong.service.request.set_raw_query(qparams_str)g API to modify query parameters string
```

We noticed that the set_raw_query() method doesn't work. I've found an existing bug report on the same issue, see the link below:
https://github.com/Kong/kong/issues/9492

We don't like calling ngx.var.args to achieve the same goal, as it invokes yet another API. Plus, we've found that even this API doesn't work if the original HTTP request doesn't have any query parameters. 

Our conclusion, after experimenting with both API, is that they don't work reliably for all situations. We want it to work consistently as specified below:

### Expected Behavior

1) We use Kong API to read the existing query parameters from the incoming request. It should return either nil or ""  or empty table if there is none
2) We then manipulate the existing query parameters, either as a string or a LUA table, by appending, removing or modifying query parameter
3) We call Kong API to set the query parameters

The above 3 steps should always work in the Kong plugin access phase.

### Steps To Reproduce

```markdown
1. Create a kong 3.0 custom plugin
2. In its access function, try to use Kong API to retrieve the raw query parameters
3. Try to append a new query parameter, remove an existing parameter, and modify an existing parameter
4. Call Kong API set_raw_qeury() to set the updated query parameter string
5. Observe if the service/upstream received the request with the updated query parameters
```

### Anything else?

_No response_

samugi:
### Summary

Avoid accessing `ngx.var.args` and fetch the raw query string from the request_uri instead.

Alternatives could be restricting access to happen only via ngx.var.* or implementing some index clearing logic in [lua-kong-nginx-module](https://github.com/Kong/lua-kong-nginx-module/blob/fec73310688b16da974a97e52bf27aa9c698bfaa/lualib/resty/kong/var.lua#L153-L187) by patching set_uri_args(). 

Possibly going to close this in favor of: https://github.com/Kong/lua-kong-nginx-module/pull/59

Fix #10080

[KAG-410]

[KAG-410]: https://konghq.atlassian.net/browse/KAG-410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,12 @@
 - **OAuth2**: `refresh_token_ttl` is now limited between `0` and `100000000` by schema validator. Previously numbers that are too large causes requests to fail.
   [#10068](https://github.com/Kong/kong/pull/10068)
 
+#### PDK
+
+- Fix an issue where service.request.set_raw_query fails when called after
+  request.get_raw_query
+  [#10164](https://github.com/Kong/kong/pull/10164)
+
 ### Changed
 
 #### Hybrid Mode
--- a/kong/pdk/request.lua
+++ b/kong/pdk/request.lua
@@ -454,7 +454,9 @@ local function new(self)
   function _REQUEST.get_raw_query()
     check_phase(PHASES.request)
 
-    return var.args or ""
+    local uri = var.request_uri or ""
+    local q = find(uri, "?", 2, true)
+    return q and sub(uri, q + 1, -1) or ""
   end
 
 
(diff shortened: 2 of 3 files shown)


samugi:
closing in favor of: https://github.com/Kong/lua-kong-nginx-module/pull/59

PR #10171, opened 2023-01-26
fix(key-auth): return unauthorized if api key includes symbols that are not allowed

### LINKED ISSUE #5627: key-auth plugin fails with internal server errors for utf-8 api keys
### Summary

This is a PenTest finding, that has no specific criticality, but it seems to be something, that could be fixed. If an attacker (or a wrongly configured system) sends a request to a route, that has a key-auth plugin configured, and includes a key with non-ascii symbols (like `ý`), instead of responding with 401 Unauthorized, Kong responds with 500 Internal Server Error.

### Steps To Reproduce

1. Create any service
2. Create a route /pentest-apikey
3. Add key-auth to the route, setting key names to `apikey`
4. Perform the following Request

> GET /pentest-apikey HTTP/1.1 
> Host: <yourHostHere> 
> apikey: ý

 > An unexpected error occurred

### Additional Details & Logs

- Kong version (`$ kong version`): 2.0.1
- Kong error logs:
> 2020/02/28 15:37:32 [error] 112#0: *513622 [kong] handler.lua:156 [key-auth] failed to get from node cache: [postgres] ERROR: invalid byte sequence for encoding "UTF8": 0xfd, client: 10.128.8.1, server: kong, request: "GET /api-manager HTTP/1.1", host: "gateway.hub-test.db.de"
--
  | 10.128.8.1 - - [28/Feb/2020:15:37:32 +0000] "GET /api-manager HTTP/1.1" 500 28 "-" "PostmanRuntime/7.22.0"

It would be wise to validate the input, that is parsed from the header before using it in the handler, and handling any unparseable values for the apikey as if it is an invalid key, and thus returning 401.

This issue is related to https://github.com/Kong/kong/issues/5043, and the Implementation of that issue could be extended to catch the case that is described in this issue.

Niladri24dutta:
Currently the key-auth plugin returns HTTP 500 (Internal Server Error) status code if the api key includes non-ascii symbols. This is not appropriate .This updates the key-auth plugin to send HTTP 401 (Unauthorized) status code on such cases.

* return 401 when the api key includes non-ascii symbols
* add a new test case in key-auth plugin

Fix #5627

### Summary

Currently the key-auth plugin returns HTTP 500 (Internal Server Error) status code if the api key includes non-ascii symbols. This is not appropriate .This updates the key-auth plugin to send HTTP 401 (Unauthorized) status code on such cases.

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

Fix #5627

### DIFF
--- a/kong/plugins/key-auth/handler.lua
+++ b/kong/plugins/key-auth/handler.lua
@@ -30,6 +30,7 @@ local _realm = 'Key realm="' .. _KONG._NAME .. '"'
 
 local ERR_DUPLICATE_API_KEY   = { status = 401, message = "Duplicate API key found" }
 local ERR_NO_API_KEY          = { status = 401, message = "No API key found in request" }
+local ERR_INVALID_API_KEY     = { status = 401, message = "Invalid symbol(s) found in API key" }
 local ERR_INVALID_AUTH_CRED   = { status = 401, message = "Invalid authentication credentials" }
 local ERR_INVALID_PLUGIN_CONF = { status = 500, message = "Invalid plugin configuration" }
 local ERR_UNEXPECTED          = { status = 500, message = "An unexpected error occurred" }
@@ -168,6 +169,12 @@ local function do_authentication(conf)
   if not key or key == "" then
     kong.response.set_header("WWW-Authenticate", _realm)
     return nil, ERR_NO_API_KEY
+  end
+
+  -- this request contains invalid symbol in API key, HTTP 401
+  if not key:find("^[%w%-%_]+$") then
+    kong.response.set_header("WWW-Authenticate", _realm)
+    return nil, ERR_INVALID_API_KEY
   end
 
   -- retrieve our consumer linked to this API key
--- a/spec/03-plugins/09-key-auth/02-access_spec.lua
+++ b/spec/03-plugins/09-key-auth/02-access_spec.lua
@@ -245,6 +245,19 @@ for _, strategy in helpers.each_strategy() do
         local json = cjson.decode(body)
         assert.same({ message = "No API key found in request" }, json)
       end)
+      it("returns Unauthorized on invalid symbol in key header", function()
+        local res = assert(proxy_client:send {
+          method  = "GET",
+          path    = "/status/200",
+          headers = {
+            ["Host"] = "key-auth1.com",
+            ["apikey"] = "5SRmk6gý",
+          }
+        })
+        local body = assert.res_status(401, res)
+        local json = cjson.decode(body)
+        assert.same({ message = "Invalid symbol(s) found in API key" }, json)
+      end)
       it("returns WWW-Authenticate header on missing credentials", function()
         local res = assert(proxy_client:send {
           method  = "GET",


kong/plugins/key-auth/handler.lua
```diff
@@ -170,6 +172,13 @@ local function do_authentication(conf)
     return nil, ERR_NO_API_KEY
   end
 
+  -- this request contains invalid non-ascii symbol in API key, HTTP 401
+  matchposition = key:find('[\192-\255][\128-\191]*')
```
hanshuebner, line 176: What is the actual requirement for the API key?  Are control characters and/or DEL (\127) allowed?  Why is the pattern not '[\128-\255]*'?

Niladri24dutta: @hanshuebner as per the below documentation I found that API key is alphanumeric ([A-Z][a-z][0-9]) even though i could not find any exact requirement for it. 
https://docs.konghq.com/hub/kong-inc/key-auth/

>  Are control characters and/or DEL (\127) allowed? Why is the pattern not [\128-\255]*

As for this PR , I have used regex to not allow utf-8 characters like **à,á,â,ã,ä,å** .  I agree that [\128-\255] covers all the extended characters list. Should I make the regex like this `'[\128-\255]'`

hanshuebner: I'd rather whitelist what's in the documentation ("The key names may only contain [a-z], [A-Z], [0-9], [_] underscore, and [-] hyphen.")

Niladri24dutta: @hanshuebner ok I am using this regex `^[%w%-%_]+$` to allow only ` [a-z], [A-Z], [0-9], [_] underscore, and [-] hyphen` . But do you think if this could handle the scenario if the non ascii character is at the last character of the string ?

locao: > But do you think if this could handle the scenario if the non ascii character is at the last character of the string ?

You could add a test to check that scenario. You'd make sure your solution is working and no one will mess with it in the future.

hanshuebner:
KAG-721

hanshuebner (approved):

kong/plugins/key-auth/handler.lua
```diff
@@ -170,6 +171,12 @@ local function do_authentication(conf)
     return nil, ERR_NO_API_KEY
   end
 
+  -- this request contains invalid symbol in API key, HTTP 401
+  if not key:find("^[%w%-%_]+$") then
```
sabertobihwy, line 175: `string.find` cannot be optimized by LuaJIT when doing regular matching, might there be performance issues here?

hanshuebner: Sorry, I've seen your new comment only after merging.  Do you believe that checking the syntax of the API key with some optimized function is warranted?

sabertobihwy: This line of code is on the hot path, so I think yes.

sabertobihwy: If you also agree I will try to optimize here.

hanshuebner: In my measurements, the check is taking in the order of 230 nanoseconds.  I'm not confident that putting time into optimizing this is time well spent, in particular if it makes the code less readable.

sabertobihwy: I verified 53 nanoseconds, but I found that the took time increased linearly with the size of the key.

kong/plugins/key-auth/handler.lua
hanshuebner: I misread the documentation - It is the _key name_ that must contain [a-zA-Z0-9_], but nothing is said about the key itself.  I will need to revisit.

hanshuebner:
I tried reproducing the originally reported problem but I could not.  @Niladri24dutta can you please provide a test case that demonstrates the problem in standalone form?

PR #10174, opened 2023-01-27
fix(build): passthrough rust toolchain env vars

flrgh:
This adds RUSTUP_HOME and CARGO_HOME, but there are others we may want to consider adding at some point:

* https://doc.rust-lang.org/cargo/reference/environment-variables.html
* https://rust-lang.github.io/rustup/environment-variables.html

This fixes the fatal build errors that happen when rustup/cargo are installed in a non-standard location:

```
error: rustup could not choose a version of cargo to run, because one wasn't specified explicitly, and no default is configured.
help: run 'rustup default stable' to download the latest stable release of Rust and set it as your default toolchain.
Makefile:24: *** "cargo not found in PATH, consider doing \"curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh\"".  Stop.
Target //build:kong failed to build
```

### DIFF
--- a/build/kong_bindings.bzl
+++ b/build/kong_bindings.bzl
@@ -17,6 +17,9 @@ def _load_vars(ctx):
     # Temporarily fix for https://github.com/bazelbuild/bazel/issues/14693#issuecomment-1079006291
     for key in [
         "PATH",
+        "RUSTUP_HOME",
+        "CARGO_HOME",
+
         # above should not be needed
         "GITHUB_TOKEN",
         "RPM_SIGNING_KEY_FILE",


fffonion:
@flrgh we don't want to use the system-wise rust, the bazel build of atc-router will download the toolchain
with its own https://github.com/Kong/atc-router/pull/20/files#diff-cbbd1a1262f26bce2ee4fbf3f94a652224c94d6c652c14d719c960a0d70a84cfR18, this will make sure
the build is hermetic.

The error you saw is from https://github.com/Kong/atc-router/blob/main/Makefile#L25, where we use the Makefile
to install lualibs, there's a hack we are currently doing at https://github.com/Kong/kong/blob/master/build/BUILD.bazel#L33, but we should really be making fix to the atc-router's Makefile so that 1) it only checks for cargo when it tries to build libatc_router.so, 2) we can have a seperate make target that only installs lualibs

flrgh:
> @flrgh we don't want to use the system-wise rust, the bazel build of atc-router will download the toolchain with its own https://github.com/Kong/atc-router/pull/20/files#diff-cbbd1a1262f26bce2ee4fbf3f94a652224c94d6c652c14d719c960a0d70a84cfR18, this will make sure the build is hermetic.

Gotcha. At this point I'm still very confused as to why I'm the only one who seems to get this error :shrug:

I had previously opened a PR for the router Makefile before this one, so I'm re-opening that:

https://github.com/Kong/atc-router/pull/28

sabertobihwy:
> 2) we can have a seperate make target that only installs lualibs. (And for 1 i haven't figured how to do yet)

The error I am experiencing maybe is related to this.
The short version is that I have successfully built openresty with the following command.

```
./kong-ngx-build -p ${BUILDROOT} \
  --openresty ${RESTY_VERSION} \
  --openssl ${RESTY_OPENSSL_VERSION} \
  --luarocks ${RESTY_LUAROCKS_VERSION} \
  --pcre ${RESTY_PCRE_VERSION} \
  --atc-router ${ATC_ROUTER} \
  --resty-websocket ${RESTY_WEBSOCKET} \
  --debug
```

but I found no lua module for `atc-router` under `${OPENRESTY_PREFIX}/lualib`

PR #10199, opened 2023-01-31
chore(deps) bump resty.session from 3.10 to 4.0.0

bungle:
### Summary

Bumps resty.session session from 3.10 to 4.0.0.

This is huge change as the session library was basically rewritten. The session plugin was modified accordingly.

And the clustering compatibility was added.

### DIFF
--- a/kong-3.2.0-0.rockspec
+++ b/kong-3.2.0-0.rockspec
@@ -40,7 +40,7 @@ dependencies = {
   "lua-resty-counter == 0.2.1",
   "lua-resty-ipmatcher == 0.6.1",
   "lua-resty-acme == 0.10.1",
-  "lua-resty-session == 3.10",
+  "lua-resty-session == 4.0.0",
   "lua-resty-timer-ng == 0.2.0",
 }
 build = {
--- a/kong/clustering/compat/init.lua
+++ b/kong/clustering/compat/init.lua
@@ -235,7 +235,16 @@ local function delete_at(t, key)
 end
 
 
-local function invalidate_keys_from_config(config_plugins, keys, log_suffix)
+local function rename_field(config, name_from, name_to, has_update)
+  if config[name_from] ~= nil then
+    config[name_to] = config[name_from]
+    return true
+  end
+  return has_update
+end
+
+
+local function invalidate_keys_from_config(config_plugins, keys, log_suffix, dp_version_num)
   if not config_plugins then
     return false
   end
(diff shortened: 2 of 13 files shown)


kong-3.2.0-0.rockspec
```diff
@@ -40,7 +40,7 @@ dependencies = {
   "lua-resty-counter == 0.2.1",
   "lua-resty-ipmatcher == 0.6.1",
   "lua-resty-acme == 0.10.1",
-  "lua-resty-session == 3.10",
+  "lua-resty-session == 4.0.0.beta.1",
```
bungle, line 43: I will update this to `4.0.0` if everything goes fine.

kong/plugins/session/session.lua
```diff
@@ -9,66 +9,67 @@ local ipairs = ipairs
...
 --- Open a session based on plugin config
 -- @returns resty.session session object
 function _M.open_session(conf)
-  return resty_session.open(get_opts(conf))
+  kong.log.inspect(conf.response_headers)
+
+  return resty_session.open({
+    secret                    = conf.secret,
+    audience                  = conf.audience,
+    storage                   = conf.storage == "kong" and kong_storage,
```
samugi, line 20: just checking: storage being false loads cookie storage, right?

kong/plugins/session/storage/kong.lua
```diff
@@ -10,103 +28,112 @@ local storage = {}
...
-    decode      = session.encoder.decode,
-  }, storage)
+local function load_session_from_db(key)
+  return kong.db.sessions:select_by_session_id(key)
 end
 
 
-local function load_session(id)
-  return kong.db.sessions:select_by_session_id(id)
+local function load_session_from_cache(key)
```
samugi, line 36: just a thought, since this falls back to loading from db in case of cache miss, should we call it load_session_**with**_cache?

samugi (approved):
approving with a couple of comments

kong/plugins/session/session.lua
bungle: yes `nil` and `"cookie"` are synonyms.

kong/clustering/compat/init.lua
```diff
@@ -246,8 +255,17 @@ local function invalidate_keys_from_config(config_plugins, keys, log_suffix)
     local config = t and t["config"]
     if config then
       local name = gsub(t["name"], "-", "_")
-
       if keys[name] ~= nil then
+        -- 3.2.0.0
```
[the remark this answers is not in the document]
bungle: I changed this.

kong/plugins/session/storage/kong.lua
bungle: I leace this as is, it kinda loads it from cache, if it is not found it loads it from db and puts it in cache, and then gets value from cache and returns it.

PR #10201, opened 2023-02-01
fix(*): reduce bazel memory usage on messy local state and fix bin/busted

fffonion:
### Summary

This patch remove watch of dangeling local files so that bazel memory usage can remain small.

This patch also makes `bin/busted` be able to run with `resty-cli` with args. Consider this
```
/path/to/resty -I lualib_path --nginx nginx_path bin/busted x
```
the current logic will be broken because arg will be
```
{
 [-5] = "/path/to/resty",
 [0] = "bin/busted",
 [1] = "x",
}
```

In this patch, `$0` (resty-cli path) is not read from `arg[-1]` but using `/usr/bin/env resty` instead.

### Full changelog

* [fix(busted): allow busted to be called with resty with extra args](https://github.com/Kong/kong/commit/973f2674efe6424b1a9628fe46944dcd3a49776c)
* [chore(build): reduce memory usage by removing unused filegroup](https://github.com/Kong/kong/commit/25bf4ecf6bfddcd115c16d432c7565abfd11f76e)

### Issue reference

Fix #_[issue number]_

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -3,12 +3,6 @@ load("//build/nfpm:rules.bzl", "nfpm_pkg")
 load("@bazel_skylib//lib:selects.bzl", "selects")
 
 filegroup(
-    name = "srcs",
-    srcs = glob(["**"]),
-    visibility = ["//visibility:public"],
-)
-
-filegroup(
     name = "distribution_srcs",
     srcs = glob(["distribution/**"]),
     visibility = ["//visibility:public"],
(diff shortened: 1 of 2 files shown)


BUILD.bazel
```diff
@@ -2,12 +2,6 @@ load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 load("//build/nfpm:rules.bzl", "nfpm_pkg")
 load("@bazel_skylib//lib:selects.bzl", "selects")
 
-filegroup(
-    name = "srcs",
-    srcs = glob(["**"]),
```
fffonion, line 7: not including all files (even it's not from kong), as they are not needed during the build, is helpful when reducing bazel ram

ADD-SP (approved):

PR #10211, opened 2023-02-02
style(runloop/events): remove unnecessary `set_workspace`

chronolaw:
### Summary

`set_workspace()` is useless in CE, we should remove this line.

NOTICE: in EE we MUST have this line.

### Checklist

### DIFF
--- a/kong/runloop/events.lua
+++ b/kong/runloop/events.lua
@@ -277,6 +277,8 @@ end
 
 
 local function crud_consumers_handler(data)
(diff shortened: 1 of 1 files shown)


dndx:
If this line is harmless, maybe leave it for now to reduce the discrepancy between OSS and EE.

You can add a comment signaling this is no-op in OSS.

chronolaw:
It seems that Adding a comment does not make any sense, let's keep it in the codebase.

PR #10224, opened 2023-02-03
chore(*): extend fixture `mock_upstream` to data plane (FTI-4782)

outsinre:
### Summary

This is just a test fixture. So, tests for data plane can make use of `mock_upstream`.

1. `control_plane` still uses port `15555/15556`.
2. `data_plane` uses a designated *new* port `16665/16666`. So, in `data_plane`, we set up mock upstream. 

For example, the RLA tests current we have to use `httpbin.org`, which makes tests flaky. Check <https://github.com/Kong/kong-ee/blob/master/plugins-ee/rate-limiting-advanced/spec/03-access_spec.lua#L1894>.

FTI-4782

### DIFF
--- a/spec/fixtures/custom_nginx.template
+++ b/spec/fixtures/custom_nginx.template
@@ -455,12 +455,17 @@ http {
     }
 > end -- role == "control_plane"
 
-> if role ~= "data_plane" then
     server {
         server_name mock_upstream;
 
+> if role ~= "data_plane" then
         listen 15555;
         listen 15556 ssl;
+
+> else
+        listen 16665;
+        listen 16666 ssl;
+> end -- role ~= "data_plane"
 
 > for i = 1, #ssl_cert do
         ssl_certificate     $(ssl_cert[i]);
@@ -706,7 +711,6 @@ http {
             }
         }
     }
-> end -- role ~= "data_plane"
 
     include '*.http_mock';
 
--- a/spec/helpers.lua
+++ b/spec/helpers.lua
@@ -20,6 +20,8 @@ local MOCK_UPSTREAM_PORT = 15555
 local MOCK_UPSTREAM_SSL_PORT = 15556
 local MOCK_UPSTREAM_STREAM_PORT = 15557
 local MOCK_UPSTREAM_STREAM_SSL_PORT = 15558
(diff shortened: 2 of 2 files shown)


outsinre:
Close this PR just in case both DP and CP require `mock_upstream`. Regarding DP, please use custom fixtures.

kikito (dismissed):
I'd like us to avoid merging unused code - this has caused us trouble in the past, where the unused code was merged but then we realized it needed further changes when we started using it.

Could you please do a first pass at the tests that could use this new feature (like the ratelimiting test you mentioned in the pr description - perhaps others as well?) and change the tests so that they use this?

Once that is done, I believe this pr will be a `tests` PR instead of a `chore`

outsinre:
> I'd like us to avoid merging unused code - this has caused us trouble in the past, where the unused code was merged but then we realized it needed further changes when we started using it.
> 
> Could you please do a first pass at the tests that could use this new feature (like the ratelimiting test you mentioned in the pr description - perhaps others as well?) and change the tests so that they use this?

Sure. Let me test it locally.

outsinre:
Blocks https://github.com/Kong/kong-ee/pull/4581

StarlightIbuki:
> Blocks [Kong/kong-ee#4581](https://github.com/Kong/kong-ee/pull/4581).
> 
> For example, connection to `mock_upstream_url` would result in 502 as depicted in https://github.com/Kong/kong-ee/actions/runs/4181433505/jobs/7243462564#step:9:2750

Please use this feature at #4581 and change the targeting branch of it to this branch(`kong_template_15555`), so we can verify the change.

outsinre:
> > Blocks [Kong/kong-ee#4581](https://github.com/Kong/kong-ee/pull/4581).
> > For example, connection to `mock_upstream_url` would result in 502 as depicted in https://github.com/Kong/kong-ee/actions/runs/4181433505/jobs/7243462564#step:9:2750
> 
> Please use this feature at #4581 and change the targeting branch of it to this branch(`kong_template_15555`), so we can verify the change.

The two PRs are from CE and EE respectively.

outsinre:
As there is some concern. I would firstly do this PR in EE and cherry-pick back to CE if required.

dndx (approved):

PR #10229, opened 2023-02-04
feat(ip-restriction): Add TCP support (#6679)

### LINKED ISSUE #6679: IP Restriction at TCP Level.
### Summary

merusso:
### Summary

Note: This change is based on branch `release/2.8.x` so it targets this branch. However, since it is a new feature, it would probably be added in a 2.9.x release instead. There isn't currently a `release/2.9.x` or `release/2.x` branch.

### Checklist

### Full changelog

* Implement preread function

### Issue reference

Fix #6679

### DIFF
--- a/kong/plugins/ip-restriction/handler.lua
+++ b/kong/plugins/ip-restriction/handler.lua
@@ -1,3 +1,4 @@
+local cjson = require "cjson"
 local ipmatcher = require "resty.ipmatcher"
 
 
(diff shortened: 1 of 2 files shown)


kong/plugins/ip-restriction/handler.lua
```diff
@@ -28,29 +29,63 @@ local function match_bin(list, binary_remote_addr)
...
-    return kong.response.error(403, "Cannot identify the client IP address, unix domain sockets are not supported.")
+    local status = 403
+    local message = "Cannot identify the client IP address, unix domain sockets are not supported."
+
+    do_exit(status, message, is_http)
   end
 
   local status = conf.status or 403
-  local message = conf.message or "Your IP address is not allowed"
+  local message = conf.message or string.format("IP address not allowed: %s", ngx.var.remote_addr)
```
merusso, line 61: I forgot to note that we updated the rejection message format also. This is causing tests to fail.

Will this string format change be accepted? It is technically unrelated to enabling TCP, but it helps us troubleshoot issues with rejected requests.

scrudge:
> Can you please rebase this on top of 'master' and target it for 'master'? New features always go into the mainline development branch (named master) first.

We have a separate [PR for master](https://github.com/Kong/kong/pull/10245).  The code has changed a bit between the 2 versions.

hbagdi:
In that case, please get the other PR merged first and then we will move forward with this.
We don't ship features in patch releases so I'm not sure if we can even move forward with this or not.
cc @dndx

kong/plugins/ip-restriction/handler.lua
```diff
@@ -28,29 +29,63 @@ local function match_bin(list, binary_remote_addr)
...
+    return ngx.exit()
+  end
+end
+
+
+local function handler(conf, is_http)
   local binary_remote_addr = ngx.var.binary_remote_addr
   if not binary_remote_addr then
-    return kong.response.error(403, "Cannot identify the client IP address, unix domain sockets are not supported.")
+    local status = 403
```
locao, line 54: Why does this line overwrite the `conf.status` value?

merusso: Good question. This preserves the existing behavior which also uses a hardcoded status `403` instead of `conf.status`.

merusso:
> In that case, please get the other PR merged first and then we will move forward with this.
> We don't ship features in patch releases so I'm not sure if we can even move forward with this or not.

I'm happy to target a `2.9.x` or `2.x` branch for this feature change if one becomes available.

hbagdi:
Let's first get the PR targeting the master branch merged. We can then cherry-pick it back.

merusso:
Sounds good. We'll continue working on PR #10245.

hanshuebner:
Closing as we intend to backport #10245 instead.

PR #10243, opened 2023-02-06
feat(ip-restriction): Add TCP support

### LINKED ISSUE #6679: IP Restriction at TCP Level.

scrudge:
### Summary

### Full changelog

* Implement preread function 

### Issue reference
Fix #6679


merusso:
Why was this closed?

PR #10245, opened 2023-02-06
feat(ip-restriction): Add TCP Support

### LINKED ISSUE #6679: IP Restriction at TCP Level.

scrudge:
### Summary

### Checklist

- [x] Link to docs pr: https://github.com/Kong/docs.konghq.com/pull/5736

### Full changelog
* Implement preread function

### Issue reference

Fix #6679 

KAG-722


kong/plugins/ip-restriction/handler.lua
```diff
@@ -1,3 +1,4 @@
+local cjson = require "cjson"
```
chronolaw, line 1: "cjson.safe" may be better.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +53,64 @@ local function match_bin(list, binary_remote_addr)
...
+  if is_http then
+    return kong.response.error(status, message)
+
+  else
+    local tcpsock, err = ngx.req.socket(true)
+    if err then
+      error(err)
+    end
+
+    tcpsock:send(cjson.encode({
```
chronolaw, line 66: We should localize `cjson.encode` at the module level.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +53,64 @@ local function match_bin(list, binary_remote_addr)
...
+    local message = "Cannot identify the client IP address, unix domain sockets are not supported."
+
+    do_exit(status, message, is_http)
   end
 
   local deny = conf.deny
   local allow = conf.allow
   local status = conf.status or 403
-  local message = conf.message or "Your IP address is not allowed"
+  local message = conf.message or string.format("IP address not allowed: %s", binary_remote_addr)
```
chronolaw, line 88: localize `string.format`, and `binary_remote_addr` may not be formated as `%s`.

chronolaw:
This PR is familiar with #10229, What is the relation?

kong/plugins/ip-restriction/handler.lua
scrudge: We work on the same team.  We need this feature in 2.8.x because we can't upgrade to v3 yet.

scrudge:
> 

We work on the same team.  We need this feature in 2.8.x because we can't upgrade to v3 yet.

kong/plugins/ip-restriction/handler.lua
scrudge: I'm not sure what you mean by localize.  Put the value in a local variable first?

chronolaw: like `local string_format = string.format`.

kong/plugins/ip-restriction/handler.lua
scrudge: Done

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +53,67 @@ local function match_bin(list, binary_remote_addr)
 end
 
 
-function IpRestrictionHandler:access(conf)
+local function do_exit(status, message, is_http)
```
chronolaw, line 56: I think we can remove this arg `is_http`.

In `stream` subsystem we will call `preread()`, and in `http` subsystem we will call `access(()`.

So we could do this to eliminate the runtime cost:
```
local is_http_subsystem = ngx.config.subsystem == "http"
local do_exit
if is_http_subsystem then
  do_exit = function() ... end
else
  do_exit = function() ... end
end
```

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +53,67 @@ local function match_bin(list, binary_remote_addr)
...
+  if is_http then
+    return kong.response.error(status, message)
+
+  else
+    local tcpsock, err = ngx.req.socket(true)
```
chronolaw, line 61: Same as others, localize `ngx.req.socket`

kong/plugins/ip-restriction/handler.lua
chronolaw: I think you misunderstand my meaning. The code should be:
```
local cjson_encode = cjson.encode
```

chronolaw:
Could you add some test cases for this PR?  Thanks.

scrudge:
> Could you add some test cases for this PR? Thanks.

Sure, but I'm completely lost on how testing works here.  is this the test for this plugin :
https://github.com/Kong/kong/blob/master/t/01-pdk/03-ip/01-is_trusted.t

Also, is there some docs on how testing works? Thanks.

chronolaw:
This plugin's test should be `spec/03-plugins/17-ip-restriction`.
Perhaps we do not have a proper doc for the test, but don't worry, we can implement the feature first.

scrudge:
Yeah, I found the tests right after I asked :-)

I'm actually working on the other suggested changed ATM.
I'll also try to fix the test that are broken from the change in response message.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -29,6 +31,31 @@ do
...
+
+
+local do_exit
+if is_http_subsystem then
+  do_exit = function(status, message)
+    return kong.response.error(status, message)
+  end
+else
+  do_exit = function(status, message)
+    local cjson_encode = cjson.encode
```
chronolaw, line 44: Move this line out of function

kong/plugins/ip-restriction/handler.lua
```diff
@@ -29,6 +31,31 @@ do
...
 
+local is_http_subsystem = ngx.config.subsystem == "http"
+
+
+local do_exit
+if is_http_subsystem then
+  do_exit = function(status, message)
+    return kong.response.error(status, message)
+  end
+else
```
chronolaw, line 42: blank line before `else`

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +79,45 @@ local function match_bin(list, binary_remote_addr)
...
+    local message = "Cannot identify the client IP address, unix domain sockets are not supported."
+
+    do_exit(status, message)
   end
 
   local deny = conf.deny
   local allow = conf.allow
   local status = conf.status or 403
-  local message = conf.message or "Your IP address is not allowed"
+  local default_message = string.format("IP address not allowed: %s", binary_remote_addr)
```
chronolaw, line 94: We can not output `binary_remote_addr` as `%s`, perhaps we should use the original message `Your IP address is not allowed`

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +79,45 @@ local function match_bin(list, binary_remote_addr)
 end
 
 
-function IpRestrictionHandler:access(conf)
+local function handler(conf)
```
sabertobihwy, line 82: I think change the func name `handler` to `do_restrict` is better.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -29,6 +31,31 @@ do
...
+    if err then
+      error(err)
+    end
+
+    tcpsock:send(cjson_encode({
+      status  = status,
+      message = message
+    }))
+
+    return ngx.exit()
```
sabertobihwy, line 55: 1. It is better to localize `ngx.exit`

```lua
local ngx_exit = ngx.exit

...

ngx_exit()
```

2. we should use `ngx.exit()` with a status argument, e.g. `ngx.exit(1)`

kong/plugins/ip-restriction/handler.lua
scrudge: Sorry, we had fixed this in our code and I forgot to copy it over.  I changed it to the string address variable.  We would really like to keep this feature.  It's been very useful when trouble shooting with users and identifying holes in the allow lists.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -29,6 +33,31 @@ do
...
+    if err then
+      error(err)
+    end
+
+    tcpsock:send(cjson_encode({
+      status  = status,
+      message = message
+    }))
+
+    return ngx_exit(1)
```
sabertobihwy, line 57: sorry, I misled you, maybe `ngx_exit(status)` would be better here, right? cc @dndx

kong/plugins/ip-restriction/handler.lua
```diff
@@ -29,6 +33,31 @@ do
...
+else
+  do_exit = function(status, message)
+    local tcpsock, err = ngx_req.socket(true)
+    if err then
+      error(err)
+    end
+
+    tcpsock:send(cjson_encode({
+      status  = status,
+      message = message
```
hbagdi, line 54: Why send a 'status' property for TCP/UDP connections? HTTP status is an L7 concept and we should not be sending that to a client who is using TCP/UDP.

scrudge: I understand your point.  What type of response would you recommend?

scrudge: @hbagdi Any suggestion on the response format and content?

kong/plugins/ip-restriction/handler.lua
scrudge: @chronolaw is the fix acceptable?

kong/plugins/ip-restriction/handler.lua
hbagdi: `{"message":"rate-limit exceeded"}` would make more sense.

merusso: So would the following be acceptable?

```lua
    tcpsock:send(cjson_encode({
      message = message
    }))
```

scrudge: Status removed

hanshuebner:
@scrudge Did you mean to add a test for the new functionality?  I only see changes to the existing test.  Please rebase to the `master` branch as well.  Thanks!

hanshuebner (changes requested):
Missing test for new functionality.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -29,6 +33,30 @@ do
...
+    local tcpsock, err = ngx_req.socket(true)
+    if err then
+      error(err)
+    end
+
+    tcpsock:send(cjson_encode({
+      message = message
+    }))
+
+    return ngx_exit(status)
```
bungle, line 56: Could this just use `kong.response.exit`?

```suggestion
    kong.response.exit(status, { message = message })
```

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +80,45 @@ local function match_bin(list, binary_remote_addr)
...
 
-function IpRestrictionHandler:access(conf)
+local function do_restrict(conf)
   local binary_remote_addr = ngx_var.binary_remote_addr
   if not binary_remote_addr then
-    return kong.response.error(403, "Cannot identify the client IP address, unix domain sockets are not supported.")
+    local status = 403
+    local message = "Cannot identify the client IP address, unix domain sockets are not supported."
+
+    do_exit(status, message)
```
bungle, line 89:
```suggestion
    return do_exit(status, message)
```

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +80,45 @@ local function match_bin(list, binary_remote_addr)
...
   local status = conf.status or 403
-  local message = conf.message or "Your IP address is not allowed"
+  local default_message = string.format("IP address not allowed: %s", ngx_var.remote_addr)
+  local message = conf.message or default_message
 
   if not isempty(deny) then
     local blocked = match_bin(deny, binary_remote_addr)
     if blocked then
-      return kong.response.error(status, message)
+      do_exit(status, message)
```
bungle, line 101:
```suggestion
      return do_exit(status, message)
```

kong/plugins/ip-restriction/handler.lua
```diff
@@ -52,31 +80,45 @@ local function match_bin(list, binary_remote_addr)
...
     end
   end
 
   if not isempty(allow) then
     local allowed = match_bin(allow, binary_remote_addr)
     if not allowed then
-      return kong.response.error(status, message)
+      do_exit(status, message)
```
bungle, line 108:
```suggestion
      return do_exit(status, message)
```

bungle (dismissed):
Could you please add some return statements (it makes code cleaner about its intent).

hbagdi:
@scrudge Could you please address the latest feedback?

kong/plugins/ip-restriction/handler.lua
scrudge: Can kong.response.exit(status) work with the TCP connection?

scrudge:
> Missing test for new functionality.

As I mentioned early, I have know idea how to test TCP.  Is there some example you can suggest.  Thanks.

kong/plugins/ip-restriction/handler.lua
chronolaw: Yes, see `pdk/response.lua`

spec/03-plugins/17-ip-restriction/02-access_spec.lua
```diff
@@ -248,7 +248,7 @@ for _, strategy in helpers.each_strategy() do
         })
         local body = assert.res_status(403, res)
         local json = cjson.decode(body)
-        assert.same({ message = "Your IP address is not allowed" }, json)
+        assert.is_true(string.find(json, "IP address not allowed"))
```
chronolaw, line 251: We could use `assert.matches()` with the special pattern.

hbagdi:
@chronolaw How do we test this feature? Could you please share some thoughts on this here?

chronolaw:
I think we could find some example code in `spec/02-integration/05-proxy`, such as `28-stream_plugins_triggering_spec.lua` and `19-grpc_proxy_spec.lua`.

scrudge:
Yes, thanks.

jjchambl:
Who do I need to yell at to reopen this

scrudge:
@hbagdi can we re-open this?  We are about to update with some tests.

jeffrey-arndt:
Reopening. Tests to be added soon

spec/03-plugins/17-ip-restriction/02-access_spec.lua
scrudge: Can you send a link to the docs for assert.matches()?  I can't find it.

scrudge:
Sorry, I tried applying a last minute suggestion and it broke the build.
I've rolled it back and added tests, which are all passing.  I believe this is ready for review.

kong/plugins/ip-restriction/handler.lua
hanshuebner: Seems like this is not addressed.

hanshuebner (changes requested):
One remaining comment to be addressed, otherwise good to go.  I have added a [docs PR](https://github.com/Kong/docs.konghq.com/pull/5736) as well.

kong/plugins/ip-restriction/handler.lua
scrudge: The Kong response caused an error when I tested it.
However, I'll remove the json response that should make it more efficient.

scrudge: I removed the json response.

scrudge: @hanshuebner is this acceptable?

hanshuebner: I think @bungle pointed out that `kong.response.exit` should be preferred to `ngx.exit`.  Any reason not to follow his advice?

scrudge: Yes, I tried using kong.response.exit and it failed sending the reponse.

scrudge: @hanshuebner can we close this?  I went with one of @bungle suggestions.  The other didn't work.

scrudge:
@hanshuebner @bungle 

There's one remaining comment.  I think I've resolved the issue, just waiting for feedback.
https://github.com/Kong/kong/pull/10245#discussion_r1123420637

kong/plugins/ip-restriction/handler.lua
scrudge: It's not sending the message.  I'm seeing an empty string:
https://github.com/scrudge/kong/actions/runs/5391318477/jobs/9787991403#step:9:1342

hbagdi:
Please add an entry to CHANGELOG.md.

kong/plugins/ip-restriction/handler.lua
```diff
@@ -29,6 +31,28 @@ do
...
+  end
+
+else
+  do_exit = function(status, message)
+    local tcpsock, err = ngx_req.socket(true)
+    if err then
+      error(err)
+    end
+
+    tcpsock:send(message)
```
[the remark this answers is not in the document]
scrudge: We use that for debugging, But I can see issues with protocols like ssh.
Getting rid of the message would also break all of the tests we added.
@hbagdi how would you like us to proceed?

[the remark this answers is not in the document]
hbagdi: @chronolaw @hanshuebner Opinions?
Why not only close the socket and do nothing more?
(please hold until we arrive at a solution)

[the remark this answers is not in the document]
hanshuebner: @hbagdi I agree that sending a message through the TCP channel is unhygienic.  Furthermore, one could argue that from a security perspective, it may be undesirable to reveal the fact that there is a black- or whitelist that causes the closure of the connection.  From a usability perspective, however, I feel that sending a message before closing the connection is very helpful in debugging.  Making the message sending configurable would be an option that'd also help with testability.  Another option will be to change the tests so that instead of looking for the in-band rejection message, they will look at the log file.

Personally, I'll be fine keeping things how they are.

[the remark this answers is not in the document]
scrudge: I vote for keeping the message, but am happy to do whatever.  @hbagdi who is going to make the final call?  Please let me know how to proceed.  Would like like to wrap this up, we've been working on this for quite a while and are currently having to use a custom build for now.  Thanks.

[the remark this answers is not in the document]
hbagdi: Mixing up strings on an arbitrary TCP connection doesn't seem the right thing to do unless someone can prove enough value for doing the wrong thing.

@hanshuebner Could you please check if Kong does something similar on a TCP connection either in the core or plugin? I don't recall anything like that.
If this is a net new thing, let's remove the message and merge this. If a message is needed, we will do an opt-in flag on the plugin configuration for it.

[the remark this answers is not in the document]
hanshuebner: I could not find any evidence that we do such in-band error signaling on TCP sockets at this point.

@scrudge Can you remove the message and update your tests so that they look at specific log entries instead?  This is done in several other tests as well, here's an example: https://github.com/Kong/kong/blob/96eb92f2e301b4b249eae710d51715931a683656/spec/02-integration/05-proxy/06-ssl_spec.lua#L270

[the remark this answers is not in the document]
scrudge: I refactored to use kong respnse error (which does not return the message on TCP connections.  Also added a log message at level warn when a connection is denied.  Updated the test to use the log message to validate the deny.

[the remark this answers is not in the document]
scrudge: @hanshuebner @hbagdi 
Do you mind reviewing my changes?
Thanks!

hanshuebner (approved):

scrudge:
> Could you please add some return statements (it makes code cleaner about its intent).

@bungle can you please approve changes you requested?  Thank you.

merusso (approved):

scrudge:
> > Could you please add some return statements (it makes code cleaner about its intent).
> 
> @bungle can you please approve changes you requested? Thank you.

@hanshuebner @hbagdi 
I believe this PR is held waiting for a response from @bungle but I haven't heard from him since his original comment.  Is there anyway to get this completed?
Thanks for all the help.

hanshuebner:
@scrudge Thank your for your contribution!

scrudge:
Thank you for all your help.

PR #10252, opened 2023-02-07
fix(*): prevent queues from growing without bounds (#10046)

hanshuebner:
This commit implements an upper limit on the number of batches that may be waiting on a queue for processing.  Once the limit has been reached, the oldest batch is dropped from the queue and an error message is logged.  The maximum number of batches that can be waiting on a queue is configured through the max_queued_batches parameter of the queue, which defaults to 100 and can be globally overriden with the max_queued_batches parameter in kong.conf

KAG-303

This reverts commit 218cc0a81500986b0b94e6b4e350302326097018.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,10 @@ under certain circumstances.
 
 - Bumped `atc-router` to `1.0.2` to address the potential worker crash issue.
   [#9927](https://github.com/Kong/kong/pull/9927)
+- Update the batch queues module so that queues no longer grow without bounds if
+  their consumers fail to process the entries.  Instead, old batches are now dropped
+  and an error is logged.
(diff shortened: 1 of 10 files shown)


kikito (approved):

seh:
When can we expect a Kong release including this patch? We believe that it's causing unbounded memory use in our Kong proxy containers.

hanshuebner:
@seh We can't give you a firm date, but Kong Gateway 3.2.0 is in the final stages of our release process and will be in GA in a couple of weeks.

seh:
I asked here because this back-part targets the 3.1 minor version range. Will there be patch release for that version soon?

hanshuebner:
We are not planning to release a minor release of 3.1.  If you need the fix on 3.1, you will need to create your own images.

PR #10256, opened 2023-02-07
fix(dbless): drain error table when flattening errors

flrgh:
This only affect `POST /config?flatten_errors=1`.

When nested errors are flattened for the response body, they should be removed from the original error table so that they are not included in the final `fields` error object. Errors that are not flattened should remain.

It could be argued that this is a backwards-incompatible change, but:

1. This is opt-in only via the `flatten_errors` query param.
2. Because `flatten_errors` is a new, yet unreleased feature, there is no defined behavior to be broken.
3. It's arguably a behavioral fix since we are effectively de-duplicating the errors in the response.
4. This makes the response object much, much cleaner.
5. The overall "shape" of the `fields` object has not changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,9 +242,11 @@
 
 #### Admin API
 
-- In dbless mode, `/config` API endpoint can now flatten all schema validation
-  errors to a single array via the optional `flatten_errors` query parameter.
+- In dbless mode, `/config` API endpoint can now flatten entity-related schema
+  validation errors to a single array via the optional `flatten_errors` query
+  parameter. Non-entity errors remain unchanged in this mode.
   [#10161](https://github.com/Kong/kong/pull/10161)
+  [#10256](https://github.com/Kong/kong/pull/10256)
 
 #### PDK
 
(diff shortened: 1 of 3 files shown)


flrgh:
@RobSerafini This is a fix, but it's not really a blocker for anyone, so feel free to remove from the 3.2 milestone if we're too strapped for time.

rainest (approved):
No concern from our side. We aren't using `fields` at all (with or without the availability of flattened errors--we do use `messages` for logging though), and I agree that if you've opted into flattened that you shouldn't need the old format also.

RobSerafini:
@tyler-ball - I leave it to you to decide the milestone question.

flrgh:
This will need to wait; removed from milestone.

flrgh:
Converting this back to a draft because I will need to make a changelog edit or two before it can be merged.

flrgh:
Changelog entry was added, and I've just rebased on master. This is good to go once CI is :green_circle:.

hbagdi:
Does this need an explicit cherry pick from CE to EE?

hbagdi:
@flrgh @rainest  Please cherry-pick this PR to EE.

PR #10257, opened 2023-02-08
fix(conf_loader): propagate values for aliased/deprecated properties

flrgh:
## summary

This updates the behavior for aliased and deprecated properties.

### aliases

> Property `X` is an alias of property `Y` (`X.alias.replacement == "Y"`).

A) the user sets `X` but not `Y`: set `Y = X`
B) the user does not set `X`: set `X = <default>`, `Y = X`
C) the user sets both `X` and `Y`: `warn()`

A and B are skipped when a custom `alias.alias()` handler is registered for `X`.

### deprecated properties

> Property `X` is deprecated and superseded by `Y` (`X.deprecated.replacement == "Y"`).

A) the user sets `X` but not `Y`: set `Y = X`, `warn()`
B) the user does not `X`: set `X = Y` (for backwards compatibility)
C) the user sets both `X` and `Y`: set `X = Y`, `warn()`

A and C are skipped when a custom `deprecated.alias()` handler is registered for `X`.

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -155,6 +155,7 @@ local HEADER_KEY_TO_NAME = {
 
 
 local EMPTY = {}
+local NONE = "NONE"
 
 
 -- NOTE! Prefixes should always follow `nginx_[a-z]+_`.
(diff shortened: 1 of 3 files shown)


flrgh:
I'm working on some refinements to this, so don't merge yet.

flrgh:
This is now way too big of a change for what is a last minute bugfix :cry: I will save this for later and re-push with a more simple changeset.

PR #10263, opened 2023-02-08
fix(conf_loader): propagate renamed tracing conf values

flrgh:
This is the fix from #10257, but isolated to the two properties that were renamed.

## changelog

* clear default values for `tracing_sampling_rate` and `tracing_instrumentations`*
* for each property, if no user value is set for `$new`, set `$new = $old || default($old)`

\* we will set defaults for these fields when it is time to remove the old, deprecated ones

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -559,6 +559,20 @@ local CONF_PARSERS = {
 
   tracing_instrumentations = {
     typ = "array",
+    alias = {
+      replacement = "opentelemetry_tracing",
+      alias = function(conf)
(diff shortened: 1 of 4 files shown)


flrgh:
The tests in `spec/02-integration/14-tracing/01-instrumentations_spec.lua` are failing for me locally, but they are also failing on `master`, so... :crossed_fingers:

gszr (approved):

flrgh:
Please do not merge this for the time being. I will revisit this soon to decide if it's still needed.

flrgh:
Closing for now. We can revisit this if any users run into problems with the current behavior.

PR #10274, opened 2023-02-09
fix(gha): align docker platform output name

curiositycasualty:
### Summary

A fix I already applied to EE: https://github.com/Kong/kong-ee/blob/master/.github/workflows/release.yml#L349-L366
Needed in CE to **actually** build arm64 images.


saisatishkarra (approved):
gtm

PR #10293, opened 2023-02-13
fix(db/lmdb): hash the cache_key that exceeds LMDB max_key_size

### LINKED ISSUE #10219: Problems with keys greater than 511 in declarative configuration

### Kong version 3.1.1

kong 3.1.1

### Current Behavior

we are upgrading Kong from 2.x to Kong 3.x, and we are facing a blocking bug :cry:
We use Kong Databaseless, and with the new LMDB mechanisms from Kong 3.x, our configurations are not loading.

In the tests in declarative config file the length of a consumer's API Key is causing it (but other places may also have the same problem)

When the value has more than 511 characters Kong starts the process, but with empty configurations, also Kong reports:
2023/02/02 14:21:35 [crit] 1129#0: *8 [lua] init.lua:697: init_worker(): worker initialization error: failed to load declarative config file: MDB_BAD_VALSIZE: Unsupported size of key/DB name/data, or wrong DUPFIXED size; this node must be restarted, context: init_worker_by_lua*

### Expected Behavior

kong in version 3.1.1 can load declarative configuration without errors

### Steps To Reproduce

1. Using the image kong:3.1.1
2. Load this configuration file bellow: (following the steps in the image [readme](https://hub.docker.com/_/kong))
```yaml
---
_format_version: "1.1"

consumers:
  - username: consumer-with-credential-bigger-than-512-chars
    keyauth_credentials:
      - key: 6y8fivuu6jj822b77hcuxp6z33pxgie3t5ypx8mwuxne59gp3yvgt62m2vazuu8b3ahyjhzdmri8bk768qpdp745ijt362f3z5nnwgn785iakpwia5pegwpxki4k8xjzrac2nbg2ff3adrbbgmw7ihj8mj4itwmjaxkkrany5qnb7za46imhc9vwqmg4vpqf5vkza5dfgkr8fphpfrk6bcgz2mxuqgvcvadjb974jpjuctfjptw4j5izb68pywg9a7h85wfdddxh7j7j9yd2cxtvyk7n4cggibec9qtjhci8rdhuwcf2ixtfk4vm6hzj5j6ece9xwuxicdtrizwgm6mrx99av4ukw4c9qp7yr23ij9tutph8gek698765ydjrmgkm88pyzuq7jabpeqe7ae9j8qiew2d9rjdmmd27nmtkdgajrg78347dmrbgvxzjw3f4ewakckxnia3izv7itqa9bufnjn7rd9mwkvzfmuba4hrwk49f2ktzw9npw94

services:
  - name: httpbin
    url: https://httpbin.org/anything

routes:
  - name: httpbin
    service: httpbin
    paths:
      - /
    plugins:
      - name: key-auth
        config:
          key_names:
            - X-API-Key
```
3. see in the log the following error in loading the configuration:
```
2023/02/02 16:40:15 [error] 1127#0: *889 [kong] config.lua:124 failed loading declarative config into cache: MDB_BAD_VALSIZE: Unsupported size of key/DB name/data, or wrong DUPFIXED size, client: 172.17.0.1, server: kong_admin, request: "POST /config HTTP/1.1", host: "localhost:8001"

```

### Anything else?

_No response_

sabertobihwy:
### Summary

Fix bug in #10219 to use sha256 to generate fixed size cache_key.

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

Fix #10219

### DIFF
--- a/kong/db/declarative/import.lua
+++ b/kong/db/declarative/import.lua
@@ -29,6 +29,7 @@ local sha256 = utils.sha256_hex
 
 local DECLARATIVE_HASH_KEY = constants.DECLARATIVE_HASH_KEY
 local DECLARATIVE_EMPTY_CONFIG_HASH = constants.DECLARATIVE_EMPTY_CONFIG_HASH
+local LMDB_MAXKEYSIZE = 511 -- LMDB default max key size
 
 
 local function find_or_create_current_workspace(name)
@@ -294,7 +295,12 @@ local function load_into_cache(entities, meta, hash)
 
       if schema.cache_key then
         local cache_key = dao:cache_key(item)
-        t:set(cache_key, item_marshalled)
+        if #cache_key > LMDB_MAXKEYSIZE then
+          local hash_cache_key = sha256(cache_key)
+          t:set(hash_cache_key, item_marshalled)
+        else
+          t:set(cache_key, item_marshalled)
+        end
       end
 
       for i = 1, #uniques do


kong/db/declarative/import.lua
```diff
@@ -294,7 +295,12 @@ local function load_into_cache(entities, meta, hash)
 
       if schema.cache_key then
         local cache_key = dao:cache_key(item)
-        t:set(cache_key, item_marshalled)
+        if #cache_key > LMDB_MAXKEYSIZE then
+          local hash_cache_key = sha256(cache_key)
```
chronolaw, line 299: It may not be the correct place.
If we change the cache key here, we will not find it with the original cache_key.

should we take a look at `dao:cache_key`

chronolaw:
We need a test case to verify this PR.

kong/db/declarative/import.lua
bungle: Yes, we cannot just start hashing cache keys in a single place. Either it has to be everywhere (which is too risky), or you need to create a custom dao for key_auth (https://github.com/Kong/kong/blob/master/kong/plugins/key-auth/daos.lua) -> a `dao` property. see: https://github.com/Kong/kong/blob/master/kong/plugins/oauth2/daos.lua#L92 and https://github.com/Kong/kong/blob/master/kong/plugins/oauth2/daos/oauth2_tokens.lua#L8-L10.

sabertobihwy:
> We also need a test case to verify this PR.

Add it later

kong/db/declarative/import.lua
sabertobihwy: > Either it has to be everywhere (which is too risky), or you need to create a custom dao for key_auth

It looks like we need to do the length check at the only place where the cache_key is generated.

When loading the configuration into the cache (lmdb), the cache_key needs to be generated, where the key is directly the table, and the content:

```json
{
  consumer = {
    id = "075415e8-e1cf-546e-8f8c-b7e15251e3d7"
  },
  created_at = 1676408487,
  id = "3d9ec867-3822-5729-adbf-775e972cbb74",
  key = "6y8fivuu6jj822b77hcuxp6z33pxgie3t5ypx8mwuxne59gp3yvgt62m2vazuu8b3ahyjhzdmri8bk768qpdp745ijt362f3z5nnwgn785iakpwia5pegwpxki4k8xjzrac2nbg2ff3adrbbgmw7ihj8mj4itwmjaxkkrany5qnb7za46imhc9vwqmg4vpqf5vkza5dfgkr8fphpfrk6bcgz2mxuqgvcvadjb974jpjuctfjptw4j5izb68pywg9a7h85wfdddxh7j7j9yd2cxtvyk7n4cggibec9qtjhci8rdhuwcf2ixtfk4vm6hzj5j6ece9xwuxicdtrizwgm6mrx99av4ukw4c9qp7yr23ij9tutph8gek698765ydjrmgkm88pyzuq7jabpeqe7ae9j8qiew2d9rjdmmd27nmtkdgajrg78347dmrbgvxzjw3f4ewakckxnia3izv7itqa9bufnjn7rd9mwkvzfmuba4hrwk49f2ktzw9npw94",
  ws_id = "0dc6f45b-8f8d-40d2-a504-473544ee190b"
}
```

stack traceback:
```
	./kong/db/dao/init.lua:1539: in function 'cache_key'
	./kong/db/declarative/import.lua:300: in function 'load_into_cache'
	./kong/init.lua:516: in function <./kong/init.lua:511>
	[C]: in function 'pcall'
	./kong/concurrency.lua:54: in function 'with_worker_mutex'
	./kong/init.lua:511: in function 'load_declarative_config'
	./kong/init.lua:763: in function 'init_worker'
	init_worker_by_lua:2: in main chunk, context: init_worker_by_lua*
```

When querying the cache, we also need to generate the cache_key, where the key is a string, 

stack traceback:
```
	./kong/db/dao/init.lua:1498: in function 'cache_key'
	./kong/plugins/key-auth/handler.lua:177: in function 'do_authentication'
	./kong/plugins/key-auth/handler.lua:227: in function <./kong/plugins/key-auth/handler.lua:215>
```

This seems like a good choice for the `DAO:cache_key` function to avoid making changes to potentially long cache_key that are distributed across different plugins.

kong/db/dao/init.lua
```diff
@@ -22,6 +22,8 @@ local log          = ngx.log
 local fmt          = string.format
 local match        = string.match
 local run_hook     = hooks.run_hook
+local sha256       = utils.sha256_hex
+local LMDB_MAX_KEY_SIZE = 511 -- LMDB default max key size
```
chronolaw, line 26: `sha256` and `LMDB_MAX_KEY_SIZE` are only used by `hash_by_length`,
so we could move them near `hash_by_length` and into a `do ... end` block.

kong/db/dao/init.lua
```diff
@@ -1481,6 +1483,15 @@ local function get_cache_key_value(name, key, fields)
 end
 
 
+local function hash_by_length(str)
+  if str and #str > LMDB_MAX_KEY_SIZE then
+    return sha256(str)
+  else
```
chronolaw, line 1489: In Kong's code we should add a blank line before `else`.

kong/db/dao/init.lua
chronolaw: 511 may not be a proper number, cache_key is "key:ws_Id", a key with 511 bytes may still generate a long key.

Perhaps we should change it to `256`?

kong/db/dao/init.lua
sabertobihwy: ok

kong/db/dao/init.lua
sabertobihwy: ok

kong/db/dao/init.lua
```diff
@@ -1494,6 +1509,9 @@ function DAO:cache_key(key, arg2, arg3, arg4, arg5, ws_id)
   -- the generic code below, but building the cache key
   -- becomes a single string.format operation
   if type(key) == "string" then
+
+    key = hash_by_length(key)
```
chronolaw, line 1513: `hash_by_length` may not be a good name, could it be `normalize_string` or else?

sabertobihwy: replace

kong/db/dao/init.lua
```diff
@@ -1481,6 +1481,21 @@ local function get_cache_key_value(name, key, fields)
...
 
+local hash_by_length
+do
+  local sha256       = utils.sha256_hex
+  local MAX_KEY_SIZE = 256
+  hash_by_length = function (str)
+    if str and #str > MAX_KEY_SIZE then
+      return sha256(str)
+
+    else
```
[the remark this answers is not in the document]
sabertobihwy: ok

kong/db/dao/init.lua
```diff
@@ -1481,6 +1481,21 @@ local function get_cache_key_value(name, key, fields)
 end
 
 
+local normalize_string
+do
+  local sha256       = utils.sha256_hex
+  local MAX_KEY_SIZE = 256
+
+  normalize_string = function (str)
```
chronolaw, line 1489: minor style: remove space between `function` and `(str)`

chronolaw:
We should add a change log entry for this PR.

sabertobihwy:
> We should add a change log entry for this PR.

added

spec/02-integration/11-dbless/03-config_persistence_spec.lua
```diff
@@ -152,3 +152,62 @@ describe("dbless persistence with a declarative config #off", function()
...
+        plugins:
+        - name: key-auth
+        routes:
+        - name: my-route
+          paths:
+          - /
+      consumers:
+      - username: credential-bigger-than-512-chars
+        keyauth_credentials:
+        - key: 6y8fivuu6jj822b77hcuxp6z33pxgie3t5ypx8mwuxne59gp3yvgt62m2vazuu8b3ahyjhzdmri8bk768qpdp745ijt362f3z5nnwgn785iakpwia5pegwpxki4k8xjzrac2nbg2ff3adrbbgmw7ihj8mj4itwmjaxkkrany5qnb7za46imhc9vwqmg4vpqf5vkza5dfgkr8fphpfrk6bcgz2mxuqgvcvadjb974jpjuctfjptw4j5izb68pywg9a7h85wfdddxh7j7j9yd2cxtvyk7n4cggibec9qtjhci8rdhuwcf2ixtfk4vm6hzj5j6ece9xwuxicdtrizwgm6mrx99av4ukw4c9qp7yr23ij9tutph8gek698765ydjrmgkm88pyzuq7jabpeqe7ae9j8qiew2d9rjdmmd27nmtkdgajrg78347dmrbgvxzjw3f4ewakckxnia3izv7itqa9bufnjn7rd9mwkvzfmuba4hrwk49f2ktzw9npw94
```
chronolaw, line 189: Discussion.
Perhaps we should not use the real data int the test case, could we change it to `string.rep` or others?

sabertobihwy: updated

StarlightIbuki:
Theoretically, should we consider the possibility of collisions?

kong/db/dao/init.lua
```diff
@@ -1528,6 +1546,9 @@ function DAO:cache_key(key, arg2, arg3, arg4, arg5, ws_id)
       if value ~= nil then
         use_pk = false
       end
+
+      value = normalize_string(value)
```
StarlightIbuki, line 1550: If we have multiple sources of `cache_key`, they will be concatenated with ":", which then may still break the limit:

https://github.com/Kong/kong/blob/13c7c6f8336075575d44ac297019f7e35bbc5145/kong/db/dao/init.lua#L1551

Should we instead do the hash at the very end? Or with more complex logic, to hash more parts of the keys until the total length is within the limit?

sabertobihwy: This is a bit of an aggressive approach, and some functionality that was matched via cache_key prefix would be lost if we do it this way. What do you think? cc @chronolaw @dndx

sabertobihwy: > If we have multiple sources of `cache_key`, they will be concatenated with ":",

This is the point where I am not sure if such a case exists, and if so, a test case needs to be added to cover it.

chronolaw: As I know, currently we do not have this case, I think the code of @sabertobihwy is OK.

chronolaw:
> Theoretically, should we consider the possibility of collisions?

sha256 is good enough, it is almost impossible to find a collision.

kong/db/dao/init.lua
```diff
@@ -1494,6 +1509,9 @@ function DAO:cache_key(key, arg2, arg3, arg4, arg5, ws_id)
   -- the generic code below, but building the cache key
   -- becomes a single string.format operation
   if type(key) == "string" then
+
+    key = normalize_string(key)
```
dndx, line 1513: Should we do this inside the LMDB specific codepath instead? Doing this for non-LMDB codepath seems to make little sense.

sabertobihwy: use `self.db.strategy == off` to determine if  LMDB specific codepath

hbagdi: Wouldn't keeping the behavior consistent across all code paths be better here?
Datong, is there a specific concern here?

chronolaw: I also agree with @hbagdi , we should have the same code path.

chronolaw:
Due to PR #10038, This PR's change log entry should move to the `unreleased` section.

sabertobihwy:
> Due to PR #10038, This PR's change log entry should move to the `unreleased` section.

move

CHANGELOG.md
```diff
@@ -73,6 +73,8 @@
 
 - Bumped lua-resty-session from 4.0.2 to 4.0.3
   [#10338](https://github.com/Kong/kong/pull/10338)
+- Generate fixed size keys using hashing for keys that exceed the max key size restriction in LMDB.
```
chronolaw, line 76: This entry should in section `Fixes - Core`

kong/db/dao/init.lua
chronolaw: @dndx, could you confirm this change? I think the condition `strategy == "off"` is unnecessary.

CHANGELOG.md
sabertobihwy: move

kong/db/dao/init.lua
dndx: No specific concerns, this can be left as is. Thanks for the comments.

hbagdi:
Please rebase so we can proceed with merge.

kong/db/dao/init.lua
sabertobihwy: so I need to remove `strategy == "off"`? cc @dndx

chronolaw: I talked with dndx, I think we should remove `strategy == "off"`, which will make code more clear.

hbagdi:
@chronolaw Could you please help review this one?

chronolaw (approved):
thank you, @sabertobihwy

kong/db/dao/init.lua
```diff
@@ -1482,6 +1482,22 @@ local function get_cache_key_value(name, key, fields)
 end
 
 
+local normalize_string
+do
+  local sha256       = utils.sha256_hex
+  local MAX_KEY_SIZE = 256
```
dndx, line 1488: Should we use a higher limit instead? We can safely go way over 256 and this can reduce the number of hashing we need to do.

kong/db/dao/init.lua
dndx: I had second thought and it still seems that hashing the `cache_key` makes little sense, and this could make debugging shdict related caching issue harder. Is there a reason why we do **not** do this in place where we actually set keys inside LMDB?

The argument "Wouldn't keeping the behavior consistent across all code paths be better here?" would not apply, because I feel the `cache_key` function is simply the wrong place to do it.

dndx (changes requested):

kong/db/dao/init.lua
chronolaw: Do we have any suggestion? `384` or `448`?

chronolaw: As dndx metioned, we should use number as big as we can, so I think 448 is a proper number?

512 > 448 = 256 + 128 + 64

sabertobihwy: We need to reserve enough length for the rest of the components of the cache key, like `arg2` ~ `arg5`, `ws_id`

kong/db/dao/init.lua
chronolaw: No, we need to query LMDB by key, only changing set is not enough.

kong/db/dao/init.lua
sabertobihwy: adjust `MAX_KEY_SIZE` to `448`

sabertobihwy:
cc @dndx need your confirm again

chronolaw:
After a short discussion, we decide to use another solution for this issue. The fix is in lua-rest-lmdb(https://github.com/Kong/lua-resty-lmdb/pull/27).

@sabertobihwy could you take a look at it? if you think it is ok could you close this PR?

sabertobihwy:
get it

PR #10311, opened 2023-02-16
chore(build): install wasmx lua libs

flrgh:
This should unblock the rest of our integration work.

Please share any bazel feedback because I'm still very bad at it :) Just know that the wasmx build stuff is still very very WIP, so warts here will likely be fixed at a later date (before we merge into master).


build/BUILD.bazel
```diff
@@ -36,6 +36,25 @@ install_lualib_deps_cmd = "\n".join([
...
+    "//conditions:default": [],
+})
+
+wasmx_lua_libs_cmd = select({
+    "@kong//:wasmx_flag": """
+    for fname in $(locations @ngx_wasm_module//:lua_libs); do
+        base=${fname##*/ngx_wasm_module/lib/}
+        dest="${BUILD_DESTDIR}/openresty/lualib/$base"
+        mkdir -p "$(dirname "$dest")"
+        cp -P --preserve=all "$fname" "$dest"
```
fffonion, line 52: this _might_ not work under macos 🤔

flrgh: hmm, good call; I think we can probably get away with dropping the flags for the time being

locao (approved):

PR #10315, opened 2023-02-16
chore(bazel) remove unneeded nginx modules

bungle:
### Summary

This commit removes couple of modules that OpenResty builds by default that Kong doesn't use for anything:

- https://github.com/openresty/xss-nginx-module
- https://github.com/FRiCKLE/ngx_coolkit
- https://github.com/openresty/set-misc-nginx-module
- https://github.com/calio/form-input-nginx-module
- https://github.com/openresty/srcache-nginx-module
- https://github.com/openresty/lua-upstream-nginx-module
- https://github.com/openresty/array-var-nginx-module
- https://github.com/openresty/memc-nginx-module
- https://github.com/openresty/redis2-nginx-module
- https://github.com/openresty/redis-nginx-module
- https://github.com/openresty/rds-json-nginx-module
- https://github.com/openresty/rds-csv-nginx-module
- https://github.com/vision5/ngx_devel_kit


fffonion:
i would assume most of them are indeed not needed, but bit worried if someone comes up that they are using it. so intead of --without, what about make them "dynamic modules"?
so that we don't turn it on by default, and people can still choose flip on if they need.
we need a patch though

```patch
diff --git a/configure b/configure
index d461294..649e7eb 100755
--- a/configure
+++ b/configure
@@ -69,27 +69,27 @@ if ($OS =~ /solaris|sunos/i) {
 
 my @modules = (
     [ndk => 'ngx_devel_kit'],
-    [http_iconv => 'iconv-nginx-module', 'disabled'],
-    [http_echo => 'echo-nginx-module'],
-    [http_xss => 'xss-nginx-module'],
+    [http_iconv => 'iconv-nginx-module', 'dynamic'],
+    [http_echo => 'echo-nginx-module', 'dynamic'],
+    [http_xss => 'xss-nginx-module', 'dynamic'],
     [http_coolkit => 'ngx_coolkit'],
-    [http_set_misc => 'set-misc-nginx-module'],
-    [http_form_input => 'form-input-nginx-module'],
-    [http_encrypted_session => 'encrypted-session-nginx-module'],
+    [http_set_misc => 'set-misc-nginx-module', 'dynamic'],
+    [http_form_input => 'form-input-nginx-module', 'dynamic'],
+    [http_encrypted_session => 'encrypted-session-nginx-module', 'dynamic'],
     [http_drizzle => 'drizzle-nginx-module', 'disabled'],
     [http_postgres => 'ngx_postgres', 'disabled'],
-    [http_srcache => 'srcache-nginx-module'],
+    [http_srcache => 'srcache-nginx-module', 'dynamic'],
     [http_lua => 'ngx_lua'],
-    [http_lua_upstream => 'ngx_lua_upstream'],
-    [http_headers_more => 'headers-more-nginx-module'],
-    [http_array_var => 'array-var-nginx-module'],
-    [http_memc => 'memc-nginx-module'],
-    [http_redis2 => 'redis2-nginx-module'],
-    [http_redis => 'redis-nginx-module'],
+    [http_lua_upstream => 'ngx_lua_upstream', 'dynamic'],
+    [http_headers_more => 'headers-more-nginx-module', 'dynamic'],
+    [http_array_var => 'array-var-nginx-module', 'dynamic'],
+    [http_memc => 'memc-nginx-module', 'dynamic'],
+    [http_redis2 => 'redis2-nginx-module', 'dynamic'],
+    [http_redis => 'redis-nginx-module', 'dynamic'],
     #[http_upstream_keepalive => 'upstream-keepalive-nginx-module'],
     #[http_auth_request => 'auth-request-nginx-module'],
-    [http_rds_json => 'rds-json-nginx-module'],
-    [http_rds_csv => 'rds-csv-nginx-module'],
+    [http_rds_json => 'rds-json-nginx-module', 'disabled'],
+    [http_rds_csv => 'rds-csv-nginx-module', 'disabled'],
     [stream_lua => 'ngx_stream_lua'],
 );
 
@@ -1224,7 +1224,11 @@ _EOC_
 
         my $dir = auto_complete "../$prefix";
 
-        $opts_line .= " \\\n  --add-module=$dir";
+        if ($attr eq 'dynamic') {
+            $opts_line .= " \\\n  --add-dynamic-module=$dir";
+        } else {
+            $opts_line .= " \\\n  --add-module=$dir";
+        }
     }
 
     return $opts_line;
```

bungle:
>  worried if someone comes up that they are using it. so intead of --without, what about make them "dynamic modules"?

Yes, that is indeed safer. Let's see what others think about, if someone opposes this altogether.

bungle:
I added back:
- echo-nginx-module
- headers-more-nginx-module
- ngx_http_stub_status_module

Because of test dependency.

dndx:
@fffonion I think building them as dynamic modules does not have a huge amount of advantages as these modules are fairly small. If we don't build them at all at least we can argue this reduces the size of the package we build. With dynamic module these benefits also doesn't exist. Having these moduled built but not enabled also has no impact over performance. So I would like to understand the reason of doing this change first.

bungle:
> So I would like to understand the reason of doing this change first.

Many of those modules came to OpenResty prior almost everything moved to Lua. Some of those modules we already have removed such as https://github.com/openresty/encrypted-session-nginx-module. Most of them are unmaintained or get very little maintenance. In some ways they may widen attack surface, may contain code that is caught by security analyzers etc. (And because "Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.") ;-)

I would love to ask this otherway around. Why ship these? I know that it keeps compatibility with vanilla OpenResty, some might use Kong with those modules, but other than that, they feel like bloat.

hanshuebner:
We should not ship stuff that we or our paying customers don't need or use because we don't want to waste time on fixing issues if someone starts using any of these modules and finds issues with it.  Let's just go ahead.

hanshuebner (approved):

PR #10317, opened 2023-02-17
feat(wasmx): DB entities and admin API

flrgh:
This is still WIP and not ready for a final review, but feel free to share feedback in the meantime.

For the sake of simplicity, this implementation bundles the entire filter chain array into a JSON-serialized array. We can break filter entities off into their own table/entity with a FK relationship if need be (should decide on this before release).

## changes

* new `wasm_filter_chains` entity
  * schema
  * migrations
  * DB tests
  * admin API tests

## todo

- [ ] ~filter registration/metadata/schema integration~
  - [ ] ~validate `wasm_filter_chains.filters[].name` against known filters~
  - [ ] ~validate `wasm_filter_chains.filters[].config`~


hbagdi:
Were we not going for a single plugin that takes in the lists of filters to execute (and their configurations) to start small?

hishamhm:
@hbagdi No, we ultimately reversed on that decision because the execution flow between Lua plugins and Wasm filters (each happening on their own Nginx module) imposes some constraints that wouldn't be correctly represented if Wasm filters were encapsulated by a plugin.

thibaultcha:
> Were we not going for a single plugin that takes in the lists of filters to execute (and their configurations) to start small?

That was the 2022 Tech Preview model.

thibaultcha (commented):
LGTM so far.

flrgh:
I've been banging my head on the wall about filter config schema validation for a minute now and still don't have it fully functional and sane, so I'm punting on that for the moment. I'd rather we just get this PR merged as-is so that everyone is unblocked.

locao (approved):

PR #10318, opened 2023-02-17
fix(cmd): log format error

catbro666:
Log line isn't correctly formatted.

FTI-4801

### DIFF
--- a/kong/cmd/utils/prefix_handler.lua
+++ b/kong/cmd/utils/prefix_handler.lua
@@ -113,8 +113,8 @@ local function gen_default_ssl_cert(kong_config, target)
     end
 
     if not exists(ssl_cert) and not exists(ssl_cert_key) then
-      log.verbose("generating %s SSL certificate (%s) and key (%s) for %s",
-        target or "proxy", ssl_cert, ssl_cert_key, "listener")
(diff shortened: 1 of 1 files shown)


chronolaw (approved):

chronolaw:
I think that we may have more log format errors in cmd, could we have a complete check?

catbro666:
> I think that we may have more log format errors in cmd, could we have a complete check?

Checked. No others in cmd.

kong/cmd/utils/prefix_handler.lua
```diff
@@ -113,7 +113,8 @@ local function gen_default_ssl_cert(kong_config, target)
     end
 
     if not exists(ssl_cert) and not exists(ssl_cert_key) then
-      log.verbose("generating %s SSL certificate (", ssl_cert, ") and key (", ssl_cert_key, ") for ", target or "proxy", " listener")
+      log.verbose("generating %s SSL certificate (%s) and key (%s) for %s",
+        target or "proxy", ssl_cert, ssl_cert_key, "listener")
```
dndx, line 117:
```suggestion
        target or "proxy", ssl_cert, ssl_cert_key, "listener")
```

Why do we use the hardcoded `"listener"` in here? Should it be written in the format string instead?

dndx (approved):

PR #10325, opened 2023-02-20
fix(runloop): use upstream status code in passive health check

### LINKED ISSUE #10281: Kong upstream passive health check uses the HTTP code set by a plugin instead of  the one returned by the target server

3.0.0.0

### Current Behavior

We noticed that when our custom Kong plugin set HTTP code to '502' via its exit hook ( function exit(status, body, headers) ), the involved upstream's passive unhealthy healthcheck using HTTP status code will count this as an incident of "unhealthy" state.  When this occurs a few times in a short duration, as in a load test, the upstream will report "{"message":"failure to get a peer from the ring-balancer"}" in the errors log, essentially mark the upstream as "down". 

### Expected Behavior

We expect the upstream passive unhealthy healthcheck on HTTP status code should only mark the targets as "down" (not reachable or proxiable) using the returned HTTP status code from the target, and it should not be affected by the final HTTP status code set by Kong plugin, which is returned to the client.

### Steps To Reproduce

On Kong 3.0.0.0:
1. Create a custom Kong plugin that contains the following: 
1a) In its init_worker(), register itself as an exit hook:
  kong.response.register_hook("exit", self.exit, self)
1b) In its body_filter function, generate an exit event as:
  return kong.response.exit(500)
1c) In its exit hook function exit(status, body, headers), simply add a return statement as:
  return status, body, headers

2. Create a Kong upstream with a target server, e.g. you can use any HTTP echo service, like http://httpbin.org/headers as the target server. Make sure the passive unhealthy health check is enabled. Leave everything else as default
3. Assign the upstream to a Kong service, and then attach the custom plugin to the service
4. Send a burst of HTTP requests that triggers the service and the cusotm plugin, and verify in the errors log that you see "{"message":"failure to get a peer from the ring-balancer"}"

### Anything else?

O/S: Linux

attenuation:
### Summary

We expect the upstream passive unhealthy healthcheck on HTTP status code should only mark the targets as "down" (not reachable or proxiable) using the returned HTTP status code from the target, and it should not be affected by the final HTTP status code set by Kong plugin, which is returned to the client.

So we change the passive health check implementation to use nginx var `upstream_code`, this will not be changed by any plugin.

### Checklist

### Full changelog

* [Implement ...]

### Issue reference
FTI-4841
Fix #10281


kong/runloop/handler.lua
```diff
@@ -1513,7 +1513,7 @@ return {
       -- Report HTTP status for health checks
       local balancer_data = ctx.balancer_data
       if balancer_data and balancer_data.balancer_handle then
-        local status = ngx.status
+        local status = ngx.var.upstream_status
```
sabertobihwy, line 1516: If an upstream retry occurs, then `ngx.var.upstream_status` will be an array containing the upstream status of each retry, is there any special handling required?

hbagdi:
The reason for the fix makes sense. Please add a test.

kong/runloop/handler.lua
attenuation: Yeah, I just check lua-nginx-module code, It's a string that contains upstream stats code array, I will change this way to solve it.

kong/runloop/handler.lua
```diff
@@ -1513,7 +1513,21 @@ return {
       -- Report HTTP status for health checks
       local balancer_data = ctx.balancer_data
       if balancer_data and balancer_data.balancer_handle then
+        -- https://github.com/nginx/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_upstream.c#L5656
+        -- because of the way of nginx do the upstream_status variable, it can be
+        -- a string or a number, so we need to parse it to get the status
         local status = ngx.var.upstream_status
+        if tonumber(status) == nil then
```
chronolaw: Here we have a simple solution to get the last upstream status code,
using `sub(ngx.var.upstream_status, -3)`.

There is a example code in `pdk/service/response.lua`.

attenuation: good idea

kong/runloop/handler.lua
```diff
@@ -1513,7 +1513,14 @@ return {
       -- Report HTTP status for health checks
       local balancer_data = ctx.balancer_data
       if balancer_data and balancer_data.balancer_handle then
-        local status = ngx.status
+        -- https://github.com/nginx/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_upstream.c#L5656
+        -- because of the way of nginx do the upstream_status variable, it can be
+        -- a string or a number, so we need to parse it to get the status
+        local status = tonumber(sub(ngx.var.upstream_status or "", -3))
```
chronolaw, line 1519: should use already cached `var` to get nginx var value, like `ngx.var.upstream_status`.

More simple code style:
```
local status = tonumber(sub(var.upstream_status or "", -3)) or ngx.status
```

hbagdi:
Please add a changelog entry.

attenuation:
> 

pending for 3.2 release.

attenuation:
> Please add a changelog entry.

pending for 3.2 release, now changelog have unrelease in 3.2

kong/runloop/handler.lua
```diff
@@ -1513,7 +1513,10 @@ return {
       -- Report HTTP status for health checks
       local balancer_data = ctx.balancer_data
       if balancer_data and balancer_data.balancer_handle then
-        local status = ngx.status
+        -- https://github.com/nginx/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_upstream.c#L5656
+        -- because of the way of nginx do the upstream_status variable, it can be
+        -- a string or a number, so we need to parse it to get the status
```
chronolaw, line 1518:
```suggestion
        -- because of the way of Nginx do the upstream_status variable, it may be
        -- a string or a number, so we need to parse it to get the status
```

kong/runloop/handler.lua
```diff
@@ -1513,7 +1513,10 @@ return {
       -- Report HTTP status for health checks
       local balancer_data = ctx.balancer_data
       if balancer_data and balancer_data.balancer_handle then
-        local status = ngx.status
+        -- https://github.com/nginx/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_upstream.c#L5656
+        -- because of the way of Nginx do the upstream_status variable, it may be
+        -- a string or a number, so we need to parse it to get the status
+        local status = tonumber(sub(ngx.var.upstream_status or "", -3)) or ngx.status
```
chronolaw, line 1519: `ngx.var` => `var`

chronolaw, line 1519:
```suggestion
        local status = tonumber(sub(var.upstream_status or "", -3)) or ngx.status
```

chronolaw (approved):

kong/runloop/handler.lua
```diff
@@ -1513,7 +1513,10 @@ return {
       -- Report HTTP status for health checks
       local balancer_data = ctx.balancer_data
       if balancer_data and balancer_data.balancer_handle then
-        local status = ngx.status
+        -- https://github.com/nginx/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_upstream.c#L5656
+        -- because of the way of Nginx do the upstream_status variable, it may be
+        -- a string or a number, so we need to parse it to get the status
+        local status = tonumber(sub(var.upstream_status or "", -3)) or ngx.status
```
hbagdi, line 1519: Should we consider moving this into the PDK or some other helper function?
I can see other modules, especially around metrics and logging that would like to use this capability.

chronolaw: Good idea. Perhaps we should do it in another PR and not this.

attenuation: Thanks @hbagdi , I will open another PR to do this.

fffonion: a link to possible values of `var.upstream_status` (like http://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_addr) instead of pointing to code would be more straightforward for future us

dndx (approved):

kong/runloop/handler.lua
```diff
@@ -1511,7 +1511,10 @@ return {
       -- Report HTTP status for health checks
       local balancer_data = ctx.balancer_data
       if balancer_data and balancer_data.balancer_handle then
-        local status = ngx.status
+        -- https://github.com/nginx/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_upstream.c#L5656
```
chronolaw, line 1514:
```suggestion
        -- https://nginx.org/en/docs/http/ngx_http_upstream_module.html#variables
```

PR #10327, opened 2023-02-20
feat(request-transformer): honor untrusted_lua configuration values

samugi:
### Summary

This PR enables support for the following configuration parameters and values in the `request-transformer` plugin:

* untrusted_lua = "on", "off", "sandbox"
* untrusted_lua_sandbox_requires
* untrusted_lua_sandbox_environment

### Checklist

### Full changelog

* [Implement ...]

[KAG-168]
[KAG-169]

[KAG-168]: https://konghq.atlassian.net/browse/KAG-168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KAG-169]: https://konghq.atlassian.net/browse/KAG-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/03-plugins/36-request-transformer/02-access_spec.lua
+++ b/spec/03-plugins/36-request-transformer/02-access_spec.lua
@@ -2228,6 +2228,433 @@ describe("Plugin: request-transformer(access) [#" .. strategy .. "]", function()
   end)
 end)
 
+describe("Plugin: request-transformer(access) [#" .. strategy .. "] untrusted_lua=off", function()
+  local client
+
+  lazy_setup(function()
+    local bp = helpers.get_db_utils(strategy, {
+      "routes",
+      "services",
+      "plugins",
+    })
+
+    local route1 = bp.routes:insert({
+      hosts = { "testLuaOff.test" }
+    })
+
+    local route2 = bp.routes:insert({
+      hosts = { "testLuaOff2.test" }
+    })
+
+    bp.plugins:insert {
+      route = { id = route1.id },
+      name = "request-transformer",
+      config = {
+        add = {
+          headers = {"x-added:some_value"},
+        }
+      }
+    }
+
+    bp.plugins:insert {
+      route = { id = route2.id },
+      name = "request-transformer",
+      config = {
+        add = {
+          headers = {"x-added:$(require('inspect')(string.sub(query_params.user, 1, 3)))"},
+        }
+      }
+    }
+
+    assert(helpers.start_kong({
+      database = strategy,
+      untrusted_lua = "off",
+      plugins = "bundled, request-transformer",
+      nginx_conf = "spec/fixtures/custom_nginx.template",
+    }))
+  end)
+
+  lazy_teardown(function()
+    helpers.stop_kong()
+  end)
+
+  before_each(function()
+    client = helpers.proxy_client()
+  end)
+
+  after_each(function()
+    if client then client:close() end
+  end)
+
+
+  it("correctly handles plain text value", function()
+    local r = assert(client:send {
+      method = "GET",
+      path = "/",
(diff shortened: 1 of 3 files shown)


spec/03-plugins/36-request-transformer/02-access_spec.lua
```diff
@@ -2227,4 +2227,400 @@ describe("Plugin: request-transformer(access) [#" .. strategy .. "]", function()
...
+
+  before_each(function()
+    client = helpers.proxy_client()
+  end)
+
+  after_each(function()
+    if client then client:close() end
+  end)
+
+  describe("using template", function()
```
dndx, line 2275: Should we also test use of expressions when `untrusted_lua=off` to make sure it does not work?

samugi: I added a better unhappy path to the test, thanks @dndx .

Currently the behaviour of `untrusted_lua=off` is to just not render the template which will be treated as just a regular string; This allows us to not do any parsing of the template to determine whether it contains inline Lua, and [just disable it](https://github.com/Kong/kong/pull/10327/files#diff-41ae27766e102f74d9deb4899fc8ba265b0dcd602e2ed8a0a150d5fc03ab406fR47) on penlight. If we wanted to actually fail and return some error when Lua is detected though, we'd need to do some pattern matching, I'm still unsure which is the best approach.

spec/03-plugins/36-request-transformer/02-access_spec.lua
```diff
@@ -2227,4 +2227,427 @@ describe("Plugin: request-transformer(access) [#" .. strategy .. "]", function()
...
+        user = "foo"
+      },
+      headers = {
+        host = "testLuaOff2.test",
+        ["Content-Type"] = "application/x-www-form-urlencoded",
+      }
+    })
+    assert.response(r).has.status(200)
+    local value = assert.request(r).has.header("x-added")
+    assert.equals("$(require('inspect')(string.sub(query_params.user, 1, 3)))", value)
```
kikito, line 2318: I am not sure this (rendering the template without any evaluation) is equivalent to what we do in the serverless function plugin(s). I think on those we [always call the sandbox](https://github.com/Kong/kong/blob/master/kong/plugins/pre-function/_handler.lua#L22), and then the sandbox [throws an error](https://github.com/Kong/kong/blob/master/spec/01-unit/20-sandbox_spec.lua#L111) like with the contents of `sandbox.configuration.err_msg`:

```suggestion
    assert.response(r).has.status(500)
    assert.request(r).not.has.header("x-added")
    -- and perhaps "the error log now shows the contents of sandbox.configuration.err_msg somewhere
```

In other words - ideally the plugin code should not "look" at the config at all. Instead it should always run the sandbox, passing the config options to it. I don't know how feasible this is, given that we run the code using penlight's template thingie.

kikito (changes requested):

spec/03-plugins/36-request-transformer/02-access_spec.lua
samugi: discussed and decided to "detect" lua inline expressions and throw error instead of returning non-parsed template

kikito (approved):

kikito:
@samugi this PR needs an update in the request-transformer docs page. Our page for request-transformer does not have a changelog at the end yet, but use the one in request-transformer-advanced as an example.

PR #10331, opened 2023-02-21
perf(postgres): optimize the expired rows cleanup routine in postgres connector

windmgc:
### Summary

This PR tries to optimize the expired rows cleanup routine in the Postgres strategy connector.

Currently, the cleanup logic is managed by a `timer.every`, which will run multiple `DELETE` sub-queries in one big query and delete those rows that got expired. Concerns have been raised about such queries may lead to high load on database if the number of expired rows is large. The idea originates from [FTI-4746](https://konghq.atlassian.net/browse/FTI-4746) and [KAG-516](https://konghq.atlassian.net/browse/KAG-516), but the code that got changed in this PR may not be directly relevant to the root cause of the production issue. The logic itself is definitely worth some optimizing after all.

This PR brings the following changes to the cleanup logic:
- Makes the cleanup interval configurable. A new config parameter `pg_expired_rows_cleanup_interval` is introduced to control the interval of triggering the timer.
- A cluster mutex is added to ensure that in a "traditional" deployment cluster, only one cleanup will be running at the same time.
- The big query which contains multiple `DELETE` sub-queries has been split into multiple queries, with more obvious logging, so that if any errors are encountered, the table which caused such an error can be located easily. And each query will use its own connection so that if any of them got timed out on Kong's side(for example, waiting for locks in the Pg but the connection itself is timed out), the following cleanup will not be affected.
- Debug logging which shows the total time usage of each table is added.
- Batch delete is implemented in every DELETE query to avoid letting extremely slow query consuming database resources.

### Checklist

### Full changelog

* Optimize the expired rows cleanup routine in Postgres connector.
* Add a new config parameter `pg_expired_rows_cleanup_interval` to control the interval of expired rows cleanup routine.

### Issue reference

- [FTI-4746](https://konghq.atlassian.net/browse/FTI-4746)
- [KAG-516](https://konghq.atlassian.net/browse/KAG-516)

[FTI-4746]: https://konghq.atlassian.net/browse/FTI-4746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KAG-516]: https://konghq.atlassian.net/browse/KAG-516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FTI-4746]: https://konghq.atlassian.net/browse/FTI-4746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -337,7 +337,7 @@ function _mt:init_worker(strategies)
 
     local cleanup_statement = concat(cleanup_statements, "\n")
 
-    return timer_every(60, function(premature)
+    return timer_every(self.config.expired_rows_cleanup_interval, function(premature)
       if premature then
         return
       end
@@ -934,6 +934,7 @@ function _M.new(kong_config)
     sem_timeout = (kong_config.pg_semaphore_timeout or 60000) / 1000,
     pool_size   = kong_config.pg_pool_size,
     backlog     = kong_config.pg_backlog,
+    expired_rows_cleanup_interval = kong_config.pg_expired_rows_cleanup_interval or 60,
 
     --- not used directly by pgmoon, but used internally in connector to set the keepalive timeout
     keepalive_timeout = kong_config.pg_keepalive_timeout,
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -36,6 +36,7 @@ local insert       = table.insert
 
 local WARN                          = ngx.WARN
 local ERR                           = ngx.ERR
+local DEBUG                         = ngx.DEBUG
 local SQL_INFORMATION_SCHEMA_TABLES = [[
 SELECT table_name
   FROM information_schema.tables
@@ -318,6 +319,7 @@ end
 
 
 local function cleanup_expired_rows_in_table(connector, table_name)
+  local time1 = now()
   local ttl_escaped = connector:escape_identifier("ttl")
   local expired_at_escaped = connector:escape_identifier("expire_at")
   local column_name = table_name == "cluster_events" and expired_at_escaped
@@ -332,6 +334,10 @@ local function cleanup_expired_rows_in_table(connector, table_name)
   }
 
   local ok, err = connector:query(cleanup_statement)
+  local time2 = now()
+  local time_elapsed = tonumber(fmt("%.3f", time2 - time1))
+  log(DEBUG, "cleaning up expired rows from table '", table_name,
+            "' took ", time_elapsed, " seconds")
   if not ok then
     if err then
       log(WARN, "failed to clean expired rows from table '",
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -319,7 +319,7 @@ end
 
 local function cleanup_expired_rows_in_table(connector, table_name)
   local ttl_escaped = connector:escape_identifier("ttl")
-  local expired_at_escaped = connector:escape_identifier("expired_at")
+  local expired_at_escaped = connector:escape_identifier("expire_at")
   local column_name = table_name == "cluster_events" and expired_at_escaped
                                                       or ttl_escaped
 
--- a/kong/db/strategies/postgres/connector.lua
+++ b/kong/db/strategies/postgres/connector.lua
@@ -9,6 +9,7 @@ local constants = require "kong.constants"
 local knode        = kong and kong.node
                      or require "kong.pdk.node".new()
 
+local tablex = require "pl.tablex"
 
 local setmetatable = setmetatable
 local encode_array = arrays.encode_array
@@ -55,6 +56,7 @@ local OPERATIONS = {
 local ADMIN_API_PHASE = kong_global.phases.admin_api
 local CORE_ENTITIES = constants.CORE_ENTITIES
 local EXPIRED_ROWS_CLEANUP_LOCK_KEY = "db_cluster_expired_rows_cleanup"
+local EMPTY_T = tablex.readonly {}
 
 
 local function now_updated()
@@ -318,7 +320,10 @@ function _mt:init()
 end
 
 
-local function cleanup_expired_rows_in_table(connector, table_name)
+local function cleanup_expired_rows_in_table(config, table_name)
+  -- Create new connection on each table to avoid reusing a connection that might be already timed out
+  local connector = connect(config)
+
   local time1 = now()
   local ttl_escaped = connector:escape_identifier("ttl")
   local expired_at_escaped = connector:escape_identifier("expire_at")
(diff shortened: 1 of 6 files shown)


kong/db/strategies/postgres/connector.lua
```diff
@@ -314,55 +322,94 @@ function _mt:init()
...
 
 
-function _mt:init_worker(strategies)
-  if ngx.worker.id() == 0 then
+local function cleanup_expired_rows_in_table(config, table_name)
+  -- Create new connection on each table to avoid reusing a connection that might be already timed out
+  local connector = connect(config)
+
+  local cleanup_start_time = now_updated()
+  local ttl_escaped = connector:escape_identifier("ttl")
```
liverpool8056, line 330: Would it be better if we define `ttl_escaped` as a global variable so that we won't have to run escape_identifier every time for the same result.
By the way, personally I don't think either `ttl` or `expired_at` need to be escaped, neither of them is a keyword in current postgres-sql syntax.

kong/db/strategies/postgres/connector.lua
```diff
@@ -314,55 +322,94 @@ function _mt:init()
...
+                                                      or ttl_escaped
+
+  local cleanup_statement = concat {
+    "DELETE FROM ",
+    connector:escape_identifier(table_name),
+    " WHERE ctid in (SELECT ctid FROM ",
+    connector:escape_identifier(table_name),
+    " WHERE ",
+    column_name,
+    " < TO_TIMESTAMP(%s) AT TIME ZONE 'UTC' LIMIT ",
```
liverpool8056, line 342: Would it be helpful if we force db to use index here? I think it might not be better, but at least it wouldn't be worse than seq scan when using index.

kong/db/strategies/postgres/connector.lua
```diff
@@ -314,55 +322,94 @@ function _mt:init()
...
+        log(WARN, "failed to clean expired rows from table '",
+                  table_name, "' on PostgreSQL database (",
+                  err, ")")
+      else
+        log(WARN, "failed to clean expired rows from table '",
+                  table_name, "' on PostgreSQL database")
+      end
+    end
+
+    if ok.affected_rows == 0 then
```
liverpool8056, line 364: Maybe we could use `ok.affected_rows <= (10% of EXPIRED_ROW_BATCH_SIZE)` so that we will never hit an unnecessary loop especially an unnecessary query.
If every time there is a 10% left, it could be cleared at next window. Even we may consider using `ok.affected_rows <  EXPIRED_ROW_BATCH_SIZE` for making use of the capacity of batch fully.

kong/db/strategies/postgres/connector.lua
```diff
@@ -314,55 +322,94 @@ function _mt:init()
...
+      break
     end
+  end
+
+  local cleanup_end_time = now_updated()
+  local time_elapsed = tonumber(fmt("%.3f", cleanup_end_time - cleanup_start_time))
+  log(DEBUG, "cleaning up expired rows from table '", table_name,
+            "' took ", time_elapsed, " seconds")
+  connector:disconnect()
```
liverpool8056, line 373: If there is a connection pool and the usage of connection pool could be in our control, we should try to use it rather than disconnecting. Processing connection is a expensive action for DB.

kong/db/strategies/postgres/connector.lua
windmgc: Good suggestion, I've changed to 
```
tonumber(ok.affected_rows) < EXPIRED_ROW_BATCH_SIZE
```
to check when the last loop happens

kong/db/strategies/postgres/connector.lua
windmgc: I did a google and found no way of using force index in Postgres... it seems that this is not doable
`SELECT`ing a batch of rows here should be fine since we already have `ttl` index for these table, so let's just leave the job for query engine to decide whether it should use index

kong/db/strategies/postgres/connector.lua
windmgc: I chose to create a new connection for each table here intentionally, to avoid problem caused by connection reuse(such as former connection waiting for locks and timed-out, re-use on this connection will be problematic)

kong/db/strategies/postgres/connector.lua
windmgc: The `escape_identifier` depends on creating a Postgres connector in pgmoon, and it is merely just string manipulation  so the cost is small, so I think it's okay to keep the current state

bungle:
Question: what if we drop the Postgres cleanup routine altogether, and instead create Postgres triggers on entities that have `ttl=true`? We could run the trigger `AFTER INSERT`. I think that would be a better solution. This issue only affects `traditional` deployments. Very little the `Control Plane` and none the `Data Planes` or `DBless`.

windmgc:
Hmm I'm a bit concerned about the `AFTER INSERT` trigger(if we're going to add a `DELETE *` instead of a batch deleting), it could also lead to more concurrent queries and high resource usage, right? Say there is a user's scenario that oauth2_tokens are created frequently, since hybrid mode cannot be used under the oauth2 case, there will still be some users having large traditional clusters deployed.

But yes, dropping the background timer and adding a trigger(and batch deleting in the function perhaps?) is more straightforward, we could re-consider this option after the CP-DP refactor thing happens. What do you think about it? @bungle

windmgc:
```
The ttl fields on all tables (where schema says ttl=true) need proper indexes as they are also used with SELECTs.
```
@bungle Could you please show me where can I change to make this guarantee? I can confirm that all these ttl enabled tables have a proper index on ttl fields in our migration files, but I don't know if there is any DAO mechanics that can guarantee the schema field is indexed.

bungle:
> @bungle Could you please show me where can I change to make this guarantee?

Yes, just by looking our Postgres schemas that `ttl=true` tables have the index. If they don't we need to add index for `ttl`.  We have added them during the years, but there might still be some entities that are missing Postgres index.

bungle:
> Hmm I'm a bit concerned about the `AFTER INSERT` trigger(if we're going to add a `DELETE *` instead of a batch deleting)


The batch deleting as you added here is basically Postgres feature. Thus it is easy to add it in trigger delete too.

> it could also lead to more concurrent queries and high resource usage, right? Say there is a user's scenario that oauth2_tokens are created frequently, since hybrid mode cannot be used under the oauth2 case, there will still be some users having large traditional clusters deployed.

We can tune this inside trigger. E.g. delete only sometimes etc. 

What I mean that the current logic is very complicated as in this PR. I think in most cases what we already had is better than what we have here. So then we have probably a single pain point that is oauth2 tokens. Perhaps oauth2 tokens ttling should be implemented differently while the rest are fine with current or triggers. Perhaps triggers can be made to work with oauth2 tokens problem too.

windmgc:
superseded by #10389

PR #10332, opened 2023-02-21
fix(plugin): OTEL exporting with translate

StarlightIbuki:
### Summary

We represent trace spans with a universal format for all kinds of mainstream tracers and translate them to different formats when propagating headers and exporting traces. However, OTEL misses the handling of trace ID length.

### Checklist

N/A There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE

### Issue reference

Fix KAG-652

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -195,7 +195,7 @@
     [#10160](https://github.com/Kong/kong/pull/10160)
   - For `http.flavor`. It should be a string value, not a double.
     [#10160](https://github.com/Kong/kong/pull/10160)
-- **OpenTelemetry**: Fix a bug that when get trace of other formats, the trace ID reported and propagated could be of incorrect length.
+- **OpenTelemetry**: Fix a bug that when getting the trace of other formats, the trace ID reported and propagated could be of incorrect length.
     [#10332](https://github.com/Kong/kong/pull/10332)
 - **OAuth2**: `refresh_token_ttl` is now limited between `0` and `100000000` by schema validator. Previously numbers that are too large causes requests to fail.
   [#10068](https://github.com/Kong/kong/pull/10068)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,8 +193,10 @@
     [#10069](https://github.com/Kong/kong/pull/10069)
   - For `http.status_code`. It should be present on spans for requests that have a status code.
     [#10160](https://github.com/Kong/kong/pull/10160)
-  - For `http.flavor`. It should be a string value, not a double.
+  - For `http.flavor`. It should be a string value, not a double
     [#10160](https://github.com/Kong/kong/pull/10160)
+- **OpenTelemetry**: Fix a bug that when get trace of other formats, the trace ID reported and propagated could be of incorrect length.
+    [#10332](https://github.com/Kong/kong/pull/10332)
 - **OAuth2**: `refresh_token_ttl` is now limited between `0` and `100000000` by schema validator. Previously numbers that are too large causes requests to fail.
   [#10068](https://github.com/Kong/kong/pull/10068)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,7 +193,7 @@
     [#10069](https://github.com/Kong/kong/pull/10069)
   - For `http.status_code`. It should be present on spans for requests that have a status code.
     [#10160](https://github.com/Kong/kong/pull/10160)
-  - For `http.flavor`. It should be a string value, not a double
+  - For `http.flavor`. It should be a string value, not a double.
     [#10160](https://github.com/Kong/kong/pull/10160)
 - **OpenTelemetry**: Fix a bug that when get trace of other formats, the trace ID reported and propagated could be of incorrect length.
     [#10332](https://github.com/Kong/kong/pull/10332)
--- a/kong/plugins/opentelemetry/otlp.lua
+++ b/kong/plugins/opentelemetry/otlp.lua
@@ -10,6 +10,7 @@ local insert = table.insert
 local tablepool_fetch = tablepool.fetch
 local tablepool_release = tablepool.release
 local deep_copy = utils.deep_copy
+local shallow_copy = utils.shallow_copy
 local table_merge = utils.table_merge
 
 local POOL_OTLP = "KONG_OTLP"
@@ -84,7 +85,7 @@ local function translate_span(span)
     trace_id = string.rep("\0", 16 - #trace_id) .. trace_id
   end
 
-  local translated = deep_copy(span)
+  local translated = shallow_copy(span)
   translated.trace_id = trace_id
   return translated
 end
--- a/kong/plugins/opentelemetry/otlp.lua
+++ b/kong/plugins/opentelemetry/otlp.lua
@@ -15,6 +15,7 @@ local table_merge = utils.table_merge
 local getmetatable = getmetatable
 local setmetatable = setmetatable
 
+local TRACE_ID_LEN = 16
 local NULL = "\0"
 local POOL_OTLP = "KONG_OTLP"
 local EMPTY_TAB = {}
@@ -83,11 +84,11 @@ local function translate_span(span)
   local new_id = trace_id
 
   -- make sure the trace id is of 16 bytes
-  if len > 16 then
-    new_id = trace_id:sub(-16)
+  if len > TRACE_ID_LEN then
+    new_id = trace_id:sub(-TRACE_ID_LEN)
 
-  elseif len < 16 then
-    new_id = NULL:rep(16 - len) .. trace_id
+  elseif len < TRACE_ID_LEN then
+    new_id = NULL:rep(TRACE_ID_LEN - len) .. trace_id
   end
 
   if new_id ~= trace_id then
--- a/kong/plugins/opentelemetry/otlp.lua
+++ b/kong/plugins/opentelemetry/otlp.lua
@@ -83,6 +83,10 @@ local function translate_span(span)
   local len = #trace_id
   local new_id = trace_id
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -177,8 +177,10 @@
     [#10069](https://github.com/Kong/kong/pull/10069)
   - For `http.status_code`. It should be present on spans for requests that have a status code.
     [#10160](https://github.com/Kong/kong/pull/10160)
-  - For `http.flavor`. It should be a string value, not a double.
+  - For `http.flavor`. It should be a string value, not a double
```
StarlightIbuki, line 180:
```suggestion
  - For `http.flavor`. It should be a string value, not a double.
```

flrgh:
I've updated the PR with a couple fixes. I _think_ it meets the correct requirements now, but there are some warts:

1. We only have test coverage for the trace ID zero-pad case. There is no test for the trace ID truncate case, because an incoming trace ID that is too long will be [discarded as invalid](https://github.com/Kong/kong/blob/93d3168bc9436febf0b6e79cb9d1d8375ecec5b1/kong/tracing/propagation.lua#L231-L233).
2. The way we are copying the span table seems like it may have unintended consequences. [Spans use table pools under the hood](https://github.com/Kong/kong/blob/93d3168bc9436febf0b6e79cb9d1d8375ecec5b1/kong/pdk/tracing.lua#L337-L348), so the PDK expects to manage their lifecycle. Copying spans outside of the PDK can therefore lead to situations where tables are orphaned and gc-ed instead of being returned to the pool as intended.

StarlightIbuki:
> I've updated the PR with a couple fixes. I _think_ it meets the correct requirements now, but there are some warts:
> 
> 1. We only have test coverage for the trace ID zero-pad case. There is no test for the trace ID truncate case, because an incoming trace ID that is too long will be [discarded as invalid](https://github.com/Kong/kong/blob/93d3168bc9436febf0b6e79cb9d1d8375ecec5b1/kong/tracing/propagation.lua#L231-L233).
> 2. The way we are copying the span table seems like it may have unintended consequences. [Spans use table pools under the hood](https://github.com/Kong/kong/blob/93d3168bc9436febf0b6e79cb9d1d8375ecec5b1/kong/pdk/tracing.lua#L337-L348), so the PDK expects to manage their lifecycle. Copying spans outside of the PDK can therefore lead to situations where tables are orphaned and gc-ed instead of being returned to the pool as intended.

Answering question 2: the copied span won't survive the request so it will be GC-ed, and its reference to the members will also be gone. So the only reference to those members should be from the original span, which will be returned to the pool. If the PDK decides to also reuse the member, it will still be fine as the table returned to the pool has reference to the members.

kong/plugins/opentelemetry/otlp.lua
```diff
@@ -72,9 +76,37 @@ local function transform_events(events)
...
 end
 
+-- this function is to translate the span to otlp span, but of universal span format
+local function translate_span(span)
+  local trace_id = span.trace_id
+  local len = #trace_id
+  local new_id = trace_id
+
+  -- make sure the trace id is of 16 bytes
+  if len > 16 then
```
chronolaw, line 86: We should not use the magic number `16`, what does it mean? 
Could we define a readable const for it?

StarlightIbuki:
> I've updated the PR with a couple fixes. I _think_ it meets the correct requirements now, but there are some warts:
> 
> 1. We only have test coverage for the trace ID zero-pad case. There is no test for the trace ID truncate case, because an incoming trace ID that is too long will be [discarded as invalid](https://github.com/Kong/kong/blob/93d3168bc9436febf0b6e79cb9d1d8375ecec5b1/kong/tracing/propagation.lua#L231-L233).
> 2. The way we are copying the span table seems like it may have unintended consequences. [Spans use table pools under the hood](https://github.com/Kong/kong/blob/93d3168bc9436febf0b6e79cb9d1d8375ecec5b1/kong/pdk/tracing.lua#L337-L348), so the PDK expects to manage their lifecycle. Copying spans outside of the PDK can therefore lead to situations where tables are orphaned and gc-ed instead of being returned to the pool as intended.

To answer question 1: I agree that we need a test for the truncating case. But the code you're referring to (handling to OT header) does not mean truncating doesn't work. We're translating traces, and a hypothetical trace format may have a valid trace ID larger than that.

StarlightIbuki:
However there's not a type of ID that is long enough to go through the truncating path, so we skip it for now.

CHANGELOG.md
```diff
@@ -195,6 +195,8 @@
     [#10160](https://github.com/Kong/kong/pull/10160)
   - For `http.flavor`. It should be a string value, not a double.
     [#10160](https://github.com/Kong/kong/pull/10160)
+- **OpenTelemetry**: Fix a bug that when get trace of other formats, the trace ID reported and propagated could be of incorrect length.
```
chronolaw, line 198:
```suggestion
- **OpenTelemetry**: Fix a bug that when getting the trace of other formats, the trace ID reported and propagated could be of incorrect length.
```

kong/plugins/opentelemetry/otlp.lua
```diff
@@ -72,9 +77,37 @@ local function transform_events(events)
   return pb_events
 end
 
+-- this function is to translate the span to otlp span, but of universal span format
+local function translate_span(span)
+  local trace_id = span.trace_id
+  local len = #trace_id
+  local new_id = trace_id
+
```
chronolaw, line 85: should return `span` if len == 16

chronolaw (approved):
We could do more code clean for it, but I think it is OK now.

kong/plugins/opentelemetry/otlp.lua
```diff
@@ -72,9 +77,39 @@ local function transform_events(events)
   return pb_events
 end
 
+-- this function is to translate the span to otlp span, but of universal span format
+local function translate_span(span)
```
kikito, line 81: This comment + function name don't clarify what the function is doing. Since it only truncates/elongates the trace id:

kong/plugins/opentelemetry/otlp.lua
```diff
@@ -72,9 +77,39 @@ local function transform_events(events)
   return pb_events
 end
 
+-- this function is to translate the span to otlp span, but of universal span format
```
kikito, line 80: I don't understand what "but of universal span format" means.

kong/plugins/opentelemetry/otlp.lua
```diff
@@ -72,9 +77,39 @@ local function transform_events(events)
...
+    new_id = NULL:rep(TRACE_ID_LEN - len) .. trace_id
+  end
+
+  local translated = shallow_copy(span)
+  setmetatable(translated, getmetatable(span))
+
+  translated.trace_id = new_id
+  span = translated
+
+  return span
```
kikito, line 104: Nitpick, not needed if we don't do a shallow copy.
```suggestion
  return translated
```

kong/plugins/opentelemetry/otlp.lua
```diff
@@ -72,9 +77,39 @@ local function transform_events(events)
...
+
+  -- make sure the trace id is of 16 bytes
+  if len > TRACE_ID_LEN then
+    new_id = trace_id:sub(-TRACE_ID_LEN)
+
+  elseif len < TRACE_ID_LEN then
+    new_id = NULL:rep(TRACE_ID_LEN - len) .. trace_id
+  end
+
+  local translated = shallow_copy(span)
```
kikito, line 98: Why is a `shallow_copy` needed?

I would have expected this function to do something like:

```
local function adjust_trace_id_for_oltp(trace_id)
```

And the usage would be, when creating the root span:
```
root_span.trace_id = adjust_trace_id_for_oltp(trace_id)
```

And inside `transform_span`:
```
local pb_span = {
    trace_id = adjust_trace_id_for_oltp(span.trace_id),
```

kikito: Also, I don't recommend using abstract names (like `transform_span` or `translate_span`) when the function is only doing one modification on one field. The tradeoff ("making the code easier to modify in the future if we find that we need to do more modifications in the future") is not worth the upfront cost ("making the code more difficult to read now"). Future us can always group several modifications into one if the need arises.

kong/plugins/opentelemetry/otlp.lua
StarlightIbuki: The original design to support arbitrary types of tracing plugins has not made it clear:
1. We have an internal representation for span, and it should preserve what it received from downstream requests at best effort;
2. when exporting the data to collectors, it needs to transform the span to meet the requirement with its field (DD, OTEL, etc), and then translate it into a format used for the reporting API (gRPC, JSONRPC, etc). This function does the first step but not the second one.

kong/plugins/opentelemetry/otlp.lua
StarlightIbuki: > Why is a `shallow_copy` needed?

The span passed in may also be used by other plugins. We need to preserve as much original information as possible, and only do the transformation for exporting/propagating.

StarlightIbuki: I agree naming is a problem, but `transform_span` is used as some kind of interface for tracing plugins (we can see a function with the same name from datadog-tracing).
I'm adding comments and renaming the internal function.

StarlightIbuki: And the function is in the file "otlp.lua" so I think we do not need to suffix it with "\_otlp".

StarlightIbuki: > And the usage would be, when creating the root span:

We cannot just modify the `root_span` given the reason: https://github.com/Kong/kong/pull/10332/files#r1115411177

kikito (approved):

Issue #7418, opened 2021-06-03
 a client request body is buffered to a temporary file 


highwayliu:
### Summary
- Kong version (`$ kong version`)
- 2.4.1
- A concise summary of the bug.
- I user kong for grpc, and the kong error log always write " a client request body is buffered to a temporary file /usr/local/kong/client_body_temp/0000000181" , I have inject the nginx variable by docker environment. But ie doesn't work.
# injected nginx_proxy_* directives         
    client_body_buffer_size 8m;                  
    client_max_body_size 8m;                    
    proxy_buffer_size 128k;                      
    proxy_buffers 4 256k;                       
    proxy_busy_buffers_size 256k;
    real_ip_header X-Real-IP;
    real_ip_recursive off;          

> spec:

      containers:
      - env:
        - name: KONG_DATABASE
          value: postgres
        - name: KONG_PG_HOST
          value: 172.26.0.45
        - name: KONG_PG_PORT
          value: "5432"
        - name: KONG_PG_USER
          value: kong
        - name: KONG_PG_PASSWORD
          value: kong
        - name: KONG_PG_DATABASE
          value: kong
        - name: KONG_PLUGINS
          value: bundled,grpc-itop-verify,grpc-api-log,grpc-error-handler,p-prometheus,correlationid,itop-verify,g-gateway
        - name: KONG_LUA_PACKAGE_PATH
          value: ./?.lua;./?/init.lua;/usr/local/custom/?.lua;;
        - name: KONG_PROXY_LISTEN
          value: 0.0.0.0:8000, 0.0.0.0:8443 ssl, 0.0.0.0:9080 http2, 0.0.0.0:9081 http2 ssl
        - name: KONG_ADMIN_LISTEN
          value: 0.0.0.0:8001
        - name: KONG_PROXY_ACCESS_LOG
          value: /usr/local/kong/logs/proxy-access.log
        - name: KONG_ADMIN_ACCESS_LOG
          value: /usr/local/kong/logs/admin-access.log
        - name: KONG_PROXY_ERROR_LOG
          value: /usr/local/kong/logs/proxy-error.log
        - name: KONG_ADMIN_ERROR_LOG
          value: /usr/local/kong/logs/admin-error.log
        - name: KONG_SSL_CERT
          value: /usr/local/custom/kong/ssl/server.pem
        - name: KONG_SSL_CERT_KEY
          value: /usr/local/custom/kong/ssl/server.key
        - name: KONG_LOG_LEVEL
          value: info
[...]

PR #10333, opened 2023-02-21
chore(session) do not read body by default and only on certain HTTP methods

bungle:
### Summary

On comment: https://github.com/Kong/kong/issues/7418#issuecomment-1411729346 @PidgeyBE mentioned that session plugin reads bodies by on every HTTP request that is not an GET request.

Because it is quite common to use bodies to send large files, reading body makes features like route.request_buffering=off, not working. Thus, the default value for ´logout_post_arg` in session plugin was removed. The bodies are only read when this is configured. This might change behaviuor on scripts that create session plugin and which also think that logout by body argument works as before. On the other hand it is a way more common to read session than it is to log out session, thus it should be better value for future of us.


kong/plugins/session/schema.lua
```diff
@@ -94,7 +94,7 @@ return {
           { request_headers = headers },
           { logout_methods = logout_methods },
           { logout_query_arg = {  type = "string",  default = "session_logout" } },
-          { logout_post_arg = { type = "string", default = "session_logout" } },
```
samugi, line 97: my concern would be, as you pointed out, those who are relying on the removed default. Seems like manager might be one of them: https://github.com/Kong/kong-admin/blob/4dfc92fe311a21d6e24142e2d7807c71f3b768e7/src/stores/auth.ts#L73 (?)

bungle: Manager should force the setting to whatever they need for this logout parameter, and not rely on user configuration on it. Manager owns the logout and they know how it is to be used. We need to check this with Manager code.

The line you pointed out uses `DELETE` and parameter is added to query, so this change should not affect at least that line of code. The `logout_query_arg` still has the default.

bungle: Here is a link to EE code (not publicly visible):
https://github.com/Kong/kong-ee/blob/master/kong/api/routes/kong.lua#L398-L408

That `local session_conf`  should contain all the settings that Manager needs, e.g. it should force the `logout_post_arg=xxx` if it really needs that (despite what user might have configured - it should be a setting fully controlled by Manager code itself).

samugi: > The line you pointed out uses DELETE and parameter is added to query, so this change should not affect at least that line of code
you are right. I made a quick search and noticed the same value, in that case applied to the query arg. Still not ideal that it appears to be relying on the default value of `logout_query_arg`.

> it should be a setting fully controlled by Manager code itself, not the user configuring manager
agree

hbagdi:
Aapo, is there a way to fix the performance problem without a breaking change?
Perhaps giving users a knob to disable reading bodies (and then calling that out in documentation)?

I'm all for fixing performance issues but I don't agree that this warrants a breaking change in Kong. Sessions library changes in 3.2 have enough breaking changes, we should not pile on that and instead try to limit them as much as possible.
What do you think?

PidgeyBE:
> Aapo, is there a way to fix the performance problem without a breaking change? Perhaps giving users a knob to disable reading bodies (and then calling that out in documentation)?
> 
> I'm all for fixing performance issues but I don't agree that this warrants a breaking change in Kong. Sessions library changes in 3.2 have enough breaking changes, we should not pile on that and instead try to limit them as much as possible. What do you think?

If you are looking for a non-breaking compromise, you could always try to check the `Content-Length` header and only get the body if it is below a certain threshold... In this way the performance issue is fixed for (large) file uploads and it won't break logout flows (which should be small requests in size).

bungle:
> I'm all for fixing performance issues but I don't agree that this warrants a breaking change in Kong.

As explained it is not only a performance fix, it also makes `route.request_buffering=false` to not work, so it can be viewed as a bug too. It does not break any existing configurations as they already have the default value stored in db. Just some scripts that rely on default value. Not a huge breakage.

bungle:
> If you are looking for a non-breaking compromise, you could always try to check the `Content-Length` header and only get the body if it is below a certain threshold... In this way the performance issue is fixed for (large) file uploads and it won't break logout flows (which should be small requests in size).

Just wondering whether it is too magical.

bungle:
> Aapo, is there a way to fix the performance problem without a breaking change? Perhaps giving users a knob to disable reading bodies (and then calling that out in documentation)?

The knob is there, aka set `logout_post_arg` actively to `null`. But that has turned out to be complicated with yamls and stuff.

hbagdi:
> It does not break any existing configurations as they already have the default value stored in db. Just some scripts that rely on default value. Not a huge breakage.


That's simplifying the breaking change problem a bit.
Whenever these defaults change, alternate control planes like Koko which work with DPs of multiple versions at the same time have a hard time adopting this change since there can be only one default (Kong schemas are not versioned whatsoever).

If the default value before this patch is captured in decK, KIC, etc, then the effectiveness of this "bug-fix" is questionable as well since fixing the bug actually requires more work than upgrading the gateway.

Are you more open to an alternate where we introduce a net new schema field that toggles the behavior of reading bodies or not?
This is still a breaking change in terms of the behavior of Kong but it is not a breaking change to the schema of Kong. We are sensitive to both kinds of breaking changes but the schema ones are more problematic currently. This further reduces the likelihood that a user will still run into this bug (the bug being reading bodies even with buffering disabled) after an upgrade.

bungle:
> Are you more open to an alternate where we introduce a net new schema field that toggles the behavior of reading bodies or not?

As said, there is already an option, just actively set `logout_post_arg` to null. The change in this PR was trying to make it default as having better defaults means better usability and that adding session plugin with default settings to Kong will not screw up the performance and other features. I am open adding feature `read_bodies` or something with a default value of `false`, but I don't think it makes sense to add one with default value of `true`.

kong/plugins/session/schema.lua
```diff
@@ -92,6 +92,7 @@ return {
           { remember_absolute_timeout = { type = "number", default = 2592000 } },
           { response_headers = headers },
           { request_headers = headers },
+          { read_bodies = { type = "boolean", default = false } },
```
hbagdi, line 95:
```suggestion
          { read_request_body_on_logout = { type = "boolean", default = false } },
```

nit: if we plan on expanding the usage on flows other than logout, then we can drop `_on_logout`. Not a blocker for merging.

bungle: The `logout` is kinda magical as it is always enabled (so this plugin always tries to check if the request is logout based on config parameters). Perhaps we should add option like `disable_logout=true`, would that make more sense? Ping @hbagdi (@hanshuebner what's your take?)

hbagdi: > would that make more sense?

How would that fix the problem of reading bodies by default though?

bungle: @hbagdi because it only reads bodies when it detects logout. If it is not detecting logout it does not.

samugi: just to make sure I understood, @bungle, you mean we could add a new `enable_logout` option to replace this `read_bodies`, but serving the same purpose as `read_bodies` in the current implementation, i.e.  if `enable_logout` then read `logout_post_arg`? Also, this new option needs a new entry in [removed_fields](https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua) right?

bungle: @samugi:

The options we have:
1. remove default (tried, but @hbagdi wanted new field)
2. `read_bodies` added, but was thinking, should we have `enable_logout` instead, which is disabled by default, but when enabled, it would read bodies.
4. setting actively `logout_post_arg` to `null`. This works already, but `null` is considered problematic with Yaml, CRDs, etc. (the 1. was removing default so it would be null/nil by default).

CHANGELOG.md
```diff
@@ -27,6 +27,11 @@
   `max_retry_delay` must now be `number`s greater than 0.001
   (seconds).
   [#10840](https://github.com/Kong/kong/pull/10840)
+- **Session**: a new configuration field `read_bodies` was added with a default value of `false`,
+  that changes behavior of `logout_post_arg` in a way that it is not anymore considered if the
+  `read_bodies` is not explicitly set to `true`. This is to avoid session plugin from reading
+  request bodies by default on e.g. `POST` request for logout detection.
+  [#10333](https://github.com/Kong/kong/pull/10333)
```
hbagdi, line 34: Can we move this up into the "Breaking changes" section?

hbagdi:
I'm adding the 3.4 milestones to this patch. It seems awfully close and is a worthwhile performance change. Recommend changing the commit type to 'perf'.

hanshuebner (changes requested):
I have tried to understand the pull request description and the CHANGELOG entry, but I found both very difficult to follow.  It seems that there is consensus that the implementation makes sense, but maybe the verbiage can be improved.  Happy to jump onto a Zoom to help with this.

kikito:
I'm removing this from the milestone, we are too close to the GA date and have no people available to review/approve this in the meantime.

hanshuebner:
@hbagdi We discussed whether we should really treat this as a breaking change in the PR queue review meeting earlier and these is some consensus that it is unlikely that someone relies on the current undocumented behavior.  Thus, merging this PR should be fine, even if we document it as breaking change in the change log.  What do you think?

jschmid1:
@bungle this seems to not made any progress. Do you still think this should be targeted to 3.5?

kong/plugins/session/schema.lua
bungle: @samugi I meant that `enable_logout` will disable logout totally by default, and if you need plugin to work for logout you do `enable_logout=true`. It means that we skip the whole logout handling with this.

bungle: `enable_logout` would enable this block:
https://github.com/Kong/kong/blob/master/kong/plugins/session/access.lua#L80-L93

By default it is never executed then.

jschmid1 (changes requested):
To summarize this patch:

Currently we're reading the body for every request as `logout_post_arg` is set to a non-null default. This forces us to look into every request body by default to detect if a logout is requested.
Reading bodies can become expensive, especially when requests with large payloads are sent. To resolve this issue there are a couple of options.

1. Explicitly disable the logout functionality.

This affects a rather large group of people and doesn't really solve the issue at hand. When enabling this flag, you'd still have to set `logout_post_arg` to null to avoid body reads alltogether.

2. Disable body reads by default

This will also affect some people but the surface is smaller in comparison to (1). To get this functionality back, a user can re-enable this flag.

3. Change the default of `logout_post_arg`

This is _technically_ speaking the most straightforward fix. Changing defaults will affect other products though and should be avoided if possible (@hbagdi agrees here)


This leaves us with (2) as it has the lowest impact and achieves what this patch tries to solve.

What I'm not happy with is the name of the config flag as it doesn't reflect the scope this parameter is used in. My proposal is 

`read_body_for_logout`  (move from bodies to body, define the scope  - for logout)

jschmid1 (approved):

PR #10346, opened 2023-02-22
fix(plugin): compat issue for config sync

StarlightIbuki:
### Summary

For 3.2, we added `aws_imds_protocol_version` for `aws_lambda` plugin, and `aws_imds_protocol_version` for `zipkin` plugin, but we did not add corresponding `removed_fields` entries.

### Checklist

The Pull Request has tests -- N/A, we removed tests to `removed_fileds` as it does not mean much but causes flakiness.

There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com -- N/A, it's a bug fix

### Issue reference

Fix KAG-725

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,6 +168,8 @@
   [#10247](https://github.com/Kong/kong/pull/10247)
 - Fix an issue where 'X-Kong-Upstream-Status' cannot be emitted when response is buffered.
   [#10056](https://github.com/Kong/kong/pull/10056)
+- Fix an issue where control plane does not downgrade config for `aws_lambda` and `zipkin` for older version of data planes.
+  [#10346](https://github.com/Kong/kong/pull/10346)
 
 #### Plugins
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,13 @@
 - Bumped lua-resty-session from 4.0.2 to 4.0.3
   [#10338](https://github.com/Kong/kong/pull/10338)
 
+### Fix
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -168,6 +168,8 @@
   [#10247](https://github.com/Kong/kong/pull/10247)
 - Fix an issue where 'X-Kong-Upstream-Status' cannot be emitted when response is buffered.
   [#10056](https://github.com/Kong/kong/pull/10056)
+- Fix an issue where control plane does not downgrade config for `aws_lambda` and `zipkin` for older version of data planes.
```
chronolaw, line 171: Because 3.2.0 has been released, should we move this entry to `unreleased` section?

dndx: Unreleased means the next released version. In this case we should change the release number to `3.2.1` and mention "This version has been retracted because of known issues, please use Kong 3.2.1 instead" in the `3.2.0` section.

dndx: I think we can keep this as-is for now. We need to fix the Changelog before 3.2.1 release anyway and best place to do this is in a separate PR.

dndx (approved):

StarlightIbuki:
I've manually tested the change and it works with 3.1.1.2.

https://gist.github.com/StarlightIbuki/ee9f47617c165961e5aca2e60a23bf39

Using the compose and config it with new fields, we see the config is synced.

samugi:
Adding a comment just to confirm that the fix worked fine in my manual tests too.

PR #10353, opened 2023-02-22
feat(conf_loader): load all WASM filters from conf path

locao:
### Summary

Load WASM filters from configured path.

### Checklist

### Full changelog

- load wasm filters from configured path.
- do not expect individual files.
- added new directive to template.

### Issue reference

KAG-500

### DIFF
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1819,3 +1819,5 @@
                                   # a string that identifies the plugin causing the
                                   # problem.  Queues are used by the http-log, statsd,
                                   # opentelemetry and datadog plugins.
+
+#wasm_filters_path=               # Path to WASM filters that Kong will load.
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1820,5 +1820,8 @@
                                   # problem.  Queues are used by the http-log, statsd,
                                   # opentelemetry and datadog plugins.
 
-#wasm_filters_path=               # Path to the directory containing WASM filters
+#wasm = off                       # Use this setting to enable WasmX, this allows running
+                                  # Wasm filters to process request data.
+
+#wasm_filters_path=               # Path to the directory containing Wasm filters
                                   # that Kong must load on startup.
--- a/kong.conf.default
+++ b/kong.conf.default
@@ -1820,4 +1820,5 @@
                                   # problem.  Queues are used by the http-log, statsd,
                                   # opentelemetry and datadog plugins.
 
-#wasm_filters_path=               # Path to WASM filters that Kong will load.
+#wasm_filters_path=               # Path to the directory containing WASM filters
+                                  # that Kong must load on startup.
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -1985,10 +1985,14 @@ local function load(path, custom_conf, opts)
         filter_files[pathname] = pathname
 
         local extension = pl_path.extension(entry)
-        insert(wasm_modules, {
-          name = entry:sub(0, -#extension - 1),
-          path = pathname,
-        })
+        if string.lower(extension) == ".wasm" then
+          insert(wasm_modules, {
+            name = entry:sub(0, -#extension - 1),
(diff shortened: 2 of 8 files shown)


kong.conf.default
```diff
@@ -1819,3 +1819,5 @@
                                   # a string that identifies the plugin causing the
                                   # problem.  Queues are used by the http-log, statsd,
                                   # opentelemetry and datadog plugins.
+
+#wasm_filters_path=               # Path to WASM filters that Kong will load.
```
hbagdi, line 1823:
```suggestion
#wasm_filters_path=               # Path to the directory containing WASM filters
                                  # that Kong must load on startup.
```

thibaultcha:
The tech preview branch had the `wasm = on/off` flag that superseded the `wasm_modules` property, but this PR removed it to fully depend on whether or not `.wasm` files are present in the directory we point to. I guess the question is do we want to enable wasm support only depending on whether or not `.wasm` bytecode is detected, or do we want a boolean flag to enable on top of that first condition. Personally I have no opinion, only wanted to clarify what we have and what we can do.

locao:
> disable WASM (for any reason) without removing the .wasm bytecode files, which may be baked into the container image

Having the wasm files baked in the container image was something I wasn't considering. So, yes, it makes sense to have a boolean flag.

kong/conf_loader/init.lua
```diff
@@ -581,7 +581,7 @@ local CONF_PARSERS = {
   max_queued_batches = { typ = "number" },
 
   wasm = { typ = "boolean" },
-  wasm_modules = { typ = "array" },
+  wasm_filters_path = { typ = "string" },
```
flrgh, line 584: Since we're leaning towards using the boolean `wasm = on|off` setting to toggle wasm, would it make sense to remove `wasm_filters_path` as a setting and use a standardized location (i.e. `{{KONG_PREFIX}}/wasm_filters`) instead?

We can always add `wasm_filters_path` later on if there is a compelling case to use a non-default location.

@locao what do you think?

locao: I don't have a strong opinion on this. I like a lot of knobs to adjust even if they were never touched, but that's not how a good product is built 😅 

For the sake of keeping things simpler, I'd say let's try without `wasm_filters_path` and see if we miss using it during development phase.

kong.conf.default
```diff
@@ -1819,3 +1819,6 @@
                                   # a string that identifies the plugin causing the
                                   # problem.  Queues are used by the http-log, statsd,
                                   # opentelemetry and datadog plugins.
+
+#wasm_filters_path=               # Path to the directory containing WASM filters
+                                  # that Kong must load on startup.
```
hbagdi, line 1824: Please document the 'wasm' configuration option.

kong/conf_loader/init.lua
```diff
@@ -1228,17 +1228,10 @@ local function check_and_parse(conf, opts)
...
-      local extension = pl_path.extension(path)
-
-      if not pl_path.isfile(path)
-         or not pl_path.exists(path)
-         or extension ~= ".wasm"
-      then
-        errors[#errors + 1] = fmt("wasm_module at '%s' not .wasm or " ..
-                                  "does not exists", path)
-      end
+  if conf.wasm and conf.wasm_filters_path then
```
hbagdi, line 1231: Can we split these into nested if blocks or better create a new "validate_wasm" to limit the length of this already long function?

kong/conf_loader/init.lua
flrgh: I can imagine it being kind of annoying for our integration tests if the path is not configurable, _but_ in my opinion this gives us the incentive to make sure symlinks are supported. Ideally something like this should be supported out-of-the box:

```sh
ln -s \
  /path/to/git/kong/spec/fixtures/wasm/a/b/c/d/my_filter.wasm \
  /path/to/my_kong_prefix/wasm_filters/my_filter.wasm
```

PR #10355, opened 2023-02-23
feat(balancer): add KONG_UPSTREAM_DNS_TIME ctx

attenuation:
### Summary
add KONG_UPSTREAM_DNS_TIME to `ngx.ctx`, So kong plugin can get upstream DNS resolve time in log phase.

### Checklist

### Issue reference

FTI-4826


kong/runloop/balancer/targets.lua
```diff
@@ -511,15 +512,21 @@ local function targetExpired(target)
   return not target.lastQuery or target.lastQuery.expire < ngx_now()
 end
 
+local function get_updated_now_ms()
+  update_time()
+  return ngx_now() * 1000 -- time is kept in seconds with millisecond resolution.
+end
```
sabertobihwy, line 518: Just a small suggestion, this function appears twice, could it be placed in a module like tools or utils.

kong/tools/utils.lua
```diff
@@ -37,6 +37,8 @@ local re_match      = ngx.re.match
 local inflate_gzip  = zlib.inflateGzip
 local deflate_gzip  = zlib.deflateGzip
 local stringio_open = pl_stringio.open
+local now           = ngx.now
+local update_time   = ngx.update_time
```
chronolaw, line 41: These two lines are only used by `get_updated_now_ms`, 
we should move these two lines into a `do end` block,
like `sha256_hex` did.

hbagdi:
@locao understands balancer and must sign off to make sure that the latency is being measured correctly.
We also need a test that puts Kong in a stat where it performs a synchronous DNS resolution while proxying a request and we observe the appropriate latency in here.

We want to capture this metric in the log-serializer and then add a test to make sure that this is emitted in `http-log` and other metric plugins.

locao:
I labeled this PR as "discussion" so we can quickly talk about it offline before merging.

hbagdi:
When and where is this discussion going to happen?

locao:
> When and where is this discussion going to happen?

It happened in FTI-4826. I think we are good to continue.

locao:
I converted this to draft PR, as the comments in FTI-4826 are still to be addressed.

kong/init.lua
```diff
@@ -1279,6 +1273,10 @@ do
       ctx.KONG_PROXY_LATENCY = ctx.KONG_RESPONSE_START - ctx.KONG_PROCESSING_START
     end
 
+    if not ctx.KONG_UPSTREAM_DNS_TIME and ctx.KONG_UPSTREAM_DNS_END and ctx.KONG_UPSTREAM_DNS_START then
+      ctx.KONG_UPSTREAM_DNS_TIME = ctx.KONG_UPSTREAM_DNS_END - ctx.KONG_UPSTREAM_DNS_START
+    end
```
locao, line 1278:
```suggestion
    else
      ctx.KONG_UPSTREAM_DNS_TIME = 0
    end
```
Let's keep the metric always present, even when we get a cache hit.

locao (changes requested):
Can you add some testing, please?

hbagdi:
ping @locao for another round

hanshuebner (dismissed):
@attenuation Please update the pull request description to include an actual description and make sure that all checklist items are addressed.

locao (changes requested):
@attenuation please, add some tests to validate the new feature, a changelog entry and fix the merge conflict.

hanshuebner (commented):
Why is a new function introduced that works with non-standard millisecond resolution?  Is it a requirement that `KONG_UPSTREAM_DNS_TIME` is presented in milliseconds?

All the extra localization makes the change much bigger than it needs to be, with questionable benefit.

attenuation:
> Why is a new function introduced that works with non-standard millisecond resolution? Is it a requirement that `KONG_UPSTREAM_DNS_TIME` is presented in milliseconds?
> 
> All the extra localization makes the change much bigger than it needs to be, with questionable benefit.

We need to get accurate time, and the time returned by `ngx.now()` is cached by Nginx, which is refreshed each time an epoll event loop is waited. As we need to get the time during the connection, we have to call `ngx.update_time` to refresh the cached time.
Also, I encapsulated this logic in a function since it's being used in many places, and there is no problem with that.

locao:
> Why is a new function introduced that works with non-standard millisecond resolution? Is it a requirement that `KONG_UPSTREAM_DNS_TIME` is presented in milliseconds?

That's not a new function, it was moved to `utils` to be used by the load-balancer code. Besides that, all other latencies are measured in ms (using the same function), so I think it's right to keep the same resolution.

locao (approved):
This change seems good to go. Good job!

I'll just wait for the tests to go green before merging.

PR #10364, opened 2023-02-24
fix(tracing): trace hook error out when see nil

StarlightIbuki:
### Summary

"instrument" hooks `toip`, `db_connect` and other functions that widely used by core and plugins, and collect data for tracing.

I surprisingly find the hooking function will error out when `set_attribute` got a nil value.  For example, A failing DNS query would emit an error with traceback: "invalid value".

This will definitely change the behavior of the hooked function. `toip` would have returned `(nil, err)` if it's not hooked, but instead, it throws an error and the error has nothing to do with the origin error. This could cause more serious issues for other hooked functions.

Luckily, we do not have more cases where `set_attributes` receives invalid values.

The complete list of hooked functions will be shown offline.

### Checklist

N/A There is a user-facing docs PR against https://github.com/Kong/docs.konghq.com - PUT DOCS PR HERE

### Full changelog

* Change "error" to "log".

### Issue reference

KAG-749


fffonion:
seems the failing test is related

StarlightIbuki:
> seems the failing test is related

It is. But let's wait until the 3.2 released

spec/01-unit/26-tracing/01-tracer_pdk_spec.lua
```diff
@@ -138,16 +178,21 @@ describe("Tracer PDK", function()
...
     end)
 
     it("fails add_event", function ()
       local span = c_tracer.start_span("meow")
-      assert.error(function() span:set_attribute("key1") end)
-      assert.error(function() span:set_attribute("key1", function() end) end)
-      assert.error(function() span:set_attribute(123, 123) end)
+      assert.error(function() span:add_event("key1", 123) end)
+      assert.error(function() span:add_event("key1", function() end) end)
+      assert.error(function() span:add_event(123, {}) end)
```
StarlightIbuki, line 195: This seems a mistake in the original test. I'm fixing this by the way.

kong/globalpatches.lua
```diff
@@ -467,6 +467,9 @@ return function(options)
       -- to "all".
       local instrumentation = require "kong.tracing.instrumentation"
       instrumentation.set_patch_dns_query_fn(toip, function(wrap)
+        -- TODO: this call has no side effect.
+        -- maybe client.toip = wrap?
+        -- but that doesn't work to client = require("kong.tools.dns")() case
```
StarlightIbuki, line 472:
```suggestion
```

It has a side effect. The function exported uses this local variable as its up value.

kong/pdk/tracing.lua
```diff
@@ -253,13 +255,18 @@ end
 -- span:set_attribute("net.peer.port", 443)
 -- span:set_attribute("exception.escaped", true)
 function span_mt:set_attribute(key, value)
+  -- key is decided by the programmer, so if it is not a string, we should
```
fffonion, line 258:
```suggestion
  -- key is a user input, so if it is not a string, we should
```

CHANGELOG.md
```diff
@@ -84,6 +84,8 @@
   [#10352](https://github.com/Kong/kong/pull/10352)
 - Fix an issue where validation to regex routes may be skipped when the old-fashioned config is used for DB-less Kong.
   [#10348](https://github.com/Kong/kong/pull/10348)
+- FIx and issue where tracing may cause unexpected behavior.
```
fffonion, line 87:
```suggestion
- Fix and issue where tracing may cause unexpected behavior.
```

kong/pdk/tracing.lua
StarlightIbuki: It's not user input, that's why we can error out without worrying this would fail requests.

spec/01-unit/26-tracing/01-tracer_pdk_spec.lua
fffonion: i'm bit confused on why this change is needed, what was the `assert.error`previously tested upon?

StarlightIbuki:
I tried reverting the change and the test still fails for `t/01-pdk/08-response/11-exit.t` (39, 41, 42, 43).
The master is not red. No idea why it fails at this branch.

spec/01-unit/26-tracing/01-tracer_pdk_spec.lua
StarlightIbuki: It's not needed. But the original test is not testing what it says. I'm fixing it.

ADD-SP (approved):

PR #10365, opened 2023-02-24
fix(db): didn't consider all workspaces when doing sni name collision check

catbro666:
### Summary

As sni names should be unique across workspaces, so collision check should be done across all workspaces.

This affects admin API `certificates`. The post request may fail without reverting to the original state when a sni name already exists in another workspace.

### Checklist

### Full changelog

* fix an issue with dao API `_SNIs:check_list_is_new` which may cause the endpoint `certificates` post request to fail without reverting to the original state when a sni name already exists in another workspace.

### Issue reference

Fix FTI-4846


catbro666:
Because ce doesn't support workspace, move this PR to ee.

PR #10372, opened 2023-02-25
Revert "fix(key-auth): return unauthorized if api key includes non-ascii symbols (#10171)"

hanshuebner:
This reverts commit d10101955bc554e9e453d5968392babbdf1bf4f9 (#10171).

The fix is too restrictive on the key syntax due to a misunderstanding on my end.  Before reimplementing, we need a reproduction case that demonstrates the problem, as we may need to fix the issue at another level.

https://konghq.atlassian.net/browse/KAG-721


Tieske (approved):
lgtm

PR #10374, opened 2023-02-25
feat(proxy): error templates customization

### LINKED ISSUE #7093: Improve error handling
### Improve and allow a more consistent error handling

Hi,

The error handling in kong is somewhat inconsistent:
 - If the error come from nginx and is catched by the directive defined in the [nginx configuration](https://github.com/Kong/kong/blob/4da9f0a0b1c5d0d7905f12d3fb244950980affdd/kong/templates/nginx_kong.lua#L77-L78) it will trigger the [error handler](https://github.com/Kong/kong/blob/4da9f0a0b1c5d0d7905f12d3fb244950980affdd/kong/error_handlers.lua#L31-L63) response to the client (that will render body depending on mime types of the request)
- If the error come from nginx and is not catched by the directive above, it will trigger the default nginx error page
- If the error come from kong code, as in [these](https://github.com/Kong/kong/blob/4da9f0a0b1c5d0d7905f12d3fb244950980affdd/kong/init.lua#L272-L275) [examples](https://github.com/Kong/kong/blob/4da9f0a0b1c5d0d7905f12d3fb244950980affdd/kong/init.lua#L777) it will generally just return a json response, no matter what the mime type of the request was...
- IF the error come from `kong.response.error` call, it will respond with a body build depending on mime type

When kong is a installed as a gateway that serve html along with json API's, these json errors are perceived like a bug...

Furthermore, when the errors are built depending on the mime type of the request, it use some [hardcoded templates](https://github.com/Kong/kong/blob/4da9f0a0b1c5d0d7905f12d3fb244950980affdd/kong/tools/utils.lua#L1275-L1301). I think it will be good if the kong user can override these templates to allow customisation.

That cause us some difficulties as well, when kong serve website that return HTML and do not allow to serve a branded error page.

I tried to fix this in [this PR](https://github.com/Kong/kong/pull/7087/files) (still draft) by systematically calling to `kong.response.error` function in kong code to allow consistent errors response and also added a error template file configuration to allow error template override.

I would like to receive some feedback on this

Thanks !

samugi:
### Summary

Allow customizing error templates via kong.configuration parameters

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

Fix #7093

[KAG-723](https://konghq.atlassian.net/browse/KAG-723)
[KAG-751]

[KAG-751]: https://konghq.atlassian.net/browse/KAG-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[KAG-723]: https://konghq.atlassian.net/browse/KAG-723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kong/conf_loader/init.lua
+++ b/kong/conf_loader/init.lua
@@ -577,6 +577,11 @@ local CONF_PARSERS = {
 
   proxy_server = { typ = "string" },
   proxy_server_ssl_verify = { typ = "boolean" },
+
+  error_template_html = { typ = "string" },
+  error_template_json = { typ = "string" },
+  error_template_xml = { typ = "string" },
+  error_template_plain = { typ = "string" },
 }
 
 
--- a/kong/pdk/response.lua
+++ b/kong/pdk/response.lua
@@ -26,9 +26,7 @@ local find = string.find
 local lower = string.lower
 local error = error
 local pairs = pairs
-local ipairs = ipairs
 local concat = table.concat
-local tonumber = tonumber
 local coroutine = coroutine
 local cjson_encode = cjson.encode
 local normalize_header = checks.normalize_header
@@ -36,7 +34,6 @@ local normalize_multi_header = checks.normalize_multi_header
 local validate_header = checks.validate_header
 local validate_headers = checks.validate_headers
 local check_phase = phase_checker.check
-local split = utils.split
 local add_header
 local is_http_subsystem = ngx and ngx.config.subsystem == "http"
 if is_http_subsystem then
@@ -1061,39 +1058,6 @@ end
   end
 
 
-  local function get_response_type(content_header)
-    local type = CONTENT_TYPE_JSON
-
-    if content_header ~= nil then
-      local accept_values = split(content_header, ",")
-      local max_quality = 0
-      for _, value in ipairs(accept_values) do
-        local mimetype_values = split(value, ";")
-        local name
-        local quality = 1
-        for _, entry in ipairs(mimetype_values) do
-          local m = ngx.re.match(entry, [[^\s*(\S+\/\S+)\s*$]], "ajo")
-          if m then
-            name = m[1]
-          else
-            m = ngx.re.match(entry, [[^\s*q=([0-9]*[\.][0-9]+)\s*$]], "ajoi")
-            if m then
-              quality = tonumber(m[1])
-            end
-          end
-        end
-
-        if name and quality > max_quality then
-          type = utils.get_mime_type(name)
-          max_quality = quality
-        end
-      end
-    end
-
-    return type
-  end
-
-
   ---
   -- This function interrupts the current processing and produces an error
   -- response.
@@ -1195,11 +1159,8 @@ end
       if is_grpc_request() then
         content_type = CONTENT_TYPE_GRPC
       else
-        content_type_header = ngx.req.get_headers()[ACCEPT_NAME]
-        if type(content_type_header) == "table" then
-          content_type_header = content_type_header[1]
-        end
-        content_type = get_response_type(content_type_header)
+        local accept_header = ngx.req.get_headers()[ACCEPT_NAME]
+        content_type = utils.get_response_type(accept_header)
       end
     end
 
--- a/kong/tools/utils.lua
+++ b/kong/tools/utils.lua
@@ -14,6 +14,7 @@ local pl_stringx = require "pl.stringx"
 local pl_stringio = require "pl.stringio"
 local pl_utils = require "pl.utils"
 local pl_path = require "pl.path"
+local pl_file = require "pl.file"
 local zlib = require "ffi-zlib"
 
 local C             = ffi.C
@@ -1295,6 +1296,34 @@ do
 ]],
   }
 
+  local ngx_log = ngx.log
+  local ERR     = ngx.ERR
+  local custom_error_templates = setmetatable({}, {
+    __index = function(self, format)
+      local template_path = kong.configuration["error_template_" .. format]
+      if not template_path then
+        rawset(self, format, false)
+        return false
+      end
+
+      local template, err
+      if pl_path.exists(template_path) then
+        template, err = pl_file.read(template_path)
+      else
+        err = "file not found"
+      end
+
+      if template then
+        rawset(self, format, template)
+        return template
+      end
+
+      ngx_log(ERR, fmt("failed reading the custom %s error template: %s", format, err))
+      rawset(self, format, false)
+      return false
+    end
+  })
+
   get_mime_type = function(content_header, use_default)
     use_default = use_default == nil or use_default
     content_header = _M.strip(content_header)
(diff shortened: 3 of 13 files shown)


kong/tools/utils.lua
```diff
@@ -1295,6 +1296,36 @@ do
 ]],
   }
 
+  local ngx_log = ngx.log
+  local ERR     = ngx.ERR
+  local custom_error_templates = setmetatable({}, {
+    __index = function(self, format)
+      local template_path_k = "error_template_" .. format
+      local template_path = kong.configuration[template_path_k]
```
gszr, line 1304: PDK tests are failing because those tests only initialize the kong global (and contents of it) on demand. See [this](https://github.com/Kong/kong/blob/a9b3f5deff0b51897091dcbea4d4eeb1d77be73c/t/01-pdk/05-client/08-get_protocol.t#L269) for example.

samugi: thanks @gszr! I had left the tests running and didn't notice the failure yet

kong/pdk/response.lua
```diff
@@ -1196,10 +1160,7 @@ end
         content_type = CONTENT_TYPE_GRPC
       else
         content_type_header = ngx.req.get_headers()[ACCEPT_NAME]
```
hanshuebner, line 1162: `content_type_header` should be named `accept_header`

kong/tools/utils.lua
```diff
@@ -1295,6 +1298,71 @@ do
...
+  get_response_type = function(accept_header)
+    local content_type = MIME_TYPES[CONTENT_TYPE_DEFAULT]
+    if type(accept_header) == "table" then
+      accept_header = join(",", accept_header)
+    end
+
+    if accept_header ~= nil then
+      local accept_values = split(accept_header, ",")
+      local max_quality = 0
+      for _, value in ipairs(accept_values) do
```
hanshuebner, line 1339: I would use one regular expression to parse each accept value, after stripping any white space from the beginning and the end.  According to the syntax of the accept header defined in [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2), the following regular expression should be usable to parse the type, subtype and q-value out of any accept header:
```
([a-z0-9-!#$&^_+.]+|\*)\/([a-z0-9-!#$&^_+.]+|\*)(?:;q=(1(?:\.0{0,3}|)|0(?:\.\d{0,3}|))|)
```
I am certain that some would object to the expression as being pretty complex, but a manually written parser would probably be a lot of code and not be easier to review.

This regular expression parser would also be more correct in that it'd parse quality values `0` and `1`, which would not be parsed by yours.

I used https://regexr.com/ to develop this expression and found it pretty nifty and helpful.

hanshuebner (changes requested):

kong/tools/utils.lua
hanshuebner: While you're at it, you could also implement the wildcard for `*` to match any type or subtype that we have content for.

samugi: we might need to tune the regex a little to match q-values even when a `level` property is defined, e.g. `text/html;level=1` --> `([a-z0-9][a-z0-9-!#$&^_+.]+|\*)\/([a-z0-9][a-z0-9-!#$&^_+.]+|\*)(?:;level=\d|)(?:;q=(1(?:\.0{0,3}|)|0(?:\.\d{0,3}|))|)`. I'm analyzing different scenarios to ensure this is compatible with the code we currently have [in the response module](https://github.com/Kong/kong/blob/dbe8d947e7d6a6c12f8464fc2f08276be1cd392f/kong/pdk/response.lua#L1064-L1094) and cause no regressions.

hanshuebner: Please note this section from RFC7231:

> Each media-range might be followed by zero or more applicable media
> type parameters (e.g., charset), an optional "q" parameter for
> indicating a relative weight ([Section 5.3.1](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.1)), and then zero or more
> extension parameters.  The "q" parameter is necessary if any
> extensions (accept-ext) are present, since it acts as a separator
> between the two parameter sets.

This means that an `Accept` header with `text/html;level=2;q=0.4` would not be conforming to the spec.  In that case, the regular expression would not see the `q` value, which I think would be acceptable.

samugi: after discussing this offline we agreed to adapt the regex pattern to support said parameters.
I updated the code with the new pattern, with a very small variation where I used a single capturing group to get the media type instead of splitting type/subtype. I did it because `utils.get_mime_type` expects the media_type to be passed as a single string so there should be no need to capture type and subtype separately. The wildcard for subtype matching is already implemented in `get_mime_type` based on that input.

kong/tools/utils.lua
```diff
@@ -1295,6 +1298,74 @@ do
...
+    if accept_header ~= nil then
+      local token_pattern = "[a-z0-9][a-z0-9-!#$&^_+.]+"
+      local pattern = string.format([[
+        ((?:%s|\*) \/ (?:%s|\*))
+        (?:
+          \s*;\s*
+          q = ( 1(?:\.0{0,3}|) | 0(?:\.\d{0,3}|) )
+          | \s*;\s* %s (?:=[^;]*|)
+        )*
+      ]], token_pattern, token_pattern, token_pattern)
```
hanshuebner, line 1345: I'm not a huge fan of putting the token pattern into a variable and then interpolating because it mixes two languages, making it overall harder to understand what's going on.  If you must do it, then I'd recommend doing it like this:
```
local pattern = string.gsub([[
        ((?:TOKEN|\*) \/ (?:TOKEN|\*))
        (?:
          \s*;\s*
          q = ( 1(?:\.0{0,3}|) | 0(?:\.\d{0,3}|) )
          | \s*;\s* TOKEN (?:=[^;]*|)
        )*
      ]], "TOKEN", "[a-z0-9][a-z0-9-!#$&^_+.]+")
```
That way, it is easier to distinguish the TOKEN replacement string from `\s`, and it also does not require counting.  But I'd just leave the token pattern inside of the main regular expression.

In any case, you want to push the possible assembly and compilation of the regular expression into the global scope as there is no need to do it for every invocation of `get_response_type`, and the performance gang would immediately complain.

hanshuebner (changes requested):

kong/tools/utils.lua
samugi: > you want to push the possible assembly and compilation of the regular expression into the global scope as there is no need to do it for every invocation of get_response_type

since we are using the option `o` we should be ok I believe: _the regex pattern in question will only be compiled once, cached in the worker-process level, and shared among all requests in the current Nginx worker process_ see: https://github.com/openresty/lua-nginx-module#ngxrematch

hanshuebner (approved):

kong/conf_loader/init.lua
```diff
@@ -577,6 +577,11 @@ local CONF_PARSERS = {
 
   proxy_server = { typ = "string" },
   proxy_server_ssl_verify = { typ = "boolean" },
+
+  error_template_html = { typ = "string" },
+  error_template_json = { typ = "string" },
+  error_template_xml = { typ = "string" },
+  error_template_plain = { typ = "string" },
```
esatterwhite, line 584: @hanshuebner 
What data gets passed to these templates?

samugi: @esatterwhite the templates are rendered by passing them _the error message_ as a string, see [here](https://github.com/Kong/kong/blob/fa90a685b5a7ac03774460a92f856293184d704d/kong/error_handlers.lua#L67) for example. We will provide proper documentation before this is released.

PR #10379, opened 2023-02-27
feat(otel): add http_response_header_for_traceid config to add trace_id for downstream header

attenuation:
### Summary

add http_response_header_for_traceid config to add trace_id to the downstream header.

### Checklist

### Full changelog

* [Implement ...]

### Issue reference

FTI-4655

### DIFF
--- a/kong/clustering/compat/removed_fields.lua
+++ b/kong/clustering/compat/removed_fields.lua
@@ -62,7 +62,7 @@ return {
     },
     proxy_cache = {
       "ignore_uri_case",
-    }
+    },
     opentelemetry = {
       "http_response_header_for_traceid",
     },
--- a/kong/clustering/compat/removed_fields.lua
+++ b/kong/clustering/compat/removed_fields.lua
@@ -63,5 +63,8 @@ return {
     proxy_cache = {
       "ignore_uri_case",
     }
+    opentelemetry = {
+      "http_response_header_for_traceid",
+    },
   },
 }
--- a/kong/plugins/opentelemetry/handler.lua
+++ b/kong/plugins/opentelemetry/handler.lua
@@ -21,7 +21,6 @@ local null = ngx.null
 local encode_traces = otlp.encode_traces
 local translate_span_trace_id = otlp.translate_span
 local encode_span = otlp.transform_span
-local ctx = ngx.ctx
 
 local _log_prefix = "[otel] "
 
@@ -153,7 +152,7 @@ end
 
 function OpenTelemetryHandler:header_filter(conf)
   if conf.http_response_header_for_traceid then
-    local trace_id = ctx.plugin.trace_id
+    local trace_id = kong.ctx.plugin.trace_id
     if not trace_id then
       local root_span = ngx.ctx.KONG_SPANS and ngx.ctx.KONG_SPANS[1]
       trace_id = root_span and root_span.trace_id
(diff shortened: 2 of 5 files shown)


kong/plugins/opentelemetry/handler.lua
```diff
@@ -150,6 +150,17 @@ function OpenTelemetryHandler:access()
   propagation_set("preserve", header_type, translate_span_trace_id(root_span), "w3c")
 end
 
+function OpenTelemetryHandler:header_filter(conf)
+  if conf.http_response_header_for_traceid then
+    local trace_id = kong.ctx.plugin.trace_id
+    if not trace_id then
+      local root_span = ngx.ctx.KONG_SPANS and ngx.ctx.KONG_SPANS[1]
+      trace_id = root_span and root_span.trace_id
+    end
```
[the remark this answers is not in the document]
attenuation: if trace_id is still nil, kong.response.add_header can cover nil value. so we do not consider this.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -150,6 +150,17 @@ function OpenTelemetryHandler:access()
...
+    kong.response.add_header(conf.http_response_header_for_traceid, trace_id)
+  end
```
ms2008, line 161:
```suggestion
  if not conf.http_response_header_for_traceid then
    return
  end
  
  local trace_id = kong.ctx.plugin.trace_id
  if not trace_id then
    local root_span = ngx.ctx.KONG_SPANS and ngx.ctx.KONG_SPANS[1]
    trace_id = root_span and root_span.trace_id
  end
  kong.response.add_header(conf.http_response_header_for_traceid, trace_id)
```

kong/plugins/opentelemetry/schema.lua
```diff
@@ -49,6 +49,7 @@ return {
         { connect_timeout = typedefs.timeout { default = 1000 } },
         { send_timeout = typedefs.timeout { default = 5000 } },
         { read_timeout = typedefs.timeout { default = 5000 } },
+        { http_response_header_for_traceid = { type = "string", default = nil }},
```
fffonion, line 52: how about `traceid_response_header`?

fffonion: nvm, i saw this is also the name used in zipkin; let's be consistent then

kong/plugins/opentelemetry/handler.lua
attenuation: I think we can not do this, because the header filter may have more capabilities in the future, if add `return` early, it may be hard to extension this function.

ms2008: Code should reduce nesting where possible by handling error cases/special conditions first and returning early or continuing the loop. Reduce the amount of code that is nested multiple levels.

Ref: https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

ms2008: I suggest doing this, and I think it's a good habit to get into. If you don't think it's necessary, it's OK.😉

kong/clustering/compat/removed_fields.lua
```diff
@@ -59,6 +59,9 @@ return {
   [3003000000] = {
     acme = {
       "account_key",
-    }
+    },
+    statsd = {
```
GGabriele, line 63:
```suggestion
    opentelemetry = {
```

ms2008 (approved):

ms2008 (commented):
Please rebase from the master, and resolve the new conflicts.

PR #10382, opened 2023-02-27
feat(key-sets): allow remote sourcing of JWKs 

jschmid1:
### Summary
Be able store a URL, pointing to a resource with `application/jwk-set+json`, in a `key-set` instance to load JSON Web Keys (as defined in https://www.rfc-editor.org/rfc/rfc7517#section-5) to a `key` entity.

Example:

``` shell
http :8001/key-sets name=google jwks_url=https://www.googleapis.com/oauth2/v3/certs

{
    "created_at": 1677147024,
    "id": "ed6d2e01-1e38-4e86-a7ef-fc1d7de5b472",
    "jwks_url": "https://www.googleapis.com/oauth2/v3/certs",
    "name": "google",
    "tags": null,
    "updated_at": 1677147024
}

# this should have triggered a key retrieval from the specified `jwks_url`

http  :8001/key-sets/google/keys


{
    "data": [
        {
            "created_at": 1677147024,
            "id": "88d18f0e-014f-4b75-b730-8168aa5fb967",
            "jwk": ".."
            "name": null,
            "pem": null,
            "set": {
                "id": "ed6d2e01-1e38-4e86-a7ef-fc1d7de5b472"
            },
            "tags": null,
            "updated_at": 1677147024
        },
        {
            "created_at": 1677147024,
            "id": "9138ac51-d3d6-4efa-84e8-979766b209ef",
            "jwk": "..."
            "kid": "d25f8dbcf97dc7ec401f0171fb6e6bda9ed9e792",
            "name": null,
            "pem": null,
            "set": {
                "id": "ed6d2e01-1e38-4e86-a7ef-fc1d7de5b472"
            },
            "tags": null,
            "updated_at": 1677147024
        }
    ],
    "next": null
}
```

Keys will

* inherit the `kid` from the remote key-set. 
* not have a name

Keys must be rejected when they

* don't match the expected format
* can't be loaded by our openssl lib
* there is no `kid`

## Rotation

Keys and key-sets behind a remote resource can be subject to change. As there is no canonical way  in Kong to get notified about remote content change, we have to support time/schedule based key-rotation. 

Keys will only be updated when

* the `JWK` (identified by `kid`) has changed in the remote location
	* or does not yet exist

Rotation is a process that needs to be kicked off when:

* Kong starts (starts a timer with a 10 minute cooldown)
* A key-set was
	* created
	* updated
	* deleted

Cache invalidation events need to be emitted in these scenarios. This is handled by the entity for all CRUD events)

### Interfaces

The timeframe to trigger key-rotation is relatively large(10m). Within Kong, we need mechanisms to interface with this feature to forefully trigger a key rotation.

#### Plugins

Within plugins we can use the `dao` to trigger the `rotate` function

``` lua
-- kong/plugin/example/handler.lua

local key_set = kong.db.key_sets:select("key-set name or ID you get from the plugin's configuration")

locak rotate_ok, rotate_err = kong.db.key_sets:rotate(key_set)
--- .. 

```

#### Admin Api

If a user is aware that the content in the remote resource has changed, he can kick-off a `rotate` operation by targeting the respective key-set

``` shell
http :8001/key-sets/:key-set/rotate
{
    "message": "ok"
}
```

### FAQ:

* What happens with keys that are associated to a key-set but can't be found on a remote URL?

A) Remove 
B) Keep 

(B). A user may want to mix and match remote keys with self-issues keys. There can be a partial outage of the remote server where it will appear healty (status->200) but does not deliver content. In this case we should not remove keys.

* Do we keep previously rotated keys?

No, keys are separate, loosely coupled, entities that have no notion of `previous` or `current`.  A successful rotation will only update existion keys (detected by their `kid` attribute). There is no "hot swap" where full key-sets are flipped from `previous->current`. 

* When _new_ appear in the resource will they be included automatically?

Yes.

### Checklist

- [ ] docs -> https://github.com/Kong/docs.konghq.com/pull/5221

### Issue reference

Fix: KAG-718
Design: https://docs.google.com/document/d/1u6LoKuFZUyFshtsMQ0dHABzj27UmWwIaxWmkpxTPlMA/edit?usp=sharing


hbagdi:
This requires a small one-pager design document since this has product usage implications as well as will lead to additional work in downstream projects based on Kong. Please do not merge the PR until we have this in place. Most of the content from the PR description can be reused.


> this should have triggered a key retrieval from the specified `jwks_url`


What will happen if the Control-Plane is in a restricted network environment where it cannot resolve these URLs while a Data-Plane can?
That is very much the assumption that was baked into the design of secrets management as well and that assumption should remain true in this case as well.

jschmid1:
@hbagdi 
> This requires a small one-pager design document since this has product usage implications as well as will lead to additional work in downstream projects based on Kong. Please do not merge the PR until we have this in place. Most of the content from the PR description can be reused.

done, link in the description.

jschmid1:
> What will happen if the Control-Plane is in a restricted network environment where it cannot resolve these URLs while a Data-Plane can?

The current implementation would raise an error at creation-time when the specified URL is not reachable. Keys are directly loaded from the endpoint mainly to prevent misconfiguration and to provide a quicker feedback loop. 

> That is very much the assumption that was baked into the design of secrets management as well and that assumption should remain true in this case as well.

If this _is_ an requirement I'll need to make further changes. However, we should be aware that this _can_ hide misconfiguration  and makes this process less insightful as you don't get immediate feedback.

hbagdi:
> If this is an requirement I'll need to make further changes. However, we should be aware that this can hide misconfiguration and makes this process less insightful as you don't get immediate feedback.


I see the problem but I don't have a solution. Do you?
I know many installations where CP is running in a completely different environment than DP and something like this is going to fail.

jschmid1:
> I see the problem but I don't have a solution. Do you?

There are the obvious paths like controlling the behavior when keys can't be loaded on the CP with configuration parameters. ("skip_initial_key_imports" or something along those lines). There may be better alternatives.. I'll keep digging

> I know many installations where CP is running in a completely different environment than DP and something like this is going to fail.

I guess it's reasonable to block egress from the CP but not from the DP.

kong/db/dao/key_sets.lua
```diff
@@ -27,45 +157,72 @@ function key_sets:each(size, options)
   return self.super.each(self, size, options)
 end
 
+function key_sets:start_rotation_timer()
+  kong.log.notice("starting key rotation timer")
+  return timer_every(ROTATION_TIMER, self.rotate_all)
```
bungle, line 162: Should we use named one instead:
https://github.com/Kong/lua-resty-timer-ng#named_every

hbagdi:
There are no plans to move forward on this currently so closing it. The linked Jira is tracking the work.

PR #10383, opened 2023-02-27
[JK] Opentelemetry  with K8s collector support

jaswanthikolla:
### Opentelemety with K8s collector support

Opentelemetry doesn't support reaching the collector running on the KubeNode with host IP address.  We need ability to pass in the environment variable as well part from DNS name.

### Full changelog
It adds environment support to Opentelemetry config.


kong/plugins/opentelemetry/handler.lua
```diff
@@ -78,17 +78,27 @@ local function http_export_request(conf, pb_data, headers)
...
+local parsed_endpoint = nil
+
 local function http_export(conf, spans)
   local start = ngx_now()
   local headers = get_cached_headers(conf.headers)
   local payload = encode_traces(spans, conf.resource_attributes)
 
-  local ok, err = http_export_request(conf, payload, headers)
+  -- caching because comparing string is much faster than applying regex
+  if conf.endoint ~= prev_endpoint then
```
jaswanthikolla, line 90: Is there a race condition with using variables like this?

slavaPeloton, line 90: Typo:
conf.endoint -> conf.endpoint

hbagdi:
Datadog plugin had the same issue and it was addressed in these two commits: https://github.com/kong/kong/commit/19369e1c2 https://github.com/kong/kong/commit/bd85d8c88.
Is it possible to implement the same solution here?

hbagdi:
xref https://github.com/Kong/kong/pull/9466

@Tieske Given the hostility of user-experience will `null` values, especially with downstream projects like decK, KIC, etc, it would be worthwhile to explore an explicit configuration option (like the `use_env` approach that you removed in the above-referenced commit).
Configuring a detail like this reliably would go a long way, especially with the ever-expanding user base that runs kong on k8s.

jaswanthikolla:
> Datadog plugin had the same issue and it was addressed in these two commits: [19369e1c2](https://github.com/Kong/kong/commit/19369e1c2) [bd85d8c88](https://github.com/Kong/kong/commit/bd85d8c88).
> Is it possible to implement the same solution here?

There are 2 different use cases. Here we have full URL, where in datadog we just have host and port.  I think it get's complicated on what scenario environment variables are used. Implictly telling, use environment variable when the config values are `nil` is not really great way to expose a plugin config.  IMO, It should be explicit and simple at the same time.

I think a simple configuration with `http://${KONG_OT_COLLECTOR_IP}:8444/traces` is much more user friendly not error prone.

kong/plugins/opentelemetry/handler.lua
```diff
@@ -78,17 +78,28 @@ local function http_export_request(conf, pb_data, headers)
...
 local function http_export(conf, spans)
   local start = ngx_now()
   local headers = get_cached_headers(conf.headers)
   local payload = encode_traces(spans, conf.resource_attributes)
 
-  local ok, err = http_export_request(conf, payload, headers)
+  -- caching because comparing string is much faster than applying regex
+  if conf.endpoint ~= prev_endpoint then
+    prev_endpoint = conf.endpoint
+    parsed_endpoint = string.gsub(conf.endpoint, ENV_IDENTIFIER_REGEX, os.getenv)
```
fffonion, line 93: unfortunately env vars are not available to read once Kong is started, so we can't really
have dynamic env var names. it can only be static like datadog plugin does.

jaswanthikolla: Yes, That's similar to any app, right? We need to set the environment variable before starting the process. Here, user will set the environment variables upfront in Kong's helm values.yaml with following syntax, and refer to them in the plugin conf
```
env:
  NGINX_MAIN_ENV: "KONG_HOST_IP"
  KONG_HOST_IP: 
      valueFrom:
      fieldRef:
        apiVersion: v1
        fieldPath: status.hostIP
```

fffonion: The problem here is nginx clears all environment variable after it's restarted (i.e. post the `init` phase). In regards to lua code, that means only `os.getenv` that's called at module level will return exepcted result. The approach in this PR that reads env var in a function, that's called during run-time, will not work.

jaswanthikolla: >  The approach in this PR that reads env var in a function, that's called during run-time, will not work.

@fffonion  I tested this PR plugin and  It's already working. It's runtime, and it's able to pick up the environment variable, `NGINX_MAIN_ENV` directive is to pass the environment variables to kong, right? 

What do you mean after the `restart`, it won't work?  Can you give me a scenario to reproduce that?

fffonion: That way it would work yes. But this also makes the design more hackier.
We appreciate your continous effort on pushing this forward, but I'm afraid these are not the propriate approach we feel would solve the issue properly as the reason I explained in https://github.com/Kong/kong/pull/9466#issuecomment-1451302916.

On the other side, if you would like to give a stab on making the datadog and otel plugin able to be use with secrets management, your contributions are always welcomed.

hbagdi: ping @fffonion

fffonion: @jaswanthikolla right, that wasn't the right/supported syntax for referencable values, consider using a different approach that does the templating at helm chart side and pass in the whole string in.

hbagdi:
@jaswanthikolla Could you please rebase?

hbagdi:

PR #10389, opened 2023-02-28
feat(db): add postgres triggers to cleanup expired rows on ttl-enabled schema tables

bungle:
### Summary

This is alternative to #10331. @windmgc please check this out. What do you think? If ok, please take over and do the rest of the changes (aka add migrations so that cleanup is in other entities too, similar to what I did for `oauth2_tokens` here).


hanshuebner:
This is a clever approach.  Are we fine with the random response time increase that we see for those insertions that perform a garbage collection?

hbagdi:
This programs the database in a way that is very hard to control when something goes wrong or needs tweaking.
Is creating such stored procedures and then invoking them from control-plane nodes a better balance? That helps us invoke the procedure on a schedule and connection that can be controlled by the CP and we (developers) can fine-tune that based on the feedback we receive.
Kong is sensitive to database latencies, especially when there is a short burst of write traffic, and such a change could potentially worsen database performance and increase the time it takes for the configuration of Kong to propagate across the cluster in hybrid mode.

I must say that the cleverness is appealing especially since it offloads all the work to the database server!

hanshuebner:
> Is creating such stored procedures and then invoking them from control-plane nodes a better balance? That helps us invoke the procedure on a schedule and connection that can be controlled by the CP and we (developers) can fine-tune that based on the feedback we receive.

I think that if we have a mechanism for that, scheduling the garbage collection from th CP instead of randomly invoking it from insertions is better.  The concern that I have is that when we'd piggyback the GC onto insertions, we'd see the additional cost as latency increases in our proxy path.  I'm sure that we have other non-obvious contributing factors, but I'd generally want to keep clear of adding cost to the proxy path that'd be very difficult to analyze when looking at the latency profile.

bungle:
> > Is creating such stored procedures and then invoking them from control-plane nodes a better balance? That helps us invoke the procedure on a schedule and connection that can be controlled by the CP and we (developers) can fine-tune that based on the feedback we receive.
> 
> I think that if we have a mechanism for that, scheduling the garbage collection from th CP instead of randomly invoking it from insertions is better.

All this needs to be measured. Perhaps we don't have any problems with the solution I am proposing. Perhaps having cluster of 50 data planes that are scheduled to bombard database is a way worse. Perhaps adding cluster mutex lock there only makes it more complex and complicated (the cluster mutex is still an additional query to database and then you need to deal with possible situations when lock is not released).

bungle:
> The concern that I have is that when we'd piggyback the GC onto insertions, we'd see the additional cost as latency increases in our proxy path.

Data planes and DBless proxy nodes don't insert anything. CP inserts and there we don't have the latency issue that much (on CPs insertions are a low factor). The only problem we have is OAuth2 plugin in TRADITIONAL mode.

windmgc:
I've been thinking about this the whole day, and so far I still have some concerns here:
- The backporting problem. I'm okay with the trigger way if we're going to fix it in a new version but it's hard to backport this to an older version when solving FTI problems. Manually creating triggers on the older versions seems not sustainable enough and might cause problems when upgrading to newer versions.
- The latency increase, which needs some test(haven't got time to do that yet, will do)

Do you think it is a good idea that we remove the `RANDOM` call and just `DELETE` two rows on every `INSERT`? Since we already have proper indexes

bungle:
@hanshuebner and @hbagdi, thanks for great feedback from @windmgc I modified this PR so that now on each insert statement the trigger removes maximum of 2 rows. I also removed usage of RANDOM. The penalty of deleting maximum of 2 rows per insert statement should be miniscule.

hanshuebner (approved):

hbagdi:
>  The penalty of deleting maximum of 2 rows per insert statement should be minuscule.


I have forever been told my DBA to not use triggers or use them minimally and be careful with them. I've never found a good reason when I push for why not.
Provided that the overhead of running on the trigger on every insert is low, we are good. If that is not the case, we may be adding cost. I'm out of depth with my knowledge of databases here.


> Inserting new access tokens and authorization codes is not that frequent. It usually means one per user per certain amount of time (e.g. session). If there are say 100-1000 logins per second (which is quite a bit already).

I could be wrong but isn't this change about all tables that have TTLs? At least that was what the original PR (#10331) was about.

> CP inserts and there we don't have the latency issue that much

Write traffic via CP is generally bursty in nature and latency differences do matter. I do see your point that the latency matters less.

Anyways, this change looks much better than it was before.
Should this move out from Draft at this point?

hanshuebner:
> I have forever been told my DBA to not use triggers or use them minimally and be careful with them. I've never found a good reason when I push for why not.
It is basically a matter of architectural choice whether you use or don't use triggers in a database.  Using them has benefits (fewer DB interactions with the application, business-level constraints can be enforced on all database clients) and downsides (cost of insertions/updates becomes non-obvious due to the triggered activity, hence making analysis and capacity planning more difficult, triggers are not portable across database systems).

It should be a conscious decision to use or not use triggers in an application.  It seems that we're already using them to maintain the tags table.  If that implementing that functionality using triggers was a concious decision, I see no reason why this cleanup task can't be implemented using triggers as well.

bungle:
> I could be wrong but isn't this change about all tables that have TTLs? At least that was what the original PR (#10331) was about.

Yes, we need to add triggers to all the tables where entity says `ttl=true` and to `cluster_events` too which does not have `ttl=true` but it has column `expire_at` (the `batch_delete_expired_rows` function expects `ttl` column, so for `cluster_events` I think we just re-create this or rename `cluster_events` column to `ttl`). I cannot think of any other entity than oauth2 plugins two entities that are constantly created by external users. Most other entities are admin managed, so those don't have this issue of potentially huge amounts of rows. The cluster events may also be plenty especially if creating oauth2 entities also creates cluster event. The OAuth 2.0 plugin is not supported in DP and DBless so its surface is smaller, but important to those that use it.

bungle:
One thing worth noting is that I am not sure if Kong cache is currently aware of entity `ttl`, but this problem is with current cleanup too. Aka when it cleans it will just delete rows on db, but cache of those entities is not invalidated.

windmgc:
Tables covered in this PR:
`cluster_events`, `clustering_data_planes`, `acme_storage`, `keyauth_credentials`, `oauth2_authorization_codes`, `oauth2_tokens`, `ratelimiting_metrics`, `sessions`

locao:
I labeled this as "do not merge" for now as the migrations tests are failing.

vm-001:
There may have a potential behavior change in the Database trigger solution.

Kong not only collects the TTL schemas from the DAO of bundled plugins, but also from the enabled custom plugins. If a custom plugin is relying on the built-in TTL scheduled task to clean up the expired data of itself, then it won't work like the previous. Although adding a plugin migration to add a trigger(like we're creating), or even creating a timer task to do so could avoid the behavior change, Both options require manual action. This may be something that needs to be considered.

vm-001:
There may have a potential behavior change in the Database trigger solution.

Kong not only collects the TTL schemas from the DAO of bundled plugins, but also from the enabled custom plugins. If a custom plugin is relying on the built-in TTL scheduled task to clean up the expired data of itself, then it won't work like the previous. Although adding a plugin migration to add a trigger(like we're creating), or even creating a timer task to do so could avoid the behavior change, Both options require manual action. This may be something that needs to be considered.

vm-001:
For brainstorming. I was thinking an alternative solution, which is still using the database trigger, and isn't triggered by the `INSERT` statement, but is triggered by `UPDATE` statement which is called by the previous timer. 

Let's say we have a new table, and a trigger triggered by the UPDATE statement. And also a function `delete_expired_rows` is called by trigger to clean up the expired data.

```sql
create table ttl_task (
    execute_time int,
    ttl_schemas text
);
insert into ttl_task values(0); -- a default row

-- create function delete_expired_rows
CREATE OR REPLACE FUNCTION delete_expired_rows() RETURNS TRIGGER
LANGUAGE plpgsql
AS $$
DECLARE
    v_table text;
BEGIN
    -- TODO "oauth2_tokens,oauth2_authorization_codes" should be fetched from ttl_task.ttl_schemas
    foreach v_table in array STRING_TO_ARRAY('oauth2_tokens,oauth2_authorization_codes', ',') loop
        EXECUTE format('delete from %s WHERE ttl < (CURRENT_TIMESTAMP AT TIME ZONE ''UTC'')' , v_table);
    end loop;
    return NULL;
END;
$$;
```


The postgres clean up timer can be simplified to the following without any lock or cluster mutex consideration. Because the database layer will take care of the rest.

```lua
return timer_every(60, function(premature)
    -- execute SQL
    -- sending a collected_ttl_schemas is for to be used by a trigger
    local collected_ttl_schemas = "oauth2_tokens,oauth2_authorization_codes,...."
    local sql = "update ttl_task set execute_time = floor(EXTRACT(epoch FROM current_timestamp)), ttl_schemas = " .. collected_ttl_schemas .. " where execute_time < floor(EXTRACT(epoch FROM current_timestamp)) - 60"
    execute_sql(sql)
end

```




Here is the whole sql script, which is very rough.

```sql
drop table if exists ttl_task;
create table ttl_task (
    execute_time int,
    ttl_schemas text
);
insert into ttl_task values(0); -- insert a default row


-- create function delete_expired_rows
CREATE OR REPLACE FUNCTION delete_expired_rows() RETURNS TRIGGER
LANGUAGE plpgsql
AS $$
DECLARE
    v_table text;
BEGIN
    -- TODO "oauth2_tokens,oauth2_authorization_codes" should be fetched from ttl_task.ttl_schemas
    foreach v_table in array STRING_TO_ARRAY('oauth2_tokens,oauth2_authorization_codes', ',') loop
        EXECUTE format('delete from %s WHERE ttl < (CURRENT_TIMESTAMP AT TIME ZONE ''UTC'')' , v_table);
    end loop;
    return NULL;
END;
$$;


-- define ttl_task_trigger
DROP TRIGGER IF EXISTS "ttl_task_trigger" ON "ttl_task";
DO $$
BEGIN
CREATE TRIGGER "ttl_task_trigger"
		AFTER UPDATE ON "ttl_task"
		FOR EACH ROW
		WHEN (OLD.execute_time <> NEW.execute_time)
		EXECUTE PROCEDURE delete_expired_rows();
EXCEPTION WHEN UNDEFINED_COLUMN OR UNDEFINED_TABLE THEN
        -- Do nothing, accept existing state
END$$;



delete from oauth2_tokens;
insert into oauth2_tokens(id, ttl) values('00000000-0000-0000-0000-000000000000', '2022-11-14 18:45:17');
insert into oauth2_tokens(id, ttl) values('00000000-0000-0000-0000-000000000001', '2022-11-14 18:45:17');

delete from oauth2_authorization_codes;
insert into oauth2_authorization_codes(id ,ttl) values('00000000-0000-0000-0000-000000000000', '2022-11-14 18:45:17');
insert into oauth2_authorization_codes(id ,ttl) values('00000000-0000-0000-0000-000000000001', '2022-11-14 18:45:17');

update ttl_task set execute_time = floor(EXTRACT(epoch FROM current_timestamp)), ttl_schemas = 'oauth2_tokens,oauth2_authorization_codes,not_exist' where execute_time < floor(EXTRACT(epoch FROM current_timestamp)) - 60;

select * from oauth2_tokens;
select * from oauth2_authorization_codes;


-- insert again
delete from oauth2_tokens;
insert into oauth2_tokens(id, ttl) values('00000000-0000-0000-0000-000000000000', '2022-11-14 18:45:17');
insert into oauth2_tokens(id, ttl) values('00000000-0000-0000-0000-000000000001', '2022-11-14 18:45:17');

-- insert again
delete from oauth2_authorization_codes;
insert into oauth2_authorization_codes(id ,ttl) values('00000000-0000-0000-0000-000000000000', '2022-11-14 18:45:17');
insert into oauth2_authorization_codes(id ,ttl) values('00000000-0000-0000-0000-000000000001', '2022-11-14 18:45:17');


-- run update again, and should not trigger the delete_expired_rows() function
update ttl_task set execute_time = floor(EXTRACT(epoch FROM current_timestamp)), ttl_schemas = 'oauth2_tokens,oauth2_authorization_codes,not_exist' where execute_time < floor(EXTRACT(epoch FROM current_timestamp)) - 60;
select count(*) from oauth2_tokens; -- should be 2
select count(*) from oauth2_authorization_codes; -- should be 2

```

windmgc:
@hanshuebner @bungle  Could you please review again? I've add the triggers that CE needs and fixed the migration test.

hbagdi:
Let's avoid the breaking change for custom plugins, our bar for breaking change is much higher and there are far too many custom plugins in the world and it is hard to know the impact of such a breaking change. And, I also don't see a big enough risk that keeping the old logic around.

Is it possible to check the DB for custom tables and then fire up the timer only if there are custom tables?

hbagdi:
> @windmgc could you check how possible that is? @hbagdi it is not that easy to detect whether custom table by a custom plugin has added a trigger or should it do timer cleanup. Perhaps we could still have timer (only on custom entities) as it just does almost nothing if there is a trigger.

This is exactly what I meant.

windmgc:
> Is it possible to check the DB for custom tables and then fire up the timer only if there are custom tables?

I'd like to keep all of the origin timer running in the background after consideration. It seems a bit hard to tell if the entity is a "custom" one, from Kong's perspective the "oauth2_tokens" is also a custom entity loaded by plugin. And it also seems not reliable to judge by `bundled_plugins` since the custom entity may not be provided by plugins. I can tell for sure if the entity we're cleaning is a core entity(based on `CORE_ENTITES`) but running the timer on non-core entities sounds the same as running it on all entities to me.

@bungle @hbagdi If you all agree with keeping the original timer code I'll go this way

bungle:
> > Is it possible to check the DB for custom tables and then fire up the timer only if there are custom tables?
> 
> I'd like to keep all of the original timer running in the background after consideration. It seems a bit hard to tell if the entity is a "custom" one, from Kong's perspective the "oauth2_tokens" is also a custom entity loaded by plugin. And it also seems not reliable to judge by `bundled_plugins` since the custom entity may not be provided by plugins. I can tell for sure if the entity we're cleaning is a core entity(based on `CORE_ENTITES`) but running the timer on non-core entities sounds the same as running it on all entities to me.
> 
> And `SKIP LOCKED` is really a good improvement here, it can prevent us from potentially conflicting when running multiple cleanup logic on one table. With batch query and `SKIP LOCKED` I think it should be relatively safe to keep both the trigger and the timer.
> 
> @bungle @hbagdi If you all agree with keeping the original timer code I'll go this way

Yes, I approve this approach. Now that we don't need it that much for core anymore, perhaps also change that it runs less often. I do not know what would be a good value. Once per 5 mins? Once per hour? But I feel once per minute is too much.

hbagdi:
I don't have strong opinions either way.
If making the timer run less frequently can result in a large behavior change then we should publicly document it. Otherwise, this is ok.

windmgc:
I think increasing it to per 5min is enough, and it won't cause a huge behavior change.

One thing to note that is from 3.X we replaced timer library to timer-ng which can also guarantee that, if a timer is created by `timer.every`, only one will run at the same time. Before 3.X we don't have such a guarantee.

windmgc:
@bungle  Could you please review it again?

I've restored the original timer code and increase the interval to 300s.

hbagdi:
@windmgc Please cherry pick this PR to EE.

hbagdi:
ping @windmgc

hbagdi:
ping @windmgc

outsinre:
As not yet cherry-picked, now covered by CE2EE master merge.

kikito:
@outsinre notice that this has now been cherrypicked to ee

outsinre:
@kikito noticed. Thanks to @bungle on the debug for CE2EE master merge.

PR #10390, opened 2023-02-28
fix(request-terminator): add the missing captures in the echo

Tieske:
### Summary

request terminator 'echo' option would not return the uri-captures. And when  it did, the array part was not returned as an array.

### DIFF
--- a/spec/03-plugins/14-request-termination/02-access_spec.lua
+++ b/spec/03-plugins/14-request-termination/02-access_spec.lua
@@ -343,6 +343,10 @@ for _, strategy in helpers.each_strategy() do
           },
           raw_body = 'cool body',
           scheme = 'http',
+          uri_captures = {
+            named = {},
+            unnamed = {}
+          },
         }, json.request)
       end)
       it("doesn't echo a request if the trigger is set but not specified", function()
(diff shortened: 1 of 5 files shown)


spec/03-plugins/14-request-termination/02-access_spec.lua
```diff
@@ -387,6 +391,10 @@ for _, strategy in helpers.each_strategy() do
           },
           raw_body = 'cool body',
           scheme = 'http',
+          uri_captures = {
+            named = {},
+            unnamed = {}
+          },
```
hbagdi, line 397: Can we please add a test that has this populated since we are adding a new field to the response?

Tieske: Thought about it, but didn’t. This test already proves that the output of the PDK function gets injected. The tests for the PDK function cover its output. So this is covered. Hence didn’t add it. 

Let me know if extra tests are needed.

hbagdi: I expected such a response.
One could replace the code in the plugin handler to output a couple of empty tables under the 'uri_captures' table key and no test would fail. I recommend adding a test or updating an existing test to logically cover as much of the code as possible.

Tieske: added

hbagdi (approved):
Need someone else with more experience with PDK to approve.

hanshuebner:
@jschmid1 Can you have a look, please?

jschmid1:
Scheduling for tomorrow

jschmid1 (approved):
my main concern would've been testing more complex capture groups within the plugin but I see that the PDK tests cover this scenario.

